PR #442, opened 2018-08-21
KIALI-1389 Fix service metric support

jshaughn:
- use source telemetry and add more label qualifiers to ensure only
  service metrics are retrieved.


jshaughn:
@jotak Please have a look when you get back from PTO, even if it is already merged.

lucasponce:
Adding DNM. Testing locally.

lucasponce:
As discussed, with this PR we miss the inbound traffic at service level for unknown sources:
image

So, perhaps we need to figure out in some solution as discussed in :
https://issues.jboss.org/browse/KIALI-1344

I understand that this could be tricky to automatically merge both kind of traffic (reporter="source", reporter="destination").

But also, in the other hand, from user perspective is going to be weird to see that there is traffic to productpage service but it is not collected at service level.

I don't know, I can't suggest any direction at implementation level, I understand this is not an easy one.

jshaughn:
Rats, yes, although the destination-side telemetry is not required for the graphing case, it is required for general service inbound-metric charts in the details side of the house.  Need to think more...

PR #445, opened 2018-08-21
KIALI-1379 Fix service metric support

jshaughn:
- update getMetrics to return both source and destination telemetry to allow the client to use what he needs.
- use source telemetry and add more label qualifiers to ensure only service metrics are retrieved.

### DIFF
--- a/handlers/apps_test.go
+++ b/handlers/apps_test.go
@@ -39,10 +39,10 @@ func TestAppMetricsDefault(t *testing.T) {
 		assert.Contains(t, query, "[1m]")
 		if strings.Contains(query, "histogram_quantile") {
 			// Histogram specific queries
-			assert.Contains(t, query, " by (le)")
+			assert.Contains(t, query, " by (le,reporter)")
 			atomic.AddUint32(&histogramSentinel, 1)
 		} else {
-			assert.NotContains(t, query, " by ")
+			assert.Contains(t, query, " by (reporter)")
 			atomic.AddUint32(&gaugeSentinel, 1)
 		}
 		assert.Equal(t, 15*time.Second, r.Step)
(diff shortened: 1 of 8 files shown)


jshaughn:
@lucasponce, @jotak ,  This PR makes an API change based on our earlier discussion and the [resulting Jira comment[(https://issues.jboss.org/browse/KIALI-1344?focusedCommentId=13622180&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13622180)

The basic idea is that GetMetrics now returns both Source and Destination telemetry and the caller can then use what is appropriate given his situation.  In the graph we can now special case "unknown" nodes and service nodes, etc.  In various detail charts it should now be possible to show two lines is desired, one for each telemetry source.

Let me know what you think, or if you can see a hole in the idea, but I'm fairly jazzed with this approach because it still performs it does not increase the number of prom queries and avoids any sort of horrible data merge logic.

jshaughn:
There is also a sibling UI PR that must be used if you want to run this server version. See https://github.com/kiali/kiali-ui/pull/605

jshaughn:
I'm actually going to remove the DNM label here.  I've been using this code and updating the UI, especially in the graph side panels, to make use of the dual telemetry.  I have not played much with the detail view charts but in a some cursory checks they looked good to me, and in some cases we could enhance them to also use the dual telemetry.

lucasponce:
I will add back the DNM as this PR needs a counterpart in the detail pages.
As it throws the error, so, UI would need a PR and all should be reviewed/merged in a similar timeline.
image

lucasponce:
@jshaughn You are right, I was testing a wrong UI PR.
Now things look much better:
image

PR #446, opened 2018-08-22
[KIALI-1373] Revert "Moving ImagePullPolicy Always (#441)"

gbaufake:
This reverts commit bc3934ce5e97c5636e5238585bba81285f433849 and removes `imagepullpolicy:always` from deployment template.

Best Regards,
Guilherme Baufaker Rêgo


mwringe:
Approved. This might not be the final solution we are doing to go with, but for now we should revert until we figure things out a bit better.

PR #447, opened 2018-08-22
KIALI-1373 Accommodating different types of pull policy for different needs

gbaufake:
Hello everyone,

Considering the situation presented by @lucasponce on https://github.com/kiali/kiali/pull/441#issuecomment-414937306, using the pull policy always as default is not the best solution for developer side.

On this PR, I've made IfNotpresent as the default option to use on makefile (like it was before the inclusion of pull policy) but also will include the possibility for different users (istio ansible playbooks, qe automation) to use always to ensure the image is always the latest from that build.

Best Regards,
Guilherme Baufaker Rẽgo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,18 @@ DOCKER_NAME ?= kiali/kiali
 DOCKER_VERSION ?= dev
 DOCKER_TAG = ${DOCKER_NAME}:${DOCKER_VERSION}
 
+
+ifndef PULL_POLICY
+	ifeq ("${DOCKER_VERSION}", "latest")
+		PULL_POLICY = "Always"
+	endif
+
+	ifeq ("${DOCKER_VERSION}", "dev")
+		PULL_POLICY = "IFNotPresent"
+	endif
+endif
+
+
 # Indicates the log level the app will use when started.
 # <4=INFO
 #  4=DEBUG
@@ -118,7 +130,7 @@ test-race:
 	@echo Running tests with race detection, excluding third party tests under vendor
 	go test -race $(shell go list ./... | grep -v -e /vendor/)
 
-## test-e2e-setup: Setup Python environment for running test suite 
+## test-e2e-setup: Setup Python environment for running test suite
 test-e2e-setup:
 	@echo Setting up E2E tests
 	cd tests/e2e && ./setup.sh
@@ -235,7 +247,7 @@ openshift-deploy: openshift-undeploy
 	@echo Deploying to OpenShift project ${NAMESPACE}
 	cat deploy/openshift/kiali-configmap.yaml | VERSION_LABEL=${VERSION_LABEL} envsubst | ${OC} create -n ${NAMESPACE} -f -
 	cat deploy/openshift/kiali-secrets.yaml | VERSION_LABEL=${VERSION_LABEL} envsubst | ${OC} create -n ${NAMESPACE} -f -
-	cat deploy/openshift/kiali.yaml | IMAGE_NAME=${DOCKER_NAME} IMAGE_VERSION=${DOCKER_VERSION} NAMESPACE=${NAMESPACE} VERSION_LABEL=${VERSION_LABEL} VERBOSE_MODE=${VERBOSE_MODE} envsubst | ${OC} create -n ${NAMESPACE} -f -
+	cat deploy/openshift/kiali.yaml | IMAGE_NAME=${DOCKER_NAME} IMAGE_VERSION=${DOCKER_VERSION} NAMESPACE=${NAMESPACE} VERSION_LABEL=${VERSION_LABEL} VERBOSE_MODE=${VERBOSE_MODE} PULL_POLICY=${PULL_POLICY} envsubst | ${OC} create -n ${NAMESPACE} -f -
 
 ## openshift-undeploy: Undeploy from Openshift project.
 openshift-undeploy: .openshift-validate
@@ -257,7 +269,7 @@ k8s-deploy: k8s-undeploy
 	@echo Deploying to Kubernetes namespace ${NAMESPACE}
 	cat deploy/kubernetes/kiali-configmap.yaml | VERSION_LABEL=${VERSION_LABEL} envsubst | ${KUBECTL} create -n ${NAMESPACE} -f -
 	cat deploy/kubernetes/kiali-secrets.yaml | VERSION_LABEL=${VERSION_LABEL} envsubst | ${KUBECTL} create -n ${NAMESPACE} -f -
-	cat deploy/kubernetes/kiali.yaml | IMAGE_NAME=${DOCKER_NAME} IMAGE_VERSION=${DOCKER_VERSION} NAMESPACE=${NAMESPACE} VERSION_LABEL=${VERSION_LABEL} VERBOSE_MODE=${VERBOSE_MODE} envsubst | ${KUBECTL} create -n ${NAMESPACE} -f -
+	cat deploy/kubernetes/kiali.yaml | IMAGE_NAME=${DOCKER_NAME} IMAGE_VERSION=${DOCKER_VERSION} NAMESPACE=${NAMESPACE} VERSION_LABEL=${VERSION_LABEL} VERBOSE_MODE=${VERBOSE_MODE} PULL_POLICY=${PULL_POLICY} envsubst | ${KUBECTL} create -n ${NAMESPACE} -f -
 
 ## k8s-undeploy: Undeploy docker image in Kubernetes namespace.
 k8s-undeploy: .k8s-validate
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ GO_VERSION_KIALI = 1.8.3
 
 # Identifies the docker image that will be built and deployed.
 DOCKER_NAME ?= kiali/kiali
-DOCKER_VERSION ?= dev
+DOCKER_VERSION ?= latest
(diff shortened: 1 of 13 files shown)


mwringe:
-1

Just remove the image pull policy and use the defaults. This is probably the more expected behaviour.

This means:
If the tag is 'latest' it will use an imagepullpolicy of 'always'. This allows people to use latest and easily get an updates (it also useful if you are running on a cluster and you want to pull in latest for testing purposes, otherwise it will never go out and pull in a newer version)

If the tag is not 'latest' then it will use an imagepullpolicy of 'ifnotpresent'. This means for everyone not using latest they won't pull in a new image if they already have it. This is the expected and wanted behaviour if the image is tagged for a specific version and is not expected to change.

gbaufake:
Some of the concerns that PR try to address:

1- If we move to a branch like we had in the pre-release scenario.
Images were not updated as it should because we had the only version, this might lead to a problem on openshift with more than one node (you would need to delete manually image on this nodes, then pull it again)

2 - It won't impact developers work which are using `dev` tag and not uploading it docker hub because the default would be `IfNotPresent`

3 - It will help QE because scenario 1 - (we had to look on our every cluster that we had and manually confirm that image was the desirable one).

4 - It allows us to use `always` for Istio Ansible and Helm Installers. That will guarantee that images released and by somehow weren't not good, the images will pull out preventing some deprecated bugs reports. 

5 - If do find a major bug, we can easily re-tag that image and everyone will have the correct image like we had last week with istio-1.0 image.

6 - On the latest Scenario, you would just switch `always` and then you have the latest docker images.


Best Regards,
Guilherme Baufaker Rêgo

mwringe:
I would rather have this PR do the following

- if a PULL_POLICY is set, then use that pull policy
- if no PULL_POLICY is set and the version matches vX.Y then set the imagePullPolicy to always

For everything else we should use the default. Eg
latest -> will use the default of 'always'
dev -> will use the default of 'ifNotPresent'
vX.Y.Z -> will use the default of 'ifNotPresent'

Makefile
```diff
@@ -30,6 +30,10 @@ DOCKER_NAME ?= kiali/kiali
 DOCKER_VERSION ?= dev
 DOCKER_TAG = ${DOCKER_NAME}:${DOCKER_VERSION}
 
+
+# Change that IfNotPresent to use on dev mode
```
jmazzitelli: I'm not entirely sure what this comment is saying... do you mean "Change the Always below to IfNotPresent when in dev mode" or something like that?

In other words, "Change that IfNotPresent" is confusing because there is no IfNotPresent here, so what is "that IfNotPresent" referring to?

jmazzitelli: BTW: if this does mean "PULL POLICY should be set to IfNotPresent if in dev mode" then we can add a if-else here - look at the value of DOCKER_VERSION and if "dev" then we should set the PULL_POLICY (unless PULL_POLICY is already set - in which case use it since the user is telling us specifically what value to use).

jmazzitelli:
The change to the yaml here will require us to make additional changes in other places:

1) we need to change the Istio helm chart/ansible scripts
2) we need to change the getting started docs to tell the user to pass in the PULL_POLICY on the command line - see here: https://www.kiali.io/gettingstarted/ in the "Install Kiali" and the "Getting started on Kubernetes" sections.

gbaufake:
Thanks @jmazzitelli.

I'll address all your comments on the following changes.

gbaufake:
@mwringe  @jmazzitelli 
I know that 01a3f6a is not everything that you commented on last reviews.

it addresses:
-  if a PULL_POLICY is set, then use that pull policy
- latest -> will use the default of 'always'
- dev -> will use the default of 'ifNotPresent'

I didn't find a "safe way" to guarantee with regex that (vX.Y.Z -> will use the default of 'ifNotPresent' and  vX.Y then set the imagePullPolicy to always) but I don't have too much experience with these makefiles.

If this PR is not needed anymore (or too complicated to be merged as it is - or too many people on -1), I will close it and find a way to patch the files.

Best Regards,
Guilherme Baufaker Rẽgo

jmazzitelli:
I think this PR looks ok except for the nested ifeq inside the ifndef. Did you test all the combinations to ensure that all works? I thought you could not indent ifndef/ifeq conditionals like that (i.e. they all had to be on column 1, even if you nested an ifeq in another ifndef). Anyway, I didn't test it - but if it works for you, then I'm ok with it. But for sure, make sure all that ifeq/ifndef stuff works.

> I didn't find a "safe way" to guarantee with a regex that (vX.Y.Z -> will use the default of 'ifNotPresent' and vX.Y then set the imagePullPolicy to always) but I don't have too much experience with these makefiles.

As for the above,  create another JIRA and assign it to me for this "regex to look for vX.Y.Z or v.X.Y" and I can look at it. I have an idea how it can be done.

mwringe:
I think I have it working on my machine, I will submit a PR for it.

mwringe:
See https://github.com/kiali/kiali/pull/470

jmazzitelli:
@mwringe is #470 a replacement for this PR? i.e. do we close this PR in favor of #470?

gbaufake:
@mwringe  and @jmazzitelli 

I merged #470 with #473 on the same file on c6acbb7d

Can you review it?

Makefile
```diff
@@ -27,9 +27,31 @@ GO_VERSION_KIALI = 1.8.3
 
 # Identifies the docker image that will be built and deployed.
 DOCKER_NAME ?= kiali/kiali
-DOCKER_VERSION ?= dev
+DOCKER_VERSION ?= latest
```
jmazzitelli: do we want to default to "latest" here?

Devs are the ones normally working with the Makefile, so I assume we want to keep this at "dev". CIs will normally be passing in the DOCKER_VERSION they want, thus they will override this default.

deploy/openshift/kiali.yaml
```diff
@@ -83,6 +83,7 @@ spec:
       serviceAccount: kiali
       containers:
       - image: ${IMAGE_NAME}:${IMAGE_VERSION}
+        imagePullPolicy: ${PULL_POLICY}
```
jmazzitelli: I think we want what Matt did before - he had something like ${IMAGE_PULL_POLICY} whose value is the entire string.

This way, if you don't set it, this is just empty and will just pick up the k8s default (and thus we are backwards compatible with our current getting started instructions and we don't have to worry about helm chart/ansible updates right now.

Makefile
gbaufake: ah... sorry

deploy/kubernetes/kiali.yaml
```diff
@@ -81,6 +81,7 @@ spec:
       serviceAccount: kiali
       containers:
       - image: ${IMAGE_NAME}:${IMAGE_VERSION}
+        imagePullPolicy: ${PULL_POLICY}
```
jmazzitelli: see my comment below about the image pull policy.

deploy/openshift/kiali.yaml
jmazzitelli: see https://github.com/kiali/kiali/pull/470/files#diff-47101f922225a5e44356824f41906c92R86

jmazzitelli:
OK, I just tested this and here is what I see - you tell me if this is what we want:

DOCKER_VERSION=latest
        imagePullPolicy: Always

DOCKER_VERSION=dev
        imagePullPolicy: IfNotPresent

DOCKER_VERSION=v0.7
        imagePullPolicy: Always

DOCKER_VERSION=v0.7.0
       (not set in the yaml)

Makefile
```diff
@@ -30,6 +30,30 @@ DOCKER_NAME ?= kiali/kiali
...
+		IMAGE_PULL_POLICY = Always
+	endif
+
+	ifeq ("${DOCKER_VERSION}", "latest")
+		IMAGE_PULL_POLICY = Always
+	endif
+
+	ifeq ("${DOCKER_VERSION}", "dev")
+		IMAGE_PULL_POLICY = IfNotPresent
+	endif
```
mwringe: We don't need to check for 'latest' or 'dev' here. Those are the defaults.

jmazzitelli: I wondered about that - the docs indicate the default for "latest" as "Always" is deprecated:

https://kubernetes.io/docs/concepts/configuration/overview/#container-images

`An alternative, but deprecated way to have Kubernetes always pull the image is to use the :latest tag, which will implicitly set the imagePullPolicy to Always.`

so I wonder if we should keep this and explicitly set it to avoid relying on deprecated features?

Makefile
```diff
@@ -30,6 +30,30 @@ DOCKER_NAME ?= kiali/kiali
...
+	SHORT_VERSION=$(shell if [[ ${DOCKER_VERSION} =~ ^v[0-9]+\.[0-9]+$$ ]]; then echo true; fi)
+	ifeq (${SHORT_VERSION}, true)
+		IMAGE_PULL_POLICY = Always
+	endif
+
+	ifeq ("${DOCKER_VERSION}", "latest")
+		IMAGE_PULL_POLICY = Always
+	endif
+
+	ifeq ("${DOCKER_VERSION}", "dev")
```
jmazzitelli: This check we can remove because "IfNotPresent" is the default policy (and that is not a deprecated feature).

Makefile
mwringe: oh, interesting. Ok, I guess this is good then

PR #44, opened 2018-02-18
fail the openshift-deploy target if the namespace doesn't exist

jmazzitelli:

PR #36, opened 2018-02-17
install istio into openshift

jmazzitelli:
I still can't get prometheus to run successfully. Getting:

level=error
ts=2018-02-17T02:10:53.269935415Z
caller=main.go:323
msg="Opening storage failed"
err="mkdir data/: permission denied"

PR #35, opened 2018-02-16
script to install istio into openshift

jmazzitelli:

PR #21, opened 2018-02-14
remove the unnecessary "pkg" package level.

jmazzitelli:
Everything is a package, no need to have "pkg" at the top. It is superfluous.

Also, run gofmt - these files weren't formatted originally

PR #448, opened 2018-08-22
KIALI-616 Support limited access to namespaces

mwringe:
** Describe the change **

Allows user to run Kiali without granting them cluster wide permissions.

A user will be able to remove the cluser-wide 'kiali' role and the be able to select which namespaces it wants Kiali to have access to.

From the console, it will just appear like that namespace doesn't exist in the OpenShift cluster.

Caveats
- this only affects access to namespaces using the k8s RBAC. Kiali still has access to all the metrics and we don't do anything new in the PR to prevent access to these metrics. The purpose of this is to reduce and control access Kiali has to k8s objects.
- this only really works with Openshift. There really isn't a way in k8s to ask what namespaces a user has access to. Once k8s has that ability it should be easy to add it in. But for now its really only for Openshift (or if someone were to run https://github.com/openshift/kube-projects but I don't think that is being maintained anymore).

How to use (before we have official docs for it)
1) install Kiali as normal on OpenShift
2) remove the clusterrolebinding for Kiali (eg `oc delete clusterrolebinding kiali`)
3) grant permissions to Kiali only for specific namespaces (eg `oc adm policy add-role-to-user kiali system:serviceaccount:istio-system:kiali -n $NAMESPACE_NAME`)
4) Notice that only the namespaces the user has access to show up in the Kiali console.

Short Gif: we can see that there is a few namespaces in this cluster but the cluster role has been removed so we don't see them namespace drop down. We then grant the role in various namespaces and the namespace data starts to populate the dropdown and the graph. This works the same way in the details pages.

kiali-namespaces-2

** Issue reference **

https://issues.jboss.org/browse/KIALI-616

** Backwards incompatible? **

There shouldn't be any perceivable change to the user if they use the default cluster wide permissions.

If they decide to remove the cluster wide permissions, then namespaces they don't have access to anymore wont show up in the console.

### DIFF
--- a/graph/cytoscape/cytoscape.go
+++ b/graph/cytoscape/cytoscape.go
@@ -47,6 +47,7 @@ type NodeData struct {
 	HasVS           bool            `json:"hasVS,omitempty"`           // true (has route rule) | false
 	IsDead          bool            `json:"isDead,omitempty"`          // true (has no pods) | false
 	IsGroup         string          `json:"isGroup,omitempty"`         // set to the grouping type, current values: [ 'version' ]
+	IsInaccessible  bool            `json:"isInaccessible,omitempty"`  // true if the node exists in an inaccessible namespace
 	IsMisconfigured string          `json:"isMisconfigured,omitempty"` // set to misconfiguration list, current values: [ 'labels' ]
 	IsOutside       bool            `json:"isOutside,omitempty"`       // true | false
 	IsRoot          bool            `json:"isRoot,omitempty"`          // true | false
@@ -175,6 +176,11 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
 		// node may be unused
 		if val, ok := n.Metadata["isUnused"]; ok {
 			nd.IsUnused = val.(bool)
+		}
+
+		// node is not accessible to the current user
+		if val, ok := n.Metadata["isInaccessible"]; ok {
+			nd.IsInaccessible = val.(bool)
 		}
 
 		// node may have a circuit breaker
@@ -369,6 +375,12 @@ func groupByVersion(nodes *[]*NodeWrapper) {
 				if n.HasVS {
 					n.HasVS = false
 					hasVirtualService = true
+				}
+
+				// If we have any nodes which are inaccessible, then mark the group as being inaccessible
+				// Note: all nodes here would have the same inaccessible value since they all belong to the same namespace
+				if n.IsInaccessible {
+					nd.IsInaccessible = true
 				}
 			}
 
--- a/graph/options/options.go
+++ b/graph/options/options.go
@@ -2,6 +2,8 @@
 package options
 
 import (
+	"errors"
+	"fmt"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -12,7 +14,7 @@ import (
 
 	"github.com/kiali/kiali/graph"
 	"github.com/kiali/kiali/graph/appender"
-	"github.com/kiali/kiali/services/models"
+	"github.com/kiali/kiali/services/business"
 )
 
 const (
@@ -45,11 +47,12 @@ type VendorOptions struct {
 
 // Options are all supported graph generation options.
 type Options struct {
-	Appenders          []appender.Appender
-	Duration           time.Duration
-	IncludeIstio       bool // include istio-system services. Ignored for istio-system ns. Default false.
-	InjectServiceNodes bool // inject destination service nodes between source and destination nodes.
-	Namespaces         []string
+	AccessibleNamespaces map[string]bool
+	Appenders            []appender.Appender
+	Duration             time.Duration
+	IncludeIstio         bool // include istio-system services. Ignored for istio-system ns. Default false.
+	InjectServiceNodes   bool // inject destination service nodes between source and destination nodes.
+	Namespaces           []string
 	NodeOptions
 	QueryTime int64 // unix time in seconds
 	Workload  string
@@ -78,24 +81,32 @@ func NewOptions(r *http.Request) Options {
 	namespaces := params.Get("namespaces") // csl of namespaces. Overrides namespace path param if set
 	vendor := params.Get("vendor")
 
+	accessibleNamespaces := getAccessibleNamespaces()
+
 	var namespaceNames []string
 	fetchNamespaces := namespaces == NamespaceAll || (namespaces == "" && (namespace == NamespaceAll))
 	if fetchNamespaces {
-		namespaces, err := models.GetNamespaces()
-		checkError(err)
-
-		for _, namespace := range namespaces {
+		for namespace, _ := range accessibleNamespaces {
 			// The istio-system namespace is only shown when explicitly requested. Note that the
 			// 'includeIstio' option doesn't apply here, that option affects what is done in
 			// non-istio-system namespaces.
-			if namespace.Name != NamespaceIstioSystem {
-				namespaceNames = append(namespaceNames, namespace.Name)
+			if namespace != NamespaceIstioSystem {
+				namespaceNames = append(namespaceNames, namespace)
 			}
 		}
 	} else if namespaces != "" {
 		namespaceNames = strings.Split(namespaces, ",")
+		for _, namespaceName := range namespaceNames {
+			if _, found := accessibleNamespaces[namespaceName]; !found {
+				checkError(errors.New(fmt.Sprintf("Requested namespace [%s] is not accessible.", namespaceName)))
+			}
+		}
 	} else if namespace != "" {
-		namespaceNames = []string{namespace}
+		if _, found := accessibleNamespaces[namespace]; !found {
+			checkError(errors.New(fmt.Sprintf("Requested namespace [%s] is not accessible.", namespace)))
+		} else {
+			namespaceNames = []string{namespace}
+		}
 	} else {
 		// note, some global handlers do not require any namespaces
 		namespaceNames = []string{}
@@ -124,12 +135,13 @@ func NewOptions(r *http.Request) Options {
 	}
 
 	options := Options{
-		Duration:           duration,
-		IncludeIstio:       includeIstio,
-		InjectServiceNodes: injectServiceNodes,
-		Namespaces:         namespaceNames,
-		QueryTime:          queryTime,
-		Vendor:             vendor,
+		AccessibleNamespaces: accessibleNamespaces,
+		Duration:             duration,
+		IncludeIstio:         includeIstio,
+		InjectServiceNodes:   injectServiceNodes,
+		Namespaces:           namespaceNames,
+		QueryTime:            queryTime,
+		Vendor:               vendor,
 		NodeOptions: NodeOptions{
 			App:      app,
 			Service:  service,
@@ -199,10 +211,31 @@ func parseAppenders(params url.Values, o Options) []appender.Appender {
 		appenders = append(appenders, appender.IstioAppender{})
 	}
 	if csl == AppenderAll || strings.Contains(csl, "sidecars_check") {
-		appenders = append(appenders, appender.SidecarsCheckAppender{})
+		appenders = append(appenders, appender.SidecarsCheckAppender{
+			AccessibleNamespaces: o.AccessibleNamespaces,
+		})
 	}
 
 	return appenders
+}
+
+// getAccessibleNamespaces returns a Set of all namespaces accessible to the user.
+// The Set is implemented using the map[string]bool convention.
+func getAccessibleNamespaces() map[string]bool {
+	// Get the namespaces
+	business, err := business.Get()
+	checkError(err)
+
+	namespaces, err := business.Namespace.GetNamespaces()
+	checkError(err)
+
+	// Create a map to store the namespaces
+	namespaceMap := make(map[string]bool)
+	for _, namespace := range namespaces {
+		namespaceMap[namespace.Name] = true
+	}
+
+	return namespaceMap
 }
 
 func checkError(err error) {
--- a/handlers/graph.go
+++ b/handlers/graph.go
@@ -94,16 +94,19 @@ func graphNamespaces(o options.Options, client *prometheus.Client) graph.Traffic
 	// we can make some final adjustments:
 	// - mark the outsiders (i.e. nodes not in the requested namespaces)
 	// - mark the insider traffic generators (i.e. inside the namespace and only outgoing edges)
-	markOutsiders(trafficMap, o.Namespaces)
+	markOutsiders(trafficMap, o)
 	markTrafficGenerators(trafficMap)
 
 	return trafficMap
 }
 
-func markOutsiders(trafficMap graph.TrafficMap, namespaces []string) {
+func markOutsiders(trafficMap graph.TrafficMap, o options.Options) {
 	for _, n := range trafficMap {
-		if isOutside(n, namespaces) {
+		if isOutside(n, o.Namespaces) {
 			n.Metadata["isOutside"] = true
+			if isInaccessible(n, o.AccessibleNamespaces) {
+				n.Metadata["isInaccessible"] = true
+			}
 		}
 	}
 }
@@ -118,6 +121,14 @@ func isOutside(n *graph.Node, namespaces []string) bool {
 		}
 	}
 	return true
+}
+
+func isInaccessible(n *graph.Node, accessibleNamespaces map[string]bool) bool {
+	if _, found := accessibleNamespaces[n.Namespace]; !found {
+		return true
+	} else {
+		return false
+	}
 }
 
 func markTrafficGenerators(trafficMap graph.TrafficMap) {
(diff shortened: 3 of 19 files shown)


mwringe:
Note: the Kiali UI should work as is right now, except if you look in the console you will see some errors about not being able to fetch health for some cross namespace nodes if you don't have permission in the other namespace (I have a PR that skills this check on nodes with the 'inaccessible' attribute).

If you click on a pentagon that is outside of a namespace you have access to, you will still try and load the graph for that namespace but will receive an error message about access being denied.

Things seem to just work in the details view, but I have tested it less there, there might be some corner cases I have not hit yet.

mwringe:
Rebased to master

hhovsepy:
kialiqeci

lucasponce:
I will fetch and review it locally too.

hhovsepy:
it needs to be rebased with master to include ssl support

graph/graph.go
```diff
@@ -22,15 +22,16 @@ const (
...
+	ID         string                 // unique identifier for the node
+	NodeType   string                 // Node type
+	Namespace  string                 // Namespace
+	Workload   string                 // Workload (deployment) name
+	App        string                 // Workload app label value
+	Version    string                 // Workload version label value
+	Service    string                 // Service name
+	Edges      []*Edge                // child nodes
+	Metadata   map[string]interface{} // app-specific data
+	Accessible bool                   // If the node is accessible or not
```
lucasponce: In graph/cytoscape/cytoscape.go you use the "Inaccessible" semantic and here the positive one.
It sounds to me better to use same semantic in both places.

services/business/namespaces.go
```diff
@@ -0,0 +1,34 @@
...
+
+// Returns a list of the given namespaces / projects
+func (in *NamespaceService) GetNamespaces() ([]models.Namespace, error) {
+	// First try and see if we are running on OpenShift of Kubernetes with the kube-project component.
+	// This will allow us to fetch the namespaces that only our serviceaccount has permission to access.
+	//
+	// If we encounter an error here, we are assuming its because the project.openshift.io component is unknown
+	// and it means we are running in k8s without kube-projet. In that case, get the namespace list from Kubernetes
+	// (which requires a higher permission level)
+	projects, err := in.k8s.GetProjects()
```
lucasponce: This logic is too aggressive to be queried always IMO.
I mean, perhaps we can perform a 1-time GetProjects() when Kiali startup to check if it is under OpenShift or Kubernetes.
Falling into two queries per k8s scenarios sounds not necessary as it is something is not going to change between queries.

kubernetes/kubernetes_service.go
```diff
@@ -65,6 +67,19 @@ func (in *IstioClient) GetNamespaces() (*v1.NamespaceList, error) {
 	return namespaces, nil
 }
 
+func (in *IstioClient) GetProjects() (*osv1.ProjectList, error) {
+
+	result := &osv1.ProjectList{}
+
+	err := in.k8s.RESTClient().Get().Prefix("apis", "project.openshift.io", "v1", "projects").Do().Into(result)
```
lucasponce: This is plain/raw REST.
I suggest to add this API into the os_route_client.go as it is also used to queried OpenShift specific API, just for consistency.
Perhaps os_route_client.go is more generic and can be refactored to os_client.go.

jshaughn:
I added the DNM label as I'm just seeing this PR and am in the middle of a review.

graph/options/options.go
```diff
@@ -64,7 +64,10 @@ func NewOptions(r *http.Request) Options {
 	var namespaceNames []string
 	fetchNamespaces := namespaces == NamespaceAll || (namespaces == "" && (namespace == NamespaceAll))
```
jshaughn: I'm not exactly sure where is the best place to put my overall review comment so I'll put it here.  First let me make sure I understand what we're trying to do.  I think what this PR wants to do is two things: first, if a caller requests an inaccessible namespace return an error.  Second, for [outsider] nodes in inaccessible namespaces, flag the node as inaccessible so that the UI can handle it correctly.  If that is a valid synopsis I think the impl can be simplified to be a lot less intrusive. 

I'd suggest the following:

options.go is responsible for the up front processing of the requested graph options.  It's a good place to determine the accessible namespaces and validate the requested namespaces.  I'd suggest always fetching the accessible namespaces.  For 'all' namespace graphs, you'd then use that list.  If the user requested explicit namespaces then do the validate here and return an error immediately if a requested namespace is inaccessible.  We already have Options.Namespaces to list the namespaces to include in the graph,  Add a new field, Options.AccessibleNamespaces that provides the full list of accessible namespaces.  If we do this the graph.go code can assume that Options.Namespaces has only accessible namespaces.  And it is provided the accessible namespaces to be used to flag outsider nodes.

Then, generate the graph normally, just like before.  No need for new parameters or queries for accessible namespaces.

Finally, after all of the graph generation is done, add some code, probably in the graph.isOutside() function, to set the isAccessible flag.  If I understand correctly, only outsider nodes could be inaccessible as we are guaranteed that the "insider" nodes are already for accessible namespaces.

Maybe I am misunderstanding, but I think if you take this approach you can eliminate a lot of the changes that added complexity to this PR, and leave the main graph generation logic untouched.

services/business/namespaces.go
jshaughn: +1, that check should be done one time most likely.

kubernetes/kubernetes_service.go
mwringe: Yeah, that makes sense, the os_route_client.go is a bit messy as it deals with plain json and casting things to interfaces.

graph/cytoscape/cytoscape.go
```diff
@@ -51,6 +51,7 @@ type NodeData struct {
 	IsOutside       bool            `json:"isOutside,omitempty"`       // true | false
 	IsRoot          bool            `json:"isRoot,omitempty"`          // true | false
 	IsUnused        bool            `json:"isUnused,omitempty"`        // true | false
+	Inaccessible    bool            `json:"inaccessible,omitempty"`    // true if the node exists in an inaccessible namespace
```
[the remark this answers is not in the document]
mwringe: 'IsInaccessible' might be better.

Our convention is also not to add something to the json if its false (eg we don't have 'IsRoot:false' on any of the nodes, only when its true do we mention this)

graph/cytoscape/cytoscape.go
```diff
@@ -177,6 +178,11 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
 			nd.IsUnused = val.(bool)
 		}
 
+		// node is not accessible to the current user
+		if val, ok := n.Metadata["inaccessible"]; ok {
```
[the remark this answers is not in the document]
mwringe: See above, I think it should be 'IsInaccessible' or something like that, to have the convention where we only add the tag if its true (and most of the cases the node will be accessible, so I don't know if we want to add it on all nodes, but I guess that is a possibility).

graph/cytoscape/cytoscape.go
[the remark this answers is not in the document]
jshaughn: Good point, we only include true bools.  IsInaccessible is a little long but OK if you prefer it. Other possibilities: isLocked, isRestricted, isUnauthorized....

mwringe:
Updated, we still need to move all the OpenShift stuff out on its own and out of the kubernetes parts, but I think that can be done in another pr (this is already touching a lot of stuff).

graph/options/options.go
```diff
@@ -179,6 +180,24 @@ func parseAppenders(params url.Values, o Options) []appender.Appender {
 	return appenders
 }
 
+// Gets a map containing all the namespaces available
```
jshaughn: The Go convention is to lead the comment with the function name.  Also, maybe describe why this is a map (I assume it's to act like a Set):
`// getAccessibleNamespaces returns a Set of all namespaces accessible to the user. The Set is implemented using the map[string]bool convention.`

graph/options/options.go
```diff
@@ -61,18 +62,17 @@ func NewOptions(r *http.Request) Options {
...
 			// The istio-system namespace is only shown when explicitly requested. Note that the
 			// 'includeIstio' option doesn't apply here, that option affects what is done in
 			// non-istio-system namespaces.
-			if namespace.Name != NamespaceIstioSystem {
-				namespaceNames = append(namespaceNames, namespace.Name)
+			if namespace != NamespaceIstioSystem {
+				namespaceNames = append(namespaceNames, namespace)
 			}
 		}
 	} else if namespaces != "" {
```
jshaughn: After the namespaceNames array is set you should ensure that all of the names are in the accessibleNamespaces set.  If not fail fast like:

`checkError(errors.New(fmt.Sprintf("Requested namespace [%s] is not accessible", namespaceName))`

By validating the requested namespaces here the graph generation code does not have to worry about it at all.

handlers/graph.go
```diff
@@ -81,9 +81,15 @@ func graphNamespaces(o options.Options, client *prometheus.Client) graph.Traffic
 
 	trafficMap := graph.NewTrafficMap()
 	for _, namespace := range o.Namespaces {
+		// If we don't have access to the namespaces being requested then return an error.
```
jshaughn: Don't do this check here, handle it in the options processing.  This code should be able to assume that the requested namespaces are valid and just focus on graph generation, not security concerns.

handlers/graph.go
```diff
@@ -81,9 +81,15 @@ func graphNamespaces(o options.Options, client *prometheus.Client) graph.Traffic
...
+		if _,found := o.AccessibleNamespaces[namespace]; !found {
+			checkError(errors.New(fmt.Sprintf("Kiali does not have access to the %v namespace", namespace)))
+		}
 		log.Debugf("Build traffic map for namespace [%s]", namespace)
 		namespaceTrafficMap := buildNamespaceTrafficMap(namespace, o, client)
 
+                markInaccessible(namespaceTrafficMap, o)
```
jshaughn: I don't think you want to do this here.  You'd be better off doing this in the markOutsiders logic.  You want to wait for all of the appenders to run first.  Also, only outsider nodes can be inaccessible.

mwringe: The problem is that we can't wait for all the appenders to run first, some of the appenders will fail if it ties to access information in a namespace where it doesn't have access (eg the sidecar check appender)

graph/options/options.go
```diff
@@ -35,6 +35,7 @@ type VendorOptions struct {
 
 // Options are all supported graph generation options.
 type Options struct {
+        AccessibleNamespaces map[string]bool
```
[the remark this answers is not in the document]
mwringe: I wonder if we can make the formatter run before doing a commit, like what we have with kiali-ui

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#21](https://paste.fedoraproject.org/paste/PyeYmp36Kptwu0voncQUMw)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1224]](https://paste.fedoraproject.org/paste/vznz88Rv9BqkRMmRgaszQQ)
  * :red_circle: Kiali-on-Openshift [#[572]](https://paste.fedoraproject.org/paste/mlmvO7iwh7t6nLD0iBMN6w)


KIALI_NAME=kialicorepr448-istio-system

mwringe:
Updated PR.

I didn't change where the inaccessible check happens as it needs to occur before the appenders (some of the appenders will try and access information it doesn't have access to, such as the missing sidecar appender). But the other requested changes should be in there.

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#24](https://paste.fedoraproject.org/paste/-29ahLl5NMsS1Ke5-Rp7Jg)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1233]](https://paste.fedoraproject.org/paste/E6BpooS3Nb28sORYmL77mQ)
  * :red_circle: Kiali-on-Openshift [#[578]](https://paste.fedoraproject.org/paste/86c2syMeAffPsgcTgyqPPA)


KIALI_NAME=kialicorepr448-istio-system

lucasponce:
I see your last comment.
I will review the changes locally as well, sorry for my delay.

mwringe:
I updated the appender to check if it has access or not and moved marking the nodes as inaccessible to after the appenders have run (and its now part of the check if its outside or not, since a node can't be inaccessible unless its also outside the current namespace)

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#35](https://paste.fedoraproject.org/paste/FuptT7khP-rBt3dxP4PFPw)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1258]](https://paste.fedoraproject.org/paste/7~dCcU2SZgTe3PkyP~DtRg)
  * :heavy_check_mark: Istio-Upstream [#[146]](https://paste.fedoraproject.org/paste/fElWlcoXFhfzNhJ4E9DyhQ)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1030, 1031]](https://paste.fedoraproject.org/paste/dwE60lvoSZhnFkDFQLcTiw)
  * :heavy_check_mark: install-bookinfo [#[608]](https://paste.fedoraproject.org/paste/Qh62GIzPnp9~3UOytbGD9Q)
  * :heavy_check_mark: Kiali-on-Openshift [#[602]](https://paste.fedoraproject.org/paste/FxdxnxmA-yeRGKAaWS0cqQ)
  * :red_circle: run-kiali-e2e-tests [#[75]](https://paste.fedoraproject.org/paste/wkYBzlFgOgrgxplMOzjdGg)


KIALI_NAME=kialicorepr448-istio-system

mwringe:
Just a note that once this is reviewed there is still the kiali-ui aspect of this https://github.com/kiali/kiali-ui/pull/611

I don't think they need to be merged together (the UI part makes it work better, but I don't think there are any problems if you don't have have ui one merged yet)

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#36](https://paste.fedoraproject.org/paste/~N8Bw~Q6p4AibDqaKTGpBg)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1260]](https://paste.fedoraproject.org/paste/CnnhrhgtXnbeBInN6MAkDA)
  * :heavy_check_mark: Istio-Upstream [#[148]](https://paste.fedoraproject.org/paste/PmciQzpJPSYfUXekArTtkg)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1034, 1035]](https://paste.fedoraproject.org/paste/1Gtx-4EehWWLhgKRSFQXSw)
  * :heavy_check_mark: install-bookinfo [#[610]](https://paste.fedoraproject.org/paste/-r1kovFDMb6mGa3vSUJmcA)
  * :heavy_check_mark: Kiali-on-Openshift [#[604]](https://paste.fedoraproject.org/paste/vTDSr9s7PeJT3tydVyRsSQ)
  * :red_circle: run-kiali-e2e-tests [#[77]](https://paste.fedoraproject.org/paste/r0QMDj7aIY6mXKyXFsZXkw)


KIALI_NAME=kialicorepr448-istio-system

lucasponce:
Please @mwringe, could you rebase with master ?
I see refactoring done KIALI-1479 impacted in your work.
Let me know if there is something you would need help.
Thanks.

mwringe:
Rebased

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#44](https://paste.fedoraproject.org/paste/3gkVt2oVoJFVARx7rDpGxQ)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1282]](https://paste.fedoraproject.org/paste/GAI~8V5NtZPLnc0edBG5fg)
  * :heavy_check_mark: Istio-Upstream [#[167]](https://paste.fedoraproject.org/paste/TvKgJDlUm9FcJu3jxtLWyg)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1082, 1083]](https://paste.fedoraproject.org/paste/Fnbpy9kAqS3hwf3fnQvLcQ)
  * :heavy_check_mark: install-bookinfo [#[632]](https://paste.fedoraproject.org/paste/PxP1S4oIkFFm0bLgOhM39w)
  * :heavy_check_mark: Kiali-on-Openshift [#[632]](https://paste.fedoraproject.org/paste/FqUPzVwEHcJ3Z1v6N9~hXQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[99]](https://paste.fedoraproject.org/paste/oH0rTTgE3zet18RYrxzxGg)


KIALI_NAME=kialicorepr448-istio-system

handlers/graph.go
```diff
@@ -120,6 +123,14 @@ func isOutside(n *graph.Node, namespaces []string) bool {
 	return true
 }
 
+func isInaccessible(n *graph.Node, accessibleNamespaces map[string]bool) bool {
+	if _, found := accessibleNamespaces[n.Namespace]; !found {
+		return true
+	} else {
```
jshaughn: This is an unnecessary else clause.  Not a blocker but would probably be better to just:
```
_, found := accessibleNamespaces[n.Namespace]
return !found
```

PR #451, opened 2018-08-23
KIALI-1399 Keep Kiali versions in sync

mwringe:
Currently the Kiali UI version is set to 0.7.0 snapshot and the Kiali version is set to 0.6.1 snapshot. They versions are suppose to be kept in sync and they should both be 0.7.0 snapshot.


lucasponce:
Note that the UI doesn't use the snapshot flag, but that is minor
https://github.com/kiali/kiali-ui/blob/master/package.json#L3

PR #452, opened 2018-08-23
hack script now uses the new istiooc - installs OS 3.10, Istio 1.0

jmazzitelli:
This uses the latest istiooc - do not merge unless you can confirm this works. I tested on my box and it is working, but I would like others to try it, too.

Also adds `--mongo` and `--mysql` options to the bookinfo hack script in case you want to add those to your bookinfo install.


jmazzitelli:
@abonas see if these work for you.

PR #462, opened 2018-08-27
Recreate swagger.json in each travis build

xeviknal:
** Describe the change **

I have found the following scenario:
1. User1 adding invalid go-swagger comments into a field.
2. User1 not calling `make swagger-generate`
3. User1 getting merged his/her PR.
4. User2 adding go-swagger comments into a completely different feature.
5. User2 calling `make swagger-generate swagger-validate` and getting multiple errors unrelated to his/her feature.

I think we have to prevent user to push swagger-code that is not validated. Here, my proposal to fix it.
It has to be safe to re-create `swagger.json` in each build because any developer should modify that file. It has to be done always via go-swagger comments.


josejulio:
but we are only creating and not committing this file?

If we want to be more strict, maybe we can check in travis if the generated file is exactly the same as the latest copy (ensuring that the dev comited and pushed their `swagger.json`)

lucasponce:
I understand that this change lets travis to perform an additional validation and it will catch more missings issues in the doc; I also like @josejulio comments that perhaps we should be more strict.
All comments are complementary and it can be done in following interations.

PR #466, opened 2018-08-28
KIALI-1203 Check ServiceEntries for host validations

burmanm:


kubernetes/istio_details_service.go
```diff
@@ -439,8 +440,16 @@ func FilterByRoute(spec map[string]interface{}, protocols []string, service stri
...
 												if host, ok := mDestinationW["host"]; ok {
-													if sHost, ok := host.(string); ok && FilterByHost(sHost, service, namespace) {
-														return true
+													if sHost, ok := host.(string); ok {
+														if FilterByHost(sHost, service, namespace) {
+															return true
+														}
+														if serviceEntries != nil {
+															// We have ServiceEntry to check
+															if _, found := serviceEntries[strings.ToLower(protocol)+sHost]; found {
```
lucasponce: I guess I see a problem here.
This map is protocol+host, where protocol from the map is one of these [1]:
HTTP|HTTPS|GRPC|HTTP2|MONGO|TCP|TLS

But the protocol in the VirtualService will be one of [2]:
http, tls or tcp

Then something is not going to work when we test other usecases beyond http - I think -.

Perhaps in the ServiceEntryHostnames() method it should map the protocols used into the three ones expected by the VirtualServices ?

[1] https://istio.io/docs/reference/config/istio.networking.v1alpha3/#Port
[2] https://istio.io/docs/reference/config/istio.networking.v1alpha3/#VirtualService

burmanm: Our NoHostChecker does not actually understand "TLS" at all at this point, only "http/tcp":

https://github.com/kiali/kiali/blob/master/services/business/checkers/virtual_services/no_host_checker.go#L20

So should I change that too?

lucasponce: TLS is being addressed by @xeviknal in a different PR (I guess there is a PR with same work).

lucasponce:
cc @xeviknal Perhaps this is related too https://github.com/kiali/kiali/pull/467

services/business/checkers/virtual_services/single_host_checker.go
```diff
@@ -24,10 +24,8 @@ func (in SingleHostChecker) Check() models.IstioValidations {
 	validations := models.IstioValidations{}
 
 	for _, vs := range in.VirtualServices {
-		if hosts, ok := getHost(vs); ok {
-			for _, host := range hosts {
-				storeHost(hostCounter, vs, host)
-			}
+		for _, host := range getHost(vs) {
```
xeviknal: Good point!

services/business/istio_validations.go
```diff
@@ -87,21 +88,56 @@ func (in *IstioValidationsService) GetIstioObjectValidations(namespace string, o
...
+		if len(errChan) == 0 {
+			istioDetails.ServiceEntries = ses
+			istioDetails.VirtualServices = vss
+			istioDetails.DestinationRules = drs
+			virtualServiceChecker := checkers.VirtualServiceChecker{Namespace: namespace, VirtualServices: istioDetails.VirtualServices, DestinationRules: istioDetails.DestinationRules}
+			objectCheckers = []ObjectChecker{noServiceChecker, virtualServiceChecker}
+		}
+
 		if vss, err = in.k8s.GetVirtualServices(namespace, ""); err == nil {
 			if drs, err := in.k8s.GetDestinationRules(namespace, ""); err == nil {
```
xeviknal: If I understood your code correctly, this line and the previous one are redundant, right? Because you already have VirtualServices and DestinationRules (fetched in paralel in L#116, L#117). And in case we there are errors before, do you want to perform the same fetching calls again?

services/business/istio_validations.go
```diff
@@ -87,21 +88,56 @@ func (in *IstioValidationsService) GetIstioObjectValidations(namespace string, o
...
-				istioDetails.VirtualServices = vss
-				istioDetails.DestinationRules = drs
-				virtualServiceChecker := checkers.VirtualServiceChecker{Namespace: namespace, VirtualServices: istioDetails.VirtualServices, DestinationRules: istioDetails.DestinationRules}
-				objectCheckers = []ObjectChecker{noServiceChecker, virtualServiceChecker}
+				if ses, err := in.k8s.GetServiceEntries(namespace); err == nil {
```
xeviknal: I guess we have the same situation as before here.

services/business/istio_validations.go
burmanm: Heh, no. That's a rebase mistake, I'll remove the old code.

PR #376, opened 2018-07-23
KIALI-665 SSL Support

jotak:
- Get Openshift generates certificate
- Mount in  pod volume and configure to use them in kiali config
- Use them also for jaeger proxy
- Route: use reencrypting termination

PR #469, opened 2018-08-29
KIALI-665 Support SSl

aljesusg:
Continue @jotak work in https://github.com/kiali/kiali/pull/376

SSL is working
screenshot from 2018-08-29 12-33-58

Gif with SSL and Jaeger

ssl_kiali

It was  practically done by @jotak ^^


deploy/openshift/kiali.yaml
```diff
@@ -89,10 +93,15 @@ spec:
         volumeMounts:
         - name: kiali-configuration
           mountPath: "/kiali-configuration"
+        - name: kiali-cert
```
pilhuhn: Should we make this `optional=true`?

PR #470, opened 2018-08-29
KIALI-1373 Set the imagePullPolicy to always for vX.Y tags

mwringe:
** Describe the change **

If we are using an image tag of vX.Y then we should be using an imagePullPolicy of 'Always'.

The vX.Y releases represent the latest vX.Y.Z release, and if someone is using vX.Y we want them to always pull in the latest since those will have bug fixes.

The PR also allows someone to override the imagePullPolicy so they can set it to whatever they want.

Note: we don't explicitly set the imagePullPolicy if its doesn't match vX.Y, we let the system use the defaults in that case (which is IfNotPresent unless its 'latest' which will use 'Always'). This is why we use the IMAGE_PULL_POLICY_TOKEN in the yaml files, if its not specified then nothing extra is added to the yaml.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1373

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,6 +30,21 @@ DOCKER_NAME ?= kiali/kiali
 DOCKER_VERSION ?= dev
 DOCKER_TAG = ${DOCKER_NAME}:${DOCKER_VERSION}
 
+# If the IMAGE_PULL_POLICY is not defined and its the short version (eg vX.Y and not vX.Y.Z)
+# then we need to use the image pull policy of 'Always' otherwise the latest update to that
+# branch may not be brought in for users
+ifndef IMAGE_PULL_POLICY
+	SHORT_VERSION=$(shell if [[ ${DOCKER_VERSION} =~ ^v[0-9]+\.[0-9]+$$ ]]; then echo true; fi)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -30,6 +30,21 @@ DOCKER_NAME ?= kiali/kiali
...
 DOCKER_TAG = ${DOCKER_NAME}:${DOCKER_VERSION}
 
+# If the IMAGE_PULL_POLICY is not defined and its the short version (eg vX.Y and not vX.Y.Z)
+# then we need to use the image pull policy of 'Always' otherwise the latest update to that
+# branch may not be brought in for users
+ifndef IMAGE_PULL_POLICY
+        SHORT_VERSION=$(shell if [[ ${DOCKER_VERSION} =~ ^v[0-9]+\.[0-9]+$$ ]]; then echo true; fi)
+        ifeq (${SHORT_VERSION}, true)
+		IMAGE_PULL_POLICY = Always
+	endif
```
jmazzitelli: OCD alarm going off - indentation is off - ifeq..endif aren't on same column.

mwringe: Updated the spacing.

PR #471, opened 2018-08-29
trivial - check error arg in intermittent failing test

jshaughn:
to see if we can learn something...


lucasponce:
There was a planned JIRA for this https://issues.jboss.org/browse/KIALI-1429
I guess we can consider done unless we have more information.

PR #472, opened 2018-08-30
KIALI-1135 Check that VirtualService gateway exists

burmanm:
If gateway is defined in VirtualService, validates that such Gateway also exists (by name)

### DIFF
--- a/services/business/checkers/virtual_services/no_gateway_checker.go
+++ b/services/business/checkers/virtual_services/no_gateway_checker.go
@@ -1,0 +1,23 @@
+package virtual_services
+
+import (
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/services/models"
+)
+
+type NoGatewayChecker struct {
+	VirtualService kubernetes.IstioObject
+	GatewayNames   map[string]struct{}
+}
+
+// Check validates that all the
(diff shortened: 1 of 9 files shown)


services/business/checkers/virtual_services/no_gateway_checker.go
```diff
@@ -0,0 +1,23 @@
...
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/services/models"
+)
+
+type NoGatewayChecker struct {
+	VirtualService kubernetes.IstioObject
+	GatewayNames   map[string]struct{}
+}
+
+// Check validates that all the
```
lucasponce: Minor, missing comment ?

services/business/istio_validations.go
```diff
@@ -106,15 +107,26 @@ func (in *IstioValidationsService) GetIstioObjectValidations(namespace string, o
 		}
 	}
 
+	// Identical to above, but since k8s layer has both (namespace, serviceentry) and (namespace) queries, we need two different functions
+	fetchNoEntry := func(rValue *[]kubernetes.IstioObject, namespace string, fetcher func(string) ([]kubernetes.IstioObject, error), errChan chan error) {
```
lucasponce: Yeah.
I think that we have some duplication in the kubernetes packages and some methods can be joined to unified parameters, but yes, this is correct.

PR #25, opened 2018-02-15
Allow deployments to be successful using minishift

xeviknal:
I have OpenShift installed through minishift. It means that when you are using `make openshift-deploy` the sws image built is not available within the cluster. Then, OpenShift goes to docker.io to look for it but it only finds the @jmazzitelli latest version.

This is only a humble proposal. I don't know if it has any sense to do it this way, so please let me know :)

1. I create a repo into docker.io image registry
2. I changed `Makefile` to allow the name of the docker image as a parameter.

This way, openshift goes to my personal docker.io account and pulls it from there.

```bash
export DOCKER_NAME=xeviknal/sws
export DOCKER_VERSION=latest
make docker
docker push xeviknal/sws:latest
make openshift-deploy
```
Here is the output:

minishift-deployment-own-dockerio

Should perhaps add another target to Makefile for this minishift deploy? 
```bash
minishift-deploy:
  @echo Pushing current docker image to ${DOCKER_TAG}
  docker push ${DOCKER_TAG}
  openshift-deploy
```

Just an idea :dancing_men:

PR #473, opened 2018-08-30
KIALI-1136 Add new GroupChecker for DestinationRules

burmanm:
Checks that only a single DestinationRule per host is created

### DIFF
--- a/services/business/checkers/destination_rules/multi_match_checker.go
+++ b/services/business/checkers/destination_rules/multi_match_checker.go
@@ -1,6 +1,8 @@
 package destination_rules
 
 import (
+	"strings"
+
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/services/models"
 )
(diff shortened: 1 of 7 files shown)


lucasponce:
Is this considered a validation to show at VirtualService level or DestinationRule level ?

services/business/checkers/destination_rules/multi_match_checker.go
```diff
@@ -0,0 +1,44 @@
...
+						ObjectType: "destinationrules",
+						Valid:      true,
+						Checks: []*models.IstioCheck{
+							&checks,
+						},
+					}
+
+					validations.MergeValidations(models.IstioValidations{key: rrValidation})
+				} else {
+					seenHosts[dHost] = empty
```
xeviknal: I think that using a one-field-key map like this one, you are going to miss some duplicates due to the different ways istio allow users to specify hosts. 
e.g. `reviews` host might appear in a destination rule as `reviews.bookinfo.svc.cluster.local`. 
also `reviews` host can fell into a wildcard host like this one `*.bookinfo.svc.cluster.local' or simply to a `*` host.

Therefore, if you have on DR with `reviews` and another one with `reviews.bookinfo.svc.cluster.local`, this validation won't show up.

We did some work connected to determine duplicated by host for Virtual Services: https://github.com/kiali/kiali/blob/master/services/business/checkers/virtual_services/single_host_checker.go#L122
At the end, we have to compare host not just by name but conceptually (name, namespace, cluster).

services/business/checkers/virtual_service_checker.go
```diff
@@ -50,6 +51,14 @@ func (in VirtualServiceChecker) runGroupChecks() models.IstioValidations {
 		validations = validations.MergeValidations(checker.Check())
 	}
 
+	enabledDRCheckers := []GroupChecker{
```
xeviknal: IMO, I would say that this validation takes `DestinationRules` as subject so I would place it into the DestinationRule checker.

services/business/checkers/destination_rules/multi_match_checker.go
burmanm: I tried to search for the rules for those hostnames, but I could not find anything in the documention (now browsing the istio source code). Are all regexps allowed for the destionationrules, or is this the definition:

https://github.com/istio/istio/blob/master/pilot/pkg/model/validation.go#L43

Thus only allowing the leading ``*`` or ``*-`` ?

services/business/checkers/virtual_service_checker.go
burmanm: Yep, moving.

services/business/checkers/destination_rules/multi_match_checker.go
burmanm: And the secondary question, are these harmful definitions or are they alright:

* *.bookinfo.svc.cluster.local
* bookinfo.svc.cluster.local

It shouldn't match the same hosts, but how does it do sub-hosts internally?

xeviknal: I use to follow the descriptions in the reference page: https://istio.io/docs/reference/config/istio.networking.v1alpha3/#DestinationRule

> REQUIRED. The name of a service from the service registry. Service names are looked up from the platform’s service registry (e.g., Kubernetes services, Consul services, etc.) and from the hosts declared by ServiceEntries. Rules defined for services that do not exist in the service registry will be ignored.

> Note for Kubernetes users: When short names are used (e.g. “reviews” instead of “reviews.default.svc.cluster.local”), Istio will interpret the short name based on the namespace of the rule, not the service. A rule in the “default” namespace containing a host “reviews will be interpreted as “reviews.default.svc.cluster.local”, irrespective of the actual namespace associated with the reviews service. To avoid potential misconfigurations, it is recommended to always use fully qualified domain names over short names.

From that description, I understand that  you can have both 'short name' (reviews) and 'FQDN' (reviews.bookinfo.svc.cluster.local). What I can't tell is whether wildcard expressions are allowed or no. It would be great to prepare an example mesh with that an play with it.

lucasponce: I don't think that wildcards will work well for DestinationRule.
So I would consider a first scenario where inner service can be expressed in short name or FQDN.
If wildcards are supported at this level, that can be clarified and fixed later.

burmanm: This seems to indicate that wildcard could be usable? https://github.com/istio/istio/blob/master/tests/testdata/config/destination-rule-all.yaml#L24

burmanm: Yeah, it is allowed, from the Istio blog: "The host of a DestinationRule can include wildcard prefixes, allowing a single rule to be specified for many actual services."

services/business/checkers/destination_rules/multi_match_checker.go
```diff
@@ -0,0 +1,103 @@
...
+	Service   string
+	Namespace string
+	Cluster   string
+}
+
+func (m MultiMatchChecker) Check() models.IstioValidations {
+	validations := models.IstioValidations{}
+
+	var empty struct{}
+	seenHosts := make(map[string]map[string]map[string]struct{}) // Poor man's trie
```
xeviknal: For your comment at the end of the line, I guess there is something you don't like that much in this type. Probably lack of readiness.
What do you think about moving this type to sub-types? I could help other's devs to better understand what we are doing here. Besides that, I have the feeling that this code will be reused by other components.
I was thinking of something like the following. Perhaps, different naming and without the string types.

```golang
type Cluster string
type Namespace string
type Service string
type ClusterSearchMap map[Cluster]NamespaceSearchMap
type NamespaceSearchMap map[Namespace]ServiceSearchMap
type ServiceSearchMap map[Service]struct{}
```

burmanm: Readability is one, yes. The other thing is that it's pretty limited structure. But as long as we can guarantee that there's no need for other query types than these, then it's going to be fine. But what if we need something like ``*-frontend.namespace.cluster`` and the ``*-frontend`` could match ``service1-frontend`` as well as ``service1-qa-frontend``. That's when we would need a better sorted data structure (or other prefix-searchable)

lucasponce:
I fetched locally and I found some issue when checking from UI.
So, in the IstioConfig list page we have an error:
image
But in the details the error is gone:
image

I haven't investigated, but it is a hint that there is a mistmatch when asking validations per a namespace vs validations for a single object.

Perhaps, something is missing.

Let me know if you can take it a look.

burmanm:
@lucasponce It's not this PR that causes that error, the error is that bookinfo points to a gateway that is not found (but .. it is later, odd). I'll investigate the reason in any case, but probably in another PR.

burmanm:
Well, this was tricky to understand at first.. but it was just because the original base branch was too old. Rebase from the master fixed the issue.

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#25](https://paste.fedoraproject.org/paste/WMd8sU~IYR5QgWxKt~P2Fw)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1237]](https://paste.fedoraproject.org/paste/BdDRrUQZUoU6y3EuFR3P1w)
  * :red_circle: Kiali-on-Openshift [#[582]](https://paste.fedoraproject.org/paste/5wEiYtD7uxMq1Dj59W2-VQ)


KIALI_NAME=kialicorepr473-istio-system

PR #478, opened 2018-09-04
Readme: change intro question

jotak:
I think it's more correct to reformulate the question, because Kiali doesn't know anything about what a service is doing, beyond its interactions. We have no idea of the services business value.

See also https://github.com/kiali/kiali.io/pull/20

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -8,7 +8,7 @@ image:https://img.shields.io/badge/license-Apache2-blue.svg["Apache 2.0 license"
 
 == Introduction
 
-Kiali provides answers to the question _What are the microservices in my Istio service mesh doing?_
+Kiali provides answers to the questions: _What microservices are part of my Istio service mesh and how are they connected?_
 
 image::https://www.kiali.io/images/features/feature3.png[Kiali Graph]
 


README.adoc
```diff
@@ -6,7 +6,7 @@ toc::[]
 
 == Introduction
 
-Kiali provides answers to the question _What are the microservices in my Istio service mesh doing?_
+Kiali provides answers to the questions: _What are the microservices in my Istio service mesh, and how are they connected?_
```
abonas: Maybe "What microservices exist in my Istio service mesh" instead of "what are ....", otherwise I feel it misses a verb :)

JStickler: I would choose either "What microservices are part of my service mesh and how are they connected."  or "Which microservices exist in my service mesh and how are they connected."

pilhuhn:
I'd merge that now, as the counterpart on kiali.io has been merged.

README.adoc
abonas: @jotak have you addressed @JStickler comments?

jotak: done

PR #479, opened 2018-09-04
KIALI-1235 Security appender sensitive to policy switch

xeviknal:
** Describe the change **

Making security appender sensitive to policy switch (from non-secure to secure and viceversa).
Prom queries only took secure data points (connection_security_policy != none) which isn't an accurate query. It turns out that secure data points are also present in queries even if mtls is switched off in istio.
This PR takes all the data points and compute the `isMTLS` flag.

In this solution, there is only one unknown scenario which is when you have secure and non-secure data points with a 0 value. Meaning that there is no traffic for that edge. In that case, `isMTLS` is set to `false` (taking a conservative approach).

** Issue reference **
When switching from mtls to non-tls security policy, graph doesn't change.

https://issues.jboss.org/browse/KIALI-1235

** Backwards incompatible? **
Yes.


xeviknal:
@jshaughn you are right! thank you very much for that piece of advice.

PR #480, opened 2018-09-05
KIALI-1480 Provide workload name and namespace for service source deps

jotak:
https://issues.jboss.org/browse/KIALI-1480

Related frontend PR: https://github.com/kiali/kiali-ui/pull/647

### DIFF
--- a/services/models/service.go
+++ b/services/models/service.go
@@ -28,14 +28,14 @@ type ServiceList struct {
 }
 
 type ServiceDetails struct {
-	Service          Service             `json:"service"`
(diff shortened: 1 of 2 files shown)


lucasponce:
As I commented, there is a missing discussion around this.
It is right that service is invoked from workloads and not services.
But it is not clear if the solution was to move this to workloads.

This approach is right, but I was hoping to discuss if we want to just show workloads or to show the services related to the workloads, to let Kiali add more value.

In any case, I'm testing it locally.

PR #482, opened 2018-09-06
Remove some dead code and fix a log in token generation

jotak:
About dead code: what I removed is some error handling part, but:

- it never ran into that code (see line 65)
- the initial error message is actually already accurate so it's not necessary to replace it

### DIFF
--- a/config/authentication.go
+++ b/config/authentication.go
@@ -14,7 +14,7 @@ func AuthenticationHandler(next http.Handler) http.Handler {
 		if strings.Contains(r.Header.Get("Authorization"), "Bearer") {
 			err := ValidateToken(strings.TrimPrefix(r.Header.Get("Authorization"), "Bearer "))
 			if err != nil {
-				log.Warning("Error token %+v", err)
+				log.Warning("Token error: ", err)
 				statusCode = http.StatusUnauthorized
 			}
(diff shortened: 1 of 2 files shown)


config/authentication.go
```diff
@@ -14,7 +14,7 @@ func AuthenticationHandler(next http.Handler) http.Handler {
 		if strings.Contains(r.Header.Get("Authorization"), "Bearer") {
 			err := ValidateToken(strings.TrimPrefix(r.Header.Get("Authorization"), "Bearer "))
 			if err != nil {
-				log.Warning("Error token %+v", err)
+				log.Warning("Token error: ", err)
```
jshaughn: It took me a moment to accept that this was OK, I originally thought the fix should be:

`  log.Warningf("Error token %+v", err)`

But your way should work as well.

lucasponce:
Please @jshaughn, let's give to the explicit reviewer requested possibility to comment before merge.
Perhaps some PR sounds trivial, but if @jotak requested to @aljesusg to take a look as original author, I would expect some comment from him in a reasonable timeframe.

jshaughn:
@lucasponce OK, I just thought for such a small PR that I would do Alberto a favor, but I will avoid in the future.

PR #485, opened 2018-09-07
KIALI-1464 labels need to come from pods not the deployments

aljesusg:
** Issue reference **

[KIALI-1464](https://issues.jboss.org/browse/KIALI-1464)

### DIFF
--- a/services/business/workloads_test.go
+++ b/services/business/workloads_test.go
@@ -37,11 +37,21 @@ func TestDeploymentListHandler(t *testing.T) {
 
 	assert.Equal(3, len(workloads))
 	assert.Equal("httpbin-v1", workloads[0].Name)
+	assert.Equal(true, workloads[0].AppLabel)
+	assert.Equal(false, workloads[0].VersionLabel)
 	assert.Equal("httpbin-v2", workloads[1].Name)
+	assert.Equal(true, workloads[1].AppLabel)
+	assert.Equal(true, workloads[1].VersionLabel)
 	assert.Equal("httpbin-v3", workloads[2].Name)
+	assert.Equal(false, workloads[2].AppLabel)
+	assert.Equal(false, workloads[2].VersionLabel)
 }
 
 func fakeDeploymentList() *v1beta1.DeploymentList {
+	conf := config.NewConfig()
+	config.Set(conf)
+	appLabel := conf.IstioLabels.AppLabelName
+	versionLabel := conf.IstioLabels.VersionLabelName
 	t1, _ := time.Parse(time.RFC822Z, "08 Mar 18 17:44 +0300")
 	return &v1beta1.DeploymentList{
 		Items: []v1beta1.Deployment{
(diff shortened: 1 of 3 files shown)


services/models/workload.go
```diff
@@ -145,9 +145,9 @@ func (workload *WorkloadListItem) Parse(d v1beta1.Deployment) {
 	workload.Name = d.Name
 	workload.Type = DEPLOYMENT
 
-	/** Check the labels app and version required by Istio*/
-	_, workload.AppLabel = d.Labels["app"]
-	_, workload.VersionLabel = d.Labels["version"]
+	/** Check the labels app and version required by Istio in tempalte Pods*/
```
lucasponce: typo "Template"

services/models/workload.go
```diff
@@ -145,9 +145,9 @@ func (workload *WorkloadListItem) Parse(d v1beta1.Deployment) {
...
+	_, workload.AppLabel = d.Spec.Template.Labels["app"]
```
lucasponce: As Spec.Template could not exist, it needs some protection.

aljesusg: That is why has

```go
 _, workload.AppLabel = d.Spec.Template.Labels["app"]
```

aljesusg: On the other hand Istio create the object Tempalte with empty values.

Template: v1.PodTemplateSpec{
						ObjectMeta: meta_v1.ObjectMeta{
							Labels: map[string]string{},
						},
					},

So the value will be "" for the first parameter **_** and for second **workload.AppLabel** if the key exist or not

jotak:
Side question - I know this is not the topic of this jira, but... we're hardcoding "app" and "version" strings whereas they can be configured in kiali configmap. Is it expected? Like, do we expect the validation warning to pop up because the user doesn't use exactly "app" and "version", or should we warn because he doesn't use the labels that he defined in kiali config?
I would opt for the latter.

lucasponce:
> we're hardcoding "app" and "version" strings 

Good point. Yes in most places this is fetched from the config, I guess it can go in the PR as well.

aljesusg:
No, it was my fault there are values in the configuration I think that this was done before I am gonna change it. Thanks @jotak

jotak:
@aljesusg ok, if you change it in this PR then I've seen another place where the same is done: file `business/services.go` in `buildServiceList`

aljesusg:
Changed

aljesusg:
Wait I am moving all app to config

jotak:
@aljesusg I didn't see other places with the same problem ... except maybe in graph but I'm not sure if it's intentional or not ; and in tests but to me it's ok to hardcode in tests.

services/business/workloads_test.go
```diff
@@ -37,23 +37,37 @@ func TestDeploymentListHandler(t *testing.T) {
...
 	assert.Equal("httpbin-v2", workloads[1].Name)
+	assert.Equal(true, workloads[1].AppLabel)
+	assert.Equal(true, workloads[1].VersionLabel)
 	assert.Equal("httpbin-v3", workloads[2].Name)
+	assert.Equal(false, workloads[2].AppLabel)
+	assert.Equal(false, workloads[2].VersionLabel)
 }
 
 func fakeDeploymentList() *v1beta1.DeploymentList {
+	conf := config.Get()
```
lucasponce: did you setup the config on the test ? otherwise this may fail.

aljesusg: I see....maybe it's a good moment to move all mocks to the same file ^^

aljesusg: Anyway...this config comes from "github.com/kiali/kiali/config"  and it's working

lucasponce: There is a separated JIRA for that to be planned in Sprint 11.
But yes, it's some of the areas identified to improve.

PR #520, opened 2018-09-26
KIALI-1621 - Replace CB Rule Yaml

mattmahoneyrh:
Replace CB rule Yaml file:
Removed: assets/circuitbreaker.yaml
Added: assets/bookinfo-reviews-all-cb.yaml
Updated: tests/conftest.py


rhqci:

PR #524, opened 2018-09-28
KIALI-1635 cluster-openshift does not work with sh as it uses bashism

burmanm:
Since cluster-openshift.sh is not a standard sh script, we should make it use bash instead so it works on systems with standard compliant /bin/sh (such as Ubuntu which uses dash instead of bash in /bin/sh)


rhqci:

PR #529, opened 2018-10-01
Make Istio download scripts use posix sh or bash directly

burmanm:
Istio download scripts do not work under Ubuntu either without some fixes to make them standard compliant - or use bash instead of sh


lucasponce:
Please @jmazzitelli, could you give a try to this as you are the reviewer and main author of the script ?

PR #537, opened 2018-10-03
[WIP] KIALI-1669 Move k8sQPS, k8sBurst to config package

### LINKED ISSUE #532: Move QPS and Burst parameters to config
There are a couple of parameters fixed as constants that can be moved and read them from the config package:

https://github.com/kiali/kiali/blob/c1c9551daf6cf8eea3053ac42056ac5fd3b1c14a/kubernetes/client.go#L26

bhavin192:
** Describe the change **
- This move the k8sQPS, k8sBurst to config package as const
- Also changes those to K8sBurst and K8sQPS
- Update the kubernetes/client.go accordingly

** Issue reference **
- Fixes #532 
- Fixes https://github.com/kiali/hacktoberfest18/issues/8

Edit: seems like I have missed some of the references, will update them in few hours


rhqci:

bhavin192:
Should I make this change in this as well?

https://github.com/kiali/kiali/blob/master/status/versions.go#L165-L172

jotak:
Hi @bhavin192 and thanks for your contribution.

I think the ticket description lacks of clarity, I'm going to update it asap. The intent is actually not to move constants from a package to another, but instead to remove them as constants and have them as configurable options instead, with default values.

Basically, all of Kiali's configurable options can be read either from the config file (ConfigMap) or from environment variable. They're also documented in the readme.

I hope you're still interested in doing it? :-)

In my opinion, what could be a correct implementation would be to have a new `KubernetesConfig` struct  similar what we already have for `IstioConfig`, `JaegerConfig` and so on: https://github.com/kiali/kiali/blob/master/config/config.go#L102 . It would just contain Burst and QPS.

As you can see here [ https://github.com/kiali/kiali/blob/master/config/config.go#L153-L173 ] , all configurable options can be read from environment variables and have default values, so QPS and Burst default would be their current constant values.

bhavin192:
@jotak yeah! that's what I thought the intention was but later reading the ticket I thought about just moving the config variables from one package to other, I will update my PR with new changes :)

jotak:
Great, thank you!
And about your other question:

> Should I make this change in this as well?
> https://github.com/kiali/kiali/blob/master/status/versions.go#L165-L172

You're right, it would make sense to change them as well, using the same config values.

bhavin192:
Sorry for having no updates on this for last few days, I have worked on this. Will push the code once I'm done with remaining file.

bhavin192:
@jotak hey, I have updated the PR with correct changes, can you please review?

PR #157, opened 2018-04-16
remove that erroneous third column in IN_CLUSTER row

jmazzitelli:
Also made the default verbage consistent across all rows -- now indicates default values with "(default is `<value>`)"

PR #549, opened 2018-10-09
[WIP] KIALI-1661 Provide endpoints to check Kiali authorization

jotak:
Uses k8s "SelfSubjectAccessReview"
+ additional readonly flag from config

JIRA: https://issues.jboss.org/browse/KIALI-1661

TODO:

- [ ] Decide on readonly flag
- [ ] Swagger doc
- [ ] Unit tests

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -356,6 +356,13 @@ server:
 in_cluster: (true\|false)
 ----
 
+|`READ_ONLY`
+|Tells if Kiali must run in read-only mode or not. This mode prevents any modification to a configuration, to Istio objects or other changes of any other kind. (default is `true`)
+[source,yaml]
+----
+read_only: (true\|false)
+----
+
 |`PROMETHEUS_SERVICE_URL`
 |The URL used to access and query the Prometheus Server. It must be accessible from Kiali pod. (default is `http://prometheus.istio-system:9090`)
 [source,yaml]
--- a/config/config.go
+++ b/config/config.go
@@ -24,6 +24,7 @@ const (
 	EnvIstioIdentityDomain    = "ISTIO_IDENTITY_DOMAIN"
 	EnvIstioSidecarAnnotation = "ISTIO_SIDECAR_ANNOTATION"
 	EnvIstioUrlServiceVersion = "ISTIO_URL_SERVICE_VERSION"
+	EnvReadOnly               = "READ_ONLY"
 
 	EnvServerAddress                    = "SERVER_ADDRESS"
 	EnvServerPort                       = "SERVER_PORT"
@@ -126,6 +127,7 @@ type Config struct {
 	LoginToken       LoginToken        `yaml:"login_token,omitempty"`
 	IstioNamespace   string            `yaml:"istio_namespace,omitempty"`
 	IstioLabels      IstioLabels       `yaml:"istio_labels,omitempty"`
+	ReadOnly         bool              `yaml:"read_only,omitempty"`
 }
 
 // NewConfig creates a default Config struct
@@ -138,6 +140,7 @@ func NewConfig() (c *Config) {
 	c.IstioNamespace = strings.TrimSpace(getDefaultString(EnvIstioNamespace, "istio-system"))
 	c.IstioLabels.AppLabelName = strings.TrimSpace(getDefaultString(EnvIstioLabelNameApp, "app"))
 	c.IstioLabels.VersionLabelName = strings.TrimSpace(getDefaultString(EnvIstioLabelNameVersion, "version"))
+	c.ReadOnly = getDefaultBool(EnvReadOnly, false)
 
 	// Server configuration
 	c.Server.Address = strings.TrimSpace(getDefaultString(EnvServerAddress, ""))
--- a/handlers/authorize.go
+++ b/handlers/authorize.go
@@ -1,0 +1,41 @@
+package handlers
+
+import (
+	"net/http"
+
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+)
+
+// CheckAuthorization checks Kiali authorization against the Kubernetes API (SelfSubjectAccessReview) for a given resource
+func CheckAuthorization(w http.ResponseWriter, r *http.Request) {
+	params := r.URL.Query()
+	namespace := params.Get("namespace")
+	verb := params.Get("verb")
+	group := params.Get("group")
+	resource := params.Get("resource")
+
+	if config.Get().ReadOnly && verb != "get" && verb != "list" {
+		RespondWithError(w, http.StatusUnauthorized, "Kiali configured in read-only mode")
+		return
+	}
+
+	client, err := kubernetes.NewClient()
+	if err != nil {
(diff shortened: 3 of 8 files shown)


jotak:
I'm suggesting two different approaches to restrict Kiali capabilities in the frontend:

- Using the Kubernetes API to check what Kiali can actually do
- With a "readonly" config flag that would override (if true) the kube permissions

I'm not 100% sure about the latter, so discussion is welcomed. I think it can be helpful to restrict Kiali globally without having to fine-tune kube's permissions. On the other hand it adds more configure options so can eventually add confusion. What do you think?

@mwringe , @jmazzitelli you may want to have a look too

pilhuhn:
I think the readonly-flag (also) needs to be enforced in the backend. Otherwise it is possible for a user to open the JS console and send requests that way even if the user would have the restrict flag set in the frontend.

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#157](https://paste.fedoraproject.org/paste/9bhzL2dTHFfC2nt8KUbhpA)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1550]](https://paste.fedoraproject.org/paste/eDegP19wWAZa~DtiAg6Jrw)
  * :heavy_check_mark: Istio-Upstream [#[407]](https://paste.fedoraproject.org/paste/pW7pxy4Rd5Pd-5P6UTyoNQ)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1654, 1655]](https://paste.fedoraproject.org/paste/zl8dmgvhMwO8bnJwmMvAkQ)
  * :heavy_check_mark: install-bookinfo [#[920]](https://paste.fedoraproject.org/paste/pDraHWMvUJk3II2GOKAVug)
  * :heavy_check_mark: Kiali-on-Openshift [#[927]](https://paste.fedoraproject.org/paste/lKZfmISzLtZz5wbTIkp7lw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[348]](https://paste.fedoraproject.org/paste/~51k84fHXaX~b87WS15wiw)


KIALI_NAME=kialicorepr549-istio-system

mwringe:
Why do we want a readonly option?

The RBAC of OpenShift/K8S should be the single source for determining if a user has these permissions or not. If you don't want to have Kiali have write permissions, they you shouldn't grant the Kiali service account write permissions.

In OpenShift we are also updating things to allow individual users to use their own tokens to access things. In this case a system wide readonly flag doesn't make a lot of sense (some users may have read, some may also have write).

We also shouldn't have a new endpoint to check things here. It should be returned as part of the request to get the information.

Eg when I go to the istio detail page I get a list of the detail objects (eg api/namespaces/istio-bookinfo/istio) this should include in there if I have read/write permissions or not. From there there UI can figure out if the user should be presented with an option to edit or just read the yaml.

jotak:
> Why do we want a readonly option?
> 
> The RBAC of OpenShift/K8S should be the single source for determining if a user has these permissions or not. If you don't want to have Kiali have write permissions, they you shouldn't grant the Kiali service account write permissions.

I don't necessarily disagree (and that's why I opened the discussion), but a global readonly flag would just be an easy and helpful shortcut for people who don't want to spend too much time on finding out all kiali permissions. Think about it: they would have to analyze the whole kiali deployment yaml to search for "delete", "create", "update" verbs across potentially several different groups and apis ... it's quite easy to miss one. That's the argument to put in balance, now I'm not saying we have to do it.

> 
> In OpenShift we are also updating things to allow individual users to use their own tokens to access things. In this case a system wide readonly flag doesn't make a lot of sense (some users may have read, some may also have write).

I would say on the contrary, it's a good scenario where it makes sense. Ability to deactivate writes system-wide, for whatever reason, instead of reviewing all permissions on a user basis one by one.

> 
> We also shouldn't have a new endpoint to check things here. It should be returned as part of the request to get the information.
> 
> Eg when I go to the istio detail page I get a list of the detail objects (eg api/namespaces/istio-bookinfo/istio) this should include in there if I have read/write permissions or not. From there there UI can figure out if the user should be presented with an option to edit or just read the yaml.

Agree, it makes sense.

mwringe:
> I don't necessarily disagree (and that's why I opened the discussion), but a global readonly flag would just be an easy and helpful shortcut for people who don't want to spend too much time on finding out all kiali permissions. Think about it: they would have to analyze the whole kiali deployment config to search for "delete", "create", "update" verbs across potentially several different groups and apis ... it's quite easy to miss one. That's the argument to put in balance, now I'm not saying we have to do it.

We should have roles to help define this (eg 'kiali-read' and 'kiali-write', or something like that). The admin would then be able to configure which roles they want to grant to users.

Users shouldn't have to know more than that. I wouldn't expect them to look through the yaml and need to understand that or figure out how that should work and make modifications themselves.

> I would say on the contrary, it's a good scenario where it makes sense. Ability to deactivate writes system-wide, for whatever reason, instead of reviewing all permissions on a user basis one by one.

I don't know if I really understand this use case properly. Can you please explain this a bit more as to why it is needed?

I see this as we now have two things that we need to keep in sync: the value in the configmap to enable/disable readonly mode, and the roles that we grant to users (either a read role or a write role). This is going to potentially introduce more frustration to users as thing could get out of sync (eg they grant a write role to the user but the readonly parameter is still set).

I would think just setting the role would make things a lot easier. I would expect users would be more concerned over the permissions it needs to grant the Kiali service account. Granting the service account the ability to do a write and disabling it with the readonly configuration wouldn't necessarily alleviate their concerns. It still has permissions to do the write, so something like a bug or if the pod got compromised would still present problems.

If its a performance issue, then maybe we want to make sure we can do things in a way that we only ever fetch if a user has read/write permission when its necessary. I suspect most people will be running with the write ability enabled, so we need to make sure it works well in that situation.

jotak:
> > I would say on the contrary, it's a good scenario where it makes sense. Ability to deactivate writes system-wide, for whatever reason, instead of reviewing all permissions on a user basis one by one.
> 
> I don't know if I really understand this use case properly. Can you please explain this a bit more as to why it is needed?

I cannot foresee all potential usages. One example:
You notice that some Istio objects are modified unexpectedly through Kiali. You want to investigate. First step: turn off the modification switch (readonly flag). Second step: investigate who has more privilege than it should, and fix permissions. Third step: remove readonly flag.

The thing is we haven't focused a lot on security from the beginning for an obvious reason, all that we are doing til a near future is read-only actions. Introducing modification capabilities will undoubtedly bring more security concerns. A readonly flag is also a security switch.

> 
> I see this as we now have two things that we need to keep in sync: the value in the configmap to enable/disable readonly mode, and the roles that we grant to users (either a read role or a write role). This is going to potentially introduce more frustration to users as thing could get out of sync (eg they grant a write role to the user but the readonly parameter is still set).
> 

There's nothing to keep in sync, why would anyone set the readonly flag if he's not willing to put the whole in readonly? It's quite self-explanatory, I don't see the ambiguity. it's much like a general water valve that takes priority over all the taps in your house.

Again, I don't have myself a clear preference for a solution over the other. So as the interest of a read-only flag is not obvious (looks like it's not) - then no problem I'll remove it, question solved.

jotak:
will reopen another one

PR #556, opened 2018-10-11
KIALI-1730 Fix consistent workload name between list and detail

lucasponce:


rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#174](https://paste.fedoraproject.org/paste/1GP~23VNc9XPm3BRaG-l2g)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1591]](https://paste.fedoraproject.org/paste/x~qwTgL~3Hmam7Gc4cl7CQ)
  * :heavy_check_mark: Istio-Upstream [#[444]](https://paste.fedoraproject.org/paste/N3ooI0FLR8DAC9L6RcjM5A)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1746, 1748]](https://paste.fedoraproject.org/paste/2Zf1Pru-m~TlNq~akCaxwg)
  * :heavy_check_mark: install-bookinfo [#[965]](https://paste.fedoraproject.org/paste/WZz7mdkrJ0Q-NKJLDfPUZg)
  * :heavy_check_mark: Kiali-on-Openshift [#[980]](https://paste.fedoraproject.org/paste/bqeS093D6OpvoA9I~wMgRQ)
  * :red_circle: run-kiali-e2e-tests [#[389]](https://paste.fedoraproject.org/paste/T9GpiJpQ8Z~v8KN2jjHz6w)


KIALI_NAME=kialicorepr556-istio-system

jotak:
Question: If I understand correctly the problem is that two different objects (like a statefulset and a deployment) can share the same name, right? And this is introducing prioritization to detect which one prevails on the other.
But does it means that, when several objects of different kinds share the same name, only one of them would be displayed? Is it the intended behaviour?

lucasponce:
> But does it means that, when several objects of different kinds share the same name, only one of them would be displayed? Is it the intended behaviour?

Yes, this is the intended behaviour. For Istio on this specific case, it only takes one workload and ignores the other. So I sync-ed with the Istio code to at least reflect the same which is rendered in the graph.

jotak:
@lucasponce ok, it makes sense, it might be a (edge-case) problem but it would be essentially an istio issue, so we make best effort here. +1

PR #230, opened 2018-05-30
[KIALI-599] Init Style Code Guide

aljesusg:
Initial Style Code Guide, we'll update this

PR #220, opened 2018-05-28
KIALI-521 metrics, make start and end times multiple of step

jotak:

Issue #219, opened 2018-05-28
Empty Service Graph


debianmaster:
<img width="1197" alt="screen shot 2018-05-28 at 7 01 20 pm" src="https://user-images.githubusercontent.com/1439169/40611699-0a611b50-62aa-11e8-841c-56bf62fa129d.png">

unable to see service graph data for any namespace except istio-system.   not much help in kiali pod logs.  i can see data in istio service graph and as well as grafana.

Any clues where to check?

lucasponce:
Hello @debianmaster, can you see the services list ? Clicking on Services menu tab ?
Could you access to the kiali pod log and post it here ?
That could help us to check what is going on there.

lucasponce:
I assume you are testing latest master, or if not, could you post the commit / version you are using ?

debianmaster:
i can see the list of services.  

<img width="1017" alt="screen shot 2018-05-28 at 7 31 13 pm" src="https://user-images.githubusercontent.com/1439169/40612622-eebd33e4-62ad-11e8-825c-f8f5479815ad.png">


kiali logs 

```
I0528 11:31:02.545437       1 kiali.go:44] Kiali: Version: 0.3.1.Alpha-SNAPSHOT, Commit: b05ff8972a474df52db7854b3753fcc59a9c2b21
I0528 11:31:02.547369       1 kiali.go:45] DEBUG: Kiali: Command line: [/opt/kiali/kiali -config /kiali-configuration/config.yaml -v 4]
I0528 11:31:02.547389       1 config.go:232] DEBUG: Reading YAML config from [/kiali-configuration/config.yaml]
I0528 11:31:02.547984       1 kiali.go:65] Kiali: Console version: 0.2.10-SNAPSHOT.1149
I0528 11:31:02.549705       1 server.go:51] Server endpoint will start at [:20001]
I0528 11:31:02.549733       1 server.go:52] Server endpoint will serve static content from [/opt/kiali/console]
I0528 11:31:05.680298       1 graph.go:78] DEBUG: Build roots (root destination services nodes) for [cytoscape] namespace graph with options [{Appenders:[{} {Duration:1m0s Quantile:0.95 QueryTime:1527507065} {} {} {}] Duration:1m0s Metric:istio_request_count Namespace:istio-system QueryTime:1527507065 Service: Vendor:cytoscape VendorOptions:{GroupByVersion:true Timestamp:1527507065}}]
[...]

PR #200, opened 2018-05-18
KIALI-744 badge orphan services

jshaughn:
This is the necessary server-side change.  There will also be a ui-side follow-on.

PR #560, opened 2018-10-12
KIALI-1714 Limit the lower bound of Prometheus queries

israel-hdez:
Lower bound is set only if needed, in order to not query time ranges that don't fall into the time of life of the namespace whose data is being queried.

Both, prometheus query ranges and rate intervals are subject to be adjusted. For instant queries, the rate interval can be adjusted to not let fetch data before the creation date of the namespace.

For range queries, both the start time of the query and the rate interval could be adjusted. The rate interval is only changed if, from the "end date" of the range, the query would try to fetch data before the creation date of the namespace. The start date will be adjusted if, in combination with the rate interval, the query would fetch data before the creation date of the namespace.

https://issues.jboss.org/browse/KIALI-1714


rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#180](https://paste.fedoraproject.org/paste/SpogfBhSBuTX~VuIBWMIIQ)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1604]](https://paste.fedoraproject.org/paste/ZTr26JupUZZgns9A3byV0g)
  * :heavy_check_mark: Istio-Upstream [#[456]](https://paste.fedoraproject.org/paste/XIRwBjWuKrK4fxzh3ZzvTQ)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1771, 1772]](https://paste.fedoraproject.org/paste/VxgZKiBiIaNRfR7PhOnu-w)
  * :heavy_check_mark: install-bookinfo [#[977]](https://paste.fedoraproject.org/paste/0WqefoLCV1nwyeEK2v9Ztw)
  * :heavy_check_mark: Kiali-on-Openshift [#[991]](https://paste.fedoraproject.org/paste/JbVRu3du~UtPZAdAHf2UNg)
  * :heavy_check_mark: run-kiali-e2e-tests [#[402]](https://paste.fedoraproject.org/paste/mFOgyvGo1369dYm5fjf15A)


KIALI_NAME=kialicorepr560-istio-system

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#181](https://paste.fedoraproject.org/paste/7hvkcqhPecqi23cYNO7wwQ)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1607]](https://paste.fedoraproject.org/paste/NbLcX8WREhhQbivHOdHLxQ)
  * :heavy_check_mark: Istio-Upstream [#[459]](https://paste.fedoraproject.org/paste/X5IchIStLG-ZqOcU~57WlQ)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1777]](https://paste.fedoraproject.org/paste/JIHCO7PdNB3o6OWQfqZ5PQ)
  * :heavy_check_mark: install-bookinfo [#[980]](https://paste.fedoraproject.org/paste/SI17pqNpXhRJIStrYxZQWA)
  * :red_circle: Kiali-on-Openshift [#[994]](https://paste.fedoraproject.org/paste/Bik6JCv3bxevVbPjoIYzrA)


KIALI_NAME=kialicorepr560-istio-system

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#182](https://paste.fedoraproject.org/paste/UxgjxiRRNgdmnEZo-lA6gA)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1608]](https://paste.fedoraproject.org/paste/mGe5fHN92b8YH94m8y-skA)
  * :heavy_check_mark: Istio-Upstream [#[460]](https://paste.fedoraproject.org/paste/bj590uAmHexLgztccj~R2Q)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1778, 1779]](https://paste.fedoraproject.org/paste/iGhy2sgNzdWJ7kdazIBd7w)
  * :heavy_check_mark: install-bookinfo [#[981]](https://paste.fedoraproject.org/paste/n84MTw0BK3iovt3gVaa0nQ)
  * :heavy_check_mark: Kiali-on-Openshift [#[995]](https://paste.fedoraproject.org/paste/F-Jv-DP6V~UV40MfbN2JxA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[405]](https://paste.fedoraproject.org/paste/5qFuD4CqPQIAqCTl1WNH~Q)


KIALI_NAME=kialicorepr560-istio-system

prometheus/metrics.go
```diff
@@ -12,7 +13,9 @@ import (
 	"github.com/prometheus/common/model"
 
 	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
```
jotak: Having the prometheus client depending on the k8s client is somewhat annoying ... they are both meant to be kind of 'low level' parts of kiali, hence autonomous.

Would it be possible to move the related parts to an upper level?
For instance, iiuc the kubernetes client is mostly needed to adjust the query range low bound, and in `getMetrics` we have the `q *MetricsQuery` object that contains the range, so this object could be updated before being sent to the prom client.
This way, prom client still is agnostic of any k8s consideration.

Another advantage is that it would allow some optimization, if metrics are fetched several times for the same namespace, to fetch the namespace creation time only once.

jotak: Note also that in `handlers/metrics.go`, the start time is adjusted to be a multiple of `step`, to avoid so-called "jumpy metrics effect"
https://github.com/kiali/kiali/blob/master/handlers/metrics.go#L85

I believe the range adjustment that takes place here should be performed _before_ the former.

prometheus/metrics.go
```diff
@@ -459,52 +495,104 @@ func getAllRequestRates(api v1.API, namespace string, ratesInterval string) (mod
...
+
+	if startTime.Before(namespaceData.CreationTimestamp.Time) {
+		interval := queryTime.Sub(namespaceData.CreationTimestamp.Time)
+		return fmt.Sprintf("%vs", int(interval.Seconds())), nil
+	}
+
+	return ratesInterval, nil
+}
+
+func getStartForInterval(baseTime time.Time, ratesInterval string) (time.Time, error) {
```
jotak: I wonder if the prometheus go library already has a method to parse rate interval strings. @simonpasquier certainly knows?

simonpasquier: You're probably looking for [`github.com/prometheus/common/model.Duration`](https://godoc.org/github.com/prometheus/common/model#Duration).

simonpasquier: more specifically `ParseDuration`
https://github.com/prometheus/common/blob/master/model/time.go#L182

graph/options/options.go
```diff
@@ -47,7 +47,7 @@ type VendorOptions struct {
```
jshaughn: @israel-hdez , I haven't looked too deeply at the metrics side of things, I wanted to start with the graph code.  My initial comment is that there may be too much logic being deferred to the lower levels of the code.  I may be wrong but I think we could handle most of it right here in the options pre-processing.    We know here which namespaces are requested, as well as the queryTime (i.e. endTime) and the duration.  The difference now is that we may have a different duration for each namespace, right?   Instead of AccessibleNamespaces having the side-effect of carrying namespace creation time, maybe keep that info local to this file and instead change Options.Namespaces to a map[string]time.Duration (or maybe it could be a map to new struct like map[string]NamespaceInfo). Then populate the map by calculating a valid duration per requested namespace (based on the requested duration, queryTime and namespace creation time).  You can then remove Options.Duration.   If you do this I think the downstream impact would be almost negligible.  The only difference would be replacing accesses to options.Duration with something like options.Namespaces[<namespace>].Duration.

What do you think?

israel-hdez: @jshaughn Yes, indeed with your suggestion makes sense. It will minimize changes in the inner levels of the code.

I was trying to see the code with a 3-tier architecture, since there already are some "business" and "models" packages. In my point of view, the options parsing step is still part of the "presentation" tier and, probably, that was the reason I didn't do it here (I was trying to not mix business with presentation logic).

But... here in the options the namespaces are already being filtered out if they are not accessible which I see also as part of "business logic". So, I think it doesn't hurt if I move the logic to determine per-namespace duration to the parsing of the options. I'll do that.

israel-hdez: @jshaughn Changed!... I'll think how to change the metrics part.

prometheus/metrics.go
israel-hdez: Right... You had the same comment as @jshaughn. I will move the range adjustment to the extractMetricsQueryParams.

But I have the same observation: this lead to mixing some small "business logic" in a part of the code that I see belonging to the "presentation tier".

And yes, you are right in that the adjustment of the start time to be a multiple of `step` should happen after.

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#189](https://paste.fedoraproject.org/paste/D4vn4YpVp1c7imn-rtNbMA)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1621]](https://paste.fedoraproject.org/paste/uYDI4vfgo1ZVMyhPKeN37A)
  * :heavy_check_mark: Istio-Upstream [#[470]](https://paste.fedoraproject.org/paste/Z3rN6FE5tvEIgIU1sO6cxg)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1806, 1807]](https://paste.fedoraproject.org/paste/GjQsV7PWFbFNhErZL17tHA)
  * :heavy_check_mark: install-bookinfo [#[995]](https://paste.fedoraproject.org/paste/l46odOMTjiM30Pv4NZ3pjw)
  * :heavy_check_mark: Kiali-on-Openshift [#[1010]](https://paste.fedoraproject.org/paste/fmNn0BjeZSMKQ5mZGKzjTA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[418]](https://paste.fedoraproject.org/paste/sEbk-Iz4Pz5zxBAsXYGxxg)


KIALI_NAME=kialicorepr560-istio-system

prometheus/metrics.go
jotak: I see the problem, for lists / details queries we've introduced the business layers that sits between handlers and the underlying clients. I don't know how easy it would be to move some graph code in this layer, but maybe it's the solution? (If yes, I let you judge if it has to be part of a bigger refactoring task or not...)

graph/options/options.go
jshaughn: I understand your thinking about the tiers.  And it can be a little confusing because the graph code and metrics code grew somewhat independently.  As for the graph code, I guess I don't really consider any of the server-side code as part of the presentation layer.   I think of it mainly as a separation of concerns.  "Options" is concerned with describing the graph to be generated, it is the nexus between the user's request and the graph code.  It can immediately fail given an impossible request, or it can massage the request as needed.  The graph code is responsible for generating the logical graph topology and should not be concerned with the graph definition.  The appenders are responsible for optional decoration of the graph, and the vendor code (we only have the one cytoscape vendor now, but we used to have vizceral) translates the graph into a particular format for consumption by the caller.

graph/appender/sidecars_check.go
```diff
@@ -34,7 +34,7 @@ func (a *SidecarsCheckAppender) applySidecarsChecks(trafficMap graph.TrafficMap,
 	for _, n := range trafficMap {
 
 		// Skip the check on the sidecars if we don't have access to the namespace
-		if _, found := a.AccessibleNamespaces[n.Namespace]; !found {
+		if _, found := a.Namespaces[n.Namespace]; !found {
```
jshaughn: I think there is a mistake here.  Maybe my original comment was confusing.  When I suggested not adding duration information to AccessibleNamespaces I did not mean remove AccessibleNamespaces.  I think you need to keep the use of AccessibleNamespaces as it was before.  There are two things going on.  First there are the namespaces requested for the graph, say A and B.   Then there are the AccessibleNamespaces for the user, say A, B, C.

First, if a requested namespace is not in the set of accessible namespaces the request should fail fast in the Options processing.  But remember now that when we generate a graph for A and B it can have leaf nodes outside of A and B, those nodes either request services in A or B, or are requested by services in A or B.  For those outsider nodes we need to know, in various places, if the user has full access.  This is one of those places.

So, basically, a,Namespaces deals with the *requested* namespaces (all of which must be accessible), but is only a subset of AccessibleNamespaces.

graph/appender/sidecars_check.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 type SidecarsCheckAppender struct {
-	AccessibleNamespaces map[string]bool
```
jshaughn: I think you still need this, see my longer comment below.

graph/graph.go
```diff
@@ -41,6 +43,11 @@ type Edge struct {
 	Metadata map[string]interface{} // app-specific data
 }
 
+type NamespaceInfo struct {
+	models.Namespace
```
jshaughn: Do we really need models.Namespace or could it just be the name, as before?  I don't think anything but the name is used outside of Options processing.

graph/options/options.go
```diff
@@ -47,14 +48,13 @@ type VendorOptions struct {
 
 // Options are all supported graph generation options.
 type Options struct {
-	AccessibleNamespaces map[string]bool
```
jshaughn: I think you still need this, see my related comment.

graph/options/options.go
```diff
@@ -132,19 +102,68 @@ func NewOptions(r *http.Request) Options {
...
+	}
+
 	// Service graphs require service injection
 	if graphType == graph.GraphTypeService {
 		injectServiceNodes = true
 	}
 
+	// Build options struct
 	options := Options{
-		AccessibleNamespaces: accessibleNamespaces,
```
jshaughn: Same comment, I think we need AccessibleNamepaces.

handlers/graph.go
```diff
@@ -133,26 +133,26 @@ func markOutsiders(trafficMap graph.TrafficMap, o options.Options) {
...
-		if n.Namespace == ns {
+		if n.Namespace == ns.Name {
 			return false
 		}
 	}
 	return true
 }
 
-func isInaccessible(n *graph.Node, accessibleNamespaces map[string]bool) bool {
+func isInaccessible(n *graph.Node, accessibleNamespaces map[string]graph.NamespaceInfo) bool {
```
jshaughn: I think this should go back to using the original AccessibleNamespaces map.

handlers/graph.go
```diff
@@ -632,15 +633,20 @@ func graphNode(w http.ResponseWriter, r *http.Request, client *prometheus.Client
...
 
-	namespace := o.Namespaces[0]
-	n := graph.NewNode(namespace, o.NodeOptions.Workload, o.NodeOptions.App, o.NodeOptions.Version, o.NodeOptions.Service, o.GraphType)
+	// Here, it's true that o.Namespaces has only one item. So, it's safe to use "for" knowing
+	// that only one iteration will happen.
+	var n graph.Node
+	var namespace graph.NamespaceInfo
+	for _, namespace = range o.Namespaces {
+		n = graph.NewNode(namespace.Name, o.NodeOptions.Workload, o.NodeOptions.App, o.NodeOptions.Version, o.NodeOptions.Service, o.GraphType)
+	}
```
jshaughn: Not a big deal, but why replace the 2 lines with 7?

graph/appender/sidecars_check.go
israel-hdez: No, I misunderstood the usage of `AccessibleNamespaces`. I thought that both will have the same items, but `AccessibleNamespaces` can have more items. I'll fix this.

graph/graph.go
israel-hdez: Hmmm... Yes, it can be only the name. I'll change.

handlers/graph.go
israel-hdez: Since `o.Namespaces` is no longer an array, it's not possible to do namespace := o.Namespaces[0].
Perhaps, by removing the comments it will be only 5 lines?

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#198](https://paste.fedoraproject.org/paste/q0CoucBZXJTpQNH3~yIGdw)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1638]](https://paste.fedoraproject.org/paste/pOTXcUbgWN4IbTxH6yiPUg)
  * :heavy_check_mark: Istio-Upstream [#[486]](https://paste.fedoraproject.org/paste/we67rq8ljEQ9xG208IvdgA)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1842, 1843]](https://paste.fedoraproject.org/paste/3bSuvD~6~IJD8oZy53DyAQ)
  * :heavy_check_mark: install-bookinfo [#[1013]](https://paste.fedoraproject.org/paste/85dF-eK8TAWEpYHGZzKdaA)
  * :heavy_check_mark: Kiali-on-Openshift [#[1029]](https://paste.fedoraproject.org/paste/kRPQxziwKUAZwGLI-d1zeA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[435]](https://paste.fedoraproject.org/paste/kGPLF2EIJJLgJc64Q3MlFw)


KIALI_NAME=kialicorepr560-istio-system

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#199](https://paste.fedoraproject.org/paste/eO4Skp6JXjaqKzhwusZSyQ)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1639]](https://paste.fedoraproject.org/paste/0etYyJjw5bLR0ONojudyKg)
  * :heavy_check_mark: Istio-Upstream [#[487]](https://paste.fedoraproject.org/paste/~a5iZJMowe~H7Oow3mngUQ)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1844, 1845]](https://paste.fedoraproject.org/paste/PV9-qItLCkzDrayEaHJ3gg)
  * :heavy_check_mark: install-bookinfo [#[1014]](https://paste.fedoraproject.org/paste/41PA5LEOALf398QsIEjYeQ)
  * :heavy_check_mark: Kiali-on-Openshift [#[1030]](https://paste.fedoraproject.org/paste/Dvdm-3Sh7OdpfFpsuIj4Yg)
  * :heavy_check_mark: run-kiali-e2e-tests [#[436]](https://paste.fedoraproject.org/paste/npQ~22EB9Q-kJx3SS75XZA)


KIALI_NAME=kialicorepr560-istio-system

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#200](https://paste.fedoraproject.org/paste/znIzrZgqg2uZcQYuVhy~lw)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1640]](https://paste.fedoraproject.org/paste/zhoDWFYr4X0pcnu1UO0sSA)
  * :heavy_check_mark: Istio-Upstream [#[488]](https://paste.fedoraproject.org/paste/abnRFv6Z-x1~xcvPMiKuTg)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1846, 1847]](https://paste.fedoraproject.org/paste/0NQ7MKp86Pe2cYMw1W0tmw)
  * :heavy_check_mark: install-bookinfo [#[1015]](https://paste.fedoraproject.org/paste/kGEJtMCBbA~iX3WXMz7cBg)
  * :heavy_check_mark: Kiali-on-Openshift [#[1031]](https://paste.fedoraproject.org/paste/YczW-b-REsWXaz8qvHEVCA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[437]](https://paste.fedoraproject.org/paste/pTyL0XYb1xt11wht1v5uSA)


KIALI_NAME=kialicorepr560-istio-system

handlers/metrics.go
```diff
@@ -80,6 +84,37 @@ func extractMetricsQueryParams(r *http.Request, q *prometheus.MetricsQuery) erro
...
+	}
+	if intervalStartTime.Before(namespace.CreationTimestamp.Time) {
+		q.RateInterval = fmt.Sprintf("%vs", int(q.End.Sub(namespace.CreationTimestamp.Time).Seconds()))
+		intervalStartTime = namespace.CreationTimestamp.Time
+		log.Debugf("[extractMetricsQueryParams] Interval set to: %v", q.RateInterval)
+	}
+	// If needed, adjust query start time (bound to namespace creation time)
+	log.Debugf("[extractMetricsQueryParams] Requested query start time: %v", q.Start)
+	intervalDuration := q.End.Sub(intervalStartTime)
+	allowedStart := namespace.CreationTimestamp.Time.Add(intervalDuration)
```
jotak: I have a doubt here, why allowedStart isn't just namespace.CreationTimestamp.Time ?
In other words, why not just:

```
if q.Start.Before(namespace.CreationTimestamp.Time) {
   q.Start = namespace.CreationTimestamp.Time
```

I must be missing something?

handlers/metrics.go
```diff
@@ -80,6 +84,37 @@ func extractMetricsQueryParams(r *http.Request, q *prometheus.MetricsQuery) erro
...
+	if err != nil {
+		return err
+	}
+	// If needed, adjust interval -- Make sure query won't fetch data before the namespace creation
+	intervalStartTime, err := util.GetStartTimeForRateInterval(q.End, q.RateInterval)
+	if err != nil {
+		return err
+	}
+	if intervalStartTime.Before(namespace.CreationTimestamp.Time) {
+		q.RateInterval = fmt.Sprintf("%vs", int(q.End.Sub(namespace.CreationTimestamp.Time).Seconds()))
```
jotak: Not a big deal (I won't request change for that), but the advantage of using accurate formatting types like "%d" "%s" etc. instead of the catch-all solution %v , is that if there's a refactoring later, if by mistake wrong parameters types being changed and mistakenly passing a wrong type, the problem will be caught earlier ; and tools such as govet or golint (can't remember which) would report this issue. So IMO that's a good practice.

handlers/metrics.go
israel-hdez: Because of the _interval_. We don't want to query data logged before the namespace creation date. Setting `q.Start = namespace.CreationTimestamp.Time` may lead to prometheus evaluate the query at `namespace.CreationTimestamp.Time` and using data for an `interval` past that date. This is my reasoning.

From reading the docs, I understood that `q.Start` and `q.End` only sets a range to evaluate the query in steps of length `q.Step`. This doesn't really set "boundaries" when evaluating a query and fetching data. But, I may be misunderstanding how prometheus evaluates queries. So, If you think that `q.Start = namespace.CreationTimestamp.Time` is enough, I'll change.

handlers/metrics.go
israel-hdez: Ok! I'll change the format types.

handlers/metrics.go
jotak: Ok I see the problem.

But it creates another problem, it also means that when you create a new namespace and immediately start to look at it in kiali, you will always miss the first datapoints (or even get errors), until the delay of "rate interval" is passed, right? Which is probably ok for a 1m rate interval, but could be annoying for something bigger like 30m.

israel-hdez: Probably, that case is already covered by the combination of adjusting the rate interval (some lines before this one) and adjusting the start date. The "rate interval" should be adjusted to the time of live of the namespace and here, the the start of the range should be set so that `q.Start == q.End`. And that will, at least, return some data.

I don't know what's better. So, if you think that is better to use irate, I'll change this.

graph/options/options.go
```diff
@@ -227,7 +244,7 @@ func parseAppenders(params url.Values, o Options) []appender.Appender {
 
 // getAccessibleNamespaces returns a Set of all namespaces accessible to the user.
 // The Set is implemented using the map[string]bool convention.
-func getAccessibleNamespaces() map[string]bool {
+func getAccessibleNamespaces() map[string]time.Time {
```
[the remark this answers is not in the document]
israel-hdez: The `Options` struct now has a `Namespaces` (type `map[string]NamespaceInfo`) and an `AccessibleNamespaces` (type `map[string]bool`) fields. I need to fill both in the `NewOptions` method. In order to fill the `Namespaces` field I need the timestamps to calculate the duration of each namespace. So, I choose to return the time, instead of booleans which are always `true`. This saves another call to the cluster API to only get timestamps -- And this should answer your other comment.

I can restore the map of bools. But if we want to save calls to the cluster API, this method is the best place to work on; I would still need that `map[string]time.Time` in addition to the `map[string]bool` one. So, this method would return two maps instead of one --- And this way that little additional logic you also commented can go away.

graph/options/options.go
```diff
@@ -132,17 +102,63 @@ func NewOptions(r *http.Request) Options {
...
 	if graphType == graph.GraphTypeService {
 		injectServiceNodes = true
 	}
 
+	allowedNamespaces := make(map[string]bool, len(accessibleNamespaces))
+	for ns := range accessibleNamespaces {
+		allowedNamespaces[ns] = true
+	}
+
+	// Build options struct
```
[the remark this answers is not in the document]
israel-hdez: I reverted this block of code.

graph/options/options.go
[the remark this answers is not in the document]
israel-hdez: Ok... I reverted, but also, as I said, I changed this method to return two maps instead of only one.

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#219](https://paste.fedoraproject.org/paste/6mFHRPD3Kc5ruNzrE3wZIw)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1674]](https://paste.fedoraproject.org/paste/Bg0neZe6cU9kzHPKB-ZYIQ)
  * :heavy_check_mark: Istio-Upstream [#[520]](https://paste.fedoraproject.org/paste/iWxnuqoSI6ZwHa4nx~3MKw)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1922, 1923]](https://paste.fedoraproject.org/paste/cedJ0ThzyiaBAsR~rmIgWA)
  * :heavy_check_mark: install-bookinfo [#[1049]](https://paste.fedoraproject.org/paste/eTxY8JZ4pe4mUczhLZGW-A)
  * :heavy_check_mark: Kiali-on-Openshift [#[1067]](https://paste.fedoraproject.org/paste/NBe0CDn~1X9s16HjDhWK3g)
  * :heavy_check_mark: run-kiali-e2e-tests [#[474]](https://paste.fedoraproject.org/paste/LM2t3579NrpkE6AeWTZuww)


KIALI_NAME=kialicorepr560-istio-system

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#220](https://paste.fedoraproject.org/paste/vgbfI6tH7p3ngtFLiz4IRw)
  * :heavy_check_mark: Istio-and-Kiali-on-Openshift [#[1675]](https://paste.fedoraproject.org/paste/tYQZ5NB0~fP67C1gA-gMxA)
  * :heavy_check_mark: Istio-Upstream [#[521]](https://paste.fedoraproject.org/paste/vUyXhzhUUBnzEYXTI1fLUw)
  * :heavy_check_mark: Kiali-Istio-Checker [#[1924, 1925]](https://paste.fedoraproject.org/paste/p7QGzNkyp~Knb49K84Y~LA)
  * :heavy_check_mark: install-bookinfo [#[1050]](https://paste.fedoraproject.org/paste/e8vlLG9FvupnlfjNn~Evyg)
  * :heavy_check_mark: Kiali-on-Openshift [#[1068]](https://paste.fedoraproject.org/paste/RKNKBNbqmeOFwWuGPDLhiA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[475]](https://paste.fedoraproject.org/paste/zYO6ceeKoA39Sd06R47-kA)


KIALI_NAME=kialicorepr560-istio-system

graph/options/options.go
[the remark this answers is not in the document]
jshaughn: OK, I understand now why you were assigning timestamps to the accessibleNamespaces.  It can be useful to add an inline comment when performing side-effects/optimizations.  Without the explanation it looked to me like you had added some sort of time logic to AccessibleNamespaces and I was confused.  We can leave it as is but if you want to change it back with a comment that is fine as well.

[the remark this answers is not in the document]
israel-hdez: I'll leave it as is for now. I like that it has less loops.

israel-hdez:
@jshaughn Rebased.
I was going to merge, but the rebase dropped your approval :(

rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#230](https://paste.fedoraproject.org/paste/eFPnzF3mz5AHcjlCkfe2Cg)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1721]](https://paste.fedoraproject.org/paste/BgMiNXUSIhukZuSisyID2A)
  * :heavy_check_mark: Istio-Upstream [#[569]](https://paste.fedoraproject.org/paste/YpixSkyjkdyPulOp2fSo8A)
  * :heavy_check_mark: Kiali-Istio-Checker [#[2006]](https://paste.fedoraproject.org/paste/YdKsZFOScSi8Ds03W-CQZw)
  * :red_circle: install-bookinfo [#[1091]](https://paste.fedoraproject.org/paste/yA9oTU5CVNbvHD1KTTOY6w)


KIALI_NAME=kialicorepr560-istio-system

jotak:
+1, I didn't have time to investigate more on the adjustment algorithm, I think I don't totally understand everything but don't block the PR because I run out of time

PR #563, opened 2018-10-15
KIALI-1497 Create liveliness and ready probes for Kiali pod (fixes #540)

### LINKED ISSUE #540: Create liveliness and ready probes for Kiali pod (hacktober fest)
**Describe the bug**
Create liveliness and ready probes for Kiali pod
Health probes documentation: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/

Kiali pod definition:
- Kubernetes:  https://github.com/kiali/kiali/blob/master/deploy/kubernetes/kiali.yaml
- OpenShift: https://github.com/kiali/kiali/blob/master/deploy/openshift/kiali.yaml

matejgera:
…

** Describe the change **

The change adds liveness and readiness probes for the Kiali pods for Kubernetes and OpenShift. It does so by probing path "/api/status".

** Issue reference **

KIALI-1497 

** Backwards incompatible? **

- [ ] Is your change introducing backwards incompatible changes? - No

### DIFF
--- a/deploy/kubernetes/kiali.yaml
+++ b/deploy/kubernetes/kiali.yaml
@@ -71,6 +71,20 @@ spec:
         - "/kiali-configuration/config.yaml"
         - "-v"
         - "${VERBOSE_MODE}"
+        readinessProbe:
+          httpGet:
+            path: /api/status
+            port: 20001
+          initialDelaySeconds: 3
+          periodSeconds: 3
+          timeoutSeconds: 45
+        livenessProbe:
+          httpGet:
+            path: /api/status
+            port: 20001
+          initialDelaySeconds: 3
+          periodSeconds: 3
+          timeoutSeconds: 45
         env:
         - name: ACTIVE_NAMESPACE
           valueFrom:
--- a/deploy/kubernetes/kiali.yaml
+++ b/deploy/kubernetes/kiali.yaml
@@ -77,14 +77,12 @@ spec:
             port: 20001
           initialDelaySeconds: 3
           periodSeconds: 3
-          timeoutSeconds: 45
         livenessProbe:
           httpGet:
             path: /api/status
             port: 20001
           initialDelaySeconds: 3
           periodSeconds: 3
-          timeoutSeconds: 45
         env:
         - name: ACTIVE_NAMESPACE
           valueFrom:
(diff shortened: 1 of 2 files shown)


rhqci:

pilhuhn:
I think that looks good (haven't tried it though). In #540 you say "It takes 30-45 seconds to load". I would expect that this only applies to the readiness probe until Kiali is fully deployed. The liveness probes to see if it is still alive should be much quicker - or did you ever encounter something >> 3 secs?

matejgera:
Apologies, it seems the wait time to load the status was probably caused by insufficient resources in my testing cluster when deploying Istio. After re-testing it again today the response is instant and there is therefore no need to override the default value for `timeoutSeconds`. Please see the latest commit in my pull request.

matejgera:
@israel-hdez thank you for that important information, I now commited changes reflecting this fact, i.e. the specification for OpenShift liveness/readiness probe now expclitly sets `scheme: HTTPS`

israel-hdez:
@matejgera Thanks for your contribution.

matejgera:
@israel-hdez thank you, if you have suggestion on aspects which need improving I'll be happy to work on it further

israel-hdez:
@matejgera Basically, the improvement would be to have a proper endpoint for the probes. The existent `/api/status` endpoint is a good one, but it queries the cluster api, prometheus and istio to get their versions. And having these data fetched every 3 seconds sounds too much - especially if the returned data is not used.

Moreover, if there is some error fetching the versions, `/api/status` will still return success. That's fine, because, even if dependencies are down, Kiali itself is "ready". So, there is really no point in fetching versions of the dependencies that frequently. It's only adding overhead in the cluster.

matejgera:
@israel-hdez I see, this is what I was touching upon in the other thread (https://github.com/kiali/hacktoberfest18/issues/9). If I understand you correctly, that would require creating a completely new API endpoint to serve this purpos (something like `/api/health` ?) which would not unnecessarily query all the other service to simply get their version.

israel-hdez:
@matejgera Yes, that's the idea.
We already use "health" with other meaning, so you may want to use another term if you decide to work on this.

jmazzitelli:
We need to change this. We are getting spammed in the kiali logs every couple settings with the debug message about the versions being detected. We don't want to be calling status/ just to determine health and wellness of the app - its causing Kiali to not only spam the logs but now we are forced to collect all versions of the products, parse the version strings, and return json.

We should have an endpoint that simply returns 4xx or 5xx status codes and that's it.

Issue #578, opened 2018-10-22
Support Envoy stats natively


liamawhite:
Currently, Kiali only supports stats that mixer produces. As Envoy will be being scraped directly in the near future (and for mixerless deployments) Kiali should support Envoy stat labels as scraped by Prometheus.

abonas:
@pilhuhn @lucasponce @mwringe we should move this to JIRA.
@liamawhite thank you for the notice! Do you have any information/point to docs/timeline regarding the change of Envoy being scraped directly?

jmazzitelli:
Does this mean in the near future it is possible for the istio-* metrics to be missing/go away? That would be bad because those are the metrics that Kiali uses to build the graph visualizations.

jotak:
Oh, I see the Envoy stats are now labelled in a "friendly" way (in 1.1.0 snapshot). That's very nice!

Ex:
`envoy_cluster_membership_healthy{app="stadium",cluster_name="BlackHoleCluster",instance="172.17.0.19:15090",job="envoy-stats",namespace="mesh-arena",pod_name="stadium-small-66888649cb-rsjk5",pod_template_hash="2244420576",version="small"}`

@liamawhite thanks for letting us know. I would like to understand what the `cluster_name` refers to, I can see three of them for every pod: `BlackHoleCluster`, `xds-grpc` and `zipkin`. I wonder if it makes sense to show the aggregate of all in Kiali or if some selection must be made.
Also, there's no reference of any workload, is it intentional? We could retrieve the workload based on pod name but it adds some cost.

We actually do already show some envoy stats but it's restricted about cluster membership as part of a service live health, it's not displayed as a timeseries.
We already have a JIRA to show timeseries about opened CB or such, iirc.

By the way, beyond cluster membership healthy versus total, do you think some stats should be particularly relevant to show?

@jmazzitelli I guess there's absolutely no impact on usual istio stats (at least I don't see any in 1.1.0-snap)

liamawhite:
I may have misunderstood an earlier discussion with some of the telemetry team. Let me get back to you on the specifics or just how we're going to do this. I think it may be a combination of the two.

PR #283, opened 2018-06-14
Move config and Readme to Istio 8.0

aljesusg:
Set istio 8.0 in configuration and Readme

PR #604, opened 2018-10-29
KIALI-1813 / istio-1.0.4 Envoy Health Membership model changed

jotak:
- Envoy metrics are now more prometheus-friendly, with labels
- Now related to app instead of services. This makes it a lot easier to manage (lot of code removed because of that)
- Some improvements in mocks/tests

UI PR: https://github.com/kiali/kiali-ui/pull/775/files

JIRA: https://issues.jboss.org/browse/KIALI-1813

This change **is backward incompatible with Istio < 1.1.0**

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/prometheus"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
+	"github.com/kiali/kiali/util"
 )
 
 // HealthService deals with fetching health from various sources and convert to kiali model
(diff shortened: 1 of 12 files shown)


rhqci:
:heavy_check_mark: Jenkins CI: kiali-core-pr-e2e-test [#283](https://paste.fedoraproject.org/paste/Z-vTgwyrk4EYLOaXncOMLw)
  * :red_circle: Istio-and-Kiali-on-Openshift [#[1815]](https://paste.fedoraproject.org/paste/vvoK8~Vv6zKx8i4TAhmwRQ)
  * :heavy_check_mark: Istio-Upstream [#[656]](https://paste.fedoraproject.org/paste/DWlywNqo8TQdvXhms6Idsg)
  * :heavy_check_mark: Kiali-Istio-Checker [#[2180, 2181]](https://paste.fedoraproject.org/paste/jGb10z83JESBS5KSoy9msg)
  * :heavy_check_mark: install-bookinfo [#[1194]](https://paste.fedoraproject.org/paste/Xg9m7eQL85MRj15oED1vpw)
  * :heavy_check_mark: Kiali-on-Openshift [#[1189]](https://paste.fedoraproject.org/paste/PuiDGXqmZ2ItDHXDLtbHTQ)
  * :red_circle: run-kiali-e2e-tests [#[587]](https://paste.fedoraproject.org/paste/RLjbpIUyN7sNDi4HsPk67w)


KIALI_NAME=kialicorepr604-istio-system

jotak:
I'm going to refresh this one: it seems that the envoy change is live in istio 1.0.4, they haven't wait for 1.1

lucasponce:
@jotak starting to review this one, sorry for my delay

business/health.go
```diff
@@ -22,133 +23,73 @@ type HealthService struct {
...
-				ports = append(ports, port.Port)
-			}
-			envoy, _ := in.prom.GetServiceHealth(namespace, service.Name, ports)
-			health.Envoy = append(health.Envoy, models.EnvoyHealthWrapper{
-				Service:            service.Name,
-				EnvoyServiceHealth: envoy,
-			})
-		}(service)
-	}
+	envoy, errEnvoy := in.prom.GetEnvoyMembershipRatio(namespace, app)
```
lucasponce: So, the services per app relation is included in the GetEnvoyMembershipRatio ?

business/health.go
```diff
@@ -22,133 +23,73 @@ type HealthService struct {
...
-	health.Requests = in.getAppRequestsHealth(namespace, app, rateInterval, queryTime)
+	rate, errRate := in.getAppRequestsHealth(namespace, app, rateInterval, queryTime)
+	health.Requests = rate
 
 	// Deployment status
 	health.WorkloadStatuses = castWorkloadStatuses(ws)
 
-	wg.Wait()
-	return health
+	return health, util.CoalesceErrors(errEnvoy, errRate)
```
lucasponce: I suspect that for non-english speakers, perhaps "CombineErrors" can be more clear :-).
I didn't use "Coalesce" verb before today.

prometheus/metrics.go
```diff
@@ -20,89 +19,49 @@ var (
...
-func getServiceHealth(api v1.API, namespace, servicename string, ports []int32) (EnvoyServiceHealth, error) {
-	ret := EnvoyServiceHealth{}
-	if len(ports) == 0 {
-		return ret, nil
-	}
+func getEnvoyMembershipRatio(api v1.API, namespace, app string) (EnvoyMembershipRatio, error) {
+	ret := EnvoyMembershipRatio{}
 
 	// In these Prometheus queries, only the last values are of interest. Since the
 	// <servicename> is received as a parameter from the frontend, we can assume the
```
lucasponce: Perhaps this comment is not updated with the query it performs below ?

business/health.go
jotak: To be honest, I've never seen it used elsewhere than in code. Originally from SQL syntax: https://www.w3schools.com/sql/func_mysql_coalesce.asp
and from time to time I also see it used in some libs.
"Combine" wouldn't be very accurate, it's not really a combination, however I can rename `FirstNonNil` if it's better.

jotak: See also: https://en.wikipedia.org/wiki/Null_coalescing_operator

lucasponce: "FirstNonNil" sounds more descriptive to me.
I suggested "combine" as synonym of "coalesce" given by Collins dic.
But as perhaps not everyone has used this operator "FirstNonNil" is clear what it does.

lucasponce:
Another comment:
image

So, perhaps this new value is counting inbound + outbound 'sidecar' per pod ?

prometheus/metrics.go
```diff
@@ -20,89 +19,49 @@ var (
...
-			ret.Inbound.Total += int(v)
-		case v := <-outboundHealthyChan:
-			ret.Outbound.Healthy += int(v)
-		case v := <-outboundTotalChan:
-			ret.Outbound.Total += int(v)
-		case err = <-errChan:
-			// No op
+	go func() {
+		defer wg.Done()
+		vec, err := fetchTimestamp(api, "sum(envoy_cluster_membership_total{"+labels+"})", now)
```
lucasponce: Perhaps this comment https://github.com/kiali/kiali/pull/604#discussion_r243568274 is more appropiate under this line.

lucasponce: Not sure if the sum() is correct here.
Looking at the values:
image
It sounds like we could use just one single value ?

jotak:
tl;dr : I suggest to close this PR and remove "Envoy Health" from our health checks.

So... I spent the afternoon investigating envoy stats. The reason why we have twice the number of pods represented as "Envoy Health" is because for each pod there's two different "cluster names" reported by Envoy. The bad part is that, I don't understand what these cluster names are for, what they really represent, but I don't think they represent really what we want to show: they are "zipkin" and "xds-grpc", for all services in our mesh. Some time ago I already questioned that, but it went out of my mind: see https://github.com/kiali/kiali/issues/578#issuecomment-432118648 .

Further investigation showed that there are more metrics available from Envoy that could be more interesting to show Health, but those metrics are being filtered out by Istio in the default prometheus config. They could be, in theory, interesting for us, they are the ones that really concern our mesh health. Exampe: `envoy_cluster_membership_change{cluster_name="outbound|8080|ball-v2|ball.mesh-arena.svc.cluster.local"}`

But I say "in theory", because even if I undrop them, I'm not seeing exactly what I would expect - or with some drawback:
- First, we were initially tracking "envoy_cluster_membership_healthy" and "envoy_cluster_membership_total". But for some reason they don't show what I would expect, for instance even with a pod ejected with CB open, it's still reported as healthy. [1]
- We could use another metric that seems to be more reliable: "envoy_cluster_outlier_detection_ejections_enforced_total" : it's a counter showing number of times CB was open, basically. But we need to figure out how it's labelled, because it seems inconsistent, sometimes being reported with the outgoing workload name that's being evicted, sometimes without it:

```
envoy_cluster_outlier_detection_ejections_enforced_total{cluster_name="outbound|8080|ball-v2|ball.mesh-arena.svc.cluster.local"} 3
envoy_cluster_outlier_detection_ejections_enforced_total{cluster_name="outbound|8080||ball.mesh-arena.svc.cluster.local"} 36
```
- Also, we must note that the labels "app" and "version" on envoy metrics are defining the source of the request, not the destination, for the metrics that are interesting for us here. So it's not the workload that is actually being unhealthy, but it's the workload that tries to contact an unhealthy one. Hopefully we can still figure out which is the unhealthy one so we could still display health as before.

[1] Maybe a new ticket to open on Envoy repo, for 2019

lucasponce:
@jotak catching up with this one

> tl;dr : I suggest to close this PR and remove "Envoy Health" from our health checks.

I'm +1 if we close this and investigate more in the direction of your previous comment.

My understanding is that the "envoy health" was related if proxy was good or it might have some problem user wants to know in an early phase, but it seems this info is not exactly as we thought.

Just let me know how to proceed.

jotak:
@lucasponce just created https://issues.jboss.org/browse/KIALI-2209

Now closing this

PR #348, opened 2018-07-11
KIALI-1149 Improve verbosiy of no_service checker

lucasponce:

PR #627, opened 2018-11-05
KIALI-1819 Support generic handlers for Istio 1.1.x

lucasponce:
This fix allows to support generic handlers introduced in Istio 1.1.x.
Instead to use a specific type "i.e. kind: prometheus", in Istio 1.1.x there will be a "kind: handler".

Note, this patch allows to support 1.0.x and 1.1.x in the same codebase.
On this scenario is possible.

Potentially it can solve also https://issues.jboss.org/browse/KIALI-1579 but only on 1.1.x onwards.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1819


lucasponce:
cc @jmazzitelli @gbaufake another example of PR that modifies config.
I can create an specific label if that helps to spot them better.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#347](https://paste.fedoraproject.org/paste/tb1d2AYMXguQAi4q5NRRZw)
  * :red_circle: run-kiali-e2e-tests [#[713]](https://paste.fedoraproject.org/paste/SiffG4nJvIfDafyLVHYdJA)

lucasponce:
I've sent another commit in the config to unify the "config.istio.io". 
I think it makes sense to unify them.
Not sure why they were splitted.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#348](https://paste.fedoraproject.org/paste/-ZrVOzqKp0LWqkIKUFX0Eg)
  * :red_circle: run-kiali-e2e-tests [#[714]](https://paste.fedoraproject.org/paste/92Cq7jgRTxsIHemYmIYbUQ)

deploy/kubernetes/kiali.yaml
```diff
@@ -148,12 +149,6 @@ rules:
   - quotas
   - reportnothings
   - servicecontrolreports
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups: ["config.istio.io"]
-  resources:
```
jotak: They were split because of the "delete" verb, we only delete a few objects among them. I think we should still require "delete" permission only for what we really need.

deploy/openshift/kiali.yaml
```diff
@@ -170,12 +171,6 @@ rules:
   - quotas
   - reportnothings
   - servicecontrolreports
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups: ["config.istio.io"]
-  resources:
```
jotak: Same comment here, there was a reason for splitting

deploy/kubernetes/kiali.yaml
lucasponce: Most of these objects are handlers and templates that depends on the rules (which is editable); so, I think it can be unified in the config.

jotak: but are they supposed to be deleted when we delete a rule? I guess not, the delete command needs to be explicit on what we delete... so we'll never delete them.

I don't really like to ask for wider permissions than we actually need, just for having a prettier config file

lucasponce: I'm not fully convinced that we cannot delete adapters/template.
Also, this is affecting the on-going work for update verb.
So, I suggest to maintain this change and we can refine continue discussing when the update is in place as I think we can delete/update adapters/templates inside rules.

aljesusg:
This Fixed https://issues.jboss.org/browse/KIALI-1867 too. Thanks

PR #628, opened 2018-11-06
KIALI-1868 - define the namespace explicitly

jmazzitelli:
This requires a change in the gettingstarted/ docs, too - because it requires NAMESPACE env var to be set when running envsubstr - for that change, see https://github.com/kiali/kiali.io/pull/56

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,7 +270,7 @@ docker-push:
 openshift-deploy: openshift-undeploy
 	@if ! which envsubst > /dev/null 2>&1; then echo "You are missing 'envsubst'. Please install it and retry. If on MacOS, you can get this by installing the gettext package"; exit 1; fi
 	@echo Deploying to OpenShift project ${NAMESPACE}
-	cat deploy/openshift/kiali-configmap.yaml | VERSION_LABEL=${VERSION_LABEL} JAEGER_URL=${JAEGER_URL} GRAFANA_URL=${GRAFANA_URL} envsubst | ${OC} create -n ${NAMESPACE} -f -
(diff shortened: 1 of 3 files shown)


jmazzitelli:
DNM yet - this PR may be deprecated if/when https://github.com/kiali/kiali/pull/629 is merged.

gbaufake:
@jmazzitelli  I think we can merge your PR first then I can mine to adapt this change.

jmazzitelli:
> I think we can merge your PR first then I can rebase mine to adopt this change.

But you probably aren't going to adopt anything in this PR - this PR changes:

1) the makefile targets to pass the namespace to the oc apply command - but you are going to completely change the make targets to invoke helm, not the oc apply command - so this PR changes are going to be completely removed

2) the configmap yaml to include the namespace, but you are going to delete these yaml files entirely

gbaufake:
@jmazzitelli I thought we would keep the openshift part of the files and deprecate the kurbenetes part in favor Istio Helm charts.

Did I misunderstand it?

jmazzitelli:
> @jmazzitelli I thought we would keep the openshift part of the files and deprecate the kurbenetes part in favor Istio Helm charts.
> 
> Did I misunderstand it?

Actually, I was going to say - we might not be able to do what we want. What happens when someone is developing new features, and they need to add, modify, or remove things from the yaml (like things from the cluster role). How can they do this? The Makefile will use the upstream helm, so the dev can't use that. We may need to rethink that.

jmazzitelli:
@gbaufake OK, I am going to merge this. At least some of it is going to be needed (the openshift) one.
Also, I'm going to merge the kiali.io documentation PR too: https://github.com/kiali/kiali.io/pull/56 - that is a file you will be changing too but some of those changes about the OpenShift install will still be valid (even if you have to modify the docs a bit).

PR #361, opened 2018-07-17
KIALI-1127 [1.0] Update Graph

jshaughn:
This gets the graph to generate using the 1.0 istio metrics.

** It does not yet handle the appenders **

The appenders compile but getting them working will be follow-up PRs.
Also, the Service graph endpoint is not functional (although it's not used)

PR #640, opened 2018-11-08
KIALI-1185 - Implement oAuth support for OCP/OKD/Maistra clients

cfcosta:
** Describe the change **

This PR implements the server-side support for oauth-proxy, enabled only on Openshift. It uses the oauth token (the same that we can use to connect to k8s) and handles most of the use cases.

** What still needs to be done **

* [x] Handle signout gracefully.

** Backwards incompatible? **

No.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -529,7 +529,7 @@ api:
 ----
 
 |`AUTH_STRATEGY`
-|The authentication strategy to be used by Kiali. The possible values are `none`, for no checking of credentials, `oauth` for using Openshift oauth, and `login` for the default login page (default).
+|The authentication strategy to be used by Kiali. The possible values are `none`, for no checking of credentials, `openshift` for using Openshift's oauth, and `login` for the default login page (default).
 [source,yaml]
 ----
 auth_strategy: VALUE
--- a/README.adoc
+++ b/README.adoc
@@ -528,6 +528,12 @@ api:
     - etc..
 ----
 
+|`AUTH_STRATEGY`
+|The authentication strategy to be used by Kiali. The possible values are `none`, for no checking of credentials, `oauth` for using Openshift oauth, and `login` for the default login page (default).
+[source,yaml]
+----
+auth_strategy: VALUE
+----
 |`LOGIN_TOKEN_SIGNING_KEY`
 |The signing key used to generate tokens for user authentication. (default is `kiali`)
 [source,yaml]
--- a/deploy/openshift/clusterrole.yaml
+++ b/deploy/openshift/clusterrole.yaml
@@ -81,6 +81,16 @@ rules:
   - get
   - list
   - watch
+- apiGroups: ["authentication.k8s.io"]
+  resources:
+  - tokenreviews
+  verbs:
+  - create
+- apiGroups: ["authorization.k8s.io"]
+  resources:
+  - subjectaccessreviews
+  verbs:
+  - create
 - apiGroups: ["project.openshift.io"]
   resources:
   - projects
--- a/deploy/openshift/deployment.yaml
+++ b/deploy/openshift/deployment.yaml
@@ -44,6 +44,7 @@ spec:
         - -cookie-secret=kiali-16bytes-1234
         - -skip-provider-button
         - -pass-access-token
+        ${AUTH_BYPASS}
         ports:
         - containerPort: 20002
           name: proxy
--- a/deploy/openshift/deployment.yaml
+++ b/deploy/openshift/deployment.yaml
@@ -33,7 +33,7 @@ spec:
         - -http-address=
         - -email-domain=*
         - -upstream=http://localhost:20001
-        - -openshift-service-account=kiali
+        - -openshift-service-account=kiali-service-account
         - '-openshift-sar={"namespace":"istio-system","resource":"services","name":"kiali","verb":"get"}'
         - '-openshift-delegate-urls={"/": {"resource": "namespaces", "verb": "get"}}'
         - -tls-cert=/etc/tls/private/tls.crt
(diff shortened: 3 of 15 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#361](https://paste.fedoraproject.org/paste/JsoZqwSzPE3LorWS0CiGJw)

deploy/openshift/clusterrole.yaml
```diff
@@ -79,6 +79,16 @@ rules:
...
+- apiGroups: ["authentication.k8s.io"]
+  resources:
+  - tokenreviews
+  verbs:
+  - create
+- apiGroups: ["authorization.k8s.io"]
+  resources:
+  - subjectaccessreviews
+  verbs:
+  - create
```
mwringe: I don't understand why we need this. Once we get the token we shouldn't have to do anything extra with it unless I am missing something.

cfcosta:
This is necessary for the proxy, it doesn’t work without those permissions.

Cheers,
Cainã Costa
http://www.cfcosta.com

deploy/openshift/route.yaml
```diff
@@ -10,5 +10,5 @@ spec:
     termination: reencrypt
   to:
     kind: Service
-    targetPort: 20001
+    targetPort: 20002
```
mwringe: Why are we changing the port?

cfcosta:
Because the proxy listens on the new port, and proxies to the old one.

Cheers,
Cainã Costa
http://www.cfcosta.com

mwringe:
I am a bit confused how this is working. It looks like we are just checking if we get a token back from the oauth proxy and then continuing. We shouldn't be allowing just anyone to access Kiali just because they have an account on OpenShift.

deploy/openshift/deployment.yaml
```diff
@@ -25,6 +25,33 @@ spec:
...
+      - image: openshift/oauth-proxy:v1.1.0
+        name: kiali-proxy
+        args:
+        - -provider=openshift
+        - -https-address=:20002
+        - -http-address=
+        - -email-domain=*
+        - -upstream=http://localhost:20001
+        - -openshift-service-account=kiali
+        - '-openshift-sar={"resource": "namespaces", "verb": "get"}'
```
mwringe: I don't think we want this. We don't want the proxy to be doing any sort of permission checks. We just need to the token and then Kiali uses that to do all the calls to the k8s master. That will control who has access to what.

This isn't going to work for our usecase since this means the user who accesses Kiali has to have permission to get all the namespaces (which isn't a normal user)

cfcosta: I think we can fix this for 1186, and for this one just have one permission that might be enough for our user. I'd say it's the responsibility of the proxy to do that, else we need to check the permission directly, on every request, where they only check it on the moment of the auth.

mwringe: >I'd say it's the responsibility of the proxy to do that, else we need to check the permission directly, on every request, where they only check it on the moment of the auth.

With KIALI-1186 we don't want to check the permission, we just use the token provided directly when communicating with the k8s master. The k8s master verifies the permissions before returning the results.

I don't know why we want this proxy when we get to KIALI-1186 since we just want to do a normal OAuth workflow where we request the token from OpenShift and use that instead of the Kiali service account token.

cfcosta: Yeah, that's the confusing part: the service account is necessary to create the correct oauth configuration. It does not use the service account token, but the user one. Having oauth directly, without using the service account is unstable, in the sense that it doesn't always return to the correct page, and for some reason, the returned token is not an access token that we can use, but a HMAC signed key that we can use for auth, but not for connecting.

Also, this proxy is the same that Prometheus uses on Openshift (https://github.com/openshift/cluster-monitoring-operator/blob/9c97908591f4b4fc3b12251c4b6ab0d6289fdbc0/assets/prometheus-k8s/prometheus.yaml) and the one that Jaeger is evaluating, afaik.

cfcosta:
> I am a bit confused how this is working. It looks like we are just checking if we get a token back from the oauth proxy and then continuing. We shouldn't be allowing just anyone to access Kiali just because they have an account on OpenShift.

Agree, that's what 1186 is for. And since that's a feature flag, we can just disable it if necessary.

deploy/openshift/deployment.yaml
mwringe: Yes, its the same proxy that others are using or considering, but its a different usecase than ours.

README.adoc
```diff
@@ -493,6 +493,12 @@ api:
     - etc..
 ----
 
+|`AUTH_STRATEGY`
+|The authentication strategy to be used by Kiali. The possible values are `none`, for no checking of credentials, `oauth` for using Openshift oauth, and `login` for the default login page (default).
```
mwringe: We might want to rename this to something like 'openshift' or 'openshift-oauth'. We might want to support other oauth types in the future.

cfcosta:
@mwringe updated with the default strategy and with the strategy name change.

Also, did some tests on two machines: my small server (running latest CentOS) and my machine (Fedora), and got both to work like this:

```bash
AUTH_STRATEGY=openshift CONSOLE_VERSION=local CONSOLE_LOCAL_DIR=<console-dir> make build docker-build openshift-deploy
```

It also worked fine on minishift.

mwringe:
Ok, I have this working now. I don't know why it was failing for me before.

A couple of things I have seen:
1) the logout button doesn't seem to work. When I click logout I am taken to the kiali login page, but if I just refresh the page then I am back into kiali and authenticated. I would expect to be logged out and taken to the OpenShift OAuth page instead

2) if I log in as a user that doesn't have permission I am taken to a page with a '403 Permission Denied'  error. Which is expected, but this page has a link to to the login page, clicking that takes me to a page asking if I want to login with OpenShift, clicking that takes me to the actual OpenShift login page. But now if I do use the right credentials then I never get redirected back to the Kiali console. I am logged into Kiali at this point, since if I go to the Kiali URL directly it works

cfcosta:
Closed in favour of #768 and https://github.com/kiali/kiali-ui/pull/906.

PR #647, opened 2018-11-09
KIALI-1873 Provide inbound and outbound error ratios in RequestHealth

israel-hdez:
The overall error rate is detailed into inbound and outbound error rates. This will bring a better idea about where to look for issues.

https://issues.jboss.org/browse/KIALI-1873

** Backwards compatible? ** *yes*

Related UI PR: https://github.com/kiali/kiali-ui/pull/810

### DIFF
--- a/models/health.go
+++ b/models/health.go
@@ -27,12 +27,16 @@ type AppHealth struct {
 	Requests         RequestHealth        `json:"requests"`
 }
 
+func NewEmptyRequestHealth() RequestHealth {
+	return RequestHealth{ErrorRatio: -1, InboundErrorRatio: -1, OutboundErrorRatio: -1}
+}
+
 // EmptyAppHealth create an empty AppHealth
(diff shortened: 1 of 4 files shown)


models/health.go
```diff
@@ -57,21 +61,48 @@ type WorkloadStatus struct {
...
 type RequestHealth struct {
-	requestRate float64
-	errorRate   float64
-	ErrorRatio  float64 `json:"errorRatio"`
+	inboundErrorRate    float64
+	outboundErrorRate   float64
+	inboundRequestRate  float64
+	outboundRequestRate float64
+
+	ErrorRatio         float64 `json:"errorRatio"`
```
jotak: I guess you kept the global error ratio because it is also displayed separately in different lists views?
I don't know if it still makes sense to have error rate displayed both here in in health. Maybe it could be removed in the future. But not for this PR.

PR #648, opened 2018-11-09
Update RELEASING.adoc

theute:
Adding instructions for JIRA


abonas:
LGTM. Should this be duplicated to similar doc in kiali-ui?

pilhuhn:
@abonas I think the main README in the ui already talks about UI releases. The jira process is only triggered when we do a kiali-all release (that includes the UI), which is triggered from this repo, so I'd say we are good.

PR #655, opened 2018-11-12
[KIALI-1863] E2E test to validate if clusterrole matches the one on I…

gbaufake:
## Describe the change 

This PR will include some e2e testing on clusterrole validation (bewteen kiali and istio helm charts).

## Issue reference 

 _`KIALI-1863 E2E Test to verify sync Kiali Yaml Files Through Maistra and Istio Helm Charts`_


mattmahoneyrh:
0.10 should perhaps be in the env.yaml and not embedded in the test code.
Consider moving clusterrole file get methods into their own class.

gbaufake:
0.10 is used as string to make YAML works.
I don't use the metadata part of YAML files for comparision

PR #351, opened 2018-07-11
Add changing package.json to release instructions for the UI

cfcosta:
From v0.5.0 onwards we started doing the correct tagging for kiali-ui, but it still needs to change the `package.json` so we can get the correct version from npm when building.

The `Makefile` already gets the latest version of the console, so I don't think we need to change much on the build scripts.

PR #663, opened 2018-11-13
go back to using major/minor - this should be OK

jmazzitelli:


deploy/openshift/deploy-kiali-to-openshift.sh
```diff
@@ -38,7 +38,7 @@ if [ "${KIALI_PASSPHRASE_BASE64}" == "" ]; then
 fi
 
 export IMAGE_NAME="${IMAGE_NAME:-kiali/kiali}"
-export IMAGE_VERSION="${IMAGE_VERSION:-v0.10.0}"
+export IMAGE_VERSION="${IMAGE_VERSION:-v0.10}"
 export VERSION_LABEL="${VERSION_LABEL:-$IMAGE_VERSION}"
```
pilhuhn: This only works if there is a branch with image_version (iiuc)

jmazzitelli: No, this only affects that the k8s Deployment pulls down. It doesn't mean anything wrt github branches.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#351](https://paste.fedoraproject.org/paste/euQrbt3ExP5kiLQ3ILSHcA)
  * :red_circle: run-kiali-e2e-tests [#[730]](https://paste.fedoraproject.org/paste/P68~NASgBtGs9-lDmmx6yw)

deploy/openshift/deploy-kiali-to-openshift.sh
pilhuhn: VERSION_LABEL is used in line 105 as part of the url:

    yaml_url="https://raw.githubusercontent.com/kiali/kiali/${VERSION_LABEL}/deploy/openshift/${yaml_file}"

so if V_L = IMAGE_VERSION  and I_V = v0.10  then the url has v0.10 in it, which only works when there is such a branch

jmazzitelli: Note that we have a v0.10 branch, too. But that branch does not have the split out yaml so that won't work.

pilhuhn: @jmazzitelli I got the V_L=master - but not everyone seeing that script has you available as personal tutor :-)

jmazzitelli: Yes, understood. We are not advertising this script yet. When we do the next release, all of this will work properly and we will document it on the gettingstarted page. (in fact, I'm thinking of moving this to kiali.io (we should put this on kiali.io/getLatestKiali so we have more control).

Note that regardless of this PR, a user would still need to say V_L=master today.

gbaufake: @jmazzitelli 
Can we reconcile this IMAGE_VERSION and VERSION_LABEL to just one thing?

jmazzitelli: no - they should be separate because they might be different under special use-cases (like in the case now - we need VERSION_LABEL to be master). VERSION_LABEL is used as part of github URLs (its typically the name of our branches/tags) but IMAGE_VERSION is not. This let's us pull the yaml files from github but let's the user choose a different image (for example, I may want to pull "latest" - I can't use VERSION_LABEL=latest because that invalidates the URL).

But, by default, they do take the same values
(see the line `export VERSION_LABEL="${VERSION_LABEL:-$IMAGE_VERSION}"`

jmazzitelli: Here's an example - run `make-openshift-deploy` or `make k8s-deploy` and notice the values of I_V and V_L:

```
IMAGE_NAME="kiali/kiali" \
IMAGE_VERSION="dev" \
IMAGE_PULL_POLICY_TOKEN="imagePullPolicy: IfNotPresent" \
VERSION_LABEL="v0.11.0-SNAPSHOT" \
NAMESPACE="istio-system" \
JAEGER_URL="http://jaeger-query-istio-system.127.0.0.1.nip.io" \
GRAFANA_URL="http://grafana-istio-system.127.0.0.1.nip.io"  \
VERBOSE_MODE="3" \
deploy/kubernetes/deploy-kiali-to-kubernetes.sh
```

jmazzitelli:
Engage.

PR #666, opened 2018-11-15
KIALI-1947 Split Istio Rules from Adapters/Templates

lucasponce:
** Describe the change **

Handle Adapters/Templates as independent object from Istio Rules as a previous step to edit them.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1947

** Backwards incompatible? **

Yes, changes in the API.

Also this PR needs an update in the UI project to update the changes.


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#364](https://paste.fedoraproject.org/paste/9dA~PNSUTSEmWDJd~~rEVw)
  * :red_circle: run-kiali-e2e-tests [#[771]](https://paste.fedoraproject.org/paste/qwK1sTgQfvT6H8oZHBDEHg)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#370](https://paste.fedoraproject.org/paste/DsiR5bH0pkdgBUB6RAlo0g)
  * :red_circle: run-kiali-e2e-tests [#[777]](https://paste.fedoraproject.org/paste/sT3xJ13gL-rg8hmwvTjRGQ)

business/istio_config.go
```diff
@@ -147,16 +172,15 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 	return istioConfigList, nil
 }
 
-func (in *IstioConfigService) GetIstioConfigDetails(namespace string, objectType string, object string) (models.IstioConfigDetails, error) {
+func (in *IstioConfigService) GetIstioConfigDetails(namespace, objectType, objectSubtype, object string) (models.IstioConfigDetails, error) {
```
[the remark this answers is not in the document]
lucasponce: +1

kubernetes/istio_rules_service.go
```diff
@@ -37,153 +39,97 @@ func (in *IstioClient) GetIstioRuleDetails(namespace string, istiorule string) (
...
+	for name, plural := range pluralsMap {
+		go func(name, plural string) {
+			results, err := in.istioConfigApi.Get().Namespace(namespace).Resource(plural).Do().Get()
+			istioObjects := istioResponse{}
+			resultList, ok := results.(*GenericIstioObjectList)
+			if !ok {
+				istioObjects.results = nil
+				istioObjects.err = fmt.Errorf("%s doesn't return a %s list", namespace, plural)
+			}
+			if err == nil {
```
jotak: should be `else if`, no? (I guess if `!ok`, then `resultList` is going to be nil)

kubernetes/istio_rules_service.go
```diff
@@ -37,153 +39,97 @@ func (in *IstioClient) GetIstioRuleDetails(namespace string, istiorule string) (
...
+			}
+			resultsChan <- istioObjects
+		}(name, plural)
+	}
+
+	results := make([]IstioObject, 0)
+	for i := 0; i < len(pluralsMap); i++ {
+		adapter := <-resultsChan
+		if adapter.err != nil {
+			return nil, adapter.err
```
jotak: Just a note here: I think it's ok to leave without closing the channel as [it will be garbage-collected](https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open).
However maybe we should want to continue collecting responses before returning immediately? (and store the error in a variable). That's particularly interesting in the case of read-permission denied on some objects: we would still continue to display objects where we have read permissions granted, rather than failing for everything. (That's a pattern that we don't apply in many parts of the code but I think it should be nice to start thinking like that)

kubernetes/istio_rules_service.go
lucasponce: Correct, I wanted to return on this "if" but I changed and introduced this bug.
Good catch.

kubernetes/istio_rules_service.go
lucasponce: Yes, in this KIALI-1947 I planned to work only on the split of rules/templates/adapters and consolidate it from the backend to the ui.
But you are right, in KIALI-1654 I wanted to link this logic with the permissions one, just to separate them.
But yes, it makes sense, is something to take into consideration.

lucasponce:
>  (I'm not sure if you want to take my last comment in consideration right now or for later, so I let you merge when you want)

Sure, I added a last commit with it, but this PR needs also some UI one, so DNM label is still in place until I can prepare the next PR.

Thanks !

business/istio_config_test.go
```diff
@@ -137,29 +137,29 @@ func TestGetIstioConfigDetails(t *testing.T) {
...
 	assert.Equal("checkfromcustomer", istioConfigDetails.Rule.Name)
 	assert.Nil(err)
 
-	istioConfigDetails, err = configService.GetIstioConfigDetails("test", "serviceentries", "googleapis")
+	istioConfigDetails, err = configService.GetIstioConfigDetails("test", "serviceentries", "", "googleapis")
 	assert.Equal("googleapis", istioConfigDetails.ServiceEntry.Name)
 	assert.Nil(err)
 
-	istioConfigDetails, err = configService.GetIstioConfigDetails("test", "rules-bad", "stdio")
+	istioConfigDetails, err = configService.GetIstioConfigDetails("test", "rules-bad", "", "stdio")
```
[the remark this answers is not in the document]
lucasponce: +1
Yes, I've added a couple more testing of this business method to support all scenarios.

business/istio_config.go
```diff
@@ -220,11 +261,16 @@ func GetIstioAPI(resourceType string) string {
 }
 
 // DeleteIstioConfigDetail deletes the given Istio resource
-func (in *IstioConfigService) DeleteIstioConfigDetail(api, namespace, resourceType, name string) (err error) {
+func (in *IstioConfigService) DeleteIstioConfigDetail(api, namespace, resourceType, resourceSubtype, name string) (err error) {
```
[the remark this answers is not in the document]
lucasponce: +1
It was missing; I've added in the next commit.

PR #672, opened 2018-11-21
KIALI-1319 Add multi-namespace gateway multi-match checker

burmanm:
Checks that there are no duplicate gateways in any of the namespaces that the application can see. Some notes on scaling though:

* Uses linear search to check for matching hosts
  * If performance becomes a real issue, then something more complex like a hostname graph structure or a trie is required. This feels like an overkill in my opinion however at this point.
* Since golang has no unbounded channels, the errChan could get blocked. I've increased the size to 64 and try to avoid writing there, but in theory there's a possibility that the errChan returns len = 0 and it is actually full. In theory.

### DIFF
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -35,7 +35,7 @@ func (in *IstioValidationsService) GetServiceValidations(namespace, service stri
 
 	// Get Gateways and ServiceEntries to validate VirtualServices
 	wg := sync.WaitGroup{}
-	errChan := make(chan error, 2)
+	errChan := make(chan error, 64)
 
 	istioDetails := kubernetes.IstioDetails{}
 	vs := make([]kubernetes.IstioObject, 0)
@@ -48,6 +48,7 @@ func (in *IstioValidationsService) GetServiceValidations(namespace, service stri
 	go fetch(&drs, namespace, service, in.k8s.GetDestinationRules, &wg, errChan)
 	go in.serviceFetcher(&services, namespace, errChan, &wg)
 	go in.fetchWorkloads(&workloads, namespace, errChan, &wg)
+	gwChecker := in.getGatewayChecker(errChan, &wg) // Block
 	wg.Wait()
 	if len(errChan) != 0 {
 		err = <-errChan
@@ -59,6 +60,7 @@ func (in *IstioValidationsService) GetServiceValidations(namespace, service stri
 		checkers.VirtualServiceChecker{namespace, drs, vs},
 		checkers.DestinationRulesChecker{DestinationRules: drs},
 		checkers.NoServiceChecker{Namespace: namespace, Services: services, IstioDetails: &istioDetails, WorkloadList: workloads},
+		gwChecker,
 	}
 
 	// Get group validations for same kind istio objects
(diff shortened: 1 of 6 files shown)


lucasponce:
> Since golang has no unbounded channels, the errChan could get blocked. I've increased the size to 64 and try to avoid writing there, but in theory there's a possibility that the errChan returns len = 0 and it is actually full. In theory.

Could you ellaborate this ? Not sure if I understand correctly in the code this change.

business/istio_validations.go
```diff
@@ -35,7 +35,7 @@ func (in *IstioValidationsService) GetServiceValidations(namespace, service stri
 
 	// Get Gateways and ServiceEntries to validate VirtualServices
 	wg := sync.WaitGroup{}
-	errChan := make(chan error, 2)
+	errChan := make(chan error, 64)
```
lucasponce: I wonder if we need to refactor this method.
To me it starts to get complex and hard to follow.
I understand that increasing this to 64 it's because we don't control that max # of invocations could be 2 and we need some kind of higher level.
But it so, perhaps it's a matter to reorganize how are we grouping these methods or to use just a semaphore with a shared variable, I don't know, just thinking loud.

hhovsepy:
I found a situation when one of gateway's warning is not shown, 
this happens when there are 3 gateways with different names in different namespaces:
1. Create bookinfo-gateway-auto for bookinfo2 namespace.
2. Create bookinfo-gateway for bookinfo namespace.

screenshot from 2018-11-21 17-05-06
screenshot from 2018-11-21 17-05-28
screenshot from 2018-11-21 17-05-44
screenshot from 2018-11-21 17-06-49
screenshot from 2018-11-21 17-07-04

burmanm:
Actually that one was on purpose, it allows one to be valid and rest are invalid (on the fetch order). But I can make it invalidate all of them (they just need to filtered out).

business/istio_validations.go
burmanm: I'll answer the previous question also. The question is that if we would have 65 namespaces and fetching data from each of them would fail at the same time, then it's possible that the last fetch is blocked indefinitely as the channel is never read before it is being written to.

In theory that could happen, but it would probably require large amount of bad luck with a scheduler + very beefy machine with large amount of cores.

Now, we could use something like the "context" that would cancel after first error and then prevent writing to it. Theoretically we could also do changes to the err-variable and let just any of them win (who cares which), but test -race does not like that although it's quite valid in this case. Maybe Golang has Java style atomic swap for variable also, I can't remember right now (at least unsafe has, but I don't remember if otherwise there is).

Also, sadly the goroutine can't return anything to the caller so err can't be handled that way (and channel can't be expanded on the fly). 

But all these methods are extremely complicated already and duplicating code, so a refactor is clearly necessary. Maybe it should be done in the context of KIALI-1671.

lucasponce: > But all these methods are extremely complicated already and duplicating code, so a refactor is clearly necessary. Maybe it should be done in the context of KIALI-1671.

+1

Yes this logic has grown very fast and we need to revisit it.
KIALI-1671 sounds the right place to address it.

hhovsepy:
Verified the solution.
I think it is better to show warn for all Gateways to not case confusion.
Because in older commit,  after deletion and recreateion the first gateway, anyway all were shown as warn.

xeviknal:
@burmanm I just recall that we should update kiali.io adding this new validation up. In documentation/overview/istio config (section 2.5.1) we have the list of all of them. Would you mind adding that up?

PR #674, opened 2018-11-22
KIALI-1984 Refactor Kiali model to include metadata/spec in Istio config

lucasponce:
** Describe the change **

Istio Objects mapped in Kiali that will be editable need to bring explicit metadata/spec fields.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1984

** Backwards incompatible? **

Yes, this requires changes in e2e test and also a PR in the UI to map the changes.

### DIFF
--- a/models/destination_rule.go
+++ b/models/destination_rule.go
@@ -1,6 +1,8 @@
 package models
 
 import (
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 )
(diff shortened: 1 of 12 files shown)


models/destination_rule.go
```diff
@@ -23,21 +25,12 @@ type DestinationRules struct {
...
-	CreatedAt string `json:"createdAt"`
-	// The resource version of the destinationRule
-	//
-	// required: true
-	ResourceVersion string      `json:"resourceVersion"`
-	Host            interface{} `json:"host"`
-	TrafficPolicy   interface{} `json:"trafficPolicy"`
-	Subsets         interface{} `json:"subsets"`
+	Metadata meta_v1.ObjectMeta `json:"metadata"`
+	Spec     struct {
```
jotak: Wow I didn't know we could do that :)

jotak:
One remark however, that maybe we could think about for a future task: as we're now showing the original yml for edition purpose, maybe we should remove our model layer for these objects and directly send the k8s model?

It would be safer by removing any risk of discrepancy between the two models, hence risk of saving something inappropriate.

Also would make kiali still able to show & edit yaml even if it's not recognized because using another version of Istio. In which case only the validations can be broken, without affecting the rest.

lucasponce:
> One remark however, that maybe we could think about for a future task: as we're now showing the original yml for edition purpose, maybe we should remove our model layer for these objects and directly send the k8s model?

Kiali one could enrich the Istio one, for example adding validations in a future step and also permissions.
But having this metadata/spec is handy to manage it for editing.

Probably if we see that there is not value we can consider to remove the kiali model, but I don't see it clear yet.

We need to progress more in the edit use cases to decide it.

PR #676, opened 2018-11-23
KIALI-1140 Add destination services for workload details endpoint

xeviknal:
** Describe the change **

Adding the services that are called by an specific workload.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1140

This goes along a UI feature that will list those services in the workload details: https://github.com/kiali/kiali-ui/pull/839

destination-services

### DIFF
--- a/business/layer.go
+++ b/business/layer.go
@@ -51,8 +51,8 @@ func NewWithBackends(k8s kubernetes.IstioClientInterface, prom prometheus.Client
 	temporaryLayer.Health = HealthService{prom: prom, k8s: k8s}
 	temporaryLayer.Svc = SvcService{prom: prom, k8s: k8s, businessLayer: temporaryLayer}
 	temporaryLayer.IstioConfig = IstioConfigService{k8s: k8s}
-	temporaryLayer.Workload = WorkloadService{k8s: k8s}
 	temporaryLayer.Validations = IstioValidationsService{k8s: k8s, ws: temporaryLayer.Workload}
+	temporaryLayer.Workload = WorkloadService{k8s: k8s, prom: prom, businessLayer: temporaryLayer}
 	temporaryLayer.App = AppService{k8s: k8s}
 	temporaryLayer.Namespace = NewNamespaceService(k8s)
 	temporaryLayer.k8s = k8s
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -16,12 +16,15 @@ import (
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
+	"github.com/kiali/kiali/prometheus"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
 )
 
 // Workload deals with fetching istio/kubernetes workloads related content and convert to kiali model
 type WorkloadService struct {
-	k8s kubernetes.IstioClientInterface
+	prom          prometheus.ClientInterface
+	k8s           kubernetes.IstioClientInterface
+	businessLayer *Layer
 }
 
 // GetWorkloadList is the API handler to fetch the list of workloads in a given namespace.
@@ -61,11 +64,41 @@ func (in *WorkloadService) GetWorkload(namespace string, workloadName string, in
 	}
 
 	if includeServices {
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		errChan := make(chan error, 1)
+		var destService []prometheus.Service
+
+		go func() {
+			defer wg.Done()
+			var err2 error
+			ns, err2 := in.businessLayer.Namespace.GetNamespace(namespace)
+			if err2 != nil {
+				log.Errorf("Error fetching details of namespace %s: %s", namespace, err2)
+				errChan <- err2
+				return
+			}
+
+			destService, err2 = in.prom.GetDestinationServices(ns.Name, ns.CreationTimestamp, workloadName)
+			if err2 != nil {
+				log.Errorf("Error fetching SourceWorkloads per namespace %s and service %s: %s", namespace, workloadName, err2)
+				errChan <- err2
+			}
+		}()
+
 		services, err := in.k8s.GetServices(namespace, workload.Labels)
 		if err != nil {
 			return nil, err
 		}
 		workload.SetServices(services)
+
+		wg.Wait()
+		if len(errChan) != 0 {
+			err = <-errChan
+			return nil, err
+		}
+
+		workload.SetDestinationServices(destService)
 	}
 
 	return workload, nil
--- a/prometheus/client.go
+++ b/prometheus/client.go
@@ -43,6 +43,14 @@ type Workload struct {
 	Version   string
 }
 
+// Service describes a service with contextual information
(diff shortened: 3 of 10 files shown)


business/layer.go
```diff
@@ -51,7 +51,7 @@ func NewWithBackends(k8s kubernetes.IstioClientInterface, prom prometheus.Client
 	temporaryLayer.Health = HealthService{prom: prom, k8s: k8s}
 	temporaryLayer.Svc = SvcService{prom: prom, k8s: k8s, businessLayer: temporaryLayer}
 	temporaryLayer.IstioConfig = IstioConfigService{k8s: k8s}
-	temporaryLayer.Workload = WorkloadService{k8s: k8s}
+	temporaryLayer.Workload = WorkloadService{k8s: k8s, prom: prom, ns: NewNamespaceService(k8s)}
```
lucasponce: I think you might want to use here the temporaryLayer instead to create a new one.
In similar code as 
```
	temporaryLayer.Svc = SvcService{prom: prom, k8s: k8s, businessLayer: temporaryLayer}
```
businessLayer is passed inside SvcService to access Health service, for WorkloadService you need this to access NamespaceService, I think we can use the same pattern instead to create a new ns variable.

business/workloads.go
```diff
@@ -16,12 +16,15 @@ import (
...
+	"github.com/kiali/kiali/prometheus"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
 )
 
 // Workload deals with fetching istio/kubernetes workloads related content and convert to kiali model
 type WorkloadService struct {
-	k8s kubernetes.IstioClientInterface
+	prom prometheus.ClientInterface
+	k8s  kubernetes.IstioClientInterface
+	ns   NamespaceService
```
lucasponce: Similar comment here as layer.go

business/workloads.go
```diff
@@ -61,11 +64,40 @@ func (in *WorkloadService) GetWorkload(namespace string, workloadName string, in
...
 	if includeServices {
+		wg := sync.WaitGroup{}
+		wg.Add(1)
+		errChan := make(chan error, 1)
+		var destService []prometheus.Service
+
+		go func() {
+			defer wg.Done()
+			var err2 error
+			ns, err2 := in.ns.GetNamespace(namespace)
```
lucasponce: Same comment here.

business/workloads.go
```diff
@@ -61,11 +64,40 @@ func (in *WorkloadService) GetWorkload(namespace string, workloadName string, in
...
+			if err2 != nil {
+				log.Errorf("Error fetching details of namespace %s: %s", namespace, err2)
+				errChan <- err2
```
jotak: I guess you should `return` here

prometheus/client.go
```diff
@@ -117,6 +126,44 @@ func (in *Client) GetSourceWorkloads(namespace string, namespaceCreationTime tim
...
+		reporter = "destination"
+	}
+
+	queryTime := util.Clock.Now()
+	queryInterval := queryTime.Sub(namespaceCreationTime)
+	groupBy := "(destination_service_namespace, destination_service_name, destination_service)"
+	query := fmt.Sprintf("sum(rate(istio_requests_total{reporter=\"%s\",source_workload=\"%s\",source_workload_namespace=\"%s\"}[%vs])) by %s",
+		reporter, workloadname, namespace, int(queryInterval.Seconds()), groupBy)
+	log.Debugf("GetDestinationServices query: %s", query)
+	log.Infof("GetDestinationServices query: %s", query)
```
jotak: Remove one of the above logging ?

prometheus/client.go
```diff
@@ -43,6 +44,14 @@ type Workload struct {
 	Version   string
 }
 
+// Service describes a service with contextual information
+type Service struct {
+	Namespace   string
+	App         string
+	ServiceName string
+	Service     string
```
jotak: What "Service" stands for? is it like the full qualified name "svc.namespace.cluster"? This is weird to see both "Service" and "ServiceName" here

business/layer.go
xeviknal: We also have the line right below following this other pattern. Besides that, adding the whole layer will add recursion into the structure. For example, if you are using `SvcService` you would be able to call all the `SvcService` public methods from `businessLayer`. IMO, we can avoid this behavior.
Moreover, It sounds to me a little bit overkiller to add the whole other services if you just need one more.

Following what you suggest about avoiding create a new instance, probably I should use the instance created in L57.
```go
temporaryLayer.Workload = WorkloadService{k8s: k8s, prom: prom, ns: temporaryLayer.Namespace}
```
what do you think about that?

lucasponce: I would say to follow the same pattern as in SvcService.
We have a mixed of styles in these packages and we should normalize.
It will be refactored in the future all at same time for consistency, but let's not create more combinations of possibilities, I think if there is an example in the same file for what you want to do, the best choice is to replicate it.

xeviknal:
@lucasponce comments approached.

xeviknal:
@hhovsepy I did changes to avoid the behavior you realized (https://github.com/kiali/kiali-ui/pull/839#issuecomment-442161393)

jotak:
@xeviknal I think reporter=source was good, actually. This was not a bug ; or maybe, a bug in the graph, as per the JIRA ticket opened to show unknown services in graph.
It is correct to report unknown destination when a workloads sends requests to an unknown / external service.

xeviknal:
@jotak you are right! then the scenario proposed by hayk is right? Should we leave it like that? 
It sounds weird to me that we have an unknown service because of an error on VS.

jotak:
@xeviknal I think this scenario is correct, the workload tries to connect to a service that doesn't exist so it's reported as we see in telemetry. It's the consequence of VS badly configured **and** being actually used in a request.

xeviknal:
@jotak good! And as discussed with team, we are going as it is and as you suggested. So this PR is ready to merge. Thanks everyone :)

business/layer.go
```diff
@@ -51,8 +51,8 @@ func NewWithBackends(k8s kubernetes.IstioClientInterface, prom prometheus.Client
 	temporaryLayer.Health = HealthService{prom: prom, k8s: k8s}
 	temporaryLayer.Svc = SvcService{prom: prom, k8s: k8s, businessLayer: temporaryLayer}
 	temporaryLayer.IstioConfig = IstioConfigService{k8s: k8s}
-	temporaryLayer.Workload = WorkloadService{k8s: k8s}
 	temporaryLayer.Validations = IstioValidationsService{k8s: k8s, ws: temporaryLayer.Workload}
+	temporaryLayer.Workload = WorkloadService{k8s: k8s, prom: prom, businessLayer: temporaryLayer}
```
xeviknal: Here comes the main problem of this PR. `ws` field from `IstioValidationsService` is not a pointer, so it is ws got a copy of `temporaryLayer` that was an empty object. So anytime we wanted to access to any page that consumed a IstioValidations an error raised.
This is fixed in: https://github.com/kiali/kiali/pull/684/files#diff-59657b2347e4d9f79961f3c52ce83db9R54

PR #678, opened 2018-11-26
update cluster-openshift hack script to use latest maistra

jmazzitelli:
cluster-openshift.sh will now install latest maistra (includes OS 3.11, Istio 1.0.4).

### DIFF
--- a/hack/cluster-openshift.sh
+++ b/hack/cluster-openshift.sh
@@ -28,10 +28,10 @@ debug() {
 cd "$(dirname "${BASH_SOURCE[0]}")"
 
 # The default version of the istiooc command to be downloaded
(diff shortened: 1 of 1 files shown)


jmazzitelli:
@abonas this update installs the latest maistra. I know you use this hack script, so I made you a reviewer.

You can install things like this:

* `hack/cluster-openshift.sh up`
* At the prompt, enter "1" to tell the script you want cluster admin rights for your new user
* `hack/istio/download-istio.sh`
** this `download-istio.sh` script will by default pull down the latest istio (1.0.4) into your _output directory. In the resulting `_output/istio-1.0.4/bin` directory will be `istioctl` - you need that istioctl executable in your PATH to do the next step - installing bookinfo
* `hack/istio/install-bookinfo-demo.sh --mongo -tg`
** the options `--mongo -tg` will install a mongo DB backend and a traffic generator so you can get a populated kiali graph without having to do anything.

PR #680, opened 2018-11-27
hack config to create external service in bookinfo

jmazzitelli:
this hack config helps us dev/test with external services.
see kiali-2014 for example why we need this

### DIFF
--- a/hack/istio/bookinfo-ext-service-entry.yaml
+++ b/hack/istio/bookinfo-ext-service-entry.yaml
@@ -1,0 +1,76 @@
+##########
+# This creates a simulated external service for the ratings bookinfo service such that
+# when the reviews service requests ratings from the ratings service, that request will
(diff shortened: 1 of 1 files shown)


hack/istio/bookinfo-ext-service-entry.yaml
```diff
@@ -0,0 +1,76 @@
...
+  hosts:
+  - foo.bookinfo.ext # this is an arbitrary string; only used in VS
+  location: MESH_EXTERNAL
+  ports:
+  - number: 9080
+    name: http-ext
+    protocol: HTTP
+  resolution: STATIC
+  endpoints:
+  - address: 192.168.1.19
```
pilhuhn: You could put a comment at the end of line `# edit here`

jmazzitelli: done

PR #681, opened 2018-11-27
KIALI-1654 Add support to modify Istio config objects

lucasponce:
Expose a PATCH endpoint to support modification on Istio objects [1][2] with more flexibility.

The new enpoint can be tested as described in the following examples.

Update a DestinationRule
```
PATCH http://admin:admin@localhost:8000/api/namespaces/bookinfo/istio/destinationrules/reviews
Content-Type: application/json

{
  "spec": {
    "host": "reviews",
    "trafficPolicy": null,
    "subsets": [
      {
        "labels": {
          "version": "v1"
        },
        "name": "v1"
      },
      {
        "labels": {
          "version": "v2"
        },
        "name": "v2"
      },
      {
        "labels": {
          "version": "v3"
        },
        "name": "v3"
      },
      {
        "labels": {
          "version": "v4"
        },
        "name": "v4"
      }
    ]
  }
}
```

Update an Adapter/Template:
```
PATCH http://admin:admin@localhost:8000/api/namespaces/istio-system/istio/templates/quotas/requestcount
Content-Type: application/json

{
  "spec": {
    "dimensions": {
      "destination": "destination.labels[\"app\"] | destination.service | \"unknown\"",
      "destinationVersion": "destination.labels[\"version\"] | \"unknown\"",
      "source": "source.labels[\"app\"] | source.service | \"unknown\"",
      "sourceVersion": "source.labels[\"version\"] | \"unknown\""
    }
  }
}
```

This PR won't break the UI as it is adding new feature without modifying existing one.

[1] https://kubernetes.io/docs/tasks/run-application/update-api-object-kubectl-patch/
[2] https://tools.ietf.org/html/rfc7386

### DIFF
--- a/kubernetes/istio_details_service.go
+++ b/kubernetes/istio_details_service.go
@@ -2,6 +2,8 @@ package kubernetes
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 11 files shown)


kubernetes/istio_details_service.go
```diff
@@ -266,6 +268,27 @@ func (in *IstioClient) GetQuotaSpecBinding(namespace string, quotaSpecBindingNam
 	return quotaSpecBinding.DeepCopyIstioObject(), nil
 }
 
+// UpdateIstioObject updates an Istio object from either config api or networking api
+func (in *IstioClient) UpdateIstioObject(api, namespace, resourceType, name, jsonPatch string) (IstioObject, error) {
+	log.Infof("UpdateIstioObject input: %s / %s / %s / %s", api, namespace, resourceType, name)
```
jotak: If it's for auditing (I believe it is), shouldn't we also log `jsonPatch` ?

lucasponce: This is just for informational.
Auditing logging will have a specific format and I planned to address it on KIALI-1771.
(As it could have more config logic to decide the verbosity, etc, and I didn't plan to do it on this one).

But yeah, comment is valid.

lucasponce:
> In the current code while getting objects, we're checking for "update" permissions (in func "getUpdateDeletePermissions") ; however since you're using "patch", we should change it accordingly.

True, I forgot to update the verb, perhaps it's a corner case, but yes, as it is using PATCH verb that should be used to fetch permissions.

Thanks !

lucasponce:
@jotak I was rebasing and revisiting your comment and not sure if we need to change "update" by "patch".

I mean, reading this:
https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#resourceattributes-v1-authorization-k8s-io

```
Verb is a kubernetes resource API verb, like: get, list, watch, create, update, delete, proxy. "*" means all.
--
```

So, there is no distinction on "POST/PUT/PATCH" so, it seems create/update maps lowlevel verbs.

lucasponce:
@jotak updated verbs
@jmazzitelli @gbaufake this PR will update the clusterrole.yaml do you remember what more steps I need to update ?
I remember @gbaufake created some e2e for this, can you point me to them ?
I can also update them on this PR, too.

jotak:
@lucasponce as discussed it seems that "patch" really exists despite what the doc says. I can re-test without Maistra when you want (as Maistra won't let you see the permission issues).

lucasponce:
@jotak added "patch". I didn't split "patch" / "update" I guess in the context of kiali we are using both in the same semantic here.

PR #685, opened 2018-11-28
KIALI-2014 make sure dead nodes appender doesn't strip egress nodes

jmazzitelli:
** Describe the change **

ServiceEntry objects with MESH_EXTERNAL definitions were getting stripped out by the dead node appender. This fixes it.

** Issue reference **

KIALI-2014 Support ServiceEntry in Graph

** Backwards incompatible? **

No

Here's what the new graph will look like - notice the egress service node is now showing:

image

### DIFF
--- a/graph/appender/dead_node.go
+++ b/graph/appender/dead_node.go
@@ -1,8 +1,11 @@
 package appender
 
 import (
+	"time"
+
 	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/graph"
+	"github.com/kiali/kiali/log"
 )
 
 const DeadNodeAppenderName = "deadNode"
@@ -15,7 +18,9 @@ const DeadNodeAppenderName = "deadNode"
 //   terminal service node and we want to show it even if it has no incoming traffic
 //   for the time period.
 // Name: deadNode
-type DeadNodeAppender struct{}
+type DeadNodeAppender struct {
+	AccessibleNamespaces map[string]time.Time
+}
 
 // Name implements Appender
 func (a DeadNodeAppender) Name() string {
@@ -39,10 +44,10 @@ func (a DeadNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *G
 		namespaceInfo.WorkloadList = &workloadList
 	}
 
-	applyDeadNodes(trafficMap, globalInfo, namespaceInfo)
+	a.applyDeadNodes(trafficMap, globalInfo, namespaceInfo)
 }
 
-func applyDeadNodes(trafficMap graph.TrafficMap, globalInfo *GlobalInfo, namespaceInfo *NamespaceInfo) {
+func (a DeadNodeAppender) applyDeadNodes(trafficMap graph.TrafficMap, globalInfo *GlobalInfo, namespaceInfo *NamespaceInfo) {
 	numRemoved := 0
 	for id, n := range trafficMap {
 		switch n.NodeType {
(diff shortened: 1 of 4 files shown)


jmazzitelli:
Once this is merged, we'll need to then fix this: https://issues.jboss.org/browse/KIALI-2026

pilhuhn:
That looks (almost) good for my use case.

bildschirmfoto 2018-11-28 um 17 07 51

It is a bit confusing that the arrow is not marked with the VS icon (and/or does not originate from the serviceb-service)

graph/appender/dead_node.go
```diff
@@ -123,28 +122,33 @@ func applyDeadNodes(trafficMap graph.TrafficMap, globalInfo *GlobalInfo, namespa
...
-			IncludeServiceEntries: true,
-			Namespace:             namespaceInfo.Namespace,
-		})
-		checkError(err)
-
-		for _, entry := range istioCfg.ServiceEntries {
-			if entry.Spec.Hosts != nil && entry.Spec.Location == "MESH_EXTERNAL" {
-				for _, host := range entry.Spec.Hosts.([]interface{}) {
-					namespaceInfo.ExternalServices[host.(string)] = true
+		for ns := range a.AccessibleNamespaces {
```
israel-hdez: I'm wondering how to make this work in the _multi-tenancy_ or _reduced privileges_ use cases.

Since external services * just work *, we should look for them in all namespaces (not only accessible ones). But Kiali is running with reduced permissions... So... well... I don't know. This is just a comment to have this point in mind.

jmazzitelli: We can only look at what is accessible to the user. So he will only see accessible namespaces, but if restricted access is given to the user, then this will only show what the user can see, which is basically "working as expected" in that case.

graph/appender/dead_node.go
```diff
@@ -123,28 +122,33 @@ func applyDeadNodes(trafficMap graph.TrafficMap, globalInfo *GlobalInfo, namespa
...
-// All ServiceEntries are needed, because Istio does not distinguish where the
-// ServiceEntries are created when routing egress traffic.
-func isExternalService(service string, namespaceInfo *NamespaceInfo, globalInfo *GlobalInfo) bool {
+// All ServiceEntries are needed because Istio does not distinguish where the
+// ServiceEntries are created when routing egress traffic (i.e. a ServiceEntry
+// can be in any namespace and it will still work).
+// However, an egress service will have its namespace set to "default" in the telemetry.
+func (a DeadNodeAppender) isExternalService(service string, namespaceInfo *NamespaceInfo, globalInfo *GlobalInfo) bool {
 	if namespaceInfo.ExternalServices == nil {
 		namespaceInfo.ExternalServices = make(map[string]bool)
```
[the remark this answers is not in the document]
jmazzitelli: done

graph/appender/dead_node.go
```diff
@@ -123,28 +122,33 @@ func applyDeadNodes(trafficMap graph.TrafficMap, globalInfo *GlobalInfo, namespa
...
 
-		// Currently no other appenders use ServiceEntries, so they are not cached in NamespaceInfo
```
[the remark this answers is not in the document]
jmazzitelli: done.

graph/appender/dead_node.go
israel-hdez: I know. That's OK from that point of view. However, workloads will still be able to use the *external services*.

I mean... External services needs to be defined somewhere. But they really don't belong to a namespace... they are... well... external to the mesh. If we hide them just because the definition is in namespace that the user can't access doesn't make much sense to me.

jshaughn: I agree with @israel-hdez , it doesn't make a lot of sense that we can show the node but not be able to flag it as an egress node.  I feel like we're going to have to revisit this going forward.  As it stands now we can only flag egress nodes defined in a namespace kiali can access.

PR #686, opened 2018-11-29
KIALI-1973 Create CRD and register related k8s API

jotak:
- Add get role for these resources
- Update deploy script to install CRD
- Temporary sample vertx dashboard
- Some small refactoring in k8s (new function "newClientForAPI" shared by Istio client, OS client and the new Kiali client)

JIRA: https://issues.jboss.org/browse/KIALI-1973


deploy/openshift/crds.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
lucasponce: deploy/kubernetes/crds.yaml is missing ? or we don't plan to add runtimes outside openshift ?

Makefile
```diff
@@ -281,7 +281,7 @@ deploy/openshift/deploy-kiali-to-openshift.sh
 ## openshift-undeploy: Undeploy from Openshift project.
 openshift-undeploy: .openshift-validate
 	@echo Undeploying from OpenShift project ${NAMESPACE}
-	${OC} delete all,secrets,sa,templates,configmaps,deployments,clusterroles,clusterrolebindings,virtualservices,destinationrules,ingresses --selector=app=kiali -n ${NAMESPACE}
+	${OC} delete all,secrets,sa,templates,configmaps,deployments,clusterroles,clusterrolebindings,virtualservices,destinationrules,ingresses,customresourcedefinitions --selector=app=kiali -n ${NAMESPACE}
```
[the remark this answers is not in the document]
jotak: it's different, here with `customresourcedefinitions` I'm asking to remove the *definitions* (the CRDs), not the custom objects themselves. So it's still necessary to explicitly remove virtualservices and destinationrules

deploy/openshift/crds.yaml
jotak: you're right I forgot it

kubernetes/os_route_client.go
```diff
@@ -19,7 +16,7 @@ type OSRouteClient struct {
 
 // NewOSRouteClient creates a new client able to fetch OpenShift Routes API.
 func NewOSRouteClient() (*OSRouteClient, error) {
-	config, err := rest.InClusterConfig()
+	config, err := ConfigClient()
```
lucasponce: I think that the whole os_route_client.go is not being used anymore.

Makefile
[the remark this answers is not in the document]
jotak: (however the first item we delete is "all", so I wonder why we need all the others)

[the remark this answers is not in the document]
pilhuhn: "all" never has been _all_, but only a small subset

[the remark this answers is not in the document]
lucasponce: Ok so "customresourcedefinitions" can delete all definitions from the selector, so, without selector it can delete all definitions from the given namespace.

kubernetes/os_route_client.go
jotak: I removed old references to openshift routes

deploy/openshift/crds.yaml
jotak: @lucasponce added

Makefile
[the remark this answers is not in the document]
jotak: @lucasponce I guess yes

jotak:
@jmazzitelli of course, no problem to discuss.
Note, if having CRD is a problem upstream, we can just not have it installed. This CRD should anyway not be a _requirement_ for Kiali to run, but it's enabling new features when present.

jmazzitelli:
If its a feature people are going to be able to use (even optionally), we should get istio to install what we need. People most likely aren't going to be using our deploy/ scripts - they are going to be installing Kiali with Istio using the built-in Istio helm charts. So we'd want it to be done there, too.

deploy/kubernetes/clusterrole.yaml
```diff
@@ -71,4 +71,9 @@ rules:
...
   - get
   - list
-  - watch
\ No newline at end of file
+  - watch
+- apiGroups: ["monitoring.kiali.io"]
+  resources:
+  - monitoringdashboards
+  verbs:
+  - get
```
jmazzitelli: We also need to remember to get this changed in upstream Istio and Maistra

deploy/dashboards/vertx.yaml
```diff
@@ -0,0 +1,25 @@
+apiVersion: "monitoring.kiali.io/v1alpha1"
+kind: MonitoringDashboard
+metadata:
+  name: vertx
```
jmazzitelli: Missing labels for app and version. This will not be removed with selector=app=kiali without it.

Makefile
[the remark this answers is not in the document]
jmazzitelli: will this delete all objects of the kind:

```
apiVersion: "monitoring.kiali.io/v1alpha1"
kind: MonitoringDashboard
```
??

(assuming they all have an app label - which our vert.x one does not - see my previous comment about that)

deploy/kubernetes/crds.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: monitoringdashboards.monitoring.kiali.io
+  annotations:
+    "helm.sh/hook": crd-install
+  labels:
+    app: kiali
```
jmazzitelli: missing version label. We should be consistent across all our objects - use app and version labels

deploy/openshift/clusterrole.yaml
```diff
@@ -84,8 +84,8 @@ rules:
   - projects
   verbs:
   - get
-- apiGroups: ["route.openshift.io"]
+- apiGroups: ["monitoring.kiali.io"]
   resources:
-  - routes
+  - monitoringdashboards
```
jmazzitelli: maistra needs to know about this

deploy/openshift/crds.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: monitoringdashboards.monitoring.kiali.io
+  labels:
+    app: kiali
```
jmazzitelli: missing version label

deploy/dashboards/vertx.yaml
jotak: At the moment it's not meant to be tied to Kiali - not deployed with Kiali and not removed with it.
It's currently here as an example, at least until the whole epic is finished.

Makefile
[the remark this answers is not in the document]
jotak: no (see response on previous comment)

jotak:
I'm removing the DNM label. As explained on the mailing list, having it merged doesn't mean that it really goes "live" as it's an experimental feautre and still necessitate more plumbing to be fully functional (more plumbing = creating and installing dashboard templates)

The CRD/ConfigMap question is not fully decided yet, but this PR is necessary for us to continue the work. We can still change to a config map if we need to, before enabling the whole feature up & downstream.

xeviknal:
@jotak given the definition of the new CRD (MonitoringDashboard), how Kiali knows in which Details page will have to place that dashboard? Is that dashboard appearing in service, app and workload details page?

jotak:
@xeviknal as discussed, this will be done through this task: https://issues.jboss.org/browse/KIALI-1975 (by reading pods annotations)
The dashboard is going to appear in Workloads and App, not Service (as Service isn't directly running any container)

Issue #689, opened 2018-11-29
Go Panic when Grafana returns 401 Unauthorized


leosunmo:
**General remarks**
I've encountered a Go panic when running Kiali against a password protected Grafana instance as the Grafana handler doesn't check for HTTP response code and doesn't marshal in to a struct, but rather uses type assertion.

**Describe the bug**
When trying to connect the my Grafana instance Kiali panics with the following:
```
2018/11/29 20:07:25 http: panic serving 172.22.10.163:56216: interface conversion: interface {} is map[string]interface {}, not []interface {}
goroutine 136363 [running]:
net/http.(*conn).serve.func1(0xc4203f9680)
	/home/travis/.gimme/versions/go1.9.4.linux.amd64/src/net/http/server.go:1697 +0xd0
panic(0x11ca4a0, 0xc420e68240)
	/home/travis/.gimme/versions/go1.9.4.linux.amd64/src/runtime/panic.go:491 +0x283
github.com/kiali/kiali/handlers.findDashboardPath(0xc420551660, 0x20, 0x133b5dd, 0x1b, 0x0, 0x0, 0x0, 0x0)
	/home/travis/gopath/src/github.com/kiali/kiali/handlers/grafana.go:102 +0x578
github.com/kiali/kiali/handlers.getGrafanaInfo(0x13ae390, 0x13ae388, 0xc420391da0, 0x1c17100, 0x19ff4faa164b, 0xbef82f9f65a9aad8)
	/home/travis/gopath/src/github.com/kiali/kiali/handlers/grafana.go:66 +0x316
github.com/kiali/kiali/handlers.GetGrafanaInfo(0x1bb9280, 0xc4207840e0, 0xc4200f6600)
	/home/travis/gopath/src/github.com/kiali/kiali/handlers/grafana.go:24 +0x3d
net/http.HandlerFunc.ServeHTTP(0x13ae2b8, 0x1bb9280, 0xc4207840e0, 0xc4200f6600)
	/home/travis/.gimme/versions/go1.9.4.linux.amd64/src/net/http/server.go:1918 +0x44
github.com/kiali/kiali/routing.metricHandler.func1(0x1bb9280, 0xc4207840e0, 0xc4200f6600)
	/home/travis/gopath/src/github.com/kiali/kiali/routing/router.go:73 +0x97
net/http.HandlerFunc.ServeHTTP(0xc4203bd680, 0x1bb9280, 0xc4207840e0, 0xc4200f6600)
[...]

PR #690, opened 2018-11-30
KIALI-2048 Don't panic on Grafana error

jotak:
- Better handling of grafana errors & unmarshalling
- Propagate error message
- More tests

JIRA: https://issues.jboss.org/browse/KIALI-2048

### DIFF
--- a/handlers/grafana.go
+++ b/handlers/grafana.go
@@ -17,12 +17,12 @@ import (
 
 type osRouteSupplier func(string, string) (string, error)
 type serviceSupplier func(string, string) (*v1.ServiceSpec, error)
-type dashboardSupplier func(string, string) (string, error)
+type dashboardSupplier func(string, string) ([]byte, int, error)
 
 // GetGrafanaInfo provides the Grafana URL and other info, first by checking if a config exists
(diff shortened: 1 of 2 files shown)


jotak:
Fixing issue #689 cc @leosunmo

leosunmo:
It should be pointed out that the documentation on the main README for `GRAFANA_SERVICE_NAMESPACE` is incorrect as it states that the setting will be ignored if `GRAFANA_URL` is supplied, which is not the case. If I run the above configuration _without_ `external_services.grafana.service_namespace` it does _not_ panic and it simply "can't find grafana service".

jotak:
Thanks @leosunmo 

> It should be pointed out that the documentation on the main README for `GRAFANA_SERVICE_NAMESPACE` is incorrect as it states that the setting will be ignored if `GRAFANA_URL` is supplied, which is not the case.

That's true, I think that was modified after istio 1.0 release, we had to change our way to query grafana but the doc wasn't updated as it should. I'll do it in another PR. Thanks for pointing it out.

aljesusg:
Please @jotak rebase this PR

PR #698, opened 2018-12-03
KIALI-1969 KIALI-1994 Some Endpoints have double /api when swagger file is parsed / Root Swagger API is directing to an error

aljesusg:
** Describe the change **

- Rename description swagger to be consistent with convetion
  - GetToken => getToken
  - Root => root
- Set StrictSlash to true to 

```go
When true, if the route path is "/path/", accessing "/path" will perform a redirect
// to the former and vice versa. In other words, your application will always
// see the path as specified in the route.
```

** Issue reference **

JIRA: [KIALI-1969](https://issues.jboss.org/browse/KIALI-1969
)  [KIALI-1994](https://issues.jboss.org/browse/KIALI-1994)

@gbaufake Can you check if tehre is /api/api routes? I didn't see them

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -17,7 +17,7 @@ func NewRouter() *mux.Router {
 	webRoot := conf.Server.WebRoot
 	webRootWithSlash := webRoot + "/"
 
-	rootRouter := mux.NewRouter().StrictSlash(false)
+	rootRouter := mux.NewRouter().StrictSlash(true)
 	appRouter := rootRouter
 
 	// Due to PathPrefix matching behavoir on sub-routers
--- a/routing/routes.go
+++ b/routing/routes.go
@@ -386,7 +386,7 @@ func NewRoutes() (r *Routes) {
 			handlers.NamespaceList,
 			true,
 		},
-		// swagger:route GET /api/namespaces/{namespace}/services/{service}/metrics services serviceMetrics
+		// swagger:route GET /namespaces/{namespace}/services/{service}/metrics services serviceMetrics
 		// ---
 		// Endpoint to fetch metrics to be displayed, related to a single service
 		//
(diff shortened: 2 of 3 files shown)


gbaufake:
@aljesusg  I think this PR solves 
[KIALI-1994](https://issues.jboss.org/browse/KIALI-1994) but it doesn't solve KIALI-1969.


Double api when the swagger is parsed: (/api/namespaces/{namespace}/apps/{app}/health) - https://github.com/kiali/kiali/blob/master/swagger.json#L41

Without double api when the swagger is parsed: /namespaces/{namespace}/applications/{app}/graph (https://github.com/kiali/kiali/blob/master/swagger.json#L896)

The problem is the /api before namespaces on line 41, so when it parses, it add root path (/api/), becoming double api

routing/router.go
```diff
@@ -17,7 +17,7 @@ func NewRouter() *mux.Router {
 	webRoot := conf.Server.WebRoot
 	webRootWithSlash := webRoot + "/"
 
-	rootRouter := mux.NewRouter().StrictSlash(false)
```
mwringe: This was changed as part of the custom context root work (https://github.com/kiali/kiali/commit/3629100a3ca78a6c3831305c2086f32ba00e9998)

Are we sure this continues to work when the context root is something else?

gbaufake: @mwringe 
I tested on https://github.com/kiali/kiali-client-python/commit/3e3c06dc1c5ed2eaeb6f19d296d4b6cb4de98326 and it worked with custom context

gbaufake:
Verified with new kiali-python-client test and the double api error is gone
https://github.com/kiali/kiali-client-python/blob/11417f88810536ad2d995ad26d6bc89a531fc129/tests/test_api_swagger.py#L14-L15

Verified root_path as well (https://github.com/kiali/kiali-client-python/blob/11417f88810536ad2d995ad26d6bc89a531fc129/tests/test_api_methods.py#L51-L52)

Verified tested custom context on kiali/kiali-client-python@3e3c06d and it worked as expected.

PR #703, opened 2018-12-04
KIALI-2078 Return Forbidden when requesting inaccessible namespace

jshaughn:


graph/util.go
```diff
@@ -0,0 +1,38 @@
...
+	})
+}
+
+// Forbidden panics with Forbidden and the provided message
+func Forbidden(message string) Response {
+	panic(Response{
+		Message: message,
+		Code:    http.StatusForbidden,
+	})
+}
```
jmazzitelli: not that it matters, but why don't Forbidden and Error both just delegate to Panic?

```
func Error(message string) Response { Panic(message, http.StatusInternalServerError) }
func Forbidden(message string) Response { Panic(message, http.StatusForbidden) }
```

jshaughn: yeah, that was silly. I blame Monday.  I'll change it...

PR #704, opened 2018-12-04
KIALI-2081 Only show unknown node when it has traffic

jshaughn:
- this actually managed to uncover a small issue in service graph reduction, now fixed

### DIFF
--- a/graph/appender/appender.go
+++ b/graph/appender/appender.go
@@ -36,6 +36,10 @@ func NewNamespaceInfo(namespace string) *NamespaceInfo {
 }
 
 func getWorkload(workloadName string, workloadList *models.WorkloadList) (*models.WorkloadListItem, bool) {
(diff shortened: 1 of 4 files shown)


jshaughn:
@jmazzitelli One note, this currently eliminates the unknown node if it is generating no traffic.  It keeps the unknown node if it generating *any* traffic.  Do you think we should do it a bit differently and maybe eliminate all grey edges emanating from unknown?

jmazzitelli:
> Do you think we should do it a bit differently and maybe eliminate all grey edges emanating from unknown?

I actually think that should be a separate feature... have some option "Remove edges with no traffic". Didn't we have a JIRA for this already?

jshaughn:
I don't think there is a specific Jira for that, I believe it's an idea we've talked about and relates to KIALI-1774.  OK, well if you are good then you can approve.

PR #705, opened 2018-12-04
KIALI-1954 Subset Presence Checker does not invalidate any object

xeviknal:
** Describe the change **
'Subset not found' validations doesn't leave the validation as invalid since it is a warning.

** Issue reference **
https://issues.jboss.org/browse/KIALI-1954

** Backwards incompatible? **
yes

Before:
48569741-d5bdeb00-e902-11e8-9835-36df690d267b

After:
screenshot of https___kiali-istio-system 192 168 42 18 nip io_api_namespaces_bookinfo_istio_virtualservices_reviews_istio_validations

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -69,7 +69,7 @@ func addError(validations models.IstioValidations, gatewayRuleName string, serve
 	rrValidation := &models.IstioValidation{
 		Name:       gatewayRuleName,
 		ObjectType: GatewayCheckerType,
-		Valid:      false,
+		Valid:      true,
 		Checks: []*models.IstioCheck{
 			&checks,
 		},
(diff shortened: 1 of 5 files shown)


xeviknal:
Before:
before-1954

After:
after-1954

xeviknal:
@jotak That is a good point. I guess that this field is pretty useful for sorting items in the ui. IMO, it gives a more direct way to compare items by config status. However, the fact that we have a boolean instead of a integer/string is limiting and also adds some extra computing.

xeviknal:
I have also fixed a Gateway Multimatch Checker. It returned a valid = false but it only adds warnings. In the same way as Subset Presence checker, the sort got broken because of it.

xeviknal:
@hhovsepy do you mind check this PR?

hhovsepy:
Verified the solution:
screenshot from 2018-12-06 09-04-34

PR #708, opened 2018-12-04
KIALI-2090 refactor: consolidate telem value validation to a method

jshaughn:
- simplify by getting rid of the multiple "unknown" constants

### DIFF
--- a/graph/appender/unused_node_test.go
+++ b/graph/appender/unused_node_test.go
@@ -1,8 +1,9 @@
 package appender
 
 import (
-	"github.com/kiali/kiali/models"
 	"testing"
+
+	"github.com/kiali/kiali/models"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/graph"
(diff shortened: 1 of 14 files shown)


graph/appender/unused_node_test.go
```diff
@@ -1,9 +1,10 @@
 package appender
 
 import (
-	"github.com/kiali/kiali/models"
 	"testing"
 
+	"github.com/kiali/kiali/models"
+
```
jmazzitelli: remove this space, it should be grouped with the rest of kiali imports

graph/appender/unused_node_test.go
```diff
@@ -1,9 +1,10 @@
...
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/graph"
 	"github.com/stretchr/testify/assert"
```
jmazzitelli: this is a third-party, non-kiali, import. It should be moved up above the kiali group of imports into its own group to match our coding standards.

go-imports
blank-line
third-party imports
blank-line
kiali imports

graph/util.go
```diff
@@ -25,8 +27,14 @@ func Panic(message string, code int) Response {
...
 
+// CheckError panics with the supplied error if it is non-nil
 func CheckError(err error) {
 	if err != nil {
 		panic(err.Error)
 	}
 }
+
+// IsOK just validates that a telemetry label value is not empty or unknown
+func IsOK(telemetryVal string) bool {
```
jmazzitelli: "IsOK" for a generic util function seems very unintuitive especially when it just takes a string and returns a boolean and is part of a util package. Perhaps rename it "IsNotUnknown" or "IsValidTelemetryLabel" or something? I don't care too much about fixing this, but something to think about.

jshaughn: I thought about that but opted for the short name and a comment.  It's a util but on;y for the graph package.

PR #702, opened 2018-12-03
KIALI-2038 - Refactor to support redesigned Kiali-Client

mattmahoneyrh:
Refactor to support Kiali-Client redesign KIALI-1766.

PR #709, opened 2018-12-04
KIALI-2038 - Refactor to support redesigned Kiali-Client

gbaufake:
- Refactor to support Kiali-Client redesign KIALI-1766.
- Updating with Kiali Client test (Updates from #702)
- 65 tests on total


mattmahoneyrh:
kiali-client version in requirements.txt needs to be changed to "kiali-client==0.5.0"

PR #710, opened 2018-12-05
KIALI-2086 Optimize and simplify metrics fetching

jotak:
- Queries now must specify reporter and direction (in/out), avoiding to load much more metrics than necessary
- Response model is adapated, not providing reporter/direction anymore

This is removing from backend some of the business logic about reporter and direction

JIRA: https://issues.jboss.org/browse/KIALI-2086

Frontend PR required: https://github.com/kiali/kiali-ui/pull/871

** Backwards incompatible? **

- [x] Is your change introducing backwards incompatible changes?

Yes, metrics API changes, needs frontend PR

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -177,23 +177,23 @@ type AvgParam struct {
 }
 
 // swagger:parameters serviceMetrics appMetrics workloadMetrics
-type ByLabelsInParam struct {
-	// List of labels to use for grouping inbound metrics (via Prometheus 'by' clause).
+type ByLabelsParam struct {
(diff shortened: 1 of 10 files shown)


jotak:
@jshaughn @israel-hdez can you check this please?
As explained in the JIRA, this is a change in metrics API, to optimize many of the requests. It's also going to help in the "runtimes monitoring" feature as with this model I'll be able to mutualize some parts of the model and the API.

jotak:
@jshaughn yes, about the graph side panel, there's some cases where we have to make 2 calls (but never more than 2 at the moment).

PR #711, opened 2018-12-05
KIALI-2050 Grafana API Key/Basic auth support

leosunmo:
** Describe the change **

This is a first pass at adding API key and basic auth support to Grafana web requests.

I was able to successfully authenticate and access the Grafana dashboards from Kiali using an API key with "viewer" permissions in Grafana.
I _wasn't_ able to authenticate with basic auth as it seems that doesn't work in Grafana if you have LDAP/SAML enabled.

Not sure if any of this functionality requires any additional tests to be added, but if you deem it necessary I'll see what I can do!

** Issue reference **

This adds functionality referenced in https://github.com/kiali/kiali/issues/689.

** Backwards incompatible? **

There should be no backwards incompatibilities issues.


leosunmo:
I'm considering building the bearer token in a function and pass that string (`$username:$password` or `api_key:$apiKey` ) so I don't have to pass around a struct, but I noticed that the [API key as basic auth](https://github.com/grafana/grafana/pull/12416) is a [new-ish feature](https://github.com/grafana/grafana/issues/13065#issuecomment-416860939).

It felt like a safer move to add it directly as a bearer token.

jotak:
@leosunmo do you use configmap or directly environment variable to configure in your environment?
For API key I think it's better to use a Secret, which is "translated" into the environment variable via the deployment yaml.

See how we do it for server credentials: https://github.com/kiali/kiali/blob/master/deploy/kubernetes/secret.yaml and https://github.com/kiali/kiali/blob/master/deploy/kubernetes/deployment.yaml#L42-L51
I guess the same could be done for Grafana credentials (and we can reuse the same Secret file, I guess). This is just some additional plumbing, your code in `config.go` looks fine.

Also, could you remove the commit about `glide.lock` (I think there's no library upgrade here, correct?). If the vendor files need to be updated that must be done via a dedicated PR, but I guess you don't need it for this feature.

leosunmo:
I used straight environment variables in my Deployment for testing but would use Secrets in prod. From inside the container it's all the same so I didn't bother. I didn't see that you had some example deployment files, would you like me to add to the secret file?

I don't know if adding it to https://github.com/kiali/kiali/blob/master/deploy/kubernetes/deploy-kiali-to-kubernetes.sh like you've currently got with the Kiali password makes sense, since adding Grafana access is not mandatory. Thoughts?

leosunmo:
I tidied up the credentials building a bit by removing the new struct and building the auth header early on instead.

leosunmo:
You are absolutely correct. Looking at https://golang.org/src/net/http/request.go?s=29952:30009#L906 that's probably what I should be using. I just pushed a change, please give it another try if it's not too much of a hassle!

PR #718, opened 2018-12-05
KIALI-2087 use citadel certificate to put kiali behind https on kubernetes

jmazzitelli:


jmazzitelli:
This is working. Just need some more time to test this.

jmazzitelli:
Here's how I tested - using the new hack script for k8s testing - this uses minikube - so you need minikube and kubectl available:

1. Start minikube (assumes you use virtualbox): `hack/k8s-minikube.sh up`
2. Install Istio: `hack/k8s-minikube.sh istio`
3. Put minikube's docker environment in your shell - run the command that `hack/k8s-minikube.sh docker` tells you
4. In the same shell window, install kiali using the makefile: `make build docker-build k8s-deploy`
5. Get the gateway url for minikube: `k8s-minikube.sh gwurl` - use this to access kiali
6. Now you should be able to access kiali over https via the gateway url: `https://host:port/kiali`

Because the trusting CA authority is not a real one, you need to tell the browser to accept the certificate.

sdake:
@jmazzitelli thanks a bunch for looking into this!  I think adding such functionality is of very high value for Istio 1.1.

Cheers
-steve

jmazzitelli:
@gbaufake you said you tested on GKE and it worked. If that is true, can you approve this PR so we can merge it? thx!

jmazzitelli:
PR https://github.com/kiali/kiali/pull/733 *must* be merged first before this PR. Please review and merge that before this one.

jmazzitelli:
Note this only works when going through k8s ingress - it doesn't work with gateways. I'm still looking at getting istio gateways and https configured.

hack/k8s-minikube.sh
```diff
@@ -99,16 +107,16 @@ Valid options:
...
+  up:                starts the minikube cluster
+  down:              stops the minikube cluster
+  status:            gets the status of the minikube cluster
+  delete:            completely removes the minikube cluster VM destroying all state
+  docker:            information on the minikube docker environment
+  dashboard:         enables access to the Kubernetes GUI within minikube
+  ingress:           enables access to the Kubernetes ingress URL within minikube
+  istio:             installs Istio into the minikube cluster
+  bookinfo:          installs Istio's bookinfo demo (make sure Istio is installed first)
+  gwurl [port name]: displays the Ingress Gateway URL - if port name is given, the gateway port is also shown.
```
pilhuhn: Is 'port name' literal? Or how do users know what value to supply?

jmazzitelli: it is not literal - its the actual port name specified in the istio-gateway config. For example, you can get the gateway port for "http2-prometheus" (that's the port name you can see here being defined in the helm values file: https://github.com/istio/istio/blob/ce90c45dfe8aaf46315aca1b31ddbc2d12a4d090/install/kubernetes/helm/subcharts/gateways/values.yaml#L53) - thus getting the IP:port needed to access Prometheus.

jmazzitelli: Obviously, the names of these ports vary - depends on what the helm charts use, or what users add to the istio-gateway.

jmazzitelli:
@pilhuhn I changed the README to add details about disabling the SSL stuff. I also noticed there are also things wrong in there, so I changed some other things.

hack/k8s-minikube.sh
pilhuhn: Do we know what the 'default' is, so that we can suggest that name here? I guess we don't want to return arbitrary ingress-gw+ports by default , but Kiali - no?

pilhuhn:
@gbaufake Could you please go over this (again)? Basically the diff for the readme

hack/k8s-minikube.sh
jmazzitelli: There is no "default" really - it depends what URL you want. For example, I do pass in "http2" to get the Bookinfo URL (because I know the bookinfo demo will be going over that port) - see https://github.com/kiali/kiali/blob/6e946ed231d31d9386f5cf70132b67a7b995852e/hack/k8s-minikube.sh#L178 
That way when you install bookinfo, the script will echo out the proper URL.

But, for example, the reason I did it this way is I need to know the URLs for **all** the different add-ons (not just kiali). There isn't a "default" there in that case so I needed a command to run that would get me these:

./k8s-minikube.sh gwurl http2-prometheus
./k8s-minikube.sh gwurl http2-grafana
./k8s-minikube.sh gwurl http2-tracing
./k8s-minikube.sh gwurl http-kiali

All of those return different ports - and they only work if the istio gateway has those port names defined.

We could create another command "gwurldefaults" to spit those four out?

PR #721, opened 2018-12-06
rename Config.Api -> Config.API

jsloyer:
I was starting to go through the kiali code and noticed some of it doesn't line up with some basic go linting.  I am volunteering to go through the code and clean it up a bit... This will be the first of many PR's if you are interested. 

** Issue reference **
n/a

** Backwards incompatible? **
Yes

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -58,7 +58,7 @@ func (in *NamespaceService) GetNamespaces() ([]models.Namespace, error) {
 	}
 
 	result := namespaces
-	excludes := config.Get().Api.Namespaces.Exclude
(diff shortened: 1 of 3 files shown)


pilhuhn:
Thank you @jsloyer

jotak:
Just as a side note, golint (govet?) isn't yet in our build process so most people in the team, I think, don't even see this kind of issue. Which means that similar issues could reappear at any moment.
So ... should we introduce golint in our process? (I vote for)  :-)

jshaughn:
@jotak see https://github.com/kiali/kiali/issues/722

PR #723, opened 2018-12-06
exclude ibm namespaces

jsloyer:
** Describe the change **
This PR excludes ibm vendor namespaces from kiali.  I am assuming other vendor's will want to do this as well so I am more than happy to modify the logic a bit to make it easier for vendors to exclude their namespaces

** Backwards incompatible? **
Yes

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -166,7 +166,7 @@ func NewConfig() (c *Config) {
 	c.IstioNamespace = strings.TrimSpace(getDefaultString(EnvIstioNamespace, "istio-system"))
 	c.IstioLabels.AppLabelName = strings.TrimSpace(getDefaultString(EnvIstioLabelNameApp, "app"))
 	c.IstioLabels.VersionLabelName = strings.TrimSpace(getDefaultString(EnvIstioLabelNameVersion, "version"))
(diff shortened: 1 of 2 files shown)


pilhuhn:
Thanks @jsloyer . I wonder (as you seem to hint) if this exclusion list should be put into a ConfigMap entry and be read at startup, as I expect that folks that have deployed Istio and Kiali will have their own list of namespaces to ignore.

jsloyer:
@pilhuhn I am more than happy to submit a PR for that route if you want...

jsloyer:
So here is the output... Most of the offenders are about comments for exported methods.

```

**Comment by jmazzitelli** [comment]
> Thanks @jsloyer . I wonder (as you seem to hint) if this exclusion list should be put into a ConfigMap entry and be read at startup,

it already is. see the README - its the

```
api:
  namespaces:
    exclude:
    ...the list of namespaces to exclude.
```

We implemented this I believe back in v0.10.

jsloyer:
closing as this was already done

jmazzitelli:
for @jshaughn to review too

I'm OK with this change.

jmazzitelli:
@jsloyer we can keep this - just have the default exclude ibm* namespaces. This is fine - its just a default. Users can customize their exclude list via the configmap if this default is not what they want.

PR #724, opened 2018-12-06
KIALI-2099 Fix issue in istio appender when processing outsider nodes

jshaughn:


jshaughn:
@jmazzitelli If you have any comments we'll take care of it in a future PR.

PR #725, opened 2018-12-07
KIALI-2035 Add new prometheus url for runtime metrics dashboard system

xeviknal:
** Describe the change **

When fetching metrics for runtime monitoring, we must be able to use a different prometheus URL than the one included with istio. If this alternative URL isn't set up, then fallback to istio's prometheus.

So, an alternative prometheus URL hasn't been added in config.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2035

** Backwards incompatible? **
yes


xeviknal:
@jotak what I am not fully sure about this PR is the name of the env variable name: 'PROMETHEUS_RUNTIME_METRICS_URL'. wdyt?

xeviknal:
screenshot of openshift web console

jshaughn:
naive question: Is there possibly a need for more than one additional prometheus URL?

jotak:
@jshaughn I don't see such a need at the moment... My position is that we should start refactoring if number of urls > 2 :-)

jmazzitelli:
@xeviknal @jotak I'm adding config to the new operator - and I see this custom metrics URL config setting but it is not used anywhere in the server-side that I can see. I do not see a link to any UI-related PR, so I'm not sure where this is actually used. Is this obsolete? Can I get rid of it?

jotak:
@jmazzitelli i noticed just last week it was not used, but it should be. Please do not remove, I'm already fixing it in a commit (not pushed yet)

PR #726, opened 2018-12-07
KIALI-1974 Load custom metrics & dashboard

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-1974

* First commit:
  - Load custom dashboards from CRD
  - Expose via new route & handler

* Second commit:
  - Introduce a new hard-coded Istio dashboard that reuses the dashboard model to build Inbound & Outbound Istio metrics dashboard. The goal is to have the exact same Inbound & Outbound Metrics pages as before, but using new model.
  - Create new routes & handlers for these dashboards (for workload, app and service)

Note that these changes should not break anything in the UI. "Old" metrics endpoints are still valid and can still be consumed as before. Graph side panel will continue to use the old _metrics_ routes.

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -49,7 +49,7 @@ func (in *DashboardsService) GetDashboard(params prometheus.CustomMetricsQuery, 
 		go func(idx int, chart kubernetes.MonitoringDashboardChart) {
 			defer wg.Done()
 			filledCharts[idx] = models.ConvertChart(chart)
-			if chart.MetricType == "counter" {
+			if chart.MetricType == kubernetes.Counter {
 				filledCharts[idx].CounterRate = in.prom.FetchRateRange(chart.MetricName, labels, grouping, &params.BaseMetricsQuery)
 			} else {
 				filledCharts[idx].Histogram = in.prom.FetchHistogramRange(chart.MetricName, labels, grouping, &params.BaseMetricsQuery)
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -1,0 +1,59 @@
+package business
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/models"
+	"github.com/kiali/kiali/prometheus"
+)
+
+// DashboardsService deals with fetching dashboards from k8s client
+type DashboardsService struct {
+	prom prometheus.ClientInterface
+	mon  *kubernetes.KialiMonitoringClient
+}
+
+// NewDashboardsService initializes this business service
+func NewDashboardsService(mon *kubernetes.KialiMonitoringClient, prom prometheus.ClientInterface) DashboardsService {
+	return DashboardsService{prom: prom, mon: mon}
+}
+
+// GetDashboard returns a dashboard filled-in with target data
+func (in *DashboardsService) GetDashboard(params prometheus.CustomMetricsQuery, template string) (*models.MonitoringDashboard, error) {
+	dashboard, err := in.mon.GetDashboard(params.Namespace, template)
+	if err != nil {
+		return nil, err
+	}
+
+	labels := fmt.Sprintf(`{namespace="%s",app="%s"`, params.Namespace, params.App)
+	if params.Version != "" {
+		labels += fmt.Sprintf(`,version="%s"`, params.Version)
+	}
+	labels += "}"
+	grouping := strings.Join(params.ByLabels, ",")
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(dashboard.Spec.Charts))
+	filledCharts := make([]models.Chart, len(dashboard.Spec.Charts))
+
+	for i, c := range dashboard.Spec.Charts {
+		go func(idx int, chart kubernetes.MonitoringDashboardChart) {
+			defer wg.Done()
+			filledCharts[idx] = models.ConvertChart(chart)
+			if chart.MetricType == "counter" {
+				filledCharts[idx].CounterRate = in.prom.FetchRateRange(chart.MetricName, labels, grouping, &params.BaseMetricsQuery)
+			} else {
+				filledCharts[idx].Histogram = in.prom.FetchHistogramRange(chart.MetricName, labels, grouping, &params.BaseMetricsQuery)
+			}
+		}(i, c)
+	}
+
+	wg.Wait()
+	return &models.MonitoringDashboard{
+		Title:  dashboard.Spec.Title,
+		Charts: filledCharts,
+	}, nil
+}
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/prometheus"
@@ -13,11 +14,11 @@ import (
 // DashboardsService deals with fetching dashboards from k8s client
 type DashboardsService struct {
 	prom prometheus.ClientInterface
-	mon  *kubernetes.KialiMonitoringClient
+	mon  kubernetes.KialiMonitoringInterface
 }
 
 // NewDashboardsService initializes this business service
-func NewDashboardsService(mon *kubernetes.KialiMonitoringClient, prom prometheus.ClientInterface) DashboardsService {
+func NewDashboardsService(mon kubernetes.KialiMonitoringInterface, prom prometheus.ClientInterface) DashboardsService {
 	return DashboardsService{prom: prom, mon: mon}
 }
 
@@ -25,7 +26,12 @@ func NewDashboardsService(mon *kubernetes.KialiMonitoringClient, prom prometheus
 func (in *DashboardsService) GetDashboard(params prometheus.CustomMetricsQuery, template string) (*models.MonitoringDashboard, error) {
 	dashboard, err := in.mon.GetDashboard(params.Namespace, template)
 	if err != nil {
-		return nil, err
+		// Dashboard might be in Kiali namespace
+		cfg := config.Get()
+		dashboard, err = in.mon.GetDashboard(cfg.IstioNamespace, template)
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	labels := fmt.Sprintf(`{namespace="%s",app="%s"`, params.Namespace, params.App)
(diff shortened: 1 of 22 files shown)


business/dashboards.go
```diff
@@ -0,0 +1,59 @@
...
+	grouping := strings.Join(params.ByLabels, ",")
+
+	wg := sync.WaitGroup{}
+	wg.Add(len(dashboard.Spec.Charts))
+	filledCharts := make([]models.Chart, len(dashboard.Spec.Charts))
+
+	for i, c := range dashboard.Spec.Charts {
+		go func(idx int, chart kubernetes.MonitoringDashboardChart) {
+			defer wg.Done()
+			filledCharts[idx] = models.ConvertChart(chart)
```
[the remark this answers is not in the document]
jotak: Well, I know in some places it's done differently. But to me it looks simpler to just call a "converter" than first create an empty struct, then have it fill data from a source. It makes the conversion function fully responsible of building the output object, and I don't have to worry about questions like "is my object correctly initialized before sending to the converter?". If constructors existed in go, that would be a constructor.
Plus (but it's minor detail) I've always seen the "Parse" word being associated to text, maybe I'm biased, but I feel like I'm not "parsing" anything here :)

handlers/apps.go
```diff
@@ -85,3 +85,56 @@ func getAppMetrics(w http.ResponseWriter, r *http.Request, promSupplier promClie
 	metrics := prom.GetMetrics(&params)
 	RespondWithJSON(w, http.StatusOK, metrics)
 }
+
+// CustomMetrics is the API handler to fetch runtime metrics to be displayed, related to a single app
+func CustomMetrics(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	namespace := vars["namespace"]
+	app := vars["app"]
+	template := vars["template"]
```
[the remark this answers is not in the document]
jotak: basically there's _handlers_ > _business services_ > _clients_ , and `metrics` is a client and is sometimes directly manipulated from handlers, when there is no "business" code identified as such.

However here with dashboards I'm creating a dashboards service but not getting it through `business.Get()`, but building it directly instead. I don't see much value in initializing 10 services when we need only one (even if the initialization cost is cheap). At the beginning when we created `business.Get()` it was mostly because we saw in 90% of our handlers we were initializing almost the same services every time, but I would say it's not a hard rule.

jotak:
@xeviknal ok I've found & fixed the swagger issue

business/dashboards.go
```diff
@@ -0,0 +1,59 @@
...
+			if chart.MetricType == "counter" {
```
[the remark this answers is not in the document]
jotak: done

business/dashboards.go
```diff
@@ -0,0 +1,148 @@
...
+
+// NewDashboardsService initializes this business service
+func NewDashboardsService(mon kubernetes.KialiMonitoringInterface, prom prometheus.ClientInterface) DashboardsService {
+	return DashboardsService{prom: prom, mon: mon}
+}
+
+// GetDashboard returns a dashboard filled-in with target data
+func (in *DashboardsService) GetDashboard(params prometheus.CustomMetricsQuery, template string) (*models.MonitoringDashboard, error) {
+	dashboard, err := in.mon.GetDashboard(params.Namespace, template)
+	if err != nil {
```
lucasponce: I don't get this logic (as external reviewer).
I understand that if first fetch fails 404/500 it goes to the default fetch in Kiali namespace.
Probably I'm missing something, but as this logic is pretty new I would suggest to add more comments just to follow the main flow of the logic.

lucasponce: For example, I don't understand if query is asking for ns-1 and that doesn't exist, why it shouldn't return a 404 instead to fall into the istio-ns one.

business/dashboards.go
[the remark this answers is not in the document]
lucasponce: Yes. This is one things I don't like about golang, that hybrid syntax between pure C vs C++ model.
As classes and objects doesn't really exist, the "entry points" some times are confused to read and I can understand different (but all correct) opinions here.

Personally, sometimes I prefer to have a helper function rather than an entry point, specially for minor tasks, but that's pretty biased as well.
But I agree, not on this PR, but code has several inconsistencies that at some point we may want to address.

business/dashboards.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// NewDashboardsService initializes this business service
+func NewDashboardsService(mon kubernetes.KialiMonitoringInterface, prom prometheus.ClientInterface) DashboardsService {
+	return DashboardsService{prom: prom, mon: mon}
+}
+
+// GetDashboard returns a dashboard filled-in with target data
+func (in *DashboardsService) GetDashboard(params prometheus.CustomMetricsQuery, template string) (*models.MonitoringDashboard, error) {
+	dashboard, err := in.mon.GetDashboard(params.Namespace, template)
```
lucasponce: "in.mon.GetDashboard" fetches the "definition" of the dashboard, meanwhile "DashboardsService.GetDashboard" fetches the dashboard populated with data. Am I right ?
If so I feel more easy to read if the "in.mon.GetDashboard" -> "in.mon.GetDashboardDef/Spec/Whatever" just to differentiate.

business/dashboards.go
```diff
@@ -0,0 +1,148 @@
...
+	}, nil
+}
+
+type istioChart struct {
+	models.Chart
+	refName string
+}
+
+var istioCharts = []istioChart{
+	istioChart{
```
lucasponce: Minor:
istioChart is not necessary here as it's declared in the array.

business/dashboards.go
```diff
@@ -0,0 +1,148 @@
...
+		Chart: models.Chart{
```
lucasponce: These values are good to be hardcoded ?
It sounds natural to read them from a config file as well.

business/dashboards.go
```diff
@@ -0,0 +1,148 @@
...
+			Name:  "TCP sent",
+			Unit:  "bps",
+			Spans: 12,
+		},
+		refName: "tcp_sent",
+	},
+}
+
+// GetIstioDashboard returns Istio dashboard (currently hard-coded) filled-in with metrics
+func (in *DashboardsService) GetIstioDashboard(params prometheus.IstioMetricsQuery) (*models.MonitoringDashboard, error) {
```
lucasponce: "GetIstioDashboard" is the refactoring of the current metrics tab coming from the Istio telemetry ?
To differentiate from the runtime metrics ?

xeviknal:
@jotak is this the expected output of /namespace/{ns}/service/{svc}/dashboard?

screenshot of https___kiali-istio-system 192 168 42 86 nip io_api_namespaces_bookinfo_services_details_dashboard

business/apps.go
```diff
@@ -192,3 +195,31 @@ func fetchNamespaceApps(k8s kubernetes.IstioClientInterface, namespace string, a
...
+			}
+		}
+	}
+	mon, err := kubernetes.NewKialiMonitoringClient()
+	if err != nil {
+		log.Error("Cannot get dashboards - cannot load Kiali Monitoring Client")
+		return
+	}
+	for _, ref := range uniqueRefs {
+		title := getCustomDashboardTitle(mon, namespace, ref)
```
lucasponce: Ok, I understand that the Pods have the annotation if there is some custom dashboard available.
Then, with those ref, we fetch just the title or the whole dashboard definition ?

I miss some comments with the intention of the logic here as well.

business/workloads.go
```diff
@@ -989,3 +992,52 @@ func controllerPriority(type1, type2 string) string {
...
+var customDashboardTitle = make(map[string]string)
+
+// TODO: put this function in DashboardsService when related PR in merged
+func getCustomDashboardTitle(mon *kubernetes.KialiMonitoringClient, namespace, name string) string {
+	// @ is forbidden charatecter in k8s resource name, so safe to use here
+	key := name + "@" + namespace
+	if title, ok := customDashboardTitle[key]; ok {
+		return title
+	}
+	dashboard, err := mon.GetDashboard(namespace, name)
```
lucasponce: Similar comments here than in the apps.go code

models/dashboards.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	sort.Slice(aggs, func(i, j int) bool {
+		return aggs[i].DisplayName < aggs[j].DisplayName
+	})
+	return aggs
+}
+
+func buildIstioAggregations(local, remote string) []Aggregation {
+	return []Aggregation{
+		Aggregation{
```
lucasponce: Minor: Aggregation here is not necessary.

models/pod.go
```diff
@@ -22,6 +23,7 @@ type Pod struct {
 	Status              string            `json:"status"`
 	AppLabel            bool              `json:"appLabel"`
 	VersionLabel        bool              `json:"versionLabel"`
+	CustomDashboards    []string          `json:"-"`
```
lucasponce: Why this mapping has not a "customDashboards" ?

routing/routes.go
```diff
@@ -497,6 +497,90 @@ func NewRoutes() (r *Routes) {
...
+		//      200: dashboardResponse
+		//
+		{
+			"WorkloadDashboard",
+			"GET",
+			"/api/namespaces/{namespace}/workloads/{workload}/dashboard",
+			handlers.WorkloadDashboard,
+			true,
+		},
+		// swagger:route GET /namespaces/{namespace}/apps/{app}/cstmetrics/{template} apps customDashboard
```
lucasponce: cstmetrics -> custommetrics
I prefer to have descriptive names, mixing long names with acronyms creates a own acronyms game hard to memorize.
I don't mind to type but have clear the endpoint.

routing/routes.go
```diff
@@ -497,6 +497,90 @@ func NewRoutes() (r *Routes) {
...
+		//
+		// responses:
+		//      400: badRequestError
+		//      503: serviceUnavailableError
+		//      200: dashboardResponse
+		//
+		{
+			"CustomDashboard",
+			"GET",
+			"/api/namespaces/{namespace}/apps/{app}/cstdashboard/{template}",
```
lucasponce: here the swagger comment is different

business/apps.go
jotak: Looks like I messed up with my commits, this one was about KIALI-1975... it shouldn't be here. I'll push again, sorry :-(
(which doesn't invalidate your remark, I can add a code comment)

business/dashboards.go
jotak: Well, that's just layers doing stuff at their levels... we have the same for GetService, GetNamespace, probably several others... should we change everything?

business/dashboards.go
jotak: You're right that it should be commented.
This is an override mechanism. The reason is that we're going to provide some default dashboards that may not have to be installed in every namespaces. But those dashboards can be overriden per namespace, or user can create new dashboards in their app namespace.

business/dashboards.go
jotak: I think it could eventually go in a future step, TBH  I haven't thought a lot if and how it would be doable (because there's some particularities with istio metrics, such as the reporter and source/destination concepts) so it wouldn't fit in the generic format. So the first-easy step is to hardcode.

business/dashboards.go
jotak: Yes

business/workloads.go
jotak: wrong commit

models/pod.go
jotak: we don't need it at the pods level, in the UI. But it's used in the go code.

xeviknal:
Mesh arena working on istio 1.0.2. Players are sweating and appending prom metrics:
screenshot of https___kiali-istio-system 192 168 42 86 nip io_api_namespaces_mesh-arena_apps_ui_cstdashboard_vertx_duration 3600
As observed with @jotak, mesh arena is not working on istio 1.0.3 (telling this for future reviewers).

jotak:
Thanks for your time @xeviknal !
For your information, it's now fixed with istio 1.0.4: adding this annotation for prometheus did the trick
`prometheus.io/scheme: "http"`

PR #727, opened 2018-12-07
KIALI-2112 fix default reporter value when absent from query

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-2112


jotak:
cc @mattmahoneyrh

jotak:
@jshaughn you're right, I've just added more validations

Issue #722, opened 2018-12-06
Adding a linter


jsloyer:
I am a big fan of linters and noticed there currently isn't a linter in kiali for the code base.  How would you feel about adding a linter and start working towards cleaning up the code base a bit and then eventually putting the linter into the travis PR flow to catch PR's that dont lint?

I am a big fan of [GoMetaLinter](https://github.com/alecthomas/gometalinter) but open to other suggestions as well.

I wanted to get the conversation going from a dialog perspective

jshaughn:
Do you have an example of what the linter spits out for the current code base?   We have avoided linters in the past due to the tools generating too many unwanted recommendations, and the fact that refactors introduce regressions.

jsloyer:
Let me do a run with it and show what it flags... hold please :)

jsloyer:
So here is the output... Most of the offenders are about comments for exported methods as well as using const's instead of repeating declaring similar values.

```
(⎈ |zues:default) jsloyer@MacBook-Pro  ~/dev/go/src/github.com/kiali/kiali   master  gometalinter --disable-all --enable=vet --enable=golint --enable=vetshadow --enable=ineffassign --enable=goconst --tests  --vendor ./...
business/workloads.go:245:29:warning: 3 other occurrence(s) of "Pod" found in: business/workloads.go:505:8 business/workloads.go:694:29 business/workloads.go:895:8 (goconst)
business/workloads.go:505:8:warning: 3 other occurrence(s) of "Pod" found in: business/workloads.go:245:29 business/workloads.go:694:29 business/workloads.go:895:8 (goconst)
business/workloads.go:694:29:warning: 3 other occurrence(s) of "Pod" found in: business/workloads.go:245:29 business/workloads.go:505:8 business/workloads.go:895:8 (goconst)
business/workloads.go:895:8:warning: 3 other occurrence(s) of "Pod" found in: business/workloads.go:245:29 business/workloads.go:505:8 business/workloads.go:694:29 (goconst)
business/workloads.go:254:15:warning: 5 other occurrence(s) of "ReplicaSet" found in: business/workloads.go:369:27 business/workloads.go:433:8 business/workloads.go:703:15 business/workloads.go:802:27 business/workloads.go:841:8 (goconst)
[...]

PR #730, opened 2018-12-07
add gometalinter but dont lint

jsloyer:
** Describe the change **
This adds a make task to install gometalinter and also adds it to the travis run but does not enforce anything, it just passes no matter what.

** Issue reference **
https://github.com/kiali/kiali/issues/722

** Backwards incompatible? **

- [x] Is your change introducing backwards incompatible changes?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,13 @@ env:
 # See: https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory
 install:
  - make swagger-install
+ - make gometalinter-install
 #- make dep-install
 #- make dep-update
 
 script:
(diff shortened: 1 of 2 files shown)


jotak:
Thanks @jsloyer 
If I understand correctly, running `make lint` will just do nothing at the moment (due to `--disable-all`), and we will enable linters one by one when we think they're useful.

I'll start a topic on kiali-dev mailing list for wider discussion on linter. This is a topic that was already discussed previously but we haven't reached a consensus yet.

jotak:
See https://groups.google.com/forum/#!topic/kiali-dev/ZjJ5MgSmO9Y

israel-hdez:
@jotak, I think the good PR to review is #732.
Most likely, we should close this one, since #732 already covers changes in this PR.

jsloyer:
@israel-hdez I am fine with that, I was just opening it up so you could see how the linter would run, I am fine to just close #730 and review #732 instead

jsloyer:
closing in preference of #732

PR #733, opened 2018-12-07
KIALI-2009 change service account name to match maistra and upstream helm

jmazzitelli:
We need this to remain consistent no matter how kiali is installed. This is important because citadel names the secrets with the name of the service account.

### DIFF
--- a/deploy/kubernetes/clusterrolebinding.yaml
+++ b/deploy/kubernetes/clusterrolebinding.yaml
@@ -11,5 +11,5 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 subjects:
 - kind: ServiceAccount
-  name: kiali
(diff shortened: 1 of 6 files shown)


jmazzitelli:
I need someone to review this and merge because it is needed for https://github.com/kiali/kiali/pull/718

jmazzitelli:
don't merge yet - the name might actually change to REMOVE -service-account from the other names. See: https://github.com/istio/istio/pull/10362#issuecomment-445503314
If that proposal is merged, then we need to reject this PR and make sure all the SA names do NOT have -service-account on them.

jmazzitelli:
It might be a week or more (if ever) this name change happens. Let's get everything conforming to the SA names and we'll change it if upstream changes. Taking off DNM label. Let's do this now.

PR #734, opened 2018-12-10
Fix subset host+port validation issue

burmanm:
Istio's type for port is uint32 only in documentation and internally, it exports it as int and turns to int64 in parsing

Thus, we now expect int64 and hope that Istio doesn't fix anything..

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -20,7 +20,7 @@ const (
 )
 
 type Host struct {
-	Port            uint32
+	Port            int
 	Hostname        string
 	ServerIndex     int
 	HostIndex       int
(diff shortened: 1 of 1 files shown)


israel-hdez:
@burmanm What do you mean by "internally in Istio". Looking at the API definition, it's also `uint32`:

https://github.com/istio/api/blob/c85aada9bdc9f27471c3ef1ad2d43a499bfe0534/networking/v1alpha3/gateway.pb.go#L479

But maybe at parsing time it is `int` as you say. But I cannot find where it is parsed.

burmanm:
Yes, API definition says so. But I think we request it as a JSON (at least that's how the RESTClient is configured) and at that point the protobuf definitions do not hold.

PR #739, opened 2018-12-11
Fix inconsistent patch/update permissions reported

jotak:
Race condition here: when patch and update permissions are different, only the last that comes up is taken


business/istio_config.go
```diff
@@ -352,5 +354,5 @@ func getUpdateDeletePermissions(k8s kubernetes.IstioClientInterface, namespace,
 			log.Errorf("Error getting permissions [namespace: %s, api: %s, resourceType: %s]: %v", namespace, api, resourceType, permErr)
 		}
 	}
-	return canUpdate, canDelete
+	return canUpdate || canPatch, canDelete
```
lucasponce: +1
It solves the issue and it also maintains the original logic to merge "update" / "patch" verbs for Kiali.

PR #746, opened 2018-12-14
Add the ability to remove Jaeger and ES from the cluster

burmanm:
** Describe the change **

Adds a flag -nj that removes ES & Jaeger from the cluster after it has been setup. This reduces memory consumption and CPU of the running cluster.


jmazzitelli:
+1000 :)

PR #751, opened 2018-12-19
KIALI-2172 Fix error traffic aggregation in service graph reducer

jshaughn:
We were only aggregating 200 traffic.
- add test

### DIFF
--- a/handlers/graph.go
+++ b/handlers/graph.go
@@ -264,6 +264,15 @@ func addServiceGraphTraffic(target, source *graph.Edge) {
 	switch protocol {
 	case "http":
 		addToMetadataValue(target.Metadata, "rate", source.Metadata["rate"].(float64))
+		if val, ok := source.Metadata["rate3xx"]; ok {
+			addToMetadataValue(target.Metadata, "rate3xx", val.(float64))
+		}
+		if val, ok := source.Metadata["rate4xx"]; ok {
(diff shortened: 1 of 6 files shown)


handlers/graph.go
```diff
@@ -264,6 +264,15 @@ func addServiceGraphTraffic(target, source *graph.Edge) {
...
 		addToMetadataValue(target.Metadata, "rate", source.Metadata["rate"].(float64))
+		if val, ok := source.Metadata["rate3xx"]; ok {
+			addToMetadataValue(target.Metadata, "rate3xx", val.(float64))
+		}
+		if val, ok := source.Metadata["rate4xx"]; ok {
+			addToMetadataValue(target.Metadata, "rate4xx", val.(float64))
+		}
+		if val, ok := source.Metadata["rate5xx"]; ok {
+			addToMetadataValue(target.Metadata, "rate5xx", val.(float64))
+		}
```
jmazzitelli: You are not aggregating "rate2xx" - which are also valid metadata attributes - see https://github.com/kiali/kiali/blob/master/handlers/graph.go#L478-L479

jshaughn: This is resolved from removing 2xx from the edge metadata in your referenced code.  It isn't used and so we may as well reduce processing rather than add more.  The PR is updated.

PR #757, opened 2018-12-21
compare with single equals in sh

cisco-:
** Describe the change **

Substitutes the "==" comparison operator with "=" for compatibility with "sh" shell, in case you prefer this over running the install script with bash.

### DIFF
--- a/deploy/kubernetes/deploy-kiali-to-kubernetes.sh
+++ b/deploy/kubernetes/deploy-kiali-to-kubernetes.sh
@@ -20,17 +20,17 @@
 
 # The credentials can be specified either as already base64 encoded, or in plain text.
 # If the username or passphrase plain text variable is set but empty, the user will be asked for a value.
-if [ "${KIALI_USERNAME_BASE64}" == "" ]; then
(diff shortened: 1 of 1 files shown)


jmazzitelli:
The header is #!/bin/bash - these are bash scripts. I prefer to change the documentation than the scripts. I'm writing a PR for the docs to say "| bash" ... we should close this.

jmazzitelli:
@cisco- what is your preference here? Change the code or the docs?

jmazzitelli:
See the docs PR - https://github.com/kiali/kiali.io/pull/86

It's either this PR or that kiali.io PR - one or the other. Keep in mind, the header of the scripts say `#!/bin/bash` but I could go either way.

jmazzitelli:
Also, if the script is going to change to run in sh, we should change the header to be `#!/bin/sh`

jmazzitelli:
OK, talking on #kiali about this - we're just going to change the docs. The scripts already say #!/bin/bash and since most modern systems (all?) have bash, we should be good.

PR #762, opened 2018-12-28
[hack] cluster-openshift.sh - use the proper command to install kiali, latest by default

jmazzitelli:


hack/cluster-openshift.sh
```diff
@@ -264,6 +281,20 @@ else
...
+if [ "${KIALI_ENABLED}" == "true" -a "${KIALI_VERSION}" == "lastrelease" ]; then
+  get_downloader
+  eval ${DOWNLOADER} /tmp/kiali-release-latest.json https://api.github.com/repos/kiali/kiali/releases/latest
+  KIALI_VERSION=$(cat /tmp/kiali-release-latest.json |\
+    grep  "tag_name" | \
+    sed -e 's/.*://' -e 's/ *"//' -e 's/",//')
+  if [ "${KIALI_VERSION}" == "" ]; then
+    echo "ERROR: Cannot determine the latest Kiali version to install"
+    exit 1
+  fi
```
aljesusg: What do you think about to show how set the Kiali version to install like the step of install curl or get ?

jmazzitelli: I added a blurb to this error message to tell the user what env var to set.

PR #763, opened 2018-12-28
KIALI-2187 use the downloader var rather than hardcoding curl

jmazzitelli:
** Describe the change **

In the deploy scripts for both openshift and k8s, the use of curl was explicit when downloading the github json which is used to determine the latest release. They should be using the `$downloader` var since some systems don't have curl, but do have wget. We want to support both.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2187

** Backwards incompatible? **

yes

### DIFF
--- a/deploy/kubernetes/deploy-kiali-to-kubernetes.sh
+++ b/deploy/kubernetes/deploy-kiali-to-kubernetes.sh
@@ -18,6 +18,28 @@
 # is installed and in your PATH.
 ##############################################################################
 
+# Determine what tool to use to download files. This supports environments that have either wget or curl.
+# After return, $downloader will be a command to stream a URL's content to stdout.
+get_downloader() {
+  if [ ! "$downloader" ] ; then
+    # Use wget command if available, otherwise try curl
+    if which wget > /dev/null 2>&1 ; then
+      downloader="wget -q -O -"
+    fi
+    if [ ! "$downloader" ] ; then
+      if which curl > /dev/null 2>&1 ; then
+        downloader="curl -s"
+      fi
+    fi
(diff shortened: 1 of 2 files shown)


deploy/kubernetes/deploy-kiali-to-kubernetes.sh
```diff
@@ -69,7 +91,8 @@ fi
 # If asking for the last release (which is the default), then pick up the latest release.
 # Note that you could ask for "latest" - that would pick up the current image built from master.
 if [ "${IMAGE_VERSION}" == "lastrelease" ]; then
-  version_we_want=$(curl https://api.github.com/repos/kiali/kiali/releases/latest 2> /dev/null |\
+  get_downloader
+  version_we_want=$(${downloader} https://api.github.com/repos/kiali/kiali/releases/latest 2> /dev/null |\
```
abonas: can this be named a bit more clearly? perhaps kiali_version ?

jmazzitelli: @abonas done - "$kiali_version_we_want"

jmazzitelli: just to be clear -- I left the "_we_want" portion to make it more obvious to the person reading the code that this is going to be the kiali version that the user wants to install based on what the github API returns to us - it isn't the version we have downloaded/installed yet. Note that the curl (downloader) here is only grabbing the version string of the latest github release from the github API - we are saying "the github API is returning the version of kiali that we want to use." Later on in the script we will download this version that we want.

PR #771, opened 2019-01-07
KIALI-2205 Improve graph JSON field names and precision

jshaughn:
- explicit http in/out fields
- explicit tcp in/out fields
- improved percentage fields
- reduce percentage precision from 2 to 1
  - allow us to show more precision on edge
  - allows Graph Find to more easily support numeric =, != operands


jmazzitelli:
The code looks OK. Haven't run it yet.

jmazzitelli:
I assume this will require a change in kiali-ui as well. Post the PR for that in here.

PR #382, opened 2018-07-26
Rebase on latest master

jmazzitelli:

PR #773, opened 2019-01-08
KIALI-2207 Do not round timeseries when rounded result affect signifi…

jotak:
…cant decimals

JIRA: https://issues.jboss.org/browse/KIALI-2207

Screenshot before:
Before

Screenshot after:
After

### DIFF
--- a/prometheus/metrics.go
+++ b/prometheus/metrics.go
@@ -212,10 +212,11 @@ func fetchRateRange(api v1.API, metricName, labels, grouping string, q *BaseMetr
 	var query string
 	// Example: round(sum(rate(my_counter{foo=bar}[5m])) by (baz), 0.001)
 	if grouping == "" {
-		query = fmt.Sprintf("round(sum(%s(%s%s[%s])), 0.001)", q.RateFunc, metricName, labels, q.RateInterval)
+		query = fmt.Sprintf("sum(%s(%s%s[%s]))", q.RateFunc, metricName, labels, q.RateInterval)
 	} else {
-		query = fmt.Sprintf("round(sum(%s(%s%s[%s])) by (%s), 0.001)", q.RateFunc, metricName, labels, q.RateInterval, grouping)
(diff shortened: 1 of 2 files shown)


jotak:
@jshaughn / @israel-hdez I think this should be ok with graph side-panel too, but let me know if you think it's not

jshaughn:
@jotak That's interesting PromQL.  It doesn't cause the inner query to be performed twice, right?  So, no real perf impact?

jotak:
@jshaughn I'm not sure if Prometheus caches the result internally, I hope so, but at least I guess it wouldn't run the right operand of "OR" if the left operand is true, which should stands for the vast majority of the cases.

jotak:
@simonpasquier any idea about the performance impact of this?

simonpasquier:
So unless I'm mistaken, Prometheus is going to evaluate twice the inner query as it doesn't do deduplicating when parsing the whole query.

In practice if we run `round(rate(foo[1m]), 0.1) > 0.1 or rate(foo[1m])`, Prometheus should evaluate the first `rate(foo[1m])` then `round(..., 0.1) > 0.1` drops all samples below the threshold then it will evaluate the second `rate(foo[1m])`and finally it will return all series from the first query + all series from the second query that aren't already present in the first set.

Obviously it is a performance hit but I think it should be reasonable and hardly noticeable unless you're returning 10s of thousands of series.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#382](https://paste.fedoraproject.org/paste/g91nhwyQqDgAbas7JQs5ig)
  * :heavy_check_mark: run-kiali-e2e-tests [#[963]](https://paste.fedoraproject.org/paste/73pmC59lOdJTPqjdnJAutg)

rhqci:

jotak:
@jshaughn @israel-hdez so I'm merging it but also opened a ticket here for client-side rounding: https://issues.jboss.org/browse/KIALI-2217
I haven't assigned a team so feel free to take it if you want, or I will do later.

PR #381, opened 2018-07-26
KIALI-1258 issue with compound node inclusion

jshaughn:

PR #774, opened 2019-01-08
KIALI-2188 Fix root path redirection loop

israel-hdez:
https://issues.jboss.org/browse/KIALI-2188

Because of the `StrictSlash(true)` on the root URL, the muxer was redirecting from `$webRoot/` -> to -> `$webRoot` (without trailing slash). However, there is a manual redirection in the opposite way when $webRoot is not '/' (custom context case). This was causing a redirection loop making impossible to access the UI using the root Kiali URL when custom context is configured.

The loop is removed by turning on "StrictSlash" after the manual redirection is configured.

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -17,7 +17,7 @@ func NewRouter() *mux.Router {
 	webRoot := conf.Server.WebRoot
 	webRootWithSlash := webRoot + "/"
 
-	rootRouter := mux.NewRouter().StrictSlash(true)
+	rootRouter := mux.NewRouter().StrictSlash(false)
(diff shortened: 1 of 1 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#381](https://paste.fedoraproject.org/paste/jfvKIoCwAf7rz8GZujRS1w)
  * :heavy_check_mark: run-kiali-e2e-tests [#[962]](https://paste.fedoraproject.org/paste/VTvu2Ghfdqwhc22gn7gZgw)

rhqci:

PR #735, opened 2018-12-10
KIALI-1975 Runtime monitoring: list available runtimes for app and workload

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-1975

- Add CustomDashboards in app & workload details
- Get dashboards by parsing annotation kiali.io/dashboards on pods
- Memoize title to avoid reloading resource to often

PR #405, opened 2018-08-06
KIALI-1290 [1.0] graph, app box not namespaced

jshaughn:
Add proper namespacing (i.e. proper ids) to the app box nodes

PR #785, opened 2019-01-15
KIALI-2155 Update the graph VS detection

mwringe:
The way we were determining if a service in a graph had its traffic affected by a virtual service was wrong. We were assuming that the virtual service 'host' value would correspond to the hostname of the service, when it can be any valid hostname.

This change brings how the graph does its VS detection in line with how the service detail pages does things (which does correctly detect the virtual services).

** Issue reference **

https://issues.jboss.org/browse/KIALI-2155

** Backwards incompatible? **

This will change how VS badges are detected in the graph. They will now be in line with how the details page does things.

Meshes should now have their VS badge applied where appropriate to indicate that the traffic could have been modified by a VS.

### DIFF
--- a/models/virtual_service.go
+++ b/models/virtual_service.go
@@ -59,12 +59,15 @@ func (vService *VirtualService) IsValidHost(namespace string, serviceName string
 	if serviceName == "" {
 		return false
 	}
-	if hosts, ok := vService.Spec.Hosts.([]interface{}); ok {
-		for _, host := range hosts {
-			if kubernetes.FilterByHost(host.(string), serviceName, namespace) {
-				return true
-			}
-		}
+
+	protocolNames := []string{"http", "tcp"}
+	protocols := map[string]interface{}{
+		"http": vService.Spec.Http,
+		"tcp":  vService.Spec.Tcp,
+	}
+
+	if kubernetes.FilterByRoute(protocols, protocolNames, serviceName, namespace, nil) {
+		return true
 	}
 	return false
 }
--- a/graph/appender/istio_details_test.go
+++ b/graph/appender/istio_details_test.go
@@ -16,7 +16,7 @@ import (
 	"github.com/kiali/kiali/kubernetes/kubetest"
 )
 
-func setupTrafficMap() (map[string]*graph.Node, string, string, string, string, string) {
+func setupTrafficMap() (map[string]*graph.Node, string, string, string, string, string, string) {
 	trafficMap := graph.NewTrafficMap()
 
 	appNode := graph.NewNode("testNamespace", "ratings", "testNamespace", graph.Unknown, "ratings", "", graph.GraphTypeVersionedApp)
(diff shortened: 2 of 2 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#405](https://paste.fedoraproject.org/paste/uX6-F-vlb4U5lTV15LqR~g)
  * :red_circle: run-kiali-e2e-tests [#[988]](https://paste.fedoraproject.org/paste/ALwBdgDUIiI024bf0nfxwA)

models/virtual_service.go
```diff
@@ -57,12 +57,15 @@ func (vService *VirtualService) IsValidHost(namespace string, serviceName string
...
 		return false
 	}
-	if hosts, ok := vService.Spec.Hosts.([]interface{}); ok {
-		for _, host := range hosts {
-			if kubernetes.FilterByHost(host.(string), serviceName, namespace) {
-				return true
-			}
-		}
+
+	protocolNames := []string{"http", "tcp"}
```
lucasponce: I'm not reviewing this one, just looked by change, but I think the protocol names is incomplete as "tls" is also a valid protocol for a VS.

mwringe: We are following what the service details page is doing here, which is only checking for http and tcp (https://github.com/kiali/kiali/blob/master/kubernetes/istio_details_service.go#L79).

If we want to handle tls as well for virtualservices, we probably also need to update how the service details handles things

lucasponce:
For some reason, github hid this PR from my list.
Would you mind to rebase it ?
Thx.

lucasponce:
@burmanm you did some work related this area. Would you mind to take a look to this change ?
Thanks

mwringe:
Rebased

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#735](https://paste.fedoraproject.org/paste/VJPjqfnHMWWTaenZmd9LLA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1496]](https://paste.fedoraproject.org/paste/dgcOko8rp7lSShjYsy7mpg)

models/virtual_service.go
burmanm: Is there a reason we don't allow https then? Validators do accept it and also validate it.

burmanm: We should rather fix service details than repeat the mistake here also.

mwringe: This PR just brings what the graph is doing in line with that is happening with the detail pages (which just uses http and tcp, see https://github.com/kiali/kiali/blob/master/kubernetes/istio_details_service.go#L188).

If we want to improve that, then lets open another jira to add in support for tls (which is the only other option https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/#VirtualService).

lucasponce: +1 to merge this one and fix the pending tls prototocol in a following PR
@mwringe CI is claiming about some minor style issue, if you can fix it we can move on.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#752](https://paste.fedoraproject.org/paste/OdhQ1V6j-iQyQNQjIYuo-w)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1530]](https://paste.fedoraproject.org/paste/oGXbexgEMppYF8kISDrI3A)

PR #418, opened 2018-08-09
KIALI-1304 Service details update - workloads added, deployments/pods removed

xeviknal:
** Describe the change **
This PR works towards istio1.0 upgrade. Updates the service details endpoint:

- Removing Pods
- Removing Deployments
- Adding workloads

#404 Needs to be merged first.

** Issue reference **

[KIALI-1304](https://issues.jboss.org/browse/KIALI-1304)

** Backwards incompatible? **
No. It needs frontend to be adapted to this change.
There is actually an ongoing PR for that:  https://github.com/kiali/kiali-ui/pull/566 ([KIALI-1218](https://issues.jboss.org/browse/KIALI-1218))

_describe the changes if appropriate_

New output for service details:
screenshot of kiali-istio-system 192 168 42 123 nip io_api_namespaces_bookinfo_services_reviews

PR #792, opened 2019-01-18
KIALI-2272 Improve protocol support

jshaughn:
- Introduce generic JSON structure for reporting traffic rates
- Add new protocol.go to centralize protocol-specific logic
  - with the exception of protocol-specific queries (like for TCP)

THIS PR REQUIRES UI PR https://github.com/kiali/kiali-ui/pull/945 . When both are approved the DNM label can be removed and the two PRs should be merged in tandem.

### DIFF
--- a/graph/appender/unused_node.go
+++ b/graph/appender/unused_node.go
@@ -80,6 +80,7 @@ func (a UnusedNodeAppender) buildUnusedTrafficMap(trafficMap graph.TrafficMap, n
 			if _, found = unusedTrafficMap[id]; !found {
 				log.Debugf("Adding unused node for workload [%s] with labels [%v]", w.Name, labels)
 				node := graph.NewNodeExplicit(id, namespace, w.Name, app, version, "", nodeType, a.GraphType)
+				// note: we don't konw what the protocol really should be, http is most common, it's a dead edge anyway
 				node.Metadata = map[string]interface{}{"httpIn": 0.0, "httpOut": 0.0, "isUnused": true}
 				unusedTrafficMap[id] = &node
 			}
--- a/graph/cytoscape/cytoscape.go
+++ b/graph/cytoscape/cytoscape.go
@@ -22,37 +22,36 @@ import (
 	"github.com/kiali/kiali/graph/options"
 )
 
+type ProtocolTraffic struct {
+	Protocol string            `json:"protocol"` // protocol
+	Rates    map[string]string `json:"rates"`    // map[rate]value
+}
+
 type NodeData struct {
 	// Cytoscape Fields
 	Id     string `json:"id"`               // unique internal node ID (n0, n1...)
 	Parent string `json:"parent,omitempty"` // Compound Node parent ID
 
 	// App Fields (not required by Cytoscape)
-	NodeType        string          `json:"nodeType"`
-	Namespace       string          `json:"namespace"`
-	Workload        string          `json:"workload,omitempty"`
-	App             string          `json:"app,omitempty"`
-	Version         string          `json:"version,omitempty"`
-	Service         string          `json:"service,omitempty"`         // requested service for NodeTypeService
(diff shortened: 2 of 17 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#418](https://paste.fedoraproject.org/paste/NcjEov~V~Ab~Fo4fy8tdtg)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1015]](https://paste.fedoraproject.org/paste/8uHVTmgiTu18YhQDI5Zdeg)

graph/appender/unused_node.go
```diff
@@ -80,6 +80,7 @@ func (a UnusedNodeAppender) buildUnusedTrafficMap(trafficMap graph.TrafficMap, n
 			if _, found = unusedTrafficMap[id]; !found {
 				log.Debugf("Adding unused node for workload [%s] with labels [%v]", w.Name, labels)
 				node := graph.NewNodeExplicit(id, namespace, w.Name, app, version, "", nodeType, a.GraphType)
+				// note: we don't konw what the protocol really should be, http is most common, it's a dead edge anyway
```
jmazzitelli: misspelled "know"

graph/cytoscape/cytoscape.go
```diff
@@ -261,41 +254,85 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
...
+		total := 0.0
+		err := 0.0
+		var percentErr, percentReq graph.Rate
+		for _, r := range p.EdgeRates {
+			rateVal := getRate(e.Metadata, r.Name)
+			switch {
+			case r.IsTotal:
+				total = rateVal
+			case r.IsErr:
+				err += rateVal
```
jmazzitelli: comments would be nice around here. I think you `+=` because there can be multiple "IsErr" rates defined and thus you need to sum them all up. Total, IsPercentErr, IsPercentReq - those do not get summed via `+=` only because you don't expect more than one of those error, so setting them with `=` rather then summing them with `+=` works.

graph/cytoscape/cytoscape.go
```diff
@@ -261,41 +254,85 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
...
+				if percentReq.Name != "" {
+					rateVal := 0.0
+					for _, r := range p.NodeRates {
+						if !r.IsOut {
+							continue
+						}
+						rateVal = total / getRate(e.Source.Metadata, r.Name) * 100.0
+						break
+					}
+					if rateVal > 0.0 && rateVal < 100.0 {
```
jmazzitelli: You don't want this if rateVal == 100? (though using equality with floating point isn't usually valid, but, seems odd you don't include this if its 100.

jshaughn: We don't currently include 100% in the JSON because that value is assumed if not provided.  It's the same as not including a 0 rate.

jmazzitelli: I'm pretty sure UXD told us not to do that - they want to see 100%

graph/cytoscape/cytoscape.go
jshaughn: Unless you think it's a blocker I'll add comments here in the follow-up gRPC PR.

graph/cytoscape/cytoscape.go
jshaughn: You will see 100%, but it's done on the UI side.

graph/appender/unused_node.go
jshaughn: I fixed this in the upcoming gRPC PR.

graph/cytoscape/cytoscape.go
jshaughn: I added comments in the upcoming gRPC PR.

PR #798, opened 2019-01-22
KIALI-2262 Make ServiceEntry validation work even without any services

burmanm:
** Describe the change **

Previous validator for ServiceEntries required that some services exists, otherwise the path was never triggered. If there are no v1.Services, this would fail and the host would get flagged invalid.

** Issue reference **

KIALI-2262

@simonpasquier


lucasponce:
@burmanm I've merged the PR for KIALI-2260, I think it can affect on this.
It needs a rebase, also.

burmanm:
So this is kinda stupid.. I think I fixed this issue in KIALI-2260 also (without knowing it). Other than additional tests, there's nothing to merge from here. I'll close this.

hhovsepy:
kialiqeci

PR #437, opened 2018-08-16
KIALI-1350: Prepare for 0.6.0 release

mwringe:
Prepare for the 0.6.0 release.

PR #801, opened 2019-01-23
KIALI-2287 Remove severity hardcoding from validation checkers

xeviknal:
** Describe the change **
Small refactor that remove magic numbers from validation system's severities. This way, system is a bit less error-prone since it doesn't allow devs to add severity in open literals.
Now, severity is an enumerator. Sounds easier to me.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2287

** Backwards incompatible? **
yes.

### DIFF
--- a/business/checkers/destinationrules/multi_match_checker_test.go
+++ b/business/checkers/destinationrules/multi_match_checker_test.go
@@ -53,6 +53,7 @@ func TestMultiHostMatchInvalid(t *testing.T) {
 	assert.True(validation.Valid) // As long as it is warning, this is true
 	assert.NotEmpty(validation.Checks)
 	assert.Equal(models.WarningSeverity, validation.Checks[0].Severity)
+	assert.Equal(models.CheckMessage("destinationrules.multimatch"), validation.Checks[0].Message)
 }
 
 func TestMultiHostMatchWildcardInvalid(t *testing.T) {
--- a/business/checkers/destinationrules/no_dest_checker_test.go
+++ b/business/checkers/destinationrules/no_dest_checker_test.go
@@ -51,7 +51,7 @@ func TestNoValidHost(t *testing.T) {
 	assert.False(valid)
 	assert.NotEmpty(validations)
 	assert.Equal(models.ErrorSeverity, validations[0].Severity)
-	assert.Equal("This host has no matching workloads", validations[0].Message)
+	assert.Equal(models.CheckMessage("destinationrules.nodest.matchingworkload"), validations[0].Message)
 	assert.Equal("spec/host", validations[0].Path)
 }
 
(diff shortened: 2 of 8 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#437](https://paste.fedoraproject.org/paste/-tQrvIlUemcsLjuvtyNpJg)
  * :red_circle: run-kiali-e2e-tests [#[1042]](https://paste.fedoraproject.org/paste/1vdC2raEz5WsqpLi7sE4jQ)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#438](https://paste.fedoraproject.org/paste/mWAIzEIQItW6cYPP~OAxYg)
  * :red_circle: run-kiali-e2e-tests [#[1043]](https://paste.fedoraproject.org/paste/Z6RHiCvniBSOmWEdwEvHPA)

burmanm:
Might sound like a stupid proposal, but let me know your opinion. What if we had instead of models.BuildWarning a defined list of validations? So, calling for example ``models.Build(.., "destinationrules.multimatcher")`` and then the builder would add the "warning/error/etc" status.

That way we could potentially make these user-configurable at some point in the future. If that makes sense. In the tests we could use the default-values.

business/checkers/destinationrules/multi_match_checker.go
```diff
@@ -114,8 +114,8 @@ func checkCollisions(validations models.IstioValidations, destinationRulesName s
 func addError(validations models.IstioValidations, destinationRuleNames []string) models.IstioValidations {
 	for _, destinationRuleName := range destinationRuleNames {
 		key := models.IstioValidationKey{Name: destinationRuleName, ObjectType: DestinationRulesCheckerType}
-		checks := models.BuildCheck("More than one DestinationRules for the same host subset combination",
-			"warning", "spec/host")
+		checks := models.BuildWarningCheck("More than one DestinationRules for the same host subset combination",
```
lucasponce: I don't fully understand this approach.
Moving from using a "warning" string to a custom method BuildWarningCheck doesn't make really a difference.
As @burmanm comments, if we have a generic Build() and we have a table configurable somewhere perhaps it may sense, but for this scenario I'd recommend to change the "warning" for just a constant WARNING, but I don't think that adding a explicit BuildWarningCheck simplifies it.
So, I'd prefer to move the literall to constants that can be changed "warning" / "warn".

lucasponce: I would avoid to over-arch with this.

xeviknal:
@burmanm @lucasponce is this last commit similar to what you imagined?

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#445](https://paste.fedoraproject.org/paste/YfqgXn-H8Y2ZnptwxWnIRg)
  * :red_circle: run-kiali-e2e-tests [#[1057]](https://paste.fedoraproject.org/paste/yqna4GxYmhOlTfcs0RnsNQ)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#464](https://paste.fedoraproject.org/paste/E02CH3afqsABcBuRo~SO9A)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1078]](https://paste.fedoraproject.org/paste/xQqjrDIxrXnAo7F~rb3xjQ)

hhovsepy:
Regression testing pass for all error messages.

xeviknal:
@burmanm Just removed the hardcoded messages from suite. Removing them, we make tests independent from a future I18n system, imo. We don't test the message itself but the connection from the check with the message. great!

Issue #474, opened 2018-08-31
DestinationWeight on route doesn't have a valid service (host not found)


wansuiye:
**General remarks**

**Describe the bug**
the virtualservice work well, but kiali show the error:"DestinationWeight on route doesn't have a valid service (host not found)".
![image](https://user-images.githubusercontent.com/11864894/44901296-920f0880-ad39-11e8-9cf5-39da0e9d2112.png)

```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: configserver
spec:
  hosts:
  - "configserver.test.io"
  gateways:
  - istio-gateway
  http:
  - route:
    - destination:
         port:
           number: 8888
         host: configserver.infra.svc.cluster.local
---
```

For UI issues please also add a screenshot that shows the issue.

**Versions used**
Kiali: 0.6
Istio: 1.0.1
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9)

**To Reproduce**

**Expected behavior**

lucasponce:
Thanks for the report @wansuiye. 
I suspect where the problem is, as the full FQDN seems not checked correctly.
Could you run a couple of checks ?
- Is the VirtualService deployed under "infra" namespace ?
- If so, could you change the "configserver.infra.svc.cluster.local" to just "configserver" ? 

This is just to spot the error, as I think that host under the route should be detected correctly but it seem there is a bug.

wansuiye:
@lucasponce hi, the VirtualService is deployed under default namespace and the service is deployed under infra namespace, so i used the fqdn~

wansuiye:
@lucasponce hi, in my scenario the networking objects are defined in default namespace, there are several reasons: first is that  istio don't limit it, if using a fqdn it can work well in both namespace; and if a virtualservice is crossing namespace, such as a host is used by several services（experiment or gateway, it do not belong to one), which may depend how to distinguish namespace. so it's better to support it～

pilhuhn:
@burmanm Do you know if that is still an issue?

burmanm:
@pilhuhn We haven't added support for cross-namespace host checks. So it would still fail.

ceastman-ibm:
i have a similar issue but what im trying to do is setup a reverse proxy: such as www.demoistio.com/google should show the www.google.com webpage

here is my virtual service that generates the same error message as above:
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
[...]

PR #807, opened 2019-01-24
KIALI-2245 Add gRPC support

jshaughn:
also:
    - no longer default %traffic to 100%, it's confusing. Default to 0 like everything else and report 100s.
    - change edge to a single traffic protocol, multiple protocols don't make sense.
    - allow ResponseTime to accept grpc success code (0)
    - Handle KIALI-2297 service edge response time issue
    - Handle KIALI-2296 100% label issue
    - Handle KIALI-2290 App box side panel issue
    - Fix bug with istio namespace unknown traffic rates being doubled

THIS PR REQUIRES UI PR https://github.com/kiali/kiali-ui/pull/957

### DIFF
--- a/graph/appender/response_time.go
+++ b/graph/appender/response_time.go
@@ -60,53 +60,57 @@ func (a ResponseTimeAppender) appendGraph(trafficMap graph.TrafficMap, namespace
 	log.Debugf("Generating responseTime using quantile [%.2f]; namespace = %v", quantile, namespace)
 	duration := a.Namespaces[namespace].Duration
 
+	// create map to quickly look up responseTime
+	responseTimeMap := make(map[string]float64)
+
 	// query prometheus for the responseTime info in three queries:
 	// 1) query for responseTime originating from "unknown" (i.e. the internet)
 	groupBy := "le,source_workload_namespace,source_workload,source_app,source_version,destination_service_namespace,destination_service_name,destination_workload,destination_app,destination_version"
-	query := fmt.Sprintf(`histogram_quantile(%.2f, sum(rate(%s{reporter="destination",source_workload="unknown",destination_service_namespace="%v",response_code=~"2[0-9]{2}"}[%vs])) by (%s))`,
+	query := fmt.Sprintf(`histogram_quantile(%.2f, sum(rate(%s{reporter="destination",source_workload="unknown",destination_service_namespace="%v",response_code=~"%s"}[%vs])) by (%s))`,
 		quantile,
 		"istio_request_duration_seconds_bucket",
 		namespace,
+		"2[0-9]{2}|^0$",         // must match success for all expected protocols
 		int(duration.Seconds()), // range duration for the query
 		groupBy)
 	unkVector := promQuery(query, time.Unix(a.QueryTime, 0), client.API(), a)
+	a.populateResponseTimeMap(responseTimeMap, &unkVector)
 
 	// 2) query for responseTime originating from a workload outside of the namespace. Exclude any "unknown" source telemetry (an unusual corner case)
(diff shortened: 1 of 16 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#450](https://paste.fedoraproject.org/paste/SdkBrsvPwNt-XjQyiEiWKQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1062]](https://paste.fedoraproject.org/paste/RSsEiN~34S2uyAO5LZS4JQ)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#451](https://paste.fedoraproject.org/paste/KmHP5GIbFK1FvlLHur-D-w)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1063]](https://paste.fedoraproject.org/paste/D7fnQNT17b89BombCsi~UQ)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#452](https://paste.fedoraproject.org/paste/RrnI3k1EG01wFVCjLNTSOQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1064]](https://paste.fedoraproject.org/paste/bVQe62-THzyvnmtm7druDg)

jshaughn:
@jmazzitelli Can you review the server-side in addition to @israel-hdez ?  You have a lot of familiarity with the graph generation now that you have produced your tome on the subject.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#462](https://paste.fedoraproject.org/paste/MG5j79EiuKNFuMK8C5Ccig)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1076]](https://paste.fedoraproject.org/paste/6GCrBRtA555qLOIrRvp8iA)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#474](https://paste.fedoraproject.org/paste/UJiEkAvVYvSjw3Fj4n1dMA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1096]](https://paste.fedoraproject.org/paste/ueIMQpze~5ji6c1fxJt~4w)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#475](https://paste.fedoraproject.org/paste/yrXFWVyFlguMZkB5FfL3wQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1097]](https://paste.fedoraproject.org/paste/x1sP-Uc8WEwYDcRw6R110w)

jshaughn:
@israel-hdez I have pushed the fix for the service graph generation. I will now look at the side panel issue.

jshaughn:
> Also, I consistently don't get sparklines for nodes with incoming HTTP and outgoing GRPC:

OK, I think the answer here is that it's an unfortunate corner case and I'm not sure we can do very much about it.  Here is what I think is happening...   You clicked on a node in bookinfo namespace.  For incoming traffic we use destination telemetry because the node is the request destination.  So you will get that incoming traffic and it seems you do, looking at the screenshot.  But the outgoing traffic is source telemetry because the node is the source.  But istio telemetry is destination-only, so the grpc traffic flowing to istio services will not show up.  If you actually select an  istio-system node we are smart and use destination telemetry for incoming and outgoing.  But this is a corner case, the node is in bookinfo so we don't get the dest-reported traffic to istio,

I don't see a way of really fixing this.  And actually, the same would be true if it were outgoing http traffic.  What do you think we should do?

israel-hdez:
> OK, I think the answer here is that it's an unfortunate corner case

Ouch... I understand. So, I also think that the effort to fix that is not worth.
Perhaps, it can be enough to show a legend similar to the one for the unknown node (reusing image):

image

So, for outgoing traffic going to istio namespace, we should direct the user to look at the edges or destination nodes. For me, this would be simple enough and acceptable.

PR #488, opened 2018-09-10
kiali-1383 support multiple versions check

vnugent:
** Describe the change **
Change version check logic to support multiple versions.  Kiali config map now has an entry where we can specify one or multiple version check expressions. There will be no warnings if actual Istio version satisfies _any_ of the expressions.  Note: make sure you have a space between `>=` and `1.0`.
```
    external_services:
      istio:
        istio_version_supported: [">= 1.0", ">= 0.1.0"]
```

** Issue reference **

KIALI-1383 Kiali is showing mismatch version when is running with Maistra

** Backwards compatible? **
Yes

Verification instructions:
---

1. Deploy Kiali to Openshift with Maistra
2. The version check only performs once at login so make sure you log out and relogin
3. Verify version mismatch warning no longer there. 

if you are running upstream instead of Maistra you can edit Kiali config map and change `istio_version_supported` to something different than Istio version to trigger the warning. (again you need to logout/re-login).

You can also inspect Kiali pod console log for the debug message `Checking actual Istio version`.

No warnings
kiali-1383-console log

With warnings
kiali-1838-warning-console
kiali-1383-ui-warning

PR #818, opened 2019-01-31
Use correct oc command in hack for status checks

burmanm:
** Describe the change **

The current hack/cluster-openshift.sh version uses a hardcoded command ``oc`` for status check after installation. If these return incorrect results (or command not found) then this script will never terminate and we'll see the output:

```
Waiting for Istio Deployments to be created.........................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
```

This will never recover. This will replace ``$(oc`` to use ``${MAISTRA_ISTIO_OC_COMMAND}`` which is the one rest of the script uses.

### DIFF
--- a/hack/cluster-openshift.sh
+++ b/hack/cluster-openshift.sh
@@ -503,14 +503,14 @@ if [ "$_CMD" = "up" ]; then
       echo "Wait for Istio to fully start (this is going to take a while)..."
 
       echo -n "Waiting for Istio Deployments to be created..."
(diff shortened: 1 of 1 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#488](https://paste.fedoraproject.org/paste/AKhFEi1sSt8~ULstg2ZNxw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1121]](https://paste.fedoraproject.org/paste/fgGkXQrI9JanduwvIPYN2w)

PR #819, opened 2019-02-01
Accept both istiooc and oc as valid tools when deploying Kiali to OS

burmanm:
** Describe the change **

When running ``make openshift-deploy`` the tool would look for a ``which oc`` which is not available if only the Maistra version is installed. This change polls for the istiooc from the $PATH also

### DIFF
--- a/deploy/openshift/deploy-kiali-to-openshift.sh
+++ b/deploy/openshift/deploy-kiali-to-openshift.sh
@@ -149,8 +149,9 @@ export KIALI_PASSPHRASE_BASE64
 
 # Make sure we have access to all required tools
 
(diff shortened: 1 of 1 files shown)


jmazzitelli:
> When running `make openshift-deploy` the tool would look for a `which oc` which is not available if only the Maistra version is installed.

OK, I see the problem now. We do have this in the Makefile already:

```
@$(eval OC ?= $(shell which istiooc 2>/dev/null || which oc))
```

But when we switched the `openshift-deploy` target over to using the install script rather than directly using ${OC} to do the commands, we don't pass the 'oc' executable name to the script. We just wrote the script to assume "oc"

I'm OK with this.

PR #830, opened 2019-02-07
KIALI-1711 Audit config changes

lucasponce:
This PR adds additional logging tagged with "AUDIT" for easy filtering by external logging processors (thinking on EFK systems).

It adds traces on Write operations.

** Issue reference **

https://issues.jboss.org/browse/KIALI-1711

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -342,7 +342,7 @@ server:
 ----
 
 |`SERVER_AUDIT_LOG`
-|When true, allows additional audit logging on Write operations.
+|When true, allows additional audit logging on Write operations. (default is true)
 [source,yaml]
 ----
 server:
--- a/README.adoc
+++ b/README.adoc
@@ -341,6 +341,14 @@ server:
   cors_allow_all: (true\|false)
 ----
 
+|`SERVER_AUDIT_LOG`
+|When true, allows additional audit logging on Write operations.
+[source,yaml]
+----
+server:
+  audit_log: (true\|false)
+----
+
 |`SERVER_STATIC_CONTENT_ROOT_DIRECTORY`
 |The file server will serve all static content found under this root directory.
 [source,yaml]
--- a/handlers/istio_config.go
+++ b/handlers/istio_config.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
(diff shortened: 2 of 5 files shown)


README.adoc
```diff
@@ -341,6 +341,14 @@ server:
   cors_allow_all: (true\|false)
 ----
 
+|`SERVER_AUDIT_LOG`
+|When true, allows additional audit logging on Write operations.
```
jotak: Can you add that it's true by default (if I believe what's in config.go) ?

handlers/istio_config.go
```diff
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"github.com/kiali/kiali/config"
```
jotak: wrong order

handlers/istio_config.go
```diff
@@ -315,9 +318,17 @@ func IstioConfigCreate(w http.ResponseWriter, r *http.Request) {
...
 func checkObjectType(objectType string) bool {
 	return business.GetIstioAPI(objectType) != ""
 }
+
+func audit(r *http.Request, message string) {
+	if config.Get().Server.AuditLog {
+		user := r.Header.Get("Kiali-User")
+		log.Infof("AUDIT User [%s] Msg [%s]", user, message)
+	}
+}
```
aljesusg: If we are going to start to manage and create audit logs for debug and control I think that we should move this audit to the logs folder. What do yo think?

aljesusg: In fact, I would create a method to print the message and have consistency with a Datetime.

Are these only audit logs for authorization with Istio or audit about kiali auth too?

lucasponce: The goal of the JIRA is to add additional audit log on write operations.
So, if enabled, it will have an external line with a proper AUDIT keyword for debugging.
I reused the existing log to not write just in plain fmt or to split the logs, I mean, I think the output of the container should be unique, so as this is going to be processed via EFK I guess with the keyword is just enough for the goal.
As we audit only write operations, I put the audit() function there (just for simplicity), but yes, when this need to be propagated into more places, it makes sense to move it.

jotak:
thanks @lucasponce

PR #831, opened 2019-02-07
KIALI-2311 Show MeshPolicies into IstioConfig section

xeviknal:
** Describe the change **
MeshPolicies are now listed and detailed into Istio Config section.
UI changes in here: https://github.com/kiali/kiali-ui/pull/986

** Issue reference **
https://issues.jboss.org/browse/KIALI-2311

** Backwards incompatible? **
yes

### DIFF
--- a/handlers/istio_config.go
+++ b/handlers/istio_config.go
@@ -9,7 +9,6 @@ import (
 	"github.com/gorilla/mux"
 
 	"github.com/kiali/kiali/business"
-	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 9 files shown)


xeviknal:
@hhovsepy I added one more commit for supporting creation and update of a MeshPolicy. Thanks for catching that.

handlers/istio_config.go
```diff
@@ -136,6 +138,12 @@ func parseCriteria(namespace string, objects string) business.IstioConfigCriteri
 	if checkType(types, business.Policies) {
 		criteria.IncludePolicies = true
 	}
+
+	// MeshPolicies are not namespaced. They will be only listed for the namespace
+	// where istio is deployed.
+	if checkType(types, business.MeshPolicies) && namespace == config.Get().IstioNamespace {
```
lucasponce: I guess this filter should be done in a inner layer (business).
Handler should propagate the query with IncludeMeshPolicies = true if the param has been passed to the request.

models/istio_config.go
```diff
@@ -36,6 +37,7 @@ type IstioConfigDetails struct {
 	QuotaSpec        *QuotaSpec          `json:"quotaSpec"`
 	QuotaSpecBinding *QuotaSpecBinding   `json:"quotaSpecBinding"`
 	Policy           *Policy             `json:"policy"`
+	MeshPolicy       *Policy             `json:"meshPolicy"`
```
lucasponce: if MeshPolicy is implemented with a different Kind in the backend, I'd add a new type here.
I know redundant, but it can help to maintain one-to-one mapping.
I understand the optimization, but perhaps doing this redundant type will help.

PR #834, opened 2019-02-07
KIALI-2041 Support for edge releases in jenkins

josejulio:
** Describe the change **

Enables Jenkins to do the edge releases and disables travis from doing so.


deploy/jenkins-ci/Makefile
```diff
@@ -20,9 +20,9 @@ ifeq ($(RELEASE_TYPE),minor)
   # For minor, remove the -SNAPSHOT suffix from the Makefile
   BACKEND_VERSION_TO_RELEASE ?= $(shell semver bump release $(BACKEND_VERSION))
 endif
-ifeq ($(findstring snapshot,$(RELEASE_TYPE)),snapshot)
-  # For end-of-week snapshot release, substitute the -SNAPSHOT prefix in the Makefile with
-  # the value of $(RELEASE_TYPE) -- assuming value will be snapshot.{X}
+ifeq ($(findstring $(RELEASE_TYPE),snapshot edge),$(RELEASE_TYPE))
```
israel-hdez: Same comment here as in the ui PR. Looks like for snapshots this will expand to something like:
`ifeq ($(findstring snapshot.X,snapshot edge),snapshot.X)`
and because `snapshot.X` is not in `snapshot edge` it shouldn't evaluate to true.

I haven't tried the code, so I may be wrong :)

deploy/jenkins-ci/Makefile
```diff
@@ -45,13 +45,18 @@ BUMP_BRANCH_ID ?= $(BUILD_TAG)
 
 DOCKER_NAME ?= docker.io/kiali/kiali
 
-ifeq ($(findstring snapshot,$(RELEASE_TYPE)),snapshot)
+ifeq ($(findstring $(RELEASE_TYPE),snapshot edge),$(RELEASE_TYPE))
```
israel-hdez: (same comment here)

deploy/jenkins-ci/Makefile
josejulio: You are correct here, good catch.

deploy/jenkins-ci/Makefile
josejulio: (same answer here)

deploy/jenkins-ci/Makefile
josejulio: Updated

israel-hdez:
@josejulio A side request: could you please add this line in the Jenkinsfile?
`echo "Resolved release type: ${releaseType}"`

It should go after this block:
https://github.com/kiali/kiali/blob/master/deploy/jenkins-ci/Jenkinsfile#L102-L107

It's for the cases when the job is run as `auto` to have a log about what was the actual release "mode" .

josejulio:
> One remaining issue: The image uploaded to docker is tagged as `v0.15.0-edge`, but it should be tagged as `latest` to mimic the Travis behavior.

done

> @josejulio A side request: could you please add this line in the Jenkinsfile?
> `echo "Resolved release type: ${releaseType}"`

and done

deploy/jenkins-ci/Makefile
```diff
@@ -20,10 +20,22 @@ ifeq ($(RELEASE_TYPE),minor)
...
+  RELEASE_TYPE_IS_SNAPSHOT_OR_EDGE = n
+endif
+
 ifeq ($(findstring snapshot,$(RELEASE_TYPE)),snapshot)
   # For end-of-week snapshot release, substitute the -SNAPSHOT prefix in the Makefile with
   # the value of $(RELEASE_TYPE) -- assuming value will be snapshot.{X}
   BACKEND_VERSION_TO_RELEASE ?= $(shell semver bump prerel $(RELEASE_TYPE) $(BACKEND_VERSION))
+else ifeq ($(RELEASE_TYPE),edge)
+  # For edge releases, substitute -SNAPSHOT prefix in the Makefile with 'latest'
+  BACKEND_VERSION_TO_RELEASE ?= $(shell semver bump prerel 'latest' $(BACKEND_VERSION))
```
israel-hdez: I think I didn't explain well myself. This code will push to dockerhub something similar to this:
image

(this was from my last try). The difference is that this code will put `-latest` suffix instead of `-edge`. But what Travis is currently doing is pushing a `latest` tag without any version attached:

image

I think that what you had previously was correct. It's only the docker tag that should be fixed.

deploy/jenkins-ci/Jenkinsfile
```diff
@@ -105,6 +105,7 @@ node('kiali-build') {
         returnStdout: true,
         script: "${backendDir}/deploy/jenkins-ci/bin/determine-release-type.sh").trim()
     }
+    echo "Resolved release type: ${releaseType}"
```
israel-hdez: Thanks :)

deploy/jenkins-ci/Makefile
josejulio: Sorry i got it wrong. I just pushed a change.

deploy/jenkins-ci/Makefile
```diff
@@ -20,10 +20,22 @@ ifeq ($(RELEASE_TYPE),minor)
...
+  RELEASE_TYPE_IS_SNAPSHOT_OR_EDGE = n
+endif
+
 ifeq ($(findstring snapshot,$(RELEASE_TYPE)),snapshot)
   # For end-of-week snapshot release, substitute the -SNAPSHOT prefix in the Makefile with
   # the value of $(RELEASE_TYPE) -- assuming value will be snapshot.{X}
   BACKEND_VERSION_TO_RELEASE ?= $(shell semver bump prerel $(RELEASE_TYPE) $(BACKEND_VERSION))
+else ifeq ($(RELEASE_TYPE),edge)
+  # Use 'latest' for edge releases
+  BACKEND_VERSION_TO_RELEASE ?= latest
```
israel-hdez: I thought it was OK, but I laughed after seeing that I got this:
image

Haha... still not right :)

josejulio: Ha!

I didn't see that one coming.

deploy/jenkins-ci/Makefile
```diff
@@ -45,13 +57,18 @@ BUMP_BRANCH_ID ?= $(BUILD_TAG)
 
 DOCKER_NAME ?= docker.io/kiali/kiali
 
-ifeq ($(findstring snapshot,$(RELEASE_TYPE)),snapshot)
+ifeq ($(RELEASE_TYPE_IS_SNAPSHOT_OR_EDGE),y)
   IS_SNAPSHOT ?= y
```
josejulio: Is this OK?

I'm wondering because i see that this also release a `v$(BACKEND_VERSION_BRANCH)`
```
	docker tag "$(DOCKER_NAME):v$(BACKEND_VERSION)" "$(DOCKER_NAME):v$(BACKEND_VERSION_BRANCH)"
```

israel-hdez: I think it's OK. The `v$(BACKEND_VERSION_BRANCH)` should only be pushed when it's an actual release (major/minor/patch) and not a snapshot. Last time this happened was for v0.10:
image

The *v0.10* is the "BACKEND_VERSION_BRANCH". It should always point to the latest patch release of a minor stream. When minor v0.10.0 was released, the tag v0.10 was also created and it was an alias of v0.10.0. When patch v0.10.1 was released, the tag v0.10 was updated to point to that patch release.

deploy/jenkins-ci/Makefile
josejulio: Ok, Removed the `v`

deploy/jenkins-ci/Makefile
```diff
@@ -106,18 +126,19 @@ ifdef DOCKER_PASSWORD
...
 	  CONSOLE_VERSION="$(CONSOLE_VERSION)" \
 	  CONSOLE_LOCAL_DIR="$(CONSOLE_LOCAL_DIR)" \
 	  make docker-build docker-push
 ifneq ($(IS_SNAPSHOT),y)
-	docker tag "$(DOCKER_NAME):v$(BACKEND_VERSION)" "$(DOCKER_NAME):v$(BACKEND_VERSION_BRANCH)"
+	docker tag "$(DOCKER_NAME):$(DOCKER_TAG)" "$(DOCKER_NAME):v$(BACKEND_VERSION_BRANCH)"
 	docker push "$(DOCKER_NAME):v$(BACKEND_VERSION_BRANCH)"
 	docker rmi "$(DOCKER_NAME):v$(BACKEND_VERSION_BRANCH)"
 endif
 	docker rmi "$(DOCKER_NAME):v$(BACKEND_VERSION)"
```
israel-hdez: It fails here... You missed to replace the `v$(BACKEND_VERSION)`

josejulio: Thanks for the heads up

I wonder why it didn't fail on my env... I suppose i need a way to clean it before a test.

PR #840, opened 2019-02-13
KIALI-2094 Destination Rule validation when TLS not mentioned and non-local DR present

xeviknal:
** Describe the change **
This PR adds a warning into IstioConfig section whenever a Destination Rule doesn't specify any tls and there is one Destination Rule setting up a non-local TLS rule.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2094

** Backwards incompatible? **
yes

** screenshots **

With warnings:

No trafficPolicy mentioned:
screenshot of kiali console 38

No TLS references in trafficPolicy level:
screenshot of kiali console 39

No TLS references in Port Level
screenshot of kiali console 34

Without warning scenarios:
There are no warnings when you have TLS settings in either trafficPolicy or port level.

screenshot of kiali console 37
screenshot of kiali console 36
screenshot of kiali console 33
screenshot of kiali console 35

All the yamls tested are compressed in here: [kiali-2094.zip](https://github.com/kiali/kiali/files/2861244/kiali-2094.zip)

### DIFF
--- a/business/checkers/destinationrules/traffic_policy_checker.go
+++ b/business/checkers/destinationrules/traffic_policy_checker.go
@@ -1,1 +1,60 @@
 package destinationrules
+
+import (
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/models"
+)
+
+type TrafficPolicyChecker struct {
+	DestinationRules []kubernetes.IstioObject
+}
+
+func (t TrafficPolicyChecker) Check() models.IstioValidations {
+	validations := models.IstioValidations{}
+
+	// When mTLS is not enabled, there is no validation to be added.
+	if !t.hasEnablingmTLSDR() {
+		return validations
+	}
+
+	// Check whether DRs override mTLS.
+	for _, dr := range t.DestinationRules {
+		if !enablesNonLocalmTLS(dr) {
+			check := models.Build("destinationrules.trafficpolicy.meshmtls", "spec")
+			key := models.BuildKey(DestinationRulesCheckerType, dr.GetObjectMeta().Name)
+			validation := buildDestinationRuleValidation(dr, check, true)
+
+			if _, exists := validations[key]; !exists {
+				validations.MergeValidations(models.IstioValidations{key: validation})
+			}
+		}
+	}
+
+	return validations
+}
+
+func (t TrafficPolicyChecker) hasEnablingmTLSDR() bool {
+	enablesTLS := false
+
+	for _, dr := range t.DestinationRules {
+		enablesTLS = enablesNonLocalmTLS(dr)
+		if enablesTLS {
+			break
+		}
+	}
+
+	return enablesTLS
+}
+
+func buildDestinationRuleValidation(dr kubernetes.IstioObject, checks models.IstioCheck, valid bool) *models.IstioValidation {
+	validation := &models.IstioValidation{
+		Name:       dr.GetObjectMeta().Name,
+		ObjectType: DestinationRulesCheckerType,
+		Valid:      valid,
+		Checks: []*models.IstioCheck{
+			&checks,
+		},
+	}
+
+	return validation
+}
--- a/business/checkers/destinationrules/traffic_policy_checker.go
+++ b/business/checkers/destinationrules/traffic_policy_checker.go
@@ -1,0 +1,1 @@
+package destinationrules
--- a/business/checkers/destinationrules/traffic_policy_checker.go
+++ b/business/checkers/destinationrules/traffic_policy_checker.go
@@ -7,6 +7,7 @@ import (
 
 type TrafficPolicyChecker struct {
 	DestinationRules []kubernetes.IstioObject
+	MTLSDetails      kubernetes.MTLSDetails
 }
 
 func (t TrafficPolicyChecker) Check() models.IstioValidations {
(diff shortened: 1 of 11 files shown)


business/checkers/destinationrules/traffic_policy_checker.go
```diff
@@ -0,0 +1,102 @@
...
+	_, trafficPresent := dr.GetSpec()["trafficPolicy"]
+	return trafficPresent
+}
+
+func hasmTLSSettings(dr kubernetes.IstioObject) bool {
+	return enablesAnyTLS(dr) || enablesPortTLS(dr)
+}
+
+// enablesPortTLS returns true when there is one port that specifies any TLS settings
+func enablesPortTLS(dr kubernetes.IstioObject) bool {
```
lucasponce: I'm not checking other code yet.
as we have "hasTrafficPolicy" kinda syntax it sounds to me interesting to rename this to "isPortTLSEnabled".
When I read enablesPortTLS I was expecting some side effect on the dr.

business/istio_validations.go
```diff
@@ -13,8 +14,8 @@ import (
 )
 
 type IstioValidationsService struct {
-	k8s kubernetes.IstioClientInterface
-	ws  WorkloadService
+	k8s           kubernetes.IstioClientInterface
+	businessLayer *Layer
```
lucasponce: Not sure if this breaks the pattern with the *Service struct to pass the clients.
I need to test locally to have a better look at it.

hhovsepy:
Question about a scenario I tested.
1. Make sure mTLS is NOT enabled, local or global.
2. on Bookinfo/reviews service create Traffic Routing via wizard by specifying some TLS.
screenshot from 2019-02-14 09-47-03
3. on Bookinfo/details or any other service create Traffic Routing without TLS.
Warning message will be displayed on that DR:
screenshot from 2019-02-14 09-47-50
4. Delete the Traffic Routing created in step 2.  then the Warning message is gone in DR from step 3. .

business/checkers/destinationrules/traffic_policy_checker.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	return validations
+}
+
+func (t TrafficPolicyChecker) hasEnablingmTLSDR() bool {
+	enablesTLS := false
+
+	for _, dr := range t.MTLSDetails.DestinationRules {
+		enablesTLS = enablesmTLS(dr)
```
burmanm: Wouldn't it be simpler to do:

```go
for _, dr := range t.MTLSDetails.DestinationRules {
if enablesmTLS(dr) { return true }
}
return false
```

Nitpicking .. ;)

business/istio_validations.go
```diff
@@ -97,11 +101,12 @@ func (in *IstioValidationsService) GetIstioObjectValidations(namespace string, o
 
 	// Get all the Istio objects from a Namespace
 	if objectType != Gateways {
-		wg.Add(3)
+		wg.Add(4)
 		// Gateways has limited fetching, others will require this information currently
 		go in.fetchDetails(&istioDetails, namespace, errChan, &wg)
 		go in.fetchServices(&services, namespace, "", errChan, &wg)
 		go in.fetchWorkloads(&workloads, namespace, errChan, &wg)
+		go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
```
burmanm: Shouldn't this be for DRs only?

business/checkers/destinationrules/traffic_policy_checker.go
xeviknal: Always open to this kind of things. Specially if they follow go's idiosyncrasy. :+1:

business/istio_validations.go
xeviknal: Yep, DR are only needed right now. I'll remove the non-needed-yet meshpolicies.

xeviknal: btw, any idea of how to fix the race condition I have in the tests?

business/istio_validations.go
xeviknal: I don't fully-follow you here. Do you mind elaborating a little bit here?
Do you think that I shouldn't remove `ws WorkloadService`?

business/istio_validations.go
burmanm: You do append inside each goroutine which run simultaneously. append is not thread safe, you need to use channel or equivalent.

burmanm: That is, at line 634 in ``istio_config.go``, the state of ``allDestinationRules`` is shared by each goroutine.

burmanm: https://gist.github.com/burmanm/ff447943407c9ccf5b3b79d92d6d6703

Should do the job

xeviknal: Thank you! approached and tests passed.

hhovsepy:
Verified the previous issue is also gone:
screenshot from 2019-02-15 15-09-22
screenshot from 2019-02-15 15-08-13
screenshot from 2019-02-15 15-08-57

business/istio_validations.go
burmanm: I'll let Lucas comment on this before merging, I kinda share his issue with this expose (and linking) businessLayer directly here instead of only the Service layers seeing each other.

Issue #767, opened 2019-01-01
Virtualservice's FQDN gateway validation error:"virtualservice is pointing to a no-existing gateway"


dreadbird:
**General remarks**
The gateway `default` is created at namespace: `istio-system`,and the virtualservice is created at other namespaces.So,we used FQDN,like:`default.istio-system.svc.cluster.local`to describe the gateway which the virtualservice related.

**Describe the bug**
The gateway is created at ns: `istio-system`
```
apiVersion: networking.istio.io/v1alpha3                                                                                                                                                                                                                                      
kind: Gateway                                                                                                                                                                                                                                                                 
metadata:                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      
  name: default                                                                                                                                                                                                                                                               
  namespace: istio-system"
spec:
  selector:
    istio: ingressgateway
  servers:
  - hosts:
    - '*.xip.io'
    port:
      name: http
      number: 80
      protocol: HTTP
  - hosts:
    - '*.xip.io'
    port:
      name: https
      number: 443
      protocol: HTTPS
    tls:
      mode: SIMPLE
      privateKey: /etc/istio/ingressgateway-certs/tls.key
      serverCertificate: /etc/istio/ingressgateway-certs/tls.crt
```

and the virtualservice yaml:

```
name: cas-service
createdAt: '2018-11-30T07:17:43Z'
resourceVersion: '22902650'
hosts:
  - service.inspurcloud.cn
gateways:
  - default.istio-system.svc.cluster.local
http:
  - match:
      - uri:
          prefix: /cas
    route:
      - destination:
          host: cas-service.cas.svc.cluster.local
          subset: stable
    timeout: 60s
  - match:
      - headers:
          cookie:
            regex: ^(.*\s*)?(cas-canary)(.*)?$
        uri:
          prefix: /cas
    route:
      - destination:
          host: cas-service.cas.svc.cluster.local
          subset: canary
    timeout: 60s
```

For UI issues please also add a screenshot that shows the issue.
![image](https://user-images.githubusercontent.com/2113395/50571006-acd1b680-0dd9-11e9-9b34-db3f1bc95e08.png)

**Versions used**
Kiali:0.9
Istio:1.0.4
Kubernetes flavour and version: v1.11

abonas:
@xeviknal @lucasponce fyi

stephen-dahl:
[...]

PR #841, opened 2019-02-14
KIALI-2390 Allow VirtualService gateway check to validate all namespaces

burmanm:
** Describe the change **

If FQDN format is used for the given gateway host, check against all namespaces and their gateways for a match. We already check all namespaces for gateway existing, so doing it for the VirtualServices makes sense also.

Also, added Host type to the kubernetes/types.go since this format is used in multiple places. Might be worth doing all internal checks with FQDN in the future (but this PR only adds that to few places).

** Issue reference **

KIALI-2390, issue #767 

** Backwards incompatible? **

Yes

### DIFF
--- a/business/checkers/destinationrules/multi_match_checker.go
+++ b/business/checkers/destinationrules/multi_match_checker.go
@@ -1,8 +1,6 @@
 package destinationrules
 
 import (
-	"strings"
-
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 )
@@ -11,12 +9,6 @@ const DestinationRulesCheckerType = "destinationrule"
 
 type MultiMatchChecker struct {
 	DestinationRules []kubernetes.IstioObject
-}
-
-type Host struct {
-	Service   string
-	Namespace string
-	Cluster   string
 }
 
 type subset struct {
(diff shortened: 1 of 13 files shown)


hhovsepy:
What is the functionality change in this PR, from user's perspective?

burmanm:
If you create a gateway in for example ``istio-system`` namespace, you can point to it from a ``bookinfo`` vs (using FQDN)

hhovsepy:
Like this?, gateway is in bookinfo2, but VS is in bookinfo:
screenshot from 2019-02-15 08-23-07

burmanm:
Yes, as such (although doesn't Istio require clusterName to be included to FQDN?)

pilhuhn:
@burmanm No, that cluster name is not required. Those partial FQDNs are allowed.
I would support those, but that can be done in a followup-PR.

hhovsepy:
FQDN is accepted:
screenshot from 2019-02-15 14-44-52

Without cluster name it fails:
 
screenshot from 2019-02-15 14-45-05

burmanm:
@pilhuhn Where is such mentioned? From Istio docs the only definition I could find (other than all the examples of fully qualified giving cluster name also):

```
The hosts field specifies, implicitly or explicitly, one or more fully qualified domain names (FQDN). The short name reviews, above, would implicitly expand to an implementation specific FQDN. For example, in a Kubernetes environment the full name is derived from the cluster and namespace of the VirtualSevice (for example, reviews.default.svc.cluster.local).
```

That is from [https://istio.io/docs/concepts/traffic-management/#rule-destinations]. To me the use of non-cluster-name seems like something that might work, but is not actually guaranteed in any specification?

burmanm:
And added to that, ``Note for Kubernetes users: When short names are used (e.g. “reviews” instead of “reviews.default.svc.cluster.local”), Istio will interpret the short name based on the namespace of the rule, not the service.``

So to me these indicate that short is without namespace, FQDN is with namespace and cluster.

hhovsepy:
So will it be changed to support "gateway.namespace" format?

burmanm:
No

PR #848, opened 2019-02-18
KIALI-1469 Performance improvement on health fetching

jotak:
Avoid fetching health from istio metrics when there's no sidecar
This is a side perf improvement related to https://issues.jboss.org/browse/KIALI-1469


jotak:
WIP/DNM: will add one or two tests before merge

PR #849, opened 2019-02-18
KIALI-2375 mTLS padlock considers PERMISSIVE meshpolicies

xeviknal:
** Describe the change **
Don't show 'mesh-wide mTLS partially enabled' when user has a mesh policy with mTLS in `permissive` mode (and there is any mesh-wide DR enabling mTLS)

** Issue reference **
https://issues.jboss.org/browse/KIALI-2375

** Backwards incompatible? **
yes

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -542,8 +542,6 @@ func (in *IstioConfigService) hasMeshPolicyEnabled(namespaces []string) (bool, e
 		return false, err
 	}
 
-	mtlsEnabled := false
-
 	for _, mp := range mps {
 
 		// It is mandatory to have default as a name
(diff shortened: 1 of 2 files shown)


business/istio_config.go
```diff
@@ -566,14 +564,21 @@ func (in *IstioConfigService) hasMeshPolicyEnabled(namespaces []string) (bool, e
...
+			if mtls, present := peerMap["mtls"]; present {
+				if mtlsMap, ok := mtls.(map[string]interface{}); ok {
+					// mTLS enabled in case there is an empty map or mode is STRICT
+					if mode, found := mtlsMap["mode"]; !found || mode == "STRICT" {
+						return true, nil
+					}
+				} else {
+					// mTLS enabled in case mtls object is empty
+					return true, nil
+				}
```
jotak: I'm not sure if it applies to a real-world case, but shouldn't be there another `else` case to deal with `present` being false? It seems to me that the comment above ( `// mTLS enabled in case mtls object is empty`) would also apply to that 'else' condition, no?

jotak: ( @xeviknal tell me, I might be wrong :) )

PR #850, opened 2019-02-18
[hack] work with the latest istio 1.1 snapshot.6

jmazzitelli:


jmazzitelli:
> although do you really need both of these?
> --set global.mtls.enabled=${MTLS} --set global.controlPlaneSecurityEnabled=${MTLS}

Yes, see step 2 in the new istio 1.1 install docs here: https://preliminary.istio.io/docs/setup/kubernetes/helm-install/#option-1-install-with-helm-via-helm-template

PR #851, opened 2019-02-18
KIALI-2403 looks like the latest Istio 1.1 changed istio-pilot port. We need thi…

jmazzitelli:
We need this to get the version string of Istio. This bad URL causes the about box to break and api/status to fail


jmazzitelli:
@hhovsepy @mattmahoneyrh @gbaufake I think this fixes the About box bug that Hayk found earlier.

jmazzitelli:
We should probably harden the code around this so we don't blow up. We should default to some "unknown" version string or something rather than just have the api/status endpoint fail entirely.

jshaughn:
One more comment, does anyone know why we are persisting the Status info with redux-persist, ui-side?  It doesn't seem like a good idea to hold that info over in cookies because it can present the wrong info.

PR #855, opened 2019-02-19
KIALI-2412 Handle workload-to-workload edges in service graph generation

jshaughn:
Terminate paths that are not service request paths.


jshaughn:
@jmazzitelli When this gets merged I'll update the Graph Gen doc.

PR #859, opened 2019-02-20
Add yaml to hack folder to enable mesh-wide mTLS

xeviknal:
** Describe the change **
Adding a hack yaml which enables meshwide mTLS to the whole Istio installation.

### DIFF
--- a/hack/istio/istio-mtls-meshwide.yaml
+++ b/hack/istio/istio-mtls-meshwide.yaml
@@ -1,0 +1,24 @@
+##########
+# This enables mTLS at service mesh level.
+# The MeshPolicy allows all the services to receive traffic in mTLS mode.
+# The DestinationRule allows all the entities to start communications in mTLS mode.
+##########
+---
+apiVersion: "authentication.istio.io/v1alpha1"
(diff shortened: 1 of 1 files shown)


hack/istio-mtls-meshwide.yaml
```diff
@@ -0,0 +1,25 @@
+##########
+# This enables mTLS for the whole istio installation.
```
abonas: istio -> Istio

abonas: also, it's not for Istio itself, right? Istio installation might be misleading

xeviknal: Comment approached, phrase reformulated :)

jmazzitelli:
Please put this yaml file in the "istio" directory under "hack" with the rest of them ... we have a bunch of these istio yaml files there - keep them all together. The "hack" directory is mainly for scripts starting and managing the k8s/OS cluster.

xeviknal:
@jmazzitelli moved. You are right. I was mistaken when I created that file.

PR #860, opened 2019-02-20
KIALI-2425 Fix ServiceEntry parsing for validation

burmanm:
** Describe the change **

Changes the type checking to match []interface{} first for hosts to match correctly the type. Also, changed the parsed datatype internally to make it easier to call for different purposes (such as when no protocol is known).

** Issue reference **

KIALI-2425


business/checkers/no_service_checker.go
```diff
@@ -23,6 +24,8 @@ func (in NoServiceChecker) Check() models.IstioValidations {
 		return validations
 	}
 
+	log.Infof("ServiceEntries: %v\n", in.IstioDetails.ServiceEntries)
```
lucasponce: Debug trace instead ?
It sounds odd as we don't have these info logging on these area, I wonder if it was intentional.

PR #863, opened 2019-02-21
KIALI-2426 Write Public Config to env.js from server, merge with webRoot

jotak:
Exploding config.go handler to other layers, business and config, so that it can be called from handlers and from kiali root
For the moment, handler and routes for public config are kept even if unused; exposing for hypothetical consumer.

Move webRoot into PublicConfig

JIRA: https://issues.jboss.org/browse/KIALI-2426
Frontend PR: https://github.com/kiali/kiali-ui/pull/1012


jotak:
cc @aljesusg @lucasponce while I was introducing a periodic job to check prom config, it looks like the same should be done for that Jaeger flag "JaegerAvailable" ... what do you think?

Basically what is done here is that instead of fetching anything just once at startup, it's fetched periodically (every minute), hence making kiali less dependant to startup sequence with other services (ex: if kiali and jaeger are started altogether, if kiali is ready before jaeger then it wouldn't be aware of it)

jotak:
@jshaughn as written in the JIRA, we can have a discussion prior to merging if necessary. If after some time we end up having a periodic serverConfig polling from the UI, then there's no need to do this env.js generation.

israel-hdez:
@jotak I think @jshaughn was referring to the duplicated logging.

jshaughn:
@jotak I don't think we want to poll from the UI.  I thought perhaps just pulling updated serverConfig at login or on F5 (i.e checkCreds) would be sufficient.  In other words update based on user action, not a scheduled interval.  That code was already in place and remember that the call in checkCreds was actually done to check the auth token, so getting the serverConfig was basically free.

I think we all (me, you, @israel-hdez) agree that keeping the serverConfig in persisted redux is what we'd like to remove.  It adds complexity and touches a lot of files.  The reason it's in redux is because I originally could not find a way to ensure serverConfig was present before we started rendering on an F5 (i.e. Navigation would start rendering say, OverviewPage before serverConfig was returned from the credentials check).   I think env.js solves that issue.  I agree with @israel-hdez that it does introduce a stronger tie to the server, and it's not something a typical client would be able to leverage, but I am OK with it because we know our deployment strategy and the UI will be deployed with the server.

So what we are talking about here is an additional call at login to grab the server config.  I think it's worth doing to ensure we capture any updated change in prometheus (however unlikely).

I'll leave it up to you as to how you want to proceed.  To move things along I will approve this PR now but please resolve the comment above about the duplicate logging.  I will also  approve the UI PR but I encourage you to revert the LoginThunkActions code and restore calls to getServerConfig.  Although, I am not requiring it.

jotak:
@jshaughn so what is the alternative if we don't go with env.js (because it's tying client too much to the server) ? I can try with a memoized promise approach, but in the end it's going to be close to the current redux approach and i'm not sure if it's better or worse.

jshaughn:
@jotak, to remove the persisted redux storage of serverConfig we need another way that assures us of always having *some* serverConfig available.  The specific problem use-case is F5, because we start rendering before the credentials check is complete. (Note that prior to this PR the credentials check was performed via a call to get the serverConfig, thus validating the token).  So, the non-redux options I can see are:
1. env.js
- pros: it works and it's not too complex, no calls needed for serverConfig (on login/refresh)
- cons: it may not pick up a change (changes are rare), ties server-ui, a bit of work on server to supply
2. static default serverConfig in the UI.
- pros: super easy, just return default values until serverConfig is set properly set
- cons: may not reflect reality, requires calls for serverConfig (on login/refresh), unless redux is somehow involved pages won't re-render automatically when serverConfig is updated
3. Prevent rendering until serverConfig request completes
- pros: clean, guarantees we use the proper serverConfig when rendering
- cons: requires some code changes in navigation

These are the non-redux approaches that I can see,  To be honest, option 3 is probably the best, if we can do it.  It's what I should have tried to do in the first place.  Maybe it's easy.  By the time I had the redux approach working I had taken so much time I didn't want to spend more time. Ironically, Now we are both spending time, so it was a failure...

jshaughn:
@jotak OK, so after talking about this with Edgar we decided to at least try option 3 above, which would be to "do the right thing" and not render until the user is properly authorized.  This also plays into the recent work for OAuth support, and also moving auth token handling to cookies.  So, I hope you don't mind but for now I am going to change these PR's to DNM until we have some results from Edgar with the option 3 approach.  I will create another Jira for him to use.  What do you think?

jotak:
@jshaughn +1 for trying option 3, I hope it's not going to be too complicated :)

jotak:
closing in favor of KIALI-2479

PR #864, opened 2019-02-21
KIALI-2434 Check the portname form

burmanm:
** Describe the change **

Adds validation to the portname for gateways as it must match the Istio rules for the routing to work correctly

https://istio.io/docs/setup/kubernetes/spec-requirements/

** Issue reference **

KIALI-2434

### DIFF
--- a/business/checkers/gateway_checker.go
+++ b/business/checkers/gateway_checker.go
@@ -6,12 +6,46 @@ import (
 	"github.com/kiali/kiali/models"
 )
 
+const GatewayCheckerType = "gateway"
+
 type GatewayChecker struct {
 	GatewaysPerNamespace [][]kubernetes.IstioObject
+	Namespace            string
 }
 
+// Check runs checks for the all namespaces actions as well as for the single namespace validations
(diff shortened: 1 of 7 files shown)


business/checkers/gateway_checker.go
```diff
@@ -6,12 +6,46 @@ import (
...
+	// Single namespace
+	for _, nssGw := range g.GatewaysPerNamespace {
+		for _, gw := range nssGw {
+			if gw.GetObjectMeta().Namespace == g.Namespace {
+				validations.MergeValidations(runSingleChecks(gw))
+			}
+		}
+	}
+
+	return validations
```
xeviknal: Wdyt about following the same pattern as in the `virtual_service_checkers.go`?
Perhaps re-using that `runChecks` and adapting to make it broader?

burmanm: Yeah, although I guess as a broader refactoring (for all different types - since we have more types coming in any case and perhaps multitenancy). Current one won't directly work here as the gateway validators get gateways from all the namespaces instead of a single namespace / single object like the VS validator.

PR #611, opened 2018-10-31
update the hack script so it can install Istio 1.1.

jmazzitelli:

PR #880, opened 2019-02-28
KIALI-2473 Use Spec.Replicas instead of Status.Replicas

lucasponce:
It seems Status.Replicas may not be equal to Spec.Replicas which is the "desired" replicas field.

### DIFF
--- a/models/workload.go
+++ b/models/workload.go
@@ -137,7 +137,9 @@ func (workload *Workload) ParseDeployment(d *v1beta1.Deployment) {
 
 	workload.CreatedAt = formatTime(d.CreationTimestamp.Time)
 	workload.ResourceVersion = d.ResourceVersion
-	workload.Replicas = d.Status.Replicas
+	if d.Spec.Replicas != nil {
+		workload.Replicas = *d.Spec.Replicas
+	}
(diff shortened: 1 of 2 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#611](https://paste.fedoraproject.org/paste/F6NEJthyvZWqlHnG7Ru-SQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1321]](https://paste.fedoraproject.org/paste/C~H-VJJRA8V5xg2-eTdGZQ)

lucasponce:
Yes, I had same thought when we worked on that feature.
My understanding is that Status.Replicas is pure runtime info, so all Pods in non-terminated state; and Spec.Replicas is the "desired" value added by the user.
In most cases that value is the same, but I think it might happen that for some reason there is a gap.
I.e. user modify the Deployment spec replica, but for some issue, that value is not propagated into the child controller (ReplicaSet) and Pod is not even started.

It sounds a corner case, but just to map the real "desired" value versus the pod counter.

I also find difficult to build an example to test this in a real scenario, my guess is that the controller will always initiate the Pod.

PR #616, opened 2018-11-01
hack script to build upstream istio binaries, including istioctl

jmazzitelli:
This is a hack script that you can use to build Istio. I want to build images for the istio release-1.1 branch, so this script came out of that need. Run with `--help` for usage. But in short, you can do this:

`build-istio.sh -c -b release-1.1 -y -d`

This should build the binaries, build the yaml, and install the images in your local docker.

Unfortunately, I can't get the go tests to pass. So I have not seen this run fully. Would like someone to try to verify this. Just grab the script, put it in your hack/istio directory, and run it with the command I show above and see what you get.

PR #881, opened 2019-02-28
should not be using hardcoded strings - use constants

jmazzitelli:
I was looking in this code and noticed we are using hardcoded strings throughout for the strategy names - these should be constant variables so we don't screw it up later with typos


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#616](https://paste.fedoraproject.org/paste/rJ-CDRgSvIeldq0rPUllBQ)

PR #620, opened 2018-11-02
KIALI-1823 Reconciling Kiali Secrets with Istio Helm Charts Secrets

gbaufake:
** Describe the change **
When Kiali was incorporated on Istio, Istio Engineers asked us to change our secrets from password to passphrase (https://github.com/istio/istio/pull/7967/files#r210915209).

That was done on this PR (https://github.com/istio/istio/pull/7326), but it led to a mismatch between our secrets on helm and ansible installer and on the repository, because kiali stills uses password on the master

** Issue reference **

`KIALI-1823`

cc @jmazzitelli

PR #882, opened 2019-02-28
KIALI-1186 Use oauth token when contacting the k8s master

mwringe:
**Describe the change**

If we are using the Oauth login mechanism for OpenShift, we will now use this token when contacting the k8s master. This means the permissions the user has in the OpenShift cluster will affect what they can access in Kiali.

If a user is not using the Oauth login mechanism, then we use the service accounts token when contacting the k8s master.

This basically brings us up to the 'mulitenancy' support that we want to have with OpenShift.

**Issue reference**

https://issues.jboss.org/browse/KIALI-1186

**Documentation**

We do need documentation to be updated. Will submit a PR for docs once this PR has been reviewed.

**Known Issues**

There are a few issues I have seen so far:

1) if you are in the graph seeing one namespace and then later login with a user that doesn't have access to the namespace you are shown an error.
2) if your user doesn't have access to the kiali-system namespace you will get an error about not being able to generate the grafana links.

These are also issues that we have if you were running with reduced permissions. I would rather fix those in a separate jira and keep this one more focused on the task we are trying to do here.

User who are logging into Kiali may not have all the permissions that the `kiali` role grants, so we might experience some issues with access denied problems (and will need to work towards making them work better). The easier way to get around this is to grant the user the `kiali` role for that namespace (`oc adm policy add-role-to-user kiali ${USER} -n ${NAMESPACE}`)

**Other Considerations**
We should also see how this affects performance. Before we were reusing the k8s client across all connections and now we are creating new ones per user requests. Its unknown at this point how much of a performace hit this will cause.

**How to test**
1) build Kiali like you would if you were using oauth authentication 
ie `AUTH_STRATEGY=openshift make openshift-deploy`

2) log into openshift using a user account. You should only be displayed data that you have access to. If you want to grant a user access to another namespace, you can grant them the `kiali` role:
`oc adm policy add-role-to-user kiali ${USER} -n ${NAMESPACE}`

### DIFF
--- a/business/layer.go
+++ b/business/layer.go
@@ -18,33 +18,48 @@ type Layer struct {
 	OpenshiftOAuth OpenshiftOAuthService
 }
 
-// Global business.Layer; currently only used for tests to inject mocks,
-//	whereas production code recreates services in a stateless way
-var layer *Layer
+// Global clientfactory and prometheus clients.
+var clientFactory kubernetes.ClientFactory
+var prometheusClient prometheus.ClientInterface
 
-// Get the business.Layer, create it if necessary
-func Get() (*Layer, error) {
-	if layer != nil {
-		return layer, nil
-	}
-	k8s, err := kubernetes.NewClient()
-	if err != nil {
-		return nil, err
-	}
-	prom, err := prometheus.NewClient()
-	if err != nil {
-		return nil, err
-	}
-
-	// Business needs to maintain a minimal state as kubernetes package will maintain a cache
-	SetWithBackends(k8s, prom)
-	return layer, nil
+func GetUnauthenticated() (*Layer, error) {
+	return Get("")
 }
 
-// SetWithBackends creates all services with injected clients to external APIs
-func SetWithBackends(k8s kubernetes.IstioClientInterface, prom prometheus.ClientInterface) *Layer {
-	layer = NewWithBackends(k8s, prom)
-	return layer
+// Get the business.Layer
+func Get(token string) (*Layer, error) {
+	// Use an existing client factory if it exists, otherwise create and use in the future
+	if clientFactory == nil {
+		userClient, err := kubernetes.NewClientFactory()
+		if err != nil {
+			return nil, err
+		}
+		clientFactory = userClient
+	}
+
+	// Creates a new k8s client based on the current users token
+	k8s, err := clientFactory.NewClient(token)
+	if err != nil {
+		return nil, err
+	}
+
+	// Use an existing Prometheus client if it exists, otherwise create and use in the future
+	if prometheusClient == nil {
+		prom, err := prometheus.NewClient()
+		if err != nil {
+			return nil, err
+		}
+		prometheusClient = prom
+	}
+
+	return NewWithBackends(k8s, prometheusClient), nil
+}
+
+// SetWithBackends allows for specifying the ClientFactory and Prometheus clients to be used.
+// Mock friendly. Used only with tests.
+func SetWithBackends(cf kubernetes.ClientFactory, prom prometheus.ClientInterface) {
+	clientFactory = cf
+	prometheusClient = prom
 }
 
 // NewWithBackends creates the business layer using the passed k8s and prom clients
--- a/graph/appender/dead_node.go
+++ b/graph/appender/dead_node.go
@@ -1,7 +1,6 @@
 package appender
 
 import (
-	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/graph"
 )
 
@@ -27,11 +26,6 @@ func (a DeadNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *G
 		return
 	}
 
-	var err error
-	if globalInfo.Business == nil {
-		globalInfo.Business, err = business.Get()
-		graph.CheckError(err)
-	}
 	if namespaceInfo.WorkloadList == nil {
 		workloadList, err := globalInfo.Business.Workload.GetWorkloadList(namespaceInfo.Namespace)
 		graph.CheckError(err)
--- a/graph/options/options.go
+++ b/graph/options/options.go
@@ -146,7 +146,20 @@ func NewOptions(r *http.Request) Options {
 
 	// Process namespaces options:
 	namespaceMap := make(map[string]graph.NamespaceInfo)
-	accessibleNamespaces := getAccessibleNamespaces()
+
+	tokenContext := r.Context().Value("token")
+	var token string
+	if tokenContext != nil {
+		if tokenString, ok := tokenContext.(string); !ok {
+			graph.Error("token is not of type string")
+		} else {
+			token = tokenString
+		}
+	} else {
+		graph.Error("token missing in request context")
+	}
+
+	accessibleNamespaces := getAccessibleNamespaces(token)
 
 	// If path variable is set then it is the only relevant namespace (it's a node graph)
 	// Else if namespaces query param is set it specifies the relevant namespaces
@@ -314,9 +327,9 @@ func parseAppenders(params url.Values, o Options) []appender.Appender {
 // The Set is implemented using the map convention. Each map entry is set to the
 // creation timestamp of the namespace, to be used to ensure valid time ranges for
 // queries against the namespace.
-func getAccessibleNamespaces() map[string]time.Time {
+func getAccessibleNamespaces(token string) map[string]time.Time {
 	// Get the namespaces
-	business, err := business.Get()
+	business, err := business.Get(token)
 	graph.CheckError(err)
 
 	namespaces, err := business.Namespace.GetNamespaces()
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -1,7 +1,9 @@
 package handlers
 
 import (
+	"context"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strconv"
 	"strings"
@@ -16,6 +18,10 @@ import (
 const (
 	missingSecretStatusCode = 520
 )
+
+type AuthenticationHandler struct {
+	saToken string
+}
 
 type AuthInfo struct {
 	Strategy              string      `json:"strategy"`
@@ -144,7 +150,7 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
 
 	expiresOn := time.Now().Add(time.Second * time.Duration(expiresInNumber))
 
-	business, err := business.Get()
+	business, err := getBusiness(r)
 	if err != nil {
 		RespondWithJSONIndent(w, http.StatusInternalServerError, "Error retrieving the OAuth package.")
 	}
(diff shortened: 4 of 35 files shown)


pilhuhn:
💯for the very detailed description - great job 🎉

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#620](https://paste.fedoraproject.org/paste/UGU947c1Xl3zHiCUGaEXNg)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1328]](https://paste.fedoraproject.org/paste/EaTSMAn5jeCRnAo2IhJZ4w)

israel-hdez:
> There are a few issues I have seen so far:

Perhaps as a follow-up improvement, we could create a kiali role with the very minimum privileges that a user needs so that Kiali works OK. I think we could also use that role to give/restrict per-user access to Kiali because, for now, any user with an account in the cluster will be able to login to Kiali. It depends on how much we want to control access.

mwringe:
> Perhaps as a follow-up improvement, we could create a kiali role with the very minimum privileges that a user needs so that Kiali works OK.

There is a jira to look into a 'read only' role for Kiali (https://issues.jboss.org/browse/KIALI-2325)

There is also a jira to look and see if we can minimise the permissions in the permissions required for the existing kiali role (https://issues.jboss.org/browse/KIALI-2324)

> I think we could also use that role to give/restrict per-user access to Kiali because, for now, any user with an account in the cluster will be able to login to Kiali.

There is https://issues.jboss.org/browse/KIALI-2326 which would allow us to support 'soft multitenancy'. Essentially the Kiali service account would be granted permission to certain namespaces and a user logging into Kiali would only be able to see those namespaces.

business/layer.go
```diff
@@ -18,33 +18,48 @@ type Layer struct {
...
+	// Creates a new k8s client based on the current users token
+	k8s, err := clientFactory.NewClient(token)
 	if err != nil {
 		return nil, err
 	}
-	prom, err := prometheus.NewClient()
-	if err != nil {
-		return nil, err
+
+	// Use an exisiting Prometheus client if it exists, otherwise create and use in the future
```
lucasponce: Typo: "exisiting" > "existing"

graph/appender/dead_node.go
```diff
@@ -27,11 +26,6 @@ func (a DeadNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *G
 		return
 	}
 
-	var err error
```
lucasponce: Is this safe ?
I get that the initialization of globalInfo.Business happens in a different area and this "nil" condition won't happen again ?

graph/appender/dead_node.go
```diff
@@ -27,11 +26,6 @@ func (a DeadNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *G
...
-	if globalInfo.Business == nil {
-		globalInfo.Business, err = business.Get()
-		graph.CheckError(err)
-	}
 	if namespaceInfo.WorkloadList == nil {
 		workloadList, err := globalInfo.Business.Workload.GetWorkloadList(namespaceInfo.Namespace)
```
lucasponce: If this condition happens, globalInfo.Business will be properly initialized ?

handlers/authentication.go
```diff
@@ -9,30 +11,60 @@ import (
...
+	})
+}
+
+func (ahandler *AuthenticationHandler) Handle(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		statusCode := http.StatusOK
 		conf := config.Get()
 
+		//set the token to be the SA token. If we are using OpenShift OAuth then we will use the OAuth token instead
+		token := "" // ahandler.saToken;
```
jmazzitelli: The comment says this code should be setting the token to the SA token, however that part is commented out and token is being set to an empty string "".

graph/options/options.go
```diff
@@ -146,7 +146,16 @@ func NewOptions(r *http.Request) Options {
 
 	// Process namespaces options:
 	namespaceMap := make(map[string]graph.NamespaceInfo)
-	accessibleNamespaces := getAccessibleNamespaces()
+
+	tokenContext := r.Context().Value("token")
+	var token string
+	if tokenContext != nil {
+		token = tokenContext.(string)
```
lucasponce: Can this fail ? Better to protect it with a `[...] token, ok := tokenContext.(string); ok  [...]`

handlers/authentication.go
```diff
@@ -9,30 +11,60 @@ import (
...
+
 		switch conf.Auth.Strategy {
 		case config.AuthStrategyOpenshift:
-			business, err := business.Get()
+			// Use the SA token to validate the oauth token
+			business, err := business.Get(ahandler.saToken)
```
lucasponce: If the saToken is something fixed.
I think the client inside could be cached.
Unless I missed we create on every request two clients:
[1] One with the sa to validate oath.
[2] User client with it's token once validated.

The [1] won't change on every request.

business/layer.go
```diff
@@ -18,33 +18,48 @@ type Layer struct {
...
+	if prometheusClient == nil {
+		prom, err := prometheus.NewClient()
+		if err != nil {
+			return nil, err
+		}
+		prometheusClient = prom
```
lucasponce: Good cache.

business/layer.go
```diff
@@ -18,33 +18,48 @@ type Layer struct {
...
+		userClient, err := kubernetes.NewClientFactory()
+		if err != nil {
+			return nil, err
+		}
+		clientFactory = userClient
 	}
-	k8s, err := kubernetes.NewClient()
+
+	// Creates a new k8s client based on the current users token
+	k8s, err := clientFactory.NewClient(token)
```
lucasponce: I wonder if we can maintain a map with token -> client with some time to expire (configurable).
That would be easy to implement and it would help to reduce the client operation at cost to maintain some memory increasy.
It is expected that same token will be re-used multiple times so that probably could make sense.

handlers/graph.go
```diff
@@ -69,14 +69,21 @@ func graphNamespaces(w http.ResponseWriter, r *http.Request, client *prometheus.
 	promtimer := internalmetrics.GetGraphGenerationTimePrometheusTimer(o.GetGraphKind(), o.GraphType, o.InjectServiceNodes)
 	defer promtimer.ObserveDuration()
 
-	trafficMap := buildNamespacesTrafficMap(o, client)
+	// Create a 'global' object to store the business. Global only to the request.
+	globalInfo := appender.NewGlobalInfo()
+
+	business, err := getBusiness(r)
+	graph.CheckError(err)
+	globalInfo.Business = business
```
lucasponce: Ok, this responds one of my previous questions.

business/layer.go
mwringe: Yeah, we could do this.

I wonder if we want to do this purely based on expiration, or do something like store X many clients and when we go over X we kick out the one that was used last (so we known a bound on the memory size).

handlers/root.go
```diff
@@ -18,7 +18,14 @@ func Root(w http.ResponseWriter, r *http.Request) {
...
 
 func getStatus(w http.ResponseWriter, r *http.Request) {
-	RespondWithJSONIndent(w, http.StatusOK, status.Get())
+
+	tokenContext := r.Context().Value("token")
+	if tokenContext != nil {
+		token := tokenContext.(string)
+		RespondWithJSONIndent(w, http.StatusOK, status.Get(token))
+	} else {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, "Token missing in request context")
```
lucasponce: Note that "Root" handlers is not authenticated:

```
		// swagger:route GET / Root
		// ---
		// Endpoint to get the status of Kiali
		//
		//     Produces:
		//     - application/json
		//
		//     Schemes: http, https
		// responses:
		//      500: internalError
		//      200: statusInfo
		{
			"Root",
			"GET",
			"/api",
			handlers.Root,
			false,
		},
```

So, that needs to be updated in the doc or let not authenticated scenarios.
Not sure if it's valid a token required in all cases.

mwringe: It has an empty string token in this case (which is an unauthenticated user). So the token value is set here (its set for all the endpoints that don't require authentication, see handlers.authentication HandleUnauthenticated)

handlers/utils.go
```diff
@@ -14,15 +15,21 @@ import (
...
-	return kubernetes.NewClient()
-}
 
-func getService(namespace string, service string) (*v1.ServiceSpec, error) {
-	client, err := kubernetes.NewClient()
+var clientFactory kubernetes.ClientFactory
+
+func getService(token string, namespace string, service string) (*v1.ServiceSpec, error) {
+	if clientFactory == nil {
+		userClient, err := kubernetes.NewClientFactory()
```
lucasponce: Minor: userClient -> userClientFactory looks more clear to me

handlers/utils.go
```diff
@@ -48,23 +59,37 @@ func checkNamespaceAccess(w http.ResponseWriter, k8s kubernetes.IstioClientInter
...
 	}
-	return prom, k8s, nsInfo
+	return prom, nsInfo
+}
+
+// getToken retrieves the token from the request's context
+func getToken(r *http.Request) (string, error) {
+	tokenContext := r.Context().Value("token")
+	if tokenContext != nil {
+		return tokenContext.(string), nil
```
lucasponce: Cast operations are better to be proctected.

lucasponce:
Adding @hhovsepy to verify potential regressions.

business/layer.go
lucasponce: I simple first version could be ok, with some configuration for some tweaks.
But I wouldn't add too much effort yet on it.
A map with last accessed and a clean/purge if last access was > threshold could work.
That could give us also hints if we can apply k8s caches on it or if a next effort is not worth.

lucasponce:
Some comments from my side that perhaps are worth to address, specially related to cache clients.
Tested locally and things look good.
@hhovsepy will run more specific tests looking regressions but in general it's good.

mwringe:
Ok, this has been updated to reflect the changes and rebased onto the latest from master.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#648](https://paste.fedoraproject.org/paste/AodQy8~GCuYRubnHhbGGSw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1374]](https://paste.fedoraproject.org/paste/u6Xstjp5EJYXDDMKsN60Aw)

lucasponce:
@hhovsepy would you mind to give a final check ?

I can also review the code locally.

hhovsepy:
UI Regression testing pass from Jupiter pov.

lucasponce:
I think that was commented before
image
But just to confirm that probably we need a follow-up JIRA to address this.

mwringe:
Note I have a PR for the Grafana issue, but I am waiting for this to get merged first so that I can send out a separate PR.

lucasponce:
@jmazzitelli last comment about this one ? I'm +1 to merge to not block @mwringe.
Just let me know or update the PR.
Thanks

Issue #652, opened 2018-11-12
Support auto login using url query params


jeremyxu2010:
**Describe the bug**
I want embed kiali's web page into my system. So i need auto login using url query params. eg. i use 
`http://kiali.k8s.local/console/graph/namespaces?layout=dagre&duration=60&edges=hide&graphType=versionedApp&injectServiceNodes=false&username=admin&password=admin` link to view graph page directly(skip login page).

For UI issues please also add a screenshot that shows the issue.

**Versions used**
Kiali: v0.9
Istio: 1.0.3
Kubernetes flavour and version: 1.10

pilhuhn:
As discussed in https://github.com/kiali/kiali-ui/pull/813 this is too dangerous. So I am closing the request.

jmazzitelli:
See https://github.com/kiali/kiali-ui/pull/830 - I believe that will do what you want.

PR #887, opened 2019-03-04
KIALI-2482 Improve kiali install script with dashboards

jotak:
- Moved all dashboard files into a big 'all.yaml', easing the installation
- Change warning => note in message displayed in case the dashboards couldn't be installed


jotak:
@jmazzitelli I think this new "error" message is much less stressful now, when kiali is installed without dashboards

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#634](https://paste.fedoraproject.org/paste/HJ~zp9-6wUeFdImk30DMMQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1356]](https://paste.fedoraproject.org/paste/jHqsGKLtImro8anzkhlDvA)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#635](https://paste.fedoraproject.org/paste/DW2r2NdzhAwdKCA0e2ZdTA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1357]](https://paste.fedoraproject.org/paste/1mKecZmUH-3tjiv86vUx3g)

pilhuhn:
Do you think using the all.yaml will scale? Of course it is easier to curl for it, then for many files. 
I could imagine using something like `oc apply dashboards/` could be good (?)

jotak:
@pilhuhn this is what I was doing initially but it doesn't work when downloading files from the repo, so I switched to explicit file names.

I think at the moment it's fine as we don't plan to host tons and tons of dashboards (or then, if we do so, maybe we would have a dedicated repository "a la Grafana"?)

Or another alternative could be to keep individual files and have maybe a pre-commit script that generates all.yaml .. but i would say it's not necessary at the moment.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#652](https://paste.fedoraproject.org/paste/ISZy3x1U-aRBvqS3mYqBfw)

pilhuhn:
I know this ship has sailed, but I wonder if we should (partially) go back to the previous state and listing the available dashboards as individual files.
That would allow users to browse them easily and to more easily copy them for their own purposes.
And then when we create a release, we put them all together into one big file, like all.yaml and make them available on the GH release page, so that there is a single URL to download and kubectl apply -f it

jotak:
@israel-hdez @jmazzitelli regarding @pilhuhn 's comment above, what would be the right place to generate a file that must be part of the releases? in `deploy/jenkins-ci/Makefile` ?

israel-hdez:
@jotak @pilhuhn I think it depends on what is defined as a "release".
Right now, our main artifact for a release is the image in DockerHub. We also have the released UI in NPM and the tagged code in GitHub.

Probably the place to put the processed yaml would be the tagged code in GitHub? If so, then yes the `deploy/jenkins-ci/Makefile` file sound like the right place to generate it. Although I don't like too much the idea of creating a GitHub tag for "processed" source code. To me it sounds like a motivation for building one huge yaml file that integrates all yaml files in the `deploy/{platform}/` dirs and publish it as an asset in the github release (similar to how istio generates built assets in https://github.com/istio/istio/releases).

PR #889, opened 2019-03-04
KIALI-2457 Tidying up validations - Failed - more golang background needed

xeviknal:
** Describe the change **

Such a failure. I still have in my mind the inheritance mechanism from Ruby and Java.
I tried to specify some common methods on the parents, some other methods on the children and in some other cases, override some methods in children that are already specified on parent.

The thing is, GoLang doesn't offer inheritance itself, it just let embed structs. So the embedded struct doesn't know anything about the struct embedding it. Then, when a child call a method defined in the parent and this last one calls to another method overrode in the child, the call goes to the parent instead of the child.

This makes that the embedded structs are more likely self-contained. Methods defined in that embedded method can't rely on the embeddee struct, so they can only use data/methods specified in its struct.

### DIFF
--- a/business/checkers/checker.go
+++ b/business/checkers/checker.go
@@ -8,7 +8,7 @@ import (
 type ObjectChecker interface {
 	GetType() string
 	GetGroupCheckers() []GroupChecker
-	GetIndividualCheckers() []IndividualChecker
+	GetIndividualCheckers(object kubernetes.IstioObject) []IndividualChecker
 	GetObjects() []kubernetes.IstioObject
 	Check() models.IstioValidations
 }
--- a/business/checkers/checker.go
+++ b/business/checkers/checker.go
@@ -1,11 +1,87 @@
 package checkers
 
-import "github.com/kiali/kiali/models"
+import (
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/models"
+)
 
-type Checker interface {
+type IndividualChecker interface {
 	Check() ([]*models.IstioCheck, bool)
 }
 
 type GroupChecker interface {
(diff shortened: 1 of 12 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#638](https://paste.fedoraproject.org/paste/5cOkDh8JLG~hVcpjTo9IIA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1360]](https://paste.fedoraproject.org/paste/flmRJJHbO-0ehtwayC0rVg)

business/checkers/checker.go
```diff
@@ -1,11 +1,95 @@
...
+
+type IndividualChecker interface {
 	Check() ([]*models.IstioCheck, bool)
 }
 
 type GroupChecker interface {
 	Check() models.IstioValidations
 }
+
+type ObjectCheck struct{}
```
jotak: I guess you could remove all these Getters (GetType, GetGroupCheckers etc.) and have the corresponding stuff declared as struct fields instead. They would be provided at construction, for instance:

```
func NewDestinationRulesChecker() {
  return DestinationRulesChecker{
    Type: "destinationrule",
    // ...
  }
}
```

(you can pass lambdas as well)

Forget OOP ;-)

xeviknal: It seems that I haven't learnt that lesson yet, hahaha

israel-hdez: @xeviknal You are not alone. It's also hard for me to stop thinking in a OOP way.
Then I read this: https://medium.com/@cscalfani/goodbye-object-oriented-programming-a59cda4c0e53 haha I've had almost all problems described in that article.

PR #908, opened 2019-03-11
KIALI-2506 Cache K8S clients between requests

mwringe:
** Describe the change **

Instead of creating a new k8s client on each request, we store in memory the clients. If a client is not used in 15 minutes, it is deleted. This is done for performance reasons.

A new metric ('kiali_kubernetes_clients') tracks how many clients are in use at any given time.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2506

### DIFF
--- a/kubernetes/client_factory.go
+++ b/kubernetes/client_factory.go
@@ -1,0 +1,137 @@
+package kubernetes
+
+import (
+	"sync"
+	"time"
+
+	"k8s.io/client-go/rest"
+
+	"github.com/kiali/kiali/log"
+	"github.com/kiali/kiali/prometheus/internalmetrics"
+)
+
+var factory *clientFactory
+
+// Mutex for when modifying the stored clients
+var mutex = &sync.RWMutex{}
+
+const expirationTime = time.Minute * 15
+
+// ClientFactory interface for the clientFactory object
+type ClientFactory interface {
+	GetClient(token string) (IstioClientInterface, error)
+}
+
+// clientFactory used to generate per users clients
+type clientFactory struct {
+	ClientFactory
+	baseIstioConfig *rest.Config
+	clientEntries   map[string]*clientEntry
+}
+
+// clientEntry stored the client and its created timestamp
+type clientEntry struct {
+	client  IstioClientInterface
+	created time.Time
+}
+
+// GetClientFactory returns the client factory. Creates a new one if necessary
+func GetClientFactory() (ClientFactory, error) {
+	if factory == nil {
+		// Get the normal configuration
+		config, err := ConfigClient()
+		if err != nil {
+			return nil, err
+		}
+
+		// Create a new config based on what was gathered above but don't specify the bearer token to use
+		istioConfig := rest.Config{
+			Host:            config.Host,
+			TLSClientConfig: config.TLSClientConfig,
+			QPS:             config.QPS,
+			Burst:           config.Burst,
+		}
+
+		return getClientFactory(&istioConfig, expirationTime)
+
+	}
+	return factory, nil
+}
+
+// newClientFactory allows for specifying the config and expiry duration
+// Mock friendly for testing purposes
+func getClientFactory(istioConfig *rest.Config, expiry time.Duration) (*clientFactory, error) {
+	mutex.Lock()
+	if factory == nil {
+		clientEntriesMap := make(map[string]*clientEntry)
+
+		factory = &clientFactory{
+			baseIstioConfig: istioConfig,
+			clientEntries:   clientEntriesMap,
+		}
+
+		go watchClients(clientEntriesMap, expiry)
+	}
+	mutex.Unlock()
+	return factory, nil
+}
+
+// NewClient creates a new IstioClientInterface based on a users k8s token
+func (cf *clientFactory) newClient(token string) (IstioClientInterface, error) {
+	config := cf.baseIstioConfig
+
+	config.BearerToken = token
+
+	return NewClientFromConfig(config)
+}
+
+// GetClient returns a client for the specified token. Creating one if necessary.
+func (cf *clientFactory) GetClient(token string) (IstioClientInterface, error) {
+	clientEntry, err := cf.getClientEntry(token)
+	if err != nil {
+		return nil, err
+	}
+	return clientEntry.client, nil
+}
+
+// getClientEntry returns a clientEntry for the specified token. Creating one if necessary.
+func (cf *clientFactory) getClientEntry(token string) (*clientEntry, error) {
+	mutex.RLock()
+	cEntry, ok := cf.clientEntries[token]
+	mutex.RUnlock()
+	if ok {
+		return cEntry, nil
+	} else {
+		client, err := cf.newClient(token)
+		if err != nil {
+			log.Errorf("Error fetching the Kubernetes client: %v", err)
+			return nil, err
+		}
+
+		cEntry := clientEntry{
+			client:  client,
+			created: time.Now(),
+		}
+
+		mutex.Lock()
+		cf.clientEntries[token] = &cEntry
+		mutex.Unlock()
+		internalmetrics.SetKubernetesClients(len(cf.clientEntries))
+		return &cEntry, nil
+	}
+}
+
+// watchClients loops over clients and removes ones which are too old
+func watchClients(clientEntries map[string]*clientEntry, expiry time.Duration) {
+	for {
+		time.Sleep(expiry)
+		mutex.Lock()
+		for token, clientEntry := range clientEntries {
+			if time.Since(clientEntry.created) > expiry {
+				delete(clientEntries, token)
+			}
+		}
+		internalmetrics.SetKubernetesClients(len(clientEntries))
+		mutex.Unlock()
+	}
+}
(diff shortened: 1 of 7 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#675](https://paste.fedoraproject.org/paste/ep6SWtMDWiJNySLBzrKkGA)
  * :red_circle: run-kiali-e2e-tests [#[1412]](https://paste.fedoraproject.org/paste/J6NcIKXzxyMGIq8TWKChIg)

lucasponce:
The CI is failing; I'm fetching this one locally to see if I can reproduce same issues there.

lucasponce:
I'm having this locally
```
[lponce@recopolis kiali]$ make test
Running tests, excluding third party tests under vendor
go test github.com/kiali/kiali github.com/kiali/kiali/business github.com/kiali/kiali/business/checkers github.com/kiali/kiali/business/checkers/destinationrules github.com/kiali/kiali/business/checkers/gateways github.com/kiali/kiali/business/checkers/meshpolicies github.com/kiali/kiali/business/checkers/virtual_services github.com/kiali/kiali/config github.com/kiali/kiali/config/security github.com/kiali/kiali/graph github.com/kiali/kiali/graph/appender github.com/kiali/kiali/graph/cytoscape github.com/kiali/kiali/graph/options github.com/kiali/kiali/handlers github.com/kiali/kiali/kubernetes github.com/kiali/kiali/kubernetes/kubetest github.com/kiali/kiali/log github.com/kiali/kiali/models github.com/kiali/kiali/prometheus github.com/kiali/kiali/prometheus/internalmetrics github.com/kiali/kiali/prometheus/prometheustest github.com/kiali/kiali/routing github.com/kiali/kiali/server github.com/kiali/kiali/status github.com/kiali/kiali/tests/data github.com/kiali/kiali/util github.com/kiali/kiali/util/intutil
# github.com/kiali/kiali/server
server/metrics_server.go:18: Info call has possible formatting directive %v
?   	github.com/kiali/kiali	[no test files]
ok  	github.com/kiali/kiali/business	0.104s
ok  	github.com/kiali/kiali/business/checkers	0.043s
ok  	github.com/kiali/kiali/business/checkers/destinationrules	0.050s
ok  	github.com/kiali/kiali/business/checkers/gateways	0.047s
ok  	github.com/kiali/kiali/business/checkers/meshpolicies	0.022s
ok  	github.com/kiali/kiali/business/checkers/virtual_services	0.028s
ok  	github.com/kiali/kiali/config	0.061s
ok  	github.com/kiali/kiali/config/security	0.001s
?   	github.com/kiali/kiali/graph	[no test files]
ok  	github.com/kiali/kiali/graph/appender	0.030s
?   	github.com/kiali/kiali/graph/cytoscape	[no test files]
?   	github.com/kiali/kiali/graph/options	[no test files]
ok  	github.com/kiali/kiali/handlers	0.065s
ok  	github.com/kiali/kiali/kubernetes	0.510s
?   	github.com/kiali/kiali/kubernetes/kubetest	[no test files]
?   	github.com/kiali/kiali/log	[no test files]
ok  	github.com/kiali/kiali/models	0.027s
?   	github.com/kiali/kiali/prometheus	[no test files]
ok  	github.com/kiali/kiali/prometheus/internalmetrics	1.034s
ok  	github.com/kiali/kiali/prometheus/prometheustest	0.011s
ok  	github.com/kiali/kiali/routing	0.009s
FAIL	github.com/kiali/kiali/server [build failed]
ok  	github.com/kiali/kiali/status	0.008s
?   	github.com/kiali/kiali/tests/data	[no test files]
?   	github.com/kiali/kiali/util	[no test files]
?   	github.com/kiali/kiali/util/intutil	[no test files]
make: *** [Makefile:137: test] Error 2
```

I wonder if we've introduced some specific version syntax that is failing here.
(Need to check if master is working well)

mwringe:
I likely need to rebase to pull in https://github.com/kiali/kiali/commit/800d6760b925ce9d465d22da2ac30fb9e1a5ecba

kubernetes/client_factory.go
```diff
@@ -0,0 +1,132 @@
...
+func (cf *clientFactory) GetClient(token string) (IstioClientInterface, error) {
+	clientEntry, err := cf.getClientEntry(token)
+	if err != nil {
+		return nil, err
+	}
+	return clientEntry.client, nil
+}
+
+// getClientEntry returns a clientEntry for the specified token. Creating one if necessary.
+func (cf *clientFactory) getClientEntry(token string) (*clientEntry, error) {
```
lucasponce: I haven't tested, but under high load, we are forcing all calls to pass with the same mutex.Lock(), that sounds on paper that could be a penalty.

In the other hand, I wonder if using a default expiration could be enough, in that way, we can use a RW semaphore, and let multiple reads, only blocking when adding a new element (which can be configurable).

On this way, everytime a client is get, it's locked in the mutex.

On a expiration strategy, without marking last access time, but just the creation time, it can be more performance with the penalty to have one time re-creation of the clients.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#681](https://paste.fedoraproject.org/paste/MQ5a9wlguvuHj4UkVDDerA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1420]](https://paste.fedoraproject.org/paste/gXQ4CLLkVJ2d3LzK4NiQIA)

kubernetes/client_factory.go
mwringe: I've updated how we are handling the locking here.

Its now using a RWMutex. We only RLock when we fetch from the map (updating the timestamp doesn't need to be protected in a lock. We can have multiple threads concurrently updating this value).

We still need a lock when we create a new entry in the map and when we iterate over the map.

@lucasponce what do you think?

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#688](https://paste.fedoraproject.org/paste/hTsWXxHttMnYsh1dCqUGXQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1429]](https://paste.fedoraproject.org/paste/QMGKjjIZgpmBZKq3wUMdjA)

kubernetes/client_factory.go
lucasponce: Not sure if we can update the lastAccessed concurrently without enter in a race condition (this why the CI is failing).
And also, it could also miss the intention of your code, I guess that a check on expiration is more safer, in the worst case, a expired session but active will need to create a new client.

Maintaining the lastAccessed unprotected it can end in some undefined state (which also can lead to the same scenario) but having a race in the code.

I ask cc @simonpasquier here, as he helped us with some concurrency discussion in the past and it was good to see if something was missing in our comments.

mwringe: If we have two threads trying to concurrently set the lastAccessedTime, it doesn't really matter the order its applied. We don't need high precision here for expiration purposes.

> Maintaining the lastAccessed unprotected it can end in some undefined state (which also can lead to the same scenario) but having a race in the code.

Can it be in an undefined state? I don't believe its possible for a variable in Go to be undefined, the worst case it it would be in its default state (which is Jan 1 0000)

[Note this could actually happen in the code since we don't set the lastAccessedTime time in a mutex when we create the client, so we could have the expiration checker check the client's expiration time before its set, and since it would be year 0000 it would get marked for removal. If we want to keep the lastAccessedTIme we might want to fix this part]

There is a race condition here, but I don't believe it can actually cause a problem. Running simulations with a 10ks threads concurrently creating and deleting clients didn't have any issues.

I can remove this if we think its more safe.

@lucasponce Do we want to just remove the update to the expiration time and just expire them based on creation time? Or do you want this handled another way?

lucasponce: @mwringe I think that expire them on creation time is safe.

Probably you are right and the race condition sounds "correct" for this use case, but leaving a race condition in the code creates a weird feeling.

So, if we can make it safe I'd prefer it.

We can see in a following PR if we can implement the same scenario but without having a race in the code.

mwringe: Sounds good, I will update the PR to expire based on creation time and will remove the race conditions.

simonpasquier: Sorry for the late reply... Basically @lucasponce is right: when the race detector reports a problem, there's 99.999% changes that it is right, false positives are extremely rare. Any race conditions can generate heisenbugs even though they seem innocuous at first.

mwringe:
Rebased and updated

@lucasponce can you take another look?

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#734](https://paste.fedoraproject.org/paste/-nV8ugIXowpKizsO2ZtHSQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1495]](https://paste.fedoraproject.org/paste/3z6eQDCt2MfyDwvjgGQdxg)

lucasponce:
make test-race still failing in 
```
		mutex.Lock()
		for token, clientEntry := range clientEntries {
			if time.Since(clientEntry.created) > expiry {
				delete(clientEntries, token)
			}
		}
		internalmetrics.SetKubernetesClients(len(clientEntries))
		mutex.Unlock()
```
perhaps due `clientEntries map[string]*clientEntry` structure I mean the map is protected, but as what we store is pointer I guess that it is assuming the clientEntry.created is not protected ?

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#741](https://paste.fedoraproject.org/paste/vYf-UkYg7lY5eIsOZ7WBjA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1505]](https://paste.fedoraproject.org/paste/Uo03HBwdCHFJoktDQ8IhwQ)

PR #920, opened 2019-03-15
KIALI-2561 Ask browser to not use cache

israel-hdez:
https://issues.jboss.org/browse/KIALI-2561

Background:
- When user was already logged in, front-end was wrongly displaying the login page even if user was already logged in. This was causing a confusing experience, because any random credentials could be entered and, then, the user was being able to access the main console. Also, just refreshing with F5 caused the main console to render.
- On other less critical case, the contrary was also happening: while user was not yet logged in, front-end was displaying the main console but before any data was loaded, the UI switched immediately to the login page.

This is caused because the UI makes a request to the /api/auth/info endpoint to check if there is an active session. Browser was storing in cache the response of this endpoint on the first load of Kiali. On subsequent page loads, browser was using the cached response and the UI was determining an incorrect logged in status of the user. It could even determine an incorrect authentication strategy.

Currently, front-end is assuming that all requests hit the back-end. That was not happening always. This commit makes a change to ask the browser to not use its cache so that the front-end assumption becomes true.

Ultimately, the browser will choose. If browser keeps using cache, front-end changes will be needed.

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -278,6 +278,8 @@ func NewAuthenticationHandler() (AuthenticationHandler, error) {
 
 func (aHandler AuthenticationHandler) Handle(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
+
 		statusCode := http.StatusOK
 		conf := config.Get()
 
@@ -314,6 +316,7 @@ func (aHandler AuthenticationHandler) Handle(next http.Handler) http.Handler {
 
 func (aHandler AuthenticationHandler) HandleUnauthenticated(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
 		context := context.WithValue(r.Context(), "token", "")
 		next.ServeHTTP(w, r.WithContext(context))
 	})


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#703](https://paste.fedoraproject.org/paste/AzgbQY431lzOe68qnH7b-w)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1442]](https://paste.fedoraproject.org/paste/HEtYLZgvz7p0KepidwY09w)

routing/router.go
```diff
@@ -66,6 +67,13 @@ func NewRouter() *mux.Router {
 	return rootRouter
 }
 
+func disableBrowserCache(next http.Handler) http.Handler {
```
lucasponce: But could this affect also to the UI resources ?
I mean, if this is only to not cache the API calls, it sounds correct.
But if this could have side effect to re-fetch a JS/CSS/image resource, perhaps could have a penalty in performance.

Just thinking loud, not tested but worth to take a look.

pilhuhn: I had the same thoughts as @lucasponce

mtho11:
This is kind of the opposite of: https://issues.jboss.org/browse/KIALI-2508 - turn on caching headers for kiali ui assets

routing/router.go
israel-hdez: Yesterday I tested, and had the idea that the FileServer handler was resetting the headers, but it doesn't. Probably I got confused with the request headers in the browser's console.

I moved the code to the handlers. It should only affect the /api endpoints now.

israel-hdez:
@mtho11 Yes and KIALI-2508 seems still valid.

israel-hdez:
@jmazzitelli OK, that one is fixed in the UI. See PR: https://github.com/kiali/kiali-ui/pull/1070

The issue this back-end PR is fixing is more related to the "Duplicate tab" browser's feature. I saw the login page when Duplicating the tab and thought it was the same issue. Sorry for the confusion.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#710](https://paste.fedoraproject.org/paste/i2wnmJvk4nDrPtFyJjzlqw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1451]](https://paste.fedoraproject.org/paste/Tclws-LUqinjAlsZcFcxig)

israel-hdez:
The bug stated in the Jira looks fixed. Probably we can close this if the other behaviors described are not reproducible.

israel-hdez:
Apparently, this is not needed. Closing it.

PR #943, opened 2019-03-22
KIALI-2556 Fix node type of unknown destination

israel-hdez:
https://issues.jboss.org/browse/KIALI-2556

This prevents the UI from creating links to service pages for the "unknown" node when this node is a destination.

### DIFF
--- a/graph/graph.go
+++ b/graph/graph.go
@@ -145,7 +145,7 @@ func Id(serviceNamespace, service, workloadNamespace, workload, app, version, gr
 	// Every other field is unknown. Allow one unknown service per namespace to help reflect these
 	// bad destinations in the graph,  it may help diagnose a problem.
 	if Unknown == workload && Unknown == app && Unknown == service {
-		return fmt.Sprintf("svc_%s_unknown", namespace), NodeTypeService
+		return fmt.Sprintf("svc_%s_unknown", namespace), NodeTypeUnknown
 	}
 
 	workloadOk := IsOK(workload)


israel-hdez:
@jshaughn This seems like a simple fix, but looks like this piece of code is for a corner case.

The bug described in KIALI-2556 is for the traffic tab. But since the traffic tab consumes from the graph endpoints, graph is also affected:

image

The graph was also showing links for unknown in the side panel and this seems to remove the links. The only thing I see pending to fix in graph page is the sparklines error when selecting the edge - the error can be seen in the image.

But, well... Not sure if I'm affecting something else in a negative way. So adding you as a reviewer since you know better this part.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#742](https://paste.fedoraproject.org/paste/yPeeDNkhxI9S0Ff4zQ~kiw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1506]](https://paste.fedoraproject.org/paste/UN-0t90oKbbrWO4OpC3fBg)

graph/graph.go
```diff
@@ -145,7 +145,7 @@ func Id(serviceNamespace, service, workloadNamespace, workload, app, version, gr
 	// Every other field is unknown. Allow one unknown service per namespace to help reflect these
 	// bad destinations in the graph,  it may help diagnose a problem.
 	if Unknown == workload && Unknown == app && Unknown == service {
-		return fmt.Sprintf("svc_%s_unknown", namespace), NodeTypeService
+		return fmt.Sprintf("svc_%s_unknown", namespace), NodeTypeUnknown
```
jshaughn: No, you can't do this.  The unknown node is a traffic generator and is handled above.  This is specifically to allow one unknown destination *service*.  It means that some legitimate source generated telemetry such that the destination_service_name was "unknown".   Unfortunately, it means there does have to be special case handling to prevent links to an "unknown" service.  Perhaps you could update SummaryLink.isInaccesible() to check for the unknown service.

israel-hdez: Ok, so `NodeTypeUnknown` is reserved for unknown source. I'm wondering why it is bad to extend its definition to include unknown destination. Sounds "logically" correct.

Anyway, here is the PR fixing this issue in the UI side, rather than the back-end side: https://github.com/kiali/kiali-ui/pull/1095

I'm closing this one.

jshaughn: Yes, "unknown" is an endless source of annoyance.  I'm not sure, really.  I was mainly afraid that it would lead to other issues where we may have assumed unknown node type was a source.  And I also thought we should keep the triangle for the destination, not a diamond, which has traditionally been a source of requests.  I realize that handling it client-side is a bit messy because  we likely need to check on nodeType = service and also namespace = unknown, or something like that.

Maybe what we could do is mark it as isInaccessible in the Cytoscape generation code server-side.

israel-hdez: Marking it as inaccessible sounds quite fine. I'm more comfortable having this kind of logic in the back-end, rather than letting the UI decide. Looking at the UI code, seems like the PR will probably be thin, or not needed at all, if we mark the unknown destination as inaccessible. 

I like more that approach, so if you are OK, I can open another back-end PR to set isInaccessible for the unknown destination and review the changes of the UI.

jshaughn: Yes, of course, open another.  Lo siento por bouncing this around from server to ui and back to server.

PR #944, opened 2019-03-24
Make prometheus service url configureable

DaspawnW:
** Describe the change **
This change allows to make the prometheus service url configureable during deployment. This is interesting in case of a deployment where prometheus is not deployed with the istio helm chart or for example by using istio-operator by banzaicloud. 

** Issue reference **

* No Issue but I think it is useful.

** Backwards incompatible? **

- no

** Documentation **

* no

### DIFF
--- a/deploy/kubernetes/configmap.yaml
+++ b/deploy/kubernetes/configmap.yaml
@@ -16,6 +16,7 @@ data:
         url: ${JAEGER_URL}
       grafana:
         url: ${GRAFANA_URL}
+      prometheus_service_url: ${PROMETHEUS_SERVICE_URL}
     identity:
       cert_file: /kiali-cert/cert-chain.pem
       private_key_file: /kiali-cert/key.pem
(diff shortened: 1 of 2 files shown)


rhqci:

abonas:
@jmazzitelli @jotak

jmazzitelli:
We are soon moving away from these deploy/ files/scripts and going to an operator model. I'm putting a DNM label on this PR but I will make an analogous change to [my operator PR](https://github.com/kiali/kiali/pull/974) so this functionality goes in there. Operator will be released/published within a week or so hopefully.

jmazzitelli:
These deploy/* files have been deleted - we now use the operator. The functionality this PR wants now exists in the operator - set the prometheus.url setting in the Kiali CR to tell Kiali where prom is: https://github.com/kiali/kiali/blob/v0.18/operator/deploy/kiali/kiali_cr.yaml#L149-L156

PR #745, opened 2018-12-13
KIALI-2039 Support for patch releases through Jenkins Pipeline

israel-hdez:
https://issues.jboss.org/browse/KIALI-2039

PR #744, opened 2018-12-13
KIALI-2070 - Release Notifier - Deploy/Run tests

mattmahoneyrh:

PR #947, opened 2019-03-25
KIALI-2634 provide endpoint for readiness and liveness probes

jmazzitelli:


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#744](https://paste.fedoraproject.org/paste/oEto1yJLDr2Svtz0yuIyKQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1518]](https://paste.fedoraproject.org/paste/zsjwjSocIH934adJySuZXg)

jmazzitelli:
One problem with this now is that there is no longer a "one-stop" to change the server port and whether we want the server to be behind https or http. It used to be we just are able to change the configmap with those new config settings, restart the pod, and it would work. But now that the probes need to be told what scheme to use and what port to use, it is no longer simple to just change the server_port and the server identity to change these. You must also change the deployment yaml as well.

I think I will add the ability to set the port in the install script, at least that will allow us to configure the server_port using that, rather than having to understand you need to change the configmap AND the deployment yaml.

mwringe:
@jmazzitelli You don't have to have an endpoint for health checks, you can also have it run any executable command. If you want it to be a more one-stop change, you can just have the probes run a bash script that does something simple like a curl (and have the bashscript determine from the system if it should use http/https or what ports it should use)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#745](https://paste.fedoraproject.org/paste/PbH3YFYMOEe-MF9caCnekQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1519]](https://paste.fedoraproject.org/paste/hItBQLHd-gx02LtDszVUfg)

jmazzitelli:
OK, right now this PR is ready IF we want to use the special endpoint.
Let me now look into a different implementation that @mwringe mentioned - using a command. We'll see if its any better.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#746](https://paste.fedoraproject.org/paste/tRm7orDr4YtviprV8WCjAQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1520]](https://paste.fedoraproject.org/paste/f~pKxnF6WsrNgSTXYmT46g)

mwringe:
@jmazzitelli This PR looks good.

We don't need to use a script for the health checks, what we have here should be fine. It was just a comment that we could run a script to call the healthz endpoint if that made things easier (eg to determine values that could change like the context path, port, http/https)

jmazzitelli:
@mwringe I agree. I think we should just merge this.

PR #749, opened 2018-12-17
KIALI-1671 Refactor validations

burmanm:
Requires kiali-UI PR 892 to be merged at the same time.

This change will remove the validation endpoints and refactor them to the requested items. This requires changes to the UI part, which are not completed yet. 

Here for comments. I also changed the "istio_validations" to "validate" parameter in the REST-interface as it in my opinion fit the change better.

PR #948, opened 2019-03-25
KIALI-2644 Add in ability to specify the login type during install

mwringe:
** Describe the change **

When installing using OpenShift and there is no AUTH_STRATEGY specified it will ask if they want to use the login or OAuth option.

It will also check if the AUTH_STRATEGY is 'login' before asking for login credentials to be added (a username and password are only needed if the auth strategy is login. So there is no need to ask if we are not using 'login').

It also fixes a bug where if you run the install multiple times it will not update the Kiali installation. For instance, we could update the configmap or secret but since the deployment is not updated the Kiali pod wont get redeployed. Now we explicitly delete the Kiali deployment before doing the install.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2644

### DIFF
--- a/deploy/kubernetes/deploy-kiali-to-kubernetes.sh
+++ b/deploy/kubernetes/deploy-kiali-to-kubernetes.sh
@@ -16,6 +16,11 @@
 # To customize the behavior of this script, you can set one or more of the
 # following environment variables:
 #
+# AUTH_STRATEGY
+#    Determines what authentication strategy to use. Choose "login" to use a username and password.
+#    Choose "anonymous" to allow full access to Kiali without requiring any credentials.
+#    Default: "login"
+#
 # KIALI_USERNAME
 #    This is the username that will be required when logging into Kiali.
 #    If this is not set, or if it is set to an empty string, you will be prompted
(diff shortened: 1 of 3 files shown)


deploy/openshift/deploy-kiali-to-openshift.sh
```diff
@@ -16,6 +16,12 @@
 # To customize the behavior of this script, you can set one or more of the
 # following environment variables:
 #
+# AUTH_STRATEGY
+#    Determines what authentication strategy to use. Choose "login" to use a username and password.
+#    Choose "openshift" to use the OpenShift OAuth which presents data based on the individuals
+#    OpenShift RBAC.
+#    Default: "openshift"
+#
```
jmazzitelli: we need this same comment at the top of the k8s script.

deploy/openshift/deploy-kiali-to-openshift.sh
```diff
@@ -116,23 +122,39 @@ get_downloader() {
...
-if [ "${KIALI_USERNAME_BASE64}" == "" ]; then
-  KIALI_USERNAME="${KIALI_USERNAME=}" # note: the "=" inside ${} is on purpose
-  if [ "$KIALI_USERNAME" == "" ]; then
-    KIALI_USERNAME=$(read -p 'What do you want to use for the Kiali Username: ' val && echo -n $val)
-  fi
-  KIALI_USERNAME_BASE64="$(echo -n ${KIALI_USERNAME} | base64)"
+# Check the login strategy. If using "openshift" there is no other checks to perform,
+# but if we are using "login" then we need to make sure there is a username and password set
+if [ "${AUTH_STRATEGY}" == "" ]; then
+  AUTH_STRATEGY=$(read -p 'Choose a login strategy of either "login" or "openshift" [openshift]: ' val && echo -n $val)
```
jmazzitelli: do we not want to tell them about "anonymous"?

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#747](https://paste.fedoraproject.org/paste/08tlznTh7olstcSbcICsLA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1521]](https://paste.fedoraproject.org/paste/umg1Js9kwO0P3b1JbbVF3w)

deploy/kubernetes/deploy-kiali-to-kubernetes.sh
```diff
@@ -116,23 +116,25 @@ get_downloader() {
   fi
 }
 
-# The credentials can be specified either as already base64 encoded, or in plain text.
-# If the username or passphrase plain text variable is set but empty, the user will be asked for a value.
-if [ "${KIALI_USERNAME_BASE64}" == "" ]; then
-  KIALI_USERNAME="${KIALI_USERNAME=}" # note: the "=" inside ${} is on purpose
-  if [ "$KIALI_USERNAME" == "" ]; then
-    KIALI_USERNAME=$(read -p 'What do you want to use for the Kiali Username: ' val && echo -n $val)
+if [ "${AUTH_STRATEGY}" == "login" ]; then
```
jmazzitelli: We are missing the strategy in the configmap - so this actually won't work if the user wants "anonymous" here.
There is no strategy setting here: https://github.com/kiali/kiali/blob/master/deploy/kubernetes/configmap.yaml

We need this to look like the openshift configmap so the installer can set the strategy.. see this: https://github.com/kiali/kiali/blob/v0.16.0/deploy/openshift/configmap.yaml#L19-L20

mwringe:
@jmazzitelli Thoughts on including 'anonymous' here? I originally had it since I thought some people might shoot themselves in the foot and enabled anoymous without knowing that it makes Kiali wide open to anyone who has the endpoint. But I guess I can add it back (along with a warning).

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#749](https://paste.fedoraproject.org/paste/H~1P9R~kyRf8C22ZHFhz-Q)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1523]](https://paste.fedoraproject.org/paste/G06oD5kXdJDMoIxznlJA7A)

PR #950, opened 2019-03-25
fix some things I noticed while testing KIALI-2644

jmazzitelli:
-- the Makefile didn't pass through the strategy for the k8s-deploy target - this fixes that
-- the install scripts now echo a message when the secret creation is skipped so its obvious what is going on
-- get the scripts to look more identical (trying to keep these more or less in sync with each other, minus differences required between k8s and OS)

### DIFF
--- a/deploy/openshift/deploy-kiali-to-openshift.sh
+++ b/deploy/openshift/deploy-kiali-to-openshift.sh
@@ -140,7 +140,6 @@ if [ "${AUTH_STRATEGY}" != "login" ] && [ "${AUTH_STRATEGY}" != "openshift" ] &&
   exit 1
 fi
 
-
 if [ "${AUTH_STRATEGY}" == "login" ]; then
   # The credentials can be specified either as already base64 encoded, or in plain text.
   # If the username or passphrase plain text variable is set but empty, the user will be asked for a value.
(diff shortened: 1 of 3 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#750](https://paste.fedoraproject.org/paste/~lj~~Ts8y19MoIm6TsYF1A)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1524]](https://paste.fedoraproject.org/paste/pH48Rs2h-RNBG~HYxJyeiw)

deploy/openshift/deploy-kiali-to-openshift.sh
```diff
@@ -267,21 +266,26 @@ apply_dashboard() {
...
+
 # Only deploy the secret if we are using the login AUTH_STRATEGY
 if [ "${AUTH_STRATEGY}" == "login" ]; then
   apply_resource secret
   if [ "$?" != "0" ]; then
     echo "ERROR: Failed to deploy the Kiali secret. Aborting."
     exit 1
   fi
+else
+  echo "Using strategy [${AUTH_STRATEGY}] - a secret is not needed so one will not be created."
```
jshaughn: next time you make an edit:
" a secret is not needed so one will not be created" ->  "a secret is not needed so none will be created."

jmazzitelli: done

PR #951, opened 2019-03-26
make results length available when fetchAllMetrics

ChangyuWang:
** Describe the change **
make a available length slice to store results of fetching metrics


rhqci:

jotak:
Hi @ChangyuWang , thanks for your contrib.
Just to make sure I understand the intent: this is not solving any actual bug, but is optimizing the array size when not all metrics are being fetched, correct?

ChangyuWang:
> Hi @ChangyuWang , thanks for your contrib.
> Just to make sure I understand the intent: this is not solving any actual bug, but is optimizing the array size when not all metrics are being fetched, correct?

true it is.

PR #824, opened 2019-02-04
KIALI-2147 Change DR validations to support mesh/ns-wide mTLS enabled

xeviknal:
** Describe the change **
Destination Rules validations now takes into account the mTLS enablers (DR + MeshPolicy combo).

** Issue reference **
https://issues.jboss.org/browse/KIALI-2147

** Backwards incompatible? **
yep.

PR #788, opened 2019-01-17
Ignore VSCode directories for git

burmanm:
** Describe the change **

Adds .vscode/ to .gitignore to prevent those directories from appearing in the git status / add / commits / etc. These appear when for example using pipenv to run our e2e tests with Python and the VSCode has different settings to defaults (per project).

PR #787, opened 2019-01-17
[hack] adds mirroring to bookinfo reviews

jmazzitelli:
oc apply this yaml to your bookinfo demo and see mirroring in action (look at traffic going from productpage to reviews).

Issue #786, opened 2019-01-17
Kiali cannot get prometheus data when prometheus enable https


clyang82:
From kiali document:
`prometheus_service_url` The URL used to access and query the Prometheus Server. It must be accessible from Kiali pod. (default is http://prometheus.istio-system:9090)

When I configured `prometheus_service_url=https://prometheus.istio-system:9090`, the kiali cannot collect prometheus data and there has exception thrown from kiali pod
```
E0117 09:47:28.726655       1 graph.go:853] Get https://prometheus.istio-system:9090/api/v1/query?query=round%28sum%28rate%28istio_requests_total%7Breporter%3D%22destination%22%2Csource_workload%3D%22unknown%22%2Cdestination_service_namespace%3D%22cert-manager%22%2Cresponse_code%3D~%22%5B2345%5D%5B0-9%5D%5B0-9%5D%22%7D+%5B60s%5D%29%29+by+%28source_workload_namespace%2Csource_workload%2Csource_app%2Csource_version%2Cdestination_service_namespace%2Cdestination_service_name%2Cdestination_workload%2Cdestination_app%2Cdestination_version%2Cresponse_code%29%2C0.001%29&time=2019-01-17T09%3A47%3A28Z: x509: certificate signed by unknown authority: goroutine 21 [running]:
runtime/debug.Stack(0xc42063a280, 0x26e, 0xc42069c3a0)
	/usr/local/go/src/runtime/debug/stack.go:24 +0xa7
github.com/kiali/kiali/handlers.handlePanic(0x192a100, 0xc4206e8b60)
	/root/go/src/github.com/kiali/kiali/handlers/graph.go:853 +0xa6
panic(0x1005f20, 0xc42069c3a0)
	/usr/local/go/src/runtime/panic.go:491 +0x283
github.com/kiali/kiali/handlers.checkError(0x1913d80, 0xc42026c900)
	/root/go/src/github.com/kiali/kiali/handlers/graph.go:836 +0xaa
github.com/kiali/kiali/handlers.promQuery(0xc42056a900, 0x17a, 0x0, 0xed3d24630, 0x19a1460, 0x1920d80, 0xc420223e10, 0x0, 0x0, 0x0)
	/root/go/src/github.com/kiali/kiali/handlers/graph.go:822 +0x468
[...]

PR #961, opened 2019-04-01
KIALI-2546 Runtimes metrics discovery 

jotak:
This is enabling auto-discovery of dashboards, removing the need for pods annotations.

Pods annotations mechanism is not removed however: it can still be useful for dashboards where we are unable to provide a distinctive criteria for discovery, or also to speed up details display (annotation processing is much faster than discovery).

Some performance consideration, measuring the loading time of Workload Details page:
- (1) Before this PR, with no explicit runtimes dashboard: avg ~= 13ms
- (2) Before this PR, with explicit runtimes dashboards: avg ~= 20ms
- (3) After this PR, with explicit runtimes dashboards: avg ~= 20ms (no change, as expected)
- (4) After this PR, with no explicit dashboard / auto-discovery: avg ~= 180ms
[EDIT] / (4-bis) got better figures by adding goroutines, (4) now has avg ~= 75ms

pic

With (4-bis):
pic2

This peak due to discovery was kinda expected, see description in https://issues.jboss.org/browse/KIALI-2546 ; suggestion is to cache dashboards.

Note however that the increased latency *only* affects the Workloads / App details page, (not list pages), which makes it less critical as there's no scaling issue (we only show workload details one by one).

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -153,9 +153,10 @@ func (in *DashboardsService) GetDashboard(params prometheus.CustomMetricsQuery, 
 	}
 
 	aggLabels := models.ConvertAggregations(dashboard.Spec)
-	labels := fmt.Sprintf(`{namespace="%s",app="%s"`, params.Namespace, params.App)
+	cfg := config.Get()
+	labels := fmt.Sprintf(`{namespace="%s",%s="%s"`, params.Namespace, cfg.IstioLabels.AppLabelName, params.App)
 	if params.Version != "" {
-		labels += fmt.Sprintf(`,version="%s"`, params.Version)
+		labels += fmt.Sprintf(`,%s="%s"`, cfg.IstioLabels.VersionLabelName, params.Version)
 	} else {
 		// For app-based dashboards, we automatically add a possible aggregation/grouping over versions
 		versionsAgg := models.Aggregation{
@@ -347,9 +348,10 @@ func (in *DashboardsService) buildRuntimesList(namespace string, templatesNames 
 }
 
 func (in *DashboardsService) fetchMetricNames(namespace, app, version string) []string {
-	labels := fmt.Sprintf(`{namespace="%s",app="%s"`, namespace, app)
+	cfg := config.Get()
+	labels := fmt.Sprintf(`{namespace="%s",%s="%s"`, namespace, cfg.IstioLabels.AppLabelName, app)
 	if version != "" {
-		labels += fmt.Sprintf(`,version="%s"`, version)
+		labels += fmt.Sprintf(`,%s="%s"`, cfg.IstioLabels.VersionLabelName, version)
 	}
 	labels += "}"
 	metrics, err := in.prom.GetMetricsForLabels([]string{labels})
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -2,6 +2,7 @@ package business
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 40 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#765](https://paste.fedoraproject.org/paste/eX9we6qi6Fyl5c42nOIAqQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1570]](https://paste.fedoraproject.org/paste/2UuiIZ~yyXt-P~wVF4H~-A)

jotak:
Ideas to minimize impact on performances, non-exclusive:

- Caching dashboards (**pros**: efficient solution; **cons**: need to handle cache updates / invalidation)
- Flag-enabled feature (**pros**: can turn out completely to have 0 perf impact; **cons**: still perf issue when enabled)
- Async loading of available dashboards list (**pros**: fixes latency issue on Workload Details main info; **cons**: total loading time still unchanged (or even increased)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#766](https://paste.fedoraproject.org/paste/xWk5W2sN7uhIUYCLCMnUIA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1571]](https://paste.fedoraproject.org/paste/yCkSqX58kWFY2hFrttIe~g)

lucasponce:
> Note however that the increased latency only affects the Workloads / App details page, (not list pages), which makes it less critical as there's no scaling issue (we only show workload details one by one).

If this is under reasonable parameters, it doesn't sound bad.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#773](https://paste.fedoraproject.org/paste/h~c9oQNqyp~d0B4RMOMW~w)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1580]](https://paste.fedoraproject.org/paste/O0zN97S2b0GNE8m4Kugadw)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#786](https://paste.fedoraproject.org/paste/Gx3JNEDlYJTRXA0SFh8iOg)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1597]](https://paste.fedoraproject.org/paste/7RmTD5RYzMt~msEyS~1NEA)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#787](https://paste.fedoraproject.org/paste/ibMJixtVB1SXe8XbMJ0SRQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1598]](https://paste.fedoraproject.org/paste/7mXWYI~perYrkjPZrx0Gkw)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#788](https://paste.fedoraproject.org/paste/qGtaFz54OmCDzP6FwF-VqQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1599]](https://paste.fedoraproject.org/paste/ERLVjML51VtenRkFe3RlSw)

jotak:
@lucasponce @hhovsepy  this is ready for test/review

business/dashboards.go
```diff
@@ -243,26 +341,98 @@ func (in *DashboardsService) buildRuntimesList(namespace string, templatesNames
...
-				rtObj.DashboardRefs = append(rtObj.DashboardRefs, ref)
-				found = true
-				break
+		runtimes = addDashboardToRuntimes(dashboard, runtimes)
+	}
+	return runtimes
+}
+
+func (in *DashboardsService) fetchMetricNames(namespace, app, version string) []string {
+	labels := fmt.Sprintf(`{namespace="%s",app="%s"`, namespace, app)
```
lucasponce: Probably it's documented somewhere.
Runtime metrics follow this format, or it is proposed to follow, right ?
I mean "app" here is the same "app" label with same meaning used in Istio, right ?
I wonder if that's case, perhaps we can use the "app" constant from istio config ? 
Just in case if that changes, or at least, in the code to try to not code "app" but use that constant.

But it sounds a cornercase, so if it's unlikely just disregard my comment, that can be done in a future PR.

business/dashboards.go
```diff
@@ -243,26 +341,98 @@ func (in *DashboardsService) buildRuntimesList(namespace string, templatesNames
...
+	if version != "" {
+		labels += fmt.Sprintf(`,version="%s"`, version)
```
lucasponce: yeah, same thought here, as we are not using app/version but reading from config, I wonder if makes sense to align here.
At least to easy maintain in the future and tell developer that it's same constant.

business/dashboards.go
```diff
@@ -243,26 +341,98 @@ func (in *DashboardsService) buildRuntimesList(namespace string, templatesNames
...
+	// In all dashboards, finds the ones that match the metrics set
+	// We must exclude from the results included dashboards when both the including and the included dashboards are matching
+	runtimesMap := make(map[string]*v1alpha1.MonitoringDashboard)
+	for _, d := range allDashboards {
+		dashboard := d // sticky reference
+		matchReference := strings.TrimSpace(dashboard.Spec.DiscoverOn)
+		if matchReference != "" {
+			for _, metric := range metrics {
+				if matchReference == strings.TrimSpace(metric) {
+					if _, exists := runtimesMap[dashboard.Name]; !exists {
```
lucasponce: I wonder what may happens if there is some metric shared between dashboards.
I.e. same metric in nodejs or vertx.
What dashboard would find ?

skondkar:
Verified on:
kiali-ui 0.17.0 (0af3b25fe70da258149773a96468f8164db575aa)
kiali v0.17.0-SNAPSHOT (f0a2fb7775c46de9cf1a18272a90cf3cede5fd87)

Used bookinfo-runtimes and removed kiali.io/runtimes annotations from deplyment.yamls. After deploying,  all dashboards are discovered in Kiali with metrics.

business/dashboards.go
jotak: Actually, you're pointing me to something i didn't take care about: here the "app" label comes from prometheus label-mapping of k8s labels, so I should not hardcode "app" but use the same configurable App label that we use to identify k8s apps.

The same issue probably exists in the current code for custom dashboards.

jotak: If you don't mind I will enter a JIRA to check and fix this potential issue here and in other places

business/dashboards.go
jotak: since we don't break the `for _, d := range allDashboards` loop, all matching dashboards are going to appear if they share the same "discovery" metric. This is as I would expect...

jotak:
@jmazzitelli sure I will do

business/dashboards.go
jotak: https://issues.jboss.org/browse/KIALI-2674

jotak: (actually, I've fixed it here)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#824](https://paste.fedoraproject.org/paste/SCTkaP9IA6n2k5rwgJc1zQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1645]](https://paste.fedoraproject.org/paste/iRcxgvX7oPZ3bJE97lW0PA)

jotak:
So ... this is quite "funny". I'm seeing a race on the Config object, just because I've added a couple of `config.Get()`. The race makes sense, but what doesn't make sense to me is why it wasn't detected before? Because this PR doesn't change fundamentally anything here.

I'm currently seeing how to fix that... adding a mutex on Config works, but is costy. We should probably do something different and return a copy instead of pointer but i need to check all the code for impacts.
(cc @jmazzitelli @lucasponce )

lucasponce:
> So ... this is quite "funny". I'm seeing a race on the Config object, just because I've added a couple of config.Get(). The race makes sense, but what doesn't make sense to me is why it wasn't detected before? Because this PR doesn't change fundamentally anything here.

I'm currently seeing how to fix that... adding a mutex on Config works, but is costy. We should probably do something different and return a copy instead of pointer but i need to check all the code for impacts.
(cc @jmazzitelli @lucasponce )

A RW semaphore would help ?

I guess W op only happens at init, right ?

jmazzitelli:
> A RW semaphore would help ?
> 
> I guess W op only happens at init, right ?

Yes, that would be what I think, too. Add a RW lock with the W lock acquired at init time and released once the config is loaded. This is a * pointer being passed around for the config, but I wonder if there are things inside the config that are passed around by value that might be causing issues?? I'm just thinking out loud, but that's what I would be doing - see if a RW lock helps, but if this issue isn't happening at init time, then something else is going out, probably with the pass-by-ref vs. pass-by-value semantics in the config object itself.

jotak:
@jmazzitelli the issue happens on `test-race` which, if I understand correctly, runs several tests concurrently. So in that context the "init time" occurs several times, not just once, I think...

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#840](https://paste.fedoraproject.org/paste/RNcq83KtS1vSmDWZrohqJA)

FilipB:
kialiqeci

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#841](https://paste.fedoraproject.org/paste/An9WyVWdPvHc6-NGWiT3VA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1663]](https://paste.fedoraproject.org/paste/VT576UbfEFGBMLeOYnb0KQ)

jotak:
@lucasponce it's rebased, tests are green now

PR #972, opened 2019-04-03
KIALI-2668 Fixing validations around both ns-wide and mesh-wide mtls

xeviknal:
** Describe the change **

Scenarios fixed:

Scenario 1. 
MeshPolicy that uses PERMISSIVE mode + DestinationRule enabling mTLS traffic.
Problem: DR rule shows validation error: 'MeshPolicy enabling mTLS is missing'.
Expected: No validations shown for that DR.
Explanation: Permissive mode allows all kinds of traffic, mTLS included.

Scenario 2.
MeshPolicy in STRICT mode + DestinationRule enabling mTLS mesh-wide (classic scenario)
Policy ns-level in PERMISSIVE mode + DR disabling mTLS ns-wide.
Problem: DR rule shows validation error: MeshPolicy in strict mode has found, should be permissive.
Expected: No validations shown for that DR.
Explanation: ns-level Policy in PERMISSIVE mode allows disabled mtls connections.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2668

### DIFF
--- a/business/checkers/destinationrules/disabled_namespacewide_mtls_checker.go
+++ b/business/checkers/destinationrules/disabled_namespacewide_mtls_checker.go
@@ -21,11 +21,18 @@ func (m DisabledNamespaceWideMTLSChecker) Check() ([]*models.IstioCheck, bool) {
 
 	// otherwise, check among Policies for a rule enabling mTLS
 	for _, mp := range m.MTLSDetails.Policies {
-		// If Policy has mTLS enabled in the namespace
-		// traffic going through DestinationRule won't work
-		if strictMode := kubernetes.PolicyHasStrictMTLS(mp); strictMode {
-			check := models.Build("destinationrules.mtls.policymtlsenabled", "spec/trafficPolicy/tls/mode")
-			return append(validations, &check), false
+		if enabled, mode := kubernetes.PolicyHasMTLSEnabled(mp); enabled {
+			// If Policy has mTLS enabled in STRICT mode
+			// traffic going through DestinationRule won't work
+			if mode != "PERMISSIVE" {
+				check := models.Build("destinationrules.mtls.policymtlsenabled", "spec/trafficPolicy/tls/mode")
+				return append(validations, &check), false
+			} else {
+				// If Policy has mTLS enabled in PERMISSIVE mode
+				// traffic going through DestinationRule will work
+				// no need for further analysis in MeshPolicies
+				return validations, true
+			}
 		}
 	}
 
--- a/business/checkers/destinationrules/disabled_namespacewide_mtls_checker_test.go
+++ b/business/checkers/destinationrules/disabled_namespacewide_mtls_checker_test.go
@@ -31,6 +31,29 @@ func TestDRNSWideDisablingTLSPolicyPermissive(t *testing.T) {
 }
 
 // Context: DestinationRule ns-wide disabling mTLS connections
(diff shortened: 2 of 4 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#807](https://paste.fedoraproject.org/paste/3aFek9gY8Ra8creltABrKA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1622]](https://paste.fedoraproject.org/paste/hk3C1w9ULzIV1xkT3kFRxA)

business/checkers/destinationrules/disabled_namespacewide_mtls_checker.go
```diff
@@ -21,11 +21,18 @@ func (m DisabledNamespaceWideMTLSChecker) Check() ([]*models.IstioCheck, bool) {
 
 	// otherwise, check among Policies for a rule enabling mTLS
 	for _, mp := range m.MTLSDetails.Policies {
-		// If Policy has mTLS enabled in the namespace
-		// traffic going through DestinationRule won't work
-		if strictMode := kubernetes.PolicyHasStrictMTLS(mp); strictMode {
-			check := models.Build("destinationrules.mtls.policymtlsenabled", "spec/trafficPolicy/tls/mode")
-			return append(validations, &check), false
+		if enabled, mode := kubernetes.PolicyHasMTLSEnabled(mp); enabled {
```
burmanm: I'm slightly confused with this change. Wouldn't that "PolicyHasStrictMTLS" be basically the same thing? Or are there other modes than "PERMISSIVE" && "STRICT" ?

xeviknal: > Or are there other modes than "PERMISSIVE" && "STRICT" ?

No, there aren't more modes than that. 

> Wouldn't that "PolicyHasStrictMTLS" be basically the same thing?

No, because PolicyHasStrictMTLS returns only true for STRICT mode. And here we want to treat PERMISSIVE items too. If we found a Permissive policy, then the validation has to finish with no checks and true as validity.
If we don't take into account the Permissive mode, then validation checks for MeshPolicies and it might found a STRICT mode MeshPolicy and might return an invalid validation.


It fixes the Scenario 2:
MeshPolicy in STRICT mode + DestinationRule enabling mTLS mesh-wide (classic scenario)
Policy ns-level in PERMISSIVE mode + DR disabling mTLS ns-wide.
Problem: DR rule shows validation error: MeshPolicy in strict mode has found, should be permissive.
Expected: No validations shown for that DR.
Explanation: ns-level Policy in PERMISSIVE mode allows disabled mtls connections.

Perhaps another way to implement that would be (although I think is running almost same code twice):
```golang
// If there is a Policy in strict Mode then traffic from Destination Rule won't work
if strictMode := kubernetes.PolicyHasStrictMTLS(mp); strictMode {
	check := models.Build("destinationrules.mtls.policymtlsenabled", "spec/trafficPolicy/tls/mode")
	return append(validations, &check), false
}

// If we found a Policy in permissive mode, then traffic from Destination Rule will work
if permissiveMode := kubernetes.PolicyHasPermissiveMTLS(mp); permissiveMode {
	return validations, true
}
```

burmanm: Ah, the github view got me. It looked like the processing ended in both implementations.

PR #974, opened 2019-04-04
KIALI-2682 Move to using Kiali Operator

jmazzitelli:
This is going to change how devs work on Kiali.

Use operator/Makefile to deploy kiali as opposed to the install script and the two sets of deploy/ yaml files.

The main Makefile is still going to have the docker build targets so our CI builds still work,
but the main Makefile now delegates to the operator/Makefile for building the operator image.

Operator image is assumed to go into quay.io.

More documentation is needed - kiali.io gettingstarted will have to be updated.

PR #985, opened 2019-04-08
KIALI-2671: Update endpoints for OpenShift 4

mwringe:
** Describe the change **

Some of the endpoints we were using have now changed in OpenShift 4. This updates to use the newer endpoints.

They updates also work in OpenShift 3, so it should still continue to work for using currently on OpenShift 3. This change doesn't affect people running on kubernetes.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2671

### DIFF
--- a/deploy/openshift/clusterrole.yaml
+++ b/deploy/openshift/clusterrole.yaml
@@ -20,6 +20,11 @@ rules:
   - get
   - list
   - watch
+- apiGroups: ["route.openshift.io"]
+  resources:
+  - routes
+  verbs:
+  - get
 - apiGroups: ["extensions", "apps"]
   resources:
   - deployments
@@ -168,6 +173,11 @@ rules:
   - get
   - list
   - watch
+- apiGroups: ["route.openshift.io"]
+  resources:
+  - routes
+  verbs:
+  - get
 - apiGroups: ["extensions", "apps"]
   resources:
   - deployments
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -14,6 +14,8 @@ import (
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
+
+	"github.com/kiali/kiali/log"
 )
 
 type OpenshiftOAuthService struct {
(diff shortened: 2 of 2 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#866](https://paste.fedoraproject.org/paste/ud3vGTuZ7TlcsjEJ-gampA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1678]](https://paste.fedoraproject.org/paste/oyjoXbbrwInNWWPuZe8Lfw)

deploy/openshift/clusterrole.yaml
```diff
@@ -20,6 +20,11 @@ rules:
   - get
   - list
   - watch
+- apiGroups: ["route.openshift.io"]
```
jmazzitelli: there is a "routes" in the "" apiGroup - I think we need to delete that, no?

jmazzitelli: These same changes need to be made to the operator cluster role templates for the kiali and kiali-viewer roles... so make these same changes here:

* https://github.com/kiali/kiali/blob/master/operator/roles/kiali/templates/openshift/clusterrole.yaml
* https://github.com/kiali/kiali/blob/master/operator/roles/kiali/templates/openshift/clusterrole-viewer.yaml

Keep in mind, this deploy/clusterrole.yaml file (all of these under deploy/) are going away in favor of the operator. When I merge my operator PR these will be deleted. So we need to remember to keep the operator templates up to date going forward.

jmazzitelli: actually, @mwringe - I will add these to the operator PR right now. That way it will be in the same PR with everything else. so never mind my last comment :)

jmazzitelli: Here's the commit containing the changes to the operator that incorporate the new cluster role data: https://github.com/kiali/kiali/pull/974/commits/b56fe93311d0a7d6c4a75020b87e67e5229dad92

mwringe: > there is a "routes" in the "" apiGroup - I think we need to delete that, no?

I left it in there to make sure it wont affect anything when running on 3.x. Leaving it in doesn't cause any problems, but we could likely remove it.

Issue #898, opened 2019-03-07
Graph nodes don't appear


andrebriggs:
**Describe the bug**
The graph nodes don't appear in the graph view. I do see the edges (arrows). Below is a screenshot of what I see. I have added blue boxes to signify where I expected to see node

<img width="1148" alt="screen shot 2019-03-07 at 12 02 35 pm" src="https://user-images.githubusercontent.com/3119513/53936863-57fc3600-40d1-11e9-9d80-7e64d6d9de82.png">
 
**Versions used**

- kiali-ui 0.12.0 (356ea85ab943d34e30ccca50152dde5f901c927a)
- kiali v0.12.0 (2a44b37707bec1b0a954dd5d71f5687ad0672935)
- Istio 1.0.6
- Prometheus 2.7.2
- Kubernetes v1.12.5 (via Azure Kubernetes Services)
- JaegerURL http://jaeger-query.jaeger.svc.cluster.local
- GrafanaURL http://grafana.grafana.svc.cluster.local

**To Reproduce**
1. Go to `http://127.0.0.1:20001/console/graph/namespaces/?edges=trafficRatePerSecond&graphType=service&namespaces=default&injectServiceNodes=true&duration=86400&pi=900000&layout=cose-bilkent`

Here are the logs from the Kiali pod when I refresh the page
`
I0307 06:38:27.509652       1 versions.go:146] DEBUG: Detected Istio version [root@464fc845-2bf8-11e9-b805-0a580a2c0506-docker.io/istio-1.0.6-98598f88f6ee9c1e6b3f03b652d8e0e3cd114fa2-dirty-Modified]
I0307 06:38:27.814803       1 graph.go:87] DEBUG: Build [service] graph for [1] namespaces [map[default:{default 21h30m34s}]]
I0307 06:38:27.814856       1 graph.go:93] DEBUG: Build traffic map for namespace [{default 21h30m34s}]
I0307 06:38:27.814931       1 graph.go:986] DEBUG: Graph query:
round(sum(rate(istio_requests_total{reporter="destination",source_workload="unknown",destination_service_namespace="default",response_code=~"[2345][0-9][0-9]"} [77434s])) by (source_workload_namespace,source_workload,source_app,source_version,destination_service_namespace,destination_service_name,destination_workload,destination_app,destination_version,response_code),0.001)@time=2019-03-07 06:38:27 (now=2019-03-07 06:38:27, 1551940707)
[...]

PR #897, opened 2019-03-06
KIALI-2501 Fix browser console deprecated warning message

israel-hdez:
https://issues.jboss.org/browse/KIALI-2501

Front-end is using momentjs library to parse the dates the back-end returns. momentjs library works correctly only with ISO and RFC2822 date formats.

This is returning the date in an RFC2822 compatible date format.

This is the console error that this PR is fixing:

image

PR #999, opened 2019-04-11
Kiali 2665 update apis and qualify imports

jshaughn:
* Deployment batch_v1beta1 -> apps_v1
* ReplicaSet batch_v1beta2 -> apps_v1
* StatefulSet batch_v1beta2 -> apps_v1

* use explicit prefixes for versioned imports for better identification and future update.
* use consistent import prefix names/format. Some examples:
      - v1 -> core_v1, meta_v1
      - osappsv1 -> osapps_v1
      - osv1 -> osproject_v1

* fix some import ordering


graph/appender/util.go
```diff
@@ -8,13 +8,13 @@ import (
 	"github.com/kiali/kiali/graph"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
-	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	prom_v1 "github.com/prometheus/client_golang/api/prometheus/v1"
```
jmazzitelli: again, I see "prom" when you would think its "prometheus_v1". I think I can infer that you are restricting the name with four-characters - what happens if four characters aren't enough to make an alias unique? Then you really are going to trigger my OCD to an 11.

handlers/grafana_test.go
```diff
@@ -6,7 +6,8 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
-	"k8s.io/api/core/v1"
+	core_v1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
jmazzitelli: why is the same import here twice with different aliases?

kubernetes/types.go
```diff
@@ -5,9 +5,9 @@ import (
 	"strings"
 
 	"github.com/kiali/kiali/config"
-	"k8s.io/api/apps/v1beta1"
-	autoscalingV1 "k8s.io/api/autoscaling/v1"
-	v1 "k8s.io/api/core/v1"
+	apps_v1 "k8s.io/api/apps/v1"
+	autoscaling_v1 "k8s.io/api/autoscaling/v1"
```
jmazzitelli: MY EYES! MY EYES!! OK, this is how I would want it to be, but following your pattern, why didn't you call this autos_v1? You are arbitrarily abbreviating some places but not others, it seems.

josejulio: `auto_v1` *

business/tls_test.go
```diff
@@ -3,8 +3,8 @@ package business
 import (
 	"testing"
 
-	osv1 "github.com/openshift/api/project/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	osproj_v1 "github.com/openshift/api/project/v1"
```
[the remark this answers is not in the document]
jshaughn: OK, well I reckoned the abbrev was reasonably clear but I agree that if we are going to be explicit we may as well be explicit.  I'll change to osproject_v1.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#897](https://paste.fedoraproject.org/paste/bX3dA7rptQu-hQiRQQqSvQ)

handlers/grafana_test.go
jshaughn: good catch.

graph/appender/util.go
jshaughn: Yes, OK, so prometheus_v1 just starts to get long.  Here is what I will do, I will meet you in the middle.  If 4-letters (or less) is not sufficiently clear I'll use the whole name.  I believe "prom" is fine here, so it will stay.

kubernetes/types.go
jshaughn: OK, so because "auto" is not very descriptive, and there are not many occurrences, I am keeping the explicit naming.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#898](https://paste.fedoraproject.org/paste/eOKnjrHoQEzGVaIVddWCAw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1695]](https://paste.fedoraproject.org/paste/KksJIqUnLN3h504IkGL5SQ)

PR #1011, opened 2019-04-16
KIALI-2718 Don't confuse no response time with rt=0

jshaughn:
- add test for this case
- fix some linter suggestions

Before (rt < 50):
image
After  (rt < 50):
image


rhqci:

Issue #984, opened 2019-04-07
Kiali graph not showing traffic to some services of the mesh


codependent:
**Describe the bug**
I have a mesh consisting of 4 services, one of them aggregates calls to the other 3 ones. The Kiali UI graph shows a green arrow from the istio-ingressgateway to the first one (global-position-service), but it doesn't show traffic to the others although the mesh is working fine:

<img width="723" alt="Screenshot 2019-04-07 at 12 26 43" src="https://user-images.githubusercontent.com/6399869/55682249-4658b900-5931-11e9-8ae6-2bc10492fa3f.png">

**Versions used**
Kiali: 0.7.2
Istio: 1.0.2
Kubernetes flavour and version: OpenShift 3.10

**To Reproduce**
Gateway configuration:
```apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: insuranceinc-gateway
  namespace: istio-insuranceinc
spec:
  selector:
    istio: ingressgateway # use istio default controller
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - "*"```
Virtual service configuration:
```apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: insuranceinc
  namespace: istio-insuranceinc
spec:
  hosts:
  - "*"
  gateways:
  - insuranceinc-gateway
  http:
  - match:
    - uri:
        prefix: /insuranceinc
    rewrite:
      uri: " "
    route:
    - destination:
        host: global-position-service
        port:
          number: 8080
```

Every service has its associated k8s service, e.g.:

```
- apiVersion: v1
  kind: Service
  metadata:
    creationTimestamp: 2019-04-06T16:40:18Z
    labels:
      app: global-position-service
      template: springbootembedded-build
      version: 1.0.0
    name: global-position-service
    namespace: istio-insuranceinc
    resourceVersion: "88666"
    selfLink: /api/v1/namespaces/istio-insuranceinc/services/global-position-service
    uid: a9b92f9a-588a-11e9-aae1-66e73c3101a7
  spec:
    clusterIP: 172.30.90.0
    ports:
    - name: tcp-8080
      port: 8080
      protocol: TCP
      targetPort: 8080
    selector:
      deploymentconfig: global-position-service
    sessionAffinity: None
    type: ClusterIP
  status:
    loadBalancer: {}
- apiVersion: v1
  kind: Service
  metadata:
[...]

PR #970, opened 2019-04-03
Enable code coverage for system tests

FilipB:
** Describe the change **

This commit adds possibility to build kiali which can produce code coverage report even for system tests which run against kiali deployed in OpenShift. Note this does not do anything for code coverage of unit tests.

PR #963, opened 2019-04-01
KIALI-577 Server-side REST support for getting pod details and pod logs

jshaughn:
Starting with a draft PR for this potential feature for dislaying pod logs via the Kiali console.  This is the initial server-side support, adding REST API endpoints for both Pod details and Pod logs.

@jotak If you'd like to take a look or give it a try please do so.  @jmazzitelli , @mwringe, Also take a look if you'd like to test/review/provide feedback.  I'll add you all as reviewers but feel free to take your name off if you're not interested in the draft PR.

*** REQUIRES UI PR: https://github.com/kiali/kiali-ui/pull/1101

** TODO create an upstream PR to change the helm chart for pods/log permission addition

Issue #954, opened 2019-03-27
Question: Can use more than one label to group my app?


HeathLee:
https://github.com/kiali/kiali/blob/25707c3a2fb00aa12bd93083668cb64df208a7b0/config.yaml#L44
I want use two label to identify my app, is this possible?

jotak:
This is not currently possible. To better understand your request, do you want to use two labels interchangeably ( `app1=foo` or `app2=foo` ) or do you want to have a combination of labels to identify the app? (`app1=foo && app2=bar`)

jmazzitelli:
Note, this isn't possible with out-of-the-box Istio either - you would need to alter the metrics where the telemetry comes from in order to even get Prometheus to store data for apps that use multiple labels. So this would require changes to both Kiali and Istio.

abonas:
at the moment Istio is taking into account the app and the version labels (app label can be configured to a different one in Kiali), but indeed as Joel asked above, it's important to understand what use case are you looking for.

HeathLee:
I  want have a combination of labels to identify the app? `app1=foo && app2=bar`

HeathLee:
My apps provide rest api and grpc both. I labeld them with `app=foo tier=api` or `app=foo tier=rpc`

jmazzitelli:
> My apps provide rest api and grpc both. I labeld them with `app=foo tier=api` or `app=foo tier=rpc`

If you label your apps that way (and assuming you are using Istio telemetry definitions that come out of the box), and you look in the Istio telemetry (look in Istio's Prometheus) you will see that all your tiers (api and rpc) will be aggregated into a single app=foo. So even Istio won't be splitting out the telemetry between your different tiers like you want. And Kiali just displays what Istio stores in telemetry (in Prometheus) - so it will so the aggregated Istio telemetry.

Perhaps one workaround for this could be having three labels "app=foo-api" along with "base-app=foo" and "tier=api" -- this will still give you the separate labels in case you want to do your own querying (base-app and tier) but provides the combined app label for Istio (and thus Kiali) to use.

PR #1022, opened 2019-04-22
KIALI-2446 AutoDiscover Jaeger and Grafana from backend

aljesusg:
[KIALI-2446](https://issues.jboss.org/browse/KIALI-2446)

To test this we need to remove https://github.com/kiali/kiali/blob/master/operator/roles/kiali-deploy/tasks/openshift/os-jaeger.yml where we assign the route http://tracing-istio-system.127.0.0.1.nip.io , the autodiscover detect if there is SSL.

Actually if you have jaeger with SSL the kiali operator tasks that returns the tracing url is not working because return http instead https.

cc @jmazzitelli 

Autodiscover for OpenShift


rhqci:

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#942](https://paste.fedoraproject.org/paste/LZZw9bc~5FDTahVA4qw2uA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1754]](https://paste.fedoraproject.org/paste/4NmKVLavngfKf8y0eFukCw)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#945](https://paste.fedoraproject.org/paste/DBJT6St-cwX5S0E80J7cqw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1757]](https://paste.fedoraproject.org/paste/7UzUqA3gcZknVuPKIpq7MQ)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#951](https://paste.fedoraproject.org/paste/7QhYA7nqh1PtuysTI9h-UQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1762]](https://paste.fedoraproject.org/paste/gIrygQtFLlhRJZFBs02-6Q)

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -54,7 +54,8 @@ kiali_defaults:
       istio_sidecar_annotation: "sidecar.istio.io/status"
       url_service_version: ""
     jaeger:
-      service: "jaeger-query"
+      service: "tracing"
+      jaeger_service: "jaeger-query"
```
jmazzitelli: Two things here:
1. You are missing service_namespace. Even if empty string, explicitly specify it. Virtual all of these settings should be set to their actual defaults - even if empty string.
2. Alphabetize the list. All settings are in alphabetical order. So under jaeger, you should list the children settings in this order: jaeger_service, service, service_namespace, url

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#952](https://paste.fedoraproject.org/paste/Tkkdy8POmdVIt-G27oKB2Q)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1763]](https://paste.fedoraproject.org/paste/qvA6U2m--8SRNbALvZVVWw)

kubernetes/client.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	osapps_v1 "github.com/openshift/api/apps/v1"
 	osproject_v1 "github.com/openshift/api/project/v1"
+	osroutesv1 "github.com/openshift/api/route/v1"
```
jmazzitelli: Last week, we just introduced consistent patterns to our import aliases. This doesn't match the new pattern all our import alias are following now. Imports of the form .../api/ABC/vN results in import aliases of abc_vn - if it is an openshift specific import, prefix with "os"... so in this case, change this import alias to "osroutes_v1".

kubernetes/kubernetes_service.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	osapps_v1 "github.com/openshift/api/apps/v1"
 	osproject_v1 "github.com/openshift/api/project/v1"
+	osroutesv1 "github.com/openshift/api/route/v1"
 	apps_v1 "k8s.io/api/apps/v1"
```
jmazzitelli: import alias name should be osroutes_v1

kubernetes/kubetest/mock.go
```diff
@@ -5,6 +5,7 @@ import (
 
 	osapps_v1 "github.com/openshift/api/apps/v1"
 	osproject_v1 "github.com/openshift/api/project/v1"
+	osroutesv1 "github.com/openshift/api/route/v1"
```
jmazzitelli: import alias should be osroutes_v1

status/discover.go
```diff
@@ -0,0 +1,76 @@
+package status
+
+import (
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+	"io/ioutil"
+)
```
jmazzitelli: kiali has a pattern to our imports. We put standard imports first, then blank line, then third-party imports, then blank line, then kiali imports. Each section in alphabetical order (this sorting is done automatically by gofmt, which is why we break the groups up with newlines). So this import section should be:

```
import (
	"io/ioutil"

)
```

kubernetes/client.go
aljesusg: we should set this in the https://github.com/kiali/kiali/blob/master/STYLE_GUIDE.adoc

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#953](https://paste.fedoraproject.org/paste/cpHWfyZn68soCyZqRz~pYA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1764]](https://paste.fedoraproject.org/paste/YaL2qyrJKfsmIlz-SF8jbA)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#954](https://paste.fedoraproject.org/paste/OYESIZcj4mktoxDOSzKK2A)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1765]](https://paste.fedoraproject.org/paste/RgULF5twWWVGLhS~SuQndQ)

status/versions.go
```diff
@@ -187,15 +181,15 @@ type p8sResponseVersion struct {
 func jaegerVersion() (*ExternalServiceInfo, error) {
 	product := ExternalServiceInfo{}
 	product.Name = "Jaeger"
-	product.Url = config.Get().ExternalServices.Jaeger.URL
+	product.Url, _ = DiscoverJaeger()
```
lucasponce: I wonder if DiscoverJaeger() should be a one time operation or invoked everytime that jaegerVersion() is called.
If I understand we call /status from the UI very often.
Perhaps this is a thing to be cached some way, as I don't think the location of jaeger/grafana will change to invoke the Discover on every request of the UI.

aljesusg: mmm Maybe we should create another endpoint to force reload the status like /api/status/refresh and let this to be a one time operation

pilhuhn: +1 on caching that and not evaluating every time /status is called.
It could be from the UI called the moment the About... box is shown.

lucasponce: Creating and maintaining a new endpoint sounds complex here, I mean, once Kiali has discovered there endpoints, probably they won't change unless some troubleshooting/demo where then probably a workaround could be to recreate kiali pod to refresh it.

I've just proposed something simple here, I mean, if we find some trouble with this design (we need to refresh that info) the suggested /status/refresh or similar operation can be implemented, but I don't see a reason to do it in a normal operation (and to recreate the pod sounds reasonable for a troubleshooting workaround).

pilhuhn: If the URLs are not set in config, we call the k8s API for this every time we call /status (if I get the code above). I don't think those additional calls every few seconds are really needed. 
But my Golang code reading-fu is not strong ..

aljesusg:
@jmazzitelli I am checking this....I don't know why is not set the service name ...in the setup of the configuration in config.go 
```go
c.ExternalServices.Jaeger.ServiceNamespace = strings.TrimSpace(getDefaultString(EnvJaegerServiceNamespace, ISTIO_NAMESPACE)) 
```
It's correctly but after that when the code goes to set method this value is empty I don't understand this :( I need to debug this

```go
                                                                                                      
// Set the global Config                                                                              
func Set(conf *Config) {                                                                              
	log.Debugf("Conf Set is %s", conf.ExternalServices.Jaeger.ServiceNamespace)                       
	rwMutex.Lock()                                                                                    
	defer rwMutex.Unlock()                                                                            
	configuration = *conf                                                                             
                                                                                                      
}                                                                                                     
`

**Comment by rhqci** [comment]

**Comment by rhqci** [comment]
Jenkins CI: kiali-core-pr-e2e-test [#960](https://paste.fedoraproject.org/paste/NmQz8VjoRBbEJnJ3acEkcA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1774]](https://paste.fedoraproject.org/paste/Gh1Zwl95HpfKQUFImnXDKA)

**Comment by hhovsepy** [comment]
kialiqeci

**Comment by rhqci** [comment]

File: config/config.go [hunk 7]
```diff
@@ -138,10 +138,10 @@ type GrafanaConfig struct {
...
 // JaegerConfig describes configuration used for jaeger links
 type JaegerConfig struct {
-	URL              string `yaml:"url"`
-	Service          string `yaml:"service"`
-	JaegerService    string `yaml:"jaeger_service"`
-	ServiceNamespace string `yaml:"service_namespace"`
+	URL           string `yaml:"url"`
+	Service       string `yaml:"service"`
+	JaegerService string `yaml:"jaeger_service"`
+	Namespace     string `yaml:"namespace"`
```
jmazzitelli: did you change these settings in the operator too? 

https://github.com/kiali/kiali/pull/1022/files#diff-f8f6c9abc6c5969045b61451c949e922R174
https://github.com/kiali/kiali/pull/1022/files#diff-f8f6c9abc6c5969045b61451c949e922R181
https://github.com/kiali/kiali/pull/1022/files#diff-c924dfb634d9c40dbd97889668ab1efdR59

hhovsepy:
kialiqeci

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#963](https://paste.fedoraproject.org/paste/kD36RwEO0dQvoBa3Lse8qg)

config/config.go
aljesusg: No @jmazzitelli, because I was checking why the service Namespace wasn't work , with this change is working, I'll update the code asap

aljesusg:
Updated

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#970](https://paste.fedoraproject.org/paste/bRg3EypnTxBaDNEt01IHsg)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1783]](https://paste.fedoraproject.org/paste/3KG4R5JYHCKSPndTIG2yTQ)

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -171,14 +171,14 @@ spec:
 # **Jaeger-specific settings:
 # jaeger_service: The Kubernetes service name for jaeger. Kiali uses this to connect within the cluster to Jaeger.
 # service: The Kubernetes service name for Tracing. Kiali uses this to connect within the cluster to Tracing.
-# service_namespace: The Kubernetes namespace that holds the Jaeger/Tracing service (if empty, assumes the same value as deployment.namespace)
+# namespace: The Kubernetes namespace that holds the Jaeger/Tracing service (if empty, assumes the same value as deployment.namespace)
```
jmazzitelli: please alphabetize - all settings are alphanumerically sorted .... as per my previous comment:

> (don't forget to put them in the proper sorted order -- now that its "namespace" and not "service_namespace" it will reorder what you have currently)

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -171,14 +171,14 @@ spec:
...
 # url: The URL that Kiali uses when integrating with Jaeger. This URL must be accessible to clients external to
 #      the cluster in order for the integration to work properly. If empty, an attempt to auto-discover it is made.
 #    ---
 #    jaeger:
 #      jaeger_service : "jaeger-query"
 #      service: "tracing"
-#      service_namespace: ""
+#      namespace: ""
```
jmazzitelli: please alphabetize - all settings are alphanumerically sorted.

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -56,7 +56,7 @@ kiali_defaults:
     jaeger:
       jaeger_service: "jaeger-query"
       service: "tracing"
-      service_namespace: ""
+      namespace: ""
```
jmazzitelli: please alphabetize - all settings are alphanumerically sorted.

status/discover.go
```diff
@@ -60,18 +80,21 @@ func discoverUrlService(ns string, service string) (url string, err error) {
 	}
 	// If the client is not openshift return and avoid discover
 	if !client.IsOpenShift() {
+		log.Debugf("Kiali detect a no Openshift Cluster")
```
jmazzitelli: I don't think we need all of these debug messages in here - perhaps make them Trace level if you want to keep them. This one doesn't explain where in the processing it is - so seeing this message "not in OpenShift" will be confusing because its not obvious why this debug message is saying this. Also, reword this message... say something like "Discovering external service within a non-OpenShift cluster". That way at least people reading the log will know this message is coming from an attempt to auto-discover jaeger or grafana or some other external service.

status/discover.go
```diff
@@ -13,32 +14,51 @@ var saToken string
...
+	ns := conf.ExternalServices.Jaeger.Namespace
 	// Check if there is a Tracing service in the namespace
-	url, err = discoverUrlService(config.Get().ExternalServices.Jaeger.ServiceNamespace, config.Get().ExternalServices.Jaeger.Service)
+	url, err = discoverUrlService(ns, tracing)
 	if err != nil || url == "" {
 		// Check if there is a Jaeger-Query service in the namespace
-		url, err = discoverUrlService(config.Get().ExternalServices.Jaeger.ServiceNamespace, config.Get().ExternalServices.Jaeger.JaegerService)
+		url, err = discoverUrlService(ns, jaeger)
+		if err != nil || url == "" {
+			log.Debugf("Services %s or %s not detected  in namespace %s", jaeger, tracing, ns)
```
jmazzitelli: trivial - there are two spaces in between "detected' and "in"

status/discover.go
```diff
@@ -13,32 +14,51 @@ var saToken string
...
+	log.Debugf("Kiali is looking for Grafana service ...")
+	url, err = discoverUrlService(config.Get().ExternalServices.Grafana.ServiceNamespace, config.Get().ExternalServices.Grafana.Service)
+	conf.ExternalServices.Grafana.URL = url
+	config.Set(conf)
+	return url, err
 }
 
 func discoverUrlService(ns string, service string) (url string, err error) {
 	if saToken == "" {
+		log.Debugf("Reading sa Token")
```
jmazzitelli: This debug message will be confusing because just seeing "Reading sa Token" doesn't indicate why/what it is doing this for. I recommend changing to Tracef. Also, perhaps say, "Reading the SA token in order to detect external service".

status/discover.go
```diff
@@ -60,18 +80,21 @@ func discoverUrlService(ns string, service string) (url string, err error) {
 	}
 	// If the client is not openshift return and avoid discover
 	if !client.IsOpenShift() {
+		log.Debugf("Kiali detect a no Openshift Cluster")
 		return "", nil
 	}
 	route, err := client.GetRoute(ns, service)
 	if err != nil {
+		log.Debugf("Kiali not detect the service %s in the namespace %s", service, ns)
```
jmazzitelli: add the word "did" for slightly better English: "Kiali did not detect the service [%s] in the namespace [%s]"

status/discover.go
```diff
@@ -60,18 +80,21 @@ func discoverUrlService(ns string, service string) (url string, err error) {
...
 		return "", err
 	}
 
 	host := route.Spec.Host
-
 	if route.Spec.TLS != nil {
+		log.Debugf("Kiali detect the service %s in the namespace %s with SSL in %s", service, ns, "https://"+host)
```
jmazzitelli: add the word "did" for slightly better English: "Kiali did not detect the service [%s] in the namespace [%s] with SSL in [%s]"

status/discover.go
```diff
@@ -60,18 +80,21 @@ func discoverUrlService(ns string, service string) (url string, err error) {
...
 		return "https://" + host, nil
 	} else {
+		log.Debugf("Kiali detect the service %s in the namespace %s without SSL in %s", service, ns, "http://"+host)
```
jmazzitelli: add the word "did" for slightly better English: "Kiali did not detect the service [%s] in the namespace [%s] without SSL in [%s]"

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#984](https://paste.fedoraproject.org/paste/fPHtu8yYBnEeuClaSi8i~g)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1798]](https://paste.fedoraproject.org/paste/XQHYKFTM~43j~A4N0mJC4w)

status/discover.go
```diff
@@ -42,14 +36,11 @@ func DiscoverGrafana() (url string, err error) {
...
 	if !conf.ExternalServices.Grafana.DisplayLink {
-		log.Debugf("Grafana display link is disabled in the configuration.")
 		return "", nil
 	}
 	if conf.ExternalServices.Grafana.URL != "" {
-		log.Debugf("Detected url %s configured for tracing service", config.Get().ExternalServices.Grafana.URL)
 		return conf.ExternalServices.Grafana.URL, nil
 	}
-	log.Debugf("Kiali is looking for Grafana service ...")
 	url, err = discoverUrlService(config.Get().ExternalServices.Grafana.ServiceNamespace, config.Get().ExternalServices.Grafana.Service)
```
jmazzitelli: I see you took out all the debug messages, but you should have AT LEAST one debug message that logs what the newly discovered URL is. This is going to be helpful to let us know that this discovery actually was performed and it let's us know what it found.

Same is true for both the grafana and jaeger discovery... log a message for each one that is found.

config/config.go
```diff
@@ -205,6 +209,9 @@ type Config struct {
 	Auth             AuthConfig        `yaml:"auth,omitempty"`
 }
 
+// Istio System namespace by default
+var ISTIO_NAMESPACE = "istio-system"
```
jotak: that should be `const`

status/discover.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if conf.ExternalServices.Jaeger.URL != "" {
+		return conf.ExternalServices.Jaeger.URL, nil
+	}
+	tracing := conf.ExternalServices.Jaeger.Service
+	jaeger := conf.ExternalServices.Jaeger.JaegerService
+	ns := conf.ExternalServices.Jaeger.Namespace
+	// Check if there is a Tracing service in the namespace
+	url, err = discoverUrlService(ns, tracing)
+	if err != nil || url == "" {
```
jotak: I don't understand this part. It seems like the URL we will used can be picked up of any of "jaeger" service or "tracing" service? What use case is it intended to solve? And why wouldn't we have just 1 service name config and the user is expected to provide the one that works?

Also, I find the config names a bit confusing: "Jaeger.Service" and "Jaeger.JaegerService" it's not easy to understand what they relate to.

aljesusg: This fetch if there is a tracing service first, if not we get the jaeger-query service

status/versions.go
jotak: So at the moment it's set in config so not re-discovering every time.
Now the drawback is that the only way to refresh it (if the user changed grafana/jaeger url), kiali needs to be restarted. But I guess it's a reasonable-enough limitation.

status/discover.go
jotak: My question was more about why?

jotak: I mean, can really jaeger-query service and tracing service be used interchangeably? That seems odd.

aljesusg: My concern is that there may be environments without the tracing service, so in that case we should look for jaeger service @jotak

rhqci:
PRT Failed, Please Contact QE

rhqci:
PRT Failed, Please Contact QE

status/discover.go
jotak: But if someone has an environment where Jaeger is accessible from another service name... then he would just have to put that service name in config, no? I'm still missing something.

aljesusg:
> While testing, there is still an issue, front-end reporting that the grafana URL isn't set up in config.
> The error is due to the ui calling both endpoints /grafana and /status , and only the latter triggers the auto-discovery stuff.
> 
> So I think from the Grafana handler we should also call the discovery function when URL is not set.

makes sense

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#999](https://paste.fedoraproject.org/paste/NV~r2b2APvX-RunODQ-y~Q)

rhqci:
PRT Failed, Please Contact QE

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1025](https://paste.fedoraproject.org/paste/Ta~ZSWk9HqTttAZxpD8A-Q)
  * :red_circle: run-kiali-e2e-tests [#[1845]](https://paste.fedoraproject.org/paste/DROBnAV6Sp~4ZWfabIdKzw)

handlers/grafana.go
```diff
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/kiali/kiali/status"
```
jmazzitelli: put this import in the proper place according to our coding standards (its a kiali import, so its in the bottom import group).

operator/roles/kiali-deploy/defaults/main.yml
jmazzitelli: the "tracing" setting should move below "prometheus" to maintain alphabetical sort order of the settings.

```
prometheus:
  custom_metrics_url: ""
  url: ""
tracing:
  namespace: ""
  service: ""
  url: ""

**Comment by hhovsepy** [comment]
AutoDiscovery works, tested the cases:
1. Grafana and Jaeger URLs are not set:
1.1 The Grafana and Jaeger Routes exists on OC side. Kiali starts and discovers both. Here in Kiali configmap by default the Grafana URL is set, but the Jaeger URL is not. Is finding and setting Grafana URL by operator design? @jmazzitelli 
1.2 Roues does not exist on OC side, Kiali started, both URL are undefined in about box, routes are created, Kiali discovers both.

File: config/config.go [hunk 21]
```diff
@@ -134,10 +134,12 @@ type GrafanaConfig struct {
 	Password                 string `yaml:"password"`
 }
 
-// JaegerConfig describes configuration used for jaeger links
-type JaegerConfig struct {
-	URL     string `yaml:"url"`
-	Service string `yaml:"service"`
+// TracingConfig describes configuration used for tracing links
+type TracingConfig struct {
+	EnableJaeger bool
```
jmazzitelli: This boolean is confusing to me. What is it used for? I do not see it being set in any configuration setting, and its hardcoded default is "false". Perhaps you should add a comment somewhere to explain what this is for. If this is NOT meant to come from the config file, it might also be good to add "yaml:-" here to make it clear it doesn't belong in the configmap yaml and will be skipped:

```
   EnableJaeger bool `yaml:-`
```

config/config.go
```diff
@@ -255,9 +266,9 @@ func NewConfig() (c *Config) {
 		log.Error("Grafana username (\"GRAFANA_USERNAME\") requires that Grafana password (\"GRAFANA_PASSWORD\") is set.")
 	}
 
-	// Jaeger Configuration
-	c.ExternalServices.Jaeger.URL = strings.TrimSpace(getDefaultString(EnvJaegerURL, ""))
-	c.ExternalServices.Jaeger.Service = strings.TrimSpace(getDefaultString(EnvJaegerService, "jaeger-query"))
+	// Tracing Configuration
+	c.ExternalServices.Tracing.EnableJaeger = false
```
jmazzitelli: this is hardcoded to false - it isn't coming from any config setting. a comment here explaining why this is set differently than all the rest (e.g. with getDefaultBoolean) would be helpful.

jmazzitelli:
> 1. The Grafana and Jaeger Routes exists on OC side. Kiali starts and discovers both. Here in Kiali configmap by default the Grafana URL is set, but the Jaeger URL is not. Is finding and setting Grafana URL by operator design? @jmazzitelli

No, the URLs are not going to be set in the configmap if you have not set them in the Kiali CR. The operator will NOT auto-detect anymore... @aljesusg has removed that code from the ansible operator in this PR.

hhovsepy:
> > 1. The Grafana and Jaeger Routes exists on OC side. Kiali starts and discovers both. Here in Kiali configmap by default the Grafana URL is set, but the Jaeger URL is not. Is finding and setting Grafana URL by operator design? @jmazzitelli
> 
> No, the URLs are not going to be set in the configmap if you have not set them in the Kiali CR. The operator will NOT auto-detect anymore... @aljesusg has removed that code from the ansible operator in this PR.

@aljesusg this means that the described issue is in this PR.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1031](https://paste.fedoraproject.org/paste/w4Za2gj32QyfAH380B8L9g)
  * :red_circle: run-kiali-e2e-tests [#[1851]](https://paste.fedoraproject.org/paste/i6z5bf3J7350Bw4XaSCpqQ)

config/config.go
aljesusg: it's hardcoded but change in the discover, this flag let kiali if he should use jaeger or not. I'll add comments

config/config.go
aljesusg: This must be false until discover

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1033](https://paste.fedoraproject.org/paste/rf4x5DC3UJUM~n4RRwdoPw)
  * :red_circle: run-kiali-e2e-tests [#[1854]](https://paste.fedoraproject.org/paste/v4FV5KW6B-uXvS99kS7jUg)

hhovsepy:
Verified:
Grafana and Jaeger are discovered, also when both routes do not exist at first and added later, still discovered.

handlers/grafana.go
jmazzitelli: don't forget to fix this - I don't see it fixed yet in the PR

aljesusg:
Moved @jmazzitelli  I forgot it :( can you approve this again ?

PR #1026, opened 2019-04-22
KIALI-2524 Fix setting headers for audit info

israel-hdez:
Kiali-User was being set to response headers but it is expected to be in the request headers.


rhqci:

PR #1029, opened 2019-04-23
Fix operator role on monitoringdashboards

jotak:
The "list" role is missing in operator for monitoringdashboards.
It was necessary for dashboards auto-discovery

### DIFF
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -218,3 +218,4 @@ rules:
   - monitoringdashboards
   verbs:
   - get
+  - list
--- a/operator/roles/kiali-deploy/templates/kubernetes/clusterrole-viewer.yaml
+++ b/operator/roles/kiali-deploy/templates/kubernetes/clusterrole-viewer.yaml
@@ -122,3 +122,4 @@ rules:
   - monitoringdashboards
   verbs:
   - get
+  - list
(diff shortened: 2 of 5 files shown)


jotak:
@jmazzitelli , also, it looks like the dashboards objects are not deployed anymore along with kiali deployment. Is it done on purpose? I have to manually deploy them.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#961](https://paste.fedoraproject.org/paste/tzaiDFGa22MVeY~7wuIiYQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1775]](https://paste.fedoraproject.org/paste/5JPKnUSmo~NQMFPORJDmBA)

jmazzitelli:
> also, it looks like the dashboards objects are not deployed anymore along with kiali deployment. Is it done on purpose? I have to manually deploy them.

What do you mean "with kiali deployment" ?? These are deployed when you do a "make openshift-deploy"  because that triggers the operator to deploy kiali and you should be getting them installed via the operator.

jotak:
> These are deployed when you do a "make openshift-deploy"
doesn't seem to work for me. I'll check that later, I thought you would have an idea about that.

jmazzitelli:
> doesn't seem to work for me. I'll check that later, I thought you would have an idea about that.

Oh, @jotak - yeah, I forgot about something. There is a bug in upstream operator-sdk - it breaks our installation of the dashboard CRD and thus the dashboards. I posted some messages on this. But suffice it to say, yes, with this bug, it won't work. The bug is fixed upstream, just need to wait for it to be merged and a new operator base-image is released, then we can pull that in.

PR #968, opened 2019-04-03
Discover Jaeger and Grafana URL

aljesusg:
AutoDiscover Jaeger and Grafana URL,

Openshift API provide the data that we need in https//127.0.0.1:8443/apis/route.openshift.io/v1/namespaces/istio-system/routes/jaeger-query

```json
{
  "kind": "Route",
  "apiVersion": "route.openshift.io/v1",
  "metadata": {
    "name": "jaeger-query",
    "namespace": "istio-system",
    "selfLink": "/apis/route.openshift.io/v1/namespaces/istio-system/routes/jaeger-query",
    "uid": "17acdd79-3454-11e9-bdca-54e1ad6bad86",
    "resourceVersion": "8016",
    "creationTimestamp": "2019-02-19T14:38:59Z",
    "labels": {
      "app": "jaeger-query",
      "jaeger-infra": "query-service"
    },
    "annotations": {
      "openshift.io/host.generated": "true"
    }
  },
  "spec": {
    "host": "jaeger-query-istio-system.127.0.0.1.nip.io",
    "to": {
      "kind": "Service",
      "name": "jaeger-query",
      "weight": 100
    },
    "port": {
      "targetPort": "jaeger-query"
    },
    "tls": {
      "termination": "edge"
    },
    "wildcardPolicy": "None"
  },
  "status": {
    "ingress": [
      {
        "host": "jaeger-query-istio-system.127.0.0.1.nip.io",
        "routerName": "router",
        "conditions": [
          {
            "type": "Admitted",
            "status": "True",
            "lastTransitionTime": "2019-02-19T14:38:59Z"
          }
        ],
        "wildcardPolicy": "None"
      }
    ]
  }
}
```

PR #1032, opened 2019-04-23
to support system tests / code coverage

jmazzitelli:
A possible replacement for PR #970 

I have not run the Makefile target or the main_test.go defined here - I simply found them in PR #970 and assumed Jenkins somehow uses them.

But the .sh script I did run with and that did work (even though the pod logged an error because kiali didn't have the systemTests option compiled in it - which I assume is what main_test.go provides)

This is just an example PR. QE can take it and use it, alter it, duplicate it and change it.. whatever is best. I just created this as an example of how we can do this without injecting this systemTest test all throughout the operator.

### DIFF
--- a/deploy/jenkins-ci/deploy-kiali-for-system-tests-code-coverage.sh
+++ b/deploy/jenkins-ci/deploy-kiali-for-system-tests-code-coverage.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+bash <(curl -L https://git.io/getLatestKialiOperator)
+
+echo -n "Waiting for Kiali to start"
+for run in {1..60}
+do
+  oc get pods -l app=kiali -n istio-system 2>/dev/null | grep "^kiali.*Running" > /dev/null && _STARTED=true && break
+  echo -n "."
+  sleep 5
+done
+echo
+
+if [ -z ${_STARTED} ]; then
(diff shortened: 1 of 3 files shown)


deploy/jenkins-ci/deploy-kiali-for-system-tests-code-coverage.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+bash <(curl -L https://git.io/getLatestKialiOperator)
```
jmazzitelli: Note that I just did this "curl" download simply to pick up the released version. This is simply running the stock [operator/deploy/deploy-kiali-operator.sh script](https://github.com/kiali/kiali/blob/master/operator/deploy/deploy-kiali-operator.sh). You can execute that script however you do it normally - if you are running this out of a git cloned repo, you could make this something like `../../operator/deploy/deploy-kiali-operator.sh` instead.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#968](https://paste.fedoraproject.org/paste/N8-OWBDfU0Z-TN010O2lzQ)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1779]](https://paste.fedoraproject.org/paste/FAU9ljeIQqvarh~zkDfKkg)

jmazzitelli:
this isn't needed

Issue #1006, opened 2019-04-13
Support native envoy 


rverma-nikiai:
Aws app mesh also uses envoy proxy similar to Istio. Will it be possible to extend the implementation to support that.
I am willing to send a PR, please let me know. Also would be nice to provide a design doc so that contributors can quickly visualize the components.

abonas:
Can you elaborate more? What does it mean to support native Envoy? Aws App mesh is also mentioned in the description, please clarify.

PR #1048, opened 2019-04-29
KIALI-2677 Assume sidecar exists unless proven otherwise

jshaughn:
- return true when checking sidecar and there are no pods
- set true when checking apps sidecar and there are no workloads
- defer to Workload.IstioSidecar when possible, at times we were re-calling workload.pods.HasIstioSidecar()
- introduce Workload.HasIstioSidecar()
  - optimize for Deployments where we can assume all pods are the same
- fix case consistency: SideCar -> Sidecar
- some clarity changes in health.go

### DIFF
--- a/models/pod.go
+++ b/models/pod.go
@@ -127,15 +127,19 @@ func lookupImage(containerName string, containers []core_v1.Container) string {
 	return ""
 }
 
-func (pods Pods) HasIstioSideCar() bool {
-	for _, pod := range pods {
-		if pod.HasIstioSideCar() {
-			return true
+// HasIstioSidecar returns true if there are no pods or all pods have a sidecar
+func (pods Pods) HasIstioSidecar() bool {
+	if len(pods) > 0 {
+		for _, p := range pods {
+			if !p.HasIstioSidecar() {
+				return false
+			}
 		}
 	}
-	return false
+	return true
 }
 
-func (pod Pod) HasIstioSideCar() bool {
+// HasIstioSidecar returns true if the pod has an Isio proxy sidecar
+func (pod Pod) HasIstioSidecar() bool {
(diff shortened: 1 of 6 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1006](https://paste.fedoraproject.org/paste/ajDHcQu4ibFS20G-Ry~CjA)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1819]](https://paste.fedoraproject.org/paste/a0IUjqvFnpA-nEVYdxtbHg)

models/pod.go
```diff
@@ -127,15 +127,19 @@ func lookupImage(containerName string, containers []core_v1.Container) string {
 	return ""
 }
 
-func (pods Pods) HasIstioSideCar() bool {
-	for _, pod := range pods {
-		if pod.HasIstioSideCar() {
-			return true
+// HasIstioSidecar returns true if there are no pods or all pods have a sidecar
```
jotak: Ok, so the logic is reversed and it took me a while to figure out, at first glance it looked incorrect to me because now we return false if at least one pod has no sidecar. Before we were returning true if there was at least one sidecar.

This is indeed what we want, as this is used to report any missing sidecar in the UI. Perhaps in some places the variables are not named very accurately and this is why it can be confusing. Like "HasSidecar" is actually "HasNoMissingSidecar", that's a slight difference :-)

jshaughn:
@mattmahoneyrh to repro:
Choose a workload (I used bookinfo ratings V1) and cycle to 0 pods
Various detail screens, (Workload, Service) will start to show missing sidecar

Fix: Should not show missing sidecar.  No pods does not mean that a pod, if it existed, would lack the sidecar.  The missing sidecar icon should only show when running pods lack the sidecar.

PR #1027, opened 2019-04-22
KIALI-2524 Fix setting headers for audit info

israel-hdez:
Kiali-User was being set to response headers but it is expected to be in the request headers.

PR #1054, opened 2019-05-06
fix the kiali_cr.yaml file

jmazzitelli:
The kiali_cr.yaml file, though all comments, is a reference document.

The values you see in here are the actual defaults. So people can look up
kiali_cr.yaml and know what the defaults are if the user doesn't set it
(for example, if they use the minimal cr.yaml).

Therefore, this PR is changing the comments in kiali_cr.yaml to fix
the label_selector - the default is not "myLabel=myValue" so this removes
that. The example value is put in the description of the setting which
is where it belongs.

As the top-comment describing this file mentions, "If the setting is not defined by default, its name will be prefixed with "#"" - thus, kiali_cr.yaml will show "#label_selector:" informing the reader that this setting is not set by default.

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -35,13 +35,14 @@ spec:
 #
 # Allows for controlling what namespaces/projects are returned by Kiali.
 #
-# 'exclude' is optional takes an list of namespace/projects to be excluded from the list 
-# of namespaces provided by the API and UI. Regex is supported. This does not affect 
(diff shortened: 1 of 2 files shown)


rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1027](https://paste.fedoraproject.org/paste/ufzfID4Ebrana2ssODYI0g)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1847]](https://paste.fedoraproject.org/paste/3K~Yt6EgekGda1ucx73V-Q)

PR #1058, opened 2019-05-07
KIALI-2820 if secret already exists, don't create one and don't abort

jmazzitelli:


rhqci:
PRT Failed, Please Contact QE

jotak:
Testing OK, works as expected

Issue #1059, opened 2019-05-11
Run kiali behind the proxy application just like kubectl proxy


inyee786:
**Describe the bug**
I have an application which works like a kubeclt proxy, when I run kiali behind the application proxy the UI is not coming up because the `base href` is point to incorrect path

So when i hit the kiali service  url
ie: `https://xyz.com/abc/clustername/api/v1/namespaces/istio-system/services/kiali:80/proxy/kiali/` 

in the index.html file the base tag  `<base href="/api/v1/namespaces/istio-system/services/kiali:80/proxy/kiali/" />` is point the incorrect path

and it can't able to load all the other file(env.js manifest.json etc) because of the incorrect path in `base tag`

**Versions used**
Kiali: 0.16
Istio: 1.1.5
Kubernetes flavour and version: GKE v1.9.7-gke.7

**Expected behavior**

base href tag should point to
`abc/clustername/api/v1/namespaces/istio-system/services/kiali:80/proxy/kiali`

jmazzitelli:
Can you provide your kiali configuration? Specifically, the kiali configmap found in istio-system namespace: `kubectl get configmap kiali -n istio-system -o yaml`

I assume you installed Kiali using the upstream helm chart - if that is correct, what was the command you used to install Kiali and what is the values.yaml (if you changed it?). If you used the Kiali operator, what does your Kiali CR look like (`kubectl get kiali kiali -n kiali-operator -o yaml`).

All of this is to figure out what you set the kiali web_root configuration setting to.

inyee786:
Download the latest binary from istio repo
there was istio-demo.yaml inside `istio-1.1.5/install/kubernetes/istio-demo.yaml` which i applied

kiali config map
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: kiali
  namespace: istio-system
  labels:
    app: kiali
    chart: kiali
    heritage: Tiller
    release: istio
data:
  config.yaml: |
    istio_namespace: istio-system
    server:
      port: 20001
    external_services:
      istio:
        url_service_version: http://istio-pilot:8080/version
      jaeger:
        url: 
      grafana:
        url: 
```

jmazzitelli:
Look in your deployment yaml:

```
 - name: SERVER_WEB_ROOT
   value: /kiali
```

[...]

PR #1060, opened 2019-05-13
KIALI-2836 enhance webroot validation

jmazzitelli:
allow web root validation to support things as mentioned in issue #1059

### DIFF
--- a/kiali_test.go
+++ b/kiali_test.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/kiali/kiali/config"
+)
+
+func TestValidateWebRoot(t *testing.T) {
+	// create a base config that we know is valid
+	conf := config.NewConfig()
+	conf.Server.StaticContentRootDirectory = "."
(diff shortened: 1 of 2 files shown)


kiali_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// now test some web roots, both valid ones and invalid ones
+	validWebRoots := []string{
+		"/",
+		"/kiali",
+		"/abc/clustername/api/v1/namespaces/istio-system/services/kiali:80/proxy/kiali",
+		"/a/0/-/./_/~/!/$/&/'/(/)/*/+/,/;/=/:/@/%aa",
+		"/kiali0-._~!$&'()*+,;=:@%aa",
+	}
+	invalidWebRoots := []string{
```
jotak: not sure, but should we disallow `x/../y` from a security perspective ? (the `..`)

jmazzitelli: I am not sure either, but "just in case", see my latest commit where that is now disallowed and you get an explicit error message if web root is trying to use that "..".

PR #1061, opened 2019-05-13
KIALI-2838 install script can uninstall old kiali

jmazzitelli:
If Kiali is found to already be installed, the script will ask the user if they want it uninstalled. If the user is OK  with that, the script will remove all the old Kiali resources and then move forward with installing the operator and then optionally installing a new Kiali via the Kiali CR.

This also adds a env var option so the user can tell us they are OK with uninstalling any existing kiali without having to be prompted about it.

### DIFF
--- a/operator/deploy/deploy-kiali-operator.sh
+++ b/operator/deploy/deploy-kiali-operator.sh
@@ -350,6 +350,24 @@ else
   echo "Kiali will now be installed."
 fi
 
+# Give the user an opportunity to tell us if they want to uninstall if they did not set the envar yet.
+# The default to the prompt is "yes" because the user will normally want to uninstall an already existing Kiali.
+if [ -z "${UNINSTALL_EXISTING_KIALI}" ]; then
+  if ${CLIENT_EXE} get deployment kiali -n "${NAMESPACE}" > /dev/null 2>&1 ; then
+    read -p 'It appears Kiali has already been installed. Do you want to uninstall it? [Y/n]: ' _yn
+    case "${_yn}" in
+      [yY][eE][sS]|[yY]|"")
+        echo "The existing Kiali will be uninstalled."
+        UNINSTALL_EXISTING_KIALI="true"
+        ;;
+      *)
+        echo "The existing Kiali will NOT be uninstalled."
+        UNINSTALL_EXISTING_KIALI="false"
+        ;;
+    esac
+  fi
+fi
+
 # Some settings specific to Kiali installations
 NAMESPACE="${NAMESPACE:-istio-system}"
 SECRET_NAME="${SECRET_NAME:-kiali}"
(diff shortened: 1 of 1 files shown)


jmazzitelli:
this doesn't provide a prompt yet - I'm adding it now

jmazzitelli:
The latest commit adds the prompt. If the user didn't set the env var telling us what he wants, and there is a Kiali installed already, a prompt will ask the user if they want to uninstall Kiali or not.

jotak:
Thanks, will test it asap

operator/deploy/deploy-kiali-operator.sh
```diff
@@ -328,6 +350,30 @@ else
   echo "Kiali will now be installed."
 fi
 
+# Give the user an opportunity to tell us if they want to uninstall if they did not set the envar yet.
+# The default to the prompt is "yes" because the user will normally want to uninstall an already existing Kiali.
+if [ -z "${UNINSTALL_EXISTING_KIALI}" ]; then
+  if ${CLIENT_EXE} get deployment kiali -n "${NAMESPACE}" > /dev/null 2>&1 ; then
+    read -p 'It appears Kiali has already been installed. Do you want to uninstall it? [Y/n]: ' _yn
```
jotak: So, two questions:
- If it's uninstalled, then I guess the ConfigMap is deleted as well? Maybe we should warn about that?
- If not uninstalled, then what happens exactly? The operator is installed without Kiali CR, or with Kiali CR but it won't create a new deployment or fail to do so?

jmazzitelli: First: Everything goes - configmap included. This is why we have a prompt - to make sure they are OK with deleting it. We could add something to this prompt saying "Do you want to lose all your Kiali configuration and remove Kiali?" - but I was thinking it was intuitively obvious that the user will lose their kiali config when they tell us to uninstall kiali :)

Second: If they do not delete kiali - things may get weird. Because when the operator attempts to install its own kiali, it will attempt to "patch" the existing kiali resources (its the same as if you do a "kubectl apply" IIRC how the ansible k8s tasks work). So this might actually upgrade the existing kiali by accident :)

jmazzitelli: @jotak perhaps we should abort the script with an error if it detects kiali and the user doesn't want to remove it? thoughts?

jmazzitelli: I added a new commit - I don't think the operator should be installed at all if kiali already exists. So the new change is - if kiali exists, and the user tells us they don't want to uninstall it, we abort the script and do not install anything. This way, the user either tells us explicitly they want to uninstall it via the option, or we warn them and they answer the prompt that they are OK with uninstalling it.

PR #1066, opened 2019-05-14
Increase linter deadline 300s -> 500s

jotak:


jotak:
BTW, locally my linting fails although it's working on travis. It seems related to the version of GO (I'm on 1.10.3), linting is stricter with higher version it seems

PR #1067, opened 2019-05-14
KIALI-2856 Jaeger URL autodiscovery fails when is jaeger deployed und…

aljesusg:
https://issues.jboss.org/browse/KIALI-2856

### DIFF
--- a/status/discover.go
+++ b/status/discover.go
@@ -1,7 +1,10 @@
 package status
 
 import (
+	"errors"
 	"io/ioutil"
+
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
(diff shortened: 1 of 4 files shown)


xeviknal:
@aljesusg I can't get it working so far. This is the result:

aljesusg:
Sorry I did a small change , I found a bug here. Can you test it @xeviknal ?

status/discover.go
```diff
@@ -13,6 +16,73 @@ var saToken string
...
+	client, err := getClient()
+
+	if err != nil {
+		return path, err
+	}
+	svc, err := client.GetService(ns, service)
+	if err != nil {
+		return path, err
+	}
+	labelSelectorService := labels.Set(svc.Labels).String()
```
xeviknal: I think the correct way to fetch deployments from a service is fetch it from service selector (svc.Spec.Selector) instead of service labels. 

https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service

status/discover.go
```diff
@@ -13,6 +16,73 @@ var saToken string
...
+	deployments, err := client.GetDeploymentsByLabel(ns, labelSelectorService)
+
+	if err != nil {
+		return path, nil
+	}
+
+	switch len(deployments) {
+	case 0:
+		log.Debugf("Kiali not found a deployment for service %s", service)
```
xeviknal: nitpicking. I would put `Kiali didn't find any deployment for the service %s`

status/discover.go
aljesusg: IF we check the spec we can't find the deployment :(

aljesusg:
@hhovsepy Can we test that this works for maistra and qe too? Thanks  !!!

aljesusg:
Thanks @hhovsepy

PR #1068, opened 2019-05-14
KIALI-2858 Restructure backend graph code for better separation of core and vendor code

jshaughn:
Producing a graph has two main phases:
1. Telemetry processing
2. Generation

Telemetry processing takes a vendor's telemetry and produces a TrafficMap.  The TrafficMap is a Kiali construct that describes all of the traffic (for the time period): nodes, edges, rates, badging, etc..  (current vendors: Istio)

Generation takes the TrafficMap and creates a vendor-consumable configuration that can be rendered by the vendor's tool.  (current vendors: Cytoscape)

This PR mostly relates to step one.  There are a few goals:
- Clearly define the TrafficMap such that it can:
  - Be produced by different telemetry vendors.
  - Be consumed by different generation vendors.
- Separate core code from vendor code
- Clean up some code and simplify packages

To help define the TrafficMap this PR introduces `meta.go`.   This new core file defines the non-protocol keys that can be populated in the Node and Edge metadata.  It also ensures that we can use constants as opposed to string literals when populating the Metadata, which is much more safe.  Protocol.go (which existed prior to the PR) is another core file and is responsible for defining the protocol-specific keys.

It also introduces 2 new interfaces: TelemetryVendor and ConfigVendor, for implementations of the two required APIs for vendors.  The Istio code now implements TelemetryVendor and the cytoscape code now implements ConfigVendor.  These complement the existing Appender interface.

The updated code is currently structure like:
```
/handlers
    graph.go        # rest api handler
/graph
    ...                    # core files
    /api                # entry point api (called from rest api handler)
    /telemetry
        /istio          # istio telemetry impl
    /config
        /cytoscape # cytoscape config impl
```


jshaughn:
@jmazzitelli @israel-hdez , I'd like you guys to take a look at this draft PR, not necessarily the code but perhaps just the description above, and let me know any initial thoughts about whether it's useful, a good approach, etc...

israel-hdez:
Ok, so, not checking the code :)

The TelemetryVendor interface, I think is quite important if we want to support other kind of telemetries (perhaps Istio's telemetry lite, if it manages to be released?). Also, it is good if we opt to support vendors other than Istio, or if community chooses to contribute with their "adapters". +1 to that.

The GenerationVendor I'm not sure if we really need it. I'm not against it, but I'd prefer to see a generic output not tied to any graph control so that the front-end can "transform" it to its need to produce the graph. This is because I would like the backend to not generate any response tied to front-end specific controls. But I'm neutral so, if you think it's worth it, it's quite OK.

I guess most code in handlers/graph.go will be moved to other files, right?

Probably the appenders may also need a similar mechanism of interfacing? Just a though because other data sources may have telemetry with different granularity and may need less or more appenders. This is just a note. Probably it's something to think later.

I'm in completely in favor of using constants for Metadata. But also leave an "open field" for custom unstructured data that other Telemetry vendors may need to use. It also would be a nice way to do experimentation to later evolve to proper metadata fields ;)

jotak:
It looks like a great step forward.
What I can envision next, is to have an generic implementation of TelemetryVendor (so , `GenericTelemetryVendor`) that reads some custom resource of telemetry adapter to know which metric to look for, and build the graph from that. With this architecture we're able to:

- Have some specific implementations, like the Istio one, that is fine-tuned to manage some edge cases that couldn't be managed easily from the generic impl (ex: reporter subtleties)

- Have contributors able to add their adapters using the Generic implementation, without having to submit any PR to Kiali.

josejulio:
> The GenerationVendor I'm not sure if we really need it. I'm not against it, but I'd prefer to see a generic output not tied to any graph control so that the front-end can "transform" it to its need to produce the graph. This is because I would like the backend to not generate any response tied to front-end specific controls. But I'm neutral so, if you think it's worth it, it's quite OK.

This is an interesting idea. Let the client adapt the `TelemetryVendor` to their needs, in this case we could add to kiali-ui a layer to transform it to cytoscape format.

jshaughn:
@israel-hdez 
> The GenerationVendor I'm not sure if we really need it. I'm not against it, but I'd prefer to see a generic output not tied to any graph control so that the front-end can "transform" it to its need to produce the graph. This is because I would like the backend to not generate any response tied to front-end specific controls.

GenerationVendor is not a big restructuring step and the code already had the notion of vendor-specific outputs.  We originally supported vizceral and cytoscape.  We dropped vizceral but the vendor idea is still there.  GenerationVendor just formalizes it with an Interface (1 function).  As for a generic output, the cytoscape output could likely be considered our generic output :)

> I guess most code in handlers/graph.go will be moved to other files, right?

Yes, it is now very thin in the PR, you can take a look.  Actually, it may makes sense to migrate the tests in the handler package to the graph package.

> Probably the appenders may also need a similar mechanism of interfacing?

Yes.  Actually, there is already an Appender interface.

> I'm completely in favor of using constants for Metadata. But also leave an "open field" for custom unstructured data that other Telemetry vendors may need to use.

This is an interesting one.  Until we have a UI/consumer that can deal with that sort of thing I don't think we need it.  But it can be added as soon as the need is realized.  I am sure that as soon as someone tries to actually add a TelemetryVendor it will shake out several things, this is intended as a first step.

jshaughn:
@jotak 
> What I can envision next, is to have an generic implementation of TelemetryVendor (so , GenericTelemetryVendor) 

I'm not sure I completely follow just yet.  Are you saying that we define a set of time-series labels for Prometheus that our generic vendor understands?   And then externally someone populates that metric, points us to it and then we query it?   In other words, a replacement for istio_requests_total that is likely less complex but could still provide enough information to determine nodes and edges, rates, response times, etc... ?

graph/options.go
```diff
@@ -56,16 +54,17 @@ type VendorOptions struct {
 // Options are all supported graph generation options.
 type Options struct {
 	AccessibleNamespaces map[string]time.Time
-	Appenders            []appender.Appender
```
jmazzitelli: I still think you should keep this appender.Appender, even if it is only a marker interface now.. in the future, it will provide a place to add things when needed. It also makes it more strictly typed (now it is back to just string, which is too generic for my taste).

jshaughn: The thing is that this option is no longer an array of Appenders but rather Appender names.  Just the appenders requested via query params.  The appender names are then passed to the actual telemetry vendor (i.e. istio) where they actually get converted into the actual Appender instances supplied by the vendor impl.  I could make this more clear by adding something like:

type AppenderName string

Is that what you are looking for?

jmazzitelli: `type AppenderName string`

yes, that would be better

jshaughn:
I feel like this improves the code structure for graph generation while stopping short of actually adding a new vendor. I am sure adding a vendor will shake out more structural changes, but this at least moves us down the path.  So, I am promoting the draft PR and asking some folks for a code-level review.

jotak:
> @jotak
> 
> > What I can envision next, is to have an generic implementation of TelemetryVendor (so , GenericTelemetryVendor)
> 
> I'm not sure I completely follow just yet. Are you saying that we define a set of time-series labels for Prometheus that our generic vendor understands? And then externally someone populates that metric, points us to it and then we query it? In other words, a replacement for istio_requests_total that is likely less complex but could still provide enough information to determine nodes and edges, rates, response times, etc... ?

Not exactly, not our own metric being put in Prometheus, but more an adapter pattern to be able to fetch metrics from other providers. Actually, it would be close to what you're building with this Vendoring interface, except that we understand which metric we have to fetch not from hard-coded implementation, but from yaml resource. And for that to work, we would implement a "generic vendor" that reads these resources.

graph/api/api.go
```diff
@@ -0,0 +1,79 @@
+package api
```
israel-hdez: Personally, I would have used the "business" folder for this code.
No action really needed, it is just my preference.

jshaughn:
@jmazzitelli Do you have any additional comments before a merge?

jmazzitelli:
> @jmazzitelli Do you have any additional comments before a merge?

no

jotak:
So ... now the SMI spec is out, we can say that the `GenericTelemetryVendor` I was talking about would be more or less the implementation of the SMI metrics specs :-)

jshaughn:
Yes, time to look more closely at that and perhaps just start an impl based on that API, updating as it evolves.

PR #1074, opened 2019-05-16
KIALI-2877 Remove outdated documentation in config.yaml

israel-hdez:
Reported in https://github.com/kiali/kiali/issues/1059#issuecomment-492785576.

Support for credentials in config.yaml was removed in PR #883. So this looks safe to remove, since credentials in config.yaml are no longer honored.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -2,13 +2,6 @@ server:
   address: localhost
   port: 8000
 
-  # Uncomment credentials to enable basic authentication for server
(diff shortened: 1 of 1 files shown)


jmazzitelli:
FWIW: there is other things wrong with this file :)

like the prometheus URL setting is wrong now. I'm not sure what else is. This is an obsolete file - i don't think anyone is using it or the make run target. If it is being used, someone should go through and confirm it all works.

PR #1075, opened 2019-05-17
KIALI-2759 Don't show workloads when service doesn't have selector labels

xeviknal:
** Describe the change **

Don't show workloads for a Service that doesn't have selector labels. Also, Istio sidecar shouldn't be computer either bc there aren't pods running under its umbrella.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2759

** Backwards incompatible? **
yes

** Screenshoots **
Before:
Screenshot of Kiali Console (38)

After:
Screenshot of Kiali Console (39)

Test data: https://issues.jboss.org/browse/KIALI-2759?focusedCommentId=13734911&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13734911

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -120,19 +120,25 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime 
 	var nsmtls models.MTLSStatus
 
 	wg := sync.WaitGroup{}
-	wg.Add(9)
+	wg.Add(7)
 	errChan := make(chan error, 6)
 
 	labelsSelector := labels.Set(svc.Spec.Selector).String()
+	// If service doesn't have any selector, we can't know which are the pods and workloads applying.
+	if labelsSelector != "" {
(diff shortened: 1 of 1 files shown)


business/services.go
```diff
@@ -120,19 +120,25 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime
...
 
-	go func() {
-		defer wg.Done()
-		var err2 error
-		pods, err2 = in.k8s.GetPods(namespace, labelsSelector)
-		if err2 != nil {
-			errChan <- err2
-		}
-	}()
+	if labelsSelector != "" {
```
lucasponce: Minor comment, perhaps the goroutine for GetPods and fetchWorkloads can be under same check of labelSelector.
But it's Ok.

PR #1076, opened 2019-05-17
KIALI-2880 KIALI-2879 Modify the FQDN parsing

burmanm:
** Describe the change **

Modify the FQDN parsing to parse ServiceEntries to fqdn.Service part only, as the SEs are global scoped and do not adhere to the cluster parsing otherwise

This simplifies our code in some parts and makes it hopefully more reliable when catching SEs. Also, change certain parsing to use prefix matching instead of only "*" to avoid false positives.

Same PR adds support for wildcard ServiceEntries, which would in certain cases fail otherwise.

** Issue reference **

KIALI-2879, KIALI-2880

** Backwards incompatible? **

Yes.

### DIFF
--- a/business/checkers/destinationrules/multi_match_checker.go
+++ b/business/checkers/destinationrules/multi_match_checker.go
@@ -1,6 +1,8 @@
 package destinationrules
 
 import (
+	"strings"
+
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 )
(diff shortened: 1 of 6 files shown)


hhovsepy:
"gateway.namespace" format is not supported anymore?
Screenshot from 2019-05-21 10-54-44

burmanm:
It has never been.

hhovsepy:
Then what about this in master?
Screenshot from 2019-05-21 16-55-43

hhovsepy:
And what are the expected functionalities by this PR?

burmanm:
``servicename.namespace`` working is previously was just an accident (a problem in parsing the FQDN), it didn't work everywhere. It's not a documented supported format by Istio and thus shouldn't be used. Some Istio components use format such as ``servicename/namespace``, thus it is unwise to use any non-official format.

The change is to catch few edge cases of ServiceEntries, since they support wildcards also - but must be different from Service wildcards. With old master, you can create a ServiceEntry that will match "*" for example for ServiceName - and that's wrong.

burmanm:
I was going to create the additional tests with (upcoming) KIALI-2880 PR.

burmanm:
So for now, if everything works correctly (FQDN, short names, ServiceEntry domain names etc) then that's the proper use. So ``google.com`` should be ``google.com`` and not service: google, namespace: com.

burmanm:
Added KIALI-2880 also here and some tests to, to make sure we can support ServiceEntries with wildcards also. This PR hopefully now addresses all the edge cases with ServiceEntries (other than visibility in cluster).

PR #1082, opened 2019-05-20
KIALI-2892 Jaeger traces not work when the user use Istio upstream with a QUERY_BASE_PATH

aljesusg:
Detected some problems with QUERY_BASE_PATh and jaeger.

- We are checking the services of jaeger to check if is available or not before the autodiscover so if the user not set the service name of tracing service jaegerAvailable will be false (https://github.com/kiali/kiali/blob/master/business/jaeger_helper.go#L29) and errorTraces will be -1 (Fix with https://github.com/kiali/kiali/pull/1082/files#diff-c19e5972b564f8f7ebd30c662b1974cbR104)

- If user is working with istio community the correct patch to check the services is `<endpoint>/<QUERY_BASE_PATH>/api` so we added a new param in configuration called Path to control this.

Thanks @hhovsepy

### DIFF
--- a/business/jaeger_helper.go
+++ b/business/jaeger_helper.go
@@ -36,7 +36,7 @@ func getErrorTracesFromJaeger(namespace string, service string) (errorTraces int
 		return -1, errors.New("jaeger is not available")
 	}
 	if config.Get().ExternalServices.Tracing.EnableJaeger {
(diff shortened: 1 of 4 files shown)


hhovsepy:
It worked on Istio upstream, will do regression check on Maistra.

PR #1083, opened 2019-05-21
the time format is consistent with the MarshalJSON of k8s

stone4774:
** Describe the change **

kiali's time format: 2019-05-06T10:56:58+08:00
k8s's time format: 2019-05-06T02:56:58Z

kiali's time format should be consistent with k8s, and UTC should be used by default.

### DIFF
--- a/models/pod_test.go
+++ b/models/pod_test.go
@@ -39,7 +39,7 @@ func TestPodFullyParsing(t *testing.T) {
 	pod := Pod{}
 	pod.Parse(&k8sPod)
 	assert.Equal("details-v1-3618568057-dnkjp", pod.Name)
-	assert.Equal("2018-03-08T17:44:00+03:00", pod.CreatedAt)
+	assert.Equal("2018-03-08T14:44:00Z", pod.CreatedAt)
 	assert.Equal(map[string]string{"apps": "details", "version": "v1"}, pod.Labels)
 	assert.Equal([]Reference{Reference{Name: "details-v1-3618568057", Kind: "ReplicaSet"}}, pod.CreatedBy)
 	assert.Len(pod.IstioContainers, 1)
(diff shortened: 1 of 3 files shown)


pilhuhn:
@stone4774 Can you please describe the underlying issue?
I think - even if k8s is doing that - the Zulu time is not well known outside of military

stone4774:
Kiali currently mainly manages istio resources, in addition, we need to manage k8s resources, such as creating services, deployment and so on; so the client gets different time formats, which is very troublesome to handle.

jotak:
I would have imagined users want to see the times for their timezone, no?
@kiali/ux-reviewers , any recommendation?

stone4774:
I don't think we should pay attention to time zones in APIs. Time zones should be converted and displayed by UI. Usually, only the display side knows its own time zone.

jotak:
@stone4774 Is it because you're consuming the REST API directly?

jotak:
So @stone4774 I guess your argument is good but there must be a related change on the UI as well, no?

stone4774:
I'm developing a micro service management platform; the backend loads kiali as a package; we implement our requirements by calling kiali's internal methods; therefore, we hope that kiali's functions can be designed with smaller granularity and more convenient for us to compose.

burmanm:
I agree with @stone4774. The "+08:00" is not sane in an API, it doesn't even mention +8 to what. To UTC? To GMT? To something else? With daylight changes? Which region daylight?

Applications should never use anything else than UTC.

stone4774:
> So @stone4774 I guess your argument is good but there must be a related change on the UI as well, no?

createdAt: <LocalTime time={service.createdAt} />
In kiali-ui, most of the time is displayed in this way, so the change of time format has no effect on the current UI display.

jotak:
> I agree with @stone4774. The "+08:00" is not sane in an API, it doesn't even mention +8 to what. To UTC? To GMT? To something else? With daylight changes? Which region daylight?
> 
> Applications should never use anything else than UTC.

Yes, +1
I didn't realize at first that we were talking about the rest api and not the UI.

jotak:
@stone4774 some unit tests are failing can you take care of that?

```
--- FAIL: TestPodFullyParsing (0.00s)
	pod_test.go:42: 
			Error Trace:	pod_test.go:42
			Error:      	Not equal: 
			            	expected: "2018-03-08T17:44:00+03:00"
			            	actual  : "2018-03-08T14:44:00Z"
			Test:       	TestPodFullyParsing

```

serenamarie125:
I think we should remain consistent with OpenShift.  They use elapsed time in list views.  On details pages they have a tooltip as well 
Screen Shot 2019-05-21 at 9 07 40 PM

It may be best for us to create a UX proposal to remain consistent with OpenShift.  what do you think?

jotak:
@serenamarie125 sorry for the noise, I misunderstood this PR as it was only impacting the REST API and not the UI. So I'm going to merge anyway.

BTW, to be honest I prefer the current way times are displayed in Kiali compared to OpenShift: in OS I have the english am/pm format, it seems to ignore my locale (which is French / 24hrs format). (speaking about openshift 3.x , i haven't checked for os4)

PR #1087, opened 2019-05-22
KIALI-2370 Add validation for gateway selector

burmanm:
** Describe the change **

Gateway object's selector should a match a workload from the same namespace as per Istio documentation. This validation adds that, but also includes a workaround for "istio-ingressgateway" and "istio-egressgateway" which do not match Istio's documentation.

Marking as warning instead of error because older versions did not have this requirement in their documentations. They stated that selector can match any pod in any namespace in Kubernetes, but this has been changed in Istio 1.1 documentation with a note that it will become mandatory to match same namespace.

** Issue reference **

KIALI-2370

### DIFF
--- a/business/checkers/gateway_checker.go
+++ b/business/checkers/gateway_checker.go
@@ -11,6 +11,7 @@ const GatewayCheckerType = "gateway"
 type GatewayChecker struct {
 	GatewaysPerNamespace [][]kubernetes.IstioObject
 	Namespace            string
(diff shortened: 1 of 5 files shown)


hhovsepy:
Verified:
Screenshot from 2019-05-24 19-05-36
Screenshot from 2019-05-24 19-05-50
Screenshot from 2019-05-24 19-08-00

PR #1090, opened 2019-05-24
Update the base url to run kiali in relative path

inyee786:
** Describe the change **

 Update the base url to take the relative url not the absolute path
  while running the kiali under reverse proxy for the url like
 `abc.com/xyz/def/api/v1/namespaces/istio-system/services/kiali:20001/proxy/`
   it takes the wrong base url URL `abc.com/api/v1/namespaces/istio-system/services/kiali:20001/proxy/`

 To solve this I have taken the relative URL and it starts taking the relative url 

** Issue reference ** 
 https://github.com/kiali/kiali/issues/1059

** Backwards incompatible? **

  NO

### DIFF
--- a/kiali.go
+++ b/kiali.go
@@ -184,13 +184,10 @@ func validateConfig() error {
 		return fmt.Errorf("server static content root directory does not exist: %v", config.Get().Server.StaticContentRootDirectory)
 	}
 
-	validPathRegEx := regexp.MustCompile(`^\/[a-zA-Z0-9\-\._~!\$&\'()\*\+\,;=:@%/]*$`)
+	validPathRegEx := regexp.MustCompile(`^[a-zA-Z0-9\-\._~!\$&\'()\*\+\,;=:@%/]*$`)
 	webRoot := config.Get().Server.WebRoot
 	if !validPathRegEx.MatchString(webRoot) {
 		return fmt.Errorf("web root must begin with a / and contain valid URL path characters: %v", webRoot)
-	}
-	if webRoot != "/" && strings.HasSuffix(webRoot, "/") {
-		return fmt.Errorf("web root must not contain a trailing /: %v", webRoot)
 	}
 	if strings.Contains(webRoot, "/../") {
 		return fmt.Errorf("for security purposes, web root must not contain '/../': %v", webRoot)
@@ -268,7 +265,7 @@ func updateBaseURL(webRootPath string) {
 	html := string(b)
 
 	searchStr := `<base href="/"`
-	newStr := `<base href="` + webRootPath + `/"`
+	newStr := `<base href="` + webRootPath + `"`
 	newHTML := strings.Replace(html, searchStr, newStr, -1)
 	if html != newHTML && strings.Contains(newHTML, newStr) {
 		log.Debugf("Base URL has been updated to [%v]", newStr)
--- a/kiali_test.go
+++ b/kiali_test.go
@@ -16,17 +16,19 @@ func TestValidateWebRoot(t *testing.T) {
 	validWebRoots := []string{
 		"/",
 		"/kiali",
+		"/kiali/",
+		"./kiali",
 		"/abc/clustername/api/v1/namespaces/istio-system/services/kiali:80/proxy/kiali",
 		"/a/0/-/./_/~/!/$/&/'/(/)/*/+/,/;/=/:/@/%aa",
 		"/kiali0-._~!$&'()*+,;=:@%aa",
 	}
 	invalidWebRoots := []string{
-		"/kiali/",
-		"kiali/",
+		"/../",
+		"/../bar",
 		"/^kiali",
 		"/foo/../bar",
 		"/../bar",
-		"../bar",
+		"/../bar/..",
 	}
 
 	for _, webroot := range validWebRoots {


jmazzitelli:
This needs alot of testing before merge. IIRC, this base URL starting with "/" is assumed in some code. I think we assume that for a reason, but I can't remember what.

So, we need very good testing here - I'm not 100% sure putting "." in front is always going to work.

israel-hdez:
I'm not sure how to make this work.
The only what I manage to get is a blank page, because javascript files don't load correctly.

Of course, I don't have a reverse proxy. But it should work OK with or without a reverse proxy.
Perhaps, could you provide how to setup the reverse proxy case? We could test your with setup and also ensure it works without the reverse proxy.

inyee786:
@jmazzitelli  @israel-hdez  The reverse proxy is our private project, 
let me know what are the test case you want to perform, I will provide the test case result in full details,

I will talk to my team about how to set up a reverse proxy with our private project.

israel-hdez:
I think the regex change is a better approach, since this PR breaks a "simple" development setup.

inyee786:
Great @jmazzitelli, I will rest the change

rhqci:

kiali.go
```diff
@@ -184,14 +184,11 @@ func validateConfig() error {
 		return fmt.Errorf("server static content root directory does not exist: %v", config.Get().Server.StaticContentRootDirectory)
 	}
 
-	validPathRegEx := regexp.MustCompile(`^\/[a-zA-Z0-9\-\._~!\$&\'()\*\+\,;=:@%/]*$`)
+	validPathRegEx := regexp.MustCompile(`^[a-zA-Z0-9\-\._~!\$&\'()\*\+\,;=:@%/]*$`)
 	webRoot := config.Get().Server.WebRoot
 	if !validPathRegEx.MatchString(webRoot) {
 		return fmt.Errorf("web root must begin with a / and contain valid URL path characters: %v", webRoot)
 	}
-	if webRoot != "/" && strings.HasSuffix(webRoot, "/") {
```
inyee786: @israel-hdez  and @jmazzitelli I am removing this check as for reverse proxy '/' is not needed

jmazzitelli: @inyee786 Are you saying that the reverse proxy CAN have a trailing slash `/` at the end of the configured web_root?

This check ensures that `web_root` never ends with a slash `/` character. We have seen this break in the past when users append "/" to the end of web root (that is why this check is there). In other words, we have seen it where users set the web root to things like `/kiali/` and the system breaks (something about the request processor - I forget the reason why).

So please check other use-cases (other than your reverse proxy use-case) to make sure that change will not break things. Specifically, you will want to test that web_root="/kiali/" works. If it does not work but we still want to allow this (for those with reverse proxies that end the web_root with a `/`) we will  need to log a WARN message here to let people know about potential issues.

Note also that the operator will strip any ending slash that is found in web_root - see: https://github.com/kiali/kiali/blob/master/operator/roles/kiali-deploy/tasks/main.yml#L86-L91

So if we are to change this behavior, we will need the operator to log a debug message here and not change web_root if, in fact, we can allow web_root to end with a slash. So that operator ansible script will need to change as well.

inyee786: No, I am saying that it should not contain trailing slash `/` at the end of webroot but
here 
https://github.com/kiali/kiali/blob/b8fc8fe782042fea27db363c3a1d16b9ce4d3757/kiali.go#L271  it getting updated with `/` at the end

jmazzitelli: > No, I am saying that it should not contain trailing slash `/` at the end of webroot but here

Yes, that is what this code does. I am confused then. This code:

```
	if webRoot != "/" && strings.HasSuffix(webRoot, "/") {
```

Is merely confirming that the user did not configure web_root with the trailing slash. Thus eliminating the possibility that we end up with `<base href=/base_root_here//`> (notice the duplicate trailing slash). What is the purpose of removing that check `strings.HasSuffix(webRoot, "/")` ? It sounds like the reverse proxy wants the same as the normal root configurations (where the user should not put a trailing slash because Kiali will add one for you - this eliminates duplicate trailing slashes).

kiali.go
```diff
@@ -268,7 +265,7 @@ func updateBaseURL(webRootPath string) {
 	html := string(b)
 
 	searchStr := `<base href="/"`
-	newStr := `<base href="` + webRootPath + `/"`
+	newStr := `<base href="` + webRootPath + `"`
```
jmazzitelli: Oh, I see... you are removing the trailing slash here. We need to think about this some more, because I am wondering why the original developer added that  trailing slash here. Clearly it was needed.

We just need to confirm the normal use cases work in addition to the reverse proxy before merging this PR.

Did you run with web_root as `/` and `/kiali` and see if that works?

inyee786: with `/` i had tested it, let me test it with `/kiali` and  will let you know

mtho11:
Where does this stand?

jmazzitelli:
I just tested this PR - it does not work. I installed on OpenShift with web_root set to `/` (which is the default) and it works. I then changed web_root to be `/kiali` and it fails to serve the UI properly to the browser. I then set it back to `/` and it works fine.

So this PR does break the current behavior.

PR #1092, opened 2019-05-24
Change label and make text more explicit

pilhuhn:
Change the stalebot label from wontfix to stale


israel-hdez:
I think it is fine... We will need to manually change the wontfix label to the stale label.
I'll do that once this is merged.

PR #1095, opened 2019-05-27
KIALI-2850 Disable displaying tracing information in the UI

aljesusg:
JIRA: [KIALI-2850]( https://issues.jboss.org/browse/KIALI-2850)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -54,6 +54,7 @@ const (
 	EnvGrafanaUsername                 = "GRAFANA_USERNAME"
 	EnvGrafanaPassword                 = "GRAFANA_PASSWORD"
 
+	EnvTracingEnabled          = "TRACING_ENABLED"
 	EnvTracingURL              = "TRACING_URL"
 	EnvTracingServiceNamespace = "TRACING_SERVICE_NAMESPACE"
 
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -41,6 +41,7 @@ const (
 	EnvServerMetricsPort                = "SERVER_METRICS_PORT"
 	EnvServerMetricsEnabled             = "SERVER_METRICS_ENABLED"
 
+	EnvGrafanaEnabled                  = "GRAFANA_ENABLED"
```
jotak: The `EnvGrafanaDisplayLink` already does pretty much the same, so I think either don't add the new one or remove the old one.

aljesusg: Removed. Thanks @jotak

aljesusg:
Now the information is removed in /api/status when is disabled

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -191,12 +191,14 @@ spec:
 #      url: ""
 # **Tracing-specific settings:
 #  - Right now we only support Jaeger
+# enabled: When true, Kiali shows Jaeger and autodiscover
```
jmazzitelli: Trivial - but can we reword this slightly? The sentence seems to be cut off here. Say something like:

```suggestion
# enabled: When true, Kiali shows Jaeger and will attempt to autodiscover it.
```

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -191,12 +191,14 @@ spec:
...
 # namespace: The Kubernetes namespace that holds the Tracing service (if empty, assumes the same value as deployment.namespace)
 # service: The Kubernetes service name for tracing. Kiali uses this to connect within the cluster to Jaeger.
 # url: The URL that Kiali uses when integrating with Tracing. This URL must be accessible to clients external to
 #      the cluster in order for the integration to work properly. If empty, an attempt to auto-discover it is made.
 #    ---
 #    tracing:
+#      enabled: "true"
```
jmazzitelli: looking at the defaults/main.yml, the default is actually `true` not `"true"` (actual boolean, not string). So this should be:

```suggestion
#      enabled: true
```

aljesusg: so display_link should be boolean too. I found it strange to see it

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -58,6 +58,7 @@ kiali_defaults:
       custom_metrics_url: ""
       url: ""
     tracing:
+      enabled: true
```
jmazzitelli: Just a question: do you really want `true` here and not the string `"true"`? - see my comment above

aljesusg:
done @jmazzitelli

Issue #865, opened 2019-02-21
Kiali assumes Grafana is cluster internal, in instio-system namespace and has k8s service named grafana


iftachsc:
**Describe the bug**
I've set Grafana url, user , and password for an external grafana where we send all metrics.
kiali it self has all its metrics from a federated prometheus inside the cluster but we dont hold a grafana there.
everything works. kiali generates correct links and send me to grafana. only thing is that kiali still complain on welcome screen with the following:

![image](https://user-images.githubusercontent.com/28829551/53188883-ede28c00-360e-11e9-9ef0-ccb4fcd9b803.png)

**Versions used**
Kiali: 0.15
Istio: 1.0.3-gke.3
GKE

**To Reproduce**
i think its related ot any installation where you set grafana url, user and passowrd correctly but its external and doesnt sit in side istio-system and has k8s service named grafana

**Expected behavior**
if kiali ables to connect succesfully to the grafana set in grafana.url configuration - dont try to find a k8s service named grafana.

I currently fix this with defining a fake externalName k8s service with the ip of the grafana but i think i could set any ip there.

jotak:
@iftachsc thanks for reporting, I'll have a look

jotak:
@iftachsc , I have some questions for you. Actually I don't really understand how Kiali would generate correct links to grafana if it doesn't have access to its http API.

For instance the link displayed here:

![sshot](https://screenshotscdn.firefoxusercontent.com/images/95dbcce0-8108-4aaa-aa32-3f9cd2f30175.png)
Looks like this:
`http://(grafana_server)/d/UbsSZTDik/istio-workload-dashboard?var-namespace=myns&var-workload=myworkload`

The *UbsSZTDik* part is not hard-written, Kiali has to query grafana's rest api to get that dashboard ID.
Note that older versions of Grafana did not have this generated part.

So, I don't see how you can get valid links without that connection between Kiali and Grafana, and that connection is currently only performed with in-cluster queries. Do you have an explanation for that.

Doing the same with external Grafana would necessitate more plumbing with egress definition and so on, which isn't something that we cover now.

Maybe what we could do, however, is when a Grafana url is configured but is inaccessible from within kiali, have a fallback scenario to show that url but without pointing to the exact service/workload dashboard, as we cannot guess its ID.

iftachsc:
i do have a connection. but kiali complains regardless. i have no issue between kiali and grafana. 
[...]

PR #1096, opened 2019-05-27
KIALI-2920 Make Grafana off-cluster still accessible

jotak:
- Also fix KIALI-2921 for Grafana / potential race condition while setting config.
- Add a new config "InCluster" to tell if Grafana is in cluster or not
- Factorize code for access to service account token, and fix test failing while running that code

JIRA: https://issues.jboss.org/browse/KIALI-2920
See also #865

### DIFF
--- a/business/jaeger_helper.go
+++ b/business/jaeger_helper.go
@@ -4,13 +4,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/kiali/kiali/log"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"time"
 
 	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/log"
 )
 
 type Trace struct {
--- a/config/config.go
+++ b/config/config.go
@@ -42,6 +42,7 @@ const (
 	EnvServerMetricsEnabled             = "SERVER_METRICS_ENABLED"
 
 	EnvGrafanaDisplayLink              = "GRAFANA_DISPLAY_LINK"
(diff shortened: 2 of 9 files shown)


jotak:
To test Grafana not in cluster, we now have to explicitly set the config `in_cluster` to `false`:

```yaml
  external_services:
    grafana:
      in_cluster: false
      url: "http://my-grafana-server:3000/"
```

jotak:
@hhovsepy between 2 and 3, the same URL is used for link generation, however when it's in-cluster, Kiali uses the cluster domain name (with "http://service.namespace") to connect to the grafana API instead of the external URL. This is done to avoid issues like when Kiali is not able to reach the outside.

hhovsepy:
> Verified:
> 
>     1. The default case when url is not provided and in_cluster=true. Grafana is recognised.
> 
>     2. URL is provided off-cluster and in_cluster=true. Grafna URL is set correctly.
> 
>     3. URL is provided off-cluster and in_cluster=false. Still uses that URL. Is this the expected behavior?
> 
>     4. Wrong URL and in_cluster=false. Error message is shown.

@jotak sorry I had commented wrongly here, should be:

2. URL is provided to off-cluster and in_cluster=false. Grafna URL is set correctly.
3. URL is provided to off-cluster and in_cluster=true. Still uses that off-cluster provided URL. Is this the expected behavior? Or it should set the internal URL?

hhovsepy:
4.  in_cluster: true, url: 'off-cluster url' , in-cluster Grafana is down:  the off-cluster URL is set.

PR #1097, opened 2019-05-27
KIALI-2902 Missing tracing tab

aljesusg:
Fix control in kiali when tracing is not enabled

Related with: [KIALI-2902](https://issues.jboss.org/browse/KIALI-2902)
cc @mwringe


lucasponce:
Hi @mwringe, you were commenting about this PR in the grooming call, right ?
Do we want to consolidate this and continue with other separate PR or add here the suggestions ?

aljesusg:
we must merge this @lucasponce @mwringe we can discuss in another PR about what happens in his environment but when the user disable jaeger kiali shouldn't try to discover it.

PR #1056, opened 2019-05-06
KIALI-2804 [operator] restart the kiali pod if the configmap changed

jmazzitelli:

PR #1101, opened 2019-05-29
KIALI-2921 Fix possible race on config write

jotak:
Introduce 'appstate' package, which can hold states more domain-specific, in order to avoid writes on global config.
The idea is to try to not mutate config (exception for kiali main and tests), hence avoiding concurrent writes.

Also try to clarify a couple of things:
- Renaming business.GetServices to business.GetJaegerServices
- In k8s client GetRoute, clarify that argument is route name and not service name (although they match most of the time)
- More consistent logging in discovery function

JIRA: https://issues.jboss.org/browse/KIALI-2921

### DIFF
--- a/appstate/jaeger_state.go
+++ b/appstate/jaeger_state.go
@@ -1,0 +1,12 @@
+package appstate
+
+import "github.com/kiali/kiali/config"
+
+// JaegerEnabled is a memo-flag to tell if Jaeger was configured or succesfully discovered
+var JaegerEnabled = false
+
+// JaegerAvailable tells if Jaeger is ready
+var JaegerAvailable = true
+
+// JaegerConfig is a copy of config.TracingConfig from global config, that can be mutated (e.g. for URL discovery)
+var JaegerConfig config.TracingConfig
--- a/kiali.go
+++ b/kiali.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/golang/glog"
 
+	"github.com/kiali/kiali/appstate"
 	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/config/security"
@@ -103,9 +104,9 @@ func main() {
 	// we need first discover Jaeger
 	if config.Get().ExternalServices.Tracing.Enabled {
 		status.DiscoverJaeger()
-		_, err := business.GetServices()
+		_, err := business.GetJaegerServices()
 		if err != nil {
-			business.JaegerAvailable = false
+			appstate.JaegerAvailable = false
 			log.Errorf("Jaeger is not available : %s", err)
 		}
 	}
--- a/status/discover.go
+++ b/status/discover.go
@@ -5,6 +5,8 @@ import (
 	"net/url"
 	"strings"
 
+	"github.com/kiali/kiali/appstate"
+
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/kiali/kiali/config"
@@ -12,7 +14,8 @@ import (
 	"github.com/kiali/kiali/log"
 )
 
-var grafanaDiscoveredURL string
+// Route names to lookup for discovery
+var tracingLookupRoutes = [...]string{"tracing", "jaeger-query"}
 
 var clientFactory kubernetes.ClientFactory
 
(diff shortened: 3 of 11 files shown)


appstate/jaeger_state.go
```diff
@@ -0,0 +1,12 @@
+package appstate
+
+import "github.com/kiali/kiali/config"
+
+// JaegerEnabled is a memo-flag to tell if Jaeger was configured or succesfully discovered
+var JaegerEnabled = false
```
jotak: @aljesusg Currently I see we have 3 different flags that serve similar purposes with however some slight differences: there's `JaegerEnabled`, `JaegerAvailable` and `TracingConfig.Enable`. I wonder if we could simplify that.

status/discover.go
```diff
@@ -88,119 +91,110 @@ func getPathURL(endpoint string) (path string) {
...
 
-func checkTracingService() (url string, err error) {
-	conf := config.Get()
-	tracing := config.TracingDefaultService
-	jaeger := config.JaegerQueryDefaultService
-	ns := config.IstioDefaultNamespace
-	service := tracing
+func checkTracingService() (url string) {
+	tracingConfig := config.Get().ExternalServices.Tracing
+	ns := config.IstioDefaultNamespace // TODO: don't use default!
```
jotak: @aljesusg I didn't change it because I'm not 100% sure, but it seems wrong to me to use `IstioDefaultNamespace` instead of the configured istio namespace. WDYT? Shouldn't we use `config.IstioNamespace` instead?

kiali.go
```diff
@@ -103,9 +104,9 @@ func main() {
 	// we need first discover Jaeger
 	if config.Get().ExternalServices.Tracing.Enabled {
 		status.DiscoverJaeger()
-		_, err := business.GetServices()
+		_, err := business.GetJaegerServices()
 		if err != nil {
-			business.JaegerAvailable = false
+			appstate.JaegerAvailable = false
```
jotak: @aljesusg I think we should improve this , if I understand correctly here, when Jaeger cannot be reached at Kiali startup, it's definitely turned off. One would have to restart Kiali to get this `JaegerAvailable` flag to true again.
This happened to me while testing, the call to Jaeger ended in timeout [1] so `JaegerAvailable` was set to false even if it was available again shortly after.
Maybe this check to jaeger availability shouldn't be done just once at startup, WDYT?

[1] logs: `E0529 09:17:05.520601       1 kiali.go:109] Jaeger is not available : Get http://tracing/jaeger/api/services:  net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers)`

appstate/jaeger_state.go
aljesusg: yes, @jotak I know.... I don't know how added JaegerEnabled...let me introduce the flags:

## Enabled

User set this in the configuration to enable or disable jaeger in the UI (like displaylink of grafana)

## jaegerAvailable

This is a internal configuration for us, in the discover step we check if kiali discovered the service and endpoint or the user set this information, if all is ok this variable will be true. So if there is a service with route all is conmfigured.

## jaegerEnabled 

This variable is false if kiali can't reach http://<jaeger>/api/services in business.GetServices().

So **jaegerEnabled** check if the application is available and up and **jaegerAvailable** if there is a tracing service with a route defined.

Here is the problem. jaegerEnabled is ok because the user can have the correct setting in his environment but for some reason jaeger is not available, for example elasticsearch is wrong. Should we warn the user about this? there is so use cases here.

status/discover.go
aljesusg: Yes, matt told me about we should use the namespace where is deployed the environment.
Here is the line that I saw and I failed ^^
```go
c.IstioNamespace = strings.TrimSpace(getDefaultString(EnvIstioNamespace, IstioDefaultNamespace))
```

so it should be like

```go
c.ExternalServices.Tracing.Namespace = strings.TrimSpace(getDefaultString(EnvTracingServiceNamespace, c.IstioNamespace))
```

instead

```go
c.ExternalServices.Tracing.Namespace = strings.TrimSpace(getDefaultString(EnvTracingServiceNamespace, IstioDefaultNamespace))
```

?

kiali.go
aljesusg: This is why I proposed about the refresh endpoint to avoid restart pod and reset the discover, I talked about this with @lucasponce .

status/discover.go
jotak: Yes I guess, something like that.. is there already a JIRA about that? or I can change it from this PR

kiali.go
jotak: Cool, ok!
Is there already a JIRA?

aljesusg: no, because we decided not do it (Discussion: https://github.com/kiali/kiali/pull/1022/files#r277584377 You were in discussion ) , but I am +1 in do it..... @lucasponce  ?? ^^.

I think that the problem here is when there are more than 1 pod and you want to refresh...

jotak: That discussion was about the `JaegerEnabled` one, not `JaegerAvailable`.
I think the case with `JaegerAvailable` is much worst because it depends on Kiali being able to reach Jaeger at startup, so it makes it very dependant on the startup sequence of the services.

rhqci:

jotak:
kialiqeci

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1056](https://paste.fedoraproject.org/paste/om71hLCW9N8aJUrPkThcEw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1951]](https://paste.fedoraproject.org/paste/LDBydxYCI4ie6bcfleL2Bw)

status/discover.go
jotak: Created https://issues.jboss.org/browse/KIALI-2963

kiali.go
jotak: JIRA created: https://issues.jboss.org/browse/KIALI-2964

PR #1103, opened 2019-05-29
KIALI-2947 Fix detection of valid+false pair of gateways in VS

burmanm:
** Describe the change **

If the first item in the gateways list was valid, the VS validation did not return the correct result. Validates correctly all the cases now.

** Issue reference **

KIALI-2947


hhovsepy:
Verified:
Screenshot from 2019-05-30 10-45-06

PR #1063, opened 2019-05-13
KIALI-2847 support cmd line options in deploy script

jmazzitelli:
cc @jotak 

See if this provides what you want. It is kind of weird syntax (opts go after the <() argument), but to pass cmdline opts when using <() notation, you do it like this

`bash <(curl -L https://git.io/getLatestKialiOperator) --help`

e.g.

`bash <(curl -L https://git.io/getLatestKialiOperator) -ju http://my-jaeger --auth-strategy anonymous `

PR #1111, opened 2019-06-03
KIALI-2963 Use configured istio namespace for grafana/jaeger/prometheus

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-2963

As discussed in comments, I'm also renaming / cleaning some of the grafana config: https://issues.jboss.org/browse/KIALI-2966

Frontend PR: https://github.com/kiali/kiali-ui/pull/1234

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -228,9 +228,6 @@ type Config struct {
 	Auth             AuthConfig        `yaml:"auth,omitempty"`
 }
 
-// Istio System namespace by default
-const IstioDefaultNamespace = "istio-system"
-
 // NewConfig creates a default Config struct
 func NewConfig() (c *Config) {
 	c = new(Config)
@@ -240,7 +237,7 @@ func NewConfig() (c *Config) {
 	c.Identity.CertFile = getDefaultString(EnvIdentityCertFile, "")
 	c.Identity.PrivateKeyFile = getDefaultString(EnvIdentityPrivateKeyFile, "")
 	c.InCluster = getDefaultBool(EnvInCluster, true)
-	c.IstioNamespace = strings.TrimSpace(getDefaultString(EnvIstioNamespace, IstioDefaultNamespace))
+	c.IstioNamespace = strings.TrimSpace(getDefaultString(EnvIstioNamespace, "istio-system"))
 	c.IstioLabels.AppLabelName = strings.TrimSpace(getDefaultString(EnvIstioLabelNameApp, "app"))
 	c.IstioLabels.VersionLabelName = strings.TrimSpace(getDefaultString(EnvIstioLabelNameVersion, "version"))
 	c.API.Namespaces.Exclude = getDefaultStringArray(EnvApiNamespacesExclude, "istio-operator,kube.*,openshift.*,ibm.*")
@@ -262,14 +259,14 @@ func NewConfig() (c *Config) {
 	c.Server.MetricsEnabled = getDefaultBool(EnvServerMetricsEnabled, true)
 
 	// Prometheus configuration
-	c.ExternalServices.Prometheus.URL = strings.TrimSpace(getDefaultString(EnvPrometheusServiceURL, fmt.Sprintf("http://prometheus.%s:9090", IstioDefaultNamespace)))
+	c.ExternalServices.Prometheus.URL = strings.TrimSpace(getDefaultString(EnvPrometheusServiceURL, fmt.Sprintf("http://prometheus.%s:9090", c.IstioNamespace)))
 	c.ExternalServices.Prometheus.CustomMetricsURL = strings.TrimSpace(getDefaultString(EnvPrometheusCustomMetricsURL, c.ExternalServices.Prometheus.URL))
 
 	// Grafana Configuration
 	c.ExternalServices.Grafana.DisplayLink = getDefaultBool(EnvGrafanaDisplayLink, true)
 	c.ExternalServices.Grafana.InCluster = getDefaultBool(EnvGrafanaInCluster, true)
 	c.ExternalServices.Grafana.URL = strings.TrimSpace(getDefaultString(EnvGrafanaURL, ""))
-	c.ExternalServices.Grafana.ServiceNamespace = strings.TrimSpace(getDefaultString(EnvGrafanaServiceNamespace, IstioDefaultNamespace))
+	c.ExternalServices.Grafana.ServiceNamespace = strings.TrimSpace(getDefaultString(EnvGrafanaServiceNamespace, c.IstioNamespace))
 	c.ExternalServices.Grafana.Service = strings.TrimSpace(getDefaultString(EnvGrafanaService, "grafana"))
 	c.ExternalServices.Grafana.WorkloadDashboardPattern = strings.TrimSpace(getDefaultString(EnvGrafanaWorkloadDashboardPattern, "Istio%20Workload%20Dashboard"))
 	c.ExternalServices.Grafana.ServiceDashboardPattern = strings.TrimSpace(getDefaultString(EnvGrafanaServiceDashboardPattern, "Istio%20Service%20Dashboard"))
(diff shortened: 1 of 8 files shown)


jotak:
@jmazzitelli can you have a look at this please?
I started by just modifying the defaults in `config.go`, but now with the operator it seems that all these defaults are useless, right? So I had to change the ansible script. What we have in our config.go is a bit misleading then.
Don't  you think we could have all of our defaults back in the config.go again?

rhqci:
PRT Failed, Please Contact QE

jotak:
Hmm actually, probably just by adding "omitempty" to the yaml deserialization directives would make it work again with the config.go defaults

jmazzitelli:
> but now with the operator it seems that all these defaults are useless, right? So I had to change the ansible script. What we have in our config.go is a bit misleading then.
> Don't you think we could have all of our defaults back in the config.go again?

They are not useless in the sense that if the configmap doesn't have the settings, the Go defaults take effect. It just so happens that the operator will set all of them. But remember, not everyone is using the installer (think upstream Istio helm charts) so the config.go defaults still are used in that case.

We need to have the operator have all its default values set because otherwise it makes the ansible script more difficult to write/maintain (you now have to worry about doing checks "is this setting defined? No? Do this. Yes?  Do something else).

jotak:
> We need to have the operator have all its default values set because otherwise it makes the ansible script more difficult to write/maintain (you now have to worry about doing checks "is this setting defined? No? Do this. Yes? Do something else).

It seems I could remove these ansible checks if I add `omitempty` as yaml deserializer annotation in config.go ; so that when an empty namespace is given, it will be defaulted (which makes sense as empty namespace is not something possible), so I think I could do that both for grafana namespace and jaeger namespace.

jotak:
@jmazzitelli I've created this JIRA to fix what you mentioned (and more) : https://issues.jboss.org/browse/KIALI-2966

jmazzitelli:
> so that when an empty namespace is given, it will be defaulted (which makes sense as empty namespace is not something possible),

I mean these checks have to go into all the ansible tasks... everywhere you use these variables in the ansible scripts, we would then have to add a "when: the_var_name is defined" otherwise ansible aborts with a "variable not defined error".

Please don't remove the default settings in the ansible roles defaults/main.yml files - they are there very specifically.

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1061](https://paste.fedoraproject.org/paste/zeHqa4KGXS4s~gM5hyIyHw)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1961]](https://paste.fedoraproject.org/paste/qbdfFvAXPDgLtljWy91s~g)

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1063](https://paste.fedoraproject.org/paste/onW~5y8ICL4~mB50EFLG0Q)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1963]](https://paste.fedoraproject.org/paste/Jn9vhqCkScZ~oHWxGw05-w)

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -138,36 +138,25 @@ spec:
...
-# in_cluster: Set true to indicate that Grafana service is accessible in the Kubernetes cluster.
-# password: Password to be used when making requests to Grafana. User only requires viewer permissions.
-# service: The Kubernetes service name for Grafana. Kiali uses this to connect within the cluster to Grafana.
-#          If "service" and "service_namespace" are empty then "url" is used for communication.
-# service_dashboard_pattern: Search pattern for Grafana Service dashboard.
-# service_namespace: The Kubernetes namespace that holds the Grafana service (if empty, assumes the same value as deployment.namespace)
+# in_cluster: Set true to indicate that the Grafana service is accessible in the Kubernetes cluster.
+# password: Password to be used when making requests to Grafana, for basic authentication. User only requires viewer permissions.
+# service: The Kubernetes service name for Grafana. Only used when in_cluster is true.
+# namespace: The Kubernetes namespace that holds the Grafana service. Only used when in_cluster is true. If empty, assumes the same value as deployment.namespace.
```
jmazzitelli: trivial: all of these settings in this file are ordered alphabetically - can you move this "namespace" up a few lines - after " in_cluster" but before "password".

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -138,36 +138,25 @@ spec:
...
 #    ---
 #    grafana:
 #      api_key: ""
 #      display_link: true
 #      in_cluster: true
 #      password: ""
 #      service: "grafana"
-#      service_dashboard_pattern: "Istio%20Service%20Dashboard"
-#      service_namespace: ""
+#      namespace: ""
```
jmazzitelli: same as above - move "namespace" up so it is alphabetically listed (after in_cluster, before password).

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -43,14 +43,9 @@ kiali_defaults:
       in_cluster: true
       password: ""
       service: "grafana"
-      service_dashboard_pattern: "Istio%20Service%20Dashboard"
-      service_namespace: ""
+      namespace: ""
```
jmazzitelli: same as above - move "namespace" up so it is alphabetically listed (after in_cluster, before password).

config/config.go
```diff
@@ -262,20 +254,15 @@ func NewConfig() (c *Config) {
...
 
 	// Grafana Configuration
 	c.ExternalServices.Grafana.DisplayLink = getDefaultBool(EnvGrafanaDisplayLink, true)
 	c.ExternalServices.Grafana.InCluster = getDefaultBool(EnvGrafanaInCluster, true)
 	c.ExternalServices.Grafana.URL = strings.TrimSpace(getDefaultString(EnvGrafanaURL, ""))
-	c.ExternalServices.Grafana.ServiceNamespace = strings.TrimSpace(getDefaultString(EnvGrafanaServiceNamespace, IstioDefaultNamespace))
+	c.ExternalServices.Grafana.Namespace = strings.TrimSpace(getDefaultString(EnvGrafanaNamespace, c.IstioNamespace))
 	c.ExternalServices.Grafana.Service = strings.TrimSpace(getDefaultString(EnvGrafanaService, "grafana"))
-	c.ExternalServices.Grafana.WorkloadDashboardPattern = strings.TrimSpace(getDefaultString(EnvGrafanaWorkloadDashboardPattern, "Istio%20Workload%20Dashboard"))
-	c.ExternalServices.Grafana.ServiceDashboardPattern = strings.TrimSpace(getDefaultString(EnvGrafanaServiceDashboardPattern, "Istio%20Service%20Dashboard"))
```
jmazzitelli: Question: these are now hardcoded because we think these will never change. Are we sure? I am thinking mainly about our multi-tenancy use case when Maistra will be installing several control planes -- is it possible for there to be different Grafana/Jaeger installations (or a single installation?) that has different workload/service dashboard patterns to graph different control planes? I do not know the answer, but I figured it was worth asking before merge this.

jotak: I'm not sure about how multi-tenancy will be implemented, but if we consider there's one grafana per tenant then the URL will discriminate them, not the dashboards.

And if really there's just a single Grafana instance, I think there are good chances that this part would need to be reworked anyway...

operator/roles/kiali-deploy/defaults/main.yml
jotak: done

jotak:
@jmazzitelli ordering fixed

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1064](https://paste.fedoraproject.org/paste/oySnZD2VzEIN~hICRg8Qgg)
  * :heavy_check_mark: run-kiali-e2e-tests [#[1964]](https://paste.fedoraproject.org/paste/IYhdUT-8BYEAboW2WSpJQg)

hhovsepy:
Verified for Grafana poiting to another namespace, and it it autodiscovered.

PR #748, opened 2018-12-17
KIALI-1906 Make changes in the Kiali-UI to integrate Jaeger Components

aljesusg:
[KIALI-1906](https://issues.jboss.org/browse/KIALI-1906)

PR #1113, opened 2019-06-03
KIALI-2964 Remove JaegerAvailable flag

jotak:
When Jaeger is temporary not available, it shows an error in Kiali console.
If Jaeger is not available at all then it should be configured with "Enabled: false".

JIRA: https://issues.jboss.org/browse/KIALI-2964

### DIFF
--- a/business/jaeger_helper.go
+++ b/business/jaeger_helper.go
@@ -29,7 +29,7 @@ type JaegerServices struct {
 func getErrorTracesFromJaeger(namespace string, service string) (errorTraces int, err error) {
 	errorTraces = 0
 	err = nil
-	if !appstate.JaegerAvailable || !config.Get().ExternalServices.Tracing.Enabled {
+	if !config.Get().ExternalServices.Tracing.Enabled {
 		return -1, errors.New("jaeger is not available")
 	}
 	if appstate.JaegerEnabled {
--- a/appstate/jaeger_state.go
+++ b/appstate/jaeger_state.go
@@ -5,8 +5,5 @@ import "github.com/kiali/kiali/config"
 // JaegerEnabled is a memo-flag to tell if Jaeger was configured or succesfully discovered
 var JaegerEnabled = false
 
-// JaegerAvailable tells if Jaeger is ready
-var JaegerAvailable = true
-
 // JaegerConfig is a copy of config.TracingConfig from global config, that can be mutated (e.g. for URL discovery)
 var JaegerConfig config.TracingConfig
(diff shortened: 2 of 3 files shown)


jotak:
@hhovsepy can you check if this is ok with you? The JaegerAvailable flag originated from a comment that you made a while ago : https://github.com/kiali/kiali/pull/748#issuecomment-460524516

But it seems that we can get rid of it now that we have explicit `Enable` config flag.

jotak:
What I did to test this PR is:

- Run Kiali with default config & jaeger available => works as expected
- Turn off (scale down) Jaeger => errors are displayed in Kiali console because Jaeger cannot be reached
- Restart Kiali with Jaeger still turned off => errors still displayed
- Turn on Jaeger => errors are gone, Tracing works in Kiali

And to fully get rid of the errors we just have to explicitly disable Jaeger from Kiali config.

business/jaeger_helper.go
```diff
@@ -29,7 +29,7 @@ type JaegerServices struct {
 func getErrorTracesFromJaeger(namespace string, service string) (errorTraces int, err error) {
 	errorTraces = 0
 	err = nil
-	if !appstate.JaegerAvailable || !config.Get().ExternalServices.Tracing.Enabled {
+	if !config.Get().ExternalServices.Tracing.Enabled {
 		return -1, errors.New("jaeger is not available")
 	}
 	if appstate.JaegerEnabled {
```
aljesusg: But now we can't check if the problem is in Jaeger or in Kiali.

config.Get().ExternalServices.Tracing.Enabled // User enable configuration
JaegerEnabled  // Kiali discover jaeger o check that the information provided by user is ok
JaegerAvailable // Jaeger application is not available.

jotak: I'm not sure to understand. When jaeger is not available, we get an error message in Kiali. So the user will know that something is wrong with Jaeger.

aljesusg: The point was to check if Jaeger was ok but the application is failling for example by elastic search, but maybe we can improve this in Kiali 2.0 :)

jotak: We can check this scenario with jaeger-query up and elastic down, I'm not sure how the error will be reported / maybe will have a timeout. However, either way I think an error should still be caught at fetch time rather than at Kiali startup.

jotak: So, after checking with @hhovsepy it's basically the same from Kiali standpoint to say elastic is failing, or Jaeger is failing. The scenario with failing Jaeger has been tested here and seems OK, as expected.

PR #1120, opened 2019-06-06
KIALI-2778 Use same kiali oauth token for grafana API

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-2778

With some refactoring to get consistent config with Prometheus
This PR is based on https://github.com/kiali/kiali/pull/1129 , so it should be merged only after that.


jotak:
@xeviknal this is a draft but you can see what i've done so far.
FYI, the config map I use will be like that:

```
    in_cluster_url: "https://grafana.istio-system.svc:3000"
    insecure_skip_verify: true
    use_oauth_token: true
```

rhqci:
Jenkins CI: kiali-core-pr-e2e-test [#1074](https://paste.fedoraproject.org/paste/H67Yqh9~wfbeHx9lxo22jA)

mwringe:
There is also this PR to add in support for oauth access of Prometheus: https://github.com/kiali/kiali/pull/1129

We should align on the configmap naming of things so that they are consistent.

jotak:
@mwringe +1 to coordinate both PRs. Your PR on prometheus also adds CA appending that I should add here too.
The grafana also config supports basic auth and token auth (not using kiali's token, but a config-defined one) so I think it's a mix of the two PRs that we should do.

jotak:
Suggestion:

We can define a new struct "auth" that will be used by both grafana and prometheus (and jaeger?), make sure we remove any inconsistency. That would be something like:

```yaml
prometheus|grafana:
  auth: 
    type: string # "none" or "basic" or "bearer"
    username: string # used only for basic type
    password: string # used only for basic type
    token: string # used only for bearer type
    use_kiali_token: bool # used only for bearer type; if true, the "token" config is ignored
```

We can do with or without `use_kiali_token` ; an empty `token` could be implicit way to tell to use kiali token. It's just a matter of preference, do we want to make it implicit, or explicit with an extra config.

jotak:
cc @mwringe I've merged your PR in this one.
If you can have a look, especially at the last commit https://github.com/kiali/kiali/pull/1120/commits/c9ee970f3949f4e971892d511fe36a7f3560e1df and tell me what you think

jotak:
@aljesusg if you're working based on this PR: I've updated it to fix an issue with the way we get the user token (it took me a while to figure out I was actually not picking the right token).
I've tested this one with Grafana behind proxy and it works fine. The config I'm using is:

```yaml
      prometheus:
        url: "https://prometheus.istio-system.svc:9090" 
        auth:
          type: "bearer"
          insecure_skip_verify: true
          use_kiali_token: true
      grafana:
        in_cluster_url: "https://grafana.istio-system.svc:3000" 
        auth:
          type: "bearer"
          insecure_skip_verify: true
          use_kiali_token: true
```

aljesusg:
Thanks @jotak I'll check it this night

operator/roles/kiali-deploy/tasks/main.yml
```diff
@@ -46,12 +46,6 @@
   when:
   - kiali_vars.istio_namespace == ""
```
jotak: @jmazzitelli the grafana config is changing so there's no explicit "namespace" config anymore and I'm removing this script.

However, now the config has a default url instead: `in_cluster_url: "http://grafana.istio-system:3000"`

I wonder if we should introduce some kind variable replacement mechanism in the config. For instance, we would write something like:

`in_cluster_url: "http://grafana.${ISTIO_NAMESPACE}:3000"`
And that would be automatically replaced in code.

aljesusg:
@jotak please rebase https://github.com/Maistra/istio-operator/pull/80 and https://github.com/kiali/kiali/pull/1129 were merged we can go ahead with this ;)

config/config.go
```diff
@@ -128,26 +124,30 @@ type Server struct {
 	MetricsEnabled             bool                 `yaml:"metrics_enabled,omitempty"`
 }
 
+// Auth provides authentication data for external services
+type Auth struct {
+	Type               string `yaml:"type"`
+	Username           string `yaml:"username"`
+	Password           string `yaml:"password"`
+	Token              string `yaml:"token"`
+	UseKialiToken      bool   `yaml:"use_kiali_token"`
```
jotak: Team, can you review this config please and comment?

jotak:
Just updated some in-code doc... now it can be merged imo

aljesusg:
can we merge this @jotak ? Remove the label please

xeviknal:
Removing DNM labels per @jotak comment and merging.

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -39,25 +39,32 @@ kiali_defaults:
...
+        type: "none"
+        use_kiali_token: false
+        username: ""
       display_link: true
-      in_cluster: true
-      insecure_skip_verify: false
-      namespace: ""
-      password: ""
-      service: "grafana"
+      in_cluster_url: "http://grafana.istio-system:3000"
```
jmazzitelli: This is not good - it assumes grafana is in istio-system which won't be the case in multitenancy. 

We need to set this default to `""` empty string and define this in roles/kiali-deploy/tasks/main.yml - it should set this to `http://grafana.{{ kiali_vars.istio-namespace }}:3000`

operator/roles/kiali-deploy/tasks/main.yml
```diff
@@ -46,12 +46,6 @@
...
 
-- name: Set default Grafana namespace
```
jmazzitelli: this section should be resurrected and should build `in_cluster_url` in the same way

Something like:

```
set_fact:
  kiali_vars: "{{ kiali_vars | combine({'external_services': {'grafana': {'in_cluster_url': 'http://grafana.' + kiali_vars.istio_namespace + ':3000'}}}, recursive=True) }}"
when
  kiali_vars.external_services.grafana.auth.in_cluster_url == ""
```

jmazzitelli: of course, change the `name:` to be `Set default Grafana in_cluster_url`

jmazzitelli: all this work is done -- #1145 is merged

operator/roles/kiali-deploy/defaults/main.yml
jotak: @jmazzitelli I prefer to have a default that works in the most common cases, and having overrides performed in less common cases. For the multi-tenant / Maistra scenario, I assume should be able to override any config ; is that not the case?

jotak: BTW what you're suggesting is more or less what I suggesting above (2 days ago) : https://github.com/kiali/kiali/pull/1120#discussion_r292786599
So if that works, sure we can do that instead

config/config.go
```diff
@@ -274,26 +274,13 @@ func NewConfig() (c *Config) {
...
-	c.ExternalServices.Grafana.Namespace = strings.TrimSpace(getDefaultString(EnvGrafanaNamespace, c.IstioNamespace))
-	c.ExternalServices.Grafana.Service = strings.TrimSpace(getDefaultString(EnvGrafanaService, "grafana"))
-	c.ExternalServices.Grafana.APIKey = strings.TrimSpace(getDefaultString(EnvGrafanaAPIKey, ""))
-	c.ExternalServices.Grafana.Username = strings.TrimSpace(getDefaultString(EnvGrafanaUsername, ""))
-	c.ExternalServices.Grafana.Password = strings.TrimSpace(getDefaultString(EnvGrafanaPassword, ""))
-	if c.ExternalServices.Grafana.Username != "" && c.ExternalServices.Grafana.Password == "" {
-		log.Error("Grafana username (\"GRAFANA_USERNAME\") requires that Grafana password (\"GRAFANA_PASSWORD\") is set.")
-	}
-	c.ExternalServices.Grafana.InsecureSkipVerify = getDefaultBool(EnvGrafanaInsecureSkipVerify, false)
+	c.ExternalServices.Prometheus.Auth = getAuthFromEnv("GRAFANA")
```
mousavian: Is this intentional? Because it's overwriting prometheus auth config in line 277.
i think it should be 
```
c.ExternalServices.Grafana.Auth = getAuthFromEnv("GRAFANA")
```

pilhuhn:
@mousavian Excellent question :)

config/config.go
pilhuhn: Excellent question :)

jotak: oh, good catch!

pilhuhn: See #1197

mousavian: Thank you for quick reply and PR! :)

PR #1121, opened 2019-06-06
KIALI-2955 Support multiple oauthclients

mwringe:
** Describe the change **

Previously the oauthclient name has hard coded to 'kiali'

This change allows the user to specify another name if they wish.

If no name is specified, then the operator will default to 'kiali-${NAMESPACE}' where ${NAMESPACE} is the namespace where Kiali has been deployed.

** Issue reference **

https://issues.jboss.org/browse/KIALI-2955

### DIFF
--- a/operator/roles/kiali-remove/tasks/main.yml
+++ b/operator/roles/kiali-remove/tasks/main.yml
@@ -95,7 +95,7 @@
   - os_item.metadata is defined
   - os_item.metadata.name is defined
   with_items:
-  - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='OAuthClient', resource_name='kiali', api_version='oauth.openshift.io/v1') }}"
+  - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='OAuthClient', resource_name='kiali-' + kiali_vars.deployment.namespace, api_version='oauth.openshift.io/v1') }}"
   - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='Route', resource_name='kiali', api_version='route.openshift.io/v1') }}"
   loop_control:
     loop_var: os_item
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -57,6 +57,8 @@ type OAuthRouteTLSSpec struct {
 
 const serverPrefix = "https://kubernetes.default.svc/"
 
+var kialiNamespace string
+
 func (in *OpenshiftOAuthService) Metadata() (metadata *OAuthMetadata, err error) {
 	redirectURL, err := getKialiRoutePath()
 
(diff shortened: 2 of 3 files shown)


operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -23,6 +23,7 @@ kiali_defaults:
 
   auth:
     strategy: ""
+    oauth_name: ""
```
jmazzitelli: trivial - All of these settings are alphabetically ordered. Keep that convention - move `oauth_name` above `strategy`.

operator/roles/kiali-remove/defaults/main.yml
```diff
@@ -1,6 +1,8 @@
 kiali_defaults:
   deployment:
     namespace: "istio-system"
+  auth:
```
jmazzitelli: trivial - the top level "auth" should appear alphabetically above "deployment" so move this above like this:

```
auth:
  oauth_name: ""
deployment:
  namespace: "istio-system"
```

operator/roles/kiali-remove/tasks/main.yml
```diff
@@ -26,6 +26,12 @@
   - is_openshift == False
   - is_k8s == False
 
+- name: Set default oauth name when using the openshift strategy
+  set_fact:
+    kiali_vars: "{{ kiali_vars | combine({'auth': {'oauth_name': 'kiali-' + kiali_vars.deployment.namespace}}, recursive=True) }}"
+  when:
+  - kiali_vars.auth.oauth_name == ""
```
jmazzitelli: Based on the name ("...when using the openshift strategy"), I think you want an additional when clause here:

```
- is_openshift == True
```

operator/roles/kiali-remove/vars/main.yml
```diff
@@ -5,3 +5,10 @@ kiali_vars:
     {%- else -%}
     {{ kiali_defaults.deployment }}
     {%- endif -%}
+  auth: |
```
jmazzitelli: trivial - keep alphabetical ordering convention - move this "auth" block above the "deployment" block

jmazzitelli:
@mwringe the more I think about these changes, the more I think we should do it differently. I don't think we should allow the user to configure this name. It is going to make things harder on us. If someone edits the Kiali CR and changes this oauth_name, the operator now has to know to delete the old one and create the new one.

I think we should rip out all this oauth_name config setting and just always assume the name of the OAuthclient is going to be "kiali-" + deployment.namespace.

I see no benefit of allowing a user control over this name - we should just always name it "kiali-" + deployment.namespace. That way, it will always be unique per Kiali installation but it will also always be fixed/known.

mwringe:
Closing PR to open another one that is more simple and just hardcodes the oauthname to be kiali-${NAMESPACE}

PR #1123, opened 2019-06-07
KIALI-2955 Allow for multiple oauthclients

mwringe:
** Describe the change **

We used to use a hardcoded oauth name of "kiali" which doesn't work well if you have multiple Kiali's installed in the same cluster.

This update changes it to be hardcoded to "kiali-${NAMESPACE}" so that multiple Kiali instances can be installed in the cluster without a collision.

Note that installing multiple Kiali instances in the same namespace is not supported. 

** Issue reference **

https://issues.jboss.org/browse/KIALI-2955

### DIFF
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -57,6 +57,8 @@ type OAuthRouteTLSSpec struct {
 
 const serverPrefix = "https://kubernetes.default.svc/"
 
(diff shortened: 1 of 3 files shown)


jmazzitelli:
LGTM, although it is failing the format check.. looks like you need to run gofmt?

PR #1129, opened 2019-06-10
KIALI-2776 Support bearer tokens when accessing Prometheus

mwringe:
**Describe the change**

Allows for connection to Prometheus servers that are behind the oauth proxy. Such as what Maistra is going to be configurable in Maistra.

This introduced two new configuration options under the Prometheus option in external_services:

**ca_file**: The CA file that was used to sign the certificate used by Prometheus. If you are running with Maistra then the endpoint is signed by the service signer and you will need to use `/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`. Leaving this empty means that no additional file is used.

**auth**: Setting this to `bearer` means that we send the Kiali service account bearer token in the `Authorization` header when contacting Prometheus. Leaving this as `none` means that it works as it does today where no authentication information is passed to Prometheus.

If you are using the PR from https://github.com/Maistra/istio-operator/pull/80 then your external_services in the kiali configmap should look as follows:

````
external_services:
  prometheus:
    url: https://prometheus.istio-system.svc:9090
    ca_file: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
    auth: bearer
````
    

**Issue reference**

https://issues.jboss.org/browse/KIALI-2776

**Backwards incompatible?**

Everything should work as before if these values are not specified in the configmap or kiali CR.

### DIFF
--- a/prometheus/client.go
+++ b/prometheus/client.go
@@ -3,7 +3,9 @@ package prometheus
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"time"
 
@@ -12,6 +14,8 @@ import (
 	"github.com/prometheus/common/model"
 
 	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/log"
 )
 
 // ClientInterface for mocks (only mocked function are necessary here)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: prometheus/client.go [hunk 1]
```diff
@@ -38,18 +42,62 @@ type Client struct {
...
+
+		certPool.AppendCertsFromPEM(cert)
+
+		transportConfig.TLSClientConfig = &tls.Config{
+			RootCAs: certPool,
+		}
+	}
+
+	if cfg.Auth == "bearer" {
+		token, err := kubernetes.GetKialiToken()
```
jotak: Here you're reading the Kiali SA token to pass it to prometheus ... i don't think it's what we want, or is it? I would imagine we had to "forward" the user token that comes from the request that triggers the prom api call.

aljesusg:
This throw me connection refused

```bash
Failed to fetch Prometheus flags: Get https://prometheus-istio-system.192.168.86.71.nip.io:9090/api/v1/status/flags: dial tcp 192.168.86.71:9090: connect: connection refused

```

prometheus/client.go
mwringe: We want to use the Kiali service account token here and not the user token.

The user token is still used, but its used by Kiali determine if the user has access to those metrics or not. Its Kiali that does the metric filtering, not Prometheus (since Prometheus doesn't have the concept of metric permissions).

For the other secured components, such as Grafana and Jaeger, we do want to use the user token. Its only for Prometheus that we don't since Kiali already does the filtering.

jshaughn: Why not comment the code with this kind of information?  No one will remember all the good info in the PR.

mwringe:
> This throw me connection refused
>
> Failed to fetch Prometheus flags: Get https://prometheus-istio-system.192.168.86.71.nip.io:9090/api/v1/status/flags: dial tcp 192.168.86.71:9090: connect: connection refused

You should have configured your url value to be the one mentioned in https://github.com/kiali/kiali/pull/1129#issue-286785264

`url: https://prometheus.istio-system.svc:9090`

Note: the .svc at the end is important here if you are using the service signing ca, using 'prometheus.istio-system' is not valid since the ca doesn't sign it with that hostname.

A lot of the time you can't use the route  URL from within the pod itself, it doesn't always resolve properly (at least on my machine)

aljesusg:
I still get an error

Failed to fetch Prometheus flags: Get https://prometheus.istio-system.svc:9090/api/v1/status/flags: x509: certificate signed by unknown authority

mwringe:
@aljesusg You need to see the `ca_file` value as mentioned in https://github.com/kiali/kiali/pull/1129#issue-286785264

You can also set it to be insecure, but that should never be recommended to users.

aljesusg:
I set the value to ca_file: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt @mwringe  like the comment so i don't know what happen, I am going to test again or add insecure. Thanks

prometheus/client.go
lucasponce: +1
To add extended comments in the code for this.
It will help a lot to maintain it for others.

Thanks !

aljesusg:
Ok this is valid @mwringe , tested  it after recreate environment!!! Thanks !!

aljesusg:
@mwringe can we rebase this and merge it ?

mwringe:
Rebased and a bit more explanation of why Prometheus uses the Kiali service account token and not the user's token directly.

PR #1131, opened 2019-06-11
Update Go dashboard

simonpasquier:
**Describe the change**
CPU and RSS memory panels are added at the top. The rate of heap
allocations is also added.

The 'pod_name' label is added to all panels.

The 'go_gc_duration_seconds' panel is removed because it is based on a
summary metric that can't be aggregated. Instead the rate of GC events
is added.

The 'go_threads' panel is removed, tracking the number of goroutines is
usually enough.

image

image

image

**Issue reference**

None. It was discussed directly with @jotak 

**Backwards incompatible?**

No

**Documentation**

If this is merged, [kiali.io](https://www.kiali.io/documentation/runtimes-monitoring/#_go) would need to be updated too.

### DIFF
--- a/operator/roles/kiali-deploy/templates/dashboards/go.yaml
+++ b/operator/roles/kiali-deploy/templates/dashboards/go.yaml
@@ -8,27 +8,53 @@ spec:
   discoverOn: "go_info"
   items:
   - chart:
-      name: "Threads"
+      name: "CPU ratio"
       spans: 6
-      metricName: "go_threads"
+      metricName: "process_cpu_seconds_total"
+      dataType: "rate"
+      aggregations:
+      - label: "pod_name"
+        displayName: "Pod"
+  - chart:
+      name: "RSS Memory"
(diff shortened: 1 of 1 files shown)


rhqci:

operator/roles/kiali-deploy/templates/dashboards/go.yaml
```diff
@@ -8,27 +8,53 @@ spec:
...
       name: "Goroutines"
       spans: 6
       metricName: "go_goroutines"
       dataType: "raw"
+      aggregations:
+      - label: "pod_name"
+        displayName: "Pod"
+  - chart:
+      name: "Heap allocation rate"
+      unit: "bytes/s"
```
jotak: I will probably need to handle it like bitrate for auto-scaling : https://docker.pkg.github.com/kiali/k-charted/blob/master/web/react/src/utils/formatter.ts#L15

PR #1133, opened 2019-06-11
KIALI-3007 support watching CR in a different namespace

jmazzitelli:
https://issues.jboss.org/browse/KIALI-3007

I think this is going to be needed for https://issues.jboss.org/browse/KIALI-3007

NOTE: this PR does not affect current behavior. The defaults are set such that the behavior is left unchanged. This just allows someone to tell the install script to have the operator look elsewhere for the Kiali CR (as opposed to the same namespace where the operator is located). The default is the operator will look in its own namespace (which is how it works today).

### DIFF
--- a/operator/deploy/deploy-kiali-operator.sh
+++ b/operator/deploy/deploy-kiali-operator.sh
@@ -93,6 +93,10 @@
 #    to the value of OPERATOR_IMAGE_VERSION env var value.
 #    Default: See above for how the default value is determined
 #
+# OPERATOR_WATCH_NAMESPACE
+#    The namespace in which the operator looks for the Kiali CR.
+#    Default: The configured OPERATOR_NAMESPACE
+#
 # -----------
 # Environment variables that affect Kiali installation:
 #
@@ -276,6 +280,10 @@ while [[ $# -gt 0 ]]; do
       OPERATOR_VERSION_LABEL="$2"
       shift;shift
       ;;
+    -own|--operator-watch-namespace)
+      OPERATOR_WATCH_NAMESPACE="$2"
+      shift;shift
+      ;;
     -sn|--secret-name)
       SECRET_NAME="$2"
       shift;shift
@@ -339,6 +347,9 @@ Valid options for the operator installation:
       A Kubernetes label named "version" will be set on the Kiali operator resources.
       The value of this label is determined by this setting.
       Default: Determined by the operator image version being installed
+  -own|--operator-watch-namespace
+      The namespace in which the operator looks for the Kiali CR.
+      Default: The configured operator namespace (-on)
 
 Valid options for Kiali installation (if Kiali is to be installed):
   -an|--accessible-namespaces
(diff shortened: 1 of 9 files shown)


jmazzitelli:
The merge-operator-yaml.sh is here because I think we are going to need this when we have to build the yaml to give to Maistra. Rather than have to manually build/change the single yaml for Maistra to use, I think we can use this script to build it for us. I'm still not 100% sure it is going to be used, but I put it in here because I think it might used.

jmazzitelli:
Here's how to test this.

First, make sure you build everything from the top directory: `make docker-build`.

The remaining steps I will assume your current working directory is `operator/deploy`.

Now test 2 things - one using SingleNamespace mode with namespace-scoped operator. Then AllNamespace mode with cluster-scoped operator.

## TEST 1

Test SingleNamespace mode, namespace-scoped operator

Delete everything to make sure you start from scratch: `./deploy-kiali-operator.sh -um true`

Run the deploy script and tell it you want to change the operator watch namespace to a specific namespace that is NOT kiali-operator. Let's test with `myproject` namespace.

`./deploy-kiali-operator.sh -oiv dev -own myproject`

Answer the prompts. This will now install the Kiali CR in the `myproject` namespace - confirm that:

`oc get kiali kiali -n myproject`

Confirm also that Kiali is installed in istio-system namespace:

`oc get deployment kiali -n istio-system`

## TEST 2

Test AllNamespace mode, cluster-scoped operator

Delete everything to make sure you start from scratch: `./deploy-kiali-operator.sh -um true`

Run the deploy script and tell it you want to change the operator watch namespace to all namespaces. That means setting the watch namespace to `**`. Note that because the script won't know where to install Kiali, it will not attempt to do so (we'll install Kiali in the next step):

`./deploy-kiali-operator.sh -oiv dev -own '**'`

Now use the Makefile to create a Kiali CR in some custom namespace - we'll create a namespace called "foo" and put it in there:

```
oc create namespace foo
OPERATOR_WATCH_NAMESPACE=foo make -C .. kiali-create
```

Confirm the Makefile put the Kiali CR in the `foo` namespace - confirm that:

`oc get kiali kiali -n foo`

After you give the operator some time to complete everything, confirm that Kiali is installed in istio-system namespace:

`oc get deployment kiali -n istio-system`

operator/deploy/deploy-kiali-operator.sh
```diff
@@ -797,27 +818,41 @@ fi
...
+  local _ns="${OPERATOR_WATCH_NAMESPACE}"
+  if [ "${_ns}" == '""' ]; then
+    _ns="<any namespace you choose>"
   fi
   echo "=========================================="
   echo "Skipping the automatic Kiali installation."
-  echo "To install Kiali, create a Kiali custom resource in the namespace [${OPERATOR_NAMESPACE}]."
+  echo "To install Kiali, create a Kiali custom resource in the namespace [$_ns]."
   echo "An example Kiali CR with all settings documented can be found here:"
   echo "  https://raw.githubusercontent.com/kiali/kiali/${_BRANCH}/operator/deploy/kiali/kiali_cr.yaml"
```
jshaughn: This URL is bad. Use:
https://github.com/kiali/kiali/blob/${_BRANCH}/master/operator/deploy/kiali/kiali_cr.yaml

Also, below.

jmazzitelli:
This is needed for the Maistra work that is starting: https://issues.jboss.org/browse/MAISTRA-369 so I'm going to merge.

PR #1134, opened 2019-06-12
Kiali 2779 Handle Jaeger connections with authentication

aljesusg:
JIRA: [KIALI-2779](https://issues.jboss.org/browse/KIALI-2779)

Require: 

1. https://github.com/kiali/kiali/pull/1129
2. https://github.com/kiali/kiali/pull/1120

Configmap

```yaml
  tracing:
    auth:
      ca_file: /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
      insecure_skip_verify: true
      password: ''
      token: ''
      type: bearer
      use_kiali_token: true
      username: ''
    enabled: true
    namespace: istio-system
    service: ''
    url: ''
```

PR #1135, opened 2019-06-12
KIALI-2976 Support hiding Jaeger and Grafana based on authorization

xeviknal:
** Describe the change **
Preventing URL discovery of Jaeger and grafana in case user doesn't have access to pods of istio install namespace.

** Issue reference **
https://issues.jboss.org/browse/KIALI-2976

** Backwards incompatible? **
yes

Needs: https://github.com/kiali/kiali-ui/pull/1262

PRs:
This PR contents: https://github.com/kiali/kiali/pull/1129, https://github.com/kiali/kiali/pull/1120 and https://github.com/kiali/kiali/pull/1134. They should go first.

Outcome:
Admin with cluster-admin permissions:
admin - multitenancy

Foo with permissions only to bookinfo:
foo - multitenancy (1)

Foo with disabled `authorize_external_services: false` option:
Screen recording (15)
Note that Grafana url discovery actually needs to ping grafana for dashboard discovery. Then Kiali triggers an error.

### DIFF
--- a/handlers/jaeger.go
+++ b/handlers/jaeger.go
@@ -54,8 +54,7 @@ func getJaegerInfo(requestToken string) (*models.JaegerInfo, int, error) {
 		auth.Token = requestToken
 	}
 
-	ha, code, err := canAccessURL(apiURL, &auth)
-	if !ha {
+	if ha, code, err := canAccessURL(apiURL, &auth); !ha {
 		return nil, code, errors.New("can't reach Jaeger query service")
 	} else if err != nil {
 		return nil, http.StatusInternalServerError, err
--- a/handlers/jaeger.go
+++ b/handlers/jaeger.go
@@ -1,31 +1,76 @@
 package handlers
 
 import (
+	"errors"
 	"net/http"
+	"time"
 
 	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
+	"github.com/kiali/kiali/status"
+	"github.com/kiali/kiali/util/httputil"
 )
 
-// Get JaegerInfo provides the proxy Jaeger URL
+// Get JaegerInfo provides the Jaeger URL and other info, first by checking if a config exists
+// then (if not) by inspecting the Kubernetes Jaeger service in Istio installation namespace
 func GetJaegerInfo(w http.ResponseWriter, r *http.Request) {
-	jaegerConfig := config.Get().ExternalServices.Tracing
-	info := models.JaegerInfo{
-		URL: jaegerConfig.URL,
+	requestToken, err := getToken(r)
+	if err != nil {
+		RespondWithError(w, http.StatusInternalServerError, "Token initialization error: "+err.Error())
+		return
 	}
 
-	// Check if URL is in the configuration
-	if info.URL == "" {
-		RespondWithError(w, http.StatusNotFound, "You need to set the Jaeger URL configuration.")
+	info, code, err := getJaegerInfo(requestToken)
+	if err != nil {
+		log.Error(err)
+		RespondWithError(w, code, err.Error())
 		return
+	}
+	RespondWithJSON(w, code, info)
+}
+
+func getJaegerInfo(requestToken string) (*models.JaegerInfo, int, error) {
+	jaegerConfig := config.Get().ExternalServices.Tracing
+
+	if !jaegerConfig.Enabled {
+		return nil, http.StatusNoContent, nil
+	}
+
+	apiURL := status.DiscoverJaeger()
+	if apiURL == "" {
+		return nil, http.StatusServiceUnavailable, errors.New("Jaeger URL is not set in Kiali configuration")
 	}
 
 	// Check if URL is valid
-	_, error := validateURL(info.URL)
-	if error != nil {
-		RespondWithError(w, http.StatusNotAcceptable, "You need to set a correct format for Jaeger URL in the configuration error: "+error.Error())
-		return
+	_, err := validateURL(apiURL)
+	if err != nil {
+		return nil, http.StatusServiceUnavailable, errors.New("wrong format for Jaeger URL: " + err.Error())
 	}
 
-	RespondWithJSON(w, http.StatusOK, info)
+	// Be sure to copy config.Auth and not modify the existing
+	auth := jaegerConfig.Auth
+	if auth.UseKialiToken {
+		auth.Token = requestToken
+	}
+
+	ha, code, err := canAccessURL(apiURL, &auth)
+	if !ha {
+		return nil, code, errors.New("can't reach Jaeger query service")
+	} else if err != nil {
+		return nil, http.StatusInternalServerError, err
+	}
+
+	info := &models.JaegerInfo{
+		URL: apiURL,
+	}
+
+	return info, http.StatusOK, nil
+}
+
+func canAccessURL(url string, auth *config.Auth) (bool, int, error) {
+	timeout := time.Duration(1000 * time.Millisecond)
+	_, code, _ := httputil.HttpGet(url, auth, timeout)
+
+	return code == 200, code, nil
 }
--- a/handlers/jaeger.go
+++ b/handlers/jaeger.go
@@ -54,10 +54,8 @@ func getJaegerInfo(requestToken string) (*models.JaegerInfo, int, error) {
 		auth.Token = requestToken
 	}
 
-	if ha, code, err := canAccessURL(apiURL, &auth); !ha {
-		return nil, code, errors.New("can't reach Jaeger query service")
-	} else if err != nil {
-		return nil, http.StatusInternalServerError, err
+	if ha, err := canAccessURL(apiURL, &auth); !ha {
+		return nil, http.StatusServiceUnavailable, err
 	}
 
 	info := &models.JaegerInfo{
@@ -67,7 +65,7 @@ func getJaegerInfo(requestToken string) (*models.JaegerInfo, int, error) {
 	return info, http.StatusOK, nil
 }
 
-func canAccessURL(url string, auth *config.Auth) (bool, int, error) {
+func canAccessURL(url string, auth *config.Auth) (bool, error) {
 	_, code, err := httputil.HttpGet(url, auth, 1000*time.Millisecond)
-	return code == 200, code, err
+	return code == 200, err
 }
--- a/handlers/jaeger.go
+++ b/handlers/jaeger.go
@@ -68,8 +68,6 @@ func getJaegerInfo(requestToken string) (*models.JaegerInfo, int, error) {
 }
 
 func canAccessURL(url string, auth *config.Auth) (bool, int, error) {
-	timeout := time.Duration(1000 * time.Millisecond)
-	_, code, _ := httputil.HttpGet(url, auth, timeout)
-
-	return code == 200, code, nil
+	_, code, err := httputil.HttpGet(url, auth, 1000*time.Millisecond)
+	return code == 200, code, err
 }
--- a/handlers/grafana.go
+++ b/handlers/grafana.go
@@ -22,7 +22,7 @@ const (
 )
 
 // GetGrafanaInfo provides the Grafana URL and other info, first by checking if a config exists
-// then (if not) by inspecting the Kubernetes Grafana service in namespace istio-system
+// then (if not) by inspecting the Kubernetes Grafana service in Istio installation namespace
 func GetGrafanaInfo(w http.ResponseWriter, r *http.Request) {
 	requestToken, err := getToken(r)
 	if err != nil {


status/versions.go
```diff
@@ -259,3 +274,27 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
+
+func canIGetIstioPods(authToken string) (bool, error) {
+	c := config.Get()
+
+	client, err := clientFactory.GetClient(authToken)
+	if err != nil {
+		return false, err
+	}
+
+	ssars, err := client.GetSelfSubjectAccessReview(c.IstioNamespace, "", "pods", []string{"list"})
```
jotak: I'm not sure it's a subjectaccessreview that we need here. I would rather contact the related services (grafana/jaeger) to make sure access is not forbidden.

Subject access review is using k8s API to get permissions on k8s resources, e.g. pods (as being done here); but does the user really have to get pods list permission to access Grafana? Seems to me that it's unrelated.

jotak: cc @mwringe

mwringe: The subject access review is because the oauth server has been configured using that.

I don't think there is a problem is we also just check if we can access it or not. That might actually be better since its makes configuration easier.

mwringe:
@xeviknal Can you post your kiali configmap that you are using?

xeviknal:
```yaml
istio_namespace: istio-system
    server:
      port: 20001
      static_content_root_directory: /opt/kiali/console
    external_services:
      istio:
        url_service_version: http://istio-pilot:8080/version
      jaeger:
        namespace: istio-system
        url: 
      prometheus:
        url: "https://prometheus.istio-system.svc:9090" 
        auth:
          type: "bearer"
          insecure_skip_verify: true
          use_kiali_token: true
      grafana:
        in_cluster_url: "https://grafana.istio-system.svc:3000" 
        auth:
          type: "bearer"
          insecure_skip_verify: true
          use_kiali_token: true
    identity:
      cert_file: /kiali-cert/tls.crt
      private_key_file: /kiali-cert/tls.key
```

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -76,6 +76,14 @@ spec:
 #    ---
 #    strategy: ""
 
+##########
+# Determines whether to show links of Grafana and Traces or not depending on user
+# access to Istio control plane pods. Kiali won't show Grafana and Tracing integration
+# in case the user doesn't have permissions to list pods of IstioNamespace.
+# When empty, its default will be `true`.
+#    ---
+#  authorize_external_services: false
```
mwringe: Should we not show the default value of 'true' here instead of 'false'?

kiali.go
```diff
@@ -101,7 +101,7 @@ func main() {
 	// check if Jaeger is available
 	// we need first discover Jaeger
 	if config.Get().ExternalServices.Tracing.Enabled {
-		status.DiscoverJaeger()
+		status.DiscoverJaeger("")
```
mwringe: Can we have a comment here why we are passing ""

operator/deploy/kiali/kiali_cr.yaml
jmazzitelli: @mwringe I think the default is explicitly set to "false" ... when you explicitly set this value to "", it will default to "true" (or, at least, that's how I understand these couple of lines)

config/config.go
```diff
@@ -66,6 +66,7 @@ const (
 	EnvLoginTokenSigningKey        = "LOGIN_TOKEN_SIGNING_KEY"
 	EnvLoginTokenExpirationSeconds = "LOGIN_TOKEN_EXPIRATION_SECONDS"
 	EnvIstioNamespace              = "ISTIO_NAMESPACE"
+	EnvAuthorizeExternalServices   = "AUTHORIZE_EXTERNAL_SERVICES"
```
mwringe: I think we want to have this as a configuration option for each service.

For instance, I could have just Jeager be behind an oauth proxy but not grafana.

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -76,6 +76,14 @@ spec:
 #    ---
 #    strategy: ""
 
+##########
+# Determines whether to show links of Grafana and Traces or not depending on user
+# access to Istio control plane pods. Kiali won't show Grafana and Tracing integration
+# in case the user doesn't have permissions to list pods of IstioNamespace.
+# When empty, its default will be `true`.
+#    ---
```
jmazzitelli: if this setting is a top level setting, then format this "---" line - all of these separators are supposed to be on the same indentation as the level that the settings are at. Helps keep a visual clue as to what level these settings are at.
```suggestion
#  ---
```

mwringe:
Awesome, this appears to be working well for me :)

I left a few comments, but they are minor.

I think I also agree with @jotak in that instead of doing a subjectaccessreview we determine this by seeing if the user can access the endpoints or not. Checking with a subjectaccessreview I think is too close to the maistra implementation and wont work in other situations (or if maistra decides to change what their oauth proxies require)

operator/deploy/kiali/kiali_cr.yaml
jmazzitelli: OK, after looking at this closely, this is wrong. All commented out settings here have values set to their defaults. I think you want this to say "true". This .yaml file is used for documentation - people go here to see the settings, descriptions, and the default values (read the top comment in this file, it explains this). So change this to:
```suggestion
#  authorize_external_services: true
```

jmazzitelli: Also, any settings you add must be added to the roles/kiali-deploy/defaults/main.yml. So add this there:

```
...
auth:
  strategy: ""

authorize_external_services: true

deployment:
...
```

config/config.go
jmazzitelli: +1. I think we should get rid of this top level "authorize_external_service" setting and put one under each grafana and tracing - it makes sense here under external_services since this affects ... external services :)

```
external_services:
  grafana:
    confirm_access: true
  tracing:
    confirm_access: true
```      

I don't understand the use of the word "authorize" here (what does "authorize_external_service" mean?). But you get the idea. Perhaps "confirm_access" or something? Then again, I'm not quite sure what this is doing - sounds like its just making a request to confirm you can access the external service - if so, I think this should be "confirm_access".

xeviknal:
@mwringe @jmazzitelli I approached your changes. Let me know what do you think.
Instead of checking the RBAC I reach the discovered URL for both services and returning unknown url in case user don't have access.

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -169,6 +169,7 @@ spec:
 #   use_kiali_token: When true and if auth.type is "bearer", the same OAuth token used for authentication in Kiali will be used for the API calls to Grafana,
 #       and auth.token config is ignored then.
 #   username: Username to be used when making requests to Grafana, for basic authentication. User only requires viewer permissions.
+# confirm_access: When true, Kiali will check if logged user has access to Grafana endpoint. It will show Grafana links in case of having access, It won't instead.
```
jmazzitelli: Very trivial change - just re-wording description:

```suggestion
# confirm_access: When true, Kiali will check if logged-in user has access to the Grafana endpoint. If there is access, Grafana links will be shown in Kiali UI.
```

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -244,6 +246,7 @@ spec:
 #   use_kiali_token: When true and if auth.type is "bearer", the same OAuth token used for authentication in Kiali will be used for the API calls to Grafana,
 #       and auth.token config is ignored then.
 #   username: Username to be used when making requests to Jaeger, for basic authentication. User only requires viewer permissions.
+# confirm_access: When true, Kiali will check if logged user has access to Tracing service. It will show Tracing integration in case of having access, It won't instead.
```
jmazzitelli: Very trivial change - just re-wording description:

```suggestion
# confirm_access: When true, Kiali will check if logged-in user has access to the Tracing service. If there is access, the Tracing integration will be shown in the Kiali UI.
```

status/versions.go
```diff
@@ -259,3 +264,21 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
+// Given a url and a user authentication token,
+// check if the user can access to the specific url.
+func canAccessURL(url, authToken string, auth *config.Auth) (bool, error) {
+	// If either url or authToken is empty, abort check
+	if url == "" || authToken == "" {
+		return false, nil
+	}
+
+	if auth.UseKialiToken {
+		auth.Token = authToken
```
jotak: we should not mutate `auth` as it's a pointer to the global config. If you just pass `auth` by value instead, it would be fine. Or explicitly make a copy of it before, and use that copy.

status/discover.go
```diff
@@ -136,10 +136,16 @@ func checkTracingService() (url string) {
 	return tracingConfig.URL
 }
 
-func DiscoverJaeger() string {
+// DiscoverJaeger will return the Jaeger URL if it has been configured.
+// AuthToken is the users auth token. When is not empty, it'll check if
+// user has access to Jaeger service.
+func DiscoverJaeger(authToken string, auth *config.Auth) string {
+	url := ""
+
```
jotak: The return path below is not covered by the authentication check. Since the state-defined URL can be accessed with different user tokens, it should be checked as well.

xeviknal:
@jotak, @aljesusg what do you think about this last approach? adding only the check into /jaeger.
Needs a PR on the console to set to redux the value from /jaeger instead of status. As it is done with Grafana.

handlers/jaeger.go
```diff
@@ -1,31 +1,76 @@
...
-	RespondWithJSON(w, http.StatusOK, info)
+	// Be sure to copy config.Auth and not modify the existing
+	auth := jaegerConfig.Auth
+	if auth.UseKialiToken {
+		auth.Token = requestToken
+	}
+
+	ha, code, err := canAccessURL(apiURL, &auth)
+	if !ha {
+		return nil, code, errors.New("can't reach Jaeger query service")
```
aljesusg: I would change this if by this because you are going to use this variables only in the if
```go
if ha, code, err := canAccessURL(apiURL, &auth);!ha {
		return nil, code, errors.New("can't reach Jaeger query service")
	} else if err != nil {
		return nil, http.StatusInternalServerError, err
	}
```

handlers/jaeger.go
```diff
@@ -1,31 +1,76 @@
...
+
+	info := &models.JaegerInfo{
+		URL: apiURL,
+	}
+
+	return info, http.StatusOK, nil
+}
+
+func canAccessURL(url string, auth *config.Auth) (bool, int, error) {
+	timeout := time.Duration(1000 * time.Millisecond)
```
jotak: nitpicking: I think you can write just `timeout := 1000 * time.Millisecond`

handlers/jaeger.go
```diff
@@ -1,31 +1,76 @@
...
+	_, code, _ := httputil.HttpGet(url, auth, timeout)
+
+	return code == 200, code, nil
```
jotak: it seems `error` is always `nil`, so maybe you can remove? (or make use of it)

jotak:
I've dismissed my review because it looks good (just 2 minor comments), but need to check with the frontend PR before approving

xeviknal:
@jotak @aljesusg @jmazzitelli changes are ready. Need to check for the frontend too.

handlers/jaeger.go
```diff
@@ -1,31 +1,73 @@
...
+
+	// Be sure to copy config.Auth and not modify the existing
+	auth := jaegerConfig.Auth
+	if auth.UseKialiToken {
+		auth.Token = requestToken
+	}
+
+	if ha, err := canAccessURL(apiURL, &auth); !ha {
+		return nil, http.StatusServiceUnavailable, errors.New("can't reach Jaeger query service")
+	} else if err != nil {
```
jotak: Actually, I'm not sure to understand this part. It would mean that we have a code 200 but still with an error? Is that even possible?

Also, when `ha` is false, here we don't forward the error to the console... I would prefer to send it, so that the user can troubleshoot more easily why his jaeger integration is turned off. For instance I hit the case currently, I have this error:  `Get http://tracing-istio-system.127.0.0.1.nip.io/jaeger: dial tcp 127.0.0.1:80: connect: connection refused` but in the console it only shows "can't reach Jaeger query service" (I have to check server logs to get more clues).

What do you think if we do just:
```go
if ha, err := canAccessURL(apiURL, &auth); !ha {
  return nil, http.StatusInternalServerError, err
}
```
?

lucasponce:
@xeviknal @skondkar Can you give an status of this one ?
It's basically our priority for v1.0 branch.

xeviknal:
@lucasponce development is finished. Helping @skondkar setting up the environment with both PRs.

mwringe:
> Foo user does not have access to integrations + receive the alerts like below:
>
> Could not fetch Grafana info. Turning off links to Grafana., Error: [ unknown error from Grafana (403) ]
> Could not fetch Jaeger info. Turning off Jaeger integration., Error: [ Service Unavailable ]

The whole point of this PR was to not have these kindof alerts.

jotak:
@mwringe the messages are much less visible than before, there is not that annoying popup that we see before every time we log in. But still, I don't think we want to turn Kiali into a black box when things are not working as intended. Missing permissions could be unintentional, a config issue, that the user would like to fix. This is also what Jay suggested in the JIRA ("moving this to informational and not displaying it in a popup on login"), IMO it makes sense to not totally hide the information.

lucasponce:
> IMO it makes sense to not totally hide the information.

+1

PR #1136, opened 2019-06-12
KIALI-3010 support http by allowing the Kiali CR to define empty identity cert/key files

jmazzitelli:
https://issues.jboss.org/browse/KIALI-3010

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -224,12 +224,16 @@ spec:
 #
 # Certificate file used to identify the file server. If set, you must go over https to access Kiali.
 # The operator will set these if it deploys Kiali behind https.
+# When left undefined, the operator will assign a cluster-specific cert file to provide https by default.
+# When set to an empty string, https will be disabled.
 #    ---
-#    cert_file: ""
+#    #cert_file:
 #
 # Private key file used to identify the server. If set, you must go over https to access Kiali.
+# When left undefined, the operator will assign a cluster-specific private key file to provide https by default.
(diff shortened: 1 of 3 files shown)


jmazzitelli:
How I tested:

```
deploy/deploy-kiali-operator.sh -oiv dev -kiv latest  -an "**" -gu "http://35.242.178.39:15031" -ju "http://35.242.178.39:15032"
```

(I had to use "dev" operator image version because I'm testing my local build of this PR -- cannot use latest because the latest up on quay.io doesn't have my fix :))

At this point, the server is configured for https because it got the default certs/keys as before - you will see in the kiali logs this message:

```
I0612 19:58:12.224980       1 server.go:55] Server endpoint will require https
```

And you will see in the configmap via `kubectl get cm kiali -n istio-system -o yaml` something like:

```
  identity:
      cert_file: /kiali-cert/tls.crt
      private_key_file: /kiali-cert/tls.key
```

This is all default behavior and how it has been working in the past.

Now edit the Kiali CR and set the identity cert/key files to empty string  `""`:

`kubectl edit kiali kiali -n kiali-operator`

and add this (under `spec:`)

```
spec:
  identity:
    cert_file: ""
    private_key_file: ""
```

Save the file and soon the operator will detect the change, and will change the configmap with the appropriate identity settings and will restart the Kiali pod. When the kiali pod restarts, look at the logs and you will NOT see the log message "Server endpoint will require https" this time. You will also see the identity cert_file and private_key_file flip to empty strings in the configmap.

lucasponce:
> Perhaps, that could be used in a flag from the operator script.

Nevermind, with the other PR, once we can pass a CR file on the script, that's resolved.

PR #1137, opened 2019-06-12
KIALI-2999 Fix issues with ServiceEntry identification

jshaughn:
Given a telemetry destination_servicename we attempt to determine if that
destination is actually a serviceEntry.  We do it by matching the service
name to the hosts defined for the accessible serviceEntrys.  I'm pretty sure
this PR does not cover all of the cases but it is an improvment over the
exact string match previously required.  This PR adds:
- wildcard handling for external service entries
- prefix handling for internal service entries with host set to the fully qualified service

This also fixes two bugs:
- we were failing to cache serviceEntry info in the appender globalInfo
- we called for the serviceDefinition of a serviceEntry (in badging appender), which does not exist

### DIFF
--- a/graph/telemetry/istio/appender/service_entry.go
+++ b/graph/telemetry/istio/appender/service_entry.go
@@ -10,9 +10,21 @@ import (
 
 const ServiceEntryAppenderName = "serviceEntry"
 
-// ServiceEntryAppender is responsible for identifying service nodes that are
-// Istio Service Entries.
+// ServiceEntryAppender is responsible for identifying service nodes that are Istio Service Entries.
 // Name: serviceEntry
+// Doc Links
+// - https://istio.io/docs/reference/config/networking/v1alpha3/service-entry/#ServiceEntry
+// - https://istio.io/docs/examples/advanced-gateways/wildcard-egress-hosts/
+//
+// A note about wildcard hosts. External service entries allow for prefix wildcarding such that
+// many different service requests may be handled by the same service entry definition.  For example,
+// host = *.wikipedia.com would match requests for en.wikipedia.com and de.wikipedia.com. The Istio
+// telemetry will produce separate service nodes for each distinct address. Nothing in the
+// graph code, including this appender, will aggregate the service nodes into one service entry node,
+// although each will be flagged as a serviceEntry due to the wildcard matching. On the plus side,
+// users can track request traffic to each distinct url. On the negative side, many distinct urls
+// will generate many distinct serviceEntry nodes in the graph.  Unless we get negative community
+// feedback on the current behavior we will stick with the current, simpler, approach.
 type ServiceEntryAppender struct {
 	AccessibleNamespaces map[string]time.Time
 }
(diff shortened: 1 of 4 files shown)


graph/telemetry/istio/appender/service_entry.go
```diff
@@ -72,14 +72,46 @@ func (a ServiceEntryAppender) getServiceEntry(service string, globalInfo *graph.
...
+			}
+		}
+		// handle wildcard
+		if serviceEntryHost.location == "MESH_EXTERNAL" {
+			hostTokens := strings.Split(serviceEntryHost.host, ".")
+			serviceNameTokens := strings.Split(serviceName, ".")
+			match := len(hostTokens) == len(serviceNameTokens)
+			if match {
+				for i, t := range hostTokens {
+					if t != "*" && t != serviceNameTokens[i] {
```
josejulio: Could a wildcard also be used for MESH_INTERNAL?
`*.my-namespace.*.*.*`

graph/telemetry/istio/appender/service_entry.go
```diff
@@ -72,14 +72,46 @@ func (a ServiceEntryAppender) getServiceEntry(service string, globalInfo *graph.
...
+		if serviceEntryHost.location == "MESH_INTERNAL" {
+			if strings.Split(serviceEntryHost.host, ".")[0] == serviceName {
+				return serviceEntryHost.location, true
+			}
+		}
+		// handle wildcard
+		if serviceEntryHost.location == "MESH_EXTERNAL" {
+			hostTokens := strings.Split(serviceEntryHost.host, ".")
+			serviceNameTokens := strings.Split(serviceName, ".")
+			match := len(hostTokens) == len(serviceNameTokens)
```
josejulio: Can the hostTokens be shorter than the serviceName?
e.g.
Would hostToken= ".my-domain.com" match serviceName="one.two.three.my-domain.com" ?
What about the other side, "mail." match serviceName="mail.google.com" ?

graph/telemetry/istio/appender/service_entry.go
jshaughn: Not as far as I know: see the "hosts" description [here](https://istio.io/docs/reference/config/networking/v1alpha3/service-entry/#ServiceEntry).

For MESH_INTERNAL I'm only trying to support exact match, or to match the serviceName portion of a full service.

graph/telemetry/istio/appender/service_entry.go
jshaughn: I actually don't know.  Perhaps a different approach is needed here.  It may be that the wildcard is allowed only at the beginning at that it should be able to match anything as long as the domain "tail" matches exactly.  Let me know what you think after you read up.

josejulio: Sounds good!

jshaughn: @josejulio After reading it again, and also [reading this](https://istio.io/docs/examples/advanced-gateways/wildcard-egress-hosts/), I think I should change the approach to just be a prefix wildcard that matches any prefix in front of the common domain "tail".   I'll make the change...

jshaughn: @josejulio I updated the wildcard logic to just handle prefix wildcarding. Please take a look.

PR #1140, opened 2019-06-13
KIALI-2924 option to uninstall operator if exists

jmazzitelli:
This now provides an option to uninstall the operator if it already exists.


jmazzitelli:
@lucasponce this appears to be working for me now. see what you think

lucasponce:
@jmazzitelli can we have an option to just "uninstall" ?
I mean, I want to clean my environment, check manually that things are ok, and also to resume later.

jmazzitelli:
> can we have an option to just "uninstall" ?

This PR now includes an option `--uninstall-mode true` (aka `-um true`) which does this.

PR #1141, opened 2019-06-14
KIALI-3023 be able to tell the operator not to install ingress/route

jmazzitelli:
https://issues.jboss.org/browse/KIALI-3023

If deployment.ingress_enabled is false, the route/ingress is not created on either k8s or openshift.

If ingress_enabled is false but you are using auth strategy of openshift, the operator will abort the install because OAuth Client needs the route.


jmazzitelli:
> * Install initially with ingress_enabled: true and then update to ingress_enabled: false

This test should now pass with this PR now

jmazzitelli:
> operator logged a fatal error for the conflict with OS auth but the kiali pod and route remained,

I'll assume the fatal error you are talking about is: "The auth.strategy is 'openshift' which requires a Route, but deployment.ingress_enabled is false. Aborting."

If so, then, yes, that error means your new Kiali CR is misconfigured and therefore the operator aborted and stopped before doing anything for that Kiali CR. You need to correct the configuration, save the Kiail CR, and try again. This is expected. With bad configuration like this, the operator shouldn't be expected to do anything. It isn't going to delete a route if your auth-strategy and ingress_enabled are in conflict (i.e. auth-strategy of openshift and ingress_enabled of false).

lucasponce:
@jmazzitelli is this required for v0.21 branch ?

Thanks

jmazzitelli:
> is this required for v0.21 branch ?

Yes, it is.  I'll label.

gbaufake:
@jmazzitelli 
So it will not create an ingress for OCP? Is that correct?

jmazzitelli:
> So it will not create an ingress for OCP? Is that correct?

Correct - it will give you an openshift Route as it has always done, but you will not get a k8s Ingress.

gbaufake:
@jmazzitelli 
On OCP4, you can have both, so I think you could create an ingress as well.

jmazzitelli:
you can, but I don't see why. You can in  OCP only because its a kubernetes thing. But OpenShift's Route feature provides Ingress features and more. I never hear of someone actually using k8s Ingress resources on OpenShift - that's what Routes are for.

PR #1145, opened 2019-06-14
make default grafana in_cluster_url dynamic

jmazzitelli:
Addresses something that was missed in https://github.com/kiali/kiali/pull/1120

We want the grafana URL to be dynamically generated since the istio namespace may be different.

### DIFF
--- a/operator/roles/kiali-deploy/defaults/main.yml
+++ b/operator/roles/kiali-deploy/defaults/main.yml
@@ -48,7 +48,7 @@ kiali_defaults:
         use_kiali_token: false
         username: ""
       display_link: true
-      in_cluster_url: "http://grafana.istio-system:3000"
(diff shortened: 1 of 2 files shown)


operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -48,7 +48,7 @@ kiali_defaults:
         use_kiali_token: false
         username: ""
       display_link: true
-      in_cluster_url: "http://grafana.istio-system:3000"
+      in_cluster_url: ""
```
jotak: Why didn't you write as you suggested in my PR : `http://grafana.{{ kiali_vars.istio-namespace }}:3000` ? It seems better to me ; so that we can have an empty in_cluster_url if it's what we want

PR #1146, opened 2019-06-14
KIALI-3025 Wrong Jaeger base query when defined

aljesusg:
jira: [kiali-3025](https://issues.jboss.org/browse/KIALI-3025)

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -237,12 +237,14 @@ spec:
 # service: The Kubernetes service name for tracing. Kiali uses this to connect within the cluster to Jaeger.
 # url: The URL that Kiali uses when integrating with Tracing. This URL must be accessible to clients external to
 #      the cluster in order for the integration to work properly. If empty, an attempt to auto-discover it is made.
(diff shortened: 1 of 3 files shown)


operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -237,12 +237,14 @@ spec:
...
 # url: The URL that Kiali uses when integrating with Tracing. This URL must be accessible to clients external to
 #      the cluster in order for the integration to work properly. If empty, an attempt to auto-discover it is made.
+# query_base_path: The value stored in QUERY_BASE_PATH in tracing service
 #    ---
 #    tracing:
 #      enabled: true
 #      namespace: ""
 #      service : ""
 #      url: ""
+#      query_base_path: ""
```
jmazzitelli: all of these settings are in alphabetical sort order. Move `query_base_path` after `namespace` and before `service`
Do the same with the comment line above.

Ex:

```
# query_base_path: The value stored in QUERY_BASE_PATH in tracing service
# url: The URL that Kiali uses when integrating with Tracing. This URL must be accessible to clients external to
#    ---
#    tracing:
#      enabled: true
#      namespace: ""
#      query_base_path: ""
#      service : ""
#      url: ""
```

PR #1150, opened 2019-06-14
explicitly patch the finalizer lists and delete Kiali CRs when we need to purge kiali

jmazzitelli:


jmazzitelli:
The one question I have is how are we expecting dependent operators to be installed?
This PR currently just has the Kiali CR in the kiali helm chart templates directory (along with the secret). The Kiali Operator CRD and its required roles, service accounts, etc are next to the maistra-operator/servicemesh-operator yaml files in the deploy/ directory. See: https://github.com/kiali/kiali/pull/1150/files#diff-9ed7738fd066fc54d1b80f0da319d8cf

But I don't know how/who will be installing these - I assume the person/process that installs the maistra operator CRD yaml will also install all other CRDs that are needed (which now include the Kiali CRD for the operator CR and for kiali dashboard CR).

If that is not how it is going to work, then this PR will need to change - the `charts/kiali/templates` directory will need to include the necessary clusterroles, service accounts, etc needed for the Kiali *operator* (not Kiali itself, those yaml template files have already been removed via this PR). But based on how I understand how things are going to work, I don't think we want to do it this way. I think we want to manage the kiali operator CRD in the same way as the maistra CRD.

I'll need input from the Maistra team here as for the proper direction this needs to go.

PR #1156, opened 2019-06-18
KIALI-2693 - Adding Automated Tests on Kiali Operator

gbaufake:
## Describe the change 

- Adding Operator Testing


gbaufake:
cc @jmazzitelli

operator/molecule/deploy/role.yaml
```diff
@@ -0,0 +1,282 @@
+---
+# NOTE: Changes to this file must be reflected in the OperatorHub.io CSV file
```
jmazzitelli: Is there something we can do here rather than hardcode the same role.yaml content that is found in the main operator/deploy directory? e.g., can we have a Makefile, and a "make test" will first copy the operator/deploy/role.yaml to some tmp directory and then the make target will run molecule with that file.

The same with the other files - looks like molecule/deploy just has copies of what is in operator/deploy (or at least a close copy - could probably be able to reuse them).

I really don't want to have to maintain another role.yaml (or other deploy yamls) - the ones in operator/deploy are authoritative and should be used if at all possible, rather than maintaining two separate copies (the "real" ones and the molecule ones).

gbaufake: Yes, 
I think I will need to some sort of envbust parse.

gbaufake:
Fixed @jmazzitelli.
Can you try it again? 
`molecule test -s multi-tenancy` should be enough for testing MT positive test case

PR #1162, opened 2019-06-19
KIALI-3042 keep only existing namespaces (including the deployment namespace) in the list

jmazzitelli:
Turns out, the operator already did the bulk of what we need - it only puts existing namespaces in the accessible_namespaces list in the configmap.

This PR does one additional thing though - it adds the deployment namespace to the list if it isn't already specified.


jmazzitelli:
To test this, in your Kiali CR do NOT put "istio-system" in the accessible_namespaces and confirm it makes it way into the configmap and you can see istio-system in Kiail.

PR #1164, opened 2019-06-19
KIALI-2859 Replace gometalinter with golangci-lint

israel-hdez:
https://issues.jboss.org/browse/KIALI-2859

Also, fixing the newly reported issues with the additional linter introduced by golangci-lint.

From the default linters set, only "errcheck" is disabled because it generates too many errors to fix under the scope of this PR/JIRA.

### DIFF
--- a/handlers/istio_config.go
+++ b/handlers/istio_config.go
@@ -51,7 +51,7 @@ func IstioConfigList(w http.ResponseWriter, r *http.Request) {
 			defer wg.Done()
 			// We don't filter by objects when calling validations, because certain validations require fetching all types to get the correct errors
 			istioConfigValidationResults, errValidations := business.Validations.GetValidations(namespace, "")
-			if errValidations != nil && err == nil {
+			if errValidations != nil && *err == nil {
 				*err = errValidations
 			} else {
 				if len(parsedTypes) > 0 {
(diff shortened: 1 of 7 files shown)


handlers/istio_config.go
```diff
@@ -51,7 +51,7 @@ func IstioConfigList(w http.ResponseWriter, r *http.Request) {
 			defer wg.Done()
 			// We don't filter by objects when calling validations, because certain validations require fetching all types to get the correct errors
 			istioConfigValidationResults, errValidations := business.Validations.GetValidations(namespace, "")
-			if errValidations != nil && err == nil {
+			if errValidations != nil && *err == nil {
```
israel-hdez: @lucasponce @mwringe This and the other similar changes look like a bug.
Not sure if it is worth to cherry-pick these changes.

jotak: I don't think it's worth cherry-picking as an hotfix, because anyway `err` will anyway never be `nil` so it's not going to panic due to potential dereference. Only problem is that `errValidations` is never going to be caught, I don't think it's an urgent fix.

But good catch anyway! I'm glad we can rely on a better linter soon :-)

jotak:
PS: we'll have to tell folks to bump their go and run `make lint-install` locally when it's merged

israel-hdez:
Follow up JIRA to enable errcheck linter: https://issues.jboss.org/browse/KIALI-3047

PR #1178, opened 2019-06-24
Adds istio-egressgateway to upstream installer

josejulio:
** Describe the change **

This adds istio-egressgateway to istio upstream hackscript. 
Maistra seems to install it by default, but using this script doesn't do the same, to test, verify that previous to this you get no egress.

I'm unsure if Kiali depends on it to be present or if we don't care.

Before:
```bash
$ oc get pod -l istio=egressgateway -n istio-system
No resources found.
```
After:
```bash
$ oc get pod -l istio=egressgateway -n istio-system
NAME                                   READY     STATUS    RESTARTS   AGE
istio-egressgateway-75964c5997-7dgj8   1/1       Running   0          3d
```

### DIFF
--- a/hack/istio/install-istio-kiali-via-helm.sh
+++ b/hack/istio/install-istio-kiali-via-helm.sh
@@ -29,6 +29,7 @@ KIALI_CREATE_SECRET="false"
 DASHBOARDS_ENABLED="false"
 USE_DEMO_VALUES="false"
 USE_DEMO_AUTH_VALUES="false"
+ISTIO_EGRESSGATEWAY_ENABLED="true"
 #HELM_REPO_TO_ADD="https://gcsweb.istio.io/gcs/istio-prerelease/daily-build/release-1.1-latest-daily/charts/"
 
 # process command line args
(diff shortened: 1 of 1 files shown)


jmazzitelli:
we can merge this if you want, but realize that you could have done this already by passing in the --set option to the script:

`install-istio-kiali-via-helm.sh --set gateways.istio-egressgateway.enabled=true`

You can merge this if you think devs are going to always want this egress all the time with no way of disabling it. Because as this PR is written, you will be getting the egressgateway always (i.e. you won't be able to test how Kiali behaves when egressgateways are disabled).

josejulio:
> we can merge this if you want, but realize that you could have done this already by passing in the --set option to the script:
> 
> `install-istio-kiali-via-helm.sh --set gateways.istio-egressgateway.enabled=true`
> 
> You can merge this if you think devs are going to always want this egress all the time with no way of disabling it. Because as this PR is written, you will be getting the egressgateway always (i.e. you won't be able to test how Kiali behaves when egressgateways are disabled).

I'm not really sure, but probably we should be able to test without the gateway but have this on by default (like maistra).

I'll extend this to be able to disable with command line args.

josejulio:
@jmazzitelli now is configurable, by default is enabled.

PR #1187, opened 2019-06-25
KIALI-3057 Fixing MeshPolicy error when user hasn't access to any namespace

xeviknal:
** Describe the change **

Users receive an error regarding to MeshPolicies when they don't have access to any namespace.

Screenshot of Kiali Console (47)

I rename the error to a proper error message. Kiali is unable to determine the mesh-wide mTLS status without accessing to, at least, one namespace. Removing the error is not a solution.
Once the user has one namespaces added, the MeshPolicy could be fetch. However, the total mTLS status might be inaccurate.

Screenshot of Kiali Console (51)

Goes with: https://github.com/kiali/kiali-ui/pull/1284

** Issue reference **

https://issues.jboss.org/browse/KIALI-3057

** Backwards incompatible? **
yes

### DIFF
--- a/business/tls.go
+++ b/business/tls.go
@@ -1,7 +1,6 @@
 package business
 
 import (
-	"fmt"
 	"sync"
 
 	"github.com/kiali/kiali/kubernetes"
@@ -45,7 +44,7 @@ func (in *TLSService) MeshWidemTLSStatus(namespaces []string) (models.MTLSStatus
 
 func (in *TLSService) hasMeshPolicyEnabled(namespaces []string) (bool, error) {
 	if len(namespaces) < 1 {
-		return false, fmt.Errorf("can't find MeshPolicies without a namespace")
+		return false, nil
 	}
 
 	// MeshPolicies are not namespaced. So any namespace user has access to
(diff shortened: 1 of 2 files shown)


mwringe:
>  Removing the error is not a solution.

Why is this not a solution? Not having access to a namespace is a valid scenario and we shouldn't be showing an error message to the user when this happens.

Is there not something else we can do here? Perhaps just not show the mtls status if there is no namespaces available.

lucasponce:
I agree with @xeviknal here.
Kiali query namespaces to fetch information about namespaces.
Having a user with access to 0 namespaces sounds a corner case to me, and probably it can happen due some unintential removing permissions.

To me sounds good to show a warning with an error "Kiali can access to N namespaces" rather than silent this.

pilhuhn:
Changing the message sounds good to me.

business/tls.go
```diff
@@ -45,7 +45,7 @@ func (in *TLSService) MeshWidemTLSStatus(namespaces []string) (models.MTLSStatus
 
 func (in *TLSService) hasMeshPolicyEnabled(namespaces []string) (bool, error) {
 	if len(namespaces) < 1 {
-		return false, fmt.Errorf("can't find MeshPolicies without a namespace")
+		return false, fmt.Errorf("Unable to determine mesh-wide mTLS status without access to a namespace")
```
pilhuhn: Can we reword from 'a' namespace to either
* 'any namespace' if we only need access to one to determine this
or
* name the namespace that is needed to be accessed for that operation?

xeviknal:
@mwringe it is a matter of deciding what we prefer to show. We can either not show the mtls status and silent this or show the warning message. I guess that if we show messages for grafana/jeager integrations disabled, we should show also messages in this scenario.

PR #1172, opened 2019-06-21
KIALI-3034 ServiceEntry nodes should represent an actual ServiceEntry

jshaughn:
Service entry nodes have previously represented the actual requested endpoint and not an actual serviceEntry.  For example, the node may represent requests for  "en.wikipedia.com".  But in reality the actual "service" handling the request is a serviceEntry with a host that matches "en.wikipedia.com".  Moreover its hosts may match many different service requests (e.g. "en.wikipedia.com", "de.wikipedia.con" may both be handled by serviceEntry named "wikipedia" with host "*.wikipedia.com".

This PR changes the handling to generate nodes that represent the serviceEntry, not just a matching destination.  The new node will aggregate the traffic for the different requested endpoints.

*** REQUIRES UI PR: https://github.com/kiali/kiali-ui/pull/1271

PR #1193, opened 2019-06-26
KIALI-3034 ServiceEntry nodes should represent an actual ServiceEntry (#1172)

jshaughn:
- aggregate se-host service nodes into one serviceEntry node
  - add destServices aggregation
- fix bug in graph.NewNode
- don't mark serviceEntries as inaccessible so that the UI can generate a link to the detail.
  - to specialCase serviceEntries test against the isServiceEntry metadata
- Handle non-leaf serviceEntry with outgoing traffic to an edgress gateway
- Fix 2 bugs(!) in edge aggregation (these affected serviceGraph generation as well)
- minor fix to prevent fetch of unknown service in badging appender


lucasponce:
I've reviewed and tested according v1.0 branch
image

PR #1195, opened 2019-06-26
update Makefiles to support devs working with CRC VM environments

jmazzitelli:


jmazzitelli:
I'm so close I can see the ribbon at the finish line. I believe all that is left to do here is to give the correct permissions to the Kiali and Kiali Operator Service Accounts so they can access the internal images pushed into the kiali namespace.

jmazzitelli:
This should be working, but I don't know why it is not. All permissions should be granted. Just have to test to figure out why it isn't.

jmazzitelli:
Ta-Da! :)

CRC now seems to be working for our typical local deployment environment/Makefile.

I am able to push images directly into the CRC cluster from our make targets.

When using the crc-openshift.sh hack script, the targets you need to use are:

`make crc-push` -- this builds and pushes the kiali and operator images into CRC VM's image registry via docker. You need to do this after you've changed the code and need to push the new instance to the CRC cluster.

`make crc-operator-create` -- this deploys the operator (but not kiali) to the CRC VM cluster, ensuring it uses the image that was pushed to the internal registry.

`make crc-kiali-create` -- this deploys Kiali to the CRC VM cluster, ensuring it uses the image that was pushed to the internal registry.

To undeploy Kiali, you use the the same ones as before (either `make openshift-undeploy` or the one in the operator Makefile - `make -C operator kiali-delete`). To remove the operator, again, it's the same as before: `make -C operator operator-delete`.

PR #1198, opened 2019-06-27
KIALI-3072 MTLS validations fixed when user doesn't have access to istio ns

xeviknal:
** Describe the change **
given a user without access to istio installation ns.
has access to at least one ns.
that ns has one service.

When user run the validations in service details page, validations crash.

Page now is up and also validations are returned.

** Issue reference **

https://issues.jboss.org/browse/KIALI-3072

** Backwards incompatible? **
yes

Screenshot from 2019-06-27 17-28-16


hhovsepy:
cc @skondkar

PR #1202, opened 2019-06-28
Add gogo/protobuf to glide.yaml (same version as in k8s)

jotak:
Also updating vendors

I'm doing this change because, almost everytime we run `make dep-update` (which is more or less a `glide up`), this `gogo/protobuf` gets updated.

This lib is essentially used by k8s / openshift clients. In vendor/k8s.io/api/Godeps/Godeps.json it's declared with a SHA (the same I'm setting in our glide.yaml). I have the impression that glide & godeps are not interacting well and we need to do that, but I'm not totally sure. Now I don't know how, going forward, we can keep this lib automatically up-to-date if it changes in k8s - we would have to update manually.

@jmazzitelli , @simonpasquier , do you think it's correct?


vendor/github.com/gogo/protobuf/.travis.yml
```diff
@@ -1,10 +1,7 @@
-apt:
-  update: true
-
 env:
   - PROTOBUF_VERSION=2.6.1
   - PROTOBUF_VERSION=3.0.2
-  - PROTOBUF_VERSION=3.7.1
+  - PROTOBUF_VERSION=3.2.0
```
jotak: Note, this seems to be a big step backward, but we were not aligned with what is used in k8s / openshift

jmazzitelli:
This seems right. Of course, I'm not sure these changes (particularly going backwards in that one lib version) will not break anything.

This would require some testing, obviously. Maybe @gbaufake can do his magic on this PR to see if his testing can flush out any problems.

jotak:
Ok.
@gbaufake (also @hhovsepy maybe) , I cannot give much clue about some specific stuff to test, the lib is a dependency of k8s client and to be honest I'm not sure it's ever used in the context of kiali, but if impacting anything it would probably be some call that kiali performs to the cluster API (listings, details, logs, ...). I've run some smoke tests and didn't detect any issue. Maybe some additional non-regression tests would be nice.

jotak:
thanks @hhovsepy

PR #1206, opened 2019-06-29
KIALI-3084 [operator] check create-cluster-role permissions if it needs them

jmazzitelli:
if operator does not have permissions to create cluster roles or cluster role bindings, abort if it is asked to support accessible_namespaces=**

https://issues.jboss.org/browse/KIALI-3084

This is important to check (and fail-fast when this condition occurs) because if the operator was not given  permissions when it was installed, you cannot then change the accessible namespaces setting to "**" and expect it to work. Without failing fast here, kiali will be left in a broken state when this condition happens.

### DIFF
--- a/operator/roles/kiali-deploy/tasks/main.yml
+++ b/operator/roles/kiali-deploy/tasks/main.yml
@@ -190,6 +190,17 @@
     role_kind: "{{ 'ClusterRole' if '**' in kiali_vars.deployment.accessible_namespaces else 'Role' }}"
     role_binding_kind: "{{ 'ClusterRoleBinding' if '**' in kiali_vars.deployment.accessible_namespaces else 'RoleBinding' }}"
 
+- name: Determine if the operator can support accessible_namespaces=**
+  vars:
+    role_json: "{{ lookup('k8s', api_version='rbac.authorization.k8s.io/v1', kind='ClusterRole', resource_name='kiali-operator') | default({}) }}"
+    query_cr: "rules[?apiGroups.contains(@, 'rbac.authorization.k8s.io')].resources.contains(@, 'clusterroles')"
+    query_crb: "rules[?apiGroups.contains(@, 'rbac.authorization.k8s.io')].resources.contains(@, 'clusterrolebindings')"
+  fail:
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Here is how I tested (these replication procedures use CRC and the crc hack script - if you want to test on some other cluster, go ahead. This should work anywhere. Just create the operator and kiali using the deploy-kiali-operator.sh script, wait for kiali to be installed, then edit the Kiali CR and set accessible namespaces to '**').

1. Start an openshift cluster using `hack/crc-openshift.sh start`
2. Log into the CRC VM using the "oc login" and "docker login" commands that you see in the output of `hack/crc-openshift.sh status` - this logs you into the CRC OS cluster as well as its image registry.
3. Build and push the new operator as well as kiali via `make crc-push`
4. Create a Kiali CR without accessible_namespace=** - I did this via `make crc-kiali-create`
5. Wait for the operator to install Kiali.
6. Now edit the Kiali CR via `oc edit kiali kiali -n kiali-operator` and add the setting `deployment.accessible_namespaces` making sure to set its value to `['**']`
7. Save the Kiali CR.

At this point, the Kiali operator is going to need to create a cluster role and cluster role binding for Kiali (in order to support accessible_namespaces=**). However, because you installed the operator initially withOUT accessible namespaces set to **, the operator doesn't have permissions to create cluster roles/bindings. Because of this, you will see the kiali operator abort quickly with the following error as seen in the operator pod logs:

```
FAILED! => {"changed": false, "msg": "The operator cannot support deployment.accessible_namespaces set to '**' because it does not have permissions to create clusterroles or clusterrolebindings"}
```

jmazzitelli:
this is built on top of https://github.com/kiali/kiali/pull/1208 (which is just a PR that fixes the crc makefile targets)

jmazzitelli:
@jshaughn would you like to review this? could be an opportunity for you to also try out the crc-openshift.sh hack script (see my test procedures in my earlier comment which uses it)

jshaughn:
@jmazzitelli OK, I will try, it will be interesting to see whether my VM has any chance of running OS4.

Issue #1196, opened 2019-06-27
Traffic graph and Response Codes didn't match


sniperking1234:
**Describe the bug**
![no3xx](https://user-images.githubusercontent.com/7834655/60226055-1da2e880-98bc-11e9-98c2-dfc3ebb86484.png)
![no3xx2](https://user-images.githubusercontent.com/7834655/60226058-1f6cac00-98bc-11e9-95cb-072dfe2bfd04.png)
duration=21600s
And the percentage also seems incorrect

**Versions used**
Kiali: 0.21.0
Istio: 1.1.2

pilhuhn:
Thanks for reporting this. Do you have a way, we can easily reproduce this? E.g. some specific VS/DR for bookinfo?

lucasponce:
cc @israel-hdez

sniperking1234:
@pilhuhn I think this is a problem of float precision when the traffic rates is very small.

jshaughn:
Yes, when dealing with very low request rates the rate can get rounded to 0 as we do use a precision of 2 .  The response percentages have a precision of 1 but the numbers are relative to each other and only suffer a 0 if the slice is tiny.  I don't think there is much we can do here, luckily there is some detail to be found in the Responses.  We are likely going to introduce a FAQ for the graph, this would make a good entry.

israel-hdez:
Ok, looking at the first image, I understand that the %success = 66% of the traffic tab is not close to the 72% of the 200 code of the Response codes tab, which is misleading.

What I see weird is that even when numbers in both tabs are being calculated from the same queries, numbers don't add up to a close value. So, yes, the floating point precision is causing quite bad results in this case.

I think it will be hard to make numbers in both tabs to match. But if the current deviation is a concern, I think there are ways to reduce the precision error to, at least, get numbers that are closer. Although, I'm not sure how much we can improve it.

sniperking1234:
I tried to fix this problem @jshaughn @israel-hdez

jshaughn:
Did this proposed code change solve your problem?  Do you have an "after" screenshot with this change applied?

sniperking1234:
@josejulio I think the change can solve it.
![image](https://user-images.githubusercontent.com/7834655/60406430-5f979b80-9be8-11e9-98dc-ba486793261f.png)
![image](https://user-images.githubusercontent.com/7834655/60406438-6b835d80-9be8-11e9-867d-0f9314e858a7.png)
[...]

PR #1207, opened 2019-06-29
fix precision about traffic rate

sniperking1234:
** Describe the change **

Fix traffic graph and Response Codes didn't match when dealing with very low request rates

** Issue reference **

https://github.com/kiali/kiali/issues/1196

** Backwards incompatible? **

no

### DIFF
--- a/graph/config/cytoscape/cytoscape.go
+++ b/graph/config/cytoscape/cytoscape.go
@@ -277,7 +277,8 @@ func addNodeTelemetry(n *graph.Node, nd *NodeData) {
 				if protocolTraffic.Rates == nil {
 					protocolTraffic.Rates = make(map[string]string)
 				}
-				// handle low request rates precision trouble
+				// handle low request rates precision trouble.
+				// If rate < 0 then increase precision
 				delta := calPrecisionDelta(rateVal)
 				rate := fmt.Sprintf("%.*f", r.Precision+delta, rateVal)
 				// remove redundant zero and keep at least two decimals
@@ -330,6 +331,7 @@ func addEdgeTelemetry(e *graph.Edge, ed *EdgeData) {
 					protocolTraffic.Rates = make(map[string]string)
 				}
 				// handle low request rates precision trouble
+				// If rate < 0 then increase precision
 				delta := calPrecisionDelta(rateVal)
 				rate := fmt.Sprintf("%.*f", r.Precision+delta, rateVal)
 				// remove redundant zero and keep at least two decimals
--- a/graph/config/cytoscape/cytoscape.go
+++ b/graph/config/cytoscape/cytoscape.go
@@ -278,7 +278,7 @@ func addNodeTelemetry(n *graph.Node, nd *NodeData) {
 					protocolTraffic.Rates = make(map[string]string)
 				}
 				// handle low request rates precision trouble.
-				// If rate < 0 then increase precision
+				// If rate < 1 then increase precision
 				delta := calPrecisionDelta(rateVal)
 				rate := fmt.Sprintf("%.*f", r.Precision+delta, rateVal)
 				// remove redundant zero and keep at least two decimals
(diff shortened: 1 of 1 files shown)


rhqci:

pilhuhn:
ok to test

graph/config/cytoscape/cytoscape.go
```diff
@@ -275,7 +277,12 @@ func addNodeTelemetry(n *graph.Node, nd *NodeData) {
 				if protocolTraffic.Rates == nil {
 					protocolTraffic.Rates = make(map[string]string)
 				}
-				protocolTraffic.Rates[string(r.Name)] = fmt.Sprintf("%.*f", r.Precision, rateVal)
+				// handle low request rates precision trouble.
+				// If rate < 0 then increase precision
```
abonas: rate < 0? 
why not <= 0?

graph/config/cytoscape/cytoscape.go
```diff
@@ -438,3 +450,12 @@ func generateGroupCompoundNodes(appBox map[string][]*NodeData, nodes *[]*NodeWra
 		}
 	}
 }
+
+func calPrecisionDelta(rateVal float64) int {
+	delta := 0
+	if rateVal < 1 {
+		base := int(1 / rateVal)
+		delta = len(strconv.Itoa(base))
```
abonas: is there way to achieve a manipulation on a number without converting it to string?

graph/config/cytoscape/cytoscape.go
sniperking1234: I made a mistake. It should be rate < 1

graph/config/cytoscape/cytoscape.go
sniperking1234: That can only be done in a loop

jotak: You could use `log10(...) + 1` to get the same value with only maths.

graph/config/cytoscape/cytoscape.go
```diff
@@ -438,3 +450,12 @@ func generateGroupCompoundNodes(appBox map[string][]*NodeData, nodes *[]*NodeWra
 		}
 	}
 }
+
+func calPrecisionDelta(rateVal float64) int {
+	delta := 0
+	if rateVal < 1 {
+		base := int(1 / rateVal)
```
jotak: I know that at the moment this function is only called for a rate that is strictly positive, but nevertheless can you add a check `&& rateVal > 0` ? That would prevent bad surprises if it changes in the future

jotak:
BTW, in other parts of Kiali we're rounding significant decimals directly from within the promql query : https://github.com/kiali/kiali/blob/master/prometheus/metrics.go#L249-L251 . This isn't used for the graph afaik, but it's an alternative way to achieve the same, I think. Not sure however how easy it would be to use that in graph code.

Just saying as an information... To me, the proposed fix looks good, but @jshaughn or @israel-hdez are better placed to approve.

graph/config/cytoscape/cytoscape.go
sniperking1234: check added

israel-hdez:
nice approach @sniperking1234 
rounding was made to keep the generated JSON small and this looks like a good trade-off for small values.

one question: do you see the same issue when clicking an edge?
The `addEdgeTelemetry` function of the file you modified does something very similar. So, I wonder if you would like to apply the same fix there.

graph/config/cytoscape/cytoscape.go
```diff
@@ -275,7 +277,12 @@ func addNodeTelemetry(n *graph.Node, nd *NodeData) {
 				if protocolTraffic.Rates == nil {
 					protocolTraffic.Rates = make(map[string]string)
 				}
-				protocolTraffic.Rates[string(r.Name)] = fmt.Sprintf("%.*f", r.Precision, rateVal)
```
jshaughn: Clever work.  Although I'd like to see the special handling only for rateVal < 1, otherwise just keep the simple Sprintf.  If you prefer just hide everything in a func, maybe like:

`   protocolTraffic.Rates[string(r.Name)] = rateToString(r.Precision, rateVal)`

sniperking1234:
@israel-hdez I have modified both the `addEdgeTelemetry ` function and `addNodeTelemetry` function.

graph/config/cytoscape/cytoscape.go
sniperking1234: It's a good way to keep code clean. I have modified it.

PR #1213, opened 2019-07-02
Kiali-GH-1196 fix small rate handling

jshaughn:
Earlier today we merged https://github.com/kiali/kiali/pull/1207 but when running things it seemed the fix may not be quite right.  I added some tests and as a result I've tweaked the algorithm.  It's not quite as slick but seems to work.

### DIFF
--- a/graph/config/cytoscape/cytoscape.go
+++ b/graph/config/cytoscape/cytoscape.go
@@ -17,9 +17,7 @@ package cytoscape
 import (
 	"crypto/md5"
 	"fmt"
-	"math"
 	"sort"
-	"strings"
 
 	"github.com/kiali/kiali/graph"
 )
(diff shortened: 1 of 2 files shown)


jshaughn:
> I don't like the "brute force" approach with the loop...

I know, it's less slick, but better results and if it makes you feel better, look at what log10 does under the covers ;)

graph/config/cytoscape/cytoscape.go
```diff
@@ -441,20 +439,26 @@ func generateGroupCompoundNodes(appBox map[string][]*NodeData, nodes *[]*NodeWra
...
+	}
+
+	return fmt.Sprintf("%.*f", precision, rateVal)
 }
 
-func calPrecisionDelta(rateVal float64) int {
-	delta := 0
-	if rateVal < 1 && rateVal > 0 {
-		base := 1 / rateVal
-		delta = int(math.Log10(base) + 1)
```
jotak: At first, he was doing `delta = len(strconv.Itoa(base))` instead of `log10` ... would it have been better with the first approach?

jshaughn: I'm not sure without running the tests against that first approach.  I'm OK as is, but I'm fine with a different algo too if you think it's better.  If you or he want to try it that's fine with me, otherwise  let's just press on :)

As an aside, it was a mistake for me to accept the PR without some sort of tests.

jotak: I'm reading the tests, and to be honest I'm not sure to understand exactly why `calcPrecision(0.05, 5)` should be `1` and not `2` as before ? It means that we're loosing precision for any number where significant digit is >5, looks like an arbitrary rule, no?

However I see there's an error at boundaries, with the previous formula: when the value is like `0.0(...)01` we shouldn't add 1 as we do ; and your impl does not have this issue

Anyway that's nitpicking ; i think in any case, the initial problem is resolved.

israel-hdez: My understanding is that `calcPrecision(0.05, 5)` is 1, because 0.05 can be rounded up to 0.1. So, it's enough a precision of 1.

jotak: I still think it's weird:

- 0.01 is 0.01
- 0.02 is 0.02
- 0.03 is 0.03
- 0.04 is 0.04
but
- 0.05 is 0.1
- 0.06 is 0.1

what's the rationale? I don't get it.

jshaughn: OK, let's make this more consistent as the rounding approach here is making it confusing.  I'll revisit right now...

jshaughn: Updated in https://github.com/kiali/kiali/pull/1222

PR #1216, opened 2019-07-04
KIALI-3081 Alternative Auth POC (Token auth)

israel-hdez:
This is not meant to be merged, just a PoC.

This is adding a login mechanism similar to how the [kubernetes dashboard](https://github.com/kubernetes/dashboard/wiki/Access-control#login-view) works (the "token" one).

So, instead of providing a username and a password, Kiali asks for a token. The provided token is meant to be a service account token. Once you provide a valid token, kiali uses that for the session. This works fine when using Kiali with Openshift and works so-so with kubernetes because of RBAC differences. 

image

Adding new auth methods looks easy, but probably, some abstractions are required to make this more "pluggable" and to change less files, and also to make the auth API accept all auth kinds under the same endpoint.

In the UI side (see https://github.com/kiali/kiali-ui/pull/1297) also multiple files needs to be changed to add more auth methods. The LoginDispatcher looks like the abstracted interface. But the login form still needs to be modified and take into account the different auth methods, so probably its still not optimal.

Operator also validates the auth method set and rejects to setup kiali if there is not a valid auth method. The operator, probably, should not validate the auth method and let kiali display an error. Else, it's hard to find the error on all those operator logs.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -90,10 +90,12 @@ const (
 	AuthStrategyOpenshift = "openshift"
 	AuthStrategyLogin     = "login"
 	AuthStrategyAnonymous = "anonymous"
+	AuthStrategyToken     = "token"
 
 	TokenCookieName             = "kiali-token"
 	AuthStrategyOpenshiftIssuer = "kiali-openshift"
 	AuthStrategyLoginIssuer     = "kiali-login"
+	AuthStrategyTokenIssuer     = "kiali-token"
 
 	// These constants are used for external services auth (Prometheus, Grafana ...) ; not for Kiali auth
 	AuthTypeBasic  = "basic"
--- a/config/token.go
+++ b/config/token.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"github.com/kiali/kiali/log"
 	"time"
 
 	"github.com/dgrijalva/jwt-go"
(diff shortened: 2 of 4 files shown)


jmazzitelli:
I know this is just a PoC and you say it is not to be merged - but in case it grows up to be a real boy, need to update the kiali_cr.yaml comments in the CR yaml and the deploy script

kiali_cr.yaml https://github.com/kiali/kiali/blob/v1.0.0/operator/deploy/kiali/kiali_cr.yaml#L69-L75
deploy sh: https://github.com/kiali/kiali/blob/v1.0.0/operator/deploy/deploy-kiali-operator.sh#L117-L122

as well as changes to the deploy script for this new value:

help text: https://github.com/kiali/kiali/blob/v1.0.0/operator/deploy/deploy-kiali-operator.sh#L368
default prompts: https://github.com/kiali/kiali/blob/v1.0.0/operator/deploy/deploy-kiali-operator.sh#L429-L436
validate: https://github.com/kiali/kiali/blob/v1.0.0/operator/deploy/deploy-kiali-operator.sh#L894

jmazzitelli:
> The operator, probably, should not validate the auth method and let kiali display an error. Else, it's hard to find the error on all those operator logs.

IMO If the operator detects a misconfiguration (especially if the strategy is just invalid; but this is true not just for auth_strategy - if anything is misconfigured to cause kiali to just not function) it should fail fast and not install all of the Kial resources. You should not create all the Kiali resources in a way that we know is broken. For example, it seems odd to have the operator say successfully installed kiali only for the users not to even be able to login, and the operator knows it.

The operator logs are there to tell you what is wrong - you can see the logs in the OpenShift console UI which should be fine.

PR #1220, opened 2019-07-05
Vendors update: upgraded k-charted to get bug fix (KIALI-2985)

jotak:
Frontend PR: https://github.com/kiali/kiali-ui/pull/1299


jotak:
I don't know why the build is failing. It says that swagger file is valid ... however the file comparison with generated file doesn't match.
When running locally on my machine, it matches.
Did someone ever see that?

PR #1227, opened 2019-07-09
Fixed a couple of descriptions in the reference CR

jpkrohling:
A couple of places in the Jaeger-specific parts had references to Grafana and Prometheus, when they really meant Jaeger. This PR fixes that, as well as a similar problem under the Grafana properties.

Signed-off-by: Juraci Paixão Kröhling <juraci@kroehling.de>


rhqci:

PR #1230, opened 2019-07-10
KIALI-3115 Check TargetPort with string and port number, fixes #1228

### LINKED ISSUE #1228: Service port and deployment port do not match
**Describe the bug**

Probably not really a bug, but something is obviously wrong with my configuration.
Kiali says the port configuration of my service is not valid.

kiali_invalid_port_config

I mean obviously, the service and container port don't match, but that's what a service is for at the end? Why is it a problem from kialis point of view?

So the service is configured to listen on port 80 TCP. Port ist named http. Mapped to targetPort 8080.
```yaml
spec:
  selector:
    app: auth
  ports:
    - protocol: TCP
      port: 80
      targetPort: 8080
      name: http
```
The corresponding pod listens on 8080.

```yaml
        ports:
        - containerPort: 8080
```

**Versions used**
Kiali: 0.20.0
Istio: 1.2.2
Kubernetes flavour and version: 1.13.7

**To Reproduce**

1. Visit "Services" in the UI
2. View details of a service

**Expected behavior**
Port configuration is recoginzed as valid.

burmanm:
** Describe the change **

Modify port validation to allow TargetPort to override Port definition (default is Port == TargetPort). In case of string in the TargetPort, search by the Port name in the deployment definition.

** Issue reference **

KIALI-3115, github issue #1228 

** Backwards incompatible? **

Yes

** Documentation **

Will need kiali.io update to the validation page, kiali/kiali.io#154

### DIFF
--- a/business/checkers/services/port_mapping_checker.go
+++ b/business/checkers/services/port_mapping_checker.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/intstr"
+
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 	apps_v1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


hhovsepy:
Verified, including the validations and Error message:
1. targetPort is integer, compared with containerPort.
2. targetPort is string, compared with container port name.

lucasponce:
Is this affecting to v1.0 ?

burmanm:
Yes, the same bug / missing feature is in 1.0 also.

PR #1234, opened 2019-07-10
KIALI-3102 Always add the protocol to the edge

josejulio:
** Describe the change **

This lets the UI know the protocol an edge has, even if there is no traffic.

** Issue reference **

[KIALI-3102](https://issues.jboss.org/browse/KIALI-3102)

** Backwards incompatible? **

No


josejulio:
I added the do-not-merge tag because i'm still testing if there is any bug with this changes.

PR #1244, opened 2019-07-15
[PoC]: When kubernetes cache is enabled, create a cluster role to help with cache operation

jmazzitelli:
_Just a PoC. Do not merge._

cluster role for get/list/watch on namespaces for kubernetes and namespaces/projects for openshift is created if the cache is enabled.

This also gives Kiali project list and and project watch permissions on its non-cluster openshift role.

@burmanm I have not attempted to built, run, or test this :) I have no idea if this is what you would need. But all this does is give cluster access to namespace and projects only (literally just those 2 resources - there is no cluster-wide access to the resources in those namespaces/projects - which I think you might need. If so, this PR is useless).  If the cache DOES need full cluster-wide access to all these resources inside namespaces/projects, then that cache will need to be implemented differently because we will not be allowed to create such cluster roles when in "multitenancy mode" (i.e. accessible_namespaces NOT set to `[**]`). To get this cluster-wide access to all resources in all namespaces/projects, you will need to install with "accessible_namespaces" set to `[**]` but this is essentially assuming a single cluster-wide mesh and is not how Maistra will deploy Kiali.

Another note: if you have multiple Kialis installed in the cluster, then the cache enabled flag needs to be the same value for all of them, or else you will get indeterminate results (either the cluster roles will exist or not depending on which Kiali you last installed). For example, if you install kiali with cache_enabled == False, the cluster roles will be removed. If you then follow up by installing a Kiali with cache_enabled=True, the cluster roles will be created. So the last Kiali installed wins (either the cluster roles will be removed or created depending on the cache_enabled flag of the last kiali installed).

Related to https://github.com/kiali/kiali/issues/1386

### DIFF
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -269,6 +269,8 @@ rules:
   - projects
   verbs:
   - get
+  - list
+  - watch
 - apiGroups: ["route.openshift.io"]
   resources:
   - routes
--- a/operator/roles/kiali-deploy/tasks/kubernetes/k8s-main.yml
+++ b/operator/roles/kiali-deploy/tasks/kubernetes/k8s-main.yml
@@ -50,3 +50,16 @@
   when:
   - is_k8s == True
   - '"**" not in kiali_vars.deployment.accessible_namespaces'
+
+- name: Create cache cluster roles on Kubernetes if required
+  include_tasks: process-resource.yml
+  vars:
+    process_resource_cluster: "kubernetes"
+  loop:
+  - cache-clusterrole
+  - cache-clusterrolebinding
+  loop_control:
+    loop_var: process_resource_item
+  when:
+  - is_k8s == True
+  - kiali_vars.kubernetes_config.cache_enabled == True
--- a/operator/roles/kiali-deploy/tasks/main.yml
+++ b/operator/roles/kiali-deploy/tasks/main.yml
@@ -327,6 +327,13 @@
   - '"**" in current_accessible_namespaces'
   - '"**" not in kiali_vars.deployment.accessible_namespaces'
 
+- name: Delete Kiali cache cluster roles if no longer required
+  vars:
+    role_namespace: "{{ kiali_vars.deployment.namespace }}"
+  include_tasks: remove-cache-clusterroles.yml
+  when:
+  - kiali_vars.kubernetes_config.cache_enabled != True
+
 - name: Delete all Kiali roles from namespaces if view_only_mode is changing since role bindings are immutable
   include_tasks: remove-roles.yml
   loop: "{{ kiali_vars.deployment.accessible_namespaces }}"
--- a/operator/roles/kiali-deploy/tasks/openshift/os-main.yml
+++ b/operator/roles/kiali-deploy/tasks/openshift/os-main.yml
@@ -51,6 +51,19 @@
   - is_openshift == True
   - '"**" not in kiali_vars.deployment.accessible_namespaces'
 
+- name: Create cache cluster roles on OpenShift if required
+  include_tasks: process-resource.yml
+  vars:
+    process_resource_cluster: "openshift"
+  loop:
+  - cache-clusterrole
+  - cache-clusterrolebinding
+  loop_control:
+    loop_var: process_resource_item
+  when:
+  - is_openshift == True
+  - kiali_vars.kubernetes_config.cache_enabled == True
+
 - name: Process Kiali OAuth client on OpenShift
   include: openshift/os-oauth.yml
   when:
--- a/operator/roles/kiali-deploy/tasks/remove-cache-clusterroles.yml
+++ b/operator/roles/kiali-deploy/tasks/remove-cache-clusterroles.yml
@@ -1,0 +1,25 @@
+- name: Delete cache-related Kiali roles that are cluster-scoped
+  ignore_errors: yes
+  k8s:
+    state: absent
+    api_version: "{{ k8s_item.apiVersion }}"
+    kind: "{{ k8s_item.kind }}"
+    namespace: "{{ role_namespace }}"
+    name: "{{ k8s_item.metadata.name }}"
+  register: delete_result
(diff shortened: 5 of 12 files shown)


jmazzitelli:
This setting turns on and off the cache - if this is enabled (true) then the cache cluster roles are created. If disabled (false) then the cache cluster roles are removed if they exist (so no cache cluster roles will exist):

https://github.com/kiali/kiali/blob/v1.1.0/operator/deploy/kiali/kiali_cr.yaml#L316

burmanm:
@jmazzitelli Are you aware if somethinghas  changed in the master related to this PR lately? I found out that I'm still getting (after rebase) the fetching errors for OS v1.ProjectList (system:anonymous has no rights etc) when trying to continue the caching work.

jmazzitelli:
> Are you aware if somethinghas changed in the master related to this PR lately?

I can't think of any changes that would affect this. "system:anonymous has no rights" - not sure why its saying "system:anonymous" - sounds like whatever made a request didn't have a service account associated with it??? Make sure that cache clusterrole and its binding are created properly and assigned to the kiali service account (which they should be - that's what this PR does).

burmanm:
Yeah, that appears to have been the case - the kiali-cache role is no longer (in my branch) created. Maybe some rebase issue, I'll have to investigate more.

burmanm:
Right, the operator fails (but gives me no clue..):

```
{
    "EventData.FailedTaskPath": "/opt/ansible/roles/kiali-deploy/tasks/process-resource.yml:1",
    "EventData.Task": "Create resource [cache-clusterrole] on [openshift]",
    "EventData.TaskArgs": "_ansible_version=2.7.10, definition=apiVersion: rbac.authorization.k8s.io/v1\nkind: ClusterRole\nmetadata:\n  name: kiali-cache\n  labels:\n    app: kiali\n    version: dev\nrules:\n- apiGroups: [\"\"]\n  resources:\n  - namespaces\n  verbs:\n  - get\n  - list\n  - watch\n- apiGroups: [\"project.openshift.io\"]\n  resources:\n  - projects\n  verbs:\n  - get\n  - list\n  - watch\n, _ansible_selinux_special_fs=['fuse', 'nfs', 'vboxsf', 'ramfs', '9p'], _ansible_no_log=False, _ansible_module_name=k8s, _ansible_debug=False, _ansible_verbosity=2, _ansible_keep_remote_files=False, _ansible_syslog_facility=LOG_USER, _ansible_socket=None, state=present, _ansible_diff=False, _ansible_remote_tmp=~/.ansible/tmp, _ansible_shell_executable=/bin/sh, _ansible_check_mode=False, _ansible_tmpdir=/opt/ansible/.ansible/tmp/ansible-tmp-1568020277.9-139779556739932/",
    "error": "[playbook task failed]",
    "event_type": "runner_on_failed",
    "gvk": "kiali.io/v1alpha1, Kind=Kiali",
    "job": "8190398549307930729",
    "level": "error",
    "logger": "logging_event_handler",
    "msg": "",
    "name": "kiali",
    "namespace": "kiali-operator",
    "stacktrace": "github.com/go-logr/zapr.(*zapLogger).Error\n\tpkg/mod/github.com/go-logr/zapr@v0.1.1/zapr.go:128\ngithub.com/operator-framework/operator-sdk/pkg/ansible/events.loggingEventHandler.Handle\n\tsrc/github.com/operator-framework/operator-sdk/pkg/ansible/events/log_events.go:84",
    "ts": 1568020279.0601788
}
```

burmanm:
Ah, it was hidden.. apparently the operator needs more rights.

```
fatal: [localhost]: FAILED! => {"attempts": 6, "changed": false, "error": 403, "msg": "Failed to create object: {\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"clusterroles.rbac.authorization.k8s.io is forbidden: User \\\"system:serviceaccount:kiali-operator:kiali-operator\\\" cannot create clusterroles.rbac.authorization.k8s.io at the cluster scope: no RBAC policy matched\",\"reason\":\"Forbidden\",\"details\":{\"group\":\"rbac.authorization.k8s.io\",\"kind\":\"clusterroles\"},\"code\":403}\n", "reason": "Forbidden", "status": 403}
```

burmanm:
Yarr spamming, but this helps in any case to that issue:

```diff
diff --git a/operator/deploy/role.yaml b/operator/deploy/role.yaml
index fffc3e15..40b43dbd 100644
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -73,6 +73,8 @@ rules:
   resources:
   ${OPERATOR_ROLE_CLUSTERROLEBINDINGS}
   ${OPERATOR_ROLE_CLUSTERROLES}
+  - clusterroles
+  - clusterrolebindings
   - rolebindings
   - roles
   verbs:
```

burmanm:
Although it didn't fix the original issue even when the kiali-cache role now exists.

jmazzitelli:
```
   ${OPERATOR_ROLE_CLUSTERROLEBINDINGS}
   ${OPERATOR_ROLE_CLUSTERROLES}
+  - clusterroles
+  - clusterrolebindings
```
Well, that's the point of the ${OPERATOR_ROLE_xxx}... those are set to `- clusterroles` and `-clusterrolebindings` when it is determined the operator needs those super-user rights. We do not blindly give the operator those rights, hence why they aren't hardcoded but we use the env vars instead.

You'll see this here - these rights are required when accessible_namespaces are set to ** - in that case, the operator needs the ability to create clusterroles/bindings. Otherwise, we do not grant the operator those rights - see https://github.com/kiali/kiali/blob/master/operator/deploy/deploy-kiali-operator.sh#L732-L737

lucasponce:
Note that with PR https://github.com/kiali/kiali/pull/1749 a cluster role won't be necessary.

lucasponce:
As PR #1749 and others are merged, I think this one is outdated.

So, please, @jmazzitelli feel free to close, at least from my side.

PR #1248, opened 2019-07-16
KIALI-3128 be able to specify imagePullSecrets in the Deployment 

gbaufake:
Porting back to 1.0

### DIFF
--- a/operator/roles/kiali-deploy/defaults/main.yml
+++ b/operator/roles/kiali-deploy/defaults/main.yml
@@ -29,6 +29,7 @@ kiali_defaults:
     #additional_service_yaml:
     image_name: "kiali/kiali"
     image_pull_policy: "IfNotPresent"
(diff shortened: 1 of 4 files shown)


operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -29,6 +29,7 @@ kiali_defaults:
     #additional_service_yaml:
     image_name: "kiali/kiali"
     image_pull_policy: "IfNotPresent"
+    image_pull_secrets: []
```
lucasponce: The default value will be always an empty array ?

jmazzitelli: yes. by default you have no secrets define (and none are required to pull from docker.io or quay.io). You can optionally set this to one or more secret names if you need to pull images from a private image registry.

PR #1250, opened 2019-07-16
KIALI-3130 have operator installer support a dry run mode

jmazzitelli:
The yaml will be written to a file rather than sent to the cluster. The yaml can be used to install the operator at a later time.

### DIFF
--- a/operator/deploy/deploy-kiali-operator.sh
+++ b/operator/deploy/deploy-kiali-operator.sh
@@ -648,7 +648,7 @@ if [ "${KIALI_IMAGE_VERSION:-lastrelease}" == "lastrelease" ]; then
     echo ${github_api_url}
     exit 1
   fi
-  echo "Will use the last Kiali operator release: ${kiali_version_we_want}"
+  echo "Will use the last Kiali release: ${kiali_version_we_want}"
   KIALI_IMAGE_VERSION=${kiali_version_we_want}
 else
   if [ "${KIALI_IMAGE_VERSION}" == "latest" ]; then
--- a/operator/deploy/deploy-kiali-operator.sh
+++ b/operator/deploy/deploy-kiali-operator.sh
@@ -23,6 +23,12 @@
 # -----------
 # Environment variables that affect the overall behavior of this script:
 #
+# DRY_RUN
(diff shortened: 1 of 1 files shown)


jmazzitelli:
All the deploy script options are still valid - but if you supply `--dry-run <output .yaml file>` all the resources that would have been created are not created but instead the yaml is written to that output yaml file.

The idea is that you can use that yaml file to install the operator.

The real reason for this is we want to be able to product an "uber-yaml" file that we can publish for people to use to install the Kiali Operator without having to run the script. This will be useful when Kiali is released and we want people to be able to install a fixed version of the operator with a fixed set of options already enabled/disabled (such as not setting accessible_namespaces to `**` and not giving the operator permissions to create cluster roles, etc.).

The deploy script will still be very useful because it allows users to customize the kiali operator's configuration. This dry run uber-yaml file will simply be a fixed operator config for a particular version with a particular default configuration.

jmazzitelli:
If you want to produce a yaml file that just has yaml for the operator (and not for a Kiali CR) make sure you pass in `--operator-install-kiali false` in addition to the `--dry-run <your .yaml file>` option.

jmazzitelli:
@gbaufake see if this produces a single yaml file that you can use. I don't want to merge this until I hear from you that it does what you need it to do.

jmazzitelli:
> There was message `Will use the last Kiali operator release: v1.1.0`

Opp.. this is just a copy-n-paste error that as been here for a while. I'll fix it in this PR. Notice:

https://github.com/kiali/kiali/blob/v1.1.0/operator/deploy/deploy-kiali-operator.sh#L592-L612

line 592 is the message about the operator version and line 612 is the message about the kiali version - but like 612 is just a copy-n-paste of 592. It should be its own message (should not mention operator).

jmazzitelli:
> The yaml seems to be correct, but it seems we will need to change because of https://issues.jboss.org/browse/KIALI-3133.

We are going to handle this separately. This is due to a problem with a productized version (not an upstream version) so we will not change the behavior of this PR. However, the workaround for that productized issue will be to manually remove the offending 'ansible' container from the uber-yaml and that would be the version of the uber-yaml we publish. So, in short, nothing to be done here in this PR.

PR #1254, opened 2019-07-17
KIALI-3136 - use deployment.secret_name

jmazzitelli:
https://issues.jboss.org/browse/KIALI-3136


jmazzitelli:
I confirmed this works now.

PR #1256, opened 2019-07-18
KIALI-3139 Use internal url to determine if kiali user has access to jaeger

xeviknal:
** Describe the change **

In order to determine if kiali user has access to jaeger, we need to use internal urls

** Issue reference **
https://issues.jboss.org/browse/KIALI-3139

** Backwards incompatible? **
yes

### DIFF
--- a/business/jaeger_helper.go
+++ b/business/jaeger_helper.go
@@ -40,7 +40,7 @@ func getErrorTracesFromJaeger(namespace string, service string, requestToken str
 			auth.Token = requestToken
 		}
 
-		u, errParse := url.Parse(fmt.Sprintf("http://%s%s/api/traces", appstate.JaegerConfig.Service, appstate.JaegerConfig.Path))
+		u, errParse := GetJaegerInternalURL("/api/traces")
 		if errParse != nil {
 			log.Errorf("Error parse Jaeger URL fetching Error Traces: %s", err)
 			return -1, errParse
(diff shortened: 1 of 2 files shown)


business/jaeger_helper.go
```diff
@@ -77,6 +77,10 @@ func getErrorTracesFromJaeger(namespace string, service string, requestToken str
 	return errorTraces, err
 }
 
+func GetJaegerInternalURL(path string) (*url.URL, error) {
+	return url.Parse(fmt.Sprintf("http://%s%s%s", appstate.JaegerConfig.Service, appstate.JaegerConfig.Path, path))
```
jotak: The URL is built from service without port ... is that supposed to work ?
For me it seems it doesn't. E.g. from kiali pod I have to curl on `http://jaeger-query:16686/...` and not just `http://jaeger-query/...`

aljesusg:
Screenshot from 2019-07-18 15-20-52

xeviknal:
@aljesusg does your jaeger-query service runs in port 80? I think maistra 0.11 changes the jaeger tracing to use port 80. That is why you could fetch jaeger without the port.
In maistra 12, jaeger is under 443 (https). And istio mainstream uses 16646 port.
So I understand that the traces count in the traces tab is only working with jaeger under port 80.
https://github.com/kiali/kiali/blob/v1.1/business/jaeger_helper.go#L43

I propose to autodiscover the port in the autodiscovery process.

I did fill in the JIRA: https://issues.jboss.org/browse/KIALI-3151

aljesusg:
@xeviknal tests that we did in the bluejeans were in  GKE with Istio 1.1.9 , it's not related with Maistra that's the point ^^

hhovsepy:
Verified, when Jeager config refers to a namespace Kiali does not have access.
Screenshot from 2019-07-26 13-37-23

PR #1257, opened 2019-07-18
Trivial: allow to use Jaeger out-cluster

lucasponce:
This is a developer helper.
When working with an external Kiali process, as a developer I'd like to configure the jaeger service to fetch traces errors from service details.

It's transparent to normal behaviour.


hhovsepy:
Verified

Issue #1259, opened 2019-07-18
operator: port name for service is incorrect when https is disabled


mikebryant:
**Describe the bug**

I wanted to use the operator to install kiali, with no backend https, but mtls and a virtualservice.

The resulting service is:
```
  spec:
    clusterIP: 172.20.48.68
    ports:
    - name: tcp
      port: 20001
      protocol: TCP
      targetPort: 20001
    selector:
      app: kiali
      version: v1.1.0
    sessionAffinity: None
    type: ClusterIP
```

This means istio isn't able to route traffic correctly, as the port should be named http not tcp

**Versions used**
Kiali: 1.1.0
Istio: 1.1.9
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9) EKS 1.11

**To Reproduce**
```
---
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  name: kiali
  namespace: kiali-operator
spec:
  auth:
    strategy: anonymous
  deployment:
    accessible_namespaces: "**"
    image_version: v1.1.0
    namespace: kiali
    service_type: ClusterIP
    view_only_mode: true
  external_services:
    grafana:
      url: ""
    jaeger:
      url: ""
  identity:
    cert_file: ""
    private_key_file: ""
  istio_labels:
    app_label_name: app.kubernetes.io/name
    version_label_name: app.kubernetes.io/version
  istio_namespace: istio-system
  server:
    web_root: /
---
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: kiali
  namespace: kiali
spec:
  gateways:
  - gateway.blah...
  hosts:
  - kiali.url...
  http:
  - match:
    - uri:
        prefix: /
    route:
    - destination:
        host: kiali.kiali.svc.cluster.local
        port:
          number: 20001
```

**Expected behavior**
When backend https is disabled I expected the port name to be http so it can be used with istio

israel-hdez:
Closing, because was fixed in master in #1260

PR #1260, opened 2019-07-18
fix: Use correct port name when https is disabled

mikebryant:
** Describe the change **

Use the correct port name when https is disabled

** Issue reference **

https://github.com/kiali/kiali/issues/1259

** Backwards incompatible? **

Maybe, if someone is targeting the tcp port name

** Documentation **

I don't think this needs a readme change

### DIFF
--- a/operator/roles/kiali-deploy/templates/kubernetes/service.yaml
+++ b/operator/roles/kiali-deploy/templates/kubernetes/service.yaml
@@ -9,7 +9,7 @@ metadata:
 spec:
   type: {{ kiali_vars.deployment.service_type }}
   ports:
-  - name: tcp
+  - name: {{ 'http' if kiali_vars.identity.cert_file == "" else 'tcp' }}
     protocol: TCP
     port: {{ kiali_vars.server.port }}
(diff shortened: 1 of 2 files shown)


rhqci:

operator/roles/kiali-deploy/templates/kubernetes/service.yaml
```diff
@@ -9,7 +9,7 @@ metadata:
 spec:
   type: {{ kiali_vars.deployment.service_type }}
   ports:
-  - name: tcp
+  - name: {{ 'http' if kiali_vars.identity.cert_file == "" else 'tcp' }}
     protocol: TCP
```
jmazzitelli: Does the Protocol needs to change to "HTTP" too?

mikebryant: No, that's the layer 4 protocol

From https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#serviceport-v1-core this can only be TCP, UDP, or SCTP

PR #1263, opened 2019-07-19
Attempting to re-enable caching..

burmanm:
** Describe the change **

Trying to re-enable caching. Unfortunately, Openshift API starts to timeout if you try..

** Issue reference **

KIALI-2572
related: https://github.com/kiali/kiali/issues/1386


glide.yaml
```diff
@@ -34,3 +34,5 @@ import:
   version: 91cfa479c814065e420cee7ed227db0f63a5854e
 - package: github.com/kiali/k-charted
   version: 378e9360b5d146e419ce8b6203b35b0fa1fcfb07
+- package: github.com/openshift/client-go
+  version: release-3.11
```
lucasponce: Two comments:
- vender changes should be in separated commits / PRs to not mix them with logic (desirable).
- if this client also compatible with openshift 4 ?

operator/roles/kiali-deploy/vars/main.yml
```diff
@@ -79,3 +79,9 @@ kiali_vars:
     {{ kiali_defaults.server }}
     {%- endif -%}
 
+  cache_config: |
```
jmazzitelli: trivial: All of these lists of settings are alphabetical - move "cache_config" group above "deployment" but below "auth".

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -91,8 +91,9 @@ kiali_defaults:
   kubernetes_config:
     burst: 200
     qps: 175
+
+  cache_config:
```
jmazzitelli: * move the cache_config group up above, below the "auth" group to maintain alphabetical order.
* unless there is a reason not to, define every possible setting in here with their default values. It typically makes the ansible code easier to write and less prone to errors (if, for example, we later need to reference these settings in the ansible scripts).
```
cache_config:
  cache_duration: 30000000
  cache_enabled: false
  istio_objects: false
  kubernetes_objects: false
```

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -309,16 +309,33 @@ spec:
 # The QPS value of the Kubernetes client.
 #    ---
 #    qps: 175
+
+##########
+#  ---
+#  cache_config:
```
jmazzitelli: Maintain alphabetical order - move this group up above, under the "auth" group and before "deployment".

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -309,16 +309,33 @@ spec:
...
 #
-# Flag to use a Kubernetes cache for watching changes and updating pods and controllers data asynchronously.
-# Caution: Kubernetes cache is not compatible with reduced permissions scenearios.
+# Flag to use a cache for watching changes in Istio / Kubernetes. Note, enabling cache only
+# caches projects / namespaces. To cache more objects, enable istio_objects and/or kubernetes_objects
+# to cache more items.
+# Caution: Cache is not compatible with reduced permissions scenarios. Performance issues may arise
+# on the Kubernetes registry if too many namespaces are cached.
 #    ---
 #    cache_enabled: false
```
jmazzitelli: maintain alphabetical ordering. Doc the settings in the order:
* cache_duration
* cache_enabled
* istio_objects
* kubernetes_objects

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -309,16 +309,33 @@ spec:
...
 # The ratio interval (expressed in nanoseconds) used for the cache to perform a full refresh.
 # Only used when cache_enabled is true.
 #    ---
-#    cache_duration: 300000000
+#    cache_duration: 3000000000
+#
+# Cache Kubernetes objects such as Pods, Services, Deployments, etc.
+# Only used when cache_enabled is true.
+#    ---
+#    kubernetes_objects: true
```
jmazzitelli: The commented settings here are also used as documentation - whatever values here are supposed to be the defaults so users know what the default values are if they are not explicitly set. I see in config.go that these settings have a default of "false" so make this "false" as well
```suggestion
#    kubernetes_objects: false
```

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -309,16 +309,33 @@ spec:
...
+#
+# Cache Istio objects such as VirtualServices, Gateways, DestinationRules, etc
+# Only used when cache_enabled is true.
+#    ---
+#    istio_objects: true
```
jmazzitelli: The commented settings here are also used as documentation - whatever values here are supposed to be the defaults so users know what the default values are if they are not explicitly set. I see in config.go that these settings have a default of "false" so make this "false" as well
```suggestion
#    istio_objects: false
```

glide.yaml
burmanm: No idea, it's the same version as we currently use for Openshift object versions. I don't think making them unsynced makes sense.

config/config.go
```diff
@@ -213,10 +213,15 @@ type IstioLabels struct {
...
 type KubernetesConfig struct {
-	Burst         int     `yaml:"burst,omitempty"`
-	QPS           float32 `yaml:"qps,omitempty"`
-	CacheEnabled  bool    `yaml:"cache_enabled,omitempty"`
-	CacheDuration int64   `yaml:"cache_duration,omitempty"`
+	Burst int     `yaml:"burst,omitempty"`
+	QPS   float32 `yaml:"qps,omitempty"`
+}
+
+type CacheConfig struct {
```
lucasponce: I wonder if it makes sense (or not) to also check which namespaces are cached.

config/config.go
```diff
@@ -368,8 +374,12 @@ func NewConfig() (c *Config) {
 	// Kubernetes client Configuration
 	c.KubernetesConfig.Burst = getDefaultInt(EnvKubernetesBurst, 200)
 	c.KubernetesConfig.QPS = getDefaultFloat32(EnvKubernetesQPS, 175)
-	c.KubernetesConfig.CacheEnabled = getDefaultBool(EnvKubernetesCacheEnabled, false)
-	c.KubernetesConfig.CacheDuration = getDefaultInt64(EnvKubernetesCacheDuration, time.Duration(5*time.Minute).Nanoseconds())
+
+	// Cache config
+	c.CacheConfig.CacheEnabled = getDefaultBool(EnvKubernetesCacheEnabled, false)
+	c.CacheConfig.CacheDuration = getDefaultInt64(EnvKubernetesCacheDuration, time.Duration(10*time.Minute).Nanoseconds())
```
lucasponce: Question.
In theory client-go client caching mechanism are dealing with the TTL of the cache.
I mean, at least for the k8s objects is listening the events and it refreshes them under the hood.

In that scenario, why do we need to define a duration ?

glide.yaml
lucasponce: But I wonder if this has not problems with the 
```
 package: k8s.io/client-go
```
or the openshift one only works with the os specific objects ?

(Just asking I didn't test locally the PR yet).

kubernetes/cache.go
```diff
@@ -41,18 +49,38 @@ type (
...
+		GetNamespaces() ([]core_v1.Namespace, error)
+
+		// Openshift caches
+		GetProjects() ([]osproject_v1.Project, error)
+
+		// Istio caches
+		GetVirtualServices(namespace string) (*GenericIstioObjectList, error)
+
+		// Allow subset of cache to be used
+		KubeCached(namespace string) bool
```
lucasponce: Ok, this means that k8s objects for namespace NS are cached, right ?

kubernetes/cache.go
```diff
@@ -41,18 +49,38 @@ type (
...
+		// OpenshiftCached(namespace string) bool
+		IstioCached(namespace string) bool
```
lucasponce: Same here but with the Istio objects.

kubernetes/cache.go
```diff
@@ -41,18 +49,38 @@ type (
...
+		stopChans         map[string]chan struct{}
+		controlStopChan   chan struct{}
+		syncCount         int
+		maxSyncCount      int
+		isErrorState      bool
+		lastError         error
+		lastErrorLock     sync.Mutex
+		controllers       map[string]map[string]cache.SharedIndexInformer
+		istioControllers  map[string]map[string]cache.SharedIndexInformer
+		projectInformer   cache.SharedIndexInformer
```
lucasponce: This can be OpenShift projects or Kubernetes namespaces.
Perhaps a comment to miss the reference in the next methods.

kubernetes/cache.go
```diff
@@ -110,26 +149,135 @@ func initControllers(clientset kube.Interface, refreshDuration time.Duration) ma
...
+		},
+		DeleteFunc: func(obj interface{}) {
+			project := obj.(*osproject_v1.Project)
+			namespace := project.GetObjectMeta().GetName()
+			c.stopNamespace(namespace)
+		},
+	}
+}
+
+func (c *controllerImpl) NamespaceListener() cache.ResourceEventHandlerFuncs {
```
lucasponce: Interesting, I thought that the k8s/client-go should have already a Listener for Namespaces.

kubernetes/cache.go
```diff
@@ -110,26 +149,135 @@ func initControllers(clientset kube.Interface, refreshDuration time.Duration) ma
...
+	if c.istioEnabled {
+		c.initIstioControllersNamespace(namespace)
+		run(c.istioControllers[namespace], stopChan)
+	}
+
+	// TODO Previous istioEnabled && kubernetesEnabled are allowing isEnabled too soon, before the
+	// caches are synced
 
-	return controllers
+	c.WaitForSync()
```
lucasponce: I wonder if we should add a timeout here. The scenario that caches never finished for some reason should not wait forever here, right ?
I mean, if the caches have a problem I guess it can make it crash with some reason sooner than wait in a slow machine.

Just an idea, I know that the previous code was pretty basic and it didn't cover this either.

glide.yaml
burmanm: The OS client-go has only the types

config/config.go
burmanm: We give it to the client-go, which refreshes (resyncs) the cache. The 10 mins is the default setting in the client-go currently.

kubernetes/cache.go
burmanm: Yes.

kubernetes/cache.go
burmanm: Correct.

kubernetes/cache.go
burmanm: Listener yes, but we want to attach the ResourceEventHandlerFunc to it so we can act on the events.

kubernetes/cache.go
burmanm: Yeah, probably makes sense. Although so far it has always failed to something else.  This should probably happen on line 310 I think instead of here though.

kubernetes/cache.go
```diff
@@ -454,3 +670,118 @@ func (c *controllerImpl) GetServices(namespace string) ([]core_v1.Service, error
...
+	if err := c.checkStateAndRetry(); err != nil {
+		return nil, err
+	}
+	if _, found := c.istioControllers[namespace]; !found {
+		log.Debugf("No %s informers for %s found", istioType, namespace) // Debugf
+		return nil, fmt.Errorf("Istio caching is not enabled for namespace %s", namespace)
+	}
+
+	vss := c.istioControllers[namespace][istioType].GetIndexer().List()
+	log.Infof("Len of vss: %d for %s\n", len(vss), namespace)
```
lucasponce: Minor. as this is generic, the vss perhaps can be changed for the type it is fetching.

kubernetes/cache.go
```diff
@@ -454,3 +670,118 @@ func (c *controllerImpl) GetServices(namespace string) ([]core_v1.Service, error
...
+			return itemList, nil
+		}
+		log.Infof("Got %d %s items back from cache", len(vss), istioType) // Debugf
+	}
+
+	return nil, nil
+}
+
+// Allow subset of cache to be used - could be enabled all the way to namespace level
+// TODO There's no config option for namespace level configuration yet
```
lucasponce: yeah, I guess I put a comment as well for this TODO in other area of the code, too.

kubernetes/cache.go
```diff
@@ -454,3 +670,118 @@ func (c *controllerImpl) GetServices(namespace string) ([]core_v1.Service, error
...
+func (c *controllerImpl) KubeCached(namespace string) bool {
```
lucasponce: Yeah, I guess if here also we may want to discriminate per type too, or not.
KubeCached("bookinfo", "VirtualServices")

I'm more sure that we can save some performance if we skip rare objects (i.e thinking in not used controllers for workloads and also not all Istio objects deserves to be cached as well).

burmanm: I don't know if it makes sense to have that distinction here. Unless we then allow dynamic configuration of types to be cached. If instead we just want to skip "rare objects", it's more simple to just not make any ``GetRareObject()`` methods in the cache.go

burmanm: How that affects things like fetching ``IstioConfigList`` is something we need to consider also. It still takes time to fetch those from the k8s API, even if they're not used for other things.

kubernetes/client.go
```diff
@@ -304,12 +305,39 @@ func NewClientFromConfig(config *rest.Config) (*IstioClient, error) {
...
+
 	client.istioConfigApi = istioConfigAPI
 	client.istioNetworkingApi = istioNetworkingAPI
 	client.istioAuthenticationApi = istioAuthenticationAPI
 	client.istioRbacApi = istioRbacApi
 	client.maistraAuthenticationApi = maistraAuthenticationAPI
 	client.maistraRbacApi = maistraRbacApi
+
+	// Init client cache
+	// Note that cache will work only in full permissions scenarios (similar permissions as mixer/istio-telemetry component)
```
lucasponce: yeah, no full permissions scenarios need to be revisited how to use them with the cache.
Probably namespaces/project can't be cached, but perhaps workloads and istio objects can get benefits.

kubernetes/istio_details_service.go
```diff
@@ -196,14 +196,27 @@ func (in *IstioClient) getRbacResources() map[string]bool {
 // If serviceName param is provided it will filter all VirtualServices having a host defined on a particular service.
 // It returns an error on any problem.
 func (in *IstioClient) GetVirtualServices(namespace string, serviceName string) ([]IstioObject, error) {
-	result, err := in.istioNetworkingApi.Get().Namespace(namespace).Resource(virtualServices).Do().Get()
-	if err != nil {
-		return nil, err
+	var virtualServiceList *GenericIstioObjectList
+	if in.k8sCache != nil && in.k8sCache.IstioCached(namespace) {
```
lucasponce: yeah, here I guess it can also check the (namespace, type), and probably with that tradeoff between caching / performance may work.

kubernetes/kubernetes_service.go
```diff
@@ -22,6 +22,21 @@ import (
 // GetNamespace fetches and returns the specified namespace definition
 // from the cluster
 func (in *IstioClient) GetNamespace(namespace string) (*core_v1.Namespace, error) {
+	if in.k8sCache != nil {
+		nss, err := in.GetNamespaces("")
```
lucasponce: yeah, this operation is clusterwide in k8s, in openshift projects make the filter per user "visibiliy".

kubernetes/kubernetes_service.go
```diff
@@ -60,8 +96,24 @@ func (in *IstioClient) GetNamespaces(labelSelector string) ([]core_v1.Namespace,
 // the specified name by querying the cluster API. GetProject will fail
 // if the underlying cluster is not Openshift.
 func (in *IstioClient) GetProject(name string) (*osproject_v1.Project, error) {
-	result := &osproject_v1.Project{}
 
+	if in.k8sCache != nil {
+		projects, err := in.GetProjects("")
```
lucasponce: I remember that k8s caches by default had a default index on namespace and name IIRC.
Perhaps this is because we need to create the indexer for projects, right ?

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -127,6 +127,33 @@ spec:
...
```
lucasponce: as discussed, perhaps we may want something more flexible as namespace x object.
By default, it can go with all/nothing, but it would be interested to open potential tweaks just to give options for tuning.

operator/deploy/role.yaml
```diff
@@ -72,6 +73,8 @@ rules:
   resources:
   ${OPERATOR_ROLE_CLUSTERROLEBINDINGS}
   ${OPERATOR_ROLE_CLUSTERROLES}
+  - clusterroles
```
lucasponce: This is what we need to review, that cache should work well in clusterroles or no elevated privileges scenarios.
So, that may affect some queries (namespaces/projects), but other objects can get benefit of them.

lucasponce:
Nice work @burmanm.
I did a brief review of code but my next step is to fetch locally and experiment with some of the ideas I leave in the comments.

lucasponce:
@burmanm thanks a lot for your work on this PR.
We have consolidated and merged these ideas on #1749 and others, so, now master has a caching system that we can improve in following iterations.

I'm waiting feedback from QE team about it.

I think this one can be closed, too.

Issue #1237, opened 2019-07-12
Feature : Show API documentation in Kiali service tab


couniojc:
**Describe the bug**
We are proposing to add service API info in Kiali. We could display the type of API of the service (rest, graphql, grpc) in the services list, and add a new 'API Doc' tab in the service info, that would display api documentation generated from the spec served by the service (for example  openapi sepc for rest services).
All information is provided by annotating the services, the extra info being displayed only for annotated services.
We've done some screenshots of the work we've done so far :

<img width="1379" alt="kiali-apidocs-list" src="https://user-images.githubusercontent.com/1235410/61105953-62885c80-a430-11e9-822c-6abbcae23e57.png">

<img width="1375" alt="kiali-apidoc" src="https://user-images.githubusercontent.com/1235410/61105965-6caa5b00-a430-11e9-9075-57f1b28748ce.png">

**Versions used**
Kiali: 1.2.0-snapshot.0
Istio: 1.2.2

pilhuhn:
That is as discussed here: https://groups.google.com/forum/#!topic/kiali-dev/99E9VDoj0fU

lucasponce:
I think this one was implemented on #1325 so we are going to close this issue.
There is a followup on https://github.com/kiali/kiali/issues/1567

PR #1266, opened 2019-07-19
Add backend for ApiDocumentation feature

couniojc:
** Describe the change **

This adds backend to support API Documentation feature. We are collecting the api type and api spec fields from annotations if they are present and add it to the JSON that is sent to the FrontEnd. This also add a new endpoint to proxy the api spec

** Issue reference **

https://github.com/kiali/kiali/issues/1237

** Backwards incompatible? **

No

** Documentation **

https://github.com/kiali/kiali/pull/1266/commits/94a0441f7976b7039f1d45dd0f941bba3a173876

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -333,12 +333,12 @@ spec:
 
 == Services API Documentation
 
-Kiali can display API Documentation of your services. See https://user-images.githubusercontent.com/1235410/61569413-0e870480-aa3c-11e9-9527-bbea18eff475.png[Api documentation screen capture] and https://user-images.githubusercontent.com/1235410/61569419-1777d600-aa3c-11e9-82cc-f9510e5b9808.png[Api type list screen capture].
+Kiali can display API Documentation of your services. See https://user-images.githubusercontent.com/1235410/61569413-0e870480-aa3c-11e9-9527-bbea18eff475.png[API documentation screen capture] and https://user-images.githubusercontent.com/1235410/61569419-1777d600-aa3c-11e9-82cc-f9510e5b9808.png[API type list screen capture].
 
 === Configure your services
 
-Your services must be annotated with the type of api ('rest', 'grpc', 'graphql') and a url to the spec of the api. 
-If the api spec is served from the service itself, Kiali will infer the hostname and port :
+Your services must be annotated with the type of API ('rest', 'grpc', 'graphql') and a URL to the spec of the API. 
+If the API spec is served from the service itself, Kiali will infer the hostname and port :
 
 [source,yaml]
 ----
@@ -353,7 +353,7 @@ spec:
 ...
 ----
 
-The api spec can also be served from any http/s url, internal or external to the cluster :
+The API spec can also be served from any http/s URL, internal or external to the cluster :
 
 [source,yaml]
 ----
@@ -368,12 +368,12 @@ spec:
 ...
 ----
 
-For now, only REST apis have their spec displayed but we are working to support gRpc and GraphQL soon.
+For now, only REST APIs have their spec displayed but we are working to support gRpc and GraphQL soon.
 A live console to test your APIs directly with Kiali is also being worked on.
 
-=== Customize api docs annotations
+=== Customize API docs annotations
 
-You can configure Kiali to use your own annotation names with the operator cr
+You can configure Kiali to use your own annotation names with the Kiali CR
 
 [source,yaml]
 ----
--- a/README.adoc
+++ b/README.adoc
@@ -331,6 +331,60 @@ spec:
 ...
 ----
 
+== Services API Documentation
+
+Kiali can display API Documentation of your services. See https://user-images.githubusercontent.com/1235410/61569413-0e870480-aa3c-11e9-9527-bbea18eff475.png[Api documentation screen capture] and https://user-images.githubusercontent.com/1235410/61569419-1777d600-aa3c-11e9-82cc-f9510e5b9808.png[Api type list screen capture].
+
+=== Configure your services
+
+Your services must be annotated with the type of api ('rest', 'grpc', 'graphql') and a url to the spec of the api. 
+If the api spec is served from the service itself, Kiali will infer the hostname and port :
+
+[source,yaml]
+----
+apiVersion: v1
+kind: Service
+metadata:
+  name: myservice
+  annotations:
+    kiali.kubernetes.io/api-type: rest
+    kiali.kubernetes.io/api-spec: /v1/api-spec  
+spec:
+...
+----
+
+The api spec can also be served from any http/s url, internal or external to the cluster :
+
+[source,yaml]
+----
+apiVersion: v1
+kind: Service
+metadata:
+  name: petstore
+  annotations:
+    kiali.kubernetes.io/api-type: rest
+    kiali.kubernetes.io/api-spec: https://petstore.swagger.io/v2/swagger.json  
+spec:
+...
+----
+
+For now, only REST apis have their spec displayed but we are working to support gRpc and GraphQL soon.
+A live console to test your APIs directly with Kiali is also being worked on.
+
+=== Customize api docs annotations
+
+You can configure Kiali to use your own annotation names with the operator cr
+
+[source,yaml]
+----
+...
+apidocs:
+  annotations:
+    api_spec_annotation_name: "my-annotation-for-api-spec"
+    api_type_annotation_name: "my-annotation-for-api-type"
+...
+---
+
 == Additional Notes
 
 === Customize the UI web context root
--- a/business/services.go
+++ b/business/services.go
@@ -11,7 +11,6 @@ import (
 	apps_v1 "k8s.io/api/apps/v1"
 	core_v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
@@ -99,7 +98,7 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 		/** Check if Service has the label app required by Istio */
 		_, appLabel := item.Spec.Selector[conf.IstioLabels.AppLabelName]
 		/** Check if Service has the api annotation */
-		apiTypeFromAnnotation, _ := item.ObjectMeta.Annotations[conf.ApiDocumentation.ApiTypeAnnotationName]
+		apiTypeFromAnnotation := item.ObjectMeta.Annotations[conf.ApiDocumentation.Annotations.ApiTypeAnnotationName]
 		services[i] = models.ServiceOverview{
 			Name:         item.Name,
 			IstioSidecar: hasSidecar,
@@ -123,7 +122,7 @@ func (in *SvcService) GetServiceApiDocumentation(namespace, service string) (str
 		log.Errorf("Error fetching Service per namespace %s and service %s: %s", namespace, service, err)
 		return "", errors.NewInternalError(err)
 	}
-	apiSpecPath, _ := svc.ObjectMeta.Annotations[conf.ApiDocumentation.ApiSpecAnnotationName]
+	apiSpecPath := svc.ObjectMeta.Annotations[conf.ApiDocumentation.Annotations.ApiSpecAnnotationName]
 
 	if apiSpecPath == "" {
 		qualifiedResource := schema.GroupResource{
@@ -149,17 +148,16 @@ func (in *SvcService) GetServiceApiDocumentation(namespace, service string) (str
 	}
 	defer resp.Body.Close()
 
+	data, err3 := ioutil.ReadAll(resp.Body)
 	if resp.StatusCode != http.StatusOK {
 		log.Errorf("Status error: %v", resp.StatusCode)
-		return "", &errors.StatusError{meta_v1.Status{
-			Status:  meta_v1.StatusFailure,
-			Code:    int32(resp.StatusCode),
-			Reason:  "",
-			Message: "",
-		}}
+		qualifiedResource := schema.GroupResource{
+			Group:    "",
+			Resource: "",
+		}
+		return "", errors.NewGenericServerResponse(resp.StatusCode, "GET", qualifiedResource, "Proxied request error", string(data), -1, true)
 	}
 
-	data, err3 := ioutil.ReadAll(resp.Body)
 	if err3 != nil {
 		log.Errorf("Read body: %v", err)
 		return "", errors.NewInternalError(err3)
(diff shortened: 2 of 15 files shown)


rhqci:

pilhuhn:
Build failure due to
```
These files need to be formatted: ./handlers/apidocumentation.go
./config/config.go
./models/service.go
./business/services.go
./routing/routes.go
```

config.yaml
```diff
@@ -39,3 +39,12 @@ external_services:
   # Uncomment service_label_name to set up a different label name for grouping all resources of an app version.
   # Default is "version"
   # version_label_name: version
+
```
jmazzitelli: Don't bother updating this file. Its obsolete and out of date.

config/config.go
```diff
@@ -241,6 +250,7 @@ type Config struct {
 	API              ApiConfig         `yaml:"api,omitempty"`
 	Auth             AuthConfig        `yaml:"auth,omitempty"`
 	Deployment       DeploymentConfig  `yaml:"deployment,omitempty"`
+        ApiDocumentation ApiDocAnnotations `yaml:"apidoc_annotations,omitempty"`
```
jmazzitelli: This is going to require some additions to the operator's `operator/roles/kiali_deploy/vars/main.yml` and `operator/roles/kiali_deploy/defaults/main.yml` as well as documented in the `operator/deploy/kiali_cr.yaml` file. I will add a more detailed comment in this PR once I finish this review.

jmazzitelli: Add the documentation to this file following the same syntax/format as the rest:

https://github.com/kiali/kiali/blob/v1.1.0/operator/deploy/kiali/kiali_cr.yaml#L64

This is an example Kiali CR that has all the defaults commented out. It is formatted in way that lets a user easily create a Kiali CR (e.g. all the indentation is correct). All of these settings are alphabetically sorted - so put this new `apidoc_annotations" section after the "api" but before the "auth" section.

Make sure you set the default values in the comments (all values you see in here are the defaults - its how people will know what the defaults are; by just using this file as a reference).

jmazzitelli: Because you are adding a new top-level group of settings, you need to add a section to this file, following the same syntax and format as the other groups - again, these are alphabetically sorted so put the new apidoc_annotations group after the "api" and before the "auth" groups:

https://github.com/kiali/kiali/blob/v1.1.0/operator/roles/kiali-deploy/vars/main.yml#L25

jmazzitelli: Finally, you want to define the full config section (with all available settings and their default values) in here:

https://github.com/kiali/kiali/blob/v1.1.0/operator/roles/kiali-deploy/defaults/main.yml#L23

Again, alphabetically sort everything. So if I understand this PR correctly, it will look something like:

```
apidoc_annotations:
  api_spec_annotation_name: "kiali.kubernetes.io/api-spec"
  api_type_annotation_name: "kiali.kubernetes.io/api-type"

auth:
  strategy:
... the rest of the file ...
```

jmazzitelli: As a suggestion, you should think about renaming this "apidoc_annotations" section. If additional configuration settings are needed in the future, it would be best to name this section "apidocs" - that way any new settings related to the apidocs feature can be added here even if they are not related to the annotations themselves. So something like this should be considered:

```
apidocs:
  api_spec_annotation_name: "kiali.kubernetes.io/api-spec"
  api_type_annotation_name: "kiali.kubernetes.io/api-type"
```

Or something along those lines.

couniojc: Ok, I will have a look to the operator, thanks !

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -21,6 +21,11 @@ kiali_defaults:
       - "kiali-operator"
       #label_selector:
 
+  apidocs:
+    annotations:
+      api_type_annotation_name: "kiali.kubernetes.io/api-type"
```
jmazzitelli: trivial change request - all peer settings should be alphabetically sorted (as they already are in your comments up above on lines 75-76), so just swap the type and spec setting here to maintain alphabetical ordering
```
  apidocs:
    annotations:
      api_spec_annotation_name: "kiali.kubernetes.io/api-spec"
      api_type_annotation_name: "kiali.kubernetes.io/api-type"
```

couniojc:
@jmazzitelli Can you have a quick look to latest changes ? Thanks !

mtho11:
@couniojc @jmazzitelli is out on vacation. @jshaughn or @mwringe perhaps can take a peek?

README.adoc
```diff
@@ -331,6 +331,60 @@ spec:
...
+spec:
+...
+----
+
+For now, only REST apis have their spec displayed but we are working to support gRpc and GraphQL soon.
+A live console to test your APIs directly with Kiali is also being worked on.
+
+=== Customize api docs annotations
+
+You can configure Kiali to use your own annotation names with the operator cr
```
jshaughn: operator cr => Kiali Operator CR

README.adoc
```diff
@@ -331,6 +331,60 @@ spec:
 ...
 ----
 
+== Services API Documentation
+
+Kiali can display API Documentation of your services. See https://user-images.githubusercontent.com/1235410/61569413-0e870480-aa3c-11e9-9527-bbea18eff475.png[Api documentation screen capture] and https://user-images.githubusercontent.com/1235410/61569419-1777d600-aa3c-11e9-82cc-f9510e5b9808.png[Api type list screen capture].
+
+=== Configure your services
+
+Your services must be annotated with the type of api ('rest', 'grpc', 'graphql') and a url to the spec of the api.
```
jshaughn: minor, inconsistent capitalization. Should probably use "API" consistently. Also, URL.

config.yaml
jshaughn: The conversation is resolved but the file is still updated.

README.adoc
jmazzitelli: We call this the `Kiali CR` in all other places where we reference it. So I would continue to use that name -> Kiali CR

README.adoc
couniojc: Fixed

README.adoc
couniojc: Fixed

couniojc:
@jmazzitelli Test steps :
- Build Kiali along with the UI (https://github.com/kiali/kiali-ui/pull/1331)
- Start Kiali
- Click on the 'Services' menu tab
- Select 'istio-system' namespace
- You should see the 'api: ' column filled for kiali service with open-api logo
- Click on the kiali service line
- You should see an 'API Doc' tab in the service page
- Click on the 'API Doc' tab and you should see Kiali Rest API doc displayed
- Go back to the service list page
- Click on jaeger-query service
- You should not see an 'API Doc' tab in the service page
- Go back to the service list page
- In your k8s cluster, add annotations to 'istio-pilot' service (or another service)
- kubectl annotate service istio-pilot -n istio-system kiali.kubernetes.io/api-type=grpc
- Refresh service list page and you should see the grpc api  logo along with istio-pilot line
- In your k8s cluster, add annotations to 'istio-policy' service (or another service)
- kubectl annotate service istio-policy -n istio-system kiali.kubernetes.io/api-type=rest
- kubectl annotate service istio-policy -n istio-system kiali.kubernetes.io/api-spec=https://petstore.swagger.io/v2/swagger.json
- Refresh service list page,  and you should see the open-api logo along with istio-policy line
- Click on the istio-policy line
- Click on the 'API Doc' tab and you should see the PetStore API doc

business/services.go
```diff
@@ -118,6 +182,20 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime
...
 	var ws models.Workloads
 	var nsmtls models.MTLSStatus
+	var apidoc models.ApiDocumentation
+
+	conf := config.Get()
+	apiSpecFromAnnotation := svc.ObjectMeta.Annotations[conf.ApiDocumentation.Annotations.ApiSpecAnnotationName]
+	apiTypeFromAnnotation := svc.ObjectMeta.Annotations[conf.ApiDocumentation.Annotations.ApiTypeAnnotationName]
+	apiBaseUrl := ""
+	if apiSpecFromAnnotation != "" {
+		apiBaseUrl = conf.Server.WebRoot + "/api/namespaces/" + namespace + "/services/" + service
```
jotak: It doesn't work for me: my WebRoot is `/` so it's ending up in something like: `//api/namespaces/istio-system/services/istio-pilot/apispec` , which fails & I can't see the petstore doc that i've configured in annotations.

Actually I think this `apiBaseUrl` is not needed at all because we can build it from the frontend, just like we build any other API url. I'll propose a patch to your frontend PR about that.

jotak: @couniojc can you rebase your 2 PRs from master? You'd have to do that anyway, and it's also needed to fix an issue on the frontend to fix the build.

jotak:
I've tested and it's almost good (see inline comments), very nice work!

Then I'll also have a couple of minor comments but we can do them in another PR, to not block this one for too long. I'm listing them right here (not inline) to retrieve more easily later (both frontend and backend):

- In go/GetServiceApiDocumentation we should set a timeout on remote calls.
- in UI, loading indicator is still displayed upon failure; also it's not in masthead as we do usually
- in JS console I'm seeing several 404 not found about fonts ... I guess we should either not try to load them, or bundle them in Kiali
- In Services list maybe the API icons should be a bit bigger?
- Add API logo on ApiDocumentation page too
- Looks like rapidoc is logging some debug info to the console ... we probably want to avoid that.

Again, I think we can go without them now and have a follow up - the only blocking issue I see is about the webroot URL. Feature looks great

business/services.go
couniojc: @jotak Thanks, ok I will rebase, and yes I had back and forth about this baseUrl, but you're right, it will be cleaner if it is built in the frontend, let me see if I can change it

couniojc:
@jotak : I moved the url spec build to the FE, you can see the commit on the FE PR as well

- About logo : Yes grpc is small because it is not square like the others, I'll try to find a more square icon, or have something on FE side to make it bigger for this one
- About fonts & rapidoc :  I probably will try to convert the web-component to pure React to reuse stuff for other renderings, we can move those to assets at that time
- Add API logo on doc page : Yes, definitely a good idea

jotak:
@couniojc about the fonts, we will have to check with @kiali/ux-reviewers because Kiali follows conventions and style defined by Patternfly ( https://www.patternfly.org/ ) , so maybe we will want to use our usual fonts instead.

jotak:
Actually, I've just verified the special case of kiali serving its own API doc and it doesn't work for me. I've identified two reasons:
- Because as mentioned previously my web_root is "/" so the apispec annotation becomes `//swagger.json` ; the double-slash is an issue
- Because in the code it's assumed kiali runs on http, but it's https for me, so it wouldn't work

BTW I wonder if instead of this loopback connection we could read the file directly, for the special case of Kiali apidoc?

Anyway, I think Kiali self-serving its apidoc is not absolutely required for a first iteration. So I keep my approval and we can fix that later.

jmazzitelli:
> * Because in the code it's assumed kiali runs on http, but it's https for me, so it wouldn't work

Isn't this going to be a problem for any apidoc behind https?

jotak:
> Isn't this going to be a problem for any apidoc behind https?

Scheme "http://" is added only when it's not explicitly provided in the annotation, so for https the user would have to write it explicitly and it should work.

jmazzitelli:
>Because as mentioned previously my web_root is "/" so the apispec annotation becomes //swagger.json ; the double-slash is an issue

> Because in the code it's assumed kiali runs on http, but it's https for me, so it wouldn't work

> Scheme "http://" is added only when it's not explicitly provided in the annotation, so for https the user would have to write it explicitly and it should work.

We can address these two issues. We know if Kiali is behind https or https using the ansible variables. Also, the double-slash should be fixable very easily too.

How about something like (this is in the openshift and kubernetes service templates for the value of the annotation:

```
  annotations:
    {{ kiali_vars.apidocs.annotations.api_spec_annotation_name }}: {{ 'http' if kiali_vars.identity.cert_file == "" else 'https' }}://WHAT_HOST_GOES_HERE?/{{ '' if kiali_vars.server.web_root == '/' else kiali_vars.server.web_root}}/swagger.json
    {{ kiali_vars.apidocs.annotations.api_type_annotation_name }}: rest
```

The only question I have is the host. Is this annotation a full URL when you need to specify protocol? Can we have this api doc stuff be told what protocol to use without having to specify the host? (presumably the host is already known).

operator/roles/kiali-deploy/templates/kubernetes/service.yaml
```diff
@@ -6,6 +6,9 @@ metadata:
   labels:
     app: kiali
     version: {{ kiali_vars.deployment.version_label }}
+  annotations:
+    {{ kiali_vars.apidocs.annotations.api_spec_annotation_name }}: {{ kiali_vars.server.web_root }}/swagger.json
```
jmazzitelli: If the web_root is just `/` then this causes a problem. But this is easily fixable by using an if-else expression:

```suggestion
    {{ kiali_vars.apidocs.annotations.api_spec_annotation_name }}: {{ '' if kiali_vars.server.web_root == '/' else kiali_vars.server.web_root}}/swagger.json
```

operator/roles/kiali-deploy/templates/openshift/service.yaml
```diff
@@ -8,6 +8,8 @@ metadata:
     version: {{ kiali_vars.deployment.version_label }}
   annotations:
     service.alpha.openshift.io/serving-cert-secret-name: kiali-cert-secret
+    {{ kiali_vars.apidocs.annotations.api_spec_annotation_name }}: {{ kiali_vars.server.web_root }}/swagger.json
```
jmazzitelli: If the web_root is just `/` then this causes a problem. But this is easily fixable by using an if-else expression:

```suggestion
    {{ kiali_vars.apidocs.annotations.api_spec_annotation_name }}: {{ '' if kiali_vars.server.web_root == '/' else kiali_vars.server.web_root}}/swagger.json
```

operator/roles/kiali-deploy/templates/kubernetes/service.yaml
couniojc: Thank you !

jotak:
I'll test again today

jotak:
@jmazzitelli , I'm not sure to understand your patch, you've fixed the "/" webroot issue but not the http vs https scheme, right? So we still need to apply something as you suggested : `{{ 'http' if kiali_vars.identity.cert_file == "" else 'https' }}` ?

jmazzitelli:
> I'm not sure to understand your patch, you've fixed the "/" webroot issue but not the http vs https scheme, right?

That is correct.

> So we still need to apply something as you suggested : `{{ 'http' if kiali_vars.identity.cert_file == "" else 'https' }}` ?

Right - and the problem is I don't know how to complete the URL because you'd need the host. That's why I did not suggest a recommended fix for that. See my [earlier comment](https://github.com/kiali/kiali/pull/1266#issuecomment-517770149) when I say

` {{ kiali_vars.apidocs.annotations.api_spec_annotation_name }}: {{ 'http' if kiali_vars.identity.cert_file == "" else 'https' }}://WHAT_HOST_GOES_HERE?/{{ '' if kiali_vars.server.web_root == '/' else kiali_vars.server.web_root}}/swagger.json`

 `The only question I have is the host. Is this annotation a full URL when you need to specify protocol? Can we have this api doc stuff be told what protocol to use without having to specify the host? (presumably the host is already known).`

jotak:
Hmm ... looks like "localhost" should work, no? As I said above ( https://github.com/kiali/kiali/pull/1266#issuecomment-517720663 ) we could as well read the swagger file directly from fs instead of loopback connection

jotak:
superseded by https://github.com/kiali/kiali/pull/1325

PR #1273, opened 2019-07-23
KIALI-3153 Add response_flags to metrics

jotak:
JIRA: https://issues.jboss.org/browse/KIALI-3153

As mentioned in the JIRA, adding this new label is an easy one however the codes can be kind of cryptic (FI, UF, URX, etc.), so I suggest to prioritize task https://issues.jboss.org/browse/KIALI-1690 for an in-app help on the metrics page, which would add info about all codes.

codes

### DIFF
--- a/business/dashboards_test.go
+++ b/business/dashboards_test.go
@@ -32,7 +32,7 @@ func TestGetIstioDashboard(t *testing.T) {
 
 	assert.Nil(err)
 	assert.Equal("Inbound Metrics", dashboard.Title)
-	assert.Len(dashboard.Aggregations, 4)
+	assert.Len(dashboard.Aggregations, 5)
 	assert.Equal("Local version", dashboard.Aggregations[0].DisplayName)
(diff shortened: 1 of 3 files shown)


abonas:
this is a great enhancement. what is a response code "-" ? is it for "none" or something else?

jotak:
Yes, it stands for "none" , this is how it appears in prometheus (label raw values are kept unchanged as they can be pretty much anything). Not sure how to help the user to understand these codes. What we are showing here is the raw prometheus labels, in KIALI-1690 I'm suggesting to create a help dialog in this page.

hhovsepy:
Verified:
Screenshot from 2019-07-24 17-45-10

jotak:
@israel-hdez I agree, that would be better than a separate help dialog. I'm thinking about that (will create JIRA) but I'm merging this one as a first iteration.

jshaughn:
Just saw this, great addition @jotak .

PR #1275, opened 2019-07-24
KIALI-3151 Add namespace and port for Jaeger integration test

xeviknal:
** Describe the change **

Don't assume that jaeger is in port 80 when testing if jaeger integration is enabled.
Also fixed the auto-discovery of the path.

** Issue reference **
https://issues.jboss.org/browse/KIALI-3151

** Backwards incompatible? **
yes

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -57,7 +57,7 @@ const (
 	EnvTracingEnabled          = "TRACING_ENABLED"
 	EnvTracingURL              = "TRACING_URL"
 	EnvTracingServiceNamespace = "TRACING_SERVICE_NAMESPACE"
-	EnvTracingServicePort 	   = "TRACING_SERVICE_PORT"
+	EnvTracingServicePort      = "TRACING_SERVICE_PORT"
 
 	EnvThreeScaleAdapterName = "THREESCALE_ADAPTER_NAME"
 	EnvThreeScaleServiceName = "THREESCALE_SERVICE_NAME"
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -157,6 +158,8 @@ type TracingConfig struct {
 	Enabled   bool   `yaml:"enabled"`
 	Namespace string `yaml:"namespace"`
 	Service   string `yaml:"service"`
+	Port      int32  `yaml:"port"`
+	Protocol  string `yaml:"protocol"`
```
jotak: Hmm Protocol is not used or are you going to add it later?

config/config.go
```diff
@@ -157,6 +158,8 @@ type TracingConfig struct {
 	Enabled   bool   `yaml:"enabled"`
 	Namespace string `yaml:"namespace"`
 	Service   string `yaml:"service"`
+	Port      int32  `yaml:"port"`
```
jotak: That must be added in the CR files too (as doc in `kiali_cr.yaml`, and `main.yml`)

xeviknal: Good point

config/config.go
xeviknal: Removed. Not adding it now.

hhovsepy:
Verified

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -272,6 +273,7 @@ spec:
 #      enabled: true
 #      namespace: ""
 #      service : ""
+#      port: 16686
```
jmazzitelli: Oh man!! you guys screwed up my alphabetically sorting in here. I knew this would happen when I went on vacation :)

jotak: Yeah, we broke as much as we could during your vacation!

PR #1280, opened 2019-07-25
KIALI-3165 - Adding Molecule Test for whole Multi-tenancy stack (all operators)

gbaufake:
## Describe the change 

Adding a test which will install whole operators needed (kiali, maistra, jaeger), two control planes with Multi-tenancy and CNI enabled. This test will test 
- Kiali configmaps, 
- Kiali pods 
- Kiali crs
- Maistra service mesh memberrolls
- Maistra service mesh controlplanes

## Issue reference 
[KIALI-3165](https://issues.jboss.org/browse/KIALI-3165)


operator/.gitignore
```diff
@@ -0,0 +1 @@
+molecule/maistra-e2e/assets/secret.yaml
```
jmazzitelli: Can you explain what this secret.yaml is? It is being ignored, but is it required to run this test?

gbaufake: It is only when you are going to deploy from maistra on quay.
It is possible to deploy from docker.io/maistra ubi images so the secret is not needed.

PR #1283, opened 2019-07-25
KIALI-3082 Enable check to enforce QE approvals

israel-hdez:
https://issues.jboss.org/browse/KIALI-3082

This will enable the kiali-bot to require a review from, at least, one of the QE people.

The PR will still be mergeable. In order to get real mandatory checks, the repository will need to be configured turn the checks from the kiali-bot as required (this is good until we know that the bot works OK).

### DIFF
--- a/.github/kiali.yml
+++ b/.github/kiali.yml
@@ -1,0 +1,4 @@
+checks:
+  pull_requests:
+    required_approvals:
+      - [gbaufake, mattmahoneyrh, hhovsepy, skondkar]


jotak:
The JIRA state "This could be requiring that one of the reviews is from the QE team, or require certain labels to be added if it doesn't require a QE review. "

How can we implement the last part? (having labels to bypass the rule) ?

israel-hdez:
@jotak 

> How can we implement the last part? (having labels to bypass the rule) ?

To be honest, I omitted that part for now. My reasoning is that, if we provide that option, then the Jira loses its value, since anybody can set labels. IMO, QE should always approve at least to acknowledge that the PR doesn’t need a review from their part.

Since this won’t block merging, I’d like to have that feedback. But if you think it’ll be problematic, I can implement that part.

PR #1284, opened 2019-07-26
Gateway validations: tighten up duplicate host matching regex.

### LINKED ISSUE #1272: Incorrect warning about duplicate host/port config on Gateways
**Description**

I have one `Gateway`, configured as follows:

```yaml
kind: Gateway
metadata:
  name: xyz-admin-gateway
  namespace: default
  selfLink: >-
    /apis/networking.istio.io/v1alpha3/namespaces/default/gateways/xyz-admin-gateway
spec:
  servers:
    - hosts:
        - web-admin-bff.xyz.mydev.io
        - xyz.mydev.io
      port:
        name: https-443-web-admin-ui-platform-dev
        number: 443
        protocol: https
      tls:
        credentialName: cert-star.xyz.mydev.io
        mode: SIMPLE
        privateKey: sds
        serverCertificate: sds
  selector:
    istio: ingressgateway
```

Kiali still reports on this one `Gateway` that there is "More than one gateway for the same host port combination".

I had a quick look at the code, and i think the error might be in this line:

https://github.com/kiali/kiali/blob/2c10cc9bae44037f21d811c141089fb6e9cf02ee/business/checkers/gateways/multi_match_checker.go#L128

If i understand correctly, we're regex matching host A against host B, both directions.  When one of the hostnames is a substring of the other one, it'll give a false positive.

A possible fix might be to wrap the match string in `^....$` to force matching the entire string, but that feels hacky.

Or, maybe i'm misunderstanding, and my configuration as pasted above is actually problematic?  It seems to work fine for me.

**Versions used**
Kiali: `quay.io/kiali/kiali:v0.21`
Istio: v1.2.2
Kubernetes flavour and version: Amazon EKS 1.13, nodes running `v1.13.7-eks-c57ff8`

**To Reproduce**
1. Deploy a `Gateway` object as above
1. Observe Kiali giving a warning on http://localhost:20001/kiali/console/namespaces/default/istio/gateways/xyz-admin-gateway

**Expected behavior**
There should be no warning about a duplicate host/port combination.

toothbrush:
This is to address an issue where a perfectly legitimate `Gateway`
with hostnames `example.com` and `foo.example.com` would get flagged
with a duplicate-host warning.  This was as a result of an unanchored
regular expression.  This PR introduces a test case to illustrate it,
as well as a fix.

Closes #1272.

* business/checkers/gateways/multi_match_checker.go (findMatch):
  Anchor the hostname with `^$` when it is used as a regular
  expression.  This avoids incorrect substring matching behaviour.
* business/checkers/gateways/multi_match_checker_test.go (TestNoMatchOnSubdomainHost):
  Introduce a regression test of the above.

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -125,7 +125,15 @@ func (m MultiMatchChecker) findMatch(host Host) (bool, []Host) {
 			current := strings.Replace(host.Hostname, "*", ".*", -1)
 			previous := strings.Replace(h.Hostname, "*", ".*", -1)
 
-			if regexp.MustCompile(current).MatchString(previous) || regexp.MustCompile(previous).MatchString(current) {
+			// We anchor the beginning and end of the string when it's
+			// to be used as a regex, so that we don't get spurious
+			// substring matches, e.g., "example.com" matching
+			// "foo.example.com".
+			current_re := strings.Join([]string{"^", current, "$"}, "")
+			previous_re := strings.Join([]string{"^", previous, "$"}, "")
+
+			if regexp.MustCompile(current_re).MatchString(previous) ||
+				regexp.MustCompile(previous_re).MatchString(current) {
 				duplicates = append(duplicates, h)
 				break
 			}
--- a/business/checkers/gateways/multi_match_checker_test.go
+++ b/business/checkers/gateways/multi_match_checker_test.go
@@ -99,6 +99,31 @@ func TestWildCardMatchingHost(t *testing.T) {
 
 }
 
+func TestNoMatchOnSubdomainHost(t *testing.T) {
+	conf := config.NewConfig()
+	config.Set(conf)
+
+	assert := assert.New(t)
+
+	gwObject := data.AddServerToGateway(data.CreateServer(
+		[]string{
+			"example.com",
+			"thisisfine.example.com",
+		}, 80, "http", "http"),
+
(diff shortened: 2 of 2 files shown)


rhqci:

abonas:
@burmanm please review

xeviknal:
@toothbrush it looks good so far. I will give an extra eye next Monday. Thanks for the PR. Reach you out later.

israel-hdez:
@toothbrush For completeness, can you please add a test to check that a gateway with `*.example.com` and `thisisNotfine.example.com` will raise the warning?
I think your changes are good and will pass the test.

hhovsepy:
Have created 2 gateways, different names but the same two hosts in each of them.
Only the fist hosts are validated.
Screenshot from 2019-07-29 13-01-59
Screenshot from 2019-07-29 13-02-45

burmanm:
@hhovsepy That's a separate issue, please create ticket for that (and assign to me).

xeviknal:
@hhovsepy is faster than wind: https://issues.jboss.org/browse/KIALI-3171 :smile:

xeviknal:
@hhovsepy I have also tested this one in master and looks good to me:
Screenshot of Kiali  istio-system  (4)
Screenshot of Kiali  istio-system  (3)

hhovsepy:
@xeviknal your case is the case sensitive one #1286

xeviknal:
Wrong upload files. The following should be the ones:
Screenshot of Kiali  istio-system  (6)
Screenshot of Kiali  istio-system  (5)

PR #1286, opened 2019-07-29
Gateway host/port matching: make case-insensitive.  

### LINKED ISSUE #1285: Missing warning on duplicate host/port when case doesn't match
This is quite similar to my previous issue, #1272.  I expect to get a warning when adding both `foo.com` and `FOO.com` as hosts to a `Gateway`, while in fact, [DNS should be case insensitive](https://tools.ietf.org/html/rfc4343).

I've got a fix incoming, stay tuned.

toothbrush:
It turns out that Kiali doesn't throw a warning if adding both `FOO.com` and `foo.com` to a gateway, while in fact, [DNS should be case insensitive](https://tools.ietf.org/html/rfc4343).

I've updated the matcher to downcase both hostnames before comparison to achieve this.  I could've use a [case-insensitive flag on Golang's `regexp` library](https://tip.golang.org/pkg/regexp/syntax/) but that felt like overkill.

Closes #1285.

* business/checkers/gateways/multi_match_checker.go(findMatch): Adapt  the function to do case-insensitive match on hostnames.
* business/checkers/gateways/multi_match_checker_test.go(TestCaseMatching):   Add test case for the above.

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -122,8 +122,8 @@ func (m MultiMatchChecker) findMatch(host Host) (bool, []Host) {
 			}
 
 			// Either one could include wildcards, so we need to check both ways and fix "*" -> ".*" for regexp engine
-			current := strings.Replace(host.Hostname, "*", ".*", -1)
-			previous := strings.Replace(h.Hostname, "*", ".*", -1)
+			current := strings.ToLower(strings.Replace(host.Hostname, "*", ".*", -1))
+			previous := strings.ToLower(strings.Replace(h.Hostname, "*", ".*", -1))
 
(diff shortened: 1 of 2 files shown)


rhqci:

hhovsepy:
The issue still exists in the case when 2 gateways are created.
Screenshot from 2019-07-29 13-13-06
Screenshot from 2019-07-29 13-13-21

xeviknal:
@hhovsepy I have the following behavior tested in master:
Screenshot of Kiali  istio-system  (2)
Screenshot of Kiali  istio-system  (1)

I think is correct.

hhovsepy:
@xeviknal thanks, you are right, my issue was happening on v1.0 only: https://github.com/kiali/kiali/pull/1288

PR #1288, opened 2019-07-29
[v1.0] Gateway validations fixes: #1284 and #1286

### LINKED ISSUE #1272: Incorrect warning about duplicate host/port config on Gateways

### LINKED ISSUE #1285: Missing warning on duplicate host/port when case doesn't match

xeviknal:
Cherry-pick from 
https://github.com/kiali/kiali/pull/1284
https://github.com/kiali/kiali/pull/1286

* Gateway validations: tighten up duplicate host matching regex.

This is to address an issue where a perfectly legitimate `Gateway`
with hostnames `example.com` and `foo.example.com` would get flagged
with a duplicate-host warning.  This was as a result of an unanchored
regular expression.  This PR introduces a test case to illustrate it,
as well as a fix.

Closes #1272.

* Gateway host/port matching: make case-insensitive.
It turns out that Kiali doesn't throw a warning if adding both FOO.com and foo.com to a gateway, while in fact, DNS should be case insensitive.

I've updated the matcher to downcase both hostnames before comparison to achieve this. I could've use a case-insensitive flag on Golang's regexp library but that felt like overkill.

Closes #1285.

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -125,7 +125,15 @@ func (m MultiMatchChecker) findMatch(host Host) (bool, []Host) {
 			current := strings.Replace(host.Hostname, "*", ".*", -1)
 			previous := strings.Replace(h.Hostname, "*", ".*", -1)
 
-			if regexp.MustCompile(current).MatchString(previous) || regexp.MustCompile(previous).MatchString(current) {
+			// We anchor the beginning and end of the string when it's
+			// to be used as a regex, so that we don't get spurious
+			// substring matches, e.g., "example.com" matching
+			// "foo.example.com".
+			current_re := strings.Join([]string{"^", current, "$"}, "")
+			previous_re := strings.Join([]string{"^", previous, "$"}, "")
+
+			if regexp.MustCompile(current_re).MatchString(previous) ||
+				regexp.MustCompile(previous_re).MatchString(current) {
 				duplicates = append(duplicates, h)
 				break
 			}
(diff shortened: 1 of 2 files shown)


xeviknal:
Thank you @hhovsepy :)

PR #1295, opened 2019-07-31
add kiali-operator psp

Sallyan:
Add Pod Security Policy for Kiali-operator.

### DIFF
--- a/operator/deploy/operator-psp.yaml
+++ b/operator/deploy/operator-psp.yaml
@@ -1,0 +1,22 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: kiali-operator-psp
+  labels:
+    app: kiali-operator
+    version: v1.1.0
+spec:
+  fsGroup:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+  - downwardAPI
+  - hostPath
+  - secret
+  - configMap
+  - emptyDir
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -8,6 +8,11 @@ metadata:
     app: kiali-operator
     version: ${OPERATOR_VERSION_LABEL}
 rules:
+- apiGroups: ["policy"]
+  resources: ["podsecuritypolicies"]
+  verbs: ["use"]
+  resourceNames:
+    - kiali-operator-psp
 - apiGroups: [""]
   resources:
   - configmaps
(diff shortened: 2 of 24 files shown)


rhqci:

pilhuhn:
This is great, thanks for submitting!

Sallyan:
There should be one more change in deploy-kiali-operator.sh. 
`# Now deploy all the Kiali operator components.
echo "Deploying Kiali operator to namespace [${OPERATOR_NAMESPACE}]"

for yaml in namespace crd service_account **operator_psp** role role_binding operator
do
  apply_operator_resource ${yaml}

  if [ "$?" != "0" ]; then
    echo "ERROR: Failed to deploy Kiali operator. Aborting."
    exit 1
  fi
done`

operator/deploy/role.yaml
```diff
@@ -8,6 +8,11 @@ metadata:
     app: kiali-operator
     version: ${OPERATOR_VERSION_LABEL}
 rules:
+- apiGroups: ["policy"]
```
jmazzitelli: for consistency, use the same syntax and indentation that is used throughout this file. So it should be:

```
- apiGroups: ["policy"]
  resourceNames:
  - kiali-operator
  resources:
  - podsecuritypolicies
  verbs:
  - use
```

operator/deploy/operator-psp.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: kiali-operator-psp
```
jmazzitelli: by convention, all operator entities (such as the Deployment, ClusterRole, ServiceAccount, etc) are named "kiali-operator". So change this to follow that convention:

```suggestion
  name: kiali-operator
```

operator/deploy/operator-psp.yaml
```diff
@@ -0,0 +1,22 @@
...
+  labels:
+    app: kiali-operator
+    version: v1.1.0
```
jmazzitelli: These files here are templates - do not hardcode the version. It will be set by the deploy script. So change this to the following:

```suggestion
    version: ${OPERATOR_VERSION_LABEL}
```

operator/deploy/operator-psp.yaml
jmazzitelli: Note that by convention the names of the files here are not prefixed with "operator-" (so you do not see things like "operator-role.yaml" or "operator-namespace.yaml" - its just "role.yaml" and "namespace.yaml".

So rename this file from operator-psp.yaml to simply `psp.yaml`

operator/deploy/operator-psp.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
```
jmazzitelli: @mwringe can you review this PSP spec. I'm not familiar with the security implications here - perhaps you do?

operator/deploy/operator-psp.yaml
Sallyan: Thanks! I will change the names convention.

Sallyan:
Sorry for the confuse. Will create new PR.

PR #1306, opened 2019-08-02
fix gateway checker does not recognize numbers

sniperking1234:
** Describe the change **

Increase recognition range


rhqci:

burmanm:
Is this a real problem somewhere? I remember this feature as it should be uint32 according to the documentation, but our parsing (JSON) returned it back as int64 everytime. In which cases did this come back as uint32 / int32 from the parser?

sniperking1234:
@burmanm This happens when I run unit tests `TestCorrectGateways`.
System: Windows10 x64_86
go version:  go1.12.5 windows/amd64
image

burmanm:
@hhovsepy Do test for anything else than Linux at all? This is interesting if the behavior is different with Windows and Linux versions of Go.

business/checkers/gateways/multi_match_checker.go
```diff
@@ -86,8 +87,9 @@ func parsePortAndHostnames(serverDef map[string]interface{}) []Host {
 	if portDef, found := serverDef["port"]; found {
 		if ports, ok := portDef.(map[string]interface{}); ok {
 			if numberDef, found := ports["number"]; found {
-				if portNumber, ok := numberDef.(int64); ok {
-					port = int(portNumber)
+				portNumber, e := intutil.Convert(numberDef)
+				if e != nil {
```
burmanm: Isn't this incorrect in any case? If err == nil, then we can use the value. Otherwise it is the default value. And then merge the lines:

```go
if portNumber, err := intutil.Convert(numberDef); err == nil {
  port = portNumber
}
```

sniperking1234: You're right

israel-hdez:
Thanks @sniperking1234

hhovsepy:
Regression test passed.

Issue #1294, opened 2019-07-31
Kiali Operator to support Pop Security Policy


Sallyan:
**Issue**: 
Our GCP cluster is enabled Pod Security Policy. After run the kiali-operator deploy script, the containers are not started. 
Pod status: CreateContainerConfigError
Error: container has runAsNonRoot and image will run as root

**Solution**: 
Need to add PSP for kiali-operator and update the role spec, deploy/merge scripts.

jmazzitelli:
This issue needs more details. Please fill in the "new issue" template so it is more clear what you are asking for. I am including the template below so you don't have to press the "new issue" button again:

----

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

**Describe the solution you'd like**
A clear and concise description of what you want to happen.

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've considered.

**Additional context**
Add any other context or screenshots about the feature request here. If it is about a new design, please attach a drawing of a potential solution.
For backend changes also describe how one could test the new feature and how it impacts display on the UI.

jshaughn:
@Sallyan @jmazzitelli I'm adding the Stale label here.  Should this be closed?

lucasponce:
No new info from this issue.
I'm going to close it, please, if it's still open, add an update and we can re-open and plan an action on it.

jmazzitelli:
FYI, PSP (PodSecurityPolicy) will soon be deprecated in Kubernetes. So we won't be doing anything wrt PSPs since it is going away soon. See: https://github.com/kubernetes/kubernetes/pull/97171

`The PodSecurityPolicy API is deprecated in 1.21, and will no longer be served starting in 1.25.`

PR #1307, opened 2019-08-02
Add Pod Security Policy for kiali-operator

Sallyan:
**Why Kiali-operator needs PSP?**
If the cluster enables Pod Security Policy, the kiali-operator pods will fail to start with error "Error: container has runAsNonRoot and image will run as root"

 **Describe the change** 

Add PodSecurityPolicy for kiali-operator, Permits the container to run with root privileges
update the role and rolebinding definition to bind the sa to adopt the PSP. 
Accordingly modify the deploy and merge script to process the psp file. 

 **Issue reference** 
https://github.com/kiali/kiali/issues/1294

### DIFF
--- a/operator/deploy/deploy-kiali-operator.sh
+++ b/operator/deploy/deploy-kiali-operator.sh
@@ -568,7 +568,7 @@ delete_operator_resources() {
     ${CLIENT_EXE} delete --ignore-not-found=true customresourcedefinitions --selector="app=kiali-operator"
 
     # now purge all operator resources
-    ${CLIENT_EXE} delete --ignore-not-found=true all,sa,deployments,roles,rolebindings,clusterroles,clusterrolebindings --selector="app=kiali-operator" -n "${OPERATOR_NAMESPACE}"
+    ${CLIENT_EXE} delete --ignore-not-found=true all,sa,deployments,roles,rolebindings,clusterroles,clusterrolebindings,podsecuritypolicies --selector="app=kiali-operator" -n "${OPERATOR_NAMESPACE}"
   fi
 
   # Clean up the operator namespace entirely but only if there are no pods running in it.
@@ -854,7 +854,7 @@ apply_operator_resource() {
 # Now deploy all the Kiali operator components.
 echo "Deploying Kiali operator to namespace [${OPERATOR_NAMESPACE}]"
 
-for yaml in namespace crd service_account role role_binding operator
+for yaml in namespace crd service_account psp role role_binding operator
 do
   apply_operator_resource ${yaml}
 
--- a/operator/deploy/psp.yaml
+++ b/operator/deploy/psp.yaml
@@ -1,0 +1,22 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: kiali-operator
+  labels:
+    app: kiali-operator
+    version: ${OPERATOR_VERSION_LABEL}
+spec:
+  fsGroup:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+  - downwardAPI
+  - hostPath
+  - secret
+  - configMap
+  - emptyDir
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -8,6 +8,11 @@ metadata:
     app: kiali-operator
     version: ${OPERATOR_VERSION_LABEL}
 rules:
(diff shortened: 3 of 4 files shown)


Sallyan:
Old PR and communications: 
https://github.com/kiali/kiali/pull/1295

rhqci:

Sallyan:
@jmazzitelli @gbaufake
Please help to review. Thanks!

operator/deploy/role.yaml
```diff
@@ -8,6 +8,11 @@ metadata:
     app: kiali-operator
     version: ${OPERATOR_VERSION_LABEL}
 rules:
+- apiGroups: ["policy"]
```
jmazzitelli: see the old communications on this part of the yaml: https://github.com/kiali/kiali/pull/1295#discussion_r309305838

I'll repeat it here:

for consistency, use the same syntax and indentation that is used throughout this file. So it should be:

```
- apiGroups: ["policy"]
  resourceNames:
  - kiali-operator
  resources:
  - podsecuritypolicies
  verbs:
  - use
```

operator/deploy/deploy-kiali-operator.sh
```diff
@@ -568,7 +568,7 @@ delete_operator_resources() {
     ${CLIENT_EXE} delete --ignore-not-found=true customresourcedefinitions --selector="app=kiali-operator"
 
     # now purge all operator resources
-    ${CLIENT_EXE} delete --ignore-not-found=true all,sa,deployments,roles,rolebindings,clusterroles,clusterrolebindings --selector="app=kiali-operator" -n "${OPERATOR_NAMESPACE}"
+    ${CLIENT_EXE} delete --ignore-not-found=true all,sa,deployments,roles,rolebindings,clusterroles,clusterrolebindings,podsecuritypolicies --selector="app=kiali-operator" -n "${OPERATOR_NAMESPACE}"
```
jmazzitelli: This same kind of change should be applied to the Makefiles where appropriate. I just looked and I think there is only one place where you need to change - it is here: https://github.com/kiali/kiali/blob/v1.2/operator/Makefile#L155

Make sure that Makefile is deleting the same resources as this line 571 of deploy-kiali-operator.sh is

operator/deploy/psp.yaml
```diff
@@ -0,0 +1,22 @@
...
+  fsGroup:
+    rule: RunAsAny
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+  - downwardAPI
```
jmazzitelli: As per https://kubernetes.io/docs/concepts/policy/pod-security-policy/#volumes-and-file-systems they recommend a minimum set of the following:

```
- configMap
- downwardAPI
- emptyDir
- persistentVolumeClaim
- secret
- projected
```

Of that list, it looks like you are missing `persistentVolumeClaim` and `projected`. I think you should add those unless you have a reason not to.

Sallyan:
> See requested changes.
> 
> @Sallyan Briefly, can you explain the purpose/justification of adding the PSP - why Kiali operator needs it and what the current PSP configured in this PR will do? Note that PSP is currently only in beta stage, so we want to make sure we are not going to introduce something that we don't need right now and may change in the future.
> 
> Also, this looks to be a PSP for the operator only - is there a reason you want a PSP for the operator and not for Kiali itself (i.e. do you think Kiali itself will need a PSP?).

Hi @jmazzitelli 
1. If you really think Kiali-operator doesn't need Pop Security Policy right now, i can customize the deploy script according to my requirements and add the PSP file automatically to the cluster which has enabled PSP.  Then we can ignore this PR change. 
2. Why Kiali itself not need PSP?
By default Kiali is installed in namespace istio-system where istio is deployed. If the cluster has PSP enabled, the istio has configured pod security policy and according clusterrole , clusterrolebinding
The kiali deployment ansible playbook will create sa, and add this sa to istio clusterrolebind. So the Kiali pod will pass the Security policy and can run successfully.

jmazzitelli:
> 1. If you really think Kiali-operator doesn't need Pop Security Policy right now, i can customize the deploy script according to my requirements and add the PSP file automatically to the cluster which has enabled PSP.  Then we can ignore this PR change.

I think that would be the best thing to do right now. We can leave this PR open but I'll put a "do not merge" label on it, just so we still have it around but we won't merge now. Let's keep an eye on the PSP feature as it matures. But for now, I think we'll want to avoid using it.

Sallyan:
@jmazzitelli Ok, Thanks!

lucasponce:
@jmazzitelli @Sallyan thanks for working on this issue. Any update of the status ?

PR #1319, opened 2019-08-06
KIALI-2938 Better support detailed error messages

jshaughn:
- add RespondWithDetailedError() for reporting an error with additional
  detail (stack trace, or whatever)
- use formal type for error responses
- graph code now returns a stack trace for app errors
- fix multiple-response bug in error response handling in graph code

*** UI PR: https://github.com/kiali/kiali-ui/pull/1353

### DIFF
--- a/handlers/base.go
+++ b/handlers/base.go
@@ -5,10 +5,15 @@ import (
 	"net/http"
 )
 
+type responseError struct {
+	Error  string `json:"error,omitempty"`
+	Detail string `json:"detail,omitempty"`
+}
+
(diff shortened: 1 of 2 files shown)


gbaufake:
I just see some UX issue because you need to slide vertically to see the error.

jshaughn:
> I just see some UX issue because you need to slide vertically to see the error.

True, but only when expanding.  I think it's better than the old single-wrapped-string but if UX has some suggestion going forward it's fine by me.

PR #1320, opened 2019-08-07
[WIP] [operator] issue 1315 - operator now supports istio_component_namespaces

jmazzitelli:
See issue #1315

@jshaughn code we can use to get operator to support the new istio_component_namespaces

PR #1321, opened 2019-08-07
KIALI-3195 Support Istio component namespaces

jshaughn:
https://issues.jboss.org/browse/KIALI-3195

- add config support for multiple namespaces
- remove obsolete includeIstio query parameter support
- add support for multi-namespace intra-istio and inter-istio traffic

This operator PR supports configuration of the component namespaces: 
https://github.com/kiali/kiali/pull/1320

*** UI PR *** https://github.com/kiali/kiali-ui/pull/1361


config/config.go
```diff
@@ -230,20 +230,25 @@ type DeploymentConfig struct {
...
-	API              ApiConfig         `yaml:"api,omitempty"`
-	Auth             AuthConfig        `yaml:"auth,omitempty"`
-	Deployment       DeploymentConfig  `yaml:"deployment,omitempty"`
+	Identity         security.Identity        `yaml:",omitempty"`
+	Server           Server                   `yaml:",omitempty"`
+	InCluster        bool                     `yaml:"in_cluster,omitempty"`
+	ExternalServices ExternalServices         `yaml:"external_services,omitempty"`
+	LoginToken       LoginToken               `yaml:"login_token,omitempty"`
+	IstioNamespace   string                   `yaml:"istio_namespace,omitempty"` // default component namespace
+	IstioNamespaces  IstioComponentNamespaces `yaml:"istio_namespaces,omitempty"`
```
jmazzitelli: yaml is istio_component_namespaces
```suggestion
	IstioNamespaces  IstioComponentNamespaces `yaml:"istio_component_namespaces,omitempty"`
```

config/config.go
```diff
@@ -517,3 +522,12 @@ func getAuthFromEnv(prefix string) Auth {
 	auth.CAFile = strings.TrimSpace(getDefaultString(prefix+EnvAuthSuffixCAFile, ""))
 	return auth
 }
+
+// IsIstioNamespace returns true if the namespace is the default istio namespace or an Istio component namespace
+func IsIstioNamespace(namespace string) bool {
+	if namespace == configuration.IstioNamespace {
+		return true
+	}
+	_, found := configuration.IstioNamespaces[namespace]
```
jmazzitelli: I think this is backwards. The key is the component name, the value is the namespace. You want to see if 
"namespace" is a value anywhere in any key of IstioNamespaces (which, IMO, should be called IstioComponentNamespaces)

config/config.go
jshaughn: Yes, I had already changed it locally, will be updated in next commit.

config/config.go
jshaughn: yeah, it's wrong.  will fix.

config/config.go
```diff
@@ -398,6 +391,21 @@ func getDefaultStringArray(envvar string, defaultValue string) (retVal []string)
...
 
+func getDefaultStringMap(envvar string, defaultValue string) (retVal map[string]string) {
+	csv := os.Getenv(envvar)
+	if csv == "" {
+		csv = defaultValue
+	}
+	retVal = map[string]string{}
+	for _, token := range strings.Split(csv, ",") {
+		mapEntry := strings.Split(token, "=")
+		if len(mapEntry) == 2 {
```
jmazzitelli: you should add an else-stmt here and in that else you should log a warning message here because if the length here is 1, it means the person misconfigured this (forgot the "=" for example -- "aa=bb,cc,dd" instead of "aa=bb,cc=dd")

config/config.go
```diff
@@ -398,6 +391,21 @@ func getDefaultStringArray(envvar string, defaultValue string) (retVal []string)
...
 }
 
+func getDefaultStringMap(envvar string, defaultValue string) (retVal map[string]string) {
+	csv := os.Getenv(envvar)
+	if csv == "" {
+		csv = defaultValue
+	}
+	retVal = map[string]string{}
+	for _, token := range strings.Split(csv, ",") {
+		mapEntry := strings.Split(token, "=")
```
jmazzitelli: you should force this to split with a max of 2 using `SplitN` (I don't think namespaces can have = chars, but still we should force this to only split on the first = it sees anyway).
```suggestion
		mapEntry := strings.SplitN(token, "=", 2)
```

config/config.go
jshaughn: Added warning.

config/config.go
jshaughn: Good tip, changed to SplitN.

PR #1253, opened 2019-07-17
Support ldap

Maneshme:
** Describe the change **

This is change is to integrate LDAP based authentication into the Kiali API

** Backwards incompatible? **

- Is your change introducing backwards incompatible changes?
No
_describe the changes if appropriate_

Related UI PR: https://github.com/kiali/kiali-ui/pull/1335

Issue #1053, opened 2019-05-02
LDAP integration


sujithvs74:
We are looking for an option to integrate LDAP with kiali-ui, which will enable us to share kiali to application team to view status of services. Is there any option for it.

israel-hdez:
Hi, @sujithvs74 

Currently, Kiali does not support integration with LDAP. There are plans for supporting other authentication systems, but are for medium to long term. Also, we will be happy to accept PRs if you know and have the time to implement it.

Another user has previously asked something similar. See: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kiali-dev/n1J07xRmMzQ/w3DoOBvwAQAJ. Apparently, his workaround was to put Kiali behind a reverse proxy capable of doing the authentication. I suggest you try that route if a reverse proxy is feasible for you and you need it soon.

israel-hdez:
Hi, @sujithvs74 
I'm wondering if you would like to collaborate and implement this feature.
Nodbody in the team knows well LDAP. If you have the knowledge, I would like to know if you want to help for this.

Maneshme:
I'm current have implemented this feature

jmazzitelli:
I'm in the process of finalizing the two PRs (one server side, one UI side). Molecule tests show the feature working. So this should be ready to merge fairly soon.

jmazzitelli:
PR with server-side changes: https://github.com/kiali/kiali/pull/1323
PR with UI changes: https://github.com/kiali/kiali-ui/pull/1335

jmazzitelli:
This feature has been implemented and is now in master.

PR #1323, opened 2019-08-08
issue 1053 - LDAP integration

jmazzitelli:
** Describe the change **

This is based on PR https://github.com/kiali/kiali/pull/1253 (and requires the UI PR https://github.com/kiali/kiali-ui/pull/1335). This adds a new supported `auth.strategy` of `ldap`. When users attempt to log in, they provide credentials that are authenticated via LDAP on the backend.

There is a molecule test that shows this stuff working - it tests unsecure and secure access to an LDAP server. To run the test, oc login to a cluster that has OpenShift deployed and then execute `cd operator; MOLECULE_SCENARIO=ldap-test make molecule-test` (you will have to edit the molecule.yml to point to your image versions and repos).

The `operator/deploy/kiali/kiali_cr.yaml` documents the new `auth.ldap` settings that you provide to the Kiali CR to enable this new feature. README also has a blurb on this, too.

** Issue reference **

https://github.com/kiali/kiali/issues/1053

** Backwards incompatible? **

No


jmazzitelli:
the glide deps haven't been updated. I will have to do that.

jmazzitelli:
This is now ready for review and merge. I do not expect any more changes or commits - unless in the future we need to rebase yet again on master. But as of right now, this is rebased on current master.

jmazzitelli:
that didn't take long - I'll need to rebase and resolve the conflicts

jmazzitelli:
@gbaufake I put you as a reviewer here (as well as to the UI PR that goes with this). See if there is anything here that might cause QE to add additional tests. This PR does already have a molecule test. 

I want to merge this early next week if possible.

jmazzitelli:
After talking with QE they agree we can merge this now, and we'll formally test it through QE process later. We have a molecule test that does show this working.

README.adoc
```diff
@@ -385,6 +385,28 @@ apidocs:
 ...
 ---
 
+=== LDAP Configuration
+To configure authentication strategy to LDAP add the following configuration to the yaml file. If the auth strategy is LDAP. The application will not start without the basic LDAP settings like ldap_host, ldap_port, ldap_base and ldap_bind_dn.
```
jshaughn: Not really a blocker but it may be helpful to indicate that the values below are just example values that should be changed.

Also,  omit "If the auth strategy is LDAP."

handlers/authentication.go
```diff
@@ -282,6 +328,22 @@ func checkKialiSession(w http.ResponseWriter, r *http.Request) int {
 	return http.StatusOK
 }
 
+// checkLDAPSession is to check validaity of the LDAP session
```
jshaughn: typo: validaity -> validity

ldap/jwt.go
```diff
@@ -0,0 +1,101 @@
...
+		return u, err
+	}
+
+	// Token is valid so fill in the rest of u with happy state and return it
+	auth = true
+	u.Status.Authenticated = &auth
+	u.Status.User = &User{Username: claims.Username, UID: claims.UID, Groups: claims.Groups}
+
+	return u, nil
+
```
jshaughn: unnecessary newline

ldap/jwt.go
```diff
@@ -0,0 +1,101 @@
...
+func ValidateToken(token string) (UserInfo, error) {
+	return validate(token)
+}
+
+// validate does much of the work of ValidateToken
+func validate(bearerToken string) (UserInfo, error) {
+
+	signingKey := config.Get().LoginToken.SigningKey
+
+	var auth bool
```
jshaughn: This may be more clear as:  `auth := false`

ldap/jwt.go
```diff
@@ -0,0 +1,101 @@
...
+
+	token, err := jwt.ParseWithClaims(bearerToken, &claims, func(token *jwt.Token) (interface{}, error) {
+		if !strings.HasPrefix(token.Method.Alg(), "HS") { // HMAC are the only allowed signing methods
+			log.Errorf("Unexpected signing method: %s", token.Method.Alg())
+			return nil, fmt.Errorf("Unexpected signing method: %s", token.Method.Alg())
+		}
+		return []byte(signingKey), nil
+	})
+
+	u := UserInfo{ // user we'll return, initially in error state
```
jshaughn: I find it odd that this is placed between the call above and the error check below.  It would be better above, the call to ParseWithClaims.

ldap/jwt.go
```diff
@@ -0,0 +1,101 @@
...
+		APIVersion: "authentication.k8s.io/v1beta1",
+		Kind:       "TokenReview",
+		Status: &Status{
+			Authenticated: &auth,
+			User:          nil,
+		},
+	}
+
+	if !token.Valid {
```
jshaughn: This looks a little unsafe to me, maybe change to:  `if err != nil || !token.valid`

ldap/jwt.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+// GetTokenStringFromRequest is to get the token string from the request
+func GetTokenStringFromRequest(r *http.Request) string {
+	tokenString := "" // Default to no token.
+
+	// Token can be provided by a browser in a Cookie or
+	// in an authorization HTTP header.
+	// The token in the cookie has priority.
+	if authCookie, err := r.Cookie(config.TokenCookieName); err != http.ErrNoCookie {
```
jshaughn: Is that really the only err that can be returned?

ldap/ldap.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/jtblin/go-ldap-client"
+	ldapv2 "gopkg.in/ldap.v2"
+
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/log"
+)
+
+//ValidateUser validates user by extracting creds from the request body, then
+//    authenticating against an ID backend and filling out the User struct therefrom
```
jshaughn: minor:
//V -> // V
therefrom -> from there.

ldap/ldap.go
```diff
@@ -0,0 +1,127 @@
...
+		BindDN:             bindDN,
+		BindPassword:       pwd,
+		UserFilter:         authConfig.LDAP.LDAPUserFilter,
+		GroupFilter:        authConfig.LDAP.LDAPGroupFilter,
+		Attributes:         ldapAttributes,
+	}
+	defer client.Close()
+
+	// uname and password (ID) correct?
+	authOk, err := isAuthenticated(&client, username, pwd)
```
jshaughn: I think we need to check for err != nil

ldap/ldap.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if authOk { // get the rest of the info for this person -- groups, app ownershop etc.
```
jshaughn: Similar to above, I think you want:  `if err != nil && authOk`.  Return the failure case first, then getUser.

ldap/jwt.go
jmazzitelli: This I do not agree with. I don't know this API, but if the token is not valid, we should not care what err is (nil or otherwise). If token is not valid, we need to return regardless if err is nil or not

ldap/jwt.go
jmazzitelli: This is odd. I don't get this. If err is non-nil and NOT ErrNoCookie, then its still going in here - but in that case, authCookie most likely will be nil and access to nil here will blow up. I'll change this around to be more fault tolerant. This is a good catch.

ldap/ldap.go
jmazzitelli: See my comments below.

ldap/ldap.go
jmazzitelli: This is not correct. I will add comments to isAuthenticated to make this clear though. If you look, isAuthenticated may return false with an err == nil (a side effect of the call to ldapClient.Authenticate). So we can't assume anything based on err (i.e. things may have failed even if err is nil).

ldap/jwt.go
jshaughn: OK, then:  `token == nil || !token.Valid`

README.adoc
jmazzitelli: I re-wrote this a bit.

handlers/authentication.go
jmazzitelli: done

ldap/jwt.go
jmazzitelli: done

ldap/jwt.go
jmazzitelli: done

ldap/jwt.go
jmazzitelli: done

ldap/jwt.go
jmazzitelli: done

ldap/jwt.go
jmazzitelli: I changed the if expression to check "err == nil && authCookie != nil" - in other words, do not look at the cookie UNLESS there is no error AND authCookie is not nil.

ldap/ldap.go
jmazzitelli: done

ldap/ldap.go
jshaughn: OK, since authOk is a bool it's safe as is.

pilhuhn:
Thank you @Maneshme for all your work on this!

ldap/ldap.go
```diff
@@ -0,0 +1,130 @@
...
+
+	for _, entry := range sr.Entries {
+		for _, attr := range entry.Attributes {
+			if attr.Name == authConfig.LDAP.LDAPUserIDKey {
+				user.UID = attr.Values[0] //There will always be one CN
+			} else if attr.Name == authConfig.LDAP.LDAPMailIDKey {
+				user.Username = attr.Values[0] //There will always be one mail
+			} else {
+				for _, value := range attr.Values {
+					groupName := value[strings.Index(value, "CN=")+3 : strings.Index(value, ",")]
```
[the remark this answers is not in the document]
jmazzitelli: @Maneshme thoughts?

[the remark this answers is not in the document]
kenske: @Maneshme have you had a chance to see this?

PR #1330, opened 2019-08-12
(Trivial) give more context on APIDoc client calls errors

jotak:
Trivial, this is only changing some error logs to give context about API Documentation being involved.

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -143,14 +143,14 @@ func (in *SvcService) GetServiceApiDocumentation(namespace, service string) (str
 
 	resp, err2 := http.Get(apiSpecPath)
 	if err2 != nil {
-		log.Errorf("GET error: %v", err)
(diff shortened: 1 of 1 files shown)


abonas:
is API documentation feature documented on kiali.io?

business/services.go
```diff
@@ -143,14 +143,14 @@ func (in *SvcService) GetServiceApiDocumentation(namespace, service string) (str
 
 	resp, err2 := http.Get(apiSpecPath)
 	if err2 != nil {
-		log.Errorf("GET error: %v", err)
+		log.Errorf("API Documentation GET error: %v", err)
```
abonas: why is there a need to hardcode "GET"?

jotak: I was keeping the old message, just adding context. Now rewritten to give even more info

PR #1332, opened 2019-08-13
KIALI-3220 Fetch ServiceMeshPolicies from control-plane

lucasponce:
In Maistra scenarios, ServiceMeshPolicies need to be queried always from control-plane namespace.
This created some problems in the required object used by Kiali validations to perform DestinationRule checks.

### DIFF
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -2,6 +2,7 @@ package business
 
 import (
 	"fmt"
+	"github.com/kiali/kiali/config"
 	"sync"
 
 	apps_v1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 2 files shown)


business/istio_validations.go
```diff
@@ -2,13 +2,15 @@ package business
 
 import (
 	"fmt"
+	"github.com/kiali/kiali/config"
```
jmazzitelli: FYI: this doesn't follow our standard practice of putting the kiali imports in the bottom section. Same with the tls.go imports.

jotak: Isn't it something that can be handled automatically for us in goland (i think it's what lucas is using) ? Personally with vscode I almost never have this issue as imports are automatically grouped as we want on save.

lucasponce: I had to install a workspace in the IDE and I had an "automatically optimization" in the imports.
Thanks for the pointer.

business/tls.go
```diff
@@ -2,9 +2,11 @@ package business
 
 import (
 	"fmt"
+	"github.com/kiali/kiali/config"
```
jmazzitelli: this should be moved also

PR #1336, opened 2019-08-13
Add NPM registry parameter

israel-hdez:
Adds NPM_CONFIG_REGISTRY parameter to the sample job.

Unfortunately, since different envs may need to use different package registries it's no longer possible to hardcode the parameters to make the Pipeline to "auto configure" itself if created manually (harcoding configs replaces any customized configs of the Pipeline on every run)

This means, that the Pipeline will need to be configured manually if the config.xml file cannot be placed in the Jenkins instance.

### DIFF
--- a/deploy/jenkins-ci/Jenkinsfile
+++ b/deploy/jenkins-ci/Jenkinsfile
@@ -1,73 +1,54 @@
-properties([
-  disableConcurrentBuilds(),
-  parameters([
-    string(name: 'RELEASE_TYPE',
-      defaultValue: 'auto',
-      trim: true,
-      description: 'Valid values are: auto, minor, snapshot.0, snapshot.1, edge. When "auto" is\n' +
-                    'specified, the type of the release will be determined based on the current date.'),
-    string(name: 'BACKEND_RELEASING_BRANCH',
-      defaultValue: 'refs/heads/master',
-      trim: true,
-      description: 'Branch of the backend to release'),
-    string(
-      name: 'UI_RELEASING_BRANCH',
-      defaultValue: 'refs/heads/master',
-      trim: true,
-      description: 'Branch of the UI to release'),
-    string(
-      name: 'BACKEND_GITHUB_URI',
-      defaultValue: 'git@github.com:kiali/kiali.git',
-      trim: true,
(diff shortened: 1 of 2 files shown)


josejulio:
Can't we default to `NPM_CONFIG_REGISTRY` not being set and only override that one as needed?

israel-hdez:
@josejulio I think I don't understand :/

josejulio:
> Adds NPM_CONFIG_REGISTRY parameter to the sample job.
> 
> Unfortunately, since different envs may need to use different package registries it's no longer possible to hardcode the parameters to make the Pipeline to "auto configure" itself if created manually (harcoding configs replaces any customized configs of the Pipeline on every run)

I saw that you commented all the defaults (I think that's what you did :P) 
But is not clear to me why we can't let the pipeline to "auto configure" with the param `NPM_CONFIG_REGISTRY` being "unset" and only set when we need it.

> 
> This means, that the Pipeline will need to be configured manually if the config.xml file cannot be placed in the Jenkins instance.

israel-hdez:
@josejulio Oh! I think I understand... So you want the NPM_CONFIG_REGISTRY to be present in the `properties` but without the `defaultValue` attribute...
That may work and will remove the need of commenting out. I'll check.

israel-hdez:
@josejulio It didn't work.

Basically, Jenkins is syncing the config with whatever is specified in the `properties`, if present. This happens on every run. I just tried and it sets the default to empty even if no `defaultValue` is specified.

Perhaps "auto configure" is the wrong term. It syncs on each run. Since we need a specific NPM_CONFIG_REGISTRY default value on our "prod" Jenkins instance, the `properties` become a problem.

My goal with the `properties` was to avoid setting up all those parameters each time I need to recreate my local dev env. So, what I was doing is to create a "bare" pipeline (only configuring the repo) and run it once so that it fetches the configs and in subsequent runs the parameters will be there. But since I already made some work to have a pre-configured Jenkins docker image, I think the hardcoded properties are no longer needed and the config.xml may enough?

I don't know... you say :)

josejulio:
> I think the hardcoded properties are no longer needed and the config.xml may enough?

+1

Issue #1337, opened 2019-08-14
Constantly get error message: Mesh-wide mTLS status feature disabled


jhasselgren:
**Describe the bug**
After the summer we suddenly get the following error message in Kiali in all of our environments (dev, test, prod):

>Mesh-wide mTLS status feature disabled., Info: [ the server could not find the requested resource (get meshpolicies.authentication.istio.io) ]
8 in 2 minutes
07/08/201909:46:45

After some troubleshooting done in [Kiali Forum](https://groups.google.com/d/msgid/kiali-users/24a3bf74-8c13-4b1a-a08d-612a02f49a24%40googlegroups.com?utm_medium=email&utm_source=footer)

This is our findings:
Kiali makes the following call:
> http://localhost:8080/apis/authentication.istio.io/v1alpha1/namespaces/anyNamespace/meshpolicies 

And gets a 404 as a reply

But if we makes the same call and ignoring the namespace part we get a reply
> http://localhost:8080/apis/authentication.istio.io/v1alpha1/meshpolicies

And the reason for this looks to be that meshpolicies in APIResourceList has "namespaced": false

**Versions used**
Kiali: 1.1.0 & 1.2.0
Istio: 1.2.0
Kubernetes flavour and version: version: v1.11.10-eks-f12431 (AWS EKS)

**Expected behavior**
That Kiali should be able to retrive the default meshpolicy with no errors

lucasponce:
@jhasselgren I suspect this is related the specific built of AWS EKS.
I'm testing Istio 1.2.1 upstream and I can run that query without a 404 error.

I need to investigate more.

lucasponce:
I've used for my tests a kubernetes version kubernetes v1.11.0+d4cacc0
 based on OpenShift 3.11 and deploying Istio 1.2.2 as described in Istio upstream https://istio.io/docs/setup/kubernetes/platform-setup/openshift/.

This is a starting point, just to spot if the potential issue was in Istio / Kiali code base of it's related to the platform as it looks like.

I'm trying to setup a AWS EKS environment to also confirm/reproduce the issue.

lucasponce:
Ok, I can get the same 404 error in the API.
Yes, there is a difference between how k8s platform deal with cluster-wide objects, it seems on this case AWS EKS is forcing in a strict way.
Anyway, I will study a fix in Kiali codebase.

[...]

PR #1346, opened 2019-08-16
KIALI-3223 Query cluster scoped resources without namespace

lucasponce:
Related to 
https://github.com/kiali/kiali/issues/1337 
https://issues.jboss.org/browse/KIALI-3223

As discussed, Istio cluster-wide MeshPolicy and ClusterRbacConfig will continue being managed in kiali from Istio control plane (typically "istio-system") namespace.

But, internally, queries are fixed with the right scope.

Note that this issue doesn't affect Maistra scenarios, as these objects have been removed there and Kiali logic has been updated.


xeor:
I was still getting `GetMeshPolicies failed during a TLS validation. Probably Kiali doesn't have cluster permissions. Error: meshpolicies.authentication.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "meshpolicies" in API group "authentication.istio.io" at the cluster scope` using `1.4.2` so I tried the `kiali/kiali:v1.5.0-snapshot.1`, and even that is getting the same error. Regression, or new bug? Anyone else?

I am using
* kiali-operator
* kiali v1.5.0-snapshot.1
* istio 1.3-20190914
* kubernetes v1.15.3

lucasponce:
@xeor how are yous installing upgrading Kiali ?

that sounds that serviceaccount doesn't have clusteradmin permissions as meshpolicies is a cluster wide object.

The workaround would be to provide cluster permissions to kiali serviceaccount.

A followup can be to disable the TLS check if kiali can't access to that object.

jmazzitelli:
Note that we have this permission: https://github.com/kiali/kiali/blob/master/operator/roles/kiali-deploy/templates/kubernetes/role.yaml#L111

However, if you do not grant Kiali access to the cluster as a whole (i.e. if you set accessible_namespaces to something other than **) then Kiali will not be given a ClusterRole. This could be the issue.

We'd need to see the Kiali CR, specifically its "deployment.accessible_namespaces" setting to confirm. Or look at the roles - if there isn't a cluster role named "kiali" or "kiali-viewer", that could be the issue.

lucasponce:
Actually, this shouldn't fail:
https://github.com/kiali/kiali/blob/master/business/tls.go#L53
if Kiali has not cluster wide permissions, it should log a warning in kiali server but not showing an error.

@xeor could you give us more details ? I guess we have something new here.

jmazzitelli:
@xeor confirm the Kiali version you have - in the UI About box AND look in the Kiali pod and confirm the container image is what we expect (i.e. v1.5 or v1.4)

lucasponce:
@xeor ok, could you confirm if that message is a warning in the kiali log ?
If yes, and UI is clean with no errors, yes, that's expected.
Probably we may want to change the log.

xeor:
Some more info..

```
$ kubectl -n istio-system get role kiali -o yaml | grep -C 3 meshpolicies
- apiGroups:
  - authentication.istio.io
  resources:
  - meshpolicies
  - policies
  verbs:
  - create
```

```
$ k get -n istio-system svc kiali -o wide
NAME    TYPE        CLUSTER-IP    EXTERNAL-IP   PORT(S)     AGE   SELECTOR
kiali   ClusterIP   10.233.0.96   <none>        20001/TCP   2d    app=kiali,version=v1.5.0-snapshot.1
```

```
$ k logs -l app=kiali,version=v1.5.0-snapshot.1 -f
I0917 19:25:21.932834       1 kiali.go:66] Kiali: Version: v1.5.0-snapshot.1, Commit: 836c9bdf1f3a68640c38f9ae06238f2db48364a4
I0917 19:25:21.934131       1 kiali.go:210] Using authentication strategy [anonymous]
W0917 19:25:21.934158       1 kiali.go:218] Kiali auth strategy is configured for anonymous access - users will not be authenticated.
I0917 19:25:21.934233       1 kiali.go:87] Kiali: Console version: 1.5.0-snapshot.1
I0917 19:25:21.934252       1 kiali.go:272] Updating base URL in index.html with [/kiali]
I0917 19:25:21.934743       1 kiali.go:253] Generating env.js from config
I0917 19:25:21.971611       1 server.go:55] Server endpoint will start at [:20001/kiali]
.... doing some requests from UI ....
E0917 19:26:27.328537       1 jaeger.go:27] Jaeger URL is not set in Kiali configuration
W0917 19:26:27.545523       1 tls.go:56] GetMeshPolicies failed during a TLS validation. Probably Kiali doesn't have cluster permissions. Error: meshpolicies.authentication.istio.io is forbidden: User "syst
em:serviceaccount:istio-system:kiali-service-account" cannot list resource "meshpolicies" in API group "authentication.istio.io" at the cluster scope
W0917 19:26:42.525421       1 tls.go:56] GetMeshPolicies failed during a TLS validation. Probably Kiali doesn't have cluster permissions. Error: meshpolicies.authentication.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "meshpolicies" in API group "authentication.istio.io" at the cluster scope
.... a bunch more TLS validation errors
```

There are no errors in the UI, except there are no graphs, even tho I generate some simple traffic... Also, it complains about jaeger and grafana, but I think that is irrelevant..

The kiali-cr looks like this

```
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  name: kiali
  namespace: kiali-operator
annotations:
  ansible.operator-sdk/reconcile-period: "0s"
spec:
  deployment:
    image_version: "v1.5.0-snapshot.1"

  auth:
    strategy: anonymous

  identity:
    cert_file: ""
    private_key_file: ""

  external_services:
    prometheus:
      custom_metrics_url: "http://prometheus-server.monitor:80"
      url: "http://prometheus-server.monitor:80"


```

jmazzitelli:
> $ kubectl -n istio-system get role

Yup - OK, you were given a "Role" not a "ClusterRole". Looking at the Kiali CR it is because you did not set "deployment.accessible_namespaces" to "**" (double-asterisk) which is what we recommend people use when running Istio  (as opposed to Maistra, which has soft-multi-tenancy in play, and why Kiali supports this accessible_namespace thing). Because Istio is cluster-wide, you want accessible_namespaces set to **.

See:

* https://github.com/kiali/kiali/blob/master/operator/deploy/kiali/kiali_cr.yaml#L134-L142
* https://www.kiali.io/documentation/getting-started/#_accessible_namespaces

(I just noticed the docs aren't that clear wrt using this with upstream Istio (as opposed to, say, Maistra) -- for sure, using accessible_namespaces of "**" is how most people should use this. Doing this, Kiali will essentially have the same roles as when installed via the Istio helm charts).

Note that if you installed the operator without specifying `-an "**"` then you need to re-install the operator because you need to give the operator the proper roles (so it can, in turn, assign the proper roles to Kiali).

e.g:

```
  bash <(curl -L https://git.io/getLatestKialiOperator) -an '**'
```

xeor:
Thanks! I'll try this when I get back.. :)

lucasponce:
Minor comment, also note that these messages:

```
W0917 19:26:27.545523       1 tls.go:56] GetMeshPolicies failed during a TLS validation. Probably Kiali doesn't have cluster permissions. Error: meshpolicies.authentication.istio.io is forbidden: User "syst
em:serviceaccount:istio-system:kiali-service-account" cannot list resource "meshpolicies" in API group "authentication.istio.io" at the cluster scope
W0917 19:26:42.525421       1 tls.go:56] GetMeshPolicies failed during a TLS validation. Probably Kiali doesn't have cluster permissions. Error: meshpolicies.authentication.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "meshpolicies" in API group "authentication.istio.io" at the cluster scope
.... a bunch more TLS validation errors
```
are warnings, not real erros, we decided to give visibility in the log about a query that can't be executed (reason is what we discussed, that Kiali ServiceAccount can't access to that object).

PR #1348, opened 2019-08-16
KIALI-3051 [operator] do a rolling upgrade rather than deleting the pod

jmazzitelli:
When we need to restart Kiali, limit the downtime that the user will see by using the trick of adding an annotation to the deployment's template metadata. When a change to an annotation is made,
k8s will do a rolling upgrade to the pods when updating the annotation, thus restarting the pod.

https://issues.jboss.org/browse/KIALI-3051

### DIFF
--- a/operator/roles/kiali-deploy/tasks/main.yml
+++ b/operator/roles/kiali-deploy/tasks/main.yml
@@ -412,7 +412,8 @@
   when:
   - '"**" not in kiali_vars.deployment.accessible_namespaces'
 
-- name: Delete Kiali deployment if image is changing since image is immutable
+# TODO: do we want to remove the label selector "version" so the Deployment only tracks by app=kiali label?
+- name: Delete Kiali deployment if image is changing since label selectors are immutable and we have version in a label
   k8s:
     state: absent
     api_version: apps/v1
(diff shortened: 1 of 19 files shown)


jmazzitelli:
@mwringe can I ask you for a review on this?  Since it was your idea, you can look at it to see if what I did here makes sense :) It works for me, FWIW.

operator/roles/kiali-deploy/tasks/main.yml
```diff
@@ -424,15 +425,44 @@
   - current_image_name is defined and current_image_version is defined
   - (current_image_name != kiali_vars.deployment.image_name) or (current_image_version != kiali_vars.deployment.image_version)
 
+# Get the deployment's custom annotation we set that tells us when we last updated the Deployment.
+# We need this to ensure the Deployment we update retains this same timestamp unless changes are made
+# that requires a pod restart - in which case we update this timestamp.
```
mwringe: I would have thought when we update the deployment we just set the annotation to force the update. Why do we need the previous value?

jmazzitelli: We need to get the previous value because if there IS NO previous value, we have to set it to the current time.

See line 451, "if the current_deployment_last_update" is not defined (meaning there is no deployment right now) then line 448 creates one with the current time and we set `deployment_is_new` to true. But without this attempt to get the current deployment, we have no idea if the deployment is new or if it already exists from a prior reconciliation run.

We need that `deployment_is_new` flag as well as part of the condition to know when we need to update the deployment so it does the rolling update. Line 511 - so if the deployment is new we don't do the task on line 504 ("- name: Force the Kiali pod to restart if necessary") regardless of the fact that the configmap changed because we already know, since its new, the pod is coming up fresh anyway.

mwringe:
@jmazzitelli Is there a molecule test we could add for this?

jmazzitelli:
> @jmazzitelli Is there a molecule test we could add for this?

I could add one yes. We can hold off on merging and I can put one together.

jmazzitelli:
molecule test showing nothing breaks : https://github.com/kiali/kiali/pull/1348/commits/8e984f1e7efb0e64459c14d109790c0145fc2392

jmazzitelli:
@mwringe molecule test added... this is ready for review/approval/merge

PR #1352, opened 2019-08-21
KIALI-3235 MeshPolicies queries shouldn't fail without cluster rights

lucasponce:
After we changed the MeshPolicies and ClusterRbacConfig queries.
If Kiali doesn't have cluster wide permissions, these queries can fail.

When Kiali doesn't have cluster rights, these queries will return empty.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -240,7 +240,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 		}
 	}(errChan)
 
-	go func(errChan chan error) {
+	go func() {
 		defer wg.Done()
 		// MeshPolicies are not namespaced. They will be only listed for an Istio namespace.
 		// Only listed in non Maistra environments.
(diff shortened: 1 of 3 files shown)


jshaughn:
What is the client-side impact?  Do we need to handle the fact that, say, global mTLS is unknown?

lucasponce:
> What is the client-side impact? Do we need to handle the fact that, say, global mTLS is unknown?

We don't have that state yet.
This situation was also known as to calculate MTLS it is needed a pair of MeshPolicy plus DestinationRule resources.

DestinationRule can live on a namespace that Kiali can't access for some reason and then Kiali would calculate the mTLS state with the info it has.

In this case, we are following the same pattern used for Maistra too, where Kiali can't access to the IstioNamespace.

@xeviknal and me have discussed about to introduce a new mTLS unknown state, but that would be handled in a next feature, as they have more implications.

For now, Kiali simply won't get those resources and it will calculate with the information it has.

Issue #1419, opened 2019-08-24
Link offending validations


burmanm:
In certain validations, such as the duplicate gateways or mismatch between mTLS rules and DestinationRule settings, we don't actually link the resources in any way. Instead, we only add a validation issue to both resources separately.

We should add to our validation model the link to the other resource as well, such as:

```go
References: models.IstioValidations
```

This should be relatively easy thing to do in the backend. On the front-end, we could further improve the UI validation text, so linking both to the documentation as well as items which are related to this validation issue and allow the user to click that one to go to the other offending ones. Or add a new tab which would list all the resources that are part of this validation issue (and allow quick jumping between them).

PR #1551, opened 2019-08-27
Add references to IstioValidation type and validators

burmanm:
** Describe the change **

As described in the issue #1419, certain resources are causing each other validation errors. This change adds links to those other objects which could have caused the validation issue to each other (since we don't know which one the user has incorrectly defined).

This PR includes only a sample for multimatch gateway validations, nothing else so far. Waiting for input first.

Also makes the logic in multi_match_checker a little more sane (instead of relying only on the duplicate removal).

** Issue reference **

Closes issue #1419 

** Backwards incompatible? **

Yes. If we want to take advantage of the feature in the UI, then UI will require changes.

### DIFF
--- a/business/checkers/checker.go
+++ b/business/checkers/checker.go
@@ -11,13 +11,13 @@ type GroupChecker interface {
 }
 
 // EmptyValidValidation returns a stub validation object which can be used by checkers
-func EmptyValidValidations(name, objectType string) models.IstioValidations {
-	key, emptyValidation := EmptyValidValidation(name, objectType)
+func EmptyValidValidations(name, namespace, objectType string) models.IstioValidations {
+	key, emptyValidation := EmptyValidValidation(name, namespace, objectType)
 	return models.IstioValidations{key: emptyValidation}
 }
 
-func EmptyValidValidation(name, objectType string) (models.IstioValidationKey, *models.IstioValidation) {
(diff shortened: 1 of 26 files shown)


lucasponce:
I've added the do not merge and UI frontend changes required.
But if this PR can be consolidated in backend without impact UI, then please, don't hesitate to update them.

Thanks !

burmanm:
Removal of labels meant that this should be ready for review and does not need frontend changes (it does not break current frontend)

hhovsepy:
As discussed in a chat, the namespace adding can cause problems, as it requires much more changes and is out of scope.

hhovsepy:
It seems "More than one virtual service for same host" validation is gone. I can not see that by the same steps before.

burmanm:
Can you list your steps? It worked for me (so might be something related to input material / key creation)

hhovsepy:
Screenshot from 2019-09-30 09-39-11
Screenshot from 2019-09-30 09-41-46

xeviknal:
@burmanm I know it is one more thing to take care of but it looks like you need to re-build the swagger files. If you want me to do it, I can take this PR from now on, no probs.
I'll do the review today or perhaps Monday, I want to make sure I understand all the changes.
Thanks for the patience on this one.

PR #1576, opened 2019-08-27
getting ready for operator phase 2

### LINKED ISSUE #1618: Investigate OLM and update mechanism
We should do some research into OLM and how this might help with our upgrade process.

We also need to figure out how this will affect upstream users of Istio running on k8s.

JIRA:
https://issues.jboss.org/browse/KIALI-3080

jmazzitelli:
fixes #1618 

This will be work to see the operator work with OLM as a "phase 2" operator (meaning seamless upgrades).

Phases are part of the maturity model as described [here](https://github.com/operator-framework/operatorhub-docs/blob/master/building-operators/building-operators/types-of-operators.md#operator-maturity-model) and [here](https://github.com/operator-framework/operator-sdk/blob/master/doc/images/operator-capability-level.png)

### DIFF
--- a/operator/deploy/deploy-kiali-operator.sh
+++ b/operator/deploy/deploy-kiali-operator.sh
@@ -157,7 +157,9 @@
 #
 # KIALI_IMAGE_NAME
 #    Determines which image of Kiali to download and install.
-#    Default: "kiali/kiali"
+#    If you set this, you must make sure that image is supported by the operator.
+#    If left empty (the default), the operator will use a known supported image.
+#    Default: ""
 #
 # KIALI_IMAGE_PULL_POLICY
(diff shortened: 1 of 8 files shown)


jmazzitelli:
This PR https://github.com/operator-framework/operator-lifecycle-manager/pull/988 will introduce docs on operator upgrades.. will be useful as reference.

jmazzitelli:
While this hasn't been putting the operator through vigorous testing to ensure it supports phase-2-like functionality, it is preliminary work that has to get done anyway; and, in fact, this does appear to at least perform a basic upgrade of a previous installation (it doesn't fail and the new kiali starts).

So I want to merge this to get this work in and continue on with additional work to be done in a followup PR.

PR #1628, opened 2019-08-30
Fix InClusterURL for jaeger

aljesusg:
Fix #1626

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -257,6 +257,7 @@ spec:
 # namespace: The Kubernetes namespace that holds the Tracing service (if empty, assumes the same value as deployment.namespace)
 # service: The Kubernetes service name for tracing. Kiali uses this to connect within the cluster to Jaeger.
 # port: The kubernetes port where the tracing service is mount. Default port is Jaeger port, 16686. Autodiscovered when service is not specified.
+# in_cluster_url: Set URL for in-cluster access. Example: "http://jaeger-query.istio-system:16686".
 # url: The URL that Kiali uses when integrating with Tracing. This URL must be accessible to clients external to
 #      the cluster in order for the integration to work properly. If empty, an attempt to auto-discover it is made.
 #      If tracing service is deployed in a QUERY_BASE_PATH set this in the url like https://<hostname>/<QUERY_BASE_PATH> . EX: https://tracing-service:8080/jaeger
@@ -275,6 +276,7 @@ spec:
 #      service : ""
 #      port: 16686
 #      url: ""
+#      in_cluster_url: ""
 
 ##########
 #  ---
--- a/operator/roles/kiali-deploy/defaults/main.yml
+++ b/operator/roles/kiali-deploy/defaults/main.yml
@@ -80,6 +80,7 @@ kiali_defaults:
       namespace: ""
       service: ""
       port: 16686
+      in_cluster_url: ""
       url: ""
 
   identity: {}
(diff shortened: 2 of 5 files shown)


operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -257,6 +257,7 @@ spec:
 # namespace: The Kubernetes namespace that holds the Tracing service (if empty, assumes the same value as deployment.namespace)
 # service: The Kubernetes service name for tracing. Kiali uses this to connect within the cluster to Jaeger.
 # port: The kubernetes port where the tracing service is mount. Default port is Jaeger port, 16686. Autodiscovered when service is not specified.
+# in_cluster_url: Set URL for in-cluster access. Example: "http://jaeger-query.istio-system:16686".
```
lucasponce: As default value is empty. Perhaps worth to indicate if when empty what happens (i.e. it defaults to autodiscovery, or service would be disabled, etc.
Perhaps this comment can apply to other fields too.

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -80,6 +80,7 @@ kiali_defaults:
       namespace: ""
       service: ""
       port: 16686
+      in_cluster_url: ""
```
lucasponce: Also, I imagine that a change in the operator would trigger a change in the "users" of the operators.
In particular, if I'm not wrong, the operator populates these data from maistra data (or it's the maistra operator who creates the kiali CR).
So probably this will need to be propagated in order to be working.

operator/deploy/kiali/kiali_cr.yaml
jmazzitelli: please put all these new values in alphabetical order. We need to keep these in order to help diff future versions and to just make things clean and orderly to easily find things. Move those "in_cluster_url" above "insecure_skip_verify"

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -275,6 +276,7 @@ spec:
 #      service : ""
 #      port: 16686
 #      url: ""
+#      in_cluster_url: ""
```
jmazzitelli: put in alphabetical order - put it above insecure_skip_verify

operator/roles/kiali-deploy/defaults/main.yml
jmazzitelli: put in alphabetical order - put it above insecure_skip_verify

jmazzitelli: > or it's the maistra operator who creates the kiali CR

Yes, this is a good point. See this template that Maistra uses when creating the Kiali CR:
 
https://github.com/Maistra/istio-operator/blob/014f116a7e2291cf7e89c4612700724a636a5ccd/helm/istio/charts/kiali/templates/kiali-cr.yaml#L65-L79

jmazzitelli: Note also this Go code in the Maistra operator that attempts to auto-detect Jaeger and re-configured the Kiali CR:

https://github.com/Maistra/istio-operator/blob/014f116a7e2291cf7e89c4612700724a636a5ccd/helm/istio/charts/kiali/templates/kiali-cr.yaml#L65-L79

aljesusg: PR sent https://github.com/Maistra/istio-operator/pull/244

aljesusg:
> see previous comments - we just need to keep these settings in alphabetical order.

Done sorry :(

hhovsepy:
Verified:
Screenshot from 2019-08-30 15-46-18

lucasponce:
I've merged this one, but keep in mind we can't release a 1.0.6 if this one Maistra/istio-operator#244 is addressed.

Issue #1605, opened 2019-08-28
If using https endpoint, redirect if http is accidentally entered


mwringe:
**Is your feature request related to a problem? Please describe.**

Its been mentioned that it's easy when trying to teach and demo Kiali for users to accidentally use the http endpoint instead of the https one.

Currently if you are using the http endpoint instead of the https one you will get an error message and its easy for users to think that Kiali is down or not installed properly.

**Describe the solution you'd like**

If possible, when a user access the http endpoint instead of the https endpoint that we redirect them to the https endpoint instead.

I am not sure if this is possible with the type of route we are using in OpenShift, or if its possible with the ingress we are using on k8s. But we should at least investigate and close if there is no option.

PR #1630, opened 2019-08-30
KIALI#1605 Update route to redirect http to https

mwringe:
**Describe the change**

Updates the route so that if a use enters the url using 'http' it will redirect them to the 'https' endpoint instead.

closes https://github.com/kiali/kiali/issues/1605


jmazzitelli:
shoot - I thought we merged this. This actually would have fixed the upstream Istio issue they hit during the 1.3 release testing ... we should still merge this.

PR #1633, opened 2019-08-30
Re-label telemetry destination_service={*.global} to hostname only

### LINKED ISSUE #1481: Multi-cluster: Adequately identify service nodes when being called remotely (multiple control plane)
Given a multi-cluster setup and a bookinfo example setup. Assume that a cluster is calling the details and reviews services of a remote cluster.

In the "remote cluster", Kiali is generating the following graph:

image

Kiali manages to create a graph with incoming traffic from the source cluster. However, there is an issue: the _details.bookinfo.global service node_ is a duplicate of the bare _details service node_.

The second issue (although I'm not sure of this one) is that the _unknown_ node, in this case, could be changed to a node representing incoming traffic from other clusters. This could be done by looking at the _global_ suffix in the telemetry.

Fixing these two issues the graph could better represent traffic coming from other clusters in the mesh.

israel-hdez:
In a multi-cluster scenario, in the cluster that is the destination of
the traffic, Istio will record telemetry with
destination_service="svc.namespace.global". This was generating a graph
with dummy service nodes labeled with "svc.namespace.global". These
represent the same "svc" kubernetes services without any suffix.

This changes perform some renaming to generate a traffic map that will
lead to proper unified service nodes.

Resolves #1481.

BEFORE:

image

AFTER:

image

### DIFF
--- a/graph/telemetry/istio/istio.go
+++ b/graph/telemetry/istio/istio.go
@@ -387,24 +387,7 @@ func populateTrafficMap(trafficMap graph.TrafficMap, vector *model.Vector, o gra
 		flags := string(lFlags)
 
 		val := float64(s.Value)
-
-		destSvcNameEntries := strings.Split(destSvcName, ".")
-		if sourceWlNs == graph.Unknown && sourceWl == graph.Unknown &&
-			len(destSvcNameEntries) == 3 && destSvcNameEntries[2] == config.IstioMultiClusterHostSuffix {
-			// For multi-cluster setup, if "we are" the DESTINATION cluster of traffic,
-			// the destination_service_name label will contain the
-			// hostname that the source cluster requested. Since we know that the requested hostname
-			// has the form "svc_name.ns_name.global", we can truncate the destSvcName to the
-			// prefix to correctly unify svc nodes. This way, the graph will show only one "svc_name" node
-			// instead of duplicating and having one "svc_name" and one "svc_name.ns_name.global" node which,
-			// in practice, are same.
-			//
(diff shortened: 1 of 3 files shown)


israel-hdez:
I'm not doing the part of adding a "remote-cluster" node instead of using the "unknown" node. I'm not sure if that change would be worth it.

I'm doing only minimal changes to properly support the multi-cluster with multiple-control-plane use case. If you think it's better to differentiate "unknown" vs "remote-cluster" I can do it.

graph/telemetry/istio/istio.go
```diff
@@ -388,6 +388,24 @@ func populateTrafficMap(trafficMap graph.TrafficMap, vector *model.Vector, o gra
...
+			// instead of duplicating and having one "svc_name" and one "svc_name.ns_name.global" node which,
+			// in practice, are same.
+			//
+			// This is done only if source workload is "unknown" which is what is recorded in telemetry
+			// when traffic is coming from a remote cluster. If source workload IS known,
+			// then, traffic is most-likely going to a ServiceEntry and being routed out of the cluster (i.e.
+			// the "source cluster" case). That case is handled in the service_entry.go file.
+			destSvcName = destSvcNameEntries[0]
+		}
+
```
jshaughn: Let's extract these two large, duplicated code blocks into a helper function so that we can just have something like:

`  destSvcName = handleMultiClusterServiceName(sourceWlNs, sourceWl, destSvcName)`

Minor: don't do the split prior to checking the for the unknown preconditions.

israel-hdez: Change done.
I also rebased the PR.

PR #1636, opened 2019-09-03
Update CRD count in helm installer

jshaughn:
- 1.3.0 RC1 has 24, up from 23 in RC0

### DIFF
--- a/hack/istio/install-istio-kiali-via-helm.sh
+++ b/hack/istio/install-istio-kiali-via-helm.sh
@@ -358,7 +358,7 @@ else
   ${HELM_EXE} template "${ISTIO_DIR}/install/kubernetes/helm/istio-init" --name istio-init --namespace ${NAMESPACE} | ${CLIENT_EXE} apply -f -
   _crd_count="0"
   echo -n "Waiting for the CRDs to be created"
-  while [ "$_crd_count" -lt "23" ]; do
+  while [ "$_crd_count" -lt "24" ]; do
     sleep 1
     echo -n "."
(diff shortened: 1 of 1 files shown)


hack/istio/install-istio-kiali-via-helm.sh
```diff
@@ -358,7 +358,7 @@ else
   ${HELM_EXE} template "${ISTIO_DIR}/install/kubernetes/helm/istio-init" --name istio-init --namespace ${NAMESPACE} | ${CLIENT_EXE} apply -f -
   _crd_count="0"
   echo -n "Waiting for the CRDs to be created"
-  while [ "$_crd_count" -lt "23" ]; do
+  while [ "$_crd_count" -lt "24" ]; do
```
mtho11: Is this really a good way of doing this? Because:

1. It changes over time and versions
2. If cert-manager is enabled it is different (like 28 instead)

Seems like a pretty fickle solution.

jmazzitelli: its just our little hack script - so not end-user-facing... so at the end of the day "it works" is good enough.

It's also what the Istio docs say to do: https://istio.io/docs/setup/kubernetes/install/helm/#option-1-install-with-helm-via-helm-template

```
Verify that all 23 Istio CRDs were committed to the Kubernetes api-server using the following command:

If cert-manager is enabled, then the CRD count will be 28 instead.

"$ kubectl get crds | grep 'istio.io\|certmanager.k8s.io' | wc -l"
```

mtho11: Fine then 🍅 🚽 👅

abonas: @mtho11 translation please ;)

PR #1658, opened 2019-09-10
make secret optional when http is configured

### LINKED ISSUE #1657: if https not to be used, don't expect a istio secret to be mounted
If identity.cert_file is empty, it means kiali should be exposed over http (not https).

In this case, our kubernetes deployment should not expect to see an istio cert mounted, it should be optional. So this:

https://github.com/kiali/kiali/blob/v1.4/operator/roles/kiali-deploy/templates/kubernetes/deployment.yaml#L82-L84

```
      - name: kiali-cert
        secret:
          secretName: istio.kiali-service-account
```

Should have "optional: true" in that case.

jmazzitelli:
fixes #1657

### DIFF
--- a/operator/roles/kiali-deploy/templates/kubernetes/deployment.yaml
+++ b/operator/roles/kiali-deploy/templates/kubernetes/deployment.yaml
@@ -82,6 +82,9 @@ spec:
       - name: kiali-cert
         secret:
           secretName: istio.kiali-service-account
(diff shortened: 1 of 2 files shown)


jmazzitelli:
This came up in conversation with the Istio folks. If Istio is not generating certificates, we can't out of box deploy with https so we need to make this certificate optional when deployed with http.

Issue #1662, opened 2019-09-11
Session expired dialog - Logout fails


hhovsepy:
When Kiali session is expired, it shows a dialog box, which has an option to Log Out.
But in fact Logout fails with a loop of error notifications:

![Screenshot from 2019-09-11 09-45-50](https://user-images.githubusercontent.com/604313/64678148-48dbb380-d479-11e9-987e-ba5038935fdd.png)

lucasponce:
In which Kiali version this problem happens ?

hhovsepy:
Kiali-ui 1.5.0 (79f36a929f854e3d20fa54ad73b0f4b30d6c7594)
Kiali v1.5.0-SNAPSHOT (20300260cce93481c9f28f4a75d0dd69875f6ecb)

jotak:
In a first attempt, I didn't reproduce. I've set expiration to 60 seconds, and it works as expected on logout.
@hhovsepy can you tell what is the whole error message? (it's truncated on the screenshot)
Also, is there any more information in backend logs?

jotak:
Also, which authentication strategy are you using? Openshift? Basic auth?

hhovsepy:
It is using OAuth. The error message is "Token is invalid: token contains an invalid number of segments", the same in UI ans logs.

jotak:
ok, i was trying with basic auth, that certainly explains the difference

lucasponce:
@hhovsepy just to confirm, this issue has been fixed in master, but it affect to v1.0.5 ?
Has @jotak to backport the PR merged into the v1.0 branch ?

I put the label but not sure if I misunderstood it.

hhovsepy:
@lucasponce yes it should be backported to v1.0 branch

jotak:
@lucasponce @hhovsepy while cherry-picking the patch & resolving conflicts I'm seeing that the bug was certainly not present in v1.0 branch, because we were silently ignoring the error on logout hence it wouldn't have triggered what Hayk saw on the frontend.

See incriminated code, just before my patch: https://github.com/kiali/kiali/blob/2a287358e12d74968fa3ee1b3f973b3bb3b5b072/handlers/authentication.go#L493-L499

And in v1: https://github.com/kiali/kiali/blob/828e31e1cfec6424b8844056689965fc25777f29/handlers/authentication.go#L422-L426 
=> no InternalServerError returned in case of failure in v1

So, I think it's not necessary to backport the patch.

PR #1664, opened 2019-09-11
Do not return error 500 on missing cookie during logout

jotak:
Return 401 instead, so that Kiali UI will get back to the login screen
Also, fixes an issue with response being written twice (`RespondWithError` call followed by `RespondWithCode`)

Closes https://github.com/kiali/kiali/issues/1662

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"strconv"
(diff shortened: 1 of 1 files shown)


jotak:
Note for testers: the issue described in #1662 happens on "openshift" login strategy.

To reduce the expiration duration, on openshift 3.x you need to edit your `master-config.yml` file (I've 3 of them in my openshift data directory, not sure which one matters, I edited all of them) and change `accessTokenMaxAgeSeconds` to something smaller.

In openshift 4 I haven't tried but it seems there's a `OAuth` resource where it can be managed: https://docs.openshift.com/container-platform/4.1/authentication/configuring-internal-oauth.html#oauth-configuring-internal-oauth_configuring-internal-oauth

hhovsepy:
@jotak is it expected to auto-logout when session has timeout? This is what I saw on my env.

jotak:
@hhovsepy you mean at the end of the timer countdown? I think it is. WDYT @lucasponce @aljesusg  ?

lucasponce:
> is it expected to auto-logout when session has timeout?

That sounds correct, unless I missed something.
Or not sure if "auto", but if on a request the token/session is not valid, it should logout (I guess it gets the "auto" behaviour when it's polling health).

PR #1673, opened 2019-09-14
[hack] OpenShift on AWS hack script

jmazzitelli:


jmazzitelli:
this is ready for merge. It deployed a cluster in AWS and optionally installs Service Mesh with jaeger and kiali (which is the default behavior - pass in `-ie false` to just get OS cluster without service mesh - you would have to install istio yourself in this case).

gbaufake:
Should we include the OLM option here?

jmazzitelli:
> Should we include the OLM option here?

What do you mean "OLM option"?

This hack script goes through OLM to install the latest Service Mesh - it installs the OLM-required resources (a CatalogSourceConfig that includes all service-mesh related operators (which, in the future, we won't need because service mesh operator will define its dependencies) as well as all the Subscriptiosn required.).

It will then install the SMCP and SMMR to install service mesh in a control plane named "istio-system" and includes "bookinfo" in the member roll.

aljesusg:
Worked for me

gbaufake:
@jmazzitelli 
My apologies, I saw the YAML installation but I didn't see the Subcription and Catalog Source just up on the hack script.

It worked for me as well.

jmazzitelli:
OK, so it worked for two other people, can someone approve this then so I can merge it :)

PR #1683, opened 2019-09-17
WIP - Batch namespace health operations in a single call

### LINKED ISSUE #1637: Get all health data for the graph on only one request
Part of kiali/kiali#1623.

Every time that we do refresh on the graph, we do at least 5 HTTP requests, one for metrics, one for the nodes, and 3 for Health.

Those requests are done in parallel, but each take 40ms to reply (locally, OCP 3.11 with hack scripts), and joining them in just one request will make things faster. I did a small test that is really hard to show, but removing all 3 requests made the graph refresh close to 16ms faster, which is one "frame"/unit of time that you can wait.

In general, speed updates in JS run speed can only be done in increments of 16ms (or less if they do decide to stop forcing 60hz on the engine), so most small changes are not worth it.

Here's an example of the requests:

josejulio:
** Describe the change **

Batches the health operations in a single call, to avoid multiple calls for each health type (workload, service, app)

 @cfcosta is this what you had in mind?

This PoC uses GET, but we should probably move to POST to avoid hitting limits on the URL.

I was running some tests, and there are pro's and con's about this approach.

All the tests were run on my local minikube cluster, so maybe someone else should take a look

For a single namespace, there wasn't any noticeable speed gain, it was about the same time, i suppose it has something to do with the persistent connections (`connection: keep-alive`)
For 3 namespaces, times was about the same, but there was an increase in the download size, as a single result gets compressed and tiny separate results do not.
The downside of this is that if there is a namespace with 1 workload; the result reach the UI faster without the batch, as the batch has to wait for every other health result to be finished.

Perhaps a web socket would be a better fit for this kind of optimizations?

Closes #1637

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -1,6 +1,7 @@
 package business
 
 import (
+	"sync"
 	"time"
 
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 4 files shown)


josejulio:
I'm going to close this as there is no real gain by using only 1 call, the gains are minimal and could potentially complicate the code even more with small benefits.

jotak:
Thanks @josejulio for trying this

cfcosta:
Thanks for trying this!

Issue #1405, opened 2019-08-23
List hostnames of rejected external connections (and make it easy to accept them)	


lucasponce:
In an Istio mesh, external connections are rejected unless a serviceentry exists that explicitly grants access.

Right now in Kiali, if a workload tries to connect to an external host we show a red 'unknown' triangle. This isn't very helpful (and is likely confusing to users).

"As a Kiali user, I would like to see what host was attempted to be accessed and provided with an option to 'allow access' (eg create a serviceentry that would grant access). Knowing that some host was denied access is not that helpful"

Note: by default the metrics that we have do not provide this information. But Istio can be configured to start collecting the request.host which does give this information (https://istio.io/docs/reference/config/policy-and-telemetry/attribute-vocabulary/)

We might also want to discuss with UXD on the best way to visualise this in the mesh. Perhaps we need a new shape or something else to represent rejected connections.

JIRA:
https://issues.jboss.org/browse/KIALI-2167

jshaughn:
We need to revisit the behavior with Istio 1.3 and changes with telemetry reporting for the PassthroughCluster,

jshaughn:
> In an Istio mesh, external connections are rejected unless a serviceentry exists that explicitly grants access.

This is no longer true by default.  By default Istio sets `global.outboundTrafficPolicy.mode` to `ALLOW_ANY` for external requests.  In such case the request is allowed to exit the mesh, via the `PassthroughCluster`, which acts as sort of a default ServiceEntry. ` PassthroughCluster` is not an actual Service or ServiceEntry, but shows up in the telemetry.    When configured to `REGISTRY_ONLY` requests not routed through a ServiceEntry will be blocked.

Either way the actual requested host is not available unless, as the description points out, request.host is configured as an additional telemetry field.  The question is whether or not we want to try and collect these hosts (and possibly success/fail rates for each) if and when that additional information is present.

If so, I'd probably recommend that we just display the hosts in the side panel of the `PassthroughCluster `service (perhaps PC should be shown as a service entry?).  Also, we need to test in istio 1.3 the failed telemetry resulting from `REGISTRY_ONLY` and no matching ServiceEntry.  The hope would be that it shows a failed request for `PassthroughCluster`.

jshaughn:
> Also, we need to test in istio 1.3 the failed telemetry resulting from REGISTRY_ONLY and no matching ServiceEntry. The hope would be that it shows a failed request for PassthroughCluster.

[...]

PR #1687, opened 2019-09-18
Capture and report request.host information when it is configured

jshaughn:
Users have no way to easily see the requested host for their requests. This
can be very useful for failed requests, especially external requests, but is
also has informational value in other cases.

The request.host info must (currently 1.3) be configured manually for Istio, but
if done this PR will capture the information and report it back as part of
the "responses" data, along with rates and response_flags.

UI PR: https://github.com/kiali/kiali-ui/pull/1430
Website PR: https://github.com/kiali/kiali.io/pull/167

Test Instructions:
For this feature to fully work the user needs to do additional configuration of Istio telemetry, adding a field to the istio_requests_total metric.  To begin, test without the customization to ensure things work in the out-of-box scenario.

1. Install the Server and UI PRs.
2. Generate a graph, click an edge, and look at the summary panel.
3. There are now 3 tabs, check the tooltips for the non-default tabs.
4. Click the new Hosts tab.
4.1 It should say something like "Not configured" and direct you to the FAQ
4.2 Install the website PR or just look at the text in the new FAQ file
4.3 Customize the telemetry as described in the FAQ
5. Re-test with the updated telemetry, the Hosts information should now show up.

https://github.com/kiali/kiali/issues/1405


jshaughn:
> I don't like the idea of Kiali now asking users to add new fields to telemetry (even though optional). Adding dimensions increases storage on prometheus and could have performance implications. Also, we are now supporting non-standard Istio telemetry. What happens if in the future Istio adds a field that does this same thing but doesn't match "our" schema. We now have to support both.

We're not asking them to add a new field, this is adding support for users that need the request.host information and are willing to add it to the telemetry. IMO, this field will not have detrimental impact on perf, there is typically a 1:1 relationship between request.host and destination_service.  In the scope of all the telemetry generated, this should be nominal.  As for them adding a field for this and naming it something other than "request_host", it's possibly but I'd say it would be a 99% chance that request.host would map to a request_host label.

> I would rather we push this up to the Istio community, have them add this field (or something like it) and THEN support it.

We could wait for Istio 1.3.1, or possibly for the 1.4 designs to be complete.  But this could be very useful to users of Istio <= 1.3.0.  It solves two existing Kiali community-reported issues.

jshaughn:
> I don't like the idea of Kiali now asking users to add new fields to telemetry (even though optional). 

@jmazzitelli , I held this PR to see what they may do in 1.3.1 and in fact they've changed the telemetry for egress traffic such that destination_service now gets the request.host value.  For non-egress it is set to destination.service.host, which is maybe not as useful, but still maybe good to see.  So, now this PR will collect and display this info for all edges.  Backend processing increases, as does the JSON size, but it should be nominal, and the egress info is very valuable.

Bottom line, no longer asking users to add special telemetry, nor do we do anything with it if they do.

jshaughn:
@josejulio , @jmazzitelli , this PR (and the UI PR) have now been updated given the changes in Istio 1.3.1.  The request.host telemetry addition is no longer required.  Istio 1.3.1 is not required, although without it you won't get the useful egress information.  Please review.

Issue #1694, opened 2019-09-19
Investigate Istio mixer-less telemetry (experimental in 1.3.0)


jshaughn:
Istio 1.3.0 provides an experimental mixer-less telemetry option.  It can be configured using 1.3.0 with the following instructions:

https://github.com/istio/istio/wiki/Mixerless-HTTP-Telemetry

Related issues or Jiras should be linked here.

jshaughn:
I configured mixer-less and immediately saw some issues in the graph.  This Istio issue deals with the lack of parity between mixer and mixer-less values in istio_requests_total.  Confirmed as an issue by Istio:

https://github.com/istio/istio/issues/17231

jshaughn:
Mixer-less has changed the metric for latency from istio_request_duration_seconds to istio_request_duration_milliseconds.   For backward compatibility we'll need to change our queries to support both.  I expect  istio_request_duration_seconds will likely be deprecated/removed at some point at which point we can remove support for the old metric.

I'll work on a PR for this change.

jshaughn:
I'm not seeing any more work for this at the moment.  The reported bug is targeted for 1.4, at which point we can test this again.

PR #1711, opened 2019-09-23
Support request_duration_milliseconds for latency reporting in mixer-less telemetry

jshaughn:
Istio 1.3.0 experimental introduces mixer-less telemetry.  In this new approach latency is reported via
a new metric: request_duration_milliseconds, as opposed to what has has been done with request_duration (in seconds).

I'm not aware of a way to drive off config to know which will be in service for a time period, so until the legacy metric is removed we need to support both.  This makes for some sort-of-hacky code. 
 In the graph's response time appender we solved it mainly in the prom queries (this is the way Istio itself is doing it).  But the metrics code is table-driven/generic and so it's handled in a different way, based on determining which metric actually returns valid data.

See relevant Istio issue here: https://github.com/istio/istio/issues/16799.

I'm not really asking reviewers to run the code with mixer-less, rather to ensure no regression using the existing telemetry.  This mainly means that the detail charts should work as expected, as should the graph side panel sparklines for edges, and of course the response time edge labels.

@jotak If you could take a look at the metrics part it would be appreciated.

https://github.com/kiali/kiali/issues/1694

UI PR: https://github.com/kiali/kiali-ui/pull/1436

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -1,6 +1,8 @@
 package business
 
 import (
+	"math"
+
 	kbus "github.com/kiali/k-charted/business"
 	kconf "github.com/kiali/k-charted/config"
 	kxconf "github.com/kiali/k-charted/config/extconfig"
@@ -95,6 +97,8 @@ var istioCharts = []istioChart{
 		},
 		refName: "request_count",
 	},
+	// TODO: Istio is transitioning from duration in seconds to duration in ms (a new metric). When
+	//       complete we should reduce the next two entries to just one entry.
 	{
 		Chart: kmodel.Chart{
 			Name:  "Request duration",
@@ -102,6 +106,14 @@ var istioCharts = []istioChart{
 			Spans: 6,
 		},
 		refName: "request_duration",
+	},
+	{
+		Chart: kmodel.Chart{
+			Name:  "Request duration",
+			Unit:  "ms",
+			Spans: 6,
+		},
+		refName: "request_duration_millis",
 	},
 	{
 		Chart: kmodel.Chart{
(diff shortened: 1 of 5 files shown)


graph/telemetry/istio/appender/response_time.go
```diff
@@ -70,15 +71,25 @@ func (a ResponseTimeAppender) appendGraph(trafficMap graph.TrafficMap, namespace
...
+		int(duration.Seconds()), // range duration for the query
+		groupBy)
+	secondsQuery := fmt.Sprintf(`histogram_quantile(%.2f, sum(rate(%s{reporter="destination",source_workload="unknown",destination_workload_namespace="%v",response_code=~"%s"}[%vs])) by (%s))`,
 		quantile,
 		"istio_request_duration_seconds_bucket",
 		namespace,
 		"2[0-9]{2}|^0$",         // must match success for all expected protocols
 		int(duration.Seconds()), // range duration for the query
 		groupBy)
+	query := fmt.Sprintf(`((%s > 0) OR ((%s > 0) * 1000.0))`, millisQuery, secondsQuery)
```
josejulio: Is the query OK? I don't know much of it, is `1000` multiplying the result of `%s > 0` ?
 `(%s > 0) * 1000.0)`

business/dashboards.go
```diff
@@ -103,6 +107,14 @@ var istioCharts = []istioChart{
 		},
 		refName: "request_duration",
 	},
+	{
+		Chart: kmodel.Chart{
+			Name:  "Request duration",
+			Unit:  "ms",
```
jotak: I believe this unit is not going to scale properly when displayed on charts if it has big or low values (e.g. 1000 ms might be displayed as "1 kms" for "kilo-millisecond" :man_facepalming: 

Have you seen something like that? (I haven't enabled telemetry lite so I can't confirm right now)

This is not something you can fix in this PR, I would need to do something in k-charted.

jotak:
Tested non-regression on old metric + reviewed dashboards code => LGTM but we probably need to open a bug in k-charted to better handle unit scaling

graph/telemetry/istio/appender/response_time.go
jshaughn: yes, it's tricky but it works.  in ((%s > 0) * 1000.0)) the %s is the main query that returns the quantile value in seconds, then we throw out any time-series where value is 0, and then we multiply by 1000 to turn seconds into milliseconds, so the values from both sides of the OR return in ms.

PR #1722, opened 2019-09-24
Add TCP security badging 

### LINKED ISSUE #1709: TCP security badging not working
This was originally reported in https://github.com/istio/istio/issues/17110.

The TCP security policy is not being handled server-side.  The UI-side is likely OK.

jshaughn:
Add TCP support to to the graph security_policy appender.

Fixes #1709


jshaughn:
I wasn't able to generate the actual time-series but was able to simulate it.  So, assuming Istio is reporting the mtls it should now show up in Kiali.

image

israel-hdez:
Hmm... So, timeseries didn't get generated. I'm wondering if Istio supports mTLS for TCP connections.
Although TLS can happen over TCP, I'm unsure that Istio can actually "wrap" TCP data over TLS. In tutorials and examples I have seen only mTLS over HTTP traffic, and that may mean that Istio may not cover the TCP case for mTLS. It's probably worth to ask.

jshaughn:
@israel-hdez See https://github.com/istio/istio/issues/17110#issuecomment-533745960

PR #1736, opened 2019-09-26
Fix wrong k-charted version in glide.yaml

jotak:
It seems like the k-charted version was mistakenly screwed up in commit 4db2a6cf, erasing my previous change
It doesn't affect the vendors directory (probably was just a merge conflict on a previous rebase) so has no impact in code until someone runs `make dep-update` again.


jotak:
I'm merging  as there's no code impact anyway

Issue #1738, opened 2019-09-27
Fix unit scaling for mixer-less telemetry


jotak:
This is a follow-up of https://github.com/kiali/kiali/pull/1711 and https://github.com/kiali/kiali/issues/1715

- https://github.com/kiali/kiali/pull/1711 introduced metric with mixer-less telemetry
- https://github.com/kiali/kiali/issues/1715 fixed k-charted to manage scaled unit in prometheus (used in #1711)

Now we need to update Kiali to use this pre-scaling mechanism.

PR #1739, opened 2019-09-27
Fix unit scaling for mixer-less telemetry

jotak:
- Change unit of request_duration_millis from "ms" to "seconds" with scaling 0.001, so that the scaling mechanism on time units is preserved
- Fix an issue with k-charted loggers (was all set to Warning)
- Update version of k-charted

Closes https://github.com/kiali/kiali/issues/1738

PR required: https://github.com/kiali/k-charted/pull/56

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -78,14 +78,15 @@ func DashboardsConfig() (kconf.Config, klog.LogAdapter) {
 		}, klog.LogAdapter{
 			Errorf:   log.Errorf,
 			Warningf: log.Warningf,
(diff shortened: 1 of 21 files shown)


jotak:
@jshaughn I haven't tested yet with the mixer-less telemetry as I don't have it configured. Will you be able to test it, if you still have this setup? Essentially to test that the units are still viewed in milliseconds with the change to "seconds" + scaling 0.001

jotak:
thanks @jshaughn @hhovsepy

Issue #1139, opened 2019-06-13
Make readonly & admin access possible together


hdimitriou:
I have read that you are working on supporting RBAC ( https://github.com/kiali/kiali/issues/671#issuecomment-440564792 ) which is what I really want.

However, until that is in place, a workaround would be to allow the typical admin user, but also allow a readonly user as well. 
This way, we can let people use a readonly account to view stuff in Kiali and provide only a few people with the admin account. 

Right now, I see you can turn Kiali to readonly globally, which is not what I want because we need some people to have admin access

israel-hdez:
@hdimitriou If you are using an OpenShift based cluster, you can have this possibility by enabling authentication through cluster credentials.

If you are using Kubernetes or a cluster flavor similar to it, then, as you say, it's not possible yet. I guess that the very minimum would be that Kiali supports more users that just an "admin" one.

For RBAC, as pointed in the comment you are referring, we are trying to delegate to the cluster RBAC. Here, the goal is to login into Kiali turhoug the cluster auth mechanisms. I'm curious to know if you would use any of the authentication mechanisms described here: https://kubernetes.io/docs/reference/access-authn-authz/authentication/ Your feedback about this help us to prioritize work.

hdimitriou:
> I'm curious to know if you would use any of the authentication mechanisms described here: https://kubernetes.io/docs/reference/access-authn-authz/authentication/ 
I have gone through those, I would be more keen on service account token for groups of users. 
At the moment I am reluctant of giving non admin person access to people. 

Let's say that in terms of priority it makes more sense to be able to give groups of users some specific access levels. If you wanted you could even make personal service account tokens.

israel-hdez:
Hi @hdimitriou 
I'm working into making it easier to integrate other auth mechanisms into Kiali, so that people can collaborate and include their wanted mechanisms.

As part of that, I may impement a "Token login", similar to the [login mechanism of the kubernetes dashboard](https://github.com/kubernetes/dashboard/wiki/Access-control#login-view) which seems to be "compatible" with current Kiali's RBAC Openshift mechanisms, thus probably making kubernetes RBAC support doable with small changes to Kiali. I hope this may cover your needs.

I'll update you with that progress.

lucasponce:
@israel-hdez any update on this issue ?

israel-hdez:
@lucasponce My plan is to do some changes to the auth mechanism.
[...]

PR #1742, opened 2019-09-30
Supporting setting more than one user credential at once (for web login)

cfcosta:
Part of #1139.

This PR enables the users to create more than one weblogin user, by using an array instead of an object for the user. It does not break backwards compatibility.

To create multiple users, change the secret as such:

```diff
--- before.yml  2019-09-30 11:12:22.214379071 -0300
+++ after.yml   2019-09-30 11:12:37.007582588 -0300
@@ -1,7 +1,9 @@
 apiVersion: v1
 data:
-  passphrase: YWRtaW4=
-  username: YWRtaW4=
+  - passphrase: YWRtaW4=
+    username: YWRtaW4=
+  - passphrase: YWRtaW4=
+    username: YWRtaW4=
 kind: Secret
 metadata:
   creationTimestamp: "2019-09-30T14:11:45Z"
```

Obviously all the users should have different passwords ;)

### How to test?

- Install kiali as expected.
- Create a secret with more than one user.
- Log in with both users.
- Both should work.

### DIFF
--- a/kiali.go
+++ b/kiali.go
@@ -155,9 +155,10 @@ func waitForSecret() {
 		}
 	}()
 	secret := <-foundSecretChan
+
 	cfg := config.Get()
-	cfg.Server.Credentials.Username = secret.Username
-	cfg.Server.Credentials.Passphrase = secret.Passphrase
+	cfg.Server.Credentials = security.CredentialList{security.Credentials{Username: secret.Username, Passphrase: secret.Passphrase}}
+
 	config.Set(cfg)
 }
 
@@ -183,7 +184,7 @@ func validateConfig() error {
 		return fmt.Errorf("server port is negative: %v", config.Get().Server.Port)
 	}
 
-	if err := config.Get().Server.Credentials.ValidateCredentials(); err != nil {
+	if _, err := config.Get().Server.Credentials.ValidateCredentials(); err != nil {
 		return fmt.Errorf("server credentials are invalid: %v", err)
 	}
 	if strings.Contains(config.Get().Server.StaticContentRootDirectory, "..") {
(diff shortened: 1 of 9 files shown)


kiali.go
```diff
@@ -155,9 +155,10 @@ func waitForSecret() {
 		}
 	}()
 	secret := <-foundSecretChan
+
 	cfg := config.Get()
-	cfg.Server.Credentials.Username = secret.Username
-	cfg.Server.Credentials.Passphrase = secret.Passphrase
+	cfg.Server.Credentials = security.CredentialList{security.Credentials{Username: secret.Username, Passphrase: secret.Passphrase}}
```
jmazzitelli: I don't think this is what you want.

This code here sleeps until a secret is detected (this allows Kiali to start up first, and the user can create the secret later). Before, the secret only had a single username/passphrase in it - so when the secret was mounted, the Credentials (single user/pass) was sent to the channel. The code outside the loop then reads the single credentials found in the channel and stores it in the cfg.Server.Credentials as a kind of cache - so the authentication code doesn't have to keep cracking open the secret everytime we need to authenticate a user request.

But this new code appears to be creating a CredentialsList with just that single item still (secret.Username, secret.Passphrase). But now that we support multiple ones, shouldn't the cfg.Server.Credentials be the full list of all the names found in the secret? This code seems to only support just a single set of credentials?

cfcosta: That's true, I changed this to get something compiling but the tests/compilation did not catch it.

jmazzitelli: Maybe not for this PR, but when the full effort you are working on is done, update the molecule test found here to test this (this molecule tests that the read-only and read/write roles are created properly - we can enhance this test to actually make sure we can log in with these different users):

https://github.com/kiali/kiali/blob/master/operator/molecule/roles-test

This test here is for the LDAP stuff - but it provides an example how to log in as different users within the molecule framework:

https://github.com/kiali/kiali/tree/master/operator/molecule/ldap-test

cfcosta:
Closed in favour of the Token authentication implemented by @israel-hdez.

PR #1749, opened 2019-10-01
Consolidate Cache work

### LINKED ISSUE #1386: Improve cache support in Kiali backend
Kiali introduced caches in kubernetes/cache.go.
With the multitenancy work these caches were switched off in order to not introduce additional complexity to the backend.

Now that multitenancy work is consolidated, it's time to revisit the use of these caches to improve overall performance of Kiali backend, specifically in list queries that involve multiple namespaces and multiple resources.

Tasks:
- [ ] Implement cache controllers for Istio API
JIRA: https://issues.jboss.org/browse/KIALI-2571
- [ ] Enable kubernetes/cache.go in multi-tenancy scenarios
JIRA: https://issues.jboss.org/browse/KIALI-2572

lucasponce:
@burmanm I'm taking your PR https://github.com/kiali/kiali/pull/1263 and consolidate your work here as well.

Also I'm going to simplify some things: like the hack to capture internal errors inside the client-go/cache package.

Another action I want to add is to improve the Cache config with two directions: user can define if they want query all controllers for workloads, and also which types they want to cache.

This is not yet covered, but cache will need to use sa tokens instead to user ones (otherwise it would be impossible as we need to maintain a cache per user/namespace/type), but that logic will be placed outside of caching.

I've just started, but I'm going to cherrypick your commits and also consolidate here to work actively on that.

Fixes #1386 

For testing (cc @hhovsepy @gbaufake):

This PR adds more configuration cache options to be activated via the kiali_cr.yaml:
https://github.com/kiali/kiali/pull/1749/files#diff-f8f6c9abc6c5969045b61451c949e922

When cache is enabled, now it expects that user defines the namespace to cache (or a regex used).

Also, important, there are two parameters to use in the cache to define a refresh rate for objects, and another one for Namespaces / Projects.

As an example, this config would cache bookinfo namespace, and also would have a 300 seconds duration for full refresh (that means that cache is synced via events at any time, but after 300 seconds, it creates a full refresh).

Also, kiali caches the namespaces per user/token, this cache is usually more short lived and it can be configured in a different way, on my example I used 120 seconds but by default it uses 10 second.

```
  cache_enabled: true
  cache_duration: 300
  cache_namespaces:
  - bookinfo
  cache_token_namespace_duration: 120
```

If you would like to test to cache every namespace you can use this config:

```
kubernetes_config:
  cache_enabled: true
  cache_duration: 300
  cache_namespaces:
  - .*
  cache_token_namespace_duration: 120
```

Kiali is not going to cache every resource, this implementation is focus in the heavy queries only.
With the feedback from QE we can adjust this logic and check what it's worth to introduce in cache or not.

Full documentation is written in the kiali_cr.yaml, but please, let me know if there is any doubt.


kubernetes/cache/cache.go
```diff
@@ -0,0 +1,146 @@
...
+
+	return true
+}
+
+func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
+	if !c.isCached(namespace) {
+		return false
+	}
+	if _, exist := c.nsCache[namespace]; !exist {
+		return c.createCache(namespace)
```
burmanm: From my testing, you need to have a method to slow down these requests. If too many createCache are called at the same time, you will run into the timeout as the K8s API will throttle the amount of informers that are created. Thus, too many at once will give you timeouts (once there are enough namespaces).

lucasponce: Yeah, there are three areas we can touch this:
- Explicit namespaces (only namespaces selected by user are cached).
- Explicit resources (mainly for Istio, but also this can be a flag and let user to decide if he considers some workloads - jobs, replicationcontrollers - I see there are room space here too).
- Also tweak the connection settings, there are some parameters we can also introduce.

And for createCache, one thing can be queue them in a sequential way. In theory that will be a "one time" operation and that would reduce the overhead as well.

kubernetes/cache/istio.go
```diff
@@ -20,8 +20,7 @@ type (
 
 func (c *kialiCacheImpl) CheckIstioResource(resource string) bool {
 	// This list can be populated from configuration in the future
-	istioResources := []string{"VirtualService", "DestinationRule", "Gateway", "ServiceEntry"}
-	for _, r := range istioResources {
+	for _, r := range c.cacheIstioTypes {
```
burmanm: Considering this method is called so often (every lookup), it makes sense to use a map here instead of always doing a linear search.

lucasponce: +1

lucasponce:
I've added the changes in the operator, @burmanm (cc @jmazzitelli)

But this PR is not ready, it has a security flaw that I need to resolve, in the case of being deployed in OpenShift, we need to "cache" a list of projects per user in a way that is not an overkill, and also this design needs to work with no cluster permissions, too.

lucasponce:
@jotak Added you to this draft as well.
It's still in draft, tested only locally but is getthing in shape with the final idea I had for this.

So, I'd appreciate feedback.
There is some pending work, but I guess it can be started to be reviewed as well.

gbaufake:
I can start the performance testing when this PR is ready.

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -413,6 +409,36 @@ spec:
 # Only used when cache_enabled is true.
 #    ---
 #    cache_duration: 300000000
+#
```
jmazzitelli: Note: all values in this kiali_cr.yaml document for the user what the actual defaults are. So users can come to this file and see what the operator's default values are in case the user does not set these values themselves.

I see some values here are not the actual defaults. Change these so it matches the defaults yaml. I'll give suggestions below.

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -413,6 +409,36 @@ spec:
...
+#    - "bookinfo"
+#
+# List of controllers that won't be used for Workload calculation.
+# Kiali queries Deployment,ReplicaSet,ReplicationController,DeploymentConfig,StatefulSet,Job and CronJob controllers.
+# Deployment and ReplicaSet will be always queried, but ReplicationController,DeploymentConfig,StatefulSet,Job and CronJobs
+# can be skipped from Kiali workloads query if they are present in this list.
+#    ---
+#    excluded_workloads:
+#    - "CronJob"
+#    - "Job"
```
jmazzitelli: `DeploymentConfig` is also found in the default array for this value - add it here in alphabetical sort order:

```suggestion
#    - "DeploymentConfig"
#    - "Job"
```

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -413,6 +409,36 @@ spec:
...
+#    ---
+#    cache_istio_types:
+#    - "VirtualService"
+#    - "DestinationRule"
+#    - "Gateway"
+#    - "ServiceEntry"
+#
+# List of namespaces or regex defining namespaces to include in a cache.
+#    ---
+#    cache_namespaces:
```
jmazzitelli:
```suggestion
#    cache_namespaces: []
```

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -413,6 +409,36 @@ spec:
...
+#    - "bookinfo"
```
jmazzitelli: `bookinfo` is not a true default and thus should be removed.

```suggestion
```

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -127,9 +127,20 @@ kiali_defaults:
...
     burst: 200
-    qps: 175
     cache_enabled: false
+    cache_istio_types:
+    - "VirtualService"
+    - "DestinationRule"
+    - "Gateway"
+    - "ServiceEntry"
     cache_duration: 300000000
+    excluded_workloads:
```
jmazzitelli: Make sure to specify default values (even if empty) for all settings. Only in rare cases do we not define values in here. This ensures the values get into the configmap and are not left undefined (which could cause errors in the operator ansible scripts or the kiali server itself).
```suggestion
    cache_namespaces: []
    excluded_workloads:
```

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -127,9 +127,20 @@ kiali_defaults:
 
   kubernetes_config:
     burst: 200
-    qps: 175
     cache_enabled: false
+    cache_istio_types:
+    - "VirtualService"
+    - "DestinationRule"
+    - "Gateway"
+    - "ServiceEntry"
```
jmazzitelli: You should alphabetize this list. "VirtualService" should go at the end.

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -413,6 +409,36 @@ spec:
...
 #    cache_duration: 300000000
+#
+# Kiali can cache VirtualService,DestinationRule,Gateway and ServiceEntry Istio resources if they are present
+# on this list of Istio types. Other Istio types are not yet supported.
+#    ---
+#    cache_istio_types:
+#    - "VirtualService"
+#    - "DestinationRule"
+#    - "Gateway"
+#    - "ServiceEntry"
```
jmazzitelli: You should alphabetize this list. "VirtualService" should go at the end.

business/apps.go
```diff
@@ -151,7 +152,15 @@ func fetchNamespaceApps(k8s kubernetes.IstioClientInterface, namespace string, a
 	go func() {
 		defer wg.Done()
 		var err error
-		services, err = k8s.GetServices(namespace, nil)
+		// Check if namespace is cached
+		if kialiCache != nil && kialiCache.CheckNamespace(namespace) {
+			// Cache uses Kiali ServiceAccount, check if user can access to the namespace
+			if _, err := layer.Namespace.GetNamespace(namespace); err == nil {
```
jotak: I guess it should be simple equal `=` instead of `:=`, no? Else the variable is shadowed and the error will not be reported on chanel

lucasponce:
@hhovsepy @gbaufake I'm going to move this from Draft to Ready, looking forward to get QE feedback in two ways: a) regression, to be sure everything is working as expected, b) performance impact.

This PR effectly saves Kiali from doing queries into the Kubernetes API which was identified as a source of potential perf penalty.

Also, configuration is rich to adjust and tweak parameters and see what config works better.

This work is also focus in the heavy scenarios identified, in other words it doesn't cache everything, so, it gives room to improvement depending on QE results.

Let me know how it looks like there, and don't hesitate to ask me any detail needed.

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -413,6 +409,42 @@ spec:
...
+#    - "VirtualService"
+#
+# List of namespaces or regex defining namespaces to include in a cache.
+#    ---
+#    cache_namespaces: []
+#
+# Cache duration expressed in nanoseconds
```
jotak: As it's configuration, shouldn't it be something more human-friendly like seconds? I understand we're targeting short-lived cache, but to the point of nanos? :-)

lucasponce: Yeah, good point, I used the native field of the API which is in nanos, that's why I propagated.

business/namespaces.go
```diff
@@ -105,6 +115,12 @@ func (in *NamespaceService) GetNamespace(namespace string) (*models.Namespace, e
 	promtimer := internalmetrics.GetGoFunctionMetric("business", "NamespaceService", "GetNamespace")
 	defer promtimer.ObserveNow(&err)
 
+	if kialiCache != nil {
+		if ns := kialiCache.GetNamespace(in.k8s.GetToken(), namespace); ns != nil {
+			return ns, nil
+		}
+	}
+
```
jotak: What if cache is enabled but namespace was not found / outdated ? Shouldn't we put it back to cache below?

business/services.go
```diff
@@ -47,7 +47,15 @@ func (in *SvcService) GetServiceList(namespace string) (*models.ServiceList, err
 	go func() {
 		defer wg.Done()
 		var err2 error
-		svcs, err2 = in.k8s.GetServices(namespace, nil)
+		// Check if namespace is cached
+		if kialiCache != nil && kialiCache.CheckNamespace(namespace) {
+			// Cache uses Kiali ServiceAccount, check if user can access to the namespace
+			if _, err := in.businessLayer.Namespace.GetNamespace(namespace); err == nil {
```
jotak: Here again err is not reported to channel

business/services.go
```diff
@@ -57,7 +65,15 @@ func (in *SvcService) GetServiceList(namespace string) (*models.ServiceList, err
 	go func() {
 		defer wg.Done()
 		var err2 error
-		pods, err2 = in.k8s.GetPods(namespace, "")
+		// Check if namespace is cached
+		if kialiCache != nil && kialiCache.CheckNamespace(namespace) {
+			// Cache uses Kiali ServiceAccount, check if user can access to the namespace
+			if _, err := in.businessLayer.Namespace.GetNamespace(namespace); err == nil {
```
jotak: here also

business/services.go
```diff
@@ -204,7 +228,15 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime
 		go func() {
 			defer wg.Done()
 			var err2 error
-			pods, err2 = in.k8s.GetPods(namespace, labelsSelector)
+			// Check if namespace is cached
+			if kialiCache != nil && kialiCache.CheckNamespace(namespace) {
+				// Cache uses Kiali ServiceAccount, check if user can access to the namespace
+				if _, err := in.businessLayer.Namespace.GetNamespace(namespace); err == nil {
```
jotak: here also

business/services.go
lucasponce: Yeah, make lint catched some but not all.
I will look at them, thanks !

business/namespaces.go
lucasponce: Yeah, I relied on GetNamespaces(), but your comment is quite valid.
It should also trigger the cache refresh on this case.

lucasponce:
I've missed a use case for the workloads page that may impact in the performance.
So, I will add a dnm label until is finished.

lucasponce:
Thanks @jotak; I'm +1 with your comment :).

Please @gbaufake @hhovsepy, happy if you can test this one and break it.
I guess it's in the right direction but probably we need to make adjustments with QE feedback from tests environments.

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -127,9 +127,22 @@ kiali_defaults:
...
     burst: 200
-    qps: 175
     cache_enabled: false
-    cache_duration: 300000000
+    cache_duration: 300
+    cache_istio_types:
+      - "DestinationRule"
+      - "Gateway"
+      - "ServiceEntry"
+      - "VirtualService"
```
jmazzitelli: trivial - but all lists in the operator yamls do not have items indented. These items should be moved over to the left - just like you have it for the `excluded_workloads` list down below a little further. So this cache_istio_types should be:
```
cache_istio_types:
- "DestinationRule"
- "Gateway"
- "ServiceEntry"
- "VirtualService"
```

lucasponce: Done. 
Yeah, I missed, the IDE doesn't modify the indentation, but when you edit first time, it proposes the pattern.

business/istio_config.go
```diff
@@ -133,7 +134,18 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 	go func(errChan chan error) {
 		defer wg.Done()
 		if criteria.IncludeGateways {
-			if gg, ggErr := in.k8s.GetGateways(criteria.Namespace); ggErr == nil {
+			var gg []kubernetes.IstioObject
+			var ggErr error
+			// Check if namespace is cached
+			if kialiCache != nil && kialiCache.CheckIstioResource("Gateway") && kialiCache.CheckNamespace(criteria.Namespace) {
```
burmanm: This is just nitpicking (applies to many rows here), but instead of rewriting these "Gateway", maybe the types.go's defined variables could have been used - not that I expect them to ever change.

business/istio_config.go
```diff
@@ -144,7 +156,18 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 	go func(errChan chan error) {
 		defer wg.Done()
 		if criteria.IncludeVirtualServices {
-			if vs, vsErr := in.k8s.GetVirtualServices(criteria.Namespace, ""); vsErr == nil {
+			var vs []kubernetes.IstioObject
+			var vsErr error
+			// Check if namespace is cached
+			if kialiCache != nil && kialiCache.CheckIstioResource("VirtualService") && kialiCache.CheckNamespace(criteria.Namespace) {
+				// Cache uses Kiali ServiceAccount, check if user can access to the namespace
+				if _, vsErr = in.businessLayer.Namespace.GetNamespace(criteria.Namespace); vsErr == nil {
```
burmanm: Does it make sense to use them in this order? First CheckIstioResource && CheckNamespace and only after that the user rights are checked?

lucasponce: The flow used is:
1) Is cached enabled ?
2) Is VirtualService included in cache ?
3) Is Namespace included in cache ?
4) Can user see Namespace X ?

I guess when using cache, 2) will shortcut if the VS is not included, and 3) and 4) is the most logical order, but probably it can be some cornercase where 4) 3) could have more sense. 
I don't know, this flow makes more sense intially to me.

business/layer.go
```diff
@@ -23,13 +28,38 @@ type Layer struct {
...
 
 func GetUnauthenticated() (*Layer, error) {
 	return Get("")
 }
 
 // Get the business.Layer
 func Get(token string) (*Layer, error) {
+	// Kiali Cache will be initialized once at first use of Business layer
+	if kialiCache == nil {
+		initKialiCache()
```
burmanm: Just to confirm - the intention here is that if enabling cache failed the first time, it will never tried again before restart? But then, why is the kialiCache == nil check important? Isn't it repetition to the once.Do() ?

business/istio_config.go
lucasponce: Yeah, it was in my todo list to move this strings to constants.
I can do it in the PR.

business/layer.go
lucasponce: Yeah, kialiCache == nil is redundant here.
I will make it more clear, perhaps putting the once.Do() outside of initKialiCache.

business/workloads.go
```diff
@@ -28,6 +29,30 @@ type WorkloadService struct {
...
+}
+
+func setExcludedWorkloads(m map[string]bool) {
+	defer ew.Unlock()
+	ew.Lock()
+	excludedWorkloads = m
+}
+
+func isWorkloadIncluded(workload string) bool {
+	if excludedWorkloads == nil {
```
burmanm: Why is ``isExcludedWorkloadsEmpty`` using mutex to protect ``excludedWorkloads == nil``, but this one isn't? And why isn't it just calling that method first? Actually, the ``!excludedWorkloads[workload]`` is not protected by the mutex also, it could have issues with ``setExcludedWorkloads``

lucasponce: Indeed, this mutex probably it's not necessary at all, as excludedWorkloads is a one time initialization, similar as kialiCache.
I guess originally it was placed in a different code and that case introduced the mutex.
I'll review & check.

kubernetes/cache/namespaces.go
```diff
@@ -0,0 +1,66 @@
...
+		SetNamespaces(token string, namespaces []models.Namespace)
+		GetNamespaces(token string) []models.Namespace
+		GetNamespace(token string, namespace string) *models.Namespace
+	}
+)
+
+func (c *kialiCacheImpl) SetNamespaces(token string, namespaces []models.Namespace) {
+	defer c.tokenLock.Unlock()
+	c.tokenLock.Lock()
+	nameNamespace := make(map[string]models.Namespace)
```
burmanm: ``, len(namespaces)``

lucasponce:
@burmanm I've addressed your comments.

hhovsepy:
Run full set of UI automations tests on following configs:
1. Cache off.
2. Cache on.
2.1. cache_namespaces "*", cache_token_namespace_duration: 120.
2.2. cache_namespaces 'bookinfo', cache_token_namespace_duration: 120.
2.3. cache_namespaces 'bookinfo', cache_token_namespace_duration: 10.

In all cases with cache on, no performance drop or improvement was found.
No regression was found.

lucasponce:
@hhovsepy @gbaufake 
Thanks, one note, prometheus queries are not cached, just the k8s.
I guess you can notice the cache impact when accessing list pages, when cached, the results should be loaded faster.

In any case, when we have more data with large population we can refine these adjustments.
Also interesting if the Kiali Metrics show good info.

Anyway, thanks for checking, I'm going to merge this one and we can refine on iterations.

lucasponce:
@hhovsepy @gbaufake 
When I answered I thought that I missed to ask a trivial thing; it's weird you don't see any performace (unless the k8s queries are very fast compared with p8s).

So in the kiali log, can you confirm some log like this one:

```
```

For namespaces queried ? 
Just to confirm that there was not a problem passing the config from the CR to Kiali.

Thanks

PR #1765, opened 2019-10-04
[operator] create preliminary manifests for the next versions.

jmazzitelli:
Placeholder version of X.X.X is in place - we'll replace X.X.X with the version at the time of the release when we know what version to use


jmazzitelli:
I'll wait to merge this until next week so as not to interfere with the automated release process that will happen in a couple hours. I don't want those placeholder files to be assigned to the next release tag.

jmazzitelli:
i think i'm going to combine this with #1773

Issue #1713, opened 2019-09-23
ServiceMeshApiDocumentation is returning 500


gbaufake:
**General remarks**

Installed Kiali and bookinfo queried the ServiceMeshApiDocumentation API with 
``api/namespaces/bookinfo/services/details/apispec``

Found by E2E #1712 

**Describe the bug**

It returned 500 instead of 200 with API description with ``error    "Internal error occurred: services \"details\" not found"``

**Versions used**
Kiali-ui: 1.6.0 (74a636489ef6b64d61a220cdd9215eb5851d6204)
Kiali: v1.6.0-SNAPSHOT (10aeec7358482f0d7a401d0c8bd5b9501e32b4a2)
Service Mesh: 1.0.0
Kubernetes flavour and version: Openshift 4.1

**To Reproduce**
1. Install Service Mesh, Kiali and Bookinfo
2. Query KIALI_URL/api/namespaces/bookinfo/services/details/apispec

**Expected behavior**
- Request should return 200.

mwringe:
Are you sure you have the right URL here? I can't reproduce this problem.

I do get the error mentioned if I put a typo in the namespace name (eg 'api/namespaces/foo/services/details/apispec') or if I put a typo in the service name.

Even with the typos in the URL, we shouldn't be returning a 500 response but a 404

gbaufake:
closed by #1769

PR #1769, opened 2019-10-07
Return 404, not 500, when service is not found in apidoc call

jshaughn:
Fixes https://github.com/kiali/kiali/issues/1713

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -119,8 +119,7 @@ func (in *SvcService) GetServiceApiDocumentation(namespace, service string) (str
 	conf := config.Get()
 	svc, err := in.k8s.GetService(namespace, service)
 	if err != nil {
-		log.Errorf("Error fetching API documentation for service [%s:%s]: %s", namespace, service, err)
-		return "", errors.NewInternalError(err)
+		return "", err
 	}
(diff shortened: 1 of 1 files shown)


jotak:
How can we query for apidoc on a service that doesn't exist? isn't it something else that is flawed?

jshaughn:
> How can we query for apidoc on a service that doesn't exist? isn't it something else that is flawed?

@jotak We can't, so with this change you get a 404 if the service does not exist or if it does exist but has no apidoc.  Is that not correct?

jotak:
Well, I mean we should be *trying* to run that query in a first place if the service doesn't exist, that's where I think there's another problem.
If the service has no apidoc you won't get an error on `in.k8s.GetService(namespace, service)` call so it's probably handled at another place.

jotak:
@jshaughn it's ok, I've read more carefully the issue and it seems the bug is not reproducible (or not easily?) so it was probably something more like a timing issue with a deleted service or something like that?

So I'm ok with this fix as it just forwards the k8s error to the frontend (hence returning 404 when service is not found) - but it would not solve the original bug if it comes to be reproducible, as a code 200 was actually expected in e2e test.

jshaughn:
@jotak OK, thanks.  I can only guess that the original problem had to do with a typo of a manual api call in the browser.  Otherwise, I agree, it is something odd.

PR #1774, opened 2019-10-08
OSSM-102 CVE-2019-11253

jshaughn:
- update version of gopkg/yaml to v2.2.4

@lucasponce  I put DNM label on this only to wait until 1.0.1 is released.  Because this is targeted for v1.0.2.


lucasponce:
Thanks @jshaughn. 
@hhovsepy can we run a regression test on this ?
This change looks trivial but worth if we can validate that there are not any side effect.

PR #1775, opened 2019-10-08
OSSM-103 change the name of the operator deployment. OLM should see t…

jmazzitelli:
I talked to the OLM guys.  They said changing the name of the operator deployment in the CSV will cause OLM to destroy and re-create the deployment, thus avoiding that update error.

### DIFF
--- a/operator/manifests/kiali-ossm/1.0.6/kiali.v1.0.6.clusterserviceversion.yaml
+++ b/operator/manifests/kiali-ossm/1.0.6/kiali.v1.0.6.clusterserviceversion.yaml
@@ -212,7 +212,7 @@ spec:
     strategy: deployment
     spec:
       deployments:
-      - name: kiali-operator
+      - name: kiali-operator2
         spec:
           replicas: 1
           selector:


lucasponce:
I am confused.

Why master branch should handle a kiali-ossm/1.0.6/kiali.v1.0.6.clusterserviceversion.yaml file which looks like specific version dependent ?

jmazzitelli:
@lucasponce these operator manifest files are the "golden" copies for every OLM manifest package everywhere. It is the "one place" where we keep the "true" copies.

This is to avoid the chaos and mess that we have today where "golden" copies are strewn all over the  place in different git repos like:

community manifest:
https://github.com/operator-framework/community-operators/tree/master/community-operators/kiali

upstream manifest: https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/kiali

productization manifest:
http://XXX.redhat.com/cgit/containers/...kiali-operator/tree/manifests?h=rhossm-1.0-rhel-7

(and potentially -- soon to be -- manifests for QE which need to be slightly different)

Now they are in one place, and we can just blindly copy them over to where they need to be.

And now when our Kiali devs need to add a new permission to the Kiali role, they do not need to edit files in all of those external git repo places - all changes that devs make go into the files in kiali github.com repo. When we update the various OLM manifests in those different external places, we just blindly copy them -- we do not have to go through the mistake-prone process of remembering what has to change where.

Note: Maistra is doing the same thing - they put all their manifests [here](https://github.com/Maistra/istio-operator/tree/maistra-1.1/manifests-maistra) and [here](https://github.com/Maistra/istio-operator/tree/maistra-1.1/manifests-servicemesh) in their github.com repo.

jshaughn:
@jmazzitelli Maybe this would be useful info for kiali-dev list?

jmazzitelli:
> @jmazzitelli Maybe this would be useful info for kiali-dev list?

I'll send an email out

Issue #1786, opened 2019-10-11
Can't open service details


jotak:
Hi can't open any service details, here's what I'm getting:

![Capture d’écran de 2019-10-11 12-04-22](https://user-images.githubusercontent.com/2153442/66644037-414a2e80-ec20-11e9-8e77-a92ef285c537.png)

I first imagined it was something wrong in my environment but doesn't seem so.
On the REST query for service details, an empty object is returned for `validations`.

Also there's some code smell UI-side regarding this, in several place I'm seeing this type cast: `{} as ObjectValidation` which looks quite wrong, especially in `ServiceInfoDescription`, as straight after the validation object is accessed as `this.getValidations().checks` which results in the error I'm seeing.

I'm setting this issue as a blocker as I can't see any service details.

xeviknal:
It might be related to https://github.com/kiali/kiali-ui/pull/1440 @aljesusg wdyt?

lucasponce:
Mmm I don't get it.

I'm using a fresh master with kiali backend:

commit a52035c7c801c70ae76e616535cd0414ad4804b4 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Merge: 7e0e54a2 1ad19f43
Author: Guilherme Baufaker Rêgo <gbaufake@redhat.com>
Date:   Thu Oct 10 22:15:30 2019 -0300


and kiali-ui:

commit d490082f8c3b58bbd060ed0598e9ef065c11ffd2 (HEAD -> master, upstream/master, origin/master, origin/HEAD)
Merge: 19c7c2db 12519199
Author: Joel Takvorian <jtakvori@redhat.com>
Date:   Fri Oct 11 12:21:00 2019 +0200

    Merge pull request #1474 from jotak/pf4-metrics-settings
    
    Move metrics settings to PF4

And I have a bogus Service:

![image](https://user-images.githubusercontent.com/1662329/66647073-35626a80-ec28-11e9-9b7b-b5ba8fce223e.png)

![image](https://user-images.githubusercontent.com/1662329/66647085-3d220f00-ec28-11e9-851d-00f7e1e711a2.png)

![image](https://user-images.githubusercontent.com/1662329/66647111-48753a80-ec28-11e9-997c-025c2a7a4b67.png)

I'm using a localhost mode, but if the problem is between backend/fronend I guess I should reproduce it here as well.

xeviknal:
Tried to reproduce the bug but I couldn't. I am using the same version as mentioned right below.

[...]

PR #1788, opened 2019-10-11
Fix potentially ignored error from validation checks

jotak:
Closes https://github.com/kiali/kiali/issues/1786

@burmanm @xeviknal If i'm correct, the error resulting from validation checks was ignored. Now at least it's logged.
If I'm correct having an error might result in empty validations object.

### DIFF
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -73,7 +73,7 @@ func (in *IstioValidationsService) GetValidations(namespace, service string) (mo
 	close(errChan)
 	for e := range errChan {
 		if e != nil { // Check that default value wasn't returned
-			return nil, err
+			return nil, e
 		}
 	}
 
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -6,6 +6,7 @@ import (
 
 	apps_v1 "k8s.io/api/apps/v1"
 	core_v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/kiali/kiali/business/checkers"
 	"github.com/kiali/kiali/config"
@@ -332,10 +333,8 @@ func (in *IstioValidationsService) fetchNonLocalmTLSConfigs(mtlsDetails *kuberne
 		if !in.k8s.IsMaistraApi() {
 			if meshPolicies, iErr := in.k8s.GetMeshPolicies(); iErr == nil {
 				details.MeshPolicies = meshPolicies
-			} else {
-				// This query can return false if user doesn't have cluster permissions
-				// On this case we log internally the error but we return an empty list
-				log.Warningf("GetMeshPolicies failed during a TLS validation. Probably Kiali doesn't have cluster permissions. Error: %s", iErr)
+			} else if !checkForbidden("GetMeshPolicies", iErr, "probably Kiali doesn't have cluster permissions") {
+				errChan <- iErr
 			}
 		} else {
 			// ServiceMeshPolicies are namespace scoped.
@@ -347,10 +346,8 @@ func (in *IstioValidationsService) fetchNonLocalmTLSConfigs(mtlsDetails *kuberne
 			controlPlaneNs := config.Get().IstioNamespace
 			if serviceMeshPolicies, iErr := in.k8s.GetServiceMeshPolicies(controlPlaneNs); iErr == nil {
 				details.ServiceMeshPolicies = serviceMeshPolicies
-			} else {
-				// This query can return false if user can't access to controlPlaneNs
-				// On this case we log internally the error but we return an empty list
-				log.Warningf("GetServiceMeshPolicies failed during a TLS validation. Probably user can't access to %s namespace. Error: %s", controlPlaneNs, iErr)
+			} else if !checkForbidden("GetServiceMeshPolicies", iErr, fmt.Sprintf("probably user can't access to %s namespace", controlPlaneNs)) {
+				errChan <- iErr
 			}
 		}
 	}(mtlsDetails)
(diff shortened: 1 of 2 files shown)


handlers/services.go
```diff
@@ -102,8 +103,11 @@ func ServiceDetails(w http.ResponseWriter, r *http.Request) {
 		go func(istioConfigValidations *models.IstioValidations, err *error) {
 			defer wg.Done()
 			istioConfigValidationResults, errValidations := business.Validations.GetValidations(namespace, service)
-			if errValidations != nil && *err == nil {
-				*err = errValidations
+			if errValidations != nil {
+				if *err == nil {
```
jotak: BTW that part doesn't look thread-safe

burmanm: It isn't, but the end result will still always be correct (as it doesn't matter which thread wins - one will).

jotak: The error handling is quite confusing here. `err` is read / written in goroutine as well as in the main thread for different purposes. I guess we should clear that up.

burmanm: To clarify, if the error was created by the main thread, it will exit at line 118 but if the validation fails it will most likely exit at line 128 (as there's a wg.Wait()). 

In any case, both will respond with the error so why log it?

burmanm: I don't think it matters to the caller, right? They will get an error that prevented the call in any case, it doesn't matter if both failed.

jotak: Hmm in my understanding, the validation failure is sent as response only if the call to `business.Validations.GetValidations(namespace, service)` returns after `business.Svc.GetService(namespace, service, rateInterval, queryTime, requestToken)` ... which is not guaranteed

burmanm: It's guaranteed in line 124 by the ``wg.Wait()`` if the validations were requested. If error happened with the ``GetService`` then we don't wait for the validations as it doesn't make any sense.

jotak: Ok, I'll take another look at it with the fresh eyes of a Monday morning :-)

jotak: So I still don't get it. Here's the scenario that I think results in ignoring the error:

1. `GetValidations` call begins (l.105)
2. `GetServices` call begins (l.122)
3. `GetValidations` call ends, `err` filled with a validation error (l.108)
4. `GetServices` call ends, `err` being overwritten with `nil`, given there's no error while getting the service (l.122)
5. `wg.Wait` is not blocking as the goroutine has already finished (l.124)
6. `err` is `nil` => responding success & losing track of the validation error that happened.

Makes sense?

burmanm: Ah yeah, in that scenario it does, GetService can overwrite the GetValidations' error.

jotak:
@burmanm I've edited my commit, what do you think of this?

handlers/services.go
```diff
@@ -95,19 +95,18 @@ func ServiceDetails(w http.ResponseWriter, r *http.Request) {
...
 	var istioConfigValidations = models.IstioValidations{}
+	var errValidations error
 
 	wg := sync.WaitGroup{}
 	if includeValidations {
 		wg.Add(1)
-		go func(istioConfigValidations *models.IstioValidations, err *error) {
+		go func(istioConfigValidations *models.IstioValidations) {
 			defer wg.Done()
 			istioConfigValidationResults, errValidations := business.Validations.GetValidations(namespace, service)
```
burmanm: This errValidations is not necessarily the pointer to the above errValidations as you use ``:=``. So inside the goroutine we have a different struct. You need to either pass it to the go func in the parameters or not use ``:=``

jotak: oops you're right

jotak:
Soooo this is now "fixed" as the error is now really coming down to the UI. We can explain now what happened here #1786 .
BUT the funny thing is... it's certainly not what we want!
Now a call to service details returns 500 internal server error (hence I can't see any service) because of this, on my setup: 
`clusterrbacconfigs.rbac.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list clusterrbacconfigs.rbac.istio.io at the cluster scope: no RBAC policy matched`

It should be a warning, but it should not be something necessary that prevents seeing any service.
I don't know if you have a suggestion on how to fix that.

jotak:
Put differently: what should we do with errors during validation?

- Just log, without informing frontend?
- Inform frontend but without a code 500 to still allow service display
- Inform frontend with code 500 (but I think it's bad especially if the errors during validation are something related to the user setup - which isn't even incorrect, like in my case here)

lucasponce:
@jotak backend errors should be moved to UI but there are a couple of exceptions where we may want to put just a warning.

I'm interested in that, because, clusterrbacconfigs should be empty when kiali doesn't have cluster rights.

Perhaps due the error you are fixing here, it was missed in one of my previous PRs for that.

I can take a look to confirm what should send.

jotak:
@lucasponce just investigating a little bit more: this error was not triggered from `GetMeshPolicies`, but from `fetchAuthorizationDetails` (where there's no similar code to return empty data)

jotak:
@lucasponce take a look on my last commit please : I'm checking explicitly if errors are "forbidden", in which case we log silently ; else error is sent to channel. Doing this both for 'Get[Service]MeshPolicies' as before, plus 'GetAuthorizationDetails' which is where it was failing for me.

lucasponce:
Yes, the GetAuthorizationDetails is the offending one.
It should have a similar logic as GetClusterRbacConfigs which shouldn't propagate fail but just send a warning.

I'd suggest to unify the logic in both, I mean if we have a check() method to validate the error, worth to use it in both places or viceversa, just send a warning when that method fails.

business/istio_validations.go
```diff
@@ -399,3 +399,17 @@ func (in *IstioValidationsService) fetchAuthorizationDetails(rValue *kubernetes.
 		}
 	}
 }
+
+func checkForbidden(caller string, err error, context string) bool {
+	// Some checks return 'forbidden' errors if user doesn't have cluster permissions
+	// On this case we log it internally but we don't consider it as an internal error
+	if errors.IsForbidden(err) {
```
lucasponce: +1

PR #1798, opened 2019-10-14
[hack] move to openshift 4.2

jmazzitelli:


jmazzitelli:
This appears to be working now. I'm still testing and tweaking that README - but so far things are good as-is. Would be great if someone else could try it out, too.

This will install OpenShift 4.2 on AWS and installs ServiceMesh 1.0.1 and its dependencies (jaeger and kiali).

jmazzitelli:
I need to get the sm-uninstall and k-uninstall commands working now that subscriptions are created a bit differently

jmazzitelli:
sm-install, sm-uninstall, and k-uninstall are working again. this PR is again ready for review and merge

PR #1801, opened 2019-10-15
Add Jaeger configuration for namespaceSelector

aljesusg:
Fix https://github.com/kiali/kiali/issues/1785

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -120,14 +120,15 @@ type GrafanaVariablesConfig struct {
 
 // TracingConfig describes configuration used for tracing links
 type TracingConfig struct {
-	Auth         Auth   `yaml:"auth"`
-	Enabled      bool   `yaml:"enabled"` // Enable autodiscover and Jaeger in Kiali
-	InClusterURL string `yaml:"in_cluster_url"`
-	Namespace    string `yaml:"namespace"`
-	Path         string `yaml:"-"` // Path stores the value of QUERY_BASE_PATH, not configurable in ConfigMap
-	Port         int32  `yaml:"port"`
-	Service      string `yaml:"service"`
-	URL          string `yaml:"url"`
+	Auth              Auth   `yaml:"auth"`
+	Enabled           bool   `yaml:"enabled"` // Enable autodiscover and Jaeger in Kiali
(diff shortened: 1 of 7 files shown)


jmazzitelli:
Can we hold off merging this until I get this PR merged (I have a lot of changes going into config stuff): PR #1840 ... It will be easier to reconcile the conflicts if we merge that PR first and then this PR after.

aljesusg:
ok @jmazzitelli

jmazzitelli:
@aljesusg that PR of mine was approved and merged. So you'll have conflicts here - contact me if you have any questions how to resolve the conflicts.

config/config.go
```diff
@@ -82,6 +82,7 @@ const (
 	EnvThreeScaleServicePort                = "THREESCALE_SERVICE_PORT"
 	EnvTracingEnabled                       = "TRACING_ENABLED"
 	EnvTracingInClusterURL                  = "TRACING_IN_CLUSTER_URL"
+	EnvTracingNamespaceSelector             = "TRACING_NAMESPACE_SELECTOR"
```
jmazzitelli: This can be deleted - we no longer have env vars for all config settings (we only use env vars for sensitive values like passwords).

config/config.go
```diff
@@ -384,6 +386,7 @@ func NewConfig() (c *Config) {
 	c.ExternalServices.Tracing.Path = ""
 	c.ExternalServices.Tracing.InClusterURL = strings.TrimSpace(getDefaultString(EnvTracingInClusterURL, ""))
 	c.ExternalServices.Tracing.URL = strings.TrimSpace(getDefaultString(EnvTracingURL, ""))
+	c.ExternalServices.Tracing.NamespaceSelector = getDefaultBool(EnvTracingNamespaceSelector, true)
```
jmazzitelli: This line is now changed in the latest master. You can actually delete this. If the unit tests need this "true" value to be defined in the default Config object created by NewConfig() function, you can add it in the the NewConfig() function, but for the most part, we don't need to do this anymore because the operator will put this in the ConfigMap explicitly and thus we know it will always be set.

aljesusg:
I did the changes @jmazzitelli . Thanks

PR #1806, opened 2019-10-15
[operator] upstream manifests are now nested

jmazzitelli:
now that operatorhub.io supports it, the kiali manifests for upstream now use the new nested format


jmazzitelli:
OperatorHub.io now supports the new nested format for manifests. Kiali upstream has been updated. This just keeps our golden copies in sync. These files match those that were merged here: https://github.com/operator-framework/community-operators/pull/775

Issue #1804, opened 2019-10-15
False positives on validation "More than one Gateway for the same host port combination"


nabam:
**Describe the bug**
Due to use of regex matching in [business/checkers/gateways/multi_match_checker.go#L124](https://github.com/kiali/kiali/blob/v1.7.0/business/checkers/gateways/multi_match_checker.go#L124) gateway host names with only difference in `.` symbol cause false positives on validation "More than one Gateway for the same host port combination"

**Versions used**
Kiali: 1.7.0
Istio: 1.1.13-gke.0
Kubernetes flavour and version: GKE 1.13.7-gke.24

**To Reproduce**
1. Create a gateway with 2 hosts: "api.dev.example.com" and "api-dev.example.com"
2. Go to 'Istio Config' section 
3. Observe 'More than one Gateway for the same host port combination' validation warning

**Expected behavior**
Configuration above is valid and should not cause warnings

gateway example:
```
kind: Gateway
apiVersion: networking.istio.io/v1alpha3
metadata:
  name: flames-gateway
  namespace: default
spec:
  servers:
    - hosts:
        - api.dev.example.com
        - api-dev.example.com
      port:
        name: https-443
        number: 443
        protocol: HTTPS
      tls:
        credentialName: example-com-cert
        mode: SIMPLE
  selector:
    istio: ingressgateway
```
<img width="1249" alt="Screenshot 2019-10-15 at 16 47 35" src="https://user-images.githubusercontent.com/751681/66844932-7b446900-ef6f-11e9-8f0a-035dde75ee6a.png">

lucasponce:
cc @burmanm @xeviknal 

can you take a look at this one ?

xeviknal:
Hey @nabam! Thanks for submitting the issue and letting us know about the bug.
I just reproduced the bug so I can confirm it.
Thanks again!

nabam:
Thanks! I'm impressed by the speed of resolution!

PR #1809, opened 2019-10-16
KIALI-1804 Remove false positive coming from . chars in RegExps

xeviknal:
Closes https://github.com/kiali/kiali/issues/1804

Before:

After:
Screenshot of Kiali Console (74)


lucasponce:
Please don't use KIALI-XXXX codes in PRs or commits.
Those are from the legacy JIRA.

PR #1830, opened 2019-10-22
[operator] fix molecule tests so they use the route and LoadBalancer

### LINKED ISSUE #1832: [operator] molecule tests fail when testing AWS clusters
Because we are using mainly AWS clusters now (CRC VMs cannot run Service Mesh/Istio + Kiali so we have no way of testing with local clusters anymore), we need the molecule tests to run against remote AWS clusters. But today the molecule tests fail when testing with AWS clusters. We need to fix the tests.

jmazzitelli:
fixes #1832 

The molecule tests will fail on an AWS cluster due to the way the kiali_base_url was being defined.  The LDAP test also failed because the LDAP service type was incompatible with running on a remote AWS cluster.

This fixes it to use the Kiali Route, which should always be there.
This makes the LDAP service of type "LoadBalancer" and the test uses the external load balancer URL.

Because of the use of Route, this means the molecule tests will only run on OpenShift clusters (which I think we assumed anyway, but I just wanted to point this out).

@gbaufake I'm running tests over here myself, but you might want to review this and run those e2e/olm tests you have.

### DIFF
--- a/operator/molecule/common/tasks.yml
+++ b/operator/molecule/common/tasks.yml
@@ -46,21 +46,15 @@
     - app = kiali
   register: kiali_service
 
-- name: Get cluster node IP that can be used to communicate to cluster services
+- name: Get Kiali Route
   k8s_facts:
-    api_version: v1
-    kind: Node
-  register: cluster_node
-- set_fact:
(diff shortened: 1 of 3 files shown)


jmazzitelli:
Looks like the ldap-test is still failing. Need to figure out way to deploy an LDAP server and have it accessible to the molecule test.

Issue #1255, opened 2019-07-18
Operator - where do I store sensitive values?


omerlh:
**Is your feature request related to a problem? Please describe.**
Currently, the operator CRD allows to define various information, including sensitive values - like authentication credentials used to consume Prometheus or Grafan.

**Describe the solution you'd like**
Allow specifying a secret that will contain the relevant secret, like [Jaeger operator](https://github.com/jaegertracing/jaeger-operator#secrets-support).

**Describe alternatives you've considered**
Storing them hardcoded in the CRD.

jmazzitelli:
I'm going to research how Service Mesh/Maistra and Jaeger does it and see if we can mimic how they do it.

jmazzitelli:
Here's what my first attempt at an implementation will look like.

I am going to leverage the fact that almost all of the ConfigMap settings can be overridden by environment variables. In this case, the server will look for PROMETHEUS_PASSWORD, GRAFANA_PASSWORD and TRACING_PASSWORD env vars and use those for values in place of any ConfigMap setting for `external_services.[prometheus,grafana,tracing].auth.password` respectively. The analogous thing will happen for "token" values as well (as opposed to "password").

What the Kiali operator will then do is create an env var for these settings if the external_services settings has a value of the form `secret:<secretName>:<secretKey>`. For example, if the  Kiali CR has a setting `external_services.grafana.auth.password` whose value is `secret:myGrafanaSecret:myPassword` then the operator will create the kiali deployment with the following env var set:

```
spec:
  containers:
    env:
    - name: GRAFANA_PASSWORD
      valueFrom:
        secretKeyRef:
          name: myGrafanaSecret
          key: myPassword
```

With this in place, it *should* be the case that no code changes need to be made inside the Kiali server Go code. This should "just work" considering the env var should bind its value to the secret's key.

This also provides full backward compatibility because the operator and kiali will behave exactly as it did before if the password or token value is not prefixed with "secret:" (meaning, the password or token is encoded directly into the Kiali CR and ConfigMap).

NOTE: This requires the user to manage the secrets himself, but this is true for any solution to this issue because the whole idea is that the kiali operator and kiali cannot be directly told what the actual password is - thus, it will require an administrator to create the secret with the password encoded in it and just pass the secret details to the Kiali operator.

jmazzitelli:
PR #1827 is ready for review and testing.

PR #1840, opened 2019-10-23
rip out the obsolete env var and default settings

### LINKED ISSUE #1837: get rid of the references to environment variables in config.go
Back in the early days of the project, users were to hand-edit the ConfigMap to configure Kiali (now we use the operator and the Kiali CR). To avoid having the user enter EVERY possible setting in the ConfigMap, in `config.go` we hardcoded default values for each setting. If a user did not enter a value for a specific setteing in the ConfigMap, the value would obtain a hardcoded default value defined in config.go.

Because we didn't want to be stuck with hardcoded values for defaults (in case we needed to change it for, say, a hotfix and we didn't want to respin a new image) we provided a backdoor way of changing those hardcoded default values via environment variables.

But now things have changed. The Kiali CR is the user-facing config (not the ConfigMap), and the operator is the one that defines all the defaults (not the server in config.go). If the Kiali CR does not define a particular setting, the operator is responsible for setting it to some default value. In effect, this means the Kiali ConfigMap will have virtually every setting defined - thus the hardcoded defaults in config.go are rarely used and the environment variable overrides are essentially useless since the ConfigMap will have virtually all the values already set (I am hedging here with "virtually" because there may be one or a couple settings in the ConfigMap that the operator allows to be undefined (for special purposes that the Kiali server knows about - but these are rare, special cases and wouldn't have env var overrides and thus we can ignore those special cases).

We should rip out all of the code in config.go that refers to environment variables and hardcoded defaults and just assume the ConfigMap has everything defined. Or, at minimum, get rid of the backdoor environment variable lookups and just set the default ConfigMap structure with hardcoded defaults (if not just empty values). This will make the code leaner, easier to maintain, and easier to understand.

jmazzitelli:
fixes #1837


jmazzitelli:
The first commit cleans up the config so we remove the obsolete "env var override" feature that isn't used anymore and just made the code ugly for no reason. This code still hardcodes some defaults because the unit tests need them (they fail if we don't fill in the Config object with some data).

In an upcoming commit to this PR, I'm going to add in some new env var override capabilities, at least for some settings (specifically for sensitive configuration settings so that we can store them in a secret and pass the secret's value through k8s's feature of binding secret values to env vars). This is to support things like issue #1255).

jmazzitelli:
Note that some of the structs in config.go now have their fields sorted alphabetically - this is to make it easy to match the Go struct with the yaml code in the operator yamls (which are already sorted) - its easy now to find the related code between Go and yaml files and this helps to do things like find settings that might be missing (I already found one missing setting in the yaml).

lucasponce:
I can help to review/test this one.
I develop with a kiali process outside of the cluster and I think I tweak config for this use case.
I'll check if everything look ok from that side.

config/config.go
```diff
@@ -174,54 +107,52 @@ type GrafanaDashboardConfig struct {
...
-	Tracing    TracingConfig    `yaml:"tracing,omitempty"`
 	ThreeScale ThreeScaleConfig `yaml:"threescale,omitempty"`
+	Tracing    TracingConfig    `yaml:"tracing,omitempty"`
 }
 
 // LoginToken holds config used in token-based authentication
 type LoginToken struct {
-	SigningKey        string `yaml:"signing_key,omitempty"`
 	ExpirationSeconds int64  `yaml:"expiration_seconds,omitempty"`
+	SigningKey        string `yaml:"signing_key,omitempty"`
```
lucasponce: Should SigningKey be treated as a potential overwritten property ? As a token or password ?

config/config.go
```diff
@@ -312,135 +243,117 @@ type IstioComponentNamespaces map[string]string
...
-	c.KubernetesConfig.Burst = getDefaultInt(EnvKubernetesBurst, 200)
-	c.KubernetesConfig.QPS = getDefaultFloat32(EnvKubernetesQPS, 175)
-	c.KubernetesConfig.CacheEnabled = getDefaultBool(EnvKubernetesCacheEnabled, false)
-	c.KubernetesConfig.CacheDuration = getDefaultInt(EnvKubernetesCacheDuration, 5*60)
-	c.KubernetesConfig.CacheNamespaces = getDefaultStringArray(EnvKubernetesCacheNamespaces, ".*")
-	c.KubernetesConfig.CacheIstioTypes = getDefaultStringArray(EnvKubernetesCacheIstioTypes, "VirtualService,DestinationRule,Gateway,ServiceEntry")
-	c.KubernetesConfig.CacheTokenNamespaceDuration = getDefaultInt(EnvKubernetesCacheTokeNamespaceDuration, 10)
-	c.KubernetesConfig.ExcludeWorkloads = getDefaultStringArray(EnvKubernetesExcludeWorkloads, "CronJob,DeploymentConfig,Job,ReplicationController,StatefulSet")
-
-	trimmedExclusionPatterns := []string{}
```
lucasponce: This logic is being removed without an equivalent

jmazzitelli: This is one reason why this PR is necessary - the env var default stuff was confusing and rarely used. If you notice, this code was actually virtually useless because it didn't do anything except *if* the exclusion list was specified by an environment variable *and* the same setting wasn't found in the ConfigMap yaml. Well, it never was the former (we never overrode via env var) and since the operator came online it never was the latter either (because this setting is always found in the Kiali CR so even if an env var was specified, it is never used - again, that's the problem with this env var override stuff - it isn't actually getting used now).

So, effectively, this was dead code and so no equivalent is needed.

config/config.go
jmazzitelli: Yeah, that's a good one. We can add that. Let me add another commit for that.

jmazzitelli:
@lucasponce added LOGIN_TOKEN_SIGNING_KEY override env var in the latest commit

lucasponce:
Travis is complaining about formatting, but otherwise, LGTM

jmazzitelli:
I'm not seeing any format changes with "gofmt" or "make format" on my local machine. I need to find out why travis is failing

jmazzitelli:
I pushed a manual edit of the file to get travis to pass. I'll find out why my gofmt isn't matching travis, but for now, this should be good.

PR #1846, opened 2019-10-25
[hack] move to OS 4.2.x and support ssh key

jmazzitelli:
This moves to OpenShift 4.2.2 - this version fixes the operator dependency bug we are hitting when installing ServiceMesh (where it was installing community versions of Jaeger and Kiali when it should not have been).

This also adds support for passing an SSH key to the installer for debugging. See: https://docs.openshift.com/container-platform/4.1/installing/installing_aws/installing-aws-customizations.html#ssh-agent-using_install-customizations-cloud

### DIFF
--- a/hack/aws-openshift.sh
+++ b/hack/aws-openshift.sh
@@ -390,7 +390,7 @@ SCRIPT_ROOT="$( cd "$(dirname "$0")" ; pwd -P )"
 cd ${SCRIPT_ROOT}
 
 # The default version of OpenShift to be downloaded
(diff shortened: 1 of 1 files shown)


jmazzitelli:
@jshaughn you can use any ssh key you have currently installed on your system in your ssh-agent. So if you have a `$HOME/.ssh/id_rsa.pub` for example, run:

`aws-openshift.sh create -sk $HOME/.ssh/id_rsa.pub ...and the rest...`

PR #1853, opened 2019-10-29
[hack] install istio using istioctl. this is how istio 1.4 is installed now

jmazzitelli:

PR #1856, opened 2019-10-30
[hack] this is to support installing upstream istio 1.4.

jmazzitelli:
This is the old hack script that installs Istio using the helm charts. The current istio 1.4 beta is broken when attempting to install via istioctl, so this PR updates the hack script that installs 1.4 using the old helm method. Helm installation like this is deprecated in 1.4 but still works.

### DIFF
--- a/hack/istio/install-istio-kiali-via-helm.sh
+++ b/hack/istio/install-istio-kiali-via-helm.sh
@@ -16,7 +16,7 @@
 ##############################################################################
 
 # ISTIO_DIR is where the Istio download is installed and thus where the Helm charts are found.
-# CLIENT_EXE_NAME must be either "oc", "kubectl", or "istiooc"
+# CLIENT_EXE_NAME must be either "oc", "kubectl".
 ISTIO_DIR=
 CLIENT_EXE_NAME="oc"
 NAMESPACE="istio-system"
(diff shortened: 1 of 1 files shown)


jshaughn:
@jmazzitelli, maybe we should have versioned hack scripts, or add some back compat to ensure we support the supported versions of Istio.  We could just rename the current one with suffix 1_3 and make this one 1_4.

jmazzitelli:
> maybe we should have versioned hack scripts, or add some back compat to ensure we support the supported versions of Istio. We could just rename the current one with suffix 1_3 and make this one 1_4.

We already have poor man's versions - if you look back at our tags or branches you can just use the older ones.

jshaughn:
Well, yes, but that's not convenient when running master :)

jmazzitelli:
This is difficult - because I just tried to create a 1.3 and 1.4 versions, and I just realized some of these changes are not related to the istio version but rather of the openshift version. Unless you want to maintain a matrix of versions "(openshift 3.11, istio 1.3, openshift 4.1, istio 1.3, openshift 4.1, istio 1.4).

And notice I said unless **you** want to maintain... :-p

jmazzitelli:
FWIW: there will be a new hack script for the non-helm installation mechanism - see PR #1853 - so we WILL have different hack scripts here. One for the helm installation (which is for 1.3 and below, still works but deprecated for 1.4) and one for istioctl.

jmazzitelli:
Istio 1.4 has been released - let's review this hack script now and merge it.

jmazzitelli:
to test -- delete (or rename) all of your `_output/istio*` directories and then run

```bash
hack/istio/install-istio-kiali-via-help.sh -ud true
```

This should download Istio 1.4, and install with the demo profile.

jmazzitelli:
I think they changed something and now this doesn't work - missing cni setting in the demo profile. I don't think we can use the demo profile anymore on openshift

jmazzitelli:
OK, I had to make a slight change to the hack script but it is working now. We can merge this to get the hack script to install 1.4 via helm

jmazzitelli:
cc @gbaufake @jshaughn

Issue #1838, opened 2019-10-23
Validate Kiali for Istio 1.4 release


jshaughn:
Kiali should be validated against the Istio 1.4 RC releases in preparation for the GA.

The Assignee is responsible for ensuring the validation happens but not necessarily for doing all the validation.

RC1 Validation
bits: https://gcsweb.istio.io/gcs/istio-release/releases/1.4.0-alpha.0/
docs: https://preliminary.istio.io/

- Is Kiali Installed?
- Is the Kiali version correct?
- Does this command work: > istioctl dashboard kiali 
- Did a smoke test of Kiali pass?
- Did a smoke test of Kiali pass using mixer-less telemetry?

Notes:
[jshaughn] There is some slack convo about potential breakage/oom issues in the RC, waiting before installing.

RC2 Validation

- Is Kiali Installed?
- Is the Kiali version correct?
- Does this command work: > istioctl dashboard kiali 
- Did a smoke test of Kiali pass?
- Did a smoke test of Kiali pass using mixer-less telemetry?

jmazzitelli:
Installing Istio 1.4 using istioctl via new hack script found here: https://github.com/kiali/kiali/pull/1853

Istio not starting on my OS 4.2 cluster, so I am not getting as far as to even test kiali:

```
$ oc get pods
NAME                                      READY   STATUS              RESTARTS   AGE
grafana-6b58d744b6-95rdc                  1/1     Running             0          10m
istio-citadel-5d79889c58-txkvl            1/1     Running             0          10m
istio-egressgateway-7645d99dcf-r22zf      0/1     Running             0          10m
istio-galley-95c8c794f-d6pq2              0/1     ContainerCreating   0          10m
istio-ingressgateway-557898dff9-xb888     0/1     Running             0          10m
istio-pilot-7457565f64-6bzt2              0/1     Running             0          10m
istio-policy-994d94c66-d9t46              0/1     CrashLoopBackOff    7          10m
istio-sidecar-injector-866669d699-jzbds   0/1     ContainerCreating   0          10m
istio-telemetry-d76c7b87d-2twhr           0/1     CrashLoopBackOff    7          10m
istio-tracing-86cc56bcd4-xfc5b            1/1     Running             0          10m
kiali-567b6cbfdb-rnzs7                    1/1     Running             0          10m
prometheus-7c84fcfc59-krgvp               1/1     Running             0          10m
```

jmazzitelli:
[...]

Issue #1847, opened 2019-10-28
Support MUTUAL tls method in wizard


lucasponce:
Goal: 
Add support in Kiali routing wizards for mutual tls method (NOT the ISTIO_MUTUAL which is already supported).

Problem: 
Kiali doesn’t have the necessary support for configuring tls mutual method in the routing wizards.

Why is this important: 
Providing a convenient UI to add security configuration for users who choose to use other non default to Istio methods is important for user experience and for securing Istio and the mesh.

PR #1870, opened 2019-11-01
Fix wizards for Istio 1.4

israel-hdez:
The JSON coming from kiali-ui may not be "cleaned" from unset properties and Istio 1.4 seems to have more strict schemas for its CRDs.

This is ensuring that a "clean" JSON is used. Only CRDs involved in Kiali Wizards are fixed.

Related to kiali/kiali#1838, istio/istio#18558.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -587,16 +587,15 @@ func (in *IstioConfigService) ParseJsonForCreate(resourceType, subresourceType s
 	} else {
 		kind = kubernetes.PluralType[resourceType]
 	}
+	var pivot interface{}
 	switch resourceType {
 	case Gateways:
 		istioConfigDetail.Gateway = &models.Gateway{}
 		err = json.Unmarshal(body, istioConfigDetail.Gateway)
 	case VirtualServices:
-		istioConfigDetail.VirtualService = &models.VirtualService{}
-		err = json.Unmarshal(body, istioConfigDetail.VirtualService)
+		pivot = &models.VirtualService{}
 	case DestinationRules:
-		istioConfigDetail.DestinationRule = &models.DestinationRule{}
-		err = json.Unmarshal(body, istioConfigDetail.DestinationRule)
+		pivot = &models.DestinationRule{}
 	case ServiceEntries:
 		istioConfigDetail.ServiceEntry = &models.ServiceEntry{}
 		err = json.Unmarshal(body, istioConfigDetail.ServiceEntry)
@@ -633,11 +632,27 @@ func (in *IstioConfigService) ParseJsonForCreate(resourceType, subresourceType s
 	default:
 		err = fmt.Errorf("object type not found: %v", resourceType)
 	}
+	if pivot != nil {
+		var marshalledBytes []byte
+
+		// Unmarshall, just to validate JSON sent by client.
+		err = json.Unmarshal(body, pivot)
+		if err == nil {
+			// Re-marshall, to get the cleaned JSON.
+			marshalledBytes, err = json.Marshal(pivot)
+		}
+		if err == nil {
+			// Use the cleaned JSON to create the CRD.
+			marshalled = string(marshalledBytes)
+		}
+	} else {
+		// Use JSON from client as is (assume it's already validated in the above switch).
+		marshalled = string(body)
+	}
 	if err != nil {
 		return "", err
 	}
 	// Append apiVersion and kind
-	marshalled = string(body)
 	marshalled = strings.TrimSpace(marshalled)
 	marshalled = "" +
 		"{\n" +
--- a/models/destination_rule.go
+++ b/models/destination_rule.go
@@ -29,7 +29,7 @@ type DestinationRule struct {
 	Metadata meta_v1.ObjectMeta `json:"metadata"`
 	Spec     struct {
 		Host          interface{} `json:"host"`
-		TrafficPolicy interface{} `json:"trafficPolicy"`
+		TrafficPolicy interface{} `json:"trafficPolicy,omitempty"`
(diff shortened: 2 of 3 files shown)


israel-hdez:
It seems to fix the wizards for Istio1.4. Although, I don't know if this is the right way to do the fix.
I think we should still decide to either fix Kiali or wait for Istio people.

In case we merge this, then we will need to decide to either backport to Kiali1.4, or to bump the kiali version shipped with Istio1.4.

jmazzitelli:
Before we merge, I want to wait to hear from @lucasponce at least... let's see what he thinks of this fix.

business/istio_config.go
```diff
@@ -587,16 +587,15 @@ func (in *IstioConfigService) ParseJsonForCreate(resourceType, subresourceType s
 	} else {
 		kind = kubernetes.PluralType[resourceType]
 	}
+	var pivot interface{}
 	switch resourceType {
 	case Gateways:
 		istioConfigDetail.Gateway = &models.Gateway{}
 		err = json.Unmarshal(body, istioConfigDetail.Gateway)
```
lucasponce: Note that Wizards also create Gateways, so same issue would happen there as well.

business/istio_config.go
```diff
@@ -633,11 +632,27 @@ func (in *IstioConfigService) ParseJsonForCreate(resourceType, subresourceType s
...
 		err = fmt.Errorf("object type not found: %v", resourceType)
 	}
+	if pivot != nil {
+		var marshalledBytes []byte
+
+		// Unmarshall, just to validate JSON sent by client.
+		err = json.Unmarshal(body, pivot)
+		if err == nil {
+			// Re-marshall, to get the cleaned JSON.
+			marshalledBytes, err = json.Marshal(pivot)
```
lucasponce: Mmm I haven't tested locally. The re-marshall procedure would work in all the cases ?
I guess this should be also aligned with the json omitempty.

Also I guess if it may have some implication in the edit page of the config, I guess not, but worth to test as well.
I mean, once that Wizard is created, resources should be available in the Istio Config section and editable.

jshaughn:
@israel-hdez @lucasponce , So, is it our feeling that this is our issue and not Istio?   As Edgar mentions above, if it is our fix then we need an updated version with the fix, that is then applied to Istio 1.4.

israel-hdez:
@jshaughn I lean towards fixing Kiali. The Istio.14 CRDs are stricter than in Istio1.3 and I think there is nothing wrong with this. This is why I think the issue is in Kiali.

jshaughn:
@lucasponce, @israel-hdez , OK, I agree that there really isn't anything for Istio to change here, this is just fallout that we need to deal with.  Lucas, the thinking for Saturn is we fix this, perform a mid-sprint 1.9 release, and use it for upstream's 1.4 release.  Let me know if you are OK with this approach.

lucasponce:
If we are in time, I'd prefer to release an upstream Kiali 1.9, rather than backport to Kiali 1.4.

As I commented to @israel-hdez the fix may need more check and work as we need to check more features related, it may hide some side effects in the editor.

I'll help to review/check it as well.

jshaughn:
@lucasponce We are waiting on a new 1.4 beta with a fixed installer (a problem that @jmazzitelli reported to them).  When we have it more people can try to qualify our master against 1.4.  When this fix goes in we can also involve QE.

jmazzitelli:
> I'd prefer to release an upstream Kiali 1.9, rather than backport to Kiali 1.4.

+1

israel-hdez:
@lucasponce I've tried the gateway creation in the wizards and it's not affected.
The editor also seems to work OK. So, I see no changes required.

business/istio_config.go
israel-hdez: I tested and seems to work OK. I guess it's because the wizards don't try to create `null` attributes.

business/istio_config.go
israel-hdez: I just did a small test on the edit page. Both VirtualServices and DestinationRules update correctly in the editor.

lucasponce:
@israel-hdez that's great, I'm also coding #1847 which is going to touch that same area, so I can check the fix and merge if it looks ok.

jmazzitelli:
Let's be clear with version numbers since "1.4" is overloaded. There is a new upstream Istio 1.4 coming... which today uses Kiali 1.4

We are thinking of releasing a new Kiali and replacing the Kiali 1.4 with the new Kiali release inside Istio 1.4 (so Istio 1.4 will have Kiali 1.9 - this  is the hope, at least, if we can do this in time).

lucasponce:
@hhovsepy can run a regression test on it, but I've merged in order to not block any other work related to 1.4.

jshaughn:
If @hhovsepy can validate he regression fix that will be great.  When validated we can evaluate readiness for a 1.9 release for Istio 1.4.

hhovsepy:
@jshaughn @lucasponce trying to install Istio 1.4 by the hack script from PR#1853, facing issues with CNI on my local env. If anyone has a setup Istio 1.4 on OCP then the regression check will take much faster.

jmazzitelli:
@hhovsepy you need to get 1.4-beta2 in order for CNI to work. it will not work on 1.4.beta1

lucasponce:
@israel-hdez I'm revisiting this one and unfortunately it has some side effects, I reviewed too fast.
In short: null values can't be ommited in all cases are those are used in updates.

I'm working in a fix in the context of https://github.com/kiali/kiali/issues/1883 that probably will touch this fix.

hhovsepy:
I confirm that issue https://github.com/kiali/kiali/issues/1883 also happens on Maistra.

Issue #1564, opened 2019-08-27
Allow to configure which annotations should be displayed per entity in UI


lucasponce:
Allow to configure which annotations should be displayed per entity in UI.
Based on discussion here
https://github.com/kiali/kiali/issues/1299

JIRA:
https://issues.jboss.org/browse/KIALI-3192

jotak:
I'm thinking about a new config, that would look like this:

```yaml
- entityAnnotations:
  - name: "mydomain/git-sha"
    displayName: "GIT SHA"
  - name: "mydomain/ci-link"
    displayName: "Build URL"
```

Reading the initial request from @omerlh ( #1299 )  it seems obvious that we'll want links in some cases, so either we auto-detect URLs (starts with http) and always display them as link, or we add an extra config flag such as `showAsLink: true` within the annotation config. I would rather go first 1st option.

That would give something like this: 

![Capture d’écran de 2019-11-04 11-53-11](https://user-images.githubusercontent.com/2153442/68115975-f5766680-fef9-11e9-83f7-77192517ab1c.png)

jmazzitelli:
> ```yaml
> - entityAnnotations:
> ```

I would call it something else - just saying "entityAnnotations" doesn't give any indication what this setting does.

A better name would be something like:

```yaml
annotations_to_display:
- name: "mydomain/git-sha"
  display_name: "GIT SHA"
```

What about labels? Is someone going to want to restrict the labels we display? If so, how about:

```yaml
for_display:
  annotations:
  - name: "mydomain/git-sha"
    display_name: "GIT SHA"
  labels:
  - name: app
    display_name: Application
```

I actually don't like the idea of restricting labels, but **if** we think people are going to want this, prepare for it now by creating that outer setting (`for_display` or something like it) so we can put both annotations and labels under it.

lucasponce:
@jotak also I guess it would need some kind of regex to restrict show all "kiali*", o simply to show all/nothing.

About labels, I'm not sure that we want to restrict them, actually, the UI covers that use cases with the "more labels" (not sure if I like :-), but it was proposed by UX). Hiding labels can create more problems when debugging (why I don't see that pod if it was included in the selector), but annotations are not used in selectors and it's good to restrict what can be seen or not.

In any case, if in the future, that changes, it's good to adjust the config's structure in that way, just in case.

jotak:
@jmazzitelli ok for the naming, I agree that something like `annotations_for_display` is more meaningful.

[...]

PR #1872, opened 2019-11-04
Add annotations for display config

jotak:
- Add related annotation values in Workload and Service models
- Some refactoring to mutualize meta-related conversions in all workload kinds
- Add some tests

Part of https://github.com/kiali/kiali/issues/1564
Frontend PR: https://github.com/kiali/kiali-ui/pull/1527

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -378,6 +378,16 @@ spec:
 
 ##########
 #  ---
+#  additional_display_details: []
+#
+# A list of additional details that Kiali will look for in annotations and display, for every workloads and services, in their respective details pages.
+# Each item in the list is an object with "annotation" and "title" fields to indicate which annotation is to be displayed and how it should be displayed.*
+# For example: 
+#    - annotation: "domain/my-annotation"
+#      title: "My Annotation"
+
+##########
+#  ---
 #  identity:
 #
 # Certificate file used to identify the file server. If set, you must go over https to access Kiali.
--- a/operator/roles/kiali-deploy/defaults/main.yml
+++ b/operator/roles/kiali-deploy/defaults/main.yml
@@ -12,6 +12,8 @@ kiali_defaults:
   istio_component_namespaces: {}
   istio_namespace: ""
 
+  additional_display_details: []
+
   api:
     namespaces:
       exclude:
--- a/operator/roles/kiali-deploy/vars/main.yml
+++ b/operator/roles/kiali-deploy/vars/main.yml
@@ -17,6 +17,13 @@ kiali_vars:
   istio_component_namespaces: "{{ istio_component_namespaces | default(kiali_defaults.istio_component_namespaces) }}"
   istio_namespace: "{{ istio_namespace | default(kiali_defaults.istio_namespace) }}"
 
+  additional_display_details: |
+    {%- if additional_display_details is defined and additional_display_details is iterable -%}
+    {{ additional_display_details }}
+    {%- else -%}
+    {{ kiali_defaults.additional_display_details }}
+    {%- endif -%}
+
   api: |
     {%- if api is defined and api is iterable -%}
     {{ kiali_defaults.api | combine(api, recursive=True) }}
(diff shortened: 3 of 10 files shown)


operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -376,6 +376,16 @@ spec:
...
 #      url: ""
 
+##########
+#  ---
+#  for_display:
+#
+# A list of annotations that Kiali will look for and display, for every workloads and services, in their respective details pages.
+#    ---
+#    annotations:
+#    - display_name: "My Annotation"
```
jmazzitelli: This cr .yaml file is used to document the default settings. This "My Annotation" is not a default (the default is empty list). So change this so it says `annotations: []`. You can put examples in the description text above. So something like:

```
##########
#  ---
#  for_display:
#
# A list of annotations that Kiali will look for and display, for every workload and service, in their respective details
# pages. Each item in the list is a dict that has "name" and "display_name" to indicate which annotation is to be displayed
# and how it should be displayed. For example: 
# - name: "domain/my-annotation"
#   display_name: "My Annotation"
#
#    ---
#    annotations: []
```

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -117,6 +117,9 @@ kiali_defaults:
       service: ""
       url: ""
 
+  for_display:
+    annotations:
```
jmazzitelli: I believe you are going to need to explicitly put an empty list here. This should be:
```suggestion
    annotations: []
```

jmazzitelli:
I must say, even though it was my initial thought, I don't like the name `for_display` for this new top level setteing :) Someone think of a better name.

jmazzitelli:
I forgot in my last review, since you are adding a new top-level group of settings ("for_display" - or hopefully a new name :), you need to add a new group to the vars/main.yaml. Look in there and just copy-n-paste-and-tweek.. you'll know what to do when you look in there ;) See: https://github.com/kiali/kiali/blob/master/operator/roles/kiali-deploy/vars/main.yml

jotak:
ok @jmazzitelli 
About the name, let's try to figure out what this feature is exactly / how we could best describe it. To me, it's about enriching / augmenting the information (meta data) associated with the services and the workloads. The fact that it's based on annotations is more an implementation detail. So what about something like:

```yaml
  entities_extra_details:
  - annotation: "domain/my-annotation"
    display_name: "My annotation"
```

Moreover, that would leave a place for further enhancement, e.g. if one day we have the same based on labels or anything else, we can replace `annotation` with something else within the same structure.

Now, wdyt of `entities_extra_details` ? Any better name?

jmazzitelli:
> Now, wdyt of `entities_extra_details` ? Any better name?

I think my hesitation on the name is the following: this is a config option for display in the UI only. It is really specific/used only in our browser GUI client (it is not configuring anything on the server-side - or... is it?? Does it affect our API in any significant way?).

So I wanted to somehow make it clear in the name that this is solely for display purposes only. I'm not sure how to do that.

`additional_display_details` ???

operator/roles/kiali-deploy/vars/main.yml
```diff
@@ -17,6 +17,13 @@ kiali_vars:
   istio_component_namespaces: "{{ istio_component_namespaces | default(kiali_defaults.istio_component_namespaces) }}"
   istio_namespace: "{{ istio_namespace | default(kiali_defaults.istio_namespace) }}"
 
+  additional_display_details: |
+    {%- if additional_display_details is defined and additional_display_details is iterable -%}
+    {{ additional_display_details }}
```
jotak: @jmazzitelli I didn't do a dummy copy'n paste here because it was generating an error in kiali operator. Maybe it's due to `combine` function, here `additional_display_details` is an array, unlike the other examples where it's all objects. Might explain the difference? Removing `combine` (hence *not* combining the default and the override) works ; as the default here is an empty array it shouldn't be a problem anyway, however if some day we change the default then it's going to be a problem.

jotak:
@xeviknal / @hhovsepy for testing you need to add a config in Kiali and add some annotations on services and/or workloads.
E.g.
Kiali Config:

```yaml
  additional_display_details:
    - annotation: my-annotation
      title: My annotation
```

Workload/service:

```yaml
#...
annotations:
  my-annotation: 'some value'
```

PR #1881, opened 2019-11-05
Remove legacy adapters/templates

### LINKED ISSUE #1880: Istio 1.4: Update old Rules/Adapters/Templates format
In Istio 1.4 seems that the old legacy adapters/templates are gone and Kiali should be updated.

lucasponce:
Istio 1.4 has removed old legacy adapters API.

This should be enough to minimally Kiali supports Istio 1.4 without errors.

Fixes #1880


hhovsepy:
Full regression test on Maistra passed, only already logged issue caused by other commit is happening: https://github.com/kiali/kiali/issues/1883

PR #1888, opened 2019-11-06
Remove null values from JSON on Create operations

### LINKED ISSUE #1883: Routing Wizard - TLS non Disabled - Fails
In any type of Traffic Routing Wizard, if you choose TLS something different that DISABLED, it fails to create the DR:
spec.trafficPolicy.loadBalancer in body must be of type object: "null"

Screenshot from 2019-11-06 09-39-55

lucasponce:
Fixes #1883 

This is a continuation of #1870. Main issue in Kiali is that our model in backend doesn't map entire types, so Marshall/Unmarshall didn't work in some use cases.

Also, this is not trivial, as empty values are valid when using patches, so UI will continue sending them.
But now with this patch, all JSON on CREATE operations should be cleaned from nil values if whole type is mapped or not.

Note that Golang has not a "omitempty" tag for generic.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -14,6 +14,7 @@ import (
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
+	"github.com/kiali/kiali/util"
 )
 
 type IstioConfigService struct {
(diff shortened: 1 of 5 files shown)


lucasponce:
> BTW, I understand that the Unmarshalls done inside the switch are validating that the received data complies with the schema. I’m thinking that, probably, we could now omit that schema validation on Kiali side, since the cluster will be doing it for us. This will remove the need to keep up to date those structures on Kiali codebase. Commenting for consideration in the future (perhaps when 1.3 reaches EOL).

That's correct, probably with Galley and schema validation, that would be enough, too.

Issue #1884, opened 2019-11-06
False validation in DestinationRule


lucasponce:
An error marked in the validations for a service that exists:
![image](https://user-images.githubusercontent.com/1662329/68283292-94769c00-007b-11ea-8b68-2f8f7ccbbc13.png)

![image](https://user-images.githubusercontent.com/1662329/68283343-aa845c80-007b-11ea-873b-ea71fbd08e19.png)

PR #1889, opened 2019-11-06
Covering scenario where host in DR is formatted as follows: svc.ns

xeviknal:
** Describe the change **
DestinationRules might have a host in the format `servicename.namespace` without the domain.

** Issue reference **
Fixes https://github.com/kiali/kiali/issues/1884

** Backwards incompatible? **
yes

### DIFF
--- a/business/checkers/destinationrules/no_dest_checker.go
+++ b/business/checkers/destinationrules/no_dest_checker.go
@@ -77,11 +77,18 @@ func (n NoDestinationChecker) hasMatchingWorkload(service string, subsetLabels m
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


hhovsepy:
Solution works.
But here comes another thing to resolve with 2 DRs in the same namespace:
DR1 host: reviews
DR2 host: reviews.bookinfo
Both refer to the same host, but the warning message "More than one Destination Rule ..." is not displayed.
Will it be fixed within this PR?

lucasponce:
@hhovsepy is right, perhaps internally we need to manage canonical representation of services, and always working with that.

PR #1916, opened 2019-11-11
Return 403 instead 500 for accessible_namespaces filtered queries

### LINKED ISSUE #1906: Incorrect Error Code Returned
There are a few API automation tests that fail when run against the master branch due to the unexpected error codes being returned for some negative tests. The three examples below return the error of 404 in the 1.0 branch, though in master they return an error 500.

api/namespaces/invalid/workloads/details-v1
api/namespaces/invalid/apps/ratings
api/namespaces/invalid/istio/rules/promtcp

E2E Test:  test_api_methods -> test_engative_404

lucasponce:
Ignore a namespace included in accessible_namespaces that does not exist

Fix #1906

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -5,10 +5,11 @@ import (
 
 	osproject_v1 "github.com/openshift/api/project/v1"
 	core_v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 
-	"fmt"
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
 )
@@ -18,6 +19,19 @@ type NamespaceService struct {
 	k8s                    kubernetes.IstioClientInterface
 	hasProjects            bool
 	isAccessibleNamespaces map[string]bool
+}
+
+type AccessibleNamespaceError struct {
+	msg string
+}
+
+func (in *AccessibleNamespaceError) Error() string {
+	return in.msg
+}
+
+func IsAccessibleError(err error) bool {
+	_, isAccessibleError := err.(*AccessibleNamespaceError)
+	return isAccessibleError
 }
 
 func NewNamespaceService(k8s kubernetes.IstioClientInterface) NamespaceService {
@@ -79,13 +93,19 @@ func (in *NamespaceService) GetNamespaces() ([]models.Namespace, error) {
 			}
 			namespaces = models.CastNamespaceCollection(nss)
 		} else {
-			k8sNamespaces := make([]core_v1.Namespace, len(accessibleNamespaces))
-			for i, ans := range accessibleNamespaces {
+			k8sNamespaces := make([]core_v1.Namespace, 0)
+			for _, ans := range accessibleNamespaces {
 				k8sNs, err := in.k8s.GetNamespace(ans)
 				if err != nil {
-					return nil, err
+					// If a namespace is not found, then we skip it from the list of namespaces
+					if errors.IsNotFound(err) {
+						log.Warningf("Kiali has an accessible namespace [%s] which doesn't exist", ans)
+					} else {
+						return nil, err
+					}
+				} else {
+					k8sNamespaces = append(k8sNamespaces, *k8sNs)
 				}
-				k8sNamespaces[i] = *k8sNs
 			}
 			namespaces = models.CastNamespaceCollection(k8sNamespaces)
 		}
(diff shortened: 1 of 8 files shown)


lucasponce:
@mattmahoneyrh if you can test this one against the i-tests and validate it looks ok, that would be great.

Thanks.

jshaughn:
@mattmahoneyrh Yes, I think that is the new, anticipated status code.  403 means Forbidden.  We can't distinguish between a namespace that doesn't exist or is simply inaccessible.  So 403/Forbidden seems to make sense.

mattmahoneyrh:
@ jshaughn, that's fine - just sanity checking - good to merge from my testing.

lucasponce:
@jshaughn thanks for the feedback.

> /api/namespaces/invalid/health. That one still returns a 500.

Let me address this, as this one fetches from p8s perhaps something else was missing in my fix.

> This is probably unrelated and maybe has to do with cache refresh? But when sitting on the Overview page, and configuring a new, valid namespace, it generates errors for a while before settling down. For example, I added bookinfo and the bookinfo card appeared, but it was unable to fill in the card for about 30-45 seconds I think. During that time I repeatedly got the same toast messages [...]

I'm going to investigate this. Can you give me more details ? Did you enabled caching ? In which environment you tested ? And, what steps did you follow to add a new namespace (I guess you added into the ServiceMeshMemberRoll entry, and next you changed the Kiali CR) ?

I'm going to try to reproduce it locally. I will update this thread.

lucasponce:
I've address the health enpoints.

About your last comment @jshaughn I'd like to investigate more, but I need more details. 
Caching is disabled by default, so unless you explicictly enabled it, we may be hitting a different thing here (i.e. wondering if it's the case where Kiali pod is being recreated and the UI is again constantly hitting).

Perhaps one thing we can do in Kiali is to have a general poll to kiali backend, if that /status query doesn't respond, just to block all queries to the backend and re-enable them when the basic health is up again.

Then, Kiali UI would handle nicely when pod is down like "backend is not available" instead to fill the UI with error messages.

This is not a minor thing so it would require a separate issue as it would need a non small refactoring in the code.

jshaughn:
@lucasponce OK, it doesn't sound like caching, then, because I did not enable anything.  Basically I did the following:
1. created OS 4.2 env on AWS with Istio 1.4
2. build/push/install Kiali operator/kiali built from master

At this point bookinfo doesn't exist and is not an accessible namespace

3, install bookinfo (but I think this could just be the creation of any new namespace)
4. leave kiali UI refreshing at 10s interval on Overview page
5. edit Kiali CR and add:
```
deployment:
  accessible_namespaces:
  - default
  - istio-system
  - bookinfo
```

Go back and watch overview page while kiali operator applies the CR change.  Bookinfo card appears but is not filled in, errors generate/repeat for a while, then it settles down.

lucasponce:
Ok, @jshaughn that info is good to start, I suspect where it could be

lucasponce:
@jshaughn ok, I think I found the problem.
accessible_namespaces is not used in OpenShift environments for the GetNamespaces() query.
But we use to filter the access in general.

So, it looks there is a mismatch here, that affects in scenarios as you point.

jshaughn:
> accessible_namespaces is not used in OpenShift environments for the GetNamespaces() query.

interesting

lucasponce:
If userA can see ns1, ns2 and ns3 projects, but accessible_namespaces is ns1, and ns2 only.
Then GetNamespaces() returns [ns1, ns2 and n3] BUT a GetNamespace() is filtering the accessible_namespaces.

So, on this case, probably a ns3 based query can lead in the issues you are facing there.

lucasponce:
I understand that how this was designed, was that if ** is set, then all ns1, ns2 and ns3 should be shown. But if accessible_namespaces is set, it should also be filtered by that.

Am I right here @jshaughn ?

jmazzitelli:
> (I guess you added into the ServiceMeshMemberRoll entry, and next you changed the Kiali CR) ?

FYI: you should never do this. If you are using Kiali with Service Mesh, you should never edit Kiali CR. You ONLY edit ServiceMeshMemeberRole CR and ServiceMesh will manage Kiali CR for you (just like it manages any other resource Service Mesh operator creates like Deployments, ConfigMaps, etc). If you edit Kiali CR, you will clash with the ServiceMesh operator making its own edits in the Kiali CR. The only time you edit Kiali CR yourself is if you installed Kiali operator yourself (i.e. it was not a dependency of another operator) -  such as is the case with upstream Istio today. If you install upstream Istio and you install Kiali operator, then you can edit the Kiali CR.

> I understand that how this was designed, was that if ** is set, then all ns1, ns2 and ns3 should be shown. But if accessible_namespaces is set, it should also be filtered by that.

If accessible_namespaces is set to "**" then it is assumed Kiali has been given cluster role access to every namespace that exists, and new namespaces that might be created in the future. If accessible_namespaces is set to something other than **, then only those namespaces are assumed to be accessible to Kiali so only those should be shown (if they do not exist, Kiali should ignore them -- but I actually thought the Kiali Operator will strip those out and not put them in accessible_namespaces (see [this code](https://github.com/kiali/kiali/blob/v1.9.1/operator/roles/kiali-deploy/tasks/main.yml#L332-L344)) - but there is nothing stopping a user from editing the ConfigMap and putting bad/non-existence namespaces in the ConfigMap - this is another reason why operators are good - they help to ensure correct/valid data only shows up in the ConfigMap.

Issue #1757, opened 2019-10-03
[operator] need to provide schema definitions to CRDs


jmazzitelli:
Kubernetes v1.16 introduces api version `apiextensions.k8s.io/v1` of CRDs. The old `v1beta1` is deprecated.

The big change here is that our CRDs ([Kiali](https://github.com/kiali/kiali/blob/v1.5/operator/deploy/crd.yaml#L19-L40) and [MonitoringDashboard](https://github.com/kiali/kiali/blob/v1.5/operator/deploy/crd.yaml#L2-L17)) must now have a definition of a structural schema - this is mandatory for CRDs of version `apiextensions.k8s.io/v1`

So we need to get ready to bump up our versions of our CRDs and add schema definitions to them. This is going to probably involve lots of changes across different places (think OLM and the operator manifest definitions for one example).

For more, read these:

* How to migrate an operator to a new CRD: https://github.com/operator-framework/operator-sdk/blob/0481703cbf7ae25b0862f14ff4a4f5b652632897/doc/user/migrating-existing-apis.md
* Upgrading CRDs within OLM: https://github.com/operator-framework/operator-lifecycle-manager/blob/0.12.0/doc/design/dependency-resolution.md#customresourcedefinition-crd-upgrade
* How OLM upgrades work: https://github.com/operator-framework/operator-lifecycle-manager/blob/0.12.0/doc/design/how-to-update-operators.md
* How to specify a structural schema in a CRD: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#specifying-a-structural-schema

jotak:
Ok, for monitoringdashboard it was kept schema-less which I think helped quite a lot at the beginning to let time for the API to stabilize. Now it's time to change.

jotak:
For the record, I'm putting this as a gist but haven't had the chance to test and verify yet as I need to upgrade my k8s : https://gist.github.com/jotak/e335ecc6b19229b1920c55897195d1a4

lucasponce:
I'd removed from Sprint 32 as it's waiting external.
It can be reviewed and added in Sprint 32 if it gets unblocked and it can progress.

jotak:
@israel-hdez @jmazzitelli is this task only for Kiali CR or also for monitoring dashboards CR? (put differently, should I create a separate task?)

jmazzitelli:
[...]

PR #1920, opened 2019-11-11
[operator] Add schema to Kiali operator CRD

israel-hdez:
This is adding a schema to the Kiali operator CRD.

Related to kiali/kiali#1757

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -48,7 +48,7 @@ const (
 
 	// These constants are used for external services auth (Prometheus, Grafana ...) ; not for Kiali auth
 	// If you need to add more, make sure to update the schema of the operator crd:
-	//   $/operator/deploy/crd.yaml, the deployment.external_services property
+	//   (i.e. $/operator/deploy/crd.yaml and the CRDs defined in the OLM operator manifests), the deployment.external_services property
 	AuthTypeBasic  = "basic"
 	AuthTypeBearer = "bearer"
 	AuthTypeNone   = "none"
--- a/config/config.go
+++ b/config/config.go
@@ -47,6 +47,8 @@ const (
 	AuthStrategyTokenIssuer     = "kiali-token"
 
 	// These constants are used for external services auth (Prometheus, Grafana ...) ; not for Kiali auth
+	// If you need to add more, make sure to update the schema of the operator crd:
+	//   $/operator/deploy/crd.yaml, the deployment.external_services property
 	AuthTypeBasic  = "basic"
 	AuthTypeBearer = "bearer"
 	AuthTypeNone   = "none"
--- a/operator/deploy/crd.yaml
+++ b/operator/deploy/crd.yaml
@@ -38,3 +38,309 @@ spec:
   - name: v1alpha1
     served: true
     storage: true
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        spec:
+          type: object
+          properties:
+            installation_tag:
+              type: string
+            istio_component_namespaces:
+              type: object
+            istio_namespace:
+              type: string
+            api:
+              type: object
+              properties:
+                namespaces:
+                  type: object
+                  properties:
+                    exclude:
+                      type: array
+                      items:
+                        type: string
+                label_selector:
+                  type: string
+            apidocs:
+              type: object
+              properties:
+                annotations:
+                  type: object
+                  properties:
+                    api_spec_annotation_name:
+                      type: string
+                    api_type_annotation_name:
+                      type: string
+            auth:
+              type: object
+              properties:
+                strategy:
+                  type: string
+                ldap:
+                  type: object
+                  properties:
+                    ldap_base:
+                      type: string
+                    ldap_bind_dn:
+                      type: string
+                    ldap_group_filter:
+                      type: string
+                    ldap_host:
+                      type: string
+                    ldap_insecure_skip_verify:
+                      type: boolean
+                    ldap_mail_id_key:
+                      type: string
+                    ldap_member_of_key:
+                      type: string
+                    ldap_port:
+                      type: number
+                    ldap_role_filter:
+                      type: string
+                    ldap_search_filter:
+                      type: string
+                    ldap_use_ssl:
+                      type: boolean
+                    ldap_user_filter:
+                      type: string
+                    ldap_user_id_key:
+                      type: string
+            deployment:
+              type: object
+              properties:
+                accessible_namespaces:
+                  type: array
+                  items:
+                    type: string
+                additional_service_yaml:
+                  type: object
+                affinity:
+                  type: object
+                  properties:
+                    node:
+                      type: object
+                    pod:
+                      type: object
+                    pod_anti:
+                      type: object
+                image_name:
+                  type: string
+                image_pull_policy:
+                  type: string
+                  enum:
+                  - Always
+                  - Never
+                  - IfNotPresent
+                image_pull_secrets:
+                  type: array
+                  items:
+                    type: string
+                image_version:
+                  type: string
+                ingress_enabled:
+                  type: boolean
+                namespace:
+                  type: string
+                secret_name:
+                  type: string
+                service_type:
+                  type: string
+                  enum:
+                  - NodePort
+                  - ClusterIP
+                  - LoadBalancer
+                  - ExternalName
+                tolerations:
+                  type: array
+                  items:
+                    type: object
+                verbose_mode:
+                  type: number
+                  minimum: 0
+                version_label:
+                  type: string
+                view_only_mode:
+                  type: boolean
+            external_services:
+              type: object
+              properties:
+                grafana:
+                  type: object
+                  properties:
+                    auth:
+                      type: object
+                      properties:
+                        ca_file:
+                          type: string
+                        insecure_skip_verify:
+                          type: boolean
+                        password:
+                          type: string
+                        token:
+                          type: string
+                        type:
+                          type: string
+                          enum:
+                          - basic
+                          - bearer
+                          - none
+                        use_kiali_token:
+                          type: boolean
+                        username:
+                          type: string
+                    dashboards:
+                      type: array
+                      items:
+                        type: object
+                    enabled:
+                      type: boolean
+                    in_cluster_url:
+                      type: string
+                    url:
+                      type: string
+                istio:
+                  type: object
+                  properties:
+                    istio_identity_domain:
+                      type: string
+                    istio_sidecar_annotation:
+                      type: string
+                    url_service_version:
+                      type: string
+                prometheus:
+                  type: object
+                  properties:
+                    auth:
+                      type: object
+                      properties:
+                        ca_file:
+                          type: string
+                        insecure_skip_verify:
+                          type: boolean
+                        password:
+                          type: string
+                        token:
+                          type: string
+                        type:
+                          type: string
+                          enum:
+                          - basic
+                          - bearer
+                          - none
+                        use_kiali_token:
+                          type: boolean
+                        username:
+                          type: string
+                    custom_metrics_url:
+                      type: string
+                    url:
+                      type: string
+                tracing:
+                  type: object
+                  properties:
+                    auth:
+                      type: object
+                      properties:
+                        ca_file:
+                          type: string
+                        insecure_skip_verify:
+                          type: boolean
+                        password:
+                          type: string
+                        token:
+                          type: string
+                        type:
+                          type: string
+                          enum:
+                          - basic
+                          - bearer
+                          - none
+                        use_kiali_token:
+                          type: boolean
+                        username:
+                          type: string
+                    enabled:
+                      type: boolean
+                    in_cluster_url:
+                      type: string
+                    namespace:
+                      type: string
+                    port:
+                      type: number
+                      minimum: 1
+                      maximum: 65535
+                    service:
+                      type: string
+                    url:
+                      type: string
+            identity:
+              type: object
+              properties:
+                cert_file:
+                  type: string
+                private_key_file:
+                  type: string
+            istio_labels:
+              type: object
+              properties:
+                app_label_name:
+                  type: string
+                version_label_name:
+                  type: string
+            kubernetes_config:
+              type: object
+              properties:
+                burst:
+                  type: number
+                cache_duration:
+                  type: number
+                cache_enabled:
+                  type: boolean
+                cache_istio_types:
+                  type: array
+                  items:
+                    type: string
+                cache_namespaces:
+                  type: array
+                  items:
+                    type: string
+                cache_token_namespace_duration:
+                  type: number
+                excluded_workloads:
+                  type: array
+                  items:
+                    type: string
+                qps:
+                  type: number
+            login_token:
+              type: object
+              properties:
+                expiration_seconds:
+                  type: number
+                signing_key:
+                  type: string
+            server:
+              type: object
+              properties:
+                address:
+                  type: string
+                audit_log:
+                  type: boolean
+                cors_allow_all:
+                  type: boolean
+                gzip_enabled:
+                  type: boolean
+                metrics_enabled:
+                  type: boolean
+                metrics_port:
+                  type: number
+                  minimum: 1
+                  maximum: 65535
+                port:
+                  type: number
+                  minimum: 1
+                  maximum: 65535
+                web_root:
+                  type: string
--- a/operator/deploy/crd.yaml
+++ b/operator/deploy/crd.yaml
@@ -143,6 +143,8 @@ spec:
                   type: boolean
                 namespace:
                   type: string
+                resources:
+                  type: object
                 secret_name:
                   type: string
                 service_type:
--- a/operator/manifests/kiali-upstream/kiali.package.yaml
+++ b/operator/manifests/kiali-upstream/kiali.package.yaml
@@ -1,7 +1,7 @@
 packageName: kiali
 channels:
 - name: alpha
-  currentCSV: kiali-operator.v1.11.0
+  currentCSV: kiali-operator.v1.12.0
 - name: stable
-  currentCSV: kiali-operator.v1.11.0
+  currentCSV: kiali-operator.v1.12.0
 defaultChannel: stable
--- a/operator/manifests/README.adoc
+++ b/operator/manifests/README.adoc
@@ -303,7 +303,7 @@ annotations:
 spec:
   deployment:
     namespace: ${CONTROL_PLANE_NAMESPACE}
-    verbose_mode: "4"
+    verbose_mode: 4
 EOF
 ```
 
--- a/operator/manifests/kiali-community/1.12.0/kiali.crd.yaml
+++ b/operator/manifests/kiali-community/1.12.0/kiali.crd.yaml
@@ -124,6 +124,8 @@ spec:
                   type: boolean
                 namespace:
                   type: string
+                resources:
+                  type: object
                 secret_name:
                   type: string
                 service_type:
--- a/operator/manifests/kiali-community/1.12.0/kiali.crd.yaml
+++ b/operator/manifests/kiali-community/1.12.0/kiali.crd.yaml
@@ -1,0 +1,327 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: kialis.kiali.io
+  labels:
+    app: kiali-operator
+spec:
+  group: kiali.io
+  names:
+    kind: Kiali
+    listKind: KialiList
+    plural: kialis
+    singular: kiali
+  scope: Namespaced
+  subresources:
+    status: {}
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        spec:
+          type: object
+          properties:
+            installation_tag:
+              type: string
+            istio_component_namespaces:
+              type: object
+            istio_namespace:
+              type: string
+            api:
+              type: object
+              properties:
+                namespaces:
+                  type: object
+                  properties:
+                    exclude:
+                      type: array
+                      items:
+                        type: string
+                label_selector:
+                  type: string
+            apidocs:
+              type: object
+              properties:
+                annotations:
+                  type: object
+                  properties:
+                    api_spec_annotation_name:
+                      type: string
+                    api_type_annotation_name:
+                      type: string
+            auth:
+              type: object
+              properties:
+                strategy:
+                  type: string
+                ldap:
+                  type: object
+                  properties:
+                    ldap_base:
+                      type: string
+                    ldap_bind_dn:
+                      type: string
+                    ldap_group_filter:
+                      type: string
+                    ldap_host:
+                      type: string
+                    ldap_insecure_skip_verify:
+                      type: boolean
+                    ldap_mail_id_key:
+                      type: string
+                    ldap_member_of_key:
+                      type: string
+                    ldap_port:
+                      type: number
+                    ldap_role_filter:
+                      type: string
+                    ldap_search_filter:
+                      type: string
+                    ldap_use_ssl:
+                      type: boolean
+                    ldap_user_filter:
+                      type: string
+                    ldap_user_id_key:
+                      type: string
+            deployment:
+              type: object
+              properties:
+                accessible_namespaces:
+                  type: array
+                  items:
+                    type: string
+                additional_service_yaml:
+                  type: object
+                affinity:
+                  type: object
+                  properties:
+                    node:
+                      type: object
+                    pod:
+                      type: object
+                    pod_anti:
+                      type: object
+                image_name:
+                  type: string
+                image_pull_policy:
+                  type: string
+                  enum:
+                  - Always
+                  - Never
+                  - IfNotPresent
+                image_pull_secrets:
+                  type: array
+                  items:
+                    type: string
+                image_version:
+                  type: string
+                ingress_enabled:
+                  type: boolean
+                namespace:
+                  type: string
+                secret_name:
+                  type: string
+                service_type:
+                  type: string
+                  enum:
+                  - NodePort
+                  - ClusterIP
+                  - LoadBalancer
+                  - ExternalName
+                tolerations:
+                  type: array
+                  items:
+                    type: object
+                verbose_mode:
+                  type: number
+                  minimum: 0
+                version_label:
+                  type: string
+                view_only_mode:
+                  type: boolean
+            external_services:
+              type: object
+              properties:
+                grafana:
+                  type: object
+                  properties:
+                    auth:
+                      type: object
+                      properties:
+                        ca_file:
+                          type: string
+                        insecure_skip_verify:
+                          type: boolean
+                        password:
+                          type: string
+                        token:
+                          type: string
+                        type:
+                          type: string
+                          enum:
+                          - basic
+                          - bearer
+                          - none
+                        use_kiali_token:
+                          type: boolean
+                        username:
+                          type: string
+                    dashboards:
+                      type: array
+                      items:
+                        type: object
+                    enabled:
+                      type: boolean
+                    in_cluster_url:
+                      type: string
+                    url:
+                      type: string
+                istio:
+                  type: object
+                  properties:
+                    istio_identity_domain:
+                      type: string
+                    istio_sidecar_annotation:
+                      type: string
+                    url_service_version:
+                      type: string
+                prometheus:
+                  type: object
+                  properties:
+                    auth:
+                      type: object
+                      properties:
+                        ca_file:
+                          type: string
+                        insecure_skip_verify:
+                          type: boolean
+                        password:
+                          type: string
+                        token:
+                          type: string
+                        type:
+                          type: string
+                          enum:
+                          - basic
+                          - bearer
+                          - none
+                        use_kiali_token:
+                          type: boolean
+                        username:
+                          type: string
+                    custom_metrics_url:
+                      type: string
+                    url:
+                      type: string
+                tracing:
+                  type: object
+                  properties:
+                    auth:
+                      type: object
+                      properties:
+                        ca_file:
+                          type: string
+                        insecure_skip_verify:
+                          type: boolean
+                        password:
+                          type: string
+                        token:
+                          type: string
+                        type:
+                          type: string
+                          enum:
+                          - basic
+                          - bearer
+                          - none
+                        use_kiali_token:
+                          type: boolean
+                        username:
+                          type: string
+                    enabled:
+                      type: boolean
+                    in_cluster_url:
+                      type: string
+                    namespace:
+                      type: string
+                    port:
+                      type: number
+                      minimum: 1
+                      maximum: 65535
+                    service:
+                      type: string
+                    url:
+                      type: string
+            identity:
+              type: object
+              properties:
+                cert_file:
+                  type: string
+                private_key_file:
+                  type: string
+            istio_labels:
+              type: object
+              properties:
+                app_label_name:
+                  type: string
+                version_label_name:
+                  type: string
+            kubernetes_config:
+              type: object
+              properties:
+                burst:
+                  type: number
+                cache_duration:
+                  type: number
+                cache_enabled:
+                  type: boolean
+                cache_istio_types:
+                  type: array
+                  items:
+                    type: string
+                cache_namespaces:
+                  type: array
+                  items:
+                    type: string
+                cache_token_namespace_duration:
+                  type: number
+                excluded_workloads:
+                  type: array
+                  items:
+                    type: string
+                qps:
+                  type: number
+            login_token:
+              type: object
+              properties:
+                expiration_seconds:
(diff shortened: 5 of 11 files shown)


jmazzitelli:
This is a good start. It begins to address https://github.com/kiali/kiali/issues/1757

There is a lot more that we have to do. Changing the manifests across the OLM landscape (see `operator/manifest` directory) is a big thing.

Before we merge we also have to figure out what needs to change in upstream Istio.

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+  preserveUnknownFields: false
   scope: Namespaced
-  subresources:
-    status: {}
-  version: v1alpha1
   versions:
-  - name: v1alpha1
-    served: true
-    storage: true
+    - name: v1alpha1
```
mtho11: Nice to see the typed schemas. Cuts down on config errors.

jmazzitelli: The array indentation in here looks wrong. All of our other yaml indentations (at least the yaml in the operator) has arrays on the same indentation level as the parent. So:

```yaml
versions:
- name
```

not

```yaml
versions:
  - name
```

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                  type: object
+                  properties:
+                    namespaces:
+                      type: object
+                      properties:
+                        exclude:
+                          type: array
+                          items:
+                            type: string
+                          default:
```
jmazzitelli: somewhere (either here, or maybe in `operator/roles/kiali-deploy/default/main.yaml`) we need a comment to tell people "if you change this default, you also need to change the default found in default/main.yaml AND in the kiali_cr.yaml file"  (and whereever else we define defaults).

Question: do we even need to define the defaults here? Can we avoid this? Otherwise, this is going to be a maintenance problem because now we have to keep the defaults all in sync with this CRD as well as the operator's ansible defaults and the kiali_cr example yaml.

If we have to define the defaults here (or if it just is best practice to do so) then we should remove the notion that kiali_cr defines the defaults and instead in kiali_cr comments we should tell the reader to come here to the CRD to find out what the defaults are (this would eliminate the problem if needing to sync these files).

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                          default: kiali.io/api-type
+                auth:
+                  type: object
+                  properties:
+                    strategy:
+                      type: string
+                      enum:
+                        - login
+                        - openshift
+                        - anonymous
```
jmazzitelli: this is missing `ldap` auth strategy option.

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                        - ExternalName
+                    tolerations:
+                      type: array
+                      items:
+                        type: object
+                        additionalProperties: true
+                    verbose_mode:
+                      type: number
+                      minimum: 0
+                      maximum: 5
```
jmazzitelli: I don't think the maximum is really 5. The glog library allows for anything >= 0. We should allow for any value. We don't know, for example, what go libraries Kiali pulls in that logs messages at a verbose mode higher than 5. So we should not limit to any maximum here in case we need to ask for trace logging at that higher level.

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                        - anonymous
+                    ldap:
+                      type: object
+                      properties:
+                        ldap_base:
+                          type: string
+                        ldap_bind_dn:
+                          type: string
+                        ldap_group_filter:
+                          type: string
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `"(cn=%s)"`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                          type: boolean
+                          default: false
+                        ldap_mail_id_key:
+                          type: string
+                          default: "mail"
+                        ldap_member_of_key:
+                          type: string
+                          default: "memberOf"
+                        ldap_port:
+                          type: number
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `0`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                        ldap_role_filter:
+                          type: string
+                        ldap_search_filter:
+                          type: string
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `"(&(name={USERID}))"`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                        ldap_use_ssl:
+                          type: boolean
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `false`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                        ldap_user_filter:
+                          type: string
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `"(cn=%s)"`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                        ldap_user_id_key:
+                          type: string
+                          default: "cn"
+                deployment:
+                  type: object
+                  properties:
+                    accessible_namespaces:
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `["^((?!(istio-operator|kube.*|openshift.*|ibm.*|kiali-operator)).)*$"]`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                      type: array
+                      items:
+                        type: string
+                    additional_service_yaml:
+                      type: object
+                      additionalProperties: true
+                    affinity:
+                      type: object
+                      properties:
+                        affinity:
```
jmazzitelli: this is the wrong name - this property name is `node`  (`node` is under `affinity` )

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                          additionalProperties: true
+                        pod:
+                          type: object
+                          additionalProperties: true
+                        pod_anti:
+                          type: object
+                          additionalProperties: true
+                    image_name:
+                      type: string
+                    image_pull_policy:
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is "IfNotPresent"

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                      type: array
+                      items:
+                        type: string
+                    image_version:
+                      type: string
+                    ingress_enabled:
+                      type: boolean
+                      default: true
+                    namespace:
+                      type: string
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `istio-system`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                        - NodePort
+                        - ClusterIP
+                        - LoadBalancer
+                        - ExternalName
+                    tolerations:
+                      type: array
+                      items:
+                        type: object
+                        additionalProperties: true
+                    verbose_mode:
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `3`

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                      type: number
+                      minimum: 0
+                      maximum: 5
+                    version_label:
+                      type: string
+                    view_only_mode:
+                      type: boolean
```
jmazzitelli: if we ARE going to put defaults in this schema, then we need to set this default which is `false`

(see what a pain it is to keep defaults in sync here :)) I won't continue with these comments - suffice it to say, there are alot of defaults missing in here. Either we document them all, or we do not document any defaults. But we shouldn't document just some of them)

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                          properties:
+                            ca_file:
+                              type: string
+                            insecure_skip_verify:
+                              type: boolean
+                            password:
+                              type: string
+                            token:
+                              type: string
+                            type:
```
jmazzitelli: type should be an enum - it accepts 3 values - `basic`, `bearer`, and `none` - see: https://github.com/kiali/kiali/blob/v1.9/config/config.go#L47-L50

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
```
jmazzitelli: type should be an enum - it accepts 3 values - `basic`, `bearer`, and `none` - see: https://github.com/kiali/kiali/blob/v1.9/config/config.go#L47-L50

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
```
jmazzitelli: type should be an enum - it accepts 3 values - `basic`, `bearer`, and `none` - see: https://github.com/kiali/kiali/blob/v1.9/config/config.go#L47-L50

operator/deploy/crd.yaml
```diff
@@ -30,11 +30,335 @@ spec:
...
+                      type: number
+                    cache_duration:
+                      type: number
+                    cache_enabled:
+                      type: boolean
+                    cache_istio_types:
+                      type: array
+                      items:
+                        type: string
+                        enum:
```
jmazzitelli: cache_istio_types is not an enum - it is a list of strings - similar to the "excluded_workloads" below. The values here under "enum" is the default value.

operator/deploy/crd.yaml
jmazzitelli: We also want to finally bump up our schema version officially to "v1". I think with this change of adding a full schema, we can say we are no longer in alpha (v1alpha1).

This means we have to investigate what we need to do to properly upgrade CRD versions, particularly within OLM.

jmazzitelli: When we bump up the version to v1, all kiali CR files in operator/deploy/kiali need to be moved up to the new version. For example:

https://github.com/kiali/kiali/blob/v1.9/operator/deploy/kiali/kiali_cr.yaml#L1

should be using `apiVersion: kiali.io/v1`

jmazzitelli: Also, the operator needs to be told to watch for the new version:

https://github.com/kiali/kiali/blob/v1.9/operator/watches.yaml#L2

I think we should still be supporting the v1alpha1 - so we'll need to add the new version to the watches.yaml rather than just replacing the current v1alpha1 version.

israel-hdez: I used the indentation as in this sample: https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#create-a-customresourcedefinition

I'm not sure if there is any difference.

operator/deploy/crd.yaml
israel-hdez: > Question: do we even need to define the defaults here? Can we avoid this?
We don't need the defaults :)
I'll remove them. These are remnants of my first try.

I was putting all defaults I saw and I realized that it was going to be too much work to keep this in sync. It's better to omit the defaults and let the operator do its job of setting those values..

operator/deploy/crd.yaml
jmazzitelli: There isn't a difference so long as a yaml file uses one way or the other.

But all the operator yaml files use indentation of lists on the same level as the parent item. For example, in the current CRD, you will see this:

https://github.com/kiali/kiali/blob/v1.9/operator/deploy/crd.yaml#L38

```yaml
  versions:
  - name: v1alpha1
    served: true
    storage: true
```

And this is how the indentation is everywhere else. So we want to keep this file consistent and use the same indentation as every other yaml file.

operator/deploy/crd.yaml
israel-hdez: The comment in 
https://github.com/kiali/kiali/blob/9dda43d4bc5bac06e1399da8238cc9bfb83ecc82/operator/deploy/kiali/kiali_cr.yaml#L443-L444

gives the idea that these CRDs are the only supported ones. So, I though it was "good" to put the enum :/
By removing this, perhaps that comment should be fixed.

operator/deploy/crd.yaml
israel-hdez: Right, I'll fix all instances.

operator/deploy/crd.yaml
jmazzitelli: Oh, I get it now. I didn't notice this is a list of "items" of enum type.

That said, I think we should not use enum here and just use string. This will allow for future expansion of the cache WITHOUT requiring us to update the version of the CRD.

Remember, now that we have a schema, we are now required to update the version of the CRD each and every time something is added (and it isn't a trivial thing to bump up the CRD version - people will have created Kiali CRs that needs to eventually be modified, Service Mesh operator will be using a specific version of the Kiali CRD when it writes out the Kiali CR, upstream Istio will eventually be using this, and [OLM has rules about CRD updates](https://github.com/operator-framework/operator-lifecycle-manager/blob/0.12.0/doc/design/dependency-resolution.md#customresourcedefinition-crd-upgrade)). It will be a pain to have to bump the CRD version simply because we added a new type that can be cached. That really shouldn't involve a new CRD version being created.

israel-hdez: Hmm I see and I agree. OK, I'll remove the enum.
This makes me think that also the `strategy` shouldn't have the enum.

jmazzitelli: > This makes me think that also the `strategy` shouldn't have the enum.

That is a good point.

israel-hdez:
@jmazzitelli All your comments are fixed. It needs a test to see if the operator upgrade works OK

operator/watches.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-- version: v1alpha1
+- version: v1
```
jmazzitelli: This is going to cause any new operator build to not install Kiali if the user is using v1alpha1 version of Kiali CR. I think we want to duplicate this so it watches for both v1alpha1 and v1. I'm not 100% sure because I have not done this kind of thing yet with operators, but I think that is what we need to do (especially since the CRD now has "served=true" for both v1alpha1 and v1 versions - that means a user could have v1alpha1 on the system).

I will test this kind of thing and report back - I don't know how the operator SDK deals with watching for multiple CRDs and how that affects the watches.yaml.

jmazzitelli:
> t needs a test to see if the operator upgrade works OK

Oh, speaking of tests... that reminds me. We now need to update all the molecule tests to use the new CRD version.

Run this: `grep -R v1alpha1 operator/molecule` and you will see all the places that need to change.

Note that we still need a schema for the other CRD that Kiali owns - the monitoring dashboard CRD. The molecule tests also reference that one ("api_version: monitoring.kiali.io/v1alpha1")

operator/watches.yaml
israel-hdez: AFAIU, even if you have different versions of the CRD, when you query the API, it fetches the one that has `storage: true` but it just changes the `apiVersion` header. So, watching both versions would be like watching twice the same thing. At least, this is what I understand.

israel-hdez:
> Note that we still need a schema for the other CRD that Kiali owns - the monitoring dashboard CRD. The molecule tests also reference that one ("api_version: monitoring.kiali.io/v1alpha1")

I'd like to the dashboard CRDs in another PR.

jmazzitelli:
> I'd like to the dashboard CRDs in another PR.

I don't think we should split this effort into multiple PRs, however, if you want to make sure everything else is done in this PR (for example, updating the operator manifests also) then we can have a second PR to do the monitoring dashboard CRD separately.

But for this PR, we need to get everything done related to the kiali.io CRD (the big thing is the operator manifests have to be updated).

I still have to test the watches.yaml to see what needs to get done in there.

config/config.go
```diff
@@ -45,6 +45,8 @@ const (
 	AuthStrategyLoginIssuer     = "kiali-login"
 
 	// These constants are used for external services auth (Prometheus, Grafana ...) ; not for Kiali auth
+	// If you need to add more, make sure to update the schema of the operator crd:
+	//   $/operator/deploy/crd.yaml, the deployment.external_services property
```
jmazzitelli: We need to remind people the CRDs are all over the place - it isn't just one yaml file. I do not know the best way to do this. But, for example, a new CRD requires us to add a new CSV in the manifests (we have 3 different manifest locations today, [upstream](https://github.com/kiali/kiali/tree/v1.9/operator/manifests/kiali-upstream), [community](https://github.com/kiali/kiali/tree/v1.9/operator/manifests/kiali-community), and [ossm](https://github.com/kiali/kiali/tree/v1.9/operator/manifests/kiali-ossm)) and in there we need to add a new CRD. For example: https://github.com/kiali/kiali/blob/v1.9/operator/manifests/kiali-community/1.7.0/kiali.crd.yaml

Maybe this is good enough:

```suggestion
	//   (i.e. $/operator/deploy/crd.yaml and the CRDs defined in the OLM operator manifests), the deployment.external_services property
```

israel-hdez: OK, I used your suggestion

israel-hdez:
@jmazzitelli I've added the manifests.
I tried to test this using OLM, but I didn't succeed. Operator remains in "Unknown" state and it's not pulled.

I looked at the logs of the catalog-operator pod and *apparently* it doesn't like the `additionalProperties: true` attributes that are set in some parts of the CRD schema. I'll further investigate to check if this is really the issue.

operator/manifests/kiali-upstream/1.10.0/kiali.v1.10.0.clusterserviceversion.yaml
```diff
@@ -0,0 +1,596 @@
...
+          "apiVersion": "kiali.io/v1alpha1",
+          "kind": "Kiali",
+          "metadata": {
+            "name": "kiali"
+          },
+          "spec": {
+            "installation_tag": "My Kiali",
+            "istio_namespace": "istio-system",
+            "deployment": {
+              "namespace": "istio-system",
```
jmazzitelli: It looks like you have exact copies of these manifest files in both upstream and community. But you cannot copy the CSV yaml file from one to the other - they are actually slightly different. Take a look here:

https://github.com/operator-framework/community-operators/blob/master/upstream-community-operators/kiali/1.9.1/kiali.v1.9.1.clusterserviceversion.yaml#L27

versus here:

https://github.com/operator-framework/community-operators/blob/master/community-operators/kiali/1.9.1/kiali.v1.9.1.clusterserviceversion.yaml#L27

Notice the namespace is different. This is because when we place these files in a PR over in the operator framework/OLM github repo (for publication to OperatorHub.io), those PRs trigger travis tests. The travis tests for one is run on Kubernetes (the upstream manifest) and the travis tests for the other is run on OpenShift (the community manifest). And with namespace set to "istio-system", the tests fail on the Kubernetes environment (it has to go in the "default" namespace). But we can't use "default" for this as an OpenShift example because Kiali should not go in OpenShift's default namespace. And when these PR fail, the Kiali manifest will not get published to OperatorHub.io or in the OpenShift quay repo.

So you want this line to be:

```suggestion
              "namespace": "default",
```

I don't know if there are any other changes between the community and upstream manifests (I can't think of any - do a diff and see). But this is why I never copy files from upstream manifest to community manifest (or the other way around). Note: the CRD and the package files are the same, but the CSV yaml files are not.

(Hopefully you can now see why working with the operator/OLM stuff is a pain :-) This is also why I am meticulous in formatting and sorting the yaml in a consistent way - because I'm constantly having to look at files for tiny differences and having everything sorted and formatted consistently helps my eyes find things.)

israel-hdez:
I'm going to rearrange this PR.
OS4 don't have support for apiextensions.k8s.io/v1 yet. So, we must stay at apiextensions.k8s.io/v1beta1, and provide schema under that version.

operator/manifests/kiali-upstream/1.10.0/kiali.v1.10.0.clusterserviceversion.yaml
israel-hdez: Fixed...

> Hopefully you can now see why working with the operator/OLM stuff is a pain :-)

Well... What can I add? :)
I expected a better experience.

israel-hdez:
@jmazzitelli I stepped back and this is no longer moving the CRD definition to apiextensions.k8s.io/v1, because this API version is not available on OS4.2.

So, also the Kiali CRD is no longer on v1, which makes things a little easier, since there is no need to change the watches.yaml file nor the tests, etc.

I successfully installed it using OLM. But I'm not sure if an upgrade would work. I have a cluster with operator 1.9.1 and pushed 1.10.0... and it says it's still "up to date" even if it hasn't fetched 1.10.0... Not sure how long OLM will take to see the new version and do the upgrade :/

israel-hdez:
OLM does not upgrade the operator successfully.
These are the logs of OLM: https://paste.fedoraproject.org/paste/7MU1AfNtwhtMVxXiI9euzQ

Suspicious line in the logs is #228:

```
time="2019-11-14T20:46:17Z" level=error msg="Update existing cluster role failed: &ClusterRole{ObjectMeta:k8s_io_apimachinery_pkg_apis_meta_v1.ObjectMeta{Name:monitoringdashboards.monitoring.kiali.io-v1alpha1-edit,GenerateName:,Namespace:,SelfLink:,UID:,ResourceVersion:,Generation:0,CreationTimestamp:0001-01-01 00:00:00 +0000 UTC,DeletionTimestamp:<nil>,DeletionGracePeriodSeconds:nil,Labels:map[string]string{olm.opgroup.permissions/aggregate-to-edit: global-operators,olm.owner: kiali-operator.v1.10.0,olm.owner.kind: ClusterServiceVersion,olm.owner.namespace: openshift-operators,rbac.authorization.k8s.io/aggregate-to-edit: true,},Annotations:map[string]string{},OwnerReferences:[],Finalizers:[],ClusterName:,Initializers:nil,ManagedFields:[],},Rules:[{[create update patch delete] [monitoring.kiali.io] [monitoringdashboards] [] []}],AggregationRule:nil,}" error="clusterroles.rbac.authorization.k8s.io \"monitoringdashboards.monitoring.kiali.io-v1alpha1-edit\" not found"
time="2019-11-14T20:46:17Z" level=info msg="couldn't ensure clusterroles for provided api types" csv=kiali-operator.v1.10.0 error="clusterroles.rbac.authorization.k8s.io \"monitoringdashboards.monitoring.kiali.io-v1alpha1-edit\" not found" id=Ek2F+ namespace=openshift-operators phase=Succeeded
```

Then it keeps logging forever:

```
time="2019-11-14T20:49:09Z" level=info msg="requirements were not met" csv=kiali-operator.v1.10.0 id=LdHxd namespace=openshift-operators phase=Pending
time="2019-11-14T20:49:09Z" level=info msg="couldn't ensure RBAC in target namespaces" csv=kiali-operator.v1.10.0 error="no owned roles found" id=MSBnB namespace=openshift-operators phase=Pending
E1114 20:49:09.611694       1 queueinformer_operator.go:282] sync {"update" "openshift-operators/kiali-operator.v1.10.0"} failed: no owned roles found
time="2019-11-14T20:49:10Z" level=info msg="csv in operatorgroup" csv=kiali-operator.v1.10.0 id=FQ9A3 namespace=openshift-operators opgroup=global-operators phase=Pending
```

What works OK is uninstall previous version, then install new version :/

jmazzitelli:
PR https://github.com/kiali/kiali/pull/1943 is going to introduce another setting in the Kiali CR (deployment.resources). When that PR gets merged (which should be soon), we need to update the Kiali CRD in this PR. The type for `deployment.resources` is `object` (it contains a yaml dict for additional k8s information - similar to our `additional_service_yaml` setting we already have).

israel-hdez:
After a brief talk with OLM people, apparently, there is a bug in OLM that is preventing the operator to upgrade correctly. Issue with OLM is reported here: https://bugzilla.redhat.com/show_bug.cgi?id=1774732

@jshaughn @jmazzitelli Apparently, the issue with OLM is specific when using **manual** upgrades. I have tested the upgrade using the automatic strategy and it succeeds in that case. So, we could wait until the fix is rolled out, or we could move forward this PR and document the broken manual upgrade as a known issue :/

jmazzitelli:
> So, we could wait until the fix is rolled out, or we could move forward this PR and document the broken manual upgrade as a known issue :/

Until OLM can successfully upgrade in all cases, we should NOT merge this. OLM needs to work in order for us to incorporate this.

israel-hdez:
This is now working.
I think it can move forward.

operator/manifests/kiali-upstream/kiali.package.yaml
```diff
@@ -1,7 +1,7 @@
 packageName: kiali
 channels:
 - name: alpha
-  currentCSV: kiali-operator.v1.10.0
+  currentCSV: kiali-operator.v1.11.0
 - name: stable
-  currentCSV: kiali-operator.v1.10.0
+  currentCSV: kiali-operator.v1.11.0
```
jmazzitelli: This is the version I'm referring to - we should move these to 1.12.0. The names of the CSV yamls need to be renamed too (e.g. kiali.v1.12.0.clusterserviceversion.yaml)

jmazzitelli: fyi: I'm going to create a 1.11 set of manifests when we release this sprint

israel-hdez:
I've re-tested this PR and looks ready for a final review round.
I'd like to know if it is possible to have it merged for this release.

jotak:
@israel-hdez i've tested this monitoring dashboard spec with crd version `apiextensions.k8s.io/v1`, it seems to work fine : https://gist.github.com/jotak/e335ecc6b19229b1920c55897195d1a4

If you want to integrate within your PR.
There shouldn't be any code dependency, I didn't change the resource version (still v1alpha1) since it's not an API change, so I hope it will work seamlessly.

jmazzitelli:
I think we need to update all CRD versions to v1.

(a) this is not alpha anymore :)
(b) we are saying this is the first version containing the schema - a new schema (even though not truly a new  API) is new validation that must be adhered to

jotak:
@jmazzitelli what about planning in the short term a move to v1, but decouple that task from this purely schema work?

Because moving to v1 has more impacts. For Kiali CR I don't know, but for dashboards it means obviously upgrading every dashboard but also modify our clients to look for the correct versions (as here: https://github.com/kiali/k-charted/blob/master/kubernetes/client.go#L43 ) and also we'd need to answer the question, should we immediately shutdown the support for v1alpha1 or should we be able to support both, maybe with some deprecation mechanism and warnings (I'd be more in favor of a "gentle" deprecation).

All in all, we can answer these questions separately from the schema definition if we choose to separate tasks.

jmazzitelli:
I'm of the opinion we need to rip the bandaid off and do it all. We are adding new details to the CRD, it should probably be a new version. We need to figure this out sooner rather than later - because as soon as someone adds a new field, modifies a new field, or deletes a new field from a CRD we are going to have to figure out how upgrades work. So we need to figure this out. Whether its one commit, two commits, one PR, or two PRs, its all got to go into a release, and soon.

So I don't care if we split out into multiple PRs, mulitple commits, or whatever we deem appropriate.  But it has to go into a single release (i.e. I don't think we want to have one release update to "v1" CRDs and then the next release have these v1 CRDs with new schemas).

israel-hdez:
@jotak which OS version are you using?
I try in OS 4.2.4 and I get `error: unable to recognize "md-crd.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1"` when applying it.

@jmazzitelli I'm more on Joel's side. Moreover, IMO it's still alpha until we are sure that the strong schemas won't cause issues to users.

jotak:
@israel-hdez I tested on kubernetes v1.16.2 (no openshift). The whole task is about k8s 1.16, no? This is what I understand from @jmazzitelli description in the issue. Currently we have `apiextensions.k8s.io/v1beta1` without schema ; and we want to move to `apiextensions.k8s.io/v1` with schema ;

However, the spec is a little bit different between those two: in beta, the schema is under `spec.validation.openAPIV3Schema` whereas in v1 it's under `spec.versions.schema.openAPIV3Schema` . So, should we see this current work as an intermediate step to have a schema but still using the beta v1beta1 spec? Which means we'll have another similar work to do later?

jotak:
Well I'm now reading this comment from you: https://github.com/kiali/kiali/pull/1920#issuecomment-553662045
I think we'll need to have a manifest for kubernetes 1.16 but we shouldn't do that for openshift as long as it's not using 1.16, no?

jmazzitelli:
This is all getting too complicated. I was hoping this would be an easy "move our CRDs to v1, add the schemas, done." We don't want multiple sets of CRDs for Kubernetes and for Openshift.

Since this is new for k8s 1.16 (and OpenShift won't even be on 1.16 until at least 4.4 if not later) then we have time since this deprecated v1beta1 version of CRD will work for a while.

Let's not merge anything. At least the work we did here shows what the schemas will look like.  But until we are forced to use these schemas, let's just keep things the way they are now. It makes updates via OLM easier anyway.

israel-hdez:
@jotak 
> @israel-hdez I tested on kubernetes v1.16.2 (no openshift). The whole task is about k8s 1.16, no?

At first, I also tried on k8s v1.16. But when I tested on OS (because of OLM), it exploded. So, I reverted to `apiextensions.k8s.io/v1beta1` API.

@jmazzitelli 
> We don't want multiple sets of CRDs for Kubernetes and for Openshift.

Indeed, this is the reason why I choose to go back to `apiextensions.k8s.io/v1beta1`, because I didn't want to maintain two CRD sets.

---

@jotak 
> So, should we see this current work as an intermediate step to have a schema but still using the beta v1beta1 spec? Which means we'll have another similar work to do later?

...and @jmazzitelli 
> Let's not merge anything. At least the work we did here shows what the schemas will look like. But until we are forced to use these schemas, let's just keep things the way they are now.

I'm with one foot on each side. Although, I lean more towards merging as intermediate work because the way CRDs work in k8s/OS would mean that if we use `apiextensions.k8s.io/v1beta1` to support OpenShift, then newer k8s versions (like 1.16) should transparently make available the CRDs also under `apiextensions.k8s.io/v1` and k8s 1.16 should also take advantage of the added schemas.

But as you wish, if you think it's better to wait, let's wait.

israel-hdez:
This will be on-hold until OS supports the k8s 1.16 API (or until we find a safe way to implement this without maintaining two versions of the CRDs).

jmazzitelli:
Here's one issue I am not sure about and one reason I'm hesitant to implement this PR. We need to at least understand what the ramifications are of this:

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#overview

```
Since custom resource objects need to be able to be served at both versions, that
means they will sometimes be served at a different version than their storage version.
In order for this to be possible, the custom resource objects must sometimes be
converted between the version they are stored at and the version they are served at.
If the conversion involves schema changes and requires custom logic, a conversion
webhook should be used. If there are no schema changes, the default None
conversion strategy may be used and only the apiVersion field will be modified when
serving different versions.
```

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects

```
When an object is written, it is persisted at the version designated as the storage
version at the time of the write. If the storage version changes, existing objects are
never converted automatically. However, newly-created or updated objects are
written at the new storage version. It is possible for an object to have been written
at a version that is no longer served.
```

Now, consider the above when you think about the times when we need to introduce new CRDs that are not merely changes to schema but also indicate different versions of Kiali (and its Operator) entirely.

Assuming we have a single stream of versions in a single OLM package (today we have "stable" channel which includes all Kiali versions up to and including 1.13), this all works. Every Kiali version can read in and support (or at least can parse and understand) the CR no matter what API version it is based on.

But now consider we have two streams of Kiali versions (say, 1.x and 2.x) and consider 2.x is not compatible with 1.x BUT we can have a cluster that has Kiali 1.x and Kiali 2.x installed.

Now a problem appears (I think?). We do not want to convert Kiali 1.x CRs into Kiali 2.x CRs because Kiali 1.x might still be running and needs to read in the 1.x CR. 

Also, the operator-sdk does not support conversion webhooks today - so there is no easy solution to get users to be able to convert one CR to another when conversions are required.

I don't know all the ramifications of this. But these are the questions and issues that make me hesitant to introduce schemas and thus new CRD versions right now.

jmazzitelli:
I blocked user "ravikantcool2019" from the kiali organization. I just looked, this username was only created 4 days ago, and since then he has posted the same message "Approved the files" in over 50 PRs in 19 different reps. I don't know what's up - but this is very suspicious.

israel-hdez:
I'll close this PR. I'll reopen or open a new one when this work is really needed.

Issue #1886, opened 2019-11-06
Async load of SMMR causes failure of loading all objects


hhovsepy:
Happens on master.
Make sure SMMR has access to bookinfo.
Reinstall Maistra and Kiali so SMMR is recreated.
Add bookinfo into SMMR, but it takes some time load up all accessible namespaces.
There is kind a async load of accessible namespaces from SMMR between Namespace selector and the loading of for instance Workloads, which causes failure of loading of other accessible namespace items as well.
The namespaces names are shown in Namespace selector, but their items not in the list of objects, and this causes failing load of all other objects.
![Screenshot from 2019-11-06 10-45-59](https://user-images.githubusercontent.com/604313/68287607-4a91b400-0083-11ea-9fd1-c660b39a115f.png)

lucasponce:
@hhovsepy let me see if I understand the scenario.

ServiceMeshMemberRoll has "bookinfo" as a member.
User kube:admin has access to "bookinfo"

Kiali shows an error about "bookinfo" is not accessible ?
Does Kiali recover when it is refreshed ? (Not to solve the issue but to check the impact).

hhovsepy:
@lucasponce right, it takes up to a minute, then Kiali recovers and shows all accessible namespaces, including bookinfo.

lucasponce:
Fixed by #1925

PR #1925, opened 2019-11-12
Filter OS projects with accessible_namespaces for consistency

### LINKED ISSUE #1924: Adding an accessible namespace can cause GUI errors
Although I didn't test this on minikube, after some discussion with @lucasponce it is likely this issue occurs only on OpenShift.

To reproduce:
1. created OS 4.2 env on AWS with Istio 1.4
2. build/push/install Kiali operator/kiali built from master

At this point bookinfo doesn't exist and is not an accessible namespace

3, install bookinfo (but I think this could just be the creation of any new namespace)
4. leave kiali UI refreshing at 10s interval on Overview page
5. edit Kiali CR and add:
```
deployment:
  accessible_namespaces:
  - default
  - istio-system
  - bookinfo
```

Go back and watch overview page while kiali operator applies the CR change. Bookinfo card appears but is not filled in, errors generate/repeat for a while, then it settles down.

image

lucasponce:
Fixes #1924 

@jshaughn I've prepared a fix, but I couldn't test it yet, I'm preparing an AWS environment (but you know it takes time...).

If you can test it, I guess this should solve the issue.

Let me know, thanks !

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -72,20 +72,29 @@ func (in *NamespaceService) GetNamespaces() ([]models.Namespace, error) {
 	labelSelector := config.Get().API.Namespaces.LabelSelector
 
 	namespaces := []models.Namespace{}
+	_, queryAllNamespaces := in.isAccessibleNamespaces["**"]
 	// If we are running in OpenShift, we will use the project names since these are the list of accessible namespaces
 	if in.hasProjects {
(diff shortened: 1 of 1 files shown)


hhovsepy:
Also fixes this issue: https://github.com/kiali/kiali/issues/1886

PR #1927, opened 2019-11-12
Refresh Kiali caches after Create/Update/Delete

### LINKED ISSUE #1877: Refresh/Sync Kiali cache after edition to avoid to read outdated info
After editing a VS, it looks like that validations are not correctly updated:

image

It seems the last backend response is correct but is not propagated into the editor.

lucasponce:
Fixes #1877 

@hhovsepy I couldn't test it locally, but I guess it should work.

Tomorrow I will perform more tests but let me know if you can test it.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -674,6 +674,11 @@ func (in *IstioConfigService) DeleteIstioConfigDetail(api, namespace, resourceTy
 	} else {
 		err = in.k8s.DeleteIstioObject(api, namespace, resourceType, name)
 	}
+
+	// Cache is stopped after a Create/Update/Delete operation to force a refresh
+	if kialiCache != nil && err == nil {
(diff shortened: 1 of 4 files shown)


hhovsepy:
@lucasponce I can still see the issue, when cache enabled, each "Reload" button click show different warning list on YAML editor.
Also the "Actions" menu gets disabled when deleting routing, then only gets enabled after reload of page.
Screenshot from 2019-11-13 10-19-19

lucasponce:
Yes, the patch was not fully tested.
I'm still working on it.

lucasponce:
@hhovsepy Ok, now this PRs performs two changes:
- Invalidates / Refresh Kiali Cache on well known CUD operations.
- It uses cache when those can be considered Synced (from k8s/cache interface).

Let me know how it looks like.

Issue #1556, opened 2019-08-27
Add indication on status of validations on the cards in overview page


lucasponce:
Add indication on validation err/warn on the cards in overview page
Currently the overview cards per namespace show traffic and health.
It will be helpful to add also indication of validation issues of configuration that belongs to that namespace.

JIRA:
https://issues.jboss.org/browse/KIALI-3158

lucasponce:
Completed

Issue #1930, opened 2019-11-13
Validations shown in empty namespaces


xeviknal:
MultiMatchGateway checker returns warnings even when there aren't any Gateways in the selected namespace. See the following:
![Screenshot from 2019-11-13 17-51-09](https://user-images.githubusercontent.com/613814/68785342-53aff180-063e-11ea-9f43-12b20f6fccfa.png)

The shown validations come from the gateways in the bookinfo namespace:
![Screenshot from 2019-11-13 17-53-49](https://user-images.githubusercontent.com/613814/68785519-940f6f80-063e-11ea-871b-f0a35e52505d.png)

Expected output:
Return an empty set of validations when there are no Istio Objects for a given namespace.

xeviknal:
Steps to reproduce:

```
cat <<EOF | oc apply -n bookinfo -f -
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: bookinfo-gateway
spec:
  selector:
    istio: ingressgateway # use istio default controller
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - "*"
---
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: bookinfo-gateway-copy
spec:
  selector:
    istio: ingressgateway # use istio default controller
  servers:
  - port:
      number: 80
      name: http
      protocol: HTTP
    hosts:
    - "*"
```

Since Gateway are shared between namespaces and the gateways are pointing to '*', all namespace in the cluster has the warning that they have a more than one gateway pointing to the same host. Actually, it is useful because if the user adds a new gateway it might be clashing with other gateways defined in the cluster.

So I guess that this is not a but. It is a legit scenario.
@lucasponce @burmanm wdyt?

Same case happened to @jshaughn in https://github.com/kiali/kiali/pull/1932#issuecomment-555629143

xeviknal:
I have just tried to change the Gateways to point to other host than *. The result is that the validations are shown as well for empty namespaces.
I am thinking of remove the validations in case there aren't gateways.
On the other side, it might be useful to have a global indicator of the gateways enabled in the cluster. Perhaps it doesn't make sense to show gateways per namespace if they can be used in any namespace. Thinking of a quick tooltip on a corner of IstioObject List showing which are the hosts covered by the gateways of the current servicemesh. Show a warning error in case there are duplications. Plus, those hosts should be linked to the gateway details page.

[...]

PR #1932, opened 2019-11-13
Adding NamespaceValidations endpoint with summary of validations

xeviknal:
** Describe the change **
Backend work for https://github.com/kiali/kiali-ui/pull/1541
Returns Validation aggregation count for each namespace:
Screenshot of https___kiali-istio-system 192 168 42 56 nip io_api_namespaces_bookinfo_validations

** Issue reference **
kiali/kiali#1556

### DIFF
--- a/handlers/namespaces.go
+++ b/handlers/namespaces.go
@@ -1,13 +1,12 @@
 package handlers
 
 import (
-	"github.com/kiali/kiali/models"
 	"net/http"
-	"sync"
 
 	"github.com/gorilla/mux"
 
 	"github.com/kiali/kiali/log"
+	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/prometheus"
 )
 
@@ -58,9 +57,9 @@ func getNamespaceMetrics(w http.ResponseWriter, r *http.Request, promSupplier pr
 	RespondWithJSON(w, http.StatusOK, metrics)
 }
 
-//NamespaceValidations the API handler to fetch validations to be displayed.
+// NamespaceValidationSummary is the API handler to fetch validations summary to be displayed.
 // It is related to all the Istio Objects within the namespace
-func NamespaceValidations(w http.ResponseWriter, r *http.Request) {
+func NamespaceValidationSummary(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	namespace := vars["namespace"]
 
@@ -73,18 +72,13 @@ func NamespaceValidations(w http.ResponseWriter, r *http.Request) {
 
 	var validationSummary models.IstioValidationSummary
 
-	wg := sync.WaitGroup{}
-	wg.Add(1)
-	go func(namespace string, validationSummary *models.IstioValidationSummary, err *error) {
-		defer wg.Done()
-		istioConfigValidationResults, errValidations := business.Validations.GetValidations(namespace, "")
-		if errValidations != nil && *err == nil {
-			*err = errValidations
-		} else {
-			*validationSummary = istioConfigValidationResults.SummarizeValidation()
-		}
-	}(namespace, &validationSummary, &err)
+	istioConfigValidationResults, errValidations := business.Validations.GetValidations(namespace, "")
+	if errValidations != nil {
+		log.Error(errValidations)
+		RespondWithError(w, http.StatusInternalServerError, errValidations.Error())
+	} else {
+		validationSummary = istioConfigValidationResults.SummarizeValidation()
+	}
 
-	wg.Wait()
 	RespondWithJSON(w, http.StatusOK, validationSummary)
 }
--- a/handlers/namespaces.go
+++ b/handlers/namespaces.go
@@ -1,7 +1,9 @@
 package handlers
 
 import (
+	"github.com/kiali/kiali/models"
 	"net/http"
+	"sync"
 
 	"github.com/gorilla/mux"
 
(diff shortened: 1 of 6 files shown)


handlers/namespaces.go
```diff
@@ -55,3 +57,34 @@ func getNamespaceMetrics(w http.ResponseWriter, r *http.Request, promSupplier pr
...
+	if err != nil {
+		log.Error(err)
+		RespondWithError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+
+	var validationSummary models.IstioValidationSummary
+
+	wg := sync.WaitGroup{}
+	wg.Add(1)
```
lucasponce: This wg is not necessary, the main thread is not doing anything else.

handlers/namespaces.go
```diff
@@ -55,3 +57,34 @@ func getNamespaceMetrics(w http.ResponseWriter, r *http.Request, promSupplier pr
 	metrics := prom.GetMetrics(&params)
 	RespondWithJSON(w, http.StatusOK, metrics)
 }
+
+//NamespaceValidations the API handler to fetch validations to be displayed.
```
lucasponce: Two things:
- Typo, missing space at the beginning.
- This endpoint won't return validations, there is an existing endpoint that returns istio objects + validations. This enpoint returns a summary.

I can understand we don't want the existing endpoint as we only need the summary, but description should clarify that to avoid misunderstandings.

handlers/namespaces.go
```diff
@@ -55,3 +57,34 @@ func getNamespaceMetrics(w http.ResponseWriter, r *http.Request, promSupplier pr
...
+// It is related to all the Istio Objects within the namespace
+func NamespaceValidations(w http.ResponseWriter, r *http.Request) {
```
lucasponce: NamespaceValidationSummary ?

routing/routes.go
```diff
@@ -793,6 +793,27 @@ func NewRoutes() (r *Routes) {
 			handlers.NamespaceHealth,
 			true,
 		},
+		// swagger:route GET /namespaces/{namespace}/validations namespaces namespaceValidations
+		// ---
+		// Get validations for all objects in the given namespace
```
lucasponce: This should describe a "summary" not validations.

routing/routes.go
```diff
@@ -793,6 +793,27 @@ func NewRoutes() (r *Routes) {
...
+		//
+		//     Schemes: http, https
+		//
+		// responses:
+		//      200: namespaceValidationsResponse
+		//      400: badRequestError
+		//      500: internalError
+		//
+		{
+			"NamespaceHealth",
```
lucasponce: Wrong name in the ID. "NamespaceHealth" -> "NamespaceValidationsSummary" or something similar.

jshaughn:
minor, not sure this really needs "frontend-changes-required".  When ready I think it could be merged as it only adds to the API.

routing/routes.go
```diff
@@ -793,6 +793,27 @@ func NewRoutes() (r *Routes) {
...
+		//
+		// responses:
+		//      200: namespaceValidationSummaryResponse
+		//      400: badRequestError
+		//      500: internalError
+		//
+		{
+			"NamespaceValidationSummary",
+			"GET",
+			"/api/namespaces/{namespace}/validations",
```
jshaughn: @lucasponce Perhaps this should be /validationsummary ?

lucasponce: As there is no other "/validations" endpoint, this is fine.
For consistency we have used one single term to refer the endpoints (unless there  is a conflict).

jshaughn: OK, just making sure because the other comments were to change from validations to validationSummaries.

xeviknal:
@jshaughn do you mind showing me the output of the following endpoint?

```
/api/namespaces/default/istio?validate=true
```
Could you paste the json part of the validations?
Besides that, in the UI, the default namespace shows any validation message? Perhaps a N/A?

I do have this issue open that is quite related. It might seem an error but it has a rationale.

jshaughn:
@xeviknal , it's interesting, there are two configs I guess, which apply to all of the namespaces.  They are both valid, I think perhaps they are both being counted as invalid?  I don't know.
```
{
  "namespace": {
    "name": "default"
  },
  "gateways": [],
  "virtualServices": {
    "permissions": {
      "create": false,
      "update": false,
      "delete": false
    },
    "items": []
  },
  "destinationRules": {
    "permissions": {
      "create": false,
      "update": false,
      "delete": false
    },
    "items": []
  },
  "serviceEntries": [],
  "rules": [],
  "adapters": [],
  "templates": [],
  "quotaSpecs": [],
  "quotaSpecBindings": [],
  "policies": [],
  "meshPolicies": [],
  "serviceMeshPolicies": [],
  "clusterRbacConfigs": [],
  "rbacConfigs": [],
  "serviceMeshRbacConfigs": [],
  "serviceRoles": [],
  "serviceRoleBindings": [],
  "sidecars": [],
  "validations": {
    "gateway": {
      "bookinfo-gateway": {
        "name": "bookinfo-gateway",
        "objectType": "gateway",
        "valid": true,
        "checks": [
          {
            "message": "More than one Gateway for the same host port combination",
            "severity": "warning",
            "path": "spec/servers[0]/hosts[0]"
          }
        ],
        "references": [
          {
            "objectType": "gateway",
            "name": "ingressgateway",
            "namespace": "istio-system"
          }
        ]
      },
      "ingressgateway": {
        "name": "ingressgateway",
        "objectType": "gateway",
        "valid": true,
        "checks": [
          {
            "message": "More than one Gateway for the same host port combination",
            "severity": "warning",
            "path": "spec/servers[0]/hosts[0]"
          }
        ],
        "references": [
          {
            "objectType": "gateway",
            "name": "bookinfo-gateway",
            "namespace": "bookinfo"
          }
        ]
      }
    }
  }
}
```

xeviknal:
yes, @jshaughn I think this is a legit scenario. See comments here: https://github.com/kiali/kiali/issues/1930#issuecomment-555631461

In the UI, what I did is hide errors in case there are no applications/services. I think is useful to see validation information only if there are subjects that may be affected.

jshaughn:
@xeviknal But what about non-empty namespaces (with applications/services)?  For example, if I go to overview bookinfo namespace complains of two warnings.  If I then click on Config icon I get to the list view and see only one warning because only one of the gateways is defined in the bookinfo namespace.

I expect to see the same number of warnings/errors in this summary as I do when I restrict the Istio Config list view to that namespace.  Otherwise I think it's confusing.  I understand that these gateway definitions can affect namespaces outside of the one in which it is defined.  I also understand that the column/filter is on the *defining* namespace.   But I just don't think we can't tell a user the namespace has 2 warnings and then give him no way to easily see the two contributing configs.

I'm OK if we merge before figuring this out, but for me it's weird the way it is working.

xeviknal:
@jshaughn you are right. It could be even worse. You have apps/services/istio objects see the warning and go to istio object list and see no objects listed.
Yep, I'll give a second thought. Thanks for pointing that out.

jshaughn:
@xeviknal As I mentioned above, I don't necessarily want to block the PRs if they are otherwise ready to go.  In fact, my PR for the graph is currently built on both of yours.  But I do think we need a story here.

lucasponce:
It's an interesting comment.
Also, cross-namespace validations if not fully complete, so it can be more corner cases.

I think that is related to this feature, but we can consolidate first, we can iterate in the backend with the consistency in the scenarios.

@xeviknal would you mind to open a new Issue to track @jshaughn comments ?

I'm going to review and merge this feature (backend + frontend).

PR #1943, opened 2019-11-16
[operator] be able to specify amount of resources allocated to the kiali pod

### LINKED ISSUE #1942: [operator] be able to specify amount of resources allocated to the kiali pod
You can specify resource allocations needed for a specific pod's containers. We should be able to tell the operator what resources we want for the kiali pod. Upstream Istio installer allows you to specify the "resources" (a section in a k8s Deployment's containers) for the different istio pods, we want to support it as well as upstream Istio users can tell what resources are to be given to kiali.

For background on the "resources" data in the Deployment, see https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/

jmazzitelli:
fixes #1942 

Kiali CR now supports a new "deployment.resources" dict which allows you to say what resources are to be allocated to the Kiali pod's container. So it would look like:

```yaml
spec:
  deployment:
    resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "500m"
```


jmazzitelli:
I tested and this works. If you pass a Kiali CR with `deployment.resources` in its spec that follows the [syntax found in the k8s docs here](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/), for example:

```yaml
spec:
  auth:
    strategy: openshift
  deployment:
    namespace: istio-system
    resources:
      requests:
        memory: "64Mi"
        cpu: "250m"
      limits:
        memory: "128Mi"
        cpu: "500m"
```

The operator will install the Kiali Deployment with this in its container (notice the `resources` element):

```yaml
      containers:
        - resources:
            limits:
              cpu: 500m
              memory: 128Mi
            requests:
              cpu: 250m
              memory: 64Mi
```

jmazzitelli:
I'm going to submit another commit for this PR with a molecule test.

Note that I found a limitation. If a user already has "resources" defined in the CR (the same will be true for other settings like affinity and tolerations as well) and he then removes it, that won't do anything. When the operator updates the Deployment, the old resource valuesremain (unless the CR overrides existing settings).

For details, see: https://github.com/ansible/ansible/issues/43206

We need to wait for an ansible operator base image that has ansible 2.9 in which case it looks like we can use [the new `apply` setting for the `k8s` task](https://docs.ansible.com/ansible/latest/modules/k8s_module.html#parameters). 

A workaround is to set the values to literal string `"null"` - this tells the k8s api server that this setting needs to be removed. I may have to update the operator code to support this, though. I'll submit that as part of this PR if its doable.

jmazzitelli:
I found another workaround for that limitation and doesn't require anything special from the user. It should just work now if the user removes the deployment.[affinity,tolerations,resources] sections now.

Molecule test has been added, and it also tests this removal of those settings as well as tests this new 'resources' feature.

PR #1945, opened 2019-11-17
[operator] the default deployment namespace should be the namespace where the CR is

### LINKED ISSUE #1944: [operator] default value for deployment.namespace should be where the CR is
When the operator examines the Kiali CR, it looks at `deployment.namespace` to determine where it should install Kiali. If that setting is not defined, its default today is `istio-system`. So Kiali will always be installed in the namespace `istio-system` by default, regardless of where the Kiali CR is located.

This is actually a poor default. The default should be where the Kiali CR is found. So if a Kiali CR is created in the namespace `foo`, and that CR does not define `deployment.namespace`, then Kiali should be deployed in the same namespace where that CR is (in other words, `foo` namespace). This is the more natural default.

So we need to comment out this: https://github.com/kiali/kiali/blob/v1.9.1/operator/roles/kiali-deploy/defaults/main.yml#L59

And if the operator sees that that `deployment.namespace` setting is undefined, it should use the namespace where the Kiali CR is - which should be (I hope) found in [here](https://github.com/kiali/kiali/blob/v1.9.1/operator/roles/kiali-deploy/tasks/main.yml#L58) using `current_cr.metadata.namespace`

jmazzitelli:
fixes #1944

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -198,9 +198,10 @@ spec:
 #    ---
 #    ingress_enabled: true
 #
-# The namespace into which Kiali is to be installed.
+# The namespace into which Kiali is to be installed. If this is empty or not defined,
+# the default will be the namespace where the Kiali CR is located.
 #    ---
-#    namespace: "istio-system"
+#    namespace: ""
 #
 # The name of the secret where the Kiali credentials are found.
 # These credentials will be required when logging into Kiali.
--- a/operator/molecule/accessible-namespaces-test/kiali-cr.yaml
+++ b/operator/molecule/accessible-namespaces-test/kiali-cr.yaml
@@ -7,7 +7,7 @@ spec:
   auth:
     strategy: {{ kiali.auth_strategy }}
   deployment:
-    namespace: {{ istio.control_plane_namespace }}
+    namespace: {{ kiali.install_namespace }}
     image_name: {{ kiali.image_name }}
     image_pull_policy: {{ kiali.image_pull_policy }}
     image_version: {{ kiali.image_version }}
--- a/operator/molecule/accessible-namespaces-test/kiali-cr.yaml
+++ b/operator/molecule/accessible-namespaces-test/kiali-cr.yaml
@@ -7,9 +7,9 @@ spec:
   auth:
     strategy: {{ kiali.auth_strategy }}
   deployment:
+    namespace: {{ istio.control_plane_namespace }}
     image_name: {{ kiali.image_name }}
     image_pull_policy: {{ kiali.image_pull_policy }}
     image_version: {{ kiali.image_version }}
     accessible_namespaces: {{ kiali.accessible_namespaces }}
     service_type: NodePort
-
--- a/operator/molecule/accessible-namespaces-test/molecule.yml
+++ b/operator/molecule/accessible-namespaces-test/molecule.yml
@@ -22,6 +22,7 @@ provisioner:
         istio:
           control_plane_namespace: istio-system
         kiali:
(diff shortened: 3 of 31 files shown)


jmazzitelli:
This will help upstream Istio when it moves to using the kiali operator.

jmazzitelli:
@gbaufake this changes the default for `deployment.namespace`. Before this PR, the default was hardcoded `istio-system`, but now it is the location of the Kiali CR. All the molecule test CRs that I found did not specify `deployment.namespace` so it always defaulted to `istio-system`, which is fine because all of the tests want that value. But now the default for `deployment.namespace` (at least for some of these tests) will be `kiali-operator` namespace (because that is where some tests put the Kiali CR). Obviously, we do not want Kiali installed in `kiali-operator` so I had to change some of the molecule kiali CR yaml files to explicitly set `deployment.namespace`.

Hopefully, none of this affects your e2e tests - but please double check that these tests are OK. I tested some, but not the e2e tests (e.g. olm-test) that you work with.

jmazzitelli:
There is now a molecule test (`default-namespace-test`) in this PR that shows this feature working.

jmazzitelli:
I need this carefully looked at -  I may want to tweek that molecule test to ensure it is testing what I think it is testing. Many tests are assuming "istio-system" is where kiali is supposed to be deployed, but for this feature I want to move Kiali somewhere else to see this feature working.

As it is now, I think this is working, but I'm not confident the test is really testing things the way I want.

jmazzitelli:
> I may want to tweek that molecule test to ensure it is testing what I think it is testing.

The last commit tweeks the test by putting the Kiali CR in a separate namespace and making sure
Kiali is installed in that namespace.

Also fixed the tests to make sure they are checking in the right places by using the correct variables.

jmazzitelli:
> so now we need to explicitly to tell kiali to be installed on that namespace?

Because the only difference is the default "deployment.namespace" (It used to be a hardcoded default of "istio-system". Now the default is the location where the CR is located), if you do not explicitly tell Kiali where to be installed, it will be installed in the namespace where the CR is.

So you **will** have to explicitly tell Kiali which namespace to be installed IF you installed the CR in one namespace (i.e. kiali-operator) but want Kiali installed somewhere else (i.e. istio-system).

But that is how Maistra works (i.e. Kiali is installed in the same namespace where the CR is located) - so for the Maistra use case, nothing really has to change.

But for these tests, I did have to add some additional variables (cr_namespace, kiali.install_namespace) because the hardcoded "istio-system" cannot be assumed anymore. You have to explicitly tell the tests where to expect kiali to be installed.

NOTE: The users of deploy-kiali-operator.sh (getLatestKialiOperator) should not see any changes. That script already explicitly defines deployment.namespace so they never relied on this default value. See https://github.com/kiali/kiali/blob/master/operator/deploy/deploy-kiali-operator.sh#L480 and https://github.com/kiali/kiali/blob/master/operator/deploy/deploy-kiali-operator.sh#L1109 where the script sets it explicitly (i.e. it always explicitly sets that value to "istio-system" if the user didn't tell the script what to set NAMESPACE, so it never relies on the defaults - thus this PR doesn't affect people using that script).

jmazzitelli:
@gbaufake 

> I'll need to update some tests because of that.

If you want, submit a PR to this PR if you want to include those test changes with this. I can add your changes to this PR so they all go in together.

jmazzitelli:
I need to merge this now - needed to simplify my work with the other branch for https://github.com/kiali/kiali/pull/1943

@gbaufake if you need to fix any of your tests, just submit another PR. I didn't think olm-test or maistra-e2e test would be affected though since they are installing Service mesh and not authoring its own Kiali CR files.

Issue #1550, opened 2019-08-27
Improve Istio Config Access Error Message When User Does Not Have Permission


lucasponce:
When a user does not have permission to access a namespace, attempting to access Kiali Istio Config eventually times out with error "Could not fetch Istio objects list, Error: [ Gateway Time-out ]". The message should be more user friendly as to why the timeout occurred, such as perhaps an access denied message.

Repro:
1) Create new OS user 'foo'
2) Create Role Binding for user 'foo' that has "Edit" role access to namespace 'bookinfo'
3) Log into Kiali as user 'foo', and click on Istio Config

JIRA:
https://issues.jboss.org/browse/KIALI-3140

xeviknal:
Steps to reproduce using `./aws-openshift`:

1. Create a new user [1] / [2]
```
htpasswd -c -B -b users.htpasswd user user
htpasswd -b users.htpasswd user user
oc create secret generic htpass-secret --from-file=htpasswd=users.htpasswd -n openshift-config
cat <<EOF | oc apply -f -
apiVersion: config.openshift.io/v1
kind: OAuth
metadata:
  name: cluster
spec:
  identityProviders:
  - name: my_htpasswd_provider 
    mappingMethod: claim 
    type: HTPasswd
    htpasswd:
      fileData:
        name: htpass-secret
EOF

oc login -u user
```
2. Create bookinfo:
```
./aws-openshift.sh bi-install
```

3. Add bookinfo to ServiceMeshMemberRoll
```
cat <<EOF | oc apply -f -
apiVersion: maistra.io/v1
kind: ServiceMeshMemberRoll
metadata:
  # name must be default
  name: default
spec:
  members:
  # a list of projects/namespaces that should be joined into the service mesh
  # for example, the bookinfo project/namespace
  - bookinfo
EOF
```

4. Create role and assign it to `user`
```
cat <<EOF | oc apply -f -
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  name: my_configview
  namespace: bookinfo
rules:
- apiGroups:
  - ""
  resources:
  - configmaps
  - endpoints
  - namespaces
  - nodes
  - pods
  - pods/log
  - replicationcontrollers
  - services
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - extensions
  - apps
  resources:
  - deployments
  - replicasets
  - statefulsets
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - autoscaling
  resources:
[...]

PR #1948, opened 2019-11-18
Removing deadlock in IstioConfigList

xeviknal:
** Describe the change **

Tricky situation. In case there were two or more errors in the goroutines, the channel blocked the execution until the error was read. The problem was that there was a semaphore waiting for all the goroutines to finish. Deadlock. The result of this was a Timeout.
The execution is not blocked after adding a buffered channel of length 18. Each attempt to add will have room and won't block.

Right now, the response is:

Screenshot of Kiali  istio-system  (9)

Then, when the user has permissions for the resource, the list will be displayed. Meaning, we are following the same pattern as before.

** Issue reference **

Closes https://github.com/kiali/kiali/issues/1550

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -133,7 +133,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 		return models.IstioConfigList{}, err
 	}
 
-	errChan := make(chan error, 1)
+	errChan := make(chan error, 18)
 
 	var wg sync.WaitGroup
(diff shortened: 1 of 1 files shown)


lucasponce:
@xeviknal interesting, probably I introduced this bug some time ago.
Let me review it.

lucasponce:
Nice catch.
It's true, it was not protected for +2 concurrent errors.
We followed a pattern of 1 size buffer for errors used in validations, but without checking the status of the buffer.

PR #1959, opened 2019-11-19
[operator] customize priority class name and pod annotations

jmazzitelli:
upstream istio expects components to be configured with a custom priorityClassName and pod annotations. its a good feature to have. this PR adds this.


jmazzitelli:
do not merge this yet. I want to test it some more and maybe get a molecule test in

jmazzitelli:
I tested this, works fine.

Issue #1963, opened 2019-11-20
Validations - Destination Rule traffic policy override error


xeviknal:
There are two Destination Rules living in `istio-system` showing a warning. For instance, the following one:
![Screenshot of Kiali  istio-system  (10)](https://user-images.githubusercontent.com/613814/69257554-8700fc00-0bbb-11ea-930a-dc2db5cf0c4c.jpg)

The offending object is the following:
![Screenshot of Kiali  istio-system  (11)](https://user-images.githubusercontent.com/613814/69257536-823c4800-0bbb-11ea-9313-ecb6a5aeb39a.jpg)

Error reference: https://kiali.io/documentation/validations/#_mtls_settings_of_a_non_local_destination_rule_are_overridden

Either the offending object is incorrect or the validation itself has a problem.

PR #1965, opened 2019-11-21
Validations - Destination Rule traffic policy override error

xeviknal:
** Describe the change **
Steps to reproduce:
1. Add a DR rule pointing to host `*.bookinfo.svc.cluster.local` enabling mtls.
2. See DR in istio-system, for example `istio-telemetry`

Kiali warns DR in istio-system. It shouldn't be.
This PR fixes that.

** Issue reference **
Closes https://github.com/kiali/kiali/issues/1963

### DIFF
--- a/business/checkers/destinationrules/traffic_policy_checker.go
+++ b/business/checkers/destinationrules/traffic_policy_checker.go
@@ -19,16 +19,26 @@ func (t TrafficPolicyChecker) Check() models.IstioValidations {
 		return validations
 	}
 
-	refKeys := make([]models.IstioValidationKey, 0, len(refdMtls))
-	for _, dr := range refdMtls {
-		refKeys = append(refKeys, models.BuildKey(DestinationRulesCheckerType, dr.GetObjectMeta().Name, dr.GetObjectMeta().Namespace))
-	}
-
 	// Check whether DRs override mTLS.
 	for _, dr := range t.DestinationRules {
+		drSameHosts := sameHostDestinationRules(dr, refdMtls)
+
+		// Continue if there aren't DestinationRules enabling mTLS non-locally
+		// and pointing to same host as dr.
+		if len(drSameHosts) == 0 {
+			continue
+		}
+
+		// Invalid if there isn't trafficPolicy specified or trafficPolicy doesn't specify TLSSettings
 		if !hasTrafficPolicy(dr) || !hasTLSSettings(dr) {
 			check := models.Build("destinationrules.trafficpolicy.notlssettings", "spec/trafficPolicy")
 			key := models.BuildKey(DestinationRulesCheckerType, dr.GetObjectMeta().Name, dr.GetObjectMeta().Namespace)
+
+			refKeys := make([]models.IstioValidationKey, 0, len(refdMtls))
+			for _, dr := range refdMtls {
+				refKeys = append(refKeys, models.BuildKey(DestinationRulesCheckerType, dr.GetObjectMeta().Name, dr.GetObjectMeta().Namespace))
+			}
+
 			validation := buildDestinationRuleValidation(dr, check, true, refKeys)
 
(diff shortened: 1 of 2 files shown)


xeviknal:
@lucasponce @burmanm do you guys know where `*.global` is used and why? Because it could be a scenario to cover not only here but in whole validations.

business/checkers/destinationrules/traffic_policy_checker.go
```diff
@@ -55,6 +58,31 @@ func (t TrafficPolicyChecker) drsWithNonLocalmTLSEnabled() []kubernetes.IstioObj
 	return mtlsDrs
 }
 
+func sameHostDestinationRules(dr kubernetes.IstioObject, mdrs []kubernetes.IstioObject) []kubernetes.IstioObject {
+	var drHost kubernetes.Host
+	shdrs := make([]kubernetes.IstioObject, 0, len(mdrs))
+
+	if host, ok := dr.GetSpec()["host"]; ok {
+		if dHost, ok := host.(string); ok {
+			drHost = kubernetes.ParseHost(dHost, dr.GetObjectMeta().Namespace, dr.GetObjectMeta().ClusterName)
```
burmanm: Maybe this was just some environments (or tests only?), but the ClusterName was in some cases empty..

burmanm: Arr, nm.. the ParseHost takes care of that.

xeviknal: yes, I assume that ParseHost it will populate those 3 fields(service, ns, cluster). Except for "*.local" that it only will have service with "*.local" value.

lucasponce:
> @lucasponce @burmanm do you guys know where *.global is used and why? Because it could be a scenario to cover not only here but in whole validations.

I don't remember the "*.global", worth to check the k8s specification:
https://github.com/kubernetes/dns/blob/master/docs/specification.md

lucasponce:
But not sure if we need to review the whole validation.
As now Destinations within DR can be anything, not only internal services names but also declared ServiceEntries.

@burmanm @xeviknal do you remember if we cover this case as well ?

burmanm:
We do check the ServiceEntries in other DR validations, yes.

xeviknal:
> As now Destinations within DR can be anything, not only internal services names but also declared ServiceEntries.

Do you have an example?

I think this validation is correct for ServiceEntries too:

1. Policies only applies to workloads in the ServiceMesh. Therefore, it doesn't apply to ServiceEntries. Am I wrong?
2. DestinationRules may point to hosts defined within a ServiceMesh (www.kiali.io).
3. If you define a Policy with non-local mTLS setting, it won't apply to serviceentries communications.
4. If you have a DR pointing to `www.kiali.io` not specifying `TLSSettings`, Kiali shouldn't return any warning because the Destination rule pointing to all workloads in the NS doesn't apply to it.

```yaml
kind: Policy
apiVersion: authentication.istio.io/v1alpha1
metadata:
  name: default
  namespace: bookinfo
spec:
  targets: ~
  peers:
    - mtls:
        mode: STRICT
---
kind: DestinationRule
metadata:
  name: enable-mtls
  namespace: bookinfo
spec:
  host: '*.bookinfo.svc.cluster.local'
  trafficPolicy:
    tls:
      mode: ISTIO_MUTUAL
---
kind: ServiceEntry
metadata:
  name: facebook
  namespace: bookinfo
spec:
  hosts:
    - www.kiali.io
  addresses: ~
  ports:
    - name: http
      number: 80
      protocol: HTTP
  location: MESH_EXTERNAL
  resolution: DNS
---
kind: DestinationRule
metadata:
  name: kiali-dr
  namespace: bookinfo
spec:
  host: www.kiali.io
  subsets:
  - name: testversion
    labels:
      version: dev
  - name: prodversion
    labels:
      version: prod
```

PR #1967, opened 2019-11-22
issue 1561 the operator SDK exposes metrics; tell prometheus to scrape it

### LINKED ISSUE #1561: [operator] expose operator metrics / service
If/when we get to phase 5 for operators and we want to have "auto pilot" features, we need to expose metrics. I'm just documenting what I've found out about this so far.

Operator-SDK 0.8.x and onwards automatically exposes the Operator's metrics endpoint [1] and exposes that as a service [2]. I think we need to expose this port in an operator Service resource.
But I'm not sure - we may not have to expose this as a service. We do know the operator will get a metric endpoint created, listening at port 8383 [3]. That may be all we need, as long as we put this in our operator deployment:
```
spec:
  metadata:
   annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "8383"
```
The problem is this might not work if the operator is deployed in a different namespace than where Prometheus is. Need to do some research as to how those annotations work when Prom and the metric endpoint are in different namespaces.

In our case, the Operand is Kiali, and Kiali already has an exposed metrics port in its service.

[1] https://github.com/operator-framework/operator-sdk/blob/v0.9.x/pkg/ansible/run.go#L76
[2] https://github.com/operator-framework/operator-sdk/blob/v0.9.x/pkg/ansible/run.go#L95-L101
[3] https://github.com/operator-framework/operator-sdk/blob/v0.9.x/pkg/ansible/run.go#L43

JIRA:
https://issues.jboss.org/browse/KIALI-3184

jmazzitelli:
fixes #1561 

This is part of the operator maturity model phase 4 which kiali operator supports.

This PR includes changes to the deploy script as well as the several gold copy of our OLM manifests.

To test this works, install operator normally using the deploy script:

```
cd operator/deploy
./deploy-kiali-operator.sh
```
Once the kiali operator pod is running, wait a short amount of time for prometheus to collect the metrics, then go to the Prometheus console (on OpenShift, you can expose a route to it) and look for metrics with the prometheus query `{app="kiali-operator"}`

You should see metrics like these (plus a lot more):

```
workqueue_retries_total{app="kiali-operator",name="kiali-controller",namespace="kiali-operator",pod_name="kiali-operator-57588b7b45-4gcv4", ... }
```

```
go_threads{app="kiali-operator",namespace="kiali-operator",pod_name="kiali-operator-57588b7b45-4gcv4", ...}
```

### DIFF
--- a/operator/deploy/operator.yaml
+++ b/operator/deploy/operator.yaml
@@ -17,6 +17,9 @@ spec:
       labels:
         app: kiali-operator
         version: ${OPERATOR_VERSION_LABEL}
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "8383"
     spec:
       serviceAccountName: kiali-operator
       containers:
--- a/operator/manifests/kiali-community/1.10.0/kiali.crd.yaml
+++ b/operator/manifests/kiali-community/1.10.0/kiali.crd.yaml
@@ -1,0 +1,21 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: kialis.kiali.io
+  labels:
+    app: kiali-operator
+spec:
+  group: kiali.io
+  names:
+    kind: Kiali
+    listKind: KialiList
+    plural: kialis
+    singular: kiali
+  scope: Namespaced
+  subresources:
+    status: {}
+  version: v1alpha1
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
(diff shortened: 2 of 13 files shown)


hhovsepy:
On OpenShift Service Mesh 1.0.2 Metrics did not show for me on OCP 4.1.

jmazzitelli:
> On OpenShift Service Mesh 1.0.2 Metrics did not show for me on OCP 4.1.

Most likely because the productized ansible operator base image is old and does not have the metrics. So this won't work on the productized images, but should work with the images that are currently up on quay.io/kiali-operator.

So to test, you need to test with the community image, not the productized images.

Issue #1854, opened 2019-10-30
Enable correlation between metrics and traces in Details Pages / Metrics Tab


lucasponce:
Details pages offer a tab for detailed view of metrics (Inbound, Outbound, Runtimes).

One interesting feature could be to correlate the results of these metrics with other signals that Kiali collects.

On this case, this issue propose to enable a new button / option on these pages to enable to show / correlate two different signals within same chart as described in the following mockup:

![image](https://user-images.githubusercontent.com/1662329/67855799-6d681980-fb13-11e9-8a82-bf2f0020c524.png)

A "Show/Hide Traces" feature may enrich the interpretation of the information and "correlate" in the sense to visualize together both scenarios for analysis.

This feature is just scoped to traces, as a similar calculation was made before for the traces tab title, where kiali fetched the number of errors for a particular service.

lucasponce:
cc @jotak @aljesusg

jshaughn:
cc Everybody :)

jotak:
https://docs.google.com/document/d/1z2u1Kw7_WeHBL5aaR2hvKVbYr_qjdRRfoqQoKBtPOm4

jotak:
I'm splitting the work, with some pre-requisite in k-charted: https://github.com/kiali/kiali/issues/1939

lucasponce:
> I'm splitting the work, with some pre-requisite in k-charted: #1939

It's fine but also you can add subtasks in the same issue (without the need to create new issues).
Whatever it works .

There is no need to 1 issue per Pull Request and this could be an example where it can perfectly fit in a subtask (it that simplifies).

jotak:
done (all merged)

PR #1968, opened 2019-11-22
[Temp] Fetch traces / spans

jotak:
- Fetch traces and extract spans
- In case limit is reached, try better

Note: this is temporary as to be combined with Alberto's work
Also, TODO:
- update glide.yml for jaeger deps
- service is assumed to be "service.namespace", which is true for envoy-powered tracing but what about traces coming from the application code? Should assume the same will be used, or is there a convention (e.g. based on k8s annotations) to know declared service name(s)?
- there are some inconsistencies in the query results, looks like it's not returning all the results even when it doesn't reach the limit. Tested with istio 1.3.1 / all-in-one 1.12 / badger
- **unit tests**

Part of https://github.com/kiali/kiali/issues/1854

### DIFF
--- a/business/jaeger_helper.go
+++ b/business/jaeger_helper.go
@@ -7,7 +7,11 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"sort"
+	"strconv"
 	"time"
+
+	jaeger "github.com/jaegertracing/jaeger/model/json"
 
 	"github.com/kiali/kiali/appstate"
 	"github.com/kiali/kiali/config"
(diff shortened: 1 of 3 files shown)


jotak:
@aljesusg here's some temporary backend code that we can refactor together with your parts

jotak:
Uh, this was supposed to be a draft :-( putting DNM

business/jaeger_helper.go
```diff
@@ -84,6 +100,143 @@ func getErrorTracesFromJaeger(namespace string, service string, requestToken str
...
+	}
+
+	// Map to list
+	ret := []Span{}
+	for _, span := range spansMap {
+		ret = append(ret, span)
+	}
+	log.Infof("Found %d relevant spans", len(ret))
+	return ret, nil
+}
```
jotak: @jpkrohling @objectiser when the limit of traces has been reached (limit set to 100 at the moment) I'm trying here to fetch more relevant spans, taking the ones in error + the ones with duration above 90th percentile. WDYT? Does the algorithm sound ok ? Or do you have better suggestion?

jpkrohling: I think I'm missing some context. Are you then retrieving the traces where the relevant spans occurred? What makes a span more relevant?

jotak: @jpkrohling Well, I might be wrong, but on the UI we will display Spans and not Traces (see https://github.com/kiali/kiali-ui/pull/1560 ) - only the Span has really a duration, and is directly correlated to a Service, isn't it?

So, my endpoint here will be returning a list of spans to be displayed in the UI. Note that I'm adding the trace size (number of spans) within my span model so that it display smaller or bigger circles accordingly, hence keeping the notion of trace size. The intent is also to be able to jump to a Trace from a Span, maybe with a link to Jaeger UI, although it's not for this PR / will come later.

jpkrohling: I think I see what you mean now. One thing to consider: should you be selecting only the proxy spans?

jotak: @jpkrohling I don't think we should select only the proxy spans, as correlating user-defined spans also brings value I think.
However we still perform some selection as we're using the service name as defined by the proxies, which formats as "service.namespace". So if user-defined traces or spans use a different format, they wouldn't be caught here. We can probably enhance this later.

jotak:
Closing, will re-open based on Alberto's work

PR #1972, opened 2019-11-22
[hack] ignore version warnings

jmazzitelli:
for now, hack script will ignore version warnings since openshift installer/client releases do not report the expected versions.

This also bumps up to OS 4.2.8

### DIFF
--- a/hack/aws-openshift.sh
+++ b/hack/aws-openshift.sh
@@ -417,6 +417,9 @@ DEFAULT_CONTROL_PLANE_NAMESPACE="istio-system"
 # Default namespace where bookinfo is to be installed
 DEFAULT_BOOKINFO_NAMESPACE="bookinfo"
 
+# Temporarily ignore the version check by default since the current installer/client releases do not report correct versions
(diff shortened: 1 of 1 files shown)


jmazzitelli:
do NOT merge. problem in 4.2.8:
```
ERROR                                              
ERROR Error: Provider produced inconsistent result after apply 
ERROR                                              
ERROR When applying changes to module.vpc.aws_vpc_dhcp_options.main, provider "aws" 
ERROR produced an unexpected new value for was present, but now absent. 
ERROR                                              
ERROR This is a bug in the provider, which should be reported in the provider's own 
ERROR issue tracker.                               
ERROR                                              
FATAL failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply using Terraform 
```

jmazzitelli:
reverting back to using 4.2.4 - something isn't right with 4.2.8. But this does flip the default of the -ivc to true so we don't need to specify it all the time. We'll flip that default back to false when the OS releases get back to more meaningful "version" output for both installer and oc.

PR #1975, opened 2019-11-22
[operator] publish using 1.10.1 version

jmazzitelli:
we don't want to publish 1.10.0 to OLM. Change all manifests to use the upcoming 1.10.1


jmazzitelli:
When @jshaughn releases 1.10.1 (I'm sure it will happen this weekend), merge this.

jshaughn:
@jmazzitelli , I think you can close this as I don't anticipate a 1.10.1 release prior to 1.11.

PR #1979, opened 2019-11-25
Remove redundant type on struct definitions

xeviknal:
** Describe the change **

Removing redundant types when defining structs.
In our codebase we have definitions with both: redundant types and without. IMO, we should just use one way, without redundant type.
This PR removes redundant types. wdyt?


jotak:
+1
Actually I thought that it was necessary to have this redundancy... or is it a change in a recent version of go?

jshaughn:
> Actually I thought that it was necessary to have this redundancy... 

@jotak me too, something must have changed.

PR #1980, opened 2019-11-25
Create Jaeger Endpoints for Traces

aljesusg:
cc @jotak 

- [X] Clean Configuration
- [X] Clean discovery
- [X] Add endpoint get Traces
- [X] Add endpoint get specific Trace
- [X] Add endpoint get services
- [X] Move endpoint get Error Traces
- [X] Remove appstate for Jaeger (Now we'll get the configuration not more discovery)
- [x] Clean & Refactor

Require https://github.com/kiali/kiali/pull/1991


aljesusg:
## Get Traces
"/api/namespaces/{namespace}/services/{service}/traces",
**For this case all queryparams added to the endpoint will be passed to Jaeger so if we call /api/namespaces/{namespace}/services/{service}/traces?limit=2 this will call to jaeger with this queryparam**

### Sample
```json
[{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spans":[{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spanID":"b6416f18906f4c40","operation_name":"tracing.istio-system.svc.cluster.local:9411/*","start_time":1574683640482514,"duration":68972,"tags":[{"key":"node_id","type":"string","value":"router~10.0.2.7~istio-ingressgateway-f5cc4fb98-mnwb7.istio-system~istio-system.svc.cluster.local"},{"key":"zone","type":"string","value":"europe-west2-b"},{"key":"guid:x-request-id","type":"string","value":"6c1e1b74-16ed-9fcf-8183-0048bd055fc2"},{"key":"http.url","type":"string","value":"http://35.242.156.23:15032/jaeger/api/traces?service=istio-ingressgateway"},{"key":"http.method","type":"string","value":"GET"},{"key":"downstream_cluster","type":"string","value":"-"},{"key":"user_agent","type":"string","value":"Go-http-client/1.1"},{"key":"http.protocol","type":"string","value":"HTTP/1.1"},{"key":"request_size","type":"string","value":"0"},{"key":"response_size","type":"string","value":"9734"},{"key":"component","type":"string","value":"proxy"},{"key":"upstream_cluster","type":"string","value":"outbound|9411||tracing.istio-system.svc.cluster.local"},{"key":"http.status_code","type":"string","value":"200"},{"key":"response_flags","type":"string","value":"-"},{"key":"span.kind","type":"string","value":"client"},{"key":"internal.span.format","type":"string","value":"zipkin"}]},{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spanID":"9cae14eb2832f904","operation_name":"async outbound|9091||istio-policy.istio-system.svc.cluster.local egress","start_time":1574683640482712,"duration":3004,"tags":[{"key":"upstream_cluster","type":"string","value":"outbound|9091||istio-policy.istio-system.svc.cluster.local"},{"key":"component","type":"string","value":"proxy"},{"key":"grpc.status_code","type":"string","value":"0"},{"key":"span.kind","type":"string","value":"client"},{"key":"internal.span.format","type":"string","value":"zipkin"}]},{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spanID":"7e3d768f7445805d","operation_name":"/istio.mixer.v1.Mixer/Check","start_time":1574683640484284,"duration":64,"tags":[{"key":"component","type":"string","value":"istio-mixer"},{"key":"hostname","type":"string","value":"istio-policy-6cb85c5fc-plhmg"},{"key":"span.kind","type":"string","value":"server"},{"key":"internal.span.format","type":"string","value":"zipkin"}]}]}]
```

## Get Specific trace
"/api/namespaces/{namespace}/services/{service}/traces/{traceID}"

### Sample
```json
{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spans":[{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spanID":"b6416f18906f4c40","operation_name":"tracing.istio-system.svc.cluster.local:9411/*","start_time":1574683640482514,"duration":68972,"tags":[{"key":"node_id","type":"string","value":"router~10.0.2.7~istio-ingressgateway-f5cc4fb98-mnwb7.istio-system~istio-system.svc.cluster.local"},{"key":"zone","type":"string","value":"europe-west2-b"},{"key":"guid:x-request-id","type":"string","value":"6c1e1b74-16ed-9fcf-8183-0048bd055fc2"},{"key":"http.url","type":"string","value":"http://35.242.156.23:15032/jaeger/api/traces?service=istio-ingressgateway"},{"key":"http.method","type":"string","value":"GET"},{"key":"downstream_cluster","type":"string","value":"-"},{"key":"user_agent","type":"string","value":"Go-http-client/1.1"},{"key":"http.protocol","type":"string","value":"HTTP/1.1"},{"key":"request_size","type":"string","value":"0"},{"key":"response_size","type":"string","value":"9734"},{"key":"component","type":"string","value":"proxy"},{"key":"upstream_cluster","type":"string","value":"outbound|9411||tracing.istio-system.svc.cluster.local"},{"key":"http.status_code","type":"string","value":"200"},{"key":"response_flags","type":"string","value":"-"},{"key":"span.kind","type":"string","value":"client"},{"key":"internal.span.format","type":"string","value":"zipkin"}]},{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spanID":"9cae14eb2832f904","operation_name":"async outbound|9091||istio-policy.istio-system.svc.cluster.local egress","start_time":1574683640482712,"duration":3004,"tags":[{"key":"upstream_cluster","type":"string","value":"outbound|9091||istio-policy.istio-system.svc.cluster.local"},{"key":"component","type":"string","value":"proxy"},{"key":"grpc.status_code","type":"string","value":"0"},{"key":"span.kind","type":"string","value":"client"},{"key":"internal.span.format","type":"string","value":"zipkin"}]},{"traceID":"bd49b4dd0355a37db6416f18906f4c40","spanID":"7e3d768f7445805d","operation_name":"/istio.mixer.v1.Mixer/Check","start_time":1574683640484284,"duration":64,"tags":[{"key":"component","type":"string","value":"istio-mixer"},{"key":"hostname","type":"string","value":"istio-policy-6cb85c5fc-plhmg"},{"key":"span.kind","type":"string","value":"server"},{"key":"internal.span.format","type":"string","value":"zipkin"}]}]}

cc @jotak 
```

jotak:
There might be some changes I'd like to make, but instead of modifying this PR I suggest that we take it as is and iterate from there (beside usual code review and approval). See in my temporary PR https://github.com/kiali/kiali/pull/1968 there's a couple of things that are different and in particular I think we're interested in Spans and not in Traces to correlate more accurately to a service. But that can be done later.

business/jaeger.go
```diff
@@ -0,0 +1,261 @@
...
+		tags = append(tags, models.JaegerKeyValue{
+			Key:   tag["key"].(string),
+			Type:  tag["type"].(string),
+			Value: tag["value"].(string),
+		})
+	}
+	return
+}
+
+func getTraceFromJson(data map[string]interface{}) (trace models.JaegerTrace) {
```
jotak: I think it would be more typesafe (and easier to use/read) to unmarshal using jaeger's model instead of several nested maps. Is there a reason to not do it?

E.g. 

```go
import (
    jaeger "github.com/jaegertracing/jaeger/model/json"
)

// ...

	var traces RequestTrace
	if errMarshal := json.Unmarshal(body, &traces); errMarshal != nil {
		log.Errorf("Error unmarshalling Jaeger response: %s", errMarshal)
		return []jaeger.Trace{}, errMarshal
	}
// ...
```

business/jaeger.go
```diff
@@ -0,0 +1,261 @@
...
+		return
+	}
+	defer resp.Body.Close()
+	response, err = ioutil.ReadAll(resp.Body)
+	status = resp.StatusCode
+	return
+}
+
+func (in *JaegerService) getClient() (client http.Client, err error) {
+	timeout := time.Duration(1000 * time.Millisecond)
```
jotak: When I had a lot of traces, this timeout was too easily reached, so I moved to 5 seconds. Not sure what's the best to use ... or maybe we should let the caller specify the timeout, 1s is probably ok the get a single trace, but not to get all traces.

business/jaeger.go
```diff
@@ -0,0 +1,261 @@
+package business
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"github.com/gorilla/mux"
```
jotak: Usually we try to have a clear separation between handlers layer and business layer, and keep any of gorilla/mux and such dependency out of the business layer. Could we do the same here?
It means that business functions would not take *Request* argument or such, but more explicit arguments list instead.

status/discover.go
```diff
@@ -39,147 +37,6 @@ func getClient() (kubernetes.IstioClientInterface, error) {
 	return client, nil
 }
```
jotak: Is there no discovery anymore?

aljesusg: No, at least for Jaeger, we are going to force the user set the endpoint

business/jaeger.go
aljesusg: ofc, this is a first work I need to clean and order thigns

business/jaeger.go
aljesusg: clean and less memory but I understand your point that this is more not readable for developers

aljesusg:
I don't know if it's a good idead add jaeger in vendor see all these additions ^^ @jotak 

The code is cleaner now

jotak:
thanks @aljesusg , looks good, yes there's a lot of additions in vendors, personally I think it's ok to import all jaeger stuff but if you or other don't like I'm fine either way.

jotak:
@aljesusg I hope you're going to squash, because your commit messages are terrible!! 😂

aljesusg:
Yes  @jotak  sorry I don't know wtf is happening in my environment but swagger fails :(

jotak:
@aljesusg maybe you have a wrong version of go-swagger?
When I run swagger-gen on your PR I got swagger.json changed:

```diff
diff --git a/swagger.json b/swagger.json
index 75a064d7..53f4c883 100644
--- a/swagger.json
+++ b/swagger.json
@@ -2327,10 +2327,28 @@
         "tags": [
           "traces"
         ],
-        "operationId": "tracesDeatil",
+        "operationId": "tracesDetail",
+        "parameters": [
+          {
+            "type": "string",
+            "x-go-name": "Name",
+            "description": "The namespace name.",
+            "name": "namespace",
+            "in": "path",
+            "required": true
+          },
+          {
+            "type": "string",
+            "x-go-name": "Name",
+            "description": "The service name.",
+            "name": "service",
+            "in": "path",
+            "required": true
+          }
+        ],
         "responses": {
           "200": {
-            "$ref": "#/responses/serviceDetailsResponse"
+            "$ref": "#/responses/tracesDetailResponse"
           },
           "404": {
             "$ref": "#/responses/notFoundError"

```

business/layer.go
```diff
@@ -70,6 +71,7 @@ func Get(token string) (*Layer, error) {
 	if err != nil {
 		return nil, err
 	}
+	// Use an existing Jaeger client if it exists, otherwise create and use in the future
```
jotak: I guess this comment is to remove?

handlers/jaeger.go
```diff
@@ -47,20 +96,14 @@ func getJaegerInfo(requestToken string) (*models.JaegerInfo, int, error) {
...
-
 	// Be sure to copy config.Auth and not modify the existing
 	auth := jaegerConfig.Auth
 	if auth.UseKialiToken {
 		auth.Token = requestToken
 	}
 
-	if ha, err := canAccessURL(apiURL, &auth); !ha {
-		return nil, http.StatusServiceUnavailable, err
+	if ha, _ := canAccessURL(jaegerConfig.InClusterURL, &auth); !ha {
```
jotak: why not keeping the error detail? I guess it could give some valuable details to troubleshoot?

operator/roles/kiali-deploy/tasks/main.yml
```diff
@@ -171,12 +171,6 @@
   when:
     kiali_vars.external_services.grafana.in_cluster_url == ""
```
jotak: I guess we should do the same inference than in Grafana for namespace in `in_cluster_url` , no? (basically copy/paste/modify the block above)

hhovsepy:
This is how it looks:
Screenshot from 2019-11-27 12-45-32
Screenshot from 2019-11-27 12-46-12
Screenshot from 2019-11-27 12-47-44

business/jaeger.go
jotak: It's still there :)
Due to `GetJaegerTraceDetail`

jotak:
@hhovsepy this is using still using the old iframe I guess, as this PR is only backend work (or is there some frontend work that I missed?)

Maybe @aljesusg you could edit the PR/commit title to remove this ambiguity? (it's actually not (yet) about the iframe & react component, but about server endpoints and fetching jaeger)

aljesusg:
yep @hhovsepy there is not change in the frontend , the UI will change in another PR


totally @jotak

hhovsepy:
> @hhovsepy this is using still using the old iframe I guess, as this PR is only backend work (or is there some frontend work that I missed?)
yeah, good point, at least no regression shown in UI, but I could not find the old iFrame in my UI with this PR.

aljesusg:
Done @jotak

aljesusg:
> yeah, good point, at least no regression shown in UI, but I could not find the old iFrame in my UI with this PR.

Did you set the  url in the configuration ? There is not dicovery now. user must set the configuration

hhovsepy:
Screenshot from 2019-11-27 17-49-36
Screenshot from 2019-11-27 17-49-56
Screenshot from 2019-11-27 17-49-20

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -399,10 +396,7 @@ spec:
 #        username: ""
 #      enabled: true
 #      in_cluster_url: ""
-#      namespace: ""
 #      namespace_selector: true
-#      port: 16686
-#      service : ""
```
jmazzitelli: remove the corresponding default settings from the default/main.yaml here: https://github.com/kiali/kiali/blob/v1.10/operator/roles/kiali-deploy/defaults/main.yml#L121-L124

status/discover.go
```diff
@@ -39,147 +34,6 @@ func getClient() (kubernetes.IstioClientInterface, error) {
...
-
-func getPathURL(endpoint string) (path string) {
-	u, err := url.Parse(endpoint)
-	if err != nil {
-		return ""
-	}
-	return u.Path
-}
-
-func discoverTracingService() (service string, port int32) {
```
jmazzitelli: does this mean Kiali server no longer attempts to auto-discover jaeger?

operator/deploy/kiali/kiali_cr.yaml
jmazzitelli: this has been done.

status/discover.go
aljesusg: No, the user must set the configuration we dicussed about it

aljesusg: There are so uses cases like proxys etc....so this task is not perfect so we decided to let the user configure it

lucasponce: I think we may want to align with Prometheus here which is not auto-discovered.
Auto-Discovery has introduced more complexity and I think we may want to reduce the number of cornercases and have a more clean config process.

aljesusg:
@mazz @jotak are you ok with this? Can we merge it?

business/services.go
```diff
@@ -327,7 +327,7 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime
 	go func() {
 		// Maybe a future jaeger business layer
 		defer wg.Done()
-		eTraces, err = getErrorTracesFromJaeger(namespace, service, requestToken)
+		eTraces, err = in.businessLayer.Jaeger.jaeger.GetErrorTraces(namespace, service)
```
jotak: I find a bit odd writing `in.businessLayer.Jaeger.jaeger...`
As I understand, the first `Jaeger` is the Jaeger service and the second `jaeger` is the jaeger client?
I would say, the function `GetErrorTraces` should be promoted into the service, so that another service doesn't access (pseudo-)private members of the Jaeger service (ie. the client). No?

handlers/jaeger.go
```diff
@@ -2,14 +2,13 @@ package handlers
 
 import (
 	"errors"
-	"github.com/kiali/kiali/business"
+	"github.com/gorilla/mux"
```
jotak: The import order is incorrect ; remember this should be configured in IDE

handlers/services.go
```diff
@@ -104,13 +104,8 @@ func ServiceDetails(w http.ResponseWriter, r *http.Request) {
...
 		}()
 	}
-	requestToken, err := getToken(r)
-	if err != nil {
-		RespondWithError(w, http.StatusInternalServerError, "Token initialization error: "+err.Error())
-		return
-	}
 
-	serviceDetails, err := business.Svc.GetService(namespace, service, rateInterval, queryTime, requestToken)
+	serviceDetails, err := business.Svc.GetService(namespace, service, rateInterval, queryTime)
```
jotak: I don't understand why the `requestToken` has been removed. I think this is breaking an authentication case.

business/health_test.go
```diff
@@ -1,6 +1,7 @@
 package business
 
 import (
+	"github.com/kiali/kiali/jaeger/jaegertest"
```
jotak: here also, wrong import order

handlers/services.go
jotak: Oh, I think I understand, the token is passed before, when creating the client, right?

business/services.go
aljesusg: It's not working without that Jaeger I don't know why I need to check it

handlers/services.go
aljesusg: Yes The token is passed in the client

jaeger/traces.go
```diff
@@ -0,0 +1,106 @@
+package jaeger
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/kiali/kiali/config"
```
jotak: here also

status/versions.go
```diff
@@ -207,9 +207,14 @@ type p8sResponseVersion struct {
...
 func jaegerVersion() (*ExternalServiceInfo, error) {
+	jaegerConfig := config.Get().ExternalServices.Tracing
+
+	if !jaegerConfig.Enabled {
+		return nil, nil
+	}
 	product := ExternalServiceInfo{}
 	product.Name = "Jaeger"
-	product.Url = DiscoverJaeger()
+	product.Url = config.Get().ExternalServices.Tracing.URL
```
jotak: Please replace with `jaegerConfig.URL` as config is read just above (remember that getting config is not without cost as it involves a mutex)

business/services.go
jotak: @aljesusg I think it's because you didn't define `GetErrorTraces` in the business service

aljesusg: sorry but I don't understand why

```go
business/services.go:330:41: in.businessLayer.Jaeger.GetErrorTraces undefined (type JaegerService has no field or method GetErrorTraces)
```

aljesusg: the point is that it's defined in the client
```go
// ClientInterface for mocks (only mocked function are necessary here)
type ClientInterface interface {
	GetJaegerServices() (services []string, code int, err error)
	GetTraces(namespace string, service string, rawQuery string) (traces []*jaegerModels.Trace, code int, err error)
	GetTraceDetail(traceId string) (trace []*jaegerModels.Trace, code int, err error)
	GetErrorTraces(ns string, srv string) (errorTraces int, err error)
}
```
and I can set GetJaegerServices, GetTraces and GetTraceDetail but not GetErrorTraces

jotak: Look at your file `business/jaeger.go` : you didn't define that function

aljesusg: ooooo you're right

aljesusg: pushed, thanks @jotak

business/layer.go
```diff
@@ -80,7 +82,13 @@ func Get(token string) (*Layer, error) {
 		prometheusClient = prom
 	}
 
-	return NewWithBackends(k8s, prometheusClient), nil
+	// Create Jaeger client
```
jotak: I think this is not optimal because we're creating a jaeger client every time someone makes an access to the business layer (even when it's not going to use jaeger at all). Probably we should do some lazy loading instead.

As you wish, I'm fine if you prefer not doing it in this PR, but then we should create a ticket to not forget about that, wdyt?

aljesusg: Ok  @jotak , I added this like a client thinking about to improve integrations with more endpoints or providers, let me fix it

aljesusg: If user with admin rights requests client, this is stored with his token in the instance, if another user request with another rights what happen?  How kiali deal with this in prometheus? Am I missing code?

jotak:
@jmazzitelli note that discovery for grafana is still there

business/layer.go
jotak: With prometheus it's a bit different because Kiali has its own method to grant permission for metrics per user, so we always use Kiali's own service account token and never the user one. You can see that mentioned here: https://github.com/kiali/kiali/blob/master/prometheus/client.go#L51-L53 . Also in the kiali_cr documentation:

```yaml
#   use_kiali_token: When true and if auth.type is "bearer", Kiali Service Account token will be used for the API calls to Prometheus
```

whereas for Grafana or Jaeger it says:

```yaml
#   use_kiali_token: When true and if auth.type is "bearer", the same OAuth token used for authentication in Kiali will be used for the API calls to Grafana,
```

So that makes the things a bit easier to code for Prometheus.

jotak: @aljesusg I think we could have lazy loading that deals with the token by capturing it via a loader function. I can give some help if you want

jotak: That could be something around these lines:

- In Jaeger service:

```go
type JaegerLoader = func() (jaeger.ClientInterface, error)

type JaegerService struct {
	loader        JaegerLoader
	loaderErr     error
	jaeger        jaeger.ClientInterface
	businessLayer *Layer
}
```

Then we can create a function `client()` that performs the lazy loading. When the service wants to get the client, we can do:

```go
func (in *JaegerService) client() (jaeger.ClientInterface, error) {
	if in.jaeger != nil {
		return in.jaeger, nil
	} else if in.loaderErr != nil {
		return nil, in.loaderErr
	}
	in.jaeger, in.loaderErr = in.loader()
	return in.jaeger, in.loaderErr
}

// In GetJaegerServices for example:
	client, err := in.client()
	if err != nil {
		return nil, 0, err
	}
	return client.GetJaegerServices()
```

- And from layer.go to create the service with the loader, using the captured token:

```go
// ...
	// Create Jaeger client loader
	loader := func() (jaeger.ClientInterface, error) {
		return jaeger.NewClient(token)
	}
	return NewWithBackends(k8s, prometheusClient, loader), nil
```

And `NewWithBackends` creates the service with just the loader:

```go
        JaegerService{loader: jaegerLoader, businessLayer: temporaryLayer}
```

mazz:
I understand it’s an honest mistake but please do not @ me on your PRs

PR #1992, opened 2019-11-27
Move to Go1.12.13

### LINKED ISSUE #1970: Move to, at minimum, Go1.12.13
This issue is mainly for seeking consistency:

* In the README file (https://github.com/kiali/kiali#building) go 1.8.3 is being asked.
* In the Makefile (https://github.com/kiali/kiali/blob/master/Makefile#L26), also, go 1.8.3 is being required.

But:

* If the Jenkins box hasn't changed, the builds are using go 1.12.1.
* Travis is using go 1.12.1 (see https://github.com/kiali/kiali/blob/master/.travis.yml#L9)

The [Go Release Cycle](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance) gives the understanding that, currently, any version prior to go1.12 has reached EoL. So, we shouldn't be mentioning go1.8.

Additionally, since we have not been building Kiali with go 1.8, it even may no longer build with that version.

We should bump to go1.12.13 (which is the latest patch), but if we can move to 1.13 it would be great.

israel-hdez:
With these changes, Makefiles will ask for a minimum go version of 1.12.13. Additionally, the Makefile is changed so that it's possible to specify the go binary using an environment variable; ex: `GO=go1.13.4 make build`. This is useful if user has installed several go version (see https://golang.org/doc/install#extra_versions).

This also is:
* removing references to istiooc, because it's no longer supported in Maistra;
* a quick README.md cleanup;

Fixes #1970.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,12 +1,12 @@
-sudo: required
-
-services:
-- docker
+# sudo: required
+#
+# services:
+# - docker
 
 language: go
 
 go:
-- 1.12.1
+- 1.13.4
 
 notifications:
   irc:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@
 language: go
 
 go:
-- 1.13.4
+- 1.12.13
 
 notifications:
   irc:
--- a/Makefile
+++ b/Makefile
@@ -342,4 +342,4 @@ lint-install:
 ## lint: Runs golangci-lint
 # doc.go is ommited for linting, because it generates lots of warnings.
 lint:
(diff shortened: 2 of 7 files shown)


israel-hdez:
I reported an issue in `go-swagger`: https://github.com/go-swagger/go-swagger/issues/2122
I'll wait to see if there is a fix soon to move forward. Aternatively, I'll rearrange this PR to using go1.12.13.

I think it will be better if we can use go1.13 so that we have more time before another need to bump the go version.

israel-hdez:
Falling back to using go1.12.13, because I don't see any comments in the go-swagger issue.

israel-hdez:
I'm waiting for an update of the Jenkins box before merging.

Issue #1981, opened 2019-11-25
Traces & metrics correlation: metrics analysis


jotak:
An idea for more correlation (could be something like correlation v2): a user could right-click on a chart (e.g. on a span, or actually anywhere in a chart), which opens a menu with the option: "More correlation" (or a better name) which, based on the related timestamp, would start some analysis of *all* metrics in prometheus for the related service (or even other services in the same namespace) to find the ones that have "unusual" changes close to that timestamp. The goal would to help the user to locate "strange behaving" metrics among all the available ones - and possibly correlate things beyond a single service.

I think this analysis can be done by comparing the standard deviation close to that timestamp. I've tested that locally and it seems to give interesting results.

Goal of the RS would be to determine if the kind of analysis we can do would be promising to actually troubleshoot.

lucasponce:
This feature is progressed in the traces tab, where users can check if a particular trace/span is "good/bad" compared with the standard deviation.

![image](https://user-images.githubusercontent.com/1662329/128191083-60cdbde3-38bd-4f8f-9eba-a72f86f7e527.png)

Also the charts give some level of correlation to identify suspected peaks that a user may want to zoom in to investigate better:

![image](https://user-images.githubusercontent.com/1662329/128191214-8931dc04-f691-4220-aab2-fb4ce58d5878.png)

Today spans are correlated with logs as well, so there is some good level of correlation support.

Probably, in the future, Kiali way want to integrate with more powerful analytics solutions and show them rather than compute from metrics itself.

So I think we can close this one in favor to future integration with other solutions for this particular space (metrics analysis).

PR #1994, opened 2019-11-29
Jaeger: fetch spans

jotak:
- Get spans extracted from traces
- Small refactor in fetch traces to reuse queries when fetching spans
- In case limit is reached, try better

Also, TODO:
<strike>- service is assumed to be "service.namespace", which is true for envoy-powered tracing but what about traces coming from the application code? Should assume the same will be used, or is there a convention (e.g. based on k8s annotations) to know declared service name(s)?</strike> actually not really a problem because we're supposed to be in a service mesh, so with envoy sidecars, and this service name pattern is used to query *traces* (not spans) which should in theory contain also user-defined spans as well - and these one are kept in the returned spans list

- there are some inconsistencies in the query results, looks like it's not returning all the results even when it doesn't reach the limit. Tested with istio 1.3.1 / all-in-one 1.12 / badger

Part of https://github.com/kiali/kiali/issues/1854

Frontend: https://github.com/kiali/kiali-ui/pull/1560


jaeger/spans.go
```diff
@@ -0,0 +1,121 @@
...
+type Span struct {
+	jaegerModels.Span
+	TraceSize int `json:"traceSize"`
+}
+
+func getSpans(client http.Client, endpoint *url.URL, namespace, service, startMicros, endMicros string) ([]Span, error) {
+	u := endpoint
+	queryService := service
+	if config.Get().ExternalServices.Tracing.NamespaceSelector {
+		queryService = service + "." + namespace
```
lucasponce: Is this correct ?
I understood from some comments with @aljesusg we are going to fetch the "internal url" from config, instead to build the service using "namespace + selector" then we can reach same scenario but in and out cluster.

lucasponce: ah, ok, nevermind. It adds the tracing namespace selector always ?

jaeger/spans.go
```diff
@@ -0,0 +1,121 @@
...
+	return spans, nil
+}
+
+func tracesToSpans(traces []jaegerModels.Trace, service, namespace string) []Span {
+	spans := []Span{}
+	for _, trace := range traces {
+		// First, get the desired processes for our service
+		processes := make(map[jaegerModels.ProcessID]bool)
+		for pId, process := range trace.Processes {
+			if process.ServiceName == service || process.ServiceName == service+"."+namespace {
```
lucasponce: minor, I'd add a local variable like in https://github.com/kiali/kiali/pull/1994/files#diff-f5d5eff94f17a5f6374cb3b80d4608e2R30 as it's same concept here.

jaeger/spans.go
```diff
@@ -0,0 +1,121 @@
...
+					TraceSize: len(trace.Spans),
+				})
+			}
+		}
+	}
+	log.Infof("Found %d spans in the %d traces for service %s", len(spans), len(traces), service)
+	return spans
+}
+
+func findRelevantSpans(client http.Client, spansSample []Span, u *url.URL, q url.Values, service, namespace string) ([]Span, error) {
```
lucasponce: Nice logic here.
I know it's quite far, but conceptually it opens to add things like this in the future:
https://www.hawkular.org/blog/2017/08/alerts-nelson-rules.html

jaeger/spans.go
jotak: To be honest I'm not 100% sure why this config of NamespaceSelector (a bool flag) exists, I guess someone asked for that at some point. This pattern of naming `service + "." + namespace` is what Envoy does when it creates traces, so that's why we need to do the same to identify service names as they're stored by Jaeger.

jaeger/spans.go
jotak: Thanks! This is very interesting in particular in regard of what I'd like to experiment here https://github.com/kiali/kiali/issues/1981 . We should definitely talk about that tomorrow if possible!

jaeger/spans.go
lucasponce: Yes, I remember @aljesusg comments about it.

aljesusg: Yes if namespaceSelector is true (by default) we look by service.namespace.

With false we look by service, there are users that not use the envoy to do this.

lucasponce:
Note this will need to bump once k-charted 0.4.1-rc0 is moved to 0.4.1.

business/jaeger.go
```diff
@@ -33,15 +33,23 @@ func (in *JaegerService) GetJaegerServices() (services []string, code int, err e
 	return client.GetJaegerServices()
 }
 
-func (in *JaegerService) GetJaegerTraces(ns string, srv string, query string) (traces []*jaegerModels.Trace, code int, err error) {
+func (in *JaegerService) GetJaegerSpans(namespace, service, startMicros, endMicros string) ([]jaeger.Span, error) {
```
lucasponce: Minor, but I see some APIs using the "jaeger." package and other the imported "jaegerModels."
It would be nice to some kind of alignment even if that means just to create an alias to some "jaegerModels" to consolidate models used for the business logic.

jotak: This `jaeger` is a package defined in Kiali whereas `jaegerModels` is an import alias of a package from jaeger lib, so, how would we do? Alias everything we use from jaeger lib within our jaeger package?

jotak:
@lucasponce now that release 1.11.0 is done, I think it's a good time to merge?

lucasponce:
> @lucasponce now that release 1.11.0 is done, I think it's a good time to merge?

Yes, my only question is if we want to sync the k-charted dependencies here and in the kiali-ui or it doesn't matter.

jotak:
@lucasponce I think it doesn't matter, I created another task to sync & bump k-charted : https://github.com/kiali/kiali/issues/2016

lucasponce:
ok

PR #1996, opened 2019-11-29
Implement token authentication

israel-hdez:
This is adding a login mechanism inspired on the [kubernetes dashboard login view](https://github.com/kubernetes/dashboard/blob/master/docs/user/access-control/README.md#login-view).

Instead of providing a username and a password, Kiali asks for a token. The provided token is meant to be a service account token and Kiali will use it during the session of the user to make calls to the cluster API.

By using the provided token to make calls to the cluster API, the cluster's RBAC mechanisms takes effect and write actions in Kiali can be limited using Roles and RoleBindings.

Currently, in Kubernetes, there is a limitation: the resulting _accessible namespaces_ given the Kiali ConfigMap should all be, at least, readable via the kiali-viewer role. Kiali will fail if a user logs in using a token that does not have read privileges in, at least, one of the "accessible namespaces". This limitation is because the Kubernetes API always returns all namespaces regardless if the user does not have access to any of the resources on it. In contrast, in OpenShift Kiali takes advantage of the Projects API which truncates the list to the readable namespaces.

Despite the limitation with Kubernetes, the new Token auth mechanism allows to login with privileges to modify the cluster, or to login in view mode by taking advantage of the cluster RBAC.

See also kiali/kiali-ui#1572

Related #1139

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -15,6 +15,7 @@ import (
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/ldap"
 	"github.com/kiali/kiali/log"
+	"github.com/kiali/kiali/util"
 )
 
 const (
@@ -210,6 +211,62 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
 	return true
 }
 
+func performTokenAuthentication(w http.ResponseWriter, r *http.Request) bool {
+	err := r.ParseForm()
+
+	if err != nil {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, fmt.Errorf("error parsing form info: %+v", err))
+		return false
+	}
+
+	token := r.Form.Get("token")
+
+	if token == "" {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, "Token is empty.")
+		return false
+	}
+
+	business, err := business.Get(token)
+	if err != nil {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, "Error retrieving the OAuth package.")
+	}
+
+	// Using the namespaces API to check if token is valid. In Kubernetes, the version API seems to allow
+	// anonymous access, so it's not feasible to use the version API for token verification.
+	_, err = business.Namespace.GetNamespaces()
+	if err != nil {
+		RespondWithJSONIndent(w, http.StatusUnauthorized, "Token is not valid or is expired: "+err.Error())
+		return false
+	}
+
+	timeExpire := util.Clock.Now().Add(time.Second * time.Duration(config.Get().LoginToken.ExpirationSeconds))
+	tokenClaims := config.IanaClaims{
+		SessionId: token,
+		StandardClaims: jwt.StandardClaims{
+			Subject:   "token",
+			ExpiresAt: timeExpire.Unix(),
+			Issuer:    config.AuthStrategyTokenIssuer,
+		},
+	}
+	tokenString, err := config.GetSignedTokenString(tokenClaims)
+	if err != nil {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, err)
+		return false
+	}
+
+	tokenCookie := http.Cookie{
+		Name:     config.TokenCookieName,
+		Value:    tokenString,
+		Expires:  timeExpire,
+		HttpOnly: true,
+		// SameSite: http.SameSiteStrictMode, ** Commented out because unsupported in go < 1.11
+	}
+	http.SetCookie(w, &tokenCookie)
+
+	RespondWithJSONIndent(w, http.StatusOK, TokenResponse{Token: tokenString, ExpiresOn: timeExpire.Format(time.RFC1123Z), Username: "token"})
+	return true
+}
+
 func checkOpenshiftSession(w http.ResponseWriter, r *http.Request) (int, string) {
 	tokenString := getTokenStringFromRequest(r)
 	if claims, err := config.GetTokenClaimsIfValid(tokenString); err != nil {
@@ -383,6 +440,8 @@ func (aHandler AuthenticationHandler) Handle(next http.Handler) http.Handler {
 		case config.AuthStrategyLogin:
 			statusCode = checkKialiSession(w, r)
 			token = aHandler.saToken
+		case config.AuthStrategyToken:
+			statusCode, token = checkOpenshiftSession(w, r)
 		case config.AuthStrategyAnonymous:
 			log.Tracef("Access to the server endpoint is not secured with credentials - letting request come in. Url: [%s]", r.URL.String())
 			token = aHandler.saToken
@@ -425,6 +484,8 @@ func Authenticate(w http.ResponseWriter, r *http.Request) {
 		if !performKialiAuthentication(w, r) {
 			writeAuthenticateHeader(w, r)
 		}
+	case config.AuthStrategyToken:
+		performTokenAuthentication(w, r)
 	case config.AuthStrategyAnonymous:
 		log.Warning("Authentication attempt with anonymous access enabled.")
 	case config.AuthStrategyLDAP:
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -240,11 +240,22 @@ func performTokenAuthentication(w http.ResponseWriter, r *http.Request) bool {
 		return false
 	}
 
+	// Now that we know that the ServiceAccount token is valid, parse/decode it to extract
+	// the name of the service account. The "subject" is passed to the front-end to be displayed.
+	tokenSubject := "token" // Set a default value
+
+	parsedClusterToken, _, err := new(jwt.Parser).ParseUnverified(token, &jwt.StandardClaims{})
+	if err == nil {
+		tokenSubject = parsedClusterToken.Claims.(*jwt.StandardClaims).Subject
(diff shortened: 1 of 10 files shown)


jshaughn:
@israel-hdez @jmazzitelli If we have this feature do we still need the view-only mode feature?

handlers/authentication.go
```diff
@@ -210,6 +211,62 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
 	return true
 }
 
+func performTokenAuthentication(w http.ResponseWriter, r *http.Request) bool {
+	err := r.ParseForm()
+
+	if err != nil {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, fmt.Errorf("error parsing form info: %+v", err))
```
jshaughn: minor, here and in other places you're not returning a JSON string, are you?  I'd suggest:

`RespondWithDetailedError(w, http.StatusInternalServerError, "Error parsing form info", err.Error())`

handlers/authentication.go
```diff
@@ -210,6 +211,62 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
...
+	token := r.Form.Get("token")
+
+	if token == "" {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, "Token is empty.")
+		return false
+	}
+
+	business, err := business.Get(token)
+	if err != nil {
+		RespondWithJSONIndent(w, http.StatusInternalServerError, "Error retrieving the OAuth package.")
```
jshaughn: OAuth?

handlers/authentication.go
```diff
@@ -210,6 +211,62 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
...
+		RespondWithJSONIndent(w, http.StatusInternalServerError, err)
+		return false
+	}
+
+	tokenCookie := http.Cookie{
+		Name:     config.TokenCookieName,
+		Value:    tokenString,
+		Expires:  timeExpire,
+		HttpOnly: true,
+		// SameSite: http.SameSiteStrictMode, ** Commented out because unsupported in go < 1.11
```
jshaughn: I thought we required go 1.12?  Or is that still pending?

handlers/authentication.go
```diff
@@ -383,6 +440,8 @@ func (aHandler AuthenticationHandler) Handle(next http.Handler) http.Handler {
 		case config.AuthStrategyLogin:
 			statusCode = checkKialiSession(w, r)
 			token = aHandler.saToken
+		case config.AuthStrategyToken:
+			statusCode, token = checkOpenshiftSession(w, r)
```
jshaughn: Is it always Openshift?

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -94,13 +94,18 @@ spec:
 #  auth:
 #
 # Determines what authentication strategy to use when users log into Kiali.
-# Options are "login", "anonymous", "openshift".
+# Options are "login", "anonymous", "openshift", "token".
```
jshaughn: ldap?

jmazzitelli:
> @israel-hdez @jmazzitelli If we have this feature do we still need the view-only mode feature?

i think so - an admin might want to force all users (regardless of who they are) to only be able to view the mesh and not modify it. The view_only_mode setting can be considered independent of this feature

operator/deploy/kiali/kiali_cr.yaml
jmazzitelli: yeah, while we are here, let's add the missing `ldap` , too. We must have forgot this spot when we added ldap support.

handlers/authentication.go
israel-hdez: Ops, yes! I fixed it.
This came from the old PR (https://github.com/kiali/kiali/pull/1216) when the detailed error function didn't exist yet.

handlers/authentication.go
israel-hdez: OAuth is unrelated :)
Changed to "business layer".

handlers/authentication.go
israel-hdez: Still pending, but PR is ready: https://github.com/kiali/kiali/pull/1992
So, I uncommented. I think it won't cause any problems.

handlers/authentication.go
israel-hdez: `checkOpenshiftSession` works OK.
But I created a dedicated function, per my comment in `performTokenAuthentication` (`In Kubernetes, the version API seems to allow anonymous access, so it's not feasible to use the version API for token verification`)

Good catch.

operator/deploy/kiali/kiali_cr.yaml
israel-hdez: Yes, I tried to add ldap in all places where it was omitted. I missed this one.
Thanks for pointing it.

israel-hdez:
> > @israel-hdez @jmazzitelli If we have this feature do we still need the view-only mode feature?
> 
> i think so - an admin might want to force all users (regardless of who they are) to only be able to view the mesh and not modify it. The view_only_mode setting can be considered independent of this feature

I agree with Mazz.  But also note that, AFAIK, the "view-only" mode is just creating RoleBindings to the kiali-viewer role instead of binding to kiali role. So, "view-only" is for convenience. Users can adjust rolebindinds thereafter. (The website has a note about this; read https://kiali.io/documentation/getting-started/#_reducing_permissions_in_openshift)

israel-hdez:
BTW, the "token" auth strategy is compatible with OpenShift. So, you can try this PR on OS.

jmazzitelli:
@israel-hdez this is a really cool addition. I want to test this out. I'm in the middle of working on another PR (was supposed to be only 30 minutes, taking me over a day :) but once I'm done that, I will test this PR.

jmazzitelli:
Hold on merging.. I'm going to see if I can write a molecule test for this. Not sure how easy it will be, but I think it would be nice to have a molecule test for this.

jshaughn:
One more thing, you'll need a kiali.io PR as well, for this section: https://kiali.io/documentation/getting-started/#_login_options

jmazzitelli:
see https://github.com/israel-hdez/swscore/pull/18 for the commit that adds the token molecule test. it all passes.

jmazzitelli:
I believe @israel-hdez your PR branch is the one that needs to be rebased: https://github.com/israel-hdez/swscore/commits/kiali-3081

But as I say, you could just take the 4 files in my PR branch and directly copy them over to your PR branch and don't worry about merging my PR. The files are:

* operator/molecule/asserts/token-test/assert-token-access.yml
* operator/molecule/token-test/molecule.yml
* operator/molecule/token-test/playbook.yml
* operator/molecule/token-test/prepare-token.yml

jmazzitelli:
note: if you want to run the molecule test yourself, build and push the kiali from this PR into your AWS cluster using "make ocp-push" and then in the molecule.yml, you have to comment out/uncomment the lines that switch from the quay.io images to the AWS image repo image (see e.g. operator_image_name in molecule.yml).

operator/deploy/deploy-kiali-operator.sh
```diff
@@ -462,13 +466,13 @@ CLIENT_EXE=$(which istiooc 2>/dev/null || which oc 2>/dev/null)
 if [ "$?" == "0" ]; then
   echo "Using 'oc' located here: ${CLIENT_EXE}"
   _AUTH_STRATEGY_DEFAULT="openshift"
-  _AUTH_STRATEGY_PROMPT="Choose a login strategy of either 'login', 'openshift' or 'anonymous'. Use 'anonymous' at your own risk. [${_AUTH_STRATEGY_DEFAULT}]: "
+  _AUTH_STRATEGY_PROMPT="Choose a login strategy of either 'login', 'openshift', 'token', 'ldap' or 'anonymous'. Use 'anonymous' at your own risk. [${_AUTH_STRATEGY_DEFAULT}]: "
```
cfcosta: Great, I was going to ask the changes on the script as well :)

jmazzitelli:
@israel-hdez a two line change to the molecule test so it can now test the new feature you just added where the login response gives you the username of the token. Merge this PR: https://github.com/israel-hdez/swscore/pull/19

israel-hdez:
@jmazzitelli It's merged!

israel-hdez:
OK, I'm merging. This seems good enough.

@mattmahoneyrh I'll appreciate any post-merge feedback, if you find something.

githubrotem:
Hi
Does anyone knows when Kiali will support full oidc authentication, similar like Kubernetes dashboard?

Current token implementation only support manual token usage via gui/api but its not possible to perform pass throughput login with the service account token, you need to exchange it for a token with Kiali

Thanks

jmazzitelli:
> Does anyone knows when Kiali will support full oidc authentication, similar like Kubernetes dashboard?

Currently, this is not being worked on. We are hoping those in the community that wish to have further enhancements to the login/auth functionality to contribute that to the Kiali codebase.

See this issue for tracking this future work: https://github.com/kiali/kiali/issues/2056

[side note: Kiali is not bereft of token auth, however, as it now does have token auth similar to [this Kubernetes dashboard login feature](https://github.com/kubernetes/dashboard/blob/master/docs/user/access-control/README.md#bearer-token).]

Issue #1855, opened 2019-10-30
Replace traces iFrame with a Kiali React component


lucasponce:
Service Details page offers a "traces" tab providing an integration with Jaeger/Tracing solution.

Today, Kiali implements the following features:
- Number of errors reported by Jaeger for a specific service.
- Customization of the search controls, to be aligned with the rest of Kiali controls.
- An IFrame to integrate Jaeger within Kiali, to offer an unified experience to the Kiali's users.

The iFrame solution was intended as something temporal as it has several disadvantages:

- When Kiali and Jaeger are protected, a Single Sign On solution is needed, otherwise, the iFrame creates more technical problem rather than offers added value.

Using an iFrame limits Kiali features to progress and add / enrich traces info with: Service Details and correlation with other signals (metrics/health) already collected by Kiali.

So, this issue is to replace the current iFrame with a component as is described in the mockup:

![image](https://user-images.githubusercontent.com/1662329/67858005-714a6a80-fb18-11e9-9abe-ff1f7add4394.png)

- The first red rectangle will provide a similar timeline representation of traces, x-asis using the timeline, y-line using the response time, circle size will represent request size.
- Additionally, in this component, Kiali can differentiate in the timeline traces with errors vs healthy.
- When selecting an element in the timeline, the details of the trace should be showed in the second red rectangle, combining tracing info (provided by Jaeger, summary of spans) with relevant Service Details info (i.e. we can start with the health of the service linked/attached with the service). Also there should be a link to see this trace in Jaeger to examine further details.

Also, this issue propose to limit the Jaeger integration in the context of the Services tab.

lucasponce:
cc @aljesusg 
We have commented several ideas on this, I hope this is good start for this effort.

lucasponce:
Completed via https://github.com/kiali/kiali-ui/pull/1577 and others PRs in backend.

PR #2017, opened 2019-12-10
Move jaegerInfo and return if there is an integration and errorMessage

aljesusg:
Related with https://github.com/kiali/kiali/issues/1855

### DIFF
--- a/handlers/jaeger.go
+++ b/handlers/jaeger.go
@@ -1,28 +1,22 @@
 package handlers
 
 import (
-	"errors"
 	"net/http"
-	"time"
 
 	"github.com/gorilla/mux"
-
-	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/log"
-	"github.com/kiali/kiali/models"
-	"github.com/kiali/kiali/util/httputil"
 )
 
 // Get JaegerInfo provides the Jaeger URL and other info, first by checking if a config exists
 // then (if not) by inspecting the Kubernetes Jaeger service in Istio installation namespace
 func GetJaegerInfo(w http.ResponseWriter, r *http.Request) {
-	requestToken, err := getToken(r)
+	business, err := getBusiness(r)
 	if err != nil {
 		RespondWithError(w, http.StatusInternalServerError, "Token initialization error: "+err.Error())
 		return
 	}
 
-	info, code, err := getJaegerInfo(requestToken)
+	info, code, err := business.Jaeger.GetJaegerInfo()
 	if err != nil {
 		log.Error(err)
 		RespondWithError(w, code, err.Error())
@@ -49,7 +43,7 @@ func GetJaegerServices(w http.ResponseWriter, r *http.Request) {
 func TraceServiceDetails(w http.ResponseWriter, r *http.Request) {
 	business, err := getBusiness(r)
 	if err != nil {
-		RespondWithError(w, http.StatusInternalServerError, "Services initialization error: "+err.Error())
+		RespondWithError(w, http.StatusInternalServerError, "Trace Service Details initialization error: "+err.Error())
 		return
 	}
 	params := mux.Vars(r)
@@ -67,7 +61,7 @@ func TraceServiceDetails(w http.ResponseWriter, r *http.Request) {
 func TraceDetails(w http.ResponseWriter, r *http.Request) {
 	business, err := getBusiness(r)
 	if err != nil {
-		RespondWithError(w, http.StatusInternalServerError, "Services initialization error: "+err.Error())
+		RespondWithError(w, http.StatusInternalServerError, "Trace Detail initialization error: "+err.Error())
 		return
 	}
 	params := mux.Vars(r)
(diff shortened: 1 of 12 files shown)


lucasponce:
Swagger seems failing.

lucasponce:
Code LGTM. But swagger needs to be updated correctly.

handlers/jaeger.go
```diff
@@ -34,7 +28,7 @@ func GetJaegerInfo(w http.ResponseWriter, r *http.Request) {
 func GetJaegerServices(w http.ResponseWriter, r *http.Request) {
 	business, err := getBusiness(r)
 	if err != nil {
-		RespondWithError(w, http.StatusInternalServerError, "Services initialization error: "+err.Error())
+		RespondWithError(w, http.StatusInternalServerError, "Token initialization error: "+err.Error())
```
jotak: Not sure why it's changed to `Token` (here and below), the function `getBusiness` may fail in case of client initialization error for instance, it's not necessarily about the token

jaeger/info.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"errors"
+	"net/http"
+	"net/url"
+	"path"
+
+	"github.com/kiali/kiali/config"
+)
+
+func getJaegerInfo(client http.Client, endpoint *url.URL) (*JaegerInfo, int, error) {
```
jotak: can we elaborate a little bit with the changes in our dependencies to Jaeger? I'm not sure to have followed everything:

- do we still need a configured external URL or not? In other words, are we going to keep some links to jaeger from within the browser?

- What does mean "Integration" here: is it that any kind of integration with Jaeger is enabled/disabled when this flag is true/false? For instance in the Spans overlay that I did, is it supposed to be turned off when Integration is false? Even when the in_cluster_url is working correctly but not the external one?

jaeger/info.go
```diff
@@ -0,0 +1,48 @@
...
+	externalUrl := jaegerConfig.URL
+	if externalUrl == "" {
+		return nil, http.StatusServiceUnavailable, errors.New("Jaeger URL is not set in Kiali configuration")
+	}
+
+	if jaegerConfig.InClusterURL == "" {
+		return nil, http.StatusServiceUnavailable, errors.New("Jaeger URL in cluster is not set in Kiali configuration")
+	}
+
+	// Be sure to copy config.Auth and not modify the existing
```
jotak: I guess this residual comment can be removed :)

jaeger/info.go
aljesusg: In that way,

If user configure the InClusterUrl, the user will see the traces and the stuff provided by Kiali,
If user configure the external URL the user will see the links to the Jaeger UI in kiali.

Casuistics

- User can set only the internal URL => The user will see the traces tab
- User can set only the  external URL => The user will see the traces tab but this tab will be a link to the JaegerUI traces for that service, and he'll see the distributed tracing link in the menu
- User can set boths

For example /api/jaeger returns

```json
{
"url":"http://localhost:16686/",
"namespace_selector":true,
"integration":false,
"integration_message":"Error with internal connection with Jaeger: Get http://localhost:16686/api/services: dial tcp [::1]:16686: connect: connection refused"}
```

This means that user will see links to Jaeger UI in Kiali.
The integration is false, this means that kiali can't reach the jaeger service internally and the integration_message throw the cause, so the user won't see the traces tab with the scatter plot, this tab will be a link to the Jaeger UI
And the namespace_selector indicate that we need to look the traces by service.namespace

cc @lucasponce @jotak

handlers/jaeger.go
aljesusg: You're right I'll change this

jaeger/client.go
```diff
@@ -17,11 +15,19 @@ import (
...
-	GetTraceDetail(traceId string) (trace []jaegerModels.Trace, code int, err error)
+	GetJaegerInfo() (*JaegerInfo, int, error)
+	GetTraces(namespace string, service string, rawQuery string) (traces *JaegerResponse, code int, err error)
+	GetTraceDetail(traceId string) (trace *JaegerResponse, code int, err error)
 	GetErrorTraces(ns string, srv string) (errorTraces int, err error)
 }
 
+type JaegerInfo struct {
+	URL                string `json:"url"`
+	NamespaceSelector  bool   `json:"namespace_selector"`
```
jotak: we use camel case for json (unless there's a special reason to do otherwise) ; should be changed here?

jaeger/info.go
jotak: thanks!

jaeger/client.go
aljesusg: changed I forgot it

jaeger/info.go
```diff
@@ -0,0 +1,46 @@
...
+		return nil, http.StatusServiceUnavailable, errors.New("Jaeger URL in cluster is not set in Kiali configuration")
+	}
+
+	u := endpoint
+	u.Path = path.Join(u.Path, "/api/services")
+
+	_, code, err := makeRequest(client, u.String(), nil)
+	if err != nil || code != 200 {
+		integration = false
+		error = "Error with internal connection with Jaeger: " + err.Error()
```
jotak: you will have an issue here when `err` is `nil` (when `err` is `nil` and `code` is anything but 200, the `if` condition is true)

aljesusg: yep, the point is taht jaeger returns 400 if there is any error and we are showing these errors in kliali UI,

for example :
/api/traces?end=1578581310315000&limit=20&lookback=1h&maxDuration=12ms&minDuration=22ms&service=details.bookinfo&start=1578577710315000


return this json with code 400, that means that jaeger is working but the query is wrong
```json
{"data":null,"total":0,"limit":0,"offset":0,"errors":[{"code":400,"msg":"'maxDuration' should be greater than 'minDuration'"}]}
```

jotak: Yes, but given your `if` condition above, `err` can be nil. So your code will crash (`err.Error()` will crash)

aljesusg: o.O you're right OMG

jaeger/traces.go
```diff
@@ -6,78 +6,81 @@ import (
...
 }
 
-func getErrorTraces(client http.Client, endpoint *url.URL, namespace string, service string) (errorTraces int, err error) {
+func getErrorTraces(client http.Client, endpoint *url.URL, namespace string, service string, interval string) (errorTraces int, err error) {
 	errorTraces = 0
 	err = nil
 	u := endpoint
 	u.Path = path.Join(u.Path, "/api/traces")
-
+	log.Errorf("%s", interval)
```
lucasponce: This is going to print always a error in the log.
Perhaps this was left for debugging ?

jaeger/traces.go
```diff
@@ -6,78 +6,81 @@ import (
...
+	conv, _ := strconv.ParseInt(interval[0:len(interval)-1], 10, 64)
```
lucasponce: Perhaps better to protect against an len(interval) == 0 ?

jaeger/traces.go
```diff
@@ -6,78 +6,81 @@ import (
...
+	inter := conv * 1000 * 1000
+	unit := interval[len(interval)-1:]
```
lucasponce: same here, I think

aljesusg:
> Missed protection in one of my comments, I think it deserves to fix it before to merge.

Lucas I am checking before that code if interval len is = then return err so I don't need to check it again

```go
if len(interval) == 0 {
		return -1, errors.New("Empty interval getting error traces")
	}
	conv, err := strconv.ParseInt(interval[0:len(interval)-1], 10, 64)
	if err != nil {
		return -1, err
	}
	inter := conv * 1000 * 1000
	unit := interval[len(interval)-1:]
	for _, u := range []string{"s", "m", "h"} {
		if u == unit {
			break
		}
		inter *= 60
	}
```

jaeger/info.go
jotak: @aljesusg you wrote ` If user configure the external URL the user will see the links to the Jaeger UI in kiali.` but it doesn't seem to be true today: when no external URL is provided, I don't see the Traces tab at all (whereas I can see the spans within metrics charts). Should I open a ticket?

aljesusg: Yes please open a ticket, if you have the internal configured but not the external one you should see the traces tab.

PR #2020, opened 2019-12-10
issue 1698 do not log a warning if its expected

### LINKED ISSUE #1698: Logs pollution tls.go
My Kiali logs are full of the same following message:
`1 tls.go:56] GetMeshPolicies failed during a TLS validation. Probably Kiali doesn't have cluster permissions. Error: meshpolicies.authentication.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list meshpolicies.authentication.istio.io at the cluster scope: no RBAC policy matched`

We should avoid logging it again and again. Maybe just log it once then set a flag to not log it again?

Or even better: have a flag to disable permanently this periodic check, and that would be set by the operator when Kiali has no cluster permissions ?

jmazzitelli:
fixes #1698 

I took the path of least resistance. There should be no need to log this unless we have a cluster role and we know we have a cluster role when accessible namespaces is \*\* (and we know we do not have a cluster role when accessible namespaces is NOT \*\*).

This means it's a very small, localized change. No need to go through a huge change in the operator and adding a config setting to the configmap just to turn off this message - that seems like killing an ant with a sledgehammer.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -290,7 +290,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			} else {
 				// This query can return false if user doesn't have cluster permissions
 				// On this case we log internally the error but we return an empty list
-				log.Warningf("GetMeshPolicies failed during a GetIstioConfigList. Probably Kiali doesn't have cluster permissions. Error: %s", mpErr)
+				checkForbidden("GetMeshPolicies", mpErr, "probably Kiali doesn't have cluster permissions")
 			}
 		}
 	}()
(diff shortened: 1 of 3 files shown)


jotak:
I'm wondering btw (ping @xeviknal @lucasponce see my comment above) : do you think it's normal that I cannot get GetAuthorizationDetails and GetMeshPolicies when I don't have cluster role but I have istio-system in my accessible namespaces?

lucasponce:
> do you think it's normal that I cannot get GetAuthorizationDetails and GetMeshPolicies when I don't have cluster role but I have istio-system in my accessible namespaces?

Yes, because those objects are at cluster level not at namespace level.

jmazzitelli:
> I guess the same fix can be applied there (through this PR or the follow-up)

Do we know all the places where this happens? We should just put it all in this PR. And then maybe extract out the common code into a utility function so we aren't repeating this code in multiple places:

```
an := config.Get().Deployment.AccessibleNamespaces
if len(an) == 1 && an[0] == "**" {
  // we have a cluster Role
} else {
  // we do NOT have a cluster Role
}
```

jmazzitelli:
@jotak @lucasponce comment in here if you know of other places that we know require cluster role (and will fail otherwise).

lucasponce:
@jmazzitelli you need to trace GetClusterRbacConfigs() and GetMeshPolicies() calls in TLS and validations.

jmazzitelli:
@jotak try it now. The forbidden messages should only show up on time per k8s API type (so once by GetMeshPolicies, once by GetClusterRbacConfigs, etc).

PR #2022, opened 2019-12-10
[operator] script to generate new OLM manifest files

### LINKED ISSUE #2021: [operator] create a simple script to automate creation of new operator manifest files
right now we have to do some basic steps to create new versions of OLM manifest files. These can be automated, at least for the simple things we need.

jmazzitelli:
fixed #2021 

This just does the grunt work of copying the previous version of the manifest files and editing the parts that correspond to the new version.

If anything extra needs to change per version  (such a new role permissions or new descriptions) then you will obviously still need to hand-edit the new manifest files after generating the new version with this script.

### DIFF
--- a/operator/manifests/create-new-version.sh
+++ b/operator/manifests/create-new-version.sh
@@ -1,0 +1,127 @@
+#!/bin/bash
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
+
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case $key in
+    -mn|--manifest-name)
+      MANIFEST="$2"
+      shift;shift
+      ;;
+    -nv|--new-version)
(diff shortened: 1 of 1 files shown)


jmazzitelli:
i think this is working.

@mattmahoneyrh  since you were asking for something like this, feel free to review and try it out.

Script needs to be in the same directory where the manifests are (like how it is here - it's in the same directory where `kiali-community`, `kiali-upstream`, and `kiali-ossm` manifests are).

Then just run it like so (this creates a new 1.1.0 version of the kiali-ossm manifest):

```
./create-new-version.sh -ov 1.0.8 -nv 1.1.0 -mn kiali-ossm
```

See `--help` to the script to explain what `-ov` , `-nv`, and `-mn` are (heads up - it isn't earth-shattering :-)

PR #2029, opened 2019-12-12
Test CLIENT_EXE_NAME instead of CLIENT_EXE which might not be set

FilipB:
CLIENT_EXE is not required and is auto-detected when not set but it was
tested before it was auto-detected.


jmazzitelli:
hmm.. this is a bug, but the fix is not right because a user could pass in "-c" to change to kubectl - but the check has already been done prior to it.

jmazzitelli:
i'll submit a new PR that fixes this.. but thanks for finding the bug. the original script is definitely broken for kubectl use-cases

jmazzitelli:
This fix won't work if the user specified `-c kubectl` or something like `-cep /usr/bin/kubectl`

See this PR for a fix that supports those use cases too: https://github.com/kiali/kiali/pull/2030

PR #2030, opened 2019-12-12
[hack] only check if we are on OpenShift after the client-exe variable is determined

jmazzitelli:
This fixes the problem that was reported here: https://github.com/kiali/kiali/pull/2029


jmazzitelli:
@FilipB see if this fixes it. Note that this fix will also work when the user specified `-c kubectl` or something like `-cep /usr/bin/kubectl`

Issue #1969, opened 2019-11-22
(workflow) Some automation of issue/pr labeling


israel-hdez:
**Is your feature request related to a problem? Please describe.**
I keep *partially* forgetting to apply labels (it's partially because I usually realize that labels aren't right). 

**Describe the solution you'd like**
Some labeling could be automatic:
* New issues should have `triage` by default (this can be done using the github template)
* New bugs should have `bug` and `backlog` by default (this can be done using the github template)
* When an issue is assigned, the `team/*` label associated with the assignee should be applied.
* When an issue is assigned, the current Sprint project should be applied (or, at least, a warning msg could be generated).
* When a PR is opened, the `team/*` label associated with the opener should be applied.
* When a PR is opened, the assignee of the associated issue should be checked to match.
* When a PR is opened, the current sprint project should be applied.
* When a PR is opened, the `bug`, `enhancement`, `docs` or `maintenance` labels should be inherited from the associated issue.
* When assigning reviewers to a PR, the `team/*` label associated with the reviewer should be applied.

Most label automation would be happening when opening the issue/pr, so it's a one-time action, except for the `team/*` labels. In this case, automatic labeling of a particular `team/*` must become disabled for an issue/PR when a manual adjustment is done.

**Describe alternatives you've considered**
Try to not forget :)

lucasponce:
Please, let's discuss workflow issues in the document we created:
https://docs.google.com/document/d/1ztlgTlUVtSbXioqO9bOFMeZdDInMPtu8iWEkjjitO8E/edit#heading=h.7qq70lsd7kyz

jshaughn:
I agree with Lucas, let's discuss in the doc and reserve the github issue for anything actionable.  I have copied the proposal to the doc and added my comments.

israel-hdez:
Closing as not worthwhile.

PR #2048, opened 2019-12-17
Fix issue templates

israel-hdez:
Lately, looks like no labels are being added to new issues when using the templates. But issue templates have labels that no longer exist because the emoji was removed some time ago.

Probably, this is the reason behind no labels being added when using the templates. This may fix the problem and we may, again, see the labels automatically added.

Related #1969.


israel-hdez:
@jshaughn @lucasponce So, what we agreed on the GDoc about the labels was already in place. But apparently it's currently broken. This should be fixing it :)

PR #2061, opened 2019-12-20
Simplify to use queryParams.Get whenever possible, instead of direct array access

jotak:


jotak:
@jshaughn following your remark, I've removed the from/to new parameters ; so this should be a transparent change, I've only kept some simplification about query-params retrieval - minor change, nothing mandatory.

jotak:
@jshaughn done
Can you also check https://github.com/kiali/k-charted/pull/68 ? This is basically the same

Issue #2066, opened 2020-01-03
[operator] Kubernetes Ingress always assumes HTTPS when it should not


jmazzitelli:
In the Kiali CR there is a "spec.deployment.ingress_enabled" setting that when true (the default) tells the operator to create a Kubernetes Ingress when installing on Kubernetes (non-OpenShift) environment.

When testing on minikube for example, you will see an Ingress created with the nginx annotations set to indicate HTTPS. But if you do not have Kiali configured for HTTPS (that is, the `identity` settings set to empty strings which disables HTTPS) these Ingress nginx annotations need to be set accordingly to turn off the expectation of HTTPS on the backend.

The problem is here:

https://github.com/kiali/kiali/blob/v1.11/operator/roles/kiali-deploy/templates/kubernetes/ingress.yaml#L9-L14

```
  annotations:
    # For ingress-nginx versions older than 0.20.0
    # (see: https://github.com/kubernetes/ingress-nginx/issues/3416#issuecomment-438247948)
    nginx.ingress.kubernetes.io/secure-backends: "true"
    # For ingress-nginx versions 0.20.0 and later
    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
```

We need to do some similar templating as we do in the [service template](https://github.com/kiali/kiali/blob/v1.11/operator/roles/kiali-deploy/templates/kubernetes/service.yaml#L14).

PR #2067, opened 2020-01-03
[operator] create proper annotations on k8s ingress when https is disabled on the backend

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2066

### DIFF
--- a/operator/roles/kiali-deploy/templates/kubernetes/ingress.yaml
+++ b/operator/roles/kiali-deploy/templates/kubernetes/ingress.yaml
@@ -9,9 +9,9 @@ metadata:
   annotations:
     # For ingress-nginx versions older than 0.20.0
     # (see: https://github.com/kubernetes/ingress-nginx/issues/3416#issuecomment-438247948)
-    nginx.ingress.kubernetes.io/secure-backends: "true"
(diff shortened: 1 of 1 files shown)


jmazzitelli:
To test:

First, ensure you install Istio and the Kiali Operator in minikube. Then ...

1. Create a Kiali CR with an `identity` section declaring empty keys (to disable https):

```
identity:
  cert_file: ''
  private_key_file: ''
```

2. See that the Ingress is created with the appropriate nginx annotations to indicate HTTP only:

```
$ kubectl get -n istio-system ingress kiali -o jsonpath='{.metadata.annotations}'
map[nginx.ingress.kubernetes.io/backend-protocol:HTTP nginx.ingress.kubernetes.io/secure-backends:false ...]
```

3. Update the Kiali CR and remove the `identity` section (this enabled https with the default keys)

4. See that the ingress is updated with the nginx annotations changed to reflect the backend is now using HTTPS:

```
$ kubectl get -n istio-system ingress kiali -o jsonpath='{.metadata.annotations}'
map[nginx.ingress.kubernetes.io/backend-protocol:HTTPS nginx.ingress.kubernetes.io/secure-backends:true ...]
```

israel-hdez:
I did a post-merge check and it works.

Issue #2073, opened 2020-01-08
[operator] replace k8s_facts with k8s_info


jmazzitelli:
```
[DEPRECATION WARNING]: The 'k8s_facts' module has been renamed to 'k8s_info'. 
This feature will be removed in version 2.13. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
```

This was something I found here: https://github.com/operator-framework/operator-sdk/issues/2168

Looks like Ansible 2.9 deprecated that k8s_facts test. We use it in our playbooks, so we need to fix this too.

jmazzitelli:
From that operator SDK issue, it sounds like a trivial search-n-replace:

`This is pretty easy to fix; we just need to change from using k8s_facts to k8s_info. I verified with a default test setup that this works identically and drops the deprecation warnings.`

jmazzitelli:
fixed here: https://github.com/kiali/kiali/pull/2076

jmazzitelli:
Ansible Operator SDK moved to Ansible 2.9 in this PR: https://github.com/operator-framework/operator-sdk/pull/2203
It went in 0.13.0: https://github.com/operator-framework/operator-sdk/commit/cf1ca34ce262c45e94504083ff51ef1606a6f212

So we do not need to do this work until the productized Ansible image is at least 0.13.0 (but today it is only at 0.12.). In fact, when last I tried this, I saw problems while running with current productized ansible images. So we should wait until they release a newer base image.

jmazzitelli:
Base image 4.5 is now at 0.16.0 which provides the collection so we can use k8s_info now.

Issue #1584, opened 2019-08-28
Evicted Kiali Operator will not deploy Kiali


lucasponce:
Evicted Kiali Operator pods will be on hanging state until the evicted pods are up.

Kiali operator will have this log on the pod.

```
{"level": "info", "ts": 1566293874.0547807, "logger": "leader", "msg": "Not the leader. Waiting."
```

Logged this issue upstream some time ago: https://github.com/operator-framework/operator-sdk/issues/1305

JIRA:
https://issues.jboss.org/browse/KIALI-3239

abonas:
@jmazzitelli fyi

jmazzitelli:
This is a problem in the underlying operator SDK code - they plan on addressing it in a future release. See comments in https://github.com/operator-framework/operator-sdk/issues/1305

abonas:
@jmazzitelli is there any AI/workaround in Kiali for that? in other words - should this issue remain open on Kiali side?

jmazzitelli:
We should leave this open if for no other reason than we need to remember to keep tracking this issue up on the operator-sdk side of things and test the fix when they have one.

I know of no workaround.

jmazzitelli:
I think this [SDK issue](https://github.com/operator-framework/operator-sdk/issues/1305) is fixed now - see https://github.com/operator-framework/operator-sdk/pull/2210

jshaughn:
The fix is in but not in any release.  We are waiting for a release, probably 0.13, and then we will need to move to that version.

https://github.com/operator-framework/operator-sdk/releases

jmazzitelli:
> The fix is in but not in any release. We are waiting for a release, probably 0.13, and then we will need to move to that version.

To be clear.. the fix is in but not in any release **of operator SDK**. We are waiting for a release, probably 0.13, **of operator SDK**, and then we will need to move to that as the operator's base image.

I asked the operator SDK team when this fix will be available in a release - see:

https://github.com/operator-framework/operator-sdk/pull/2210#issuecomment-564108920

jshaughn:
@jmazzitelli , please proceed with the update to 0.13, which has the fix.

jmazzitelli:
fixed here: https://github.com/kiali/kiali/pull/2076

Issue #2074, opened 2020-01-08
[operator] jmepath is included now, kiali operator doesn't need to add it 


jmazzitelli:
See: https://github.com/operator-framework/operator-sdk/pull/2252

The jmespath library seems to be now included in the ansible base image. The kiali operator adds this because it did not exist in the base image before. See: https://github.com/operator-framework/operator-sdk/pull/2252

So we need to remove that once we bump up the next base image version - which is happening soon via issue https://github.com/kiali/kiali/issues/1584 / PR https://github.com/kiali/kiali/pull/2079

Also, see if we need to change or remove jmespath mentions in other places: https://github.com/kiali/kiali/search?q=jmespath&type=Code

jmazzitelli:
The work to remove this out of the dockerfile is done in PR #2079.
The other mentions of jmespath is still required because those uses are for when running locally and in that case the developer will need jmespath. So I'm going to close this since no additional work is needed.

PR #2076, opened 2020-01-08
[operator] update base ansible image and change to k8s_info task

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/1584
fixes https://github.com/kiali/kiali/issues/2073

Move up to base image of ansible operator SDK 0.13, which also means we need to move from k8s_facts to k8s_info due to deprecation of the former.

### DIFF
--- a/operator/build/Dockerfile
+++ b/operator/build/Dockerfile
@@ -1,10 +1,5 @@
-FROM quay.io/operator-framework/ansible-operator:v0.10.0
+FROM quay.io/operator-framework/ansible-operator:v0.13.0
 
 COPY roles/ ${HOME}/roles/
 COPY playbooks/ ${HOME}/playbooks/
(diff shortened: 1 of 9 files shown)


jmazzitelli:
this also addresses part of https://github.com/kiali/kiali/issues/2074 - since the new base image has jmespath we do not need to add it explicitly

jmazzitelli:
this k8s_info was introduced in ansible 2.7, so this should work for prior versions of the operator base image (including the productized versions of the base image): https://docs.ansible.com/ansible/latest/modules/k8s_info_module.html

so this should be safe to merge

jmazzitelli:
after further testing , i do not have confidence the k8s_info will work on earlier versions. I'm closing this PR and will create another one that does not touch that stuff.

Issue #1957, opened 2019-11-19
Link to console when running on Openshift


cfcosta:
JIRA Issue: https://issues.jboss.org/browse/OSSM-4

Goal:
Provide easy access to the service mesh console to users with access to namespaces in the mesh by creating a link from OpenShift UI for the managed namespaces.

Problem:
The Service Mesh console (Kiali) will be deployed in one namespace, and the application components that are managed by the service mesh will be deployed into N other namespaces. Users administering an application component who want to access the service mesh console will not know in what namespace the Kiali console is running. Information on which namespace the Kiali console is in will need to be communicated out of band unless a link is provided in the managed namespace.

Why is this important:
The value of the console is zero if users cannot locate and access it.

Dependencies (internal and external):
OpenShift console extensibility. At this time we do not believe there is a proper extension point for us to implement.

Prioritized epics + deliverables (in scope / not in scope):
As a developer I want a link to the Service Mesh Console in every project that's part of the mesh so that I do not have to navigate out of my project to get to the console

lucasponce:
Minor comment.
Please, add a link from downstream -> upstream issue in the OSSM-4.
Then Program/Product Managet can browse to this one.

abonas:
> Dependencies (internal and external):
> OpenShift console extensibility. At this time we do not believe there is a proper extension point for us to implement.

@cfcosta  this part is probably redundant in upstream issue?

cfcosta:
@abonas that has been updated since the issue has been created, OCP 4.3 is developing an extension point for this.

lucasponce:
Please, let's progress on this issue. 
It's part of our next release and today it's in the radar with a potential delay.

Let's use this work as base:
https://github.com/openshift/console/pull/3176

And let's prepare a draft with the Maistra/Kiali part, according with the info that is merged in openshift there will be needed a CR for this.

From Maistra/Kiali side we can start prototyping how when a new namespace is added to the mesh that the operator should create a CR representing this link with the project.

It doesn't matter if that CR is the final one needed for OCP 4.3 or we can start working on this with a demo one but I'm sure that we will need to check / resolve issues at Maistra/Kiali side first.

jshaughn:
@cfcosta I'm removing the Waiting on External label.  I think progress can be made.  As far as I know the support has been merged into 4.3 and is available in a nightly build.  Also, on our side we should be able to work on the needed support.  You may want to talk with @jmazzitelli (before the break) regarding any operator work.

PR #2078, opened 2020-01-08
[operator] Add console links to OS 4.3

cfcosta:
Add kiali links to the OpenShift console, one in the Application Launcher for the dashboard, and a contextual link on the project details page, which opens directly to the graph on the namespace.

Closes https://github.com/kiali/kiali/issues/1957.

This is the first part of the PRs needed for this, and can be merged separately. It does however need testing and proper review, as this is my first operator change.

There are some current limitations on the implementation:

* There is no way to set an icon to the app launcher item.
* There is no way to have a placeholder on the link to set the correct namespace, so I needed to create one `ConsoleLink` for each namespace in the cluster.
* I found a bug where logging in with a link for an internal page redirects the user to the dashboard, which breaks behaviour if the user is not signed in when it clicks on the contextual link.

The UI implementation for this (adding the ApplicationLauncher to Kiali-UI) still depends on external input from the Console team.

Screenshot with both links:
image


operator/roles/kiali-deploy/tasks/openshift/os-consolelinks.yml
```diff
@@ -0,0 +1,49 @@
...
+    kiali_route_protocol: "https"
+  when:
+  - is_openshift == True
+  - kiali_route_tls_termination != ""
+
+- name: Create URL for Kiali OpenShift route
+  set_fact:
+    kiali_route_url: "{{ kiali_route_protocol }}://{{ kiali_route_raw['resources'][0]['status']['ingress'][0]['host'] }}"
+  when:
+  - is_openshift == True
```
jmazzitelli: let's extract out everything up to line 43 into a separate yaml file `openshift/os-get-kiali-route-url.yml`. Let it create the kiali_route_url fact - and then we can just use it here and in the osauth.yaml.

operator/roles/kiali-deploy/tasks/openshift/os-main.yml
```diff
@@ -54,3 +54,8 @@
   include: openshift/os-oauth.yml
   when:
   - is_openshift == True
+
+- name: Process Kiali Console Links on OpenShift
+  include: openshift/os-consolelinks.yml
+  when:
+  - is_openshift == True
```
jmazzitelli: since we are refactoring os-consolelinks.yaml and removing the bulk of it into a common yaml to create kiali_route_url, that means only about 5 lines are left. So we can delete the os-consolelinks.yaml entirely and put those 5 lines here (so rather than doing an `include` just do the `k8s.definition` here.

```
- name: Process OpenShift Console Links
  k8s:
    definition: "{{ lookup('template', 'templates/openshift/console-links.yaml') }}"
  when:
  - is_openshift == True
```

operator/roles/kiali-deploy/tasks/openshift/os-main.yml
```diff
@@ -54,3 +54,8 @@
   include: openshift/os-oauth.yml
   when:
   - is_openshift == True
+
```
jmazzitelli: above this "Process Kiali OAuth client on OpenShift" task is where we should now do the include of the yaml that will create the kiali_route_url fact.

```
- name: Get the Kiali Route URL
  include: openshift/os-get-kiali-route-url.yml
  when:
￼  - is_openshift == True
```

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,22 @@
...
+kind: ConsoleLink
+metadata:
+  name: "kiali-namespace-{{ namespace }}-link"
+spec:
+  href: {{ kiali_route_url }}/{{ kiali_vars.server.web_root }}/console/graph/namespaces?namespaces={{ namespace }}
+  location: NamespaceDashboard
+  text: Open on Kiali
+  namespaceDashboard:
+    namespaces:
+      - {{ namespace }}
```
jmazzitelli: indentation is off - all list indentations in the operator yamls are on the same level as the list container (`namespaces` in this context).. .so:

```suggestion
    - {{ namespace }}
```

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: kiali-app-link
```
jmazzitelli: There is no indication what namespace this goes into - I assume it goes into the deployment namespace (e.g. "istio-system") - we need to add that to the metadata.

We also need more metadata here to be consistent with the rest of the resources created by the operator.

```suggestion
  name: kiali-app-link
  namespace: {{ kiali_vars.deployment.namespace }}
  labels:
    app: kiali
    version: {{ kiali_vars.deployment.version_label }}
```

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,22 @@
...
+spec:
+  href: {{ kiali_route_url }}
+  location: ApplicationMenu
+  text: Kiali
+{% for namespace in kiali_vars.deployment.accessible_namespaces %}
+---
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: "kiali-namespace-{{ namespace }}-link"
```
jmazzitelli: Same as above - need more metadata:

```suggestion
  name: "kiali-namespace-{{ namespace }}-link"
  namespace: {{ kiali_vars.deployment.namespace }}
  labels:
    app: kiali
    version: {{ kiali_vars.deployment.version_label }}
```

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
```
jmazzitelli: Because we are creating new types, we need to tell the remove playbook to delete these. This is required when Kiali is to be undeployed. So you need to add ConsoleLink to this list: https://github.com/kiali/kiali/blob/master/operator/roles/kiali-remove/tasks/main.yml#L164

Or, we may have to do something differently because the ConsoleLinks don't have easily hardcoded names (like "kiali" so you will have to delete them based on a query of the metadata labels (which is one reason why adding the metadata labels is needed - see below).

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
cfcosta: I'm not exactly sure if the `ConsoleLink` has a namespace, but will merge those and see what happens.

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
cfcosta: Ok, will give it a try.

operator/roles/kiali-deploy/tasks/openshift/os-get-kiali-route-url.yml
```diff
@@ -1,6 +1,8 @@
-# All of this is ultimately to obtain the kiali_route_url that is needed by oauth.yaml
+# All of this is ultimately to obtain the kiali_route_url that is needed by the console links
+# TODO: Extract this into a better named file
```
jmazzitelli: Trivial - delete this TODO
```suggestion
```

operator/roles/kiali-deploy/tasks/openshift/os-get-kiali-route-url.yml
```diff
@@ -1,6 +1,8 @@
-# All of this is ultimately to obtain the kiali_route_url that is needed by oauth.yaml
+# All of this is ultimately to obtain the kiali_route_url that is needed by the console links
```
jmazzitelli: It's needed for more than just the console links

```suggestion
# All of this is ultimately to obtain the kiali_route_url
```

operator/roles/kiali-deploy/tasks/openshift/os-get-kiali-route-url.yml
```diff
@@ -13,43 +15,35 @@
...
   set_fact:
     kiali_route_url: "{{ kiali_route_protocol }}://{{ kiali_route_raw['resources'][0]['status']['ingress'][0]['host'] }}"
   when:
   - is_openshift == True
-  - kiali_vars.auth.strategy == "openshift"
-
-# Now that we have the kiali_route_url we can process the oauth.yaml
 
-- name: Process OpenShift OAuth client
+- name: Process OpenShift Console Links
```
jmazzitelli: this is in os-main.yml now - all that this os-get-kiali-route-url.yml should do is ultimately set the kiali_route_url fact. So lines 45 and following should be deleted.

cfcosta: Good catch, thanks.

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
cfcosta: Should be ok now.

operator/roles/kiali-deploy/tasks/openshift/os-consolelinks.yml
cfcosta: Done.

operator/deploy/role.yaml
```diff
@@ -223,6 +223,16 @@ rules:
   - routes
   verbs:
   - get
+- apiGroups: ["console.openshift.io"]
```
jmazzitelli: also need to update the golden copies of the manifests (the OLM stuff). See the `openshift/manifests` directory. Need to prepare new versions with this new stuff.

operator/roles/kiali-deploy/tasks/openshift/os-main.yml
```diff
@@ -65,5 +65,9 @@
 - name: Process OpenShift Console Links
   k8s:
     definition: "{{ lookup('template', 'templates/openshift/console-links.yaml') }}"
+  vars:
+    # For some reason, in some cases when the operator runs, the kiali.io/member-of tag is not set yet.
```
jmazzitelli: Ah.. yes. That is because those labels are only ever added when accessible_namespaces is NOT `**` as you see here:

https://github.com/kiali/kiali/blob/v1.12/operator/roles/kiali-deploy/tasks/main.yml#L400-L405

So those are not good to look for since you need something when it is `**`

So I would take out the comment "for some reason in some cases" -- its in all cases when the operator runs with accessible namespaces of ** it does not create those labels because, up to now, those labels were never needed and we didn't want to add a label to every single namespace in the cluster.

operator/roles/kiali-deploy/tasks/openshift/os-main.yml
```diff
@@ -65,5 +65,9 @@
...
+    # Maistra-ones are always present however.
```
jmazzitelli: Of course, this requires you to be running with ServiceMesh installed. Which, again, is why I say if you are going to assume Maistra is installed, then don't worry about any of this since when running with ServiceMesh you will NEVER get accessible_namespaces set to `**`.

operator/roles/kiali-deploy/tasks/openshift/os-main.yml
```diff
@@ -65,5 +65,9 @@
...
+    namespaces: "{{ lookup('k8s', api_version='v1', kind='Namespace', label_selector=('maistra.io/member-of=' + kiali_vars.deployment.namespace)) | default({}) | json_query('[].metadata.name') if '**' in all_accessible_namespaces else all_accessible_namespaces }}"
```
jmazzitelli: This is fine by me - it just means we need to say we only support ConsoleLinks when running with ServiceMesh. But I don't see a problem with this - the only time you will run into a non-supported configuration is if you are on OpenShift, *and* you are using upstream Istio... but that will rarely, if ever, happen since ServiceMesh is shipping with the latest OpenShift releases so virtually everyone will use it and not upstream Istio.

operator/deploy/role.yaml
jmazzitelli: I just noticed that these are permissions for the operator itself not kiali, so we need to move these up above the comment:

`# The permissions below are for Kiali itself; operator needs these so it can escalate when creating Kiali's role`

I'm going to submit a PR to this PR to introduce a molecule test -- I'll submit a change for this as well as changes to the gold copies of the manifests.

jmazzitelli:
I created a PR to this PR to add molecule tests (as well as create the proper golden copies of the manifests). You just need to add some ansible asserts in playbook.yaml (I stubbed those out - you can see where you need to add the asserts).

jmazzitelli:
some assertion code for the molecule tests are in a second PR to this PR..see: https://github.com/cfcosta/swscore/pull/2/commits/1aa6e3fad96870f592a56c9c594cd7735e3efc11

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,30 @@
...
+metadata:
+  name: kiali-app-link
+  namespace: {{ kiali_vars.deployment.namespace }}
```
jmazzitelli: This should not be here because ConsoleLink resources are cluster-scoped.
```suggestion
```

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,30 @@
...
+  href: {{ kiali_route_url }}
+  location: ApplicationMenu
+  text: Kiali
+{% for namespace in namespaces %}
+---
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: "kiali-namespace-{{ namespace }}-link"
+  namespace: {{ kiali_vars.deployment.namespace }}
```
jmazzitelli: This should not be here because ConsoleLink resources are cluster-scoped.

```suggestion
```

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: kiali-app-link
```
jmazzitelli: This should not be hardcoded as a fixed string because now this breaks multi-tenancy scenarios. In other words, suppose I have control plane A and B where kiali is installed in A and B... how can I have two ConsoleLinks pointing to kiali A and B? ConsoleLinks are cluster scoped so there can only be a single link named "kiali-app-link".

We need a way to name these ConsoleLinks specific to the kiali in question.

We can use kiali_vars.deployment.namespace as a unique identifier. So I would make this name have that inside it like this:

```suggestion
  name: "kiali-app-link-{{kiali_vars.deployment.namespace }}"
```

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,30 @@
...
+spec:
+  href: {{ kiali_route_url }}
+  location: ApplicationMenu
+  text: Kiali
+{% for namespace in namespaces %}
+---
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: "kiali-namespace-{{ namespace }}-link"
```
jmazzitelli: Let's get the names of the namespace links consistent with the app link name. As proposed earlier, the app name link will be something like `"kiali-app-link-{{kiali_vars.deployment.namespace}}"` so let's call the namespace links in the same form only instead of `kiali-app` call it `kiali-namespace` so this:

```suggestion
  name: "kiali-namespace-link-{{ namespace }}"
```

If we want to make these links easily associated with the kiali app they belong to, we could put the deployment namespace in here as well, like this:

```
  name: "kiali-namespace-link-{{ kiali_vars.deployment.namespace }}-{{ namespace }}"
```

jmazzitelli:
get this PR merged into this PR to get molecule tests that test basic functionality. I see some failures when namespaces are removed from accessible namespaces: https://github.com/cfcosta/swscore/pull/2

operator/roles/kiali-remove/tasks/main.yml
```diff
@@ -162,6 +162,7 @@
   with_items:
   - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='OAuthClient', resource_name='kiali-' + kiali_vars.deployment.namespace, api_version='oauth.openshift.io/v1') if is_openshift == True else [] }}"
   - "{{ query('k8s', namespace=kiali_vars.deployment.namespace, kind='Route', resource_name='kiali', api_version='route.openshift.io/v1') if is_openshift == True else [] }}"
+  - "{{ query('k8s', kind='ConsoleLink', label_selector='app=kiali')}}"
```
jmazzitelli: I just realized this is wrong too - because in a multi-tenant scenario, if you remove one Kiali CR from one tenant, this is going to delete every consolelink for every tenant. Not good.

jmazzitelli: we might need to add another label to our ConsoleLinks - whatever its name is, the value should be the deployment namespace of Kiali.. so we can add it to the label selector here. something like

`label_selector='app=kiali tenant='+kiali_vars.deployment.namespace`

jmazzitelli:
> get this PR merged into this PR to get molecule tests that test basic functionality. I see some failures when namespaces are removed from accessible namespaces: [cfcosta#2](https://github.com/cfcosta/swscore/pull/2)

I adding the deletion code to that PR - so the ConsoleLinks are now deleted for all namespaces that have been removed from accessible_namespaces.

jmazzitelli:
@cfcosta you should merge PR https://github.com/cfcosta/swscore/pull/2 now - once you do, I believe this PR will be ready for merge, too. We have molecule tests showing things working (well, at least is shows the ConsoleLinks getting created and destroyed properly - it obviously doesn't test that the OS UI console can actually use them).

cfcosta:
Ok, thanks for the help! Will coordinate with @mattmahoneyrh tomorrow to get this merged ASAP.

operator/roles/kiali-deploy/templates/openshift/console-links.yaml
```diff
@@ -0,0 +1,30 @@
...
+apiVersion: console.openshift.io/v1
+kind: ConsoleLink
+metadata:
+  name: kiali-namespace-{{ namespace }}
+  labels:
+    app: kiali
+    version: {{ kiali_vars.deployment.version_label }}
+    kiali.io/home: {{ kiali_vars.deployment.namespace }}
+spec:
+  href: {{ kiali_route_url }}/{{ kiali_vars.server.web_root }}/console/graph/namespaces?namespaces={{ namespace }}
```
jmazzitelli: we should fix this - if kiali_vars.server.web_root is "/" that this will show up as "///"

jmazzitelli:
```suggestion
  href: {{ kiali_route_url }}{{ '/' if kiali_vars.server.web_root == '/' else '/'+kiali_vars.server.web_root }}/console/graph/namespaces?namespaces={{ namespace }}
```

jmazzitelli: this is a separate issue - we'll fix after we merge this PR: https://github.com/kiali/kiali/issues/2132

Issue #2084, opened 2020-01-10
Have a correct default in-cluster URL for Jaeger without Maistra


jotak:
It should be `http://tracing.istio-system/jaeger` instead of `http://tracing.istio-system:16686`

PR #2085, opened 2020-01-10
Have a correct default in-cluster URL for Jaeger without Maistra

jotak:
Fixes https://github.com/kiali/kiali/issues/2084

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -326,10 +326,9 @@ func NewConfig() (c *Config) {
 				Auth: Auth{
 					Type: AuthTypeNone,
 				},
-				Enabled:           false,
+				Enabled:           true,
 				NamespaceSelector: true,
-				InClusterURL:      "https://tracing.istio-system.svc:16686",
-				URL:               "http://tracing.istio-system:16686",
+				InClusterURL:      "http://tracing.istio-system/jaeger",
 			},
 		},
 		IstioLabels: IstioLabels{
--- a/operator/roles/kiali-deploy/tasks/main.yml
+++ b/operator/roles/kiali-deploy/tasks/main.yml
@@ -12,6 +12,7 @@
   set_fact:
     is_openshift: "{{ True if 'route.openshift.io' in api_groups else False }}"
     is_k8s: "{{ False if 'route.openshift.io' in api_groups else True }}"
+    is_maistra: "{{ True if 'maistra.io' in api_groups else False }}"
   when:
   - is_openshift == False
   - is_k8s == False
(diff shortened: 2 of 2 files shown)


jotak:
This seems to work fine, but I'm going to try to be more accurate to differentiate maistra and not just openshift (as @jmazzitelli suggested me)

jotak:
I've done like @jmazzitelli suggested.

Tested without Maistra and half-tested the Maistra scenario (actually I did just add a maistra crd to my minikube cluster and see if it changed the URL .. it does). More tests should be done but I'll switch to openshift cluster later ... anyway it's opened for review.

jotak:
@aljesusg ok but what's the expected behaviour here? I guess this is used only when the operator isn't in place, so this is the istio-upstream scenario, right?

aljesusg:
> @aljesusg ok but what's the expected behaviour here? I guess this is used only when the operator isn't in place, so this is the istio-upstream scenario, right?

Yes

config/config.go
```diff
@@ -326,10 +326,9 @@ func NewConfig() (c *Config) {
 				Auth: Auth{
 					Type: AuthTypeNone,
 				},
-				Enabled:           false,
+				Enabled:           true,
```
jotak: Actually it was disabled by default, so the URLs set below had no effect.
But I think we can enable it by default, set the standard istio in-cluster url, and have no external URL by default (so no external link to jaeger).
Does it sound good @aljesusg  ?

config/config.go
```diff
@@ -326,10 +326,9 @@ func NewConfig() (c *Config) {
...
 				NamespaceSelector: true,
-				InClusterURL:      "https://tracing.istio-system.svc:16686",
-				URL:               "http://tracing.istio-system:16686",
```
jmazzitelli: Is this URL setting getting removed? If so, we need to remove it in the operator in several places:

(the comment and the setting itself): https://github.com/kiali/kiali/blob/v1.11.0/operator/deploy/kiali/kiali_cr.yaml#L385-L401
(the default): https://github.com/kiali/kiali/blob/v1.11.0/operator/roles/kiali-deploy/defaults/main.yml#L122

operator/roles/kiali-deploy/tasks/main.yml
```diff
@@ -12,6 +12,7 @@
   set_fact:
     is_openshift: "{{ True if 'route.openshift.io' in api_groups else False }}"
     is_k8s: "{{ False if 'route.openshift.io' in api_groups else True }}"
+    is_maistra: "{{ True if 'maistra.io' in api_groups else False }}"
```
jmazzitelli: you are going to want to set this separately from is_openshift/is_k8s... look at the "when" clause - this only is executed when is_openshift and is_k8s are both false. This is almost always going to be the case, but not 100% of the time, mainly for testing/development when an operator developer wants to forcibly set one of these to true in the playbook (which I do rarely, but it is done).

So remove this line 15 that was added, and simply add a new task just below... so you want something like this:

```
- name: Determine the Istio implementation
  set_fact:
￼   is_maistra: "{{ True if 'maistra.io' in api_groups else False }}"
```

config/config.go
jmazzitelli: I may be wrong. Looks like the URL setting is still there: https://github.com/kiali/kiali/blob/56e5ef762f5cc9eb8633fafcd63fc630429eeea9/config/config.go#L129

Looks like we just remove the default value here in config.go?

So resolve this conversation if URL remains in the schema.

jotak: Yes, URL still exists but we cannot attribute any default value. I could have written `URL: ""` as well.

jmazzitelli: That's fine - I think we are leaving out any default values of empty string so we don't make that structure bigger than it actually is.

We really only need this long structure of defaults for the unit tests - IIRC, this is never used when actually deployed (in upstream istio or maistra) - because the configmap will be used and since the operator puts every value in the configmap, all of these defaults are eventually overridden anyway (again, IIRC)

hhovsepy:
Regression tested.

Issue #2095, opened 2020-01-14
Purge sensitive data from logs


jotak:
See https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/kiali-users/OkRfZF2b9gA/u1l7zWFlDgAJ

PR #2096, opened 2020-01-14
Purge sensitive config from logs

jotak:
Fixes https://github.com/kiali/kiali/issues/2095

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -91,6 +91,13 @@ type Auth struct {
 	Username           string `yaml:"username"`
 }
 
+func (a *Auth) Obfuscate() {
+	a.Token = "xxx"
+	a.Password = "xxx"
+	a.Username = "xxx"
+	a.CAFile = "xxx"
+}
+
 // PrometheusConfig describes configuration of the Prometheus component
 type PrometheusConfig struct {
 	Auth             Auth   `yaml:"auth,omitempty"`
@@ -156,6 +163,10 @@ type ExternalServices struct {
 type LoginToken struct {
 	ExpirationSeconds int64  `yaml:"expiration_seconds,omitempty"`
 	SigningKey        string `yaml:"signing_key,omitempty"`
+}
+
+func (lt *LoginToken) Obfuscate() {
+	lt.SigningKey = "xxx"
 }
 
 // IstioLabels holds configuration about the labels required by Istio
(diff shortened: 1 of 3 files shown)


jotak:
@jmazzitelli @lucasponce I'm obfuscating not only tokens/password but also some sensitive information such as private key file location, I think it's better.

config/config.go
```diff
@@ -395,7 +406,14 @@ func getDefaultStringFromFile(filename string, defaultValue string) (retVal stri
 
 // String marshals the given Config into a YAML string
```
jmazzitelli: @jotak I haven't looked at the code that calls this String() function - the comment says it returns a YAML string... can you confirm that all the callers to that String() function only use the returned string for logging? i.e. the returned YAML is not used for additional processing of the YAML? I'm thinking maybe some unit tests do that??? I dunno.

Also, can you change the comment to this function -- in the comment, tell people this function is not to be used to process the YAML in any way since some of the data is obfuscated. We need people to know they cannot simply re-parse the returned YAML string and expect to get back to original YAML. The way around that would be to pass in a boolean argument to the String() function -- `obfuscate` -- and when true do what you are doing but if false don't call the new "Obfuscate" functions. I don't think we need this, though, unless there are callers to String() that need to re-parse the YAML for some reason.

jotak: Finding actual callers of `String()` is not easy and changing the signature is not possible, for the same reason: it's actually an implementation of the `Stringer` interface and is used in a wide variety of stuff, like when you do something like `fmt.Sprintf("%v", obj)` the `String()` function of obj would be invoked. In our code we don't explicitly call `config.String()`. Much like java's `Object::toString()`.

But to be honest, this isn't really a problem, we *should* not be using such a formatting function for any business code. I mean, if we really wanted to read/parse as yaml, we would have something like `config.ToYAML()`, not `config.String()`, right? To me it is quite clear than a `String()` implementation is essentially used for logging purpose.

I tried to search in code for config occurrences that could be used in such a way as you describe, but didn't find any. I seriously doubt there's such a thing somewhere.

I think it was transformed into YAML for logging purpose because the native object logging in go was not satisfying for config (it doesn't show keys iirc)

PS: I'll update the function comment

config/config_test.go
```diff
@@ -61,6 +62,35 @@ func TestEnvVarOverrides(t *testing.T) {
...
+		},
+	}
+
+	printed := fmt.Sprintf("%v", conf)
+
+	assert.NotContains(t, printed, "my-username")
+	assert.NotContains(t, printed, "my-password")
+	assert.NotContains(t, printed, "my-token")
+	assert.NotContains(t, printed, "my-signkey")
+	assert.Contains(t, printed, "12345")
```
jmazzitelli: you should probably assert the original values in conf remain as before - to test that you are not overwriting the original values

jotak: good idea

PR #2104, opened 2020-01-15
update makefile to support building and deploying to other non-minikube/non-OS clusters

### LINKED ISSUE #2101: Deploying a custom kiali image via Kubernetes, without OpenShift or Minikube, when following the Kiali readme building steps
Hello, 

I want to be able to deploy a Kiali image I locally built via Kubernetes. 
This does not seem to be possible currently, as apparently only OpenShift and Minikube are supported when following the steps described by the readme on Github.
Building the container image via `make container-build` works fine, but I am unable to use `make cluster-push` and `make operator-create`. 
Both fail at the `.prepare-ocp` step, or with the environment variable `IS_MINIKUBE=true` at `.prepare-minikube`. 

make-cluster-push

I'd like to build and deploy my own Kiali image via Kubernetes/kubectl. It used to be supported in older versions (v1.09 and down).

Downgrading to older versions where building and deploying via kubectl was still supported.

**Additional context**
Using the current master/v1.12. 
kubectl versions are v1.14 client, v1.13 server. 

This probably is a seperate issue, maybe a configuration mistake on my part: 
I've mentioned that I used older Kiali versions. However I've had issues with building it with my locally built UI, as it always used the default kiali UI, even when building with `CONSOLE_VERSION=local` and specifying the UI path via `CONSOLE_LOCAL_DIR`. I've developed my custom UI via proxying, but now I need to implement some backend methods. However, I currently can not build both the custom backend and custom frontend together in any way.

jmazzitelli:
fixes #2101 

If you do not have a minikube or remote openshift cluster, you can build and deploy using the  `CLUSTER_TYPE=local` env var. like this:

`CLUSTER_TYPE=local make container-build operator-create kiali-create`

This assumes the cluster can pull from your local repository.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,11 +64,11 @@ GO_BUILD_ENVVARS = \
 # Determine if we should use Docker OR Podman - value must be one of "docker" or "podman"
 DORP ?= docker
 
-# Set this to true if you want images to be tagged/pushed for minikube as opposed to OpenShift/AWS
-IS_MINIKUBE ?= false
+# Set this to 'minikube' if you want images to be tagged/pushed for minikube as opposed to OpenShift/AWS. Set to 'local' if the image should not be pushed to any remote cluster (requires the cluster to be able to pull from your local image repository).
+CLUSTER_TYPE ?= openshift
 
 # Find the client executable (either oc or kubectl). If minikube, only look for kubectl.
-ifeq ($(IS_MINIKUBE),true)
+ifeq ($(CLUSTER_TYPE),minikube)
 OC ?= $(shell which kubectl 2>/dev/null || echo "MISSING-KUBECTL-FROM-PATH")
 else
 OC ?= $(shell which oc 2>/dev/null || which kubectl 2>/dev/null || echo "MISSING-OC/KUBECTL-FROM-PATH")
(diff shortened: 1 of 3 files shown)


jmazzitelli:
@Exzelsios try this out and see if it works for you.

jmazzitelli:
I confirmed this still works with the openshift (default CLUSTER_TYPE) and minikube use-cases (`CLUSTER_TYPE=minikube`)

jotak:
What happens if the cluster cannot pull from local? I guess we get an ErrImagePull or such, and we need to manually push to the cluster, right?
Maybe a next step could be to automate this, it seems quite similar to what we have with our aws workflow, no?

jmazzitelli:
> What happens if the cluster cannot pull from local? I guess we get an ErrImagePull or such, and we need to manually push to the cluster, right?

Correct.

I am assuming @Exzelsios has a set up that works like it used to work with minishift where you deploy to your local repo and the cluster was able to pull from it. If you do not have this setup, then you have to push to an image repo that is specific to your cluster. This is the magic of `CLUSTER_TYPE=(openshift|minikube)` - the makefiles know how to build the image tag with the appropriate repo location.

Exzelsios:
Yes, this solved my problem. Thank you very much!

jmazzitelli:
@jotak can you approve/merge this PR? Everything looks good to go.

Issue #1528, opened 2019-08-27
Validation of VS converted from Knative channel is red due to weight =0


lucasponce:
In an environment with KNative, a Channel entity was (I'm assuming) converted by Knative into VirtualService Istio entity.
Kiali displays weight =0 as red validation.
it is not clear if weight = 0 is valid (it sounds not per Istio docs)
https://istio.io/docs/reference/config/networking/v1alpha3/virtual-service/#HTTPRouteDestination
so either KNative has some bug in conversion, or if in Istio this is a valid configuration, Kiali should not show an error in validation (remove it or at least reduce to warning)

Related:
https://github.com/istio/istio/issues/15157

JIRA:
https://issues.jboss.org/browse/KIALI-3037

xeviknal:
This validation is not valid anymore. Galley allow users to put routes with 0 weight.

![Screen recording (22)](https://user-images.githubusercontent.com/613814/72520645-26f4a200-385a-11ea-83fb-54cc0c171970.gif)

In the case there is only one route, Istio warns that the weight [will be assumed to be 100](https://istio.io/docs/reference/config/networking/virtual-service/#HTTPRouteDestination).

Opening a PR to remove the scenarios Galley is taking care of and adding the warning above instead.

PR #2108, opened 2020-01-16
VirtualService#Route validation: one route under 100 weight warning

xeviknal:
** Describe the change **
1. Removing all the outdated validations that Galley takes care of.
2. Adding one warning validation for the following scenario:

```yaml
kind: VirtualService
metadata:
  name: reviews
  namespace: bookinfo
spec:
  hosts:
    - reviews
  http:
    - route:
        - destination:
            host: reviews
            subset: v1
          weight: 80
```
This is the scenario where there is only one destinationRouteWeight but there is a weight different to 100. In this case, Istio assumes the weight to be 100. Independently the value set in the weight field.

The validation message is the following:
`The weight is assumed to be 100 because there is only one route destination`

** Issue reference **
Relates to https://github.com/kiali/kiali/issues/1528

** Screeeenshot **
Screenshot of localhost_8000_api_namespaces_bookinfo_istio_virtualservices_reviews_validate=true

### DIFF
--- a/business/checkers/virtual_service_checker_test.go
+++ b/business/checkers/virtual_service_checker_test.go
@@ -74,8 +74,8 @@ func TestVirtualServiceMixedChecker(t *testing.T) {
 	assert.True(ok)
 	assert.Equal(validation.Name, "reviews-mixed")
 	assert.Equal(validation.ObjectType, "virtualservice")
-	assert.False(validation.Valid)
-	assert.Len(validation.Checks, 3)
+	assert.True(validation.Valid)
+	assert.Len(validation.Checks, 2)
 }
 
 func TestVirtualServiceMultipleIstioObjects(t *testing.T) {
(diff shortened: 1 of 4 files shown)


xeviknal:
@skondkar I realized that the validation scope could be improved. I changed that to point only to the weight field inside the route field.

It should be something like this:
Screenshot of Kiali Console - 2020-01-17T121404 373

Thank you for showing me that!

lucasponce:
@xeviknal do we want to backport this to v1.12 branch ?

lucasponce:
LGTM
Then as it's a galley validation, when trying to save a multiple weights, Galley will send the error in the MessageCenter.

xeviknal:
@lucasponce cherry-picked in here: https://github.com/kiali/kiali/pull/2124
Good to go!

PR #2116, opened 2020-01-17
[operator] allow user to override ingress spec and annotation sections to be able to customize it

### LINKED ISSUE #2111: Allow more control over ingress creation

We have a Kubernetes cluster with both the Istio ingress and nginx ingress enabled. The nginx ingress has a specific IP address and routes traffic based on host header along with IP address.

The ingress configuration in the Kiali CR only allows me to enable or disable the ingress; I have no ability to add a `host` entry to the ingress, specify that the ingress should be nginx vs Istio (e.g, `kubernetes.io/ingress.class: nginx`), or provide any other guiding annotations. You also can't specify any `tls` settings to get LetsEncrypt or `cert-manager` into the mix.

Ability to control the ingress (`annotations`, `spec.rules.host`, `spec.tls`, etc.) at a more granular level; or more guidance on writing a custom ingress definition - where to get the service name/port, where to deploy the ingress, what things in the CR YAML correspond to things that show up in the ingress.

I'm currently exporting the existing ingress, setting the CR to disable the ingress, and deploying a custom ingress that's based on what the operator generates but has proper markup for our environment.

jmazzitelli:
fixed #2111 

Because an ingress into a cluster can vary wildly in their desired configuration, a new setting in the Kiali CR is being introduced to allow the user to override complete portions of the ingress resource configuration (Ingress on Kubernetes and Route on OpenShift).

See the new setting in the Kiali CR for an example and the docs commenting the new setting.

For users that use these overrides, they are required to make sure the new configuration is valid - the operator will just blindly copy those settings into the resource - no validation is performed other than the validation the k8s master performs when it applies the new config to the resource.

@tillig can you see if this provides what you need? Based on your comments in the issue, you will want something like this in your Kiali CR:

```
spec:
  deployment:
    override_ingress_yaml:
      metadata:
        annotations:
          kubernetes.io/ingress.class: nginx
      spec:
        tls:
          ...your tls settings...
        rules:
        - host: your.host.here
          http:
            paths:
            - path: /kiali
              backend:
                serviceName: kiali
                servicePort: 20001
```

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -214,7 +214,7 @@ spec:
 #    ---
 #    namespace: ""
 #
-# Because an ingress into a cluster can vary wildly in their desired configuration,
+# Because an ingress into a cluster can vary wildly in its desired configuration,
 # this setting provides a way to override complete portions of the ingress resource
 # configuration (Ingress on Kubernetes and Route on OpenShift). It is up to the user
 # to ensure this override YAML configuration is valid and supports the cluster environment
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -214,6 +214,33 @@ spec:
 #    ---
 #    namespace: ""
 #
+# Because an ingress into a cluster can vary wildly in their desired configuration,
+# this setting provides a way to override complete portions of the ingress resource
+# configuration (Ingress on Kubernetes and Route on OpenShift). It is up to the user
+# to ensure this override YAML configuration is valid and supports the cluster environment
+# since the operator will blindly copy this custom configuration into the resource it
+# creates.
+# This setting is not used if deployment.ingress_enabled is set to 'false'.
+# Note that only 'metadata.annotations' and 'spec' is valid and only they will
+# be used to override those same sections in the created resource. You can define
+# either one or both.
+# Example:
+#   override_ingress_yaml:
+#     metadata:
+#       annotations:
+#         nginx.ingress.kubernetes.io/secure-backends: "true"
+#         nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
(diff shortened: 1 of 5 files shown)


tillig:
@jmazzitelli Yes! That's exactly what I need. 🎉

jmazzitelli:
This is an enhancement and not currently used or passed through via the Maistra CR config, so this isn't technically required to be backported to the 1.12 branch, however, it is an interesting feature that may be a nice-to-have in the Maistra release, so it would be useful to have this in the 1.12 branch - this is why I put the backport label on this PR. If we decide not to backport, we can remove the label.

jmazzitelli:
After discussing with team, we are NOT going to backport to v1.12 branch. This feature will be in the next v1.13 release.

jmazzitelli:
@jshaughn you win the lottery - no one else is reviewing this, so I'm requesting you to review and approve.

operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -214,6 +214,33 @@ spec:
 #    ---
 #    namespace: ""
 #
+# Because an ingress into a cluster can vary wildly in their desired configuration,
```
jshaughn: in its desired configuration

operator/roles/kiali-deploy/defaults/main.yml
```diff
@@ -59,6 +59,7 @@ kiali_defaults:
     image_version: ""
     ingress_enabled: true
     namespace: ""
+    #override_ingress_yaml:
```
jshaughn: is this really an override or a customization?

jmazzitelli: its really an override- you are completely overriding what the operator would have used.

operator/deploy/kiali/kiali_cr.yaml
jmazzitelli: done

PR #2121, opened 2020-01-20
Improve log messages for 3scale and traces

### LINKED ISSUE #2109: Separate errorTraces call from services
In a case that we have problems getting information from a service

Use Case that we found
**/api/threescale/namespaces/bookinfo/services/productpage**

the information returned is 
```json
{"error":"rules.config.istio.io \"threescale-bookinfo-productpage\" not found"}
```

Problems found:

1. We can't see this error in Kiali-Ui messageCenter. So the user can't know what is happening.
2. Should this error block all the information about the service? Can we create somekind of "error_found" in the json and return the information that we found?
3. ErrorTraces Icon is hidden because there is not information about them because kiali return this information in the same call and the user could see error Traces in the chart.

Impact https://github.com/kiali/kiali-ui/pull/1577

cc @xeviknal @lucasponce @jotak @hhovsepy

lucasponce:
Fixes #2109 

It doesn't log an error when a 3scale rule is not found, because it's a valid business request, UI is handled correctly, but in backlog it was printing an unnecessary Error trace.

Also, moved to Trace a repetitive log when fetching spans.

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -327,7 +327,11 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime 
 	go func() {
 		// Maybe a future jaeger business layer
 		defer wg.Done()
-		eTraces, err = in.businessLayer.Jaeger.GetErrorTraces(namespace, service, interval)
+		var err2 error
+		eTraces, err2 = in.businessLayer.Jaeger.GetErrorTraces(namespace, service, interval)
+		if err2 != nil {
+			errChan <- err2
+		}
 	}()
 
 	wg.Wait()
--- a/handlers/threescale.go
+++ b/handlers/threescale.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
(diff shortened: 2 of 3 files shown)


lucasponce:
@hhovsepy not really a fix but just to put better context of these error log traces

aljesusg:
We should add this errors to the MessageCenter in the UI too @lucasponce

lucasponce:
> We should add this errors to the MessageCenter in the UI too @lucasponce

No, as it's a business error, not sure if I understand you.
The threescale error is already added to the MessageCenter when it's a non business condition.
Otherwise is handled correctly.

aljesusg:
> > We should add this errors to the MessageCenter in the UI too @lucasponce
> 
> No, as it's a business error, not sure if I understand you.
> The threescale error is already added to the MessageCenter when it's a non business condition.
> Otherwise is handled correctly.

Mmm I thought that we didn't add it. I didn't see it in the hayk env maybe it was a temporal issue so Frome me side we can merge it

PR #2122, opened 2020-01-20
(cherry pick 2116 to v1.12 branch) allow ingress/route config to be customizable

jmazzitelli:
cherry picks https://github.com/kiali/kiali/pull/2116 to the v1.12 branch.

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -214,6 +214,33 @@ spec:
 #    ---
 #    namespace: ""
 #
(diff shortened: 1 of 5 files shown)


jmazzitelli:
This is an enhancement and not currently used or passed through via the Maistra CR config, so this isn't technically required to be backported to the 1.12 branch, however, it is an interesting feature that may be a nice-to-have in the Maistra release, so it would be useful to have this in the 1.12 branch. If we decide not to backport, we can close this PR without merging.

jmazzitelli:
After discussing with team, we are NOT going to backport to v1.12 branch. This feature will be in the next v1.13 release.

Issue #2136, opened 2020-01-22
[operator] cannot create ConsoleLinks on OpenShift 4.2


jmazzitelli:
The ConsoleLinks that get created use location: NamespaceDashboard:

https://github.com/kiali/kiali/blob/d99f35fd45234d4a7167829196e05ea96a2e848b/operator/roles/kiali-deploy/templates/openshift/console-links.yaml#L25

However, those are not valid on OpenShift 4.2. That location is only available/supported on OpenShift 4.3+.

We need the operator to determine what version of OpenShift it is on and then have it skip creating those ConsoleLinks if 4.2 or earlier.

jmazzitelli:
This is the information I need, but it probably requires special permissions to obtain this  ClusterOperator object (though I'm not sure of that yet).

```
oc get ClusterOperator/kube-apiserver -o jsonpath={.status.versions}
[map[name:raw-internal version:4.2.4] map[name:kube-apiserver version:1.14.6] map[name:operator version:4.2.4]]
```

jmazzitelli:
This gets the version - but I'm not sure if this requires special permissions to get this ClusterOperator object. I also need to confirm this object instance will exist on 4.x clusters (it does on 4.2.4).

```
- name: Determine the OpenShift version
  k8s_facts:
    api_version: config.openshift.io/v1
    kind: ClusterOperator
    name: kube-apiserver
  register: kube_apiserver_cluster_op_raw
  when:
  - is_openshift == True
- set_fact:
    openshift_version: "{{ kube_apiserver_cluster_op_raw.resources[0] | json_query(query) | join }}"
  vars:
    query: "status.versions[?name == 'raw-internal'].version"
  when:
  - kube_apiserver_cluster_op_raw is defined
```

jmazzitelli:
This gets the k8s version, too - in case we need it. I suspect in the future we might need to get the k8s version when not on openshift. but for now, since we don't need it, I don't look up k8s_version if not on openshift:

```
- name: Determine the OpenShift version
  k8s_facts:
    api_version: config.openshift.io/v1
    kind: ClusterOperator
    name: kube-apiserver
  register: kube_apiserver_cluster_op_raw
  when:
  - is_openshift == True
- set_fact:
    openshift_version: "{{ kube_apiserver_cluster_op_raw.resources[0] | json_query(ri_query) | join }}"
    k8s_version: "{{ kube_apiserver_cluster_op_raw.resources[0] | json_query(ka_query) | join }}"
[...]

PR #2138, opened 2020-01-22
[operator] do not create consolelinks if not supported on the openshift version

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2136

Need to not create consolelinks if not supported on the openshift version we are on.


jmazzitelli:
This is ready for review and testing.

Issue #976, opened 2019-04-04
Not seeing edges for kafka events


MathieuGi:
Hello, 

I have some microservices communicating between each other using kafka events but these connections are not shown in kiali graph. Is there a way the take these events into account in kiali ?

thank you very much

abonas:
Can you elaborate a bit more on your system?
kubernetes/openshift?
which version of Istio and Kiali?
Are sidecars injected properly?
Is any other type of communication between microservices showing?
What kind of communication are kafka events? (http/tcp/grpc etc.)
Are those events handled and reported to p8s correctly by Istio and only missing from Kiali UI?
thanks

jmazzitelli:
I believe you need to get Istio to publish the telemetry for those kafka connections. I'm not sure Istio supports that in the telemetry (i.e. if you look in Prometheus, do you see kafka data for those connections in any timeseries?). Without publishing the telemetry, Istio won't be able to tell us anything about the kafka connections.

If there ARE kafka data in Prom, then Kiali would need to implement an enhancement to show this data.

MathieuGi:
@abonas I'm using kubernetes 1.14 with istio 1.0.6 and kiali 0.12.0. Kafka events are using tcp and sidecars seems to be injected properly. Here is what I get in kiali UI
<img width="855" alt="Screen Shot 2019-04-04 at 13 53 55" src="https://user-images.githubusercontent.com/10846231/55553685-28951500-56e1-11e9-8f60-e7ea99fc7295.png">

@jmazzitelli That's interesting, I will take a look at these stuff and come back later if I still don't find anything

MathieuGi:
Apparently prometheus seems to detect kafka events. When I search for : 

> istio_tcp_received_bytes_total{destination_service="kafka-broker"}

 I get 3 elements like this one for example:

> istio_tcp_received_bytes_total{connection_security_policy="unknown",destination_app="unknown",destination_principal="unknown",destination_service="kafka-broker",destination_service_name="kafka-broker",destination_service_namespace="default",destination_version="unknown",destination_workload="unknown",destination_workload_namespace="unknown",instance="172.17.0.10:42422",job="istio-mesh",reporter="source",source_app="conversation-service",source_principal="unknown",source_version="unknown",source_workload="conversation-service",source_workload_namespace="default"}

I tried some other stuffs but i'm still unable to see these kafka events in kiali

jshaughn:
[...]

PR #2144, opened 2020-01-24
Make external_mesh service entry hosts recognized on other namespaces

jotak:
Related to https://github.com/kiali/kiali/issues/976

This is a placeholder PR ; there's other ways to fix the same problem ; needs to be discussed

### DIFF
--- a/graph/telemetry/istio/appender/service_entry.go
+++ b/graph/telemetry/istio/appender/service_entry.go
@@ -105,7 +105,7 @@ func (a ServiceEntryAppender) applyServiceEntries(trafficMap graph.TrafficMap, g
 
 		// A service node represents a serviceEntry when the service name matches serviceEntry host. Map
 		// these "se-service" nodes to the serviceEntries that represent them.
-		if se, ok := a.getServiceEntry(n.Service, globalInfo); ok {
+		if se, ok := a.getServiceEntry(n.Service, n.Namespace, globalInfo); ok {
 			if nodes, ok := seMap[se]; ok {
 				seMap[se] = append(nodes, n)
 			} else {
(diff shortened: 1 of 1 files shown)


graph/telemetry/istio/appender/service_entry.go
```diff
@@ -201,7 +201,7 @@ func (a ServiceEntryAppender) getServiceEntry(serviceName string, globalInfo *gr
 		// handle exact match
 		// note: this also handles wildcard-prefix cases because the destination_service_name set by istio
 		// is the matching host (e.g. *.wikipedia.com), not the rested service (e.g. de.wikipedia.com)
-		if host == serviceName {
+		if host == serviceName || host == serviceName+"."+namespace {
```
jotak: @jshaughn Looking at the whole function, it seems that the behaviour I want for my MESH_EXTERNAL SE is actually the same as what is below defined for MESH_INTERNAL. So maybe we can refactor this a little bit.

jshaughn:
@josejulio You were the service entry expert and I know we have iterated on the SE detection several times.  What do you think of @jotak changes? (if you don't have time for this review please just unassign yourself)

graph/telemetry/istio/appender/service_entry.go
josejulio: I don't recall this very well, how the `namespace` appears in the host?

I though the host was whatever you wrote in the `host` field of the SE, s you could have `foo.bar.whatever` and wasn't related to the namespace.
Is it because `exportTo` is set to an explicit namespace (I think i never tested this part)?

josejulio:
> @josejulio You were the service entry expert and I know we have iterated on the SE detection several times. What do you think of @jotak changes? (if you don't have time for this review please just unassign yourself)

I don't really remember it very well, but checking :)

graph/telemetry/istio/appender/service_entry.go
jotak: Yes indeed, `host` is whatever we write in SE, but it doesn't have to match a service name afaict. 
But wait... I'm not even sure this is right direction. I'm trying to solve a problem for which, in theory, I don't have to create a service entry ... i'm just creating a dummy one as a workaround because else Kiali won't show my metrics... need to step back a little bit :)

jotak:
I'm going to close this one. Not that the problem is solved, but we need to think about it in a more general context of kafka/messaging/stateful-sets compliance effort that is more covered here: https://github.com/kiali/kiali/issues/2197

PR #2152, opened 2020-01-24
[hack] move aws-openshift.sh to OpenShift 4.3.0

jmazzitelli:
OS 4.3.0 is available now. Have the hack script install it by default.

You can still have it use 4.2.4 by passing in `-ov 4.2.4` as a command line argument.

### DIFF
--- a/hack/aws-openshift.sh
+++ b/hack/aws-openshift.sh
@@ -401,7 +401,7 @@ SCRIPT_ROOT="$( cd "$(dirname "$0")" ; pwd -P )"
 cd ${SCRIPT_ROOT}
 
 # The default version of OpenShift to be downloaded
-DEFAULT_OPENSHIFT_DOWNLOAD_VERSION="4.2.4"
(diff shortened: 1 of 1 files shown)


jmazzitelli:
I just tried and got an error - I don't think its related to the 4.3 installer - perhaps its an AWS issue. But for now, leaving this PR in draft state since I cannot be 100% sure this isn't a problem in 4.3.
```
ERROR                                              
ERROR Error: timeout while waiting for state to become 'INSYNC' (last state: 'PENDING', timeout: 30m0s) 
ERROR                                              
ERROR   on ../../../../tmp/openshift-install-487165084/route53/base.tf line 33, in resource "aws_route53_record" "api_external": 
ERROR   33: resource "aws_route53_record" "api_external" { 
ERROR                                              
ERROR                                              
FATAL failed to fetch Cluster: failed to generate asset "Cluster": failed to create cluster: failed to apply using Terraform 
```

jmazzitelli:
This is caused by an outage with AWS Route53. I'll take this out of draft mode. Anyone wants to try and test this out - feel free to approve and merge.

Issue #2155, opened 2020-01-27
Service Details load Error when Jaeger is configured wrongly


hhovsepy:
Happens on master.
Jaeger configured to wrong URI, but the Service Details page load crashes:

![Screenshot from 2020-01-27 10-29-15](https://user-images.githubusercontent.com/604313/73163299-faf5de00-40ef-11ea-88af-bef13fee7c07.png)
![Screenshot from 2020-01-27 10-28-34](https://user-images.githubusercontent.com/604313/73163304-fe896500-40ef-11ea-9211-8156cfe80457.png)

PR #2156, opened 2020-01-27
Service Details load Error when Jaeger is configured wrongly

aljesusg:
Fix https://github.com/kiali/kiali/issues/2155

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -130,7 +130,8 @@ type GrafanaVariablesConfig struct {
 // TracingConfig describes configuration used for tracing links
 type TracingConfig struct {
 	Auth              Auth   `yaml:"auth"`
-	Enabled           bool   `yaml:"enabled"` // Enable autodiscover and Jaeger in Kiali
+	Enabled           bool   `yaml:"enabled"` // Enable Jaeger in Kiali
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -130,7 +130,8 @@ type GrafanaVariablesConfig struct {
 // TracingConfig describes configuration used for tracing links
 type TracingConfig struct {
 	Auth              Auth   `yaml:"auth"`
-	Enabled           bool   `yaml:"enabled"` // Enable autodiscover and Jaeger in Kiali
+	Enabled           bool   `yaml:"enabled"` // Enable Jaeger in Kiali
+	Integration       bool   `yaml:",omitempty"`
```
lucasponce: `yaml:",omitempty"` > "`yaml:"integration,omitempty"` ?

Note that config changes should be propagated to the CRs

aljesusg: This is a prop for use internally with kiali, I think that we shouldn't add to the CRs

jotak: if so maybe that shouldn't be in config, but just some global variable?

aljesusg: I moved it in the last commit. Thanks !!!

aljesusg:
I'll need to make some changes in the UI too.

PR #2169, opened 2020-01-28
(Trivial) avoid too many config.Get()

jotak:
config.Get() involves a lock, so it must not be used when not necessary

@aljesusg @lucasponce this is a small one to avoid unnecessary calls to config.Get()


lucasponce:
+1 and probably it can be touched in more places as well, I couldn't review it locally but worth to look

Issue #2174, opened 2020-01-29
deploy-kiali-operator.sh shows invalid error messages on k8s/minikube install


mtho11:
**General remarks**
k8s installs show 2 error messages that are not relevant to the k8s install (only Openshift)

```
error: the server doesn't have a resource type "routes"
error: the server doesn't have a resource type "oauthclients"
```

k8s/minikube doesn't have routes or oauthclients so these messages should not appear.

Anyway, nobody wants to see error messages when they are installing :)

**Describe the bug**
This bug is related to lines:
https://github.com/kiali/kiali/blob/v1.12/operator/deploy/deploy-kiali-operator.sh#L589-L591

**To Reproduce**
1. Do a dev install:   
`bash <(curl -L https://git.io/getLatestKialiOperator) --accessible-namespaces '**'`
into a minikube/k8s cluster
2. note the log messages

**Expected behavior**
No error messages on install in k8s/minikube.

PR #2175, opened 2020-01-29
don't spit out errors if route or oauthclient types don't exist

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2174


jmazzitelli:
You won't see these errors anymore when deleting/uninstalling kiali on k8s:

```
error: the server doesn't have a resource type "routes"
error: the server doesn't have a resource type "oauthclients"
```

PR #2176, opened 2020-01-29
Set url jaeger default to empty

aljesusg:
Set in the configuration default jaeger Url to empty to avoid that user not set this value.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -340,6 +340,7 @@ func NewConfig() (c *Config) {
 				Enabled:           true,
 				NamespaceSelector: true,
 				InClusterURL:      "http://tracing.istio-system/jaeger",
+				URL:               "",
 			},
 		},
 		IstioLabels: IstioLabels{


lucasponce:
My fault, PR looked good and I merged without test it locally, but it breaks master:

image

hhovsepy:
@lucasponce this was broken before this PR. Happened on upstream Istio 1.4.3 + Kiali master.

jmazzitelli:
If you install via the operator, those config.go defaults are typically not even used (technically, they are overridden) because the operator defines virtually all settings in the config map. The config.go defaults are mainly used during the go unit tests.

So if a default is missing - you usually have to do it in the operator's `roles/kiali-deploy/default/main.yml`

But FYI: this is tracing URL already set to "" by default by the operator - see: https://github.com/kiali/kiali/blob/master/operator/roles/kiali-deploy/defaults/main.yml#L123

I suspect the problem is the upstream Istio's helm chart has the problem.

PR #2184, opened 2020-01-30
[operator] get k8s version and simplify task getting OS version

### LINKED ISSUE #2182: [operator] determine the kubernetes version
The operator should determine the k8s version, at minimum for logging purposes so the user (and support) can know what version of the cluster the operator is on. In the future, the operator may need to know this in order to know if certain features are available to it.

jmazzitelli:
fixed #2182


jmazzitelli:
@israel-hdez yup, output indicates that it is working.

Issue #2186, opened 2020-01-30
[operator] support multiple version streams


jmazzitelli:
Right now, we have a single stream of versions. In OLM, the metadata has a single channel where this stream of versions is found. The OLM channel is called "stable". See the [community metadata here](https://github.com/operator-framework/community-operators/blob/master/community-operators/kiali/kiali.package.yaml#L5). (note: ignore the alpha channel - that is just for future use but it tracks the same version as "stable" and we might remove "alpha" soon since it really isn't used anyway).

The same channel name "stable" is used in the kiali-ossm package (this is the Red Hat certified operator) - as you see [here in the golden copy of the metadata](https://github.com/kiali/kiali/blob/master/operator/manifests/kiali-ossm/kiali-ossm.package.yaml#L3).

However, we can't have just a single stream/channel because the upstream Istio versions that are supported by Kiali differ. And the way OLM works for automatic updates is once a new operator is released in a channel, the previous operator is automatically updated. But this will break certain users. For example, suppose we have a release compatibility chart consisting of the following fictitious versions (this isn't reality - these version numbers are just used for illustration):

Kiali 1.0 through Kiali 1.9 supports Istio 1.0 and 1.1
Kiali 1.10 through Kiali 1.19 supports Istio 1.1 and 1.2

Suppose a user has installed Istio 1.0 and Kiali 1.9. Kiali then releases 1.10. The user SHOULD NOT install this latest Kiali release because Kiali 1.10 does not support Istio 1.0.

But the way OLM and automatic updates work is, as soon as 1.10 is released in the "stable" channel, the user will get upgraded to 1.10 automatically, thus breaking the user due to the incompatibility.

[...]

PR #2190, opened 2020-01-30
[operator] change kiali-ossm manifest so it has 2 channels

jmazzitelli:
This attempts to address https://github.com/kiali/kiali/issues/2186 for the kiali-ossm product release.
This does not address the upstream community OLM metadata, which is a bit more complicated should we wish to have multiple channels for that. However, we have decided for community releases we will only have a single "stable" channel stream of versions.

UPDATE: this PR is likely obsolete and probably will be closed without merging in the near future.


jmazzitelli:
I want to point out that the old "stable" channel is going away in this PR.
The new channels are tentatively named:
* OpenShift Service Mesh 1.0
* OpenShift Service Mesh 1.1

I thought naming them "1.0" and "1.1" was not good because then it isn't clear what those versions are. People will think they are Kiali versions but they are not. e.g. The 1.1 stream does not have Kiali 1.1.x in them. These versions are the OS SM versions.

jmazzitelli:
we are not going to implement this with multiple channels... not merging, closing this PR

Issue #2187, opened 2020-01-30
Avoid fetch spans when integration is disable


aljesusg:
If there is not integration with jaeger we should avoid the fetch spans and remove the errorMessage

 * We can control this in the UI or in the Backend like we did with fetchErrorTraces

jotak:
So I talked a little bit with Xavi about that.
I've always been a bit reluctant with this "integration" flag because it means Kiali is turning off jaeger stuff definitively (until next restart) while jaeger could just be unavailable for some temporary reason. To me this is not the right way to proceed.

If people don't want to have kiali connected to jaeger, then they should turn it off via configuration (there's an "enabled" flag iirc). But having kiali deciding on behalf of the user is, IMO, not the right thing to do.

However I agree the situation with alerts constantly displayed is annoying. Afaik we have some mechanism in Alert center to avoid repeating alerts ... I should try to find out how to use it here.

aljesusg:
No jotak, this info is checked Everytime UI fetch /API/Jaeger (every F5 , reload site, login UI). 

On the other hand I agree that user should be set this to disable if he doesn't want Jaeger.

Cc @lucasponce @xeviknal

lucasponce:
To me if jaeger integration is set disabled in configuration it shouldn't fail.
If jaeger integration is enabled but it fails, then having the errors is the expected behaviour.

I understand that the context of this issue is that jaeger was disabled, but spans were queried.

xeviknal:
My environment was messy. I had tracing `enabled: true` but jaeger wasn't properly working. Then we were running the scenario where tracing was enabled but jaeger wasn't available.
So I think I am quite aligned with Joel comment. If we could make this error not be popping all the time it should be better UX.

jotak:
@lucasponce the trick is that there's two "enabling" flags: one coming from config, and one determined at runtime. The latter is set when Jaeger is enabled (from config), Kiali checking if it can be reached or not.

Maybe we could remove the second flag and have a smarter error management to avoid constantly poping errors. And keep only the flag from config as the only source of truth to know whether we want jaeger integration or not.

jotak:
btw I'm going to test what happens when we disable jaeger from config, it's not something I remember to have tested

lucasponce:
> Maybe we could remove the second flag and have a smarter error management to avoid constantly poping errors. And keep only the flag from config as the only source of truth to know whether we want jaeger integration or not.

+1

aljesusg:
[...]

PR #2192, opened 2020-01-31
Reworked some jaeger error handling and enabling workflow

jotak:
- "Integration" flag removed in favour or just the "enabled" user config. Kiali assumes Jaeger integration is always turned on as long as the user set it to "enabled" in config (which is the default)
- Because of that, some prior checks are not necessary anymore - so removing the "GetJaegerInfo" functions in business/client layers as it now just maps jaeger config to some info exposed for UI. The idea is that Kiali keeps trying to reach Jaeger when it's configured as enabled. Some mechanisms can be setup in the UI to avoid repeating popups.
- Also remove GetJaegerServices that is unused
- Fix issues with returned http code from client calls: some errors were silently ignored because the code was "0". Basically an http code resulting from client queries should not be reused as-is in Kiali UI-backend API. It must be caught in backend and eventually reinterpreted. Here, it's added to the error message.

UI PR: https://github.com/kiali/kiali-ui/pull/1645

Fixes https://github.com/kiali/kiali/issues/2187 and, I think, all the various points discussed in that ticket


aljesusg:
So are we going to remove the "check integration" ? So the point is if user set this to enable so throw one time the error (in metrics/spans and in traces tab). If user disable the configuration but he has the url so kialis shows the links. Am I right @jotak ?

My concern is about the "traces tab" and the behaviour I'll comment in the PR. 

For example if there is no integration (we check this in errTraces errChan <- err2) the traces tab should be disable or a link to the external tracing I think. And again we have the problem that if this fail the serviceDetails fetch fail so I think taht we should have this fetch in another endpoint

jotak:
> For example if there is no integration (we check this in errTraces errChan <- err2) the traces tab should be disable or a link to the external tracing I think.

Yes, I think this is still working as you say, no?

> And again we have the problem that if this fail the serviceDetails fetch fail so I think taht we should have this fetch in another endpoint

Either different endpoints or have the ServiceDetails endpoint able to not fail in this case (but still report an issue) .. i thought actually it was already done as we had the discussion previously, but I can do that

aljesusg:
> > For example if there is no integration (we check this in errTraces errChan <- err2) the traces tab should be disable or a link to the external tracing I think.
> 
> Yes, I think this is still working as you say, no?
mmm I am not sure I need to check the UI PR that you sent, we can discuss this in the UI PR because is more related with the behaviour of the UI.
> 
> > And again we have the problem that if this fail the serviceDetails fetch fail so I think taht we should have this fetch in another endpoint
> 
> Either different endpoints or have the ServiceDetails endpoint able to not fail in this case (but still report an issue) .. i thought actually it was already done as we had the discussion previously, but I can do that

yes...this is why I requested a bj for talk about this . I agree about have another endpoint for this. Do you agree @lucasponce ?

aljesusg:
Tested in minikube, with the UI I only see one "error message" but the traces tab is active and enable if we get some error, I think that we should merge this because fix some problems that we have but we should talk about what behaviour we want to have for this tab. What do you think @lucasponce @jotak ? 
Screenshot from 2020-02-03 12-49-29

PR #2193, opened 2020-01-31
Return ObjectCount and validations only of current ns

xeviknal:
Backend part of this one: https://github.com/kiali/kiali-ui/pull/1646

### DIFF
--- a/models/istio_validation.go
+++ b/models/istio_validation.go
@@ -20,6 +20,10 @@ type IstioValidationSummary struct {
 	// required: true
 	// example: 2
 	Errors int `json:"errors"`
+	// Number of Istio Objects analyzed
+	// required: true
+	// example: 6
+	ObjectCount int `json:"objectCount"`
 	// Number of validations with warning severity
 	// required: true
 	// example: 4
@@ -314,10 +318,12 @@ func (iv IstioValidations) MergeReferences(validations IstioValidations) IstioVa
 	return iv
 }
 
-func (iv IstioValidations) SummarizeValidation() IstioValidationSummary {
+func (iv IstioValidations) SummarizeValidation(ns string) IstioValidationSummary {
 	ivs := IstioValidationSummary{}
-	for _, v := range iv {
-		ivs.mergeSummaries(v.Checks)
+	for k, v := range iv {
+		if k.Namespace == ns {
+			ivs.mergeSummaries(v.Checks)
+		}
 	}
 	return ivs
 }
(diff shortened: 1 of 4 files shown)


xeviknal:
@jshaughn it passes the swagger check now. Thank you for checking this out :)

models/istio_validation_test.go
```diff
@@ -63,7 +63,7 @@ func TestSummarizeValidations(t *testing.T) {
 		},
 	}
 
-	summary := validations.SummarizeValidation()
+	summary := validations.SummarizeValidation("bookinfo")
 
 	assert.Equal(2, summary.Warnings)
 	assert.Equal(2, summary.Errors)
```
jotak: would be nice to add an assertion on ObjectCount too

models/istio_validation.go
```diff
@@ -314,10 +318,12 @@ func (iv IstioValidations) MergeReferences(validations IstioValidations) IstioVa
 	return iv
 }
 
-func (iv IstioValidations) SummarizeValidation() IstioValidationSummary {
+func (iv IstioValidations) SummarizeValidation(ns string) IstioValidationSummary {
 	ivs := IstioValidationSummary{}
```
jotak: maybe here there should be an explicit initialization? (see my comment on frontend PR)

xeviknal: Default are 0 in this case. Then, the empty object looks like this:
```json
{
"errors": 0,
"objectCount": 0,
"warnings": 0
}
```
Does it make sense not to add any initial data?

jotak: yep, that's fine ; tbh I'm sometimes puzzled with go handling of zero-value objects and i wasn't sure if that would come as 0 or undefined on serialization. It's fine to keep as you did.

models/istio_validation.go
```diff
@@ -329,6 +335,7 @@ func (summary *IstioValidationSummary) mergeSummaries(cs []*IstioCheck) {
 		} else if c.Severity == WarningSeverity {
 			summary.Warnings = summary.Warnings + 1
 		}
+		summary.ObjectCount = summary.ObjectCount + 1
```
jshaughn: I see this is consistent with the above, and it's subjective.  But for me either var++ or  var += 1 is a more clear way to increment.  Certainly still fine as is.

xeviknal: Yep, it is an old habit from ruby times. Ruby core decided not to add this feature. Don't remember the rationale behind. Changed :beer:

Issue #1515, opened 2019-08-27
Add validation for repeating subset in VS


lucasponce:
Need to add a validation for repeating subset in VS
Today if user specifies the same subset more than once in VS, for example with different weights, the config is green but the mesh goes crazy 
It turns red, etc which is expected given contradictory configuration.
Attached example.

JIRA:
https://issues.jboss.org/browse/KIALI-2905

lucasponce:
I think this is outdated for Istio 1.4 as Galley take care of it.
Please @xeviknal feel free to reopen if you think I missed something.

xeviknal:
Galley doesn't do the checking. However, the mesh doesn't get broken.
Reopening it again. Adding this validation as warning should be the right approach.

The following VS generates the graph right after:
```yaml
kind: VirtualService
apiVersion: networking.istio.io/v1alpha3
metadata:
  name: reviews
  namespace: default
spec:
  hosts:
    - reviews
  http:
    - route:
        - destination:
            host: reviews
            subset: v1
          weight: 80
        - destination:
            host: reviews
            subset: v1
          weight: 20
```

![Screenshot of Kiali Console - 2020-01-30T152711 035](https://user-images.githubusercontent.com/613814/73458176-2a694c80-4375-11ea-828b-5bce0aade62d.jpg)

PR #2206, opened 2020-02-04
Adding new validation: same subset in different route destination

xeviknal:
** Describe the change **
Adding new validation for VirtualServices. When there are two destination routes referencing to the same subset. 

Screenshot of Kiali Console (3)

** Issue reference **
Fixes https://github.com/kiali/kiali/issues/1515


xeviknal:
There is only pending the documentation part. Coming tomorrow morning.

hhovsepy:
@xeviknal  It shows warning only on second subset, but on your initial screenshot on all:
Screenshot from 2020-02-05 10-14-10

xeviknal:
@hhovsepy you are right. Commit not pushed. it should work now. sorry.

PR #2209, opened 2020-02-05
[hack] Get the latest released version of istio instead of last release

FilipB:
Original solution was using
https://api.github.com/repos/istio/istio/releases/latest which is not
always pointing to the latest released version of istio but to last
release which can be older stream release.
Right now it's pointing to 1.3.7 but the latest version is 1.4.3.
New solution (taken from https://git.io/getLatestIstio) should always
pick the latest released version (now 1.4.3).

### DIFF
--- a/hack/istio/download-istio.sh
+++ b/hack/istio/download-istio.sh
@@ -56,10 +56,12 @@ OUTPUT_DIR="$(pwd)" # remove the .. references
 echo "Output Directory: ${OUTPUT_DIR}"
 
 if [ -z "${ISTIO_VERSION}" ]; then
-   VERSION_WE_WANT=$(curl https://api.github.com/repos/istio/istio/releases/latest 2> /dev/null |\
-         grep  "tag_name" | \
-         sed -e 's/.*://' -e 's/ *"//' -e 's/",//')
+   # get the latest released version
+   VERSION_WE_WANT=$(curl -L -s https://api.github.com/repos/istio/istio/releases | \
+         grep tag_name | sed "s/ *\"tag_name\": *\"\\(.*\\)\",*/\\1/" | \
+         grep -v -E "(alpha|beta|rc)\.[0-9]$" | sort -t"." -k 1,1 -k 2,2 -k 3,3 -k 4,4 | tail -n 1)
    echo "Will use the latest Istio version: $VERSION_WE_WANT"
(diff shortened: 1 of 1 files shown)


hack/istio/download-istio.sh
```diff
@@ -56,10 +56,12 @@ OUTPUT_DIR="$(pwd)" # remove the .. references
 echo "Output Directory: ${OUTPUT_DIR}"
 
 if [ -z "${ISTIO_VERSION}" ]; then
-   VERSION_WE_WANT=$(curl https://api.github.com/repos/istio/istio/releases/latest 2> /dev/null |\
-         grep  "tag_name" | \
-         sed -e 's/.*://' -e 's/ *"//' -e 's/",//')
+   # get the latest released version
+   VERSION_WE_WANT=$(curl -L -s https://api.github.com/repos/istio/istio/releases | \
+         grep tag_name | sed "s/ *\"tag_name\": *\"\\(.*\\)\",*/\\1/" | \
+         grep -v -E "(alpha|beta|rc)\.[0-9]$" | sort -t"." -k 1,1 -k 2,2 -k 3,3 -k 4,4 | tail -n 1)
```
jmazzitelli: This will support future versions as well as current, correct? In other words, when  Istio 1.5 and 1.6 , etc. are released, we don't have to update this code,

Did you look to see how Kiali does it for its own versions? We ran into this same issue in getLatestKialiOperator, and this is how we get the lastest release (not necessarily the last):

https://github.com/kiali/kiali/blob/master/operator/deploy/deploy-kiali-operator.sh#L550-L556

```
resolve_latest_kiali_release() {
  get_downloader
  github_api_url="https://api.github.com/repos/kiali/kiali/releases"
  kiali_version_we_want=$(${downloader} ${github_api_url} 2> /dev/null |\
    grep  "tag_name" | \
    sed -e 's/.*://' -e 's/ *"//' -e 's/",//' | \
    grep -v "snapshot" | \
    sort -t "." -k 1.2g,1 -k 2g,2 -k 3g | \
    tail -n 1)
  if [ -z "${kiali_version_we_want}" ]; then
    echo "ERROR: Failed to determine latest Kiali release."
    echo "Make sure this URL is accessible and returning valid results:"
    echo ${github_api_url}
    exit 1
  fi
}
```

PR #2211, opened 2020-02-05
Support GRPC status codes

### LINKED ISSUE #1235: GRPC errors, response_flags non-empty, graph looks green
**Describe the bug**

Hi.  I have a deployment of Istio 1.2.2, and i've bumped Kiali to tag v0.21.  I've deployed a service which is supposed to serve GRPC on a particular port, but it doesn't.  I'd expect to see the edge or the node in Kiali go red, but instead it stays green.  In the sidebar it correctly reports Envoy's `response_flags`, but flags indicate errors and (IMHO) should be visualised as such.

**Versions used**
Kiali: v0.21
Istio: 1.2.2
Kubernetes flavour and version: Amazon EKS 1.13

**To Reproduce**

Deploy the following services and such:

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: hello-world-service
  namespace: default
  labels:
    app: hello-world-service
    env: development
spec:
  replicas: 1
  selector:
    matchLabels:
      app: hello-world-service
      env: development
  template:
    metadata:
      labels:
        app: hello-world-service
        version: latest
        env: development
    spec:
      containers:
      - name: hello-world-service
        image: debian
        command: ["bash", "-c", "while true; do echo hullo; sleep 30; done"]
        env:
          - name: PORT
            value: "50051"
        ports:
        - containerPort: 50051
        readinessProbe:
          exec:
            command:
            - "true"
          initialDelaySeconds: 5
          periodSeconds: 5
        livenessProbe:
          exec:
            command:
            - "true"
          initialDelaySeconds: 5
          periodSeconds: 5
---
apiVersion: v1
kind: Service
metadata:
  name: hello-world-service
  namespace: default
spec:
  ports:
    - name: grpc-hello-world
      port: 50051
      targetPort: 50051
  selector:
    app: hello-world-service
    env: development
---
kind: Gateway
metadata:
  name: my-gateway-hello-world
  namespace: default
spec:
  selector:
    istio: ingressgateway # use istio default controller
  servers:
  - hosts:
    - "hello-world.scratchpro.link"
    port:
      name: grpc-50051-hello-world-service-platform-dev # this must be unique
      number: 50051
      protocol: grpc # https://github.com/istio/istio/issues/7843#issuecomment-413622721
    # tls:
    #   credentialName: cert-hello-world.scratchpro.link
    #   mode: SIMPLE
    #   privateKey: sds # this value can probably be omitted.
    #   serverCertificate: sds # this value can probably be omitted.
---
kind: VirtualService
metadata:
  name: hello-world-service-dev
  namespace: default
spec:
  hosts:
  - "hello-world.scratchpro.link"
  gateways:
  - my-gateway-hello-world
  http:
  - route:
    - destination:
        host: hello-world-service
        port:
          number: 50051
```

Go to Kiali Graph > workload graph > find the hello-world-service.  I expect that grpc edge to go red, indicating the upstream failures.  Instead, it stays green.  In the sidebar, response codes, you see the URX flag.  See https://github.com/istio/istio/issues/15400 for more context on my issue.

**Expected behavior**

I expect the edge to go red, indicating that upstream is failing.

jshaughn:
In Istio 1.5 (likely) and SM 1.1 (definite) GRPC status codes will be reported for requests using protocol="grpc".  We had partially coded for this in the graph but the code assumed that the GRPC code would be set in the existing response_code telemetry field.  Because Istio was afraid that existing consuming code may assume response_code contains only HTTP codes, it elected to add a new field, grpc_response_status.  See https://github.com/istio/istio/pull/20036 for more.

Although this protects existing consumer code it also makes it more difficult to add support as consumers now need to deal with two, protocol-specific, fields.

In this PR the graph code deals with it by basically setting our existing response code field with the appropriate protocol-specific code in the telemetry.  In that way most code stays the same.

In metrics things are little complicated because we support queries that fetch all error time-series.  That results in the need for OR queries, or a post-query merge of data.  Merge is hard so this PR opts for the OR logic, which does produce some longer/more-complicated query strings.  In Health we aggregate errors and so the code now uses a regex pattern to capture both HTTP and GRPC error codes.

Fixes #1235

### DIFF
--- a/prometheus/metrics.go
+++ b/prometheus/metrics.go
@@ -20,7 +20,7 @@ func getMetrics(api prom_v1.API, q *IstioMetricsQuery) Metrics {
 	return metrics
 }
 
-func buildLabelStrings(q *IstioMetricsQuery) (string, string) {
+func buildLabelStrings(q *IstioMetricsQuery) (string, []string) {
 	labels := []string{fmt.Sprintf(`reporter="%s"`, q.Reporter)}
 	ref := "destination"
 	if q.Direction == "outbound" {
@@ -48,15 +48,23 @@ func buildLabelStrings(q *IstioMetricsQuery) (string, string) {
 
 	full := "{" + strings.Join(labels, ",") + "}"
 
-	labels = append(labels, `response_code=~"[5|4].*"`)
-	errors := "{" + strings.Join(labels, ",") + "}"
+	errors := []string{}
+	protocol := strings.ToLower(q.RequestProtocol)
+	if protocol == "" || protocol == "grpc" {
+		grpcLabels := append(labels, `grpc_response_status!="0"`)
+		errors = append(errors, ("{" + strings.Join(grpcLabels, ",") + "}"))
+	}
+	if protocol == "" || protocol == "http" {
+		httpLabels := append(labels, `response_code=~"[5|4].*"`)
+		errors = append(errors, "{"+strings.Join(httpLabels, ",")+"}")
+	}
 
 	return full, errors
 }
 
-func fetchAllMetrics(api prom_v1.API, q *IstioMetricsQuery, labels, labelsError, grouping string) Metrics {
+func fetchAllMetrics(api prom_v1.API, q *IstioMetricsQuery, labels string, labelsError []string, grouping string) Metrics {
 	var wg sync.WaitGroup
-	fetchRate := func(p8sFamilyName string, metric **Metric, lbl string) {
+	fetchRate := func(p8sFamilyName string, metric **Metric, lbl []string) {
 		defer wg.Done()
 		m := fetchRateRange(api, p8sFamilyName, lbl, grouping, &q.BaseMetricsQuery)
 		*metric = m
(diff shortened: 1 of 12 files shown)


prometheus/metrics.go
```diff
@@ -122,15 +130,24 @@ func fetchAllMetrics(api prom_v1.API, q *IstioMetricsQuery, labels, labelsError,
...
+			query += fmt.Sprintf("sum(%s(%s%s[%s]))", q.RateFunc, metricName, labelsInstance, q.RateInterval)
+		} else {
+			query += fmt.Sprintf("sum(%s(%s%s[%s])) by (%s)", q.RateFunc, metricName, labelsInstance, q.RateInterval, grouping)
+		}
+	}
+	if len(labels) > 1 {
+		query = fmt.Sprintf("(%s)", query)
 	}
 	query = roundSignificant(query, 0.001)
+	fmt.Println("QUERY:\n" + query)
```
jotak: this ^^^
Or we could as well keep this log but it should be done with the logger then

jshaughn: yes, I had this here for dumping to the console while trying to update the tests.  I will remove it. 
I don't think we want to dump queries to the log either, it's too noisy, even in debug.  If you want to see the queries being generated run the prom clien tests with these Printlns.

jshaughn:
> When looking at frontend code, the two of them (with and without errors) are always fetched together. So we could get them both within a single promql call because they actually relate to the same metric. I'm sure we could refactor that to make it simpler. We could just fetch "request_count" grouping on errors, and leave further processing to the frontend responsibility. Of course it's going to add a little bit of work on the frontend, but I think what we'll gain on one side exceed what we'll lose on the other side.

@jotak  yes, I had made these changes completely trying to keep API parity, and ignoring the client.  But because our API is still not supported and only [officially] serves our UI, I agree that it could be a good refactor.    Another approach, even in this PR, would be for request_error_count to  "fetch 'request_count' grouping on errors",  and then filter the results server-side as opposed to client-side.  It is probably less efficient than letting Prometheus do it but would avoid the giant OR queries.

jshaughn:
I found a problem in the response code appender work.  Added DNM until i resolve the problem.

jshaughn:
Fixed issue in response time, removing DNM label.

jshaughn:
@mattmahoneyrh, @pbajjuri20  Initial testing for this work should basically be regression testing around the graph, response time of graph edges, the metric charts, and health.  In all cases it would be good to have a combination of success and error traffic.  GRPC traffic would be a bonus if you can generate GRPC traffic.  I had to fake it locally.

@gbaufake When SM supports the GRPC  status code you can further verify the change using Kiali, once this is in.

mattmahoneyrh:
When deploying this PR with Istio 1.3, it can be seen that GRPC traffic is showing 100% error on the spark-line graph, even though the GRPC edge is 'green' (200 return codes)

Selection_127

jshaughn:
@mattmahoneyrh  I finally found the *bad* problem and the PR has been updated with the fix.

pbajjuri20:
@mattmahoneyrh  I have deployed and tested the PR. The spark-line graph issue has been fixed.
@jshaughn  Thanks for fixing the Issue.
IstioGraph

pbajjuri20:
@mattmahoneyrh I have deployed and tested the PR. The spark-line graph issue has been fixed.
@jshaughn Thanks for fixing the Issue.


IstioGraph

jshaughn:
@israel-hdez @mattmahoneyrh @pbajjuri20 I think this is good to go, just need dev re-approvals and QE approval.  Thanks.

prometheus/metrics.go
```diff
@@ -51,11 +51,14 @@ func buildLabelStrings(q *IstioMetricsQuery) (string, []string) {
 	errors := []string{}
 	protocol := strings.ToLower(q.RequestProtocol)
 	if protocol == "" || protocol == "grpc" {
-		grpcLabels := append(labels, `grpc_response_status!="0"`)
+		// this is intentionally not `grpc_response_status!="0"`. We need to be backward compatible and
+		// handle the case where grpc_response_status does not exist.  In Prometheus, negative tests on a
+		// non-existent label match everything, but positive tests match nothing. So, we stay positive.
+		grpcLabels := append(labels, `grpc_response_status=~"^[1-9]$|^1[0-6]$"`)
```
israel-hdez: Interesting find.

PR #1773, opened 2019-10-08
[operator] support operator upgrades

jmazzitelli:
fixes #1772

PR #2219, opened 2020-02-06
[operator] support multiple versions of kiali within a single operator

jmazzitelli:
Another attempt at addressing #2186 

This attempt changes the design of the operator such that the one operator implementation can install multiple versions of Kiali. This also keeps the original OLM channel "stable" (there will not be multiple channels - we can just have a single stream as long as the operator is able to install previous versions, which it can right now).

For each breaking, incompatible release, we would need to duplicate the ansible roles and add some additional metadata.  Then the user can tell the operator which specific version to install (this PR supports "default" (which will install the Kiali whose version is the same as the operator itself), "v1.0" which uses the v1.0 tag and the ansible roles from the v1.0 branch, and also similarly "v1.12".). You specify the version in the Kiali CR's new "version" field.

If backward compatible releases follow (for example, support v.1.12.1 or v1.13) then we just bump up the supported version in the default-supported-versions.yaml but we can keep everything as-is.


FilipB:
@jmazzitelli Would it be possible to add more info to kiali_cr.yaml to spec.version? e.g. what are the options, how one knows which option to choose based on used istio version. Also a relation with spec.deployment.image_name - e.g. what happens if I set image_name to some kiali version which is not compatible with option specified in spec.version?

I don't see any changes of deploy-kiali-operator.sh in this PR - what is the plan with deploy-kiali-operator.sh when this PR is merged?

jmazzitelli:
> Would it be possible to add more info to kiali_cr.yaml to spec.version? e.g. what are the options, how one knows which option to choose based on used istio version. Also a relation with spec.deployment.image_name - e.g. what happens if I set image_name to some kiali version which is not compatible with option specified in spec.version?

I added some more details to that file to explain further.

> I don't see any changes of deploy-kiali-operator.sh in this PR - what is the plan with deploy-kiali-operator.sh when this PR is merged?

There are no changes to that file that are required. Installing with deploy-kiali-operator.sh leaves this version unset and thus (as per the comments in kiali_cr.yaml) the "default" version is installed - which means the Kiali version that is the same version as the operator will be installed (which is the current behavior of the script, so no backward incompatible change is occurring here).

jmazzitelli:
Note that it is possible for different OLM manifests to define different environment variables for the operator Deployment - we use this feature to support overriding the default deployment.image_[name,version] per "version" setting. See `KIALI_IMAGE_default`, `KIALI_IMAGE_v1_0` and `KIALI_IMAGE_v1_12` environment variables that are set in this PR's kiali-ossm manifest for v0.12 (operator/manifests/kiali-ossm/1.12.0/kiali.v1.12.0.clusterserviceversion.yaml). Users will never care about this particular override feature (it is only something the operator itself takes advantage of) so it isn't described in detail in kiali_cr.yaml. All the user will care about is that the "version" setting will by default install a correct version of Kiali image. We (Kiali operator devs) make sure the operator does this.

FilipB:
> > Would it be possible to add more info to kiali_cr.yaml to spec.version? e.g. what are the options, how one knows which option to choose based on used istio version. Also a relation with spec.deployment.image_name - e.g. what happens if I set image_name to some kiali version which is not compatible with option specified in spec.version?
> 
> @FilipB I added some more details to that file to explain further.
Thank you. It's clear now.
> 
> > I don't see any changes of deploy-kiali-operator.sh in this PR - what is the plan with deploy-kiali-operator.sh when this PR is merged?
> 
> There are no changes to that file that are required. Installing with deploy-kiali-operator.sh leaves this version unset and thus (as per the comments in kiali_cr.yaml) the "default" version is installed - which means the Kiali version that is the same version as the operator will be installed (which is the current behavior of the script, so no backward incompatible change is occurring here).

Maybe I'm misusing the deploy-kiali-operator.sh but what I do is:

1. get the master version of deploy-kiali-operator.sh
2. set OPERATOR_VERSION_LABEL to one of master,v1.0,v1.12 based on which version of kiali I wan to install
3. set other variables OPERATOR_IMAGE_NAME,OPERATOR_IMAGE_VERSION,KIALI_IMAGE_VERSION,KIALI_IMAGE_NAME which are pointing to our kiali-qe repo where we push our nightly/PR kiali builds.

I'm missing a part how I can use deploy-kiali-operator.sh to deploy e.g. our qe build of kiali from v1.0 branch when this PR is merged when the deploy-kiali-operator.sh is not touching spec.version

jmazzitelli:
Yeah, OK, you are using this is custom way, but, we can change this to support your use-case. I'll add a env var (and its corresponding command line argument) to be able to set this version string, too.

jmazzitelli:
@FilipB see commit https://github.com/kiali/kiali/pull/2219/commits/e21ac332b4d2a562c39d702530863ca459394a6f

There is now a new "--version" command line option (or you can optionally use VERSION env var if you just want to pass env vars around).

Try it out please. I didn't really test it - a quick smoke test shows the script doesn't blow up. But I think this should do what you want - you can now set what "version" you want to go in the Kiali CR by using the --version option or VERSION env var.

FilipB:
@jmazzitelli Thank you. I will play with that tomorrow.

FilipB:
So far I tried:
- VERSION env var passed to deploy-kiali-operator.sh is correctly propagated to kiali cr
- successfully installed kiali using spec.version v1.12 with image_name and image_version empty (version 1.12 was installed)
- successfully installed kiali using spec.version v1.12 and default with image_name and image_version set explicitly to latest qe build of kiali
Found problems:
- warning in kiali operator log:
/usr/local/lib/python3.6/site-packages/kubernetes/config/kube_config.py:509: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  config_dict=yaml.load(f),
- failure when having spec.version: v1.0 and image_name and image_version set explicitly to qe build of kiali v1.0 branch:
[0;31mfatal: [localhost]: FAILED! => {"attempts": 6, "changed": false, "error": 422, "msg": "Failed to patch object: b'{\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"Deployment.apps \\\\\"kiali\\\\\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\\\\\"app\\\\\":\\\\\"kiali\\\\\", \\\\\"version\\\\\":\\\\\"corev1.0uiv1.0\\\\\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable\",\"reason\":\"Invalid\",\"details\":{\"name\":\"kiali\",\"group\":\"apps\",\"kind\":\"Deployment\",\"causes\":[{\"reason\":\"FieldValueInvalid\",\"message\":\"Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\\\\\"app\\\\\":\\\\\"kiali\\\\\", \\\\\"version\\\\\":\\\\\"corev1.0uiv1.0\\\\\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable\",\"field\":\"spec.selector\"}]},\"code\":422}\\n'", "reason": "Unprocessable Entity", "status": 422}

Questions:
- with kiali operator supporting more versions (all versions?) of kiali, I assume that only the latest version of operator is needed, there is no situation when older version (not the latest) is required (e.g. to install older version of kiali)?
- if previous is true, does it make sense to still use OPERATOR_VERSION_LABEL for deploy-kiali-operator.sh? We will always want to install the latest? Maybe just to differentiate master and last released version?

jmazzitelli:
> * warning in kiali operator log:
>   /usr/local/lib/python3.6/site-packages/kubernetes/config/kube_config.py:509: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
>   config_dict=yaml.load(f),

I always see this - nothing to do with this PR. Just comes from python that is included in the ansible operator sdk base image. Nothing to do here.

> * failure when having spec.version: v1.0 and image_name and image_version set explicitly to qe build of kiali v1.0 branch:
>   [0;31mfatal: [localhost]: FAILED! => {"attempts": 6, "changed": false, "error": 422, "msg": "Failed to patch object: b'{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Deployment.apps \\"kiali\\" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\\"app\\":\\"kiali\\", \\"version\\":\\"corev1.0uiv1.0\\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable","reason":"Invalid","details":{"name":"kiali","group":"apps","kind":"Deployment","causes":[{"reason":"FieldValueInvalid","message":"Invalid value: v1.LabelSelector{MatchLabels:map[string]string{\\"app\\":\\"kiali\\", \\"version\\":\\"corev1.0uiv1.0\\"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable","field":"spec.selector"}]},"code":422}\n'", "reason": "Unprocessable Entity", "status": 422}

This looks like the old problem where we had "version" in the selector labels. This should not happen anymore because we removed that. Check how you are deploying the operator (do you have your own custom versions of the deployment yaml?) I suspect you have "version" in the selector that should be removed. We fixed this last year in PR https://github.com/kiali/kiali/pull/1773 - notice we removed the "version" field here: https://github.com/kiali/kiali/commit/1dfbbba93ea272b01f5ee17eb6a72ec589bf9513#diff-0bbda191c507cecd6b2af170a7551f6a

> Questions:
> 
> * with kiali operator supporting more versions (all versions?) of kiali, I assume that only the latest version of operator is needed, there is no situation when older version (not the latest) is required (e.g. to install older version of kiali)?
> * if previous is true, does it make sense to still use OPERATOR_VERSION_LABEL for deploy-kiali-operator.sh? We will always want to install the latest? Maybe just to differentiate master and last released version?

Yes, the latest version of the operator is all that is needed to install v1.0, v1.12 or the latest version. You should not be installing, say, the operator v1.0 anymore to install v1.0. Use the latest operator to do this.

OPERATOR_VERSION_LABEL simply labels the operator Deployment. By default, the deploy script will use the latest version.

jmazzitelli:
> This looks like the old problem where we had "version" in the selector labels. This should not happen anymore because we removed that.

Hmm.. but looks like they are still there in the deployment script v1.0 versions: https://github.com/kiali/kiali/blob/v1.0/operator/deploy/operator.yaml#L14

But this just means you might be installing using the v1.0 deployment script and files, which you should not do. Install using the latest scripts (not the ones out of the v1.0 branch).

FilipB:
> This looks like the old problem where we had "version" in the selector labels. This should not happen anymore because we removed that. Check how you are deploying the operator (do you have your own custom versions of the deployment yaml?) I suspect you have "version" in the selector that should be removed. We fixed this last year in PR #1773 - notice we removed the "version" field here: [1dfbbba#diff-0bbda191c507cecd6b2af170a7551f6a](https://github.com/kiali/kiali/commit/1dfbbba93ea272b01f5ee17eb6a72ec589bf9513#diff-0bbda191c507cecd6b2af170a7551f6a)

I used deploy-kiali-operator.sh with env var OPERATOR_VERSION_LABEL=master which should get yaml files from master branch. So I'm not sure how the "version" label got there if it's only in v1.0.

> OPERATOR_VERSION_LABEL simply labels the operator Deployment. By default, the deploy script will use the latest version.

OPERATOR_VERSION_LABEL also determines from which kiali branch are the yaml files for operator deployment downloaded. I used that to get matching yaml files for given version of operator. So I guess I will still use it for cases where I need last released operator instead of operator built from master branch.

jmazzitelli:
> OPERATOR_VERSION_LABEL also determines from which kiali branch are the yaml files for operator deployment downloaded.

Ah. OK. I see that. We might have to change how this works. Perhaps the script should use VERSION for determining which branch to use (that seems the most logical thing to do?).

FilipB:
> > OPERATOR_VERSION_LABEL also determines from which kiali branch are the yaml files for operator deployment downloaded.
> 
> Ah. OK. I see that. We might have to change how this works. Perhaps the script should use VERSION for determining which branch to use (that seems the most logical thing to do?).

I think it's fine as it is now - e.g. OPERATOR_VERSION_LABEL and OPERATOR_IMAGE_VERSION will specify which version of operator will be installed while VERSION will specify which version of kiali will be installed.

jmazzitelli:
The travis error is unrelated to anything within this PR - I wrote an issue up for the test error here: https://github.com/kiali/kiali/issues/2254

jmazzitelli:
Once this PR is approved and merged, we need to cherry-pick it over to backport into the v1.12 branch.

jmazzitelli:
The backport PR is https://github.com/kiali/kiali/pull/2255

PR #2248, opened 2020-02-14
[operator] do not convert pod annotations to snake case

### LINKED ISSUE #2243: [operator] pod annotations needs to not convert annotation names to snake case
Due to [this issue](https://github.com/operator-framework/operator-sdk/issues/1770) the operator SDK converts all CR settings to snake_case. Sometimes we don't want this, so in the operator we work around this by looking at an internal variable the operator SDK sets for us.

In the case of the [pod annotations](https://github.com/kiali/kiali/blob/v1.13/operator/deploy/kiali/kiali_cr.yaml#L244-L246), it looks like we need to do the same workaround as we do for the tolerations and affinity settings.

(This came up in this issue: https://github.com/kiali/kiali/issues/2216)

jmazzitelli:
fixes #2243

### DIFF
--- a/operator/molecule/affinity-tolerations-resources-test/playbook.yml
+++ b/operator/molecule/affinity-tolerations-resources-test/playbook.yml
@@ -6,7 +6,7 @@
   tasks:
   - import_tasks: ../common/tasks.yml
   - import_tasks: ../asserts/pod_asserts.yml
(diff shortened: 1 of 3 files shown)


jmazzitelli:
this has a molecule test showing that you can now create new pod annotations with camelCase key names.

Issue #2031, opened 2019-12-12
[operator] track changes to operator manifest bundle format


jmazzitelli:
This issue is just to remind us to track future changes coming to the operator manifest bundle format and to change the kiali operator manifests to match that new format.

For details, see: https://github.com/openshift/enhancements/blob/master/enhancements/olm/operator-bundle.md

jmazzitelli:
https://github.com/openshift/enhancements/blob/master/enhancements/olm/operator-registry.md

https://github.com/operator-framework/operator-registry/tree/master/docs/design

PR #2278, opened 2020-02-21
[operator] a new script that builds new operator bundles using existing metadata

jmazzitelli:
This is part of https://github.com/kiali/kiali/issues/2031 - this PR contains a new script that converts the existing metadata into new operator bundles. Builds and uses "opm".

See: https://github.com/operator-framework/operator-registry/blob/master/docs/design/operator-bundle.md

### DIFF
--- a/operator/manifests/create-op-bundles.sh
+++ b/operator/manifests/create-op-bundles.sh
@@ -1,0 +1,89 @@
+#!/bin/bash
+
+################################################
+# This script creates operator bundles, one
+# bundle per version of the operator.
+################################################
+
+log() {
+  echo "LOG: ${1}"
+}
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd -P)"
+cd ${SCRIPT_DIR}
+
+IMAGE_BUILDER="docker"
+
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case $key in
+    -ib|--image-builder)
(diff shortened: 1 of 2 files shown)


operator/manifests/create-op-bundles.sh
```diff
@@ -0,0 +1,89 @@
...
+      package="kiali-ossm"
+    elif [ "${project_dir}" == "kiali-community" ]; then
+      tag="quay.io/kiali/kiali-operator-community.v${version}:latest"
+      package="kiali"
+    elif [ "${project_dir}" == "kiali-upstream" ]; then
+      tag="quay.io/kiali/kiali-operator-upstream.v${version}:latest"
+      package="kiali"
+    fi
+    echo -n "${tag} found in: ${version_dir} | Building..."
+    ${OPM} alpha bundle build --directory ${version_dir} --package ${package} --channels stable --default stable --image-builder ${IMAGE_BUILDER} --tag ${tag} > /dev/null 2>&1
```
gallettilance: You shouldn't be using conversion in your release script. The goal is to use this conversion **once** to migrate existing content, but from then on you are in the bundle format and should not be converting.

jmazzitelli: Right, this script's purpose was just to be used once, in the initial conversion. At that point, the script should never be used again.

jmazzitelli:
we don't need this anymore. conversions have already been done.

Issue #2440, opened 2020-02-24
Istio 1.5 no longer generates secrets - Kiali needs to handle this by default


jmazzitelli:
Re: https://github.com/istio/istio/issues/21407

Istio 1.5 no longer generates secrets by default. Kiali needs this secret to obtain a cert for https by default when installed on K8s (like minikube).

By default, we should probably not expect the secret and just put Kiali behind http when on Kubernetes. This means those who want secure access (https) to the Kiali service should either (a) generate their own secret and tell Kiali about it via the Kiali CR's `spec.identity` setting or (b) put a Gateway or Ingress or something like that in front of Kiali.

When on OpenShift, we use a secret generated for us by OpenShift (so we are not reliant on Istio generating the secret). So we are still good with default behavior on OpenShift.

PR #2441, opened 2020-02-24
Istio no longer generating secrets/certs - so go behind http by default now

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2440

### DIFF
--- a/operator/roles/default/kiali-deploy/tasks/main.yml
+++ b/operator/roles/default/kiali-deploy/tasks/main.yml
@@ -253,12 +253,12 @@
 
 - name: Set default identity cert_file based on cluster type
   set_fact:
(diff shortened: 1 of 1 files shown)


jmazzitelli:
i tested on minikube, all looks good.

lwrigh:
This was just brought up on the Istio UX community call so it might be worth connecting with the Istio team again on this issue. They said they would be happy to do testing along side with it.

jmazzitelli:
FYI: This fix only affects the Kiali Operator, which is not used by Istio nor their istioctl installer/operator (in fact, this bug does not exist if you use Istio's own istioctl installer). In order to test this fix, you would need to actually NOT install Kiali via istioctl and instead test using the Kiali Operator and the kiali.io docs: https://kiali.io/documentation/getting-started/ docs.

See: https://github.com/istio/istio/issues/21407#issuecomment-590642510

Issue #2454, opened 2020-02-27
Enable Kiali to create and manipulate iter8 experiments


fabolive:
This entails create iter8 Experiment CRs and performing actions on them.

This is a subtask of epic #2007

lucasponce:
Closed by #2503

Issue #2453, opened 2020-02-27
Enable Kiali to visualize iter8 experiments


fabolive:
This covers listing iter8 Experiment CRs and their details on Kiali.

This is a subtask of epic #2007

lucasponce:
Close by https://github.com/kiali/kiali/pull/2503

Issue #2007, opened 2019-12-05
Research Iter8 integration with Kiali


lucasponce:
Iter8 project https://iter8.tools/ provide automatic canary deployment features and it shares some synergies with Kiali use cases like described in https://medium.com/kialiproject/kiali-observability-in-action-for-istio-service-mesh-69127f792103

This issue is defined as a research task about potential next steps about how Kiali could integrate some operations and reporting linked with Iter8 scenario.

Subtasks:
- [x] Enable Kiali to create and manipulate iter8 experiments https://github.com/kiali/kiali/issues/2454
- [x] Enable Kiali to visualize iter8 experiments https://github.com/kiali/kiali/issues/2453
- [x] Enable refresh control in Iter8 list page https://github.com/kiali/kiali/issues/2628
- [ ] Create Iter8 experiments as a Service details Wizard https://github.com/kiali/kiali/issues/2629
- [x] Integrate Iter8 generated Istio Config inside Kiali Wizards https://github.com/kiali/kiali/issues/2630
- [x] Enhance Iter8 details page https://github.com/kiali/kiali/issues/2633
- [x] Improve baseline/candidate representation in Iter8 list page https://github.com/kiali/kiali/issues/2634
- [x] Add more examples under demos repo to be used for Iter8 integration https://github.com/kiali/kiali/issues/3235

kalantar:
I’m going to document some initial thoughts (developed together with @fabolive @sriumcp @kimikowang and @sushmarchandran). There is lots about the extensibility of Kiali that we do not understand so many statements are a bit speculative. Please help give shape to the right path forward.

Looking at https://medium.com/kialiproject/kiali-observability-in-action-for-istio-service-mesh-69127f792103, a manual a/b or canary test is documented. Using this as a starting point, we have identified three use canary use cases: (a) Manual (b) Automated and (c) User-Assisted Feedback on the use cases might be the best first step; then more technically on the integration with Kiali.

kalantar:
**(a) Manual**
The flow for a manual canary test is similar to the blog article. I’ve highlighted concepts we’ve added from iter8.

- User deploys a new version of a service
- Using Kiali, the user can inspect metrics [1] (on Service > Inbound Metrics)
- Using Kiali, the user can inspect/update/add success criterion [2] (new function)
- Using Kiali, the user can change the routing. If there are success criteria defined, the _Update Weighted Routing_ dialog can be augmented with a “set to recommended weights” widget
[...]

PR #2461, opened 2020-02-28
Add Kiali Iter8 Extension - EPIC #2007

jadeyliu939:
Kiali Backend changes to support  kiali-ui issue #2453 and issue #2454 ( for EPIC #2007 )

** Describe the change **

Add Iter8 support for Kiali backend 
For Iter8, please reference [Iter8 Tools]( https://iter8.tools) 

Add 3 Iter8 Experiment REST API

1. Get Experiments from specified namespaces ```/api/iter8/experiments``` ```GET```
2. Get Experiment by namespace and namespace ```/api/iter8/experiment/namespaces/{namespace}/name/{name}``` ```GET```
3. Create Experiment ```/api/iter8/experiment``` ```POST```

** Issue reference **

Issue [#2453](https://github.com/kiali/kiali/issues/2453) and [#2454](https://github.com/kiali/kiali/issues/2454) 

** Documentation **

To install Iter8, please follow [install Iter8](https://github.com/iter8-tools/docs/blob/master/doc_files/iter8_install.md). 
(For minikube - it has been tested with with kubernetes version 1.14.2  ```minikube start --memory=32768 --cpus=4 --kubernetes-version=v1.14.2```)
To use Iter8 - please follow the [tutorial](https://github.com/iter8-tools/docs/blob/master/doc_files/iter8_bookinfo_istio.md)

** Known Issues **
1. When retrieving Iter8 Experiment CR, get error ``` forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot get path "/namespaces/istio-system/experiments".
Temporary fix is to Grant super-user access to all service accounts cluster-wide 

```
kubectl create clusterrolebinding serviceaccounts-cluster-admin \
  --clusterrole=cluster-admin \
  --group=system:serviceaccounts
```

### DIFF
--- a/business/iter8.go
+++ b/business/iter8.go
@@ -2,278 +2,148 @@ package business
 
 import (
 	"encoding/json"
+	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	kube "k8s.io/client-go/kubernetes"
-	"github.com/kiali/kiali/config"
-	"k8s.io/client-go/rest"
-	"time"
-	"fmt"
-	"github.com/kiali/kiali/log"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	corev1 "k8s.io/api/core/v1"
-	iter8v1alpha1 "github.com/iter8-tools/iter8-controller/pkg/apis/iter8/v1alpha1"
-	kialiConfig "github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/prometheus/internalmetrics"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const GroupName = "iter8.tools"
-const GroupVersion = "v1alpha1"
-
 type Iter8Service struct {
-	k8s kubernetes.IstioClientInterface
+	k8s           kubernetes.IstioClientInterface
 	businessLayer *Layer
 }
 
-func (in *Iter8Service) GetIter8Info() (models.Iter8Info, error) {
+func (in *Iter8Service) GetIter8Info() models.Iter8Info {
+	var err error
+	promtimer := internalmetrics.GetGoFunctionMetric("business", "Iter8Service", "GetIter8Info")
+	defer promtimer.ObserveNow(&err)
 
-	var SchemeGroupVersion = schema.GroupVersion{Group: GroupName, Version: GroupVersion}
-	var iter8Kind = SchemeGroupVersion.WithKind("Experiment")
+	conf := config.Get()
 
-	if (iter8Kind == schema.GroupVersionKind{}) {
-		return models.Iter8Info{}, nil
+	// It will be considered enabled if the extension is present in the Kiali configuration and the CRD is enabled on the cluster
+	if conf.Extensions.Iter8.Enabled && in.k8s.IsIter8Api() {
+		return models.Iter8Info{
+			Enabled: true,
+		}
 	}
-	log.Infof("Experiment CRD exists")
 	return models.Iter8Info{
-		Enabled: true,
-	}, nil
+		Enabled: false,
+	}
 }
 
-func (in *Iter8Service) GetIter8Experiment(namespace string, name string) (experiment models.ExperimentDetail, err error) {
-	result := &iter8v1alpha1.Experiment{}
-	cacheToken := ""
-	kConfig := kialiConfig.Get()
-	if kConfig.InCluster {
-		if saToken, err := kubernetes.GetKialiToken(); err != nil {
-			return models.ExperimentDetail{}, err
-		} else {
-			cacheToken = saToken
-		}
-	}
-	config, err := kubernetes.ConfigClient()
+func (in *Iter8Service) GetIter8Experiment(namespace string, name string) (models.Iter8ExperimentDetail, error) {
+	var err error
+	promtimer := internalmetrics.GetGoFunctionMetric("business", "Iter8Service", "GetIter8Experiment")
+	defer promtimer.ObserveNow(&err)
+
+	iter8ExperimentDetail := models.Iter8ExperimentDetail{}
+
+	iter8ExperimentObject, err := in.k8s.GetIter8Experiment(namespace, name)
 	if err != nil {
-		return models.ExperimentDetail{}, err
-	}
-	istioConfig := rest.Config{
-		Host:            config.Host,
-		TLSClientConfig: config.TLSClientConfig,
-		QPS:             config.QPS,
-		BearerToken:     cacheToken,
-		Burst:           config.Burst,
+		return iter8ExperimentDetail, err
 	}
 
-	istioClient, err := kubernetes.NewClientFromConfig(&istioConfig)
-	clientset := istioClient.GetK8sApi()
-	path := fmt.Sprintf("/apis/iter8.tools/v1alpha1/")
-	err = clientset.CoreV1().RESTClient().Get().AbsPath(path).
-		Namespace(namespace).
-		Resource("experiments").
-		Name(name).
-		Do().Into(result)
-
-	criterias := make([]models.CriteriaDetail, len(result.Spec.Analysis.SuccessCriteria))
-	for i, c := range result.Spec.Analysis.SuccessCriteria {
-		metricName := c.MetricName
-		criteriaDetail := models.CriteriaDetail {
-			Name: c.MetricName,
-			Criteria : models.Criteria {
-				Metric : c.MetricName,
-				SampleSize : c.GetSampleSize(),
-				Tolerance : c.Tolerance,
-				ToleranceType : string(c.ToleranceType),
-				StopOnFailure:c.GetStopOnFailure(),
-			},
-			Metric: models.Metric {
-				AbsentValue: result.Metrics[metricName].AbsentValue,
-				IsCounter: result.Metrics[metricName].IsCounter,
-				QueryTemplate: result.Metrics[metricName].QueryTemplate,
-				SampleSizeTemplate: result.Metrics[metricName].SampleSizeTemplate,
-			},
-		}
-		criterias[i] = criteriaDetail
-	}
-    trafficControl :=  models.TrafficControl {
-    	Algorithm : result.Spec.TrafficControl.GetStrategy(),
-    	Interval: result.Spec.TrafficControl.GetInterval(),
-    	MaxIteration : result.Spec.TrafficControl.GetMaxIterations(),
-    	MaxTrafficPercentage: result.Spec.TrafficControl.GetMaxTrafficPercentage(),
-    	TrafficStepSize: result.Spec.TrafficControl.GetStepSize(),
-    }
-	return models.ExperimentDetail{
-		ExperimentItem : models.ExperimentListItem {
-			Name: result.Name,
-			Status: result.Status.Message,
-			Baseline: result.Spec.TargetService.Baseline,
-			BaselinePercentage: result.Status.TrafficSplit.Baseline,
-			Candidate: result.Spec.TargetService.Candidate,
-			CandidatePercentage: result.Status.TrafficSplit.Candidate,
-		},
-		CriteriaDetails: criterias,
-		TrafficControl: trafficControl,
-
-	}, nil
+	iter8ExperimentDetail.Parse(iter8ExperimentObject)
+	return iter8ExperimentDetail, nil
 }
 
-func (in *Iter8Service) GetIter8Experiments(namespaces []string) ( []models.ExperimentListItem, error) {
-	experiments := make([]models.ExperimentListItem, 0)
-	if (len(namespaces) == 0) {
+func (in *Iter8Service) GetIter8ExperimentsByNamespace(namespace string) ([]models.Iter8ExperimentItem, error) {
+	var err error
+	promtimer := internalmetrics.GetGoFunctionMetric("business", "Iter8Service", "GetIter8ExperimentsByNamespace")
+	defer promtimer.ObserveNow(&err)
+
+	return in.fetchIter8Experiments(namespace)
+}
+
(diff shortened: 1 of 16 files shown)


rhqci:

lucasponce:
Thanks @jadeyliu939, I'll fetch it today to test it locally and start adding comments.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
+package business
+
+import (
+	"encoding/json"
```
lucasponce: Code style: we try to follow this convention on imports:

```
<standard imports>

<third party>

<kiali ones>

i.e.

	"encoding/json"
	"errors"
	"fmt"
	"strings"
	"sync"

	errors2 "k8s.io/apimachinery/pkg/api/errors"

	"github.com/kiali/kiali/log"
	"github.com/kiali/kiali/util"
```

We don't have yet any automatic task to order these so it depends of IDE settings and manual check.

Don't worry, it still happens to any veteran Kiali developer.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+	"fmt"
+	"github.com/kiali/kiali/log"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	iter8v1alpha1 "github.com/iter8-tools/iter8-controller/pkg/apis/iter8/v1alpha1"
+	kialiConfig "github.com/kiali/kiali/config"
+)
+
+const GroupName = "iter8.tools"
```
lucasponce: Note that these public variables are package scoped to whole "business" package.
So, to avoid clashes better to add a prefix i.e. "Iter8GroupName", "Iter8GroupVersion" I think.

Again, I'm just starting to look the code, probably I can comment later when I put all pieces together.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+const GroupVersion = "v1alpha1"
+
+type Iter8Service struct {
+	k8s kubernetes.IstioClientInterface
+	businessLayer *Layer
+}
+
+func (in *Iter8Service) GetIter8Info() (models.Iter8Info, error) {
```
lucasponce: If I understand the intention of this method, it tries to respond if Iter8 is active or not isn't it ?
I think the current implementation creates only two constants (SchemeGroupVersion and iter8Kind) that don't perform any runtime check.
Perhaps this method could also check if Iter8 is enabled/disabled from config instead ?

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+	istioConfig := rest.Config{
+		Host:            config.Host,
+		TLSClientConfig: config.TLSClientConfig,
+		QPS:             config.QPS,
+		BearerToken:     cacheToken,
+		Burst:           config.Burst,
+	}
+
+	istioClient, err := kubernetes.NewClientFromConfig(&istioConfig)
```
lucasponce: I think that in the in *Iter8Service the iter8 REST client can be already created and to avoid to create it on every invocation.

Perhaps, this can be moved to these:
https://github.com/kiali/kiali/blob/master/kubernetes/client.go#L123
https://github.com/kiali/kiali/blob/master/kubernetes/client.go#L313

I know that perhaps this may trigger a couple of refactoring inside.
Let me know what do you think, I can also try to send a PR to your branch with this particular change as I admit that it may need some re-arrange.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+	clientset := istioClient.GetK8sApi()
+	path := fmt.Sprintf("/apis/iter8.tools/v1alpha1/")
+	err = clientset.CoreV1().RESTClient().Get().AbsPath(path).
+		Namespace(namespace).
+		Resource("experiments").
+		Name(name).
+		Do().Into(result)
+
+	criterias := make([]models.CriteriaDetail, len(result.Spec.Analysis.SuccessCriteria))
+	for i, c := range result.Spec.Analysis.SuccessCriteria {
```
lucasponce: Normally the parsing between the raw result from a REST call and Kiali model is done inside the model/* types, trying to add a Parse() set of methods (for collections and items, i.e. model/virtual_service.go or similar). 

On this case we see a new scenario and we need to decide what to do, I tend to think that following the Kiali convetion will help, even if that means that perhaps that it needs to re-organize a little bit the existing code.

Hopefully it's a basic call & parse scenario and I think it shouldn't be a big effort on this.

I can also to try something locally and send a PR to this branch as well, I knew that probably we will face new scenarios like importing a new API ("/apis/iter.tools/v1alpha1") and in the Extensions documentations I didn't detail this.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+			Name: c.MetricName,
+			Criteria : models.Criteria {
+				Metric : c.MetricName,
+				SampleSize : c.GetSampleSize(),
+				Tolerance : c.Tolerance,
+				ToleranceType : string(c.ToleranceType),
+				StopOnFailure:c.GetStopOnFailure(),
+			},
+			Metric: models.Metric {
+				AbsentValue: result.Metrics[metricName].AbsentValue,
```
lucasponce: Probably yes.
If this case doesn't need protection against panic or runtime errors, perhaps some comment like "metricName" will be always present in the result may help for future revisions.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      newexperimentSpec.Name,
+			Namespace: newexperimentSpec.Namespace,
+		},
+		Spec: iter8v1alpha1.ExperimentSpec {
+			TargetService: iter8v1alpha1.TargetService {
+				ObjectReference: &corev1.ObjectReference {
+					Name: newexperimentSpec.Service,
+					Namespace: "bookinfo-iter8",
```
lucasponce: I guess these constants were left for a demo scenario.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+		Body(&experiment).
+		Timeout(timeout).
+		Do().
+		Into(result)
+	if err != nil {
+		log.Infof("Create return error %s", err)
+	}
+	return result, err
+}
+
```
lucasponce: Okey, I think this class is interesting.
Feature wise, functionality is clear what it tries to do.
Also it's good that this class has triggered some questions in kiali backend, mostly about how to organize REST client declaration and where to put some code (it's good, I expected this).

So, @jadeyliu939 I suggest that I can work on this PR and try to address some of my comments related to re-organization.

Nothing wrong with the PR, just as this is the first extension added in Kiali from a different API and perhaps it requires some previous organization in Kiali in order to add it.

handlers/iter8.go
```diff
@@ -0,0 +1,101 @@
...
+	experimentlists := []models.ExperimentListItem{}
+	params := r.URL.Query()
+	namespaces := params.Get("namespaces") // csl of namespaces
+	if len(namespaces) > 0 {
+		ns := strings.Split(namespaces, ",")
+		experimentlists, err = business.Iter8.GetIter8Experiments(ns)
+	} else {
+		ns := []string{}
+		experimentlists, err = business.Iter8.GetIter8Experiments(ns)
+	}
```
lucasponce: This handler should also have an error response when business methods return an error.

handlers/iter8.go
```diff
@@ -0,0 +1,101 @@
...
+	params := mux.Vars(r)
+	business, err := getBusiness(r)
+	namespace := params["namespace"]
+	name := params["name"]
+	if err != nil {
+		RespondWithError(w, http.StatusInternalServerError, "Services initialization error: "+err.Error())
+		return
+	}
+	log.Infof("Calling get etIter8Experiment with Namespace %s, and name %s ", namespace, name)
+	experiment, err := business.Iter8.GetIter8Experiment(namespace, name)
```
lucasponce: This handler should also have an error response when business methods return an error.

handlers/iter8.go
```diff
@@ -0,0 +1,101 @@
...
+
+func Iter8ExperimentCreate(w http.ResponseWriter, r *http.Request) {
+	business, err := getBusiness(r)
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		RespondWithError(w, http.StatusInternalServerError, "Services initialization error: "+err.Error())
+		return
+	}
+
+	experiment, err := business.Iter8.Iter8ExperimentCreate(body)
```
lucasponce: This handler should also have an error response when business method returns and error.

handlers/iter8.go
```diff
@@ -0,0 +1,101 @@
...
+	params := mux.Vars(r)
+	business, err := getBusiness(r)
+	namespace := params["namespace"]
+	name := params["name"]
+	if err != nil {
+		RespondWithError(w, http.StatusInternalServerError, "Services initialization error: "+err.Error())
+		return
+	}
+
+	experiment, err := business.Iter8.GetIter8Experiment(namespace, name)
```
lucasponce: This handler should also have an error response when business method returns and error.

handlers/iter8.go
```diff
@@ -0,0 +1,101 @@
...
```
lucasponce: This handler should also have an error response when business method returns and error.

models/iter8.go
```diff
@@ -0,0 +1,68 @@
...
+	Permissions ResourcePermissions `json:"permissions"`
+}
+/*
+type ExperimentList struct {
+	Namespace string `json:"namespace"`
+
+	Experiments []ExperimentListItem `json:"experiments"`
+}
+*/
+type ExperimentListItem struct {
```
lucasponce: One suggestion, as these classes are public for the "models" package, I think it could be good if Iter8 classes are namespaced with a "Iter8" suffix.
Perhaps Experiment could have not clash, but classes like CriteriaDetail and Metric are generic and other areas of Kiali may use similar naming so a "Iter8CriteriaDetail" o "Iter8Metric" may help on these cases.

In next iterations we can decide if we put all extension under it's own model/subpackage or we maintain this pattern.

lucasponce:
Thanks @jadeyliu939 for the PR ! I know the effort to catchup with a different project.
The feature proposed is pretty clear, I have some comments related to code organization that I think that I suggest to send you a PR to your branch, so, we can consolidate on this PR.

Nothing wrong, but probably it was me as I didn't document well some aspects:
- How to add a new client for a new API.
- What approach to do for raw data: duplicate some structs from iter8 instead of add whole controller as a new dependency.
- How to place the parsing methods when dealing with REST calls.

Let me send you a PR with my proposal and we can comment on this.

Thanks.

business/iter8.go
```diff
@@ -0,0 +1,279 @@
...
+	kube "k8s.io/client-go/kubernetes"
+	"github.com/kiali/kiali/config"
+	"k8s.io/client-go/rest"
+	"time"
+	"fmt"
+	"github.com/kiali/kiali/log"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	corev1 "k8s.io/api/core/v1"
+	iter8v1alpha1 "github.com/iter8-tools/iter8-controller/pkg/apis/iter8/v1alpha1"
```
lucasponce: Also, as commented, this import may have implications to also add the whole iter8-controller as a dependency.
I wonder if we can just clone the structs needed as we did in Istio too, I'll tackle that too.

business/iter8.go
jadeyliu939: Thanks, will follow this convention  for all the files.

business/iter8.go
jadeyliu939: I notice that all the dependencies package were also checkin to git. I am not sure if I should do that or not.

business/iter8.go
jadeyliu939: Yes, will update to "Iter8GroupName" and "Iter8GroupVersion"

business/iter8.go
jadeyliu939: The intention is to check if the CRD exists in the cluster or not. Of course, this is not enough check to see if the Iter8 is enabled or not ( or running normal or not).  I will check more if all the components of iter8 are running.

business/iter8.go
jadeyliu939: I will take a look at client.go. There might already the k8s core rest client I can use. Thank you

business/iter8.go
jadeyliu939: Yes, I can re-organize the existing code to confirm to Kiali model.

business/iter8.go
jadeyliu939: Yes, understand.

business/iter8.go
jadeyliu939: Oops., sorry about that. Thank you for catching that up.

business/iter8.go
jadeyliu939: Thank you. Sorry about the "not so clean" code :-) And thanks for taking the time to look thru it. But I am not sure what the next step, do i fix the code according to your comments? I can commit to my branch per your comments ( one commit per comment?) And do another PR with detail about what I have done?  ( Sorry, not very familiar with Git's bell and whistle).

lucasponce: Hi @jadeyliu939, I propose you that I'm going to send you a PR to your branch with my suggestions, then if you are agree, we can consolidate them on this PR and merge together once is ok.

lucasponce:
Note, most of my suggestions are consolidated in 
https://github.com/jadeyliu939/kiali/pull/1

Basically the goal of that PR is:
- Avoid to pull all iter8-tools/itercontroller package: it fetches knative and other transitive dependencies that have conflicts with Kiali ones. I don't think we want to update whole Kiali dep tree just for this.
- Organize a little bit the structure of code between kubernetes (native queries to K8S), business (manipulation, create models), handlers.
- A little bit of error handling.
- Other minor changes I saw in the original PR.

lucasponce:
@jadeyliu939 and all, as this one is getting bigger, I'm going to consolidate it in master and send my followup PR to master directly.

It will break CI but I prefer to consolidate first.

lucasponce:
Closing in favor of #2503 which contains fixes for linter, swagger and final naming.
We will iterate changes on following PR against master.

PR #2466, opened 2020-03-02
[hack] new stuff needed for CNI

jmazzitelli:


jmazzitelli:
we are going to need to add a "delete" command to the install-bookinfo-demo.sh script since there is now this extra object we have to delete (not to mention the permission stuff too).

jmazzitelli:
`-db` , `--delete-bookinfo` has been added.

Issue #1541, opened 2019-08-27
Include Sidecar (resource, not the proxy) validations


lucasponce:
I would like to see validations running on my sidecar resources: https://istio.io/docs/reference/config/networking/v1alpha3/sidecar/

> NOTE: Each namespace can have only one Sidecar resource without any workload selector. The behavior of the system is undefined if more than one selector-less Sidecar resources exist in a given namespace. The behavior of the system is undefined if two or more Sidecar resources with a workload selector select the same workload instance.

It also should include the following validations:
- [x] There are workloads with the provided labels in the `workloadSelector` field.
- [x] There aren't two sidecars in the same namespace without `workloadSelector`.
- [x] There aren't two sidecars pointing to the same workload
- [x] IstioEgressLister `hosts` have to be in namespace/dnsName format.
- [x] IstioEgressListener `bind` must have format: x.x.x.x or unix:///path/to/uds or unix://@foobar
- [x] IstioIngressListener `bind` must have  `x.x.x.x` format.
- [x] Global Sidecar (placed in root Namespace) should not have `workloadSelector`

~~IstioEgressListener `hosts` field to ensure access to istio-system services (for both telemetry and policy services)~~ (Only need if you use mixer. Istio 1.5 deprecated Mixer. Not necessary anymore)

JIRA:
https://issues.jboss.org/browse/KIALI-3120

lucasponce:
Now that Kiali can generate Sidecars, it would be worth if validations logic is updated with these scenarios.

xeviknal:
Adding the following validations:
- [ ] There are workloads with the provided labels in the `workloadSelector` field.
- [ ] There aren't two sidecars in the same namespace without `workloadSelector`.
- [ ] Global Sidecar (placed in root Namespace) should not have `workloadSelector`
- [ ] IstioEgressListener `hosts` field to ensure access to istio-system services (for both telemetry and policy services)
- [ ] IstioEgressLister `hosts` have to be in namespace/dnsName format.
- [ ] IstioEgressListener `bind` must have format: x.x.x.x or unix:///path/to/uds or unix://@foobar
- [ ] IstioIngressListener `bind` must have  `x.x.x.x` format.

xeviknal:
All validations in the list are finished. Closing the issue.

PR #2468, opened 2020-03-02
Sidecar validations: no matching workloads and multi sidecar match

xeviknal:
** Describe the change **
Add validations for sidecar resources.

1. There are workloads with the provided labels in the workloadSelector field.
2. There aren't two sidecars in the same namespace without workloadSelector.

Requires to merge quick one in the frontend: https://github.com/kiali/kiali-ui/pull/1679

Next validation will come. Splitting https://github.com/kiali/kiali/issues/1541 for better review.
1.
Screenshot of Kiali Console (26)

2. 
Screenshot of Kiali Console (22)

### DIFF
--- a/models/istio_validation.go
+++ b/models/istio_validation.go
@@ -99,6 +99,7 @@ var ObjectTypeSingular = map[string]string{
 	"servicerolebindings":   "servicerolebinding",
 	"clusterrbacconfigs":    "clusterrbacconfig",
 	"authorizationpolicies": "authorizationpolicy",
+	"sidecars":              "sidecar",
 }
 
 var checkDescriptors = map[string]IstioCheck{
--- a/models/istio_validation.go
+++ b/models/istio_validation.go
@@ -198,36 +198,40 @@ var checkDescriptors = map[string]IstioCheck{
 		Message:  "KIA0903 ServiceRole does not exists in this namespace",
 		Severity: ErrorSeverity,
 	},
+	"sidecar.selector.workloadnotfound": {
+		Message:  "KIA1001 No matching workload found for authorization policy selector in this namespace",
+		Severity: WarningSeverity,
+	},
 	"virtualservices.nohost.hostnotfound": {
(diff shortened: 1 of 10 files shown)


prachiyadav:
Screenshot from 2020-03-04 20-12-15

prachiyadav:
Screenshot from 2020-03-04 20-15-39

prachiyadav:
Screenshot from 2020-03-04 20-22-24

models/istio_validation.go
```diff
@@ -198,36 +199,44 @@ var checkDescriptors = map[string]IstioCheck{
...
+		Message:  "KIA1001 No matching workload found for authorization policy selector in this namespace",
+		Severity: WarningSeverity,
+	},
+	"sidecar.multimatch": {
+		Message:  "KIA1002 More than one selector-less Sidecar in the same namespace",
+		Severity: ErrorSeverity,
+	},
 	"virtualservices.nohost.hostnotfound": {
-		Message:  "KIA1001 DestinationWeight on route doesn't have a valid service (host not found)",
+		Message:  "KIA1101 DestinationWeight on route doesn't have a valid service (host not found)",
```
lucasponce: Let's be careful when changing codes, probably not good idea to update all, I think new codes should be added at the end.

xeviknal: I wanted to keep the list of validations sorted alphabetically. I'll keep next new resources at the end, not to force this change.

jshaughn: +1 to alphabetizing!

lucasponce: These are two different things.
alphabetizing things is good.
But if in version 1.x KIA1001 has a code, you can't change that code in version 1.1 just because we are adding a new element starting by "a".

That's not right.

Otherwise any user would become crazy that KIA1001 in 1.x means something different than KIA1101.

jshaughn: @lucasponce Absolutely correct.  Can not change codes.

Issue #2253, opened 2020-02-17
Add StatefulSet controllers in default Kiali config


lucasponce:
Kiali supports multiple controllers as Workloads.
For performance reasons some of them where disabled by default from the config (i.e. Jobs).

https://github.com/kiali/kiali/blob/master/operator/deploy/kiali/kiali_cr.yaml#L546

There have been more issues asking to add StatefulSet as part of default, as example, for Kafka and related deployments:

> Quite often, Kafka brokers are deployed as a StatefulSet. Because of that, Kiali won't recognize the workload out of the box: StafefulSet fetching needs to be explicitly enabled in config. Maybe we could look for a solution to make it work more easily.

jotak:
@lucasponce I was wondering if this could be controlled somehow from the UI. Something like a workloads type filter. I don't know what is better, just thinking out loud.

jshaughn:
@jotak I was going to make a similar comment.  Themain problem here is that users probably don't know what is being excluded and so it's a mystery when expected workloads don't show up.  Perhaps this should be a filter/queryparam such that the user can see what is selected and not selected.  Or something in that area.  It may be weird because it may need to be a global/masthead type filter, like namespace selection.  I'm not sure, also just thinking out loud...

jotak:
@jshaughn why should it be global? I don't remember, does this workload type restriction affect the graph too?

lucasponce:
Personally, I think that enables UI config just for this, perhaps is too much.
I mean, moving things from the Kiali CR to be modified from the UI and change it perhaps it open more edges that we need to resolve this.

Motivation of not pulling all controllers was mainly performance, but I think that adding all controllers except Jobs (that we know that they have still open issues with istio), and enabling cache (now that QE have performed good testing around it) could be just enough for this case.

Controlling Kiali config from UI sounds interesting but thinking about community perspective, I guess we need to progress better in user/roles for non openshift deployments, otherwise I think we will open more scenarios.

jshaughn:
@jotak The graph gen does pull the workloads to help determine dead nodes (i.e. nodes that aren't backed by pods).

@lucasponce If we think perf is OK by expanding the default workload types to include the types mot often causing problems for users, that's fine too.

jotak:
@lucasponce so enabling the cache would be part of this task, right? (or a pre-requisite). Do we still have some known issues about cache or everything we knew about is now fixed? (like using cache in multi-user environment)

lucasponce:
[...]

PR #2470, opened 2020-03-02
Enable cache, fetch StatefulSets by default

jotak:
Fixes https://github.com/kiali/kiali/issues/2253

Screenshot showing stateful sets in workloads list:

Capture d’écran de 2020-03-02 17-47-40


jotak:
Thanks. I'll drop a quick message on kiali-dev so that everyone is aware that cache is now enabled by default

PR #2472, opened 2020-03-02
Include GRPC failures in Health calculation

### LINKED ISSUE #2457: Istio 1.5: fix health error aggregation with respect to grpc errors
Health error aggregation does not take into account that grpc requests report result in grpc_response_status.  The response_code field only reports http request result.

jshaughn:
Starting is Istio 1.5 GRPC requests will generate request_protocol="grpc" grpc_request_status=<grpcStatus>.  Note that when request_protocol is not "grpc" the additional field is not provided (i.e. not present on the time series).

The health calculation was considering only http errors reported in response_code.  We need to include grpc errors.

Fixes #2457

### DIFF
--- a/business/health_test.go
+++ b/business/health_test.go
@@ -219,6 +219,26 @@ var (
 		Value:     model.SampleValue(3.5),
 		Timestamp: model.Now(),
 	}
+	sampleReviewsToHttpbinGrpc0 = model.Sample{
+		Metric: model.Metric{
+			"source_service":       "reviews.tutorial.svc.cluster.local",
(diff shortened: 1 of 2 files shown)


lucasponce:
> Starting is Istio 1.5

I don't think v1.12 targets Istio 1.5, just to confirm the backport label.

jshaughn:
> I don't think v1.12 targets Istio 1.5, just to confirm the backport label.

@lucasponce Sorry, I understand the confusion.  The GRPC status reporting is introduced upstream in Istio 1.5 but [downstream] it is being applied to Maistra 1.1 on top of istio 1.4.

PR #2478, opened 2020-03-03
Enable cache (missing config)

jotak:
Follow-up on https://github.com/kiali/kiali/pull/2470 (cache wasn't trully enabled)

### DIFF
--- a/operator/deploy/kiali/kiali_cr.yaml
+++ b/operator/deploy/kiali/kiali_cr.yaml
@@ -552,7 +552,8 @@ spec:
 #
 # List of namespaces or regex defining namespaces to include in a cache.
 #    ---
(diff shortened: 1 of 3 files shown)


operator/deploy/kiali/kiali_cr.yaml
```diff
@@ -552,7 +552,8 @@ spec:
 #
 # List of namespaces or regex defining namespaces to include in a cache.
 #    ---
-#    cache_namespaces: []
+#    cache_namespaces:
+#    - ".*"
```
lucasponce: Is it missing the default value also in config.go perhaps ?

jotak: good point, it's updated

PR #2481, opened 2020-03-03
Support canonical_service fields in Istio 1.5+ telemetry

### LINKED ISSUE #2276: Support canonical_service fields in Istio 1.5+ telemetry
Istio 1.5 introduces destination_canonical_service and source_canonical_service into the telemetry metrics.   This is basically the solution to the problem of having different label approaches for specifying what Istio calls the "app" label.  The approach is basically to set the canonical_service field using a prioritized list of possible sources.  There should be new doco for this in Istio but I don't see it yet.  From the original proposal (may have changed) the order is:

1. The service.istio.io/canonical-name label.
2. The app.kubernetes.io/name label.
3. The [Istio] app label
4. The workload name (e.g. Deployment or Job name).

A canonical_revision is also proposed to replace 'version', but it is not included in 1.5.  I expect it has a good chance for 1.6.

UPDATE: canonical_revision is also included in 1.5.

jshaughn:
Istio 1.5 introduces` canonical_service` and `canonical_revision` telemetry labels and deprecates `app` and `version`.   This affects graph, metrics and health as they now need to prefer the new labels when the istio version is 1.5+.  This PR introduces the necessary version-specific handling.  When support for the deprecated fields is removed so can be the version-specific handling.  

Also added Istio dev builds as recognized istio versions.

Fixes #2276

### DIFF
--- a/models/dashboards.go
+++ b/models/dashboards.go
@@ -34,7 +34,7 @@ func buildIstioAggregations(local, remote string) []kmodel.Aggregation {
 	verLabel := "version"
 	if status.IstioSupportsCanonical() {
 		appLabel = "canonical_service"
-		// verLabel = "canonical_revision"  TODO UNCOMMENT WHEN FIELD IS THERE
+		verLabel = "canonical_revision"
 	}
 
 	aggs := []kmodel.Aggregation{
--- a/models/dashboards.go
+++ b/models/dashboards.go
@@ -6,6 +6,8 @@ import (
 
 	"github.com/kiali/k-charted/kubernetes/v1alpha1"
 	kmodel "github.com/kiali/k-charted/model"
+
+	"github.com/kiali/kiali/status"
 )
 
 // ConvertAggregations converts a k8s aggregations (from MonitoringDashboard k8s resource) into this models aggregations
@@ -28,9 +30,16 @@ func ConvertAggregations(from v1alpha1.MonitoringDashboardSpec) []kmodel.Aggrega
 }
 
 func buildIstioAggregations(local, remote string) []kmodel.Aggregation {
+	appLabel := "app"
+	verLabel := "version"
+	if status.IstioSupportsCanonical() {
+		appLabel = "canonical_service"
+		// verLabel = "canonical_revision"  TODO UNCOMMENT WHEN FIELD IS THERE
+	}
+
 	aggs := []kmodel.Aggregation{
 		{
-			Label:       fmt.Sprintf("%s_version", local),
+			Label:       fmt.Sprintf("%s_%s", local, verLabel),
 			DisplayName: "Local version",
 		},
 	}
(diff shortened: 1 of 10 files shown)


jmazzitelli:
How does this affect: https://github.com/kiali/kiali/blob/v1.14.0/operator/deploy/kiali/kiali_cr.yaml#L516-L525 ??

jshaughn:
> How does this affect [our app/version label name config]?

@mazz, right, good question.  This is where things get pretty subtle.  So, they have introduced these canonical fields, set heuristically on the workloads, and then assigned to the _telemetry_.  The value depends on what they find on the service (in order):
1. The service.istio.io/canonical-name label.
2. The app.kubernetes.io/name label.
3. The [Istio] app label
4. The workload name (e.g. Deployment or Job name).
It's all well and good for telemetry because we are guaranteed a consistent value.   But then what, what if we want to match telemetry back to the service?  Well then we're f'd because the telemetry doesn't tell us which value they used.  So if we see canonical_service='foo' we still don't know if that's from option 1,2,3 or 4.  So currently we'll still use the label name set in the config you pointed at.  But actually, we may need to follow up with a series of label selectors because in reality they are now supporting a mix.

mazz:
@jshaughn please remove me from this DL.

jmazzitelli:
@jshaughn My nick is @jmazzitelli :)

jotak:
hi @mazz (and I really mean you), sorry for the troubles, I'd like to say it won't happen again but mistakes always happen. Perhaps you should mute notifications for all the kiali repositories?

Capture d’écran de 2020-03-04 18-35-25

It's mostly https://github.com/kiali/kiali and https://github.com/kiali/kiali-ui (there's a couple of others, less used)

jotak:
@jshaughn 
> But actually, we may need to follow up with a series of label selectors because in reality they are now supporting a mix.

At the moment we expect people to have consistent labelling for a given kiali installation, so they would either use the 1. or the 2. or the 3. or the 4. This has been our assumption til now. Maybe it's still "good enough" if nobody complains ?

jshaughn:
> Maybe it's still "good enough" if nobody complains ?

@jotak I think that is exactly the approach we should take.  We still maintain the same config as before letting people declare the single, preferred label names for app and version.    And we only select using that one pair.  We should still use the the canonical fields in telemetry, but we can wait on anything furthr with respect to getting the config objects.

models/dashboards.go
```diff
@@ -42,17 +51,18 @@ func buildIstioAggregations(local, remote string) []kmodel.Aggregation {
...
 		{
-			Label:       fmt.Sprintf("%s_version", remote),
+			Label:       fmt.Sprintf("%s_%s", remote, verLabel),
 			DisplayName: "Remote version",
 		},
 		{
 			Label:       "response_code",
 			DisplayName: "Response code",
 		},
+		// TODO: Do we need to aggrgate also on grpc_response_status?
```
jotak: yes, we should add that

jshaughn: OK, thanks.  I can open another issue for that, it's not related to this issue.

jshaughn: https://github.com/kiali/kiali/issues/2485

mazz:
@jshaughn ok good idea, thanks.

Issue #2488, opened 2020-03-06
Error: Kiali cache doesn't support sidecars


jotak:
I'm running Kiali master with Istio upstream 1.5

This error is showing up continuously in the UI (overview page): "Could not fetch validations status: Internal Server Error"

Logs show:
```
E0306 09:15:44.805011       1 namespaces.go:77] Kiali cache doesn't support [resource: Sidecar]
```

cc @lucasponce @xeviknal

PR #2489, opened 2020-03-06
Fix error: Kiali cache doesn't support sidecars

jotak:
Need to change the logic in istio_validations, as istio objects can be part or not of cache, we need to fetch types, in cache or not, one by one and not all at the same time.

In the end, some logic in kubernetes package is moved to business package.

Fixes https://github.com/kiali/kiali/issues/2488


hhovsepy:
validation regression test passed

PR #2491, opened 2020-03-09
Add Sidecar egress host validation

xeviknal:
** Describe the change **
Add validation to egress host field[1].

Each field is a string formatted like this: "namespace/fqdn". In the namespace there can be 4 different values: *, ~, ., namespace name.

*/*, ~/*, ./* are always valid scenarios.
ns/*, ns/*.ns.svc.cluster.local and ns/svc.ns.svc.cluster.local are valid when ns = sidecar namespace. Otherwise, cross-ns validation is shown.

ns/*.ns2.svc.cluster.local and ns/svc.ns2.svc.cluster.local is not a valid scenario unless ns = ns2. The service references is in another namespace. cross-ns validation is shown.

ServiceEntries are taken into account.

1. Check if the host is present in the namespace: 
Screenshot of Kiali Console (32)

2. Warns user that the host is in another namespace. kiali can cover this scenario

See the exact spec for this field:
[1] https://istio.io/docs/reference/config/networking/sidecar/#IstioEgressListener

** Issue reference **
https://github.com/kiali/kiali/issues/1541

Bug detected: https://github.com/kiali/kiali/issues/2492

### DIFF
--- a/business/checkers/sidecars/egress_listener_checker.go
+++ b/business/checkers/sidecars/egress_listener_checker.go
@@ -1,0 +1,155 @@
+package sidecars
+
+import (
+	"fmt"
+	"strings"
+
+	core_v1 "k8s.io/api/core/v1"
+
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/models"
+)
+
+type EgressHostChecker struct {
+	Sidecar        kubernetes.IstioObject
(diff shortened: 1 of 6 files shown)


lucasponce:
This is not related to the backend but the icon in the margin for the grey/info case seems gone:
image

xeviknal:
@lucasponce I put it in the description. It is a bug that probably came up from the PF3 clean up.

> Bug detected: #2492

Issue #1511, opened 2019-08-27
Add new filter in list by label


lucasponce:
It will be great if the user can click a label and see all resources with that label in the listcomponent.

New feature post GA filter by labels in list

JIRA:
https://issues.jboss.org/browse/KIALI-2878

lucasponce:
Probably in App List could use an aggregation of labels of Workloads/Services but doable.
Information is stored in UI or being fetched, perhaps with some changes in the current model this feature could be added.

Issue #2233, opened 2020-02-11
Improve filtering capabilities of Overview and List pages


lucasponce:
Overview and List pages provide filtering feature on several criterias.
Additional criteria can be added like filtering by labels or validation status.

These are some of the tasks identified for this effort:
- [x] Filtering by validation status in Istio Config List https://github.com/kiali/kiali/issues/1557
- [x] Filtering by label in Overview page https://github.com/kiali/kiali/issues/1522
- [x] Filtering by label List pages https://github.com/kiali/kiali/issues/1511
- [x] Show labels in Kiali Lists and Overview  https://github.com/kiali/kiali/issues/2547
- [x] Labels filtering: "all of" / "any of" switch  https://github.com/kiali/kiali/issues/2734
- [x] Labels filtering: click on labels to filter https://github.com/kiali/kiali/issues/2735

jotak:
I'd have a couple of suggestions of improvements, not sure if they could be part of this epic or should be standalone tasks (let me know what you prefer).

Currently here's how it looks like:
![Capture d’écran de 2020-05-05 11-05-47](https://user-images.githubusercontent.com/2153442/81051743-93a3b300-8ec2-11ea-9426-c9d30fa627d7.png)

1. It's not obvious what is the boolean operation performed with multiple labels: OR or AND ? So, the screenshot above shows it's a OR. Which is good to widen the search space but not to narrow it. Perhaps we could show a small switch in the UI, showing "all of" / "any of", to leave this choice to the user?

2. It would be really nice if one could just click on a label from the list to automatically add it to the filters. If you ever used Kibana for instance, it's a simple yet powerful way to navigate in data & refine searches.

jotak:
I've created two feature requests, perhaps it can go under this epic, as you prefer:

- https://github.com/kiali/kiali/issues/2734
- https://github.com/kiali/kiali/issues/2735

aljesusg:
The research should be removed from this epic. I'll close this one

aljesusg:
> I've created two feature requests, perhaps it can go under this epic, as you prefer:
> 
> * #2734
> * #2735

Oh I didn't see this. Agree

lucasponce:
I think this EPIC is almost done and merged in master.
Any followup issue can be handled like a regular bug.

Issue #1522, opened 2019-08-27
[Overview page] Filter namespaces by label


lucasponce:
There is this scenario where cluster admins label each namespace with the team name. This way allows admins determine really quickly who has to ping to in case of troubles in that namespace.

The feature would be adding one more filter option into Overview page for filtering ns by label(s). It could be one label or more than one.

JIRA:
https://issues.jboss.org/browse/KIALI-2996

lucasponce:
I even think that this feature can be extended to all list pages too.
I think labels are already mapped and brought to the UI, and adding a similar a text filter option based on labels can be easy to implement.

aljesusg:
We require some changes in backend here...

Options

1. New endpoint to check labels for namespaces
2. Add a queryparam in namespaces endpoint to filter by labels

I prefer the 2º option

jotak:
Question: currently in the merged PR, if I understand correctly, on the overview page the labels that are looked for are the ones on namespaces, is it correct?

It's positive because now I can see, for instance, the "istio-enabled=true" label that I've set on them, it's a valuable information that totally makes sense to show.

However, for filtering when coming from other list pages, it is a bit weird because the filters refer to different objects, it's like comparing apples and oranges. For instance, if from the application page I filter with "version=v1" and see 3 matching applications, then go to overview page: the filter is kept but here it shows 0 namespace / applications because the filter is unrelated.

I'm raising this issue but to be honest I'm not sure how it should be addressed. What perhaps I would expect here is to see all the overview cards, with health items filtered to show only those for version=v1, as it was intended when I filtered from apps page. But it clearly contradicts with the goal to have namespace-based label filters. So there's a dissonance.

Perhaps a solution in-between could be just to clear the active label filters while switching from *list pages to overview, because they don't relate to the same thing ?

aljesusg:
This was merged

PR #2496, opened 2020-03-10
Improve filtering capabilities of Overview and List pages

aljesusg:
Related with https://github.com/kiali/kiali/issues/2233
Fix https://github.com/kiali/kiali/issues/1511 & https://github.com/kiali/kiali/issues/1522
Now we show Labels in 

- Namespaces endpoint
- Services endpoint
- Apps endpoint

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -2,6 +2,7 @@ package business
 
 import (
 	"fmt"
+	"strings"
 	"sync"
 
 	core_v1 "k8s.io/api/core/v1"
@@ -25,7 +26,11 @@ type AppService struct {
 func joinMap(m1 map[string]string, m2 map[string]string) map[string]string {
 	result := m1
 	for k, v := range m2 {
-		m1[k] = v
+		value := v
+		if val, ok := result[k]; ok && !strings.Contains(val, v) {
+			value = fmt.Sprintf("%s,%s", val, value)
+		}
+		result[k] = value
 	}
 	return result
 }
--- a/business/apps.go
+++ b/business/apps.go
@@ -22,6 +22,14 @@ type AppService struct {
 	businessLayer *Layer
 }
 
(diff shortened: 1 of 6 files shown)


business/apps.go
```diff
@@ -22,6 +22,14 @@ type AppService struct {
 	businessLayer *Layer
 }
 
+func joinMap(m1 map[string]string, m2 map[string]string) map[string]string {
```
lucasponce: Not sure if this is right, a join should support multi values.
Example, map1 == version=v1; map2 == version=v2
After a join you would overwrite and have only version=v2.
This is going to cause perhaps some inconsistency and would miss real scenario, I think.

Example: I have a reviews app with version=v1, version=v2, version=v3, I'd like to filter by:
- version label is present (ok).
- version=v1 is present or version=v2 or version=v3 (all of these should send a valid app)

So, if the version=v1 is present in the service or any of the workload labels, if should belong to the app.

I know app are more complex because are not real entities, but a virtual group, but I think in that case would help.

lucasponce: So, for the app then perhaps we may need that the join is something internally like
vesion=v1,v2,v3
As "," is not a valid character for a label, instead to create a map of arrays, it can be used a list of valid values for a label.
Then in the UI it can be used to filter in multivalue, just an idea.

aljesusg: Omg @lucasponce I asked it to you and we talked about overwrite last value XD I'll change it

lucasponce: Really ?
I remember I used the same example.
Well, next time I'll add some notes and an example in the issue to clarify :-)

prachiyadav:
Screenshot from 2020-03-13 00-15-11
Screenshot from 2020-03-13 00-12-53

lucasponce:
As commented, there is a missing case in the overview page labels filter, when label has multiple values.

PR #2500, opened 2020-03-10
Validation: multiple sidecars pointing to one workload

xeviknal:
** Describe the change **
Adding validation that checks whether multiple sidecars are pointing to the same workloads or not.
Istio says:

> The behavior of the system is undefined if two or more Sidecar resources with a workload selector select the same workload instance.

Screenshot of Kiali Console (39)

Test yaml demo:
https://github.com/xeviknal/istio-lab/blob/master/sidecars-examples/multimatch-deployment.yaml


xeviknal:
Docs added in https://github.com/kiali/kiali.io/pull/210

PR #2511, opened 2020-03-11
Remove aggressive filtering of bad telemetry from DeadNodeAppender

### LINKED ISSUE #976: Not seeing edges for kafka events

jshaughn:
Remove aggressive filtering of bad telemetry from DeadNodeAppender because it was also removing good telemetry for headless services.  Instead, filter unwanted telemetry immediately when building the TrafficMap.

This change is based on the discussion in https://github.com/kiali/kiali-ui/pull/1680

Fixes #976

@jotak Try out this change and see how it affects the Kafka situation.

### DIFF
--- a/graph/telemetry/istio/appender/dead_node.go
+++ b/graph/telemetry/istio/appender/dead_node.go
@@ -40,40 +40,7 @@ func (a DeadNodeAppender) applyDeadNodes(trafficMap graph.TrafficMap, globalInfo
 	for id, n := range trafficMap {
 		switch n.NodeType {
 		case graph.NodeTypeService:
-			// a service node with outgoing edges is never considered dead (or egress)
-			if len(n.Edges) > 0 {
-				continue
-			}
-
-			// A service node that is a service entry is never considered dead
-			if _, ok := n.Metadata[graph.IsServiceEntry]; ok {
-				continue
-			}
-
-			// A service node that is an Istio egress cluster is never considered dead
-			if _, ok := n.Metadata[graph.IsEgressCluster]; ok {
-				continue
-			}
-
-			// a service node with no incoming error traffic and no outgoing edges, is dead.
-			// Incoming non-error traffic can not raise the dead because it is caused by an
-			// edge case (pod life-cycle change) that we don't want to see.
-			isDead := true
-		ServiceCase:
-			for _, p := range graph.Protocols {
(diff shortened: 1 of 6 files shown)


jotak:
Just tested the initial use case with headless services, it looks good!

Capture d’écran de 2020-03-12 17-46-09
_(messaging-kafka-brokers is the headless service that was previously unseen)_

I will run a couple of other tests...

jotak:
@jshaughn perhaps this is too aggressive because when I remove all workloads I keep seeing dead (service) nodes:

Capture d’écran de 2020-03-12 17-55-52

jotak:
I believe what we should continue to filter out here, are services that don't show incoming telemetry for the considered time period (meaning, with rate == 0), no?

jshaughn:
> I believe what we should continue to filter out here, are services that don't show incoming telemetry for the considered time period (meaning, with rate == 0), no?

Good point. Now we can treat the surviving service nodes like other nodes and set them dead when they have no incoming or outgoing traffic.  I'll add that back.

jshaughn:
@jotak Pushed another try.    My AWS cluster got reaped while I was testing but I think it may be good.

jotak:
@jshaughn I continue to see the same, but turns out, also on master, so it's not actually a regression. I see this when I delete all objects in my namespace.

There's still a difference between how dead nodes are checked for workloads versus services: deleted workloads are removed by dead node, but deleted services are not.

So I believe there's two things that could be done:
- Either do same as for Workloads, check if the service still exists (with the price of querying k8s, but at least it's cached now by default)
- Or continue to show deleted Workloads (so, I guess the service-to-workload edge would appear, in grey). In any case, it should go away by itself when the selected time frame stops to include past traffic on these nodes.

Since it doesn't look like a regression I guess it's fine to merge this as is. I've retested the case with kafka out-of-mesh, and it still works well!

jshaughn:
@jotak I'll merge this as is so we can get this in for today's 1.15 sprint build and get that headless handling in better shape.  I'll create a new issue to enhance DNA with your first suggestion, filter out service nodes without a backing service.  We definitely do not want to show the deleted workload nodes, that's why we added this appender in the first place, those nodes are unwanted.  Merci!

Issue #2508, opened 2020-03-11
Build failure


israel-hdez:
Starting March 3rd, something went in and started causing a build failure. This is the reason of the missing 1.15.0-snapshot.1 release (no snapshot was released March 6th).

The log of a failed build is attached:

[build-log.txt](https://github.com/kiali/kiali/files/4320026/build-log.txt)

Looks like a mock is missing on GetToken function.

israel-hdez:
Adding the blocker label, because this needs to be fixed so that the release is generated correctly.

lucasponce:
I think is related the config of the CI
I'm goinjg to take a look to see if I can help with the build-log.txt and reproduce locally.

lucasponce:
The command used by Jenkins CI:
```
 make -f deploy/jenkins-ci/Makefile -C . backend-test
```

seems to work locally.

So, still looking at the build-log.txt.

lucasponce:
Ah, I think I got something:

```
GOROOT=/home/lponce/Software/go #gosetup
GOPATH=/home/lponce/Projects/gopath #gosetup
/home/lponce/Software/go/bin/go test -c -o /tmp/___TestAppGraph_in_github_com_kiali_kiali_graph_api github.com/kiali/kiali/graph/api #gosetup
/home/lponce/Software/go/bin/go tool test2json -t /tmp/___TestAppGraph_in_github_com_kiali_kiali_graph_api -test.v -test.run ^TestAppGraph$ #gosetup
=== RUN   TestAppGraph
!!! Set up standard mock
E0313 12:01:04.944846   87522 layer.go:40] Error initializing Kiali Cache. Details: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
--- PASS: TestAppGraph (0.32s)
PASS
```

Some tests seem to swallow some error.
I'll try to address this.

lucasponce:
Fixed by #2515

PR #2515, opened 2020-03-13
Disable cache on mocked api_test

lucasponce:
Related to #2508 

Not sure if it fixed completely, but I'm preparing one PR with my findings.

### DIFF
--- a/graph/api/api_test.go
+++ b/graph/api/api_test.go
@@ -31,7 +31,10 @@ import (
 // Setup mock
 
 func setupMocked() (*prometheus.Client, *prometheustest.PromAPIMock, *kubetest.K8SClientMock, error) {
-	config.Set(config.NewConfig())
+	conf := config.NewConfig()
+	conf.KubernetesConfig.CacheEnabled = false
+	config.Set(conf)
+
 	k8s := new(kubetest.K8SClientMock)
 
 	k8s.On("GetNamespaces").Return(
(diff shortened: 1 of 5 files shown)


lucasponce:
Note, this one is not yet finished, there are a lot of tests that create a custom conf and it requires additional checking.

israel-hdez:
Throwing an idea.
Like other config variables, perhaps it would be OK to read a `CACHE_ENABLED` environment value? The difference is that this would need to be done in the NewConfig method.
If it exists, the value is set to the envvar. If not, then `true` by default.

That way we could set the envvar in the Makefile targets (see https://github.com/kiali/kiali/blob/master/make/Makefile.build.mk#L53-L65) with minimum changes to the code and tests.

jmazzitelli:
> Like other config variables, perhaps it would be OK to read a `CACHE_ENABLED` environment variable?

Several months ago, we ripped out the code to have all the config settings to be read from env. variables. It made the code complex, and hard to understand where the values were coming from. So the ONLY settings we pass through via env vars today are security related (so, for example, you can pass credential or signing key values via env. vars - usually used in combination with secrets).

I would avoid re-introducing more env. variables that are un-related to security values unless we have a good reason to do so. For this particular test issue, isn't there a way we can pass in test config .yaml?

israel-hdez:
@jmazzitelli 

> For this particular test issue, isn't there a way we can pass in test config .yaml

In unit tests, we are not reading a yaml :-(
It's just the hardcoded defaults and tests customize the defaults depending on the goals of the tests.
Not sure really how we could avoid that.

Issue #2485, opened 2020-03-05
1.5: Add dashboard aggregation for grpc_response_status


jshaughn:
Similar to what we do[ here ](https://github.com/kiali/kiali/blob/master/models/dashboards.go#L52-L55)for response-code, we need to add for the new grpc_response_status.

PR #2519, opened 2020-03-16
Add GRPC status as dashboards available aggregation/filters

jotak:
Fixes https://github.com/kiali/kiali/issues/2485


jotak:
@jshaughn @hhovsepy I haven't really tested yet. What demo can I use to check GRPC statuses?

PR #1115, opened 2019-06-03
KIALI-2958: Move to UBI7 as our base image

mwringe:
** Describe the change **

Changes the base image to be UBI7

** Issue reference **

https://issues.jboss.org/browse/KIALI-2958

** Backwards incompatible? **

Everything should work the same as before, we are just using a different base image now that seems to have more updated packages.

PR #2532, opened 2020-03-19
Add arm64 support for kiali container image

### LINKED ISSUE #1091: Feature request: Build ARM/ARM64 images
It would be great to have kiali/kiali on docker hub to work on ARM machines.

MrXinWang:
** Describe the change **

Modified `Makefile` and `Dockerfile` to support building kiali container image on arm64, without affecting current images.

Operator image will still be x86_64 only.

Signed-off-by: Henry Wang <<Henry.Wang@arm.com>>

** Issue reference **

Fixes #1091 

** Backwards incompatible? **

No.

** Documentation **

Updated.


rhqci:

Makefile
```diff
@@ -56,9 +56,11 @@ NAMESPACE ?= istio-system
 GOPATH ?= ${HOME}/go
 
 # Environment variables set when running the Go compiler.
+GOOS ?= $(shell go env GOOS)
+GOARCH ?= $(shell go env GOARCH)
```
jmazzitelli: I just want to make sure this is going to work when Travis and Jenkins performs our builds.
What if the build machine reports different GOOS and GOARCH? (i.e. are Travis/Jenkins build machines linux/arch64?)

deploy/docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.access.redhat.com/ubi7-minimal
+FROM registry.access.redhat.com/ubi8-minimal
```
jmazzitelli: This needs a large discussion before we merge this. Moving from UBI 7 and UBI 8 should not be taken lightly. We need to find out what ramifications this has.

MrXinWang: Yes I agree, but as far as I am concerned, ubi7 does not support arm64...I am wondering if the cross build would work, if so the change of base image can be removed.

Makefile
MrXinWang: Same as below, we still need to discuss how we build the multiarch image (by either cross building or building on arm64 natively). But the idea of this change is: it allows the buiding can be carried out on aarch64, without affecting the current building on x86_64. I think the CI machine is x86_64 only, but by trying configuation in https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support, we can build on both x86_64 and aarch64 machines by applying this change here.

make/Makefile.container.mk
```diff
@@ -44,7 +44,14 @@ container-build-operator: .ensure-operator-sdk-exists
...
 ## container-build: Build Kiali and Kiali operator container images
-container-build: container-build-kiali container-build-operator
+container-build:
+# On x86_64 machine, build both kiali and operator images.
+ifeq ($(GOARCH),amd64)
+	make container-build-kiali container-build-operator
+# On other achitectures, only build kiali image.
+else
+	make container-build-kiali
+endif
```
israel-hdez: You will need a similar conditional for the `container-push` target, to avoid pushing the operator image that won't be built if `GOARCH` is not amd64.

Makefile
israel-hdez: Travis reports GOOS=linux and GOARCH=amd64. (See first and last lines in highlighted range https://travis-ci.org/github/kiali/kiali/builds/664245023#L182)

Jenkins also reports the same. This looks safe.

deploy/docker/Dockerfile
israel-hdez: I see this:

```bash
$ docker buildx imagetools inspect registry.access.redhat.com/ubi7-minimal:latest
Name:      registry.access.redhat.com/ubi7-minimal:latest                                                                                                                                        
MediaType: application/vnd.docker.distribution.manifest.list.v2+json                                                                                                                             
Digest:    sha256:8c74fc2c392e6816049068ef1bcd71068eaf40ae6d94b233c5c98d9129e17c84              
                                                
Manifests:                                                                                                                                                                                       
  Name:      registry.access.redhat.com/ubi7-minimal:latest@sha256:adf8765f5a564df0888c441d4c285eb34800ab175e079c93581870741e6954c3                                                              
  MediaType: application/vnd.docker.distribution.manifest.v2+json                               
  Platform:  linux/amd64    
                                                                                                
  Name:      registry.access.redhat.com/ubi7-minimal:latest@sha256:4902f66bb401bd4b2171b7ab0c267c24ce21e7bd4debc749d55210baa25a8197
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/ppc64le
              
  Name:      registry.access.redhat.com/ubi7-minimal:latest@sha256:6e1e81408e49b1e0f54c6bb23ce9b385ea543ecd1258b844825b20860cd220a8
  MediaType: application/vnd.docker.distribution.manifest.v2+json
  Platform:  linux/s390x
```

ARM64 is not listed for UBI7. So, looks like @MrXinWang is right and this change is probably required to move forward with builds for ARM64.

I don't think this move is that bad. In #1115 we moved from centos7 to UBI7. Then in #1199 we moved from UBI7 to UBI7-minimal. In both cases, only sanity checks were done. So, I think a sanity check is all we need in order to merge this.

jmazzitelli: moving to UBI8 may (IIRC) cause this to not be able to run on Red Hat 7.x systems. I will test that out to see.

make/Makefile.container.mk
MrXinWang: Done. I will check how to make the operator work and the process of that feature release. If they support arm64, we can also remove both conditionals.

deploy/docker/Dockerfile
MrXinWang: @jmazzitelli Thanks!

jmazzitelli: Smoke test on my end did not show any problems. But I remember hearing people (I think it was during OLM/Operator discussions) that UBI8 images cannot run or are not supported on RHEL 7.x platforms. So we need to research this more. If UBI8 is not supported on RHEL 7 platforms, then we need to discuss this PR further because we can't lose support on RHEL 7 systems.

MrXinWang: Any suggestions for me to do for researching this problem? :)))) Is there a convenient way to setup an development environment of RHEL 7 to test this?

aljesusg: Maybe we can have two images one for rhel7 and another for rhel8

jmazzitelli: I will ask around about the UBI8 / RHEL issue. I don't have any details, just something I thought I remember hearing. I'll report back.

jmazzitelli: I was correct in my original thinking.

https://access.redhat.com/support/policy/rhel-container-compatibility

I confirmed that the RHEL team has said they will support UBI7 images on top of RHEL8 host, but the reverse is not necessarily true. While it is believed that most UBI8 images would work on a RHEL7 host, those images are not officially supported by RHEL. So if there is a problem the responsibility falls on the container owner to work it out.

So, IMO, this is a no-go if the amd64 images produced by our builds are based on UBI8.

Perhaps think about two Dockerfiles and somehow switching between them (when ARM, use the Dockerfile-UBI8, when amd, use Dockerfile-UBI7). That's just one thought. Not sure if there are other possible solutions. But we can't currently have the amd64 image based on UBI8.

make/Makefile.container.mk
```diff
@@ -44,7 +44,14 @@ container-build-operator: .ensure-operator-sdk-exists
 	cd "${ROOTDIR}/operator" && "${OP_SDK}" build --image-builder ${DORP} --image-build-args "--pull" "${OPERATOR_QUAY_TAG}"
 
 ## container-build: Build Kiali and Kiali operator container images
-container-build: container-build-kiali container-build-operator
+container-build:
+# On x86_64 machine, build both kiali and operator images.
+ifeq ($(GOARCH),amd64)
+	make container-build-kiali container-build-operator
```
jmazzitelli: Don't call "make" directly. At minimum use `$(MAKE)` but even doing that is going to have unintended consequences.

Since all you are doing is restricting the dependent targets, do something like:

```
ifeq ($(GOARCH),amd64)
container-build: container-build-kiali container-build-operator
else
container-build: container-build-kiali
endif
```

We do similar things in `make/Makefile.operator.mk` , target `kiali-create` - so see that for an example.

make/Makefile.container.mk
```diff
@@ -78,4 +85,11 @@ else
...
 ## container-push: Pushes all container images to quay and docker hub.
-container-push: container-push-kiali-quay container-push-kiali-docker container-push-operator-quay
+container-push:
+# On x86_64 machine, push both kiali and operator images.
+ifeq ($(GOARCH),amd64)
+	make container-push-kiali-quay container-push-kiali-docker container-push-operator-quay
+# On other achitectures, only push kiali image.
+else
+	make container-push-kiali-quay container-push-kiali-docker
+endif
```
jmazzitelli: Same as above. Do not call `make` directly. Do this instead:

```
ifeq ($(GOARCH),amd64)
container-push: container-push-kiali-quay container-push-kiali-docker container-push-operator-quay
else
container-push: container-push-kiali-quay container-push-kiali-docker
endif
```

deploy/docker/Dockerfile
aljesusg: Goof to know

IMO in any moment we'll move to rhel8 so maybe is not a bad idea have 2 dockerfile and start to test rhel8. What do you think?

jmazzitelli: I think having two Dockerfiles - one for UBI7 and one for UBI8 - would be nice. We should have a flag to switch between the two (rather than the GOARCH - define an env var "KIALI_DOCKER_FILE" - the default should use UBI7 if amd64 and the default should use UBI8 if ARM, but you should be able to override by passing the env var in which would allow us to build amd64 Kiali using UBI8 so we can test it). This would do all of that - put this in the root `Makefile` after the line that defines GO_BUILD_ENVVARS:

```
ifeq ($(GOARCH),amd64)
KIALI_DOCKER_FILE ?= Dockerfile-ubi7-minimal
else
KIALI_DOCKER_FILE ?= Dockerfile-ubi8-minimal
endif
```

If KIALI_DOCKER_FILE env var is already defined, it will be used (that's what the "?=" does).

Then, where ever we run the `docker build` commands, we need to pass in `-f ${KIALI_DOCKER_FILE}`.

aljesusg: great @jmazzitelli :)

make/Makefile.container.mk
MrXinWang: Done.

deploy/docker/Dockerfile
MrXinWang: @jmazzitelli Great solution indeed! I have updated the code, please review again. Thanks!

make/Makefile.container.mk
MrXinWang: Done. Thanks for this :))

make/Makefile.container.mk
```diff
@@ -44,7 +44,14 @@ container-build-operator: .ensure-operator-sdk-exists
 	cd "${ROOTDIR}/operator" && "${OP_SDK}" build --image-builder ${DORP} --image-build-args "--pull" "${OPERATOR_QUAY_TAG}"
 
 ## container-build: Build Kiali and Kiali operator container images
+container-build:
```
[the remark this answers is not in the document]
MrXinWang: Ohhhhhhh...sorry for this, removed. Thanks for the carefully checking!

make/Makefile.container.mk
```diff
@@ -78,4 +85,11 @@ else
 endif
 
 ## container-push: Pushes all container images to quay and docker hub.
+container-push:
```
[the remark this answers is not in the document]
MrXinWang: Same as above..

jmazzitelli:
The changes look good. I won't have time to test this soon. We'll want to confirm this is all backwards compatible with the way we currently build (it looks like it is, but without running it, I cannot say with certainty).

jmazzitelli:
I did a quick build test and everything built fine on amd64.  I tried "make container-build" to use the default (KIALI_DOCKER_FILE was set to the UBI7 one). I then set KIALI_DOCKER_FILE to point to the UBI8 file and ran the build and it seemed to do the correct thing.

MrXinWang:
Tested on arm64 machine and the base image can be correctly set to ubi8 and the image can also be built successfully.

README.adoc
```diff
@@ -151,7 +151,7 @@ cd ${GOPATH}/src/github.com/kiali/kiali
 make container-build
 ----
 
-This builds the Kiali operator image, too.
+On x86_64 machine, this will build both the Kiali and the Kiali operator image. On aarch64 machine, this will only build the Kiali image.
```
israel-hdez: aarch64 :question:

MrXinWang: Created an issue in https://github.com/operator-framework/operator-sdk/issues/2686 in order to enable the building of operator image on arm. Please feel free to leave any comments there.

I checked the method of building from source, in order to achive this, we need to use a few `sed` command to apply changes to their `Makefile` , I personally would not prefer that way, but if the community has request, I will do that change also.

jmazzitelli: > aarch64

See: https://en.wikichip.org/wiki/arm/aarch64

jmazzitelli:
I'm ok with this. Though without an ARM build of the operator, you will not be able to install the ARM build of the Kiali image.

israel-hdez:
Merging...
Thanks @MrXinWang . I await the follow-up PRs :)

PR #2544, opened 2020-03-20
Sidecar in istio ns shouldn't have workloadSelector

xeviknal:
** Describe the change **
Sidecars placed in istio namespace will be defaulted to all sidecars in the mesh that don't have any sidecar applied to. 

** Issue reference **

https://github.com/kiali/kiali/issues/1541

** Documentation **
Screenshot of Kiali Console (25)


hhovsepy:
Screenshot from 2020-03-20 15-30-41

PR #2545, opened 2020-03-20
Fix service graph issue with percent traffic values

### LINKED ISSUE #2483: In Graph Service View 200% traffic?
**General remarks**
In Service graph while under service graph mode.

A picture says 1000 words:

Kiali_Console

Here is the versioned app graph view:

Kiali_Console

**Versions used**
Kiali: Master
Istio: 1.5.0-beta.5
On Minishift

jshaughn:
When building the service graph, ensure outgoing metrics reflect only traffic from edges reflected in the final graph.

Pay attention to the edge(s) from reviews to ratings...

Workload Graph:
image

Before:
image

After:
image

This has been wrong for quite a while but the issue can be subtle as edge labeling must be %traffic and there must be multiple workload versions in play.  This fixes at least one other analogous scenario.

Fixes #2483

### DIFF
--- a/graph/api/testdata/test_service_graph.expected
+++ b/graph/api/testdata/test_service_graph.expected
@@ -20,8 +20,7 @@
             {
               "protocol": "http",
               "rates": {
-                "httpIn": "20.00",
-                "httpOut": "20.00"
+                "httpIn": "20.00"
               }
             }
           ],
@@ -48,8 +47,7 @@
                 "httpIn": "80.00",
                 "httpIn3xx": "20.00",
                 "httpIn4xx": "20.00",
-                "httpIn5xx": "20.00",
-                "httpOut": "80.00"
+                "httpIn5xx": "20.00"
               }
             }
           ]
@@ -72,7 +70,13 @@
               "protocol": "http",
               "rates": {
                 "httpIn": "170.00",
-                "httpOut": "170.00"
+                "httpOut": "160.00"
+              }
+            },
+            {
+              "protocol": "tcp",
+              "rates": {
+                "tcpOut": "31.00"
               }
             }
           ]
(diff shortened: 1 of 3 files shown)


jshaughn:
@pbajjuri20 In addition to the "percent traffic" labeling you should probably validate the request rate labeling.  Note that response time labeling will not show anything, that's expected for service graph because it reflects only traffic flow in a more logical sense.  Also note that TCP edges will not show %request, that's normal as TCP metrics don't reflect request rate, only bytes sent.

pbajjuri20:
Service graph: Request Percentage 

percentage

Service graph: Request Time

percentage time


@jshaughn: Thanks, It having the perfect information.

Tested and passed on Kiali UI 1.16.0 (2d7e2aa3b3d1d985f7def5bb783e309637f23adc)

pbajjuri20:
Service Graph: Request percentage.

percentage

It having perfect information.

pbajjuri20:
Service Graph: Response Time

It is showing some time response from Kiali-traffic-generator to unknown service.

percentage time

jshaughn:
@pbajjuri20 That's an interesting finding.  Despite what I said ("...response time labeling will not show anything, that's expected for service graph") it's not a regression or behavioral change related to this fix.  It's a corner case  of "unknown" service.  The edge from traffic-generator to "unknown" is a service-to-service edge that is present even in the underlying workload graph and has been carried over to the service graph (maintaining its response time information).  If you confirm that this edge is also in the workload graph then  we can safely ignore it here.

pbajjuri20:
Workload: Response Time

The Workload Graph also showing the time response from "Kiali-traffic-generator" to "unknown service".


workload

It has been carried from the Workload graph so that we can ignore it safely.

pbajjuri20:
Tested and Passed on Kiali UI 1.16.0 (2d7e2aa3b3d1d985f7def5bb783e309637f23adc)

PR #2553, opened 2020-03-24
Fix Makefile cluster-push target

israel-hdez:
It got broken by the renaming of the Dockerfiles. This seems to fix it.

### DIFF
--- a/make/Makefile.cluster.mk
+++ b/make/Makefile.cluster.mk
@@ -84,11 +84,11 @@ cluster-build-operator: .prepare-cluster container-build-operator
 ## cluster-build-kiali: Builds the Kiali image for development with a remote cluster
 cluster-build-kiali: .prepare-cluster container-build-kiali
 ifeq ($(DORP),docker)
-	@echo Building container image for Kiali for a remote cluster using docker
-	docker build -t ${CLUSTER_KIALI_TAG} ${OUTDIR}/docker
+	@echo Re-tag the already built Kiali container image for a remote cluster using docker
+	docker tag ${DOCKER_TAG} ${CLUSTER_KIALI_TAG}
 else
-	@echo Building container image for Kiali for a remote cluster using podman
-	podman build -t ${CLUSTER_KIALI_TAG} ${OUTDIR}/docker
+	@echo Re-tag the already built Kiali container image for a remote cluster using podman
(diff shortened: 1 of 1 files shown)


jmazzitelli:
we probably could have done that differently in the original code - instead of docker build, just use docker tag. But for now, this is fine.

jmazzitelli:
I only suggest that because as it is now, that target is building the container twice. With that change, it builds one via container-build and then just tags the image with the remote cluster repo name (so no need to do that extra build which it doesn't need to do).

make/Makefile.cluster.mk
```diff
@@ -85,10 +85,10 @@ cluster-build-operator: .prepare-cluster container-build-operator
 cluster-build-kiali: .prepare-cluster container-build-kiali
 ifeq ($(DORP),docker)
 	@echo Building container image for Kiali for a remote cluster using docker
-	docker build -t ${CLUSTER_KIALI_TAG} ${OUTDIR}/docker
+	docker build -t ${CLUSTER_KIALI_TAG} -f ${OUTDIR}/docker/${KIALI_DOCKER_FILE} ${OUTDIR}/docker
```
jmazzitelli:
```suggestion
	docker tag ${CONTAINER_NAME}:${CONTAINER_VERSION} ${CLUSTER_KIALI_TAG}
```

make/Makefile.cluster.mk
```diff
@@ -85,10 +85,10 @@ cluster-build-operator: .prepare-cluster container-build-operator
...
 else
 	@echo Building container image for Kiali for a remote cluster using podman
-	podman build -t ${CLUSTER_KIALI_TAG} ${OUTDIR}/docker
+	podman build -t ${CLUSTER_KIALI_TAG} -f ${OUTDIR}/docker/${KIALI_DOCKER_FILE} ${OUTDIR}/docker
```
jmazzitelli:
```suggestion
	podman tag ${CONTAINER_NAME}:${CONTAINER_VERSION} ${CLUSTER_KIALI_TAG}
```

jmazzitelli:
travis did pass this build, but for some reason this PR is waiting for results: https://travis-ci.org/github/kiali/kiali/builds/666567000

jmazzitelli:
re-opening to see if it wakes up the travis notification

PR #2560, opened 2020-03-25
Fixes for CVE-2020-1764 and CVE-2020-1762

israel-hdez:
Bringing to master fixes for bulleting [KIALI-SECURITY-001](https://kiali.io/news/security-bulletins/kiali-security-001/)

---

* [operator] generate random signing key by default
* Strengthen security of Kiali session with some additional validations

---

This is fixing some issues around authentication mechanisms.

For **all** auth strategies: 
* All sessions are invalidated if the signing key is changed
* If subject and/or expiration claims are missing in the Kiali token, the token is rejected and user is treated as not logged in.
* Usage of empty signing key (which provides no security) and *kiali* signing key (which is a well-known key) are now banned. Kiali server won't start and/or will panic if these keys are configured.

For **login** auth strategy:
* All sessions will be invalidated if username and/or passphrase are changed (mismatch between configured username/password and the data from the token)
* Add tests validation the behavior of this strategy.

For **openshift** auth strategy:
* Require presence of sid claim in the Kiali token
* Do proper validation of the OpenShift session token
  * Looks like ValidateToken method in openshift_oauth.go source is using endpoints of the cluster API which are unprotected. So, the session token was being ignored by the cluster API effectively meaning that the token wasn't being validated. This was allowing malicious requests to be processed by Kiali.
  * Removing that ValidateToken method and using GetUserInfo as a replacement.

For **token** auth strategy:
* Require presence of sid claim in the Kiali token

For **ldap** auth strategy:
* Add the "issuer" claim in the token, and require it when validating it on every request.

### DIFF
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -32,6 +32,23 @@ rules:
   - get
   - list
   - patch
+- apiGroups: [""]
+  resources:
+  - secrets
+  verbs:
+  - create
+- apiGroups: [""]
+  resourceNames:
+  - kiali-signing-key
+  resources:
+  - secrets
+  verbs:
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
 - apiGroups: ["apps"]
   resources:
   - deployments
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -37,6 +37,8 @@ rules:
   - secrets
   verbs:
   - create
+  - list
+  - watch
 - apiGroups: [""]
   resourceNames:
   - kiali-signing-key
--- a/operator/manifests/kiali-community/1.15.1/kiali.v1.15.1.clusterserviceversion.yaml
+++ b/operator/manifests/kiali-community/1.15.1/kiali.v1.15.1.clusterserviceversion.yaml
@@ -334,6 +334,8 @@ spec:
           - secrets
           verbs:
           - create
+          - list
+          - watch
         - apiGroups: [""]
           resourceNames:
           - kiali-signing-key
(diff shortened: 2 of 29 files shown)


operator/deploy/role.yaml
```diff
@@ -32,6 +32,23 @@ rules:
   - get
   - list
   - patch
+- apiGroups: [""]
+  resources:
+  - secrets
+  verbs:
+  - create
```
jmazzitelli: let's do this now while we have it in a PR... add "list" and "watch" permissions here.
```suggestion
  - create
  - list
  - watch
```

operator/manifests/kiali-community/1.15.1/kiali.v1.15.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,570 @@
...
+          - namespaces
+          verbs:
+          - get
+          - list
+          - patch
+        - apiGroups: [""]
+          resources:
+          - secrets
+          verbs:
+          - create
```
jmazzitelli:
```suggestion
          - create
          - list
          - watch
```

operator/manifests/kiali-upstream/1.15.1/kiali.v1.15.1.clusterserviceversion.yaml
```diff
@@ -0,0 +1,570 @@
...
+          - namespaces
+          verbs:
+          - get
+          - list
+          - patch
+        - apiGroups: [""]
+          resources:
+          - secrets
+          verbs:
+          - create
```
jmazzitelli:
```suggestion
          - create
          - list
          - watch
```

jmazzitelli:
re-opening to try and kick-start travis

PR #2563, opened 2020-03-25
[kiali] [v1.15] update role for cve

jmazzitelli:
this is a cherry pick of https://github.com/kiali/kiali/pull/2564 and also adds some metadata for v1.15.1

### DIFF
--- a/operator/deploy/role.yaml
+++ b/operator/deploy/role.yaml
@@ -37,6 +37,8 @@ rules:
   - secrets
   verbs:
   - create
+  - list
+  - watch
 - apiGroups: [""]
   resourceNames:
   - kiali-signing-key
--- a/operator/manifests/kiali-community/1.15.1/kiali.v1.15.1.clusterserviceversion.yaml
+++ b/operator/manifests/kiali-community/1.15.1/kiali.v1.15.1.clusterserviceversion.yaml
@@ -334,6 +334,8 @@ spec:
           - secrets
           verbs:
           - create
+          - list
+          - watch
         - apiGroups: [""]
           resourceNames:
           - kiali-signing-key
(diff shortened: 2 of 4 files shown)


operator/deploy/role.yaml
```diff
@@ -32,6 +32,25 @@ rules:
   - get
   - list
   - patch
+- apiGroups: [""]
+  resources:
+  - secrets
+  verbs:
+  - create
+  - list
+  - watch
```
israel-hdez: I though you were going to add just these two lines. Isn't this duplicating the permission. I see again the same a few lines down here.

jmazzitelli: Technically, yes, lines 50 and 53 duplicate the perms on 40 and 41.. .but for consistency sake, we list all the perms we need for "kiali-signing-key" down below just to make it clear to anyone who looks at this code that, yes, we need list and watch for signing key. If we take away lines 50 and 53 it isn't clear that we DO need them but we are just getting them from lines 40 and 41. The way it is here, this is more explicit and makes it clear what permissions we need for kiali-signing-key. It is OK to have permissions listed like that (nothing breaks because of it).

israel-hdez: I was meaning that lines 35-41 are the same as lines 54-58. I though you were going to add list and watch to that section instead of duplicating.

Analogously, lines 42-53 ~= 59-70.

jmazzitelli: eeekkk... I didn't see that.

jmazzitelli: OK, good catch. I didn't even notice the "secrets" perms were already here. I fixed, just added the two list and watch.

jmazzitelli:
OK, this is now ready for merge I believe

PR #2569, opened 2020-03-26
Fix rest.Config use https when incluster is false

shalk:
** Describe the change **

use https instead of http

** Issue reference **

None

** Backwards incompatible? **

no

** Documentation **

### DIFF
--- a/kubernetes/client.go
+++ b/kubernetes/client.go
@@ -198,7 +198,7 @@ func ConfigClient() (*rest.Config, error) {
 	}
 	return &rest.Config{
 		// TODO: switch to using cluster DNS.
-		Host:  "http://" + net.JoinHostPort(host, port),
+		Host:  "https://" + net.JoinHostPort(host, port),
 		QPS:   kialiConfig.Get().KubernetesConfig.QPS,
 		Burst: kialiConfig.Get().KubernetesConfig.Burst,
 	}, nil


rhqci:

shalk:
> Can you ellaborate more this change ?
> 
> InCluster variable is used for development and typically don't require https port.
> This can break some developers config, so just trying to understand the changes.

when I use login Options "token", i found  that when incluster  is true, api address is internal ip, so only secret `istiod-service-account-token-xxx`  works , I can not use token of  other namespace.

So I found i can config incluster = false, "KUBERNETES_SERVICE_HOST" = my-api-server-domain-name,"KUBERNETES_SERVICE_PORT"=443,  but when incluster = false， it's hard code to http。

 We have a proxy in front of our api-server and  support  `self defined short token` (works like beared token). 

**If there a way to config rest.Config.HOST that shoud be better.**

I think there are a way to solve this  and not break developer's config。

when  InCluster is false,
try to 
read value from  env KUBERNETES_API_SERVER_URL_CUSTOM or ,
read value from yaml `kubernetes_api_server_url`
if value is exist ,  the url   can overide rest.config.HOST

Can you give some advise please?


Thank you

lucasponce:
Thanks @shalk for the comment.

Technically we can introduce a KUBERNETES_SERVICE_PROTOCOL Getenv for this.

But, not sure if I understand from your comment, is your Kiali deployed out of your cluster ?

Why the ```func InClusterConfig()``` doesn't work in your environment ?

shalk:
I'm not familar with k8s api server.
My Kiali is deployed in my cluster.  

```
$ echo $KUBERNETES_SERVICE_HOST
10.254.0.1
$ echo $KUBERNETES_SERVICE_PORT
443
```
I tried  some service accounts  from other namespace,  
these service accounts works with `curl` command ([refer doc](https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#without-kubectl-proxy))

But kiali-ui just return

```
Unauthorized. The provided credentials are not valid to access Kiali.

Please check your credentials and try again.
```
I opened debug log  but found nothing.

Someone tell me **Pod can only use its own token to connect to internel ip**.
Is these word is wrong ?

kiali could work with  user token  ,  func InClusterConfig()  should works with different user token?

Maybe I meet a bug...

lucasponce:
Ok, if I understand the initial scenario is that you have configured kiali to with "auth" token policy and it doesn't work as expected.

I'm going to cc my colleague @israel-hdez that worked close in this area as I think perhaps we need to collect more info to investigate this scenario.

@shalk would you mind to describe which k8s cluster, Istio version and how did you install/configure Kiali ?

Thanks

israel-hdez:
Hi @shalk 

I understand you are using _token_ authentication.

In Kiali website, under [Login Options section](https://kiali.io/documentation/getting-started/#_login_options), this paragraph is important:

`When using the token strategy in Kubernetes, Kiali assumes that all accessible namespaces are readable. Thus, for Kiali to work correctly the Service Account used to log in must have, at least, read privileges in all accessible namespaces.`

You can use any ServiceAccount you want, as long as it has the required privileges for Kiali to work correctly. If you configured Kiali to read all namespaces in your cluster, Kiali requires the `list namespaces` privilege on the ServiceAccount you are using. Else, you need the `get namespace` privilege on each accessible namespace.

So, please make sure that the ServiceAccount you want to use has the required privileges.

By default, the Kiali ServiceAccount has the needed privileges, thus you can login with it. As a starting point, you could to copy the privileges of the Kiali ServiceAccount to the other ServiceAccounts you want to use. Then, reduce privileges as needed.

shalk:
hi @israel-hdez  
Thank you for your tips,  

My kiali config is configured by istioctl default:
```
deployment:
   accessible_namespaces:['**']
```
My ServiceAcount  do not have  `list namespaces` privilege,  so i can not login.



In my scenario, every team have one namespace and one serviceaccount.

It looks like  accessible_namespaces is unchangable when kiali is running.

If i configure accessible_namespaces = `namespace-A`, team A can login kiali. But team B can not .

How can i implement multi-tanent?

By the way 
I installed kiali by istioctl.
kiali version is 1.14
istio version is 1.5.2

israel-hdez:
Hi @shalk 

> In my scenario, every team have one namespace and one serviceaccount.
> 
> It looks like accessible_namespaces is unchangable when kiali is running.
> 

Yes, this is by design. It's meant to hide namespaces from the namespaces dropdown. A good use is to hide infrastructure namespaces like the _kubernetes_ one which hosts coredns, kube-proxy and other infrastructure services.

> If i configure accessible_namespaces = `namespace-A`, team A can login kiali. But team B can not .
> 
> How can i implement multi-tanent?

Unfortunately, this is a known limitation. **Kiali uses the _list namespaces_ endpoint to validate the token**. The best I can suggest is to keep the default config of `accessible_namespaces:['**']` and add the _list namespaces_ privilege to the ServiceAccounts you need to use Kiali. Those ServiceAccounts will be able to see the available namespaces in the cluster, but they won't be able to list items that belong to the namespaces and also Kiali shouldn't let the users to inspect the namespaces.

If this does not suit your needs (i.e. is not OK for you to add the _list namespaces_ privilege)... well... I really need ideas to better validate the token used to login ;-) If you have one, let me know and I can try to implement it.

shalk:
@israel-hdez Thank you for  solve  my doubts.    I will discuss with my colleague about ` list namespaces privilege`，If i have some idea, I will  reply my word here.  Thank you for your time again.

jmazzitelli:
> It's meant to hide namespaces from the namespaces dropdown. A good use is to hide infrastructure namespaces like the _kubernetes_ one which hosts coredns, kube-proxy and other infrastructure services.

I just want to point out that I think you confused [deployment.accessible_namespaces](https://github.com/kiali/kiali/blob/v1.15.1/operator/deploy/kiali/kiali_cr.yaml#L167-L175) with the [api.namespaces.exclude](https://github.com/kiali/kiali/blob/v1.15.1/operator/deploy/kiali/kiali_cr.yaml#L95-L101) setting.

`api.namespaces.exclude` is what hides infrastructure namespaces from the namespaces dropdown.
`deployment.accessible_namespaces` is used primarily by the operator to help it determine what roles to create and which namespaces to create those roles in.

I'm not sure it matters for this particular issue, but I wanted to point that out in case it causes confusion.

jshaughn:
@israel-hdez This is interesting fallout from using listNamespaces as part of auth.  Perhaps this deserves a FAQ entry?

israel-hdez:
@jshaughn I partially agree. I'll create an issue for the FAQ entry.
I'm thinking about submitting Kiali to this: https://developers.google.com/season-of-docs. They start accepting submissions on April 13th.

Our docs on the website are not well organized. So, perhaps we need somebody to "fix" our docs.

Issue #2548, opened 2020-03-23
Missing ServiceEntries in Graph


drcrees:
**Describe the bug**
After upgrading from Kiali v1.13.0 to v1.15.0 we noticed that traffic to external services is no longer displaying in graphs.

We are using Istio's `REGISTRY_ONLY` external traffic policy, with defined service entries.  

Could this no longer be the default behavior, and we're just missing a configuration setting?

**Versions used**
Kiali: v1.15.0
Istio: 1.3.5 & 1.4.6
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9)
GCP 1.14.10-gke.17 

**To Reproduce**
1. Generate traffic to a registered external service from application in the mesh
2. Navigate to the graph in Kiali
3. See that the traffic to the external service does not show 

**Expected behavior**
After generating traffic to an external service from an application in the mesh, I would expect to see that traffic represented as a service entry in the graph like it does in v1.13.0.

jshaughn:
That is strange considering there has not been a recent change to service entry detection code.  The only change has been kiali version from 1.13 to 1.15?  And now the the edges to service entry nodes (and the nodes) are not displayed?

drcrees:
@jshaughn correct.  We have verified that the traffic is also making it to Prometheus.

The only other thing we did was disable `threescale`
```
... 

extensions:
  threescale:
    enabled: false
```

jshaughn:
@drcrees I have one thought as to what may be happening, could you share one or two prometheus time-series that previously showed as service entry traffic?

drcrees:
```
istio_tcp_sent_bytes_total{connection_security_policy="unknown",destination_app="unknown",destination_principal="unknown",destination_service="ldap.example.com",destination_service_name="ldap.example.com",destination_service_namespace="test-app",destination_version="unknown",destination_workload="unknown",destination_workload_namespace="unknown",instance="172.18.234.175:42422",job="istio-mesh",reporter="source",response_flags="-",source_app="test-app",source_principal="unknown",source_version="unknown",source_workload="test-app",source_workload_namespace="test-app"}
```

```
[...]

PR #2571, opened 2020-03-26
Relax filtering of bad telemetry

jshaughn:
We did work in #2511 to support headless services.  It allowed headless services to survive "dead" node filtering but introduced other service filtering that has turned out to be too aggressive, removing service entry telemetry when the requested host matches exactly the defined host in the service entry.

The solution here is to limit that filtering to IP addresses and to ignore qualified host names.  We've seen Istio generate bad telemetry exposing specific IPs, but not qualified hosts.  It is possible for users to define service entries using IPs but it is discouraged.  If it too becomes a problem we may have to think further, or make the filtering optional.

Fixes https://github.com/kiali/kiali/issues/2548

@jotak You should not see any differences in your Kafka/Headless env with this change.  It should continue to filter on the IP case only.


jshaughn:
Thanks @jotak .

PR #2598, opened 2020-04-01
Add Istio Components Status

xeviknal:
This PR adds a new endpoint that returns the status of each component of Istio. For each components it returns if it is a core component and whether it is Unhealthy or Not Found.

Unhealthy => The component has missing pods
Not Found => There isn't any deployment in that namespace for that component.

Healthy deployments aren't returned.

#### List of components along its severity:
**warning**: means that if a component is not found or unhealthy, the UI will show a warning symbol.
**error**: if the component is not found or unhealthy, it will show an error symbol.

monolith (demo profile from > 1.5):
```
		"istio-egressgateway":  warning,
		"istio-ingressgateway": error,
		"istiod":               error,
		"grafana":              warning,
		"istio-tracing":        warning,
		"prometheus":           error,
``` 

mixer-base installation (tipically < 1.5):

```
		"istio-citadel":          error,
		"istio-egressgateway":    warning,
		"istio-galley":           error,
		"istio-ingressgateway":   error,
		"istio-pilot":            error,
		"istio-policy":           error,
		"istio-sidecar-injector": error,
		"istio-telemetry":        error,
		"grafana":                warning,
		"istio-tracing":          warning,
		"prometheus":             error,
```

Screenshot of localhost_20001_kiali_api_istio_status (1)

This PR supports mixer-based istio installations (pilot, citadel, telemetry,...) and monolith base Istio installation (istiod, ingress, egress).

If Grafana or Tracing is disabled, the components are not considered to check the status.

Needs https://github.com/kiali/kiali-ui/pull/1733

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -26,7 +26,7 @@ type ComponentStatus struct {
 	// required: true
 	Status string `json:"status"`
 
-	// When true, the component is part of istio core. Otherwise, it is an addon
+	// When true, the component is necessary for Istio to function. Otherwise, it is an addon
 	//
 	// example:  true
 	// required: true
@@ -41,42 +41,89 @@ const (
 	NotFound  string = "NotFound"
 )
 
-// List of workloads part of a Istio deployment and if whether it is mandatory or not
-var components = map[string]bool{
-	// Core components, mandatory
-	"istio-egressgateway":  true,
-	"istio-ingressgateway": true,
-	"istiod":               true,
-	// Addon components, not mandatory
-	// Kiali not included.
-	"grafana":    false,
-	"jaeger":     false,
-	"prometheus": false,
+// List of workloads part of a Istio deployment and if whether it is mandatory or not.
+// It follows the default profile
+var components = map[string]map[string]bool{
+	"monolith": {
+		"istio-egressgateway":  false,
+		"istio-ingressgateway": true,
+		"istiod":               true,
+		"grafana":              false,
+		"istio-tracing":        false,
+		"prometheus":           true,
+	},
+	"mixer": {
+		"istio-citadel":          true,
+		"istio-egressgateway":    false,
+		"istio-galley":           true,
+		"istio-ingressgateway":   true,
+		"istio-pilot":            true,
+		"istio-policy":           true,
+		"istio-sidecar-injector": true,
+		"istio-telemetry":        true,
+		"grafana":                false,
+		"istio-tracing":          false,
+		"prometheus":             true,
+	},
 }
 
 func (iss *IstioStatusService) GetStatus() (IstioComponentStatus, error) {
-	isc := IstioComponentStatus{}
-
 	if !config.Get().ExternalServices.Istio.IstioStatusEnabled {
-		return isc, nil
+		return IstioComponentStatus{}, nil
 	}
-
-	cf := map[string]bool{}
 
 	// Fetching workloads from control plane namespace
 	ds, error := iss.k8s.GetDeployments(config.Get().IstioNamespace)
 	if error != nil {
-		return isc, error
+		return IstioComponentStatus{}, error
 	}
 
-	// Map workloads there by app name
+	arch := iss.detectArchitecture(ds)
+	if arch != "monolith" && arch != "mixer" {
+		return IstioComponentStatus{}, nil
+	}
+
+	return iss.getStatusOf(arch, ds)
+}
+
+func (iss *IstioStatusService) detectArchitecture(ds []apps_v1.Deployment) string {
+	monArch := false
+	mixArch := false
+
 	for _, d := range ds {
-		appName := d.Labels[config.Get().IstioLabels.AppLabelName]
+		appName := d.Name
 		if appName == "" {
 			continue
 		}
 
-		isCore, found := components[appName]
+		monArch = monArch || appName == "istiod"
+		mixArch = mixArch || appName == "istio-pilot"
+	}
+
+	arch := "notfound"
+	if monArch && !mixArch {
+		arch = "monolith"
+	} else if !monArch && mixArch {
+		arch = "mixer"
+	} else if monArch && mixArch {
+		arch = "multiple"
+	}
+
+	return arch
+}
+
+func (iss *IstioStatusService) getStatusOf(arch string, ds []apps_v1.Deployment) (IstioComponentStatus, error) {
+	isc := IstioComponentStatus{}
+	cf := map[string]bool{}
+
+	// Map workloads there by app name
+	for _, d := range ds {
+		appName := d.Name
+		if appName == "" {
+			continue
+		}
+
+		isCore, found := components[arch][appName]
 		if !found {
 			continue
 		}
@@ -84,17 +131,19 @@ func (iss *IstioStatusService) GetStatus() (IstioComponentStatus, error) {
 		// Component found
 		cf[appName] = true
 
-		// Check status
-		isc = append(isc, ComponentStatus{
-			Name:   appName,
-			Status: GetDeploymentStatus(d),
-			IsCore: isCore,
-		},
-		)
+		if status := GetDeploymentStatus(d); status != Healthy {
+			// Check status
+			isc = append(isc, ComponentStatus{
+				Name:   appName,
+				Status: status,
+				IsCore: isCore,
+			},
+			)
+		}
 	}
 
 	// Add missing deployments
-	for comp, isCore := range components {
+	for comp, isCore := range components[arch] {
 		if _, found := cf[comp]; !found {
 			isc = append(isc, ComponentStatus{
 				Name:   comp,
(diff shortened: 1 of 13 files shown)


business/istio_status.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+)
+
+// SvcService deals with fetching istio/kubernetes services related content and convert to kiali model
+type IstioStatusService struct {
+	k8s kubernetes.IstioClientInterface
+}
+
+type Status string
```
lucasponce, line 16: I'd really appreciate if we can unify the style of code.
This is nothing wrong, but at some point we need to try to "align" and unify otherwise code will look different.

To me, a wrapper type Status to not use a string is not justified.

Sorry, but in other parts of the code is mapped directly:

```
type Pod struct {
	Name                string            `json:"name"`
	Labels              map[string]string `json:"labels"`
	CreatedAt           string            `json:"createdAt"`
	CreatedBy           []Reference       `json:"createdBy"`
	Containers          []*ContainerInfo  `json:"containers"`
	IstioContainers     []*ContainerInfo  `json:"istioContainers"`
	IstioInitContainers []*ContainerInfo  `json:"istioInitContainers"`
	AppLabel            bool              `json:"appLabel"`
	VersionLabel        bool              `json:"versionLabel"`
	Annotations         map[string]string `json:"annotations"`
}
```

I can understand a helper type when it's a complex type (i.e. map/array to pointers) and there is a clear advantaged in the use.

On situations like these, it just a different style and it diverges the style of the code.
Should other developer refactor all Name string, to Name ComponentName ?

So, I don't enter here if this is right or wrong, but let's try to keep an eye at least in the same package and try to unify criterias.

xeviknal:
Rabased with master.

business/istio_status.go
```diff
@@ -0,0 +1,116 @@
...
+
+	// Fetching workloads from control plane namespace
+	ds, error := iss.k8s.GetDeployments(config.Get().IstioNamespace)
+	if error != nil {
+		return isc, error
+	}
+
+	// Map workloads there by app name
+	for _, d := range ds {
+		appName := ComponentName(d.Labels[config.Get().IstioLabels.AppLabelName])
```
lucasponce, line 72: These kind of mapping are not justified.
Sorry.
AppLabelName is a string.
ComponentName is a wrapper of a string.
It's not necessary.

I understand and appreciate the intention, but on this case it adds an overhead in the code.

xeviknal:
Totally agree with your comments. I guess that I wanted to use the type aliases for increase readability of data structures but at the end wasn't that necessary.
Pushed one commit removing those 3 aliases. Thus, reducing the abuse of type aliases.

business/istio_status.go
```diff
@@ -0,0 +1,118 @@
...
+	Unhealthy string = "Unhealthy"
+	NotFound  string = "NotFound"
+)
+
+// List of workloads part of a Istio deployment and if whether it is mandatory or not
+var components = map[string]bool{
+	// Core components, mandatory
+	"istio-egressgateway":  true,
+	"istio-ingressgateway": true,
+	"istiod":               true,
```
jotak, line 49: Perhaps that will be for a next PR to split the work, but I think we'll want compatibility with older Istio components. Not only for istio < 1.5 but also because Istio can still run without istiod, with old mixer & co (as I just learnt here https://github.com/kiali/kiali/issues/2669#issuecomment-616596458 )

xeviknal: Good catch. wdyt of the following behavior?

```
istiod  | pilot  | outcome
-------------------------------------
Not Found   | Not Found  | Error message: Pilot not found
Not Found   | Found      | Status for non-monolithic istio control plane stack (mixer and co)
Found       | Not Found  | Status for monolithic istio control plane stack (istiod)
Found       | Found      | Error message: Multiple Pilot installed
```

I think this will cover both scenarios. Even weird installations (two pilots) (because they are weird, right?)

jotak: Looks good! I agree that having both istiod and pilot is probably not supposed to happen and can be reported as a problem (until proven otherwise)

xeviknal: Thinking out loud. Does it make sense to show the 'Not found' component status?

Istio installation can be really diverse: ones using tracing, others don't; ones using grafana; others don't; ones egress, others don't and so on[1]. If we show the not found, the icon on the masthead might lose some meaning bc it will be always present and warning.

Therefore, what we could do is just show the status of the istio components (monolith and non-monolith ones) because we consider that users aren't deleting Istio components carelessly. 

Side effect, implementation will be easier: check status of Istio components that are deployed in the control plane namespace. If they are not present, don't show any status. Include both monolith and non-monolith components in the list.

Otherwise, this PR will be very demo profile oriented. IMO.

[1] Minimal or remote profiles are quite skinny: https://istio.io/docs/setup/additional-setup/config-profiles/

lucasponce: Just a comment.
Let's try to differentiate errors/warnings.
Error: some situation we know that is cause a problem.
Warning: something that may show something unusual i.e. tracing not enabled. We know Istio may work, but probably Kiali will miss some feature.

Perhaps we can't cover all combinations of situations on a single PR.

This work was defined to offer a first feature that can be extended with more use cases.

But no need to cover all of them, it can go in following iterations if some of them is not trivial.

jotak: About third party not found, I think we have to match that against Kiali config. If Jaeger is disabled in Kiali, we won't show the not found issue, but if it's enabled we should show it. Same for Grafana.

jotak: Actually for Jaeger and Grafana, I don't know, if you choose to remove them from the scope of this task, I think it's fine as well because we have other obvious signals / messages that comes up when they're missing. So, it's not a big value added to have them here as well.

For other components, it's different. Having them missing can be a real problem that goes unnoticed in Kiali, without those checks.

xeviknal:
@lucasponce comments approached. Ready to review.

business/istio_status.go
```diff
@@ -0,0 +1,194 @@
...
+	} else if arch == "notfound" {
+		unex, isc = true, IstioComponentStatus{
+			Message: "Istio Status disabled: Pilot not found",
+		}
+	}
+
+	return unex, isc
+}
+
+func (iss *IstioStatusService) detectArchitecture(ds []apps_v1.Deployment) string {
```
jotak, line 115: Instead of this, is it possible to use what is being done for telemetry v1: https://github.com/kiali/kiali/pull/2682 ?
Or is it different use cases that I don't see?

(I'm suggesting that they rename func `IstioSupportsCanonical` to `IsMixerEnabled`, so that's it's more appropriate here)

xeviknal: Does enabling telemetry v1 on Istio 1.5 means that also non telemetry related components (citadel, galley, injector, etc) are deployed? Which are the deployments added into the 1.5 installation to support telemetry v1?

I think that the new method from #2682 might help to add telemetry deployments to `istiod`-like installation. But I'd need to install 1.5 with that flag changed to see new deployments added (if they are).

xeviknal:
Ready to be reviewed! Thanks to all involved.

config/config.go
```diff
@@ -140,6 +140,7 @@ type TracingConfig struct {
 
 // IstioConfig describes configuration used for istio links
 type IstioConfig struct {
+	IstioStatusEnabled     bool   `yaml:"istio_status_enabled,omitempty"`
```
lucasponce, line 143: Note that this requires a change in the operator.
In the Kiali CR and in the defaults.

xeviknal: I've just created this PR into the kiali-operator to support this one.
Do I have to change any other place for the default? I've added one default at Line 329 at this same file.

kubernetes/kubernetes_service.go
```diff
@@ -356,3 +365,39 @@ func (in *IstioClient) GetSelfSubjectAccessReview(namespace, api, resourceType s
...
+
+	meshConfigYaml, ok := istioConfig.Data["mesh"]
+	log.Tracef("meshConfig: %v", meshConfigYaml)
+	if !ok {
+		log.Warningf("IsMixerDisabled: Cannot find Istio mesh configuration.")
+		return true
+	}
+
+	meshConfig := istioMeshConfig{}
+	err = yaml.Unmarshal([]byte(meshConfigYaml), &meshConfig)
```
lucasponce, line 389: Why don't use the json.Unmarshall in the ""encoding/json" library instead ?
Just learning here why two different libraries are used for same function.

xeviknal: This is code inherited from @israel-hdez from previous PR. I had to move it into this package for better testability. Didn't refactor the method itself though. And I think I prefer not to in this PR. See it here: https://github.com/kiali/kiali/pull/2598/files#diff-87fb62602e6a5ead1f80e010eb9609deL356

And besides that, the MeshConfig file is in YAML format not in JSON. So I guess JSON.Unmarshall would fail.

Issue #2556, opened 2020-03-25
figure out a good way to default the "version" setting or at least document it


jmazzitelli:
Right now the spec.version setting is not required, but its default setting is "v1.0" which is tripping people up (many people expect it to have a default of "default" so the latest version is installed, not the old v1.0 version). It used to have a default value of "default" but because of the Maistra use-case we needed to have this default to "v1.0".

Should probably figure out a way to not have to have a default value of "v1.0".

Or at minimum we should have a loud document somewhere reminding people to put "spec.version: default"

lucasponce:
In which scenario ? Using upstream opereator ?

I guess we have three main situations:
- Installing a specific Istio version, a specific Kiali version is installed with that.
- Installing Maistra, a specific version is installed.

So, using the upstream operator perhaps leaves that the default case is to upgrade an existing version, and perhaps we want to point always to our latest "stable".

jmazzitelli:
This will happen every time someone has a Kiali CR in which they do not set "spec.version". In this case, it is going to install using the v1.0 playbooks. And only in some situations do we want that (the only time we want this is if someone has Maistra 1.0 and want to automatically upgrade. In that case, Maistra 1.0 did not put "spec.version" in the kiali CR.

But I'm seeing the case that people using Kiali in other situations (upstream Istio mainly) where they are used to creating Kiali CR's without "spec.version" - when they do this now using the latest Kiali, they are getting the wrong Kiali installed.

This doesn't affect those installing Kiali using istioctl or the upstream istio helm charts because that does not involve the kiali operator.

PR #2609, opened 2020-04-03
[operator] provide a small file to be able to determine what default playbook to use

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2556

### DIFF
--- a/operator/playbooks/default-playbook.yml
+++ b/operator/playbooks/default-playbook.yml
@@ -1,0 +1,1 @@
+playbook: default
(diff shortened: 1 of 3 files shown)


jmazzitelli:
This moves the default back to the "default" playbook, to help avoid confusion of people using older Kiali CRs that do not specify "spec.version".

This allows us to easily change the default used by the operator - during builds, we can change that one-line file to indicate what the version should be. This way we are not hardcoding the default version throughout the different playbook/role files

PR #2632, opened 2020-04-08
Disable LoginToken.SigningKey validation with Anonymous auth strategy

koalaty-code:
** Describe the change **

This disables the validation for `LoginToken.SigningKey` when using the `anonymous` Authentication Strategy. Currently, the value is required and cannot equal `"kiali"` which is not relevant when using an `anonymous` strategy.

** Issue reference **

[Issue](https://github.com/kiali/kiali/issues/2548)([comment](https://github.com/kiali/kiali/issues/2548#issuecomment-610630144))

** Backwards incompatible? **

This change is **not** backwards incompatible. 

TBD

### DIFF
--- a/config/token.go
+++ b/config/token.go
@@ -38,7 +38,7 @@ func GetSigningKey() string {
 	cfg := Get()
 	signKey := cfg.LoginToken.SigningKey
 
-	if len(signKey) == 0 || signKey == "kiali" {
+	if cfg.Auth.Strategy != AuthStrategyAnonymous && (len(signKey) == 0 || signKey == "kiali") {
 		// "kiali" is a well-known signing key reported in a CVE. We ban it's usage.
 		// An empty key is also just not allowed.
 		panic("signing key for login tokens is invalid")
--- a/config/token.go
+++ b/config/token.go
@@ -38,10 +38,8 @@ func GetSigningKey() string {
 	cfg := Get()
 	signKey := cfg.LoginToken.SigningKey
 
-	if cfg.Auth.Strategy != AuthStrategyAnonymous && (len(signKey) == 0 || signKey == "kiali") {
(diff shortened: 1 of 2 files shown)


rhqci:

israel-hdez:
@koalaty-code Same conditional is in https://github.com/kiali/kiali/blob/master/kiali.go#L218-L224
It also needs to be fixed.

koalaty-code:
My in-familiarity with Go led me down the easiest path to make this change. I would like to remove the duplicated validation logic between `kiali.go` and `token.go`. To do that, I was planning on:

1. Creating a new function in `token.go`
```go
func ValidateSigningKey(signingKey string) {
	if cfg.Auth.Strategy != AuthStrategyAnonymous && (len(signingKey) == 0 || signingKey == "kiali") {
		// "kiali" is a well-known signing key reported in a CVE. We ban it's usage.
		// An empty key is also just not allowed.
		panic("signing key for login tokens is invalid")
	}
}
```
2. Update `token.go#GetSigningKey` to remove the validation and replace with a call to `ValidateSigningKey(signingkey)`
3. Update `kiali.go#validateConfig` to replace the duplicated logic with a call to `config.ValidateSigningKey`.

If you favor that approach, I can re-work those files accordingly.

israel-hdez:
@koalaty-code It is perfectly fine to remove the duplicated logic :)
I see there is a difference: one place calls `panic()` while the other returns an error. Both places have in common that the pod will die. Just test that the message is properly logged so that is visible in the pod logs when unifying with `panic` instead of returning error in one place.

israel-hdez:
@koalaty-code Thanks for the fix!

israel-hdez:
This fix will be released on Kiali v1.17

Issue #2622, opened 2020-04-07
Graph fails to show up with "Failed ID gen"


uncycler:
**Describe the bug**
Using Istio with mTLS enabled, I have a pod that post a metric on a prometheus-pushgateway located in a different namespace which is destroyed right after (it's a gitlab runner's helper). Somehow, this break Kiali graph with the following error:

```
graph.go:87] Failed ID gen: namespace=[gitlab] workload=[unknown] app=[unknown] version=[latest] service=[] graphType=[versionedApp]: goroutine 405 [running]:
runtime/debug.Stack(0xc0012e4408, 0x1240620, 0xc000959b10)
	/home/jmazzite/bin/go-install/go/src/runtime/debug/stack.go:24 +0x9d
github.com/kiali/kiali/handlers.handlePanic(0x1676940, 0xc001052460)
	/home/jmazzite/source/kiali/kiali/src/github.com/kiali/kiali/handlers/graph.go:86 +0x155
panic(0x1240620, 0xc000959b10)
	/home/jmazzite/bin/go-install/go/src/runtime/panic.go:967 +0x15d
```

The rogue prometheus query seems to be:
```
round(sum(rate(istio_requests_total{reporter="destination",source_workload="unknown",destination_workload_namespace="mynamespace"} [60s])) by (source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,request_protocol,response_code,grpc_response_status,response_flags),0.001)
```
I guess prometheus don't have time to scrap this pod to get full information.

**Versions used**
Kiali: 1.15.1
Istio: 1.5.1
Kubernetes flavour and version: AWS EKS 1.15

**To Reproduce**
1. Enable mutual TLS
2. Create two pod with the Istio sidecar in different namespaces
3. Do a post request from one pod to another and delete it.
4. Try to load kiali graph.

**Expected behavior**
Graph should just ignore those and show up.

jshaughn:
@uncycler You are definitely hitting unexpected/incomplete telemetry being generated by Istio.  This is very likely a new corner-case in Istio's Telemetry V2, the new default for Istio 1.5.    The query you posted is fine, it's more rogue data than a rogue query.

[...]

PR #2636, opened 2020-04-08
Protect against bad source telemetry

jshaughn:
Fixes https://github.com/kiali/kiali/issues/2622

After discussing with the community member in the issue we decided that the problematic telemetry is "bad" telemetry, likely due to a telemetry V2 corner case with reporting or scraping.  I asked in the istio Telemetry working group and they also said it seemed unintended to have source workload set, but with no other source information.

So, this PR adds a filter for telemetry with that signature.

@israel-hdez This is probably just a visual review of the code as we can't easily replicate the problem.  @mattmahoneyrh, the validation is likely just to ensure that bookinfo or any other traffic remains the same as before the PR (basically show that the new filtering doesn't lose any of the existing traffic we expect).

### DIFF
--- a/graph/telemetry/istio/util/util.go
+++ b/graph/telemetry/istio/util/util.go
@@ -62,12 +62,20 @@ func HandleResponseCode(protocol, httpResponseCode string, grpcResponseStatusOk 
 	return grpcResponseStatus
 }
 
-// IsBadTelemetry tests for known issues in generated telemetry given indicative label values.
+// IsBadSourceTelemetry tests for known issues in generated telemetry given indicative label values.
+// 1) source namespace is provided without workload or app
+// 2) no more conditions known
+func IsBadSourceTelemetry(ns, wl, app string) bool {
+	// case1
+	return graph.IsOK(ns) && !graph.IsOK(wl) && !graph.IsOK(app)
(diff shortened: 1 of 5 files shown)


graph/telemetry/istio/util/util.go
```diff
@@ -62,12 +62,20 @@ func HandleResponseCode(protocol, httpResponseCode string, grpcResponseStatusOk
 	return grpcResponseStatus
 }
 
-// IsBadTelemetry tests for known issues in generated telemetry given indicative label values.
+// IsBadSourceTelemetry tests for known issues in generated telemetry given indicative label values.
+// 1) source namespace is provided without workload or app
```
israel-hdez, line 66: Should the comment say "**nor** app"?

jshaughn: You are right, the wording here is a bit vague.  I won't update it for the PR but I will update it in passing.  Usually "nor" is used with "neither", so something like, "source namespace is provided but with neither workload nor app"

jshaughn: actually, I will update it to see if I can snyk to run again, Not sure what the problem is there...

Issue #2627, opened 2020-04-07
[operator] set proper relatedImages in metadata


jmazzitelli:
To support disconnected use-cases, our metadata sets some environment variables telling the operator what images to use. We also need to set relatedImages to the images we need.

We should use fieldRefs to refer back to the relatedImages fields so there is less changes needed to be made in the CSV yaml.

https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/contributors/design-proposals/related-images.md

So we will need to do something like this in our CSVs:

```
spec:
  relatedImages:
  - name: kiali-v1.0
    annotation: kiali-v1.0
    required: true
    image: registry.redhat.io/openshift-service-mesh/kiali-rhel7:1.0.12
  - name: kiali-v1.12
    annotation: kiali-v1.12
    required: true
    image: registry.redhat.io/openshift-service-mesh/kiali-rhel7:1.12.7
...
                - name: KIALI_IMAGE_default
                  valueFrom:
                    fieldRef:
                      fieldPath: metadata.annotations['olm.relatedImage.kiali-v1.0']
                - name: KIALI_IMAGE_v1_0
                  valueFrom:
                    fieldRef:
                      fieldPath: metadata.annotations['olm.relatedImage.kiali-v1.0']
                - name: KIALI_IMAGE_v1_12
                  valueFrom:
                    fieldRef:
                      fieldPath: metadata.annotations['olm.relatedImage.kiali-v1.12']
```

jmazzitelli:
I was told that the annotation feature is not implemented yet. So for now, we have to hardcode the image names in the env var definition themselves.

PR #2639, opened 2020-04-09
[operator] additional changes needed to support disconnected use-case

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/2627

These are changes to the OLM metadata required for disconnected use-case to work.

Do not merge this yet. My current testing on OpenShift 4.3 shows that the OLM annotations are not getting created. Need to verify that my testing was correct and if it was we need to figure out why those annotations are missing. I followed this doc: https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/contributors/design-proposals/related-images.md#clusterserviceversion-spec-changes

### DIFF
--- a/operator/manifests/kiali-ossm/1.12.8/kiali.crd.yaml
+++ b/operator/manifests/kiali-ossm/1.12.8/kiali.crd.yaml
@@ -1,0 +1,21 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: kialis.kiali.io
+  labels:
+    app: kiali-operator
+spec:
+  group: kiali.io
+  names:
+    kind: Kiali
+    listKind: KialiList
+    plural: kialis
+    singular: kiali
+  scope: Namespaced
+  subresources:
+    status: {}
+  version: v1alpha1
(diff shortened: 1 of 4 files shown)


jmazzitelli:
We can't use the fieldRefs in the env var definitions yet because OLM isn't creating those annotations yet. We can add those in later. For now, we just have to provide `relatedImages`.

When OLM does implement the annotations, we can change the env var definitions to look like this:

```
   - name: KIALI_IMAGE_default
     valueFrom:
       fieldRef:
         fieldPath: metadata.annotations['olm.relatedImage.kiali-v1.0']
   - name: KIALI_IMAGE_v1_0
     valueFrom:
       fieldRef:
         fieldPath: metadata.annotations['olm.relatedImage.kiali-v1.12']
   - name: KIALI_IMAGE_v1_12
     valueFrom:
       fieldRef:
         fieldPath: metadata.annotations['olm.relatedImage.kiali-v1.12']
```

jmazzitelli:
this needs to create and modify a new 1.12.8 rather than update the existing .7

jmazzitelli:
This PR now adds 1.12.8 with the necessary relatedImages.

Issue #2653, opened 2020-04-14
create new kiali/kiali-operator github repo 


jmazzitelli:
Create new repo and prepopulate it with kiali/kiali's operator code. Make sure to retain git history.

epic: https://github.com/kiali/kiali/issues/2618

jmazzitelli:
This is considered complete once these PRs are merged:

* kiali PR: https://github.com/kiali/kiali/pull/2659
* operator PR: https://github.com/kiali/kiali-operator/pull/1

PR #2659, opened 2020-04-14
Remove and ignore the operator directory, and build operator out of new repo

jmazzitelli:
This removes the operator code - it is now located at https://github.com/kiali/kiali-operator

fixes: https://github.com/kiali/kiali/issues/2653

You must merge this after this operator PR is merged: https://github.com/kiali/kiali-operator/pull/1


jmazzitelli:
@israel-hdez take a look at the last commit in this PR - the commit with the message "be able to build the operator via jenkins" (currently that commit is https://github.com/kiali/kiali/pull/2659/commits/450123e4373169ad2614e7a25605a43a53c46a25)

It modifies the Jenkins stuff so that we can build the operator from the new repo but still use the same Jenkins that is set up in this repo. This PR has a `Makefile.operator.jenkins` in `deploy/jenkins-ci` which calls the `Makefile` that is in the kiali-operator PR here: https://github.com/kiali/kiali-operator/pull/1/files#diff-b67911656ef5d18c4ae36cb6741b7965

I do not know if this is the right thing to do, but I think this should work (I haven't tested).

The idea is that you should be able to build the operator separately, without having to build the UI or backend. You can also build the operator with everything, too, but this change lets you skip the UI and backend builds and just release the operator.

Again, I don't know if this is good, or if we want to create a new Jenkins integration inside the kiali-operator (that would mean moving over the `Makefile.operator.jenkins` file to kiali-operator repo and probably copying the `deploy/jenkins-ci` files over, too, since we need many of the same things in there). But, rather than duplicate code and bloat the kiali-operator repo with stuff we already have here in kiali/kiali, I just tried to do it this way (share the Jenkins job we already have). If, for some reason, we should not do it this way, then let's figure out how to get Jenkins to build directly out of kiali-operator repo to get a kiali-operator release built and pushed to quay.

Note also we currently don't have any commit hooks in kiali-operator (so I do not think this will be able to build "latest" operator images). I don't know how we do it today, but I thought every commit causes Jenkins to build and release with the "latest" tag.

jmazzitelli:
I just tested the Jenkins infrastructure and was able to build snapshot and minor builds of backend and operator (skipping UI). I then tried to build a minor release of JUST the operator (skipping the UI and backend) and that worked as well.

jmazzitelli:
> Lots of files

all of them (except maybe 10) are all deletions :) this is the `operator/` directory getting removed.

kenske:
:/ this broke the installation instructions in the docs:

```
bash <(curl -L https://git.io/getLatestKialiOperator) --operator-install-kiali false
```
https://kiali.io/documentation/getting-started/#_advanced_install_operator_only

jmazzitelli:
> :/ this broke the installation instructions in the docs:
> 
> ```
> bash <(curl -L https://git.io/getLatestKialiOperator) --operator-install-kiali false
> ```
> https://kiali.io/documentation/getting-started/#_advanced_install_operator_only

There have been various announcements about this earlier today - on #kiali room on Istio slack, on kiali-dev on googlegroups, and kiali-users on googlegroups:

* https://istio.slack.com/archives/CJVRXB09F/p1587147486030400
* https://groups.google.com/d/msg/kiali-dev/0S3-AfIW_E0/zVY0sz0WBAAJ
* https://groups.google.com/d/msg/kiali-users/mmOYH7fOVns/20yJSkWyBgAJ

And if you look at the link you posted:

https://kiali.io/documentation/getting-started/#_advanced_install_operator_only

It does not say to use the `git.io` link. It specifies `kiali.io` now (https://kiali.io/getLatestKiaiOperator)... the doc states:

```
  bash <(curl -L https://kiali.io/getLatestKialiOperator) --operator-install-kiali false
```

PR #2662, opened 2020-04-15
Refresh namespace level cache

kinsolee:
This PR adds a method `RefreshNamespace` in `KialiCache` providing a way to clear the cache only in a specific namespace.

When updating istio object, caches from all namespaces will be refreshed. It takes 10s+ to finish this job when the number of namespaces is more than 50.

This PR is to solve this problem as it is not necessary to refresh all caches.

### DIFF
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -146,7 +146,7 @@ func (c *kialiCacheImpl) createCache(namespace string) bool {
 	c.createIstioInformers(namespace, &informer)
 	c.nsCache[namespace] = informer
 
-	if _, exist := c.stopChan[namespace]; !exist {
+	if nsChan, exist := c.stopChan[namespace]; !exist || nsChan == nil {
 		c.stopChan[namespace] = make(chan struct{})
 	}
 
@@ -194,6 +194,12 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
 
 // RefreshNamespace will delete the specific namespace's cache and create a new one.
 func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
+	if nsChan, exist := c.stopChan[namespace]; exist {
+		if nsChan != nil {
+			close(nsChan)
+			nsChan = nil //nolint
+		}
+	}
 	defer c.cacheLock.Unlock()
 	c.cacheLock.Lock()
 	delete(c.nsCache, namespace)
@@ -203,8 +209,10 @@ func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
 func (c *kialiCacheImpl) Stop() {
 	log.Infof("Stopping Kiali Cache")
 	for _, nsChan := range c.stopChan {
-		close(nsChan)
-		nsChan = nil
+		if nsChan != nil {
+			close(nsChan)
+			nsChan = nil //nolint
+		}
 	}
 	defer c.cacheLock.Unlock()
 	c.cacheLock.Lock()
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -25,6 +25,9 @@ type (
 		// Control methods
 		// Check if a namespace is listed to be cached; if yes, creates a cache for that namespace
 		CheckNamespace(namespace string) bool
+
+		// Clear a namespace's cache
+		RefreshNamespace(namespace string)
 		// Stop all caches
 		Stop()
 
@@ -176,6 +179,14 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
 		return c.createCache(namespace)
 	}
 	return c.isKubernetesSynced(namespace) && c.isIstioSynced(namespace)
+}
+
+// RefreshNamespace will delete the specific namespace's cache and create a new one.
+func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
+	defer c.cacheLock.Unlock()
+	c.cacheLock.Lock()
+	delete(c.nsCache, namespace)
+	c.createCache(namespace)
 }
 
 func (c *kialiCacheImpl) Stop() {
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -53,7 +53,7 @@ type (
 		refreshDuration        time.Duration
 		cacheNamespaces        []string
 		cacheIstioTypes        map[string]bool
-		stopChan               chan struct{}
+		stopChan               map[string]chan struct{}
 		nsCache                map[string]typeCache
 		cacheLock              sync.Mutex
 		tokenLock              sync.RWMutex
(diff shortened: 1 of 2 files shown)


rhqci:

lucasponce:
@kinsolee thanks for your proposal.
I put myself to fetch this PR locally and review any potential side effect.

Thanks !

lucasponce:
I think this proposal may have a leak. Let me know if you see the same.

If you look to the createCache()

```
func (c *kialiCacheImpl) createCache(namespace string) bool {
	if _, exist := c.nsCache[namespace]; exist {
		return true
	}
	informer := make(typeCache)
	c.createKubernetesInformers(namespace, &informer)
	c.createIstioInformers(namespace, &informer)
	c.nsCache[namespace] = informer

	go func() {
		for _, informer := range c.nsCache[namespace] {
			go informer.Run(c.stopChan)
		}
		<-c.stopChan
		log.Infof("Kiali cache for [namespace: %s] stopped", namespace)
	}()

	log.Infof("Waiting for Kiali cache for [namespace: %s] to sync", namespace)
	isSynced := func() bool {
		hasSynced := true
		for _, informer := range c.nsCache[namespace] {
			hasSynced = hasSynced && informer.HasSynced()
		}
		return hasSynced
	}
	if synced := cache.WaitForCacheSync(c.stopChan, isSynced); !synced {
		c.stopChan <- struct{}{}
		log.Errorf("Kiali cache for [namespace: %s] sync failure", namespace)
		return false
	}

	return true
}
```

You'd see there is a goroutine running the informer and waiting for c.stopChan signal.

So, in your proposal

```
func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
	defer c.cacheLock.Unlock()
	c.cacheLock.Lock()
	delete(c.nsCache, namespace)
	c.createCache(namespace)
}
```

When deleting the cache, it doesn't stop the goroutine, so at least it would leak in the process on deletion, not checked if some race could happen to have two informers on the same cache too.

So, the proposal doesn't look bad, but it would need a refactoring to have a stopChan per namespace and sync those signals.

In that case perhaps the refresh would work safely.

Let me know your opinion on that.

kinsolee:
@lucasponce  You are right! I made the changes, now every namespace has a `stopChan` to stop the goroutine.

kubernetes/cache/cache.go
```diff
@@ -178,11 +192,27 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
 	return c.isKubernetesSynced(namespace) && c.isIstioSynced(namespace)
 }
 
+// RefreshNamespace will delete the specific namespace's cache and create a new one.
+func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
+	if nsChan, exist := c.stopChan[namespace]; exist {
+		if nsChan != nil {
+			close(nsChan)
+			nsChan = nil //nolint
```
lucasponce, line 200: I think this is not going to delete the map entry.
I would require a 
```
delete(c.stopChan, namespace)
```

kubernetes/cache/cache.go
```diff
@@ -178,11 +192,27 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
...
+
 func (c *kialiCacheImpl) Stop() {
 	log.Infof("Stopping Kiali Cache")
-	if c.stopChan != nil {
-		close(c.stopChan)
-		c.stopChan = nil
+	for _, nsChan := range c.stopChan {
+		if nsChan != nil {
+			close(nsChan)
+			nsChan = nil //nolint
```
lucasponce, line 214: Same here, it would need to get the key to delete entry from map:
```
	for namespace, nsChan := range c.stopChan {
		if nsChan != nil {
			close(nsChan)
			delete(c.stopChan, namespace)
		}
	}
```

kubernetes/cache/cache.go
```diff
@@ -178,11 +192,27 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
 	return c.isKubernetesSynced(namespace) && c.isIstioSynced(namespace)
 }
 
+// RefreshNamespace will delete the specific namespace's cache and create a new one.
+func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
+	if nsChan, exist := c.stopChan[namespace]; exist {
```
lucasponce, line 197: I think the close and deletion of the stopChannel should be protected by the lock.

kubernetes/cache/cache.go
```diff
@@ -178,11 +192,27 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
...
+	delete(c.nsCache, namespace)
+	c.createCache(namespace)
+}
+
 func (c *kialiCacheImpl) Stop() {
 	log.Infof("Stopping Kiali Cache")
-	if c.stopChan != nil {
-		close(c.stopChan)
-		c.stopChan = nil
+	for _, nsChan := range c.stopChan {
```
lucasponce, line 211: As similar as the other comment, I think that the stop channel should be protected by the lock.

kubernetes/cache/cache.go
```diff
@@ -136,11 +146,15 @@ func (c *kialiCacheImpl) createCache(namespace string) bool {
 	c.createIstioInformers(namespace, &informer)
 	c.nsCache[namespace] = informer
 
+	if nsChan, exist := c.stopChan[namespace]; !exist || nsChan == nil {
```
lucasponce, line 149: I think that if the entry has been removed correctly using the delete() invokation, there is no need to check if nsChan is nil here

kubernetes/cache/cache.go
kinsolee: I think there is also no need to add if nsChan here.

kinsolee:
I tested to update istio objects and stop terminal, there is no panic after changing the code to delete entry from map. It seems fine this time. Please review again and thank you for your patience.

`I0417 11:01:08.172403   30993 cache.go:158] Kiali cache for [namespace: bookinfo] stopped
I0417 11:01:08.473183   30993 cache.go:174] Kiali cache for [namespace: bookinfo] started
I0417 11:01:08.473226   30993 istio_config.go:343] AUDIT User [token] Msg [UPDATE on Namespace: bookinfo Type: destinationrules Subtype:  Name: reviews Patch: {"metadata":{"namespace":"bookinfo","name":"reviews","labels":{"kiali_wizard":"weighted_routing"}},"spec":{"host":"reviews.bookinfo.svc.cluster.local","subsets":[{"name":"v1","labels":{"version":"v1"}},{"name":"v2","labels":{"version":"v2"}},{"name":"v3","labels":{"version":"v3"}}],"trafficPolicy":null}}]
I0417 11:01:08.473291   30993 cache.go:161] Waiting for Kiali cache for [namespace: bookinfo] to sync
I0417 11:01:08.473339   30993 cache.go:158] Kiali cache for [namespace: bookinfo] stopped
I0417 11:01:11.273513   30993 cache.go:174] Kiali cache for [namespace: bookinfo] started
I0417 11:01:11.273557   30993 istio_config.go:343] AUDIT User [token] Msg [UPDATE on Namespace: bookinfo Type: virtualservices Subtype:  Name: reviews Patch: {"metadata":{"namespace":"bookinfo","name":"reviews","labels":{"kiali_wizard":"weighted_routing"}},"spec":{"http":[{"route":[{"destination":{"host":"reviews.bookinfo.svc.cluster.local","subset":"v1"},"weight":79},{"destination":{"host":"reviews.bookinfo.svc.cluster.local","subset":"v2"},"weight":21},{"destination":{"host":"reviews.bookinfo.svc.cluster.local","subset":"v3"},"weight":0}]}],"hosts":["reviews.bookinfo.svc.cluster.local"],"gateways":null}}]
`

lucasponce:
Latest changes looks ok.
Thanks @kinsolee.

I'm adding @hhovsepy as a reviewer for final check.
Hayk, I've reviewed locally and I haven't seen any regression or side effect, but if you could test it and perform more testing we can definitely merge it.

The focus of testing could be if after a modification or delete of a Istio config from Kiali, if there is any side effect in other parts of the app due namespace cache invalidation.

Thanks

PR #2671, opened 2020-04-16
Added get permissions method

pbajjuri20:


rhqci:

jmazzitelli:
Make sure to squash this - we don't want 8 commits in the git history that only end up changing 3 lines :)

Issue #2669, opened 2020-04-16
Kiali fails if Telemetry V1 is used with Istio 1.5+


jshaughn:
Istio 1.5 introduced istiod and Telemetry V2 (aka "mixerless" telemetry).  Kiali supports V2 telemetry for Istio v1.5.  Kiali supports V1 or V2 telemetry in v1.4.  But it no longer supports V1 telemetry for Istio 1.5+ users, and this is a problem as V1 telemetry, no longer the default, is still configurable and potentially necessary for users using helm install (as I understand things).

The reason for this is that we based on a version check of istio we now default to using the new "canonical" fields in the metrics, but they don't exist in Telemetry V1.  We need to improve our check to be based on the Telemetry version, not the product version.  Or, make this configurable in Kiali (i.e. set Telemetry version in CR, defaulting the correct way based on istio version).

cc @seeruk

jshaughn:
The problem here relates back to the work to support the new "canonical" fields in V2 telemetry, introduced in v1.5 (see https://github.com/kiali/kiali/issues/2276).   To support these fields we needed a way to know they were supported.   There is no easy way to query Istio API to know which version of telemetry is being run.  So, the solution was to query the Istio version via the API, and if the version is >= 1.5 then use the canonical fields.  The problem results from the fact that Istio did not provide backward compatibility for V1 telemetry, the canonical fields don't exist.

Although not the default, V1 telemetry can be configured in V1.5 (scheduled to be fully removed in Istio 1.7).  So, the version check is insufficient.  It means only that the canonical fields will exist if V2 telemetry is being used.

As mentioned above, there is no way [afaik] to query Istio to explicitly determine the telemetry version.  As such I see two ways of doing this, we could query prometheus for a sample time-series and check for the presence of the canonical fields.  Or, we could add Kiali config to specify the telemetry version (it would default to V1 for istio < v1.5, otherwise V2.

I think for cleanliness I'm leaning towards the config change.  Most users would use the default.  The time-series approach requires a special prom-query, and may not be deterministic if there is not yet any traffic, or may have a corner case in an upgrade scenario.

israel-hdez:
@jshaughn I haven't moved to Istio 1.5, so I cannot see how telemetry v2 is setup.
[...]

PR #2682, opened 2020-04-21
Add compatibility for Telemetry V1 on Istio 1.5

israel-hdez:
IstioSupportsCanonical function is updated to read Istio's config map and check a flag that dictates if telemetry through mixer is enabled (aka Telemetry V1). If flag is true, TelemetryV2 is active.

IstioSupportsCanonical is using Kiali's Service Account always rather than the account associated with the logged in user. This assumes that Kiali's ServiceAccount has privileges to read Istio's namespace, in contrast to reduced privileges when using authentication methods that involve the RBAC system. It should be so-so safe to do this, because Kiali ServiceAccount tends to always have privileges in Istio system namespace (at least, this is the default when first installing Kiali).

Fixes kiali/kiali#2669

### DIFF
--- a/status/versions.go
+++ b/status/versions.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-version"
+	"gopkg.in/yaml.v2"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kversion "k8s.io/apimachinery/pkg/version"
 	kube "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
@@ -20,7 +22,13 @@ import (
 	"github.com/kiali/kiali/util/httputil"
 )
 
+const ISTIO_CONFIGMAP_NAME = "istio"
+
 type externalService func() (*ExternalServiceInfo, error)
+
+type istioMeshConfig struct {
+	DisableMixerHttpReports bool `yaml:"disableMixerHttpReports,omitempty"`
+}
 
 var (
 	// Example Maistra product version is:
@@ -312,18 +320,52 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
 // did that that version change will matter, and the kiali pod could be bounced as a workaround.
 var istioSupportsCanonical *bool
 
-// IstioSupportsCanonical returns true if Istio version can be determined and is >= 1.5.
-// TODO: This test can be removed when Kiali stops supporting any Istio versions < 1.5
+// IstioSupportsCanonical returns true if Telemetry V2 is enabled
+// TODO: This test can be removed when Kiali stops supporting Istio versions with Mixer Telemetry
 func IstioSupportsCanonical() bool {
 	if istioSupportsCanonical != nil {
 		return *istioSupportsCanonical
 	}
 
-	istioVersion, err := istioVersion()
+	k8sConfig, err := kubernetes.ConfigClient()
 	if err != nil {
+		log.Warningf("IstioSupportsCanonical: Cannot create config structure Kubernetes Client.")
 		return false
 	}
-	valid := validateVersion(">= 1.5", istioVersion.Version)
-	istioSupportsCanonical = &valid
+
+	k8s, err := kube.NewForConfig(k8sConfig)
+	if err != nil {
+		log.Warningf("IstioSupportsCanonical: Cannot create Kubernetes Client.")
+		return false
+	}
+
+	cfg := config.Get()
+	istioConfig, err := k8s.CoreV1().ConfigMaps(cfg.IstioNamespace).Get(ISTIO_CONFIGMAP_NAME, meta_v1.GetOptions{})
+	if err != nil {
+		log.Warningf("IstioSupportsCanonical: Cannot retrieve Istio ConfigMap.")
+		return false
+	}
+
+	meshConfigYaml, ok := istioConfig.Data["mesh"]
+	log.Tracef("meshConfig: %v", meshConfigYaml)
+	if !ok {
+		log.Warningf("IstioSupportsCanonical: Cannot find Istio mesh configuration.")
+		return false
+	}
+
+	meshConfig := istioMeshConfig{}
+	err = yaml.Unmarshal([]byte(meshConfigYaml), &meshConfig)
+	if err != nil {
+		log.Warningf("IstioSupportsCanonical: Cannot read Istio mesh configuration.")
+		return false
+	}
+
+	log.Infof("IstioSupportsCanonical: %t", meshConfig.DisableMixerHttpReports)
+
+	// References:
+	//   * https://github.com/istio/api/pull/1112
+	//   * https://github.com/istio/istio/pull/17695
+	//   * https://github.com/istio/istio/issues/15935
+	istioSupportsCanonical = &meshConfig.DisableMixerHttpReports
 	return *istioSupportsCanonical
 }


israel-hdez:
@seeruk It will be great if you can try this fix.

jotak:
Isn't it possible to coordinate with @xeviknal ? Or there is a good reason to end up with two concurrent algorithms to detect the same thing?

seeruk:
Happy to test, but will hold off for the time being. Just gone through the steps to build the Kiali image, so should be able to give it a shot once it's ready.

status/versions.go
```diff
@@ -312,18 +320,52 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
 // did that that version change will matter, and the kiali pod could be bounced as a workaround.
 var istioSupportsCanonical *bool
 
-// IstioSupportsCanonical returns true if Istio version can be determined and is >= 1.5.
-// TODO: This test can be removed when Kiali stops supporting any Istio versions < 1.5
+// IstioSupportsCanonical returns true if Telemetry V2 is enabled
+// TODO: This test can be removed when Kiali stops supporting Istio versions with Mixer Telemetry
 func IstioSupportsCanonical() bool {
```
jotak, line 325: Perhaps this function can be renamed in something like `IsMixerEnabled`, to be more relevant for other uses.

israel-hdez:
> Isn't it possible to coordinate with @xeviknal ? Or there is a good reason to end up with two concurrent algorithms to detect the same thing?

I looked at @xeviknal PR #2598 and I though it was more like for finding monolith vs non-monolith and finding components status. So, I ended up with this, which I find more dependable with regards to the active Telemetry version. This seems to work when installing Istio 1.5 vía istioctl, helm and also seems to work when installing Istio 1.4.

But this is just covering the version of Telemetry **supposed** to be active. It's not checking if the right Istio services are up like in @xeviknal PR. I think both codes can compliment each other. @xeviknal could use this method to check with more precision which deployments should be up (or not).

jotak:
@israel-hdez yes, I didn't mean this was entirely overlapping, but the detection of istio-kind of installation bit is overlapping I think (I can be wrong).
Knowing if istio is monolith or not, and knowing if it has telemetry v1 or not, isn't actually the same thing?

I agree that he could use your method, just asking for coordination to know which method to use.

jshaughn:
> But this is just covering the version of Telemetry supposed to be active. It's not checking if the right Istio services are up like in @xeviknal PR. I think both codes can compliment each other.

I agree with this, I think @israel-hdez has come up with the correct approach for the definitive way of making the the V1 vs V2 telemetry determination.  Using runtime component status is perhaps less deterministic.  Instead, as suggested, this could indicate which components should or should not be running.

Renaming the method to IsMixerEnabled() is a good idea.

xeviknal:
Could you guys tell me how to enable telemetry v1 on Istio 1.5? Is it something done at installation time? Or can I just change that field on configmap?
Second, having the telemetry v1 on istio 1.5 implies that other components like galley, citadel, and so on are installed?

seeruk:
Install Istio via Helm instead of using Istioctl - then you'll get Telemetry V1, and the rest of the old stuff.

israel-hdez:
@xeviknal If you are using Istio 1.5 and installing via istioctl or via the hack script, pass these additional flags (accepted by both istioctl and the hack script): `--set values.telemetry.v1.enabled=true --set values.telemetry.v2.enabled=false --set components.citadel.enabled=true --set components.telemetry.enabled=true`

I'm writing the flags by memory, so if some are wrong, I think you get the idea :)

(Taken from https://istio.io/docs/tasks/observability/mixer/metrics/collecting-metrics/#before-you-begin)

status/versions.go
```diff
@@ -312,18 +320,52 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
+	}
+
+	k8s, err := kube.NewForConfig(k8sConfig)
+	if err != nil {
+		log.Warningf("IstioSupportsCanonical: Cannot create Kubernetes Client.")
+		return false
+	}
+
+	cfg := config.Get()
+	istioConfig, err := k8s.CoreV1().ConfigMaps(cfg.IstioNamespace).Get(ISTIO_CONFIGMAP_NAME, meta_v1.GetOptions{})
```
xeviknal, line 343: Does it make sense to move this function into the KubernetesClient instead? This way all calls to k8s are group into the same client. Plus, It also helps testing this feature.
Maybe adding a GetConfigMap(name, namespace string).
wdyt?

jshaughn:
@xeviknal So today I asked the guys on the telemetry wg call what to expect about the services that should be present in both upgrade and native 1.5 installs, when mixer is either carried forward in the update or configured for 1.5, and they actually could not answer.  There have been so many install options and update scenarios that they have lost track.  So, in short we need to experiment and figure it out ourselves.  That may just mean we assume all present istio components are desired.   Also, I learned that Mixer deprecation may last through 1.7 and be removed in 1.8.  They really are not very interested in mixer at this point and are working to close all [feasible] feature gaps for 1.7.

status/versions.go
jshaughn: I like the idea of a IstioMixerEnabled() but it probably should be in addition to IstioCanonicalSupported() because it's possible in versions < 1.5 to run telem V2 (experimental) but without canonical fields, which were introduced in 1.5.  In short the test for canonical is (!mixer && version >= 1.5).

status/versions.go
```diff
@@ -312,18 +320,52 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
+		return false
+	}
+
+	meshConfig := istioMeshConfig{}
+	err = yaml.Unmarshal([]byte(meshConfigYaml), &meshConfig)
 	if err != nil {
+		log.Warningf("IstioSupportsCanonical: Cannot read Istio mesh configuration.")
 		return false
 	}
-	valid := validateVersion(">= 1.5", istioVersion.Version)
```
jshaughn, line 326: So we still need to have version >= 1.5 to support canonical.

israel-hdez:
@jshaughn So, I renamed as `IsMixerDisabled` which is actually a new function. Then, added back the old `IstioSupportsCanonical` but renamed as `AreCanonicalMetricsAvailable` which will take into account both the Istio version and the mixer enabled/disabled flag to determine if metrics should be queried using canonical labels.

I hope this is now right.

status/versions.go
israel-hdez: I moved to the k8s client :)

xeviknal:
This LGTM. I prepared an integration with this for https://github.com/kiali/kiali/pull/2598. Once this is merged, I'll proceed with changes.

status/versions.go
```diff
@@ -308,22 +315,83 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
+// TODO: This test can be removed when Kiali stops supporting Istio versions with Mixer Telemetry
+func IsMixerDisabled() bool {
+	if isMixerDisabled != nil {
+		return *isMixerDisabled
+	}
+
+	k8sConfig, err := kubernetes.ConfigClient()
+	if err != nil {
+		log.Warningf("IsMixerDisabled: Cannot create config structure Kubernetes Client.")
+		return false
```
jshaughn, line 332: I think this should probably return true.  Going forward more users will have Telem V2, I think.  Honestly there isn't really a good default, if we can't determine this it's a pretty big problem.

status/versions.go
```diff
@@ -308,22 +315,83 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
 
-// IstioSupportsCanonical returns true if Istio version can be determined and is >= 1.5.
-// TODO: This test can be removed when Kiali stops supporting any Istio versions < 1.5
-func IstioSupportsCanonical() bool {
-	if istioSupportsCanonical != nil {
-		return *istioSupportsCanonical
+	k8s, err := kubernetes.NewClientFromConfig(k8sConfig)
+	if err != nil {
+		log.Warningf("IsMixerDisabled: Cannot create Kubernetes Client.")
+		return false
```
jshaughn, line 338: Again, I think true is the better default going forward.

status/versions.go
```diff
@@ -308,22 +315,83 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
 	}
-	valid := validateVersion(">= 1.5", istioVersion.Version)
-	istioSupportsCanonical = &valid
-	return *istioSupportsCanonical
+
+	meshConfigYaml, ok := istioConfig.Data["mesh"]
+	log.Tracef("meshConfig: %v", meshConfigYaml)
+	if !ok {
+		log.Warningf("IsMixerDisabled: Cannot find Istio mesh configuration.")
+		return false
```
jshaughn, line 352: again

status/versions.go
```diff
@@ -308,22 +315,83 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
+	}
+
+	meshConfig := istioMeshConfig{}
+	err = yaml.Unmarshal([]byte(meshConfigYaml), &meshConfig)
+	if err != nil {
+		log.Warningf("IsMixerDisabled: Cannot read Istio mesh configuration.")
+		return false
```
jshaughn, line 359: again

status/versions.go
```diff
@@ -308,22 +315,83 @@ func kubernetesVersion() (*ExternalServiceInfo, error) {
...
+
+// AreCanonicalMetricsAvailable returns true if canonical labels are present in Istio Telemetry.
+// TODO: This test can be removed when Kiali stops supporting Istio versions with Mixer Telemetry
+func AreCanonicalMetricsAvailable() bool { // AreCanonicalMetricsAvailable() bool {
+	if istioSupportsCanonical == nil {
+		// First, check Istio version because canonical labels were first introduced in Istio v1.5. Prior
+		// Istio versions won't have canonical labels in metrics regardless of active Telemetry version.
+		istioVersion, err := istioVersion()
+		if err != nil {
+			return false
```
jshaughn, line 384: it may seem countr-intuitive given my other comments but I think this is still the correct default.  If we don't know for sure that canonical is supported then we should fall back to app/version because they will exist in older installs and will still be there (deprecated) for several more versions.

jshaughn, line 384: it may seem counter-intuitive given my other comments but I think this is still the correct default.  If we don't know for sure that canonical is supported then we should fall back to app/version because they will exist in older installs and will still be there (deprecated) for several more versions.

jshaughn, line 384: it may seem counter-intuitive given my other comments but I think this is still the correct default.  If we don't know for sure that canonical is supported then we should fall back to app/version because they will exist in older installs and will still be there (deprecated) for several more versions.

pbajjuri20:
Tested and Passed on Kiali UI 1.17.0 (047bab80842dae0df4711a4e0f0e7deab4c4e61f)

pbajjuri20:
Tested and Passed without any errors.

status/versions.go
israel-hdez: I was keeping false as default, just because it is the "default" in Istio side at least when speaking about how that flag is defined their code. But the real "default" when installing Istio 1.5 is "true" and Kiali is focusing support on the latest Istio version. So, I agree this should be `true`.

israel-hdez:
Thanks all for the reviews & comments!
Merging...

Issue #2623, opened 2020-04-07
Links to Grafana should allow base URL with params


jotak:
In Grafana config, base URL must be set as something like `http://mygrafana:3000`.
However, users may want to add static query parameters such as org id:

`http://mygrafana:3000?orgId=1`

Today, it would break the links generation: would produce `http://mygrafana:3000?orgId=1?var_namespace=default` for instance, instead of `http://mygrafana:3000?orgId=1&var_namespace=default` (notice `?` instead of `&`).

So it needs to be fixed...

PR #2687, opened 2020-04-22
Allow pre-set query params in Grafana URL

jotak:
Example: Configured Grafana URL 'http://grafana:3000?orgId=1' to not screw up link generation

Fixes https://github.com/kiali/kiali/issues/2623

Frontend PR: https://github.com/kiali/kiali-ui/pull/1752

### DIFF
--- a/handlers/grafana.go
+++ b/handlers/grafana.go
@@ -73,14 +73,20 @@ func getGrafanaInfo(requestToken string, dashboardSupplier dashboardSupplier) (*
 
 	// Call Grafana REST API to get dashboard urls
 	links := []kmodel.ExternalLink{}
+	urlParts := strings.Split(externalURL, "?")
 	for _, dashboardConfig := range grafanaConfig.Dashboards {
 		dashboardPath, err := getDashboardPath(apiURL, dashboardConfig.Name, &auth, dashboardSupplier)
 		if err != nil {
 			return nil, http.StatusServiceUnavailable, err
 		}
 		if dashboardPath != "" {
+			// E.g.: http://localhost:3000?orgId=1 transformed into http://localhost:3000/d/LJ_uJAvmk/istio-service-dashboard?orgId=1
+			externalURL = strings.TrimSuffix(urlParts[0], "/") + "/" + strings.TrimPrefix(dashboardPath, "/")
+			if len(urlParts) > 1 {
+				externalURL = externalURL + "?" + urlParts[1]
+			}
 			externalLink := kmodel.ExternalLink{
-				URL:  strings.TrimSuffix(externalURL, "/") + "/" + strings.TrimPrefix(dashboardPath, "/"),
+				URL:  externalURL,
 				Name: dashboardConfig.Name,
 				Variables: v1alpha1.MonitoringDashboardExternalLinkVariables{
 					App:       dashboardConfig.Variables.App,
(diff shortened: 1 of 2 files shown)


jotak:
@hhovsepy as discussed, I've added a change to allow adding orgId / query params to in_cluster_url too.

A trick to test is to configure Kiali with these query params in URL, shut down grafana then go to a metrics page where links are supposed to be displayed => Kiali logs will show in the error the grafana URL that it tries to reach. E.g. for me I see:

```
E0422 15:31:54.081187       1 grafana.go:35] Get "http://grafana.istio-system:3000/api/search?query=Istio%20Service%20Dashboard&orgId=1": dial tcp 10.103.203.126:3000: connect: connection refused
```

handlers/grafana.go
```diff
@@ -73,14 +73,20 @@ func getGrafanaInfo(requestToken string, dashboardSupplier dashboardSupplier) (*
 
 	// Call Grafana REST API to get dashboard urls
 	links := []kmodel.ExternalLink{}
+	urlParts := strings.Split(externalURL, "?")
```
lucasponce, line 76: Is this expected format documented in Kiali CR template ?
(It should be the main source for these formats doc, also good if kiali.io can be updated).

jotak: What do you mean? I don't think there's anything to change in CR documentation, simple URL will continue to work as before. It's just that now we *also* accept URLs that contain query params.

lucasponce: Yes, specially as "http://grafana:3000?orgId=1" is a non standard format and it should be well documented to redirect users when something goes wrong.

lucasponce: This needs to be updated:
https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L405

How the user would know that this format is available ?

jotak: but the point is that it's not wrong anymore with this PR. Would you like us to say in doc: "if you set an URL with query params, such as http://grafana:3000?orgId=1 , it's going to work as well" ?

lucasponce: I think it would be good to mention that in the Kiali CR comment.

jotak: Also I don't understand why you say it's non standard. Query params in URL are standard (?)

jotak: > And what is the expected one ? "http://grafana:3000" ? "http://grafana:3000/" ? "http://grafana:3000/?orgId=1" ?

All are correct, that's my point. With this PR we're not tightening expectations, on the contrary, we're relaxing them.

jotak: Note that when we write `urlParts := strings.Split(externalURL, "?")` , if the URL doesn't contain `?`, then it all goes in `urlParts[0]`, that's why it doesn't break the existing format.

jotak: There's also a couple of tests in `grafana_test.go` to ensure old formats continue to work.

lucasponce: I don't understand why user need to guess that static parameters are allowed.

What I'm saying is that some similar comment in the Kiali CR (because the CR is the source of truth for config documentation) like

"Users may want to add static query parameters such as org id:
http://mygrafana:3000?orgId=1"

jotak: Ok, I'm going to add that, although I'm not really convinced. Query params *are* standard in URL, so to me the fact that we expect an URL should implicitly mean it's allowed.

This is the opposite that would be problematic IMO: if setting query params in a URL would break anything, that should be mentioned because it means we don't accept any kind of URL. Which is the current situation for Grafana without this PR, and perhaps (would need to check) the situation with jaeger as well. I believe it's more important to tell when we don't follow standard, than when we do follow it.

Anyway, it's not a big deal to add it to doc, so i'll do it.

jotak: hmm looks like it's in the operator repo so it's going to be a separate PR anyway

jotak: BTW if we start documenting to that level of granularity, we should do the same for everywhere we expect an URL: tell if query params are allowed or not?

lucasponce: If someone has a reasonable doubt about a config parameter, asking to add some comment in the Kiali CR is always good.

We have not started anything new. 
We are reviewing a PR and there was a comment about the format of a URL when we have several comments about what is the well formed one.

Or perhaps I don't understand your question or the intention.

PR #2713, opened 2020-04-29
[hack] script to install SM 1.1

jmazzitelli:
Ripping out the obsolete code that installed the old SM stuff from the AWS hack script - that script should only be responsible for installing the AWS cluster (just like how the CRC hack script works).

We will introduce different scripts to install different versions of SM. We currently have install-sm10.sh and this PR will introduce install-sm11.sh to install SM 1.1.

(this is ready to be merged)


jmazzitelli:
note: a followup PR will be coming to move the AWS hack script to OpenShift 4.4. But that is a task for another day. This PR is to get SM 1.1 installed in any OS cluster (currently testing on openshift 4.3 but should work with 4.4 as well)

Issue #2711, opened 2020-04-29
Issue updating ConfigMap from Kiali Operator


lucasponce:
Please @jmazzitelli, take a look to this problem.

I have a dev environment on minikube.
After update the operator in this way:

```
  namespace: kiali-operator
  resourceVersion: "250912"
  selfLink: /apis/kiali.io/v1alpha1/namespaces/kiali-operator/kialis/kiali
  uid: 136500c1-8898-11ea-98b0-e0f4c42844e7
spec:
  auth:
    strategy: login
  deployment:
    accessible_namespaces:
    - '**'
    image_name: localhost:5000/kiali/kiali
    image_pull_policy: Always
    image_version: dev
    namespace: istio-system
    service_type: ClusterIP
    verbose_mode: 3
  extensions:
    iter8:
      enabled: true
    threescale:
      adapter_name: threescale
      adapter_port: 3333
      adapter_service: threescale-istio-adapter
      enabled: true
  version: default
```

I can see that reconciled ConfigMap has:

```
[lponce@newrecopolis k-charted]$ kubectl get configmaps/kiali -n istio-system -o yaml
apiVersion: v1
data:
  config.yaml: |
    additional_display_details:
    - annotation: kiali.io/api-spec
      icon_annotation: kiali.io/api-type
      title: API Documentation
    api:
      namespaces:
        exclude:
        - istio-operator
        - kube.*
        - openshift.*
        - ibm.*
        - kiali-operator
    auth:
      ldap:
        ldap_base: ''
        ldap_bind_dn: ''
        ldap_group_filter: (cn=%s)
        ldap_host: ''
        ldap_insecure_skip_verify: false
        ldap_mail_id_key: mail
        ldap_member_of_key: memberOf
        ldap_port: 0
        ldap_role_filter: ''
        ldap_search_filter: (&(name={USERID}))
        ldap_use_ssl: false
        ldap_user_filter: (cn=%s)
        ldap_user_id_key: cn
      strategy: login
    deployment:
      accessible_namespaces:
      - '**'
      affinity:
        node: {}
        pod: {}
        pod_anti: {}
      image_name: localhost:5000/kiali/kiali
      image_pull_policy: Always
      image_pull_secrets: []
      image_version: dev
      ingress_enabled: true
      namespace: istio-system
      node_selector: {}
      pod_annotations: {}
      priority_class_name: ''
      replicas: 1
      resources: {}
      secret_name: kiali
      service_annotations: {}
[...]

PR #2714, opened 2020-04-29
Fixes Iter8 literal in config to please ansible operator

lucasponce:
I won't refactor all code, just the mapping into the yaml.

Fixes https://github.com/kiali/kiali/issues/2711

Note, this requires a PR on the operator side (on the way).


lucasponce:
cc @kiali/iter8 just to be aware this minor bug.
It's not critial, but it may affect to @kalantar work with the Iter8 operator as config in Kiali will have this minor change.

PR #2724, opened 2020-05-03
[operator] be able to specify what images to test via env vars.

jmazzitelli:
This is just for making running molecule tests with different images easier.
This PR goes with: https://github.com/kiali/kiali-operator/pull/25

You can now run the molecule-test target with the env var `MOLECULE_USE_DEV_IMAGES=true` in order to test locally build images (not specifying that will test the "latest" image from quay.io). For example:

```
MOLECULE_USE_DEV_IMAGES=true MOLECULE_SCENARIO=config-values-test make molecule-test
```

### DIFF
--- a/make/Makefile.molecule.mk
+++ b/make/Makefile.molecule.mk
@@ -7,6 +7,15 @@
 # To run multiple scenarios sequentially: MOLECULE_SCENARIO="token-test roles-test" make molecule-test
 MOLECULE_SCENARIO ?= default
 
+# Set MOLECULE_USE_DEV_IMAGES to true to use your local Kiali dev builds (not released images from quay.io).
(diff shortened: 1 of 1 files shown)


make/Makefile.molecule.mk
```diff
@@ -7,6 +7,15 @@
 # To run multiple scenarios sequentially: MOLECULE_SCENARIO="token-test roles-test" make molecule-test
 MOLECULE_SCENARIO ?= default
 
+# If you want to use your local Kiali dev builds (not released images from quay.io) then set this to 'true'.
```
jshaughn, line 10: Change first sentence to:

> Set MOLECULE_USE_DEV_IMAGES to true to use your local Kiali dev builds (not released images from quay.io).

jmazzitelli: done.

Issue #2674, opened 2020-04-17
auto-generate all-in-one yaml during release process


jmazzitelli:
Some people do not want to run a script (either over the Internet or locally after a git clone) but still want to install Kiali Operator and Kiali. An all-in-one yaml would be helpful. But we do not want to have to manually create and maintain such a yaml.

We should look into adding a new target to the kiali-operator Makefile that executes the [merge-operator-yaml.sh](https://github.com/kiali/kiali-operator/blob/master/deploy/merge-operator-yaml.sh) script in order to build a default all-in-one yaml. We could then invoke this make target as part of our Jenkins release builds and commit the auto-generated all-in-one yaml to the git repo (perhaps inside [the operator's deploy directory](https://github.com/kiali/kiali-operator/blob/master/deploy).

(of course, we could investigate providing helm charts, but that would involve a lot more work and the helm charts themselves would have to be maintained. It would be best if we do go with helm that we remove the entire deploy script mechanism and simply maintain helm charts - but that work is not covered in this enhancement request).

PR #2730, opened 2020-05-04
jenkins generation of all-in-one operator deployment yaml

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2674
This goes with https://github.com/kiali/kiali-operator/pull/28

This generates the all-in-one yaml and commits it to the deploy/ directory.

This all-in-one yaml should only be used for those that do not have nor wish to use Operator Lifecycle Manager to install and manage the Kiali operator installation.

### DIFF
--- a/deploy/jenkins-ci/Dockerfile
+++ b/deploy/jenkins-ci/Dockerfile
@@ -56,6 +56,8 @@ RUN set -eux; \
     # Install tools: build tools, gosu, docker-cli \
     apt-get -qq update; \
     apt-get install --no-install-recommends -qq build-essential gosu docker-ce-cli; \
+    # Install some things to support generating operator all-in-one yaml (gettext-base has 'envsubst') \
+    apt-get -qq install gettext-base; \
     # Create Docker group \
     groupadd -r docker; \
     usermod -aG docker jenkins; \
--- a/deploy/jenkins-ci/Makefile.operator.jenkins
+++ b/deploy/jenkins-ci/Makefile.operator.jenkins
@@ -90,6 +90,11 @@ operator-build-release:
 	sed -i -r 's/^VERSION \?= v.*/VERSION \?= $(OPERATOR_VERSION_TO_RELEASE)/' Makefile
 	OPERATOR_QUAY_NAME="$(QUAY_OPERATOR_NAME)" \
 	  $(MAKE) clean build
+ifneq ($(IS_SNAPSHOT),y)
+	OPERATOR_IMAGE_VERSION="$(OPERATOR_VERSION_TO_RELEASE)" \
+	  $(MAKE) generate-all-in-one
+	cp _output/kiali-operator-all-in-one.yaml deploy/
+endif
 
 operator-push-quay:
 ifdef QUAY_USER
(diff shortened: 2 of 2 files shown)


jmazzitelli:
@israel-hdez The Jenkinsfile will now auto-generate an all-in-one operator install yaml (see the community request for this feature here: https://github.com/kiali/kiali/issues/2674)

One major thing I need you to look at is the change to the Dockerfile. We need "envsubst" (because the new operator makefile target uses envsubst to generate the new all-in-one file) - this requires `apt-get install gettext-base` in the Dockerfile. I think this means we need to rebuild/restart the real Jenkins server so it gets this new dependency installed.

I tested this on a local Jenkins server and things seem to work. Here's what the two commits will look like after a minor release (I ran the test against a copy of kiali-operator repo that I have in my personal github account):
* The release commit: https://github.com/jmazzitelli/kiali-org-kiali-operator/commit/cb61e8edaab230e5b46a20e1e2e12b77d87deb73
* The prepare-for-next-version commit: https://github.com/jmazzitelli/kiali-org-kiali-operator/commit/27f12244acf6f6d872410649a8f8f10a017a39d9

Notice that not only is the operator's Makefile changed (to update VERSION) but an updated all-in-one yaml file is committed. This ensures any changes we needed to make in the deploy/\*.yaml files are reflected in this all-in-one. With Jenkins now auto-generating this all-in-one file, we do not have to maintain it manually. We just need to maintain the individual deploy/\*.yaml files (which we already are doing now anyway).

deploy/jenkins-ci/Dockerfile
```diff
@@ -62,7 +62,9 @@ RUN set -eux; \
     # Clean-up \
     rm -rf /var/lib/apt/lists/*; \
     # Set exec permissions to entrypoint; \
-    chmod +x /usr/local/bin/entrypoint.sh;
+    chmod +x /usr/local/bin/entrypoint.sh; \
+    # Do some things to support generating operator all-in-one yaml \
+    apt-get -qq update; apt-get -qq install gettext-base;
```
israel-hdez, line 67: Could you move these new lines before the apt clean-up (before line 63)? Just to have an image with reduced size. If you move this before the clean-up you may not need to invoke `apt-get -qq update`.

deploy/jenkins-ci/Makefile.operator.jenkins
```diff
@@ -156,8 +162,12 @@ endif
 operator-prepare-master-next-version:
 	# Only minor releases require to prepare the master branch for the next release
 ifeq ($(RELEASE_TYPE),minor)
+	OPERATOR_IMAGE_VERSION="latest" \
+	  $(MAKE) generate-all-in-one
+	cp _output/kiali-operator-all-in-one.yaml deploy/
 	sed -i -r "s/^VERSION \?= (.*)/VERSION \?= v$(OPERATOR_BUMPED_VERSION)-SNAPSHOT/" Makefile
 	git add Makefile
+	git add deploy/kiali-operator-all-in-one.yaml
```
israel-hdez, line 170: There is the unfortunate outcome that the all-in-one YAML file will be updated on master only when the operator is released. Not sure if you would prefer to update it "automagically" every time is needed.

Just a comment. Not sure how we really could deal with this.

jmazzitelli: Yes, this is the intention. We auto-generate this all-in-one when the operator is released. If we ever need to update it any other time (for whatever reason) we can always manually generate it. The point here is that we release an all-in-one for each minor release, and we just maintain a copy in master. But the master copy is less important than the ones in the release branch.

deploy/jenkins-ci/Dockerfile
jmazzitelli: this is done - there is now just 2 lines added to dockerfile (one is a comment) and I took out the apt-get update.

Issue #2727, opened 2020-05-04
gracefully ignore missing Policy CRD


jmazzitelli:
Upstream Istio removed the Policy CRD in 1.6. Kiali expects it to exist. Kiali should assume this is "optional" so it supports older versions of Istio but yet won't spit out errors if its missing.

See: https://github.com/istio/istio/issues/23484

lucasponce:
@xeviknal I put this on you, but let me know if you want me to take it.

xeviknal:
The way to properly fix this bug is to cover https://github.com/kiali/kiali/issues/2520. I think that covering the first 3 bullets, could be a good start. I can approach them.

xeviknal:
That error message is not displayed anymore. Fixed in #2738

PR #2738, opened 2020-05-05
Allow Policy to return an empty array on Istio without Authentication API

xeviknal:
This PR has the responsability to allow Kiali to run into Istio 1.5 and Istio 1.6 without triggering alerts because of the consumption of the Policy CRD.

This is preventing me to implement further features based on Istio 1.6.

The behavior is the following:

On Istio 1.5:
Screen recording (27)

On Istio 1.6:
Screen recording (26)

Makes #2727 less ugly. However, I couldn't consider that issue close because this workaround makes the MTLS icons disappear into istio 1.6 (except for the graph).

### DIFF
--- a/kubernetes/client.go
+++ b/kubernetes/client.go
@@ -149,6 +149,11 @@ type IstioClient struct {
 	// See istio_details_service.go#HasSecurityResource() for more details.
 	securityResources *map[string]bool
 
+	// securityResources private variable will check which resources kiali has access to from security.istio.io group
+	// It is represented as a pointer to include the initialization phase.
+	// See istio_details_service.go#HasSecurityResource() for more details.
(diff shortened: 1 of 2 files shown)


kubernetes/client.go
```diff
@@ -149,6 +149,11 @@ type IstioClient struct {
 	// See istio_details_service.go#HasSecurityResource() for more details.
 	securityResources *map[string]bool
 
+	// securityResources private variable will check which resources kiali has access to from security.istio.io group
+	// It is represented as a pointer to include the initialization phase.
+	// See istio_details_service.go#HasSecurityResource() for more details.
```
lucasponce, line 154: Minor: Update the comments, are copied from securityResources var and this is authenticationResources with also uses a different method.

PR #2742, opened 2020-05-06
[Kiali 1.12] OSSM-199 Fix missing service details when jaeger disabled

jotak:


lucasponce:
@jshaughn Note for Kiali v1.12.x there will be changes on both backend/frontend repos.

Issue #2737, opened 2020-05-05
Add Support of RequestAuthentication in Kiali backend


lucasponce:
As part of RBAC deprecation.
Kiali should support new RequestAuthentication object in backend:

https://preliminary.istio.io/docs/reference/config/security/jwt/
https://preliminary.istio.io/docs/reference/config/security/request_authentication/

Issue #2226, opened 2020-02-10
Review Kiali with Istio 1.6+


lucasponce:
Istio 1.6+ telemetry and configuration models are changing and it may impact in Kiali features.

We need an on-going EPIC where we can coordinate all tasks needed to identify potential imcompatibiliy areas and properly plan changes.

Preliminary tasks:
- [ ] Telemetry model
- [ ] Networking model
- [ ] Security model
- [ ] Mixer model and how extensions used in Kiali based on Adapters may be affected

lucasponce:
Related:
is:open is:issue milestone:1.6

jmazzitelli:
v2 Telemetry changed in 1.5 that is causing annoying "unknown" telemetry traffic to be recorded: https://github.com/istio/istio/issues/21762

jotak:
I opened this issue to test Kiali with production-scale prom setup, it relates to Istio 1.5 and more: https://github.com/kiali/kiali/issues/2518

jotak:
This issue will also fall under the 1.6 epic umbrella: https://github.com/kiali/kiali/issues/2582

According to [this doc PR](https://github.com/istio/istio.io/pull/7112/files#diff-976e9322ab00eac62f7bdf7a8b709f0bR41-R53), I expect some easier management of prometheus scraping for envoy and/or runtimes metrics.

(PS @lucasponce perhaps you can edit the top comment to add #2582 to the list of related issue?)

lucasponce:
We will need to update https://preliminary.istio.io/docs/tasks/observability/kiali/ as part of this Epic.
It wording is still valid, perhaps update the screenshots with the version shipped/recommended in Istio would be enough.

PR #2746, opened 2020-05-07
Add WorkloadEntry and RequestAuthentication for Istio 1.6

lucasponce:
This PRs adds WorkloadEntry and RequestAuthentication CRD into Kiali backend.

It's related to issues:
https://github.com/kiali/kiali/issues/2226
https://github.com/kiali/kiali/issues/2737

UI counterpart:
https://github.com/kiali/kiali-ui/pull/1778

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -877,6 +877,12 @@ func (in *IstioConfigService) modifyIstioConfigDetail(api, namespace, resourceTy
 	case PeerAuthentications:
 		istioConfigDetail.PeerAuthentication = &models.PeerAuthentication{}
 		istioConfigDetail.PeerAuthentication.Parse(result)
+	case RequestAuthentications:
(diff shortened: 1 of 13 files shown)


business/istio_config.go
```diff
@@ -152,10 +161,10 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 		return models.IstioConfigList{}, err
 	}
 
-	errChan := make(chan error, 19)
+	errChan := make(chan error, 20)
```
xeviknal, line 164: Should this be a 21? You adding to new calls, so you expect up to two more errors in that channel.

lucasponce: I think it was out-of-sync and I've adjusted.
image

There are 22 goroutines there, but only errChan is passed to 20.

I think that I don't miss it but let me know if I missed something perhaps I coded it fast.

Issue #2670, opened 2020-04-16
Labels in Apps list are randomly changed after refresh


hhovsepy:
Happens on Kiali master. And issue coming from Kiali backend.

List applications.
Search for "jaeger"
Refresh the results several times.
The "jaeger-infra: " label's value is changing with refresh.
This happens in the REST API call as well. "api/namespaces/istio-system/apps".

![Screenshot from 2020-04-16 17-35-24](https://user-images.githubusercontent.com/604313/79477084-0cbd9200-800a-11ea-9971-3fac293e98ec.png)
![Screenshot from 2020-04-16 17-35-26](https://user-images.githubusercontent.com/604313/79477087-0d562880-800a-11ea-86ba-31094f70882b.png)
![Screenshot from 2020-04-16 17-35-35](https://user-images.githubusercontent.com/604313/79477092-0deebf00-800a-11ea-86a8-c544fc8fef62.png)
![Screenshot from 2020-04-16 17-36-22](https://user-images.githubusercontent.com/604313/79477093-0e875580-800a-11ea-837e-5bbaec21f280.png)
![Screenshot from 2020-04-16 17-36-49](https://user-images.githubusercontent.com/604313/79477095-0f1fec00-800a-11ea-9414-e5219c56acc5.png)
![Screenshot from 2020-04-16 17-37-14](https://user-images.githubusercontent.com/604313/79477097-0f1fec00-800a-11ea-964c-ad8ebe23096f.png)

aljesusg:
Ok so the "thing" that change is the value, I thought it was related with the order/sort. 

Is this happen when you deploy the environment in the first time or ever?

I think that maybe the jaeger-infra is not 100% ready in this case because is adding new components.

1. collector
2. collector, jaeger-service
3. collector, jaeger-service, agent-service

Is this changing in refresh after the 3º result? In that case there is a problem. On other case I think that this value change after that app is ready.

Maybe the Jaeger team can teach us about this.
Can you confirm this @objectiser, @pavolloffay  or @jpkrohling ?

hhovsepy:
The change happens  much later after environment deployment, and not in order 1,2,3 but randomly, so after 3 labels it can switch to any of 1,2.

aljesusg:
Ok, I'm going to check it because it's weird.
Is this only with Jaeger service?

pavolloffay:
Which component is managing the object that has these labels? You can find that by owner reference in labels/annotations.

aljesusg:
[...]

PR #2768, opened 2020-05-12
Fix apps labels being incomplete when duplicates are found

jotak:
When a duplicate is found, it must just be ignored, not erase the previous values

Adding some unit tests

Fixes https://github.com/kiali/kiali/issues/2670

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -2,6 +2,7 @@ package business
 
 import (
 	"fmt"
+	"sort"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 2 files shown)


jotak:
PS: I've found another (small) problem that I can try to fix in the same PR: labels are not sorted and their ordering may change upon refresh. We could sort them to display more consistently

jotak:
@aljesusg I've pushed a new one to sort labels by value.
Example, `jaeger-infra: agent-service,collector-service,jaeger-service`, values are sorted alphabetically to avoid changing order on refresh

aljesusg:
> @aljesusg I've pushed a new one to sort labels by value.
> Example, `jaeger-infra: agent-service,collector-service,jaeger-service`, values are sorted alphabetically to avoid changing order on refresh

Maybe we should sort the values from backend.. I thought that istio had them sorted

jotak:
(oops sorry I edited your comment instead of quoting it, my bad)

I don't know if they come sorted after being fetched, but since they're stored in a map, their order isn't guaranteed anyway

aljesusg:
> (oops sorry I edited your comment instead of quoting it, my bad)
> 
> Hmm actually it's the order of workloads/services related to the app that may change, and make the ordering of related labels changing

I thought that go lets sort when transform to json for marshall...ok so from me side is ok tos ort them in UI

jotak:
It's sorted in my last commit so no need to do it from UI

jotak:
thanks @hhovsepy @aljesusg

Issue #2772, opened 2020-05-13
Overview page: Services show more items in Istio 1.6


hhovsepy:
Happens on Kiali master with Istio 1.6.0.

In Overview page, select Show Healt for Services.
For 'bookinfo' namespace it shows 7 services and 5 healthy.
But in fact bookinfo has 6 services and 4 healthy.
Additional service can be found in tooltip:
![Screenshot from 2020-05-13 09-21-13](https://user-images.githubusercontent.com/604313/81796995-7bf5ab80-950e-11ea-8308-18de8754d974.png)
![Screenshot from 2020-05-13 09-21-37](https://user-images.githubusercontent.com/604313/81797004-7e580580-950e-11ea-87cb-f60003dcabad.png)

PR #2774, opened 2020-05-13
Fix overview services health showing inexisting services

jotak:
It was adding services based on istio metrics in addition to the ones strictly resulting from k8s api

Fixes https://github.com/kiali/kiali/issues/2772

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -200,12 +200,9 @@ func (in *HealthService) getNamespaceServiceHealth(namespace string, services []
 	lblDestSvc := model.LabelName("destination_service_name")
 	for _, sample := range rates {
 		service := string(sample.Metric[lblDestSvc])
-		health, ok := allHealth[service]
-		if !ok {
-			health = &models.ServiceHealth{Requests: models.NewEmptyRequestHealth()}
-			allHealth[service] = health
+		if health, ok := allHealth[service]; ok {
+			health.Requests.AggregateInbound(sample)
 		}
-		health.Requests.AggregateInbound(sample)
 	}
 
 	return allHealth


jotak:
I started to fix that right away, but actually, perhaps it needs a little bit of discussion before merging.
Here, we're showing health of services within a namespace. At a first glance, only "real" services should be counted, the ones that actually refer to a k8s object.

BUT if we don't show health related to istio's "special" services, such as passthrough/blackhole clusters, then we're going to hide potentially unhealthy things.

What's the best thing to do according to you? cc @lucasponce @jshaughn ...

jotak:
perhaps the right thing to do would be to ignore them unless they're unhealthy

jshaughn:
@jotak Passthrough and BlackHole are not really services but sort of request handlers.  As entities themselves I'd ignore them.    As an indicator of health, anything going to BlackHole is likely a problem, but those requests will already return with a failure status.

jotak:
@jshaughn yes, but then in that particular view (namespaces, service-based health) we would see everything fine because we're only looking at heath of services. We would have to switch to app/workload-based health to see the errors.

Actually I start wondering if having a service-based health in overview makes any sense at all to be honest.

jshaughn:
@jotak   Service health is a little hard to define, I guess it comes down to the success rate of requests for that destination service.  But only for destination services that have an explicit k8s service defined.  I don't think health applies to every "terminal node" that shows up in the graph, which is to say that I don't think every request maps to the health of some service.  Long way of saying that I think the initial fix is the right one.  The removal of service health itself is probably overkill.

jotak:
I'll merge this change, but I still think we could simplify this overview page a little bit. What I think people expect from an overview is something that offers a complete overview from a single place, and not having some particularities depending on the angle of view that we're taking.

jotak:
PS @jshaughn while you're here could you review? It's a small one :-)

jotak:
Thanks!

Issue #2770, opened 2020-05-12
Disable the use of TLS versions below 1.2


jmazzitelli:
Kiali server process should not use TLS versions below 1.2.

Specifically: https://github.com/kiali/kiali/blob/973a34c98cae1680849daf9153bcafcf6d36502b/server/server.go#L62-L67

At a minimum, resolving this requires adding a TLSConfig to the http.Server definition which specifies the minimum TLS version as 1.2, e.g.:

```
index aa7ae5b7..68f6359c 100644
--- a/server/server.go
+++ b/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+       "crypto/tls"
        "fmt"
        "net/http"
 
@@ -40,9 +41,14 @@ func NewServer() *Server {
        http.DefaultServeMux = mux
        http.Handle("/", handler)
 
+       tlsConfig := &tls.Config{
+               MinVersion: tls.VersionTLS12,
+       }
+
        // create the server definition that will handle both console and api server traffic
        httpServer := &http.Server{
-               Addr: fmt.Sprintf("%v:%v", conf.Server.Address, conf.Server.Port),
+               Addr:      fmt.Sprintf("%v:%v", conf.Server.Address, conf.Server.Port),
+               TLSConfig: tlsConfig,
        }
 
        // return our new Server
```

We'd need to add a test when implementing this. See:
https://github.com/kiali/kiali/blob/93f5cd0b6698e8fe8772afb8f35816f6c086aef1/server/server_test.go#L156

mattmahoneyrh:
verified on v1.12 branch

PR #2782, opened 2020-05-13
server should not allow TLS 1.1 or lower

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/2770

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -1,6 +1,7 @@
 package server
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net/http"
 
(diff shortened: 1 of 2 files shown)


jmazzitelli:
This has a unit test showing it working.

jmazzitelli:
The request from Brian "redbeard" is to support TLS 1.2 but to reject anything lower. So I assume we'll still want 1.2 support for now.

brianredbeard:
Late to the party, but _personally_ I'm fine going all the way to 1.3.  The present urgency was that 1.0 and 1.1 fail a number of security audits.  Thanks though, folks are going to be excited.

PR #2786, opened 2020-05-14
AutoEnableMTLS by default is true

xeviknal:
I've detected a change from 1.6-beta.0 to beta.1. The Istio config map no longer includes the 'enableAutoMtls' field in the default installation. It means that they defaults to true.

My previous deployment defaulted to false.
This PR includes this default to true when the flag is not present.

### DIFF
--- a/kubernetes/types.go
+++ b/kubernetes/types.go
@@ -437,7 +437,7 @@ type IstioObjectList interface {
 }
 
 type IstioMeshConfig struct {
-	DisableMixerHttpReports bool `yaml:"disableMixerHttpReports,omitempty"`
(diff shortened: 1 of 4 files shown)


kubernetes/types.go
```diff
@@ -641,3 +641,10 @@ func (in *GenericIstioObjectList) DeepCopyObject() runtime.Object {
 	}
 	return nil
 }
+
+func (imc IstioMeshConfig) GetEnableAutoMtls() bool {
```
lucasponce, line 645: I have a concern to grow business logic inside kubernetes/ package instead of business/ model/ ones.
Not a blocker, but at some point we need to revisit this logic to try to not have a mix of logic in the kubernetes/ and business + model ones.

hhovsepy:
Post merge verification.

Issue #2056, opened 2019-12-19
Support OAuth login (OpenID connect)


harpratap:
**Is your feature request related to a problem? Please describe.**
We use Keycloak for user authentication for all of our components, would like to have in built support in Kiali (like Grafana). Right now we have a sidecar proxy that does this for us, but it's very inefficient as every single request (including prometheus queries) require to go through the authentication first which makes everything slow. 

**Describe the solution you'd like**
Kiali supports OAuth 2.0 natively with OpenID connect integration (Keycloak)

**Describe alternatives you've considered**
Used Gatekeeper for Keycloak as a sidecar but it makes everything extremely slow.

**Additional context**
UI could bring a SSO button instead of username and password which redirects to our OpenID system based on Keycloak. Rest of the changes are in backend. Would prefer only user authentication to happen via this mechanism, prometheus/Jaeger/Grafana queries can still without auth.

lucasponce:
Thanks for the issue @harpratap.
For the specific OpenShift platform there is a native integration, what platform are you using ?

Also, in OpenShift, Kiali deletages in the user's RBAC rights, I guess with the feature request you propose here, you mean only authentication, right ? But do you envision also to touch the authorization part managed by Keycloack ?

harpratap:
> Thanks for the issue @harpratap.
> For the specific OpenShift platform there is a native integration, what platform are you using ?

It's an on-prem cluster we built on bare metal. Not using any provider

> Also, in OpenShift, Kiali deletages in the user's RBAC rights, I guess with the feature request you propose here, you mean only authentication, right ? But do you envision also to touch the authorization part managed by Keycloack ?

Yes, the RBAC controls what Kiali can/cannot do in the kubernetes cluster. But anyone who has access to the UI of Kiali is able to make use of whatever permissions Kiali has access to. So we want to provide access to only the authenticated users (which we manage via groups made in Keycloak). We personally don't need authorization (we just use RBAC for it and works fine for our use case).

lucasponce:
> Yes, the RBAC controls what Kiali can/cannot do in the kubernetes cluster. But anyone who has access to the UI of Kiali is able to make use of whatever permissions Kiali has access to. So we want to provide access to only the authenticated users (which we manage via groups made in Keycloak). We personally don't need authorization (we just use RBAC for it and works fine for our use case).

[...]

PR #2798, opened 2020-05-18
Add support for authentication through an OpenId provider

israel-hdez:
Add support for authentication through an OpenId provider

Part of kiali/kiali#2056


jmazzitelli:
I know this is currently a draft and WIP and not ready for review, but I want to bring this up now... I'm sure you know you will need unit tests for this to ensure no security holes exist ;)

israel-hdez:
@jmazzitelli I hope my tests won't be buggy ;)

handlers/authentication.go
```diff
@@ -682,9 +683,16 @@ func AuthenticationInfo(w http.ResponseWriter, r *http.Request) {
...
+		if !strings.Contains(scopes, "openid") {
+			scopes = "openid " + scopes
+		}
+
+		response.AuthorizationEndpoint = fmt.Sprintf("%s?client_id=%s&response_type=id_token&redirect_uri=%s&scope=%s&nonce=%s",
+			conf.Auth.OpenId.AuthorizationEndpoint,
+			url.QueryEscape(conf.Auth.OpenId.ClientId),
+			url.QueryEscape(httputil.GuessKialiURL(r)),
+			url.QueryEscape(scopes),
 			"asdf123456")
```
jmazzitelli, line 696: hardcoded strings like this scare me now - what is this? If this is some kind of key or seed, then we need to randomize somewhere.

israel-hdez: Yes, `nonce` should be a random string. It's usage is to mitigate replay attacks. I'll replace this dummy string in a subsequent commit.

israel-hdez:
Rebased up to now. Added a small unit test for GuessKialiURL.
Promoting, because IMO it's good enough.

israel-hdez:
@jshaughn @jmazzitelli For a Minikube+Keycloak setup, you could use this repo: https://github.com/mrbobbytables/oidckube. The author of the repo has a guide with screenshots here: https://medium.com/@mrbobbytables/kubernetes-day-2-operations-authn-authz-with-oidc-and-a-little-help-from-keycloak-de4ea1bdbbe#172c.

I haven't tried by myself that guide. I tried with dex. Just referring those links because I think you prefer to try with Keycloak ;-)

israel-hdez:
Asking a review from @jotak to have some eyes from Jupiter side (who are nearer from the Earth where all this will be installed).

business/openid_auth.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// Create HTTP client
+	httpTransport := &http.Transport{}
+	if cfg.InsecureSkipVerifyTLS {
+		httpTransport.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true,
+		}
+	}
+
+	httpClient := http.Client{Transport: httpTransport}
```
jotak, line 76: There should be a timeout (even if a long one) : https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779

business/openid_auth.go
```diff
@@ -0,0 +1,140 @@
...
+	// Validate there is an authorization endpoint
+	if len(metadata.AuthURL) == 0 {
+		return nil, errors.New("the OpenID provider does not expose an authorization endpoint")
+	}
+
+	// Log warning if OpenId provider Metadata does not expose "id_token" in it's supported response types.
+	// It's possible to try authentication. If metadata is right, the error will be evident to the user when trying to login.
+	responseTypes := strings.Join(metadata.ResponseTypesSupported, " ")
+	if !strings.Contains(responseTypes, "id_token") {
+		log.Warning("Configured OpenID provider informs response_type=id_token is unsupported. Users may not able to login.")
```
jotak, line 116: typo "s/able to/be able to" ?

business/openid_auth.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	err = json.Unmarshal(rawMetadata, &metadata)
+	if err != nil {
+		return nil, fmt.Errorf("cannot parse OpenId Metadata: %s", err.Error())
+	}
+
+	// Validate issuer == issuerUri
+	if metadata.Issuer != cfg.IssuerUri {
+		return nil, errors.New("mismatch between the configured issuer_uri and the exposed Issuer URI in OpenId provider metadata")
```
jotak, line 104: perhaps add `metadata.Issuer` in the message to help users debugging?

business/openid_auth.go
```diff
@@ -0,0 +1,140 @@
...
+		isScopeSupported := false
+		for _, supportedScope := range metadata.ScopesSupported {
+			if scope == supportedScope {
+				isScopeSupported = true
+				break
+			}
+		}
+
+		if !isScopeSupported {
+			log.Warning("Configured OpenID provider informs some of the configured scopes are unsupported. Users may not able to login.")
```
jotak, line 132: typo "s/able to/be able to" ?

handlers/authentication.go
```diff
@@ -197,14 +200,112 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
...
 
 	RespondWithJSONIndent(w, http.StatusOK, TokenResponse{Token: tokenString, ExpiresOn: expiresOn.Format(time.RFC1123Z), Username: user.Metadata.Name})
 	return true
 }
 
+func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
+	// Check if the nonce cookie is present
+	nonceCookie, err := r.Cookie(openIdNonceCookieName)
+	if err != nil {
+		RespondWithError(w, http.StatusUnauthorized, "No nonce code present. Login window timed out.")
```
jotak, line 215: type s/code/cookie?

util/httputil/http_util.go
```diff
@@ -84,3 +86,49 @@ func AuthTransport(auth *config.Auth, transportConfig *http.Transport) (http.Rou
 
 	return newAuthRoundTripper(auth, transportConfig), nil
 }
+
+func GuessKialiURL(r *http.Request) string {
```
jotak, line 90: I wonder if there are corner cases where this would not work. Perhaps as a safety net we could add a field in openid config to override the redirection URL ? GuessKialiURL would be called only if this field isn't present?

Or are you confident that it should work in any case?

jmazzitelli:
> I tried with dex.

@israel-hdez how did you do it? I don't want to use that other stuff - that is a wrapper around minikube itself... I already have minikube and I use our hack script to start it and configure it. Is there something that explains "here is how you install an OpenID provider in an existing minikube cluster"?

business/openid_auth.go
```diff
@@ -0,0 +1,140 @@
...
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/log"
+)
+
+type OpenIdMetadata struct {
+	// Taken from https://github.com/coreos/go-oidc/blob/8d771559cf6e5111c9b9159810d0e4538e7cdc82/oidc.go
```
jmazzitelli, line 17: Isn't there a library we can pull in for this structure? What if something changes in the future? We would need to keep up-to-date with this metadata structure and manually update our code (rather than just updating a library).

util/random.go
```diff
@@ -1,17 +1,56 @@
...
+}
+
+// CryptoRandomString returns a securely generated random string.
+// It will return an error if the system's secure random
+// number generator fails to function correctly, in which
+// case the caller should not continue.
+//
+// Found at https://gist.github.com/dopey/c69559607800d2f2f90b1b1ed4e550fb
+func CryptoRandomString(n int) (string, error) {
+	const letters = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-"
```
jmazzitelli, line 47: I would prefer to (a) not use ASCII-sequential chars and (b) put special characters in here.

Suggest something like (which is different than the above random letters):

```
const letters = "gl9z2p1a%8c4Er3@XFyLvGf!ZMHubAYeTdUB$&nV6o#jRkWmC)qIPtS*QJ0h7(^K5OwDiNxs"
```

Where I randomized the order of the chars (simply via this command):

```
echo '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz!@#$%^&*()' | fold -w1 | shuf | tr -d '\n' ;echo
```

util/random.go
```diff
@@ -1,17 +1,56 @@
...
+)
 
 // RandomString generates a random string of length n. Before calling this function, you should call
 // rand.Seed() to initialize the default source.
 //
 // Found at https://ispycode.com/Blog/golang/2016-10/How-to-generate-a-random-string-of-a-fixed-length
+//  - Adapted for more characters
 func RandomString(n int) string {
-	var letterRunes = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	var letterRunes = []rune("0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz")
```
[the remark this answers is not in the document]
jotak: I'm not sure there is any benefit in randomized order ... and perhaps quite the opposite, as sequential order makes it easy to spot any weakness in the runes. For instance, imagine there's duplicated chars, which would result in less evenly distributed function: it would be very hard to notice with a pre-randomized order.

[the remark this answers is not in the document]
jmazzitelli: my concern here was that any internal randomize generator that messes with this "letters" (if its algorithm results can be predetermined) might make it more easy to reverse engineer the string if the reverse-engineer knows (or suspects) the seed was an ASCII-sequential string (of course, they could hunt down this line of code and see the actual string, so this is limited).

If you feel keeping an ASCII-sequential string that gets passed in is better, feel free to keep it. 

But in my opinion, at minimum, we should be using special characters in here too (not just alphanumeric chars) - thus increasing the span of possible characters used in the randomization process.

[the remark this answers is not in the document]
jotak: Actually I realize my comment applies for `CryptoRandomString` below, and not this `RandomString`. I wonder if we really have to keep `RandomString` at all?

Talking only about `CryptoRandomString`: the runes are not a seed, they don't impact the underlying random bytes generation which is (supposed to be) secure - I agree that adding more characters (and perhaps increasing size of the output string, currently it's 15) would increase randomness ; but I still don't think we have to feel worried about the initial order, perhaps I'm wrong.

But yes, +1 for adding a couple of special characters, it doesn't cost anything.

business/openid_auth.go
israel-hdez: Yes. The mismatch would be a very weird corner case. But I'll add the Issuer to the message.

util/httputil/http_util.go
israel-hdez: I was thinking about it yesterday.
The "success" of this function depends 100% on the client. If it doesn't send the HTTP headers, it's impossible to determine the public URL.

So, I think this is not totally safe. Thus, I'll add the new config value. I think I'll put it under the `Server` next to the `WebRoot` config which is closely related.

util/random.go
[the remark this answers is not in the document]
israel-hdez: Citing the spec (https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes):

> The nonce parameter value needs to include per-session state and be unguessable to attackers. One method to achieve this for Web Server Clients is to store a cryptographically random value as an HttpOnly session cookie and use a cryptographic hash of the value as the nonce parameter. In that case, the nonce in the returned ID Token is compared to the hash of the session cookie to detect ID Token replay by third parties. A related method applicable to JavaScript Clients is to store the cryptographically random value in HTML5 local storage and use a cryptographic hash of this value.

The important statement is that this string must be **unguessable to attackers**. As @jotak says, that sequential string is not a seed. A random string of 15 chars is being generated using the `crypto/rand` package which is supposed to be "cryptographycally secure" (see https://golang.org/pkg/crypto/rand/). So, I agree that it's not a problem to have that sequential string. I also don't see the need to add special characters (it's not a password :-) because the only requirement is to be unguessable. But, I'll anyway add a few chars.

What I see important in the `implementation notes` of the spec is the statement that there are two parts: (1) the random string, and (2) the cryptographic hash of that random string. The string (1) should live in the cookie, while the string (2) is the one passed to the IdP. Indeed, this makes sense to effectively block/mitigate replay attacks. The code, as it is, is using (1) for both and this doesn't result in a right mitigation (if you steal an `id_token`, you can craft the cookie to authenticate into Kiali) -- surprisingly, lots of examples in the internet suggest using the random string (1) for both the cookie and sending it to the IdP :worried: 

I'll check how to generate a `hash` to send to the IdP. It is passed in the URL, so there is a limitation in size. Once this is in place, I'm not sure if it's really useful to add special characters, but I'll add them anyway.

sandhose:
I just setup Kiali in my cluster and was wondering if there was any OpenID Connect auth support, and stumbled upon this PR, nice!

I did a lot around OIDC recently, so here are a few thoughts:

 - you really should implement the authorization code flow. I saw in one of your comments that for some reason you can't?
 - there is a nice RFC draft about security best practices in oauth2/oidc: https://tools.ietf.org/html/draft-ietf-oauth-security-topics-14
 - currently you're just parsing the ID token without verifying it. You really should be! Right now, anyone can craft a token with the right informations in it and log in with it. Using the code flow would avoid this issue
 - you might want users to allow specifying additional parameters to the redirect URI. Things like locale preferences, prompt style, etc. can be specified there
 - the nonce cookie should be signed somehow to avoid being crafted (a jwt would do it)
 - you might want to implement things like logging out, but this can be done outside this PR I guess
 - it seems like you're getting the ID token in the form data. That depends on the provider. It is usually advertised in the `response_modes_supported` metadata (can be `query`, `fragment` and `form_post`)
 - you might want to check for `error` in the callback. There are a few [well known errors](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#extensions-error), and the provider might include a `error_description` field with a human-readable error message

I've never contributed to Kiali and am very new to this project, but I can help you with all of that if you want. Would you mind if I do a PR on your branch in the next few days?

jmazzitelli:
> Would you mind if I do a PR on your branch in the next few days?

@sandhose Any help (by an OpenID expert especially ;-) would be greatly appreciated!

israel-hdez:
@jmazzitelli Try using the script in this gist:
https://gist.github.com/israel-hdez/6dd09251d4d96ee17de6a4a8e1ce339a

It will boot minikube, configure dex, restart it and OpenId connect will be configured. Then, all what remains is installing Kiali.

israel-hdez:
> * you really should implement the authorization code flow. I saw in one of your comments that for some reason you can't?

Kiali doesn't have back-end storage. So, we anyway would need to store the id_token in the cookie sent to the browser. I understand that one of the main ideas of the `authorization code flow` is to never expose the tokens to the browser. Without back-end storage, it doesn't make a lot of sense to implement it.

> * currently you're just parsing the ID token without verifying it. You really should be! Right now, anyone can craft a token with the right informations in it and log in with it. Using the code flow would avoid this issue

It's being verified in https://github.com/kiali/kiali/pull/2798/files#diff-7b14eebb8414820b40c718efcf921867R240-R250. Although, that validation is going to change.

> * you might want users to allow specifying additional parameters to the redirect URI. Things like locale preferences, prompt style, etc. can be specified there

I didn't know :)
Anyway, in this first iteration, I'll omit this. If you want to contribute, this would be a nice addition.

> * the nonce cookie should be signed somehow to avoid being crafted (a jwt would do it)

I'll use the suggested method as in https://openid.net/specs/openid-connect-core-1_0.html#NonceNotes.

By sending a cryptographic hash to the IdP instead of the randomly generated nonce code, there should be no need to sign the cookie.

> * you might want to implement things like logging out, but this can be done outside this PR I guess

Yes, I prefer to do this in follow-up PRs.

> * it seems like you're getting the ID token in the form data. That depends on the provider. It is usually advertised in the `response_modes_supported` metadata (can be `query`, `fragment` and `form_post`)

If I understand correctly, the Implicit Flow should always return the parameters via the hash portion of the URL, because it takes into account JavaScript applications. Since this PR is using Implicit Flow, I think it should be fine.

> * you might want to check for `error` in the callback. There are a few [well known errors](https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml#extensions-error), and the provider might include a `error_description` field with a human-readable error message

I also would like to do this on a follow-up PR. 


> I've never contributed to Kiali and am very new to this project, but I can help you with all of that if you want. Would you mind if I do a PR on your branch in the next few days?

Pull requests are very welcome!
Looks like you have reviewed the code of this PR and your review is very welcome!. After fixing the issue with the nonce validation, I would like to merge this PR because I think it will be good enough after that. Then, open the PR directly in kiali/kiali repository. We will prefer that you open the PR directly against kiali/kiali :) The whole team will be happy to provide guidance for the developer setup. Although if you really prefer to open a PR over my branch, that's also fine!

sandhose:
@israel-hdez 

> Kiali doesn't have back-end storage. So, we anyway would need to store the id_token in the cookie sent to the browser. I understand that one of the main ideas of the `authorization code flow` is to never expose the tokens to the browser. Without back-end storage, it doesn't make a lot of sense to implement it.

The authorization code flow does not mean you have to store anything. The overhaul flow stays the same: instead of posting the id_token to the backend, you can pass the code and the backend does the token exchange.

> It's being verified in https://github.com/kiali/kiali/pull/2798/files#diff-7b14eebb8414820b40c718efcf921867R240-R250. Although, that validation is going to change.

Not really. Notice the `ParseUnverified`: any token generated by any party would work. If you really want to use the implicit flow, the ID token should be checked with the keys published at the JWKS URI. More info here: https://auth0.com/docs/tokens/guides/validate-jwts

> If I understand correctly, the Implicit Flow should always return the parameters via the hash portion of the URL, because it takes into account JavaScript applications. Since this PR is using Implicit Flow, I think it should be fine.

I did not get that part of the flow was done on the frontend side. I saw the other PR, it makes more sense now :)

> Pull requests are very welcome!
> Looks like you have reviewed the code of this PR and your review is very welcome!. After fixing the issue with the nonce validation, I would like to merge this PR because I think it will be good enough after that. Then, open the PR directly in kiali/kiali repository. We will prefer that you open the PR directly against kiali/kiali :) The whole team will be happy to provide guidance for the developer setup. Although if you really prefer to open a PR over my branch, that's also fine!

If you want to merge that PR fast, please please fix at least the token verification first: in the current state, it is really insecure and lets anyone to login as any user.

jmazzitelli:
> it is really insecure and lets anyone to login as any user.

I put a block on this PR so it will not be merged as-is. We absolutely must NOT merge code that lets anyone login as any user. @sandhose we'd highly welcome any additional changes/PRs to this code that solves the problems you mention.

util/random.go
israel-hdez: I added the special characters. Not randomizing the order, though.

config/config.go
```diff
@@ -302,6 +316,15 @@ func NewConfig() (c *Config) {
 		},
 		Auth: AuthConfig{
 			Strategy: "login",
+			OpenId: OpenIdConfig{
+				AuthenticationTimeout: 300,
+				AuthorizationEndpoint: "",
+				ClientId:              "",
+				InsecureSkipVerifyTLS: false,
+				IssuerUri:             "",
+				Scopes:                []string{"profile", "email"},
```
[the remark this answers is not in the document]
israel-hdez: Done!

util/httputil/http_util.go
israel-hdez: Fixed. Now there are settings to configure all parts of the public URL

business/openid_auth.go
israel-hdez: Fixed.

business/openid_auth.go
israel-hdez: Fixed.

business/openid_auth.go
israel-hdez: Fixed.

business/openid_auth.go
israel-hdez: fixed.

israel-hdez:
@sandhose 

> 
> The authorization code flow does not mean you have to store anything. The overhaul flow stays the same: instead of posting the id_token to the backend, you can pass the code and the backend does the token exchange.
> 

With regards to exchanging the Authorization Code that's right. No back-end storage needed. But once the code is exchanged with a token, we do need back-end storage. The spec says *"This provides the benefit of not exposing any tokens to the User Agent and possibly other malicious applications with access to the User Agent."* (see https://openid.net/specs/openid-connect-core-1_0-final.html#CodeFlowAuth). So, the idea is to never expose the exchanged id_token to the user agent. Without back-end storage, I don't think it's possible to do this part, so it doesn't makes sense to implement the Authorization Code Flow --- but if you have a clever way to do this without back-end storage, you should contribute on that :)

> Not really. Notice the `ParseUnverified`: any token generated by any party would work. If you really want to use the implicit flow, the ID token should be checked with the keys published at the JWKS URI. More info here: https://auth0.com/docs/tokens/guides/validate-jwts
> 
> If you want to merge that PR fast, please please fix at least the token verification first: in the current state, it is really insecure and lets anyone to login as any user.

The verification of the nonce code has changed. (see https://github.com/kiali/kiali/pull/2798/files#diff-7b14eebb8414820b40c718efcf921867R255-R265). But it has the same pattern about using `ParseUnverified`.

In my opinion, I don't think there is the need to verify the signature of the id_token. This PR is assuming that the cluster is also configured with OpenId connect. The lines following the nonce code check are doing an API call to the cluster (see https://github.com/kiali/kiali/pull/2798/files#diff-7b14eebb8414820b40c718efcf921867R270-R282). The cluster should do it's job of properly verifying the id_token and it's signature. If it's invalid, it should return an error, and Kiali will just reject the login. I don't think that can be bypassed using an arbitrary token with an invalid signature.

util/random.go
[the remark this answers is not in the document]
israel-hdez: I added the special characters and also did the change about sending a hash to the IdP instead of the plain nonce code.

sandhose:
> In my opinion, I don't think there is the need to verify the signature of the id_token. This PR is assuming that the cluster is also configured with OpenId connect. The lines following the nonce code check are doing an API call to the cluster (see https://github.com/kiali/kiali/pull/2798/files#diff-7b14eebb8414820b40c718efcf921867R270-R282). The cluster should do it's job of properly verifying the id_token and it's signature. If it's invalid, it should return an error, and Kiali will just reject the login. I don't think that can be bypassed using an arbitrary token with an invalid signature.

Oh, you're right. I did not get that part when I looked through the code: I thought it was just implementing OIDC login in Kiali, not that it then uses the token got from the provider to do API calls. If the token is verified by the k8s API server it's definitely good enough ; sorry about that!

util/httputil/http_util_test.go
```diff
@@ -0,0 +1,111 @@
...
+	assert.Equal(t, "https://kiali/custom/kiali", guessedUrl)
+}
+
+func TestGuessKialiURLPrioritizesConfig(t *testing.T) {
+	request := setupAndCreateRequest()
+
+	conf := config.NewConfig()
+	conf.Server.WebRoot = "/foo/bar"
+	conf.Server.WebFQDN = "subdomain.domain.dev"
+	conf.Server.WebSchema = "http"
```
jmazzitelli, line 102:
```suggestion
	// do not give Kiali an identity which by definition turns off https
	conf.Identity.CertFile = ""
	conf.Identity.PrivateKeyFile = ""
```

util/httputil/http_util.go
```diff
@@ -84,3 +86,58 @@ func AuthTransport(auth *config.Auth, transportConfig *http.Transport) (http.Rou
...
+	cfg := config.Get()
+
+	// Take default values from configuration
+	schema := cfg.Server.WebSchema
```
jmazzitelli, line 94:
```suggestion
	schema := "http"
	// if Kiali has been given an identity, then we know it will be using https (since it is via https that its identity gets known)
	if cfg.Identify.CertFile != "" && cfg.Identity.PrivateKeyFile != "" {
		schema = "https"
	}
```

util/random.go
```diff
@@ -1,17 +1,56 @@
...
+	cryptoRand "crypto/rand"
+	mathRand "math/rand"
+)
 
 // RandomString generates a random string of length n. Before calling this function, you should call
 // rand.Seed() to initialize the default source.
 //
 // Found at https://ispycode.com/Blog/golang/2016-10/How-to-generate-a-random-string-of-a-fixed-length
+//  - Adapted for more characters
 func RandomString(n int) string {
```
jotak, line 13: I just checked the current usage of this `RandomString`, and it's used for generating signing keys. So I wonder, what's the point of having a secured version and a less secured version? I think we should keep only one version, obviously the stronger one.
We can do it in a separate PR though.

util/httputil/http_util.go
israel-hdez: For explanation, see comment: https://github.com/kiali/kiali-operator/pull/42#discussion_r433946709

I think the identity setting configures the protocol that Kiali uses to listen, but that is not necessarily the public-facing protocol.

util/random.go
israel-hdez: Yes, I introduced the `RandomString` just for tests -- all usages are in unit tests. OpenId spec requires cryptographic secure randomness, so I had to introduce the other version.

I agree that we can remove this "less secure" one in another PR. Although this is supposed to be faster than the secure one, I don't think the added time to tests will be noticeable.

jshaughn: Helpful to just comment code that is explicitly a test hook.

israel-hdez:
Requesting QA.

israel-hdez:
@jmazzitelli "sub" claim check.added in the new commit

jmazzitelli:
> "sub" claim check.added in the new commit

Works. I see this in the logs when I try to spoof another user name:

```
W0603 19:10:36.736799       1 authentication.go:553] Kiali token rejected because of subject claim mismatch
```

israel-hdez:
@sandhose All is merged.
About the enhancements you would like to see, can you open an issue with your suggestions so that we can discuss there?
I really would like to hear how we can strengthen this OpenID mechanism.

PR #2804, opened 2020-05-19
[hack] move to CRC 0.10.0 / OpenShift 4.4

jmazzitelli:
This also adds some more information in the status output - uptime, memory usage (via free), and cpu usage (via top).


jmazzitelli:
This is working for me. I can deploy upstream Istio 1.6 and Kiali from master build (using `make cluster-push operator-create kiali-create`)

Issue #1424, opened 2019-08-26
Configurable thresholds for error rates


lucasponce:
As a user I want to be able to configure the error rate thresholds so that they display an appropriate status in health.

I want also to be able to configure different thresholds depending on the service. Ex: service A may still be considered as OK with 5% errors whereas service B is not OK above 1%.

JIRA:
https://issues.jboss.org/browse/KIALI-579

jshaughn:
@lucasponce I'm somewhat surprised no one has upvoted this but perhaps this can be closed given that it is stale and no one has asked for it.  Maybe users are happy enough with the fixed settings.

lucasponce:
@jshaughn +1
Let's re-open if we have more feedback about it.

kenske:
I'd be very interested in having this functionality. It is normal and expected for one of our APIs to return a high percentage of 4xx errors (around 12%). Right now this is always marked as being over the fixed threshold. If we ever do get an abnormal amount, it won't be easily detectable.

jotak:
@kenske I guess that in this situation you would like to make a difference between 4xx and 5xx errors, no? Like having one threshold for 4xx and another for 5xx and they are checked separately?
Also, would it be on a per-service basis?

kenske:
@jotak Correct, having around 12% 4xx errors for our API service is fine, but having any 5xx errors would be a problem. For other services in our cluster, having anything over 1% for 4xx would be a problem.

lucasponce:
@kenske thanks for the feedback :-).
Having a comment from users is enough reason to keep it open.

jshaughn:
I have nothing against a WIP to help flesh out ideas but I think this is a pretty big topic that can probably benefit from more up-front exposure and definition.  Moving from two simple, fixed values to a fully configurable namespace/object/errorCode granularity is pretty big.  We need to make sure that changes here are applied consistently across the affected areas.  For example, the thresholds are not used explicitly for health calculation but also have shared use in the graph when determining edge color.

A couple of notes/questions:
- Will there still be global settings like what we have today, as the defaults?  Meaning, are the detailed configs applied as overrides?
- Will it be possible to inspect/configure the settings from the UI?
- Remember that not everything is HTTP and different protocols have different error codes (or no error codes for TCP).
[...]

PR #2807, opened 2020-05-20
Configurable thresholds for error rates

aljesusg:
Fixes https://github.com/kiali/kiali/issues/1424

Actual Config sample

```yaml
thresholds_health:
  - namespace: bookinfo
    threshold_checks:
      - rule: ">20"
        kind: "all"
        label: "response_code"
        filter:
          - regex: "views"
            label_filter:
              - labels:
                    app: "details"
                    version: "v1"
                operation: "and"
            prom_filter:
              - labels:
                    destination_service: "reviews"
        exp: "400>x"
        alert: "warning"
```

- Namespace is the ns to apply: 
  - We can set this value to "**" or "" to set all or a list of namespaces "bookinfo,istio-system"

For each alert

- rule => When I want more than a x% (>, >=, <=, <, == or !=)
- Kind can be "**" or ""(all) | "services" | "workloads" | "apps" or list "services,workloads"
- metric: metric to check in prometheus, by default response_code
- filter(optional) =>
  - We can filter by a reg for resource name
  - labels that the resource has
  - Working in optProm => filter by a parameter of metric promethus like destination_source: "reviews-v2"
- expression => status code operation than  (x<400 , 400>x, 300<x<400)
- alert => kind of alert

Example:

Warning alert when in the Last X minutes of user the requests with status code > 400 are more than 20%

PR Docs: https://github.com/aljesusg/kiali/blob/improve_configuration_custom/models/threshold/Threshold.adoc

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -11,6 +11,7 @@ import (
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
+	"github.com/kiali/kiali/models/threshold"
 	"github.com/kiali/kiali/prometheus"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
 	"github.com/kiali/kiali/status"
@@ -29,8 +30,8 @@ func (in *HealthService) GetServiceHealth(namespace, service, rateInterval strin
 	promtimer := internalmetrics.GetGoFunctionMetric("business", "HealthService", "GetServiceHealth")
 	defer promtimer.ObserveNow(&err)
 
-	rqHealth, err := in.getServiceRequestsHealth(namespace, service, rateInterval, queryTime)
-	return models.ServiceHealth{Requests: rqHealth}, err
+	rqHealth, thresholds, err := in.getServiceRequestsHealth(namespace, service, rateInterval, queryTime)
+	return models.ServiceHealth{Requests: rqHealth, Thresholds: thresholds}, err
 }
 
 // GetAppHealth returns an app health from just Namespace and app name (thus, it fetches data from K8S and Prometheus)
@@ -56,7 +57,6 @@ func (in *HealthService) GetAppHealth(namespace, app, rateInterval string, query
 
 func (in *HealthService) getAppHealth(namespace, app, rateInterval string, queryTime time.Time, ws models.Workloads) (models.AppHealth, error) {
 	health := models.EmptyAppHealth()
-
 	// Perf: do not bother fetching request rate if there are no workloads or no workload has sidecar
 	fetchRate := false
 	for _, w := range ws {
@@ -69,8 +69,9 @@ func (in *HealthService) getAppHealth(namespace, app, rateInterval string, query
 	// Fetch services requests rates
 	var errRate error
 	if fetchRate {
-		rate, err := in.getAppRequestsHealth(namespace, app, rateInterval, queryTime)
+		rate, thresholdAux, err := in.getAppRequestsHealth(namespace, app, rateInterval, queryTime)
 		health.Requests = rate
+		health.Thresholds = thresholdAux
 		errRate = err
 	}
 
@@ -105,10 +106,11 @@ func (in *HealthService) GetWorkloadHealth(namespace, workload, rateInterval str
 		}, nil
 	}
 
-	rate, err := in.getWorkloadRequestsHealth(namespace, workload, rateInterval, queryTime)
+	rate, threshold, err := in.getWorkloadRequestsHealth(namespace, workload, rateInterval, queryTime)
 	return models.WorkloadHealth{
 		WorkloadStatus: status,
 		Requests:       rate,
+		Thresholds:     threshold,
 	}, err
 }
 
@@ -287,17 +289,19 @@ func fillWorkloadRequestRates(allHealth models.NamespaceWorkloadHealth, rates mo
 	}
 }
 
-func (in *HealthService) getServiceRequestsHealth(namespace, service, rateInterval string, queryTime time.Time) (models.RequestHealth, error) {
+func (in *HealthService) getServiceRequestsHealth(namespace, service, rateInterval string, queryTime time.Time) (models.RequestHealth, []models.Threshold, error) {
 	rqHealth := models.NewEmptyRequestHealth()
+	srv, _ := in.businessLayer.Svc.GetServiceDefinition(namespace, service)
 	inbound, err := in.prom.GetServiceRequestRates(namespace, service, rateInterval, queryTime)
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
 	}
-	return rqHealth, err
+	return rqHealth, generateThreshold(inbound, namespace, service, "service", srv.Service.Labels), err
 }
 
-func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval string, queryTime time.Time) (models.RequestHealth, error) {
+func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval string, queryTime time.Time) (models.RequestHealth, []models.Threshold, error) {
 	rqHealth := models.NewEmptyRequestHealth()
+	appInfo, _ := fetchNamespaceApps(in.businessLayer, namespace, app)
 	inbound, outbound, err := in.prom.GetAppRequestRates(namespace, app, rateInterval, queryTime)
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
@@ -305,11 +309,12 @@ func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval strin
 	for _, sample := range outbound {
 		rqHealth.AggregateOutbound(sample)
 	}
-	return rqHealth, err
+	return rqHealth, generateThreshold(inbound, namespace, app, "app", getLabelsApp(appInfo[app].Workloads, appInfo[app].Services)), err
 }
 
-func (in *HealthService) getWorkloadRequestsHealth(namespace, workload, rateInterval string, queryTime time.Time) (models.RequestHealth, error) {
+func (in *HealthService) getWorkloadRequestsHealth(namespace, workload, rateInterval string, queryTime time.Time) (models.RequestHealth, []models.Threshold, error) {
 	rqHealth := models.NewEmptyRequestHealth()
+	wInfo, _ := in.businessLayer.Workload.GetWorkload(namespace, workload, false)
 	inbound, outbound, err := in.prom.GetWorkloadRequestRates(namespace, workload, rateInterval, queryTime)
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
@@ -317,7 +322,7 @@ func (in *HealthService) getWorkloadRequestsHealth(namespace, workload, rateInte
 	for _, sample := range outbound {
 		rqHealth.AggregateOutbound(sample)
 	}
-	return rqHealth, err
+	return rqHealth, generateThreshold(inbound, namespace, workload, "workload", wInfo.Labels), err
 }
 
 func castWorkloadStatuses(ws models.Workloads) []models.WorkloadStatus {
@@ -332,4 +337,11 @@ func castWorkloadStatuses(ws models.Workloads) []models.WorkloadStatus {
 
 	}
 	return statuses
+}
+
+func generateThreshold(requests model.Vector, ns string, srv string, kind string, labels map[string]string) models.Thresholds {
+	thAlerts := threshold.FilterBy(config.Get().ThresholdsHealth, ns, srv, kind, labels)
+	thresholds := models.Thresholds{}
+	thresholds.Parse(thAlerts, &requests, "service")
+	return thresholds
 }
--- a/business/health.go
+++ b/business/health.go
@@ -296,7 +296,7 @@ func (in *HealthService) getServiceRequestsHealth(namespace, service, rateInterv
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
 	}
-	return rqHealth, generateThreshold(inbound, namespace, service, "service", srv.Service.Labels), err
+	return rqHealth, generateThreshold(inbound, model.Vector{}, namespace, service, "service", srv.Service.Labels), err
 }
 
 func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval string, queryTime time.Time) (models.RequestHealth, []models.Threshold, error) {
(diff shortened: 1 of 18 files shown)


models/threshold.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/prometheus/common/model"
+
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/models/threshold"
+)
+
+
+type Thresholds []Threshold
+// Customize Thresholds error rates
+type Threshold struct {
```
lucasponce, line 16: About naming, better "ThresholdHealth" for consistency wutg ServiceHealth, AppHealth, and others.

models/threshold.go
```diff
@@ -0,0 +1,99 @@
...
+	Percent int   `json:"percent"`
+	Rule    string    `json:"rule"`
+	Alert   string    `json:"alert"`
```
lucasponce, line 19: I don't think "alert" is good terminology, it has a different meaning.
If it's related to the "condition" of "properties" of the "ThresholdHealth" that would be understood better.

config/config.go
```diff
@@ -265,6 +265,27 @@ type DeploymentConfig struct {
 // defaults to the namespace configured for IstioNamespace (which itself defaults to 'istio-system').
 type IstioComponentNamespaces map[string]string
 
+
+// Threshold configuration
+type Threshold struct {
```
lucasponce, line 270: ThresholdHealth sounds better description what represents I think.
Threshold is too generic.

config/config.go
```diff
@@ -265,6 +265,27 @@ type DeploymentConfig struct {
...
+	Namespace string `yaml:"namespace"`
+	ThresholdAlerts    []ThresholdAlert  `yaml:"threshold_alerts"`
+}
+
+type ThresholdAlert struct {
```
lucasponce, line 275: ThresholdAlert -> ThresholdCondition or ThresholdProperties.

Alert is a complex term and using it here it may create amgibuity.

config/config.go
```diff
@@ -265,6 +265,27 @@ type DeploymentConfig struct {
...
+	Rule       string   `yaml:"rule"`
+	Kind       string   `yaml:"kind"`
+	Metric     string   `yaml:"metric"`
+	Filter     ThresholdFilter `yaml:"filter,omitempty"`
+	Expression string   `yaml:"exp"`
+	Alert      string   `yaml:"alert"`
```
lucasponce, line 281: May you document in comments how to use these fields ?
That would help me to review better the tests.

jotak:
Inline comments:

> Actual Config sample
> 
> ```yaml
> thresholds:
>   - namespace: bookinfo
>     threshold_alerts:
>       - rule: ">20"
>         kind: "all"
>         metric: "response_code"
>         filter:
>           - reg: "views"
>             labels:
>               - app: "details"
>         expression: "400>x"
>         alert: "warning"
> ```
> 
>     * Namespace is the ns to apply:
>       
>       * We can set this value to "**" or "" to set all or a list of namespaces "bookinfo,istio-system"
> 
> 
> For each alert
> 
>     * rule => When I want more than a x% (>, >=, <=, <, == or !=)
> 
>     * Kind can be "**" or ""(all) | "services" | "workloads" | "apps" or list "services,workloads"

I wonder if this field is really useful (at least unless someone asks for it). I would imagine that in most cases, if not always, we will leave it to "all", no?

> 
>     * metric: metric to check in prometheus, by default response_code

This should be named `label` rather than `metric` (response_code is a label)

> 
>     * filter(optional) =>
>       
>       * We can filter by a reg for resource name
>       * labels that the resource has
>       * Working in optProm => filter by a parameter of metric promethus like destination_source: "reviews-v2"
> 
>     * expression => status code operation than  (x<400 , 400>x, 300<x<400)

in the case of response_code, `expression` will deal with numbers indeed... but if you consider other labels such as the envoy's response_flags, it's not going to be a number comparison. So I suggest that `expression` is a regex instead. Eg: `[4|5]..` (for response code) or `UO|FI` (for response flags).

> 
>     * alert => kind of alert
> 
> 
> Example:
> 
> Warning alert when in the Last X minutes of user the requests with status code > 400 are more than 20%

aljesusg:
> Inline comments:
> 
> > Actual Config sample
> > ```yaml
> > thresholds:
> >   - namespace: bookinfo
> >     threshold_alerts:
> >       - rule: ">20"
> >         kind: "all"
> >         metric: "response_code"
> >         filter:
> >           - reg: "views"
> >             labels:
> >               - app: "details"
> >         expression: "400>x"
> >         alert: "warning"
> > ```
> > 
> > 
> > ```
> > * Namespace is the ns to apply:
> >   
> >   * We can set this value to "**" or "" to set all or a list of namespaces "bookinfo,istio-system"
> > ```
> > 
> > 
> > For each alert
> > ```
> > * rule => When I want more than a x% (>, >=, <=, <, == or !=)
> > 
> > * Kind can be "**" or ""(all) | "services" | "workloads" | "apps" or list "services,workloads"
> > ```
> 
> I wonder if this field (`Kind`) is really useful (at least unless someone asks for it). I would imagine that in most cases, if not always, we will leave it to "all", no?
> 
> > ```
> > * metric: metric to check in prometheus, by default response_code
> > ```
> 
> This (`metric`) should be named `label` rather than `metric` (response_code is a label)
> 
> > ```
> > * filter(optional) =>
> >   
> >   * We can filter by a reg for resource name
> >   * labels that the resource has
> >   * Working in optProm => filter by a parameter of metric promethus like destination_source: "reviews-v2"
> > 
> > * expression => status code operation than  (x<400 , 400>x, 300<x<400)
> > ```
> 
> in the case of response_code, `expression` will deal with numbers indeed... but if you consider other labels such as the envoy's response_flags, it's not going to be a number comparison. So I suggest that `expression` is a regex instead. Eg: `[4|5]..` (for response code) or `UO|FI` (for response flags).
> 
> > ```
> > * alert => kind of alert
> > ```
> > 
> > 
> > Example:
> > Warning alert when in the Last X minutes of user the requests with status code > 400 are more than 20%

Thanks for comments I am still working about that, I was thinking about to do this for all the labels but first por resposne code, for this is the regex like you see :) let me finish the PR and I'll ping you

aljesusg:
Added some docs to help in https://github.com/aljesusg/kiali/blob/improve_configuration_custom/models/threshold/Threshold.adoc

jotak:
@aljesusg could you please answer to my previous comments? ( https://github.com/kiali/kiali/pull/2807#issuecomment-634691680 ). Especially about "expression".
I think the current implementation with "expression" will quickly show some limitations and perhaps having just a regexp here would be sufficient and handle more usages.

business/health.go
```diff
@@ -287,37 +289,40 @@ func fillWorkloadRequestRates(allHealth models.NamespaceWorkloadHealth, rates mo
...
+	appInfo, _ := fetchNamespaceApps(in.businessLayer, namespace, app)
 	inbound, outbound, err := in.prom.GetAppRequestRates(namespace, app, rateInterval, queryTime)
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
 	}
 	for _, sample := range outbound {
 		rqHealth.AggregateOutbound(sample)
 	}
-	return rqHealth, err
+	return rqHealth, generateThreshold(inbound, namespace, app, "app", getLabelsApp(appInfo[app].Workloads, appInfo[app].Services)), err
```
jotak, line 312: Thresholds are checked only for inbound here ... I think it makes sense to define some thresholds only for inbounds, but then we'd also like to define some thresholds for outbounds. In other words, we could have something similar to your "kind", but for traffic direction: either "inbound", "outbound" or "both".

business/health.go
```diff
@@ -333,3 +338,10 @@ func castWorkloadStatuses(ws models.Workloads) []models.WorkloadStatus {
 	}
 	return statuses
 }
+
+func generateThreshold(requests model.Vector, ns string, srv string, kind string, labels map[string]string) models.Thresholds {
+	thAlerts := threshold.FilterBy(config.Get().ThresholdsHealth, ns, srv, kind, labels)
+	thresholds := models.Thresholds{}
+	thresholds.Parse(thAlerts, &requests, "service")
```
jotak, line 345: I guess "service" should be `kind` here?

aljesusg:
> @aljesusg could you please answer to my previous comments? ( [#2807 (comment)](https://github.com/kiali/kiali/pull/2807#issuecomment-634691680) ). Especially about "expression".
> I think the current implementation with "expression" will quickly show some limitations and perhaps having just a regexp here would be sufficient and handle more usages.

@jotak your comments are done, for response_code we have a specific regular expression but for other labels we'll use the input like regex like you comment.

business/health.go
aljesusg: Yes this was last commit. I forgot it sorry

config/config.go
```diff
@@ -281,6 +281,33 @@ type DeploymentConfig struct {
...
 
+// Threshold configuration
+type ThresholdHealth struct {
+	Namespace       string           `yaml:"namespace"`
+	ThresholdChecks []ThresholdCheck `yaml:"threshold_checks"`
+}
+
+type ThresholdCheck struct {
+	Rule       string          `yaml:"rule"`
+	Kind       string          `yaml:"kind"`
```
jotak, line 292: About naming, "Kind" is a very generic word so I don't like it very much but I fail to find something better :-) If anyone has a good idea ...

config/config.go
```diff
@@ -281,6 +281,33 @@ type DeploymentConfig struct {
...
+	Label      string          `yaml:"label"`
+	Filter     ThresholdFilter `yaml:"filter,omitempty"`
+	Expression string          `yaml:"exp"`
```
jotak, line 295: For consistent names between code & yaml, it should be `expression`

config/config.go
```diff
@@ -281,6 +281,33 @@ type DeploymentConfig struct {
...
+	Alert      string          `yaml:"alert"`
+}
+
+type ThresholdFilter struct {
+	Regex       string                `yaml:"regex,omitempty"`
+	LabelFilter *ThresholdLabelFilter `yaml:"label_filter,omitempty"`
+	OptProm     *ThresholdLabelFilter `yaml:"prom_filter,omitempty"`
```
jotak, line 302: same here, code not consistent with yaml

models/threshold/Threshold.adoc
```diff
@@ -0,0 +1,83 @@
...
+=== Filter configuration
+
+We can filter the Threshold by:
+
+- Regex: Match the resource name by this regex
+- LabelFilter: Filter by labels of the resource
+- PromFilter: Filter by Prometheus Labels of each request
+
+
+==== LabelFilter and PromFilter configuration
```
jotak, line 56: I'm must say I'm not very confident about that part. I understand the goal is to have some powerful matching capabilities on labels, but I've the feeling that writing complex matching rules exploded in yaml structures, with potentially several levels of nested objects, could turn into a bad experience. I'd like to hear others about it. Perhaps it would need further discussion / brainstorming .. Perhaps more research including about if there's some DSL parsers that already exist in the wild that we could reuse here ?

jotak:
> > @aljesusg could you please answer to my previous comments? ( [#2807 (comment)](https://github.com/kiali/kiali/pull/2807#issuecomment-634691680) ). Especially about "expression".
> > I think the current implementation with "expression" will quickly show some limitations and perhaps having just a regexp here would be sufficient and handle more usages.
> 
> @jotak your comments are done, for response_code we have a specific regular expression but for other labels we'll use the input like regex like you comment.

Hmm I see .. but that makes special cases in code, with some quite complicated parts, for something that IMO could have been entirely managed by regex quite simply, no?

aljesusg:
> > > @aljesusg could you please answer to my previous comments? ( [#2807 (comment)](https://github.com/kiali/kiali/pull/2807#issuecomment-634691680) ). Especially about "expression".
> > > I think the current implementation with "expression" will quickly show some limitations and perhaps having just a regexp here would be sufficient and handle more usages.
> > 
> > 
> > @jotak your comments are done, for response_code we have a specific regular expression but for other labels we'll use the input like regex like you comment.
> 
> Hmm I see .. but that makes special cases in code, with some quite complicated parts, for something that IMO could have been entirely managed by regex quite simply, no?

The reported issue was create threshold for use cases like have response_code greater than X, here is the problem.

models/threshold/Threshold.adoc
aljesusg: My idea is have a powerful filtering here :) I think that the most complex filter will have 2 levels
AND
--> Labels with OR
--> Labels with OR

You can use a regex pattern for or values in the same label too so I didn't expect a huge yaml config for this. But yes we can can discuss about this and have a brainstorming session

jotak:
> The reported issue was create threshold for use cases like have response_code greater than X, here is the problem.

As put in my previous comment we can handle these use cases via regex so no need for a special case. Something like `/4../` for all 4xx codes, for instance, and so on.

aljesusg:
> > The reported issue was create threshold for use cases like have response_code greater than X, here is the problem.
> 
> As put in my previous comment we can handle these use cases via regex so no need for a special case. Something like `/4../` for all 4xx codes, for instance, and so on.

mmm I understand now xD +1 sorry monday morning I need an extra coffee

models/threshold/Threshold.adoc
```diff
@@ -0,0 +1,83 @@
+= ThresholdHealth
+:toc: macro
+:toc-title:
+
+== Introduction
+
+Kiali provides a customize configuration for ThresholdAlert from requests in your environment.
```
lucasponce, line 7: First comment about naming.

"ThresholdAlert" is not correct. 
The goal of this feature is to extract the health logic to Kiali configuration in a global way, so a user could change it.

But using terms like "alert", "rule", "threshold" creates confusion with an alerting engine, which is not the goal of this feature.

Perhaps it's better the think on this like "HealthConfig", about "rules", "threshold" I will continue reading the whole PR before to suggest different naming but I understand some of concerns from @jshaughn related naming as these wording is used in differente environments.

models/threshold/Threshold.adoc
```diff
@@ -0,0 +1,83 @@
+= ThresholdHealth
```
lucasponce, line 1: I don't like to have a separated .adoc for this.

This should be defined at least in two places:
- Where all info related configuration is placed (typically in operator Kiali CR descriptor).
- Also, in Kiali.io where I guess this user-friendly style is very good (and also we would need to move the rest of config in that way).

models/threshold/Threshold.adoc
lucasponce: I'm in the middle of the review, but I share some of the concerns as well.

I also understand that this feature can be very open, but as it is, it could be seen like a proto-alerting-engine and that's not really the goal.

The idea is to have a first step to extract implicit hardcoded logic used to calculate health in a way the user can modify globally from the config.

But I haven't yet finished to review completely the PR I need to understand better the approach.

lucasponce:
I tried to summarize my comments here:
https://github.com/kiali/kiali/issues/1424#issuecomment-642573600

Let's maintain the do not merge on this PR, there are good ideas there but I think we need to combine them with more context as discussed in the issue.

Let's discuss there.

models/threshold/Threshold.adoc
aljesusg: Ok I'll create another PR for Doc after discussion about what do do it with his

lucasponce:
I think main focus is being developed on new PR #2915 
I'll close this for clarity.

Both PRs are linked and we can refer to this one for tracking all discussion and comments.

PR #2811, opened 2020-05-21
[hack] we want to give CRC 5 CPUs by default since 4 isn't enough

jmazzitelli:
If we install everything (istio/kiali/bookinfo) then we need at least 5 CPUs, so make that the default. You can always ask for less (`-ccpus`) if you want.

This PR also outputs the # of  CPUs allocated to the VM in the status.


jmazzitelli:
With 4 CPUs, I had Istio 1.6, Kiali, and bookinfo+traffic generator installed. The Traffic Generator pod failed to start with the error `0/1 nodes are available: 1 Insufficient cpu`. Thus, we need 5 CPUs minimum.

Issue #2633, opened 2020-04-08
Enhance Iter8 Experiment details page


lucasponce:
Once we have a created an Iter8 experiment, Kiali offers a basic details page:

![image](https://user-images.githubusercontent.com/1662329/78799487-7d393300-79ba-11ea-84d7-a5b226822af3.png)

The goal of the issue is to discuss with @kiali/iter8 team how we want to engage this page:
- What interesting info we want to add.
- Do we want to add this info at Service Details levels too ? New tab, or some link to the Iter8 experiment details page (as it happens with Istio config).
- Today we have implemented a basic create/delete actions. What actions make sense for an existing experiment ? Is a modification a valid scenarios or better to approach with a cancel/delete/new experiment flow ?

The idea is to first scope the feature itself (information to display, actions to perform, relationship with other entities).

When feature is almost defined, then we can involve UX team to help us to better design it in terms of usability.

I know it's an open topic, so, we are seeking a ready-action-pull-request to send, but to share ideas and iterate on prototypes on it.

jadeyliu939:
Prototype from the iter8 team on the experiment detail page. ( This is view only design, no actions yet)

**Tab 1: Overview** - contains detail information about the experiment, and a request volume chart.
The chart will show live view while the experiment is running. and once the experiment had ended, it will show a snapshot view with start/end time of the experiment as long there are prometheus data available. 

_- Live View_
<img width="1634" alt="details-LiveView" src="https://user-images.githubusercontent.com/4615187/78841221-d27d4080-79ca-11ea-8bd1-86062fd3efd8.png">

-_SnapShot View_ 
<img width="1636" alt="detail-snapshot" src="https://user-images.githubusercontent.com/4615187/78841247-de690280-79ca-11ea-8081-a7a4c273f905.png">


**Tab 2: Criteria** 

View - Collapsed
<img width="1639" alt="criteria" src="https://user-images.githubusercontent.com/4615187/78841312-0b1d1a00-79cb-11ea-9438-1df90636dc01.png">

View - Expanded 

<img width="1630" alt="criteria-expand" src="https://user-images.githubusercontent.com/4615187/78841326-14a68200-79cb-11ea-9cc8-c7ce985adf2e.png">

lucasponce:
@jadeyliu939 for this isssue I think it's ready.
I'm going to bring it to the UX meeting next Tuesday.
I will add you in the invitation via email.

lucasponce:
[...]

PR #2815, opened 2020-05-21
New Feature - Experiment Action: Pause, Resume, Terminate with status

jadeyliu939:
** Describe the change **

Adding new action for Iter8 Experiment
- Pause
- Resume
- Terminate with success status
- Terminate with failure status

** Issue reference **
[Main issue #2007](https://github.com/kiali/kiali/issues/2007)
and 
[Issue #2633](https://github.com/kiali/kiali/issues/2633)

** Changes **
Introduce a new Iter8 Endpoint for update experiment with action.
Change the Iter8 Experiment CRD -- add new field "action"

** UI Changes **
The new action will require changes in the Kiali-UI [PR#1805](https://github.com/kiali/kiali-ui/pull/1805)


lucasponce:
Travis CI reports swagger needs to be updated due changes in the handlers.

business/iter8.go
```diff
@@ -145,6 +148,37 @@ func (in *Iter8Service) CreateIter8Experiment(namespace string, body []byte) (mo
 	return iter8ExperimentDetail, nil
 }
 
+func (in *Iter8Service) UpdateIter8Experiment(namespace string, name string, body []byte) (models.Iter8ExperimentDetail, error) {
+	var err error
+	promtimer := internalmetrics.GetGoFunctionMetric("business", "Iter8Service", "CreateIter8Experiment")
```
lucasponce, line 153: CreateIter8Experiment -> UpdateIter8Experiment to collect metrics for this method

business/iter8.go
```diff
@@ -145,6 +148,37 @@ func (in *Iter8Service) CreateIter8Experiment(namespace string, body []byte) (mo
...
+	newExperimentSpec.Parse(experiment)
+	newExperimentSpec.Action = action.Action
+	var newObject []byte
+	newObject, err = json.Marshal(newExperimentSpec)
+	json, err := in.ParseJsonForCreate(newObject)
+	if err != nil {
+		return iter8ExperimentDetail, err
+	}
+
+	iter8ExperimentObject, err := in.k8s.UpdateIter8Experiment(namespace, name, string(json))
```
lucasponce, line 173: The kubernetes/UpdateIter8Experiment is a generic patch which is good.
But perhaps the business/UpdateIter8Experiment should be also more specific.
Please @jadeyliu939 correct me if I'm wrong, but basically this business/UpdateIter8Experiemt only apply a change in the action, isn't it ?

Then, perhaps at business package level could be renamed or focused for just actions ?

Also, I'm still reading the whole PR to understand the other changes, perhaps I can resolve my question later.

business/iter8.go
```diff
@@ -172,7 +206,18 @@ func (in *Iter8Service) ParseJsonForCreate(body []byte) (string, error) {
 	object.Spec.TrafficControl.MaxTrafficPercentage = newExperimentSpec.TrafficControl.MaxTrafficPercentage
 	object.Spec.TrafficControl.MaxIterations = newExperimentSpec.TrafficControl.MaxIterations
 	object.Spec.TrafficControl.TrafficStepSize = newExperimentSpec.TrafficControl.TrafficStepSize
+	object.Spec.TrafficControl.Interval = newExperimentSpec.TrafficControl.Interval
 	object.Spec.Analysis.AnalyticsService = "http://iter8-analytics.iter8:" + strconv.Itoa(in.GetAnalyticPort())
+	rr, err := in.GetIter8RoutingReferences(newExperimentSpec.Namespace, newExperimentSpec.Service)
```
lucasponce, line 211: I wouldn't make any call inside a Parse() method.
it can be added as parameter, but calls should be done from the business method, otherwise we are mixing concerns and then it's harder to follow.

business/iter8.go
```diff
@@ -241,3 +289,90 @@ func (in *Iter8Service) GetAnalyticPort() int {
...
+	gwNames := make([]string, 0)
+	if err != nil {
+		return routingReferences, err
+	}
+
+	Gateways.Parse(gws)
+	for _, gw := range gws {
+		gwNames = append(gwNames, gw.GetObjectMeta().Name)
+	}
+	vs, err := in.k8s.GetVirtualServices(namespace, servicename)
```
lucasponce, line 309: I think this can be done in a better way.
I mean, here the Iter8Service is duplicating basically a feature that is dont in the IstioConfigService.
As the Iter8Service has access to the Layer object, it can access to that service and query the VirtualServices or others objects getting advantage of the caching and parsing.

Then the Iter8Service should only consume that service.

Let me know if it makes sense, I can also help to propose a PR if perhaps there is some difficulty but I guess if you check the "Layer" type probably you will get examples about what you want to do.

handlers/iter8.go
```diff
@@ -49,6 +49,19 @@ func Iter8ExperimentGet(w http.ResponseWriter, r *http.Request) {
...
 	}
+	workloadList, err := business.Workload.GetWorkloadList(namespace)
+	if err != nil {
+		handleErrorResponse(w, err)
+		return
+	}
+	workloads := workloadList.Workloads
+	for _, w := range workloads {
+		if w.Name == experiment.ExperimentItem.Baseline {
+			experiment.ExperimentItem.BaselineVersion = w.Labels["version"]
```
lucasponce, line 60: Note, Istio can change this label, so in Kiali code we use the config to fetch this value:

```
version := workload.Labels[conf.IstioLabels.VersionLabelName]
```

business/iter8.go
jadeyliu939: Yes, currently it is only a change in the action. But I am not sure if in the future, there will be other `apply` possible. So i kind of left it generic.

business/iter8.go
jadeyliu939: understand. thank you

business/iter8.go
jadeyliu939: yes, it makes sense. I just didn't realize the existing capabilities.

jadeyliu939: @lucasponce I am replacing the GetGateway and GetVirtualServices with the following code section. Is this what you have in mind? 
```
istioCfg, err := in.businessLayer.IstioConfig.GetIstioConfigList(IstioConfigCriteria{
		IncludeGateways : true,
		IncludeVirtualServices:  true,
		Namespace:               namespace,
	})
```

lucasponce: @jadeyliu939 yes, using the IstioConfig API enables the caching.
Thanks !

jadeyliu939:
@prachiyadav Thank you for the info. 

Can you check if the iter8-analytics service is running and on which port `kubectl get svc iter8-analytics -n iter8`. And also check what image version the analytics is using. 
As for the Action, the resume should be grey-out. I will fix that.

jadeyliu939:
@prachiyadav The action `resume` did work, but since the network problem still exists, it failed again. My previous comment that the resume should be grey-out is wrong, the resume should still be there, and once the network glitch is gone, user can resume.

lucasponce:
Hi @jadeyliu939,
I'd like to progress with this feature and consolidate it on master for this Sprint.

I see that the branch is mixed with a lot commits, I wonder if you could rebase with master and have a PR with just the commits for the feature.

I can give a final review and merge if no blockers.

I think we can refine/fix when it's consolidated

jadeyliu939:
Close this PR base on suggestion from @lucasponce.  A new PR will be created asap.

jadeyliu939:
NEW PR created [#2907 ](https://github.com/kiali/kiali/pull/2907)

Issue #2783, opened 2020-05-14
Deleted k8s services are persisted within the Graph UI


robert-blackman:
**Describe the bug**
When a service is deleted from kubernetes, it still remains within the Graph UI. We are running  kiali across multiple environments and are seeing this behaviour in each of them. This behaviour was observed only once traffic was generated for the service.

Service doesn't exist
![kiali_no-svc](https://user-images.githubusercontent.com/57333677/81878202-b0c73880-95ca-11ea-9b5c-4992d0d40b6f.png)

![kiali_service_list](https://user-images.githubusercontent.com/57333677/81877194-cbe47900-95c7-11ea-97a9-34fd4496a6c3.png)

Service still appears in UI (_**Indefinitely**_)
![kiali_phantom_service](https://user-images.githubusercontent.com/57333677/81876987-511b5e00-95c7-11ea-947f-59423983b815.png)

**Versions used**
Kiali: 1.18.0-edge
Istio: 1.5.1
Kubernetes flavour and version: AWS EKS 1.15

**Additional Info**
Mutual TLS Enabled
Automatic Sidecar Injection Enabled
<details open>
<summary>DEBUG</summary>

```
{
  "backendConfigs": {
    "authenticationConfig": {
      "strategy": "login"
    },
    "computedServerConfig": {
      "extensions": {
        "threescale": {
          "enabled": false
        },
        "iter8": {
          "enabled": false
        }
      },
      "istioStatusEnabled": true,
      "istioIdentityDomain": "svc.cluster.local",
      "istioNamespace": "istio-system",
      "istioComponentNamespaces": {
        "grafana": "istio-system",
        "pilot": "istio-system",
        "prometheus": "istio-system",
        "tracing": "istio-system"
      },
      "istioLabels": {
        "appLabelName": "app",
        "versionLabelName": "version"
      },
      "prometheus": {
        "globalScrapeInterval": 15,
        "storageTsdbRetention": 21600
      },
      "durations": {
        "60": "1m",
        "300": "5m",
        "600": "10m",
        "1800": "30m",
        "3600": "1h",
        "10800": "3h",
        "21600": "6h"
      }
    }
  },
  "currentURL": "https://<REMOVED>/kiali/console/graph/namespaces/?edges=noEdgeLabels&graphType=versionedApp&namespaces=kiali-test&unusedNodes=false&injectServiceNodes=true&duration=60&refresh=15000&layout=dagre",
  "reduxState": {
    "authentication": {
[...]

PR #2830, opened 2020-05-26
Fix dead node corner case

jshaughn:
The injected service node was not being killed when:
- Service injection on
- Both the source and dest nodes were dead:
  - No traffic between the source and dest
  - No backing workloads for the source and dest

Fixes: #2783 

Before:
image

After:
image

### DIFF
--- a/graph/telemetry/istio/appender/dead_node.go
+++ b/graph/telemetry/istio/appender/dead_node.go
@@ -102,13 +102,26 @@ func (a DeadNodeAppender) applyDeadNodes(trafficMap graph.TrafficMap, globalInfo
 		return
 	}
 
-	for _, s := range trafficMap {
+	for _, n := range trafficMap {
 		goodEdges := []*graph.Edge{}
-		for _, e := range s.Edges {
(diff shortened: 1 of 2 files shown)


jshaughn:
@pbajjuri20 This is a little bit difficult to replicate:
- Deploy bookinfo with traffic generator
- Wait for typical graph to form (display service nodes should be enabled)
- Cycle traffic generator to 0 pods
- Wait for graph to go gray
- Delete the ProductPage deployment

See the orphan productpage service node.  (actually, you may also need to delete the ingressgateway deployment)

With the fix you should no longer see the node.  Let me know if you have any issues verifying the fix.

pbajjuri20:
I have reproduced the issue and Tested it with this PR. The issue has been fixed perfectly. 

Tested and Passed on Kiali UI 1.19.0 (d5e1ed641f9d98ea6c6d9715577ac4ab83256e89)

pbajjuri20:
I have reproduced the issue and Tested it with this PR. The issue has been fixed perfectly.

Issue #2732, opened 2020-05-05
Improve separation of concerns in tracing / service details


jotak:
- [x] In UI, service details class is too big, has multiple purposes. It fetches tons of stuff from backend before dispatching to sub-components. It should probably be possible that the specialized subcomponents fetch what they need themselves: https://github.com/kiali/kiali-ui/pull/1800
- [ ] Tracing fetching params are mixed between UI params (for Kiali) and backend params (for Jaeger), which resulted in some troubles when we had to make a distinction. They should be separate objects & types.

jotak:
As part of this task, make sure the backend endpoint to fetch traces suits well for consumption for in-graph tracing (it should)

jotak:
Related: https://github.com/kiali/kiali/issues/2720

jotak:
It's a big work so i'm going to split in two PRs: one for service details, another for tracing params

jotak:
(half-done)

PR #2835, opened 2020-05-27
Separate jaeger queries and kiali tracing queries

jotak:
Using TracingQuery when kiali is queried, decoupled from jaeger query API.
Share more code between spans.go and traces.go

Fixes https://github.com/kiali/kiali/issues/2732

Frontend PR required: https://github.com/kiali/kiali-ui/pull/1810

### DIFF
--- a/jaeger/spans.go
+++ b/jaeger/spans.go
@@ -6,47 +6,32 @@ import (
 	"net/url"
 	"path"
 	"sort"
-	"strconv"
 	"time"
 
 	jaegerModels "github.com/jaegertracing/jaeger/model/json"
 
-	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/log"
+	"github.com/kiali/kiali/models"
 )
-
-// TODO / Question: make limit configurable? Selected from UI?
-const tracesLimit = 100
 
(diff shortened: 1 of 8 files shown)


jaeger/spans.go
```diff
@@ -6,47 +6,32 @@ import (
...
 	jaegerModels "github.com/jaegertracing/jaeger/model/json"
 
-	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/log"
+	"github.com/kiali/kiali/models"
 )
 
-// TODO / Question: make limit configurable? Selected from UI?
-const tracesLimit = 100
-
```
aljesusg, line 20: This was for the first check that we do when the user go to the service details. We fetch the last 100 traces to check if there is an error in the tab (https://github.com/kiali/kiali-ui/blob/master/src/pages/ServiceDetails/ServiceDetailsPage.tsx#L121) where we fetch https://github.com/kiali/kiali/blob/master/jaeger/traces.go#L48 . The point is taht I forgot remove this because we fetch by tag error :( So good one

jotak: no, it's my old code :) i was using it for spans

PR #2839, opened 2020-05-28
[hack] fix up some things for minikube that are broken

jmazzitelli:
1. need to specify insecure registry when starting minikube now
2. default to the most stable release of k8s
3. make should not try to delete templates (k8s doesn't have that type and we don't use templates anymore anyway)
4. make should only try to delete openshift resources when cluster type is openshift
5. xdg-open is deprecated; use gio open now

### DIFF
--- a/hack/k8s-minikube.sh
+++ b/hack/k8s-minikube.sh
@@ -7,7 +7,8 @@
 # The typical order of commands used is the following:
 #   up - starts the Kubernetes cluster via minikube
 #   istio - installs Istio using Kiali's install hack script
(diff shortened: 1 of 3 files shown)


jmazzitelli:
asking @israel-hdez to review because I need these changes to test his OpenID feature :)

jmazzitelli:
do not merge yet, something else I want to add.

jmazzitelli:
OK, now its ready to merge (I just wanted to add a simple thing to the cluster-status make target and a new command "podman" to the k8s-minikube hack for those that use podman)

PR #2845, opened 2020-05-28
molecule make target to add to the container host table

jmazzitelli:
Some clusters have hosts that are not resolveable within the molecule docker container so add some smarts to the makefile to try to pass the IPs into the container's host table.


jmazzitelli:
I was going to try to get this to not blow up when CLUSTER_TYPE=minikube, but these tests don't work with minikube anyway, and if they did, you can always just pass in MOLECULE_ADD_HOST_ARGS="" so avoid the make target blowing up.

Issue #995, opened 2019-04-11
Feature request: Envoy Filter


Buri:
**Describe the bug**
Istio has support for [EnvoyFilter](https://istio.io/docs/reference/config/networking/v1alpha3/envoy-filter/), but Kiali has not. Is it possible to add type EnvoyFilter to config section, so it is possible to edit it from UI?

**Versions used**
Kiali: 0.16.2
Istio: 1.1.1
Kubernetes flavour and version: AWS EKS v1.11.8-eks-7c34c0

**To Reproduce**
1. Go to 'Istio Config'
2. Click on 'Istio type selector'
3. EnvoyFilter is not present

**Expected behavior**
Be able to select EnvoyFilter

pilhuhn:
Thank you for your suggestion. Displaying the presence of such filters and potentially their details may be "easily" doable. 
How do you envision the editing - especially as the actual filter config seems not to be simple yaml or json to just show, but some more complex data type.
Would you be able to help us on this?

Buri:
I think that for now it would be enough to just show yaml editor for now, especially since there is some ongoing discussion over at Istio about changes that are coming to Envoy Filters (Envoy is deprecating some functionality on their end).

israel-hdez:
@Buri Looks like the team has been busy :)
If you have the time to prepare a PR to implement this, we will be more than happy to incorporate your feature.

I see that "Gateways" have the same behavior you describe about showing the YAML with no overview page, so the involved code could bring some guidance about how to implement this.

Buri:
Might aswell give it a shot :)

lucasponce:
Thanks @Buri, indeed the EnvoyFilter is the missing resource to be added to Kiali.

If you like to work on this feature, please let me know and I can assign this issue to you.

Also happy to help you to jump into the project, it should be easy to implemented as there are only needed a couple of changes in backend/frontend.

Please, feel free to contact us at anytime and we can help you on this.

josejulio:
@Buri If you are still interested in adding support for `EnvoyFilter`, you can check the PRs that added support for the `Sidecar` resource.

https://github.com/kiali/kiali/pull/1108
https://github.com/kiali/kiali/pull/1130
https://github.com/kiali/kiali-ui/pull/1230

lucasponce:
Hi @Buri, I'm reviewing this issue.
Would you like to work on it ?
Please, let us know and we can help to jump into the project.

lucasponce:
[...]

PR #2853, opened 2020-06-02
Add EnvoyFilter,AttributeManifest,HttpApiSpec{Binding}

lucasponce:
This PRs adds support to missing Istio Config objects.

EnvoyFilter was missing from Networking API.
AttributeManifest,HttpApiSpec and HttpApiSpecBindings are from mixer and probably when mixer is completely out we can remove it but as those APIs are present in Istio 1.6 crds, to me it looked weird not to add them to have minimal view/edit capabilities on them.

This requires an UI PR to match these new objects.
https://github.com/kiali/kiali-ui/pull/1813

If it looks good it should be backported to 1.18 to reach Kiali 1.18.2 and offer complete support of Istio config.

Fixes https://github.com/kiali/kiali/issues/995


business/istio_config.go
```diff
@@ -662,6 +722,34 @@ func (in *IstioConfigService) GetIstioConfigDetails(namespace, objectType, objec
...
+		} else {
+			err = iErr
+		}
+	case HttpApiSpecs:
+		if hs, iErr := in.k8s.GetHttpApiSpec(namespace, object); iErr == nil {
+			istioConfigDetail.HttpApiSpec = &models.HttpApiSpec{}
+			istioConfigDetail.HttpApiSpec.Parse(hs)
+		} else {
+			err = iErr
+		}
```
xeviknal, line 752: I believe we do want to stll users to modify this new objects using, right?
If so, I guess, there is need to update the method `modifyIstioConfigDetail` (around L#92)

lucasponce:
@xeviknal I've added support for update but those objects are not yet supported for create.
Let me know if this solves the issue.

PR #2871, opened 2020-06-05
[hack] some scripts we can use for testing

jmazzitelli:
1. hack/istio/install-travel-agency-demo.sh
* Another microservice demo with added benefit of a couple defined operations in telemetry.
2. hack/jwt-[en,de]code.sh
* Two scripts that let us test passing in and examining JWT tokens. Useful to make sure hacking JWT tokens is not possible. Also useful to pass in when testing Kiali APIs as authenticated users.

I used these hack scripts for several things. First to confirm the correctness of our new OpenID integration. Also the demo let's us test the graph as well as producing telemetry with classified metric data (the operation ID stuff).

### DIFF
--- a/hack/istio/install-travel-agency-demo.sh
+++ b/hack/istio/install-travel-agency-demo.sh
@@ -1,0 +1,261 @@
+#/bin/bash
+
+# This deploys the travel agency demo
+
+: ${CLIENT_EXE:=oc}
+: ${NAMESPACE_AGENCY:=travel-agency}
+: ${NAMESPACE_PORTAL:=travel-portal}
+: ${ENABLE_OPERATION_METRICS:=false}
+
+while [ $# -gt 0 ]; do
+  key="$1"
+  case $key in
+    -c|--client)
+      CLIENT_EXE="$2"
+      shift;shift
+      ;;
+    -eo|--enable-operation-metrics)
+      ENABLE_OPERATION_METRICS="$2"
+      shift;shift
+      ;;
+    -na|--namespace-agency)
+      NAMESPACE_AGENCY="$2"
+      shift;shift
+      ;;
+    -np|--namespace-portal)
+      NAMESPACE_PORTAL="$2"
+      shift;shift
+      ;;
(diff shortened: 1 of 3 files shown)


jmazzitelli:
putting @lucasponce and/or @israel-hdez as reviewers - they are familiar with the travel agency demo and the jwt scripts respectively.

hack/istio/install-travel-agency-demo.sh
```diff
@@ -0,0 +1,261 @@
...
+kind: NetworkAttachmentDefinition
+metadata:
+  name: istio-cni
+EOF
+fi
+
+# Deploy the demo
+
+${CLIENT_EXE} apply -f <(curl -L https://raw.githubusercontent.com/lucasponce/travel-comparison-demo/master/travel_agency.yaml) -n ${NAMESPACE_AGENCY}
+${CLIENT_EXE} apply -f <(curl -L https://raw.githubusercontent.com/lucasponce/travel-comparison-demo/master/travel_portal.yaml) -n ${NAMESPACE_PORTAL}
```
lucasponce, line 85: This is nice.
I wonder if we could add some parameter to also install the version 2 in an optional manner (Step [2.5.0]):
https://lucasponce.github.io/presentations/observability-in-action.html

hack/istio/install-travel-agency-demo.sh
```diff
@@ -0,0 +1,261 @@
...
++                     }
++                   }]
+                 }
+               root_id: stats_inbound
+               vm_config:
+EOF
+
+cat <<EOF | ${CLIENT_EXE} -n istio-system apply -f -
+apiVersion: networking.istio.io/v1alpha3
+kind: EnvoyFilter
```
lucasponce, line 119: Nice EnvoyFilters.

hack/istio/install-travel-agency-demo.sh
jmazzitelli: added that to the script

Issue #2181, opened 2020-01-30
Reorganize documentation for kiali.io


lucasponce:
Today Kiali holds documentation for kiali.io in a separate repo and it presents some challenges:

- Hard to maintain a documentation updated when there are changes on the features.
- Impossible to track which feature is included in which version, there is a de-facto, it refers to "latest".

These two issues are serious and make difficult for everyone to contribute in an efficient way to doc.

So, this issue proposes an EPIC to define:

- How documentation content can be separated from the html template and hosted together with the code.
- With that in mind, in the same PR where a modification is done, reviewer can ask to modify the content of the documentation, without wait to open a new issue and update a separate task (that in practice, it happens very few times and badly).
- Also, as a side effect, when Kiali performs a release, content is versioned too, and it would be simplified how to track documentation in a better way.

It's clear that this affects in how site is built but I think is worth to put priority on this.

- [x] Versioned content
- [x] Add section for security notifications https://github.com/kiali/kiali/issues/2536
- [x] Add section for release information [notes, changes, etc] https://github.com/kiali/kiali/issues/2536
- [x] Investigate Google Season of Docs https://github.com/kiali/kiali/issues/2619

jotak:
I have a comment about config documentation: currently, we just point to Kiali CR (ex: https://github.com/kiali/kiali/blob/c78f195fcbe214e56a918d616ed2f241be24c968/operator/deploy/kiali/kiali_cr.yaml#L433-L470 ) when we want to refer to configuration. I see two problems with that:

- This is not pretty. I think people could expect more than just yaml comments
- We can either link to `master` (latest) or to a specific commit (like the link I put above). Both are problematic: links to master can be inaccurate if we put anchors to specific lines, when master changes ; and links to commits wouldn't show the most up-to-date content.

So, as part of this task, I think it would be nice if we could start thinking about having a better documented configuration, integrated into kiali.io.

jshaughn:
This is an RH internal link, but is an [internal review of the web site](https://docs.google.com/document/d/1KrGbE9rW_XA9Ph9pDNMajinQ2L9ZjVqCc44PwCZCCWA/edit?ts=5e1dcbfd#heading=h.l1s9y5ya55qe).

aljesusg:
I started the poc about versions in https://github.com/kiali/kiali.io/pull/142 so time ago but I moved it by other tasks I can resume this one

PR #2879, opened 2020-06-08
Prepare CI for releasing the website

cfcosta:
Needs major testing, but it works for me.

### DIFF
--- a/deploy/jenkins-ci/Jenkinsfile
+++ b/deploy/jenkins-ci/Jenkinsfile
@@ -61,9 +61,9 @@
  *   - SKIP_OPERATOR_RELEASE
  *      defaultValue: n
  *      description: Set to 'y' if you don't want to release the operator
- *   - SKIP_SITE_RELEASE
- *      defaultValue: n
- *      description: Set to 'y' if you don't want to release the website
+ *   - RELEASE_SITE
+ *      defaultValue: auto
+ *      description: Set to 'y' to force release of the website even for patch versions
  *   - UI_VERSION
  *       defaultValue: ''
  *       description: If you are skipping UI release. Specify the UI version to package, or leave
@@ -81,7 +81,7 @@ node('kiali-build') {
   def buildUi = params.SKIP_UI_RELEASE != "y"
   def buildBackend = params.SKIP_BACKEND_RELEASE != "y"
   def buildOperator = params.SKIP_OPERATOR_RELEASE != "y"
-  def buildSite = params.SKIP_SITE_RELEASE != "y"
+  def buildSite = (params.RELEASE_SITE == "auto" && params.RELEASE_TYPE == "minor") || params.RELEASE_SITE != "y"
   def quayTag = ""
 
   try {
--- a/deploy/jenkins-ci/Jenkinsfile
+++ b/deploy/jenkins-ci/Jenkinsfile
@@ -13,6 +13,9 @@
  *   - UI_RELEASING_BRANCH
  *      defaultValue: refs/heads/master
  *      description: Branch of the UI to release
+ *   - SITE_RELEASING_BRANCH
+ *      defaultValue: refs/heads/master
+ *      description: Branch of the website to release
  *   - OPERATOR_RELEASING_BRANCH
  *      defaultValue: refs/heads/master
  *      description: Branch of the operator to release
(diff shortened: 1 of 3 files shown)


cfcosta:
Depends on https://github.com/kiali/kiali.io/pull/247

jshaughn:
@cfcosta I defer to @israel-hdez for this review.  I really don't know anything about the jenkins automation.  My only question is whether there is github issue for this work, so that we know what the PR is actually supposed to accomplish.

cfcosta:
Closes #2181.

deploy/jenkins-ci/Jenkinsfile
```diff
@@ -126,6 +139,27 @@ node('kiali-build') {
...
+            submoduleCfg: [],
+            userRemoteConfigs: [[
+            credentialsId: 'kiali-bot-gh-ssh',
+            url: params.SITE_GITHUB_URI]]
+        ])
+
+          sh "cd ${siteDir}; git config user.email 'kiali-dev@googlegroups.com'"
+          sh "cd ${siteDir}; git config user.name 'kiali-bot'"
+          sh "ls src"
+          sh "ls src/github.com"
```
israel-hdez, line 160: Perhaps remove these `ls` commands that won't be useful for the final code?

deploy/jenkins-ci/Makefile.site.jenkins
```diff
@@ -0,0 +1,60 @@
...
+website-push-version-tag:
+ifeq ($(PUSH_CHANGES),y)
+	git push $(SITE_GITHUB_URI) $$(git rev-parse HEAD):refs/tags/v$(SITE_VERSION)
+ifdef GH_TOKEN
+	@echo "Creating GitHub release..."
+	@curl -H "Authorization: token $(GH_TOKEN)" \
+	  -H "Content-Type: application/json" \
+	  -d '{"name": "Kiali.io $(SITE_VERSION)", "tag_name": "$(VERSION_TAG)"}' \
+	  -X POST $(SITE_RELEASE_URI)
+endif
```
israel-hdez, line 55: I think we don't need to create releases for kiali.io repository
It's OK to create them, but I'd prefer to not :)

deploy/jenkins-ci/Makefile.site.jenkins
```diff
@@ -0,0 +1,60 @@
...
+
+.PHONY: website-build-archive website-push-version-tag
+.PHONY: all release
+
+all:
+	$(error You must explicitly specify a target)
+
+website-build-archive:
+ifneq ($(PUSH_CHANGES),y)
```
israel-hdez, line 34: isn't this conditional inverted?

deploy/jenkins-ci/Jenkinsfile
```diff
@@ -52,6 +61,9 @@
  *   - SKIP_OPERATOR_RELEASE
  *      defaultValue: n
  *      description: Set to 'y' if you don't want to release the operator
+ *   - SKIP_SITE_RELEASE
+ *      defaultValue: n
+ *      description: Set to 'y' if you don't want to release the website
```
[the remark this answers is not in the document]
cfcosta: The idea was to allow the user to decide to even run or not the website release, for tests and so on.

Glad to remove if you see no value in having this.

cfcosta:
> As a side comment, this is becoming too big IMO and starts to become a little hard to maintain. I think we should break this up in three parts: kiali app (back and front-ends), operator and website parts. I'll create issues for breaking this in parts hoping that this can be prioritized... cc @jshaughn .

Agree, that could be the next step (and I think we could also simplify the release process itself, now that operator is its own thing, we can make things simpler).

deploy/jenkins-ci/Jenkinsfile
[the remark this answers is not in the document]
israel-hdez: I was thinking this should be a variable accepting 3 piossible values?
* `auto` (the default) to do the current behavior
* `y` to force releasing the site release regardless of the RELEASE_TYPE
* `n` to force skipping the site release regardless of the RELEASE_TYPE

jshaughn:
> As a side comment, this is becoming too big... cc @jshaughn

I trust you guys to do the right thing here.  Proceed any way you both agree on.

israel-hdez:
I created the issue for splitting this big pipeline in smaller parts: https://github.com/kiali/kiali/issues/2895

Let's move forward this PR as is and work later on splitting/simplifying.

cfcosta:
@israel-hdez as it turns out, if you run two different tasks from a Makefile, it gets evaluated twice. The version is bumped on the first release, and then re-evaluated when pushing the tags. Should be ok now.

PR #2881, opened 2020-06-09
Refactor kubernetes package

lucasponce:
This PR have several goals:
- Remove technical debt around legacy mixer objects. In the earlier versions where any adapter has its own CRD Kiali created a virtual sub-type to handler this in the adapter/template category. This is now clean and removed using handlers/instances.
- Reorganice a little bit kubernetes package, trying to use a single API to fetch IstioObjects when possible.
- A couple of renamings and better distribution of interfaces.

This requires UI PR:
https://github.com/kiali/kiali-ui/pull/1822


lucasponce:
@jotak @xeviknal I'm performing this refactoring to simplify the legacy mixer objects and also to improve (a little bit) the kubernetes package.

It's not yet ready, but I put you on cc on this as it may touch several areas and introduce some conflict but I think it will help in long term.

lucasponce:
@hhovsepy last commit was missed for 3scale problems.
Let me know if the UI works now as expected.

business/istio_config.go
```diff
@@ -193,7 +125,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			if kialiCache != nil && kialiCache.CheckIstioResource(kubernetes.GatewayType) && kialiCache.CheckNamespace(criteria.Namespace) {
 				gg, ggErr = kialiCache.GetIstioResources(kubernetes.GatewayType, criteria.Namespace)
 			} else {
-				gg, ggErr = in.k8s.GetGateways(criteria.Namespace)
+				gg, ggErr = in.k8s.GetIstioObjects(criteria.Namespace, kubernetes.Gateways, "")
```
jotak, line 128: Some day golang will reach modern age and have generics :-)

business/istio_config.go
```diff
@@ -285,18 +217,40 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
...
+				errChan <- hhErr
+			}
+		}
+	}(errChan)
+
+	go func(errChan chan error) {
+		defer wg.Done()
+		if criteria.IncludeInstances {
+			if ii, iiErr := in.k8s.GetIstioObjects(criteria.Namespace, kubernetes.Instances, ""); iiErr == nil {
+				istioConfigList.Instances = models.CastIstioInstancesCollection(ii)
```
jotak, line 243: I can see sometimes we use `CastXXX` and other times `xxx.Parse`, I believe this is two different ways to do the same thing? If so, that's something we could refactor too? (here or another PR)

Personally I prefer the `CastXXX` way as it handles object initialization.

business/istio_config.go
lucasponce: Yeap, there are missing features in Golang I miss, too, like to differentiate signatures by arguments, etc.

business/istio_config.go
lucasponce: Yes, there is some technical debt in that area, due mixed objects.
At it's not really urgent, I left that untouched until we have some strong changes in that area and we can apply it.

handlers/istio_config.go
```diff
@@ -331,7 +202,6 @@ func IstioConfigCreate(w http.ResponseWriter, r *http.Request) {
 	params := mux.Vars(r)
 	namespace := params["namespace"]
 	objectType := params["object_type"]
-	objectSubtype := params["object_subtype"]
```
jotak, line 334: I can see a bunch of remaining `object_subtype` in code, mostly related to swagger / doc, I guess they should go away too?
(Just searched for "object_subtype" in all files)

kubernetes/istio.go
```diff
@@ -0,0 +1,663 @@
...
+		return *in.isMixerDisabled
+	}
+
+	meshConfig, err := in.GetIstioConfigMap()
+	if err != nil {
+		log.Warningf("IsMixerDisabled: Cannot read Istio mesh configuration.")
+		return true
+	}
+
+	log.Infof("IsMixerDisabled: %t", meshConfig.DisableMixerHttpReports)
```
jotak, line 457: It should be log.Trace or Debug I guess

kubernetes/istio.go
```diff
@@ -0,0 +1,663 @@
+package kubernetes
```
jotak, line 1: I haven't carefully reviewed all the business logic here but I guess this result from moving code without functional changes, right?

handlers/istio_config.go
lucasponce: Ops, you are right, those should be cleaned as well.

kubernetes/istio.go
lucasponce: +1

kubernetes/istio.go
lucasponce: Right. There are a couple of functional changes: adding Instances/Handlers as plain Istio config. But most of the refactor is just to simplify the kubernetes package using shared methods when possible.

Also a couple of renaming changes that were always there and I think it makes sense to simplify for better separation of concerns between the package.

lucasponce:
@hhovsepy @jotak addressed Joel's comments.

Let me know if it's ok to merge.

hhovsepy:
Re-verified.

PR #2899, opened 2020-06-11
Remove unnecessary "istio namespace" prom queries in graph generation

### LINKED ISSUE #2891: Remove unnecessary "istio namespace" prom queries in graph generation
In telem V2, and actually I think in later versions of mixer, Istio telemetry no longer generates intra-istio service traffic telemetry.  In graph generation we had special-case support to capture this telemetry when the user selected the istio-system namespace (or whichever ns was configured for the control plane).

We can remove this support, simplify the code, and make the slight perf improvement.

jshaughn:
With istiod (maybe even before given the overhead) Istio no longer reports  telemetry between its components. It also no longer offers component deployment across namespaces. This removes a bunch of special case code dealing with the internal istio time-series.  Note that things like the ingress and egress gateways are unaffected.

Fixes #2891


graph/types.go
```diff
@@ -13,6 +13,7 @@ const (
 	GraphTypeVersionedApp string = "versionedApp"
 	GraphTypeWorkload     string = "workload"
 	NodeTypeApp           string = "app"
+	NodeTypeOperation     string = "operation" // The special request_operation aggregate node
 	NodeTypeService       string = "service"
```
jmazzitelli, line 17: This is unrelated to this PR (which should just be about removing the component namespace stuff) so this should be reverted. This "operation" thing is for the upcoming work on the experimental metric classification stuff.

jshaughn: good catch, looks like I had PR spillage... Fixed.

jshaughn:
@mattmahoneyrh This is a code removal, just looking to ensure no graph regressions, with and without the istio namespace selected.

Issue #2897, opened 2020-06-11
Display throughput on the graph edges


sebcrozet:
**Is your feature request related to a problem? Please describe.**
My distributed system follows a producer-consumer logic. In order to identify the bottlenecks, I need to be able to visualize the throughput of each service (in kB/sec, the request rate is not relevant in my case).

**Describe the solution you'd like**
Right now Kiali allows the display of the number of requests per second on the graph edges. An option to display the throughput (kB per second) instead would be great.

**Describe alternatives you've considered**
I can go into the details of each service to see their inbound/outbound metrics but this makes it difficult to get a global view of the system.

jshaughn:
@sebcrozet Note that the request rates for HTTP/GRPC edges are requestsPerSecond but for TCP edges the request rates are in bytesPerSecond (for bytes sent).  

Does that help you in any way?  Or are you trying to get response data size information?

jotak:
@jshaughn this is indeed something that we don't query from prometheus but it's available to us. That's valid for graph stats as well as for metrics details pages.

Currently we only use the `istio_request_bytes` metric to get histogram distribution, which gives us the average / median / percentiles of request sizes. However these metrics can also provide throughput:

- For histogram distribution, as we currently do, it's queried as something like: `histogram_quantile(0.99, sum(rate(istio_request_bytes_bucket[1m])) by (le))`

- For throughput, we could query something like: `rate(istio_request_bytes_sum[1m])`

So, +1 for me to add these stats, both in graph and metrics pages

jotak:
I did a quick capture of how it would be in Metrics page:

![Capture d’écran de 2020-06-15 14-38-50](https://user-images.githubusercontent.com/2153442/84658974-fe5efa00-af16-11ea-853b-038f77fcad1a.png)

jshaughn:
@jotak This could be done for graph edges by adding a new appender and a display option to toggle its use (the same as we do for response time).  Let's also wait for @sebcrozet to be specific about request/response data throughput.  When it's well-defined we can add to the backlog.  Of course a contribution would be welcome as well.

lucasponce:
I think this is valid.
Adding an optional appender to show request and response throughput in the graph to be in parity of metrics info is useful.

Perhaps that may need also to update the sidepanel charts that are also focused on operations per sec in http+gRPC.

TCP traffic already provides a similar info.

jshaughn:
@sebcrozet Are you still out there?  Is this something you would still like to see?

[...]

PR #2912, opened 2020-06-17
Add rq/rs throughput in metrics page

jotak:
Related to https://github.com/kiali/kiali/issues/2897

Fix / add tests

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -111,11 +111,29 @@ func getIstioCharts(durationMillis bool) []istioChart {
 		},
 		{
 			Chart: kmodel.Chart{
+				Name:  "Request throughput",
+				Unit:  "bitrate",
+				Spans: 6,
+			},
+			refName: "request_throughput",
+			scale:   8, // Bps to bps
+		},
+		{
+			Chart: kmodel.Chart{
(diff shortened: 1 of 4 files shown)


jotak:
Screenshot:
Capture d’écran de 2020-06-15 14-38-50



@lucasponce this is a small and interesting addition to add rq/rs throughput to the list of metrics.

Throughput adds value from requests per second in several situations ; for instance when streaming data or high volume of data per request, it can be an interesting metric.

lucasponce:
@jotak It looks great to me

lucasponce:
I've found something that is not related to this PR but perhaps worth to take a look.
From this PR I get the metrics:
image

But when I enable the spans:
image

The scale changes.

It seems correct, because:
image
image

The peak is shown in a very short scale when a single metric.

Anyway, the PR looks ok, just commenting this in case it may help for other refinment in the charts.

jotak:
Thanks @lucasponce 
About this spike that disappears: yet, this is a side-effect of having the y-axis autoscaled. A possibility could be to have y-axis always 0-based. It may not make sense for every metrics in general, but for the istio metrics that we display here, I think it should be fine.

PR #2915, opened 2020-06-18
Configurable thresholds for error rates

aljesusg:
Related with https://github.com/kiali/kiali/pull/2807

Fix https://github.com/kiali/kiali/issues/1424
Require: https://github.com/kiali/kiali-operator/pull/82
UI: https://github.com/kiali/kiali-ui/pull/1831
Config

```go
// ToleranceConfig
type Tolerance struct {
	Code     string `yaml:"code,omitempty"`
	Degraded int    `yaml:"degraded,omitempty"`
	Failure  int    `yaml:"failure,omitempty"`
}

// RateConfig
type Rate struct {
	Namespace       string    `yaml:"namespace,omitempty"`
	Kind            string    `yaml:"kind,omitempty"`
	RequestType     string    `yaml:"request_type,omitempty"`
	RequestProtocol string    `yaml:"request_protocol,omitempty"`
	Name            string    `yaml:"name,omitempty"`
	Tolerance       Tolerance `yaml:"tolerance,omitempty"`
}

// HealthConfig
type HealthConfig struct {
	Rate []Rate `yaml:"rate,omitempty"`
}
```

This let us to add in the future another Healths like tracing one

health Return

```go
// RequestHealth holds several stats about recent request errors
type RequestHealth struct {
	InboundRequestCode  map[string]map[string]float64 `json:"inboundRequestsCode"`
	OutboundRequestCode map[string]map[string]float64 `json:"outboundRequestsCode"`
}
```


jotak:
My understanding of Lucas' position was that we wouldn't have so much info in HealthConfig. No rule/regexp/code/alerttype , but something much more limited like just "warningThreshold" and "failureThreshold", as 2 numbers, and that's it. Is it what you thought @lucasponce  ?

aljesusg:
> My understanding of Lucas' position was that we wouldn't have so much info in HealthConfig. No rule/regexp/code/alerttype , but something much more limited like just "warningThreshold" and "failureThreshold", as 2 numbers, and that's it. Is it what you thought @lucasponce ?

I thought that we talked about some minimum filter...so Is this going to apply to all resources @lucasponce ?

lucasponce:
Let me put some examples, perhaps I can better detail my previous comments.

From the Health config, I'd expect that user may define something in this way:

```
    healthConfig:
        rate:
        - tolerance:
            degraded: 5%
            failure: 10%
        - namespace: default
          tolerance:
            degraded: 10%
            failure: 15%
        - namespace: bookinfo
          app: reviews
          tolerance:
            code: 4xx
            degraded: 2%
            failure: 5%
```

This info is available in the external config and it may be used but the UI to calculate the health.
First case, may apply to all namespaces (except if there is a specific one), second example is the most clear to me, for a given namespace `default` I want to raise the tolerance for `degraded` and `failure` cases, the third case is less clear, but possible because Kiali backend provides health at several levels: namespace, app, workload and service, and it may personalize for a specific app and/or code.

To me it's open to discuss if here we may or may not want to introduce a simple regex for names (both in app/service/workload or code).

To implement this, the only change I expect from the health is to send request rate aggregated (as it is today) or splitted:

```
        ErrorRatio         float64 `json:"errorRatio"`
	InboundErrorRatio  float64 `json:"inboundErrorRatio"`
	OutboundErrorRatio float64 `json:"outboundErrorRatio"`

        CodeErrorRatio         map[string]float64 `json:"codeErrorRatio"`
	CodeInboundErrorRatio  map[string]float64 `json:"codeInboundErrorRatio"`
	CodeOutboundErrorRatio map[string]float64 `json:"codeOutboundErrorRatio"`
```

Backend shouldn't need to calculate anything extra, just send this info, as today the health is calculated in the UI (both for details and also for graph).

To me is up to discuss if on this aggretation we want to put together all 4xx, 5xx in two values, or allow them to be individually (i'd rather prefer the group of 4xx vs 5xx, but I think here it can be a minor deal to play with both cases).

Then, the UI has access to the server config, and today the main change is that instead to perform a `errorRatio>0` it may add more edges for these cases.

jotak:
Also, perhaps if we provide through the Health not a ratio, but the absolute number of requests, grouped by error code (or error code family) then we remove more processing from the backend and we leave that to the frontend. This way we can also get rid of `ErrorRatio` / `CodeErrorRatio` because it can be computed on the frontend with inbound+outbound data.

aljesusg:
> Also, perhaps if we provide through the Health not a ratio, but the absolute number of requests, grouped by error code (or error code family) then we remove more processing from the backend and we leave that to the frontend. This way we can also get rid of `ErrorRatio` / `CodeErrorRatio` because it can be computed on the frontend with inbound+outbound data.

+1 I'm going to start in this way

lucasponce:
> To me is up to discuss if on this aggretation we want to put together all 4xx, 5xx in two values, or allow them to be individually (i'd rather prefer the group of 4xx vs 5xx, but I think here it can be a minor deal to play with both cases).

Also, let's not forget that we need the grpc/tcp case that doesn't have codes. 
I always miss to put that in my draft/examples.

aljesusg:
> > To me is up to discuss if on this aggretation we want to put together all 4xx, 5xx in two values, or allow them to be individually (i'd rather prefer the group of 4xx vs 5xx, but I think here it can be a minor deal to play with both cases).
> 
> Also, let's not forget that we need the grpc/tcp case that doesn't have codes.
> I always miss to put that in my draft/examples.

yep... I am thinking about how to tell the UI that requests are GRPC or HTTP.

If user not set how to deal with health defaults are

```ts
const grpcErrorRegexp = new RegExp('^[1-9]$|^1[0-6]$'); // 1..16
const httpErrorRegexp = new RegExp('^[4-5]\d\d$');      // 4xx, 5xx
```

So maybe we should return with requests the "request_protocol"....but Can we have different protocols in the communication? IMO yes...so we should return an object more complex.

lucasponce:
> So maybe we should return with requests the "request_protocol"....but Can we have different protocols in the communication? IMO yes...so we should return an object more complex.

Perhaps a map with protocols, we have two at the moment grpc or http/https.
I don't know, less familiar with this model than @jotak, but yes, separating data by protocols looks correct to me.

jotak:
Yes, it's either nested maps, or a flat collection with something like tags.
With nested maps you would have something like this:

```
inbound:
  http:
    200:
      value: 1234
```

With flat collection + tags:

```
tags:
  direction: inbound
  protocol: http
  code: 200
value: 1234
```

Alberto started with nested maps, it's fine for me. It has the advantage to aggregate values more easily. Perhaps if we continue to add more discriminators in the future, which would increase the maps depth, the flat collection will be better then, but now it's fine I think.

aljesusg:
there are many casuistry here...for example now we are checking errorCodes for GRPC and http for a X service.

We can have different protocol for a namespace (Service A with Service B in http and Service B with Service C in GRPC)
Now, we support it.

If user configure a health with protocol **http**, we won't have problems....but if we let the user set _* for all protocol_ the "code regex" is not the same for **http** than **grpc** so I don't know hot to deal with this case.

- No let regex
- Let regex but let to add more codeCheckers

aljesusg:
On the other hand foer example

```yaml
    healthConfig:
        rate:
        - tolerance:
            degraded: 5%
            failure: 10%
        - namespace: default
          tolerance:
            degraded: 10%
            failure: 15%
        - namespace: default
          app: reviews
          tolerance:
            code: 4xx
            degraded: 2%
            failure: 25%
```

In this case.
If there are 10% failure in namespace default this will match 

```yaml
        - tolerance:
            degraded: 5%
            failure: 10%
        - namespace: default
          tolerance:
            degraded: 10%
            failure: 15%
```
Should I get the "major" match or use the major percent error rate ?

With the sample example with the app reviews.
If I detect percent error 10% Should I show error by
```yaml
 - tolerance:
            degraded: 5%
            failure: 10%
```
that apply to all

or not because I have error if is greater than 25
```yaml
   - namespace: default
          app: reviews
          tolerance:
            code: 4xx
            degraded: 2%
            failure: 25%
```

jotak:
Yes I see there are some questions that need an answer. Perhaps we can have a quick call @lucasponce @aljesusg .
Another problematic example, let's consider:

```
   - namespace: default
     app: reviews
       tolerance:
         degraded: 2%
         failure: 25%
   - namespace: default
     tolerance:
       code: 4xx
       degraded: 20%
       failure: 30%
```

One is discriminated by app, the other is discriminated by code. Which one prevails if I have a 4xx error on app:review ?
I think there's two things we could do:
- Put "code" not in "tolerance" but at the same level as namespace and app, as they all are discriminators for selection.
- Say that the ordering matters. We could say that the first (or last) rule that applies is taken.

lucasponce:
> Put "code" not in "tolerance" but at the same level as namespace and app, as they all are discriminators for selection

+1

> Say that the ordering matters. We could say that the first (or last) rule that applies is taken.

+1

Yes, it's like usually these config works: first entry that matches the namespace/app works, and let the default one in last position.

jotak:
@aljesusg I think for a given service, when multiple rules apply (because we can have a rule for 4xx and another one for 5xx and another one for grpc, as you said), we compute the status for every rule, and then keep the worst status. That would work, no?

aljesusg:
> @aljesusg I think for a given service, when multiple rules apply (because we can have a rule for 4xx and another one for 5xx and another one for grpc, as you said), we compute the status for every rule, and then keep the worst status. That would work, no?
> 
> [so actually, I'm not sure anymore about moving `code` to the upper level :-) ]

^^ agree we should return the worst status and somekind of message how "hey you requests match code XXX > Y%"

aljesusg:
@prachiyadav I added the documentation here https://github.com/kiali/kiali/pull/2915/files#diff-7bd4a925c695c2eb0eced3872b9b965f

models/health.go
```diff
@@ -74,53 +67,33 @@ type WorkloadStatus struct {
...
 type RequestHealth struct {
-	inboundErrorRate    float64
-	outboundErrorRate   float64
-	inboundRequestRate  float64
-	outboundRequestRate float64
-
-	ErrorRatio         float64 `json:"errorRatio"`
-	InboundErrorRatio  float64 `json:"inboundErrorRatio"`
-	OutboundErrorRatio float64 `json:"outboundErrorRatio"`
+	InboundRequestCode  map[string]map[string]float64 `json:"inboundRequestsCode"`
```
jotak, line 70: We could simplify a little bit the naming here, changing `InboundRequestCode` to just `inbound`.
That would give us:

```json
{
  "requests": {
    "inbound": {
      "http": {
        "200": ...,
        "503": ...
      }
    },
    "outbound": {
      "http": {
        "200": ...
      }
    }
  }
}
```

README.adoc
```diff
@@ -285,6 +285,104 @@ spec:
 ...
 ----
 
+== Configure Health
+
+If you want customize the health alerts in Kiali you can do it in the configuration yaml.
```
jotak, line 290: I would remove this sentence as the section title `Configure Health` is self-explanatory.

README.adoc
```diff
@@ -285,6 +285,104 @@ spec:
...
+For each Rate Kiali will use the first match by:
+
+* Namespace (string) is a regular expression that match the namespace name (bookinfo, istio-system...)
+* Kind (Optional) (string) is a regular expression that match the resource type (app/workload/service)
+* Name (Optional) (string) is a regular expression that match the resource name (reviews, productpage...)
+
+If a configuration match kiali will use the first tolerance configuration that match:
+
+* Request protocol is a plain string (http/grpc...)
+* Request type (string) is a regular expression that match the request (inbound/outbound)
```
jotak, line 340: I would prefer `direction` than `request_type`, it's more meaningful.

jotak: Also, I don't think it's worth it to run regexp checks to match these strings when there's only 2 or 3 possible values. It's not the same cost as a direct string match, so in this case it sounds like a an overkill. As for "Kind" above.

README.adoc
```diff
@@ -285,6 +285,104 @@ spec:
...
+          request_type: "outbound"
+          code: "400"
+          degraded: 1
+          failure: 5
+----
+
+In this case we created :
+
+*  For namespace default:
+  ** A "degraded" alert in http protocol and if request type match the regular expression bound (In this case for inbound and outbound), when code mach the regular expression 400 rate is >=5% and failure alert if >=10%
```
jotak, line 378: It's not an alert here ... it's actually a threshold (I know Lucas didn't want to use this word, but "alert" is even worse, no?)

README.adoc
aljesusg: mmm I was thinking about ...the user want to have the health for inbound and outbound...or in kind case for wokload and app...

jotak: But if you don't set these settings (driection or kind), I guess it stands for "all", no?
Unset direction means both inbound and outbound ; unset kind means any kind of entity?

The only missing possibility would be if you want to apply something, for instance, only on a app and service but not the related workloads ... but it seems really like an unusual corner case and anyway it's still possible to do by adding more rate configs, no?

config/config.go
```diff
@@ -289,6 +311,7 @@ type Config struct {
 	Deployment               DeploymentConfig         `yaml:"deployment,omitempty"`
 	Extensions               Extensions               `yaml:"extensions,omitempty"`
 	ExternalServices         ExternalServices         `yaml:"external_services,omitempty"`
+	HealthConfig             HealthConfig             `yaml:"health_config,omitempty" json:"healthConfig"`
```
jotak, line 314: Look below in `NewConfig` : this is where I guess you should create the default config.

jotak:
I guess there must be a missing PR on the operator to be able to setup a config via Kiali CR, no?
I tried to setup health config in CR but it wasn't working .. still had a "null" health rates in UI.

aljesusg:
> I guess there must be a missing PR on the operator to be able to setup a config via Kiali CR, no?
> I tried to setup health config in CR but it wasn't working .. still had a "null" health rates in UI. Not sure if I did something wrong.

Yes, I didn't do it yet only working in local xD.

Here you have https://github.com/kiali/kiali-operator/pull/82

lucasponce:
What do I need to test this and https://github.com/kiali/kiali-ui/pull/1831 ?

I'm having an error like:

image

the config fetching from the server

image

In my config I have nothing for the healthConfig, so, it should delegates into the default one in that scenario.

aljesusg:
> What do I need to test this and [kiali/kiali-ui#1831](https://github.com/kiali/kiali-ui/pull/1831) ?
> 
> I'm having an error like:
> 
> image
> 
> the config fetching from the server
> 
> image
> 
> In my config I have nothing for the healthConfig, so, it should delegates into the default one in that scenario.

Hi @lucasponce you need this PR https://github.com/kiali/kiali-operator/pull/82 too. The Operator set the rate value to []. Anyway I am going to create another check in the UI if the rate is null

config/config.go
```diff
@@ -397,6 +397,9 @@ func NewConfig() (c *Config) {
 				WhiteListIstioSystem: []string{"jaeger-query", "istio-ingressgateway"},
 			},
 		},
+		HealthConfig: HealthConfig{
+			Rate: []Rate{},
```
jotak, line 401: I think you should put good defaults here that ensure Kiali behaves the same as today without any configuration ? ie. degraded threshold > 0.1% , failure > 20% ?

aljesusg: I'm keeping the defaults in the UI https://github.com/kiali/kiali-ui/pull/1831/files#diff-0e0ca9c6860e3c885bf7cb495a0ad59bR89 . So the proposal is move this default to the config?. My doubt is if user set an empty config we won't have health in the UI

jotak: I think the default has to be duplicated in UI *and* in backend. The reason is, they have different purposes:
- in the UI, the default is used just as temporary data until server config is fetched; after it's fetched, it has strictly no effect
- in the backend, the default is used to provide data when user hasn't configured anything

jotak: PS just to let you know: I've re-tested your PR, and the health isn't working because of that, defaults are missing.

aljesusg: Added defaults

jotak:
Hey @aljesusg , I've been testing and can't get any valid health. After investigating what could be wrong, I'm finding that the data model between frontend and backend just doesn't match: see here [1] and there [2]: backend was changed to `inbound` / `outbound` while frontend is still using `inboundRequestsCode` and `outboundRequestsCode`. So I can't see how it could work, even the nominal case.

Could you please test it before pinging reviewers back, or mark it somehow as WIP until it can be tested and reviewed? Thank you.

[1] https://github.com/kiali/kiali/pull/2915/files#diff-e52a19cb969be4a75acba5e9232a1f71R77-R78
[2] https://github.com/kiali/kiali-ui/pull/1831/files#diff-0e0ca9c6860e3c885bf7cb495a0ad59bR34-R35

jotak (dismissed):
doesn't work (see comment above)

jotak:
Ok, then please in this situation mark it explicitly as WIP, or draft, or something like that. I've changed the title in this way now.

lucasponce:
> Ok, then please in this situation mark it explicitly as WIP, or draft, or something like that. I've changed the title in this way now.

A label is the way :).
I've added back the "do not merge" one, which is the way to indicate this.

@aljesusg let us know when you want us to start testing and reviewing more closely.

lucasponce:
> Hi @lucasponce you need this PR kiali/kiali-operator#82 too. The Operator set the rate value to []. Anyway I am going to create another check in the UI if the rate is null

Note that this shouldn't be correct.
Kiali should load defaults without need to use the operator or having an explicit config.
(Just if not present, then everything works as the previous scenario which should be the default one).

README.adoc
```diff
@@ -285,6 +285,102 @@ spec:
 ...
 ----
 
+== Configure Health
```
lucasponce, line 288: I am not sure this feature belongs to README.adoc but kiali.io
If we document a feature, we need to document ALL features here.

So, I prefer a better documentation in kiali.io and leave README.adoc more clean (or point to the proper kiali.io place).

README.adoc
```diff
@@ -285,6 +285,102 @@ spec:
...
+
+=== Kiali defaults
+
+By default Kiali health is:
```
lucasponce, line 292: This is not clear.
It's assuming what is health in all context.
Health happens in several places: Graph, Namespace, Application, Workload and Service.
And for each component there are several items that are used to compute them (istio metrics and also controller data current pods, requested, etc).

This need to be documented properly in a way that the user understand what is referring to.

README.adoc
```diff
@@ -285,6 +285,102 @@ spec:
...
+* Failure when Error Rate is >= 20%
+
+For Kiali an error request is when:
+
+* For http the code status match the expression */^[4-5]\d\d$/*
+* For grpc the code status match the expression */^ [1-9]$|^1[0-6]$/*
+
+=== Configuration
+
+Kiali works with a FIFO configuration, so Kiali will use the first configuration that applies with the resource requested.
```
lucasponce, line 304: FIFO configuration is unclear.

I think I understand what you want to describe, I had a good example in Istio VS:
https://istio.io/latest/docs/reference/config/networking/virtual-service/

"An ordered list of configuration entries. [...]. The first matching configuration is used."

README.adoc
```diff
@@ -285,6 +285,102 @@ spec:
...
+
+So the granularity and criteria should be configured from highest to lowest.
```
lucasponce, line 306: You assume that user knows about granularity and criteria and that's not described on the feature.

README.adoc
```diff
@@ -285,6 +285,102 @@ spec:
...
+          ...
+    ...
+
+
+----
+
+For each Rate Kiali will use the first match by:
+
+* Namespace (string) is a regular expression that match the namespace name (bookinfo, istio-system...)
+* Kind (Optional) (string) is a regular expression that match the resource type (app/workload/service)
```
lucasponce, line 332: I think I missed the discussion about filtering by "kind".
I see that's optional but I think this can introduce more edges and corner cases rather than really help
(i.e. app is a composition of workload + services, if it has a different config, perhaps user can get easily some discrepancies, "apps are health, but services look in red").

I will continue looking the details and examples to add more comments about this.

README.adoc
```diff
@@ -285,6 +285,102 @@ spec:
...
+* Name (Optional) (string) is a regular expression that match the resource name (reviews, productpage...)
+
+If a configuration match kiali will use the first tolerance configuration that match:
```
lucasponce, line 335: It would be good to indicate user where these matching fields are coming from.
It implicitly requires that user understand that these are istio collected metrics.
It would be worth to add better description here.

README.adoc
```diff
@@ -285,6 +285,102 @@ spec:
...
+
+*  For namespace default:
+  ** A "degraded" threshold in http protocol and if direction match the regular expression bound (In this case for inbound and outbound), when code mach the regular expression 400 rate is >=5% and failure alert if >=10%
+  ** A "degraded" threshold in grpc protocol and if direction match the regular expression inbound (In this case for inbound), when code mach the regular expression 1 rate is >=10% and failure alert if >=15%
+* For namespace bookinfo, app reviews:
+  ** A "degraded" threshold in http protocol and if direction match the regular expression outbound (In this case for outbound and outbound), when code mach the regular expression 400 rate is >=1% and failure alert if >=5%
+
+
+
+
```
lucasponce, line 383: It's good to have documentation.
I've added some comments as this feature is new and it needs additional description to understand what it does (it requires a lot of implicit knowledge).

I don't think the README.adoc is the proper place for this.
Better to add a section in the Features where we are consolidating things in the kiali.io.

config/config.go
```diff
@@ -374,6 +397,31 @@ func NewConfig() (c *Config) {
 				WhiteListIstioSystem: []string{"jaeger-query", "istio-ingressgateway"},
 			},
 		},
+		HealthConfig: HealthConfig{
```
lucasponce, line 400: Is this default config the previous existing logic for calculate health before this feature ?
Or are we introducing a new default here ?

models/health.go
```diff
@@ -74,53 +74,33 @@ type WorkloadStatus struct {
...
 type RequestHealth struct {
-	inboundErrorRate    float64
-	outboundErrorRate   float64
-	inboundRequestRate  float64
-	outboundRequestRate float64
-
-	ErrorRatio         float64 `json:"errorRatio"`
-	InboundErrorRatio  float64 `json:"inboundErrorRatio"`
-	OutboundErrorRatio float64 `json:"outboundErrorRatio"`
+	Inbound  map[string]map[string]float64 `json:"inbound"`
```
lucasponce, line 77: When introducing nested maps, it'd be good to describe in a comment the potential values for keys and other relevant info in the structure (what values should or shouldn't be there).

README.adoc
aljesusg: ok

README.adoc
aljesusg: This impact in  Graph, Namespace, Application, Workload and Service

README.adoc
aljesusg: is a regex related with the resource

config/config.go
aljesusg: This is the default config that we had

lucasponce:
I think it's in good shape.
I have some comments but I would be fine to consolidate it if @jotak doesn't have additional thoughts on it.

Please, prepare a PR on kiali.io with the comments addressed, I'd like to update the doc at same time we merge this.

config/config.go
```diff
@@ -452,6 +475,37 @@ func NewConfig() (c *Config) {
...
+	// Health default configuration
+	healthConfig := HealthConfig{
+		Rate: []Rate{
+			{
+				Namespace: ".*",
+				Kind:      ".*",
+				Name:      ".*",
+				Tolerance: []Tolerance{
+					{
+						Code:      "^5\\d\\d$",
```
lucasponce, line 489: I don't remember the previous defaults, if any, but it could be good time to ask folks.
Personally I think we should have some tolerance on the 404, but fail on the rest 4xx and 5xx.

I know this is opinionated but also @jshaughn @jotak have some thoughts on it.

Also, I wouldn't use the "0.1" value, just "0" or empty I think, a "0.1" tend to be read as 10% instead of 0.1% which perhaps has less sense.

jshaughn: I don't really know what to say about 404s as a special case of 4xx client errors.  It would be simpler semantics if the default applied to 4xx completely.  Either ignore or set a high tolerance for client errors, maybe like 50%.  If a service is getting killed with client errors it could be a DNS attack or something like that.

As for the 0.1, I'm OK if Degraded is changed to be >0 errors for the time period.

lucasponce: @aljesusg @jshaughn I'm ok to apply tolerance for 4xx (client errors).
Perhaps we can set something like starting to degraded in 25% of client errors and show failure state in 50% (as default).
And this is just a starting point, I can imagine when this feature is in place there were more cases.

aljesusg: WE must add it again to be align with traffic tab and graph

lucasponce: @aljesusg Not sure what you mean, your comment is confusing.
Graph should be synced if 4xx are present or not.
All elements that show health state should read the health config from the same source.

Another discussion is if we consider 4xx error healthy or unhealthy.

jotak: +1 for more tolerance on 4xx

lucasponce: ok, let's tweak the defaults:
- 5xx and grpc errors degraded starting in 0, failure starts on 10%.
- 4xx starts to degraded in 20%, failure on 30% (I know this is pure opinion, just to have a starting point).

lucasponce (changes requested):
Backend code LGTM.
One comment to tweak defaults.

Perhaps we may revisit how to change the defaults one, but that can be a separated issue, following this, when this first step is consolidated.

config/config.go
aljesusg: Agree with that. what do you think @jotak @jshaughn ?

jotak: I would put something in-between for 4xx, like 10% / 20% ; because, as I said above, for outbound connections this is *really* errors that lies in user's services.

jotak: (but no problem to do as @lucasponce suggests and eventually tweak that later)

aljesusg: so...proposals:

5xx and grpc errors degraded starting in 0, failure starts on 10%.
4xx starts to degraded in 20%, failure on 30%

or

5xx and grpc errors degraded starting in 0, failure starts on 10%.
4xx starts to degraded in 10%, failure on 20%

jotak: I don't mean to slow down things by commenting. It's fine to use what lucas suggests.

aljesusg: we have time while https://github.com/kiali/kiali-ui/pull/1831 is reviewed

jshaughn: I'm fine with 5xx 0%/10% because there is some indication of a possible issue as soon as there are any 5xx.   For 4xx I think we need to assume that client errors are potentially bad and so I'm in favor of the lower setting for degraded, 10%.  I'm less concerned with the Failure percentage but to be consistent with a 10% delta above degraded, I'd go with 20%.

aljesusg: ok so 10-20. Anyway the user can overwrite this configuration if he wants other values

lucasponce (approved):
LGTM
It's in a stable point.
There will be more minor followup tasks around this topic, but it's good to consolidate this work.

Issue #2903, opened 2020-06-12
Kiali cannot access after using openid login port mapping


shinynz:
**Describe the bug**
My domain name has been mapped from kubernetes to port 80, but after using keyclock, the original port 20001 has been brought back and cannot be accessed

**Versions used**
Kiali:  v1.19.
Istio:  v1.6.1
Kubernetes flavour and version: kubernetes 1.18.2

**Expected behavior**
After k8s port mapping, kiali can use the mapped port instead of taking out the original port

![image](https://user-images.githubusercontent.com/15795798/84510297-88c91300-acf7-11ea-8476-488907ac93f5.png)

shinynz:
After using keyclock authentication, the return port of kiali should be the address exit port, not the internal port of the kiali container

shinynz:
kiali for keycloak login error
![image](https://user-images.githubusercontent.com/15795798/84638752-f6f51c00-af29-11ea-8f6d-a46d914e3c12.png)

lucasponce:
Thanks for the reporting @shinynz.

I cc our colleague @israel-hdez which has been working on this area.

jmazzitelli:
I think this means "keycloak" right? If so, please change the title of this issue as it is confusing :)

jmazzitelli:
@shinynz We are going to need more information. I assume you are using the new auth strategy of "openid" - is that correct?

If so, we will need to see your Kiali CR configuration, particularly the "auth" section so we can see how you configured Kiali to integrate with your OpenID Connect implementation of KeyCloak.

In addition, please provide the logs messages found in the Kiali pod. Please do not provide screenshots of the logs, as we will need to see more lines than a screenshot can provide, and we would need to search the logs for relevant details. Feel free to scrub the logs of sensitive domain names and things like that if you wish.

Thanks!

israel-hdez:
@shinynz I also would like to see your Kiali CR configuration.
Also, what cloud provider are you using for your Kubernetes cluster?

shinynz:
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #2919, opened 2020-06-19
fix guess url for default port 

silenceshell:
** Describe the change **

We met this problem when we connect kiali web from the service(the service ip is accessible out of k8s cluster). 
We use openid auth strategy. 
As the kiali service port is 80, we connect kiali by `http://1.1.1.1`, then the browser will redirect to openid login page; After the user login, the browser will redirect to `http://1.1.1.1:20001`, which should be `http://1.1.1.1`.

We check the func `GuessKialiURL` and find this maybe a bug. 
If the http request uses default port(80, 443) both in url and in host header, the func `GuessKialiURL` will use `cfg.Server.WebFQDN`, which maybe 20001.

So we add a step to check if it is a default port, if so, use default port instead.

Look forward to your response :)

** Issue reference **
NA

* Please, provide the link to the GitHub issue here.
NA

** Backwards incompatible? **
None

** Documentation **
NA

NA

### DIFF
--- a/util/httputil/http_util.go
+++ b/util/httputil/http_util.go
@@ -121,19 +121,27 @@ func GuessKialiURL(r *http.Request) string {
 		}
 	}
 
+	var isDefaultPort = false
 	// Guess the port. In this case, the port in configuration doesn't take
 	// priority, because this is the port where the pod is listening, which may
 	// be mapped to another public port via the Service/Ingress. So, HTTP headers
 	// take priority.
 	if fwdPort, ok := r.Header["X-Forwarded-Port"]; ok && len(fwdPort) == 1 {
 		port = fwdPort[0]
-	} else if len(r.URL.Port()) != 0 {
-		port = r.URL.Port()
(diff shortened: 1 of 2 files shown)


jmazzitelli:
This might also fix https://github.com/kiali/kiali/issues/2903 - asking @israel-hdez to confirm and review.

israel-hdez:
Thanks for this contribution, @silenceshell :)
If you can, let us know how the `openid` strategy is working for you. It was added recently and we will appreciate any feedback you can provide.

silenceshell:
@israel-hdez 
Thanks for the openid strategy, it helps us a lot.
We use dex as a OIDC provider, an we add a connecter for dex, which is a sso login web site.
And we create rolebinding for different user in different namespace.
People login from kiali, and will be redirected to dex and then the sso login web site, after login, they could use kiali to manage their **own** service mesh network.

israel-hdez:
@silenceshell Thanks for letting us know how you are using Kiali.

Issue #2860, opened 2020-06-03
Ability to reuse a handler when linking 3scale to workload


PhilipGough:
**Describe the solution you'd like**
cc @lucasponce: 
Currently it seems to be that when selecting a workload to be integrated with 3scale through the wizard, each integration creates a separate handler, instance and rule.

Ideally however, we would create only a single handler per account and provide the service ID directly by editing the deployment, assuming there are no issues in doing so on the Kiali side

Detailed description can be read here https://github.com/3scale/3scale-istio-adapter#routing-service-traffic-through-the-adapter

lucasponce:
Hi @PhilipGough,

I think in Kiali (master branch) we have refactored this to address some of the behaviour you comment.

Now, when 3scale extension is present, there is a new top level menu in Kiali:

![image](https://user-images.githubusercontent.com/1662329/83727816-f74a0900-a645-11ea-833b-653750a284f6.png)

On the 3scale main page, there is a list of existing 3scale handlers:

![image](https://user-images.githubusercontent.com/1662329/83727913-15176e00-a646-11ea-879e-e62f58ae67c8.png)

As "3scale handler" is more a logical entity (in implementation is a rule/instance of mixer that can be changed in the future), Kiali manages it from "logical" perspective, so, that's reason those handlers are not namespaced on this page as they should live in the control-plane namespace.

The form to create a handler is placed here:

![image](https://user-images.githubusercontent.com/1662329/83728051-4859fd00-a646-11ea-816f-826d2c6e2dd5.png)

In this way, handlers are in practice shared between all services, we have eliminated the option to create a handler from the wizard but selecting an existing one (creation of handlers will happen from the 3scale extension section).

![image](https://user-images.githubusercontent.com/1662329/83728172-74757e00-a646-11ea-8e79-741323391182.png)

All of this is consolidated in master.

This refactoring was motivated to: a) address this re-use usecase, b) also to shape extensions in a more generic way, allowing other features to extend Kiali in a more organized way.

Please, if you have possibility to test this, we can try to see if the current work is valid or if we would need to change some aspect.

lucasponce:
Probably, we can use this issue to study how to change the rule I think.
Today, every handler is linked with a single Rule with a explicit service condition.

PhilipGough:
@lucasponce - thanks for the explanation

This all looks like valid changes, it still does tie the service ID directly to the handler though as opposed to coming from the target services deployment via a label as I linked above.

[...]

PR #2936, opened 2020-06-25
Backend refactoring triggered by 3scale changes

lucasponce:
Work in Progress for:
- Remove 3scale abstraction model in Kiali backend, now 3scale extension will deal directly with Istio Mixer objects: Handlers, Instances and Rules.
- Expose labelSelector to GetIstioConfig endpoint, used mainly to discriminate some Istio config generated by Kiali in 3scale scenario.
- Refactoring cache naming to better alignment with kubernetes package API.
- Expose a new parameter in the config used for 3scale config, this would need a PR in kiali-operator repo.
- Introduce a new handler PATCH /workloads to be used for label workloads (needed to link/unlink 3scale credentials).

It requires https://github.com/kiali/kiali-ui/pull/1833 and https://github.com/kiali/kiali-operator/pull/92

Note that this PR also requires changes on the kiali-operator as it needs access to a new verb (patch) and also it changes the configuration.

Related to https://github.com/kiali/kiali/issues/2860


business/workloads.go
```diff
@@ -1062,6 +1082,53 @@ func fetchWorkload(layer *Layer, namespace string, workloadName string) (*models
...
+	workloadTypes := []string{
+		kubernetes.DeploymentType,
+		kubernetes.ReplicaSetType,
+		kubernetes.ReplicationControllerType,
+		kubernetes.DeploymentConfigType,
+		kubernetes.StatefulSetType,
+		kubernetes.JobType,
+		kubernetes.CronJobType,
+		kubernetes.PodType,
+	}
```
jotak, line 1101: If I understand correctly, for a single workload patch request we're going to try to patch every possible kind of workloads with that name?
I guess it works in general because only one type will match with the provided name and the others will return errors / not found?

Is it possible to improve this, and provide the correct type directly from the query params / ui ?
This would give better performances (avoid unnecessary kube queries) and also prevent the corner case where several workloads of different types would have the same name (I guess this corner case would trigger patch errors)

lucasponce: In Kiali we are assuming that we won't have two different controllers with same name in the namespace.
Basically, Istio didn't support that cornercase and we maintained that:
https://github.com/kiali/kiali/blob/master/business/workloads.go#L1065

Yes, perhaps in the handler we could pass an optional parameter with the workloadType and if present to follow that, otherwise fall into fetching all.

But this is a helper, to proper fix this we should revisit the workload model, and I tried to avoid that as it may trigger more major changes that perhaps are better to adress them in a next major version.

xeviknal: I agree with @jotak. It looks like we can easily avoid the multiple calls to k8s API with passing this extra param. I see that the workload model on the backend already have the type field that can be useful in here.
I might be missing something over here though.

lucasponce: Yes, an extra param may help here, but just keeping optional to not break the current model.
In short, kubernetes delegates into the user to not have two different controllers with same name, Istio followed that implicitly in their model and we mapped the workloads using that assumption.
But adding a check for get/update a single workload with a type may help here.
I'll add it during the week to the PR.

What I don't want to open is to introduce any kind of big refactoring as adding constratints into the workload model in Kiali for now, that's something we can discuss in a major iteration.

lucasponce:
@xeviknal @jotak I've addressed your comments and added an optional type for workload details (fetch and update flows).
Let me know if it's ok.

I don't remove the "do not merge" label because I'm waiting to finish the UI changes and prepare the PR with the operator.

kubernetes/filters.go
```diff
@@ -184,3 +184,13 @@ func FilterByRoute(spec map[string]interface{}, protocols []string, service stri
 	}
 	return false
 }
+
+func FilterIstioObjectsForSelector(selector labels.Selector, allObjects []IstioObject) []IstioObject {
+	var istioObjects []IstioObject
```
jotak, line 189: If the resulting array is to be serialized as json, it's probably safer to initialize this array `istioObjects := []IstioObject{}` ?

lucasponce: It's not exposed but the initialization makes sense

lucasponce:
Thanks for the review @xeviknal @hhovsepy @jotak going to merge these ones (backend + frontend) and check with @jmazzitelli if the operator one looks good.

Issue #2856, opened 2020-06-02
Design: request classification information in graph


jshaughn:
Epic: https://github.com/kiali/kiali/issues/2854

This is to discuss the design for adding[ request classification](https://istio.io/docs/tasks/observability/metrics/classify-metrics/#classify-metrics-by-request) information to the graph.

jshaughn:
Request classification is about being able to aggregate metrics for requests matching various request/route conditions.  Services can perform a variety of "operations" and it allows the observer to better understand how a service is being, used at a more customized, granular level.

When applicable the current Istio metrics will be assigned the "request_operation" label.  What we want to be able to see are these operation aggregates in the graph.

1. One idea is to introduce a new graph type, "Op Graph", that focused specifically on operations, showing only source->op->destination connections, likely resulting in "bow-tie" like graphs:

```
    S1 \          / D1
    S2  - OP1 - D2
    S3 /          \ D3
```

But this loses a lot of the context of a normal graph and also calls into question the type we're looking at for the Sources and Destinations.

2. Another idea is to not change the graph but to make the operation information available in the side-panel for relevant destination nodes, perhaps badging destination nodes that have operation information available.

This would likely be the easiest approach, perhaps designing a new Tab for this information.  And it won't further complicate an already complicated graph with nodes that are essentially not part of the actual topology.

But this approach does not provide much, if any, visual cue to a problematic operation.

3. The most visual approach is to maintain our current graph types and to inject operations into them.  This is the more difficult thing to do but the idea would be to have a new Display option, "Show Op Nodes", and when enabled show the operations in the graph.  It would likely be disabled by default.

This implies the need for a new node type, because operations don't fit as a service, app or workload node.   This is actually a fundamental change for the graph as it would be the first node to represent "meta-information".  It's not really a config object nor does it really have to do with request routing.

As a new node type it could have a dedicated side-panel implementation.  It likely would not offer a node-detail graph, but should show up in the node-detail graph for a relevant destination node.

I'd vote for option 2 or 3.   I don't think option 1 buys us very much.

jshaughn:
I'm starting to lean towards option 2 for a couple of reasons:
- It avoids introducing non-topology nodes into the graph.
  - Non-topology nodes may confuse users as today the graph reflects routing/traffic only.
  - It could add a significant number of nodes making the graph more complicated.
- It avoids having to figure out where those nodes should be injected.
  - It's not 100% clear to me whether they would lead into or out of a service node.
[...]

Issue #2854, opened 2020-06-02
Add request classification information to graph


jshaughn:
Istio 1.6 introduced a new [experimental] [feature for request classification metrics](https://istio.io/docs/tasks/observability/metrics/classify-metrics/).   Istio users have long-requested route-specific metrics but these are not technically feasible due to the cardinality explosion that results from route-specific time-series.  The solution is to group similar requests/routes into a more limited set of "operations".  Cardinality will increase but if following best practices it should be manageable.

The end-result is that metrics may now provide a "request_operation" label that contains the operationId.  Note that the label name "request_operation" is actually configurable by the user but I think it's fine for Kiali to require "request_operation".

There is also the ability to perform response code classification but that is out of scope for this RFE and I don't currently see a use for this in Kiali, but rather for user-specific dashboards or alerting.

Epic Tasks:
- [x] Graph design work https://github.com/kiali/kiali/issues/2856
- [x] Graph generation work #2908 
- [x] Graph side-panel work #2908
- [x] Documention  #2969
- [ ] Blog (optional)

Note that this Epic is specific to showing this information in the graph and side-panel.   We can decide later whether some sort of detail view makes sense.

jotak:
>  Note that the label name "request_operation" is actually configurable by the user but I think it's fine for Kiali to require "request_operation"

IMO it's better to make that name configurable. The name "request_operation" is really taken as an example in this doc/task I think. It's not even a default value.

jshaughn:
May still do a blog, but the main items are complete.

PR #2944, opened 2020-06-27
Introduce operation aggregate nodes

### LINKED ISSUE #2908: graph gen: request classification information in graph
Epic: #2854 

This is poc/impl work for adding request operation classification into the graph.

jshaughn:
Introduce "Operation Nodes" into the graph based on the experimental request classification feature introduced in Istio 1.6.  Users can now classify requests into "operations", by performing regex matching on the urls.  See the epic and design issues for more.  This server-side PR adds the aggregate_node appender to graph generation, as well as new API support for aggregates.  An "aggregate" is basically a metric attribute that we expect to hold "bucket" values.  The backend support is generic although we default to "request_operation", as it is the name used in the istio documentation example for the request classification feature.  The aggregate nodes are supplied to the graphing client (our ui) where they are presented as "operation nodes".  We can decide on whether we like that naming, but the main point is that the backend is generic but what we ultimately present in the UI should likely be more specific.

UI PR: https://github.com/kiali/kiali-ui/pull/1836
Epic #2854 
Design: #2856
Fixes #2908


jshaughn:
@jmazzitelli I think the PRs are OK to test and review.  I assigned you to only review the server-side code.  I'll ask someone else to review the client-side.

You can install the travel-agency demo via `hack/istio/install-travel-agency-demo.sh -eo true`.   You can look at the hack script for the request classification definitions and the request_operation attribute.

@jotak I'm adding you as a reviewer but you don't have to look at it all if you don't like, but please take a look at the metrics additions for aggregates.

I will keep the DNM label until I add tests for the new appender.

graph/telemetry/istio/appender/aggregate_node.go
```diff
@@ -0,0 +1,276 @@
...
+	groupBy := fmt.Sprintf("source_workload_namespace,source_workload,source_%s,source_%s,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_%s,destination_%s,request_protocol,response_code,grpc_response_status,response_flags,%s", appLabel, verLabel, appLabel, verLabel, a.Aggregate)
+	httpQuery := fmt.Sprintf(`sum(rate(%s{reporter="destination",source_workload_namespace!="%s",destination_service_namespace="%v",%s!="unknown"}[%vs])) by (%s) > 0`,
+		"istio_requests_total",
+		namespace,
+		namespace,
+		a.Aggregate,
+		int(duration.Seconds()), // range duration for the query
+		groupBy)
+	tcpQuery := fmt.Sprintf(`sum(rate(%s{reporter="destination",source_workload_namespace!="%s",destination_service_namespace="%v",%s!="unknown"}[%vs])) by (%s) > 0`,
+		"istio_tcp_sent_bytes_total",
```
jmazzitelli, line 73: I still don't understand this use case. And you are doing two more queries (to support istio_tcp_sent_bytes_total) so we need to make sure these really are going to be useful. Are you sure users can categorize tcp traffic in envoy filters? Is that a valid thing to do?

The only thing I can think of is for gRPC traffic -- is gRPC telemetry in the istio_tcp_sent_bytes_total timeseries? If so, how does one create envoy filters to categorize gRPC traffic (is that even possible)?

If it isn't possible or doesn't make sense, I would remove the code involving istio_tcp_sent_bytes_total since it just adds 2 more prom queries that are being executed but are not useful, and clutters up the code.

handlers/metrics.go
```diff
@@ -8,18 +8,52 @@ import (
...
+		return
+	}
+
+	params := prometheus.IstioMetricsQuery{Namespace: namespace, Aggregate: aggregate, AggregateValue: aggregateValue}
+	err := extractIstioMetricsQueryParams(r, &params, namespaceInfo)
+	if err != nil {
+		RespondWithError(w, http.StatusBadRequest, err.Error())
+		return
+	}
+	// Ensure direction=inbound because aggregates always reflect inbound traffic to the target workload.  This ensures
```
jotak, line 42: Isn't it going to produce unexpected/wrong results if params.Direction was outbound? Perhaps return a StatusBadRequest instead?

graph/telemetry/istio/appender/aggregate_node.go
jshaughn: After some itsio slack conversation it seems the gRPC may be less useful, or not useful, but it's a no-op for us because it's reported as a request counter in requests_total.  It seems the TCP support may not be needed, so I'll remove it because it does cost us some queries.  If someone shows it's useful we can always add it back (I may leave the queries commented out).

handlers/metrics.go
jshaughn: @jotak OK, I can return a StatusBadRequest instead of forcing the Direction.

graph/telemetry/istio/appender/aggregate_node.go
jshaughn: PR Updated.

handlers/metrics.go
jshaughn: PR Updated.

graph/telemetry/istio/appender/appender.go
```diff
@@ -103,6 +107,24 @@ func ParseAppenders(o graph.TelemetryOptions) []graph.Appender {
 		}
 		appenders = append(appenders, a)
 	}
+	if _, ok := requestedAppenders[AggregateNodeAppenderName]; ok || o.Appenders.All {
+		aggregate := o.NodeOptions.Aggregate
+		aggregateValue := o.NodeOptions.AggregateValue
+		if aggregate == "" {
```
jotak, line 113: One thing perhaps I didn't understand correctly: how will a user change the default aggregate? IIUC "aggregate" is controlled by UI, so, will it be hard-coded in UI, or editable from UI?
Initially I thought that would be something configurable from configmap.

jshaughn: @jotak In the initial PR there is no support to change the default aggregate, it is fixed to "request_operation".  The graph request supports a query param that can override the default but the UI is not using that option.  So in this initial PR it is fixed.   If you think we need to make it immediately configurable I can add support.  It seemed OK to require "request_operation" for now only because it's brand new and that's what Istio uses in the example.  let me know...

jotak: No problem with me, just I wasn't aware.
Also, in the same idea of potential future changes, perhaps we could have multiple custom aggregates and not just a single one, which would add complexity to the model ... but we're not there yet, let's see what people wants :-)

Issue #2591, opened 2020-03-31
need to run the molecule tests periodically


jmazzitelli:
We have a good suite of molecule tests. These are currently run manually only. We should get these to run automatically through some CI job.

jmazzitelli:
One thing to note - if running on an OpenShift cluster and the tests will be done with dev images (`MOLECULE_USE_DEV_IMAGES=true`), this requires the dev images pushed to the 'kiali' namespace to be allowed to be pulled into the `anothernamespace` namespace as part of the test `default-namespace-test`. Otherwise, the kiali pod will not be authorized to pull the kiali dev image and this error results: 
```
Failed to pull image "image-registry.openshift-image-registry.svc:5000/kiali/kiali:dev": rpc error: code = Unknown desc = Error reading manifest dev in image-registry.openshift-image-registry.svc:5000/kiali/kiali: unauthorized: authentication required
```
In short, if a test installs Kiali in any namespace other than istio-system, then you have to give authorization to that other namespace because the current Makefile only prepares authorization to the namespace you give it in the cluster-push target. So something like this is needed:
```
oc policy add-role-to-user system:image-puller system:serviceaccount:anothernamespace:kiali-service-account --namespace=kiali 
```

jmazzitelli:
The PR https://github.com/kiali/kiali/pull/2943 provides a script that can run any set of molecule tests that you want. It performs the necessary preparations for tests that need it (like that "default-namespace-test" I mentioned in the earlier comment).

jmazzitelli:
If you install OpenShift on a bare metal machine using https://github.com/kxr/ocp4_setup_upi_kvm (taking most of the defaults) then you can run the molecule tests by installing a script like this in the machine's root crontab (or, if not root, you need to give the user read/execute access to everything under /root/ocp4_setup_ocp4):

```
#!/bin/bash

# abort on any error
set -e

ID="/root/ocp4_setup_ocp4/install_dir"
OC="/root/ocp4_setup_ocp4/oc"
SRC="/tmp/KIALI-TESTS"
DORP="podman"

echo "Create a clean github repo location"
rm -rf /tmp/KIALI-TESTS
mkdir -p $SRC

echo "Make sure everything exists"
test -d $ID || (echo "$ID is missing"; exit 1)
test -x $OC || (echo "$OC is missing"; exit 1)
test -d $SRC || (echo "$SRC is missing"; exit 1)
[...]

PR #2945, opened 2020-06-28
[hack] script for running Molecule tests periodically

jmazzitelli:
This doesn't actually setup a CI system to periodically run the tests, but this PR does address #2591 in that it provides hack scripts that can be run within a cronjob or other automatic system. It will run the molecule tests against any cluster you currently are logged into via "oc login". If you are not logged in, this will attempt to install an AWS cluster and install either Service Mesh or upstream Istio.

fixes: https://github.com/kiali/kiali/issues/2591


jmazzitelli:
note that the script as checked in now currently only supports running tests on OpenShift. I will plan on enhancing these to run against minikube.

Issue #2933, opened 2020-06-23
Invalid OIDC "/authorize" request due to missing state param


asmith60:
**Describe the bug**
When using the `openid` auth strategy I receive the following error from my IDP (Okta):

```error=invalid_request&error_description=The+authentication+request+has+an+invalid+%27state%27+parameter```

I see in the `/authorize` request that the `state` parameter is not included. Many IDPs (including Okta) require the `state` parameter to be present. See [Okta's docs](https://developer.okta.com/docs/reference/api/oidc/#parameter-details) for their explanation (scroll down to the `state parameter description`).

**Versions used**
Kiali: v1.19
Istio: v1.6.2
Kubernetes flavour and version: Kops on AWS v1.17.6

**To Reproduce**
1. Configure auth strategy to `openid` with appropriate supporting properties
2. Navigate to Kiali console in browser and click on "Log In With OpenID"

**Expected behavior**
Be logged in via OIDC without error

israel-hdez:
@asmith60 is it possible for you to post the metadata of your IdP? I want to learn if the provider exposes in the metadata that `state` is required.

You can download it by going to `https://{your_issuer_uri}/.well-known/openid-configuration`.
Metadata should be public, so it should be safe to post. But if you are using a private IdP, feel free to mask any private data.

asmith60:
I scrubbed some of the info, but I believe all of the relevant stuff should be there.

```json
{
    "response_types_supported": [
        "code",
        "id_token",
        "code id_token",
        "code token",
        "id_token token",
        "code id_token token"
    ],
    "response_modes_supported": [
        "query",
        "fragment",
        "form_post",
        "okta_post_message"
    ],
    "grant_types_supported": [
        "authorization_code",
        "implicit",
        "refresh_token",
        "password"
    ],
    "subject_types_supported": [
        "public"
    ],
    "id_token_signing_alg_values_supported": [
        "RS256"
    ],
    "scopes_supported": [
        "openid",
        "email",
        "profile",
        "address",
        "phone",
        "offline_access",
        "groups"
    ],
    "token_endpoint_auth_methods_supported": [
        "client_secret_basic",
        "client_secret_post",
        "client_secret_jwt",
        "private_key_jwt",
        "none"
    ],
    "claims_supported": [
        "iss",
        "ver",
        "sub",
        "aud",
        "iat",
        "exp",
        "jti",
        "auth_time",
        "amr",
        "idp",
        "nonce",
        "name",
        "nickname",
[...]

PR #2963, opened 2020-07-03
Add `state` parameter when requesting autorization to OpenId IdP

israel-hdez:
State parameter is a recommended parameter in the OpenId Spec. Despite just "recommended", some IdPs are _requiring_ it. So, adding `state` parameter for better compatibility.

Fixes kiali/kiali#2933

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -213,7 +213,7 @@ func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
 	// Check if the nonce cookie is present
 	nonceCookie, err := r.Cookie(openIdNonceCookieName)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


handlers/authentication.go
```diff
@@ -213,7 +213,7 @@ func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
 	// Check if the nonce cookie is present
 	nonceCookie, err := r.Cookie(openIdNonceCookieName)
 	if err != nil {
-		RespondWithError(w, http.StatusUnauthorized, "No nonce code present. Login window timed out.")
+		RespondWithError(w, http.StatusBadRequest, "No nonce code present. Login window timed out.")
```
cfcosta, line 216: These make a lot more sense :)

PR #2966, opened 2020-07-06
Format goimports

jotak:
This is for auto-formatting imports.

This "Works for me"© but would be nice to test by others. Especially I did already have goimports in my go bin path, not sure if it's the same for everyone.

To test it, mess around with some imports then run `make format`.

First commit includes the new scripts, second commit holds the generated changes.

Note 1: I'm not a Makefile guru; had some troubles with having multilines awk within the makefile, so I ended up putting everything in a separate shell script
Note 2: I'm not an awk wizard either. My awk file looks probably ugly to others. Feel free to suggest changes.

### DIFF
--- a/business/checkers/authorization/service_checker_test.go
+++ b/business/checkers/authorization/service_checker_test.go
@@ -3,13 +3,13 @@ package authorization
 import (
 	"testing"
 
-	"github.com/kiali/kiali/config"
-	"github.com/kiali/kiali/models"
-	"github.com/kiali/kiali/tests/data"
-
 	"github.com/stretchr/testify/assert"
 	core_v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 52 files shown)


make/fix_imports.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
jmazzitelli, line 1: I think you want to put this script in `hack/` directory. We already do this kind of thing in our makefiles - specifically, see:

https://github.com/kiali/kiali/blob/master/make/Makefile.build.mk#L20

where we call `hack/check_go_version.sh` to make sure the make build has a supported Go version. That script is here in the hack directory: https://github.com/kiali/kiali/blob/master/hack/check_go_version.sh

Other than that, all the imports look prestine now :)

jotak: done!

jotak:
thanks!

Issue #2925, opened 2020-06-23
OpenId authorization with Auth0


mikolajwrzesinski:
**Describe the bug**

I'm trying to configure openid authorization using Auth0. I'm already using Auth0 to authorize a cluster. Kiali has been set to openid auth strategy with the same data as the cluster. It looks like Auth0 implicit flow doesn't return all the data Kiali wants. 

https://auth0.com/docs/flows/concepts/implicit

![104439120_570255956992565_1761605522059263736_n](https://user-images.githubusercontent.com/8101457/85363543-815e0100-b564-11ea-8a39-51bdcfd6aa0c.png)

Auth0 returns only `id_token` after successfull login but `OAuthLogin` service expects `access_token` hash parameter.
```
if (window.location.hash.startsWith('#access_token')) {
  return AuthResult.CONTINUE;
} else {
  return AuthResult.HOLD;
}
```
It is possible for Auth0 to return both `id_token` and `access_token`, but then different `response_type` must be used.

![104768864_266348537924540_2134122429249052415_n](https://user-images.githubusercontent.com/8101457/85364370-55437f80-b566-11ea-822e-5afec6adc5e3.png)

Going deeper into code I can see that this parameter is hardcoded in the backend `handers/authentication.go` 

```
redirectUri := fmt.Sprintf("%s?client_id=%s&response_type=id_token&redirect_uri=%s&scope=%s&nonce=%s",
	authorizationEndpoint,
	url.QueryEscape(conf.Auth.OpenId.ClientId),
	url.QueryEscape(httputil.GuessKialiURL(r)),
	url.QueryEscape(scopes),
	url.QueryEscape(fmt.Sprintf("%x", nonceHash)),
)
```

**Versions used**
Kiali: 1.19
Istio: 1.6.3
Kubernetes flavour and version: vanilla 1.18.3

**To Reproduce**
Setup Kiali to use openid auth startegy with Auth0 provider.

**Expected behavior**
Kiali shoud authenticate the user only with `id_token`, just like Kubernetes cluster.

jshaughn:
cc @israel-hdez , @jmazzitelli

israel-hdez:
Essentially, the front-end is blocking the authentication because of the missing `access_token`, which shouldn't be required for OpenId.

OK, I'll fix this.

israel-hdez:
@mikolajwrzesinski The fix is going to be released today in Kiali v1.20.

mikolajwrzesinski:
[...]

PR #2981, opened 2020-07-12
Make expires in optional for OpenId authentication

### LINKED ISSUE #2954: OpenID authentication strategy should not require "expires_in" in the callback
As noted in comment this comment of another issue: https://github.com/kiali/kiali/issues/2925#issuecomment-650837094, OpenID is not working properly if the IdP does not include the `expires_in` URL parameter when invoking the callback url. This leads to a failed authentication.

[Section 3.2.2.5 of the OpenID spec](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse) states that `expires_in` parameter is *optional*. Thus, authentication should succeed even if it's not present.

cfcosta:
Closes #2954

Fix a problem where some authentication providers (such as Okta) do not provide an expiration date for the token. It is optional on the spec, so we make it optional here as well with a value of **1 day**.

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -25,6 +25,7 @@ import (
 const (
 	missingSecretStatusCode = 520
 	openIdNonceCookieName   = config.TokenCookieName + "-openid-nonce"
+	defaultSessionDuration  = 86400
 )
 
 type AuthenticationHandler struct {
@@ -242,8 +243,7 @@ func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
 
 	token := r.Form.Get("id_token")
 	state := r.Form.Get("state")
-	expiresIn := r.Form.Get("expires_in")
-	if token == "" || expiresIn == "" {
+	if token == "" {
 		RespondWithError(w, http.StatusBadRequest, "Token is empty or invalid.")
 		return false
 	}
@@ -252,10 +252,17 @@ func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
 		return false
 	}
 
-	expiresInNumber, err := strconv.Atoi(expiresIn)
-	if err != nil {
-		RespondWithDetailedError(w, http.StatusBadRequest, "Token expiration is invalid.", err.Error())
-		return false
+	expiresIn := r.Form.Get("expires_in")
+	expiresOn := time.Now().Add(time.Second * time.Duration(defaultSessionDuration))
+
+	if expiresIn != "" {
+		expiresInNumber, err := strconv.Atoi(expiresIn)
+		if err != nil {
+			RespondWithDetailedError(w, http.StatusBadRequest, "Token expiration is invalid.", err.Error())
+			return false
+		}
+
+		expiresOn = time.Now().Add(time.Second * time.Duration(expiresInNumber))
 	}
 
 	// CSRF mitigation
@@ -286,8 +293,6 @@ func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
 	}
 
 	// Create business layer using the received id_token
-	expiresOn := time.Now().Add(time.Second * time.Duration(expiresInNumber))
-
 	business, err := business.Get(token)
 	if err != nil {
 		RespondWithDetailedError(w, http.StatusInternalServerError, "Error instantiating the business layer", err.Error())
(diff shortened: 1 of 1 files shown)


mikolajwrzesinski:
I think this issue should be solved differently. As noted in https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse `expires_in` is the expiry time of the Access Token, not ID Token. To validate `id_token` you need to analize `exp` claim of the token.

cfcosta:
> I think this issue should be solved differently. As noted in https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse `expires_in` is the expiry time of the Access Token, not ID Token. To validate `id_token` you need to analize `exp` claim of the token.

Thanks for the feedback! I need to investigate a bit further to make sure, but I think you're right.

@israel-hdez what do you think?

cfcosta:
@mikolajwrzesinski commited a draft on how that should look like, what do you think? I still need to test it a little bit more to be confident, but should be ok.

jmazzitelli:
don't forget to make sure the openid integration test passes - making necessary updates to it as appropriate: https://github.com/kiali/kiali-operator/tree/master/molecule/openid-test

cfcosta:
@jmazzitelli will do it.

handlers/authentication.go
```diff
@@ -285,9 +279,37 @@ func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
...
+			RespondWithDetailedError(w, http.StatusBadRequest, "Token exp claim is present, but invalid.", err.Error())
+			return false
+		}
+
+		expiresOn = time.Now().Add(time.Second * time.Duration(expiresInNumber))
+	}
 
+	// Then we check the expiration date for the access on the access token
+	// We give priority to it because it is what we use to access the cluster,
+	// so if it is valid we are ok.
```
israel-hdez, line 299: @mikolajwrzesinski has already noted that the OpenId spec tells that the `expires_in` parameter is tied to the `access_token` (see https://openid.net/specs/openid-connect-core-1_0.html#ImplicitAuthResponse).

For OpenID we are *not* using the `access_token`, but the `id_token`. We can safely ignore `expires_in` parameter. So, this block should be removed.

handlers/authentication.go
```diff
@@ -285,9 +279,37 @@ func performOpenIdAuthentication(w http.ResponseWriter, r *http.Request) bool {
...
 	}
 
-	// Create business layer using the received id_token
-	expiresOn := time.Now().Add(time.Second * time.Duration(expiresInNumber))
+	// Set a default value for expiration date
+	expiresOn := time.Now().Add(time.Second * time.Duration(defaultSessionDuration))
+
+	// We check for the expiration date from the claim
+	if expClaim := idTokenClaims["exp"].(string); expClaim != "" {
+		expiresInNumber, err := strconv.Atoi(expClaim)
```
israel-hdez, line 287: Can you double-check this?
AFAIK, the `exp` claim is a Unix timestamp. If this is true, you should take the value as is (as the expiration time), rather than using it as a duration.

handlers/authentication.go
cfcosta: Since we do use the `access_token` to to the requests to kubernetes, wouldn't it be the case where the expiration provided on `expires_in` be more reliable than the exp on the id_token?

israel-hdez: The `id_token` is the one used for OpenID (see the checks at https://github.com/kiali/kiali/pull/2981/files#diff-7b14eebb8414820b40c718efcf921867R244-R249). The `access_token` is ignored in this case. So, `expires_in` is not that reliable.

For OpenShift it is different. In OpenShift, we DO use the `access_token` and it's better to use `expires_in`. But for this PR, we should move to using the `exp` claim.

israel-hdez: See also an already fixed issue: https://github.com/kiali/kiali/issues/2925. It's mentioned that some IdPs won't send an `access_token`.

handlers/authentication.go
cfcosta: Should be correct now.

cfcosta:
@israel-hdez can you verify it again?

cfcosta:
Build has failed, looking at why.

cfcosta:
@israel-hdez this latest commit should take care of the panic.

mikolajwrzesinski:
Thank you guys! v1.21 works with Keycloak

jshaughn:
@mikolajwrzesinski Excellent news!

PR #2985, opened 2020-07-13
Fix Snyk

### LINKED ISSUE #2965: Snyk reports in backend kiali repo are misleading
Just an example of a PR I'm working on:
https://github.com/kiali/kiali/pull/2936

Tracis CI is ok but the Snyk reports are out-of-sync/wrong.

I think our repos should try to be clean, I mean, I prefer a no-report than a false-positive-report that we need to ignore to progress.

So, with this issue I open the discussion what we want to do with this tool (continue using in both backend/frontend, only frontend, etc).

israel-hdez:
Running Snyk within CI, because Snyk's GitHub integration does not support go modules.

Once this is merged, and after the first Snyk CLI run completes, we should see a new entry in the Snyk dashboard for Kiali. Thereafter, we can disable Snyk's GitHub integration for this repo (and leave enabled the integration for the front-end). On the web dashboard, we should be able to see the dependency tree alike in the front-end case.

This is not that ideal because this will run a Snyk check and create a snapshot on each merged PR (while the GitHub integration does a full check only if dependencies change). But it's OK from the point of view that Snyk is working again.

See an example of a run in https://travis-ci.org/github/kiali/kiali/builds/708086245 - there is a new "Snyk" job doing the checks. It says `✓ Tested 293 dependencies for known issues, no vulnerable paths found.` :-)

Since it's a new job, we *may* get a notification on IRC if Snyk check fails with a vulnerability (but I'm not sure, it depends on the exit code of the snyk CLI). Not sure if we later will receive a notification by e-mail.

Fixes #2965 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,21 @@ jobs:
     - arch: amd64
     - arch: arm64
     - arch: ppc64le
+    - name: Snyk
+      if: branch == master AND fork == false
+      node_js:
+        - 12
+      env:
+        - NODE_OPTIONS=--max-old-space-size=8192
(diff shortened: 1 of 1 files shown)


israel-hdez:
We will loose the Snyk indicator from the GitHub checks section. But well... as it is right now, it's really doing any good checks. So, I think having it inside a job in CI much better that what we have now.

israel-hdez:
First run suceeded: https://travis-ci.org/github/kiali/kiali/builds/708427593.
It's now visible in https://app.snyk.io/org/kiali/projects (make sure to check "Without issues" at the left).

I removed the GitHub integration for this repo to keep only the CLI entry.

PR #3012, opened 2020-07-21
CVE-2020-14040

jshaughn:
https://issues.redhat.com/browse/MAISTRA-1636

Marking as backport to 1.12 but note that this can not be cherry-picked because master is using go modules and 1.12 is using glide for dep handling.  The 1.12 fix will need to be manual.


pbajjuri20:
Tested and passed on Kiali UI 1.22.0 (e1008d8be8c0768414d00ce6e26ce5c6560d6354)

pbajjuri20:
Tested and Passed, LGTM!

PR #3033, opened 2020-07-23
Supporting Iter8-v0.2

jadeyliu939:
** Describe the change **

1. Add Host section in CR
2. Remove RoutingPreference in the CR ( Iter8 v0.1)
3. Enforce Supporting level. (Iter8 v0.2)

*** Refer to Issue
#2007 

** Backwards incompatible? **
With this changes, we only support Iter8 v0.2 and above.
Old version of Kiali-UI will work with these changes, but will not be able to take advantage of new feature introduced in V0.2

*** Note ***

### DIFF
--- a/business/iter8.go
+++ b/business/iter8.go
@@ -2,7 +2,6 @@ package business
 
 import (
 	"encoding/json"
-	"github.com/hashicorp/go-version"
 	"regexp"
 	"strconv"
 	"strings"
@@ -16,6 +15,7 @@ import (
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
+	"github.com/kiali/kiali/status"
 )
 
 type Iter8Service struct {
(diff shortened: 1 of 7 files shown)


business/iter8.go
```diff
@@ -24,19 +25,60 @@ type Iter8Service struct {
...
+	var controllerImgVersion string
+	var analyticsImgVersion string
 	promtimer := internalmetrics.GetGoFunctionMetric("business", "Iter8Service", "GetIter8Info")
 	defer promtimer.ObserveNow(&err)
 
 	conf := config.Get()
 
 	// It will be considered enabled if the extension is present in the Kiali configuration and the CRD is enabled on the cluster
 	if conf.Extensions.Iter8.Enabled && in.k8s.IsIter8Api() {
+		if kialiCache != nil && kialiCache.CheckNamespace("iter8") {
```
lucasponce, line 38: "iter8" literal should come from the config https://github.com/kiali/kiali/blob/master/config/config.go#L162 
It can be initialized with "iter8" if not present in the Kiali yaml config and we can propagate it in the future in the operator if we want to change it.

lucasponce: Here we assume that if iter8 extension is enabled every user has access.
That may be fine for this PR, but in the future we may check if "iter8" namespace is accessible (invoking a GetNamespace() first).

But I'm ok without dealing with that check know, it's not a bad assumption that iter8 should be accessible for now.

business/iter8.go
```diff
@@ -291,83 +365,24 @@ func (in *Iter8Service) GetAnalyticPort() int {
...
-	defer promtimer.ObserveNow(&err)
-	istioCfg, err := in.businessLayer.IstioConfig.GetIstioConfigList(IstioConfigCriteria{
-		IncludeGateways:        true,
-		IncludeVirtualServices: true,
-		Namespace:              namespace,
-	})
-	rf := core_v1.ObjectReference{}
-	if err != nil {
-		return rf, err
+func validateVersion(requiredVersion string, installedVersion string) bool {
```
lucasponce, line 368: I wonder if you want just to reuse existing function:
https://github.com/kiali/kiali/blob/master/status/versions.go#L80

And just make it public to be imported here.

Issue #2982, opened 2020-07-13
Renamed k8s services are persisted within the Graph UI


robert-blackman:
**General remarks**

Similar to [this](https://github.com/kiali/kiali/issues/2783#issuecomment-647266321) issue. 

Same repro steps. Except when renaming the k8s service. i.e. `my-svc-v1` -> `my-svc-v2`.

Same expected behaviours.

**Versions used**
Kiali: 1.19.0
Istio: 1.6.3
Kubernetes flavour and version: AWS EKS 1.15

Happy to provide debug info if requested. Thanks

jshaughn:
@robert-blackman So it's the same situation where you have a service node in the graph, with no incoming or outgoing edges?

robert-blackman:
@jshaughn I can confirm it is the same situation, when there is no incoming or outgoing traffic. This is with the old service name. So `my-svc-v1` renamed to `my-svc-v2` the `v1` is persisted. Additionally, i've noticed that there appears to be some discrepancies when faceting using different durations, whereby the old nodes exist within some time periods but not others. Perhaps this indicates that the old nodes are only visible when there is no traffic going to the new nodes?

jshaughn:
@robert-blackman With the exception of "Unused Nodes" I'm fairly sure you should never see an orphan node, especially a service node.  (Unused nodes are those that have never seen traffic but seem to be validly defined, displayed only when the option is enabled.  They are rendered using a dashed border).   But the fact is that any metrics reported into Prometheus will be there until the time that they are dropped via a time-to-live setting or are somehow cleaned out in some other fashion (Kiali does not do any cleaning).  And so given a time-period that extends into the past, it's quite possible that there could be metrics for both the old and new service.  It may have something to do with that sort of scenario, I'll need to investigate.

robert-blackman:
@jshaughn Also, the application itself is not renamed, only the k8s service. I should also mention that when I say "old nodes", I mean _all_ of the old nodes, going back many weeks. Thanks a lot for looking into this!

jshaughn:
[...]

PR #3043, opened 2020-07-28
Ensure removal of leaf service nodes with no incoming traffic

jshaughn:
Because dead node removal can then make other nodes eligible we now iterate until no nodes need to be removed.

Fixes https://github.com/kiali/kiali/issues/2982

This isn't easy to test unless you are suffering from the issue.  But I have added new test code to make sure the use case is handled.  So, I think we just need a visual pass of the code and a regression test from QE to ensure that all of the expected graph nodes always show up.

### DIFF
--- a/graph/telemetry/istio/appender/dead_node.go
+++ b/graph/telemetry/istio/appender/dead_node.go
@@ -10,7 +10,8 @@ const DeadNodeAppenderName = "deadNode"
 // - nodes for which there is no traffic reported and a backing workload that can't be found
 //   (presumably removed from K8S). (kiali-621)
 //   - this includes "unknown"
-// - service nodes that are not service entries (kiali-1526) and for which there is no incoming
+// - service nodes that are not service entries (kiali-1526), egress handlers and for which there is no
+//   incoming traffic or outgoing edges
 //   error traffic and no outgoing edges (kiali-1326).
 // Name: deadNode
 type DeadNodeAppender struct{}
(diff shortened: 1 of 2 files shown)


graph/telemetry/istio/appender/dead_node.go
```diff
@@ -32,11 +33,16 @@ func (a DeadNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *g
 		namespaceInfo.Vendor[workloadListKey] = &workloadList
 	}
 
-	a.applyDeadNodes(trafficMap, globalInfo, namespaceInfo)
+	// Apply dead node removal iteratively until no dead nodes are found.  Removal of dead nodes may
+	// alter the graph such that new nodes qualify for dead-ness by being orphaned, lack required
+	// outgoing edges, etc.. so we repeat as needed.
+	applyDeadNodes := true
+	for applyDeadNodes {
+		applyDeadNodes = a.applyDeadNodes(trafficMap, globalInfo, namespaceInfo) > 0
```
jmazzitelli, line 41: Scary - if this a.applyDeadNodes code has a problem or some graph has some weirdness, this will lock the server in an infinite loop. For example, if applyDeadNodes fails to remove a dead node, over and over we go.

jshaughn: well there are only so many nodes :)  If we remove all of them then we're done.   I'm not sure what you are suggesting, this is basically a while-loops, lots of loops, recursive algorithms, whatever, are hosed if the invariant is never satisfied.  I'm not adding some additional maxTries if that's what you are intoning.

jshaughn: Not that it was necessary but to ensure I never have to listen to you whine about it, I added a check.

pbajjuri20:
Tested and Passed on Kiali UI 1.22.0 (e4a7e060b63d6f369575284f8fbf50ee7c97372c)

pbajjuri20:
Tested and Passed, LGTM!

Issue #3017, opened 2020-07-22
remove login and ldap auth strategy


jmazzitelli:
We have already deprecated the login auth strategy. We should remove it soon. This would involve removing it from the deploy script, the makefile, the operator playbook, and the docs (and I might be missing other places, too).

Same with ldap, now that openid can replace it.

jmazzitelli:
@israel-hdez knows where the touch points are - particularly on the frontend and backend. I also know where they are on the operator side.

jshaughn:
Adding to the sprint as it was deprecated 2 releases ago and we need it removed prior to the next downstream release.

jmazzitelli:
The following PRs implement this fix:
* backend: https://github.com/kiali/kiali/pull/3044
* frontend: https://github.com/kiali/kiali-ui/pull/1861
* operator: https://github.com/kiali/kiali-operator/pull/99
* docs: https://github.com/kiali/kiali.io/pull/270

NOTE: the backend, operator, and docs PRs are all rebased on [the recent helm work](https://github.com/kiali/kiali/issues/195) - do not merge any of the above until that helm work is merged first.

jshaughn:
I want to hold off on this until we ensure OpenID is working for folks.  Setting as wait-for-eng until https://github.com/kiali/kiali/issues/3042 is resolved.

jmazzitelli:
We also want to hold off until we fix the `handlers/authentication_test.go` unit tests - see my comment: https://github.com/kiali/kiali/pull/3044#issuecomment-665266568

jmazzitelli:
I'm assigning this to @israel-hdez for him to look at the tests in authorization_test.go.

jmazzitelli:
@israel-hdez I changed all the PRs from "draft" to "open" (they are now ready for review). I made you a reviewer on all of them (there are 4 PRs - see the PR list in https://github.com/kiali/kiali/issues/3017#issuecomment-665273600)

israel-hdez:
@jmazzitelli OK, I'll start the reviews.

jmazzitelli:
all work done and PRs merged. Au revoir, ldap and login.

PR #3044, opened 2020-07-28
remove login and ldap auth strategies

jmazzitelli:
https://github.com/kiali/kiali/issues/3017


jmazzitelli:
some of the tests still fail. Need to figure out another way for the tests to login (they all used the "login" strategy to get into the server).

israel-hdez:
Hmmm... Looking at the failing tests, I don't think they are mockable.
Most tests (if not all) of the `authentication_test.go` file could be classified as system/integration tests, because they are testing the requests rather than functions. Thus, these tests need a full stack setup. The `login` strategy was an exception, because the k8s api wasn't used on that strategy. But regarding the other strategies... well... that's the reason why they are not being tested here (because they would need a running k8s cluster) :-(

So, I think we will need to remove most of the failing tests and rely on the molecule tests. I see we already have `token-test` and `openid-test`. Probably we will need to migrate the logic on these go tests to the molecule tests.

israel-hdez:
Hmm... there is probably a way to mock...

kiali.go
```diff
@@ -203,16 +153,8 @@ func validateConfig() error {
...
 	auth := config.Get().Auth
 	log.Infof("Using authentication strategy [%v]", auth.Strategy)
-	if auth.Strategy == config.AuthStrategyLogin {
-		creds := config.Get().Server.Credentials
-		if creds.Username == "" && creds.Passphrase == "" {
-			// This won't cause Kiali to abort, but users won't be able to log in, so immediately log a warning
-			log.Warningf("Credentials are missing. Create a proper secret. Please refer to the documentation for more details.")
-		}
-	} else if auth.Strategy == config.AuthStrategyAnonymous {
+	if auth.Strategy == config.AuthStrategyAnonymous {
```
israel-hdez, line 156: I think this should do more validation.
Manually changing the configmap to use "login" leads to a running pod. The error in config is evident only after trying to login; this is when Kiali rejects the request with error and logs the misconfiguration. 

I think here we should check for a valid strategy and return an error to drive the pod to an failed state. Otherwise, the UI behavior can be undefined.

jmazzitelli: done.

I left the hack scripts alone - those are for older versions that still support login strategy. At some point we'll want to remove those, though - the cluster-openshift.sh is really old and should never be used anyway.

Issue #3055, opened 2020-07-30
[Istio 1.7] Missing custom dashboards


jotak:
Custom dashboards are missing when I installed istio 1.7 even though the needed metrics are present in prometheus.
Note: when installing, I did *not* set the flag "meshConfig.enablePrometheusMerge=true" (not sure if it's still valid in 1.7).

Here's an example of metric I have which in theory should trigger dashboard recognition:

```
vertx_eventbus_handlers{address="init-session",app="ui",instance="172.17.0.17:15020",istio_io_rev="default",job="kubernetes-pods",kubernetes_namespace="default",kubernetes_pod_name="ui-8fc585c8f-gjghn",pod_template_hash="8fc585c8f",project="mesh-arena",security_istio_io_tlsMode="istio",service_istio_io_canonical_name="ui",service_istio_io_canonical_revision="v1",version="v1"}
```

At first glance, looks like the problem is with `kubernetes_namespace` label that has changed. Perhaps we should make that configurable.

Also, TODO: test with install flag meshConfig.enablePrometheusMerge=true

PR #3056, opened 2020-07-30
Add more custom dashboards configurations

jotak:
- New NamespaceLabel field
- Prometheus' Auth now decoupled from main prometheus config
- Reorganized config for custom metrics

Part of #3055

Requires k-charted PR: https://github.com/kiali/k-charted/pull/106
Requires operator PR: https://github.com/kiali/kiali-operator/pull/100

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -34,7 +34,12 @@ func NewDashboardsService(prom prometheus.ClientInterface) DashboardsService {
 
 func DashboardsConfig() (kconf.Config, klog.LogAdapter) {
 	cfg := config.Get()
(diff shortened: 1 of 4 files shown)


jotak:
DNM as changes on the operator is needed.
Adding @jmazzitelli as reviewer => there would be some changes not only in config but also I think in the ansible scripts.
We used to populate the custom prometheus URL from main prometheus URL. My suggestion here is to just leave it blank, and from code we'll pick a non-empty one.
But ideally I'd like also to pre-populate the new "NamespaceLabel" field based on the version of Istio. Not sure if we can do that.

Issue #3001, opened 2020-07-17
KIA0104 error severity


MarkDeckert:
**General Remarks**

When applying an AuthorizationPolicy on an istio-ingressgateway it's quite possible that there is legitimately and deliberately no service or ServiceEntry directly matching the host in the AuthorizationPolicy.

**Bug description**
Let's say you have a Gateway object and a VirtualService object for foo.example.com.  There are two routes configured in the VirtualService - /bar goes to an actual service in the cluster called "bar".  /foo goes to a ServiceEntry called "bar.vms.svc.cluster.local" that refers to an external service.

If I want to set AuthorizationPolicy that applies to both services (or maybe just the external service).  The only way to have it apply to the external service as well is to apply the policy to the istio-ingressgateway itself.

In this example there is no service or ServiceEntry that refers to a "foo.example.com" host.  Only the Gateway object and VirtualService object know about the hostname.

Additionally these objects may not even be in the same namespace as the istio-ingressgateway deployment itself.

Apply an authorization policy on the istio-ingressgateway using 
```
  - to:
    - operation:
        hosts:
          - "foo.example.com"
```

Under the hood the envoy http rbac filter rule results in:
```
{
    "header": {
        "name": ":authority",
        "exactMatch": "foo.example.com"
    }
}
```
Authorization policy works correctly and matches requests that have -- depending on protocol -- the http1.1  host header "foo.example.com" or http2 authority header "foo.example.com" as it should according to the AuthorizationPolicy reference documentation. However in Kiali it reports a KIA0104 error. https://kiali.io/documentation/v1.15/validations/#_kia0104_this_host_has_no_matching_entry_in_the_service_registry

**Expected behavior**
This should not be an error because its a legitimate -- though perhaps obscure -- AuthorizationPolicy configuration. Again, only the Gateway and VirtualService need to know about the hostname.

I propose modifying the KIA0104 error so that it takes into account VirtualService and Gateway entries.

**Versions used**
Kiali: 1.18.2
Istio: 1.6.5

lucasponce:
cc @xeviknal

xeviknal:
hi @MarkDeckert, thanks for taking the time and issue this scenario.
I see a couple of things that might be happening here. Do you have the chance to share the four pieces of config you are saying in the description (obscured if it has to)? Virtual Service, Gateway, ServiceEntry and the AuthzPolicy itself (including the namespaces where they are living). It would really help me to better identify the underlying problem.

[...]

PR #3063, opened 2020-07-31
Support VirtualServices hosts in Authz Policies

xeviknal:
Fixes https://github.com/kiali/kiali/issues/3001

AuthorizationPolicies might apply to specific request matching the target host. This host might not only be internal services or service entries in the mesh but all the possible hosts that the cluster allow to get in (through gateways and virtual services).

This PR removes the false positive of the following scenario where the AuthzPolicy is applied to a host declared in a Virtual Service.

Therefore, this PR includes VirtualService's hosts in the host checking process.

ServiceEntry
```
---
kind: ServiceEntry
metadata:
  name: foo-dev
  namespace: istio-system
spec:
  endpoints:
  # ip is a vm/service outside the cluster
  - address: 10.15.1.121
  exportTo:
  - "."
  hosts:
  - foo-dev.istio-system.svc.cluster.local
  location: MESH_EXTERNAL
  ports:
  - name: http
    number: 49000
    protocol: HTTP
  resolution: STATIC
```
VirtualService:
```
---
kind: VirtualService
metadata:
  name: foo-dev
  namespace: istio-system
spec:
  hosts:
  - foo-dev.example.com
  gateways:
  - foo-dev
  http:
  - name: foo-svc
    match:
    - uri:
        prefix: /foo
    rewrite:
      uri: "/foo"
    route:
    - destination:
        host: foo-dev.istio-system.svc.cluster.local
        port:
          number: 49000
```
Gateway:
```
---
kind: Gateway
metadata:
  name: foo-dev
  namespace: istio-system
spec:
  selector:
    istio: ingressgateway # use istio default controller
  servers:
  - hosts:
    - "foo-dev.example.com"
    port:
      number: 80
      name: http
      protocol: HTTP
    tls:
      httpsRedirect: true # sends 301 redirect for http requests
  - hosts:
    - "foo-dev.example.com"
    port:
      number: 443
      name: https
      protocol: HTTPS
    tls:
      mode: SIMPLE
      serverCertificate: /etc/istio/ingressgateway-certs/tls.crt
      privateKey: /etc/istio/ingressgateway-certs/tls.key
```
And finally AuthorizationPolicy
```
---
kind: AuthorizationPolicy
metadata:
  name: allow-foo
  namespace: istio-system
spec:
  action: DENY
  selector:
    matchLabels:
      app: istio-ingressgateway
  rules:
  - to:
    - operation:
        hosts:
          - "foo-dev.example.com"
        paths:
          - "/foo"
    when:
    - key: request.headers[x-token]
      notValues:
      - "fooreader"
```

Huge thanks to @MarkDeckert for the energy and level of detail in his issue!

### DIFF
--- a/business/checkers/authorization/mtls_enabled_checker.go
+++ b/business/checkers/authorization/mtls_enabled_checker.go
@@ -2,6 +2,7 @@ package authorization
 
 import (
 	"fmt"
+
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/util/mtls"
--- a/business/checkers/authorization/no_host_checker.go
+++ b/business/checkers/authorization/no_host_checker.go
@@ -17,6 +17,7 @@ type NoHostChecker struct {
 	Namespaces          models.Namespaces
 	ServiceEntries      map[string][]string
 	Services            []core_v1.Service
+	VirtualServices     []kubernetes.IstioObject
 }
 
 func (n NoHostChecker) Check() ([]*models.IstioCheck, bool) {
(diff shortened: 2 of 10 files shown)


hhovsepy:
Post merge verification.
Regression testing passed.
Verified the VS validation as well.
Screenshot from 2020-08-03 13-04-54

Issue #2543, opened 2020-03-20
Support RequestAuthentication resources


xeviknal:
Istio 1.5 has introduced the RequestAuthentication CRD.

Upgrade notes to 1.5: https://istio.io/news/releases/1.5.x/announcing-1.5/upgrade-notes/#authentication-policy
Reference: https://istio.io/docs/reference/config/security/request_authentication/

- [x] Add the CRD into Istio Object section
- [ ] Add validations to that object

PR #3083, opened 2020-08-05
Add workload validation support to RequestAuthentication

xeviknal:
Add basic support to RequestAuthentication objects:

- Multiple ns-wide objects found
- Two objects pointing to the same workload
- Workload not found

Screenshot of Kiali Console (30)

Testing scenarios: https://github.com/xeviknal/istio-lab/blob/master/2543.yaml

Fixes https://github.com/kiali/kiali/issues/2543


xeviknal:
I put the DNM label in this one. I'd prefer to aggregate this validation into a generic one and therefore kiali.io won't have 3 duplicated validation descriptions per each object using it.

xeviknal:
Documentation PR has been created: https://github.com/kiali/kiali.io/pull/279. Removing the DNM.
In this PR, sidecars, peerauthentication and AuthorizationPolicies validations should be tested. Specially the ones regarding to the workload checks (same as description).

business/istio_validations.go
```diff
@@ -463,6 +466,15 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
 			}
 			go fetchIstioObjects(&istioDetails.Sidecars, namespace, getSidecars, &wg2, errChan2)
 		}
+		if nsCached && kialiCache.CheckIstioResource(kubernetes.RequestAuthentications) {
```
lucasponce, line 469: This is good.
No blocker, probably good to have a follow up PR to add all validations objects to the cache.
For Istio, only it's cached VirtualServices,DestinationRules,Gateways and Sidecars, perhaps we can add more.
(It's also a tradeoff).

I can prepare a next PR for that, too.

prachiyadav:
Screenshot from 2020-08-07 16-48-15

Issue #1471, opened 2019-08-26
VS/DR can share same host in some cases


lucasponce:
Currently, we warn if the same host is shared by multiple Destination Rules or multiple Virtual Services. There are good reasons to it, but sadly there are also cases where this is acceptable.

Istio documentation gives an instruction to split the definition in certain cases as long as one avoids the potential pitfalls.

We should probably modify our validation processes to accept these multiple hosts in those cases where these pitfalls are not an issue. Otherwise we might see people complain that the istio documentation says "do it" and we say "don't do it".

https://istio.io/help/ops/traffic-management/deploy-guidelines/#multiple-virtual-services-and-destination-rules-for-the-same-host

JIRA:
https://issues.jboss.org/browse/KIALI-2490

bryonwinger:
We have a use case where we have a multi-tenant application where a particular service will be used across many domains. The mesh doesn't otherwise need to be aware of the tenant and all incoming requests are sanitized outside of the mesh. This works as expected when using `*` for hosts in the Virtual Service spec, but we would really like to avoid the warnings in Kiali.

Ideally we want to avoid the need to update the VS every time we add a new host.Is there any way to configure Kiali to disable this warning (either entirely or per resource)?

lucasponce:
Perhaps we can introduce some flag to consider wildcars as this case on Kiali validations.

lucasponce:
@bryonwinger can you share a little bit more your details ?

I mean, I understand that you may have VirtualService A and VirtualService B with host wildcards "*".
So, I understand that you have a different gateway per VS ? or how do you resolve the potential ambiguity ?

I mean if I invoke host.test.com, both VS A and VS B will process it (unless they have different gateways).

I'm thinking loud on this to see how we can refine the validation (perhaps same wildcard but different gateways is valid, but same wildcard with same or no gateway should remain a warning).

I mean, disable a potential validation is an option, but I know what this may fall: just always keep something disabled. I don't discard this at the moment but with more info about use cases perhaps we can adapt the validation to be more useful for your use cases.

(Also, I'd like to code an example like you use, to help us testing more general scenarios).

Thanks

lucasponce:
I've commented with @xeviknal about to relax the warning conditions in Kiali 
https://github.com/kiali/kiali/issues/3074

But I think that probably there can be cases where 2 VS sharing same host could lead in a potential problem, so adding a warning it's something that at least should inform to the user.

bryonwinger:
[...]

PR #3102, opened 2020-08-07
Validations: More than one Virtual Service for same host considers gateways

xeviknal:
Fixes: https://github.com/kiali/kiali/issues/1471

This is a validation on Virtual Services. It appears now when both Virtual Services shares the same gateway and same host. In case, the VS doesn't have Gateway, it is compared among other VS without Gateways.

Another scenario has been covered:
1. 2 Destination Rules pointing to the same host. Each DR defines one subset.
2. 1 Virtual Service pointing to one subset.
3. VS has a `no subset found` validation. (the checker only considered one DR pointing to same host).
(user-gateway/admin-gateway should not be presented with this validation)

Testing scenario: https://github.com/xeviknal/istio-lab/blob/master/1471/issue.yaml
Testing scenario with duplicates: https://github.com/xeviknal/istio-lab/blob/master/1471/issue-dups.yaml

Before:
Screenshot of Kiali Console (34)
Screenshot of Kiali Console (35)
Screenshot of Kiali Console (36)

After:

### DIFF
--- a/business/checkers/virtual_services/single_host_checker.go
+++ b/business/checkers/virtual_services/single_host_checker.go
@@ -92,6 +92,11 @@ func storeHost(hostCounter map[string]map[string]map[string]map[string][]*kubern
 		gwList = []string{"no-gateway"}
 	}
 
+	if !host.CompleteInput {
+		host.Cluster = config.Get().ExternalServices.Istio.IstioIdentityDomain
+		host.Namespace = vs.GetObjectMeta().Namespace
+	}
+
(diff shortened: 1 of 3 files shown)


lucasponce:
Thanks @xeviknal would you mind to not forget to add a change in kiali.io too ?

Now we have updated it, it would be good to review to not forget update features in the site.

xeviknal:
Good point. PR created for documentation: https://github.com/kiali/kiali.io/pull/289

Issue #3093, opened 2020-08-06
Jenkins should release the helm charts


jmazzitelli:
There are two new helm charts in Kiali (operator and server).

The source code for these charts are located in the helm charts github repo at https://github.com/kiali/helm-charts.

They are published via GitHub Pages from that github repo's master branch and they appear here: https://kiali.org/helm-charts 
 (Helm index.yaml is at https://kiali.org/helm-charts/index.yaml)

To publish new helm charts for the community to be able to use for demos and such, you simply have to do this:

```
git clone https://github.com/kiali/helm-charts
cd helm-charts
VERSION="$(VERSION_TO_RELEASE)" $(MAKE) update-helm-charts
git push
```

Jenkins does not release these charts automatically - we have to do it manually. It would be nice to have Jenkins build the new repos and push them to the helm-charts git repo as I show above.

We might have to think about this some more, particularly for z-stream releases because all releases are coming out of the master branch in the helm-charts repo, so we can't utilize and build out of other branches or tags.

PR #3107, opened 2020-08-08
jenkins build for helm charts

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3093

### DIFF
--- a/deploy/jenkins-ci/Jenkinsfile
+++ b/deploy/jenkins-ci/Jenkinsfile
@@ -31,6 +31,9 @@
  *   - OPERATOR_GITHUB_URI
  *      defaultValue: git@github.com:kiali/kiali-operator.git
  *      description: SSH Url of the kiali-operator GitHub repository
+ *   - HELM_GITHUB_URI
+ *      defaultValue: git@github.com:kiali/helm-charts.git
+ *      description: SSH Url of the helm-charts GitHub repository
  *   - QUAY_NAME
  *      defaultValue: quay.io/kiali/kiali
  *      description: The name of the Quay repository to push the release
@@ -49,6 +52,9 @@
  *   - OPERATOR_PULL_URI
  *      defaultValue: https://api.github.com/repos/kiali/kiali-operator/pulls
  *      description: The URL of the GitHub API to use to create pull requests for the operator (changes to prepare for next version)
+ *   - HELM_PULL_URI
+ *      defaultValue: https://api.github.com/repos/kiali/helm-charts/pulls
+ *      description: The URL of the GitHub API to use to create pull requests for the helm charts (changes to prepare for next version)
  *   - NPM_DRY_RUN
  *      defaultValue: n
  *      description: Set to "y" if you want to make a "dry run" of the front-end release process
@@ -61,6 +67,9 @@
  *   - SKIP_OPERATOR_RELEASE
  *      defaultValue: n
  *      description: Set to 'y' if you don't want to release the operator
+ *   - SKIP_HELM_RELEASE
+ *      defaultValue: n
+ *      description: Set to 'y' if you don't want to release the helm charts
  *   - SHOULD_RELEASE_SITE
  *      defaultValue: auto
  *      description: Set to 'y' to force release of the website even for patch versions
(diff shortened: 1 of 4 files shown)


deploy/jenkins-ci/Jenkinsfile
```diff
@@ -75,19 +84,20 @@
...
+  def (backendForkUri, uiForkUri, operatorForkUri,helmForkUri) = ['git@github.com:kiali-bot/kiali.git', 'git@github.com:kiali-bot/kiali-ui.git', 'git@github.com:kiali-bot/kiali-operator.git', 'git@github.com:kiali-bot/helm-charts.git']
+  def (backendDir, uiDir, operatorDir, helmDir, siteDir) = ['src/github.com/kiali/kiali', 'src/github.com/kiali/kiali-ui', 'src/github.com/kiali/kiali-operator', 'src/github.com/kiali/helm-charts', 'src/github.com/kiali/kiali.io']
+  def (backendMakefile, uiMakefile, operatorMakefile, helmMakefile, siteMakefile) = ['deploy/jenkins-ci/Makefile', 'Makefile.jenkins', 'Makefile.operator.jenkins', 'Makefile.helm.jenkins', 'Makefile.site.jenkins']
   def buildUi = params.SKIP_UI_RELEASE != "y"
   def buildBackend = params.SKIP_BACKEND_RELEASE != "y"
   def buildOperator = params.SKIP_OPERATOR_RELEASE != "y"
-  def buildSite = (params.SHOULD_RELEASE_SITE == "auto" && params.RELEASE_TYPE == "minor") || params.SHOULD_RELEASE_SITE != "y"
+  def buildHelm = params.SKIP_HELM_RELEASE != "y"
+  def buildSite = (params.SHOULD_RELEASE_SITE == "auto" && params.RELEASE_TYPE == "minor") || params.SHOULD_RELEASE_SITE == "y"
```
jmazzitelli, line 94: I want to point out that here is a bug fix unrelated to the helm chart build. Notice that the old code would set `buildSite` to true if `SHOULD_RELEASE_SITE` is *not* `y` which is the opposite of what we want. I changed this so if `SHOULD_RELEASE_SITE` is equal to `y` then it will set `buildSite` to true.

deploy/jenkins-ci/Jenkinsfile
```diff
@@ -139,6 +149,24 @@ node('kiali-build') {
         sh "cd ${operatorDir}; git config user.name 'kiali-bot'"
         sh "cp ${backendDir}/deploy/jenkins-ci/${operatorMakefile} ${operatorDir}"
       }
+      if ( buildHelm ) {
+        checkout([
+          $class: 'GitSCM',
+          branches: [[name: 'master']],
```
jmazzitelli, line 155: I would like to point out that notice you cannot define the branch here (there is no HELM_RELEASING_BRANCH). It must always be `master` because that is where GitHub Pages will serve its content from. So `master` is the branch where all the release artifacts must go, such as the helm chart tarballs and the helm `index.yaml`.

deploy/jenkins-ci/Makefile.helm.jenkins
```diff
@@ -0,0 +1,160 @@
...
+
+helm-build-release:
+	@echo "Will build version: $(HELM_VERSION_TO_RELEASE)"
+	sed -i -r 's/^VERSION \?= v.*/VERSION \?= $(HELM_VERSION_TO_RELEASE)/' Makefile
+	$(MAKE) clean update-helm-repos
+
+helm-push-version-tag:
+ifeq ($(PUSH_GITHUB_TAG),y)
+	git add Makefile
+	git add docs
```
jmazzitelli, line 95: This entire Makefile is essentially a direct copy of `Makefile.operator.jenkins` with the exception of:

1. The two lines here at 90 and 95 (where it calls the helm-charts make target `update-helm-repos` and it commits the `docs` folder which is where the helm artifacts are copied so GitHub Pages can find it)
2. Line 71 (`OMIT_VERSION_BRANCH` will always be `y` now - we never want to release into version branches - everything has to be `master` because that is where GitHub Pages gets its content from).
3. Anything dealing with pushing to quay.io has been removed - there are no Helm artifacts getting pushed into quay.

The rest of this file is essentially identical with the obvious exceptions of renaming variables so they say things with "HELM" rather than "OPERATOR".

jmazzitelli:
This is ready for review and merge now. I put some notes in comments above.

jmazzitelli:
> Do we want patch releases to also build the helm charts? If we do, this needs some little additional changes.

I assume patch releases are "special" like they are with the main builds (don't we have to do things manual for patch builds)?

I don't know what would need to be done here for patch builds - submit a PR to a PR if you think some changes need to be made.

Just remember, we can't build out of patch branches - at least, we need to check INTO master in order to publish.

israel-hdez:
> I assume patch releases are "special" like they are with the main builds (don't we have to do things manual for patch builds)?
> 

So, so. The manual part is the backports to the vX.Y branch. Once code is backported, you fire the build with the right params.

> I don't know what would need to be done here for patch builds - submit a PR to a PR if you think some changes need to be made.
> 
> Just remember, we can't build out of patch branches - at least, we need to check INTO master in order to publish.

I think it depends on how much the charts have diverted, no?
I mean, I understand we are tying Kiali v1.22 to Istio v1.7.

So, as an example, assume that the charts evolve for several Kiali releases after v1.22. But, then, assume we want to build Kiali v1.22.something because of a bug. Clearly, we won't want to build helm-charts from master; we will want to build them from the v1.22 branch (because they are the ones that work), yet we will want to publish the build artifacts in `master`.

I think that, for the scope of this PR, we should block building the helm charts for patch releases. I think it requires a few more steps.

jmazzitelli:
> yet we will want to publish the built artifacts in `master`

Right, that's the part I want to make sure people understand. The artifacts need to go into the /docs directory - the make target `update-helm-repos` does this.

>  I think that, for the scope of this PR, we should block building the helm charts for patch releases. I think it requires a few more steps.

That's fine. Just submit a PR to this PR if you know how to do it - which I assume you do :)

deploy/jenkins-ci/Jenkinsfile
```diff
@@ -75,19 +84,20 @@
...
+  def (backendForkUri, uiForkUri, operatorForkUri,helmForkUri) = ['git@github.com:kiali-bot/kiali.git', 'git@github.com:kiali-bot/kiali-ui.git', 'git@github.com:kiali-bot/kiali-operator.git', 'git@github.com:kiali-bot/helm-charts.git']
+  def (backendDir, uiDir, operatorDir, helmDir, siteDir) = ['src/github.com/kiali/kiali', 'src/github.com/kiali/kiali-ui', 'src/github.com/kiali/kiali-operator', 'src/github.com/kiali/helm-charts', 'src/github.com/kiali/kiali.io']
+  def (backendMakefile, uiMakefile, operatorMakefile, helmMakefile, siteMakefile) = ['deploy/jenkins-ci/Makefile', 'Makefile.jenkins', 'Makefile.operator.jenkins', 'Makefile.helm.jenkins', 'Makefile.site.jenkins']
   def buildUi = params.SKIP_UI_RELEASE != "y"
   def buildBackend = params.SKIP_BACKEND_RELEASE != "y"
   def buildOperator = params.SKIP_OPERATOR_RELEASE != "y"
-  def buildSite = (params.SHOULD_RELEASE_SITE == "auto" && params.RELEASE_TYPE == "minor") || params.SHOULD_RELEASE_SITE != "y"
+  def buildHelm = params.SKIP_HELM_RELEASE != "y"
```
israel-hdez, line 93: This should be enough to block Helm on patch releases:

```suggestion
  def buildHelm = params.SKIP_HELM_RELEASE != "y" && params.RELEASE_TYPE != "patch"
```

jmazzitelli: committed.

israel-hdez:
Ouch, something is failing in the real Jenkins. This is the logs chunk of the error:

```
W0810 21:26:09.865107     765 loader.go:223] Config not found: /var/lib/origin/openshift.local.config/master/admin.kubeconfig
==> Linting /home/jenkins/agent/workspace/kiali-workflow/kiali-release/src/github.com/kiali/helm-charts/_output/charts/kiali-operator

1 chart(s) linted, 0 chart(s) failed
W0810 21:26:09.998660     781 loader.go:223] Config not found: /var/lib/origin/openshift.local.config/master/admin.kubeconfig
Error: Invalid Semantic Version
make[1]: *** [.build-helm-chart-operator] Error 1
make[1]: Leaving directory `/home/jenkins/agent/workspace/kiali-workflow/kiali-release/src/github.com/kiali/helm-charts'
make: *** [helm-build-release] Error 2
make: Leaving directory `/home/jenkins/agent/workspace/kiali-workflow/kiali-release/src/github.com/kiali/helm-charts'
```

Probably, should the build also omit helm charts on `edge` builds?
It's the only reason I can think for the `Error: Invalid Semantic Version` message.

jmazzitelli:
The helm versions must follow semver semantics. It can't be something like "edge". We should definitely disable edge builds for helm. We should only be doing releases for new versions (not latest).

israel-hdez:
OK, I'll create a small PR to also block "edge" builds.

PR #3109, opened 2020-08-10
when deleting istio, remove the addons too

jmazzitelli:


jmazzitelli:
The test failure has nothing to do with this change or this hack script in general. merging.

PR #3114, opened 2020-08-11
[hack] be able to tell the hack script what istio addons to install

jmazzitelli:
You can optionally tell the hack script what addons to install. Default is "prometheus grafana jaeger" But you can pass any combination of `grafana`, `jaeger`, and `kiali`.  Note that "prometheus" will always be installed so you do not have to explicitly mention it in the `-a` option:

This will install all addons:

```
install-istio-via-istioctl.sh -a "grafana jaeger kiali"
```

This will install nothing but prometheus:

```
install-istio-via-istioctl.sh -a ""
```

Note: when uninstalling, it will attempt to uninstall everything (since you are destroying Istio, all components must go).

This is something @FilipB asked for.


FilipB:
Tried `./install-istio-via-istioctl.sh -a "jaeger "` and `./install-istio-via-istioctl.sh` both worked fine. Thank you!

Issue #3119, opened 2020-08-12
Namespace Enable/Disable Auto Injection - Fails on OCP cluster


hhovsepy:
Tested on OpenShift Version 4.5.4.

Any action to Enable or Disable Auto Injection on namespace fails with error:
"Project.project.openshift.io "bookinfo" is invalid: metadata.labels[istio-injection]: Invalid value: "disabled": field is immutable, try updating the namespace"
![Screenshot from 2020-08-12 13-56-36](https://user-images.githubusercontent.com/604313/90013384-77739700-dca5-11ea-8636-039923858f42.png)

Manual injection of "oc label namespace bookinfo istio-injection=enabled --overwrite" works for the same user.

![Screenshot from 2020-08-12 14-13-25](https://user-images.githubusercontent.com/604313/90013697-126c7100-dca6-11ea-8abd-8c09481a51f8.png)

[configmap-kiali.txt](https://github.com/kiali/kiali/files/5062549/configmap-kiali.txt)

lucasponce:
> Manual injection of "oc label namespace bookinfo istio-injection=enabled --overwrite" works for the same user.

Ok, I used the "project" resource in OpenShift resources but it seems it's better to use in an uniform way the "namespace" one.

I'll apply the changes.

Thanks @hhovsepy

PR #3120, opened 2020-08-12
Use Namespace resources for updates in all scenarios

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3119

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -232,11 +232,7 @@ func (in *NamespaceService) UpdateNamespace(namespace string, jsonPatch string) 
 		return nil, err
 	}
 
-	if in.hasProjects {
-		_, err = in.k8s.UpdateProject(namespace, jsonPatch)
-	} else {
-		_, err = in.k8s.UpdateNamespace(namespace, jsonPatch)
-	}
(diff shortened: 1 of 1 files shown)


hhovsepy:
@lucasponce it works now, thank you.
Just a small improvement request: When clicking on "Enable/Disable Auto Injection" actions dropdown item, it still keeps the dropdown open, however it updates the button for instance from "Enable.." to "Disable...".
Improvement suggestion: What if it closes the actions dropdown after click and then shows notification in messages area that particular "namespace" is enabled/disabled from auto injection?

lucasponce:
> Just a small improvement request: When clicking on "Enable/Disable Auto Injection" actions dropdown item, it still keeps the dropdown open, however it updates the button for instance from "Enable.." to "Disable...".
Improvement suggestion: What if it closes the actions dropdown after click and then shows notification in messages area that particular "namespace" is enabled/disabled from auto injection?

We can review it, I remember that when using this action in the table view, it seems useful to keep it open and see the update in the labels.

But in any case, we can comment it, I'd separate from this fix which targets the backend code.

The other one would need to be in the UI repo.

Issue #1615, opened 2019-08-29
Mesh status view, proxy-status health


lucasponce:
Probably, complimentary to the overview.
See details in https://github.com/kiali/kiali/issues/400

JIRA:
https://issues.jboss.org/browse/KIALI-2959

PR #3126, opened 2020-08-13
Add proxy status in the health

xeviknal:
Fixes https://github.com/kiali/kiali/issues/1615
Needs https://github.com/kiali/kiali-ui/pull/1884
Needs https://github.com/kiali/kiali-operator/pull/140
Needs https://github.com/kiali/helm-charts/pull/9

Steps to help testing this PR (bottom of the description) :arrow_down: 

This PR incorporates the proxy status to the health. 

Screenshot of Kiali Console (52)
Screenshot of Kiali Console (51)
Screenshot of Kiali Console (50)
Screenshot of Kiali Console (49)
Screenshot of Kiali Console (48)

Steps to Test:
My approach to test this is creating a sidecar object on one workload and update it many times and in a short time period.
I created the following script to force productpage to be in `STALE` status: https://github.com/xeviknal/istio-lab/tree/master/kiali-1615

1. Download both files in the link.
2. Run the script.
3. If you want to start the test again, update the bind-sidecar.yaml to have `bookinfo/reviews-v0` as a host. Go to step 2.

The status shown in Kiali should match the resulting status in the following command: `istioctl proxy-status`. Take in account that the is a cache that might delay some changes.

TODO:
- [x] Add test to the current work
- [x] Proxy status at Application level (list and details page)
- [x] Proxy status at Overview page (app, workload level)
- [x] Edit permissions to allow the pod/proxy subresource (https://github.com/kiali/kiali-operator/pull/140)

Subsequent PR:
- see the [diff](https://istio.io/latest/docs/ops/diagnostic-tools/proxy-cmd/#retrieve-diffs-between-envoy-and-istiod) / [related](https://github.com/kiali/kiali/issues/3203)


lucasponce:
Note that this will have implications with https://github.com/kiali/kiali/issues/1424 work with is expected to be consolidated soon.

jmazzitelli:
TODO: ["Are You Altering a Kiali Server Role Permission?"](https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission) checklist

business/health.go
```diff
@@ -75,7 +75,12 @@ func (in *HealthService) getAppHealth(namespace, app, rateInterval string, query
 	}
 
 	// Deployment status
-	health.WorkloadStatuses = castWorkloadStatuses(ws)
+	health.WorkloadStatuses = ws.CastWorkloadStatuses()
+
+	// Proxy status
+	if fetchRate {
```
jotak, line 81: perhaps you can rename this `fetchRate` since it's not only about rates anymore

business/proxy_status.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	ps := getProxyStatusByName(workloadName, namespace, proxyStatuses)
+	if ps != nil {
+		workloadStatus.ProxyStatus = castProxyStatus(*ps)
+	}
+	return workloadStatus
+}
+
+func (in *ProxyStatus) GetWorkloadsProxyStatuses(namespace string, workloadStatuses []models.WorkloadStatus) []models.WorkloadStatus {
```
jotak, line 29: I'm trying to run your PR, but it seems there are some errors unnoticed due to this function and the one above (`GetWorkloadProxyStatus`), and in spite of this, the health shows as "proxy is synced" in the UI, which seems wrong.

In kiali logs I see: `Error while fetching app health: unknown (get pods istiod-5b9d44c58b-l6nzx)`. Not sure why I have this (I'll investigate more), but anyway, I think here and above you shouldn't ignore the error.

Perhaps instead of taking a `workloadStatus` as param and send it mutated as return, you should just return `[]ProxyStatus` (or map of apps+`[]ProxyStatus`) + the error, and let the caller (`health.go`) handle that error?

business/proxy_status.go
```diff
@@ -0,0 +1,129 @@
...
+		if ps != nil {
+			ws.WorkloadStatus.ProxyStatus = castProxyStatus(*ps)
+		}
+		res[wl] = ws
+	}
+
+	return res, nil
+}
+
+func getProxyStatusByName(name, namespace string, proxyStatus []*kubernetes.ProxyStatus) *kubernetes.ProxyStatus {
```
jotak, line 89: Hmm, if there's several pods per workload, only the first match will be taken?

business/proxy_status.go
```diff
@@ -0,0 +1,129 @@
...
+			return ps
+		}
+	}
+	return nil
+}
+
+func castProxyStatus(ps kubernetes.ProxyStatus) []models.ProxyStatus {
+	statuses := make([]models.ProxyStatus, 0, 4)
+
+	r := reflect.ValueOf(ps)
```
jotak, line 101: I'm not sure why reflection is needed here; aren't those fields explicitly defined in `istio.go` `SyncStatus` ? Why can't we have direct/explicit access on those?

(btw if we have to go through reflection I think some defensive checks are necessary, in the case fields don't exist)

business/proxy_status.go
xeviknal: > in spite of this, the health shows as "proxy is synced" in the UI, which seems wrong.
Good point!

The main problem here is that you need to allow new permissions to your cluster. Specifically, allow GET to `pod/proxy` subresource. I am now working on the operator side to make the change.

> Perhaps instead of taking a workloadStatus as param and send it mutated as return, you should just return []ProxyStatus (or map of apps+[]ProxyStatus) + the error, and let the caller (health.go) handle that error?

It sound pretty obvious as you said now. Since the logic in this file is only about ProxyStatus, why should we add an extra responsability to know about the WorkloadStatus. Yep, like it.

> PS: also in general, it would be nice to embrace using as much as possible pure functions whenever it's possible (and easy) :-)
Good to connect with the thoughts in the article. If you have some more, happy to exchange and read.

business/proxy_status.go
```diff
@@ -0,0 +1,129 @@
...
+	for component, key := range map[string]string{"Cluster": "CDS", "Endpoint": "EDS", "Listener": "LDS", "Route": "RDS"} {
+		cSent := reflect.Indirect(r).FieldByName(fmt.Sprintf("%s%s", component, "Sent")).String()
+		cAck := reflect.Indirect(r).FieldByName(fmt.Sprintf("%s%s", component, "Acked")).String()
+		if xdsStatus := xdsStatus(cSent, cAck); xdsStatus != models.Synced {
```
jotak, line 105: This is arguable but I'd be more for sending all statuses, even when it's "synced". It makes it explicit for the user that Kiali did that check.

jotak (commented):
I've added a couple of comments (I know it's still draft) ; nice work btw it will be an excellent feature!

business/proxy_status.go
xeviknal: You are right. In case of deployment which might have more than one workload. This will make the workload health display the proxy status differently.

business/proxy_status.go
xeviknal: Yes, the fields are specified in the SyncStatus. However, I wanted to avoid the repetition of performing the same action with every pair of fields.

```
if xdsStatus(ps.ClusterSent, ps.ClusterAcked) != models.Synced {
  statuses = append(statuses, models.ProxyStatus{})
}

if xdsStatus(ps.EndpointSent, ps.EndpointAcked) != models.Synced {
  statuses = append(statuses, models.ProxyStatus{})
}

if xdsStatus(ps.ListenerSent, ps.ListenerAcked) != models.Synced {
  statuses = append(statuses, models.ProxyStatus{})
}


if xdsStatus(ps.RouteSent, ps.RouteAcked) != models.Synced {
  statuses = append(statuses, models.ProxyStatus{})
}
```

Is there any other way to perform this differently?
How would you have approached that?

business/proxy_status.go
jotak: > > PS: also in general, it would be nice to embrace using as much as possible pure functions whenever it's possible (and easy) :-)

> Good to connect with the thoughts in the article. If you have some more, happy to exchange and read.

For sure, open a parenthesis (and do some underrated value of code-reviews :-) ). That's one of the principle of functional programming ... We're using go, which isn't a functional-style language so we could just say f*** & YOLO but nonetheless I think the benefits of such principles are valid although we're not forced to applying them.

The article says:

-    They’re easier to reason about
-    They’re easier to combine
-    They’re easier to test
-    They’re easier to debug
-    They’re easier to parallelize

To me, in this particular case the main point is the first, easier to reason about:

Simply, just looking at this function signature `func GetWorkloadProxyStatus(workloadName, namespace string) []ProxyStatus` , as a caller, it looks very obvious: it will find proxy statuses out of my workload & namespace. No doubt.

Whereas `func GetWorkloadProxyStatus(workloadName, namespace string, workloadStatus WorkloadStatus) WorkloadStatus` works by side-effect. Just looking at the function signature isn't enough, as a caller, I'll be asking myself: what do I have to pass as workloadStatus input? Some empty object, something pre-initialized with important values that are going to be used internally? It's less clear, I'll have to step into it to make sure.

Also as you put it, not having to deal with responsibility of `WorkloadStatus` makes it simpler, and will still be working if we want to use the same function in different contexts.

business/proxy_status.go
jotak: I see, perhaps it wouldn't be as much simplified as you'd like, but maybe with some temporary struct?

Something like this:

```go

	type cpntStatus struct {
		name string
		sent string
		ack  string
	}
	cluster := cpntStatus{name: "CDS", sent: ps.ClusterSent, ack: ps.ClusterAcked}
	endpoint := cpntStatus{name: "EDS", sent: ps.EndpointSent, ack: ps.EndpointAcked}
	listener := cpntStatus{name: "LDS", sent: ps.ListenerSent, ack: ps.ListenerAcked}
	route := cpntStatus{name: "RDS", sent: ps.RouteSent, ack: ps.RouteAcked}
	for _, cpnt := range []cpntStatus{cluster, endpoint, listener, route} {
// etc.
```

But to be honest, personally I probably wouldn't have cared much about this repetition (at least as long as it's not very big). Reflection has a cost not only in performances but also in maintenance, being more error-prone (e.g. if in the future we change the name of some fields in the underlying struct: reflection we miss the help of the compiler to detect where they're used,  making refactorings harder and more risky). Surely if we had many more repetitions, the pros of reflection would surpass the cons.

jotak: Could be condensed:

```go
	type cpntStatus struct {
		name, sent, ack string
	}
	for _, cpnt := range []cpntStatus{{"CDS", ps.ClusterSent, ps.ClusterAcked}, {"EDS", ps.EndpointSent, ps.EndpointAcked}, {"LDS", ps.ListenerSent, ps.ListenerAcked}, {"RDS", ps.RouteSent, ps.RouteAcked}} {
```

models/health.go
```diff
@@ -66,10 +66,28 @@ type WorkloadHealth struct {
...
-	CurrentReplicas   int32  `json:"currentReplicas"`
-	AvailableReplicas int32  `json:"availableReplicas"`
+	Name              string        `json:"name"`
+	DesiredReplicas   int32         `json:"desiredReplicas"`
+	CurrentReplicas   int32         `json:"currentReplicas"`
+	AvailableReplicas int32         `json:"availableReplicas"`
+	ProxyStatus       []ProxyStatus `json:"proxyStatus"`
+}
+
+type ProxyStatuses string
```
lucasponce, line 76: Please, don't model a primitive type with an abstraction.

Also, we have this in the pods

```
```

business/proxy_status.go
lucasponce: `business` package has responsability to populate the model.
So, I agree that perhaps it's clear to let to do
```
workloadStatus.ProxyStatus = in.businessLayer.ProxyStatus.GetWorkloadProxyStatus(workload, namespace)
```
Instead to populated it inside the method which is less clear and it tends to add complexity when we want to make changes.

I'm more pragmatic related to functional style: if it helps to simplify something, perfect.

But side effects are not bad in all scenarios and I tend to not generalize.

business/proxy_status.go
lucasponce: Mmm, note that a workload IS not a pod.
Workload is the controller + their pods.
So perhaps I wonder if the proxy status is something that should be tied to Pod instead and summarized into the Workload.

But I'm starting to review the whole feature, so take this comment with a pinch of salt.

lucasponce:
We had a first review with @xeviknal.
The feature itself it's great, we are going to work in two main areas:
- Proxy Status should be linked with Pod level and workload and application should have a similar view as Pod status (where we may have status of multiple proxies). 
- Proxy Status details should be added at Pod level table instead to propagate it to all health level. For health is more interesting to now that there are some pods/proxies unsynced.
- Caching: we need to add this info into the cache because querying this info on every health invocation is not scalable.

aljesusg:
> We had a first review with @xeviknal.
> The feature itself it's great, we are going to work in two main areas:
> 
> * Proxy Status should be linked with Pod level and workload and application should have a similar view as Pod status (where we may have status of multiple proxies).
> * Proxy Status details should be added at Pod level table instead to propagate it to all health level. For health is more interesting to now that there are some pods/proxies unsynced.
> * Caching: we need to add this info into the cache because querying this info on every health invocation is not scalable.

We need to create something similar to go-cache library, something general to store motre things , I think that we'll need it soon for other use cases.
> 
> So, the plan is to continue working on that direction on these PRs, I will help @xeviknal with the cache staff as it's something similar is going to happen with tracing health and it would need some refactoring on that area to support new use cases.

lucasponce:
> We need to create something similar to go-cache library, something general to store motre things , I think that we'll need it soon for other use cases.

It's already created in Kiali :), used today for Kubernetes, Istio and Namespaces caching with interfaces for each domain, and yeap, we are going to extend it to more use cases (proxy status, jaeger results, and potentially prometheus results as well).

jotak:
> It's already created in Kiali :), used today for Kubernetes, Istio and Namespaces caching with interfaces for each domain, and yeap, we are going to extend it to more use cases (proxy status, jaeger results, and potentially prometheus results as well).

So, what kind of cache would it be? With k8s objects there is the "Informer" API which is great because it's automatically updated, can we use the same with proxy status?

For Jaeger results we need to be careful because, unlike Prometheus traces are not immutable, they are still modified after they were created. Just something to keep in mind.

lucasponce:
> So, what kind of cache would it be? With k8s objects there is the "Informer" API which is great because it's automatically updated, can we use the same with proxy status?

I'm afraid nope, so we will have a short time cache as similar we have for namespaces.
I'm working actively on it, probably it can be linked with other events (i.e. pod change) but "Informer" is basically an implementation of the list/watch pattern used.

> For Jaeger results we need to be careful because, unlike Prometheus traces are not immutable, they are still modified after they were created. Just something to keep in mind.

Yes, we can discuss later for this use case, but main idea here is to cache results and avoid repetition of same queries with in a short interval, with the cost that info may be updated for X time.

I don't want to focus this PR in the Jaeger caching my comment is that KialiCache is designed to add more interfaces for these scenarios as they were on the radar in terms of next steps for performance considerations.

xeviknal:
Method to test this: upscale reviews-v2 to 2 pods + kill one pod of the two.

Overview page:
Screenshot of Kiali Console (56)

App list page:
Screenshot of Kiali Console (57)

App details page:
Screenshot of Kiali Console (58)

Workload list page:
Screenshot of Kiali Console (59)

Workload details page (health):
Screenshot of Kiali Console (60)

Workload details page (pods): (forced by backend)
Screenshot of Kiali Console (61)

models/health.go
```diff
@@ -70,6 +70,26 @@ type WorkloadStatus struct {
...
+	NotSent ProxyStatuses = "NOT_SENT"
+	Stale   ProxyStatuses = "Stale"
+	StaleNa ProxyStatuses = "Stale (Never Acknowledged)"
+)
+
+// ProxyStatus gives the sync status of the sidecar proxy.
+// In healthy scenarios all variables should be true.
+// If at least one variable is false, then the proxy isn't fully sync'ed with pilot.
+type ProxyStatus struct {
+	CDS ProxyStatuses `json:"CDS"`
```
lucasponce, line 89: I don't think this mapping "ProxyStatuses -> string" makes sense, a mapping of a primitive should be justify when it's used always as an abstract type and it can be changed for a composed type in the future without changing the APIs.

A clear example is like a PrimaryKey mapping or similar, but it's not the case on this so I suggest to not add that abstraction.

Also in the UI this is mapped directly with the raw type

https://github.com/kiali/kiali-ui/pull/1884/files?file-filters%5B%5D=.ts#diff-0e0ca9c6860e3c885bf7cb495a0ad59bR78

So, I don't see why CDS and other members can be just `string`.

lucasponce (commented):
A minor comment on the code but LGTM

lucasponce (dismissed):
Nice work.
Thanks @xeviknal

lucasponce (approved):
LGTM

Thanks for the latest changes !

Issue #3042, opened 2020-07-28
Kiali not working with OIDC for 1.19 


widdix123:
**Describe the bug**
Configuring OIDC with Kiali configmap , i am not able to login to UI 

![image](https://user-images.githubusercontent.com/38529736/88658608-529df200-d0f1-11ea-9dde-1d1128fe65ac.png)

**Versions used**
Kiali: 1.19
Istio: 1.65 
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9) - 1.17 EKS

**To Reproduce**
1. Configure Configmap of Kiali to point to dex 

```
auth:
  strategy: openid
  openid:
    client_id: example-app
    issuer_uri: https://dex.xxxx.xxxxx.com/dex
    username_claim: email
    scopes: ["openid", "profile", "email", "groups"]

```

2. Create a ClusterRoleBinding to add user to have access to Kiali 

3. Login to Kiali Console and Click on "LogIn with OpenId"

4. It will throw error as "Unauthorised"

**Expected behavior**

I should be able to login to the UI console with OIDC

jmazzitelli:
Can you provide the kiali pod logs so we can see what the error is?

widdix123:
@jmazzitelli  - I dont see much in the logs as it does not seem to update even on errors

```
I0728 11:25:01.105204       1 kiali.go:66] Kiali: Version: v1.19.0, Commit: 000642e202cb61f49b65c5077bd05b3cb6274135
I0728 11:25:01.105893       1 kiali.go:205] Using authentication strategy [openid]
I0728 11:25:01.105957       1 kiali.go:87] Kiali: Console version: 1.19.0
I0728 11:25:01.105974       1 kiali.go:284] Updating base URL in index.html with [/kiali]
I0728 11:25:01.106473       1 kiali.go:265] Generating env.js from config
I0728 11:25:01.110513       1 server.go:67] Server endpoint will start at [:20001/kiali]
I0728 11:25:01.110552       1 server.go:68] Server endpoint will serve static content from [/opt/kiali/console]
I0728 11:25:01.110591       1 metrics_server.go:18] Starting Metrics Server on [:9090]
```

jmazzitelli:
We run a nightly integration test using Dex on minikube and it runs successfully so I wonder if there is something in your Dex setup? Have you been able to install a test app (outside of Kiali) that can perform an OIDC login with it?

[...]

PR #3142, opened 2020-08-20
Add support for cluster API proxies on OpenId strategy

israel-hdez:
Since some cloud providers won't support configuring OpenID integration in the managed Kubernetes, users may want to use workarounds. One of them is to place a proxy in front of the Kubernetes API handing the OpenID auth (an example is kube-oidc-proxy).

This is adding support for such API proxies.

Fixes kiali/kiali#3042

### DIFF
--- a/business/layer.go
+++ b/business/layer.go
@@ -63,10 +63,6 @@ func IsResourceCached(namespace string, resource string) bool {
 	return ok
 }
 
-func GetUnauthenticated() (*Layer, error) {
-	return Get("")
-}
-
 // Get the business.Layer
 func Get(token string) (*Layer, error) {
 	// Kiali Cache will be initialized once at first use of Business layer
--- a/config/config.go
+++ b/config/config.go
@@ -264,6 +264,8 @@ type OpenShiftConfig struct {
 
 // OpenIdConfig contains specific configuration for authentication using an OpenID provider
 type OpenIdConfig struct {
+	ApiProxy              string   `yaml:"api_proxy,omitempty"`
+	ApiProxyCAData        string   `yaml:"api_proxy_ca_data,omitempty"`
 	AuthenticationTimeout int      `yaml:"authentication_timeout,omitempty"`
 	AuthorizationEndpoint string   `yaml:"authorization_endpoint,omitempty"`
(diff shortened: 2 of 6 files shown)


israel-hdez:
User has confirmed these changes work. See https://github.com/kiali/kiali/issues/3042#issuecomment-680077563.

So, I'm expecting only code review and, if needed, a quick smoke test.
Asking for a review on Jupiter side because this touches the k8s client factory code, closely related to the caching mechanism.

lucasponce:
> Asking for a review on Jupiter side because this touches the k8s client factory code, closely related to the caching mechanism.

Thanks @israel-hdez I'll take a look I didn't see the github notification, also feel free to ping me in the IRC / Slack if you see I don't answer within the day :)

lucasponce:
> Asking for a review on Jupiter side because this touches the k8s client factory code, closely related to the caching mechanism.

I don't see risks in your code, it seems ok but I can't test, as by default caching is enabled, I guess if you can perform smoke test you can easily validate regressions.

The only relevant info to know about caching is that it's always using the ServiceAccount token:

```
	if kConfig.InCluster {
		if saToken, err := kubernetes.GetKialiToken(); err != nil {
			return nil, err
		} else {
			cacheToken = saToken
		}
	}
```

Because, otherwise we should keep a cache per namespace x resource x token.

So, as long as the SA token doesn't change it, this PR should be ok.

lucasponce:
Also I realize it needs to rebase, there are a couple of conflicts.

israel-hdez:
@lucasponce OK, the conflict is solved.

And yes, the usage of the ServiceAccount shouldn't be affected.
I already tried this with the new feature turned off and also with an artificial smoke test. I see it worked fine, but I wanted to be sure the caching won't be affected.

lucasponce:
> I already tried this with the new feature turned off and also with an artificial smoke test. I see it worked fine, but I wanted to be sure the caching won't be affected.

I can't test. It looks good, if you can test your feature, as caching is enabled, just simply browsing any List or Details page without errors good be an indication that it works.

israel-hdez:
OK, I already tried and seems to work fine.
I'll move this forward.

lucasponce (approved):
Code seems ok, not tested locally.

Issue #3139, opened 2020-08-20
Kiali access for developer in their Namespace(s) only - Kiali throws errors when checking the Istio deployment status.


ssurovich:
**Describe the bug**
Im using Token based access for now and its working well, in general.  Developers can only see their namespace in the Kiali console, but Kiali seems to insist on checking the status of the Istio deployment and every 10-15 seconds the developer received a warning that they are forbidden to view deployments in the istio-system namespace.

This makes sense - but is there a way to stop Kiali from checking the status of the components so it doesnt spam the developers every 10-15 seconds?

I could potentially add a role to view deployments in the istio-system namespace for service accounts, but Im not sure if this would be the "preferred" way to remediate this?

**Versions used**
Kiali: 1.22.0
Istio: 1.6.6
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9)  GKE On Prem 1.4

**To Reproduce**
Use a service account in a namespace to log into Kiali via the token.  In a few seconds you will start to receive the error in the upper right hand corner.

**Expected behavior**
Developers should see their namespace and services, without a warning about the Istio deployment and the istio-system namespace.

jmazzitelli:
Can you provide the exact messages you are getting, both in the UI console and in the kiali pod logs? Just so we are clear what is repeatedly occuring?

lucasponce:
@ssurovich @jmazzitelli I think this is due is expected that the ServiceMesh / Istio user should have access to the istio-system.

Probably we could see if we can use the SA Kiali account to query this info, but in theory an Istio user should have access to the istio-system (at least that was a requeriment in Maistra - OpenShift scenarios and that was reason we didn't see this before).

I'm familiar with the code, I'll take it to see if it's doable to address this.

PR #3146, opened 2020-08-24
Use cache for fetching Istio Status deployments

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3139

Note, that Istio Status is a perf operation, so using cache is something reasonable.

Cache is enabled by default in all namespaces:

https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L188

In the cornercase that the user has an explicit disabled cache, it will show a warning in Kiali log and it will query the Istio deployments with the user's token.

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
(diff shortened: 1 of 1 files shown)


hhovsepy (approved):
Regression tested.

Issue #3147, opened 2020-08-24
Update Kiali Cache with Istio resources used in validations


lucasponce:
Kiali performs validations on several Istio objects, as this is a potential perf operation we decided to cache those objects (not all Istio resources, but those involved in validations).

We have recently added more resources into the validations so it makes sense to add them into the Kiali Cache defaults.

lucasponce:
It's a minor/trival task, so, I'm going to add it into the backlog and send a PR.

PR #3149, opened 2020-08-25
Update Cache with potential perf scenarios

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3147

Added to the cache:
- StatefulSets
- ConfigMaps
- Istio: Sidecar, PeerAuthentication, RequestAuthentication, AuthorizationPolicy

All of them used for common operations of tls, istio status and validations.

Refactor some logic placed on kubernetes package that makes sense to use it in the business package.

It requires a followup to update operator defaults.

@hhovsepy main testing of this PR is to validate there is no regression or any obvious perf penalty.
Otherwise it should be transparent from feature perspective.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -126,7 +126,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			var gg []kubernetes.IstioObject
 			var ggErr error
 			// Check if namespace is cached
-			if kialiCache != nil && kialiCache.CheckIstioResource(kubernetes.Gateways) && kialiCache.CheckNamespace(criteria.Namespace) {
+			if IsResourceCached(criteria.Namespace, kubernetes.Gateways) {
 				gg, ggErr = kialiCache.GetIstioObjects(criteria.Namespace, kubernetes.Gateways, criteria.LabelSelector)
 			} else {
 				gg, ggErr = in.k8s.GetIstioObjects(criteria.Namespace, kubernetes.Gateways, criteria.LabelSelector)
@@ -145,7 +145,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			var vs []kubernetes.IstioObject
 			var vsErr error
 			// Check if namespace is cached
-			if kialiCache != nil && kialiCache.CheckIstioResource(kubernetes.VirtualServices) && kialiCache.CheckNamespace(criteria.Namespace) {
+			if IsResourceCached(criteria.Namespace, kubernetes.VirtualServices) {
 				vs, vsErr = kialiCache.GetIstioObjects(criteria.Namespace, kubernetes.VirtualServices, criteria.LabelSelector)
 			} else {
 				vs, vsErr = in.k8s.GetIstioObjects(criteria.Namespace, kubernetes.VirtualServices, criteria.LabelSelector)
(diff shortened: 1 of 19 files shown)


business/tls.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/util/mtls"
+	core_v1 "k8s.io/api/core/v1"
```
jmazzitelli, line 11: trivial - requires gofmt - this import needs to move up to the middle section of imports.

business/istio_config.go
```diff
@@ -328,7 +328,14 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 	go func(errChan chan error) {
 		defer wg.Done()
 		if criteria.IncludeAuthorizationPolicies {
-			if ap, apErr := in.k8s.GetIstioObjects(criteria.Namespace, kubernetes.AuthorizationPolicies, criteria.LabelSelector); apErr == nil {
+			var ap []kubernetes.IstioObject
+			var apErr error
+			if kialiCache != nil && kialiCache.CheckIstioResource(kubernetes.AuthorizationPolicies) && kialiCache.CheckNamespace(criteria.Namespace) {
```
jmazzitelli, line 333: I'm seeing code like this repeated a lot:

```
if kialiCache != nil && kialiCache.CheckIstioResource(kubernetes.AuthorizationPolicies) && kialiCache.CheckNamespace(criteria.Namespace)
```

or similar but does not check Istio resource type:

```
if kialiCache != nil && kialiCache.CheckNamespace(namespace)
```

Should we have a utility func that performs this? e.g.

```
func IsCached(kc KialiCache, ns string, rt string) (bool) {
    ok := kc != nil && kc.CheckNamespace(ns)
    if ok && rt != nil {
        ok = kc.CheckIstioResource(rt)
    }
    return ok
}
```

jmazzitelli: Then that repeated code and be replaced with things like:

`if IsCached(kialiCache, criteria.Namespace)`

or

`if IsCached(kialiCache, criteria.Namespace, kubernetes.AuthorizationPolicies)`

lucasponce: Yeah, good suggestion

lucasponce:
> I'm not seeing these UI Errors in master but saw it in the PR on first deployment, and then again on a sanity check. In general backend calls are failing:

This PR needs operator changes, forgot to add in the description, as it seems that some verbs are missing for some resources added into the cache, that should explain that problem.

I'll add the operator PR.

lucasponce:
@jshaughn Even the operator PR is coming there was a bug in the sync logic of the cache that was visible when some of the new resources were not in the config.
Would you mind to test again if you don't see those errors ?

hhovsepy:
Error in Kiali logs, on OpenShift cluster:
[kiali-76b656789-7lnmw-kiali.txt](https://github.com/kiali/kiali/files/5128341/kiali-76b656789-7lnmw-kiali.txt)

lucasponce:
> Error in Kiali logs, on OpenShift cluster, causing kiali pod crash:
> [kiali-76b656789-7lnmw-kiali.txt](https://github.com/kiali/kiali/files/5128341/kiali-76b656789-7lnmw-kiali.txt)

This sounds the same error that @jshaughn and me were having, I thinkg it's addressed in the last commit I've sent.

hhovsepy:
Post merge verification and regression tested.

Issue #3015, opened 2020-07-22
Better link Istio configuration under Workload and Service details.


lucasponce:
Kiali links DestinationRules and VirtualServices object under Kiali Service details page.

This scenario can be extended as there are more networking and security objects that are linked directly at workload and/or service details level.

This issue is to work on:
- [x] Join "Virtual Services" and "Destination Rules" tabs into a single "Istio" tab under Service Details.
-  ~~Extend "Istio" tabs with other Istio objects related with a Service (Networking: Gateway, Security).~~
- [x] Extend "Istio" tabs under Workload details to link specific Istio objects linked with workload selectors only.
- ~~Revisit "Delete all" action to avoid conflicts with shared objects (i.e. do not delete a Gateway, probably it's already done but it would require to check potential regressions)~~

Update:
- Security objects are linked under the Workload details.
- "Delete All" action under Service Details may need more thinking as there are some scenarios which may include new Gateway and PeerAuthentication, so at the moment it will be related to the direct VS/DR objects created, rest of the Istio resources are deleted individually from Istio Config details.

lucasponce:
One option I was thinking about is to have a single tab called "Istio" and put all objects there and have a list of DR, VS, Peer, etc.
In that case, we could normalize the same tab under all App/Workload/Service details.
Also, we have some inconsistencies in the tab as in the VR we only show basic fields meanwhile in the DR there are some sort of mini-summary in the row, that's also an aspect that could be good to normalize.

It doesn't seem a pretty big effort, but something cheap that can help user to understand all Istio configuration related for an element.

In practice is like to have a Istio Config list filtered by the element (app, workload, service).

xeviknal:
I like the idea. However, I do have one concern: there aren't much space in that page for all the filtering and pagination capabilities.
I do like also the fact of standardizing the columns of each row in there. I don't know if the details in there are helping that much.

jshaughn:
> One option I was thinking about is to have a single tab called "Istio"

A complete side-comment, but the idea of compartmentalizing mesh-specific handling could be useful for setting a "vendor" mode across Kiali.

lucasponce:
> A complete side-comment, but the idea of compartmentalizing mesh-specific handling could be useful for setting a "vendor" mode across Kiali.

+1 Yes, in practice is something we address today, for example, specific forms for Istio are located in the "Istio Config" -> "New" objects, meanwhile under the Services the "Wizards" use an abstract layer.

[...]

PR #3157, opened 2020-08-27
Add WorkloadSelector filtering on Istio Config List

lucasponce:
Related to https://github.com/kiali/kiali/issues/3015

Frontend https://github.com/kiali/kiali-ui/pull/1901

Backend support for filtering Istio Objects that support WorkloadSelector labels


business/istio_config.go
```diff
@@ -45,14 +45,77 @@ type IstioConfigCriteria struct {
...
 	IncludeHttpApiSpecBindings    bool
 	IncludeHttpApiSpecs           bool
 	LabelSelector                 string
+	WorkloadSelector              string
+}
+
+func (icc IstioConfigCriteria) Include(resource string) bool {
+	// Flag used to skip object that are not used in a query when a WorkloadSelector is present
+	isWorkloadSelector := icc.WorkloadSelector != ""
+	switch resource {
```
xeviknal, line 62: Similarly to the previous comment in the UI PR review. Do you think there is a way to reduce the switch by using the `reflect` or dynamic access methods? Probably this is because of the architecture of the package so it might change after the already commented refactor.

xeviknal (commented):
LGTM! There is a comment I'd like to hear from @lucasponce though.

business/istio_config.go
lucasponce: I don't think it can be changed without migrating more things in the model.

Basically the IstioConfigCriteria is taking the flags of which types to use because we have a single istio endpoint to fetch all types with a single IstioConfig abstraction.

Also this Criteria is used to decide what to check or not.

Point is that now we have another scenario, IstioConfig which is related/linked with a Workload, so that logic is done on this helper method.

What I don't want is to move this ParseCriteria to typeless, i.e. just storing a "type string" and checking if included or not, I'm less sure, I prefer to keep this.

At some point this package will be revisit, there are deprecated types that will be removed and then we can think more movements, but I think having an Include() and not propagating this complexity beyond the helper is a good idea.

hhovsepy (approved):
Verified.

business/istio_config.go
xeviknal: Yes, I was a bit unsure of this one. I guess that when we revisit this package we will be able to get rid of some snippets like this one.

xeviknal (approved):
LGTM!

Issue #3071, opened 2020-08-03
[Tracing] Tracing tab in Service page


jotak:
The trace tab is being removed from the Service page, because it actually relates to App, not Service: https://github.com/kiali/kiali/issues/3045
But this doesn't mean it cannot be under Services as well ; just that some adjustments are necessary [1]. And we need to fetch the app label related to the service prior to querying jaeger.

So, as an enhancement, add it (back) to the Service Details page as a tab.

[1] post-filtering of Service name, based on operation name?

aljesusg:
Pfff For this ...we can query jaeger to check if "service name" is in the list /api/services that jaeger provides. In that case we can show the tab

jotak:
but this would still be matching against the jaeger service name which corresponds in reality to kiali's app. Whereas Kiali service name is actually visible in jaeger in operation name.

jotak:
I mean, perhaps yes we can do as you suggest and consider it's "good enough".
But we would still miss the services that are != from app names.

PR #3169, opened 2020-09-02
Fetch traces for service

jotak:
Fixes https://github.com/kiali/kiali/issues/3071

- New server endpoint
- New function to get app name for a service
- Filter app traces for service based on operation name in spans
- Update swagger

Frontend PR: https://github.com/kiali/kiali-ui/pull/1905

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -370,3 +370,47 @@ func (in *SvcService) getServiceValidations(services []core_v1.Service, deployme
 
 	return validations
 }
+
+// GetServiceAppName returns the "Application" name (app label) that relates to a service
+// This label is taken from pods, not the service object itself.
+// Consistency accross pods is assumed, ie. only one app label for all the pods under this service (only one will be returned).
+func (in *SvcService) GetServiceAppName(namespace, service string) (string, error) {
+	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
+	// if namespace is accessible from Kiali (Deployment.AccessibleNamespaces)
+	if _, err := in.businessLayer.Namespace.GetNamespace(namespace); err != nil {
+		return "", err
+	}
+
+	svc, eps, err := in.getServiceDefinition(namespace, service)
+	if err != nil {
+		return "", err
+	}
+
+	labelsSelector := labels.Set(svc.Spec.Selector).String()
+	// If service doesn't have any selector, we can't know which are the pods applying.
+	if labelsSelector == "" {
+		return "", nil
+	}
+
+	var pods []core_v1.Pod
+	// Check if namespace is cached
+	// Namespace access is checked in the upper caller
+	if IsNamespaceCached(namespace) {
+		pods, err = kialiCache.GetPods(namespace, labelsSelector)
+	} else {
+		pods, err = in.k8s.GetPods(namespace, labelsSelector)
+	}
+	if err != nil {
+		return "", err
+	}
+
+	appLabelName := config.Get().IstioLabels.AppLabelName
+	for _, pod := range kubernetes.FilterPodsForEndpoints(eps, pods) {
+		if app := pod.Labels[appLabelName]; app != "" {
+			return app, nil
+		}
+	}
+
+	// No app label found
+	return "", nil
+}
--- a/business/services.go
+++ b/business/services.go
@@ -372,8 +372,7 @@ func (in *SvcService) getServiceValidations(services []core_v1.Service, deployme
 }
 
 // GetServiceAppName returns the "Application" name (app label) that relates to a service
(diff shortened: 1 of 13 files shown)


business/services.go
```diff
@@ -370,3 +370,47 @@ func (in *SvcService) getServiceValidations(services []core_v1.Service, deployme
...
+	if _, err := in.businessLayer.Namespace.GetNamespace(namespace); err != nil {
+		return "", err
+	}
+
+	svc, eps, err := in.getServiceDefinition(namespace, service)
+	if err != nil {
+		return "", err
+	}
+
+	labelsSelector := labels.Set(svc.Spec.Selector).String()
```
lucasponce, line 389: Perhaps you can assume in Istio that service selector would contain the app level.
Because, if it's not also present in the Pods result would be the same (empty result).
So in that way no need to fetch pods and endpoints.
(I mean, the only corner case here would be that service selector don't use the app, but probably that would have other implications everywhere -> telemetry, istio config, etc).

jaeger/util.go
```diff
@@ -1,26 +1,29 @@
...
 	"github.com/kiali/kiali/config"
-	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 )
 
-func prepareQuery(u *url.URL, namespace, app string, query models.TracingQuery) {
-	queryApp := app
-	if config.Get().ExternalServices.Tracing.NamespaceSelector && namespace != config.Get().IstioNamespace {
-		queryApp = fmt.Sprintf("%s.%s", app, namespace)
+type JaegerServiceName = string
```
lucasponce, line 14: In other code areas we have discuss to avoid to map primitive types.
It's fine to deal directly with string.
Otherwise code gets populated with a lot of abstractions.
Also, in the code q.Set("service", jsn) is expecting a string, too, so abstraction is weak here.

routing/routes.go
```diff
@@ -373,7 +373,7 @@ func NewRoutes() (r *Routes) {
 			handlers.ServiceDetails,
 			true,
 		},
-		// swagger:route GET /namespaces/{namespace}/apps/{app}/spans traces spansList
+		// swagger:route GET /namespaces/{namespace}/apps/{app}/spans traces appSpans
 		// ---
 		// Endpoint to get Jaeger spans for a given service
```
lucasponce, line 378: I guess this comment would need to be updated.

I see that for spans we have "AppSpans", I wonder if now that there are a "AppTraces" / "ServiceTraces" if that would require a "AppSpans" / "ServiceSpans".

Perhaps not, and "AppSpans" work for both scenarios, but just leaving a note here.

business/services.go
jotak: Well, I *think* we saw that case precisely of having services that don't select pods from the app label, but I'm not totally sure. And I don't think it causes any trouble in istio telemetry. I would prefer to play safe here, and since cache is generally enabled and very likely to already contains this data, the cost of this call should be minor.

jaeger/util.go
jotak: I'm puzzled because I thought the compiler would expect explicit type conversion when using string or something derived from string, but it seems it doesn't. Not sure why, sometimes it does, sometimes not. My intention is to avoid future possible errors where we would call `prepareQuery` with something not formatted as a jaeger service name. But if the compiler doesn't enforce explicit type conversion, it is meaningless... so I'm fine to remove.

business/services.go
lucasponce: But look into the Service Details, it's using the selector for the very same scenario.

lucasponce: I mean if we are using this pattern:
image

I guess using the service selector will just work and we can avoid that for one scenario people fetches the pod and for others the service and others the workload selector.

Also, if app is not present it will be marked with a validation as well.

routing/routes.go
jotak: I'm updating the comment, it works for apps indeed, not services.

That's right that we should as well create endpoints for service spans and workload spans. I'll create a follow-up task.

lucasponce: Ok, so then in this scenario:
image
Spans are fetched by app but it should filter by service, too.

jotak: https://github.com/kiali/kiali/issues/3175

business/services.go
jotak: Ok, you're right, I didn't remember we were assuming this all throughout our code

jotak:
@lucasponce all 3 comments addressed

lucasponce:
> @lucasponce all 3 comments addressed

Thanks, I think it requires an update of the swagger.

lucasponce (approved):

PR #3185, opened 2020-09-08
Traces workload matching, use "hostname" as well

jotak:
This works in general for non-envoy spans.

Related frontend PR: https://github.com/kiali/kiali-ui/pull/1914


jotak:
PS: I should add a couple of unit tests ...

jshaughn (approved):
LGTM

jotak:
thanks
merging right now, it doesn't need to be tied with the frontend PR ... and I need it for other work :-)

Issue #3088, opened 2020-08-05
OIDC - Add support for "Authentication code" flow


ssurovich:
**Is your feature request related to a problem? Please describe.**
Its a problem from a security side - assuming I understand how Kiali integrates with OIDC.
I was trying to use our OIDC provider, OpenUnison, and it failed to work.   We did some tracing and discovered that Kiali wants the OIDC provider to send back the id_token in the response.  The GET response is logged by any system that sees the traffic (ie: the K8s Ingress, the F5, etc...) - So if you can see the log, you have the K8s id_token.

Second - we use short lived id_tokens, and it appears that Kiali's OIDC implementation does not use refresh_tokens?  With our timeout on the id_tokens, a user will need to re-authenticate every few minutes.

**Describe the solution you'd like**
A more secure method to access Kiali using OIDC with support for refresh_tokens.

**Describe alternatives you've considered**
My original thought is to use the auth_header method that the K8s Dashboard supports - This is an easy and secure method, using a reverse proxy that will send the auth_header from a link on our OIDC page.
OR
Change Kiali code to include a response_type=form_post in the auth request to a form post instead of a 302 redirect.

**Additional context**
I cant think of anything, if I can offer any information, please let me know.

If my understanding is way off base, let me know.

israel-hdez:
Hi, @ssurovich 

AFAIK, the `response_type=form_post` is not part of the OpenID spec (probably it's an OAuth feature). At least, that's what I understand from section 3 of this document: https://openid.net/specs/openid-connect-core-1_0.html#Authentication. I wouldn't like to introduce features outside of the OpenID spec, so the only alternative I see is to use `response_type=code` which triggers the "Authorization code" flow, but to support that my understanding is that Kiali will need back-end storage (which does not have). If you have any idea/suggestion/though about how to implement the Authorization code flow without back-end storage (i.e, in a state-less way), please share your ideas!

The current Kiali OIDC implementation is using `nonce` and `state` codes to mitigate replay and CSRF attacks (see https://auth0.com/docs/api-auth/tutorials/nonce and https://auth0.com/docs/protocols/oauth2/oauth-state for some details). I know these are just "mitigations", but all systems logging the traffic shouldn't be able to use the logged credentials to gain access to Kiali -- Kiali should reject because of the codes.

[...]

PR #3189, opened 2020-09-08
Initial implementation of OpenId's auth code flow

israel-hdez:
This initial implementation of the "authorization code" flow for OpenID authentication let's Kiali negotiate authentication with an extenal IdP provider as described in [section 3.1 of the OpenID spec](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth). Kiali is acting as an "unauthenticated client" (i.e. no client secret is used).

This is using AES-GCM encryption to encode the cookie which prevents revealing the OpenId id_token to the network and to the browser.

Related to #3088

### DIFF
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -3,6 +3,7 @@ package business
 import (
 	"crypto/aes"
 	"crypto/cipher"
+	"crypto/sha256"
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
@@ -19,6 +20,10 @@ import (
 	"github.com/kiali/kiali/log"
 )
 
+const (
+	OpenIdNonceCookieName = config.TokenCookieName + "-openid-nonce"
+)
+
 type OpenIdMetadata struct {
 	// Taken from https://github.com/coreos/go-oidc/blob/8d771559cf6e5111c9b9159810d0e4538e7cdc82/oidc.go
 	Issuer      string   `json:"issuer"`
@@ -33,24 +38,107 @@ type OpenIdMetadata struct {
 	ResponseTypesSupported []string `json:"response_types_supported"`
 }
 
+type OpenIdCallbackParams struct {
+	Code          string
+	ExpiresOn     time.Time
+	IdToken       string
+	Nonce         string
+	NonceHash     []byte
+	ParsedIdToken *jwt.Token
+	State         string
+	Subject       string
+}
+
 var cachedOpenIdMetadata *OpenIdMetadata
 
-func ExtractOpenIdClaimsOfInterest(token string) (sub, nonce string, expiresOn time.Time, errorMessage string, detailedError error) {
-	// Parse the received id_token from the IdP
-	parsedIdToken, _, err := new(jwt.Parser).ParseUnverified(token, jwt.MapClaims{})
-	if err != nil {
-		errorMessage = "Cannot parse received id_token from the OpenId provider"
-		detailedError = err
-		return
+func BuildOpenIdJwtClaims(openIdParams *OpenIdCallbackParams) *config.IanaClaims {
+	return &config.IanaClaims{
+		SessionId: openIdParams.IdToken,
+		StandardClaims: jwt.StandardClaims{
+			Subject:   openIdParams.Subject,
+			ExpiresAt: openIdParams.ExpiresOn.Unix(),
+			Issuer:    config.AuthStrategyOpenIdIssuer,
+		},
 	}
+}
+
+func ExtractOpenIdCallbackParams(w http.ResponseWriter, r *http.Request) (params *OpenIdCallbackParams, err error) {
+	params = &OpenIdCallbackParams{}
+
+	// Get the nonce code hash
+	var nonceCookie *http.Cookie
+	if nonceCookie, err = r.Cookie(OpenIdNonceCookieName); err == nil {
+		params.Nonce = nonceCookie.Value
+
+		hash := sha256.Sum224([]byte(nonceCookie.Value))
+		params.NonceHash = make([]byte, sha256.Size224)
+		copy(params.NonceHash, hash[:])
+	}
+
+	// Parse/fetch received form data
+	err = r.ParseForm()
+	if err != nil {
+		err = fmt.Errorf("error parsing form info: %w", err)
+	} else {
+		// Read relevant form data parameters
+		params.Code = r.Form.Get("code")
+		params.IdToken = r.Form.Get("id_token")
+		params.State = r.Form.Get("state")
+	}
+
+	// Cleanup: Delete the nonce cookie since we no longer need it.
+	deleteNonceCookie := http.Cookie{
+		Name:     OpenIdNonceCookieName,
+		Expires:  time.Unix(0, 0),
+		HttpOnly: true,
+		Path:     config.Get().Server.WebRoot,
+		SameSite: http.SameSiteStrictMode,
+		Value:    "",
+	}
+	http.SetCookie(w, &deleteNonceCookie)
+
+	return
+}
+
+//func CheckOpenIdImplicitFlowParams(params *OpenIdCallbackParams) string {
+//	if params.NonceHash == nil {
+//		return "No nonce code present. Login window timed out."
+//	}
+//	if params.State == "" {
+//		return "State parameter is empty or invalid."
+//	}
+//	if params.IdToken == "" {
+//		return "Token is empty or invalid."
+//	}
+//
+//	return ""
+//}
+
+func CheckOpenIdAuthenticationCodeFlowParams(params *OpenIdCallbackParams) string {
+	if params.NonceHash == nil {
+		return "No nonce code present. Login window timed out."
+	}
+	if params.State == "" {
+		return "State parameter is empty or invalid."
+	}
+	if params.Code == "" {
+		return "No authentication code is present."
+	}
+
+	return ""
+}
+
+func ParseOpenIdToken(openIdParams *OpenIdCallbackParams) error {
+	// Parse the received id_token from the IdP
+	parsedIdToken, _, err := new(jwt.Parser).ParseUnverified(openIdParams.IdToken, jwt.MapClaims{})
+	if err != nil {
+		return fmt.Errorf("cannot parse received id_token from the OpenId provider: %w", err)
+	}
+	openIdParams.ParsedIdToken = parsedIdToken
 	idTokenClaims := parsedIdToken.Claims.(jwt.MapClaims)
 
-	if nonceClaim, ok := idTokenClaims["nonce"]; !ok {
-		nonce = nonceClaim.(string)
-	}
-
 	// Set a default value for expiration date
-	expiresOn = time.Now().Add(time.Second * time.Duration(config.Get().LoginToken.ExpirationSeconds))
+	openIdParams.ExpiresOn = time.Now().Add(time.Second * time.Duration(config.Get().LoginToken.ExpirationSeconds))
 
 	// If the expiration date is present on the claim, we use that
 	expiresInNumber := int64(0)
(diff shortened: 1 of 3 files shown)


jmazzitelli:
Oooo.. interesting enhancement! We should contact the Kiali users who expressed interest in this feature to see if they can test it out.

israel-hdez:
@jmazzitelli Merging this may break the open-id molecule tests, I'm not sure. I'll fix later if they fail.

israel-hdez:
@cfcosta @jmazzitelli @jshaughn Pinging all you asking for a review from somebody :-)
Still more work needed. There will be a follow-up PR.

business/openid_auth.go
```diff
@@ -1,19 +1,29 @@
...
 	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
+	"github.com/dgrijalva/jwt-go"
```
jmazzitelli, line 18: trivial - a newline should be on the next line to separate the third-party deps from the kiali deps.

business/openid_auth.go
```diff
@@ -28,8 +38,141 @@ type OpenIdMetadata struct {
...
+
+	// Parse/fetch received form data
+	err = r.ParseForm()
+	if err != nil {
+		err = fmt.Errorf("error parsing form info: %w", err)
+	} else {
+		// Read relevant form data parameters
+		params.Code = r.Form.Get("code")
+		params.IdToken = r.Form.Get("id_token")
+		params.State = r.Form.Get("state")
```
jmazzitelli, line 86: Are these all form parms necessary/required? If so, should we be checking to see if Code, IdToken, and/or State are `nil` and if so generate an error here? I am wondering if we should be performing some sanity checks here in case we hit some odd OpenID implementation that doesn't provide one or more of these.

business/openid_auth.go
```diff
@@ -28,8 +38,141 @@ type OpenIdMetadata struct {
...
+//	}
+//	if params.State == "" {
+//		return "State parameter is empty or invalid."
+//	}
+//	if params.IdToken == "" {
+//		return "Token is empty or invalid."
+//	}
+//
+//	return ""
+//}
```
jmazzitelli, line 115: We should remove this commented code if it will never be used.

business/openid_auth.go
jmazzitelli: HA!  I just noticed in code further below (`CheckOpenIdAuthenticationCodeFlowParams`) it does check Code and State. But what about IdToken?

business/openid_auth.go
```diff
@@ -142,3 +326,142 @@ func GetOpenIdMetadata() (*OpenIdMetadata, error) {
...
+func ValidateOpenIdState(openIdParams *OpenIdCallbackParams) (validationFailure string) {
+	state := openIdParams.State
+
+	separator := strings.LastIndexByte(state, '-')
+	if separator != -1 {
+		csrfToken, timestamp := state[:separator], state[separator+1:]
+		csrfHash := sha256.Sum224([]byte(fmt.Sprintf("%s+%s+%s", openIdParams.Nonce, timestamp, config.GetSigningKey())))
+
+		if fmt.Sprintf("%x", csrfHash) != csrfToken {
+			//RespondWithError(w, http.StatusForbidden, "Request rejected because of CSRF mitigation.")
```
jmazzitelli, line 437: should we delete this unused/commented code?

business/openid_auth.go
```diff
@@ -142,3 +326,142 @@ func GetOpenIdMetadata() (*OpenIdMetadata, error) {
...
+	// Using the namespaces API to check if token is valid. In Kubernetes, the version API seems to allow
+	// anonymous access, so it's not feasible to use the version API for token verification.
+	nsList, err := business.Namespace.GetNamespaces()
+	if err != nil {
+		return http.StatusUnauthorized, "Token is not valid or is expired", err
+	}
+
+	// If namespace list is empty, return unauthorized error
+	if len(nsList) == 0 {
+		return http.StatusUnauthorized, "Not enough privileges to login", nil
```
jmazzitelli, line 463: This has caused confusion with some users in the past. I think we should have a more explicit error message here. When we say "Not enough privileges to login" - what we specifically mean here is "User cannot get even a single namespace". Perhaps we should indicate this more explicitly (that they need to get able to get at least one namespace). 

The error message right now is ambiguous - the user has no idea what privileges he needs - he'll have to hunt that down in the docs. Granted, just getting the minimum privileges here ("get at least one namespace") won't be enough to really use Kiali, but maybe this error message should help the user a little more. Maybe tell them "Look at the Kiali documentation for the proper permissions needed to access Kiali, including the requirement that you must be able to get at least one namespace". I can't think of a good way to say it - but I know the way it is right now should change :)

handlers/authentication.go
```diff
@@ -696,14 +719,125 @@ func OpenIdRedirect(w http.ResponseWriter, r *http.Request) {
...
+	// "IanaClaims" type just for convenience to avoid creating new types and
+	// to bring some type convergence on types for the auth source code.
+	sessionData := business.BuildOpenIdJwtClaims(openIdParams)
+	sessionDataJson, err := json.Marshal(sessionData)
+	if err != nil {
+		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials", err.Error())
+		return true
+	}
+
+	// Cihper the session data
```
jshaughn, line 807: typo

business/openid_auth.go
```diff
@@ -28,8 +38,141 @@ type OpenIdMetadata struct {
...
+//		return "State parameter is empty or invalid."
+//	}
+//	if params.IdToken == "" {
+//		return "Token is empty or invalid."
+//	}
+//
+//	return ""
+//}
+
+func CheckOpenIdAuthenticationCodeFlowParams(params *OpenIdCallbackParams) string {
```
jshaughn, line 117: isn't it actually the OpenID "Authorization" code flow? As part of an authentication request?

jshaughn (commented):
I don't see anything wrong but I have no idea what I'm looking at.  I defer to those with more expertise.

handlers/authentication.go
```diff
@@ -696,14 +719,125 @@ func OpenIdRedirect(w http.ResponseWriter, r *http.Request) {
...
+	// developer console. So, we cipher the session data using AES-GCM
+	// which allows to leave aside the usage of JWT tokens. So, this
+	// builds a bare JSON serialized into a string, cipher it and
+	// set a cookie with the ciphered string. Yet, we use the
+	// "IanaClaims" type just for convenience to avoid creating new types and
+	// to bring some type convergence on types for the auth source code.
+	sessionData := business.BuildOpenIdJwtClaims(openIdParams)
+	sessionDataJson, err := json.Marshal(sessionData)
+	if err != nil {
+		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials", err.Error())
```
jmazzitelli, line 803: I see the same error message repeated here and 3 other places. To make it easier on someone debugging this, let's make each error message different so when someone says, `I see "Error when creating credentials" in the logs` we won't have to wonder which of the 4 places it happened in the code.

```suggestion
		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials - failed to marshal json", err.Error())
```

handlers/authentication.go
```diff
@@ -696,14 +719,125 @@ func OpenIdRedirect(w http.ResponseWriter, r *http.Request) {
...
+		return true
+	}
+
+	// Cihper the session data
+	block, err := aes.NewCipher([]byte(config.GetSigningKey()))
+	if err != nil {
+		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials", err.Error())
```
jmazzitelli, line 810:
```suggestion
		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials - failed to create cipher", err.Error())
```

handlers/authentication.go
jmazzitelli, line 807: trivial - typo
```suggestion
	// Cipher the session data
```

handlers/authentication.go
```diff
@@ -696,14 +719,125 @@ func OpenIdRedirect(w http.ResponseWriter, r *http.Request) {
...
+		return true
+	}
+
+	aesGcm, err := cipher.NewGCM(block)
+	if err != nil {
+		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials", err.Error())
```
jmazzitelli, line 816:
```suggestion
		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials - failed to create gcm", err.Error())
```

handlers/authentication.go
```diff
@@ -696,14 +719,125 @@ func OpenIdRedirect(w http.ResponseWriter, r *http.Request) {
...
+		return true
+	}
+
+	aesGcmNonce, err := util.CryptoRandomBytes(aesGcm.NonceSize())
+	if err != nil {
+		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials", err.Error())
```
jmazzitelli, line 822:
```suggestion
		RespondWithDetailedError(w, http.StatusInternalServerError, "Error when creating credentials - failed to generate random bytes", err.Error())
```

jmazzitelli (changes requested):

business/openid_auth.go
israel-hdez: Ops! Yes, I'll fix the naming.

business/openid_auth.go
israel-hdez: Unused right now, but it's here so that I don't forget it.
All these new functions are duplicating code inside the `performOpenIdAuthentication`. The code flow is more or less the same in both "openid flows", but order of execution changes. So, in a follow up PR, my idea is to change `performOpenIdAuthentication` to unify code and de-duplicate.  So, this commented code will be un-commented :)

business/openid_auth.go
israel-hdez: Hmm, yes, I'll delete.

handlers/authentication.go
jshaughn: +1000 unique messages are priceless.  If you want the same text add a numeric or some sort of identifier.

business/openid_auth.go
israel-hdez: IdToken is for the other flow. Not used in this PR, but it will be used in the follow-up PR.

handlers/authentication.go
israel-hdez: This is using `RespondWithDetailedError`. I was hoping that the "detailed" part would give context. But yes, it doesn't hurt if we add the context ourselves.

israel-hdez:
@jshaughn @jmazzitelli All your feedback solved.

handlers/authentication.go
jmazzitelli: The detailed part does give context - but sometimes it doesn't, and because it's runtime generated, we can't just grep for it to find where it came from.

When a user hits an error, what do they say? "I have an error in my logs that says "Error when creating credentials" with details "foo bar"". We then (at least I know I do) immediately do a grep for "Error when creating credentials" to find out where it came from in the code. In this case, we'll find 4 places and we don't know which one to immediately look at... "foo bar" may or may not help make it more obvious, but even then we aren't absolutely sure and we have to look at the 4 places to eliminate the possibilities. And of course we can't grep on "foo bar" because that isn't in the code.

For this reason, I never (or I try to never) have the same exact error message in multiple places. In fact, I've been known to do little tricks like add a period at the end of one sentence in one message, but leave it off the sentence in the other :) Anything to make it unique so someone debugging can quickly find the one spot where the log/error message came from by just doing a simple grep.

business/openid_auth.go
jmazzitelli: What about this one?

jmazzitelli:
> @jshaughn @jmazzitelli All your feedback solved.

What say you all about https://github.com/kiali/kiali/pull/3189#discussion_r487247559

business/openid_auth.go
israel-hdez: We could say "Not enough privileges to login. Check Kiali's RBAC documentation", hoping users will visit https://kiali.io/documentation/latest/configuration/rbac/.

jmazzitelli (approved):

Issue #3048, opened 2020-07-29
In-graph tracing: handle more nodes and graph types


jotak:
Epic: #2677 

Currently, the entry point to show traces is Service nodes.

First thing: it's actually wrong as pointed out there: https://github.com/kiali/kiali/issues/3045 . It just works accidentally when service name == app label. So the real entry point should be apps.

Second thing: we can expand this a little bit, by allowing any kind of entry point as long as there's an app label defined (it could work for apps, services and workloads nodes, although the last two need some post-fetch filtering)

jshaughn:
@jotak As mentioned in my comment in #3045 the "app" is typically set for Service, App and Workload nodes when using telemetry V2.  This is because canonical_service is always set in the telemetry, defaulting to workload name.  Of course "app" may not be set on certain [service] nodes that don't directly tie to a workload.

jotak:
Yes, and in jaeger it's similar to telemetry: when "app" is not set, then the workload name is used instead as the jaeger service identifier. So if we want to go further, when app label isn't defined, we could also get traces from workload names.

jshaughn:
I like service nodes because they can appear in any sort of graph type.  But feasibly any node with app set, which would include workload nodes as well.

PR #3190, opened 2020-09-09
Increase limit while fetching traces for workloads and services

jotak:
... because they are fetched for apps, and narrowed-down later. Re-apply initial limit after post-filtering.

This should attenuate the effect of having an inconsistent number of traces displayed for workloads and services (though not entirely eliminating that possibility)

Plus properly recognize "trace not found" error as a 404 instead of 503 / unmarshalling issue

Part of https://github.com/kiali/kiali/issues/3048

### DIFF
--- a/business/jaeger.go
+++ b/business/jaeger.go
@@ -55,7 +55,7 @@ func (in *JaegerService) GetServiceTraces(ns, service string, query models.Traci
 	reqLimit := query.Limit
 	query.Limit *= 2
 	r, err := client.GetServiceTraces(ns, app, service, query)
-	if reqLimit > 0 && r != nil {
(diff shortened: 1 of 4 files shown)


jotak:
@mattmahoneyrh this is for what we discussed while you tested https://github.com/kiali/kiali-ui/pull/1914

As explained, the issue that you saw (number of traces displayed looked volatile) is due to post-filtering traces for workload after they were queried with a limit.

This fix is not a silver bullet, but should mitigate the problem.

jshaughn (approved):
Sometimes you just need a hack.

mattmahoneyrh (approved):

Issue #3187, opened 2020-09-08
Add new config `rev` into kiali settings


xeviknal:
Right now, Kiali assumes that the istio config map will be named as 'istio'.

That is not always the case:
- Multiple kiali installations in the same namespace: https://github.com/kiali/kiali/issues/2883, https://github.com/kiali/kiali/issues/2884
- Maistra installation: it uses the istio revistion and their configmaps always include the revision name.

The goal for this issue is then to add a new config field to take in the revision of the control plane installation.

This issue won't cover the scenario where multiple control planes are installed in the same ns however it will ease the task to achieve it.

Relates to https://github.com/kiali/kiali/issues/2884

PR #3193, opened 2020-09-09
Add revision field into config + use it to fetch istio configmap

xeviknal:
Fixes https://github.com/kiali/kiali/issues/3187
Requires: https://github.com/kiali/kiali-operator/pull/126

When istio is installed using the canary upgrade feature [1], Kiali have errors in service details page. This is the following error:

lclKJB6

In order to indicate the installation revision, there is now available the following setting:

```yaml
external_services:
  istio:
    revision: canary
```
In case istio is installed without revision, then this revision setting should be `""` (or left it empty).

[1] https://istio.io/latest/docs/setup/upgrade/

### DIFF
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -556,9 +556,9 @@ func (in *IstioValidationsService) fetchNonLocalmTLSConfigs(mtlsDetails *kuberne
 		var istioConfig *core_v1.ConfigMap
 		var err error
 		if IsNamespaceCached(cfg.IstioNamespace) {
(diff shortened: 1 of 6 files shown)


jmazzitelli (dismissed):
code looks good.

hhovsepy (approved):
Verified, regression tested on istio upstream.

lucasponce (changes requested):
Would you mind to rebase with master ?

This PR has some problems with UI master due lack of updates there, if you don't mind, I can test the whole cycle after rebased.

lucasponce (approved):
Regression test passed.
After rebase, PR is synced with UI.

Issue #3175, opened 2020-09-04
[Tracing+Metrics] Filter spans for service/workloads


jotak:
In service/workload metrics page with spans correlation, spans should be filtered for service/workloads as we're doing for tracing page.

lucasponce:
Added to the project, it's a subtask of https://github.com/kiali/kiali/issues/2776

PR #3200, opened 2020-09-10
Filter spans for workloads and services

jotak:
Involves some refactoring, as more and more business logic were entering into jaeger client, which obviously shouldn't => move business logic to business layer.
Client is now back to a (quite) dummy interface to Jaeger API, which essentially allows to fetch traces out of an app.

Also adding more tests

Fixes https://github.com/kiali/kiali/issues/3175

Frontend PR: https://github.com/kiali/kiali-ui/pull/1919


hhovsepy (approved):
Verified.

jotak:
@aljesusg does code look OK to you? (@hhovsepy has already verified testing)

aljesusg (approved):
LGFM

PR #3206, opened 2020-09-14
Add ConfigMapName setting to kiali CR

xeviknal:
Rename revision by config map name setting.
Reduce kiali code and remove the hardcoded istio config-map also.

related to https://github.com/kiali/kiali-operator/pull/127


jmazzitelli (approved):
I did not run this, but the code looks good. Now allows the user to specify whatever the configmap name is rather than kiali relying on a pattern that may or may not change in the future.

hhovsepy (approved):
Verified on maistra and istio.
'istio:
  config_map_name: istio'

lucasponce (approved):

Issue #2924, opened 2020-06-22
Wizard to create AuthorizationPolicy based on the graph of the current traffic


olaf-meyer:
**Is your feature request related to a problem? Please describe.**
From my experience, AuthorizationPolicies are usually created to match the traffic at a certain time between services and deny all other traffic. 

**Describe the solution you'd like**
Would it be possible to create a wizard for AuthorizationPolicy that is based on the graph in Kiali? By this, the manual mapping of the graph can be skipped and a user only needs to adjust the resources instead of creating them from scratch.

**Describe alternatives you've considered**
Creating a way to export the graph in a text form as YAML or JSON, so that it can be transformed into other resources like AuthorizationPolicy

**Additional context**

lucasponce:
Thanks @olaf-meyer.
We have had some discussions in a very similar topic and it makes all of sense.

Also, we are looking for feedback like this for more high level additions on the Wizards area.

jshaughn:
@olaf-meyer @lucasponce This is a good idea, it sounds similar to https://github.com/kiali/kiali/issues/1542.

olaf-meyer:
Please let me know, if you need a tester.

lucasponce:
@olaf-meyer count on it :-) thanks 

We are actively working on Wizards area on these sprints, so, I will update the issue soon.

jshaughn:
@lucasponce Assigning you and adding to backlog just to keep on the radar.

lucasponce:
I'm actively working on this one, but I guess we need to propagate a couple of fields coming from the Istio prometheus.
Basically from the graph, Kiali should have access to the "source_principal" and "destination_principal" fields.

![image](https://user-images.githubusercontent.com/1662329/93124901-a036e400-f6ca-11ea-9992-f2589cfe1ad0.png)
With those fields, from the telemetry we could generate AuthorizationPolicy resources per namespace "lock-in" the current traffic.

I guess most straightforward way is to add those fields in the graph (even if those are not used), other alternatives, like using Sidecars or trying to lock by IPs are costly and less standard.

So, goal would be to have those fields mapped into the "GraphDefinition" object used to fetch telemetry in the UI.

cc @jshaughn

jshaughn:
[...]

PR #3208, opened 2020-09-14
Add principals into security appender

lucasponce:
@jshaughn: I've started to add the discussed changes for https://github.com/kiali/kiali/issues/2924 into the security appender.

I think these changes make sense but perhaps I'm less familiar with the structure.

Feel free to comment.

Note: it's not yet finished, tests and other checks are missing.

### DIFF
--- a/graph/config/cytoscape/cytoscape.go
+++ b/graph/config/cytoscape/cytoscape.go
@@ -88,9 +88,11 @@ type EdgeData struct {
 	Target string `json:"target"` // child node ID
 
 	// App Fields (not required by Cytoscape)
-	Traffic      ProtocolTraffic `json:"traffic,omitempty"`      // traffic rates for the edge protocol
-	ResponseTime string          `json:"responseTime,omitempty"` // in millis
-	IsMTLS       string          `json:"isMTLS,omitempty"`       // set to the percentage of traffic using a mutual TLS connection
+	Traffic      	ProtocolTraffic `json:"traffic,omitempty"`      // traffic rates for the edge protocol
+	ResponseTime 	string          `json:"responseTime,omitempty"` // in millis
+	IsMTLS       	string          `json:"isMTLS,omitempty"`       // set to the percentage of traffic using a mutual TLS connection
+	DestPrincipal 	string		 	`json:"destPrincipal,omitempty"`
+	SourcePrincipal string       	`json:"sourcePrincipal,omitempty"`
 }
 
 type NodeWrapper struct {
@@ -281,6 +283,12 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
 				Traffic: ProtocolTraffic{
 					Protocol: protocol,
 				},
+			}
+			if e.Metadata[graph.SourcePrincipal] != nil {
+				ed.SourcePrincipal = e.Metadata[graph.SourcePrincipal].(string)
+			}
+			if e.Metadata[graph.DestPrincipal] != nil {
+				ed.DestPrincipal = e.Metadata[graph.DestPrincipal].(string)
 			}
 			addEdgeTelemetry(e, &ed)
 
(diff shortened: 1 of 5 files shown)


graph/config/cytoscape/cytoscape.go
```diff
@@ -88,9 +88,11 @@ type EdgeData struct {
...
 
 	// App Fields (not required by Cytoscape)
-	Traffic      ProtocolTraffic `json:"traffic,omitempty"`      // traffic rates for the edge protocol
-	ResponseTime string          `json:"responseTime,omitempty"` // in millis
-	IsMTLS       string          `json:"isMTLS,omitempty"`       // set to the percentage of traffic using a mutual TLS connection
+	Traffic      	ProtocolTraffic `json:"traffic,omitempty"`      // traffic rates for the edge protocol
+	ResponseTime 	string          `json:"responseTime,omitempty"` // in millis
+	IsMTLS       	string          `json:"isMTLS,omitempty"`       // set to the percentage of traffic using a mutual TLS connection
+	DestPrincipal 	string		 	`json:"destPrincipal,omitempty"`
+	SourcePrincipal string       	`json:"sourcePrincipal,omitempty"`
```
jshaughn, line 95: Just needs format a short comments for each new field.  For bonus points alphabetize below // App Fields.

graph/config/cytoscape/cytoscape.go
```diff
@@ -282,6 +284,12 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
 					Protocol: protocol,
 				},
 			}
+			if e.Metadata[graph.SourcePrincipal] != nil {
+				ed.SourcePrincipal = e.Metadata[graph.SourcePrincipal].(string)
+			}
+			if e.Metadata[graph.DestPrincipal] != nil {
+				ed.DestPrincipal = e.Metadata[graph.DestPrincipal].(string)
+			}
```
jshaughn, line 292: Alphabetize (just flip them)

graph/meta.go
```diff
@@ -30,6 +30,8 @@ const (
 	IsUnused        MetadataKey = "isUnused"
 	ProtocolKey     MetadataKey = "protocol"
 	ResponseTime    MetadataKey = "responseTime"
+	SourcePrincipal MetadataKey = "sourcePrincipal"
+	DestPrincipal   MetadataKey = "destPrincipal"
```
jshaughn, line 34: Alphabetize

graph/telemetry/istio/appender/security_policy.go
```diff
@@ -90,6 +90,12 @@ func (a SecurityPolicyAppender) appendGraph(trafficMap graph.TrafficMap, namespa
 	a.populateSecurityPolicyMap(securityPolicyMap, &inVector)
 
 	applySecurityPolicy(trafficMap, securityPolicyMap)
+
+	principalMap := make(map[string]map[graph.MetadataKey]string)
+	a.populatePrincipalMap(principalMap, &outVector)
+	a.populatePrincipalMap(principalMap, &inVector)
```
jshaughn, line 96: Not sure you really need to do all of this separately, I'd suggest maybe just working this into the existing 'populate' and 'apply' and not have to walk through everything twice.  If you are worried about getting the feature in then this is fine for now and I'll probably refactor it later.

jshaughn (changes requested):
I think this is the right way to do it, piggybacking on the security appender is nice.  I think it's perfectly reasonable to require the 'Security' Display option.  Note that opetions can be automatically enabled if and when that makes sense.  For example, if you type 'has mtls' into graph find you'll see that the security option gets enabled automagically.

A few minor comments about alphabetizing, a more major comment about removing the redundancy, but the main blocker is that you need to add to the test code.  If you need help let me know, you could progress with the main feature and when I have time I can offer a sub-pr.

lucasponce:
Thanks @jshaughn, let me know how it looks like now.

jshaughn (approved):
Seems fine, maybe we can optimize a little more but certainly looks good enough to progress.

Issue #3201, opened 2020-09-10
[Operator/UI] istio_namespace doesn't work with istio_component_namespaces


PatrickDuncan:
**General remarks**

https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L47-L65

**Describe the bug**

`istio_namespace` is ignored when `istio_component_namespaces` is present. 

![Screen Shot 2020-09-10 at 2 54 33 PM](https://user-images.githubusercontent.com/6889074/92787100-caa93a00-f376-11ea-887e-fb4a44432ae5.png)

**Versions used**
Kiali: 1.23.0
Istio: 1.7.0
Kubernetes flavour and version: 1.17.5

**To Reproduce**
```
istiod -> istio-system
istio-ingressgateway -> istio-system
kiali -> istio-observe
prometheus -> istio-observe
```

```yaml
istio_component_namespaces:
  prometheus: istio-observe
istio_namespace: istio-system
```

**Expected behavior**

The namespace for istiod and istio-ingressgateway should be inherited from `istio_namespace` since it's not defined in `istio_component_namespaces`. It shouldn't use Kiali's deploy namespace.

Minor bug, the workaround is just to define the istiod and istio-ingressgateway namespaces separately

jmazzitelli:
FWIW: I think we want to deprecate the istio_component_namespace stuff - that was needed for a-now-obsolete Istio installation mechanism where things like pilot, citadel, etc would be in different places. Now there is one monolith (istiod) there is no really need for this and in fact creates confusion.

IIRC, the thinking was if you had anything in istio_component_namespace, that would be assumed to have everything. istio_namespace wasn't a fallback. I can't remember exactly.

But, be that as it may, there is no real need for istio_component_namespace anymore. and we will most likely deprecate it/remove it at some point soon.

jmazzitelli:
Here's one of the PRs where we were going to do this - but pulled back. We'll probably do this in the future soon, though. https://github.com/kiali/kiali-operator/pull/54

PatrickDuncan:
@jmazzitelli Thanks for the reply. If you guys deprecate `istio_component_namespace` how would you go about something like this?
```
istiod -> istio-system
istio-ingressgateway -> istio-system
kiali -> istio-observe
prometheus -> istio-observe
```

I think it's still a reasonable use case to have istio core and istio addon resources in separate namespaces.

jmazzitelli:
[...]

PR #3211, opened 2020-09-15
Allow users to specify namespace for each Istio health component

xeviknal:
Fixes https://github.com/kiali/kiali/issues/3201
Requires https://github.com/kiali/kiali-operator/pull/128

Due to the removal of `istio_component_namespaces` from the settings, the istio component health feature needs other ways to let users specify where are their components located in.

This PR adds a new namespace field in each `component_status` object. In case, the namespace is empty, it defaults to the `istioNamespace` setting (typically `istio-system`). This way, users might specify the components of tracing, grafana, prometheus and the istio components (ingress, egress and istiod).

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -65,13 +65,7 @@ func (iss *IstioStatusService) getComponentNamespacesWorkloads() ([]apps_v1.Depl
 	nss := map[string]bool{}
 	deps := make([]apps_v1.Deployment, 0)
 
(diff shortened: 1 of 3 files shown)


lucasponce (approved):
Code LGTM, not tested.

hhovsepy:
Verified, the default empty '', correct namespace 'istio-system', and wrong namespaces cases.

PR #3214, opened 2020-09-15
Add Support for requests with no response

### LINKED ISSUE #3194: Add support for requests with no response
_This is a borderline bug but we can consider it a high priority enhancement as it's not a use-case we previously knew existed or tried to handle._

In the case of http/grpc downstream connection loss (i.e. the client abandons the request, usually based on client-side timeout) the request actually has *no response*.  That means there is no response code set and it defaults to 0.  In general there will be a response_code set as DC, although there are possibly other scenarios.

For grpc requests if the  [http] response_code=0 the grpc_response_status should be ignored.

For Kiali we need to handle this situation and visusalize it as needed.  We do not necessarily have to report the response code as a 0, that is really just the "indicator" used by istio to report the lack of response.  The main issue is for use to ensure that we don't miss these requests in our queries, or discard/skip these requests when processing query results.

jshaughn:
Fixes #3194 

Requests that have no response must also be treated as errors because the request never completes, and a response is never sent.  Error handling up to this point assumed a valid response code for the request protocol. But when there is no response Istio sets the response_code=0 to indicate the inability to set a proper code.  Kiali needs to treat the "no response" case as an error.
- For the graph generation and appenders add support for the new case.
- For health enhance the defaults with a new error rate config and update rate eggregation
- For metrics capture the no response case in the errors queries

Also:
- fix issue by making service nodes with unknown namespace inaccessible.

**Requires UI PR: https://github.com/kiali/kiali-ui/pull/1925**

For Testing:
One way to force the 0 response_code scenario is to deploy bookinfo and then apply our hack:

`> oc apply -n bookinfo -f hack/istio/bookinfo-ratings-delay.yaml`

It should result in some various failure scenarios giving you red edges and red nodes.  The edges should likely show code="-", flags="DC" in the "flags" tab of the side panel.  The nodes should reflect traffic rate issues in the side panel health tooltip.

@jmazzitelli please review the graph code changes
@jotak  please review the metrics code changes
@aljesusg please review the health code changes

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -515,7 +515,7 @@ func (conf *Config) AddHealthDefault() {
 						Code:      "^-$", // no response is indicated with a "-" code
 						Protocol:  "http|grpc",
 						Direction: ".*",
-						Degraded:  0.1,
+						Degraded:  0,
 						Failure:   10,
 					},
 				},
--- a/config/config.go
+++ b/config/config.go
@@ -515,7 +515,6 @@ func (conf *Config) AddHealthDefault() {
 						Code:      "^-$", // no response is indicated with a "-" code
 						Protocol:  "http|grpc",
 						Direction: ".*",
-						Degraded:  0,
 						Failure:   10,
 					},
 				},
--- a/config/config.go
+++ b/config/config.go
@@ -482,7 +482,7 @@ func NewConfig() (c *Config) {
 	return
 }
 
-// Add Health Default Configuration
+// AddHealthDefault Configuration
 func (conf *Config) AddHealthDefault() {
 	// Health default configuration
 	healthConfig := HealthConfig{
@@ -509,6 +509,13 @@ func (conf *Config) AddHealthDefault() {
 						Code:      "^[1-9]$|^1[0-6]$",
 						Protocol:  "grpc",
 						Direction: ".*",
+						Failure:   10,
+					},
+					{
+						Code:      "^-$", // no response is indicated with a "-" code
+						Protocol:  "http|grpc",
+						Direction: ".*",
+						Degraded:  0.1,
 						Failure:   10,
 					},
 				},
--- a/prometheus/metrics.go
+++ b/prometheus/metrics.go
@@ -14,6 +14,11 @@ import (
 	"github.com/kiali/kiali/status"
 )
 
+const (
+	regexGrpcResponseStatusErr = "^[1-9]$|^1[0-6]$"
+	regexResponseCodeErr       = "^0$|^[4-5]\\d\\d$"
+)
+
 func getMetrics(api prom_v1.API, q *IstioMetricsQuery) Metrics {
 	labels, labelsError := buildLabelStrings(q)
 	grouping := strings.Join(q.ByLabels, ",")
@@ -61,7 +66,7 @@ func buildLabelStrings(q *IstioMetricsQuery) (string, []string) {
 
 	// both http and grpc requests can suffer from no response (response_code=0) or an http error
 	// (response_code=4xx,5xx), and so we always perform a query against response_code:
-	httpLabels := append(labels, `response_code=~"^0$|^[4-5]\\d\\d$"`)
+	httpLabels := append(labels, fmt.Sprintf(`response_code=~"%s"`, regexResponseCodeErr))
 	errors = append(errors, "{"+strings.Join(httpLabels, ",")+"}")
 
 	// if necessary also look for grpc errors. note that the grpc test intentionally avoids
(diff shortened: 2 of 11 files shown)


jmazzitelli (dismissed):
quick review of the code and nothing stands out. I do question how the config/config.go settings are becoming the default when using the operator and a Kiali CR. I assume an empty list defined in the CR will end up setting these defaults on the server.

prometheus/metrics.go
```diff
@@ -58,17 +58,21 @@ func buildLabelStrings(q *IstioMetricsQuery) (string, []string) {
...
+	httpLabels := append(labels, `response_code=~"^0$|^[4-5]\\d\\d$"`)
+	errors = append(errors, "{"+strings.Join(httpLabels, ",")+"}")
+
+	// if necessary also look for grpc errors. note that the grpc test intentionally avoids
+	// `grpc_response_status!="0"`. We need to be backward compatible and handle the case where
+	// grpc_response_status does not exist, or if it is simply unset. In Prometheus, negative tests on a
+	// non-existent label match everything, but positive tests match nothing. So, we stay positive.
+	// furthermore, make sure we only count grpc errors with successful http status.
+	if protocol != "http" {
+		grpcLabels := append(labels, `grpc_response_status=~"^[1-9]$|^1[0-6]$",response_code!~"^0$|^[4-5]\\d\\d$"`)
```
jotak, line 73: since that exact same regex is used twice, here and above, perhaps put it as a const to make sure one isn't modified without the other

jotak (approved):
LGTM for the metrics API part. Thanks for adding code doc.

Not for this PR, but sometimes I'm wondering if we shouldn't remove all that specific errors-querying parts from the backend and leave that to the frontend. Sure it would only move complexity, not remove it, but it would also run 1 prom query instead of 2 ; and I feel that scanning errors from responses is easier than fine-tuning a query for that. wdyt?

lucasponce:
Have the reviewers test the whole feature ?

This feature has a lot of potential side effects and it starts to worry me if approvals are only in the code.

config/config.go
```diff
@@ -510,6 +510,13 @@ func (conf *Config) AddHealthDefault() {
 						Direction: ".*",
 						Failure:   10,
 					},
+					{
+						Code:      "^-$", // no response is indicated with a "-" code
+						Protocol:  "http|grpc",
+						Direction: ".*",
+						Degraded:  0,
+						Failure:   10,
```
aljesusg, line 518: it's ok but we don't need to set degraded to 0 (it's 0 by default) we can remove that line if you want

aljesusg (dismissed):
LGFM

aljesusg:
> Have the reviewers test the whole feature ?
> 
> This feature has a lot of potential side effects and it starts to worry me if approvals are only in the code.

I tested it and it's working at least the error-rates demo

lucasponce:
> I tested it and it's working at least the error-rates demo

I've asked @jshaughn to update the error-rates demo to also cover this scenario as well.

I haven't tested it, but that one can be checked after merge if you like, my comment was to avoid that no regressions were introduced.

Thanks for testing.

aljesusg:
> > I tested it and it's working at least the error-rates demo
> 
> I've asked @jshaughn to update the error-rates demo to also cover this scenario as well.
> 
> I haven't tested it, but that one can be checked after merge if you like, my comment was to avoid that no regressions were introduced.
> 
> Thanks for testing.

Totally , I'll test it . I don't want any regressions on this please :)

jshaughn:
>  ...sometimes I'm wondering if we shouldn't remove all that specific errors-querying parts from the backend and leave that to the frontend.

@jotak I understand the motivation but in the end I think I prefer it in the back-end such that multiple clients would not have to duplicate the same work, and because I'd rather let prom filter data rather than pull it, send it, and then filter it.  BUt maybe more than anything I don't want to break what is already working :-)

jotak (dismissed):

jmazzitelli (approved):

lucasponce:
@aljesusg we need to merge this today, if you can check a non-regression test we can consolidate.

@jshaughn has changes on demos/error-rates app that I'm going to review, but if it looks good I don't mind to consolidate first.

aljesusg (approved):
LGFM

aljesusg:
> @aljesusg we need to merge this today, if you can check a non-regression test we can consolidate.
> 
> @jshaughn has changes on demos/error-rates app that I'm going to review, but if it looks good I don't mind to consolidate first.

Tested with https://github.com/kiali/demos/pull/10 I was checking it :)

PR #3215, opened 2020-09-15
Support for acting as authenticated OIDC client (authorization code flow)

israel-hdez:
When OIDC client should authenticate, we need the client-secret provided/configured by/in the OIDC server. For this, we re-use our old/reserved "kiali" secret that was used in the past for the already removed login authentication strategy avoiding another set of changes in the operator.

This also:
* Fixes logout for the auth code flow
* Fixes broken OIDC "implicit flow" (nonce cookie deleted early)
* Deduplicates code doing a small rework of the `performOpenIdAuthentication` function in the authenticatin.go file

Related #3088

### DIFF
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -63,7 +63,20 @@ func BuildOpenIdJwtClaims(openIdParams *OpenIdCallbackParams) *config.IanaClaims
 	}
 }
 
-func ExtractOpenIdCallbackParams(w http.ResponseWriter, r *http.Request) (params *OpenIdCallbackParams, err error) {
+func CallbackCleanup(w http.ResponseWriter) {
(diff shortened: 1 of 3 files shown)


jmazzitelli:
> For this, we re-use our old/reserved "kiali" secret that was used in the past

I just knew we would end up using that again!!! :-D I didn't delete that because I just had a feeling we would use it again, I just didn't think it would be so soon :-)

israel-hdez:
@cfcosta If possible, please try and review :)

cfcosta (approved):
Did a quick smoke test (with the normal authorization workflow, did not do anything different from the default  hack setup) and everything seems ok. Code also seem ok, good we are bringing back commented code ;)

israel-hdez:
Thanks @cfcosta 
Merging...

PR #3220, opened 2020-09-17
Bump k-charted 0.6.2

jotak:
Note, there's actually no code change here, it's only on the frontend, but this PR is to keep version consistent with kiali-ui


lucasponce (approved):

aljesusg (approved):
LGFM

Issue #3224, opened 2020-09-17
Kiali is not working with non-root user


a8j8i8t8:
**Describe the bug**
Kiali is not working with non-root user.

**Versions used**
Kiali: 1.23
Istio: 1.7.0
Kubernetes flavour and version: Kubernetes 1.18.0

**To Reproduce**
Started Kiali as non-root user using below: -
```
        securityContext:
          runAsGroup: 1000
          runAsNonRoot: true
          runAsUser: 1000
```
Following are the start-up logs: - 
```
I0917 13:21:35.281821       1 kiali.go:64] Kiali: Version: v1.23.0, Commit: 61d506bb658265552ab835c5e4183ccf8d86dcec
I0917 13:21:35.283900       1 kiali.go:155] Using authentication strategy [anonymous]
W0917 13:21:35.283914       1 kiali.go:157] Kiali auth strategy is configured for anonymous access - users will not be authenticated.
I0917 13:21:35.283955       1 kiali.go:85] Kiali: Console version: 1.23.0
I0917 13:21:35.283965       1 kiali.go:230] Updating base URL in index.html with [/kiali]
E0917 13:21:35.284008       1 kiali.go:256] File I/O error [open /opt/kiali/console/index.html: read-only file system]
I0917 13:21:35.284021       1 kiali.go:211] Generating env.js from config
E0917 13:21:35.284040       1 kiali.go:256] File I/O error [open /opt/kiali/console/env.js: read-only file system]
I0917 13:21:35.286095       1 server.go:67] Server endpoint will start at [:20001/kiali]
I0917 13:21:35.286109       1 server.go:68] Server endpoint will serve static content from [/opt/kiali/console]
I0917 13:21:35.286118       1 metrics_server.go:18] Starting Metrics Server on [:9090]
```
**Expected behavior**
We should not see any File I/O error.

jmazzitelli:
Kiali is already running as a non-root user as defined in the image : https://github.com/kiali/kiali/commit/a6affb4f74664f866cf6b27c532a9435ff656d57

I'm not sure if what you are doing is breaking that ?

a8j8i8t8:
@jmazzitelli Not doing anything special. Any pointers would help. 🤔

jmazzitelli:
> @jmazzitelli Not doing anything special. Any pointers would help.

[...]

PR #3228, opened 2020-09-18
3224 - define USER as numeric uid in dockerfile

jmazzitelli:
** Describe the change **

If someone wants to use a PodSecurityPolicies with runAsNonRootUser (even though Kiali already does this), Kiali Dockerfile needs to be set with a numeric USER. See https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups where it says, `MustRunAsNonRoot - Requires that the pod be submitted with a non-zero runAsUser or have the USER directive defined (using a numeric UID) in the image`.

** Issue reference **

https://github.com/kiali/kiali/issues/3224

** Backwards incompatible? **

Yes. This PR simply uses the same UID as what was used before (1000) except now the Dockerfile is explicit about it (rather than use the uesr name "kiali").

### DIFF
--- a/deploy/docker/Dockerfile-ubi7-minimal
+++ b/deploy/docker/Dockerfile-ubi7-minimal
@@ -10,7 +10,7 @@ WORKDIR $KIALI_HOME
 RUN microdnf install -y shadow-utils && \
     microdnf clean all && \
     rm -rf /var/cache/yum && \
-    adduser kiali
+    adduser --uid 1000 kiali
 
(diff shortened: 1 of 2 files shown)


jmazzitelli:
I just tested this for backward compatibility and it all works as before. Nothing breaks.

jshaughn (approved):
This seems fine to me, but let's wait for Edgar to confirm.  But if he can't review before the release then just merge it.

Issue #3233, opened 2020-09-22
Error fetching spans on idle workloads


lucasponce:
Running an experiment I scale to 0 a workload and I got an internal error while getting details in Kiali:

![image](https://user-images.githubusercontent.com/1662329/93878553-8e60cc80-fcda-11ea-9244-c7affe0b2fa5.png)

aljesusg:
I can't replicate this

![Screenshot from 2020-09-22 14-33-41](https://user-images.githubusercontent.com/3019213/93882604-ba7f4c00-fce0-11ea-8df5-30637a8d8822.png)

In fact the call to spans when there is not pods return []

jaeger api when we fetch a traces of a service that not exist this return

```json
{"data":[],"total":0,"limit":0,"offset":0,"errors":null}
```

so it seems that this error is related with another thing @lucasponce, maybe jaeger crashed because it says internal server error, I'm going to make another tests to be sure 

cc @jotak

jotak:
@lucasponce perhaps you have more information in kiali server logs?

lucasponce:
I get this details from the browser, I'm running from localhost.
![image](https://user-images.githubusercontent.com/1662329/93886573-e6510080-fce5-11ea-8539-669853e16213.png)

I'm not using the normal bookinfo, but a different example:

```
kubectl apply -f https://iter8.tools/tutorials/namespace.yaml
kubectl --namespace bookinfo-iter8 apply -f https://iter8.tools/tutorials/bookinfo-tutorial.yaml
kubectl --namespace bookinfo-iter8 apply -f https://iter8.tools/tutorials/reviews-v3.yaml
```

And try to set replica to 0

![image](https://user-images.githubusercontent.com/1662329/93886872-3b8d1200-fce6-11ea-871d-4c8ff3c8753c.png)


Jaeger traces tab shows no errors, but this info:

![image](https://user-images.githubusercontent.com/1662329/93886949-5790b380-fce6-11ea-8df1-912ee8283a11.png)


But the Metrics tabs fails with a similar like this:

![image](https://user-images.githubusercontent.com/1662329/93887025-70996480-fce6-11ea-875e-04499094586f.png)

Kiali logs are clean, it may be something else, I don't know I haven't started to debug, just to describe the problem.

lucasponce:
```
GET http://admin:admin@localhost:8000/api/namespaces/bookinfo-iter8/workloads/reviews-v2/spans?startMicros=1600780418154000
```

```
[...]

PR #3236, opened 2020-09-22
Do not refine spans query when no spans were found

jotak:
It means that the workload which we're looking at, probably doesn't produce any more spans
(Also it was crashing with out-of-bound error)

Fixes https://github.com/kiali/kiali/issues/3233


aljesusg (approved):
LGFM

lucasponce (approved):
LGTM Tested locally.
@jotak would you mind to prepare a PR from the same branch to the v1.24 branch ?

Thanks

PR #3238, opened 2020-09-23
[1.24] Do not refine spans query when no spans were found

jotak:
It means that the workload which we're looking at, probably doesn't produce any more spans
(Also it was crashing with out-of-bound error)

Fixes https://github.com/kiali/kiali/issues/3233


lucasponce (approved):
LGTM

Issue #3248, opened 2020-09-24
panic in server , openid code


jmazzitelli:
I'm sending some requests I do not expect to give me access via curl. Even though I know I should not have access because I did not log in via openID, we should not panic the server. We should fix this by performing necessary checks (I assume there is a nil here somewhere that we need to check):

```
2020/09/24 11:44:29 http: panic serving 127.0.0.1:58904: runtime error: invalid memory address or nil pointer dereference
goroutine 385 [running]:
net/http.(*conn).serve.func1(0xc00033afa0)
	/home/jmazzite/bin/go-install/go/src/net/http/server.go:1800 +0x139
panic(0x12f2700, 0x200fb10)
	/home/jmazzite/bin/go-install/go/src/runtime/panic.go:975 +0x3e3
github.com/kiali/kiali/handlers.performOpenIdAuthentication(0x16cd1e0, 0xc000180a80, 0xc00025fd00, 0xc000158380)
	/home/jmazzite/source/kiali/kiali/src/github.com/kiali/kiali/handlers/authentication.go:160 +0x913
github.com/kiali/kiali/handlers.Authenticate(0x16cd1e0, 0xc000180a80, 0xc00025fd00)
	/home/jmazzite/source/kiali/kiali/src/github.com/kiali/kiali/handlers/authentication.go:488 +0x203
net/http.HandlerFunc.ServeHTTP(0x1526800, 0x16cd1e0, 0xc000180a80, 0xc00025fd00)
	/home/jmazzite/bin/go-install/go/src/net/http/server.go:2041 +0x44
github.com/kiali/kiali/handlers.AuthenticationHandler.HandleUnauthenticated.func1(0x16cd1e0, 0xc000180a80, 0xc00025fc00)
	/home/jmazzite/source/kiali/kiali/src/github.com/kiali/kiali/handlers/authentication.go:478 +0x166
net/http.HandlerFunc.ServeHTTP(0xc00036d680, 0x16cd1e0, 0xc000180a80, 0xc00025fc00)
	/home/jmazzite/bin/go-install/go/src/net/http/server.go:2041 +0x44
github.com/kiali/kiali/routing.metricHandler.func1(0x16cd1e0, 0xc000180a80, 0xc00025fc00)
	/home/jmazzite/source/kiali/kiali/src/github.com/kiali/kiali/routing/router.go:89 +0xa3
[...]

PR #3249, opened 2020-09-24
fix panic in openid check

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3248
backport PR: https://github.com/kiali/kiali/pull/3250


jmazzitelli:
I confirmed this fixes the panic. We now get a error 400 when doing a `.../kiali/api/authenticate` without any credentials. Which is how the openid implementation behaved in the past before the bug was introduced.

israel-hdez (approved):
Ouch, yes! This definitely is a bug.
An ugly one.

PR #3276, opened 2020-09-30
update make targets to work better with OpenShift image registry

jmazzitelli:
Never saw the need to do this until just now - need to make sure image registry is Managed. This also removes some hardcoded 'oc' commands that accidentally came in during past commits


jotak (approved):
LGTM !

PR #3282, opened 2020-10-02
Support parsed log messages for the workload logs

### LINKED ISSUE #3264: Add log entry support to the server-side logging API
The k8s logging API is fairly limited.  It does not support many query restraints and returns logs as a single string.   To better support querying logs it willbe useful to perform server-side pre-processing into log entries.  Parsing the logs string into log entries would initially give us a set of entries something like:

```
LogEntry
  Timestamp
  Severity
  Message
```

This would allow us to support new server-side filters like endTime and minSeverity.  Also to support return formats like String or LogEntries.

cfcosta:
Closes #3264.

This PR adds some support to parsing log messages from the workload, so we can apply some better filtering (as an example, kubernetes pod logs api supports a start time, but not an end time).

Some implementation details:

- Timestamp support uses ISO8601 instead of RFC3339 (which is go default), because no other stack seems to be using it from what I've seen.
- I could not find a good ISO8601 (or RFC3339 for that matter) regexp that could cover all our cases, because go does not support Perl-like lookarounds, like many mature regexp engines around.

For log severity, I'm currently supporting only named ones (INFO, WARN, ERROR). Log formats are really, really unreliable.

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -209,7 +209,7 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
 
 		timestamp := timestampRegexp.FindString(line)
 		if timestamp != "" {
-			parsed, _ := time.Parse("2006-01-02T15:04:05+0000", string(timestamp))
+			parsed, err := time.Parse("2006-01-02T15:04:05+0000", string(timestamp))
 
 			if err == nil {
 				message.Timestamp = timestamp
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -39,12 +39,16 @@ type PodLog struct {
 
 type LogEntry struct {
 	Message   string `json:"message,omitempty"`
-	Timestamp int64  `json:"timestamp,omitempty"`
 	Severity  string `json:"severity,omitempty"`
+	Timestamp int64  `json:"timestamp,omitempty"`
 }
 
 var (
 	excludedWorkloads map[string]bool
+
+	// Matches an ISO8601 full date
+	timestampRegexp = regexp.MustCompile(`^\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d(\.\d+)?(([+-]\d\d:\d\d)|Z)?(\+\d{4})?`)
+	severityRegexp  = regexp.MustCompile(`(\s+|^)(INFO|WARN|DEBUG|TRACE)\s+`)
 )
 
 func isWorkloadIncluded(workload string) bool {
@@ -190,10 +194,6 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
 	if err != nil {
 		return nil, err
 	}
-
-	// Matches an ISO8601 full date
-	timestampRegexp := regexp.MustCompile("^\\d{4}-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\d(\\.\\d+)?(([+-]\\d\\d:\\d\\d)|Z)?(\\+\\d{4})?")
-	severityRegexp := regexp.MustCompile("(\\s+|^)(INFO|WARN|DEBUG|TRACE)\\s+")
 
 	lines := strings.Split(podLog.Logs, "\n")
 	messages := make([]LogEntry, 0)
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -48,8 +48,7 @@ var (
 	excludedWorkloads map[string]bool
 
 	// Matches an ISO8601 full date
-	timestampRegexp = regexp.MustCompile(`^\d{4}-\d\d-\d\dT\d\d:\d\d:\d\d(\.\d+)?(([+-]\d\d:\d\d)|Z)?(\+\d{4})?`)
-	severityRegexp  = regexp.MustCompile(`(\s+|^)(INFO|WARN|DEBUG|TRACE)\s+`)
+	severityRegexp = regexp.MustCompile(`(\s+|^)(INFO|WARN|DEBUG|TRACE)\s+`)
 )
 
 func isWorkloadIncluded(workload string) bool {
@@ -207,18 +206,16 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
 			Severity:      "",
 		}
 
-		timestamp := timestampRegexp.FindString(line)
-		if timestamp != "" {
-			parsed, err := time.Parse("2006-01-02T15:04:05+0000", string(timestamp))
+		splitted := strings.Split(line, " ")
+		timestamp := splitted[0]
+		parsed, err := time.Parse("2006-01-02T15:04:05+0000", string(timestamp))
 
-			if err == nil {
-				message.Timestamp = timestamp
-				message.TimestampUnix = parsed.Unix()
-			}
+		if err == nil {
+			message.Timestamp = timestamp
+			message.TimestampUnix = parsed.Unix()
 		}
 
-		line = string(timestampRegexp.ReplaceAll([]byte(line), []byte("")))
-		message.Message = line
+		line = strings.Join(splitted[1:], " ")
 
 		severity := severityRegexp.FindString(line)
 		if severity != "" {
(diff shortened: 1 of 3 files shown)


business/workloads.go
```diff
@@ -29,6 +31,18 @@ type WorkloadService struct {
 	businessLayer *Layer
 }
 
+// Structures for workload log messages
+type PodLog struct {
+	Logs     string       `json:"logs,omitempty"`
+	Messages []LogMessage `json:"messages,omitempty"`
+}
+
+type LogMessage struct {
```
jshaughn, line 40: Maybe rename this to LogEntry as the Message is really just that text part.

business/workloads.go
```diff
@@ -29,6 +31,18 @@ type WorkloadService struct {
...
+	Message   string `json:"message,omitempty"`
+	Timestamp int64  `json:"timestamp,omitempty"`
+	Severity  string `json:"severity,omitempty"`
```
jshaughn, line 43: Alphabetizing is a great habit.

business/workloads.go
```diff
@@ -170,8 +184,63 @@ func (in *WorkloadService) GetPod(namespace, name string) (*models.Pod, error) {
...
+func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.PodLogOptions) (*PodLog, error) {
+	podLog, err := in.k8s.GetPodLogs(namespace, name, opts)
+
+	if err != nil {
+		return nil, err
+	}
+
+	// Matches an ISO8601 full date
+	timestampRegexp := regexp.MustCompile("^\\d{4}-\\d\\d-\\d\\dT\\d\\d:\\d\\d:\\d\\d(\\.\\d+)?(([+-]\\d\\d:\\d\\d)|Z)?(\\+\\d{4})?")
+	severityRegexp := regexp.MustCompile("(\\s+|^)(INFO|WARN|DEBUG|TRACE)\\s+")
```
jshaughn, line 196: Perhaps declare these as global vars so we only eval one time.  They can't be constants because you can't have a regexp const, but you can have a var that is basically treated like a constant.

business/workloads.go
```diff
@@ -170,8 +184,63 @@ func (in *WorkloadService) GetPod(namespace, name string) (*models.Pod, error) {
...
+
+	message := PodLog{
+		Logs:     podLog.Logs,
+		Messages: messages,
+	}
+
+	return &message, err
+}
+
+func (in *WorkloadService) GetPodLogs(namespace, name string, opts *core_v1.PodLogOptions) (*PodLog, error) {
```
jshaughn, line 242: Although our API is not supported I'm not sure we should break the client by changing the return structure for the API call.  I think perhaps we should either have two calls, one that returns the logs as a string, like today, and one that returns the new format.  Or, maybe we shoud have an option (queryParam) that specifies the desired return format, defaulting to today's string.

What do you think?

jshaughn (changes requested):
@cfcosta good progress, a couple of potential changes.  Also, travis is unhappy.

business/workloads.go
cfcosta: I don't really agree, no, as we do not break the compatibility, the client code can stay the same. Also this means that a partial integration is possible, before commiting to adding the changes to the backend.

jshaughn: OK, so I see now what you are doing.  We can temporarily return both the logs blob ('Logs') and log entries ('Entries') until we support only the entries on the client side.  I can't really see the need to maintain the blob format as the client could reconstruct lines or the whole blob as needed.

jshaughn (commented):
Looks good, only 1 or 2 more comments.

business/workloads.go
```diff
@@ -29,8 +31,24 @@ type WorkloadService struct {
...
+// Structures for workload log messages
+type PodLog struct {
+	Logs    string     `json:"logs,omitempty"`
+	Entries []LogEntry `json:"entries,omitempty"`
+}
+
+type LogEntry struct {
+	Message   string `json:"message,omitempty"`
+	Severity  string `json:"severity,omitempty"`
+	Timestamp int64  `json:"timestamp,omitempty"`
```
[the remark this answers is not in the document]
cfcosta: Unix timestamp, really, but I like the idea.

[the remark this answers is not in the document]
mtho11: I like the idea of the timestamp as it is easy to convert from json and doesn't require parsing.
The downside is that it is not easily human readable.

business/workloads.go
```diff
@@ -200,17 +201,19 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
...
-			Severity:  "",
+			Message:       line,
+			Timestamp:     "",
+			TimestampUnix: 0,
+			Severity:      "",
 		}
 
 		timestamp := timestampRegexp.FindString(line)
 		if timestamp != "" {
 			parsed, _ := time.Parse("2006-01-02T15:04:05+0000", string(timestamp))
```
jshaughn, line 212: I think you do want to capture 'err', no?

business/workloads.go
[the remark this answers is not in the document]
cfcosta: Yeah, I kept both because unix timestamps are easily parsed, but also kept the string part so we can reconstruct the string later.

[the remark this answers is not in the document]
jshaughn: Latest rev has both formats so I think that is good and satisfies both comments.

jshaughn (changes requested):
@cfcosta code looks good but when testing it seems the parsing is not working as expected.  I think the timestamp is not parsing out.  We use the Timestamps=true option of the k8s API call, so every line in the log blob should be prefixed with a  RFC3339 or RFC3339Nano timestamp string.  For me it looks like (for example);

"2020-10-06T17:46:19.865055293Z <the app log entry>"

If you can clip that and parse it you should be able to get the unix timestamp.  As for severity, that may or may not be found, I'd suggest just looking for ERROR | WARN | DEBUG | TRACE and if not found just assume Info level.

jshaughn (changes requested):
@cfcosta See https://github.com/cfcosta/kiali/pull/1

cfcosta:
@jshaughn merged your PR, should be ready to merge now.

jshaughn (changes requested):
@cfcosta from my prior comments: "You may need to update tests, I'm not sure."

cfcosta:
Pushed a fix, tests are ok here, let's see if travis agree :)

jshaughn (dismissed):
OK, tests are now passing. Looks OK to me.

pbajjuri20:
Here, It should be "INFO" because it is case-sensitive.

info

jshaughn:
> Here, It should be "INFO" because it is case-sensitive.

Actually, I think this is fine.  QE for this PR is actually just ensuring that logs are still being presented.  The UI has not been changed so unless what you are seeing is different than what you see using master it's not a problem.  I think the log message is actually writing it as lowercase.  Let us know,

cfcosta:
@jshaughn explaining it a little bit better: the code did not parse the `warn` correctly as the severity because the regexps are case-sensitive. It is a quick fix. It should not change the log display.

Also, something for later: we did find that because we add kubernetes timestamps to the logs, all logs on istio show two timestamps (on master, we don't change anything).

pbajjuri20:
Thanks for the Clarification @jshaughn 

"warn" issue is fixed perfectly @cfcosta 


before:

before warn


after:
warn 200

pbajjuri20:
Tested and Passed on Kiali UI 1.25.0 (843d76d6cbe3bda4f7156ccfd370336b930baed0)

pbajjuri20 (approved):

israel-hdez:
Kubernetes structured logs (introduced in 1.19) is something to keep an eye on, which is related to this: https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/

jshaughn:
> Kubernetes structured logs (introduced in 1.19)

Very good to know.  I guess apps need to use the new functions to gain the benefits, so not sure when we can leverage this.  And unfortunately the actual API for fetching logs doesn't seem to have been improved [yet].

PR #3292, opened 2020-10-05
molecule tests - allow dump logs on error to be turned off

jmazzitelli:
Sometimes it is annoying seeing all the logs dump on error (the real error sometimes scrolls off the console window and you lose it). So be able to turn that off.

This goes with PR https://github.com/kiali/kiali-operator/pull/156


jshaughn (approved):
Fine.

Issue #3219, opened 2020-09-17
KIA0301 validation rule does not work for multiple ingress-gateways


tehleet:
**Describe the bug**
I get KIA0301 warning in Kiali UI, that my Gateway configuration is bad, but Kiali doesn't check that I have 2 different ingress-gateways.

KIA0301 documentation says:

> Gateway creates a proxy that forwards the inbound traffic for the exposed ports. If two different gateways expose the same ports for the same host, this creates ambiguity inside Istio as either of these gateways could handle the traffic. This is most likely a configuration error. This check is done across all namespaces the user has access to.

but it doesn't check if used istio-ingress resources are different

**Versions used**
Kiali: 1.23.0
Istio: 1.6.5
Kubernetes flavour and version: AWS EKS 1.16

**To Reproduce**

1. I have 2 istio-ingresses - one is private and the other one is public facing. This is 2 different AWS NLB.
```BASH
➜ kubectl -n istio-system get svc
NAME                       TYPE           CLUSTER-IP       EXTERNAL-IP                                                                     PORT(S)                                                      AGE
istio-ingressgateway-prv   LoadBalancer   10.100.16.14     supersecret.amazonaws.com   15021:30378/TCP,80:32260/TCP,443:30289/TCP,15443:32552/TCP   13d
istio-ingressgateway-pub   LoadBalancer   10.100.227.209   supersecret.amazonaws.com   15021:32017/TCP,80:30360/TCP,443:31177/TCP,15443:30967/TCP   13d
istiod                     ClusterIP      10.100.137.149   <none>                      15010/TCP,15012/TCP,443/TCP,15014/TCP,853/TCP                13d
➜ 
```
2. I have 2 istio Gateways private and public which map istio-ingress private or public.
```YAML
---
kind: Gateway
apiVersion: networking.istio.io/v1alpha3
metadata:
  name: istio-gateway-prv
  namespace: istio-system
spec:
  servers:
    - hosts:
        - '*'
      port:
        name: http-prv
        number: 80
        protocol: HTTP
      tls:
        httpsRedirect: true
    - hosts:
        - '*'
      port:
        name: https-prv
        number: 443
        protocol: HTTPS
      tls:
        credentialName: supersecret
        mode: SIMPLE
  selector:
    istio: istio-ingressgateway-prv
---
kind: Gateway
apiVersion: networking.istio.io/v1alpha3
metadata:
  name: istio-gateway-pub
  namespace: istio-system
spec:
  servers:
[...]

PR #3296, opened 2020-10-06
Gateways can share host+port if they point to different ingress gws

xeviknal:
fixes https://github.com/kiali/kiali/issues/3219

The warning appears when two gateways share the same host, port and gateway labels.
We consider that if two gateways are pointing to the same gateway with different labels, then users should unify the criteria by using the same selector in both gateways.

Testing scenario: https://github.com/xeviknal/istio-lab/blob/master/kiali-3219/gateways.yaml

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -5,6 +5,8 @@ import (
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


hhovsepy (approved):
Verified, regression tested:
Screenshot from 2020-10-06 18-05-14

lucasponce (approved):

Issue #3274, opened 2020-09-30
Invalid dates in Iter8 Overview Card


lucasponce:
It seems at some point of the bump into Iter8 1.0.0-rcX we introduced a minor bug:

![image](https://user-images.githubusercontent.com/1662329/94676435-9c869c80-031b-11eb-9aa5-4e25bf0f7832.png)

cc-ing @jadeyliu939

PR #3308, opened 2020-10-08
Fix Iter8 time format

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3274

image

Now created/started/ended at have same format used in Kiali.


jotak (approved):
LGTM

Issue #3084, opened 2020-08-05
Support external OIDC providers


christian-posta:
Right now to use OIDC you have to tie it into the Kube API. On cloud-hosted k8s, it's not easy (or is impossible) to alter the kube api startup config. 

Can we support a standalone (Auth0, Keycloak, Dex, etc) OIDC provider?

israel-hdez:
Hi, @christian-posta 
We have heard of users that has successfully setup Kiali with Auth0, Keycloak and Dex. But, as you say, Kube API also needs to be setup for integration with the providers.

We can remove the requirement of the OIDC integration in Kube API, but RBAC capabilities will be lost. All users logging into Kiali will share the same privileges, because the OpenID users will be unknown to the API server. 

Another approach is to add support for API proxies (like `kube-oidc-proxy`) where the proxy translates OpenID users to regular kubernetes ServiceAccounts. This was requested in #3042.

What would be your expectation? Do you need RBAC, or are you OK with the shared privileges?

christian-posta:
Supporting cloud customers who cannot modify kube api with #3042 sounds interesting!
Otherwise having an option w/o RBAC for global admin users w/out tying to kube-api would work too.

ceastman-ibm:
And related to this would be our use case:
Allow a read only view (mainly for our development teams) but have the read only view limited to people that passed the openid authentication.

xFayre:
Hi @israel-hdez,
Managed clusters like AKS that support group RBAC rules also do not work. 

In our setup we have an AKS cluster with AAD integration enabled. Our idea was to let people use their AAD credentials to access Kiali.
Seeing that Kiali checks with the kube API to check permissions and the fact we cannot edit --oidc flags on a managed cluster makes it impossible to integrate with Azure.
Also, our AKS + AAD integration is based on Rolebindings to Azure's group Ids and User Ids. The fact that groups claim are not supported also makes it difficult to use Kiali in our scenario.

Either using code flow or implicit flow (and generating an otherwise valid JWT) does not seem to work.

I would be happy to provide more details

israel-hdez:
@xFayre perhaps, can you open new github issue with your details?
If your AKS + AAD integration is working, you should be able to use Kiali with OpenID. The omission of group claims in Kiali is because we delegate the permission check to Kubernetes. So, rolebindings to GIDs should work fine in Kiali as long as the underlying AKS+AAD works.

I would be interested in knowing what's the error that Kiali is outputting.

[...]

PR #3310, opened 2020-10-08
[OpenId] Support for clusters without OIDC integration (no RBAC will be possible)

israel-hdez:
Add support for using OpenId providers to login into Kiali even if the Kubernetes cluster has no OIDC integration. This is done by turning off RBAC and using the Kiali ServiceAccount token regardless of the logged in user.

Most of the code is related to adding several checks on the OpenId token. On the "integrated" scenario, these checks were delegated to the Kubernetes cluster. But on the "non-integrated" scenario (which this PR is adding support for) we are no longer passing the openid token to Kubernetes (because we are using the Kiali ServiceAccount token). Thus, without these checks, the token would not be fully validated.

This is adding two new dependencies:
* golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e -- used for multi-thread synchronization when fetching "common" data from the OpenId Server
* gopkg.in/square/go-jose.v2 v2.5.1 -- used for validating openid token signatures

Operator side PR: https://github.com/kiali/kiali-operator/pull/165

Related kiali/kiali#3084

### DIFF
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -1,7 +1,6 @@
 package business
 
 import (
-	"crypto"
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/sha256"
@@ -408,16 +407,6 @@ func GetOpenIdJwks() (*jose.JSONWebKeySet, error) {
 			return nil, fmt.Errorf("cannot parse OpenId JWKS document: %s", err.Error())
 		}
 
-	// TODO: Remove this loop. It's only for debugging
-	log.Debugf("OIDC has %i keys", len(oidcKeys.Keys))
-	for _, key := range oidcKeys.Keys {
-		thumbprint, e := key.Thumbprint(crypto.SHA256)
-		if e != nil {
-			thumbprint = []byte{0}
-		}
-		log.Debugf("OIDC Key: ID = %s, ALG = %s, Use = %s, Key thumbprint = %x", key.KeyID, key.Algorithm, key.Use, thumbprint)
-	}
-
 		cachedOpenIdKeySet = &oidcKeys // Store the keyset in a "cache"
 		return cachedOpenIdKeySet, nil
 	})
@@ -635,7 +624,7 @@ func ValidateOpenTokenInHouse(openIdParams *OpenIdCallbackParams) error {
 		// Let's do the minimal check to ensure that the token wasn't issued in the future
 		// we add a little offset to "now" to add one minute tolerance
 		iatTime := time.Unix(parsedIat, 0)
-		nowTime := util.Clock.Now().Add(60*time.Second)
+		nowTime := util.Clock.Now().Add(60 * time.Second)
 		if iatTime.After(nowTime) {
 			return fmt.Errorf("we don't like people living in the future - enjoy the present!; iat = '%d'", parsedIat)
 		}
@@ -655,7 +644,6 @@ func ValidateOpenTokenInHouse(openIdParams *OpenIdCallbackParams) error {
 	//       need to provide it nor we need to verify it.
 	//   - auth_time: we are not asking for this claim at authorization, so the IdP doesn't
 	//	     need to provide it nor we need to verify it.
-
 
 	// If execution flow reached this point, all claims look valid, but that won't guarantee that
 	// the id_token hasn't been tampered. So, we check the signature to find if
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/log"
+	"github.com/kiali/kiali/util"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


israel-hdez:
@sandhose You may want to review this, because it is adding the missing JWT checks you mentioned in https://github.com/kiali/kiali/pull/2798#issuecomment-636168758.

In this, users can opt to not configuring OIDC integration in Kubernetes and still be able to use OpenId in Kiali, but this requires a full validation of the JWT token.

I hope this is doing it correctly. Your review will be appreciated.

business/openid_auth.go
```diff
@@ -247,86 +272,150 @@ func GetOpenIdMetadata() (*OpenIdMetadata, error) {
 		return cachedOpenIdMetadata, nil
 	}
 
-	cfg := config.Get().Auth.OpenId
+	fetchedMetadata, fetchError, _ := openIdFlightGroup.Do("metadata", func() (interface{}, error) {
+		cfg := config.Get().Auth.OpenId
 
-	// Remove trailing slash from issuer URI, if needed
-	trimmedIssuerUri := strings.TrimRight(cfg.IssuerUri, "/")
+		// Remove trailing slash from issuer URI, if needed
```
cfcosta, line 278: I think your editor did something bad here :)

business/openid_auth.go
```diff
@@ -459,6 +548,134 @@ func ValidateOpenIdState(openIdParams *OpenIdCallbackParams) (validationFailure
...
+		nowTime := util.Clock.Now().Add(60 * time.Second)
+		if iatTime.After(nowTime) {
+			return fmt.Errorf("we don't like people living in the future - enjoy the present!; iat = '%d'", parsedIat)
+		}
+	}
+
+	// Check exp (expiration time) claim
+	// The OIDC spec says: "The current time MUST be before the time represented by the exp Claim"
+	// No tolerance for this check.
+	if !util.Clock.Now().Before(openIdParams.ExpiresOn) {
```
cfcosta, line 636: I think `if util.Clock.Now().After(openIdParams.ExpiresOn)` might look better here, wdyt?

handlers/authentication.go
```diff
@@ -699,14 +713,29 @@ func OpenIdCodeFlowHandler(w http.ResponseWriter, r *http.Request) bool {
...
+		// the token must be fully validated, as we no longer pass the OpenId token to the cluster API server.
+		// Since the configuration indicates RBAC is off, we do the validations:
+		err = business.ValidateOpenTokenInHouse(openIdParams)
+		if err != nil {
+			RespondWithDetailedError(w, http.StatusForbidden, "the OpenID token was rejected", err.Error())
+			return true
+		}
+	}
+
+	if !conf.Auth.OpenId.DisableRBAC {
```
cfcosta, line 727: You could just use a else here, or short circuit on the first if, keeping this one without a conditional.

handlers/authentication.go
```diff
@@ -751,13 +780,12 @@ func OpenIdCodeFlowHandler(w http.ResponseWriter, r *http.Request) bool {
 		Value:    base64.StdEncoding.EncodeToString(cipherSessionData),
 		Expires:  openIdParams.ExpiresOn,
 		HttpOnly: true,
-		Path:     config.Get().Server.WebRoot,
+		Path:     conf.Server.WebRoot,
```
cfcosta, line 783: good one.

cfcosta (changes requested):
I did some smoke testing with this, setting up a server without setting up RBAC, and setting the configuration option. Could use it as expected.

Code seems ok, just some small changes that I think can make it better.

business/openid_auth.go
israel-hdez: Is it because of the extra indentation?
If so, I think it's fine. All this code is now enclosed in an anonymous function. on line 275 (two lines before this one):

`fetchedMetadata, fetchError, _ := openIdFlightGroup.Do("metadata", func() (interface{}, error) {`

or is it something else?

business/openid_auth.go
israel-hdez: I also though about it, but I would need something like `AfterOrEquals`. I'm not sure if there is that function. Otherwise, I could try `util.Clock.Now().After(openIdParams.ExpiresOn) || util.Clock.Now().Equal(openIdParams.ExpiresOn)`

Thoughts?

handlers/authentication.go
israel-hdez: My eyes didn't saw that... I'll change...

israel-hdez:
@cfcosta Fixed one issue
The other two are still open for discussion.

business/openid_auth.go
cfcosta: You are correct, no need to change then :)

business/openid_auth.go
cfcosta: Yeah, the indentation is mixed with tabs/spaces.

israel-hdez: Hmmmmmm... I reformatted, but it does not detect any changes...

I consistently see tabs on all changed lines... the added indentation are tabs and the existent indentation are tabs too :/ Even here in GitHub when selecting text the browser is highlighting tabs...

Can you check? I would think these formatting issues would be detected by the linters in the CI.

jshaughn: I checked out the PR and the formatting is fine for me.  I think this comment can be resolved.  If this is the only issue then we can approve.

jshaughn:
@mattmahoneyrh (same comment as in the related operator PR) We have upstream users waiting for this issue to be resolved and willing to test. So I am asking @israel-hdez to merge for the 1.26 release. Does QE have the ability to test OpenID related issues? It is something we may need to address going forward.

jshaughn (approved):
LGTM

Issue #3315, opened 2020-10-13
Short term caching of Prometheus queries in Health Service


lucasponce:
During review and testing of https://github.com/kiali/kiali-ui/pull/1957#issuecomment-706261371

It was found some scenarios where same health queries were used in the context of a single high level request / operation.

More details in comment:
https://github.com/kiali/kiali/issues/3252#issuecomment-704813959

A possible inmediate strategy for this could be to allow to short-term cache results, same queries for sure, and potentially short intervals could be also cached (a repeated query but in a <2-5seconds).

This issue is to PoC this scenario and analyze if same technique applied into namespace, proxy-status could be applied for these queries as well.

PR #3316, opened 2020-10-13
Short term expiration cache for Health Prom queries

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3315 

Analyzing the queries involved in a graph + healh generation we can see a pattern like this [1].
There are a level of repetitive queries when involved multiple namespaces and health.

This PR introduces caching on the prometheus/client.go API, similar as used for namespaces and proxy-status in the kubernetes/cache/cache.go.

As a first result, with a short term expiration of 10 seconds (used by default) it gets a hit ratio of ~76%.

It approaches only the `Get*` API used mainly for Health.

Also, it's configurable (enabled/disabled and duration).

In terms of design I think it's better to keep separated the three major domains: kubernetes, prometheus and jaeger, as each APIs may have their own particularities. I was tempted to propose a single abstraction for all of them but I think in this way design is simplified and clean.

Note, this requires operator and helm-chart additions for new config.
https://github.com/kiali/kiali-operator/pull/162.

[1] [get-all-requests-rates-no-cache.txt](https://github.com/kiali/kiali/files/5377070/get-all-requests-rates-no-cache.txt)

### DIFF
--- a/prometheus/cache.go
+++ b/prometheus/cache.go
@@ -71,7 +71,7 @@ func (c *promCacheImpl) GetAllRequestRates(namespace string, ratesInterval strin
 
 	if nsRates, okNs := c.cacheAllRequestRates[namespace]; okNs {
 		if rtInterval, okRt := nsRates[ratesInterval]; okRt {
-			if queryTime.Sub(rtInterval.queryTime) < c.cacheDuration {
+			if !queryTime.Before(rtInterval.queryTime) && queryTime.Sub(rtInterval.queryTime) < c.cacheDuration {
 				log.Tracef("[Prom Cache] GetAllRequestRates [namespace: %s] [ratesInterval: %s] [queryTime: %s]", namespace, ratesInterval, queryTime.String())
 				return true, rtInterval.inResult
 			}
--- a/prometheus/cache.go
+++ b/prometheus/cache.go
@@ -37,6 +37,7 @@ type (
 
 	promCacheImpl struct {
 		cacheDuration          time.Duration
+		cacheExpiration        time.Duration
 		cacheAllRequestRates   map[string]map[string]timeInResult
 		cacheAppRequestRates   map[string]map[string]map[string]timeInOutResult
 		cacheNsSvcRequestRates map[string]map[string]timeInResult
@@ -54,14 +55,19 @@ func NewPromCache() PromCache {
 	kConfig := kialiConfig.Get()
 
 	cacheDuration := time.Duration(kConfig.ExternalServices.Prometheus.CacheDuration) * time.Second
+	cacheExpiration := time.Duration(kConfig.ExternalServices.Prometheus.CacheExpiration) * time.Second
 	promCacheImpl := promCacheImpl{
 		cacheDuration:          cacheDuration,
+		cacheExpiration: 		cacheExpiration,
 		cacheAllRequestRates:   make(map[string]map[string]timeInResult),
 		cacheAppRequestRates:   make(map[string]map[string]map[string]timeInOutResult),
 		cacheNsSvcRequestRates: make(map[string]map[string]timeInResult),
 		cacheSvcRequestRates:   make(map[string]map[string]map[string]timeInResult),
 		cacheWkRequestRates:    make(map[string]map[string]map[string]timeInOutResult),
 	}
+
+	go promCacheImpl.watchExpiration()
+
 	return &promCacheImpl
 }
 
@@ -233,4 +239,32 @@ func (c *promCacheImpl) SetWorkloadRequestRates(namespace string, workload strin
 		outResult: outResult,
 	}
 	log.Tracef("[Prom Cache] SetAppRequestRates [namespace: %s] [workload: %s] [ratesInterval: %s] [queryTime: %s]", namespace, workload, ratesInterval, queryTime.String())
+}
+
+// Expiration is done globally, this cache is designed as short term, so in the worst case it would populated the queries
+// Doing an expiration check per item is costly and it's not necessary in this particular context
+func (c *promCacheImpl) watchExpiration() {
+	for {
+		time.Sleep(c.cacheExpiration)
+		c.allRequestRatesLock.Lock()
+		c.cacheAllRequestRates = make(map[string]map[string]timeInResult)
+		c.allRequestRatesLock.Unlock()
+
+		c.appRequestRatesLock.Lock()
+		c.cacheAppRequestRates = make(map[string]map[string]map[string]timeInOutResult)
+		c.appRequestRatesLock.Unlock()
+
+		c.nsSvcRequestRatesLock.Lock()
+		c.cacheNsSvcRequestRates = make(map[string]map[string]timeInResult)
+		c.nsSvcRequestRatesLock.Unlock()
+
+		c.svcRequestRatesLock.Lock()
+		c.cacheSvcRequestRates = make(map[string]map[string]map[string]timeInResult)
+		c.svcRequestRatesLock.Unlock()
+
+		c.wkRequestRatesLock.Lock()
+		c.cacheWkRequestRates = make(map[string]map[string]map[string]timeInOutResult)
+		c.wkRequestRatesLock.Unlock()
+		log.Infof("[Prom Cache] Expired")
+	}
 }
--- a/prometheus/cache.go
+++ b/prometheus/cache.go
@@ -58,7 +58,7 @@ func NewPromCache() PromCache {
 	cacheExpiration := time.Duration(kConfig.ExternalServices.Prometheus.CacheExpiration) * time.Second
 	promCacheImpl := promCacheImpl{
 		cacheDuration:          cacheDuration,
-		cacheExpiration: 		cacheExpiration,
+		cacheExpiration:        cacheExpiration,
 		cacheAllRequestRates:   make(map[string]map[string]timeInResult),
 		cacheAppRequestRates:   make(map[string]map[string]map[string]timeInOutResult),
 		cacheNsSvcRequestRates: make(map[string]map[string]timeInResult),
--- a/prometheus/cache.go
+++ b/prometheus/cache.go
@@ -102,7 +102,7 @@ func (c *promCacheImpl) GetAppRequestRates(namespace string, app string, ratesIn
 	if nsRates, okNs := c.cacheAppRequestRates[namespace]; okNs {
 		if appInterval, okApp := nsRates[app]; okApp {
 			if rtInterval, okRt := appInterval[ratesInterval]; okRt {
-				if queryTime.Sub(rtInterval.queryTime) < c.cacheDuration {
+				if !queryTime.Before(rtInterval.queryTime) && queryTime.Sub(rtInterval.queryTime) < c.cacheDuration {
 					log.Tracef("[Prom Cache] GetAppRequestRates [namespace: %s] [app: %s] [ratesInterval: %s] [queryTime: %s]", namespace, app, ratesInterval, queryTime.String())
 					return true, rtInterval.inResult, rtInterval.outResult
 				}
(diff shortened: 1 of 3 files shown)


jotak:
Playing a little exercise to try to see how the duration impact refreshes, and which would be the best tradeoff for a default value.
The lag latency is already bound to prometheus scrape period which is usually 15s. Depending on how scrapes and cache refresh are synced / out of sync, we will have a different lag (plus to combine with the refresh period set up in the UI). 

Here's some scenarios:

```
|----|----|----|----|----|----|
0    5    10   15   20   25   30

1. With d=10s

Synced case:
S--------------S--------------S
C---------C---------C---------C
               <---->
               Best = 5s additional lag

Unsynced case:
S--------------S--------------S
----C---------C---------C---------C
               <-------->
               Worst = 9s additional lag

2. With d=5s

Synced case:
S--------------S--------------S
C----C----C----C----C----C----C
               |
               Best = 0s additional lag

Unsynced case:
S--------------S--------------S
----C----C----C----C----C----C----C
               <--->
               Worst = 4s additional lag

S = Scrape
C = Cache refresh
```

So, no surprise that having a smaller value reduces the lag, and of course on the other hand it reduces the cache efficiency.
What do you think? Personally I'd be in favour of something like 5s as default.

prometheus/cache.go
```diff
@@ -0,0 +1,236 @@
...
+}
+
+func (c *promCacheImpl) GetAppRequestRates(namespace string, app string, ratesInterval string, queryTime time.Time) (bool, model.Vector, model.Vector) {
+	defer c.appRequestRatesLock.RUnlock()
+	c.appRequestRatesLock.RLock()
+
+	if nsRates, okNs := c.cacheAppRequestRates[namespace]; okNs {
+		if appInterval, okApp := nsRates[app]; okApp {
+			if rtInterval, okRt := appInterval[ratesInterval]; okRt {
+				if queryTime.Sub(rtInterval.queryTime) < c.cacheDuration {
```
jotak, line 105: Take care here if I'm correct, `queryTime` can be in past, so possibly lower than the query time stored previously in cache, hence diff being negative. In that case I guess we shouldn't hit cache?

jotak: (I'm not talking for `GetAppRequestRates` specifically, but more in general when fetching metrics)

jotak: I realize this cache is at the moment only for health-kind of queries, so it eliminates the case of fetching raw metrics in the past.... but still I think in the context of Graph Replay we do query health in the past, right? cc @jshaughn

jotak:
Also something to consider perhaps: using this (quite popular) cache implementation instead of implementing ours? https://github.com/patrickmn/go-cache
One difference would be to have flat cache instead of nested structures with maps. Which implies some way to generate a cache key out of parameters. Perhaps it's easier to reason about? I don't know, just suggesting to check the pros/cons.

prometheus/cache.go
lucasponce: In which scenarios ? Perhaps I missed some use case.

lucasponce:
> Also something to consider perhaps: using this (quite popular) cache implementation instead of implementing ours? https://github.com/patrickmn/go-cache

The design we have in place is working for quite some time without major issues.
I don't see a good reason to move to this framework.

hhovsepy (approved):
Regression testing passed.

prometheus/cache.go
jotak: Actually, that's probably ok. `time.Sub` is not signed, so no matter that `queryTime` might be smaller than `rtInterval.queryTime`.
I'm just trying to figure out from code, not tested yet, in any case I guess it's important to test if health is correctly displayed during a Graph Replay as it may have some impacts there.

prometheus/cache.go
```diff
@@ -0,0 +1,236 @@
...
+		SetWorkloadRequestRates(namespace, workload, ratesInterval string, queryTime time.Time, inResult model.Vector, outResult model.Vector)
+	}
+
+	promCacheImpl struct {
+		cacheDuration          time.Duration
+		cacheAllRequestRates   map[string]map[string]timeInResult
+		cacheAppRequestRates   map[string]map[string]map[string]timeInOutResult
+		cacheNsSvcRequestRates map[string]map[string]timeInResult
+		cacheSvcRequestRates   map[string]map[string]map[string]timeInResult
+		cacheWkRequestRates    map[string]map[string]map[string]timeInOutResult
```
jshaughn, line 44: What is the advantage of map-of-maps[-of-maps] as opposed to constructing a single key and using a single map?

prometheus/cache.go
jshaughn: > `if queryTime.Sub(rtInterval.queryTime) < c.cacheDuration`

Looking,  I don't think this is safe.  The cache only works if the queryTime is equal to or later than when the cache entry was stored.  queryTime.Sub will return a negative duration if queryTime is before rtInterval.queryTime.  And a negative duration would always be < c.cacheDuration.  I think you need something like, 
'''
`if !queryTime.Before(rtInterval.queryTime) && queryTime.Sub(rtInterval.queryTime) < c.cacheDuration`

jshaughn (commented):
Given your investigation showing duplicate queries it does seem like we could get some benefit from the caching approach when everything other than the queryTime is equal and the queryTime is within a small delta of the cached value time.  Memory is of course negatively impacted and that is a little concerning given that some of this retrieved data could be large.  I think the cacheDuration should be fairly small, no more than 1/2 the prom scrape interval (rounded down), to minimize worst-case lag.

prometheus/cache.go
lucasponce: Several advantages:
- It structures caching per categories (namespace, item, rate interval).
- It allows to invalidate / refresh per namespace (not yet introduced in this iteration but something to consider later).
- It saves to manipulate and build single primary keys for items.
- It follows similar design as used in kubernetes/caching.

lucasponce:
> Given your investigation showing duplicate queries it does seem like we could get some benefit from the caching approach when everything other than the queryTime is equal and the queryTime is within a small delta of the cached value time. 

+1

> Memory is of course negatively impacted and that is a little concerning given that some of this retrieved data could be large. 

Well I'm less concerning about this on this case, yes, it will produce a little stress on memory, but don't assume that querying p8s is free memory consumption, potentially would consume similar memory plus network jump. It would depend of golang memory planner to release the memory, potentially having buffers still used on repeated queries. So, using caching may be even less memory intense on stress situations.

> I think the cacheDuration should be fairly small, no more than 1/2 the prom scrape interval (rounded down), to minimize worst-case lag.

+1 let's start with that.

Of course there will be more possibilities here, this iteration was the most obvious one.

lucasponce:
> The lag latency is already bound to prometheus scrape period which is usually 15s.

+1

scrape / 2 could be a good default number to start with.

prometheus/cache.go
lucasponce: Oh, I think Github losses my response on this comment.
+1 on it.

jotak: > queryTime.Sub will return a negative duration

Sure about that? Duration can be negative? That was my initial concern but it isn't clear in the documentation I've read

jotak:
7s is better than 10 but IMO still a lot of lag when people want to see changing health asap. But if you both agree on this value, then ok...

jotak:
> > Also something to consider perhaps: using this (quite popular) cache implementation instead of implementing ours? https://github.com/patrickmn/go-cache
> 
> The design we have in place is working for quite some time without major issues.
> I don't see a good reason to move to this framework.

Sorry but I don't think it should stop us from thinking about it. Especially when we add more and more caches. Not saying this cache solution is better, just asking to consider it.

One quick example, from what I'm seeing in this PR, it seems there is no cache invalidation, am I wrong? I just see that cache gets overwritten when we add fresher data. So, if the cache keys change (eg: user renames his workloads ; changes his rate intervals ; ...) then old data stays forever? It's a kind a thing that a well known / tested cache solution would take care of. I think establishing a list of pros/cons wouldn't hurt.

lucasponce:
> Sorry but I don't think it should stop us from thinking about it. Especially when we add more and more caches. Not saying this cache solution is better, just asking to consider it.

> One quick example, from what I'm seeing in this PR, it seems there is no cache invalidation, am I wrong? I just see that cache gets overwritten when we add fresher data. So, if the cache keys change (eg: user renames his workloads ; changes his rate intervals ; ...) then old data stays forever? It's a kind a thing that a well known / tested cache solution would take care of. I think establishing a list of pros/cons wouldn't hurt.

I see this comment a little bit out of the context, focus to introduce a library instead to address a specific problem we want to solve.

Caching is introducing in Kiali for some time at several levels: tokens, namespaces, proxy-status, kubernetes objects, and now extending design to Prom Queries.

To me, to introduce a new dependency, and change the design should have a better clear motivation to solve a problem.

In short I see that the way we structure the caching is better suit to our needs and introducing a new plan key, value cache would imply to change.

Expiration is a good feature we don't have by default, I take that as part of the motivation of the change in a future if we see need, but before that we need that feature probably we would consider other areas to address in terms of performance.

But on this PR I wouldn't perform such change, or I doesn't see fit to plan it in short term, in that sense I'd see other aspects related to performance that could go first in the backend.

lucasponce:
Talked with @jotak about this PR to avoid misunderstandings and put more context to positions.

- To review https://github.com/patrickmn/go-cache is definitely a good idea and it would be good to see if it matches our future requirements.
- That task would be outside of this PR, and as we are in the same page that introducing a new library is costly, we need to review if it suits to our business mode. Some caching features have a cost (more locking, potentially more CPU penalty) and it's a tradeof to have a balance solution.
- I'd like to treat expiration globally, not only for this effort, but also in all caching we are working and with more data and more business driven (instead to delegate in a library, which tends to add locking, to capture some events - namespace deleting - or see if we need other library with more sophisticated LRU for it).

lucasponce:
@jotak putting together all discussions I've added a last commit to address both concerns:
- Adding a global expiration guard.
- But avoiding to lock per item, as discussed this is one my concerns to not rely on individual items expirations, as it tend to create a stop block.

It requires a change in the operator PR.

prometheus/cache.go
jshaughn: @jotak Yes, I tested it yesterday just to make sure: For example:

```
func main() {
	start := time.Date(2000, 1, 1, 0, 0, 0, 0, time.UTC)
	end := time.Date(2000, 1, 1, 12, 0, 0, 0, time.UTC)

	difference := end.Sub(start)
	fmt.Printf("difference = %v\n", difference)

	difference = start.Sub(end)
	fmt.Printf("difference = %v\n", difference)
	d10s, _ := time.ParseDuration("10s")
	fmt.Printf("%s < 10s = %v", difference, (difference < d10s))
}

difference = 12h0m0s
difference = -12h0m0s
-12h0m0s < 10s = true
```

hhovsepy (approved):
Retested only the full functionality regression.

lucasponce:
@jotak @jshaughn 

I think most of your comments and suggestions have been addressed.

I'd consider out of the scope of this PRs the valid discussion to study if we need additional libraries when we grow in cache using.

@hhovsepy has tested twice this work and it address a real problem we have today with duplication of queries.

jotak (approved):
LGTM, this approach looks good.
Though I think we should write some tests to cover cache + expiration. Perhaps in a next PR, but to keep in mind (sooner is better).

jshaughn (approved):
This looks good to me.  I know we proved there were cache hits but did we ever get any timing comparisons?

lucasponce:
> I know we proved there were cache hits but did we ever get any timing comparisons?

In short, not in a scientific way, meaning that when tracing is enabled you can see filter and compare Get vs Set calls, but I didn't collect time differences between Get and Set. Set implies more calls to a different service with all it requires to create a new result. A Get allows to reuse the previous result.

Empirically, you see that the health is pretty fast in graph + list pages when hits cache.

Issue #3271, opened 2020-09-29
VS icon missing in Kiali-UI for application graph of a VS with TLS routes.


abhioncbr:
**General remarks**

I have a very simple sample microservice applications configured based on istio secure gateway & istio SSL passthrough gateway settings. In the case of the SSL passthrough example, the virtual service icon is missing in Kiali UI for application graph. Manifests for SSL passthrough gateway and VS are as follows

```yaml
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  namespace: istio-ssl-passthrgh-sample
  name: istio-ssl-passthrgh-gateway
spec:
  selector:
    istio: internal-ingressgateway # use istio default controller
  servers:
  - port:
      number: 8443
      name: https
      protocol: HTTPS
    tls:
      mode: PASSTHROUGH
    hosts:
      - istio-ssl-passthrgh-nginx-exmple.aks.com
---
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  namespace: istio-ssl-passthrgh-sample
  name: frontend-nodejs-app-vs
spec:
  gateways:
  - istio-ssl-passthrgh-gateway.istio-ssl-passthrgh-sample.svc.cluster.local
  hosts:
  - istio-ssl-passthrgh-nginx-exmple.aks.com
  tls:
  - match:
    - port: 8443
      sniHosts:
      - istio-ssl-passthrgh-nginx-exmple.aks.com
    route:
      - destination:
          host: frontend-nodejs-app-svc
          port: 
            number: 8443
          subset: v1
        weight: 50
      - destination:
          host: frontend-nodejs-app-svc
          port: 
            number: 8443
          subset: v2
        weight: 50  
```

Added below the image for your reference. You can see in the lower app, VS icon is missing.

<img width="931" alt="Screen Shot 2020-09-29 at 4 11 13 PM" src="https://user-images.githubusercontent.com/1076944/94611777-3abd2880-0270-11eb-9ecd-095088b02d6d.png">

**Describe the bug**

For UI issues please also add a screenshot that shows the issue.

**Versions used**
Kiali: 1.17.9
Istio: 1.6.5
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9) AKS, 1.17.9

**To Reproduce**

**Expected behavior**

jmazzitelli:
cc @jshaughn

jshaughn:
[...]

PR #3319, opened 2020-10-13
Add 'tls' Spec to virtual service matching.

jshaughn:
Fixes https://github.com/kiali/kiali/issues/3271

Virtual Service matching for services was limited to 'http' and 'tcp' protocols, this PR adds 'tls'.  This is purely for visualization, the badging applied in the graph should now show VS badges for relevant service nodes.

To test, apply the following virtual service to bookinfo:

```
cat <<VSD | kubectl -n bookinfo create -f -
kind: VirtualService
metadata:
  namespace: bookinfo
  name: badge-test
spec:
  hosts: 
  - badge-test.test.com
  tls:
  - match:
    - port: 8443
      sniHosts:
      - badge-test.test.com
    route:
      - destination:
          host: details
          port: 
            number: 8443
          subset: v1
        weight: 100
VSD
```

Note that this just a bogus VS, the PR is only related to detecting the VS for badging purposes.  Without the PR the details service node in the graph should not show a VS badge.  WIth the PR, it should.

### DIFF
--- a/models/virtual_service.go
+++ b/models/virtual_service.go
@@ -58,9 +58,10 @@ func (vService *VirtualService) IsValidHost(namespace string, serviceName string
 		return false
 	}
 
-	protocolNames := []string{"http", "tcp"}
+	protocolNames := []string{"http", "tls", "tcp"} // ordered by matching preference
 	protocols := map[string]interface{}{
 		"http": vService.Spec.Http,
+		"tls":  vService.Spec.Tls,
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):

mattmahoneyrh (approved):

jmazzitelli:
Just an FYI: this is what you should expect to see now (creating the VS in the description, and then look at `details` badging):

tls-virtual-service

Issue #3318, opened 2020-10-13
Proxy-status reporting out of sync for all pods


jshaughn:
I'm not sure what the issue is but it seems that the proxy-status reporting is not working for me.
openshift 4.5
bookinfo
istio 1.7.3
kiali master (1.25+)

![image](https://user-images.githubusercontent.com/2104052/95886450-985f7380-0d4c-11eb-92dd-74f29d62eb55.png)

Every workload/app node reports "0 synced proxy", yet from the CLI:

```
$ _output/istio-1.7.3/bin/istioctl proxy-status
NAME                                                   CDS        LDS        EDS        RDS          ISTIOD                      VERSION
details-v1-558b8b4b76-x9242.bookinfo                   SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
istio-egressgateway-669785ffb4-sm4kj.istio-system      SYNCED     SYNCED     SYNCED     NOT SENT     istiod-75c5776d98-lcmlm     1.7.3
istio-ingressgateway-76447f8f85-9gg65.istio-system     SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
kiali-traffic-generator-n8pjf.bookinfo                 SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
mongodb-v1-78d87ffc48-88rk9.bookinfo                   SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
productpage-v1-6987489c74-7g99w.bookinfo               SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
ratings-v1-7dc98c7588-w8cml.bookinfo                   SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
ratings-v2-86f6d54875-25btb.bookinfo                   SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
reviews-v1-7f99cc4496-8txf7.bookinfo                   SYNCED     SYNCED     SYNCED     SYNCED       istiod-75c5776d98-lcmlm     1.7.3
[...]

PR #3321, opened 2020-10-14
Fix missing ProxyStatus check

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3318

I'm sure this is going to fix a flawn that is happening in your scenario.
Added an extra warning for a hidden error that perhaps may show up something extra (i.e. not access to the istiod for some reason).


jshaughn (approved):
LGTM.  The main problem was in my env, I was using an outdated kiali operator with the newer kiali server.  But changes in the PR still look valid and useful.

Issue #3313, opened 2020-10-12
Session not cleared correctly on log-out


israel-hdez:
**General remarks**

The session is not cleared correctly after logging out.

In OpenShift, when a user logs out, Kiali asks to OpenShift to invalidate the token. Since the session is not cleared correctly in Kiali it becomes stale. If you try to access Kiali again (either trying to re-login, or by accessing the Kiali root URL), it "thinks" that there is still a session (because cookies are not cleared correctly). But since OpenShift has already banned the token, Kiali enters in a blocked state where it detects that the token is invalid, but trying to log-in is blocked because of the stale session. So, it isn't possible to login again.

Workaround is to clear browser cookies.

This was seen in OpenShift strategy. It may affect other login strategies.

**Versions used**
Kiali: 1.24
Istio: N/A
Kubernetes flavour and version: Seen on OpenShift 4.5

**To Reproduce**
1. Login successfully into Kiali
2. Use the "Log out" option from the dropdown menu on the username on the top-right of the screen.
3. Try to re-login. It will fail.

**Expected behavior**
User should be able to login again.

israel-hdez:
Back-port to v1.24 is #3329 .

PR #3326, opened 2020-10-15
Fix the path of the cookie for OpenShift authentication strategy

israel-hdez:
The missing path directive for OpenShift authentication is creating a cookie targeting https://{fqdn}/{webroot}/api (notice the "api" sub-path).

When logging out, cookies are cleared using the Kiali root path (i.e. https://{fqdn}/{webroot}). This causes a mismatch in the browser and the session cookie is not cleared for OpenShift strategy. As a consequence, Kiali ends in a bad state where it's not possible to login again (workaround is to manually clear browser cookies).

This fixes the creation of the session cookie for OpenShift by specifying the root path. Also, when deleting the session cookie, in addition to setting an expiry time in the past, the MaxAge attribute is also passed with negative value.

Fixes kiali/kiali#3313

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -138,6 +138,7 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
 		Value:    tokenString,
 		Expires:  expiresOn,
 		HttpOnly: true,
+		Path:     config.Get().Server.WebRoot,
 		SameSite: http.SameSiteStrictMode,
 	}
 	http.SetCookie(w, &tokenCookie)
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
The Path change is good.  You can remove Expires when you already set MaxAge: -1 but I guess it can't hurt.

I couldn't reproduce after these changes. I think we can merge and backport and wait to see if anyone can surface it again.  It's not easy to reproduce so my tests may not be conclusive.

israel-hdez:
@jshaughn Thanks for the extensive testing.

PR #3331, opened 2020-10-20
Implement duration filtering for logs

### LINKED ISSUE #3303: Support "duration" parameter for pod logs API
Epic: https://github.com/kiali/kiali/issues/2232

Now that the server-side has logEntry support we can perform some more smart filtering to the backend and avoid pushing dead data over the wire.

The primary use case is to have log queries be able to support the same thing we have for metrics, and that's time-range support.  Logs already supports "&sinceTime", which is equivalent to "&queryTime" support in the metrics API.  We need to add "&duration" to the logs API.

The k8s logs API has no support for duration/endTime and so it will inefficiently return everything after sinceTime.  We will need to manually perform "tailLines" support when duration is specified (or more specifically, when sinceTime+duration < now).

cfcosta:
Closes #3303.

This adds support to a new `duration` param to the logs endpoint. It supports the usual duration format we use everywhere else (`1m`, `1h` and so on).

It has less priority than tail lines filtering, so their behaviours should not clash.

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -206,9 +206,7 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 	}
 
 	lines := strings.Split(podLog.Logs, "\n")
-	messages := make([]LogEntry, 0)
-
-	lineCount := int64(0)
+	entries := make([]LogEntry, 0)
 
 	var startTime *time.Time
 	if k8sOpts.SinceTime != nil {
@@ -216,14 +214,6 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 	}
 
 	for _, line := range lines {
-		if tailLines != nil {
-			if *tailLines <= lineCount {
-				break
-			} else {
-				lineCount += 1
-			}
-		}
-
 		entry := LogEntry{
 			Message:       "",
 			Timestamp:     "",
@@ -268,12 +258,16 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 			entry.Severity = strings.ToUpper(severity)
 		}
 
-		messages = append(messages, entry)
+		entries = append(entries, entry)
+	}
+
+	if tailLines != nil && len(entries) > int(*tailLines) {
+		entries = entries[len(entries)-int(*tailLines):]
 	}
 
 	message := PodLog{
 		Logs:    podLog.Logs,
-		Entries: messages,
+		Entries: entries,
 	}
 
 	return &message, err
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -209,6 +209,7 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 	entries := make([]LogEntry, 0)
 
 	var startTime *time.Time
+	var endTime *time.Time
 	if k8sOpts.SinceTime != nil {
 		startTime = &k8sOpts.SinceTime.Time
 	}
@@ -243,8 +244,15 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 				startTime = &parsed
 			}
 
-			if opts.Duration != nil && parsed.After(startTime.Add(*opts.Duration)) {
-				break
+			if opts.Duration != nil {
+				if endTime == nil {
+					end := parsed.Add(*opts.Duration)
+					endTime = &end
+				}
+
+				if parsed.After(*endTime) {
+					break
+				}
 			}
 
 			entry.TimestampUnix = parsed.Unix()
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -45,6 +45,11 @@ type LogEntry struct {
 	TimestampUnix int64  `json:"timestampUnix,omitempty"`
 }
 
+type LogOptions struct {
+	Duration *time.Duration
+	core_v1.PodLogOptions
+}
+
 var (
 	excludedWorkloads map[string]bool
 
@@ -189,8 +194,9 @@ func (in *WorkloadService) GetPod(namespace, name string) (*models.Pod, error) {
 	return &pod, nil
 }
 
-func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.PodLogOptions) (*PodLog, error) {
-	podLog, err := in.k8s.GetPodLogs(namespace, name, opts)
+func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOptions) (*PodLog, error) {
+	k8sOpts := opts.PodLogOptions
+	podLog, err := in.k8s.GetPodLogs(namespace, name, &k8sOpts)
 
 	if err != nil {
 		return nil, err
@@ -198,8 +204,17 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
 
 	lines := strings.Split(podLog.Logs, "\n")
 	messages := make([]LogEntry, 0)
+	lineCount := int64(0)
 
 	for _, line := range lines {
+		if k8sOpts.TailLines != nil {
+			if *k8sOpts.TailLines <= lineCount {
+				break
+			} else {
+				lineCount += 1
+			}
+		}
+
 		entry := LogEntry{
 			Message:       "",
 			Timestamp:     "",
@@ -247,7 +262,7 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
 	return &message, err
 }
 
-func (in *WorkloadService) GetPodLogs(namespace, name string, opts *core_v1.PodLogOptions) (*PodLog, error) {
+func (in *WorkloadService) GetPodLogs(namespace, name string, opts *LogOptions) (*PodLog, error) {
 	return in.getParsedLogs(namespace, name, opts)
 }
 
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -196,6 +196,9 @@ func (in *WorkloadService) GetPod(namespace, name string) (*models.Pod, error) {
 
 func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOptions) (*PodLog, error) {
 	k8sOpts := opts.PodLogOptions
+	tailLines := k8sOpts.TailLines
+	k8sOpts.TailLines = nil
+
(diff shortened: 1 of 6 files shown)


handlers/workloads.go
```diff
@@ -178,7 +178,7 @@ func PodLogs(w http.ResponseWriter, r *http.Request) {
 	queryParams := r.URL.Query()
 
 	// Get business layer
-	business, err := getBusiness(r)
+	businessLayer, err := getBusiness(r)
```
lucasponce, line 181: All handlers are using same naming pattern on this code:
image

I'd suggest to not renaming existing variables, otherwise code style will change.

cfcosta: I did this because, in this context, business can mean both the variable and the package name (and I need types defined inside the package).

Can you think another solution that would make sense here?

lucasponce: I haven't review your PR in deep, just that came to my attention in a first look.
If you have a clash it seems you are adding some model into the business package that you need to import.
You can use the model package for that, it's where typically all models used by Kiali on this logic are used.

cfcosta: I feel like those are more internal types to the package, but sounds good, will move them there.

lucasponce: On the contrary. The Kiali model exposed to the UI is mainly located under the package "model".
At least for all logic related App/Service/Workload, Istio Validations and Istio Config.

jshaughn: The LogOptions is more about the type expected by the business method than any Model used for a return.  I don't think it should be moved to models.  In my opinion the "business" variable convention is the issue, it shadows the package name and seems like bad style.  I actually think the variable name should change or, in this instance we could preserve the variable name and alias the package like: `businesspkg "github.com/kiali/kiali/business"`.

business/workloads.go
```diff
@@ -189,17 +194,36 @@ func (in *WorkloadService) GetPod(namespace, name string) (*models.Pod, error) {
...
+	k8sOpts := opts.PodLogOptions
+	podLog, err := in.k8s.GetPodLogs(namespace, name, &k8sOpts)
 
 	if err != nil {
 		return nil, err
 	}
 
 	lines := strings.Split(podLog.Logs, "\n")
 	messages := make([]LogEntry, 0)
+	lineCount := int64(0)
```
jshaughn, line 207: Maybe I'm missing something but I don't think you want to do this TailLines logic, that starts here, on every call.  In normal cases, when querying to `now` we can just call k8s API normally.   This issue is only when `(sinceTime + duration) < now`.  In that case we can't pass PodLogOptions.TailLines to the k8s API because it will apply the tail to the most recent logs.   Instead, I think we need to:
1) trim logs with time > endTime (i.e. `endTime = sinceTime + duration`)
2) apply TailLines to the remaining logs

jshaughn (changes requested):
A few comments in line.  Also, the `duration` query parameter needs to be documented in doc.go and it should be made clear that `sinceTime + duration` specifies the time range whereas in other API's `queryTime - duration` specifies the time range.

handlers/workloads.go
cfcosta: I like changing the package name, sounds like a good compromise.

business/workloads.go
```diff
@@ -189,17 +194,36 @@ func (in *WorkloadService) GetPod(namespace, name string) (*models.Pod, error) {
...
+
+	var startTime *time.Time
+	if k8sOpts.SinceTime != nil {
+		time := k8sOpts.SinceTime
+
+		if time != nil {
+			startTime = &time.Time
+		}
+	}
```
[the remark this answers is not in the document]
cfcosta: The weirdest thing is that I had something pretty similar to it but was getting weird segfaults, but now that I changed it back it is not erroring out anymore :D

business/workloads.go
cfcosta: Would it fix the case if I just set the k8s param for it to nil before the query, and then apply the tailLines directly on the code? From what I see, we need to parse all lines anyway, so we can just stop parsing then we have enough lines (tailLines if set, then durations).

jshaughn: Yes, I suppose we could do it in all cases, the only cost is the k8s API returning more lines than it really may have to return, but I guess that is fairly minor.  If you prefer that approach I'm OK with it.

business/workloads.go
```diff
@@ -199,7 +208,22 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
...
+
 	for _, line := range lines {
+		if tailLines != nil {
+			if *tailLines <= lineCount {
+				break
+			} else {
+				lineCount += 1
+			}
+		}
+
```
jshaughn, line 226: The logic here is still not right.  I'm OK if we apply the TailLines after the k8s fetch but TailLines=N means return the *last* N lines in the queried Time range, not the first N lines.  As I said in the prior comment, you need to:

1) trim logs with time > endTime (i.e. endTime = sinceTime + duration)
2) apply TailLines to the remaining logs

I'd suggest setting your endTime above and as soon as hit a log entry with timestamp > endTime break out of the loop.  Then run backwards through the entries keeping at most TailLines entries.

business/workloads_test.go
```diff
@@ -456,17 +457,99 @@ func TestGetPodLogs(t *testing.T) {
...
 
-	assert.Equal("2018-01-02T04:34:28+00:00", podLogs.Entries[3].Timestamp)
-	assert.Equal(int64(1514867668), podLogs.Entries[3].TimestampUnix)
+	assert.Equal("2018-01-02T06:34:28+00:00", podLogs.Entries[3].Timestamp)
+	assert.Equal(int64(1514874868), podLogs.Entries[3].TimestampUnix)
 	assert.Equal("error Log Entry With LowerCase Severity", podLogs.Entries[3].Message)
 	assert.Equal("ERROR", podLogs.Entries[3].Severity)
 }
 
+func TestGetPodLogsTailLines(t *testing.T) {
```
jshaughn, line 471: I didn't look at this test but it should not have been passing given the above impl issue.  perhaps its expectations were wrong.

jshaughn (changes requested):

cfcosta:
@jshaughn should be ok now, addressed your concerns (as well as made the tests easier to understand/test stuff).

jshaughn (changes requested):

business/workloads.go
```diff
@@ -225,6 +240,21 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
...
 		parsed, err := time.Parse(time.RFC3339, entry.Timestamp)
 		if err == nil {
+			if startTime == nil {
+				startTime = &parsed
+			}
+
+			if opts.Duration != nil {
+				if endTime == nil {
+					end := parsed.Add(*opts.Duration)
+					endTime = &end
```
cfcosta, line 250: If you know of a better way to do this I'm happy to change. The error is that for some reason I can not do something like this `endTime = &parsed.Add(*opts.Duration)`, it fails on compilation.

jshaughn: It's fine, that's normal for Go as you can't take the address of a funtion return.

lucasponce:
Naming and code organization discussions won't never get a winner.
But I think on this code you are changing a minor pattern and that's triggering this naming issues.

In short, don't bring kubernetes types into the handlers, handlers should work with business and models.

After a look into the code, it seems the pattern you are using is something done similar in other handlers: I understand you are building a criteria from the params and passed to the business layer, that's fine.

If you move that parsing into the business layer probably code will be better organized.

For example, istio_config.go handler (the others are similar):

```
	criteria := business.ParseIstioConfigCriteria(namespace, objects, labelSelector, workloadSelector)

	// Get business layer
	business, err := getBusiness(r)
        [...]
        istioConfig, err := business.IstioConfig.GetIstioConfigList(criteria)
```

You will see the only handler it brought some details from kubernetes into the handler package was this one.

Just take a look, and probably you may do that minor change and apply that.

Renaming all handlers just because the logs are bringing types to the handlers layer doesn't seem a good reason to that.

If after the separation of concerns between layers, you want to rename the package to make more clear change between variable name vs package name is fine, but I'd respect the (intention) with the separation of layers.

Perhaps we are over-commenting, I'm sure if you take a look to the other existing handlers you can even find a better suggestion for that.

cfcosta:
@lucasponce did you check the code after the change proposed by @jshaughn? The idea why I left the types in business is that they are not kubernetes models, but just the type parameter for the argument. I don't need to access the business package to get the kubernetes parameters, just the ones explictly required by the business layer.

If you still think it is a problem, I can just move them to models and be done with it, however there are other places in the code that use the business package (in this handler, I'm sure that others do as well), they just don't conflict because the endpoint is not instantiating the business layer.

lucasponce:
> @lucasponce did you check the code after the change proposed by @jshaughn? The idea why I left the types in business is that they are not kubernetes models, but just the type parameter for the argument. I don't need to access the business package to get the kubernetes parameters, just the ones explictly required by the business layer.

That's fine, I looked into all PR and I think that you are just using a "criteria" pattern.

I don't think the "core and metadata" classes belongs to the handler package, only this handler is using it.

But it's not a blocker, we can refactor this in other iteration.

My only suggestion is that people get familiar with the "abstraction" criteria used on this code to allow write similar code, otherwise it will tend to deviate.

business/workloads.go
```diff
@@ -225,6 +240,21 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *core_v1.P
...
+				}
+
+				if parsed.After(*endTime) {
+					break
+				}
+			}
+
 			entry.TimestampUnix = parsed.Unix()
 		} else {
```
jshaughn, line 259: Not a blocker but the code would flow a little nicer if it were like:

```
if err != nil {
        log.Debugf("Failed to parse log timestamp (skipping) [%s], %s", entry.Timestamp, err.Error())
	continue
}

// the rest here without indentation...
```

jshaughn (approved):
OK, seeing that @lucasponce is OK with a follow-up to handle his comment, I can approve this.

lucasponce:
> OK, seeing that @lucasponce is OK with a follow-up to handle his comment, I can approve this.

Yes, no intention to slow down the work, on the contrary. Just happy to see more people working on these packages.

I'll address some of the refactoring commented in the next work to remove mixer (which is a pure refactoring PR).

Issue #3293, opened 2020-10-06
Use metrics-based comparison / trends for span's duration info


jotak:
Follow-up on #3095

In Trace Page, Span table, next to duration cell: add some comparison value with metrics avg/percentiles over last X min/hour. (E.g. one comparison versus last 30 min, and another versus last 3 hours)
This would apply only to envoy spans. Metric must be selected for same source and same destination as the span.

An informative tooltip needs to be added about that, because it's important that the user understands the bias in this comparison, given the span granularity is higher that the metric's.

lucasponce:
This is good, but we are missing the tracking of this effort.
Please, @jotak would you mind to update the main epic https://github.com/orgs/kiali/projects/19 with this one as a subtask ?

jotak:
FYI I'm doing a first iteration reusing our existing API for metrics.
But given the number of queries perhaps we'll have to introduce some new "batch" endpoint for metrics.

jotak:
So, I'm hesitating with the approach to have, between something more generic or something more opinionated.
The existing API in any case isn't sufficient to deal with the needs here because it's designed to query datapoints in a range, and we want here a single value from a rate interval (like what is done in health & graph edges). Moreover as said above we will probably want to avoid to many front->back queries.

In the details, we want to query stats for:
- several apps and/or workloads (since each span may relate to a different app/workload)
- grouped and not grouped by peer
- with several time intervals (e.g. stats for last 30m and stats for last 3 hours)
- several histogram functions: e.g. avg, p90... (on this point, the current API was fine)

All these points multiply, so that's many queries.
At the moment my preference is for something quite opinionated, with a single grouped "stats" query, through perhaps a new endpoint like `/stats/metrics`, and query params for `apps` / `workloads`, `intervals`, `percentiles` . Perhaps even some POST for that, as the parameters could grow more complex (e.g. what if the apps are across several namespaces? more structured params needed?).

Perhaps that would clear the path for future features like `stats/traces` and so on, that would share the same idea.

@aljesusg , @xeviknal , @lucasponce , does it sound good to you?

jotak:
Note: the new endpoint to be created might eventually be reused by https://github.com/kiali/kiali/issues/2708

jotak:
So, to track some progress here: I have a working draft with back & front:

![Capture d’écran de 2020-10-27 13-53-58](https://user-images.githubusercontent.com/2153442/97304644-4f371580-185c-11eb-87e3-d76ed4744904.png)

[...]

PR #3347, opened 2020-10-22
New metrics stats endpoint

jotak:
- New handler, in/out model
- New MetricsService
- Tests

Fixes https://github.com/kiali/kiali/issues/3293

Note: there's a prior refactoring due to introducing a new Metrics Service. It's in a separate commit (the first one) in order to allow distinct reviews. It includes:
- Moving some istio-related functions from prometheus client to this new metrics service (keep client more low level)
- Unifying usage of the Metrics Handler file: it already exist but was only used for Aggregation-kind metrics. Now other kind of metrics handlers are moved there. It will also hold the new metrics stats handler.
- Creating metrics model file (the model for REST API was previously in prometheus client package; now having the new business layer allows to decouple client // model)
- Moving existing tests to their relevant package

Frontend PR: https://github.com/kiali/kiali-ui/pull/1987


lucasponce:
I can help reviewing this one.
@jotak would you mind to rebase ? I see a minor conflict with master.

routing/routes.go
```diff
@@ -1433,6 +1433,28 @@ func NewRoutes() (r *Routes) {
...
+		//    503: serviceUnavailableError
+		//		500: internalError
+		//		200: metricsStatsResponse
+		{
+			Name:          "MetricsStats",
+			Method:        "POST",
+			Pattern:       "/api/stats/metrics",
+			HandlerFunc:   handlers.MetricsStats,
+			Authenticated: true,
+		},
```
jmazzitelli, line 1457: Note: we now have a molecule test "api-test" that can do basic tests again our API endpoints. If you would like to add a test for this new "api/stats/metrics" that runs nightly, you can add a `stats-metrics.test.yaml` [here](https://github.com/kiali/kiali-operator/tree/master/molecule/api-test). You can look at the other *.test.yaml files in there for examples - we do have a very basic `metrics.test.yaml` already - you can see it [here](https://github.com/kiali/kiali-operator/blob/master/molecule/api-test/metrics.test.yaml).

jotak:
> I can help reviewing this one.
> @jotak would you mind to rebase ? I see a minor conflict with master.

done, rebased

routing/routes.go
jotak: thanks, I will take a look

business/labels_builder.go
```diff
@@ -0,0 +1,142 @@
...
+)
+
+type LabelsBuilder struct {
+	side     string
+	peerSide string
+	protocol string
+	labelsKV []string
+}
+
+func NewLabelsBuilder(direction string) *LabelsBuilder {
```
lucasponce, line 22: I'm not sure what is this.
In the business package is better to give more meaninful naming like "metrics_labels_builder", otherwise it sounds like an aux/helper class.

Note that there are other "helper" similar methods for k8s label management, that's why a "labels_builder" looks odd to me.

lucasponce (commented):
I did a brief review on the code, I'd address the label_builder naming by something with more context i.e. metrics_labels / metrics_labels_builder / ... as labels are also used in kubernetes context.

business/labels_builder.go
jotak: Right, I'll disambiguate by adding "Metrics" in the name here

lucasponce (approved):

hhovsepy (approved):

jotak:
Thanks

Issue #3067, opened 2020-07-31
Add Date/time (start/stop) filtering for logs tab


mtho11:
**Is your feature request related to a problem? Please describe.**
Enable filtering the logs by date/time

**Describe the solution you'd like**
Add a start/stop date/time filters to the logs tab to further filter to a specific time period.

jotak:
Related: https://github.com/kiali/kiali/issues/2147

jshaughn:
> Related: #2147

@mtho11 @jotak , right this is something to think about because I think for correlation purposes the user may want a time range to carry over between tabs, so before doing anything here maybe there should be some disussion about whether this sort of setting should be at a higher level.   I had actually suggested a global time range in the 2147 but it couldbe more restricted.  The views themselves would just limit results to the configured time range.

lucasponce:
I don't know why this one was added into the backlog and the Sprint.

Please, before to progress on this, I'd like that team fix the consistency problem with the controls in the logs:
https://github.com/kiali/kiali/issues/3011

Thanks.

jshaughn:
@lucasponce It's in the sprint because the epic is in the sprint and as per my comment above your comment, we're just discussing it at the moment.  Moreover, you're 6 hours ahead, so give me a chance to read your comments.  I had *just* assigned the bug to Mike and commented that it was the priority, then I read this.  Also, we have a sync meeting in 28 minutes so you could have waited to talk there.

jshaughn:
So, to be clear, at this time we still need to discuss what it means to have time range filtering for logs.  It does NOT currently mean adding timepickers explicitly to the page.  It probably means that this tab and other tabs may all need to be able to handle a custom start time and duration as properties and that that custome range be determined outside of the tab.

mtho11:
@jshaughn got it!

lucasponce:
For this one: please, review the Metrics tabs, that already have a custom datepicker just for those scenarios.

jshaughn:
@mtho11 Given the dificulties of doing this filtering client-side I think we can defer the actual filtering to the pending server-side work in https://github.com/kiali/kiali/issues/3303.  The scope here can be reduced to just supplying the anticipated "duration" query parameter.  Using the "custom" support in TimeRangeComponent hopefully provides a lot of what is needed.

Let me know what you think, we can also continue with the client-side effort although if it's not going to make sprint 46 I think we should punt and instead defer to the backend, so that we can avoid transporting the dead logs over the wire.

[...]

PR #3352, opened 2020-10-23
workload logs: remove support for logs blob

jshaughn:
- leave support only for log entries
- be more careful about sending empty log messages

part of https://github.com/kiali/kiali/issues/3067

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -278,10 +278,10 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 		}
 
 		entries = append(entries, entry)
+	}
 
-		if isBounded && tailLines != nil && len(entries) >= int(*tailLines) {
-			break
-		}
+	if isBounded && tailLines != nil && len(entries) > int(*tailLines) {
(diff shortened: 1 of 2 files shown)


pbajjuri20 (dismissed):

cfcosta (commented):
Looks good. Is there any reason why you did the `isBounded` change? It looks better and more consistent, just curious. Build seems to be broken though.

jshaughn:
> any reason why you did the isBounded change?

@cfcosta Remember when we talked about letting k8s API handle tailLines for current time ranges (i.e. no Duration)?  At the time we thought there would be no penalty to always applying TailLines manually, but I did some testing and in fact letting k8s API apply tailLines was significantly more efficient, especially as the ration of totalLines to tailLines grows.  So, I changed it to only manually perform tailLines when the time range is bounded (i.e Duration is specified and we can't trust k8s to do it).

jshaughn:
@cfcosta or @pbajjuri20 Can one of you re-approve, I lost the approvals when I had to fix a test.

pbajjuri20 (approved):

cfcosta (approved):

Issue #3336, opened 2020-10-21
DR - No labels on subset - Error in list but not in overview


hhovsepy:
Remove label in DR's subset.
In Istio config list of Service details it shows invalid config, but in DR Overview page no error is displayed.
It seems issue is in config list.

![Screenshot from 2020-10-21 11-03-48](https://user-images.githubusercontent.com/604313/96698599-9a00eb00-138d-11eb-8abe-ace159bd6c75.png)
![Screenshot from 2020-10-21 11-03-30](https://user-images.githubusercontent.com/604313/96698606-9bcaae80-138d-11eb-9600-0c1926a4e300.png)

lucasponce:
Yes, there is a missing corner case here.
Basically in Service Details we see the problem due the validations happen in the context of all info fetched for a Service: Service details, Workloads, VirtualServices, Destination Rules.

But in the DestinationRule details a missing labels in the DR destination is not triggering at least a warning.

So I think we can add this warning/error there.

lucasponce:
It's closed but it requires an update in the Docs, so a PR is on the way.

PR #3355, opened 2020-10-26
Add No Labels checker for DR

lucasponce:
Needed for https://github.com/kiali/kiali/issues/3336

### DIFF
--- a/business/checkers/destinationrules/no_dest_checker.go
+++ b/business/checkers/destinationrules/no_dest_checker.go
@@ -59,6 +59,11 @@ func (n NoDestinationChecker) Check() ([]*models.IstioCheck, bool) {
 										valid = false
 									}
 								}
+							} else {
+								validation := models.Build("destinationrules.nodest.subsetnolabels",
+									"spec/subsets["+strconv.Itoa(i)+"]")
+								validations = append(validations, &validation)
+								valid = false
 							}
 						}
 					}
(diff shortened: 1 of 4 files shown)


xeviknal:
I think you might have a Subset in a DestinationRule which has no labels:

The istio.io doc says:

> One or more labels are typically required to identify the subset destination, however, when the corresponding DestinationRule represents a host that supports multiple SNI hosts (e.g., an egress gateway), a subset without labels may be meaningful. In this case a traffic policy with ClientTLSSettings can be used to identify a specific SNI host corresponding to the named subset.

I am not familiar with SNI scenario, and I am wondering if you might have a specific subset that just point to a host but with no different labels but different trafficPolicy. For instance,

```yaml
kind: DestinationRule
metadata:
  name: destination-rule-reviews-test
spec:
  host: reviews
  trafficPolicy:
    loadBalancer:
      simple: RANDOM
  subsets:
  - name: oultlierdetection
    trafficPolicy:
      outlierDetection:
        consecutiveErrors: 7
        interval: 5m
        baseEjectionTime: 15m
```

I don't know if that makes sense though.

lucasponce:
> I don't know if that makes sense though.

One thing could be to move this to a warning, what do you think ?
It sounds very specific to me.

Perhaps we can put a warning and see how it goes.

hhovsepy:
> Perhaps we can put a warning and see how it goes.

+1 on warning

xeviknal (approved):
LGTM!

xeviknal (approved):
LGTM

xeviknal:
@lucasponce forgot to mention that this PR would need an update on the validation documentation page.

hhovsepy (approved):
Verified in a scope of UI PR.

lucasponce:
> @lucasponce forgot to mention that this PR would need an update on the validation documentation page.

Right, I'm going to merge this one and prepare the doc's PR.

PR #3358, opened 2020-10-27
Workload Logs: Avoid business package name shadowing by using criterias

cfcosta:
This is a followup to #3331, using criterias to generate the options so we don't have to access the underlying types directly, and thus avoiding the `business` package to conflict with the `business` variable we use for the business layer.

### DIFF
--- a/handlers/workloads.go
+++ b/handlers/workloads.go
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/gorilla/mux"
-	core_v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	businesspkg "github.com/kiali/kiali/business"
@@ -187,7 +186,7 @@ func PodLogs(w http.ResponseWriter, r *http.Request) {
 	pod := vars["pod"]
 
 	// Get log options
-	opts := businesspkg.LogOptions{PodLogOptions: core_v1.PodLogOptions{Timestamps: true}}
+	opts := business.Workload.BuildLogOptionsCriteria()
 	if container := queryParams.Get("container"); container != "" {
 		opts.Container = container
 	}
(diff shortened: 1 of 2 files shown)


handlers/workloads.go
```diff
@@ -8,7 +8,6 @@ import (
 	"time"
 
 	"github.com/gorilla/mux"
-	core_v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	businesspkg "github.com/kiali/kiali/business"
```
jshaughn, line 13: We can remove this prefix.

jshaughn (changes requested):
@cfcosta One small change andthen I can test it out.

handlers/workloads.go
cfcosta: done

jshaughn (dismissed):
LGTM, I think this should satisfy Lucas' want.

handlers/workloads.go
```diff
@@ -187,7 +186,7 @@ func PodLogs(w http.ResponseWriter, r *http.Request) {
 	pod := vars["pod"]
 
 	// Get log options
-	opts := businesspkg.LogOptions{PodLogOptions: core_v1.PodLogOptions{Timestamps: true}}
+	opts := business.Workload.BuildLogOptionsCriteria()
```
lucasponce, line 189: This is good.
But I wonder why wouldn't you like to move the parsing into this method instead to use it just like an initialization.

Look this example in the other handler:

```
criteria := business.ParseIstioConfigCriteria(namespace, objects, labelSelector, workloadSelector)
```

The paremeters are passed into this `ParseIstioConfigCriteria` and then the criteria is used for the business method.
If the parse error fails you can send a back a 500 or 400 (depends if you want to notify user that parameters were incorrect).

Benefit of this is that handler doesn't touch business logic, only invoking business hence you can save to bring the meta_v1 package.

It's just style, this code works, but in rest of the handlers that's the style we used.

In any case this solves the previous request (thank you!).

jshaughn (approved):
I think this satisfies what @lucasponce is asking for.   Handler now knows only about request/respone handling and business package understands how to handle/parse the query param values and populate the criteria.

jshaughn:
I tested this and assuming travis also passes, and because this is just a small refactor to a PR already passed by QE, I think it is OK to merge as-is.

cfcosta:
@jshaughn tests are ok (travis did take forever to run though), need an approval :)

jshaughn (approved):
LGTM

Issue #3332, opened 2020-10-20
Allow custom health configuration to be deployed as resources/yaml or annotations


erSitzt:
**Is your feature request related to a problem? Please describe.**

Currently custom health configuration has to be done in the kiali operator CRD or via Helm `--set` or in the configmap, but custom error rates are much more dependent on the behavior of the services that get deployed and i think it would make more sense to be able to define them there

**Describe the solution you'd like**
I would like to be able to deploy those configs like any other resource or at least be able to use annotations to set these values directly on a service/virtualservice.

Our services display very different behavoir and error-rates so a general config for a whole namespace would not suffice and i think our devs know best how their services are working and what their "error" baseline is.
This way they could specify thresholds that are used when the services are deployed.
The format of the config looks good to me, only the way to apply those config seems counter intuitive for a feature related to the resources that kiali is displaing/graphing

**Describe alternatives you've considered**
My only other alternative would be to do namespace wide custom values and explicit ones for services/workloads that are showing more extreme behavior

**Additional context**
I did not yet implement something like this, so i dont know how much work/complexity it adds to read/validate configs from CRs
And i think it might be a lot to put in annotations... but i have seen good use of annotation for other tools, so this was the first think that came to mind :)

I would be happy if this could be considered for an already great tool :)

lucasponce:
Thanks for the feedback @erSitzt 

When we started the Health Config feature [1] [2] we know that probably the way config is defined may evolve.

I like the proposal to use annotations rather than custom CRDs for this.

Perhaps CRDs would be more canonical, but adding new CRDs to maintain increases complexity in other places, so having the possibility to use annotation (in Services and/or Workloads) may be a good point to explore.

I'll also cc @aljesusg which is one of the original authors of this work as well.

[1] https://kiali.io/documentation/latest/features/#_health_configuration
[2] https://kiali.io/documentation/latest/configuration/health/

jotak:
+1 for both the feature request (decentralized health config), and @lucasponce 's comment that CRD adds complexity in the installation / update processes
Let's just try to not make it too complex, and make it easy to reason about (ideally I would expect regex in config only when it's necessary).

jmazzitelli:
+1 on CRDs make things more complicated and +1 on trying annotations first.

My contribution to this thread would be to suggest we keep the current way we do it (with the config) but allow annotations to OVERRIDE the config (hopefully that would be possible).

[...]

PR #3362, opened 2020-10-28
Allow custom health configuration to be deployed as annotations

aljesusg:
Fix https://github.com/kiali/kiali/issues/3332
Fix https://github.com/kiali/kiali/issues/3435

# Support Health Config Annotations

- [X] Service
- [X] Workload

# Graph side
- [X] Service
- [x] Workload

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -22,6 +22,9 @@ type HealthService struct {
 	businessLayer *Layer
 }
 
+// Annotation Filter for Health
+var HealthAnnotation = []models.AnotationKey{models.RateHealthAnnotation}
+
 // GetServiceHealth returns a service health (service request error rate)
 func (in *HealthService) GetServiceHealth(namespace, service, rateInterval string, queryTime time.Time) (models.ServiceHealth, error) {
 	var err error
@@ -188,6 +191,7 @@ func (in *HealthService) getNamespaceServiceHealth(namespace string, services []
 	// Prepare all data (note that it's important to provide data for all services, even those which may not have any health, for overview cards)
 	for _, service := range services {
 		h := models.EmptyServiceHealth()
+		h.Requests.HealthAnnotations = models.GetHealthAnnotation(service.Annotations, HealthAnnotation).(map[string]string)
 		allHealth[service.Name] = &h
 	}
 
@@ -228,6 +232,7 @@ func (in *HealthService) getNamespaceWorkloadHealth(namespace string, ws models.
 	allHealth := make(models.NamespaceWorkloadHealth)
 	for _, w := range ws {
 		allHealth[w.Name] = models.EmptyWorkloadHealth()
+		allHealth[w.Name].Requests.HealthAnnotations = models.GetHealthAnnotation(w.HealthAnnotations, HealthAnnotation).(map[string]string)
 		allHealth[w.Name].WorkloadStatus = w.CastWorkloadStatus()
 		if w.IstioSidecar {
 			hasSidecar = true
@@ -288,9 +293,17 @@ func fillWorkloadRequestRates(allHealth models.NamespaceWorkloadHealth, rates mo
 func (in *HealthService) getServiceRequestsHealth(namespace, service, rateInterval string, queryTime time.Time) (models.RequestHealth, error) {
 	rqHealth := models.NewEmptyRequestHealth()
 	inbound, err := in.prom.GetServiceRequestRates(namespace, service, rateInterval, queryTime)
+	if err != nil {
+		return rqHealth, err
+	}
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
 	}
+	svc, err := in.businessLayer.Svc.getService(namespace, service)
+	if err != nil {
+		return rqHealth, err
+	}
+	rqHealth.HealthAnnotations = models.GetHealthAnnotation(svc.Annotations, HealthAnnotation).(map[string]string)
 	rqHealth.CombineReporters()
 	return rqHealth, err
 }
@@ -311,11 +324,21 @@ func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval strin
 func (in *HealthService) getWorkloadRequestsHealth(namespace, workload, rateInterval string, queryTime time.Time) (models.RequestHealth, error) {
 	rqHealth := models.NewEmptyRequestHealth()
 	inbound, outbound, err := in.prom.GetWorkloadRequestRates(namespace, workload, rateInterval, queryTime)
+	if err != nil {
+		return rqHealth, err
+	}
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
 	}
 	for _, sample := range outbound {
 		rqHealth.AggregateOutbound(sample)
+	}
+	w, err := in.businessLayer.Workload.GetWorkload(namespace, workload, "", false)
+	if err != nil {
+		return rqHealth, err
+	}
+	if len(w.Pods) > 0 {
+		rqHealth.HealthAnnotations = models.GetHealthAnnotation(w.HealthAnnotations, HealthAnnotation).(map[string]string)
 	}
 	rqHealth.CombineReporters()
 	return rqHealth, err
--- a/business/health.go
+++ b/business/health.go
@@ -191,7 +191,7 @@ func (in *HealthService) getNamespaceServiceHealth(namespace string, services []
 	// Prepare all data (note that it's important to provide data for all services, even those which may not have any health, for overview cards)
 	for _, service := range services {
 		h := models.EmptyServiceHealth()
-		h.Requests.HealthAnnotations = models.GetHealthAnnotation(service.Annotations, HealthAnnotation).(map[string]string)
+		h.Requests.HealthAnnotations = models.GetHealthAnnotation(service.Annotations, HealthAnnotation)
 		allHealth[service.Name] = &h
 	}
 
@@ -232,7 +232,7 @@ func (in *HealthService) getNamespaceWorkloadHealth(namespace string, ws models.
 	allHealth := make(models.NamespaceWorkloadHealth)
 	for _, w := range ws {
 		allHealth[w.Name] = models.EmptyWorkloadHealth()
-		allHealth[w.Name].Requests.HealthAnnotations = models.GetHealthAnnotation(w.HealthAnnotations, HealthAnnotation).(map[string]string)
+		allHealth[w.Name].Requests.HealthAnnotations = models.GetHealthAnnotation(w.HealthAnnotations, HealthAnnotation)
 		allHealth[w.Name].WorkloadStatus = w.CastWorkloadStatus()
 		if w.IstioSidecar {
 			hasSidecar = true
@@ -303,7 +303,7 @@ func (in *HealthService) getServiceRequestsHealth(namespace, service, rateInterv
 	if err != nil {
 		return rqHealth, err
 	}
-	rqHealth.HealthAnnotations = models.GetHealthAnnotation(svc.Annotations, HealthAnnotation).(map[string]string)
+	rqHealth.HealthAnnotations = models.GetHealthAnnotation(svc.Annotations, HealthAnnotation)
 	rqHealth.CombineReporters()
 	return rqHealth, err
 }
@@ -338,7 +338,7 @@ func (in *HealthService) getWorkloadRequestsHealth(namespace, workload, rateInte
 		return rqHealth, err
 	}
 	if len(w.Pods) > 0 {
-		rqHealth.HealthAnnotations = models.GetHealthAnnotation(w.HealthAnnotations, HealthAnnotation).(map[string]string)
+		rqHealth.HealthAnnotations = models.GetHealthAnnotation(w.HealthAnnotations, HealthAnnotation)
 	}
 	rqHealth.CombineReporters()
 	return rqHealth, err
--- a/business/health.go
+++ b/business/health.go
@@ -23,7 +23,7 @@ type HealthService struct {
 }
 
 // Annotation Filter for Health
-var HealthAnnotation = []models.AnotationKey{models.RateHealthAnnotation}
+var HealthAnnotation = []models.AnnotationKey{models.RateHealthAnnotation}
 
 // GetServiceHealth returns a service health (service request error rate)
 func (in *HealthService) GetServiceHealth(namespace, service, rateInterval string, queryTime time.Time) (models.ServiceHealth, error) {
--- a/business/services.go
+++ b/business/services.go
@@ -123,7 +123,7 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 			IstioSidecar:           hasSidecar,
 			AppLabel:               appLabel,
 			AdditionalDetailSample: models.GetFirstAdditionalIcon(conf, item.ObjectMeta.Annotations),
-			HealthAnnotations:      models.GetHealthAnnotation(item.Annotations, models.GetHealthConfigAnnotation()).(map[string]string),
+			HealthAnnotations:      models.GetHealthAnnotation(item.Annotations, models.GetHealthConfigAnnotation()),
 			Labels:                 item.Labels,
 		}
 	}
--- a/business/services.go
+++ b/business/services.go
@@ -123,7 +123,7 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 			IstioSidecar:           hasSidecar,
 			AppLabel:               appLabel,
 			AdditionalDetailSample: models.GetFirstAdditionalIcon(conf, item.ObjectMeta.Annotations),
-			KialiAnnotations:       models.GetHealthAnnotation(item.Annotations, models.GetHealthConfigAnnotation()).(map[string]string),
+			HealthAnnotations:      models.GetHealthAnnotation(item.Annotations, models.GetHealthConfigAnnotation()).(map[string]string),
 			Labels:                 item.Labels,
 		}
 	}
--- a/business/services.go
+++ b/business/services.go
@@ -123,6 +123,7 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 			IstioSidecar:           hasSidecar,
 			AppLabel:               appLabel,
 			AdditionalDetailSample: models.GetFirstAdditionalIcon(conf, item.ObjectMeta.Annotations),
+			KialiAnnotations:       models.GetHealthAnnotation(item.Annotations, models.GetHealthConfigAnnotation()).(map[string]string),
 			Labels:                 item.Labels,
 		}
 	}
--- a/business/services.go
+++ b/business/services.go
@@ -276,6 +276,26 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime 
 	return &s, nil
 }
 
+func (in *SvcService) UpdateService(namespace, service string, interval string, queryTime time.Time, jsonPatch string) (*models.ServiceDetails, error) {
+	var err error
+	promtimer := internalmetrics.GetGoFunctionMetric("business", "SvcService", "GetService")
+	defer promtimer.ObserveNow(&err)
+
+	// Identify controller and apply patch to workload
+	err = updateService(in.businessLayer, namespace, service, jsonPatch)
+	if err != nil {
+		return nil, err
+	}
+
+	// Cache is stopped after a Create/Update/Delete operation to force a refresh
+	if kialiCache != nil && err == nil {
+		kialiCache.RefreshNamespace(namespace)
+	}
+
+	// After the update we fetch the whole workload
+	return in.GetService(namespace, service, interval, queryTime)
+}
+
 // GetServiceDefinition returns a single service definition (the service object and endpoints), no istio or runtime information
 func (in *SvcService) GetServiceDefinition(namespace, service string) (*models.ServiceDetails, error) {
 	var err error
@@ -412,4 +432,14 @@ func (in *SvcService) GetServiceAppName(namespace, service string) (string, erro
 	appLabelName := config.Get().IstioLabels.AppLabelName
 	app := svc.Spec.Selector[appLabelName]
 	return app, nil
+}
+
+func updateService(layer *Layer, namespace string, service string, jsonPatch string) error {
+	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
+	// if namespace is accessible from Kiali (Deployment.AccessibleNamespaces)
+	if _, err := layer.Namespace.GetNamespace(namespace); err != nil {
+		return err
+	}
+
+	return layer.k8s.UpdateService(namespace, service, jsonPatch)
 }
--- a/graph/config/cytoscape/cytoscape.go
+++ b/graph/config/cytoscape/cytoscape.go
@@ -54,6 +54,9 @@ type ProtocolTraffic struct {
 	Responses Responses         `json:"responses,omitempty"` // see comment above
 }
 
+// HealthConfig maps annotations information for health
+type HealthConfig map[string]string
+
 type NodeData struct {
 	// Cytoscape Fields
 	ID     string `json:"id"`               // unique internal node ID (n0, n1...)
@@ -71,6 +74,7 @@ type NodeData struct {
 	DestServices    []graph.ServiceName `json:"destServices,omitempty"`    // requested services for [dest] node
 	Traffic         []ProtocolTraffic   `json:"traffic,omitempty"`         // traffic rates for all detected protocols
 	HasCB           bool                `json:"hasCB,omitempty"`           // true (has circuit breaker) | false
+	HasHealthConfig HealthConfig        `json:"HasHealthConfig,omitempty"` // set to the health config override
 	HasMissingSC    bool                `json:"hasMissingSC,omitempty"`    // true (has missing sidecar) | false
 	HasVS           bool                `json:"hasVS,omitempty"`           // true (has route rule) | false
 	IsBox           string              `json:"isBox,omitempty"`           // set for NodeTypeBox, current values: [ 'app', 'cluster', 'namespace' ]
@@ -213,6 +217,11 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
 		}
 
 		addNodeTelemetry(n, nd)
+
+		// set annotations, if available
+		if val, ok := n.Metadata[graph.HasHealthConfig]; ok {
+			nd.HasHealthConfig = val.(map[string]string)
+		}
 
 		// node may have deployment but no pods running)
 		if val, ok := n.Metadata[graph.IsDead]; ok {
--- a/graph/config/cytoscape/cytoscape.go
+++ b/graph/config/cytoscape/cytoscape.go
@@ -74,7 +74,7 @@ type NodeData struct {
 	DestServices    []graph.ServiceName `json:"destServices,omitempty"`    // requested services for [dest] node
 	Traffic         []ProtocolTraffic   `json:"traffic,omitempty"`         // traffic rates for all detected protocols
 	HasCB           bool                `json:"hasCB,omitempty"`           // true (has circuit breaker) | false
-	HasHealthConfig HealthConfig        `json:"HasHealthConfig,omitempty"` // set to the health config override
+	HasHealthConfig HealthConfig        `json:"hasHealthConfig,omitempty"` // set to the health config override
 	HasMissingSC    bool                `json:"hasMissingSC,omitempty"`    // true (has missing sidecar) | false
 	HasVS           bool                `json:"hasVS,omitempty"`           // true (has route rule) | false
 	IsBox           string              `json:"isBox,omitempty"`           // set for NodeTypeBox, current values: [ 'app', 'cluster', 'namespace' ]
--- a/graph/meta.go
+++ b/graph/meta.go
@@ -18,6 +18,7 @@ const (
 	DestPrincipal   MetadataKey = "destPrincipal"
 	DestServices    MetadataKey = "destServices"
 	HasCB           MetadataKey = "hasCB"
+	HasHealthConfig MetadataKey = "hasHealthConfig"
 	HasMissingSC    MetadataKey = "hasMissingSC"
 	HasVS           MetadataKey = "hasVS"
 	IsDead          MetadataKey = "isDead"
--- a/graph/telemetry/istio/appender/appender.go
+++ b/graph/telemetry/istio/appender/appender.go
@@ -25,6 +25,8 @@ func ParseAppenders(o graph.TelemetryOptions) []graph.Appender {
 				requestedAppenders[AggregateNodeAppenderName] = true
 			case DeadNodeAppenderName:
 				requestedAppenders[DeadNodeAppenderName] = true
+			case HealthConfigAppenderName:
+				requestedAppenders[HealthConfigAppenderName] = true
 			case IdleNodeAppenderName:
 				requestedAppenders[IdleNodeAppenderName] = true
 			case IstioAppenderName:
@@ -110,6 +112,10 @@ func ParseAppenders(o graph.TelemetryOptions) []graph.Appender {
 		}
 		appenders = append(appenders, a)
 	}
+	if _, ok := requestedAppenders[HealthConfigAppenderName]; ok || o.Appenders.All {
+		a := HealthConfigAppender{}
+		appenders = append(appenders, a)
+	}
 	if _, ok := requestedAppenders[IdleNodeAppenderName]; ok || o.Appenders.All {
 		hasNodeOptions := o.App != "" || o.Workload != "" || o.Service != ""
 		a := IdleNodeAppender{
@@ -170,6 +176,18 @@ func getServiceDefinitionList(ni *graph.AppenderNamespaceInfo) *models.ServiceDe
 		return sdl.(*models.ServiceDefinitionList)
 	}
 	return nil
+}
+
+func getService(serviceName string, ni *graph.AppenderNamespaceInfo) (*models.Service, bool) {
+	if serviceName == "" || serviceName == graph.Unknown {
+		return nil, false
+	}
+	for _, srv := range getServiceDefinitionList(ni).ServiceDefinitions {
+		if srv.Service.Name == serviceName {
+			return &srv.Service, true
+		}
+	}
+	return nil, false
 }
 
 func getServiceEntryHosts(gi *graph.AppenderGlobalInfo) (serviceEntryHosts, bool) {
--- a/graph/telemetry/istio/appender/health_config.go
+++ b/graph/telemetry/istio/appender/health_config.go
@@ -7,8 +7,7 @@ import (
 
 const HealthConfigAppenderName = "healthConfig"
 
-// HealthConfigAppenderis responsible for adding healtch configuration annotion to the graph.
-// the appender currently add only mutual_tls security although is written in a generic way.
+// HealthConfigAppenderis responsible for adding health configuration annotation to the graph.
 // Name: healthConfig
 type HealthConfigAppender struct{}
 
--- a/graph/telemetry/istio/appender/health_config.go
+++ b/graph/telemetry/istio/appender/health_config.go
@@ -1,0 +1,60 @@
+package appender
+
+import (
+	"github.com/kiali/kiali/graph"
+	"github.com/kiali/kiali/models"
+)
+
+const HealthConfigAppenderName = "healthConfig"
+
+// HealthConfigAppenderis responsible for adding healtch configuration annotion to the graph.
+// the appender currently add only mutual_tls security although is written in a generic way.
+// Name: healthConfig
+type HealthConfigAppender struct{}
+
+// Name implements Appender
+func (a HealthConfigAppender) Name() string {
+	return HealthConfigAppenderName
+}
+
+// AppendGraph implements Appender
+func (a HealthConfigAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo, namespaceInfo *graph.AppenderNamespaceInfo) {
+	if len(trafficMap) == 0 {
+		return
+	}
+
+	if getWorkloadList(namespaceInfo) == nil {
+		sdl, err := globalInfo.Business.Workload.GetWorkloadList(namespaceInfo.Namespace)
+		graph.CheckError(err)
+		namespaceInfo.Vendor[workloadListKey] = &sdl
+	}
+
+	if getServiceDefinitionList(namespaceInfo) == nil {
+		sdl, err := globalInfo.Business.Svc.GetServiceDefinitionList(namespaceInfo.Namespace)
+		graph.CheckError(err)
+		namespaceInfo.Vendor[serviceDefinitionListKey] = sdl
+	}
+
+	a.applyHealthConfigPresence(trafficMap, namespaceInfo)
+}
+
+func (a *HealthConfigAppender) applyHealthConfigPresence(trafficMap graph.TrafficMap, namespaceInfo *graph.AppenderNamespaceInfo) {
+	for _, n := range trafficMap {
+		if n.Namespace != namespaceInfo.Namespace {
+			continue
+		}
+		// get the workload for the node and check to see if they have health configuration.
+		switch n.NodeType {
+		case graph.NodeTypeWorkload:
+			if workload, found := getWorkload(n.Workload, namespaceInfo); found {
+				n.Metadata[graph.HasHealthConfig] = models.GetHealthAnnotation(workload.HealthAnnotations, models.GetHealthConfigAnnotation())
+			}
+		case graph.NodeTypeService:
+			if srv, found := getService(n.Service, namespaceInfo); found {
+				n.Metadata[graph.HasHealthConfig] = models.GetHealthAnnotation(srv.HealthAnnotations, models.GetHealthConfigAnnotation())
+			}
+		default:
+			continue
+		}
+	}
+}
--- a/models/health.go
+++ b/models/health.go
@@ -30,6 +30,7 @@ func NewEmptyRequestHealth() RequestHealth {
 	return RequestHealth{
 		Inbound:            make(map[string]map[string]float64),
 		Outbound:           make(map[string]map[string]float64),
+		HealthAnnotations:  make(map[string]string),
 		inboundSource:      make(map[string]map[string]float64),
 		inboundDestination: make(map[string]map[string]float64),
 	}
@@ -96,6 +97,7 @@ type ProxyStatus struct {
 type RequestHealth struct {
 	Inbound            map[string]map[string]float64 `json:"inbound"`
 	Outbound           map[string]map[string]float64 `json:"outbound"`
+	HealthAnnotations  map[string]string             `json:"healthAnnotations"`
 	inboundSource      map[string]map[string]float64
 	inboundDestination map[string]map[string]float64
 }
--- a/models/health_config.go
+++ b/models/health_config.go
@@ -5,7 +5,7 @@ type AnotationKey string
 
 const (
 	AllHealthAnnotation  AnotationKey = ".*"
-		RateHealthAnnotation AnotationKey = "health.kiali.io/rate"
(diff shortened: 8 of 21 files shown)


aljesusg:
Some concerns.

# Workload case

I need to check all the pods start with workload name requested. After that check all annotations (could be different, example : manual update and have a reviews-v1-323232 and reviews-v1-4343 with different annotation) What to do?

# App case

I think that we should remove this improvement for this kind resource

# Cache

I saw the cache work for namespaces. I was thinking about to store for a specific resource the configuration to avoid to fetch it each time. Is the cache refreshing when annotations change in kubernetes?

cc @lucasponce @jotak

jotak:
> Some concerns.
> Workload case
> 
> I need to check all the pods start with workload name requested.

why checking pod name and workload name? why not workload-versus-workload? (see other point below maybe it's related)

> After that check all annotations (could be different, example : manual update and have a reviews-v1-323232 and reviews-v1-4343 with different annotation) What to do?

I think we can assume pods within the same workload share the same annotations. Actually I think we don't have other choice than to assume that (we won't have health per pod anyway). So, I would say just picking just the first pod is fine :)

> App case
> 
> I think that we should remove this improvement for this kind resource

I see why you say that... the annotation could only be set a the workload level, but the health that we fetch for apps are all merged into a single RequestHealth object. Perhaps we should change the model here, to allow per-workload config within app health, but it's fine to postpone that for later, one step after another.

> Cache
> 
> I saw the cache work for namespaces. I was thinking about to store for a specific resource the configuration to avoid to fetch it each time. Is the cache refreshing when annotations change in kubernetes?
> 
> cc @lucasponce @jotak

aljesusg:
> > Some concerns.
> > Workload case
> > I need to check all the pods start with workload name requested.
> 
> why checking pod name and workload name? why not workload-versus-workload? (see other point below maybe it's related)
> 
> > After that check all annotations (could be different, example : manual update and have a reviews-v1-323232 and reviews-v1-4343 with different annotation) What to do?
> 
> I think we can assume pods within the same workload share the same annotations. Actually I think we don't have other choice than to assume that (we won't have health per pod anyway). So, I would say just picking just the first pod is fine :)

Yes...there is not another way that assume that...I hope that kubernetes will update it for all the pods

> 
> > App case
> > I think that we should remove this improvement for this kind resource
> 
> I see why you say that... the annotation could only be set a the workload level, but the health that we fetch for apps are all merged into a single RequestHealth object. Perhaps we should change the model here, to allow per-workload config within app health, but it's fine to postpone that for later, one step after another.

Yes..I'll start adding this feature to service and workload and continue in another step with app.

> 
> > Cache
> > I saw the cache work for namespaces. I was thinking about to store for a specific resource the configuration to avoid to fetch it each time. Is the cache refreshing when annotations change in kubernetes?
> > cc @lucasponce @jotak

lucasponce:
> I need to check all the pods start with workload name requested. After that check all annotations (could be different, example : manual update and have a reviews-v1-323232 and reviews-v1-4343 with different annotation) What to do?

I'd check only the workload / controller template (or pod in the corner case that there is no controller).

> I think that we should remove this improvement for this kind resource

It's fine, also, Application doesn't exist as a strong entity, so, it's fine to not consider it in a first step. If we cover the Workload/Services sounds a good progress. We can see on testing if there is some clash between workload and app.

lucasponce:
@aljesusg I guess it's still in active development and perhaps it's earlier to jump into the code or doing more testing.
Just ping when you think it's ready to further review or local testing and we can help there.

aljesusg:
Ok I added a new appender to return the annotations for workloads graph type in first step, I am going to check the changes in the UI and we can think about to add services/apps

aljesusg:
Ok @lucasponce now we can see anntoations in graph endpoint. I am filtering the annotations to return only rates

Annotation
```json
{"kiali": {"health_config": {"rate":[{"code": "^4\\d\\d$", "protocol": "http",
      "degraded": 5, "failure": 10}]}}
```

Example node
```json
{
        "data": {
          "id": "e8a4c5a8a5a937ec63d1da940d4b68a1",
          "nodeType": "service",
          "namespace": "bookinfo",
          "app": "reviews",
          "service": "reviews",
          "annotation": [
            {
              "code": "^4\\d\\d$",
              "degraded": 5,
              "failure": 10,
              "protocol": "http"
            }
          ],
          "destServices": [
            {
              "namespace": "bookinfo",
              "name": "reviews"
            }
          ]
        }
      }
```

I don't know if we should call **kialiAnnotation** instead **annotation** in the return object

aljesusg:
# How to test this 

We can check for example in service reviews that the health is green.

We are going to force this having a tolerance for all codes

```bash
kubectl annotate --overwrite svc reviews kiali="{\"health_config\":{\"rate\": [{\"code\": \"^.*$\", \"failure\": 20, \"degraded\": 11, \"protocol\": \"http\"}]}}" -n bookinfo
```
## Screenshot
Screenshot from 2020-11-13 09-36-37

aljesusg:
# Fast example

Deploy error-rates with this config in operator

```bash
{
  "healthConfig": {
    "rate": [
      {
        "namespace": "/alpha/",
        "kind": "/app/",
        "name": "/x-server/",
        "tolerance": [
          {
            "code": "/[45]\\d[^\\D4]/",
            "degraded": 0,
            "failure": 10,
            "protocol": "/http/",
            "direction": "/.*/"
          }
        ]
      }    
    ]
  }
}
```

We'll see the next picture in the graph
Screenshot from 2020-11-16 10-19-05
Screenshot from 2020-11-16 10-33-19


Where X-server is in green status but we can check that we have 9.0% of 400 error code

Screenshot from 2020-11-16 10-20-02

We want to show this error for x-server when the error rate is >5%, so we need to add the next annotation 

kubectl annotate --overwrite svc x-server kiali="{\"health_config\":{\"rate\": [{\"code\": \"^4\\\d\\\d$\", \"failure\": 20, \"degraded\": 5, \"protocol\": \"http\"}]}}" -n alpha


and now we can check the graph with this degraded status

Screenshot from 2020-11-16 10-32-47
Screenshot from 2020-11-16 10-32-57

hhovsepy:
Have increased the degraded rates for service and workload to 10%, but still it shows warning icon for both:
kubectl annotate --overwrite svc y-server kiali="{"health_config":{"rate": [{"code": "^4\\\d\\\d$", "failure": 20, "degraded": 10, "protocol": "http"}]}}" -n alpha
kubectl annotate --overwrite deployment b-client kiali="{"health_config":{"rate": [{"code": "^4\\\d\\\d$", "failure": 20, "degraded": 10, "protocol": "http"}]}}" -n beta

Screenshot from 2020-11-20 09-08-39
Screenshot from 2020-11-20 09-09-31

aljesusg:
> Have increased the degraded rates for service and workload to 10%, but still it shows warning icon for both:
> kubectl annotate --overwrite svc y-server kiali="{"health_config":{"rate": [{"code": "^4\\d\\d$", "failure": 20, "degraded": 10, "protocol": "http"}]}}" -n alpha
> kubectl annotate --overwrite deployment b-client kiali="{"health_config":{"rate": [{"code": "^4\\d\\d$", "failure": 20, "degraded": 10, "protocol": "http"}]}}" -n beta
> 
> Screenshot from 2020-11-20 09-08-39
> Screenshot from 2020-11-20 09-09-31

The problem is that in that PR we are not throwing errors, we are working in another PR.

So You can check if the config is correct if the health endpoint return that config. In that case /api/namespaces/alpha/services/y-server/health not return the configuration. If you go to /api/namespaces/alpha/services/y-server you can check the kiali annotation is 
```json
{health_config:{rate: [{code: ^4\\d\\d, failure: 20, degraded: 10, protocol: http}]}}
```
and should be
```json
{\"health_config\":{\"rate\": [{\"code\": \"^4\\d\\d\", \"failure\": 20, \"degraded\": 10, \"protocol\": \"http\"}]}}
```


The cli command should be
```bash
kubectl annotate --overwrite svc x-server kiali="{\"health_config\": {\"rate\": [{\"code\": \"^4\\\d\\\d$\", \"degraded\": 10, \"failure\": 20, \"protocol\": \"http\"}]}}" -n alpha
```

hhovsepy (dismissed):
Thanks @aljesusg verified now.

lucasponce:
Hi @aljesusg would you mind to rebase this and the UI one ?

I see the branch is a little bit old
image

I'm going to start to test & review it.

lucasponce:
We have discussed a couple of suggestions on this feature:
- Adjust the annotation format in a simplified structure.
- Introduce some "sugar" syntax for the code.

Inspired in [1] this annotation could be something named like:

```
health.kiali.io/rate
```

In that way, this syntax allows to extend and add more categories in the future and it's sync-ed with istio ones.

Also, the tolerance rate is defined with the following structure:

```
	Code      string  `yaml:"code,omitempty" json:"code"`
	Degraded  float32 `yaml:"degraded,omitempty" json:"degraded"`
	Failure   float32 `yaml:"failure,omitempty" json:"failure"`
	Protocol  string  `yaml:"protocol,omitempty" json:"protocol"`
	Direction string  `yaml:"direction,omitempty" json:"direction"`
```

In a configuration for the namespace is used like:

```
    - namespace: "alpha"
      tolerance:
        - code: "404"
          failure: 10
          protocol: "http"
        - code: "[45]\\d[^4]"
          protocol: "http"
```

But this tolerance could be represented in an annotation for a service / workload like:

```
health.kiali.io/rate=404,0,10;[45]x[^4],10,20
```

Which is better readable for a user.

Two details:
- As it happens with the global config, when a field is not present it takes the default value.
- For **code** regex we expect to have only numeric values, we could introduce a "sugar syntax" with x or X to represent `\d`, we'll reduce the number of parsing problems and also it's readable for the user expression like 4xx or 5xx (or 4XX, 5XX) in this context.

This "sugar syntax" could be applied to the general config as well:

```
    - namespace: "alpha"
      tolerance:
        - code: "404"
          failure: 10
          protocol: "http"
        - code: "[45]X[^4]"
          protocol: "http"
```

I think this is going to trigger some refactoring in the current PR but it shouldn't change the result of the feature.


[1] https://istio.io/latest/docs/reference/config/annotations/

aljesusg:
I did the changes in backend and UI side @lucasponce. I'm going to continue working in the UI.

Another thing....For workload we are fetching the annotations of the first pod....maybe we should fetch the deployment instead this? What do you think @jotak @lucasponce ?

lucasponce:
> Another thing....For workload we are fetching the annotations of the first pod....maybe we should fetch the deployment instead this? What do you think @jotak @lucasponce ?

Yes, I have that comment in my list for the review, but as we were discussing more the changes in the feature I didn't want to mix all feature + low level code reviews.

I'll resume the review and I'll add more comments.

aljesusg:
> > Another thing....For workload we are fetching the annotations of the first pod....maybe we should fetch the deployment instead this? What do you think @jotak @lucasponce ?
> 
> Yes, I have that comment in my list for the review, but as we were discussing more the changes in the feature I didn't want to mix all feature + low level code reviews.
> 
> I'll resume the review and I'll add more comments.

I thought that deployment change propagate this changes....:(

models/workload.go
```diff
@@ -127,6 +131,10 @@ func (workload *WorkloadListItem) ParseWorkload(w *Workload) {
 	workload.PodCount = len(w.Pods)
 	workload.AdditionalDetailSample = w.AdditionalDetailSample
 
+	if len(w.Pods) > 0 {
```
lucasponce, line 134: I'd expect the annotation on the workload, no need to get it from the Pod.
Just place the annotation inside the `parseObjectMeta` and that's good.
Note, that all workload annotations manipulation are located there.

models/health.go
```diff
@@ -1,9 +1,14 @@
 package models
 
 import (
+	"fmt"
+
 	"github.com/prometheus/common/model"
 )
 
+const HealthConfiguration = "health_config"
```
lucasponce, line 9: Unused

models/health.go
```diff
@@ -1,9 +1,14 @@
...
+const HealthConfigurationRates = "rate"
```
lucasponce, line 10: Unused

models/health.go
```diff
@@ -157,3 +163,10 @@ func (ps ProxyStatus) IsSynced() bool {
 func isComponentStatusSynced(componentStatus string) bool {
 	return componentStatus == "Synced"
 }
+
+func GetHealthConfiguration(annotations map[string]string) (string, error) {
+	if kialiConf, ok := annotations["health.kiali.io/rate"]; ok {
```
lucasponce, line 168: Naming is hard and opnionated, just thinking in future maintenance, to me it would be more clear "kialiConfig" -> "healthConfig".
It's a minor comment.

models/health.go
```diff
@@ -89,6 +94,7 @@ type ProxyStatus struct {
 type RequestHealth struct {
 	Inbound  map[string]map[string]float64 `json:"inbound"`
 	Outbound map[string]map[string]float64 `json:"outbound"`
+	Config   string                        `json:"config"`
```
lucasponce, line 97: This should be `omitempty` as this is optional and only used on the case where user annotates a specific Service and/or Workload.

models/health.go
```diff
@@ -157,3 +163,10 @@ func (ps ProxyStatus) IsSynced() bool {
 func isComponentStatusSynced(componentStatus string) bool {
 	return componentStatus == "Synced"
 }
+
+func GetHealthConfiguration(annotations map[string]string) (string, error) {
```
lucasponce, line 167: We reserve the "Get*" verbs for backend calls, for these kind of operations a "Parse*" / "Cast*" fits better because it's just extracting data from one format to another.

models/health.go
```diff
@@ -157,3 +163,10 @@ func (ps ProxyStatus) IsSynced() bool {
...
+	if kialiConf, ok := annotations["health.kiali.io/rate"]; ok {
+		return kialiConf, nil
+	}
+	return "", fmt.Errorf("No Kiali Health configuration in Annotations")
```
lucasponce, line 171: I don't think you should return an error here.
Not every Service would expect annotations.
So, I would send just empty string, then returning error may make no sense on this method (and all propagations).

business/health.go
```diff
@@ -291,9 +293,23 @@ func fillWorkloadRequestRates(allHealth models.NamespaceWorkloadHealth, rates mo
...
 	rqHealth := models.NewEmptyRequestHealth()
 	inbound, err := in.prom.GetServiceRequestRates(namespace, service, rateInterval, queryTime)
+	if err != nil {
+		return rqHealth, err
+	}
 	for _, sample := range inbound {
 		rqHealth.AggregateInbound(sample)
 	}
+
+	svc, err := in.k8s.GetService(namespace, service)
```
lucasponce, line 303: Ok, there is a missing case in the kialiCached that I missed.
This query is ok, but it should be done via cache.
No changes for the moment as this requires more changes.

I'll send a followup PR https://github.com/kiali/kiali/issues/3467 and I will update this later.

business/health.go
```diff
@@ -312,11 +328,26 @@ func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval strin
...
 	}
 	for _, sample := range outbound {
 		rqHealth.AggregateOutbound(sample)
 	}
+
+	w, err := in.businessLayer.Workload.GetWorkload(namespace, workload, "", false)
+	if err != nil {
+		return rqHealth, err
+	}
+	if len(w.Pods) > 0 {
```
lucasponce, line 345: Annotation should be provided by the model.Workload instead to get parsed from the Pod.

business/health.go
```diff
@@ -312,11 +328,26 @@ func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval strin
...
+		var errh error
+		rqHealth.Config, errh = models.GetHealthConfiguration(w.Pods[0].Annotations)
+		if errh != nil {
+			log.Debugf("Error Getting Kiali Health Configuration. %s", errh.Error())
```
lucasponce, line 349: This sounds like a debug, but it's not an error, because the normal scenario is that annotations are not present.

models/health.go
lucasponce: I'm reviewing the appender and looking that code I guess we should call this parameter "healthConfig", also here and in the graph to get consistency.
Rationale is that /health endpoint and /graph fetch the info needed to calculate health: rate request, if "healthConfig" is present in both, then instead to use the global "healthConfig" config it personalizes for this item (service/workload) using this info.

graph/config/cytoscape/cytoscape.go
```diff
@@ -59,26 +59,27 @@ type NodeData struct {
...
+	App              string              `json:"app,omitempty"`
+	Version          string              `json:"version,omitempty"`
+	Service          string              `json:"service,omitempty"`          // requested service for NodeTypeService
+	Aggregate        string              `json:"aggregate,omitempty"`        // set like "<aggregate>=<aggregateVal>"
+	DestServices     []graph.ServiceName `json:"destServices,omitempty"`     // requested services for [dest] node
+	Traffic          []ProtocolTraffic   `json:"traffic,omitempty"`          // traffic rates for all detected protocols
+	HasCB            bool                `json:"hasCB,omitempty"`            // true (has circuit breaker) | false
+	HasMissingSC     bool                `json:"hasMissingSC,omitempty"`     // true (has missing sidecar) | false
+	HasVS            bool                `json:"hasVS,omitempty"`            // true (has route rule) | false
+	HealthAnnotation []interface{}       `json:"healthAnnotation,omitempty"` // requested healthAnnotation for NodeType
```
lucasponce, line 74: I guess "Annotation" shouldn't belong to this domain, I guess it's better to name it "HealthConfig" and I think type should be a `string` if I'm not wrong and I missed something.

graph/meta.go
```diff
@@ -13,6 +13,7 @@ func NewMetadata() Metadata {
 
 // Metadata keys to be used instead of literal strings
 const (
+	Annotation      MetadataKey = "annotation"
```
lucasponce, line 16: I have feeling using "annotation" naming will be confusing, "healthConfig" would help with the semantic.

graph/telemetry/istio/appender/appender.go
```diff
@@ -22,6 +22,8 @@ func ParseAppenders(o graph.TelemetryOptions) []graph.Appender {
 	if !o.Appenders.All {
 		for _, appenderName := range o.Appenders.AppenderNames {
 			switch appenderName {
+			case AnnotationAppenderName:
```
lucasponce, line 25: The previous comment about naming would propagate into these changes as well.

graph/telemetry/istio/appender/health_annotation.go
```diff
@@ -0,0 +1,75 @@
+package appender
+
+import (
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/graph"
+	"github.com/kiali/kiali/models"
+)
+
+const AnnotationAppenderName = "healthAnnotation"
```
lucasponce, line 9: "healthAnnotation" is a little confusing to me, I'd prefer a "healthConfig" naming.
Another consideration is that this appender is pretty close to the sidecar one, and perhaps both could be merged together.

But I'd let @jshaughn to comment here.

lucasponce (changes requested):
Conceptually feature is clear:
- Both sources of info used to calculate health (/graph and /health) can be augmented with a specific "healthConfig" info that will be used to compute health instead of using the global config defined. This optional "healthConfig" can be added as part of a kiali annotation in Workloads and/or Services.

Some requests about code organization and naming.

Also, this PR should wait to https://github.com/kiali/kiali/issues/3467 as it's missing to use the cache for Service Detail in order to not invoke unnecessary calls to the backend.

graph/telemetry/istio/appender/health_annotation.go
aljesusg: I am goign to replace this by the new syntax "health.kiali.io/rate" , if we add more annotations we can change this to "kiali annotations" to return all annotations like kiali.io

models/workload.go
jotak: I don't know, the advantage of checking at pod level is that, in the end, there's always a pod ; whereas deployments are not mandatory, pods can result from various kind of controllers.
Like istio annotations in general are taken from pods, so I thought we could follow the same pattern.

jotak: (see how many istio annotations are read from pods: https://istio.io/latest/docs/reference/config/annotations/ )

aljesusg: In fact I checked that yesterday thanks @jotak.

In my last commit I am filtering by kiali Annotations.

lucasponce: Deployments scaled to 0 are possible, a corner case but possible.
For this particular code, the suggestion is to move it into the parseObjectMeta() method that is performing a similar operation.
In that case, it will cover all cases, Pod included.

graph/telemetry/istio/appender/health_annotation.go
lucasponce: The suggestion is to use a "healthConfiguration" appender instead of "healthAnnotations".
When we cover a more generic case we can reconsider, but "healthAnnotation" is a little bit confusing.

business/health.go
lucasponce: Note, that PR #3467 is merged so business package can get this from cached.
i.e. used from validations
```
svc, err := in.businessLayer.Svc.getService(namespace, service)
```

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
+package appender
+
+import (
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/graph"
+	"github.com/kiali/kiali/models/annotation"
+)
+
+const AnnotationAppenderName = "kialiAnnotation"
```
jshaughn, line 9: This is similar to the Sidecars appender and could feasibly be combined into that appender, but there are a few differences and I'm not opposed to adding a new appender specifically to handle these config overrides.   As mentioned above, I do think the name should change to better reflect what it is doing.  Assuming the proposed name change:

`const HealthConfigAppenderName = "healthConfig"`

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Name: annotation
+type AnnotationAppender struct{}
```
jshaughn, line 12: This would change to `HealthConfigAppender`.  Also, the appenders typically have a fairly robust comment above type decl.

Also, before a merge we'd need test code for this appender.

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+	if len(trafficMap) == 0 {
+		return
+	}
+
+	if getWorkloadList(namespaceInfo) == nil {
+		workloadList, err := globalInfo.Business.Workload.GetWorkloadList(namespaceInfo.Namespace)
+		graph.CheckError(err)
+		namespaceInfo.Vendor[workloadListKey] = &workloadList
+	}
+
```
jshaughn, line 30: You don't want to get the ServiceDefinitionList every time, instead you want to grab the cached version that presumably would have been established by `IstioAppender`.  So insert here:

```
	if getServiceDefinitionList(namespaceInfo) == nil {
		sdl, err := globalInfo.Business.Svc.GetServiceDefinitionList(namespaceInfo.Namespace)
		graph.CheckError(err)
		namespaceInfo.Vendor[serviceDefinitionListKey] = sdl
	}
```

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+	a.applyAnnotationPresence(trafficMap, namespaceInfo)
+}
+
+func (a *AnnotationAppender) applyAnnotationPresence(trafficMap graph.TrafficMap, namespaceInfo *graph.AppenderNamespaceInfo) {
```
jshaughn, line 34: just a name change here...

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, n := range trafficMap {
+		// Skip the check if this node is outside the requested namespace, we limit badging to the requested namespaces
+		if n.Namespace != namespaceInfo.Namespace {
+			continue
+		}
```
jshaughn, line 39: I think this check is valid as we onbly have the workloads for the current namespace.  Juts need to fix the comment.

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+
+		// We whitelist istio components because they may not report telemetry using injected sidecars.
+		if config.IsIstioNamespace(n.Namespace) {
+			continue
+		}
```
jshaughn, line 44: This I'm really not sure about.  We probably want to delete these lines as it seems feasible that a user could annotate workloads in the control plane namespace.

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+
+		// dead nodes tell no tales (er, have no pods)
+		if isDead, ok := n.Metadata[graph.IsDead]; ok && isDead.(bool) {
+			continue
+		}
```
jshaughn, line 49: Dead nodes have no pods but I think it's possible they have workloads.  I think we should delete these lines.

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+
+		// get the workloads for the node and check to see if they have sidecars. Note that
+		// if there are no workloads/pods we don't flag it as missing sidecars.  No pods means
+		// no missing sidecars.  (In most cases this means it was flagged as dead, and handled above)
```
jshaughn, line 53: Fix the comment

graph/telemetry/istio/appender/kiali_annotation.go
```diff
@@ -0,0 +1,69 @@
...
+		switch n.NodeType {
+		case graph.NodeTypeWorkload:
+			if workload, found := getWorkload(n.Workload, namespaceInfo); found {
+				annotations := annotation.GetAnnotation(workload.Annotations, annotation.GetKialiAnnotation())
```
jshaughn, line 57: I think there will be name changes here based on other comments.  But if not we want to only send back healthConfig overrides.  Again, I'd like to see Kiali-level typing as opposed to just the Go map.

jshaughn (changes requested):
@aljesusg After a short meeting with @lucasponce I now understand better the approach of this feature, and I do see the need for the additional appender support.  Because we still generate the health determination on the client, and this is not the static config we pull already from the server, we do need to look for the config overrides at graph-generation-time.  The good news is that the additional overhead should be negligible, as we already pull the information needed in other, non-optional, appenders.

It looks like you found the necessary touch points in the code, and it looks pretty close, but I did have a few comments.  Also, when the appender name is decided you will want to add the appender name to line 172 of doc.go.

graph/config/cytoscape/cytoscape.go
```diff
@@ -71,6 +71,7 @@ type NodeData struct {
 	HasCB           bool                `json:"hasCB,omitempty"`           // true (has circuit breaker) | false
 	HasMissingSC    bool                `json:"hasMissingSC,omitempty"`    // true (has missing sidecar) | false
 	HasVS           bool                `json:"hasVS,omitempty"`           // true (has route rule) | false
+	KialiAnnotation map[string]string   `json:"kialiAnnotation,omitempty"` // requested kialiAnnotation for NodeType
```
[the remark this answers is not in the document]
aljesusg: understood, I was thinking more in let this for future enhancements but you're right...our first step is only focus in health. I'll address the comments, thanks @jshaughn

graph/telemetry/istio/appender/kiali_annotation.go
aljesusg: In this case we can remove them, because for workload we are fetching the annotations of their first pod, so if they have not pods so not makes sense.

graph/config/cytoscape/cytoscape.go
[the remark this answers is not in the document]
aljesusg: instead to HealthConfig this should be a []Tolerance, it makes no sense set the full configuration for a specific node.

models/health.go
aljesusg: I nned to remove this method is not used :(

graph/config/cytoscape/cytoscape.go
```diff
@@ -71,6 +71,7 @@ type NodeData struct {
 	HasCB           bool                `json:"hasCB,omitempty"`           // true (has circuit breaker) | false
 	HasMissingSC    bool                `json:"hasMissingSC,omitempty"`    // true (has missing sidecar) | false
 	HasVS           bool                `json:"hasVS,omitempty"`           // true (has route rule) | false
+	HasHealthConfig map[string]string   `json:"HasHealthConfig,omitempty"` // set to the health config override
```
jshaughn, line 74: three things:
- json should be `hasHealthConfig`
- maintain alphebetical (move up after HasCB)
- recommended: I still think map[string]string deserves to be typed.  I had suggested `HealthConfig`, you suggest ` []Tolerance`.  But for me `[]Tolerance` introduces another term that then needs definition.    Remember that the type declared here is just local to this Cytoscape config generator, declared above with the other types.  So I think something like this would be fine (update the comment as you see fit):
```
// HealthConfig is a flexible map to pass available health information, tolerances, etc, to be interpreted by the client.
type HealthConfig map[string][string]
```

graph/meta.go
```diff
@@ -13,6 +13,7 @@ func NewMetadata() Metadata {
 
 // Metadata keys to be used instead of literal strings
 const (
+	HasHealthConfig MetadataKey = "hasHealthConfig"
```
jshaughn, line 16: Name change is good, please fix alphabetical ordering.

graph/telemetry/istio/appender/appender.go
```diff
@@ -22,6 +22,8 @@ func ParseAppenders(o graph.TelemetryOptions) []graph.Appender {
 	if !o.Appenders.All {
 		for _, appenderName := range o.Appenders.AppenderNames {
 			switch appenderName {
+			case HealthConfigAppenderName:
+				requestedAppenders[HealthConfigAppenderName] = true
```
jshaughn, line 26: Name change is good. Again, maintain alphabetical ordering where it is being used, it helps in code readability.

graph/config/cytoscape/cytoscape.go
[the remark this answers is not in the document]
jshaughn: I have a inline response to this in my next review.

graph/telemetry/istio/appender/health_config.go
```diff
@@ -0,0 +1,52 @@
...
+	if len(trafficMap) == 0 {
+		return
+	}
+
+	if getWorkloadList(namespaceInfo) == nil {
+		sdl, err := globalInfo.Business.Svc.GetServiceDefinitionList(namespaceInfo.Namespace)
+		graph.CheckError(err)
+		namespaceInfo.Vendor[workloadListKey] = &sdl
+	}
+
```
jshaughn, line 29: Still missing the getServiceDefinitionList.  See my previous review comment.

jshaughn (changes requested):
Looking better, still some changes to make, ans still need test code for the new appender.

graph/config/cytoscape/cytoscape.go
aljesusg: understood. Thanks

graph/telemetry/istio/appender/health_config.go
aljesusg: I added it in L25 @jshaughn 

https://github.com/kiali/kiali/blob/a542e95b525a6e0f903d6f236e333c818d88cc87/graph/telemetry/istio/appender/health_config.go#L25-L27

lucasponce:
I'd reviewed on Monday, but is there a UI more or less in good shape to test the end-to-end cases ?
I can review the code but I'd like to start from the feature first.

lucasponce (changes requested):
It needs a counterpart in the UI updated to be reviewed together in the same context.
Otherwise, if it's merged without UI review it may require changes in backend.

PR https://github.com/kiali/kiali-ui/pull/1986 is non update since some time.

No rush, but just let reviewers when it's ready to review them together.

lucasponce:
@aljesusg would you mind to review the CI errors on this PR ?
Thanks

lucasponce:
The regular expressions used are something that I'm not 100% conviced.

As we have introduced some "sugar syntax" for these scenario, and custom Regex here will be rare, I suggest to change the defaults in something that user can read better:

Original:
```
				Tolerance: []Tolerance{
					{
						Code:      "^5\\d\\d$",
						Protocol:  "http",
						Direction: ".*",
						Failure:   10,
					},
					{
						Code:      "^4\\d\\d$",
						Protocol:  "http",
						Direction: ".*",
						Degraded:  10,
						Failure:   20,
					},
					{
						Code:      "^[1-9]$|^1[0-6]$",
						Protocol:  "grpc",
						Direction: ".*",
						Failure:   10,
					},
					{
						Code:      "^-$", // no response is indicated with a "-" code
						Protocol:  "http|grpc",
						Direction: ".*",
						Failure:   10,
					},
				},
```

Suggested:
```
				Tolerance: []Tolerance{
					{
						Code:      "5XX",
						Protocol:  "http",
						// Direction: ".*", <- An empty value could be considered as ".*"
						Failure:   10,
					},
					{
						Code:      "4XX",
						Protocol:  "http",
						// Direction: ".*", <- An empty value could be considered as ".*"
						Degraded:  10,
						Failure:   20,
					},
					{
						Code:      "^[1-9]$|^1[0-6]$",
						Protocol:  "grpc",
						// Direction: ".*", <- An empty value could be considered as ".*"
						Failure:   10,
					},
					{
						Code:      "^-$", // no response is indicated with a "-" code
						Protocol:  "http|grpc",
						// Direction: ".*", <- An empty value could be considered as ".*"
						Failure:   10,
					},
				},
```

This will also help when iterate in how to show this information to the user.

aljesusg:
I only applied the 'X' Sintax to the annotations not to the default configuration. I need to change more code for this let me address it.

lucasponce:
Also, just put on the user's role:

```
{
  "healthConfig": {
    "rate": [
      {
        "namespace": "/.*/",
        "kind": "/.*/",
        "name": "/.*/",
        "tolerance": [
          {
            "code": "/^5\\d\\d$/",
            "degraded": 0,
            "failure": 10,
            "protocol": "/http/",
            "direction": "/.*/"
          },
          {
            "code": "/^4\\d\\d$/",
            "degraded": 10,
            "failure": 20,
            "protocol": "/http/",
            "direction": "/.*/"
          },
          {
            "code": "/^[1-9]$|^1[0-6]$/",
            "degraded": 0,
            "failure": 10,
            "protocol": "/grpc/",
            "direction": "/.*/"
          },
          {
            "code": "/^-$/",
            "degraded": 0,
            "failure": 10,
            "protocol": "/http|grpc/",
            "direction": "/.*/"
          }
        ]
      }
    ]
  }
}
```

could be simplified to
```
{
  "healthConfig": {
    "rate": [
      {
        "tolerance": [
          {
            "code": "5XX",
            "degraded": 0,
            "failure": 10,
            "protocol": "http",
          },
          {
            "code": "4XX",
            "degraded": 10,
            "failure": 20,
            "protocol": "http",
          },
          {
            "code": "/^[1-9]$|^1[0-6]$/",
            "degraded": 0,
            "failure": 10,
            "protocol": "grpc",
          },
          {
            "code": "/^-$/",
            "degraded": 0,
            "failure": 10,
            "protocol": "http|grpc",
          }
        ]
      }
    ]
  }
}
```

So, when some field is not present, it could be considered empty and and empty would mean ".*" in practice, then it would simplify the read of this.

The idea here is to allow regexp, but those should be more corner cases rather than the usual scenario.

aljesusg:
> Also, just put on the user's role:
> 
> ```
> {
>   "healthConfig": {
>     "rate": [
>       {
>         "namespace": "/.*/",
>         "kind": "/.*/",
>         "name": "/.*/",
>         "tolerance": [
>           {
>             "code": "/^5\\d\\d$/",
>             "degraded": 0,
>             "failure": 10,
>             "protocol": "/http/",
>             "direction": "/.*/"
>           },
>           {
>             "code": "/^4\\d\\d$/",
>             "degraded": 10,
>             "failure": 20,
>             "protocol": "/http/",
>             "direction": "/.*/"
>           },
>           {
>             "code": "/^[1-9]$|^1[0-6]$/",
>             "degraded": 0,
>             "failure": 10,
>             "protocol": "/grpc/",
>             "direction": "/.*/"
>           },
>           {
>             "code": "/^-$/",
>             "degraded": 0,
>             "failure": 10,
>             "protocol": "/http|grpc/",
>             "direction": "/.*/"
>           }
>         ]
>       }
>     ]
>   }
> }
> ```
> 
> could be simplified to
> 
> ```
> {
>   "healthConfig": {
>     "rate": [
>       {
>         "tolerance": [
>           {
>             "code": "5XX",
>             "degraded": 0,
>             "failure": 10,
>             "protocol": "http",
>           },
>           {
>             "code": "4XX",
>             "degraded": 10,
>             "failure": 20,
>             "protocol": "http",
>           },
>           {
>             "code": "/^[1-9]$|^1[0-6]$/",
>             "degraded": 0,
>             "failure": 10,
>             "protocol": "grpc",
>           },
>           {
>             "code": "/^-$/",
>             "degraded": 0,
>             "failure": 10,
>             "protocol": "http|grpc",
>           }
>         ]
>       }
>     ]
>   }
> }
> ```
> 
> So, when some field is not present, it could be considered empty and and empty would mean ".*" in practice, then it would simplify the read of this.
> 
> The idea here is to allow regexp, but those should be more corner cases rather than the usual scenario.
> 
> I'm not sure if I'm missing some case, but that's the idea, to simplify this config, and trying to unify them with also the annotation one.

In fact if a field is empty Kiali will use "allMatch" with the value .*

lucasponce:
> In fact if a field is empty Kiali will use "allMatch" with the value .*

I haven't review the implementation details, I'm just first review this from user's point of view, then perhaps we need to update the defaults in Kiali and operator/helm etc.

aljesusg:
> > In fact if a field is empty Kiali will use "allMatch" with the value .*
> 
> I haven't review the implementation details, I'm just first review this from user's point of view, then perhaps we need to update the defaults in Kiali and operator/helm etc.

Yes, I remember that we added that to the doc to show the user the options in the example, it makes no sense when we have the "table" doc with all opts. For each expresion we use this method in the frontend.

So in the case that we have not a value we return allMatch

```ts
const allMatch = new RegExp('.*');
export const getExpr = (value: RegexConfig | undefined, code: boolean = false): RegExp => {
  if (value) {
    if (typeof value === 'string' && value !== '') {
      const v = value.replace('\\\\', '\\');
      return new RegExp(code? replaceXCode(v):v);
    }
    if (typeof value === 'object' && value.toString() !== '/(?:)/') {
      return value;
    }
  }
  return allMatch;
};
```

I added the condition to let user set X or x in the code option @lucasponce 
https://github.com/kiali/kiali-ui/pull/2056/commits/148ab1014d8acc9ce8f340023abb3f0bcb977732

so user can set 4XX or 4xx or [4|5]xx

the UI will replace it by 4\d\d or [4|5]\d\d

lucasponce:
@aljesusg I don't understand the last commit.

Please, take the time to review my comments, the idea is that user don't read by default the complex regex, but a friendly format to express 4xx and 5xx.

This has to be propagated in more places, operator included.

P.S.: Try also to rebase with master both backend + frontend as multicluster work has been merged

jshaughn (approved):
This seems to be working in the case I tried.  I did waste some time trying to figure out the latest syntax, so I look forward to some documentation to go with the feature.  a couple of very minor comments that you can fix, but approving.

models/health_config.go
```diff
@@ -0,0 +1,26 @@
...
+const (
+	AllHealthAnnotation  AnnotationKey = ".*"
+	RateHealthAnnotation AnnotationKey = "health.kiali.io/rate"
+)
+
+func GetHealthConfigAnnotation() []AnnotationKey {
+	return []AnnotationKey{RateHealthAnnotation}
+}
+
+func GetHealthAnnotation(annotations map[string]string, filters []AnnotationKey) interface{} {
```
lucasponce, line 15: Perhaps "GetHealthAnnotation" should return a map[string]string instead of "interface{}".

Also, below there are multiple lines doing the conversion on the call:

```
GetHealthAnnotation(service.Annotations, GetHealthConfigAnnotation()).(map[string]string)
```

I guess if not found or if there is a type class it can return an error or a nil.

aljesusg: I remember that i had problems in graph/telemetry/istio/appender/health_config.go when I return that to the metadata let me check

aljesusg: mmm I did the commit with changes I am going to tests all again because I remember something about this

business/health.go
```diff
@@ -188,6 +191,7 @@ func (in *HealthService) getNamespaceServiceHealth(namespace string, services []
 	// Prepare all data (note that it's important to provide data for all services, even those which may not have any health, for overview cards)
 	for _, service := range services {
 		h := models.EmptyServiceHealth()
+		h.Requests.HealthAnnotations = models.GetHealthAnnotation(service.Annotations, HealthAnnotation).(map[string]string)
```
lucasponce: I've commented in the "GetHealthAnnotation", it can return directly a "map[string]string" with nil or error if you want to check that the annotation is not well formed.

business/health.go
```diff
@@ -311,12 +324,22 @@ func (in *HealthService) getAppRequestsHealth(namespace, app, rateInterval strin
...
 		rqHealth.AggregateInbound(sample)
 	}
 	for _, sample := range outbound {
 		rqHealth.AggregateOutbound(sample)
 	}
+	w, err := in.businessLayer.Workload.GetWorkload(namespace, workload, "", false)
+	if err != nil {
+		return rqHealth, err
+	}
+	if len(w.Pods) > 0 {
```
lucasponce, line 340: It seems now that it gets the annotation from the controller, instead of first Pod, it seems this check is not necessary.

business/services.go
```diff
@@ -275,6 +276,26 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime
 	return &s, nil
 }
 
+func (in *SvcService) UpdateService(namespace, service string, interval string, queryTime time.Time, jsonPatch string) (*models.ServiceDetails, error) {
```
lucasponce, line 279: Note, in the current reduced scenario, Kiali UI is not updated the annotations.
Users typically can update directly the controller.

We can leave this method in the code as it's similar as the workload one, but edition of the annotations are out of the scope for the moment.

models/health_config.go
lucasponce: Another comment, I looked close to this code.

I guess this signature:

```
func GetHealthAnnotation(annotations map[string]string, filters []AnnotationKey) interface{} {
```

Could be simplified as "filters" is always a constat in the current use case (if I looked the code correctly), so, I'd remove the "filters []AnnotationKey" and I'd use it internally of the method as a constant.

When this annotations feature grows, it can be refactored, but for now I guess it will help to simplify it.

graph/telemetry/istio/appender/health_config.go
```diff
@@ -0,0 +1,59 @@
...
+func (a *HealthConfigAppender) applyHealthConfigPresence(trafficMap graph.TrafficMap, namespaceInfo *graph.AppenderNamespaceInfo) {
+	for _, n := range trafficMap {
+		if n.Namespace != namespaceInfo.Namespace {
+			continue
+		}
+		// get the workload for the node and check to see if they have health configuration.
+		switch n.NodeType {
+		case graph.NodeTypeWorkload:
+			if workload, found := getWorkload(n.Workload, namespaceInfo); found {
+				n.Metadata[graph.HasHealthConfig] = models.GetHealthAnnotation(workload.HealthAnnotations, models.GetHealthConfigAnnotation())
```
lucasponce, line 49: Passing a "models.GetHealthConfigAnnotation()" to "models.GetHealthAnnotation" is weird, so I put a comment to use the filters as a constant in the package or method.

lucasponce (changes requested):
This time I've left some comments more related to code style and simplification.

No big blockers but some easy changes to address.

lucasponce (approved):
OK to consolidate.
As commented in the UI PR, no regressions, this feature is in active development and probably we will interate on it, but this is good.

PR #3364, opened 2020-10-28
[v1.24] Fix no trace displayed for istio ingress service

jotak:
Fix no trace displayed for istio ingress service

For Service traces, a post-filtering based on operation names was performed.
It's generally ok, except some special situations like when trace comes from ingress (or when trace isn't generated from envoy)
Now we're by-passing this post-filtering in the case where the service name matches the app name.

### DIFF
--- a/business/jaeger.go
+++ b/business/jaeger.go
@@ -62,14 +62,20 @@ func (in *JaegerService) GetServiceSpans(ns, service string, query models.Tracin
 	if err != nil {
 		return nil, err
 	}
-	return in.getFilteredSpans(ns, app, query, svcSpanFilter(ns, service))
+	var postFilter SpanFilter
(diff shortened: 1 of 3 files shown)


jotak:
@hhovsepy I'll open a second PR for master.

as discussed, this fixes the case of istio ingress (which was a real problem indeed) but *not* the one of jaeger-query, which is more a corner case without much importance IMO

The changes in code impact fetching traces and spans for any services (not just istio-ingress) so it's probably wise to do some non-regression for side-effects. I did some tests without seeing issues.

aljesusg (approved):
LGFM

hhovsepy (approved):
Verified:
Screenshot from 2020-10-28 14-49-30

jotak:
thanks!

PR #3368, opened 2020-10-28
Fix no trace displayed for istio ingress service

jotak:
Cherry-picking https://github.com/kiali/kiali/pull/3364 from v1.24

Fix no trace displayed for istio ingress service

For Service traces, a post-filtering based on operation names was performed.
It's generally ok, except some special situations like when trace comes from ingress (or when trace isn't generated from envoy)
Now we're by-passing this post-filtering in the case where the service name matches the app name.

### DIFF
--- a/business/jaeger.go
+++ b/business/jaeger.go
@@ -62,14 +62,20 @@ func (in *JaegerService) GetServiceSpans(ns, service string, query models.Tracin
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


jotak:
@aljesusg @hhovsepy this is just cherry-picking the other PR from v1.24 , nothing different, no conflict.

aljesusg (approved):
LGFM

Issue #3265, opened 2020-09-28
Need of a warning[KIA0302 No matching workload found for gateway selector in this namespace]? 


abhioncbr:
**General remarks**

I am getting the following warning in Kiali UI, `KIA0302 No matching workload found for gateway selector in this namespace.`
following below are my Ingress & istio Gateway manifests
```yaml
apiVersion: extensions/v1beta1
kind: Ingress
metadata:
  labels:
    app: frontend-nodejs-app 
  namespace: istio-system
  name: istio-securegw-nginx-ing
  annotations:
    kubernetes.io/ingress.class: nginx
    nginx.ingress.kubernetes.io/ssl-passthrough: "true"
    nginx.ingress.kubernetes.io/rewrite-target: /$1
    nginx.ingress.kubernetes.io/backend-protocol: "HTTPS"
spec:
  tls:
  - hosts:
    - istio-securegw-nginx-exmple.aks.com
  rules:
  - host: istio-securegw-nginx-exmple.aks.com
    http:
      paths:
      - backend:
          serviceName: internal-ingressgateway
          servicePort: 443
        path: /  
---
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  namespace: istio-securegw-sample
  name: istio-securegw-sample-gateway
spec:
  selector:
    istio: internal-ingressgateway # use istio default controller
  servers:
  - port:
      number: 8443
      name: https
      protocol: HTTPS
    tls:
      mode: SIMPLE
      credentialName: aks-ingress-tls
    hosts:
      - istio-securegw-nginx-exmple.aks.com
```

The warning makes sense since `internal-ingressgateway` is in the `istio-system` namespace, however in my opinion this is always going to be the same in most cases, so does this warning is needed or not? 

**Describe the bug**

For UI issues please also add a screenshot that shows the issue.

**Versions used**
Kiali: 1.17.9
Istio: 1.6.5
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9) AKS, 1.17.9

**To Reproduce**

**Expected behavior**

xeviknal:
Hello @abhioncbr,

the basic usage of this warning is to make sure that there is at least one running workload for the specified ingress. If you don't have any, then the gateway would be doing load balancing for a non existing ingress.

The warning is useful for multiple reasons, for instance, the following two:
1. You don't have any typo while typing the labels of your ingress.
2. That specific ingress has been removed or is down for some reason.

[...]

PR #3375, opened 2020-10-29
Include ControlPlane workloads to KIA0302 no matching workload

xeviknal:
Fixes https://github.com/kiali/kiali/issues/3265#

Goal: include istio-system workloads to the look up for workloads in the gateways selector checker.

I have several thoughts regarding this change:
1. It breaks the cross-namespace validation rule. Security-wise, bad-intended users without access to istio-system could gain information about workloads in the istio-system that match that specific selector. Is it that bad?
2. I think it might be different kinds of deployments where the ingress/egress are going to be in different namespaces than the control plane. Therefore, the validation will keep failing.
3. Because of the previous thoughts, I am wondering if we should remove the warning at all. 

Thoughts?

### DIFF
--- a/business/checkers/gateway_checker.go
+++ b/business/checkers/gateway_checker.go
@@ -9,10 +9,9 @@ import (
 const GatewayCheckerType = "gateway"
 
 type GatewayChecker struct {
-	GatewaysPerNamespace     [][]kubernetes.IstioObject
-	Namespace                string
-	WorkloadList             models.WorkloadList
-	ControlPlaneWorkloadList models.WorkloadList
+	GatewaysPerNamespace  [][]kubernetes.IstioObject
+	Namespace             string
+	WorkloadsPerNamespace map[string]models.WorkloadList
 }
 
 // Check runs checks for the all namespaces actions as well as for the single namespace validations
@@ -39,9 +38,8 @@ func (g GatewayChecker) runSingleChecks(gw kubernetes.IstioObject) models.IstioV
 
 	enabledCheckers := []Checker{
 		gateways.SelectorChecker{
-			ControlPlaneWorkloadList: g.ControlPlaneWorkloadList,
-			WorkloadList:             g.WorkloadList,
-			Gateway:                  gw,
+			Gateway:               gw,
+			WorkloadsPerNamespace: g.WorkloadsPerNamespace,
 		},
 	}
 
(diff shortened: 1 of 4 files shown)


lucasponce:
> It breaks the cross-namespace validation rule. Security-wise, bad-intended users without access to istio-system could gain information about workloads in the istio-system that match that specific selector. Is it that bad?

We assume that Istio users can access to the control plane, to have access in read mode to the names of the workloads in the control plane for validation purposes seems to me a scenario under control.

Indeed we have more like this inside kiali.

Write operations are always done under the user's privileges, but some read operations have granularity of the namespace.

lucasponce:
> I think it might be different kinds of deployments where the ingress/egress are going to be in different namespaces than the control plane. Therefore, the validation will keep failing.

Well, note that we have all this info cached, so accessing to it is faster. And kiali has access to this info. There should be a corner case that may fall if there are some tricky combination of permissions on two namespaces userA -> namespaceA, and userB -> namespaceB but namespaceA has a gateway defined in namespaceB ... something in that sense, but it seems a tweak scenario in this validation context.

lucasponce:
> Because of the previous thoughts, I am wondering if we should remove the warning at all.

Not sure if remove it but perhaps iterate on them.

I guess that Kiali is doing good validation within the mesh (VS/DR) but it seems users need help on these Ingress/Egress scenarios.

So, if this validation is useful I'd keep it even if that means some change in the background.

If there is an unbalance tradeoff between feature and implementation perhaps we can reconsider.

hhovsepy:
Before:
Screenshot from 2020-11-04 11-41-46

After:
Screenshot from 2020-11-04 11-56-02

But in list still warning is displayed.
Screenshot from 2020-11-04 11-56-15

xeviknal:
I've updated this validation to consider all the available workloads. This way, only false-positive when one user doesn't have access to the namespace with ingress/egress.

Nevertheless, I am concerned in the scenario where a user goes to the istio config list page for several namespaces, plus each namespace has hundreds of services each. Since the API calls in the list are per namespace, each call will fetch from cache the workloads of all the available ns. Should we implement a booster here? Like a shared variable for both requests?

lucasponce:
> Should we implement a booster here? Like a shared variable for both requests?

Perhaps we can see later, but `GetNamespaces` and `GetWorkloadList` queries are already cached, so it shouldn't have any additional backend call on regular scenario.

lucasponce (approved):

Issue #2893, opened 2020-06-11
[FEATURE] JSON based logging for day2 operations


MarianZoll:
**Is your feature request related to a problem? Please describe.**

Kiali is designed to support systems in being operated properly. One fundamental thing to do is forwarding the logs to the right place (e.g. Kibana) to make it easier for the DevOps team to read them as well as taking (automatic) actions around the logs. 
Unfortunately, the basic Go logger in the Kiali project does not support JSON based logging.

**Describe the solution you'd like**

Rework logging with a Go logger that supports JSON logging by either providing a startup environment or even better by providing an environmental variable. 

Some alternatives which come pretty handy which we might want to consider:
https://github.com/sirupsen/logrus
https://github.com/uber-go/zap

**Describe alternatives you've considered**

Changing logs in the log collector (e.g. fluentd) - stupid thing to do as it is very complex and the logging libs provide it out of the box

**Additional context**
None but I am happy to provide it.

israel-hdez:
For logging, most of the Kiali code (if not all of it) is using the helpers available at: https://github.com/kiali/kiali/blob/1f7edad56004e2ec317d58dc845a7f7dbf6166ba/log/log.go

So, I think it's not hard to change the log library. However, without doing more work to enhance the logged messages with more metadata, the best log line we could offer short term would be of the form:

`{timestamp: 1591910720, severity: "warn", message: "The logged message"}`

Would that be enough for your case? or would you need more metadata on it?

MarianZoll:
@israel-hdez 

That would be more than helpful, thanks for the response. It will reduce the size of transferred data and would allow the ops tool to automate actions and co.

We are currently working to get a couple of approvals from our mgmt. to add some contributions. Would that be something you (speaking about the Kiali. project) would be interested?

israel-hdez:
@MarianZoll Contributions will be more than welcome :smiley: Just make sure to discuss your enhancements with the team by opening GitHub issues.

MarianZoll:
@israel-hdez Haha this took some time but I was covered by a lot of project work. 

I had another look on the logging and how we could improve it with some advanced logging package. At the moment, a Kiali internal log class does abstract most of the logic (except formatting and co.): https://github.com/kiali/kiali/blob/master/log/log.go

[...]

PR #3382, opened 2020-10-30
Implement zerolog as logger, support proper json log format and human…

MarianZoll:
** Describe the change **

Implement a new logger that support json and text logging as well as human-readable log levels out of the box. Additionally, the logger can be extended based on new features of https://github.com/rs/zerolog without all the overhead and manual effort required for glog which is used so far.

** Issue reference **

https://github.com/kiali/kiali/issues/2893 

And pull requests for the helm-charts and operator repo:
https://github.com/kiali/helm-charts/pull/15
https://github.com/kiali/kiali-operator/pull/187

** Backwards incompatible? **

No, based on the discussions we had, we made sure that the `deployment.verbose_mode` setting is still supported. Nothing else should be affected.

** Documentation **

The documentation is added in the operator CR:

https://github.com/kiali/kiali-operator/pull/187/files#diff-b05613685dff7d8a5eb70946010acb19192f6f9edc83d292af21db46aa9be89dR280-R286

In case additional documentation is required somewhere, please let me know.

Should not. Only the log output will look different, but all information should stay the same.

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -53,7 +53,7 @@ func InitializeLogger() zerolog.Logger {
 
 	logFormat := resolveLogFormatFromEnv()
 	if logFormat != "json" {
-		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: zerolog.TimeFieldFormat})
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: zerolog.TimeFieldFormat, NoColor: true})
 	}
 
 	logLevel := resolveLogLevelFromEnv()
--- a/log/log.go
+++ b/log/log.go
@@ -1,9 +1,10 @@
 package log
 
 import (
+	"os"
+
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
-	"os"
 )
 
 // Configures the global log level and log format.
@@ -40,12 +41,10 @@ func Errorf(format string, args ...interface{}) {
 	log.Error().Msgf(format, args...)
 }
 
-// Debug will log a message at verbose level 4 and will ensure the caller's stack frame is used
 func Debug(args ...interface{}) {
 	log.Debug().Msgf("%s", args...)
 }
 
-// Debugf will log a message at verbose level 4 and will ensure the caller's stack frame is used
 func Debugf(format string, args ...interface{}) {
 	log.Debug().Msgf(format, args...)
 }
@@ -54,12 +53,10 @@ func IsDebug() bool {
 	return zerolog.GlobalLevel() == zerolog.DebugLevel
 }
 
-// Trace will log a message at verbose level 5 and will ensure the caller's stack frame is used
 func Trace(args ...interface{}) {
 	log.Trace().Msgf("%s", args...)
 }
 
-// Tracef will log a message at verbose level 5 and will ensure the caller's stack frame is used
 func Tracef(format string, args ...interface{}) {
 	log.Trace().Msgf(format, args...)
 }
@@ -86,13 +83,13 @@ func resolveLogLevelFromEnv() zerolog.Level {
 
 	switch logLevel {
 	case "0":
-		return zerolog.InfoLevel
-	case "1":
-		return zerolog.WarnLevel
-	case "2":
-		return zerolog.ErrorLevel
-	case "3":
 		return zerolog.FatalLevel
+	case "1":
+		return zerolog.ErrorLevel
+	case "2":
+		return zerolog.WarnLevel
+	case "3":
+		return zerolog.InfoLevel
 	case "4":
 		return zerolog.DebugLevel
 	case "5":
--- a/log/log.go
+++ b/log/log.go
@@ -2,6 +2,7 @@ package log
 
 import (
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -10,22 +11,34 @@ import (
 )
 
 // Configures the global log level and log format.
-func InitializeLogger() {
+func InitializeLogger()  zerolog.Logger  {
+	logTimeFieldFormat, isTimeFieldFormatDefined := os.LookupEnv("LOG_TIME_FIELD_FORMAT")
 
-	logTimeFieldFormat, isDefined := os.LookupEnv("LOG_TIME_FIELD_FORMAT")
-
-	if !isDefined {
+	if !isTimeFieldFormatDefined {
 		logTimeFieldFormat = time.RFC3339
+	}
+	zerolog.TimeFieldFormat = logTimeFieldFormat
+
+	logSamplerRateAsString, isSamplerRateDefined := os.LookupEnv("LOG_SAMPLER_RATE")
+	if isSamplerRateDefined {
+		logSamplerRate, err := strconv.Atoi(logSamplerRateAsString)
+		if err != nil {
+			log.Warn().Msgf("Provided sampling rate %s cannot be parsed to int32. " +
+				"No sampling rate will be set. Error: %v", logSamplerRateAsString, err)
+		} else {
+			log.Debug().Msgf("Setting log sample rate to every %dth event", logSamplerRate)
+			log.Logger = log.Sample(&zerolog.BasicSampler{N: uint32(logSamplerRate)})
+		}
 	}
 
 	if os.Getenv("LOG_FORMAT") != "json" {
-		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: logTimeFieldFormat})
-	} else {
-		zerolog.TimeFieldFormat = logTimeFieldFormat
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: zerolog.TimeFieldFormat})
 	}
 
 	logLevel := resolveLogLevelFromEnv()
 	zerolog.SetGlobalLevel(logLevel)
+
+	return log.Logger
 }
 
 func Info(args ...interface{}) {
--- a/log/log.go
+++ b/log/log.go
@@ -13,18 +13,31 @@ import (
 type Format string
 
 const (
-	FALLBACK_LOG_FORMAT = "text"
-	FALLBACK_TIMESTAMP  = time.RFC3339
+	FallbackLogFormat       = "text"
+	FallbackTimeFieldFormat = time.RFC3339
 )
+
+var supportedTimeFormats = [...]string{
+	time.ANSIC,
+	time.UnixDate,
+	time.RubyDate,
+	time.RFC822,
+	time.RFC822Z,
+	time.RFC850,
+	time.RFC1123,
+	time.RFC1123Z,
+	time.RFC3339,
+	time.RFC3339Nano,
+	time.Kitchen,
+	time.Stamp,
+	time.StampMilli,
+	time.StampMicro,
+	time.StampNano,
+}
 
 // Configures the global log level and log format.
 func InitializeLogger() zerolog.Logger {
-	logTimeFieldFormat, isTimeFieldFormatDefined := os.LookupEnv("LOG_TIME_FIELD_FORMAT")
-
-	if !isTimeFieldFormatDefined {
-		logTimeFieldFormat = FALLBACK_TIMESTAMP
-	}
-	zerolog.TimeFieldFormat = logTimeFieldFormat
+	zerolog.TimeFieldFormat = resolveTimeFormatFromEnv()
 
 	logSamplerRateAsString, isSamplerRateDefined := os.LookupEnv("LOG_SAMPLER_RATE")
 	if isSamplerRateDefined {
@@ -32,7 +45,7 @@ func InitializeLogger() zerolog.Logger {
 		if err != nil {
 			log.Warn().Msgf("Provided sampling rate %s cannot be parsed to int32. "+
 				"No sampling rate will be set. Error: %v", logSamplerRateAsString, err)
-		} else {
+		} else if logSamplerRate != 1 {
 			log.Debug().Msgf("Setting log sample rate to every %dth event", logSamplerRate)
 			log.Logger = log.Sample(&zerolog.BasicSampler{N: uint32(logSamplerRate)})
 		}
@@ -136,11 +149,12 @@ func resolveLogLevelFromEnv() zerolog.Level {
 	}
 }
 
+// Resolves and validates the log format. FallbackLogFormat is used as a default.
 func resolveLogFormatFromEnv() string {
 	logFormatEnv, isDefined := os.LookupEnv("LOG_FORMAT")
 
 	if !isDefined {
-		return FALLBACK_LOG_FORMAT
+		return FallbackLogFormat
 	}
 
 	switch logFormatEnv {
@@ -148,6 +162,24 @@ func resolveLogFormatFromEnv() string {
 		return logFormatEnv
 	default:
 		Warningf("Provided LOG_FORMAT %s is invalid. Fallback to text.", logFormatEnv)
-		return FALLBACK_LOG_FORMAT
+		return FallbackLogFormat
 	}
+}
+
+// Resolves and validates the provided log time format. FallbackTimeFieldFormat is used as a fallback.
+func resolveTimeFormatFromEnv() string {
+	logTimeFieldFormat, isDefined := os.LookupEnv("LOG_TIME_FIELD_FORMAT")
+
+	if !isDefined {
+		return FallbackTimeFieldFormat
+	}
+
+	for _, supportedTimeFormat := range supportedTimeFormats {
+		if logTimeFieldFormat == supportedTimeFormat {
+			return logTimeFieldFormat
+		}
+	}
+
+	Warningf("Provided LOG_TIME_FIELD_FORMAT %s is not supported. Fallback to %s", logTimeFieldFormat, FallbackTimeFieldFormat)
+	return FallbackTimeFieldFormat
 }
--- a/log/log.go
+++ b/log/log.go
@@ -1,16 +1,25 @@
 package log
 
 import (
-	"os"
-
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
+	"os"
+	"time"
 )
 
 // Configures the global log level and log format.
 func InitializeLogger() {
+
+	logTimeFieldFormat, isDefined := os.LookupEnv("LOG_TIME_FIELD_FORMAT")
+
+	if !isDefined {
+		logTimeFieldFormat = time.RFC3339
+	}
+
 	if os.Getenv("LOG_FORMAT") != "json" {
-		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr})
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stderr, TimeFormat: logTimeFieldFormat})
+	} else {
+		zerolog.TimeFieldFormat = logTimeFieldFormat
 	}
 
 	logLevel := resolveLogLevelFromEnv()
--- a/log/log.go
+++ b/log/log.go
@@ -10,12 +10,19 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
+type Format string
+
+const (
+	FALLBACK_LOG_FORMAT = "text"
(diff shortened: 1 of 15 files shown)


log/log.go
```diff
@@ -1,72 +1,108 @@
 package log
 
 import (
-	"fmt"
-
-	"github.com/golang/glog"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+	"os"
```
jmazzitelli, line 6: trivial - the imports in the kiali src code base are formatted specifically - base libs first, blank line, third party libs, blank line, kiali libs last. So this should be:

```suggestion
	"os"

	"github.com/rs/zerolog"
	"github.com/rs/zerolog/log"
```

log/log.go
```diff
@@ -1,72 +1,108 @@
...
-	glog.ErrorDepth(1, args...)
+	log.Error().Msgf("%s", args...)
 }
 
 func Errorf(format string, args ...interface{}) {
-	glog.ErrorDepth(1, fmt.Sprintf(format, args...))
+	log.Error().Msgf(format, args...)
 }
 
 // Debug will log a message at verbose level 4 and will ensure the caller's stack frame is used
```
jmazzitelli, line 43: we can remove this comment as numeric log levels are not used anymore
```suggestion
```

log/log.go
```diff
@@ -1,72 +1,108 @@
...
 func Debug(args ...interface{}) {
-	if glog.V(debug) {
-		glog.InfoDepth(1, "DEBUG: "+fmt.Sprint(args...)) // 1 == depth in the stack of the caller
-	}
+	log.Debug().Msgf("%s", args...)
 }
 
 // Debugf will log a message at verbose level 4 and will ensure the caller's stack frame is used
```
jmazzitelli, line 48: we can remove this comment as numeric log levels are not used anymore
```suggestion
```

log/log.go
```diff
@@ -1,72 +1,108 @@
...
-	}
+	log.Debug().Msgf(format, args...)
 }
 
 func IsDebug() bool {
-	return bool(glog.V(debug))
+	return zerolog.GlobalLevel() == zerolog.DebugLevel
 }
 
 // Trace will log a message at verbose level 5 and will ensure the caller's stack frame is used
```
jmazzitelli, line 57: we can remove this comment as numeric log levels are not used anymore
```suggestion
```

log/log.go
```diff
@@ -1,72 +1,108 @@
...
 func Trace(args ...interface{}) {
-	if glog.V(trace) {
-		glog.InfoDepth(1, "TRACE: "+fmt.Sprint(args...)) // 1 == depth in the stack of the caller
-	}
+	log.Trace().Msgf("%s", args...)
 }
 
 // Tracef will log a message at verbose level 5 and will ensure the caller's stack frame is used
```
jmazzitelli, line 62: we can remove this comment as numeric log levels are not used anymore
```suggestion
```

log/log.go
```diff
@@ -1,72 +1,108 @@
...
+	case "1":
+		return zerolog.WarnLevel
+	case "2":
+		return zerolog.ErrorLevel
+	case "3":
+		return zerolog.FatalLevel
+	case "4":
+		return zerolog.DebugLevel
+	case "5":
+		return zerolog.TraceLevel
```
jmazzitelli, line 99: I think these levels are incorrect. Based on the current [Kiali CR documentation](https://github.com/kiali/kiali-operator/blob/v1.24/deploy/kiali/kiali_cr.yaml#L355-L358), where it says,

```
Determines which priority levels of log messages Kiali will output.
# Typical values are "3" for INFO and higher priority, "4" for DEBUG and higher priority.
```
This is based on the glog "V" levels we used before, higher numbers get noisier as you go up. So in Kiali, 3 is INFO, 4 is DEBUG and 5 is TRACE , with 2 being WARN, and 1 being ERROR and 0 being FATAL.
So Kiali users will be using lower numbers to make the log quieter.

```suggestion
	case "0":
		return zerolog.FatalLevel
	case "1":
		return zerolog.ErrorLevel
	case "2":
		return zerolog.WarnLevel
	case "3":
		return zerolog.InfoLevel
	case "4":
		return zerolog.DebugLevel
	case "5":
		return zerolog.TraceLevel
```

jmazzitelli (changes requested):
Some trivial things, except for the verbose_mode conversion. That looks wrong. See below.

jmazzitelli (commented):
Looks like a conflict in go.mod - probably a minor thing that needs to be fixed.

MarianZoll:
Sorry missed that one this morning.

MarianZoll:
But seems like Travis as some issues with linting. I will make sure that the build passes.

MarianZoll:
Linting should be fine now. Not sure why `make clean build test-race` fails due to permission errors. But locally it seems to work fine.

jmazzitelli:
(I'll try to investigate the travis failures - I doubt it has anything to do with this PR).

jmazzitelli:
travis failure has been corrected. I'll be testing this stuff soon

jmazzitelli (commented):
I'm running the molecule tests on these now. I see the logs now look like this:

```
7:28PM INF Kiali: Version: v1.27.0-SNAPSHOT, Commit: 1f3247892d60945421d92cec16a63f499a85005c

7:28PM INF Using authentication strategy [anonymous]
7:28PM WRN Kiali auth strategy is configured for anonymous access - users will not be authenticated.
```
Notice the timestamps are less-than-helpful because they are only at granularity of minutes (`7:28PM` rather than, say, `15:28:15.032` - i.e. 24-hour clock at a more fine-grained milliseconds).

Having at minimum seconds in the timestamp (if not millis) allows one to easily refer to a line in the logs ("Hey, go check out the logs at 7:28:15") and you can roughly see how fast things are going (on the order of at least seconds).

Is there some way we can configure the timestamps (either hardcode it in the log.go module or maybe even add another `logger` setting?)

MarianZoll:
Sure, I will make some changes regarding the timestamp and probable other features that might be nice to include. I will attach a snippet to this PR and will update the operator and helm charts accordingly.

jmazzitelli (changes requested):
I would also like a simple unit test `log/log_test.go` here. I just wrote up the start of one that you should use to start with. Note that I am getting failures when I set the LOG_LEVEL env. var. to non-numeric level names (tests pass when using the numerics). I haven't looked at the failures yet - I'm just writing a simple unit test that I thought should just work.

[log_test.go.txt](https://github.com/kiali/kiali/files/5477882/log_test.go.txt)

I'm getting errors like:

```
    log_test.go:73: 
        	Error Trace:	log_test.go:73
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 1
        	Test:       	TestEnvVarLogLevel
        	Messages:   	LOG_LEVEL=WARN,levelToTest={stringLevel:WARN level:2}
```

Which means I'm setting the env var LOG_LEVEL to "WARN" but the resolveLogLevelFromEnv() is always returning LOG_LEVEL=INFO. And I'm seeing a warning log message spit out on these failures saying `{"level":"warn","time":"2020-11-02T15:33:47-05:00","message":"Provided LOG_LEVEL WARN is invalid. Fallback to info."}`

log/log.go
```diff
@@ -1,72 +1,105 @@
 package log
 
 import (
-	"fmt"
+	"os"
```
jmazzitelli, line 4: Add "strings" to get the `ToLower` function
```suggestion
	"os"
	"strings"
```

log/log.go
```diff
@@ -1,72 +1,105 @@
...
+	case "2":
+		return zerolog.WarnLevel
+	case "3":
+		return zerolog.InfoLevel
+	case "4":
+		return zerolog.DebugLevel
+	case "5":
+		return zerolog.TraceLevel
+	default:
+		logLevelFromString, err := zerolog.ParseLevel(logLevel)
```
jmazzitelli, line 98: Convert to lowercase since that is what the parser wants. Some people are going to think you need to specify the level as all uppercase (e.g. "WARN" or "DEBUG") but this wants all lower. Just support any case by converting to all lower:
```suggestion
		logLevelFromString, err := zerolog.ParseLevel(strings.ToLower(logLevel))
```

jmazzitelli (changes requested):
Convert to all lowercase when needing to parse the log level name.

jmazzitelli (changes requested):
Also, as mentioned [here](https://github.com/kiali/kiali-operator/pull/187#issuecomment-720732741), there are two places we have to clean up. Delete the setting and usage of VERBOSE_MODE env var in the makefiles. We don't need to set that via env var - devs can manually set it if they need to. So these need to be deleted:

- [ ] Delete line from make/Makefile.operator.mk: https://github.com/kiali/kiali/blob/master/make/Makefile.operator.mk#L72
- [ ] Delete line from Makefile: https://github.com/kiali/kiali/blob/master/Makefile#L113

MarianZoll:
Short summary of the latest commits:

- Implemented the changes you supposed. 
- Added the possibility for log sampling based on another environment variable. That might be a reasonable feature for productive system that want to reduce the log volume collected by external ops tools. By default, this feature will be disabled. 
- Extended the `log_test.go` you provided (Thank you!) with some regression tests for the logger itself to ensure that changes to the logging process would be identified as early as possible. 

In case you agree with the made changes, I will incorporate the settings for `LOG_TIME_FIELD_FORMAT` and `LOG_SAMPLER_RATE` to the helm chart and operator and update the PRs accordingly.

log/log.go
```diff
@@ -1,72 +1,129 @@
...
-	debug glog.Level = glog.Level(4)
-	trace glog.Level = glog.Level(5)
-)
+// Configures the global log level and log format.
+func InitializeLogger() zerolog.Logger {
+	logTimeFieldFormat, isTimeFieldFormatDefined := os.LookupEnv("LOG_TIME_FIELD_FORMAT")
+
+	if !isTimeFieldFormatDefined {
+		logTimeFieldFormat = time.RFC3339
+	}
```
jmazzitelli, line 19: Is there some way to parse the time field format for correctness? Similar to what you do for LOG_SAMPLER_RATE down below when you confirm it can be parsed as a number, and what you do for LOG_LEVEL in resolveLogLevelFromEnv().

Perhaps zerolog has a way to check it for validity. I'm thinking what happens if a person enters some invalid time field - we would want this to either fail-fast or default to some valid format. Something like:
```suggestion
	} else if !zerolog.IsTimeFieldFormatValid(logTimeFieldFormat) { // I have no idea if that API exists - but you get the idea
		log.Warn().Msgf("Provided time field format [%s] is not valid. Defaulting to RFC3339.", logTimeFieldFormat)
		logTimeFieldFormat = time.RFC3339
	}
```

log/log.go
```diff
@@ -1,72 +1,129 @@
...
+				"No sampling rate will be set. Error: %v", logSamplerRateAsString, err)
+		} else {
+			log.Debug().Msgf("Setting log sample rate to every %dth event", logSamplerRate)
+			log.Logger = log.Sample(&zerolog.BasicSampler{N: uint32(logSamplerRate)})
+		}
+	}
+
+	if os.Getenv("LOG_FORMAT") != "json" {
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: zerolog.TimeFieldFormat})
+	}
```
jmazzitelli, line 36: Again, I wonder if we should test for validity of the LOG_FORMAT using some zerolog parse function?

jmazzitelli (changes requested):
> In case you agree with the made changes, I will incorporate the settings for `LOG_TIME_FIELD_FORMAT` and `LOG_SAMPLER_RATE` to the helm chart and operator and update the PRs accordingly.

Yes, I like this. Let's incorporate the new settings in the helm chart and operator PRs accordingly. Aside from my comments here in this review, this all looks good. 

If you can incorporate those validity checks in the code which I talk about in this review, that's great; but if there is no easy way to perform those validity checks, don't worry about it. In other words, my requested changes here are minor and low priority. You tell me if it is easy to incorporate those changes - if not, I'll remove my "request for changes" in this PR and we can move forward with the review/approval process to get this merged soon.

Thanks for all the help!

log/log_test.go
```diff
@@ -0,0 +1,215 @@
...
+				"LOG_TIME_FIELD_FORMAT": time.RFC1123Z,
+			},
+			expectedLogMessage: "Kiali logging test 2: map[LOG_FORMAT:json LOG_LEVEL:debug LOG_TIME_FIELD_FORMAT:Mon, 02 Jan 2006 15:04:05 -0700]",
+		},
+	}
+
+	for index, test := range tests {
+
+		t.Logf("Cleaned env vars")
+		os.Clearenv()
```
jmazzitelli, line 135: Probably should be careful here. You are clearing the entire environment which is going to affect all future unit tests that are run after this one executes. You should consider using a "defer" here and rebuild the environment back to the way it was before this test function executes so future unit tests don't lose their environment. So far, all unit tests pass, but this could get ugly in the future with this bug being VERY hard to find ("where did all my env vars go??") :)

jmazzitelli (changes requested):
Consider fixing this as well.

log/log_test.go
```diff
@@ -0,0 +1,215 @@
...
+		{
+			envSettings: map[string]string{
+				"LOG_FORMAT":            "json",
+				"LOG_LEVEL":             "debug",
+				"LOG_TIME_FIELD_FORMAT": time.RFC1123Z,
+			},
+			expectedLogMessage: "Kiali logging test 2: map[LOG_FORMAT:json LOG_LEVEL:debug LOG_TIME_FIELD_FORMAT:Mon, 02 Jan 2006 15:04:05 -0700]",
+		},
+	}
+
```
jmazzitelli, line 131: Because the test is clearing the entire environment, it will affect all future unit tests that are run after this one executes. So far, all unit tests pass, but this could get ugly in the future with this bug being VERY hard to find ("where did all my env vars go??") :)

Here's a fix - using "defer", this unit test will put the environment back to the way it was before this test function executes so future unit tests don't lose their environment. 

```suggestion

	originalEnv := os.Environ()
	defer func() {
		for _, kvp := range originalEnv {
			kvpArray := strings.Split(kvp, "=")
			os.Setenv(kvpArray[0], kvpArray[1])
		}
		return
	}()
```

jmazzitelli (changes requested):

log/log.go
MarianZoll: Hm it does rely on the standard time package (https://golang.org/pkg/time/) which to the best of my knowledge could take any malformed time format and parse it to a non-sense time object. 

I tried to do a couple of format checks on the current timestamp ála:
```
timestamp := time.Now()
parsedAsString := timestamp.Format(format)
parsedAsTimestamp, err := time.Parse(format, parsedAsString)
```
but this seems to be unlikely successful as the format can range from `stupidInput`, `2006` to a standard format like `2006-01-02T15:04:05Z07:00`. Especially the fact that you can provide wrong formats like `2020` which will completely mess up the logic (parses the year to 9090) will probably brake the check routine. 

Neither the time module itself nor zerolog has validators for the format. What we might want to consider is that we only support a subset of time formats like the ones shipped with the time module:
```
	ANSIC       = "Mon Jan _2 15:04:05 2006"
	UnixDate    = "Mon Jan _2 15:04:05 MST 2006"
	RubyDate    = "Mon Jan 02 15:04:05 -0700 2006"
	RFC822      = "02 Jan 06 15:04 MST"
	RFC822Z     = "02 Jan 06 15:04 -0700" // RFC822 with numeric zone
	RFC850      = "Monday, 02-Jan-06 15:04:05 MST"
	RFC1123     = "Mon, 02 Jan 2006 15:04:05 MST"
	RFC1123Z    = "Mon, 02 Jan 2006 15:04:05 -0700" // RFC1123 with numeric zone
	RFC3339     = "2006-01-02T15:04:05Z07:00"
	RFC3339Nano = "2006-01-02T15:04:05.999999999Z07:00"
	Kitchen     = "3:04PM"
	// Handy time stamps.
	Stamp      = "Jan _2 15:04:05"
	StampMilli = "Jan _2 15:04:05.000"
	StampMicro = "Jan _2 15:04:05.000000"
	StampNano  = "Jan _2 15:04:05.000000000"
```

We could check against those and fallback to `RFC3339` in case the validations fails. What do you think?

log/log.go
MarianZoll: Guess this is already some kind of validation. Everything else than `json` will result in a regular text output. We could validate against an enum of text and json to give some debugging info to the user. Guess that might be reasonable to do.

MarianZoll:
I fixed the environment setup and validated the `LOG_FORMAT`. I adopt the `LOG_TIME_FIELD_FORMAT` based on our discussion later on.

log/log.go
MarianZoll: Proposal see below.

jmazzitelli (commented):
OK, I ran this and it all works. Thanks @MarianZoll !

I have one thing I want to think about before giving this the final approval. It is probably a minor thing - but I want to think about it. If anyone has suggestions or comments, feel free to do so.

It has to do with the colored fonts emitted by zerolog. If you look at the logs within the k8s console (in my case, the OpenShift Console), the log messages are ugly and hard to read due to the color ASCII codes embedded in the log messages. For example:
image

While that is ugly, the `kubectl logs` (or `oc logs`) command will output the color-codes, too, but if your command line window can support them, the output looks nice.

image

So I'm not sure what we should do or which log output we should cater to (console window or command line window?).

Is it possible to add yet another config option to let the user choose to emit colored logs or not? I'm not sure if zerolog has that config option, or, even if they do, should we care about this?

jmazzitelli:
Interestingly, if you change to "json" format, it actually looks better in the console:
image
so maybe that is the "solution". If someone doesn't like the color-codes (in case they are not interpreted correctly by the viewer and you see the codes rather than the colors), they can switch to 'json' and it will look "better" albeit with a little more noise (the "keys" in the json)

jmazzitelli:
FWIW: I'm leaning towards leaving everything as-is (rather than bothering to add more config options). Users can switch to 'json' format if they'd like.

So I am inclined to merge this as-is, but I'll wait to hear other's opinions before doing so.

jshaughn:
@jmazzitelli I don't think we want colorization at all.  It could generate unexpected results in any non-terminal log display (e.g. OS console, Kiali).  In Kiali it could also confuse parsing, or regex-based show/hide. 

I don't think it needs an option, I think it should be turned off.

jmazzitelli:
> I don't think it needs an option, I think it should be turned off.

OK, it looks easy to do that:
```
log := zerolog.New(zerolog.ConsoleWriter{Out: os.Stdout, NoColor: true})
```
I'll add a suggested fix so @MarianZoll can change it quickly. Should be a simple one-line change.

log/log.go
```diff
@@ -1,72 +1,185 @@
...
+				"No sampling rate will be set. Error: %v", logSamplerRateAsString, err)
+		} else if logSamplerRate != 1 {
+			log.Debug().Msgf("Setting log sample rate to every %dth event", logSamplerRate)
+			log.Logger = log.Sample(&zerolog.BasicSampler{N: uint32(logSamplerRate)})
+		}
+	}
+
+	logFormat := resolveLogFormatFromEnv()
+	if logFormat != "json" {
+		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: zerolog.TimeFieldFormat})
```
jmazzitelli, line 56:
```suggestion
		log.Logger = log.Output(zerolog.ConsoleWriter{Out: os.Stdout, TimeFormat: zerolog.TimeFieldFormat, NoColor: true})
```

jmazzitelli (changes requested):
We'll want to turn off colorization for the "text" format.

log/log.go
jmazzitelli: This is the final change. Everything else is good and working. Once this change is incorporated, we can merge the three PRs.

MarianZoll:
Wanted to propose the same. Simply turning it off sounds reasonable to do. I committed your suggestion. 

@jmazzitelli Thank you for guiding and reviewing the PR. This was really fun implementing and we are looking forward using the features in multiple projects to optimize the day-2 ops of the apps and opps stack! 

Let me know in case something else is needed..

jmazzitelli (approved):

PR #3383, opened 2020-10-30
Add support to setting up the web port in the configuration

### LINKED ISSUE #3111: Add configuration option to specify Kiali's public port (OIDC support)
**Describe the bug**

When using Kiali behind an Istio VirtualService (or likely a Kubernetes Ingress too), the openid login button redirects me to `<schema>://<web_fqdn>:20001`. Since I'm terminating all ingress on port 443, this behaviour breaks my ability to use openid.

**Versions used**
Kiali: 1.22
Istio: 1.6
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9) Kubernetes 1.18.6 bare-meal

**To Reproduce**
1. Configure Kiali behind an Ingress with a public FQDN (in my case, `kiali.elpenguino.net`)
2. Configure openid auth
3. At the initial screen, click "Log in with OpenID"
4. You are redirected to `https://kiali.elpenguino.net:20001/kiali/api/auth/openid_redirect`

**Expected behavior**

I should be able to override the URL to remove the `:20001` component. This should also apply to the redirect_url passed during the OIDC login process.

cfcosta:
Closes #3111.

### DIFF
--- a/util/httputil/http_util.go
+++ b/util/httputil/http_util.go
@@ -128,6 +128,8 @@ func GuessKialiURL(r *http.Request) string {
 	// take priority.
 	if fwdPort, ok := r.Header["X-Forwarded-Port"]; ok && len(fwdPort) == 1 {
 		port = fwdPort[0]
+	} else if len(cfg.Server.WebPort) > 0 {
+		port = cfg.Server.WebPort
 	} else if len(r.URL.Host) != 0 {
 		if len(r.URL.Port()) != 0 {
 			port = r.URL.Port()
--- a/util/httputil/http_util.go
+++ b/util/httputil/http_util.go
@@ -126,10 +126,10 @@ func GuessKialiURL(r *http.Request) string {
 	// priority, because this is the port where the pod is listening, which may
 	// be mapped to another public port via the Service/Ingress. So, HTTP headers
 	// take priority.
-	if fwdPort, ok := r.Header["X-Forwarded-Port"]; ok && len(fwdPort) == 1 {
-		port = fwdPort[0]
-	} else if len(cfg.Server.WebPort) > 0 {
+	if len(cfg.Server.WebPort) > 0 {
 		port = cfg.Server.WebPort
+	} else if fwdPort, ok := r.Header["X-Forwarded-Port"]; ok && len(fwdPort) == 1 {
+		port = fwdPort[0]
 	} else if len(r.URL.Host) != 0 {
 		if len(r.URL.Port()) != 0 {
 			port = r.URL.Port()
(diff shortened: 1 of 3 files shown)


israel-hdez (approved):
It's working OK.

util/httputil/http_util.go
```diff
@@ -128,6 +128,8 @@ func GuessKialiURL(r *http.Request) string {
 	// take priority.
 	if fwdPort, ok := r.Header["X-Forwarded-Port"]; ok && len(fwdPort) == 1 {
 		port = fwdPort[0]
+	} else if len(cfg.Server.WebPort) > 0 {
+		port = cfg.Server.WebPort
```
israel-hdez, line 132: Can you, please make this the setting with the most priority?
I see that if both have the `X-Forwarded-Port` header is present and also the manual config is present, the value at the HTTP headers are taken. This would be misleading. I think whatever is present in the config should take priority over any other setting.

israel-hdez (changes requested):

util/httputil/http_util.go
cfcosta: In general I think that runtime configuration should have higher priority than the one on the configuration file, but from what I've seen elsewhere we're not doing this for other keys, so changed. I've also added a test to cover this case.

jshaughn: I agree that in general runtime would override the static config.  But in this case it seems the static config would only be set in order to force the web port to that value.  Sorry to stick my head in here, hopefully I'm not off base.

cfcosta: It is important to note that the issue that triggered this change happened because the webserver did not report the forwarded for headers. In this case, even if the configuration conflicted, the proxy should have the higher priority.

However, I do think that consistency is more important than being intuitive.

israel-hdez: My perception was that people were preferring to have a configuration option rather than configure the HTTP headers in the proxy (or that's what I understand from this comment: https://github.com/kiali/kiali/issues/3111#issuecomment-673056301 -- I also remember one or two other people asking for this option.). It's the main reason why I think the "static" config should take priority.

But well... we can argue :)

israel-hdez (approved):
Working OK, so approving.
Merge if you think that it's OK that the "static config" is taking the highest priority

util/httputil/http_util.go
cfcosta: We all agree here, only discussing semantics :)

cfcosta:
With reasonable certainty (there's a test :) ).

PR #3385, opened 2020-10-30
new hack script for ci when testing with openshift

jmazzitelli:
run "--help" for description of this tool. Can be used as a cron job to periodically run the molecule tests on an existing openshift cluster

### DIFF
--- a/hack/ci-openshift-molecule-tests.sh
+++ b/hack/ci-openshift-molecule-tests.sh
@@ -1,79 +1,367 @@
 #!/bin/bash
 
 #
-# This script is designed to be run via a cronjob so the latest Kiali images
-# can be tested periodically using the molecule tests.
-#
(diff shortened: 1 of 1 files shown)


jmazzitelli:
A quick basic test (that runs only a single molecule test `token-test`) that should work for everyone that has "docker" is this (note: you can run this from any directory - including `/tmp`):

```
bash <(curl -s https://raw.githubusercontent.com/jmazzitelli/kiali/hack-ci-openshift/hack/ci-openshift-molecule-tests.sh) \
 --kubeadmin-pw  YOUR-KUBEADMIN-PASSWORD-HERE \
 --openshift-api https://api.YOUR-HOST-HERE:6443 \
 -oc             $(which oc) \
 -at             token-test
```

If you require podman, pass the extra option `-dorp podman` to the command.

jshaughn (approved):
Good.

PR #3392, opened 2020-11-03
remove unused/obsolete hack files/scripts

jmazzitelli:
This script starts OpenShift 3.x and is the only thing that uses that old Maistra yaml. So remove it to clean up `hack/` directory.

This cleans up other hack/ stuff.


jshaughn (approved):
Hack and Slay Day

PR #3396, opened 2020-11-03
[hack] install SM 2.0

jmazzitelli:
hack script to install Red Hat ServiceMesh 2.0

### DIFF
--- a/hack/istio/install-sm2.sh
+++ b/hack/istio/install-sm2.sh
@@ -1,0 +1,610 @@
+#!/bin/bash
+
+##############################################################################
+# install-sm2
+#
+# Run this script to install Service Mesh 2.x.
+#
+# This script takes one command whose value is one of the following:
+#   sm-install: installs service mesh into the cluster
+# sm-uninstall: removes all service mesh components
+#   bi-install: installs bookinfo demo into the cluster
+#  k-uninstall: removes only kiali components
+#
+# This script accepts several options - see --help for details.
+#
+##############################################################################
+
+########################################
+# START FUNCTIONS
+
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Almost done - all commands work, but getting this in the k-uninstall command (though Kiali still gets uninstalled)
```
HACK: Telling ServiceMesh to disable Kiali
Error from server: admission webhook "smcp.mutation.maistra.io" denied the request: v2.ServiceMeshControlPlane.Spec: v2.ControlPlaneSpec.Addons: v2.AddonsConfig.Kiali: v2.KialiAddonConfig.Enablement: Enabled: ReadBool: expect t or f, but found ", error found in #10 byte of ...|enabled":"false","in|..., bigger context ...|"install":{},"name":"jaeger"},"kiali":{"enabled":"false","install":{},"name":"kiali"},"prometheus":{|...
```
Once I fix that, this can get ready for review and merged.

jmazzitelli:
Fixed that last error (needed to specify the value as a boolean `false` not as a string `"false"`).
This is ready for review. 
Devs now have an easy way to install Service Mesh 2.0 in an OpenShift cluster, complete with Elastic Search, Jaeger, and Kiali.
If devs want their own version of Kiali, they can uninstall the one this script installs via the `k-uninstall` command.

jshaughn (approved):
Tested on OS 4.5.  Initially it failed because the webhooks did not have enough time to initialize before the script tried to use them.  That was fixed and now it's working.  Also tested bi-install and it worked (I wish it had mongo), and sm-unintsall.

Good stuff, now I can run downstream.

PR #3397, opened 2020-11-03
Add timeout and client close

### LINKED ISSUE #2905: Connections to the OpenShift API are kept open and accumulating until no more connections can be established, turning Kiali inaccessible
Happens on Kiali latest master.
OCP 4.5
Istio 1.6.1

When logged in to Kiali from 2 different browsers (FF and Chrome), after the second browser login, the first browser session is timeout, and it sows either session timeout message, or "Error trying to get OAuth metadata".
Started happening recently and not constantly reproducible.
Only after some time login works for one of browsers.

Screenshot from 2020-06-15 16-41-03
Screenshot from 2020-06-15 16-42-34

[kiali-7d8d7dfcf9-rw89t-kiali.log](https://github.com/kiali/kiali/files/4781000/kiali-7d8d7dfcf9-rw89t-kiali.log)

jshaughn:
As pointed out by @jotak , the go http client does not have a timeout by default, which can cause a hang in API calls.  Apply a timeout when interacting with the OpenAuth API.  Also, don't leave behind any keep-alive sessions.

In testing under load we did see the timeout trigger and the issue seems to be resolved with these changes.  This fix seemed to resurface a lingering issue, #3313, which we were hopefully able to finally squash.  So that UI PR is being required here.

UI PR: https://github.com/kiali/kiali-ui/pull/1994

Fixes #2905

### DIFF
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -57,7 +57,7 @@ type OAuthRouteTLSSpec struct {
 
 const serverPrefix = "https://kubernetes.default.svc/"
 
-var defaultRequestTimeout, _ = time.ParseDuration("10s")
+const defaultRequestTimeout = 10 * time.Second
 
 var kialiNamespace string
 
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"time"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
@@ -107,7 +108,7 @@ func getOAuthAuthorizationServer() (*OAuthAuthorizationServer, error) {
 
 	if err != nil {
 		log.Error(err)
-		message := fmt.Errorf("could not send request to the Openshift OAuth API: %v", err)
+		message := fmt.Errorf("could not get OAuthAuthorizationServer: %v", err)
 		return nil, message
 	}
 
@@ -115,7 +116,7 @@ func getOAuthAuthorizationServer() (*OAuthAuthorizationServer, error) {
 
 	if err != nil {
 		log.Error(err)
-		message := fmt.Errorf("could not parse data from the Openshift API: %v", err)
+		message := fmt.Errorf("could not parse OAuthAuthorizationServer: %v", err)
 		return nil, message
 	}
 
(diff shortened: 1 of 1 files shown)


jotak:
@jshaughn as found earlier, there are other places where a timeout is missing, do you want to fix them through that same PR or do another one? I fear the same issue could happen, unclosed connection piling up, resulting in resource leaks.

business/openshift_oauth.go
```diff
@@ -56,6 +57,8 @@ type OAuthRouteTLSSpec struct {
 
 const serverPrefix = "https://kubernetes.default.svc/"
 
+var defaultRequestTimeout, _ = time.ParseDuration("10s")
```
jotak, line 60: `10*time.Second` would avoid parsing a string (is there any advantage to parse here?)

jshaughn: Is there a disadvantage?   I see it as very readable and it is only done one time.

jshaughn:
> there are other places where a timeout is missing

@jotak I wanted to limit this specifically to the interactions with OAuth API around login, because that is where we were seeing a reported problem and this fix in particular we want to backport to our supported releases.

I do think we should add Timeouts in other uses of http Client, but since we haven't had a reported problem I think that could probably just go upstream.  I can create another issue, what do you think?

business/openshift_oauth.go
jshaughn: I decided there is an advantage, I can use const as opposed to var.  Made the change.

jotak:
> > there are other places where a timeout is missing
> 
> @jotak I wanted to limit this specifically to the interactions with OAuth API around login, because that is where we were seeing a reported problem and this fix in particular we want to backport to our supported releases.
> 
> I do think we should add Timeouts in other uses of http Client, but since we haven't had a reported problem I think that could probably just go upstream. I can create another issue, what do you think?

I don't know ... since it's identified and we know it can cause stability issues, I'd rather be in favour of backporting the other as well. Also because it's an easy fix without any negative impact afaict. Could be in a different issue/PR though.

jshaughn:
@jotak Well I looked for more instances without a timeout and to my pleasant surprise I don't see any.  It seems, possibly by luck, that we are consistently applying 10s timeouts.  The exceptions are Jaeger API calls, which are using 5s.  Do you see any non-test code, outside of the PR, that I missed?

israel-hdez (approved):
I think @jotak is referring to this particular line: https://github.com/kiali/kiali/blob/d042b5cb680a39534aeecedc7ca30c9d55294076/status/versions.go#L112

It is using a "global" client which may or may not be specifying the timeout.
We didn't do testing against the involved endpoint calling that function... this would require validating if it's really impacted so as @jshaughn said, I think we should work on it (if needed) on a new issue.

So, IMO, this PR is good. Approving...

jshaughn:
Wow, I even looked for instances of http.Get() but missed that one.  Good catch. But yes, let's proceed.

mattmahoneyrh (approved):

Issue #3203, opened 2020-09-11
Envoy config dump


jotak:
It would be fairly easy, and I think quite interesting, to allow per-pod config dump of envoy configuration.

From pods list in workload details:
![Capture d’écran de 2020-09-11 11-06-51](https://user-images.githubusercontent.com/2153442/92899483-2202e500-f41f-11ea-9fb6-efca8f76a848.png)

There could be a new column like "config" with a "dump" button per pod. Clicking on it would show up dialog with the dump or provide dump file to download.

From Kiali backend, it's just an http GET on `http://istiod.istio-system:15014/debug/config_dump?proxyID=<podname>.<namespace>`

This is quite helpful for debugging envoy

lucasponce:
I think that this is duplicated to some issue that @xeviknal is investigating, perhaps it's filled with a different description, but I'm sure that it's related to fetch status of the proxies.

xeviknal:
@lucasponce @jotak in the scope of the https://github.com/kiali/kiali/issues/1615 and https://github.com/kiali/kiali/pull/3126 I had the idea of showing the diff between istiod and envoy config in order to make the debugging task easier.

However, I didn't thought about the feature of show the whole config. Good point. Nevertheless, this would share most of the code developed to get the proxy-status and the diff because it is fetch from the same source: `istiod`.

jotak:
Yes perhaps it can be merged into the same task, I don't know, do as you wish.
The "diff" is interesting to troubleshoot out-of-sync issues, but sometimes it's not enough, when playing with envoy filters I've come several times to debug by looking at the whole filters config which I found only in the full dump.

xeviknal:
After some research, the next good step should be to show the envoy config (and config diff) at workload level (taking advantage of the kiali actions in the workload details page).

The feature should provide a high level view where users could either see the full envoy config dump or the diff between the istiod version and the actual envoy config dump per each pod.

As next steps, users might have the ability to:
- see the cluster, listeners, routes or endpoints at pod level.
- inspect the virtual listeners for both inbound and outbound traffic.
- inspect routes for given ports.
- Check the connectivity between the envoy and the istiod pod.

abonas:
In indeed sounds very helpful to get this info.
I'm wondering if generating the dump is a heavy action that might influence the proxy?
If several users, or even the same user click, say, 3 times in a row on "generate dump", will it have impact on the proxy, or it's such a light and fast action (and a tiny dump) that it won't influence anything?

[...]

PR #3400, opened 2020-11-06
Add ConfigDump endpoint at pod level

xeviknal:
Fixes https://github.com/kiali/kiali/issues/3203
Needs https://github.com/kiali/kiali-ui/pull/2006
Needs https://github.com/kiali/kiali-operator/pull/213
Needs https://github.com/kiali/helm-charts/pull/23

proxy-details


jotak:
> There is the need to do a lot of ProtoBuffer UnMarshalling in order to have the config_dump data. And AFAIK, a lot of imports from the Envoy control plane needed to happen for that.
> A lot of logic is quite similar to the one needed in istioctl. Should we consider importing istioctl modules?

If these models are already in istioctl, perhaps yes. They are not in main istio codebase? I guess another option to consider could be to get the protobuf definitions and introducing the related code generation in our build steps.
The problem is similar to how we deal with the various istio objects (vs, dr, etc.) or not?

xeviknal:
There are two types of models: Ones for fetching the info from Envoy and the others for filtering and printing the information.

Importing the protobuf definition, I guess it would imply a high number of import also. Just the Cluster protobuf definition have more than 45 fields and perhaps 2/3 of them has different protobuf definitions in it (some might be shared, but not many).

Regarding to the option to import, I've tried this morning, and it looks like the import is way more bigger than just importing the envoy-control-plane API.

`The problem is similar to how we deal with the various istio objects (vs, dr, etc.) or not?`
I think it is not exactly the same. AFAIK We receive Istio Objects as plain Go structs instead of protobufs. That is because we are fetching them from K8S API. Whereas in this case, we fetch the information straight forward from the envoy. Therefore, there is the need to convert envoy's bytes to structured data through the protobuf.

The second part of the problem is filtering and printing the information. In this part, we will be replicating same logics in Kiali that in Istio side. Filtering by port, by name, by address, are tools that are already implemented there. Also sorting clusters for making it easier, showing the certificates in a meaningful way.

The feeling is that, implementing that from scratch without actually copying exact behavior, it might be less useful that the istioctl command.

lucasponce:
Hi @xeviknal, would you mind to rebase ?
image

I'd like to focus on the review, but it seems that your working branch is old.
Thx

models/config_dump.go
```diff
@@ -0,0 +1,398 @@
+package models
+
+import (
+	"fmt"
+	"strconv"
+	"strings"
+
+	adminapi "github.com/envoyproxy/go-control-plane/envoy/admin/v3"
```
lucasponce, line 8: There are a lot of imports of core features of envoy for bringing a "model".
Perhaps on this case it's safe, but also consider clonning the struct if this could bring extra non-necessary imports.

business/proxy_status.go
```diff
@@ -74,3 +76,47 @@ func xdsStatus(sent, acked string) string {
...
+	var err error
+	promtimer := internalmetrics.GetGoFunctionMetric("business", "ProxyStatus", "GetConfigDump")
+	defer promtimer.ObserveNow(&err)
+
+	dump, err := in.k8s.GetConfigDump(namespace, pod)
+	if err != nil {
+		return nil, err
+	}
+
+	cd := models.NewConfigDump(dump)
```
lucasponce, line 90: Discussed with @xeviknal if at this point, as response is received in JSON format, no really need to bring the protobuf dependencies that are being pulled from Envoy.
Perhaps in the context of this feature we could do something similar as IstioConfig and use a simple struct model with some interface{} for inner fields.

business/proxy_status.go
```diff
@@ -74,3 +75,50 @@ func xdsStatus(sent, acked string) string {
 	// Since the Nonce changes to uuid, so there is no more any time diff info
 	return "Stale"
 }
+
+func (in *ProxyStatus) GetConfigDump(namespace, pod string) (interface{}, error) {
```
lucasponce, line 79: In general, the business package should try to return types from model package.
I guess this is returning a ConfigDump, I'd add a simple struct in the model package for this.
As it happens for other Istio objects (kubernetes vs model package) in the model package Kiali adds some processing (like adding additional permissions or validations info that it's used in next steps).

So, I guess same reasoning fits here, even if it's kinda duplicated of the kubernetes/envoy.go that you've added, but just to let business package continue returning objects from model package.

business/proxy_status.go
```diff
@@ -74,3 +75,50 @@ func xdsStatus(sent, acked string) string {
...
+
+	dump, err := in.k8s.GetConfigDump(namespace, pod)
+	if err != nil {
+		return nil, err
+	}
+
+	return dump, nil
+}
+
+func (in *ProxyStatus) GetConfigDumpResourceEntries(namespace, pod, resource string) (interface{}, error) {
```
lucasponce, line 92: In this one, I'd split the methods with their own returning type.

Note, it's perfectly ok to have inside of the model.ConfigDump interfaces to not add all nested values in complex structures.

kubernetes/envoy.go
```diff
@@ -0,0 +1,150 @@
+package kubernetes
+
+import "github.com/mitchellh/mapstructure"
```
lucasponce, line 3: I haven't used this library before, how does it work ? Does it parse an interface{} as input into a strong type struct{} ?

kubernetes/istio.go
```diff
@@ -248,6 +252,63 @@ func getStatus(statuses map[string][]byte) ([]*ProxyStatus, error) {
...
+
+	// First try whether the pod exist or not
+	_, err = in.GetPod(namespace, podName)
+	if err != nil {
+		log.Errorf("Couldn't fetch the Pod: %v", err)
+		return nil, err
+	}
+
+	// Create a Port Forwarder
+	f, err := config_dump.NewPortForwarder(in.k8s.CoreV1().RESTClient(), clientConfig, namespace, podName, "localhost", "15000", writer)
```
lucasponce, line 289: I'd try to put the host and portforward post in the config, for two reasons: one for document this, if you don't stop here, you may think this is exposed externally instead to run a port-forward locally, and second, considering that we expose as default on 20001 port, I'd put together all these ports in case that user change 20001 -> 15000 and things don't work as expected.

I know that this is unlikely be a problem, but just to prevent this in the future.

lucasponce:
@xeviknal thanks ! the PR looks good, much better without all those protobuf libraries.
I have a couple of comments for your consideration.

business/proxy_status.go
xeviknal: Sure thing. I guess I've just missed in between all the work done in this PR. I guess that the handlers where the firsts methods I did to play around at the beginning.

I've added a wrapper type in the models layer for the ConfigDump and return it in here. To preserve the consistency across the layers.

business/proxy_status.go
xeviknal: I disagree in the fact to split methods just for returning specific types. I guess that just returning a common interface would be alright. I created an small interface for dealing with the summaries in the `buildDump` method right below. It makes it easier for removing the complexity of managing different type structures to achieve the very same result.

Using that, the method signature is:
`func (in *ProxyStatus) GetConfigDumpResourceEntries(namespace, pod, resource string) (models.ResourceDump, error)`

kubernetes/envoy.go
xeviknal: Yes, it is exactly what you mentioned.
See an example:

```
type Person struct {
    Name   string
    Age    int
    Emails []string
    Extra  map[string]string
}

// This input can come from anywhere, but typically comes from
// something like decoding JSON where we're not quite sure of the
// struct initially.
input := map[string]interface{}{
    "name":   "Mitchell",
    "age":    91,
    "emails": []string{"one", "two", "three"},
    "extra": map[string]string{
        "twitter": "mitchellh",
    },
}

var result Person
err := Decode(input, &result)
if err != nil {
    panic(err)
}

fmt.Printf("%#v", result)
```
Output:
```
Person{Name:"Mitchell", Age:91, Emails:[]string{"one", "two", "three"}, Extra:map[string]string{"twitter":"mitchellh"}}
```

You can see in action in this PR in here: https://github.com/kiali/kiali/pull/3400/files#diff-f27412a2c0f12f79f5d59029c2f5947bee29ebd0673f8ecc4718b6c92afb0316R127-R133
It decodes from `map[string]interface{}` to a `RouteDump` struct.

And after discovering this small lib, the Istio object parsing and the validations code came to my head. Because with this method, we would remove all the nasty type assertion and dealing with map[string]interface{}. You just decode the `interface{}` and then you deal with the object as usual.

One good thing with this lib is that you can skip attributes you don't want. So you don't have to copy the whole structure. You create the struct with the part you only care of. That is really convenient for the envoy structs because you pick the only few field you need.

kubernetes/istio.go
xeviknal: Good point!

xeviknal: Oh wait. This portforward is done in the kiali container, so it is not external and it doesn't depend on user's ports. It just depends on kiali's pod open ports. I'd have conflict with the `20001` port if the port-forwarding was performed in the client side.

I think that there is no need to add this in the config because it wouldn't affect to the user at any way.

kubernetes/istio.go
```diff
@@ -248,6 +252,63 @@ func getStatus(statuses map[string][]byte) ([]*ProxyStatus, error) {
...
+	// Start the forwarding
+	if err := f.Start(); err != nil {
+		return nil, err
+	}
+
+	// Defering the finish of the port-forwarding
+	defer f.Stop()
+
+	// Ready to create a request
+	resp, err := http.Get(fmt.Sprintf("http://localhost:15000%s", path))
```
jotak, line 303: remember, always set a timeout on client calls :)

models/config_dump.go
```diff
@@ -0,0 +1,385 @@
...
+	}
+
+	return fmt.Sprintf("Cluster: %s", r.Route.Cluster)
+}
+
+func getAppDescriptor(chainMatch *kubernetes.FilterChainMatch) string {
+	plainText := []string{"http/1.0", "http/1.1", "h2c"}
+	istioPlainText := []string{"istio", "istio-http/1.0", "istio-http/1.1", "istio-h2"}
+	httpTLS := []string{"http/1.0", "http/1.1", "h2c", "istio-http/1.0", "istio-http/1.1", "istio-h2"}
+	tcpTLS := []string{"istio-peer-exchange", "istio"}
```
jotak, line 221: Where do all these values come from, are they exhaustive or not?

jotak:
One general remark that I'm just thinking about: this feature is going to be quite tied with the versions of Envoy, which often change with new istio releases. I guess it will worth it to test with several istio versions, at least the one that we officially support, to check if this work is compatible with all of them. If they're not all compatible, I guess we can document it as known limitations and more work can be done as a follow-up if we want so ?

lucasponce:
> One general remark that I'm just thinking about: this feature is going to be quite tied with the versions of Envoy, which often change with new istio releases. I guess it will worth it to test with several istio versions, at least the one that we officially support, to check if this work is compatible with all of them. If they're not all compatible, I guess we can document it as known limitations and more work can be done as a follow-up if we want so ?

If anything relevant, it can be added a note on 

https://kiali.io/documentation/v1.27/installation-guide/#_version_compatibility

If things move smoothly on previous Istio version is fine, but I wouldn't expect to add support for them.

models/config_dump.go
xeviknal: It appears to be that yes, they are exhaustive.
I've seen that in the istioctl code: https://github.com/istio/istio/blob/master/istioctl/pkg/writer/envoy/configdump/listener.go#L175

It is useful to have this map in order to give more human readable info.

xeviknal:
@jotak good catch. And I guess that we should make sure to have good coverage in all the testing suit for this (@hhovsepy @skondkar @prachiyadav). Right now, the version compatibility matrix works says that the current and previous version of kiali needs istio 1.8 version. So we are good, no need to check for previous versions.
However, this feature would also work for istio 1.7 and most likely istio 1.6 (not 100% sure though).

xeviknal:
Thanks for the comments. I've pushed changes approaching both your comments.

kubernetes/envoy.go
lucasponce: +1
Yes, I was also exactly to update at some point that parsing we had with the interfaces in other places.

kubernetes/istio.go
lucasponce: But in the same container Kiali is opening the 20001 port, isn't it ? It won't have a conflict if you change this to "20001" ?

util/httputil/http_util.go
```diff
@@ -19,6 +19,10 @@ func HttpMethods() []string {
 		http.MethodDelete, http.MethodConnect, http.MethodOptions, http.MethodTrace}
 }
 
+func HttpGetWithBasicAuth(url string, timeout time.Duration) ([]byte, int, error) {
```
jotak, line 22: hmm why is it called "basic auth", looks like there's no auth here, no?

jotak (commented):
Beside a minor comment code LGTM, I'm going to test before finishing the review

kubernetes/istio.go
xeviknal: you are right. Had it wrong. I've pushed a commit adding the local port in the config.

config/config.go
```diff
@@ -154,6 +154,7 @@ type TracingConfig struct {
 type IstioConfig struct {
 	ComponentStatuses        ComponentStatuses `yaml:"component_status,omitempty"`
 	ConfigMapName            string            `yaml:"config_map_name,omitempty"`
+	EnvoyAdminLocalPort      int               `yaml:"envoy_admin_local_port,omitempty"`
```
jotak, line 157: Small question about the configuration & the additional rights that are required: what if the users don't want to grant this new permissions to Kiali, because they don't want Kiali to sneak in envoy internals?
Do they just have to not grant those permissions, and that's it; an error will be thrown when they click on "show envoy details", but nothing else unexpected happens?

lucasponce: >  what if the users don't want to grant this new permissions to Kiali, because they don't want Kiali to sneak in envoy internals?

I don't think this is optional for users, granularity of permissions should be the namespace.

If they want to manually ungrant permissions not this or other feature, they should expect an error, but I don't think we should support a Kiali "à la carte" with this permission removed.

For example, I see this at same level like workload or pod annotations as we have today.

But I guess it would help if the error message is clear, because I see more probably some problem if user didn't upgrade correctly or something similar.

kubernetes/istio.go
```diff
@@ -248,6 +252,64 @@ func getStatus(statuses map[string][]byte) ([]*ProxyStatus, error) {
...
+
+	// First try whether the pod exist or not
+	_, err = in.GetPod(namespace, podName)
+	if err != nil {
+		log.Errorf("Couldn't fetch the Pod: %v", err)
+		return nil, err
+	}
+
+	// Create a Port Forwarder
+	portMap := fmt.Sprintf("%d:15000", config.Get().ExternalServices.Istio.EnvoyAdminLocalPort)
```
lucasponce, line 289: I understand that here you are mapping the Envoy 15000 container port into Kiali EnvoyAdmingLocalPort.
15000 is fixed, but EnvoyAdminLocalPort for some reason can change, or the 20001 used by Kiali can be changed to 15000.

kubernetes/istio.go
```diff
@@ -248,6 +252,64 @@ func getStatus(statuses map[string][]byte) ([]*ProxyStatus, error) {
...
+	// Start the forwarding
+	if err := f.Start(); err != nil {
+		return nil, err
+	}
+
+	// Defering the finish of the port-forwarding
+	defer f.Stop()
+
+	// Ready to create a request
+	resp, code, err := httputil.HttpGet(fmt.Sprintf("http://localhost:15000%s", path), nil, 10*time.Second)
```
lucasponce, line 305: But here, I guess this query happens in Kiali container, so that 15000 shouldn't be the same from the EnvoyAdminLocalPort ?

lucasponce (approved):
LGTM
All comments approached.

jotak (approved):

hhovsepy:
Post merge verification:
Screenshot from 2020-12-09 13-36-58
Screenshot from 2020-12-09 13-37-16
Screenshot from 2020-12-09 13-37-48

PR #3404, opened 2020-11-07
Remove : from URL link

fai555:
The URL of the diagram contains an extra : that effectively breaks the URL. Removing the : fixed it.


jshaughn (approved):
Thanks!

Issue #3409, opened 2020-11-09
[multi-cluster] Find/resolve the clusters available to a certain control plane


israel-hdez:
In order to have a better understanding about how is the structure of the mesh, one first step is to find/resolve which clusters are available/visible to one specific Istio control plane (CP).

Once we have the data about what clusters are available, we can query the other clusters to fetch for services, workloads, etc to have more complete data about what's available in the mesh.

The scope of this task is to find the clusters that are "connected" to the Istio CP that is "adjacent" to a specific Kiali instance. I'm uncertain, but probably it's possible to reach portions of the mesh not reachable directly by proxying traffic first to other CPs -- it would be out of the scope of this task to find those portions of the mesh with indirect access.

**Describe the solution you'd like**

From all the [different available multi-cluster setups](https://preliminary.istio.io/latest/docs/setup/install/multicluster/) the common pattern to "connect" an Istio CP to another cluster is by creating secrets with credentials to access the API Server of the other cluster. See, for example the [Enable Endpoint Discovery section of the multi-primary/different networks instructions](https://preliminary.istio.io/latest/docs/setup/install/multicluster/multi-primary_multi-network/#enable-endpoint-discovery).

**Describe alternatives you've considered**
It would be better if Istio provides this data rather than trying to resolve by ourselves. This was already asked in a meeting with some Istio people, but it may take time until it arrives.

**Additional context**
It could be good to show in the UI the resolved clusters. A new section (new main navigation menu) is probably needed to show this information. It could be named "Mesh structure" or "Clustering".
 
Related epic: #1833.

israel-hdez:
This is targeting `multi-cluster` feature branches on the repositories.

Front-end PR: https://github.com/kiali/kiali-ui/pull/2000
Back-end PR: https://github.com/kiali/kiali/pull/3413

Both have been merged. Closing this issue.

PR #3413, opened 2020-11-09
[multi-cluster] Resolve clusters available to the adjacent istio control plane

israel-hdez:
Add an endpoint to retrieve the clusters that are visible to the adjacent control plane.

Given that Kiali is doing it's own resolution, there is no guarantee that the resolved list is what Istio is seeing. Because of misconfigurations, Kiali may recognize a different list to what Istio has registered internally. But this is the best effort we can do, for now.

Related #3409
Front-end: kiali/kiali-ui#2000


israel-hdez:
@jshaughn I think this is ready for a review.
Tests are pending.

kubernetes/secret.go
```diff
@@ -39,12 +41,34 @@ type RemoteSecret struct {
...
 		return nil, err
 	}
-	err = yaml.Unmarshal(secretFile, &secret)
+	return ParseRemoteSecretBytes(secretFile)
+}
+
+// GetSecrets returns a list of secrets for a given namespace.
+// If selectorLabels is defined, the list will only contain services matching
+// the specified label selector.
+func (in *K8SClient) GetSecrets(namespace string, labelSelector string) ([]core_v1.Secret, error) {
```
xeviknal, line 54: For me, it seems more natural to have this methods into the `client.go` file because there are all the methods dealing with pure k8s objects. This secret files sounds a bit more a helpers around dealing with secrets.

business/layer.go
```diff
@@ -111,20 +112,21 @@ func SetWithBackends(cf kubernetes.ClientFactory, prom prometheus.ClientInterfac
...
+	temporaryLayer.Namespace = NewNamespaceService(k8s)
 	temporaryLayer.OpenshiftOAuth = OpenshiftOAuthService{k8s: k8s}
-	temporaryLayer.TLS = TLSService{k8s: k8s, businessLayer: temporaryLayer}
-	temporaryLayer.Iter8 = Iter8Service{k8s: k8s, businessLayer: temporaryLayer}
-	temporaryLayer.IstioStatus = IstioStatusService{k8s: k8s}
 	temporaryLayer.ProxyStatus = ProxyStatus{k8s: k8s}
+	temporaryLayer.Svc = SvcService{prom: prom, k8s: k8s, businessLayer: temporaryLayer}
+	temporaryLayer.TLS = TLSService{k8s: k8s, businessLayer: temporaryLayer}
+	temporaryLayer.Validations = IstioValidationsService{k8s: k8s, businessLayer: temporaryLayer}
+	temporaryLayer.Workload = WorkloadService{k8s: k8s, prom: prom, businessLayer: temporaryLayer}
```
xeviknal, line 129: Yep, I think it is safe now to re-sort the services. We had problems in the past with that because of some services including other services instead of the whole temporary layer. So :+1:

xeviknal (commented):
Looks pretty straightforward :+1: There is one question I'd like to hear from @israel-hdez though.

business/clustering.go
```diff
@@ -0,0 +1,160 @@
+package business
+
+import (
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+)
+
+// ClusteringService is a support service for retrieving data about the mesh environment
+// when Istio is installed with multi-cluster enabled.
+type ClusteringService struct {
```
jshaughn, line 10: I wonder if we maybe we should just call this MeshService as a way to query anything we find relevant to a Mesh.

business/clustering.go
```diff
@@ -0,0 +1,160 @@
...
+	k8s           kubernetes.ClientInterface
+	businessLayer *Layer
+}
+
+// MeshCluster holds some metadata about a cluster that is
+// part of the mesh.
+type MeshCluster struct {
```
jshaughn, line 17: Could we just call this Cluster?

business/clustering.go
```diff
@@ -0,0 +1,160 @@
...
+	// ApiEndpoint is the URL where the Kubernetes/Cluster API Server can be contacted
+	ApiEndpoint string `json:"apiEndpoint"`
+
+	// IsHomeCluster specifies if this cluster is hosting Kiali (and the observed Mesh Control Plane)
+	IsHomeCluster bool `json:"isHomeCluster"`
```
jshaughn, line 22: Maybe we could be explicit with this and a few other characterizations of the cluster, like HasKiali, HasControlPlane, HasDataPlane etc?

business/clustering.go
```diff
@@ -0,0 +1,160 @@
...
+
+	allClusters := append(remoteClusters, *myCluster)
+
+	return allClusters, nil
+}
+
+// resolveMyControlPlaneCluster tries to resolve the metadata about the cluster where
+// Kiali is installed. This assumes that the mesh Control Plane is installed in the
+// same cluster as Kiali.
+func (in *ClusteringService) resolveMyControlPlaneCluster() (*MeshCluster, error) {
```
jshaughn, line 54: Again, we could just be explicit, and instead of inferring the Kiali point of view in the API, just resolveKialiControlPlaneCluster().

jshaughn (commented):
Functionally it looks OK to me.  Organizationally, Xavi's comment looks good from a consistency standpoint.  For me I think maybe the terminology could be a cleaner but it is not a blocker.

One thing I wonder is whether we should initially work in a branch as we explore this feature, get feedback, etc.

business/clustering.go
israel-hdez: Yes, I think your suggested renaming is OK, as this may include some things not related to clustering. I'll do it.

kubernetes/secret.go
israel-hdez: I think I'm not sure to understand the structure that good :) because `client.go` for me, looks more like interface definitions and a few helpers related to creating the client. So, I this code would be like an "outlier" if moved to `client.go` :/

Perhaps, it's something else?

business/clustering.go
israel-hdez: Hmm... Naming is hard here...

So, in my env I have two clusters. Both can be marked as HasKiali, HasControlPlane and HasDataPlane, yet only one should be marked with "IsHomeCluster" and the one that gets this flag depends completely on which Kiali instance you are accessing...

So, all three variables HasKiali, HasControlPlane and HasDataPlane are good ideas to have, but this PR is still not finding this data, so I'll not add them yet (but I'll try to add in follow-ups). 

Focusing on "IsHomeCluster"... well... It could be named like `IsHostingThisKialiInstance`???? .... I'm now sure, I don't have ideas of a good naming here. May be the comment is what I should fix?

business/clustering.go
israel-hdez: OK, I'll rename this.

business/clustering.go
jshaughn: I see your point.  A couple of ideas, isHomeKiali, isCurrentKiali, isActiveKiali, isSessionKiali.

israel-hdez:
@jshaughn Alike in the front-end, I did the renamings.

Same here... once I get the approvals, I'll close this PR and open a new one targeting a new branch that we can use to work on the MC epic (and also I'll fix conflicts there).

jshaughn (approved):
@israel-hdez OK, this looks good.  But we need to merge it into a feature branch, not into master.  We can work in a feature branch until we're satisfied with approach, details. etc.

israel-hdez:
@jshaughn OK, I changed the target branch.
Also, I added a new commit adding tests, just in case you want to review the tests.

israel-hdez:
As this is targeting a feature branch, I'm merging. But I still would like to get clarification of @xeviknal comment to move code in a follow-up PR, if needed.

Issue #3410, opened 2020-11-09
Unable to authenticate with OIDC and Google OpenID connect


sandmule:
**General remarks**
We're trying to use google OIDC to authenticate into Kiali without RBAC. It is making the connection to google but is erroring in two places. 

**Describe the bug**
Inspecting the logs shows two errors: 

`[authentication.go:406] Token error: Unauthorized` - Where I believe the issue is coming from
`[authentication.go:688] Not handling OpenId code flow authentication: No nonce code present. Login window timed out.` - Don't believe this is the problem, as it's the flow handler.

I might be way off the mark here but as the first error appears to be coming from authentication.go:L406, I can see on line 399 it is attempting to get the namespaces. Is this with the credentials passed back? If so it won't work as OIDC isn't configured with the cluster due to using AWS managed Kubernetes.

**Versions used**
Kiali: 1.26.0
Istio: 1.7.3
Kubernetes flavour and version: EKS 1.18

**To Reproduce**
1. Configure OAUTH screen in google console
2. populate Kiali config
```
  auth:
    strategy: "openid"
    openid:
      # api_proxy: ''
      # api_proxy_ca_data: ''
      authentication_timeout: 600
      # authorization_endpoint: ''
      client_id: "XXXXXXX.apps.googleusercontent.com"
      disable_rbac: true
      # insecure_skip_verify_tls: false
      issuer_uri: 'https://accounts.google.com'
      scopes:
      - openid
      - email
      - profile
      username_claim: email
```
3. Create the secret inside the cluster
4. Add `X-Forwarded-Port: 443` to the gateway
5. Navigate to kiali endpoint and click on `Log in with OpenID`

**Expected behavior**
I expect to be able to log into kiali

Any help would be much appreciated, I've spent the day going through your codebase and exhausted all ideas.

sandmule:
@israel-hdez I hope you don't mind the tag, but I saw you did a lot of the OIDC work and might be able to share some light on the problem

israel-hdez:
I'll check. I think you are right that line 399 is the problem.
This may have slipped, as I think I had a fully integrated OIDC setup when I was working on this.

If that's the problem, I'll fix ASAP and push for a patch release of 1.26.

sandmule:
Thank you very much for your fast response. More than happy to help with testing if required.

israel-hdez:
@sandmule I have a fix in #3414 .

[...]

PR #3414, opened 2020-11-09
[v1.26] [openid] Fix token validation when RBAC is off

israel-hdez:
Function checkOpenIdSession should skip token validation via fetching namespaces when RBAC is OFF. It's assumed the token will be rejected by the cluster. Also, it's assumed that the Kiali Service Account (which is used in place) has enough privileges.

Related #3410


jshaughn (approved):
LOK.  This change was tested and cnfirmed by the community user reporting the issue.

jshaughn:
@mattmahoneyrh See my approval comment, this fix has been confirmed by the upstream user that reported the problem.

PR #3415, opened 2020-11-09
[openid] Fix token validation when RBAC is off

### LINKED ISSUE #3410: Unable to authenticate with OIDC and Google OpenID connect

israel-hdez:
Function checkOpenIdSession should skip token validation via fetching namespaces. It's assumed the token will be rejected by the cluster. Also, it's assumed that the Kiali Service Account (which is used in place) has enough privileges.

Fixes #3410

Backport to v1.26 is #3414 .

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -396,16 +396,18 @@ func checkOpenIdSession(w http.ResponseWriter, r *http.Request) (int, string) {
 		return http.StatusUnauthorized, ""
 	}
 
-	_, err = business.Namespace.GetNamespaces()
-	if err == nil {
-		// Internal header used to propagate the subject of the request for audit purposes
-		r.Header.Add("Kiali-User", claims.Subject)
-		return http.StatusOK, claims.SessionId
+	if !config.Get().Auth.OpenId.DisableRBAC {
+		// If RBAC is ENABLED, check that the user has privilges on the cluster.
+		_, err = business.Namespace.GetNamespaces()
+		if err != nil {
+			log.Warning("Token error: ", err)
+			return http.StatusUnauthorized, ""
+		}
 	}
 
-	log.Warning("Token error: ", err)
-
-	return http.StatusUnauthorized, ""
+	// Internal header used to propagate the subject of the request for audit purposes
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
LOK.  This change was tested and cnfirmed by the community user reporting the issue.

jshaughn:
@mattmahoneyrh See my approval comment, this fix has been confirmed by the upstream user that reported the problem.

israel-hdez:
Thanks for your review @jshaughn

Issue #3419, opened 2020-11-10
Error in Run kiali with config file


aljesusg:
After last changes the error message is wrong. The error is related that now we can't use the **-v flag**

```bash
aljesusg@localhost:~/Projects/kiali/kiali/src/github.com/kiali/kiali (master)$ /home/aljesusg/Projects/kiali/kiali/bin/go_build_github_com_kiali_kiali -config /home/aljesusg/Projects/kiali/configKiali/config.yaml -v 4
flag provided but not defined: -v
Usage of /home/aljesusg/Projects/kiali/kiali/bin/go_build_github_com_kiali_kiali:
  -config string
    	Path to the YAML configuration file. If not specified, environment variables will be used for configuration.

```

So we need to update the Makefile too in the **run task** https://github.com/kiali/kiali/blob/master/make/Makefile.build.mk#L81

Do we want to remove the verbose option ?

lucasponce:
I can confirm this problem, well workaround is to not use the "v" but I'd need to review how the logging is configured now.

The PR that performs this change is this:

https://github.com/kiali/kiali/pull/3382

jotak:
Yes, looks like you need to set environment "LOG_LEVEL" : https://github.com/kiali/kiali/pull/3382/files#diff-894630a7ed925c7768a861c4465bc9ad393f7937df7cac5c77ab123c03921aeeR123

jmazzitelli:
People run "kiali" standalone on the command line like that??
Yeah, there is now a new logger in place ... no longer use -v.
See: https://github.com/kiali/kiali/issues/2893

aljesusg:
Closed by https://github.com/kiali/kiali/pull/3421

aljesusg:
> People run "kiali" standalone on the command line like that??
> Yeah, there is now a new logger in place ... no longer use -v.
> See: #2893

nope, I had configure this in the IDE and I didn't know why it was failing and I tried to do it manually, but I used the -v :)

suryaoruganti:
I'm trying to run kiali using the operator install and NOT through a CLI. I get the same issue. Any thoughts on how to fix this?

The Operator is installed with `helm upgrade --install kiali-operator kiali/kiali-operator -n tools`
When I install Kiali using the CR, the pod dies with the same error.
`kubectl -n tools apply -f kiali_cr.yaml`
`kubectl -n tools logs kiali-d4fd4f749-hv5qs` gives

> flag provided but not defined: -v
[...]

PR #3421, opened 2020-11-10
use LOG_LEVEL in make run target since -v is invalid now

jmazzitelli:
this should have been part of https://github.com/kiali/kiali/issues/2893
-v no longer valid - set LOG_LEVEL env var.


aljesusg:
This close https://github.com/kiali/kiali/issues/3419

aljesusg (approved):
LGFM

PR #3425, opened 2020-11-11
[hack] clean up some more things hanging around after uninstalling SM 2.x.

jmazzitelli:
The docs here are incomplete: https://docs.openshift.com/container-platform/4.6/service_mesh/v2x/installing-ossm.html#ossm-operatorhub-remove_installing-ossm

Need to delete a couple other things, too.


jshaughn (approved):
LOK

Issue #3422, opened 2020-11-11
Consolidate k-charted into kiali repo


lucasponce:
Proposal discussed here:
https://groups.google.com/g/kiali-dev/c/vI-Kg8gXBgs

jotak:
Hi @jadeyliu939 ,
So our k-charted lib will be moved, integrated into Kiali. Can you just confirm it will have no impact on iter8 (said differently, you are not consuming directly the k-charted lib) ? As far as I know there shouldn't be any impact.

PR #3427, opened 2020-11-12
Incorporate k-charted backend into kiali backend

jotak:
Includes some merge of models & functions that were duplicated.

Fixes https://github.com/kiali/kiali/issues/3422
Frontend PR: https://github.com/kiali/kiali-ui/pull/2002


jotak:
Note, there will also be a frontend PR, but they're technically independent.

jshaughn (approved):
LGTM

lucasponce (approved):
LGTM

Perhaps in future steps some naming can be unified, but it's good to have it in the same repo.

israel-hdez:
@jotak We can archive k-charted repo, right?

lucasponce:
> We can archive k-charted repo, right?

I'd say not yet, or at least until we don't need it for Kiali 1.26.x which is shipped in downstream.
It's unlikely that will need a patch there, but just in case.

israel-hdez:
> I'd say not yet, or at least until we don't need it for Kiali 1.26.x which is shipped in downstream.
> It's unlikely that will need a patch there, but just in case.

Oh! Right, Yes!
Fair enough.

PR #3428, opened 2020-11-12
[hack] Some changes to dev tools (make and hack script)

jmazzitelli:
* Have make cluster-status tell you what version of istio you have installed (if any)
* Also have the istio hack script delete istio-system ns when uninstalling.


jshaughn (approved):
Good, it should delete the ns.  Also, version info is nice. LOK

PR #3432, opened 2020-11-14
need to create NetworkPolicy when deploying dev builds inside a Maistra environment

jmazzitelli:
When the makefile create a Kiali deployment (which is done outside of Maistra/OLM), the Maistra NetworkPolicy forbids access to that Kiali deployment. So we need to create our own NetworkPolicy to enable access to our dev deployments.

fixes: https://github.com/kiali/kiali/issues/3433


jshaughn (approved):
LOK

Issue #3406, opened 2020-11-09
Use client / server timeouts (potential leaks)


jotak:
This is a follow-up on https://github.com/kiali/kiali/issues/2905 : a missing connection timeout was added to fix a specific case.
But there's other missing connection timeouts in our codebase that could result in the same bugs.

I don't think we necessarily have to see the bug happen before fixing it; using client connections without timeout is a known risk [1] that we should tackle via code best practices IMO.

- I identified just one place in the code where it is missing: `versions.go` / `istioVersion()`
- Similarly, server timeouts need to be defined to avoid file descriptors leaks [2]. To fix in `server.go`

[1] https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779
[2] https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

jotak:
Personally I think it would be wise to backport these fixes asap.

lucasponce:
Note that k8s API and Kiali uses a extensive use of watches [1]

I wonder if a generic timeout can solve some scenarios but it can add some problem on these calls.

These are connections that are kept open.

In other words, this issue looks good, but it will need more close review to avoid side effects.

[1] https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes

jotak:
@lucasponce I was not thinking of a generic timeout (not sure if it's even possible), but per client connection, as we do in many of our client calls, there's just one that is missing.

jshaughn:
Probably a separate concern, but we may also want to look at our calls to prometheus.  I think we use only context.Background and that context does not have a timeout and isn't cancelable. We may want to take a look there...

PR #3440, opened 2020-11-18
Add client and server timeouts

jotak:
- Add http client timeout to istio version call
- Add server 30s timeouts

Fixes https://github.com/kiali/kiali/issues/3406

### DIFF
--- a/server/metrics_server.go
+++ b/server/metrics_server.go
@@ -3,6 +3,7 @@ package server
 import (
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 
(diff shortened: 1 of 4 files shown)


jshaughn (approved):
This all looks good to me.   The fixed server timeouts are hopefully OK, I guess we'll see if anyone runs into trouble.  Note that there is a separate effort going on to investigate applying timeouts and/or cancelation to Prom queries.  A server timeout in Kiali may need to coordinate with lower level handling.

jotak:
Yes I've seen this ticket about prom timeouts, to be honest I though we *did already* have timeouts on those. Though I don't really know about context cancellation. Is there a specific reason why you started this investigation, a known impact?

jshaughn:
>  Is there a specific reason why you started this investigation, a known impact?

There was anissue a while back where it was theorized that prom connections were used up.  In the end I don't think that was true, but in the recent context of timeout investigation and perf/scale in general, it seemed like good knowledge to have.  If we are timing out in the server after 30s, yet the prom calls live on, it seems like it could be an issue.

jotak:
DNM: I'd like to wait next week before merging, so that it doesn't get released without being tested for some time.

PR #3444, opened 2020-11-18
[make] no need for basename here

jmazzitelli:


jotak (approved):
LGTM, I did the same locally, makes my volume mounts look better although I don't know if it could cause any harm

PR #3452, opened 2020-11-23
(trivial) Initialize stats array for non-null json

jotak:
This may happen when there's no metrics at all for the target workload (eg. when workload is in CrashLoopBackoff state)


lucasponce (approved):

jotak:
thx @lucasponce

Issue #3453, opened 2020-11-23
data race - istio status scenarios


xeviknal:
Running `make test-race` it randomly gets the following data race:

```
WARNING: DATA RACE
Read at 0x00c000268070 by goroutine 123:
  github.com/kiali/kiali/business.addAddOnRoute.func1()
      /home/jenkins/agent/workspace/kiali-workflow/kiali-release/src/github.com/kiali/kiali/business/istio_status_test.go:366 +0x1fc
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:2042 +0x51
  github.com/gorilla/mux.(*Router).ServeHTTP()
      /home/jenkins/agent/workspace/kiali-workflow/kiali-release/pkg/mod/github.com/gorilla/mux@v1.7.4/mux.go:210 +0x132
  net/http.serverHandler.ServeHTTP()
      /usr/local/go/src/net/http/server.go:2843 +0xca
  net/http.(*conn).serve()
      /usr/local/go/src/net/http/server.go:1925 +0x84c

Previous write at 0x00c000268070 by goroutine 93:
  github.com/kiali/kiali/business.addAddOnRoute.func1()
      /home/jenkins/agent/workspace/kiali-workflow/kiali-release/src/github.com/kiali/kiali/business/istio_status_test.go:366 +0x212
  net/http.HandlerFunc.ServeHTTP()
      /usr/local/go/src/net/http/server.go:2042 +0x51
  github.com/gorilla/mux.(*Router).ServeHTTP()
      /home/jenkins/agent/workspace/kiali-workflow/kiali-release/pkg/mod/github.com/gorilla/mux@v1.7.4/mux.go:210 +0x132
  net/http.serverHandler.ServeHTTP()
      /usr/local/go/src/net/http/server.go:2843 +0xca
  net/http.(*conn).serve()
      /usr/local/go/src/net/http/server.go:1925 +0x84c

Goroutine 123 (running) created at:
  net/http.(*Server).Serve()
      /usr/local/go/src/net/http/server.go:2969 +0x5d3
  net/http/httptest.(*Server).goServe.func1()
      /usr/local/go/src/net/http/httptest/server.go:308 +0xd8

Goroutine 93 (running) created at:
  net/http.(*Server).Serve()
      /usr/local/go/src/net/http/server.go:2969 +0x5d3
  net/http/httptest.(*Server).goServe.func1()
[...]

PR #3454, opened 2020-11-23
Add mutex to prevent test-race in Istio Status test scenario

xeviknal:
fixes https://github.com/kiali/kiali/issues/3453

Adding a mutex to prevent a race condition when accessing to same counter at the same time.

### DIFF
--- a/business/istio_status_test.go
+++ b/business/istio_status_test.go
@@ -361,7 +361,7 @@ func mockServer(mr *mux.Router) *httptest.Server {
 	return httptest.NewServer(mr)
 }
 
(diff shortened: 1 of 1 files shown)


jotak (approved):
LGTM
I guess you could also use the `atomic` package: https://golang.org/pkg/sync/atomic/#AddInt64
which is perhaps simpler (just call `AddInt64(callNum, 1`) 

But +1 anyway for me

xeviknal:
@jotak the thing of using the AddInt64 is that I should change all the vars initialization types, referenced param and assertions. So it is a little bit more tedious. Anyway, thanks for pointing at this method, didn't know it existed.
btw, forgot to use the mutex as pointer. It wouldn't make any effect otherwise.

jotak (approved):

PR #3464, opened 2020-11-24
Add timeouts to prometheus and jaeger calls

### LINKED ISSUE #3436: Add timeouts on Prometheus calls
Related to #3406 and #2905.

Because it is a problem in other cases, we should also verify if there is a reliable and consistent way to add timeouts to Prometheus calls, and possibly check for leaks in the process.

With an structure for timeouts in place, we can then understand how to use it effectively.

cfcosta:
Closes #3436.

Adds timeouts to prometheus. As a bonus, also sets some more strict timeouts and add timeouts to Jaeger :)

### DIFF
--- a/jaeger/client.go
+++ b/jaeger/client.go
@@ -46,7 +46,7 @@ func NewClient(token string) (*Client, error) {
 			return nil, errParse
 		}
 		timeout := time.Duration(5000 * time.Millisecond)
-		transport, err := httputil.CreateTransport(&auth, &http.Transport{}, 10*time.Second)
+		transport, err := httputil.CreateTransport(&auth, &http.Transport{}, timeout)
 		if err != nil {
 			return nil, err
 		}
--- a/jaeger/client.go
+++ b/jaeger/client.go
@@ -46,7 +46,7 @@ func NewClient(token string) (*Client, error) {
 			return nil, errParse
 		}
 		timeout := time.Duration(5000 * time.Millisecond)
-		transport, err := httputil.AuthTransport(&auth, &http.Transport{})
+		transport, err := httputil.CreateTransport(&auth, &http.Transport{}, 10*time.Second)
 		if err != nil {
 			return nil, err
 		}
--- a/prometheus/client.go
+++ b/prometheus/client.go
@@ -81,7 +81,7 @@ func NewClientForConfig(cfg config.PrometheusConfig) (*Client, error) {
 		auth.Token = token
 	}
 
-	transportConfig, err := httputil.AuthTransport(&auth, api.DefaultRoundTripper.(*http.Transport))
+	transportConfig, err := httputil.CreateTransport(&auth, api.DefaultRoundTripper.(*http.Transport), 10*time.Second)
 	if err != nil {
 		return nil, err
 	}
--- a/prometheus/client.go
+++ b/prometheus/client.go
@@ -81,7 +81,7 @@ func NewClientForConfig(cfg config.PrometheusConfig) (*Client, error) {
 		auth.Token = token
 	}
 
-	transportConfig, err := httputil.CreateTransport(&auth, api.DefaultRoundTripper.(*http.Transport), 10*time.Second)
+	transportConfig, err := httputil.CreateTransport(&auth, api.DefaultRoundTripper.(*http.Transport), httputil.DefaultTimeout)
 	if err != nil {
 		return nil, err
 	}
--- a/util/httputil/http_util.go
+++ b/util/httputil/http_util.go
@@ -6,6 +6,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"strconv"
 	"strings"
@@ -24,16 +25,13 @@ func HttpGet(url string, auth *config.Auth, timeout time.Duration) ([]byte, int,
 	if err != nil {
 		return nil, 0, err
 	}
-	var client http.Client
-	if auth != nil {
-		transport, err := AuthTransport(auth, &http.Transport{})
-		if err != nil {
-			return nil, 0, err
-		}
-		client = http.Client{Transport: transport, Timeout: timeout}
-	} else {
-		client = http.Client{Timeout: timeout}
+
+	transport, err := CreateTransport(auth, &http.Transport{}, timeout)
+	if err != nil {
+		return nil, 0, err
 	}
+
+	client := http.Client{Transport: transport, Timeout: timeout}
 
 	resp, err := client.Do(req)
 	if err != nil {
@@ -67,7 +65,19 @@ func newAuthRoundTripper(auth *config.Auth, rt http.RoundTripper) http.RoundTrip
 	}
 }
 
-func AuthTransport(auth *config.Auth, transportConfig *http.Transport) (http.RoundTripper, error) {
+func CreateTransport(auth *config.Auth, transportConfig *http.Transport, timeout time.Duration) (http.RoundTripper, error) {
+	// Limits the time spent establishing a TCP connection if a new one is
+	// needed.
+	transportConfig.Dial = (&net.Dialer{
+		Timeout:   timeout,
+		KeepAlive: timeout,
+	}).Dial
+	transportConfig.IdleConnTimeout = timeout
+
+	if auth == nil {
+		return transportConfig, nil
+	}
+
 	if auth.InsecureSkipVerify || auth.CAFile != "" {
 		var certPool *x509.CertPool
 		if auth.CAFile != "" {
--- a/util/httputil/http_util.go
+++ b/util/httputil/http_util.go
@@ -75,11 +75,12 @@ func newAuthRoundTripper(auth *config.Auth, rt http.RoundTripper) http.RoundTrip
 // some point in the near future.
 func CreateTransport(auth *config.Auth, transportConfig *http.Transport, timeout time.Duration) (http.RoundTripper, error) {
 	// Limits the time spent establishing a TCP connection if a new one is
-	// needed.
-	if transportConfig.Dial == nil && transportConfig.DialContext == nil {
-		transportConfig.Dial = (&net.Dialer{
+	// needed. If DialContext is not set, Dial is used, we only create a new one
+	// if neither is defined.
+	if transportConfig.DialContext == nil && transportConfig.Dial == nil {
+		transportConfig.DialContext = (&net.Dialer{
 			Timeout: timeout,
-		}).Dial
+		}).DialContext
 	}
 
 	transportConfig.IdleConnTimeout = timeout
--- a/util/httputil/http_util.go
+++ b/util/httputil/http_util.go
@@ -76,9 +76,11 @@ func newAuthRoundTripper(auth *config.Auth, rt http.RoundTripper) http.RoundTrip
 func CreateTransport(auth *config.Auth, transportConfig *http.Transport, timeout time.Duration) (http.RoundTripper, error) {
 	// Limits the time spent establishing a TCP connection if a new one is
 	// needed.
-	transportConfig.Dial = (&net.Dialer{
-		Timeout: timeout,
-	}).Dial
+	if transportConfig.Dial == nil && transportConfig.DialContext == nil {
+		transportConfig.Dial = (&net.Dialer{
+			Timeout: timeout,
+		}).Dial
+	}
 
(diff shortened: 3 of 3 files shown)


jaeger/client.go
```diff
@@ -46,7 +46,7 @@ func NewClient(token string) (*Client, error) {
 			return nil, errParse
 		}
 		timeout := time.Duration(5000 * time.Millisecond)
-		transport, err := httputil.AuthTransport(&auth, &http.Transport{})
+		transport, err := httputil.CreateTransport(&auth, &http.Transport{}, 10*time.Second)
```
jotak, line 49: Still referring to this page https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/ (which is becoming my ultimate golang timeout guide :-) )
So this code means we have a client timeout set to 5 seconds (which works for the whole exchange), plus more granular 10 seconds timeouts defined in `CreateTransport` which work for parts of the exchange. I guess these granular ones are useless here, no? (refer to the "Client Timeouts" section of the article, the schema is quite self-explanatory).
Or perhaps, at least if we want this more granular control, the timeout value passed to `CreateTransport` should be less than the 5s defined for the whole?

util/httputil/http_util.go
```diff
@@ -24,17 +25,14 @@ func HttpGet(url string, auth *config.Auth, timeout time.Duration) ([]byte, int,
...
-	if auth != nil {
-		transport, err := AuthTransport(auth, &http.Transport{})
-		if err != nil {
-			return nil, 0, err
-		}
-		client = http.Client{Transport: transport, Timeout: timeout}
-	} else {
-		client = http.Client{Timeout: timeout}
+
+	transport, err := CreateTransport(auth, &http.Transport{}, timeout)
```
jotak, line 29: IIUC the one big difference we have with the Prometheus client, compared to the other client calls we do, is that here the same client is reused for all connections.
So, still referring to this https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/ , what was really missing here I guess was the `http.Transport.IdleConnTimeout` which is what impacts the time spent between two requests when the connection pool is full (i guess). All the rest was already covered by the top-level client timeout.

Anyway, even if it covers more than necessary, +1 to set all granular timeouts as a security (if it hasn't any real drawback, but I'm not aware of any)

jotak (approved):
+1, just perhaps the Jaeger client call could have more consistent timeouts as I commented, but it's nitpicking.

util/httputil/http_util.go
```diff
@@ -67,7 +65,19 @@ func newAuthRoundTripper(auth *config.Auth, rt http.RoundTripper) http.RoundTrip
 	}
 }
 
-func AuthTransport(auth *config.Auth, transportConfig *http.Transport) (http.RoundTripper, error) {
+func CreateTransport(auth *config.Auth, transportConfig *http.Transport, timeout time.Duration) (http.RoundTripper, error) {
+	// Limits the time spent establishing a TCP connection if a new one is
+	// needed.
+	transportConfig.Dial = (&net.Dialer{
+		Timeout:   timeout,
+		KeepAlive: timeout,
```
israel-hdez, line 73: Default value for KeepAlive is 15 seconds.
TBH, I think setting `KeepAlive` and `Timeout` doesn't make much sense. Connection will anyway be terminated while the KeepAlive probe is on it's way. Probably, KeepAlive is more useful in long lasting connections.

So, I think we can omit KeepAlive from this PR and let the default of 15 seconds to take effect if the transport is created with longer timeouts. The KeepAlive probes will detect the broken connection and, hopefully, that will trigger an error.

util/httputil/http_util.go
```diff
@@ -67,7 +65,19 @@ func newAuthRoundTripper(auth *config.Auth, rt http.RoundTripper) http.RoundTrip
...
+	}).Dial
+	transportConfig.IdleConnTimeout = timeout
```
israel-hdez, line 75: Unless the httpClient is reused, I don't think the Idle Timeout should match the Dial Timeout.

I think this is more to enable reusing connections and have a small connection pool. But in Kiali, I think we create a new client every time. So, I think it's better to keep this to a static low value that can be overridden via an optional argument.

This is just a suggestion. If you don't agree, I think it's OK to keep as is.

jaeger/client.go
israel-hdez: After reading the cloudflare post, I agree with @jotak.
This timeout should be, at most, equal to the Client timeout (in line 53: https://github.com/kiali/kiali/pull/3464/files#diff-31753879fb48a9f42f0390e11ebd20f5be310a8a2f66f1061d3dceb3ff23e517R53).

israel-hdez (commented):
I have a few additional comments.
I'm not blocking the PR as you already got an approval from @jotak, which is more involved on these options.

You still need to fix a linting issue :)

util/httputil/http_util.go
israel-hdez: > TBH, I think setting `KeepAlive` and `Timeout` doesn't make much sense.

More context: setting both to the same value :)

util/httputil/http_util.go
jotak: > But in Kiali, I think we create a new client every time. So, I think it's better to keep this to a static low value that can be overridden via an optional argument.

True except precisely for the Prometheus client case, which is reused. I guess this is the main takeaway for this issue: have the idle timeout set for prometheus.

jotak: In the graph, the prometheus client is bound to the request lifetime. Same for a couple of other handlers.
But in parts that use the "business" layer, it's different, see here: https://github.com/kiali/kiali/blob/master/business/layer.go#L87-L99
The prometheus client is global hence re-used.
The jaeger client is created via a loader, hence it is only created on demand.

Without even questioning if it's better or not to re-use a client, I'd say it is safer to define the idle timeout anyway, like @cfcosta did, to be more future-proof.

jaeger/client.go
jotak: (updated my comment above, +1 for the safety net of having idle timout defined even if it's not necessary at the moment)

jshaughn (commented):
@cfcosta I think this is probably helpful stuff and in line with the other work done recently around timeout protection in the API.  Please respond to/resolve the comment threads, I think there are one or two changes to make.

To be honest, I originally thought this issue would result in the use of contextWithTimeout, putting a timeout on the queries themselves in order to more safely release resources in prom if we were to unleash long-running queries.  For example, instead of:
`ctx, cancel := context.WithCancel(context.Background())`
something like:
`ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)`

The only question I have with the current PR is whether a client timeout may leave a long-running prom call which eventually completes (or hits a default query timeout, if there is one), and then have its result tossed.  Not sure if you, @jotak or @israel-hdez have an opinion on this.

util/httputil/http_util.go
israel-hdez: My concern was that if you set a large timeout (say, for example, 2 minutes), when seeing this diagram: https://blog.cloudflare.com/content/images/2016/06/Timeouts-002.png.

what I understand is that the client will wait up to 2 minutes for the request and response to complete. Once that's done, the connection becomes idle and there is another 2 minute wait giving the connection a maximum of 4 minutes of lifespan. This is OK if the connection is going to be reused (which is the case for Prometheus client in Kiali). But if it's not going the be re-used, the connection is allowed to be there idle for 2 minutes.

For the scope of this PR, keeping as is this is quite fine, because we are using short timeouts. So, I'm good and as @jotak said, this is better than nothing. But I would like to, at least, see a comment in the function so that we don't forget that timeouts should be "reasonably short" to prevent any hanging connection in idle state.

prometheus/client.go
```diff
@@ -81,7 +81,7 @@ func NewClientForConfig(cfg config.PrometheusConfig) (*Client, error) {
 		auth.Token = token
 	}
 
-	transportConfig, err := httputil.AuthTransport(&auth, api.DefaultRoundTripper.(*http.Transport))
+	transportConfig, err := httputil.CreateTransport(&auth, api.DefaultRoundTripper.(*http.Transport), 10*time.Second)
```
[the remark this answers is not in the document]
cfcosta: I can do that. In general I maintained the same timeout that we are using everywhere else. I'll move this to a constant.

jaeger/client.go
cfcosta: This is correct, @jshaughn, thanks for catching it.

util/httputil/http_util.go
cfcosta: @israel-hdez I agree with your points, added a comment explaining the problem so people in the future can be aware.

cfcosta:
@jshaughn @jotak @israel-hdez thanks for the great feedback (and for the great Cloudflare link, I was literally reading through the source code/documentation to try to understand exactly what exact timeouts I should use), I added what I think is a good compromise on the points you brought up.

I tried to trigger the "IdleConnectionTimeout" problem that @israel-hdez theorized , but could not replicate the results. Either way, I added some documentation there.

@jshaughn in general, my reasoning was: adding timeouts is usually good practice everywhere, and since we are using the same "transport creation" logic in more than one place, this seemed like a good way to go. What do you think? Do you think it is still better to do the `WithTimeout` on the prometheus client?

To be quite fair it got me thinking both would be a good idea, setting sane defaults for most http requests, and at the same time adding an option for when you have to trigger a long-running request.

util/httputil/http_util.go
jotak: > But if it's not going the be re-used (which is the case for non-p8s connections), the connection is allowed to be there idle for 2 minutes.

I think (hope) it is not the case, when the client is not reused, it is closed, hence any connections that would be idle would be closed immediately as well, I guess

util/httputil/http_util.go
```diff
@@ -67,7 +67,25 @@ func newAuthRoundTripper(auth *config.Auth, rt http.RoundTripper) http.RoundTrip
...
+// Creates a new HTTP Transport with TLS and Timeouts.
+//
+// Please remember that setting long timeouts is not recommended as it can make
+// idle connections stay open for as long as 2 * timeout. This should only be
+// done in cases where you know the request is very likely going to be reused at
+// some point in the near future.
+func CreateTransport(auth *config.Auth, transportConfig *http.Transport, timeout time.Duration) (http.RoundTripper, error) {
+	// Limits the time spent establishing a TCP connection if a new one is
+	// needed.
+	transportConfig.Dial = (&net.Dialer{
```
jotak, line 79: I'm not sure about this, if there's already a dial (dialcontext?) defined in Transport, you are overwriting it. And in the case of prometheus we use the `DefaultRoundTripper` which actually defines a dialer, with KeepAlive: https://github.com/prometheus/client_golang/blob/master/api/client.go#L29-L36

(I'm not sure why there is `Transport.Dial` and `Transport.DialContext`, and how the two of them relate).

Perhaps this should be executed only if Dial and DialContext are null?

jotak (changes requested):
See my comment about the dialer and our use of prom's DefaultRoundTripper

util/httputil/http_util.go
cfcosta: In general I prefer to have it set always, but I do see your point. Added it.

util/httputil/http_util.go
```diff
@@ -76,9 +76,11 @@ func newAuthRoundTripper(auth *config.Auth, rt http.RoundTripper) http.RoundTrip
 func CreateTransport(auth *config.Auth, transportConfig *http.Transport, timeout time.Duration) (http.RoundTripper, error) {
 	// Limits the time spent establishing a TCP connection if a new one is
 	// needed.
-	transportConfig.Dial = (&net.Dialer{
-		Timeout: timeout,
-	}).Dial
+	if transportConfig.Dial == nil && transportConfig.DialContext == nil {
```
jotak, line 79: I'm seeing in my IDE that Dial is deprecated in favour of DialContext, so we can probably just remove Dial here and below, using just DialContext (explained here as well, line 140: https://golang.org/src/net/http/transport.go?s=13608:13676 )

Apart from that, I've tested and it's working fine for me

cfcosta: @jotak should be ok now. I've changed so that we only create a new DialContext if both Dial and DialContext is not set (if it is set somewhere, we shouldn't change it).

jotak (approved):
LGTM, thanks @cfcosta !

israel-hdez (approved):
It's working OK.

Issue #3467, opened 2020-11-25
Fix bypassed queries to kubernetes Services


lucasponce:
There are some queries that bypass the kiali cache and it performs unnecessary calls to the k8s API when the same object is already cached and available.

PR #3474, opened 2020-11-26
Add Service details into kialiCache

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3467

In some cases, fetching the kubernetes service bypassed the kiali cache.

This was unnoticed until we reviewed this WIP work https://github.com/kiali/kiali/pull/3362 that may need to fetch the service info more frequently so, it should use the caching mechanism.

Also adding an informer for Endpoints for consistency.

For QE: just validate that there are no regressions when deploying with operator.


hhovsepy (approved):
Verified.

aljesusg (approved):
LGFM

Issue #3478, opened 2020-11-29
when the name of namespace is started with kube, the kiali cannot graph for it .


zackzhangkai:
**Describe the bug**

when my namespace is started with kube, then kiali cannot graph for it

For UI issues please also add a screenshot that shows the issue.

```
$ k get ns kube-test
NAME        STATUS   AGE
kube-test   Active   149m
```

![image](https://user-images.githubusercontent.com/20178386/100534165-d3f3c580-3246-11eb-875b-86a28c5b2f84.png)

**Versions used**
Kiali:  v1.26.1
Istio: 1.6.10
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9)

jmazzitelli:
This is not a bug. This is working as expected. By default, Kiali will not show any namespaces that start with "kube" because typically those are internal-use only namespaces that are never part of the mesh.

If, for some reason, you want to create namespaces that start with "kube" and put them in your mesh, you can tell Kiali about it by modifying the Kiali CR setting `spec.api.namespaces.exclude` as documented here: https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L95-L114 - in that documented example CR, you can see `"kube.*"` is in the default list of excluded namespaces.

zackzhangkai:
Thanks a lot, it's really helpful for me.

PR #3479, opened 2020-11-29
add tag for exclude namespaces

zackzhangkai:
Signed-off-by: zackzhangkai <zackzhang@yunify.com>

** Describe the change **

Add tags for api configuration

** Issue reference **

https://github.com/kiali/kiali/issues/3478

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -243,12 +243,12 @@ type KubernetesConfig struct {
 
 // ApiConfig contains API specific configuration.
 type ApiConfig struct {
-	Namespaces ApiNamespacesConfig
+	Namespaces ApiNamespacesConfig `yaml:"namespaces,omitempty" json:"namespaces"`
 }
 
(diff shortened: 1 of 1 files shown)


jmazzitelli:
I don't understand why this is needed. This doesn't fix anything with respect to issue #3478 because that issue isn't a bug or anything that needs to be addressed. 

Can you write a github issue that explains the problem that this PR fixes?

zackzhangkai:
Yes, the issue is not really a bug. When I came across the situation as the issue, then I had gone through the codes and found no tags for the API configurations. Then I got confused and I considered the blacklists is not allowed to be modified. Until now, I also don't know why it had no tags, but it really takes effect to config the APIs.

```bash
apiVersion: v1
data:
  config.yaml: |
    api:
      namespaces:
        exclude:
        - kube-system
```

PR #3480, opened 2020-11-29
modify default blacklist

zackzhangkai:
Signed-off-by: zackzhangkai <zackzhang@yunify.com>

** Describe the change **

_Please describe what the code change is about_

Modify the default namespace of the blacklist, maybe filter out all the namespaces started with kube* is not elegant

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -345,7 +345,7 @@ func NewConfig() (c *Config) {
 			Namespaces: ApiNamespacesConfig{
 				Exclude: []string{
 					"istio-operator",
-					"kube.*",
+					"kube-system",
 					"openshift.*",
 					"ibm.*",
 					"kial-operator",


jmazzitelli:
First, there is *a lot* more to do if you want to change a default setting in Kiali. Refer to the large checklist here: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting

Second, I reject this change because there is more than just `kube-system` namespace as internal namespaces (`kube-public` and `kube-node-lease` are two others as documented by [Kubernetes.io](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#viewing-namespaces)). More importantly, Kubernetes documentation explicitly tells users not to create namespaces that start with `kube-` - see https://kubernetes.io/docs/tasks/administer-cluster/namespaces/#creating-a-new-namespace where it says:

```
Note: Avoid creating namespace with prefix kube-, since it is reserved for Kubernetes system namespaces.
```

zackzhangkai:
Thank you a lot. Why not just filter out the kube-*? The default configuration in the kiali is to filter out all the namespaces started with kube, it also included kubexxx, not just kube-xxx. Besides, if not filter out the namespaces, there seems having no effect to anything.

jmazzitelli:
It is true the docs say "kube-" (with the dash) and Kiali excludes all namespaces starting with "kube" (no dash). But normally users do not have namespaces in their mesh starting with "kube" (with or without dash) in order to avoid any confusion with the internal namespaces. Since it is true that virtually all production systems do not have mesh namespaces starting with "kube", Kiali by default will exclude them. You are right that technically Kubernetes docs say "kube-" (with dash) are reserved but the Kiali team has determined we should be excluding names starting with "kube" (no dash) as well.

If you really want a namespace starting with "kube" (with or without the dash) in your mesh for some reason, you are free to do so - you simply have to just configure Kiali to remove that from its list of excluded namespaces.

> Besides, if not filter out the namespaces, there seems having no effect to anything.

I do not know what this is referring to ('having no effect to anything'). If a namespace matches an api.namespaces.exclude regex, it will not show up in the menu. This api.namespaces.exclude setting is documented to have no effect on direct API access to namespaces (as documented here: https://github.com/kiali/kiali-operator/blob/v1.27.0/deploy/kiali/kiali_cr.yaml#L96-L97 where it says, "This _[the `api.namespaces.exclude` setting]_ does not affect explicit namespace access.").

zackzhangkai:
Thanks for your reply. Got it!

Issue #3489, opened 2020-12-01
be able to enable profiler in operator


jmazzitelli:
The operator image now supports running a profiler to help perform performance analysis on the operator (introduced as part of [this performance analysis issue](https://github.com/kiali/kiali/issues/3469)). We want to be able to turn that on from the command line/makefile infrastructure so devs can easily run this profiler and so the molecule tests can optionally produce the profiler report, too.

This will require a change to the operator repo, helm repo, and main kiali repo's make files.

PRs for this issue:
* https://github.com/kiali/kiali/pull/3490
* https://github.com/kiali/kiali-operator/pull/211
* https://github.com/kiali/helm-charts/pull/22

jmazzitelli:
all PRs merged.

PR #3490, opened 2020-12-01
Be able to enable operator profiler from make

jmazzitelli:
Also can turn profiler on when running molecule tests

Part of fix to https://github.com/kiali/kiali/issues/3489


jshaughn (approved):
I was a little confused by MOLECULE_OPERATOR_PROFILER_ENABLED vs OPERATOR_PROFILER_ENABLED, I though the latter needed to be set to the former for the molecult tests, but actually they are just independent things, one for molecule and one for makefile.

PR #3495, opened 2020-12-02
Add proxy to OpenId metadata http requests

### LINKED ISSUE #3451: Ability to set https_proxy on Kiali instances spawned by the operator
**Describe the bug**
My Kiali instance has no direct internet access and so must send any request destined for the internet via proxy servers.
When enabling OIDC Kiali needs to fetch information from the OIDC provider such as .well-known/openid-configuration

I can not find a way to specify a proxy server to use for kiali. I can set ENVs for the operator pod via the helm chart, but I don't see a way in the CR to set ENV variables for the kiali pod. Also another user advises that even with https_proxy env vars Kiali for OIDC retrieval does not honour the env var.

**Versions used**
Kiali: v1.24.0
Istio: 1.7.3
Kubernetes flavour and version: EKS 1.17

**To Reproduce**

```
--- 
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  name: oidc-kiali
  annotations:
    ansible.operator-sdk/verbosity: "1"
spec: 
  auth:
    strategy: openid
    openid:
      client_id: "test-dev-kiali"
      disable_rbac: true
      issuer_uri: "https://sso.dev.mycorp.com"
  deployment: 
    accessible_namespaces: 
      - "**"
    image_version: v1.24.0
    resources: 
      limits: 
        memory: 768Mi
      requests: 
        cpu: 2
        memory: 768Mi
    view_only_mode: true
  external_services: 
    prometheus: 
      component_status: 
        app_label: prometheus-operator
        namespace: prometheus
      url: "http://kube-prometheus-prometheus.prometheus.svc:9090"
  image_name: docker-registry.prod.mycorp.plc/mirrors/quay.io/kiali
  istio_namespace: istio-system
  namespaces: 
    exclude:
```

Browser error;
```
{"error":"Error fetching OpenID provider metadata.","detail":"Get \"https://sso.dev.mycorp.com/.well-known/openid-configuration\": EOF"}
```

**Expected behavior**
Kiali uses a proxy to retrieve content. I guess this could either be done via specifying the ENV vars and Kiali consuming them (http_proxy, https_proxy, no_proxy), or explicitly having a kv in the spec to set one (similar to how prometheus works when setting proxys function.

```
spec: 
  auth:
    strategy: openid
    openid:
      client_id: "mycorp-dev-kiali"
      https_proxy: mycorp.proxy:3128
```

cfcosta:
Closes #3451 .

This PR adds proxy support to OpenId's requests for metadata for the
issuer. This fixes the problem where Kiali requires a proxy to connect
to the outside world. This is not normally a problem, as the resources we
connect using Kiali are usually inside the cluster.

OpenID is the exception, as we need to connect to the issuer to get
information to move the user forward. This PR adds proxy support to this
specific request.

## How to test this?

  1. Install Squid on your machine, or other HTTP proxy.
  2. Install kiali with the correct configuration options (`http_proxy`
  and `https_proxy` on the openid section of the CR, and OpenId configured).
  3. Connect/authenticate on Kiali using OpenID
  4. Check squid log for the request.

### DIFF
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -262,6 +262,25 @@ func GetOpenIdAesSession(r *http.Request) (*config.IanaClaims, error) {
 	return &sessionData, nil
 }
 
+func getProxyForUrl(targetURL *url.URL, httpProxy string, httpsProxy string) func(req *http.Request) (*url.URL, error) {
+	return func(req *http.Request) (*url.URL, error) {
+		var proxyUrl *url.URL
+		var err error
+
+		if httpProxy != "" && targetURL.Scheme == "http" {
+			proxyUrl, err = url.Parse(httpProxy)
+		} else if httpsProxy != "" && targetURL.Scheme == "https" {
+			proxyUrl, err = url.Parse(httpsProxy)
+		}
+
+		if err != nil {
+			return nil, err
+		}
+
+		return proxyUrl, nil
+	}
+}
+
 // GetOpenIdMetadata fetches the OpenId metadata using the configured Issuer URI and
 // downloading the metadata from the well-known path '/.well-known/openid-configuration'. Some
 // validations are performed and the parsed metadata is returned. Since the metadata should be
@@ -284,6 +303,17 @@ func GetOpenIdMetadata() (*OpenIdMetadata, error) {
 			httpTransport.TLSClientConfig = &tls.Config{
 				InsecureSkipVerify: true,
 			}
+		}
+
+		parsedUrl, err := url.Parse(trimmedIssuerUri)
+
+		if err != nil {
+			return nil, err
+		}
+
+		if cfg.HTTPProxy != "" || cfg.HTTPSProxy != "" {
+			proxyFunc := getProxyForUrl(parsedUrl, cfg.HTTPProxy, cfg.HTTPProxy)
+			httpTransport.Proxy = proxyFunc
 		}
 
 		httpClient := http.Client{
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -299,7 +299,7 @@ func GetOpenIdMetadata() (*OpenIdMetadata, error) {
 
 		httpClient, err := createHttpClient(trimmedIssuerUri)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to create http client to fetch OpenId Metadata: %w", err)
 		}
 
 		// Fetch IdP metadata
@@ -383,7 +383,7 @@ func GetOpenIdJwks() (*jose.JSONWebKeySet, error) {
 		// Create HTTP client
 		httpClient, err := createHttpClient(oidcMetadata.JWKSURL)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to create http client to fetch OpenId JWKS document: %w", err)
 		}
 
 		// Fetch Keys document
@@ -458,7 +458,7 @@ func RequestOpenIdToken(openIdParams *OpenIdCallbackParams, redirect_uri string)
 
 	httpClient, err := createHttpClient(openIdMetadata.TokenURL)
 	if err != nil {
-		return err
+		return fmt.Errorf("failure when creating http client to request open id token: %w", err)
 	}
 
 	// Exchange authorization code for a token
@@ -734,7 +734,7 @@ func createHttpClient(toUrl string) (*http.Client, error) {
 	}
 
 	if cfg.HTTPProxy != "" || cfg.HTTPSProxy != "" {
-		proxyFunc := getProxyForUrl(parsedUrl, cfg.HTTPProxy, cfg.HTTPProxy)
+		proxyFunc := getProxyForUrl(parsedUrl, cfg.HTTPProxy, cfg.HTTPSProxy)
 		httpTransport.Proxy = proxyFunc
 	}
 
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -456,14 +456,6 @@ func RequestOpenIdToken(openIdParams *OpenIdCallbackParams, redirect_uri string)
 
 	cfg := config.Get().Auth.OpenId
 
-	// Create HTTP client
-	httpTransport := &http.Transport{}
-	if cfg.InsecureSkipVerifyTLS {
-		httpTransport.TLSClientConfig = &tls.Config{
-			InsecureSkipVerify: true,
-		}
-	}
-
 	httpClient, err := createHttpClient(openIdMetadata.TokenURL)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


business/openid_auth.go
```diff
@@ -286,6 +305,17 @@ func GetOpenIdMetadata() (*OpenIdMetadata, error) {
...
 		}
 
+		parsedUrl, err := url.Parse(trimmedIssuerUri)
+
+		if err != nil {
+			return nil, err
+		}
+
+		if cfg.HTTPProxy != "" || cfg.HTTPSProxy != "" {
+			proxyFunc := getProxyForUrl(parsedUrl, cfg.HTTPProxy, cfg.HTTPProxy)
```
jmazzitelli, line 315:
```suggestion
			proxyFunc := getProxyForUrl(parsedUrl, cfg.HTTPProxy, cfg.HTTPSProxy)
```

jmazzitelli (changes requested):

israel-hdez:
Although I was expecting something making use of Go's standard HTTP_PROXY and HTTPS_PROXY environment variables, this is also quite fine. Each approach have it's pros and cons. So, I'd like to hear from others about their preference about using new configs (like in this PR) vs using Go's standard environment variables.

For an example using Go's standard environment variables, see: https://golang.org/pkg/net/http/#RoundTripper where they set `Proxy: ProxyFromEnvironment`.

@jotak @jshaughn  ^^

cfcosta:
> Although I was expecting something making use of Go's standard HTTP_PROXY and HTTPS_PROXY environment variables, this is also quite fine. Each approach have it's pros and cons. So, I'd like to hear from others about their preference about using new configs (like in this PR) vs using Go's standard environment variables.
> 
> For an example using Go's standard environment variables, see: https://golang.org/pkg/net/http/#RoundTripper where they set `Proxy: ProxyFromEnvironment`.
> 
> @jotak @jshaughn ^^

I actually took a look at this before implementing. The problem of using the environment variables to do this is that EVERY request gets a proxy, even the ones inside the cluster. Given that we should allow proxy only for outside requests/openid, we shouldn't rely on this.

israel-hdez:
> I actually took a look at this before implementing. The problem of using the environment variables to do this is that EVERY request gets a proxy, even the ones inside the cluster. Given that we should allow proxy only for outside requests/openid, we shouldn't rely on this.

Hmm... that's weird. I thought the proxy was going to be skipped for the other cases, because we are creating the transport. In fact, I understand this is the reason why #3451 was reported. I understand the user tried to set the environment variables and they got ignored because we are creating a custom the transport :/

cfcosta:
@israel-hdez yes, and I think ignoring it should still be the correct way to go. This is the only point where we access the outside internet, makes sense to me to have a proper key for it (and maybe in the future support the environment variable for in-cluster requests?).

israel-hdez (changes requested):
Only the `GetOpenIdMetadata` function is using the proxy.
You also need to change:
* `GetOpenIdJwks` function
* `RequestOpenIdToken` function

This also needs operator changes.

cfcosta:
@israel-hdez Operator work is on: https://github.com/kiali/kiali-operator/pull/212. I'll take a look at the other functions.

business/openid_auth.go
```diff
@@ -491,9 +464,9 @@ func RequestOpenIdToken(openIdParams *OpenIdCallbackParams, redirect_uri string)
 		}
 	}
 
-	httpClient := http.Client{
-		Timeout:   time.Second * 10,
-		Transport: httpTransport,
```
israel-hdez, line 466: is the `httpTransport` variable unused now?

cfcosta: Yes, just removed it, thanks!

business/openid_auth.go
```diff
@@ -722,3 +717,31 @@ func parseTimeClaim(claimValue interface{}) (int64, error) {
...
+
+	httpTransport := &http.Transport{}
+	if cfg.InsecureSkipVerifyTLS {
+		httpTransport.TLSClientConfig = &tls.Config{
+			InsecureSkipVerify: true,
+		}
+	}
+
+	if cfg.HTTPProxy != "" || cfg.HTTPSProxy != "" {
+		proxyFunc := getProxyForUrl(parsedUrl, cfg.HTTPProxy, cfg.HTTPProxy)
```
israel-hdez, line 737: Looks like you "restored" the bug about using twice the HTTPProxy variable...

israel-hdez (changes requested):

business/openid_auth.go
```diff
@@ -278,17 +297,9 @@ func GetOpenIdMetadata() (*OpenIdMetadata, error) {
...
-				InsecureSkipVerify: true,
-			}
-		}
-
-		httpClient := http.Client{
-			Timeout:   time.Second * 10,
-			Transport: httpTransport,
+		httpClient, err := createHttpClient(trimmedIssuerUri)
+		if err != nil {
+			return nil, err
```
israel-hdez, line 302: Can you, please, return a descriptive error alike in the other error checks?
Like "cannot create HTTP client to fetch metadata"? (plus the inner error).

business/openid_auth.go
```diff
@@ -370,17 +381,9 @@ func GetOpenIdJwks() (*jose.JSONWebKeySet, error) {
...
-				InsecureSkipVerify: true,
-			}
-		}
-
-		httpClient := http.Client{
-			Timeout:   time.Second * 10,
-			Transport: httpTransport,
+		httpClient, err := createHttpClient(oidcMetadata.JWKSURL)
+		if err != nil {
+			return nil, err
```
israel-hdez, line 386: Also here, a more descriptive error.

business/openid_auth.go
```diff
@@ -453,17 +456,9 @@ func RequestOpenIdToken(openIdParams *OpenIdCallbackParams, redirect_uri string)
...
-			InsecureSkipVerify: true,
-		}
-	}
-
-	httpClient := http.Client{
-		Timeout:   time.Second * 10,
-		Transport: httpTransport,
+	httpClient, err := createHttpClient(openIdMetadata.TokenURL)
+	if err != nil {
+		return err
```
israel-hdez, line 461: ...and here

israel-hdez (changes requested):

business/openid_auth.go
cfcosta: That's weird, not sure what happened there.

cfcosta:
@israel-hdez done, thanks for the review, should be ok now.

israel-hdez:
@cfcosta I'm going to try this.
While I'm on it could you please update the docs? https://kiali.io/documentation/latest/configuration/authentication/openid/

I think we want the docs updated before we do the release cut, so that it's properly published.

israel-hdez (approved):
LGTM

israel-hdez:
Merged, but docs are missing...

jshaughn:
> Merged, but docs are missing...

cc @cfcosta

PR #3498, opened 2020-12-02
get the run-operator-playbook target to run all the way through and with profiler

jmazzitelli:
1. need to install operator-sdk.util via ansible-galaxy because the ansible tasks now use that library
2. also turn on profiler when appropriate

### DIFF
--- a/make/Makefile.operator.mk
+++ b/make/Makefile.operator.mk
@@ -103,9 +103,8 @@ kiali-reload-image: .ensure-oc-exists
 
 ## run-operator-playbook: Run the operator dev playbook to run the operator ansible script locally.
 run-operator-playbook: .ensure-operator-repo-exists .ensure-operator-helm-chart-exists
-	ANSIBLE_ROLES_PATH=${ROOTDIR}/operator/roles ansible-playbook -vvv -i ${ROOTDIR}/operator/dev-hosts ${ROOTDIR}/operator/dev-playbook.yml
-
-## run-operator-playbook-tag: Run a tagged set of tasks via operator dev playbook to run parts of the operator ansible script locally.
-# To use this, add "tags: test" to one or more tasks - those are the tasks that will be run.
-run-operator-playbook-tag: .ensure-operator-repo-exists .ensure-operator-helm-chart-exists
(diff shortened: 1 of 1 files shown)


make/Makefile.operator.mk
```diff
@@ -103,9 +103,8 @@ kiali-reload-image: .ensure-oc-exists
...
 ## run-operator-playbook: Run the operator dev playbook to run the operator ansible script locally.
 run-operator-playbook: .ensure-operator-repo-exists .ensure-operator-helm-chart-exists
-	ANSIBLE_ROLES_PATH=${ROOTDIR}/operator/roles ansible-playbook -vvv -i ${ROOTDIR}/operator/dev-hosts ${ROOTDIR}/operator/dev-playbook.yml
-
-## run-operator-playbook-tag: Run a tagged set of tasks via operator dev playbook to run parts of the operator ansible script locally.
-# To use this, add "tags: test" to one or more tasks - those are the tasks that will be run.
-run-operator-playbook-tag: .ensure-operator-repo-exists .ensure-operator-helm-chart-exists
-	ANSIBLE_ROLES_PATH=${ROOTDIR}/operator/roles ansible-playbook -vvv -i ${ROOTDIR}/operator/dev-hosts ${ROOTDIR}/operator/dev-playbook.yml --tags test
+ifeq ($(OPERATOR_PROFILER_ENABLED),true)
+	@$(eval ANSIBLE_CALLBACK_WHITELIST_ARG ?= ANSIBLE_CALLBACK_WHITELIST=profile_tasks)
```
jshaughn, line 107: Let's change WHITELIST to ACCEPTLIST.

jshaughn (changes requested):

make/Makefile.operator.mk
jmazzitelli: Let's not - this is a ansible env var - that is simply its name, see: https://docs.ansible.com/ansible/latest/reference_appendices/config.html#default-callback-whitelist

It's how you turn on the profiler.

jshaughn (approved):
Ansible needs to get with the times!

Issue #3500, opened 2020-12-03
Building operator/helm fails to "git pull"


jotak:
Probably for @jmazzitelli :

While I'm trying to pull @xeviknal 's work that spans across multiple repos (operator, helm-charts, core and ui), I get an error when running our make targets.

What I'm running:

```
CLUSTER_TYPE=minikube DORP=podman CONSOLE_LOCAL_DIR=/work/swsui CONSOLE_VERSION=local make clean build cluster-push operator-create kiali-create
```

At some point, a `git pull` is performed (see target `.ensure-helm-charts-repo-exists` in `Makefile.helm.mk`). This throws an error:

```
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>
```

Not sure if it's because of my version of git (it's 2.28.0), or because I have several remotes on this repo (as I always do, a `origin` and a `upstream`), but the build stops there. There's no problem when I'm on `master`.

I removed the `git pull` from the makefile and it works well. To be honest, it feels strange to pull code during a local build process, don't you think?

jmazzitelli:
This is making sure your helm charts are up to date.

The solution is simple - just set your upstream in your helm-chart repo's PR branch - something like this:

```
git branch --set-upstream-to=jotak/your-helm-branch your-helm-branch
```

jotak:
hmm that's still a problem. What if I need to get back to an old version and rebuild? It will pull master... that's a problem, no?

jotak:
I mean, I understand the intent to have this as a convenience, but on the other hand it's really something I don't expect, while I'm building from my local files, that the build script will change my git state. Or perhaps there should be two different make targets (or env var), one for helm update and another for untouched helm.

jmazzitelli:
I won't pull master - it only pulls the local branch (i.e. it doesn't checkout anything).

We can remove that `git pull` - I just fully expect people to complain "why isn't my stuff installing correctly?" - but I'll just point to this github issue :)

jmazzitelli:
You know what we could do - instead of pulling, just check to see if you match what it up on the remote repo. If it doesn't match (or there is no upstream repo set), we can spit out a warning message, but continue with the build.

jmazzitelli:
See PR https://github.com/kiali/kiali/pull/3501 that attempts to make this better.

PR #3501, opened 2020-12-03
Provide better flexibility when building helm charts.

jmazzitelli:
Be able to turn off the git pull of the helm-chart repo (via `HELM_CHARTS_REPO_PULL=false`)
If the git pull is turned on (which is the default) but it fails, a better error message is output to the user.

fixes https://github.com/kiali/kiali/issues/3500

There are three scenarios:

1) You want to pull the helm charts by default, and your helm repo has a branch checked out that has a remote upstream branch set up. In this case, the build just succeeds:

```
$ make .ensure-helm-charts-repo-exists
Pulling down latest helm-charts from remote repo to here: /home/jmazzite/source/kiali/helm-charts
```

2) You want to pull the helm charts by default, but your helm chart repo does not have a branch with a valid remote upstream branch. In this case, the build will fail, but you now get a better error message:

```
$ make .ensure-helm-charts-repo-exists
Pulling down latest helm-charts from remote repo to here: /home/jmazzite/source/kiali/helm-charts
X11 forwarding request failed on channel 0
There is no tracking information for the current branch.
Please specify which branch you want to merge with.
See git-pull(1) for details.

    git pull <remote> <branch>

If you wish to set tracking information for this branch you can do so with:

    git branch --set-upstream-to=<remote>/<branch> 2983-operator-sdk-1.0

==========
Failed to update your local helm chart repo.
You can either attempt to correct the error mentioned above,
or set 'HELM_CHARTS_REPO_PULL=false' and try again.
==========
make: *** [.ensure-helm-charts-repo-exists] Error 1
```

3) You DO NOT want to pull the helm charts - either because you want to use the local changes or your helm chart repo does not have a branch with a valid remote upstream branch. In this case, the build will now succeed:

```
$ HELM_CHARTS_REPO_PULL=false make .ensure-helm-charts-repo-exists
Using local helm-charts repo found here: /home/jmazzite/source/kiali/helm-charts
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -127,6 +127,10 @@ else
 KIALI_CR_FILE ?= ${ROOTDIR}/operator/deploy/kiali/kiali_cr_dev.yaml
 endif
 
+# When ensuring the helm chart repo exists, by default the make infrastructure will pull the latest code from git.
+# If you do not want this to happen (i.e. if you want to retain the local copies of your helm charts), set this to false.
+HELM_CHARTS_REPO_PULL ?= true
+
 include make/Makefile.build.mk
 include make/Makefile.container.mk
 include make/Makefile.cluster.mk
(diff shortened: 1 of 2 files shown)


jotak (approved):
LGTM

PR #3509, opened 2020-12-07
need to use the GO env var in makefiles

jmazzitelli:
the Makefile is hardcoding "go" which assumes Go is in the PATH and that's the version of Go to use, which is not a good assumption. This is why we have the GO env var - so that should be used.


jshaughn (approved):
LOK

PR #3510, opened 2020-12-07
[hack] Allow user to dictate what protocol (git or https) to use when git cloning repos.

jmazzitelli:
This may be needed by the QE team.

cc @mattmahoneyrh


jmazzitelli:
@mattmahoneyrh let's get this in, and then when you call the hack script, pass in `-gcp https` and see if it fixes the problem

mattmahoneyrh (approved):

Issue #3311, opened 2020-10-09
Split/Join strategy on Tracing queries


lucasponce:
This is related to EPIC https://github.com/kiali/kiali/issues/2776

In short interval the correlation works well:
![image](https://user-images.githubusercontent.com/1662329/95577605-c71ad880-0a32-11eb-8042-757167f475fa.png)
![image](https://user-images.githubusercontent.com/1662329/95577640-d863e500-0a32-11eb-9f78-c232136324b7.png)

But when we open the interval we find always the following pattern due the nature of tracing info:
![image](https://user-images.githubusercontent.com/1662329/95577699-f598b380-0a32-11eb-836d-7f49a93f2caa.png)

As I can't fetch all, I need to limit the number of results, then on this case using correlation is less effective.

This issue is to discuss a proposal to apply a new pattern on large intervals.

For example, we can "split" the interval and perform several queries (with limits) and then put them together as single result.

Then, I would have a better overview picture of what happened.

This is always configurable, but at some point if I'm selecting 3h check, instead to get the latest 20 traces of that interval (that would always return the latest 20 traces in last 5 minutes) it can split the 3h interval in several queries to give a better picture during that timeline.

Note, this is just an approach with the given API and without using other external mechanisms.

lucasponce:
cc @aljesusg @jotak

jotak:
Good point, I often have the same frustration when looking back in past.

I didn't think about a split/join strategy but it is definitely interesting, yes.

A different possibility (or complementary) could be to load more data as a kind of background job: continuing to populate the chart after the initial load, with small batches. Perhaps with a couple of control buttons, to start and stop loading more traces.

Issue #2776, opened 2020-05-13
Tracing enhancements and correlations


jotak:
Grouping here several tasks of different sizes around tracing & correlation (but each one is self-contained)

- [x] In-graph tracing #2677 (size XL?)
- [x] Add trace-based information in Health #2707 (size L?) (**Closed**)
- [x] Metrics page: more details & links from spans #2775 (size M?)
- [x] Tracing tab in service view #3071
- [x] Tracing tab in workload view #2710
- [X] _Warning check by service #2708_ (**Removed**)
- [ ] Correlation Dashboard #2777 (also related to #1981 and #2708 **closed**) (size XXL?)
- [x] Bucketed display when too many traces/spans #2198 (size L?)
- [x] Traces on workloads/services #3071 , #3048
- [x] Tracing details fix and enhancements #3095
  - [x] Follow-up: Use metrics-based comparison #3293
  - [x] Follow-up: Span filters toolbar #3278
- [ ] Trace mini-graph #3182
- [x] Comparison map per span #3293
- [x] Comparison map per trace #3279
- [ ] Correlate tracing with logging

Also, maintenance work / pre-requisite:
- [x] #2732 (done @jotak)

jotak:
I've added a new task under this epic: #3279 (showing traces heatmap)
Once this is done I want to write a comprehensive blog post to recap all the tracing things done in Kiali since my previous post on that subject [1] which is almost one year old.

But then perhaps it is also time to close this epic? Because it's kind of a never-ending thing, we will keep improving traces anyway (with or without the epic), at the moment I feel it's more of a burden to keep maintaining a sub-tasks list. It would be actually simpler to manage, I think, if we could label tasks with "areas"; like a label "area/tracing" for instance. And for us easier to retrieve where we stand in a specific area in terms of backlog. Thoughts? (@lucasponce @jshaughn )

[1] https://medium.com/kialiproject/tracing-in-kiali-correlation-with-metrics-6e7c1cc67e2a

lucasponce:
> But then perhaps it is also time to close this epic? Because it's kind of a never-ending thing, we will keep improving traces anyway (with or without the epic), at the moment I feel it's more of a burden to keep maintaining a sub-tasks list. It would be actually simpler to manage, I think, if we could label tasks with "areas"; like a label "area/tracing" for instance. And for us easier to retrieve where we stand in a specific area in terms of backlog. Thoughts? (@lucasponce @jshaughn )

[...]

PR #3513, opened 2020-12-09
Traces fetching: new "split and join" strategy

jotak:
Fixes https://github.com/kiali/kiali/issues/3311

Capture d’écran de 2020-12-09 09-07-33

### DIFF
--- a/business/jaeger.go
+++ b/business/jaeger.go
@@ -1,9 +1,8 @@
 package business
 
 import (
-	"fmt"
-	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	jaegerModels "github.com/jaegertracing/jaeger/model/json"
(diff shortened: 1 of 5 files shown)


jotak:
Some remarks:

- This strategy works both for traces & spans fetching (so, tracing page and metrics page w/ correlation), and for all target kinds (apps/workloads/services)

- The "split-join" strategy might still not be sufficient to see traces spread all over the time, when a lot of traces are generated. At the moment I have an arbitrary division of 10 intervals slices on which traces are going to be fetched. These divisions might still be too coarse-grained in high volume to really see a smooth distribution of traces. We could increase the number of intervals to fetch, but it increases the number of queries performed against Jaeger, which could cause performance penalty I fear. Perhaps we could do that if we adopt the jaeger protobuf client? (which we should probably do at some point anyway, I guess...)

- I removed the previous "find relevant spans" algorithm that we had (it was used only when fetching spans for correlation with metrics, not for the tracing page). These two algorithms are somehow competing in providing more spans when the query limit was reached. It can be debated, perhaps we can also imagine a mix of the two algorithms...

jotak:
By the way, one convenient side effect of this change is that the "adjust time" checkbox is now pretty useless since data will often be spread all over the timeline. One display setting that we could remove.

lucasponce:
Thanks @jotak, I'm catching up with some pending reviews but this one is interesting, thanks for implementing.
I'll review it locally.

hhovsepy (approved):
Regression tested for Traces and Metrics.

business/jaeger.go
```diff
@@ -99,7 +102,23 @@ func (in *JaegerService) GetAppTraces(ns, app string, query models.TracingQuery)
 	if err != nil {
 		return nil, err
 	}
-	return client.GetAppTraces(ns, app, query)
+	r, err := client.GetAppTraces(ns, app, query)
+	if err != nil {
+		return nil, err
+	}
+	if len(r.Data) == query.Limit {
```
lucasponce, line 109: Ok, so, if I understand the implementation, first it fetches until reach the limit (i.e. 20 traces).
If limit is reached, then it performs the split and try to query 10 times over the given interval, dividing the limit (20 traces) with the slices (10).

Ok, so, in that case, when I have a query with 20 limit, it'll return 40 results.

Isn't it ?

Or perhaps we can just query the very same limit on the slices, and allow the user to personalize the slices as a parameter.

In any case, it sounds ok as a first iteration, I know it's not an easy topic.

jotak: > Ok, so, if I understand the implementation, first it fetches until reach the limit (i.e. 20 traces).
> If limit is reached, then it performs the split and try to query 10 times over the given interval, dividing the limit (20 traces) with the slices (10).
> 
> Ok, so, in that case, when I have a query with 20 limit, it'll return 40 results.
> 
> Isn't it ?

Exactly

> 
> Or perhaps we can just query the very same limit on the slices, and allow the user to personalize the slices as a parameter.

Yes, perhaps. That should go with changes in the UI too, then. Perhaps we can think about that as part of our discussion about the toolbar redesign? ( here https://github.com/kiali/kiali/issues/2776 )

> 
> In any case, it sounds ok as a first iteration, I know it's not an easy topic.

lucasponce: > Yes, perhaps. That should go with changes in the UI too, then. Perhaps we can think about that as part of our discussion about the toolbar redesign? ( here #2776 )

Totally, it makes sense.

lucasponce (approved):

PR #3517, opened 2020-12-09
Governance file + Remove outdated info

lucasponce:
Added a first governance file with a better maintainers description.

Removed a couple of outdated files:
- ADOPTERS.adoc
- DEMOS.adoc

Note, for demos the idea is to consolidate internal/external demos under the repo:
https://github.com/kiali/demos


GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+A maintainer is a contributor to the Kiali project's success and a citizen helping the project succeed.
+
+## Becoming a Maintainer
+
+To become a maintainer you need to demonstrate the following:
+
+  * commitment to the project
+    * participate in discussions, contributions, code reviews for 3 months or more,
+    * perform code reviews for 10 non-trivial pull requests,
+    * contribute 10 non-trivial pull requests and have them merged into master,
```
abonas, line 21: I think the term "master" should no longer be used. Probably "main"?

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+  * ability to write good code,
```
abonas, line 22: I wonder if "good" is the right term here, "good" is subjective. Perhaps something about well structured, clear, maintainable, documented and tested?

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+They will be given write permission to the Kiali GitHub repositories https://github.com/kiali.
+
+## Changes in Governance
+
+All changes in Governance require a 2/3 majority vote by maintainers.
+
+## Other Changes
+
+Unless specified above, all other changes to the project require a 2/3 majority vote by maintainers.
+Additionally, any maintainer may request that any change require a 2/3 majority vote by maintainers.
```
abonas, line 54: What is the difference between line 53 and 54?

GOVERNANCE.md
lucasponce: Without enter in discussions, for the case of Kiali, the project continues using the "master" branch.
So, I'd add that expression "master branch" to avoid any potential context misunderstanding, but let's not use a different rename of the branch that may cause confusion.

GOVERNANCE.md
lucasponce: To me "good" is as good and subjective as "structured, clear, maintainable, documented and tested".
Both expressions are similar and subjective.

GOVERNANCE.md
lucasponce: I'm not an expert and I basically re-used the same form used by Jaeger.

My interpretation is:

- Line 53 scope refers to the explicit terms discussed previously.
- Line 54 scope refers to any new thing that is not explicitly mentioned as a guard.

abonas: OK :)

GOVERNANCE.md
abonas: I thought that github is renaming master to main for projects, but I guess I misunderstood

lucasponce: The status is that for new repos, github create a "main" branch instead of "master" branch.
For existing projects is untouched.

As a potential migration may have more implications, my suggestion is to use the "master branch" wording just for this doc and with that context.

If in the future we trigger a migration (we have several repos and it's not a transparent process) we can update the doc as well (just to refer to the same branch we use).

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+  * nominee's first and last name,
+  * nominee's email address and GitHub user name,
+  * an explanation of why the nominee should be a committer,
+  * a list of links to non-trivial pull requests (top 10) authored by the nominee.
+
+Two other maintainers need to second the nomination. If no one objects in 5 working days (U.S.), the nomination is accepted.  If anyone objects or wants more information, the maintainers discuss and usually come to a consensus (within the 5 working days). If issues can't be resolved, there's a simple majority vote among current maintainers.
+
+## Changes in Maintainership
+
+Maintainers can be removed by a 2/3 majority vote by maintainers.
```
jotak, line 40: Question: when the maintainer agrees to leave (e.g. doesn't have time for the job) or is self-resigning, perhaps we don't have to go through a vote?

GOVERNANCE.md
abonas: if the branch keeps being named master then the governance should certainly match it. as I said, my understanding around the renaming was not accurate.

GOVERNANCE.md
lucasponce: To me that situation may fall on this case as well.
There is some maintainer that would like to drop for whatever reason, then the group comment on that before.

Note, in 99% of the cases these votes don't need to be "formal", just when proposing some action, if someone desagrees and there is no unanimous decision, then it can be setup a formal vote.

I think.

jotak: The way it's written here, it sounds very formal. If we don't like the way it's written, then we should rewrite. In my (very small) experience of OSS with open governance, votes are systematic, even if it always ends up in an unanimous "+1".

lucasponce: I don't understand.
There will be a vote.
Another topic is how that vote is.

An email can work as a vote, or a thumbup or a differente mechanism, this doc shouldn't force about the how, but about the what.

What I understand is that there is a no consensus of that minimum amount of maintainers, an actions shouldn't go there.

How we setup the votes is a different discussion, in my opinion.

jpkrohling: We didn't have the case for a maintainer being removed yet, but I would expect a vote to happen, even if just a symbolic one.

jotak: @jpkrohling yes I think that's the "spirit" of this kind of governance.
Though for Kiali this is a bit weird given the number of maintainers being all from the same team (at least at the moment)

lucasponce:
I've sent a followup commit to address some comments.

Also note that this is a first version and we can discuss, modify and adapt to Kiali's needs on future iterations.

The goal of this PR in combination with the teams update is

- Provide a better structure of Kiali teams (https://github.com/orgs/kiali/teams).
- Provide an initial description of the roles used.

Please, let me know if this is in good status to merge or there is any concernt to address before we got it merged.

Thanks for all the comments.

lucasponce:
Is there any final comment before that we move forward with this file ?

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
+# Kiali Governance
+
+This document defines governance policies for the Kiali project.
+
+## Maintainers
+
+Kiali Maintainers have write access to the Kiali GitHub repositories https://github.com/kiali.
+They can merge their own patches or patches from others. The current maintainers can be found in https://github.com/orgs/kiali/teams/maintainers.
+
+This privilege is granted with some expectation of responsibility: maintainers are people who care about the Kiali project and want to help it grow and improve. A maintainer is not just someone who can make changes, but someone who has demonstrated his or her ability to collaborate with the team, get the most knowledgeable people to review code, contribute high-quality code, and follow through to fix issues (in code or tests).
```
jshaughn, line 10: Although it's OK as is, the wording is a little verbose.  Perhaps:

This privilege is granted with an expectation of responsibility: maintainers care about the Kiali project and want to help it grow and improve.  Above the ability to contribute changes, a maintainer has demonstrated the ability to collaborate well with the team, assign appropriate code reviewers, contribute high-quality code, and be thorough and timely with fixes, tests and documentation.

GOVERNANCE.md
jshaughn: +1 I'd change to, "the ability to write high quality code."

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+
+To become a maintainer you need to demonstrate the following:
+
+  * commitment to the project
+    * participate in discussions, contributions, code reviews for 3 months or more,
+    * perform code reviews for 10 non-trivial pull requests,
+    * contribute 10 non-trivial pull requests and have them merged into project branches,
+  * ability to write good code,
+  * ability to collaborate with the team,
+  * understanding of how the team works (policies, processes for testing and code review, etc),
```
jshaughn, line 24: understanding of team policies, processes,  etc.

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+  * understanding of the projects' code base and coding style.
```
jshaughn, line 25: project's

GOVERNANCE.md
```diff
@@ -0,0 +1,54 @@
...
+
+A new maintainer must be proposed by an existing maintainer by sending a message to the
+[kiali-dev@googlegroups.com](https://groups.google.com/forum/#!forum/kiali-dev)
+mailing list containing the following information:
+
+  * nominee's first and last name,
+  * nominee's email address and GitHub user name,
+  * an explanation of why the nominee should be a committer,
```
jshaughn, line 33: maintainer or committer?

GOVERNANCE.md
jshaughn: +1, "Maintainers can be removed by a 2/3 majority vote by maintainers, or by resignation."

jshaughn (changes requested):
This will be a good addition.  A few suggested changes...

jshaughn (approved):
LGTM!

PR #3521, opened 2020-12-10
[hack] when installing istio, do not pull images from docker.io anymore

jmazzitelli:
See: https://istio.io/latest/blog/2020/docker-rate-limit/#mitigations


jshaughn (approved):
LOK

PR #3525, opened 2020-12-11
create the helm charts so people don't have to remember to do it when running tests

jmazzitelli:
when running molecule tests, get make to build the updated helm chart so devs don't have to remember to do it.


jshaughn (approved):
Anything to make my life easier.

Issue #3502, opened 2020-12-03
Migrate Travis CI to Github Actions


mtho11:
**Is your feature request related to a problem? Please describe.**
We are currently facing issues with Travis CI (slow and go through our credits quickly)

**Describe the solution you'd like**
A clear option to try out is Github Actions for CI/CD. So this task would be to migrate our CI/CD to use Github Actions instead.

https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/migrating-from-travis-ci-to-github-actions

- [x] _Merged_ Add Github Actions CI to Kiali-ui repo
- [ ] _Review_ Add Github Actions CI to Kiali repo

lucasponce:
This is a blocker task.
What's the status ?

Issue #3518, opened 2020-12-09
race condition in new http_util.CreateTransport function


israel-hdez:
There is a test in the back-end that is failing in some runs. Logs are following:

```
make test-race
make[1]: Entering directory '/home/fedora/workspace/kiali-workflow/kiali-release/src/github.com/kiali/kiali'
Running tests with race detection, excluding third party tests under vendor
[...]

PR #3531, opened 2020-12-15
Add Github Actions CI to Kiali repo

mtho11:
** Describe the change **
This PR runs the CI tests on each push or PR. Effectively doing what Travis CI is doing today but with Github Actions instead.

** Issue reference **
fixes: https://github.com/kiali/kiali/issues/3502

** Backwards incompatible? **
Yes

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -13,17 +13,23 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
 
-    - name: Set up Go 1.x
-      uses: actions/setup-go@v2
-      with:
-        go-version: 1.14
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.14
 
-    - name: Check out code into the Go module directory
-      uses: actions/checkout@v2
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
 
-    - name: Build
-      run: make build 
+      - name: Swagger Install
+        run: make swagger-install
 
-    - name: Test
-      run: make test
+      - name: Lint Install
+        run: make lint-install
+
+      - name: Build
+        run: make clean build
+
+      - name: Test
+        run: make test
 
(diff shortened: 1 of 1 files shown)


mtho11:
Once we are happy with these CI actions (maybe after a week?) we can add more branches to them for the various versions that we support.

.github/workflows/kiali-ci.yml
```diff
@@ -0,0 +1,29 @@
...
+        go-version: 1.14
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2
+
+    - name: Build
+      run: make build 
+
+    - name: Test
+      run: make test
```
lucasponce, line 28: This is very simplified CI.

I guess we have the following in travis:

```
 - make swagger-install
 - make lint-install

script:
  - if [ ! -z "$(gofmt -l ${GO_FILES})" ]; then echo "These files need to be formatted:" "$(gofmt -l ${GO_FILES})";echo "Diff files:"; gofmt -d ${GO_FILES}; exit 1; fi # Gofmt Linter
  - make lint
  - make swagger-travis
  - make clean build test-race
```

Perhaps the goformatted may be packaged as an aux make target, but probably the others steps can be added in the CI in a similar was as we did for travis.

mtho11:
Things seem to be working well. 
Once the race condition tests are fixed: https://github.com/kiali/kiali/issues/3518 then we can run the `test-race` again. For now, it is just running `test` target.

lucasponce:
As commented, I think we are missing the check:

```
  - if [ ! -z "$(gofmt -l ${GO_FILES})" ]; then echo "These files need to be formatted:" "$(gofmt -l ${GO_FILES})";echo "Diff files:"; gofmt -d ${GO_FILES}; exit 1; fi # Gofmt Linter
```
We had in travis.

perhaps that can be added inside a make target and added to the action.

I'm fine to enable the race test later, but I'd try to add the gofmt ones as those are very useful.

mtho11:
@lucasponce it properly fails the build if the go files are not properly formatted

lucasponce (approved):

Issue #3508, opened 2020-12-07
All apps are degraded due to unsynced proxies


HazemElAgaty:
**General remarks**

**Describe the bug**
All my apps are degraded due to unsynced proxies, although `istioctl proxy-status` shows all proxies are synced.

![screenshot2](https://user-images.githubusercontent.com/17452220/101379810-5fabc880-38bd-11eb-93db-9bf9573be179.png)

**Versions used**
Kiali: 1.27.0
Istio: 1.7.3
Kubernetes: 1.16.13

**Installation**
Installed using helmfile
```yaml
  - name: kiali-operator
    namespace: kiali-operator
    chart: kiali/kiali-operator
    version: 1.27.0
    values:
      - cr:
          create: false
          namespace: istio-system
          spec:
            deployment:
              accessible_namespaces:
              - '**'
```
**Kiali CR**
`kubectl get kiali -n istio-system kiali -o yaml`>
```yaml
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  annotations:
    ansible.operator-sdk/verbosity: "1"
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"kiali.io/v1alpha1","kind":"Kiali","metadata":{"annotations":{"ansible.operator-sdk/verbosity":"1"},"name":"kiali","namespace":"istio-system"},"spec":{"deployment":{"accessible_namespaces":["**"],"namespace":"kiali-operator"},"external_services":{"istio_namespace":"istio-system","prometheus":{"url":"http://prometheus-chart-server.monitoring.svc.cluster.local"}}}}
  creationTimestamp: "2020-12-07T15:06:25Z"
  finalizers:
  - finalizer.kiali
  generation: 1
  name: kiali
  namespace: istio-system
  resourceVersion: "52180939"
  selfLink: /apis/kiali.io/v1alpha1/namespaces/istio-system/kialis/kiali
  uid: 3fb8a2a9-f634-4dbb-b419-260b12e42e44
spec:
  deployment:
    accessible_namespaces:
    - '**'
    namespace: kiali-operator
  external_services:
    istio_namespace: istio-system
    prometheus:
      url: http://prometheus-chart-server.monitoring.svc.cluster.local
status:
  accessibleNamespaces:
  - '**'
  conditions:
  - ansibleResult:
      changed: 1
      completion: 2020-12-07T15:08:55.280506
      failures: 0
      ok: 83
      skipped: 76
[...]

PR #3534, opened 2020-12-16
Add a notification when kiali can't find any istio component

xeviknal:
This change tries to approach [this comment](https://github.com/kiali/kiali/issues/3508#issuecomment-742617476).

The main point for them was:

1. They misplaced the `istio_namespace` config in the CR.
2. They correctly placed a different namespace for the `kiali` deployment (through: `spec.deployment.namespace`).
3. The operator set `istio_namespace` to `spec.deployment.namespace` (see [operator code](https://github.com/kiali/kiali-operator/blob/22e89317e94dc22667fe00ac98c5fdccda130b66/roles/v1.0/kiali-deploy/tasks/main.yml#L38-L47))
4. The istio components feature couldn't find any istio components (istiod, istio-ingress/egress).
5. The proxy status in health couldn't find any `istiod` pod to fetch the syncs.

In order to leverage attention in users minds, I've placed a notification like the following triggered by the Istio Status component. It is triggered when Kiali isn't able to find any Istio components (ingress/egress/istiod).

Screenshot of Kiali Console (2)

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -1,6 +1,8 @@
 package business
 
 import (
+	"errors"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


jmazzitelli (dismissed):
I like this.

business/istio_status.go
```diff
@@ -198,6 +202,14 @@ func (iss *IstioStatusService) getStatusOf(ds []apps_v1.Deployment) (IstioCompon
 		}
 	}
 
+	// When all the deployments are missing,
+	// Warn users that their kiali config might be wrong
+	if componentNotFound == len(statusComponents) {
+		return isc, errors.New(fmt.Sprintf(
+			"Are you sure Istio is installed in the %s namespace?",
```
jshaughn, line 209: This message may lead a user to think that Istio *should be* installed in the %s namespace.    I think we are trying to warn them that this is where we are unsuccessfully looking for Istio.  Maybe something like this would be more clear?:

`Kiali is unable to find any Istio deployment in namespace %s.  Are you sure the Istio namespace is configured correctly in Kiali?`

jshaughn (approved):
This could be a nice help to users.  It looks good, I have one optional suggestion for the actual message.  I'm approving as is.

xeviknal:
Message changed:
Screenshot of Kiali Console (3)
 
Thank you all!

jshaughn (approved):
LG!

Issue #3305, opened 2020-10-08
"View in Grafana" link can be wrong


kjetillilleby:
**Describe the bug**
I use the kiali-operator for installation, and have specified (among other thing)
```yaml
spec:
  external_services:
    grafana:
      url: "/system/grafana"
```
 

For UI issues please also add a screenshot that shows the issue.
![image](https://user-images.githubusercontent.com/9693382/95454265-822c6e80-096c-11eb-9c7c-6260f2e67cc7.png)

**Versions used**
Kiali: 1.24
Istio: 1.7
Kubernetes flavour and version: v1.18.6-gke.4801

**To Reproduce**
1. Go to 'Workloads -> Inbound Metrics'
2. Inspect the url on the "View in Grafana" link
3. See error: url starts with `/system/grafana/system/grafana`

**Expected behavior**
url to start with `/system/grafana`

jotak:
I will look into it, thanks for reporting.

jotak:
I don't reproduce the issue. I've set `system/grafana` in CR, which ends up as expected as `url: /system/grafana/d/LJ_uJAvmk/istio-service-dashboard` in the json fetched from UI.

Could you check kiali configmap and see if it mismatches with what you defined in the CR? (Kiali configmap is in the istio-system by default)

kjetillilleby:
Nothing suspicious in the kiali configMap. My installation of grafana is still bundled to istio, Not sure if that is important.
My kiali configMap:
```yaml
apiVersion: v1
data:
  config.yaml: |
    additional_display_details:
    - annotation: kiali.io/api-spec
      icon_annotation: kiali.io/api-type
      title: API Documentation
    api:
      namespaces:
        exclude:
        - istio-operator
        - kube.*
        - openshift.*
        - ibm.*
        - kiali-operator
        label_selector: kiali.io/member-of=istio-system
    auth:
      openid:
        api_proxy: ''
        api_proxy_ca_data: ''
        authentication_timeout: 300
        authorization_endpoint: ''
        client_id: ''
        insecure_skip_verify_tls: false
        issuer_uri: ''
        scopes:
        - openid
        - profile
        - email
        username_claim: sub
      openshift:
        client_id_prefix: kiali
      strategy: anonymous
    deployment:
      accessible_namespaces:
      - istio-system
      - fa-service
      - default
      - pact
      affinity:
        node: {}
        pod: {}
        pod_anti: {}
      custom_dashboards:
        excludes:
        - ''
        includes:
[...]

PR #3536, opened 2020-12-17
Fix grafana links when configured with absolute URLs

jotak:
Fixes https://github.com/kiali/kiali/issues/3305

Also there was some duplicated code that I left after the k-charted
merge, in handlers/grafana.go ; it's now removed, and some pieces moved
to business package like other grafana stuff.

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -181,7 +181,7 @@ func (in *DashboardsService) resolveReferences(namespace string, dashboard *v1al
 }
 
 // GetDashboard returns a dashboard filled-in with target data
-func (in *DashboardsService) GetDashboard(params models.DashboardQuery, template string) (*models.MonitoringDashboard, error) {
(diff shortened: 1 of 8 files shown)


jotak:
Note: to reproduce the initial issue, just edit the grafana deployment to add the following environment variables:

```
        - name: GF_SERVER_SERVE_FROM_SUB_PATH
          value: "true"
        - name: GF_SERVER_ROOT_URL
          value: /system/grafana/
```

Then edit kiali cm/cr grafana config to set: `url: '/system/grafana/'`

Before this patch, base URL part was doubled in the grafana links: `/system/grafana/system/grafana/d/etc`
Now we must have: `/system/grafana/d/etc`

hhovsepy (approved):
Verified.

aljesusg (approved):
LGFM

Issue #3543, opened 2020-12-21
All apps are degraded due to unsynced proxies when using different istio labels than app/version


xeviknal:
>I have the same issue[1] because I changed the istio_labels config parameter to use app.kubernetes.io/name instead of app for the app label key.
>
>The issue occurs because Kiali now expects the istiod pod to have the app.kubernetes.io/name=istiod label. However, istiod is deployed with app=istiod and there is no way to change that configuration using the Istio operator.

from https://github.com/kiali/kiali/issues/3508#issuecomment-748057812

[1] All apps are degraded due to unsynced proxies

xeviknal:
@jshaughn you mentioned that this is an issue for the istio operator. What do we expect from istio to change? As a final result, do we want the istio deployments to have an custom `app` label there?

I am having two ideas for a workaround in kiali:
1. Hardcode the app_label to "app".
2. Add a new config in kiali: "spec.external_components.istio.app_label" where the "app" label is defined for the istio components.

Wdyt?

jshaughn:
> What do we expect from istio to change?

I guess I thought that Istio should be more flexible and let the user configure the label, or possibly apply both the 'app' and the standard k8s label.  I don't think they want to force the use of the `app` label for istiod.   But that won't solve the issue in the near-term.

I guess the hardest thing is knowing when we need to ignore our configured app label name and use either the hard coding or the proposed new config.  I guess I'd suggest we take the easier route, hard code it for this use case, and possible backport to 1.28 and make an upstream release to help these users immediately.  At the same time, create an Istio issue (or PR) to resolve the problem on their side.

xeviknal:
Closing the issue per https://github.com/kiali/kiali/pull/3544

PR #3544, opened 2020-12-21
Force kiali to use "app" label name for fetching istio components

xeviknal:
This is a workaround for the https://github.com/kiali/kiali/issues/3543

Right now, Istio forces the usage of the app label as "app" while there are user which use different app label in their other deployments. It sounds a bit inconsistent.

Need this PR to quickly fix the bug in our users but we expect Istio to fix this situation. When this is fixed in upstream Istio, we'll be back to previous usage.

To reproduce change the istio_labels settings in your CR similarly to the following snippet:
```yaml
istio_labels:
  app_label_name: different_app_label
  version_label_name: different_version_label
```

Previous behavior:
- Error message shown right after login:
Screenshot of Kiali Console (4)

- All health in warning (unsync'ed workloads):
Screenshot of Kiali Console (5)
Screenshot of Kiali Console (6)

Right after, those errors shouldn't show up.

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -164,7 +164,7 @@ func (iss *IstioStatusService) getStatusOf(ds []apps_v1.Deployment) (IstioCompon
 
 	// Map workloads there by app name
 	for _, d := range ds {
-		appLabel := labels.Set(d.Spec.Template.Labels).Get(config.Get().IstioLabels.AppLabelName)
+		appLabel := labels.Set(d.Spec.Template.Labels).Get("app")
 		if appLabel == "" {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


jshaughn (approved):
@xeviknal LGTM.  Did you create an Istio issue?

xeviknal:
@jshaughn just created this one: https://github.com/istio/istio/issues/29742 and also another kiali issue tracking the istio changes https://github.com/kiali/kiali/issues/3547

skondkar (approved):
Verified on Kiali Server :    v1.29.0-SNAPSHOT (8cbe752e15741fb8567cbc5989bf910a4e73b548)

-Reproduced the issue before applying the PR
-Applied the PR and verified that there are no error messages after login and there are no health warnings on workload list and details pages.

Screenshot_Workload-list

Screenshot-Workload-Details

Issue #3539, opened 2020-12-18
Proxy status not reporting when cache is disabled


bougyman:
Due to an issue with our kiali (https://github.com/kiali/kiali/issues/3537), we have to deploy with cache disabled. Once doing so, we noticed that all of our applications are showing unsynced proxy status / degraded in the UI. In the logs, I see
```
2020-12-18T18:00:10Z WRN Kiali has not [istio-system] namespace cached. It is required to fetch Istio Status correctly
```
Which leads me to believe that disabling the cache is reponsible for the misreporting status of the proxies.

I also found #3318 and #3508 which have the same proxy status issue, but we do not seem to meet either of those conditions.

**Versions used**
Kiali: 1.28.0
Istio: 1.6.9
Kubernetes flavour and version: GKE 1.17

**To Reproduce**
1. Disable cache in the kiali cr with
```
kubernetes_config:
  cache_enabled: false
```
2. Go to the UI after this cr has deployed
3. Click on 'Applications'
4. In the 'Health' column, apps should have a yellow yield sign, hover over that to see that the proxy status shows unsynced

**Expected behavior**
As istioctl ps shows all proxies synced, these degraded icons and messages should not show up, it should be green and show the proper proxy status.

jmazzitelli:
cc @lucasponce

PR #3549, opened 2020-12-22
Disable the proxy status into health when cache is disabled (all pods doesn't have sidecar)

xeviknal:
fixes https://github.com/kiali/kiali/issues/3539

The idea behind is to disable the unsync'ed pods when the cache is disabled.
This will avoid the situation mentioned in the issue linked right above and won't show warnings in all health.
The proxy-status is disabled when cache is disabled because the query performed can be massive and will dramatically damage the whole performance in medium deployments.

To reproduce:
Add the following config in your kiali CR to disable the cache:

```yaml
kubernetes_config:
  cache_enabled: false
```

Bug:
You should be able to see all the health in the mesh with unsynced proxies.

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -100,17 +100,6 @@ func (in *HealthService) GetWorkloadHealth(namespace, workload, workloadType, ra
 		}, nil
 	}
 
-	// Add Proxy Status info
-	if w.Pods != nil || len(w.Pods) > 0 {
-		syncedProxies := int32(0)
-		for _, p := range w.Pods {
-			if p.ProxyStatus != nil && p.ProxyStatus.IsSynced() {
-				syncedProxies++
-			}
-		}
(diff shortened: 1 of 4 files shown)


skondkar (approved):
Verified on Kiali Server    v1.29.0-SNAPSHOT (fa990da74975c53f8a5f328a666792c5673c0e95)
There are no health warnings on Applications/Workloads/services list/details pages after applying the PR
LGTM.

Note: There are below warnings in kiali pod logs:
2020-12-23T07:32:37Z WRN Kiali has not [istio-system] namespace cached. It is required to fetch Istio Status correctly

prachiyadav (approved):

lucasponce:
@xeviknal I missed to review this one. I'll do it today.

lucasponce (approved):

Issue #3552, opened 2020-12-29
Kiali show service entries in another namespaces


alexandershelega:
Hi, we have create following service entry in dev namespace and use `exportTo "."` but when we open graph view for production namespace we also see that service entry.

```
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: main-mongodb-se
  namespace: dev
spec:
  hosts:
    - main-db1-1.c.internal-ai.internal
    - main-db1-2.c.internal-ai.internal
    - main-db1-3.c.internal-ai.internal
  exportTo:
  - "."
  ports:
    - name: main-mongo-clients
      number: 27017
      protocol: TCP
  location: MESH_EXTERNAL
  resolution: NONE
  ```

Please advice.

**Versions used**
Kiali: 1.28.1
Istio: 1.8.1
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9)
GKE : 1.17.13-gke.2600

**Comment by jotak** [comment]
Nothing sure, but could it be a hint that your dev service entry... is actually used in production? Kiali shows essentially whatever gets traffic in and out (based on istio telemetry), and then tries to map that to kubernetes objects in order to better identify them (in short).
So, I imagine if the service entry shows up, it must be because some of your (production) workloads are sending data to, at least, one of the hosts defined in the service entry. So that Istio telemetry shows traffic sent to `main-db1-x.c.internal-ai.internal`, that kiali interprets as being that represented by this service entry.

**Comment by jotak** [comment]
Ok, forget my previous comment.
It looks like a case that needs to be handled specifically, as stated in this doc:

```
The following example demonstrates the use of a dedicated egress gateway through which all external service traffic is 
forwarded. The ‘exportTo’ field allows for control over the visibility of a service declaration to other namespaces in the mesh. By 
default, a service is exported to all namespaces. The following example restricts the visibility to the current namespace, 
represented by “.”, so that it cannot be used by other namespaces.
```
( https://istio.io/latest/docs/reference/config/networking/service-entry/ )

cc @jshaughn @israel-hdez - I don't think we cover this kind of configuration indeed

jshaughn:
@alexandershelega , are you seeing traffic in the graph from `prod` services to the `dev` service entry?  Or is the service entry somehow showing up as an orphan node (with no incoming or outgoing edges)?    Do you have an analogous service entry defined in `prod` that you are expecting to see, but instead it is being labeled as `dev` namespace?

[...]

PR #3559, opened 2021-01-05
Handle `ExportTo` when Service Entry matching

jshaughn:
When matching Service Entries with hosts make sure the SE is exported to the current namespace.

Fixes https://github.com/kiali/kiali/issues/3552

### DIFF
--- a/graph/telemetry/istio/appender/service_entry.go
+++ b/graph/telemetry/istio/appender/service_entry.go
@@ -105,7 +105,7 @@ func (a ServiceEntryAppender) applyServiceEntries(trafficMap graph.TrafficMap, g
 
 		// A service node represents a serviceEntry when the service name matches serviceEntry host. Map
 		// these "se-service" nodes to the serviceEntries that represent them.
-		if se, ok := a.getServiceEntry(n.Service, globalInfo); ok {
+		if se, ok := a.getServiceEntry(n.Service, globalInfo, namespaceInfo); ok {
 			if nodes, ok := seMap[se]; ok {
 				seMap[se] = append(nodes, n)
 			} else {
@@ -165,10 +165,10 @@ func (a ServiceEntryAppender) applyServiceEntries(trafficMap graph.TrafficMap, g
 
 // getServiceEntry queries the cluster API to resolve service entries across all accessible namespaces
 // in the cluster.
-// TODO: I don't know what happens (nothing good) if a ServiceEntry is defined in an inaccessible namespace but exported to
-// all namespaces (exportTo: *). It's possible that would allow traffic to flow from an accessible workload
-// through a serviceEntry whose definition we can't fetch.
-func (a ServiceEntryAppender) getServiceEntry(serviceName string, globalInfo *graph.AppenderGlobalInfo) (*serviceEntry, bool) {
+// TODO: I don't know what happens (nothing good) if a ServiceEntry is defined in an inaccessible namespace
+// but exported to all namespaces (exportTo: *). It's possible that would allow traffic to flow from an
+// accessible workload through a serviceEntry whose definition we can't fetch.
+func (a ServiceEntryAppender) getServiceEntry(serviceName string, globalInfo *graph.AppenderGlobalInfo, namespaceInfo *graph.AppenderNamespaceInfo) (*serviceEntry, bool) {
 	serviceEntryHosts, found := getServiceEntryHosts(globalInfo)
 	if !found {
 		for ns := range a.AccessibleNamespaces {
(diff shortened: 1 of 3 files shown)


graph/telemetry/istio/appender/service_entry.go
```diff
@@ -225,3 +231,22 @@ func (a ServiceEntryAppender) getServiceEntry(serviceName string, globalInfo *gr
...
+		return true
+	}
+	for _, export := range se.exportTo.([]interface{}) {
+		if export == "*" {
+			return true
+		}
+		if export == "." && se.namespace == namespaceInfo.Namespace {
+			return true
+		}
+		if export == se.namespace {
```
jotak, line 246: I think this should be commented out, or removed, because according to the documentation, it is not implemented yet in istio: https://istio.io/latest/docs/reference/config/networking/service-entry/
Says:

```
NOTE: in the current release, the exportTo value is restricted to “.” or “*” (i.e., the current namespace or all namespaces).
```

That note exists since v1.0 at least (i didn't dig further), so I guess we shouldn't expect the feature to be implemented any time soon ?

graph/telemetry/istio/appender/service_entry_test.go
```diff
@@ -124,7 +127,114 @@ func serviceEntriesTrafficMap() map[string]*graph.Node {
 func TestServiceEntry(t *testing.T) {
 	assert := assert.New(t)
 
-	businessLayer := setupServiceEntries()
+	businessLayer := setupServiceEntries(nil)
```
jotak, line 130: nice to have so many tests! :-) :+1:

graph/telemetry/istio/appender/service_entry.go
jshaughn: @jotak I didn't actually see that comment, I had looked at the initial description and the fact it allowed a list of values.  Despite the current restriction I'm inclined to leave this code here as it shouldn't harm anything and will cover the use case if and when it ever shows up in Istio.  Are you OK with that?

jotak: yes, sounds good!

jshaughn: Great, thanks.  If you have no other comments feel free to approve :)

jotak (commented):
Tested several scenarios that are all fine
- with SE defined in another namespace and exportTo '*'
- with SE defined in another namespace and exportTo '.'
- with SE defined in same namespace and exportTo '*'
- with SE defined in same namespace and exportTo '.'
- without exportTo

I've seen a bug however, but it's not directly related to this PR (same thing happens in master) but now it is more problematic : when two service entries are defined and match the hosts, it seems like one of them is randomly chosen for display (so, refreshing may make them showing up alternatively). But now if one of those service entries has the "exportTo" set to '.' and is in another namespace (hence being ignored), it seems like it prevents the valid one from being found, which result in inconsistent display:

E.g. first refresh I get this:
Capture d’écran de 2021-01-06 16-42-15

Second refresh I get this:
Capture d’écran de 2021-01-06 16-42-18

graph/telemetry/istio/appender/service_entry.go
```diff
@@ -198,6 +200,10 @@ func (a ServiceEntryAppender) getServiceEntry(serviceName string, globalInfo *gr
 	}
 
 	for host, se := range serviceEntryHosts {
+		if !isExportedToNamespace(se, namespaceInfo) {
```
jotak, line 203: related to my previous comment, I guess the `isExportedToNamespace` check should be moved above, before adding the SE to the hosts map

jshaughn: @jotak Excellent discovery for the reported problem.  The issue with your suggested fix is that the code above is actually establishing a cached map (stored in globalnfo) that can be re-used for each namespace in the current graph.  So we can't exclude SEs in that code, it needs to have them all.  Let me take a look and see what can be done...

jshaughn: @jotak pushed some changes that I think ensures complete maching when a host is duplicates and also guarantees a predictable/repeatable match.  Please give it a try when you get the chance.

jotak: One of my scenarios doesn't seem to work:

-    with SE defined in same namespace and exportTo '.'

I expect to see the service entry in graph but it doesn't show up. Let me see if I find something in code...

jotak: pff I think I've found the problem but it's going to be complex to solve. Perhaps we should merge this and create a follow-up because it actually fixes the initial issue, while uncovering a new one. Basically, the namespace that we need to consider for that service entry node is the namespace of the workload that originated the request, if I'm correct. We can chat about that when you're available.

jotak: it also means that, within the same graph, a single service host might be represented both as a service entry and not a service entry, if the requests it serves comes from different workloads of different namespaces and `exportTo` is `.`.

jshaughn: OK, so I was writing some slightly complicated code that started dealing with the source nodes and such when I realized that in fact the source workload namespaces really should, I think, always be the same as the namespace being processed by the appender.  So, with that simplification I submitted a commit for your consideration.  If it's not a valid assumption then I'll have to go back and reconsider.

jotak: cool, fingers crossed, i'll test it

jotak (approved):
LGTM, all my scenarios look good, CR looks good, thanks @jshaughn !

mattmahoneyrh (approved):

Issue #3541, opened 2020-12-18
molecule test "openid-test" is failing


jmazzitelli:
The "openid-test" molecule test is consistently failing. Seemed to have started a few days ago.

To see this:

1) Set up minikube with dex: `hack/k8s-minikube.sh start --dex-enabled true -kv 1.18.0`
2) Install Istio: `hack/k8s-minikube.sh istio`
3) Run the molecule test: `make -e CLUSTER_TYPE=minikube -e MOLECULE_SCENARIO="openid-test" molecule-test`

```
TASK [With the login cookie and nonce code, make a request to the kiali auth endpoint] ***
fatal: [localhost]: FAILED! => {"changed": false, "connection": "close", "content": "{\"error\":\"Cannot view any namespaces. Please read Kiali's RBAC documentation for more details.\"}", "content_length": "96", "content_type": "application/json", "date": "Fri, 18 Dec 2020 21:29:33 GMT", "elapsed": 0, "json": {"error": "Cannot view any namespaces. Please read Kiali's RBAC documentation for more details."}, "msg": "Status code was 401 and not [200]: HTTP Error 401: Unauthorized", "redirected": false, "server": "openresty/1.15.8.2", "set_cookie": "kiali-token-openid-nonce=; Path=/kiali; Expires=Thu, 01 Jan 1970 00:00:00 GMT; HttpOnly; SameSite=Strict", "status": 401, "url": "https://192.168.99.100/kiali/api/authenticate", "vary": "Accept-Encoding"}
```

jshaughn:
@cfcosta The molecule tests are the best end-to-end tests we have.  They don't trigger anything in the UI but they can utilize the API.  Anyway, for this issue you'll need to at least run locally and see if you can solve the failure.

jmazzitelli:
I don't know why this started happening now (before this test was working - something must have changed in the server around the namespace lookup). But clearly the admin@example.com user doesn't have access to [read the list of namespaces.](https://github.com/kiali/kiali/blob/v1.28.0/business/openid_auth.go#L690). I'll see if I can figure out how to fix this and get the test to pass again.

jmazzitelli:
Problem was in the Dex setup in our k8s-minikube.sh hack script - needed to give the admin@example.com user a role binding that can see namespaces. I'll submit a PR soon to fix that.

PR #3564, opened 2021-01-07
[hack] add role/binding to the Dex admin@example.com user so the openid molecule test passes

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/3541


jshaughn (approved):
LOK

PR #3573, opened 2021-01-11
make a copy of the prometheus default round tripper http transport to avoid race condition

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3518


israel-hdez (approved):
Invoked `make test-race` locally and got the issue. It's replicated quite easily. For me, it's happening several times in a row.

With this PR issue is gone. I think this is right.

Issue #3537, opened 2020-12-17
Review client-go version used in Kiali


lucasponce:
Some users running on GKE are reporting an odd problem with caching and debugging it points to the client-go version used.

Kiali is using an old version (but compatible) of client-go [1] but I start to suspect that it may be related to these oddities reported on #3494 

[1] https://github.com/kubernetes/client-go

Issue #3494, opened 2020-12-01
Some HTTP request still pending and never load (Graph, Applications, Workloads)


JuanSeBestia:
**Describe the bug**
the application keeps processing some calls indefinitely, this can happen with various APIs in various sections of the application (such as Graph, Applications, Workloads, ... etc).

I waited half an hour and the HTTP call was still pending 

reference image
<img width="1666" alt="Screen Shot 2020-12-01 at 4 46 00 PM" src="https://user-images.githubusercontent.com/7362688/100809050-4d381780-3403-11eb-945d-f315ad063b03.png">

The strangest thing is that it only happens in a productive environment, in my company has 3 types of environments (prod, staging, ephemeral env), but this can **only** be reproduced in the prod environment.

I have seen the logs in debug mode and they are not interesting at all, I will still attach them.

I don't have a clue what happens, any suggestion is welcome :)

I roll back to Kiali Operator v1.24.0 and it works as expected, but I will continue with this version for now.

**Versions used**
Kiali Operator v1.26.1
Kiali UI 1.26.0 (35e2d8b075bd5c4bcc3f315aef9e6144aad0b608)
Kiali Server v1.26.1 (0cf243d730d45cd16e254e2c8d4d2a0ac0cf5222)
Kiali Container v1.26.1
Components
Istio 1.7.4
Promtheus 2.18.2
Kubernetes v1.17.13-gke.1400

**NOTE: ** also I test it with Kiali Operator v1.25.0, Kiali Operator v1.27.0

**To Reproduce**
1. Go to 'Workloads'

**Expected behavior**
Load all the workloads

JuanSeBestia:
Logs:
[Kiali_1_26_1.txt](https://github.com/kiali/kiali/files/5626111/Kiali_1_26_1.txt)

jotak:
@lucasponce have you ever seen this "too old resource version" log? Looks related to the watchers.

```
W1201 21:45:46.726043       1 reflector.go:289] pkg/mod/k8s.io/client-go@v11.0.1-0.20190820062731-7e43eff7c80a+incompatible/tools/cache/reflector.go:94: watch of *v1.ConfigMap ended with: too old resource version: 37868011 (37870980)
```

Perhaps when such error occurs we need to reinitialize watchers?

@JuanSeBestia 
Does that log look related with the issue, ie. do you see it constantly when the problem appears and you don't see it in your working staging environment?

lucasponce:
[...]

PR #3579, opened 2021-01-13
Align client-go, apimachinery and prometheus dependencies

lucasponce:
Kiali dependencies used for Prometheus and Kubernetes were pretty old and I have suspects those are related in some issues I'm investigating.

I performed a basic bump of those libraries following the same table used by Istio project [1].

API had some changes, i.e. passing context in k8s calls, and prometheus API added a new Warnings result as well, so I've just changed the minimal in the code base to make it compatible (but without touching any business logic).

It looks ok on my local env but I'd like to see if there is any other potential problem.

In any case, I want to separate this bump from other efforts.

cc @hhovsepy this one would be a good candidate to check if introduces regressions, I haven't tested all features, just a couple of basic smoke tests.

This is also related/part of the investigation/work spotted on https://github.com/kiali/kiali/issues/3494

Fixes https://github.com/kiali/kiali/issues/3537

[1] https://github.com/istio/istio/blob/master/go.mod


jotak:
From the `go.sum` file, it seems like there's a lot of new additions and libs that come in several versions. I wonder if a `go mod tidy` could clean some of that, did you try to run it?
(I remember we had some troubles with `go mod tidy` in the past, I think it was breaking the build, not sure if it works well now)

graph/telemetry/istio/appender/util.go
```diff
@@ -11,20 +11,21 @@ import (
...
+	"strings"
 )
 
 // package-private util functions (used by multiple files)
 
-func promQuery(query string, queryTime time.Time, api prom_v1.API, a graph.Appender) model.Vector {
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
-
+func promQuery(query string, queryTime time.Time, ctx context.Context, api prom_v1.API, a graph.Appender) model.Vector {
```
lucasponce, line 19: Note, that it's a more or less convention to put the context parameter in the first position of the signature.
I did for the k8s ones but on this util method I didn't want to churn too much the existing code, so I put it together with the API one.

graph/telemetry/istio/appender/util.go
```diff
@@ -11,20 +11,21 @@ import (
 	"github.com/kiali/kiali/graph"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
+	"strings"
```
lucasponce, line 14: Self review, this needs to go up.

graph/telemetry/istio/istio.go
```diff
@@ -29,6 +29,7 @@ import (
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/prometheus"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
+	"strings"
```
lucasponce, line 32: Here

prometheus/client.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/util/httputil"
+	"strings"
```
lucasponce, line 18: Here

prometheus/client.go
```diff
@@ -39,6 +40,7 @@ type Client struct {
 	ClientInterface
 	p8s api.Client
 	api prom_v1.API
+	ctx context.Context
```
lucasponce, line 43: Context is mandatory but we are not using, instead to create a new empty one with Background() call, we can store it at client level for future uses.

graph/telemetry/istio/appender/util.go
```diff
@@ -11,20 +11,21 @@ import (
...
 )
 
 // package-private util functions (used by multiple files)
 
-func promQuery(query string, queryTime time.Time, api prom_v1.API, a graph.Appender) model.Vector {
-	ctx, cancel := context.WithCancel(context.Background())
```
lucasponce, line 19: This cancel has not make too much sense as context.Background() is empty and it's not cancellable or used to sync with other "context" area.

jshaughn: Using WithCancel() to wrap context.Background() just provides the ability to cancel.  It's there *because* context.Background() is empty.   I still think we want to be able to cancel prom queries and when I wrote this I probably expected we would look into that sort of thing.  I agree that at the moment it's likely not needed, as we have no support for actually canceling the query, and as such the deferred cancel() is always invoked.  So for now I guess we can remove it, although the cost is nominal.

graph/telemetry/istio/appender/util.go
jshaughn: It's just a utility, if you feel the need to change it further go ahead.

prometheus/cache.go
```diff
@@ -265,6 +265,6 @@ func (c *promCacheImpl) watchExpiration() {
 		c.wkRequestRatesLock.Lock()
 		c.cacheWkRequestRates = make(map[string]map[string]map[string]timeInOutResult)
 		c.wkRequestRatesLock.Unlock()
-		log.Infof("[Prom Cache] Expired")
+		log.Tracef("[Prom Cache] Expired")
```
jshaughn, line 268: \o/

jshaughn (approved):
A couple of comments, no changes required.  No issues with build or runtime in a short test of graph and other ui featured.

jotak:
@lucasponce did `go mod tidy` work? I think last week I looked again and found the `go.sum` looked much better, but now it seems much bigger again.

Apart from that, the code looks good to me

lucasponce:
> @lucasponce did go mod tidy work? I think last week I looked again and found the go.sum looked much better, but now it seems much bigger again.

I performed a "go mod tidy" on commit:

https://github.com/kiali/kiali/pull/3579/commits/5c5ddba48a0a64304ea13748c66db190d6c4356a

Let me repeat locally to check if I missed something.

jotak:
Well, that must be good, the next commits didn't add anything. I wonder why there are so many references of libraries with different versions.

jotak (approved):
LGTM

lucasponce:
@jotak just checked that is seems correct:

```
~/Projects/gopath/src/github.com/kiali/kiali(client-go-update) $ go mod tidy
warning: ignoring symlink /home/lponce/Projects/gopath/src/github.com/kiali/kiali/operator
~/Projects/gopath/src/github.com/kiali/kiali(client-go-update) $ git status
En la rama client-go-update
nada para hacer commit, el árbol de trabajo está limpio
~/Projects/gopath/src/github.com/kiali/kiali(client-go-update) $ 
```

So, no changes from the branch.

hhovsepy (approved):
Regression tested.

Issue #3523, opened 2020-12-11
[multi-cluster] Identify current Kiali instance in a "primary" cluster


israel-hdez:
**Is your feature request related to a problem? Please describe.**
In multi-cluster enabled environments, it's possible to have several Kiali instances deployed on each control plane of the mesh. This may lead to some confusion about which Kiali instance is being accessed.

**Describe the solution you'd like**
Kiali should, somehow, clearly state what is the current instance. Probably, the better way is to show info about the environment and not Kiali itself (i.e. what's the cluster name and mesh id hosting the current Kiali instance).

**Describe alternatives you've considered**
Currently, Kiali doesn't identifies itself in any way. But if we are willing to introduce some configurable identifier in the CR, probably a Kiali instance ID? probably a label? If we don't want more config options, perhaps the pod name?

Although, a Kiali identifier is probably not enough, as users may want to know what's the environment (i.e. what's the cluster).

**Additional context**
Issue title states a "primary" cluster. This means Kiali is installed along a control plane.

Epic: https://github.com/kiali/kiali/issues/1833

israel-hdez:
Marking as done. There will be some UX discussion about the UI indicator.

PR #3580, opened 2021-01-13
Add Cluster info to the server config

israel-hdez:
The server configuration API endpoint will now return the cluster_id and the network_id of the cluster.

This code is to support UI changes: kiali/kiali-ui#2052
Related kiali/kiali#3523

### DIFF
--- a/handlers/config.go
+++ b/handlers/config.go
@@ -8,7 +8,9 @@ import (
 	"github.com/prometheus/common/model"
 	"gopkg.in/yaml.v2"
 
+	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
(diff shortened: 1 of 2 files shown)


handlers/config.go
```diff
@@ -85,6 +93,23 @@ func Config(w http.ResponseWriter, r *http.Request) {
 		},
 	}
 
+	// The following code fetches the cluster info. Cluster info is not critical.
+	// It's even possible that it cannot be resolved (because of Istio not being with MC turned on).
+	// Because of these two reasons, let's simply ignore errors in the following code.
```
jshaughn, line 98: It's OK to ignore but I also don't see a problem with just an info level log indicating that cluster was not configured.

jshaughn (approved):
One non-blocking comment

handlers/config.go
israel-hdez: I added several log messages. This should give clarity about what's the issue if no indicator is shown.

Issue #3532, opened 2020-12-16
Sorry, there was a problem. Try a refresh or navigate to a different page. - with Argo Rollouts


arturrez:
**Describe the bug**
I am getting "Sorry, there was a problem. Try a refresh or navigate to a different page." when click on ArgoRollouts object in Kiali UI

**Versions used**
Kiali UI 1.27.0 (8f8d4c0730857d7a25bd2ae613f0bff419e50e28)
Kiali Server v1.27.0 (f78420fc6e0cf565944d714ee92523e0642ca3e8)
Kiali Container v1.27.0
Components
Istio 1.7.4
Prometheus 2.22.1
Kubernetes v1.17.13-gke.2001

**To Reproduce**
1. Go to Workloads
2. Click on some Rollou Object
3. See error above

**Expected behavior**
I expect to see details for Rollout object similar what I see for deployment object

**Screenshots
![2020-12-15_18-14](https://user-images.githubusercontent.com/56270896/102296517-57c5d700-3f02-11eb-8712-e10c16602b8a.png)
![2020-12-15_18-19](https://user-images.githubusercontent.com/56270896/102296520-598f9a80-3f02-11eb-8758-64b7f471523d.png)

**
![2020-12-15_18-13](https://user-images.githubusercontent.com/56270896/102296482-48df2480-3f02-11eb-83d5-7c12a27480fc.png)

lucasponce:
Thanks @arturrez for the issue.
Yes, probably it's a bug when the controller is in rollout status.

lucasponce:
Ok, "Argo Rollout" [1] is a custom controller provided by Argo api "argoproj.io/v1alpha1" so, Kiali can't get details of this controller at the moment. On this case, the controller info is infered from the Pod and probably can't get all details that a Deployment should provide (i.e. status of desired controllers).

I will review if in this case what is the best strategy, but yes, we will fix this error, as minimum it shouldn't fail when accessing to the details.

[1] https://argoproj.github.io/argo-rollouts/#how-does-it-work

ahsanqaisar:
Hi there - I recently deployed Kiali v1.31. The workload is appearing fine but it is giving the same issue under the Application tab. 

![image](https://user-images.githubusercontent.com/73361295/111342324-bb21eb00-8650-11eb-906d-d37a64822e44.png)

Thought anyone?

jmazzitelli:
> Thought anyone?

[...]

PR #3585, opened 2021-01-15
Better support for custom controllers

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3532

Some frameworks create custom workload controllers replacing standard "Deployment" or kubernetes controller with a custom one.

This fix improves the support for those cornercases.

It assumes that, as it happens with a Deployment, a custom controller define a child controller as a ReplicaSet and try to use it for basic workload information: pod and service status.

Most information is ok but custom properties of the controller are not supported as we would need to extend the native controllers that Kiali can handle beyond the standard ones.

For QE, how to test:
- Verify no regressions on existing workloads logic.
- Install a third party framework with custom controllers, i.e. https://argoproj.github.io/argo-rollouts/#quick-start
- Deploy an example with custom controllers i.e. https://github.com/argoproj/argo-rollouts/blob/master/examples/rollout-analysis-step.yaml

Verify that Kiali has basic support for custom controllers as workloads without errors:

Workload List
image

Workload Detail
image

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -506,7 +506,7 @@ func fetchWorkloads(layer *Layer, namespace string, labelSelector string) (model
 	// Resolve ReplicationControllers from DeploymentConfigs
 	// Resolve Jobs from CronJobs
 	for cname, ctype := range controllers {
-		if ctype == "ReplicaSet" {
+		if ctype == kubernetes.ReplicaSetType {
 			found := false
 			iFound := -1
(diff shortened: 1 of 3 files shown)


hhovsepy (approved):
Verified:
Screenshot from 2021-01-21 11-57-58
Screenshot from 2021-01-21 12-00-01

pantianying:
The same problems still exist

image
image

PR #3590, opened 2021-01-18
Fix some things to work with the new header auth strategy

jmazzitelli:
Update the run-molecule-tests script so it prepares header-auth correctly.
Also change the k8s-minikube.sh hack script to only create that special role when requested.
Cleans up some import ordering, too.


jshaughn (approved):
LOK

Issue #3592, opened 2021-01-18
[multi-cluster] Initial support


jshaughn:
This issue is for initial enhancements to better support Istio's multi-cluster deployment strategies.  It addresses some of the tasks in:

Epic: https://github.com/kiali/kiali/issues/1833

It comprises various epic sub-tasks developed in the `multi-cluster` feature branch.   This work is compatible with single-cluster environments.   The main enhancments are:

- An experimental Mesh discovery List view
- Graph enhancements for Namespace and Cluster boxing

jshaughn:
PRs are merged

Issue #3591, opened 2021-01-18
Error authenticating (getting business layer) - authInfo missing from the request context


jmazzitelli:
Either I didn't build things correctly, or the openshift auth strategy is now broken due to the changes made in the header-auth strategy PRs that just went in.
I'm pretty sure this is broken and not due to a wrong build or old cookies in my browser - this is bad - openshift deployments are now broken.

jmazzitelli:
@mlbiam I'm seeing this when I try to log in while deployed on OpenShift using openshift auth strategy - perhaps you know the place to fix this:

![image](https://user-images.githubusercontent.com/2029470/104959169-94706c80-599f-11eb-88f4-fdc87170fc98.png)

mlbiam:
The error it's self is because the `authInfo` variable wasn't present when getting the business layer, so it didn't know how to talk to openshift (this replaced the `token` variable in the request).  Is this on authentication or afterwards?  what happens when you hit /kiali/api/authenticate?

jmazzitelli:
It is on authenication - trying to login (enter user/pass in the openshift openid login screen) and I immediately get the above.

jmazzitelli:
The problem is here:

https://github.com/kiali/kiali/blob/master/handlers/authentication.go#L654

mlbiam:
My guess is line 147 in business/authentication.go - `business, err := getBusiness(r)`.  `r` doesn't have the `authInfo` in there.  I'm guessing you're trying to get the business object using the kiali service account, similar to we validate the header token using the kiali service account?  Try: `business, err := business.Get(&api.AuthInfo{Token: kialiToken})` ?

jmazzitelli:
> My guess is line 147 in business/authentication.go - `business, err := getBusiness(r)`. `r` doesn't have the `authInfo` in there. I'm guessing you're trying to get the business object using the kiali service account, similar to we validate the header token using the kiali service account? Try: `business, err := business.Get(&api.AuthInfo{Token: kialiToken})` ?

I did try that already. Didn't work. Same behavior (had to use `token` in this context. there is no `kialiToken`)

mlbiam:
is `/auth/info` called after authenticating?  so the user should already be authenticated, right?

jmazzitelli:
Now that I pay closer attention, I'm not even getting to the login screen. Something is happening even before I get challenged with the openshift login page.

jmazzitelli:
[...]

PR #3593, opened 2021-01-18
[multi-cluster] initial support, server-side merge of multi-cluster branch

jshaughn:
Part of https://github.com/kiali/kiali/issues/3592

Requires ui PR: https://github.com/kiali/kiali-ui/pull/2055

Requires master operator commit: 47a3225e56d021e09a51f1cc6db890ba6f6b23c8

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -1,0 +1,168 @@
+package business
+
+import (
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+)
+
+// MeshService is a support service for retrieving data about the mesh environment
+// when Istio is installed with multi-cluster enabled.
+type MeshService struct {
+	k8s           kubernetes.ClientInterface
+	businessLayer *Layer
+}
+
+// Cluster holds some metadata about a cluster that is
+// part of the mesh.
+type Cluster struct {
+	// ApiEndpoint is the URL where the Kubernetes/Cluster API Server can be contacted
+	ApiEndpoint string `json:"apiEndpoint"`
+
+	// IsKialiHome specifies if this cluster is hosting this Kiali instance (and the observed Mesh Control Plane)
+	IsKialiHome bool `json:"isKialiHome"`
+
+	// Name specifies the CLUSTER_ID as known by the Control Plane
+	Name string `json:"name"`
+
+	// SecretName is the name of the kubernetes "remote secret" where data of this cluster was resolved
+	SecretName string `json:"secretName"`
+}
+
+// GetClusters resolves the Kubernetes clusters that are hosting the mesh. Resolution
+// is done as best-effort using the resources that are present in the cluster.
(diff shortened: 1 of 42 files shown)


business/mesh.go
```diff
@@ -0,0 +1,315 @@
...
+// newRemoteClientFunc arguments (see the MeshService struct for details). The newRemoteClientFunc
+// can be passed a nil value and a default function will be used.
+func NewMeshService(k8s kubernetes.ClientInterface, newRemoteClientFunc func(config *rest.Config) (kubernetes.ClientInterface, error)) MeshService {
+	if newRemoteClientFunc == nil {
+		// Mesh discovery will use ServiceAccount token instead of user token
+		token := ""
+		token, _ = kubernetes.GetKialiToken()
+
+		newRemoteClientFunc = func(config *rest.Config) (kubernetes.ClientInterface, error) {
+			return kubernetes.NewClientFromConfig(config, &api.AuthInfo{Token: token})
```
israel-hdez, line 57: This is misleading. We have to honor the token that is passed in `config` argument. But I tried the code and it's working anyway.

I reviewed what's inside that `NewClientFromConfig` and, indeed, it does nothing with the token that is passed to it via `authInfo` other than simply storing for later. The real client is created purely from `config`. 

I have the idea that current state of `NewClientFromConfig` is error prone, because it ignores the token, but it can merge Impersonation data regardless of any token mismatch between `config` and `authInfo`.

:/

israel-hdez: For more context: this piece of code is creating a kubernetes client for remote clusters; not the "home/local" cluster. So, any local token or auth context that there is in place is not valid for querying the remote clusters API Servers.

mattmahoneyrh:
Encountered an OAUTH error:

{"error":"Error authenticating (getting business layer)","detail":"authInfo missing from the request context"}

jmazzitelli:
> {"error":"Error authenticating (getting business layer)","detail":"authInfo missing from the request context"}

This was fixed last night: 
* https://github.com/kiali/kiali/issues/3591
* https://github.com/kiali/kiali/pull/3594

business/mesh.go
jshaughn: @israel-hdez I agree with your comments.  This initial support was only an attempt to fix the rebase issues for compilation.  I'll resolve this conversation and you can push any update you feel is more suitable.

israel-hdez: @jshaughn I chose to create a PR to let @jmazzitelli and @mlbiam take a look.
Here it is: https://github.com/kiali/kiali/pull/3602

mattmahoneyrh:
@mazz a rebuild picked up the fix.

@israel-hdez Should the cluster name be present if not configured in Multi-Cluster mode (appears to be a static " Kubernetes" value)?

image

mattmahoneyrh (approved):

PR #3594, opened 2021-01-18
fix openshift auth strategy

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3591

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -617,7 +617,7 @@ func (aHandler AuthenticationHandler) Handle(next http.Handler) http.Handler {
 
 func (aHandler AuthenticationHandler) HandleUnauthenticated(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		context := context.WithValue(r.Context(), "token", "")
+		context := context.WithValue(r.Context(), "authInfo", &api.AuthInfo{Token: ""})
 		next.ServeHTTP(w, r.WithContext(context))
 	})
 }


jmazzitelli:
This is not working.

handlers/authentication.go
```diff
@@ -617,7 +617,7 @@ func (aHandler AuthenticationHandler) Handle(next http.Handler) http.Handler {
 
 func (aHandler AuthenticationHandler) HandleUnauthenticated(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		context := context.WithValue(r.Context(), "token", "")
+		context := context.WithValue(r.Context(), "authInfo", &api.AuthInfo{Token: ""})
```
jmazzitelli, line 620: I believe this was the crux of the problem. This may be the only line that needs to be fixed. I'm still testing.

handlers/authentication.go
```diff
@@ -144,7 +144,7 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
 
 	expiresOn := time.Now().Add(time.Second * time.Duration(expiresInNumber))
 
-	business, err := getBusiness(r)
+	business, err := business.Get(&api.AuthInfo{Token: token})
```
jmazzitelli, line 147: I'm not 100% sure this line change is needed. I'm testing to see if this change is required.

jmazzitelli: this is not needed. I will revert this change.

handlers/authentication.go
jmazzitelli: This fixes the openshift auth strategy

jshaughn (approved):
Looks promising.

Issue #3514, opened 2020-12-09
Jaeger client: switch to gRPC/protobuf


jotak:
See https://www.jaegertracing.io/docs/1.19/apis/#grpcprotobuf-stable

I hope to see better perfs when switching to grpc (to be measured). If this is the case, it could probably enable a couple of enhancements (e.g. spreading traces over more interval via split/join, see https://github.com/kiali/kiali/issues/3311 )

By the way, the currently used HTTP/json api is supposed to be for jaeger internal use only.

aljesusg:
I thought about that....and about moving kiali api to grpc (throw a ball ;)) to improve it... +1 on this if we can do it

lucasponce:
I'm not against to introduce gRPC clients in the backend Kiali query, but I won't mix that with the other backends (prometheus, kubernetes).

> By the way, the currently used HTTP/json api is supposed to be for jaeger internal use only.

Yes, this was somehow discussed with Jaeger team, being Kiali responsible to adapt potential changes.

aljesusg:
> I'm not against to introduce gRPC clients in the backend that Kiali queries, but I won't mix that with the other backends (prometheus, kubernetes).
> 
> > By the way, the currently used HTTP/json api is supposed to be for jaeger internal use only.
> 
> Yes, this was somehow discussed with Jaeger team, being Kiali responsible to adapt potential changes.

That's right...maybe we should create somekind of check fro that...I am thinking in a github action to check if jaeger https endpoints and model that we are using change...

PR #3601, opened 2021-01-19
Jaeger client: switch to gRPC/protobuf

jotak:
Fixes https://github.com/kiali/kiali/issues/3514

Doc PR: https://github.com/kiali/kiali.io/pull/348
Operator PR: https://github.com/kiali/kiali-operator/pull/249


jotak:
WIP: proof of concept looks OK, I have a working GRPC-based connection with Jaeger to fetch traces.
Now it needs to be generalized in our jaeger calls, remove the old http client / clean up code.
And measure performance.

jotak:
Pretty much done, but I totally haven't tested yet the authentication stuff. Since it's the first time we use grpc in Kiali, the related code hasn't been tested at all yet. (To do before merging.)

Next step is also to run some perf tests.

jotak:
I've done some comparisons that don't really show any performance gain...
(and yes that's a bit disappointing)

jotak:
So, I could finally get some better numbers:

Capture d’écran de 2021-01-21 11-16-38

The test runs queries continuously for 2 minutes with sleeps between queries. The query is crafted to be reasonably fast on the jaeger side (heavy queries would make storage latency preponderant against kiali-jaeger latency, hence hiding any improvement here).

So, while the p99 latency is quite comparable (~250ms on master versus ~230ms on grpc), the average latency is really better (~140ms on master versus ~80ms on grpc). Which allows a better throughput too (2.2ops versus 2.9ops - there's a sleep time between queries, throughput could be much higher)

lucasponce:
> So, I could finally get some better numbers:

Good :), but not really a concern.

One thing to take into consideration is that grpc introduces a strong dependency in the protobuf or client-library used by Jaeger.
I'll also check with @FilipB that this could works ok with upstream and potential versions packaged in downstream to avoid this situation forces to import different versions on each environment (probably not, just thinking loud potential areas to check).

Also, the Oauth config on downstream for Jaeger service is something that probably it would require a closer testing.

FilipB:
Tried with istio 1.8.2 and bookinfo. Following error thrown on Services->bookinfo->details->Traces page:
`jaeger GRPC client error: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"`

Tried with both `values.meshConfig.enableAutoMtls=false` and `values.meshConfig.enableAutoMtls=true` with the same result.

lucasponce:
Thanks @FilipB, feel free to add a revision as well, just to prevent that we don't merge it without notice and we don't skip it.

jotak:
@FilipB , I suspect you were using a jaeger install serving on https, right? Mine is http and I've found a bug indeed with https (fixed now). If not, then it was something else.
By the way, mesh TLS having no effect here is totally expected (unless you have jaeger deployed as part of the mesh)

FilipB (changes requested):
I still see the `jaeger GRPC client error: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"` error even with the `86bcccf` commit.

Jaeger is installed as add-on via - https://github.com/kiali/kiali/blob/master/hack/istio/install-istio-via-istioctl.sh#L298

jotak:
Hey @FilipB ,
Can you check if you get the same in master and it is actually a regression? Isn't it a misconfiguration you have?
Even with the hack script I get jaeger deployed as http/port 80 so you should have in your Kiali CR / CM something like `in_cluster_url: http://tracing.istio-system/jaeger` (no https)

jotak:
(squashed and rebased)

FilipB:
Not ideal scenario to compare perf results but still it's visible that gRPC (top of the picture) is slightly better. Looking at avg,GetWorkloadTraces it oscillates around 40ms for gRPC (this PR) and around 50ms for master.
Jaeger and kiali pods were running on different OCP worker nodes.

metrics

FilipB:
Tested this PR with OpenShift Service Mesh 2.0.1. It was necessary to remove the '-v 3' parameter from kiali container starting command in kiali deployment yaml (spec.template.spec.containers.command) to avoid following error during kiali pod startup:
```
flag provided but not defined: -v
Usage of /opt/kiali/kiali:
  -config string
    	Path to the YAML configuration file. If not specified, environment variables will be used for configuration.
```

Kiali pod was started but traces failed to be shown with following error:
`GetTraceDetail, Jaeger GRPC client error: rpc error: code = Unavailable desc = connection closed`

Kiali log does not show any error, only following info:
`2021-02-05T12:46:57Z INF Jaeger GRPC client info: address=jaeger-query.istio-system.svc:443, auth={/kiali-cabundle/service-ca.crt false f+fdlQbTcQ+EFDMrHm6DswMIvKkQkG33nL8uUwehHWNvSfYXPWCh3p8PFUPCO7Te0IItrEEqbpyHSTAB+TKwLBzI9gA21xcGe/VKQFJyAhDB59DRFNxtZ6pVIXMx1UIGJb9bWaQY47OG4fDnBDnkRG9vupA+gCcjag6sZHHKGaZMXoscJCOpzrSqz5BWf8ByASSyx18gzJ5uxf3deXJL9V4YGNz7M/xCaaanGLrOFGEaObvoVlrSa/iq0uc7GvBlnsoXGjFVmtaWitwYSfPVUA0yXQ1I8l9SH9Z0czaJXOzHA4gyKRujLEw0czgHXBkFLZ4qMHdOUvcSX615FBrx  basic false internal}`

jotak:
At the moment, I haven't found a solution to fix this issue with ossm. I've opened an issue on oauth-proxy, I think it doesn't support grpc (to be confirmed).

In any case I think we should either implement an hybrid solution where both http and grpc clients are possible (switchable from config), or leav it aside for a moment. I'm not super confident that switching to a grpc-only solution would work in every situation, every kind of customer setup, so I would prefer a more conservative approach with the two implementations living side by side. A question would remain: which one would be enable by default? I'd say the grpc one, in order to get as many feedbacks as possible.

jotak:
tracking issue on oauth-proxy: https://github.com/openshift/oauth-proxy/issues/205

jshaughn:
>  A question would remain: which one would be enable by default? I'd say the grpc one, in order to get as many feedbacks as possible.

I'd say maybe GRPC upstream and HTTP downstream, if possible.

jmazzitelli:
> I'd say maybe GRPC upstream and HTTP downstream, if possible.

We don't have build procedures in place today that would be able to have different configurations built into upstream vs. downstream binaries/configs. But there are two ways we could do this:

1) Have the OSSM Operator create the necessary config setting in the Kiali CR that it creates (thus the OSSM Operator will tell Kiali what setting it should use).

or

2) Have one default setting if the Kiali Operator sees it is installing Kiali in an OpenShift (or even a Maistra) environment, and a different setting otherwise (we already do this kind of thing for other settings - spec.auth.strategy for one).

jotak:
> 
> I'd say maybe GRPC upstream and HTTP downstream, if possible.

Yes, since that would be configurable, we can push a config specific to ossm

jotak:
Rebased + added some code comments about untested GRPC client with secured communication. I will add a faq entry to kiali.io too.

FilipB (approved):
Tested following:
- grpc is disabled by default for service mesh and enabled for istio
- use_grpc flag works as expected
- grpc client (use_grpc=true) is not working for service mesh (which is expected - ERR GetAppTraces, Jaeger GRPC client error: rpc error: code = Unavailable desc = connection closed)
- it works fine for both use_grpc=true/false in istio
- tested with both all-in-one and production-elasticsearch jaeger templates

PR #3602, opened 2021-01-19
Fix multi-cluster build

israel-hdez:
After merging https://github.com/kiali/kiali/pull/3603, the `multi-cluster` branch requires a small fix to build correctly.

Related kiali/kiali#3592

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -6,7 +6,6 @@ import (
 	"time"
 
 	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/clientcmd/api"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
@@ -49,12 +48,8 @@ type Cluster struct {
 // can be passed a nil value and a default function will be used.
 func NewMeshService(k8s kubernetes.ClientInterface, newRemoteClientFunc func(config *rest.Config) (kubernetes.ClientInterface, error)) MeshService {
 	if newRemoteClientFunc == nil {
-		// Mesh discovery will use ServiceAccount token instead of user token
-		token := ""
-		token, _ = kubernetes.GetKialiToken()
-
 		newRemoteClientFunc = func(config *rest.Config) (kubernetes.ClientInterface, error) {
-			return kubernetes.NewClientFromConfig(config, &api.AuthInfo{Token: token})
+			return kubernetes.NewClientFromConfig(config)
 		}
 	}
 


israel-hdez:
@jmazzitelli  I'm reverting back a few changes to the `NewClientFromConfig`. I hope this is fine.
If @mlbiam can also take a look, it will be great.

jmazzitelli:
> @jmazzitelli I'm reverting back a few changes to the `NewClientFromConfig`. I hope this is fine.
> If @mlbiam can also take a look, it will be great.

I'd wait for @mlbiam to give his thoughts - that deals with the new authInfo introduced with the new header auth strategy.

jshaughn (commented):
The reasoning behind the PR is good, but I don't know enough to say that this won't affect the behavior of the recent work that added this stuff.  So, minimally we should run the Molecule tests to ensure those scenarios still pass.

This could have been a PR against master and then we could have rebased multi-cluster, but I guess it's OK here.  I'll let @jmazzitelli  decide.

jmazzitelli:
I think you want this to go into "master" first (this PR is targeting the multi-cluster branch - but if this truly needs to be reverted, it should be done in the master branch - you can then move it over to the multi-cluster branch via a rebase or cherry-pick)

jshaughn:
@jmazzitelli We can't cherry pick because this has multi-cluster file changes.  @israel-hdez can open a new PR, though.  And then update this one after a rebase.

mlbiam:
in general i disagree with the idea of passing tokens instead of objects.  A token is very limiting and can hurt forward compatibility.  For instance what about storing additional metadata about tokens such as easier expiration checking for support tokens mounted from the TokenRequest api in the future?  most of the work (and debugging) came from moving from passing tokens to the authInfo object.  Now that's done I don't see why it's a good idea to revert one component to be inconsistent.

I'm confused by removing it from the `K8SClient`, the impersonation headers should be passed whenever the work needs to be done on behalf of the user.

israel-hdez:
@jmazzitelli @jshaughn @mlbiam Here is the PR targeting master: #3603 
I think I may close this one, because `multi-cluster` will need a rebase anyway.

israel-hdez:
@mlbiam Impersionation data is still passed, but it's now in `client_factory.go` which is one level up in the call stack.

mlbiam:
@israel-hdez ok, you know this code base better so i'll defer to you on if this is the right way to go

jshaughn (approved):
LGTM!

jshaughn:
@mlbiam Thanks for looking, of course let us know if you are in any way negatively impacted by these changes.

Issue #3207, opened 2020-09-14
Support reverse proxy injection of id_token and impersonation header


mlbiam:
**Is your feature request related to a problem? Please describe.**

The current OIDC implementation will pass the user's `id_token` to the API server, but requires that the API server is configured to use OpenID Connect.  Managed clusters, such as EKS, AKS, etc, do not support OpenID Connect.  The k8s dashboard gets around this issue by supporting reverse proxy authentication and the impersonation headers supported by kubernetes letting a reverse proxy take ownership of authentication.  This allows SSO for Kiali regardless of the type of cluster it's running on.

**Describe the solution you'd like**
I want to donate the code that would let Kiali support the same rules as the kubernetes dashboard:

1. TLS must be enabled on Kiali to support the `Authorization` header or the `Impersonate-*` headers
2.  If TLS is enabled, when the `Authorization` header is present it is passed through to the API server.  If, additionally, the `Impersonate-*` headers are present then those headers are passed through too

There's no need to validate the headers, it's the responsibility of the reverse proxy to create the correct headers and the API server to validate them.  Kiali is simply passing them through.  This makes Kiali more secure as it doesn't need a privileged service account.

**Describe alternatives you've considered**

OIDC is one alternative for OIDC enabled clusters.  This doesn't handle managed clusters though.  To handle managed clusters Kiali would need to implement token management for every cloud (ie EKS, AKS, GKE, DigitalOcean, etc...).  Kiali shouldn't gatekeep to certain platforms when a platform agnostic solution is available.

**Additional context**

Diagram of the deployment architecture:

![k8s_dashboard_impersonation](https://user-images.githubusercontent.com/8249283/93100509-3a6d3c80-f677-11ea-90c2-518dfb946d37.png)

similar issue for the k8s dashboard - https://github.com/kubernetes/dashboard/issues/3664 includes test cases

israel-hdez:
@mlbiam We will be happy to incorporate your code into Kiali.

mlbiam:
fantastic!  i'll get started

GoelDeepak:
+1 I was about to file a similar enhancement and found this thread. @mlbiam looking forward for your implementation

mlbiam:
Life's gotten in the way but hoping to get started later this week. Open to help testing?

GoelDeepak:
> Life's gotten in the way but hoping to get started later this week. Open to help testing?

Definitely. Let me know once you have something ready.

[...]

PR #3603, opened 2021-01-19
Rollback a few changes to NewClientFromConfig

israel-hdez:
These changes are, apparently, unneeded. Also, it's a error prone, because there were two tokens involved when creating the client. There was one in the `config` parameter, and another one in `authInfo` parameter. Only the token in `config` was being used. The `authInfo` is used only to pass impersonation data, but this data shouldn't always be honored because it's only valid when using the `header` strategy.

So, I think it's better to rollback these changes and pass the impersonation data already embedded in the `config` parameter when it's strictly needed. The `config` parameter already has impersonation fields, so the extra `authInfo` is, apaprently, redundant.

Related kiali/kiali#3207

### DIFF
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -9,7 +9,6 @@ import (
 	kube "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
-	"k8s.io/client-go/tools/clientcmd/api"
 
 	kialiConfig "github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
@@ -101,7 +100,7 @@ func NewKialiCache() (KialiCache, error) {
 		BearerToken:     cacheToken,
 		Burst:           config.Burst,
 	}
-	istioClient, err := kubernetes.NewClientFromConfig(&istioConfig, &api.AuthInfo{Token: cacheToken})
+	istioClient, err := kubernetes.NewClientFromConfig(&istioConfig)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


israel-hdez:
@jmazzitelli do we have molecules to test the new header strategy?
I'm wondering if we can simply run the tests, or whether this needs a manual test.

jmazzitelli:
> do we have molecules to test the new header strategy?

[Yes - here](https://github.com/kiali/kiali-operator/tree/master/molecule/header-auth-test). I'm running it now... also running the other tests dealing with auth strategies.

jmazzitelli:
How I am testing. First prepare minikube with dex and istio (using k8s-minikube.sh hack script), and then run this:

```
$ make \
  -e CONSOLE_VERSION=local \
  -e CLUSTER_TYPE=minikube \
  -e MOLECULE_SCENARIO="header-auth-test openid-test token-test" \
  -e MOLECULE_USE_DEV_IMAGES=true \
  clean build test cluster-push molecule-test
```

and the header-auth-test failed - it could not log in successfully.

So this broke something.

```
    TASK [Assert that we were able to get the list of namespaces] ******************
fatal: [localhost]: FAILED! => {
    "assertion": "kiali_output.json | length > 0",
    "changed": false,
    "evaluated_to": false,
    "msg": "Assertion failed"
}
```

jmazzitelli:
If you use k8s-minikube.sh to install minikube+dex you'll get a reverse proxy setup so you can simply set spec.auth.strategy to header and you'll be able to use it (log in with the admin@example.com user). You'll want to pass in `-dun all` to give that user access to all namespaces so you can log in:

```
hack/k8s-minikube.sh start --dex-enabled true -kv 1.18.0 -dun all
```

jmazzitelli:
This might be a false alarm. I may not have given my admin@example.com user correct permissions.

jmazzitelli (approved):
molecule tests pass

israel-hdez:
Merging...

PR #3610, opened 2021-01-22
[hack] be able to tell the script to get the images from a custom hub

jmazzitelli:
don't always assume the user wants gcr.io. This allows you to tell the script to get the images from the default hub.


jshaughn (approved):
LOK, unfortunate that we have to play repo roulette.

Issue #3611, opened 2021-01-22
api-test molecule test failure


jmazzitelli:
Need to see if something is really broke or if the test needs to be fixed.

```yaml
    TASK [Results: /namespaces/graph: https://192.168.99.100/kiali/api/namespaces/graph?namespaces=simple-mesh&graphType=workload&injectServiceNodes=false] ***
    ok: [localhost] => {
        "msg": {
            "duration": 600,
            "elements": {
                "edges": [
                    {
                        "data": {
                            "destPrincipal": "spiffe://cluster.local/ns/simple-mesh/sa/default",
                            "id": "2f7ace8d53d82d6644764e714bb773cc",
                            "isMTLS": "100",
                            "responseTime": "5",
                            "source": "e26a73e7dde82950527dec642ff24f5a",
                            "sourcePrincipal": "spiffe://cluster.local/ns/simple-mesh/sa/default",
                            "target": "279175c6df0b586b508aa39563cf4c53",
                            "traffic": {
                                "protocol": "http",
                                "rates": {
                                    "http": "0.82",
                                    "httpPercentReq": "100.0"
                                },
                                "responses": {
                                    "200": {
                                        "flags": {
                                            "-": "100.0"
                                        },
                                        "hosts": {
                                            "simple-server-svc.simple-mesh.svc.cluster.local": "100.0"
                                        }
                                    }
                                }
                            }
                        }
                    }
                ],
                "nodes": [
                    {
                        "data": {
                            "app": "simple-client",
                            "cluster": "cluster-default",
                            "id": "e26a73e7dde82950527dec642ff24f5a",
                            "isRoot": true,
                            "namespace": "simple-mesh",
                            "nodeType": "workload",
                            "traffic": [
                                {
                                    "protocol": "http",
                                    "rates": {
                                        "httpOut": "0.82"
                                    }
                                }
                            ],
                            "version": "v99",
                            "workload": "simple-client"
                        }
                    },
                    {
                        "data": {
                            "app": "simple-server",
                            "cluster": "cluster-default",
                            "destServices": [
                                {
                                    "cluster": "cluster-default",
                                    "name": "simple-server-svc",
                                    "namespace": "simple-mesh"
                                }
                            ],
                            "id": "279175c6df0b586b508aa39563cf4c53",
                            "namespace": "simple-mesh",
                            "nodeType": "workload",
                            "traffic": [
                                {
                                    "protocol": "http",
                                    "rates": {
                                        "httpIn": "0.82"
[...]

PR #3612, opened 2021-01-22
Fix Idle Node detection for multi-cluster

jshaughn:
When checking for Idle Nodes we need to incorporate the proper clusterName into the ID of the candidate node.

Fixes https://github.com/kiali/kiali/issues/3611

For QE Test:
With master:
1) Install using hack/istio/install-istio-via-istioctl.sh or otherwise ensure you configure with a meshId and clusterName
2) Generate bookinfo graph with traffic
3) Enable Display Idle Nodes

See unexpected Idle Nodes appear

Repeat with PR, no Idle Nodes should appear

### DIFF
--- a/graph/telemetry/istio/appender/idle_node.go
+++ b/graph/telemetry/istio/appender/idle_node.go
@@ -31,6 +31,7 @@ func (a IdleNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *g
 
 	services := []models.ServiceDetails{}
 	workloads := []models.WorkloadListItem{}
+	clusterName := graph.Unknown
 
 	if a.GraphType != graph.GraphTypeService {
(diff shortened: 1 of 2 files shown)


jmazzitelli:
To test the fix, install Istio using hack script to OpenShift cluster and then run this:
```
make \
  -e CONSOLE_VERSION=local \
  -e MOLECULE_SCENARIO="api-test" \
  -e MOLECULE_USE_DEV_IMAGES=true \
  clean build cluster-push-kiali molecule-test
```

jmazzitelli (approved):
molecule test now passes

jshaughn:
Merging prior to QE approval as the offended molecule test is passing and I want this in the 1.29 build.  Still asking for QE to validate at their leisure.

PR #3626, opened 2021-01-26
Adding host information to clusters and routes

xeviknal:
Adding formatted information about the services listed in Clusters and Routes.
Needs https://github.com/kiali/kiali-ui/pull/2063

### DIFF
--- a/models/config_dump.go
+++ b/models/config_dump.go
@@ -26,20 +26,20 @@ type Listener struct {
 
 type Clusters []*Cluster
 type Cluster struct {
-	ServiceFQDN     string `json:"service_fqdn"`
-	Port            int    `json:"port"`
-	Subset          string `json:"subset"`
-	Direction       string `json:"direction"`
-	Type            string `json:"type"`
-	DestinationRule string `json:"destination_rule"`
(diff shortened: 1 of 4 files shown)


hhovsepy (approved):
Verified in scope of UI PR.

models/config_dump.go
```diff
@@ -26,20 +26,20 @@ type Listener struct {
...
 }
 
 type Routes []*Route
 type Route struct {
-	Name           string `json:"name"`
-	Domains        string `json:"domains"`
-	Match          string `json:"match"`
-	VirtualService string `json:"virtual_service"`
+	Name           string          `json:"name"`
+	Domains        kubernetes.Host `json:"domains"`
```
lucasponce, line 40: Minor comment, "domains" is plural, perhaps the native type expects a comma separated list of domains ?

Just to confirm, in case that parsing may fail on that case.

xeviknal: Good point, I've re-check that, and it is a filter of a list of domains. However, I guess that one string may represent multiple domains, that is why it is kept in plural.
istioctl strategy is always to choose the shortest domains representation: '*.bookinfo', 'details.bookinfo'.

lucasponce (approved):

PR #3627, opened 2021-01-26
Implement keys for UI defaults from the CR

### LINKED ISSUE #3371: provide some UI defaults within the Kiali CR
Consider adding a section to the Kiali CR to define UI-specific defaults. Something like this might be useful:

```
spec:
  ui:
    duration_default: "5m"
    refresh_period_default: "30s"
    ... and any other defaults we can think of...
```

Perhaps we can put this under "kiali_feature_flags" (just a thought):

```
spec:
  kiali_feature_flags:
    ui_defaults:
      duration: "5m"
      refresh_period: "30s"
      ...

```
The server can provide an API endpoint similar to the `/config` endpoint (if not part of the json that /config itself returns). The UI can call this at startup to obtain defaults that it will then use. We already do this kind of things for kiali_feature_flags and the health config.

This could be useful for folks that have very large meshes and want the UI to default to a duration of, for example, 5m, refresh every 1m (rather than the default which "last 1m" and "every 15s").

cfcosta:
Closes #3371.

Please see: https://github.com/kiali/kiali-ui/pull/2064

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -295,8 +295,18 @@ type DeploymentConfig struct {
 // defaults to the namespace configured for IstioNamespace (which itself defaults to 'istio-system').
 type IstioComponentNamespaces map[string]string
 
+// Feature flags available from the CR
 type KialiFeatureFlags struct {
-	IstioInjectionAction bool `yaml:"istio_injection_action,omitempty" json:"istioInjectionAction"`
+	IstioInjectionAction bool       `yaml:"istio_injection_action,omitempty" json:"istioInjectionAction"`
+	UIDefaults           UIDefaults `yaml:"ui_defaults,omitempty" json:"uiDefaults,omitempty"`
+}
+
+// Describe default settings for the UI, like the default selected namespace or
+// auto refresh interval.
+type UIDefaults struct {
+	Namespaces          []string `yaml:"namespaces,omitempty" json:"namespaces,omitempty"`
+	MetricsPerRefresh   string   `yaml:"metrics_per_refresh,omitempty" json:"metricsPerRefresh,omitempty"`
+	AutoRefreshInterval string   `yaml:"auto_refresh_interval,omitempty" json:"autoRefreshInterval,omitempty"`
 }
 
 // ToleranceConfig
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -295,8 +295,18 @@ type DeploymentConfig struct {
...
+	IstioInjectionAction bool       `yaml:"istio_injection_action,omitempty" json:"istioInjectionAction"`
+	UIDefaults           UIDefaults `yaml:"ui_defaults,omitempty" json:"uiDefaults,omitempty"`
+}
+
+// Describe default settings for the UI, like the default selected namespace or
+// auto refresh interval.
+type UIDefaults struct {
+	Namespaces          []string `yaml:"namespaces,omitempty" json:"namespaces,omitempty"`
+	MetricsPerRefresh   string   `yaml:"metrics_per_refresh,omitempty" json:"metricsPerRefresh,omitempty"`
+	AutoRefreshInterval string   `yaml:"auto_refresh_interval,omitempty" json:"autoRefreshInterval,omitempty"`
```
jshaughn, line 309: Fine but even better if alphabetized.

jshaughn (commented):
Looks OK but not approving until finished with end-to-end review.

config/config.go
jmazzitelli, line 309:
```suggestion
	RefreshInterval string   `yaml:"refresh_interval,omitempty" json:"refreshInterval,omitempty"`
```

config/config.go
```diff
@@ -454,6 +464,11 @@ func NewConfig() (c *Config) {
 		},
 		KialiFeatureFlags: KialiFeatureFlags{
 			IstioInjectionAction: true,
+			UIDefaults: UIDefaults{
+				Namespaces:          make([]string, 0),
+				MetricsPerRefresh:   "1m",
+				AutoRefreshInterval: "",
```
jmazzitelli, line 470:
```suggestion
				RefreshInterval: "15s",
```

jmazzitelli (changes requested):

config/config.go
jmazzitelli: What are the "json:" fields for? I don't think we need those and this just makes things more verbose than necessary

jshaughn:
Closing this PR in favor of https://github.com/kiali/kiali/pull/3688.

PR #3628, opened 2021-01-26
fix logging calls

jmazzitelli:
I came across this looking at https://github.com/kiali/kiali/issues/3595. But since these are separate bugs, I'm committing as a separate PR (although this kinda addressed 3595 because that issue also reports the warning message with the bad logging). This fixes the logging in places where the "f" form of the the logging messages either should have been used or was used incorrectly.


jshaughn (approved):
Never good when the debugging help is buggy. LG

PR #3630, opened 2021-01-26
make error messages unique

jmazzitelli:
This one function has *four* identical error messages that could get logged. Because all four messages are identical, when you see the log message in the pod logs, it isn't immediately easy to figure out where the error was happening.

This PR makes the messages unique so you know what failed immediately when you read the message.


jshaughn (approved):
LOK

Issue #3634, opened 2021-01-27
Istio "config_map_name" not used


joscha-alisch:
**Describe the bug**
When installing kiali via the kiali-operator, it looks like the configuration option `external_services.istio.config_map_name` is not used. We get the following log message when starting up
```
Failure when checking if mesh-id is configured: configmaps "istio" not found
```
even though we set the following configuration in the kiali CR:
```yaml
external_services:
  istio:
    config_map_name: istio-ver-1
```

I made sure that this is indeed the name of the ConfigMap resource in istio-system and that in the actual kiali configmap this setting is configured. Lastly, I enabled trace logging which shows the kiali config on startup and also there it says `config_map_name: istio-ver-1`. 

It might be important for debugging that kiali is running in a different namespace than istio, but I tried it in the same namespace as well with the same error.

**Versions used**
Kiali: v1.29.0
Istio: 1.7.3
Kubernetes flavour and version: GKE v.1.19.6-gke.600

**To Reproduce**
1. Install istio, see that the configmap is named "istio-ver-1"
2. Install kiali with the configuration option `external_services.istio.config_map_name` set to `istio-ver-1` 
3. Open the kiali UI 
4. See error in the log output of the kiali server `Failure when checking if mesh-id is configured: configmaps "istio" not found`

joscha-alisch:
I now also tried to disable the kubernetes caching in case that has anything to do with it, with no improvement.

jmazzitelli:
The code does appear to at least attempt to use that setting: https://github.com/kiali/kiali/blob/v1.29.0/business/mesh.go#L97

joscha-alisch:
Yes, this is what I found as well which confuses me immensely. I think that `configmaps "istio" not found` is the error returned by the native kubernetes clientset, which suggests that for some reason `config.IstioConfigMapName` is `istio` in that line you posted. I.e. not what I actually configured 😕

jmazzitelli:
what do you see if you curl the /api/config endpoint? It will tell you what it thinks the config_map_name is. In my case, I set `spec.external_services.istio.config_map_name` to `mazzwashere` and this is what I see:

```
$ curl -k http://localhost:20001/kiali/api/config
{
  "clusterInfo": {
    "name": "cluster-default",
    "network": "network-default"
  },
...
  "istioConfigMap": "mazzwashere",
...
  }

[...]

PR #3636, opened 2021-01-27
Use the configured name for the Istio ConfigMap - do not use the constant

jmazzitelli:
The constant isn't guaranteed to be the name of the istio configmap, so do not use it. This PR removes it as well as the one reference to it.

fixes: https://github.com/kiali/kiali/issues/3634
fixes: https://github.com/kiali/kiali/issues/3637

backport PR: https://github.com/kiali/kiali/pull/3638

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -94,7 +94,7 @@ func (in *MeshService) IsMeshConfigured() (isEnabled bool, returnErr error) {
 	isEnabled = false
 	cfg := config.Get()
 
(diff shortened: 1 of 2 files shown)


xeviknal:
@jmazzitelli good catch. The config map name can't be a constant anymore. Since the introduction of istio canary deployments, istio config maps might be renamed to be coupled to it's canary. Plus, maistra community is using the canary naming convention for the default installation (something like `istio-default`).
Good!

xeviknal (approved):

jshaughn (approved):
LOK

Issue #3644, opened 2021-01-28
operation node graph is missing the operation node [regression]


jshaughn:
This looks like a regression between v1.28 and v1.29, likely due to the multi-cluster work.

vbergbauer:
As of Kiali 1.30.0, the problem persists.
I am using Istio 1.9

If I just replace the container image with 1.28.1 (leaving everything as it was), the service nodes appear again.
If I make the exact same query against the two versions, both succeed, but only v1.28.1 returns aggregates:
```bash
curl "http://kiali.example/api/namespaces/graph?duration=60s&graphType=app&injectServiceNodes=true&namespaces=httpbin"
```
Results for v1.28.1:
```json
    // ...
      {
        "data": {
          "id": "5dbc49f0f0cbdd7d04399a4db59b6237",
          "nodeType": "aggregate",
          "namespace": "httpbin",
          "app": "httpbin",
          "service": "httpbin",
          "aggregate": "request_operation=PostStatus",
          "traffic": [
            {
              "protocol": "http",
              "rates": {
                "httpIn": "1.10",
                "httpOut": "1.10"
              }
            }
          ]
        }
      },
// ...
```
But no aggregates nodes for v.1.30.0.

PR #3645, opened 2021-01-28
Fix issue with aggregate node detail graphs

jshaughn:
This fixes a regression in Kiali v1.29

Fixes https://github.com/kiali/kiali/issues/3644

To test:
1) using istio 1.8, deploy travel demo via hack script, using the enable-operation-metrics option
    - note, you may need to have istio configured for multi-cluster, I'm not sure.

`> hack/istio/install-travel-agency-demo.sh -eo true`

2) in the UI select the travel namespaces and also enable the Show Operation Nodes Display option

3) double-click on an operation node to drill into the node detail graph

pre-fix: the operation node will be missing from the detail graph
post-fix: the operation node will appear as expected


jmazzitelli (approved):
lgtm

PR #3647, opened 2021-01-28
[hack] fix up the jwt hack utilities so they work

jmazzitelli:
I was using these utilities to debug some auth issue and needed to fix them up to work with OpenShift


jshaughn (approved):
LOK

Issue #3650, opened 2021-01-29
Metrics page: segregate peer by namespace


jotak:
In our testing setup we have a single server (`s1-service-a)` and three consumers. two consumers in the same namespace, `s1-client-x` and `s1-client-y`. There is another namespace that also contains a consumer, also called `s1-client-y`. All three consumers do 1RPS continuously. The Graph view correctly distinguishes between the different consuming workloads and adds the namespace identifier. However, in the "Inbound metrics" tab of `s1-service-a`, I only see two consumers, and the load from services called `s-client-y` are added to each other. At the least this is inconsistent. Personally, I consider this a bug and the inbound metrics should distinguish using the namespace as well.

I've added screenshots of the situation as well.
<img width="1710" alt="compounded-metrics-graph-view" src="https://user-images.githubusercontent.com/54579/106192859-066d6080-61ad-11eb-9110-8efe5f97bfb4.png">

<img width="1710" alt="kiali-workloads-view" src="https://user-images.githubusercontent.com/54579/106192868-08cfba80-61ad-11eb-8242-2e6a1673b76d.png">
<img width="1710" alt="inbound-metrics-view" src="https://user-images.githubusercontent.com/54579/106192873-0a00e780-61ad-11eb-95f6-7bf885bbf557.png">

_Originally posted by @spockz in https://github.com/kiali/kiali/discussions/3646_

jotak:
fix on its way, this will have to be backported on 1.29

PR #3651, opened 2021-01-29
Metrics: fix peer namespace segregation

jotak:
Add *_workload_namespace as an available label to group by

Fixes https://github.com/kiali/kiali/issues/3650

Frontend PR: https://github.com/kiali/kiali-ui/pull/2073


hhovsepy (approved):

jshaughn (approved):
@jotak, nice change.  It speaks well to the code architecture that the enhancement/fix was not disruptive.

models/dashboards.go
```diff
@@ -111,6 +111,10 @@ func buildIstioAggregations(local, remote string) []Aggregation {
 			Label:       fmt.Sprintf("%s_canonical_revision", local),
 			DisplayName: "Local version",
 		},
+		{
+			Label:       fmt.Sprintf("%s_workload_namespace", remote),
+			DisplayName: "Remote namespace",
```
spockz, line 116: I'm unfamiliar with Istio parlance but to me "Remote" sounds like a namespace in a different cluster. The test above uses `source` as well in `source_workload_namespace`.

jotak: "Remote" here is really like a synonym for "Peer", nothing related to clusters. We could write "peer" but just as a matter of consistency I won't do it at least in this PR, because this is the terminology we currently have (renaming would impact other places). Note that the "source_workload_namespace" / "destination_workload_namespace" are the prometheus labels to be grouped by (using "source" or "destination" depends whether we are displaying inbound or outbound metrics, and in both case it will be referred to as "Remote namespace" in Kiali)

Issue #3642, opened 2021-01-28
Gap between rate info collected in graph versus health


lucasponce:
In the context of a feature review of work done in PRs [1] and [2] I've found a potential bug that it is also present in master branch.

Use the Error Rates Demo [3] in a cluster, I used minikube but I think this can be reproduced in any env.

The error rate information collected by graph and health endpoints seem to have a mismatch that affects when user is trying to set up different Health config (or not modifying any health config, but just looking into the details and he/she may realize that info doesn't 100% match).

In the "alpha" namespace, if we focus on the "w-server" we see the inbound rate gap:
![image](https://user-images.githubusercontent.com/1662329/106094576-873b4680-6132-11eb-8d64-e07f0c620c24.png)

Health shows a 32% vs Traffic sidepanel shows a 14%

Let's take a closer look into the backend data.
These are the queries used for this calculation, both in graph and health that happens at same time:

```
GET http://localhost:8000/api/namespaces/graph?duration=60s&graphType=app&includeIdleEdges=false&injectServiceNodes=true&boxBy=app&appenders=deadNode,sidecarsCheck,serviceEntry,istio&namespaces=alpha

GET http://localhost:8000/api/namespaces/alpha/health?type=app&rateInterval=60s
```

Graph result:
```
...
    "edges": [
      {
        "data": {
          "id": "de7d3d1d23e214f06cc958ea48c4e1f6",
          "source": "1a04807ad43c689efe3e236814b06a54",
          "target": "941d852ac3644b2d2e77ddb35cc40798",
          "traffic": {
            "protocol": "http",
            "rates": {
              "http": "1.93",
              "http5xx": "0.29",
              "httpPercentErr": "15.0",
              "httpPercentReq": "100.0"
            },
            "responses": {
              "200": {
                "flags": {
                  "-": "85.0"
                },
                "hosts": {
                  "w-server.alpha.svc.cluster.local": "85.0"
                }
              },
              "503": {
                "flags": {
                  "URX": "15.0"
                },
                "hosts": {
                  "w-server.alpha.svc.cluster.local": "15.0"
                }
              }
            }
          }
        }
      },
...
```

Health result:
```
...
  "w-server": {
    "workloadStatuses": [
      {
        "name": "w-server",
        "desiredReplicas": 1,
        "currentReplicas": 1,
        "availableReplicas": 1,
        "syncedProxies": 1
      }
    ],
    "requests": {
      "inbound": {
        "http": {
[...]

PR #3653, opened 2021-01-29
Fetch health request rate from a single reporter

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3642

Health endpoints provide request rate details since Health Config feature was introduced.

This created a bug in the request rate info used on health as inbound and outbound data were accumulated twice by source and destination reporter.

This PRs will use source reporter for outbound info, and it will combine source/destination reporter for inbound data.

The previous bug creates strange situations, clearly visible when using scenarios with errors where Health Config is being tested.

This fix is also needed to continue the effort of Health Config feature and in particular to continue reviewing/testing https://github.com/kiali/kiali/pull/3362


jshaughn (commented):
I think this code will probably do what is intended, although it may just be sufficient to use source reporting for inbound and destination reporting for outbound.  The traffic graph uses source reporting except for cases where we know there is no source reporting.  Node graphs are generated from the perspective of the node, and therefore use source reporting for outbound and destination reporting for inbound.

Overall, the duplicate telemetry generated by Istio's source and destination reporting is both a waste of resources and a programmatic nuisance.  It has been actively proposed to eliminate the duplicate reporting by limiting source reporting to only the telemetry that fails to route to route to, or reach,  a destination.  The duplication occurs on requests for which a destination workload returns a response.  Until Istio makes the change,  or until Kiali actually extracts its own requisite data, we will have to keep playing these games.

I think it could be a good thing if Kiali removed the notion of the reporters entirely and manipulated the metrics as if Istio did not generate duplicate telemetry.  But that would be a fairly large undertaking.

lucasponce:
Please @hhovsepy would you mind to check any regression on this PR ?

lucasponce:
> I think it could be a good thing if Kiali removed the notion of the reporters entirely and manipulated the metrics as if Istio did not generate duplicate telemetry. But that would be a fairly large undertaking.

@jshaughn In general I agree. But an additional problem that we have spotted is that when traffic contains errors, there is no duplication telemetry, but different telemetry as results are not the same, so user may have incorrect information about what is happening.

This PR is more a fix at least to reduce the problem and adjust the goal of the "health" endpoint rather to change or start a bigger discussion about source/destination telemetry.

Again, we are reviewing the health config effort, and this PR is created in that context, I'd ask reviewers to take a look in that context.

I know is tempting but I avoid to open a big topic about if Kiali should completely remove the notion of reporter (again, nothing against that, just focusing the context of this).

models/health.go
```diff
@@ -87,18 +94,69 @@ type ProxyStatus struct {
...
+	default:
+		log.Tracef("Inbound metric without reporter %v ", sample)
+		aggregate(sample, in.Inbound)
+	}
 }
 
 // AggregateOutbound adds the provided metric sample to internal outbound counters and updates error ratios
 func (in *RequestHealth) AggregateOutbound(sample *model.Sample) {
-	aggregate(sample, in.Outbound)
+	// Outbound traffic will be aggregated per source reporter
```
jotak, line 120: I am annoyed by the absence of symmetry between how inbound and outbound metrics are processed. Inbound and outbound are really the same metrics, just the point of view differs, so if we had an issue with the inbound metrics, I would imagine the same issue exists with the outbound metrics

models/health.go
```diff
@@ -87,18 +94,69 @@ type ProxyStatus struct {
...
+	reporter := string(sample.Metric[model.LabelName("reporter")])
+	if reporter == "source" {
+		aggregate(sample, in.Outbound)
+	}
+}
+
+// RequestHealth interanlly stores Inbound rate separated by reporter
```
jotak, line 127: nit/typo: "internally"

jotak:
One thing that I'd like to know is if the missing metrics for client-side reporting is a bug in istio or envoy that we can expect to be fixed reasonably soon, or if it's really something that we have to deal with.
We know that the telemetry is not entirely symmetric between reporters, but having missing client-side telemetry for certain kind of errors is more unexpected and looks like a bug.

models/health.go
lucasponce: Again, I'll try to detach the global discussion about reporters, but narrow that on health endpoint only.
Otherwise we won't progress.

Today, Outbound metrics are reported used source, but Inbound metrics use both.
You can see this in graph sidepanels and traffic tabs.

The goal of this fix was to address the health endpoint, in that we were aggregating both, that's a problem.

What to use ?

Well, for Outbound metrics is a consensus to use the source reporter (I don't enter in the discussion if that's good or bad, yet).

For Inbound metrics, there are more logic, in the "heatlh" context as results may differ I think using the max could help to spot a problem.

I could apply same logic in the Outbound as well, but I don't think we can't duplicate same logic in graph.

To me is to choose between:
a) Fixed reporter
b) Simple logic to choose on Inbound, Outbound or both.

As commented, this problerm is more visible on scenarios which errors, where the users see one thing in one reporter and a different one in another.

On NO ERRORS, probably it's identical info, in scenarios with ERRORS I think it could be interesting to show in the healh that an element have more errors (which probably will be more close to the reality).

I insist, but this PR is not a generic reporter source/destination, it should be tied to the health enpoint.

To change the reporter in graph, traffic or any other area perhaps is a deeper task that I didn't want to tackle first.

lucasponce:
I think you'll see better the details with these captures:

Note that reality is that w-server is generating around 50% of errors, but this is what we get from the telemetry:

From w-server workload perspective:

image
image


From v-server workload perspective:
image
image


So, with this fix, Health is aligned in Inbound case with the worst scenario (also showed in the bottom chart in the side panel).

In the Outbound info from v-server, information doesn't match with the reality but at least is consistent.

Again, I don't want to discuss which reporter is good or less good but:

- Resolve the duplication (that's done).
- Better sync the health with the information that user is seeing.
- Try to align with the reality.

So this is why the Inbound logic in the health takes the worst case.

jotak:
To me it's fine to go on and merge this, but there should be two more actions IMO:

- Create a follow-up task, because as we discussed this is a partial fix, it makes the destination node health aligned with what is shown in the graph, however, it creates a new discrepancy between source node health and destination node health. So the source node health still has to be fixed, and so does the side panel metrics for source node. (Moreover that new discrepancy might now be visible also in more usual and valid cases such as fault injection)

- Bring this to the attention of Istio/Envoy teams (open an issue). We went from a situation where we could trust client-side metrics to situation where we cannot, and that's what triggered this issue in the first place.

jotak (approved):

lucasponce:
Thanks @jotak I agree and stand your suggestions.
I'll wait if @hhovsepy has any comment in terms of regression before to merge.

hhovsepy (approved):
UI regression pass.

lucasponce:
cc @aljesusg, would you mind to rebase your PRs with master (backend and frontend) to include this (and other UI changes).

I'll continue the review of the health config work.

Thanks !

aljesusg:
> cc @aljesusg, would you mind to rebase your PRs with master (backend and frontend) to include this (and other UI changes).
> 
> I'll continue the review of the health config work.
> 
> Thanks !

Done !

lucasponce:
This would be nice to be backported to 1.29 branch.

(Note to self)

Issue #3605, opened 2021-01-20
Additional Istiod Pod Causes Unhealthy Pod Sync Status For Kiali


asummers:
**Describe the bug**
Having two `istiod` pods causes `unhealthy` status to be reported in Kiali with reason `proxy not synced`, when `istioctl proxy-status` reports synced due to different read targets.  

**Versions used**
Kiali: 1.28.1
Istio: 1.7.2
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9) Rancher On Prem 1.18.6

**To Reproduce**
1. In my case a hardware issue caused the second istiod to exist, but anything that creates a second istiod should be able to reproduce.
2. Run `istioctl proxy-status` and notice SYNCED status because it is reading from new istiod
3. Go to Kiali and notice all applications have unhealthy yellow border 
4. Drill into any and notice pod status unsynced because it is reading from old istiod
5. Remove extra istiod
6. Notice situation resolved in Kiali

**Expected behavior**
Kiali more gracefully handles duplicate istiod.

jshaughn:
I guess if Kiali could detect and report on this situation specifically it would be good.

xeviknal:
Also adding this info:

> `replicas: 2` is the default number when using ASM[1]
When running the ASM, the istiod deployment have 2 replicas.
We should cover this scenario too.

[1] ASM (Anthos Service Mesh)

gfthybridlabs:
Thanks @xeviknal this is certainly the case for the embedded ASM Operator profiles in istioctl (asm-gcp, asm-multicloud etc.) not in the "usual" profiles (default, demo).

The ASM profiles setup an HPA for the istiod pod:-
 ```
$ k get hpa -n istio-system
NAME                   REFERENCE                         TARGETS   MINPODS   MAXPODS   REPLICAS   AGE
istio-egressgateway    Deployment/istio-egressgateway    5%/80%    1         4         1          5h12m
istio-ingressgateway   Deployment/istio-ingressgateway   6%/80%    2         5         2          5h12m
istiocoredns           Deployment/istiocoredns           40%/80%   1         5         1          5h12m
istiod-asm-176-1       Deployment/istiod-asm-176-1       2%/80%    2         5         2          5h12m
```

xeviknal:
Complimenting the description, the additional pod in the environment was in an unhealthy state. Meaning that it was called to get the proxy-status but wasn't responding.

@gfthybridlabs Does your current kiali installation have problems?
[...]

PR #3654, opened 2021-01-29
Proxy status: fetching status only from running istiod pods + always returning status if present

xeviknal:
fixes https://github.com/kiali/kiali/issues/3605

1. The proxy-status is only fetch from pods in the `Running` state.
2. Kiali returns the proxy-status even if there is one non-responding `istiod` (following the `istioctl` approach). The proxy-status then might be partial because each `istiod` pod controls different set of envoys (e.g. `istiod1` controls `details` and `productpage`, `istiod2` controls `ratings` and `reviews`). If this scenario happens, then the Istio Component Status modal in the mast-head will be in red. Hence, it'd mean that the mesh is unstable and the admin/op needs to act.
3. New logs has been added to better troubleshoot this scenarios.

### DIFF
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


hhovsepy (approved):
Regression test passed.

kubernetes/istio.go
```diff
@@ -214,26 +215,64 @@ func (in *K8SClient) GetProxyStatus() ([]*ProxyStatus, error) {
 		return nil, errors.New("unable to find any Pilot instances")
 	}
 
+	wg := sync.WaitGroup{}
```
lucasponce, line 218: Note, I've fetched this PR locally and I guess I missed some review in the past.
This method is always fetching Pods from the wire even if that info is cached.

I think the business layer should combine and pass the pods to this method
.

lucasponce (approved):

lucasponce:
It looks ok.
I propose some refactoring on that area, to pass the Pods to the kubernetes package from the business side when those pods are cached.
Otherwise, Kiali is bypassing caching for this logic.

But I'm fine if this refactoring is done in a followup PR.

kubernetes/istio.go
xeviknal: I don't see why we should use the pods from the cache. Using them may lead to scenarios where kiali fetches proxy status from old `istiod` that are either dead or in a not running phase. I think that the most reliable way to perform this operation is to use fresh `istiod` that at this moment are healthy.
What scenario am I missing?

lucasponce: Well, it was more to follow same pattern of usage.
But it's an optimization, I understand that when proxy is cached, queries hit the cache for high level info and this fetch to the raw pods happens when building the new proxy status from the istiod status.

PR #3655, opened 2021-01-29
fix logout

jmazzitelli:
perform a logout correctly by converting the access token to the corresponding OAuthAccessToken name and deleting it.

Molecule test PR: https://github.com/kiali/kiali-operator/pull/242
backport 1.24 PR: https://github.com/kiali/kiali/pull/3661
backport 1.12 PR: https://github.com/kiali/kiali/pull/3662

fixes: https://github.com/kiali/kiali/issues/3595

This PR uses the same mechanism that the OpenShift Console itself uses:

https://github.com/openshift/console/blob/79c6d1118674cb1a2c8dcfcfd1797f25492e699b/pkg/server/server.go#L581-L593
https://github.com/openshift/console/blob/79c6d1118674cb1a2c8dcfcfd1797f25492e699b/pkg/server/server.go#L607-L613
https://github.com/openshift/console/blob/79c6d1118674cb1a2c8dcfcfd1797f25492e699b/pkg/server/server.go#L228
https://github.com/openshift/console/blob/79c6d1118674cb1a2c8dcfcfd1797f25492e699b/frontend/public/module/auth.js#L70-L84

### DIFF
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -1,8 +1,10 @@
 package business
 
 import (
+	"crypto/sha256"
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


jmazzitelli:
This is how you test this to make sure things are working.

1. Get a list of all the OAuthAccessToken resources in your cluster (these are all active sessions - users that are logged into the cluster): `oc get oauthaccesstokens.oauth.openshift.io`
2. Log into Kiali
3. Get the list of OAuthAccessTokens again - take notice that you now have one additional one compared to the list you got from step 1. Make a note of the name of that new  OAuthAccessToken.
4. Log out of Kiali
5. Get the list of OAuthAccessTokens for a third time. You can now confirm this PR works when you see that additional new token is now gone. It has been deleted when you logged out of Kiali. Thus, the Kiali session has been successfully killed.

jmazzitelli:
backport PRs:
* 1.24: https://github.com/kiali/kiali/pull/3661
* 1.12: https://github.com/kiali/kiali/pull/3662

business/openshift_oauth.go
```diff
@@ -203,8 +205,15 @@ func (in *OpenshiftOAuthService) Logout(token string) error {
 		return fmt.Errorf("could not connect to Openshift: %v", err)
 	}
 
+	// convert the access token to the corresponding oauthaccesstoken resource name
+	// see: https://github.com/openshift/console/blob/9f352ba49f82ad693a72d0d35709961428b43b93/pkg/server/server.go#L609-L613
```
jshaughn, line 209: Do we need to note that this changed in OpenShift 4.6 and so if run on earlier versions the new approach will fail like the old approach does on 4.6+?   Should we silently try the old delete if the new one returns an error?

jmazzitelli: we could try to silently delete if this fails. I can do that. Actually, let me just do the same thing I did in the 1.12 PR for all of these. That way is one code base and will work everywhere.

jmazzitelli: this is done. I'll update the backport PRs now so they use the same code.

jshaughn (approved):
Good investigation and back compat fix.

jshaughn:
@mattmahoneyrh We merged this as it will be tested downstream via Jira.

Issue #3656, opened 2021-01-31
Services are filtered based on Selectors and not on Labels


daur88:
**Describe the bug**
When fetching `fetchNamespaceApps`, services are collected based on wrong filter. 

**Expected behavior**
Kiali should not inspect `labelSelector` of the Services, but should inspect `labels` of the Services instead.

lucasponce:
Service's labels may be different from workload/pod labels, and hence, potentially create more misunderstanding when a service is not linked with a particular app.

I know that there is some duplication in the models used, Kiali decided to tie to the selector to have a strong connection with the pods -> workload which is the source of the telemetry in Istio.

Could you describe an example of your use case, just trying to learn how we can solve this gap without breaking any logical connection between application/service/workload.

Thanks !

daur88:
I see, now that i've learned a bit more, my initial decision does not seem that right, either. A little backstory here, i have a controller (workload) with label `app: myapp`, pods with label `app: myapp`, and a service with label `app: myapp`. However my service does not target pods based on the `app` label, for that i use my labels of my own. `app: myapp` is there merely to conform with Kiali's requirements. Right now, without this entry in my services' selectors, UI responds as if my application does not have any services (http://localhost:20001/kiali/console/namespaces/default/applications/whatever-app). That is the problem i want to solve.

Right now i assume, we can track services down if we start from pods. If we filter out from all pods, those with label `app: myapp`, then we can perform an inner join of this list with the list of available endpoints. This will give us a reduced list of endpoints that are attached to the services we want. This way we then can lift the requirements of having `app: maypp` in both service's selectors and labels - only pods' and workloads' label will matter.

jotak:
Filtering from pods, then tying pods to services also has a drawback: it only works when you have some pods. There's a number of situations where you wouldn't have any (serverless, failing to start, etc.)

What about this alternative: filter from selector **and** from labels? @lucasponce , do you see any drawback doing this?

jshaughn:
Note that using the `app` label is not really "Kiali's requirement" but is inherited from Istio.

lucasponce:
> What about this alternative: filter from selector and from labels? @lucasponce , do you see any drawback doing this?

I think we can augment the use of labels in that direction, but using the label selector is something we will need in one or another way.

daur88:
> Note that using the `app` label is not really "Kiali's requirement" but is inherited from Istio.

Thanks for correcting on that :+1:

> What about this alternative: filter from selector and from labels?

[...]

PR #3657, opened 2021-02-01
Augment Service filtration

daur88:
Service filtration is now based on "selectors vs labels", and not "selectors vs selectors"

** Describe the change **

Service filtration is now based on "selectors vs labels", and not "selectors vs selectors"

** Issue reference **

3656

** Backwards incompatible? **

No

### DIFF
--- a/kubernetes/filters.go
+++ b/kubernetes/filters.go
@@ -68,7 +68,7 @@ func FilterPodsForController(controllerName string, controllerType string, allPo
 func FilterServicesForSelector(selector labels.Selector, allServices []core_v1.Service) []core_v1.Service {
 	var services []core_v1.Service
 	for _, svc := range allServices {
-		if selector.Matches(labels.Set(svc.Spec.Selector)) {
+		if selector.Matches(labels.Set(svc.Spec.Selector)) || selector.Matches(labels.Set(svc.ObjectMeta.Labels)) {
 			services = append(services, svc)
 		}
 	}


lucasponce (changes requested):
I'm not sure if this is 100% correct.
Otherwise you'll pull pods that are not exposed by a Service.

I understand the intention as described here:
https://github.com/kiali/kiali/issues/3656#issuecomment-770656354

I suggest if we can work in some examples in the issue to understand in which scenarios a service label will be different to a service selector, and potential consequences in the model.

Note that Istio telemetry is tied to the workload (it is reported from the proxy deployed) so that's one of the reasons of this choice.

I agree that there is a gap between labeling the service or deployment and selector, but as this may introduce some changes, I think a discussion about it would benefit to build a solid model.

Again, thanks a lot for opening the issue and proposing a PR !
But on this case it was not a bug but a decision made on purspose due all the context I've commented, happy if you can participate with us to discuss it.

daur88:
Following the discussion in https://github.com/kiali/kiali/issues/3656#issuecomment-770656354, 
I have made the changes that will keep `selectors vs selectors` comparison, and will append the available services list with the result of `selector vs labels` comparison.

jotak:
Testing, it doesn't work for me, but that's because the services are filtered (again) in function `castAppDetails` with their Selector.
To be honest, the code in `fetchNamespaceApps` should be rewritten a little bit. The same function being used for quite different purposes isn't nice. If you don't mind I'll add a commit to your PR

lucasponce:
Not sure if before this PR progress we have thought that telemetry will differ if it collects app elements from different sources.
https://github.com/kiali/kiali/issues/3656#issuecomment-771463890

I mean, this will work if the app label is also located in the service AND selector, otherwise the application details show a list of services and workloads that won't match the telemetry collected.

I'm not opposed to discuss about this but I'd like to see the picture complete, not only to modify one tab.

jotak:
I think it's the user responsibility to guarantee that the pods covered by the service still have the correct app label, even if they aren't selected by that label. This is what I understand from @daur88 's comment here: https://github.com/kiali/kiali/issues/3656#issuecomment-770725424 . The purpose here is to cover the case where service doesn't select pods out of app label, which I think is a totally valid scenario wrt. istio.

@daur88 perhaps you could post an example of the service labels and selector + pod labels, to clarify the case?

lucasponce:
>  even if they aren't selected by that label. This is what I understand from @daur88 's comment here: #3656 (comment) 

That sounds valid to me, perhaps a little bit cornercase but valid.

Yes, +1 to an example, perhaps I image potential situations where Kiali could get "broken" with data that doesn't match and that's why I'm adding my comments.

jotak:
Yes, I agree that it's more easy to make kiali look broken if the user messes up with the labels

daur88:
@jotak , @lucasponce  thank you for taking time to tackle this issue.

One example i can give is this. Imagine I have following resources:

```yaml
kind: Deployment
metadata:
  labels:
    app: "app-name-that-i-use-for-istio-kiali-only"
    app.kubernetes.io/name: "my-app"
    app.kubernetes.io/instance: "my-release"
spec:
  selector:
    app.kubernetes.io/name: "my-app"
    app.kubernetes.io/instance: "my-release"
  template:
    metadata:
      labels:
        app: "app-name-that-i-use-for-istio-kiali-only"
        app.kubernetes.io/name: "my-app"
        app.kubernetes.io/instance: "my-release"
---
kind: Service
metadata:
  labels:
    app: "app-name-that-i-use-for-istio-kiali-only"
    app.kubernetes.io/name: "my-app"
    app.kubernetes.io/instance: "my-release"
spec:
  selector:
    app.kubernetes.io/name: "my-app"
    app.kubernetes.io/instance: "my-release"
```
In our setup we have many charts like this. They have been initially built with helm-recommended labels, but when we started adding istio and kiali into our setup we came across this case. Of course we can uninstall all the charts, add new selector and install everything back, but that is something we would rather avoid. 

Right now kiali and istio work fine. My problem is with one page in the kiali-ui (http://localhost:20001/kiali/console/namespaces/default/applications/app-name-that-i-use-for-istio-kiali-only). The page with application details does not show the service (since its selector lacks `app` label). 

However the very same service is listed fine in the all services page (http://localhost:20001/kiali/console/services). This makes me think, this PR is more about `frontend-ui` <-> `backend-api` communication, it should not affect telemetry data collection.

jmazzitelli:
> ```yaml
>     app: "app-name-that-i-use-for-istio-kiali-only"
>     app.kubernetes.io/name: "my-app"
>     app.kubernetes.io/instance: "my-release"
> ```

I have no idea if this is relevant (I haven't followed this discussion closely as to what the issue is being reported), but I just want to make it known that you CAN tell Kiali to look at the `app.kubernetes.io/name` as opposed to `app` by simply setting these values in your Kiali CR:

https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L626-L635

So you don't need that `app: "app-name-that-i-use-for-istio-kiali-only"` - you can set `spec.istio_labels.app_label_name:   app.kubernetes.io/name` and Kiali will look for that label.

You would also need to tell Istio to use/look for that label too IIRC, I don't remember the steps to do that, but I know its possible.

lucasponce:
> You would also need to tell Istio to use/look for that label too IIRC, I don't remember the steps to do that, but I know its possible.

Correct.

It sounds the motivation of this work is to align in the same label used to define an "app", I agree that user should decide that and propagate that defaults on the kubernetes cluster, Istio and Kiali. 

No need to deal with augment labeling filtering if that's the case, I'd prefer to work to align or to check how to make it easier to change them instead.

lucasponce:
Hi @daur88,

What do you think of latest comments on this topic ?

https://github.com/kiali/kiali/pull/3657#issuecomment-775945572

I wonder if that would solve your scenario ? Augmenting filtering of workloads on this way it may affect to the metrics collected per app level and probably it would create some gap in Kiali (i.e. you see 4 workloads but only get collected telemetry for 3)

daur88:
Hi, @lucasponce . I fully agree, no need to deal with augment labeling filtering in my case.

lucasponce:
Thanks for confirming @daur88.
Feel free th close the PR in that case to note that is not active.

And again, thank you for your contribution, it was a nice discussion and we appreciate your effort to catch up with the code and proposes a change.

Note, we have a lot of open issues in our backlog and we will be more than happy to help you if you'd like to collaborate on any of them.

PR #3663, opened 2021-02-01
use UBI8 for all builds

jmazzitelli:
This may affect people running a cluster on older node hardware where UBI8 is not officially supported.

For example, I think this may effect those with nodes running RHEL7 - it should still work, but if problems crop up that are caused by the underlying node operating system, the only solution will be to upgrade the node environment.

From what I was told a while ago (and why we did not move amd64 builds to UBI8 originally when we added ARM64/UBI8 support):

> We believe that most UBI8 images would work on a RHEL7 host. However
> those images are not officially supported by RHEL. So if there is a
> problem the responsibility falls on you, the container owner, to work it out.


jshaughn (approved):
I understand that this opens us up to potential RHEL7 compatibility issues but the expectation is that it works, and ubi8 is not fairly mature.   We should run QE on a RHEL7 env, if possible.  Because this will protect us from CVEs, as well as the fact that I think SM builds on ubi8, I say we make the change.

jmazzitelli:
@lucasponce we don't need this for 1.29 - is there a reason you put the backport label on this?

jshaughn:
Removing backport-v1.29 label, users can update to 1.30+ if they feel the need.

lucasponce:
> @lucasponce we don't need this for 1.29 - is there a reason you put the backport label on this?

I understand (correct me if I'm wrong) that this PR was to address issues in the base image used.

If the image used in Kiali 1.29 is vulvernable, then it should be backported, if not, then it won't be necessary.

Issue #3643, opened 2021-01-28
kiali ui failing to allow login after a session is externally deleted


jmazzitelli:
I got into a state where I can no longer log into the Kiali UI after multiple attempts. The only way for me to be able to log in again is to go into my browser settings and delete the kiali cookie.

This is using the `openshift` auth strategy.

How to replicate:

1. Log into Kiali UI normally
2. Delete the oauthaccesstoken.oauth.openshift.io via `oc` command
3. Notice your Kiali UI session immediately brings you to the login screen (it sees the session has been terminated).
4. Now try to log in - you can't. I tried closing my browser tab, I tried F5 refresh, nothing works. I cannot log in anymore. You have to delete the kiali cookie manually to be able to log in again.

This is the screen I see:

![image](https://user-images.githubusercontent.com/2029470/106160949-62f06180-6154-11eb-8d4d-bba9845076b9.png)

This is the log messages I see:

```
2021-01-28T15:37:04Z ERR Failed to get OK status from api endpoint [apis/user.openshift.io/v1/users/~] for oauth consumption, error: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}

2021-01-28T15:37:04Z WRN Token error: could not get user info from Openshift: Failed to get OK status from api endpoint [apis/user.openshift.io/v1/users/~] for oauth consumption, error: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}

2021-01-28T15:37:04Z ERR Failed to get OK status from api endpoint [apis/user.openshift.io/v1/users/~] for oauth consumption, error: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}

2021-01-28T15:37:04Z WRN Token error: could not get user info from Openshift: Failed to get OK status from api endpoint [apis/user.openshift.io/v1/users/~] for oauth consumption, error: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"Unauthorized","reason":"Unauthorized","code":401}

[...]

PR #3666, opened 2021-02-02
whenever the server detects the client is unauthorized, delete the token cookies

jmazzitelli:
delete the token cookies so the client never uses them again

fixes: https://github.com/kiali/kiali/issues/3643

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -152,6 +152,7 @@ func performOpenshiftAuthentication(w http.ResponseWriter, r *http.Request) bool
 
 	user, err := business.OpenshiftOAuth.GetUserInfo(token)
 	if err != nil {
+		deleteTokenCookies(w, r)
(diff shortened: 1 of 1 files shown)


jmazzitelli:
I believe I got it now. This is working in my tests. See the replication procedures in the issue https://github.com/kiali/kiali/issues/3643

jshaughn (approved):
This makes sense to me.  In the past we cleared the cookies only on logout, but we should also be clearing the [bad/stale] cookies whenever there is an auth problem with the token.

I see a couple of `RespondWithError(w, http.StatusUnauthorized, err.Error())` instances without additional messages, is that OK?

jmazzitelli:
> I see a couple of `RespondWithError(w, http.StatusUnauthorized, err.Error())` instances without additional messages, is that OK?

I didn't touch that stuff - I assume its all OK.

jmazzitelli:
The openshift-auth and token-auth molecule tests pass. I think this is good to go for merge.

Issue #3660, opened 2021-02-01
Performance problems with custom dashboards discovery


lucasponce:
Querying workload details [1] shows a significant amount of queries between 1.26 and v1.29 in a highly populated cluster.
 
Info is provided by @primeroz (thanks a lot!)

![image](https://user-images.githubusercontent.com/1662329/106493653-b860a180-64b9-11eb-9ced-598b4dc749e2.png)

So, this would deserve an investigation about what could be going on.

[1] http://localhost:20001/kiali/console/namespaces/twodotoh/workloads/gateway-external-write?tab=info&duration=600&refresh=0

primeroz:
For reference, this is a relatively big cluster 

* 80 nodes
* 250 pods within the mesh, mostly in the same namespace as the workload page that fails to load 

when loading the workload page , the workload has about 30 running pods , kiali will often OOM jumping from a normal usage of few hundreds MB to its limits of 6GB.

Sometimes , especially on 1.29, it also caused prometheus to OOM raising its ram usage from an average of 14GB to 24GB (its limit)

jotak:
hey @primeroz @lucasponce ,
I've found a quite serious bug in our prom client code but I have no reason to think it can cause your issue. Still ... might be interesting to test in your environment!
I've also added some trace-level logs.
All is there: https://github.com/kiali/kiali/pull/3664

Checking the new logs at a small scale, I don't find anything unexpected. When loading a Workload Details page just triggers 5 prometheus queries, all of them are expected (although it's certainly possible to optimize, there's some redundancy, but nothing huge as we saw in your case):

```
2021-02-02T10:32:30Z INF getRequestRatesForLabel: rate(istio_requests_total{destination_workload_namespace="default",destination_workload="ball-base"}[60s]) > 0
2021-02-02T10:32:30Z INF getRequestRatesForLabel: rate(istio_requests_total{source_workload_namespace="default",source_workload="ball-base"}[60s]) > 0
2021-02-02T10:32:30Z INF getRequestRatesForLabel: rate(istio_requests_total{destination_service_namespace="default",source_workload_namespace!="default"}[60s]) > 0
2021-02-02T10:32:30Z INF getRequestRatesForLabel: rate(istio_requests_total{destination_service_namespace="default"}[60s]) > 0
2021-02-02T10:32:30Z INF getRequestRatesForLabel: rate(istio_requests_total{source_workload_namespace="default"}[60s]) > 0
```

[...]

PR #3668, opened 2021-02-03
Performance / custom dashboards: new configs

jotak:
- discovery_enabled (true/false/auto) to switch discovery mode
- discovery_auto_threshold: pods threshold above which discovery is
  skipped in auto mode

Part of https://github.com/kiali/kiali/issues/3660

Operator PR: https://github.com/kiali/kiali-operator/pull/243

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -569,14 +569,19 @@ func (in *DashboardsService) GetCustomDashboardRefs(namespace, app, version stri
 
 	if len(runtimes) == 0 {
 		cfg := config.Get()
-		filters := make(map[string]string)
-		if app != "" {
-			filters[cfg.IstioLabels.AppLabelName] = app
+		discoveryEnabled := cfg.ExternalServices.CustomDashboards.DiscoveryEnabled
+		if discoveryEnabled == config.DashboardsDiscoveryEnabled ||
+			(discoveryEnabled == config.DashboardsDiscoveryAuto &&
+				len(pods) <= cfg.ExternalServices.CustomDashboards.DiscoveryAutoThreshold) {
+			filters := make(map[string]string)
+			if app != "" {
+				filters[cfg.IstioLabels.AppLabelName] = app
+			}
+			if version != "" {
+				filters[cfg.IstioLabels.VersionLabelName] = version
+			}
+			runtimes = in.discoverDashboards(namespace, filters)
 		}
-		if version != "" {
-			filters[cfg.IstioLabels.VersionLabelName] = version
-		}
-		runtimes = in.discoverDashboards(namespace, filters)
 	}
 	return runtimes
 }
--- a/config/config.go
+++ b/config/config.go
@@ -59,6 +59,12 @@ const (
 	OidcClientSecretFile        = "/kiali-secret/oidc-secret"
 )
 
+const (
+	DashboardsDiscoveryEnabled  = "true"
+	DashboardsDiscoveryDisabled = "false"
+	DashboardsDiscoveryAuto     = "auto"
+)
+
 // Global configuration for the application.
 var configuration Config
 var rwMutex sync.RWMutex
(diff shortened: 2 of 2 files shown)


jotak:
cc @primeroz
If you want to test, I pushed this image: `quay.io/jotak/kiali:dev`

jotak:
Testers: a simple way to test is to have some runtime metrics demo and scale up pods. Example:

- Run mesh-arena with metrics
```bash
kubectl label namespace default istio-injection=enabled
kubectl apply -f <(curl -L https://raw.githubusercontent.com/jotak/demo-mesh-arena/zizou/quickstart-metrics.yml) -n default
```

- Configure new pods threshold to something lower, e.g. `discovery_auto_threshold: 5`
- Check Kiali, application "ai" => runtime dashboards are visible
- Scale `kubectl scale deployment ai-visitors --replicas=5`
- Check Kiali, application "ai" => runtime dashboards are now hidden

primeroz:
@jotak i ll test this afternoon when i can get the change merged. 

So, to understand this correctly, with this change i should get custom dashboards in istio-system namespace for kiali ( since there are only a few pods there ) but not in my busy namespaces. right ?

thanks for the quick turn around

jotak:
@primeroz yes that's the idea, though it's not per namespace decision, but per app/workload

hhovsepy (approved):
Verified:
1. discovery_enabled=true it discovers custom dashboards in any case (discovery_auto_threshold is ignored)
2. discovery_enabled=false it does not discover custom dashboards (discovery_auto_threshold is ignored)
3. discovery_enabled=auto custom dashboards are discovered when pods count are less than discovery_auto_threshold value

aljesusg (approved):
LGFM

primeroz:
@jotak Is working for me as well sort of but there are still a couple of issues: 

>  yes that's the idea, though it's not per namespace decision, but per app/workload

I don't know exactly how this bit works but i have a situation like this: 

* App is composed of many workloads , 8, with different number of pods in each 
```
 app-workload1: 20 / 20
 app-workload2: 20 / 20
 app-workload3: 30 / 30
 app-workload4: 20 / 20
 app-workload5: 10 / 10
 app-workload6: 10 / 10
 app-workload7: 1 / 1
 app-workload8: 1 / 1
```
* When loading the `Application` page the change in this PR works and the autodiscovery is disabled 
* When i open the `Workload page` for any of the workload > 10 pods the PR works and the autodiscovery is disabled and the page loads up super quick 
* When i open the `Workload page` for any of the workload with <= 10 pods though the autodiscovery is **enabled** but it must still be doing the `series` query that returns a huge cardinality **causign Kiali to OOM like before** 

Also for some reasons, probably unrelated to your change but maybe part of 1.30 branch, in the workload page in the `graph overivew` panel i now get 
```
No namespace is selected
There is currently no namespace selected, please select one using the Namespace selector.
``` 

even though the path is correct and it includes the namespace

jotak:
>     * When i open the `Workload page` for any of the workload with <= 10 pods though the autodiscovery is **enabled** but it must still be doing the `series` query that returns a huge cardinality **causign Kiali to OOM like before**
>       
>       * something like using the whole `app` rather than the `workload` ?

It's expected that, when autodiscovery is run, it continues to run that "Series" query.
About the OOM, I think it really loads the metrics for a single workload (not app) but that's still too high and it still OOM. Pods is not the only thing that impacts cardinality, actually it's the whole mesh topology (number of interconnections), plus the variety of protocols used, variety of error codes and so on.

I think the quickest thing to do for you is just to turn off the discovery (you'll still be able to have dashboards from annotations).

But there's also an alternative, because even if with this simple flag turned off you get decent performances, you may still run into troubles later at higher scale. What's recommended when you hit the metrics cardinality problem, is to change your prometheus setup to decrease the retention time on istio's prometheus, have that prometheus strictly limited to scraping istio metrics (no more), and have a second prometheus instance that can scrape more targets and also rewrites istio metrics by discarding a couple of data (essentially pod information). The two prometheus are connected through federation. This is described more in details here: https://istio.io/latest/docs/ops/best-practices/observability/#using-prometheus-for-production-scale-monitoring
And how it impacts Kiali is described here: https://medium.com/kialiproject/kiali-with-production-scale-prometheus-c53ddfa20570, long story short, you can point in Kiali config to the istio's prometheus as the main URL, and to the second prometheus as the URL for custom dashboards.


>       * In that case than should the autodiscovery consider the whole `app` when deciding if is above the triggering threshold ?
> 
> 
> Also for some reasons, probably unrelated to your change but maybe part of 1.30 branch, in the workload page in the `graph overivew` panel i now get
> 
> ```
> No namespace is selected
> There is currently no namespace selected, please select one using the Namespace selector.
> ```
> 
> even though the path is correct and it includes the namespace

My bad, I messed up this while building the image, don't worry it's not part of this pull request (it's what I did yesterday to disable the mini-graph when we weren't sure yet what was causing your issue)

primeroz:
> I think the quickest thing to do for you is just to turn off the discovery (you'll still be able to have dashboards from annotations).

perfect, that is good enough for me. I will wait for this feature to release and will set _just the autodiscovery off_ while keeping the _custom dashboards on_ 

thanks!

jotak:
Thanks for tests and reviews, I'll merge and backport to 1.29

PR #3678, opened 2021-02-05
Prevent Proxy Status endpoint to hang + Add unreachable istiod into Istio Component Statuses

xeviknal:
This PR attempts to add prevention fixes to the forever taking request when the kube-api doesn't have access to istiod pods.
Fixes https://github.com/kiali/kiali/issues/3494

Three additions:
1. Adding timeouts to Pod Proxy calls.
2. Early return when Kiali can't check one healthy Istiod liveliness probe endpoint.
3. The Istio Component Status (masthead light) incorporates the liveliness probe result for each istiod pod (only shows the non reachable istiods)

To reproduce the scenario:
```
watch -n 3 kubectl delete pod -n istio-system --selector=app=istiod
```

Screenshot of Kiali (4)

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -69,12 +69,22 @@ func (iss *IstioStatusService) GetStatus() (IstioComponentStatus, error) {
 
 func (iss *IstioStatusService) getIstioComponentStatus() (IstioComponentStatus, error) {
 	// Fetching workloads from component namespaces
-	ds, error := iss.getComponentNamespacesWorkloads()
-	if error != nil {
-		return IstioComponentStatus{}, error
+	ds, err := iss.getComponentNamespacesWorkloads()
+	if err != nil {
+		return IstioComponentStatus{}, err
 	}
 
-	return iss.getStatusOf(ds)
+	deploymentStatus, err := iss.getStatusOf(ds)
+	if err != nil {
+		return IstioComponentStatus{}, err
+	}
+
+	istiodStatus, err := iss.getIstiodReachingCheck()
+	if err != nil {
+		return IstioComponentStatus{}, err
+	}
+
+	return deploymentStatus.merge(istiodStatus), nil
 }
 
 func (iss *IstioStatusService) getComponentNamespacesWorkloads() ([]apps_v1.Deployment, error) {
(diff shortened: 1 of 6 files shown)


business/istio_status.go
```diff
@@ -251,6 +261,30 @@ func (iss *IstioStatusService) getAddonComponentStatus() IstioComponentStatus {
...
+	cfg := config.Get()
+	ics := IstioComponentStatus{}
+
+	istiods, err := iss.k8s.GetPods(cfg.IstioNamespace, labels.Set(map[string]string{"app": "istiod"}).String())
+	if err != nil {
+		return ics, err
+	}
+
+	for _, istiod := range istiods {
+		_, err := iss.k8s.GetPodProxy(istiod.Namespace, istiod.Name, "/ready")
```
jotak, line 274: Just asking: wouldn't it be equivalent to get just the pod status? (ready/notready/ ...) hence relying on k8s API rather than crafting a custom query? Or this is not equivalent?

jshaughn (dismissed):
Looks OK to me.  No obvious problems when running.  One small comment.  I'll approve but defer to other approvals as well.

business/istio_status.go
xeviknal: Here is the main point. If we only rely on the k8s API we can't detect whether the cluster allow traffic between k8s API server and istiods.

We need this because when we fetch the proxy status, we connect to each istiod pod to fetch its `/debug/syncz` endpoint. This connection is done via k8s API: `kiali -> k8s API (proxy)-> istiod`.

The issue with private clusters usually has a really restricted K8S namespace access. Therefore, this check allows us to know whether the permissions are in place or not.

business/istio_status.go
```diff
@@ -251,6 +261,30 @@ func (iss *IstioStatusService) getAddonComponentStatus() IstioComponentStatus {
...
+			ics.merge([]ComponentStatus{
+				{
+					Name:   istiod.Name,
+					Status: Unreachable,
+					IsCore: true,
+				},
+			})
+		}
+	}
+	return ics, nil
```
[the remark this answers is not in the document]
xeviknal: Yes, it is on purpose. Thanks!
The istio component status (masthead) only shows components with trouble. Hence, when all is green, nothing is appended (for traffic reduction purposes).

jotak (dismissed):
LGTM (not tested)

business/istio_status.go
jshaughn: A simple one-line comment in the code is useful when there is a design decision that may not be clear to the next dev.  You put thought into the approach, share that thought in the code when it may not be obvious.

jshaughn (approved):
LGTM

hhovsepy:
When 'istio-egressgateway' or 'istiod' pods are failing it shows warn or error as expected:
Screenshot from 2021-02-08 09-36-05

But when these pods are not deployed at all, no any error or warn is shown in Kiali:
Screenshot from 2021-02-08 10-04-06
Screenshot from 2021-02-08 10-04-36

xeviknal:
How did you manage to have 0 pods? Through the deployment replicas field?

If so, I like the idea to cover this scenario although it is something that users have explicitly created. Kiali being transparent of what happens at istio component deployment level.
As this is a new scenario not covered before, what do you think of adding an issue for that and I cover this in a subsequent PR?

hhovsepy (approved):
@xeviknal agreed, approving this PR.

kubernetes/kubernetes.go
```diff
@@ -289,6 +290,17 @@ func (in *K8SClient) GetPodLogs(namespace, name string, opts *core_v1.PodLogOpti
 	return &PodLogs{Logs: buf.String()}, nil
 }
 
+func (in *K8SClient) GetPodProxy(namespace, name, path string) ([]byte, error) {
+	return in.k8s.CoreV1().RESTClient().Get().
+		Timeout(10 * time.Second).
```
lucasponce, line 295: Minor, I remember there were other timeouts for authentication endpoints IIRC perhaps worth to put the value in a constant and exposed to configuration, but not a blocker, just a side comment to future iterations.

lucasponce (dismissed):

kubernetes/kubernetes.go
xeviknal: Good! I've used the `DefaultTimeout` value already defined for the http utils.

lucasponce (approved):

Issue #2887, opened 2020-06-10
Show network latencies in graph


pavolloffay:
**Is your feature request related to a problem? Please describe.**

I would like to see network latencies between workloads/services in the graph.

**Describe the solution you'd like**

This is related to https://github.com/kiali/kiali/issues/2677

Calculate network latencies (e.g. client span start time - server span start time) from traces and show the durations in the graph. It should be possible to navigate from the duration to trace instance. The system could also highlight durations that are higher as the other durations for the same pair of pods (client - server).

If I am not mistaken this metric information can be only extracted from traces (e.g. client span timestamp - server span timestamp).

**Describe alternatives you've considered**

I am assuming that Envoy does not expose client-side metrics partitioned by the destination pod. Even if it would provide such information it would be probably collected as a histogram and with no direct link to a trace.

**Additional context**
related to https://github.com/kiali/kiali/issues/2677

jshaughn:
@jotak Do you think this is in scope for Kiali?

jotak:
If I understand correctly the suggestion, the goal is to show pure network latency (ie. without server processing time) in graph.
I think we can reformulate the initial request, the fact that it comes from traces or not is an implementation detail.

We could have that with metrics (hence at a high aggregation level), with basically: `duration{reporter=source} - duration{reporter=destination}` => that should in theory give the pure network latency. It could be a new graph edge option?

A similar operation can be done with traces, so when a trace is selected we could add in the trace details (or displayed on edges?) the network latency, as Pavol says, with `server.startTime - client.startTime`. That could be a second subtask.

lucasponce:
> Indeed we could have that with metrics (hence at a high aggregation level, which matches better the scope of whole graph), with basically: duration{reporter=source} - duration{reporter=destination} => that should in theory give the pure network latency. It could be a new graph edge label option?

I was about to write a similar comment, the differences between reporters can add value in a similar was as @jotak described.

Probably it would require a little bit of more experimentation, but this is also linked with other discussions about reporters we are having.

I like to see it as an opportunity to show value to the user.

lucasponce:
I'm revisiting this one.
I'd suggest to change the scope but in the same direction to help user to troubleshoot better differences between client/servers.

Perhaps, in the graph, we have a good coverage with the recent changes:
![image](https://user-images.githubusercontent.com/1662329/128339911-446d74a2-696c-4647-a0dd-e88480cfb751.png)

[...]

PR #3680, opened 2021-02-05
Update graph query approach to use complementary source and dest reporting

### LINKED ISSUE #3673: Better handle single requests that generate multiple destination telemetry entries
This is a follow-up to https://github.com/kiali/kiali/issues/3642 and also [this Istio Slack thread](https://istio.slack.com/archives/C382V8Q92/p1612194626015700).

That thread basically confirms that it is possible, and expected, that a single request could generate multiple destination telemetry entries when retry is applied:

"...assuming downstream and upstream retry policy are the same (let’s assume 2), then downstream proxy will record only 1 failure, while upstream proxy will record 2. Since every retry from downstream proxy is considered as a new request at upstream proxy. All of the records should have URX attached."

"...From upstream proxy point of view, there are indeed two requests coming in, which results in failure..."

Kiali typically assumes that for a single request from `A-downstreamProxy --> upstreamProxy-B` that there will always be 1 entry added in source telemetry by the downstream proxy, and 0 or 1 entry in destination telemetry by the upstream proxy.  0 if the request never reaches the downstream workload, and 1 if it does.

This can result in confusing, or seemingly conflicting, error rates when looking at a destination node using source telemetry.  Using the example above, traffic from A -> B may show 1/2 as many errors using source telemetry as it would using dest telemetry, when retry is involved.

The problem, I think, is mainly due to the graph code using mostly source telemetry to generate the traffic topology.  It does this so as to not miss telemetry generated only by the source/upstream proxy, in cases where the request never reaches a destination workload.

It may be time to revisit the graph query approach such that it can capture the complementary source telemetry and destination telemetry, while avoiding duplicates.

jshaughn:
Fixes #3673 

See the Issue for the motivation and details of the new approach.  In short, use source telem only for unserviced requests (i.e. those that did not reach a destination workload for the requested service), and dest telem for everything else.  This ensures we capture all telemetry when a single source request can result in multiple dest requests, due to things like retry at the source proxy.  The end result are error rates in the graph that are more consistent with those generated by health, or metric charts.

UPDATE 2/20/21:
new query approach as the original ended up not being able to handle various corner cases.  In short the new strategy queries for incoming and outgoing rates in separate queries, and although the queries may fetch overalapping time series, the data is not duplicated as the results are assigned to only the source or dest (with some protection for not duplicating edge rates).  Removing the DNM label as testing looks good so far and the cases where the original approach failed, are now passing.


jshaughn:
@lucasponce Please give this a try and let me know whether you see more consistent results with the "rate gap" issue.
@israel-hdez Please give this a try in your MC setup and ensure things are handled as well as before.
@mattmahoneyrh Please regression test this, the general result should be graphs that likely look the same.  If you see any discrepancies we can look to see if they make sense.  It is actually possible that the new queries pick up some valid "idle" traffic that was missed in the past.

mattmahoneyrh:
Encountered error in Kiali Message Center:
 Could not fetch health: error in metric request_count: server_error: server error: 503

lucasponce:
I've just merged the pending work for health config that may side impact to this.
Would you mind to rebase with master ?

israel-hdez:
@jshaughn Testing in the MC setup with Istio 1.9. I see the same graph in "clusterA (kukulcan)":

image

In "clusterB (tzotz)", there is a difference. This is how it looks:

image

I think the new graph is much better for clusterB, because we now have insight about where traffic is coming from. This is awesome.

I'll check the code tomorrow, if I can.

lucasponce:
From feature perspective I don't see differences in the scenrios I'm testing.

I mean, we have a "v-server" talking with a "w-server" and from the "v-server" perspective the error rate is around a 15%:
image

But from the "w-server" perspective the error rate is higher, which it maches with the reality:
image

And, in the graph we still having some data hard to explain to the user, or at least, user requires deep understanding about what is happening:
image

I see in the sidepanel, that both data are now flipped but different, and health info used is showing the perspective of "v-server".

So, the health tooltip matches with the bottom graph summary, but not with the first one.

I think we need to iterate on this, now we understand why this is happening because we have investigated, but presenting the info in that way is confusing and probably most of users will feel that Kiali has a problem instead that telemetry reported from source/destination is different on purpose.

jshaughn:
> I think we need to iterate on this, now we understand why this is happening because we have investigated, but presenting the info in that way is confusing and probably most of users will feel that Kiali has a problem instead that telemetry reported from source/destination is different on purpose.

@lucasponce I agree that we we need to get these numbers aligned or it will confuse the users.  The issue you point out, I think, is due to the fact that the graph aggregates the outgoing edge rates, and those rates now combine source and destination reporting.  But metrics/health, I think, is using solely source reporting for outgoing.

Let's talk about it, although I think this new approach still has benefits over the prior query approach.

jshaughn:
@lucasponce @israel-hdez @pbajjuri20 I am seeing a potential issue but I'm not sure if it's our problem, or an issue in upstream Istio/Envoy telemetry.  I'm asking about it in Slack.   The problem is that upstream/destination telemetry is not providing the URX flag and so the 503 failures are not showing that the issue is a retry or connection limit problem.

israel-hdez:
BTW, I was looking closer at the graph in the MC setup. There IS a change in the "clusterA" graph. This is how it looks with these changes:

image

This is how it looked before:

image

Notice that reviews-v2 and reviews-v3 are missing.
Sorry for noticing this late. I didn't look carefully.

jshaughn:
> Notice that reviews-v2 and reviews-v3 are missing.  Sorry for noticing this late. I didn't look carefully.

OK, that's a problem for sure.  So what we've managed to do is lose outgoing from the source cluster and gain incoming from the dest cluster.  This is for sure due to the source telemetry generated for inter-cluster traffic.  Let's take a look at that telemetry.  I'll flag this as Do Not Merge until we figure out what to do about this issue.

jshaughn:
@israel-hdez Another update to the PR, please retry again in the MC env.  Gracias!

jshaughn:
@israel-hdez I think this is now ready for any final review, thanks for all of your help with the multi-cluster testing thus far.

lucasponce:
Last time I checked this I also get some weird results.
I don't have bandwidth but I cc @hhovsepy as I'd like to know his opinion before this PR is merged.

hhovsepy:
The inconsistency mentioned by @lponce I can see on master as well. The issue mentioned by Edgar is fixed now.

jshaughn:
@lucasponce, since the last time you looked at this PR it has basically been re-written.  It now addressed several issues present in the first approach.  But the inconsistency you and @hhovsepy are observing  is expected, and out of scope for this PR, as explained in [this document](https://docs.google.com/document/d/1k0TKtaLFxbSNX5r6WcIh5XJoinzg0--QlLfTK_djNbw/edit?usp=sharing).   From a testing perspective the new approach is better in thatit captures more multi-cluster traffic, more destination telemetry in certain corner cases (like retry), and more telemetry in one-sided reporting (telem reported only from either source or destination).

jshaughn:
@pbajjuri20 or @mattmahoneyrh this PR is now valid for QE regression testing.

spockz:
I can confirm this worked for us with the docker image provided by @jshaughn.

pbajjuri20 (approved):

graph/telemetry/istio/appender/response_time.go
```diff
@@ -27,6 +27,9 @@ var (
 // is represented as a percentile value. The default is 95th percentile, which means that
 // 95% of requests executed in no more than the resulting milliseconds. ResponeTime values are
 // reported in milliseconds.
+// Response Times are reported using destination proxy telemetry, when available, which should remove
+// network latency fluctuations.
+// TODO: Should we report both source and destination when possible?
```
israel-hdez, line 32: As we could fetch both response time from source and also response time from destination, rather than reporting both, perhaps we could report something similar to "Jitter"?

See: https://en.wikipedia.org/wiki/Jitter#Packet_jitter_in_computer_networks

Possibly related to: #2887

graph/telemetry/istio/istio.go
```diff
@@ -70,91 +70,67 @@ func BuildNamespacesTrafficMap(o graph.TelemetryOptions, client *prometheus.Clie
...
+	idleCondition := "> 0"
+	if o.IncludeIdleEdges {
+		idleCondition = ""
+	}
 
-	// 3) query for internal traffic, originating from a workload inside of the namespace
-	query = fmt.Sprintf(`sum(rate(%s{reporter="source",source_workload_namespace="%s"} [%vs])) by (%s)`,
-		requestsMetric,
+	// 1) Incoming: query destination telemetry to capture namespace services' incoming traffic
+	query := fmt.Sprintf(`sum(rate(%s{reporter="destination",destination_service_namespace="%s"} [%vs])) by (%s) %s`,
```
israel-hdez, line 84: I see in some queries you are preferring `destination_service_namespace` rather than `destination_workload_namespace`. I'm wondering if this would introduce a regression: did you try the direct workload-to-workload case? I remember the graph was crashing in this case. I'm not sure if this is covered by tests.

I guess this workload-to-workload case could be catched by the 2nd query. But I think it's only partially covered, as I think a workload in "another" namespace which is calling directly a workload in "this" namespace won't be visible in the graph.

IMO, it's a corner case. But I'm not sure if this case is that common in the field so that we need to worry about it. So, it's up to you to check it.

israel-hdez (approved):
I'm good in general.
But read my comment about that corner case. I hope it's works as expected ;)

graph/meta.go
```diff
@@ -25,7 +25,6 @@ const (
 	IsEgressCluster MetadataKey = "isEgressCluster" // PassthroughCluster or BlackHoleCluster
 	IsIdle          MetadataKey = "isIdle"
 	IsInaccessible  MetadataKey = "isInaccessible"
-	IsMisconfigured MetadataKey = "isMisconfigured"
```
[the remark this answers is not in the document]
jshaughn: @israel-hdez This was some deprecated code that I cleaned up to simplify the code actually changing in the PR.  It had to do with detecting issues during the time when the "canonical" attributes may or may not exist.  Given that they now are guaranteed the code was not useful.

graph/telemetry/istio/appender/response_time.go
jshaughn: @israel-hdez It's something to look into and definitely relevant to #2887.  It's not directly relevant to this PR.  I think the first thing to verify is the assumption that source proxies do in fact include latency and dest proxies do not.  I have been assuming yes, but it's something to make sure of before we do any work.

graph/telemetry/istio/istio.go
jshaughn: @israel-hdez   Well, it's a good question.  I think this may be a holdover from previous revisions of the queries.  This is destination telemetry and so it should be true that destination_workload would be set, and so it should be fine, and potentially better, to use destination_workload_namespace.

I'm fairly sure there would be no negatives to making the change, and maybe there is an advantage for the scenario you mention.  Having said that, I'm not sure it is ever true that destination_service_namespace != destination_workload_namespace when destination_workload is set.  So, I'm OK  to make the change.

The more troubling thing is something related to your scenario (a workload in another namespace which is calling directly a workload in 'this' namespace).  When thinking about it I realized that the PR currently will not capture traffic from outside the namespace that fails to reach its intended workload inside the namepace.  Put another way (using bookinfo as the target namespace):

reporter="source",
source_workload_namespace!="bookinfo", 
destination_service_namespace="bookinfo",      // actually better: destination_service=~"^.*\.bookinfo\..*$"
destination_workload_namespace="unknown", 
destination_workload="unknown"

The question is whether to add this query to the current approach or to just let users add the source namespace to the selected namespaces.

Using bookinfo I cycled down productpage to 0 pods to test this scenario.  To my surprise master also does not show the failed traffic because istio-ingressgateway does not set destination_service_namespace, only destination_service and destination_service_name.  That is why if we were to add a query we should use destination_service.   In master or in the PR I can see it fine by adding istio-system to the selected namespaces.

Any thoughts as to whether I should add the third query, another RT to Prom, for this situation?

jshaughn: After thinking more on this overnight, as much as I'd prefer not to have to add a 3rd query I think it probably makes sense to add the additional query and get a complete view of incoming traffic.  In general the graph is supposed to show incoming/outgoing edge traffic from/to unselected namespacs.  And this addition, in particular, would likely show failing requests, which is valuable.

graph/telemetry/istio/appender/response_time.go
israel-hdez: I fully about verifying that source proxies include latency.

graph/telemetry/istio/istio.go
israel-hdez: > Having said that, I'm not sure it is ever true that destination_service_namespace != destination_workload_namespace when destination_workload is set. So, I'm OK to make the change.

I don't think services can point to workloads on a different namespace. So, I think you are right. However, I found that a service can refer to another service in another namespace. See https://stackoverflow.com/a/44329470 (it's a use-case [mentioned in official docs](https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors)) . I've never tried this. Not sure what Istio would do :-/ Anyway, we haven't heard a case. So, let's not worry about that.

> as much as I'd prefer not to have to add a 3rd query I think it probably makes sense to add the additional query and get a complete view of incoming traffic.

I agree. I was also thinking about it a little in after-work hours, and I think it's worth to add the 3rd query. I agree that the failing requests are important.

jshaughn: @israel-hdez I have made another commit with both of the changes we discusses.  First, a change from destination_service_namespace to destination_workload_namespace for the destination reported telemetry.  Second, I added the new query to capture source-only telemetry for request originating outside the namespace, and failing to be serviced by a workload.  I have again removed the DNM label.   Please take a look and re-approve if it seems OK.  @pbajjuri20 the new commit will also force you to re-approve, thanks.

lucasponce:
Before to approve this one, please test this step from the tutorial:
https://kiali.io/documentation/latest/tutorial/#03-open-application-to-outside

In the current 1.29.x the Graph is unable to draw edges for workloads if destination ones have no sidecar included, but that traffic exists and that was working in 1.28.x version.

I'll test closely next week, but I'd like if we are in the same page of expectations about what should be draw in such scenario.

jshaughn:
> Before to approve this one, please test this step from the tutorial:
https://kiali.io/documentation/latest/tutorial/#03-open-application-to-outside

@lucasponce I will test this scenario.  And we will not merge until you have had a chance to try things next week.  At the moment things are looking good, though, as the PR is catching more traffic than at any time in the past.

pbajjuri20 (dismissed):

lucasponce:
> @lucasponce I will test this scenario. And we will not merge until you have had a chance to try things next week. At the moment things are looking good, though, as the PR is catching more traffic than at any time in the past.

Thanks ! I'm catching up today with this one. I'll add a comment shortly.

lucasponce (changes requested):
I think there are a couple of problems:

image

Control workload has outbound traffic that should be visible here.

Also, when going into the details there is an error in the workload:

image

```
error in metric request_count: bad_data: 1:134: parse error: missing unit character in duration
```

So, I couldn't debug more, but these steps are from the tutorial, which should work.

The gap between health data and telemetry is flipped now in the charts of the sidepanel but as discussed we'd need to work on that in a separate effort.

jshaughn:
@lucasponce So, actually, the issue seems not in the query approach of the PR but rather a telemetry issue.  I'm just not sure if it's a bug or something we need to work around.  I'll investigate more tomorrow but the problem is that destination_cluster is not being set for the unproxied destination services and the time series are being filtered as "bad telemetry".  When I remove the filtering the graph looks as expected.  I feel like if Istio can route to the destination it should know which cluster it's on.

lucasponce (approved):
Thanks for the update @jshaughn.
As we are going to actively work on this area we can bring again this topic as well.

I'd say to just add a minor comment in the FAQ in case some user hits this one.

I tell you, it will happen, as the current strategy of istio is to not need a sidecar in destination if it's not necessary.

Telemetry addons is a clear example of a scenario that will be impacted by this.

lucasponce:
@jshaughn it works ok, as you commented this is my result:

image

And it cover the use case where a workload (control) is communicating with other workloads without sidecar:

image

One minor topic, perhaps the workloads in travel-portal on this case should have the "missind sidecar" badge, what do you think ?

jshaughn:
> perhaps the workloads in travel-portal on this case should have the "missind sidecar" badge, what do you think ?

good point, I wonder if somewhere along the line we regressed on that badging...  I'll take a look...

lucasponce:
@jshaughn I'm ok to merge and work in a followup task.

Just to be sure that the goal should be to get back at least the previous experience for the user:
image

I mean, there is outbound traffic from control to a number of workloads that don't have sidecar.

israel-hdez (dismissed):
I'm hoping this PR will catch this release :)

jshaughn:
@lucasponce Unfortunately I think I see another issue.  Putting DNM back in place, no need to look at this until I'm sure it's ok.

lucasponce:
@jshaughn ok, thanks for the update. Let me know if I can help.

lucasponce:
The scenario is back :)
image

lucasponce (dismissed):
The scenario that I commented in my previous review is fixed.
Thanks @jshaughn

jshaughn:
@israel-hdez @lucasponce @mattmahoneyrh, Although the coding was done before the 1.31.0 sprint release I thought that the PR was still a little "hot" for a last second merge.  I know Matt, Edgar and myself have all played with the latest bits, Lucas has not had a chance but validated several prior versions.  Anyway, for sanity I thought I'd wait until minimally, next Monday.  When everyone is OK with it we can merge and then I'd like to backport for a 1.31.1 version.

jshaughn:
One more change, this time to handle better the istio bug Istio#29373.  When the dest_cluster is errantly unset assume it is the same as the source_cluster.  This gives much better behavior in the typical case.

Before 
image

After
image

jshaughn:
@israel-hdez @lucasponce @mattmahoneyrh I squashed the PR into a single commit with a sane comment.  If/when you approve feel free to merge this, you don't need to wait for me.

mattmahoneyrh (approved):

lucasponce:
I've also tested the last commit with the original usecase that triggered the changes:
image

Also, some smoke testing with other examples:
image

So it's ok from my side as well.

lucasponce (approved):

lucasponce:
I was testing more things and I see a regression in the traffic animation.
It's not critical, but I'm going to check if it's present in master or just in this PR.

What I see:
- Set up traffic animation -> it works.
- Move to other page (i.e. Overview) and back to Graph -> animation seems broken.

israel-hdez (approved):
In this new revision, I'm seeing some new edges that weren't present in past revisions of the PR:

image

EDIT: (completeness) the new edges I'm seeing are the TCP ones.

I verified that the associated metrics are present in Prometheus. I see TCP queries were changed so, most likely, edges were missing with the original queries and perhaps the new queries are providing a more complete view. Notice that the TCP edges involve nodes that are the "boundaries" to the ´istio-eastwestgateway´ -- these TCP edges are visible only when selecting the ´istio-system´ namespace. So, I think Kiali was missing these data in past revisions.

Now, the question is why Isito is recording this telemetry that is, apparently, redundant because there is the HTTP telemetry. IMO, these TCP edges are more an Istio detail rather than something with this PR.

So, I'm approving, I'll ask in Istio channels if this TCP telemetry is expected. Everything else looks good.

israel-hdez:
I see others are OK with this PR. I'll merge in a moment...

Issue #3481, opened 2020-11-30
Enable Github Actions on Kiali repos


lucasponce:
As it happens with other OSS projects, Kiali has reached some limit of free usage of Travis.

So, we are in a moment that perhaps we need to review what to do:

- Upgrade the Travis plan (if doable).
- Migrate Travis to Github Actions (which it's not clear where the limit for free OSS usage are).

- [x] Migrate Travis CI to Github Actions https://github.com/kiali/kiali/issues/3502
- [ ] Enable Kiali Epic Action for better maintain Epic/Subtasks in Kiali Project https://github.com/kiali/kiali-epic-action

lucasponce:
Team has discussed that it's agree to start using actions for:
- CI/CD in our PR
- Project Management to enable https://github.com/kiali/kiali-epic-action to automate Epics/Subtasks.

lucasponce:
The important task, to move to Github Actions is completed.
Adding the kiali-epic-action is a nice to have that has not priority for now, it can enabled in some free / spare cycle.

PR #3700, opened 2021-02-12
Homogenate GH-Actions CI with old Travis CI

israel-hdez:
* Run linters to fail if there is coding with bad syntax.
* Check swagger file to avoid outdated API docs.
* Run tests with race detection.
* Fix linting issues
* Fix swagger file
* Restore snyk snapshot capture. We have been receiving security reports for an old snapshot from Nov 2020.
* Delete the .travis.yml file which is no longer needed.

Related kiali/kiali#3481


jshaughn (approved):
No expert but it seems fine.

PR #3705, opened 2021-02-15
Fix Snyk GitHub action

israel-hdez:
The official Snyk action didn't work. Using something that is more close to what was being done in Travis. This is better, because docker is not involved.

Related #3481

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -16,21 +16,20 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '12'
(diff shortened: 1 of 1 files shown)


israel-hdez:
@jshaughn Sorry, the actions provided by snyk didn't work.
I'm fixing with the approach used at travis. Here you can see a successful red run (it's red because there is a warn): https://github.com/israel-hdez/swscore/pull/20/checks

israel-hdez:
@jshaughn Can you take a look?
I would like to get rid of those failed CI checks in `master` before I'm the only saturner ;)

jshaughn (approved):
OK, this looks a good change, thanks.

israel-hdez:
Thanks! Merging...

PR #3726, opened 2021-02-22
Adding concurrency to istio components status

xeviknal:
While computing the istio component status, in particular the istiod reachability, kiali might introduce a sequential 10s waiting.
This PR approach this adding recurrence in each call to istiod. 

The scenario to reproduce this should be having 2+ istiod pods not responding. Then, Kiali should be waiting 10s for each istiod. After this PR, Kiali should only wait for 10s max (due to the concurrence).

In addition, there it fixes a bug that might also impact the computation of the `proxy status`. A bad usage of pointers might kiali make fetch envoy status to non-healthy istiods.

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	apps_v1 "k8s.io/api/apps/v1"
+	core_v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


hhovsepy (approved):
Verified by regression testing.

jotak (approved):
LGTM

Issue #3737, opened 2021-02-25
invalid parameter \"match[]\": 1:44: parse error: unexpected character inside braces: '.'"


larsrnielsen:
Trying to browse an application's page in Kiali I sometimes get the error below. The strange part is that I am not always getting this error. My platform is AWS EKS 1.19 and I have installed Kiali with 

```
istio_labels:
    app_label_name: "app.kubernetes.io/name"
    version_label_name: "app.kubernetes.io/version"
```

```
❯ helm search repo kiali-server
NAME              	CHART VERSION	APP VERSION	DESCRIPTION
kiali/kiali-server	1.30.0       	v1.30.0    	Kiali is an open source project for service mes...
```

```
{"level":"error","time":"2021-02-25T11:22:30Z","message":"runtimes discovery failed, cannot load metrics for labels: {kubernetes_namespace=\"epg\",app.kubernetes.io/name=\"epg-front\"}. Error was: bad_data: invalid parameter \"match[]\": 1:44: parse error: unexpected character inside braces: '.'"}
```

jotak:
Hi,
I think there's something wrong in code anyway, as this is kubernetes labels being used in place of prometheus labels. I'll check that.
Though I've no idea why it would be failing randomly and not consistently.

jotak:
So indeed, it was missing a call to labels sanitizer. I sent a pull request.
(app and version labels of custom metrics in prometheus should be "app_kubernetes_io_name" and "app_kubernetes_io_version" in your case)

PR #3739, opened 2021-02-25
Fix sanitizing label names

jotak:
Fixes https://github.com/kiali/kiali/issues/3737

Necessary when app/version labels are not the default ones


aljesusg (approved):
LGFM

lucasponce (approved):
LGTM
Tested locally with no regressions on graph+metrics logic.

PR #3750, opened 2021-03-05
Add missing token validation for implicit flow

israel-hdez:
The code handling OIDC authentication for the "implicit flow" is not
fully validating the received id_token *when* RBAC is turned off. Under
these cofiguration, only the CSRF mitigation check and the replay attack
mitigation is happening.

This is adding the missing lines of code to fully verify the id_token on
an "implicit flow + RBAC off" configuration.


jshaughn (approved):
LG!

Issue #3725, opened 2021-02-22
Istio gateway not found when deployed as DaemonSet


hitman99:
**General remarks**

We switched from kubernetes deployment to daemonset for istio gateway(s) and Kiali is not able to find them since.

**Describe the bug**
Kiali is not able to detect any traffic coming in our out of the mesh for any service.

In UI it shows "Not Found" for all Istio ingress gateways

**Versions used**
Kiali: v1.27.0
Istio: 1.6.5
Kubernetes flavour and version: (AWS EKS v1.16.15)

**To Reproduce**
1. Change istio ingress gateway from deployment to daemonset
2. See error

**Expected behavior**
Kiali should also try to retrieve statuses for istio components deployed as daemonsets (primarily the ingress gateways)

lucasponce:
@hitman99 take a look to this FAQ
https://kiali.io/documentation/latest/faq/#workload-is-not-found

But yes, probably Kiali is expecting to see it as a deployment rather than a daemonset for internal troubleshooting.

That case probably is not well covered.

cc @xeviknal

hitman99:
I've checked the code and it looks like for Istio Core Components Kiali only queries `[]apps_v1.Deployment`: https://github.com/kiali/kiali/blob/master/business/istio_status.go#L90

xeviknal:
yes, you are right. We only cover deployments. We can easily tackle this one. Thanks for opening the issue.

xeviknal:
hi @hitman99,

I am approaching this one. Could you please show me your way to install the ingress as a DaemonSet? I'd like to make sure we are covering your needs. Thanks.

PR #3764, opened 2021-03-10
Istio ingress may be installed as a daemonset

xeviknal:
fixes https://github.com/kiali/kiali/issues/3725
needs operator PR: https://github.com/kiali/kiali-operator/pull/262
needs helm-chart PR: https://github.com/kiali/helm-charts/pull/51

To reproduce this one:

1. Delete the istio ingress deployment
2. Apply the following daemonset
`k apply -f https://raw.githubusercontent.com/xeviknal/istio-lab/master/istio-ingressgateway-daemonset.yaml -n istio-system`

DNM - needs https://github.com/kiali/kiali-operator/pull/261 to be merged first.

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -5,7 +5,6 @@ import (
 	"sync"
 	"time"
 
-	apps_v1 "k8s.io/api/apps/v1"
 	core_v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
@@ -18,7 +17,8 @@ import (
 
 // SvcService deals with fetching istio/kubernetes services related content and convert to kiali model
 type IstioStatusService struct {
-	k8s kubernetes.ClientInterface
+	k8s           kubernetes.ClientInterface
+	businessLayer *Layer
 }
 
 type ComponentStatus struct {
@@ -71,12 +71,12 @@ func (iss *IstioStatusService) GetStatus() (IstioComponentStatus, error) {
 
 func (iss *IstioStatusService) getIstioComponentStatus() (IstioComponentStatus, error) {
 	// Fetching workloads from component namespaces
-	ds, err := iss.getComponentNamespacesWorkloads()
+	workloads, err := iss.getComponentNamespacesWorkloads()
 	if err != nil {
 		return IstioComponentStatus{}, err
 	}
 
-	deploymentStatus, err := iss.getStatusOf(ds)
+	deploymentStatus, err := iss.getStatusOf(workloads)
 	if err != nil {
 		return IstioComponentStatus{}, err
 	}
@@ -89,15 +89,15 @@ func (iss *IstioStatusService) getIstioComponentStatus() (IstioComponentStatus, 
 	return deploymentStatus.merge(istiodStatus), nil
 }
 
-func (iss *IstioStatusService) getComponentNamespacesWorkloads() ([]apps_v1.Deployment, error) {
+func (iss *IstioStatusService) getComponentNamespacesWorkloads() ([]*models.Workload, error) {
 	var wg sync.WaitGroup
 
 	nss := map[string]bool{}
-	deps := make([]apps_v1.Deployment, 0)
+	wls := make([]*models.Workload, 0)
 
 	comNs := getComponentNamespaces()
 
-	depsChan := make(chan []apps_v1.Deployment, len(comNs))
+	wlChan := make(chan []*models.Workload, len(comNs))
 	errChan := make(chan error, len(comNs))
 
 	for _, n := range comNs {
@@ -105,28 +105,20 @@ func (iss *IstioStatusService) getComponentNamespacesWorkloads() ([]apps_v1.Depl
 			wg.Add(1)
 			nss[n] = true
 
-			go func(n string, depsChan chan []apps_v1.Deployment, errChan chan error) {
+			go func(n string, wliChan chan []*models.Workload, errChan chan error) {
 				defer wg.Done()
-				var ds []apps_v1.Deployment
+				var wls models.Workloads
 				var err error
-				if IsNamespaceCached(n) {
-					ds, err = kialiCache.GetDeployments(n)
-				} else {
-					// Adding a warning to enable cache for fetching Istio Status.
-					// It should use cache, as it's an intensive operation but we won't fail otherwise
-					// If user doesn't have access to istio namespace AND it doesn't have enabled cache it won't get the Istio status
-					log.Warningf("Kiali has not [%s] namespace cached. It is required to fetch Istio Status correctly", n)
-					ds, err = iss.k8s.GetDeployments(n)
-				}
-				depsChan <- ds
+				wls, err = fetchWorkloads(iss.businessLayer, n, "")
+				wliChan <- wls
 				errChan <- err
-			}(n, depsChan, errChan)
+			}(n, wlChan, errChan)
 		}
 	}
 
 	wg.Wait()
 
-	close(depsChan)
+	close(wlChan)
 	close(errChan)
 	for err := range errChan {
 		if err != nil {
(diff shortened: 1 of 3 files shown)


lucasponce (changes requested):
I don't think is ready.
I'm going to look closer, but it fails when trying to get the workload details in the UI.

My first impression is that there is something missing I'll take a look and @xeviknal let me send a followup commit on top of your PR, perhaps that's easier for me.

lucasponce:
Ok, it seems it may require this https://github.com/kiali/kiali/pull/3771
@xeviknal would you mind to rebase when ready ?

business/istio_status.go
```diff
@@ -89,38 +95,73 @@ func (iss *IstioStatusService) getIstioComponentStatus() (IstioComponentStatus,
...
 
-	depsChan := make(chan []apps_v1.Deployment, len(comNs))
-	errChan := make(chan error, len(comNs))
+	depsChan := make(chan []workloadInfo, len(comNs)*2)
+	errChan := make(chan error, len(comNs)*2)
 
 	for _, n := range comNs {
 		if !nss[n] {
-			wg.Add(1)
+			wg.Add(2)
```
lucasponce, line 111: I'm thinking on this logic.
And the same problem we have here (to use a DaemonSet vs a Deployment) is going likely to happen with other workload type.

So I wonder if this logic should just pull "workloads" from the business layer; then, we will solve the scenario where the istio workload is from one type or another, and we won't duplicate a similar logic that is keep synced in the workload package.

Or perhaps 99% will be always a Deployment, and a 1% a cornercase like a DaemonSet, but if we have another similar personalization, we'll have same problem again (thinking on Job controller or a ReplicationController, or another controller different than Deployment).

I don't know, I'll keep studying these logic.

As it happens in other area, as the Graph Appenders, I think that when using workloads, they should use de business package instead of query directly.

lucasponce: Another comment around this, it happens also, that this doesn't update the DaemonSet controller view in the Workload level.

(DaemonSets and Jobs were "cornercases" for Istio when we develop this logic a.k.a. there were not even supported in Istio in first versions, so, for that case when a DaemonSet/Job was present we tried to resolve from Pods instead from the controllers. That has some known limitations, like non visibility of these controllers scaled to 0 for example, or the lack of fetching info from the controller itself).

business/istio_status.go
```diff
@@ -169,14 +208,14 @@ func istioCoreComponents() map[string]bool {
...
+func (iss *IstioStatusService) getStatusOf(wls []workloadInfo) (IstioComponentStatus, error) {
 	statusComponents := istioCoreComponents()
 	isc := IstioComponentStatus{}
 	cf := map[string]bool{}
 
 	// Map workloads there by app name
-	for _, d := range ds {
-		appLabel := labels.Set(d.Spec.Template.Labels).Get("app")
+	for _, wl := range wls {
+		appLabel := wl.Labels.Get("app")
```
lucasponce, line 218: Note that these constants can be also changed in Istio, so better to read from the main config to get the IstioAppLabel

business/istio_status.go
```diff
@@ -189,11 +228,11 @@ func (iss *IstioStatusService) getStatusOf(ds []apps_v1.Deployment) (IstioCompon
 		// Component found
 		cf[appLabel] = true
 
-		if status := GetDeploymentStatus(d); status != Healthy {
+		if wl.Status != Healthy {
```
lucasponce, line 231: This logic is linked with my previous comment.
Basically it's duplicating what business workloads is doing, so I propose to use that instead and unify.

Perhaps we can discuss if we want to do it in this or other PR, but I'm sure we need to move in that direction.

business/istio_status.go
xeviknal: This was changed in https://github.com/kiali/kiali/pull/3544. Istio doesn't change the `app` label in their deployments (https://github.com/istio/istio/issues/29742)

hhovsepy:
Logs in ingress pods:
2021-03-18T09:02:43.105741Z	warn	Envoy proxy is NOT ready: config not received from Pilot (is Pilot running?): cds updates: 0 successful, 0 rejected; lds updates: 0 successful, 0 rejected
2021-03-18T09:02:44.993614Z	info	xdsproxy	connected to upstream XDS server: istiod.istio-system.svc:15012
2021-03-18T09:02:44.995477Z	warn	xdsproxy	upstream terminated with unexpected error rpc error: code = Unknown desc = authentication failure
2021-03-18T09:02:44.995832Z	error	xdsproxy	upstream send error for type url type.googleapis.com/envoy.config.cluster.v3.Cluster: EOF

xeviknal:
@lucasponce as we discussed yesterday offline. I leave this PR in standby. The idea is to fine-tune the method fetchWorkloads (in the business) to include the `DaemonSets` controller. Once this is done, I'll rebase the changes and apply it in this branch logics.

lucasponce:
Thanks @xeviknal I'll put on my todo list !

xeviknal:
@hhovsepy @lucasponce I've approached this issue using the `fetchWorkloads` methods. This way, this features supports all kinds of controllers for istio components.
Thanks @lucasponce for approaching the task of supporting daemonsets.

hhovsepy:
`2021-04-13T08:21:47.684018Z`	warn	Envoy proxy is NOT ready: config not received from Pilot (is Pilot running?): cds updates: 0 successful, 0 rejected; lds updates: 0 successful, 0 rejected
2021-04-13T08:21:48.094949Z	info	xdsproxy	connected to upstream XDS server: istiod.istio-system.svc:15012
2021-04-13T08:21:48.096951Z	warn	xdsproxy	upstream terminated with unexpected error rpc error: code = Unauthenticated desc = authentication failure
2021-04-13T08:21:48.097664Z	warning	envoy config	StreamAggregatedResources gRPC config stream closed: 0, 
2021-04-13T08:21:48.100136Z	error	xdsproxy	upstream send error for type url type.googleapis.com/envoy.config.cluster.v3.Cluster: EOF
2021-04-13T08:21:49.684362Z	warn	Envoy proxy is NOT ready: config not received from Pilot (is Pilot running?): cds updates: 0 successful, 0 rejected; lds updates: 0 successful, 0 rejected`

xeviknal:
@hhovsepy I think this is an issue of the deamonset itself not about kiali. I know you mentioned that it might be related to a known issue for your environment.
@lucasponce It would be useful to check if that happen to you in your environment as well. It doesn't happen to my minikube.

hhovsepy:
When the ingress Deployment is deleted, istiod pod fails with error in logs:
`2021-04-15T05:32:31.823301Z	error	ads	Failed to authenticate client from 10.129.0.87:41414: Authenticator ClientCertAuthenticator: no verified chain is found; Authenticator KubeJWTAuthenticator: could not get cluster Kubernetes's kube client
2021-04-15T05:32:31.911736Z	warn	serverca	Authentication failed for 10.130.0.62:40664: Authenticator ClientCertAuthenticator at index 0 got error: no verified chain is found. Authenticator KubeJWTAuthenticator at index 1 got error: could not get cluster Kubernetes's kube client.
`
Then the ingress DaemonSet pods fail with error:
`2021-04-15T05:35:57.094732Z	warn	sds	failed to warm certificate: failed to generate workload certificate: create certificate: rpc error: code = Unauthenticated desc = request authenticate failure
2021-04-15T05:35:57.215112Z	warning	envoy config	StreamAggregatedResources gRPC config stream closed: 8, grpc: received message larger than max (24494078 vs. 4194304)
2021-04-15T05:35:57.220589Z	info	xdsproxy	connected to upstream XDS server: istiod.istio-system.svc:15012
2021-04-15T05:35:57.220703Z	warn	xdsproxy	downstream terminated with unexpected error rpc error: code = ResourceExhausted desc = grpc: received message larger than max (24494078 vs. 4194304)
`

lucasponce (approved):
Tested locally and code looks good to me.
In minikube I hadn't problems to reproduce @xeviknal steps

lucasponce:
> business/istio_status_test.go

@xeviknal I see it requires a rebase, but I tested the PR as it's if the rebase requires additional test, let me know.

lucasponce:
I couldn't reproduce your error @hhovsepy but I'm using a minikube platform:

image

So, things look ok from here.

lucasponce (approved):

hhovsepy (approved):
Regression tested the PR only.
The DaemonSet example is not official and does not work on OCP (causing https://github.com/istio/istio/issues/23879).

Issue #3526, opened 2020-12-11
[multi-cluster] Discover other Kiali instances and provide a way to jump to those other instances


israel-hdez:
**Is your feature request related to a problem? Please describe.**
In a multi-cluster enabled service mesh, it's possible to deploy a Kiali instance per control plane. Currently, each Kiali is capable of observing only it's adjacent control plane. Thus, it's desirable to be able to know what other Kialis are available in the mesh. Also, it's desirable to be able to jump to other Kiali instances within a Kiali.

**Describe the solution you'd like**
We already have a "Mesh" page listing the clusters that are part of the service mesh. It could be enhanced to list discovered Kialis.

**Additional context**
We may need to assume:
* All control planes are installed in a namespace named the same.
* Kialis are installed in the istio namespace?
  * We could try to look for the Kiali CR and resolve from there.
* There is _only one_ Kiali per control plane

Epic: https://github.com/kiali/kiali/issues/1833

PR #3768, opened 2021-03-12
Extract external-url annotation when discovering Kialis

israel-hdez:
The annotation value will be passed to the front-end and this will make cross-linking possible.

Related front-end PR: https://github.com/kiali/kiali-ui/pull/2102

Related to kiali/kiali#3526


jshaughn (approved):
This seems OK to me.

PR #3770, opened 2021-03-15
Replace remaining jaeger word with tracing

ohkinozomu:
Probably this is rest of https://github.com/kiali/kiali/pull/1022


config.yaml
```diff
@@ -22,7 +22,7 @@ external_services:
   # Uncomment istio_identity_domain to set a different value. This value must match the Istio configuration.
   # Default is "svc.cluster.local" (which matches Istio's default)
   # istio_identity_domain: svc.cluster.local
-  jaeger:
+  tracing:
```
jmazzitelli, line 25: This entire config.yaml should be deleted. This whole `external_services` section is obsolete and I'm pretty sure won't even work.

ohkinozomu: I found config.yaml referrenced in 2 files.
https://github.com/kiali/kiali/blob/8df7c28c68b367b3e75b47b12921556702b818ab/business/iter8.go#L421
https://github.com/kiali/kiali/blob/c49238fc81b486f199c06e7ce69efed5081a3069/make/Makefile.build.mk#L89

Should I only fix external_services section?

ohkinozomu: GetAnalyticPort(iter8.go) is not used in this repository, so perhaps it can be removed.
About Makefile, if not configuration file specified, kiali cli will use environment variables and work, so it can be removed.

jmazzitelli: kiali/business/iter8.go isn't referencing that file - it is referencing the file mounted from the ConfigMap - that is unrelated.
The Makefile "run" target is interesting... maybe someone is using that somewhere, but I don't know who or why.

ohkinozomu: > kiali/business/iter8.go isn't referencing that file - it is referencing the file mounted from the ConfigMap - that is unrelated.

Ok, thank you.

> The Makefile "run" target is interesting... maybe someone is using that somewhere, but I don't know who or why.

The README.adoc introduces it in Running Standalone section.
https://github.com/kiali/kiali/blob/master/README.adoc#running-standalone

ohkinozomu:
I deleted config.yaml.
"make run" fails, but it did so even before I removed it.

```
Running...
2021-03-17T07:25:17+09:00 INF Kiali: Version: v1.32.0-SNAPSHOT, Commit: 95aeeb2f2d880b59575e5a939e420d2de5e7ddbe

2021-03-17T07:25:17+09:00 DBG Kiali: Command line: [/Users/nozomu/go/bin/kiali]
2021-03-17T07:25:17+09:00 INF No configuration file specified. Will rely on environment for configuration.
2021-03-17T07:25:17+09:00 FTL server static content root directory does not exist: /opt/kiali/console
make: *** [run] Error 1
```

make/Makefile.build.mk
```diff
@@ -86,7 +86,7 @@ test-e2e:
 ## run: Run kiali binary
 run:
 	@echo Running...
-	@LOG_LEVEL="debug" ${GOPATH}/bin/kiali -config config.yaml
+	@LOG_LEVEL="debug" ${GOPATH}/bin/kiali
```
jmazzitelli, line 89: This is always going to break. If no one uses the "run" target, we should just delete the entire target.
@lucasponce do you (or anyone) use this "run" target anymore??

lucasponce: @jmazzitelli I use a local run configuration managed by the IDE
image

But I don't run the local config through the make run target.

jmazzitelli: OK, so you just explicitly run the kiali binary with our own custom config.yaml from your IDE - you don't use that "run" target that uses the old "config.yaml" that is the repo today.

I don't think anyone uses this "run" target today - I say that because [the config.yaml](https://github.com/kiali/kiali/blob/master/config.yaml) we have is outdated (hasn't been updated in over a year) and I doubt it will work.

We should just remove that config.yaml (which this PR does) and remove the "run" target (this PR should be updated to do that). If someone wants to run locally, they will just do what Lucas does - run the binary in their own environment with their own custom config.yaml.

In fact, we should probably document this in the README - it might be useful to tell people how to run this locally. Notice people have to know to set KUBERNETES_SERVICE_[HOST,PORT] env vars to connect to the remote k8s api server (and notice our "run" target doesn't do this - so that's another reason why it won't work and why we should remove it). The `-config` argument is set to a config.yaml that is nothing more than the ConfigMap data - you can take the `spec` field of [the example kiali_cr.yaml](https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml) and use it as a basis for your custom config.yaml. We can do this in another PR, though.

In short, I think this PR should be updated to remove the "run" make target entirely.

make/Makefile.build.mk
```diff
@@ -86,7 +86,7 @@ test-e2e:
 ## run: Run kiali binary
```
jmazzitelli, line 86: Delete lines 86-90 entirely - remove the "run" make target because it is obsolete and won't work even if you try it.

jmazzitelli (changes requested):

make/Makefile.build.mk
jmazzitelli: I created a new github issue to track my request to update the README with the instructions on how to run locally. See: https://github.com/kiali/kiali/issues/3781

lucasponce: Yes, +1 on that, in short my feedback is I don't envision coding Kiali in a remote way :) I always have my local processes in my ide, so perhaps we need to refresh the howto/readme to make it more clear and avoid confusion.

I guess that same process work for other ides as well, I use Goland, but in a standard way to run a go program with extra parameters to tell where to connect and where to read the config doc.

jmazzitelli (approved):
I'm OK with this. I will approve.
I'll wait a little while to give others a chance to review this. If someone uses the 'run' make target, speak now or forever hold your peace :)

make/Makefile.build.mk
xeviknal: I am having thoughts about removing this file. While working in a local environment, you need to pass to kiali the "config map" with all the settings. However, in local, the config map is not available. So you need a yaml that just contains the settings. For me, that yaml should be this `config.yaml`. It might be just a starting point: copy it and tune it for your local development.

I find this template useful as part of the developer experience. Actually, I think we should update this template to perfectly work with the `demo` profile. This way, a developer should be able to run kiali in local just doing:
`kiali -config config.yaml` (using their IDE).

If we want people to open PR's: we should reduce friction in the process of running kiali in dev mode.
Istio does't it in a wiki format: https://github.com/istio/istio/wiki/Preparing-for-Development

ohkinozomu: I'm very newcomer about Kiali, but as the creator of this PR, I would like to offer my opinion.

I installed Kiali with helm, but there was not much information about external_services in values.yaml( https://github.com/kiali/helm-charts/blob/72cfb8f829266e67e9fa7cc729bbf51ed2e0cd1d/kiali-server/values.yaml#L65-L67 ), so I referred to config.yaml in kiali repository(and created this PR).
In this respect, config.yaml can be useful.
However, if we keep config.yaml as a template, need to rename it to e.g. config.yaml.example and add config.yaml to .gitignore.

Or, this will be a big another issue, we may need  document about kiali configuration. Sorry if it's already there, but I referred to config.go many times.

jmazzitelli: @xeviknal I think we should do that as part of the work for that new github issue I created (https://github.com/kiali/kiali/issues/3781)

We can have this PR do what it is doing (remove "run" and the config.yaml, because they don't work right now).
Then we can create a "better dev experience" as part of the new PR for that issue.

This probably means we should put an "example" config.yaml somewhere (probably not in the root dir where it is now, that's just confusing because it makes it look like it is a really important file to be in the top root dir, when it really is a dev-only file). Perhaps we should have a "dev" directory or put it in our already existing "hack" directory -- then, in that directory, we put an example `dev-configmap-data.yaml` file and also maybe a simple script that will do something like this:

`hack/run-local-kiali.sh`:
```
#!/bin/bash

test -z "${KUBERNETES_SERVICE_HOST}" && echo "Please set KUBERNETES_SERVICE_HOST" && exit 1
test -z "${KUBERNETES_SERVICE_PORT}" && echo "Please set KUBERNETES_SERVICE_PORT" && exit 1
export KUBERNETES_SERVICE_HOST KUBERNETES_SERVICE_PORT

KIALI_CONFIG_FILE="${KIALI_CONFIG_FILE:-$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)/dev-configmap-data.yaml}"
KIALI_EXE="${KIALI_EXE:-${GOPATH}/bin/kiali}"

"${KIALI_EXE}" -config "${KIALI_CONFIG_FILE}"
```
Of course we need to update the README too - but all of that should be part of the new PR. In fact, I should probably copy this comment into that github issue :)

lucasponce: +1 but if we open the scope, perhaps we can make that in a followup PR.

jmazzitelli: > This way, a developer should be able to run kiali in local just doing:
> `kiali -config config.yaml` (using their IDE).

AFAIK, they can't run it by just that command - the problem is they need the KUBERNETES_SERVICE_xyz environment variables in order to connect to the remote k8s api. That's why I recommend a script (or we resurrect the "run" make target if we wish, but I have a feeling people don't like running make for something like this.)

jmazzitelli: > I'm very newcomer about Kiali, but as the creator of this PR, I would like to offer my opinion.
> 
> I installed Kiali with helm, but there was not much information about external_services in values.yaml( https://github.com/kiali/helm-charts/blob/72cfb8f829266e67e9fa7cc729bbf51ed2e0cd1d/kiali-server/values.yaml#L65-L67 ), so I referred to config.yaml in kiali repository(and created this PR).
> In this respect, config.yaml can be useful.
> However, if we keep config.yaml as a template, need to rename it to e.g. config.yaml.example and add config.yaml to .gitignore.
> 
> Or, this will be a big another issue, we may need document about kiali configuration. Sorry if it's already there, but I referred to config.go many times.

Actually, you should have been referred to this example file (this is where we document all the settings): https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml

But your comment there illustrates what I was talking about when I said earlier we should not put it in the root dir because, "that's just confusing because it makes it look like it is a really important file". People are going to quickly see "oh, there's a config file here" and it really isn't what they want.

jmazzitelli: > +1 but if we open the scope, perhaps we can make that in a followup PR.

Yes. I agree. I believe that work should be done as part of this: https://github.com/kiali/kiali/issues/3781

jmazzitelli: let's move this discussion over to that other issue - we can fine-tune what we want by discussing it there: https://github.com/kiali/kiali/issues/3781

@lucasponce @xeviknal is there an objection to just merge this PR as-is? We can always resurrect things (the make target, or config.yaml, or both, or none) in the new PR for that issue. I think I am leaning towards doing something in the `hack/` directory (since that is where all our other dev-environment scripts are) but we can discuss over in that issue.

lucasponce: Not from my side, as commented, I guess there is a WIP to clean things and move things in a separate effort.

xeviknal: Agree to add more 'developer-friendly' material as part of #3781

xeviknal (approved):
LGTM! Nice PR that triggered conversations to improve the developer experience. 
See more comments in #3781

Issue #3772, opened 2021-03-16
Deployment models: to support managed control plane


feuyeux:
**Is your feature request related to a problem? Please describe.**
Thinking about the Kiali topoloy, it's strong dependency with istiod and assumed that Kiali deploy with istiod in a same kubernetes cluster. And the interfaces, client and cache are not decoupled from kube for data plane cluster and mesh for control plane cluster.
Indeed, ASM([Alibabacloud ServiceMesh](https://www.alibabacloud.com/product/servicemesh)) has a different topology, which is illustrated below. 
![image](https://user-images.githubusercontent.com/769711/111264187-0abce280-8662-11eb-93bc-a48fa823af98.png)

Like istio1.8+ primary-remote, we provided a managed control plane to our customers, and,  in the managed cluster(primary cluster), considering security and scalability, it's not recommended to let the components to access external network. So, Kiali should not be deployed on primary cluster, and it should be deployed in users' data plane(remote cluster). Furthermore,  we cannot collect user's data, so prometheus should be deployed on remote. Therefore, kiali and prometheus are deployed in a same cluster with users' pods and services.

**Describe the solution you'd like**
To let kiali support this topology, we need a set of decouping interfaces defined clearly to separate accessing primary(istio) and remote(kubernetes). We should use mesh_interfaces to visit primary cluster to CRUD service mesh api, and use kube_interfaces to visit remote cluster to CRUD kubernetes api. In addtion, we need to use mesh_client and mesh_cache in network layer to connect to primary api-server, and to use kube_client and kube_cache to talk with remote cluster.

lucasponce:
Thanks @feuyeux for your inputs.
Indeed, this is an area of active development and it helps us to drive the multicluster/federation epics which we are working on.

Technically speaking, when developing it's normal to have a external Kiali decoupled from the control plane, just telling Kiali where to connect.

Side comment, would you like to actively collaborate on this effort from ASM perspective ?

Perhaps it's a nice area where we can join forces and work together in these efforts as well.

feuyeux:
Thanks @lucasponce for your reply.
I like to learn more about that. If our community starts this work, please let me know more details, e.g. which branch developing.
I did some changes from branch v1.31, which were passed all tests by executing the below script:
```bash
export PRIMARY_KUBE_CONFIG=$HOME/shop_config/kubeconfig/asm_production
export REMOTE_KUBE_CONFIG=$HOME/shop_config/kubeconfig/ack_production
[...]

PR #3773, opened 2021-03-16
to support managed control plane topology

feuyeux:
** Describe the change **
1. decouple the interfaces/clients/caches of mesh from kube.
2. use kubeconfig files env running on local.
3. uniform timeout defination.

** Issue reference **

https://github.com/kiali/kiali/issues/3772

** Backwards incompatible? **

_describe the changes if appropriate_

** Documentation **


lucasponce:
@feuyeux 

First of all, thanks a lot for the Pull Request.
I really appreciate the effort to take time to prepare this.
It helps to understand better your proposal.

We are going to review it and keep discussing about it.

From high level, I understand that we want to progress to make Kiali available in muliple control plane deployment strategies.

Perhaps the PR as it is it may require some changes as the support of multiple control planes is in active development and we need to make sure that we don't introduce regressions or choose any design that make block other strategy as well.

But we will work on this with you study a proposal that may help to support all strategies proposed.

israel-hdez:
So, I understand you are breaking the k8sclient into a KubeClient and a MeshClient. I understand the KubeClient is the one that talks to the _data plane_ cluster, while the MeshClient talks to the _control plane_ cluster.

I think this partially solves the Primary-Remote topology, but I'm not sure this is the right way to go because this works when there is only one primary cluster and only one remote cluster. However, the Primary-Remote topology can involve multiple remote clusters managed by a single primary cluster. So, I don't think this can be merged as is.

Also, I've heard that Istio may remove the need to create CRDs in the control plane cluster making it read-only, and allow CRDs to be created in the data plane. Probably, this is something worth to ask in Istio forums. If this is really planned, this may remove the need of splitting the k8sclient.

Anyway, I think I get the idea behind all these changes.

lucasponce:
I like comment from @israel-hdez to adjust the naming to "control plane" or "data plane".
So, in topologies where there is a single "control planes" managing several "data planes", the "data planes" could be better identified.

Perhaps this open the top-down discussion how we want to support this topology.

But in any case I'm in favor to make compatible the code to support multiple scenarios, perhaps @feuyeux you could also to give some info about the plans, from architecture perspective ?

feuyeux:
There's only one control plane and multiple data plane, including vm. And we read/write mesh crd with mesh client. This is our long term design and it's not easy to change. 
But we don't want to use write in kiali, since we have a gui console to do that. 
For kiali perspective, to support multiple is satisfied for us, even considering mesh crd read only.

lucasponce:
Hi @feuyeux sorry for the delay, the deployment models discussion is quite active and we'd like to have your feedback to help us to set up the priorities for this work I've created a discussion to set up a community session around the topic:

https://github.com/kiali/kiali/discussions/4013

feuyeux:
> Hi @feuyeux sorry for the delay, the deployment models discussion is quite active and we'd like to have your feedback to help us to set up the priorities for this work I've created a discussion to set up a community session around the topic:
> 
> #4013

Got it and thanks a lot !

lucasponce:
Hi @feuyeux,

Please, feel free jump on this discussion https://github.com/kiali/kiali/discussions/4081 where we are defining the high level design of the support for multi deployment models.

Probably we'll reuse some of the ideas that you proposed in the PR but I suggest to close it for the moment and we can re-open it and check if we can re-use some of the work.

Let me know what do you think.
Thanks !

Issue #3731, opened 2021-02-24
[hack] because on OpenShift we use CNI, no need to make the containers privileged


jmazzitelli:
Our hack scripts do things like this when on openshift:

```
$ oc adm policy add-scc-to-group privileged system:serviceaccounts:<target-namespace>
$ oc adm policy add-scc-to-group anyuid system:serviceaccounts:<target-namespace>
```

This is no longer needed as CNI is required on OpenShift and with CNI there is no need to have privileged containers.

See the old docs: https://istio.io/latest/docs/setup/platform-setup/openshift/ - but I believe this page needs to be updated

PR #3774, opened 2021-03-16
no need for these extra permissions anymore due to the use of CNI on OpenShift

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3731


jmazzitelli:
I have not tested any of the scripts yet, but I did go through and remove all the `adm policy` calls that create these permissions. Just need to confirm these really are not needed and that the hack scripts all still work.

jmazzitelli:
Confirmed that `install-istio-via-istioct.sh` and `install-bookinfo-demo.sh` work

jmazzitelli:
Confirmed `install-travel-agency-demo.sh` works (I am not testing the 1.6 one - but the same changes were made to it as this newer one. I think we should delete the 1.6 one anyway).

jmazzitelli:
the error rates demo install script works too. So I can confirm all works. We can merge.

jmazzitelli:
@jshaughn do we want to just delete the 1.6 version of the travel agent script?

jshaughn:
> @jshaughn do we want to just delete the 1.6 version of the travel agent script?

I think that's fine at this point as 1.6 is out of support and the documented config for operation_id has changed significantly for the stats filter.  Actually, I'd like for the operation_node support to be configurable as a command line option instead of the way it's done currently with manual edits and support only for a specific version.  I haven't had time to do that, and it would also likely require a different way of updating the stats filter as the patch approach is pretty brittle.

jmazzitelli:
1.6 file has been removed

jshaughn (approved):
LOK

PR #3779, opened 2021-03-16
[hack] this is still needed - this perm is required in the control plane ns

jmazzitelli:
taking away this permission was wrong - we still need it.


jshaughn (approved):
LOK

PR #3780, opened 2021-03-16
Bump golang version to 1.16.2

### LINKED ISSUE #3722: Move to golang 1.16
Golang version 1.16 was released recently. See announcement in the official golang blog: https://blog.golang.org/go1.16.

Per their [release maintenance policy](https://github.com/golang/go/wiki/Go-Release-Cycle#release-maintenance), golang 1.14 (which Kiali is using) is now unmaintained.

Kiali should move to golang 1.16 which will give about a year of patch updates. Alternatively, it's possible to move to go1.15, but this will give only about 6 months of patch updates before we need to consider (again) bumping to a newer Y-stream release.

israel-hdez:
Because go 1.14 is now unmaintained.

Fixes #3722

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-20.04
     steps:
 
-      - name: Set up Go 1.x
+      - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.14
+          go-version: 1.16.2
 
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


jmazzitelli:
This is going to affect productization - we should not move to another Go version until we can confirm we can productize on it as well.

israel-hdez:
Jenkins is ready to do the builds

jshaughn (approved):
Sinc Jenkins is OK with it I'm fine with the change to 1.16.2.  Hopefully brew will also support it but I suppose we could build prod with an older version as long as our code doesn't use any newer Golang feature/syntax.

Issue #2895, opened 2020-06-11
Divide the release pipeline


israel-hdez:
This task is about the release pipeline in https://github.com/kiali/kiali/tree/master/deploy/jenkins-ci.

Currently, it handles releasing the front-end to NPM, the back-end (bundled with the front-end) as a container image, the operator as a container image, the website, and kiali helm charts. It also updates the repositories in preparation for the next release.

The pipeline is becoming big and it's doing unrelated tasks. The only thing that share in common is that all tasks are ran when it's time for a new Kiali release.

In order to bring better maintenance to the pipeline, the _divide-and-conquer_ principle should be applied to it:
* [x] Extract the _release operator_ part to it's own pipeline kiali/kiali#3567, kiali/kiali-operator#233
* [x] Extract the _release website_ part to it's own pipeline kiali/kiali#3540, kiali/kiali.io#335
* [ ] Extract the _release Kiali app_ to it's own pipeline
* [x] Extract the _release helm charts_ to it's own pipeline kiali/kiali#3546, kiali/helm-charts#27
* [ ] Then, the original _kiali-release_ pipeline would become a simple pipeline that orchestrates running the other pipelines in the right order (because there is some implied order).
* [x] While doing the refactoring, it will be good if the number of parameters can be reduced, because some of these can be built from a more simple parameter. #3783
* [x] Update Jenkins's README.md. #3783
- [x] Add notifications for failed builds. #3783 (only for main pipeline)

israel-hdez:
I've finished most of the planned tasks.
The remaining ones, I'm not doing them for now. They are, probably, not worth the time at the moment.

So, I'm closing as done.

PR #3783, opened 2021-03-17
Simplify pipeline parameters and update README

israel-hdez:
This is also adding mail notifications for failed builds... only if back-end or front-end fails. Operator, Helm charts and website won't have failure notifications, because that needs to be added in the corresponding repos.

Related to #2895

### DIFF
--- a/deploy/jenkins-ci/Dockerfile
+++ b/deploy/jenkins-ci/Dockerfile
@@ -21,9 +21,6 @@ USER root
 COPY bin/entrypoint.sh /usr/local/bin/
 
 RUN set -eux; \
-    # Install a requirement to un-tar nodejs.
-    apt-get -qq update; \
-    apt-get install --no-install-recommends -qq xz-utils; \
     # Install golang \
     curl -fsSL https://golang.org/dl/go1.14.7.linux-amd64.tar.gz -o go.tar.gz; \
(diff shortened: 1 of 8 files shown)


israel-hdez:
This is ready.

This is not _very_ backwards compatible. I'm thinking, mainly, on those older Kiali versions that we release from time to time, like v1.12. I'll fix any issue when it's time to build an old Kiali.

deploy/jenkins-ci/Dockerfile
```diff
@@ -32,10 +29,10 @@ RUN set -eux; \
     # curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -; \
     # echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list; \
     # Install NodeJs \
-    curl -fsSL https://nodejs.org/dist/v10.19.0/node-v10.19.0-linux-x64.tar.xz -o node.tar.xz; \
+    curl -fsSL https://nodejs.org/dist/v14.16.0/node-v14.16.0-linux-x64.tar.gz -o node.tar.gz; \
```
jmazzitelli, line 32: just a side comment: we should update our productization instructions - I believe we are still using node10

jmazzitelli (approved):
lgtm (for whatever that is worth :))

Issue #3781, opened 2021-03-17
Update the "Running Standalone" section in the README


jmazzitelli:
Document in the [README.adoc "Running Standalone" section](https://github.com/kiali/kiali/blob/master/README.adoc#running-standalone) how to run Kiali locally, while connecting to a remote cluster's k8s master api endpoint.

Note that the "run" make target is obsolete and should not be used (it should be removed). You have to provide your own custom config.yaml (just the ConfigMap - the `spec` section of [the example kiali_cr.yaml](https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml) can be a starting point) and you have to provide environment variables KUBERNETES_SERVICE_[HOST,PORT] to tell Kiali where the cluster's k8s api endpoint is.

For more, see: https://github.com/kiali/kiali/pull/3770#discussion_r595857368

jmazzitelli:
We should open the scope of this issue to not only update the README but also make it easier for the user, too.

See: https://github.com/kiali/kiali/pull/3770#discussion_r596077625

My ideas are:

Create a "better dev experience" as part of this issue by putting an "example" config.yaml somewhere (probably not in the root dir where it is now, that's just confusing because it makes it look like it is a really important file to be in the top root dir, when it really is a dev-only file). Perhaps we should have a "dev" directory or put it in our already existing "hack" directory -- then, in that directory, we put an example `dev-configmap-data.yaml` file and also maybe a simple script that will do something like this:

`hack/run-local-kiali.sh`:
```
#!/bin/bash

test -z "${KUBERNETES_SERVICE_HOST}" && echo "Please set KUBERNETES_SERVICE_HOST" && exit 1
test -z "${KUBERNETES_SERVICE_PORT}" && echo "Please set KUBERNETES_SERVICE_PORT" && exit 1
export KUBERNETES_SERVICE_HOST KUBERNETES_SERVICE_PORT

KIALI_CONFIG_FILE="${KIALI_CONFIG_FILE:-$(cd "$(dirname "${BASH_SOURCE[0]}")" &> /dev/null && pwd)/dev-configmap-data.yaml}"
KIALI_EXE="${KIALI_EXE:-${GOPATH}/bin/kiali}"

"${KIALI_EXE}" -config "${KIALI_CONFIG_FILE}"
```
Of course we need to update the README too - but all of that should be part of the new PR.

jmazzitelli:
[...]

PR #3784, opened 2021-03-17
script that you can use to run kiali on the local machine

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3781

This script will allow you to run Kiali on the local machine so you can do quick code/build/test cycles without having to deploy the server on the cluster.

It attempts to auto-discover the k8s api master endpoint - but there are cmdline options to let you override it.
It attempts to auto-discover the location of your UI console build - but there is a cmdline option to tell it where it is.

There are other cmdline options available, such as being able to tell the script to download a specific version of the UI console in order to be able to test a specific UI release.

The config can be overridden, but it will provide you with a set of defaults out of box. Nothing special today, but I suspect in the future (maybe part of this PR if I have time) we can add cmdline options that let you configure the ConfigMap yaml without you having to maintain your own yaml file somewhere.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -211,24 +211,18 @@ make operator-delete
 [NOTE]
 After this completes, the `kiali-create` and `kiali-delete` targets will be ineffective until you run the `operator-create` target to re-deploy the Kiali operator again.
 
-=== Running Standalone(WIP)
+=== Running Standalone
 
-Rarely, you may want to run Kiali outside of any cluster environment, perhaps for debugging purposes. To do this, run:
-
-[source,shell]
-----
-cd ${GOPATH}/src/github.com/kiali/kiali
-make install
-make run
-----
-
-The "install" target installs the Kiali executable in your GOPATH /bin directory so you can run it outside of the Makefile:
+Rarely, you may want to run Kiali outside of any cluster environment, perhaps for debugging purposes. To do this, you
+will want to use the link:./hack/run-kiali.sh[run-kiali.sh hack script] located in the
+link:./hack[hack directory]. See the `--help` output for the options you can set.
+The default configuration it uses is found in the link:./hack/run-kiali-config-template.yaml[config template file]
+also located in the `hack` directory. Read the comments at the tops of both files for more details.
 
 [source,shell]
 ----
-cd ${GOPATH}/src/github.com/kiali/kiali
-make install
-${GOPATH}/bin/kiali -config <your-config-file>
+cd ${GOPATH}/src/github.com/kiali/kiali/hack
+./run-kiali.sh
 ----
 
 == Configuration
--- a/hack/run-kiali-config-template.yaml
+++ b/hack/run-kiali-config-template.yaml
@@ -1,0 +1,51 @@
+###################################################################
+# run-kiali-config-template.yaml
+#
+# This is intended to be used as a template configuration file
+# for the developer tool 'run-kiali.sh'.
+#
+# This file is not intended to be used by Kiali other than when
+# launched locally via 'run-kiali.sh'.
+#
+# If you want to know what settings Kiali supports, please see
+# the spec field in the example Kiali CR found here:
+# https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml
+#
+###################################################################
+
+istio_namespace: "${ISTIO_NAMESPACE}"
+
+auth:
+  strategy: "anonymous"
+
+deployment:
+  accessible_namespaces: ['**']
+
+external_services:
+  custom_dashboards:
+    enabled: false
+
+  grafana:
+    enabled: true
+    in_cluster_url: "${GRAFANA_URL}"
+    url: "${GRAFANA_URL}"
+
+  prometheus:
+    url: "${PROMETHEUS_URL}"
+
+  tracing:
+    enabled: true
+    in_cluster_url: "${TRACING_URL}"
+    url: "${TRACING_URL}"
+
+kubernetes_config:
+  cache_enabled: true
+
+login_token:
+   signing_key: "notsecure"
+
+server:
+  static_content_root_directory: "${UI_CONSOLE_DIR}"
+
+# in_cluster must be false - do not change
+in_cluster: false
--- a/hack/run-kiali.sh
+++ b/hack/run-kiali.sh
@@ -294,6 +294,13 @@ if [ -z "${PROMETHEUS_URL:-}" ]; then
         exit 1
       else
         warnmsg "Cannot auto-discover Prometheus on OpenShift. If you exposed it, you can specify the Prometheus URL via --prometheus-url. For now, this session will attempt to port-forward to it."
+        prom_remote_port="$(${CLIENT_EXE} get service -n ${ISTIO_NAMESPACE} prometheus -o jsonpath='{.spec.ports[0].targetPort}')"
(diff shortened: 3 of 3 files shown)


jmazzitelli:
This almost works. I am seeing these errors at startup, even though my KUBERNETES_SERVICE_HOST and _PORT env vars appear set correctly:

```
===== SETTINGS =====
...
KUBERNETES_SERVICE_HOST=api.jmazziteos4.lab.upshift.rdu2.redhat.com
KUBERNETES_SERVICE_PORT=6443
...
===== SETTINGS =====
Starting the Kiali server...
2021-03-17T16:30:56-04:00 INF Kiali: Version: v1.32.0-SNAPSHOT, Commit: 398cb24f5de7c655283ec76647072093ce8342f6

...
2021-03-17T16:31:33-04:00 ERR Error initializing Kiali Cache. Details: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
2021-03-17T16:31:33-04:00 ERR open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
2021-03-17T16:31:33-04:00 INF [Prom Cache] Enabled
2021-03-17T16:31:33-04:00 ERR Failed to fetch Prometheus configuration: Get "/api/v1/status/config": unsupported protocol scheme ""
2021-03-17T16:31:33-04:00 ERR Failed to fetch Prometheus flags: Get "/api/v1/status/flags": unsupported protocol scheme ""
2021-03-17T16:31:33-04:00 WRN Failed to fetch Kiali token when resolving cluster info: open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory
```

If someone can tell me what I'm doing wrong, it would be appreciated :)

@lucasponce - since you have this working, perhaps you can tell me what I am missing? Are there other env vars I need to set?

jmazzitelli:
I created a Prometheus Route and used that in my config and I no longer see the Prom errors. So for now, you have to expose the Prom service somehow. I can see if I can automate this, but for now, it is a manual step.

Because we are running out of the cluster, I think it means we need to set this hidden config setting "in_cluster: false". I did that, but now I see this in the logs:

```
ERR Error fetching the Kubernetes client: parse "http://http://api....com:6443\x1b[0m": net/url: invalid control character in URL
```

This doesn't make sense because I ensure that KUBERNETES_SERVICE_HOST does NOT contain the schema (it is only the host name) so I do not know where the double `http://http://` is coming from. I also do not know where those weird characters come from at the end (`\x1b[0m`).

jmazzitelli:
The script is now pulling the token and certs from the kiali service account and putting them in /var/run/secrets/kubernetes.io/serviceaccount. I no longer seen messages about `open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory`

However, I still get those mysterious double `http://http://` - haven't figure out why yet.

jmazzitelli:
OK, here's the progress so far. Here's the setup:
* I have Kiali installed already in the cluster in istio-system.
* I have exposed Prometheus via a OpenShift route.
* I can run the script `run-kiali.sh` and it starts the server.
* The server has KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT set to what I think are the correct values (the host and the  port of the API server)
* I point my browser to `http://localhost:20001/kiali` and I see the Kiali screen but I get errors all revolving around this problem - if I can fix this, I think I will have the basics working:

```
ERR Error fetching the Kubernetes client: parse "https://https://api.jmazziteos4.lab.upshift.rdu2.redhat.com:6443\x1b[0m": net/url: invalid control character in URL
```

Once that is solved, I'll work on automating the creation of the Prometheus route (on OpenShift) or Ingress (on k8s).

Also, it would be nice if the server did not hardcode the `/var/run/secrets/kubernetes.io/serviceaccount` location - it would be nice if we could override that with an env var like we do for the KUBERNETES_SERVICE_HOST/PORT settings. Because as of right now, its a manual step that requires root access - the user has to create that directory location. It would be nice to be able to put the token and certs in a user directory and just point the server to read that user directory (not under /var/run).

jmazzitelli:
The basics are done.

This utilizes kubectl proxy and, if required, port-forward. It runs a kiali server on your local machine - the Kiali Server has the same access as the Kiali Service Account (so permissions are exactly how it will be when the server is running inside the cluster). It only works with "anonymous" login, however.

You can package the UI that is built on your own local machine (it will attempt to auto-detect your git repo if it is in the parent tree hierarchy of your directory structure - or you can pass in the cmdline opt if it can't auto-detect your setup) or it can package the UI that is released (it pulls from npm - you can ask to run with any version of the UI, including "latest").

If you expose Prometheus as a route on OpenShift, it will use that route rather than start a port-forward process.

One note: you MUST have deployed Kiali in the cluster - this script needs the roles/rolebindings/serviceaccount so it can mimic the permissions of a true in-cluster Kiali Server.  But once you start the script, you can access the UI directly from your local host, not from the cluster.

jmazzitelli:
Two things:

1) This does nothing about detecting or setting up the grafana or jaeger endpoints. We would have to add additional port-forwarding code (identical to the way it does it now with prometheus) in order to access grafana and jaeger.

2) When using port-forwarding, I find it annoying that if the port-forward process losing communications with the server end (not matter how briefly the disconnect is), the port-forward process dies and you lose access. This script needs to have added to it some kind of restart mechanism that restarts the port-forward process, but yet is still smart enough to know when it really should die when the script tells it to exit.

jmazzitelli:
I confirm this works with both OpenShift and minikube/Kubernetes. Just make sure your current kubectl context is pointing to the cluster you want your kiali server to "talk" to.

jmazzitelli:
This now also works even if you do not have a true Kiali deployed in the cluster. Pass in "--kube-context current" to use your own cluster credentials that will be used by Kiali - Kiali will start even if there is no service account in the cluster. This will cause some errors in the server (because it sometimes needs the service account to do things) but for those use cases where you don't care, this can be handle because it means you just need a cluster with istio installed - no kiali deployment needed.

lucasponce:
Thanks @jmazzitelli I missed the notification, I'll take a look today.

jmazzitelli:
Added `--enable-server` option - if you set it to `false`, the server won't start - but tells you how you can start it. This allows you to work with a debugger IDE - just have the debugger start the server with these options after you run the script (this is a message the script prints out so you can see what values the debugger IDE should use):

```
The server has not been enabled - you must start it manually with the following environment variables and command:
export KUBERNETES_SERVICE_HOST="127.0.0.1"
export KUBERNETES_SERVICE_PORT="8001"
export LOG_LEVEL="info"
"/home/gopath/here/source/kiali/kiali/bin/kiali" -config "/tmp/run-kiali/run-kiali-config.yaml"
```

jmazzitelli:
Last commit now performs an additional auto-discover check - as opposed to hardcoding the remote target ports for the three services (9090, 3000, 16686), the script will now try to determine the target port for the three services (prometheus, grafana, tracing) and will use those as the remote ports when setting up the port-forward if the port-forward is to be created. If any of those services are not found, then the hardcoded ports are used (probably won't work, but the script will try it anyway in case it does work).

lucasponce:
I tested the script and it works ok. it didn't interfere with my own setup.
I could follow the script instructions to setup my own local urls to the external services and everything worked.

lucasponce (approved):

Issue #3499, opened 2020-12-02
Enhanced logs viewing and correlation


mtho11:
The original title here was, "Correlate Traces with Logs".  After the discussion below we decided to make this into a more general Epic for enhancing what we do with logs.  I'm going to add some initial subtasks to try and capture some of the ideas, roughly ordered in a way that may make sense:

- [x] Update the currrent logs view to be more flexible and space efficient, with an eye towards composability. https://github.com/kiali/kiali-ui/pull/2104
- [x] Update the current logs view to better display access logs https://github.com/kiali/kiali-ui/pull/2104
- [ ] Provide a composable (popup/dialog) logs view with a configurable time period
- [ ] Use the composable logs view for better correlation

Original Description Below
=================
This issue starts a conversation on showing tracing information alongside the logging information to better understand what is going on with application performance and/or errors.

Logging information by itself is very useful for understanding what an application is doing performance-wise. That combined with jaeger tracing (possibly as an overlay) would allow easier identification of hot spots and their relevant logging/tracing data in an easily digestible format.

jotak:
And that could also work the other way around, from the tracing page, have the ability to show logs from a limited time frame as contextual information of a given span...

lucasponce:
> Logging information by itself is very useful for understanding what an application is doing performance-wise. That combined with jaeger tracing (possibly as an overlay) would allow easier identification of hot spots and their relevant logging/tracing data in an easily digestible format.

I agree, I'd suggest to build a story.

One thing that shares logging and tracing is the huge amount of information they provide and the difficulties to "stop" in the right place due lack of histogram or summarized info.

So, in that direction, the current work with tracing has been to combine with the telemetry data to basically respond to that: "from all information I get, which is the interesting one I should stop by".

Sharing the time controls is a good start, I guess now those controls are better normalized and it can be extended the sync between time selection from logs/metrics + tracing.

Also another suggestion from my side would be to transform the proxy-logs in some format better structured due we know what to expect, I see this impossible / hard to do in the application container due any application and use its own format, but proxy log format is more or less fixed.

That would help to jump to next steps which could be as @jotak thinking how to correlate this info, which from high level user I'd say:
- To use the telemetry to spot the desired interval to investigate.
- Move across tracing, metrics, logs, with the desired interval, and help to analyze what happened.
- Also, try that user use raw data in proxy-logs but structure data that may help to spot easily a problem and examine / link with the app log data (which is what it is).

mtho11:
[...]

PR #3785, opened 2021-03-17
Enhanced logs view with access log detail

jshaughn:
Part of Epic #3499 

UI PR: https://github.com/kiali/kiali-ui/pull/2104

This adds backend support for access log parsing.

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -38,17 +38,25 @@ type PodLog struct {
 	Entries []LogEntry `json:"entries,omitempty"`
 }
 
-// LogEntry holds a single log entry
-type LogEntry struct {
-	Message       string `json:"message,omitempty"`
-	Severity      string `json:"severity,omitempty"`
+// AccessLogEntry provides parsed info from a single proxy access log entry
+type AccessLogEntry struct {
 	Timestamp     string `json:"timestamp,omitempty"`
 	TimestampUnix int64  `json:"timestampUnix,omitempty"`
+}
+
+// LogEntry holds a single log entry
+type LogEntry struct {
+	Message        string         `json:"message,omitempty"`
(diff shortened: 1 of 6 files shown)


jmazzitelli (approved):
LGTM (though I did add a change-request to the UI PR).

business/workloads.go
```diff
@@ -318,6 +328,41 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 			entry.Severity = strings.ToUpper(severity)
 		}
 
+		// If this is an istio access log, then parse it out. Prefer the access log time over the k8s time
+		// as it is the actual time as opposed to the k8s store time.
+		if opts.IsProxy {
+			engardeParser := parser.New(parser.IstioProxyAccessLogsPattern)
```
lucasponce, line 334: I've noticed that the new parser is more heavy and it's including a new lag.
Is this new parser something that perhaps can be added as a global var ?

jshaughn: @lucasponce I suppose we could make it optional.  Do you have any numbers on the measured lag?  I did some informal tests and it didn't seem too bad.  But of course it will increase with the number of access log entries.  I went back and forth on whether to pre-parse all of the access logs, because I knew it would add fetch time and the parsed data may not be viewed.  But I really wanted the parsed timestamp (for the reasons I explained in my UI comment), and so I decided to take the hit.   An alternative would be to just parse an entry on-demand, and introduce a new API service to do it.  But I'd still probably try and pre-parse the timestamp as the integrated view is not as useful using strict k8s timestamps.

pbajjuri20 (approved):

jshaughn:
None of @lucasponce comments are blocking so I'm going to merge.  There will be more work with logs going forward.

PR #3788, opened 2021-03-18
Make hack scripts easier for deploying local demo versions

jshaughn:
Make hack scripts easier for deploying local demo versions of travel or error-rates demo

adds -s|--source option


jmazzitelli (approved):
love those hack scripts

Issue #2884, opened 2020-06-09
how to handle istio canary updates (multiple control planes in the same namespace)


jmazzitelli:
There is a new feature in Istio 1.6 that allows Istio itself to be upgraded in a "canary" fashion - multiple control planes in the same namespace.

See:
* https://istio.io/blog/2020/multiple-control-planes/
* https://istio.io/docs/setup/upgrade/

I do not believe that Kiali today can handle this (at least, we haven't done anything wrt handling a use case where multiple Kialis are installed in a single namespace OR that Kiali can run with multiple Isitos in the same control plane namespace).

We need to figure out how to handle this gracefully.

lucasponce:
Related:
https://github.com/kiali/kiali/blob/master/kubernetes/istio_details_service.go#L24

xeviknal:
Related: when having two control planes, the about modal gets a bit crazy too.

![Screenshot of Kiali Console - 2020-06-10T122933 328](https://user-images.githubusercontent.com/613814/84257713-11885900-ab16-11ea-8e6e-17c6dec9de1a.png)

jshaughn:
@lucasponce @jmazzitelli I think maybe we should hold off work here until we see better what happens in Istio 1.7 and their changes re: installation and add-ons.

xeviknal:
I think it is time to move this one to backlog. Maistra uses the `rev` label to provide the soft multitenancy:

![Screenshot of istiod-basic-install · Details · Red Hat OpenShift Container Platform](https://user-images.githubusercontent.com/613814/92477382-cb0fcc80-f1e0-11ea-8274-3232b8904242.png)

![Screenshot of istiod-basic-install-79c775b54d-xzpm5 · Details · Red Hat OpenShift Container Platform](https://user-images.githubusercontent.com/613814/92477387-cd722680-f1e0-11ea-984e-99d568bc8a68.png)

However, I think that the main change for Kiali is only to allow users specify either the `rev` name (which will be the suffix of CM name) or the configmap name. I guess that the first one is the most appropriate.
With that new setting, kiali will be able to accurately look up the correct configmap.

@lucasponce @jshaughn

jmazzitelli:
[...]

PR #3796, opened 2021-03-22
Add the IstiodDeploymentName to config and use it in the mesh business

xeviknal:
Istiod deployment might have different names depending of the instalation. for example, istio upgrade via canary deployments changes the name in the following convention `istiod-${istio-version}` (e.g. istiod-1-9-1). However, the other bits of the deployment aren't changed at all (image, labels, template labels, so on).

This PR approaches the problem as it was approached for the config map: adding a new config into the kiali CR to specify the specific name.

```
external_services:
  istio:
    istiod_deployment_name: "istiod-1-9-1"
    config_map_name: "istio-1-9-1"
```

This partially fixes https://github.com/kiali/kiali/issues/2884 (at least the original description).
This PR needs an operator PR: https://github.com/kiali/kiali-operator/pull/270

**Steps to test:**
One way to try this fix is by following the task of upgrading istio: https://istio.io/latest/docs/setup/upgrade/canary/#control-plane
Note: you'll need to change both config listed above: deployment name and config map name.

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -158,7 +158,7 @@ func (in *MeshService) ResolveKialiControlPlaneCluster(r *http.Request) (*Cluste
 
 	// The "cluster_id" is set in an environment variable of
 	// the "istiod" deployment. Let's try to fetch it.
-	istioDeployment, err := in.k8s.GetDeployment(conf.IstioNamespace, "istiod")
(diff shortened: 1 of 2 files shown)


hhovsepy (approved):
Regression tested.

israel-hdez (approved):
LGTM!

Issue #800, opened 2019-01-23
Why my istio configuration below gets warning and error?


bangau1:
Hi Guys, I have problem on **Istio Config** menu found at Kiali web.
My Kiali version is:
```
kiali-ui 0.11.0 (f0a8d441e763e62d7a39682eb60bb9b2c12d4609)
kiali v0.11.0 (bff6219618030a012c2ead4c361c6db97bbb5433)
Components
Istio 1.0.0 
```

## Subset is not found on the host

![screen shot 2019-01-23 at 16 09 06](https://user-images.githubusercontent.com/352737/51595487-692a2280-1f29-11e9-9bec-96a1c8d9d944.png)

But in fact, i have 2 deployment `stable` and `canary` both has correct `track` label on it.
```
$ kubectl get deploy -n data-devops  -l track=stable
NAME                  DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
corn-app-stable       1         1         1            1           47m

$ kubectl get deploy -n data-devops  -l track=canary
NAME              DESIRED   CURRENT   UP-TO-DATE   AVAILABLE   AGE
corn-app-canary   1         1         1            1           21m

$ kubectl get svc -n data-devops corn-app-global -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: 2019-01-23T08:25:01Z
  name: corn-app-global
  namespace: data-devops
  resourceVersion: "148223782"
  selfLink: /api/v1/namespaces/data-devops/services/corn-app-global
  uid: 60af40cc-1ee8-11e9-af67-42010a9401fd
spec:
  clusterIP: 10.32.77.78
  ports:
  - name: http-corn-app
    port: 3000
    protocol: TCP
    targetPort: 3000
  selector:
    app: corn-app
  sessionAffinity: None
  type: ClusterIP
status:
  loadBalancer: {}
```

And as much as I can tell, I can access both deployment just fine through Istio Ingress Gateway (using traffic splitting for example). So I am not sure why the Kiali indicates that error above.

## More than 1 gateways with the same host port combinaton
![screen shot 2019-01-23 at 16 06 44](https://user-images.githubusercontent.com/352737/51595692-ea81b500-1f29-11e9-83cb-d1d59b2b5f53.png)

i've checked it multiple times, but the corn-app-stg.tvlk-data.com only listed on `corn-app-gateway` gateway. 
```
[...]

PR #3800, opened 2021-03-23
KIA0203 - If host not found, kiali will never find workloads matching labels

xeviknal:
Since cross-ns validations are not in-place, when the host is not found for a DR, there is no need to check if there are workloads serving traffic for that host and matching the subset. It will always result in a not found place: the host is not found, therefore no workloads under the service listening to that host.

fixes https://github.com/kiali/kiali/issues/800#issuecomment-804741626


hhovsepy (approved):
Verified:
Screenshot from 2021-03-23 18-20-04

lucasponce (approved):

PR #3802, opened 2021-03-23
[hack] needs update to install sm2 on os 4.7

jmazzitelli:


jshaughn (approved):
This is weird but OK, I guess the elastic thing is because we need if for ossm/jaeger.

PR #3804, opened 2021-03-23
molecule tests to be able to skip operator install

jmazzitelli:
be able to pass in the new operator installer setting so you can tell the molecule tests to skip installing the operator or to use the original way - use the operator helm chart).

This requires:
* https://github.com/kiali/kiali-operator/pull/271

fixes: https://github.com/kiali/kiali/issues/3811


jshaughn (approved):
LOK

PR #3809, opened 2021-03-24
[hack] clean up the kube-system namespace of the CNI node

jmazzitelli:


jmazzitelli:
Rejecting - this isn't needed afterall. The hack script uninstalls using istioctl and it handles removing this already -- as seen in the delete output we already have:

```
daemonset.apps "istio-cni-node" deleted
```

And I confirmed it is already gone.

Issue #3826, opened 2021-03-29
Update cache for workloads


lucasponce:
Required to unblock this task:
https://github.com/kiali/kiali/pull/3764#issuecomment-801813907

To update the workloads logic with DaemonSet and other potential controllers to avoid that other logic query directly a specific controller (on this case it was Istio status, but this pattern can be used in other logic).

Note that operator and helm changes were already merged on https://github.com/kiali/kiali-operator/pull/262

PR #3827, opened 2021-03-29
Support for DaemonSet controller

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3826

Used for https://github.com/kiali/kiali/pull/3764#issuecomment-801813907, now business.WorkloadService would provide/centralize workload queries.

For testing:
- General regression on workloads logic.
- Check that DaemonSet are showed in Kiali from controllers instead from the Pods. That's seen that in DaemonSet workloads, Kiali detail will show minimal app/version label and it will only show full labels on the Pod tab perspective.


hhovsepy (approved):
Verified, regression tested:
Screenshot from 2021-03-30 09-30-34
Screenshot from 2021-03-30 09-31-13

PR #3832, opened 2021-03-30
Pass dorp to run-molecule-tests.sh

jmazzitelli:
also pass env vars to make using -e option since that's the way it should be done


jmazzitelli:
i'm a dork for forgetting to set dorp

mattmahoneyrh (approved):

PR #3846, opened 2021-04-06
support kind

jmazzitelli:
For podman support, we had to do some "tricks" with "podman save" to get it to work.

This should eventually not be needed. See https://github.com/kubernetes-sigs/kind/issues/2027 and https://github.com/kubernetes-sigs/kind/issues/2038


jshaughn (approved):
Ran it and it seems good.

PR #3848, opened 2021-04-06
[hack] be able to easily access kiali deployed in k8s via port-forward and get image names correct with podman

jmazzitelli:


hack/kiali-port-forward.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash
+
+##############################################################################
+# kiali-port-forward.sh
+#
+# This is a convienence script that is useful when Kiali is installed
```
jshaughn, line 6: convenience

jshaughn (approved):
\o/

PR #3849, opened 2021-04-07
Allow users to override add-on urls to check health for istio component status

### LINKED ISSUE #3631: Possibility to overwrite component status url
Prometheus component status is always showing **Unreachable** when using Victoria Metrics as prometheus source. Seems like it is just checking if prometheus url returns http code less than 400. Victoria Metrics returns 400, and therefore always showing **Unreachable**.

Would be nice to be able to define a custom check url instead of using the prometheus url.

No solution atm.

xeviknal:
fixes #3631 

This PR allows users to specify which URL should kiali ping to compute the health check for each add-on (jaeger, prom, grafana and custom dashboards).

There are a lot of different deployments for each add-on, so this give flexibility.

This PR needs **operator changes**: https://github.com/kiali/kiali-operator/pull/277

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -104,11 +104,6 @@ func (a *Auth) Obfuscate() {
 	a.CAFile = "xxx"
 }
 
-type AddonComponentStatus struct {
-	IsCore         bool   `yaml:"is_core,omitempty"`
-	HealthCheckUrl string `yaml:"health_check_url,omitempty"`
-}
-
 // PrometheusConfig describes configuration of the Prometheus component
 type PrometheusConfig struct {
 	Auth Auth `yaml:"auth,omitempty"`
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -104,6 +104,11 @@ func (a *Auth) Obfuscate() {
 	a.CAFile = "xxx"
 }
 
+type AddonComponentStatus struct {
```
lucasponce, line 107: Do you think is necessary to add a nest struct at this level ?
I mean, why not to keep the IsCore as the current place and add the HealthCheckUrl at that level ?
If present is used, if not the behaviour is like the previous approach.

I comment this, because, config is growing and growing and not sure if adding more nest elements add a little bit of complexity to read it.

hhovsepy (dismissed):
Verified the combinations of both params values.
Regression tested.
All are core, health_check_url is empty, pods are failing:
Screenshot from 2021-04-08 11-31-03
All are not core, health_check_url is wrong, pods are failing:
Screenshot from 2021-04-08 11-32-59
All are not core, health_check_url is wrong, pods are running.
Screenshot from 2021-04-08 11-35-50

xeviknal:
@lucasponce comments approached. Operator changes approached to.
@hhovsepy do you mind testing this last commit? Or even waiting for potential lucas subsequent comments.
thank you both

hhovsepy:
After the last commit, it seems 'health_check_url' is ignored.

lucasponce:
@xeviknal perhaps you missed to push the latest commit ?

xeviknal:
@lucasponce 's right. Push didn't make it. Commit is finally here and green light.
@hhovsepy test can be done know. sorry for the noise.

hhovsepy (approved):
Verified:
Screenshot from 2021-04-09 19-10-52

lucasponce (approved):

lucasponce:
If @skondkar can leave a comment, it's ready from my side as well.

skondkar:
Verified. LGTM

Screenshot from 2021-04-16 12-40-58

PR #3855, opened 2021-04-09
[molecule] set image version when testing non-default spec.version

jmazzitelli:
This is needed when we want to start testing different versions via the molecule tests.


jshaughn (approved):
LOK2ME

PR #3838, opened 2021-03-31
OpenID Azure AKS+AAD support

israel-hdez:
Azure requires:
* Usage of access_token rather than the id_token. This is because Azure doesn't use a real "openid" configuration. Instead, it uses webhooks to validate bearer tokens and check privileges.
* A target resource_id when redirecting the user for authentication. However, rather than adding a specific setting for this, an additional_request_params configuration is added to make it more generic.

Authentication with azure is going to work **only on the authorization code flow**.

Operator PR: https://github.com/kiali/kiali-operator/pull/289

Related #3294

PR #3898, opened 2021-04-16
Fix login failure with long OpenId tokens

### LINKED ISSUE #3874: OpenID integration issue with large tokens (caused by a long list of groups)
**Describe the bug**

With a Kiali instance deployed with the operator,  with openid integration (in my case keycloak) and groups being returned for RBAC purposes, the login process fails if the token being returned is larger than ~2894 characters  (encoded).   In my case, large tokens are returned because we have a lot of groups being returned and it could be common to reach  8K characters.

Ref: discussion https://github.com/kiali/kiali/discussions/3870

**Versions used**
Kiali: 1.32
Istio: 1.9.2
Kubernetes flavour and version: Kubeadm 1.19.2

**To Reproduce**
1. Deploy a Kiali instance with OpenID auth against a Keycloak endpoint
2. Make sure to return group membership information in the openid client so it is part of the token body
3. Create a test user in keycloak
4. Create a series of groups, assign them to the test user up to the point the array is ~1500 characters
5. Try to login.  Once authenticated, rather than getting the kiali UI you are redirected back to the Kiali login page.  No error logged.

Notes:
- See the ingress annotation below were needed allow the large token through nginx. 
- You can reproduce the problem with rbac disabled as long as the large token is returned

My kilia CR
```
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  namespace: istio-system
  name: kiali
  labels:
    helm.sh/chart: kiali-operator-1.32.0
    app: kiali-operator
    app.kubernetes.io/name: kiali-operator
    app.kubernetes.io/instance: kiali-operator
    version: "v1.32.0"
    app.kubernetes.io/version: "v1.32.0"
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/part-of: "kiali-operator"
annotations:
  ansible.sdk.operatorframework.io/verbosity: "1"
spec:
  api:
    namespaces:
      exclude:
      - "istio-operator"
      - "kube.*"
      - "default"
  auth:
    strategy: "openid"
    openid:
      client_id: "kiali-lab"
      issuer_uri: "https://keycloak-001.kd.io/auth/realms/kubernetes"
      insecure_skip_verify_tls: true
      username_claim: "email"
      disable_rbac: false
      scopes: ["openid", "groups", "email"]
  deployment:
    logger:
      log_level: debug
    accessible_namespaces:
    - '**'
    image_version: "v1.32.0"
    ingress_enabled: true
    override_ingress_yaml:
      metadata:
        annotations:
          kubernetes.io/ingress.class: my-nginx-ingress
          nginx.ingress.kubernetes.io/secure-backends: "false"
          nginx.ingress.kubernetes.io/backend-protocol: "HTTP"
          nginx.ingress.kubernetes.io/proxy-buffer-size: "64k"
          nginx.ingress.kubernetes.io/proxy-buffers-number: "8"
      spec:
        tls:
        - hosts:
          - kiali.my-nginx-ingress.lab01.kd.io
          secretName: kiali-tls
        rules:
        - host: kiali.my-nginx-ingress.lab01.kd.io
          http:
            paths:
            - path: /kiali
              backend:
                serviceName: kiali
                servicePort: 20001
```

**Expected behavior**
The JWT token size should not be a problem and allow Kiali UI login

israel-hdez:
In the authorization code flow, tokens returned by IdP can be long to the point that the session cookie also becomes long. Although RFCs don't set a maximum cookie size (see https://tools.ietf.org/html/rfc2965), in practice, major browsers use the RFC minimum limit as it's supported maximum - which is 4K. This means that longs tokens can cause the session token to reach this limit and be discarded by browsers.

This is overcoming these limitations by splitting session data in chunk no longer than 4K and setting multiple cookies, if needed.

The multi-cookie approach is applied only to the "authorization code" flow of OpenID. The "implicit flow" is left unchanged because IDPs tend to omit group membership data, which is the data that users have reported to generate larger tokens.

Fixes #3874

### DIFF
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -262,7 +262,7 @@ func GetOpenIdAesSession(r *http.Request) (*config.IanaClaims, error) {
 
 		// It's known that major browsers have a limit of 180 cookies per domain.
 		if numChunks <= 0 || numChunks > 180 {
-			return nil, fmt.Errorf("number of session cookies is %d, but limit is 180", numChunks)
+			return nil, fmt.Errorf("number of session cookies is %d, but limit is 1 through 180", numChunks)
 		}
 
 		// Read session data chunks and save into a buffer
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -30,7 +30,7 @@ const (
 	OpenIdNonceCookieName = config.TokenCookieName + "-openid-nonce"
 
 	// Maximum size of session cookies. This is 3.5K.
-	// RFCs tell that the max cookie size is 4K, but this includes
+	// Major browsers limit cookie size to 4K, but this includes
 	// metadata like expiration date, the cookie name, etc. So
 	// use 3.5K for cookie data and leave 0.5K for metadata.
 	SessionCookieMaxSize = 3584
@@ -261,7 +261,7 @@ func GetOpenIdAesSession(r *http.Request) (*config.IanaClaims, error) {
 		}
 
 		// It's known that major browsers have a limit of 180 cookies per domain.
-		if numChunks > 180 {
+		if numChunks <= 0 || numChunks > 180 {
 			return nil, fmt.Errorf("number of session cookies is %d, but limit is 180", numChunks)
 		}
 
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -12,6 +12,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"strconv"
 	"strings"
 	"time"
 
@@ -27,6 +28,12 @@ import (
 
 const (
 	OpenIdNonceCookieName = config.TokenCookieName + "-openid-nonce"
+
+	// Maximum size of session cookies. This is 3.5K.
+	// RFCs tell that the max cookie size is 4K, but this includes
+	// metadata like expiration date, the cookie name, etc. So
+	// use 3.5K for cookie data and leave 0.5K for metadata.
+	SessionCookieMaxSize = 3584
 )
 
 type OpenIdMetadata struct {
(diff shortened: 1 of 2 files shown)


handlers/authentication.go
```diff
@@ -924,15 +949,46 @@ func OpenIdCodeFlowHandler(w http.ResponseWriter, r *http.Request) bool {
...
+	base64SessionData := base64.StdEncoding.EncodeToString(cipherSessionData)
+
+	// If resulting session data is large, it may not fit in one cookie. So, the resulting
+	// session data is broken in chunks and multiple cookies are used, as is needed.
+	sessionDataChunks := chunkString(base64SessionData, business.SessionCookieMaxSize)
+	for i, chunk := range sessionDataChunks {
+		var cookieName string
+		if i == 0 {
+			// Set a cookie with the regular cookie name with the first chunk of session data.
+			// This is for backwards compatibility
```
jmazzitelli, line 961: It's also nice that in most use-cases, you still only ever have one cookie as opposed to also needing a `-chunks` cookie with a value of 1.

jmazzitelli: or do you always still have a `-chunks` cookie? I think this implementation does. Oh, well, never mind :)

jmazzitelli (approved):
I looked at the code and it all looks correct to me. I don't have an openid setup so I didn't run this.

israel-hdez:
@somejfn I wonder if you could try this image and check that the issue with Keycloak is gone: quay.io/edgarhz/kiali:rc-20210416-1131

@kulbhusanazad perhaps can you also try that image ^^? It includes PR #3838 which are the changes to support Azure. In the operator's KialiCR you should be able to configure these new options:

```yaml
spec:
  auth:
    openid:
      api_token: "access_token"
      additional_request_params:
        resource: {your_resource_id}
```

...and all your other configs should be the same.

---

If you can confirm this works, this can go in today's Kiali release. But try ASAP.

handlers/authentication.go
israel-hdez: It's always there!
But it would be a simple condition to add around these lines: https://github.com/kiali/kiali/pull/3898/files#diff-3dd5896ffb3bb3685544e6a40338180ed95a5864b929b908fd1a70ec7628b251R983-R991

I can put that condition if you prefer not having the `-chunks` with value "1" :-)

jmazzitelli: You can leave it alone. "No biggie"

israel-hdez:
@jmazzitelli Ouch! I forgot the code to properly do "logout".
I added the code to a new commit. So, I think this deserves a follow-up review from you.

Also, I added the conditional to omit the "-chunks" cookie if there is only one chunk to keep number of cookies to a minimum ;-)

business/openid_auth.go
```diff
@@ -238,7 +245,49 @@ func GetOpenIdAesSession(r *http.Request) (*config.IanaClaims, error) {
...
+	numChunksCookie, chunksCookieErr := r.Cookie(config.TokenCookieName + "-chunks")
+	if chunksCookieErr == nil {
+		numChunks, convErr := strconv.Atoi(numChunksCookie.Value)
+		if convErr != nil {
+			return nil, convErr
+		}
+
+		// It's known that major browsers have a limit of 180 cookies per domain.
+		if numChunks <= 0 || numChunks > 180 {
+			return nil, fmt.Errorf("number of session cookies is %d, but limit is 180", numChunks)
```
jmazzitelli, line 265:
```suggestion
			return nil, fmt.Errorf("number of session cookies is %d, but limit is 1 through 180", numChunks)
```

jmazzitelli, line 265:
```suggestion
			return nil, fmt.Errorf("number of session cookies is %d, but limit is 1 through 180", numChunks)
```

jmazzitelli (approved):
I'll approve with that minor comment/suggestion, in case you want to do that.
Again, I didn't run, but I see what's going on and the code looks ok.

My only question is that in the code I see we continue even if a -aes cookie is missing (`log.Warningf("Apparently, there is no AES session: %s ", aes.Error())` but yet the deleteTokenCookies function always drops it. I think this is OK (you are just telling the client to drop a cookie that might not exist - you would think that would be a no-op) but I bring it up just in case.

israel-hdez:
> My only question is that in the code I see we continue even if a -aes cookie is missing (`log.Warningf("Apparently, there is no AES session: %s ", aes.Error())` but yet the deleteTokenCookies function always drops it. I think this is OK (you are just telling the client to drop a cookie that might not exist - you would think that would be a no-op) but I bring it up just in case.

@jmazzitelli I'm not sure if I understand your question, but I'll answer this:

The `AuthenticationInfo` func where this `log.Warning` located, is expected to be called by the front-end to check if there is an active session. If there is a session, the front-end expects some user details (e.g. his display name) and it will show the dashboard rather than the login page. Then, if there is no session, the back-end sends empty user info and the front-end requests a login. So, we want to gracefully fail here.

In the `deleteTokenCookies`, yes it's a no-op. Indeed, it checks for cookie presence to send the deletion. See: https://github.com/kiali/kiali/pull/3898/files#diff-3dd5896ffb3bb3685544e6a40338180ed95a5864b929b908fd1a70ec7628b251R1028

jmazzitelli (approved):

PR #3902, opened 2021-04-16
need to allow ad hoc namespace and images when running the playbook target

jmazzitelli:
Needed for devs that run the operator playbook via make, like this:

`make -e OPERATOR_PROFILER_ENABLED=true -e CLUSTER_TYPE=minikube run-operator-playbook `


jshaughn (approved):
fine

Issue #1404, opened 2019-08-23
Enhance visualization of label validation


lucasponce:
This JIRA is to improve the feature that indicates validation of labels presence.
Presence of labels app, version is a recommended practice for Istio:
https://istio.io/docs/setup/kubernetes/spec-requirements/
See attached screenshot from Kiali
We have some support of this, but it should be enhanced and refined.
When I read "label validation: app, version" it is not necessarily meaning "validating for presence of labels named app, version, they are both present, or 1 is missing". It could also potentially mean: "I've found 2 labels named X, Y, lets validate their content for Z based on some rule". And when there is a row with no statement at all for "label validation", it is impossible to understand that something is missing.
IMO it should be like with missing sidecar situation (after the redesign) - when it is missing, it says it's missing. and here's with the labels it's a mixed situation with no clear indication what is wrong.
For jaeger pod for instance there's only the app label, so it says "Label validation: app" - might be "validated the content of label app" (and the results/implications are nowhere to be seen).

JIRA:
https://issues.jboss.org/browse/KIALI-2143

lucasponce:
This issue is valid and Kiali may benefit of a minor change.

As it was introduced in other areas, some features may benefit with an "(i)" icon with additional information about the context.

The Label validation is one that may show a similar icon where user could understand what that means or what side effects may have a missing sidecar or a missing "app/version" label.
![image](https://user-images.githubusercontent.com/1662329/107936402-ebf3ff00-6f82-11eb-9957-cfd49804b0a1.png)

lucasponce:
Discussed in the PR.
This issue is a little bit outdated and the strategy will be consolidate the backend validations in the standard business logic, instead to propose a only validation in the UI.

But the UI proposal in https://github.com/kiali/kiali-ui/pull/2128 is good and it's aligend with https://github.com/kiali/kiali/issues/3958 so, I'm going to cherrypick those changes and adjust them as part of this issue.

PR #3906, opened 2021-04-19
Enhance visualization of label validation

aljesusg:
Fix https://github.com/kiali/kiali/issues/1404
Ui https://github.com/kiali/kiali-ui/pull/2128

Configuration require example

```yaml
label_validation:
  - namespace: "alpha"
    presence:
      - hello
    filter_label:
      app: "avion"
      version: "v1"
```
The configuration like health let granularity like namespace, kind and name by regex.
After that we have 2 checks:

- Presence of label (array of strings)
- filter of label (map key => regex)

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -229,10 +229,9 @@ type LabelValidation struct {
 
 // IstioLabels holds configuration about the labels required by Istio
 type IstioLabels struct {
-	AppLabelName       string            `yaml:"app_label_name,omitempty" json:"appLabelName"`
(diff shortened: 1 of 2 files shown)


aljesusg:
Would it make sense to put it inside the health_config? In some way is related with warning/errors what do you think?

lucasponce:
I like the UI proposal of this work, but the backend should be consolidated as a standard validation.
This work is addressing the original issue, but I admit that as it is written it may confuse how to do that.

So, I'll close this one in favor to consolidate this validation in the future as a standard validation.

PR #3909, opened 2021-04-19
Speed up CI tests by passing env var to make commands

nrfox:
Passing the `IS_MAISTRA=false` env var to the make commands skips a part of the makefile that is taking awhile to timeout when a local cluster environment (e.g. minikube) exists but is not accessible.


nrfox:
Closing in favor of #3954

Issue #3923, opened 2021-04-21
The istio-sidecar-injector config map name should be configurable


bougyman:
**Describe the bug**

**Versions used**
Kiali:
Istio:
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9)

**To Reproduce**
1. Install istio with a revision: tag set to 1-9
2. Deploy kiali with external_services.istio.deployment_name as istiod-1-9 and external_services.istio.config_map_name as istio-1-9
3. Check logs on the deployed kiali instance
4. You will see the error: 2021-04-21T15:55:03Z WRN Cannot resolve the network ID of the cluster where Kiali is hosted: cannot get the sidecar injector config map :configmaps "istio-sidecar-injector"     not found

**Expected behavior**
The config map for istio-sidecar-injector would be configurable to find the istio-sidecar-injector-1-9 config map

israel-hdez:
Closing, because you already helped us with this issue :-)

lir2002:
Hope everyone can find the answer here:
https://kiali.io/docs/configuration/istio/#sidecar-injection-canary-upgrade-management-and-istio-revisions

PR #3925, opened 2021-04-21
Adds configuration for the istio-sidecar-injector configmap name

bougyman:
** Describe the change **

This change addresses #3923 by adding a new configuration attribute for istio_sidecar_injector_config_map_name

** Issue reference **

#3923 

** Backwards incompatible? **

Should be backwards compatible

** Documentation **

Updated kiali_cr.yaml in https://github.com/kiali/kiali-operator/pull/295

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -341,7 +341,7 @@ func (in *MeshService) findRemoteKiali(clusterName string, kubeconfig *kubernete
 func (in *MeshService) resolveKialiNetwork() (string, error) {
 	conf := config.Get()
 
-	istioSidecarConfig, err := in.k8s.GetConfigMap(conf.IstioNamespace, "istio-sidecar-injector")
+	istioSidecarConfig, err := in.k8s.GetConfigMap(conf.IstioNamespace, conf.ExternalServices.Istio.IstioInjectorName)
 	if err != nil {
 		// Don't return an error, as this may mean that Kiali is not installed along the control plane.
 		// This setup is OK, it's just that it's not within our multi-cluster assumptions.
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -341,7 +341,7 @@ func (in *MeshService) findRemoteKiali(clusterName string, kubeconfig *kubernete
 func (in *MeshService) resolveKialiNetwork() (string, error) {
 	conf := config.Get()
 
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -171,6 +171,7 @@ type IstioConfig struct {
 	EnvoyAdminLocalPort      int               `yaml:"envoy_admin_local_port,omitempty"`
 	IstioIdentityDomain      string            `yaml:"istio_identity_domain,omitempty"`
 	IstioInjectionAnnotation string            `yaml:"istio_injection_annotation,omitempty"`
+	IstioInjectorName        string            `yaml:"istio_injector_name,omitempty"`
```
jmazzitelli, line 174: I think this should be called "IstioInjectorConfigMapName" (istio_injector_config_map_name) since that is what it is. Just saying "injector name" is confusing because the name gives no indication it is a config map name. Clarity is a good thing, even if it increases verbosity slightly.

(side note: we don't have to do this now, but "ConfigMapName" should probably be called IstiodConfigMapName to be more clear. But again no need to do that now because that will affect user-facing Kiali CRs).

jmazzitelli (changes requested):

business/mesh.go
```diff
@@ -341,7 +341,7 @@ func (in *MeshService) findRemoteKiali(clusterName string, kubeconfig *kubernete
 func (in *MeshService) resolveKialiNetwork() (string, error) {
 	conf := config.Get()
 
-	istioSidecarConfig, err := in.k8s.GetConfigMap(conf.IstioNamespace, "istio-sidecar-injector")
+	istioSidecarConfig, err := in.k8s.GetConfigMap(conf.IstioNamespace, conf.ExternalServices.Istio.IstioSidecarInjectorConfigBapName)
```
jmazzitelli, line 344: Bap != Map :)
```suggestion
	istioSidecarConfig, err := in.k8s.GetConfigMap(conf.IstioNamespace, conf.ExternalServices.Istio.IstioSidecarInjectorConfigMapName)
```

jmazzitelli (changes requested):

jmazzitelli (approved):
this looks correct, but I did not run it. Perhaps someone like @israel-hdez or @jshaughn can also review.

israel-hdez:
I tried this with the usual name and also changing to a non-existent name.
I do see a warning in the logs when using that random string. I think that means this is fine.

israel-hdez (approved):

israel-hdez:
Thanks @bougyman !

PR #3926, opened 2021-04-21
better support for machines with podman

jmazzitelli:
some changes to make it easier to use the hack scripts for some environments.


jshaughn (approved):
LOK

PR #3935, opened 2021-04-23
Ensure idle nodes are created in the homecluster

### LINKED ISSUE #3917: Idle nodes are not synced with traffic nodes
This looks very confusing to me:

image

It seems that "idle" nodes (generated from kubernetes definitions) and "traffic" nodes are not joined in the graph.

I understand that there should be some kind of "sync" here (if an idle node is present as a traffic node, don't show).

Otherwise it creates a lot of noise and in examples as showed it makes the graph hard to use.

This is the business scenario I'm running: in a moment of the time, I have a workload that is out of the graph (i.e. missing sidecar or simply scaled down/not ready), I'd expect I can see those in the graph combined with the existing one.

nrfox:
Adds idle nodes to the graph in the homecluster instead of "unknown".

Behavior before the change:

Screenshot from 2021-04-23 13-46-10

Behavior after the change:

Screenshot from 2021-04-23 13-39-41

Fixes #3917

### DIFF
--- a/graph/telemetry/istio/appender/idle_node.go
+++ b/graph/telemetry/istio/appender/idle_node.go
@@ -31,7 +31,14 @@ func (a IdleNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *g
 
 	services := []models.ServiceDetails{}
 	workloads := []models.WorkloadListItem{}
-	clusterName := globalInfo.HomeCluster // This can be empty if another appender hasn't populated this yet.
+	if globalInfo.HomeCluster == "" {
+		globalInfo.HomeCluster = "unknown"
+		c, err := globalInfo.Business.Mesh.ResolveKialiControlPlaneCluster(nil)
+		graph.CheckError(err)
+		if c != nil {
+			globalInfo.HomeCluster = c.Name
+		}
+	}
 
 	if a.GraphType != graph.GraphTypeService {
 		workloads = getWorkloadList(namespaceInfo.Namespace, globalInfo).Workloads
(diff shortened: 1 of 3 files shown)


mattmahoneyrh (dismissed):
Fix looks good:

image

graph/telemetry/istio/appender/idle_node.go
```diff
@@ -31,7 +31,7 @@ func (a IdleNodeAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *g
 
 	services := []models.ServiceDetails{}
 	workloads := []models.WorkloadListItem{}
-	clusterName := graph.Unknown
+	clusterName := globalInfo.HomeCluster // This can be empty if another appender hasn't populated this yet.
```
jshaughn, line 34: Don't worry about just setting globalInfo.HomeCluster immediately.   Look at dead_node.go#AppendGraph and take that same code an plunk it here.  Then you can assume it's set.

jshaughn (changes requested):
@nrfox Good job with the fix, and glad to see test code.  I have one requested change for consistency, in general for the appender we checkfor/set the required global or namespace info immediately so that the main logic can assume it's there.

graph/telemetry/istio/appender/idle_node.go
nrfox: The dead node appender simply ignores errors if they happen: https://github.com/kiali/kiali/blob/master/graph/telemetry/istio/appender/dead_node.go#L31-L36 whereas the idle node implementation panics if any errors occur: https://github.com/kiali/kiali/blob/master/graph/telemetry/istio/appender/idle_node.go#L45. Which is the correct way we want to handle errors in this case?

jshaughn: hmm, good catch, not sure why dead_node is ignoring an error there.  It should be using graph.CheckError(err).  Feel free to fix that as part of this PR.

nrfox: I tried [adding a test for the check/panic](https://github.com/nrfox/kiali/compare/kiali-3917...nrfox:debug-test?expand=1#diff-5a6963c484616b4837918ea3b52565f7ca434a7aa00d4f3a94be6965f31b3836R211-R229) and got pretty close but if another test ran before that one the cached value of cluster info would get returned causing the test to pass/fail based on the order of the test runs. I couldn't figure out a way to edit that global cache value so if you've got any suggestions there I'm happy to add the test back in.

jshaughn: don't worry about the panic test.

jshaughn (approved):
Changes look good!

nrfox:
@mattmahoneyrh you mind taking another look when you get a chance?

mattmahoneyrh (approved):

Issue #3939, opened 2021-04-23
Service tracing tab display options resulting in error 


brendarearden:
**Describe the bug**
When attempting to change the Display value on a service's tracing page, an error occurs saying `cannot not parse minDuration: time: missing unit duration "49265"` After this error shows on the page, the the display dropdown becomes disabled. Sometimes the error stays until you re-deploy the pod, sometimes the error corrects itself by navigating to a different serve and then back.

I am able to see the tracing information if I do not touch the display drop down at all. 

![Screen Shot 2021-04-23 at 4 50 36 PM](https://user-images.githubusercontent.com/19455540/115940029-f801ae80-a454-11eb-9e90-ead0467d780b.png)

**Versions used**
Kiali: v1.32.0
Istio: v1.9.3
Kubernetes flavour and version: (e.g. OpenShift Origin 3.9): k3d

**To Reproduce**
1. Install Istio v1.9.3 by installing the latest istioctl and running `istioctl install --set profile=demo
3. Install jaeger using the https://istio.io/latest/docs/ops/integrations/jaeger/#option-1-quick-start instructions - or deploy the jaeger all in one docker container https://www.jaegertracing.io/docs/1.22/getting-started/#all-in-one 
2. Use Helm 3 to install Kiali v1.32.0 kiali-server helm chart from https://github.com/kiali/helm-charts/tree/master/kiali-server 
  - when installing kiali, configure the external_services.tracing.in_cluster_url=<your in_cluster jaeger url>. Do not set the url, or any auth stuff: 
 Example: 
```
  external_services:
    tracing:
      in_cluster_url: "http://tracing.istio-system.svc:16686/jaeger"
```
3. Deploy a sample application that has several services that you can create traffic to view in kiali (I used the istio bookinfo application https://istio.io/latest/docs/examples/bookinfo/#start-the-application-services) 
4. After generating enough traffic to be picked up by kiali, navigate to a service  and then click on the tracing tab
5. Click on the display drop down and select a different display percentile. 
6. See error

**Expected behavior**
The display percentile call should work or should fail gracefully if the data is not available. 

**Additional Notes**
https://github.com/kiali/kiali/blob/f548b562303ebd7d4dc82ec3f24ae1336b5bccd8/jaeger/traces.go#L53 << it does look like you have recently made changes to functionality that is currently erroring 

[...]

PR #3950, opened 2021-04-27
Fix missing units in jaeger query

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3939

cc @brendarearden 

Kiali is using GRPC and HTTP clients to connect Jaeger.
It keeps both because some systems may use an OAuth sidecar to protect Jaeger and we've found that GRPC + OAuth sidecars may introduce some problems.

So, in short, there is a bug in the legacy HTTP query that it was missing a unit in the backend code when present a parameter.

How to test:
- Navigate into the traces tab in any of the demos examples we use with traces enabled and set the percentile options.
- Validate that no errors are found and Traces tab is stable.

image

### DIFF
--- a/jaeger/http_client.go
+++ b/jaeger/http_client.go
@@ -53,7 +53,14 @@ func getTraceDetailHTTP(client http.Client, endpoint *url.URL, traceID string) (
 }
 
 func queryTracesHTTP(client http.Client, u *url.URL) (*JaegerResponse, error) {
-	resp, code, reqError := makeRequest(client, u.String(), nil)
+	// HTTP and GRPC requests co-exist, but when minDuration is present, for HTTP it requires a unit (ms)
+	// https://github.com/kiali/kiali/issues/3939
+	if u.Query().Get("minDuration") != "" {
+		query := u.Query()
+		query.Set("minDuration", u.Query().Get("minDuration") + "ms")
+		u.RawQuery = query.Encode()
+	}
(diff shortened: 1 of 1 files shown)


aljesusg (approved):
lgfm

jaeger/http_client.go
```diff
@@ -53,6 +53,13 @@ func getTraceDetailHTTP(client http.Client, endpoint *url.URL, traceID string) (
 }
 
 func queryTracesHTTP(client http.Client, u *url.URL) (*JaegerResponse, error) {
+	// HTTP and GRPC requests co-exist, but when minDuration is present, for HTTP it requires a unit (ms)
+	// https://github.com/kiali/kiali/issues/3939
+	if u.Query().Get("minDuration") != "" {
```
hhovsepy, line 58: For the safe side, isn't it better to check strings.HasSuffix(u.Query().Get("minDuration"), "ms") ?

lucasponce: Right ! Added a fix

hhovsepy (approved):
LGTM

prachiyadav (commented):
steps followed:

deployed bookinfo with traces
navigate appliation > details, select the traces tab
select one trace
 changed the display options, to filter per percentile
Screenshot from 2021-04-29 21-56-42

prachiyadav (approved):

Issue #3830, opened 2021-03-29
Workload_health_replica_set endpoint error


pbajjuri20:
Workload_health_replica_set endpoint:

**query  param** 
rateInterval: 60s
type: ReplicaSet

https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/kiali-traffic-generator/health?rateInterval=60s&type=ReplicaSet

bug:

Replacing the **type: ReplicaSet with type: invalid**

**query  param** 
rateInterval: 60s
type: invalid

https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/kiali-traffic-generator/health?rateInterval=60s&**type=invalid**

Instead of an error, It is getting a **200 Status code**.

pbajjuri20:
Here is the list of workload_types:

![List of workload types](https://user-images.githubusercontent.com/55099523/112909933-5d65b800-90c0-11eb-944a-fdecb5f3279a.png)

lucasponce:
I don't think this is 100% a bug.

if you have this query:

```
https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/kiali-traffic-generator/health?rateInterval=60s&type=invalid
```

Kiali will find a workload of name "kiali-traffic-generator" in all types of workloads.

Could you confirm if is that the case for this test ?

PR #3953, opened 2021-04-27
Fix invalid workload types

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3830

This fix will provide better support for custom controllers and invalid types.

QE:
- Test a invalid/valid query:

```
GET http://admin:admin@localhost:8000/api/namespaces/default/workloads/reviews-v3/health?rateInterval=60s&type=invalid
```
404 expected

```
GET http://admin:admin@localhost:8000/api/namespaces/default/workloads/reviews-v3/health?rateInterval=60s&type=Deployment
```

200 expected

- Regression testing for custom controllers:

https://github.com/argoproj/argo-rollouts
https://github.com/argoproj/argo-rollouts/blob/master/docs/getting-started.md

Check that RollOut controllers are seeing in Kiali as Workloads without any crash

cc @hhovsepy I'm starting to adding you on PR as Dev role :) just with the aim to get you familiar with the code.

Note, the workloads logic is more complex as expected as it has to deal with:
- Controllers with no pods.
- Workloads that are created from a controllers A (ReplicaSet) that is also created from controller B (Deployment) (two levels).
- Custom controllers; pods created by a controller A (ReplicaSet) that is also created from a controller C (not known).
- Invalid situations.

So, those are some reason why the business logic have several edges cases (no to get in detail into this PR but just to share knowledge and context about why it has that design).

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -1302,7 +1302,7 @@ func fetchWorkload(layer *Layer, namespace string, workloadName string, workload
 		cnFound := true
 		switch ctype {
 		case kubernetes.DeploymentType:
-			if dep.Name == workloadName {
+			if dep != nil && dep.Name == workloadName {
 				selector := labels.Set(dep.Spec.Template.Labels).AsSelector()
 				w.SetPods(kubernetes.FilterPodsForSelector(selector, pods))
 				w.ParseDeployment(dep)
@@ -1347,7 +1347,7 @@ func fetchWorkload(layer *Layer, namespace string, workloadName string, workload
 				cnFound = false
 			}
 		case kubernetes.DeploymentConfigType:
-			if depcon.Name == workloadName {
+			if depcon != nil && depcon.Name == workloadName {
 				selector := labels.Set(depcon.Spec.Template.Labels).AsSelector()
 				w.SetPods(kubernetes.FilterPodsForSelector(selector, pods))
 				w.ParseDeploymentConfig(depcon)
@@ -1356,7 +1356,7 @@ func fetchWorkload(layer *Layer, namespace string, workloadName string, workload
 				cnFound = false
 			}
 		case kubernetes.StatefulSetType:
-			if fulset.Name == workloadName {
+			if fulset != nil && fulset.Name == workloadName {
 				selector := labels.Set(fulset.Spec.Template.Labels).AsSelector()
 				w.SetPods(kubernetes.FilterPodsForSelector(selector, pods))
 				w.ParseStatefulSet(fulset)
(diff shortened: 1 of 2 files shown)


pbajjuri20:
@lucasponce :

type: Deployment

200

`https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/reviews-v3/health?rateInterval=60s&type=Deployment
`

before PR:

502 Bad Gateway
The server returned an invalid or incomplete response.

`https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/reviews-v3/health?rateInterval=60s&type=invalid
`

after PR: 

404 http error 
{"error":"Workload.Kiali \"reviews-v3\" not found"}

`https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/reviews-v3/health?rateInterval=60s&type=invalid
`

lucasponce:
ok @pbajjuri20 I've added an edge case and better test coverage for this corner case.

Summary:
- If the type parameter is not known, Kiali will consider it as a custom controller, and it will try to resolve it from the pods, not querying the main controller (even if that exists).
- Health endpoints shouldn't fail, but it may present some limitation (i.e. controller information may not be available).

But considering this is a corner case I think is a valid tradeoff.

hhovsepy (approved):
The solution looks good to me.

pbajjuri20:
@lucasponce 

after new commits in the code:

type = invalid

202 status code

https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/reviews-v3/health?rateInterval=60s&type=invalid

expected: 404 HTTP error (before commit)

lucasponce:
Hi @pbajjuri20,

It's a tricky corner case, but I think the current behaviour is the desired one.

In the cluster we have a reviews-v3 workload as a Deployment, so the main API

```
https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/reviews-v3/health?rateInterval=60s
```

will work because the identifier for the workload is the main path `/workloads/reviews-v3` -> that will tell Kiali "try to find a controller called reviews-v3 (from pods and from known controllers)".

As discussed, and due Kiali API is only used for Kiali UI, there is an optional helper parameter called "type" that will help Kiali to not search on unnecessary controllers:

```
https:/{kiali.hostname.com}/api/namespaces/bookinfo/workloads/reviews-v3/health?rateInterval=60s&type=invalid
```

But if `workloads/reviews-v3` exists but `type=invalid` doesn't match, Kiali should try to find what controller it finds.

On top of that, we need to deal with custom controllers (alternative "Deployments" objects provided by third parties, like frameworks as commented).

I added more support for this case in the tests:

https://github.com/kiali/kiali/pull/3953/files#diff-08f937f38346ee39a854081b00b7e7423c61bb09425a630c9535435fee31fe9bR450

But in short, if the workload exists, Kiali will try to find it, and it will ignore the type parameter.

Hope this helps.

pbajjuri20 (approved):

PR #3954, opened 2021-04-27
Adds a check to the makefile to see if OC is ready

### LINKED ISSUE #3910: Makefile env var assignment hangs
When setting up global environment variables, the [makefile](https://github.com/kiali/kiali/blob/a810da82c812c053cb46371b1e12f7ed7941e84c/Makefile#L132) tries to determine the value of `IS_MAISTRA` by making a call to the local cluster environment (minikube, openshift, etc). If you have minikube installed but it's not accessible, this command will block until a timeout is reached and the make command can then proceed. Since these environment variables are evaluated each time make is run, this adds additional time to make commands by locally and in the CI runner. You can reproduce this locally by running `minikube stop` then `make test`.

A couple options for refactoring are:
1. Moving the `IS_MAISTRA` env var to the one target where it's used. Maybe the initial conditional assignment can be something like `IS_MAISTRA ?= unknown` and the check is performed in the `kiali-create` target.
2. Keep the env var where it is and set a short timeout for the call.

nrfox:
Adds an additional env var to the makefile to check if the cluster is ready. Useful if other env vars need to talk to the cluster to determine their value and might otherwise timeout such as with `IS_MAISTRA`.

Should also help speed up unit tests run in CI where there isn't a working cluster.

Fixes #3910

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,7 +95,7 @@ KIND_NAME ?= kind
 
 # Determine if the OC is operational or not. Useful for other commands that might timeout if the OC exists but is not responding.
 ifeq ($(CLUSTER_TYPE),minikube)
-OC_READY ?= $(shell if minikube -p ${MINIKUBE_PROFILE} status &>/dev/null ; then echo "true" ; else echo "false" ; fi)
+OC_READY ?= $(shell if ${MINIKUBE} -p ${MINIKUBE_PROFILE} status &>/dev/null ; then echo "true" ; else echo "false" ; fi)
 else ifeq ($(CLUSTER_TYPE),kind)
 OC_READY ?= $(shell if ${OC} cluster-info --context=kind-${KIND_NAME} --request-timeout=1s &>/dev/null ; then echo "true" ; else echo "false" ; fi)
 else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -93,6 +93,15 @@ MINIKUBE_PROFILE ?= minikube
 KIND ?= $(shell which kind 2>/dev/null || echo "MISSING-KIND-FROM-PATH")
 KIND_NAME ?= kind
 
+# Determine if the OC is operational or not. Useful for other commands that might timeout if the OC exists but is not responding.
+ifeq ($(CLUSTER_TYPE),minikube)
+OC_READY ?= $(shell if minikube -p ${MINIKUBE_PROFILE} status &>/dev/null ; then echo "true" ; else echo "false" ; fi)
```
jmazzitelli, line 98: We have a ${MINIKUBE} variable for this - https://github.com/kiali/kiali/blob/master/Makefile#L89

In some contexts you can't rely it being on path - hence, some times we need to pass in `MINIKUBE=/full/path/here`

```suggestion
OC_READY ?= $(shell if ${MINIKUBE} -p ${MINIKUBE_PROFILE} status &>/dev/null ; then echo "true" ; else echo "false" ; fi)
```

jmazzitelli (changes requested):

jmazzitelli (approved):
LGTM

Issue #3948, opened 2021-04-27
Research Metrics tab main layout


lucasponce:
As part of #3817, we are revisiting the details tabs to improve the visualization of the information.

Today the metrics tab offers a fixed height that forces the user to scroll down to see all the charts, also the user may not know that there are more metrics showed.

This issue is to experiment with a different layout for the main view of the metrics charts, we have discussed that probably user may benefit of "smaller" charts that can be expanded for more details.

In short the scope of this work should be:
- Propose and test different layouts where all metrics are available for the user, when user maximize or expand a particular chart, previous behaviour is still working.
- Minor align in terms of buttons with the logs tab (in particular, same button for "expand" should be used, also the logs tab proposes a collapse mode of the tab options that may be applied here as well).
- Some default options can be hidden by default (in particular percentiles 9x) and let user select them on chart expanded view.
- This is not a change in the metrics showed or a change in the feature, it's a work in terms of visualization with the goal that the user knows all metrics available and may have information about "which chart I need to pay attention".

Goal is to have some PoC that we can iterate together with UX and the team.

lucasponce:
cc @lwrigh 

As commented, this is one area that will be in the UX meetings when it has more traction, at the moment is in an early research phase where we are studying the options.

lucasponce:
Issue https://github.com/kiali/kiali/issues/4087 will be included as part of the follow-up work pending for this effort.

lucasponce:
Closed by https://github.com/kiali/kiali/pull/4091 and https://github.com/kiali/kiali-ui/pull/2176

PR #3956, opened 2021-04-28
Research Metrics tab main layout

aljesusg:
Relates to https://github.com/kiali/kiali/issues/3948


lucasponce (approved):

lucasponce:
As discussed, we are going to consolidate this work.
Runtimes are not yet addressed but there will a folloup work also in sync with the CRD refactoring that is on the way.

PR #3961, opened 2021-04-28
Create service entry nodes in homecluster

### LINKED ISSUE #3938: Service entries should be assigned to the same cluster as the relevant service
This originated in but really isn't related to #3879.  @lucasponce had this comment:

... I see this weird:
image

I understand it, as I created that example, and I see that ServiceEntry belongs to that namespace because is where the definition lives, but to move it to an unknown cluster sounds weird. I understand that is because it comes from telemetry, but I wonder if there is a different possibility of wording.

My reply and the source of this issue:

@lucasponce Service Entry handling is a constant work in progress. Looking at this I agree with you, a service entry should not have an unknown cluster as it's supposed to be associated with a service, which itself is in a cluster. It may be exported to 1..N namespaces, but those namespaces should all be within a cluster. We need to figure out if why this is "unknown", it could be a problem in Istio as part of the SE definitions returned from k8s, or it could be a problem in Kiali making a proper assignment of the cluster. I'd probably lean towards the latter case and maybe a missed case in the recent cluster work on our side.

nrfox:
Fixes #3938 

Creates Service Entry nodes in the homecluster instead of `unknown`.

Before change:
Screenshot from 2021-04-28 16-46-15

After:
Screenshot from 2021-04-28 17-10-08

### DIFF
--- a/graph/telemetry/istio/appender/service_entry.go
+++ b/graph/telemetry/istio/appender/service_entry.go
@@ -53,6 +53,15 @@ func (a ServiceEntryAppender) AppendGraph(trafficMap graph.TrafficMap, globalInf
 		return
 	}
 
+	if globalInfo.HomeCluster == "" {
+		globalInfo.HomeCluster = "unknown"
+		c, err := globalInfo.Business.Mesh.ResolveKialiControlPlaneCluster(nil)
+		graph.CheckError(err)
+		if c != nil {
+			globalInfo.HomeCluster = c.Name
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
Looks good to me.  As discussed in the issue, at some point we may have to look at making this more robust, but for now this solution should suffice.  Adding @lucasponce as a reviewer since he was the original source of the issue.

lucasponce (approved):
image

mattmahoneyrh (approved):

Issue #3964, opened 2021-04-29
fatal error: concurrent map read and map write on IsNamespaceCached


juliantaylor:
in our larger clusters kiali crashes constantly with:

```
2021-04-29T12:20:24Z INF Kiali: Version: v1.32.0, Commit: 166ffcc9edfc6fdd8c6ce8e74eb5670414b949f5

2021-04-29T12:20:24Z INF Using authentication strategy [anonymous]
2021-04-29T12:20:24Z WRN Kiali auth strategy is configured for anonymous access - users will not be authenticated.
2021-04-29T12:20:24Z INF Kiali: Console version: 1.32.0
2021-04-29T12:20:24Z INF Generating env.js from config
2021-04-29T12:20:24Z INF Server endpoint will start at [:20001/]
2021-04-29T12:20:24Z INF Server endpoint will serve static content from [/opt/kiali/console]
2021-04-29T12:20:24Z INF Starting Metrics Server on [:9090]
2021-04-29T12:20:40Z INF Kiali Cache is active for namespaces [.*]
2021-04-29T12:20:40Z INF [Prom Cache] Enabled

2021-04-29T12:20:40Z INF Waiting for Kiali cache for [namespace: mynamespace] to sync
2021-04-29T12:20:41Z INF Kiali cache for [namespace: mynamespace] started
fatal error: concurrent map read and map write

goroutine 2377 [running]:
runtime.throw(0x1b0b984, 0x21)
        /usr/local/go/src/runtime/panic.go:1117 +0x72 fp=0xc003848e08 sp=0xc003848dd8 pc=0x437072
runtime.mapaccess2_faststr(0x18ae760, 0xc00061c210, 0xc002275d64, 0x6, 0x0, 0x0)
        /usr/local/go/src/runtime/map_faststr.go:116 +0x4a5 fp=0xc003848e78 sp=0xc003848e08 pc=0x414265
github.com/kiali/kiali/kubernetes/cache.(*kialiCacheImpl).CheckNamespace(0xc000269a40, 0xc002275d64, 0x6, 0x0)
        /home/fedora/workspace/kiali-workflow/kiali-release/src/github.com/kiali/kiali/kubernetes/cache/cache.go:200 +0x9b fp=0xc003848ed0 sp=0xc003848e78 pc=0x165287b
github.com/kiali/kiali/business.IsNamespaceCached(...)
[...]

PR #3967, opened 2021-04-30
Use RW for cache lock

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3964

@juliantaylor I've prepared a first fix that I'd like if you have a chance to test in your environment.

I'm still working on the case as I don't consider it closed until we can check properly.

Basically this fix will protect the creation of new namespaces at some scale.

This issues has not been detected before in the test-race and in our testing, I'm ccing @FilipB to help me to test this in our perf tests as well to see if we have any chance to reproduce.

Again, I can't simulate in an easy way a high level of kubernetes informers to create a mockup of the same situation, that will be done in a future step, but I'm curious if with this protection, it will prevent your environment to fail.

Thanks !

### DIFF
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -63,7 +63,7 @@ type (
 		cacheIstioTypes        map[string]bool
 		stopChan               map[string]chan struct{}
 		nsCache                map[string]typeCache
-		cacheLock              sync.Mutex
+		cacheLock              sync.RWMutex
 		tokenLock              sync.RWMutex
 		tokenNamespaces        map[string]namespaceCache
 		tokenNamespaceDuration time.Duration
@@ -197,7 +197,12 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
 	if !c.isCached(namespace) {
 		return false
 	}
-	if _, exist := c.nsCache[namespace]; !exist {
+
+	c.cacheLock.RLock()
+	_, isNsCached := c.nsCache[namespace]
+	c.cacheLock.RUnlock()
+
+	if !isNsCached {
 		defer c.cacheLock.Unlock()
 		c.cacheLock.Lock()
 		return c.createCache(namespace)


hhovsepy (approved):
Solution LGTM

kubernetes/cache/cache.go
```diff
@@ -197,7 +197,12 @@ func (c *kialiCacheImpl) CheckNamespace(namespace string) bool {
 	if !c.isCached(namespace) {
 		return false
 	}
-	if _, exist := c.nsCache[namespace]; !exist {
+
+	c.cacheLock.RLock()
+	_, isNsCached := c.nsCache[namespace]
+	c.cacheLock.RUnlock()
```
burmanm, line 203: Between this RUnlock() and the next !isNsCached there's a possibility of schedulling to write the cache. It would probably just make sense to keep the writelock and then unlock it as soon as possible (without defer).

lucasponce: The write is a one time operation with a high cost, so this fix aims to protect the read on this method.

So, in a situation as you proposed:

```
No cache existing:

T1:
  c.cacheLock.RLock()
	_, isNsCached := c.nsCache[namespace]
	c.cacheLock.RUnlock()

T1 stop
T2:
	c.cacheLock.RLock()
	_, isNsCached := c.nsCache[namespace]
	c.cacheLock.RUnlock()

	if !isNsCached {
		defer c.cacheLock.Unlock()
		c.cacheLock.Lock()
		return c.createCache(namespace)
T2 quits
T1 continues:
	if !isNsCached {
		defer c.cacheLock.Unlock()
		c.cacheLock.Lock()
		return c.createCache(namespace)
```

Even if T2 creates the cache and T1 enters again in the c.createCache(namespace), in that method it will detect that cache is created and it will return the correct value, but readers of this will be protected.

juliantaylor:
Tested the branch and it cherry picked to our current 1.32.0 version with GOMAXPROCS set to 80 in the cluster with 640 namespaces. It does not crash after startup anymore as it did before so the fix seems to work.

lucasponce:
> Tested the branch and it cherry picked to our current 1.32.0 version with GOMAXPROCS set to 80 in the cluster with 640 namespaces. It does not crash after startup anymore as it did before so the fix seems to work.

Thank you very much @juliantaylor

FilipB (approved):
I'm not able to get any OCP cluster for perf testing today. I will revisit this later when the infra is fixed. Approving not to block this change.

lucasponce:
Thanks @FilipB.
I think we can consolidate and use the info of this PR to add more use cases to test similar scenarios like this one.

FilipB:
Infra issues are resolved and our perf pipeline is running again. There is no perf regression introduced by this change ( at least not visible by perf tests we currently have).

lucasponce:
Thanks @FilipB

PR #3970, opened 2021-04-30
Reduce number of API calls to discover remote Kialis

### LINKED ISSUE #3947: Multi-cluster discovery leads to time out while fetching server configs, blocking access
I installed kiali 1.33.1 and it is not loading on one cluster in nonprod and prod, but it works on every other cluster (although sometimes slow login times). This was same experience with 1.33.0 as well.

When I revert back to 1.32.0 it works.

The browser says:

> You are logged in, but there was a problem when fetching some required server configurations. Please, try refreshing the page.

All I see in the logs are

```
2021-04-26T19:00:50Z INF Not handling OpenId code flow authentication: No nonce code present. Login window timed out.
```

Kiali server helm values

```
auth:
  openid:
    client_id: [redacted]
    disable_rbac: true
    issuer_uri: [redacted]
    username_claim: email
  strategy: openid
deployment:
  affinity:
    pod_anti:
      preferredDuringSchedulingIgnoredDuringExecution:
      - podAffinityTerm:
          labelSelector:
            matchExpressions:
            - key: app
              operator: In
              values:
              - prometheus
          topologyKey: topology.kubernetes.io/zone
        weight: 100
  ingress_enabled: false
  node_selector:
    cloud.google.com/gke-nodepool: monitoring
  pod_anti:
    hack: true
  view_only_mode: true
external_services:
  grafana:
    auth:
      password: [redacted]
      type: basic
      username: admin
    in_cluster_url: http://prometheus-stack-core-grafana:80/
    url: https://core-forbes-development.grafana.forbes.com
  prometheus:
    url: http://core-prometheus:9090/
fullnameOverride: kiali-core
istio_namespace: istio-system
nameOverride: kiali-core
```

israel-hdez:
A "walk" through all remote cluster namespaces was being done to discover remote Kiali instances (that is, one query per namespace, per cluster).

This changes this "walk" for a simple query to fetch all services in a remote cluster. This should improve performance.

Fixes #3947.


business/mesh.go
```diff
@@ -304,34 +311,22 @@ func (in *MeshService) findRemoteKiali(clusterName string, kubeconfig *kubernete
...
-	// Kiali may be installed as an add-on.
-	if kialiInstances == nil && conf.IstioNamespace != os.Getenv("ACTIVE_NAMESPACE") {
-		kialiInstances = findKialiInNamespace(conf.IstioNamespace, clusterName, clientSet)
+	// - The operator and the helm charts set this well
+	//   known "app.kubernetes.io/name=kiali" label. It's also present in the
+	//   Istio addon manifest of Kiali.
+	// - We are using the "istio-reader-service-account" to connect to the
+	//   remote cluster. A typical Istio installation gives privileges to
+	//   this SA to list services in a cluster-wide way.
+	services, getSvcErr := clientSet.GetAllServicesByLabels("app.kubernetes.io/name=kiali")
```
jmazzitelli, line 320: People using the server-side helm chart can customize this name. I hate that feature, but some community people want it and use it, so we offer it. So you are not guaranteed `app.kubernetes.io/name` is "kiali". See: https://github.com/kiali/helm-charts/blob/master/kiali-server/values.yaml#L1-L2 and https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/_helpers.tpl#L63

jmazzitelli (commented):
See my comment - this is not going to work for some community members who installed Kiali using the server helm chart with a custom name.

business/mesh.go
jmazzitelli: I will assume there is a reason why "kiali.io/external-url" is an annotation not a label (I know there are limitations of label values such as length and allowed characters, maybe URLs are not available for label values). But if "kiali.io/external-url" is a label not an annotation then you can search on that very label. But, again, I'll assume there is a reason it is an annotation.

jmazzitelli: I just realized there is another label you can use: https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/_helpers.tpl#L56
```
app.kubernetes.io/part-of: kiali
```

We attach this label to indicate this resource is a part of the Kiali ecosystem. This is most likely what you want to use.

israel-hdez: It apparently works when using `app.kubernetes.io/part-of: kiali`.
I see the label is also present in the Istio add-on, so it shouldn't be an issue.

I changed it.

business/mesh.go
```diff
@@ -262,20 +275,16 @@ func findKialiInNamespace(namespace string, clusterName string, clientSet kubern
 	if kialiNs != nil {
 		// The operator and the helm charts set this fixed label. It's also
 		// present in the Istio addon manifest of Kiali.
-		services, getSvcErr := clientSet.GetServicesByLabels(kialiNs.Name, "app.kubernetes.io/name=kiali")
+		services, getSvcErr := clientSet.GetServicesByLabels(kialiNs.Name, "app.kubernetes.io/part-of=kiali")
```
lucasponce, line 278: @israel-hdez I'd like to ask you a change, perhaps not directly related to this PR but I guess it's important to introduce it or we'll add more technical debt to address.

In Kiali, all communication to the Kubernetes API should be done through the cache for local clusters, with the exceptions to some queries that doesn't have a proper Informer to build a cache.

Otherwise, we will introduce easily performance penalties very easily.

I understand that at the moment, we won't cache queries that are performed to a remote cluster, that's fine for now, but I'd ask if we can identify them in a better way, like renaming "clienteSet" to something like "remoteCluster" or similar.

That will help us to check if something is missing.

business/mesh.go
```diff
@@ -249,6 +250,18 @@ func (in *MeshService) ResolveKialiControlPlaneCluster(r *http.Request) (*Cluste
 	return kialiControlPlaneCluster, nil
 }
 
+// convertSvcToKialiInstance converts a svc Service data structure of the
+// Kubernetes client to a KialiInstance data structure.
+func convertSvcToKialiInstance(svc *core_v1.Service) KialiInstance {
```
jshaughn, line 255: This is only for Kiali services, right?  It's not  a generic find me the Kiali instance that is managing some random service?  IF so I think it would be more clear to rename this, `kialiServiceToInstance()` or something like that.  Because then it's mpre clear that, for example, the app.kubernetes.io/version label is reliable.  Because for random istio services it could very well just be `version`.

kubernetes/kubernetes.go
```diff
@@ -65,6 +66,19 @@ type OSClientInterface interface {
 	UpdateProject(project string, jsonPatch string) (*osproject_v1.Project, error)
 }
 
+// GetAllServicesByLabels fetches and returns all services in the whole cluster
+// that match the optional labelSelector. This is using the cluster-wide call
+// to fetch the services. The client will need to be created with an account that
+// has cluster-wide privileges to list services.
+func (in *K8SClient) GetAllServicesByLabels(labelsSelector string) ([]core_v1.Service, error) {
```
jshaughn, line 73: Maybe name this more explicitly?  `GetClusterServicesByLables()` ?   Perhaps it could be a convention for Cluster-wide operations.  I think `All` is a little vague.

jshaughn (commented):
Nothing blocking but a couple of comments for your consideration.

kubernetes/kubernetes.go
israel-hdez: I was using the "All" naming like in k8s docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.21/#list-all-namespaces-service-v1-core

But yes, using `Cluster` make a lot more sense. I renamed.

business/mesh.go
israel-hdez: I renamed to `convertKialiServiceToInstance`.

business/mesh.go
israel-hdez: Well, in this particular line it does not make sense a renaming. Here, the `clientSet` can be either a local or a remote client, depending on the caller. I think the clientSet is already abstracted from that fact.

However, I did a renaming on line 308 and also on line 487, it makes more sense on those lines.

lucasponce: I'm not agree.
When I read this code, I interpret "you are calling a direct k8s without using any informer/cache, why ?"

In a local cluster we must not do that but using the cache, in a remote cluster it's ok, as we don't want to apply informers in that api yet.

So, if this function is intended to be used remotely,

```
func findKialiInNamespace(namespace string, clusterName string, clientSet kubernetes.ClientInterface) 
```

my suggestion is to change the "clientSet" to "remoteClientSet" or something that clarify that, otherwise, these kind of calls are a performance penalty that we need to review and be sure that caching techniques are not bypassed for other features.

lucasponce: I haven't reviewed in deep this mesh.go class, but this is what it should be changed:

```
	// Discover ourselves
	kialiInstances := findKialiInNamespace(os.Getenv("ACTIVE_NAMESPACE"), myClusterName, in.k8s)
	if len(kialiInstances) > 0 && r != nil {
		for i := range kialiInstances {
			// If URL is already populated (because of an annotation), trust that because it's user configuration.
			if len(kialiInstances[i].Url) == 0 {
				kialiInstances[i].Url = httputil.GuessKialiURL(r)
			}
		}
	}
```
If that call is local, it's bypassing the services that are already cached.

You could fetch them from the cache and apply a local filter for labels, that would be better than perform a k8s call out of the cache scope.

jshaughn: +1, I thought this was covered in Lucas' original comments, but local cluster queries should utilize the k8s cache.

israel-hdez: I wish the cache could implement our `kubernetes.ClientInterface` and encapsulate all this if/else logic :-/

A added a commit that should solve this.

nrfox: > I wish the cache could implement our `kubernetes.ClientInterface` and encapsulate all this if/else logic :-/

Not knowing most of this codebase yet... that sounds like a good refactor. The client implementation (or a separate `CacheReader` impl) could do cache reads by default and the callers wouldn't need to worry about when to use the cache. If you still wanted the option of reading from the API server you could add an `APIReader()` that just returned a non cache reading impl. That way callers would need to opt into non-cache reads and it would be easier to audit. Obviously out of scope for this PR though.

business/mesh.go
```diff
@@ -249,33 +259,56 @@ func (in *MeshService) ResolveKialiControlPlaneCluster(r *http.Request) (*Cluste
...
 	}
 	if kialiNs != nil {
 		// The operator and the helm charts set this fixed label. It's also
 		// present in the Istio addon manifest of Kiali.
-		services, getSvcErr := clientSet.GetServicesByLabels(kialiNs.Name, "app.kubernetes.io/name=kiali")
+		var services []core_v1.Service
+		var getSvcErr error
+		if IsNamespaceCached(kialiNs.Name) {
+			var tmpSvc []core_v1.Service
+			tmpSvc, getSvcErr = kialiCache.GetServices(kialiNs.Name, nil)
```
lucasponce, line 291: Instead to add the filter logic here, you could reuse/add the filter logic into the kubernetes/filters.go

israel-hdez: OK, I moved the logic to kubernetes/filters.go
I hope this time all is OK.

lucasponce (approved):
Thanks Edgar.
Code looks good to me, also I performed a small smoke test but I couldn't test the multi-cluster area.
I'll wait until QE can test a real multi-cluster scenario before to merge.

As we have some users interested, this one can be a candidate to cherry-pick for a 1.34.1 version when ready.

pbajjuri20:
@israel-hdez @lucasponce: I have run E2E automation test cases on a Single Cluster. No issues found

pbajjuri20 (approved):

Issue #3966, opened 2021-04-30
Potential regression in one of the k8s APIs


lucasponce:
Recently we introduced some changes into the workloads calculation for TLS and it seems that it has a side effect to perform more calls than expected, creating a saturation in the API

```
I0430 10:57:41.640407   66727 request.go:655] Throttling request took 1.001500513s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 10:57:55.791107   66727 request.go:655] Throttling request took 1.002594522s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 10:58:10.796484   66727 request.go:655] Throttling request took 1.003164356s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 10:58:25.762255   66727 request.go:655] Throttling request took 1.001188882s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 10:59:40.775257   66727 request.go:655] Throttling request took 1.005343143s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 10:59:55.828042   66727 request.go:655] Throttling request took 1.001596827s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 11:00:10.765209   66727 request.go:655] Throttling request took 1.001956418s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 11:00:25.749659   66727 request.go:655] Throttling request took 1.005304468s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 11:00:40.775047   66727 request.go:655] Throttling request took 1.00559409s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 11:00:55.773801   66727 request.go:655] Throttling request took 1.003661159s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
I0430 11:01:10.775249   66727 request.go:655] Throttling request took 1.00277611s, request: POST:http://localhost:8080/apis/authorization.k8s.io/v1/selfsubjectaccessreviews
[...]

PR #3975, opened 2021-05-03
Fix get permissions queries

lucasponce:
Fixes https://github.com/kiali/kiali/issues/3966

Kiali performs a "Get permissions" query to enable/disable write features (in the UI, to render actions or create/update/delete buttons).

This query was doing some extra calls, also now that the list of verbs are explicit in the operator config, one of the calls was wrong and introduced a potential performance penalty.

This PR reduces the calls needed for the "get permissions" logic.

How to test:
- Deploy Kiali in a the default scenario and check no regressions are introduced.
- Deploy Kiali in "view_only_mode" [1] and check no regressions are introduced.
- No suspicious error/warning in Kiali logs.

[1] https://kiali.io/documentation/latest/installation-guide/#_reducing_permissions_in_openshift

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -66,13 +66,18 @@ func (icc IstioConfigCriteria) Include(resource string) bool {
 }
 
 // IstioConfig types used in the IstioConfig New Page Form
-var newIstioConfigTypes = []string{
-	kubernetes.AuthorizationPolicies,
+// networking.istio.io
+var newNetworkingConfigTypes = []string{
 	kubernetes.Sidecars,
 	kubernetes.Gateways,
+	kubernetes.ServiceEntries,
+}
+
+// security.istio.io
+var newSecurityConfigTypes = []string{
+	kubernetes.AuthorizationPolicies,
 	kubernetes.PeerAuthentications,
 	kubernetes.RequestAuthentications,
-	kubernetes.ServiceEntries,
 }
 
 // GetIstioConfigList returns a list of Istio routing objects, Mixer Rules, (etc.)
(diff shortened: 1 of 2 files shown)


hhovsepy (approved):
Regression tested,

skondkar (approved):
Verified on default and view only mode. LGTM

lucasponce:
Thanks @FilipB

lucasponce:
@FilipB another question that I'm thinking about.

```
/api/namespaces/bookinfo/services/ endpoint tests show 50% improvement in mean response time
/api/namespaces/bookinfo/istio/endpoint tests show 24% improvement in mean response time
```

These tests are for the details endpoint, the list one or both ? Do you have some figure between list vs details ? I ask because perhaps I need to check if there is an unnecessary query in the list endpoints.

FilipB:
Adding more details:
- all endpoints which are tested for services are visible here - https://gitlab.cee.redhat.com/istio/servicemesh-qe/jenkins-csb-declaration/-/blob/master/resources/kiali/performance/gatling/ServicesSim.scala#L46
- overall (combined for all endpoints) mean response time improvements are visible in previous comment
- response times for all endpoints were improved except for following: `/api/namespaces/bookinfo/services/details/graph?duration=60s&graphType=workload&includeIdleEdges=false&injectServiceNodes=true&appenders=deadNode,sidecarsCheck,serviceEntry,istio,healthConfig` this one is now worse. Was 78ms, no it is 306ms
- in case of istio configs, response times for all measured endpoints were improved

@lucasponce is the performance drop for `/api/namespaces/bookinfo/services/details/graph?....` is expected?

lucasponce:
> response times for all endpoints were improved except for following: /api/namespaces/bookinfo/services/details/graph?duration=60s&graphType=workload&includeIdleEdges=false&injectServiceNodes=true&appenders=deadNode,sidecarsCheck,serviceEntry,istio,healthConfig

The comparison is between master vs this PR ? or this PR vs a released version ?

I remember that the graph change the algorithm to get better accurate info between inbound/outbound and there were some extra calls that perhaps explains what you're seeing here 

cc @jshaughn to see if it's worth to treat that in a separate issue.

The GetPermissions logic is used in the details pages to render if a user can perform a write operation, so your results meet the expected improvement, so it seems that the graph perf drop is not related to this work.

FilipB:
The drop is not visible when using version built from 26bd3db19caaac94f739bf60d44c32c22a2c70b6 but is visible when using build from 88ded0e1827fc3a419b27a46fcc4d9aaf74442fe. So it's caused by this PR. I can retest to be 100% sure if you don't see any way how it could be caused by this PR. But even with this drop for one endpoint the overall combined response time is still better.

FilipB:
Please hold on with some deeper investigation. Test results are now very unstable for some reason. Need to find out why.

FilipB:
@jshaughn @lucasponce I revisited this and I can confirm my previous comments. This PR improves overall mean response time but RT for `/api/namespaces/bookinfo/services/details/graph?duration=60s&graphType=workload&includeIdleEdges=false&injectServiceNodes=true&appenders=deadNode,sidecarsCheck,serviceEntry,istio,healthConfig` is now worse. Was 78ms before this PR, and it is 306ms after merge. Shall I open new issue for this?

lucasponce:
>  Shall I open new issue for this?

Yes please, I think that this would need some refactoring, as there are pages that require to check the permissions (i.e. the overview page) because we need to show to the user if they can perform that operation, but that's not the case for the Graph appenders, so permissions query could be bypassed on such cases.

But this will trigger is cascade other changes in the business package that can be done isolated, so, better o fill a new issue to not forget.

jshaughn:
Thanks, Filip, nice find.

business/istio_config.go
```diff
@@ -606,33 +611,79 @@ func (in *IstioConfigService) GetIstioConfigPermissions(namespaces []string) mod
...
+			securityRP := make(models.ResourcesPermissions, len(newSecurityConfigTypes))
+			networkingPermissions[ns] = &networkingRP
+			securityPermissions[ns] = &securityRP
+			/*
+				We can optimize this logic.
+				Instead of query all editable objects of networking.istio.io and security.istio.io we can query
+				only one per API, that will save several queries to the backend.
+
+				Synced with:
+				https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/role.yaml#L62
```
jmazzitelli, line 631: just FYI - this link will eventually grow stale because its linking to "master" rather than a fixed tag. Probably should have used "v1.36.0".

business/istio_config.go
```diff
@@ -606,33 +611,79 @@ func (in *IstioConfigService) GetIstioConfigPermissions(namespaces []string) mod
...
 	if api, ok := kubernetes.ResourceTypesToAPI[objectType]; ok {
 		resourceType := objectType
-		ssars, permErr := k8s.GetSelfSubjectAccessReview(namespace, api, resourceType, []string{"create", "patch", "update", "delete"})
+		/*
+			Kiali only uses create,patch,delete as WRITE permissions
+
+			"update" creates an extra call to the API that we know that it will always fail, introducing extra latency
+
+			Synced with:
+			https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/role.yaml#L62
```
jmazzitelli, line 684: just FYI - this link will eventually grow stale because its linking to "master" rather than a fixed tag. Probably should have used "v1.36.0".

business/istio_config.go
```diff
@@ -652,7 +701,38 @@ func getPermissions(k8s kubernetes.ClientInterface, namespace, objectType string
...
+
+func getPermissionsApi(k8s kubernetes.ClientInterface, namespace, api string) (bool, bool, bool) {
+	var canCreate, canPatch, canDelete bool
+	/*
+		Kiali only uses create,patch,delete as WRITE permissions
+
+		"update" creates an extra call to the API that we know that it will always fail, introducing extra latency
+
+		Synced with:
+		https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/role.yaml#L62
```
jmazzitelli, line 715: just FYI - this link will eventually grow stale because its linking to "master" rather than a fixed tag. Probably should have used "v1.36.0".

business/services.go
```diff
@@ -243,15 +243,19 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime
...
-
 	var drCreate, drUpdate, drDelete bool
 	go func() {
 		defer wg.Done()
-		drCreate, drUpdate, drDelete = getPermissions(in.k8s, namespace, kubernetes.DestinationRules)
+		/*
+			We can safely assume that permissions for VirtualServices will be similar as DestinationRules.
+
+			Synced with:
+			https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/role.yaml#L62
```
jmazzitelli, line 253: just FYI - this link will eventually grow stale because its linking to "master" rather than a fixed tag. Probably should have used "v1.36.0".

Issue #3977, opened 2021-05-03
Istio gateway pod logs are not treated as Envoy proxy logs


jshaughn:
This may be a bug, or it may be expected, it depends on what our expectation is for Pod.IstioContainers.  My thinking was that Pod.IstioContainers was really just equivalent to "Pod.IstioProxy", as the expectation seems to be that this will always be empty or just hold one container, the proxy.  If that is the case then for istio-ingressgateway and istio-egressgateway the container should be returned in  Pod.IstioContainers and not Pod.Containers, as the container IS a proxy.

The problem currently is that despite ingress and egress generating envoy proxy logs, the new workload pod logs view is not treating them that way, and so it doesn't offer the parsed logs.

This problem could also be solved on the client, but I wrote this up because a server-side change may be the correct thing to do, as it would then apply to any client caller.  Looking for an opinion from @lucasponce or anyone else.

jshaughn:
@lucasponce Did you have an opinion on this?  I have a server-side fix local on my box but waiting for feedback.

lucasponce:
> Pod.IstioContainers and not Pod.Containers, as the container IS a proxy.

Not sure if this will work in all cases, but +1 to see in the PR and test.
I understand that the workload logs tab doesn't work well in those cases so a edge case is perfectly valid to address that.

(Sorry, I completely missed this notification in github)

PR #3981, opened 2021-05-06
Make sure Istio gateways are captured as proxies

jshaughn:
@lucasponce Here is a server-side solution to #3977.  It could also be solved client-side, either way requires some special-casing of the istio pods (gateways) that are themselves proxies, not injected, but native.  This change allows clients to treat them as proxies, in particular the workload logs can then offer the special handling of the access logs.

So, let me know if this solution seems OK to you, or if you think a client-side solution would be better.

Fixes https://github.com/kiali/kiali/issues/3977

REQUIRES UI PR: https://github.com/kiali/kiali-ui/pull/2155
Related Operator PR: https://github.com/kiali/kiali-operator/pull/314

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -185,6 +185,7 @@ type ComponentStatuses struct {
 type ComponentStatus struct {
 	AppLabel  string `yaml:"app_label,omitempty"`
 	IsCore    bool   `yaml:"is_core,omitempty"`
+	IsProxy   bool   `yaml:"is_proxy,omitempty"`
 	Namespace string `yaml:"namespace,omitempty"`
 }
 
@@ -435,14 +436,17 @@ func NewConfig() (c *Config) {
 						{
 							AppLabel: "istio-egressgateway",
 							IsCore:   false,
+							IsProxy:  true,
 						},
 						{
 							AppLabel: "istio-ingressgateway",
 							IsCore:   true,
+							IsProxy:  true,
 						},
 						{
 							AppLabel: "istiod",
 							IsCore:   true,
+							IsProxy:  false,
 						},
 					},
 				},
--- a/models/pod.go
+++ b/models/pod.go
@@ -2,6 +2,7 @@ package models
 
 import (
 	"encoding/json"
+	"strings"
 
 	core_v1 "k8s.io/api/core/v1"
 
@@ -117,7 +118,7 @@ func (pod *Pod) Parse(p *core_v1.Pod) {
 }
 
 func isIstioProxy(pod *core_v1.Pod, container *core_v1.Container, conf *config.Config) bool {
-	return pod.Namespace == conf.IstioNamespace && (pod.Name == "istio-ingressgateway" || pod.Name == "istio-egressgateway" || container.Name == "istio-proxy")
+	return pod.Namespace == conf.IstioNamespace && (container.Name == "istio-proxy" || strings.HasPrefix(pod.Name, "istio-ingressgateway") || strings.HasPrefix(pod.Name,"istio-egressgateway")
 }
 
 func lookupImage(containerName string, containers []core_v1.Container) string {
--- a/models/pod.go
+++ b/models/pod.go
@@ -103,13 +103,21 @@ func (pod *Pod) Parse(p *core_v1.Pod) {
 			Name:  c.Name,
 			Image: c.Image,
 		}
-		pod.Containers = append(pod.Containers, &container)
+		if isIstioProxy(p, &c, conf) {
+			pod.IstioContainers = append(pod.IstioContainers, &container)
+		} else {
+			pod.Containers = append(pod.Containers, &container)
+		}
 	}
 	pod.Status = string(p.Status.Phase)
 	pod.StatusMessage = string(p.Status.Message)
 	pod.StatusReason = string(p.Status.Reason)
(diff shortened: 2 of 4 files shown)


models/pod.go
```diff
@@ -112,6 +116,10 @@ func (pod *Pod) Parse(p *core_v1.Pod) {
 	_, pod.VersionLabel = p.Labels[conf.IstioLabels.VersionLabelName]
 }
 
+func isIstioProxy(pod *core_v1.Pod, container *core_v1.Container, conf *config.Config) bool {
+	return pod.Namespace == conf.IstioNamespace && (pod.Name == "istio-ingressgateway" || pod.Name == "istio-egressgateway" || container.Name == "istio-proxy")
```
lucasponce, line 120: Note that pod would never be a "istio-ingressgateway" but it should be typically a "istio-ingressgateway-79df7c789f-gtw5j"

jshaughn: good point :-)  That should be a "startsWith" logic.  Or, we could limit this to the container.Name test.

jshaughn: Fixed

models/pod.go
```diff
@@ -112,6 +117,10 @@ func (pod *Pod) Parse(p *core_v1.Pod) {
 	_, pod.VersionLabel = p.Labels[conf.IstioLabels.VersionLabelName]
 }
 
+func isIstioProxy(pod *core_v1.Pod, container *core_v1.Container, conf *config.Config) bool {
+	return pod.Namespace == conf.IstioNamespace && (container.Name == "istio-proxy" || strings.HasPrefix(pod.Name, "istio-ingressgateway") || strings.HasPrefix(pod.Name, "istio-egressgateway"))
```
jmazzitelli, line 121: My radar goes off when I see hard-coded names like this. Are we guaranteed that these are the names Istio always uses? Are these customizable by the user via istioctl? If so, we need to either (a) document that they MUST be named this for Kiali to work or (b) make these customizable in the Kiali CR also

jshaughn: In short, I have no idea.  But the fallout is fairly minimal if these names don't match, and as such I'm not going to overblow this PR twith configurability.  If a real user has a real problem we can address it at that time.

jshaughn:
@lucasponce @jmazzitelli , OK, so I've pushed another possible approach.  As discussed with Lucas, the initial approach is not valid, we can't treat a native proxy container (like the gateways) the same way as an injected proxy.  Injected containers have semantics surrounding them, and so we want to reserve Pod.IstioContainers for injected containers in a dataPlane pod.

In the latest revision I'm proposing we add a flag to the ContainerInfo itself, indicating whether the container is a proxy.  This would have the advantage of still being a server-side solution, and letting other server or ui code be able to do special handling for proxy containers, whether they are in a control plane or data plane pod.  Meanwhile, Mazz will be more happy as this also includes a config-driven way to identify control-plane pods.

models/pod.go
```diff
@@ -81,15 +83,19 @@ func (pod *Pod) Parse(p *core_v1.Pod) {
 		if err == nil {
 			for _, name := range scs.InitContainers {
 				container := ContainerInfo{
-					Name:  name,
-					Image: lookupImage(name, p.Spec.InitContainers)}
+					Name:    name,
+					Image:   lookupImage(name, p.Spec.InitContainers),
+					IsProxy: false,
```
lucasponce, line 88: Note that if you propose the "IsProxy" attribute/field to mark "this is container is a Proxy", the Istio init-container is typically a proxy as well: 
```
    image: docker.io/istio/proxyv2:1.9.1
    imagePullPolicy: Always
    name: istio-init
```

lucasponce:
Propose a IsProxy flag in the ContainerInfo sounds harmless to the existing logic.
As this PR touches the config + UI, it needs a counter part in those repos, also note that the model needs to be in sync with 
https://github.com/kiali/kiali-ui/blob/master/src/types/IstioObjects.ts#L115 as well.

models/pod.go
jshaughn: Good point, I will makethe change.  The logs generated by the init-container are unlikely to benefit from access-log parsing but it is a proxy and should be flagged correctly for client-side consumption.  I will make the change.

jshaughn:
> As this PR touches the config + UI, it needs a counter part in those repos, 

@lucasponce Yes, I already had the UI-side changes locally, I was just waiting on a review of the server-side proposal.  I have now created the UI PR and have updated this PR description with the link...

config/config.go
```diff
@@ -185,6 +185,7 @@ type ComponentStatuses struct {
 type ComponentStatus struct {
 	AppLabel  string `yaml:"app_label,omitempty"`
 	IsCore    bool   `yaml:"is_core,omitempty"`
+	IsProxy   bool   `yaml:"is_proxy,omitempty"`
```
lucasponce, line 188: Note that his change requires a operator change at least in the def and defaults
https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L508

lucasponce (approved):
Code looks good, but remember that it needs a PR in the kiali-operator at least to update the main CR + defaults file.

config/config.go
jshaughn: oh wow, good catch.

jmazzitelli (approved):
lgtm

pbajjuri20 (approved):

PR #3999, opened 2021-05-11
Assign anyuid scc to correct group

FilipB:
Assigning anyuid to system:serviceaccounts is too broad and is causing
error in OCP internal openshift-service-ca (The expected SCC for service-ca pods is restricted, any other SCC being applied to it will cause the CreateContainerConfigError issue.) anyuid scc was added to service-ca pod via system:serviceaccounts, now
this script only adds anyuid to system:serviceaccounts:<namespace>

### DIFF
--- a/hack/istio/install-istio-via-istioctl.sh
+++ b/hack/istio/install-istio-via-istioctl.sh
@@ -260,7 +260,7 @@ if [ "${DELETE_ISTIO}" != "true" ]; then
     fi
 
     echo Performing additional commands for OpenShift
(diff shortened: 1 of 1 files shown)


jmazzitelli:
We should check to see if we even need this. I remember some permissions requirements changed in newer versions of Istio. We should check the OpenShift-specific install instructions on istio.io and see if this is still required (and even with that, we should confirm by testing if we really need this - I remember recently those docs were incorrect and were telling us to give out permissions that were no longer required).

FilipB:
It's still required according to https://istio.io/latest/docs/setup/platform-setup/openshift/ I have not tested if it's needed but at least this new version does no harm as it's specific to `system:serviceaccounts:<namespace>` group

PR #4006, opened 2021-05-12
Reduce number of API calls to discover remote Kialis (#3970)

israel-hdez:
Back port of #3970


israel-hdez:
@jshaughn Asking for a review from you, as you were involved in the PR for `master`.

Do we also want a sanity check from QE on this one?

jshaughn (approved):
LG

PR #4022, opened 2021-05-17
Allow read-only file system in Kiali's pod

### LINKED ISSUE #3810: Modify manifests to use readOnlyRootFilesystem
It's a good security practice to use readOnlyRootFilesystems when deploying workloads. During kiali bootstrap, kiali tries to update the env.js and index.html (rootPath), which violate the readOnlyRootFilesystem condition.

Update the documentation manifests to restrict readOnlyRootFilesystem.

I'm including a sample manifest that I'm using that uses an initContainer to copy the static files over to a memory volume.

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: kiali
spec:
  replicas: 1
  selector:
    matchLabels:
      app: kiali
  strategy:
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
    type: RollingUpdate
  template:
    metadata:
      name: kiali
      labels:
        app: kiali
      annotations:
        prometheus.io/scrape: "true"
        prometheus.io/port: "9090"
        kiali.io/runtimes: go,kiali
    spec:
      serviceAccountName: kiali
      initContainers:
        - name: copy
          securityContext:
            runAsUser: 1000
            runAsGroup: 1000
            runAsNonRoot: true
            allowPrivilegeEscalation: false
            readOnlyRootFilesystem: true
            privileged: false
            capabilities:
              drop:
                - ALL
          image: "quay.io/kiali/kiali:v1.31.0"
          command:
            - sh
            - -c
            - cp -r /opt/kiali/console/* /tmp/static
          volumeMounts:
            - mountPath: /tmp/static
              name: static-content
      containers:
        - image: "quay.io/kiali/kiali:v1.31.0"
          imagePullPolicy: Always
          name: kiali
          securityContext:
            runAsUser: 1000
            runAsGroup: 1000
            runAsNonRoot: true
            allowPrivilegeEscalation: false
            readOnlyRootFilesystem: true
            privileged: false
            capabilities:
              drop:
                - ALL
          command:
            - "/opt/kiali/kiali"
            - "-config"
            - "/kiali-configuration/config.yaml"
          ports:
            - name: api-port
              containerPort: 20001
            - name: http-metrics
              containerPort: 9090
          readinessProbe:
            httpGet:
              path: /kiali/healthz
              port: api-port
              scheme: HTTP
            initialDelaySeconds: 5
            periodSeconds: 30
          livenessProbe:
            httpGet:
              path: /kiali/healthz
              port: api-port
              scheme: HTTP
            initialDelaySeconds: 5
            periodSeconds: 30
          env:
            - name: ACTIVE_NAMESPACE
              valueFrom:
                fieldRef:
                  fieldPath: metadata.namespace
            - name: LOG_LEVEL
              value: "info"
            - name: LOG_FORMAT
              value: "text"
            - name: LOG_TIME_FIELD_FORMAT
              value: "2006-01-02T15:04:05Z07:00"
            - name: LOG_SAMPLER_RATE
              value: "1"
          volumeMounts:
            - name: kiali-configuration
              mountPath: "/kiali-configuration"
            - name: kiali-cert
              mountPath: "/kiali-cert"
            - name: kiali-secret
              mountPath: "/kiali-secret"
            - name: static-content
              mountPath: /opt/kiali/console
      volumes:
        - name: kiali-configuration
          configMap:
            name: kiali
        - name: kiali-cert
          secret:
            secretName: istio.kiali-service-account
            optional: true
        - name: kiali-secret
          secret:
            secretName: kiali
            optional: true
        - name: static-content
          emptyDir:
            medium: Memory
            sizeLimit: 50Mi
```

israel-hdez:
Specifically, stop writing changes to the index.html and env.js files, which come from the UI package.

These files are now generated on the back-end and served directly, rather than changing the files on the pod filesystem.

This allows using `readOnlyRootFilesystem: true` for the securityContext in the Kiali pod.

Related FAQ entry is removed in: https://github.com/kiali/kiali.io/pull/375

Fixes #3810

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -139,6 +139,7 @@ func serveIndexFile(w http.ResponseWriter) {
 	b, err := ioutil.ReadFile(path)
 	if err != nil {
 		log.Errorf("File I/O error [%v]", err.Error())
+		handlers.RespondWithDetailedError(w, http.StatusInternalServerError, "Unable to read index.html template file", err.Error())
 		return
 	}
 
--- a/routing/router.go
+++ b/routing/router.go
@@ -1,12 +1,18 @@
 package routing
 
 import (
+	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
+	"path/filepath"
+	"strings"
 
 	"github.com/gorilla/mux"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/handlers"
+	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/prometheus/internalmetrics"
 )
 
@@ -45,6 +51,16 @@ func NewRouter() *mux.Router {
 		webRootWithSlash = "/"
 	}
 
+	fileServerHandler := func(w http.ResponseWriter, r *http.Request) {
+		if r.RequestURI == webRootWithSlash || r.RequestURI == webRoot || r.RequestURI == webRootWithSlash+"index.html" {
+			serveIndexFile(w)
+		} else if r.RequestURI == webRootWithSlash+"env.js" {
+			serveEnvJsFile(w)
+		} else {
+			staticFileServer.ServeHTTP(w, r)
+		}
+	}
+
 	appRouter = appRouter.StrictSlash(true)
 
 	// Build our API server routes and install them.
@@ -67,7 +83,7 @@ func NewRouter() *mux.Router {
 	// All client-side routes are prefixed with /console.
 	// They are forwarded to index.html and will be handled by react-router.
 	appRouter.PathPrefix("/console").HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		http.ServeFile(w, r, conf.Server.StaticContentRootDirectory+"/index.html")
+		serveIndexFile(w)
 	})
 
 	if conf.Auth.Strategy == config.AuthStrategyOpenId {
@@ -75,12 +91,12 @@ func NewRouter() *mux.Router {
 			if !handlers.OpenIdCodeFlowHandler(w, r) {
 				// If the OpenID handler does not handle the request, pass the
 				// request to the file server.
-				staticFileServer.ServeHTTP(w, r)
+				fileServerHandler(w, r)
 			}
 		})
 	}
 
-	rootRouter.PathPrefix(webRootWithSlash).Handler(staticFileServer)
+	rootRouter.PathPrefix(webRootWithSlash).HandlerFunc(fileServerHandler)
 
 	return rootRouter
 }
@@ -91,4 +107,53 @@ func metricHandler(next http.Handler, route Route) http.Handler {
 		defer promtimer.ObserveDuration()
 		next.ServeHTTP(w, r)
 	})
+}
+
+// serveEnvJsFile generates the env.js file needed by the UI from Kiali configs. The
+// generated file is sent to the HTTP response.
+func serveEnvJsFile(w http.ResponseWriter) {
+	conf := config.Get()
+	var body string
+	if len(conf.Server.WebHistoryMode) > 0 {
+		body += fmt.Sprintf("window.HISTORY_MODE='%s';", conf.Server.WebHistoryMode)
+	}
+
+	if webRoot := strings.TrimSuffix(config.Get().Server.WebRoot, "/"); len(webRoot) > 0 {
+		body += fmt.Sprintf("window.WEB_ROOT='%s';", webRoot)
+	}
+
+	w.Header().Set("content-type", "text/javascript")
+	_, err := io.WriteString(w, body)
+	if err != nil {
+		log.Errorf("HTTP I/O error [%v]", err.Error())
+	}
+}
+
+// serveIndexFile takes UI's index.html as a template to generate a modified index file that takes
+// into account the web_root path configured in the Kiali CR. The result is sent to the HTTP response.
+func serveIndexFile(w http.ResponseWriter) {
+	webRootPath := config.Get().Server.WebRoot
+	webRootPath = strings.TrimSuffix(webRootPath, "/")
+
+	path, _ := filepath.Abs("./console/index.html")
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Errorf("File I/O error [%v]", err.Error())
+		return
+	}
+
+	html := string(b)
+	newHTML := html
+
+	if len(webRootPath) != 0 {
+		searchStr := `<base href="/"`
+		newStr := `<base href="` + webRootPath + `/"`
+		newHTML = strings.Replace(html, searchStr, newStr, -1)
+	}
+
+	w.Header().Set("content-type", "text/html")
+	_, err = io.WriteString(w, newHTML)
+	if err != nil {
+		log.Errorf("HTTP I/O error [%v]", err.Error())
+	}
 }
--- a/routing/router_test.go
+++ b/routing/router_test.go
@@ -104,14 +104,14 @@ func TestRedirectWithSetWebRootKeepsParams(t *testing.T) {
 		t.Fatal(err)
 	}
 	body, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 200, resp.StatusCode, "Response should not redirect")
+	assert.Equal(t, 500, resp.StatusCode, "Response should not redirect")
 
 	resp, err = client.Get(ts.URL + "/test/")
 	if err != nil {
 		t.Fatal(err)
 	}
 	body2, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 200, resp.StatusCode, string(body2))
+	assert.Equal(t, 500, resp.StatusCode, string(body2))
 
 	assert.Equal(t, string(body), string(body2), "Response with and without the trailing slash on the webroot are not the same")
 }
--- a/routing/router_test.go
+++ b/routing/router_test.go
@@ -104,14 +104,14 @@ func TestRedirectWithSetWebRootKeepsParams(t *testing.T) {
 		t.Fatal(err)
 	}
 	body, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 500, resp.StatusCode, "Response should not redirect")
+	assert.Equal(t, 200, resp.StatusCode, "Response should not redirect")
 
 	resp, err = client.Get(ts.URL + "/test/")
 	if err != nil {
 		t.Fatal(err)
 	}
 	body2, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 500, resp.StatusCode, string(body2))
+	assert.Equal(t, 200, resp.StatusCode, string(body2))
 
 	assert.Equal(t, string(body), string(body2), "Response with and without the trailing slash on the webroot are not the same")
 }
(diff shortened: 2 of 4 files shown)


routing/router.go
```diff
@@ -92,3 +108,52 @@ func metricHandler(next http.Handler, route Route) http.Handler {
...
+// into account the web_root path configured in the Kiali CR. The result is sent to the HTTP response.
+func serveIndexFile(w http.ResponseWriter) {
+	webRootPath := config.Get().Server.WebRoot
+	webRootPath = strings.TrimSuffix(webRootPath, "/")
+
+	path, _ := filepath.Abs("./console/index.html")
+	b, err := ioutil.ReadFile(path)
+	if err != nil {
+		log.Errorf("File I/O error [%v]", err.Error())
+		return
```
jmazzitelli, line 142: Shouldn't we return some kind of HTTP error status code and message here? It seems like a simple return will not provide the client any indication that a problem occurred. We should return some kind of 500 error here I think (writing it to "w")

jmazzitelli (changes requested):
@israel-hdez I didn't propose any implementation for the change I asked for - but I think we need to do something there in case of an error to let the user know something went wrong.

jmazzitelli:
One thing I did not comment on in my review because I think it could be something we do later if we need such an enhancement for performance, but this new code does not cache those files in memory. Those files (both env.js and index.html) never change for the lifetime of the Kiali server, so why couldn't we build the files one time (say, on the first request), cache their contents, and then serve the cached contents thereafter?

I don't know how often the env.js file is requested (that isn't requested for every UI request right? Just when the UI is initialized/loaded?). I don't know how much we save if we cache, but it seems like such a low-hanging fruit/easy-win that I don't see any reason why we shouldn't build these just once and cache them.

jmazzitelli:
Now that we do not require write access, we should lock down the pod security a bit more.

We should change these files:

* https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml
* https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/deployment.yaml
* https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/deployment.yaml

such that in the Kiali Deployment's `kiali` container, we add these constraints:

```
        securityContext:
          runAsUser: 1000
          runAsGroup: 1000
          runAsNonRoot: true
          allowPrivilegeEscalation: false
          readOnlyRootFilesystem: true
          privileged: false
          capabilities:
            drop:
            - ALL
```

I will create a separate github issue to do this work, as it is not required for this PR and can be done separately.

routing/router.go
israel-hdez: I added the message.
However, it broke some tests. I fixed them, but I don't like the fixes.

israel-hdez:
> One thing I did not comment on in my review because I think it could be something we do later if we need such an enhancement for performance, but this new code does not cache those files in memory. Those files (both env.js and index.html)

Both files are loaded on each load/refresh.

For the `env.js` everything is generated at runtime from memory. It just creates a string from two configs. So I think performance impact should be negligible. I would say it's tie, because before this, it was a read from filesystem. Now, it's an in-memory process.

For the index.html, there is the penalty of an added search&replace operation on top of the usual filesytem read. TBH I think the performance impact if that added search&replaces operation should be negligible because this is an in-pod operation. Usually, we suffer from performance issues when we do a trip to the network (query the API Server, p8s, jaeger, etc.). So, I would like to try first without any cache (i.e. this time, I prefer to save memory rather than save CPU cycles).

jmazzitelli:
> I would like to try first without any cache (i.e. this time, I prefer to save memory rather than save CPU cycles).

Actually, I meant cache it in the *server* (not the client). So the server would generate it once and only once. Store it in some go variable someplace.

But I agree we should wait to see how it goes. We can add that later if we need to.

routing/router_test.go
```diff
@@ -104,14 +104,14 @@ func TestRedirectWithSetWebRootKeepsParams(t *testing.T) {
 		t.Fatal(err)
 	}
 	body, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 200, resp.StatusCode, "Response should not redirect")
+	assert.Equal(t, 500, resp.StatusCode, "Response should not redirect")
```
jmazzitelli, line 107: is this even testing anything now? It should result in a successful request, but if we can't write to index.html because we don't have a console/index.html file, then this isn't testing what it was testing before, right?

routing/router_test.go
```diff
@@ -104,14 +104,14 @@ func TestRedirectWithSetWebRootKeepsParams(t *testing.T) {
...
 
 	resp, err = client.Get(ts.URL + "/test/")
 	if err != nil {
 		t.Fatal(err)
 	}
 	body2, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 200, resp.StatusCode, string(body2))
+	assert.Equal(t, 500, resp.StatusCode, string(body2))
```
jmazzitelli, line 114: same as my other comment

server/server_test.go
```diff
@@ -75,12 +75,12 @@ func TestRootContextPath(t *testing.T) {
 	checkHTTPReady(httpClient, serverURL)
 
 	// we should be able to get to our custom web root
-	if _, err = getRequestResults(t, httpClient, serverURL+testCustomRoot, noCredentials); err != nil {
+	if _, err = getRequestResults(t, httpClient, serverURL+testCustomRoot, noCredentials); err != nil && err.Error() != "Bad status: 500" {
```
jmazzitelli, line 78: same as my other comment - this isn't really testing successfully that we are able to get to the custom root path.

server/server_test.go
```diff
@@ -75,12 +75,12 @@ func TestRootContextPath(t *testing.T) {
...
 		t.Fatalf("Failed: Shouldn't have failed going to the web root: %v", err)
 	}
 
 	// we should be able to get to "/" root - this just forwards to our custom web root
-	if _, err = getRequestResults(t, httpClient, serverURL, noCredentials); err != nil {
+	if _, err = getRequestResults(t, httpClient, serverURL, noCredentials); err != nil && err.Error() != "Bad status: 500" {
```
jmazzitelli, line 83: same as my other comment - this isn't really testing successfully that we are able to get to the custom root path.

jmazzitelli (changes requested):
let's think about this some more and figure out how to get this to actually test what was being tested before (unless I am missing something with these changed tests).

jmazzitelli:
@israel-hdez I got the tests to pass with this change:

```diff
diff --git a/routing/router_test.go b/routing/router_test.go
index f22d47d5..15975d58 100644
--- a/routing/router_test.go
+++ b/routing/router_test.go
@@ -4,6 +4,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"os"
 	"testing"
 
 	"github.com/gorilla/mux"
@@ -85,6 +86,12 @@ func TestRedirectWithSetWebRootKeepsParams(t *testing.T) {
 	oldConfig := config.Get()
 	defer config.Set(oldConfig)
 
+	oldWd, _ := os.Getwd()
+	defer os.Chdir(oldWd)
+	os.Chdir(os.TempDir())
+	os.MkdirAll("./console", 0777)
+	os.Create("./console/index.html")
+
 	conf := new(config.Config)
 	conf.Server.WebRoot = "/test"
 	config.Set(conf)
@@ -104,14 +111,14 @@ func TestRedirectWithSetWebRootKeepsParams(t *testing.T) {
 		t.Fatal(err)
 	}
 	body, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 500, resp.StatusCode, "Response should not redirect")
+	assert.Equal(t, 200, resp.StatusCode, "Response should not redirect")
 
 	resp, err = client.Get(ts.URL + "/test/")
 	if err != nil {
 		t.Fatal(err)
 	}
 	body2, _ := ioutil.ReadAll(resp.Body)
-	assert.Equal(t, 500, resp.StatusCode, string(body2))
+	assert.Equal(t, 200, resp.StatusCode, string(body2))
 
 	assert.Equal(t, string(body), string(body2), "Response with and without the trailing slash on the webroot are not the same")
 }
diff --git a/server/server_test.go b/server/server_test.go
index 330b6b2b..3aab4a2c 100644
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -30,6 +30,12 @@ const (
 var tmpDir = os.TempDir()
 
 func TestRootContextPath(t *testing.T) {
+	oldWd, _ := os.Getwd()
+	defer os.Chdir(oldWd)
+	os.Chdir(os.TempDir())
+	os.MkdirAll("./console", 0777)
+	os.Create("./console/index.html")
+
 	testPort, err := getFreePort(testHostname)
 	if err != nil {
 		t.Fatalf("Cannot get a free port to run tests on host [%v]", testHostname)
@@ -75,12 +81,12 @@ func TestRootContextPath(t *testing.T) {
 	checkHTTPReady(httpClient, serverURL)
 
 	// we should be able to get to our custom web root
-	if _, err = getRequestResults(t, httpClient, serverURL+testCustomRoot, noCredentials); err != nil && err.Error() != "Bad status: 500" {
+	if _, err = getRequestResults(t, httpClient, serverURL+testCustomRoot, noCredentials); err != nil {
 		t.Fatalf("Failed: Shouldn't have failed going to the web root: %v", err)
 	}
 
 	// we should be able to get to "/" root - this just forwards to our custom web root
-	if _, err = getRequestResults(t, httpClient, serverURL, noCredentials); err != nil && err.Error() != "Bad status: 500" {
+	if _, err = getRequestResults(t, httpClient, serverURL, noCredentials); err != nil {
 		t.Fatalf("Failed: Shouldn't have failed going to / root: %v", err)
 	}
```

routing/router_test.go
israel-hdez: My thought is that this is testing that a redirection is *not* happening (or that's what I understand from the "Response should not redirect" string). 
Specifically, I *think* this is testing that this line is not being hit: https://github.com/kiali/kiali/blob/master/routing/router.go#L28. That line would return a `302` status code. So, my rationale was that as long as we don't get the `302` status code, the test should be green.

I had a similar rationale for `server_test.go`: as long as we get that `500` response, which is the expected within the test, I think it should be OK.

But I think mocking the index.html at runtime is fine. I'll do the changes.

israel-hdez:
@jmazzitelli OK, I applied your suggestion to the tests.
Let's hope we won't see sporadic failures, because CI runs `make test-race` which may behave slightly different than `make test`.

jmazzitelli:
linter failed - IMO, we need to skip the test code when lint is run

israel-hdez:
@jmazzitelli OK, CI is fixed by assigning errs to the underscore :D

jmazzitelli (approved):
lint made this so much easier to read! :)

ltgm

jmazzitelli:
Let's merge these three all together:
* this PR
* https://github.com/kiali/helm-charts/pull/76
* https://github.com/kiali/kiali-operator/pull/316

lucasponce:
@israel-hdez @jmazzitelli please, why this work has been merged without QE in the review / approval ?

jmazzitelli:
This work would have required QE to pull and review/test 3 separate PRs together as one. Figured it was easy enough for me and Edgar to test this stuff - it is easy enough to validate.

lucasponce:
> This work would have required QE to pull and review/test 3 separate PRs together as one. Figured it was easy enough for me and Edgar to test this stuff - it is easy enough to validate.

These kind of changes, unless you have performed regression tests, are better that are in QE radar as well to validate that things look ok.

jmazzitelli:
@mattmahoneyrh is putting his master build into OSSM 2.1 and see how it works there (which is the main reason why this needs to be fixed - so he's verifying this on OSSM 2.1).

Issue #4014, opened 2021-05-14
REST API panic if no Prometheus


esnible:
**Describe the bug**

Panic via REST API if there is no Prometheus.

To reproduce, I installed a dev version of Istio 1.10, but I did not install Prometheus.

```bash
kubectl -n istio-system exec deployment/kiali -- curl "http://localhost:20001/kiali/api/namespaces/graph?namespaces=default"
```

The response was
```json
[...]

PR #4026, opened 2021-05-18
Assume a connectivity issue on Prometheus API failures

jshaughn:
Assume a connectivity issue on Prometheus API failures. A CX issue is fairly typical, an application error is very unlikely.  This stops stack traces returned to the user on a prom connectivity problem.

Fixes https://github.com/kiali/kiali/issues/4014

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/prometheus/common/model"
 	core_v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/kiali/kiali/config"
@@ -148,16 +149,17 @@ func (in *HealthService) getNamespaceAppHealth(namespace string, appEntities nam
 		}
 	}
 
-	var errRate error
 	if sidecarPresent {
 		// Fetch services requests rates
 		rates, err := in.prom.GetAllRequestRates(namespace, rateInterval, queryTime)
-		errRate = err
+		if err != nil {
+			return allHealth, errors.NewServiceUnavailable(err.Error())
+		}
 		// Fill with collected request rates
 		fillAppRequestRates(allHealth, rates)
 	}
 
-	return allHealth, errRate
+	return allHealth, nil
 }
 
 // GetNamespaceServiceHealth returns a health for all services in given Namespace (thus, it fetches data from K8S and Prometheus)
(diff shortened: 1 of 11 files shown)


lucasponce (changes requested):
I guess this is not feature complete if we want to point that, this is only covering the graph feature but not health nor metrics:

image

Otherwise some endpoint may report an 503 and others a 500 for same scenario: Prometheus connectivity.

(I'm fine with the change, just requesting to use the same code in those endpoints, too).

jshaughn:
> ...this is only covering the graph feature but not health nor metrics:

@lucasponce OK, I wasn't sure if we cared about it so in the initial offering I responded specifically to the reported problem.  The error handling is not the same across the subsystems, so it will take some specific work in each place.  I'll follow up with another commit.

jshaughn:
@lucasponce OK, failed calls to Metrics will now also be returning 503s.

lucasponce:
I guess the health endpoints are still sending a 500
image

Probably it can be fix it in the kiali/handlers/errors.go

jshaughn:
@lucasponce OK, I think I plugged the remaining holes.  I'm seeing only 503s with Prom at 0 pods.

lucasponce (approved):
The 500 errors are moved into 503.
I see a couple of imrprovements in the UI as the health promises throw an error that is not properly caught in the AlertCenter, but that's ok, I think it's out of the scope of this PR.

jshaughn:
@mattmahoneyrh This is more of an internal change but if you want to perform QE you can cycle down prometheus to 0 pods and ensure that any requests that fail due to prometheus return 503 (ServiceUnavailable) as opposed to 500 (ApplicationError).

nrfox:
This is probably out of scope for this change since this centers around prometheus and the responses returned by prom connectivity but should the api be returning a stack trace at all to the client? It seems that was part of the concern in #4014 and it just so happened that prom connectivity was the cause of the error and the stack trace to be included in the response. Perhaps the api can log the stack [here](https://github.com/kiali/kiali/blob/5bcdd2f372ff2fcca4746e321bd9e93e677e1b0c/handlers/graph.go#L85-L90) when recovering from a panic in the graph but exclude it from the response.

jshaughn:
>  but should the api be returning a stack trace at all to the client?

I have no idea whether a stack trace is a bad thing to return to the client.  I know it was brought up as a, I guess, a potential security concern in the original issue, but up to this point no one has complained and it has made it easier to diagnose a couple of bugs by easily asking the user for a trace.  So today we return it in the response and we log it, but here's the thing, the logged trace is just as accessible to a client, as we already have a way of displaying kiali pod logs via the workload detail (or an API call).

So,I guess my vote is to leave it as is.

pbajjuri20:
Before PR:

bfd

After PR:

bbd

pbajjuri20 (approved):

Issue #321, opened 2018-06-27
Feature request: Configurable Context Root


sbueringer:
Hi,

I'm deploying Istio together with various addons like Grafana/Jager .. and Kiali . Because I didn't want to host all those addons on different hostnames on context root, I tried to deploy Kiali behind an Ingress controller (in this case Traefik) under /kiali. Obviously, because Kiali is listening on context root, I only get 404s. 

Is there any way to configure the context root on which Kiali is listening on?

jotak:
I fear it is not possible at the moment. I opened a feature request about that: https://issues.jboss.org/browse/KIALI-1059

vnugent:
@sbueringer the feature has been merged. Can you have a look when you get a chance?

sbueringer:
@vnugent I deployed the latest image on our environment with `/kiali` and it works perfectly. Thanks :)

PR #78, opened 2018-03-01
SWS-152 Handle Self-Requesting Services

jshaughn:

PR #4032, opened 2021-05-19
Kiali internal metrics revisit

### LINKED ISSUE #3244: Internal metrics revisit
_This is something that was discussed in a call and, IMO, the perception was that this is good to have. However, it's unsure if it's worth to do._

Kiali have a few internal metrics in place. For example, we expose how much it takes for Kiali to complete the processing of different paths within the code flow and we expose this metrics in a prometheus format. So, if the user has prometheus, it could collect Kiali internal metrics.

The number of internal Kiali metrics is little and Kiali is lacking some screens to give visibility and provide real usage for the users. Also, the scope is limited to the back-end and are quite technical metrics. These metrics were supposed to enable users to provide feedback when users report issues, but are probably underused.

The scope of these internal metrics probably should be extended to also cover the UI, and also to generate more less-technical, more high-level metrics. Some ideas:

* Measure which pages are the most used
* Measure which options/features are mostly turned on or off
* Error metrics
* Response time metrics

Some of these metrics could easily be generated by adding an Istio sidecar to Kiali. This could help to understand what Kiali features are the most used and how and this would empower users to share more refined feedback to the project, and can make it easier to ask for feedback when users report bugs/issues.

We could re-use the prometheus instance used for istio to store these metrics and provide some screens summarizing these metrics.

leandroberetta:
**Description**

The motivation behind this PR is to reduced the amount of metrics exposed via the Kiali internal metrics dashboard keeping just the API routes metrics (hit rate, duration and failures) and also adding the graph generation duration metric (with aggregations like graph kind and type).

The API routes metrics were already tracked in a middleware configured in the router but a solution to continue tracking errors was needed so this PR adds a new metric for counting API failure calls (500 and 503 errors) and the code to get these errors in the middleware. 

The removal of the Go function metrics tracking code is also a task done in this PR.

As a future effort, the following issue represents the addition of tracing to the code: #4036

**Issue**

Fixes #3244

**Related PRs**

- Helm Charts ([#78](https://github.com/kiali/helm-charts/pull/78))
- Kiali Operator ([#321](https://github.com/kiali/kiali-operator/pull/321))

**Documentation**

With the updated metrics and monitoring dashboard the metrics will look as follows (with the graphics collapsed):

dashboard1
dashboard2


prometheus/internalmetrics/internal_metrics.go
```diff
@@ -103,6 +104,13 @@ var Metrics = MetricsType{
 		},
 		[]string{},
 	),
+	APIFailures: prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "kiali_api_failures_total",
+			Help: "Counts the total number of failures encountered by a particular API handler.",
+		},
+		[]string{labelRoute},
+	),
```
jshaughn, line 113: I'm not against this new metric, I'm only wondering why the metric is specific to failures and not just recording all response codes?  Ignore this if it a silly question, I have never really looked at out metrics.   Assuming the new metric is just recording failures, why is it limited (below) to just 500 and 503?  Finally, would it be possible to record the API being called so we could tie together API with responses?

routing/router.go
```diff
@@ -101,11 +102,38 @@ func NewRouter() *mux.Router {
...
+}
+
+// updateMetric evaluates the StatusCode, if there is an error, increase the API failure counter, otherwise save the duration
+func updateMetric(route string, srw *statusResponseWriter, timer *prometheus.Timer) {
+	// Increase the error counter on 500 and 503 errors
+	if srw.StatusCode == http.StatusInternalServerError || srw.StatusCode == http.StatusServiceUnavailable {
+		internalmetrics.GetAPIFailureMetric(route).Inc()
+	}
+	// Always measure the duration even if the API call ended in an error
+	timer.ObserveDuration()
```
jmazzitelli, line 124: I know its very little time, but timer.ObserveDuration should be called as soon as we know the API call is done. Now we are timing not only the API call but also the call to get into this function, and the time it takes to invoke that if-stmt and the call to `GetAPIFailureMetric(route).Inc()`. I would just move the timer.ObserveDuration() as the very first thing this method does to get that timer to more accurately reflect the time. Not a big deal, but is more accurate this way. In fact, the most accurate way would be to put this back the way it was - do a `defer timer.ObserveDuration()` as the first defer call in metricHandler. Then you can call this method to check the failures. And really, since its only three lines, I don't think we even need an extra function here - just put that in the metricHandler to make it simpler - but that's just aesthetics - keeping an extra function is fine if it makes things more understandable/readable.
 
```suggestion
	// Always measure the duration even if the API call ended in an error
	timer.ObserveDuration()
	// Increase the error counter on 500 and 503 errors
	if srw.StatusCode == http.StatusInternalServerError || srw.StatusCode == http.StatusServiceUnavailable {
		internalmetrics.GetAPIFailureMetric(route).Inc()
	}
```

jmazzitelli (changes requested):

prometheus/internalmetrics/internal_metrics.go
jmazzitelli: > Finally, would it be possible to record the API being called so we could tie together API with responses?

I'm pretty sure this stuff does that - its an attribute on the time series IIRC.

leandroberetta: @jshaughn like @jmazzitelli said, it does record the API being called.

I'm limiting  the errors to 503 and 500 because I check edthe comments in every method and there wasn't a reference to another 5xx but I can re check that.

This API failure metric came out as an idea to replace the old "failures" that was used to show business failures, however It's possible to add a new dimension (label) to record status codes so that would introduce a new aggregation in the graph to show any status code return by the server. Instead of counting failures, it will count every response's status code (including 4xx, 3xx).

leandroberetta (commented):
ObserveDuration moved to the beginning of the updateMetric function.

prometheus/internalmetrics/internal_metrics.go
leandroberetta: I tried to add a new label to the metric: kiali_api_processing_duration_seconds to let users filter API calls by response code but after some testing, I realized I cannot create the timer with the label status code because in that moment of time the status code is not known. 

Anyway, this is not necessarily what we were discussing, I guess the decision is to leave the failure counter as it is right now or extend the counter to count every status code. 

My thoughts are that this counter's objective is to have more information in case of problems, relating the failure with an specfici API route.

hhovsepy:
Screenshot from 2021-05-21 15-11-25
Screenshot from 2021-05-21 15-11-49
Screenshot from 2021-05-21 15-12-20

lucasponce (commented):
This PR looks good to me.
As commented in https://github.com/kiali/kiali/issues/4036 it makes sense to instrument kiali code and move some of the "tracing" logic into jaeger and perhaps in the future, even the metrics per endpoints can be reduced.

But as a starting point I see positive to remove some of the instrumentation from the business package and keep it in the routes handlers.

The dashboard may need some adjustment when the PoC with the new metrics dashboard is complete but I have tested in a local environment

image

jmazzitelli:
FYI: before merging any of this, make sure the ["metrics-test" molecule test](https://github.com/kiali/kiali-operator/tree/master/molecule/metrics-test) still works (and perhaps modify it to incorporate a new metric that you add to see that the new stuff works). You don't have to test every metric, but testing one is good to make sure the infrastructure is working (which is what the molecule test is trying to do).

You can do this on minikube via:

```
hack/run-molecule-tests.sh --cluster-type minikube --minikube-exe $(which minikube) --client-exe $(which kubectl) -dorp $DORP  -at metrics-test
```

or on OpenShift via:

```
hack/run-molecule-tests.sh --cluster-type openshift --client-exe $(which oc) -dorp $DORP  -at metrics-test
```

lucasponce:
Hi @leandroberetta, probably you may need to rebase this work with master as I merged a PR that touched some files.

leandroberetta:
@lucasponce yes, i've just done the rebase and fixed the conflicts. 

I've also run the Molecule tests for the metrics and everything looks good:

metrics tests

I will remove the "do not merge" label from this PR and all related to this one.

jshaughn (approved):
This looks good to me.

prachiyadav (approved):

prachiyadav (approved):

lucasponce:
@jmazzitelli is this PR looking good to you ?

jmazzitelli (approved):

lucasponce (approved):

PR #4033, opened 2021-05-19
[hack] fix purge script so it purges correctly on minikube

jmazzitelli:
either this is new behavior in the latest kubectl or I never noticed this before but the purge script isn't deleting cluster resources on minikube because the command fails when it can't find any oauth or consolelink times (those are openshift only).

this fixes that problem


jshaughn (approved):
LOK

Issue #4027, opened 2021-05-18
Use Istiod service registry for Kiali Istio validations


lucasponce:
Kiali validation can be improved using Istiod service registry as a source of truth for hostname resolution.

In that way, some of the validations that doesn't depend of a Kubernetes Service can be implemented within Kiali.

PR #4034, opened 2021-05-20
Use Istiod Registry in Kiali validations

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4027

This PR fetches the ServiceRegistry from the IstioD component to be used as an additional source in the Kiali validations.

The goal is not to change how validations are processed but enrich the cases where the registry can complement the information needed to validate the hosts.

When @xeviknal is back we can discuss more changes on this area but for this PR this PR will fix the "false errors" in validations due the remove services (exported).

How to test:
- First, check that the Maistra federation scenarios are covered https://github.com/maistra/istio/tree/maistra-2.1/pkg/servicemesh/federation/example/config-poc (configurations with no "false validation errors").
- Second, check that similar scenarios defined in multi-cluster upstream are covered https://istio.io/latest/docs/setup/install/multicluster/ (configurations with no "false validations errors").
- Third, a regression test in validation features.
- Forth, there is an additional fetch to the IstioD registry that is cached, so it shouldn't introduce any perf penalty, but worth to test if there is any change in that aspect.


hhovsepy:
Regression testing with existing validation scenarios pass.

hhovsepy (approved):
Code LGTM

FilipB:
Perf tests in progress.

lucasponce:
@FilipB @skondkar in case you need it, we can merge this one if in single cluster scenarios works well and no regression, then this PR will make the next kiali 1.35 and it can help the testing in other environments.

skondkar (approved):
Verified existing validations scenarios manually. There are no regressions on single cluster. LGTM.

lucasponce:
Thanks @skondkar !

As commented, we'll consolidate this on master and we'll iterate on the multicluster and federation scenarios when we have the proper infrastructure available.

FilipB:
Adding performance results.
There is a 3.8% degradation in mean response time for  `/api/namespaces/bookinfo/istio/` endpoints overall (average for all tested endpoints https://gitlab.cee.redhat.com/istio/servicemesh-qe/jenkins-csb-declaration/-/blob/master/resources/kiali/performance/gatling/IstioConfigSim.scala)
The one which is affected the most is `/api/namespaces/bookinfo/istio?validate=true`. Mean response time in master was 51ms, in this pr it's ~100ms.

FilipB:
@lucasponce Do you want to investigate the drop or it's expected and I should just update baselines?

lucasponce:
@FilipB the degradation may be expected as there is an additional query/fetch to the Istiod registry, which is cached but it would impact in the context of the overall performance.

I'd say to let's update the baselines, as kiali 1.36.0 will be the base for SM 2.1 we can compare them in next iterations.

Issue #4049, opened 2021-05-27
Possible false positive reported as violating KIA0202


ferrarimarco:
**Describe the bug**

Kiali is reporting a violation of the KIA0202 ([description](https://kiali.io/documentation/latest/validations/#_kia0202_this_host_has_no_matching_entry_in_the_service_registry_service_workload_or_service_entries), [source](https://github.com/kiali/kiali/blob/master/business/checkers/destinationrules/no_dest_checker.go)) rule on a DestinationRule that points to a ServiceEntry, which I believe it's a false positive.

The workload is running in a VM that I added to the Istio mesh.

```yaml
---
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: productpage
spec:
  hosts:
    - productpage.default.svc.cluster.local
  location: MESH_INTERNAL
  ports:
    - number: 9080
      name: http
      protocol: HTTP
  resolution: DNS
  workloadSelector:
    labels:
      app: productpage
      instance-id: source-environment-vm
...
---
apiVersion: networking.istio.io/v1alpha3
kind: DestinationRule
metadata:
  name: productpage
spec:
  host: productpage.default.svc.cluster.local # <- Kiali reports the KIA0202 validation error here
  trafficPolicy:
    tls:
      mode: ISTIO_MUTUAL
...
```

**Versions used**
```
Kiali UI: 1.34.1 (cda505c688a0e8b6ce65dd80ca0203a5153d977d)
Kiali Server: v1.34.1 (31fc3596219485ab2e37d3968dc54910f3378992)
Kiali Container: v1.34.1

Istio: 1.10.0
Prometheus: 2.24.0
Kubernetes: v1.20.6-gke.1000
Grafana URL: N/A
Jaeger URL: N/A
```

**To Reproduce**
1. Provision a GKE cluster
2. Deploy the resources above
3. Open Kiali.
4. Open the Istio config page
5. Inspect the DestinationRule object
6. See the validation error

**Expected behavior**

Kiali shouldn't report a violation of the KIA0202, given that the hosts match.

Thanks!

ferrarimarco:
Maybe https://github.com/kiali/kiali/pull/4034 fixes this?

lucasponce:
@ferrarimarco thanks ! yes I suspect #4034 should address this as if the host is present in the Istiod registry it should be used for validation if no other Kubernetes Service or ServiceEntry is found.

I'll test.

lucasponce:
[...]

PR #4056, opened 2021-06-01
Use FQDN in Istiod Registry validations

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4049

This PRs fixes the use of FQDN when validating if a host is present in the Istiod registry.
This work also "adjusts" the behaviour of several cross-namespace validation on the DestinationRules, now that the registry has all services available, a DR can detect when a host is present or not.

How to test:

- Check that a scenario like https://github.com/kiali/kiali/issues/4049 is supported.
- Check that a service defined in a ns1 namespace is validated in a DR/VS if it's defined in a ns2 namespace.
- Validate that the only changes/regressions will be the "cross-namespace" ones that now should be adjusted in the integration tests.

@skondkar @hhovsepy I expect that this PR may be not 100% complete and some corner case may be missed, let me know if you find some missing scenario or don't hesitate to ask me more details if the QE description is not complete. (thx !)

cc @xeviknal

### DIFF
--- a/business/checkers/authorization/no_host_checker.go
+++ b/business/checkers/authorization/no_host_checker.go
@@ -129,7 +129,7 @@ func (n NoHostChecker) hasMatchingService(host kubernetes.Host, itemNamespace st
 	for _, rStatus := range n.RegistryStatus {
 		// We assume that on these cases the host.Service is provided in FQDN
 		// i.e. ratings.mesh2-bookinfo.svc.mesh1-imports.local
-		if kubernetes.FilterByRegistryStatus(host.Service, rStatus) {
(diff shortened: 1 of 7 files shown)


hhovsepy:
Now I see error in DR only, not in ServiceEntry.

hhovsepy (approved):
Code LGTM:
Before:
Screenshot from 2021-06-03 09-00-57

After:
Screenshot from 2021-06-03 08-31-38

skondkar (approved):
Verified cross namespace support and some combinations including validations for wrong namespaces/servicenames.
Also had a UI automation run on validation tests. No regressions there.  LGTM

Issue #4057, opened 2021-06-02
Consolidate Dashboards CRDs into main Kiali config, also handled via Kiali Operator


lucasponce:
Custom Dashboards is a feature that allows Kiali to extend the metrics tabs with specific dashboards definitions [1].

This isssue doesn't touch the business of the feature but it proposes a change in the implementation of the definitions.

I think it could be good to move the defaults definitions into the regular Kiali config, also managed via helm/manual or via the operator.

So, same definitions populated via the kiali moniroting API could be found from the config.

[1] https://kiali.io/documentation/latest/runtimes-monitoring/

jmazzitelli:
I have not thought about this a lot, so I have no firm opinion. But just to point out the "pros and cons":

PRO: This will make the implementation and installation of the dashboards much easier.
CON: But this will make the Kiali CR (or helm values.yaml, if using helm) much larger and complex for the end user to manage.

If I had to choose right now what do to, I would say, yes, let's do this. In fact, if we do this, it solves [a problem we have today](https://github.com/kiali/kiali-operator/pull/313#issuecomment-841528868) with managing multiple instances of Kiali in the same namespace. I'll repeat what I said over there, just to have it documented in this issue:

===

This is a rare edge case. The following will not be handled gracefully, but can be worked around.

If you elect to install multiple Kialis in the same deployment namespace, the dashboards you enabled in each Kiali CR will overwrite each other. This is fine when installing multiple Kialis, since dashboards have the same spec content whether installed by one or N Kiali instances. The problem is when you remove one Kiali CR to uninstall the one Kiali instance, the dashboards will be removed, thus all other Kiali instances will see them disappear.

The workaround is fairly simple - just tell one Kiali CR to install the dashboards again (to do this, you can simply touch the Kiali CR, thus forcing the operator to run a reconciliation). Another alternative is the user manage the dashboards manually by creating them outside of having the Kiali operator deploy them (this is how you would do it, for example, if you have your own custom dashboards, anyway).

There is talk about changing the way dashboards are published/deployed, so I won't spend any time trying to make this work more gracefully. And, again, this is only a problem if you want to install multiple Kialis in the same namespace and then you delete only one of them and want to keep the others.

lucasponce:
Thanks @jmazzitelli for the feedback.

[...]

PR #4059, opened 2021-06-03
consolidate dashboards

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4057

Here are all the PRs (including this one) that are related to this issue:
* https://github.com/kiali/kiali/pull/4059
* https://github.com/kiali/kiali-operator/pull/326
* https://github.com/kiali/helm-charts/pull/81
* https://github.com/kiali/kiali.io/pull/380


jmazzitelli:
Just a PoC for now

jmazzitelli:
Hmm... I think this isn't right:
```
if strings.TrimSpace(val.DiscoverOn) != "" || len(val.Items) >; 0 {
  newList = append(newList, val);
}
```
I think it is allowed that "DiscoverOn" can be empty (at least that's what the original documentation says). I think I will need to change this and add a unit test.

lucasponce:
In a first pass, things look good on my side.

There is an UI issue that is reported https://github.com/kiali/kiali-ui/pull/2132#issuecomment-854738056 but it will require a followup work but it's unrelated to this CRD work.

I'm going to perform more tests but nothing relevant at the moment.

lucasponce (commented):
LGTM
I haven't tested the deep personalizations of the runtimes (i.e. validating that a dashboard can be changed at Pod level) but the basic examples defined in [1] works without problem.

As commented, there is a pending task commented on [2] that is visible but it is not related to this work, indeed, I'm working on that task and I'd like to see this merge to consolidate both as probably it will touch the defaults.

[1] https://github.com/kiali/demos/tree/master/runtimes-demo
[2] https://github.com/kiali/kiali-ui/pull/2132#issuecomment-854738056

lucasponce (approved):

Issue #4079, opened 2021-06-08
clean up hack/istio/bookinfo* resources


jmazzitelli:
We have a bunch of `bookinfo*.yaml` files used to create Istio objects for dev testing. These are in here: https://github.com/kiali/kiali/tree/master/hack/istio

Last time I used this was many Istio versions ago and I'm sure some of these are obsolete (using older Istio api versions that are probably not supported anymore).

Also check this to make sure it is still valid: https://github.com/kiali/kiali/blob/master/hack/istio/istio-mtls-meshwide.yaml

We should go through these and (a) confirm they are not broken and (b) perhaps organize these in sub-folders by istio version. For example, we should create "hack/istio/bookinfo-1.9" that contains bookinfo resources that work with Istio 1.9. Similarly we should have 1.10 directories and put any new resources that only work with 1.10+, etc. etc. Or just create a subfolder "bookinfo" and put all of these in there so as not to crowd the hack/istio directory with potentially obsolete files.

Side note - part of this clean up should look at the build-istio.sh and see if that is worth keeping - I suspect we'll want to just delete that. I am thinking that no longer works.

jshaughn:
@jmazzitelli I'm not even sure these are needed any longer, as it's likely preferable to use the demo repo content for various test scenarios.  If there is something unique in these yamls then maybe we should just migrate it to one of our other demos.

jmazzitelli:
@jshaughn  do you know if one of our demo have things like service entries and mLTS settings on/off, delays/fault injection/aborts, mirroring? If so we can remove these. If not, we'll want to maintain these (even if we just move them over to the demos repo).

PR #4088, opened 2021-06-09
clean up some bookinfo hack yaml

jmazzitelli:
delete some obsolete ones, put some in new bookinfo-traffic dir. This is just to be able to quickly get Kiali to show things like long request delays or fault-injected errors or mirroring.

also updates the envoy filter for 1.10 in the install travel agency script.

fixes: https://github.com/kiali/kiali/issues/4079


jshaughn (approved):
LOK

Issue #3904, opened 2021-04-19
Wrong Health info at Service level


lucasponce:
In an scenario where a service have several workloads, and some of them don't have sidecars, there is a wrong health information reported as this:

![image](https://user-images.githubusercontent.com/1662329/115227640-02e4d800-a111-11eb-8b64-8ea696fe697b.png)

But there are traffic activity:

![image](https://user-images.githubusercontent.com/1662329/115227752-2ad43b80-a111-11eb-91c6-77ae1637ef43.png)

It looks like a bug.

Note that health UI is being refactored (a little bit the presentaiton) in the context of https://github.com/kiali/kiali-ui/pull/2122 but I think the bug is independent of the logic as I can see it in previous work.

lucasponce:
Co-asigned this issue to @hhovsepy and @lucasponce
This one can a good one for Hayk to catch up with the code and getting familiar with this logic.

hhovsepy:
The issue comes from this logic: "s.IstioSidecar = mPods.HasIstioSidecar()" in service.go.
Which means that if any of pods belonging to this service does not have a sidecar, then consider this service without sidecar, thus no health information can be available. This happens in both service list and details page.
Question is: Should we consider health on service where not all workloads have sidecars?

lucasponce:
> Question is: Should we consider health on service where not all workloads have sidecars?

Yes, I think if we have a service but 2 workloads in the mesh and 1 out the mesh, from "Kiali" perspective the health should be used with the mesh workloads as those will have telemetry.

lucasponce:
Closed by https://github.com/kiali/kiali/pull/4092

PR #4092, opened 2021-06-10
Service Details sidecar is now checked in workload's sidecars.

hhovsepy:
Fix for issue https://github.com/kiali/kiali/issues/3904.

When counting sidecar persistence on service via logic that all pods have sidecars of this service, then health of the service is not shown.
But service can have some workloads which have all pods with sidecars. For those workloads the health exists and should be shown for service.
The purpose of this change is to consider the workloads sidecars existence for service sidecar.

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -109,7 +109,7 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 		/** Check if Service has istioSidecar deployed */
 		mPods := models.Pods{}
 		mPods.Parse(sPods)
-		hasSidecar := mPods.HasIstioSidecar()
+		hasSidecar := mPods.HasAnyIstioSidecar()
 		/** Check if Service has the label app required by Istio */
 		_, appLabel := item.Spec.Selector[conf.IstioLabels.AppLabelName]
 		/** Check if Service has additional item icon */
@@ -265,6 +265,7 @@ func (in *SvcService) GetService(namespace, service, interval string, queryTime 
 	s := models.ServiceDetails{Workloads: wo, Health: hth, NamespaceMTLS: nsmtls, AdditionalDetails: additionalDetails}
 	s.SetService(svc)
 	s.SetPods(kubernetes.FilterPodsForEndpoints(eps, pods))
+	s.SetIstioSidecar(wo)
 	s.SetEndpoints(eps)
 	s.SetVirtualServices(vs, vsCreate, vsUpdate, vsDelete)
 	s.SetDestinationRules(dr, drCreate, drUpdate, drDelete)
--- a/models/pod.go
+++ b/models/pod.go
@@ -168,6 +168,18 @@ func (pods Pods) HasIstioSidecar() bool {
 	return true
 }
 
+// HasAnyIstioSidecar returns true if there are pods and any of pods have a sidecar
+func (pods Pods) HasAnyIstioSidecar() bool {
+	if len(pods) > 0 {
+		for _, p := range pods {
+			if p.HasIstioSidecar() {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 // HasIstioSidecar returns true if the pod has an Istio proxy sidecar
 func (pod Pod) HasIstioSidecar() bool {
 	return len(pod.IstioContainers) > 0
--- a/models/service.go
+++ b/models/service.go
@@ -107,7 +107,10 @@ func (s *ServiceDetails) SetEndpoints(eps *core_v1.Endpoints) {
 func (s *ServiceDetails) SetPods(pods []core_v1.Pod) {
 	mPods := Pods{}
 	mPods.Parse(pods)
-	s.IstioSidecar = mPods.HasIstioSidecar()
+}
+
+func (s *ServiceDetails) SetIstioSidecar(workloads WorkloadOverviews) {
+	s.IstioSidecar = workloads.HasIstioSidecar()
 }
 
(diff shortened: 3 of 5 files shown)


hhovsepy:
@lucasponce can you please take a look into this PR?

hhovsepy:
For reproducing the original issue, it requires to have a container in a pod which always fails, for this purpose I have created "hhovsepy/exit1:latest" image in dockerhub.
So  create a deployment in OCP and refer to this image.

xeviknal:
@hhovsepy I've checked the code and seems good to me. However, I am testing in my local environment and I am not able to see any difference with the master branch. In both branches I see the service details pare with a green health check at service level but not health check in each workload:

Branch (4f005e4c64dfd7e5855c329570d0f5b1e2af5a96):

Screenshot of Kiali Console (10)

Screenshot of Kiali Console (11)


Master (ae04a3e7):

Screenshot of Kiali Console (8)

Screenshot of Kiali Console (9)

business/services.go
```diff
@@ -105,13 +105,19 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 	validations := in.getServiceValidations(svcs, deployments, pods)
 	// Convert each k8s service into our model
 	for i, item := range svcs {
-		sPods := kubernetes.FilterPodsForService(&item, pods)
-		/** Check if Service has istioSidecar deployed */
-		mPods := models.Pods{}
-		mPods.Parse(sPods)
-		hasSidecar := mPods.HasIstioSidecar()
```
lucasponce, line 112: Perhaps the same goal can be achieved using the pods.
If you see the caller "buildServiceList" is invoked from the parent "GetServiceList" which already is fetching pods + deployments for validations.
I suspect that perhaps the deployments in that level can be changed per a generic workloads.
But if we don't want to refactor that in this work, perhaps the mPods.HasIstioSidecar() may have this logic to define if pods have sidecar.
Otherwise I see that in the caller we fetch pods + deployments and in the inner "buildServiceList" we fetch the workloads again that internally will pull also similar objects, even cached, but it may introduce some redundance here.

In the pods we can check if there is Istio sidecar available looking into the containers and if necessary some labeling I think.

lucasponce: This is the change (sorry @jmazzitelli for the screenshot), I guess:
image

This what we could tweak, to check if for a service, if there is a pod with an Istio sidecar could be enough to collect Service Health.

Main issue in https://github.com/kiali/kiali/issues/3904 is that when there are services with workloads in the mesh and out of the mesh, it doesn't show the "service health" (rate traffic) and if at least it should be one, that health should be shown.

As that method is used at workload level, perhaps you can add a flag or another method to differentiate, better to add a little of duplication if the logic responds to two difference scenarios (no problem, in that sense).

hhovsepy:
> @hhovsepy I've checked the code and seems good to me. However, I am testing in my local environment and I am not able to see any difference with the master branch. In both branches I see the service details pare with a green health check at service level but not health check in each workload:
Thanks for review @xeviknal . Actually the original issue was saying that in case of workloads failing pods, the Service Details health was not showing. So my fix is in Service List/Details page only, showing  health in all cases even if some workload is failing.

business/services.go
hhovsepy: I was thinking about using "mPods.HasIstioSidecar()", but figured our that in a case when workload has 1+ pods, or saying that there is not 1-to-1 mapping between workloads and pods of service, that solution can fail.
So decided to create another method for workloads sidecars.
Yes I agree that this makes additional backend call to load workloads per services but did not find a way to skip above mentioned potential problem.

lucasponce: But, why don't create a 
```
func (pods Pods) HasServiceIstioSidecar() bool {
...
}
```

Or similar with the check that if at least one pod has the sidecar, then the service can be considered that it has sidecars ?

I guess that is more or less what the PR is trying to do, or perhaps I'm missing some edge ?

hhovsepy: Added new method HasAnyIstioSidecar which checks if any of pods have sidecar. No additional backend call is done.

lucasponce (approved):
Tested.
Before:
image

After:
image

For QE:
- Create a Service with multiple Workloads (travels may work).
- Some of the Workloads may not have sidecar.
- With at least one workload with sidecar, the service should use the "traffic rate" to compute the service health, before to this PR, when there was a workload with missing sidecar, the whole service missed the health info.

hhovsepy:
Thanks @lucasponce for the approval.
For QE: this can be tested in bookinfo namespace as well.
1. just 'Disable Auto Injection' on 'reviews-v3' workload actions, and see that sidecar is not injected anymore on that workload.
2. then check the service list and details pages, and verify that 'reviews' service has health icon now instead of undefined before.
Screenshot from 2021-06-16 18-43-13
Screenshot from 2021-06-16 18-43-29

hhovsepy:
> Thanks @lucasponce for the approval.
> For QE: this can be tested in bookinfo namespace as well.
> 
>     1. just 'Disable Auto Injection' on 'reviews-v3' workload actions, and see that sidecar is not injected anymore on that workload.
> 
>     2. then check the service list and details pages, and verify that 'reviews' service has health icon now instead of undefined before.
>        Screenshot from 2021-06-16 18-43-13
>        Screenshot from 2021-06-16 18-43-29

hhovsepy:
sorry for close/reopen, it was by mistake click

prachiyadav (approved):
Screenshot from 2021-06-18 00-42-20

xeviknal (approved):
It make much more sense! LGTM!

PR #4095, opened 2021-06-11
hack script so your OpenShift cluster can authenticate with docker.io

jmazzitelli:
gets around the docker.io ratelimiting for anonymous users.


jshaughn (approved):
Does this mean faster pulls?  LOK

PR #4105, opened 2021-06-15
ci script now sends results to libera chat room.

jmazzitelli:
We lost our freenode registrations for the room and users, so we are moving over to libera for hopefully better stability and sanity within their network.
How dumb freenode is - we'll be learning about this debacle in business schools across the country for decades :)


aljesusg (approved):
LGFM

Issue #4054, opened 2021-05-29
Can not fetching the proxy-status


yunerou:
**General remarks**

I configured a simple stack istio-operator, kiali-operator and prometheus (helm-chart) on the same namespace (istio-system).
The issue occurs when navigating to 'istio config' tab in Kiali dashboard.

![image](https://user-images.githubusercontent.com/15893577/120083299-b5418f00-c102-11eb-95be-a872922e135b.png)

A brief of logs in deploy/kiali

```
2021-05-29T19:12:53Z WRN GetPodProxyStatus is failing for [namespace: default] [pod: details-v1-79f774bdb9-p85g4]: Error fetching the proxy-status in the following pods: istiod-77c5f5cd5-lrxg6: the server has asked for the client to provide credentials (get pods istiod-77c5f5cd5-lrxg6) 
2021-05-29T19:12:53Z WRN GetPodProxyStatus is failing for [namespace: default] [pod: productpage-v1-6b746f74dc-95zcr]: Error fetching the proxy-status in the following pods: istiod-77c5f5cd5-lrxg6: the server has asked for the client to provide credentials (get pods istiod-77c5f5cd5-lrxg6) 
```

**Describe the bug** **To Reproduce**
- Step1: Start a cluster as described below (all were deployed in istio-system namespace)
- Step2: Apply ["getting started"](https://istio.io/latest/docs/setup/getting-started/) of istio tutorial (Namespace: default)
```
kubectl apply -f samples/bookinfo/platform/kube/bookinfo.yaml
kubectl apply -f samples/bookinfo/networking/bookinfo-gateway.yaml
```
- Step3: Navigation to kiali-dashboard -> istio config -> can't load istio component's config
**Expected behavior**
- The result should be the same as when [disabling cache](https://github.com/kiali/kiali-operator/blob/v1.35.0/deploy/kiali/kiali_cr.yaml#L770)
```
  kubernetes_config:
    cache_enabled: false
```
![image](https://user-images.githubusercontent.com/15893577/120083876-d3a98980-c106-11eb-8a25-edd2a57699cf.png)

**a note when try to fix**
- I have tried the following ways for debugging
	-  #3494  -> config `Security Group` to open all port from anywhere
	- istioctl proxy-status 
	  ```
	  NAME                                                   CDS        LDS        EDS        RDS        ISTIOD                     VERSION
[...]

PR #4108, opened 2021-06-17
Replace PodProxy calls by plain HTTP GET calls

xeviknal:
fix https://github.com/kiali/kiali/issues/4054
needs https://github.com/kiali/helm-charts/pull/84
needs https://github.com/kiali/kiali-operator/pull/336

What do you think if we replace the usage of the sub-resource `pod/proxy` by simple HTTP Get calls?
The idea is instead of using the Kube API to send GET requests to each IstioD pods, use the associated IP of each IstioD pod and send the request. This change would remove the requirement of opening the port 8080 from control plane to the istio system.

The initial rationale of using the sub-resource was basically following the steps of the istioctl. However, tried to consume the debug endpoints at service level or using pod names, but didn't work. It seems that using pod's IPs works.

However, I don't know how robust is the usage of POD's IPs networkwise. Thinking of different cloud providers (GCP, AWS..) or kube flavours (k8s, ocp ...). Any idea?

This change would also trigger a reduction of permissions by removing the pod/proxy from the roles. Implying that no one could proxy any pod of the mesh.

What do you think?

This PR is fully working on my env. However, I put this on draft, not only for seeking opinion, but also because it should need the removal of the GetPodProxy method from the interface, remove the role binding from operator and also remove/update the kiali.io FAQ entry.

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -275,14 +275,14 @@ func (iss *IstioStatusService) getIstiodReachingCheck() (IstioComponentStatus, e
 	wg.Add(len(healthyIstiods))
 	syncChan := make(chan ComponentStatus, len(healthyIstiods))
 
-	for _, istiod := range healthyIstiods {
+	for i, istiod := range healthyIstiods {
 		go func(name, namespace string) {
 			defer wg.Done()
 			// Using the proxy method to make sure that K8s API has access to the Istio Control Plane namespace.
 			// By proxying one Istiod, we ensure that the following connection is allowed:
 			// Kiali -> K8s API (proxy) -> istiod
 			// This scenario is no obvious for private clusters (like GKE private cluster)
-			_, err := httputil.ForwardGetRequest(iss.k8s, namespace, name, 8080, 8080, "/ready")
+			_, err := iss.k8s.ForwardGetRequest(namespace, name, 15014 + i, 15021, "/healthz/ready")
 			if err != nil {
 				syncChan <- ComponentStatus{
 					Name:   name,
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -266,7 +266,7 @@ func (iss *IstioStatusService) getIstiodReachingCheck() (IstioComponentStatus, e
 
 	healthyIstiods := make([]*core_v1.Pod, 0, len(istiods))
 	for i, istiod := range istiods {
-		if istiod.Status.Phase == "Running" {
+		if istiod.Status.Phase == core_v1.PodRunning {
 			healthyIstiods = append(healthyIstiods, &istiods[i])
 		}
 	}
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -282,7 +282,7 @@ func (iss *IstioStatusService) getIstiodReachingCheck() (IstioComponentStatus, e
 			// By proxying one Istiod, we ensure that the following connection is allowed:
 			// Kiali -> K8s API (proxy) -> istiod
 			// This scenario is no obvious for private clusters (like GKE private cluster)
-			_, err := iss.k8s.GetPodProxy(namespace, name, "/ready")
+			_, err := httputil.ForwardGetRequest(iss.k8s, namespace, name, 8080, 8080, "/ready")
 			if err != nil {
 				syncChan <- ComponentStatus{
 					Name:   name,
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -275,14 +275,14 @@ func (iss *IstioStatusService) getIstiodReachingCheck() (IstioComponentStatus, e
 	wg.Add(len(healthyIstiods))
 	syncChan := make(chan ComponentStatus, len(healthyIstiods))
 
-	for i, istiod := range healthyIstiods {
+	for _, istiod := range healthyIstiods {
 		go func(name, namespace string) {
 			defer wg.Done()
 			// Using the proxy method to make sure that K8s API has access to the Istio Control Plane namespace.
 			// By proxying one Istiod, we ensure that the following connection is allowed:
 			// Kiali -> K8s API (proxy) -> istiod
 			// This scenario is no obvious for private clusters (like GKE private cluster)
-			_, err := iss.k8s.ForwardGetRequest(namespace, name, 15014 + i, 15021, "/healthz/ready")
+			_, err := iss.k8s.ForwardGetRequest(namespace, name, httputil.GetFreePort(), 8080, "/ready")
 			if err != nil {
 				syncChan <- ComponentStatus{
 					Name:   name,
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -282,7 +282,7 @@ func (iss *IstioStatusService) getIstiodReachingCheck() (IstioComponentStatus, e
 			// By proxying one Istiod, we ensure that the following connection is allowed:
 			// Kiali -> K8s API (proxy) -> istiod
 			// This scenario is no obvious for private clusters (like GKE private cluster)
-			_, err := iss.k8s.ForwardGetRequest(namespace, name, httputil.GetFreePort(), 8080, "/ready")
+			_, err := iss.k8s.ForwardGetRequest(namespace, name, httputil.Pool.GetFreePort(), 8080, "/ready")
 			if err != nil {
 				syncChan <- ComponentStatus{
 					Name:   name,
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -212,6 +212,10 @@ func (in *K8SClient) getIstiodDebugStatus(debugPath string) (map[string][]byte, 
 
 	// Check if the kube-api has proxy access to pods in the istio-system
 	// https://github.com/kiali/kiali/issues/3494#issuecomment-772486224
+	// The 8080 port is not accessible from outside of the pod. However, it is used for kubernetes to do the live probes.
+	// Using the port-forwarding, the call is made as it was in the pod itself, as a localhost call.
+	// Also the port-forwarding to a pod is done via the KubeAPI. Therefore if the call doesn't return any error,
+	// it means that Kiali has access to the KubeAPI and that the KubeAPI has access to the Istiod (control plane).
 	_, err = in.ForwardGetRequest(c.IstioNamespace, istiods[0].Name, httputil.GetFreePort(), 8080, "/ready")
 	if err != nil {
 		return nil, fmt.Errorf("unable to proxy Istiod pods. " +
@@ -228,6 +232,9 @@ func (in *K8SClient) getIstiodDebugStatus(debugPath string) (map[string][]byte, 
 		go func(name, namespace string) {
 			defer wg.Done()
 
+			// The 15014 port on Istiod is open for control plane monitoring.
+			// Here's the Istio doc page about the port usage by istio:
+			// https://istio.io/latest/docs/ops/deployment/requirements/#ports-used-by-istio
 			res, err := in.ForwardGetRequest(namespace, name, httputil.GetFreePort(), 15014, debugPath)
 			if err != nil {
 				errChan <- fmt.Errorf("%s: %s", name, err.Error())
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -211,7 +211,7 @@ func (in *K8SClient) getIstiodDebugStatus(debugPath string) (map[string][]byte, 
 
 	// Check if the kube-api has proxy access to pods in the istio-system
 	// https://github.com/kiali/kiali/issues/3494#issuecomment-772486224
-	_, err = in.ForwardRequestToPod(c.IstioNamespace, istiods[0].Name, 8080, 8080, "/ready")
+	_, err = in.ForwardGetRequest(c.IstioNamespace, istiods[0].Name, 15014, 15014, "/healthz/ready")
 	if err != nil {
 		return nil, fmt.Errorf("unable to proxy Istiod pods. " +
 			"Make sure your Kubernetes API server has access to the Istio control plane through 8080 port")
@@ -221,14 +221,14 @@ func (in *K8SClient) getIstiodDebugStatus(debugPath string) (map[string][]byte, 
 	wg.Add(len(healthyIstiods))
 	errChan := make(chan error, len(healthyIstiods))
 	syncChan := make(chan map[string][]byte, len(healthyIstiods))
+	envoyLocalPort := config.Get().ExternalServices.Istio.EnvoyAdminLocalPort
 
 	result := map[string][]byte{}
-	for _, istiod := range healthyIstiods {
+	for i, istiod := range healthyIstiods {
 		go func(name, namespace string) {
 			defer wg.Done()
 
-			envoyLocalPort := config.Get().ExternalServices.Istio.EnvoyAdminLocalPort
-			res, err := in.ForwardRequestToPod(namespace, name, envoyLocalPort, 15014, debugPath)
+			res, err := in.ForwardGetRequest(namespace, name, envoyLocalPort + i, 15014, debugPath)
 			if err != nil {
 				errChan <- fmt.Errorf("%s: %s", name, err.Error())
 			} else {
@@ -316,10 +316,7 @@ func parseRegistryServices(registries map[string][]byte) ([]*RegistryStatus, err
 func (in *K8SClient) GetConfigDump(namespace, podName string) (*ConfigDump, error) {
 	// Fetching the config_dump data, raw.
 	envoyLocalPort := config.Get().ExternalServices.Istio.EnvoyAdminLocalPort
-	// TODO: need to remove envoy local port -> fetch available port dynamically
-	// Scenario: Two users consuming kiali at the same time. The pod will open the same port. One request will crash.
-	// However, there is a cache in place. So this is likely to happen
-	resp, err := in.ForwardRequestToPod(namespace, podName, envoyLocalPort, 15014, "/config_dump")
+	resp, err := in.ForwardGetRequest(namespace, podName, envoyLocalPort, 15014, "/config_dump")
 	if err != nil {
 		log.Errorf("Error fetching config_map: %v", err)
 		return nil, err
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -321,7 +321,11 @@ func parseRegistryServices(registries map[string][]byte) ([]*RegistryStatus, err
 }
 
 func (in *K8SClient) GetConfigDump(namespace, podName string) (*ConfigDump, error) {
-	// Fetching the config_dump data, raw.
+	// Fetching the Config Dump from the pod's Envoy.
+	// The port 15000 is open on each Envoy Sidecar (managed by Istio) to serve the Envoy Admin  interface.
+	// This port can only be accessed by inside the pod.
+	// See the Istio's doc page about its port usage:
+	// https://istio.io/latest/docs/ops/deployment/requirements/#ports-used-by-istio
 	resp, err := in.ForwardGetRequest(namespace, podName, httputil.Pool.GetFreePort(), 15000, "/config_dump")
 	if err != nil {
 		log.Errorf("Error forwarding the /config_dump request: %v", err)
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -322,9 +322,9 @@ func parseRegistryServices(registries map[string][]byte) ([]*RegistryStatus, err
 
 func (in *K8SClient) GetConfigDump(namespace, podName string) (*ConfigDump, error) {
 	// Fetching the config_dump data, raw.
-	resp, err := in.ForwardGetRequest(namespace, podName, httputil.GetFreePort(), 15014, "/config_dump")
+	resp, err := in.ForwardGetRequest(namespace, podName, httputil.GetFreePort(), 15000, "/config_dump")
 	if err != nil {
-		log.Errorf("Error fetching config_map: %v", err)
+		log.Errorf("Error forwarding the /config_dump request: %v", err)
 		return nil, err
 	}
 
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -1,14 +1,12 @@
 package kubernetes
 
 import (
-	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
 	"regexp"
 	"strings"
 	"sync"
-	"time"
 
 	"gopkg.in/yaml.v2"
 	core_v1 "k8s.io/api/core/v1"
@@ -20,8 +18,6 @@ import (
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/log"
-	"github.com/kiali/kiali/util/config_dump"
-	"github.com/kiali/kiali/util/httputil"
 )
 
 var (
@@ -215,7 +211,7 @@ func (in *K8SClient) getIstiodDebugStatus(debugPath string) (map[string][]byte, 
 
 	// Check if the kube-api has proxy access to pods in the istio-system
 	// https://github.com/kiali/kiali/issues/3494#issuecomment-772486224
-	_, err = in.GetPodProxy(c.IstioNamespace, istiods[0].Name, "/ready")
+	_, err = in.ForwardRequestToPod(c.IstioNamespace, istiods[0].Name, 8080, 8080, "/ready")
 	if err != nil {
 		return nil, fmt.Errorf("unable to proxy Istiod pods. " +
 			"Make sure your Kubernetes API server has access to the Istio control plane through 8080 port")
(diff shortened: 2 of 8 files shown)


lucasponce:
Why don't do a ProxyGet ?

https://github.com/istio/istio/blob/f38465f8f5c1edd944ed7776c1253ffa2d9768b2/pkg/kube/client.go#L649

But in the other hand not sure if this is we are aiming the problem in the right direction.

I mean, if "istioctl" works for a user, "kiali" should work for this request, and viceversa, so in that sense, when we have this akward problems the easy we have to debug and test, the better.

More than tweaking the code, I'd like to try to understand the root cause, to check if "istioctl" works or not, and also use it in Kiali to debug when there can be connectity problems.

xeviknal:
> Why don't do a ProxyGet ?

If I understood that correctly, the ProxyGet method is still using the proxy sub-resource. It is a specific implementation. Therefore, it doesn't remove the dependency to the KubeAPI. Check it out the source code:
https://github.com/kubernetes/client-go/blob/v0.21.1/kubernetes/typed/core/v1/pod_expansion.go#L53
Plus, the `ProxyGet` method wasn't in place in our kube client version when this was implemented. The method was introduced in `v0.19` and we had version `v11.0.1`

> if "istioctl" works for a user, "kiali" should work for this request, and viceversa

The main difference between `istioctl` and `kiali` is that (in the current supported deployment models) `kiali` runs inside the cluster whereas the `istioctl` is always outside the cluster. 
While being outside the cluster the only way to communicate at pod level from external machines to the cluster is using either a `port-forward` (GET|POST|PUT|* requests) or using the `pods/proxy` sub-resource (only GET requests).

> the easy we have to debug and test, the better

This is a matter of messaging I believe. Saying: "make sure kiali pod can connect with istiod-292318-asux pod using the 172.17.0.17 at port 15014".

But again, I don't know the implications this change has at network level. So I don't know how safe or how many more implication this would have. Nevertheless I think that ALL pods must have network access to IstioD pods otherwise the control plane couldn't control a thing.

lucasponce:
Right

image

So, in short any workload connected to the mesh should be able to communicate to the 15014.

It sounds we could go ahead with this proposal then.

lucasponce:
But if the port 15014 is the required one, then we should use the service, not the pod IP.

xeviknal:
> then we should use the service, not the pod IP

That is the thing. If we were consuming the service, there wouldn't have needed the usage of the Pod Proxy. Consuming the service would lead you to inconsistent data when the control plane would have more than one IstioD pod. Each request to the service returns only a response from one single pod and it is not possible to control to which pod you want a response from. So definitely, consuming service endpoints is not part of the solution.

I am not very versed in the kubernetes network model, but as far as I understood pods on a node can communicate with all pods on all nodes[1]. Do you guys see any major blocker using the pod's IP?

> 15014 is the required one

Yep, we have to change the port number from 8080 to 15014.

[1]

lucasponce:
>  Consuming the service would lead you to inconsistent data when the control plane would have more than one IstioD pod.

I'm not sure about this.

I can understand that we have two scenarios here:

a) we consume the information from IstioD like other user, then I don't see a reason we can't use the service.
b) we can to check the health check of the IstioD and yes, we can check if the workload/pods are correct.

Perhaps just syncing if the pods of the istiod are not synced doesn't need that kiali bypass the service.

xeviknal:
See the example:

I am running a basic istio deployment (istio-system with demo profile) and basic bookinfo (bookinfo ns). There are two IstioD pods though.

```
$ kubectl get pods -l app=istiod -n istio-system
istiod-8546f98dcb-pncnm   1/1     Running   7          9d
istiod-8546f98dcb-qwhzx   1/1     Running   3          3d3h
```

Port-forwarding to both pods:
```
$ kubectl port-forward -n istio-system istiod-8546f98dcb-qwhzx 15014
[...]
$ curl http://localhost:15014/debug/syncz
[
{
"proxy": "reviews-v2-7bf8c9648f-5hs6s.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=4f54250a-dab6-4ece-a443-751028b55b09",
"cluster_acked": "C6JqN6wI4GU=4f54250a-dab6-4ece-a443-751028b55b09",
...
},
{
"proxy": "ratings-v1-b6994bb9-d9hnn.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=99960760-f79f-4abe-a671-511bdfc327eb",
"cluster_acked": "C6JqN6wI4GU=99960760-f79f-4abe-a671-511bdfc327eb",
...
},
{
"proxy": "reviews-v1-545db77b95-vnl78.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=5a7657b1-9f6c-4f85-939e-bc53b02cd277",
"cluster_acked": "C6JqN6wI4GU=5a7657b1-9f6c-4f85-939e-bc53b02cd277",
...
},
{
"proxy": "details-v1-79f774bdb9-mtvsz.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=468f47b7-42b6-4e90-bf35-d5166dfbaa7b",
"cluster_acked": "C6JqN6wI4GU=468f47b7-42b6-4e90-bf35-d5166dfbaa7b",
...
},
{
"proxy": "istio-ingressgateway-8657768d87-sm2fw.istio-system",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=1a1b48fb-a318-407d-903c-066063fdc7ee",
"cluster_acked": "C6JqN6wI4GU=1a1b48fb-a318-407d-903c-066063fdc7ee",
...
},
{
"proxy": "kiali-traffic-generator-kk5lh.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=268ec259-f7fe-41a6-80ab-b280b7493718",
"cluster_acked": "C6JqN6wI4GU=268ec259-f7fe-41a6-80ab-b280b7493718",
...
}
]
```

Let's port-forward the other istiod:

```
$ kubectl port-forward -n istio-system istiod-8546f98dcb-pncnm 15014 
[...]
$ curl http://localhost:15014/debug/sync
[
{
"proxy": "istio-egressgateway-585f7668fc-sbcbv.istio-system",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=73593a3f-4ebb-4a94-abd0-3058ebb8f066",
"cluster_acked": "C6JqN6wI4GU=73593a3f-4ebb-4a94-abd0-3058ebb8f066",
...
},
{
"proxy": "ratings-v2-59658c8774-llwc4.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=1a65034a-1eb7-40b2-a4cc-b37bf898b863",
"cluster_acked": "C6JqN6wI4GU=1a65034a-1eb7-40b2-a4cc-b37bf898b863",
...
},
{
"proxy": "reviews-v3-84779c7bbc-45jsb.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=eeff9dad-797c-40d1-9546-da525889901d",
"cluster_acked": "C6JqN6wI4GU=eeff9dad-797c-40d1-9546-da525889901d",
...
},
{
"proxy": "mongodb-v1-64d4666575-qrtth.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=7d96e8ca-fa99-4fa1-bc0a-1c12681c72a5",
"cluster_acked": "C6JqN6wI4GU=7d96e8ca-fa99-4fa1-bc0a-1c12681c72a5",
...
}
]
```

If you can see, both `/debug/syncz` calls return different sync information. Each IstioD tracks the sync of its controlled workloads. IstioDs don't sync this info between them.

Let's see the case of fetching info straight from the service:

```
# connecting from the kiali pod for convenience
$ k exec -n istio-system --stdin --tty kiali-54f8c76f96-t6kfg -- /bin/bash
$ curl http://localhost:15014/debug/sync
[
{
"proxy": "reviews-v2-7bf8c9648f-5hs6s.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=4f54250a-dab6-4ece-a443-751028b55b09",
"cluster_acked": "C6JqN6wI4GU=4f54250a-dab6-4ece-a443-751028b55b09",
...
},
{
"proxy": "ratings-v1-b6994bb9-d9hnn.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=99960760-f79f-4abe-a671-511bdfc327eb",
"cluster_acked": "C6JqN6wI4GU=99960760-f79f-4abe-a671-511bdfc327eb",
...
},
{
"proxy": "reviews-v1-545db77b95-vnl78.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=5a7657b1-9f6c-4f85-939e-bc53b02cd277",
"cluster_acked": "C6JqN6wI4GU=5a7657b1-9f6c-4f85-939e-bc53b02cd277",
...
},
{
"proxy": "details-v1-79f774bdb9-mtvsz.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=468f47b7-42b6-4e90-bf35-d5166dfbaa7b",
"cluster_acked": "C6JqN6wI4GU=468f47b7-42b6-4e90-bf35-d5166dfbaa7b",
...
},
{
"proxy": "istio-ingressgateway-8657768d87-sm2fw.istio-system",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=1a1b48fb-a318-407d-903c-066063fdc7ee",
"cluster_acked": "C6JqN6wI4GU=1a1b48fb-a318-407d-903c-066063fdc7ee",
...
},
{
"proxy": "kiali-traffic-generator-kk5lh.bookinfo",
"istio_version": "1.10.0",
"cluster_sent": "C6JqN6wI4GU=268ec259-f7fe-41a6-80ab-b280b7493718",
"cluster_acked": "C6JqN6wI4GU=268ec259-f7fe-41a6-80ab-b280b7493718",
...
}
]
```

If you perform the curl different time you eventually get the response from the other istiod. But not in a deterministic fashion.
Do I miss something? I think the only way to have the full sync info is just fetch it from pods (as it is done in the istioctl).

lucasponce:
Ok @xeviknal, then I guess what's going on here:
- The istiod has state that it's synced across IstioD replicas, otherwise dataplane would be inconsistent (i.e. different registries in the envoy clusters, it doesn't make sense that would happen).
- The Proxy Status seems to be supported by informers (a.k.a. similar cache used in kiali) that is purely async, then yes, your comment is right, for that service (that is not used by the data plane, but for istioctl / kiali) it needs to manually collect the status from all pods.

For some reason I was taking the assumption of both being internally sync and that triggered my discussion.

business/istio_status.go
```diff
@@ -271,26 +271,35 @@ func (iss *IstioStatusService) getIstiodReachingCheck() (IstioComponentStatus, e
...
+	for i, istiod := range healthyIstiods {
+		go func(name, namespace, ip string) {
 			defer wg.Done()
 			// Using the proxy method to make sure that K8s API has access to the Istio Control Plane namespace.
 			// By proxying one Istiod, we ensure that the following connection is allowed:
 			// Kiali -> K8s API (proxy) -> istiod
 			// This scenario is no obvious for private clusters (like GKE private cluster)
-			_, err := iss.k8s.GetPodProxy(namespace, name, "/ready")
-			if err != nil {
+			debugURI := fmt.Sprintf("http://%s:8080/ready", ip)
```
lucasponce, line 293: We should discuss this.
Port 8080 is deprecated according to [1]

[1] https://istio.io/latest/docs/ops/deployment/requirements/#ports-used-by-istio

lucasponce: Just looking at the endpoint:

```
$ kubectl get endpoints/istiod -n istio-system -o yaml
apiVersion: v1
kind: Endpoints
metadata:
  annotations:
    endpoints.kubernetes.io/last-change-trigger-time: "2021-06-22T07:27:14Z"
  creationTimestamp: "2021-06-22T07:27:07Z"
  labels:
    app: istiod
    install.operator.istio.io/owning-resource: unknown
    install.operator.istio.io/owning-resource-namespace: istio-system
    istio: pilot
    istio.io/rev: default
    operator.istio.io/component: Pilot
    operator.istio.io/managed: Reconcile
    operator.istio.io/version: 1.10.1
    release: istio
  name: istiod
  namespace: istio-system
  resourceVersion: "608"
  uid: ce94b645-2fe2-4ffa-b49a-6280fb5000d7
subsets:
- addresses:
  - ip: 172.17.0.5
    nodeName: minikube
    targetRef:
      kind: Pod
      name: istiod-85c958cd6-kcfk9
      namespace: istio-system
      resourceVersion: "606"
      uid: f52827e2-2726-4846-ac22-943720e83452
  ports:
  - name: https-dns
    port: 15012
    protocol: TCP
  - name: grpc-xds
    port: 15010
    protocol: TCP
  - name: https-webhook
    port: 15017
    protocol: TCP
  - name: http-monitoring
    port: 15014
    protocol: TCP

```

xeviknal: I know. However, I haven't found the way to re-use the same readiness probes used for kubernetes to manage pod's life-cycle. It looks like is deprecated for the debug endpoints (/debug/[syncz|registryz|...]), but they still use it for readiness probes. I couldn't find another way to fetch that endpoint otherwise.

We could use another endpoint to consume from the 15014 port. However, I don't know if it would be as light and accurate as this one.

```
$ kubectl get pod -n istio-system istiod-12312882 -o yaml
apiVersion: v1
kind: Pod
metadata:
  name: istiod-8546f98dcb-qwhzx
  namespace: istio-system
  annotations:
    prometheus.io/port: "15014"
    prometheus.io/scrape: "true"
    sidecar.istio.io/inject: "false"
...
    readinessProbe:
      failureThreshold: 3
      httpGet:
        path: /ready
        port: 8080
        scheme: HTTP
      initialDelaySeconds: 1
      periodSeconds: 3
      successThreshold: 1
      timeoutSeconds: 5
....
```

xeviknal:
After giving some thoughts around this, we've decided to keep all the comms between kiali and istiod(s)/sidecar(s) through the KubeAPI and using the `port-forwarding` sub-resource. Therefore, this PR is slightly changing the initial direction.

The main issues this PR is approaching are:
1) Removing the usage of deprecated ports in both `proxy-status` and `envoy info` endpoints[1]. This makes some ports stop using the 8080 port and therefore the `pod/proxy` subresource is not working anymore.
2) Moving the consumption of the `istiod`s healthz endpoint at 8080 from `pod/proxy` to `pod/portforward`.
3) Supporting the [new authentication](https://github.com/kiali/kiali/issues/4132) protection that `istiod(s)` now have in place.

@howardjohn I've built and deployed Istio from master and seems that the error you reported is not happening anymore. Is using the port-forwarding like I do in this PR something similar to [what you mentioned](https://github.com/kiali/kiali/issues/4132#issuecomment-870741780).

#### Caveats ####

There is a small trick for port assignment. This solution implies that the kiali pod will have reserved 100 ports (more than enough) for starting all the port-forwarding: from 15000 to 15099.The problem behind this is that kiali starts multiple async requests which need to start port-forwarding (only for filling the cache, not at every request ever). Therefore, each request should use one available port.I found three different strategies:

1) Manually assign serial port numbers to each request: try to hard-code the local port numbers in each call. Not picked: really error-prone. Two simultaneous requests would fail. Some calls are called in a loop, so they need multiple ports. There are two endpoints that fetch Istiods almost at the same time => port clash.
2) Use the 0 port to let the SO pick one port for you. Not picked: it might introduce a bit of latency for resolving the issue (I haven't tested it though).
3) Reserve a pool of ports to async assign to each port-forward call. Picked: easy to deal with concurrency, no network latency added resolving the port.

[1] https://istio.io/latest/docs/ops/deployment/requirements/#ports-used-by-istio

howardjohn:
Just to clarify if you use port forward you do not need auth. Auth is required only from non-localhost calls; port-forward comes from localhost.

If you ever expect to have an Istio sidecar in front of Kiali, which I think some users do, I would recommend avoiding the 150xx port  range

lucasponce:
I added @prachiyadav to the thread as this PR is still under Dev review but it will require a full regression in whole Kiali.

kubernetes/istio.go
```diff
@@ -215,7 +212,7 @@ func (in *K8SClient) getIstiodDebugStatus(debugPath string) (map[string][]byte,
 
 	// Check if the kube-api has proxy access to pods in the istio-system
 	// https://github.com/kiali/kiali/issues/3494#issuecomment-772486224
-	_, err = in.GetPodProxy(c.IstioNamespace, istiods[0].Name, "/ready")
+	_, err = in.ForwardGetRequest(c.IstioNamespace, istiods[0].Name, httputil.GetFreePort(), 8080, "/ready")
```
lucasponce, line 215: Code suggestion.
It would be nice if we add a note on the destination port, what is used, etc.
That may help to identify in the future if there is a change, and also I'm just thinking for us.
For example, we've reviewed that the 8080 would be deprecated for one type of service but it's still used for readiness.

kubernetes/istio.go
```diff
@@ -231,7 +228,7 @@ func (in *K8SClient) getIstiodDebugStatus(debugPath string) (map[string][]byte,
 		go func(name, namespace string) {
 			defer wg.Done()
 
-			res, err := in.GetPodProxy(namespace, name, debugPath)
+			res, err := in.ForwardGetRequest(namespace, name, httputil.GetFreePort(), 15014, debugPath)
```
lucasponce, line 231: Here as well, the 15014 is well known when you are dealing with these ports, but a comment would help (us) to follow the whole call.

kubernetes/kubernetes.go
```diff
@@ -326,6 +352,34 @@ func (in *K8SClient) GetPods(namespace, labelSelector string) ([]core_v1.Pod, er
...
+	writer := new(bytes.Buffer)
+
+	clientConfig, err := ConfigClient()
+	if err != nil {
+		log.Errorf("Error getting Kubernetes Client config: %v", err)
+		return nil, err
+	}
+
+	// First try whether the pod exist or not
+	_, err = in.GetPod(namespace, name)
```
lucasponce, line 371: Perhaps worth here to fail on the pod phase [1]

image

That may save some cycles on initializations and in case something is odd, to not propagate the timeout of the http.

[1] https://github.com/kubernetes/kubernetes/blob/v1.6.0-alpha.0/pkg/kubectl/cmd/portforward.go#L107

util/httputil/port_pool.go
```diff
@@ -0,0 +1,66 @@
...
+	"fmt"
+	"sync"
+)
+
+// [1500] => true // busy
+// [1501] => true // busy
+// [1502] => false // free
+// last port assigned => 1501
+var portRangeInit = 14100
+var portRangeSize = 100
```
lucasponce, line 13: I have two comments.
One is that this info should be confiburable, as user may want to tweak, 99.9% the defaults should be ok, but better to have it available as we have with other connection parameters.

Another idea is we need to think in the concurrency scenarios, in two ways, when a single user access to a high populated cluster and when there are multiple users accesing Kiali.

Probably nothing to change that just to calculate in which certain situations this may fail, i.e. we fix some concurrency issue in a very big cluster for a similar case.

lucasponce: I think that using the portforward is mostly on fetching proxy status, control plane status, and fetching envoy details.
So, per request it should be a few, and then the limit would be in the number of users that may access to Kiali, but just wanted to check if I missed some case (i.e. when fetching a list of workloads) that may impact when number is > 100.

lucasponce:
I've started to review the PR, thanks @xeviknal.
From my environment I have some problems:

image

```
2021-07-08T10:29:45+02:00 INF Kiali cache for [namespace: default] started
E0708 11:21:48.259608  125588 portforward.go:400] an error occurred forwarding 14196 -> 15014: error forwarding port 15014 to pod 9551fbd7bb6466de2016bbd721ea6c69772847ec2dc036d8d81c152f10e7394a, uid : exit status 1: 2021/07/08 09:21:48 socat[76240] E connect(5, AF=2 127.0.0.1:15014, 16): Connection refused
2021-07-08T11:21:48+02:00 ERR Error fetching config_map: Get "http://localhost:14196/config_dump": EOF
2021-07-08T11:21:48+02:00 ERR Get "http://localhost:14196/config_dump": EOF
E0708 11:21:51.631412  125588 portforward.go:400] an error occurred forwarding 14197 -> 15014: error forwarding port 15014 to pod 9551fbd7bb6466de2016bbd721ea6c69772847ec2dc036d8d81c152f10e7394a, uid : exit status 1: 2021/07/08 09:21:51 socat[76248] E connect(5, AF=2 127.0.0.1:15014, 16): Connection refused
2021-07-08T11:21:51+02:00 ERR Error fetching config_map: Get "http://localhost:14197/config_dump": EOF
2021-07-08T11:21:51+02:00 ERR Get "http://localhost:14197/config_dump": EOF
E0708 11:21:52.601286  125588 portforward.go:400] an error occurred forwarding 14198 -> 15014: error forwarding port 15014 to pod 9551fbd7bb6466de2016bbd721ea6c69772847ec2dc036d8d81c152f10e7394a, uid : exit status 1: 2021/07/08 09:21:52 socat[76253] E connect(5, AF=2 127.0.0.1:15014, 16): Connection refused
2021-07-08T11:21:52+02:00 ERR Error fetching config_map: Get "http://localhost:14198/config_dump": EOF
2021-07-08T11:21:52+02:00 ERR Get "http://localhost:14198/config_dump": EOF
```

cc @prachiyadav this PR is still under development, so just keep an eye, but not yet ready for QE.

util/httputil/port_pool.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"fmt"
+	"sync"
+)
+
+// [1500] => true // busy
+// [1501] => true // busy
+// [1502] => false // free
+// last port assigned => 1501
```
jmazzitelli, line 11: I don't understand what these comments are telling me.

util/httputil/port_pool.go
xeviknal: I have pros and cos of adding the pool params into settings. 
Will other processes in the kiali pod open ports in the same port space? Won't Kiali serve will always be the only process opening ports in the kiali pod?

xeviknal: And agree with the second message. The port range size limits the number of users hitting the same button at the same time. However, it is pretty unlikely to reach the number of 100. Let's list the number of calls done and how they are cache:

1) Istio Component status - One port open per each IstioD pod (1-N). This request is cached and it is regenerated every X seconds. (Extreme scenario: 100 users send the same request at the same precise moment and the server processes the request at the same moment and the cache is not generated/expired).
2) Service Registry status - One port for each IstioD pod (1-N). Same as before. Content cached. There should be one request after the cache is expired.
3) Proxy Status - One port for each IstioD pod (1-N). Same as before. Content is cached. There should be one request after the cache is expired.
4) Envoy Config endpoints: One port per user click in the envoy modal. Therefore, if kiali is used with high concurrency, this might reach the limit soon. We might consider applying a short cache for this. However, this cache could be huge because is linearly depending on how many workloads you have in the mesh.


We could improve this pool mechanism for high saturated networks where the calls to the sidecars or Istiods takes a lot of time. User starts to send requests. Those requests are stacked/queued because of waiting for istiod/sidecar to respond. The pool port is empty. Then Kiali starts erroring that there are no available ports.

I don't think there are istio meshes with 25 istiod pods managing the mesh. This would mean that only there is the possibility of 25 users clicking the envoy tab at the same time. The number 26th would get an "port is already is use".
100 ports = 25 ports for Istio Component status + 25 ports for Service Registy + 25 ports for Proxy Status + 25 concurrent users clicking the envoy tab.

jmazzitelli: Is there a reason you use 100 there? Ports are cheap, no? Use 1,000 and forget about it :)

lucasponce: I support @jmazzitelli comment :) it's safe to put a gib number in the ports as soon as we control that range and probably we can revisit this in the context of high scalability scenarios (where we are not yet there).

xeviknal:
@lucasponce @jmazzitelli I think this is quite stable now. Do you mind doing another look at it? Remember that there is also a change in the role (removing the Pod Proxy role).

lucasponce:
The PR is still not working on my dev env.
Now it looks there is a timeout and some missing logs in the backend log.
I'll review in detail.

lucasponce:
@xeviknal I'm having problems to make it work with my env, could you rebase the PR with backend master ?
I think some issues I'm seeing could be related to out-of-sync between backend/frontend.

Thanks

xeviknal:
@lucasponce rebase done and pushed. I've done a quick smoke test and seemed good on my in-cluster deployment

kubernetes/istio.go
```diff
@@ -318,9 +322,9 @@ func parseRegistryServices(registries map[string][]byte) ([]*RegistryStatus, err
 
 func (in *K8SClient) GetConfigDump(namespace, podName string) (*ConfigDump, error) {
 	// Fetching the config_dump data, raw.
-	resp, err := in.EnvoyForward(namespace, podName, "/config_dump")
+	resp, err := in.ForwardGetRequest(namespace, podName, httputil.Pool.GetFreePort(), 15000, "/config_dump")
```
lucasponce, line 325: I'd add a similar comment as used for port 15014, with the links of the ports used.

lucasponce (approved):
LGTM
I minimal comment to update a comment that can be addressed on this PR or next one.
Also I've tested locally from a out-cluster configuration and no regressions.

I'd suggest @prachiyadav to run some full regression testing on this, as typically QE environments are remote and with more network latency, if there is any suspicius problem that should help to find it.

xeviknal:
Thank you @lucasponce for thorough review and make sure that this PR is fully aligned with part of the vision of the project.
I've also added the comment you mention. I see that GH has dismissed your review. Do you mind adding the last check?

lucasponce (approved):

prachiyadav (approved):

Issue #4050, opened 2021-05-27
Support for custom CA certificates in OpenID authentication


AbrohamLincoln:
**Is your feature request related to a problem? Please describe.**
OpenID authentication does not support custom CA certificates.  The only recourse is to disable TLS verification.

**Describe the solution you'd like**
Add the ability to add in a custom root CA to the OpenID config.

Do something similar to [this](https://github.com/kiali/kiali/blob/master/business/openshift_oauth.go#L248-L257) in the [OpenID http client](https://github.com/kiali/kiali/blob/5bcdd2f372ff2fcca4746e321bd9e93e677e1b0c/business/openid_auth.go#L778)

**Describe alternatives you've considered**
Supporting custom CAs could potentially solve this as well: https://github.com/kiali/kiali/issues/1675

jshaughn:
cc @israel-hdez

israel-hdez:
@AbrohamLincoln I wonder if you can workaround this requirement by building a customized Kiali image with the needed certificates?

For #1675, Kiali would need a writable file-system in order to update the trusted certificate chains. We [recently added security constraints that run Kiali with a read-only file system](https://github.com/kiali/kiali-operator/blob/f0c07d1323d67b392d29e46073d81e14d032495e/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml#L54-L58). So, #1675 is blocked for now.

However, if you are able to build and use a customized Kiali image, additional certificates can be added. A sample Dockerfile is (I haven't tried, but this should be a good start point):

```Dockerfile
FROM quay.io/kiali/kiali:latest

USER root

COPY your_ca_file /etc/pki/ca-trust/source/anchors/
RUN /bin/update-ca-trust

USER 1000
```

I would like to avoid a new ca_file option, because we have such options for almost every single service that Kiali talks to (see, for example, [this option for prometheus](https://github.com/kiali/kiali-operator/blob/f0c07d1323d67b392d29e46073d81e14d032495e/deploy/kiali/kiali_cr.yaml#L588-L590)). I think this way is simply adding too much config properties and also support in the code is required. So, I want to avoid a new one. Adding the needed certs to the image is, IMO, a cleaner solution.

If you can't use a custom image, I think we can do the effort of adding yet another one `ca_file` setting. Similarly to the OpenShift auth, only a single cert file would be supported.

AbrohamLincoln:
Custom images are technically possible, but more or less a non-starter policy wise in our environment.

[...]

PR #4112, opened 2021-06-17
Support for custom CA certificate for OpenID auth

israel-hdez:
Adding support for custom CA certificate for the OpenID authentication.

The CA certificate is provided through a volume mount. Thus, this checks for existence of the certificate file and uses it, if available.

Related operator PR: https://github.com/kiali/kiali-operator/pull/333

Related #4050

### DIFF
--- a/business/openid_auth.go
+++ b/business/openid_auth.go
@@ -13,6 +13,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -29,6 +30,8 @@ import (
 
 const (
 	OpenIdNonceCookieName = config.TokenCookieName + "-openid-nonce"
+
+	OpenIdServerCAFile = "/kiali-cabundle/openid-server-ca.crt"
 
 	// Maximum size of session cookies. This is 3.5K.
 	// Major browsers limit cookie size to 4K, but this includes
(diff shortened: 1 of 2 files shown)


business/openid_auth.go
```diff
@@ -783,10 +787,30 @@ func createHttpClient(toUrl string) (*http.Client, error) {
...
+			certPool = x509.NewCertPool()
+			if ok := certPool.AppendCertsFromPEM(cafile); !ok {
+				return nil, fmt.Errorf("supplied OpenId CA file cannot be parsed")
+			}
+		}
+
 		httpTransport.TLSClientConfig = &tls.Config{
-			InsecureSkipVerify: true,
+			InsecureSkipVerify: cfg.InsecureSkipVerifyTLS,
+			RootCAs:            certPool,
```
jmazzitelli, line 813: In the if-stmt on 802, I see we get in this if-clause if "InsecureSkipVerifyTLS" is true **even if** cafile == nil.

If cafile is nil, that means certPool is never set (thus is "nil" pointer). When you go to set this RootCAs field, it is getting set to a nil point.

Is that an error? That seems like it could be a problem. Have you validated that code path? i.e. set InsecureSkipVerifyTLS to true and set cafile to nil

kubernetes/types.go
```diff
@@ -1,10 +1,11 @@
 package kubernetes
 
 import (
+	"time"
+
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"time"
```
jmazzitelli, line 7: This change was already done - see https://github.com/kiali/kiali/blob/master/kubernetes/types.go#L4-L8

Looks like this PR needs a rebase.

business/iter8.go
```diff
@@ -2,14 +2,14 @@ package business
...
 	"encoding/json"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"regexp"
 	"strconv"
 	"strings"
 	"sync"
 
 	"gopkg.in/yaml.v2"
 	core_v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
```
jmazzitelli, line 12: This change was already done - see https://github.com/kiali/kiali/blob/master/business/iter8.go#L4-L12

Looks like this PR needs a rebase.

jmazzitelli (commented):
@israel-hdez My main concern is that nil pointer "certPool". Can you confirm that is not a problem?

The rebase isn't necessary since there are no conflicts, but wouldn't be a bad thing to do.

business/openid_auth.go
israel-hdez: It is similar to what the HTTP util is doing.
See: https://github.com/kiali/kiali/blob/3e032198c053f4cc9565c7a8f0421f51561b31f9/util/httputil/http_util.go#L103-L124

So, I think it shouldn't fail. But I suspect it may have an unwanted behavior. Let me double check.

israel-hdez: It does not fail with the nil value.

jmazzitelli (approved):
lgtm

Issue #4070, opened 2021-06-04
gRPC Streaming traffic


emcake:
**Is your feature request related to a problem? Please describe.**
We run a large number of gRPC bidi streaming services talking to each other over long-lived streams, and wanted to use Kiali for observability. At the minute this doesn't show up in the metrics that kiali uses to show gRPC traffic. We see one request on completion where this may consist of many hundreds of messages.

**Describe the solution you'd like**
Istio concurrently exposes message stats under `istio_request_messages_total` and `istio_response_messages_total`, and some rudimentary prometheus querig looks like they would produce suitable metrics.

I think adding it as a new traffic type (gRPC streaming?) and logging message counts would work, and I can see the approximate location (https://github.com/kiali/kiali/blob/master/graph/telemetry/istio/istio.go#L82) where streaming traffic sniffing could be added. Then it comes down to how to represent this on the UI - I think this could either be summed into gRPC traffic or represented as a different type. The former is probably simpler and doesn't require choosing between gRPC and gRPC streaming for visual representations (e.g traffic animations).

 I may experiment with replacing my request_total messages with the sum of these two, just to see something in the traffic, but it'd be great to see it correctly implemented so that I could see both requests and message stats.

lucasponce:
Hi @emcake, side comment, I have a gRPC example what we use for testing the protocol in Kiali:

https://github.com/kiali/demos/tree/master/fraud-detection

Would you mind to take a look ?
I'd like to check that is a "valid" example that we can use to model your feature.

Thanks

emcake:
Possibly yeah, it looks like you'd get stats from `istio_response_messages_total`. By and large we are using long-lived gRPC streams with low-N requests messages and high-N response messages.

jshaughn:
@emcake OK, so I guess istio_requests_total is only incrementing at the end, but istio_request_messages_total/istio_response_messages_total increments whenever messages are initiated on the existing stream.  I don't see those metrics locally, but that may be because I don't have this scenario.  Are those metrics enabled by default in Istio, or did you have to configure them explicitly?

You are on the right track with the changes, although it wouldn't be trivial.  We'd likely have to add an option to calculate grpc edge rates via the alternate metrics and then perform several more queries, generating the edges from the different metric source.

emcake:
> Are those metrics enabled by default in Istio, or did you have to configure them explicitly?

[...]

Issue #3730, opened 2021-02-23
Consider using tcp_received telemetry for graph generation


jotak:
It's something I often struggle with when testing with kafka. Wrt. TCP, the graph uses only tcp_sent metric, not tcp_received. I think both of them should be used.

To understand better, let's consider this screenshot:
![Capture d’écran de 2021-02-23 11-07-24](https://user-images.githubusercontent.com/2153442/108828660-6a761f80-75c7-11eb-82c4-bdaea123d4e9.png)

It shows TCP telemetry of my workload "ball", which connects to Kafka, and subscribes to a topic.
As you can see, it emits very few data to kafka (must be some kind of ping) but however receives a lot. When choosing a narrow time frame in the graph, like last 1 min, there is good chances that there was no telemetry on "tcp_send". Because my kafka is off mesh, it's a service entry, it doesn't generate telemetry of its own. So the parity that exists between "send" and "receive" when all workloads are in the mesh, doesn't exist here when a service entry is used. Because of that, often I don't see the edges between Kafka and its consumers, I have to increase the time span in order to see them.

![Capture d’écran de 2021-02-23 11-16-40](https://user-images.githubusercontent.com/2153442/108829621-af4e8600-75c8-11eb-9ee5-0efda2376b60.png)

So perhaps it's possible to consider using tcp_receive for graph building, even if it's often (but not always) redundant with "tcp"send".

jshaughn:
@jotak  It's a trade-off between potentially incomplete data and the cost of ensuring complete data.  We already need to make 3 TCP queries per namespace for the main graph-generation.  And potentially more depending on the appenders that need to run.  The idea of doubling the queries is not very attractive.    I suppose it could always be made configurable.   If using a single metric I do wonder if we should be using received_bytes as opposed to sent_bytes, as perhaps there is more received data than sent data, in general.  I suppose an argument could be made that we'd just be reversing the problem.   Still, I don't think I'd like to see us double the TCP query load (which would increase the total queries by a third) on every graph generation.

jotak:
Sure, I understand. The workaround for users facing this issue would be, whenever possible, to move the external service inside the mesh (obviously it won't always be possible).

jshaughn:
@jotak What do you think about switching to received_bytes as our default metric and possibly also making it configurable to use sent_bytes OR received_bytes?  And then, if users demand it, potentially an option for both?

jotak:
It would be nice if it can be configurable, yes.
[...]

PR #4113, opened 2021-06-18
Add flexible rate option support for graph generation

jshaughn:
Supports https://github.com/kiali/kiali/issues/3730
Supports https://github.com/kiali/kiali/issues/4070

**UI PR:** https://github.com/kiali/kiali-ui/pull/2181
**OPERATOR PR:** https://github.com/kiali/kiali-operator/pull/342


graph/options.go
```diff
@@ -33,11 +33,19 @@ const (
...
+	RateRequests              string = "requests" // request count
+	RateSent                  string = "sent"     // tcp bytes sent, grpc request messages, etc
+	RateTotal                 string = "total"    // Sent+Received
 	defaultBoxBy              string = BoxByNone
 	defaultDuration           string = "10m"
 	defaultGraphType          string = GraphTypeWorkload
 	defaultIncludeIdleEdges   bool   = false
 	defaultInjectServiceNodes bool   = false
+	defaultRateHttp           string = RateRequests
+	defaultRateGrpc           string = RateRequests
```
jmazzitelli, line 47: RED ALERT!!!! What blasphemous code doth my eyes witness?

```suggestion
	defaultRateGrpc           string = RateRequests
	defaultRateHttp           string = RateRequests
```

jmazzitelli (commented):
Aside from the egregious non-alphabetized order of some of this code, I built and ran this and all looks good.

graph/options.go
jshaughn: O M G, will fix.

jmazzitelli (dismissed):
LGTM

lucasponce:
I see an increasing concern to desconect the story that the user wants to see.

If I want to generate a topology from "GRPC received messages" when the user would like to examine a chart in detail, that metrics won't be available as this is fixed by requests.

The original story of the details metrics was to give the user the ability to get more details and combine with other information.

I think that story should be keep, and probably the details could be organized by protocols (grpc, http, tcp) or another combination that follows the same story.

I haven't found anything broken in that sense, a couple of problems due performance, but I need to repeat some tests tomorrow to discard that it has nothing to do with the PR.

I'll take a look close to the code, in a first pass I focused on the feature.

jshaughn:
> I see an increasing concern to desconect the story that the user wants to see.
If I want to generate a topology from "GRPC received messages" when the user would like to examine a chart in detail, that metrics won't be available as this is fixed by requests.
The original story of the details metrics was to give the user the ability to get more details and combine with other information.I think that story should be keep,

I agree and I think the question is how much change to introduce incrementally.  The need to better handle different traffic types has been growing, first came TCP issues when sentBytes was unable to visualize some of the TCP traffic.  And now the gRPC streaming problem.  If we introduce this into the main graph users will expect the support to extend in some logical way to potentially relevant areas like side-panel, mini-graph, charts.   

One thing that may not be able to wait for a follow-up PR is some work in the side-panel.   As it stands now if you select a gRPC or TCP edge and you have selected a non-default rate, the data doesn't really match up very well.  Let me know what you think about breaking up the work or adding it to this initial PR.

> ... and probably the details could be organized by protocols (grpc, http, tcp) or another combination that follows the same story.

Note that gRPC is a pita, because it offers both request-based rates OR streaming-type rates (based on message counts), so once protocol can still force different metrics based on the desired rate.

> I haven't found anything broken in that sense, a couple of problems due performance, but I need to repeat some tests tomorrow to discard that it has nothing to do with the PR.

Note that certain options will increase graph-gen time.  As you may expect, more queries are required depending on what is being asked for.  On the bright side, eliminating a protocol, or requests,  may speed things up.

I'll take a look close to the code, in a first pass I focused on the feature.

jshaughn:
UPDATE Jul 1:  Looking more at the side-panel for a gRPC edge using message rate, we need to fix the side panel as part of this work.  So, I will work on that now...

lucasponce:
This one would require to rebase with master now that https://github.com/kiali/kiali/pull/4171 is merged.

jshaughn:
@lucasponce The server and UI PRs are now rebased with your #4171 and any other recent changes.

pbajjuri20:
Tested and verified the TCP, HTTP and GRPC traffic - 1.38.0 (4bdd7cb40fe80e856a62250ac1830a67d770b9bd)

pbajjuri20 (approved):

jmazzitelli (approved):
in praneenth I trust.
lgtm

Issue #3988, opened 2021-05-07
Support for custom istio injection labels and values


lucasponce:
Hi all,

By default,  the Kiali UI add/remove the _istio-injection_ label then set to _enabled/disabled_ values to setup a namespace for sidecar injection.

With Istio's canary control plane support,  this label is deprecated in favor of the **istio.io/rev** label set to the Istio control plane release value (set by the admin).  I.e.  istio.io/rev=1-9-4

I believe the Kiali CR can set a custom label for injection as defined here: https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L679

But what about the value ?   Or is this ^^^ not doing what I think ?

thanks !

_Originally posted by @somejfn in https://github.com/kiali/kiali/discussions/3985_

[1] https://istio.io/latest/docs/setup/upgrade/canary/

jmazzitelli:
Can I ask for a discussion around the names of the new settings being introduced. I supplied a couple suggestions in the operator PR but figured start a discussion here.

The two new settings `istio_current_revision` and `istio_upgrade_revision` aren't clearly related to each other - by that I mean it isn't clear to the user they are part of the same feature AND that if you set on you need to set the other.

I suggested we rename these and put them inside a struct like:

```
istio_canary_revision:
  current: "1.8"
  upgrade: "1.9"
```

This makes it more clear that (a) these involve canary version numbers and (b) you need to set both together.

lucasponce:
> This makes it more clear that (a) these involve canary version numbers and (b) you need to set both together.

+1

hhovsepy:
Question to community:
wdyt, should "Enable/Disable/Remove Auto Injection" functionality still be available on Overview Page when Canary upgrade functionality is enabled?
Because if can cause mess in labeling on namespace. I think it is better to hide "Enable/Disable/Remove" actions when Canary upgrade is enabled.

lucasponce:
> Because if can cause mess in labeling on namespace. I think it is better to hide "Enable/Disable/Remove" actions when Canary upgrade is enabled.

It sounds ok to me, as both features have feature flags to enable/disable is good to choose one approach, to document/describe/demo the feature to collect feedback and iterate on it.

The feature itself doesn't add additional value (i.e. just adding labels at namespace level) but the context of identifying the dataplane namespaces that are pointing to one version or another of the control plane can be very useful.

So, as you're working on the area, I suggest you to take the approach that makes more sense and let's iterate.

[...]

PR #4122, opened 2021-06-24
Added new config in Kiali CR for Istio upgrade scenario.

hhovsepy:
Fix for issue https://github.com/kiali/kiali/issues/3988
UI changes required: https://github.com/kiali/kiali-ui/pull/2184

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -169,15 +169,21 @@ type TracingConfig struct {
 
 // IstioConfig describes configuration used for istio links
 type IstioConfig struct {
-	ComponentStatuses                 ComponentStatuses `yaml:"component_status,omitempty"`
-	ConfigMapName                     string            `yaml:"config_map_name,omitempty"`
-	EnvoyAdminLocalPort               int               `yaml:"envoy_admin_local_port,omitempty"`
-	IstioIdentityDomain               string            `yaml:"istio_identity_domain,omitempty"`
-	IstioInjectionAnnotation          string            `yaml:"istio_injection_annotation,omitempty"`
-	IstioSidecarInjectorConfigMapName string            `yaml:"istio_sidecar_injector_config_map_name,omitempty"`
-	IstioSidecarAnnotation            string            `yaml:"istio_sidecar_annotation,omitempty"`
-	IstiodDeploymentName              string            `yaml:"istiod_deployment_name,omitempty"`
-	UrlServiceVersion                 string            `yaml:"url_service_version"`
+	ComponentStatuses                 ComponentStatuses   `yaml:"component_status,omitempty"`
+	ConfigMapName                     string              `yaml:"config_map_name,omitempty"`
+	EnvoyAdminLocalPort               int                 `yaml:"envoy_admin_local_port,omitempty"`
+	IstioCanaryRevision               IstioCanaryRevision `yaml:"istio_canary_revision,omitempty"`
+	IstioIdentityDomain               string              `yaml:"istio_identity_domain,omitempty"`
+	IstioInjectionAnnotation          string              `yaml:"istio_injection_annotation,omitempty"`
+	IstioSidecarInjectorConfigMapName string              `yaml:"istio_sidecar_injector_config_map_name,omitempty"`
+	IstioSidecarAnnotation            string              `yaml:"istio_sidecar_annotation,omitempty"`
+	IstiodDeploymentName              string              `yaml:"istiod_deployment_name,omitempty"`
+	UrlServiceVersion                 string              `yaml:"url_service_version"`
+}
+
+type IstioCanaryRevision struct {
+	Current string `yaml:"current,omitempty"`
+	Upgrade string `yaml:"upgrade,omitempty"`
 }
 
 type ComponentStatuses struct {
@@ -227,6 +233,7 @@ func (lt *LoginToken) Obfuscate() {
 type IstioLabels struct {
 	AppLabelName       string `yaml:"app_label_name,omitempty" json:"appLabelName"`
 	InjectionLabelName string `yaml:"injection_label,omitempty" json:"injectionLabelName"`
+	InjectionLabelRev  string `yaml:"injection_label_rev,omitempty" json:"injectionLabelRev"`
 	VersionLabelName   string `yaml:"version_label_name,omitempty" json:"versionLabelName"`
 }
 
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -333,6 +334,9 @@ type UIDefaults struct {
 // KialiFeatureFlags available from the CR
 type KialiFeatureFlags struct {
 	IstioInjectionAction bool       `yaml:"istio_injection_action,omitempty" json:"istioInjectionAction"`
+	IstioUpgradeAction bool       `yaml:"istio_upgrade_action,omitempty" json:"istioUpgradeAction"`
+	IstioCurrentRevision string       `yaml:"istio_current_revision,omitempty" json:"istioCurrentRevision"`
+	IstioUpgradeRevision string       `yaml:"istio_upgrade_revision,omitempty" json:"istioUpgradeRevision"`
```
jmazzitelli, line 339: I'm not sure what these are (I assume these are related to the Istio canary deployments), but they don't appear to be appropriate for the Kiali Feature Flags. If this is related to istio canary deployment versions, consider putting these in `external_services.istio` along with the other istio-specific things (you'll see we already have things like IstiodDeploymentName which is also used mainly to support canary deployments)

hhovsepy: These configurations are for new Kiali UI Overviews page's feature to inject the canary version Istio injection label on namespace and trigger pods restart. Are quite similar to 'IstioInjectionAction' flag and coming from Kiali CR and used in kiali-ui.

jmazzitelli: OK. I think we should consider renaming these - because I still don't understand what these are. When I see the name "IstioCurrentRevision" - that to me seems like the value of this setting will be something like "1.10" and when I see the name "IstioUpgradeRevision" - that to me seems like the value of this setting will be something like "1.11". And those don't seem to be related to just a Kiali UI function. That seems to be metadata about the Istio that is running. But I don't think my assumption is correct - based on your comment, these aren't version numbers, but instead are something that relates to injecting values into resources?

Should these be renamed to "IstioInjectCurrentVersion" and "IstioInjectUpgradeRevision"?

config/config.go
```diff
@@ -333,6 +334,9 @@ type UIDefaults struct {
 // KialiFeatureFlags available from the CR
 type KialiFeatureFlags struct {
 	IstioInjectionAction bool       `yaml:"istio_injection_action,omitempty" json:"istioInjectionAction"`
+	IstioUpgradeAction   bool       `yaml:"istio_upgrade_action,omitempty" json:"istioUpgradeAction"`
+	IstioCurrentRevision string     `yaml:"istio_current_revision,omitempty" json:"istioCurrentRevision"`
```
lucasponce, line 338: One comment, as we have splitted the logical of the config in several sections, my thoughts:

- IstioLabels -> InjectionLabelRev. Correct. Typically it will be a fix label similar as app/version used by Istio.
- KialiFeatureFlags -> IstioUpgradeAction. Correct. This should section show enable/disable the feature. (I have no idea who approved the UIDefaults, that one doesn't belong here).
- I'd move IstioCurrentRevision -> IstioUpgradeRevision to the IstioConfig, as it's where I'd look for when I want to personalize something related to my Istio mesh (i.e. I'm going to upgrade between 1.9 and 1.10, so I'd look here).

But anyway, +1 to enable this feature from the config. 

Naming is complex as everyone wants to put his/her criteria, the idea here is to try to put together the same info in the sections that user would look for.

jmazzitelli: > * I'd move IstioCurrentRevision -> IstioUpgradeRevision to the IstioConfig, as it's where I'd look for when I want to personalize something related to my Istio mesh (i.e. I'm going to upgrade between 1.9 and 1.10, so I'd look here).

Yes. To be clear, [IstioConfig](https://github.com/kiali/kiali/blob/v1.36.0/config/config.go#L170-L181) is this section in the CR: https://github.com/kiali/kiali-operator/blob/v1.36.0/deploy/kiali/kiali_cr.yaml#L549-L569

This is also what I recommended in my earlier comment https://github.com/kiali/kiali/pull/4122#discussion_r657958774 : `consider putting these in external_services.istio along with the other istio-specific things`

config/config.go
```diff
@@ -499,6 +507,7 @@ func NewConfig() (c *Config) {
 		IstioLabels: IstioLabels{
 			AppLabelName:       "app",
 			InjectionLabelName: "istio-injection",
+			InjectionLabelRev:  "istio.io/rev",
```
jmazzitelli, line 510: I do not see this in the operator PR: https://github.com/kiali/kiali-operator/pull/347/files

Should this be in the Kiali CR or is this not configurable by the user?

hhovsepy: added

hhovsepy:
Ready for review.

config/config.go
```diff
@@ -499,6 +507,7 @@ func NewConfig() (c *Config) {
...
 			VersionLabelName:   "version",
 		},
 		KialiFeatureFlags: KialiFeatureFlags{
```
lucasponce, line 513: The IstioUpgradeAction would require a default value.
As an upgrade is not a daily task I'd define it as "false" by default.

hhovsepy: this is default 'false' in operator

hhovsepy: now default 'false' here as well

lucasponce: Yes, even if it sounds redundant, but there are some explicit defaults in code as well.

lucasponce:
@hhovsepy thanks for the work, to me, this backend + frontend initial work looks ready to be tested and consolidated.

The upgrade scenario [1] is a larger one and I propose to start showing how this feature can help users on this operation.
So, would you like to prepare a demo (in the platform you want, not necessarily local) showing this for this Friday ?

It would be a good one.

[1] https://istio.io/latest/docs/setup/upgrade/canary/

lucasponce (approved):
I've tested locally.
The feature works as expected.
Also, to have a feature flag helps as this is an optional feature, so I'm +1 to consolidate and iterate.

QE couldn't make it, but I'm in favor to merge and let @skondkar to provide a post-verification to make this feature available in Kiali 1.37

jmazzitelli (approved):
lgtm

skondkar:
Post merge verification is done and added a test case for all scenarios.
Test case > SWSQE-1862

PR #4126, opened 2021-06-24
provide some make targets to allow you to set operator feature settings easily

jmazzitelli:
These are useful when installing the operator via OLM (one use case is when installing with OLM but want to install ad hoc images for testing)

The need arose for this when I was installing the operator via OLM and the allow ad hoc namespace and image features are disabled by default and thus prohibiting someone from running molecule tests or deploying local dev images. Now you just run things like `make operator-set-config-allow-ad-hoc-kiali-namespace` or `make operator-set-config-allow-ad-hoc-kiali-image` to allow this.

Another nice thing is this allows you to turn on the operator profiler - `make operator-set-config-ansible-profiler-on`

Here's the `make help` for the new targets:

```
  operator-set-config-allow-ad-hoc-kiali-namespace   Tells the operator if it should allow Kiali CRs in an ad hoc namespace. Default OPERATOR_ENV_VALUE=true
  operator-set-config-allow-ad-hoc-kiali-image       Tells the operator if it should allow ad hoc images. Default OPERATOR_ENV_VALUE=true
  operator-set-config-ansible-debug-logs             Configures the logger within the operator. Default OPERATOR_ENV_VALUE=true
  operator-set-config-ansible-verbosity              Sets the logging verbosity within the operator. Default OPERATOR_ENV_VALUE=7
  operator-set-config-ansible-profiler-on            Turns on the profiler within the operator.
  operator-set-config-ansible-profiler-off           Turns off the profiler within the operator.
```

### DIFF
--- a/make/Makefile.operator.mk
+++ b/make/Makefile.operator.mk
@@ -2,13 +2,6 @@
 # Targets to deploy the operator and Kiali in a remote cluster.
 #
 
-.ensure-operator-is-running: .ensure-oc-exists
-	@${OC} get pods -l app.kubernetes.io/name=kiali-operator -n kiali-operator 2>/dev/null | grep "^kiali-operator.*Running" > /dev/null ;\
-	RETVAL=$$?; \
-	if [ $$RETVAL -ne 0 ]; then \
-	  echo "The Operator is not running. Cannot continue."; exit 1; \
-	fi
-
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
Fine.

jmazzitelli:
I would like to backport this so we can use then when testing the 1.36 builds.

PR #4127, opened 2021-06-25
the makefile target was broken when trying to set the config when installed via OLM

jmazzitelli:


jshaughn (approved):
LOK

jmazzitelli:
I would like to backport this so we can use then when testing the 1.36 builds.

Issue #1446, opened 2019-08-26
Services list should show when a VirtualService/DestinationRule is applied


lucasponce:
As a user I want to quickly see if a service has a routing rule (VS/DR) applied to it.

The list of services does not show if a VS/DR is applied and I need to go into each service individually to find out.

JIRA:
https://issues.jboss.org/browse/KIALI-1993

lucasponce:
This is a valid use case.
It's a similar effort to check if a VS/DR is applied to an individual Service that per namespace.

lucasponce:
This use case can be very useful.
As now all the info required for this feature is cached, when the user fetches a list of Services, it can be matched with VirtualServices/DestinationRules given.

This feature would be to extend the same feature that happens for a service details, but for a service list.

In terms of visualization it could reuse same badging used in graph for consistency, but more important, it would be good if user can filter by presence of Istio Config on services.

PR #4136, opened 2021-06-30
Show Istio configuration linked with Services and Workloads (and Applications) on list pages

lucasponce:
Fixes https://github.com/kiali/kiali/issues/1446

Frontend https://github.com/kiali/kiali-ui/pull/2192


skondkar (approved):
Verified along with UI PR: https://github.com/kiali/kiali-ui/pull/2192

business/apps.go
```diff
@@ -63,19 +210,63 @@ func (in *AppService) GetAppList(namespace string) (models.AppList, error) {
...
+			svcVirtualServices := kubernetes.FilterVirtualServices(virtualServices, srv.Namespace, srv.Name)
+			apVsNames = append(apVsNames, getIstioResourcesNames(svcVirtualServices)...)
+
+			svcDestinationRules := kubernetes.FilterDestinationRules(destinationRules, srv.Namespace, srv.Name)
+			apDrNames = append(apDrNames, getIstioResourcesNames(svcDestinationRules)...)
 		}
+		appItem.VirtualServices = apVsNames
+		appItem.DestinationRules = apDrNames
+
+		apGwNames := make([]string, 0)
```
xeviknal, line 228: Why don't have one single array of references (as done in the validations) {type,name,namespace?}? The frontend will be way easier to print. Removing switches and ifs.

In the validations there's this object:
https://github.com/kiali/kiali/blob/de7041d3c7b4b5e4b9185cea70572e092cd21318/models/istio_validation.go#L11-L15

And also, in the frontend there it is pretty simplified:
https://github.com/kiali/kiali-ui/blob/master/src/pages/IstioConfigDetails/ValidationReferences.tsx#L18-L28

business/apps.go
```diff
@@ -63,19 +210,63 @@ func (in *AppService) GetAppList(namespace string) (models.AppList, error) {
...
+			apEfNames = append(apEfNames, getIstioResourcesNames(wEf)...)
 		}
-		appItem.Labels = buildFinalLabels(labels)
+		appItem.Labels = buildFinalLabels(applabels)
+		appItem.Gateways = apGwNames
+		appItem.AuthorizationPolicies = apApNames
+		appItem.PeerAuthentications = apPaNames
+		appItem.Sidecars = apScNames
+		appItem.RequestAuthentications = apRaNames
+		appItem.EnvoyFilters = apEfNames
```
xeviknal, line 262: Another thing I see, just a proposal, is that you can get rid of some repetition and initialization by using an associative/struct array. The proposal feels more (a bit) closed to modification.

```go
	for _, wrk := range valueApp.Workloads {
		joinMap(applabels, wrk.Labels)

		wSelector := labels.Set(wrk.Labels).AsSelector().String()

		data := []struct {
			source []kubernetes.IstioObject
			destination *[]string
		} {
			{ source: gateways, destination: &appItem.Gateways },
			{ source: authorizationPolicies, destination: &appItem.AuthorizationPolicies },
			{ source: peerAuthentications, destination: &appItem.PeerAuthentications },
			{ source: sidecars, destination: &appItem.Sidecars },
			{ source: requestAuthentications, destination: &appItem.RequestAuthentications },
			{ source: envoyFilters, destination: &appItem.EnvoyFilters },
		}

		for _, pair := range data {
			wEf := kubernetes.FilterIstioObjectsForWorkloadSelector(wSelector, pair.source)
			*(pair.destination) = append(*(pair.destination), getIstioResourcesNames(wEf)...)
		}
	}
	appItem.Labels = buildFinalLabels(applabels)
```

business/services.go
```diff
@@ -75,31 +82,71 @@ func (in *SvcService) GetServiceList(namespace string) (*models.ServiceList, err
...
 		if IsNamespaceCached(namespace) {
-			deployments, err = kialiCache.GetDeployments(namespace)
+			deployments, err2 = kialiCache.GetDeployments(namespace)
 		} else {
-			deployments, err = in.k8s.GetDeployments(namespace)
+			deployments, err2 = in.k8s.GetDeployments(namespace)
 		}
-		if err != nil {
+		if err2 != nil {
 			log.Errorf("Error fetching Deployments per namespace %s: %s", namespace, err)
```
xeviknal, line 94: Shouldn't this log print `err2` instead?

business/services.go
```diff
@@ -75,31 +82,71 @@ func (in *SvcService) GetServiceList(namespace string) (*models.ServiceList, err
...
+		go func() {
+			defer wg.Done()
+			var err2 error
+			if IsNamespaceCached(namespace) {
+				virtualServices, err2 = kialiCache.GetIstioObjects(namespace, kubernetes.VirtualServices, "")
+			} else {
+				virtualServices, err2 = in.k8s.GetIstioObjects(namespace, kubernetes.VirtualServices, "")
+			}
+			if err2 != nil {
+				log.Errorf("Error fetching Istio VirtualServices per namespace %s: %s", namespace, err)
```
xeviknal, line 109: Same here. I guess that `err` should be replaced by `err2`.

business/services.go
```diff
@@ -75,31 +82,71 @@ func (in *SvcService) GetServiceList(namespace string) (*models.ServiceList, err
...
+		go func() {
+			defer wg.Done()
+			var err2 error
+			if IsNamespaceCached(namespace) {
+				destinationRules, err2 = kialiCache.GetIstioObjects(namespace, kubernetes.DestinationRules, "")
+			} else {
+				destinationRules, err2 = in.k8s.GetIstioObjects(namespace, kubernetes.DestinationRules, "")
+			}
+			if err2 != nil {
+				log.Errorf("Error fetching Istio DestinationRules per namespace %s: %s", namespace, err)
```
xeviknal, line 123: Same here. I guess that `err` should be replaced by `err2`.

business/apps.go
```diff
@@ -48,13 +48,160 @@ func buildFinalLabels(m map[string][]string) map[string]string {
...
+		go func() {
+			defer wg.Done()
+			var err2 error
+			if IsNamespaceCached(namespace) {
+				virtualServices, err2 = kialiCache.GetIstioObjects(namespace, kubernetes.VirtualServices, "")
+			} else {
+				virtualServices, err2 = in.k8s.GetIstioObjects(namespace, kubernetes.VirtualServices, "")
+			}
+			if err2 != nil {
+				log.Errorf("Error fetching Istio VirtualServices per namespace %s: %s", namespace, err)
```
xeviknal, line 98: Shouldn't this log print `err2` instead? If so, it should be amended in other similar LOCs.

xeviknal (commented):
Sharing some thoughts. I'd like to hear what you think of.

business/apps.go
```diff
@@ -48,13 +48,160 @@ func buildFinalLabels(m map[string][]string) map[string]string {
...
 		Namespace: models.Namespace{Name: namespace},
 		Apps:      []models.AppListItem{},
 	}
-	apps, err := fetchNamespaceApps(in.businessLayer, namespace, "")
-	if err != nil {
+
+	var err error
+	var apps namespaceApps
+
+	var virtualServices []kubernetes.IstioObject
```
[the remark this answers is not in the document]
lucasponce: Yes, but those are different types.
Having a single IstioObject interface is something that we need to address, even in the frontend we have different types.

business/apps.go
```diff
@@ -48,13 +48,160 @@ func buildFinalLabels(m map[string][]string) map[string]string {
...
+		var err2 error
+		apps, err2 = fetchNamespaceApps(in.businessLayer, namespace, "")
+		if err2 != nil {
+			log.Errorf("Error fetching Applications per namespace %s: %s", namespace, err2)
+			errChan <- err2
+		}
+	}()
+
+	if linkIstioResources {
+		go func() {
```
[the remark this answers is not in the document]
lucasponce: I prefer this comment.
Those objects are using the IstioObject more from a technical debt but from business perspective are different types being fetched from different APIs, so keeping the split but applying this optimization sounds good.

P.S.: I think we need to bring strong types sooner than later, there is an old issue open about this.

business/apps.go
lucasponce: Oh, it makes sense, I missed that type, as it was named as "Validation" I missed for this.

business/apps.go
lucasponce: +1

business/apps.go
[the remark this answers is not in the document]
xeviknal: +1 on bringing strong types but with good interfaces and more logic distributed in each type in order to reduce tons of switch and if/else statements.

business/apps.go
xeviknal: Yeah, feel free to move the struct into another module and maybe rename it. At the implementation time felt that it was the only moment when references were needed. So agree on adjusting this one.

business/apps.go
```diff
@@ -48,13 +48,75 @@ func buildFinalLabels(m map[string][]string) map[string]string {
...
+				} else {
+					*dest, err2 = in.k8s.GetIstioObjects(namespace, resourceType, "")
+				}
+				if err2 != nil {
+					log.Errorf("Error fetching Istio %s per namespace %s: %s", resourceType, namespace, err2)
+					errChan <- err2
+				}
+			}(namespace, linkedResource, dest, errChan)
+		}
+	}
```
xeviknal, line 115: Sorry there is still one thing bothering me a bit. I thought the individual declaration on lines L60-L67 were necessaries to avoid race conditions. But I guess it is not. So you can just fill a hashmap with all the resources by resource name.

```go
	resources := []string{ kubernetes.VirtualServices, kubernetes.DestinationRules, kubernetes.Gateways,
		kubernetes.AuthorizationPolicies, kubernetes.PeerAuthentications, kubernetes.Sidecars,
		kubernetes.RequestAuthentications, kubernetes.EnvoyFilters,
	}

	linkedResources := map[string]*[]kubernetes.IstioObject{}
	if linkIstioResources {
		for _, resource := range resources {
			var resourceObjects []kubernetes.IstioObject
			linkedResources[resource] = &resourceObjects
			go func(namespace, resourceType string, dest *[]kubernetes.IstioObject, errChan chan error) {
				defer wg.Done()
				var err2 error
				if IsNamespaceCached(namespace) {
					*dest, err2 = kialiCache.GetIstioObjects(namespace, resourceType, "")
				} else {
					*dest, err2 = in.k8s.GetIstioObjects(namespace, resourceType, "")
				}
				if err2 != nil {
					log.Errorf("Error fetching Istio %s per namespace %s: %s", resourceType, namespace, err2)
					errChan <- err2
				}
			}(namespace, resource, &resourceObjects, errChan)
		}
	}
```

With this you can get rid of the declaration of one var for each resource type. Adding one more resource will be as easy as adding the type in the `resources` array (L60-L67).

business/apps.go
```diff
@@ -63,14 +125,63 @@ func (in *AppService) GetAppList(namespace string) (models.AppList, error) {
...
+		svcReferences := make([]*models.IstioValidationKey, 0)
 		for _, srv := range valueApp.Services {
-			joinMap(labels, srv.Labels)
+			joinMap(applabels, srv.Labels)
+			if linkIstioResources {
+				svcVirtualServices := kubernetes.FilterVirtualServices(*linkedResources[kubernetes.VirtualServices], srv.Namespace, srv.Name)
+				svcDestinationRules := kubernetes.FilterDestinationRules(*linkedResources[kubernetes.DestinationRules], srv.Namespace, srv.Name)
+				allFiltered := append(svcVirtualServices, svcDestinationRules...)
+				for _, a := range allFiltered {
+					ref := models.IstioValidationKey{
```
xeviknal, line 137: There's a method for creating `IstioValidationKey`s. And it is good to use it to hide implementation details and consistency, I guess.

```go
models.BuildKey(a.GetObjectMeta().Name, a.GetObjectMeta().Namespace, a.GetTypeMeta().Kind)
```

business/apps.go
```diff
@@ -63,14 +125,63 @@ func (in *AppService) GetAppList(namespace string) (models.AppList, error) {
...
+						}
+						exist := false
+						for _, r := range wkdReferences {
+							if r.Name == ref.Name && r.Namespace == ref.Namespace && r.ObjectType == ref.ObjectType {
+								exist = true
+							}
+						}
+						if !exist {
+							wkdReferences = append(wkdReferences, &ref)
+						}
```
xeviknal, line 177: Are there going to be duplicates here? `wkdReferences` is filled each time with a set of one single type of objects (i.e. Gateways, Sidecars...) and it is impossible to have two equal instances between each set because they have different object type. And within the same set of objects, it is no possible because k8s don't allow that.

Anyway, if that is the case, I'd suggest to use the native `== operator` for comparing references. Golang implements the `DeepEquals()`. So you could do something like this:

```go
for _, r := range wkdReferences {
  exist = exist || *r == ref
}
```

xeviknal (commented):
I left some comments. There is one nitpicking. I have one comment regarding the duplicate removal logic.

business/apps.go
lucasponce: Yes, see that we are at application level which may iterate over the workloads.
So, for example, an AuthorizationPolicy that has a selector like "app=travels" would bring same object for multiple workloads like "travels-v1", "travels-v2", "travels-v3".

Perhaps at service and workload level is not necessary, as we don't iterate, but yes, for App is necessary.

business/apps.go
lucasponce: Oh, true, there are some intermediate variables that are not needed.
Thanks

business/apps.go
lucasponce: Well, I may argue :).
We don't have constructors for all models, as it's not the standard way in golang, but as this one exists, I'm +1 to update it.

lucasponce:
Ok @xeviknal thanks a lot for the review, good catches !

business/services.go
```diff
@@ -75,31 +82,64 @@ func (in *SvcService) GetServiceList(namespace string) (*models.ServiceList, err
...
-		if err != nil {
-			log.Errorf("Error fetching Deployments per namespace %s: %s", namespace, err)
-			errChan <- err
+		if err2 != nil {
+			log.Errorf("Error fetching Deployments per namespace %s: %s", namespace, err2)
+			errChan <- err2
 		}
 	}()
 
+	linkedResources := map[string]*[]kubernetes.IstioObject{
```
xeviknal, line 99: Don't you want to apply the same changes in here?

business/workloads.go
```diff
@@ -75,19 +75,96 @@ func isWorkloadIncluded(workload string) bool {
...
+				} else {
+					*dest, err2 = in.k8s.GetIstioObjects(namespace, resourceType, "")
+				}
+				if err2 != nil {
+					log.Errorf("Error fetching Istio %s per namespace %s: %s", resourceType, namespace, err2)
+					errChan <- err2
+				}
+			}(namespace, linkedResource, dest, errChan)
+		}
+	}
```
xeviknal, line 137: I guess you missed it here as well. It can be refactored to avoid static declarations for each type.

xeviknal (commented):
I think there's two places where my previous comments could be applied too. And think that would be it.

business/services.go
lucasponce: Ops I missed, sorry.

xeviknal (approved):
LGTM!

lucasponce:
Thanks @xeviknal 
cc @skondkar, we've added some optimizations to the code, but worth if you could give it another pass from your side (and sorry I should wait until all Dev churn was done).

skondkar (approved):
Reverified with UI PR#2192. LGTM

Issue #4145, opened 2021-07-03
Use annotations to personalize CustomDashboards


lucasponce:
When the refactoring of the MonitoringDashboard CRD was consolidated into the Kiali CRD, the personalization of the dashboard per namespace was planned for a next iteration as commented:
https://github.com/kiali/kiali/issues/4057#issuecomment-854480502

It has been reported that this feature is actively used for users:
https://github.com/kiali/kiali/issues/4057#issuecomment-873415952

So it could be necessary to propose an alternative to the same personalization feature.
Kiali followed Istio to personalize configuration using annotations (today that's used for health) and it would make sense to follow same pattern to personalize a dashboard: a specific dashboard could be personalized per the service owner when annotated in a given namespace, otherwise, Kiali will use the templates provided by default or added globally in the Kiali configuration.

PR #4146, opened 2021-07-03
Fix regression in personalized CustomDashboards at namespace level

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4145

When Kiali moved the MonitoringDashboard into the Kiali CR config, the personalization of templates per namespace was not covered.

As reported by @vbergbauer, this feature was used as there could be different personas that write the templates (typically Kiali CR would be managed by an operator that could be different from who can operate a given namespace - another operator and/or developer -).

So, this PR proposes to bring back the personalization of templates at namespace level but instead to use the old MonitoringDashboard CRD, the mechanism would be to use an annotation in the namespace.

Note, this strategy is widely used in Istio and other Kiali features as well.

So, to create a personalized template, just annotate it using the MonitoringDashboard syntax in the namespace as it follows:

```
apiVersion: v1
kind: Namespace
metadata:
  name: default
  annotations:
    dashboards.kiali.io/templates: |
      - name: custom_envoy
        title: Custom_Envoy
        discoverOn: "envoy_server_uptime"
        items:
          - chart:
              name: "Pods uptime"
              spans: 12
              metricName: "envoy_server_uptime"
              dataType: "raw"
```

Then a given namespace can have this template as it's shown in the screenshot:

image


lucasponce:
I'm thinking that this should be backported in Kiali 1.36.x, but it's something to be discussed as PR touches several APIs.

business/apps_test.go
```diff
@@ -15,6 +15,7 @@ import (
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes/kubetest"
 	"github.com/kiali/kiali/prometheus/prometheustest"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jmazzitelli, line 18: trivial - but this should be formatted as per standard (put this in the middle set of imports - i.e after line 13).

business/workloads_test.go
```diff
@@ -20,6 +20,7 @@ import (
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/kubernetes/kubetest"
 	"github.com/kiali/kiali/prometheus/prometheustest"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
jmazzitelli, line 23: trivial - but this should be formatted as per standard (put this in the middle set of imports - i.e after line 17).

config/dashboards/dashboards_config.go
```diff
@@ -10,6 +10,8 @@ import (
 )
 
 const (
+	// Namespace annotation for dashboard templates
+	DashboardTemplateAnnotation string = "dashboards.kiali.io/templates"
```
jmazzitelli, line 14: [Standard godoc convention](https://blog.golang.org/godoc) is that the comment should be a complete sentence that `begins with the name of the element it describes.` So this should be something like:

```suggestion
	// DashboardTemplateAnnotation is the key name of the namespace annotation for dashboard templates
	DashboardTemplateAnnotation string = "dashboards.kiali.io/templates"
```

config/dashboards/dashboards_config.go
```diff
@@ -128,6 +130,18 @@ func GetBuiltInMonitoringDashboards() MonitoringDashboardsList {
 	}
 }
 
+func GetNamespaceMonitoringDashboards(annotations map[string]string) MonitoringDashboardsList {
```
jmazzitelli, line 133: Since this is a public API, should add a godoc comment to this function:

```suggestion
// GetNamespaceMonitoringDashboards will examine the given namespace annotations and return any dashboard yaml found.
func GetNamespaceMonitoringDashboards(annotations map[string]string) MonitoringDashboardsList {
```

config/dashboards/dashboards_config.go
```diff
@@ -128,6 +130,18 @@ func GetBuiltInMonitoringDashboards() MonitoringDashboardsList {
...
+	if dashboardYaml, ok := annotations[DashboardTemplateAnnotation]; ok {
+		if v, err := unmarshal(dashboardYaml); err == nil {
+			return MonitoringDashboardsList(v)
+		} else {
+			log.Errorf("Failed to unmarshall namespace dashboard yaml. err=%v", err)
```
jmazzitelli, line 138: Is there a way we can get the namespace name that owns these annotations so we can add it to this error message? Or can we bubble up the err so the caller (which I assume knows what namespace these annotations came from) can log the name of the namespace along with this error message?

Think about what will happen if the user put invalid yaml in the annotation (which is very likely, given the complexity of the dashboard yaml). When this happens, the user will see the log message about the bad yaml, but the user will not be told WHICH namespace has the bad annotation. Without knowing the namespace, it could be difficult to find the bad yaml in order to fix it (especially if they have hundreds (or more) of namespaces).

business/dashboards.go
```diff
@@ -39,12 +39,20 @@ func NewDashboardsService() *DashboardsService {
 	if nsLabel == "" {
 		nsLabel = "kubernetes_namespace"
 	}
+
+	// Overwrite Custom dashboards defined at Namespace level
+	builtInDashboards := cfg.CustomDashboards.OrganizeByName()
+	nsDashboards := dashboards.GetNamespaceMonitoringDashboards(namespace.Annotations)
```
jmazzitelli, line 45: See my earlier comment about wanting to log the namespace if this function `GetNamespaceMonitoringDashboards` results in an error. This appears to be the only place where this function is called. So we should either (a) pass in `namespace` as a new parameter to this function or (b) bubble up the err from this function and log that error along with `namespace` so the user can easily find the bad annotation in order to fix it.

jmazzitelli (changes requested):
mostly trivial change requests, but there is an important one here -- the one about logging the namespace in the error message if the annotation has invalid dashboard yaml.

business/dashboards.go
lucasponce: +1
I'm keeping the original design, so, if some problem happens at namespace level, Kiali logs the error with the namespace (thanks for the catch) and it will fall with the default one.

My first draft had that but I refactored to not create a circular dependency here, but yes, passing a namespace parameter is necessary.

jmazzitelli (dismissed):
I ran this and things look good, with a minor issue that I don't think we need to really worry about - but I will document what I found here. I'm approving this PR since this now immediately fixes the problem our community reported.

I first put in some bad yaml and saw this in the logs - you can see which namespace had the bad yaml - so this is good:

```
ERR Failed to unmarshall yaml dashboard in namespace bookinfo. err=failed to parse monitoring dashboard yaml data. error=yaml: line 6: mapping values are not allowed in this context
```

I then corrected the yaml and I see the dashboard getting installed - so this is good:

image

I then clicked on one of the other tabs in the workload page (I tried this with the "Overview" tab and the "Logs" tab) and when I REMOVE the dashboard, the custom dashboard tab disappears. When I ADD the dashboard back, the custom dashboard tab re-appears. So this is good.

But while looking at the custom dashboard's graphs (in this example, I navigated to the "Custom_Envoy" tab), I then REMOVED the annotation (without changing where my browser was pointing to - I did not do a refresh or navigate to any different page) and I got an error in the Kiali UI:

image

image

To fix this, all I had to do was either (a) refresh the UI page or (b) navigate away from the workload page and then go back to it (it seems the dashboards are re-loaded at that point and so the now-deleted dashboard is gone).  So it isn't horrible. And this is going to be a rare edge-case - where the annotation is removed at the same time someone (who is unaware that this happened) is looking at that custom dashboard that is removed. We can either document this as expected behavior (with the solution being, "Simply refresh the web page") or we can write this up as a "bug" github issue with the intention of fixing it later. I, personally, think we can just document this behavior since it will rarely happen and the workaround is simple and easy.

lucasponce:
Thanks @jmazzitelli for the review.
Discussing with @vbergbauer seems that also extending annotations to the Workload could be a good idea.
I'm going to see if it can be added to this one or if it may require a followup work.

I guess the hit you get is due the namespace caching, without cluster permissions Namespace are refreshed per short defined interval and probably it's the root of the error you saw. I'll take a look in case I missed something, but as you say, I'm +1 to document.

leandroberetta (dismissed):
LGTM and worked in my environment.

lucasponce:
I've updated the PR to cover the Workload case.
It requires a frontend UI to pass a parameter https://github.com/kiali/kiali-ui/pull/2197

But with this addition, when using the annotation at deployment level, the dashboard is used only for the given workload:
image

Compared with
image

jmazzitelli (approved):
I didn't run this - but new code looks OK.

It looks like there needs to be several areas that need automated testing to make sure this doesn't break in the future. We have two places where dashboards can be overridden (first namespace, then workload - with workload taking precendence) - I can try to add a molecule test for these, but that should not hold up the merge of this PR.

jmazzitelli:
I'm working on https://github.com/kiali/kiali/issues/4160 by writing some more molecule tests.
In my PR https://github.com/kiali/kiali-operator/pull/352 I'm seeing an error that looks to be a bug in this PR. When I get more details, I'll post them.

jmazzitelli:
OK, the problem appears to be that workload dashboard annotations are not getting picked up. It could be my test is not correct, or it could be a problem.

@lucasponce for my test, I'm adding an annotation to a pod and then I'm trying to obtain the dashboards for that workload. Here's the annotation in the deployment template: https://github.com/kiali/kiali-operator/pull/352/files#diff-945566bf57bd49d70a2efbf039b59cb263d945144e46e29444f00693999f3ed0R112-R124
This isn't working. The dashboards I'm getting back are the dashboards from the namespace annotation, whereas I think the workload dashboard annotation is supposed to take precedence and override the namespace dashboards.

jmazzitelli:
(also, you need to rebase this on master to fix the conflict)

jmazzitelli:
I'm seeing the workload annotation not working. Here's how you can test this yourself.

This test below will create a namespace annotation and a workload annotation. The namespace annotation will create two dashboards named `custom_envoy1` and `custom_envoy2`. The workload annotation will create a dashboard `custom_envoy3` and it will remove `custom_envoy1` (it does this using the feature of setting the name but not defining any charts).
The problem you will see is there will be a custom_envoy1 and custom_envoy2 but not a custom_envoy3 - this tells me the namespace annotation works but the workload annotation seems to be ignored.

Here's the test:

1. Run this to create a `simple-server` pod and a `simple-client` pod inside a single namespace called `simple-mesh` - there are annotations on the namespace and the simple-server pod: "kubectl apply -f [simple-mesh.yaml.txt](https://github.com/kiali/kiali/files/6787288/simple-mesh.yaml.txt)"
2. Double check the namespace annotation exists: `kubectl get ns simple-mesh -o yaml`
3. Double check the workload annotation exists on the pod: `kubectl get pods -n simple-mesh -l app=simple-server -o yaml`
4. Now ask the Kiali API to return the custom_envoy1 dashboard - this should fail because the workload annotation told Kiali to remove it (do not include it in the list of dashboards for this workload). But I do see it:
```
$ curl http://localhost:20001/api/namespaces/simple-mesh/customdashboard/custom_envoy1?labelsFilter=app:simple-server,version:v99

{"name":"custom_envoy1","title":"Custom_Envoy_1","charts":[{"name":"Pods uptime","unit":"","spans":3,"startCollapsed":false,"metrics":[{"labels":{},"datapoints": ....
```
5. Ask the Kiali API to return the custom_envoy2 dashboard - this works as expected:
```
$ curl http://localhost:20001/api/namespaces/simple-mesh/customdashboard/custom_envoy2?labelsFilter=app:simple-server,version:v99

{"name":"custom_envoy2","title":"Custom_Envoy_2","charts":[{"name":"Allocated memory","unit":"bytes","spans":3,"startCollapsed":false,"min":0,"metrics":[{"labels":{},"datapoints":...
```
6. Ask the Kiali API to return the custom_envoy3 dashboard - this should work because this dashboard is defined in the workload annotation, but the request fails:
```
$ curl http://localhost:20001/api/namespaces/simple-mesh/customdashboard/custom_envoy3?labelsFilter=app:simple-server,version:v99

{"error":"Dashboard [custom_envoy3] does not exist or is disabled"}
```

lucasponce:
Hi @jmazzitelli, thanks for testing.

I think the query used in the test misses the workload(s) parameter as defined in https://github.com/kiali/kiali-ui/pull/2197/files#diff-b2f58cb27e98ffce9139f886bd06aad1e2fb90c05ffb1360624523accceaf22dR18 

From the original design, the custom dashboards

```
/api/namespaces/simple-mesh/customdashboard/custom_envoy3?labelsFilter=app:simple-server,version:v99
```

Are defined globally (now via the Kiali CR) or per Namespace (annotation), and originally, a Deployment can add a "dashboard reference" to the template used.

Now, to allow this personalization in the workload, the "customdashboard" should add a parameter of the workload (and workloadType), this is to tell user to use that workload as it's optional, otherwise we'd need to check all workloads in everyquery, which looked a potential performance penalty (even using caching).

So, if you use the workload/workloadType parameters as modified in the UI, that should work.

As this call is only used in the context of a workload details, the workload would be always present in that scenario, but yes, in the molecule test would need those parameters as well.

jmazzitelli:
@lucasponce 

> Now, to allow this personalization in the workload, the "customdashboard" should add a parameter of the workload (and workloadType)

OK. That makes sense. In the case of this test I'm running, what would the values of those parameters be? I guess "workloadType" would be "Pod" and "workload" would be the name of the pod (i.e. the "metadata.name" of the pod itself)?

jmazzitelli:
This doesn't appear to be working. Here's what I even see in the UI after I rebuilt with https://github.com/kiali/kiali-ui/pull/2197

image

Notice "Custom_Envoy_1" is there, but it should not be. I also do not see Custom_Envoy_3 which should be there because it is in the workload annotation:

```
$ kubectl get pod -n simple-mesh simple-server-8567587fc4-6qhl6 -o yaml

apiVersion: v1
kind: Pod
metadata:
  annotations:
    dashboards.kiali.io/templates: |
      - name: custom_envoy1
      - name: custom_envoy3
        title: Custom_Envoy_3
        discoverOn: "envoy_server_uptime"
        items:
          - chart:
              name: "Heap size"
              unit: "bytes"
              spans: 3
              metricName: "envoy_server_memory_heap_size"
              dataType: "raw"
              min: 0
  name: simple-server-8567587fc4-6qhl6
  namespace: simple-mesh
...
```

jmazzitelli:
I updated [my test instructions](https://github.com/kiali/kiali/pull/4146#issuecomment-876752008) but I see the same thing (incorrect dashboards) as I see in the UI itself. So something isn't working when specifying dashboards in both namespace and workload annotations.

lucasponce:
> I updated my test instructions but I see the same thing (incorrect dashboards) as I see in the UI itself. So something isn't working when specifying dashboards in both namespace and workload annotations.

Thanks for testing, I'll review in case I missed something

business/dashboards.go
```diff
@@ -39,12 +39,28 @@ func NewDashboardsService() *DashboardsService {
...
+			builtInDashboards[name] = dashboard
+		}
+	}
+	if workload != nil {
+		wkDashboards := dashboards.GetWorkloadMonitoringDashboards(namespace.Name, workload.Name, workload.DashboardAnnotations)
+		for name, dashboard := range wkDashboards.OrganizeByName() {
+			builtInDashboards[name] = dashboard
+		}
+	}
+
```
jmazzitelli, line 57: OK. Almost there. The UI looks OK now. However, the [molecule test](https://github.com/kiali/kiali-operator/pull/352/files#diff-44588589631a74bd989b470bcf4a3c449be1939cd1d55c781098ed6ff5aa790dR115) is failing because custom_envoy1 still shows up (returns 200 status code), though it is empty. This API should return a 500 and say the dashboard does not exist but instead it returns this:

```
$ curl "http://localhost:20001/kiali/api/namespaces/simple-mesh/customdashboard/custom_envoy1?labelsFilter=app:simple-server,version:v99&workloadType=Deployment&workload=simple-server"

{"name":"custom_envoy1","title":"","charts":[],"aggregations":[],"externalLinks":[]}
```

The reason for this is the way the dashboards are aggregated in this PR is not correct. This code in the PR is not using function [AddMonitoringDashboards](https://github.com/kiali/kiali/blob/v1.36.0/config/dashboards/dashboards_config.go#L99-L103) but it should.

The way it should be working is - if a dashboard is declared but has NO charts and NO discoverOn then that dashboard should be removed [as you see commented here](https://github.com/kiali/kiali/blob/v1.36.0/config/dashboards/dashboards_config.go#L102) and [implemented here](https://github.com/kiali/kiali/blob/v1.36.0/config/dashboards/dashboards_config.go#L114). We use this feature already when users declare dashboards in the Kiali CR in order to override the built-in dashboards. A user can REMOVE a built-in dashboard by declaring the dashboard by name only (with no additional data). We test this kind of thing already... see our [current molecule test](https://github.com/kiali/kiali-operator/blob/v1.36.0/molecule/api-test/dashboards.test.yaml#L87-L98) that checks for that functionality of removing a built-in dashboard ([this is what the yaml looks like](https://github.com/kiali/kiali-operator/blob/v1.36.0/molecule/api-test/kiali-cr.yaml#L11-L12) when removing a dashboard). The same thing should happen up the chain - I should be able to remove built-in dashboards (and those declared by the Kiali CR) by overriding it in the namespace annotation. And individual workloads should be able to remove dashboards even declared in namespace annotations by overriding it in the workload annotations (and this is what my molecule test is testing and failing).

The solution is to make this change:

```suggestion
	// Overwrite Custom dashboards defined at Namespace level and Workoad level
	builtInDashboards := cfg.CustomDashboards
	if namespace != nil {
		nsDashboards := dashboards.GetNamespaceMonitoringDashboards(namespace.Name, namespace.Annotations)
		builtInDashboards = dashboards.AddMonitoringDashboards(builtInDashboards, nsDashboards)
	}
	if workload != nil {
		wkDashboards := dashboards.GetWorkloadMonitoringDashboards(namespace.Name, workload.Name, workload.DashboardAnnotations)
		builtInDashboards = dashboards.AddMonitoringDashboards(builtInDashboards, wkDashboards)
	}
```

This requires an additional change down a few lines -- on line 63, you need to change that to be the following (I'll add a suggested change comment next - github won't let me do a single suggested change here):

```
dashboards:      builtInDashboards.OrganizeByName(),
```

With this fixed, things work:

```
$ curl "http://localhost:20001/api/namespaces/simple-mesh/customdashboard/custom_envoy1?labelsFilter=app:simple-server,version:v99&workloadType=Deployment&workload=simple-server"

{"error":"Dashboard [custom_envoy1] does not exist or is disabled"}
```

business/dashboards.go
```diff
@@ -39,12 +39,28 @@ func NewDashboardsService() *DashboardsService {
...
 	return &DashboardsService{
 		CustomEnabled:   customEnabled,
 		promConfig:      prom,
 		globalNamespace: cfg.Deployment.Namespace,
 		namespaceLabel:  nsLabel,
-		dashboards:      cfg.CustomDashboards.OrganizeByName(),
+		dashboards:      builtInDashboards,
```
jmazzitelli, line 63: See previous comment - this change goes with the bigger change in the earlier suggested change
```suggestion
		dashboards:      builtInDashboards.OrganizeByName(),
```

jmazzitelli (changes requested):
Just need to fix the aggregation code

business/dashboards.go
lucasponce: Mmm ok, I was using the premise that the builtin were re-built on every query, so, if the user remove them, then they are not present and they should fall in the default.

I've added your suggestions, let me know how it looks like.

I'll continue on Monday, also this one would require some good QE testing as well.

jmazzitelli (dismissed):
This all works now. We also have new molecule tests that confirm it working.

When  we merge this PR, we need to also merge the operator PR with the new molecule tests.. this one--> https://github.com/kiali/kiali-operator/pull/352

jmazzitelli:
I should update the molecule test when run on OpenShift - I didn't realize we have to worry about Project vs. Namespaces.

jmazzitelli (approved):
[molecule tests](https://github.com/kiali/kiali-operator/pull/352) pass again after the latest commit

prachiyadav:
Test case > SWSQE-1847

prachiyadav (approved):
Screenshot from 2021-07-14 00-54-21


Screenshot from 2021-07-13 23-07-52

Issue #4090, opened 2021-06-09
Display traffic scenario badges when present


nrfox:
This is a follow up to #1477.

The current implementation only displays the traffic routing badges if they were created through the Kiali wizard. The badges should be displayed on the nodes if they are present on the virtual service or destination rule regardless of how they were created.

This would involve updating the backend appender to check the virtual service spec for the supported traffic routing scenarios similar to how badging is added for circuit breakers. In addition ensuring the front end displays the badges correctly when multiple scenarios are active at once.

The following scenarios need to be covered:
- Request Timeouts
- Traffic Shifting
- TCP Traffic Shifting
- Request Routing
- Fault Injection

Since the two traffic shifting scenarios share a single icon, only 1 badge can be displayed when both are present.

jshaughn:
@lucasponce To me this seems like a high priority item, as users could find it confusing to be missing expected badges just because the definitions were made outside of Kiali.  What do you think?   Alternatively, should we try and have Kiali wizards work better with externally defined definitions?

lucasponce:
+1

To make work any Istio definition with the Wizard is part of the ongoing effort we are doing on that area but yes, we can use the same strategy on badging.

This is what I had brainstormed for the feature in the Wizards and can be applied in the badges in a direct 1:1 way:

- If a VS is present under a service, by default it will be consider a "Request Routing" (that will be the more complex Wizard that will try to fit the corner cases).

- If the VS has the presence of "Request Timeouts" and/or "Fault Injection", it can enable these badges.

- If there is only one rule with multiple destination it can use the "Traffic Shifting" / "TCP Traffic Shifting" depending on the type of destination (we don't support yet the SSL destination, but that can be addressed in the future).

In cases of multiple combinations or some ambiguity, I'd use the "Request Routing" as it allows to combine the previous situations.

PR #4183, opened 2021-07-15
Add kiali scenarios to graph node from VS spec

nrfox:
Updates the istio appender to add the kiali scenario node metadata labels based on the VS spec instead of the presence of the wizard label.

Fixes: #4090

### DIFF
--- a/graph/types.go
+++ b/graph/types.go
@@ -20,12 +20,6 @@ const (
 	NodeTypeWorkload              string = "workload"
 	TF                            string = "2006-01-02 15:04:05" // TF is the TimeFormat for timestamps
 	Unknown                       string = "unknown"             // Istio unknown label value
-	WizardFaultInjectionLabel     string = "fault_injection"
-	WizardLabelKey                string = "kiali_wizard" // Label the front-end wizards add to objects created by them
-	WizardRequestRoutingLabel     string = "request_routing"
-	WizardRequestTimeoutsLabel    string = "request_timeouts"
-	WizardTCPTrafficShiftingLabel string = "tcp_traffic_shifting"
-	WizardTrafficShiftingLabel    string = "traffic_shifting"
 	// private
 	passthroughCluster string = "PassthroughCluster"
 	blackHoleCluster   string = "BlackHoleCluster"
--- a/graph/types.go
+++ b/graph/types.go
@@ -8,18 +8,18 @@ import (
 )
 
 const (
-	GraphTypeApp                  string = "app"
-	GraphTypeService              string = "service" // Treated as graphType Workload, with service injection, and then condensed
-	GraphTypeVersionedApp         string = "versionedApp"
-	GraphTypeWorkload             string = "workload"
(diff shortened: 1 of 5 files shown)


models/virtual_service.go
```diff
@@ -67,3 +67,129 @@ func (vService *VirtualService) IsValidHost(namespace string, serviceName string
...
+
+	if routes, isSlice := vService.Spec.Http.([]interface{}); isSlice {
+		for _, route := range routes {
+			if routeMap, isMap := route.(map[string]interface{}); isMap {
+				if _, hasTimeout := routeMap["timeout"]; hasTimeout {
+					return true
+				}
+			}
+		}
+	}
```
nrfox, line 85: imo ideally the `VirtualService` model would be fully typed and then these methods just performing type checking would no longer be necessary. I'm happy to make that change and remove these methods. Could Kiali simply import and use the upstream istio api types e.g. something like `networkingv1beta1.VirtualService`?

graph/types.go
```diff
@@ -8,24 +8,18 @@ import (
...
-	NodeTypeUnknown               string = "unknown" // The special "unknown" traffic gen node
-	NodeTypeWorkload              string = "workload"
-	TF                            string = "2006-01-02 15:04:05" // TF is the TimeFormat for timestamps
-	Unknown                       string = "unknown"             // Istio unknown label value
-	WizardFaultInjectionLabel     string = "fault_injection"
-	WizardLabelKey                string = "kiali_wizard" // Label the front-end wizards add to objects created by them
-	WizardRequestRoutingLabel     string = "request_routing"
-	WizardRequestTimeoutsLabel    string = "request_timeouts"
-	WizardTCPTrafficShiftingLabel string = "tcp_traffic_shifting"
-	WizardTrafficShiftingLabel    string = "traffic_shifting"
```
jshaughn, line 28: Easy come, easy go :-)

models/virtual_service.go
jshaughn: This is probably a better question for @lucasponce .  I think perhaps the models are parsed out the way they in an effort to simplify the structure to something more easy for us to deal with at higher levels.  Perhaps this badging stuff could be parsed out immediately, or stored lazily on the model, to possibly avoid doing the work many times.

jshaughn: @nrfox Now that @lucasponce is on PTO, I think we should leave this alone until he comes back and can comment.

jshaughn: I will test this as-is.

jshaughn (approved):
Looks good to me.  The model impl can be discussed later with @lucasponce, and perhaps we could look at parsing out some of this info directly into the model, I'm not sure.  But this seems to be working fine as-is.

image

pbajjuri20 (approved):
Tested and verified  on Kiali UI 1.38.0 (156692a97f02b2d82716e2f78deb49913fdd9598)


kiali_wizard

Issue #3061, opened 2020-07-31
Validations: Support `exportTo` field


xeviknal:
As of today, kiali validations doesn't consider the `exportTo` field. This field allows users to specify the scope of each config.

> This feature provides a mechanism for service owners and mesh administrators to control the visibility of services across namespace boundaries.

One example:

```yaml
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: external-svc-https
  namespace: bookinfo
spec:
  hosts:
  - api.dropboxapi.com
  location: MESH_EXTERNAL
  ports:
  - number: 443
    name: https
    protocol: TLS
  resolution: DNS
  exportTo:
  - '.'
  - foo
  - bar
```

This service entry is only available for `bookinfo, foo, bar` namespaces only.

Validations has to be adapted to filter those config by the exportTo field. This might trigger thoughts of cross-namespace validations.

- [x] #4314
- [x] #4315
- [x] #4316
- [x] #4317

xeviknal:
I am clarifying this issue.

The goal of this issue is to make sure that the validations are using the objects visible for that namespace. The model of visibility that the `exportTo` introduced made possible that objects that have visibility mesh-wide can only be visible in a very specific list of namespaces.

Using the `ServiceEntry` of the description. Any workload running within the bookinfo namespace can't call the `api.dropboxapi.com` because it is not exported.

Therefore, in the validations we shouldn't be managing entities that are not visible in that namespace.

hhovsepy:
All epic subtasks are done.

PR #4185, opened 2021-07-16
ExportTo Cross Namespace validations added: Initial commit

hhovsepy:
Added cross namespace validations for VirtualService, DestinationRule and ServiceEntry exportTo namespaces.

RFE: https://github.com/kiali/kiali/issues/3061


business/istio_validations.go
```diff
@@ -408,15 +408,8 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
 		errChan2 := make(chan error, 5)
 		istioDetails := kubernetes.IstioDetails{}
 
-		if IsResourceCached(namespace, kubernetes.VirtualServices) {
```
xeviknal, line 411: I'd probably leave this method as it is. Because the VS/DR fetched in here are the base of the validations (at namespace level). If you modify this, then you might end up validating VS/DR that are not in the namespace but they are accessible from the namespace.

I propose you to create a new method just fetching VS/DR/ServiceEntries that are available for the ns of the validation scope. The resulting arrays of VS/DR/SE will be sent in each checker validation.

For example:

```golang
go in.fetchExportedResources(&exportedResources, namespace, errChan, &wg)

checkers.VirtualServiceChecker{Namespace: namespace, Namespaces: namespaces, DestinationRules: exportedResources.DestinationRules, VirtualServices: istioDetails.VirtualServices, ExportedVirtualServices: exportedResources.VirtualServices}

checkers.DestinationRulesChecker{Namespaces: namespaces, DestinationRules: istioDetails.DestinationRules, MTLSDetails: mtlsDetails, ServiceEntries: exportedResource.ServiceEntries}
```

Note that the checkers might need to have an extra field for the exported Resources or not. Depending on each validation. Let's take `VirtualServiceCheckers` for instance. If it contains a checker that needs to compare each virtual services with other virtual services, then we would need to compare the Virtual Service (subject of the validation) of the namespace with all the virtual services exported to that namespace. Does it make sense?

business/istio_validations.go
```diff
@@ -480,6 +473,62 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+		nss, err := in.businessLayer.Namespace.GetNamespaces()
+		if err != nil {
+			errChan <- err
+			return
+		}
+
+		currentVSs, err := in.fetchVirtualServices(namespace)
+		if err != nil {
+			select {
+			case errChan <- err:
```
xeviknal, line 488: I guess you can simply do `errChan <- err` as you do in the L481.

business/istio_validations.go
```diff
@@ -480,6 +473,62 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+func (in *IstioValidationsService) fetchAllVirtualServices(istioDetails *kubernetes.IstioDetails, namespace string, errChan chan error, wg *sync.WaitGroup) {
+	defer wg.Done()
+	if len(errChan) == 0 {
+		nss, err := in.businessLayer.Namespace.GetNamespaces()
+		if err != nil {
+			errChan <- err
+			return
+		}
+
+		currentVSs, err := in.fetchVirtualServices(namespace)
```
xeviknal, line 485: If you'd like to follow the approach in my previous comment and thinking this method as the `fetchExportedResources` method, I guess that this is not necessary anymore.
The logic should go within the iteration of the `nss`.

business/istio_validations.go
```diff
@@ -480,6 +473,62 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+		if err != nil {
+			select {
+			case errChan <- err:
+			default:
+			}
+		} else {
+			for _, ns := range nss {
+				vsList, err := in.fetchVirtualServices(ns.Name)
+				if err != nil {
```
xeviknal, line 494: In go it is pretty common to avoid huge indentation by applying `early returns`. So I'd encourage you to use the `continue` statement and remove the `else` clause.

Plus, I think you don't need to use the `select` clause to send the errors to a channel.

business/istio_validations.go
hhovsepy: +1

business/istio_validations.go
```diff
@@ -480,6 +473,62 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+					select {
+					case errChan <- err:
+					default:
+					}
+				} else {
+					for _, vs := range currentVSs {
```
xeviknal, line 500: Shouldn't we use the `vsList`? Because we want to go through all VS in the mesh (or available ns). Plus, I guess for each iteration on the `nss` loop, we'd be checking the exact array of VS (from the current ns).

business/istio_validations.go
```diff
@@ -480,6 +473,62 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+						var namespaceFound = false
```
xeviknal, line 501: It sounds non-idiomatic go in here. I'd go for `namespaceFound := false`. However, I guess it is matter of style, tbh.

hhovsepy: yes, python style :)

business/istio_validations.go
```diff
@@ -480,6 +473,62 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+						}
+						if namespaceFound {
+							istioDetails.VirtualServices = append(istioDetails.VirtualServices, vsList...)
+						}
+					}
+				}
+			}
+		}
+	}
+}
```
xeviknal, line 522: One thing we should consider is to add test to this enhancement. Sooner than later. I guess that thinking the tests upfront would help you better understand the feature from the bottom to the top. For example, add a validation for virtual services that fails when there is one virtual service outside of the ns but exported to that ns.

business/istio_validations.go
hhovsepy: 'vsList' is the list of VS from exportedTo 'namespace'.
So here it loops through the VS from current namespace, and finds a VS which has 'exportTo' fields.

business/istio_validations.go
```diff
@@ -148,6 +151,7 @@ func (in *IstioValidationsService) GetIstioObjectValidations(namespace string, o
 	wg.Add(9)
```
xeviknal, line 151: I think you should increase this to 10.
The [working group](https://gobyexample.com/waitgroups) `wg` is used to manage the concurrence. Each `go routine` uses it for the sync `defer wg.Done()`. If there are more Done() than the expected, then you've got the following error:

```
panic: sync: negative WaitingGroup counter
```

business/istio_validations.go
```diff
@@ -480,6 +485,102 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+		seList, err := in.fetchServiceEntries(ns.Name)
+		if err != nil {
+			errChan <- err
+			return
+		}
+		if containsExportToNamespacesIstioObjects(ns, currentSEs) {
+			exportedResources.ServiceEntries = append(exportedResources.ServiceEntries, seList...)
+		}
+	}
+}
```
xeviknal, line 542: For me to understand, what the `exportedResources` will hold at the end of the call?

I see that if one of the VirtualServices living in the `namespace` (currentVSs) has the `exportTo` field matching to this very same namespace (aka `containsExportToNamespacesIstioObject` returns true) then you are adding all the VirtualServices of all namespaces into the `exportedResources.VirtualServices`. Because `containsExportToNamespacesIstioObjects(ns, currentVSs)` will have always the same value in each iteration. Therefore, I'd append each `vsList` in each iteration.

So in 99% of the cases, the `exportedResources` would have all the `VirtualServices` in the cluster because 99% of the times there will be one VirtualService with no `exportedTo` field (meaning that is available to all the namespaces). And so, I'd include VirtualServices in other namespaces not exported to this namespace.

business/istio_validations.go
```diff
@@ -480,6 +485,102 @@ func (in *IstioValidationsService) fetchDetails(rValue *kubernetes.IstioDetails,
...
+
+func containsExportToNamespacesIstioObjects(ns models.Namespace, currentIstioObjects []kubernetes.IstioObject) bool {
+	for _, vs := range currentIstioObjects {
+		if exportToSpec, found := vs.GetSpec()["exportTo"]; found {
+			if namespaces, ok := exportToSpec.([]interface{}); ok {
+				for _, exportToNs := range namespaces {
+					// take only namespaces where it is exported to, or if it is exported to current namespace
+					if (exportToNs == "." && ns.Name == vs.GetObjectMeta().Namespace) || exportToNs == ns.Name || exportToNs == "*" {
```
xeviknal, line 550: What if the IstioObject doesn't have the `exportTo` field?
Theoretically, when it doesn't exist it means that is exported to all the namespaces[0].

[0] https://istio.io/latest/docs/reference/config/networking/virtual-service/#VirtualService

business/istio_validations.go
```diff
@@ -165,13 +169,13 @@ func (in *IstioValidationsService) GetIstioObjectValidations(namespace string, o
 			checkers.GatewayChecker{GatewaysPerNamespace: gatewaysPerNamespace, Namespace: namespace, WorkloadsPerNamespace: workloadsPerNamespace},
 		}
 	case kubernetes.VirtualServices:
```
xeviknal, line 171: Right now, in the validations we pass from validating objects living in the same namespace to validate object from other namespaces and exported to the current namespace.

The translation of the above sentence into code is that that all the checkers which uses VS, DR, SE now will use the exported VS, DR, SE instead. Because it is not enough with the ones just living in the same namespace.

On the other side, the VS, DR, SE checkers do need two inputs: one is the subject to validate (which remains the same: the ones living in the ns) + the exported ones.
If you'd mixed both inputs you'd end up validating both subjects and exported. Which is not necessary.

So I guess that you'd need to add the exported resources to the SidecarChecker, AuthorizationChecker, PeerAuthnChecker and other checkers which have VS/DR/SE as input.

xeviknal: Same pattern applies in the namespace-validation method: https://github.com/kiali/kiali/pull/4185/files#diff-4afe91ed13da4c9e145fdccf0596bee0b6c0fc2e29d3da22f112584428159dd3R113

business/checkers/virtual_service_checker.go
```diff
@@ -33,7 +35,7 @@ func (in VirtualServiceChecker) Check() models.IstioValidations {
 func (in VirtualServiceChecker) runIndividualChecks() models.IstioValidations {
```
xeviknal, line 35: I think the VirtualServicesChecker is the best checker to explain a bit the architecture of the validations.
One ObjectChecker runs two different validations: `IndividualChecks` and `GroupChecks`.
- `IndividualChecks` are all the checkers that are focused on one single Object. For example, checking if the `exportTo` field has valid values. It doesn't consider other Istio Objects to see whether there are overlaps between them. Just one object and other resources to determine its validity.
- `GroupChecks` are all the checkers that need a group of objects of the same type to determine its validity. It might end up with one validation for more than one object. For instance, [KIA0201](https://kiali.io/documentation/latest/validations/#_kia0201_more_than_one_destinationrules_for_the_same_host_subset_combination). It need to parse a group of DRs in order to see that there are multiple ones defined for the same host.

xeviknal: Therefore, appending both exported VS to the subject of analysis VS will make Kiali validate both sets of VS. And we just want to analyze the VS living in the current ns aka `in.VirtualServices`.

business/checkers/virtual_service_checker.go
```diff
@@ -45,7 +47,7 @@ func (in VirtualServiceChecker) runGroupChecks() models.IstioValidations {
 	validations := models.IstioValidations{}
 
 	enabledCheckers := []GroupChecker{
-		virtual_services.SingleHostChecker{Namespace: in.Namespace, Namespaces: in.Namespaces, VirtualServices: in.VirtualServices},
+		virtual_services.SingleHostChecker{Namespace: in.Namespace, Namespaces: in.Namespaces, VirtualServices: append(in.VirtualServices, in.ExportedVirtualServices...)},
```
xeviknal, line 50: In this case, I guess that we should go case by case into each checker to see whether we can append both sets of VS. We should understand the nature of the validation and see if it makes any sense.
I'd say that most of them would be completely alright to merge VS, but again, we should go case by case and most definitely adding test cases to make sure all is green.

xeviknal: The two previous comments apply to all the ObjectCheckers (DR, VS, SE, GW, SC, etc.). I know, it is pretty tough, but we should be super accurate with this.

business/istio_validations.go
```diff
@@ -2,17 +2,15 @@ package business
...
-
 	"github.com/kiali/kiali/business/checkers"
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
+	apps_v1 "k8s.io/api/apps/v1"
+	core_v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"sync"
```
xeviknal, line 13: You should run the `make lint format` to fix this unsorted imports.

business/checkers/destination_rules_checker.go
```diff
@@ -10,10 +10,12 @@ import (
...
-	DestinationRules []kubernetes.IstioObject
-	MTLSDetails      kubernetes.MTLSDetails
-	ServiceEntries   []kubernetes.IstioObject
-	Namespaces       []models.Namespace
+	DestinationRules         []kubernetes.IstioObject
+	ExportedDestinationRules []kubernetes.IstioObject
+	MTLSDetails              kubernetes.MTLSDetails
+	ServiceEntries           []kubernetes.IstioObject
+	ExportedServiceEntries   []kubernetes.IstioObject
+	Namespaces               []models.Namespace
```
xeviknal, line 18: I'd remove this from this branch unless you want to add the validations. It'd reduce dead code and wouldn't confuse us. To me this is a partial implementation. Otherwise, we should add the exported objects to all the other ObjectCheckers (because besides SE, DR and VS, other Object checkers use objects which have exportTo resources).
+1 to remove dead code from `DestinationRulesChecker` and `ServiceEntryChecker`

business/checkers/virtualservices/single_host_checker.go
```diff
@@ -124,8 +131,15 @@ func storeHost(hostCounter map[string]map[string]map[string]map[string][]*kubern
...
-	namespace, clusterName := virtualService.GetObjectMeta().Namespace, virtualService.GetObjectMeta().ClusterName
+func (s SingleHostChecker) getHosts(virtualService kubernetes.IstioObject, exportedNamespace *string) []kubernetes.Host {
+	var namespace string
+	if exportedNamespace != nil {
+		// exported to VirtualServices considered as from current namespace
+		namespace = *exportedNamespace
+	} else {
+		namespace = virtualService.GetObjectMeta().Namespace
+	}
+	clusterName := virtualService.GetObjectMeta().ClusterName
```
xeviknal, line 142: We definitely need to add tests about this new scenario.
I think we should see the current tests passing in green (after adding empty exported resources).
And add exported resources from other namespaces to see it working and failing too.
Tests are pretty crucial here otherwise we won't know which scenarios are covered and which aren't.

xeviknal: Plus, test will give me a bit more clarity about what this changes are expected to cover.

business/istio_validations.go
```diff
@@ -108,11 +108,11 @@ func (in *IstioValidationsService) getServiceCheckers(namespace string, services
...
 	return []ObjectChecker{
 		checkers.NoServiceChecker{Namespace: namespace, Namespaces: namespaces, IstioDetails: &istioDetails, Services: services, WorkloadList: workloads, GatewaysPerNamespace: gatewaysPerNamespace, AuthorizationDetails: &rbacDetails, RegistryStatus: registryStatus},
 		checkers.VirtualServiceChecker{Namespace: namespace, Namespaces: namespaces, DestinationRules: istioDetails.DestinationRules, VirtualServices: istioDetails.VirtualServices},
-		checkers.DestinationRulesChecker{Namespaces: namespaces, DestinationRules: istioDetails.DestinationRules, MTLSDetails: mtlsDetails, ServiceEntries: istioDetails.ServiceEntries},
+		checkers.DestinationRulesChecker{Namespaces: namespaces, DestinationRules: istioDetails.DestinationRules, MTLSDetails: mtlsDetails, ServiceEntries: istioDetails.ServiceEntries, ExportedDestinationRules: exportedResources.DestinationRules, ExportedServiceEntries: exportedResources.ServiceEntries},
```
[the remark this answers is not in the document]
hhovsepy: you are right

business/checkers/virtualservices/single_host_checker.go
xeviknal: I still think that this is incorrect. See the following example and how istio behave:

```yaml
kind: DestinationRule
metadata:
  name: reviews
  namespace: bookinfo # Note that this live in the bookinfo
spec:
  host: reviews
  trafficPolicy:
    tls:
      mode: ISTIO_MUTUAL
  subsets:
    - labels:
        version: v1
      name: v1
    - labels:
        version: v2
      name: v2
    - labels:
        version: v3
      name: v3
---
kind: VirtualService
metadata:
  name: reviews-cp-2
  namespace: default # Note that this live in default
spec:
  hosts:
    - reviews # as per documentation, this host is the same as the FQDN: reviews.default.svc.cluster.local
  http:
    - route:
        - destination:
            host: reviews
            subset: v1
          weight: 90
        - destination:
            host: reviews
            subset: v2
          weight: 10
  exportTo:
    - bookinfo # It is exported to Bookinfo
```

Kiali shows this behavior:

Screenshot of Kiali Console (16)

Which in my opinion respond to this documentation note:
> Note for Kubernetes users: When short names are used (e.g. “reviews” instead of “reviews.default.svc.cluster.local”), Istio will interpret the short name based on the namespace of the rule, not the service.

IMO, this `getHosts` method shouldn't need the external namespace. It needs the namespace of the rule itself which is already present in the rule definition.

business/checkers/virtualservices/single_exported_host_checker_test.go
```diff
@@ -0,0 +1,621 @@
+package virtualservices
+
+import (
```
xeviknal, line 3: I see there is no usage of the `exportTo` field in this test set. Should we at least try the feature of exporting to "."?

xeviknal: Okay, I see. There is no need to use the `exportTo` explicitely because the VS are appropriately added into the ExportedVirtualServices. It make sense! :+1:

business/checkers/virtualservices/single_host_checker.go
hhovsepy: The result of getHosts method is used in storeHost method to prepare a map between hosts and VirtualServices.
If it uses the original namespace of rule while ignoring the namespace where this rule is exported, then this rule will not be considered in a validations as an imported rule from other namespace.

hhovsepy: In the case of exportTo, then right the host should be FQDN.
Screenshot from 2021-08-10 11-21-41
Screenshot from 2021-08-10 11-22-13

hhovsepy: But in your example @xeviknal the "reviews-cp-2" is ignored and Istio generates equal traffic between subsets.

hhovsepy: Just try to delete VS and leave only DR. The traffic is equally distributed between subsets.

hhovsepy:
Bookinfo2 reviews is exported to everywhere:
Screenshot from 2021-08-10 17-06-36

bookinfo3 reviews is exporting to own namespace only, but is imports from bookinfo2, so validation warning with reference is there:
Screenshot from 2021-08-10 17-08-03

hhovsepy:
Hosts should be in FQDN, as recommended in Istio.
Screenshot from 2021-08-11 10-54-05
Screenshot from 2021-08-11 10-54-27

xeviknal (approved):
Great step! It looks great to me. All the changes are approached. Tests passing.
This PR only approaches the support of `exportTo` to the KIA1106 validation. In a subsequent PR more validation will be covered.

xeviknal:
@prachiyadav I completely forgot to add QE to verify this PR. Sorry for that. Do you mind doing a port-merge verification?

prachiyadav:
Test Case : SWSQE-1876

prachiyadav:
Approved after merge

Screenshot from 2021-08-16 13-54-56

Issue #3895, opened 2021-04-16
[molecule] run molecule tests using a KinD cluster


jmazzitelli:
PRs:
* https://github.com/kiali/kiali/pull/3894
* https://github.com/kiali/kiali-operator/pull/291

jmazzitelli:
This is currently a WIP. With the two PRs I'm close, but I need to figure out how to get the molecule docker container (which runs the molecule tests) to be able to access the Kiali service in the KinD cluster - currently getting a "Host unreachable" error. I'm pretty sure this can be done, but at the moment I do not know how.

jmazzitelli:
Here is the state of the 2 PRs.

Running this command to test:
```
hack/ci-kind-molecule-tests.sh -kof jmazzitelli -kob ci-kind -kf jmazzitelli -kb ci-kind -ir '' -udi true -at default
```
Will start running the "default" molecule scenario. But this fails at the task:
```
    TASK [Wait for Kiali to be running and accepting requests] *********************^M
```
with this error:
```
Status code was -1 and not [200]: Request failed: <urlopen error [Errno 113] Host is unreachable>
```
It is trying to get to the correct url (in this case, `"url": "https://172.18.255.71/kiali/api"`) but that IP is not reachable. I can confirm also by just getting inside the container and trying to curl that URL:
```
$ docker ps | grep kiali-molecule
752ef46da58a   kiali-molecule:latest   "molecule --debug te…"   3 minutes ago   Up 3 minutes   sharp_bardeen

$ docker exec -it sharp_bardeen bash
bash-5.1# curl https://172.18.255.71/kiali/api
curl: (7) Failed to connect to 172.18.255.71 port 443: Host is unreachable
```

We need to figure out how to open up the network from the molecule container to the Kind container because the molecule tests need to send requests to the server-under-test running in Kind.

jmazzitelli:
I no longer have a machine that can run KinD (I'm on RHEL8 that has podman only and it cannot run kind.) So I'm putting this effort on pause. If someone can pick up the work and complete this, it would be appreciated.

jshaughn:
> I'm on RHEL8 that has podman only and it cannot run kind.

Are you saying Kind requires Docker and Podman to run?

jmazzitelli:
I'm saying Kind neither either docker OR podman, but not all podmans are created equal. My podman installation (for reasons unknown to me) is not good enough to work with the experimental podman support in Kind:

```
$ kind create cluster
[...]

PR #4189, opened 2021-07-16
Run molecule tests with KinD

israel-hdez:
This is #3894 updated to changes in `master` and conflicts fixed.

Additionally, there is a change in the way the repos are mounted into the container running the molecule tests: use volumes rather than mounts, and use full fork names.

* Prefer using volumes rather than mounting directories on the host
machine into the container. Although this was mainly to add
compatibility for using a remote docker engine, in general, this
potentially works better. The rationale is that there is no need to
adjust privileges nor SELinux on the host for the container to work
properly, because the container engine is fully managing the volume.
* Use forks's fullname rather than just the username/orgname that has
the fork. This helps with older forks that have the old swscore name.

Related to https://github.com/kiali/kiali/issues/3895


hack/ci-kind-molecule-tests.sh
```diff
@@ -0,0 +1,441 @@
...
+  infomsg "Sending IRC notification to room [#${IRC_ROOM}]. msg=${irc_msg}"
+  (
+  echo 'NICK kiali-test-bot'
+  echo 'USER kiali-test-bot 8 * : kiali-test-bot'
+  sleep 10
+  echo "JOIN #${IRC_ROOM}"
+  sleep 5
+  echo "PRIVMSG #${IRC_ROOM} : ${irc_msg}"
+  echo QUIT
+  ) | nc irc.freenode.net 6667
```
jmazzitelli, line 440: make sure to change this - going to libera now. See: https://github.com/kiali/kiali/blob/master/hack/ci-openshift-molecule-tests.sh#L441
```suggestion
  ) | nc irc.libera.chat 6667
```

jmazzitelli (changes requested):
minor change (chat room we log the results to changed).

When this is all ready, I'll have to test this on my machines. Large change to use the volumes - need to make sure this works on my nightly build machine as well as RHEL 8 podman3.

hack/ci-kind-molecule-tests.sh
```diff
@@ -0,0 +1,441 @@
...
+HELM_FORK="${HELM_FORK:-kiali/helm-charts}"
+HELM_BRANCH="${HELM_BRANCH:-master}"
+KIALI_FORK="${KIALI_FORK:-kiali/kiali}"
+KIALI_BRANCH="${KIALI_BRANCH:-master}"
+KIALI_OPERATOR_FORK="${KIALI_OPERATOR_FORK:-kiali/kiali-operator}"
+KIALI_OPERATOR_BRANCH="${KIALI_OPERATOR_BRANCH:-master}"
+
+# details about the github repo where the logs are to be stored
+LOGS_PROJECT_NAME="${LOGS_PROJECT_NAME:-kiali-molecule-test-logs}"
+LOGS_FORK="${LOGS_FORK:-jmazzitelli}"
```
jmazzitelli, line 186: if we run this from github CI, we'll want to change this to somewhere else where we can store the logs - probably under the kiali umbrella

israel-hdez:
With this PR and with kiali/kiali-operator#363 I'm getting `success` on all tests that run by default - I didn't check the tests that are skipped by default. That includes your new "session timeout" test that was recently merged (https://github.com/kiali/kiali-operator/pull/366).

So, with those PRs we are able to run molecules on KinD. Once they are merged, I'll create another PR to add a GitHub actions workflow to run them.

hack/ci-kind-molecule-tests.sh
israel-hdez: I would prefer printing logs to the terminal, so that you can go to the _Checks_ tab of GitHub and inspect the logs there like any regular CI, rather than going to another place to look for the logs.

jmazzitelli (approved):
I ran a couple molecule tests with minikube and OpenShift (running OSSM 2.1) and all passed.

PR #4198, opened 2021-07-21
Add new GHActions CI job to run molecule tests

israel-hdez:
Molecules are run only when pushes are performed on `master` branch
(i.e. after PRs are merged). This prevents CI to block merges.

Small changes to hack scripts are done to allow them to print logs
messages to the console. This is to allow inspection of logs in case of
failure.

Fixes kiali/kiali#3895


israel-hdez:
There is an ongoing test run here: https://github.com/israel-hdez/swscore/pull/21/checks?check_run_id=3127193048

At the moment, I see:

```
=====================
=== TEST RESULTS: ===
=====================

              accessible-namespaces-test... success [5m 5s]
     affinity-tolerations-resources-test... success [4m 41s]
                                api-test... success [4m 36s]
```

I saw that past log lines are not available while the CI is still running. But once it finishes, all generated logs are made available.

A previous run shown that it takes about 1hour for all default molecules to finish. Quite a lot. So, I'm limiting molecules to be ran when PRs are merged into master. In case of failure, I expect GitHub to send an e-mail notification, which should have a link to the failed run where we can inspect the generated logs.

israel-hdez:
My only pending problem to solve is that, if I'm not wrong, this will pull kiali images with the `latest` tag, which is probably not what we want in CI.

This, probably, should pull `edge` images. @jmazzitelli any parameter to specify this? But then `edge` images may not be updated when tests are run. It's a little a chicken-egg problem. Still, it's better than nothing.

hack/run-molecule-tests.sh
```diff
@@ -354,6 +354,12 @@ do
     printf '%s [%s]' "$(green 'success')" "${duration}"
   else
     printf '%s [%s]' "$(red 'FAILURE')" "${duration}"
+
+    # If we are in CI, abort fast. Also, print logs of the failed test to allow for inspection
+    if [ "${CI}" == "true" ]; then
+      cat ${TEST_LOGS_DIR}/${t}.log
+      exit 1
+    fi
```
jmazzitelli, line 362: This is no good - because the "exit 1" is going to exit the entire script so it won't continue to run the other molecule tests. So if the very first molecule test fails, the entire rest of the molecule test suites never run. We don't want that. We want all of them to run so we can see which molecule tests fail and which succeed.

This is why we already have the ability to store the logs in a logs repo. In this script you already have this feature - you just have to enable it and prepare a logs repo. See:

* https://github.com/kiali/kiali/blob/master/hack/ci-kind-molecule-tests.sh#L87-L93
* https://github.com/kiali/kiali/blob/master/hack/ci-kind-molecule-tests.sh#L123-L128

You can pair this together with this other script I have so we retain the last N sets of molecule test logs in the logs repo: https://github.com/jmazzitelli/mazz-basement/blob/master/server-scripts/purge-directories-from-git.sh

I use this on mazz-basement by running the openshift molecule tests like this:

```
/home/jmazzite/source/kiali/kiali/src/github.com/kiali/kiali/hack/ci-openshift-molecule-tests.sh -dorp podman -ir kiali -ul true > /tmp/KIALI-TESTS.log 2>&1 && /home/jmazzite/source/mazz-basement/server-scripts/purge-directories-from-git.sh --do-not-push false >> /tmp/KIALI-TESTS.log 2>&1
```


The result in the logs repo will look like this (this is the mazz-basement test results - notice there is only ever the last 10 nightly runs that are retained): https://github.com/jmazzitelli/kiali-molecule-test-logs

You can see all the directories there have a "results.log" - that shows you all the molecule tests and their success/failure (and this is why your update here isn't good - as soon as one of those tests results in FAILURE, no other tests will run after that).

jmazzitelli (changes requested):

jmazzitelli:
> My only pending problem to solve is that, if I'm not wrong, this will pull kiali images with the `latest` tag, which is probably not what we want in CI.

Yes, that's the default. There is MOLECULE_USE_DEV_IMAGES, but you need to push newly build images into the kind cluster in order for that to work. You would need the script (or something before the script runs) to actually build and push the images (`make -e CLUSTER_TYPE=kind clean build test cluster-push`) into an already-running kind cluster.

israel-hdez:
> Yes, that's the default. There is MOLECULE_USE_DEV_IMAGES, but you need to push newly build images into the kind cluster in order for that to work. You would need the script (or something before the script runs) to actually build and push the images (`make -e CLUSTER_TYPE=kind clean build test cluster-push`) into an already-running kind cluster.

So, it needs some preparation:
* Something needs to checkout the kiali sources and also the kiali-operator sources.
* Then, something needs to build both kiali server and the operator and create container images for them
* Then, something needs to pre-create a cluster and push the created images to this cluster.
* Then, the script runs. Curiously, the already checked out sources cannot be used, because the script wants an empty directory. So, it will checkout again and work on the newly checked out sources :-/

I need to think more about it. The first two steps would be a duplicated effort, because Jenkins is already doing that on every push to `master` and publishing the built containers to quay. So, I think it's better to run molecules on the finished `edge` packages, rather than building a private one.

But then, Jenkins needs to finish first, which may mean that Jenkins may need to trigger the molecules rather than watching for pushes to `master`. To some extent, it makes sense, because Jenkins is already watching changes on all repos and it could launch molecules each time a push is done to any of the `kiali/*` repos and artifacts are properly published (perhaps excluding kiali.io), while this PR would be listening for changes only in `kiali/kiali`.

jmazzitelli:
> So, I think it's better to run molecules on the finished `edge` packages, rather than building a private one.

I agree.

> Jenkins may need to trigger the molecules ... Jenkins is already watching changes on all repos and it could launch molecules each time a push is done to any of the `kiali/*` repos and artifacts are properly published (perhaps excluding kiali.io), while this PR would be listening for changes only in `kiali/kiali`.

+1.  And this is important to do because changes to the kiali-operator repo and/or helm repo can just as equally cause a regression and test failure. Catching those problems early would be helpful.

lucasponce:
Folks, this PR is open for 1.5 months.
If it can't move, I'd recommend to move it to "Draft", or mark if it's waiting for some external dependency.

Issue #4197, opened 2021-07-21
Feature Request: Disable certain validations


linjmeyer:
**Is your feature request related to a problem? Please describe.**

There are some validations that do not apply to our installation, and for us would be considered false positives.  For example: KIA0203 - This subset's labels are not found in any matching host.  Our CD pipeline controls VirtualServices and DestinationRules, but the DRs often have references to pods that don't exist (e.g. a canary pod) but when this happens the VS would not reference the DR subset anyway, so it's not a problem.

**Describe the solution you'd like**

Ability to disable certain validations by their ID (e.g. "KIA0203") both:
* Per resource
* Cluster wide

**Describe alternatives you've considered**

I am not aware of any alternatives or workarounds to disable certain validations. 

**Additional context**

Discussion with @jmazzitelli in Slack and they proposed the ability to do this via the Kiali CRD (for cluster wide) and annotations for specific resources, this approach would work well in my opinion:

Operator example:
```
spec.
  disable_validations: ['KIA0203', 'KIA0412 ']
```

annotation example:
```
kind: DestinationRule
metadata
  annotations:
    kiali.io/disable-validations: "KIA0203, KIA0412"
```

jmazzitelli:
If we do think it is a good idea to have a validations disabled list Kiali-wide, then I'm not sure where to put this in the Kiali CR - the place that makes the most sense seems to be `kiali_feature_flags`:

```
spec:
  kiali_feature_flags:
    validations:
      disabled:
      - KIA0203
      - KIA0412
```

So add a new section "validations" here. This would be the place to put settings for the validations subsystem. Under here we'd have a "disabled" section which is the list of those validations that should be disabled.

If we only want to support disabling validations on a per-resource basis (via annotations) then the above is moot since we won't have anything in the global Kiali CR config.

jmazzitelli:
@xeviknal I would like to hear your thoughts on this.

xeviknal:
Thank you @linjmeyer and @jmazzitelli 

I have a love and hate relationship with this idea of silent certain validations. I think it must be really annoying to keep seeing errors/warnings everywhere that you know that are not real problems. Plus, it reduces the effectiveness of the feature value: "was this error present before or is a new one I just created?" So I guess you can't rely on it that much. Which is throwing away an important kiali capability.

[...]

PR #4205, opened 2021-07-22
Be able to ignore one or more validation checks

jmazzitelli:
Some validations may fail that cannot be corrected (either it is an expected failure for some reason (see the associated issue for one such use-case) or Kiali validation has a bug and it shows an error where there should not be one).

This allows the user to ignore one or more validations (as specified by a validation code such as `KIA0101`) via the Kiali CR:

```
kiali_feature_flags:
  validations:
    ignore:
    - KIA0101
    - KIA0200
```

When a validation is ignored, we still would like users to know about the failure, especially if it is caused by a  Kiali bug because we want them to report it. So this PR will log the validation error at "info" level in the server pod logs. The log message will look like this:

```
2021-07-22T19:14:50-04:00 INF Ignoring validation failure [&{Code:KIA0101 Message:Namespace not found
for this rule Severity:warning Path:spec/rules[0]/from[1]/source/namespaces[1]}] for
object [authorizationpolicy:httpbin] in namespace [default]
```

When ignored, you can see the UI will not indicate any issues (look at spec.rules.from.source.namespaces field):

image

This is what it would have looked like if the validation was not ignored (there were 2 errors of the same kind of validation):

image

The tooltip will still show the same message (the code will be prefixed with the message):

image

(the above examples use the yaml found at https://kiali.io/documentation/latest/validations/#_kia0101_namespace_not_found_for_this_rule)

fixes: https://github.com/kiali/kiali/issues/4197

This will require PRs to both the UI frontend and the operator.
* ui: https://github.com/kiali/kiali-ui/pull/2208
* operator: https://github.com/kiali/kiali-operator/pull/369

### DIFF
--- a/business/checkers/common/check_confirmer.go
+++ b/business/checkers/common/check_confirmer.go
@@ -1,0 +1,18 @@
+package common
+
+import (
+	"fmt"
+
+	"github.com/kiali/kiali/models"
+)
+
+// ConfirmIstioCheckMessage can be used by the validation tests to confirm the Istio Check received is what was expected
+func ConfirmIstioCheckMessage(expectedIstioCheckKey string, actualIstioCheck *models.IstioCheck) error {
+	expected := models.CheckMessage(expectedIstioCheckKey)
+	actual := actualIstioCheck.GetFullMessage()
+	if expected == actual {
+		return nil
+	} else {
+		return fmt.Errorf("IstioCheck: expected [%s], actual [%s]", expected, actual)
+	}
+}
--- a/business/checkers/authorization/namespace_method_checker_test.go
+++ b/business/checkers/authorization/namespace_method_checker_test.go
@@ -6,10 +6,10 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
-	"github.com/kiali/kiali/business/checkers/common"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/tests/data"
+	"github.com/kiali/kiali/tests/testutils"
 )
 
 func TestSourceNamespaceExisting(t *testing.T) {
@@ -36,10 +36,10 @@ func TestSourceNamespaceNotFound(t *testing.T) {
 	assert.True(valid)
 	assert.NotEmpty(validations)
 	assert.Len(validations, 2)
-	assert.NoError(common.ConfirmIstioCheckMessage("authorizationpolicy.source.namespacenotfound", validations[0]))
+	assert.NoError(testutils.ConfirmIstioCheckMessage("authorizationpolicy.source.namespacenotfound", validations[0]))
 	assert.Equal(validations[0].Severity, models.WarningSeverity)
 	assert.Equal(validations[0].Path, "spec/rules[0]/from[0]/source/namespaces[0]")
-	assert.NoError(common.ConfirmIstioCheckMessage("authorizationpolicy.source.namespacenotfound", validations[1]))
+	assert.NoError(testutils.ConfirmIstioCheckMessage("authorizationpolicy.source.namespacenotfound", validations[1]))
 	assert.Equal(validations[1].Severity, models.WarningSeverity)
 	assert.Equal(validations[1].Path, "spec/rules[0]/from[0]/source/namespaces[1]")
 }
@@ -59,7 +59,7 @@ func TestToMethodWrongHTTP(t *testing.T) {
 	assert.NotEmpty(validations)
 	assert.Len(validations, 4)
 	for i, m := range []int{3, 4, 5} {
-		assert.NoError(common.ConfirmIstioCheckMessage("authorizationpolicy.to.wrongmethod", validations[i]))
+		assert.NoError(testutils.ConfirmIstioCheckMessage("authorizationpolicy.to.wrongmethod", validations[i]))
 		assert.Equal(validations[i].Severity, models.WarningSeverity)
 		assert.Equal(validations[i].Path, fmt.Sprintf("spec/rules[0]/to[0]/operation/methods[%d]", m))
 	}
--- a/business/checkers/authorization/namespace_method_checker_test.go
+++ b/business/checkers/authorization/namespace_method_checker_test.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/stretchr/testify/assert"
 
+	"github.com/kiali/kiali/business/checkers/common"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/tests/data"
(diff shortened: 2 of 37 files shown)


business/checkers/common/check_confirmer.go
```diff
@@ -0,0 +1,18 @@
+package common
+
+import (
+	"fmt"
+
+	"github.com/kiali/kiali/models"
+)
+
+// ConfirmIstioCheckMessage can be used by the validation tests to confirm the Istio Check received is what was expected
+func ConfirmIstioCheckMessage(expectedIstioCheckKey string, actualIstioCheck *models.IstioCheck) error {
```
xeviknal, line 10: I generally don't like to put test code next to production code. So I'd try to find a place in a `_test.go` file. For me it is a bit confusing. If you are alright with the suggestion, it should definitely be under `business` or `business/checkers`. What about creating the `business/checkers/checker_test.go` (which is where the base interfaces are defined) and placing there both the logic and its test together?

xeviknal (commented):
I am placing some comment I'd like to hear from you @jmazzitelli

business/checkers/common/check_confirmer.go
jmazzitelli: I tried every way I could think possible - the first thing I tried was exactly what you suggest (I created business/checkers/checker_test.go and it didn't work), and then I tried business/checkers/common/xyz_test.go - again, doesn't work). Nothing worked without getting `import cycle not allowed in test` errors. If you can figure out how to put test code in _test files that work across package boundaries without getting a circular dependency error in golang, please tell me how :) I googled and everybody says it's impossible.

jmazzitelli:
I'm going to take this out of draft mode - it works, and all tests pass. So we can open it up for review now

business/checkers/common/check_confirmer.go
nrfox: I've seen a few places having a `testing.go` file or some `testing`, `util/testing`, `testhelpers`... package to keep common test helpers. If they are never imported into the non-testing code I'm not sure if they get included in the final binary or not.

jmazzitelli: Right - they are still "outside" of the testing framework (just like I have it) but because they aren't used outside of _test code, it shouldn't get compiled in the final binary. So the way this PR has it right now, I believe this test function won't be in the final binary because it isn't used outside of test code (but that is only going by what people say in the Internets - I really have no idea).

Note that the name is irrelevant (people use "test" or "testing" just to make it human-understandable what it is for - but Go makes no distinction (e.g. see "httptest").

xeviknal:
Is the `info` log level set in default kiali installations? I mean, by default, users are going to see these logs you've added?
Otherwise, what do you thing about adding a warning during the `kiali` boot saying something like: "we've detected you are ignoring some validations. Please open an issue to kiali team if you already haven't opened one providing the details of your scenario. This way you keep kiali validations strong for the whole community".

Am I too scared? :fearful: :laughing:

jmazzitelli:
> Is the `info` log level set in default kiali installations? I mean, by default, users are going to see these logs you've added?

Yes, that is the default - they will see them.



> adding a warning during the `kiali` boot

That is easy to do. We can do that. We'll log a WARN to say that and list the ones they are ignoring. I'll add that

business/checkers/common/check_confirmer.go
xeviknal: @nrfox thanks for passing by. I was thinking of asking you to review this part since you've got some experience of go outside of this repo.

Okay, I get the `import cycle not allowed in test` problem. Perhaps aligned with @nrfox we could name the file differently to make it explicit that it contains test-only code. Otherwise, I see the common package and I think that each file in there is one checker (common - meaning that applies to different object while the folder destinationrules, for instance, only contains checkers exclusively thought for destinationrules. But each file listed is one checker. So it'd be good to keep the consistency). Then perhaps rename the file to `testing` and `testing_test` or `testhelpers` and `testhelpers_test` or even creating a generic folder for the utils.

xeviknal: Wait. we already have a test package with helpers for the test: https://github.com/kiali/kiali/tree/master/tests
In the data folder there are a bunch of helpers for creating fixtures for the test. We might one to create a utils folder with the helpers you created. wdyt?

jmazzitelli:
See the latest commit - added a warn message at boot time. Here's what the message will look like at boot time if one or more validations are to be ignored:

```
2021-07-23T11:21:00-04:00 WRN Some validation errors will be ignored [KIA0101 KIA0602].
If these errors do occur, they will still be logged. If you think these are false errors,
please report them to the Kiali team for review.
```

jmazzitelli:
> "we've detected you are ignoring some validations. Please open an issue to kiali team if you already haven't opened one providing the details of your scenario. This way you keep kiali validations strong for the whole community".

I didn't see this part of your comment - I'll change the message to be more like this

jmazzitelli:
@xeviknal what about this, also? These ignore settings are coming over to the UI for free as part of the /api/config endpoint:

```
  "kialiFeatureFlags": {
    "istioInjectionAction": true,
...
    "validations": {
      "ignore": [
        "KIA0101"
      ]
    }
  }
```

Because the UI already has this data, we could have the UI show a message in the masthead, or in the message center, or somewhere to indicate that things are being ignored. I don't know where.
This change would be part of the UI PR: https://github.com/kiali/kiali-ui/pull/2208

jmazzitelli:
updated boot-time warn message:

```
2021-07-23T11:48:00-04:00 WRN Some validation errors will be ignored [KIA0101]. If these errors do occur,
they will still be logged. If you think the validation errors you see are incorrect, please report them to the
Kiali team if you have not done so already and provide the details of your scenario.
This will keep Kiali validations strong for the whole community.
```

business/checkers/common/check_confirmer.go
jmazzitelli: I did notice that tests folder.. It actually has things in the data/ subdirectory - I wasn't sure what it was for. If that is where we put stuff in there, I should be able to move that file from common to here. I'll try it and see.

jmazzitelli: Ahh.. yes, the test_assert.go - I tried playing with that one (that's in test/data/validations) - I was getting those weird dependency errors in there, too. But that was early on in my coding. I'll revisit this. Not everything in this test/data directory involves creating test data - this test_assert.go is one example. But I won't refactor this stuff - I'll just try to put something in here - perhaps in the test/data/validations directory.

jmazzitelli: OK, I can already see problems with moving it under test/data/validations. I'm going to do this:

> We might one to create a utils folder with the helpers you created. 

I'll put a `testutil` under tests/ directory.

jmazzitelli: OK, see latest commit - moved that stuff out into a new `testutils` package located directly under the `tests` directory.

xeviknal: Sound good to me. The `testutils` package sounds the place to be.

xeviknal (approved):
LGTM!

Issue #2212, opened 2020-02-05
[operator] Kiali resource type status 


OchiengEd:
**Is your feature request related to a problem? Please describe.**
The Kiali resource that is managed by the operator has its current status held under the field `.status.type`. This results in the status being reported as Unknown in openshift dashboard or any other monitoring system that will check `.status.phase` like is used in other native kubernetes resources such as pods

```
$ oc -n servicemesh get jaeger jaeger -o json | jq .status.phase
"Running"
$ oc -n servicemesh get kiali kiali -o json | jq .status.phase
null
$ oc -n servicemesh get kiali kiali -o json | jq .status.conditions[0].type
"Running"
$ 

```

**Describe the solution you'd like**
Report the status under `.status.phase`

**Describe alternatives you've considered**
N/A

**Additional context**
Add any other context or screenshots about the feature request here. If it is about a new design, please attach a drawing of a potential solution.
For backend changes also describe how one could test the new feature and how it impacts display on the UI.

jmazzitelli:
This `status` field is dictated entirely by the ansible operator-sdk... if you feel this status field needs to change, the recommended approach is to report it to the operator-sdk folks:

https://github.com/operator-framework/operator-sdk/issues/new/choose

OchiengEd:
Thanks. Let me look into that.

OchiengEd:
@jmazzitelli It is possible to add custom status for an operand(or custom resource) managed by an ansible operator. Below are the resources used to arrive at this.

[1] https://github.com/operator-framework/operator-sdk/blob/e4d29c8ca1ff3ccc5ebef6f1695730b84cf8729e/pkg/ansible/controller/status/types.go#L153-L180
[2] https://github.com/operator-framework/operator-sdk/blob/master/doc/ansible/dev/developer_guide.md#custom-resource-status-management

jmazzitelli:
AFAIK, we don't have access to the status that the operator-sdk is putting inside the conditions[0].type. IMO, If this truly is a problem that OS console and other things can't see the status then it is a problem across the board for all ansible operators and they (operator-sdk folks) should know about this and fix it within the ansible operator-sdk base image.

OchiengEd:
The conditions are being handled as generic status. So am totally with you, operator-sdk handle those. However, nothing stops an operator developer from adding a `.status.phase` to hold the status of the operand.

It might be a challenge for operator-sdk to effectively have generic code to tell the status of all operands.

[...]

PR #407, opened 2018-08-07
KIALI-1210 DeploymentsList endpoint renamed to WorkloadList

xeviknal:
** Describe the change **

Renaming DeploymentList by WorkloadList.
URL listening: `api/namespaces/bookinfo/workloads`

screenshot of kiali-istio-system 192 168 42 123 nip io_api_namespaces_bookinfo_workloads

** Issue reference **

[KIALI-1210](https://issues.jboss.org/browse/KIALI-1210)

** Backwards incompatible? **
Yes.

PR #89, opened 2018-03-05
[SWS-213] errorRate should be calculated as rate, not percentage

burmanm:
The multiply by 100 should not happen.

PR #4227, opened 2021-07-28
Add a new section to get internal certificates information

### LINKED ISSUE #1577: Add indication if certificates are managed by Citadel or external tool
Add indication in UI if certificates are managed by Citadel or external tool

JIRA:
https://issues.jboss.org/browse/KIALI-3226

leandroberetta:
**Description**

This PR adds the possibility to get certificates information used by Istio in internal traffic (mTLS).

**Issue**

Resolves #1577. 

**Related PRs**

- UI: [#2212](https://github.com/kiali/kiali-ui/pull/2212)
- Operator: [#378](https://github.com/kiali/kiali-operator/pull/378)
- Helm Chart: [#89](https://github.com/kiali/helm-charts/pull/89)
- Kiali.io: [#407](https://github.com/kiali/kiali.io/pull/407)

**Documentation**

The following is an example of the API response when using Chiron:

```bash
curl http://localhost:20001/kiali/api/istio/certs
```

```json
[
    {
        "secretName": "dns.example1-service-account",
        "dnsNames": [
            "example1.istio-system.svc",
            "example1.istio-system"
        ],
        "issuer": "CN=minikubeCA",
        "subject": "O=",
        "notBefore": "2021-07-27T18:37:52Z",
        "notAfter": "2022-07-27T18:37:52Z",
        "error": null
    },
    {
        "secretName": "dns.example2-service-account",
        "dnsNames": [
            "example2.istio-system.svc",
            "example2.istio-system"
        ],
        "issuer": "CN=minikubeCA",
        "subject": "O=",
        "notBefore": "2021-07-27T18:37:52Z",
        "notAfter": "2022-07-27T18:37:52Z",
        "error": null
    }
]
```

And the following is how this information is presented on Kiali:

certs_info1
certs_info2


business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+
+type CertInfo struct {
+	SecretName string    `json:"secretName"`
+	DNSNames   []string  `json:"dnsNames"`
+	Issuer     string    `json:"issuer"`
+	Subject    string    `json:"subject"`
+	NotBefore  time.Time `json:"notBefore"`
+	NotAfter   time.Time `json:"notAfter"`
+	Error      error     `json:"error"`
+}
```
xeviknal, line 36: I think I'd move this struct into the `models` folder following the other examples there.
It'll be good to make all the logics better testable, modular, reusable and so on.

I think there you could implement the logics from parsing the secret to this struct. Little bit more object oriented.
```go
func (in *CertInfo) Parse(secret) {
	block, _ := pem.Decode(secret.Data["ca-cert.pem"])
	cert, err := x509.ParseCertificate(block.Bytes)
	if err != nil {
		in.SecretName = secretName
		in.Error: err
		return
	}
	
	in.Issuer = cert.Issuer.String()
	in.Subject = cert.Subject.String()
	in.NotBefore = cert.NotBefore
	in.NotAfter = cert.NotAfter
	in.SecretName= secretName
}
```

And you can re-use it everywhere in the project.

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+		NotBefore:  cert.NotBefore,
+		NotAfter:   cert.NotAfter,
+		SecretName: secretName,
+	}, nil
+}
+
+func (ics *IstioCertsService) getChironCertificates() ([]CertInfo, error) {
+	cfg := config.Get()
+
+	istioConfigMap, err := ics.k8s.GetConfigMap(cfg.IstioNamespace, "istio")
```
xeviknal, line 81: I think there is a method for picking the istio config map, but I might be wrong. Anyway, the magic string should be replaced by the config variable already present in the configmap.

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+					return
+				}
+				certChan <- CertInfo{
+					Issuer:     cert.Issuer.String(),
+					Subject:    cert.Subject.String(),
+					NotBefore:  cert.NotBefore,
+					NotAfter:   cert.NotAfter,
+					SecretName: secretName,
+					DNSNames:   dnsNames,
+				}
```
xeviknal, line 122: I think you can reuse this piece of code somehow. Less error-prone for modifications.

xeviknal (commented):
Good PR, good contribution.
I did add some comments that I'd like to hear your thoughts.

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+	Error      error     `json:"error"`
+}
+
+func (ics *IstioCertsService) GetCertsInfo() ([]CertInfo, error) {
+	if certs, err := ics.getChironCertificates(); err == nil {
+		return certs, nil
+	} else if cert, err := ics.getCertificateFromSecret("cacerts"); err == nil {
+		return []CertInfo{cert}, nil
+	} else {
+		cert, err := ics.getCertificateFromSecret("istio-ca-secret")
```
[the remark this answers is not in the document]
jmazzitelli: This is related to my comment in the operator PR. If the names are not fixed, this entire effort worries me. It requires Kiali to be given "read-secret" permissions across the entire cluster - and this is a security hole we have avoided giving the kiali server (even the operator doesn't have this permission). We cannot grant kiali cluster-wide "read-secret" permissions - community is not going to like it.

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+			return nil, err
+		}
+		return []CertInfo{cert}, nil
+	}
+}
+
+func (ics *IstioCertsService) getCertificateFromSecret(secretName string) (CertInfo, error) {
+	cfg := config.Get()
+
+	secret, err := ics.k8s.GetSecret(cfg.IstioNamespace, secretName)
```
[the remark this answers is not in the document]
jmazzitelli: Again, if the secret can be found in any namespace under any name, we can't do this - it would require the "read-secret" permission cluster-wide. And we do not want to grant that - especially if this is just for displaying secret information (it isn't like we need this permission for Kiali to "do its job" - it is just for displaying info to the user).

We must be very careful whenever we grant read-secret permission to Kiali - and we must be VERY VERY careful if we want to grant it cluster-wide - it opens up a potentially dangerous security hole. We don't want Kiali responsible for a security breach because it's service account was granted the ability to read secrets in the cluster.

jmazzitelli (dismissed):
See my earlier comments - we need a discussion around the read-secret permissions this PR is going to require. At this time, I am a -1 on this PR because of the potential security vulnerability this introduces.

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+	"encoding/pem"
+	"errors"
+	"log"
+	"sync"
+	"time"
+
+	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/kubernetes"
+	"gopkg.in/yaml.v2"
+)
```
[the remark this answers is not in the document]
leandroberetta: Gotcha!.

business/istio_certs.go
leandroberetta: +1000

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+	Subject    string    `json:"subject"`
+	NotBefore  time.Time `json:"notBefore"`
+	NotAfter   time.Time `json:"notAfter"`
+	Error      error     `json:"error"`
+}
+
+func (ics *IstioCertsService) GetCertsInfo() ([]CertInfo, error) {
+	if certs, err := ics.getChironCertificates(); err == nil {
+		return certs, nil
+	} else if cert, err := ics.getCertificateFromSecret("cacerts"); err == nil {
```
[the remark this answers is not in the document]
leandroberetta: It's not configurable, but I need to investigate what happens with canaries, thanks!.

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+		return []CertInfo{cert}, nil
+	} else {
+		cert, err := ics.getCertificateFromSecret("istio-ca-secret")
+		if err != nil {
+			return nil, err
+		}
+		return []CertInfo{cert}, nil
+	}
```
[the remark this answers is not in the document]
leandroberetta: It's true, in idiomatic Go, return faster.

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (ics *IstioCertsService) getCertificateFromSecret(secretName string) (CertInfo, error) {
+	cfg := config.Get()
+
+	secret, err := ics.k8s.GetSecret(cfg.IstioNamespace, secretName)
+
+	if err != nil {
+		log.Println(err)
```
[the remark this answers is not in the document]
leandroberetta: Yes, i'm debugging too! Sorry about forgetting to remove this.

business/istio_certs.go
leandroberetta: I will check, thanks!.

business/istio_certs.go
leandroberetta: True, will do that.

leandroberetta (commented):
@xeviknal thanks for your review!.

leandroberetta:
@lucasponce @xeviknal @jmazzitelli 

I restricted a little bit the out of the box functionality for granting less permissions to Kiali and operator service account.

I guess that in the majority of the cases, the secrets we need to inspect are named **istio-ca-secret** (default) and **cacerts** (user provided), so I'm giving these permissions to the service accounts (get, list, watch).

By doing this, we cover a great part of the use case, but also we are leaving out the case of secrets (certificates) generated by Chiron (it can be multiple secrets, with names defined by the user at Istio install). For this case, my proposal is to nicely indicate that there is an error regarding permissions, and if the users really need to inspect the content, give proper permissions on their own, after that, Kiali will be able to read them.

## Cases

### Istio's default certificates

certs_istio

### Plugin CA certificates

[https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/](https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/)

certs_plugin

### Chiron 

[https://istio.io/latest/docs/tasks/security/cert-management/dns-cert/](https://istio.io/latest/docs/tasks/security/cert-management/dns-cert/)

#### With permissions

certs_chiron_ok

#### With error permissions on a secret

certs_chiron_error


What do you think about this proposal?.

@xeviknal I will review again the code before asking a review for the new changes, and I also need to add some test cases to cover the errors.

jmazzitelli:
> What do you think about this proposal?.

I like restricting it as you have described. One thing you should do is re-word that error message:

image

I would change it so it indicates this is an expected situation, it isn't broken. I don't know how to word it :) But everything about that current message - from the red font color to the first words "An error occurred" - makes it seem like something unexpected and wrong happened and the user has a bad situation that they should immediately fix, when really this is an expected harmless condition and that the user has the option to add functionality here if they want.

Maybe something like:

```
For security purposes, Kiali has not been granted permission to view this certificate.
If you want Kiali to provide details about this certificate then you must grant the
Kiali service account permission to read the secret 'foo' found in namespace 'bar'.
Refer to the Kiali documentation for details on how you can add this permission.
```

Then in our kiali.io docs, you can have an FAQ on this and provide details. The details would include an explanation on why we don't provide this permission out-of-box, what the implications are if a user does grant these permissions, and instructions on how a user can add these permissions (i.e. give them the kubectl commands necessary to add those permissions).

business/istio_certs.go
jmazzitelli: > Anyway, the magic string should be replaced by the config variable already present in the configmap.

The config map name is found in our config here - you should use this rather than hardcode "istio" because it isn't always going to be that name: https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L543

business/istio_certs.go
```diff
@@ -0,0 +1,143 @@
...
+	DNSNames   []string `yaml:"dnsNames"`
+}
+
+type istioConfig struct {
+	Certificates []certConfig `yaml:"certificates"`
+}
+
+const (
+	IstioDefaultCASecret string = "istio-ca-secret"
+	UserProvidedCASecret string = "cacerts"
```
jmazzitelli, line 31: confirm these are the names in OSSM, too.

And if they can be different (OSSM? or canary deployments?) these should be configurable in our configmap somewhere (in [the same way the name of the configmap is configurable](https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L543)).

leandroberetta: @jmazzitelli thanks for your comments.

I confirm that secrets names in OSSM are the same.

Regarding canary deployments, not an expert but I did a test upgrading from 1.9.6 to 1.10.3 seeing that there is not a new secret created after the upgrade, it seems to be using the same "istio-ca-secret".

I will continue with this new strategy and will update the message as you proposed.

leandroberetta: With the new message:

certs_chiron_new_msg

lucasponce: Sorry, I'm late to the review, I see several colleagues are working together to review this one.
A first comment from my side, in this screen capture, I'd suggest a minor change in the design:

"From dns.example1-service-account secret"
(Warning icon) Access Denied + Tooltip with the original message

In that case, if this list grow it can handle it better, or if you have a list of several certificates without access, you can benefit of this design (note that other cards are designed in that way).

leandroberetta: @lucasponce something like the following?

certs_new_error

jmazzitelli:
I did a bunch of testing with different permutations of the permissions. Everything in the operator and helm charts work, but I have a question about what this `/api/istio/certs` is supposed to return. Look at the results I was getting and see if this is expected. I don't know - I'll leave it up to @leandroberetta to determine if those results are OK. See: https://github.com/kiali/kiali-operator/pull/378#issuecomment-902039512 and comments that follow it.

leandroberetta:
> I did a bunch of testing with different permutations of the permissions. Everything in the operator and helm charts work, but I have a question about what this `/api/istio/certs` is supposed to return. Look at the results I was getting and see if this is expected. I don't know - I'll leave it up to @leandroberetta to determine if those results are OK. See: [kiali/kiali-operator#378 (comment)](https://github.com/kiali/kiali-operator/pull/378#issuecomment-902039512) and comments that follow it.

Thanks for testing @jmazzitelli. 

For the case:

```
$ kubectl get kiali kiali -n kiali-operator -o yaml
...
  kiali_feature_flags:
    certificates_information_indicators:
      enabled: false
...
```
Results in:

```
$ curl -s http://localhost:20001/kiali/api/istio/certs | jq
[
  {
    "secretName": "cacerts",
    "secretNamespace": "istio-system",
    "dnsNames": null,
    "issuer": "",
    "notBefore": "0001-01-01T00:00:00Z",
    "notAfter": "0001-01-01T00:00:00Z",
    "error": "",
    "accessible": false
  }
]
```

Yes, this is a new case that is introduced by the new way the operator works and needs to be fixed in the server. I guess that when false, the server may return an empty list and in the UI not even show the section.

For the case:

with all defaults (operator and kiali cr) then all permissions are granted and the feature is enabled. This results:

```
$ curl -s http://localhost:20001/kiali/api/istio/certs | jq
[
  {
    "secretName": "istio-ca-secret",
    "secretNamespace": "istio-system",
    "dnsNames": null,
    "issuer": "O=cluster.local",
    "notBefore": "2021-08-18T08:04:32Z",
    "notAfter": "2031-08-16T08:04:32Z",
    "error": "",
    "accessible": true
  }
]
```

This is OK, in a standard out of the box installation, the only secret found is "istio-ca-secret", cacerts is not found, falling back to this one (cacerts may be created by the user).

For the case:

using --set "secretReader={cacerts}" and with this in kiali cr:

```
  kiali_feature_flags:
    certificates_information_indicators:
      enabled: true
      secrets:
      - cacerts
```

this results:

```
$ curl -s http://localhost:20001/kiali/api/istio/certs | jq
[
  {
    "secretName": "istio-ca-secret",
    "secretNamespace": "istio-system",
    "dnsNames": null,
    "issuer": "",
    "notBefore": "0001-01-01T00:00:00Z",
    "notAfter": "0001-01-01T00:00:00Z",
    "error": "",
    "accessible": false
  }
]
```

This is a case of bad configuration, you're saying that cacerts secret will be granted for reading but does not exists, so the logic is falling back again to istio-ca-secret, which is not accessible.

Note: The secrets list in the operator is just used to generated permissions, the server uses a different logic:

1) Look for Chiron certificates
2) Look for cacerts
3) Look for istio-ca-secret

IMO I need to fix the first case using the "enable" flag. Also, a comment about why in the first case is not falling back to istio-ca-secret, I treat the forbidden errors different than for example a not found error, if I found that is forbidden, I set the "accessible" field to false.

jmazzitelli:
> This is a case of bad configuration, you're saying that cacerts secret will be granted for reading but does not exists, so the logic is falling back again to istio-ca-secret, which is not accessible.

OK. Then this was just a case of me not understanding the use-case :)

Don't forget to write up a kiali.io PR with documentation for this stuff. We'll need to explain what this is all doing. I don't know if it goes in "Features" or "FAQ" in the docs, but we need a write up somewhere.

I would not merge any of these PRs until after the 1.39 sprint release is built and gone out this Friday, Aug 20, 2021. These permission changes need a couple weeks to bake among us devs before we release it out in the wild. I did a good amount of testing, so I think it should be good, but I don't trust myself :)

leandroberetta:
> Don't forget to write up a kiali.io PR with documentation for this stuff. We'll need to explain what this is all doing. I don't know if it goes in "Features" or "FAQ" in the docs, but we need a write up somewhere.

Yes, needs to be clear for sure, thanks for reminding me this.

> I would not merge any of these PRs until after the 1.39 sprint release is built and gone out this Friday, Aug 20, 2021. These permission changes need a couple weeks to bake among us devs before we release it out in the wild. I did a good amount of testing, so I think it should be good, but I don't trust myself :)

Sure, makes sense. Thanks a lot for your help and ideas about this feature and the security concerns.

business/istio_certs.go
lucasponce: @leandroberetta yes +1

business/istio_certs.go
```diff
@@ -0,0 +1,150 @@
...
+
+	cert.Parse(secret.Data[certName])
+
+	return cert, nil
+}
+
+func (ics *IstioCertsService) getCertsConfigFromIstioConfigMap() ([]certConfig, error) {
+	cfg := config.Get()
+
+	istioConfigMap, err := ics.k8s.GetConfigMap(cfg.IstioNamespace, "istio")
```
jmazzitelli, line 98: Don't forget to update this - this configmap name is configurable and will be different for OSSM and also (IIRC) for canary deployments, too. https://github.com/kiali/kiali-operator/blob/v1.38.0/deploy/kiali/kiali_cr.yaml#L540

jmazzitelli (changes requested):
need to use configured configmap name

business/istio_certs.go
leandroberetta: Ok, thanks for letting me know. It's true. Fixed.

jmazzitelli (approved):

lucasponce:
@leandroberetta as this effort is getting mature, I've added @skondkar to the PRs to give visibility.
Please, discuss with them in the Scrum meeting when this is ready to test and probably define a QE flow that may help them in what to test (the demo you did was pretty good in that sense, it's already uploaded).

@skondkar, please also coordinate with @ScriptingShrimp when he is also ready to jump into QE PR testing, thanks !

business/istio_certs.go
```diff
@@ -0,0 +1,155 @@
...
+			return nil, err
+		}
+	}
+
+	certs := make([]models.CertInfo, 0)
+	for cert := range certChan {
+		certs = append(certs, cert)
+	}
+
+	sort.Slice(certs, func(i, j int) bool {
```
xeviknal, line 150: +1!

xeviknal (dismissed):
LGTM! Great great job! Long work but great feature :muscle:

leandroberetta:
@skondkar, some comments for testing this PR:

This feature is enabled by default (feature flag) and with permissions set for reading 2 secrets (istio-ca-secret and cacerts):

Kiali CR:

```yaml
spec:
  kiali_feature_flags:
    certificates_information_indicators:
      enabled: true
      secrets:
      - cacerts
      - istio-ca-secret
```

There are 3 possible configurations to test that mainly depends on how Istio was installed and configured:

### Default

Istio will use its own CA, using the secret istio-ca-secret. 

The information expected to see in Kiali comes from that secret.

### Plug in CA

The user will configured their own CA, to test this I follow the following example at Istio:

https://istio.io/latest/docs/tasks/security/cert-management/plugin-ca-cert/#plug-in-certificates-and-key-into-the-cluster

After creating the cacerts secret, Kiali will show information about this secret instead of the default one.

### DNS Certificates

The user will configure some secret names and dns entries for these secrets prior to install Istio resulting in some secrets created with certificates information.

To test this I follow the next example at Istio:

https://istio.io/latest/docs/tasks/security/cert-management/dns-cert/

Kiali CR needs to be modified with the following config:

```yaml
spec:
  kiali_feature_flags:
    certificates_information_indicators:
      enabled: true
      secrets:
      - dns.example1-service-account
      - dns.example2-service-account
```

Kiali will show information about these certs but there will be some permissions issues:

1. Check that the UI shows permissions denied warnings
2. Fix permissions and check that the information is shown properly

For fixing permissions, follow the next steps:

```yaml
# Update Kiali operator cluster role to read the new secrets
kubectl patch $(kubectl get clusterroles -o name | grep kiali-operator) --type "json" -p '[{"op":"add","path":"/rules/0","value":{"apiGroups":[""],"resources":["secrets"],"verbs":["get"],"resourceNames":["dns.example1-service-account"]}}]'
kubectl patch $(kubectl get clusterroles -o name | grep kiali-operator) --type "json" -p '[{"op":"add","path":"/rules/0","value":{"apiGroups":[""],"resources":["secrets"],"verbs":["get"],"resourceNames":["dns.example2-service-account"]}}]'
```

Finally, if the feature is disabled, the UI won't show the "View certificates info" item in Help menu.

skondkar (commented):
Update:
Verified that the feature flag is enabled by default
Verified that the UI displays istio default secret istio-ca-secret and also verified for Plug in CA.

Screenshot from 2021-08-31 16-45-36

Screenshot from 2021-08-31 16-47-53

Screenshot from 2021-08-31 17-13-09

working with Leandro for testing on DNS certificate.

jmazzitelli (approved):
I assume this was just a rebase and nothing changed. Re-approving (but I did not run tests again).

leandroberetta:
> I assume this was just a rebase and nothing changed. Re-approving (but I did not run tests again).

Correct.

leandroberetta:
The configuration related with DNS certificates doesn't seem to work in OpenShift due to:

```
2021-09-07T17:34:48.136003Z	error	failed to verify the certificate chain: x509: certificate signed by unknown authority
```

This happens when Istio generates CSRs for the certificates.  What I propose for now is to merge this code and investigate why this is happening if there is interest from the community. 

Note: This configuration is not possible in Maistra (it's not implemented in Maistra operator).

lucasponce (approved):

skondkar:
Polarion Test case -> SWSQE-1982

PR #4229, opened 2021-07-28
Reorganization of the test helpers into testutils

xeviknal:
After the addition of the `testutils` folders for the test helpers, I felt that some validations helpers should be living under the same folder now.

Therefore, the situation that this PR aims is:
- `tests/data`: contains all the necessary yamls and the old data builders needed for the tests (istio data object creation)
- `tests/testutils`: contains actual helpers needed for all kinds of purposes.
- `tests/testutils/validations`: contains actual helpers needed for validations


jmazzitelli (approved):
changes look good, all unit tests pass on my machine.

nrfox (approved):
LGTM

Issue #4221, opened 2021-07-26
Issues with clustering discovery


israel-hdez:
There are a few issues with _clustering discovery_ when Istio is installed with a revision (usually for canary upgrades).

The root cause are missing configurations in Kiali, because when using revisions (or custom names in the resources) the customized/versioned resource names need to be specified in Kiali configuration.

In any case, Kiali should have better defaults or fail gracefully:

* Although local cluster discovery fails gracefully, the local cluster is assigned the "unknown" name. This results with a graph with `unknown` cluster label for idle nodes, and `Kubernetes` cluster labels for nodes found through telemetry. This is caused because the [Istio deployment name is not specified in the KialiCR](https://github.com/kiali/kiali-operator/blob/15e1aaea57b887b10386c3c04fd424d431809a87/deploy/kiali/kiali_cr.yaml#L574).
  * Idle nodes should fallback to using `Kubernetes` for it's cluster name.
  * Relevant line to fix: https://github.com/kiali/kiali/blob/f754e63bd224d8e5e13eaec1d6cae230955b82a7/business/mesh.go#L450
* If the Istio deployment name is specified correctly, BUT the [sidecar injector config map name is unset](https://github.com/kiali/kiali-operator/blob/15e1aaea57b887b10386c3c04fd424d431809a87/deploy/kiali/kiali_cr.yaml#L573) there is a crash when Kiali tries to discover the network of the home cluster. Kiali should fail gracefully.
  * Possible location of the crash: https://github.com/kiali/kiali/blob/f754e63bd224d8e5e13eaec1d6cae230955b82a7/business/mesh.go#L234-L237

jshaughn:
@israel-hdez That looks good.  In addition look for the three occurrences of `globalInfo.HomeCluster = "unknown"` in the grap appenders.   We don't want the default cluster name to every be "unknown", if anything it should be "Kubernetes" to match Istio.

israel-hdez:
@jshaughn OK, I'll do that search&replace.

PR #4231, opened 2021-07-28
Fix graph with spurious cluster boxes and a crash

israel-hdez:
* Provide better defaults when ClusterID cannot be found.
* Avoid a crash if the configured sidecar injector configmap does not exist.

Fixes kiali/kiali#4221

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -20,6 +20,8 @@ import (
 	"github.com/kiali/kiali/util/httputil"
 )
 
+const DefaultClusterID = "Kubernetes"
+
 // MeshService is a support service for retrieving data about the mesh environment
 // when Istio is installed with multi-cluster enabled. Prefer initializing this
 // type via the NewMeshService function.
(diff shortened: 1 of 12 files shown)


israel-hdez:
Backported to v1.36 in #4232

jshaughn (changes requested):
Code changes look good. Good to see dead_node_tests updated, but I think we should update all the appender test code to properly use the new DefaultClusterID.  I have sent you a sub-PR for this.  If you merge it I think we can approve this.

jshaughn (approved):
LGTM!

israel-hdez:
@mattmahoneyrh This is the fix for the duplicated bookinfo namespaces you found.
This isn't the root cause, but it avoids the need of the manual workaround.

Please, review if you think it's needed.

business/mesh.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/kiali/kiali/util/httputil"
 )
 
+const DefaultClusterID = "Kubernetes"
```
jmazzitelli, line 23: See my comment in the backport PR (i.e. should this be all lowercase?)

business/mesh.go
```diff
@@ -372,6 +374,11 @@ func (in *MeshService) resolveKialiNetwork() (string, error) {
 		// This setup is OK, it's just that it's not within our multi-cluster assumptions.
 		log.Warningf("Cannot resolve the network ID of the cluster where Kiali is hosted: cannot get the sidecar injector config map :%v", err)
 		return "", nil
+	} else if istioSidecarConfig == nil {
+		// Don't return an error, as this may mean that Kiali is not installed along the control plane.
```
jmazzitelli, line 378: see my comment in the backport PR
```suggestion
		// Don't return an error, as this may mean that Kiali is not installed in the control plane namespace.
```

business/mesh.go
jshaughn: It is correct as is.  It mirrors the default set by Istio.

mattmahoneyrh (approved):

business/mesh.go
israel-hdez: Let's discuss in the backport PR. I'll change the comment in another work, if needed.

PR #4232, opened 2021-07-28
[v1.36] Fix graph with spurious cluster boxes and a crash

israel-hdez:
Cherry-pick of #4231 to `v1.36` branch

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -20,6 +20,8 @@ import (
 	"github.com/kiali/kiali/util/httputil"
 )
 
+const DefaultClusterID = "Kubernetes"
+
 // MeshService is a support service for retrieving data about the mesh environment
 // when Istio is installed with multi-cluster enabled. Prefer initializing this
 // type via the NewMeshService function.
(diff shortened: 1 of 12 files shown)


business/mesh.go
```diff
@@ -20,6 +20,8 @@ import (
 	"github.com/kiali/kiali/util/httputil"
 )
 
+const DefaultClusterID = "Kubernetes"
```
jmazzitelli, line 23: are cluster IDs typically mixed case? I would think they would be all lowercase. If so, I would change this.

Not sure if it matters but... it would be consistent with "real" cluster IDs anyway.

```suggestion
const DefaultClusterID = "kubernetes"
```

business/mesh.go
```diff
@@ -372,6 +374,11 @@ func (in *MeshService) resolveKialiNetwork() (string, error) {
 		// This setup is OK, it's just that it's not within our multi-cluster assumptions.
 		log.Warningf("Cannot resolve the network ID of the cluster where Kiali is hosted: cannot get the sidecar injector config map :%v", err)
 		return "", nil
+	} else if istioSidecarConfig == nil {
+		// Don't return an error, as this may mean that Kiali is not installed along the control plane.
```
jmazzitelli, line 378: do you mean not installed along "with" the control plane? (i.e. it is not installed in the control plane namespace?)
```suggestion
		// Don't return an error, as this may mean that Kiali is not installed in the control plane namespace.
```

business/mesh.go
jshaughn: No, do not make this change.  "Kubernetes" is correct, we are mirroring the value set by Istio as the default.

jshaughn (changes requested):
@israel-hdez This backport needs to be updated with the additional test file changes in the upstream PR.

israel-hdez:
> @israel-hdez This backport needs to be updated with the additional test file changes in the upstream PR.

Tests are updated.
Some files in master have diverged. I hope I did correctly the cherry-pick.

business/mesh.go
israel-hdez: Nope. If I recall correctly, this only to prevent a failure in case Istio is installed as [Primary-Remote](https://istio.io/latest/docs/setup/install/multicluster/primary-remote_multi-network/) (i.e. Istio installed in only one cluster, and controlling multiple clusters), but then somebody chose to install Kiali in one of those cluster where the Istio control plane is absent.

israel-hdez:
@jmazzitelli I wonder if you could be Jay's backup for reviewing this backport?

jmazzitelli:
Oh, did this not go in? I think this might be the cause of https://github.com/kiali/kiali/issues/4246

jmazzitelli (approved):
quick test shows this ok

PR #4233, opened 2021-07-29
[v1.36] Run molecule tests with KinD (#4189)

jmazzitelli:
This is a cherry pick of: https://github.com/kiali/kiali/pull/4189 because we cherry picked [its operator peer](https://github.com/kiali/kiali-operator/pull/376) and so we might as well get this in too (allows kind to run the tests in the 1.36 branch now)

* run molecule tests with kind

* Use volumes rather than mounts, and use full fork names

* Prefer using volumes rathern than mounting directories on the host
machine into the container. Although this was mainly to add
compatibility for using a remote docker engine, in general, this
potentially works better. The rationale is that there is no need to
adjust privileges nor SELinux on the host for the container to work
properly, because the container engine is fully managing the volume.
* Use forks's fullname rather than just the username/orgname that has
the fork. This helps with older forks that have the old swscore name.

* Use Libera IRC instead of Freenode. Also add KinD config

Since Kiali has move to Lbiera Chat for IRC, use it instead of Freenode.
Also add a configuration for KinD to start with a control plane node and
a worker node.

Co-authored-by: John Mazzitelli <mazz@redhat.com>

### DIFF
--- a/hack/ci-kind-molecule-tests.sh
+++ b/hack/ci-kind-molecule-tests.sh
@@ -1,0 +1,447 @@
+#!/bin/bash
+
+#
+# Refer to the --help output for a description of this script and its available options.
+#
+
+infomsg() {
+  echo "[INFO] ${1}"
+}
+
+helpmsg() {
+  cat <<HELP
+This script will run the Kiali molecule tests within a KinD cluster.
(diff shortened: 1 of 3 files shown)


jshaughn (approved):
LOK

Issue #4215, opened 2021-07-26
Scripts not loading (404) on openid_error when Kiali is hosted in a subfolder (web_root: /kiali)


stevensystems:
**Describe the bug**
When using a sub-directory to host kiali (e.g. web_root: /kiali), some statics files are still referenced using the root "/" path, and cannot be loaded (404).

![image](https://user-images.githubusercontent.com/76811102/126973881-7a129be3-6913-44c2-ba0f-e42d9e248322.png)

**Versions used**
Kiali: v1.37.0
Istio: 1.10.3
Kubernetes flavour and version: (AKS V. 1.19.11)

**To Reproduce**
1. Install Kiali operator using Helm chart:
```sh
kubectl create namespace kiali-operator
helm install \
    --set cr.create=false \
    --set cr.namespace=istio-system \
    --namespace kiali-operator \
    --repo https://kiali.org/helm-charts \
    kiali-operator \
    kiali-operator
```
2. Setup Kiali with an openid provider and web_root "/kiali"
```yaml
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:
  name: kiali
  annotations:
    ansible.sdk.operatorframework.io/verbosity: "1"
spec:
  version: "default" 
  istio_namespace: "istio-system"
  api:
    namespaces:     
      exclude:
      - "istio-operator"
      - "kube-.*"
      - "kiali-operator"
      - "cert-manager"
  auth:
    strategy: "openid"
    openid:
      client_id: "<client id>"
      issuer_uri: "https://sts.windows.net/<tenant id>/"
      username_claim: preferred_username
      api_token: "access_token"     
      additional_request_params:
        resource: "6dae42f8-4368-4678-94ff-3960e28e3630"
  secret_name: "kiali"
  deployment:
    ingress_enabled: false
    image_pull_policy: "IfNotPresent"
    namespace: "monitoring-system"
    replicas: 1
    logger:
      log_level: info
      log_format: text
      sampler_rate: "1"

    view_only_mode: true
    external_services:
      custom_dashboards:
        enabled: true
    server:
      metrics_enabled: true
      metrics_port: 9090
      port: 20001
      web_root: /kiali
      web_fqdn: mykialihost.****azure.com
      web_port: 443
      web_schema: "https"
```
3. Create a new user on your openid provider, that has no access to kiali.
4. Open the developer toolbar of your browser and start tracing network traffic
4. Login to "Kiali" with the newly created user from step 3.
[...]

PR #4235, opened 2021-07-29
Fix root URL not loading with URL parameters

israel-hdez:
Fix non '/' root url failing to load if there are URL parameters (i.e. there is a ? symbol in the URL)

Fixes kiali/kiali#4215

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -53,9 +53,14 @@ func NewRouter() *mux.Router {
 	}
 
 	fileServerHandler := func(w http.ResponseWriter, r *http.Request) {
-		if r.RequestURI == webRootWithSlash || r.RequestURI == webRoot || r.RequestURI == webRootWithSlash+"index.html" {
+		urlPath := r.RequestURI
+		if r.URL != nil {
+			urlPath = r.URL.Path
+		}
+
+		if urlPath == webRootWithSlash || urlPath == webRoot || urlPath == webRootWithSlash+"index.html" {
 			serveIndexFile(w)
-		} else if r.RequestURI == webRootWithSlash+"env.js" {
+		} else if urlPath == webRootWithSlash+"env.js" {
 			serveEnvJsFile(w)
 		} else {
 			staticFileServer.ServeHTTP(w, r)


jmazzitelli:
Why did this unit test not find the bug in the first place? We should fix this if possible: https://github.com/kiali/kiali/blob/v1.37.0/server/server_test.go#L32

israel-hdez:
> Why did this unit test not find the bug in the first place?

I guess it's because the request does not fails. Instead, it responds with the `index.html` template.
That test would need to check that the returned HTML has the expected \<base> tag.

jmazzitelli (approved):
This works. Approving.

I don't know why that unit test didn't pick this up (UPDATE: ahh, Just saw your last comment). If you can figure it out today, good. If not, merge anyway because we need this in the release going out tomorrow.

israel-hdez:
@mattmahoneyrh I'm so-so sure that the issue being fixed this affects v1.36. Perhaps, can somebody from QE check? To see if we need to backport.

jmazzitelli:
I should be able to check this - git checkout the v1.36 branch, build it, and use `run-kiali.sh` .... i'll try it now

jmazzitelli:
```
2021-07-29T14:08:57-04:00 INF Kiali: Version: v1.36.2, Commit: f9040e2889b0482759540c95b211c63d65ea3e99
...
2021-07-29T14:08:57-04:00 INF Server endpoint will start at [:20001/kiali]
```

image

pbajjuri20 (approved):
LGTM!

jmazzitelli:
ugh.. this didn't make it into 1.38????

jmazzitelli:
Is there a backport for this PR? We need one: https://github.com/kiali/kiali/pull/4235#issuecomment-889355443

israel-hdez:
> ugh.. this didn't make it into 1.38????

Nope :-( I though it would make it. But I think I merged when the build was already running.


> Is there a backport for this PR? We need one: [#4235 (comment)](https://github.com/kiali/kiali/pull/4235#issuecomment-889355443)

There you go: https://github.com/kiali/kiali/pull/4245

Issue #4238, opened 2021-07-30
Jaeger traces & spans fetching error


pbajjuri20:
Jaeger traces & spans are not fetching in the Kiali for Istio 1.11.0-beta.3

Reproduce steps:

1. install upstream istio 1.11.0-beta.3 with upstream kiali
2. install bookinfo 
3. Navigate to Kiali
4. Click on "Applications" (left-side panel)
5. Click on "Details"
6. Click on "Traces" (Jaeger traces are not fetching)

![Jaeger_error](https://user-images.githubusercontent.com/55099523/127583815-60283835-1bf9-4ac5-bebb-52aace9c51d6.png)

![Jaeger_3](https://user-images.githubusercontent.com/55099523/127583993-c7f4fc8b-3173-4f46-af8d-4c723117551e.png)

xeviknal:
How did you install kiali?

I think this is a duplicate of this one: https://github.com/kiali/kiali/issues/4207
However, there is one thing missing in our side. Kiali installed via Istio addons won't work because the default value is still using the port 80, old one. So we need to change it.

xeviknal:
It shouldn't be broken if you installed kiali the operator in the latest version in master.
What you can also do is to go to the kiali-cr and change the `external_services.tracing.in_cluster_url` to `http://tracing.istio-system:16685/jaeger` and everything should be back to normal.

pbajjuri20:
@xeviknal: it works successfully with this port  'http://tracing.istio-system:16685/jaeger'.

xeviknal:
Purrfect. It would be wonderful to check kiali 1.38 (or latest master) with latest istio 1.11. Do you mind double checking it? Thank you :)

pbajjuri20:
@xeviknal: Yes, I have validated against Kiali 1.38 with the latest Istio 1.11.

xeviknal:
I am re-opening the issue.

I see two problems:

1) Default `use_grpc` is false but the default `in_cluster_url` uses the `gRPC` one. My fault:
Should we rephrase this sentence? https://github.com/kiali/kiali-operator/blob/98a2fb7441f600548172bf8cdd83230dd812c7ad/deploy/kiali/kiali_cr.yaml#L642
I read that gRPC defaults to `true`. However, our [config defaults](https://github.com/kiali/kiali/blob/master/config/config.go#L520) to `false`.
The problem this creates is the following: masthead and traces are not present because kiali sends HTTP traffic to gRPC port.
[...]

PR #4240, opened 2021-07-30
Jaeger v1.20+ uses new port for grpc. Default url points to it

xeviknal:
fix https://github.com/kiali/kiali/issues/4238
related https://github.com/kiali/kiali-operator/pull/370

The default in_cluster_url should point to the new grpc port of jaeger.
Notice that kiali default installation will fail for older jaeger releases than 1.20. However, users will only need to change the configmap/kiali-cr fixing the port.

@jmazzitelli do we need to update any other thing to have all the installations covered?
I guess that  https://github.com/kiali/kiali-operator/pull/370 covered the installation through the operator, and with this PR we are covering helm and addons, right?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -514,7 +514,7 @@ func NewConfig() (c *Config) {
 				},
 				Enabled:              true,
 				NamespaceSelector:    true,
-				InClusterURL:         "http://tracing.istio-system/jaeger",
+				InClusterURL:         "http://tracing.istio-system:16685/jaeger",
 				IsCore:               false,
 				URL:                  "",
 				UseGRPC:              false,


jmazzitelli:
> do we need to update any other thing to have all the installations covered?
> I guess that [kiali/kiali-operator#370](https://github.com/kiali/kiali-operator/pull/370) covered the installation through the operator, and with this PR we are covering helm and addons, right?

The upstream Istio addon yaml is built via the server helm chart. And the server helm chart does not define any of the external_services.tracing settings - so the server's defaults hardcoded in config.go (i.e. what this PR is doing) will take effect.

So as long as this new `http://tracing.istio-system:16685/jaeger` setting will work with the default Jaeger installation, this is all you need.

jmazzitelli (approved):

PR #4249, opened 2021-08-03
Using gRPC or HTTP methods to check health of the jaeger component

xeviknal:
fix https://github.com/kiali/kiali/issues/4238
fix https://github.com/kiali/kiali/issues/4238#issuecomment-891624634

requires https://github.com/kiali/kiali-operator/pull/384

Jaeger has splitted its ports in two: one for HTTP another one for gRPC.
The istio component health feature was checking jaeger accessibility through HTTP calls.
This PR changes this method for checking accessibility to support both HTTP and gRPC. When Jaeger is consumed using gRPC, the accessibility will be checked using gRPC as well. Similarly, when Jaeger is consumed using HTTP, kiali will use HTTP to discover its accessibility.

This PR also makes gRPC the default strategy to consume jaeger.

This should be tested:
- enabling / disabling grcp (http and grcp traffic)
- installing kiali via operator and via istio addons (modifying the image in the addon)


business/istio_status.go
```diff
@@ -237,7 +237,7 @@ func (iss *IstioStatusService) getAddonComponentStatus() IstioComponentStatus {
 
 	go getAddonStatus("prometheus", true, extServices.Prometheus.IsCore, &extServices.Prometheus.Auth, extServices.Prometheus.URL, extServices.Prometheus.HealthCheckUrl, staChan, &wg)
 	go getAddonStatus("grafana", extServices.Grafana.Enabled, extServices.Grafana.IsCore, &extServices.Grafana.Auth, extServices.Grafana.InClusterURL, extServices.Grafana.HealthCheckUrl, staChan, &wg)
-	go getAddonStatus("jaeger", extServices.Tracing.Enabled, extServices.Tracing.IsCore, &extServices.Tracing.Auth, extServices.Tracing.InClusterURL, extServices.Tracing.HealthCheckUrl, staChan, &wg)
```
xeviknal, line 240: I guess that we should remove the `healthCheckUrl` because the URL used now will always be present. The URL now is static and is part of the Jaeger API. So I think it doesn't have any sense anymore now.

This config was set because of this request: https://github.com/kiali/kiali/issues/3631

leandroberetta (dismissed):
LGTM. Tested on my local environment with jaeger-test molecule test.

jmazzitelli:
I'm not able to get the jaeger-test to pass over here. I'm not sure what is wrong. I'm still trying...

jmazzitelli (dismissed):
I see the jaeger-test pass now.

xeviknal:
@jmazzitelli  and @leandroberetta sorry for the noise, last commit just removed an unnecessary and unused config. I'll create a PR for the operator

jmazzitelli:
Question: should this be backported? I am thinking "no, not right now" because 1.36 is targeting istio 1.10 and under.

@xeviknal is it true the defaults in this PR will now work for istio 1.11 but NOT for 1.10 and under? If so, then no backport I guess.

xeviknal:
No, it should be backported. AFAIK 1.36 target won't suffer from port splitting, so no problems in that area.
It should be backported to 1.38 though. Otherwise istio 1.11 won't work by default

xeviknal:
@pbajjuri20 do you mind testing this branch for istio 1.11? Using latest master kiali addon pointing to this build should work without changing any configmap setting.

pbajjuri20:
@xeviknal: I will test this PR against Istio 1.11

jmazzitelli:
@pbajjuri20 when you confirm all is OK, I'll merge this and then prepare the backports.

jmazzitelli:
once merged, we will backport to `v1.38` branch

pbajjuri20:
LGTM! 
@jmazzitelli : thanks for the instructions. 

error-22

pbajjuri20 (approved):

PR #4254, opened 2021-08-03
[hack] support istio 1.11 and the openshift profile

jmazzitelli:
The openshift profile existed on istio 1.9 and 1.10 so we should be able to use this for those as well. But if something is wrong and it doesn't work with 1.9 and/or 1.10 on your machine, just pass in `-cp demo` to get back to the original way the script was working.


israel-hdez (approved):
LG

Issue #4218, opened 2021-07-26
[validations] Don't show KIA0203 when there are no VS referencing the DR subset


xeviknal:
> KIA0203 - This subset's labels are not found in any matching host. Our CD pipeline controls VirtualServices and DestinationRules, but the DRs often have references to pods that don't exist (e.g. a canary pod) but when this happens the VS would not reference the DR subset anyway, so it's not a problem.

from https://github.com/kiali/kiali/issues/4197

Plus,
> What if we remove the validation in the DR when there are no VS referencing the DR subset? When a VS used the host specified in the DR, then we should trigger the validation and warn you.
Would that change help you? I think it would keep giving value to you, right?

from https://github.com/kiali/kiali/issues/4197#issuecomment-885075152
cc @linjmeyer

hhovsepy:
But this feature can cause a confusion among those users who create DR first, make sure that it is valid and then create VS over this DR. And they will see new errors which did not happen after DR creation.
So for avoiding any confusion over existing users, I think this new feature can be configurable in Kiali CR, and disabled by default.
And for users who need to ignore those errors in DR when VS is not created yet can activate manually this feature.

linjmeyer:
Do DRs serve any purpose if not referenced by another Istio resource like a VirtualService?  That almost seems to me like it should be an error in itself "This DR is not referenced by any other resources and serves no purpose".  I've never personally had a use case for creating a DR on it's own so I'm curious what the reason would be for that?

hhovsepy:
Having a VirtualService created is not a mandatory, as it is enough to have DestinationRule with appropriate subsets, and Istio (OpenShift) will have a default VirtualService with equally distributed traffic between subsets. 
VS is necessary when traffic distribution needs to be under control and specified by % per subset.

xeviknal:
@linjmeyer you are right. DRs itselves aren't of use for istio. They are useful when they are referenced by a VirtualService.

> Note: Policies specified for subsets will not take effect until a route rule explicitly sends traffic to this subset.

from https://istio.io/latest/docs/reference/config/networking/destination-rule/

What do you think if we play with the severity of the validation? Idea:

1. When the DR is not used, keep the validation but in informative severity. Because it won't affect any behavior in the cluster. Informative messages are only seen if you are browsing the object itself and won't create noise.
2. When the DR is referenced, increase the severity to error (as it is now). Error is the severity because this definitely create wrong traffic routing.

[...]

PR #4256, opened 2021-08-04
Make DRs subset validation error to info when no VS created for this subsets. Tests added.

hhovsepy:
POC for RFE: https://github.com/kiali/kiali/issues/4218
Documentation changes: https://github.com/kiali/kiali.io/pull/402

Make DestinationRules subset validations "Error" to "Info" when no VirtualService is referencing to that DestinationRule.

Before:
Error was displayed always when DR subset workload was not found.
Screenshot from 2021-08-09 14-48-26

After:
When DR subset workload is not found but that subset is not linked in any VS, Info severity is shown.
Screenshot from 2021-08-18 19-58-29

### DIFF
--- a/business/checkers/destinationrules/no_dest_checker.go
+++ b/business/checkers/destinationrules/no_dest_checker.go
@@ -1,6 +1,7 @@
 package destinationrules
 
 import (
+	"reflect"
 	"strconv"
 	"strings"
 
@@ -16,6 +17,7 @@ type NoDestinationChecker struct {
 	Namespaces      models.Namespaces
 	WorkloadList    models.WorkloadList
 	DestinationRule kubernetes.IstioObject
+	VirtualServices []kubernetes.IstioObject
 	ServiceEntries  map[string][]string
 	Services        []core_v1.Service
 	RegistryStatus  []*kubernetes.RegistryStatus
@@ -50,8 +52,12 @@ func (n NoDestinationChecker) Check() ([]*models.IstioCheck, bool) {
 									if !n.hasMatchingWorkload(fqdn.Service, stringLabels) {
 										validation := models.Build("destinationrules.nodest.subsetlabels",
 											"spec/subsets["+strconv.Itoa(i)+"]")
+										if n.isSubsetReferenced(dHost, innerSubset["name"].(string)) {
+											valid = false
+										} else {
+											validation.Severity = models.Unknown
+										}
 										validations = append(validations, &validation)
-										valid = false
 									}
 								}
 							} else {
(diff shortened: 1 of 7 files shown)


business/checkers/destinationrules/no_dest_checker.go
```diff
@@ -1,6 +1,8 @@
 package destinationrules
 
 import (
+	"github.com/kiali/kiali/config"
```
jmazzitelli, line 4: as per coding standard, this should be moved to the kiali import section (the third section in our imports)

config/config.go
```diff
@@ -357,6 +357,7 @@ type Validations struct {
 type KialiFeatureFlags struct {
 	IstioInjectionAction bool        `yaml:"istio_injection_action,omitempty" json:"istioInjectionAction"`
 	IstioUpgradeAction   bool        `yaml:"istio_upgrade_action,omitempty" json:"istioUpgradeAction"`
+	SkipSubsetValidation bool        `yaml:"skip_subset_validation,omitempty" json:"skipSubsetValidation"`
```
jmazzitelli, line 360: this should be moved under the "Validations" struct since it is a feature flag specifically involving validations.

config/config.go
```diff
@@ -530,6 +531,7 @@ func NewConfig() (c *Config) {
 		KialiFeatureFlags: KialiFeatureFlags{
 			IstioInjectionAction: true,
 			IstioUpgradeAction:   false,
+			SkipSubsetValidation: false,
```
jmazzitelli, line 534: this should be moved under the "Validations" struct since it is a feature flag specifically involving validations.

jmazzitelli (changes requested):

jmazzitelli:
> it will ignore DestinationRules subset validations when no VirtualService is referencing to that DestinationRule.

I think this should be informational (or at most warning) - I don't think this warrants an "Error" - it really isn't an error. Things still work. But we should at least tell the user they have unused DRs (hence why it should still be flagged, but just not as an error)

xeviknal:
@jmazzitelli I agree with the information point. There is more discussion in the issue with its author: https://github.com/kiali/kiali/issues/4218

business/checkers/destinationrules/no_dest_checker.go
```diff
@@ -147,3 +155,74 @@ func (n NoDestinationChecker) hasMatchingService(host kubernetes.Host, itemNames
...
+
+					host, ok := destination["host"].(string)
+					if !ok {
+						continue
+					}
+
+					subset, ok := destination["subset"].(string)
+					if !ok {
+						continue
+					}
```
xeviknal, line 213: Do you think we can extract this method into a common area for different checkers? I see there is a similar code in here: https://github.com/kiali/kiali/blob/3977524605b3bd216d58b8d3093a27f0ae32d352/business/checkers/virtualservices/subset_presence_checker.go#L22-L66

Have you tried it?

business/checkers/destinationrules/no_dest_checker.go
```diff
@@ -48,10 +50,16 @@ func (n NoDestinationChecker) Check() ([]*models.IstioCheck, bool) {
...
+										if n.isSubsetReferenced(dHost, innerSubset["name"].(string)) {
+											validation := models.Build("destinationrules.nodest.subsetlabels",
+												"spec/subsets["+strconv.Itoa(i)+"]")
+											validations = append(validations, &validation)
+											valid = false
+										} else {
+											validation := models.Build("destinationrules.nodest.subsetlabelsnoterferenced",
+												"spec/subsets["+strconv.Itoa(i)+"]")
+											validations = append(validations, &validation)
+										}
```
xeviknal, line 62: What happens when the subset is not referenced but the subset has some matching workloads?
I guess that with this implementation the validation won't come up, right?

xeviknal: This can be reflected in a test case as well. wdyt?

xeviknal (changes requested):
I'd like to hear some thoughts I have about the implementation.
Plus, I think we should add this into kiali.io validations page as well.

business/checkers/destinationrules/no_dest_checker.go
hhovsepy: +1 to cover by test the scenario when subset is not referenced but have matching workload, for regression purposes

business/checkers/destinationrules/no_dest_checker.go
hhovsepy: yes the loop is the same, but the difference is in inner logic, can this logic be provided via local function?

hhovsepy: can you please send any example of making common functions and where we put it usually?

hhovsepy:
@xeviknal thank you for the comments, approached almost all, besides the common function one

business/checkers/destinationrules/no_dest_checker.go
```diff
@@ -48,10 +50,16 @@ func (n NoDestinationChecker) Check() ([]*models.IstioCheck, bool) {
...
-											"spec/subsets["+strconv.Itoa(i)+"]")
-										validations = append(validations, &validation)
-										valid = false
+										if n.isSubsetReferenced(dHost, innerSubset["name"].(string)) {
+											validation := models.Build("destinationrules.nodest.subsetlabels",
+												"spec/subsets["+strconv.Itoa(i)+"]")
+											validations = append(validations, &validation)
+											valid = false
+										} else {
+											validation := models.Build("destinationrules.nodest.subsetlabelsnoterferenced",
```
xeviknal, line 59: I see that you are using two different validations for this. If you want to do so, then in the documentation you should add one more entry covering this validation (KIA0210). Otherwise people will put this on google and won't found any documentation.

However, if you want to save this step, you can just change the severity for the existing validation
```golang
validation.Severity = models.Unknown
```

This should just keep the same validation but different severity. I am fine with both versions. Pick the one it works the best to your sense. Both have pros and cons we can evolve along the way.

business/checkers/destinationrules/no_dest_checker.go
xeviknal: No, I don't know exactly a way to do that. I was thinking if you wanted to get creative with go and find out something.
Let's wait until we find again the same need to reuse this code and then we refactor. So, let's close this comment.

Just for fun, I had a vague idea of having a helper which receives a function as a parameter. This function would have the specific responsibility for each scenario (return VSs or validations or any other interface{}). 
But I guess it is not the time to do so.

business/checkers/destinationrules/no_dest_checker.go
hhovsepy: thank you @xeviknal I rethink and decided to use the same error code but different severity, this will cause less complications and confusions

hhovsepy:
@xeviknal it seems all changes approached now

xeviknal (approved):
LGTM!

hhovsepy:
For QE:
Create Traffic Shifting for a "reviews" service.
In created DR "reviews" change one of the labels to "v3-wrong" for instance.
It will show error on that label.
Delete the VirtualService 'reviews'.
In DR 'reviews' the Error severity should become Info with the same message.

prachiyadav:
Test Case >> SWSQE-1882

prachiyadav (approved):
Screenshot from 2021-08-20 21-27-41


Screenshot from 2021-08-20 21-25-25

skondkar (approved):
Verified. LGTM

Screenshot from 2021-08-23 15-48-22

Issue #4261, opened 2021-08-05
helm-charts and istio addons doesn't have default grafana in_cluster_url defined


xeviknal:
I think we have another inconsistency with the definition of a default value. In this case, the grafana `in_cluster_url`.

The CR document says the default value is appropriately set:

```yaml
#      enabled: true
#      health_check_url: ""
#      in_cluster_url: "http://grafana.istio-system:3000"
#      is_core: false
#      url: ""
```
from https://github.com/kiali/kiali-operator/blob/0bcf8f0a91b939346b43d82a468cb4b0f4a37b92/deploy/kiali/kiali_cr.yaml#L524

However, there is no default value set in the defaults main:

```yaml
      health_check_url: ""
      in_cluster_url: ""
      is_core: false
      url: ""
```
from https://github.com/kiali/kiali-operator/blob/0bcf8f0a91b939346b43d82a468cb4b0f4a37b92/roles/default/kiali-deploy/defaults/main.yml#L134

And then, in the server side, we don't have any default value:
https://github.com/kiali/kiali/blob/3a5b136c095990eaaddfe174594279a7a3a83974/config/config.go#L462-L468

Therefore, it looks like we got a gap having grafana enabled but with an empty `in_cluster_url` when kiali is installed via both helm-charts and istio addons.

I've just used the default helm installation with istio 1.11-beta-3 and I got the masthead warning saying grafana is not there. Checking the configmap, I see no `in_cluster_url` setting and therefore kiali server has an empty string there. Curling and empty string obviously gives you a non 200 response status code.

Am I doing something wrong? If not, I guess we should backport it to 1.38.

There is no problem when installing kiali via the operator:
```yaml
- name: Set default Grafana in_cluster_url
  set_fact:
    kiali_vars: "{{ kiali_vars | combine({'external_services': {'grafana': {'in_cluster_url': 'http://grafana.' + kiali_vars.istio_namespace + ':3000'}}}, recursive=True) }}"
  when:
    kiali_vars.external_services.grafana.in_cluster_url == ""
```

from https://github.com/kiali/kiali-operator/blob/0bcf8f0a91b939346b43d82a468cb4b0f4a37b92/roles/default/kiali-deploy/tasks/main.yml#L279-L283

@jmazzitelli does this make sense to you?

jmazzitelli:
[...]

PR #4262, opened 2021-08-05
Adding default grafana in_cluster_url

xeviknal:
fixes https://github.com/kiali/kiali/issues/4261

Now, all the three addons have a default value: prom, jaeger and grafana.

### DIFF
--- a/business/grafana_test.go
+++ b/business/grafana_test.go
@@ -42,6 +42,7 @@ func TestGetGrafanaInfoDisabled(t *testing.T) {
 
 func TestGetGrafanaInfoExternal(t *testing.T) {
 	conf := config.NewConfig()
+	conf.ExternalServices.Grafana.InClusterURL = ""
 	conf.ExternalServices.Grafana.URL = "http://grafana-external:3001"
 	conf.ExternalServices.Grafana.Dashboards = dashboardsConfig
 	config.Set(conf)
@@ -71,6 +72,7 @@ func TestGetGrafanaInfoInCluster(t *testing.T) {
 
 func TestGetGrafanaInfoGetError(t *testing.T) {
 	conf := config.NewConfig()
+	conf.ExternalServices.Grafana.InClusterURL = ""
 	conf.ExternalServices.Grafana.URL = "http://grafana-external:3001"
 	conf.ExternalServices.Grafana.Dashboards = dashboardsConfig
 	config.Set(conf)
@@ -83,6 +85,7 @@ func TestGetGrafanaInfoGetError(t *testing.T) {
 
 func TestGetGrafanaInfoInvalidDashboard(t *testing.T) {
 	conf := config.NewConfig()
+	conf.ExternalServices.Grafana.InClusterURL = ""
 	conf.ExternalServices.Grafana.URL = "http://grafana-external:3001"
 	conf.ExternalServices.Grafana.Dashboards = dashboardsConfig
 	config.Set(conf)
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -465,6 +465,7 @@ func NewConfig() (c *Config) {
 				},
 				Enabled: true,
 				IsCore:  false,
+				InClusterURL: "http://grafana.istio-system:3000",
```
jmazzitelli, line 468: If we are sorting/alphabetizing all of these settings in here, then you should move this up a line. I'll approve in case we want to merge anyway.

jmazzitelli (approved):

config/config.go
jmazzitelli: you need to gofmt this anyway (the format linter failed) - when you do it, just move it up a line.

jmazzitelli (approved):

xeviknal:
@jmazzitelli I needed to adjust the tests about the grafana links. Tests expected to have default empty "" for the `in_cluster_url`.
I think that this have an implication. Users that didn't have `in_cluster_url` set but `URL` will now have explicitely empty the `in_cluster_url`. Is this bad?

@lucasponce do you have an opinion about that? I am no so familiar grafana.

jmazzitelli:
> Users that didn't have `in_cluster_url` set but `URL` will now have explicitly empty the `in_cluster_url`. Is this bad?

If someone sets URL, wouldn't they also set in_cluster_url? I mean, what is the use case where someone has URL set but not in_cluster_url?

xeviknal:
AFAIR the use case is when your grafana is outside the cluster. Mainly. Not 100% sure though.

jmazzitelli:
> AFAIR the use case is when your grafana is outside the cluster. Mainly. Not 100% sure though.

Just FYI: the way the code is written in the operator, the server will never see `in_cluster_url` be empty string - the operator will always set it if it is empty string: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/tasks/main.yml#L279-L283

If `in_cluster_url` == `""` is valid, then we need to change the operator to allow for this.

business/grafana_test.go
```diff
@@ -42,6 +42,7 @@ func TestGetGrafanaInfoDisabled(t *testing.T) {
 
 func TestGetGrafanaInfoExternal(t *testing.T) {
 	conf := config.NewConfig()
+	conf.ExternalServices.Grafana.InClusterURL = ""
```
xeviknal, line 45: If users are not able to set an empty string in the CR for `in_cluser_url`, then they won't be able to reproduce this test scenario. And I am not 100% sure of the implications of it.

xeviknal: But on the other side, the operator by default is always setting grafana's `in_cluster_url`. Therefore, it looks like there is a config that might not be achievable from the operator whilst it is from helm charts. Am I right?

In helm-charts but not with the operator:
```yaml
url: http://grafana.my-org.cat
in_cluster_url: ""
```

jmazzitelli: Looking at this tells me it is OK that you can set inClusterURL to the same as URL. So rather than set it to empty string, just set it to the same as what you set it to URL and it should work.

https://github.com/kiali/kiali/blob/master/business/grafana.go#L127-L132

jmazzitelli: BTW: it is not hard to change the operator to support setting this to empty string if that is what we need.

xeviknal: > So rather than set it to empty string, just set it to the same as what you set it to URL and it should work.

However, we would have two tests recreating the very same scenario. That is why I prefer to keep the suite exactly the same but making explicit the empty string. Otherwise, we would be changing behavior.

pbajjuri20 (approved):
Verified all Grafana links, Nothing is broken

PR #4273, opened 2021-08-12
Add viewOnlyMode property into config object

leandroberetta:
**Description**

This PR adds the property viewOnlyMode into the server configuration. 

This property can be used to avoid doing validations with k8s API and return faster from methods that manage resources.

**Documentation**

```
http://localhost:20001/kiali/api/config
```

```json
{
  ...
  "deployment": {
    "viewOnlyMode": true
  },
  ...
}
```

Frontend PR https://github.com/kiali/kiali-ui/pull/2219

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -9,11 +9,14 @@ import (
 
 	errors2 "k8s.io/apimachinery/pkg/api/errors"
 
+	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/util"
 )
+
+const allResources string = "*"
 
 type IstioConfigService struct {
 	k8s           kubernetes.ClientInterface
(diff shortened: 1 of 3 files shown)


lucasponce:
Hi @leandroberetta, thanks for the PR.
I'm thinking in an optimization that may help in terms of performance.

The business layer performs a permissions query to see if they should render UI actions (create, update, delete).
As this query is typically heavy, when "view_only_mode" flag is set to true, those queries can return "false" all the time and then expand this to the other buttons.

Typically this will affect to Namespace actions, Workload actions (as you are covering), Service actions as well as "Save/Delete" buttons on Istio Config details.

I guess that change can be performed in the business package, to skip the query when the flag is present and return the default value (false).

leandroberetta:
> Hi @leandroberetta, thanks for the PR.
> I'm thinking in an optimization that may help in terms of performance.
> 
> The business layer performs a permissions query to see if they should render UI actions (create, update, delete).
> As this query is typically heavy, when "view_only_mode" flag is set to true, those queries can return "false" all the time and then expand this to the other buttons.

Good idea @lucasponce , I will work on it.

> Note, when false, we need it, specially in OS platforms, where RBAC can have fine granularity and then Kiali would need to maintain that query to see what actions can be rendered.

Yes, when false, everything stay as it its now.

business/istio_config.go
```diff
@@ -9,12 +9,15 @@ import (
...
 	errors2 "k8s.io/apimachinery/pkg/api/errors"
 
+	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/util"
 )
 
+const allResources string = "*"
```
lucasponce, line 19: Minor.
String constants can skip the type:
i.e.
```
const (
	// Kubernetes Controllers
	ConfigMapType             = "ConfigMap"
	CronJobType               = "CronJob"
        ...
```

lucasponce (approved):
A very minor optional suggestion that doesn't require a change.

PR #4285, opened 2021-08-17
[hack] some enhancements and cleanup

jmazzitelli:
1. remove workaround in k8s-minikube.sh olm install code since the bug has been fixed
2. add script that lets you dynamically configure operator
3. remove obsolete aws script


jshaughn (approved):
LOK

PR #4286, opened 2021-08-18
[hack] a couple of new, general purpose scripts

jshaughn:
- A general purpose hack script for adding namespaces
- A Maistra-specific hack script for enabling injection for deployments

### DIFF
--- a/hack/istio/create-mesh-namespace.sh
+++ b/hack/istio/create-mesh-namespace.sh
@@ -55,9 +55,8 @@ Valid command line arguments:
   -in|--istio-namespace <name>: Where the Istio control plane is installed (default: istio-system).
   -n|--namespace <name>: The namespace to be created or deleted. REQUIRED.
   -sa|--service-accounts
-      The service accounts that will be able to authenticate with docker with the credentials you provide.
+      The service accounts to use for adding users to the SecurityContextConstraints.
       You can specify more than one service account separated with spaces.
-      If you are using this to be able to pull Istio addon images, this value should be "prometheus grafana default".
       Default: "default"
   -h|--help: this text
 HELPMSG
@@ -151,9 +150,6 @@ metadata:
   name: "${NAMESPACE}-scc"
 runAsUser:
   type: RunAsAny
-  UID: <none>
-  UID Range Min: <none>
-  UID Range Max: <none>
 seLinuxContext:
   type: RunAsAny
 supplementalGroups:
--- a/hack/istio/create-mesh-namespace.sh
+++ b/hack/istio/create-mesh-namespace.sh
@@ -25,6 +25,7 @@ EOM
 : ${DELETE_NAMESPACE:=false}
 : ${ENABLE_INJECTION:=true}
 : ${ISTIO_NAMESPACE:=istio-system}
+: ${SERVICE_ACCOUNTS:="default"}
 
 while [ $# -gt 0 ]; do
   key="$1"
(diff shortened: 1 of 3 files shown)


hack/istio/enable_maistra_deployment_injection.sh
```diff
@@ -0,0 +1,53 @@
...
+# Given a namepace, prepare the deployments for injection
+# This means:
+# 1. Annotate all of the namespace's Deployments with the sidecar injection annotation
+prepare_maistra() {
+  local ns="${1}"
+
+  for d in $(oc get deployments -n ${ns} -o name)
+  do
+    echo "Enabling sidecar injection for deployment: ${d}"
+    oc patch ${d} -n ${ns} -p '{"spec":{"template":{"metadata":{"annotations":{"sidecar.istio.io/inject": "true"}}}}}' --type=merge
```
jmazzitelli, line 14: I recently asked the Istio team about this and they told me the label should be used now. This is in the docs, also (using label, not annotation) - see: https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

```suggestion
    oc patch ${d} -n ${ns} -p '{"spec":{"template":{"metadata":{"labels":{"sidecar.istio.io/inject": "true"}}}}}' --type=merge
```

jmazzitelli (changes requested):
one quick thing I see that should change

hack/istio/create-mesh-namespace.sh
```diff
@@ -0,0 +1,145 @@
...
+metadata:
+  name: "${NAMESPACE}-scc"
+runAsUser:
+  type: RunAsAny
+seLinuxContext:
+  type: RunAsAny
+supplementalGroups:
+  type: RunAsAny
+users:
+- "system:serviceaccount:${NAMESPACE}:default"
```
jmazzitelli, line 142: Here's the thing with this - if your mesh namespace has a Pod that uses its own service account (that is NOT "default") then this won't work. You need to assign this SCC to all the service accounts that run the pods in the namespace.

jshaughn: OK, this is fixed

hack/istio/create-mesh-namespace.sh
```diff
@@ -0,0 +1,164 @@
...
+Valid command line arguments:
+  -c|--client: either 'oc' or 'kubectl'
+  -d|--delete: either 'true' or 'false'. If 'true' the namespace will be deleted, not installed.
+  -ei|--enable-injection: either 'true' or 'false' (default is true). If 'true' auto-inject proxies for the workloads.
+  -in|--istio-namespace <name>: Where the Istio control plane is installed (default: istio-system).
+  -n|--namespace <name>: The namespace to be created or deleted. REQUIRED.
+  -sa|--service-accounts
+      The service accounts that will be able to authenticate with docker with the credentials you provide.
+      You can specify more than one service account separated with spaces.
+      If you are using this to be able to pull Istio addon images, this value should be "prometheus grafana default".
```
jmazzitelli, line 60: at least pretend you didn't plagiarize :)

hack/istio/create-mesh-namespace.sh
```diff
@@ -0,0 +1,164 @@
...
+  cat <<SCC | $CLIENT_EXE apply -f -
+apiVersion: security.openshift.io/v1
+kind: SecurityContextConstraints
+metadata:
+  name: "${NAMESPACE}-scc"
+runAsUser:
+  type: RunAsAny
+  UID: <none>
+  UID Range Min: <none>
+  UID Range Max: <none>
```
jmazzitelli, line 156: this can't possibly be right

jmazzitelli: you took this from "oc describe" - do a "oc get" for the true yaml. But here's a hint - you don't need to specify anything regarding uid ranges.

hack/istio/create-mesh-namespace.sh
jshaughn: imitation is flattery!

hack/istio/create-mesh-namespace.sh
jshaughn: I just removed that, it was your dumb suggestion to begin with :)

jmazzitelli (approved):
hack scripts bring me joy

PR #4291, opened 2021-08-19
[hack] ci tests will now report the number of failed tests

jmazzitelli:
also, minikube CI script can now send irc message


jshaughn (approved):
LOK

PR #4297, opened 2021-08-23
[hack] update CRC script so it can run the latest

jmazzitelli:
This should help people that want an OpenShift cluster on their local machine. This still may not be enough to run a full Istio+Kiali+demo set up, but it should be able to at least provide a platform to run Istio if not also with Kiali.


jmazzitelli:
Here's one way to expose the CRC cluster to remote clients: https://cloud.redhat.com/blog/accessing-codeready-containers-on-a-remote-server/

Though I found a different way that is easier: https://github.com/code-ready/crc/issues/2667

I will add a new command to the crc script that can be used to manage these firewall rules.

jshaughn (approved):
Fine

Issue #4268, opened 2021-08-10
Support the new istio injection label


jmazzitelli:
We currently can be told what the sidecar injection _annotation_ is:

https://github.com/kiali/kiali-operator/blob/v1.38.0/deploy/kiali/kiali_cr.yaml#L571

But this annotation is deprecated (though is still supported in upstream Istio). The way that this should be done now is via a _label_ with the same name -- see https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy

We need to support the injection label, too, in addition to the annotation.

Server PR: https://github.com/kiali/kiali/pull/4309
UI PR: https://github.com/kiali/kiali-ui/pull/2230
Operator PR: https://github.com/kiali/kiali-operator/pull/413

lucasponce:
Good to know, yes, it seems that not all doc is being updated
https://istio.io/latest/docs/reference/config/annotations/

I think this would only apply to the Workload actions I guess, as detection of sidecars is done directly through Pod containers.

jmazzitelli:
More clarification from the upstream Istio team:

QUESTION: what happens if both are set? I assume when you say Label > Annotation that the label takes precedence. I just want to confirm that. If both label and annotation are set - I will assume annotation is ignored and label is used. Correct?

ANSWER: actually no, [if] either one is false it will not be injected

jmazzitelli:
One weird thing is I don't see annotations working in 1.11:

```
$ _output/istio-1.11.0/bin/istioctl version
client version: 1.11.0
control plane version: 1.11.0
data plane version: 1.11.0 (3 proxies)

$ kubectl get pods -n simple-mesh 
NAME                            READY   STATUS    RESTARTS   AGE
simple-client-5f54f4b59-5hjvz   2/2     Running   0          12m
simple-server-7957d875-s4jtz    1/1     Running   0          12m

$ kubectl get pods -n simple-mesh simple-client-5f54f4b59-5hjvz -o jsonpath='{"annotation-->"}{.metadata.annotations.sidecar\.istio\.io/inject}{"\nlabel-->"}{.metadata.labels.sidecar\.istio\.io/inject}{"\n"}'
annotation-->
label-->true

[...]

PR #4309, opened 2021-08-30
when checking for sidecar injection, check the label as well as the annotation

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4268

UI PR: https://github.com/kiali/kiali-ui/pull/2230
Operator PR: https://github.com/kiali/kiali-operator/pull/413

### DIFF
--- a/models/workload.go
+++ b/models/workload.go
@@ -12,6 +12,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/status"
 )
 
 type WorkloadList struct {
@@ -48,6 +49,8 @@ type WorkloadListItem struct {
 	ResourceVersion string `json:"resourceVersion"`
 
 	// Define if Workload has an explicit Istio policy annotation
+	// Istio supports this as a label as well - this will be defined if the label is set, too.
+	// If both annotation and label are set, if any is false, injection is disabled.
 	// It's mapped as a pointer to show three values nil, true, false
 	IstioInjectionAnnotation *bool `json:"istioInjectionAnnotation,omitempty"`
 
@@ -160,12 +163,32 @@ func (workload *Workload) parseObjectMeta(meta *meta_v1.ObjectMeta, tplMeta *met
 	if tplMeta.Annotations != nil {
 		annotations = tplMeta.Annotations
 	}
-	var annotation string
-	exist := false
-	annotation, exist = annotations[conf.ExternalServices.Istio.IstioInjectionAnnotation]
-	if value, err := strconv.ParseBool(annotation); exist && err == nil {
-		workload.IstioInjectionAnnotation = &value
+
+	// check for automatic sidecar injection - can be defined via label or annotation
+	// if both are defined, either one set to false will disable injection
+	// NOTE: the label (regardless of value) is meaningless in Maistra environments.
+	labelExplicitlyFalse := false // true means the label is defined and explicitly set to false
+	label, exist := workload.Labels[conf.ExternalServices.Istio.IstioInjectionAnnotation]
+	if exist && !status.IsMaistra() {
+		if value, err := strconv.ParseBool(label); err == nil {
+			workload.IstioInjectionAnnotation = &value
+			labelExplicitlyFalse = !value
+		}
 	}
+
+	// do not bother to check the annotation if the label is explicitly false since we know in that case injection is disabled
+	// NOTE: today, if the annotation value is set to "true", that is meaningless for non-Maistra environments.
+	if !labelExplicitlyFalse {
+		annotation, exist := annotations[conf.ExternalServices.Istio.IstioInjectionAnnotation]
+		if exist {
+			if value, err := strconv.ParseBool(annotation); err == nil {
+				if !value || status.IsMaistra() {
+					workload.IstioInjectionAnnotation = &value
+				}
+			}
+		}
+	}
+
 	workload.CreatedAt = formatTime(meta.CreationTimestamp.Time)
 	workload.ResourceVersion = meta.ResourceVersion
 	workload.AdditionalDetails = GetAdditionalDetails(conf, annotations)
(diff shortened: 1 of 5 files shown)


jmazzitelli:
I modified the code to only return a non-nil IstioInjectionAnnotation if the pod label is true or false OR if the pod annotation is false. If the pod annotation is true, that value is meaningless and so does not affect the value of IstioInjectionAnnotation.
Here's what it looks like if the label is true and if the annotation is true:

label == true
injection-true

annotation == true (meaningless value - you cannot opt in for sidecar injection with the annotation - it is opt-out only)
injection-false

jmazzitelli:
I just realized something - the annotation has different behavior when running with Maistra. The annotation value of "true" DOES have meaning. I need to rethink that change I did.

jmazzitelli:
this PR now provides a common util for use within the Kiali Server -- `status.IsMaistra()` -- this will return `true` if Kiali server is running within a Maistra environment.

This is used in this PR due to the different semantics of the auto-injection pod label and pod annotation between Maistra and non-Maistra environments. There are comments in the code to explain those different semantics so hopefully the code will make sense once you read the comments.

jmazzitelli:
asking @lucasponce to review - this is adding a common API that can be used in the server anywhere. Just want to make sure you review it first before it gets merged.

lucasponce:
@jmazzitelli I'm starting to review your PR, thanks, I guess you saw that you'd need also to update the UI too:
image

So, I think you need a combined PR for the frontend to merge all together (unless you want to merge them separated).

status/versions.go
```diff
@@ -43,6 +43,17 @@ var (
 	istioVersionExpr          = regexp.MustCompile(`([0-9]+\.[0-9]+\.[0-9]+)`)
 )
 
+var (
```
lucasponce, line 46: Minor var -> const if those are constants

models/workload.go
```diff
@@ -48,6 +49,8 @@ type WorkloadListItem struct {
 	ResourceVersion string `json:"resourceVersion"`
 
 	// Define if Workload has an explicit Istio policy annotation
+	// Istio supports this as a label as well - this will be defined if the label is set, too.
+	// If both annotation and label are set, if any is false, injection is disabled.
 	// It's mapped as a pointer to show three values nil, true, false
 	IstioInjectionAnnotation *bool `json:"istioInjectionAnnotation,omitempty"`
```
lucasponce, line 55: Probably you may need to differentiate if you are using labels vs annotations and tell that to the UI which should build a patch depending of the case

https://github.com/kiali/kiali-ui/blob/master/src/components/IstioWizards/WizardActions.ts#L1407

Suggestion a new field with "IstioInjectionLabel" in the similar way, and let the UI to use one or another.

jmazzitelli: I originally thought of that but decided against it because of:

1. the amount of change would be a lot more (it would also require a change to the operator also)
2. adding yet another config option that will likely never change and will likely never have a value different from the annotation label name

In other words, the name of this label is (and will always be) the same as the name of the annotation (`sidecar.istio.io/inject`). I don't see any benefit to creating a separate config option here because it will likely never be the case that the annotation label will be one name and the name of the label will be different than that.

Do you see a benefit to creating a separate one?

jmazzitelli:
> So, I think you need a combined PR for the frontend to merge all together (unless you want to merge them separated).

The way the code is, the UI should not have to change. Because of the different semantics of the pod injection label between Maistra and upstream, that UI code can't use the label to add auto-injection. (see my comments above for a description of these differences).

And because the name of both the label and annotation are the same (and we don't have that extra superfluous config option you suggested earlier), the UI does not have to change anything when its displaying whether auto-injection is on or off for the workload.

If you can see a reason why we would need to add a second config option (istioInjectionLabel), then that would cause us to change the UI and operator (as well as this server PR)... this is why I decided not to add that extra config option - it isn't strictly needed and would only cause some churn in both operator and UI that is not needed as this PR is written now.

lucasponce:
So, your suggestion is that the Workload action will always use the POd annotation in any case, even if the user decides to use the label in the workload instead.

I'm not initially against that, just thinking if there could be any clash.

jmazzitelli:
> So, your suggestion is that the Workload action will always use the Pod annotation in any case, even if the user decides to use the label in the workload instead.

The only way we could have the UI use either label or annotation is if we were able to tell the UI that it is running in a Maistra or non-Maistra environment. This is because the pod label is completely meaningless in a Maistra environment (i.e. Maistra never looks at the label to know if it should inject the sidecar. Maistra ONLY ever looks at the annotatio). So it would not make sense (and in fact would probably be considered a bug) if the UI offered the user to add the label (as opposed to the annotation) when running in a Maistra environment.

Since upstream Istio also supports the annotation to enable injection, it seems the easiest thing to do is to leave the UI alone - the UI would be opinionated in this case. It will always use the annotation. Thus it will support both upstream Istio and Maistra as is.

Just to further explain the label and annotation support more clearly:


_UPSTREAM ISTIO_:
* **Namespace Label**: supports "true" and "false"
* **Pod Label**: supports "true" and "false"
* **Pod Annotation**: supports "false" only

_MAISTRA_:
* **Namespace Label**: not supported
* **Pod Label**: not supported
* **Pod Annotation**: "true" and "false"

jmazzitelli:
OK, after discussing, I will change this and the UI to support different paths for non-Maistra and Maistra environs. If non-Maistra, the label will be used by the UI when it adds it to the workload; annotation will be used when on Maistra.

The UI will have an "isMaistra" flag for it to use here and anywhere else the UI needs to make this distinction.

jshaughn (commented):
Visually looks OK to me.  I'll defer to @lucasponce for a final approval as he has been the main point of contact about the feature addition.

jmazzitelli:
I added molecule testing for this - see the operator PR https://github.com/kiali/kiali-operator/pull/413

lucasponce (approved):
Reviewed and tested locally.

lucasponce:
For QE: I think covering the upstream case would be enough (I think). Trying to build a Maistra version (Istio 1.9 based) with Kiali 1.40 can be more complex and probably this use case can be tested in the future in the context of SM 2.2.

Issue #1525, opened 2019-08-27
Be able to set the logging level for istio and envoy logs from Kiali UI


lucasponce:
This feature would add being able to set the logging level from inside the Kiali UI workload -> logs tab.
This would be a big help when trying to debug something in istio using the logs and not having to go back out to change the logging level. And especially useful for the envoy proxy log level (--proxyLogLevel=debug)

JIRA:
https://issues.jboss.org/browse/KIALI-3013

lucasponce:
I'm not sure if this feature can overlaps with a regular k8s console.
I agree it would be a good idea to have some control, but not sure with all those imply: Kiali start managing Pods for restart operations, config changes, etc.

lucasponce:
This sounds valid as soon as we are adding more features around EnvoyConfig.

Istioctl allows to change the proxy config, and in particular, to modify the log level of a particular pod.

```
istioctl proxyu-config log --level ...
```

As logs tab relies on proxy logic, it sounds interesting to explore if that feature could be added at Workload level.

I mean Workload, because this would mean a re-start of the pod.

In general, Workload management tasks are out of the scope of Kiali, but as it happens with istio labels, this specific action, to change the log level of a container sounds correct and in line with other similar operations related to the EnvoyConfig.

cc @xeviknal as I suspect that the implementation used around accessing to the EnvoyConfig could be similar to this.

I don't add it to the Project, as it may require more discussion, but this issue sounds ok from my side.

lucasponce:
Related to https://github.com/kiali/kiali/discussions/4082 discussion

nrfox:
> I mean Workload, because this would mean a re-start of the pod.

In testing this by `kubectl exec` into the proxy pod and doing `curl -XPOST localhost:15000/logging?level=debug`, the log level updated successfully without a pod restart. If the istio proxy works the same way then a pod restart shouldn't be necessary for updating this.

nrfox:
How about having the dropdown just to the left of the "log lines" dropdown?

![loglevellocation](https://user-images.githubusercontent.com/6226732/130290915-f11acf3d-153d-459e-8f7c-7dc9c9ed774c.png)

jshaughn:
Is it actually possible to know the current log level?  Or just to set a new one?  I'm not sure whether I'd like to add more to the toolbar or instead control it from the kebab.

nrfox:
[...]

PR #4312, opened 2021-08-30
Add route to update proxy logging for specific pod

nrfox:
Adds a route, `/api/namespaces/{namespace}/pods/{pod}/logging`, to set the log level for a pod's proxy.

Relates to #1525 
Front end changes: https://github.com/kiali/kiali-ui/pull/2236

To test
- Deploy bookinfo 
- Choose a bookinfo pod with a proxy
- Get logs for proxy pod and note the log level: `kubectl logs -f <pod_name> -n <bookinfo_ns>  -c istio-proxy`
- curl kiali api to update pod's proxy log level `curl -k -XPOST "https://<minikube_ip>/kiali/api/namespaces/<bookinfo_namespace>/pods/<pod_name>/logging?level=debug"`
- Get logs for proxy pod and note the updated log level: `kubectl logs -f <pod_name> -n <bookinfo_ns>  -c istio-proxy`


nrfox:
@xeviknal I added you since you're more familiar with the pod proxy code :)

business/proxy_logging.go
```diff
@@ -0,0 +1,46 @@
...
+
+type ProxyLoggingService struct {
+	k8s kubernetes.ClientInterface
+}
+
+func (in *ProxyLoggingService) SetLogLevel(ctx context.Context, namespace, pod, level string) error {
+	path := fmt.Sprintf("/logging?level=%s", level)
+
+	localPort := httputil.Pool.GetFreePort()
+	f, err := in.k8s.GetPodPortForwarder(namespace, pod, fmt.Sprintf("%d:%d", localPort, envoyAdminPort))
```
nrfox, line 24: Does this need to return a pointer to an `Interface`? Can it be updated to just return a `PortForwarder` interface? Then start can be called as `f.Start()` instead of `(*f).Start()`.

util/httputil/http_util.go
```diff
@@ -44,6 +46,36 @@ func HttpGet(url string, auth *config.Auth, timeout time.Duration) ([]byte, int,
 	return body, resp.StatusCode, err
 }
 
+func HttpPost(ctx context.Context, url string, auth *config.Auth, body io.Reader) ([]byte, int, error) {
```
nrfox, line 49: This is slightly different from `HttpGet` in that it accepts a `context` as its first arg. I thought it'd be nice to passthrough the request.Context to propagate cancellation but I'm happy to update this to look more like `HttpGet`.

util/httputil/http_util_test.go
```diff
@@ -1,12 +1,16 @@
-package httputil
+package httputil_test
```
nrfox, line 1: This is just my own personal preference to ensure testing the public API but again happy to update this back.

lucasponce:
@nrfox I can help you reviewing this. @xeviknal will be on PTO for some weeks and I'm familiar with the latest changes we discussed about the pod proxy code.

nrfox:
> @nrfox I can help you reviewing this. @xeviknal will be on PTO for some weeks and I'm familiar with the latest changes we discussed about the pod proxy code.

Sounds great, thanks!

business/proxy_logging.go
lucasponce: I guess these wraps are more inspired from the client itself
image

That they use the pointer to interface to make the type opaque and just focused on the interface.

This is more code style, as @xeviknal was the original author, when he's back that design can be revisited, but looking the client-go I see some similar interfaces, I guess that's the rationale.

business/proxy_logging.go
```diff
@@ -0,0 +1,46 @@
...
+
+const (
+	envoyAdminPort = 15000
+)
+
+type ProxyLoggingService struct {
+	k8s kubernetes.ClientInterface
+}
+
+func (in *ProxyLoggingService) SetLogLevel(ctx context.Context, namespace, pod, level string) error {
```
lucasponce, line 20: I'd recommend to place the low level logic in the kubernetes/istio.go just to have in that package the logic to define the Envoy ports, endpoints and portforwarding

image

Nothing wrong, just trying to put it in the sample place for future optimizations.

util/httputil/http_util.go
lucasponce: I'd propose as you say, to keep it similar as "HttpGet" first
image

To not introduce a refactoring in the use of these two close methods, as there would be several places.
Also, how to pass the context may vary, and probably a nice refactoring would touch more files, I think.

business/proxy_logging.go
nrfox: Yeah it's definitely just code style. I just thought it looked odd when calling/mocking that package but it's pretty unimportant . The difference between the `PortForwarder` in client-go and the `PortForwarder` in kiali is that in client-go it is a [struct](https://pkg.go.dev/k8s.io/client-go@v0.22.1/tools/portforward#PortForwarder) whereas in kiali it is an [interface](https://pkg.go.dev/github.com/kiali/kiali@v1.39.0/util/httputil#PortForwarder).

util/httputil/http_util.go
nrfox: Sure I'll update to be more like httpget and save the refactoring for another change. Updating all instances of httpget as well is definitely a larger change.

business/proxy_logging.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+// ProxyLoggingService is a thin layer over the kube interface for proxy logging functions.
+type ProxyLoggingService struct {
+	k8s kubernetes.ClientInterface
+}
+
+// SetLogLevel sets the pod's proxy log level.
+func (in *ProxyLoggingService) SetLogLevel(namespace, pod, level string) error {
+	return in.k8s.SetProxyLogLevel(namespace, pod, level)
```
lucasponce, line 14: Suggestion, perhaps in the business layer you may want to check if the pod is in the mesh before to trying to perform the change.

This info is cached and you could re-use a single call to "kialiCache.GetPodProxyStatus(ns, pod)" (similar to used in the business/proxy_status.go) to let you perform that operation.

Non a blocker, you can rely on the runtime error of the "SetProxyLogLevel", but I was wondering that you may have that info available to decide if to throw the operation.

handlers/proxy_logging.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"fmt"
+	"net/http"
+	"strings"
+
+	"github.com/gorilla/mux"
+)
+
+var (
+	validLogLevels = []string{"off", "trace", "debug", "info", "warning", "error", "critical"}
```
lucasponce, line 12: Perhaps the "handler" layer should not have information the list of available levels.
I think that could be part of the business package.

To give an example, if you look into the "handlers/istio_config.go" you'll see that the there is a criteria parsed and built in the "business" package. 

Then if you need to maintain this list, that would be located under the "business" layer.

My suggerstion would be to move the "validLogLevels" and "isValidLogLevel" to the business layer and just invoke in the "handler" a business.IsValidLogLevel.

It's just code style, logic itself is ok.

lucasponce:
Thanks @nrfox, I put a couple of minor comments related to style and organization between handler/business, but the PR looks ok overall.

handlers/proxy_logging.go
nrfox: Sure that makes sense to me to move it there

business/proxy_logging.go
nrfox: If the pod is not in the mesh, would `GetPodProxyStatus` just return the runtime error returned by doing `curl localhost:15000/debug/` through the proxy and failing? That seems the same as just returning the runtime error from `SetProxyLogLevel`?

nrfox: Although it would probably be better to fail on doing a `GET` request to a port in the pod that is potentially running something else if it's not running the proxy rather than failing on a `POST` that could change something. Chances are small that the pod is running another service on `localhost:15000` that would be altered in some way from the request but not zero.

I'll update by doing that pod proxy status check first as you suggested.

nrfox:
I realized that you can restrict openapi to certain values using enum: https://github.com/kiali/kiali/pull/4312/files#diff-10cfbe600d3f92e886aea1d00df1367f9899549b1a3235442e5191b66ced7e71R2508-R2516 so added that as well

nrfox:
@jshaughn how's this look?

Screenshot from 2021-09-13 14-15-11

lucasponce:
@nrfox I've sent a minor change to your branch, to consolidate
https://github.com/nrfox/kiali/pull/1

Just to add an audit log on it.

nrfox:
> @nrfox I've sent a minor change to your branch, to consolidate
> [nrfox#1](https://github.com/nrfox/kiali/pull/1)
> 
> Just to add an audit log on it.

Sorry I missed that yesterday. Thanks, merged!

lucasponce (approved):

pbajjuri20:
LGTM!

Tested with Proxy

with_proxy



Tested without Proxy

without_proxy

pbajjuri20 (approved):

PR #4313, opened 2021-08-31
code style, keep CamelCase in RequestHealth

zhucheng01:
[ ] Configuration Infrastructure
[ ] Docs
[ ] Installation
[ ] Networking
[ ] Performance and Scalability
[ ] Policies and Telemetry
[ ] Security
[ ] Test and Release
[X ] User Experience
[ ] Developer Infrastructure

Pull Request Attributes

Please check any characteristics that apply to this pull request.

[ ] Does not have any user-facing changes. This may include CLI changes, API changes, behavior changes, performance improvements, etc.

Before:
type RequestHealth struct {
	Inbound            map[string]map[string]float64 `json:"inbound"`
	Outbound           map[string]map[string]float64 `json:"outbound"`
	HealthAnnotations  map[string]string             `json:"healthAnnotations"`
	inboundSource      map[string]map[string]float64
	inboundDestination map[string]map[string]float64
}

After:
type RequestHealth struct {
	Inbound            map[string]map[string]float64 `json:"inbound"`
	Outbound           map[string]map[string]float64 `json:"outbound"`
	HealthAnnotations  map[string]string             `json:"healthAnnotations"`
	InboundSource      map[string]map[string]float64
	InboundDestination map[string]map[string]float64
}

### DIFF
--- a/models/health.go
+++ b/models/health.go
@@ -31,8 +31,8 @@ func NewEmptyRequestHealth() RequestHealth {
 		Inbound:            make(map[string]map[string]float64),
 		Outbound:           make(map[string]map[string]float64),
 		HealthAnnotations:  make(map[string]string),
-		inboundSource:      make(map[string]map[string]float64),
-		inboundDestination: make(map[string]map[string]float64),
+		InboundSource:      make(map[string]map[string]float64),
+		InboundDestination: make(map[string]map[string]float64),
 	}
 }
 
@@ -98,8 +98,8 @@ type RequestHealth struct {
 	Inbound            map[string]map[string]float64 `json:"inbound"`
 	Outbound           map[string]map[string]float64 `json:"outbound"`
 	HealthAnnotations  map[string]string             `json:"healthAnnotations"`
-	inboundSource      map[string]map[string]float64
-	inboundDestination map[string]map[string]float64
+	InboundSource      map[string]map[string]float64
+	InboundDestination map[string]map[string]float64
 }
 
 // AggregateInbound adds the provided metric sample to internal inbound counters and updates error ratios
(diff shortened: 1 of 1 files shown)


jmazzitelli (commented):
Be careful here - case has semantic information in Go. This is not all about code-style. 

https://golang.org/ref/spec#Exported_identifiers

It is quite possible this code is utilizing lowercase for a reason (e.g. the struct does not want these variables exported). Do not approve/merge without the original author reviewing this.

jmazzitelli:
It looks like this might have been done purposefully - see PR: https://github.com/kiali/kiali/pull/3653
cc @lucasponce - you will want to review. Close this PR is you purposefully wanted these as private struct vars. Otherwise, you can approve as you see fit.

lucasponce:
@zhucheng01 first, thanks for opening a PR in Kiali, we appreciate the contributions.

But the PR sounds a little bit odd to me.

In particular the backend "RequestHealth" is used as part of the Health calculation logic, in that context the "inboundSource" and "outboundSource" are opaque as internal variables that shouldn't be used outside the "models" package:
image

So, the mapped in the UI only takes the public parameters:
image

The question is, this change is triggered to any other followup work where you'd need to make those parameters public ?

Perhaps with more context we can help you to what would be the best way to achieve what you'd like.

But I don't see a strong reason to change that visibility for an internal variable used to populate a public field.

Do I guess you are using that backend API in a different scenario than the Kiali UI ?

lucasponce:
@zhucheng01 I'm going to close this PRs. 
As discussed https://github.com/kiali/kiali/pull/4313#issuecomment-911577010 this change doesn't seem justified.

We really appreciate the effort to clone the project, review the code and propose a PR !

The whole team can help to provide guidance for new contributors.

You could review the list of issues in the "backlog" in case you may have interest in some area.

https://github.com/kiali/kiali/issues?q=is%3Aissue+is%3Aopen+label%3Abacklog

Or if you want to start with some optimizations/refactoring as this PR proposes, that is not part of the backlog list, I suggest to join the Istio Slack -> #kiali channel and try to contact the team, we are sure that there are multiple areas to collaborate.

Issue #1440, opened 2019-08-26
Add info/tooltip on virtual service that doesn't have a gateways section


lucasponce:
There is a (valid) option for Virtual Service not to have the gateways section in its config.
In such case, there is an "invisible" rule that uses the reserved word "mesh" and applies the VS to all sidecards in the mesh.
This is not a validation issue because it is a valid configuration, but at the same time this is an "invisible" rule that users might not be aware of its impact without digging into documentation.
I propose to add an info/tooltip to VS that does not have a gateways section that explains the impact of this.
Please see attached screenshot of the Istio docs.

JIRA:
https://issues.jboss.org/browse/KIALI-1787

lucasponce:
Valid.
Good one for newcomers.

PR #4335, opened 2021-09-07
Added info message on VirtualService with missing gateways

hhovsepy:
Fix for RFE https://github.com/kiali/kiali/issues/1440

Showing warning message on VS when no gateway is added.

Screenshot from 2021-09-07 15-42-53


leandroberetta (commented):
@hhovsepy great addition, this information is very helpful for Istio newcomers.

A comment about the warning message, what do you think about listing the "mesh" gateway in the Gateways list instead of the warning. 

For example: 

mesh

This is already happening when the "mesh" gateway is explicitly configured in the vs.
The tooltip describes in a similar way what the warning is describing.

hhovsepy:
You mean  in the case when gateways are missing, list 'mesh' gateway in Gateways list on right side with the tooltip?
> A comment about the warning message, what do you think about listing the "mesh" gateway in the Gateways list instead of the warning.

leandroberetta:
> Thanks you @leandroberetta . You mean in the case when gateways are missing, list 'mesh' gateway in Gateways list on right side with the tooltip?
> 
> > A comment about the warning message, what do you think about listing the "mesh" gateway in the Gateways list instead of the warning.

Yes, if you're agree and seems fine for you, I'm recommending this because when the mesh gateway is present, we are already listing it in the gateways list, so when no gateways are declared, we can list the mesh gateway by default, instead of the warning. But it's a suggestion, let me know what you think.

hhovsepy:
+1 it seems a better solution than introducing new validation message.
> Yes, if you're agree and seems fine for you, I'm recommending this because when the mesh gateway is present, we are already listing it in the gateways list, so when no gateways are declared, we can list the mesh gateway by default, instead of the warning. But it's a suggestion, let me know what you think.

hhovsepy:
@lucasponce wdyt,  when gateways are not set, is the tooltip on right panel confusing for users?

lucasponce:
I'm in favor of @leandroberetta comment here.
I guess this is not a validation but more a pure informational state.
I like to show the "mesh" in the panel when there is no gateway in the virtualservice.
Perhaps in that case you can put something like 
image

Just to map exactly the same description provided by istio.io.

leandroberetta:
> I'm in favor of @leandroberetta comment here.
> I guess this is not a validation but more a pure informational state.
> I like to show the "mesh" in the panel when there is no gateway in the virtualservice.
> Perhaps in that case you can put something like
> image
> 
> Just to map exactly the same description provided by istio.io.

+1 on the tooltip message but only when there is no explicit gateway. If there are gateways and one is the "mesh", I would mantain the original tooltip message, or something similar to what @lucasponce proposed but without mentioning the lack of the gateways field.

hhovsepy:
I am preparing a solution in UI, when there is not gateway, it still displays 'mesh' gateway but with the tooltip message from istio.io

hhovsepy:
Closing this PR in a favor of another PR in UI: https://github.com/kiali/kiali-ui/pull/2235

Issue #4318, opened 2021-08-31
Running ./ci-minikube-molecule-tests.sh fails


prary:
With '''set -u''' enabled scripts fails with '''variable unbound error'''

https://github.com/kiali/kiali/blob/2c22b05724f967a95f3204130ce5327c6e9b4a0b/hack/ci-minikube-molecule-tests.sh#L111

Steps to reproduce it:
Run ./ci-minikube-molecule-tests.sh with default flags

nrfox:
I get the same error. Judging by the name of the script, my guess is this is usually run in a CI context where that option is passed in but it seems like it should have an `if upload_logs == true` guard around those variable assignments. @jmazzitelli you might be more familiar with this.

jmazzitelli:
Well, I would say, "don't add `set -u`" to the script and it'll work :)

If you want to add a `set -u` then the fix for this would be:

```
LOGS_LOCAL_DIRNAME_ABS="${LOGS_DIR:-}"
```

This logs stuff is used if "upload logs" feature is turned on - and when it is, you will be setting that logs dir so under normal circumstances this isn't a problem. Hence why no one cared.

lucasponce:
I read then this issue can be considered a minor maintenance one (but valid). Isn't it ?

jmazzitelli:
Yes @lucasponce . We could do that fix I mention above which would produce a more human readable error. So it is very minor but valid.

jmazzitelli:
PR to fix: https://github.com/kiali/kiali/pull/4347

PR #4347, opened 2021-09-13
spit out human-readable error if upload logs is true but the user did not provide a logs dir

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4318


nrfox (approved):
Works great!

Issue #4316, opened 2021-08-31
Support it in ServicesEntries


hhovsepy:

hhovsepy:
The support was added on following validations:
"KIA0202 This host has no matching entry in the service registry (service, workload or service entries)",
"KIA0104 This host has no matching entry in the service registry",
"KIA1004 This host has no matching entry in the service registry"

PR #4358, opened 2021-09-16
Added cross-namespace validation support on validation KIA0202 for exportTo field of ServiceEntry

hhovsepy:
RFE https://github.com/kiali/kiali/issues/3061
Subtask : https://github.com/kiali/kiali/issues/4316

On validation message "KIA0202 This host has no matching entry in the service registry (service, workload or service entries)" is now added a support of cross-namespace checking of ServiceEntries.

So create a ServiceEntry:
Screenshot from 2021-09-16 10-46-24

Before when linking from DR to that SE which was in different namespace, it was showing an error:
Screenshot from 2021-09-16 08-40-52

Now error on host is not shown:
Screenshot from 2021-09-16 10-48-11

But try to export SE into local namespace only:
Screenshot from 2021-09-16 11-17-51

The error is shown, as SE is not available into DR's namespace:
Screenshot from 2021-09-16 11-18-46
Screenshot from 2021-09-16 11-19-04

### DIFF
--- a/business/checkers/destinationrules/no_dest_checker.go
+++ b/business/checkers/destinationrules/no_dest_checker.go
@@ -49,7 +49,7 @@ func (n NoDestinationChecker) Check() ([]*models.IstioCheck, bool) {
 											stringLabels[k] = s
 										}
 									}
-									if !n.hasMatchingWorkload(fqdn.Service, stringLabels) {
+									if !n.hasMatchingWorkload(fqdn.String(), stringLabels) {
 										validation := models.Build("destinationrules.nodest.subsetlabels",
 											"spec/subsets["+strconv.Itoa(i)+"]")
 										if n.isSubsetReferenced(dHost, innerSubset["name"].(string)) {
@@ -122,7 +122,7 @@ func (n NoDestinationChecker) hasMatchingWorkload(service string, subsetLabels m
 
 func (n NoDestinationChecker) hasMatchingService(host kubernetes.Host, itemNamespace string) bool {
 	// Check wildcard hosts - needs to match "*" and "*.suffix" also..
-	if strings.HasPrefix(host.Service, "*") {
+	if host.IsWildcard() {
 		return true
 	}
 
@@ -142,7 +142,7 @@ func (n NoDestinationChecker) hasMatchingService(host kubernetes.Host, itemNames
 	}
 
 	// Check ServiceEntries
-	if kubernetes.HasMatchingServiceEntries(host.Service, n.ServiceEntries) {
+	if kubernetes.HasMatchingServiceEntries(host.String(), n.ServiceEntries) {
 		return true
 	}
 
--- a/business/checkers/destinationrules/no_dest_checker_test.go
+++ b/business/checkers/destinationrules/no_dest_checker_test.go
@@ -454,6 +454,165 @@ func TestWildcardServiceEntry(t *testing.T) {
 	assert.Empty(vals)
 }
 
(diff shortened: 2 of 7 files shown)


lucasponce:
I'm seeing something strange in the validations.
I'm testing "master" just to test the scenario before to test it on this branch.

I have this scenario:

```
kind: ServiceEntry
metadata:
  name: edition-cnn-com
  namespace: alpha
spec:
  hosts:
    - edition.cnn.com
  ports:
    - number: 80
      name: http-port
      protocol: HTTP
      targetPort: 443
    - number: 443
      name: https-port
      protocol: HTTPS
  resolution: DNS
---
kind: DestinationRule
metadata:
  name: edition-cnn-com
  namespace: beta
spec:
  host: edition.cnn.com
  trafficPolicy:
    portLevelSettings:
      - port:
          number: 80
        tls:
          mode: SIMPLE # initiates HTTPS when accessing edition.cnn.com
```

But I get a KIA0202:
image

I can see that the ServiceEntry is also included in the registry:
image

So I have the suspects that perhaps there is some regression introduced in the validations.

I'm commenting this because I'm going to review this potential regression in the context of this PR as well
(so it will take more time from my side, but I wanted to give some update to put in context).

lucasponce:
Ouch, please, don't take my previous comment.
It was a cached issue from my environment, once that the registry status is correct I saw the validation fixed.

Sorry for the noise.
image

hhovsepy:
This first scenario is the expected behaviour which should be in master. The error is shown as cross-namespace validation is not supported yes.

The purpose of this PR is to consider cross-namespace ServiceEntries while validation DestinationRule, and do not show error KIA0202 in these cases.

And the second behavior @lucasponce you are facing, I also see in master or testing clusters, not sure how this happens, somehow the ServiceEntries are cached. But when running master locally with cache off, then error KIA0202 is shown as expected.

lucasponce:
Ok, I guess we can move in several directions.
I suggest that we can consolidate this work, but I think I'm going to create an Epic describing how we can use the Istio registry to perform those validations and consolidate (and hopefully simplify) the validation logic for several conditions.

In that case, validations can save to pull kubernetes services, as the registry will give you all visibility of the cluster and probably some cross-validation logic can be simplified.

hhovsepy:
+1 on that, as the registry usage is for all other validations as well
> Ok, I guess we can move in several directions.
> I suggest that we can consolidate this work, but I think I'm going to create an Epic describing how we can use the Istio registry to perform those validations and consolidate (and hopefully simplify) the validation logic for several conditions.

lucasponce (approved):
LGTM
As commented, I collected my thoughts about how we can consolidate cross-namespace validation in a wider context using the Istio registry.
https://github.com/kiali/kiali/issues/4382

That will trigger more changes, but I think it's ok to consolidate these changes meanwhile we discuss the #4382 for future Sprints.

prachiyadav (changes requested):
export SE into local namespace only but error is not shown, as SE is not available into DR's namespace

Screenshot from 2021-09-23 17-32-00
Screenshot from 2021-09-23 17-15-47

hhovsepy:
> export SE into local namespace only but error is not shown, as SE is not available into DR's namespace
It seems because of cache issue https://github.com/kiali/kiali/issues/4382
Can you please retest with cache disabled, or create a new SE with different name?

prachiyadav (approved):
Approving change as disabling the cache make this PR work.

Screenshot from 2021-09-23 17-52-47

PR #4360, opened 2021-09-16
Ensure only active ReplicaSets are returned from GetReplicaSets

### LINKED ISSUE #4141: Workloads flap between OK and Not Ready w/ Argo Rollout CR
### General remarks
Workloads intermittently show as `OK` or `Not Ready`

### Describe the bug
While using Argo Rollouts, the `Rollout` appears to flap between the `OK` and `Not Ready` statuses.

This happens even though the pods from the deployment/rollout are running without issues.

The UI says that there are no available pods from one time to another, even though there is one synced proxy.

### Versions used
**Kiali UI** 1.36.0 (ceb2b93f33f1c0445b51518691dcb48c2c801f8a)
**Kiali Server** v1.36.0 (48c624aa9950b62b96c3c43591900b0291d483cf)
**Kiali Container** v1.36.0

### Components
**Istio** 1.9.0
**Prometheus** 2.26.0
**Kubernetes** v1.19.8-eks-96780e
**Grafana URL** N/A
**Jaeger URL** N/A

### To Reproduce
1. Deploy a working Rollout
2. Click on "Workloads" within the UI
3. See error

There's no output from the kiali deployment that could hint to a polling issue of any sort.

### Expected behavior
Kiali should represent the state appropiately and consistently, or there should be a way to determine the actual reason of this behavior (troubleshooting, documentation, logs, etc).

jshaughn:
This PR aims to ensure that Kiali does not consider historical replicasets, and ignores their pods when "discovering" workloads.  Kiali identifies workloads using sort-of a bottom-up approach that starts with Pods and looks for thir owners. The owner can be a ReplicaSet. Our list of Pods is cached (as is our list of ReplicaSets) and I think what happens is that we get a cached pod for an old ReplicaSet. Our list of ReplicaSets contains all the ReplicaSets for the namespace, both active and inactive (note that inactive RS can result from Deployments that have revisionHistoryLimit > 0). So we can get a cached pod that is owned by an inactive ReplicaSet. For Kiali, I don't think old ReplicaSets for the same owner (Deployment) are useful. If we filter them away I think the related issue will be resolved.

Fixes #4141

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -501,7 +501,8 @@ func fetchWorkloads(layer *Layer, namespace string, labelSelector string) (model
 		// Check if namespace is cached
 		// Namespace access is checked in the upper caller
 		if IsNamespaceCached(namespace) {
-			repset, err = kialiCache.GetReplicaSets(namespace)
+			// repset, err = kialiCache.GetReplicaSets(namespace)
+			repset, err = layer.k8s.GetReplicaSets(namespace)
 		} else {
 			repset, err = layer.k8s.GetReplicaSets(namespace)
 		}
@@ -1081,7 +1082,8 @@ func fetchWorkload(layer *Layer, namespace string, workloadName string, workload
 		// Check if namespace is cached
 		// Namespace access is checked in the upper call
 		if IsNamespaceCached(namespace) {
-			repset, err = kialiCache.GetReplicaSets(namespace)
+			// repset, err = kialiCache.GetReplicaSets(namespace)
+			repset, err = layer.k8s.GetReplicaSets(namespace)
 		} else {
 			repset, err = layer.k8s.GetReplicaSets(namespace)
 		}
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -501,8 +501,7 @@ func fetchWorkloads(layer *Layer, namespace string, labelSelector string) (model
 		// Check if namespace is cached
 		// Namespace access is checked in the upper caller
 		if IsNamespaceCached(namespace) {
-			// repset, err = kialiCache.GetReplicaSets(namespace)
-			repset, err = layer.k8s.GetReplicaSets(namespace)
+			repset, err = kialiCache.GetReplicaSets(namespace)
 		} else {
 			repset, err = layer.k8s.GetReplicaSets(namespace)
 		}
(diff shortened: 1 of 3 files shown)


business/workloads.go
```diff
@@ -1081,7 +1082,8 @@ func fetchWorkload(layer *Layer, namespace string, workloadName string, workload
 		// Check if namespace is cached
 		// Namespace access is checked in the upper call
 		if IsNamespaceCached(namespace) {
-			repset, err = kialiCache.GetReplicaSets(namespace)
+			// repset, err = kialiCache.GetReplicaSets(namespace)
```
lucasponce, line 1085: I'll review it in detail, but disabling the cache for this doesn't look correct.
Note that most of the tools today use the informers (cache) as a source of truth for the controllers.

But I'll fetch the whole patch to add detailed comments, but this has created a warning on this PR.

jshaughn: Oh, that's totall wrong, that was for debugging.  I'll fix it, sorry.

jshaughn: pushed change, stupid of me

lucasponce:
I understand the design, but not sure if it's 100% correct.
At least I see that it misses the case then the replicaset is the top controller.
Perhaps just covering that case would work for this scenario.

This case is easy to test with [1] example.

[1] https://kubernetes.io/es/docs/concepts/workloads/controllers/replicaset/#ejemplo

jshaughn:
> I understand the design, but not sure if it's 100% correct. At least I see that it misses the case then the replicaset is the top controller.  Perhaps just covering that case would work for this scenario.

I'll take another look, thanks.

jshaughn:
@lucasponce PR updated to handle the "own-controller" replicaset case.

lucasponce (approved):
LGTM
I perform a quick smoke test using Argo Rollout and ReplicaSet
image

And think look ok:
image

image

But I haven't tested more deep ArgoCD scenarios.

To QE: I'd suggest to test regression testing using ReplicaSet as top controllers [1], more deep testing on Argo custom controllers is a little bit out of the scope of Kiali, we can get feedback from the community if there are more issues.

[1] https://kubernetes.io/es/docs/concepts/workloads/controllers/replicaset/#ejemplo

pbajjuri20:
I have validated the Multiple ReplicaS

new_rc


wl_RC
ets, it is listed in Kiali-Workloads without any errors

pbajjuri20 (approved):

Issue #3704, opened 2021-02-15
Improve discovery matcher process for Custom Dashboards


jotak:
This is a follow-up on https://github.com/kiali/kiali/issues/3660

Prometheus v2.24, the latest version, has improved its API [1] in a way that should drastically improve performances of the metrics discovery process for custom dashboards. Instead of calling `api.Series`, we should call `api.LabelValues` for label `__name__` and the labelset to match on (previously, this endpoint didn't accept the labelset to match on as parameter). This endpoint should eliminate the cardinality issue and will return just 1 item per metric family (ie. regardless the number of pods per app), hence should be much more efficient in our case.

Since this solution only works with a recent enough prometheus, it will have to live with the alternative / legacy solution (keeping both in code).

Note that this will be possible only with the next version of prom's go client (today v1.9.0 doesn't have the new endpoint, but a PR was merged to update the API, so next release should be good)

[1] https://prometheus.io/docs/prometheus/2.24/querying/api/#querying-label-values

jotak:
Just though about something so stupidly simple to tackle that, even without the new prom API...
... instead of grabbing all metrics names and check if some dashboards match, we could just extract the list of metrics used for discovery *from the available dashboards* and see if they're in prom.

lucasponce:
> ... instead of grabbing all metrics names and check if some dashboards match, we could just extract the list of metrics used for discovery from the available dashboards and see if they're in prom.

This sounds doable.
Perhaps to check which approach may be more interesting, Istio 1.11 is using Prometheus 2.26, but to avoid to introduce limitations on the dependency, it can use the old API but this workaround to query specific metrics extracted from the custom dashboards.

jmazzitelli:
Just documenting where this `api.Series` call is - it is found here (a partial call stack):

* https://github.com/kiali/kiali/blob/v1.40.0/prometheus/client.go#L294
* https://github.com/kiali/kiali/blob/v1.40.0/business/dashboards.go#L286
* https://github.com/kiali/kiali/blob/v1.40.0/business/dashboards.go#L302

jmazzitelli:
>  instead of grabbing all metrics names and check if some dashboards match, we could just extract the list of metrics used for discovery from the available dashboards and see if they're in prom.

[...]

PR #4367, opened 2021-09-17
custom dashboard discovery - only look for discoverOn metrics

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3704

This is trying to work around the problem where api.Series is being called during dashboard discovery, which could return a huge amount of data.

We want to only look for specific metric names - those metrics listed as "discoverOn" metrics in the dashboards. Those are the only ones we care about.

### DIFF
--- a/prometheus/client.go
+++ b/prometheus/client.go
@@ -38,7 +38,7 @@ type ClientInterface interface {
 	GetNamespaceServicesRequestRates(namespace, ratesInterval string, queryTime time.Time) (model.Vector, error)
 	GetServiceRequestRates(namespace, service, ratesInterval string, queryTime time.Time) (model.Vector, error)
 	GetWorkloadRequestRates(namespace, workload, ratesInterval string, queryTime time.Time) (model.Vector, model.Vector, error)
-	GetMetricsForLabels(labels []string) ([]string, error)
+	GetMetricsForLabels(metricNames []string, labels string) ([]string, error)
 }
 
 // Client for Prometheus API.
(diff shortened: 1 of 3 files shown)


jmazzitelli:
Here is one implementation - it uses a set of "OR" conditions in the prom query. I am documenting this here because I might change this implementation to something different and I want a record of this way of doing it in case I want to revert back at some point in the future (perhaps we might find this is more efficient):
```go
	// you can use "count" here instead of "sum" for possibly even more goodness
	queryString := fmt.Sprintf("sum(%v%v) by (__name__)", metricNames[0], labelQueryString)
	for i := 1; i < len(metricNames); i++ {
		queryString = fmt.Sprintf("%v OR sum(%v%v) by (__name__)", queryString, metricNames[i], labelQueryString)
	}
	results, warnings, err := in.api.Query(in.ctx, queryString, time.Now())
	if warnings != nil && len(warnings) > 0 {
		log.Warningf("GetMetricsForLabels. Prometheus Warnings: [%s]", strings.Join(warnings, ","))
	}
	if err != nil {
		return nil, errors.NewServiceUnavailable(err.Error())
	}

	// We should only get one timeseries for each metric family name. However, just in case
	// we get duplicates, store the metric name in a map and convert to an array to remove duplicates.

	namesMap := make(map[string]bool)
	for _, item := range results.(model.Vector) {
		namesMap[string(item.Metric["__name__"])] = true
	}
	names := make([]string, 0, 5)
	for n := range namesMap {
		names = append(names, n)
	}

	return names, nil
```

jmazzitelli:
This is ready enough for people to review.

prometheus/client.go
```diff
@@ -285,27 +285,39 @@ func (in *Client) GetFlags() (prom_v1.FlagsResult, error) {
...
 	}
 
-	var names []string
-	for _, labelSet := range results {
-		if name, ok := labelSet["__name__"]; ok {
-			names = append(names, string(name))
+	metricsWeAreLookingFor := make(map[string]bool, len(metricNames))
+	for i := 0; i < len(metricNames); i++ {
+		metricsWeAreLookingFor[metricNames[i]] = true
+	}
```
jshaughn, line 309: Fine but maybe nicer as:
```
for _, m := range metricNames {
    metricsWeAreLookingFor[m] = true
```

prometheus/client.go
```diff
@@ -285,27 +285,39 @@ func (in *Client) GetFlags() (prom_v1.FlagsResult, error) {
...
+
+	metricsWeFound := make([]string, 0, 5)
```
jshaughn, line 311: super-minor but you could give metricsWeFound capacity len(metricNames) (as opposed to 5)

jshaughn (approved):
LGTM with a few non-blocking comments

prometheus/client.go
jmazzitelli: The reason I did NOT do that is the majority of those metric names (the vast majority, in fact) will not match. In fact, the number of elements that will be placed inside "metricsWeFound" will usually be 0 or 1 (there are no custom dashboards for the majority of workloads except for "Envoy" for those workloads with sidecars). And for those workloads that DO have custom dashboards, there will usually never be more than 1 or 2 custom dashboards. So even the size of "5" is probably larger than what will be needed. I just put "5" to capture probably every use-case we'd run in to. Even with 5, the vast majority of the time I suspect 4 will be empty.

jmazzitelli:
I built and published a test image based on this PR - if someone wants to test it - use this image: `quay.io/jmazzitelli/kiali:pr4367`

lucasponce:
I see comments on the code, but I don't see screenshots testing the feature.

jmazzitelli:
> but I don't see screenshots testing the feature.

There are no screenshots to show because nothing will look different than how things look today. This just performs a different query that I hope makes things faster. But the UI will look identical to how it looks today.

lucasponce:
>There are no screenshots to show because nothing will look different than how things look today. This just performs a different query that I hope makes things faster. But the UI will look identical to how it looks today.

Have this work being tested against the https://github.com/kiali/demos/tree/master/runtimes-demo ?

jmazzitelli:
> Have this work being tested against the https://github.com/kiali/demos/tree/master/runtimes-demo ?

Built and deployed the server with this PR:
 
commit 0f39e20f68985eadf64abe615cdf8d1f8c0e7dd2 (HEAD -> dashboard-discovery-3704, jmazzitelli/dashboard-discovery-3704)

about-box

graph

Here's all the applications and workloads - you can see all the dashboards are discovered correctly:
w-vertx
w-springboot
w-quarkus
w-nodejs
a-vertx
a-springboot
a-quarkus
a-nodejs

PR #4375, opened 2021-09-20
Pass empty custom headers

nrfox:
Fixes args not being passed to `CreateTransport` correctly as a result of merging #4312.

### DIFF
--- a/util/httputil/http_util.go
+++ b/util/httputil/http_util.go
@@ -52,7 +52,7 @@ func HttpPost(url string, auth *config.Auth, body io.Reader, timeout time.Durati
 		return nil, 0, err
 	}
 
-	transport, err := CreateTransport(auth, &http.Transport{}, timeout)
+	transport, err := CreateTransport(auth, &http.Transport{}, timeout, map[string]string{})
 	if err != nil {
 		return nil, 0, err
 	}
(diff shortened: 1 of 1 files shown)


util/httputil/http_util.go
```diff
@@ -52,7 +52,7 @@ func HttpPost(url string, auth *config.Auth, body io.Reader, timeout time.Durati
 		return nil, 0, err
 	}
 
-	transport, err := CreateTransport(auth, &http.Transport{}, timeout)
+	transport, err := CreateTransport(auth, &http.Transport{}, timeout, map[string]string{})
```
jmazzitelli, line 55: pass in nil - that's how every other place is that I changed that added this new argument.

jmazzitelli (changes requested):

jmazzitelli (approved):

jmazzitelli:
FWIW, might not be a bad idea to support custom headers like HttpGet for future

nrfox:
> FWIW, might not be a bad idea to support custom headers like HttpGet for future

I'll make a separate issue for that. Just prioritizing getting master fixed for now.

pbajjuri20 (approved):

nrfox:
> I have verified with OpenShift, Token, And Anonymous Login Authentication. It works perfectly.

Thanks @pbajjuri20!

Issue #4295, opened 2021-08-20
Mesh internal ServiceEntry should be grouped in app box with workloads


nrfox:
**General remarks**

Service nodes are grouped with their workloads on the "versioned app" graph. But internal ServiceEntries are not despite being a part of the same "app".

**Describe the bug**
In this screenshot, on the left is a Service without a ServiceEntry that is properly boxed with the workloads. On the right is a Service with a ServiceEntry that is excluded from the box with the workloads.

![Screenshot from 2021-08-20 15-55-46](https://user-images.githubusercontent.com/6226732/130287146-c4841169-c8da-4964-823b-decd47dfaf66.png)

**Versions used**
Kiali: 1.39
Istio: 1.11
Kubernetes flavour and version: upstream 1.21

**To Reproduce**
1. Deploy bookinfo
2. Create a ServiceEntry using the same selectors as one of the existing bookinfo app's Service.
3. On the versioned app graph, observe that the ServiceEntry does not get boxed with the app workloads.

**Expected behavior**
Expected internal ServiceEntry to be boxed with the app workloads that it corresponds to on the versioned app graph.

jshaughn:
Hmm, that's interesting.  I guess the "se-aggregate" node must not be getting the "app" assigned.  Is it always true that the aggregated "se-service" nodes have the same "app" value, or is it possible to have "se-service" nodes with different apps, getting aggregated together?   If we safely aggregate, and assign the "app" value to the new serviceEntry node, it should get boxed.  Adding to the backlog.

PR #4376, opened 2021-09-20
Fix ServiceEntry not grouped in app box

israel-hdez:
Istio appender was skipping further adding missing labels of ServiceEntries. Probably, it made sense at that moment that code was introduced. However, nowdays Istio doc tells that SE resources can be used to "decorate" services from other registries. See https://istio.io/v1.10/docs/reference/config/networking/service-entry/#ServiceEntry.

So, we no longer can simply skip adding missing labels for SEs. We should check if a SE is augmenting a regular service and add missing labels if needed.

Fixes kiali/kiali#4295.

### DIFF
--- a/graph/telemetry/istio/appender/istio_details.go
+++ b/graph/telemetry/istio/appender/istio_details.go
@@ -1,6 +1,7 @@
 package appender
 
 import (
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/labels"
@@ -168,7 +169,16 @@ func addLabels(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo
 			if _, ok := n.Metadata[graph.IsServiceEntry]; ok {
 				seInfo := n.Metadata[graph.IsServiceEntry].(*graph.SEInfo)
 				for _, host := range seInfo.Hosts {
-					if svc, found := svcMap[host]; found {
+					var hostToTest string
+
+					hostSplitted := strings.Split(host, ".")
+					if len(hostSplitted) == 3 && hostSplitted[2] == config.IstioMultiClusterHostSuffix {
+						hostToTest = host
+					} else {
+						hostToTest = hostSplitted[0]
+					}
+
+					if svc, found := svcMap[hostToTest]; found {
 						if app, ok := svc.Labels[appLabelName]; ok {
 							n.App = app
 						}
--- a/graph/telemetry/istio/appender/istio_details.go
+++ b/graph/telemetry/istio/appender/istio_details.go
@@ -163,8 +163,18 @@ func addLabels(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo
 	for _, n := range trafficMap {
 		// make sure service nodes have the defined app label so it can be used for app grouping in the UI.
 		if n.NodeType == graph.NodeTypeService && n.Namespace == sdl.Namespace.Name && n.App == "" {
-			// A service node that is a service entry will not have a service definition
+			// For service nodes that are a service entries, use the `hosts` property of the SE to find
+			// a matching Kubernetes Svc for adding missing labels
 			if _, ok := n.Metadata[graph.IsServiceEntry]; ok {
+				seInfo := n.Metadata[graph.IsServiceEntry].(*graph.SEInfo)
+				for _, host := range seInfo.Hosts {
+					if svc, found := svcMap[host]; found {
+						if app, ok := svc.Labels[appLabelName]; ok {
+							n.App = app
+						}
+						continue
+					}
+				}
 				continue
 			}
 			// A service node that is an Istio egress cluster will not have a service definition
--- a/graph/telemetry/istio/appender/istio_details_test.go
+++ b/graph/telemetry/istio/appender/istio_details_test.go
@@ -300,3 +300,41 @@ func TestVSWithRoutingBadges(t *testing.T) {
 
 	assert.Equal(true, trafficMap[fooSvcNodeId].Metadata[graph.HasTrafficShifting])
 }
+
+func TestSEInAppBox(t *testing.T) {
+	check := assert.New(t)
+	config.Set(config.NewConfig())
+
+	k8s := kubetest.NewK8SClientMock()
+	k8s.On("GetProject", mock.AnythingOfType("string")).Return(&osproject_v1.Project{}, nil)
+	k8s.On("GetIstioObjects", mock.AnythingOfType("string"), "destinationrules", "").Return([]kubernetes.IstioObject{}, nil)
+	k8s.On("GetIstioObjects", mock.AnythingOfType("string"), "virtualservices", "").Return([]kubernetes.IstioObject{}, nil)
+	k8s.On("GetServices", mock.AnythingOfType("string"), mock.Anything).Return([]core_v1.Service{{
+		ObjectMeta: meta_v1.ObjectMeta{
+			Name: "foobar",
+			Labels: map[string]string{
+				"app": "fooApp",
+			},
+		},
+	}}, nil)
+
+	businessLayer := business.NewWithBackends(k8s, nil, nil)
+
+	trafficMap := graph.NewTrafficMap()
+	serviceEntryNode := graph.NewNode(business.DefaultClusterID, "testNamespace", "ratings", "", "", "", "", graph.GraphTypeVersionedApp)
+	serviceEntryNode.Metadata[graph.IsServiceEntry] = &graph.SEInfo{
+		Hosts:     []string{"foobar.com"},
+		Location:  "MESH_INTERNAL",
+		Namespace: "testNamespace",
+	}
+	trafficMap[serviceEntryNode.ID] = &serviceEntryNode
+
+	globalInfo := graph.NewAppenderGlobalInfo()
+	globalInfo.Business = businessLayer
+	namespaceInfo := graph.NewAppenderNamespaceInfo("testNamespace")
+
+	a := IstioAppender{}
+	a.AppendGraph(trafficMap, globalInfo, namespaceInfo)
+
+	check.Equal("fooApp", trafficMap[serviceEntryNode.ID].App)
+}
--- a/graph/telemetry/istio/appender/service_entry.go
+++ b/graph/telemetry/istio/appender/service_entry.go
@@ -81,7 +81,7 @@ func (a ServiceEntryAppender) applyServiceEntries(trafficMap graph.TrafficMap, g
 		}
 
 		// To match, a service entry must be exported to the source namespace, and the requested
-		// service must match a defined host.  Note that teh source namespace is assumed to be the
+		// service must match a defined host.  Note that the source namespace is assumed to be the
 		// the same as the appender namespace as all requests for the service entry should be coming
 		// from workloads in the current namespace being processed for the graph.
 		if se, ok := a.getServiceEntry(namespaceInfo.Namespace, n.Service, globalInfo); ok {


israel-hdez:
@nrfox @jshaughn Apparently, this fixes the issue. Although I'm not sure if this is the right way.

israel-hdez:
@jshaughn As you mentioned (https://github.com/kiali/kiali/issues/4295#issuecomment-902941588)[here] in a comment in the associated issue, I wonder what should happen if the same SA has several `hosts` that match several k8s services? :-/

graph/telemetry/istio/appender/istio_details.go
```diff
@@ -163,8 +163,18 @@ func addLabels(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo
...
 			if _, ok := n.Metadata[graph.IsServiceEntry]; ok {
+				seInfo := n.Metadata[graph.IsServiceEntry].(*graph.SEInfo)
+				for _, host := range seInfo.Hosts {
+					if svc, found := svcMap[host]; found {
+						if app, ok := svc.Labels[appLabelName]; ok {
+							n.App = app
+						}
+						continue
+					}
+				}
```
jshaughn, line 177: @israel-hdez (cc @nrfox)  I was originally thinking maybe the logic should be in the service_entry appender, when we create the service_entry node, but I guess this is an appropriate place as this is where we add the label for other service nodes.

The question I have is the same as what I asked in the issue, _Is it always true that the aggregated "se-service" nodes have the same "app" value, or is it possible to have "se-service" nodes with different apps, getting aggregated together?_.  Or, in other words, can a ServiceEntry node have multiple apps.  I don't really know the answer.  If we want to be very safe we could make sure only to make the assignment if there is one and only one app value.  But I'm also OK leaving this as-is.  I'll leave it up to you.

jshaughn (commented):
@israel-hdez See my comment.  You need to rebase to resolve the failure that was fixed in master.

graph/telemetry/istio/appender/istio_details.go
nrfox: You're asking if it's possible to have a ServiceEntry that looks something like:
```
kind: ServiceEntry
metadata:
  name: details-svc
spec:
  hosts:
  - details.bookinfo.com
  - reviews.bookinfo.com
  location: MESH_INTERNAL
  ports:
  - number: 80
    name: http
    protocol: HTTP
  resolution: STATIC
  workloadSelector:
    labels:
      example: bookinfo
```
where one Service Entry has two hosts each pointing to a separate app? And how Kiali should aggregate the telemetry nodes in this case?

jshaughn: I'm not necessarily questioning what we currently do to create the serviceEntry nodes.  But remember that a serviceEntry node can be an aggregate of multiple service nodes.  Because this PR is for assigning a single 'app' to the Se node, my question was whether there is actually at most one 'app' value to assign, and if not, how to handle.  So yes, in your example above, would you assign `details` or `reviews`.

nrfox: It seems like it's possible so maybe there is a valid use case out there for it.

If I'm understanding the change correctly, kiali is assuming that if a svc node without an app label has a service entry with a host that matches the name of a kube `Service`, then kiali sets the app label of the svc node to be the app label of the kube `Service` since the SE is "decorating" the kube `Service`?

What if there is no kube `Service` only the SE svc node? If the SE object has an "app: ..." label that matches the workload nodes, would it still get boxed with the workload nodes?

israel-hdez: > What if there is no kube `Service` only the SE svc node? If the SE object has an "app: ..." label that matches the workload nodes, would it still get boxed with the workload nodes?

It won't get boxed.
Anyway, yesterday I tried this case and I wasn't able to make the traffic to be routed properly. The bookinfo app always replied with something like "ratings is unavailable".

After repeatedly reading the docs ,what I understand is that, for the most part, Istio only deals with IPs. So, if you are trying to contact "reviews", you still need something to let your app to resolve "reviews" to an IP. Without the k8s svc, you will always get a "host not found" error. So, you need always a serviceentry AND a k8s service (or something else) to be able to resolve the DNS name.

nrfox: > Without the k8s svc, you will always get a "host not found" error.

When running inside of kube, you need to enable [dns proxying](https://istio.io/latest/docs/ops/configuration/traffic-management/dns-proxy/) first so that the proxy itself resolves the IP rather than kube dns. This is enabled by default on VM proxies.

israel-hdez: > When running inside of kube, you need to enable [dns proxying](https://istio.io/latest/docs/ops/configuration/traffic-management/dns-proxy/) first so that the proxy itself resolves the IP rather than kube dns. This is enabled by default on VM proxies.

I wasn't aware of it... Wow... apparently you sit and deeply studied the docs :-D

israel-hdez: > It would be nice for standalone (does not decorate an existing kube `Service`) SE nodes with an `app: ...` label to be boxed with their corresponding workload nodes.

Thinking a little bit, I think this is a completely different case IMO. I wasn't aware of the dns proxying feature, so I was unable to make it work. But, first, we would need to check what's the telemetry in that case. I'm not sure if Istio will record the name of the SE in `destination_service` labels. If not, currently, the graph may generate a workload-to-workload topology. If this is the case, we would need to inject the SE nodes and it's not that trivial to do, I think.

nrfox: > I'm not sure if Istio will record the name of the SE in `destination_service` labels. If not, currently, the graph may generate a workload-to-workload topology. If this is the case, we would need to inject the SE nodes and it's not that trivial to do, I think.

That's true it'll probably require some investigation and based on that it could greatly change the work required to implement. Perhaps that can be a separate issue from this one then and we can merge this as is? iirc from the bookinfo-vm demo, the service entry node does show up on the graph but it'd good be to verify that and get a better sense of what the telemetry is.

israel-hdez: > Perhaps that can be a separate issue from this one then and we can merge this as is?

+1 At least for the scope of the associated bug, I think this is good to go.
I can further investigate under another issue.

nrfox (dismissed):
Looks good for fixing SE nodes that "decorate" existing kube `Services`. Boxing SE nodes without kube services can be a separate issue as discussed in [this](https://github.com/kiali/kiali/pull/4376/files#r713059079) comment thread.

jshaughn (dismissed):
@israel-hdez @nrfox Good discussion.  I'm fine with this as-is.  If you'd like to create that associated issue it's OK by me although I would likely suggest it be set to Future.  I think we can let the community push this further if they run into an issue.

One last comment, if we ever do need to revisit this, we could consider a serviceEntry node per app, if it were to make any sense :-)

pbajjuri20:
Before PR:

before_se_pr


After PR:

after_se_pr

pbajjuri20:
@israel-hdez  @jshaughn : 

When we applying this SE.yaml file

```
kind: ServiceEntry
metadata:
  name: ratings-svc
spec:
  hosts:
  - ratings.bookinfo.com
  location: MESH_INTERNAL
  ports:
  - number: 80
    name: http
    protocol: HTTP
  resolution: STATIC
  workloadSelector:
    labels:
      example: bookinfo


```

host = ratings, It is populating the graph expected result.

host = ratings.bookinfo.com   --> it is not populating the graph as expected,

before_se_pr

pbajjuri20 (dismissed):

israel-hdez:
@jshaughn The second case that Praneeth is commenting, I'm not sure if it's right at all that Kiali is replacing the svc triangle with a SE pentagon. Maybe it is an issue? I'm not sure.

nrfox:
> @jshaughn The second case that Praneeth is commenting, I'm not sure if it's right at all that Kiali is replacing the svc triangle with a SE pentagon. Maybe it is an issue? I'm not sure.

I had this thought when working on WorkloadEntry badging. Workloads can be _both_ "workload entries" and "deployments". I guess the same is true of (at least mesh internal) Service Entries. A mesh's "Service" can be both a kube Service and a Service Entry. When it's both which icon does kiali pick?

jshaughn:
> The second case that Praneeth is commenting, I'm not sure if it's right at all that Kiali is replacing the svc triangle with a SE pentagon.

OK, so first off I'm confused why @pbajjuri20  approved this after showing an unexpected result, perhaps I missed some off-pr discussion.   The test case is a little bit weird, maybe, because I would have expected a MESH_INTERNAL serveice entry to be tied to VMs, but I don't see any VM badging.  But that may just be because the actual test didn't need that whole setup.  @nrfox , in general a MESH_INTERNAL SE is used with a WorkloadEntry, right?  

Anyway, @israel-hdez , I don't know if it's a problem with SE identification, but the reason we see the above is because for MESH_INTERNAL SE matching, we perform matching on the hostname portion of the fqdn (`strings.Split(hostFqdn, ".")[0]`), but we do an exact match in the PR when deciding to assign `app`.  We may need to perform the same matching to be consistent.   By the way, I don't really know why we do what we do in the SE matching, but it has been like that for a long while, I think.

israel-hdez:
> We may need to perform the same matching to be consistent. By the way, I don't really know why we do what we do in the SE matching, but it has been like that for a long while, I think.

OK, I did the changes to match SEs and SVCs using the same criteria as in the service_entry appender. IMO, it's buggy. Although community may be happy with it.

> The test case is a little bit weird, maybe, because I would have expected a MESH_INTERNAL serveice entry to be tied to VMs, but I don't see any VM badging. But that may just be because the actual test didn't need that whole setup. @nrfox , in general a MESH_INTERNAL SE is used with a WorkloadEntry, right?

A SE is a container, I think. AFAIU, it can make reference both to pods or services running in VMs to give a migration path from VMs to k8s (or viceversa).

jshaughn:
> IMO, it's buggy. 

I think you're right but I don't know what the bugs are.  I've been waiting for the community to let us know.

jshaughn (dismissed):
It looks ok to me, I'll defer to @nrfox and @pbajjuri20 .

pbajjuri20:
I have deployed and tested the ServiceEntry.yaml

hosts: ratings.bookinfo.com   --> this hostname bug is fixed perfectly. 


PR_1




hosts: ratings

PR_2

nrfox:
> @nrfox , in general a MESH_INTERNAL SE is used with a WorkloadEntry, right?

Yes, although you could use a mesh internal SE without WorkloadEntries or with a mixture of both workload entries and pods. This change won't box a service entry that doesn't have an existing/matching kube service, for example an app that was VMs only, but @israel-hdez and I thought that could be addressed as a separate issue.

> A SE is a container, I think. AFAIU, it can make reference both to pods or services running in VMs to give a migration path from VMs to k8s (or viceversa).

Yeah that's my understanding as well. ServiceEntries can "select" workloads that are some mixture of workload entries and pods _or_ just workload entries _or_ just pods. When the hostnames of the SE match a kube service it counts as a "decorator" for that service, which seems to mean that if there's additional properties on the SE then they are added to the kube service in the proxy config.

nrfox (commented):
Can you also add a test for this scenario? Otherwise +1

israel-hdez:
@nrfox I added a small test focusing only on the changed appender.
I hope this is enough.

nrfox (approved):
:+1:

israel-hdez:
@pbajjuri20 The last commit only added a test. Nothing changed for QA. You can approve when you are ready.

Issue #4303, opened 2021-08-26
Kiali Istio dashboards incompatible with thanos-query


vbergbauer:
**General remarks**

To reduce the load on the Prometheus server, we are directing Grafana and Kiali to a Thanos querier.
This works fine with Grafana, and mostly works with Kiali, as seen on:
- the graph:
  ![image](https://user-images.githubusercontent.com/8659572/131010698-e775da36-8443-42ab-9b49-22f96a9e907e.png)
- this custom dashboard:
   ![image](https://user-images.githubusercontent.com/8659572/131009951-6090cf6e-3d35-479e-bad3-480763604e0c.png)

**Bug**
The Istio dashboards do not show any data:
![image](https://user-images.githubusercontent.com/8659572/131010274-64552837-5075-4746-8414-4951adfab86d.png)

Looking at the Kiali logs, I see the following error:
```bash
kubectl kubectl -n observability logs -lapp=kiali | grep ' ERR '
2021-08-26T17:24:52Z ERR Failed to fetch Prometheus configuration: client_error: client error: 404
```
Indeed, the Thanos querier does not implement Prometheus's API `api/v1/status/config`.

However this **should not prevent Kiali from calling the query APIs** (as shown in the successful Istio Kiali graph and custom dashboards). Even if Kiali picks up information from the Prometheus config (like the default refresh), it should be able to fall back to some safe default value.

**Workaround**
I can use an Istio virtual service to route thanos-query:9090/api/v1/status/config to Prometheus (or to a static file server)
```bash
# Create Virtual Service to route /api/v1/status/config to Prometheus
cat <<EOF | kubectl apply -f -
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: thanos-query
  namespace: observability
spec:
  hosts:
  - thanos-query
  http:
  - match:
    - uri:
        exact: /api/v1/status/config
    route:
    - destination:
        port:
          number: 9090
        host: prometheus-prometheus
  - route:
    - destination:
        port:
          number: 9090
        host: thanos-query
EOF

# Inject the Istio sidecar into the Kiali deployment
kubectl -n observability patch deploy kiali --patch '{"spec":{"template":{"metadata":{"labels":{"sidecar.istio.io/inject": "true"}}}}}'
```

Then the dashboard works correctly:
![image](https://user-images.githubusercontent.com/8659572/131011569-8f9b6463-e6c1-4ce6-a54c-563c5916348e.png)

**Versions used**
[...]

PR #4386, opened 2021-09-22
Kiali Istio dashboards incompatible with thanos-query

aljesusg:
Fix https://github.com/kiali/kiali/issues/4303

PR operator : https://github.com/kiali/kiali-operator/pull/427

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -125,7 +125,7 @@ type PrometheusConfig struct {
 	CustomHeaders   map[string]string `yaml:"custom_headers,omitempty"`
 	HealthCheckUrl  string            `yaml:"health_check_url,omitempty"`
 	IsCore          bool              `yaml:"is_core,omitempty"`
-	ThanosProxy     Thanos            `yaml:"thanos_proxy,omitempty"`
+	ThanosProxy     ThanosProxy       `yaml:"thanos_proxy,omitempty"`
 	URL             string            `yaml:"url,omitempty"`
 }
 
@@ -476,6 +476,11 @@ func NewConfig() (c *Config) {
 				Enabled:                true,
 				IsCore:                 false,
 				NamespaceLabel:         "kubernetes_namespace",
+				Prometheus: PrometheusConfig{
+					ThanosProxy: ThanosProxy{
+						Enabled: false
+					}
+				}
 			},
 			Grafana: GrafanaConfig{
 				Auth: Auth{
@@ -519,7 +524,7 @@ func NewConfig() (c *Config) {
 				Auth: Auth{
 					Type: AuthTypeNone,
 				},
-				ThanosProxy: Thanos{
+				ThanosProxy: ThanosProxy{
 					Enabled: false,
 				},
 				CacheEnabled: true,
--- a/config/config.go
+++ b/config/config.go
@@ -525,7 +525,9 @@ func NewConfig() (c *Config) {
 					Type: AuthTypeNone,
 				},
 				ThanosProxy: ThanosProxy{
-					Enabled: false,
+					Enabled:         false,
+					RetentionPeriod: "7d",
+					ScrapeInterval:  "30s",
 				},
 				CacheEnabled: true,
 				// 1/2 Prom Scrape Interval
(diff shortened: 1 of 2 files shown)


jmazzitelli (changes requested):
comment from the issue:
> I'm going to create the PR with these proposed changes but I think as far as I know that this is not going to fix it.

If we don't think this is going to do anything, then we should not merge this. If it doesn't fix the problem, we should not add this "just to add it". We should look into things further and fix the issue and not add code that doesn't address the problem.

jmazzitelli:
I would set this PR to "draft" status, build the server with these changes, and ask for the person who reported the issue to test with that build and see if it does, indeed, address the issue.

aljesusg:
Totally agree, I am trying to test this in GCp but IAM is playing with me and I can’t write in bucket, but as you said maybe @vbergbauer can test it

vbergbauer:
@aljesusg : If you build a container image for me, I'll happily test it.
I know how to reproduce the problem on various clusters (KinD, GKE, OCP).

aljesusg:
Thanks @vbergbauer I pushed the image in quay.io/aljesusg/kiali:dev ...I am checking and I think it solves it

config/config.go
```diff
@@ -106,6 +106,13 @@ func (a *Auth) Obfuscate() {
 	a.CAFile = "xxx"
 }
 
+// Thanos describes configuration of the Thanos proxy component
+type Thanos struct {
```
jmazzitelli, line 110: shouldn't we call this ThanosProxy just to remain consistent with the json name?

jmazzitelli:
I'm not sure if anything else has to change, but I will bring this up ... there is another place where prometheus config is in our configuration (this is Kiali CR's `spec.external_services.custom_dashboards`):

```
// CustomDashboardsConfig describes configuration specific to Custom Dashboards
type CustomDashboardsConfig struct {
	DiscoveryEnabled       string           `yaml:"discovery_enabled,omitempty"`
	DiscoveryAutoThreshold int              `yaml:"discovery_auto_threshold,omitempty"`
	Enabled                bool             `yaml:"enabled,omitempty"`
	IsCore                 bool             `yaml:"is_core,omitempty"`
	NamespaceLabel         string           `yaml:"namespace_label,omitempty"`
	Prometheus             PrometheusConfig `yaml:"prometheus,omitempty"`
}
```

Notice there is a PrometheusConfig here also (it defines the secondary Prometheus that Kiali can query for the custom dashboard metrics.) This means there can also be a ThanosProxy defined here for the secondary Prometheus. It is probably a no-op (i.e. if a user specifies a ThanosProxy here inside this `spec.external_services.custom_dashboards.prometheus.thanos_proxy` it will have no effect and be ignored).

I want to mention this in case we actually do need to do something here with ThanosProxy config when accessing Prometheus for custom dashboard metrics.

aljesusg:
Yes If this work correctly I'll do the changes

config/config.go
```diff
@@ -468,6 +476,11 @@ func NewConfig() (c *Config) {
 				Enabled:                true,
 				IsCore:                 false,
 				NamespaceLabel:         "kubernetes_namespace",
+				Prometheus: PrometheusConfig{
+					ThanosProxy: ThanosProxy{
+						Enabled: false
+					}
+				}
```
jmazzitelli, line 483: This should fix the linter error that is causing the build to fail:
```suggestion
						Enabled: false,
					},
				},
```

jmazzitelli (changes requested):

jmazzitelli:
If this is good and will be merged, we'll need an operator PR.

* Add to example CR: https://github.com/kiali/kiali-operator/blob/v1.40.0/deploy/kiali/kiali_cr.yaml#L619
* Add defaults: https://github.com/kiali/kiali-operator/blob/v1.40.0/roles/default/kiali-deploy/defaults/main.yml#L178

vbergbauer:
Hi @aljesusg 
Sorry for the delay (busy week last week).
I tested quay.io/aljesusg/kiali:dev and it does resolve the issue with custom dashboards without resorting to my workaround.
Thank you.

aljesusg:
Great we can continue @jmazzitelli

jshaughn:
@aljesusg Same comment as in the sibling PR.

Issue #1372, opened 2019-08-23
Add strong type mapping in Istio Kiali model


lucasponce:
Kiali model maps Kubernetes and Istio objects and enrich them with validations.
As Istio objects are implemented as CRDs, they have a limitation that the API doesn't offer a native strong type.

Kiali maps a first level but it should complete with all the fields:
https://github.com/kiali/kiali/blob/master/models/virtual_service.go#L29

So, today, the real mapping is defined in the UI, and I think it should be defined from the backend itself.

JIRA:
https://issues.jboss.org/browse/KIALI-2118

lucasponce:
I think this is still valid.
We have even a gap, because in the UI we have a deeper mapping between Istio entities model than in the backend.

But it's just a pure refactoring task.

nrfox:
This came up recently in another PR. Could Kiali's backend simply import and use https://pkg.go.dev/istio.io/client-go?

lucasponce:
I think we need to check pros/cons as it's discussed on this 
https://github.com/kubernetes/client-go/tree/master/examples/dynamic-create-update-delete-deployment

I'm mostly +1 on this, as when we started these libraries were not produced and we used the dynamic strategy everywhere (and it works).

Having a strong type mapping will help, but also it may introduce other issues like that a version is tight close to specific Istio versions, but that could be ok (i.e. probably we'll miss that Kiali x.y.z can support several Istio a.b.c if those have minimal changes in the APIs).

Overall I'm in favor of this, but this is going to trigger a huge refactoring in the backend so this one was on hold to wait to a major version of Kiali to proceed.

lucasponce:
I'm thinking that this one would be good to add it for next Sprint.
Basically the scope would be:
- To use the istio.io/client-go dependency for using Istio types.
- Update generic IstioObject wrappers with proper types.
- Propagate the changes into cache, validations and business package logic (most of the work would be to get rid of the conversion of generic interface{} into proper structs).
- Potentially, patch logic can keep working as usual.
- As Kiali has a proper model that interacts with the UI, the kiali backend API shouldn't be impacted, others logic (like appenders) that use business package won't be affected.

In terms of logic, it should be transparent to the features.

lucasponce:
Updating this one, as it seems that the istio.io/client-go and the jaegertracing/jaeger project have some conflicts related the use of grpc for their models.

I've created both issues in jaeger and istio projects:
https://github.com/jaegertracing/jaeger/issues/3276
[...]

PR #4388, opened 2021-09-22
Trim jaeger project dependencies

lucasponce:
Required for https://github.com/kiali/kiali/issues/1372

Kiali queries Jaeger backend using two ways: using the legacy HTTP REST api and the gRPC api_v2.
gRPC protocol is the preferred/default one used in upstream releases but we keep the HTTP legacy one as the gRPC protocol reported problems when used in combination of sidecar containers like OAuth.

As part of a technical debt, Kiali was pulling the whole jaeger project as a dependency in the go.mod:
image

The main goal of this PR is to remove that dependency and use gRPC using the .proto files defined in Jaeger but without importing any dependency (as it's recommended by Jaeger project).

This is necessary to pull other gRPC dependencies, this work is part of the epic to incorporate istio.io/client-go model into Kiali.

So, in this work you will see:
- Remove of the jaeger dependencies.
- Bring the {model,query}.proto files.
- Update the {model,query}.proto files to remove the "gogo" dependencies to have a clean protobuffer model.
- Bring and clone the minimal dependencies of the jaeger/model to manipulate TraceID and SpanID.
- Bring and clone the minimal dependencies of the jaeger/model to keep the legacy HTTP model and the JSON converted used in the UI layer.

It is being discussed with jaeger that some of the jaeger/model files should be moved to the jaeger-lib to use that dependency with no conflicts, but it's not yet there so cloning a set of minimal classes look reasonable rather than pull the whole project (more info in the main discussion https://github.com/kiali/kiali/issues/1372#issuecomment-923089928 with links to specific issues in jaeger and istio).

In short, with this refactoring, the expectation is that everything works just fine and dependencies are clean as previous step to bring other gRPCs clients.

For QE/Testing:
- Test Kiali with an example like travel-agency that have traces.
- Use the HTTP protocol (point to your backend, but as an example):
image
- Use the gRPC protocol (point to your backend, but as an example):
image

Check that there is no regression in tracing features (graph, trace tabs, correlation with metrics and logs, trace details).
Both HTTP and gRPC works well.


leandroberetta (approved):
LGTM, tested locally with travels app using grpc.

prachiyadav (approved):
Regression Testing is done

lucasponce:
Thanks ! 
There is another PR that depends on this, so I'll consolidate it.

Issue #2911, opened 2020-06-17
Add custom label aggregation in metrics tab


jotak:
Similar to https://github.com/kiali/kiali/issues/2854 , but for metrics pages.
These would just be new labels to aggregate on in metrics pages.

Actually, we could allow to configure any list of custom labels to aggregate on, in addition to the pre-configured static ones. In terms of implementation, much of the mechanisms are already in place, we would just have to add a little bit of mapping between config and k-charted dashboard generation.

jshaughn:
> Actually, we could allow to configure any list of custom labels to aggregate on

Yes, I think this will be a really nice feature for the metric charts and I think can be added independently of the graph work.

lucasponce:
As it's commented on https://github.com/kiali/kiali/issues/2911#issue-640199767 the labels used by aggregation in istio metrics tabs are static and defined in code:

```
func buildIstioAggregations(local, remote string) []Aggregation {
	aggs := []Aggregation{
		{
			Label:       fmt.Sprintf("%s_canonical_revision", local),
			DisplayName: "Local version",
		},
		{
			Label:       fmt.Sprintf("%s_workload_namespace", remote),
			DisplayName: "Remote namespace",
		},
	}
	if remote == "destination" {
		aggs = append(aggs, Aggregation{
			Label:       "destination_service_name",
			DisplayName: "Remote service",
		})
	}
	aggs = append(aggs, []Aggregation{
		{
			Label:       fmt.Sprintf("%s_canonical_service", remote),
			DisplayName: "Remote app",
		},
		{
			Label:       fmt.Sprintf("%s_canonical_revision", remote),
			DisplayName: "Remote version",
		},
		{
			Label:       "response_code",
			DisplayName: "Response code",
		},
		{
			Label:       "grpc_response_status",
			DisplayName: "GRPC status",
		},
		{
			Label:       "response_flags",
			DisplayName: "Response flags",
		},
	}...)
	return aggs
}
```

These labels are good, also the metrics tab is an opinionated dashboard and it's not expected that user can personalize it (other tools as grafana are better for that task).

But I agree that these static labels can be extended, so, this issue could be used to add a new section in the configuration to support additional label aggregations. 

Then user may filter some metrics by an additional criteria (i.e. "istio_io_rev" to get the control plane when user is performing some canary A/B upgrade or similar).

So the goal of this task could be:
- Add a new section in the config (+operator) to add this new labels.
- Update the static section to read the aggregated labels defined in the config.
[...]

PR #4390, opened 2021-09-23
Configurable label aggregations in metric tabs

israel-hdez:
Adding possibility to configure _additional_ label aggregations in metric tabs.

For example, a Kiali CR with the following configuration:

```yaml
spec:
  kiali_feature_flags:
    ui_defaults:
      metrics_inbound:
        aggregations:
        - display_name: Istio Network
          label: topology_istio_io_network
        - display_name: Istio Revision
          label: istio_io_rev

```

will generate the following options in the metrics menu:

image

Fixes kiali/kiali#2911.

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -501,14 +501,14 @@ func GetIstioScaler() func(name string) float64 {
 	}
 }
 
-// GetIstioDashboard returns Istio dashboard (currently hard-coded) filled-in with metrics
+// BuildIstioDashboard returns Istio dashboard filled-in with metrics
 func (in *DashboardsService) BuildIstioDashboard(metrics models.MetricsMap, direction string) *models.MonitoringDashboard {
 	var dashboard models.MonitoringDashboard
 	// Copy dashboard
 	if direction == "inbound" {
-		dashboard = models.PrepareIstioDashboard("Inbound", "destination", "source")
+		dashboard = models.PrepareIstioDashboard("Inbound")
 	} else {
-		dashboard = models.PrepareIstioDashboard("Outbound", "source", "destination")
+		dashboard = models.PrepareIstioDashboard("Outbound")
 	}
 
 	istioCharts := getIstioCharts()
--- a/business/istio_certs_test.go
+++ b/business/istio_certs_test.go
@@ -3,14 +3,15 @@ package business
 import (
 	"testing"
 
-	"github.com/kiali/kiali/config"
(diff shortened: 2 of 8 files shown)


israel-hdez:
@lucasponce @jshaughn I'm wondering if it's acceptable to provide defaults and let the user override, if needed?

In the original issue, it is suggested to only allow adding custom aggregations. But I wonder why negate removing aggregation that you may not want? These few changes are allowing full override, thus allowing removing any aggregation you may not need.

If we are OK by allowing overriding defaults, I need to do more changes (both to the front- and back-end). So, I want to know your thought before I continue.

lucasponce:
Hi @israel-hdez, thanks for the PR.

I put a comment in the issue when we revisited the scope of this work https://github.com/kiali/kiali/issues/2911#issuecomment-893410726 . I don't think we want to increase the level of personalization of the Kiali dashboards, Grafana does a better job for that scenario. The idea if that Kiali offers a default opionanted dashboards that we think that they help to the Istio domain with certain level of personalization (custom dashboards) but this would be more a corner case rather than a first class feature.

So any minor step that sounds interesting, like this one, to allow some personalization through configuration, looks ok, but I wouldn't go beyond as that may potentially require a re-design of the current model.

jshaughn:
My understanding is that this issue was just to add custom aggregations and not to remove or override the existing aggegations.  I think that just the addition of non-default aggregations is sufficient scope.  Anything further could be considered later, if asked for by the community.

israel-hdez:
OK, I did the changes to allow only _adding_ custom label aggregations.

This is ready for review.

israel-hdez:
Operator PR: https://github.com/kiali/kiali-operator/pull/428

pbajjuri20:
I have deployed and tested on my environment.


cr_pr


Input_q

lucasponce:
Sorry @israel-hdez I missed the review.
I'll fetch it locally today.

lucasponce (approved):
LGTM

pbajjuri20 (commented):
LGTM!

pbajjuri20:
Tested and Verified on the Kiali UI 1.41.0 (51b548beaf7ee8e2ddb501f99b1c8f82836e9910) and Kiali Server v1.41.0-SNAPSHOT (396b11384343a71083f8d579567aee89e2e76307)

israel-hdez:
Thanks for the reviews!

Issue #4326, opened 2021-09-03
Show gateway in istio config


zackzhangkai:
Create a gateway in *Request Routing* - *advanced options*. Then it will create a Virtualservice resource , a DestinationRule resource and a Gateway resource.
In the *istio config*, it just shows *virtualservice* and *destinationrule*.

I think *gateway* resource should also be displayed.

![image](https://user-images.githubusercontent.com/20178386/131968160-82c75033-9c62-400b-a4e1-eb4e78836843.png)

lucasponce:
Thanks @zackzhangkai for the feedback.
Yes, I agree, it's not linked due the VirtualService and DestinationRule are linked only with the direct services.
But I think it could be a nice enhancement to add under the service others related Gateways that are linked with the VirtualService.

I'll put it in the backlog.

hhovsepy:
This epic requires two subtasks to be implemented:

- [ ] Add Gateways in Istio Config list of Service details
- [ ] Add Gateways in Details column of Service List. Add it's support in Filters.

PR #4389, opened 2021-09-23
WIP refactor kubernetes/istio.go to use istio.io/client-go dep

lucasponce:
Part of https://github.com/kiali/kiali/issues/1372 effort

This is WIP and experimental, just using this draft to scope the changes and implications of the refactoring.

Note, it's expected this PR will mutate.

PR #4400, opened 2021-09-30
Gateways linked in VirtualServices are listed in Service/App Lists 'Details' columns.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4326

When VirtualService have linked Gateways, then these Gateways are listed in Details column of ServiceLists and AppLists pagea.
Also Services they can be filtered by Gateways.
Gateways can be linked for instance while creating Routing via Wizard.

Screenshot from 2021-09-30 20-22-36

Screenshot from 2021-10-04 12-17-15

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -126,9 +126,10 @@ func (in *AppService) GetAppList(namespace string, linkIstioResources bool) (mod
 			if linkIstioResources {
 				svcVirtualServices := kubernetes.FilterVirtualServices(*linkedResources[kubernetes.VirtualServices], srv.Namespace, srv.Name)
 				svcDestinationRules := kubernetes.FilterDestinationRules(*linkedResources[kubernetes.DestinationRules], srv.Namespace, srv.Name)
-				allFiltered := append(svcVirtualServices, svcDestinationRules...)
+				svcGateways := kubernetes.FilterGateways(*linkedResources[kubernetes.Gateways], svcVirtualServices)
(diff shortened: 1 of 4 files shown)


lucasponce (changes requested):
I think is missing also in the app view.
The app view links the istio references to the services and workloads that belongs to an app.

So, the service list looks correct:
image

But the app list is missing the gateway:
image

Not a big change, just to apply same logic.

But I'm also thinking beyond, as the gateway is a "mix" type in the sense that is linked with VirtualServices, but it's also linked with Workloads.

Typically we assume that the "gateway" workload are the predefined in the control plane:
image

But we know that users can define "gateway" workloads in other places as well, but perhaps this last comment we can address them once the https://github.com/kiali/kiali/pull/4389 is progressed to not introduce too many new conflicts as well :-).

hhovsepy:
@lucasponce thanks for review, have added for AppLists as well, also fixed IstioConfig link from Lists page.

lucasponce (changes requested):
@hhovsepy I've found a minor cornercase/edge after I approved:
image

I think in the case of the apps it should check the duplicates.

lucasponce (approved):

skondkar (approved):
LGTM
Verified on Services and Applications list pages. Verified navigation of gateway links work and wrong gateway names in VirtualServices did not appear in the list pages.

Screenshot-ServiceList

Screenshot-AppList

PR #4408, opened 2021-10-05
Distinguish between proxy app logs and access logs

### LINKED ISSUE #4356: Workload pod proxy logs shows details for Envoy app logging
Envoy [application logs](https://www.envoyproxy.io/docs/envoy/latest/configuration/observability/application_logging) have a different format than the access logs. The logs console shows the "info" icon for both even though the log parsing does not work with the app logs.

Screenshot from 2021-09-15 15-55-28

Screenshot from 2021-09-15 15-55-50

It's possible to know what the structure of the envoy app logs will be based on the [log format string](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-log-format) but this will need to be read somewhere and could possibly change between proxies depending on where and how this is configured.

nrfox:
When getting pod logs, the proxy logs are parsed according to the access log format. The istio-proxy application logs are in a different format and instead of failing the parser returns an empty object causing the front end to misinterpret the proxy app logs as access logs.

Fixes #4356

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -391,7 +391,21 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 		if opts.IsProxy {
 			engardeParser := parser.New(parser.IstioProxyAccessLogsPattern)
 			al, err := engardeParser.Parse(entry.Message)
-			if err == nil {
+			// engardeParser.Parse will not throw errors even if no fields
+			// were parsed out. Checking here that some fields were actually
+			// set before setting the AccessLog to an empty object. See issue #4346.
+			if err != nil || isAccessLogEmpty(al) {
+				if err != nil {
+					log.Debugf("AccessLog parse failure: %s", err.Error())
+				}
+				// try to parse out the time manually
+				tokens := strings.SplitN(entry.Message, " ", 2)
+				timestampToken := strings.Trim(tokens[0], "[]")
+				t, err := time.Parse(time.RFC3339, timestampToken)
+				if err == nil {
+					parsedTimestamp = t
+				}
+			} else {
 				entry.AccessLog = al
 				t, err := time.Parse(time.RFC3339, al.Timestamp)
 				if err == nil {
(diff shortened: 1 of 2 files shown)


nrfox:
After fix, proxy app logs no longer have "info" icon in UI but access logs still do:

Screenshot from 2021-10-05 11-23-06

nrfox:
Hmmm it looks like there's an extra newline in the proxy app logs now

nrfox:
I think it's just a styling issue on the frontend and not related to the backend changes here. I'll open up a separate PR there.

business/workloads.go
```diff
@@ -391,7 +391,21 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
 		if opts.IsProxy {
 			engardeParser := parser.New(parser.IstioProxyAccessLogsPattern)
```
jshaughn, line 392: Not part of the PR, but just noticing we construct a new parser for every envoy log message, is that necessary or can we do it only once?

business/workloads.go
```diff
@@ -391,7 +391,21 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
...
 			al, err := engardeParser.Parse(entry.Message)
-			if err == nil {
+			// engardeParser.Parse will not throw errors even if no fields
```
jshaughn, line 394: Sort of annoying behavior from Engarde.  Just out of curiosity, does al.ParseError get set or is any way useful?

business/workloads.go
```diff
@@ -440,6 +445,38 @@ func (in *WorkloadService) GetPodLogs(namespace, name string, opts *LogOptions)
 	return in.getParsedLogs(namespace, name, opts)
 }
 
+func isAccessLogEmpty(al *parser.AccessLog) bool {
+	if al == nil {
```
jshaughn, line 449: I think the Timestamp is always needed to be successful,  perhaps we could move that one test up here and make this more efficient?

`if al == nil || al.Timestamp == ""`

jshaughn (commented):
It's working, still a couple of questions just in case we can make it simpler or faster.  Also, I'm +1 on fixing that excessive whitespace between log entries, on the UI side.  It wasn't always like that, something [outside this PR] changed.

business/workloads.go
nrfox: Sure I'll move that up

PR #4414, opened 2021-10-07
[molecule] test with default server image

jmazzitelli:
this is needed in case we want the operator to install the default server image.
Useful when testing with OLM and OSSM.

Operator PR: https://github.com/kiali/kiali-operator/pull/433
Backport PR: https://github.com/kiali/kiali/pull/4415


jmazzitelli:
we should backport this to v1.36 so we can easily test that branch using this new stuff.

PR #4415, opened 2021-10-08
[v1.36][molecule] test with default server image

jmazzitelli:
backport of https://github.com/kiali/kiali/pull/4414


jmazzitelli:
We do not need a new 1.36 build. I did this so we can switch to this branch and run the tests (using the current 1.36 build is fine). So, this is just so we can test.

Issue #4317, opened 2021-08-31
Support `exportTo` validation in Services (`networking.istio.io/exportTo` annotation)


hhovsepy:

hhovsepy:
Add support of exported Services in "virtualservices/NoHostChecker".

PR #4439, opened 2021-10-19
Added support of exported Services into VirtualService validation NoHostChecker

hhovsepy:
RFE #3061 
Subtask: https://github.com/kiali/kiali/issues/4317

In NoHostChecker of VirtualService, now it supports exported Services from other namespaces set into route -> destination -> host.
Exporting of Service is done via "networking.istio.io/exportTo" annotation on Service in Kubernetes.
Screenshot from 2021-10-19 08-43-33

Before, it was showing message that cross-namespace validations for hosts are not supported:
Screenshot from 2021-10-18 12-09-01

Now if Service is available it does not show any warning or error:
Screenshot from 2021-10-19 08-43-04

But when Service is exported only to own namespace, thus not available in other namespaces, error is shown on hosts of VS:

Screenshot from 2021-10-19 08-43-57

Screenshot from 2021-10-19 08-44-15

### DIFF
--- a/business/checkers/no_service_checker.go
+++ b/business/checkers/no_service_checker.go
@@ -30,12 +30,12 @@ func (in NoServiceChecker) Check() models.IstioValidations {
 		return validations
 	}
 
-	serviceNames := in.ServiceList.GetServiceNames()
 	serviceHosts := kubernetes.ServiceEntryHostnames(append(in.IstioConfigList.ServiceEntries, in.ExportedResources.ServiceEntries...))
 	gatewayNames := kubernetes.GatewayNames(in.GatewaysPerNamespace)
 
 	for _, virtualService := range in.IstioConfigList.VirtualServices {
-		validations.MergeValidations(runVirtualServiceCheck(virtualService, in.Namespace, serviceNames, serviceHosts, in.Namespaces, in.RegistryServices))
+		validations.MergeValidations(runVirtualServiceCheck(virtualService, in.Namespace, in.ServiceList, serviceHosts, in.Namespaces, in.RegistryServices))
+
 		validations.MergeValidations(runGatewayCheck(virtualService, gatewayNames))
 	}
 	for _, destinationRule := range in.IstioConfigList.DestinationRules {
@@ -44,13 +44,13 @@ func (in NoServiceChecker) Check() models.IstioValidations {
 	return validations
 }
 
-func runVirtualServiceCheck(virtualService networking_v1alpha3.VirtualService, namespace string, serviceNames []string, serviceHosts map[string][]string, clusterNamespaces models.Namespaces, registryStatus []*kubernetes.RegistryService) models.IstioValidations {
+func runVirtualServiceCheck(virtualService networking_v1alpha3.VirtualService, namespace string, services models.ServiceList, serviceHosts map[string][]string, clusterNamespaces models.Namespaces, registryStatus []*kubernetes.RegistryService) models.IstioValidations {
 	key, validations := EmptyValidValidation(virtualService.Name, virtualService.Namespace, VirtualCheckerType)
 
 	result, valid := virtualservices.NoHostChecker{
 		Namespace:         namespace,
 		Namespaces:        clusterNamespaces,
-		ServiceNames:      serviceNames,
+		ServiceList:       services,
 		VirtualService:    virtualService,
 		ServiceEntryHosts: serviceHosts,
 		RegistryServices:  registryStatus,
--- a/business/services.go
+++ b/business/services.go
@@ -41,6 +41,7 @@ func (in *SvcService) GetServiceList(criteria ServiceCriteria) (*models.ServiceL
 	var deployments []apps_v1.Deployment
 	var istioConfigList models.IstioConfigList
 	var err error
+
 	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
 	// if namespace is accessible from Kiali (Deployment.AccessibleNamespaces)
 	if _, err = in.businessLayer.Namespace.GetNamespace(criteria.Namespace); err != nil {
--- a/kubernetes/filters.go
+++ b/kubernetes/filters.go
@@ -240,11 +240,24 @@ func FilterRegistryServicesByServices(registryServices []*RegistryService, servi
 	return filtered
 }
 
-func FilterByRegistryService(hostname string, registryService *RegistryService) bool {
-	// Basic filter using Hostname
-	// TODO use the ExportTo, Namespace, ServiceRegistry and other attributes to refine the match
+func FilterByRegistryService(namespace string, hostname string, registryService *RegistryService) bool {
+	// Basic filter using Hostname, also consider exported Namespaces of Service
 	// but for a first iteration if it's found in the registry it will be considered "valid" to reduce the number of false validation errors
-	return hostname == registryService.Hostname
+	if hostname == registryService.Hostname {
+		exportTo := registryService.IstioService.Attributes.ExportTo
+		if len(exportTo) > 0 {
+			for exportToNs := range exportTo {
+				// take only namespaces where it is exported to, exported to the own namespace, or if it is exported to all namespaces
+				if exportToNs == "*" || exportToNs == namespace || (exportToNs == "." && registryService.IstioService.Attributes.Namespace == namespace) {
+					return true
+				}
+			}
+		} else {
+			// no exportTo field, means service exported to all namespaces
+			return true
+		}
+	}
+	return false
 }
 
 func FilterRequestAuthenticationsBySelector(workloadSelector string, requestauthentications []security_v1beta1.RequestAuthentication) []security_v1beta1.RequestAuthentication {
--- a/kubernetes/host.go
+++ b/kubernetes/host.go
@@ -205,8 +205,8 @@ func HasMatchingVirtualServices(host Host, virtualServices []networking_v1alpha3
 	return false
 }
 
-// HasMatchingRegistryService returns true when the FDQN of the host param matches
-// with one registry status of the registryStatus param.
+// HasMatchingRegistryService returns true when the FDQN of the host (from given namespace) param matches
+// with one registry service of the registryServices param.
 func HasMatchingRegistryService(namespace string, host string, registryServices []*RegistryService) bool {
 	for _, rStatus := range registryServices {
 		// We assume that on these cases the host.Service is provided in FQDN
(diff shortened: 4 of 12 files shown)


hhovsepy:
networking.istio.io/exportTo annotation can be a ',' separated list of namespaces, also can be "*", or "."

business/services.go
```diff
@@ -294,6 +319,7 @@ func (in *SvcService) buildRegistryServices(rSvcs []*kubernetes.RegistryService,
 			HealthAnnotations: map[string]string{},
 			Labels:            item.Attributes.Labels,
 			Selector:          item.Attributes.LabelSelectors,
+			ExportTo:          item.Attributes.ExportTo,
```
lucasponce, line 322: I think that the "ExportTo" is only required for the "AllNamespaces" flag (used only for validations). Am I right ?
I mean, the buildServiceList() complements the Kubernertes services with the Istio Registry ones, if we may want to use this field in other use case, perhaps we'd need to combine both in that point.

No action, from my side, just thinking loud on this.

kubernetes/filters.go
```diff
@@ -240,11 +240,24 @@ func FilterRegistryServicesByServices(registryServices []*RegistryService, servi
...
+func FilterByRegistryService(itemNamespace string, hostname string, registryService *RegistryService) bool {
+	// Basic filter using Hostname, also consider exported Namespaces of Service
 	// but for a first iteration if it's found in the registry it will be considered "valid" to reduce the number of false validation errors
-	return hostname == registryService.Hostname
+	if hostname == registryService.Hostname {
+		exportTo := registryService.IstioService.Attributes.ExportTo
+		if len(exportTo) > 0 {
+			for exportToNs := range exportTo {
+				// take only namespaces where it is exported to, or if it is exported to all namespaces
+				if exportToNs == "*" || exportToNs == itemNamespace {
```
lucasponce, line 251: How this filter should work with the "." case ?

Given the example:

```
  {
    "Attributes": {
      "ServiceRegistry": "External",
      "Name": "discovery.east-mesh-system.svc.west-mesh.local",
      "Namespace": "east-mesh-system",
      "Labels": {
        "topology.istio.io/network": "network-west-mesh"
      },
      "UID": "",
      "ExportTo": {
        ".": true
      },
      "LabelSelectors": null,
      "ClusterExternalAddresses": null,
      "ClusterExternalPorts": null
    },
    "ports": [
      {
        "name": "https-discovery",
        "port": 8188,
        "protocol": "HTTPS"
      }
    ],
    "creationTime": "2021-11-10T16:17:33.262606433Z",
    "hostname": "discovery.east-mesh-system.svc.west-mesh.local",
    "address": "0.0.0.0",
    "autoAllocatedAddress": "240.240.0.1",
    "Mutex": {},
    "Resolution": 1,
    "MeshExternal": true
  },
```

I understand that this service should be visible in the condition that the `ExportTo == "."` and the namespace of the service is same of the parameter, isn't it ?

kubernetes/filters.go
```diff
@@ -240,11 +240,24 @@ func FilterRegistryServicesByServices(registryServices []*RegistryService, servi
 	return filtered
 }
 
-func FilterByRegistryService(hostname string, registryService *RegistryService) bool {
-	// Basic filter using Hostname
-	// TODO use the ExportTo, Namespace, ServiceRegistry and other attributes to refine the match
+func FilterByRegistryService(itemNamespace string, hostname string, registryService *RegistryService) bool {
```
lucasponce, line 243: Minor, in the kubernetes/filters.go it's used the "namespace string" as part of the signature, perhaps we can keep that here as well, to avoid to diverge too much in the coding style.

```
func FilterByHost(host, serviceName, namespace string) bool
func FilterDestinationRulesByService(allDr []networking_v1alpha3.DestinationRule, namespace string, serviceName string) []networking_v1alpha3.DestinationRule
func FilterVirtualServicesByService(allVs []networking_v1alpha3.VirtualService, namespace string, serviceName string) []networking_v1alpha3.VirtualService {
```

kubernetes/host.go
```diff
@@ -207,11 +207,11 @@ func HasMatchingVirtualServices(host Host, virtualServices []networking_v1alpha3
 
 // HasMatchingRegistryService returns true when the FDQN of the host param matches
 // with one registry status of the registryStatus param.
-func HasMatchingRegistryService(host string, registryServices []*RegistryService) bool {
+func HasMatchingRegistryService(itemNamespace string, host string, registryServices []*RegistryService) bool {
```
lucasponce, line 210: Minor, I guess you brought the same name of the variable into the method signature.
I'd keep the variable as you added, but I'd try to normalize the signature to "namespace string".

It's just for coding style, to avoid to introduce too many edges in the naming.

business/services.go
hhovsepy: I think this line is odd now and can be removed, as Exported Services validations will be done directly in FilterByRegistryService in filters.

kubernetes/filters.go
hhovsepy: +1

models/service.go
```diff
@@ -33,6 +33,8 @@ type ServiceOverview struct {
 	Labels map[string]string `json:"labels"`
 	// Selector for Service
 	Selector map[string]string `json:"selector"`
+	// ExportTo for Service
+	ExportTo map[string]bool `json:"ExportTo,omitempty"`
```
lucasponce, line 37: The json map should start with minor case "exportTo" to be similar as other mappings.
I.e.:
```
IstioSidecar bool `json:"istioSidecar"`
AdditionalDetailSample *AdditionalItem `json:"additionalDetailSample"`
IstioReferences []*IstioValidationKey `json:"istioReferences"`
```

Also, this is related to the discussion in the business/service.go, if we may want to propagated in all services, or just use it for Istio Registry use cases, then it would be empty in rest of the cases (it's fine with me if we don't need it).

If want to render it in the UI, then we'd need to adapt the buildService() list and also propagate this field in the UI type and ServiceDetails page I think.

Just a note, I think that this comment can be a side task not directly related with the PR, just leaving this thought to not forget.

kubernetes/filters.go
hhovsepy: Right, it is correct to include '.' as well, however those services are considered as services from current namespace, not imported so would not be missed and would be matched before reaching this filters.

models/service.go
```diff
@@ -33,6 +33,8 @@ type ServiceOverview struct {
 	Labels map[string]string `json:"labels"`
 	// Selector for Service
 	Selector map[string]string `json:"selector"`
+	// ExportTo for Service
```
lucasponce, line 36: Minor: worth to comment that this ExportTo comes from the Istio annotation to have a better context.

business/services.go
```diff
@@ -27,6 +28,7 @@ type SvcService struct {
 }
 
 type ServiceCriteria struct {
+	AllNamespaces          bool
```
lucasponce, line 31: Is this being used in the PR ?
Perhaps you wanted to use it in the business/istio_validations.go ? Or that would be a followup task ?

business/checkers/no_service_checker.go
```diff
@@ -30,12 +30,12 @@ func (in NoServiceChecker) Check() models.IstioValidations {
 		return validations
 	}
 
-	serviceNames := in.ServiceList.GetServiceNames()
 	serviceHosts := kubernetes.ServiceEntryHostnames(append(in.IstioConfigList.ServiceEntries, in.ExportedResources.ServiceEntries...))
 	gatewayNames := kubernetes.GatewayNames(in.GatewaysPerNamespace)
 
 	for _, virtualService := range in.IstioConfigList.VirtualServices {
-		validations.MergeValidations(runVirtualServiceCheck(virtualService, in.Namespace, serviceNames, serviceHosts, in.Namespaces, in.RegistryServices))
+		validations.MergeValidations(runVirtualServiceCheck(virtualService, in.Namespace, in.ServiceList, serviceHosts, in.Namespaces, in.RegistryServices))
```
lucasponce, line 37: Perhaps this should need a more deep refactoring if we start using the registry ?

I see that ServiceEntries were also added to this validations:
image

But now those should be available in the ServiceList with the AllNamespaces flag.

I'm just asking, I know this is part of the https://github.com/kiali/kiali/issues/4528 and you may want to plan to add it on that effort (and it's me mixing both contexts during this PR review).

lucasponce: As I understand the context, I'm fine if we want to consolidate this work and waiting for #4528

lucasponce: Ok, now I understand the context, validations were using the RegistryServices as the last way to check the exported services.

Yes, it makes sense in the context of the PR, I think we can consolidate it and wait for #4528 to unify to only use a single source for fetching the Services.

I was confused with the introduction of the AllNamespace flag in this PR.

hhovsepy: I think the ServiceEntries from Registry could be added in a followup PRs.

kubernetes/host.go
hhovsepy: +1

models/service.go
hhovsepy: removing this line

business/services.go
hhovsepy: You are right, this is not used in this PR scope, so removing it now, will be added in further PRs if necessary.

hhovsepy:
Thank you @lucasponce for review, I approached changes, removed a bunch of odd changes related to ExportTo field and AllNamespaces which are not used within this PR. Squashed the commits.

business/checkers/no_service_checker.go
lucasponce: > I think the ServiceEntries from Registry could be added in a followup PRs.

Well, I think it's part of the #4528 effort, to revisit the information needed in the business/istio_validations.go to propagate to the checkers.

There will be more cases like this one.

kubernetes/host.go
```diff
@@ -207,11 +207,11 @@ func HasMatchingVirtualServices(host Host, virtualServices []networking_v1alpha3
 
 // HasMatchingRegistryService returns true when the FDQN of the host param matches
```
lucasponce, line 208: I'd update the comment to indicate when the namespace parameter is used in the condition.

lucasponce (dismissed):

lucasponce (approved):

skondkar (approved):
LGTM
Verified that when Service is available to other namespaces, the VS config does not show any warning or error:
Screenshot from 2021-12-02 17-12-53

After exporting the Service to own namespace via "networking.istio.io/exportTo" annotation on Service, (not available in other namespaces), verified that error is shown on hosts of VS:

Screenshot from 2021-12-02 17-01-06

Screenshot from 2021-12-02 16-59-10

Verified that adding the other namespace name to the exportTO list does not show warning/error in VS config.
Also verified using the networking.istio.io/exportTo annotations "*", and "."

skondkar:
Polarion test case >> SWSQE-2170

PR #4441, opened 2021-10-19
make format: sorting imports (format broken in the ci)

xeviknal:
Is it only me that have this bunch of modification after running the make format command (gofmt)?
Here are the files that are modified. Am I running a different version of gofmt?

I think the CI is broken. Let me check it out.

### DIFF
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -7,7 +7,6 @@ import (
 	networking_v1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
 	apps_v1 "k8s.io/api/apps/v1"
 	core_v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


lucasponce:
> I think the CI is broken. Let me check it out.

Where ?

I'm doing bug fixing and the CI didn't show a broken status in the previous steps ?

I saw some error message in the Jenkins (but not related format) but the Github Actions looked ok.

I don't run the make format for some time, I rely on the IDE and the CI actions.

Are you seeing any other issue ?

xeviknal:
> Where ?

If you look into this PR, actually into the last commit, I added a format problem on purpose.
The result of the GH workflow is green. However, the import is in an incorrect place.
Running the make format in my local machine, results into the proper import sort though.
I see this as an error.

If you run the make format in master, there are 30 files with incorrect import sorts.

PR #114, opened 2018-03-15
SWS-327 Fix fetch Istio Rule with non supported handlers/instances

lucasponce:
- A non supported handler/instance is converted in a nil object but prevent to fail

PR #4465, opened 2021-11-02
New function to check if namespace is the root namespace

### LINKED ISSUE #4448: Support getting the root namespace from Istio configuration
According to #3062, the root namespace is required to do validations.

The root namespace is configured in Istio configuration (configmap) according to the [documentation](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/).

This issue is about creating a new mechanism to get the root namespace for further use in validations.

Note: if the root namespace is not configured, it defaults to "istio-system" namespace.

leandroberetta:
This PR adds a similar function than "isIstioNamespace" but for the administrative root namespace.

Requires [#441](https://github.com/kiali/kiali-operator/pull/441)
Requires [#114](https://github.com/kiali/helm-charts/pull/114)

Resolves #4448

If not configured, the root namespace will be "istio-system".

Note: This is an alternative solution to #4448 comparing to #4447

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -186,6 +186,7 @@ type IstioConfig struct {
 	IstioSidecarInjectorConfigMapName string              `yaml:"istio_sidecar_injector_config_map_name,omitempty"`
 	IstioSidecarAnnotation            string              `yaml:"istio_sidecar_annotation,omitempty"`
 	IstiodDeploymentName              string              `yaml:"istiod_deployment_name,omitempty"`
+	RootNamespace                     string              `yaml:"root_namespace,omitempty"`
 	UrlServiceVersion                 string              `yaml:"url_service_version"`
 }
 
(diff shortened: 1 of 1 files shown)


xeviknal (approved):
LGTM! I think it is much more flexible to have this defined in the config, plus much more performant.

leandroberetta:
@lucasponce @xeviknal as this PR introduces some new operator/helm changes it would be a good idea to include some testing installing Kiali and see it is correctly installed.

The only point to check apart from Kiali being well installed is to check that in the Kiali configmap, there is the following new property with the default value set:

```yaml
  config.yaml: |
    ...
    external_services:
      ...
      istio:
        ...
        root_namespace: istio-system
        ...

```

Also if the root namespace is set in Kiali CR (or in the Helm values file) the config map should have that value instead of the default one (istio-system).

lucasponce (approved):

lucasponce:
I think this one is minor and it wouldn't need QE testing, isn't it ?

PR #4470, opened 2021-11-03
Global sidecar validation using the root namespace 

### LINKED ISSUE #4449: Support rootNamespace in Sidecar validations
According to #3062 the Sidecar validations need to support the root namespace if configured. 

Today the root namespace is defaulted to "istio-system" but it needs to be changed to the root namespace.

https://istio.io/latest/docs/reference/config/networking/sidecar/

leandroberetta:
This PR updates the global sidecar validation to honor the root namespace instead of the istio namespace.

Depends on #4465 

Fixes #4449

In the following example, the global validation applies to the Sidecar created in istio-config namespace (the root namespace).

root_namespace_sidecar

### DIFF
--- a/business/checkers/sidecars/global_checker.go
+++ b/business/checkers/sidecars/global_checker.go
@@ -14,7 +14,7 @@ type GlobalChecker struct {
 func (gc GlobalChecker) Check() ([]*models.IstioCheck, bool) {
 	checks, valid := make([]*models.IstioCheck, 0), true
 
-	if !config.IsIstioNamespace(gc.Sidecar.Namespace) {
+	if !config.IsRootNamespace(gc.Sidecar.Namespace) {
 		return checks, valid
 	}
 
--- a/business/checkers/sidecars/global_checker_test.go
+++ b/business/checkers/sidecars/global_checker_test.go
@@ -29,7 +29,7 @@ func TestSidecarWithoutSelectorInControlPlane(t *testing.T) {
 	config.Set(conf)
 
 	vals, valid := GlobalChecker{
-		Sidecar: *data.CreateSidecar("sidecar1", conf.IstioNamespace),
+		Sidecar: *data.CreateSidecar("sidecar1", conf.ExternalServices.Istio.RootNamespace),
 	}.Check()
 
 	assert.Empty(vals)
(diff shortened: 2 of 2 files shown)


xeviknal (approved):
LGTM!

leandroberetta:
The steps to test this are the following:

1. Configure Kiali to a different root namespace, for example "istio-config", this can be done via the CR:

```yaml
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:     
  name: kiali  
spec:
  external_services:    
    istio:
      root_namespace: istio-config
```

2. Wait for the operator to update Kiali configmap 

3. Create the following global Sidecar resource in istio-config namespace:

```yaml
kind: Sidecar
metadata:
  name: default
  namespace: istio-config  
spec:
  workloadSelector:
    labels:
      app: ratings
  egress:
  - hosts:
    - "./*"
    - "istio-system/*"
```

4. Expect to see the following warnings in validations:
 
root_namespace_sidecar

prachiyadav (changes requested):
This PR works when manually adding "istio-config" namespace to accessible namespace. Otherwise with default configuration it is not working.

jmazzitelli:
> This PR works when manually adding "istio-config" namespace to accessible namespace. Otherwise with default configuration it is not working.

Sounds like if istio.root_namespace is defined and not empty string AND accessible namespaces is not `['**']` then that root namespace needs to be added to the accessible_namespaces.

We do something similar with the deployment namespace in the operator and server helm chart, IIRC. Just need to do the same thing for root_namespace.

leandroberetta:
> > This PR works when manually adding "istio-config" namespace to accessible namespace. Otherwise with default configuration it is not working.
> 
> Sounds like if istio.root_namespace is defined and not empty string AND accessible namespaces is not `['**']` then that root namespace needs to be added to the accessible_namespaces.
> 
> We do something similar with the deployment namespace in the operator and server helm chart, IIRC. Just need to do the same thing for root_namespace.

@jmazzitelli thanks, yes, it's the same issue we got here: https://github.com/kiali/kiali/pull/4490

I will do that.

leandroberetta:
@prachiyadav can you try again this with the latest change on the operator (in master branch)?.

Now the root namespace should be included in the accessible_namespaces attribute.

prachiyadav (approved):

prachiyadav:
Test case >> SWSQE-2172

PR #4471, opened 2021-11-03
bugfix: free ports after using them

xeviknal:
After reading this discussion https://github.com/kiali/kiali/discussions/4461 I realized that this could be messing at the time of allocating ports to the port-forwards we have.

It might not be the problem mentioned in the discussion but it something is necessary to approach.

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -278,11 +278,13 @@ func (iss *IstioStatusService) getIstiodReachingCheck() (IstioComponentStatus, e
 	for _, istiod := range healthyIstiods {
 		go func(name, namespace string) {
 			defer wg.Done()
+			freePort := httputil.Pool.GetFreePort()
+			defer httputil.Pool.FreePort(freePort)
 			// Using the proxy method to make sure that K8s API has access to the Istio Control Plane namespace.
 			// By proxying one Istiod, we ensure that the following connection is allowed:
 			// Kiali -> K8s API (proxy) -> istiod
 			// This scenario is no obvious for private clusters (like GKE private cluster)
-			_, err := iss.k8s.ForwardGetRequest(namespace, name, httputil.Pool.GetFreePort(), 8080, "/ready")
+			_, err := iss.k8s.ForwardGetRequest(namespace, name, freePort, 8080, "/ready")
 			if err != nil {
 				syncChan <- ComponentStatus{
 					Name:   name,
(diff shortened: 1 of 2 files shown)


util/httputil/port_pool.go
```diff
@@ -28,7 +29,7 @@ type PortPool struct {
 
 var Pool = &PortPool{
 	LastBusyPort:  13999,
-	Mutex:         sync.Mutex{},
+	Mutex:         &sync.Mutex{},
 	PortsMap:      map[int]bool{},
```
lucasponce, line 33: This PR looks ok from my side.
I was reviewing the algorithm and I was thinking that perhaps it could be simplify instead to use a map, to just use an bool array of the size of the PortRangeSize, and then the logic could be just to return the first free port.
Then the LastBusyPort shouldn't be needed, then it could skip a couple of validations and if the range is short it can be increased.

lucasponce (approved):

hhovsepy:
Doing some regression testing, is there a way to see that port is free or taken, nothing logs?
Otherwise looks good.

util/httputil/port_pool.go
xeviknal: Yes, you are right. Logic could be simplified and LastBusyPort might be removed. However, the complexity of the algorithm would rise a bit because you should always need to visit the first k elements to get to the first free port.
Essentially this algorithm does what you mentioned: iterate over a range until you find the first free slot. However, it starts in better position, making the search probably closer to the next free port. In most of the cases, the free port will be `LastBusyPort + 1`.

I could think of one way to reduce logic:

```go
pool.LastBusyPort = pool.PortRangeInit + (pool.LastBusyPort+1)%pool.PortRangeSize
```
Instead of

```go
if pool.LastBusyPort >= pool.PortRangeInit+pool.PortRangeSize-1 {
	pool.LastBusyPort = pool.PortRangeInit
// If the pointer is inside the range, increment by 1
} else {
	pool.LastBusyPort++
}
```

This way, the LastBusyPort is always within the range with no need to use `if-else` statement.
The rest of the logic I guess it would remain the same: the loop condition, and the post lop `if` statement.

lucasponce: > However, the complexity of the algorithm would rise a bit because you should always need to visit the first k elements to get to the first free port.

Yes but as the iteration is max of 1000 elements, perhaps using a map instead a plain array increases the memory overhead :) and the tradeoff of moving O(N) -> O(log(N)) is adding extra cost at memory :).

Again it's looking good :) as it's now.

hhovsepy (approved):
Verified from QE perspective.

Issue #4382, opened 2021-09-21
Refactor Kiali Validations to better use the Istio registry information


lucasponce:
The Kiali validations fetch workloads, services, pods and Istio resources to perform "semantic validations". 
The term "semantic validation" refers to validate that a hostname or a resource referenced in a Istio object exists in the cluster, for example, an internal hostname should match a kubernetes service, a workload selector should have a workload in the namespace that is defined, etc.

Recently, Kiali introduced the Istio registry as another source of information, but the Istio registry overlaps some of the information collected fetching directly Kubernetes services and workloads.

This Epic is a proposal to refactor the existing Istio Validations to reduce the number of fetches and use the Istio registry as main source of information for services and workloads.

The proposal have several advantages: cross-namespace validations can be calculated in a uniform way as the registry fetches all the cluster/mesh visibility; another advantage is that validation logic will use less fetches and it will simplify the code.

The istio registry has several APIs, today is used the "registryz", but for this work it may be necessary to use the "endpointz" one.

The endpoints can give the list of workloads and labels, so validations as label selector can be calculated with this information.

In terms of feature, this is a refactoring, meaning that there should be not regression in the existing logic.

This is a brief list of potentital changes in the validation logic, probably it's not 100% complete, but it would give an idea about the changes in the existing checkers:

```
1.- istio_validations.go

#GetValidations

Service can be validated against the registry, instead to fetch the kubernetes service

Workloads used in the validations can be replaced by the registry endpoints, which links all workloads with services.
That could be enough for the validation logic, as the registry has the proper labels information.

Potentially the list of namespaces could be infered from the registry ?

2.- no_service_checker.go

Service names and service hosts can be used from the Istio registry.

3.- no_host_checker.go

#checkDestination can unify the host check and consolidate/simplify the fqdn and cross validation logic using the registry only.
No service names, service entries or namespaces should be necessary using the registry.

4.- no_gateway_checker.go

+1 Valid, no changes as this validations iterates between Istio Gateways and VirtualServices

5.- no_dest_checker.go

Host logic can be checked against registry.

The #HasMatchingWorkloads logic could be replaced with the endpoints as they contain the list of pods with the workload identified.
The #HasMatchingServices and #HasMatchingServiceEntries logic could be replaced in a similar way.

6.- route_checker.go

+1 Valid

7.- subset_presence_checker.go

+1 Valid

8.- export_to_namespace_checker.go

[...]

PR #4479, opened 2021-11-04
Istio Registry changes

lucasponce:
Follow up of #4456 
UI PR required https://github.com/kiali/kiali-ui/pull/2273

The goal of this PR is to better fetch the Istio Registry information (services, endpoints and configuration) to use it in several features of Kiali (service list and details, validations and other places).

This PR focus on the Service list and details not covering the validations code, that will be part of https://github.com/kiali/kiali/issues/4382 epic

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -14,6 +14,7 @@ import (
 	api_types "k8s.io/apimachinery/pkg/types"
 
 	"context"
+
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
@@ -28,6 +29,11 @@ type IstioConfigService struct {
 }
 
 type IstioConfigCriteria struct {
+	// When AllNamespaces is true the IstioConfigService will use the Istio registry to return the configuration
+	// from all namespaces directly from the Istio registry instead of the individual API
+	// This usecase should be reserved for validations use cases only where cross-namespace validation may create a
+	// penalty
+	AllNamespaces                 bool
 	Namespace                     string
 	IncludeGateways               bool
 	IncludeVirtualServices        bool
@@ -92,7 +98,7 @@ var newSecurityConfigTypes = []string{
 // GetIstioConfigList returns a list of Istio routing objects, Mixer Rules, (etc.)
 // per a given Namespace.
 func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (models.IstioConfigList, error) {
-	if criteria.Namespace == "" {
+	if criteria.Namespace == "" && !criteria.AllNamespaces {
 		return models.IstioConfigList{}, errors.New("GetIstioConfigList needs a non empty Namespace")
 	}
 	istioConfigList := models.IstioConfigList{
@@ -110,6 +116,36 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 		AuthorizationPolicies:  []security_v1beta1.AuthorizationPolicy{},
 		PeerAuthentications:    []security_v1beta1.PeerAuthentication{},
 		RequestAuthentications: []security_v1beta1.RequestAuthentication{},
+	}
+
+	// Use the Istio Registry when AllNamespaces is present
+	if criteria.AllNamespaces {
+		registryCriteria := RegistryCriteria{
+			AllNamespaces: true,
+		}
+		registryConfiguration, err := in.businessLayer.RegistryStatus.GetRegistryConfiguration(registryCriteria)
+		if err != nil {
+			return istioConfigList, err
+		}
+		if registryConfiguration == nil {
+			log.Warningf("RegistryConfiguration is nil. This is an unexpected case. Is the Kiali cache disabled ?")
+			return istioConfigList, nil
+		}
+		// AllNamespaces will return an empty namespace
+		istioConfigList.Namespace.Name = ""
+		istioConfigList.DestinationRules = registryConfiguration.DestinationRules
+		istioConfigList.EnvoyFilters = registryConfiguration.EnvoyFilters
+		istioConfigList.Gateways = registryConfiguration.Gateways
+		istioConfigList.VirtualServices = registryConfiguration.VirtualServices
+		istioConfigList.ServiceEntries = registryConfiguration.ServiceEntries
+		istioConfigList.Sidecars = registryConfiguration.Sidecars
+		istioConfigList.WorkloadEntries = registryConfiguration.WorkloadEntries
+		istioConfigList.WorkloadGroups = registryConfiguration.WorkloadGroups
+		istioConfigList.AuthorizationPolicies = registryConfiguration.AuthorizationPolicies
+		istioConfigList.PeerAuthentications = registryConfiguration.PeerAuthentications
+		istioConfigList.RequestAuthentications = registryConfiguration.RequestAuthentications
+
+		return istioConfigList, nil
 	}
 
 	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
@@ -163,7 +199,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			}
 			if err == nil {
 				if isWorkloadSelector {
-					istioConfigList.EnvoyFilters = kubernetes.FilterEnvoyFilters(workloadSelector, istioConfigList.EnvoyFilters)
+					istioConfigList.EnvoyFilters = kubernetes.FilterEnvoyFiltersBySelector(workloadSelector, istioConfigList.EnvoyFilters)
 				}
 			} else {
 				errChan <- err
@@ -185,7 +221,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			}
 			if err == nil {
 				if isWorkloadSelector {
-					istioConfigList.Gateways = kubernetes.FilterGateways(workloadSelector, istioConfigList.Gateways)
+					istioConfigList.Gateways = kubernetes.FilterGatewaysBySelector(workloadSelector, istioConfigList.Gateways)
 				}
 			} else {
 				errChan <- err
@@ -224,7 +260,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			}
 			if err == nil {
 				if isWorkloadSelector {
-					istioConfigList.Sidecars = kubernetes.FilterSidecars(workloadSelector, istioConfigList.Sidecars)
+					istioConfigList.Sidecars = kubernetes.FilterSidecarsBySelector(workloadSelector, istioConfigList.Sidecars)
 				}
 			} else {
 				errChan <- err
@@ -297,7 +333,7 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 			}
 			if err == nil {
 				if isWorkloadSelector {
-					istioConfigList.AuthorizationPolicies = kubernetes.FilterAuthorizationPolicies(workloadSelector, istioConfigList.AuthorizationPolicies)
+					istioConfigList.AuthorizationPolicies = kubernetes.FilterAuthorizationPoliciesBySelector(workloadSelector, istioConfigList.AuthorizationPolicies)
 				}
 			} else {
 				errChan <- err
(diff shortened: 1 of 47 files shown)


lucasponce:
@xeviknal @hhovsepy adding you in an early stage, the work is not yet finished but I'll update the PR to see the progress.

One design decision to take is when to refresh the Istio Registry info.

Today as it's not critical it works via time out, but I'm planning to use the cache events to react/refresh the cache on Services informes, then the cache will react better on new Services updates.

business/istio_validations.go
```diff
@@ -61,7 +61,7 @@ func (in *IstioValidationsService) GetValidations(namespace, service string) (mo
 	var gatewaysPerNamespace [][]networking_v1alpha3.Gateway
 	var mtlsDetails kubernetes.MTLSDetails
 	var rbacDetails kubernetes.RBACDetails
-	var registryStatus []*kubernetes.RegistryStatus
+	var registryStatus []*kubernetes.RegistryService
```
hhovsepy, line 64: might be changing var name to 'registryService' ?

lucasponce: +1

kubernetes/cache/cache_handlers.go
```diff
@@ -0,0 +1,48 @@
...
+	cache.ResourceEventHandler
+}
+
+func NewServiceHandler(cache KialiCache, namespace string) ServiceHandler {
+	log.Infof("Adding a ServiceHandler for namespace [%s]", namespace)
+	return ServiceHandler{Cache: cache, Namespace: namespace}
+}
+
+func (sh ServiceHandler) OnAdd(obj interface{}) {
+	svc, ok := obj.(*core_v1.Service)
```
xeviknal, line 22: Should only refresh the cache when Services are added/updated/removed? Perhaps ServiceEntries as well? And in general other resources that modify the registry?

lucasponce: With Endpoints and ServiceEntries I think it could be good.
Also this cache is short term as it's using the tokenNamespace timeout, but I'm thinking that with this eventing mechanism it can have an increased period here.

hhovsepy:
On OCP cluster with Istio 1.12 I see this errors: "Could not fetch Istio objects list: json: cannot unmarshal array into Go struct field .clusterVIPs.Addresses of type string"

lucasponce:
> On OCP cluster with Istio 1.12 I see this errors: "Could not fetch Istio objects list: json: cannot unmarshal array into Go struct field .clusterVIPs.Addresses of type string"

Good point, the marshalling should work well with 1.11 + 1.12, it seems I introduced some bug.

I've tested the PRs with 1.11, but I'm going to validate it with the next version.

hhovsepy:
Works with Istio v1.12
Screenshot from 2021-11-23 11-41-45

leandroberetta (commented):
Partial review, put questions and minor comments.

Also I tested this on Maistra 2.1 (Istio 1.9) and I'm seeing the following graph for the electronic shop demo:

sr_maistra_graph

I'ts different from what I see on Istio 1.11

Also, it seems that when accessing a service with a * (*.msn.com) the UI crashes:

sr_maistra_msn_crash

business/services.go
```diff
@@ -214,9 +230,57 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 			Selector:               item.Spec.Selector,
 			IstioReferences:        svcReferences,
 			KialiWizard:            kialiWizard,
+			ServiceRegistry:        "Kubernetes",
 		}
 	}
 
+	// Add Istio Registry Services that are not present in the Kubernetes list
+	rSvcs = kubernetes.FilterRegistryServicesByServices(rSvcs, svcs)
+	for _, item := range rSvcs {
```
hhovsepy, line 239: Can this loop be moved to separate method for reusing porposes?

business/services.go
```diff
@@ -214,9 +230,57 @@ func (in *SvcService) buildServiceList(namespace models.Namespace, svcs []core_v
 			Selector:               item.Spec.Selector,
 			IstioReferences:        svcReferences,
 			KialiWizard:            kialiWizard,
+			ServiceRegistry:        "Kubernetes",
 		}
 	}
 
+	// Add Istio Registry Services that are not present in the Kubernetes list
+	rSvcs = kubernetes.FilterRegistryServicesByServices(rSvcs, svcs)
```
hhovsepy, line 238: Here I see that Kiali first takes Services from K8 then takes the rest of services from Registry, right?
Wasn't the reason of introducing Registry in having listed from Registry first, then missing information from K8?

lucasponce:
@leandroberetta 

> Also, it seems that when accessing a service with a * (*.msn.com) the UI crashes:

Did you check with the UI PR linked with this one ? There are some changes in the UI that affects to the visualization.

I've just tested with the Maistra 2.1 (Istio 1.9) and connecting from out of the cluster, I can see correctly the details:

image

Let me know if you can reproduce the problem.

kubernetes/istio.go
```diff
@@ -167,20 +191,151 @@ func parseProxyStatus(statuses map[string][]byte) ([]*ProxyStatus, error) {
...
+	var fullRegistryEndpoints []*RegistryEndpoint
+	for pilot, endpoint := range endpoints {
+		var eps []*RegistryEndpoint
+		err := json.Unmarshal(endpoint, &eps)
+		if err != nil {
+			log.Errorf("Error parsing RegistryEndpoints results: %s", err)
+			return nil, err
+		}
+		for _, ep := range eps {
+			ep.pilot = pilot
```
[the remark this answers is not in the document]
lucasponce: Yes, that's correct.
In the case of the "proxy-status" information we needed to iterate over all istiod pods to collect all the state.
On the registry endpoints perhaps is not necessary but I thought it could be useful to also bring this information for debugging purposes.

It's a good comment, I'm going to test this case to check that nothing is missed nor duplicated introduced.

business/istio_config.go
```diff
@@ -92,7 +98,7 @@ var newSecurityConfigTypes = []string{
 // GetIstioConfigList returns a list of Istio routing objects, Mixer Rules, (etc.)
 // per a given Namespace.
 func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (models.IstioConfigList, error) {
-	if criteria.Namespace == "" {
+	if criteria.Namespace == "" && criteria.AllNamespaces {
```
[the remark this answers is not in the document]
lucasponce: Yes, good catch, it should return with an error when no namespace and no flag for all-namespaces.

business/istio_validations.go
```diff
@@ -99,15 +99,15 @@ func (in *IstioValidationsService) GetValidations(namespace, service string) (mo
 	return validations, nil
 }
 
-func (in *IstioValidationsService) getAllObjectCheckers(namespace string, istioConfigList models.IstioConfigList, exportedResources kubernetes.ExportedResources, services models.ServiceList, workloadsPerNamespace map[string]models.WorkloadList, workloads models.WorkloadList, gatewaysPerNamespace [][]networking_v1alpha3.Gateway, mtlsDetails kubernetes.MTLSDetails, rbacDetails kubernetes.RBACDetails, namespaces []models.Namespace, registryStatus []*kubernetes.RegistryStatus) []ObjectChecker {
+func (in *IstioValidationsService) getAllObjectCheckers(namespace string, istioConfigList models.IstioConfigList, exportedResources kubernetes.ExportedResources, services models.ServiceList, workloadsPerNamespace map[string]models.WorkloadList, workloads models.WorkloadList, gatewaysPerNamespace [][]networking_v1alpha3.Gateway, mtlsDetails kubernetes.MTLSDetails, rbacDetails kubernetes.RBACDetails, namespaces []models.Namespace, registryStatus []*kubernetes.RegistryService) []ObjectChecker {
```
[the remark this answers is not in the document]
lucasponce: +1

doc.go
```diff
@@ -831,13 +831,6 @@ type MetricsStatsResponse struct {
 	Body models.MetricsStats
 }
 
-// Return a list of Cluster items
```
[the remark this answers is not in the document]
lucasponce: I removed it as it was stopped by the Github Action due a swagger check, it pointed to this unused entry, perhaps this removal is incorrect ?

handlers/dashboards.go
```diff
@@ -159,6 +159,24 @@ func ServiceDashboard(w http.ResponseWriter, r *http.Request) {
...
+	svc, err := b.Svc.GetService(namespace, service)
+	if err != nil {
+		RespondWithError(w, http.StatusServiceUnavailable, err.Error())
+		return
+	}
+
+	// "External"/"ServiceEntry" services don't use namespace in telemetry, they need to use the "unknown" parameter
+	// to collect the relevant telemetry for those services
+	if svc.Type == "External" {
+		params.Namespace = "unknown"
```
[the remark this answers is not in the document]
lucasponce: +1 we can refine some of the constants used

business/services.go
lucasponce: Good point, but I'm not sure, what do you have in mind ?
This loop should be used inside the business/services.go, and other features like validations, appender or security should use the business.SvcService to fetch services (not replicating this logic).

I guess you are thinking on fetching the services from the validations ? Then we could add a AllNamespaces criteria in the ServiceCriteria (or any other needed filtering required for this job).

It's similar to the IstioConfigList service, the `GetIstioConfigList` returns the Istio config, and using the criteria you can fetch information from one source to another.

The goal here is to simplify the pattern and avoid that there are multiple requests for the same logic.

Perhaps, was that your use case ?

business/services.go
lucasponce: Right.
The reationale for this was:
- Keep using the Kubernetes Services, as they information that is not collected in Istio (annotations, pod).
- Use the Istio registry for those services that are visible in the sidecars but not part of the kubernetes Services.

At least for the case of the Service list/details pages.

Probably we could introduce another use case for validations or similar when we want to perform a cross-namespace query with few data (exportTo, name, namespace, attribute, hostname).

Then, you're right, we could introduce a new flag in the business.SvcServiceCriteria that would help to use this case.

leandroberetta:
> @leandroberetta
> 
> > Also, it seems that when accessing a service with a * (*.msn.com) the UI crashes:
> 
> Did you check with the UI PR linked with this one ? There are some changes in the UI that affects to the visualization.
> 
> I've just tested with the Maistra 2.1 (Istio 1.9) and connecting from out of the cluster, I can see correctly the details:
> 
> image
> 
> Let me know if you can reproduce the problem.

Oh, my bad... sorry, dismiss this then.

business/services.go
hhovsepy: Yeah, I mean having it as a separate method within the business/services.go, and be also called in a method "GetServiceList" when param AllNamespaces is true. So when AllNamespaces is true, then similar to "GetIstioConfigList" it will list Services from Registry only and will process them, but without code duplication, especially the "ServiceOverview" objects creation from Registry Services.

lucasponce: Ah, I guess you mean to have a split of the "buildServiceList" in two methods, one per kubernetes Services and another for Istio Registry Services and combine/concat.

I can try to re-org those methods in that way.

hhovsepy: right, and to be able to call Registry Services one directly from GetServiceList for AllNamespaces true case

lucasponce:
@leandroberetta @hhovsepy I've updated the PR with your comments and I've increased the test coverage.

The UI PR still needs some work.

hhovsepy:
Did regression testing, one e2e test is failing, the scenario is: request Service Details via Kiali API call of non existing service in istio-system namespace, the expected result was '502', but now Kiali returns service with empty data:
Screenshot from 2021-11-25 09-43-50

lucasponce:
> Did regression testing, one e2e test is failing, the scenario is: request Service Details via Kiali API call of non existing service in istio-system namespace, the expected result was '502', but now Kiali returns service with empty data:

+1

Perhaps it can be also sent a Not Found/404 on this case.

I mean, it will break the e2e but perhaps it made sense to update the code, perhaps this was in the code for some time and it wasn't caught before.

hhovsepy:
> +1
> 
> Perhaps it can be also sent a Not Found/404 on this case.
> 
> I mean, it will break the e2e but perhaps it made sense to update the code, perhaps this was in the code for some time and it wasn't caught before.

+1 agree

hhovsepy:
Did UI and e2e regression testing, all seems fine to me. Only missing the return of "Not Found/404" in case of not existing Service details requesting.

lucasponce:
> Did UI and e2e regression testing, all seems fine to me. Only missing the return of "Not Found/404" in case of not existing Service details requesting.

Thanks, I've just updated this case as well.

hhovsepy (approved):
Did regression testing, looks good.
Run e2e tests. 3 of them are failing, which will require fixing expected return codes of those tests, as now 404 is retuned in caseof not found namespace or service.
tests.test_api_methods.test_negative_500
tests.test_api_methods.test_invalid_service_spans_list_and_trace_list
tests.test_api_methods.test_negative_502

leandroberetta (approved):
LGTM, tested on upstream with Istio 1.11 and 1.12.

PR #4490, opened 2021-11-09
Peer authentication validations using the root namespace

### LINKED ISSUE #4450: Support rootNamespace in Peer Authentication validations
According to #3062 the Peer Authentication validations need to support the root namespace if configured.

Today the root namespace is defaulted to "istio-system" but it needs to be changed to the root namespace.

https://istio.io/latest/docs/reference/config/security/peer_authentication/

leandroberetta:
This PR updates the peer authentication validations to honor the root namespace instead of the istio namespace.

Depends on #4465

Fixes #4450

### DIFF
--- a/business/checkers/peer_authentication_checker.go
+++ b/business/checkers/peer_authentication_checker.go
@@ -42,7 +42,7 @@ func (m PeerAuthenticationChecker) runChecks(peerAuthn security_v1beta.PeerAuthe
 		matchLabels = peerAuthn.Spec.Selector.MatchLabels
 	}
 	enabledCheckers = append(enabledCheckers, common.SelectorNoWorkloadFoundChecker(PeerAuthenticationCheckerType, matchLabels, m.WorkloadList))
-	if peerAuthn.Namespace == config.Get().IstioNamespace {
+	if peerAuthn.Namespace == config.Get().ExternalServices.Istio.RootNamespace {
 		enabledCheckers = append(enabledCheckers, peerauthentications.DisabledMeshWideChecker{PeerAuthn: peerAuthn, DestinationRules: m.MTLSDetails.DestinationRules})
 	} else {
 		enabledCheckers = append(enabledCheckers, peerauthentications.DisabledNamespaceWideChecker{PeerAuthn: peerAuthn, DestinationRules: m.MTLSDetails.DestinationRules})
@@ -51,7 +51,7 @@ func (m PeerAuthenticationChecker) runChecks(peerAuthn security_v1beta.PeerAuthe
 	// MeshWide and NamespaceWide validations are only needed with autoMtls disabled
 	if !m.MTLSDetails.EnabledAutoMtls {
 		// PeerAuthentications into istio control plane namespace are considered Mesh-wide objects
-		if peerAuthn.Namespace == config.Get().IstioNamespace {
+		if peerAuthn.Namespace == config.Get().ExternalServices.Istio.RootNamespace {
 			enabledCheckers = append(enabledCheckers,
 				peerauthentications.MeshMtlsChecker{MeshPolicy: peerAuthn, MTLSDetails: m.MTLSDetails, IsServiceMesh: false})
 		} else {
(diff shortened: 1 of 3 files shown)


business/checkers/peer_authentication_checker.go
```diff
@@ -42,7 +42,7 @@ func (m PeerAuthenticationChecker) runChecks(peerAuthn security_v1beta.PeerAuthe
 		matchLabels = peerAuthn.Spec.Selector.MatchLabels
 	}
 	enabledCheckers = append(enabledCheckers, common.SelectorNoWorkloadFoundChecker(PeerAuthenticationCheckerType, matchLabels, m.WorkloadList))
-	if peerAuthn.Namespace == config.Get().IstioNamespace {
+	if peerAuthn.Namespace == config.Get().ExternalServices.Istio.RootNamespace {
```
xeviknal, line 45: Didn't you wanted to use the new method you introduced in the previous PR? I guess it is a good moment to do so.

Same happens to a couple more places.

https://github.com/kiali/kiali/blob/add2f245d9739af879fd8d1fa73287730ef4209b/config/config.go#L871-L873

leandroberetta: so true... fixed!.

xeviknal (approved):
LGTM!

leandroberetta:
The steps to test this are the following:

1. Configure Kiali to a different root namespace, for example "istio-config", this can be done via the CR:

```yaml
apiVersion: kiali.io/v1alpha1
kind: Kiali
metadata:     
  name: kiali  
spec:
  external_services:    
    istio:
      root_namespace: istio-config
```

2. Wait for the operator to update Kiali configmap 

3. Create the following global PeerAuthentication resource:

```yaml
kind: PeerAuthentication
metadata:
  name: default
spec:
  mtls:
    mode: STRICT
```

4. Expect to see in the overview page that every namespace (bookinfo for example) has mTLS configured:
 
mtls_on

prachiyadav:
Getting error on KIALI side that istio-config namespace is not accessible to istio system. I am using OCP 4.9 for my testing.

Screenshot from 2021-11-17 19-18-35

prachiyadav (changes requested):
I can not see istio-config in accessible-namespace.... I tried adding the same.... then it worked fine

xeviknal:
@prachiyadav :1st_place_medal: by any chance do you still have the kiali log available? Could you share it with us?
@leandroberetta  Should we incorporate the `rootNamespace` as part of the accessible-namespace?

leandroberetta:
> @prachiyadav 1st_place_medal by any chance do you still have the kiali log available? Could you share it with us? @leandroberetta Should we incorporate the `rootNamespace` as part of the accessible-namespace?

I guess that yes, Kiali will require permissions to get configuration from that namespace and the operator is the responsible for create all the RBAC for the accessible namespaces.

So, it should be OK to add the root namespace (if it is configured) in the list of namespaces, automatically.

In our tests, with minikube, the value for the accessible namespaces was ** so it worked, but here @prachiyadav tested on OpenShift that is more restricted.

What do you think?.

prachiyadav:
@xeviknal  attaching kiali logs
[kiali-56cc6846f6-g5gnm-kiali.log](https://github.com/kiali/kiali/files/7555600/kiali-56cc6846f6-g5gnm-kiali.log)

jmazzitelli:
> In our tests, with minikube, the value for the accessible namespaces was ** so it worked, but here @prachiyadav tested on OpenShift that is more restricted.

See my comments https://github.com/kiali/kiali/pull/4470#issuecomment-971671805 and https://github.com/kiali/kiali/pull/4470#issuecomment-971680519

leandroberetta:
@prachiyadav can you try again this with the latest change on the operator (in master branch).

Now the root namespace should be included in the accessible_namespaces attribute.

prachiyadav (approved):

prachiyadav:
Test case needs to be added

prachiyadav:
Test Case >> SWSQE-2171

PR #4496, opened 2021-11-11
Use goimports for linting in CI

### LINKED ISSUE #4442: GH workflow build - CI gives a green light even when the gofmt says in not well formated
Is it only me that have this bunch of modification after running the make format command (gofmt) in master?

I see this problem for example:

Pipeline:
https://github.com/kiali/kiali/runs/3938401801?check_suite_focus=true#step:6:1

PR with unsorted imports:
https://github.com/kiali/kiali/pull/4439/files#diff-3b471dbe35509f564d7f4429405e67504bc12fdd1d69b374077f7090519f1550R9

nrfox:
Adds `goimports` to the list of linters run by CI. This detects that the imports are sorted properly for most cases.

Fixes #4442

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -2,17 +2,15 @@ name: Kiali Go CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
 
 jobs:
-
   build:
     name: Build
     runs-on: ubuntu-20.04
     steps:
-
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
@@ -26,12 +24,6 @@ jobs:
 
       - name: Lint Install
         run: make lint-install
-
-      - name: Gofmt
-        run: |
-          GO_FILES=$(find . -iname '*.go' -type f | grep -v /vendor/) # All the .go files, excluding vendor/
-          if [ ! -z "$(gofmt -l ${GO_FILES})" ]; then echo "These files need to be formatted:" "$(gofmt -l ${GO_FILES})";echo "Diff files:"; gofmt -d ${GO_FILES}; exit 1; fi # Gofmt Linter
-          echo "Done gofmt"
 
       - name: Verify code linting
         run: make lint
--- a/.github/workflows/config/.golangci.yml
+++ b/.github/workflows/config/.golangci.yml
@@ -1,0 +1,7 @@
+linters:
+  enable:
+    - goimports
+
+linters-settings:
+  goimports:
+    local-prefixes: github.com/kiali/kiali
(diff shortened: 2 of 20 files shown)


xeviknal:
It works to me. I think we would keep all the imports better sorted. However, it just fixes the block separation. I think it can't put the first block for the built-in, second to N-1 block for external imports, last one for kiali ones.
But I am happy with this, helps a lot. it was the aim of the issue that I created in the first place.

nrfox:
Not sure why CI is failing there...

lucasponce:
I wonder if we can find a way to:
- Delegate in the IDE to organize the imports, it's very typical that IDEs automatically add/update dependencies.
- The organize provided by the IDE (VSCode, Goland) is good enough and we don't need an extra task to organize it.
- +1 to add it to the CI to check that this is not scaped.

What I'd like to avoid is that we will have a bunch of extra commits "Update the goimports" as it happens with swagger.

nrfox:
> What I'd like to avoid is that we will have a bunch of extra commits "Update the goimports" as it happens with swagger.

+1 this change just adds a check to CI that fails if things aren't formatted properly. If we wanted to perform the formatting automatically imo the right place to do that is in a pre commit hook like how the UI does it. Otherwise we can just leave it up to the deveoper's IDE like you said.

lucasponce:
@nrfox I'm +1 with this PR, I think we could consolidate it when conflicts are rebased.

These settings can be refined when we start to use it.

nrfox:
> Do we want to consider adding go-staticcheck to this PR (or a follow up PR)?? See #4614

staticheck is actually [enabled by default](https://golangci-lint.run/usage/linters/#enabled-by-default-linters) in golangci-linter. When I run on the latest version of golangci-linter I get the errors seen in #4614 but Kiali is using v1.23.8 of golangci-linter and when I run with that version I don't get any errors. I'd say let's just bump the version of `golangci-linter` in a separate PR (possibly in #4614) whenever we are ready to address the issues that it reports otherwise CI tests will fail.

.github/workflows/kiali-ci.yml
```diff
@@ -27,12 +25,6 @@ jobs:
       - name: Lint Install
         run: make lint-install
 
-      - name: Gofmt
```
lucasponce, line 30: Is the "gofmt" included in the "goimports" linter?
I reviewed the discussion but not sure if that was commented on before.

nrfox: > Is the "gofmt" included in the "goimports" linter?
> I reviewed the discussion but not sure if that was commented on before.

Yep! From the [goimports docs](https://pkg.go.dev/golang.org/x/tools/cmd/goimports#pkg-overview):

> In addition to fixing imports, goimports also formats your code in the same style as gofmt so it can be used as a replacement for your editor's gofmt-on-save hook.

lucasponce (approved):
LGTM
Tested locally using a 
```
make lint
```
and everything looks ok.

I don't think this PR would need QE flow, so we can progress on it @nrfox if you feel ok.

PR #4497, opened 2021-11-12
[hack] wait for ibm cloud cluster to completely delete before continuing; bump to openshift 4.8

jmazzitelli:
main reason for this pr is to fix the delete command - this fixes things so it can now completely delete everything on the first attempt (before you would have to run the script multiple times to get everything to delete).

Other things this PR does - installs OS 4.8 by default and defines a simple description for the apikey.


jshaughn (approved):
LOK

Issue #4486, opened 2021-11-08
k8s service appProtocol is no reflected in config checks


frankconrad:
k8s service appProtocol is no reflected in config checks (KIA0601)

For this nned to add the support of service appProtocol in config checks.
To match istio support there.

KIA0601 is related to IST0118.

szelenka:
I'm also running into this issue. Where the services have appProtocol defined, according to Istio's documentation:
https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

But Kiali marks the config as invalid:
https://kiali.io/docs/features/validations/#kia0601---port-name-must-follow-protocol-suffix-form

The Istio services still work, it's just an oversight in the Kiali checks AFAIK

nrfox:
> The Istio services still work, it's just an oversight in the Kiali checks AFAIK

Agreed I'll let @xeviknal weigh in but it looks like the service validation checks don't currently support the `appProtocol` field and just look at the suffix.

hhovsepy:
This issue is resolved in Kiali v1.44.

PR #4499, opened 2021-11-12
While validating port name in Services, consider also appProtocol parameter's value.

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4486

The https://kiali.io/docs/features/validations/#kia0601---port-name-must-follow-protocol-suffix-form validation was checking only the port name in protocol [-suffix] format, but name can be not in that format and instead appProtocol parameter has the format of protocol. As described in https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection.

So now these both formats of ports are correct for Services. 
```
  ports:
  - port: 9080
    name: http-correct
  - port: 3306
    name: database
    appProtocol: http
```

### DIFF
--- a/business/checkers/services/port_mapping_checker.go
+++ b/business/checkers/services/port_mapping_checker.go
@@ -28,6 +28,11 @@ func (p PortMappingChecker) Check() ([]*models.IstioCheck, bool) {
 		for portIndex, sp := range p.Service.Spec.Ports {
 			if strings.ToLower(string(sp.Protocol)) == "udp" {
 				continue
+			} else if sp.AppProtocol != nil {
+				if !kubernetes.MatchPortAppProtocolWithValidProtocols(sp.AppProtocol) {
+					validation := models.Build("port.appprotocol.mismatch", fmt.Sprintf("spec/ports[%d]", portIndex))
+					validations = append(validations, &validation)
+				}
 			} else if !kubernetes.MatchPortNameWithValidProtocols(sp.Name) {
 				validation := models.Build("port.name.mismatch", fmt.Sprintf("spec/ports[%d]", portIndex))
 				validations = append(validations, &validation)
--- a/models/port.go
+++ b/models/port.go
@@ -4,10 +4,10 @@ import core_v1 "k8s.io/api/core/v1"
 
 type Ports []Port
 type Port struct {
-	Name        string `json:"name"`
-	Protocol    string `json:"protocol"`
-	AppProtocol string `json:"appProtocol"`
-	Port        int32  `json:"port"`
+	Name        string  `json:"name"`
+	Protocol    string  `json:"protocol"`
+	AppProtocol *string `json:"appProtocol,omitempty"`
+	Port        int32   `json:"port"`
 }
 
 func (ports *Ports) Parse(ps []core_v1.ServicePort) {
@@ -21,9 +21,7 @@ func (ports *Ports) Parse(ps []core_v1.ServicePort) {
 func (port *Port) Parse(p core_v1.ServicePort) {
 	port.Name = p.Name
 	port.Protocol = string(p.Protocol)
-	if p.AppProtocol != nil {
-		port.AppProtocol = *p.AppProtocol
-	}
+	port.AppProtocol = p.AppProtocol
 	port.Port = p.Port
 }
 
--- a/models/port.go
+++ b/models/port.go
@@ -4,9 +4,10 @@ import core_v1 "k8s.io/api/core/v1"
 
 type Ports []Port
 type Port struct {
-	Name     string `json:"name"`
-	Protocol string `json:"protocol"`
-	Port     int32  `json:"port"`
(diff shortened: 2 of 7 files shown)


hhovsepy:
@xeviknal here the open question is, that 'appProtocol' is not displayed in Service details page. So if we add a new validation message for 'appProtocol' field's validation, then we need to display it in Service page as well.

hhovsepy:
UI PR required: https://github.com/kiali/kiali-ui/pull/2271
Docs PR required: https://github.com/kiali/kiali.io/pull/476

hhovsepy:
How to test it:

Create a Service with the following value of appProtocol.
```
apiVersion: v1
kind: Service
metadata:
  name: ratings-java-svc-suffix
  namespace: bookinfo
  labels:
    app: ratings-java
    service: ratings-svc
spec:
  ports:
  - port: 9080
    name: http
  - port: 3306
    name: database
    appProtocol: wrong
  selector:
    app: ratings-java
    version: v1
```
And on Service Details page it will show Error with the messages:
Screenshot from 2021-11-15 11-49-29
Screenshot from 2021-11-15 11-49-32

Change the appProtocol to a supported protocol, for instance to "mysql" and the Error will be gone.

hhovsepy:
Screenshot from 2021-11-15 16-53-15

hhovsepy:
The swagger error happens also on master version of swagger.json.

business/checkers/services/port_mapping_checker.go
```diff
@@ -28,6 +28,11 @@ func (p PortMappingChecker) Check() ([]*models.IstioCheck, bool) {
 		for portIndex, sp := range p.Service.Spec.Ports {
 			if strings.ToLower(string(sp.Protocol)) == "udp" {
 				continue
+			} else if sp.AppProtocol != nil {
+				if !kubernetes.MatchPortAppProtocolWithValidProtocols(sp.AppProtocol) {
```
xeviknal, line 32: Nitpick here. You might want to join this if with the previous one since you already check the nil within the `MatchPortAppProtocol` method.

models/port.go
```diff
@@ -4,9 +4,10 @@ import core_v1 "k8s.io/api/core/v1"
 
 type Ports []Port
 type Port struct {
-	Name     string `json:"name"`
-	Protocol string `json:"protocol"`
-	Port     int32  `json:"port"`
+	Name        string `json:"name"`
+	Protocol    string `json:"protocol"`
+	AppProtocol string `json:"appProtocol"`
```
xeviknal, line 9: You might want to add here the `omitempty` tag. So if it is nil, the field won't be part of the json. 
In the front-end you already have the field as optional `?` so shouldn't be a problem.

models/port.go
```diff
@@ -20,6 +21,9 @@ func (ports *Ports) Parse(ps []core_v1.ServicePort) {
 func (port *Port) Parse(p core_v1.ServicePort) {
 	port.Name = p.Name
 	port.Protocol = string(p.Protocol)
+	if p.AppProtocol != nil {
```
xeviknal, line 24: I am having a doubt in here about what's better:
1) Convert the *string to string in here and do the if statement to check the if (as you have now), or
2) Move the Port.AppProtocol field to *string and let the `json.Marshal` ([here](https://github.com/kiali/kiali/blob/87ec70e9309b04b4439b95882b516b1c0ee6ba39/handlers/base.go#L13-L23)) convert the nil to string (which is nil).

Something like this:
```golang
type Port struct {
    AppProtocol *string `json:"appProtocol,omitempty"`
}

func (port *Port) Parse(p core_v1.ServicePort) {
	port.Name = p.Name
	port.Protocol = string(p.Protocol)
        port.Port = p.Port
        port.AppProtocol = p.AppProtocol
}
```

I am happy with both ways. What do you think?

xeviknal (dismissed):
LGTM! I have a couple of comments that I'd like to hear from hayk though. Nothing blocker!

business/checkers/services/port_mapping_checker.go
hhovsepy: if I join these 2 'if', then 'else' will not be clear whether appProtocol is nil, empty or exist but with wrong format.

models/port.go
hhovsepy: agree

models/port.go
hhovsepy: +1 for the second solution

xeviknal (approved):

skondkar (approved):
LGTM.
Service Details page shows error messages:

Screenshot from 2021-11-22 14-11-58

Screenshot from 2021-11-22 14-12-04

After changing the appProtocol to a supported protocol, and changing the name of protocol, there is no error message.

Screenshot from 2021-11-22 14-15-41

skondkar:
Polarion test case ->> SWSQE-2166

Issue #4488, opened 2021-11-09
Correct graph edge for Pod to Pod communication using destination_workload


MugenTwo:
As suggested by @jmazzitelli in the following discussion: https://github.com/kiali/kiali/discussions/4454 I'm creating this feature request ticket.

### **Background**
I have a microservice architecture in which my Spring Boot Admin Server's pod to pod communication is being represented as PassthroughCluster Node even though **destination_workload** is actually a known entity/app inside the system. 

This makes it so that the traffic is hard to understand because the PassthroughCluster adds a lot of **noise** in the graph, it makes it difficult to do **traffic classification** as I can't identify which traffic is actually an external traffic and which traffic is actually coming from the Spring Boot Admin Server.

### **Use cases**
- Classify traffic of "Spring Boot Admin" type of services which does pod to pod communication for healthchecks, resource monitoring, ...
- Classify traffic for client-side load balancing (e.g. Ribbon). -> more important than dealing with Spring Boot Admin traffic.

### **Suggested behaviour**
Draw and edge between app nodes to represent pod to pod communication with a **new type of edge (maybe different color) by taking into account the destination_workload** instead of of the destination_service_name if the former is a known name and the latter is a PassthroughCluster.

jmazzitelli:
The edges still represent the same kind of data (http, or tcp, or gcp) so this would not warrant a different edge type. Edges are not different based on the endpoint node types.

Because the target node is a workload and not an app (represented with a square node) or service (represented with a triangle node), the graph should simply draw a circle node which represents "workload" as per the graph legend.

MugenTwo:
Agreed on the part that this should be pointing on the correct workload. This is definitely the most important part.

I suggested having a new type -> new color of edge to easily classify the pod to pod communication from the other types of communication... similar to how TCP is being classified with a blue color.

jmazzitelli:
> I suggested having a new type -> new color of edge to easily classify the pod to pod communication from the other types of communication... similar to how TCP is being classified with a blue color.

Yeah, that's the problem. From a UX perspective, you are mixing semantics. Today color/edge type denotes the different protocols (http vs. tcp) flowing over the edge. Suggesting changing color/edge type based on the destination node makes it confusing because now the semantics are a combination of protocol and node type. Think about it this way - what if you have 2 of these specific pod communications - one is sending HTTP traffic to the pod, one is sending TCP traffic to the pod. What color/edge type do you make those edges? Things become ambiguous and confusing to the user.

[...]

PR #4500, opened 2021-11-12
Omit intermediate PassthroughCluster nodes in the graph

jshaughn:
Fixes https://github.com/kiali/kiali/issues/4488

For direct pod requests that travel through the default handler (PassthroughCluster),
improve the visualization by omitting the intermediate PassthroughCluster service node,
thus leaving only the direct edge.  Note, this is only relevant when
service injection is enabled.  Note also that it is still possible, and
desirable, to have PassthoughCluster as a terminal node.

### DIFF
--- a/graph/api/api_test.go
+++ b/graph/api/api_test.go
@@ -3508,6 +3508,24 @@ func TestMultiClusterSourceGraph(t *testing.T) {
 		"source_cluster":                 "kukulcan",
 		"source_workload":                "productpage-v1",
 		"source_workload_namespace":      "bookinfo"}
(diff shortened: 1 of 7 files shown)


jshaughn:
I'm going to promote this PR for dev/qe review.  https://github.com/MugenTwo has not been able to provide feedback from the field but I think it's a good change.

nrfox (approved):

jshaughn:
@mattmahoneyrh I don't think it's necessary to recreate this pod-to-pod scenario for testing, but rather ensure no graph regessions for the graph types, with Show Services enabled and disabled, as well as with Response time and throughput labels, and the Show Security option enabled.

mattmahoneyrh (approved):

PR #4504, opened 2021-11-17
Add rank query to find defaults

nrfox:
Related to: https://github.com/kiali/kiali-operator/pull/452

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -592,7 +592,7 @@ func NewConfig() (c *Config) {
 							Expression:  "name = unknown",
 						},
 						{
-							Description: "Find: nodes with the 2 highest ranks",
+							Description: "Find: nodes with the 2 top rankings",
 							Expression:  "rank <= 2",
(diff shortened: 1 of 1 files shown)


israel-hdez:
@nrfox I wonder if it would be useful to add a similar hide expression?
My rationale is that if people is trying to find the most important nodes, perhaps it is better to hide from the graph what makes noise.

Something like "Hide: all nodes but the highest 2 rankings"?

nrfox:
> @nrfox I wonder if it would be useful to add a similar hide expression? My rationale is that if people is trying to find the most important nodes, perhaps it is better to hide from the graph what makes noise.
> 
> Something like "Hide: all nodes but the highest 2 rankings"?

Sure that'd be useful to show. I'll add that.

israel-hdez (approved):
LGTM

mattmahoneyrh (approved):

PR #4516, opened 2021-11-19
Update to latest k8s.io/client-go

### LINKED ISSUE #4042: Upgrade kubernetes/client-go version and update beta interfaces for workloads
A lot of warnings on Kubernetes 1.21 clusters like this:

```
W0524 13:55:34.638105  209600 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```

Worth to review the beta apis used for k8s in Kiali

israel-hdez:
My quick smoke tests look OK. Also, unit tests are passing.

In #4042 it is also mentioned to migrate to the batch/v1 api of CronJobs. First k8s version where this API is available is in v1.21. So, I don't think we should move to that new API until k8s versions prior to 1.21 reach EOL. This will happen on Feb 28th, 2022 for Kubernetes, but for OpenShift that's apparently much later. So, I think we will need to wait until August 2022 when k8s v1.25 will be released and the old API will be removed. We may need to create a separate issue for that.

Fixes #4042


jmazzitelli (approved):
`make clean build test` finishes successfully.

pbajjuri20:
API test cases run successfully against this PR

pbajjuri20 (approved):

Issue #4458, opened 2021-10-27
KIA1105: Virtual service routes may not point to any subset


xeviknal:
In Virtual services the subset destination is not mandatory. You might be sending traffic to different services without specifiying any subset (labeling). 

Therefore, `KIA1105` is not accurate enough and throws a warning. The validation it should include the pair of host+subset instead of only considering the subset.

See one example:
![Screenshot of Kiali Console (29)](https://user-images.githubusercontent.com/613814/139106377-d0596451-f8a3-4885-b1b7-409cec745ca0.png)

Here's the response. The validation points to a field that doesn't exist.
![Screenshot of localhost_20001_kiali_api_namespaces_default_istio_virtualservices_rollouts-demo-vsvc2_validate=true](https://user-images.githubusercontent.com/613814/139106342-0e30c6e7-9031-40ad-84b4-8669b2b93e5c.png)

lucasponce:
Closed by https://github.com/kiali/kiali/pull/4520

PR #4520, opened 2021-11-22
While validating VS subsets repeatings, consider hosts and subset combinations

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4458.

Create a VS without a subsets in route, like this:
```
      route:
        - destination:
            host: reviews
          weight: 80
        - destination:
            host: reviews.bookinfo2.svc.cluster.local
          weight: 10
        - destination:
            host: reviews.bookinfo.svc.cluster.local
          weight: 10
```

Before it was showing warning message in wrong place because validation was tightened to only subset:
Screenshot from 2021-11-23 10-02-14

Now it validates host+subset combination and considers host in fqdn and puts validation message on the host line:
Screenshot from 2021-11-23 09-06-35

### DIFF
--- a/business/checkers/virtualservices/route_checker.go
+++ b/business/checkers/virtualservices/route_checker.go
@@ -97,7 +97,7 @@ func (route RouteChecker) checkTcpRoutes() ([]*models.IstioCheck, bool) {
 			}
 		}
 
-		route.trackTcpTlsSubset(routeIdx, "http", destinationWeights, &validations)
+		route.trackTcpTlsSubset(routeIdx, "tcp", destinationWeights, &validations)
 	}
 
 	return validations, valid
@@ -128,7 +128,7 @@ func (route RouteChecker) checkTlsRoutes() ([]*models.IstioCheck, bool) {
 			}
 		}
 
-		route.trackTcpTlsSubset(routeIdx, "http", destinationWeights, &validations)
+		route.trackTcpTlsSubset(routeIdx, "tls", destinationWeights, &validations)
 	}
 
 	return validations, valid
--- a/business/checkers/virtualservices/route_checker.go
+++ b/business/checkers/virtualservices/route_checker.go
@@ -6,10 +6,13 @@ import (
 	api_networking_v1alpha3 "istio.io/api/networking/v1alpha3"
 	networking_v1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
 
+	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 )
 
 type RouteChecker struct {
+	Namespace      string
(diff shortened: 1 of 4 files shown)


hhovsepy:
For QE, it needs to be tested with http, tcp and tls routes, also with and without subsets.

hhovsepy:
Doc changes https://github.com/kiali/kiali.io/pull/489

models/istio_validation.go
```diff
@@ -303,7 +303,7 @@ var checkDescriptors = map[string]IstioCheck{
 	},
 	"virtualservices.route.repeatedsubset": {
 		Code:     "KIA1105",
-		Message:  "This subset is already referenced in another route destination",
+		Message:  "This host subset combination is already referenced in another route destination",
```
lucasponce, line 306: Minor rewording, "host subset" -> "host/subset" ?

business/checkers/virtualservices/route_checker.go
```diff
@@ -125,13 +128,13 @@ func (route RouteChecker) checkTlsRoutes() ([]*models.IstioCheck, bool) {
 			}
 		}
 
-		trackTcpTlsSubset(routeIdx, "http", destinationWeights, &validations)
+		route.trackTcpTlsSubset(routeIdx, "http", destinationWeights, &validations)
```
lucasponce, line 131: Here the "http" -> "tls", not for this PR, but it was a bug from the refactoring to use the Istio client, on this case, it's checking the tls routes and the string is used to check collissions.

models/istio_validation.go
hhovsepy: it is used 'host subset' also in "destinationrules.multimatch", might be changing both?

business/checkers/virtualservices/route_checker.go
```diff
@@ -94,7 +97,7 @@ func (route RouteChecker) checkTcpRoutes() ([]*models.IstioCheck, bool) {
...
```
lucasponce, line 100: Here the "http" -> "tcp" I think

models/istio_validation.go
lucasponce: Nah, it's very minor, I'm also fine with the wording and if it's used in other places I don't see a I need to change.

lucasponce (approved):

skondkar (approved):
LGTM.

Validation with http route without subset:
Screenshot from 2021-12-01 18-21-24

No message if http route is with subset:
Screenshot from 2021-12-01 18-35-17

Also verified with tcp and tls routes, with and without subsets.

Screenshot from 2021-12-01 18-44-00

Screenshot from 2021-12-01 18-55-24

skondkar:
Polarion test case >> SWSQE-2168

PR #4526, opened 2021-11-23
Fix Maistra injection script

leandroberetta:
The patch adds an annotation instead of a label.


jmazzitelli (approved):
yes, this is correct - maistra only supports the annotation today.

Issue #4527, opened 2021-11-25
Possible memory leak in /api/istio/status endpoint


FilipB:
When hitting /api/istio/status endpoint kiali pod memory usage grows steadily until it reaches 1GB limit and the pod is OOMKiled.

### Version
Kiali UI
    1.44.0 (e3238fb2f2493b23540125d4e15d4665e608dc8c)
Kiali Server
    v1.44.0-SNAPSHOT (e9a4b6f469d9bb1a0eef3bfabe5b0a9e4a83ffcd)
Kiali Container
    v1.44.0-SNAPSHOT
Istio
    1.12.0 
Prometheus
    2.26.0 
Kubernetes
    v1.22.2+5e38c72 

### Repro steps
1. install istio and kiali to istio-system namespace
2. oc login to your cluster
3. java must be available (e.g dnf install -y java-11-openjdk)
4. run attached repro.sh from /tmp dir
5. watch kiali pod memory usage -> it grows very fast until it reaches 1GB limit
6. then the kiali pod is OOMKilled and restarted and running repro test is showing 503 response codes

Attaching as txt as sh is not supported. Please rename to repro.sh
[repro.sh.txt](https://github.com/kiali/kiali/files/7604704/repro.sh.txt)

FilipB:
When removing the pod memory limit it grows to 6GB after 10 minutes. It's growing linearly.

lucasponce:
Thanks @FilipB, I'll investigate this endpoint, it sounds a clear leak from your comments.

lucasponce:
@FilipB these are the steps I did for the investigation.

I used latest "master" Kiali code but on a CRC Maistra 2.1/Istio 1.9 environment, I guess that if the problem is a leak, the version of Istio shouldn't affect here.

First step was to test the endpoint using an "anonymous" strategy, and with a similar rate of concurrent users, the backend was stable in ~150 Mb of RAM, also the garbage collector seemed to work as expected.

Second step, I tested the "openshift" strategy as it requires additional steps and logic that could be affected to this leak and same results (I used a manual loop of curl to simulate users).

![image](https://user-images.githubusercontent.com/1662329/143564813-661e7fac-958f-4ce6-b342-ca3c9cccafdb.png)

![image](https://user-images.githubusercontent.com/1662329/143564748-a104a867-0b0a-4500-9612-90f6666bc4ef.png)

[...]

PR #4530, opened 2021-11-29
Cache Jaeger client

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4527

### DIFF
--- a/business/layer.go
+++ b/business/layer.go
@@ -38,6 +38,7 @@ type Layer struct {
 
 // Global clientfactory and prometheus clients.
 var clientFactory kubernetes.ClientFactory
+var jaegerClient jaeger.ClientInterface
 var prometheusClient prometheus.ClientInterface
 var once sync.Once
 var kialiCache cache.KialiCache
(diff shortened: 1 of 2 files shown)


lucasponce:
@FilipB would you mind to test this PR ?

Checking the jaeger logs, it seems that non reusing the grpc client may saturate easily the grpc transport in the server.

The PR is not completed but I'm curious if you can give it a try.

FilipB:
The memory usage is around 400MB and stable. Trying to reproduce the original issue in master now to be sure changes in this PR are responsible for stable memory usage.

business/layer.go
```diff
@@ -102,7 +103,11 @@ func Get(authInfo *api.AuthInfo) (*Layer, error) {
 
 	// Create Jaeger client
 	jaegerLoader := func() (jaeger.ClientInterface, error) {
-		return jaeger.NewClient(authInfo.Token)
+		var err error
+		if jaegerClient == nil {
+			jaegerClient, err = jaeger.NewClient(authInfo.Token)
```
FilipB, line 108: Does this token expire? If yes the client will fail to communicate after expiration and new one will not be created?

FilipB (commented):
This PR fixes the memory leak problem. Memory usage is stable around 400 MB

business/layer.go
lucasponce: I'm not sure if this feature was ever used.
I've sent a patch to ensure that if the client fails, it's not cached, so in the case of token expiration a following call should populate refresh the client.

In downstream this is not used, as it has an internal user, and in upstream prometheus is not protected with auth as well, so, perhaps it would be needed more work in those cases, but it wouldn't modify the patch.

If reusing the jaeger works well, I think that patch looks good.

FilipB: I will keep it running over night and approve in the morning if it's working fine.

FilipB (approved):
It's running all day without any error.

leandroberetta (approved):
LGTM.

PR #4531, opened 2021-11-29
Fix syntax error on Jenkinsfile

israel-hdez:
Also, remove reference to another job that is now unused/non-existent.


jshaughn (approved):
LGTM

PR #4532, opened 2021-11-30
Instrument kiali server with tracing

### LINKED ISSUE #4036: Instrument Kiali server with Jaeger
During the revisiting of #3244 the idea of replace some internal metrics with instrumented code arose.

The motivation is to have more information in cases of performance issues reported by users.

The traces will be shown in the Traces tab at App/Workload/Service level as for any other microservice.

nrfox:
The two largest changes with this PR:

1. Addition of opentelemetry client libraries
2. Passing request context through handlers to the business layer
3. Making public API of business services an `Interface`

Tracing has only been added to a small subset of routes but more can easily be added later.

To test, follow the directions in the [README](https://github.com/nrfox/kiali/blob/add-tracing/README.adoc#enable-tracing). Once the Kiali server is configured correctly you should be able to see traces and spans for the Kiali endpoints through the Jaeger UI.

Fixes #4036


README.adoc
```diff
@@ -226,6 +226,24 @@ make lint-install
...
+----
+apiVersion: kiali.io/v1alpha1
+kind: Kiali
+metadata:
+  name: kiali
+spec:
+...
+  server:
+    collector_url: http://jaeger-collector.istio-system:14268/api/traces
+    tracing_enabled: true
```
jmazzitelli, line 243: Let's think about this some more. Whenever I see multiple settings dealing with the same type of feature, I would like to see them grouped under the same parent heading. Otherwise, when we start adding things we start getting N related settings all spread out in the config and its confusing to users. I already saw this happen that caused me to refactor the config recently because I was adding stuff to the ingress config and it was getting confusing by spreading settings around.

We already have [metrics](https://github.com/kiali/kiali-operator/blob/v1.43.0/deploy/kiali/kiali_cr.yaml#L992-L999) settings that should also be grouped, maybe we can move those and put everything all under the same heading now.

```yaml
spec:
  server:
    observability:
      metrics:
        enabled: true
        port: 9090
      tracing:
        collector_url: http://jaeger-collector.istio-system:14268/api/traces
        enabled: true
```

jmazzitelli (changes requested):

README.adoc
nrfox: That sounds good to me. It would make it easier and more clear to add settings related to each of these settings in the future. What do you think about removing `enabled` and just having the presence of `tracing` mean that it is enabled?

```
spec:
  server:
    observability:
      tracing: {}
```
would mean tracing is enabled with only defaults. Whereas omitting tracing
```
spec:
  server:
    observability:
      metrics:
        ...
```
would mean that tracing is not enabled. One less setting to add or toggle.

tracing/tracing.go
```diff
@@ -0,0 +1,54 @@
+// The tracing package provides utilities for the Kiali server
+// to instrument itself with tracing to provide better insights
+// into server performance. Currently only integrated with Jaeger.
+package tracing
```
lucasponce, line 4: Naming is hard :) and no good answers at all, but I find a little bit confusing to have a "tracing" package given the context that we deal with external services prom and jaeger as both sources and also used for instrumentation.

I saw a comment in that way related the configuration, perhaps, the internal observability aspects may go to a single location (observability, or synced with the name you choose there).

README.adoc
jmazzitelli: > What do you think about removing enabled

That would be against our convention. We use "enabled" for everything else. It makes it very explicit and more understandable.

Also, its easier on the implementation side - "true" or "false" is easier to deal with than "exists" or "does not exist". (think about disabling it - the user has to patch the YAML with "null" when removing settings from k8s resources and hope the merge accepts it. I know I have trouble sometimes getting the kubectl command correct when trying to patch something and REMOVE an item). It is much easier just having a boolean value.

nrfox: > That would be against our convention. We use "enabled" for everything else. It makes it very explicit and more understandable.

:+1: 

I'm generally in favor of less nesting instead of more, how about keeping `tracing` and `metrics` under the `server` header?
```
spec:
  server:
      metrics:
        enabled: true
        port: 9090
      tracing:
        collector_url: http://jaeger-collector.istio-system:14268/api/traces
        enabled: true
```

tracing/tracing.go
nrfox: I was a little unsure of where to put this package or name it when I created it. There's already a `jaeger` package for querying jaeger directly and a `prometheus` package for the same. Since this mostly adds just opentelemetry tracing, there is very little jaeger specific code. `observability` sounds fine to me. Internal metrics/logs integration with otel can be put there if added in the future.

README.adoc
nrfox: Also does this change need to be backwards compatible? i.e. should we keep the `metrics_enabled` field and deprecate it to be removed at a later date or no?

nrfox: > how about keeping tracing and metrics under the server header?

On second thought, if the otel integration is moving under an `observability` package, it makes sense to have this as part of the config.

jmazzitelli: > Also does this change need to be backwards compatible? i.e. should we keep the metrics_enabled field and deprecate it to be removed at a later date or no?

Right now - unless I can think of a reason otherwise - I am thinking don't worry about it. The only reason why I had to add support for the backward-compatible ingress setting was because the Maistra/OSSM operator was setting that deprecated setting, so I wanted to maintain support for it. But OSSM operator (AFAIK) does NOT set this server.metrics_enabled setting - so I don't see a need to maintain backward compatibility for it. It would be easy to add if it turns out there is a reason for it, but as of right now I don't see a need to add that technical debt to the code. This is a minor feature that even if someone missed it by using the old setting, nothing horrible will happen and its easily fixed on the user's end.

nrfox: After the refactoring, this will require some operator changes. I'll push those shortly.

README.adoc
```diff
@@ -226,6 +226,26 @@ make lint-install
 make format lint
 ----
 
+=== Enable tracing
+
+Kiali itself is instrumented with opentelemetry tracing to help provide insights and surface performance issues for the kiali server. To enable, set the `server.tracing_enabled` and `server.collector_url` configuration options.
```
jmazzitelli, line 231:
```suggestion
Kiali itself is instrumented with opentelemetry tracing to help provide insights and surface performance issues for the kiali server. To enable, set the `server.observability.tracing.enabled` and `server.observability.tracing.collector_url` configuration options.
```

server/server.go
```diff
@@ -8,29 +8,44 @@ import (
...
 // Start and Stop it with the corresponding functions.
 func NewServer() *Server {
 	conf := config.Get()
-	// create a router that will route all incoming API server requests to different handlers
 
+	// create a router that will route all incoming API server requests to different handlers
 	router := routing.NewRouter()
+	var tracingProvider *sdktrace.TracerProvider
+	if conf.Server.Observability.Tracing.Enabled {
+		log.Debugf("Tracing Enabled. Initializing tracer with collector url: %s", conf.Server.Observability.Tracing.CollectorURL)
```
lucasponce, line 36: Minor, I'd like to see this log entry always, perhaps moving to "Info" can be a good idea.

business/istio_config.go
```diff
@@ -401,7 +403,12 @@ func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (
 // - "namespace": 		namespace where configuration is stored
 // - "objectType":		type of the configuration
 // - "object":			name of the configuration
-func (in *IstioConfigService) GetIstioConfigDetails(namespace, objectType, object string) (models.IstioConfigDetails, error) {
+func (in *IstioConfigService) GetIstioConfigDetails(ctx context.Context, namespace, objectType, object string) (models.IstioConfigDetails, error) {
+	if config.Get().Server.Observability.Tracing.Enabled {
+		var span trace.Span
+		ctx, span = otel.Tracer(observability.TracerName).Start(ctx, "GetIstioConfigDetails")
```
jmazzitelli, line 409: In all these `otel.Tracer.Start` calls, I see the name of the function passed in to label the trace (here is `GetIstioConfigDetails`). 

Should these strings have additional context? Because what if other packages have the same (or very similar) names? Down below I see `getPermissions` and that seems like a name that might be identical to other package functions.

So should we have a convention to prefix the function name with the package name? i.e.

```suggestion
		ctx, span = otel.Tracer(observability.TracerName).Start(ctx, "business.GetIstioConfigDetails")
```

config/config.go
```diff
@@ -71,21 +71,38 @@ const (
...
+// Metrics provides metrics configuration for the Kiali server.
+type Metrics struct {
+	Enabled bool `yaml:"enabled,omitempty"`
+	Port    int  `yaml:"port,omitempty"`
+}
+
+// Tracing provides tracing configuration for the Kiali server.
+type Tracing struct {
+	Enabled      bool   `yaml:"enabled,omitempty"`
+	CollectorURL string `yaml:"collector_url,omitempty"` // Endpoint for Kiali server traces
```
jmazzitelli, line 83: trivial - I think we are trying to alphabetize the fields inside all of these config types 
```suggestion
	CollectorURL string `yaml:"collector_url,omitempty"` // Endpoint for Kiali server traces
	Enabled      bool   `yaml:"enabled,omitempty"`
```

jmazzitelli (changes requested):

observability/tracing.go
```diff
@@ -0,0 +1,54 @@
...
+	"go.opentelemetry.io/otel/exporters/jaeger"
+	"go.opentelemetry.io/otel/propagation"
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+)
+
+const (
+	// TracingService is the name of the kiali tracer service.
+	TracingService = "kiali-server"
```
lucasponce, line 20: This shouldn't be a constant - I think - it should have some contextual information about cluster/mesh, to identify the Kiali server, perhaps "kiali-server.istio-system" ? I was thinking in federated / multi-mesh / multi-cluster ? scenarios.

Perhaps a kiali-server.east-mesh-system.clusterA, would be nice, and then it would be prepared for the future.

business/istio_config.go
nrfox: That sounds good to me. It's also possible to set any additional attributes on the trace for more context. This can be stuff like the logged in user, request id, etc. I wasn't entirely sure what would be useful so didn't add any of it in for now but it can be added later. Prefixing with the package name seems like a good convention to start.

config/config.go
nrfox: Good eye... I always miss one of these

kubernetes/kubernetes.go
```diff
@@ -470,13 +475,18 @@ func NewNotFound(name, group, resource string) error {
 }
 
 // GetSelfSubjectAccessReview provides information on Kiali permissions
-func (in *K8SClient) GetSelfSubjectAccessReview(namespace, api, resourceType string, verbs []string) ([]*auth_v1.SelfSubjectAccessReview, error) {
+func (in *K8SClient) GetSelfSubjectAccessReview(ctx context.Context, namespace, api, resourceType string, verbs []string) ([]*auth_v1.SelfSubjectAccessReview, error) {
+	if config.Get().Server.Observability.Tracing.Enabled {
+		var span trace.Span
+		ctx, span = otel.Tracer(observability.TracerName).Start(ctx, "GetSelfSubjectAccessReview")
```
lucasponce, line 481: I'd add custom tags in the Span with the method parameters:
image

In the travels demo those are added in raw, but I guess there should be an API in the span interface.
That would help to identify the offending namespace/resource/object/etc.

observability/tracing.go
nrfox: That's a good suggestion and is more in line with how other services are named by istio by default. It looks like the namespace Kiali is deployed inside of is included in the config so that would be easy to add. The mesh/cluster id might be harder to determine. Kiali may not even be a part of any mesh since it's normal to run Kiali without a sidecar. The cluster id might be able to be fetched from istio or some kube api.

kubernetes/kubernetes.go
nrfox: Yes you can add arbitrary attributes to the spans. I wasn't sure what all would be useful so left them out. Those can easily be added to the traces though. Adding the parameters seems like a good place to start.

handlers/utils.go
```diff
@@ -17,7 +18,7 @@ type promClientSupplier func() (*prometheus.Client, error)
 var defaultPromClientSupplier = prometheus.NewClient
 
 func checkNamespaceAccess(nsServ business.NamespaceService, namespace string) (*models.Namespace, error) {
-	if nsInfo, err := nsServ.GetNamespace(namespace); err != nil {
+	if nsInfo, err := nsServ.GetNamespace(context.TODO(), namespace); err != nil {
```
lucasponce, line 21: I understand that we need to propagate the context from the request to pass the trace context.
I've just wondered if some of this TODO is missing the context.
I'd pass the context in all the cases from the handlers ?

nrfox: Yeah only some of the handlers were updated to pass the context through. I thought it would keep the PR smaller and I wanted to gauge interest on the changes a bit before updating all the handlers at once. I'm happy to go back and update them though as part of this PR.

business/istio_config.go
nrfox: Instead of prefixing the function name, another option would be to include the package as an attribute/tag on the trace. This makes it less obvious when glancing at the span which package the function came from but might allow for better filtering when querying the spans to say something like "give me all spans for this package". I don't see a way to easily search/filter by span name in the jaeger UI.

jmazzitelli: I don't care as much about how its implemented - just that SOMETHING can discern the traces alittle better - having the package name somewhere is good.

nrfox:
I changed the way tracing was being added by moving the tracing logic out of the individual service methods and it a separate struct that wraps the original services. This added a bit more boilerplate but has the nice side effect of keeping the tracing details out of the various services. I opened a [separate PR](https://github.com/nrfox/kiali/pull/2) against this branch so that it's hopefully easier to see the changes but @jmazzitelli @lucasponce let me know if you'd rather it be a separate commit on this branch.

jmazzitelli:
@nrfox can you post some testing procedures? i.e. after building and deploying, what should I click/look at in order to see traces? I assume we can look at the kiali workload for some of this? But detailed testing instructions would help.

nrfox:
> @nrfox can you post some testing procedures? i.e. after building and deploying, what should I click/look at in order to see traces? I assume we can look at the kiali workload for some of this? But detailed testing instructions would help.

Once you enable tracing in the config, you should be able to see traces in the jaeger UI for the kiali-server. You'll probably need to open kiali in your browser at least once to generate some trace data.

How I tested in minikube:
1. Deploy istio/bookinfo
2. Enable tracing in kiali config
```
server:
  observability:
    tracing:
      enabled: true
```
3. Port forward to jaeger UI pod: `kubectl port-forward service/tracing 20002:80 -n istio-system`
4. Open jaeger UI in browser: `http://localhost:20002`

Depending on your Jaeger setup, you may need to set the `server.observability.tracing.collector_url` option as well.

jmazzitelli:
Here is what I see...

image

I see the `package` label on these, too:

image

Question: Why can't I see these traces in Kiali UI itself? If I go to the "traces" tab in the Kiali workload screen, I see no traces:

image

lucasponce:
> Question: Why can't I see these traces in Kiali UI itself? If I go to the "traces" tab in the Kiali workload screen, I see no traces:

As there is no a specification, Kiali expect a common format in the way as traces are collected from the proxies, "kiali-server" is not matching with the app/workload/service name.

jmazzitelli:
> Kiali expect a common format in the way as traces are collected from the proxies, "kiali-server" is not matching with the app/workload/service name.

@nrfox can we reformat the way the traces are collected so they show up in Kiali itself. Seems a little strange that we will now have tracing in Kiali yet Kiali UI itself can't show them (we should fix that so we "eat our own dogfood").

nrfox:
> @nrfox can we reformat the way the traces are collected so they show up in Kiali itself. Seems a little strange that we will now have tracing in Kiali yet Kiali UI itself can't show them (we should fix that so we "eat our own dogfood").

That's a good observation. I'll take a look at how the istio traces are formatted and see if I can get the Kiali ones to match.

observability/tracing.go
```diff
@@ -0,0 +1,63 @@
...
+	"go.opentelemetry.io/otel/sdk/resource"
+	sdktrace "go.opentelemetry.io/otel/sdk/trace"
+	semconv "go.opentelemetry.io/otel/semconv/v1.7.0"
+
+	"github.com/kiali/kiali/config"
+)
+
+const (
+	// TracingService is the name of the kiali tracer service.
+	TracingService = "kiali"
```
jmazzitelli, line 23: I don't know if it is OK for this to be fixed to `kiali` or not, but I do want to draw your attention to the `deployment.instance_name` feature. You can have a Kiali installed with different names, and you can have multiple Kialis deployed in the same namespace. So I don't know if this `TracingService` name should be this `deployment.instance_name` (the default value is `kiali`) or if it is OK for this to be a constant `kiali`.

See docs on this:
* https://github.com/kiali/kiali-operator/blob/v1.44.0/deploy/kiali/kiali_cr.yaml#L358-L373
* https://kiali.io/docs/installation/deployment-options/#kiali-instance-name

lucasponce:
@nrfox, this one would need a rebase.
I'm happy to help you review it as I can see that this PR can quickly enter into conflicts as soon as we merge other work.

observability/tracing.go
nrfox: If `deployment.instance_name` determines the "workload" name for the app then this is probably what we want to use as the [hostname](https://github.com/kiali/kiali/pull/4532/commits/8837d47a8694fc801d028c979085a240ff9cd579#diff-296a65271be18baa8a1fe01a82d31a85bcb3b223249922564a42d7db3f999ba8R48). `TracingService` is more along the lines of `canonical_service` or the kube service name. I can't really link to the specific open telemetry constants but here's an excerpt:
> 	// Logical name of the service.
	//
	// Type: string
	// Required: Always
	// Stability: stable
	// Examples: 'shoppingcart'

from https://pkg.go.dev/go.opentelemetry.io/otel/semconv/v1.7.0#pkg-constants.

I'll update the hostname resource with the `deployment.instance_name` field as we want that to match the name of the workload since kiali uses that to determine if the trace matches the workload: https://github.com/kiali/kiali/blob/master/business/jaeger.go#L275-L286.

nrfox: Ah I was hoping `deployment.instance_name` would be available on the kiali config but it doesn't appear to be. Ideally the workload name would get pulled from the pod annotations and passed down through the downward api similar to how the pod name is set. That would be less error prone than assuming the workload will always be `kiali`. That being said, it looks like the app name when deploying with the operator will always be `kiali`: https://github.com/kiali/kiali-operator/blob/v1.44.0/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml#L11 so I think it's fine to assume here the workload will just be `kiali`.

jmazzitelli: I'm surprised it isn't available... everything in the Kiali CR will get copied to the Kiali ConfigMap, even the deployment.* settings. See: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/configmap.yaml#L9

Same should be true with the server helm chart, though interestingly, the instance_name is set to the helm chart full name (which I thought should be the instance name, but maybe not?): https://github.com/kiali/helm-charts/blob/master/kiali-server/templates/configmap.yaml#L19

Anyway, if its fine for the traces to use the app name label, then it will always be kiali. I just thought it would be weird that you could have multiple instances installed and no way of knowing WHICH trace came from WHICH instance. But maybe that doesn't matter?

nrfox: > I'm surprised it isn't available... everything in the Kiali CR will get copied to the Kiali ConfigMap, even the deployment.* settings. See: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/configmap.yaml#L9

It looks like it's there in the configmap yaml it's just doesn't exist on the config go struct: https://github.com/kiali/kiali/blob/master/config/config.go#L325-L330. That's an easy fix.

nrfox: Screenshot from 2022-01-04 12-06-49

There we go

lucasponce:
Hi @nrfox, 
A comment about the format, the "service name" used in jaeger typically follows the pattern "<service>.<namespace>":
image
image

Kiali is a workload that doesn't belong to the data plane, it's an infrastructure part of the control plane, but I guess it would be nice to follow the same naming structure.

business/apps.go
```diff
@@ -279,3 +289,33 @@ func fetchNamespaceApps(layer *Layer, namespace string, appName string) (namespa
 
 	return allEntities, nil
 }
+
+type appServiceWithTracing struct {
```
lucasponce, line 293: Ah, ok, so the "AOP" style is implemented providing a different implementation of the interface for tracing/no-tracing.
I think I understand it, I'd like to fetch it locally and review it closely, it looks correct, but as the amount of changes is high I'd review if there is no side effect.

nrfox:
> Hi @nrfox, A comment about the format, the "service name" used in jaeger typically follows the pattern ".": image image
> 
> Kiali is a workload that doesn't belong to the data plane, it's an infrastructure part of the control plane, but I guess it would be nice to follow the same naming structure.

@lucasponce Yes in fact Kiali treats traces from services in the istio system namespace as a special case: https://github.com/kiali/kiali/blob/master/jaeger/client.go#L261-L263. I'm not exactly sure why it does this but I'm happy to add a special case for the kiali app itself or else remove the special case for the istio system namespace altogether if it's no longer needed.

business/apps.go
```diff
@@ -279,3 +289,33 @@ func fetchNamespaceApps(layer *Layer, namespace string, appName string) (namespa
...
+	AppService
+}
+
+func (in *appServiceWithTracing) GetAppList(ctx context.Context, namespace string, linkIstioResources bool) (models.AppList, error) {
+	var span trace.Span
+	ctx, span = otel.Tracer(observability.TracerName()).Start(ctx, "GetAppList",
```
lucasponce, line 299: Suggestion: as we have an "observability" package, I wonder if it looks good to create a helper to propagate the trace/span attributes from the "observability" package instead to propagate the import of the "go.opentelemetry.io" packages almost on every business package.

Something similar as the "kiali/prometheus/internalmetrics" implemented just to pass the "context" and attributes and make the "go.opentelemetry.io" part of the implementation detail of the "observability" package.

lucasponce: Another idea is that if you move the tracing logic into an external "observability" package, perhaps you don't need to deal with two service implementations. The tracing method can always be invoked (I'm just thinking of the same design as the internal metrics). The tracing logic is managed internally. "observability" logic is always invoked, but only spans are added when the tracing service is enabled.

Let me know if that idea is clear, perhaps I could prototype some example, but basically, it would look similar to the usage of the internal metric we had in the past (now we have that service placed at handler level, but in the past, it was located in the same place you are proposing the tracing instrumentation).

business/apps.go
nrfox: > Ah, ok, so the "AOP" style is implemented, providing a different interface implementation for tracing/no-tracing.

I'm not familiar with "AOP" but yeah that's the gist. Only the interface gets exposed and there's a separate implementation that has all the tracing logic that just wraps the original. In another language, something like a tracing decorator would probably be nice here:

Python example
```
def tracing(func):
    def wrapper():
        start_trace()
        func()
        end_trace()
    return wrapper

class AppService:
    @tracing
    def get_apps():
    ...
    @tracing
    def list_apps():
...
```

I'm definitely open to changing this and welcome feedback. I was worried about the amount of code changes involved with this approach but keeping the tracing boilerplate out of the business logic made the integration cleaner. I even played around a bit with code generation for the various tracing interfaces so that the tracing implementations wouldn't need to be maintained individually.

business/apps.go
nrfox: > Suggestion: as we have an "observability" package, I wonder if it looks good to create a helper to propagate the trace/span attributes from the "observability" package instead to propagate the import of the "go.opentelemetry.io" packages almost on every business package.

Sure I can include all those imports and re-export them as part of the observability package.

> Another idea is that if you move the tracing logic into an external "observability" package, perhaps you don't need to deal with two service implementations. The tracing method can always be invoked (I'm just thinking of the same design as the internal metrics). The tracing logic is managed internally. "observability" logic is always invoked, but only spans are added when the tracing service is enabled.

I'll play around with that a bit. This is more of the approach I originally went with having checks for `config.Observability.Tracing.Enabled` before starting the span.

lucasponce: > I'll play around with that a bit. This is more of the approach I originally went with having checks for config.Observability.Tracing.Enabled before starting the span.

From the instrumented code, the call to the observability package could be always present and inside the implementation, it can be checked if the service is enabled/disabled. It's a tradeoff but the cost would be just an empty call instead to bring the check on any business package.

But that direction looks +1.

nrfox: @lucasponce I implemented a small change exposing the span start as part of the observability package: https://github.com/kiali/kiali/pull/4532/commits/a4f3c2f9c9cf38bf347d51ba87308f3d21904ba8. The API is probably a bit rough but I'm not convinced this adds much value over just putting checks in each of the business methods. It basically just abstracts away the `config.Get().Observability.Tracing.Enabled` check.

The problem comes in needing to call `defer span.End()` within the business function itself. Which means you either need to check `span` for nil if you return `span` or go with some kind of cleanup function. Either way it's a bit awkward to use. Open to suggestions here on if this can be made better.

Otherwise I'm fine with either adding the span management directly to the business methods or else keeping the tracing logic in a separate implementation.

business/apps.go
```diff
@@ -57,6 +57,14 @@ func buildFinalLabels(m map[string][]string) map[string]string {
 
 // GetAppList is the API handler to fetch the list of applications in a given namespace
 func (in *appService) GetAppList(ctx context.Context, namespace string, linkIstioResources bool) (models.AppList, error) {
+	var end observability.EndFunc
+	ctx, end = observability.StartSpan(ctx, "GetAppList",
```
lucasponce, line 61: +1 on this approach.
A couple of suggestions, I don't think that the parameters would need to be strong-typed, perhaps they could be moved to strings all the time (are just used to fill the tags and log).
Also, I think that all business methods should call the "StartSpan", no need to have two implementations in the business/layer, the observability package can handle the "enabled/disabled" I think. The cost would be to have an empty call, which wouldn't cause any performance, but it would reduce the boilerplate of the business/layer side.

nrfox:
@lucasponce rebased with master if you could PTAL when you get a chance

observability/tracing.go
```diff
@@ -0,0 +1,119 @@
...
+// InitTracer initalizes a TracerProvider that exports to jaeger.
+// This will panic if there's an error in setup.
+func InitTracer(jaegerURL string) *sdktrace.TracerProvider {
+	exporter, err := jaeger.New(jaeger.WithCollectorEndpoint(jaeger.WithEndpoint(jaegerURL)))
+	if err != nil {
+		panic(err)
+	}
+
+	tp := sdktrace.NewTracerProvider(
+		sdktrace.WithSampler(sdktrace.ParentBased(sdktrace.TraceIDRatioBased(0.5))), // Sample half of traces.
```
lucasponce, line 45: This parameter "0.5" is pretty sensible to performance impact. I'd suggest making it available in the configuration.

business/istio_config.go
```diff
@@ -96,7 +97,13 @@ var newSecurityConfigTypes = []string{
 
 // GetIstioConfigList returns a list of Istio routing objects, Mixer Rules, (etc.)
 // per a given Namespace.
-func (in *IstioConfigService) GetIstioConfigList(criteria IstioConfigCriteria) (models.IstioConfigList, error) {
+func (in *IstioConfigService) GetIstioConfigList(ctx context.Context, criteria IstioConfigCriteria) (models.IstioConfigList, error) {
+	var end observability.EndFunc
+	ctx, end = observability.StartSpan(ctx, "GetIstioConfigList",
+		observability.Attribute("package", "business"),
```
lucasponce, line 103: I'd say that in the methods with "criteria" it would be good to have perhaps a helper method to pass the criteria.
Not blocker, I guess that can be consolidated in a future iteration, but it could be nice to check which criteria can show more latencies.

business/services.go
```diff
@@ -34,7 +36,13 @@ type ServiceCriteria struct {
 }
 
 // GetServiceList returns a list of all services for a given criteria
-func (in *SvcService) GetServiceList(criteria ServiceCriteria) (*models.ServiceList, error) {
+func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriteria) (*models.ServiceList, error) {
+	var end observability.EndFunc
+	ctx, end = observability.StartSpan(ctx, "GetServiceList",
+		observability.Attribute("package", "business"),
```
lucasponce, line 42: Same here, it could be good to pass the "criteria" in some way to the span to identify which criteria shows a bad latency.
Not a blocker, that can be done in a future iteration.

business/workloads.go
```diff
@@ -80,7 +82,13 @@ func isWorkloadIncluded(workload string) bool {
 }
 
 // GetWorkloadList is the API handler to fetch the list of workloads in a given namespace.
-func (in *WorkloadService) GetWorkloadList(criteria WorkloadCriteria) (models.WorkloadList, error) {
+func (in *WorkloadService) GetWorkloadList(ctx context.Context, criteria WorkloadCriteria) (models.WorkloadList, error) {
```
lucasponce, line 85: This is another criteria to pass into the span.
Not a blocker, I leave a comment here to not forget but I'd recommend adding it in a future iteration.

lucasponce:
Thanks, @nrfox. I've done a pass to the code (but without testing) and changes look ok from my side.
I'm going to prepare a local environment to run some smoke tests to validate that there is no obvious regression but I hope to approve it today if no obvious blocker.

I put a couple of non-blocking comments that can be in another pass, I know how big is this PR and I think it would be good to consolidate to avoid more conflicts.

lucasponce (approved):
LGTM
Smoke tests ran in a local environment.

lucasponce:
Not commented in the PR but I think this would require propagating the new config into the operator config as well.

business/istio_config.go
nrfox: Sure `Attribute` will work with anything that implements `Stringer` so I can add a `String` method to `IstioConfigCriteria`.

nrfox:
> Not commented in the PR but I think this would require propagating the new config into the operator config as well.

Yes the metrics field in the config moved from `server.metrics` to `server.observability.metrics`. I don't think this requires an operator change for functionality except that by convention we set all the config values in the CR defaults:
https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L309-L321

It's also worth noting though that this is a backwards incompatible change to go from `server.metrics` --> `server.observability.metrics`. We should mention it in the changelog/release notes but I don't believe the kaili configuration has guarantees around backward compatibility and I don't expect the majority of users to have metrics permanently enabled for kiali as it is mostly there for observing kiali itself. @jmazzitelli does that sound right?

jmazzitelli:
> It's also worth noting though that this is a backwards incompatible change to go from `server.metrics` --> `server.observability.metrics`. We should mention it in the changelog/release notes but I don't believe the kaili configuration has guarantees around backward compatibility and I don't expect the majority of users to have metrics permanently enabled for kiali as it is mostly there for observing kiali itself. @jmazzitelli does that sound right?

Yes. We talked about this earlier and it was my opinion not to worry about it. See my comment: https://github.com/kiali/kiali/pull/4532#discussion_r759612672

nrfox:
> > It's also worth noting though that this is a backwards incompatible change to go from `server.metrics` --> `server.observability.metrics`. We should mention it in the changelog/release notes but I don't believe the kaili configuration has guarantees around backward compatibility and I don't expect the majority of users to have metrics permanently enabled for kiali as it is mostly there for observing kiali itself. @jmazzitelli does that sound right?
> 
> Yes. We talked about this earlier and it was my opinion not to worry about it. See my comment: [#4532 (comment)](https://github.com/kiali/kiali/pull/4532#discussion_r759612672)

That's right. It's been a minute :)

I'll add an operator PR to update the defaults for the kiali CR yaml but they'll be the same as what is [currently being defaulted on the server](https://github.com/kiali/kiali/pull/4532/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R687-R696) so it shouldn't change anything:  @pbajjuri20

jmazzitelli:
> I'll add an operator PR to update the defaults for the kiali CR yaml

I'll update my [operator PR](https://github.com/kiali/kiali-operator/pull/470) to include this new section in the new CRD schema.

pbajjuri20 (approved):

lucasponce:
Sorry, @nrfox, I've merged the removal of the iter8 and that change touched this PR.
I think with a minimal rebase this PR would be good.

nrfox:
> Sorry, @nrfox, I've merged the removal of the iter8 and that change touched this PR. I think with a minimal rebase this PR would be good.

No worries! Rebased with master.

lucasponce (approved):

lucasponce:
QE has tested this one, I'm going to consolidate it in master to prevent potential regressions.
Thanks for the work @nrfox.

If there is anything pending, let's add a follow-up PR on it.

PR #4535, opened 2021-12-02
Fetching Exported Istio Configs from Registry for AllNamespaces

hhovsepy:
RFE subtask https://github.com/kiali/kiali/issues/4528

Before: Kiali was listing exported configs per namespace.

Now: List all exported configs for all namespaces from registry.

### DIFF
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -58,19 +58,17 @@ func (in *IstioValidationsService) GetValidations(namespace, service string) (mo
 	var services models.ServiceList
 	var namespaces models.Namespaces
 	var workloadsPerNamespace map[string]models.WorkloadList
-	var gatewaysPerNamespace [][]networking_v1alpha3.Gateway
 	var mtlsDetails kubernetes.MTLSDetails
 	var rbacDetails kubernetes.RBACDetails
 	var registryServices []*kubernetes.RegistryService
 
-	wg.Add(9) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
+	wg.Add(8) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
 
 	// We fetch without target service as some validations will require full-namespace details
 	go in.fetchIstioConfigList(&istioConfigList, namespace, errChan, &wg)
 	go in.fetchExportedResources(&exportedResources, namespace, errChan, &wg)
 	go in.fetchNamespaces(&namespaces, errChan, &wg)
 	go in.fetchAllWorkloads(&workloadsPerNamespace, errChan, &wg)
-	go in.fetchGatewaysPerNamespace(&gatewaysPerNamespace, errChan, &wg)
 	go in.fetchNonLocalmTLSConfigs(&mtlsDetails, namespace, errChan, &wg)
 	go in.fetchAuthorizationDetails(&rbacDetails, namespace, errChan, &wg)
 	go in.fetchServices(&services, namespace, errChan, &wg)
@@ -84,7 +82,7 @@ func (in *IstioValidationsService) GetValidations(namespace, service string) (mo
 		}
 	}
 
-	objectCheckers := in.getAllObjectCheckers(namespace, istioConfigList, exportedResources, services, workloadsPerNamespace, workloadsPerNamespace[namespace], gatewaysPerNamespace, mtlsDetails, rbacDetails, namespaces, registryServices)
+	objectCheckers := in.getAllObjectCheckers(namespace, istioConfigList, exportedResources, services, workloadsPerNamespace, workloadsPerNamespace[namespace], mtlsDetails, rbacDetails, namespaces, registryServices)
 
 	// Get group validations for same kind istio objects
 	validations := runObjectCheckers(objectCheckers)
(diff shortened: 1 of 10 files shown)


lucasponce:
I think we need to refine the cross references, specially on workload cases:
image

The default/generic EnvoyFilters don't add information here for the user, but adding noise.

I'm still reviewing the work, but that comes to my attention, I'll add more comments as I progress in the review.

lucasponce:
Also, the business/apps.go has a bug (not from this PR, it was introduced before):
image

The Istio reference flips Namespace/Name making to fail in the link of the Istio config detail.

business/workloads.go
```diff
@@ -108,6 +108,7 @@ func (in *WorkloadService) GetWorkloadList(criteria WorkloadCriteria) (models.Wo
 	}()
 
 	istioConfigCriteria := IstioConfigCriteria{
+		AllNamespaces:                 true,
```
lucasponce, line 111: This change is showing a problem in the way the EnvoyFilters are filtered.

Probably this method is not 100% correct:
image

It was introduced before but now with this PR we can see the problem.

Perhaps for these object we may want to a) make the filter explictly (don't add to the result generic EF objects) or b) just ignore the EF for now, as those objects are typically low level, and rely the user to query them from the Istio Config only.

hhovsepy:
@lucasponce thank you for the review. I will revert changes with AllNamespaces in apps and workloads pages. They could be done as a part of another epic, outside of the scope of this PR.

business/istio_validations.go
```diff
@@ -58,19 +58,17 @@ func (in *IstioValidationsService) GetValidations(namespace, service string) (mo
...
-	wg.Add(9) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
+	wg.Add(8) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
 
 	// We fetch without target service as some validations will require full-namespace details
 	go in.fetchIstioConfigList(&istioConfigList, namespace, errChan, &wg)
 	go in.fetchExportedResources(&exportedResources, namespace, errChan, &wg)
 	go in.fetchNamespaces(&namespaces, errChan, &wg)
 	go in.fetchAllWorkloads(&workloadsPerNamespace, errChan, &wg)
-	go in.fetchGatewaysPerNamespace(&gatewaysPerNamespace, errChan, &wg)
 	go in.fetchNonLocalmTLSConfigs(&mtlsDetails, namespace, errChan, &wg)
```
lucasponce, line 72: A minor comment here to confirm that the scope of this work would be just the IstioConfigList query, isn't it ?
I see more improvements here but I see it would be better to do it in an iterative way.

lucasponce (approved):
Code looks ok and a smoke tests looks good.

image

lucasponce:
@mattmahoneyrh this PR would require some regression test, but it's part of a set of work that @hhovsepy is leading.
Perhaps when consolidated, we could run a set of regression tests at the end, to easy QE testing on it.

But any suggestion is welcome here.

business/istio_validations.go
hhovsepy: You are right.
There are 2 things combined in this PR, one is the IstioConfigList query, and the second is optimizing fetching Gateways per namespace and instead use the query result (which was the initial idea, but without exported query it would not be done), so I sent all together to save time in reviews.

prachiyadav:
@hhovsepy I am getting below E2E test cases failed... Is this expected?

***************************************************************************************
tests/test_api_methods.py .....s...........s............................[36m [ 20%]
[0m...................[36m                                                      [ 28%][0m
tests/test_kiali_application_endpoint.py ....[36m                            [ 30%][0m
tests/test_kiali_bookinfo_namespace_endpoint.py .F[36m                       [ 31%][0m
tests/test_kiali_external_host_node.py F[36m                                 [ 31%][0m
tests/test_kiali_graph_badges.py ........................[36m                [ 42%][0m
tests/test_kiali_graph.py .................................[36m              [ 56%][0m
tests/test_kiali_health.py .............................................[36m [ 76%]
[0m...[36m                                                                      [ 77%][0m
tests/test_kiali_istio_endpoint.py ..FF............[36m                      [ 84%][0m
tests/test_kiali_logs.py ....[36m                                            [ 86%][0m
tests/test_kiali_metrics.py ...[36m                                          [ 87%][0m
tests/test_kiali_node_detail_graph.py F.FFF.[36m                             [ 90%][0m
tests/test_kiali_service_endpoint.py F........[36m                           [ 94%][0m
tests/test_kiali_traces.py ......[36m                                        [ 96%][0m
tests/test_kiali_urls.py ..[36m                                              [ 97%][0m
tests/test_kiali_workload_endpoint.py .....[36m                              [100%][0m

=================================== FAILURES ===================================
*************************************************************************************************************************

hhovsepy:
> @hhovsepy I am getting below E2E test cases failed... Is this expected?
It is not expected to be failed, it seems env is broken or bookinfo is missing.

prachiyadav:
ok, let me re check

pbajjuri20:
Rest Automation Test cases Run Successfully without any Errors

hhovsepy:
@lucasponce I have rebased, resolved the conflicts and pushed again.

pbajjuri20 (approved):

pbajjuri20 (dismissed):

lucasponce (approved):
LGTM
Smote test performed locally

lucasponce (approved):
LGTM

lucasponce:
QE has already tested the PR:
image

So, no need to re-assign this to QE as the latest commit was just a go format fix.

pbajjuri20:
Yes, I have already tested this PR. Thanks for checking @lucasponce and @hhovsepy

PR #4536, opened 2021-12-02
Use AllNamespaces flag in IstioConfig for TLS validation

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4224

Improve the getAllDestinationRules to use the AllNamespaces flag that queries directly the Istio Registry to collect the DestinationRules needed for this cross-namespace validation.

It's still in a draft mode, tests need to be updated and registry cache can be indexed per type to reduce the O(N) -> O(1) using memory.

Also related with https://github.com/kiali/kiali/discussions/4509 discussion

PR #4549, opened 2021-12-07
Fix VS badging

### LINKED ISSUE #4541: Graph: Correctly badge service nodes with the VS/Route icon
The code that is badging service nodes in the graph is not badging some nodes being affected by VirtualServices.

Historically, the code that is adding VS badges works by matching the Service Name and the VS name to decide if a badge is added. If this still holds true, probably that code should be enhanced to do matching using the routes in the VS definition similarly to the way other appenders in the graph work, and also similarly to the way relationships are resolved in detail pages.

israel-hdez:
Badging of VSs was assuming that Istio VirtualServices resources and graph service nodes belong to the same namespace. This is not necessarily true. This fixes VS badging to take into account that a VS can route to a Service in another namespace.

Fixes #4541

### DIFF
--- a/graph/telemetry/istio/appender/istio_details_test.go
+++ b/graph/telemetry/istio/appender/istio_details_test.go
@@ -167,124 +167,125 @@ func TestCBSubset(t *testing.T) {
 	assert.Equal(nil, trafficMap[wlNodeId].Metadata[graph.HasVS])
 }
 
-func TestVS(t *testing.T) {
-	assert := assert.New(t)
-	config.Set(config.NewConfig())
-
-	k8s := kubetest.NewK8SClientMock()
-	vService := &networking_v1alpha3.VirtualService{}
-	vService.Name = "vService-1"
-	vService.Namespace = "testNamespace"
-	vService.Spec.Hosts = []string{"ratings"}
-	vService.Spec.Http = []*api_networking_v1alpha3.HTTPRoute{
-		{
-			Route: []*api_networking_v1alpha3.HTTPRouteDestination{
-				{
-					Destination: &api_networking_v1alpha3.Destination{
-						Host: "foo",
-					},
-				},
-			},
-		},
-	}
-	k8s.MockIstio(vService)
-
-	k8s.On("GetProject", mock.AnythingOfType("string")).Return(&osproject_v1.Project{}, nil)
-	k8s.On("GetProjects", mock.AnythingOfType("string")).Return([]osproject_v1.Project{}, nil)
-	k8s.On("GetEndpoints", mock.AnythingOfType("string"), mock.AnythingOfType("string")).Return(&core_v1.Endpoints{}, nil)
-	k8s.On("GetServices", mock.AnythingOfType("string"), mock.Anything).Return([]core_v1.Service{{}}, nil)
-
-	businessLayer := business.NewWithBackends(k8s, nil, nil)
-	trafficMap, appNodeId, appNodeV1Id, appNodeV2Id, svcNodeId, wlNodeId, fooSvcNodeId := setupTrafficMap()
-
-	assert.Equal(6, len(trafficMap))
-	assert.Equal(nil, trafficMap[appNodeId].Metadata[graph.HasCB])
-	assert.Equal(nil, trafficMap[appNodeV1Id].Metadata[graph.HasCB])
-	assert.Equal(nil, trafficMap[appNodeV2Id].Metadata[graph.HasCB])
-	assert.Equal(nil, trafficMap[svcNodeId].Metadata[graph.HasCB])
-	assert.Equal(nil, trafficMap[wlNodeId].Metadata[graph.HasCB])
-	assert.Equal(nil, trafficMap[appNodeId].Metadata[graph.HasVS])
-	assert.Equal(nil, trafficMap[appNodeV1Id].Metadata[graph.HasVS])
-	assert.Equal(nil, trafficMap[appNodeV2Id].Metadata[graph.HasVS])
-	assert.Equal(nil, trafficMap[svcNodeId].Metadata[graph.HasVS])
-	assert.Equal(nil, trafficMap[wlNodeId].Metadata[graph.HasVS])
-	assert.Equal(nil, trafficMap[fooSvcNodeId].Metadata[graph.HasVS])
-
-	globalInfo := graph.NewAppenderGlobalInfo()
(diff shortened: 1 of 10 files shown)


graph/telemetry/istio/appender/istio_details_test.go
```diff
@@ -167,124 +167,125 @@ func TestCBSubset(t *testing.T) {
...
-	globalInfo.Business = businessLayer
-	namespaceInfo := graph.NewAppenderNamespaceInfo("testNamespace")
-
-	a := IstioAppender{}
-	a.AppendGraph(trafficMap, globalInfo, namespaceInfo)
-
-	assert.Equal(true, trafficMap[fooSvcNodeId].Metadata[graph.HasTrafficShifting])
-}
+// TODO: Re-add following two tests. A way for testing code that *requires* the cache to be enabled is needed.
+//func TestVS(t *testing.T) {
```
nrfox, line 171: What was the issue with these tests?

nrfox (commented):
Fix looks good. Seeing the egress gateway getting badged with the electronic shop demo deployed. Just one question about tests.

graph/telemetry/istio/appender/istio_details_test.go
israel-hdez: Usually, when the tested code makes call to the Kubernetes API, we turn off the cache to be able to mock the Kubernetes client. Like here: https://github.com/kiali/kiali/blob/223bd09efb46272b13e56358e14b6dbf60cdcff3/handlers/apps_test.go#L147-L149

However, the changes that I made here: https://github.com/kiali/kiali/pull/4549/files#diff-3fa7b3ce1da5b86b1f92593027a74000ba5af834ba946e97d92a405f9efd98baR48-R56

I inspected the inner code and I saw that when you use `AllNamespaces: true` the internal code of the `GetIstioConfigList` function _requires_ that the Kiali cache is enabled. Otherwise, it returns nothing. This is the relevant code: https://github.com/kiali/kiali/blob/223bd09efb46272b13e56358e14b6dbf60cdcff3/business/istio_config.go#L121-L133

Currently, we don't have a way to mock the cache. We would need to come with something to be able to mock everything inside https://github.com/kiali/kiali/tree/master/kubernetes/cache. Perhaps we can do it incrementally. I may try to mock only what I need here, but I have the idea that it would result in a larger PR. So, I'm not sure if I should do it under the scope of this issue. Anyway, nowadays I'm not even sure that unit testing is a priority (yes... a little scary, but it is what it is).

WDYT? We could start a thread with the team to see where are we really going regarding unit tests.

lucasponce: @israel-hdez there is a WIP for adding a better mock for cache testing:

https://github.com/kiali/kiali/pull/4536/files#diff-124ae619939347dcb6d96d88a3494a5b5da0091d4bdfe4975ecfd8c99aa2553a

Also @hhovsepy is working on the same area as well, so, there are some parallel efforts around same code.

Just pure sync here, as we need test, my comment is more to organize the flow, perhaps we can add these tests later, or if someone would like to mimic something similar as that WIP, any option is welcome.

israel-hdez: From what @lucasponce is saying, I think it is probably better to re-add the tests on a follow-up PR once we learn how to do it.

@lucasponce From your draft, It looks somewhat straightforward. I was thinking it was going to be more complicated. But I'm wrong apparently :-)

nrfox: > From what @lucasponce is saying, I think it is probably better to re-add the tests on a follow-up PR once we learn how to do it.

It's unfortunate that there isn't a way to easily mock the cache currently however, you could also argue that adding test coverage _is_ fundamentally part of fixing this issue but that gets back to whether unit testing is a priority or not like you mentioned. I'll always advocate for including tests with your changes as part of the definition of done but totally agree that's a broader discussion than this particular change though so I won't hold this fix up.

nrfox (approved):

graph/telemetry/istio/appender/istio_details_test.go
lucasponce: Feel free to create a Fake cache for your testing needs as defined in the WIP.
This week was slow to me and I couldn't progress too much but I think from that draft the testing strategy could be to fake the global [1] with your Fake cache for your needs.

That test is also an hybrid using kubernetes calls and registry calls, and there is room for improvement.

I expect to continue working on that next week, I can help on this in your PR reviewing or commenting, just let me know.

[1] https://github.com/kiali/kiali/pull/4536

pbajjuri20:
Before PR:

before_4549

After PR:

after_4549

pbajjuri20 (dismissed):

pbajjuri20:
Polarion - 2177

kubernetes/filters.go
```diff
@@ -32,22 +32,22 @@ func FilterAuthorizationPoliciesBySelector(workloadSelector string, authorizatio
 	return filtered
 }
 
-func FilterByHost(host, serviceName, namespace string) bool {
+func FilterByHost(host, hostNamespace, serviceName, svcNamespace string) bool {
```
lucasponce, line 35: Perhaps this deserves a comment that the "hostNamespace" may be included into the "host" string ?

I see that change is being propagated into the validations logic and in theory it should match but the host defined in a DR shouldn't necessarily match the namespace where the DR is defined.

If tests are passing we are good, just please let's leave a comment in cases that there is any regression.

@hhovsepy I think that if make test is green, we are good here, but as you are also updating the validations, just in case that you may have any other aspect into consideration.

israel-hdez: > Perhaps this deserves a comment that the "hostNamespace" may be included into the "host" string ?
> 
> I see that change is being propagated into the validations logic and in theory it should match but the host defined in a DR shouldn't necessarily match the namespace where the DR is defined.

Yes. Indeed, it may deserve more explanation than that.
The idea is that if `host` is "simple", like `reviews` rather than the full FQDN, you should pass what is the DR namespace in `hostNamespace` to correctly differentiate if the service and the host are really referencing the same thing.

However, if the `host` is already including the namespace on it, the `hostNamespace` parameter becomes irrelevant, because the very first `if` should never match.

I'll add a doc comment to this function.

israel-hdez: I added the comment.

nrfox (approved):

PR #4565, opened 2021-12-07
Do complete purge of all istio resources when uninstalling

FilipB:
The optional --purge flag will remove all Istio resources,
including cluster-scoped resources that may be shared with other Istio control planes.

This helps to reuse a cluster as there should be no istio leftovers after uninstalling.

### DIFF
--- a/hack/istio/install-istio-via-istioctl.sh
+++ b/hack/istio/install-istio-via-istioctl.sh
@@ -19,6 +19,7 @@ CLIENT_EXE_NAME="oc"
 CLUSTER_NAME="cluster-default"
 CONFIG_PROFILE="" # see "istioctl profile list" for valid values. See: https://istio.io/docs/setup/additional-setup/config-profiles/
 DELETE_ISTIO="false"
+PURGE_UNINSTALL="true"
 ISTIOCTL=
 ISTIO_DIR=
 ISTIO_EGRESSGATEWAY_ENABLED="true"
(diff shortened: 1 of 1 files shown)


hack/istio/install-istio-via-istioctl.sh
```diff
@@ -328,7 +328,10 @@ if [ "${DELETE_ISTIO}" == "true" ]; then
   done
 
   echo Deleting Core Istio
-  ${ISTIOCTL} manifest generate --set profile=${CONFIG_PROFILE} ${MANIFEST_CONFIG_SETTINGS_TO_APPLY} | ${CLIENT_EXE} delete -f -
+  # Although the 'uninstall' command has been available since Istio 1.7 and is listed in the official istio doc,
+  # it's still available in 'experimental' (x) scope, so it might change in the future.
+  # The optional --purge flag will remove all Istio resources, including cluster-scoped resources that may be shared with other Istio control planes.
```
jmazzitelli, line 333: > including cluster-scoped resources that may be shared with other Istio control planes.

So this means this script will destroy other istio installs. If I have multiple istio installs and I want to remove one of them, I can't do it. I think we want to support that use-case. So I suggest adding another option here... "--purge-uninstall" - its default should be "true" but allow it to be "false" in case a user has multiple istio installs and only wants to uninstall one of them.

jmazzitelli (changes requested):

hack/istio/install-istio-via-istioctl.sh
FilipB: Pushed requested changes.

jmazzitelli (approved):
i tested this and it works.

Issue #4570, opened 2021-12-09
"Could not fetch services list" Error in Service view when selecting some namespaces


maxbischoff:
After upgrading our Kiali to v1.44 the Service view does not show any Services anymore, when some namespaces are selected. For other namespaces (e.g. only `istio-system`) the Service list is rendered correctly.

![Screenshot 2021-12-08 at 09 58 01](https://user-images.githubusercontent.com/12122170/145367768-8e4d8db0-3d04-4e3c-929c-665ed8451f14.png)

![Screenshot 2021-12-08 at 09 58 10](https://user-images.githubusercontent.com/12122170/145367782-8def4274-ebbc-4aef-8b7e-c96a6591fe36.png)

The browser console shows following log:

```
main.3c0ceb9f.chunk.js:1 Could not fetch services list: 
value @ main.3c0ceb9f.chunk.js:1
(anonymous) @ main.3c0ceb9f.chunk.js:1
Promise.catch (async)
value @ main.3c0ceb9f.chunk.js:1
value @ main.3c0ceb9f.chunk.js:1
value @ main.3c0ceb9f.chunk.js:1
os @ 2.69e58466.chunk.js:2
du @ 2.69e58466.chunk.js:2
t.unstable_runWithPriority @ 2.69e58466.chunk.js:2
Wi @ 2.69e58466.chunk.js:2
hu @ 2.69e58466.chunk.js:2
Zs @ 2.69e58466.chunk.js:2
(anonymous) @ 2.69e58466.chunk.js:2
t.unstable_runWithPriority @ 2.69e58466.chunk.js:2
Wi @ 2.69e58466.chunk.js:2
Yi @ 2.69e58466.chunk.js:2
Hi @ 2.69e58466.chunk.js:2
eu @ 2.69e58466.chunk.js:2
notify @ 2.69e58466.chunk.js:2
t.notifyNestedSubs @ 2.69e58466.chunk.js:2
t.handleChangeWrapper @ 2.69e58466.chunk.js:2
m @ 2.69e58466.chunk.js:2
(anonymous) @ 2.69e58466.chunk.js:2
setActiveNamespaces @ main.3c0ceb9f.chunk.js:1
n.onToggle @ main.3c0ceb9f.chunk.js:1
(anonymous) @ 2.69e58466.chunk.js:2
(anonymous) @ 2.69e58466.chunk.js:2
main.3c0ceb9f.chunk.js:1 Could not fetch services list: 
[...]

PR #4572, opened 2021-12-09
Fix Service list bug on multi-cluster scenarios

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4570

This PRs address two potential problems:
- During Istio Registry call, the assignment is protected but not the call to registry, so due https://github.com/istio/istio/issues/36448 now the refresh operation should be protected.
- In multi-cluster scenarios, the Istio Registry replicates the Kubernetes services, so, Istio registry services should be filtered to only used based on the istiod cluster he is connected to (probably this needs to be revisited when a Kiali instance may connect multiple control planes).

How to test:
- This issue requires a multi-cluster setup in https://github.com/kiali/kiali/discussions/3699 there are some good helpers to define one in minikube.
- Create a namespaces in both istio-east and istio-west, populates only services in one namespace.
- Before this fix, one of the Kiali UI would fail, as it's not filtering correctly the services, with the fix check that there are no errors when navigating to the Service list page.

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -252,11 +252,67 @@ func (in *SvcService) buildKubernetesServices(svcs []core_v1.Service, pods []cor
 	return services
 }
 
+// The istiod registry doesn't have a explicit flag when a service is deployed in a different control plane.
+// The only way to identify it is to check that the service has an address in the current cluster.
+// To avoid side effects, Kiali will process only services that belongs to the current cluster.
+// This should be revisited on more multi-cluster deployments scenarios.
+//     "hostname": "test-svc.evil.svc.cluster.local",
+//    "clusterVIPs": {
+//      "Addresses": {
+//        "istio-west": [
+//          "0.0.0.0"
+//        ]
+//      }
+//    },
+func (in *SvcService) getClusterId() string {
+	// By default Istio uses "Kubernetes" as clusterId for single control planes scenarios.
+	// This clusterId is propagated into the Istio Registry and we need it to filter services in multi-cluster scenarios.
+	// I.e.:
+	//    "clusterVIPs": {
+	//      "Addresses": {
+	//        "Kubernetes": [
+	//          "10.217.4.189"
+	//        ]
+	//      }
+	//    }
+	clusterId := DefaultClusterID
+	// Protection on tests
+	if in.businessLayer != nil {
+		if cluster, err := in.businessLayer.Mesh.ResolveKialiControlPlaneCluster(nil); err == nil {
+			clusterId = cluster.Name
+		} else {
+			log.Errorf("Cluster Id resolution failed: %s", err)
+		}
+	}
+	return clusterId
+}
+
+func filterIstioServiceByClusterId(clusterId string, item *kubernetes.RegistryService) bool {
+	if clusterId == "Kubernetes" {
+		return true
+	}
+	// External and Federation services are always local to the control plane
+	if item.Attributes.ServiceRegistry != "Kubernetes" {
+		return true
+	}
+	if _, ok := item.ClusterVIPs12.Addresses[clusterId]; ok {
+		return true
+	}
+	if _, ok := item.ClusterVIPs11[clusterId]; ok {
+		return true
+	}
+	return false
+}
+
 func (in *SvcService) buildRegistryServices(rSvcs []*kubernetes.RegistryService, istioConfigList models.IstioConfigList) []models.ServiceOverview {
-	services := make([]models.ServiceOverview, len(rSvcs))
+	services := []models.ServiceOverview{}
 	conf := config.Get()
 
-	for i, item := range rSvcs {
+	clusterId := in.getClusterId()
+	for _, item := range rSvcs {
+		if !filterIstioServiceByClusterId(clusterId, item) {
+			continue
+		}
 		_, appLabel := item.Attributes.LabelSelectors[conf.IstioLabels.AppLabelName]
 		// ServiceEntry/External and Federation will be marked as hasSidecar == true as they will have telemetry
 		hasSidecar := true
@@ -287,7 +343,7 @@ func (in *SvcService) buildRegistryServices(rSvcs []*kubernetes.RegistryService,
 		}
 		svcReferences = FilterUniqueIstioReferences(svcReferences)
 		// External Istio registries may have references to ServiceEntry and/or Federation
-		services[i] = models.ServiceOverview{
+		service := models.ServiceOverview{
 			Name:              item.Attributes.Name,
 			Namespace:         item.Attributes.Namespace,
 			IstioSidecar:      hasSidecar,
@@ -298,6 +354,7 @@ func (in *SvcService) buildRegistryServices(rSvcs []*kubernetes.RegistryService,
 			IstioReferences:   svcReferences,
 			ServiceRegistry:   item.Attributes.ServiceRegistry,
 		}
+		services = append(services, service)
 	}
 	return services
 }
--- a/business/apps_test.go
+++ b/business/apps_test.go
@@ -21,6 +21,7 @@ import (
 func setupAppService(k8s *kubetest.K8SClientMock) AppService {
 	prom := new(prometheustest.PromClientMock)
 	layer := NewWithBackends(k8s, prom, nil)
+	setupGlobalMeshConfig()
 	return AppService{k8s: k8s, prom: prom, businessLayer: layer}
 }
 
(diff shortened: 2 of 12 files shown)


lucasponce:
@mattmahoneyrh this PR would require testing when devs are ok with the fix.

lucasponce:
cc @maxbischoff in case you'd like to test/review it as well.

Thanks for reporting the problem.

maxbischoff:
> cc @maxbischoff in case you'd like to test/review it as well.
> 
> Thanks for reporting the problem.

I just deployed Kiali with your fix into our test cluster and the error appears no more. Thanks for fixing this!

hhovsepy (approved):
Code LGTM, regression tested on single OCP cluster.

israel-hdez (commented):
I only have one small comment. I'm unsure if it is viable to use that business method that I mention. If possible, it would be my preference. Else, this is quite fine.

business/services.go
```diff
@@ -252,11 +252,60 @@ func (in *SvcService) buildKubernetesServices(svcs []core_v1.Service, pods []cor
...
+	// This clusterId is propagated into the Istio Registry and we need it to filter services in multi-cluster scenarios.
+	// I.e.:
+	//    "clusterVIPs": {
+	//      "Addresses": {
+	//        "Kubernetes": [
+	//          "10.217.4.189"
+	//        ]
+	//      }
+	//    }
+	return "Kubernetes"
```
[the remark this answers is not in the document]
lucasponce: Thanks @israel-hdez, let me try it

lucasponce:
@hhovsepy @israel-hdez I've followed Edgar's suggestion and the ClusterID logic is invoking the Mesh service to resolve the id.

I think it makes completely sense, perhaps it's adding more mock changes into the testing but that's something needs to be updated.

At soon as we add more inter-depependencies between backend services, perhaps we could update the mocks that we use, as they are starting to look messy sometimes.

Let me know how this PR looks like.

hhovsepy (approved):

business/services.go
```diff
@@ -252,11 +252,67 @@ func (in *SvcService) buildKubernetesServices(svcs []core_v1.Service, pods []cor
...
+	//      "Addresses": {
+	//        "Kubernetes": [
+	//          "10.217.4.189"
+	//        ]
+	//      }
+	//    }
+	clusterId := DefaultClusterID
+	// Protection on tests
+	if in.businessLayer != nil {
+		if cluster, err := in.businessLayer.Mesh.ResolveKialiControlPlaneCluster(nil); err == nil {
```
israel-hdez, line 281: I wonder if this works for single cluster.
I remember this can return (nil, nil).

Let me try...

israel-hdez: OK. Looks like it works as long as Istio and Kiali are both installed on the same cluster. If it happens that Istio is not found, the error I see is `Could not fetch services list: unable to find any healthy Pilot instance`. So the services list is protected from a `panic`.

So, this is OK.

israel-hdez (approved):
Looks good.

prachiyadav (approved):

zijianz:
Do we have any plan to backport this to 1.44 since istio 1.12 can not use versions beyond it? @lucasponce

zijianz:
@lucasponce Do we have any plan to backport this to v1.44 since istio 1.12 can not use versions higher than it?

lucasponce:
Hi @zijianz, typically we don't backport issues for old versions.
But as in this case, it's an explicit request if the @kiali/maintainers are ok, we can consider it.

zijianz:
@lucasponce  can you guys consider it?
- The istio issue is still in open state;
- This issue was found in istio 1.12; The fix was in Kiali v1.45. Istio 1.12 can only use versions less or equal than v1.44

The bug is still open if people are using istio 1.12

jshaughn:
I am OK with a backport and a patch release for 1.44.  Istio 1.12 is still supported by Istio and this will support the Kiali community using that version.

leandroberetta:
I'm OK too, it will be an opportunity to try the release type: patch in the new release pipeline :D.

zijianz:
Hi @lucasponce , do we have an estimated time like how much time it will take approximately for this backporting?

zijianz:
If we have multiple Kiali instances running on different clusters with a shared istiod, with many users using Kiali at the same time, will this still be an issue? 

My understanding is that as long as we send requests to `/debug/registryz` in an intensive manner, there is a possibility that this bug will be triggered?

lucasponce:
> do we have an estimated time like how much time it will take approximately for this backporting?

Hi @zijianz, it can be planned during the following weeks. But the team is also bandwidth limited, so if some community member would like to prepare a Pull Request that would help as well.

> If we have multiple Kiali instances running on different clusters with a shared istiod, with many users using Kiali at the same time, will this still be an issue?

> My understanding is that as long as we send requests to /debug/registryz in an intensive manner, there is a possibility that this bug will be triggered?

Note that the Istiod information fetched by Kiali is also cached locally to protect and reduce the access to the istiod, but it could be a scalability issue, of course.

Do you have some figures about cluster sizing, users, requests? The team is focused on scalability issues at this moment that info could help a lot.

Thanks

zijianz:
@lucasponce We haven't fully released Kiali to our engineers yet so currently there is no figures. 

Have anyone experienced this issue before? 
1. Having more than 1 replica of Kiali running is possible to crash istiod although the chance is low. But crashing istiod is something we want to try our best to avoid. 
2. Kiali overview page sends couple of request for each namespace,  while having hundreds of namespaces, loading the overview page takes quite long.

Combining 1 and 2, I see that even with just one person running Kiali still sometimes getting performance issue. Has anyone seen this before?

lucasponce:
@zijianz sorry for the delay, due to the release of Kiali 1.52, this backport was pending on my todo list.

lucasponce:
> Having more than 1 replica of Kiali running is possible to crash istiod although the chance is low. But crashing istiod is something we want to try our best to avoid.

We haven't seen this on recent versions of Istio and Kiali, but I guess that "istioctl" command are not tested on high concurrent load. Kiali is caching this information, and also be sure that there is no concurrency when accessing to the Istiod service. But I'm going to review the latest changes in the 1.14 versions to see if there is progress in that side.

> Kiali overview page sends couple of request for each namespace, while having hundreds of namespaces, loading the overview page takes quite long.

Yeah, this is a known topic and the project is actively working on that, https://github.com/kiali/kiali/pull/5184 is a draft that would optimize the use of these queries in the overview page and it would address some of the scenarios you're seen.

Your feedback is very valuable, please keep in touch, as it can help us to drive these efforts.

lucasponce:
In the 1.14 version, I see that the access to the registries using the debug /registryz endpoint is locked:

image
image
image

So, it would be good to keep an eye on this and see if with Istio 1.14 and recent versions this access is mitigated.

Again, the design of Kiali is to not access concurrently to this endpoint, cache the information, and only refresh when there is a change in the Istio config, to ensure the use of that endpoint is minimal.

I know you mentioned that you can use Istio 1.14, but it would be interesting if you have some possibility to test the latest versions.

In any case, as commented, this is an active development area for optimizations.

zijianz:
I am seeing calling `endpointz` may also crashes istiod.

```
istio.io/istio/pilot/pkg/xds.writeJSON({0x36a2210, 0xc0d3622000}, {0x29e79a0, 0xc0cdb90810})
        istio.io/istio/pilot/pkg/xds/debug.go:939 +0xd6 fp=0xc054255710 sp=0xc0542556b0 pc=0x27c87b6
istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).endpointz(0xc000c606e0, {0x36a2210, 0xc0d3622000}, 0xc0d3608ed0)
        istio.io/istio/pilot/pkg/xds/debug.go:377 +0x165 fp=0xc0542558d0 sp=0xc054255710 pc=0x27c3925
istio.io/istio/pilot/pkg/xds.(*DiscoveryServer).endpointz-fm({0x36a2210, 0xc0d3622000}, 0xc0c3df1900)
```

I think everything uses `ServiceDiscovery.Services()` is not safe to be called? 
https://github.com/istio/istio/blob/release-1.14.1-patch/pilot/pkg/xds/debug.go#L371

Because the returned value itself is not concurrent safe?
https://github.com/istio/istio/blob/release-1.14.1-patch/pilot/pkg/serviceregistry/memory/discovery.go#L245

lucasponce:
Hi @zijianz,
I think that code that you mention should be protected, with the latest Kiali and Istio 1.14 you shouldn't see problems.
Let me know if you are reproducing the problem.

PR #4575, opened 2021-12-10
[hack] update to the latest Dex version so we can run on newer minikube

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4582

When this PR is merged, the molecule tests will fail until you also merge this PR: https://github.com/kiali/kiali-operator/pull/466


israel-hdez (approved):
I tried and all looks good :+1:

PR #4578, opened 2021-12-13
Fix and Optimize Gateway Host Duplicate check for hosts with target namespace

maxbischoff:
** Describe the change **

Up until now, the duplicate Gateway Host only considered Ports and Selectors for separating the Comparison Spaces for Gateways. No special consideration was given for Hosts in the `namespace/host` Syntax (see [hosts in Gateway.spec.servers](https://istio.io/latest/docs/reference/config/networking/gateway/#Server) for a detailed description of this syntax). 

This PR explicitly checks the namespace of a Gateway Host in the duplicate check and thereby accomplishes two improvements:

* The duplicate check for a Gateway in namespace `test` with the host `./example.com` and the host `test/example.com` (in the same Port / Selector) is now correctly identified as duplicate host.
* The number of regex-based comparisons is greatly reduced for Gateways that expose one host to multiple namespaces.

This comes at the small cost of computing the target Namespace for each Host once, and comparing the target Namespaces when iterating over hosts with the same selector.

** Issue reference **

Fixes https://github.com/kiali/kiali/issues/4586

** Backwards incompatible? **

The change is backwards compatible.

** Documentation **

No user-visible changes are introduced.

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -20,8 +20,10 @@ type MultiMatchChecker struct {
 }
 
 const (
-	GatewayCheckerType = "gateway"
-	wildCardMatch      = "*"
+	GatewayCheckerType     = "gateway"
+	wildCardMatch          = "*"
+	targetNamespaceAll     = "*"
+	targetNamespaceCurrent = "."
 )
 
 type Host struct {
@@ -31,6 +33,7 @@ type Host struct {
 	ServerIndex     int
 	HostIndex       int
 	GatewayRuleName string
+	TargetNamespace string
 }
 
 // Check validates that no two gateways share the same host+port combination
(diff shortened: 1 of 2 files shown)


lucasponce:
Thanks for the PR @maxbischoff cc-ing @hhovsepy to review this work as he's also currently working on the area.

hhovsepy (approved):
Code LGTM, verified locally and regression tested. Thank you @maxbischoff for the fix, can you please open an issue for this?

hhovsepy:
Tested all the combinations, good idea is to consider "*" in case of missing namespace:
Screenshot from 2021-12-14 09-49-57
Screenshot from 2021-12-14 09-50-25
Screenshot from 2021-12-14 09-50-54
Screenshot from 2021-12-14 09-51-15
Screenshot from 2021-12-14 09-51-36
Screenshot from 2021-12-14 09-55-43

skondkar (approved):
LGTM.
Verified with the namespace/host combinations. Also verified that running E2E tests is successful.

Screenshot from 2021-12-15 15-40-44

skondkar:
Polarion test case ->> SWSQE-2175

Issue #4485, opened 2021-11-08
[operator] implement best practice guidelines to support multi-tenant installations


jmazzitelli:
I believe these are the tasks needed in order to complete this issue:

- [x] support custom labels on the kiali route
- [x] ship with resource requests on the operator Deployment
- [x] ship with resource request defaults on the kiali pod

_NOTE: We are going to want to backport these changes to the 1.36 playbook and branch._

Running operators in multi-tenant OpenShift clusters - there are best practice guidelines to include these aspects:

Here's a distilled version of this:
* if your operator creates `Routes` it should allow users to specify custom labels so they can employ router sharding (very popular in larger multi-tenant clusters)
** WE NEED TO ADD THIS FEATURE TO THE OPERATOR and SERVER HELM CHART

* if your operator creates `NetworkPolicies` it should be possible to disable that or at least not create "accept traffic from everywhere" policies
** I don't think we create NetworkPolicies (we rely on OSSM to create those when needed), but we should confirm.

* your operator should ship with resource requests on its own `Deployment` (enables ResourceQuotas to be used)
** We support the user providing resource requests on the Kiali server, but not the operator itself. WE NEED TO ADD THIS ON THE OPERATOR ITSELF

* your operator should apply at least resource requests on all `Pods` it creates (enables ResourceQuotas and HorizontalPodAutoscalers to be used)
** We support being able to add HPAs and resources on the Kiali server. Therefore I think we support this already. But I do not think we set this by default on Kiali pods - I believe we need some defaults.

===

### Create resource requests

* for your operator's own `Pods`
* for `Pods` your operator creates / manages

The reasoning behind this is, to allow `ResourceQuota` configuration ([docs](https://kubernetes.io/docs/concepts/policy/resource-quotas/)) to work. This is something that every production cluster will follow as a best practice, to prevents tenants from being able to consume all resources on the cluster and block other worklaods from being scheduled. `ResourceQuota` in a namespace will cause all pods without `requests` ([docs](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits)) for CPU and RAM to be rejected.

Therefore, **always ship your operator `Deployment` with resource requests**. 

[...]

PR #4579, opened 2021-12-13
Automatically determine gomaxprocs

maxbischoff:
In our cluster with large nodes, Kiali is creating many more processes than given by our `limits.cpu` (104 procs vs 8 CPU limit). Through introduction of [automaxprocs](https://github.com/uber-go/automaxprocs) Kiali automatically sets its `GOMAXPROCS` to the kubernetes limits. I tested this change on our cluster and observed only 8 procs to be used by Kiali.


lucasponce (approved):
It sounds like a good idea, cc-ing @jmazzitelli and @FilipB as they have experience in setting parameters for production environments and I remember that Filip proposed some manual tweak of the GOMAXPROCS in some of our testing.

kiali.go
```diff
@@ -30,6 +30,8 @@ import (
 	"github.com/kiali/kiali/server"
 	"github.com/kiali/kiali/status"
 	"github.com/kiali/kiali/util"
+
+	_ "go.uber.org/automaxprocs"
```
jmazzitelli, line 34: just trivial issue - Kiali coding standards put third party libs in a second group (of three) of imports. So move this up in its own group on line 26, with newlines before and after.

```
...
	"strings"

	_ "go.uber.org/automaxprocs"

...
```

I have not validated this PR at all , I just quickly noticed the coding standard violation so I'm reporting it here :)

jmazzitelli (dismissed):

FilipB:
Please hold on with merging. Our perf tests show degradation and I need to doublecheck if it's cause by this PR.

maxbischoff:
> Please hold on with merging. Our perf tests show degradation and I need to doublecheck if it's cause by this PR.

What exactly did you test and in what environment? Are there sources for the perf tests?

FilipB:
> What exactly did you test and in what environment? Are there sources for the perf tests?

Currently we have following gatling tests - https://gitlab.cee.redhat.com/istio/servicemesh-qe/jenkins-csb-declaration/-/tree/master/resources/kiali/performance/gatling

Those tests are hitting different kiali REST API endpoints, measuring mean response times and comparing with baselines.
Tests are running against Kiali in our internal OCP cluster. Results are very unstable at the moment so I'm checking what is a cause of that instability.

maxbischoff:
Hmm, I cannot reach that gitlab. I think if your Kiali does not specify any cpu limits, this change should not affect it at all (because there won't be a Quota for automaxprocs to go by). Likewise, when you have set the `GOMAXPROCS` env var automaxprocs also does nothing.

jmazzitelli:
@FilipB you just posted a URL to internal .cee machines :)

jmazzitelli:
> I think if your Kiali does not specify any cpu limits...

That cannot be guaranteed, and as of the upcoming release, it won't be true by default. Users always had the ability to set CPU limits and today we set those by default as per published best practices. Please see:

* https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr.yaml#L418-L427
* https://github.com/kiali/kiali/issues/4485
* https://master.sdk.operatorframework.io/docs/best-practices/managing-resources/

maxbischoff:
> > I think if your Kiali does not specify any cpu limits...
> 
> That cannot be guaranteed, and as of the upcoming release, it won't be true by default. Users always had the ability to set CPU limits and today we set those by default as per published best practices. Please see:

Yes, that is clear to me, but that is also the case in which this change **should** affect GOMAXPROCS. I think setting the cpu limit should also mean that Kiali behaves accordingly and doesn't start too many processes for the given limit (e.g. in my case with large machines we had 104 processes for 8 cpu.limit). 

I just wanted to mention that the change will only affect you, if you set a CPU limit at all, in case your performance tests don't limit CPU, because if they don't this change shouldn't be the culprit.

FilipB:
Keeping perf tests aside for now. I tested following:
- no CPU limit set -> kiali logs: maxprocs: Leaving GOMAXPROCS=8: CPU quota undefined
- CPU limit set to 2 -> kiali logs: maxprocs: Updating GOMAXPROCS=2: determined from CPU quota
- CPU limit set to 500m -> kiali logs: maxprocs: Updating GOMAXPROCS=1: using minimum allowed GOMAXPROCS

So it behaves as expected.

@maxbischoff how do you get number of processes created by Kiali? Are you running something like 'ps -ef | grep kiali' on a worker node which runs kiali pod?

maxbischoff:
No, I actually looked at the pprof trace using my patch in https://github.com/maxbischoff/kiali/commit/69865f05f7b5247e256b5dc82d145e6bbd708f1f.
In the trace you can see the number of processes started.

Without pprof patched in, I think the easiest way is to print out the return value of `runtime.GOMAXPROCS(0)`, which is what [automaxprocs does in your output](https://github.com/uber-go/automaxprocs/blob/master/maxprocs/maxprocs.go#L111)

nrfox:
> No, I actually looked at the pprof trace using my patch in [maxbischoff@69865f0](https://github.com/maxbischoff/kiali/commit/69865f05f7b5247e256b5dc82d145e6bbd708f1f).

As an aside, that patch seems generally useful. I'll open up a separate issue for adding those debug endpoints.

FilipB (approved):

lucasponce (approved):

lucasponce:
@FilipB I've added the PR to the QE flow, but I guess you already reviewed it, so if QE is good, we could move both flows.
Thanks!

antoinedeschenes:
I'm investigating a throttling issue, looks like this was removed in #6100

Issue #4592, opened 2021-12-14
Number of regex.Compile() calls in multi_match_checker scales quadratically with hosts checked


maxbischoff:
Since we had problems with Kiali using very much CPU while still being very slow, I analyzed performance in our cluster using pprof.

I found that when checking for host conflicts in istio Gateways, every comparison of two hosts incorporates two regex.Compile() calls (one for each host name) that are quite slow. You can see the impact in my pprof graph (the regex Compiles make up 109 seconds of CPU time, of my 20 second capture, I assume that is because the code is run in parallel since I loaded multiple views in parallel) in the red box:

![profile002](https://user-images.githubusercontent.com/12122170/145985329-8bbcc9df-2173-4c13-bb3f-9a119d6578b1.png)

This problem is only really appearing when l having a large number of hosts in Gateways, and is already partly mitigated by https://github.com/kiali/kiali/pull/4578. In our multi-tenant setup we have a Gateway resource that exposes the same two hosts to many namespaces (playground namespaces for our users that can expose VirtualServices on a shared gateway in the same wildcard domain):

```yaml
- port:
    name: https
    number: 443
    protocol: HTTPS
  hosts:
    - "playground-user1/*.example-cluster1.com"
    - "playground-user1/*.example-mesh1.com"
    - "playground-user2/*.example-cluster1.com"
    - "playground-user2/*.example-mesh1.com"
    # ...
    - "playground-user800/*.example-cluster1.com"
    - "playground-user800/*.example-mesh1.com"
- port:
    name: http
    number: 80
    protocol: HTTP
  hosts: # same as above
```

With this configuration and https://github.com/kiali/kiali/pull/4578 already applied we still have ~800 comparisons of `*.example-cluster1.com` with `*.example-mesh1.com` that each incorporate two `regexp.Compile()` calls per Port.

This makes this issue much more apparent than I would expect it to be in other setups.
I have implemented a caching mechanism for `regexp.Regexp` objects that fixes the problem in my setup, a PR will follow shortly.

jshaughn:
Closing as the PR is merged but did not have the gh hook for auto-closing the issue.

PR #4593, opened 2021-12-14
Cache compiled regexes in multi match checker

maxbischoff:
** Describe the change **

When checking for duplicate hosts, `regexp.Compile()` is currently called twice per comparison. This is a costly operation, which is fine for clusters with few gateway hosts. In our setup we heavily utilize the namespacing of hosts, though, so we have many repeating host entries, causing many regexp Compilations on the same hosts.

This PR implements a caching mechanism for compiled `regexp.Regexp` that lives as long as the Validation is running. The change heavily reduces CPU load in our setup, and should have minimal impact on memory.

In this change I also extracted the String manipulations for converting hosts to regexes into their own function and found, that the replacement of `*` -> `.*` was never actually used in the regex and had no effect, so I removed it altogether.

** Issue reference **

https://github.com/kiali/kiali/issues/4592

** Backwards incompatible? **

No.

** Documentation **

I don't think its necessary.

### DIFF
--- a/business/checkers/gateways/multi_match_checker.go
+++ b/business/checkers/gateways/multi_match_checker.go
@@ -151,24 +151,24 @@ func (m MultiMatchChecker) findMatch(host Host, selector string) (bool, []Host) 
 						continue
 					}
 
-				    // lazily compile hostname Regex
-				    if host.hostnameRegexp == nil {
-				    	host.hostnameRegexp = regexpFromHostname(host.Hostname)
-				    }
-				    if h.hostnameRegexp == nil {
-				    	h.hostnameRegexp = regexpFromHostname(h.Hostname)
-				    }
+					// DNS is case-insensitive
+					current := strings.ToLower(host.Hostname)
+					previous := strings.ToLower(h.Hostname)
 
-				    // Either one could include wildcards, so we need to check both ways and fix "*" -> ".*" for regexp engine
-				    current := strings.ToLower(strings.Replace(host.Hostname, "*", ".*", -1))
-				    previous := strings.ToLower(strings.Replace(h.Hostname, "*", ".*", -1))
+					// lazily compile hostname Regex
+					if host.hostnameRegexp == nil {
+						host.hostnameRegexp = regexpFromHostname(current)
+					}
+					if h.hostnameRegexp == nil {
+						h.hostnameRegexp = regexpFromHostname(previous)
+					}
 
-				    if host.hostnameRegexp.MatchString(previous) ||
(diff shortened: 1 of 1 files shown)


xeviknal:
@maxbischoff thanks for the contribution and, at least for me, to showing this performance improvement. I've just allowed the CI to build and run the test. The code looks pretty good to me. Lazy loading the regexp and compiling them only once.

It is incredible that reduces the CPU load heavily.

business/checkers/gateways/multi_match_checker.go
```diff
@@ -161,3 +163,17 @@ func (m MultiMatchChecker) findMatch(host Host, selector string) (bool, []Host)
...
+func regexpFromHostname(hostname string) *regexp.Regexp {
+	// Escaping dot chars for RegExp. Dot char means all possible chars.
+	// This protects this validation to false positive for (api-dev.example.com and api.dev.example.com)
+	escaped := strings.Replace(hostname, ".", "\\.", -1)
+
+	// We anchor the beginning and end of the string when it's
+	// to be used as a regex, so that we don't get spurious
+	// substring matches, e.g., "example.com" matching
+	// "foo.example.com".
+	anchored := strings.Join([]string{"^", escaped, "$"}, "")
```
xeviknal, line 176: I thought that using `fmt.Sprintf("^%s$", escaped)` would go faster but it doesn't. Actually, it make some sense because the fmt needs to find characters within a string while the Join just appends whatever in the array into the string. No lookup necessary. I guess.
So no changes need it.

xeviknal (approved):
LGTM!

maxbischoff:
> @maxbischoff thanks for the contribution and, at least for me, to showing this performance improvement. I've just allowed the CI to build and run the test. The code looks pretty good to me. Lazy loading the regexp and compiling them only once.
> 
> It is incredible that reduces the CPU load heavily 🥇

Yes, altough the large improvements only come from the configs we have in our cluster, when only a few gateways with not that many hosts exist, the potential performance gain is of course much smaller. A large part of the savings also comes from https://github.com/kiali/kiali/pull/4578 which reduces the number of checks in our config drastically.

Thanks for taking the time to look at this :)

hhovsepy:
For QE: Regression testing is required for Gateway's validations.

hhovsepy (approved):
Code LGTM, checked locally.

maxbischoff:
Had to rebase on master since these changes conflicted with https://github.com/kiali/kiali/pull/4578

hhovsepy (approved):

pbajjuri20:
Rest Automation Run successfully against this PR.

Issue #4542, opened 2021-12-03
Auth: Phase out usage of JWTs


israel-hdez:
Most Kiali's authentication mechanisms are using own JWTs.

In the past, usage of JWTs was introduced as a mechanism to avoid credential forging. The UI stored the token and it was handy to have some data in plain text form because the UI could extract by its own means some data of the token (like the username) to display on the UI or for any purposes. Then, the signature of the JWT prevented credential forging. However, using javascript mechanisms to save the token and persist the session had a few problems at the moment of killing or starting sessions which were solved by moving session persistence to cookies.

Today, the UI does not have access to the token. So, it needs to make a request to the back-end to check if there is any active session and the back-end will reply with the data that the UI used to resolved by its own in the past. Also, the JWT library that Kiali is using is now unmaintained. So, current usage of JWTs to persist the session is mainly because of historical reasons and can be seen as technical debt. It is probably desirable to move away from the unmaintained JWT library.

The proposed alternative would be to use native golang crypto packages to set an encrypted cookie, like in [the OpenID authorization code strategy](https://github.com/kiali/kiali/blob/cef36d4f69f06ea05db766027084d00d3cd31da9/handlers/authentication.go#L980-L1064). It has the added benefit that session data will no longer be transferred in plain text and can be used to store sensitive data (which is already being done for some auth strategies). Session data will be readable only at the back-end side, provided that users keep secured the `signing_key`.

jshaughn:
This has been a source of security concern, adding to the backlog.

PR #4596, opened 2021-12-14
Remove usage of the dgrijalva/jwt library for the Token authentication

israel-hdez:
This PR:
* Moves the token authentication code to it's own module.
  * Removes usage of the dgrijalva/jwt library on the code handling token authentication.
* Creates a "cookie session persistor", to separate concerns.
* Hopefully, this adds better tests.

Related to #4542


israel-hdez:
@nrfox I choose you as my new disciple for all auth-related code.

business/authentication/session_persistor.go
```diff
@@ -0,0 +1,335 @@
...
+	Strategy  string    `json:"strategy"`
+	ExpiresOn time.Time `json:"expiresOn"`
+	Payload   string    `json:"payload,omitempty"`
+}
+
+// CreateSession starts a user session using HTTP Cookies for persistance across HTTP requests.
+// For improved security, the data of the session is encrypted using the AES-GCM algorithm and
+// the encrypted data is what is sent in cookies. The strategy, expiresOn and payload arguments
+// are all required.
+func (p CookieSessionPersistor) CreateSession(_ *http.Request, w http.ResponseWriter, strategy string, expiresOn time.Time, payload interface{}) error {
```
nrfox, line 39: I don't see where the cookie data is persisted on the backend in this method? I was expecting `CreateSession` to persist the cookie data somehow on the backend (e.g. in memory or on the filesystem) and then call `http.SetCookie(sessionID)` returning just the session id to the client. Instead all of the session data is being persisted client side by returning the full session data in the `SetCookie(chunksCookie)` call.

Perhaps we can get around needing to chunk the cookie data if the backend just returns the session id instead of the full b64 encoded/encrypted cookie contents and saves the encrypted cookie contents on the backend?

If we want to bring in another external dependency, there might be some libraries that handle saving/reading on the backend either in memory or to disk such as: https://pkg.go.dev/github.com/gorilla/sessions.

business/authentication/auth_controller.go
```diff
@@ -0,0 +1,47 @@
...
+// user sessions and any authentication related requests.
+func GetAuthController() AuthController {
+	return authController
+}
+
+// InitializeAuthenticationController initializes the authentication controller associated to the
+// given strategy and prepares it to control user sessions and handle authentication requests.
+// This should be called during Kiali startup, before starting to listen to HTTP requests.
+func InitializeAuthenticationController(strategy string) {
+	persistor := CookieSessionPersistor{}
```
nrfox, line 42: Changing strategies should also invalidate any existing cookie sessions in case a user were to try and send an old session id with the new auth strategy.

nrfox (commented):
@israel-hdez had a couple of questions. The comments and code quality made this much easier to review!

business/authentication/auth_controller.go
```diff
@@ -0,0 +1,47 @@
...
+// AuthController is the interface that all Kiali authentication strategies should implement.
+// An authentication controller is initialized during Kiali startup.
+type AuthController interface {
+	// Authenticate handles an HTTP request that contains credentials. The method to pass the credentials
+	// is chosen by the authentication controller implementation. The credentials are verified and if
+	// it is supported by the controller, RBAC permissions are verified to ensure that the logging in user
+	// has enough privileges to login to Kiali.
+	// An AuthenticationFailureError is returned if the authentication request is rejected (unauthorized). Any
+	// other kind of error means that something unexpected happened.
+	Authenticate(r *http.Request, w http.ResponseWriter) (*UserSessionData, error)
```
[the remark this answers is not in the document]
israel-hdez: Yes. Why you ask? :-)

business/authentication/auth_controller.go
israel-hdez: Yes. Although that's done once a request arrives, because the browser is the storage (because of the cookies).
That's the intention of this line: https://github.com/kiali/kiali/pull/4596/files#diff-a6762b12addd1f51f23491d16825ae2746eff746dbba3e9528015d0fc0913d7aR243

business/authentication/session_persistor.go
israel-hdez: Sending only _SIDs_ is my preference. However, as you say, back-end storage is needed.

Both memory and pod filesystem storage are volatile in Kubernetes. A pod restart would kill all sessions... and it can be benign: K8s may be trying to optimize resource usage by relocating the Kiali pod to another node.

So, we need something capable of surviving a pod restart and I can only think on out-of-pod storage. Most folks on the team are against adding storage because of the perception that storage in k8s is hard to manage. I don't have the needed experience to reject that but if you can convince the team that storage isn't bad, let's start yet another round of debate ;-)

israel-hdez: BTW, that gorilla/sessions library looks good. However, I looked at the code, and it lacks support splitting large payloads into multiple cookies :-(

nrfox: > Both memory and pod filesystem storage are volatile in Kubernetes. A pod restart would kill all sessions... and it can be benign: K8s may be trying to optimize resource usage by relocating the Kiali pod to another node.

Yes if the pod gets rescheduled onto a different node it'd wipe the ephemeral storage. That's probably not a great UX to lose your session whenever that happens. On the other hand, part of the tradeoff of storing sessions in the browser is that they cannot be invalidated on the backend.

> Most folks on the team are against adding storage because of the perception that storage in k8s is hard to manage.

I am one of them :). I wonder if it'd be feasible to store session data in a configmap like:
```
data:
  "sid-1": {}
  "sid-2": {}
```
Since Kiali already talks to the kube API, the integration would be straightforward. There's apparently a 1mb limit on the amount of data you can store in a configmap though. Either way, changing how Kiali stores sessions is probably out of scope for this PR.

business/authentication/auth_controller.go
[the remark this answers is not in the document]
nrfox: I wasn't sure if `Authenticate` should call `w.WriteHeader(401)` or if it's just expected to return `AuthFailureErr` and the caller handles `w.WriteHeader(401)`.

business/authentication/token_auth_controller.go
```diff
@@ -0,0 +1,214 @@
...
+
+	// Create a bs layer with the received token to check its validity.
+	bs, err := c.businessInstantiator(&api.AuthInfo{Token: token})
+	if err != nil {
+		return nil, fmt.Errorf("error instantiating the business layer: %w", err)
+	}
+
+	// Using the namespaces API to check if token is valid. In Kubernetes, the version API seems to allow
+	// anonymous access, so it's not feasible to use the version API for token verification.
+	nsList, err := bs.Namespace.GetNamespaces()
```
nrfox, line 132: Will this work if the user does not have access to list all namespaces? Don't we expect some users to not have this permission since Kiali can be configured with `AccessibleNamespaces`?

business/authentication/token_auth_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		c.SessionStore.TerminateSession(r, w)
+		return nil, &AuthenticationFailureError{Reason: "token is not valid or is expired", Detail: err}
+	}
+
+	// If namespace list is empty, return authentication failure.
+	if len(nsList) == 0 {
```
nrfox, line 139: Does kube return an empty list if the user does not have permissions to list all namespaces? Otherwise I don't think this situation is possible given that if Kiali is deployed then there must be a namespace it is deployed in.

business/authentication/token_auth_controller.go
```diff
@@ -0,0 +1,214 @@
...
+// is done: only token validity is re-checked by making a request to the Kubernetes API, like in the Authenticate
+// function. However, privileges are not re-checked.
+// If the session is still valid, a populated UserSessionData is returned. Otherwise, nil is returned.
+func (c tokenAuthController) ValidateSession(r *http.Request, w http.ResponseWriter) (*UserSessionData, error) {
+	// Restore a previously started session.
+	sPayload := sessionPayload{}
+	sData, err := c.SessionStore.ReadSession(r, w, &sPayload)
+	if err != nil {
+		log.Warningf("Could not read the session: %v", err)
+		return nil, nil
```
nrfox, line 165: Should `ValidateSession` return the `err` here indicating to the caller the session could not be validated? It looks like an error will get returned if the business layer could not be instantiated but not in other scenarios where the session could not be validated.

business/authentication/token_auth_controller.go
```diff
@@ -0,0 +1,214 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("could not get the business layer: %w", err)
+	}
+
+	_, err = bs.Namespace.GetNamespaces()
+	if err == nil {
+		// Internal header used to propagate the subject of the request for audit purposes
+		r.Header.Add("Kiali-User", extractSubjectFromK8sToken(sPayload.Token))
+		return &UserSessionData{ExpiresOn: sData.ExpiresOn, Username: extractSubjectFromK8sToken(sPayload.Token), Token: sPayload.Token}, nil
+	}
```
nrfox, line 182: nit: can this be moved to the end of the function and the check be changed to `err != nil` as the others above it?

business/authentication/session_persistor.go
```diff
@@ -0,0 +1,335 @@
...
+	}
+
+	// Check that the session has not expired.
+	// This is just a sanity check, because browser cookies are set to expire at this date and the browser
+	// shouldn't send expired cookies.
+	if !util.Clock.Now().Before(sData.ExpiresOn) {
+		log.Tracef("Session is invalid because it expired on %s", sData.ExpiresOn.Format(time.RFC822))
+		p.TerminateSession(r, w) // Clean the expired session
+
+		return nil, nil
```
nrfox, line 255: Should these return errors? I wonder if the client will still read the `SetCookie` header if the resp is not a 2xx.

nrfox (commented):
Couple more questions/nits

business/authentication/auth_controller.go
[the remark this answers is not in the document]
israel-hdez: I wanted to do that way and handle the response here. But I missed those `RespondWithJSONIndent`, `RespondWithError` and `RespondWithDetailedError` functions. All these functions are in the `handlers` package, so I can't use them because it would cause an import loop handlers -> authentication -> handlers.

I wanted to extract these funcs to the utils/http folder, but they are widely used in the handlers folder and moving them would cause a lot of churn.

Any better suggestions?

business/authentication/session_persistor.go
```diff
@@ -0,0 +1,335 @@
...
+
+	return nil
+}
+
+// ReadSession restores (decrypts) and returns the data that was persisted when using the CreateSession function.
+// If a payload is provided, the original data is parsed and stored in the payload argument. As part of restoring
+// the session, validation of expiration time is performed and no data is returned assuming the session is stale.
+// Also, it is verified that the currently configured authentication strategy is the same as when the session was
+// created.
+func (p CookieSessionPersistor) ReadSession(r *http.Request, w http.ResponseWriter, payload interface{}) (*sessionData, error) {
```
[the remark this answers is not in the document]
israel-hdez: It can be. In this particular case, I don't have preference of one over the other.

I can change it if you think it's better the pointer receiver.

business/authentication/token_auth_controller.go
```diff
@@ -0,0 +1,214 @@
...
+// Kiali's "token" authentication strategy. It assumes that the
+// user will use a token that is valid to be used against the Cluster API.
+// In it's simplest form, it can be a ServiceAccount token. However, it can
+// be any kind of token that can be passed using HTTP Bearer authentication
+// in requests to the Kubernetes API.
+type tokenAuthController struct {
+	// businessInstantiator is a function that returns an already initialized
+	// business layer. Normally, it should be set to the business.Get function.
+	// For tests, it can be set to something else that returns a compatible API.
+	businessInstantiator func(authInfo *api.AuthInfo) (*business.Layer, error)
```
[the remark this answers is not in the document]
israel-hdez: No, AFAIK...

I think we only have this large struct: https://github.com/kiali/kiali/blob/fcbcfa02179c782577d3fa5ad3f618f7671c8be0/business/layer.go#L17-L37

It's just fields, no methods on it. So, an iface may not solve anything... :-/

business/authentication/token_auth_controller.go
israel-hdez: Yes, it works. And yes, we users without permissions to list namespaces are covered.

The internals of that `GetNamespace` function takes care of that case.

business/authentication/token_auth_controller.go
israel-hdez: Not kube directly. But yes, an empty list can be returned.

That `bs.Namespace.GetNamespaces()` is a Kiali's function. Open it and you will see an interesting code playing with tokens :smile:

business/authentication/token_auth_controller.go
israel-hdez: Uhm... Yes... but I don't know. Hmm...

business/authentication/session_persistor.go
israel-hdez: AFAIK, the `SetCookie` should be read even if the response is not 2xx.

But as the comment says, it should be just for sanity. The expiration time field of the sessionData and the browser cookie expiration time should both be synced. If for some reason the `SetCookie` is not read, the browser should still discard the cookies by it's own.

Only a forged cookie could have unsynced times. The benign case is a machine with a clock behind the real time, but still the caller should transparently see no active session rather than an error.

So IMHO, no error should be generated and the caller should simply "see" that there is no active session.

business/authentication/token_auth_controller.go
israel-hdez: OK, I reviewed what can throw an error inside the ReadSession function. It looks fine to return the error, so I changed it.

business/authentication/token_auth_controller.go
israel-hdez: Yes

israel-hdez:
@nrfox I think I fixed all your comments.
Only a few unsolved discussions...

business/authentication/auth_controller.go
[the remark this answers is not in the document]
nrfox: This would probably cause a lot of churn too but perhaps moving the `authentication` package to `handlers`? It seems more related to handling auth for http requests and less related to business specific logic anyways.

business/authentication/session_persistor.go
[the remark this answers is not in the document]
nrfox: I wouldn't say pointer is better here, mostly just curious since I usually default to pointers for structs but I don't have a great reason for that other than cargo culting. `CookieSessionPersistor` is small and if it's going to be used concurrently then non-pointer might be better anyways.

business/authentication/auth_controller.go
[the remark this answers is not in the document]
nrfox: Not a blocker. You can decide if you want to keep it the way it is or if you'd like to change it.

nrfox (approved):

business/authentication/auth_controller.go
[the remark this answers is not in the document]
israel-hdez: I don't think it will cause more churn, because it's a new package.
Anyway, I would like to leave it as is in this PR.

Soon, I'll have ready the OpenID refactor. Once you review the upcoming one, if you still think it belongs to `handlers`, then I can move all this code back to `handlers`.

israel-hdez:
@mattmahoneyrh No bandwidth to check this PR?
If not, let me know. Since we are at the beginning of the sprint, I think this can be merged and I hope we should see issues as people uses the code.

pbajjuri20:
I Have validated with Authentication: Token, OpenShift and Anonymous.

The regression test also passed.

pbajjuri20 (approved):

israel-hdez:
@nrfox @pbajjuri20 Thanks for your reviews.

Issue #4608, opened 2021-12-16
App Details is missing services with multi-label selector


EronWright:
**Problem Statement**

Kiali seems to be unable to locate the services that are associated with an application, when the services use multiple labels in their selector.  The services do appear as expected on the associated workload details page.

Here's an example:
![image](https://user-images.githubusercontent.com/1775518/146446959-705f49f5-7f8b-4af7-a450-9dbd5a1ad5b1.png)
![image](https://user-images.githubusercontent.com/1775518/146447035-b990e469-2324-4533-963d-387f53dddd72.png)

One of the services that Kiali fails to locate:
```yaml
apiVersion: v1
kind: Service
metadata:
  name: tarrax-zk
  namespace: avengers
spec:
  selector:
    cloud.streamnative.io/component: zookeeper
    service.istio.io/canonical-name: zookeeperset-tarrax
```

And the associated stateful set.
```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: tarrax-zk
  namespace: avengers
spec:
  template:
    metadata:
      labels:
        cloud.streamnative.io/component: zookeeper
        cloud.streamnative.io/zookeeper-name: tarrax
        service.istio.io/canonical-name: zookeeperset-tarrax
        service.istio.io/canonical-revision: 3.7.0
```

Note that Kiali is configured to use `service.istio.io/canonical-name` as the app label.

Expected: app details page to show relevant services.
Actual: app details page shows "No services found".

**Probable Cause**

The root cause appears to be that the `AppService` misuses the `ServiceCriteria.ServiceSelector` field when invoking `SvcService.GetServiceList`.  The `ServiceSelector` field seems designed to filter services for a given workload, and is expected to contain pod labels.  `GetServiceList` locates the services whose selector would match those labels.  To build the app details page, the `AppService` erroneously populates `ServiceSelector` with the app label (which will not match since the service's selector also looks for other labels):
https://github.com/kiali/kiali/blob/33639196cae6e5ef689dc702046f09eca0b68568/business/apps.go#L249-L253

**Suggested Fix**
1. don't use `ServiceSelector`; get all services using `GetServiceList` and then filter using `GetServiceAppName`.
2. locate the workloads first, and then locate the services that match those workloads.

lucasponce:
Thanks @EronWright, yes, I see the issue, multiple labels on the selector can create this bug.

[...]

PR #4613, opened 2021-12-20
Changed the loading App to first load Workloads then Services.

hhovsepy:
Fix for issue https://github.com/kiali/kiali/issues/4608

Before Kiali was loading Services for App pages by "app" label name using in Service selector, and in a case when AppLabelName and Service selector were not equal, the Services for App details page was showing empty.

Now with this change the Workloads for App are loaded first, then per workload the Services are loaded by using Workload labels as Service Selector.

One thing in this PR to discuss is loading Services in the loop of workloads then adding them into Service list of results after checking their duplicity.

Screenshot from 2021-12-20 12-11-59

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -206,36 +206,34 @@ type appDetails struct {
 // NamespaceApps is a map of app_name x AppDetails
 type namespaceApps = map[string]*appDetails
 
-func castAppDetails(ss *models.ServiceList, ws models.Workloads) namespaceApps {
-	allEntities := make(namespaceApps)
+func castAppDetails(allEntities namespaceApps, ss *models.ServiceList, w *models.Workload) {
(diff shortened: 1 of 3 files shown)


lucasponce (approved):
Normal cases are ok, and I think this fix can bring more flexibility for corner cases.

I've performed an "artificial" example, like using a service for all "v1":

```
apiVersion: v1
kind: Service
metadata:
  name: v1-front
  labels:
    version: v1
spec:
  ports:
    - name: http
      port: 8000
  selector:
    version: v1
```

And the result shows correct:
image

image

The example is "artificial" as the "v1-front" service has not any sense:
image

But I think it helps to illustrate the scenario and can be used for QE for testing.

prachiyadav (approved):

EronWright:
Thanks @hhovsepy

PR #4614, opened 2021-12-20
optimize: remove unused sprintf and replace value with reference

Xunzhuo:
**Describe the change**

optimize kiali codes and fix go-staticcheck errors

### DIFF
--- a/graph/telemetry/istio/appender/aggregate_node_test.go
+++ b/graph/telemetry/istio/appender/aggregate_node_test.go
@@ -105,7 +105,7 @@ func TestNamespacesGraphWithServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeAggregate, pp.Edges[1].Dest.NodeType)
 
 	topReviews := pp.Edges[0].Dest
-	if topReviews.Metadata[graph.AggregateValue] != "Top" {
+	if "Top" != topReviews.Metadata[graph.AggregateValue] {
 		topReviews = pp.Edges[1].Dest
 	}
 	assert.Equal("request_operation", topReviews.Metadata[graph.Aggregate])
@@ -116,7 +116,7 @@ func TestNamespacesGraphWithServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeService, topReviews.Edges[0].Dest.NodeType)
 
 	allReviews := pp.Edges[1].Dest
-	if allReviews.Metadata[graph.AggregateValue] != "All" {
+	if "All" != allReviews.Metadata[graph.AggregateValue] {
 		allReviews = pp.Edges[0].Dest
 	}
 	assert.Equal("request_operation", allReviews.Metadata[graph.Aggregate])
@@ -231,7 +231,7 @@ func TestNamespacesGraphNoServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeAggregate, pp.Edges[1].Dest.NodeType)
 
 	topReviews := pp.Edges[0].Dest
-	if topReviews.Metadata[graph.AggregateValue] != "Top" {
+	if "Top" != topReviews.Metadata[graph.AggregateValue] {
 		topReviews = pp.Edges[1].Dest
 	}
 	assert.Equal("request_operation", topReviews.Metadata[graph.Aggregate])
@@ -241,7 +241,7 @@ func TestNamespacesGraphNoServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeApp, topReviews.Edges[0].Dest.NodeType)
 
 	allReviews := pp.Edges[1].Dest
-	if allReviews.Metadata[graph.AggregateValue] != "All" {
+	if "All" != allReviews.Metadata[graph.AggregateValue] {
 		allReviews = pp.Edges[0].Dest
 	}
 	assert.Equal("request_operation", allReviews.Metadata[graph.Aggregate])
--- a/graph/telemetry/istio/appender/aggregate_node_test.go
+++ b/graph/telemetry/istio/appender/aggregate_node_test.go
@@ -105,7 +105,7 @@ func TestNamespacesGraphWithServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeAggregate, pp.Edges[1].Dest.NodeType)
 
 	topReviews := pp.Edges[0].Dest
-	if "Top" != topReviews.Metadata[graph.AggregateValue] {
+	if topReviews.Metadata[graph.AggregateValue] != "Top" {
 		topReviews = pp.Edges[1].Dest
 	}
 	assert.Equal("request_operation", topReviews.Metadata[graph.Aggregate])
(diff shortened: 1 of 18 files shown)


jshaughn (commented):
In general I'm not a fan of making sweeping changes based on static analysis, as almost any change has a chance at introducing a regression.  Having said that, the proposed changes look OK to me.   But I'd like to know what other maintainers think about this sort of PR submission.

jmazzitelli:
+1 to @jshaughn -- most of these are useless churn. Changing to lowercase and removing punctuation is unnecessary and I'm -1. But hidden in here might be some we could do. I saw one that changes from pass-by-value to pass-by-reference.

The only way we should do something like this is if we add into our CI or commit checks automatic running of this "optimizer tool" (and frankly, changing case and punctuation in log messages isn't optimizing anything - its just changing it to look different).

jaeger/model/keyvalue.go
```diff
@@ -210,7 +210,7 @@ func (kvs KeyValues) Hash(w io.Writer) error {
 }
 
 // Hash implements Hash from Hashable.
-func (kv KeyValue) Hash(w io.Writer) error {
+func (kv *KeyValue) Hash(w io.Writer) error {
```
jmazzitelli, line 213: This is the only thing we should look at doing. I'm not sure if pass-by-reference would be valid here in this use case, but typically, pass-by-reference is what we want.

My suggestion - strip everything out of this PR of the useless churn (changing uppercase to lowercase and punctuation in log messages) and keep useful changes like this. Then we can review further.

jmazzitelli (changes requested):

lucasponce:
> In general I'm not a fan of making sweeping changes based on static analysis, as almost any change has a chance at introducing a regression. Having said that, the proposed changes look OK to me. But I'd like to know what other maintainers think about this sort of PR submission.

+1

I understand that fixing minor issues proposed by a linter tool or similar are easy for new comers.

I'd say that if we want to introduce a new checker, the way should be to add it in the ci, I think that "go-staticcheck" is optional.

Personally I guess we need to help new comers to take issues from the "good first issue" / "help wanted" or another task where the new comer may feel comfortable.

We want to grow a community and on-board new members, so I prefer spend more time mentoring a developer to perform a more rich feature rather than see minor changes on strings or similar.

@Xunzhuo in any case, we are happy to have the proposal PR, if you like we could also to help you to focus on some feature as well.

jaeger/model/keyvalue.go
Xunzhuo: Thanks @jmazzitelli, I will update it later :)

Xunzhuo:
@lucasponce Thanks, this is what I want! Actually, I just want to help but I do not know how and where to get started.
Good first issue is nice to new comers. :)

jaeger/model/keyvalue.go
jshaughn: I would also upvote the changes to remove the unnecessary fmt.Sprintf instances.

lucasponce:
@Xunzhuo 

https://github.com/kiali/kiali/labels/help%20wanted
https://github.com/kiali/kiali/labels/good%20first%20issue

or 

Any issue in the list that is not assigned can be a good start, we understand that it requires on-boarding and mentoring time from maintainers but it's worth if you'd like to contribute to the project in a long term.

I'd say that having a proper environment, trying to perform changes in both backend/frontend and deploy via operator into a local cluster i.e. minikube could be the first steps.

And from that, the list of issues is long and then we can help guiding new developers in which areas they can contribute.

jaeger/model/keyvalue.go
Xunzhuo: Could we move forward? PR updated. @jmazzitelli @jshaughn

graph/telemetry/istio/appender/aggregate_node_test.go
```diff
@@ -105,7 +105,7 @@ func TestNamespacesGraphWithServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeAggregate, pp.Edges[1].Dest.NodeType)
 
 	topReviews := pp.Edges[0].Dest
-	if "Top" != topReviews.Metadata[graph.AggregateValue] {
+	if topReviews.Metadata[graph.AggregateValue] != "Top" {
```
jmazzitelli, line 108: i don't understand the purpose of this churn. There's a few of these. These are identical statements in the end. I would like to see these gone.

graph/telemetry/istio/appender/aggregate_node_test.go
```diff
@@ -116,7 +116,7 @@ func TestNamespacesGraphWithServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeService, topReviews.Edges[0].Dest.NodeType)
 
 	allReviews := pp.Edges[1].Dest
-	if "All" != allReviews.Metadata[graph.AggregateValue] {
+	if allReviews.Metadata[graph.AggregateValue] != "All" {
```
jmazzitelli, line 119: see above

graph/telemetry/istio/appender/aggregate_node_test.go
```diff
@@ -231,7 +231,7 @@ func TestNamespacesGraphNoServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeAggregate, pp.Edges[1].Dest.NodeType)
 
 	topReviews := pp.Edges[0].Dest
-	if "Top" != topReviews.Metadata[graph.AggregateValue] {
+	if topReviews.Metadata[graph.AggregateValue] != "Top" {
```
jmazzitelli, line 234: see above

graph/telemetry/istio/appender/aggregate_node_test.go
```diff
@@ -241,7 +241,7 @@ func TestNamespacesGraphNoServiceInjection(t *testing.T) {
 	assert.Equal(graph.NodeTypeApp, topReviews.Edges[0].Dest.NodeType)
 
 	allReviews := pp.Edges[1].Dest
-	if "All" != allReviews.Metadata[graph.AggregateValue] {
+	if allReviews.Metadata[graph.AggregateValue] != "All" {
```
jmazzitelli, line 244: see above

jmazzitelli (dismissed):

graph/telemetry/istio/appender/aggregate_node_test.go
Xunzhuo: Done

graph/telemetry/istio/appender/aggregate_node_test.go
Xunzhuo: Done

graph/telemetry/istio/appender/aggregate_node_test.go
Xunzhuo: Done

jmazzitelli:
This looks good now; @lucasponce you authored this file `jaeger/model/keyvalue.go` (according to git blame) - perhaps you should check this out and make sure the pass-by-reference is good. I'm not sure what parts of the code uses this keyvalue.go code.

lucasponce (changes requested):
I'd remove the "jaeger/model/keyvalue.go" changes from this PR.
Those classes are cloned from the jaeger project to support the legacy http1 protocol.

While testing, I see a problem in that protocol in master, but it's not related to this work, so I'd suggest not to include those changes until we fix it in master.

Thanks

Xunzhuo:
updated and thanks @lucasponce

lucasponce (approved):

lucasponce (approved):
Trivial change.
No QE is required.

PR #4616, opened 2021-12-20
fix: remove -i flag for its deprecated

Xunzhuo:
**Describe the change**

fix: remove -i flag for its deprecated

The -i flag is deprecated. Compiled packages are cached automatically.

image


jmazzitelli:
@israel-hdez do we use this `build-test` target in our jenkins CI jobs? It is being requested that we remove that target in this PR.

israel-hdez:
@jmazzitelli Nope, not used for the releases.

israel-hdez (approved):
I never had used this target. So, I'm OK with this removal OK.

jmazzitelli (approved):

PR #4619, opened 2021-12-20
Filter pods based on full owner ref name

### LINKED ISSUE #4477: custom controllers still have problems
I use kiali operator to install kiali。
version：1.36.5

when I use own custom controllers for replicaSet called DnaApp 
that happened

show the one replicaSet's ownerReferences
image

I am sorry that I provided the wrong version information.  please seem the latest
https://github.com/kiali/kiali/issues/4477#issuecomment-969971594

nrfox:
When fetching workloads, Kiali should match the replicaset owner ref's full name instead of relying on a prefix.

Fixes #4477

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -1063,27 +1063,36 @@ func fetchWorkloads(layer *Layer, namespace string, labelSelector string) (model
 				cnFound = false
 			}
 		default:
+			// Two scenarios:
+			// 1. Custom controller with replicaset
+			// 2. Custom controller without replicaset controlling pods directly.
+			//
 			// ReplicaSet should be used to link Pods with a custom controller type i.e. Argo Rollout
-			cPods := kubernetes.FilterPodsByController(cname, kubernetes.ReplicaSetType, pods)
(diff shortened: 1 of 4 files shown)


nrfox:
Tested with argo rollouts and did not find a regression there with this patch applied:
Screenshot from 2021-12-20 14-43-20

lucasponce (approved):
The PR looks correct.
It's not easy to deal with custom controllers as those APIs are out of the standard, and it doesn't scale to use them.

So, this solution is reasonably good.

Thinking beyond, as this issue is not new, we could normalize to ask users to help Kiali using specific labels/annotations for corner cases like custom controllers.

It's not something new as Istio is also using something similar:
https://github.com/istio/istio/issues/36155#event-5670107235

Anyway, typically custom controllers have a minimal impact on the data plane, and this solution hopefully will work for them.

pbajjuri20:
Tested and Passed on Kiali UI 1.45.0 (50158c3dfa71598572dd1c945d78cbcf287390bc)

Thanks, @nrfox for the Instructions.

pbajjuri20 (approved):

PR #4621, opened 2021-12-21
Replace ServiceList usage by RegistryServices

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4528

Changed several validations to use ServiceRegistry Instead of ServiceList to decrease the validation time and optimize the validation logic

Touched validation messages:
AuthPolicy: "KIA0104 This host has no matching entry in the service registry"
VirtualService: "KIA1101 DestinationWeight on route doesn't have a valid service (host not found)"

Full regression testing is required for those validations.
Regression testing is required for all other validations.

### DIFF
--- a/business/checkers/authorization/no_host_checker.go
+++ b/business/checkers/authorization/no_host_checker.go
@@ -7,6 +7,7 @@ import (
 	networking_v1alpha3 "istio.io/client-go/pkg/apis/networking/v1alpha3"
 	security_v1beta "istio.io/client-go/pkg/apis/security/v1beta1"
 
+	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 )
@@ -16,7 +17,6 @@ type NoHostChecker struct {
 	Namespace           string
 	Namespaces          models.Namespaces
 	ServiceEntries      map[string][]string
-	ServiceList         models.ServiceList
 	VirtualServices     []networking_v1alpha3.VirtualService
 	RegistryServices    []*kubernetes.RegistryService
 }
@@ -49,7 +49,10 @@ func (n NoHostChecker) validateHost(ruleIdx int, to []*api_security_v1beta.Rule_
 	if len(to) == 0 {
 		return nil, true
 	}
-
+	namespace, clusterName := n.AuthorizationPolicy.Namespace, n.AuthorizationPolicy.ClusterName
+	if clusterName == "" {
+		clusterName = config.Get().ExternalServices.Istio.IstioIdentityDomain
+	}
 	checks, valid := make([]*models.IstioCheck, 0, len(to)), true
 	for toIdx, t := range to {
 		if t == nil {
@@ -65,8 +68,8 @@ func (n NoHostChecker) validateHost(ruleIdx int, to []*api_security_v1beta.Rule_
 		}
 
 		for hostIdx, h := range t.Operation.Hosts {
-			fqdn := kubernetes.GetHost(h, n.AuthorizationPolicy.Namespace, n.AuthorizationPolicy.ClusterName, n.Namespaces.GetNames())
-			if !n.hasMatchingService(fqdn, n.AuthorizationPolicy.Namespace) {
+			fqdn := kubernetes.GetHost(h, namespace, clusterName, n.Namespaces.GetNames())
+			if !n.hasMatchingService(fqdn, namespace) {
 				path := fmt.Sprintf("spec/rules[%d]/to[%d]/operation/hosts[%d]", ruleIdx, toIdx, hostIdx)
 				validation := models.Build("authorizationpolicy.nodest.matchingregistry", path)
 				valid = false
@@ -80,19 +83,11 @@ func (n NoHostChecker) validateHost(ruleIdx int, to []*api_security_v1beta.Rule_
 
 func (n NoHostChecker) hasMatchingService(host kubernetes.Host, itemNamespace string) bool {
 	// Covering 'servicename.namespace' host format scenario
-	localSvc, localNs := kubernetes.ParseTwoPartHost(host)
+	_, localNs := kubernetes.ParseTwoPartHost(host)
 
 	// Check wildcard hosts - needs to match "*" and "*.suffix" also..
 	if host.IsWildcard() && localNs == itemNamespace {
 		return true
-	}
-
-	// Only find matches for workloads and services in the same namespace
-	if localNs == itemNamespace {
-		// Check ServiceNames
-		if matches := n.ServiceList.HasMatchingServices(localSvc); matches {
-			return matches
-		}
 	}
 
 	// Check ServiceEntries
--- a/business/checkers/authorization_policies_checker.go
+++ b/business/checkers/authorization_policies_checker.go
@@ -58,7 +58,7 @@ func (a AuthorizationPolicyChecker) runChecks(authPolicy security_v1beta.Authori
 		common.SelectorNoWorkloadFoundChecker(AuthorizationPolicyCheckerType, matchLabels, a.WorkloadList),
 		authorization.NamespaceMethodChecker{AuthorizationPolicy: authPolicy, Namespaces: a.Namespaces.GetNames()},
 		authorization.NoHostChecker{AuthorizationPolicy: authPolicy, Namespace: a.Namespace, Namespaces: a.Namespaces,
-			ServiceEntries: serviceHosts, ServiceList: a.ServiceList, VirtualServices: a.VirtualServices, RegistryServices: a.RegistryServices},
+			ServiceEntries: serviceHosts, VirtualServices: a.VirtualServices, RegistryServices: a.RegistryServices},
 	}
 
 	for _, checker := range enabledCheckers {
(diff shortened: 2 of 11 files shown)


lucasponce:
Hi @hhovsepy,
Perhaps I'm missing some design, but wouldn't it make more sense to put together the "services" and the "registry services"?
image

In the "business/istio_config.go" we have that design, the "GetIstioConfigList" will always return the configuration, depending on the flag, if the Istio configuration is retrieved from the registry or the informers is an implementation detail.

I wonder if we should use that in the "business/services.go#GetServiceList" then the validations code wouldn't need to decide when to use one or another.

I haven't yet reviewed the work, but that was some first thought that came to my mind; if I miss something, let me know, I may not see all the pictures here.

hhovsepy:
> Hi @hhovsepy, Perhaps I'm missing some design, but wouldn't it make more sense to put together the "services" and the "registry services"? image
> 
> In the "business/istio_config.go" we have that design, the "GetIstioConfigList" will always return the configuration, depending on the flag, if the Istio configuration is retrieved from the registry or the informers is an implementation detail.
> 
> I wonder if we should use that in the "business/services.go#GetServiceList" then the validations code wouldn't need to decide when to use one or another.
> 
> I haven't yet reviewed the work, but that was some first thought that came to my mind; if I miss something, let me know, I may not see all the pictures here.

The final idea is to remove "ServiceList/fetchServices" at all in IstioValidations and leave only "RegistryServices/fetchRegistryServices" for all validations. So no need to modify the "business/services.go#GetServiceList" to accept "AllNamespaces" then convert list of RegistryServices to ServiceOverviews for returning only for validation purposes.
As in all validation checkers the "kubernetes.HasMatchingRegistryService" is called and the "ServiceList.HasMatchingServices" will be removed from everywhere.
I hope I understood the concept of RegistryServices correctly.

lucasponce:
> The final idea is to remove "ServiceList/fetchServices" at all in IstioValidations and leave only "RegistryServices/fetchRegistryServices" for all validations. So no need to modify the "business/services.go#GetServiceList" to accept "AllNamespaces" then convert the list of RegistryServices to ServiceOverviews for returning only for validation purposes.
As in all validation checkers, the "kubernetes.HasMatchingRegistryService" is called, and the "ServiceList.HasMatchingServices" will be removed from everywhere.
I hope I understood the concept of RegistryServices correctly.

Yes, we are on the same page here; my comment was more if, from a validations perspective, the responsible for fetching all the services should be the GetServiceList rather than the Registry itself, like it happens in the IstioConfigList code, but is more a refactoring criterion that we can discuss later.

From the feature perspective, to use the registry services is ok, let me to validate the PR from that perspective.

business/checkers/authorization/no_host_checker.go
```diff
@@ -49,7 +49,10 @@ func (n NoHostChecker) validateHost(ruleIdx int, to []*api_security_v1beta.Rule_
 	if len(to) == 0 {
 		return nil, true
 	}
-
+	namespace, clusterName := n.AuthorizationPolicy.Namespace, n.AuthorizationPolicy.ClusterName
```
lucasponce, line 52: I was not aware of the "ClusterName" field in the ObjectMeta; I think it's there for future use according to the documentation?

image

I see in the "kubernetes.GetHost" that it is also expecting an empty value and taking the default domain as well:
image

Perhaps it can keep passing an empty? (I'm not entirely familiar with the logic, but maybe it's not needed in another place).

business/checkers/authorization/no_host_checker.go
```diff
@@ -16,7 +17,6 @@ type NoHostChecker struct {
 	Namespace           string
 	Namespaces          models.Namespaces
 	ServiceEntries      map[string][]string
```
lucasponce, line 19: The RegistryServices should include the "ServiceEntries" as well, then perhaps the use of the "ServiceEntries" may be reconsidered ? I mean the host will be also listed as an "External" Service in the RegistryServices.

business/checkers/authorization/no_host_checker.go
```diff
@@ -16,7 +17,6 @@ type NoHostChecker struct {
 	Namespace           string
 	Namespaces          models.Namespaces
```
lucasponce, line 18: Another thought: if we assume that the list of services will contain all populations from all namespaces, do we need this list of namespaces?

business/checkers/authorization_policies_checker.go
```diff
@@ -59,7 +59,7 @@ func (a AuthorizationPolicyChecker) runChecks(authPolicy security_v1beta.Authori
 		common.SelectorNoWorkloadFoundChecker(AuthorizationPolicyCheckerType, matchLabels, a.WorkloadList),
 		authorization.NamespaceMethodChecker{AuthorizationPolicy: authPolicy, Namespaces: a.Namespaces.GetNames()},
 		authorization.NoHostChecker{AuthorizationPolicy: authPolicy, Namespace: a.Namespace, Namespaces: a.Namespaces,
-			ServiceEntries: serviceHosts, ServiceList: a.ServiceList, VirtualServices: a.VirtualServices, RegistryServices: a.RegistryServices},
+			ServiceEntries: serviceHosts, VirtualServices: a.VirtualServices, RegistryServices: a.RegistryServices},
```
lucasponce, line 62: Mentioned before, the "ServiceEntries" information should be included using the "RegistryServices" data.
image

I think it's worth simplifying the logic, then the checkers may contain fewer data to iterate, and overall it could be a good move.

business/checkers/authorization/no_host_checker.go
hhovsepy: The ClusterName is used in several places of validations, when building FQDN for the service host.
But there is code inconsistency between validations, some places they can use only the default(empty) value and generate FQDN wrongly, so I am changing it to use configured cluster name when the default value is empty.

business/checkers/authorization/no_host_checker.go
hhovsepy: I was thinking to leave this for another PR, as it will extend the scope of this PR.

business/checkers/authorization_policies_checker.go
hhovsepy: Agree, added a comment on https://github.com/kiali/kiali/issues/4528 to have this done in next iteration.

business/checkers/authorization/no_host_checker.go
```diff
@@ -80,21 +83,13 @@ func (n NoHostChecker) validateHost(ruleIdx int, to []*api_security_v1beta.Rule_
...
-	// Only find matches for workloads and services in the same namespace
-	if localNs == itemNamespace {
-		// Check ServiceNames
-		if matches := n.ServiceList.HasMatchingServices(localSvc); matches {
-			return matches
-		}
-	}
-
 	// Check ServiceEntries
 	if kubernetes.HasMatchingServiceEntries(host.String(), n.ServiceEntries) {
```
lucasponce, line 94: Already commented, that ServiceEntries would be included in the ServiceList fetched from the Istio Registry.
So, it doesn't sound necessary, but it can be improved in the next iteration.

business/istio_validations.go
```diff
@@ -174,7 +174,7 @@ func (in *IstioValidationsService) GetIstioObjectValidations(namespace string, o
 	case kubernetes.AuthorizationPolicies:
 		authPoliciesChecker := checkers.AuthorizationPolicyChecker{AuthorizationPolicies: rbacDetails.AuthorizationPolicies,
 			Namespace: namespace, Namespaces: namespaces, ServiceList: services, ExportedServiceEntries: exportedResources.ServiceEntries,
```
lucasponce, line 176: This "ServiceList" is needed here in the context of this work?

lucasponce: In the business/istio_validations.go I see other references where both ServiceList and RegistryServices are passed.
I have the same comment on those.

business/checkers/authorization_policies_checker.go
hhovsepy: @lucasponce back to this topic, actually ServiceEntries are listed in a exported Istio configs retrieved from Registry.
do we need to move ServiceEntries into RegistryServices? because RegistryServices is an array of kubernetes.RegistryService which contains IstioService struct, and it ServiceEntry will be odd within this struct, wdyt?

in my mind was to unify ServiceEntries into one array listed from Registry and pass it to all validations.

lucasponce:
@hhovsepy the PR LGTM, but I'd like to perform another pass to some validation tests.
Would you mind rebasing with the master branch?
The UI has changed, and without the rebase, I can't run an end-to-end instance.

Thanks

lucasponce (approved):
The code looks good to me.
Smoke tests were performed in a local environment.
Moved to QE.

lucasponce:
@hhovsepy after the rebase, I see a new issue/regression.
In the Istio Config list, I see a green icon in the "control-gateway," but on the details page, there is a warning.
image
image

The warning is correct/expected, so I guess regression may happen in the list.

I don't remember seeing this before, so perhaps it was introduced in the latest commits?

hhovsepy:
Thank you @lucasponce , resolved the problem now

prachiyadav (approved):

lucasponce (approved):
Thanks.
I've checked the last issue and now it's gone.

prachiyadav:
Manual Test case
SWSQE-2170 , SWSQE-1990, SWSQE-745

PR #67, opened 2018-02-26
SWS 155: kubernetes scripts

jmazzitelli:
This gets k8s deployment to work. Also includes hack scripts to get istio and bookinfo installed easily on k8s. I use minishift, but the hack scripts don't require that.

PR #4622, opened 2021-12-21
Six week sprint #67 (end of 2021)

israel-hdez:


jmazzitelli (approved):
I trust you implicitly that this will work :)

PR #4626, opened 2021-12-24
chore: add kiali prefix to match the real url path

Xunzhuo:
chore: add kiali prefix to match the real url path

### DIFF
--- a/swagger.json
+++ b/swagger.json
@@ -15,7 +15,7 @@
     "title": "Kiali",
     "version": "_"
   },
-  "basePath": "/api",
+  "basePath": "/kiali/api",
   "paths": {
     "/": {
       "get": {


jmazzitelli (changes requested):
This is not true for OpenShift environments. The default of `/api` as we have it today is correct for that environment. It is only `/kiali` for the default non-OpenShift Kubernetes environments.

And even then people [can change it.](https://github.com/kiali/kiali-operator/blob/v1.44.0/deploy/kiali/kiali_cr.yaml#L1050-L1052) So I don't see a need to change this.

Xunzhuo:
Hi @jmazzitelli , I think you are right, but here:
``` golang
		// swagger:route GET /healthz kiali healthz
		// ---
		// Endpoint to get the health of Kiali
		//
		//     Produces:
		//     - application/json
		//
		//     Schemes: http, https
		// responses:
		//		500: internalError
		//		200
		{
			"Healthz",
			"GET",
			"/healthz",
			handlers.Healthz,
			false,
		},
```
tells the routes of `/healthz`, if we use the base url, the swagger ui will create `/api/healthz`, which is wrong.
How about I start a PR to change the route: `/healthz` to `/api/healthz`?

jmazzitelli:
> How about I start a PR to change the route: /healthz to /api/healthz

The "healthz" API is at `/<base>` - that is where Kiali server's health endpoint is (i.e. it is not under `/api`).

For OpenShift, that is `/healthz`, for non-OpenShift, that is `/kiali/healthz`.

For example, in my minikube:

```
$ curl -Lv http://localhost:20001/kiali/healthz
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 20001 (#0)
> GET /kiali/healthz HTTP/1.1
> Host: localhost:20001
> User-Agent: curl/7.61.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Vary: Accept-Encoding
< Date: Fri, 24 Dec 2021 09:27:36 GMT
< Content-Length: 0
< 
```

I don't know what is broken here. Is there something actually broken in the server? If the problem is just swagger documentation, then just change that.  There is no need to move the `/healthz` endpoint to a completely different endpoint (e.g. `/api/healthz`)

PR #4627, opened 2021-12-24
fix(istio-details): trim spaces and ignore upcase

Xunzhuo:
fix(istio-details): trim spaces and ignore upcase

### DIFF
--- a/handlers/istio_config.go
+++ b/handlers/istio_config.go
@@ -88,9 +88,9 @@ func IstioConfigList(w http.ResponseWriter, r *http.Request) {
 
 func IstioConfigDetails(w http.ResponseWriter, r *http.Request) {
 	params := mux.Vars(r)
-	namespace := params["namespace"]
-	objectType := params["object_type"]
-	object := params["object"]
+	namespace := strings.TrimSpace(params["namespace"])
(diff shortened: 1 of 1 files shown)


lucasponce (changes requested):
I don't think this one is necessary.
There are multiple handlers that are fetching the namespace parameter, and a "trim" wouldn't help a lot here. A bad namespace would be validated in the business layer.

Also, just fixing these validations in one handler but not in others would be a little bit inconsistent.

@Xunzhuo 
I understand that the motivation of these PRs is to get familiar with the project and also show commitment to collaborate in the code.

We really appreciate that. If you are open to collaborating, we could find a feature from the backlog where you could both get familiar but also provide a more end-to-end flow instead to send a list of minor PRs, as I suspect some of them can get merged, but others like these one could not be merged.

Let us know, I don't like to reject a PR and create the impression that we don't appreciate the effort, but I guess we could find some tasks from the backlog that can be assigned to you and have a more long-term work.

Xunzhuo:
@lucasponce , sure, if there is any backend features, please assign it to me:)

Xunzhuo:
I will close this PR :)

PR #4629, opened 2021-12-24
fix(boxBy-appenders):  update and set default to empty

Xunzhuo:
`boxBy` and `appenders` default values are not `none` and `run all appenders`

set default to empty in `boxBy` and `appenders`

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -200,17 +200,17 @@ type AppendersParam struct {
 	//
 	// in: query
 	// required: false
-	// default: run all appenders
+	// default: ""
 	Name string `json:"appenders"`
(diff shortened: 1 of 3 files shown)


lucasponce:
If the "swagger.json" is updated, is not updated the "kiali_api.md"?

lucasponce:
@Xunzhuo this PR also needs to update the "kiali_api.md" resource.
You can do that with the commands:

```
make swagger-gen
make swagger-validate
```

Then we could move forward with this PR.

lucasponce (approved):
Trivial

PR #4630, opened 2021-12-24
fix(app/versionApp): graphType is required and types are different

Xunzhuo:
In app and versionApp graph
+ graphType is required 
+ workload and service are not supported types 

separate graphApp graphAppVersion from graphParam and create a new param for app.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -116,6 +116,15 @@ type ObjectTypeParam struct {
 	Name string `json:"object_type"`
 }
 
+// swagger:parameters istioConfigList istioConfigDetails serviceDetails serviceUpdate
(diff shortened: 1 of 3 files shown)


lucasponce:
I have the same comment: when "swagger.json" is modified, I guess that the "kiali_api.md" should be updated to keep it in sync.
It probably is missing the "make swagger-validate" command IIRC.

lucasponce:
@Xunzhuo 
This PR would need to update the "kiali_api.md" resource using the following commands:

```
make swagger-gen
make swagger-validate
```
And then we can move forward with the PR.

lucasponce (approved):

PR #4631, opened 2021-12-24
fix(metrics): add restrictions to queries

Xunzhuo:
add some restrictions to queries

(a) what are broken and missing:
1. the default value for `requestProtocol` is not `all protocols`
2. when getting query of `requestProtocol`, we did not restrict the values, just take any values which are not ""
3. when getting query of `rateInterval` , we did not check it formats, just take any values which are not "", which should be formated like `10s`, `5m`, `3h`, `1h20s`...
4. when getting query of `queryTime ` `duration` and `step `, we do not check it if it is < 0, just take values which are not "".
5. Add test cases to test the changes I make.

### DIFF
--- a/doc.go
+++ b/doc.go
@@ -433,7 +433,7 @@ type RequestProtocolParam struct {
 	//
 	// in: query
 	// required: false
-	// default: all protocols
+	// default: http
 	Name string `json:"requestProtocol"`
 }
 
--- a/handlers/metrics.go
+++ b/handlers/metrics.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strconv"
 	"time"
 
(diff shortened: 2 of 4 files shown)


jmazzitelli:
In all of these PRs you are creating, can you add more/better descriptions explaining (a) what is broken or missing and (b) what the new code is doing to fix it?

Right now, some of these just have code without any comments in the code, so it is hard to judge what you are doing in all of these code submissions.

Xunzhuo:
@jmazzitelli OK, I will update the descriptions.

jshaughn (changes requested):
It's weird, but for this PR github is not allowing me to make inline comments.  So, unfortunately, I will have to do it all in one comment.

> the default value for requestProtocol is not all protocols

Are you sure this is true?  This is basically a filter, so when not supplied there is no filtering on protocol.   So, for example, if the metrics are coming from istio_requests_total, it could include HTTP and gRPC.

> when getting query of rateInterval , we did not check it formats, just take any values which are not "", which should be formated like 10s, 5m, 3h, 1h20s...

Are you sure Prometheus doesn't already validate the parameter and return an error if necessary?

> when getting query of queryTime duration and step , we do not check it if it is < 0, just take values which are not "".

Not sure this check is necessary but, if you leave it in, then you need to update the message below because it may no longer just be a parsing error. (same comment for the second occurrence)

Although your contributions are appreciated, In general I'd discourage code changes in mature projects, where there hasn't been a reported issue.  Any code change can have unexpected consequences.  I would encourage you to work on reported issues from the kiali issue backlog.

Xunzhuo:
Sure it is OK for this is a PR before I started a discussion about assigning issues. I could close this. Thanks @jshaughn For now, all PRs before assigning issues have been merged or closed. I will work on issues to submit feature PRs.

PR #4632, opened 2021-12-26
fix(healthz): add '/api' prefix to '/healthz' route

Xunzhuo:
**Describe the change** 

`healthz` api is designed to test if kiali is healthy, I strongly suggest adding '/api' prefix to '/healthz' route.

Reasons are below:

1. In swagger doc, there is a `/api` base url, generated routes for `/healthz` actually is `/api/health`, while it is actually `/healthz`. It will be a always confilct, because we can not remove `/api` baseURL.

2. And according to api doc: https://github.com/kiali/kiali/blob/master/kiali_api.md#healthz, this route is actually designed as `/api/healthz` not `/healthz`.

3. And **I check that in kiali-ui, there is no place using `/healthz` api, so this update is compatible.** :), but need some changes in helm-chart and operator, where I will give PRs to update too.

### DIFF
--- a/kiali_api.md
+++ b/kiali_api.md
@@ -9306,7 +9306,7 @@ hosts: ["another-host.com"]
 ```
 
 You can fine tune the authorization policy to set different requirement per path. For example,
-to require JWT on all paths, except /healthz, the same `RequestAuthentication` can be used, but the
+to require JWT on all paths, except /api/healthz, the same `RequestAuthentication` can be used, but the
 authorization policy could be:
 
 ```yaml
@@ -9325,7 +9325,7 @@ source:
 requestPrincipals: ["*"]
 to:
 operation:
-paths: ["/healthz"]
+paths: ["/api/healthz"]
 ```
 
 <!-- crd generation tags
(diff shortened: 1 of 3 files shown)


jmazzitelli:
I'm -1 - we do not need this churn. `/healthz` is working perfectly fine.

>  I check that in kiali-ui, there is no place using /healthz api, so this update is compatible.

It is not used by the UI - it is used for our Kubernetes [Deployment's readiness and liveness probes.](https://github.com/kiali/helm-charts/blob/v1.44.0/kiali-server/templates/deployment.yaml#L75-L88)

This is introducing backwards-incompatible changes for no reason whatsoever other than just to have it match swagger.

I'm closing this. Unless we have a legitimate reason for introducing this change, we should not do it. Users of Kiali might be expecting this endpoint for other reasons (e.g. their own scripts for determining readiness) and changing its location is not advisable without a good reason since it will break those people using it.

If the problem is with swagger, change the swagger docs, not the actual feature/functionality (i.e the docs should change to match the functionality, not the reverse).

PR #4634, opened 2021-12-26
fix: update healthz apis

Xunzhuo:
`/api/healthz` is not the right route

update healthz APIs docs: from `/api/healthz` to `/healthz`

### DIFF
--- a/kiali_api.md
+++ b/kiali_api.md
@@ -127,7 +127,7 @@ _
 | Method  | URI     | Name   | Summary |
 |---------|---------|--------|---------|
 | GET | /api/config | [get config](#get-config) |  |
-| GET | /api/healthz | [healthz](#healthz) |  |
+| GET | /healthz | [healthz](#healthz) |  |
 | GET | /api | [root](#root) |  |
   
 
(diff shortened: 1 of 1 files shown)


Xunzhuo:
cc @jmazzitelli , I start a new PR to update healthz api. But I do not know if there is a way to fix it in swagger for there is a base url `/api`.

jmazzitelli (approved):

lucasponce:
This PR may have a problem, as it would be re-created every time that swagger is generated.
So I guess it could be better to address this on the swagger side.
One way could be to not use the "basePath" in the "swagger.json" and let the handlers to take the "/api" there as there will be endpoints that won't use that prefix.

lucasponce (changes requested):
I don't think it makes sense to modify the .md as these changes will be recreated.
I suggest investigating to add these changes in the swagger source ("swagger.json" or "doc.go" resources, or any other comment in the "routes.go")

lucasponce:
Closing this one, the "kiali_api.md" resource is auto-generated from the "swagger.json":
image

So, any change should be provided from the .json definition.

Issue #4636, opened 2022-01-05
Jaeger http legacy protocol has problems in master


lucasponce:
When using the legacy HTTP config in jaeger:
![image](https://user-images.githubusercontent.com/1662329/148208350-4fba6d36-0355-4716-8d46-d3a9db95ebb7.png)

 It seems there are some problems during the unmarshal operation.

I'm not sure if this is a bug or just an environmental problem on my IDE.
I open this issue to investigate and fix if something is broken.

PR #4637, opened 2022-01-05
Fix wrong pointer in jaeger http client

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4636

When using the HTTP configuration to access jaeger from Kiali:
```
  tracing:
    enabled: true
    use_grpc: false
    in_cluster_url: http://localhost:16686/jaeger
    url: http://localhost:16686/jaeger
    namespace: istio-system
    port: 443
    service: jaeger-query
```

There was a bug when access to a trace detail, the base URL was overwritten, which produced that following queries would fail. The problem is only visible in the HTTP mode, not GRPC.

### DIFF
--- a/jaeger/http_client.go
+++ b/jaeger/http_client.go
@@ -29,7 +29,7 @@ func getAppTracesHTTP(client http.Client, baseURL *url.URL, namespace, app strin
 }
 
 func getTraceDetailHTTP(client http.Client, endpoint *url.URL, traceID string) (*JaegerSingleTrace, error) {
-	u := endpoint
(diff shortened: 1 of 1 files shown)


leandroberetta (approved):
Verified on my dev environment.

skondkar (approved):
Verified LGTM

Screenshot-Jaeger-config

After setting 'use_grpc: false' and using jaeger http url, verified that tracing works in UI. Also verified that the e2e tests run is successful.

skondkar:
Polarion test case ->> SWSQE-2392

Issue #4641, opened 2022-01-10
Wrong KIA1106 "More than one Virtual Service for same host"


lucasponce:
In a demo environment:
![image](https://user-images.githubusercontent.com/1662329/148741206-6eb16a38-f6c2-40c2-baca-d5bd08e7c9d5.png)

I got a wrong message with 2 VS:
![image](https://user-images.githubusercontent.com/1662329/148741595-adb58821-f782-4670-bd80-3bb40f474a01.png)
![image](https://user-images.githubusercontent.com/1662329/148741643-ecd71b8d-413f-46e2-9db5-1e8fc1ad0a08.png)

It's possible that this is a side effect of the WIP related to better using the Istio Registry for host detection, but I have the impression that there is a regression in the code.

hhovsepy:
This is general regression issue for all validations, happens after https://github.com/kiali/kiali/pull/4535 merged.
The issue is: that exported Istio Configs are retrieved from Registry and they contains own namespace Configs as well, which is causing duplication, thus the validation issues happens for all configs.
Solution is to have single list fo Configs for validations.
Needs to be resolved ASAP as a blocker issue.

PR #4644, opened 2022-01-10
Removed duplications for VS, DR and SE in Istio Validations.

hhovsepy:
As now the exported resources contains all objects, skip using Istio Objects from IstioConfigList for local namespace.
Exported resources fortains Istio Objects from IstioConfigList for AllNamespaces, so using any other list will cause duplicates.

Fix for issue https://github.com/kiali/kiali/issues/4641

For QE:
Full manual regression testing is required for VirtualServices, DestinationRules and ServiceEntries validations from https://kiali.io/docs/features/validations/
Full UI/E2E automation regression testing is required.

### DIFF
--- a/business/checkers/authorization_policies_checker.go
+++ b/business/checkers/authorization_policies_checker.go
@@ -16,7 +16,6 @@ type AuthorizationPolicyChecker struct {
 	AuthorizationPolicies   []security_v1beta.AuthorizationPolicy
 	Namespace               string
 	Namespaces              models.Namespaces
-	ServiceEntries          []networking_v1alpha3.ServiceEntry
 	ExportedServiceEntries  []networking_v1alpha3.ServiceEntry
 	ServiceList             models.ServiceList
 	WorkloadList            models.WorkloadList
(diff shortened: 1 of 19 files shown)


lucasponce:
I've found a problem in this PR with the cross-validation of the gateways.
Using the travels demo app:
- Expose the "control" service using a gateway, as described in the tutorial:
image
image
- In the "travels-agency" namespace, create a simple traffic shifting under the "travels" service.
- Tweak the VS to use a gateway created in the "control" service to force a validation.
image

I understand that the validation shouldn't fail if that gateway exists.

hhovsepy:
@lucasponce thank you for the catch, fixed now.

lucasponce:
I think there is a KIA1106 validation that is still not fully working between namespaces.
For example, I have these two VS:
image
image

And I'd expect a KIA1106 there as they have the same gateway and a wildcard.

If I create another VS in the same "travel-agency" namespace, then the validation is happening:
image
image

I don't see it as a blocker, perhaps we can consolidate all these pending work and address it in a separate effort.

I think I can focus to review that all pending work is stable and no blocker/major regression is present for basic scenarios.

hhovsepy:
> I don't see it as a blocker, perhaps we can consolidate all these pending work and address it in a separate effort.
> 
> I think I can focus to review that all pending work is stable and no blocker/major regression is present for basic scenarios.

I can see this issue as well.
And it is coming from "single_host_checker.go/storeHost" which is not designed for this scenarios.
It was working before because of coincidence, as before this PR the Istio Configs were duplicated for validations.
So for this fix I would suggest to have it in a scope of other issue and PR.

lucasponce (approved):
LGTM
Tested locally some basic tests for regressions, things look good.

pbajjuri20:
**Before PR:**

after deploying the bookinfo on Kiali 1.45, it shows the warning 

error_31



**After PR:**

The warning has been gone, Istio-config has no issues.

after_4644

pbajjuri20:
E2E test cases run successfully against this PR.

skondkar (approved):
LGTM
Verified all validations for VirtualServices, DestinationRules and ServiceEntries from https://kiali.io/docs/features/validations/
Also verified that no related errors in UI automation run

PR #4655, opened 2022-01-12
Enforce specific lengths for signing key

israel-hdez:
This is to support sessions with AES-GCM encryption.

Related to #4542

### DIFF
--- a/config/token.go
+++ b/config/token.go
@@ -47,10 +47,10 @@ func GetSigningKey() string {
 }
 
 func ValidateSigningKey(signingKey string, authStrategy string) error {
-	if authStrategy != AuthStrategyAnonymous && (len(signingKey) == 0 || signingKey == "kiali") {
-		// "kiali" is a well-known signing key reported in a CVE. We ban it's usage.
-		// An empty key is also just not allowed.
-		return errors.New("signing key for login tokens is invalid")
+	if authStrategy != AuthStrategyAnonymous {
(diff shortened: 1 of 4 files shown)


config/token.go
```diff
@@ -47,10 +47,10 @@ func GetSigningKey() string {
 }
 
 func ValidateSigningKey(signingKey string, authStrategy string) error {
-	if authStrategy != AuthStrategyAnonymous && (len(signingKey) == 0 || signingKey == "kiali") {
-		// "kiali" is a well-known signing key reported in a CVE. We ban it's usage.
-		// An empty key is also just not allowed.
-		return errors.New("signing key for login tokens is invalid")
+	if authStrategy != AuthStrategyAnonymous {
+		if len(signingKey) != 16 && len(signingKey) != 24 && len(signingKey) != 32 {
```
nrfox, line 51: What's the reason for these specific lengths i.e. 16, 24, and 32?

israel-hdez: The AES-GCM cipher does not like other lengths. It's tied to the key length of the algorithm.

nrfox (approved):

israel-hdez:
I don't think this needs QE. I'll merge.

jmazzitelli:
NOTE! When we release this and we bump the istio addon to use this version of kiali, we first must ensure we remember to change this: https://github.com/istio/istio/blob/master/manifests/addons/values-kiali.yaml#L12

Issue #4585, opened 2021-12-14
Develop a mock backend server for local UI work in scalability scenarios


lucasponce:
During the scalability discussion, https://github.com/kiali/kiali/discussions/4518 or https://github.com/kiali/kiali/discussions/4080 one of the challenges identified is how to test a large deployment. 

Developers use small clusters for feature works and there is the need to improve the mocks artifacts to have tools that allow focusing on large deployments without the need to get access to such large infrastructure.

One of the areas identified is the graph usability, where there are some areas in the UI to improve, but as a preliminary step, we'd need to have a better local mock server.

This issue aims to provide an optional mock server that may work as UI proxy:

![image](https://user-images.githubusercontent.com/1662329/145963574-8100713d-b368-4f1c-88b7-bdb5a8dc00be.png)

And would let the developer easily mock the backend calls:

![image](https://user-images.githubusercontent.com/1662329/145963674-3c96463e-8785-4fdf-8b39-ac24b5174a76.png)

As a first thought, the backend calls can be grouped in folders, and as a developer, I can populate those folders as a .json from a real backend, which I can modify later to use in a pure UI standalone work.

The goal here would be to have a mock-up for very large scenarios (hundreds of namespaces) without the need of having a real backend process, in that way, the productivity UI work could be greatly improved.

I wouldn't like to reinvent any existing framework, but on the other hand, we should be able to easily mock any endpoint in a very straightforward way (i.e. I've just copied a JSON to a folder and the UI can use those data like a real endpoint).

Tasks:
- [x] Tool to generate graph json data for different large scale scenarios: https://github.com/kiali/kiali/pull/4658
- [ ] #4665

Suggested test plan (QE @mattmahoneyrh):
- The goal of this Epic is to provide a Developer tool for working "offline" in the UI using large topologies. There is no expectation to have a "formal" QE test for this, but the discussion and the effort may be interesting and it may have synergies with other QE efforts as well.

nrfox:
Thanks for opening this. I briefly looked into using kiali's openapi spec to create a mock server implementation. An advantage of using the openapi spec is that it will get updated automatically as the api changes. 

There seemed to be some circular dependencies in the openapi spec perhaps caused by the generation tool we're using. [AuthorizationPolicy](https://github.com/kiali/kiali/blob/master/swagger.json#L4562) references itself which I guess is valid until you want to generate clients/servers from the spec.

lucasponce:
[...]

PR #4658, opened 2022-01-14
Add graph generator for creating mock graph data

nrfox:
The generator creates a very basic graph but has a few configurable knobs. This was primarily tested with [this patch](https://github.com/nrfox/kiali/compare/kiali-4585...nrfox:patch-graph-generator?expand=1) but the eventual goal is to use something like cypress from https://github.com/kiali/kiali-ui/pull/2264 when it's ready to mock out the request and return the json data generated by the generator.

Also adds a simple proxy server to serve the generated graph data from memory. See [here](https://github.com/kiali/kiali/blob/d5ee726d0859886cd978cd74a9d3a3e37344f87d/tools/README.md) for instructions on running/testing both.

Sample output:
Screenshot from 2022-01-14 13-06-03

Relates to #4585

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -1,0 +1,41 @@
+# Tools for Kiali maintainers
+
+This folder contains tools for maintainers of the Kiali project.
+
+## Mock graph data
+
+There's two cli tools to help with testing large graph topologies without needing to deploy the underlying resources.
+
+**Note:** You do not need to run the generator before running the proxy. The proxy server will run the generator itself and serve up the mock data from memory by default. If you are testing locally, you probably want to use the proxy. If you want to output the mock data as a json file, then run the generator.
+
+### proxy server
+
+The proxy server runs an http proxy that intercepts all `/api/namespaces/graph` calls and returns mock graph data. The rest of the api calls are passed through to the kiali api. To test the kiali ui using the proxy server, you should start the proxy server then, in the kiali ui's `package.json`, replace the `proxy` field with the address of the proxy server e.g. `"proxy": "https://localhost:10201"`. Finally run `yarn start` and navigate to local node server endpoint: `http://localhost:3000`. When you navigate to the graph page, you should see graph rendered with the mock data.
+
+Running the following command will start the proxy server in https mode.
+
+```bash
+go run tools/cmd/proxy/main.go <kiali-url> --apps 50 --box --https
+```
+
+For more usage information:
+
+```bash
+go run tools/cmd/proxy/main.go --help
+```
+
+### generator
+
+The generator creates a json file with a mock `/api/namespaces/graph` response.
+
+Running the following command will create a json file with 50 apps (workloads + services). There is some randomness in the graph generation so running the command multiple times with the same options will yield different results.
+
+```bash
+go run tools/cmd/generate/main.go --apps 50 --box
+```
+
+For more usage information:
+
+```bash
+go run tools/cmd/generate/main.go --help
+```
--- a/go.mod
+++ b/go.mod
@@ -46,6 +46,7 @@ require (
 	github.com/google/gofuzz v1.1.0 // indirect
 	github.com/google/uuid v1.1.2 // indirect
 	github.com/googleapis/gnostic v0.5.5 // indirect
+	github.com/imdario/mergo v0.3.5 // indirect
 	github.com/json-iterator/go v1.1.11 // indirect
 	github.com/matttproud/golang_protobuf_extensions v1.0.1 // indirect
 	github.com/moby/spdystream v0.2.0 // indirect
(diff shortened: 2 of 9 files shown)


lucasponce:
Minor comment: it would be nice if, by default, the output can be added to the _output folder or at least to put the graph_*.json in the ".gitignore" to skip them from accidental commit.

nrfox:
> Minor comment: it would be nice if, by default, the output can be added to the _output folder or at least to put the graph_*.json in the ".gitignore" to skip them from accidental commit.

Sure that was something that I wasn't quite sure how others would want to consume that data or what preferences for the dev flow would be. Since the primary use case for the mock data is to test the front end but the generator lives in the `kiali` repo, I was imaging that the data would be checked into the `kiali` repo. Perhaps instead people would prefer to generate it themselves by running the generator instead of having it checked into the `kiali` repo. I'm happy to hear others thoughts on that.

nrfox:
@jshaughn @lucasponce  added you as a reviewer in case you wanted to experiment with the mock graph data with some of the layout work y'all are doing. See [here](https://github.com/kiali/kiali/blob/d5ee726d0859886cd978cd74a9d3a3e37344f87d/tools/README.md) for instructions on how to run the proxy server.

tools/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Mock graph data
+
+There's two cli tools to help with testing large graph topologies without needing to deploy the underlying resources.
+
+### generator
+
+The generator creates a json file with a mock `/api/namespaces/graph` response.
+
+Running the following command will create a json file with 50 apps (workloads + services). There is some randomness in the graph generation so running the command multiple times with the same options will yield different results.
```
jshaughn, line 13: I did this only to read below that I didn't need to do it :-)

jshaughn (changes requested):
Nick, got this working and it seems to do the trick, pretty cool.  Couple of comments.

tools/README.md
```diff
@@ -0,0 +1,39 @@
...
+```
+
+### proxy server
+
+The proxy server runs an http proxy that intercepts all `/api/namespaces/graph` calls and returns mock graph data. The rest of the api calls are passed through to the kiali api. To test the kiali ui using the proxy server, you should start the proxy server then, in the kiali ui's `package.json`, replace the `proxy` field with the address of the proxy server e.g. `"proxy": "https://localhost:10201"`. You do not need to run the generator before running the proxy. The proxy server will run the generator itself from memory by default.
+
+Running the following command will start the proxy server in https mode.
+
+```bash
+go run tools/cmd/proxy/main.go -apps 50 -box -https
```
jshaughn, line 32: Somehow my review comment got lost, I'll type it again...

This doesn't work because `-kiali-url` is required.

nit: I expected the options to use `--` as opposed to `-`.  Maybe it's just me but usually abbreviations use `-`.

The `-box` option needs to be tweaked to work like the normal ui.  Namespace boxing is only done when there are multiple namespaces, and also only when the box has multiple elements.  Here is an example of an unecessary box:

image

But overall it's pretty cool.  Just using the defaults from the doc:

image

By the way, dagre is terrible when you have an ingress gateway like this, with a lot of different destinations, because it stacks all of the ingress destinations.  Concentric looks nice because there is zero traffic between namespaces.

lucasponce:
Perhaps the tools/README.md is missing more steps, like if I need to pre-populate namespaces?

tools/README.md
nrfox: Whoops yeah I'll include that information further up

nrfox:
> Perhaps the tools/README.md is missing more steps, like if I need to pre-populate namespaces?

I'll add that to this to the README but you shouldn't have to pre-populate the namespaces. Both the generator and proxy will pre-populate them for you using your current kube context. If it can't create the namespaces for any reason, it'll log a warning but should still startup the proxy or create the graph json.

tools/README.md
nrfox: > The -box option needs to be tweaked to work like the normal ui. Namespace boxing is only done when there are multiple namespaces, and also only when the box has multiple elements. Here is an example of an unecessary box:

I wonder if I could've had the input to `Generate` be a `trafficMap` instead of a `cytoscape.Config` and been able to re-use the appenders...

jshaughn: Boxing actually does not happen in the appenders.  It happens in Cytoscape.go, which is the code that converts the TrafficMap into the cy config.   The idea being that we could take the generic TrafficMap structure and potentially generate the config for various vendors, it's just the case that we have only one vendor.  The config generators are fairly autonomous, at that point the TrafficMap is complete, there are no calls to kbs, prom, etc.

lucasponce (approved):
LGTM
I have tested the proxy with my IDE:
image

image

And the first demos work as expected, thanks @nrfox, I can see that we can add more tweaks and checks into the proxy to mock more scenarios and other endpoints (i.e. health and others).

I'm +1 to merge this and iterating from here unless you wanted to add more features for this PR.

jshaughn (approved):
My previous issues are resolved and this looks usable to me!

PR #4667, opened 2022-01-21
Sidecar EgressChecker now supports Services from Registry.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4528

Modified the EgressHostChecker to use Services from Registry instead of only local namespace ones.
So now it supports exported Services as well.

The affected validation message is:
"KIA1004 This host has no matching entry in the service registry"

For QE:
Test the https://kiali.io/docs/features/validations/#kia1004---this-host-has-no-matching-entry-in-the-service-registry
Sidecar in "bookinfo", but using host from different namespace, for instance "bookinfo2/reviews.bookinfo2.svc.cluster.local".
Then "reviews/bookinfo2" service export to "bookinfo3" (Service annotation "networking.istio.io/exportTo") and verify that Sidecar host shows warning.

### DIFF
--- a/business/checkers/sidecars/egress_listener_checker.go
+++ b/business/checkers/sidecars/egress_listener_checker.go
@@ -13,7 +13,6 @@ import (
 type EgressHostChecker struct {
 	Sidecar          networking_v1alpha3.Sidecar
 	ServiceEntries   map[string][]string
-	ServiceList      models.ServiceList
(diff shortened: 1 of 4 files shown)


lucasponce (approved):

prachiyadav:
Manual Test Case : SWSQE-2537

PR #4669, opened 2022-01-21
Refactor OIDC auth to stop using the dgrijalva/jwt library

israel-hdez:
* Moves the openid authentication code to it's own module.
* Removes usage of the dgrijalva/jwt library on the code handling openid authentication.
* Add a bunch of unit tests to the openid auth.

Related to #4542


israel-hdez:
I think this is ready for review.

I only need to add doc comments to unit tests. Also, I have the idea that a few additional units are needed, but I'm thinking about it. Anyway, the main code looks ready.

@nrfox Once you review, if you still think all this auth-related code should be moved under `handlers` (as you mentioned [here](https://github.com/kiali/kiali/pull/4596#discussion_r781376667)), let me know, and I will move it on this PR.

business/authentication/auth_controller.go
```diff
@@ -17,6 +19,22 @@ type AuthController interface {
...
+	// third-party authentication systems.
+	//
+	// This function can return a nil handler if the implementing AuthController does not handle
+	// authentication requests through the Kiali's web_root.
+	GetAuthCallbackHandler(fallbackHandler http.Handler) http.Handler
+
+	// PostRoutes adds any additional endpoints needed on the Kiali's router
+	// by the implementing AuthController.
+	PostRoutes(router *mux.Router)
+
```
nrfox, line 37: It might be better to restrict this interface to the set of methods that all implementations are expected to implement rather than add methods that will be noops for some implementations. These seem like they could be standalone functions exposed by the auth package outside of this interface.

business/authentication/openid_auth_controller.go
```diff
@@ -0,0 +1,1444 @@
...
+			Reason:     "Request rejected: State parameter is invalid",
+		}
+	}
+
+	return p
+}
+
+// requestOpenIdToken makes a request to the OpenId server to exchange the received code (of the
+// authorization code flow) with a proper identity token (id_token) and an access_token (if applicable).
+func (p *openidFlowHelper) requestOpenIdToken(redirect_uri string) *openidFlowHelper {
```
nrfox, line 823: Would the [stdlib oauth2 package](https://pkg.go.dev/golang.org/x/oauth2) help in doing some of this flow such as the token exchange?

business/authentication/openid_auth_controller.go
```diff
@@ -0,0 +1,1444 @@
...
+	if err != nil {
+		return http.StatusInternalServerError, "Error instantiating the business layer", err
+	}
+
+	// Using the namespaces API to check if token is valid. In Kubernetes, the version API seems to allow
+	// anonymous access, so it's not feasible to use the version API for token verification.
+	nsList, err := bsLayer.Namespace.GetNamespaces(context.TODO())
+	if err != nil {
+		return http.StatusUnauthorized, "Token is not valid or is expired", err
+	}
```
nrfox, line 1436: Do we need to also check other services i.e. prom, jaeger?

business/authentication/openid_auth_controller.go
israel-hdez: Yes!
I remember I was considering using that lib before the very first version of all this code. To be honest, I don't clearly remember why I didn't use it. What I remember is that the oauth2 package only deals with "OAuth", but not OpenID - although they are very similar how they work.

I also considered using [coreos/go-oidc](https://github.com/coreos/go-oidc), and the reason this one was not used is that it only supports the "authorization code flow" while back then Kiali only supported the "implicit flow". Nowdays, Kiali support both "implicit flow" and "authorization code flow". So, this lib still is still short.

I know all this doesn't really answers your question. I can't remember the reason :-(

business/authentication/openid_auth_controller.go
israel-hdez: I don't think it's needed, because those services does not really support RBAC. Kiali, mostly uses the Kubernetes API to decide what to show/hide. If we were to only use Prometheus, it would happily return any requested data.

business/authentication/auth_controller.go
nrfox: This might seem vague but it's harder for me to understand what an `AuthController` is/does with the added methods whereas before it was pretty clear. Said another way, adding routes to the router does not seem related to the rest of the `AuthController` interface. I'm not sure where the best place to put that logic is. Perhaps it could be an `authentication.PostRoutes(router *mux.Router)` method or something along those lines but you probably have a better feel for that than I do.

business/authentication/openid_auth_controller.go
nrfox: No worries I was just curious to know. Thanks for the info.

handlers/authentication.go
```diff
@@ -545,17 +329,17 @@ func (aHandler AuthenticationHandler) Handle(next http.Handler) http.Handler {
...
+		//case config.AuthStrategyOpenId:
+		//	statusCode, token = checkOpenIdSession(w, r)
+		//	if conf.Auth.OpenId.DisableRBAC {
+		//		// If RBAC is off, it's assumed that the kubernetes cluster will reject the OpenId token.
+		//		// Instead, we use the Kiali token an this has the side effect that all users will share the
+		//		// same privileges.
+		//		token = aHandler.saToken
+		//	}
+		//
+		//	authInfo = &api.AuthInfo{Token: token}
```
nrfox, line 341: Can this be removed?

israel-hdez: Yes. I removed it.
Sorry, it's a leftover.

business/authentication/openid_auth_controller.go
```diff
@@ -0,0 +1,1444 @@
...
+		return p
+	}
+
+	conf := config.Get()
+
+	if len(conf.Auth.OpenId.AllowedDomains) > 0 {
+		if err := checkDomain(p.IdTokenPayload, conf.Auth.OpenId.AllowedDomains); err != nil {
+			p.Error = &AuthenticationFailureError{Reason: err.Error()}
+		}
+	}
```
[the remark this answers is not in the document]
israel-hdez: What I understand is that if there are no AllowedDomains, it means it's "not configured". So, we don't do any validation.

business/authentication/token_auth_controller.go
```diff
@@ -63,18 +64,21 @@ type UserSessionData struct {
 	//
 	// example: zI1NiIsIsR5cCI6IkpXVCJ9.ezJ1c2VybmFtZSI6ImFkbWluIiwiZXhwIjoxNTI5NTIzNjU0fQ.PPZvRGnR6VA4v7FmgSfQcGQr-VD
 	// required: true
-	Token string
+	Token string `json:"-"`
```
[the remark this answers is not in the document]
israel-hdez: Yes. I originally didn't put any `json` annotation with the intention of not being sent to the browser. However, here I saw the token was being sent. So, I'm removing it. The caller function *do* uses the `Token` and puts it in the `context()` so that it can be used later, but it shouldn't be travelling over the network. The UI is currently not using the token...

business/authentication/openid_auth_controller.go
```diff
@@ -0,0 +1,1444 @@
...
+	webRoot := conf.Server.WebRoot
+	webRootWithSlash := webRoot + "/"
+
+	flow := openidFlowHelper{businessInstantiator: c.businessInstantiator}
+	flow.
+		extractOpenIdCallbackParams(r).
+		callbackCleanup(w).
+		checkOpenIdAuthorizationCodeFlowParams().
+		validateOpenIdState().
+		requestOpenIdToken(httputil.GuessKialiURL(r)).
```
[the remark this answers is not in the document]
israel-hdez: Yes!
I don't know what I was thinking when I originally did this :-)

But, certainly, it's better to have a dedicated callback endpoint.

My current thinking it that *perhaps* I was trying to simplify the configuration that the user had to do on the open id server side to not having to figure out what is the callback endpoint, and simply use the Kiali URL. But, I agree this makes the code a little more complex than needed on this part :-(

Currently, I would like to move to a dedicated endpoint for the callback. But I'm not sure if users will like that they need to change configs on their openid server.

business/authentication/openid_auth_controller.go
```diff
@@ -0,0 +1,1444 @@
...
+		case string:
+			if oidCfg.ClientId != ac {
+				return fmt.Errorf("the OpenId token is not targeted for Kiali; got aud = '%s'", audienceClaim)
+			}
+		case []string:
+			audFound := false
+			for _, audItem := range ac {
+				if oidCfg.ClientId == audItem {
+					audFound = true
+				}
```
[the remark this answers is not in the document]
israel-hdez: Yes, it's possible. Although I'm not sure in which cases we would see multiple ids. From the client side, apparently we cannot request the token to be issued for multiple audiences. So, it must be more a server side feature.

From what I read here: https://openid.net/specs/openid-connect-core-1_0.html#IDTokenValidation
looks like we should "trust" the other client_ids if there are more than one.

I'm not sure how we could trust other clients. So, we probably should limit this to contain ourselves :confused:

business/authentication/auth_controller.go
israel-hdez: > Another option is to do a type assertion on the interface

Yes, we can go that route... Sounds better to me.

business/authentication/openid_auth_controller.go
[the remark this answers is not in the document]
nrfox: Hmm I read that the opposite way, that if there are other auds that we don't trust then we should reject:

> The ID Token MUST be rejected if ... it contains additional audiences not trusted by the Client.

> So, we probably should limit this to contain ourselves

Agreed, we should just trust ourselves :)

business/authentication/openid_auth_controller.go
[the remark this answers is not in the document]
nrfox: No worries, I'm mostly just curious for my own context. This doesn't need to change for this PR.

Not familiar with the server side configuration for all this but perhaps just the redirect uri sent with the query params on the 302 resp for `/api/auth/openid_redirect` could be updated and it'd just work without users needing to update anything?

israel-hdez:
@nrfox See the two commits before the last one.
It's now strictly checking that the `aud` claim contains only ourselves, and also I reverted the changes to the AuthController interface to have only common methods.

business/authentication/openid_auth_controller.go
```diff
@@ -0,0 +1,1437 @@
...
+	log.Errorf("Try to check nonce cookie")
+	if nonceCookie, err = r.Cookie(OpenIdNonceCookieName); err == nil {
+		log.Errorf("Nonce cookie = %v", nonceCookie.Value)
+		p.Nonce = nonceCookie.Value
+
+		hash := sha256.Sum224([]byte(nonceCookie.Value))
+		p.NonceHash = make([]byte, sha256.Size224)
+		copy(p.NonceHash, hash[:])
+	}
+	log.Errorf("Finish try to check nonce cookie")
```
nrfox, line 550: Are these debug msgs? Can they be removed?

israel-hdez: Oops! They are. I just removed them.

nrfox (approved):
Looks great and tested this on an open id deployment

israel-hdez:
@mattmahoneyrh This is almost an upstream-only feature. In OpenShift, this cannot be enabled. So, not sure if you would like to test  this or skip it. I'm don't remember what we've done in the past.

jshaughn:
@israel-hdez I think it's best to move this to the QE project and then they can move it through their workflow.  I would suggest QE just approve it and let it get out into the community for further evaluation.  Any OpenShift auth issue would be caught in standard Maistra testing.

mattmahoneyrh:
Verified that Openshift OAUTH works as expected.

mattmahoneyrh (approved):

Issue #4502, opened 2021-11-12
Create and sync namespace caches on startup


nrfox:
Namespace caches are currently created and synced when they are first accessed. When there are many namespaces, the Kiali server can take awhile to respond to the first request while it creates and syncs each of the namespace caches. This can be tested by:

1. Creating 300-400 empty namespaces
2. Restarting the kiali pod ensuring no other traffic is going to it
3. As the first request since startup, `curl https://kiali-url/api/mesh/tls`
4. Observe kiali pod logging `Waiting for Kiali cache for ... to sync`

Response times can take 30-60s.

Instead of creating and syncing the Namespace caches when first accessed, Kiali server can create/sync them on startup before accepting any user requests.

lucasponce:
Note that this request is somehow related to https://github.com/kiali/kiali/issues/4382 issue which is working out on https://github.com/kiali/kiali/pull/4479

Main issue here is that the validations logic for tls endpoint doesn't scale well, this issue is located and identified and the strategy we're following is to use Istio Registry in the validations logic, but there are a couple of preliminary steps before to attack the tls endpoint.

Loading all namespaces at startup can be a workaround but is not going to address the real issue here and perhaps it can introduce a slow deployment instead.

nrfox:
> Loading all namespaces at startup can be a workaround but is not going to address the real issue here and perhaps it can introduce a slow deployment instead.

Yes it moves the time spent creating/syncing the caches from "during the first request" to "before the first request". It will make the Kiali deployment slower to become ready initially but I'd argue that is less perceivable to the user than having the first request to Kiali be very slow. In other words, it is part of the setup process for the Kiali server and should be completed before the Kiali server considers itself ready to serve user requests.

> Main issue here is that the validations logic for tls endpoint doesn't scale well

I wouldn't say that's the main issue. It is just circumstance that  the `/mesh/tls` endpoint is both one of the first the UI calls and it uses the kiali cache across namespaces. Regardless of what endpoint the user hits first, if that endpoint uses the kiali cache, then they will need to wait for the caches to be created and synced before the request is completed.

lucasponce:
Not objection to populate the caches on the init phase.

Logic should proabably be placed in this block:
![image](https://user-images.githubusercontent.com/1662329/141941733-dd76cc16-43eb-41b0-a051-5f2565f7b604.png)

[...]

PR #4674, opened 2022-01-26
Init Kiali Cache on Server startup

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4502

Suggested test plan (QE @skondkar):
- This PR shouldn't change the current behavior from the user's perspective, it can be tested validating that when the Kiali server starts (with NO access from the UI, just deployment), the log should show that the Kiali has been populated.
- A basic access to the Overview page should be enough to validate regressions.

### DIFF
--- a/business/layer.go
+++ b/business/layer.go
@@ -1,6 +1,7 @@
 package business
 
 import (
+	"context"
 	"sync"
 
 	"k8s.io/client-go/tools/clientcmd/api"
@@ -44,9 +45,20 @@ var kialiCache cache.KialiCache
 
 func initKialiCache() {
 	if config.Get().KubernetesConfig.CacheEnabled {
+		log.Infof("Initializing Kiali Cache")
 		if cache, err := cache.NewKialiCache(); err != nil {
 			log.Errorf("Error initializing Kiali Cache. Details: %s", err)
 		} else {
+			// Temporal NamespaceService to initialize the cache informers
+			initK8sClient := cache.GetClient()
+			initNamespaceService := NewNamespaceService(initK8sClient)
+			if nss, err := initNamespaceService.GetNamespaces(context.Background()); err != nil {
+				log.Errorf("Error fetching initial namespaces for populating the Kiali Cache. Details: %s", err)
+			} else {
+				for _, ns := range nss {
+					cache.CheckNamespace(ns.Name)
+				}
+			}
 			kialiCache = cache
 		}
 	}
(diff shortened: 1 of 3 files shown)


business/layer.go
```diff
@@ -44,9 +44,20 @@ var kialiCache cache.KialiCache
...
 func initKialiCache() {
 	if config.Get().KubernetesConfig.CacheEnabled {
+		log.Infof("Initializing Kiali Cache")
 		if cache, err := cache.NewKialiCache(); err != nil {
 			log.Errorf("Error initializing Kiali Cache. Details: %s", err)
 		} else {
+			// Temporal NamespaceService to initialize the cache informers
+			initK8sClient := cache.GetClient()
+			initNamespaceService := NewNamespaceService(initK8sClient)
+			if nss, err := initNamespaceService.GetNamespaces(); err != nil {
```
nrfox, line 54: This might be out of scope for this change but do we need to create namespace specific caches if `config.Deployment.AccessibleNamespaces` is `**`?

Could the namespace names can come from `config.Deployment.AccessibleNamespaces` instead of the namespace service saving a bunch of kube API calls to fetch the full namespace object if there are many namespaces?

nrfox (commented):
Change works great and speeds up the initial kiali load. Just a couple questions.

business/layer.go
lucasponce: > This might be out of scope for this change but do we need to create namespace specific caches if config.Deployment.AccessibleNamespaces is **?

This is exactly my comment in https://github.com/kiali/kiali/issues/4502#issuecomment-1021162055 a potential optimization about this logic is to check/consider that the user/admin may grant cluster permissions to the Kiali's ServiceAccount.

In that case, instead, to create a separated informer per namespace+API, we can create a single informer per API monitoring all namespaces and that would really save memory and network, but again, it should be an optimization that will only work under that case, otherwise, with no cluster right the all-namespaces query will fail and this design should be used.

But yes, it was something that could be introduced and it can be a flag to consider for performance "you need to handle hundreds of namespaces like other Kubernetes admin tool? -> set cluster rights to the ServiceAccount".

nrfox (approved):

lucasponce:
@nrfox @skondkar I've rebased this PR with the master branch as there was a conflict with the context propagation.
It shouldn't affect the feature, let me know if it looks ok.

nrfox (approved):

PR #4676, opened 2022-01-27
Fix "xxx is not found as xxx" issue

ErikXu:
The excluded type(s) adding to controllers seems to cause "xxx is not found as xxx" issue.  
Let's take job type as an example. As it is excluded, the job array remains empty:
``` go
// business/workloads.go

	var jbs []batch_v1.Job  // line 593

	go func() {  // line 709
		defer wg.Done()
		var err error
		if isWorkloadIncluded(kubernetes.JobType) { // job type is excluded, condition is not met
			jbs, err = layer.k8s.GetJobs(namespace)
			if err != nil {
				log.Errorf("Error fetching Jobs per namespace %s: %s", namespace, err)
				errChan <- err
			}
		}
	}()
```

And the "is found" validation will always return false, which cause the  "xxx is not found as xxx" issue:
``` go
// business/workloads.go

		case kubernetes.JobType:  // line 1048
			found := false
			iFound := -1
			for i, jb := range jbs {
				if jb.Name == cname {
					found = true
					iFound = i
					break
				}
			}
			if found {
				selector := labels.Set(jbs[iFound].Spec.Template.Labels).AsSelector()
				w.SetPods(kubernetes.FilterPodsBySelector(selector, pods))
				w.ParseJob(&jbs[iFound])
			} else {  // always meet this condition
				log.Errorf("Workload %s is not found as Job", cname)
				cnFound = false
			}
```

I try to make controllers to not contain the excluded type(s) to fix the "xxx is not found as xxx" issue.

Looking forward to your review or better solution. 

Thanks!

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -746,7 +746,7 @@ func fetchWorkloads(ctx context.Context, layer *Layer, namespace string, labelSe
 	for _, pod := range pods {
 		if len(pod.OwnerReferences) != 0 {
 			for _, ref := range pod.OwnerReferences {
-				if ref.Controller != nil && *ref.Controller {
+				if ref.Controller != nil && *ref.Controller && !isWorkloadIncluded(ref.Kind) {
 					if _, exist := controllers[ref.Name]; !exist {
 						controllers[ref.Name] = ref.Kind
 					} else {
@@ -1355,7 +1355,7 @@ func fetchWorkload(ctx context.Context, layer *Layer, namespace string, workload
 	for _, pod := range pods {
 		if len(pod.OwnerReferences) != 0 {
 			for _, ref := range pod.OwnerReferences {
-				if ref.Controller != nil && *ref.Controller {
+				if ref.Controller != nil && *ref.Controller && !isWorkloadIncluded(ref.Kind) {
 					if _, exist := controllers[ref.Name]; !exist {
 						controllers[ref.Name] = ref.Kind
 					} else {
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -746,7 +746,7 @@ func fetchWorkloads(ctx context.Context, layer *Layer, namespace string, labelSe
 	for _, pod := range pods {
 		if len(pod.OwnerReferences) != 0 {
 			for _, ref := range pod.OwnerReferences {
-				if ref.Controller != nil && *ref.Controller && !isWorkloadIncluded(ref.Kind) {
+				if ref.Controller != nil && *ref.Controller && isWorkloadIncluded(ref.Kind) {
 					if _, exist := controllers[ref.Name]; !exist {
 						controllers[ref.Name] = ref.Kind
 					} else {
@@ -1355,7 +1355,7 @@ func fetchWorkload(ctx context.Context, layer *Layer, namespace string, workload
 	for _, pod := range pods {
 		if len(pod.OwnerReferences) != 0 {
 			for _, ref := range pod.OwnerReferences {
-				if ref.Controller != nil && *ref.Controller && !isWorkloadIncluded(ref.Kind) {
+				if ref.Controller != nil && *ref.Controller && isWorkloadIncluded(ref.Kind) {
 					if _, exist := controllers[ref.Name]; !exist {
 						controllers[ref.Name] = ref.Kind
 					} else {


nrfox:
Thanks for the contribution @ErikXu. I just want to clarify that I understand the issue here.

You've got a workload type you've excluded from Kiali as set in `Config.KubernetesConfig.ExcludeWorkloads` and yet Kiali is logging errors because it is trying to match pods to workloads of that type? e.g. you've excluded `Job`s and yet Kiali is finding Job pods and logging this error?

business/workloads.go
```diff
@@ -746,7 +746,7 @@ func fetchWorkloads(ctx context.Context, layer *Layer, namespace string, labelSe
 	for _, pod := range pods {
 		if len(pod.OwnerReferences) != 0 {
 			for _, ref := range pod.OwnerReferences {
-				if ref.Controller != nil && *ref.Controller {
+				if ref.Controller != nil && *ref.Controller && !isWorkloadIncluded(ref.Kind) {
```
nrfox, line 749: Shouldn't this be the opposite `isWorkloadIncluded`?

ErikXu:
Hello, @nrfox . Sorry for replying late. Yes, that's the exactly situation I met. I checked the `config.yaml`, the `Job` type is excluded, but I would still get the error or warning logs when refreshing the kiali UI.

business/workloads.go
nrfox, line 749: This should be the opposite:
```suggestion
				if ref.Controller != nil && *ref.Controller && isWorkloadIncluded(ref.Kind) {
```

business/workloads.go
```diff
@@ -1355,7 +1355,7 @@ func fetchWorkload(ctx context.Context, layer *Layer, namespace string, workload
...
```
nrfox, line 1358: This as well:
```suggestion
				if ref.Controller != nil && *ref.Controller && isWorkloadIncluded(ref.Kind) {
```

nrfox (changes requested):
Thanks again for the contribution. I can reproduce the issue and the fix works with one small suggestion.

ErikXu:
> Thanks again for the contribution. I can reproduce the issue and the fix works with one small suggestion.

Oh, the suggestion is really helpful. Thanks for your review!

nrfox (approved):

nrfox:
Note for testing:

To reproduce issue:
1. Deploy bookinfo
2. Add "Job" to excluded kiali workloads: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.kubernetes_config.excluded_workloads
kiali-cr.yaml
```
spec:
  kubernetes_config:
    excluded_workloads:
      - "Job"
```
3. Check kiali logs for "Workload <workload_name> is not found as Job" errors.

After fix the error logs should go away.

Also need to test that non-excluded workloads still log errors after the fix:
1. Remove `spec.kubernetes_config.excluded_workloads` line
2. Check kiali logs and "Workload <workload_name> is not found as Job" errors should appear again.

pbajjuri20:
@nrfox: Thanks for the steps.

I have tested this PR, it looks good to me.

ehdieunguyen:
Hi @nrfox 
This fix only applied to Kiali version from `v1.47`. So could you bring this to previous versions?

jmazzitelli:
> So could you bring this to previous versions?

No, we only backport fixes that far back if the fix is for a very severe CVE (security issue) and the Kiali version is for supported Istio versions (i.e. not EOL Istios).

From the compatibility matrix: https://kiali.io/docs/installation/installation-guide/prerequisites/

Istio | Kiali | notes
-- | -- | -- 
1.15 | 1.55.1 or later |  
1.14 | 1.50.0 to 1.54.x |  
1.13 | 1.45.1 to 1.49.x |  
1.12 | 1.42.0 to 1.44.x | Istio 1.12 is out of support.

Kiali 1.47 is only officially supporting Istio 1.13. So if you are on Istio 1.13, and are on Kiali versions prior to 1.47, and you want this bug fix, you need to upgrade to at least 1.47 or to a maximum of 1.49.

Istio versions 1.12 and before are EOL, so all Kiali versions prior to  1.45 are not being maintained. We do not backport bug fixes to those Kiali versions.

PR #4683, opened 2022-01-27
Fix IndexError: list index out of range and print json on error

FilipB:
Fixing possible 'IndexError: list index out of range' in E2E tests


FilipB:
@pbajjuri20 Could you please test before merging? Tests are not working with my python 3.10

pbajjuri20:
Tested locally, It looks good. Thanks for the changes @FilipB

PR #4689, opened 2022-01-30
fix: add dev to istioDevVersionExpr

Xunzhuo:

**Describe the change**
_Please describe what the code change is about._

istio dev version has dev prefix like this:
`1.10-dev-65a124dc2ab69f91331298fbf6d9b4335abcf0fd-Clean`

**Issue reference**

_Please, provide the link to the GitHub issue here._

**Backwards incompatible?**

_describe the changes if appropriate._

**Documentation**

If so, please also update [README.adoc](README.adoc)

### DIFF
--- a/status/versions.go
+++ b/status/versions.go
@@ -38,7 +38,7 @@ var (
 	maistraProductVersionExpr = regexp.MustCompile(`maistra-([0-9]+\.[0-9]+\.[0-9]+)`)
 	ossmVersionExpr           = regexp.MustCompile(`(?:OSSM_|openshift-service-mesh-)([0-9]+\.[0-9]+\.[0-9]+)`)
 	maistraProjectVersionExpr = regexp.MustCompile(`(?:Maistra_|openshift-istio.*-)([0-9]+\.[0-9]+\.[0-9]+)`)
-	istioDevVersionExpr       = regexp.MustCompile(`((\d+\.\d+)-alpha\.([[:alnum:]]+)-.*)|(\d+\.\d+)-(?:dev)`)
+	istioDevVersionExpr       = regexp.MustCompile(`(\d+\.\d+)-alpha\.([[:alnum:]]+)-.*|(\d+\.\d+)-dev-([[:alnum:]]+)-.*`)
 	istioRCVersionExpr        = regexp.MustCompile(`(\d+\.\d+.\d+)-((?:alpha|beta|rc|RC)\.\d+)`)
 	istioSnapshotVersionExpr  = regexp.MustCompile(`istio-release-([0-9]+\.[0-9]+)(-[0-9]{8})`)
 	istioVersionExpr          = regexp.MustCompile(`([0-9]+\.[0-9]+\.[0-9]+)`)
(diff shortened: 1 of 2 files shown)


jmazzitelli (changes requested):
We have a unit test for these version strings. Please add additional unit tests to show that the `-dev` version strings works like the `-alpha` version strings?

See unit tests here: https://github.com/kiali/kiali/blob/master/status/versions_test.go

jshaughn:
hi @Xunzhuo , Where do you see this version format?  You show `1.10-dev-65a124dc2ab69f91331298fbf6d9b4335abcf0fd-Clean` but when I look at the latest dev/commit build I see https://gcsweb.istio.io/gcs/istio-build/dev/?marker=dev%2f1.13-alpha.a0dd3c3f1e64864366d8793467bc5b5603a7d211%2f

jmazzitelli:
@jshaughn I'll let @Xunzhuo answer (so we know where he sees it) but IIRC I think these `-dev` versions come from when you build locally from source.

jshaughn:
> @jshaughn I'll let @Xunzhuo answer (so we know where he sees it) but IIRC I think these -dev versions come from when you build locally from source.

If that is true then we should add a new entry, not change the current dev build pattern.

jmazzitelli:
> If that is true then we should add a new entry, not change the current dev build pattern.

I think it might be easier doing what he did - he just added an "or" condition to the Dev pattern (dev, I'll assume, means developer builds). Otherwise the amount of code to add to introduce a new entry will just add more churn. If this change is backward compatible and doesn't break the alpha+hash detection, then I'm +1 on it. I think considering both "alpha+hash" and "dev+hash" builds the same is fine. No sense adding more code here just for non-release build detection, imo.

Xunzhuo:
> @jshaughn I'll let @Xunzhuo answer (so we know where he sees it) but IIRC I think these `-dev` versions come from when you build locally from source.

True, this comes from local build for istio.

status/versions.go
```diff
@@ -223,6 +224,16 @@ func parseIstioRawVersion(rawVersion string) (*ExternalServiceInfo, error) {
 			}
 			// we know this is Istio upstream - either a supported or unsupported version - return now
 			return &product, nil
+		} else if strings.Contains(istioVersionStringArr[0], "dev") && len(istioVersionStringArr) > 2 {
```
jmazzitelli, line 227: because you are going to be accessing the [3] and [4] indexes down on lines 229/230, you need to check that the length is good - otherwise you will throw an array index out of bounds error here if the regex is given something we are not expecting.
```suggestion
		} else if strings.Contains(istioVersionStringArr[0], "dev") && len(istioVersionStringArr) > 4 {
```

jmazzitelli (changes requested):

status/versions.go
Xunzhuo: @jmazzitelli Done.

jshaughn (approved):
look good to me

jmazzitelli (approved):
lgtm

PR #4690, opened 2022-01-30
feat: check mesh version if compatible with kiali when startup

Xunzhuo:

**Describe the change**

_Please describe what the code change is about._

**Issue reference**

_Please, provide the link to the GitHub issue here._

**Backwards incompatible?**

_describe the changes if appropriate._

**Documentation**

If so, please also update [README.adoc](README.adoc)


kiali.go
```diff
@@ -90,6 +90,21 @@ func main() {
 	status.Put(status.CoreCommitHash, commitHash)
 	status.Put(status.ContainerVersion, determineContainerVersion(version))
 
+	authentication.InitializeAuthenticationController(config.Get().Auth.Strategy)
```
jmazzitelli, line 93: since this is being done here, delete the dup call on 108

config/config.go
```diff
@@ -431,6 +431,29 @@ type HealthConfig struct {
 	Rate []Rate `yaml:"rate,omitempty" json:"rate,omitempty"`
 }
 
+// VersionRange version ranges
+type VersionRange struct {
+	Length      int
+	MeshVersion []string
+	KialiLow    []string
+	KialiHigh   []string
```
jmazzitelli, line 439: who/what/where defines these values?

Xunzhuo: According to https://kiali.io/docs/installation/installation-guide/prerequisites/#istio-version-compatibility

kiali.go
Xunzhuo: Done.

config/config.go
jmazzitelli: Yes, but where in the code or config is it set? This is something we are going to have to maintain "somewhere" - where is that matrix coded (i.e. where are the json values coming from that fill in this "VersionRange"?)

Xunzhuo: Thanks @jmazzitelli, I have added a yaml to record versions.

config/config.go
```diff
@@ -431,6 +431,38 @@ type HealthConfig struct {
 	Rate []Rate `yaml:"rate,omitempty" json:"rate,omitempty"`
 }
 
+// Versions version ranges
+type Versions struct {
+	MeshName     string `yaml:"meshName"`
+	VersionRange []struct {
+		MeshVersion string `yaml:"meshVersion"`
+		KialiLow    string `yaml:"kialiLow"`
+		KialiHigh   string `yaml:"kialiHigh"`
```
jmazzitelli, line 440: I would prefer this more descriptive. "KialiLow" and "KialiHigh" doesn't really tell you what these are. These should be something like "KialiMinimumVersion" and "KialiMaximumVersion":
```suggestion
		KialiMinimumVersion    string `yaml:"kialiMinimumVersion"`
		KialiMaximumVersion    string `yaml:"kialiMaximumVersion"`
```

config/config.go
```diff
@@ -431,6 +431,38 @@ type HealthConfig struct {
...
+	} `yaml:"versionRange"`
+}
+
+// NewVersions return compatible kiali versions for mesh
+func NewVersions() (*Versions, error) {
```
jmazzitelli, line 445: Failing to check for version compatibility doesn't mean Kiali won't work - it just means it doesn't know (just like today :-).

We should not return an error (or at least we should not abort on any error) here. Just log a Warning message like you do and keep going. Kiali should still run as before.

kiali.go
```diff
@@ -90,6 +90,37 @@ func main() {
 	status.Put(status.CoreCommitHash, commitHash)
 	status.Put(status.ContainerVersion, determineContainerVersion(version))
 
+	// check kiali version compatibility with istio
+	meshName, meshVersion, err := status.GetMeshInfo()
```
jmazzitelli, line 94: This is just a question I have: I haven't looked but - does this status.GetMeshInfo require the user to be authenticated in order to get the MeshInfo data from the server? I assume this will work regardless of the auth strategy being used.

version.yaml
```diff
@@ -0,0 +1,24 @@
+# compatible kiali version range with istio version
```
jmazzitelli, line 1: I do not like the name of this file. `version.yaml` implies this file contains the version of Kiali, which is not what it is. Rename this file to something more descriptive of what it is - something like `version-compatibility-matrix.yaml`.

kiali.go
```diff
@@ -90,6 +90,37 @@ func main() {
...
+			log.Info("Mesh Name: %v Mesh Version: %v", meshName, meshVersion)
+			status.Put(status.MeshVersion, meshVersion)
+			status.Put(status.MeshName, meshName)
+			status.Put(status.IsCompatible, "true")
+		} else {
+			log.Warningf("Kiali %v is not compatible with %v %v", version, meshName, meshVersion)
+			status.AddWarningMessages(fmt.Sprintf("Kiali %v is not compatible with %v %v", version, meshName, meshVersion))
+			status.Put(status.IsCompatible, "false")
+		}
+	}
```
jmazzitelli, line 122: I haven't run this to confirm, but I strongly suspect this will emit warnings about incompatibility when Kiali is run in an OpenShift Service Mesh environment, which we do not want. We should enhance this feature being added in this PR by adding to the compatibility matrix all the valid Kiali <-> OSSM versions.

jmazzitelli (dismissed):

kiali.go
Xunzhuo: Sure, when doing this task, I think we can check OSSM version at the same time as well.

version.yaml
Xunzhuo: Sure, I will update this name later.

config/config.go
Xunzhuo: Yes, checking version should not impact Kiali running status. I catch the error for logging warning messages.

config/config.go
Xunzhuo: I will update these names, but I will update related codes for it is using somewhere else.

Xunzhuo: Done

version.yaml
Xunzhuo: Done

kiali.go
Xunzhuo: In my view, querying version info from mesh service does not require to be authenticated.

version-compatibility-matrix.yaml
```diff
@@ -0,0 +1,70 @@
...
+      kialiMinimumVersion: "1.26.0"
+      kialiMaximumVersion: "1.28"
+    - meshVersion: "1.7"
+      SMCPVersion: ""
+      kialiMinimumVersion: "1.22.1"
+      kialiMaximumVersion: "1.25"
+    - meshVersion: "1.6"
+      SMCPVersion: ""
+      kialiMinimumVersion: "1.18.1"
+      kialiMaximumVersion: "1.21"
```
jshaughn, line 30: Just to be complete, I think this should go all the way back to meshVersion 1.0 with a maxKiali of 1.17.    But really none of this matrix will be useful because the check won't apply for kiali versions before the one that merges in this code, unless we do some backports and releases of older Kiali lines.

jshaughn (changes requested):

status/versions.go
```diff
@@ -108,6 +108,121 @@ func validateVersion(requiredVersion string, installedVersion string) bool {
...
+			}
+		}
+	}
+
+	return ok
+}
+
+// checkIstioVersion check istio if its version is compatible with kiali
+func checkIstioVersion(istioVersion string, kialiVersion string) bool {
+	ok := false
```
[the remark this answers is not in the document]
jmazzitelli: I haven't looked at what happens when this function returns false, but if we disrupt the user session at all, I think that would be a mistake. I think regardless of whether we determine the current version of Istio is "supported" or not, the user should be allowed to continue as-is normally. We should not log them out, we should not restrict access or functionality at all - they should be able to use Kiali normally as before. This new stuff should only result in a visible "warning" to the user. That warning should say something to the effect of, "Hey, if you see something funny happen, the first thing you should do is upgrade Kiali or Istio because the error is probably due to the fact you are running an obsolete/deprecated/unsupported Kiali compared to the Istio impementation you have installed"

In other words, it shouldn't matter what this function returns - the user should continue normally.

In that case (assuming the above is true - the user session is not affected no matter what this check returns), then I'm of no opinion what this should return if we cannot determine whether kiali supports the installed istio or not.

status/versions.go
```diff
@@ -108,6 +108,121 @@ func validateVersion(requiredVersion string, installedVersion string) bool {
...
+	Versions, err := config.NewVersions()
+
+	if err != nil {
+		log.Warningf("Can not load version file.%v", err)
```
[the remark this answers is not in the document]
Xunzhuo: We should not return here better, for this is kiali.go where impacts the actions of kiali.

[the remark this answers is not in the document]
Xunzhuo: I add if .. else .. to check if we should continue in the function.

kiali.go
```diff
@@ -237,3 +222,25 @@ func determineContainerVersion(defaultVersion string) string {
...
+func checkVersionCompatibility(meshName string, meshVersion string) {
+	ok := false
+	if meshName == "Unknown" {
+		log.Warningf("Unknown Istio Implementation")
+		status.Put(status.MeshVersion, meshVersion)
+		status.Put(status.MeshName, meshName)
+		status.AddWarningMessages("Unknown Istio Implementation")
+	} else {
+		ok = status.CheckMeshVersion(meshName, meshVersion, version)
+		if ok {
```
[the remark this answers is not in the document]
Xunzhuo: Done

kiali.go
```diff
@@ -237,3 +222,25 @@ func determineContainerVersion(defaultVersion string) string {
...
+			log.Info("Mesh Name: %v Mesh Version: %v", meshName, meshVersion)
```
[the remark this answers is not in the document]
Xunzhuo: Done

status/versions.go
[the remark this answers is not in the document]
Xunzhuo: Done and fix also in checkMaistraVersion and checkOSSMVersion

version-compatibility-matrix.yaml
Xunzhuo: But if meshVersion is 1.0, what is the minimum Version of kiali, which is compatible?

status/versions.go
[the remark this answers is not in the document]
Xunzhuo: Actually I followed this rule which this task will not affect kiali normally running status, which means no matter the default ok is true or false, the user should continue normally.

version-compatibility-matrix.yaml
```diff
@@ -0,0 +1,70 @@
+- meshName: "Istio"
+  versionRange: 
+    - meshVersion: "1.12"
+      SMCPVersion: ""
+      kialiMinimumVersion: "1.42.0"
+      kialiMaximumVersion: ""
```
jshaughn, line 6:
```suggestion
    - meshVersion: "1.13"
      SMCPVersion: ""
      kialiMinimumVersion: "1.45.1"
      kialiMaximumVersion: ""
    - meshVersion: "1.12"
      SMCPVersion: ""
      kialiMinimumVersion: "1.42.0"
      kialiMaximumVersion: "1.44.0"
```

kiali.go
```diff
@@ -211,3 +222,25 @@ func determineContainerVersion(defaultVersion string) string {
...
+		status.Put(status.MeshName, meshName)
+		status.AddWarningMessages("Unknown Istio Implementation")
+	} else {
+		if ok := status.CheckMeshVersion(meshName, meshVersion, version); ok {
+			log.Infof("Mesh Name: %v Mesh Version: %v", meshName, meshVersion)
+			status.Put(status.MeshVersion, meshVersion)
+			status.Put(status.MeshName, meshName)
+			status.Put(status.IsCompatible, "true")
+		} else {
+			log.Warningf("Kiali %v is not compatible with %v %v", version, meshName, meshVersion)
```
jshaughn, line 241: This message is fine for OSSM and Maistra, but for upstream Istio it may be too agressive.  We know it's not recommended (because it's not tested) but we may not be sure if it's incompatible.   I would suggest something like:

`log.Warningf("Kiali %v may not be compatible with %v %v, and is not recommended. See kiali.io for version compatibility", version, meshName, meshVersion)`

jshaughn (changes requested):
Looks good.  Only one suggested change.

version-compatibility-matrix.yaml
jmazzitelli: what @jshaughn proposes seems reasonable.

kiali.go
jmazzitelli: Yeah, let's warn them that it MAY NOT be compatible. I think the suggested change is good.

version-compatibility-matrix.yaml
Xunzhuo: Done.

kiali.go
Xunzhuo: Done.

jshaughn (changes requested):
Please add the remaining versions to the matrix, as described in the comment.

Xunzhuo:
@jshaughn Done and Thanks. :)

jshaughn (changes requested):
Code looks good, compat matrix looks good, but did you test this using an actual build?  I don't
think `version-compatibility-matrix.yaml` is actually getting included in the container.  Try `make build cluster-push operator-create kiali-create` and I think you will see `WRN Can not load version file: open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory`

jmazzitelli:
> Code looks good, compat matrix looks good, but did you test this using an actual build? I don't think `version-compatibility-matrix.yaml` is actually getting included in the container. Try `make build cluster-push operator-create kiali-create` and I think you will see `WRN Can not load version file: open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory`

I suspect the file needs to be added [here](https://github.com/kiali/kiali/blob/master/deploy/docker/Dockerfile-multi-arch#L23) and [here](https://github.com/kiali/kiali/blob/master/deploy/docker/Dockerfile-ubi8-minimal#L17). Something like:

```
ADD <local path of the matrix yaml> $KIALI_HOME
```

Makefile
```diff
@@ -19,6 +19,9 @@ COMMIT_HASH ?= $(shell git rev-parse HEAD)
 # output in the $UI_SRC_ROOT/build/ subdirectory.
 CONSOLE_LOCAL_DIR ?= ${ROOTDIR}/../../../../../kiali-ui
 
+# The path where the version-compatibility-matrix.yaml has been git cloned.
+VERSION_METRIX_LOCAL_DIR ?= ${ROOTDIR}/version-compatibility-matrix.yaml
```
jmazzitelli, line 23: I just realized in several places the word "matrix" is being misspelled - here (and elsewhere) you have "metrix" (with an "e" not an "a") --- the correct spelling is "matrix".

make/Makefile.container.mk
```diff
@@ -9,6 +9,7 @@
 	@mkdir -p ${OUTDIR}/docker
 	@cp -r deploy/docker/* ${OUTDIR}/docker
 	@cp ${GOPATH}/bin/kiali* ${OUTDIR}/docker
+	@cp ${VERSION_METRIX_LOCAL_DIR} ${OUTDIR}/docker
```
jmazzitelli, line 12: misspelled "matrix"

version-compatibility-matrix.yaml
```diff
@@ -0,0 +1,103 @@
+# The Compatible Version Metrix between upstream Istio and Kiali
```
jmazzitelli, line 1: misspelled "matrix"

version-compatibility-matrix.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - meshVersion: "1.1"
+      SMCPVersion: ""
+      kialiMinimumVersion: "1.17"
+      kialiMaximumVersion: "1.17"
+    - meshVersion: "1.0"
+      SMCPVersion: ""
+      kialiMinimumVersion: "1.17"
+      kialiMaximumVersion: "1.17"
+
+# The Compatible Version Metrix between Maistra and Kiali
```
jmazzitelli, line 61: misspelled "matrix"

version-compatibility-matrix.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - meshVersion: "2.0"
+      SMCPVersion: "1.1"
+      kialiMinimumVersion: "1.12"
+      kialiMaximumVersion: ""
+    - meshVersion: "1.1"
+      SMCPVersion: "1.1"
+      kialiMinimumVersion: "1.12"
+      kialiMaximumVersion: ""
+
+# The Compatible Version Metrix between OpenShift Service Mesh and Kiali
```
jmazzitelli, line 89: misspelled "matrix"

jmazzitelli (changes requested):
trivial change - correct the misspellings of the word "matrix"

Makefile
Xunzhuo: Done.

make/Makefile.container.mk
Xunzhuo: Done.

version-compatibility-matrix.yaml
Xunzhuo: Done.

jmazzitelli (commented):
still testing, but so far so good - I see the matrix file has now made it into the image.

1. deploy everything to minikube
```
make -e CLUSTER_TYPE=minikube make build cluster-push operator-create kiali-create
```

2. find the matrix file:
```
$ kubectl exec -it -n istio-system $(kubectl get pod -n istio-system -l app.kubernetes.io/name=kiali -o name) -- ls -l /opt/kiali/version-compatibility-matrix.yaml

-rw-rw-r-- 1 root root 2973 Feb  9 14:26 /opt/kiali/version-compatibility-matrix.yaml
```

jmazzitelli (changes requested):
@Xunzhuo you need to test this when deploying in a cluster (i.e. unit tests won't be enough to validate).

I'm still getting the error @jshaughn got - the path name looks incorrect:

```
2022-02-09T14:27:20Z WRN Can not load version file open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory
2022-02-09T14:27:20Z WRN Can not load version file: open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory
```

To test, start up minikube, install Istio, and deploy using the make target as I document in my previous comment. We have make targets and hack scripts to help you do all of this, e.g.:

```
$ hack/k8s-minikube.sh start
$ hack/k8s-minikube.sh istio
$ make -e CLUSTER_TYPE=minikube make build cluster-push operator-create kiali-create
```

Those 3 commands will get you a minikube cluster with Kiali installed, at which time you can look at the logs via `kubectl logs` and peek into the image via `kubectl exec`

Xunzhuo:
> A secondary comment - I see the warning twice:
> 
> ```
> 2022-02-09T14:27:20Z WRN Can not load version file open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory
> 2022-02-09T14:27:20Z WRN Can not load version file: open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory
> ```
> 
> which tells me we are checking version compatibility multiple times. We only need to do it once - so just make sure we don't validate more than once. It is also possible that rather than checking twice, we are logging the message twice through the call stack (I say that because the warning message is ever so slightly different - notice the extra `:` (colon) after "file" in the second message). So just make sure we only log this once - no need to repeat the warning.

Actually I do check version once, just a duplicated log, solve this just by moving it.

Xunzhuo:
> @Xunzhuo you need to test this when deploying in a cluster (i.e. unit tests won't be enough to validate).
> 
> 
> 
> I'm still getting the error @jshaughn got - the path name looks incorrect:
> 
> 
> 
> ```
> 
> 2022-02-09T14:27:20Z WRN Can not load version file open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory
> 
> 2022-02-09T14:27:20Z WRN Can not load version file: open /opt/kiali/../version-compatibility-matrix.yaml: no such file or directory
> 
> ```
> 
> 
> 
> To test, start up minikube, install Istio, and deploy using the make target as I document in my previous comment. We have make targets and hack scripts to help you do all of this, e.g.:
> 
> 
> 
> ```
> 
> $ hack/k8s-minikube.sh start
> 
> $ hack/k8s-minikube.sh istio
> 
> $ make -e CLUSTER_TYPE=minikube make build cluster-push operator-create kiali-create
> 
> ```
> 
> 
> 
> Those 3 commands will get you a minikube cluster with Kiali installed, at which time you can look at the logs via `kubectl logs` and peek into the image via `kubectl exec`
> 
> 
> 
> For the record, here's what I see in the /opt/kiali directory inside the image:
> 
> 
> 
> ```
> 
> $ kubectl exec -it -n istio-system $(kubectl get pod -n istio-system -l app.kubernetes.io/name=kiali -o name) -- ls -l /opt/kiali
> 
> total 61468
> 
> drwxrwxr-x 1 kiali kiali     4096 Feb  9 14:26 console
> 
> -rwxrwxr-x 1 root  root  62931354 Feb  9 14:26 kiali
> 
> -rw-rw-r-- 1 root  root      2973 Feb  9 14:26 version-compatibility-matrix.yaml
> 
> ```
> 
> 

Actually I try to run these commands to test, but failed for a couple of times. I tried to set ClusterType to local, but there are still some conflicts with my env. I will try to fix them later, I think there are some bugs in these scripts. But I need to retest to confirm.

jmazzitelli:
We haven't tested the scripts on, say, MacOS. I run them on RHEL 7.x. So it might have to do with whatever operating system you are on. CLUSTER_TYPE=local won't work most likely - that's for special situations (which I forget).

jmazzitelli:
> I tried to run these commands to test, but failed for a couple of times.

You don't have to use these scripts - however you work in your environment is fine. So long as you deploy Kiali in a cluster that also has Istio.

Xunzhuo:
Thanks @jmazzitelli @jshaughn , I tested and deploy kiali into cluster. I think everything is fine now.

image

jshaughn (dismissed):
This is working for me, looks good!

jmazzitelli (changes requested):
@Xunzhuo I got around to doing a more deep-dive review and testing of this. I have some requested changes - but I did them already in a PR to your PR - please see: https://github.com/Xunzhuo/kiali/pull/1

jmazzitelli:
this is what the UI will look like:
image

Xunzhuo:
looks good @jmazzitelli !

jmazzitelli (approved):
ran this and it works. lgtm

Issue #2447, opened 2020-02-26
Ensure all validations has object references


xeviknal:
Referenced objects on validations were introduced in https://github.com/kiali/kiali/pull/1551.
However, it was a first iteration. There wasn't any task to keep adding new references.

Make sure the following object type validations have references:
- [x] #4693
- [x] Destination Rules
- [x] Gateways
- [x] MeshPolicies
- [x] ServiceEntries
- [x] Services
- [x] #4695
- [x] Sidecars

This can be a small epic.

lucasponce:
Added to the Project per priorities purposes.

lucasponce:
This effort could be use to revisit the side panel in the Istio Config editor.

Today is used in some resources but I think it could be used in a more general way covering:
- References of the workloads linked with a resource (using workload selectors).
- References of the internal services linked with a resource.
- References of the external destinations defined through ServiceEntry (a destination in a VirtualService can point to a ServiceEntry host, so the side panel could include these cases as well).

Perhaps the current decorators for the VirtualService / DestinationRules can be revisited if they are adding value.
In particular I'd say more interesting to help user to understand how a matching rule is performed rather than to show visually the distribution, in other words, in the context of this effort if some of those looks outdated I'd be ok to remove it.

hhovsepy:
These newly adding references are not validations, but a reference like we have in DestinationRule overview page linking to Service.
Question: So do we need to add reference in UI only or coming from modified backend which should support returning references without validations?

lucasponce:
I think this one is should be part of https://github.com/kiali/kiali/issues/4241

hhovsepy:
The road-plan for this epic is:

- Research, organize an open discussion to choose from two concepts.
- - Thick UI, thin back-end: First concept is to have all references generated in UI only.
- - Thin UI, thick back-end: Second concept it is creating all links in backend (similar to validations) and sending list of IstioLinkKeys (new struct for reference types besides the validations) to UI.
- Define the scope of this epic and create sub-epics.
- Complete the fist 2 PoC sub-epics (back-end, UI) to show the references to one type of objects.
- Do a demo to team.
- Collect the feedback.
- Complete the rest of the epics.
- Final demo.

lucasponce:
> Thick UI, thin back-end: First concept is to have all references generated in UI only.
> Thin UI, thick back-end: Second concept it is creating all links in backend (similar to validations) and sending list of IstioLinkKeys (new struct for reference types besides the validations) to UI.

[...]

PR #4694, opened 2022-01-31
Adding References Service for Istio Config Details page.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/2447

PoC for Istio Config References service for VirtualServices, DestinationRules and Gateways.

When setting "validation=true" param in IstioConfigDetails request, this will return "references" in json.
Which consists of "objectReferences", "serviceReferences" and "workloadReferences" arrays.


business/istio_references.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+
+	// time this function execution so we can capture how long it takes to fully validate this istio object
+	timer := internalmetrics.GetSingleValidationProcessingTimePrometheusTimer(namespace, objectType, object)
+	defer timer.ObserveDuration()
+
+	wg := sync.WaitGroup{}
+	errChan := make(chan error, 1)
+
+	// Get all the Istio objects from a Namespace and all gateways from every namespace
```
jshaughn, line 53: not necessarily gateways, right, all 'objectType's?

business/layer.go
```diff
@@ -144,6 +145,7 @@ func NewWithBackends(k8s kubernetes.ClientInterface, prom prometheus.ClientInter
 	temporaryLayer.TLS = TLSService{k8s: k8s, businessLayer: temporaryLayer}
 	temporaryLayer.TokenReview = NewTokenReview(k8s)
 	temporaryLayer.Validations = IstioValidationsService{k8s: k8s, businessLayer: temporaryLayer}
+	temporaryLayer.References = IstioReferencesService{k8s: k8s, businessLayer: temporaryLayer}
```
jshaughn, line 148: place alphabetically

business/references/virtual_service_references.go
```diff
@@ -0,0 +1,117 @@
...
+						result = append(result, getGagewayReferences(match.Gateways, namespace, clusterName)...)
+					}
+				}
+			}
+		}
+	}
+	return result
+}
+
+func getGagewayReferences(gateways []string, namespace string, clusterName string) []models.IstioReference {
```
jshaughn, line 105: typo in func name

business/istio_references.go
```diff
@@ -0,0 +1,129 @@
...
+	// time this function execution so we can capture how long it takes to fully validate this istio object
+	timer := internalmetrics.GetSingleValidationProcessingTimePrometheusTimer(namespace, objectType, object)
+	defer timer.ObserveDuration()
+
+	wg := sync.WaitGroup{}
+	errChan := make(chan error, 1)
+
+	// Get all namespaces and the requested Istio object details
+	wg.Add(2)
+	go in.fetchNamespaces(ctx, &namespaces, errChan, &wg)
```
lucasponce, line 55: I'm not sure if I understand this approach.
I think that the references should try to be an internal calculation (probably when the registry info is pulled) and the business method can pull that data.

With this approach, the "fetchNamespaces" and "fetchIstioConfigDetails" won't be cheap calls and it will add global overhead.

Also, if this service is internal (not exposed directly in the handlers, just part of the istio config) some checkers like the "GetNamespace" can be skipped (we'll know if the user can access that namespace in an earlier call in the main get Istio).

The model is good.
I'd suggest exploring the way how the model is built, some ideas:
- Let's place this logic closer to the IstioRegistry logic.
- When happens a refresh in the Istio Registry, let's trigger a "built references" process (it can be async).
- The Istio Registry will provide all Istio Config, all Services, all Endpoints; with all this info probably we can build all references required for the references (perhaps some corner case here in the very far case that a headless workload has specific Istio config - that case can be added later - but I'm sure that main scenarios would happen).
- The logic to append the references to the Istio Config looks good, but I think with this approach we won't miss any feature but we'll make it more lightweight.

We can assume that cache will be always on, yes, we need to change that and it's normal that informers the only way to access some info; I wouldn't enter in the case to make this feature available when the cache is off (probably we'll put cache on fixed in a next step).

jshaughn: I agree with this.  If we have the ability use more of a precompute approach, it will be much more lightweight.  This information does not change frequently.

hhovsepy: Agree to have References running faster and not retrieving any information which have impact on porformance.
There was an epic to make Validations triggered by Registry change to be recalculated and stored in a cache for improving the performance. I would suggest to have Validations and References triggered and cached be done together within the single epic.
For now, my suggestion will be, to move References closer to Validations and hence reuse the RegistryServices, IstioConfigs and Workloads which are already retrieved for Validations. This will make a minimal impact on performance.
Will send PoC with my suggestion.

lucasponce:
Note that this PR is indirectly related to https://github.com/kiali/kiali/pull/4710 work. In that PR there are being reviewed some optimizations in the validations code that will be relevant for this work.

hhovsepy:
@lucasponce @jshaughn now the PoC works with Validations service reusing the loaded Registry data.

lucasponce (approved):
The approach looks good to me.
It has a similar design as the validations and it's reusing the same fetched information.

I cc @leandroberetta as he is also involved in the visualization of this info, but it can be a start.

As the references are triggered in the IstioConfig detail page I don't see an issue in the processing overhead.

hhovsepy:
For QE: This is backend PR only, can be tested within https://github.com/kiali/kiali-ui/pull/2299

Issue #4691, opened 2022-01-30
[make][molecule] need to wait for SA/pull secret before we create our dev deployments


jmazzitelli:
For the gory details, see: https://bugzilla.redhat.com/show_bug.cgi?id=2048180

In short, we need to enhance our `make operator-create` target - right now it won't work all the time if deploying on OpenShift. We need to wait for the service account and its pull secret to be deployed before creating the deployment. How we are going to be able to do this I have no idea because we use helm to deploy everything. Somehow we need helm to wait to deploy the deployment until after the SA/pull secret are fully deployed. Alternative is to delete the pod when it comes up so it can restart hopefully with the fully deployed SA/pull secret (in this case we now cannot have helm undeploy and cleanup on install failure because we need the resources to remain, and we'll just delete the pod).

Until this is fixed, if the operator fails to start after a `make operator-create` you need to delete the pod and hope when the new one comes up it will have the SA/pull secret and can pull the image. You can do with like this `oc delete pod -n kiali-operator -l app=kiali-operator`

This error also shows up in our molecule tests (because the molecule testing uses the helm chart by default when installing the operator).

jmazzitelli:
One suggestion that was given to me (that doesn't sound appealing, but I will document them anyway) is: Install an initContainer on the Deployment that sits and waits for the SA and pull secret to become available.

We don't want this initContainer all the time - it should only be injected when we know we are (a) on OpenShift and (b) the container image is to be found in the OpenShift internal image registry

jmazzitelli:
There is another interesting solution provided to me in the [BZ](https://bugzilla.redhat.com/show_bug.cgi?id=2048180#c3) which said:

===

As with any other registry, either you need to make images public, or you need to provide the deployment with a pull secret. For example, you can create a service account in the namespace kiali, get its token, and create the pull secret.

Something like this:

```
oc -n kiali create serviceaccount my-service-account
oc -n kiali registry login --registry="$(oc registry info --internal)" -z my-service-account --to=./config.json
oc create secret generic my-secret --from-file=.dockerconfigjson=./config.json --type=kubernetes.io/dockerconfigjson --dry-run=client -o yaml
```

jmazzitelli:
For now, I plan on adding a make target that is analogous to our current `kiali-reload-image` target - I'll call it `operator-reload-image` so when you run `make operator-reload-image` it will simply delete the operator pod, restarting it (if you do this after the SA/pull secret is established, the operator will be able to start up).

PR #4698, opened 2022-02-01
[make] add target to restart the operator pod

jmazzitelli:
This is needed to help workaround this [issue](https://github.com/kiali/kiali/issues/4691) which does not yet have a solution.


jshaughn (approved):
Fine.  This doesn't seem like a great solution to the problem but can be useful on it's own, I guess.

jmazzitelli:
> This doesn't seem like a great solution to the problem

It isn't a solution - it's just a workaround for now.

PR #4706, opened 2022-02-03
Support being able to reliably load dev images from the internal OpenShift image registry

jmazzitelli:
This fixes the dev environment make targets.

Requires update to operator repo - some molecule code needs to change with this.

fixes: https://github.com/kiali/kiali/issues/4691


jmazzitelli:
To test that this works on OpenShift (and doesn't break minikube), run `make operator-create` multiple times in succession - seeing that the operator is deployed successfully each time:
```
make build cluster-push
make operator-create
make operator-create
```
Run the same as above with `-e CLUSTER_TYPE minikube` to see that it works on minikube.

This (combined with https://github.com/kiali/kiali-operator/pull/485) shows molecule tests passing on OpenShift without problems:
```
hack/run-molecule-tests.sh --cluster-type openshift --client-exe $(which oc) -udi true 
```

Similarly, making sure minikube isn't broken:
```
hack/run-molecule-tests.sh --cluster-type minikube --client-exe $(which kubectl) -udi true
```

jshaughn (approved):
Looks sane.

PR #4716, opened 2022-02-08
Bumping go to version 1.17

nrfox:
Note that as of go 1.17, indirect dependencies get moved to their own `require` block: 
> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.

In moving the indirect dependencies, the versions changed a bit but all tests pass.


aljesusg:
we need to change here the version https://github.com/kiali/kiali/blob/master/Makefile#L31

jmazzitelli (commented):
how is this going to affect downstream builds? IIRC we are not building with 1.17 downstream

nrfox:
> how is this going to affect downstream builds? IIRC we are not building with 1.17 downstream

There's a 1.17 builder. We need to bump the builder version we are using there.

jshaughn (approved):
Force me to update my local env.  Smoke test seems to all be working against KinD.  LG.

aljesusg:
Please not merge this until I verify the build in downstream

aljesusg (approved):
LGFM, build done in downstream

PR #4718, opened 2022-02-08
[backport v1.36] Bump go version to 1.17

nrfox:


go.mod
```diff
@@ -1,27 +1,20 @@
 module github.com/kiali/kiali
 
-go 1.14
+go 1.17
```
aljesusg, line 3: I wonder if go mod go to the latest go 1.17 (1.17.6 in this case) ..
I don't know the behaviour of this

nrfox: go.mod just specifies the minimum version of go needed to build with. As long as the build system has v1.17+ it'll use whatever version the build system is using e.g. 1.17.1, 1.17.6, 1.18.1, 1.23.0...

Reference: https://go.dev/doc/modules/gomod-ref
> The minimum version of Go required by the current module.

PR #4721, opened 2022-02-09
[v1.24] protobuf version upgraded to v1.3.2

leandroberetta:
Protobuf version upgraded to v1.3.2 due to vulnerabilities.

The testing I did was:

- Install v1.24 with Istio 1.9
- Install bookinfo demo
- Check that traffic is generating traces
- Check that the traces are accessible from Kiali, I used the API (workloads/spans) directly because I couldn't get the UI to work locally


lucasponce (approved):
I haven't tested locally but @leandroberetta smoke tests worked.
If any regression due to the change of library those endpoints should fail.

PR #4409, opened 2021-10-06
adding a checker that warns about workloads which are not covered by …

shireenf-ibm:
…any authorization policy

** Describe the change **
new checker files:

-> I have added a new checker under business/checkers/authorization called "uncovered_workload_checker.go", this checker gets lists of authorizationPolicies and Workloads and checks if there is a workload in the workloadslist that is not covered by any of the given authorization policies.

-> a test file called "uncovered_workload_checker_test.go" was added too, it tests "good" and "bad" cases of the checker.
a comment that describes the testing case was added before each call of the "Check" method of the new checker.

more changes in code are:
-> " business/checkers/authorization_policies_checker.go" file : 
since my checker's main parent is AuthorizationPolicyChecker, I have added the call for the new checker here in the "runGroupChecks" as it checks all auhorizationPolicies.

->  "models/istio_validation.go" : added a description of the new checker, to be compared with in the checker tests.
->  "tests/data/authorization_policy_data.go" : for my tests I needed to create new authorization policies such as: empty authorization policy or an authorization policy with specified namespace and selector. So new creators were added here.

** Issue reference **
Fixes https://github.com/kiali/kiali/issues/4368

UI PR: [#2278](https://github.com/kiali/kiali-ui/pull/2278)
Helm PR: [#120](https://github.com/kiali/helm-charts/pull/120)
Operator PR: [#467](https://github.com/kiali/kiali-operator/pull/467)

** Backwards incompatible? ** 
- Is your change introducing backwards incompatible changes?No

PR #4723, opened 2022-02-10
Adjust ignore validation log level

lucasponce:
Related to the AuthorizationPolicy validations.
Now that there is a default ignored validation, it could be good to move that log entry to the "debug" level to not populate the "info". Otherwise, it will add extra overhead in the Kiali pod log.


lucasponce:
Related to https://github.com/kiali/kiali/pull/4409

hhovsepy (approved):
+1 on the changes.

PR #4728, opened 2022-02-11
feat: add apiserver config when out of cluster

Xunzhuo:

**Describe the change**

Add apiserver config fields, kiali can talk to specfic apiserver.

This is only work outside of cluster.

**Backwards incompatible?**

No

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ COMMIT_HASH ?= $(shell git rev-parse HEAD)
 # have been built before trying to create a kiali server container
 # image. The UI project is configured to place its build
 # output in the $UI_SRC_ROOT/build/ subdirectory.
-CONSOLE_LOCAL_DIR ?= ${ROOTDIR}/../../../../../kiali-ui
+CONSOLE_LOCAL_DIR ?= ${ROOTDIR}/../kiali-ui
 
 # Version label is used in the OpenShift/K8S resources to identify
 # their specific instances. Kiali resources will have labels of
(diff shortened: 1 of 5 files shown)


jmazzitelli:
Kiali is not designed or expected to be run out of cluster. We have no testing, documentation, or anything that talks about running outside of the cluster - it is not supported.

Unless I hear more evidence for why we would want to encode this into the server, I'm -1 on this PR.

If you want to run Kiali out of cluster in your dev environment, you can use our hack script that does this - run-kiali.sh. For details see the comments and the `--help` output of the run-kiali.sh hack script here: https://github.com/kiali/kiali/blob/master/hack/run-kiali.sh

jmazzitelli:
@lucasponce what do you think? I know you do alot with configuring those KUBERNETES env vars for out of cluster access? Should we do this PR?

jmazzitelli:
(as a side note - for any dev that wants to add new server config options, its more than just adding to config.go - see the checklist here for the things you need to do: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting)

jmazzitelli:
But how do you set this? Are we saying we will allow Kiali to be deployed in cluster A but talk to cluster B? So I could deploy in cluster A, set those config settings in the Kiali CR, and have Kiali talk to cluster B?  Won't that require additional things in order for Kiali to be authenticated? Is it true that those 2 settings are all that really need to be set to connect to a remote cluster?

Makefile
```diff
@@ -17,7 +17,7 @@ COMMIT_HASH ?= $(shell git rev-parse HEAD)
 # have been built before trying to create a kiali server container
 # image. The UI project is configured to place its build
 # output in the $UI_SRC_ROOT/build/ subdirectory.
-CONSOLE_LOCAL_DIR ?= ${ROOTDIR}/../../../../../kiali-ui
+CONSOLE_LOCAL_DIR ?= ${ROOTDIR}/../kiali-ui
```
jmazzitelli, line 20: This is already in master (see https://github.com/kiali/kiali/blob/master/Makefile#L20) so I suspect this PR needs to be rebased on the latest master.

jmazzitelli: I'm also not convinced we want this PR yet. More discussions are probably needed around multi-cluster support.

Xunzhuo: Sure I will close this PR, and send a more complete feature PR in multi cluster area after some careful discussions:)

Issue #4605, opened 2021-12-16
Research workload/service label filters on Graph


lucasponce:
Kiali offers visualization for different pre-defined types of Graphs: applications, workload, services, and versioned applications.

During the scalability discussions, https://github.com/kiali/kiali/discussions/4080 one area to improve is to let the user define the list of elements to represent in a Graph using filtering expressions.

It's a common practice in Kubernetes to use labels to filter workloads and services and that feature could be incorporated in Kiali as well as part of the Graph generation.

Additionally, a Graph generated using filtering expressions may introduce dynamic boxing/grouping (in addition to the standard {app, namespace, boxing} capabilities that Kiali offers today).

But this feature would require some research to identify the potential subtasks needed as this effort may require changes in how the graph model is built and rendered.

Suggested test plan (QE @mattmahoneyrh):
- This Epic will research and define a new filter capability in the Kiali Graph based on workload/service labels.
- The QE test plan should define a new UI flow that validates the new changes proposed in the Epic.
- This Epic may be split into different subtasks (the definition of the tasks is part of the initial research), QE/Dev should sync if the test plan may be split into subtasks as well or consolidate a final test once the Epic is progressed.

jshaughn:
I think this is also related to https://github.com/kiali/kiali/issues/1269, where we talk about using labels as part of a global scope.

jmazzitelli:
It seems like we will need a label filtering capability similar to what the OpenShift Console has. We will need this in the Graph page as well as the list pages for Applications, Workloads and Services.

Without getting into specifics of how we should implement this in the UI, these are just illustrations to show something that we would need (notice the new label filter under the Namespace dropdown - this label filter is very similar to the OpenShift Console label filter component):

![graph-label-filter](https://user-images.githubusercontent.com/2029470/152240171-206b46c0-3557-4c78-b216-558bc1f91b2c.jpg)

![label-filter](https://user-images.githubusercontent.com/2029470/152240186-81584bd4-7567-422b-962a-bba268e65859.jpg)

[...]

PR #4731, opened 2022-02-14
Add Finalizers to the graph generation architecture

jshaughn:
This is pure code organization at the moment.  But I think it may be useful for some of the
projected features.  I am tying this to a current epic, which likely
will want to add k8s label info to graph nodes. This could be done in
an appender, but may lend itself better to being done one time, at the end.
Even if we don't add more finalizers, it's still not a bad thing.  There
isn't really a perf hit and I think it adds some structure.

relates to https://github.com/kiali/kiali/issues/4605

### DIFF
--- a/graph/telemetry/istio/finalizer/traffic_generator.go
+++ b/graph/telemetry/istio/finalizer/traffic_generator.go
@@ -1,0 +1,43 @@
+package finalizer
+
+import (
+	"github.com/kiali/kiali/graph"
+)
+
+const TrafficGeneratorFinalizerName = "trafficGenerator"
+
+// TrafficGeneratorFinalizer is responsible for marking the insider traffic generator nodes (i.e. inside the namespace and only having outgoing edges)
+// Name: trafficGenerator
+type TrafficGeneratorFinalizer struct{}
+
+// Name implements Finalizer
+func (f *TrafficGeneratorFinalizer) Name() string {
+	return OutsiderFinalizerName
+}
+
+// FinalizeGraph implements Appender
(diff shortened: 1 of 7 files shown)


graph/telemetry/istio/finalizer/traffic_generator.go
```diff
@@ -0,0 +1,43 @@
...
+// TrafficGeneratorFinalizer is responsible for marking the insider traffic generator nodes (i.e. inside the namespace and only having outgoing edges)
+// Name: trafficGenerator
+type TrafficGeneratorFinalizer struct{}
+
+// Name implements Finalizer
+func (f *TrafficGeneratorFinalizer) Name() string {
+	return OutsiderFinalizerName
+}
+
+// FinalizeGraph implements Appender
```
jmazzitelli, line 18: Do you mean this?

```suggestion
// FinalizeGraph implements Finalizer
```

jshaughn:
@mattmahoneyrh , this is the class "regression-test-only" sort of PR.  No functional change.  I think it's fine to approve and perform some sanity regression-testing when it makes sense, potentially on a schedule rather that per-PR.

mattmahoneyrh:
@jshaughn Is there a plan for running the Molecule tests against the PR or is it expected that will be done on the nightly build after the merge?

jmazzitelli (approved):
code looks good.
The only molecule test that would exercise this is `api-test` - all the other molecule tests do not generate graphs using the server side API affected by this PR.

jmazzitelli:
fyi: ran `api-test` molecule test and all pass.

pbajjuri20:
The regression test looks good ..!

pbajjuri20 (approved):

Issue #4664, opened 2022-01-18
feat(multitenancy): support additional metric label for prometheus


Xunzhuo:
## Background
>This is a feature proposal of multi-clusters and multi-meshes.

In our production environment, prometheus has large storage which can be used to receive metrics of multi-mesh or clusters. We distinguish mesh by  label `mesh` and cluster by label `cluster` like below:
![image](https://user-images.githubusercontent.com/48784001/149944727-4bf34066-92b2-4b08-b28e-8a7d63672ad7.png)

### Multi-Meshs with Multi-Clusters

**Relationship(Prometheus:Mesh:Kiali:Cluster): 1:1:N:M** (N<=M)
We distinguish metrics by additional `mesh` and `cluster` label
![image](https://user-images.githubusercontent.com/48784001/149954563-6cb4aae1-9597-4351-8a81-3b0f91544179.png)

### Single Mesh with Multi-Clusters

**Relationship(Prometheus:Mesh:Kiali:Cluster): 1:1:1:N**  
We distinguish metrics by additional `cluster` label
![image](https://user-images.githubusercontent.com/48784001/149954923-7c4637fe-ce81-4c1f-9956-7824ff330e50.png)

In kiali, it seems like we miss some additional dimensions option for receiving multi-clusters/meshes in one prometheus.

For example:
I deploy kiali with cluster `kubernetes` to observe it in meshA, the other cluster is not deployed with kiali in meshB. These two clusters are running the same apps/services/workloads in the same namespaces. Their metrics are recieved in the same prometheus.

Namespace Graph will show the other cluster`s graph(which has the namespace):
![image](https://user-images.githubusercontent.com/48784001/149940057-da7f8185-1565-4308-b557-486fec54d54d.png)

NodeGraph will show the other node`s graph(which has the same node name):
![image](https://user-images.githubusercontent.com/48784001/149940750-f88cc7f3-b31d-4588-9d07-5a0c6155df48.png)
## Changes
Add additional labels in configuration, which is a `[]map[string]string`:

``` yaml
prometheus:
  auth:
    ca_file: ''
    insecure_skip_verify: false
    password: ''
    token: ''
    type: none
    use_kiali_token: false
    username: ''
  cache_duration: 10
  cache_enabled: true
  cache_expiration: 300
  custom_headers: {}
  health_check_url: ''
  additional_label:
    - key1: value1
    - key2: value2
  is_core: true
  thanos_proxy:
    enabled: false
    retention_period: 7d
[...]

PR #4739, opened 2022-02-16
Add QueryScope to narrow promethues queries against unified metrics

jshaughn:
@Xunzhuo , This PR has suggested changes for your 'custom_labels' PR, #4664.   Please review, also @jmazzitelli please review.

I wanted to submit this PR as a sub-PR against your branch, but I was unable to find your kiali fork in the base repository list.   I am creating it against master, as I think it works that way as well.

- change CustomLabels to QueryScope to be more specific about what this setting is about.
- slightly simplify metrics-related code
- centralize the scope handling in graph code to the promQuery generator
  - this is safer because you don't need to add it to each individual query string
  - same approach applied for appender queries, which were not yet covered in the original PR
- add support to custom dashboards
- add required tests
- minor - fix alphabetization for PrometheusConfig fields

Also, I created https://github.com/kiali/kiali-operator/pull/494 for the required operator support for the new config.

*** Requires Operator PR: https://github.com/kiali/kiali-operator/pull/494


graph/telemetry/istio/util/util.go
```diff
@@ -122,3 +122,20 @@ func IsBadDestTelemetry(cluster string, clusterOK bool, svcNs, svc, svcName, wl
 	}
 	return false
 }
+
+// GetQueryScope returns "" if there is no configured queryScope, otherwise queryScope with/without a trailing comma
+func GetQueryScope(prefix, suffix string) string {
```
Xunzhuo, line 127: How about adding a unit test for this method in `util_test.go`?

Xunzhuo: BTW, can we pass the query to this method instead of pass prefix and suffix to hide some details? 

If we do not read the codes, it is probably hard to understand the meaning of prefix and suffix. And these two params usually do not change.

like func GetQueryWithScope(query string) string{}, when using this method, just like 
``` golang
query = util.GetQueryWithScope(query)
```

Xunzhuo (commented):
Thanks @jshaughn for helping me complete this PR, some suggestions I comment.:)

graph/telemetry/istio/util/util.go
jshaughn: good idea, I will push a change.

jshaughn:
@Xunzhuo I made the requested changes, please re-review.  By the way, do you know why your Kiali fork doesn't show up in the list of forks?  This is why I was unable to apply this PR to your PR, and instead targeted master.

business/dashboards.go
```diff
@@ -383,6 +383,10 @@ func (in *DashboardsService) buildLabelsQueryString(namespace string, labelsFilt
 	for k, v := range labelsFilters {
 		labels += fmt.Sprintf(`,%s="%s"`, prometheus.SanitizeLabelName(k), v)
 	}
+	for labelName, labelValue := range in.promConfig.QueryScope {
+		labels += fmt.Sprintf(`,%s="%s"`, labelName, labelValue)
```
jmazzitelli, line 387: What is this `prometheus.SanitizeLabelName(k)` I see above in line 384 - and do you need to sanitize the `labelName` here as well? i.e. should this be:
```suggestion
		labels += fmt.Sprintf(`,%s="%s"`, prometheus.SanitizeLabelName(labelName), labelValue)
```

graph/telemetry/istio/util/util.go
```diff
@@ -122,3 +123,19 @@ func IsBadDestTelemetry(cluster string, clusterOK bool, svcNs, svc, svcName, wl
...
+// it returns the query with the queryScope injected after each occurrence of a leading '{'.
+func AddQueryScope(query string) string {
+	queryScope := config.Get().ExternalServices.Prometheus.QueryScope
+	if len(queryScope) == 0 {
+		return query
+	}
+
+	scope := "{"
+	for labelName, labelValue := range queryScope {
+		scope = fmt.Sprintf("%s%s=\"%s\",", scope, labelName, labelValue)
```
jmazzitelli, line 137: see my earlier comment - do we need to sanitize `labelName` ?

jmazzitelli (commented):
I'm not requesting changes because I don't know if we need this to change - but see my comments. It looks like there needs to be some "sanitation" going on here.

business/dashboards.go
jshaughn: I saw that as well but I don't think it's applicable/needed.  The labels used here must already be valid, as they are custom and being added in the target environment to the scraped metrics.  That sanitization was needed because of the scenario in https://github.com/kiali/kiali/issues/3737.

jmazzitelli: Anytime I see the word "sanitized" in the context of software, I immediately think it is trying to prevent a security loophole (like [sanitizing input to avoid a log4j CVE](https://www.whitesourcesoftware.com/resources/blog/best-practices-for-dealing-with-log4j/) :) ) 

Are we sure this isn't needed to sanitize the labels for something security related? See the code introduced in the scenario / PR you refer to: https://github.com/kiali/kiali/pull/3739/files#diff-a21497378081b5b2b2cf631b8cf327bd5f7905b48e3fdb34518e1503ef7a6713R309-R310
```
// SanitizeLabelName replaces anything that doesn't match invalidLabelCharRE with an underscore.
// Copied from https://github.com/prometheus/prometheus/blob/df80dc4d3970121f2f76cba79050983ffb3cdbb0/util/strutil/strconv.go
func SanitizeLabelName(name string) string {
	return invalidLabelCharRE.ReplaceAllString(name, "_")
}
```

jshaughn: Fine, I'll add it, although the chance of malicious behavior here seems very limited.

jmazzitelli (dismissed):

Xunzhuo:
@jshaughn can you see my fork repo: https://github.com/Xunzhuo/kiali ?

Xunzhuo (approved):
LGTM! Thanks @jshaughn

jshaughn:
@mattmahoneyrh This feature has been tested by @Xunzhuo, so we have two choices, we can perform regression tests (now or later) or we can also perform a feature test.  To test the feature we will need to add custom labels when scraping metrics, so let me know how you want to proceed.

nrfox (approved):
Tested:
- Adding custom metric: `"destination_port": "9080"`
- Adding `destination_port: "9080"` to query_scope and verifying graph is limited
- Adding `destination_canonical_revision: v1` to query_scope and verifying graph is further limited
- Updated the Kiali CR custom dashboard settings with `query_scope` and successfully reconciled
- Rest of the kiali tutorial

jshaughn:
This PR has been dev and dev-qe reviewed and approved.  I had to resolve a conflict in a test and approvals were reset.  Tests are passing, I am merging...

jmazzitelli (approved):

nrfox (approved):

Issue #2299, opened 2020-02-24
List Filters to use dynamic filtering


mtho11:
It would give a much more dynamic feeling with immediate feedback if we could filter as we type. The results in the lists immediately reflect whatever is in the filter box without having to hit return.

<img width="1421" alt="dynamic-filters" src="https://user-images.githubusercontent.com/1312165/53893731-9f0bfd80-3fe3-11e9-9678-7df829f5086d.png">

PR #4740, opened 2022-02-17
Help messages for some key fields in Istio config details method

leandroberetta:
Work in progress.

Required by [#2299](https://github.com/kiali/kiali-ui/pull/2299).

When fetching the Istio config details for some config object, a new parameter can be configured to get help for some key object fields:

`http://localhost:20001/api/namespaces/bookinfo/istio/virtualservices/bookinfo?validate=true&help=true`

Then, this information can be displayed alongside the YAML to help the users to understand some important fields.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -720,22 +720,6 @@ func NewConfig() (c *Config) {
 	return
 }
 
-func (conf *Config) AddHelpDefault() {
-	istioConfigHelpMessages := map[string][]IstioConfigHelp{
-		"virtualservices": {
-			{ObjectField: "spec.hosts", Message: "The destination hosts to which traffic is being sent. Could be a DNS name with wildcard prefix or an IP address. Depending on the platform, short-names can also be used instead of a FQDN (i.e. has no dots in the name)."},
-			{ObjectField: "spec.gateways", Message: "The names of gateways and sidecars that should apply these routes. Gateways in other namespaces may be referred to by <gateway namespace>/<gateway name>; specifying a gateway with no namespace qualifier is the same as specifying the VirtualService’s namespace. To apply the rules to both gateways and sidecars, specify mesh as one of the gateway names."},
-			{ObjectField: "spec.http", Message: "An ordered list of route rules for HTTP traffic."},
-			{ObjectField: "spec.exportTo", Message: "A list of namespaces to which this virtual service is exported. Exporting a virtual service allows it to be used by sidecars and gateways defined in other namespaces."},
-			{ObjectField: "spec.http.match", Message: "Match conditions to be satisfied for the rule to be activated."},
-			{ObjectField: "spec.http.route", Message: "A HTTP rule can either redirect or forward (default) traffic. The forwarding target can be one of several versions of a service. Weights associated with the service version determine the proportion of traffic it receives."},
-			{ObjectField: "spec.http.route.destination.host", Message: "The name of a service from the service registry. Service names are looked up from the platform’s service registry (e.g., Kubernetes services, Consul services, etc.) and from the hosts declared by ServiceEntry."},
-			{ObjectField: "spec.http.route.destination.subset", Message: "The name of a subset within the service. Applicable only to services within the mesh. The subset must be defined in a corresponding DestinationRule."},
-		},
-	}
-	conf.IstioConfigHelpMessages = istioConfigHelpMessages
-}
-
 // AddHealthDefault Configuration
 func (conf *Config) AddHealthDefault() {
 	// Health default configuration
--- a/config/config.go
+++ b/config/config.go
@@ -244,12 +244,6 @@ type IstioLabels struct {
 	VersionLabelName   string `yaml:"version_label_name,omitempty" json:"versionLabelName"`
 }
 
-// IstioConfigHelp represents a help message for a given Istio object type and field
-type IstioConfigHelp struct {
-	ObjectField string `yaml:"object_field" json:"objectField"`
-	Message     string `yaml:"message" json:"message"`
-}
-
 // AdditionalDisplayItem holds some display-related configuration, like which annotations are to be displayed
 type AdditionalDisplayItem struct {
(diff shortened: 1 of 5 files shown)


leandroberetta:
Basically I had the choice to keep this information on the frontend, but I decided to added on the backend, seems better to me although for short messages, that is this case because we are using tooltips so better to put concise information on them.

Another comment is that is hard coded, I was thinking on save the messages on files but doesn't make sense to me as they are short messages. If in the future we decided to render more help information this strategy will need to evolve (using a file database?).

models/istio_config.go
```diff
@@ -66,3 +67,10 @@ type ResourcesPermissions map[string]*ResourcePermissions
 
 // IstioConfigPermissions holds a map of ResourcesPermissions per namespace
 type IstioConfigPermissions map[string]*ResourcesPermissions
+
+var IstioConfigHelpMessages = map[string][]IstioConfigHelp{
+	"virtualservices": {
+		{ObjectField: "spec.hosts", Message: "The destination hosts to which traffic is being sent. Could be a DNS name with wildcard prefix or an IP address. Depending on the platform, short-names can also be used instead of a FQDN (i.e. has no dots in the name)"},
```
lucasponce, line 73: Early comment.
I think these messages could come from the regular config.
Typically default hard-coded values may work just fine, but it could be nice if the user may overwrite them when needed.
(It won't change the code, just perhaps moving this to the config and organize it there, that's all)

leandroberetta: I updated the code with this in mind, sounds ok for the kiali administrators to customize the help they want to give to the users. Thanks @lucasponce .

config/config.go
```diff
@@ -478,6 +484,7 @@ type Config struct {
 	Identity                 security.Identity                   `yaml:",omitempty"`
 	InCluster                bool                                `yaml:"in_cluster,omitempty"`
 	InstallationTag          string                              `yaml:"installation_tag,omitempty"`
+	IstioConfigHelpMessages  map[string][]IstioConfigHelp        `yaml:"istio_config_help_messages,omitempty" json:"help,omitempty"`
```
jmazzitelli, line 487: I know this is just a draft, but I was looking at this and wondering -- do we really want this in the config/ConfigMap? Is this something that is user-configurable? Or shouldn't this be somewhere outside of the configmap - this does not seem like something that is user-configurable so doesn't seem like it belongs in here.

models/istio_config.go
jmazzitelli: Oh, I see it was changed... I don't know why this would be in the user config. When would a user ever want to overwrite these? Would we even want them to?

config/config.go
leandroberetta: Well, this wasn't the first place for this information, before it was at the models package. 

I guess based on suggestions, it might be configurable by users, as it is right now, is not, I guess we need to modify the operator (does the operator overwrite these fields or leave it if is not something it manages?).

Anyway, yes, this is a draft and I was testing this approach, for me it seems the same, that does not belong here, but I'm open to suggestions and feedback. 

Thanks.

jmazzitelli: yeah, this would require changes to the operator and the example crd/cr. I really don't see a need for this to be user-configurable. extra work for something that will be rarely (if ever) used.

jshaughn: I agree, I would not make this configurable unless there was an actual RFE asking for that ability.  It's not clear to me that it's necessary in the initial feature.

lucasponce: If we don't want to put in the config because of the implications with the operator, that's fine.
But it's unclear to me that this information that would easily grow and change should be hard-coded.
An option can be to use a YAML that can be easily updated.

Most of this info comes from the istio.io site and in the future, the idea is that it could be automatically generated in some way (i.e. a script that generates a map).

Xunzhuo: I`m -1 for including these messages in config too, these messages should be not a user-facing config, and it increases the complexity of current configuration. 
We could use a static json/yaml to record these info and like @lucasponce said, we can use some scripts to keep this file updated in the future.
And I suggest embedding this file directly into binary, which is a new feature in golang 1.16, make it easy to manage these static files:)

lucasponce: +1
Not a blocker, perhaps these definitions will not change and these can be just constants.

models/istio_config.go
```diff
@@ -49,9 +48,42 @@ type IstioConfigDetails struct {
...
+
+// IstioConfigHelp represents a help message for a given Istio object type and field
+type IstioConfigHelp struct {
+	ObjectField string `json:"objectField"`
+	Message     string `json:"message"`
+}
+
+var IstioConfigHelpMessages = map[string][]IstioConfigHelp{
+	"virtualservices": {
+		{ObjectField: "spec.hosts", Message: "The destination hosts to which traffic is being sent. Could be a DNS name with wildcard prefix or an IP address. Depending on the platform, short-names can also be used instead of a FQDN (i.e. has no dots in the name)."},
```
lucasponce, line 65: Is this duplicated with the https://github.com/kiali/kiali/pull/4740/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R724 ?

leandroberetta: Yes, sorry. I removed the old code. Thanks @lucasponce .

lucasponce (approved):
Tested with the new additions:
image
image
image

The combined work of the references and the side-panel is very powerful, thanks to both of you @hhovsepy @leandroberetta

hhovsepy (approved):
Verified alongside with UI PR.

models/istio_config.go
JStickler: "_When would a user ever want to overwrite these? Would we even want them to?_"

Sorry to be late to the party, I just saw the notifications for this today when it closed., but having user help in a separate file is really a best practice for localization/translation.  That way the content can be translated without worrying about introducing a bug in the actual code.

PR #4754, opened 2022-02-23
[hack] install bookinfo - use apply instead of create

jmazzitelli:
This allows you to re-run the script to re-deploy things that may not have installed correctly the first time (which happens to me when I forget to specify the minikube profile that causes the traffic generator to fail installation).


jshaughn (approved):
LOK

PR #4755, opened 2022-02-23
Add labeler appender for find/hide support

jmazzitelli:
UI PR: https://github.com/kiali/kiali-ui/pull/2311

(this required PR https://github.com/kiali/kiali/pull/4787 which has been merged to master)


jmazzitelli:
The current implementation in this PR as of right now has an interesting issue - notice when looking at the versioned-app graph, and showing `label:version=v1` it doesn't show the `reviews-v1` node. This is because that node is considered of type "app" and thus the "version" label is stripped because there are three versions in that app (so we stripped labels with different values since we don't know which one to apply to an app node). See `labeler.go` in the switch case statement for graph.NodeTypeApp where I have the comment:
```
// QUESTION: Is the following the right thing to do?
// GetApp returns a list of workloads that make up the app. Capture the superset of all labels on all workloads.
// However, throw away any labels that have different values across the different workloads (e.g. "version", which may be different across workloads).
```
There must be some magic I don't yet know about that creates those "v1", "v2", and "v3" nodes that the finalizer doesn't see.

image

jmazzitelli:
looking at bookinfo + istio-system namespaces with show expression of `label:kiali-test or label:release`:
image

When I use `label:release=prometheus` I see:
image

jmazzitelli:
this is ready for review (it works if you build it with the related UI PR). The only thing left is to update the api-test unit test so that passes.

jshaughn (changes requested):
You need to add the new query param to Doc.go (analogous to what is in there for the `appenders` param.  We will need a test class for the labeler finalizer.

graph/telemetry/istio/appender/labeler.go
```diff
@@ -0,0 +1,118 @@
...
+					log.Debugf("Failed to obtain versioned-app details for [%+v]", n)
+				}
+			} else {
+				// getApp returns a list of workloads that make up the app. Capture the superset of all labels on all workloads.
+				// However, throw away any labels that have different values across the different workloads.
+				labelsMetadata = graph.LabelsMetadata{}
+				if app, ok := getApp(n.Namespace, n.App, gi); ok {
+					differentLabels := map[string]bool{}
+					for _, w := range app.Workloads {
+						if r, ok := getWorkload(n.Namespace, w.WorkloadName, gi); ok {
```
jshaughn, line 67: Do you need to  `getWorkload`, is `w` not already the workload?

jshaughn (commented):
Looks good, just one question.

graph/telemetry/istio/appender/labeler.go
jmazzitelli: `w` is a `WorkloadItem` and only contains minimal data:
image

jmazzitelli: after calling that getWorkload(), this is the result (notice more data comes back in this, which is a WorkloadListItem:
image

jshaughn (changes requested):
Looks good, just need a test class for the new appender.

graph/telemetry/istio/appender/appender.go
```diff
@@ -316,3 +340,36 @@ func getAppWorkloads(namespace, app, version string, gi *graph.AppenderGlobalInf
...
+			return app, true
+		} else {
+			namespaceApps = existingNamespaceApps
+		}
+	} else {
+		namespaceApps = appsMap{}
+		allAppsMap[namespace] = namespaceApps
+	}
+
+	if app, err := gi.Business.App.GetApp(context.TODO(), namespace, appName); err == nil {
```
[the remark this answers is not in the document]
jshaughn: A long time ago I remember adding GetServiceDefinition() support, specifically to make it more efficient to just pull a lighter subset of information, perhaps we need something analogous.  Or maybe since then there have been other changes and what I did for GetServiceDefinition() is now dated.  Just mentioning this so that we come up with something consistent.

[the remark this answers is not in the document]
jmazzitelli: For sure, all we need here is the labels. If we can come up with a more efficient "GetAppDefinition" or whatever, we should do that now.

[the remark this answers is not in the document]
lucasponce: This part was under review, and there were N "GetServiceXYZ" methods that don't have sense.
This code is being consolidated with two main "GetAppList, GetApp" but passing criteria object to populate the result with the info you need (just the definition, the health, validations, runtimes, etc).


Also, I was thinking more to provide these labels in the first pass of the GetApp (or GetAppList, I miss some context, but I haven't studied the changes in the whole context).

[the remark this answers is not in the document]
jmazzitelli: > Also, I was thinking more to provide these labels in the first pass of the GetApp

We won't know that information at the time the call is made. The true graph filtering is done client side via the cytoscape API.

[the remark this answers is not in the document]
jmazzitelli: > If you agree I can send a PR to your work with the changes.

Yes. let's use whatever APIs you suggest to keep in sync with the latest business API.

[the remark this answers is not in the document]
lucasponce: @jmazzitelli Perhaps in the graph/telemetry/istio/appender/appender.go we could add a "getAppList" and then the "getApp" would just iterate over the cached list (similar as it does in the "getWorkloadList" and "getServiceList" in the same file).

The "business.GetAppList" would need criteria, but it's something that is also proposed https://github.com/kiali/kiali/pull/4763/files#diff-80818359e8256dc52cb343d2aa947665c308d27ffb25b1930707e115ded5061fR27 as well.

I think that way would be more efficient, than iterating over the GetApp (the GetApp would need in any case criteria as by default it fetches additional things that the graph doesn't need, as the custom dashboards).

Perhaps as there are two PR that may use this AppCriteria, I can add in master a PR with the GetApp criteria, and both branches rebase it (just thinking how to minimize conflicts here).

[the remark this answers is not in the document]
jshaughn: As an aside, I have no problem if the `GetServiceDefinition` logic is refactored to be criteria-based, it has served its purpose well for years, but can certainly be changed for consistency.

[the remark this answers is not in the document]
jmazzitelli: @lucasponce are you going to do that refactoring in another PR? If so, we can wait for that PR to go in, then we rebase this PR on that work after it is done.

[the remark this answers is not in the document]
lucasponce: @jmazzitelli yes, I'll prepare one minor refactoring for that. I hope that will help to avoid conflicts with several parallel PRs.

jmazzitelli:
To test: just build and deploy normally. Install bookinfo demo with the traffic generator installed (e.g. `hack/istio/install-bookinfo-demo.sh -tg -c kubectl`) and in the "find" text field enter `label:kiali-test or label:release` and see what gets highlighted. Next use `label:kiali-test=traffic-generator` to show that specifying a label value also works. Try the same expressions in the "hide" text field to see that that works also.

graph/telemetry/istio/appender/appender.go
```diff
@@ -180,13 +186,29 @@ func ParseAppenders(o graph.TelemetryOptions) []graph.Appender {
...
+	finalizers = append(finalizers, &TrafficGeneratorAppender{})
+
+	return appenders, finalizers
 }
 
 const (
 	serviceListKey       = "serviceListKey"       // global vendor info map[namespace]serviceDefinitionList
 	serviceEntryHostsKey = "serviceEntryHostsKey" // global vendor info service entries for all accessible namespaces
 	workloadListKey      = "workloadListKey"      // global vendor info map[namespace]workloadListKey
+	appsMapKey           = "appsMapKey"           // global vendor info map[namespace]appsMap
```
jshaughn, line 211: alphabetize these constants. not a blocker.

graph/telemetry/istio/appender/appender.go
```diff
@@ -316,3 +340,40 @@ func getAppWorkloads(namespace, app, version string, gi *graph.AppenderGlobalInf
 	}
 	return result
 }
+
+func getApp(namespace, appName string, gi *graph.AppenderGlobalInfo) (*models.AppListItem, bool) {
```
jshaughn, line 344: Any reason you didn't model this like the workloads with a `getApp` and `getAppList`?  Not a blocker.

jshaughn (commented):
It ran fine, code looks good to me.  A couple of non-blocking comments.  Did you mean to add some test code for the new labeler appender?

graph/telemetry/istio/appender/appender.go
jmazzitelli: yeeeeeees, teacherrrrrr.
done.

graph/telemetry/istio/appender/appender.go
jmazzitelli: wasn't needed so opted not to add more API just to add it. I just added the API that was needed. If anyone needs a getAppList in the future, that can be implemented.

jmazzitelli:
> Did you mean to add some test code for the new labeler appender?

I forgot. I am ashamed. The amount of test code that would need to be written is going to be substantial - anyone who groks all that test/mock code for the other appenders is welcome to contribute tests for this appender ;)

jshaughn:
> I forgot. I am ashamed.

That is all I wanted to hear :-)  Let me look, maybe I can come up with something that's not too terrible.

graph/telemetry/istio/appender/appender.go
```diff
@@ -316,3 +340,40 @@ func getAppWorkloads(namespace, app, version string, gi *graph.AppenderGlobalInf
...
+			return app, true
+		} else {
+			namespaceApps = existingNamespaceApps
+		}
+	} else {
+		namespaceApps = appsMap{}
+		allAppsMap[namespace] = namespaceApps
+	}
+
+	if appList, err := gi.Business.App.GetAppList(context.TODO(), business.AppCriteria{Namespace: namespace, IncludeIstioResources: false}); err == nil {
```
lucasponce, line 369: This is good.
@aljesusg is also adding a "criteria" here https://github.com/kiali/kiali/pull/4763/files#diff-80818359e8256dc52cb343d2aa947665c308d27ffb25b1930707e115ded5061fR27 that should have also an on/off flag to not include it in the graph.

But with this object then adjusting what that method may return would be easier.

jshaughn (approved):
WIth the addition of the test code I gave you :-), this looks good to me.  I am approving and moving to QE.    @lucasponce , can I suggest you take this as Dev-QE?

lucasponce:
> @lucasponce , can I suggest you take this as Dev-QE?

+1

lucasponce (approved):
LGTM
Tested together with UI PR https://github.com/kiali/kiali-ui/pull/2311.

Main feedback about the test will be attached to the UI PR.

aljesusg (approved):
LGFM

graph/telemetry/istio/appender/appender.go
aljesusg: If health is not true then api not return the health info

jshaughn: The labeler appender does not need health.

Issue #4766, opened 2022-03-01
Istio Config List - Configuration icon load inconsistency


hhovsepy:
Happens on Kiali master, found by @nrfox 
Each refresh of Istio Config list shows different configuration validation.

![config_health](https://user-images.githubusercontent.com/604313/156162592-81c6e35d-ea2e-463b-977d-123ecdc86a1d.gif)

PR #4768, opened 2022-03-01
For Istio Config validations result JSON, make the key 'name.namespace' instead of 'name' only to support multiple namepaces

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4766

UI changes are required: https://github.com/kiali/kiali-ui/pull/2315

Before, it was generating validation result JSON with config  'type' and 'name' in a key. So for multiple namespaces having similar config names were causing overriding the key in map, thus confusion is UI.

Now, it uses config's 'name.namespace' as a key, to make sure multiple namespaces are supported.

For QE:
Create 'bookinfo', 'bookinfo2' and 'bookinfo3' namespaces.
Make sure that several configs in 'bookinfo' only is broken, so contains warnings and errors.
List all istio configs for all namespaces.
Refresh the content several times.
Verify that all configs configuration statuses are consistent within all refreshes.

### DIFF
--- a/models/istio_validation.go
+++ b/models/istio_validation.go
@@ -469,7 +469,7 @@ func (iv IstioValidations) MarshalJSON() ([]byte, error) {
 		if !ok {
 			out[k.ObjectType] = make(map[string]*IstioValidation)
 		}
-		out[k.ObjectType][k.Name] = v
+		out[k.ObjectType][k.Name+"."+k.Namespace] = v
 	}
 	return json.Marshal(out)
 }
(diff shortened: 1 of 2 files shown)


Xunzhuo (approved):
It sounds reasonable and works fine in multi-namespaces area

nrfox (approved):
:+1: 

Tested by:
deploying bookinfo, bookinfo2, bookinfo3
breaking the istio config in bookinfo causing an error to appear
refreshing the istio config list page and confirming that results are consistent

pbajjuri20:
Deployed the Bookinfo, Bookinfo2, Bookinfo3 and istio config with some errors and warnings.
refreshed the "Istio Config" Page. it don't have any issues

pbajjuri20 (approved):

PR #4770, opened 2022-03-01
Feature/go 1.17.7 update

sukrit007wawa:
**Describe the change**

This updates the GO version to 1.17.7 that includes security fixes

**Issue reference**

Reported to kiali-security@googlegroups.com

**Backwards incompatible?**

- Changes are backward compatible

**Documentation**

It does not require README updates as it is a patch version change for golang.

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.17.2
+          go-version: 1.17.7
 
       - name: Check out code into the Go module directory
         uses: actions/checkout@v2
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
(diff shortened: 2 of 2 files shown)


lucasponce (commented):
I don't see a problem bumping to these versions.
Any comment @jshaughn? 
I remember there was a discussion related Golang versions, but probably the z releases are safe?

jshaughn:
@lucasponce We can bump this upstream but I'm not sure if the downstream build env can handle it.  Does anyone know if this go version is avail for the cpaas builds?  @aljesusg @leandroberetta @jmazzitelli ?

lucasponce:
> I'm not sure if the downstream build env can handle it

Well, I think that upstream/downstream should have different settings, not sure if a bump like this should be conditioned from a downstream constraint.

Perhaps it's a good moment to introduce some kind of "build profiles" that can keep separated both?

jshaughn:
I agree the upstream should not be affected by downstream constraints.  I'm in favor of the bump, I brought it up mainly to get clarity on whether there is an issue.

jmazzitelli:
shouldn't be an issue - especially since no code changes are being done here, so it should be compatible even if something else wants to use another 1.17.x version.

lucasponce (approved):
I'm using 1.17.7 in my local env for some time:
image

And I've built multiple PRs, so, I haven't seen any problem.

P.S.: Clean well the old dependencies with the golang.

Issue #4611, opened 2021-12-17
Refactor API services to consolidate Health logic into the backend


lucasponce:
Kiali API splits services in several endpoints that are invoked by the UI to consolidate the final view:

![image](https://user-images.githubusercontent.com/1662329/146525208-1d02d657-0a21-41f9-abae-94c08b44a251.png)

This design was initially chosen during the first steps of the project and the default action was to create a new API and let the UI integrate them.

For services like Health, this pattern presents a well-known technical debt as the logic is split and the frontend code is not only rendering the data but doing business calculations.

This design "works" but in the context of scalability scenarios, it presents some performance challenges as discussed in https://github.com/kiali/kiali/discussions/4507 that should be addressed not only for a specific area but for the whole design.

The goal of this Epic is to study and propose a plan to move the Health logic (and potentially other similar endpoints) into the backend to have a robust endpoint that simplifies the UI code.

This work should take into consideration the different efforts in work in progress to adjust the backend endpoint to improve the logic on scalability scenarios.

Suggested test plan (QE @mattmahoneyrh):
- This Epic shouldn't introduce changes in the business features, so the main QE strategy should be to check feature regressions.
- Additionally QE may validate potential performance variations compared with the previous versions.
- This Epic may involve different subtasks, QE/Dev can define if regression tests may happen at the end of the Epic or a specific PR that may require a significant review.

Dev tasks:

* [x] Reduce the number of requests to fetch health data on the graph page
  * https://github.com/kiali/kiali/pull/4772
  * https://github.com/kiali/kiali-ui/pull/2316
  * https://github.com/kiali/kiali/pull/4816
* [ ] Reduce the number of requests to fetch health data on the overview page
* [x] https://github.com/kiali/kiali/issues/4748
* [x] #4790
  * https://github.com/kiali/kiali/pull/4799
  * https://github.com/kiali/kiali-ui/pull/2329
* [ ] Health metrics from graph data (remove the "perceived" duplicated fetches from prometheus)
  * https://github.com/kiali/kiali/pull/4802
  * https://github.com/kiali/kiali-ui/pull/2330

israel-hdez:
I've been reviewing the back-end code involved to Health calculations. I still haven't checked the UI side. I've been trying to understand what we _really_ need, given what is written in this issue and also given the backing discussion https://github.com/kiali/kiali/discussions/4507.

So far, I understand this is the background:
[...]

PR #4772, opened 2022-03-02
Add health data into the GraphNamespaces endpoint

israel-hdez:
A new `includeHealth` parameter is added to the graph endpoint which is `false` by default. If true, data to calculate Health (request rates and workload status) will be appended to each node in the generated cytoscape JSON.

UI side PR: https://github.com/kiali/kiali-ui/pull/2316
Related #4611


graph/api/api.go
```diff
@@ -14,7 +16,7 @@ import (
 )
 
 // GraphNamespaces generates a namespaces graph using the provided options
-func GraphNamespaces(business *business.Layer, o graph.Options) (code int, config interface{}) {
+func GraphNamespaces(business *business.Layer, o graph.Options, ctx context.Context) (code int, config interface{}) {
```
nrfox, line 19: Conventionally context is passed as the first argument

graph/api/api.go
```diff
@@ -93,11 +95,107 @@ func generateGraph(trafficMap graph.TrafficMap, o graph.Options) (int, interface
...
+			case "app":
+				health, err := bs.Health.GetNamespaceAppHealth(ctx, namespace, o.RawDuration, time.Unix(o.QueryTime, 0))
+				if err != nil {
+					return err
+				}
+				for _, nodeWrap := range nodeWrappers {
+					if h, ok := health[nodeWrap.Data.App]; ok {
+						nodeWrap.Data.HealthData = h
+					} else {
+						nodeWrap.Data.HealthData = []int{}
```
nrfox, line 166: Why `[]int{}`?

graph/api/api_test.go
```diff
@@ -3284,13 +3284,13 @@ func TestComplexGraph(t *testing.T) {
...
 
-	var fut func(b *business.Layer, p *prometheus.Client, o graph.Options) (int, interface{})
+	var fut func(b *business.Layer, p *prometheus.Client, o graph.Options, ctx context.Context) (int, interface{})
 
 	mr := mux.NewRouter()
 	mr.HandleFunc("/api/namespaces/graph", http.HandlerFunc(
 		func(w http.ResponseWriter, r *http.Request) {
 			context := context.WithValue(r.Context(), "authInfo", &api.AuthInfo{Token: "test"})
-			code, config := fut(nil, client, graph.NewOptions(r.WithContext(context)))
+			code, config := fut(nil, client, graph.NewOptions(r.WithContext(context)), context)
```
nrfox, line 3293: Is there an easy way to test with some non-null health data?

nrfox (commented):
Tested functionality and it works great. Reduces many health requests on the graph page with large numbers of namespaces.

graph/options.go
```diff
@@ -68,10 +68,12 @@ type NodeOptions struct {
 
 // CommonOptions are those supplied to Telemetry and Config Vendors
 type CommonOptions struct {
-	Duration  time.Duration
-	GraphType string
-	Params    url.Values // make available the raw query params for vendor-specific handling
-	QueryTime int64      // unix time in seconds
+	RawDuration   string
+	Duration      time.Duration
+	IncludeHealth bool
```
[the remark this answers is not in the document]
israel-hdez: With regards to the Graph components, it is always requested.
However, this is also used in the Overview page for the feature that let you create Traffic policies to lock traffic to the current topology. I thought the Overview page wasn't fetching health, but it does.

So, you are right that this flag is apparently not really needed... I'm not sure if I should remove it :-/

graph/api/api.go
israel-hdez: Fixed.

graph/api/api.go
```diff
@@ -93,11 +95,107 @@ func generateGraph(trafficMap graph.TrafficMap, o graph.Options) (int, interface
...
+			case graph.NodeTypeApp:
+				kind = "app"
+			case graph.NodeTypeBox:
+				if node.Data.IsBox == graph.BoxByApp {
+					kind = "app"
+				}
+			case graph.NodeTypeService:
+				kind = "service"
+			}
+		}
```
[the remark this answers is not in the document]
israel-hdez: Fixed.

graph/api/api_test.go
israel-hdez: I think it is testable.
I would need to pass the `includeHealth` flag set as true and a non-nil business layer mock.

Then, the tedious part would be to mock all calls that are being done to the cluster API and prom :-/
I'll check how bad it is.

graph/api/api.go
nrfox: :+1: I don't think it matters what type it is during serialization if the array is empty

graph/options.go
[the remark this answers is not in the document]
nrfox: imo if it's not being used then it can be removed until it is but it's up to you. Not a blocker.

nrfox (dismissed):

graph/options.go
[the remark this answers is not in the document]
jshaughn: Health is an inherent feature of the graph and so I would vote to simplify the change and remove the unnecessary option.  Even if you want to keep health-generation as optional, I think it should be handled a different way, see my other comment.

graph/api/api.go
```diff
@@ -93,11 +95,108 @@ func generateGraph(trafficMap graph.TrafficMap, o graph.Options) (int, interface
...
 	switch o.ConfigVendor {
 	case graph.VendorCytoscape:
-		vendorConfig = cytoscape.NewConfig(trafficMap, o.ConfigOptions)
+		vc := cytoscape.NewConfig(trafficMap, o.ConfigOptions)
+		if o.ConfigOptions.IncludeHealth {
+			err := collectGraphHealth(ctx, &vc, o.ConfigOptions, bs)
+			if err != nil {
+				graph.Error(fmt.Sprintf("Error collecting graph health: %s", err.Error()))
+			}
+		}
```
jshaughn, line 104: I think this should be done in a graph finalizer appender.

graph/config/cytoscape/cytoscape.go
```diff
@@ -105,6 +105,7 @@ type NodeData struct {
 	HasTrafficShifting    bool                `json:"hasTrafficShifting,omitempty"`    // true (vs has traffic shifting) | false
 	HasVS                 *VSInfo             `json:"hasVS,omitempty"`                 // it can be empty if there is a VS without hostnames
 	HasWorkloadEntry      []graph.WEInfo      `json:"hasWorkloadEntry,omitempty"`      // static workload entry information | empty if there are no workload entries
+	HealthData            interface{}         `json:"healthData"`                      // data to calculate health status from configurations
```
jshaughn, line 108: Is there a reason we don't describe this datatype in more detail?

israel-hdez: I had to port some typescript code to go, so I was influenced by typescript typing system. I wanted this to be a union type and the closest that go offers is the generic `interface{}`. To remove the generic type I would need to create a struct with three fields: one field for each possible allowed type: workloadHealth, serviceHealth or appHealth. Only one field would be populated on an instance of the struct.

graph/api/api.go
jshaughn: If you go with the other comments and always supply health for graph components, removing the `includeHealth` option, then perhaps you can just use omitempty when there is nothing to report (or maybe even you can then assume healthy). 
Otherise the health data would be provided.   Just a thought.

graph/options.go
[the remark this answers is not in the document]
israel-hdez: OK, let's do that in another PR.
The `includeHealth` was a flag present only on UI side is supposed to be used to choose to fetch (or not fetch) health data by doing calls to health endpoints. Here I am simply maintaining that behavior.

But as I already commented, I didn't find a case where that flag is actually turned off. Removing that flag would imply removing conditionals on UI side, so I would like to do it as separate work.

graph/api/api.go
```diff
@@ -35,20 +37,20 @@ func GraphNamespaces(business *business.Layer, o graph.Options) (code int, confi
...
-func graphNamespacesIstio(business *business.Layer, prom *prometheus.Client, o graph.Options) (code int, config interface{}) {
+func graphNamespacesIstio(ctx context.Context, business *business.Layer, prom *prometheus.Client, o graph.Options) (code int, config interface{}) {
 
 	// Create a 'global' object to store the business. Global only to the request.
 	globalInfo := graph.NewAppenderGlobalInfo()
 	globalInfo.Business = business
 
 	trafficMap := istio.BuildNamespacesTrafficMap(o.TelemetryOptions, prom, globalInfo)
-	code, config = generateGraph(trafficMap, o)
+	code, config = generateGraph(ctx, trafficMap, o, business)
```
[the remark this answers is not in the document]
israel-hdez: OK, I'll see if I can move all this to an appender or finalizer.

graph/config/cytoscape/cytoscape.go
israel-hdez: Actually, there is also a fourth type which I also mentioned in the comment that I already pointed in your other comment: https://github.com/kiali/kiali/pull/4772#discussion_r822926789.

graph/api/api.go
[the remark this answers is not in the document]
jshaughn: Feel free to ping me if you want to talk about this.   We have 3 "finalizer" appenders at the moment.  Two of them always run (not optional) and this could be like that as well.  The third is the one just added recently, called "labeler", and it runs optionally.  So if you wanted to keep the health stuff optional you could do that easily, it's just another appender, not a new option.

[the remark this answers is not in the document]
israel-hdez: I was looking at it and this seems to need the very final graph.

It turns out that boxes can have health. Specifically, the "app boxes", like the "reviews" in the bookinfo sample app:

image

Healt is calculated differently: for the inner nodes, workload health is fetched (despite they are "app" nodes) and for the enclosing app box, app health is fetched. Regardless about how health would be calculated, I do need a way to carry metadata about app boxes in the traffic map to be able to attach health data to them. That probably would imply moving around or replicating the logic for resolving app compound boxes from the `generateGraph` function to the `BuildTrafficMap` functions :confused:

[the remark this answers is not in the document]
jshaughn: I understand what you are saying.  Let me take a look at the code and see if there is something that makes sense.  Perhaps the app box health can be inherited from the children, like some other app box information.   I really would prefer to keep the config generation from having to perform additional graph generation tasks.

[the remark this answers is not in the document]
jshaughn: Thinking a bit, there may be ways this can be done in a standard health finalizer appender.  What do you think about this approach:  When you encounter an app node in the TrafficMap, query for the app health.  Store it on the app node in the metadata.  For an "App graph", this would be the same thing we do already.  But for a "Versioned App" graph, we'd store the workload health AND the app health.  We'd cache the app health in the appender code, so we could assign it to every versioned-app node for that app.  Then, for an app box node, the config generator could just inherit the app health from any one of the child app nodes.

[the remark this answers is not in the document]
israel-hdez: OK, I think that's doable.

With that in mind, it is probably better to keep the `includeHealth` flag (which would become an optional appender). Otherwise, it is going to be a lot of work to fix all tests in `api_test.go`. Making health mandatory would mean that mocks are needed on almost all tests on that file :-/

israel-hdez:
After fixing a test, I think this is ready to go to QA.

lucasponce:
> After fixing a test, I think this is ready to go to QA.

@jshaughn @nrfox and @aljesusg and I will take the role to deep review this work as commented in the team meetings.

israel-hdez:
> @jshaughn @nrfox and @aljesusg and I will take the role to deep review this work as commented in the team meetings.

@pbajjuri20 asked me to assign him this PR to do testing
Not sure if he will have time, but I think he is interested :-)

jshaughn (approved):
Good work!

graph/telemetry/istio/appender/health_test.go
```diff
@@ -148,6 +151,10 @@ func setupHealthConfig(services []core_v1.Service, deployments []apps_v1.Deploym
 	business.SetKialiControlPlaneCluster(&business.Cluster{
 		Name: business.DefaultClusterID,
 	})
-	businessLayer := business.NewWithBackends(k8s, nil, nil)
+
+	prom := new(prometheustest.PromClientMock)
+	prom.MockNamespaceServicesRequestRates("testNamespace", "0s", time.Unix(0, 0), model.Vector{})
+	prom.MockAllRequestRates("testNamespace", "0s", time.Unix(0, 0), model.Vector{})
+	businessLayer := business.NewWithBackends(k8s, prom, nil)
```
nrfox, line 158: Are there any tests for the new functionality of appending health to the graph nodes?

israel-hdez:
> @jshaughn Good work!

Well... Also good work from you ;-)

Issue #4769, opened 2022-03-01
Envoy metrics look broken


lucasponce:
Probably there were some changes that affect the Envoy metrics, but today they look broken:
![image](https://user-images.githubusercontent.com/1662329/156208687-3f29d793-474e-4699-bd82-5c3cae1ad896.png)

gregory-v:
In Workloads, tabs Envoy -> Metrics I have the following error : "Sorry, there was a problem. Try a refresh or navigate to a different page."
I increased logs in container kiali but I am not able to find anything. I have the same behaviour with previous version of kiali (v1.46.0) and Istio (1.12.2)

Kiali UI  1.47.0 (d37fff0dad452ed0bc1b122d4543fc40bd0b8510)
Kiali Server v1.47.0 (31da692118694362f1a04b25f2d661edc5e8c11f)
Kiali Container v1.47.0  (Kiali-operator)
Components Istio 1.13.1  (Istio-operator)
Prometheus 2.29.1    (kube-prometheus)
Kubernetes v1.21.5-eks-bc4871b (EKS)

The "click" doesn't generate any network request. but an JS Error: Minified React.

In my console :
2.5e6844a0.chunk.js:2 Error: Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]= for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
    at ku (2.5e6844a0.chunk.js:2:2109490)
    at 2.5e6844a0.chunk.js:2:2055231
    at ja (2.5e6844a0.chunk.js:2:2065325)
    at gs (2.5e6844a0.chunk.js:2:2105838)
    at lu (2.5e6844a0.chunk.js:2:2096347)
    at su (2.5e6844a0.chunk.js:2:2096272)
    at Zs (2.5e6844a0.chunk.js:2:2093281)
    at 2.5e6844a0.chunk.js:2:2044666
    at t.unstable_runWithPriority (2.5e6844a0.chunk.js:2:2120247)
    at Wi (2.5e6844a0.chunk.js:2:2044375)

It seems istio_tcp* metrics are missing in prom since I am using kube-prometheus. My serviceMonitor config :
https://pastebin.com/mCXar3xb

lucasponce:
Ok, I confirm it's a problem with Istio 1.13.x. In Istio 1.12.1 those metrics are enabled by default:
[...]

PR #4778, opened 2022-03-02
Align namespace label in custom dashboards

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4769

Istio 1.13.x ships Prometheus v2.31.1 (chart 1.15) (vs Istio 1.12/Prometheus v2.26.0 - chart 1.14 -).

Prometheus v2.31.1 has introduced an alignment in the labels of the "kubernetes_pod" job, that affects the "namespace" label used in the CustomDashboards (also used by the Envoy subtab) [1].

[1] https://github.com/prometheus/prometheus/pull/9832

By default, CustomDashboards have the "kubernetes_namespace" label that needs to change to "namespace" in custom queries.

With this change, the custom dashboards work as expected.

This PR is not yet finished, it's just a draft to give visibility, it needs to apply the change into the tests and other places.

I think that the defaults should point to the default version this targets (Istio 1.13.x, also backported to the Kiali 1.47.x), and users can/should adjust this to the proper Prometheus they use.

This work has a pending UI fix to not crash on this case:
https://github.com/kiali/kiali-ui/pull/2317

Also, this will need a proper FAQ entry.

As commented, when this fix is ready, as Kiali 1.47.x targets Istio 1.13.x it would require a backport (I think, this issue looks bad).

How to test:
- Build Kiali backend and frontend PRs and use them in Istio 1.12.x cluster.
- Deploy "runtimes" demo: https://github.com/kiali/demos/tree/master/runtimes-demo
- Navigate to "vertx" App details: 
a) validate that custom dashboards don't show:
image
- Navigate to "vertx-v1" Workload details, validate that there are no custom dashboards:
a) validate that custom dashboard don't show:
image
b) validate that Envoy tab doesn't show "Metrics" subtab:
image

- Build a Kiali backend and frontend PRs and use them in Istio 1.13.x cluster. 
- Deploy "runtimes" demo: https://github.com/kiali/demos/tree/master/runtimes-demo
- Navigate to "vertx" App details: 
a) validate that JVM custom dashboards show:
image
image
- Navigate to "vertx-v1" Workload details, validate that there are JVM custom dashboards:
image
b) validate that Envoy tab show "Metrics" subtab and it works as expected:
image

Two pending topics:
- The PRs from backend/frontend should be backported to 1.47.x for community as 1.47.x is used with Istio 1.13.x and this bug looks bad, IMHO.
- This requires a pending FAQ in the kiali.io.
- This parameter should be updated in the defaults of the operator in 1.48.x if some distribution would use a different version of Prometheus different than the shipped in the Istio add-ons.

Note, these are the Prometheus versions used from the image and the charts:

- Istio 1.12.1:
```
    chart: prometheus-14.6.1
    image: "prom/prometheus:v2.26.0"
``` 
- Istio 1.13.1:
```
    chart: prometheus-15.0.1
    image: "prom/prometheus:v2.31.1"
```

### DIFF
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -37,7 +37,7 @@ func NewDashboardsService(namespace *models.Namespace, workload *models.Workload
 	}
 	nsLabel := cfg.ExternalServices.CustomDashboards.NamespaceLabel
 	if nsLabel == "" {
-		nsLabel = "kubernetes_namespace"
+		nsLabel = "namespace"
 	}
 
 	// Overwrite Custom dashboards defined at Namespace level
--- a/business/dashboards_test.go
+++ b/business/dashboards_test.go
@@ -31,7 +31,7 @@ func TestGetDashboard(t *testing.T) {
 	// Setup mocks
 	service, prom := setupService("my-namespace", []dashboards.MonitoringDashboard{*fakeDashboard("1")})
 
-	expectedLabels := "{kubernetes_namespace=\"my-namespace\",APP=\"my-app\"}"
+	expectedLabels := "{namespace=\"my-namespace\",APP=\"my-app\"}"
 	namespace := models.Namespace{
 		Name: "my-namespace",
 	}
@@ -73,7 +73,7 @@ func TestGetDashboardFromKialiNamespace(t *testing.T) {
 	// Setup mocks
 	service, prom := setupService("my-namespace", []dashboards.MonitoringDashboard{*fakeDashboard("1")})
 
-	expectedLabels := "{kubernetes_namespace=\"my-namespace\",APP=\"my-app\"}"
+	expectedLabels := "{namespace=\"my-namespace\",APP=\"my-app\"}"
 	namespace := models.Namespace{
 		Name: "my-namespace",
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -537,7 +537,7 @@ func NewConfig() (c *Config) {
 				DiscoveryAutoThreshold: 10,
 				Enabled:                true,
 				IsCore:                 false,
-				NamespaceLabel:         "kubernetes_namespace",
+				NamespaceLabel:         "namespace",
 				Prometheus: PrometheusConfig{
 					ThanosProxy: ThanosProxy{
 						Enabled: false,


jmazzitelli:
I haven't looked, but perhaps this is why our metrics-test molecule test has been failing recently on certain platforms. See:

https://github.com/kiali/kiali-operator/search?q=kubernetes_namespace&type=code

jmazzitelli:
Check to see if any of this is affected on OSSM - because I saw problems when testing on OSSM a while ago. See my comment here:

https://github.com/kiali/kiali-operator/blob/f144c1943c62257a47d4009fc9404a8df6437778/molecule/metrics-test/converge.yml#L25-L26

jmazzitelli:
Thanks, @lucasponce !! You just solved a mystery - our `metrics-test` molecule test was failing the past couple weeks and no one knew why. I just confirmed this rename of the prometheus label is the cause (I saw it fail when I installed Istio 1.13.1 on my local minikube - but when I made this one line change to the molecule test, it passes).

So, we need an operator PR to go with this in order to change this prom label from `kubernetes_namespace` to `namespace`. Note that we might need to be careful when deployed with OSSM. This has happened to us in the past regarding this label name (as you can see by [my comment from long ago](https://github.com/kiali/kiali-operator/blob/f144c1943c62257a47d4009fc9404a8df6437778/molecule/metrics-test/converge.yml#L25-L26)).

Here is the change needed in the operator to get the molecule test to pass:

```diff
diff --git a/molecule/metrics-test/converge.yml b/molecule/metrics-test/converge.yml
index d369017..504a552 100644
--- a/molecule/metrics-test/converge.yml
+++ b/molecule/metrics-test/converge.yml
@@ -21,7 +21,7 @@
       test_start_time: "{{ ansible_date_time.iso8601 }}"
 
   - set_fact:
-      prom_namespace_label: kubernetes_namespace
+      prom_namespace_label: namespace
       # This used to be needed for OSSM 2.0 but no longer for 2.1
       #prom_namespace_label: "{{ 'namespace' if is_maistra == True else 'kubernetes_namespace' }}"
```

Note that other changes might be needed because we make mention of this `kubernetes_namespace` label in 2 other places:

https://github.com/kiali/kiali-operator/search?q=kubernetes_namespace&type=code


Notice the [comment in our CRD schema docs](https://github.com/kiali/kiali-operator/blob/v1.47.0/crd-docs/crd/kiali.io_kialis.yaml#L536): `The default is "kubernetes_namespace" but it is quite common to use just "namespace" depending on your Prometheus configuration."`

jmazzitelli (approved):
I ran the test procedures specified in the [description](https://github.com/kiali/kiali/pull/4778#issue-1157406761) (for both Istio 1.12 and 1.13) and the behavior is as expected.

jmazzitelli:
operator PR that fixes the metrics-test and updates the crd doc: https://github.com/kiali/kiali-operator/pull/498

Xunzhuo (commented):
looks good! I ran it in my server and it works fine :)

Issue #4784, opened 2022-03-03
Duration dropdown showing invalid durations


jshaughn:
The duration dropdowns are always showing all options (i.e. up to 30d) despite the retention setting for prometheus.

I noticed this while investigating https://issues.redhat.com/browse/OSSM-999, because I the root issue is the same.   In short, Kiali is using a deprecated Prometheus flag to determine TSDB retention.  Unfortunately the deprecated flag is still being set, but to 0, so the issue was hidden.  In addition to the duration dropdowns, this also negatively impacts the custom datepicker minDate and minTime boundaries.

Working the fix...

PR #4785, opened 2022-03-04
Fix Prometheus TSDB retention config

jshaughn:
Prometheus deprecated "storage.tsdb.retention" many versions back, and
it is no longer valid.  Unfortunately it is still set, but to 0, invalidating
this information in serverConfig.  Switch to "storage.tsdb.retention.time",
the valid setting.  Also, add a check to ensure this setting is never 0.
Furthermore, ensure that if all fails we provide a valid default to
consumers of the config.

Fixes https://github.com/kiali/kiali/issues/4784

### DIFF
--- a/handlers/config.go
+++ b/handlers/config.go
@@ -189,7 +189,7 @@ func getPrometheusConfig() PrometheusConfig {
 				retention, err := model.ParseDuration(retentionString)
 				if checkErr(err, fmt.Sprintf("Invalid storage.tsdb.retention.time [%s]", retentionString)) {
 					if retention == 0 {
-						log.Warningf("%s", "Prometheus storage.tsdb.retention.time configured to 0, ignoring...")
+						log.Warning("Prometheus storage.tsdb.retention.time configured to 0, ignoring...")
 					} else {
 						promConfig.StorageTsdbRetention = int64(time.Duration(retention).Seconds())
 					}
--- a/handlers/config.go
+++ b/handlers/config.go
@@ -17,7 +17,8 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


Xunzhuo (commented):
lgtm :)

handlers/config.go
```diff
@@ -183,10 +185,14 @@ func getPrometheusConfig() PrometheusConfig {
...
 		flags, err := client.GetFlags()
 		if checkErr(err, "Failed to fetch Prometheus flags") {
-			if retentionString, ok := flags["storage.tsdb.retention"]; ok {
+			if retentionString, ok := flags["storage.tsdb.retention.time"]; ok {
 				retention, err := model.ParseDuration(retentionString)
-				if checkErr(err, fmt.Sprintf("Invalid storage.tsdb.retention [%s]", retentionString)) {
-					promConfig.StorageTsdbRetention = int64(time.Duration(retention).Seconds())
+				if checkErr(err, fmt.Sprintf("Invalid storage.tsdb.retention.time [%s]", retentionString)) {
+					if retention == 0 {
+						log.Warningf("%s", "Prometheus storage.tsdb.retention.time configured to 0, ignoring...")
```
jmazzitelli, line 192: just curious - why is this one big "%s" ?? :) Why not just:

```suggestion
						log.Warning("Prometheus storage.tsdb.retention.time configured to 0, ignoring...")
```

jmazzitelli (approved):

israel-hdez (approved):
LG!

aljesusg (approved):
LGFM

PR #4791, opened 2022-03-05
fix: misconfigured deployment name causes panic

### LINKED ISSUE #4788: Misconfigured `istiod_deployment_name` causes a panic
When `istiod_deployment_name` in the kiali CR != actual istiod deployment name the kiali server panics. [Here](https://github.com/kiali/kiali/blob/master/business/services.go#L289) the return value can be nil even if err is nil causing a panic when trying to access `cluster.Name`. The error should be handled gracefully instead of panicing.

Steps to reproduce:
1. Deploy kiali with an alternate `.spec.external_services.istio.istiod_deployment_name` that does not match the actual deployment name (by default `istiod`).

Xunzhuo:

**Describe the change**

misconfigured `istiod_deployment_name` causes a panic, should check if it is nil.

**Issue reference**

Fixes #4788

**Backwards incompatible?**

No


business/services.go
```diff
@@ -287,7 +287,11 @@ func (in *SvcService) getClusterId() string {
 	// Protection on tests
 	if in.businessLayer != nil {
 		if cluster, err := in.businessLayer.Mesh.ResolveKialiControlPlaneCluster(nil); err == nil {
-			clusterId = cluster.Name
+			if cluster != nil {
+				clusterId = cluster.Name
+			} else {
+				log.Debug("No Cluster ID is set in the service mesh control plane configuration. Or please check istio_deployment_name is set properly.")
```
nrfox, line 293:
```suggestion
				log.Debugf("No Cluster ID is set in the service mesh control plane configuration. Or please check 'istiod_deployment_name' is set properly. Using default Cluster ID: %s", clusterId)
```

Xunzhuo: Good suggestion! Thanks @nrfox

lucasponce (approved):

Issue #4640, opened 2022-01-09
KIA0003 for multiple Request Authentication


johgoe:
We have multiple RequestAuthentication which applies to the same workload by the selector.
It's used to extends the jwtRule with additional jwksUris of newley added trusted issuers.

This is supported from istio and working fine. 

It's also mentioned in https://istio.io/latest/docs/concepts/security/#request-authentication

>Istio can combine all matching request authentication policies to work as if they come from a single request authentication policy. Thus, you can have multiple mesh-wide or namespace-wide policies in a mesh or namespace. However, it is still a good practice to avoid having multiple mesh-wide or namespace-wide request authentication policies.

>Request authentication policies can specify more than one JWT if each uses a unique location. When more than one policy matches a workload, Istio combines all rules as if they were specified as a single policy. This behavior is useful to program workloads to accept JWT from different providers.

Kiali always shows an error of type KIA0003, but no error is expected in this case.

jshaughn:
cc @hhovsepy , any thoughts on this one?

hhovsepy:
This is a valid issue, thanks for logging @johgoe .
Prepared a solution https://github.com/kiali/kiali/pull/4794

hhovsepy:
The fix of this issue will be released with Kiali v1.48.

PR #4794, opened 2022-03-09
Skip KIA0003 Multi Match checker for RequestAuthentication.

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4640

From Istio: Request authentication policies can specify more than one JWT if each uses a unique location. When more than one policy matches a workload, Istio combines all rules as if they were specified as a single policy. 

So skipping KIA0003 Multi Match checker for RequestAuthentication.

For QE:
Create 2 RequestAuthentications both having same selector labels pointing to the same workload.
Verify that no error is displayed.
Before:
Screenshot from 2022-03-09 09-10-11

Now:
Screenshot from 2022-03-09 09-10-33

Regression test the same scenario for Sidecars and PeerAuths and see that error KIA0003 is shown for them.

### DIFF
--- a/business/checkers/common/multi_match_selector_checker.go
+++ b/business/checkers/common/multi_match_selector_checker.go
@@ -14,6 +14,7 @@ type GenericMultiMatchChecker struct {
 	Selectors    map[int]map[string]string
 	WorkloadList models.WorkloadList
 	Path         string
+	skipSelSubj  bool
 }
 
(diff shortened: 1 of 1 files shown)


lucasponce (approved):
Before:
image

After with this fix:
image

jshaughn (approved):
Looks good for me as well.

Issue #4692, opened 2022-01-31
Improve crossnamespace Istio Gateways query in ServiceDetailsPage


lucasponce:
As part of the big epic defined at https://github.com/kiali/kiali/issues/4382 the cross-namespace queries have been improved with new flags "AllNamespaces" to use the Istio Registry to fetch all configuration using a single query.

There is a particular case in the ServiceDetailsQuery that iterates over all namespaces to collect all the Istio Gateways present in the cluster. This query is required to populate the Service Wizard to let the user select an existing gateway while creating/updating Istio configuration.

This loop could be reduced with a single query as in scenarios with a large number of namespaces this may introduce performance overhead.

hhovsepy:
@lucasponce can you please add the link of code you found that iteration? cannot find it.

lucasponce:
Sure, it's in the frontend code:
https://github.com/kiali/kiali-ui/blob/master/src/pages/ServiceDetails/ServiceDetailsPage.tsx#L86

There is a first call to fetch all namespaces and then it iterates on them to fetch the gateways per namespace.
With all the work there is in place, I think it can be improved.

Also, on the Overview page, there are similar discussions that may be improved as well, but those would deserve a separate issue as the refactoring could buffer can be larger.

PR #4796, opened 2022-03-11
Added new API endpoint to list all Istio Configs for All Namespaces

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4692

It used to list Gateways per Namespace in Service Details page for Wizzard.

Now there is a new API "/api/istio" for listing Istio Configs for All Namespaces.

This PoC has a weakness:  Listing Gateways for All Namespaces means listing them from Registry, and it returns all Configs, not only Gateways.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -844,10 +844,9 @@ func checkType(types []string, name string) bool {
 	return false
 }
 
-func ParseIstioConfigCriteria(namespace, objects, labelSelector, workloadSelector string) IstioConfigCriteria {
+func ParseIstioConfigCriteria(namespace, objects, labelSelector, workloadSelector string, allNamespaces bool) IstioConfigCriteria {
 	defaultInclude := objects == ""
 	criteria := IstioConfigCriteria{}
-	criteria.Namespace = namespace
 	criteria.IncludeGateways = defaultInclude
 	criteria.IncludeVirtualServices = defaultInclude
 	criteria.IncludeDestinationRules = defaultInclude
@@ -861,6 +860,12 @@ func ParseIstioConfigCriteria(namespace, objects, labelSelector, workloadSelecto
 	criteria.IncludeEnvoyFilters = defaultInclude
 	criteria.LabelSelector = labelSelector
 	criteria.WorkloadSelector = workloadSelector
+
+	if allNamespaces {
+		criteria.AllNamespaces = true
+	} else {
+		criteria.Namespace = namespace
+	}
 
 	if defaultInclude {
 		return criteria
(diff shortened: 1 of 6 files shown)


lucasponce:
Thanks, @hhovsepy, I'm going to fetch locally this one.

I have an early thought, that perhaps this endpoint may require a check that the user can access the "istiod" pod.
Or perhaps it may be overkill, I'll play with it and I'll give you more feedback.

routing/routes.go
```diff
@@ -218,6 +218,26 @@ func NewRoutes() (r *Routes) {
...
+		//     Schemes: http, https
+		//
+		// responses:
+		//      500: internalError
+		//      200: istioConfigList
+		//
+		{
+			"IstioConfigListAll",
+			"GET",
+			"/api/istio",
```
lucasponce, line 237: Perhaps it could be better to narrow the endpoint to "/api/istio/config" as there are other endpoints like "/api/istio/permissions" or "/api/istio/certs".

It's just a naming suggestion to follow a similar convention here.

lucasponce (approved):
I think is a good addition and it will open the design to other optimizations.

lucasponce:
Adding @jshaughn for a quick review/test.
This work doesn't change the behavior in any feature, the check is to validate that when navigating to the service details, the list of gateways comes from this new endpoint instead of iterating per namespace.
There is a capture in the UI about what to expect:
https://github.com/kiali/kiali-ui/pull/2325

business/istio_config_test.go
```diff
@@ -25,68 +25,84 @@ func TestParseListParams(t *testing.T) {
...
 	assert.False(t, criteria.IncludeServiceEntries)
+	assert.False(t, criteria.AllNamespaces)
+
+	criteria = ParseIstioConfigCriteria("", objects, labelSelector, "", true)
+
+	assert.True(t, criteria.IncludeGateways)
+	assert.False(t, criteria.IncludeVirtualServices)
+	assert.False(t, criteria.IncludeDestinationRules)
+	assert.False(t, criteria.IncludeServiceEntries)
+	assert.True(t, criteria.AllNamespaces)
```
jshaughn, line 51: non-blocking: this could test that namespace is not set.

business/istio_config.go
```diff
@@ -844,10 +844,9 @@ func checkType(types []string, name string) bool {
 	return false
 }
 
-func ParseIstioConfigCriteria(namespace, objects, labelSelector, workloadSelector string) IstioConfigCriteria {
+func ParseIstioConfigCriteria(namespace, objects, labelSelector, workloadSelector string, allNamespaces bool) IstioConfigCriteria {
```
jshaughn, line 847: not blocking: I suppose this could have just driven off of `namespace == ""` and avoided having both a `namespace` and `allNamespaces` parameter.

lucasponce: There is a case where there is no namespace but no "all namespaces flag" it should return an empty list.
This flag is to cover that case.

PR #4763, opened 2022-02-28
Add health to endpoints lists

aljesusg:
Add health to endpoints /workloads, /services and /apps.

So adding queryParam health=true Backend return the health for each element, If not the health is empty

# Example

/api/namespaces/bookinfo/workloads?health=true

```json
[...]

PR #4799, opened 2022-03-14
Add health information in details

aljesusg:
Require: https://github.com/kiali/kiali/pull/4763
Related with: https://github.com/kiali/kiali/issues/4611

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -39,7 +39,10 @@ type WorkloadService struct {
 
 type WorkloadCriteria struct {
 	Namespace             string
+	WorkloadName          string
+	WorkloadType          string
 	IncludeIstioResources bool
+	IncludeServices       bool
 	Health                bool
 	RateInterval          string
 	QueryTime             time.Time
@@ -101,6 +104,9 @@ func (in *WorkloadService) GetWorkloadList(ctx context.Context, criteria Workloa
 	var end observability.EndFunc
 	ctx, end = observability.StartSpan(ctx, "GetWorkloadList",
 		observability.Attribute("package", "business"),
+		observability.Attribute("namespace", criteria.Namespace),
+		observability.Attribute("rateInterval", criteria.RateInterval),
+		observability.Attribute("queryTime", criteria.QueryTime),
 	)
 	defer end()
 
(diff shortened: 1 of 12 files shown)


aljesusg:
mmm Service details returns already the health so we only need to make changes in workloads and apps @lucasponce 

But this is not consistency ....We return health in service details by default and we are going to add "health" queryparam for app and workload. Maybe we should return the health by default here too

lucasponce:
> But this is not consistency ....We return health in service details by default and we are going to add "health" queryparam for app and workload. Maybe we should return the health by default here too

Or perhaps add an include flag, I don't think health is required in all use cases, so the flag may help on those cases.

lucasponce (approved):
I've tested locally with several combinations:
image
image
image

I've identified a couple of places where the backend can use the NamespaceHealth queries instead to iterate per element,  but as there is parallel work, I think we need to see that when all efforts are together we don't see any regression.

https://github.com/kiali/kiali/pull/4772

jshaughn (approved):
Tested with https://github.com/kiali/kiali-ui/pull/2329 and things seems to be working.  A couple of non-blocking comments.

aljesusg:
> Tested with https://github.com/kiali/kiali-ui/pull/2329 and things seems to be working.  A couple of non-blocking comments.

You're right let me address comments tomorrow

business/workloads.go
```diff
@@ -273,23 +279,25 @@ func FilterUniqueIstioReferences(refs []*models.IstioValidationKey) []*models.Is
 
 // GetWorkload is the API handler to fetch details of a specific workload.
 // If includeServices is set true, the Workload will fetch all services related
-func (in *WorkloadService) GetWorkload(ctx context.Context, namespace string, workloadName string, workloadType string, includeServices bool) (*models.Workload, error) {
+func (in *WorkloadService) GetWorkload(ctx context.Context, criteria WorkloadCriteria) (*models.Workload, error) {
```
israel-hdez, line 282: Args are different between the exported funcs `GetWorkload` and `GetWorkloadList`.

It is misleading when you want to use the funcs, because you won't know which fields you need to really fill in each case without looking at the code. So, I would prefer that each func has it's own args type.

I think this same comment applies to the other files...

Not a blocker.

handlers/apps.go
```diff
@@ -17,6 +17,7 @@ type appListParams struct {
 	//
 	// in: path
 	Namespace string `json:"namespace"`
+	AppName   string `json:"app"`
```
israel-hdez, line 20: And same comment here. I think I prefer having a type for `appListParams` and `appDetailsParams`.

Here, I think this introduces a weird user experience, because this will land in the Api docs will which will show `AppName` as a parameter for the AppList endpoint.

israel-hdez (approved):
I tried and it works.

I have non-blocker comments. Go ahead and merge without fixing if you want.

handlers/apps.go
aljesusg: I agree with that...we should create a common type params for all they are so similar but we have not time now we can do it later. Let me create a new type tomorrow

Issue #4758, opened 2022-02-24
Sidecar Validations - Workloads should be from local namespace


hhovsepy:
Happens on Kiali master.

Currently Sidecar validations for workloads selector considers workloads from all namespaces.

But in fact it should be only local Sidecar's namespace workloads. https://istio.io/latest/docs/reference/config/networking/sidecar/#WorkloadSelector

![Screenshot from 2022-02-24 14-19-16](https://user-images.githubusercontent.com/604313/155531777-5a401734-5c3e-42e1-abcc-ab2d9e3e1c30.png)

PR #4803, opened 2022-03-15
Sidecar's workload selector should take workload from own namespace.

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/4758

Sidecar's Workload Selector now checks workloads and other Sidecars only from local namespace.

Before:

After:
Screenshot from 2022-03-15 17-46-12

### DIFF
--- a/business/checkers/common/multi_match_selector_checker.go
+++ b/business/checkers/common/multi_match_selector_checker.go
@@ -72,7 +72,12 @@ func RequestAuthenticationMultiMatchChecker(subjectType string, ra []security_v1
 func SidecarSelectorMultiMatchChecker(subjectType string, sc []networking_v1alpha3.Sidecar, workloadList models.WorkloadList) GenericMultiMatchChecker {
 	keys := []models.IstioValidationKey{}
 	selectors := make(map[int]map[string]string, len(sc))
-	for i, s := range sc {
+	i := 0
+	for _, s := range sc {
+		if s.Namespace != workloadList.Namespace.Name {
+			// Workloads from Sidecar's own Namespaces only are considered in Selector
+			continue
+		}
 		key := models.IstioValidationKey{
 			ObjectType: subjectType,
 			Name:       s.Name,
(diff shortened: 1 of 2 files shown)


business/checkers/common/multi_match_selector_checker.go
```diff
@@ -72,7 +72,12 @@ func RequestAuthenticationMultiMatchChecker(subjectType string, ra []security_v1
 func SidecarSelectorMultiMatchChecker(subjectType string, sc []networking_v1alpha3.Sidecar, workloadList models.WorkloadList) GenericMultiMatchChecker {
 	keys := []models.IstioValidationKey{}
 	selectors := make(map[int]map[string]string, len(sc))
-	for i, s := range sc {
+	i := 0
+	for _, s := range sc {
```
jshaughn, line 76: Not really a problem, but sort of confusing.  The type for `selectors` is `map[int]map[string]string`.  If the keys need to be 0,1,2... then why not use an Array, which would be more clear.

jshaughn (approved):
Seems to be working and no regressions detected.  One non-blocking comment.

business/checkers/common/multi_match_selector_checker.go
hhovsepy: I just avoided touching the existing logic, not to cause any regression, only skipping unnecessary Sidecars.

jshaughn: +1 to minimize regression risk.  It's not a problem, I just found it odd to have to worry about the int-keys being sequential, for a map.  @lucasponce, any comment?  I think this was introduced in the big client-go refactor.

hhovsepy:
@lucasponce can you please verify this PR?

business/checkers/common/multi_match_selector_checker.go
lucasponce: I guess that the index in the map may not be consecutive and an array may not work in those cases, but this comes after some refactoring of this checker and I'm ok with not revisiting the original design to avoid risks.

But I have an ultra-minor go-lang issue, I don't see why you'd need the "i" variable out of the "range", isn't working with the continue ?

Not a blocker :) just in case if it has a reason that I missed.

hhovsepy: @lucasponce
actually 'i' index is required to be as a sequence, used in mapping indexes in keys and sequences, otherwise by the same index the key will be different one, that is why I did not want to touch the logic :)

jshaughn: ...and this is exactly why I pointed out this confusing area :-)

lucasponce (approved):

PR #4812, opened 2022-03-17
Remove unused health endpoints

lucasponce:
Fix duplication of fetches inside business health methods
Fix missing cache hit for DaemonSet
Fix swagger

UI https://github.com/kiali/kiali-ui/pull/2335

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -175,7 +175,7 @@ func (in *AppService) GetAppList(ctx context.Context, criteria AppCriteria) (mod
 			}
 		}
 		if criteria.IncludeHealth {
-			appItem.Health, err = in.businessLayer.Health.GetAppHealth(ctx, criteria.Namespace, appItem.Name, criteria.RateInterval, criteria.QueryTime)
+			appItem.Health, err = in.businessLayer.Health.GetAppHealth(criteria.Namespace, appItem.Name, criteria.RateInterval, criteria.QueryTime, valueApp)
 			if err != nil {
 				log.Errorf("Error fetching Health in namespace %s for app %s: %s", criteria.Namespace, appItem.Name, err)
 			}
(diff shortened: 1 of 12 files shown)


business/health.go
```diff
@@ -4,16 +4,12 @@ import (
...
-	"k8s.io/apimachinery/pkg/labels"
-
-	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
-	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/observability"
 	"github.com/kiali/kiali/prometheus"
+	"github.com/prometheus/common/model"
+	"k8s.io/apimachinery/pkg/api/errors"
```
lucasponce, line 12: Ops, wrong import

business/apps.go
```diff
@@ -175,7 +175,7 @@ func (in *AppService) GetAppList(ctx context.Context, criteria AppCriteria) (mod
 			}
 		}
 		if criteria.IncludeHealth {
-			appItem.Health, err = in.businessLayer.Health.GetAppHealth(ctx, criteria.Namespace, appItem.Name, criteria.RateInterval, criteria.QueryTime)
+			appItem.Health, err = in.businessLayer.Health.GetAppHealth(criteria.Namespace, appItem.Name, criteria.RateInterval, criteria.QueryTime, valueApp)
```
jshaughn, line 178: Not a problem but why do you drop `ctx` as the first parameter here, but keep it in, say, the analogous code in business/workloads.go?

jshaughn (approved):
Built, ran and saw no issues.  One non-blocking question.

business/apps.go
lucasponce: The trace was missing, but it's good to add them.
I'll update it.

business/workloads.go
```diff
@@ -1371,7 +1371,11 @@ func fetchWorkload(ctx context.Context, layer *Layer, criteria WorkloadCriteria)
 		}
 		var err error
 		if isWorkloadIncluded(kubernetes.DaemonSetType) {
-			ds, err = layer.k8s.GetDaemonSet(criteria.Namespace, criteria.WorkloadName)
+			if IsNamespaceCached(criteria.Namespace) {
```
lucasponce, line 1374: This missing cached was being hit several times in cascade.
I had also in my config the other workloads that don't use informers and I accumulated this long delay.

The delay was always there, but with the refactoring, it was more evident.

PR #4816, opened 2022-03-17
Parallelize health appender

nrfox:
Speed up the health appender by fetching health in parallel.

### DIFF
--- a/graph/telemetry/istio/appender/health.go
+++ b/graph/telemetry/istio/appender/health.go
@@ -119,15 +119,18 @@ func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *g
 		err            error
 	}
 	healthCh := make(chan result)
-	errCh := make(chan error)
+	var errors []error
 
 	// Start accumulating results before fetching health info
 	doneAccumulating := make(chan struct{})
 	go func() {
+		// Need to loop over healthCh until closed, otherwise any go routines
+		// sending to this chan will remain blocked forever. Ideally as soon
+		// as we get an error we'd propagate cancellation to all the running
+		// goroutines.
 		for res := range healthCh {
 			if res.err != nil {
-				errCh <- res.err
-				return
+				errors = append(errors, res.err)
 			}
 			if res.appHealth != nil {
 				appHealth[res.namespace] = *res.appHealth
@@ -182,12 +185,10 @@ func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *g
 	// Wait until all results have been sent on the chan until closing
 	wg.Wait()
 	close(healthCh)
-
-	select {
-	case <-doneAccumulating:
-	case err := <-errCh:
+	<-doneAccumulating
+	if len(errors) > 0 {
 		// Check for error needs to happen in the main routine or else panic won't get propagated up
-		graph.CheckError(err)
+		graph.CheckError(errors[0])
 	}
 
 	for namespace, kinds := range healthReqs {
--- a/graph/telemetry/istio/appender/health.go
+++ b/graph/telemetry/istio/appender/health.go
@@ -1,6 +1,7 @@
 package appender
 
 import (
+	"sync"
 	"time"
 
 	"github.com/kiali/kiali/graph"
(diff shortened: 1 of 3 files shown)


graph/telemetry/istio/appender/health.go
```diff
@@ -105,18 +106,96 @@ func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *g
...
+			if res.appHealth != nil {
+				appHealth[res.namespace] = *res.appHealth
+			} else if res.serviceHealth != nil {
+				serviceHealth[res.namespace] = *res.serviceHealth
+			} else if res.workloadHealth != nil {
+				workloadHealth[res.namespace] = *res.workloadHealth
+			}
+		}
+		doneAccumulating <- struct{}{}
+	}()
```
israel-hdez, line 144: I'd lean towards not doing parallelization of this block.
Since this is doing simple assignments, I think time reduction will be negligible.

I think that synchronization could be achieved purely by using the `healthCh`, removing the need of the `wg` WaitGroup and the `doneAccumulating` channel, making code slightly clearer. Basically, rather than using `wg.Add(i)` use a simple counter to know how many thread you've created. Then, move this block of code down after the `wg.Wait()`. You will no longer need to enclose this block inside the `go func()`. Then, the `for res := range healthCh {` would become a simpler `for i := 1; i <= goroutines_counter; i++ {`

graph/telemetry/istio/appender/health.go
```diff
@@ -105,18 +106,96 @@ func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *g
...
+	// Wait until all results have been sent on the chan until closing
+	wg.Wait()
+	close(healthCh)
+	<-doneAccumulating
+	if len(errors) > 0 {
+		// Check for error needs to happen in the main routine or else panic won't get propagated up
+		graph.CheckError(errors[0])
+	}
+
+	for namespace, kinds := range healthReqs {
```
israel-hdez, line 194: I was expecting parallelization to be done to this existing loop rather than doing it in steps.

jshaughn (approved):
Working and code looks good!

graph/telemetry/istio/util/util_test.go
```diff
@@ -38,7 +38,11 @@ func TestAddScopeMultiLabel(t *testing.T) {
...
-	assert.Equal(t, expected, scopedQuery)
+	if expected != scopedQuery {
+		// Maps are unordered so check both permutations that the labels could've been set in.
+		expected = `sum(rate(istio_requests_total{cluster="cluster1",mesh_id="mesh1",reporter="destination",destination_workload_namespace="bookinfo"} [600s])) by (source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,request_protocol,response_code,grpc_response_status,response_flags)`
+		assert.Equal(t, expected, scopedQuery)
+	}
```
israel-hdez, line 45: No assertion when expected == scopedQuery?

israel-hdez (commented):
Despite this is working, I'm not approving because I'm doing a soft-rejection.

Comments are suggestions, so feel free to discard.

My "soft-rejection" is that although doing parallelization is good, this approach may lead to too much parallelization. I would suggest to limit parallelization to some hard number (like how browsers do). Twenty concurrent requests sounds reasonable. From GitHub issues, it is a fact that some people have envs with hundreds of namespaces. Some even have thousands. Kiali basically would do a DDoS attack to prometheus for those cases. So, I'm against this as-is.

It is a "soft-rejection", so you can still merge. However, if somebody to complains that Kiali is adding too much pressure to their prom server, we would be forced to backport a fix...

jshaughn:
I believe Prometheus has its own throttling but I could be wrong.  But even if it denies a request, I don't think we want to fail.  One question is whether we have this potential issue in other places as well.  And if so, we should maybe have a convention/best practice for handling.  Maybe we already do?   Anyway, I agree that we should be careful to limit our concurrent requests.

I'd suggest we merge this for 1.48 but then follow up.  Edgar, can you creat an issue and we can first use it to capture any discussion around the general problem, and then move forward with a PR that addresses the concern, using either an existing,or new (or modified), convention.  We can consider backporting the change if we consider the issue a bug.

graph/telemetry/istio/appender/health.go
nrfox: Yeah that'd work. Basically doing?:
```
	for i := 0; i <= numberOfRequests; i++ {
		res := <-healthCh	
          ....
	}
```

graph/telemetry/istio/appender/health.go
nrfox: I found it easier to first gather all the health data first rather than jamming all the logic into the already nested block but I can play around with that a bit.

graph/telemetry/istio/util/util_test.go
nrfox: That's what line 44 is doing?

nrfox:
> My "soft-rejection" is that although doing parallelization is good, this approach may lead to too much parallelization. I would suggest to limit parallelization to some hard number (like how browsers do). Twenty concurrent requests sounds reasonable. From GitHub issues, it is a fact that some people have envs with hundreds of namespaces. Some even have thousands. Kiali basically would do a DDoS attack to prometheus for those cases. So, I'm against this as-is.

This is a good point and having a cap to the number of in flight requests sounds reasonable given that the number of namespaces is unbounded. There will be many requests to the kube API as well. Caching there should reduce the number of actual requests to the API but still something to consider.

> It is a "soft-rejection", so you can still merge. However, if somebody to complains that Kiali is adding too much pressure to their prom server, we would be forced to backport a fix...

I'd personally prefer fixing it as part of this PR rather than needing to backport a fix later unless it's necessary we get this change in.

lucasponce:
We can control the number of concurrent routines via configuration (using GOMAXPROCS).
That would give control rather than create another variable in Kiali config like "max-parallel-namespace".

I suggest we merge this if no blockers are found; as compared to having a sequential query of all namespaces, this approach will help in those cases.

The discussion is valid, but let's see performance improvements in an iterated way rather a single step.

graph/telemetry/istio/appender/health.go
lucasponce: +1, parallel operations should involve typically I/O, this goroutine can go in the main flow.

lucasponce (approved):
One minor non-blocker comment, but this approach looks good to me.
As discussed, parallel flow can be controlled with ENV variables.

I did some tests in a local env with some mid load:
image

And overall the PR provides an avg ~0.5s compared with ~0.6s from the master branch.

Performance discussions and efforts are long-term but this approach would keep the concurrent fetch triggered from the UI but not inside the backend (saving networking and marshaling steps).

Also, having this approach would let users with high machines to tweak ENV variables and tests other alternatives, having a single iterative model wouldn't give that opportunity.

I've found no blockers navigating to the details or similar.

graph/telemetry/istio/appender/health.go
israel-hdez: > Yeah that'd work. Basically doing?:
> 
> ```
> 	for i := 0; i <= numberOfRequests; i++ {
> 		res := <-healthCh	
>           ....
> 	}
> ```

You've got the idea :smile:

graph/telemetry/istio/util/util_test.go
israel-hdez: My eyes (or GitHub) may be fooling me, but I see that assert inside the `if` block. So I thought it is not being evaluated when `expected == scopedQuery` is true.

jshaughn:
After some off-pr discussion, we've decided to hold on merging this for v1.48.   The perf benefit may not merit the potential risk.    If we decide this is an unacceptable perf hit, or if the community hits a problem, we'll backport the change when maintainers are more comfortable.

Adding DNM label and moving to the next sprint.

israel-hdez:
> > We can control the number of concurrent routines via configuration (using GOMAXPROCS).
> 
> afaik this does not control how many goroutines are created but rather how many CPU cores are available to the process. You could still have many thousands of goroutines even with a small number set for `GOMAXPROCS`.

GOMAXPROCS is actually a good reference for tasks that are using the CPU... you don't really want to fire more threads than processors, because past that number there is no performance gain. Indeed, performance may drop slightly due to too much context switching in the proccessor. So it is something to take into account.

However, since this is mostly waiting for requests to reply, we can have more than `GOMAXPROCS` with no problems as long as we don't add much pressure to the other services being invoked.

nrfox:
It has been decided that the potential perf penalty of doing the health requests sequentially outweigh the risks of doing all the requests simultaneously. If need be, we can backport a fix for limiting the concurrency.

PR #4821, opened 2022-03-19
Use a tagged version of metallb

israel-hdez:
Hopefully, this will fix the failing molecules in GHActions.

---

Prefer a tagged version instead of master.

This is also removing the creation of a secret, which is no longer
needed (it is created automatically).

### DIFF
--- a/hack/ci-kind-molecule-tests.sh
+++ b/hack/ci-kind-molecule-tests.sh
@@ -375,9 +375,9 @@ EOF
   infomsg "Create Kind LoadBalancer via MetalLB"
   lb_addr_range="255.70-255.84"
 
-  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/master/manifests/namespace.yaml
-  ${CLIENT_EXE} create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
-  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/master/manifests/metallb.yaml
+  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/namespace.yaml
+  # ${CLIENT_EXE} create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/metallb.yaml
 
   subnet=$(${DORP} network inspect kind --format '{{(index .IPAM.Config 0).Subnet}}')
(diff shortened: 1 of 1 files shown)


hack/ci-kind-molecule-tests.sh
```diff
@@ -375,9 +375,9 @@ EOF
   infomsg "Create Kind LoadBalancer via MetalLB"
   lb_addr_range="255.70-255.84"
 
-  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/master/manifests/namespace.yaml
-  ${CLIENT_EXE} create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
-  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/master/manifests/metallb.yaml
+  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/namespace.yaml
+  # ${CLIENT_EXE} create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+  ${CLIENT_EXE} apply -f https://raw.githubusercontent.com/metallb/metallb/v0.12.1/manifests/metallb.yaml
```
Xunzhuo, line 380: Should we use a variable to fill this value? since it is a const now.

jmazzitelli:
FYI: our other hack scripts use the configmap method too. We are going to want to eventually update this as well -- the k8s-minikube.sh one is very important to keep working.

```
$ grep -lR metallb *
hack/ci-kind-molecule-tests.sh
hack/k8s-minikube.sh
hack/istio/multicluster/start-kind.sh
```

hack/ci-kind-molecule-tests.sh
israel-hdez: I think it is OK that this is a constant.
The upcoming version of metallb seems to have breaking changes. So, simply changing versions does not work always.

And this is just a hack script. We usually fix versions in hack scripts.

israel-hdez:
> our other hack scripts use the configmap method too. We are going to want to eventually update this as well

Looks like `k8s-minikube.sh` is using minikube's addon, which I assume it's a stable version rather than `master`. Probably, there is nothing to do on that side.

The `start-kind.sh` seems to have the same problem as here. I'll fix that one...

israel-hdez:
@jmazzitelli I fixed the `start-kind.sh` script. GitHub still doesn't reflect the change here. Apparently, GitHub status is degraded.

jmazzitelli:
> I fixed the `start-kind.sh` script. GitHub still doesn't reflect the change here. Apparently, GitHub status is degraded.

I see it in your branch: https://github.com/israel-hdez/swscore/commit/a61c96eab1f98903b023791efdde7dc1d16a0369

LGTM

jmazzitelli (approved):

PR #4822, opened 2022-03-21
use the config var - don't Get() again

jmazzitelli:
I found this by accident when I happened to read this code. We should fix this - no sense getting the config again when we already have it.


lucasponce (approved):

aljesusg (approved):
LGFM

Issue #4678, opened 2022-01-27
Optimize the Kiali Cache design under cluster rights presence


lucasponce:
This is a follow-up issue started in https://github.com/kiali/kiali/issues/4502#issuecomment-1021162055

Kiali uses k8s Informers to access k8s APIs (workload, services, and Istio configuration).
The current design creates an Informer per namespace+API, and the Informer is responsible to handle all low-level events and provide a cache for the API objects.

This design works reasonably well, but in clusters with a high level of namespaces, the number of informers can add a significant overhead to the Kiali backend (an Informer requires a connection with the API and its own memory).

If the Kiali ServiceAccount doesn't have cluster rights, this design can't change (unless Informers are removed, but then, Kiali would need to query all low-level APIs and keep them updated).

But, if the ServiceAccount has cluster rights (or accessibleNamespaces == "**") then this design could be optimized and it can change to use a single informer per API that watches all namespaces.

This could be an optional enhancement in the Kiali backend, it shouldn't change the business behavior, but under this scenario, the Kiali Cache could be optimized.

nrfox:
This has come up a few different times wrt scalability and it seems like an issue that needs to be addressed. Is this ready to be worked on?

PR #4823, opened 2022-03-21
Use pool of workers to do health requests in Health Appender

nrfox:
Followup to #4816. Creates a fixed amount of goroutines so that no more than 20 requests to the health service are made in parallel.

### DIFF
--- a/graph/telemetry/istio/appender/health.go
+++ b/graph/telemetry/istio/appender/health.go
@@ -1,9 +1,12 @@
 package appender
 
 import (
+	"context"
+	"fmt"
 	"sync"
 	"time"
 
+	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/graph"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
@@ -19,20 +22,22 @@ type HealthAppender struct {
 	Namespaces        graph.NamespaceInfoMap
 	QueryTime         int64 // unix time in seconds
 	RequestedDuration time.Duration
+
+	mu sync.Mutex
 }
 
 // Name implements Appender
-func (a HealthAppender) Name() string {
+func (a *HealthAppender) Name() string {
 	return HealthAppenderName
 }
 
 // IsFinalizer implements Appender
-func (a HealthAppender) IsFinalizer() bool {
+func (a *HealthAppender) IsFinalizer() bool {
 	return true
 }
 
 // AppendGraph implements Appender
-func (a HealthAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo, _ *graph.AppenderNamespaceInfo) {
+func (a *HealthAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo, _ *graph.AppenderNamespaceInfo) {
 	if len(trafficMap) == 0 {
 		return
 	}
@@ -63,6 +68,14 @@ func (a *HealthAppender) attachHealthConfig(trafficMap graph.TrafficMap, globalI
 			continue
 		}
 	}
+}
+
+type result struct {
+	namespace      string
+	appHealth      *models.NamespaceAppHealth
+	serviceHealth  *models.NamespaceServiceHealth
+	workloadHealth *models.NamespaceWorkloadHealth
+	err            error
 }
 
 func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo) {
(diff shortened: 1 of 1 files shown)


israel-hdez (dismissed):
The pool of workers make the code more complex. I don't think we can simplify it more. IMO, you already did it as simpler as it could be.

Approving. It's working all fine.

lucasponce:
Please @nrfox would you mind rebasing it with the master branch.
It's unrelated to this work, but to have a built one would help to have the latest changes.

Thanks

nrfox:
@lucasponce sure, it's now rebased with latest master.

lucasponce:
Thanks, @nrfox, I have mixed feelings about this proposal.
Let me describe my thoughts and if you like we can comment:
- This work would work, but conceptually, I don't think the appender should slow down the concurrent queries at that level.
- Also, if the Prometheus client talks in HTTP2 with the server, some of the queries can be multiplexed or enqueue in a transparent way (the sender shouldn't be responsible to throttle them).
- Those invocations `bs.Health.GetNamespaceAppHealth, bs.Health.GetNamespaceServiceHealth,bs.Health.GetNamespaceWorkloadHealth` are cached, so it shouldn't prevent that they don't happen concurrently.
- If we want to make a pool in Prometheus, perhaps it can be done at the Prom level layer, not on the business side (all Prometheus queries can be run by a pool of workers - in case there is not an already defined).
- Originally, this problem/discussion come with the concert that a scenario with a large number of namespaces may trigger a problem, yes, I share that concern, but I think that before the user hits this one, it would require addressing the problem of the lack of cluster rights defined here https://github.com/kiali/kiali/issues/4678.

So, in a nutshell, I'd try to check if:
- The Prometheus client offers a pool that is close to the client and does not add pools in the business logic (if we can implement one that helps for all cases). I've not investigated but I remember that Prom moved to http2 some time ago, so that would also help to those scenarios.
- And in a case where we have 200 namespaces, there will be 200 queries to Prom, but in the Kubernetes API where will be 200x20 queries, in other words, probably the real problem would be there.

I know this PR works, but it adds some complexity compared with the clarity it has today, and if we need a prom in the Prometheus, I'd like to have it close to the Prom client rather than the business logic (I think).

P.S.: Happy to have a session on the code that perhaps can help to explain better my points.

nrfox:
@lucasponce thanks for the feedback. Couple of thoughts but I mostly agree with what you suggest.

> * Also, if the Prometheus client talks in HTTP2 with the server, some of the queries can be multiplexed or enqueue in a transparent way (the sender shouldn't be responsible to throttle them).

While this will help with the overhead of managing many connections and improve the speed of responses, the concern was more about the load that sending a large number of queries to prom at once would cause on the prom server. The prom server would still need to churn through all the queries at once regardless of how many connections are open with the client.

> If we want to make a pool in Prometheus, perhaps it can be done at the Prom level layer, not on the business side (all Prometheus queries can be run by a pool of workers - in case there is not one already defined).

Agreed here. I think moving this logic up a level into either the `Health` service or the prom layer and providing something like a `GetAppHealthForNamespaces(namespaces, duration, queryTime)` method would probably be a better separation of concerns. Perhaps then we can even do a single prom query to fetch all the info needed across namespaces as opposed to per namespace. The kube api doesn't support this but all those objects are cached so it doesn't matter how many times the cache is requested at once.

> Originally, this problem/discussion come with the concern that a scenario with a large number of namespaces may trigger a problem, yes, I share that concern, but I think that before the user hits this one, it would require addressing the problem of the lack of cluster rights defined here https://github.com/kiali/kiali/issues/4678.

IMO the kubernetes layer is less of a concern here than the prom layer because of the caching that happens for kube objects. While having a single watch instead of one per namespace is definitely more efficient, I don't think it'd help much here unless kiali was hitting the limits, if there are any, on the number of watches you can have open at once. In which case, that is a broader issue. Though all of the namespace scaling issues I've seen so far have been either the UI making many requests to the backend _per_ namespace or things that wouldn't benefit from a cluster scoped cache like POST requests per namespace to the subject access review api.

lucasponce:
> Perhaps then we can even do a single prom query to fetch all the info needed across namespaces as opposed to per namespace.

Well, instead of refactoring the business logic, I was thinking to capture the "Api.Query()" call in the Prom level and use it with a pool in a first approach. Again, in case the current client with http2 doesn't support it.

> IMO the kubernetes layer is less of a concern here than the prom layer because of the caching that happens for kube objects. While having a single watch instead of one per namespace is definitely more efficient, I don't think it'd help much here unless kiali was hitting the limits, if there are any, on the number of watches you can have open at once.

Well, I'm afraid not here. Kiali opens the watches and those are permanents connections, it's a factor of 20 times the number of the namespace. I'm just comparing priorities. I think this situation would produce first a scalability problem rather than the scalability of the Prometheus that is easier addressable from configuration and scalability at the server side (also we moved to start the caches at the start).

My point is that early optimizations may miss some aspect or introduce some logic that adds complexity with less good results due to the missing context in the whole picture.

lucasponce:
> I'd actually argue the opposite. afaik establishing/maintaining watches with the kube API server is pretty cheap in terms of resources.

Establishing watches per different namespaces invokes a first "query all" for that resource, but that's the subject of the work on #4678, the current design is not optimized, IMO, I have some suspects that it causes extra pressure to the Kube API.

>  Prom on the other hand has to execute each one of the queries. Depending on the queries themselves this may or may not be resource-intensive. The Prom http API may or may not have some kind of throttling mechanism. 

That's my point, that I think this should be the focus; check if the client/server offer mechanism to configure this, and if we want to add a pool, the "kiali/prometheus/client.go" could be the better location to make it transparent to the business logic. As in the current design, the Prometheus caching is located at that level.

This discussion is good, as I think it would require iterative steps.

nrfox:
> That's my point, that I think this should be the focus; check if the client/server offer mechanism to configure this, and if we want to add a pool, the "kiali/prometheus/client.go" could be the better location to make it transparent to the business logic. As in the current design, the Prometheus caching is located at that level.

I agree that at the very least we shouldn't put this logic in the appender. I'm going to close this since this PR is focused on the appender and we can decide later if we want to include similar logic in the prom client.

PR #4824, opened 2022-03-21
[hack] do not allow using 'oc' if not logged in or not talking to OpenShift.

jmazzitelli:
This fixes the problem where the cluster resources aren't purged when on non-OpenShift but -c kubectl is not specified.
You now are told that -c kubectl is needed.


jshaughn (approved):
Fine.

PR #4831, opened 2022-03-22
hack and make changes for the kiali-ui repo merge into kiali repo

jmazzitelli:
This PR contains the changes I think need to be made when the kiali-ui repo is merged into the kiali repo.

The only thing missing would be a make target "make ui". I suggest something simple like the following in `make/Makefile.build.mk`:

```make
build-ui:
  yarn install && yarn run build
```

If we don't do that, then we should just expect people to do a `yarn install && yarn run build` in the new `frontend` directory where the kiali-ui code will go.

cc @aljesusg you can either (1) manually put these changes in your own PR, or (2) you can merge this PR into yours. I just created this PR to document what changes will be needed.

NOTE! I did NOT make any changes to the Jenkins files - there will be changes necessary for that. I'll leave that for someone else who knows more about the Jenkins stuff than I.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ COMMIT_HASH ?= $(shell git rev-parse HEAD)
 # have been built before trying to create a kiali server container
 # image. The UI project is configured to place its build
 # output in the $UI_SRC_ROOT/build/ subdirectory.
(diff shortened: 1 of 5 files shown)


aljesusg:
Commits merged in https://github.com/kiali/kiali/pull/4835

Issue #4833, opened 2022-03-22
Hack script to create a Kind cluster in CI


nrfox:
Probably [this script](https://github.com/kiali/kiali/blob/master/hack/ci-kind-molecule-tests.sh) can be reused without the molecule parts since this script is already:
- Creating a Kind cluster
- Installing Istio
- Installing bookinfo
~- Installing the Kiali operator~
- Installing Kiali

Epic: https://github.com/kiali/kiali/issues/4827

jmazzitelli:
we have this too - it creates two clusters but code can be stolen. https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/start-kind.sh

nrfox:
We'll need to take into account linked PRs even when the front-end and back-end repos are combined. When there's a kiali config change that requires an operator change, we'll need to deploy the right version of the operator to run the CI tests against. We usually link these with comments on the PR discussion like: `Operator PR: #PR-Num`. Perhaps we can formalize that where if an issue contains a "required-operator" label then we'll search for the operator PR link in the comments and run that version of the operator. It looks like the ci-kind script [has support](https://github.com/kiali/kiali/blob/master/hack/ci-kind-molecule-tests.sh#L338-L350) for something similar already, we'll just need to agree on a convention and add support for getting the label info and PR from the github API.

lucasponce:
Note that for the e2e in upstream we don't even need to build the UI or the operator.
Just create the cluster (with Istio and the demos), open a proxy, and launch a Kiali backend process pointing to it.
That would help to cover all the e2e testing, and probably would be faster than the current pipelines.

The pipelines building the UI and the operator are good but are slow and they can't run on every PR (but can be used for daily testing).

jmazzitelli:
> we don't even need to build the UI or the operator.

Right - for these e2e backend api tests, there is really no need to build the operator and install via the operator (and it will be quicker to run the tests if you don't).

We have molecule testing that will catch any operator errors.

nrfox:
> Note that for the e2e in upstream we don't even need to build the UI or the operator.

I'm assuming this hack script will be used for both the e2e API tests _and_ the e2e front end tests using cypress. In both cases you'll need the same environment setup. You're right, we don't need to deploy/test the operator here and it'll make things simpler to deploy kiali independently.

> The pipelines building the UI and the operator are good but are slow and they can't run on every PR (but can be used for daily testing).

[...]

PR #4838, opened 2022-03-23
Setup real testing env in CI to run e2e tests against

nrfox:
Updates the GH actions kiali workflow by:
- Removing the UI workflow
- Separating the frontend and backend build/unit testing into separate parallel jobs
- Adding a script to setup Kind in GH actions and install kube/istio/kiali
- Adding a job to run the cypress integration tests using the kind environment
  - When cypress tests fail, videos/screenshots of the failed tests are uploaded as artifacts to the GH job. These can be downloaded to help diagnose test failures.
- The kiali workflow is run:
  - On every PR to master
  - On every PR to a release branch
  - On every release

Fixes: #4833


hack/setup-kind-in-ci.sh
```diff
@@ -0,0 +1,149 @@
...
+
+infomsg "Installing kiali server via Helm"
+helm install \
+  --namespace istio-system \
+  --set auth.strategy="anonymous" \
+  --set deployment.service_type="LoadBalancer" \
+  --set deployment.image_name=kiali/kiali \
+  --set deployment.image_version=dev \
+  --set deployment.image_pull_policy="Never" \
+  --repo https://kiali.org/helm-charts \
```
jmazzitelli, line 115: Using the --repo option with the kiali.org chart location is most likely going to be a problem. Because this is going to install using the last released helm chart. But it is commonly the case that we change the helm charts throughout a sprint (adding permissions, adding/changing configuration settings, changing Deployment or other resource yaml templates, etc). So we will want to test the master build of the helm charts, otherwise, this could fail because it will be installing with the helm chart that may not be compatible with the master branch of kiali/kiali.

To do this, we would need to checkout the master branch of the helm chart repo, run `make build-helm-charts` and then use the helm charts in the helm chart repo's `_output` directory. The helm command would look like:

```
helm install \
  ... the --set and --namespace options ... \
  kiali-server \
 _output/charts/kiali-server-*-SNAPSHOT.tgz
```

nrfox: hmmmm wouldn't we need to merge the changes into the helm chart before running the tests here then? Otherwise we'd need to target the branch with the helm changes to coincide with the kiali PR. Same thing I was describing here: https://github.com/kiali/kiali/issues/4833#issuecomment-1075605397.

hack/setup-kind-in-ci.sh
```diff
@@ -0,0 +1,149 @@
...
+
+  infomsg "Waiting for kiali pod to be ready"
+  infomsg "Latest events"
+  kubectl get events -l app=kiali -n istio-system --sort-by='.lastTimestamp'
+  infomsg "Kiali pod"
+  kubectl get pods -l app=kiali -n istio-system -o yaml
+  sleep 5
+done
+
+infomsg "Kiali is ready."
```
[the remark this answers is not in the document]
nrfox: The readiness probe on the kiali pod should take care of that. The pod shouldn't be marked as ready until the endpoint is up and responding to traffic.

hack/setup-kind-in-ci.sh
jmazzitelli: Yes. We would need to make sure any helm chart PR dependencies are merged first.

jmazzitelli:
Is it possible to use GITHUB_REF_NAME in place of "master" in all the github action yaml/ci scripts? See: https://docs.github.com/en/actions/learn-github-actions/environment-variables (search for GITHUB_REF).

That way we can run these actions in our release branches so when we backport PRs, we can run these tests on those branches, too. So we could have as an example:

```yaml
on:
  push:
    branches: [master, v1.48]
  pull_request:
    branches: [master, v1.48]
```

nrfox:
> Is it possible to use GITHUB_REF_NAME in place of "master" in all the github action yaml/ci scripts? See: https://docs.github.com/en/actions/learn-github-actions/environment-variables (search for GITHUB_REF).
> 
> That way we can run these actions in our release branches so when we backport PRs, we can run these tests on those branches, too.

That's a good question. It looks like it supports a regex so that will probably work:
```
on:
  push:
    branches:
      - main
      - 'releases/**'
```
in our case something like
```
on:
  push:
    branches: [master, "^v[0-9]+\.[0-9]+"]
  pull_request:
    branches: [master, "^v[0-9]+\.[0-9]+"]
```
I don't know if that's a valid regex or not but you get the idea

.github/workflows/kiali-ci.yml
```diff
@@ -32,7 +47,87 @@ jobs:
...
+          cache: yarn
+          cache-dependency-path: frontend/yarn.lock
+
+      - name: Build
+        run: make clean-all build-ui
+
+      - name: Test frontend
+        run: |
+          cd frontend
+          yarn pretty-quick --check --verbose --branch origin/master
```
jmazzitelli, line 77: does `origin/master` here assume this was triggered by a change to the master branch? Shouldn't this use the GITHUB_REF_NAME? Note that the GITHUB_REF might reference a PR branch name, so I'm not sure what the right thing to do is.

hack/setup-kind-in-ci.sh
```diff
@@ -0,0 +1,157 @@
...
+hack/istio/download-istio.sh
+
+infomsg "Installing istio"
+hack/istio/install-istio-via-istioctl.sh --client-exe-path "$(which kubectl)"
+  
+infomsg "Pushing the images into the cluster..."
+make -e DORP="${DORP}" -e CLUSTER_TYPE="kind" -e KIND_NAME="ci" cluster-push-kiali
+
+infomsg "Cloning kiali helm-charts..."
+git clone --single-branch --branch master https://github.com/kiali/helm-charts.git helm-charts
```
jmazzitelli, line 108: does master here assume this was triggered by a change to the master branch? Shouldn't this use the GITHUB_REF_NAME? Note that the GITHUB_REF might reference a PR branch name, so I'm not sure what the right thing to do is.

jmazzitelli (commented):
just commenting where I see 'master' being used. You are probably already going to fix this all, but wanted to mark these spots anyway.

.github/workflows/kiali-ci.yml
nrfox: Yes! Good eye

hack/setup-kind-in-ci.sh
nrfox: hmmm we probably want the target branch name which would be master if the PR is against the master branch and something like `v1.48` if it's a feature branch. I think the target branch name is an env var too.

hack/setup-kind-in-ci.sh
```diff
@@ -91,10 +91,6 @@ data:
       addresses: ['${first_ip}-${last_ip}']
 LBCONFIGMAP
 
-# if [ -n "${ISTIO_VERSION}" ]; then
-# DOWNLOAD_ISTIO_VERSION_ARG="--istio-version ${ISTIO_VERSION}"
-# fi
-
```
jmazzitelli: if you aren't going to be able to tell this script to run against a specific Istio version, then remove all the obsolete code that deals with passing in the `-iv` command line option.

Although, I think you should keep this. Some branches might not support running against the latest version of Istio.

hack/setup-kind-in-ci.sh
```diff
@@ -91,10 +91,6 @@ data:
...
 infomsg "Downloading istio"
 hack/istio/download-istio.sh
```
jmazzitelli, line 95: you could uncomment what you deleted above and just pass in that env var here to support being able to run against a specific version:
```suggestion
hack/istio/download-istio.sh ${DOWNLOAD_ISTIO_VERSION}
```

hack/setup-kind-in-ci.sh
nrfox: I was just thinking about this. Since this script will run against backported PRs and releases, it'll need to support running the correct version of istio/kubernetes for the version of kiali we are using. Kiali provides the compatibility matrix after it starts up but in this case we need it before Kiali is running to determine which version of istio/kube to deploy. Probably can just keep a short list of the maistra supported releases to pin to a specific version of istio/kube and then use the latest istio/kube for all the rest?

hack/setup-kind-in-ci.sh
nrfox: Remembering now why I commented this out... the download-istio script fails with a weird error when `DOWNLOAD_ISTIO_VERSION=""`.

Error:
```
Unknown argument []. Aborting.
```
Relevant place in script:
```
while [ $# -gt 0 ]; do
  key="$1"
  case $key in
    -div|--dev-istio-version)
      DEV_ISTIO_VERSION="$2"
      shift;shift
      ;;
    -iv|--istio-version)
      ISTIO_VERSION="$2"
      shift;shift
      ;;
    -o|--output)
      OUTPUT_DIR="$2"
      shift;shift
      ;;
    -h|--help)
      cat <<HELPMSG
Valid command line arguments:
  -div|--dev-istio-version <version>: If you want a dev version to download, use this option and do not use -iv.
                                      The value of this option is something like "1.4-dev" or "1.5-dev". If you
                                      want the latest master build, set this to "latest".
  -iv|--istio-version <#.#.#>: Released version of Istio to download. (default will be the latest version)
  -o|--output <dir> : Output directory where Istio is (or will be downloaded to if it doesn't exist).
  -h|--help : This message.
HELPMSG
      exit 1
      ;;
    *)
      echo "Unknown argument [$key]. Aborting."
      exit 1
      ;;
  esac
done
```

Seems like it's interpreting extra whitespace as an argument?

jmazzitelli: that's odd. Wrap the arg value in quotes maybe? Not sure what github is doing weird when passing in the arguments.

Try `--istio-version "${DOWNLOAD_ISTIO_VERSION}"`

hack/setup-kind-in-ci.sh
```diff
@@ -0,0 +1,183 @@
...
+      protocol: layer2
+      addresses: ['${first_ip}-${last_ip}']
+LBCONFIGMAP
+
+if [ -n "${ISTIO_VERSION}" ]; then
+  DOWNLOAD_ISTIO_VERSION_ARG=" --istio-version ${ISTIO_VERSION}"
+fi
+
+infomsg "Downloading istio"
+hack/istio/download-istio.sh"${DOWNLOAD_ISTIO_VERSION_ARG}"
```
jmazzitelli, line 124: found the problem - don't wrap this in quotes:
```suggestion
hack/istio/download-istio.sh ${DOWNLOAD_ISTIO_VERSION_ARG}
```

nrfox: bash ¯\_(ツ)_/¯

nrfox: Hmmm won't let me commit the suggestion since I force pushed so I'll just update it locally and push that

nrfox:
@ScriptingShrimp adding you as a reviewer since I changed the pathing of the cypress BDD plugin. I believe this needed to be updated anyway since the frontend repo was combined with the backend repo but if you could review and run the changes against a live env that'd be great. Also this changes runs the cypress tests in the upstream pipelines now so it would be good for you to review that too. Thanks!

frontend/cypress/plugins/bdd.ts
```diff
@@ -3,32 +3,28 @@ const cucumber = require('cypress-cucumber-preprocessor').default;
...
-    return await config
+    config.env.auth_strategy = await getAuthStrategy(config.baseUrl);
+    return await config;
   }
 
-  return exportConfig()
+  return exportConfig();
 };
-
-
```
nrfox, line 34: Just my linter being pedantic

frontend/cypress/integration/common/kiali_homepage.ts
```diff
@@ -1,13 +1,12 @@
...
+});
 
-
-Then(`I see {string} in the title`, (title) => {
-  cy.title().should('include', title)
-})
\ No newline at end of file
+Then(`I see {string} in the title`, title => {
+  cy.title().should('include', title);
+});
```
nrfox: Just my linter being pedantic

.github/workflows/kiali-ci.yml
```diff
@@ -1,23 +1,41 @@
...
+  # This workflow can run on either a PR or release. GITHUB_BASE_REF is only
+  # set when this workflow is run against a PR. When run against a PR, we want
+  # to use the branch we are running against: GITHUB_BASE_REF. When run
+  # as a release, we want to use the current branch ref: GITHUB_REF_NAME.
+  TARGET_BRANCH: ${{ github.base_ref || github.ref_name }}
 
 jobs:
-  build:
-    name: Build
+  build_backend:
```
leandroberetta, line 18: I'm wondering if we can turn build_backend and build_frontend jobs into reusable workflows (https://docs.github.com/en/actions/using-workflows/reusing-workflows). 

Something similar to the integration one. 

This is because in the release pipeline I'm working on I have very similar steps, that, can be the same I guess.

With this approach, we will have reusable workflows and both release and ci pipelines will use them.

leandroberetta (commented):
LGTM. 

Check the only comment I made, it's not about something to fix, it's how we can structure the workflows to reuse them!.

frontend/cypress/plugins/bdd.ts
nrfox: Yes he does :)

.github/workflows/kiali-ci.yml
nrfox: It looks like just adding `workflow_call:` will make this reusable? I can include that in this change.

leandroberetta: It is just that, but what I wanted to do is to split these jobs in 3 files like:
https://github.com/leandroberetta/kiali/tree/release-pipeline/.github/workflows

So, with these common jobs, both pipelines can use them.

The test worked.

nrfox: Ah so you can't reference individual jobs within a workflow it looks like? Well if it's alright with you, I'll leave making the workflow reusable up to you in your PR. This change was mostly about getting an e2e env setup in CI and having something use it to run tests on every PR.

nrfox:
@jmazzitelli if you wouldn't mind taking another look at this when you get chance

hack/setup-kind-in-ci.sh
```diff
@@ -0,0 +1,183 @@
...
+# If a specific version of Istio hasn't been provided, try and guess the right one
+# based on the Kiali branch being tested (TARGET_BRANCH) and the compatibility matrices:
+# https://kiali.io/docs/installation/installation-guide/prerequisites/
+# https://istio.io/latest/docs/releases/supported-releases/
+if [ "${TARGET_BRANCH}" == "v1.48" ]; then
+  ISTIO_VERSION="1.13"
+elif [ "${TARGET_BRANCH}" == "v1.36" ]; then
+  ISTIO_VERSION="1.10"
+elif [ "${TARGET_BRANCH}" == "v1.24" ]; then
+  ISTIO_VERSION="1.7"
```
jmazzitelli, line 52: I don't think these version strings work with our hack script. I think you need x.y.z. So 1.13.0 or 1.10.0 or 1.7.0. If I try just with x.y this is what I get:

```bash
$ hack/istio/download-istio.sh -iv 1.10
Output Directory: /home/jmazzite/source/kiali/kiali/_output
Will use a specific Istio version: 1.10
Will look for Istio here: /home/jmazzite/source/kiali/kiali/_output/istio-1.10
Cannot find Istio 1.10 - will download it now...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  4926  100  4926    0     0   4805      0  0:00:01  0:00:01 --:--:--  4805

Downloading istio-1.10 from https://github.com/istio/istio/releases/download/1.10/istio-1.10-linux-amd64.tar.gz ...

https://github.com/istio/istio/releases/download/1.10/istio-1.10-linux-amd64.tar.gz is not found, please specify a valid ISTIO_VERSION and TARGET_ARCH
```

.github/workflows/kiali-ci.yml
```diff
@@ -32,7 +55,101 @@ jobs:
...
+
+      - name: Setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: "14"
+          cache: yarn
+          cache-dependency-path: frontend/yarn.lock
+
+      - name: Build
+        run: make clean-all build-ui
```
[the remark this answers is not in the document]
nrfox: It's fine because different jobs have totally independent workspaces. The only way to share artifacts between them is to upload/download the artifacts which is what we do in each build job. Probably don't need to clean here but it's just a precaution.

jmazzitelli (approved):
lgtm

Issue #4737, opened 2022-02-15
Feature flag to disable log browser


larhauga:
Hi, it would be nice if it was possible to turn off the logs browser for workloads.

## My use case
Logs can contain sensitive information that we dont want accessible through the kiali dashboards.

jmazzitelli:
isn't the logs a RBAC permission you can disable per user? e.g. the service account for the user can just be denied pods/log read permission.

larhauga:
> isn't the logs a RBAC permission you can disable per user? e.g. the service account for the user can just be denied pods/log read permission.

Hi, thanks for your quick reply. 
I am running an EKS with no external OIDC currently, and user RBAC is thus not enabled.
I am using the operator, and have not found any way of generally disabling the pod/log rbac for kiali.

Do you know if there is any way of removing the log rbac permission with the operator; or is the only solution available to me to roll my own deployment? 
I currently see no good solutions to ensuring that logs are not available; in a pretty basic setup of kiali (operator+view-only mode). Any help is really appreciated :)

jmazzitelli:
What auth strategy are you using? In your Kiali CR, what is ["spec.auth.strategy"](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.auth.strategy) set to?

larhauga:
> What auth strategy are you using? In your Kiali CR, what is ["spec.auth.strategy"](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.auth.strategy) set to?

Right now `auth.strategy=anonymous`, but planing to migrate to `openid` but possible with `disable_rbac: true` as rbac is currently mapped to aws sso.

jmazzitelli:
Using `auth.strategy` of `anonymous` means ANYONE with access to the Kiali URL can get into the Kiali UI. If you use that in production, I suspect you have more things to worry about than just leaking log messages :)

But regardless of that, the current workaround that you can employ is: remove the `pods/log` permission that is given to the Kiali service account. So you would need to remove this permission from the `kiali-viewer` role: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/role-viewer.yaml#L14

jshaughn:
logs come through the k8s api, not sure if you can disable logs at the k8s layer.

jmazzitelli:
> logs come through the k8s api, not sure if you can disable logs at the k8s layer.

[...]

PR #4839, opened 2022-03-23
provide the ability to disable features

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/4737
(this also fixes https://github.com/kiali/kiali/issues/4871)

server PR: https://github.com/kiali/kiali/pull/4839
operator PR: https://github.com/kiali/kiali-operator/pull/507
helm PR: https://github.com/kiali/helm-charts/pull/130

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -68,6 +68,14 @@ const (
 	FeatureLogView FeatureName = "logs-tab"
 )
 
+func (fn FeatureName) IsValid() error {
+	switch fn {
+	case FeatureLogView:
+		return nil
+	}
+	return fmt.Errorf("Invalid feature name: %v", fn)
+}
+
 // Global configuration for the application.
 var configuration Config
 var rwMutex sync.RWMutex
--- a/config/config.go
+++ b/config/config.go
@@ -61,6 +61,13 @@ const (
 	DashboardsDiscoveryAuto    = "auto"
 )
 
+// FeatureName is the enum type used for named features that can be disabled via KialiFeatureFlags.DisabledFeatures
+type FeatureName string
+
+const (
+	FeatureLogView FeatureName = "log-view"
+)
+
 // Global configuration for the application.
 var configuration Config
 var rwMutex sync.RWMutex
@@ -396,6 +403,7 @@ type CertificatesInformationIndicators struct {
 // KialiFeatureFlags available from the CR
 type KialiFeatureFlags struct {
 	CertificatesInformationIndicators CertificatesInformationIndicators `yaml:"certificates_information_indicators,omitempty" json:"certificatesInformationIndicators"`
+	DisabledFeatures                  []string                          `yaml:"disabled_features,omitempty" json:"disabledFeatures,omitempty"`
 	IstioInjectionAction              bool                              `yaml:"istio_injection_action,omitempty" json:"istioInjectionAction"`
 	IstioUpgradeAction                bool                              `yaml:"istio_upgrade_action,omitempty" json:"istioUpgradeAction"`
 	UIDefaults                        UIDefaults                        `yaml:"ui_defaults,omitempty" json:"uiDefaults,omitempty"`
@@ -617,6 +625,7 @@ func NewConfig() (c *Config) {
 				Enabled: true,
 				Secrets: []string{"cacerts", "istio-ca-secret"},
 			},
+			DisabledFeatures:     []string{},
 			IstioInjectionAction: true,
 			IstioUpgradeAction:   false,
 			UIDefaults: UIDefaults{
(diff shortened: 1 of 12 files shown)


jmazzitelli:
To test - deploy Kiali and go to a workload and view its logs. Then edit the Kiali CR so this is set:
```
spec:
  kiali_feature_flags:
    disabled_features:
    - log-view
```

You should not be able to view logs. Try to get to the API, too, and see you get a 403 Forbidden error code with an error message "Pod Logs access is disabled":

`curl 'http://localhost:20001/kiali/api/namespaces/istio-system/pods/<pod-name>/logs?container=<container-name>&sinceTime=####&tailLines=100&isProxy=false'`

jmazzitelli:
One thing that needs to be reviewed is where else in the UI would the logs be requested/displayed? This PR currently will hide the "Logs" tab in the Workload Details view, but I wonder if there are other places in the UI where logs are retrieved and displayed (perhaps in the metrics views where we overlay tracing information? I thought we also overlay log information too??).

lucasponce:
I'm not sure if this effort should grow too much.
For the logs, it's ok as the motivation was to hide some sensitive info, but I don't think that it should go beyond that.

> This PR currently will hide the "Logs" tab in the Workload Details view, but I wonder if there are other places in the UI where logs are retrieved and displayed (perhaps in the metrics views where we overlay tracing information? I thought we also overlay log information too??).

Logs are only shown in the "logs" tab, the "logs" tab may overlay tracing info, but there is no logging info shown in other pages.

I think it's a little bit anti-pattern that a user cannot access a cluster (even with view-only mode) but can access Kiali.

But I'm fine with this effort, as long as it doesn't expand to other complex scenarios, Kiali shouldn't offer such a level of personalization, it's not its business, other solutions will do a better job.

config/config.go
```diff
@@ -61,6 +61,13 @@ const (
 	DashboardsDiscoveryAuto    = "auto"
 )
 
+// FeatureName is the enum type used for named features that can be disabled via KialiFeatureFlags.DisabledFeatures
+type FeatureName string
+
+const (
+	FeatureLogView FeatureName = "log-view"
```
lucasponce, line 68: I'd suggest to use "logs-tab" instead of "log-view".
That would help to align better with the UI:
image

kiali.go
```diff
@@ -178,6 +178,11 @@ func validateConfig() error {
 		log.Infof("Some validation errors will be ignored %v. If these errors do occur, they will still be logged. If you think the validation errors you see are incorrect, please report them to the Kiali team if you have not done so already and provide the details of your scenario. This will keep Kiali validations strong for the whole community.", cfg.KialiFeatureFlags.Validations.Ignore)
 	}
 
+	// log a info message if the user is disabling some features
+	if len(cfg.KialiFeatureFlags.DisabledFeatures) > 0 {
+		log.Infof("Some features are disabled: [%v]", strings.Join(cfg.KialiFeatureFlags.DisabledFeatures, ","))
```
lucasponce, line 183: In a similar way as it happens with Jaeger and Grafana integration, if the user disables this feature it should get an INFO message in the Message center.

That would reinforce that this is an exceptional config.

lucasponce (changes requested):
A couple of requests:
- Use "logs-tab" to align with the naming used in the UI.
- An INFO message should help to tell the user that the logs tab is disabled (it can go on the login, or at the workload details).

jmazzitelli:
This will now show up in the UI (just to repeat- this is INFO level - it just looks like ERROR - we need to fix this -> https://github.com/kiali/kiali/issues/4871):
notification:
info-msg2

message:
info-msg1

jmazzitelli:
> A couple of requests:

both are done.

jmazzitelli:
I am going to fix that notification color issue since it is a simple three line fix in a single file. See: https://github.com/kiali/kiali/issues/4871#issuecomment-1082091108
info

config/config.go
```diff
@@ -396,6 +403,7 @@ type CertificatesInformationIndicators struct {
 // KialiFeatureFlags available from the CR
 type KialiFeatureFlags struct {
 	CertificatesInformationIndicators CertificatesInformationIndicators `yaml:"certificates_information_indicators,omitempty" json:"certificatesInformationIndicators"`
+	DisabledFeatures                  []string                          `yaml:"disabled_features,omitempty" json:"disabledFeatures,omitempty"`
```
lucasponce, line 406: There should be control over what is populated in that list:
image

Perhaps a fixed struct with flags would help instead of an open list.

But anyway, if the feature listed is not valid, message shouldn't be propagated and should be ignored.

lucasponce:
I think the info messages should be consistent.
Today, there is a similar feature with the "Grafana" links message, with no toast:
image

And the new "disabled feature" message , with toast:
image
image

My thinking is that all should be presented in the same way (with or without toast, but mixing them is a little bit confusing).

jshaughn:
+1 to No Toast

lucasponce (approved):
Tested.
image
image
image
image
image

PR #4841, opened 2022-03-24
Fix link kiali-ui

aljesusg:
Update REDME after merge


lucasponce (approved):
Trivial. We can keep iterating on this but I'll consolidate it.

Issue #4842, opened 2022-03-24
fatal error: concurrent map writes


FilipB:
During our perf test suite following error was thrown to kiali pod log and the pod was restarted:
```
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 775 [running]:
runtime.throw({0x1cfc0ce, 0xc000ac9ea4})
	/usr/local/go/src/runtime/panic.go:1198 +0x71 fp=0xc00157b1f0 sp=0xc00157b1c0 pc=0x4349f1
runtime.mapassign_faststr(0x1a39ce0, 0xc0003f10e0, {0x1cee197, 0xc})
	/usr/local/go/src/runtime/map_faststr.go:294 +0x38b fp=0xc00157b258 sp=0xc00157b1f0 pc=0x412fcb
github.com/kiali/kiali/status.Put(...)
	/home/fedora/go/src/github.com/kiali/kiali/status/status.go:87
github.com/kiali/kiali/status.istioVersion()
	/home/fedora/go/src/github.com/kiali/kiali/status/versions.go:264 +0x2e7 fp=0xc00157b400 sp=0xc00157b258 pc=0x15a7027
github.com/kiali/kiali/status.getVersionComponent(...)
	/home/fedora/go/src/github.com/kiali/kiali/status/versions.go:83
github.com/kiali/kiali/status.getVersions()
	/home/fedora/go/src/github.com/kiali/kiali/status/versions.go:78 +0x1b8 fp=0xc00157b4b8 sp=0xc00157b400 pc=0x15a5e58
github.com/kiali/kiali/status.Get()
```

Full pod's log is attached.
I don't have repro steps but I will try to add more details if I manage to reproduce.
[kiali-666fc6c49f-sbqpz-kiali.log](https://github.com/kiali/kiali/files/8341079/kiali-666fc6c49f-sbqpz-kiali.log)

OCP  4.10.4

Kiali UI
    1.49.0 (3c6a422344504f6771ca774bb458df4db0ac667f)
Kiali Server
    v1.49.0-SNAPSHOT (06558edfe99da98ff98f4b3baa73635f7a30e573)
Kiali Container
    v1.49.0-SNAPSHOT

Components

Istio
    1.13.2 
Prometheus
    2.31.1 
Kubernetes
    v1.23.3+e419edf

jmazzitelli:
[...]

PR #4843, opened 2022-03-24
fix: add RW lock to put status

Xunzhuo:
Signed-off-by: bitliu <bitliu@tencent.com>

** Describe the change **

fix: add RW lock to put status

** Issue reference **

resolve: https://github.com/kiali/kiali/issues/4842

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -85,7 +85,7 @@ func init() {
 
 // Put adds or replaces status info for the provided name. Any previous setting is returned.
 func Put(name, value string) (previous string, hasPrevious bool) {
-	var rw *sync.RWMutex
+	var rw sync.RWMutex
 	rw.Lock()
 	previous, hasPrevious = info.Status[name]
 	info.Status[name] = value
--- a/status/status.go
+++ b/status/status.go
@@ -86,11 +86,8 @@ func init() {
 // Put adds or replaces status info for the provided name. Any previous setting is returned.
 func Put(name, value string) (previous string, hasPrevious bool) {
 	var rw *sync.RWMutex
-	rw.RLock()
-	previous, hasPrevious = info.Status[name]
-	rw.RUnlock()
-
 	rw.Lock()
+	previous, hasPrevious = info.Status[name]
 	info.Status[name] = value
 	rw.Unlock()
 	return previous, hasPrevious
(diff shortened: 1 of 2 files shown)


jmazzitelli (changes requested):
I would consolidate the number of times you create and release a lock - unless there is a reason for why you needed to do it this way.

jmazzitelli (changes requested):
Did you actually run this? I'm getting a SEGV fault.

```
$ kubectl logs -n istio-system -l app=kiali
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x47051e]

goroutine 1 [running]:
sync.(*RWMutex).Lock(0x5e02c5)
	/home/jmazzite/bin/go-install/go/src/sync/rwmutex.go:111 +0x1e
github.com/kiali/kiali/status.Put({0x1cfcc3f, 0x15}, {0xc00023c6c0, 0x35})
	/home/jmazzite/source/kiali/kiali/status/status.go:89 +0x33
main.main()
	/home/jmazzite/source/kiali/kiali/kiali.go:88 +0x338
```

status/status.go
```diff
@@ -83,8 +85,11 @@ func init() {
 
 // Put adds or replaces status info for the provided name. Any previous setting is returned.
 func Put(name, value string) (previous string, hasPrevious bool) {
+	var rw *sync.RWMutex
```
nrfox: It seems like this will create a new mutex each time this function is called. Multiple goroutines wouldn't share the same lock but instead each have their own locks thus it wouldn't prevent writes to the `info` global var. Since `info` is a global var, I think the lock needs to be a global var as well. A simple unit test where multiple routines try to call `Put` at the same time would be good to add here to help ensure we've covered this case.

It's also better to use `defer` unlock in case there's a panic that happens to ensure that `rw.Unlock` always gets called.

jmazzitelli: I JUST realized this right when you posted this.

jmazzitelli (changes requested):
this doesn't solve the problem. "info" is global - the RW lock also needs to be global.

Use defer when unlocking to ensure that lock always gets unlocked.

status/status.go
Xunzhuo: I am goint to add an unit test here.

jmazzitelli:
testing now... code review looks good though

jmazzitelli (approved):
ran the unit tests.. passed.
built and deployed in minikube... works.

nrfox (approved):
LGTM

Issue #4849, opened 2022-03-25
unexpected ci error in TestClientExpiry


Xunzhuo:
See: https://github.com/kiali/kiali/runs/5680434347?check_suite_focus=true

the changes of the PR should not cause this CI failed.

jmazzitelli:
I've seen this failure before - we never found out why it happens. There might even already be a github issue on this.

jmazzitelli:
https://github.com/kiali/kiali/pull/2968#pullrequestreview-443706509

Xunzhuo:
I think trick should be here: https://github.com/kiali/kiali/blob/master/kubernetes/client_factory.go#L180

Xunzhuo:
Ops, I think I know what is tricky here:
the `mutex` is a global lock:
https://github.com/kiali/kiali/blob/master/kubernetes/client_factory.go#L19
In `TestClientExpiry`:
https://github.com/kiali/kiali/blob/master/kubernetes/client_factory_test.go#L14
After executing `getClientFactory`, goroutines start to run `watchClients` and they are rapidly using the global `mutex`to lock and unlock.
https://github.com/kiali/kiali/blob/master/kubernetes/client_factory.go#L181.
But in test case, it is also using the `mutex` global lock, some time, it will lead chaos, like lock failed, which will make the delete failed:
https://github.com/kiali/kiali/blob/master/kubernetes/client_factory.go#L184
That is why we see:
```
    client_factory_test.go:57: 
        	Error Trace:	client_factory_test.go:57
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 2
        	Test:       	TestClientExpiry
    client_factory_test.go:59: 
        	Error Trace:	client_factory_test.go:59
        	Error:      	Should be false
        	Test:       	TestClientExpiry
```
Because the first expired client is not properly delete from clientEntries.

Xunzhuo:
In an elegant way, we should use channel to tell goroutine to recycle this client, not for sleep and read the cliententries.

lucasponce:
> But in test case, it is also using the mutex global lock, some time, it will lead chaos, like lock failed or conflicts, which will make the delete expired client failed

I've commented in the PR, perhaps the strategy for this work could be to improve the test, but moving a global lock to a channel because a test shouldn't be the first option.

nrfox:
This continues to plague CI and should be added to the backlog.

nrfox:
[...]

PR #4855, opened 2022-03-25
feat: provide recycleChan to clean expired clients

Xunzhuo:

** Describe the change **

provide recycleChan to clean expired clients

** Issue reference **

related to https://github.com/kiali/kiali/issues/4849

### DIFF
--- a/kubernetes/client_factory.go
+++ b/kubernetes/client_factory.go
@@ -15,10 +15,14 @@ import (
 
 var factory *clientFactory
 
-// Mutex for when modifying the stored clients
+// mutex for when modifying the stored clients
 var mutex = &sync.RWMutex{}
 
-const expirationTime = time.Minute * 15
+// when a client is expired, a signal with its tokenHash will be sent to recycleChan
+var recycleChan = make(chan string, 10)
+
+// defaultExpirationTime set the default expired time of a client
+const defaultExpirationTime = time.Minute * 15
 
 // ClientFactory interface for the clientFactory object
 type ClientFactory interface {
@@ -29,13 +33,7 @@ type ClientFactory interface {
 type clientFactory struct {
 	ClientFactory
 	baseIstioConfig *rest.Config
-	clientEntries   map[string]*clientEntry
-}
-
-// clientEntry stored the client and its created timestamp
-type clientEntry struct {
-	client  ClientInterface
-	created time.Time
+	clientEntries   map[string]*ClientInterface
 }
 
 // GetClientFactory returns the client factory. Creates a new one if necessary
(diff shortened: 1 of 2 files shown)


Xunzhuo:
One more questions need to be confirmed, is it a good idea to let user to configure the expiration time?

lucasponce:
@Xunzhuo I've read and reviewed the PR several times.
I don't still see the benefit of this approach compared with the current design.

What penalty will cause to review all clients in the "watchClients" in a scenario with few clients?

Perhaps do you have any other scenario that we are not seeing and it's causing a problem that would need this approach?

Without more info that justifies it, I'm not sure this one could progress.

General comment, the code base of Kiali can be optimized (no doubt on that) but typically it's driven by some business/priorities; and also some tradeoffs are considered unless there is a clear penalty or feature that we want to address.

Xunzhuo:
> @Xunzhuo I've read and reviewed the PR several times. I don't still see the benefit of this approach compared with the current design.
> 
> What penalty will cause to review all clients in the "watchClients" in a scenario with few clients?
> 
> Perhaps do you have any other scenario that we are not seeing and it's causing a problem that would need this approach?
> 
> Without more info that justifies it, I'm not sure this one could progress.
> 
> General comment, the code base of Kiali can be optimized (no doubt on that) but typically it's driven by some business/priorities; and also some tradeoffs are considered unless there is a clear penalty or feature that we want to address.

Hi @lucasponce , sorry for replying lately for busy jobs recently.
The reason I want to refactor this is both to optimize the logic and meet some specific scenario:
For now, I believe that the expiration time is a const and can not be configured, but sometimes the expiration time should be able to be configured, for example, in some clusters, we should expire the client more frequently for some dynamically changes with apiserver, and in some clusters, we may do not want kiali to expire and create clients frequently.
So I think this is the reason that I do the refactoring in this area.

lucasponce:
Please @Xunzhuo, would you mind rebasing this one with the master branch? Thanks

Xunzhuo:
@lucasponce  sure~

lucasponce (dismissed):
I've reviewed and tested the proposal.
I think we can accept this change if it helps other scenarios where @Xunzhuo is working.

lucasponce:
Sorry to ask you again @Xunzhuo, would you mind rebasing again?
I'm fine consolidating this work and checking that there are no other use cases that may affect it.

The project is in a transition phase where we added new CI pipelines for the UI and those are not yet 100% stable and it would require updating the branches.

Thanks!

lucasponce (approved):
Re-approved.
The UI test failing is unrelated to this work and it's being addressed in a separate effort.

Xunzhuo:
> Re-approved.
> 
> The UI test failing is unrelated to this work and it's being addressed in a separate effort.

@lucasponce thanks!

Issue #4860, opened 2022-03-26
LICENSE Validation


Xunzhuo:
https://github.com/kiali/kiali/blob/1ccc1d02c71d82232ca0b0c75bde1f5a649e50a6/LICENSE#L189

Read through the Kiali license file, it seems like it is lack of validation, we need to update the copyright message to let the Apache License 2.0 valid :)

PR #4861, opened 2022-03-27
fix: license copyright

Xunzhuo:
resolve: https://github.com/kiali/kiali/issues/4860


lucasponce (approved):
It looks ok.
Thanks! It was a templated that was never updated.

PR #4866, opened 2022-03-28
[make] provide a make target to allow for an easy way to "yarn start"

jmazzitelli:
This is for developers who are not familiar with (or who forget :) the proxy setting and where/how to set it in the package.json. You just have to provide your server URL via `-e YARN_START_URL` such as `make -e YARN_START_URL="http://localhost:20001" yarn-start`. Once that updates the package.json it will invoke `yarn start`. You won't have to set that env var thereafter unless you reset package.json. If the package.json already has a proxy setting, the env var is ignored. You can just run `make yarn-start` in that case.

Here is how to test. Get a cluster up with istio installed. Then run the "run-kiali.sh" hack script to start the server.  Then run this make target to start a ui via yarn start.  At this point, you have a development environment of both the backend (you can use a debugger to connect to it) and of the frontend (you can edit the frontend UI files and have them automatically recompiled/redployed in the browser). Steps are:

1. `./hack/run-kiali.sh -kc current`
2. `make -e YARN_START_URL="http://localhost:20001" yarn-start`

Tell yarn that you want to start it on another port (the backend run-kiali.sh script will have already started a proxy on port 3000 for grafana, so yarn needs to use another port). At this point, the browser can access the UI at `http://localhost:3001`.

NOTE: you don't have to use run-kiali.sh - I just show how you can use hack scripts and make to stand up a development version of both the server and UI without deploying anything Kiali-related to the cluster. If you already have kiali deployed somewhere, use its URL for the YARN_START_URL setting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,7 @@ include make/Makefile.cluster.mk
 include make/Makefile.helm.mk
 include make/Makefile.operator.mk
 include make/Makefile.molecule.mk
+include make/Makefile.ui.mk
 
 .PHONY: help
 help: Makefile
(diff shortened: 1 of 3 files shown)


jshaughn:
Perhaps all that info in the description should be somewhere in the docs, like maybe https://github.com/kiali/kiali/blob/master/README.adoc#running-the-ui-outside-the-cluster

jmazzitelli:
> Perhaps all that info in the description should be somewhere in the docs, like maybe https://github.com/kiali/kiali/blob/master/README.adoc#running-the-ui-outside-the-cluster

I updated that, and also fixed the bad URL that was already there.

jshaughn (approved):
LOK

Issue #4826, opened 2022-03-22
Migrate e2e test suite to Golang


lucasponce:
This is a follow-up issue related to the main discussion we are having in the project to improve our testing and flows.

One area discussed is to improve the maintenance of the e2e test suite mainly maintained by the QE team.

This e2e suite (located under the "kiali/tests/e2e" folder) runs a set of tests against a live Kiali server APIs.

These APIs are considered private, as they serve only to the Kiali UI and they can change and break any contract without restriction.

So, when there are changes in the APIs, those tests should be updated at the same time and ideally be tested together with the regular unit testing (and in the pipelines triggered by PR).

This issue proposes the following tasks:

- Migrate this suite to Golang, keeping all the scenarios tested, but developers would be active in maintaining this suite.
- Using Golang (in comparison with Python) will have several benefits, both from a technical perspective and people-related.
- From a technical perspective the "marshaling" will be simplified as the test responses can be converted to the expected types, which will reduce and simplify the test migration without missing any feature. 
- Also the backend language is Golang and it will help that all the groups contribute actively to them.
- The tests should keep all the features and should be valid for "upstream" scenarios and "downstream" scenarios.

It's important that QE validates that the e2e test suite is re-usable in all the testing scenarios, goal here is to unify a single test suite for endpoints testing.

lucasponce:
Side comment, that perhaps is related to https://github.com/kiali/kiali/issues/4827.
We need a "lightweight" job that may populate a cluster but can start an external backend Kiali process to test the e2e upstream.
The current https://github.com/kiali/kiali/runs/5475832549?check_suite_focus=true is good but it's covering the whole process of building and deploying.

For the e2e we could have a similar scenario as used in dev mode, a process connected by a proxy that can execute the suite for most of the features (to improve the 1h to something that can be executed per PR granularity).

Note, the molecule job is good, and we don't plan to change it, this is more an addition to improve the automation of the e2e testing.

mattmahoneyrh:
QE Requirements for e2es:
 - Need to be integrated with QE jenkins pipelines, both upstream and downstream
 - Be able to be run against OSSM downstream
 - Results will need to be uploaded into Polarion, using a junit type output format
 - Be able to run against already provisioned cluster with already deployed test stack (ex: istio + kiali + bookinfo already deployed)

jmazzitelli:
> * Need to be integrated with QE jenkins pipelines, both upstream

[...]

PR #4868, opened 2022-03-29
Converting e2e tests into Integration tests for Kiali, initial commit with Kiali Status test for CI.

hhovsepy:
RFE: https://github.com/kiali/kiali/issues/4826

For running Integration tests, URL and TOKEN env variables are required.
URL in "https://kiali-hostname" and TOKEN in "sha256~" format.

Also Integration tests can be run separately by: "make test-integration URL="$kialiurl" TOKEN="$token"" and it will export result in "tests/integration/junit-rest-report.xml"

### DIFF
--- a/make/Makefile.build.mk
+++ b/make/Makefile.build.mk
@@ -98,7 +98,7 @@ test-integration-setup:
 	go install github.com/jstemmer/go-junit-report@latest
 
 ## test-integration: Run Integration test suite
-test-integration:
+test-integration: test-integration-setup
 	@echo Running Integration tests
 	cd tests/integration/tests && ${GO} test -v 2>&1 | go-junit-report > ../junit-rest-report.xml
 
--- a/make/Makefile.build.mk
+++ b/make/Makefile.build.mk
@@ -70,17 +70,17 @@ build-system-test:
 ## test: Run tests, excluding third party tests under vendor and frontend. Runs `go test` internally
 test:
 	@echo Running tests, excluding third party tests under vendor
-	${GO} test $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/ -e /tests/integration)
+	${GO} test $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/ -e /tests/integration/)
 
 ## test-debug: Run tests in debug mode, excluding third party tests under vendor and frontend. Runs `go test -v`
 test-debug:
 	@echo Running tests in debug mode, excluding third party tests under vendor
-	${GO} test -v $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/)
+	${GO} test -v $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/ -e /tests/integration/)
 
 ## test-race: Run tests with race detection, excluding third party tests under vendor and frontend. Runs `go test -race`
 test-race:
 	@echo Running tests with race detection, excluding third party tests under vendor
-	${GO} test -race $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/)
+	${GO} test -race $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/ -e /tests/integration/)
 
 ## test-e2e-setup: Setup Python environment for running test suite
 test-e2e-setup:
--- a/make/Makefile.build.mk
+++ b/make/Makefile.build.mk
@@ -70,7 +70,7 @@ build-system-test:
 ## test: Run tests, excluding third party tests under vendor and frontend. Runs `go test` internally
 test:
 	@echo Running tests, excluding third party tests under vendor
-	${GO} test $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/)
+	${GO} test $(shell ${GO} list ./... | grep -v -e /vendor/ -e /frontend/ -e /tests/integration)
 
 ## test-debug: Run tests in debug mode, excluding third party tests under vendor and frontend. Runs `go test -v`
 test-debug:
--- a/make/Makefile.build.mk
+++ b/make/Makefile.build.mk
@@ -100,7 +100,7 @@ test-integration-setup:
 ## test-integration: Run Integration test suite
 test-integration:
 	@echo Running Integration tests
-	cd tests/integration && ${GO} test -v 2>&1 | go-junit-report > junit-rest-report.xml
+	cd tests/integration/tests && ${GO} test -v 2>&1 | go-junit-report > ../junit-rest-report.xml
 
 #
 # Swagger Documentation
--- a/make/Makefile.build.mk
+++ b/make/Makefile.build.mk
@@ -101,6 +101,7 @@ test-integration-setup:
 test-integration: test-integration-setup
 	@echo Running Integration tests
 	cd tests/integration/tests && ${GO} test -v 2>&1 | go-junit-report > ../junit-rest-report.xml
+	@echo Test results can be found here: $$(ls -1 ${ROOTDIR}/tests/integration/junit-rest-report.xml)
 
 #
 # Swagger Documentation
--- a/make/Makefile.build.mk
+++ b/make/Makefile.build.mk
@@ -92,6 +92,16 @@ test-e2e:
 	@echo Running E2E tests
 	cd tests/e2e && source .kiali-e2e/bin/activate && pytest -s tests/
 
+## test-integration-setup: Setup go library for converting test result into junit xml
+test-integration-setup:
+	@echo Setting up Integration tests
+	go install github.com/jstemmer/go-junit-report@latest
+
+## test-integration: Run Integration test suite
+test-integration:
+	@echo Running Integration tests
+	cd tests/integration && ${GO} test -v 2>&1 | go-junit-report > junit-rest-report.xml
+
 #
 # Swagger Documentation
 #
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -25,17 +25,14 @@
 # Clone this repository
 $ git clone https://github.com/kiali/kiali.git
 
-# Install requirements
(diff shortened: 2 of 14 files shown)


make/Makefile.build.mk
```diff
@@ -92,6 +92,16 @@ test-e2e:
 	@echo Running E2E tests
 	cd tests/e2e && source .kiali-e2e/bin/activate && pytest -s tests/
 
+## test-integration-setup: Setup go library for converting test result into junit xml
+test-integration-setup:
+	@echo Setting up Integration tests
+	go install github.com/jstemmer/go-junit-report@latest
+
+## test-integration: Run Integration test suite
+test-integration:
```
jmazzitelli, line 101: I think this should always do the setup. Since the setup is going to pull `@latest` we are going to want to always install this to make sure we have the latest (and thus make sure the integration tests still work with the latest)? So:
```suggestion
test-integration: test-integration-setup
```

tests/integration/README.md
```diff
@@ -0,0 +1,47 @@
...
+$ make test-integration-setup
+
+# Login to Openshift Cluster 
+$ oc login https://<openshift>:<port> --username=<username> --password=<password> --insecure-skip-tls-verify=true
+
+# Read the Token from Openshift Cluster, to be used in next step or running tests
+$ oc whoami -t
+
+# run all tests
+$ make test-integration URL="https://<kiali-hostname>" TOKEN="<token>"
```
jmazzitelli, line 38: you can combine these so the user doesn't have to manually cut-n-paste the token to pass to the TOKEN env var:
```suggestion
# run all tests authenticating as the user that was logged in
$ make test-integration URL="https://<kiali-hostname>" TOKEN="$(oc whoami -t)"
```

tests/integration/README.md
```diff
@@ -0,0 +1,47 @@
...
+## Environment Setup
+
+* System dependencies that will need to be available
+    *  `oc`
+    *  `go`
+    *  `make`
+    *  `npm`
+    *  `yarn`
+
+* It is expected that the following have already been deployed in Openshift
```
jmazzitelli, line 19: Is there a reason this README is geared towards OpenShift? I do not see instructions here on how to run these integration tests on non-OpenShift environments (like minikube). We should have the ability to run these i-tests on an OpenShift and a non-OpenShift cluster - instructions on how to do both are needed.

hhovsepy: so far oc is not required, will remove it for now

jmazzitelli (changes requested):
need `make test` to not run the integration tests - the unit tests should not require a cluster.
```
{"level":"fatal","time":"2022-03-31T08:56:30-04:00","message":"URL and TOKEN environment variables are required. Kiali URL in 'https://kiali-hostname' and TOKEN in 'sha256~' format."}
FAIL	github.com/kiali/kiali/tests/integration/tests	0.027s

```

util/httputil/http_util.go
```diff
@@ -23,12 +23,16 @@ func HttpMethods() []string {
 		http.MethodDelete, http.MethodConnect, http.MethodOptions, http.MethodTrace}
 }
 
-func HttpGet(url string, auth *config.Auth, timeout time.Duration, customHeaders map[string]string) ([]byte, int, error) {
+func HttpGet(url string, auth *config.Auth, timeout time.Duration, customHeaders map[string]string, cookies []*http.Cookie) ([]byte, int, error) {
```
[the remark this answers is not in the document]
nrfox: See my above comment about global vars and having a separate struct as that might also eliminate the need to add extra args here.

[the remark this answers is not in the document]
jmazzitelli: I'm actually ok with this - I'd rather have one line than build a request with a bunch of lines - makes the code smaller, easier to read. People can still use the stdlib if they want. No need for us to write std lib v2.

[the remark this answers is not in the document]
hhovsepy: I would leave that as a separate task. Anyway there is upcoming refactoring of httputil.

jmazzitelli (dismissed):
my requests in a PR: https://github.com/hhovsepy/kiali/pull/1

tests/integration/util/kiali_client.go
```diff
@@ -0,0 +1,64 @@
...
+		log.Fatalf("URL and TOKEN environment variables are required. Kiali URL in 'https://kiali-hostname' and TOKEN in 'sha256~' format.")
+		return
+	}
+	tokenResult, tokenCookies := GetCookies()
+	if !tokenResult || tokenCookies == nil {
+		log.Fatalf("Unable to login to the Kiali: %s by provided token: %s", kialiURL, kialiToken)
+		return
+	}
+	kialiCookies = tokenCookies
+}
```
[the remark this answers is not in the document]
jmazzitelli: it doesn't have to be. We can pass in a token using the Kiali service account - using that, we can go through, say, header-auth strategy.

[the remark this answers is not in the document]
jmazzitelli: See my new PR with instructions on how to pass in a token of the kiali service account (we will do this when running on non-openshift environments):
```
$ make test-integration -e URL="http://localhost:20001/kiali" \
   -e TOKEN="$(kubectl get -n istio-system $(kubectl get secret -n istio-system -o name | grep 'kiali.*-token' | head -n 1) -o jsonpath={.data.token} | base64 -d)"
```

hhovsepy:
thank you @jmazzitelli , have merged your PR

jmazzitelli (dismissed):
This works for me. Though, the TOKEN doesn't seem to be needed (at least not on minikube). I'll assume it will be used it other tests when we add them later.

tests/integration/utils/kiali_client.go
```diff
@@ -0,0 +1,64 @@
...
+var kialiCookies []*http.Cookie
+
+func init() {
+	kialiURL = os.Getenv("URL")
+	kialiToken = os.Getenv("TOKEN")
+	if kialiURL == "" || kialiToken == "" {
+		log.Fatalf("URL and TOKEN environment variables are required. Kiali URL in 'https://kiali-hostname' and TOKEN in 'sha256~' format.")
+		return
+	}
+	tokenResult, tokenCookies := GetCookies()
```
israel-hdez, line 26: I am probably chiming in a little late, but I was expecting closer integration into the Kiali codebase.

Personally, I was expecting more to see tests taking the form of these ones, where you don't really need to deploy Kiali to the cluster: https://github.com/kiali/kiali/blob/master/handlers/metrics_test.go.

So, you would do a direct call to the handler bypassing all that auth machinery, which lets you directly test targeted handler, and also would let you mock any 3rd party services and would remove the need to have a full stack.

The need of the full stack, and also the fact that the API is tested through an HTTP client, gives me the idea that this would be closer to the "system tests" side that the "integration tests" side. But "system tests" is what cypress would be covering, isn't it?

I don't know... at the current state, I'm expecting to see a lot of coverage duplicity between these go tests and cypress tests :-/

israel-hdez: OK, it was my expectations. You have more experience than I about the scope of QE testing if this is going to help on Dev-QE synergy, it's quite OK.

jmazzitelli: As a side comment, I HATE having to write all that mock code that we have to do today. A dev might end up spending more time writing mock code compared to "real" code.

nrfox:
The tests fail for me when running against my local env without auth enabled:
```
make -e URL=$(kiali_minikube_url) test-integration
```

It'd be nice to be able to run these against any kiali env instead of just one that requires token auth.

tests/integration/utils/kiali_client.go
israel-hdez: > @jmazzitelli  As a side comment, I HATE having to write all that mock code that we have to do today. A dev might end up spending more time writing mock code compared to "real" code.

I do agree. I think sometimes it is neccessary evil. Otherwise, how would you achive modest coverage (i.e. test something different than the happy path)?

> @nrfox These backend api e2e tests would be owned by the dev team since we maintain the kiali api. The effort here is just to port those python tests into golang because the dev team is more familiar with golang than python.

Well... If these go test are going to be owned by the Kiali team, I *do* prefer these to take proper form of integration tests rather than being closer to systems tests :-/

Personally, I envision devs to be doing quite a lot of mocks to achive proper code coverage, while QE-owned tests take care of testing with the full stack.

Just my $0.02. Although, I probably already missed my chance to have any influence here...

hhovsepy:
Going to add support of no auth support as well, based on "/api/auth/info". So TOKEN would be not necessary in those cases.

tests/integration/utils/kiali_client.go
israel-hdez: > As a side comment, I HATE having to write all that mock code that we have to do today. A dev might end up spending more time writing mock code compared to "real" code.

BTW, look at this: https://github.com/dnaeon/go-vcr
It makes me remember... oh... well... :smile:

lucasponce:
> It'd be nice to be able to run these against any kiali env instead of just one that requires token auth.

Yes. It should cover both cases auth and anonymous.

Also, it should cover any deployment model (i.e. a deployed Kiali in the cluster or a dev environment where there is a local URL pointed to a proxy).

I haven't tested it yet, it seems it will cover both scenarios, but just leaving a note here :).

hhovsepy:
> The tests fail for me when running against my local env without auth enabled:
> 
> ```
> make -e URL=$(kiali_minikube_url) test-integration
> ```
> 
> It'd be nice to be able to run these against any kiali env instead of just one that requires token auth.

Can you retry the tests on you env @nrfox ?

nrfox (approved):
Tests are working for me now against my local env with no auth

hhovsepy:
Merging this PR with simple testing scenario, not to block CI tasks.
More test scenarios will come in further PRs.

hhovsepy:
thanks Nick and mazz for the review

Issue #4872, opened 2022-03-30
Add cypress UI tests around the Overview page


lucasponce:
Subtask for https://github.com/kiali/kiali/issues/4854 Epic

PR #4875, opened 2022-03-30
[cypress] Overview page

lucasponce:
Fixes https://github.com/kiali/kiali/issues/4872

An early draft of basic UI tests, just to give visibility and get feedback about how we want to write these tests.

What I'd try to do here is:
- Take the assumption that the integration test cluster has a set of demos deployed.
- The cypress test will check:
  - Namespaces exist.
  - User can filter a specific namespace.
  - User can change the view of the namespaces.
  - User can interact with the namespace.

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -134,6 +134,9 @@ jobs:
       - name: Setup kind
         run: hack/setup-kind-in-ci.sh
 
+      - name: Setup Error Rates demo
+        run: hack/istio/install-error-rates-demo.sh -c "kubectl"
+
       - name: Set kiali URL
         run: |
           KIALI_URL="http://$(kubectl get svc kiali -n istio-system -o=jsonpath='{.status.loadBalancer.ingress[0].ip}'):20001/kiali"
--- a/frontend/cypress/integration/common/overview.ts
+++ b/frontend/cypress/integration/common/overview.ts
@@ -1,0 +1,130 @@
+import { Before, Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+
+const url = "/console";
+
+Before(() => {
+    // Focing to not stop cypress on unexpected errors not related to the tests.
+    // There are some random failures due timeouts/loadtime/framework that throws some error in the browser.
+    // After reviewing the tests failures, those are unrelated to the app, so,
+    // it needs this event to not fail the CI action due some "slow" action or similar.
+    // This is something to review in future iterations when tests are solid, but I haven't found a better way to
+    // solve this issue.
+    cy.on("uncaught:exception", (err, runnable, promise) => {
+        // when the exception originated from an unhandled promise
+        // rejection, the promise is provided as a third argument
+        // you can turn off failing the test in this case
+        if (promise) {
+            return false
+        }
+        // we still want to ensure there are no other unexpected
+        // errors, so we let them fail the test
+    });
+});
+
+Given('user opens the overview page', () => {
+    // Forcing "Pause" to not cause unhandled promises from the browser when cypress is testing
+    cy.visit(url + '/overview?refresh=0');
+});
+
+When('user clicks in the {string} view', (view) => {
+    cy.get('button[data-display-mode="' + view + '"]').click();
+});
+
+When(`user filters {string} namespace`, (ns) => {
+    cy.get('select[aria-label="filter_select_type"]')
+        .select('Namespace')
+        .should('have.value', 'namespace_search');
+    cy.get('input[aria-label="filter_input_value"]')
+        .type(ns)
+        .type('{enter}');
+});
+
+When(`user filters {string} health`, (health) => {
+    cy.get('select[aria-label="filter_select_type"]')
+        .select('Health')
+        .should('have.value', 'health');
+    cy.get('select[aria-label="filter_select_value"]')
+        .select(health);
+});
+
+When(`user selects Health for {string}`, (type) => {
+    let innerId = '';
+    switch (type) {
+        case 'Apps':
+            innerId = 'app';
+            break;
+        case 'Workloads':
+            innerId = 'workload';
+            break;
+        case 'Services':
+            innerId = 'service';
+            break;
+    }
+    cy.get('button[aria-labelledby^="overview-type"]')
+        .click()
+        .get('button[id^="' + innerId + '"]')
+        .click();
+});
+
+When(`user sorts by name desc`, () => {
+    cy.get('button[data-sort-asc="true"]')
+        .click();
+});
+
+When(`user selects {string} time range`, (interval) => {
+    let innerId = '';
+    switch (interval) {
+        case 'Last 10m':
+            innerId = '600';
+            break;
+    }
+    cy.get('button[aria-labelledby^="time_range_duration"]')
+        .click()
+        .get('button[id^="' + innerId + '"]')
+        .click();
+});
+
+Then(`user sees the {string} namespace`, (ns) => {
+    cy.get('article[data-namespace="' + ns + '"]');
+});
+
+Then(`user doesn't see the {string} namespace`, (ns) => {
+    cy.get('article[data-namespace="' + ns + '"]').should('not.exist');
+});
+
+Then(`user sees a {string} {string} namespace`, (view, ns) => {
+    if (view === "LIST") {
+        cy.get('td[role="gridcell"]').contains(ns);
+    } else {
+        cy.get('article[data-namespace="' + ns + '"][data-display-mode="' + view + '"]');
+    }
+});
+
+Then(`user sees the {string} namespace with {string}`, (ns, type) => {
+    let innerType = '';
+    switch (type) {
+        case 'Applications':
+            innerType = 'app';
+            break;
+        case 'Workloads':
+            innerType = 'workload';
+            break;
+        case 'Services':
+            innerType = 'service';
+            break;
+    }
+    cy.get('article[data-namespace="' + ns + '"]').find('[data-overview-type="' + innerType + '"]');
+});
+
+Then(`user sees the {string} namespace list`, (nslist) => {
+    const nss = nslist.split(',');
+    cy.get('article')
+        .should('have.length', nss.length)
+        .each(($a, i) => {
+            expect($a.attr("data-namespace")).be.eq(nss[i]);
+        });
+});
+
+Then(`user sees the {string} namespace with Inbound traffic {string}`, (ns, duration) => {
+    cy.get('article[data-namespace="' + ns + '"]').find('span[data-sparkline-duration="' + duration + '"]');
+});
--- a/frontend/cypress/integration/featureFiles/overview.feature
+++ b/frontend/cypress/integration/featureFiles/overview.feature
@@ -1,0 +1,71 @@
+Feature: Kiali Overview page
+
+  User opens the Overview page and see the demo "error-rates" namespaces
+
+  Background:
+    Given user is at administrator perspective
+    And user opens the overview page
+
+  @overview-page
+  Scenario: See "alpha" and "beta" namespaces
+    Then user sees the "alpha" namespace
+    And user sees the "beta" namespace
+
+  @overview-page
+  Scenario: Doesn't see a "bad" namespace
+    Then user doesn't see the "bad" namespace
+
+  @overview-page
+  Scenario: Select the COMPACT view
+    When user clicks in the "COMPACT" view
+    Then user sees a "COMPACT" "alpha" namespace
+
+  @overview-page
+  Scenario: Select the EXPAND view
+    When user clicks in the "EXPAND" view
+    Then user sees a "EXPAND" "beta" namespace
+
+  @overview-page
+  Scenario: Select the LIST view
+    When user clicks in the "LIST" view
+    Then user sees a "LIST" "beta" namespace
+
+  @overview-page
+  Scenario: Filter by namespace
+    When user filters "alpha" namespace
+    Then user sees the "alpha" namespace
+    And user doesn't see the "beta" namespace
+
+  @overview-page
+  Scenario: Filter by health
+    When user filters "Failure" health
+    Then user sees the "alpha" namespace
+    And user sees the "beta" namespace
+    And user doesn't see the "default" namespace
+
+  @overview-page
+  Scenario: Sort by name
+    When user filters "Failure" health
+    And user sorts by name desc
+    Then user sees the "beta,alpha" namespace list
+
+  @overview-page
+  Scenario: Health for Apps
+    When user selects Health for "Apps"
+    Then user sees the "alpha" namespace with "Applications"
+
+  @overview-page
+  Scenario: Health for Workloads
+    When user selects Health for "Workloads"
+    Then user sees the "alpha" namespace with "Workloads"
+
+  @overview-page
+  Scenario: Health for Services
+    When user selects Health for "Services"
+    Then user sees the "alpha" namespace with "Services"
+
+  @overview-page
+  Scenario: Last 10 minutes
+    When user selects "Last 10m" time range
+    Then user sees the "alpha" namespace with Inbound traffic "10m"
+
--- a/frontend/src/pages/Overview/OverviewPage.tsx
+++ b/frontend/src/pages/Overview/OverviewPage.tsx
@@ -744,7 +744,7 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
                   const isLongNs = ns.name.length > NS_LONG;
                   return (
                     <GridItem sm={sm} md={md} key={'CardItem_' + ns.name} style={{ margin: '0px 5px 0 5px' }}>
-                      <Card isCompact={true} className={cardGridStyle}>
+                      <Card isCompact={true} className={cardGridStyle} data-namespace={ns.name} data-display-mode={OverviewDisplayMode[this.state.displayMode]}>
                         <CardHead>
                           <CardActions>{namespaceActions[i]}</CardActions>
                           <CardHeader className={cardHeaderStyle}>
@@ -840,7 +840,14 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
   renderStatuses(ns: NamespaceInfo): JSX.Element {
     if (ns.status) {
       if (this.state.displayMode === OverviewDisplayMode.COMPACT) {
-        return <OverviewCardContentCompact key={ns.name} name={ns.name} status={ns.status} type={this.state.type} />;
+        return (
+          <OverviewCardContentCompact
+            key={ns.name}
+            name={ns.name}
+            status={ns.status}
+            type={this.state.type}
+          />
+        );
       }
       return (
         <OverviewCardContentExpanded
--- a/frontend/src/pages/Overview/OverviewCardContentCompact.tsx
+++ b/frontend/src/pages/Overview/OverviewCardContentCompact.tsx
@@ -33,7 +33,7 @@ class OverviewCardContentCompact extends React.Component<Props> {
       <>
         <div style={{ textAlign: 'left' }}>
           <span>
-            <div style={{ display: 'inline-block', width: '125px' }}>{text}</div>
+            <div style={{ display: 'inline-block', width: '125px' }} data-overview-type={this.props.type}>{text}</div>
             <div style={{ display: 'inline-block' }}>
               {status.inNotReady.length > 0 && (
                 <OverviewStatus
--- a/frontend/src/pages/Overview/OverviewCardContentExpanded.tsx
+++ b/frontend/src/pages/Overview/OverviewCardContentExpanded.tsx
@@ -56,7 +56,7 @@ class OverviewCardContentExpanded extends React.Component<Props> {
     } else {
       text = nbItems + switchType(this.props.type, ' Applications', ' Services', ' Workloads');
     }
-    const mainLink = <div style={{ display: 'inline-block', width: '125px', whiteSpace: 'nowrap' }}>{text}</div>;
+    const mainLink = <div style={{ display: 'inline-block', width: '125px', whiteSpace: 'nowrap' }} data-overview-type={this.props.type}>{text}</div>;
     if (nbItems === status.notAvailable.length) {
       return (
         <div style={{ textAlign: 'left' }}>
--- a/frontend/src/pages/Overview/OverviewCardSparkline.tsx
+++ b/frontend/src/pages/Overview/OverviewCardSparkline.tsx
@@ -42,7 +42,7 @@ class OverviewCardSparkline extends React.Component<Props, {}> {
 
       return (
         <>
-          {'Inbound traffic, ' + getName(this.props.duration).toLowerCase()}
+          <span data-sparkline-duration={getName(this.props.duration).toLowerCase()}>{'Inbound traffic, ' + getName(this.props.duration).toLowerCase()}</span>
           <SparklineChart
             name={'traffic'}
             height={60}
--- a/frontend/src/pages/Overview/OverviewToolbar.tsx
+++ b/frontend/src/pages/Overview/OverviewToolbar.tsx
@@ -174,8 +174,9 @@ export class OverviewToolbar extends React.Component<Props, State> {
               value={this.state.sortField.id}
               label={sortTypes[this.state.overviewType]}
               options={sortTypes}
+              data-sort-field={this.state.sortField.id}
             />
-            <Button variant="plain" onClick={this.updateSortDirection} style={{ ...ThinStyle }}>
+            <Button variant="plain" onClick={this.updateSortDirection} style={{ ...ThinStyle }} data-sort-asc={this.state.isSortAscending}>
               {this.state.isSortAscending ? <SortAlphaDownIcon /> : <SortAlphaUpIcon />}
             </Button>
           </>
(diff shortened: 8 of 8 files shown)


frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -744,7 +744,7 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
                   const isLongNs = ns.name.length > NS_LONG;
                   return (
                     <GridItem sm={sm} md={md} key={'CardItem_' + ns.name} style={{ margin: '0px 5px 0 5px' }}>
-                      <Card isCompact={true} className={cardGridStyle}>
+                      <Card isCompact={true} className={cardGridStyle} data-namespace={ns.name}>
```
lucasponce, line 747: Note that the code will need to add more of these IDs to help make cleaner UI tests.
In the past QE asked Dev to add these ids, but any QE eng is free to add these IDs if that helps to simplify the selectors or not to iterate on the DOM to find a specific element.

In most of the cases those are transparent to the logic.

nrfox: +1 more of these should be added in the future to aid testing and they can be added by whoever is writing the test

frontend/cypress/integration/common/overview.ts
```diff
@@ -0,0 +1,26 @@
+import { And, Given, Then } from "cypress-cucumber-preprocessor/steps";
+
+const url = "/";
+
+/*
+    TODO Prerequisites
+
+    This test assumes that integration test cluster has installed the "errors-rate" demo app
+    https://github.com/kiali/demos/tree/master/error-rates#quick-start
+*/
```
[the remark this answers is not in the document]
jmazzitelli: just make sure whatever way we choose to install, that it has to support installing on upstream istio and on maistra/openshift (I believe the scripts in hack/ do already support installing on Openshift/maistra)

frontend/src/pages/Overview/OverviewPage.tsx
lucasponce: Yes, I'd say that when let's use it when writing the test, just to avoid that some selector/expression is complex due to a lack of data-* (i.e. "find a div with a class, search the first element of a list, iterate until the proper sibling, etc", if a "data-*" would simplify that, it's something we can use without worries).

frontend/cypress/integration/common/overview.ts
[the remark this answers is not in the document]
lucasponce: Good comment.
Well, I think we should use the "kiali/demos" for testing.
Those are designed to cover almost all the scenarios that Kiali has features (HTTP, grpc traffic), fixed error rates, all combinations of Gateways/ServiceEntries, runtimes metrics, etc.
If some scenarios are not present, let's update the demos.

I'm not convinced that the UI test should install the demo, I put a comment in the code to make it open or discuss.

I'd expect that the pipeline or a hack/script, if we can avoid the interaction from the cypress in systems tasks, perhaps may help.

But again, this is just a draft for learning and syncing in the same way to write thing, as you can see the tests are very minimal and those can be adapted to any flow.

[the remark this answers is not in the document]
nrfox: > I'd expect that the pipeline or a hack/script, if we can avoid the interaction from the cypress in systems tasks, perhaps may help.

Yes if we're deploying the demos, it is probably better to setup those outside of the cypress tests since those live in a separate repo.

lucasponce:
The "overview.feature" and "overview.ts" have some common use-cases that can be re-used in other tests.

Perhaps no need to make a "generic" function, but I hope they help them to easily copy/paste and check how to add a "click()", select a "dropdown" or populate a "filter".

hack/setup-demos-ci.sh
```diff
@@ -0,0 +1,60 @@
...
+HELP
+}
+
+# abort on any error
+set -e
+
+# set up some of our defaults
+
+infomsg "Make sure everything exists"
+which kubectl > /dev/null || (infomsg "kubectl executable is missing"; exit 1)
```
nrfox, line 23: To be able to reuse this same hack script for testing on openshift, this hack script should work with `oc` as well.

lucasponce:
Adding more reviewers to this work.
The idea is to progress in the pipelines and in the style of testing in the UI.

hack/setup-demos-ci.sh
lucasponce: There is a specific hack for downstream:
```
hack/istio/install-error-rates-demo.sh
```

As the pipelines won't be the same, the downstream pipeline can just use that script I think.

(In the same way that the kind script is only used for upstream pipelines).

But tests should work in both environments.

lucasponce: I'm reviewing that hack script and it's prepared for both environments.
I'm going to test it and update the PR.

nrfox: If the `install-error-rates-demo` script already does all the necessary setup for both upstream/downstream, why not just call that hack script from this hack script? It looks like `install-error-rates-demo` will work for both.

> As the pipelines won't be the same, the downstream pipeline can just use that script I think.

What the environment is running on will be different, I don't think downstream runs on kind, but the environment will be roughly the same. Openshift might require a couple extra objects to be created like the network definitions but it looks like the `install-error-rates-demo` already handles that. If we can use the same setup scripts for both environments, we're less likely to diverge between the two.

lucasponce: PR updated.
Using the `install-error-rates-demo.sh` script and it works well, no need to duplicate this hack.
I've removed the new `setup-demos-ci.sh` one.

frontend/cypress/integration/common/overview.ts
```diff
@@ -0,0 +1,135 @@
...
+    // it needs this event to not fail the CI action due some "slow" action or similar.
+    // This is something to review in future iterations when tests are solid, but I haven't found a better way to
+    // solve this issue.
+    cy.on("uncaught:exception", (err, runnable) => {
+        return false;
+    });
+});
+
+Given('user opens the overview page', () => {
+    // Forcing "Pause" to not cause unhandled promises from the browser when cypress is testing
```
nrfox, line 18: :+1: I've noticed these error messages before about unhandled promises. Perhaps the frontend needs to more gracefully handle request cancellations when navigating away from a page with in-flight requests?

frontend/cypress/integration/common/overview.ts
```diff
@@ -0,0 +1,135 @@
...
+    cy.get('article[data-namespace="' + ns + '"]')
+        .should('exist')
+        .find('[data-overview-type="' + innerType + '"]')
+        .should('exist');
+});
+
+Then(`user sees the {string} namespace list`, (nslist) => {
+    const nss = nslist.split(',');
+    cy.get('article')
+        .should('have.length', 2)
```
nrfox, line 124: Is this testing the number of namespaces in the dropdown? imo we don't want hardcoded numbers in the tests for things like this since they will change over time and they might also change depending on whether we are testing in upstream or downstream. If anything, we should get the number of namespaces we expect with something like `cy.exec('kubectl get ns')` and compare that to the number of namespaces in the list.

frontend/cypress/integration/common/overview.ts
```diff
@@ -0,0 +1,135 @@
...
+            break;
+        case 'Workloads':
+            innerId = 'workload';
+            break;
+        case 'Services':
+            innerId = 'service';
+            break;
+    }
+    cy.get('button[aria-labelledby^="overview-type"]')
+        .should('exist')
```
nrfox, line 58: I could be wrong but I think that `cy.get(...)` and most of the cypress API will block and timeout/fail if the button does not exist so you don't need to include the `.should('exist')` assertions.

frontend/cypress/integration/common/overview.ts
```diff
@@ -0,0 +1,135 @@
...
+
+const url = "/console";
+
+Before(() => {
+    // Focing to not stop cypress on unexpected errors not related to the tests.
+    // There are some random failures due timeouts/loadtime/framework that throws some error in the browser.
+    // After reviewing the tests failures, those are unrelated to the app, so,
+    // it needs this event to not fail the CI action due some "slow" action or similar.
+    // This is something to review in future iterations when tests are solid, but I haven't found a better way to
+    // solve this issue.
```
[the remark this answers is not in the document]
lucasponce: From what I saw, the compiled final .js had some pull errors.
So, I've spent some time on it and it was unrelated to the test, but it was adding noise with pipeline failures.
It will require some investigation but with this hack, the tests passed well.

This should be iterated, but when coding "real" tests with some complexities we'll have to introduce some code like this.

frontend/cypress/integration/common/overview.ts
lucasponce: Well, it's a combination, in theory, the "in-flight" requests are properly canceled, but this was more noise that the test takes more time than the default refresh poll. Then it creates noise and makes the test longer.

frontend/cypress/integration/common/overview.ts
lucasponce: oh, yes, good point, it should compare with the length of the nslist.
But I'm -1 to call the system for a `kubectl get ns` unless there is a good reason.

I'll update the code.

frontend/cypress/integration/common/overview.ts
lucasponce: Yes, but the "should('exit')" makes the trick run faster, instead to trigger the timeout and the retry defaults.
I've browsed a couple of examples and I see this in a common way.

frontend/cypress/integration/common/overview.ts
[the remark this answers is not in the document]
nrfox: If there are specific errors that we expect to be thrown we should catch those and ignore them scoping the error handling to the specific test. I'm worried that ignoring all unhandled exceptions will hide real issues that we could've caught as part of the tests. Have you tried removing this since setting the `refresh=0`? I'm wondering if the intermittent errors you were seeing were caused by the unhandled promise rejections.

frontend/cypress/integration/common/overview.ts
nrfox: Even better!

frontend/cypress/integration/common/overview.ts
nrfox: [From the docs](https://docs.cypress.io/guides/core-concepts/introduction-to-cypress#When-To-Assert), the default behavior with timeout/retries is a major feature of the tool. It also looks like the assertions share the same timeout as the command: https://docs.cypress.io/guides/core-concepts/introduction-to-cypress#Example-2-Additional-Assertions. If we can save having to do an assertion for existence for every element in every test since cypress is already doing that, it will save time and make the tests simpler to read/write.

frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "Instead update 'web_root' in Kaili config map in OpenShift console."
   ],
   "homepage": "./",
+  "proxy": "http://localhost:8000",
```
nrfox, line 18: Is this from testing?
```suggestion
```

frontend/cypress/integration/common/overview.ts
lucasponce: Let me review if I can remove some of them.
I remember I found some cases that I needed for the chain.

frontend/package.json
lucasponce: Och, sorry, I thought it was fixed.
Thanks

frontend/cypress/integration/common/overview.ts
lucasponce: I've updated the PR. I removed some unnecessary assertions, perhaps I was updating the code and I didn't check that it was redundant.
Let me know how it looks there.

frontend/cypress/integration/common/overview.ts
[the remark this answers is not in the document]
nrfox: Totally agree. Nobody wants a flaky pipeline. Can we just catch unhandled promise rejections for now and see if that solves the intermittent failure issue?
```
Cypress.on('uncaught:exception', (err, runnable, promise) => {
  // when the exception originated from an unhandled promise
  // rejection, the promise is provided as a third argument
  // you can turn off failing the test in this case
  if (promise) {
    return false
  }
  // we still want to ensure there are no other unexpected
  // errors, so we let them fail the test
})
```

frontend/cypress/integration/common/overview.ts
nrfox: LGTM!

frontend/cypress/integration/featureFiles/overview.feature
```diff
@@ -0,0 +1,79 @@
...
+  @focus
+  Scenario: See "alpha" and "beta" namespaces
+    Given user opens the overview page
+    Then user sees the "alpha" namespace
+    Then user sees the "beta" namespace
+
+  @focus
+  Scenario: Doesn't see a "bad" namespace
+    Given user opens the overview page
+    Then user doesn't see the "bad" namespace
```
nrfox, line 14: General comment where I'm just thinking out loud. There's some overhead involved with opening the overview page. I wonder if we'll want to limit the amount of separate scenarios in a test and combine some of these.

ScriptingShrimp: Nope, testcase should be atomic. I would prefer high granularity with single (simple) assertions. We can [reduce steps in scenarios](https://cucumber.io/docs/gherkin/reference/#background) with few `Background:` steps, like:

```gherkin
Background: 
    Given user is at administrator perspective
    And user is at overview page

@overview-page
Scenario: See "alpha" and "beta" namespaces
    Then user sees the "alpha" namespace
    And user sees the "beta" namespace
```
@lucasponce we can save ~10 lines and get better readability 

Also chaining `Then` keyword does not make much sense, see example
```gherkin
Given url 'http://myhost.com/v1/cats&#39;'
And request { name: 'Billie' }
When method post
Then status 201
And match response == { id: '#notnull', name: 'Billie' }
```
Best practice is chain steps or conditions with `And` keyword.
If we want to early merge, I can request this modifications later

frontend/cypress/integration/common/overview.ts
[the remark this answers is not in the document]
ScriptingShrimp: What are these "pull errors"? Do we have some logs, or steps to reproduce? 
I don't want to discard all uncaught exceptions.

frontend/src/pages/Overview/OverviewPage.tsx
lucasponce: Yes, the "data-*" attributes depend on the domain (Kiali) and I expect that the author of the test can add them to simplify the testing.

In the past, typically QE asked Dev to add some Ids for better selection handling, but I don't see that necessary (unless there is some complex corner case).

frontend/cypress/integration/common/overview.ts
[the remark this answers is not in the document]
lucasponce: These errors are similar to these ones [1]

image

And it seems that are related to the situation that when we click in that button, there is a request to the server, and randomly the cypress framework detects some promise not completed.

That button doesn't have a problem, it's a matter of how the framework is setup and we will find more scenarios like this, so we need to introduce some tradeoffs, as we can't have an unreliable pipeline due to some random errors (timeouts, infrastructure, etc). 

I'm testing Nick's proposal to see if this mitigates the problem.

[1] https://github.com/kiali/kiali/runs/5815529616?check_suite_focus=true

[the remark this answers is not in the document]
lucasponce: @nrfox your comment seems to work, I've updated the PR.

frontend/cypress/integration/featureFiles/overview.feature
lucasponce: > General comment where I'm just thinking out loud. There's some overhead involved with opening the overview page.

Well, +1 but I'd separate the "day 1" tasks from the "day 2" :).

For "day 1" I expect that we create a set of tests for all pages, today we have only work in progress for login and overview.

In the next iteration "day 2" we can refine and refactor the tests once we have more clarity about what's the best approach to maintain these tests.

Also, I expect that these tests will be part of the regular PR work, so any change in the UI will likely include a revision of the UI tests focused on the "feature" that is being tested.

I haven't completed all the scenarios I had in mind for the overview page (the actions on the namespaces are not yet covered), but I have interest to have a first version that may help to speed up the skeleton for other pages as well.

frontend/cypress/integration/common/overview.ts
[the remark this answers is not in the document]
lucasponce: @ScriptingShrimp I did a pass of the wording in the feature with better use of the "aliases". 
Thanks for the suggestions, let me know how it looks.

I'm less worried about the "content" as we will add new tests and modify the existing ones, but I'd like that at least this work may help others to create more tests.

nrfox (commented):
LGTM! Thanks for adding this and jumping into the framework. I'll defer to @ScriptingShrimp or other QE for approval.

nrfox:
> > 
> 
> It could be reused with minor modifications, but we need to merge #4900 first. From QE point of view, we only want to merge tests compatible with downstream environment. We can make one exception in this PR

Just to be clear, the compatibility issue is needing to update the Jenkins pipeline(s) to deploy the error rate demo for this test to work in downstream?

ScriptingShrimp:
two issues regarding this specific PR

1. this testcase will fail in every environment that does not support authentication method `anonymous`
2. we don't have working jenkis pipeline right now - and even if we do, we still do not have scripts for deploying `error rate demo` implemented, so this sort of verification in downstream environment  obviously take some time to process

lucasponce:
This is upstream work, and I don't see a reason to block the work as we have implemented a pipeline to test UI upfront.

The UI tests should work in any environment and there should be the flexibility to define the authentication method.
You can propose the changes in this PR, or when you are going to work in the downstream pipeline to propose them.

But the goal is to have an upstream set of UI tests that can be used anywhere.

The script to install this demo (and future) should be in the project, this one is under 
https://github.com/kiali/kiali/blob/master/hack/istio/install-error-rates-demo.sh

(It's waiting for a minor fix).

mattmahoneyrh:
Attempted to run the **overview** test on OCP, which failed with the below error.

Perhaps there is a dependent PR as mentioned by @ScriptingShrimp, but either way the login and home tests pass in master on OCP but do not in this PR (fail with similar error as the overview test below).

Set below environment variables
export CYPRESS_BASE_URL=<kialiBaseURL>
export CYPRESS_USERNAME=<username>
export CYPRESS_PASSWD=<passworwd>

Error: Webpack Compilation Error
./cypress/integration/featureFiles/overview.feature 1:15
Module parse failed: Unexpected token (1:15)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
**> Feature: Kiali Overview page**
| 
|   User opens the Overview page and see the demo "error-rates" namespaces
    at Watching.handle [as handler] (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/packages/server/node_modules/@cypress/webpack-preprocessor/dist/index.js:180:23)
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Watching.js:99:9
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/tapable/lib/Hook.js:154:20)
    at Watching._done (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Watching.js:98:28)
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Watching.js:73:19
    at Compiler.emitRecords (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Compiler.js:499:39)
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Watching.js:54:20
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Compiler.js:485:14
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/tapable/lib/Hook.js:154:20)
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Compiler.js:482:27
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/neo-async/async.js:2818:7
    at done (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/neo-async/async.js:3522:9)
    at AsyncSeriesHook.eval [as callAsync] (eval at create (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/tapable/lib/HookCodeFactory.js:33:10), <anonymous>:6:1)
    at AsyncSeriesHook.lazyCompileHook (/home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/tapable/lib/Hook.js:154:20)
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/webpack/lib/Compiler.js:464:33
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/packages/server/node_modules/graceful-fs/graceful-fs.js:111:16
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/node_modules/graceful-fs/graceful-fs.js:61:14
    at /home/mmahoney/.cache/Cypress/9.2.1/Cypress/resources/app/packages/server/node_modules/graceful-fs/graceful-fs.js:45:10
    at FSReqCallback.oncomplete (fs.js:179:23)

lucasponce:
Thanks for the feedback @mattmahoneyrh.
Yes, those should pass in both environments, if not the PR is not completed.
I'll rebase with the master branch.

mattmahoneyrh:
The "Error: Webpack Compilation Error" no longer occurs.

Ran all tests again, with install-error-rates-demo.sh deployed, resulted in errors such as below.

> 54 | 					cy.get('.pf-c-form').contains(provider, { matchCase: false }).click()
     | 					                     ^
  55 | 				}
  56 | 
  57 | 				if (auth_strategy === 'openshift') {

ScriptingShrimp:
> Ran all tests again, with install-error-rates-demo.sh deployed, resulted in errors such as below.

So I found error in my `cy.login()` code, that is why it fails here (but only on stage). Sorry about that @mattmahoneyrh and thanks for your review.

As Matt wrote, he deployed error rates demo on the OCP, I fixed mine part and we got the results:
image
Since there are changes in the UI code itself (and tests rely on this), the results are about right. Obviously we need to build this PR and deploy it in OCP first :smiley:

To build this PR, I need to verify if our jenkins pipelines are working after repo merge. Not sure if the UI build is still pointing to old `kiali-ui` repo. Will catch with @FilipB and let you know. 

More in general: 
- verifying new tests added upstream could be done on nightly builds (in OCP), so no need to verify downstream compatibility on  demand on PR level (but could be still done)
- we might want to add `@upstream` and `@downstream` tags in scenarios - once we will see them passing in corresponding environment (that basically means all scenarios are created with `@upstream` tag)
- creating new flows in Github Actions creates technical debt by direct proportion in Jenkins pipelines. (this basically means that with upcoming release we need to have them in the place)

With that said, PR LGTM! (I will add approval after tomorrows discussion)

lucasponce:
I have a concern. We can not block upstream block waiting for some private pipeline to test.
That's against the strategy we are following.

If some upstream scenario needs changes in the private pipelines, those changes should be proposed by QE.

But we are not going to introduce a strong dependency here, the upstream pipelines need to move forward.

nrfox:
> I have a concern.
We can not block upstream block waiting for some private pipeline to test.
That's against the strategy we are following.

Agreed but I don't think that's what's being said here. This change adds an additional setup dependency for testing which is to have the error rates demo deployed into the cluster. We will need to deploy more than just bookinfo to be able to adequately test various scenarios for Kiali but we at least need to communicate what those additional requirements are before the tests rely on them. A common setup script that both pipelines use in to order to setup the testing env would probably be enough to solve this problem.

For the short term, does the overview page need to rely on the error rates demo or can we test the same things with just bookinfo?

lucasponce:
> A common setup script that both pipelines use in to order to setup the testing env would probably be enough to solve this problem.

The script used in the public pipelines works in both environments.

> For the short term, does the overview page need to rely on the error rates demo or can we test the same things with just bookinfo?

"Bookinfo" doesn't generate errors by default, "error-rates" does. 
The main features of the "Overview" page are to filter by "health" and that sounds like the best test to start.

Unless anyone doesn't want to test errors in Kiali.

ScriptingShrimp (approved):
LGTM, thanks Lucas for this PR, It was good example.

.github/workflows/kiali-ci.yml
```diff
@@ -134,6 +134,9 @@ jobs:
       - name: Setup kind
         run: hack/setup-kind-in-ci.sh
 
+      - name: Setup Error Rates demo
+        run: hack/istio/install-error-rates-demo.sh -c "kubectl"
```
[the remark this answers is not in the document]
ScriptingShrimp: > have environment-sensitive yml files

sounds good for QE

PR #4876, opened 2022-03-30
[cypress] user logout

pbajjuri20:
Kiali user logout smoke test case


pbajjuri20:
> Is there a reason to not reuse the existing login?



> Is there a reason to not reuse the existing login?

Pavel and me, working on how to reuse the existing Cypress code. Once we explored will refactor this code.

frontend/cypress/integration/common/kiali_logout.ts
```diff
@@ -0,0 +1,66 @@
...
+
+And('user clicks on admin', () => {
+    if (auth_strategy === 'openshift') {
+        cy.get('#user-dropdown-toggle').click() 
+    }
+})
+
+Then('user logout successfully', () => {
+    if (auth_strategy === 'openshift') {
+        cy.get('.pf-c-dropdown__menu-item').click()
```
ScriptingShrimp, line 61: I would like to see better assertion here, we can connect and find out something more precise

pbajjuri20: > I would like to see better assertion here, we can connect and find out something more precise

Sure, will work on that

lucasponce: +1 I've added a similar comment in other PR.
Feel free to add together with the test the changes in the components that you'd need, typically using "data-*" attributes work.
In the PR of the "overview.feature" there are several examples that may help with naming and conventions.

frontend/cypress/integration/common/kiali_logout.ts
```diff
@@ -0,0 +1,71 @@
...
+    cy.url().should('include', 'console')
+})
+
+And('user clicks on admin', () => {
+    if (auth_strategy === 'openshift') {
+        cy.get('#user-dropdown-toggle').click() 
+    }
+})
+
+And('user logout successfully', () => {
```
ScriptingShrimp, line 59: This action cannot be called "user logout successfully". Let me explain why
- you are doing logout operation - clicking logout button. 
- this action may logout user, so it should be called something like "user clicks logout button"
- next step should be logout verification, so i.e. already mentioned  "user is logged out successfully"

ScriptingShrimp (changes requested):
Please rewrite the assertion for logout assertion. Rest of the testcase is OK

pbajjuri20:
git rebase has crashed this PR, I will create the new PR.

Issue #4877, opened 2022-03-31
compatibility matrix update and backport to v1.48


jmazzitelli:
I believe we need to add compatibility matrix entry for v1.48 and maistra 2.2 (and backport it to the v1.48 branch):

https://github.com/kiali/kiali/blob/master/config/version-compatibility-matrix.yaml#L47-L52

jmazzitelli:
master PR: https://github.com/kiali/kiali/pull/4881
backport PR: https://github.com/kiali/kiali/pull/4882

PR #4881, opened 2022-04-01
update compatibility matrix

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/4877


config/version-compatibility-matrix.yaml
```diff
@@ -47,6 +47,15 @@
...
+    - meshVersion: "2.2"
+      kialiMinimumVersion: "1.48"
+      kialiMaximumVersion: ""
+    - meshVersion: "2.2"
+      kialiMinimumVersion: "1.36"
+      kialiMaximumVersion: ""
+    - meshVersion: "2.2"
+      kialiMinimumVersion: "1.24"
+      kialiMaximumVersion: ""
     - meshVersion: "2.1"
```
jshaughn, line 59: I don't totally get this structure for maistra and ossm, how is this different from one entry with 1.24 as the Minimum Version?

jmazzitelli: Because there is a strict 1-to-1 mapping between kiali and OSSM. You don't want to say "2.2 has a minimum of 1.24" because that implies 1.25 is supported, and it is not. The next supported version is 1.36. Hence the empty "Maximum" version which is handled specially in the code. I don't recall this code explicitly so can't describe it in detail, but that's how I remember it working.

jshaughn (approved):
OK re: the weird format.  At least it's consistent with what is there for the earlier versions.  Maybe we should add something like `kialiFixedVersion` to make this more simple in the future.

PR #4886, opened 2022-04-01
Fix cypress pathing

nrfox:
Fixes the path where cypress is looking for TS. Broke because of frontend repo merge.


ScriptingShrimp (approved):
LGTM

PR #4888, opened 2022-04-01
[cypress] Remove old cypress setup

nrfox:
Removes unused cypress setup. Renames cypress-bdd --> cypress so that the bdd setup is now the default.


frontend/cypress.json
```diff
@@ -2,5 +2,15 @@
...
+  "integrationFolder": "cypress/integration",
+  "testFiles": "**/*.feature",
+  "defaultCommandTimeout": 40000,
+  "animationDistanceThreshold": 20,
+  "execTimeout": 150000,
+  "pageLoadTimeout": 90000,
+  "requestTimeout": 15000,
+  "responseTimeout": 15000,
+  "supportFile": "cypress/support/index.ts",
+  "pluginsFile": "cypress/plugins/bdd.ts",
```
ScriptingShrimp, line 14: We could use vanilla plugin file, but not a major concern. LGTM

ScriptingShrimp (approved):
LGTM

PR #4889, opened 2022-04-01
Run backend integration tests in CI

### LINKED ISSUE #4874: e2e API tests are run in CI
e2e API tests are run against every PR and release in the GH actions CI pipeline.

Blocked by: #4838 and probably #4868 

Epic: #4826

nrfox:
Run backend integration tests as part of upstream CI. Runs as a separate job with a separate e2e env in parallel with the frontend integration tests. Builds PR go binaries with the `-race` flag for race detection while running tests. Releases won't be built with this flag since there is a perf penalty associated with it.

Fixes #4874

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ GO_BUILD_ENVVARS = \
 	CGO_ENABLED=$(CGO_ENABLED)
 
 # Extra build flags passed to the go compiler.
-GO_BUILD_FLAGS = 
+GO_BUILD_FLAGS ?= 
 
 # Determine which Dockerfile is used to build the server container
 KIALI_DOCKER_FILE ?= Dockerfile-ubi8-minimal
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -58,10 +58,14 @@ GOPATH ?= ${HOME}/go
...
 GOARCH ?= $(shell ${GO} env GOARCH)
+CGO_ENABLED ?= 0
 GO_BUILD_ENVVARS = \
 	GOOS=$(GOOS) \
 	GOARCH=$(GOARCH) \
-	CGO_ENABLED=0 \
+	CGO_ENABLED=$(CGO_ENABLED)
+
+# Extra build flags passed to the go compiler.
+GO_BUILD_FLAGS =
```
jmazzitelli, line 68: I believe you want `?=` here - to tell the makefile "if this env var is already set, don't set it here"... the way this is now, this env var will always be empty even if the one invoking the make file exported that env var to this process.
```suggestion
GO_BUILD_FLAGS ?= 
```

jmazzitelli (changes requested):

Makefile
nrfox: Yep!

PR #4891, opened 2022-04-02
feat: add fixed version in matrix

Xunzhuo:

**Describe the change**

feat: add fixed version in matrix

refer to: https://github.com/kiali/kiali/pull/4881#pullrequestreview-929045299


config/version-compatibility-matrix.yaml
```diff
@@ -26,82 +26,58 @@
...
-      kialiMaximumVersion: ""
+      kialiFixedVersion: "1.48"
     - meshVersion: "2.2"
-      kialiMinimumVersion: "1.36"
-      kialiMaximumVersion: ""
+      kialiFixedVersion: "1.36"
     - meshVersion: "2.2"
-      kialiMinimumVersion: "1.24"
-      kialiMaximumVersion: ""
+      kialiFixedVersion: "1.24"
```
jshaughn, line 49: This is better but do we really need the multiple entries for the, in this case, the 2.2 meshVersion?  I mean, if you're running 2.2 you need 1.48, not 1.36 or 1.24.  Or, is this somehow because the 1.48 operator installs 2.2, 2.1 and 2.0?

jshaughn (changes requested):
Please add some tests of the new logic to versions_test.go.

config/version-compatibility-matrix.yaml
jmazzitelli: > Or, is this somehow because the 1.48 operator installs 2.2, 2.1 and 2.0?

That how I see it. If you are OSSM 2.2, you could be running Kiali Server 1.24. (OSSM 2.2 operator can install Kiali 1.24).

jmazzitelli (changes requested):
I found one problem already. The version string you get back can be in X.Y.Z form and this fails the fixed version check.

I got back "2.1.1" in my latest test - and that version string isn't in the compatibility matrix.

We may need to rethink this.

config/version-compatibility-matrix.yaml
jmazzitelli: I confirmed jay is right - we only need one entry. I included that change in my PR to this PR.

jmazzitelli:
OK. I think I have this fixed with [my PR to this PR](https://github.com/Xunzhuo/kiali/pull/6).

We only want to check X.Y versions of kiali and OSSM/Maistra (since these are fixed - e.g. OSSM 2.1.z is only supported with Kiali 1.36.z. There can be many z-stream patch releases, but it is always 2.1.z == 1.36.z. So I updated the code to support this in that PR.

Also, when you look up the version from the istiod endpoint, it returns the version of the control plane (e.g. even if I have a OSSM 2.1 operator, I can install a 2.0 SMCP and that results in a 2.0 control plane  (which reports itself as a 2.0.z control plane, e.g. 2.0.8). So we only need one entry per mesh version in the compatibility matrix file. I also made that change in my PR.

jmazzitelli:
@Xunzhuo -- https://github.com/Xunzhuo/kiali/pull/7

@jshaughn I'm removing all "Maistra" (and "Maistra Project") compatibility matrix data. We don't need it and it just makes things more complex in the matrix and code.

Xunzhuo:
@jmazzitelli Thanks for updating 
, I am on a vacation, so computer is away from me:(

jmazzitelli:
@Xunzhuo OK. I might create a separate PR with all of this so I can update it and merge. I found another issue that I need to fix. I'll take it from here. Thanks.

jmazzitelli:
closing this PR, to be replaced with https://github.com/kiali/kiali/pull/4898

Issue #4892, opened 2022-04-02
ui crash with no gateways


jshaughn:
I'm seeing a UI crash when visting service detail page.  this seems to be due to an assumption in the UI that the server will return an empty array, as oppsed to null, when querying for certain istio config, like gateways, but when no gateways are found in the registry.  It seems like a fair assumption, I think the problem is server-side...  I'll see if I can track it down.

PR #4893, opened 2022-04-02
make sure RegistryConfiguration is initialized with empty arrays.

jshaughn:
Fixes https://github.com/kiali/kiali/issues/4892

@lucasponce I'm not as familiar with this code as you, but this seems to be the root of a UI crash I've been getting, as described in the issue.  Please take a look.  If this is in fact a desierd fix, I think we may need a 1.48 backport.

### DIFF
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -242,7 +242,20 @@ func ParseRegistryEndpoints(endpoints map[string][]byte) ([]*RegistryEndpoint, e
 }
 
 func ParseRegistryConfig(config map[string][]byte) (*RegistryConfiguration, error) {
-	registry := RegistryConfiguration{}
+	registry := RegistryConfiguration{
+		DestinationRules: []networking_v1alpha3.DestinationRule{},
+		EnvoyFilters:     []networking_v1alpha3.EnvoyFilter{},
+		Gateways:         []networking_v1alpha3.Gateway{},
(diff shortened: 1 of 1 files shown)


lucasponce:
The patch looks good.
But I'm curious how you got to this error. How did you reproduce the error? I can review if there is anything missing in the logic.

jshaughn:
> But I'm curious how you got to this error.

I didn't really try to do anything special, I am running the travel demo against master,, but do not create any gateways (so not the tutorial, just the vanilla travel demo installed by hack script).  I go to the services list view and select the cars service detail.  For me I saw the problem every time.  At first I thought it was due to my PF branch, but then I was able to see it on master as well.  This is why I am thinking we may need a backport.

lucasponce:
> I go to the services list view and select the cars service detail. For me I saw the problem every time. At first I thought it was due to my PF branch, but then I was able to see it on master as well. This is why I am thinking we may need a backport.

Let me investigate this. As commented, patch was ok, but I haven't seen this problem before, so I'd like to see if it's related or there is a hidden problem around.

lucasponce:
Ouch, nice catch @jshaughn!
You're right, we populate the dev environment always the "bookinfo" gateway that this case was completely missed.

lucasponce (approved):

lucasponce:
I'll prepare a cherry-pick PR for the v1.48

PR #4898, opened 2022-04-04
add fixed version in compatibility matrix and correct some problems in the checks

jmazzitelli:
this is https://github.com/kiali/kiali/pull/4891 but with some additional fixes.


jshaughn (approved):
Much better looking and easier to mantain.  I assume we can remove "maistra" because there aren't really even any "maistra" releases.  Even if someone were to build it AND run kiali, they'd just get a warning, and that may even be valid because we've never tested it :-)

jshaughn (approved):
LOK

PR #4905, opened 2022-04-05
Add make cmd for cypress and reduce mem usage in CI

nrfox:
Adds a make command for cypress and tweaks some cypress settings in CI to reduce memory consumption (and hopefully speed up tests).

Config reference: https://docs.cypress.io/guides/references/configuration#Global
> The number of tests for which snapshots and command data are kept in memory. Reduce this number if you are experiencing high memory consumption in your browser during a test run.

Removes screenshots/videos from CI to save on GH resource consumption.

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -140,25 +140,13 @@ jobs:
           echo "::set-output name=kiali_url::$KIALI_URL"
         id: set-kiali-url
 
-      - name: Run cypress integration tests
-        uses: cypress-io/github-action@v2
-        with:
-          working-directory: frontend
-          command: yarn run cypress:run --config baseUrl=${{ steps.set-kiali-url.outputs.kiali_url }}
-
-      - name: Upload cypress screenshots when tests fail
-        uses: actions/upload-artifact@v3
-        if: failure()
-        with:
-          name: cypress-screenshots
-          path: frontend/cypress/screenshots
-
-      - name: Upload cypress videos when tests fail
-        uses: actions/upload-artifact@v3
-        if: failure()
-        with:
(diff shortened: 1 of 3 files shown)


lucasponce:
Well, yes, but we're in the free subscription :) and as now we're starting to overload the pipelines, I wonder if we need those videos, or when we have some error we can repeat the pipeline locally and save bandwidth.

nrfox:
> Well, yes, but we're in the free subscription :) and as now we're starting to overload the pipelines, I wonder if we need those videos, or when we have some error we can repeat the pipeline locally and save bandwidth.

Surprisingly I haven't found any limits to the number or size of artifacts you can store other than there's a 90 day retention policy: https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#artifact-and-log-retention-policy. I think there's some value in keeping the videos/screenshots in case there's an intermittent failure or one that only shows up in CI for whatever reason. That being said, cypress runs do take longer with video/screenshots enabled because it compresses them afterward. We can save on runtime (which there are limits for) there by skipping the compression step and having larger artifacts. Or, as you suggest, we can disable the videos/screenshots for now and add them back later if we find we want them.

lucasponce:
> Or, as you suggest, we can disable the videos/screenshots for now and add them back later if we find we want them.

Yes, they bill by time, so reducing what we don't need to run in every PR sounds like a good option.

I mean, I guess the usual flow would be when there is a problem, to run locally the test, so I don't see the need to keep generating the videos on every PR.

jmazzitelli (dismissed):

nrfox:
Rebased with master to fix conflicts

jmazzitelli (approved):

PR #4906, opened 2022-04-05
Fix install-error-rates-demo hack

lucasponce:
The hack for the error-rates demo missed the "NetworkAttachmentDefinition" for OpenShift environments.


jmazzitelli (approved):
ah,yes. Need NAD in the beta namespace, too.
LGTM

Issue #4836, opened 2022-03-23
Investigate update of Patternfly to be compatible with OS Console


jshaughn:
Epic: https://github.com/kiali/kiali/issues/4260

jshaughn:
As expected, there is some negative impact wrt popper/tippy.  This is a tricky area because cytoscape requires cytsoscape-popper, which in turn requires popper.js, which uses tippy.  And we also use popper.js independently.  So getting a combination that is happy is not easy.  Some observations:
- updating cytoscape from 3.15.5 to 3.16+ can negatively impact our label sizes, I see this for dagre on initial render of a smaller graph.
- popper.js 1 is deprecated, and is superseded by  v2 at @popperjs/core
- #patternfly/react-core depends on tippy.js, which depends on popper.js
- react-datepicker also depends on popper.js
- the latest cytoscape-popper depends on v2 @popperjs/core

So, it seems we can't update to popper v2 and we have rendering issues with cy 3.16+.  This limits our ability to move forward with those dependencies.  But, we seem to have some popper/tippy issues afther the pf-react updates, I'm not sure why but I guess it's more related to the pf changes than popper/tippy, since that didn't really change.

lucasponce:
> the latest "cytoscape-popper" depends on v2 @popperjs/core

Mmm, if this is blocking, we could review and clone this to have the freedom to control this (and future clashes).
I looked at the code and it's less complex that the layouts.

So, I don't see a "why not", it's a tradeoff but it seems it can simplify this conflict.

jshaughn:
>  we could review and clone this 

I don't think this will really enable us to move forward because pf-react itself is not using popper v2.  Until that moves forward I think we are bound to v1.  Yesterday I tried to move to v2 and failed, today I will see if I can resolve the issues I see just maintaining v1.  From what I can see PF deprecated some popper/tippy stuff AND disabled it, so not sure if that's our issue, but will investigate.

nrfox:
Can we use the [resolutions](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/) field to try and make each dependency happy?

jshaughn:
> Can we use the resolutions field to try and make each dependency happy?

As it stands right now, I don't think we want to try and incorporate popper v2.  I think it makes more sense to match pf-react.

PR #4909, opened 2022-04-05
update patternfly-react and other dependencies

jshaughn:
Closes https://github.com/kiali/kiali/issues/4836

This PR looks to update dependencies to be more in-line with OpenShift Console, help with security violations, and where possible remove unnecessary styling.  It is not intended to leave things looking *exactly* the same, but differences should be reasonable.


lucasponce:
I fetched locally and I can have first feedback: a lot of styles broken as expected, but fewer things broken than I thought :).
So we need to work, but I think it's approachable with reasonable tradeoffs.

But yes, some styles are broken badly.

Thanks for the work, I'll add more feedback with more details as soon as I find issues.

ScriptingShrimp:
That is major update. @pbajjuri20 I'm wondering if we can execute current selenium suite against this draft PR, to have an idea what is affected. 
If the UI regression test will have a lot of failures, we might prioritize them in migration to Cypress.

lucasponce:
@pavolloffay @pbajjuri20 I'd wait to start any testing exploration on this PR until it progressed.
It's still a draft and it will have more changes, especially affecting styles.

pavolloffay:
@lucasponce I am not sure what should I be looking at in this PR? Could you please clarify?

lucasponce:
Sorry @pavolloffay ! My bad the cc was for Pavel alias cc @ScriptingShrimp !

pbajjuri20:
> That is major update. @pbajjuri20 I'm wondering if we can execute current selenium suite against this draft PR, to have an idea what is affected. If the UI regression test will have a lot of failures, we might prioritize them in migration to Cypress.

@ScriptingShrimp : will look about this

jshaughn:
@pbajjuri20 @ScriptingShrimp , no problem to be aware of this effort but this is still a draft PR, no need to take any action right now.  we will evaluate how we want to proceed here and keep you up to date.

PR #4913, opened 2022-04-06
Trivial. Clean PR template messages

lucasponce:
The PR template is outdated (i.e. we don't use JIRA).
Just to have a clean format.

image


jmazzitelli (approved):
much better... people never filled in that large form but always left the text anyway.

Issue #4459, opened 2021-10-27
Kiali deployed behind proxy does not work as expected


brendarearden:
## Description

Starting in Kiali 1.35, Kiali's pod was changed to a read-only file system https://github.com/kiali/kiali/pull/4022/files. This is now causing issues for deployments of Kiali that are behind a proxy as there is no way to specify the proxy path as well as a server.web_root. 

Prior to this change, I was utilizing the `env.js file` to [manually override](https://github.com/rancher/charts/blob/dev-v2.6/charts/rancher-kiali-server/rancher-kiali-server/100.0.0%2Bup1.35.0/templates/web-root-configmap.yaml#L11) the `window.WEB_ROOT` value. This meant that I was able to leave the `server.web_root` value as the default value of `/kiali` and also have the `window.WEB_ROOT` be my proxy path + the `server.web_root`. 

## Example

http://localhost:2001/kiali is where the UI is accessible from. This means `server.web_root=/kiali`.

http://localhost:8080/proxy is the endpoint that will proxy `http://localhost:2001/kiali` so my `window.WEB_ROOT` would need to be `/proxy/kiali`. Since the `window.WEB_ROOT` is set by the `server.web_root`, then there is no value that I can set the `server.web_root` that will work for kiali to still be served from `/kiali` but my window root to be `/proxy/kiali`.

If I set `server.web_root=/proxy/kiali`in an attempt to include my proxy path, then when I hit my proxy endpoint `http://localhost:8080/proxy` it will redirect the traffic to `http://localhost:8080/proxy/proxy/kiali` because kiali is being served from `/proxy/kiali`, then the `window.WEB_ROOT` will overwrite the url path and it will be left at `http://localhost:8080/proxy/kiali` which is my proxy end point, but not the path where kiali is currently being served up from because we set that to `/proxy/kiali` and with the `window.WEB_ROOT` rewrite we are now only hitting `/kiali`.

https://github.com/brendarearden/example-kiali-redirect I have created a simple proxy example using python to help explain this a little more and you can see how the path overwrite is occurring.

brendarearden:
The actual use case looks more like the former example, I created the latter as a very basic proxy example for the sake of opening this issue to make it easier to visualize what was going on.  I can definitely see how that example implementation could be exploited for sure. 

[...]

PR #4914, opened 2022-04-06
Fix url rewrites 

doflamingo721:
**Describe the change**

Fix for url rewrites when proxies are placed in front of Kiali

**Issue reference**
https://github.com/kiali/kiali/issues/4459

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -155,9 +155,7 @@ func serveEnvJsFile(w http.ResponseWriter) {
 		body += fmt.Sprintf("window.HISTORY_MODE='%s';", conf.Server.WebHistoryMode)
 	}
 
-	if webRoot := strings.TrimSuffix(conf.Server.WebRoot, "/"); len(webRoot) > 0 {
-		body += fmt.Sprintf("window.WEB_ROOT='%s';", webRoot)
-	}
+	body += "window.WEB_ROOT = document.getElementsByTagName('base')[0].getAttribute('href').replace(/^https?:\\/\\/[^#?\\/]+/g, '').replace(/\\/+$/g, '')"
 
 	w.Header().Set("content-type", "text/javascript")
 	_, err := io.WriteString(w, body)


jmazzitelli (changes requested):
The code looks simple enough, but we would like you to provide some simple test procedures so someone can test this. Is there a simple way someone can create/deploy in minikube and be able to test this new feature? (note, not asking for tests in code - just some steps to manually test... "step 1.. install minikube and istio, step 2, install a proxy with these commands A, B., C, step 3, install kiali with these settings... etc")

doflamingo721:
@jmazzitelli Sure, I'll add the testing instructions.

jmazzitelli:
Here is how I tested, and I think it shows it is working. First, before this PR, this is the behavior that is seen: https://github.com/kiali/kiali/issues/4459#issuecomment-967745560

Now, build and deploy Kiali server with this PR and do basically the same thing as [that issue documents](https://github.com/kiali/kiali/issues/4459#issuecomment-967745560). These test procedures were done while running minikube with Istio installed already in the cluster:

1. Start the k8s api proxy: `kubectl proxy --port=8080`
2. Create the Kiali CR with `spec.server.redirect_path` set to `/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali`
3. Wait for Kiali server pod to start up.
4. Point your browser to: http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy
5. See that the browser shows the Kiali UI but has been redirected to: http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali/console/overview?duration=60&refresh=15000 (so it is still going through the proxy - before this PR, the browser would get redirected to this (thus circumventing the proxy): http://localhost:8080/kiali/console)
6. See that the env.js contains the redirected path:
```
$ curl http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali/env.js
window.WEB_ROOT='/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali';
```
7. Now edit the Kiali CR and set that redirect path to empty string (`""`) to see the old behavior (this is the default behavior - how Kiali will still behave out of box). Once the Kiali pod restarts, see the env.js has the normal web root:
```
$ curl http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali/env.js
window.WEB_ROOT='/kiali';
```
8. Point your browser back to http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy and see that the redirection goes to `/kiali/console` (http://localhost:8080/kiali/console/overview?duration=60&refresh=15000):

jmazzitelli:
I do not know what negative behavior this might cause in our auth strategies ... I only tested with 'anonymous' auth strategy. @israel-hdez do you see any problems with this? At minimum, are there any side effects that we would need to document on kiali.io (such as, `If using openid auth strategy, you cannot use the server.redirect_path feature` -- I do not know if that is true, I'm just giving an example).

doflamingo721:
**Steps to validate:**
1. Have the environment setup following the Kiali [Developer setup guide](https://github.com/kiali/kiali#developer-setup).
2. Create a kubernetes cluster and install a service mesh by following [these steps](https://github.com/kiali/kiali#create-a-kubernetes-cluster-and-install-a-service-mesh). If using minikube, make sure you set the `CLUSTER_TYPE`  environment variable to `minikube` . 
3. Set `server.redirect_path: "/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali"` in KialiCR in [kiali-operator/deploy/kiali/kiali_cr_dev.yaml](https://github.com/kiali/kiali-operator/blob/master/deploy/kiali/kiali_cr_dev.yaml) to instruct the operator to deploy Kiali.
4. Follow [these](https://github.com/kiali/kiali#building-the-container-image-and-deploying-to-a-cluster) steps to build the container images and deploy them to the cluster.
5. Run  `kubectl proxy --port=8080`  to expose the kubernetes api server.
6. Check if the kiali pod is up and running and access kiali at http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali .

**Results:**
The browser window url does not get overwritten to http://localhost:8080/kiali/console/overview?duration=60&refresh=15000  and refreshing the page does not result in 404 page not found.

jmazzitelli (approved):
LGTM. I approve.

Will wait for @israel-hdez to give his thoughts and approval.

jmazzitelli:
This failed for me when I tried to use `token-auth`. Without `redirect_path`, I was able to successfully log in with a SA token. But when I set redirect_path and go to the proxy URL, logging in with the token fails:
Screenshot from 2022-04-07 09-35-24
I don't know if this is supposed to work or not. But I would like to know if this has exposed a bug or not.

To replicate, do everything as explained above, but switch `spec.auth.strategy` to `token`. When the Kiali pod starts, go to the proxy URL as before and you should get the token login screen. Attempt to log in using the Kiali service account token. You can get it via this command:

```
kubectl get secret -n istio-system $(kubectl get sa kiali-service-account -n istio-system -o "jsonpath={.secrets[0].name}") -o jsonpath={.data.token} | base64 -d && echo
```

routing/router.go
```diff
@@ -155,8 +155,12 @@ func serveEnvJsFile(w http.ResponseWriter) {
...
 	}
 
-	if webRoot := strings.TrimSuffix(conf.Server.WebRoot, "/"); len(webRoot) > 0 {
-		body += fmt.Sprintf("window.WEB_ROOT='%s';", webRoot)
+	if redirectPath := conf.Server.RedirectPath; len(redirectPath) == 0 {
+		if webRoot := strings.TrimSuffix(config.Get().Server.WebRoot, "/"); len(webRoot) > 0 {
+			body += fmt.Sprintf("window.WEB_ROOT='%s';", webRoot)
+		}
+	} else {
+		body += fmt.Sprintf("window.WEB_ROOT='%s';", redirectPath)
```
israel-hdez, line 163: I was checking resulting outputs whe `kubectl proxy` is in use. We also have this `serveIndexFile` function which also prints the `web_root` setting in the index HTML file that is served:

https://github.com/kiali/kiali/blob/f444a09521033b1f9fbc5161e2ce65dc0c830001/routing/router.go#L171-L197

I see that `kubectl proxy` is doing the right URL rewrites in the resulting HTML. So, instead of adding yet another config setting, why not replace all this with something that does not depends on configs? So, in this `serveEnvJsFile` function, I think something like this would work:

```javascript
body += "window.WEB_ROOT = document.getElementsByTagName('base')[0].getAttribute('href')"
```

The previous line would completely replace the `if` block in this `serveEnvJsFile` function.
This way we will no longer need to add that `redirect_path` option and we would completely rely on the proxy url_rewrite logic.

I haven't tested, but I think this simpler change would be better, and it would save us from adding, and documenting, and explaining a new option, and how the new option interacts with other settings.

israel-hdez:
> This failed for me when I tried to use `token-auth`. Without `redirect_path`, I was able to successfully log in with a SA token. But when I set redirect_path and go to the proxy URL, logging in with the token fails:

I got the same issue. IMO, it completely depends on how the proxy is doing url_rewriting.

With `kubectl proxy` all auth strategies will be broken. The problem are the cookies.

`kubectl proxy` is doing url_rewrite only on resulting HTML, but any `Set-Cookie` headers remain unchanged. So, Kiali is setting cookies to be valid for the `/kiali` path and the proxy is keeping this unchanged. This leads to the browser is seeing `/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali` in the URL and having cookies for the `/kiali` path. Since there is no match on paths, the browser won't provide the cookies and Kiali won't see the credentials (effectively, the session does not persist).

I think `kubectl proxy` is too naive and does not provide the required configs to fix this. But other more advanced proxies have the needed options to properly manipulate HTTP headers and fix this situation.

routing/router.go
jmazzitelli: > So, in this serveEnvJsFile function, I think something like this would work:
> body += "window.WEB_ROOT = document.getElementsByTagName('base')[0].getAttribute('href').replace(/\/+$/g, '')"
> The previous line would completely replace the if block in this serveEnvJsFile function.

This is an interesting alternative solution. Let's try to see if that would work - as you say, it means we won't have to do anything with implementing/documenting a new config setting.

@doflamingo721 see if you can try that solution. It would require a single PR with a single line change apparently :)

But then we have to examine this - but I assume nothing has to change here: https://github.com/kiali/kiali/blob/f444a09521033b1f9fbc5161e2ce65dc0c830001/routing/router.go#L171-L197

jmazzitelli: @israel-hdez @doflamingo721 I tried that alternative solution and it does not work. Here's the env.js:

```
$ curl -L http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali/env.js && echo
window.WEB_ROOT = document.getElementsByTagName('base')[0].getAttribute('href').replace(//+$/g, '')
```

and accessing the UI at http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali produces the same error as before - we get redirected to a URL that does not include the proxy (I see it going to this: http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali). So that alternative is not viable.

israel-hdez: My fault, but probably the regex is badly escaped. I see I posted `//+$/g `. Probably it should be `/\/+$/g ` (notice the escaped slash).

If in doubt, you probably can try that whole JS expression `document.get.....` in the browser console while you are in Kiali to see if it gets the right URL

israel-hdez: What I mean is that in the resulting env.js you should see `/\/+$/g `. Probably golang is doing some escaping on its side.

jmazzitelli: oh, look at that.. it works :)

```
$ curl -L http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali/env.js && echo
window.WEB_ROOT = document.getElementsByTagName('base')[0].getAttribute('href').replace(/\/+$/g, '')
```

Going to http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali
brings the browser to: http://localhost:8080/api/v1/namespaces/istio-system/services/kiali:20001/proxy/kiali/console/overview?duration=60&refresh=15000

The diff:
```diff
diff --git a/routing/router.go b/routing/router.go
index ba6828b51..6c46341bb 100644
--- a/routing/router.go
+++ b/routing/router.go
@@ -155,9 +155,7 @@ func serveEnvJsFile(w http.ResponseWriter) {
                body += fmt.Sprintf("window.HISTORY_MODE='%s';", conf.Server.WebHistoryMode)
        }
 
-       if webRoot := strings.TrimSuffix(conf.Server.WebRoot, "/"); len(webRoot) > 0 {
-               body += fmt.Sprintf("window.WEB_ROOT='%s';", webRoot)
-       }
+       body += "window.WEB_ROOT = document.getElementsByTagName('base')[0].getAttribute('href').replace(/\\/+$/g, '')"
```

jmazzitelli: I did a little more testing, and that small alternative solution appears to be good. @doflamingo721 barring any errors you see in your official setup with this new way of implementing this, we should do that instead.

The only difference I see is it will affect everyone (in this original solution, the behavior is identical as before; only if you set redirect_path will window.WEB_ROOT be different than before). However, I don't see a problem right now. I'll keep playing with it to see if I can break it, but so far it seems to be doing what we want.

doflamingo721: @jmazzitelli @israel-hdez Thank you for your inputs on this. I'll try out the alternative solution for this at the earliest and if everything seems to workout I'll update my PR and re-request your reviews.

jmazzitelli: @doflamingo721 please test out that alternative solution and let us know. All tests pass over here.

doflamingo721: Hey @jmazzitelli @israel-hdez , The alternative solution worked for me on my minikube setup. However the same was not working on my rancher setup. The `window.WEB_ROOT` was getting set to the entire browser path including the protocol and domain name. So the browser window was getting rewritten as `https://128.xx.xx.xx/https://128.xx.xx.xx/path/to/kiali-service/kiali:20001/proxy`. To fix this I modified the regex and it is working fine on my setups. Can you please review this change and let me know if it looks good to you?

routing/router.go
```diff
@@ -155,9 +155,7 @@ func serveEnvJsFile(w http.ResponseWriter) {
 		body += fmt.Sprintf("window.HISTORY_MODE='%s';", conf.Server.WebHistoryMode)
 	}
 
-	if webRoot := strings.TrimSuffix(conf.Server.WebRoot, "/"); len(webRoot) > 0 {
-		body += fmt.Sprintf("window.WEB_ROOT='%s';", webRoot)
-	}
+	body += "window.WEB_ROOT = document.getElementsByTagName('base')[0].getAttribute('href').replace(/^https?:\\/\\/[^#?\\/]+/g, '').replace(/\\/+$/g, '')"
```
jmazzitelli, line 158: Just for clarification and to document it here, can you explain what the regex is expected to do?

Here's my guess.

The first regex will replace any `http://<hostname>` or `https://<hostname>` with an empty string, leaving just the full path.

The second regex is the one originally proposed - it replaces everything in that full path, up to and including the last forward slash, with empty string.

Thus leaving the last path in the URL, which is assigned to WEB_ROOT.

@doflamingo721 @israel-hdez please confirm or comment with what we WANT those regexs to do. I assume we want WEB_ROOT to be the last path in the URL (??)

jmazzitelli (commented):
@doflamingo721 @israel-hdez see my question below

routing/router.go
doflamingo721: > The first regex will replace any http://<hostname> or https://<hostname> with an empty string, leaving just the full path.

Yes that is correct.

doflamingo721: > The second regex is the one originally proposed - it replaces everything in that full path, up to and including the last forward slash, with empty string.
My understanding is that the second regex is replacing the forward slash at the end of the URL with an empty string thus removing it.

jmazzitelli:
This new solution works nicely. No need for a separate and new config option.

I tested with the k8s proxy and it works through that. I then started a port-forward and went directly to that and it, too, worked. So this solution works.

jmazzitelli (approved):
I approve, but I would like to give @israel-hdez a chance to comment before merging.

doflamingo721:
Hey @israel-hdez, can we get some eyes on the change so that it is good to go as mentioned by @jmazzitelli . Thanks.

jmazzitelli:
@israel-hdez is on vacation - he will be back Wednesday.

routing/router.go
israel-hdez: Right, the regex originally proposed is basically a trimEnd of the `/` slash. It is done with a regex because, AFAIK, the `trimEnd` func in Javascript only works with blanks (it does not receives what character you want to trim).

israel-hdez (approved):
I tried with `kubectl proxy`, `kubectl port-forward` and exposing with nginx-ingress and all works good.
I think this PR is good.

PR #4916, opened 2022-04-06
[cypress] login bugfix for stage

ScriptingShrimp:
Bugfix: UI login flow for OCP instances

### DIFF
--- a/frontend/cypress/support/commands.ts
+++ b/frontend/cypress/support/commands.ts
@@ -46,12 +46,16 @@ Cypress.Commands.add('login', (provider: string, username: string, password: str
 				cy.intercept('api/authenticate').as('authorized') //request setting kiali cookie
 				cy.log(
 					`provider: ${provider}, 
-						username: ${username},
-						atuhorization: ${auth_strategy}`)
+					username: ${username},
+					auth_strategy: ${auth_strategy}`)
 				cy.visit('/');
 
 				if (auth_strategy === 'openshift') {
-					cy.get('.pf-c-form').contains(provider, { matchCase: false }).click()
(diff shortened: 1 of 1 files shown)


frontend/cypress/support/commands.ts
```diff
@@ -46,12 +46,16 @@ Cypress.Commands.add('login', (provider: string, username: string, password: str
...
 					`provider: ${provider}, 
-						username: ${username},
-						atuhorization: ${auth_strategy}`)
+					username: ${username},
+					auth_strategy: ${auth_strategy}`)
 				cy.visit('/');
 
 				if (auth_strategy === 'openshift') {
-					cy.get('.pf-c-form').contains(provider, { matchCase: false }).click()
+					cy.get('.pf-c-form').contains(auth_strategy, { matchCase: false }).click()
```
lucasponce, line 54: I wonder if we can use these first steps with the cypress framework to also introduce the changes in the UI components to not rely on styles but on "data-" attributes (as we are introducing in other PRs).

In that case, the changes would be located in the same work and it would be easier to review them.

The ".pf-c-form" style and similar may change without notice.

In case you need some examples to follow the same conventions, the PR that introduced the "overview.feature" added some attributes to simplify the selector.

ScriptingShrimp: It would be nice to do the "data-" attribute properly, but currently its not possible for me to do so. 
- I do lack proper knowledge of frontend framework, but more important:
- This change needs to be reverse compatible for OCP nightly and downstream testing

Don't get me wrong, I'm all in for "data-" attributes, but this one is not right candidate.

Looking forward for better example :+1:

lucasponce (approved):
Ok, we expect to maintain these tests so we can update the selectors in future iterations.

PR #4917, opened 2022-04-06
Fix CI yaml

nrfox:
Lost this line when rebasing #4905

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -197,3 +197,4 @@ jobs:
 
       - name: Get kiali pod logs when tests fail
         if: ${{ steps.integration-tests.outcome == 'failure' }}
+        run: kubectl logs deployments/kiali -n istio-system


lucasponce (approved):

lucasponce (approved):

.github/workflows/kiali-ci.yml
```diff
@@ -197,3 +197,4 @@ jobs:
 
       - name: Get kiali pod logs when tests fail
         if: ${{ steps.integration-tests.outcome == 'failure' }}
+        run: kubectl logs deployments/kiali -n istio-system
```
jmazzitelli, line 200: It is possible the name could be different if our tests ever want to change the `deployment.instance_name` CR or helm chart value. So, FYI, it would be better if we use something like `-l app.kubernetes.io/name=kiali` - we always have that label, regardless of the instance name of the kiali pod. So:

`kubectl logs -n istio-system -l app.kubernetes.io/name=kiali`

PR #4931, opened 2022-04-06
Adjust dependabot to be less noisy

lucasponce:


jshaughn (approved):
+1 to less frequency and no prs, we'll respond to alerts manually

PR #4937, opened 2022-04-07
Update README and add hack script for demo setup

nrfox:
Update README with cypress requirements and add a hack script for installing demos necessary for cypress tests.

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -16,10 +16,10 @@ jobs:
       # set when this workflow is run against a PR. When run against a PR, we want
       # to use the branch we are running against: GITHUB_BASE_REF. When run
       # as a release, we want to use the current branch ref: GITHUB_REF_NAME.
-      target-branch: ${{ github.base_ref || github.ref_name }}    
+      target-branch: ${{ github.base_ref || github.ref_name }}
     steps:
-      # The initialize job gathers variables for later use in jobs. 
-      # We are using this technique rather environment variables because at the moment, they won't work with reusable jobs. 
+      # The initialize job gathers variables for later use in jobs.
+      # We are using this technique rather environment variables because at the moment, they won't work with reusable jobs.
       # A positive side effect of this is that we can print all variables at start for debugging and troubleshooting.
       - run: echo "target-branch -> ${{ github.base_ref || github.ref_name }}"
   build_backend:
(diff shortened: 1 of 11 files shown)


hack/istio/install-testing-demos.sh
```diff
@@ -0,0 +1,58 @@
...
+fi
+
+echo "CLIENT_EXE=${CLIENT_EXE}"
+echo "IS_OPENSHIFT=${IS_OPENSHIFT}"
+
+# Installed demos should be the exact same for both environments.
+# Only the args passed to the scripts differ from each other.
+if [[ "${IS_OPENSHIFT}" = "true" ]]; then
+  echo "Deploying bookinfo demo..."
+  /bin/bash "${SCRIPT_DIR}"/install-bookinfo-demo.sh
```
jmazzitelli, line 50: a) do you need `/bin/bash` ? The .sh files have the `#!/bin/bash` headers that tell the OS what shell to use.
b) the quotes around `"${SCRIPT_DIR}"` I assume are to protect against spaces in the path? If so, the placing of the quotes looks weird to me. I would think it should just be `"${SCRIPT_DIR}/install-bookinfo-demo.sh"` (the entire string wrapped)

.github/workflows/kiali-ci.yml
```diff
@@ -202,5 +202,5 @@ jobs:
         working-directory: tests/integration/tests
 
       - name: Get kiali pod logs when tests fail
-        if: ${{ steps.integration-tests.outcome == 'failure' }}
+        if: steps.integration-tests.outcome == 'failure'
         run: kubectl logs deployments/kiali -n istio-system
```
jmazzitelli, line 206: this is weird, i thought we already changed this? Yesterday @leandroberetta had changed this, I thought? This is to ensure we get the logs regardless of the `deployment.instance_name` we may have tested with:
```suggestion
        run: kubectl logs -l app.kubernetes.io/name=kiali -n istio-system
```

leandroberetta: @jmazzitelli my PR is not merged yet, changes are still there.

README.adoc
```diff
@@ -246,6 +246,16 @@ cd $KIALI_SOURCES/kiali/hack
...
 
+Integration testing is done with cypress. More information can be found link:./frontend/cypress/README.md[here].
+
+Requirements:
+- Istio
+- Kiali
+- bookinfo demo app
+- error rates demo app
+
+You can use link:./hack/install-testing-demos.sh[this script] to install all the neccessary demo apps for testing. The script supports both openshift and non-openshift deployments.
```
jmazzitelli, line 257:
```suggestion
You can use link:./hack/istio/install-testing-demos.sh[this script] to install all the neccessary demo apps for testing. The script supports both openshift and non-openshift deployments.
```

jmazzitelli (changes requested):

jmazzitelli (approved):

jmazzitelli (approved):

nrfox:
Some of the tests are starting to flake. My guess is it's related to the increased resource usage from deploying additional demo apps. The GH actions VMs only have 7 GB of mem and 2 CPU so they are pretty resource constrained. I'm going to try tweaking some of the resource requirements for various things to see if we can get it stable again.

nrfox:
Adding "do not merge" label until test flakiness has been resolved. Will be pushing commits for testing CI changes.

jmazzitelli (approved):

jmazzitelli (approved):

nrfox:
This is ready for :eyes: again @jmazzitelli. Turns out the flaky tests were from patternfly changing the dom id for a `Select` element when you don't pass one explicitly. Still I think it's worth keeping the resource adjustments for the proxies included in this PR to try and minimize the footprint.

PR #4943, opened 2022-04-08
Merged Kiali UI and Kiali Server can point to a single commit

aljesusg:
Backport 1.48


lucasponce (approved):
Tested on 1.48 branch:
image

lucasponce:
looks good, only have a side comment if this needs something at the operator side for 1.48? cc @jmazzitelli

jmazzitelli (approved):
This is OK. The only thing we might consider on the operator repo is backport this - https://github.com/kiali/kiali-operator/pull/512 but this is not necessary for two reasons:
1) We do not run the molecule tests on any branch other than master
2) That specific api-test is soon going to be entirely deleted (once the backend api tests are up and running in CI).

So I suggest we do not bother with backporting anything in the operator repo for this issue.

PR #4946, opened 2022-04-08
Add version to package.json

aljesusg:
Cachito require version in the package.json

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,5 +1,6 @@
 {
   "name": "@kiali/kiali-ui",
+  "version": "1.49.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
   "keywords": [
     "istio service mesh",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kiali/kiali-ui",
-  "version": "1.49.0",
+  "version": "1.50.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
   "keywords": [
     "istio service mesh",


jmazzitelli (dismissed):
we'll need to make sure we update this when we do releases (our current Jenkins job already did that for the kiali-ui package.json - just need to make sure we do the same now with the new CI release job)

aljesusg:
> we'll need to make sure we update this when we do releases (our current Jenkins job already did that for the kiali-ui package.json - just need to make sure we do the same now with the new CI release job)

Yes...In fact we should add some test in the CI to check that version in frontend/package.json is the same than core version

.github/workflows/version.yml
```diff
@@ -0,0 +1,25 @@
+name: Version
```
jmazzitelli, line 1: can we run this CI workflow only when either Makefile or package.json is touched by the commit/PR?

We should not have to run this for every commit/PR since it will be rare Makefile or package.json will be changing.

jmazzitelli:
I would just get the package.json change in now (in a separate PR)... worry about getting the CI workflow in later. In about 2 hours the sprint release will be kicked off - we want that version change in ASAP.

.github/workflows/version.yml
aljesusg: I think that yes we can

aljesusg:
Ok let me remove this latest commit and we can do it in another PR

aljesusg:
Moved validator to another PR to not block @jmazzitelli https://github.com/kiali/kiali/pull/4947

jmazzitelli (dismissed):

jmazzitelli:
companion PR: https://github.com/kiali/kiali/pull/4947

aljesusg:
Tests fails...this is not related with overview I don't know why fails

frontend/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@kiali/kiali-ui",
+  "version": "1.49.0",
```
jmazzitelli, line 3: Since 1.49.0 has already been released, this needs to be 1.50.0:
```suggestion
  "version": "1.50.0",
```

jmazzitelli (changes requested):

jmazzitelli (approved):

PR #4947, opened 2022-04-08
Add a test checker for version

aljesusg:
Working


jmazzitelli (approved):
I tested the grep/sed commands and they do what we want.
The rest looks good.
The only issue is - frontend/package.json does not (today) have a "version" field - we need that for this to work. So this PR is required: https://github.com/kiali/kiali/pull/4946

aljesusg:
Yes ...i am checking how to launch this only when Makefile or package.json change

lucasponce (changes requested):
I think this code could easily go into the "frontend" one ?
image
image

Is it a check of a value in two files, I don't think it requires launching another VM for it.

In any case @leandroberetta is organizing the structure of the flows so, let's wait until he comments.

jmazzitelli:
> I think this code could easily go into the "frontend" one ?
> ...
> Is it a check of a value in two files, I don't think it requires launching another VM for it.

I do not think it is enough for this check to just go in the frontend builder - we have to do this check when building backend, too. Because if someone changes one (frontend OR backend) we have to make sure it matches the other. In other words, if someone changes Makefile, we want to ensure it matches frontend's package.json. So that would require adding this check in the backend builder, too.

lucasponce:
But the pipeline executes both (backend, frontend).
Every flow is a new VM launched.
Do we need a new one?

I'm worried that we start adding flows, so, as Leandro is organizing this work, let's see what he thinks.

aljesusg:
> But the pipeline executes both (backend, frontend). Every flow is a new VM launched. Do we need a new one?
> 
> I'm worried that we start adding flows, so, as Leandro is organizing this work, let's see what he thinks.

But this flow is executed only when we change Makefile or package.json...We don't usually change it that often

jmazzitelli:
> But the pipeline executes both (backend, frontend).

I thought only the frontend builder would run when frontend/* files are touched. If no frontend/* files are changed in the PR, the frontend building would not run. Same with backend (if files NOT in frontend/* were touched, the backend builder would run).

If that is not the case (i.e. if the frontend and backend pipeline executes both, all the time) then I guess I don't understand what the benefit is of splitting up the frontend and the backend into two separate builders.

.github/workflows/version-checker.yml
```diff
@@ -0,0 +1,34 @@
...
+  # Run on master and release branches
+  push:
+    branches: [master, "^v\\d+\\.\\d+$"]
+    paths: 
+      - 'frontend/package.json'
+      - 'Makefile'
+  pull_request:
+    branches: [master, "^v\\d+\\.\\d+$"]   
+    paths: 
+      - 'frontend/package.json'
```
lucasponce, line 13: Ah, so, this "path" is part of the trigger condition.
Then is different.

lucasponce:
> But this flow is executed only when we change Makefile or package.json...We don't usually change it that often

Totally, I missed the "path" expression.
Then it makes sense.

lucasponce (approved):
The "on.pull_request.paths" expression will prevent this flow runs only when the Makefile and /or package.json.

I missed that in my previous review, then my concerns are gone from my side.

I'll wait to see if Leandro has any comments as it's part of his work to organize the pipelines.

leandroberetta:
> The "on.pull_request.paths" expression will prevent this flow runs only when the Makefile and /or package.json.
> 
> I missed that in my previous review, then my concerns are gone from my side.
> 
> I'll wait to see if Leandro has any comments as it's part of his work to organize the pipelines.

Hi, this LGTM.

leandroberetta (approved):
LGTM.

jmazzitelli:
> Splitting them up into two separate builders lets you run both simultaneously which speeds up the overall pipeline.

I assume that means the backend has a dependency on the frontend? Because how can the backend complete building if it doesn't have the UI frontend to include inside it?

Issue #4854, opened 2022-03-25
Create a UI suite test on cypress


lucasponce:
Kiali has no proper integration UI testing located in the upstream repo.

As part of the strategy to improve this, the "cypress" framework has been introduced in the "frontend" folder:

https://github.com/kiali/kiali/tree/master/frontend/cypress

But there are a couple of efforts to work on this area
- Migrate existing UI testing used by QE into "cypress" and/or redefine new UI test flows that cover all features of the project.
- Integrate the UI testing as part of the public pipelines that are being migrated into Github Actions.

@ScriptingShrimp is leading/coordinating the migration of the "legacy" UI test into cypress but the generation of UI tests would require effort from all the team.

I'm going to create this Epic as a reference for future subtasks.

Subtasks:
- [x] https://github.com/kiali/kiali/issues/4872
- [x] https://github.com/kiali/kiali/issues/4967
- [x] https://github.com/kiali/kiali/issues/4912
- [x] https://github.com/kiali/kiali/issues/4949
- [x] #5073
- [x] https://github.com/kiali/kiali/issues/5084
- [ ] UI tests around the Graph cytoscape toolbar
- [x] https://github.com/kiali/kiali/issues/4960
- [x] https://github.com/kiali/kiali/pull/4951
- [x] https://github.com/kiali/kiali/issues/4955
- [x] https://github.com/kiali/kiali/issues/4969
- [x] https://github.com/kiali/kiali/issues/5001
- [x] https://github.com/kiali/kiali/issues/5008
- [x] #5072
- [x] #5013
- [x] #5047
- [x] #5079
- [x] #5085

Subtasks specific to OpenShift testing support:
- [ ] authentication for jaeger, prometheus, other?

nrfox:
Can we get a list of the existing UI tests and what they cover so that we can be sure to cover the same things with the cypress tests? From there we can create subtasks for the various test cases that can be worked on in parallel by anyone. IMO we need this before we can jump into writing tests if the goal is to replace the legacy UI tests with cypress. @mattmahoneyrh @ScriptingShrimp @lucasponce

ScriptingShrimp:
@pbajjuri20 lets have a look asap.
I think we need more than day to have that list curated.

nrfox:
@ScriptingShrimp we're going to continue forward with creating some skeleton tests for the various pages and those can be compared against the existing polarion or selenium test cases and gaps can be filled iteratively.

ScriptingShrimp:
[...]

PR #4951, opened 2022-04-08
Cypress tests: Logs tab

israel-hdez:
WIP

Related to #4854


israel-hdez:
@ScriptingShrimp @nrfox OK, this is my initial attempt for a cypress test.
Just one scenario and it is still incomplete, since I'm kind-of reverse engineering patternfly to see what selectors work here...

Anyway, perhaps can I have a very, very, very, _veeeeeeeeery_ early review from you? Just to know if this is how it should be, before I start writing a whole suite.

I won't be available a few days, so you can take your time.

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,25 @@
+import { Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+
+Given("I am on the {string} workload detail page", (workload) => {
+    // Go to workloads list
+    cy.get("#Workloads").click();
+
+    // Select "bookinfo" namespace
+    cy.get("#namespace-selector").click();
+    cy.get("#bulk-select-id").uncheck();
+    cy.get('#namespace-list-item\\[bookinfo\\] input').check();
```
lucasponce, line 10: Perhaps better to pass the namespace as a parameter instead to hard-code the "bookinfo" one?

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,25 @@
...
+    cy.get("#namespace-selector").click();
+
+    // Open workload detail page
+    cy.get(`[data-cy=VirtualItem_Nsbookinfo_${workload}] a`).click();
+});
+
+When("I go to the Logs tab of the workload detail page", () => {
```
lucasponce, line 17: Same here, if possible, passing the namespace and workload name as parameters may help to reuse this expression.

frontend/src/components/VirtualList/VirtualItem.tsx
```diff
@@ -58,7 +58,7 @@ export default class VirtualItem extends React.Component<VirtualItemProps, Virtu
     const { style, className, item } = this.props;
     const key = 'VirtualItem_' + ('namespace' in item ? 'Ns' + item.namespace + '_' + item.name : item.name);
     return (
-      <tr style={style} className={className} role="row" key={key}>
+      <tr style={style} className={className} role="row" key={key} data-cy={key}>
```
lucasponce, line 61: "data-key" instead? Just trying to add "data-*" attributes more domain-specific, no need to use a "cypress" prefix for them, just annotate things that may identify the value.

https://github.com/kiali/kiali/commit/cd5cbffe1cfd9fa897a9d4a20ed584bfd3048b37#diff-9cd721a0ae66bcf3e0135496c5d43d07de1da7ae58eec132049b012b60031817R59

frontend/src/pages/WorkloadDetails/WorkloadDetailsPage.tsx
```diff
@@ -132,7 +132,7 @@ class WorkloadDetails extends React.Component<WorkloadDetailsPageProps, Workload
 
     if (!serverConfig.kialiFeatureFlags.disabledFeatures || !serverConfig.kialiFeatureFlags.disabledFeatures.includes('logs-tab')) {
       const logTab = (
-        <Tab title="Logs" eventKey={2} key={'Logs'}>
+        <Tab title="Logs" eventKey={2} key={'Logs'} data-cy={"workload-details-logs-tab"}>
```
lucasponce, line 135: Same here, "data-cy" -> "data-tab-type" or something more meaningful with the value (no need to be specific to cypress).

frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
```diff
@@ -436,7 +436,7 @@ export class WorkloadPodLogs extends React.Component<WorkloadPodLogsProps, Workl
 
   private getContainerLegend = () => {
     return (
-      <Form>
+      <Form data-cy={"workload-logs-pod-containers"}>
```
lucasponce, line 439: Same here.

frontend/src/components/VirtualList/VirtualItem.tsx
nrfox: The [recommendation](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) seems to be to have the key be static/constant and just have the value change. Seems to be common to include a custom command around this like:
```
Cypress.Commands.add('getBySel', (selector, ...args) => {
  return cy.get(`[data-test=${selector}]`, ...args)
})

cy.getBySel('value-of-data-test')
```
We should decide on what key we're using and just go with that. `data-cy` is fine with me.

lucasponce: I read that recommendation that the best practice is to provide a good context.
If there should be only one data- attribute, it can be anything, but I see that we need to add data- attributes in places where there are existing "aria-" or  "data-" ones and also we have patterns where the same data- may have different values.
In example, you may want to add a "namespace" and a "workload" name in some pattern.

So I encourage to use data attributes with context, so, if you are going to add a "namespace", better to use a "data-namespace" or "data-workload", there would be cases where you may have multiple lists/tables per namespace, rendering workloads/apps. I see it confusing to use the same "data-cy" where in some cases the value is a namespace and in others a workload.

I agree that if it's something new and there is no clear attribute, the author can define the "data-" naming that better suits.

nrfox: > So I encourage to use data attributes with context, so, if you are going to add a "namespace", better to use a "data-namespace" or "data-workload", there would be cases where you may have multiple lists/tables per namespace, rendering workloads/apps. I see it confusing to use the same "data-cy" where in some cases the value is a namespace and in others a workload.

The context goes in the value e.g. `data-cy="workload-details-logs-tab"` which is consistent with other html tags like `class="expanded"` or `aria-label="filter_select_type"`. This is how the cypress real world app is setup: https://github.com/cypress-io/cypress-realworld-app/search?q=data-test. It also seems to be the convention that the openshift console uses: https://github.com/openshift/console/search?q=data-test. The "Selector Playground" optimizes for using `data-cy` or `data-test`. It's nice to be able to quickly search for all "special" testing elements using a consistent attribute and it lets you add some helper functions like the command above. Can we start with `data-cy` or `data-test` and, if need be, add additional `data-test-*` attributes later?

lucasponce: I'd prefer that if we have a "data-namespace", we keep using the same when you need to add a "data-namespace".
What I don't see good is to use a data-cy={namespace} in one place (a <ul> element) and a data-cy={workload} in an inner one. It's better to use a data-namespace={bookinfo} or a data-workload={a-client}, then after some months, everyone should understand what type of value needs that attribute.

Or at least, I suggest to use a data-test-namespace={'bookinfo'}, data-test-workload={'a-workload'], etc

Please, let's unify conventions now that we are starting, I don't mind updating the tests I wrote, but let's not have different ways of naming depending on the author, otherwise, we will outdated these tests as soon as some fail as we simply update the value to comply (as it happened with some of the existing JEST tests the project have).

lucasponce: Or we can use a "data-cy" when there is only one attribute and use "data-cy-*" when there is more than one close to the same area (I'm thinking in a ul -> li, in cases like that I'd prefer to have a data-cy-namespace/data-cy-workload).

nrfox: > What I don't see good is to use a data-cy="namespace" in one place (a "ul" element) and a data-cy="workload" in an inner one. It's better to use a data-namespace="bookinfo" or a data-workload="a-client", then after some months, everyone should understand what type of value needs that attribute.

Agreed just having the `namespace` as the value probably doesn't give you enough context and can easily conflict with other elements that may share a `data-cy=namespace`. You can provide additional context in the value however like `data-cy=list-item-${item.type}-${item.namespace}`. Here's a similar example from the cypress demo app: https://github.com/cypress-io/cypress-realworld-app/blob/develop/src/components/UserListItem.tsx#L14.

Having a `data-namespace` attribute suffers from a similar lack of context. If you've got multiple elements on the screen with `data-namespace` attributes, then it's ambiguous which one you select with `cy.get('[data-namespace="bookinfo"]')`. It could also lead to a proliferation of attribute types: `data-workload`, `data-service`, `data-app`, `data-virtualservice`. To me, it's more unclear when you use/create a new attribute type vs just using `data-cy` or `data-test`.

> Please, let's unify conventions now that we are starting

Yes! Let's do this. Ultimately we can choose whatever convention works best for us. I would just prefer we start with the recommended way and then as we learn more and get more experience with cypress we can deviate from that. I'd also prefer we use `data-test` or `data-test-*` over `data-cy`. `cy` has too much overlap with cytoscape and `data-test` is more understandable that it's an attribute for testing purposes only.

lucasponce: I like to use the `data-test` as a general rule and `data-test-*` if there are cases that may require more.
About the concatenating in the string the context, it can be a start, but I guess it will depend on a case by case.

From the cypress examples, we can get some tips but we can adapt, the project is mature and this is more thinking about what would be easier to maintain in several months/years (we have some lessons learned from the JEST tests and that's why I wanted to discuss this even if "naming" may be over-discussed).

I'll start from my side reviewing the tests I did for the overview page and I'll adapt them to use the "data-test" / "data-test-*" to move into that direction.

frontend/cypress/integration/common/workload_logs.ts
israel-hdez: This is supposed to be used when the view is already the detail page. So, the test is supposed to have the form:
```
Given I am on the "productpage-v1" workload detail page of the "bookinfo" namespace
When I go to the Logs tab of the workload detail page
Then ...
```

Adding the namespace and the workload would be redundant and would be unused under the assumption that the view is already a detail page...

So, I think this can be easily reused as long as the proper `Given` expressions are used.

frontend/src/components/VirtualList/VirtualItem.tsx
israel-hdez: I also changed this PR to use `data-test`

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,122 @@
...
+    let containerListForm = cy.get("[data-test=workload-logs-pod-containers]");
+
+    let podLabel = containerListForm.get("label").contains(containerName);
+    podLabel.then($podLabel => {
+        let logColor = $podLabel[0].style.color;
+
+        let logPane = cy.get('#logsText');
+        logPane.get('p').each(line => {
+            expect(line[0].style.color).to.equal(logColor);
+        })
```
nrfox, line 113: Is there a way to check all the text at once as opposed to line by line? Might not be. I think the default log limit is 100 so probably not a big deal.

nrfox (changes requested):

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,122 @@
...
+    containerListForm.get('[type=checkbox]:checked').uncheck();
+
+    let podLabel = containerListForm.get("label").contains(containerName);
+    podLabel.invoke('attr', 'for')
+        .then(labelFor => {
+            cy.get(`#${labelFor}`).check();
+        });
+});
+
+When("I enable visualization of spans", function () {
```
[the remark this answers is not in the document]
israel-hdez: > Are we using first person or "the user" for these assertions? I don't care either way as long as it's consistent.

I was mainly following the style on the [cucumber docs](https://cucumber.io/docs/gherkin/reference/). I saw they tend to prefer first person, but use 3rd person if there are several personas in the scenario.

I don't have preference. You tell me what style to use.

> Same with using `function() {}` vs arrow function `() => {}`

This is me being lazy. The `function() {}` syntax was generated by my IDE. I simply left it untouched.
I think most code is currently using arrow functions. So, I think I should change this one to an arrow function for consistency.

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,122 @@
...
+
+    // Select "bookinfo" namespace
+    cy.get("#namespace-selector").click();
+    cy.get("#bulk-select-id").uncheck();
+    cy.get(`#namespace-list-item\\[${namespace}\\] input`).check();
+    cy.get("#namespace-selector").click();
+
+    // Open workload detail page
+    cy.get(`[data-test=VirtualItem_Nsbookinfo_${workload}] a`).click();
+}
```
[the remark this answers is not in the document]
israel-hdez: > IMO this should be part of the test "Background" in the feature file.

Well... It is not in the Background, but it is part of `Given` clauses. I chose to not use the Background, because explicitly writing the Given clause in each Scenario gives much better clarity to the Scenarios. Also, in the [Gherkin reference, it's mentioned that the background](https://cucumber.io/docs/gherkin/reference/#background) is for "incidental details", but in this case I think this is properly helping to describe the Scenarios.

Anyway, I'm an engineer and I think I somewhat share the idea of DRYing up the code. So, if you tell me that it's better in the Background, I won't put any resistance. So, do you really want that I move it to the Background?

> Something like `Given user is at the "/namespaces/bookinfo/workloads/details-v1?tab=logs" page`

Placing the URL in the sentence makes it handy to reuse the Given clause for other pages. But, again, I'm mainly following the docs and I understand this should be human readable. IMO, a URL becomes technical.

Again, I'm an engineer and I think we won't pass this to non-engineer people. So, if you lean towards usability rather than adhering to the Gherkin language ideas, them I'm quite OK with writing URLs on the sentences. You tell.

> Perhaps testing the link to the logs tab can be a part of the workload details page testing.

I think this is an implementation detail. Simply, it is more hypnotizing the way I have it implemented :-)
But here I give you the reason. I will remove all this code in favor of a simple `cy.visit()`.

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,122 @@
...
+
+When("I select only the {string} container", containerName => {
+    let containerListForm = cy.get("[data-test=workload-logs-pod-containers]");
+    containerListForm.get('[type=checkbox]:checked').uncheck();
+
+    let podLabel = containerListForm.get("label").contains(containerName);
+    podLabel.invoke('attr', 'for')
+        .then(labelFor => {
+            cy.get(`#${labelFor}`).check();
+        });
```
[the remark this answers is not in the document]
israel-hdez: Hmm... I didn't know that assigning to variables was not recommended. I will change this to use `within`, which seems to do what I want.

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,122 @@
...
+            cy.get(`#${labelFor}`).should('be.checked');
+        });
+});
+
+Then("I should see some {string} pod selected in the pod selector", podNamePrefix => {
+    cy.get('#wpl_pods-toggle').should('include.text', podNamePrefix);
+});
+
+Then("the log pane should only show log lines containing {string}", filterText => {
+    let logPane = cy.get('#logsText');
```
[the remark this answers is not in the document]
israel-hdez: I _do_ want to assert that each paragraph (i.e. each line in the log pane) has `filterText`.

I think if I do `cy.get('#logsText').should('contain.text', filterText);` it will be enough if one paragraph (i.e. one line in the log pane) contains `filterText` for the assertion to pass. But I want to check all log lines.

I also could try `cy.get('#logsText').get('p').should('contain.text', filterText);`. I incidentally tried this on the `Then("the log pane should show spans")` (it's at the very bottom). It turns out that, despite the `get('p')` part would match multiple elements, the `should` won't check all matched elements :-/

frontend/cypress/integration/common/workload_logs.ts
israel-hdez: Hmm, there is an `all` chainer for  the `expect` assertion. Let me try it...

israel-hdez: Hmm... Nope, that `all` chainer only lets you follow up with `keys` or `member`. Not any property (here I would want `all.css`).

frontend/cypress/integration/common/workload_logs.ts
```diff
@@ -0,0 +1,122 @@
...
+    logPane.get('p').each(line => {
+        expect(line).to.contain(filterText);
+    });
+});
+
+Then("the log pane should only show log lines not containing {string}", filterText => {
+    let logPane = cy.get('#logsText');
```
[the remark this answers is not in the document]
israel-hdez: See my comment up :point_up:  there.

frontend/cypress/integration/common/workload_logs.ts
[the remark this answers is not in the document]
israel-hdez: It should be better now.

lucasponce:
Please @israel-hdez, would you mind rebasing with the master branch?
I'm testing it locally but it's not synced with the latest ones.

Thanks

frontend/cypress/integration/featureFiles/workload_logs.feature
```diff
@@ -0,0 +1,43 @@
+Feature: Workload logs tab
+  The Logs tab of a specific workload allows to see the generated logs of
+  its associated pods. If the workload is backed by more than one pod, the user
+  can choose which pod to see its logs. If the pod has multiple containers, it
+  is possible to view logs from all containers in a single view.
+
+  Background:
+    Given I open Kiali URL
+    Given user is at administrator perspective
```
lucasponce, line 9: These "cucumber" should be more "natural language" syntax.

Per previous comments, start with a "Given" and chained with "And".

frontend/cypress/integration/featureFiles/workload_logs.feature
```diff
@@ -0,0 +1,43 @@
...
+
+  Scenario: The logs tab should show the logs of a pod
+    Given I am on the "productpage-v1" workload detail page of the "bookinfo" namespace
+    When I go to the Logs tab of the workload detail page
```
lucasponce, line 13: "Given" ... "And" ...

frontend/cypress/integration/featureFiles/workload_logs.feature
lucasponce: It's only "sugar" syntax as those will work, but just following @ScriptingShrimp recommendations here and trying to have the same criteria.

lucasponce (approved):
Thanks for the work.
Some comments, but no blocker to consolidate.
Tests look good, and most of the discussions about naming are addressed:
image

I expect that these tests will evolve and will have an active maintenance as they should be updated as soon as we change the page, so to me if the CI is happy it's ok to consolidate.

frontend/cypress/integration/featureFiles/workload_logs.feature
israel-hdez: Yes, it was a small slip. I fixed it.

frontend/cypress/integration/featureFiles/workload_logs.feature
israel-hdez: I think here the wording is fine. As you say, the "Given" followed by "When" makes sense.

israel-hdez:
@lucasponce In case you want to re-test, I just rebased.

frontend/cypress/integration/common/workload_logs.ts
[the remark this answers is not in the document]
nrfox: Personally I see it as Background where you can replace the statement `Given I open Kiali URL` with something like `Given I open the workload details page`. Visiting the root kiali URL is not relevant to this test and just adds more time having to load the page twice.

Regarding the format, I haven't read any of the Gherkin docs and only written a couple tests in this format. I'm fine with whatever format works best for us. If that's a more human readable one, perhaps that can still be parametrized like `Given user is at the "workload details" page on the "logs" tab` or something like that. I see the wording as less important than the implementation. If we can re-use the same implementation across all or most of the tests then that will make writing future tests and maintaining existing ones easier.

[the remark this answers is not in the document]
israel-hdez: > Personally I see it as Background where you can replace the statement `Given I open Kiali URL` with something like `Given I open the workload details page`. Visiting the root kiali URL is not relevant to this test and just adds more time having to load the page twice.

Hmm... Your comment made me realize that I probably don't need the `Given I open Kiali URL` in the Background. I think it is enough to have `Given user is at administrator perspective` which AFAIU cannot be replaced. That probably will solve the double-load.

> If that's a more human readable one, perhaps that can still be parametrized like `Given user is at the "workload details" page on the "logs" tab` or something like that.

Well, if we will use this format, I think I prefer to _not_ parametrize and have multiple `Given()` in the code that specifically do the needed `cy.visit()`, rather than a big `switch/if` deciding/forming the URL to visit.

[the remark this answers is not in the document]
nrfox: > Well, if we will use this format, I think I prefer to not parametrize and have multiple Given() in the code that specifically do the needed cy.visit(), rather than a big switch/if deciding/forming the URL to visit.

Either way is fine with me. The implementations can all share a common function if we want to apply options to all urls like `?refresh=0`. On the downside we'll need to implement more "Given" statements for every feature file. Trading off having more shared step definitions for better readability.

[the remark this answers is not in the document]
israel-hdez: OK, I just removed the `Given I open Kiali URL`. Now, the background only has `Given user is at administrator perspective`.

BTW @nrfox, this `goToWorkloadDetailPage` function does no longer exist. Perhaps, can you take another look to see if current state is what you would expect?

I think, at the moment, this is the only blocking comment. So, if you are happy now, I hope we can go ahead and merge.

nrfox (approved):

lucasponce (approved):

PR #4958, opened 2022-04-12
Fix flaky frontend test

nrfox:
Hopefully fixes flaky CI test. This is hard to reproduce locally but failures are frequent in CI since the tests run slower there. Some different examples:
- https://github.com/kiali/kiali/runs/5980401734?check_suite_focus=true
- https://github.com/kiali/kiali/runs/5980025014?check_suite_focus=true

This change makes separate calls to `cy.get(...)` between clicks so cypress will wait for the inner element to exist before trying to click it.


lucasponce (approved):
Tested locally and no errors.
Also, the CI looks happy as well.

aljesusg (approved):
LGFM

PR #4959, opened 2022-04-13
Added Spans and Traces endpoints tests for Service, App and Workload.

hhovsepy:

Added more tests for Workload, Services and Applications endpoints.
Graph nodes tests added.


tests/integration/tests/traces_spans_test.go
```diff
@@ -0,0 +1,91 @@
...
+				assert.Equal(span.TraceID, traces.Data[0].TraceID)
+			}
+		}
+	} else {
+		assert.True(false)
+	}
+}
+
+func assertSpans(spans []jaeger.JaegerSpan, statusCode int, err error, assert *assert.Assertions) {
+	if statusCode == 500 {
```
nrfox, line 76: A 500 from the kiali server doesn't always mean jaeger is not configured. It could be any internal server error right? That's part of what we want to catch with these tests. Does kiali give a specific error message along with the `500` status code that you can check when jaeger isn't configured?

tests/integration/tests/traces_spans_test.go
```diff
@@ -0,0 +1,91 @@
...
+		assert.Nil(err)
+		assert.NotNil(spans)
+		if len(spans) > 0 {
+			assert.NotNil(spans[0].TraceID)
+			assert.NotEmpty(spans[0].References)
+			assert.NotNil(spans[0].References[0].TraceID)
+			assert.Equal(spans[0].TraceID, spans[0].References[0].TraceID)
+		}
+	} else {
+		assert.True(false)
```
nrfox, line 89: This is kind of an odd assertion. Is the goal here to fail the test? There might be an `assert.Fail('reason for failure msg')` method you can use to give more clarity.

tests/integration/tests/traces_spans_test.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+}
+
+func assertSpans(spans []jaeger.JaegerSpan, statusCode int, err error, assert *assert.Assertions) {
+	if statusCode == 500 {
+		// skip the tests as Jaeger is not configured
+		assert.True(true)
```
nrfox, line 78: Instead of having this assertion here you can just return from this method. Then you don't need to include this in the `if/else if` block and it becomes a little easier to read.
```
if statusCode ... {
  return
}

if statusCode == 200 {
...
} else {
}
...
```

tests/integration/tests/traces_spans_test.go
nrfox: imo if we expect jaeger to always be configured properly in the e2e testing env prior to running the tests, we shouldn't ignore these cases since they could signal a real problem with Kiali.

hhovsepy:
hi @nrfox, PR is ready to be reviewed please, also added more testcases

tests/integration/tests/graph_nodes_test.go
```diff
@@ -0,0 +1,82 @@
...
+	assert := assert.New(t)
+	name := "reviews-wrong"
+	graphType := "workload"
+	assertEmptyGraphConfig("workloads", graphType, utils.BOOKINFO, name, assert)
+}
+
+func assertGraphConfig(objectType, graphType, namespace, name string, assert *assert.Assertions) {
+	graph, _, _ := utils.Graph(map[string]string{"graphType": graphType, "namespaces": namespace})
+	// TODO better way to check if there are any graph nodes at all to be able to verify requested ones
+	if len(graph.Elements.Nodes) > 0 && len(graph.Elements.Edges) > 0 {
```
nrfox, line 63: If the graph is empty, should we fail here instead of skipping the asserts when `len(Nodes) && len(Edges) == 0`?

tests/integration/tests/graph_test.go
```diff
@@ -0,0 +1,265 @@
...
+		switch key {
+		case "duration":
+			assert.Contains(value, fmt.Sprintf("%d", graph.Duration))
+		case "graphType":
+			assert.Equal(value, graph.GraphType)
+		}
+	}
+	assert.NotNil(graph.Elements.Nodes)
+	assert.NotNil(graph.Elements.Edges)
+	// TODO better way to check if there are any graph nodes at all to be able to verify requested ones
```
nrfox, line 251: This is because the test only looks at whether or not nodes exist but doesn't make any assertions about those nodes? e.g. `TestDisplayServiceNodes` doesn't verify that there are service nodes in the graph just that there are nodes in the graph?

nrfox (commented):
Some small nitpicks otherwise LGTM

tests/integration/tests/graph_nodes_test.go
hhovsepy: yes, added wait there

tests/integration/tests/graph_test.go
hhovsepy: added wait there

tests/integration/tests/graph_badges_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+var VERSIONED_APP_PARAMS = map[string]string{"graphType": "versionedApp", "duration": DURATION, "injectServiceNodes": "true"}
+var WORKLOAD_PARAMS = map[string]string{"graphType": "workload", "duration": DURATION, "injectServiceNodes": "true"}
+var APP_PARAMS = map[string]string{"graphType": "app", "duration": DURATION, "injectServiceNodes": "true"}
+var SERVICE_PARAMS = map[string]string{"graphType": "service", "duration": DURATION, "injectServiceNodes": "true"}
+
+func TestCBVersionedApp(t *testing.T) {
+	assert := assert.New(t)
+	assertGraphBadges(VERSIONED_APP_PARAMS, CIRCUIT_BREAKER_FILE, CB_BADGE, assert)
+}
```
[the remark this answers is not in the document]
hhovsepy: I was thinking to optimize tests by having one yaml file to apply and then run all checks, but this is causing failures of scenarios. So leaving tests separately to be more readable. Thank you for reviewing.

lucasponce:
> Looks like the backend tests are passing in CI and it is only the frontend tests that are flaking

I'm investigating this, I think the PR is adding the traffic generator workload, but it's platform-dependent:

image

And I think is not well prepared for kind, which creates some errors in the cluster and is shown in one of the UI tests, and potentially it adds some overhead on top of the existing one.

@hhovsepy the traffic generator is needed for all the integration tests? I'm going to see if I can prepare a fix for the CI case.

PR #4963, opened 2022-04-14
`make yarn-start`: replace proxy line when env var set

nrfox:
Remove the check for an existing `proxy` line during `make yarn-start` so that the `proxy` line will always be replaced with the value of `YARN_START_URL` even if an old `proxy` line exists in the `package.json`.

### DIFF
--- a/make/Makefile.ui.mk
+++ b/make/Makefile.ui.mk
@@ -4,13 +4,11 @@
 
 ## yarn-start: Run the UI in a local process that is separate from the backend. Set YARN_START_URL to update package.json.
 yarn-start:
-	@if ! (cat ${ROOTDIR}/frontend/package.json | grep -q "\"proxy\":"); then \
-	  if [ -z "${YARN_START_URL}" ]; then \
-	    echo "${ROOTDIR}/frontend/package.json does not have a 'proxy' setting and you did not set YARN_START_URL. Aborting."; \
-	    exit 1; \
-	  else \
-	    sed -i "2 i \ \ \"proxy\": \"${YARN_START_URL}\"," ${ROOTDIR}/frontend/package.json; \
-	  fi; \
+	@if [ -z "${YARN_START_URL}" ]; then \
+		echo "${ROOTDIR}/frontend/package.json does not have a 'proxy' setting and you did not set YARN_START_URL. Aborting."; \
+		exit 1; \
+	else \
+		sed -i "2 i \ \ \"proxy\": \"${YARN_START_URL}\"," ${ROOTDIR}/frontend/package.json; \
 	fi
 	@echo "'yarn start' will use this proxy setting: $$(grep proxy ${ROOTDIR}/frontend/package.json)"
 	@cd ${ROOTDIR}/frontend && yarn start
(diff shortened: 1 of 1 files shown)


make/Makefile.ui.mk
```diff
@@ -4,13 +4,11 @@
...
 yarn-start:
-	@if ! (cat ${ROOTDIR}/frontend/package.json | grep -q "\"proxy\":"); then \
-	  if [ -z "${YARN_START_URL}" ]; then \
-	    echo "${ROOTDIR}/frontend/package.json does not have a 'proxy' setting and you did not set YARN_START_URL. Aborting."; \
-	    exit 1; \
-	  else \
-	    sed -i "2 i \ \ \"proxy\": \"${YARN_START_URL}\"," ${ROOTDIR}/frontend/package.json; \
-	  fi; \
+	@if [ -z "${YARN_START_URL}" ]; then \
+		echo "${ROOTDIR}/frontend/package.json does not have a 'proxy' setting and you did not set YARN_START_URL. Aborting."; \
```
jmazzitelli, line 8: this error message is now incorrect. You actually might have a proxy setting - now it doesn't matter. The error is "You did not set YARN_START_URL."

This is why I had it like it was before - because if I already have a proxy URL defined in my package.json, I can just say `make yarn-start` without having to always remember what the URL is. The way it was before, I only needed to figure out the URL the first time. With this PR, now I'll always have to remember to set YARN_START_URL which means I'll always have to remember what it should be. Kinda defeats the purpose - if I have to remember to always pass that in, I might as well just edit package.json and only have to do THAT one time and just run "yarn start" thereafter.

So, I'm not crazy about this PR, unless there is a reason why we need it.

jmazzitelli (commented):
I'm a `meh` on this PR, unless you have a good reason why we need it.

make/Makefile.ui.mk
jmazzitelli: If the issue is you want to change the URL if it is already set, you can add another env var for that:

`make -e YARN_START_URL_FORCE="http://...." `

or something like that. Perhaps just a boolean:

`make -e YARN_START_FORCE_URL_UPDATE="true"`

jmazzitelli: or, if you want the feature you have, do the reverse to support my thing:
`make -e YARN_START_URL_IF_NOT_SET="http...."`
or something like that.. maybe have a boolean instead? `YARN_START_IGNORE_IF_SET=true`

nrfox: How about if you provide a `YARN_START_URL` then it should either add one if it doesn't exist or replace the existing one. If you don't provide a `YARN_START_URL` and there's an existing `proxy` line then it'll use that. If you don't provide `YARN_START_URL` and there's no proxy line, then it'll yell at you to add one.

Here's how I use this command:
```
make -C ~/Developer/redhat/kiali -e YARN_START_URL=$(kiali_minikube_url) yarn-start
```

The way it currently is, if I have an existing `proxy` setting then it will use that instead of what I passed with `YARN_START_URL`.

jmazzitelli (approved):
tried all the permutations - all works. LGTM

PR #4968, opened 2022-04-18
[cypress] Tests for services list and helper cmd

### LINKED ISSUE #4955: Add cypress UI tests around the Services List page
Add cypress tests for Services List page.

Epic: #4854

nrfox:
Adds cypress tests for the services list page. Adds a `getBySel` custom command to select an element based on the `data-test` attribute. Adds back screenshots for failed CI tests to help debug flakes. Updates hack script to deploy demos to wait for demos to be ready before running tests.

Fixes #4955

### DIFF
--- a/frontend/cypress/integration/common/namespace.ts
+++ b/frontend/cypress/integration/common/namespace.ts
@@ -1,0 +1,7 @@
+import { And } from 'cypress-cucumber-preprocessor/steps';
+
+And('user selects the {string} namespace in the NamespaceSelector', (namespace: string) => {
+  cy.getBySel('namespace-dropdown').click();
+  cy.get(`input[type="checkbox"][value="${namespace}"]`).click();
+  cy.getBySel('namespace-dropdown').click();
+});
--- a/frontend/cypress/integration/featureFiles/services.feature
+++ b/frontend/cypress/integration/featureFiles/services.feature
@@ -13,7 +13,7 @@ Feature: Kiali Services page
     Then user sees a table with headings
       | Health | Name | Namespace | Labels | Configuration | Details |
     And the "productpage" row is visible
-    And the health column on the productpage row has a health icon
+    And the health column on the "productpage" row has a health icon
     And the "Name" column on the "productpage" row has a link ending in "/namespaces/bookinfo/services/productpage"
     And the "Namespace" column on the "productpage" row has the text "bookinfo"
     And the "Labels" column on the "productpage" row has the text "app: productpage"
--- a/frontend/cypress/integration/featureFiles/services.feature
+++ b/frontend/cypress/integration/featureFiles/services.feature
@@ -14,13 +14,13 @@ Feature: Kiali Services page
       | Health | Name | Namespace | Labels | Configuration | Details |
     And the "productpage" row is visible
     And the health column on the productpage row has a health icon
-    And the "Name" column on the "productpage" row has a link matching "/console/namespaces/bookinfo/services/productpage"
+    And the "Name" column on the "productpage" row has a link ending in "/namespaces/bookinfo/services/productpage"
     And the "Namespace" column on the "productpage" row has the text "bookinfo"
     And the "Labels" column on the "productpage" row has the text "app: productpage"
     And the "Labels" column on the "productpage" row has the text "service: productpage"
-    And the "Configuration" column on the "productpage" row has a link matching "/console/namespaces/bookinfo/services/productpage"
-    And the "Details" column on the "productpage" row has a link matching "/console/namespaces/bookinfo/istio/virtualservices/bookinfo"
-    And the "Details" column on the "productpage" row has a link matching "/console/namespaces/bookinfo/istio/gateways/bookinfo-gateway"
+    And the "Configuration" column on the "productpage" row has a link ending in "/namespaces/bookinfo/services/productpage"
+    And the "Details" column on the "productpage" row has a link ending in "/namespaces/bookinfo/istio/virtualservices/bookinfo"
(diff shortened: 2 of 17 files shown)


frontend/cypress/integration/common/namespace.ts
```diff
@@ -0,0 +1,7 @@
+import { And } from 'cypress-cucumber-preprocessor/steps';
+
+And('user selects the {string} namespace in the NamespaceSelector', (namespace: string) => {
+  cy.getBySel('namespace dropdown').click();
+  cy.get('input[type="checkbox"][value="bookinfo"]').click();
+  cy.getBySel('namespace dropdown').click();
+});
```
nrfox, line 7: Not totally sure where to put this and the `navigation` step definitions. My thought was to group generic step definitions by function. In this case anything generic related to namespaces could go here. Same for `navigation.ts`.

nrfox:
Marking this as draft until I can get the CI tests passing

frontend/cypress/support/commands.ts
```diff
@@ -24,49 +24,58 @@
...
+            });
+        }
+      } else {
+        cy.log('got an auth cookie, skipping login');
+      }
+    }
+  });
 });
 
+Cypress.Commands.add('getBySel', (selector: string, ...args) => cy.get(`[data-cy="${selector}"]`, ...args));
```
lucasponce, line 81: I'm not sure if this helper would be seen as a "magic" function.
As commented, and it's shown in the code of this PR
image

I'd prefer to see the real selector there, as I think there may be cases where the attribute may change.

frontend/src/components/Filters/StatefulFilters.tsx
```diff
@@ -271,6 +271,7 @@ export class StatefulFilters extends React.Component<StatefulFiltersProps, State
           variant={SelectVariant.typeahead}
           isExpanded={this.state.isExpanded}
           aria-label="filter_select_value"
+          data-cy="filter-value"
```
lucasponce, line 274: I think we are annotating here for the same use case implemented using the "aria-*" in another test:
image

I don't really mind one way or another, but it shouldn't diverge too much IMO.

In this case, the "data-cy" looks clear

lucasponce: Maybe I can add a PR with the changes in the existing tests and try to find a sync point between all folks.
I guess I've added more text in the comments rather that the change in the naming :).

frontend/cypress/support/commands.ts
nrfox: Yes including it here add its to the `cy` interface so you can query/chain it the same way you would other `cy.` commands. It's basically the same thing as:
https://github.com/cypress-io/cypress-realworld-app/blob/develop/cypress/support/commands.ts#L34-L36
https://github.com/openshift/console/blob/master/frontend/packages/integration-tests-cypress/support/selectors.ts#L36-L41
which make it more convenient to query for `cy.get(["data-test-unique-selector"])`. Even if we have additional attributes, it'd still be nice to have some syntactic sugar around the base case of `cy.get(["data-test-${selector}"])`.

frontend/src/components/Filters/StatefulFilters.tsx
nrfox: Sure I'm happy to change one way or the other. I don't have a preference.

lucasponce:
@nrfox please, could you rebase this branch with the master branch?

nrfox:
@lucasponce rebased and CI is passing now so this is ready for review

frontend/cypress/integration/featureFiles/services.feature
```diff
@@ -0,0 +1,63 @@
...
+    And user opens the "services" page
+    And user selects the "bookinfo" namespace in the NamespaceSelector
+    And Kiali is done loading
+
+  @services-page
+  Scenario: See a table with correct info
+    Then user sees a table with headings
+      | Health | Name | Namespace | Labels | Configuration | Details |
+    And the "productpage" row is visible
+    And the health column on the productpage row has a health icon
```
lucasponce, line 16: I think in this line the "productpage" should be passed as parameter instead of hardcoding:

image

frontend/cypress/integration/common/namespace.ts
```diff
@@ -0,0 +1,7 @@
+import { And } from 'cypress-cucumber-preprocessor/steps';
+
+And('user selects the {string} namespace in the NamespaceSelector', (namespace: string) => {
+  cy.getBySel('namespace-dropdown').click();
```
lucasponce, line 4: Not a blocker, but not sure if I'd prefer these "sugar" functions for things that are more elaborated, rather than a simply selector helper.

image

Personally, to me, it helps me when I read the full selector.

But I see interesting these commands for something more elaborated, like the pattern we have to "click-and-wait", those repeated elements perhaps can be implemented under the commands.

Again, not a blocker, we an iterate on this in the future. These tests will have an active maintenance.

lucasponce (commented):
LGTM I think this comment would need a minor change
https://github.com/kiali/kiali/pull/4968#discussion_r858403778

I leave other comments about style, but not blocking, happy to see this suite growing.

Tests work locally, style/refactoring discussions can go later in future iterations.

frontend/cypress/integration/common/namespace.ts
nrfox: That's fair. I was mostly following the pattern from the docs/sample app which include this. I think there's some value in not having to get the `data-test` selector right each time you want to use it but it's a very simple command.

lucasponce (approved):
LGTM thanks!

PR #4971, opened 2022-04-18
E2E: removing the v1.0 in trace test case

pbajjuri20:


pbajjuri20:
it has pulled another commits, I'm closing this PR

PR #4973, opened 2022-04-19
Move the ignored validation message to trace level instead of debug

lucasponce:
Move the ignored validations message to "Trace" level instead of "Debug".
There is a validation disabled by default ("AuthorizationPolicies" on workloads) that populates the log pretty easily while trying to debug other things.

In the Kiali startup, there is an "Info" message that tells the user about this issue, so I think is reasonable to move the level.


leandroberetta (approved):
LGTM.

PR #4977, opened 2022-04-20
Include duration in test url

nrfox:
My working theory for why the tests are flaking in CI is that a re-render happens when the url changes. The URL gets updated with options from the overview toolbar: refresh & duration. We set `refresh` explicitly but we should also set `duration` so kiali doesn't try to update the URL with the default duration. If a re-render happens between the `cy.get('button[aria-labelledby^="overview-type"]')` and the `.click()` then this causes the reference to `cy.get(button...)` to be `undefined`.

### DIFF
--- a/frontend/cypress/integration/common/overview.ts
+++ b/frontend/cypress/integration/common/overview.ts
@@ -23,7 +23,7 @@ Before(() => {
 
 Given('user opens the overview page', () => {
     // Forcing "Pause" to not cause unhandled promises from the browser when cypress is testing
-    cy.visit(url + '/overview?refresh=0');
+    cy.visit(url + '/overview?refresh=0&duration=60');
 });
 
 When('user clicks in the {string} view', (view) => {


nrfox:
You can see the `new_url` event in the cypress browser:
Screenshot from 2022-04-20 12-00-12

It might be worth adding back the videos/screenshots for CI on failure so we can get more context on the CI flakes.

frontend/cypress/integration/common/overview.ts
```diff
@@ -23,7 +23,7 @@ Before(() => {
 
 Given('user opens the overview page', () => {
     // Forcing "Pause" to not cause unhandled promises from the browser when cypress is testing
-    cy.visit(url + '/overview?refresh=0');
+    cy.visit(url + '/overview?refresh=0&duration=60');
```
lucasponce, line 26: I'm not sure if this would trigger then the change in this other test:
image

But let's consolidate this change, as I also expect to update these tests with the changes in the "data-test" attributes discussed in the other WIP.

Issue #4869, opened 2022-03-29
Improve Kiali server release pipeline


leandroberetta:
For epic: #4827 

For historical reasons, the project uses external Jenkins systems to perform some testing and builds.

This issue aims to migrate and improve the kiali server pipeline using GitHub actions.

Issue #4850, opened 2022-03-25
Kiali image builds in Github Actions


nrfox:
Build and push Kiali container images to quay.io in Github Actions. The build/push job should be triggered when a new release is created and the images should be tagged with the version of that release. This is currently happening in Jenkins but it should be moved to Github Actions.

Epic: https://github.com/kiali/kiali/issues/4827

Xunzhuo:
I am happy to help to make this happen, please assign this to me :)
BTW, if we want this, we need to set some github secrets for ci to read:
+ secrets.QUAY_USERNAME 
+ secrets.QUAY_TOKEN

lucasponce:
Hi, @Xunzhuo I think @leandroberetta has started to work on this as part of the https://github.com/kiali/kiali/issues/4827 (thanks @nrfox for creating the subtask.

We have another big epic coming which is to improve the UI testing as we are incorporating the "cypress" framework.

That task would require a nice collaboration and it would be really nice if you'd like to help us with that.

(I'm going to create the Epic, it was on my to-do list for this week).

Xunzhuo:
Hi @lucasponce, thanks for commenting, if @leandroberetta has started, I am also willing to help on improving the UI testing, some details can be discussed in https://github.com/kiali/kiali/issues/4854

leandroberetta:
Hi everyone, as @lucasponce mentioned I started last friday with this one!.

leandroberetta:
This issue will be resolved by https://github.com/kiali/kiali/pull/4980.

Issue #4827, opened 2022-03-22
Improve the upstream pipelines


lucasponce:
In parallel to the efforts of improving the testing scenarios, one epic could be to review our current pipelines based on Github actions.

For historical reasons, the project uses external Jenkins systems to perform some testing and builds.

The goal of this Epic is to study how to move all the upstream pipelines from private systems to Github actions and also how to better organize the current flows to have a better picture of the tasks performed.

Also, there is a parallel goal to include the e2e and future cypress UI testing on the pipelines, so a better organization would help to scale on these scenarios as well.

Github actions allow defining flows that can help to understand the logic, an example like this (from another related project) may help as inspiration:

![image](https://user-images.githubusercontent.com/1662329/159447486-bdc6a77c-8799-41fb-abd5-afb4a3d076ab.png)

Compared with our current flow:

![image](https://user-images.githubusercontent.com/1662329/159447546-b6b7eef3-92f8-4194-8448-c34610f211d1.png)

leandroberetta:
The release pipelines were modified to being triggered on the following Monday after the end of the sprint.

PR #4980, opened 2022-04-21
Release pipeline for Kiali server

leandroberetta:
This is the migrated pipeline for the Kiali server.

Notes:

- I tested this as is me was the kiali bot, but in this PR there are mentions to kiali-bot, but never tried with it, we will need to create a GitHub Actions secret call GH_TOKEN
- I'm not using a fork and the reasons for that are 1) simplicity (not seeing any issue, if there is one, let me know :D) 2) that when CPaaS try to grab the tag, in the past, the commit usually doesn't belong to the repo (I guess it belongs to the fork, and the tag is created before this) and it fails, this was something we discuss with @aljesusg.
- The pipeline for now can be started manually (with workflow_dispatch option, that only works when master, so for that reason I was working on my master's branch in my fork)

Resolves https://github.com/kiali/kiali/issues/4850
Resolves https://github.com/kiali/kiali/issues/4869

Epic: https://github.com/kiali/kiali/issues/4827

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,7 +12,9 @@ on:
         - auto
         - minor
         - snapshot.0
-        - snpashot.1
+        - snapshot.1
+        - snapshot.2
+        - snapshot.3
         - edge
       repository:
         description: Repository (in owner/repo format)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -150,7 +150,7 @@ jobs:
       BRANCH_VERSION: ${{ needs.initialize.outputs.branch_version }}
       NEXT_VERSION: ${{ needs.initialize.outputs.next_version }}
       RELEASE_BRANCH: ${{ github.event.inputs.release_branch }}
-      QUAY_TAG: ${{ github.event.inputs.quay_repository }}:${{ needs.initialize.outputs.release_version }}      
+      QUAY_TAG: ${{ github.event.inputs.quay_repository }}:${{ needs.initialize.outputs.release_version }} ${{ github.event.inputs.quay_repository }}:${{ needs.initialize.outputs.branch_version }}      
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -112,14 +112,15 @@ jobs:
         BRANCH_VERSION: ${{ steps.branch_version.outputs.branch_version }}
       id: quay_tag
       run: |
-        QUAY_TAG=${{ github.event.inputs.quay_repository }}:$RELEASE_VERSION
+        QUAY_TAG="${{ github.event.inputs.quay_repository }}:$RELEASE_VERSION"
         
         if [[ $RELEASE_TYPE == "minor" ]] || [ $RELEASE_TYPE == "patch" ]]
         then
-          QUAY_TAG=$QUAY_TAG ${{ github.event.inputs.quay_repository }}:$BRANCH_VERSION
+          QUAY_TAG="$QUAY_TAG ${{ github.event.inputs.quay_repository }}:$BRANCH_VERSION"
         fi
 
         echo "::set-output name=quay_tag::$QUAY_TAG"
+    
     - name: Log information
       run: |
         echo "Release type: ${{ steps.release_type.outputs.release_type }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -89,7 +89,7 @@ jobs:
       run: |          
         if [[ $RELEASE_TYPE == "patch" ]]
         then
-            NEXT_VERSION=$(./.github/workflows/util/bump $RELEASE_TYPE $RELEASE_VERSION)
+            NEXT_VERSION=$(./.github/workflows/util/bump.py $RELEASE_TYPE $RELEASE_VERSION)
         elif [[ $RELEASE_TYPE == "minor" ]]
         then 
             NEXT_VERSION=$(./.github/workflows/util/bump.py $RELEASE_TYPE $RELEASE_VERSION)          
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,12 +10,13 @@ on:
         type: choice
         options:
         - auto
+        - edge
         - minor
+        - patch
         - snapshot.0
         - snapshot.1
         - snapshot.2
         - snapshot.3
-        - edge
       release_branch:
         description: Branch to release
         required: true
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,6 +38,7 @@ jobs:
       release_version: ${{ steps.release_version.outputs.release_version }}
       branch_version: ${{ steps.branch_version.outputs.branch_version }}
       next_version: ${{ steps.next_version.outputs.next_version }}
+      quay_tag: ${{ steps.quay_tag.outputs.quay_tag }}
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
@@ -104,6 +105,21 @@ jobs:
       if: ${{ steps.release_type.outputs.release_type != 'edge' && !contains(steps.release_type.outputs.release_type, 'snapshot') }}
       run: echo "::set-output name=branch_version::$(echo $RELEASE_VERSION | sed 's/\.[0-9]*\+$//')"
 
+    - name: Determine Quay tag
+      env:
+        RELEASE_TYPE: ${{ github.event.inputs.release_type }}
+        RELEASE_VERSION: ${{ steps.release_version.outputs.release_version }}
+        BRANCH_VERSION: ${{ steps.branch_version.outputs.branch_version }}
+      id: quay_tag
+      run: |
+        QUAY_TAG=${{ github.event.inputs.quay_repository }}:$RELEASE_VERSION
+        
+        if [[ $RELEASE_TYPE == "minor" ]] || [ $RELEASE_TYPE == "patch" ]]
+        then
+          QUAY_TAG=$QUAY_TAG ${{ github.event.inputs.quay_repository }}:$BRANCH_VERSION
+        fi
+
+        echo "::set-output name=quay_tag::$QUAY_TAG"
     - name: Log information
       run: |
         echo "Release type: ${{ steps.release_type.outputs.release_type }}"
@@ -113,6 +129,8 @@ jobs:
         echo "Next version: ${{ steps.next_version.outputs.next_version }}"
 
         echo "Branch version: ${{ steps.branch_version.outputs.branch_version }}"
+
+        echo "Quay tag": ${{ steps.quay_tag.outputs.quay_tag }}
 
   build_backend:
     name: Build backend
@@ -151,7 +169,7 @@ jobs:
       BRANCH_VERSION: ${{ needs.initialize.outputs.branch_version }}
       NEXT_VERSION: ${{ needs.initialize.outputs.next_version }}
       RELEASE_BRANCH: ${{ github.event.inputs.release_branch }}
-      QUAY_TAG: ${{ github.event.inputs.quay_repository }}:${{ needs.initialize.outputs.release_version }} ${{ github.event.inputs.quay_repository }}:${{ needs.initialize.outputs.branch_version }}      
+      QUAY_TAG: ${{ needs.initialize.outputs.quay_tag }}
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -156,9 +156,7 @@ jobs:
       BRANCH_VERSION: ${{ needs.initialize.outputs.branch_version }}
       NEXT_VERSION: ${{ needs.initialize.outputs.next_version }}
       RELEASE_BRANCH: ${{ github.event.inputs.release_branch }}
-      QUAY_TAG: ${{ github.event.inputs.quay_repository }}:${{ needs.initialize.outputs.release_version }}
-      GH_PULL_URI: https://api.github.com/repos/${{ github.event.inputs.repository }}/pulls
-      GH_RELEASE_URI: https://api.github.com/repos/${{ github.event.inputs.repository }}/releases
+      QUAY_TAG: ${{ github.event.inputs.quay_repository }}:${{ needs.initialize.outputs.release_version }}      
     steps:
     - name: Checkout code
       uses: actions/checkout@v3
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,239 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      release_type:
+        description: "Release type"
+        required: true
+        default: "auto"
+        type: choice
+        options:
+          - auto
+          - minor
+          - snapshot.0
+          - snpashot.1
+          - edge
+      repository:
+        description: Repository (in owner/repo format)
+        required: true
+        default: kiali/kiali
+        type: string
+      release_branch:
+        description: Branch to release
+        required: true
+        default: master
+        type: string
+      quay_repository:
+        description: Quay repository
+        type: string
+        default: quay.io/kiali/kiali
+        required: true
+
+jobs:
+  initialize:
+    name: Initialize
+    runs-on: ubuntu-20.04
+    outputs:
+      target_branch: ${{ github.base_ref || github.ref_name }}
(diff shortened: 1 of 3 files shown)


.github/workflows/kiali-release.yml
```diff
@@ -0,0 +1,117 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      target_branch: ${{ steps.target_branch.outputs.target_branch }}
+      release_type: ${{ steps.release_type.outputs.release_type || 'auto' }}      
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      # The initialize job gathers variables for later use in jobs.
+      # We are using this technique rather environment variables because at the moment, they won't work with reusable jobs.
+      # A positive side effect of this is that we can print all variables at start for debugging and troubleshooting.
```
jmazzitelli, line 40: this is trivial but... this comment was confusing at first because its before the "id: target_branch" so I thought it was commenting on that. I think this a comment for the `Initialize` job as a whole - so this comment should go above line 29 so it is more clear what it is describing. Then again, I'm still not sure :) When it says `this technique`, what technique does it mean?

.github/workflows/kiali-release.yml
```diff
@@ -0,0 +1,117 @@
...
+      - id: target_branch
+        run: echo "::set-output name=target_branch::${{ github.base_ref || github.ref_name }}"        
+      - name: Determine release type
+        id: release_type
+        if: ${{ github.event.inputs.release_type == 'auto'}}
+        run: echo "::set-output name=release_type::$(./.github/workflows/util/release_type.sh)"
+  build_backend:
+    name: Build backend
+    uses: ./.github/workflows/build-backend.yml
+    needs: [initialize]
```
jmazzitelli, line 50: this doesn't have `target_branch` unlike the others. Doesn't it need that?

jmazzitelli (changes requested):
my initial observations

.github/workflows/kiali-release.yml
leandroberetta: Thanks, I will move it. 

So, another solution that we tried at first was to have an env block at the jobs level, targetting all jobs, but for the reusable jobs, did not work. 

This workaround/technique is to declare needed variables as outputs of the initialize job, then use it in other blocks as envs or outputs directly.

.github/workflows/kiali-release.yml
leandroberetta: That one is not using it. The frontend does use it to compare the new changes with a yarn command (afaik). And the integration tests use it for the kind setup.

jmazzitelli (approved):

.github/workflows/kiali-release.yml
```diff
@@ -0,0 +1,117 @@
...
+      BACKEND_REPO: ${{ github.event.inputs.backend_repo }}
+      QUAY_NAME: ${{ github.event.inputs.quay_name }}
+    steps:    
+      - name: Checkout code
+        uses: actions/checkout@v2
+      
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17.7
```
israel-hdez, line 84: I wonder if it is possible to set the go version in a single place?
I see the go version spread over multiple workflow files.

Additionally, it is present on the main Makefile: https://github.com/kiali/kiali/blob/7d1c3350c142bd36c13245d7a90078d144faf7e9/Makefile#L31

It would be quite easy to forget to update all places when needed.

.github/workflows/kiali-release.yml
```diff
@@ -0,0 +1,117 @@
...
+            ${{ runner.os }}-go-
+
+      - name: Download frontend build
+        uses: actions/download-artifact@v3
+        with:
+          name: build
+          path: frontend/build       
+
+      - name: Build backend for release (Multi-arch)      
+        run: make -f ./.github/workflows/util/Makefile backend-build-release
```
israel-hdez, line 103: I think you no longer need this step, since you have previously ran the `build-backend.yml` workflow. It is redundant. Perhaps, simply unify the `sed` commands in this make-target into the `backend-push-docker`.

The important point is that you should not build the back-end yet again (just for saving some time). The next step (Push Kiali to Quay) is the one that is doing the multi-arch builds.

.github/workflows/kiali-release.yml
```diff
@@ -0,0 +1,117 @@
...
+
+      - name: Push Kiali to Quay
+        env:
+          QUAY_USER: ${{ secrets.QUAY_USER }}
+          QUAY_PASSWORD: ${{ secrets.QUAY_PASSWORD }}
+        run: make -f ./.github/workflows/util/Makefile backend-push-docker     
+
+      - name: Configure Git
+        run: git config user.email 'lea.beretta@gmail.com' && git config user.name 'leandroberetta'
```
israel-hdez, line 112: Are you sure you want this data in the generated commits?
IMO, you can already put the final email and user name here.

.github/workflows/util/Makefile
```diff
@@ -0,0 +1,197 @@
...
+  BACKEND_BUMPED_VERSION ?= $(shell $(SEMVER) bump minor $(BACKEND_VERSION))
+  BACKEND_VERSION_BRANCH ?= $(shell $(SEMVER) bump release $(BACKEND_VERSION) | sed 's/\.[[:digit:]]\+$$//')
+endif
+
+BACKEND_GITHUB_URI ?= https://github.com/$(BACKEND_REPO).git
+BACKEND_PULL_URI ?= https://api.github.com/repos/$(BACKEND_REPO)/pulls
+BACKEND_RELEASE_URI = $(subst pulls,releases,$(BACKEND_PULL_URI))
+BACKEND_MAIN_BRANCH ?= master
+
+KIALI_BOT_USER ?= leandroberetta
```
israel-hdez, line 62: Also here add a TODO comment.
It's just as a reminder. If you happen to forget it, other eyes can see the TODO comment and fix it for you.

israel-hdez (commented):
I have a few comments. Some are a little picky, but treat everything as simple suggestions (not requirements).

.github/workflows/release.yml
```diff
@@ -0,0 +1,248 @@
...
+        description: 'Release type'     
+        required: true
+        default: 'auto' 
+        type: choice
+        options:
+          - auto
+          - minor
+          - snapshot.0
+          - snpashot.1
+          - edge
```
israel-hdez, line 16: I think this indentation for lists is not what we are currently using in other YAML.
I think @jmazzitelli knows better, since he has a lot of YAML in the operator side.

That said, I'm not sure if indentation here should match the opeator's one.

.github/workflows/release.yml
```diff
@@ -0,0 +1,248 @@
...
+            ${{ runner.os }}-go-
+
+      - name: Download frontend build
+        uses: actions/download-artifact@v3
+        with:
+          name: build
+          path: frontend/build       
+
+      - name: Build image     
+        run: make clean build
```
israel-hdez, line 196: Invoking the `build` target is not needed.
Perhaps, here, simply run `make clean`.

The `Push image` step should be doing the builds.

.github/workflows/release.yml
```diff
@@ -0,0 +1,248 @@
...
+
+      - name: Push image
+        run: |
+          docker login -u ${{ secrets.QUAY_USER }} -p ${{ secrets.QUAY_PASSWORD }} quay.io
+        
+          make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-push-kiali-quay
+
+      - name: Configure git
+        run: |
+          git config user.email 'kiali-dev@googlegroups.com'
```
israel-hdez, line 206: This e-mail is no longer active.
I'm not sure what we could use as a replacement.

.github/workflows/release.yml
```diff
@@ -0,0 +1,248 @@
...
+          git push origin $(git rev-parse HEAD):refs/tags/$RELEASE_VERSION
+        
+      - name: Create release
+        run: |
+          if [[ $RELEASE_TYPE == *"snapshot"* ]]; then export PRERELEASE="true"; else export PRERELEASE="false"; fi
+
+          curl -H "Authorization: token ${{ secrets.GH_TOKEN }}" \
+            -H "Content-Type: application/json" \
+            -d "{\"name\": \"Kiali Operator $RELEASE_VERSION\", \"tag_name\": \"$RELEASE_VERSION\", \"prerelease\": $PRERELEASE}" \
+            -X POST $GH_RELEASE_URI
```
israel-hdez, line 226: I'd suggest to use an action rather than invoking the API with curl.
Here is the official but deprecated action: https://github.com/actions/create-release. It is deprecated, but it lists a few alternate ones that are still maintained.

.github/workflows/release.yml
```diff
@@ -0,0 +1,248 @@
...
+          git add Makefile
+          
+          git commit -m "Prepare for next version"
+          
+          git push origin $(git rev-parse HEAD):refs/heads/$BUILD_TAG
+          
+          curl -H "Authorization: token ${{ secrets.GH_TOKEN }}" \
+            -H "Content-Type: application/json" \
+            -d "{\"title\": \"Prepare for next version\", \"body\": \"Please, merge to update version numbers and prepare for release $NEXT_VERSION.\", \"head\": \"kiali-bot:$BUILD_TAG\", \"base\": \"$RELEASE_BRANCH\"}" \
+            -X POST $GH_PULL_URI
```
israel-hdez, line 248: Same here. I'd suggest to use an already available action, if possible.
Check the Market: https://github.com/marketplace?type=actions&query=create+pull+request+

.github/workflows/release.yml
```diff
@@ -0,0 +1,248 @@
...
+            -X POST $GH_RELEASE_URI        
+
+      - name: Create or update version branch      
+        if: ${{ needs.initialize.outputs.release_type != 'edge' && !contains(needs.initialize.outputs.release_type, 'snapshot') }}
+        run: git push origin $(git rev-parse HEAD):refs/heads/$BRANCH_VERSION
+
+      - name: Create a PR to prepare for next version
+        env:
+          BUILD_TAG: kiali-release-${{ github.run_number }}-main
+        if: ${{ needs.initialize.outputs.release_type == 'minor' }}
```
israel-hdez, line 235: No support for patch releases?

israel-hdez (commented):
> we will need to create a GitHub Actions secret call GH_TOKEN

It is not needed. You can use secrets.GITHUB_TOKEN. See https://docs.github.com/en/actions/security-guides/automatic-token-authentication

.github/workflows/release.yml
jmazzitelli: We typically indent everything in the list at the same level as the parent. 
```yaml
parent:
- starts on same level
- again on same level
```

We don't further indent items in our operator yaml.

I don't know if we care to keep that format the same as this release.yml - but if you are using an editor with indentation rules, it would be nice to have it set with the same format rules and not have to change based on the file.

leandroberetta: I formatted the YAML documents with a Red Hat YAML plugin on VS Code. But if we are using other identation, I can change but I'll wait for @jmazzitelli comments to confirm.

.github/workflows/release.yml
leandroberetta: Right, this needs to be removed.

.github/workflows/release.yml
leandroberetta: What is the mail configured in the kiali-bot account?.

israel-hdez: > @jmazzitelli  perhaps this is a reason to keep it?

I don't know. It was deprecated long ago.

> @leandroberetta  What is the mail configured in the kiali-bot account?.

I don't remember TBH. But I can login into that account and fetch it, if needed.

BTW, do you really need to configure git here? I'd think that GHActions may have something pre-configured that can be used. So, try skipping this step and see what happens.

If it fails, the best I can think right now is to use the no-reply email for that kiali-bot account.

.github/workflows/release.yml
leandroberetta: I can check actions available but I saw they are third party mantained and I thought it would be better to keep the good old curl.

.github/workflows/release.yml
leandroberetta: I'm following the same strategy from the operator pipeline in here https://github.com/kiali/kiali-operator/blob/master/Jenkinsfile#L165

For the patches, it's pushing in the step before (Create or update version branch).

I guess for master is doing the PR because is protected. Unless I'm not seeing something, I guess It's ok this way too.

israel-hdez: OK, it is fine.
I just saw where this is creating the `vX.Y` branch. It is all fine.

.github/workflows/release.yml
israel-hdez: Yes, I think it is mostly 3rd parties.
Well, if you want to stick with something official. I can suggest using the `gh` CLI. See: https://cli.github.com/manual/gh_release_create.

AFAIK, it is available in the runner and invoking `gh` will result in a more readable line than using `curl`.

israel-hdez: Also, I think the command is wrong. I see `\"head\": \"kiali-bot:$BUILD_TAG\"` which means that this would create a PR using a branch in the kiali-bot account, rather than the branch that you just created.

leandroberetta: gh CLI sounds good, I will try it. Thanks!.

leandroberetta: Now that you mention, you're right, but worked for me, because I used "leandroberetta" instead of "kiali-bot", and that is me and my repo. For this PR, needs to be removed. Good catch Edgar!.

.github/workflows/release.yml
leandroberetta: @jmazzitelli yes, I will modifiy the identation, no problem.

.github/workflows/release.yml
leandroberetta: I changed the curl for the gh cli and works fine. I also fix the head field issue.

leandroberetta:
I run this pipeline creating a minor release, the evidence is at:

PR created:
https://github.com/leandroberetta/kiali/pull/28

Branch created:
https://github.com/leandroberetta/kiali/tree/v1.50

Tag created: 
https://github.com/leandroberetta/kiali/tree/v1.50.0

Quay image:
quay.io/leandroberetta/kiali:v1.50.0

.github/workflows/release.yml
leandroberetta: From the evidence I added in the last comment, I'm using the kiali-bot user for the tests, but it seems that is not finding the user, look at:

https://github.com/leandroberetta/kiali/commit/3bc487d70ea9698c7e5aa10b109cb9f7bfdef42b

the commit is made by kiali-bot but not link to it and no avatar, so we might put a different email or keep this and won't be an issue.

(The Release and the PR is created now with a github actions bot as we are using the gh cli)

I checked the behaviour for the last release (v1.49) and its the same:

israel-hdez (approved):
All fine, I think.

leandroberetta:
> oOpS... I reject my approval.
> 
> In your Quay repo, I'm seeing the `v1.50.0` tag, but not the `v1.50` tag.
> 
> image
> 
> Both tags must be created when a minor is released. For example, this is what we have in the "official" repo:
> 
> image
> 
> Also, when doing a `patch` release, the `vX.Y` tag must be updated to point to the new patch release. So, if possible, test that case.

On it, thanks.

leandroberetta:
@israel-hdez the problem is addressed:

Minor:
image

Patch:
image

As you can see, first it was a minor, and then te patch on the v1.50 branch updated the v1.50 image pointing to the image v1.50.1

israel-hdez (approved):
It is now all good!

.github/workflows/release.yml
```diff
@@ -0,0 +1,251 @@
...
+
+    - name: Cache Go deps
+      uses: actions/cache@v3
+      with:
+        path: |
+          ~/.cache/go-build
+          ~/go/pkg/mod
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
```
nrfox, line 196: Late to comment here but why do we need `go` as part of the release pipeline? Since we already built the backend binary in a previous job, can we download that here instead of rebuilding the backend?

israel-hdez: AFAIU, the previous build is just for CI purposes (i.e. testing that it builds and fail fast). But that build is for amd64 arch. This workflow still needs to build the other arches.

nrfox: Ah right we need a separate build for each arch. The integration tests use that build to run the e2e tests against.

PR #4981, opened 2022-04-21
[cypress] login cookie renamed, cy.login() polishing

ScriptingShrimp:
In some recent change, kilai auth cookie was renamed. This PR is fixing `cy.login()` function with new kaili cookie name. There is also more isolation in smoke test itself.

Tested on stage, please test locally.


israel-hdez (approved):
I tried and It worked fine with OpenShift authentication. I got two greens from the `kiali_cookie` tests.
I hope I tried correctly.

ScriptingShrimp:
Thanks @israel-hdez for review. Github actions also looks fixed, so please feel free to merge.

PR #4982, opened 2022-04-22
[cypress] user_logout

pbajjuri20:
Kiali User_logout

### DIFF
--- a/frontend/cypress/integration/common/kiali_logout.ts
+++ b/frontend/cypress/integration/common/kiali_logout.ts
@@ -1,0 +1,22 @@
+import { And, Then } from "cypress-cucumber-preprocessor/steps";
+
+const auth_strategy = Cypress.env('auth_strategy')
+
+And('user clicks on admin', () => {
+    if (auth_strategy === 'openshift') {
+        cy.get('#user-dropdown-toggle').click()
+    }
+})
+
+And('user logout successfully', () => {
+    if (auth_strategy === 'openshift') {
+        cy.intercept('api/logout').as('logout')
+        cy.get('.pf-c-dropdown__menu-item').click()
+    }
+})
+
+Then('user verify the logout', () => {
+    if (auth_strategy === 'openshift') {
+        cy.wait('@logout').its('response.statusCode').should('eq', 204)
+    }
+})
--- a/frontend/cypress/integration/featureFiles/kiali_logout.feature
+++ b/frontend/cypress/integration/featureFiles/kiali_logout.feature
@@ -1,0 +1,14 @@
+Feature: Kiali logout
+
+  User wants to logout to Kiali 
+
+  Background: 
+    Given user is at administrator perspective
+    And user is at the "overivew" page
+
+  @smoke
+  Scenario: Kiali logout successfully
+    
+    And user clicks on admin
+    And user logout successfully
+    Then user verify the logout


frontend/cypress/integration/common/kiali_logout.ts
```diff
@@ -0,0 +1,71 @@
...
+        cy.get('button[type="submit"]').click()
+    }
+})
+
+And('user see console in URL', () => {
+    cy.url().should('include', 'console')
+})
+
+And('user clicks on admin', () => {
+    if (auth_strategy === 'openshift') {
```
ScriptingShrimp, line 54: remove `if (auth_strategy === 'openshift')` condtion to pass on GH actions

frontend/cypress/integration/common/kiali_logout.ts
```diff
@@ -0,0 +1,71 @@
...
+        cy.get('#user-dropdown-toggle').click() 
+    }
+})
+
+And('user logout successfully', () => {
+    cy.intercept('api/logout').as('logout')
+    if (auth_strategy === 'openshift') {
```
ScriptingShrimp, line 61: remove `if (auth_strategy === 'openshift')` condtion to pass on GH actions

ScriptingShrimp (changes requested):
there are still conditions remove `auth_strategy === 'openshift'` on places where they should not be

frontend/cypress/integration/common/kiali_logout.ts
pbajjuri20: removed the auth_strategy

pbajjuri20:
I run this test case multiple times in the local, it run successfully. it is failing in the Git Hub Actions.

pbajjuri20:
@ScriptingShrimp: I have executed in local with upstream cluster and it run perfectly.

It is failing in the "git-hub actions"

frontend/cypress/integration/common/kiali_logout.ts
```diff
@@ -0,0 +1,19 @@
+import { Given, When, And, Then } from "cypress-cucumber-preprocessor/steps";
```
ScriptingShrimp, line 1: `Given, When` imports are not used in this file, so there is no reason to import them

frontend/cypress/integration/featureFiles/kiali_logout.feature
```diff
@@ -0,0 +1,14 @@
...
+
+  User wants to login to Kiali and see landing page
+  
+  @smoke
+  Scenario: Kiali logout successfully
+    Given user opens base url
+    And user clicks Log In With OpenShift
+    And user clicks my_htpasswd_provider
+    And user fill in username and password
+    And user see console in URL
```
ScriptingShrimp, line 11: Sorry I did not noticed earlier: This is obsolete code, should be replaced with 
`Given user is at administrator perspective`

ScriptingShrimp (changes requested):
please add newline in all files so you don't see red sign at the end of each file

frontend/cypress/integration/featureFiles/kiali_logout.feature
pbajjuri20: @ScriptingShrimp : It is taken last commit, I have update only ts file. I will update this file as well. Thanks for the review.

pbajjuri20:
@ScriptingShrimp : I have updated the PR, according to your comments.

lucasponce:
What's the status of this PR?

pbajjuri20:
It is working perfectly in the local. It is not passing the Git-Hub Actions

Pavel requested some changes, I have changed them still it is not working. I will close this ticket ASAP.

lucasponce:
This test is valid :-), I'd prefer if we could just fix it and use a selector that is updated with the changes:
image

Also, if it's platform-dependent, it's ok, as long as it has the proper conditional logic that can help to be run in any platform.

Thanks @pbajjuri20

pbajjuri20:
@lucasponce: Actually, GitHub actions are using the "Anonymous" User. there is no LogOut button.

I'm writing the test case for User Logout. it is unable to find the "logout" button in the UI. 

that is the reason, it is failing :( 


user_1

lucasponce:
> Actually, GitHub actions are using the "Anonymous" User. there is no LogOut button.

> I'm writing the test case for User Logout. it is unable to find the "logout" button in the UI.

It's not a problem; I mean that you could add some conditional logic to the test to execute them under some scenarios (i.e. non anonymous), or similar.

Then the suite would work under any pipeline.

ScriptingShrimp (approved):
LGTM!

PR #4984, opened 2022-04-22
Try to remove traffic generator.

hhovsepy:
For backend integration tests.


hhovsepy:
Hi @nrfox , the graph tests requires a traffic generator. If not able to adjust traffic generator into CI env, then they could be disabled.

nrfox:
@hhovsepy why is the traffic generator being removed?

lucasponce:
Ok, yes, I thought there were some tests that require traffic.
Perhaps we should monitor the rate to see if there is some tweak we can do.
The error-rates workloads also have a "traffic-generator" embedded.

But in any case, if the previous PR with the changes in the test works well, and we don't see more slowness in the integration tests, we can continue with the TG and check if it's the direct cause of the problems.

hhovsepy:
This is do not merge test PR to indicate the amount of failures when traffic generator is removed. sorry for the confusion

PR #4986, opened 2022-04-22
Default Istio.UrlServiceVersion configuration value improvement #4985

### LINKED ISSUE #4985: Default Istio.UrlServiceVersion configuration value improvement
### Problem
Currently, the default value of `UrlServiceVersion` is `http://istiod:15014/version`

https://github.com/kiali/kiali/blob/9bad8de83d2ff207e10c6df1720013a8d5ccf909/config/config.go#L589

This works fine when Kiali runs in the same namespace as Istod.
If this is not the case, `UrlServiceVersion` needs to be set manually

### Suggestion
Set the `UrlServiceVersion` from `http://istiod:15014/version` to `http://istiod.istio-system:15014/version` to also follow the schema of the other URL settings.

mjobst-necls:
** Describe the change **

Use a more general default value for Istio UrlServiceVersion

** Issue reference **

Fixes #4985

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -586,7 +586,7 @@ func NewConfig() (c *Config) {
 				IstiodDeploymentName:              "istiod",
 				IstiodPodMonitoringPort:           15014,
 				RootNamespace:                     "istio-system",
-				UrlServiceVersion:                 "http://istiod:15014/version",
+				UrlServiceVersion:                 "http://istiod.istio-system:15014/version",
 			},
 			Prometheus: PrometheusConfig{
 				Auth: Auth{


lucasponce:
Sorry for the delay, we missed this PR for some reason

lucasponce:
I see there is a discussion on the issue, but the PR looks ok, assuming that the "istiod" would be placed in the "istio-system" as a default value.

This parameter can be updated via operator, but I defer to @jmazzitelli to continue the discussion

jmazzitelli:
I'm -1 on this. Adding to my comments in the github issue, I will say that it appears that all this does is move the issue around. Right now the default works if Kiali is installed in the same namespace as Istio (which may or may not be named `istio-system`). With this PR, it will now work only if Istio is installed in a namespace specifically called `istio-system`, but Kiali is elsewhere in another namespace. This PR seems like it only makes the default "better" for some people that want it, but makes it worse for everyone else (and it is backwards incompatible to boot). I can easily see next week someone proposing a PR to set this default back to the way it was because "I install Kiali and Istio together but in a namespace not called istio-system".

Also, this new default isn't a good one in general anyway, IMO. Because which is the more likely scenario (and thus the best default)? I think it is more likely that Kiali will be installed with Istio where ever it is (as opposed to what this PR wants - supporting Kiali being in a different namespace BUT it also strictly requires Istio being installed in the namespace named `istio-system` and that won't be the case, for example, in a multi-tenant scenario.)

Note this doesn't affect operator-installed Kiali instances. It will affect server-helm-chart installed Kiali instances.

Perhaps someone can better explain why this is needed and why this new default is better than the old default (other than what I explained above). If it is just "I want the default to better match my own environment", we can close this :)

Note that you can easily set this via the helm chart value option `--set external_services.istio.url_service_version=http://istiod.istio-system:15014/version` and it will work as desired for the case where Istio is installed in `istio-system` and Kiali is installed in a different namespace .

israel-hdez:
> (and it is backwards incompatible to boot)

I may think that the right way to go is to change the operator and helm-charts to include the namespace. Probably, something like `spec.istio_namespace || spec.deployment.namespace` would work to be more backwards compatible.

jmazzitelli:
> I may think that the right way to go is to change the operator and helm-charts to include the namespace. Probably, something like `spec.istio_namespace || spec.deployment.namespace` would work to be more backwards compatible.

The operator already does [do this](https://github.com/kiali/kiali-operator/blob/v1.50.0/roles/default/kiali-deploy/tasks/main.yml#L371-L375) (which is why I say this doesn't affect the operator). The server helm chart would need to change to support this, and I did suggest that very thing over in the issue in my comments here:
* https://github.com/kiali/kiali/issues/4985#issuecomment-1106504500
* https://github.com/kiali/kiali/issues/4985#issuecomment-1106509818
* https://github.com/kiali/kiali/issues/4985#issuecomment-1106518512

israel-hdez:
> The operator already does [do this](https://github.com/kiali/kiali-operator/blob/v1.50.0/roles/default/kiali-deploy/tasks/main.yml#L371-L375) (which is why I say this doesn't affect the operator). The server helm chart would need to change to support this, and I did suggest that very thing over in the issue in my comments here:

OK! I overlooked it. I may had scanned it too fast :-)

jmazzitelli:
I'm ok if people want to add this. As i say, it really only affects users of the server helm chart.

leandroberetta:
> > This PR seems like it only makes the default "better" for some people that want it, but makes it worse for everyone else (and it is backwards incompatible to boot).
> 
> This change makes the default consistent with all the rest of the external services: https://github.com/kiali/kiali/blob/master/config/config.go#L556 https://github.com/kiali/kiali/blob/master/config/config.go#L607 https://github.com/kiali/kiali/blob/master/config/config.go#L614
> 
> I think it's reasonable to expect that they'd all have the same format unless there's a special reason why the `istio` service doesn't have the namespace in the URL.

+1 on this. I would say that I'm ok with this change just for the sake of consistency.

Regarding backward compatibility, is there any update required to our docs to reflect this?.

jmazzitelli:
> Regarding backward compatibility, is there any update required to our docs to reflect this?.

no, other than in our release notes. When this goes in, this will go in the release notes when they are generated.

leandroberetta (approved):
LGTM.

jmazzitelli (approved):

israel-hdez:
Thanks @mjobst-necls

PR #4987, opened 2022-04-22
[e2e] Added tests for metrics and logs endpoints.

hhovsepy:

Added tests for Metrics of Namespace, Apps, Workloads and Services.
Added Pod's Logs tests.

### DIFF
--- a/tests/integration/tests/kiali_metrics_test.go
+++ b/tests/integration/tests/kiali_metrics_test.go
@@ -15,6 +15,10 @@ func TestNamespaceMetrics(t *testing.T) {
 
 	assert.Nil(err)
 	assert.NotEmpty(metrics)
+	assert.NotEmpty(metrics.TcpSent)
+	assert.NotEmpty(metrics.TcpSent[0].Datapoints)
+	assert.NotEmpty(metrics.TcpReceived)
+	assert.NotEmpty(metrics.TcpReceived[0].Datapoints)
 }
 
 func TestServiceMetrics(t *testing.T) {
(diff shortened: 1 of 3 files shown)


tests/integration/tests/kiali_metrics_test.go
```diff
@@ -0,0 +1,74 @@
...
+var METRICS_PARAMS = map[string]string{"direction": "outbound", "reporter": "destination"}
+
+func TestNamespaceMetrics(t *testing.T) {
+	assert := assert.New(t)
+
+	pollErr := wait.Poll(time.Second, time.Minute, func() (bool, error) {
+		metrics, err := utils.NamespaceMetrics(utils.BOOKINFO, METRICS_PARAMS)
+		assert.Nil(err)
+		assert.NotEmpty(metrics)
+		return CheckMetrics(metrics.ResponseSize, metrics.RequestSize), nil
```
nrfox, line 23: Do we expect metrics to be empty initially and then to keep polling until they are non-empty? Similar to the recent flake fix, instead of an assertion here, it'd be better to return false if metrics are empty or err if `err != nil`. Otherwise the tests will fail from the assertion in the polling loop if the metrics are ever empty even if they are eventually non-empty.

hhovsepy: Here, even if metrics do not exist, the 'error' and 'metrics' objects should not be nil, only the particular metrics as attributes of 'metrics' object are empty.

nrfox: Ah got it, thanks. I don't think `assert` will fail the test. iirc `require` does that. You might want to return an `err` anyways to exit the polling or else use `require`. Otherwise the `metrics.ReponseSize` and `metrics.RequestSize` on line 23 will likely panic.

hhovsepy: removed those assertions, returning err

nrfox (approved):

PR #4990, opened 2022-04-22
Ignore return codes in wait polls.

hhovsepy:
Patch PR for fixing intermittent non stable backend integration tests.
Ignoring API call return codes in case of wait polls.

### DIFF
--- a/tests/integration/tests/external_host_node_test.go
+++ b/tests/integration/tests/external_host_node_test.go
@@ -28,15 +28,16 @@ func assertExternalNode(params map[string]string, yaml, name string, assert *ass
 	assert.True(utils.ApplyFile(filePath, utils.BOOKINFO))
 
 	pollErr := wait.Poll(time.Second, time.Minute, func() (bool, error) {
-		return NodeMatch(params, name, assert), nil
+		return NodeMatch(params, name), nil
 	})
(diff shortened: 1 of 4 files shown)


tests/integration/tests/external_host_node_test.go
```diff
@@ -28,15 +28,16 @@ func assertExternalNode(params map[string]string, yaml, name string, assert *ass
...
 }
 
-func NodeMatch(params map[string]string, nodeName string, assert *assert.Assertions) bool {
-	graph, statusCode, err := utils.Graph(params)
-	assert.Equal(200, statusCode)
-	assert.Nil(err)
+func NodeMatch(params map[string]string, nodeName string) bool {
+	graph, statusCode, _ := utils.Graph(params)
+	if statusCode != 200 {
+		return false
```
nrfox, line 39: Can we log err if it's non-nil here? It'd be helpful to know what the error is when the tests fail.

nrfox: Yeah that'd be good. If you return `err` inside the `wait.Poll` function then it' will short circuit the polling and return the `err` which is probably what you want if `err != nil`.

nrfox (approved):
Backend tests are passing

PR #4991, opened 2022-04-25
Add a comment for the minikube setup in the bookinfo demo

lucasponce:
Added a note/comment in the install-bookinfo-demo.sh script.
There are some considerations that are platform/mesh dependent and I hope it would save time for others.

I revert the change, as some tests expect the traffic coming from out of the cluster.

### DIFF
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -297,6 +297,11 @@ if [ "${TRAFFIC_GENERATOR_ENABLED}" == "true" ]; then
     INGRESS_ROUTE=$(${CLIENT_EXE} get route istio-ingressgateway -o jsonpath='{.spec.host}{"\n"}' -n ${ISTIO_NAMESPACE})
     echo "Traffic Generator will use the OpenShift ingress route of: ${INGRESS_ROUTE}"
   else
+    # Important note for minikube users
+    # Ingress and Egress configuration depend on the platform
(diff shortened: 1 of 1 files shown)


hhovsepy:
The backend integration tests failures are fixed here https://github.com/kiali/kiali/pull/4990

hack/istio/install-bookinfo-demo.sh
```diff
@@ -297,19 +297,8 @@ if [ "${TRAFFIC_GENERATOR_ENABLED}" == "true" ]; then
     INGRESS_ROUTE=$(${CLIENT_EXE} get route istio-ingressgateway -o jsonpath='{.spec.host}{"\n"}' -n ${ISTIO_NAMESPACE})
     echo "Traffic Generator will use the OpenShift ingress route of: ${INGRESS_ROUTE}"
   else
-    # for now, we only support minikube k8s environments and maybe a good guess otherwise (e.g. for kind clusters)
-    if minikube -p ${MINIKUBE_PROFILE} status > /dev/null 2>&1 ; then
-      INGRESS_HOST=$(minikube -p ${MINIKUBE_PROFILE} ip)
-      INGRESS_PORT=$($CLIENT_EXE -n ${ISTIO_NAMESPACE} get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="http2")].nodePort}')
```
lucasponce, line 303: I know why this didn't work in my "minikube" environment.
If you use the "minikube tunnel" to provide a better emulation of the ingress traffic, then "nodePort/port" seems to not work as the "tunnel" is in place.

In any case, unless there is any specific test, I guess this minor change may help to avoid these problems with the platform.

PR #4995, opened 2022-04-25
Add ScriptingShrimp as new tester

israel-hdez:
See https://github.com/kiali/kiali/discussions/4752.

Voting was already held.
Membership was already granted since some weeks.


jshaughn (approved):
Yep, Pavel should be in the list.  Wish he was pmarek so it would be more clear, but it's all good.

PR #5003, opened 2022-04-26
Remove unused k-charted script

lucasponce:
This script is not used anymore.


Xunzhuo (approved):
LGTM, it looks like this script is no longer used.

Issue #5001, opened 2022-04-26
Add cypress UI tests around the Service Details page


lucasponce:
Subtask for https://github.com/kiali/kiali/issues/4854 Epic

PR #5004, opened 2022-04-26
[cypress] Service Details: Request Routing Wizard tests

lucasponce:
Related to https://github.com/kiali/kiali/issues/5001

I've elaborated a "Wizard" test as they are typically complex with modals, forms, confirmation, etc.

Then it may work as an example for other Wizards and Service Details tests.

I cc @hhovsepy as he also battled a lot with the automation of these UI tests in the past.

### DIFF
--- a/frontend/cypress/integration/featureFiles/wizard_request_routing.feature
+++ b/frontend/cypress/integration/featureFiles/wizard_request_routing.feature
@@ -1,0 +1,55 @@
+Feature: Service Details Wizard: Request Routing
+
+  User navigates to the service details page and open the Kiali Wizard to generate a Request Routing scenario.
+
+  Background:
+    Given user is at administrator perspective
+
+  @wizard-request-routing
+  Scenario: Create a Request Routing scenario
+    When user opens the namespace "bookinfo" and "reviews" service details page
+    And user clicks in the "Request Routing" actions
+    And user sees the "Create Request Routing" wizard
+    And user clicks in the "Request Matching" tab
+    And user clicks in the "headers" request matching dropdown
+    And user types "end-user" in the matching header input
+    And user clicks in the "exact" match value dropdown
+    And user types "jason" in the match value input
+    And user adds a match
+    And user clicks in the "Route To" tab
+    And user types "100" traffic weight in the "reviews-v2" workload
+    And user adds a route
+    And user clicks in the "Request Matching" tab
+    And user clicks in "headers [end-user] exact jason" matching selected
+    And user clicks in the "Route To" tab
+    And user types "100" traffic weight in the "reviews-v3" workload
+    And user adds a route
+    And user previews the configuration
+    And user creates the configuration
+    Then user sees the "Istio Config" table with 2 rows
+
+  @wizard-request-routing
+  Scenario: See a DestinationRule generated
+    When user clicks in the "Istio Config" table "DR" badge "reviews" name row link
+    Then user sees the "kind: DestinationRule" regex in the editor
+    And user sees the "bookinfo" "reviews" "service" reference
+    And user sees the "bookinfo" "reviews-v1" "workload" reference
+    And user sees the "bookinfo" "reviews-v2" "workload" reference
+    And user sees the "bookinfo" "reviews-v3" "workload" reference
+    And user sees the "bookinfo" "reviews" "virtualservice" reference
+
+  @wizard-request-routing
+  Scenario: See a VirtualService generated
+    When user clicks in the "bookinfo" "reviews" "virtualservice" reference
+    Then user sees the "kind: VirtualService" regex in the editor
+    And user sees the "bookinfo" "reviews" "service" reference
+    And user sees the "bookinfo" "reviews" "destinationrule" reference
+    And user sees the "end-user:[\n ]*exact: jason" regex in the editor
+
+  ## @wizard-request-routing
+  Scenario: Delete the Request Routing scenario
+    When user opens the namespace "bookinfo" and "reviews" service details page
+    And user clicks in the "Delete Traffic Routing" actions
+    And user confirms delete the configuration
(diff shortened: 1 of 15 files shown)


frontend/cypress/integration/featureFiles/wizard_request_routing.feature
```diff
@@ -0,0 +1,37 @@
...
+    And user clicks in the "Request Matching" tab
+    And user clicks in "headers [end-user] exact jason" matching selected
+    And user clicks in the "Route To" tab
+    And user types "100" traffic weight in the "reviews-v3" workload
+    And user adds a route
+    And user previews the configuration
+    And user creates the configuration
+    Then user sees a table with headings
+      | Name | Status |
+    And table length should be 2
```
hhovsepy, line 31: is it supposed to verify the content of created DR and VS? 
also to make sure that correct traffic weights are set

lucasponce: +1

Perhaps I can check that in a new scenario (in that file), before deleting the generated Istio Config.

I'll update the test.

lucasponce: I've updated the scenarios, not 100% completed but it expands to navigate to the details and perform some checks in the references and the lines.

P.S.: Parsing the ACE editor is "interesting" :D

lucasponce:
I'm still working in the intermittent errors I see.

frontend/cypress/integration/featureFiles/wizard_request_routing.feature
```diff
@@ -0,0 +1,52 @@
...
+  User navigates to the service details page and open the Kiali Wizard to generate a Request Routing scenario.
+
+  Background:
+    Given user is at administrator perspective
+
+  @wizard-request-routing
+  Scenario: Create a Request Routing scenario
+    When user opens the namespace "bookinfo" and "reviews" service details page
+    And user clicks in the "Request Routing" actions
+    And user sees the "Create Request Routing" wizard
```
hhovsepy, line 12: Should't there be pre-condition to remove existing routing first if exists?

frontend/cypress/integration/featureFiles/wizard_request_routing.feature
lucasponce: I don't think is good to combine REST API calls (or system calls like the kubectl) in the UI test.

If that API is private, I think that the UI test should be solid and support that the private API can change without affecting the test.

But in the same way we have introduced the "data-test" attribute, I think we can add a helper in the UI that would simplify to check and parse the resulting object.

frontend/cypress/integration/featureFiles/wizard_request_routing.feature
```diff
@@ -0,0 +1,52 @@
...
+    Then user sees the "bookinfo" "reviews" "service" reference
+    And user sees the "bookinfo" "reviews" "destinationrule" reference
+    And user sees the "end-user:" and "exact: jason" lines in the editor
+
+  @wizard-request-routing
+  Scenario: Delete the Request Routing scenario
+    When user clicks in the "bookinfo" "reviews" "service" reference
+    And user clicks in the "Delete Traffic Routing" actions
+    And user deletes the configuration
+
```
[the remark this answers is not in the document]
lucasponce: +1 good point, yes, if some element fails, Kiali won't probably delete the config, so it needs an assert there.

frontend/cypress/integration/featureFiles/wizard_request_routing.feature
lucasponce: I assumed that the state was correct at the beginning of the test, and at the end of the test, the state was in the same position.

Otherwise, we'll need to add extra calls to update/delete config that with that assumption is shouldn't be necessary (and I guess it will add more rounds to our pipelines).

frontend/cypress/integration/featureFiles/wizard_request_routing.feature
lucasponce: I've addressed this within the dom, the "text()" invoke in an inner content "div" used by the ACE editor works reasonably well to make assertions (and no need to add extra overhead to that component).

lucasponce:
@hhovsepy I'm checking the flaky tests but I've updated the tests, let me know how it looks.

One pending comment I haven't addressed is to ensure all the config is deleted before the test runs.

I think that can be commented on in a future iteration, I'm assuming that the test cluster is with the state that is populated with the initial setup, and any test should keep that state.

Perhaps when this "corpus" grows we can adjust this.

frontend/cypress/integration/featureFiles/wizard_request_routing.feature
lucasponce: > but one thing still is not clear for me, the step "And user types "100" traffic weight in the "reviews-v2" workload" on "Request Routing", is there wizard logic change?

lucasponce: I mean, this component:
image

The test introduces the weight in the "input text" for simplicity

lucasponce: Thanks for the review Hayk; I'm still working on a flaky test, it seems that I have a race condition.
I'll send some changes to see if it improves the testing in the CI, but It won't change the logic of the testing.

hhovsepy (approved):

lucasponce:
@nrfox @ScriptingShrimp any comment about this PR?

As usual, this is just an example in that area (Wizards) that is typically complex and I wanted to prepare something that may help as an example for other tests.

But I expect these tests will be pretty dynamic as soon as we work on these features.

lucasponce:
Merging to consolidate, for any comment we can add a follow-up work.
Also, it's expected that there will be a high maintenance of these tests when we consolidate the PF4 work

PR #5013, opened 2022-04-27
[cypress] Add scenarios for controlling sidecar injection

israel-hdez:
Add cypress test for controlling sidecar injection.

Related to #4854

### DIFF
--- a/.github/workflows/integration-tests-frontend.yml
+++ b/.github/workflows/integration-tests-frontend.yml
@@ -51,6 +51,9 @@ jobs:
       - name: Install demo apps
         run: hack/istio/install-testing-demos.sh -c "kubectl"
 
+      - name: Install the sleep app in the default namespace
+        run: kubectl apply -n default -f _output/istio-*/samples/sleep/sleep.yaml
+
       - name: Set kiali URL
         run: |
           KIALI_URL="http://$(kubectl get svc -l app.kubernetes.io/name=kiali -n istio-system -o=jsonpath='{.items[0].status.loadBalancer.ingress[0].ip}'):20001/kiali"
--- a/.github/workflows/integration-tests-frontend.yml
+++ b/.github/workflows/integration-tests-frontend.yml
@@ -51,9 +51,6 @@ jobs:
       - name: Install demo apps
         run: hack/istio/install-testing-demos.sh -c "kubectl"
 
-      - name: Install the sleep app in the default namespace
-        run: kubectl apply -n default -f _output/istio-*/samples/sleep/sleep.yaml
-
       - name: Set kiali URL
         run: |
           KIALI_URL="http://$(kubectl get svc -l app.kubernetes.io/name=kiali -n istio-system -o=jsonpath='{.items[0].status.loadBalancer.ingress[0].ip}'):20001/kiali"
--- a/frontend/cypress/integration/common/sidecar_injection.ts
+++ b/frontend/cypress/integration/common/sidecar_injection.ts
@@ -1,0 +1,76 @@
+import { Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+
+Given('a namespace without override configuration for automatic sidecar injection', function () {
+    this.targetNamespace = "default";
+
+    // Make sure that the target namespace does not have override configuration
+    cy.request('PATCH', 'api/namespaces/' + this.targetNamespace, {
+        metadata: {
+            labels: {
+                "istio-injection": null,
+                "istio.io/rev": null
+            }
+        }
+    });
+});
+
+Given('a namespace which has override configuration for automatic sidecar injection', function () {
+    this.targetNamespace = "default";
+    this.istioInjection = "enabled";
+
+    // Make sure that the target namespace has some override configuration
+    cy.request('PATCH', 'api/namespaces/' + this.targetNamespace, {
+        metadata: {
+            labels: {
+                "istio-injection": this.istioInjection,
+                "istio.io/rev": null
+            }
+        }
+    });
+});
+
+Given('the override configuration for sidecar injection is {string}', function (enabledOrDisabled) {
+    if (this.istioInjection !== enabledOrDisabled) {
+        cy.request('PATCH', 'api/namespaces/' + this.targetNamespace, {
+            metadata: {
+                labels: {
+                    "istio-injection": enabledOrDisabled,
+                    "istio.io/rev": null
+                }
+            }
+        });
+        this.istioInjection = enabledOrDisabled;
+    }
+});
+
+When('I override the default automatic sidecar injection policy in the namespace to enabled', function () {
+    cy.visit('overview');
+    cy.get('[data-test=overview-type-LIST]').click();
+    cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).click();
+    cy.get(`[data-test=enable-${this.targetNamespace}-namespace-sidecar-injection]`).click();
+    cy.get('[data-test=confirm-traffic-policies]').click();
+});
+
+When('I change the override configuration for automatic sidecar injection policy in the namespace to {string} it', function (enabledOrDisabled) {
+    cy.visit('overview');
+    cy.get('[data-test=overview-type-LIST]').click();
+    cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).click();
+    cy.get(`[data-test=${enabledOrDisabled}-${this.targetNamespace}-namespace-sidecar-injection]`).click();
+    cy.get('[data-test=confirm-traffic-policies]').click();
+});
+
+When('I remove override configuration for sidecar injection in the namespace', function () {
+    cy.visit('overview');
+    cy.get('[data-test=overview-type-LIST]').click();
+    cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).click();
+    cy.get(`[data-test=remove-${this.targetNamespace}-namespace-sidecar-injection]`).click();
+    cy.get('[data-test=confirm-traffic-policies]').click();
+});
+
+Then('I should see the override annotation for sidecar injection in the namespace as {string}', function (enabled) {
+    cy.get(`[data-test=VirtualItem_${this.targetNamespace}]`).contains(`istio-injection: ${enabled}`).should('exist');
+});
+
+Then('I should see no override annotation for sidecar injection in the namespace', function () {
+    cy.get(`[data-test=VirtualItem_${this.targetNamespace}]`).contains(`istio-injection`).should('not.exist');
+});
--- a/frontend/cypress/integration/common/sidecar_injection.ts
+++ b/frontend/cypress/integration/common/sidecar_injection.ts
@@ -206,7 +206,7 @@ When('I override the default automatic sidecar injection policy in the namespace
     cy.get('[data-test=overview-type-LIST]').click();
     cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).click();
     cy.get(`[data-test=enable-${this.targetNamespace}-namespace-sidecar-injection]`).click();
-    cy.intercept('api/namespaces').as('namespacesReload');
+    cy.intercept(Cypress.config('baseUrl') + '/api/namespaces').as('namespacesReload');
     cy.get('[data-test=confirm-traffic-policies]').click();
     cy.wait('@namespacesReload')
 });
@@ -216,7 +216,7 @@ When('I change the override configuration for automatic sidecar injection policy
     cy.get('[data-test=overview-type-LIST]').click();
     cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).click();
     cy.get(`[data-test=${enabledOrDisabled}-${this.targetNamespace}-namespace-sidecar-injection]`).click();
-    cy.intercept('api/namespaces').as('namespacesReload');
+    cy.intercept(Cypress.config('baseUrl') + '/api/namespaces').as('namespacesReload');
     cy.get('[data-test=confirm-traffic-policies]').click();
     cy.wait('@namespacesReload')
 });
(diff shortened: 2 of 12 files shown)


frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -0,0 +1,76 @@
+import { Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+
+Given('a namespace without override configuration for automatic sidecar injection', function () {
```
lucasponce, line 3: Minor, we are setting the namespaces somehow explicitly.

"A namespace..." could mean anything, and I'd need to look into the implementation to know which namespace it refers.

It could be nice to have "the default namespace...", that could help to read better these tests.

israel-hdez: > "A namespace..." could mean anything

Indeed, and I wrote it that way on purpose. The namespace name is not relevant as long as it fulfills the statements. These `Given` functions are making sure that such statements become fulfilled. So, it is quite possible to switch to another namespace in the implementation and the tests will still pass.

frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -0,0 +1,274 @@
...
+    this.targetNamespace = "default";
+
+    // Make sure that the target namespace does not have override configuration
+    cy.request('PATCH', '/api/namespaces/' + this.targetNamespace, {
```
lucasponce, line 7: I'm not sure if I like this.
You are using the implementation detail when the test should focus on the UI.

frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -0,0 +1,274 @@
...
+                    labels: {
+                        "sidecar.istio.io/inject": null
+                    }
+                }
+            }
+        }
+    });
+
+    // Restart the workload to ensure changes are applied.
+    cy.exec('kubectl delete pod -n default -l app=sleep')
```
lucasponce, line 80: Sorry, I'm -1 on this, I don't think cypress should invoke the system commands.
You can prepare the scenario in the cluster in the external scripts (as we did with the alpha/beta namespaces) and ensure that the test operates with the UI with the state of the cluster.

lucasponce (changes requested):
I know this test wants to be complete but I think the design is mixing the efforts from the e2e and the UI into one.

I suggest simplifying the design and putting the API testing in the new e2e effort that @hhovsepy is working on and focusing in the UI interactions.

Side note that any new installation of a demo app is welcome but should have a wrapper that make sure that is compatible in multiple environments (typically to be sure that downstream pipelines can install it correctly in openshift environments or others).

frontend/cypress/integration/common/sidecar_injection.ts
nrfox: Maybe not here but I think we'll eventually need the tests to execute some system commands. A good example would be the wizards. Without doing a `kubectl get <objecttype>`, there's not an easy way to fetch an object that a wizard created. We could just test the kiali api response object but that doesn't test that the actual kube object was created/edited properly. `kubectl` and/or `oc` are already prerequisites to running the UI tests.

frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -0,0 +1,274 @@
...
+});
+
+Given('a workload with a sidecar', function () {
```
nrfox, line 83: `function` is used here instead of arrow so that state can be shared with `this`?

frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -0,0 +1,274 @@
...
+    this.targetNamespace = 'default';
+    this.targetWorkload = 'sleep';
+
+    // To achieve a workload with sidecar, we turn on injection in its namespace
+    // and also remove any override annotation for sidecar injection.
+    this.namespaceAutoInjectionEnabled = true;
+    this.workloadHasSidecar = true;
+    this.workloadHasAutoInjectionOverride = false;
```
nrfox, line 91: Do these need to be shared state or can they just be const globals for now?

frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -0,0 +1,274 @@
...
+
+Then('the sidecar of the workload should vanish', function () {
+    cy.get('[data-test=missing-sidecar-badge-for-sleep-workload-in-default-namespace]').should('exist');
+});
+
+Then('I should see no override annotation for sidecar injection in the workload', function () {
+    cy.get('#WorkloadDescriptionCard').then($card => {
+        if ($card.find('label_more').length) {
+            cy.wrap($card).get('label_more').click();
+        }
```
nrfox, line 270: There's some warnings about conditional testing. Not sure if that applies here or not: https://docs.cypress.io/guides/core-concepts/conditional-testing

frontend/src/components/Label/Labels.tsx
```diff
@@ -7,6 +7,7 @@ import { KialiIcon } from '../../config/KialiIcon';
 const SHOW_MORE_TRESHOLD = 2;
 
 interface Props {
+  "data-test"?: string;
```
nrfox, line 10: Why does this need a `data-test` prop?

frontend/cypress/integration/common/sidecar_injection.ts
lucasponce: I don't think we should validate things out of the UI unless there is a good reason for that.
The wizards validate the Istio config created from the UI.

There could be always exceptions, but there is a suite of e2e that is designed for this scope.

I have the feeling that we are mixing it here and then it would add more complexity in the future.

israel-hdez: This is a `Given` step.
In the Gherkin reference, it is stated that the purpose of a `Given` is simply "_to put the system the system in a known state_". It is even mentioned that we could add data to a test database (but we don't have a DB, so that does not apply).

IMHO, the docs give the idea that `Given` steps aren't doing testing. So, I understand that it should be OK to use whatever is _at hand_ to achieve the pre-conditions. I know I'm using the API and that is not with the purpose of testing the API, but to achieve the system state as described in the sentence. Indeed, here I'm assuming the API just works (no assertions are done).

I'm also invoking some CLI tools, with that same purpose. Here, I'm using `kubectl` because, sometimes the pod does not restart automatically and, thus, desired state is not reached properly without this restart.

I know these `Given` steps may be different to what we've been writing for the other tests. But, notice that these tests are probably the first ones that are doing changes to the environment as part of testing (i.e. injecting/removing sidecars via changing annotations). If I were to purely use the UI to also attain the pre-conditions, I would end up in a chicken-egg situation, because for some `Scenario`s I would need to use UI features and assume that they "just work" and then test those same UI features in another `Scenario`s (so, if something actually breaks in the UI, tests would break in pairs: one where preconditions fail to be achieved, and the other when actual assertions would fail).

So, I thought this was the best way to write the tests: using whatever is _at hand_ to achieve pre-conditions to then run the desired and relevant UI verifications. Indeed, I invite you to look at the `Then` and `When` steps. Those ones are implemented only with UI manipulations.

.github/workflows/integration-tests-frontend.yml
```diff
@@ -51,6 +51,9 @@ jobs:
       - name: Install demo apps
         run: hack/istio/install-testing-demos.sh -c "kubectl"
 
+      - name: Install the sleep app in the default namespace
+        run: kubectl apply -n default -f _output/istio-*/samples/sleep/sleep.yaml
```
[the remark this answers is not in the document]
israel-hdez: Yes! I can move this to that script.

frontend/cypress/integration/common/sidecar_injection.ts
israel-hdez: Yes :-)

frontend/cypress/integration/common/sidecar_injection.ts
israel-hdez: I was using shared state to prevent a large churn in some future PR is someday these steps need to become parametrized. Thus, you may simply add the parameters and assign to the state and all other steps can remain unchanged.

That said, I'm OK to change this to globals if you think it will be much better.

frontend/cypress/integration/common/sidecar_injection.ts
israel-hdez: I share this concern. Indeed, I stumbled on that guide when I was finding a way to implement this specific test.

May I ask you for your help here? This is the best way I could think of to implement this. I did add a few `cy.intercept` together with a `cy.wait` to try to prevent a flaky test. But I'm not sure if that will be enough. So, suggestions are welcome!

lucasponce: You can use the pod's tooltip to get the list of annotations. That can save to find the "label_more", "click" and wait.

frontend/cypress/integration/common/sidecar_injection.ts
lucasponce: If "Given" statements are used for preparing the state of the test, it sounds reasonable.
But yes, the test should be UI only, unless there are good reasons to find a workaround (i.e. graph area).

I wanted to avoid situations where a test creates some configuration and the test uses the private API to check something instead to navigate to the proper page (because it's easier to invoke that private API than navigate in the UI and build a test).

As soon as the commands are valid in any environment (upstream/downstream), it sounds reasonable.

I'd suggest adding comments in the "Given" area making clear that those API invocations are private, otherwise, we'll have these discussions on every new test, or we may have a variety of styles for test writings and it's the moment to try align them.

frontend/cypress/integration/common/sidecar_injection.ts
nrfox: Looking at this again and what it's trying to test. I think it's fine to click the "show more" labels button so that you can see all the labels before querying for a specific one. You can also get them from the tooltip as Lucas suggested above if that's easier.

frontend/src/components/Label/Labels.tsx
nrfox: I'm not sure I like adding this as a prop to the component. It makes it confusing which component the attribute is being added to. e.g.
```
<Dropdown data-test="special-dropdown" />
```
If `Dropdown` had a `data-test` prop then you wouldn't be adding the attribute to the `Dropdown` component which is different from how the rest of the `data-test` attributes work.

nrfox: Just doing `cy.contains('label')` is probably good enough when you are searching for a specific label in a specific part of the screen.

frontend/cypress/integration/common/sidecar_injection.ts
nrfox: I don't know if sharing state between the tests will make the tests easier or harder to maintain so can't recommend one way or another. It's definitely supported by the plugin. I'd say just leave it as it is for now.

frontend/cypress/integration/common/sidecar_injection.ts
israel-hdez: @lucasponce suggestion sounded quite good to remove the conditional. However, I found that the test becomes flaky, depending on how quickly the pod re-creates. It sometimes happens that there are two pods listed: the one that is terminating and the one that is being created.

frontend/src/components/Label/Labels.tsx
israel-hdez: > If Dropdown had a data-test prop then you wouldn't be adding the attribute to the Dropdown component itself which is different from how the rest of the data-test attributes work.

Probably the Dropdown is not the best example, as you don't know what element in the DOM will get the `data-test` (it fully depends on Patternfly). But I think I get the idea.

> Just doing cy.contains('label') is probably good enough when you are searching for a specific label in a specific part of the screen.

So, I followed your suggestion and removed this `data-test` prop. It turns out that is not enough to do `cy.contains('label')`. Since this is testing for absence, I need to iterate on all visible labels to test for "not containing" the text. However, I think it is making the test a little bit flaky, because sometimes that label takes a little time to go away, and the assertions may already have ran.

I'm not sure how cypress works when doing `.should('not.exist')` on an element. Perhaps, it waits for it to dissapear??? I'm not sure. It could be that this test has been flaky even when using `.should('not.exist')`.

.github/workflows/integration-tests-frontend.yml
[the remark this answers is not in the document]
israel-hdez: Done!

frontend/src/components/Label/Labels.tsx
israel-hdez: Update: Never mind.

I double checked where this `data-test` prop is used, and now I think it was probably a left-over of some iteration while I was still figuring out what to use on `cy.get` to look for elements. So, I restored the original `.should(not.exist)`.

Yet, I still have that doubt about if it is flaky or not... ugh.

lucasponce (approved):
Thanks for the discussion and address the comments.

You've chosen a complex scenario :) as we also know that this specific feature doesn't work 100% in the same way downstream.

But I think it's in a good state and we can consolidate.

There will be changes that would need to be addressed in the downstream environment, but I think it's better to work on them in follow-up work.

lucasponce:
Side comment.
This test deletes the "sleep" pods and it leaves some ReplicaSet orphans (it's not related to Kiali).
image

In orphans cases, Kiali would work but it will log an error in the log that the "parent" is not found, but as side effects, it may populate the log with this extra info.

Not a blocker, but just leaving a comment.

frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -0,0 +1,274 @@
...
+                        labels: {
+                            "sidecar.istio.io/inject": null
+                        }
+                    }
+                }
+            }
+        });
+
+        // Restart the workload to ensure changes are applied.
+        cy.exec('kubectl delete pod -n default -l app=sleep')
```
nrfox, line 168: I'm not sure if `kubectl` will be installed in openshift environments where this test will also run. You might need some conditional to use `oc` based on the `auth` strategy. IMO this would be a good place for a custom command so you can do something like `cy.kubectl('delete pod -n default -l app=sleep')` and inside the custom command either `kubectl` or `oc` would be used depending on the auth strategy.  Not something you need to add with this change but noting here for the future.

frontend/cypress/integration/featureFiles/sidecar_injection.feature
```diff
@@ -0,0 +1,61 @@
...
+    and deployment levels. Annotations are used to override the default policy and
+    Kiali should reflect these annotations.
+
+    Background:
+        Given user is at administrator perspective
+
+    Scenario: Override the default policy for automatic sidecar injection by enabling it in a namespace
+        Given a namespace without override configuration for automatic sidecar injection
+        When I override the default automatic sidecar injection policy in the namespace to enabled
+        Then I should see the override annotation for sidecar injection in the namespace as "enabled"
```
nrfox, line 15: A few of these scenarios are sometimes failing when run locally. I think there's some race condition with clicking the kebab and new data coming in. This gif doesn't show much but you can see the test is stalled until I manually click the kebab:
TestFailure

israel-hdez: Hmmm... that particular step never failed for me.
Mabe the auto refresh is happening, making the kebab menu to go away?

nrfox (approved):

frontend/src/components/MissingSidecar/MissingSidecar.tsx
```diff
@@ -7,6 +7,7 @@ import { KialiIcon } from '../../config/KialiIcon';
 import { style } from 'typestyle';
 
 type MissingSidecarProps = {
+  "data-test"?: string;
```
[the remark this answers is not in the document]
israel-hdez: I'll check if I can fix this in a follow-up PR. And, of course, I'll try to find alternatives for new code.

PR #5016, opened 2022-04-28
Improvements to the release pipeline

leandroberetta:
This PR improves/fixes the following:

- Adds required scripts into the release.yml file and the files in the folder called "util" are removed. This is because the workflow that we are using is on master, this is a valid thing, use the workflow for a branch, and set another branch for building. The problem is that when Github triggers the pipeline using a worflow in a branch, it only gathers the workflows, but no extra files (like bump.py or release_type.sh). So for support patches (branches that are not master and also branches prior to this new pipelines) we needed to embed all the require dependencies in the pipeline.
- Parametrized the reusable jobs to checkout the branch that was selected as "release_branch", this was a bug, it was only pulling master for the release pipeline
- Adding integration tests, before they were commented.


jmazzitelli (approved):
LGTM. Do I understand this right that we will have to manually merge the PR that gets created by the release.yml script? (i.e. it will not auto-merge)?

leandroberetta:
> LGTM. Do I understand this right that we will have to manually merge the PR that gets created by the release.yml script? (i.e. it will not auto-merge)?

Exactly, the release pipelines won't auto merge PR's to master anymore because we are not using a fork and the bot. But I'm investigating a feature of GH Actions that will auto merge.

For now all pipelines will behave like the operator one, manually merge.

PR #5020, opened 2022-04-29
Fix CI branching issue for PRs

leandroberetta:
Fixing CI when run agains PR branches.

### DIFF
--- a/.github/workflows/kiali-ci.yml
+++ b/.github/workflows/kiali-ci.yml
@@ -17,13 +17,22 @@ jobs:
       # to use the branch we are running against: GITHUB_BASE_REF. When run
       # as a release, we want to use the current branch ref: GITHUB_REF_NAME.
       target-branch: ${{ github.base_ref || github.ref_name }}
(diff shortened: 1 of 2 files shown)


lucasponce:
Mmm why the CI are skipped in the PR ?
image

leandroberetta:
> Mmm why the CI are skipped in the PR ? image

So, the last commit was to fix some thing related to the release pipeline (not related to this problem). I'm fixing the value of a variable when the release_type is auto. As it was not related, I cancelled the CI, because for the previous commit it worked:

https://github.com/kiali/kiali/actions/runs/2244737308

We can start the jobs again if you want.

lucasponce:
Ah, I was expecting to see all CIs in green :), that confused me.
Yes, let's update the PR to see that the normal/default flow of the CI works.

(Typically we look when a PR is merged that all CI was ok)

lucasponce (approved):

lucasponce:
I've merged because I'm going to rebase my previous (tiny) PR to check that the CI flow works as expected.

Issue #4359, opened 2021-09-16
Client certificate based kubeconfig usage


libesz:
**Is your feature request related to a problem? Please describe.**
The current Kubernetes API client in Kiali can be only configured to use the bearer token method. I would like to use also kubeconfigs with a client certificate instead of a token.

**Describe the solution you'd like**
During the testing of https://github.com/kiali/kiali/issues/4323, I found out that Kiali does not support certificate based kubeconfigs, it deals only with the token (no matter that remote or normal API access is used). This is usually enough but if the source of the kubeconfig is not based on the projected, auto-mounted, service account token, but coming from external entity, it may have a client certificate in it. This may be an easy change for the Kiali codebase.

**Describe alternatives you've considered**
External automation may be able to install the RBAC resources in the target cluster, including a Kiali SA, using the certificate based kubeconfig, than fetch the SA token from the API and configure it for the Kiali pod.

**Additional context**
In our case, Kiali will not run directly in the mesh cluster. Our systems are providing certificate based kubeconfigs for the mesh cluster. Said that, it would be very convenient if we could just plug that into Kiali.

Such cert based kubeconfigs are [also a standard way](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#x509-client-certs) of authentication/authorization in Kubernetes
 Example:
```
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0t[...]
    server: https://s6662[...]
  name: remote
contexts:
- context:
    cluster: remote
    namespace: default
    user: remote
  name: remote
current-context: remote
kind: Config
preferences: {}
users:
- name: remote
  user:
    client-certificate-data: LS0tL[...]
    client-key-data: LS0tLS1CRUdJTiBS[...]
```

lucasponce:
Thanks @libesz.
I think this issue is valid, we have open discussions about "Deployment Models" that basically it covers from a high level if Kiali is deployed in the same cluster of the mesh or in a remote one.

Also, this issue should be part of the "design/implementation" where it may cover these cases.

But I guess we could discuss this specific topic as well, Kiali is not supported in external cluster at the moment, that configuration is only used for development modes, but I think it could be fair to re-open that discussion.

libesz:
Thanks @lucasponce! Yeah I am open to any detailed discussion.

[...]

PR #5021, opened 2022-04-29
Added authentication for remote Openshift clusters.

libesz:
** Describe the change **

Openshift authentication strategy is perfect when kiali is running inside the mesh cluster. If kiali is running outside the cluster, but the cluster APIs are hosted also outside of the cluster (common in case of managed Openshift clusters in pubic clouds), the same logic is applicable without the assumption of the in-cluster network settings.

This change makes the openshift authentication details configurable, with fallback to the original behavior.

** Issue reference **
https://github.com/kiali/kiali/issues/4359

### DIFF
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -106,7 +106,7 @@ func (in *OpenshiftOAuthService) Metadata() (metadata *OAuthMetadata, err error)
 func getOAuthAuthorizationServer(config config.OpenShiftConfig) (*OAuthAuthorizationServer, error) {
 	var server *OAuthAuthorizationServer
 
-	response, err := request("GET", config.ServerPrefix, ".well-known/oauth-authorization-server", nil, config.UseSystemCA)
+	response, err := request("GET", config.ServerPrefix, ".well-known/oauth-authorization-server", nil, config.UseSystemCA, config.CustomCA)
 
 	if err != nil {
 		log.Error(err)
@@ -129,7 +129,7 @@ func (in *OpenshiftOAuthService) GetUserInfo(token string) (*OAuthUser, error) {
 	var user *OAuthUser
 	config := config.Get().Auth.OpenShift
 
-	response, err := request("GET", config.ServerPrefix, "apis/user.openshift.io/v1/users/~", &token, config.UseSystemCA)
+	response, err := request("GET", config.ServerPrefix, "apis/user.openshift.io/v1/users/~", &token, config.UseSystemCA, config.CustomCA)
 
 	if err != nil {
 		log.Error(err)
@@ -178,7 +178,7 @@ func getKialiRoutePath(config config.OpenShiftConfig) (*string, error) {
 		return nil, fmt.Errorf("could not get openshift config client: %v", err)
 	}
 
-	response, err := request("GET", config.ServerPrefix, fmt.Sprintf("apis/route.openshift.io/v1/namespaces/%v/routes/%s", namespace, config.ClientIdPrefix), &conf.BearerToken, config.UseSystemCA)
+	response, err := request("GET", config.ServerPrefix, fmt.Sprintf("apis/route.openshift.io/v1/namespaces/%v/routes/%s", namespace, config.ClientIdPrefix), &conf.BearerToken, config.UseSystemCA, config.CustomCA)
 	if err != nil {
 		log.Error(err)
 		return nil, fmt.Errorf("could not connect to Openshift: %v", err)
@@ -227,13 +227,13 @@ func (in *OpenshiftOAuthService) Logout(token string) error {
 	log.Debugf("Logging out by deleting OAuth access token [%v] which was converted from access token [%v]", oauthTokenName, token)
 
 	// Delete the access token from the API server using OpenShift 4.6+ access token name
-	_, err = request("DELETE", config.ServerPrefix, fmt.Sprintf("apis/oauth.openshift.io/v1/oauthaccesstokens/%v", oauthTokenName), &conf.BearerToken, config.UseSystemCA)
+	_, err = request("DELETE", config.ServerPrefix, fmt.Sprintf("apis/oauth.openshift.io/v1/oauthaccesstokens/%v", oauthTokenName), &conf.BearerToken, config.UseSystemCA, config.CustomCA)
 
 	if err != nil {
 		// Try to delete the access token from the API server using the pre-4.6 access token name.
 		// If this also fails, we'll send back the err from the first attempt.
 		// If this succeeds, set err to nil to indicate a successful logout.
-		_, err2 := request("DELETE", config.ServerPrefix, fmt.Sprintf("apis/oauth.openshift.io/v1/oauthaccesstokens/%v", token), &conf.BearerToken, config.UseSystemCA)
+		_, err2 := request("DELETE", config.ServerPrefix, fmt.Sprintf("apis/oauth.openshift.io/v1/oauthaccesstokens/%v", token), &conf.BearerToken, config.UseSystemCA, config.CustomCA)
 		if err2 == nil {
 			err = nil
 		}
@@ -246,14 +246,26 @@ func (in *OpenshiftOAuthService) Logout(token string) error {
 	return nil
 }
 
-func request(method string, serverPrefix string, url string, auth *string, useSystemCA bool) ([]byte, error) {
-	return requestWithTimeout(method, serverPrefix, url, auth, time.Duration(defaultRequestTimeout), useSystemCA)
+func request(method string, serverPrefix string, url string, auth *string, useSystemCA bool, customCA string) ([]byte, error) {
+	return requestWithTimeout(method, serverPrefix, url, auth, time.Duration(defaultRequestTimeout), useSystemCA, customCA)
 }
 
-func requestWithTimeout(method string, serverPrefix string, url string, auth *string, timeout time.Duration, useSystemCA bool) ([]byte, error) {
+func requestWithTimeout(method string, serverPrefix string, url string, auth *string, timeout time.Duration, useSystemCA bool, customCA string) ([]byte, error) {
 	var tlsConfig tls.Config
 
-	if !useSystemCA {
+	if customCA != "" {
+		log.Debugf("using custom CA for Openshift OAuth [%v]", customCA)
+		certPool := x509.NewCertPool()
+		decodedCustomCA, err := base64.URLEncoding.DecodeString(customCA)
+		if err != nil {
+			return nil, fmt.Errorf("error decoding custom CA certificates: %s", err)
+		}
+		if !certPool.AppendCertsFromPEM(decodedCustomCA) {
+			return nil, fmt.Errorf("failed to add custom CA certificates: %s", err)
+		}
+		tlsConfig = tls.Config{RootCAs: certPool}
+	} else if !useSystemCA {
+		log.Debugf("Using serviceaccount CA for Openshift OAuth")
 		certPool := x509.NewCertPool()
 		cert, err := ioutil.ReadFile("/run/secrets/kubernetes.io/serviceaccount/ca.crt")
 		if err != nil {
@@ -261,6 +273,8 @@ func requestWithTimeout(method string, serverPrefix string, url string, auth *st
 		}
 		certPool.AppendCertsFromPEM(cert)
 		tlsConfig = tls.Config{RootCAs: certPool}
+	} else {
+		log.Debugf("Using system CA for Openshift OAuth")
 	}
 
 	client := &http.Client{
--- a/business/openshift_oauth.go
+++ b/business/openshift_oauth.go
@@ -57,17 +57,15 @@ type OAuthRouteTLSSpec struct {
 	Termination string `json:"termination"`
 }
 
-const serverPrefix = "https://kubernetes.default.svc/"
-
 const defaultRequestTimeout = 10 * time.Second
 
 var kialiNamespace string
 
 func (in *OpenshiftOAuthService) Metadata() (metadata *OAuthMetadata, err error) {
 	// this is the prefix of the OAuthClient name as well as the Route name
-	clientIdPrefix := config.Get().Auth.OpenShift.ClientIdPrefix
+	config := config.Get().Auth.OpenShift
 
-	redirectURL, err := getKialiRoutePath(clientIdPrefix)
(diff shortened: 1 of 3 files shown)


jmazzitelli:
when adding new config, there are a few more changes required over in kiali-operator - see: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting (for this case, checklist items 1, 3 and 6 in the list need to be done).

libesz:
Thank you @jmazzitelli for the hint!
My thinking was to first align on the concept of this change itself also with @israel-hdez (https://github.com/kiali/kiali/issues/4359#issuecomment-1113138118). One grey area is that how to handle the client ID generation. It is basically the name of the OauthClient object on the authenticating cluster side, but Kiali tries to auto-assemble it based on the namespace where it is deployed (I mean, where kiali is deployed, which is not the right cluster in this case and so it does not make any sense). See [here](https://github.com/kiali/kiali/blob/master/business/openshift_oauth.go#L152). I wonder if I could make this also configurable and intuitive with a single move.
Otherwise if you are happy with the change overall, I will proceed and add the same properties in the operator config.

business/openshift_oauth.go
```diff
@@ -57,25 +57,23 @@ type OAuthRouteTLSSpec struct {
...
-const serverPrefix = "https://kubernetes.default.svc/"
-
 const defaultRequestTimeout = 10 * time.Second
 
 var kialiNamespace string
 
 func (in *OpenshiftOAuthService) Metadata() (metadata *OAuthMetadata, err error) {
 	// this is the prefix of the OAuthClient name as well as the Route name
-	clientIdPrefix := config.Get().Auth.OpenShift.ClientIdPrefix
+	config := config.Get().Auth.OpenShift
```
israel-hdez, line 66: The comment on the previous line is no longer true :-)

business/openshift_oauth.go
```diff
@@ -158,10 +157,15 @@ func getKialiNamespace() (string, error) {
 	return kialiNamespace, nil
 }
 
-func getKialiRoutePath(routeName string) (*string, error) {
+func getKialiRoutePath(config config.OpenShiftConfig) (*string, error) {
 	var route *OAuthRoute
 	var protocol string
 
+	if config.RedirectURL != "" {
+		log.Infof("Using predefined redirect URL: %s", config.RedirectURL)
```
israel-hdez, line 165: Nit: I would log this message with `Debug` or `Trace` level.

business/openshift_oauth.go
```diff
@@ -158,10 +157,15 @@ func getKialiNamespace() (string, error) {
 	return kialiNamespace, nil
 }
 
-func getKialiRoutePath(routeName string) (*string, error) {
+func getKialiRoutePath(config config.OpenShiftConfig) (*string, error) {
 	var route *OAuthRoute
 	var protocol string
 
+	if config.RedirectURL != "" {
```
israel-hdez, line 164: I never examined in detail this `getKialiRoutePath` func until now.

I wonder if this func may work here: https://github.com/kiali/kiali/blob/918476b417c67baadd3a017cfa5c0dc90bf4beaf/util/httputil/http_util.go#L181.

It would be to prevent adding a config option for the redirect URL. If it works, I may think this `getKialiRoutePath` may be completely replaced by that `GuessKialiURL` func.

israel-hdez (commented):
I haven't tried, but I think this is looking quite fine. It removes that assumption that the OAuth server is local.

As you already have faced in the operator PR, the operator auto-configuration will broke. But I see you are already talking with Mazz about how to work through the auto-configuration. In any case, this PR looks quite fine.

What's your plan? Do you want this PR merged? or is it mainly a PoC that may graduate later?

israel-hdez:
> One grey area is that how to handle the client ID generation. It is basically the name of the OauthClient object on the authenticating cluster side, but Kiali tries to auto-assemble it based on the namespace where it is deployed (I mean, where kiali is deployed, which is not the right cluster in this case and so it does not make any sense)

Currently, it is done this way, because that's how the kiali-operator creates the `OauthClient`. But on this remote auth approach, it will depend on how creation of the `OauthClient` will be done. I think there is chance that the operator may no longer be capable of doing the auto-configuration and an external party may need to create the right configs. If this will be the case, then the `client-id` can be anything and that would translate to adding another config in the Kiali `ConfigMap`.

jmazzitelli:
> But on this remote auth approach, it will depend on how creation of the `OauthClient` will be done.

If this involves the OauthClient object, then yes, this will affect the operator. The operator creates this object for you. If this object is no longer good when using a remote cluster, then the operator should not create it, but instead it must be externally created. Or, perhaps there is more config in this auth.openshift section to help? I dunno. Does the operator even need to create this OauthClient object when using this new feature to remote auth?

This is the OauthClient the operator creates: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/openshift/oauth.yaml

business/openshift_oauth.go
libesz: It seems to work that way as well, thanks!

libesz:
So the `OauthClient` need to exist on the mesh cluster (or in the cluster which will handle the auth, but in our case this is the same) for sure. I agree it is generally out of scope of any Kubernetes Operator to operate in more than one cluster. So I agree this ought to be done by external automation before Kiali installation is kicked off.
In this case Kiali config could be extended with some new field, as you suggested.
What about a flag which would disable this auto-configuration? It would:
* disable the addition of the namespace name to the `client_id_prefix`, so the prefix would be used as client ID as is
* disable the creation of the OauthClient always (it is not installed btw in my case now because the cluster type is not openshift, but in the future, on the cluster where Kiali is installed, it may also be also Openshift)

libesz:
> What's your plan? Do you want this PR merged? or is it mainly a PoC that may graduate later?
@israel-hdez  The Kiali integration for us is in PoC phase, I am just learning what needs to be changed to run kiali completely remotely to the mesh cluster. After fixing this auth issue and prometheus access (https://github.com/kiali/kiali/issues/5028), kiali still log errors about istiod access. It seems valid (unfortunately) as in our case istiod is also running outside of the mesh cluster.

Anyways, the auth part in this PR is getting ready feature-wise from my point of view.

israel-hdez:
> In this case Kiali config could be extended with some new field, as you suggested. What about a flag which would disable this auto-configuration? It would:

Yes, an explicit flag is what I personally whould prefer.

> 
>   disable the addition of the namespace name to the `client_id_prefix`, so the prefix would be used as client ID as is
> 

I think I would prefer explicit options for each case. So, I'd prefer adding a new `client_id` for the cases when auto-configuration is turned off, and leave the `client_id_prefix` when it is on. But that would be on the operator side. On the Kiali-server side, I'd prefer to unify into a single `client_id` that the operator would populate accordingly.

>   disable the creation of the OauthClient always (it is not installed btw in my case now because the cluster type is not openshift, but in the future, on the cluster where Kiali is installed, it may also be also Openshift)

This would make it non-backwards compatible, so I am :-1: on disabling always. What we can do is to not forbid `openshift` auth when auto-configuration is turned off. If it is on, current behavior should apply.

That said. I'm wondering... are you really going to use the operator in your env??
If not, and because operators may not be suitable when working with more than one cluster, I'd not worry about changing the operator for now. It may be better to start looking at operator changes once this becomes stable and the final set of configs are in place. Although I think @jmazzitelli should give his thoughts about this.

> After fixing this auth issue and prometheus access (#5028), kiali still log errors about istiod access. It seems valid (unfortunately) as in our case istiod is also running outside of the mesh cluster.

Right, this is part of what I was commenting in the issue: https://github.com/kiali/kiali/issues/4359#issuecomment-1117440953: with the primary-remote setup, the kube-client is ready, but the rest of Kiali "does not know" that it is dealing with more than one cluster.

> Anyways, the auth part in this PR is getting ready feature-wise from my point of view.

If you want, and if this PR won't break anything, personally I'm good with merging to let you progress in small steps.

jmazzitelli:
As I am not familiar at all with these auth issues, I would just say that if the operator doesn't need to change for this to work, then don't change the operator. We can always add new options to the Kiali CR if it makes things easier later. But we also do not want the server helm chart to diverge from the operator (the server helm chart values need to be the same as the operator - we do not want, for example, the server helm chart to support a feature that the operator cannot).

Of course, the one change you do have to make in the operator is being able to set auth.strategy to "openshift" when Kiali is deployed on a non-openshift cluster.

And before we merge this, we need a detailed document explaining what all this is, how you configure it, and once configured, what is happening. Because as it is, we are going to add this config option and the only one who will know what this is even doing let alone how to configure it is @libesz  :) Perhaps a kiali.io PR to the FAQ or Configuration page to describe this new feature and its config settings would be good: [here](https://kiali.io/docs/configuration/authentication/openshift/) or [here](https://kiali.io/docs/faq/authentication/) seems like the best places.

libesz:
>  I would just say that if the operator doesn't need to change for this to work, then don't change the operator

Yep, I am not entirely sure we are going to install Kiali with it's own operator in our setup. In general I am in favor of using the stock operators as it makes our life easier (no need to invent our own lifecycle management tactics, etc. but rely on the already existing implementation), but in case of Kiali I am not sure yet. One thing I noticed is that the operator tend to be a bit slow when Kiali CR is created or changed (compared to Golang based operators). This is not at all blaming, don't get me wrong, I like and use Ansible as well 😄.
We will see how difficult is to deploy the just-enough Kiali resource set with our existing automation. Until that I am fine with only merging this PR.

> And before we merge this, we need a detailed document explaining what all this is

Sure I was planning to extend the auth documentation 👍 . I already have some thoughts in the Kiali CR schema in the operator PR.

lucasponce:
I've read the discussion in the issue and the PR, thanks for the details.

I'd like to comment that there is already a method to run Kiali out of the cluster (in_cluster flag) that is mainly used for development purposes.

Introducing more use cases about this is fine (as discussed, if it doesn't create any incompatible change with existing use cases).

But it would be nice to see all these configurations aligned if possible, to avoid having N different remote configurations.

This effort could be an opportunity to review this and revisit one feature that with minimal changes can help more users.

There is a script that shows the details of this use case https://github.com/kiali/kiali/blob/master/hack/run-kiali.sh

libesz:
@lucasponce thanks for the heads-up, I never checked that script so far. Now I understand the use of the existing `in_cluster` flag 😄 . As I read the script, the way it works would not support well any pod based deployment of kiali due to the tools it uses (overriding default kubernetes env vars, proxying ports, etc). It however is perfectly fine for local development. Would you consider the enhancement of the script or the related kiali code to be part of remote deployment support?

This PR does not give full support of any new deployment model. It does not build on the placement of kiali compared to the mesh at all, it just makes it independent from the oauth server placement, which is often a property of managed k8s clusters.

I think this PR could be in line with what you said: it adds a minimal change to support managed openshift clusters. Further changes may build on this to support more istio deployment models in openshift or mixed environments.

libesz:
> I think I would prefer explicit options for each case. So, I'd prefer adding a new client_id for the cases when auto-configuration is turned off, and leave the client_id_prefix when it is on. But that would be on the operator side. On the Kiali-server side, I'd prefer to unify into a single client_id that the operator would populate accordingly.

@israel-hdez now that the alternative to "do not add this change to the operator right now" came up (https://github.com/kiali/kiali/pull/5021#issuecomment-1118865943), how to handle `client_id` enhancement? Could I just add the new parameter to the server config which would override the prefix handling?

israel-hdez:
> how to handle `client_id` enhancement? Could I just add the new parameter to the server config which would override the prefix handling?

Well, as I commented, if you are _not_ using the operator in your env, then yes! You are free to do this. This is also true if you are using helm charts.

However, if you _are_ using the operator, I think it will make your life easier if you change the operator, because it will override any custom config on each reconciliation (i.e. you will need to re-apply the customization to the server config).

libesz:
@israel-hdez server config updated, assuming the standalone deployment of the kiali for now, without operator. I think this PR is ready from my perspective.

jmazzitelli:
FWIW: we do not want to commit code that diverges support between helm and operator. In fact, the operator should receive priority - the helm chart (especially the server helm chart) is just a secondary alternative installation mechanism.

In other words, we do not want to have a feature supported by the server helm chart that is NOT supported within an operator environment. As soon as we start diverging, we will forget to go back and add that support and then we get into the state of "oh, well, it works with the server helm chart but not the operator" - we don't want that because we really want to recommend/support the operator install mechanism for the general community users.

So, all of this is to say, I have not paid attention to this issue much in the past few days/weeks - so if what I said is not happening (that is, the operator will support whatever it is you are doing here), then you can ignore me :) Otherwise, we need to incorporate support for this into the operator if the helm charts support it.

libesz:
@jmazzitelli For sure I did not change any helm chart, just kiali server code. I am not even familiar with the helm chart based deployment so far, so I might ask for some help to answer your question , sorry.

So what I did is 100% backward compatible. I added options to the kiali configmap/auth/openshift part. I am not sure if this implies the new options will be exposed with helm deployment or not.

If it does, than I guess you suggest to complete https://github.com/kiali/kiali-operator/pull/521, and merge it together with this, right? That is not a problem either, it is already getting ready from my side (except it is missing the new clientID setting).

Sorry if I misunderstood your [comment](https://github.com/kiali/kiali/pull/5021#issuecomment-1118865943) about the need of operator change.

jmazzitelli:
Looking at that operator PR, it doesn't look like the operator needs to do anything with these settings other than pass them through to the ConfigMap. Right? If that is the case then perhaps we can just merge this PR as-is and don't do anything with the helm charts and operator for now. In effect, that would make this a "hidden" feature that is undocumented, but would still work whether installed via helm or operator (because you can put these settings in the Kiali CR and the operator will pass them blindly through to the ConfigMap). For now, maybe that is fine. We can always merge the operator PR later if we want to get it documented and more formally support it.

israel-hdez:
CI passes.
FWIW, I'm on my way to testing your changes.

israel-hdez (approved):
OK, the normal non-remote case seems to be working without any issues.

The remote case... Without he experimental remote feature it doesn't go too far once auth succeeds. With the experimental remote feature enabled, it is better. I think that's OK since this PR focuses only on authentication and it's just a small step towards a bigger goal.

So, all good.

jshaughn:
Thanks @libesz, @israel-hdez and @jmazzitelli !

libesz:
Thank you for the review and all the help!

PR #5032, opened 2022-05-03
Versions must be aligned

aljesusg:
Versions in Makefile and frontend/package.json must be aligned

Screenshot from 2022-05-03 11-38-05

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@kiali/kiali-ui",
-  "version": "1.50.0",
+  "version": "1.51.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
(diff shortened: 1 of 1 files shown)


lucasponce:
Ok, so I think that is missing in the release pipeline to release the frontend/package.json version

lucasponce (approved):
It makes sense.
But also it may need to be synced in the process of bumping/updating the release.
So, @leandroberetta can take a look to update that (probably in a different PR).

leandroberetta:
> It makes sense. But also it may need to be synced in the process of bumping/updating the release. So, @leandroberetta can take a look to update that (probably in a different PR).

Yes, this is missing in the pipeline. I will update it.

leandroberetta (approved):
LGTM

PR #5034, opened 2022-05-04
Set url for grafana and jaeger in CI environment

nrfox:
Sets the url field for grafana and jaeger in CI environment. Needed for #5018


Xunzhuo (approved):
Looks good, thanks @nrfox

PR #5036, opened 2022-05-05
Fix jaeger/grafana url in CI

nrfox:
Fixes url values for jaeger and grafana in CI environment. Adds a dashboard for grafana so that CI tests can use the response. Sets kiali log level to `trace` in CI.


nrfox:
Closing as this was added as a separate commit to #5018

PR #5042, opened 2022-05-06
add cypress-gui and cypress-run targets

jmazzitelli:
This is so it is easy to remember what they go.
cypress-gui runs the GUI so you can pick which individual tests to run.
cypress-run runs the full test suite headless.
Side note: also fixes the help text (make help).. the UI ones were getting indented too much.


nrfox (approved):
:+1: tested locally

jmazzitelli:
will merge, ignoring flaky test that has nothing to do with this PR.

Issue #5030, opened 2022-05-02
Kiali should warn the user when Grafana is not configured properly


nrfox:
Kiali does not work with the grafana installation without setting the `url` field. Other addon services such as jaeger and prometheus work out of the box without setting this field but Grafana does not. 

The kiali logs show:
```
2022-05-02T18:24:00Z DBG [GRAFANA] URL discovery for service 'grafana', namespace 'istio-system'...
2022-05-02T18:24:00Z DBG [GRAFANA] Client is not Openshift, discovery url is only supported in Openshift
```

When the `url` field for Grafana is not set, Kiali should show:
1. error/warning messages in the kiali server 
2. Warning level notification in the UI masthead
3. A descriptive error message saying something like "'url' field is not set. This is required for Kiali to work with Grafana integration"

lucasponce:
The external URL of Grafana is not enabled by default.
Perhaps it's the internal service that has changed? and now is giving an error that the addon is not found?

nrfox:
> The external URL of Grafana is not enabled by default.

The `url` field is empty but the service is enabled by default in both the server: https://github.com/kiali/kiali/blob/master/config/config.go#L555 and the operator: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L146

> Perhaps it's the internal service that has changed? and now is giving an error that the addon is not found?

`in_cluster_url` is correct in Kiali's config I copied above. Setting `url` explicitly fixes the issue but this is not required for the other external services like `jaeger` where the default setup works fine without setting `url`.

I think part of the issue is that the Kiali integration with Grafana creates links using that `url` setting. If those links aren't resolvable by the user then they will be broken. For example, if kiali defaults to using the `in_cluster_url` of `grafana.istio-system` when generating the link, then that probably won't be resolvable when the user clicks on the link. The `url` field is pretty much required since there's not a good way for Kiali to provide a default for it, at least not in non-openshift environments. What was confusing here was that the defaults didn't work. The grafana service was enabled and deployed properly but the only indication that the integration was broken was an info level message in the notification area:

![Screenshot from 2022-05-03 08-28-26](https://user-images.githubusercontent.com/6226732/166453172-7db57abe-16ea-44af-b007-8dc7bfb305cb.png)

[...]

PR #5044, opened 2022-05-09
e2e removal, deleting swagger

hhovsepy:
Removing legacy python e2e tests.
Remove swagger and related things.


lucasponce:
I've started to review this work.

I've found a problem in my environment, like if needs the "oc" client in the PATH

```
{"level":"error","time":"2022-05-11T11:00:06+02:00","message":"exec: \"oc\": executable file not found in $PATH"}
```

I'm reviewing the CI and using Kind, do you still need the "oc" client or it's using the "kubectl" ?

I wonder if we could keep the flag to choose between "oc"/"kubectl" in the integration tests as well, I'd like that those can be executed easily by every developer

lucasponce:
I'm trying to run the tests on minikube following https://github.com/kiali/kiali/tree/master/tests/integration#minikube but I have these issues:

```
=== RUN   TestKialiConfig
    kiali_urls_test.go:29: 
        	Error Trace:	kiali_urls_test.go:29
        	Error:      	Should NOT be empty, but was { }
        	Test:       	TestKialiConfig
    kiali_urls_test.go:30: 
        	Error Trace:	kiali_urls_test.go:30
        	Error:      	Should NOT be empty, but was 
        	Test:       	TestKialiConfig
--- FAIL: TestKialiConfig (0.01s)
=== RUN   TestIstioPermissions
--- PASS: TestIstioPermissions (0.00s)
=== RUN   TestJaeger
    kiali_urls_test.go:51: 
        	Error Trace:	kiali_urls_test.go:51
        	Error:      	Should NOT be empty, but was 
        	Test:       	TestJaeger
--- FAIL: TestJaeger (0.00s)
=== RUN   TestGrafana
    kiali_urls_test.go:70: 
        	Error Trace:	kiali_urls_test.go:70
        	Error:      	Status code should be '200' but was: 503 and error: %!s(<nil>)
        	Test:       	TestGrafana
--- FAIL: TestGrafana (0.00s)
```

I know it's not directly related to this PR, but I think that it's a good moment to check that the integration tests run well in all environments.

I use a clean minikube server with the demos installed using the hack script "install-testing-demos.sh".

I install Kiali in the minikube cluster and open a port-forward to the 20001 port as suggested.

nrfox:
> I've found a problem in my environment, like if needs the "oc" client in the PATH

Yes the integration tests require `oc`. iirc the original python ones being removed here also required `oc` so the two suites were largely kept the same in terms of requirements. The GH actions VMs have `oc` installed by default. That being said I'd be happy if the tests could use either one.

> I use a clean minikube server with the demos installed using the hack script "install-testing-demos.sh".

The integration tests also require you to set the `url` field for both grafana and jaeger. The url doesn't actually need to be resolvable but in order for the tests to see an API response to test the integration, those fields need to be set since there's not a good way to default the url field: https://github.com/kiali/kiali/issues/5030. These requirements should be documented if they are not already.

lucasponce:
> That being said I'd be happy if the tests could use either one.
+1

> The url doesn't actually need to be resolvable but in order for the tests to see an API response to test the integration
I think there are more cases, like the cluster Ids, that are not working well.

But it seems there are just a couple of tests and most of the work should work with one or another environment without big changes of the existing effort (thanks again!).

So I'd prefer that before removing the existing tests we can make this work robust with these couple of changes.

My goal here: is to be sure that any developer must run cypress and integration tests within their local environment and all is clear, so we don't need to chase the author of these changes, and anyone can maintain them.

nrfox:
> My goal here: is to be sure that any developer must run cypress and integration tests within their local environment and all is clear, so we don't need to chase the author of these changes, and anyone can maintain them.

Yes! And being able to use `kubectl` or `oc` depending on which one is available is probably a small change. I'm not sure what we can do about the grafana/jaeger configuration other than document it as a requirement since setting that field will depend on how you've deployed kiali which is different in each environment these tests run in.

lucasponce:
> I'm not sure what we can do about the grafana/jaeger configuration other than document it as a requirement since setting that field will depend on how you've deployed kiali which is different in each environment these tests run in.

No, it's fine, I think a test that checks that the URLs are populated (even if not available because it's a test env) it's perfectly OK.

Perhaps I was more curious about the other URLs about the cluster-mesh-id that seems missing, as probably there is some assumption about the configuration that doesn't work in all the cases.

hhovsepy:
@lucasponce , the support of `kubectl` (or any other `oc` replacement) command can be added in a follow-up tasks.
Here the scope is to remove the legacy e2e tests and swagger json.

lucasponce (approved):
LGTM

I put some comments checking that these tests can be executed without "oc" dependencies.
But as discussed, I'm +1 if those can be added in a follow-up task.

Thanks @hhovsepy

PR #5046, opened 2022-05-09
[cypress] wait for redirect to overview page after logging in

nrfox:
When running tests on openshift, after login the browser is redirected to the overview page. If the tests also navigate to a page, the redirect to the overview page can happen _after_ the tests try to navigate to a particular page causing a test failure on the first test in the test suite.

This change waits for the redirect to the overview page to happen after logging in. Also clears the username/pw fields in the login form in case there is any placeholder text there.

### DIFF
--- a/frontend/cypress/support/commands.ts
+++ b/frontend/cypress/support/commands.ts
@@ -79,6 +79,7 @@ Cypress.Commands.add('login', (provider: string, username: string, password: str
             });
           // Wait for the redirect to the overview page after a successful login.
           // Otherwise the redirect can mess with page loading on subsequent tests.
+          cy.contains('loading', {matchCase: false}).should('not.exist')
           cy.url().should('include', '/overview');
         }
       } else {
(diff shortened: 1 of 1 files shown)


nrfox:
Rebasing off master

frontend/cypress/support/commands.ts
```diff
@@ -68,15 +68,19 @@ Cypress.Commands.add('login', (provider: string, username: string, password: str
...
           cy.get('button[type="submit"]').click();
           cy.wait('@authorized').its('response.statusCode').should('eq', 200);
           cy.getCookie('kiali-token-aes', { timeout: 15000 })
             .should('exist')
             .then(() => {
               haveCookie = true;
             });
+          // Wait for the redirect to the overview page after a successful login.
+          // Otherwise the redirect can mess with page loading on subsequent tests.
+          cy.contains('loading', {matchCase: false}).should('not.exist')
```
ScriptingShrimp, line 82: @nrfox not sure if this `cy.contains` is not too general. But it looks good locally, CI run fixed few testcases.

ScriptingShrimp (approved):
LGTM!

PR #5047, opened 2022-05-09
[cypress] add health indicator tests for "applications"

israel-hdez:
Testing the health indicator for apps in list pages and also the overview page. This is not including detail pages (will add those cases in another PR).


frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -82,7 +82,8 @@ Given('a workload without a sidecar', function () {
     });
 
     // Restart the workload to ensure changes are applied.
-    cy.exec('kubectl delete pod -n default -l app=sleep')
+    cy.exec(`kubectl scale -n ${this.targetNamespace} --replicas=0 deployment/${this.targetWorkload}`);
+    cy.exec(`kubectl scale -n ${this.targetNamespace} --replicas=1 deployment/${this.targetWorkload}`);
```
nrfox, line 86: nit: Doing
```
kubectl rollout restart deployment kiali -n istio-system
kubectl rollout status deployment kiali -n istio-system
```
might be less error prone and will wait for the restart to complete. Not sure if `scale` will wait or not.

frontend/cypress/integration/common/health_indicators.ts
```diff
@@ -0,0 +1,62 @@
+import { Given, Then, When } from "cypress-cucumber-preprocessor/steps";
+
+Given('a healthy application in the cluster', function () {
+    this.targetNamespace = 'bookinfo';
+    this.targetApp = 'productpage';
```
nrfox, line 5: Just pointing out here that there's no guarantee that this app will be healthy when the test runs. It _should_ be but I've seen the various bookinfo apps go "unhealthy" for a bit just because of the nature of the environment. It'd be better to choose a healthy app from the list of applications rather than picking one ahead of time that may or may not be healthy. But I don't know how you choose a healthy app other than looking for the health icon which is what we're trying to test!

nrfox (commented):
We've currently been testing features by page and not by feature. Half of these tests could go in the overview feature and half of these could go in the applications page feature rather than a dedicated feature file for the health indicators. The organization of the step definitions could use some cleanup but I think scoping the feature files to specific pages for now makes it easier to see coverage.

frontend/cypress/integration/common/sidecar_injection.ts
israel-hdez: I was mainly doing scale down and up, because the new health indicator tests are either doing scale down or up and the env is not restored to it's original state. So, this is a way to "restart" and at the same time it is making sure that there is one replica.

That said, I think `scale` is not waiting. So, I'll do a hybrid: scale to one replica and, then restart with `rollout` to make sure it restarts in case `scale` does nothing (if there is already one replica).

frontend/cypress/integration/common/health_indicators.ts
israel-hdez: > It'd be better to choose a healthy app from the list of applications rather than picking one ahead of time that may or may not be healthy. 

Uhm... I don't know. I would prefer this to be deterministic.
Choosing at runtime may make this harder to debug/fix if it breaks, no?

Although, honestly, the env itself is far from deterministic...

frontend/cypress/integration/common/sidecar_injection.ts
israel-hdez: It works with the hybrid with `rollout`.

israel-hdez: I just updated the PR.

israel-hdez:
@nrfox 
> We've currently been testing features by page and not by feature. Half of these tests could go in the overview feature and half of these could go in the applications page feature rather than a dedicated feature file for the health indicators.

OK, I moved the relevant `Scenario`s to the `overview.feature` file. I didn't move the typescript code, though.
I will wait for your Apps PR to become merged so that I move the remainder of the `Scenario`s and, possibly, also the TS code.

> The organization of the step definitions could use some cleanup but I think scoping the feature files to specific pages for now makes it easier to see coverage.

Well, after moving the health indicator scenarios that belong to the overview page, I'm not sure I like it because I miss a proper place for describing the "health indicators" feature. Now, the description of that feature probably will be replicated on all pages/feature.files that have health indicators. Also, I see "The Overview Page" more as a "view" rather than a feature. There is a related blog post [1] in the internet which I mostly agree with.

Personally, I find better to be feature focused because you can see how well covered/tested is a feature that may be distributed on several places, rather than to see how well covered is a "view". IMHO, organizing per "view" I think is a consequence of writing Gherking _after_ develpoment, rather than writing Gherking _before_ development (the later is the recommendation in cucumber docs).

Anyway, the cypress effort already started by organizing per page/view. I won't push nor pursue the other way.

---

[1] https://cucumber.io/blog/bdd/solving-how-to-organise-feature-files/

jshaughn:
> Personally, I find better to be feature focused because you can see how well covered/tested is a feature that may be distributed on several places, rather than to see how well covered is a "view". 

I can see both points of view.  I agree with Edgar that feature-based testing is good as it may guarantee better coverage for that feature, and/or may better emulate what a user may actually do.  I'm not against that idea, especially going forward, when adding a new feature that may cross-cut various views.  But there is also a need to get as much "blanket" coverage as possible, and I think sometimes that is better suited to a page approach, or even a component approach.  As an example, I wrote one for the Display menu of the graph.  It's highly unlikely a user would run through all of those options in one session, but it's by far the easiest way to ensure we have test coverage of the options.

As with most things, I think we need to use our best judgement, there is a lot of gray area.  For this initial push I'd say we're still trying to get coverage.  We should do what makes the most sense given the feature files we already have, and what we're trying to cover.

israel-hdez:
> I'm not against that idea, especially going forward, when adding a new feature that may cross-cut various views. But there is also a need to get as much "blanket" coverage as possible, and I think sometimes that is better suited to a page approach, or even a component approach.

There is something interesting in the blog post I mentioned. From that page, I can quote this:

_I like the idea of being able to quickly pin down system-components in question that generate issues, so a supporting structure that conveys which system-component is to "blame" is very welcome - typically I do this with tags and evaluate the test-reports based on tags…_

What I understand is that, regardless of how the *.feature files are organized, you can also generate reports based on tags, which may be component/view based (or the opposite, or anything you want). That gives the idea that you can have coverage reports of both approaches and fulfill both needs: one based on *.feature files and other based on `@tags`. But, at least, there needs to be some consistency so that reports won't have mixed meanings. However, I don't know how to generate such reports :-)

Just throwing ideas, FWIW.

nrfox:
> IMHO, organizing per "view" I think is a consequence of writing Gherking after develpoment, rather than writing Gherking before development (the later is the recommendation in cucumber docs).

Yes in addition to not having much experience in writing Gherkin. It couldn't be done differently in our case because we had an existing application that we were porting to Gherkin instead of starting from scratch with Gherkin.

> As with most things, I think we need to use our best judgement, there is a lot of gray area. For this initial push I'd say we're still trying to get coverage. We should do what makes the most sense given the feature files we already have, and what we're trying to cover.

This is what I think as well. I'm not saying that dividing up the feature files by "view" or page is the _correct_ approach (they are called **feature** files after all) but it is the approach that all the other tests have followed and if everything is organized the same way, it will make it easier to see what parts of the app have coverage. We need a follow-up issue where we can capture some of these cleanup items and reorganize things in a better way though.

> What I understand is that, regardless of how the *.feature files are organized, you can also generate reports based on tags, which may be component/view based (or the opposite, or anything you want). That gives the idea that you can have coverage reports of both approaches and fulfill both needs: one based on *.feature files and other based on @tags. But, at least, there needs to be some consistency so that reports won't have mixed meanings.

That'd be a good thing to add to the "cleanup" issue but is more complex than what we want to do right now.

jshaughn:
@israel-hdez That's a good point, even if the feature files are organized in a more page/view style, it look like tags could be used to crosscut on feature/workflow or even concern (like perf or smoketest, etc).

nrfox:
@israel-hdez I've merged the apps PR if you'd like to rebase to put the apps health tests there

israel-hdez:
@nrfox Done.

nrfox (approved):

PR #5048, opened 2022-05-10
[molecule] generate a results.xml file using junit xml format

jmazzitelli:
This is useful to tools that want to consume the results of the molecule test run.


jshaughn (approved):
LOK

PR #5049, opened 2022-05-10
[v1.48] [molecule] generate a results.xml file using junit xml format

jmazzitelli:
cherry pick of PR #5048


jmazzitelli:
putting this in the v1.48 branch to allow testing infrastructure to obtain a junit xml formatted file with molecule test results.
This PR only affects a hack script for running molecule tests - it does not affect any code.

Issue #4260, opened 2021-08-05
Upgrade the Patternfly framework


lucasponce:
Subtasks:

In Review and Can be merged if no blockers:

- [x]  https://github.com/kiali/kiali/issues/4836

Issues deferred to ongoing work:
- [x] [overview styling](https://github.com/kiali/kiali/pull/4944#issuecomment-1116084808) (won't do/deferred)
- [x] [shared styling](https://github.com/kiali/kiali/pull/4944#issuecomment-1116134501) (won't do/deferred)
- [x] [preview traffic policy](https://github.com/kiali/kiali/pull/4944#issuecomment-1116085801) (won't do/deferred)
- [x] [chart tooltips](https://github.com/kiali/kiali/pull/4944#discussion_r863887582) (won't do/future)

Issues resolved post-merge:
See https://github.com/kiali/kiali/pull/4944 for a long list of issues identified during review and their status. That PR was used to track progress of all issue resolution prior to merge.   Below is the list of issues to resolve post-merge.
UI:
- [x] [button placement](https://github.com/kiali/kiali/pull/4944#issuecomment-1116085588) https://github.com/kiali/kiali/pull/5124
- [x] [find/hide styling](https://github.com/kiali/kiali/pull/4944#issuecomment-1116104500) https://github.com/kiali/kiali/issues/5125
- [x] [traffic tab columns](https://github.com/kiali/kiali/pull/4944#discussion_r863883977) https://github.com/kiali/kiali/pull/5059
- [x] [kebab placement](https://github.com/kiali/kiali/pull/4944#discussion_r863893326) https://github.com/kiali/kiali/pull/5066
- [x] [heatmap format](https://github.com/kiali/kiali/pull/4944#issuecomment-1116237903) https://github.com/kiali/kiali/pull/5066
- [x] [column headers](https://github.com/kiali/kiali/pull/4944#issuecomment-1116275027) https://github.com/kiali/kiali/pull/5059
- [x] [login screen](https://github.com/kiali/kiali/pull/4944#issuecomment-1117002636) https://github.com/kiali/kiali/pull/5101
- [x] [yaml help icon](https://github.com/kiali/kiali/pull/4944#issuecomment-1117083847) https://github.com/kiali/kiali/issues/5097
- [x] [action menu](https://github.com/kiali/kiali/pull/4944#issuecomment-1117090433) https://github.com/kiali/kiali/pull/5074

Cypress:
- [x] validate cypress feature: graph_display #5063
[...]

PR #5056, opened 2022-05-10
[cypress] Fix tests on services list

jshaughn:
Epic #4260 

[cypress] Fix tests on services list
- Fix double-slash issue in Nav (sent everything to overview)
- Fix bug in label handling using new '=' separator
- Fix services feature for updated PF

### DIFF
--- a/frontend/cypress/integration/common/services.ts
+++ b/frontend/cypress/integration/common/services.ts
@@ -12,8 +12,10 @@ And('the health column on the {string} row has a health icon', (row: string) => 
 });
 
 And('user filters for service type {string}', (serviceType: string) => {
-  cy.get('button[aria-label="Options menu"]').click();
-  cy.contains('button', serviceType).click();
+  cy.get('select[aria-label="filter_select_type"]').parent().within(() => {
+    cy.get('button').click();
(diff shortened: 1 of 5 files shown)


frontend/cypress/integration/common/navigation.ts
```diff
@@ -2,5 +2,5 @@ import { Given } from 'cypress-cucumber-preprocessor/steps';
 
 Given('user is at the {string} page', (page: string) => {
   // Forcing "Pause" to not cause unhandled promises from the browser when cypress is testing
-  cy.visit(Cypress.config('baseUrl') + `/console/${page}?refresh=0`);
+  cy.visit(Cypress.config('baseUrl') + `console/${page}?refresh=0`);
```
nrfox, line 5: Tests fail for me when run locally with the defaults from `make cypress-ui`

jshaughn: Hmm, I see the problem, I started cypress with `make -e CYPRESS_BASE_URL=http://localhost:3000/ cypress-gui`.  My trailing slash caused the problem.  I'll revert the change.

nrfox (approved):

Issue #4424, opened 2021-10-14
validation: authorization policy validation, principals not found


shireenf-ibm:
Add validation to AuthPolicy object. check the existence of peers listed into  **principals** field for each [Rule.From.Source](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Sourcehttps://istio.io/latest/docs/reference/config/security/authorization-policy/#Source)

shireenf-ibm:
> 
> 
> @shireenf-ibm hi! I agree with lucasponce with the registry in place this might be easier.
> 
> however, where did you think of getting the principals? Listing all the service accounts? Is there a way to get the all complete list of possible principles?

Hi, 
The idea is to find Matching services (listing services accounts)

jshaughn:
@lucasponce , The registry work is there, is this something we should revisit?

PR #5057, opened 2022-05-10
Added principals checkers for AuthPolicy

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4424
Istio documentation https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source

Authorization Policy Source has a "principals" field, which contains a list of ServiceAccounts names in "<TRUST_DOMAIN>/ns/<NAMESPACE>/sa/<SERVICE_ACCOUNT>" format.

This PR adds a validation on an existence of those ServiceAccounts, based on searching among Workloads SAs.
Error message is displayed on a principal, which does not refer to existing SA which is labeled to any Workload.

Screenshot from 2022-05-13 09-14-21

### DIFF
--- a/business/checkers/authorization/principals_checker_test.go
+++ b/business/checkers/authorization/principals_checker_test.go
@@ -1,0 +1,49 @@
+package authorization
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	security_v1beta "istio.io/client-go/pkg/apis/security/v1beta1"
+
+	"github.com/kiali/kiali/models"
+	"github.com/kiali/kiali/tests/data"
+	"github.com/kiali/kiali/tests/testutils/validations"
+)
+
+func TestPresentServiceAccount(t *testing.T) {
+	assert := assert.New(t)
+
+	validations, valid := PrincipalsChecker{
+		AuthorizationPolicy: *authPolicyWithPrincipals([]string{"cluster.local/ns/bookinfo/sa/default", "cluster.local/ns/bookinfo/sa/test"}),
+		ServiceAccounts:     []string{"cluster.local/ns/bookinfo/sa/default", "cluster.local/ns/bookinfo/sa/test"},
+	}.Check()
+
+	// Well configured object
+	assert.True(valid)
+	assert.Empty(validations)
+}
+
+func TestNotPresentServiceAccount(t *testing.T) {
+	assert := assert.New(t)
+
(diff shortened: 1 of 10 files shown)


hhovsepy:
Added a help message:
Screenshot from 2022-05-13 09-22-34

models/istio_validation.go
```diff
@@ -117,6 +117,11 @@ var checkDescriptors = map[string]IstioCheck{
 		Message:  "Namespace not found for this rule",
 		Severity: WarningSeverity,
 	},
+	"authorizationpolicy.source.principalnotfound": {
+		Code:     "KIA0106",
```
lucasponce, line 121: Side note, this would require a PR in the kiali.io, to have it documented in the reference.

business/checkers/authorization/principals_checker_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/stretchr/testify/assert"
+	security_v1beta "istio.io/client-go/pkg/apis/security/v1beta1"
+
+	"github.com/kiali/kiali/models"
+	"github.com/kiali/kiali/tests/data"
+	"github.com/kiali/kiali/tests/testutils/validations"
+)
+
+func TestPresentServiceAccount(t *testing.T) {
```
lucasponce, line 14: Is it covered the case when an AuthorizationPolicy doesn't include any principal?

lucasponce:
Nice work:

(The SA doesn't exist and the validation works)
image

(The SA exists and the Istio Config resource has no errors)
image

models/istio_validation.go
hhovsepy: https://github.com/kiali/kiali.io/pull/529

lucasponce:
@hhovsepy I'd like to use this PR as a start to consolidate the "unit of work".
As this PR is adding a new feature (new validation) I think it could be nice if it adds a new e2e/cypress test for that.

Perhaps an e2e suits better for this type of feature.

But this PR can be an example/reference of how we should work when dealing with features, that will help to keep the e2e suite maintained.

hhovsepy:
@lucasponce added an e2e test for the new scenario

lucasponce:
> added an e2e test for the new scenario

Many thanks, @hhovsepy, this PR will help others to see how to add more integration tests

lucasponce (approved):

Issue #4934, opened 2022-04-07
Define a structure for future plugins development


lucasponce:
The goal of this issue is to prepare a first basic structure for future plugins development.

I have prepared some local assets PoC for integration with the https://github.com/openshift/console that could be consolidated into the project to let anyone start taking a look at this effort.

My proposal is to create a main folder like

```
kiali/plugins/openshift/...
```

And under that folder, we'll locate a new web project that will provide the specific assets for it.

A first architecture could be to just add a new listener/port in the Kiali service/pod for this, this listener would be enabled/disabled via configuration.

It will not iterate into more details, and it will be very basic, but it would enable us to see some basic integration and decide how we could organize this.

lucasponce:
This issue is done.
Plugins that depend on third-party platforms will have their own separated repo.
Kiali codebase would be agnostic and that help that any specific platform-related effort would not affect the regular Kiali Lifecycle.

PR #5061, opened 2022-05-11
[plugins] Initial plugin structure

lucasponce:
Draft for https://github.com/kiali/kiali/issues/4934

Goals:
- Propose a structure for new plugins related to the Kiali "ecosystem".
- Define an experimental flag where the Kiali server can serve additional plugins for third-party environments (starting with the OpenShift console).
- Define a way where new plugins can share UI components but not necessarily compile them all together (i.e. a plugin can have static references to the UI code, but the frontend folder should not have any visibility or dependency in any plugin code).

This is in a very experimental stage but it would help to give visibility and share it in the main repo.


hack/run-kiali.sh
```diff
@@ -219,6 +224,16 @@ Valid options:
       generate the distributable package. So, make sure that you build the UI before
       using this script and then set this option to the generated build directory.
       Default: <a local build that is auto-discovered>
+  -pe|--plugin-openshift-enabled
+      If present, it enabled the Kiali OpenShift plugin in the server.
+  -pd|--plugin-openshift-dir
+      A directory on the local machine containing the OpenShift plugin code.
+      If not specified, an attempt to find it on the local machine will be made. A search up the
+      directory tree is made, looking for any directory called "plugins/openshift" that has a "build" directory under it.
```
jmazzitelli, line 232:
```suggestion
      directory tree is made, looking for any directory called "plugins/openshift" that has a "dist" directory under it.
```

hack/run-kiali.sh
```diff
@@ -219,6 +224,16 @@ Valid options:
...
+      The "dist" directory of the UI is generated after you run "yarn build" to
```
jmazzitelli, line 233:
```suggestion
      The "dist" directory of the plugin is generated after you run "make build-plugin-openshift" to
```

hack/run-kiali.sh
```diff
@@ -219,6 +224,16 @@ Valid options:
...
+      generate the distributable package. So, make sure that you build the UI before
```
jmazzitelli, line 234:
```suggestion
      generate the distributable package. So, make sure that you build the plugin before
```

hack/run-kiali.sh
```diff
@@ -219,6 +224,16 @@ Valid options:
...
+      using this script and then set this option to the generated build directory.
```
jmazzitelli, line 235:
```suggestion
      using this script and then set this option to the generated "dist" directory.
```

plugins/openshift/README.md
```diff
@@ -0,0 +1,66 @@
...
+# Install demos used for testing
+$KIALI_HOME/hack/istio/install-testing-demos.sh -c "oc"
+```
+
+In one terminal window, run Kiali in standalone and using the plugins flag:
+
+```
+cd $KIALI_HOME
+
+make build-plugin-openshift
```
jmazzitelli, line 34: This fails for me:

```
$ make build-plugin-openshift 
yarn run v1.22.17
$ yarn clean && NODE_ENV=production yarn ts-node node_modules/.bin/webpack
$ rm -rf dist
$ ts-node -O '{"module":"commonjs"}' node_modules/.bin/webpack
/bin/sh: ts-node: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [make/Makefile.build.mk:52: build-plugin-openshift] Error 127
```

jmazzitelli (commented):
trying to run this but having problems. Here's some comments.

plugins/openshift/README.md
jmazzitelli: I thought maybe I need to do `yarn install` inside the `plugins/openshift` directory, and it appears, yes, that gets me this "ts_node" that is needed. But when I do that, git wants to commit changes that are made (should these go in .gitignore? Should I even do `yarn install` here?)
```
$ pwd
/home/jmazzite/source/kiali/kiali/plugins/openshift

$ git status
On branch lucasponce-plugins-structure
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	../../node_modules/
	../../yarn.lock

```

plugins/openshift/README.md
```diff
@@ -0,0 +1,66 @@
...
+hack/run-kiali.sh --kube-context current --plugin-openshift-enabled
+```
+
+In another window, run:
+
+```
+# Make sure you are logged into OpenShift i.e. oc login
+cd $KIALI_HOME
+
+hack/run-openshift-console.sh
```
jmazzitelli, line 45: At this point.. now what? :)

Based on the output of this command, I think I need to point my browser to http://localhost:9000 -- correct?

nrfox:
Thanks for putting this together! Just a high level comment. Based on the description on what is provided and briefly scanning the diff, this doesn't seem like a Kiali "plugin" but rather a separate app that imports and uses the Kiali components. To me, a plugin is something that "extends" or "enhances" the original thing but this more "replaces" it by using components from Kiali to build a new app. Kiali is basically the main dependency here for a new app. Instead of serving this new app alongside the Kiali assets, perhaps what assets that Kiali serves can be made configurable. You could imagine any new app being able to be bundled with and served by the Kiali server that way without needing to be a special plugin with its own predefined port within the Kiali server.

lucasponce:
>  To me, a plugin is something that "extends" or "enhances" the original thing but this more "replaces" it by using components from Kiali to build a new app.

lucasponce:
Thanks, @jmazzitelli, I've updated the script with the comments.

nrfox:
> In the past we used the term "extension" [1] when some third party would like to add new capabilities to Kiali, we could use that for future use cases; but for this particular effort, if we mix the naming perhaps we can introduce more wording everywhere.

Yes it would be confusing to call it an "extension" when it's different from the previous extensions. It's also confusing to have both extensions and plugins for Kiali. What is the difference between the two?

What if we just flipped the dir structure from `kiali/plugins/openshift` --> `kiali/openshift/plugins`? Then instead of a Kiali plugin, this is an openshift plugin of Kiali. I think the Kiali server can remain largely unchanged as long as we can make the assets the Kiali server serves configurable. The Kiali server would then serve the openshift console plugin instead of the normal UI assets.

lucasponce:
> What if we just flipped the dir structure from kiali/plugins/openshift --> kiali/openshift/plugins? Then instead of a Kiali plugin, this is an openshift plugin of Kiali. I think the Kiali server can remain largely unchanged as long as we can make the assets the Kiali server serves configurable. 

I know what you mean and I share the reasoning, I think that kiali/plugins/... would be better, because the "kiosk" mode is also used by other users that may like to also explore a webpack plugin. Then, i.e. "rancher" would like to add a webpack code for their needs, the "plugins/rancher" structure suits better.

Putting "openshift" at the main level would need some justification, so it would need a generic folder to invite any other contributor to work on something similar. I think.

> The Kiali server would then serve the openshift console plugin instead of the normal UI assets.

I think that would be something in the future, but probably coexisting both standalone UI + webpack plugins is the most probably use-case in the short/mid steps.

nrfox:
> I know what you mean and I share the reasoning, I think that kiali/plugins/... would be better, because the "kiosk" mode is also used by other users that may like to also explore a webpack plugin. Then, i.e. "rancher" would like to add a webpack code for their needs, the "plugins/rancher" structure suits better.

Allowing the static assets served by Kiali to be configurable I think would make it easier for other folks to re-use the Kiali server. There wouldn't be any upstream changes necessary and whoever is integrating with Kiali wouldn't need to get a separate port open and flags added to the server for their plugin in addition to the potential helm chart/operator changes necessary to support the multiple ports and plugins. It should be as simple as creating a separate container image of Kiali with the custom assets you want to be served. Kiali might even support this today.

Not a blocker just putting my 2c out there. I do think it'd make it easier for other folks to re-use Kiali without needing to merge and maintain upstream changes as well as giving greater flexibility in cases where extension doesn't need a webpack plugin. I'd imagine you could offer your own re-skinned version of Kiali by just adding some custom css and building your Kiali container image with that.

Perhaps it's worth investigating to see if this is possible today without any changes. I haven't built the plugin yet and tried to get things running so it may not be feasible for some reason.

lucasponce:
> > I know what you mean and I share the reasoning, I think that kiali/plugins/... would be better, because the "kiosk" mode is also used by other users that may like to also explore a webpack plugin. Then, i.e. "rancher" would like to add a webpack code for their needs, the "plugins/rancher" structure suits better.
> 
> Allowing the static assets served by Kiali to be configurable I think would make it easier for other folks to re-use the Kiali server. There wouldn't be any upstream changes necessary and whoever is integrating with Kiali wouldn't need to get a separate port open and flags added to the server for their plugin in addition to the potential helm chart/operator changes necessary to support the multiple ports and plugins. It should be as simple as creating a separate container image of Kiali with the custom assets you want to be served. Kiali might even support this today.
> 
> Not a blocker just putting my 2c out there. I do think it'd make it easier for other folks to re-use Kiali without needing to merge and maintain upstream changes as well as giving greater flexibility in cases where extension doesn't need a webpack plugin. I'd imagine you could offer your own re-skinned version of Kiali by just adding some custom css and building your Kiali container image with that.
> 
> Perhaps it's worth investigating to see if this is possible today without any changes. I haven't built the plugin yet and tried to get things running so it may not be feasible for some reason.

I'm not sure if I follow you.

I think that we need to keep the port that is serving the standalone + a new port that would serve a new webpack, at least in the situations that we may need soon.

I think your proposal is to keep only one port and configure to server as a standalone vs webpack, but I think we need both at the same time.

nrfox:
> I think your proposal is to keep only one port and configure to server as a standalone vs webpack, but I think we need both at the same time.

Yes I was originally thinking that the "plugin" html/css/js assets would replace the "Kiali server" assets but it seems as though the plugin assets still rely on the `index.html` file and possibly some others. Still you can bundle the plugin assets alongside the Kiali assets and it works fine. 

Here's a prototype: https://github.com/kiali/kiali/compare/master...nrfox:plugin-bundle?expand=1. Ran through the "Run Kiali and OpenShift Console in a local environment" setup and everything works fine.

lucasponce:
> Here's a prototype: https://github.com/kiali/kiali/compare/master...nrfox:plugin-bundle?expand=1. Ran through the "Run Kiali and OpenShift Console in a local environment" setup and everything works fine.

This looks good, easier than opening another port.

Yes, I think that one direction or another is more related to the topics about the lifecycle/cadence/etc that are less related to the technical solution, but your works for me.

lucasponce:
I'm thinking to start moving this PoC into a new repository that may hold the plugin artifact + the plugin operator.
How does it sound?

It would need to re-arrange this work, but I think that will help to iterate better according to the feedback collected.

nrfox:
> I'm thinking to start moving this PoC into a new repository that may hold the plugin artifact + the plugin operator. How does it sound?

Sounds good to me!

> It would need to re-arrange this work, but I think that will help to iterate better according to the feedback collected.

+1

jmazzitelli:
> I'm thinking to start moving this PoC into a new repository that may hold the plugin artifact + the plugin operator.
> How does it sound?

I think that sounds good. I like the idea of having the plugin artifact separate. We should bundle it with a very small lightweight web server whose only job is to serve up the plugin. In other words, don't bundle it with the Kiali Server itself. When the operator installs this small lightweight plugin server, it can configure the plugin to talk to the Kiali Server. And in fact this allows us to configure the plugin to talk to more than one Kiali server if that makes sense. For example, that could allow the plugin to directly support multi-tenant installations (e.g. this could be useful when looking at resources in a particular namespace and that namespace is part of a specific tenant).

lucasponce:
+1 to all the comments.
I'm going to create a new repo called "openshift-servicemesh-plugin" that will hold the plugin + artifact and I will move this PR logic there.

Once that is ready and we have an artifact that can be "demo-ed", I will close this PR, but I will leave it open for anyone that want to recreate the dev environment.

lucasponce:
Created https://github.com/kiali/openshift-servicemesh-plugin/pull/1 to move this PoC into its own repository where plugin + operator could be consolidated.

I put @jmazzitelli main reviewer to check if something is broken, but anyone is invited to test and comment.

I'll close this PR when that work is consolidated in the new repo.

PR #5063, opened 2022-05-11
fix cypress graph display  feature in CI

jshaughn:
Epic #4260 

- add "--profile" to the build. This option adds a little bit of overhead but has the advantage of not munging names that can be important for understanding profiler information, but in this case, also for searching on react component names in tests.

also:
- move #root into config, as suggested in crs doco
- turn off video when running headless via make
- update one test that was failing due to the slightly different demo app setup in CI, as compared to what I had locally.

### DIFF
--- a/frontend/cypress/integration/common/graph_display.ts
+++ b/frontend/cypress/integration/common/graph_display.ts
@@ -124,8 +124,9 @@ Then('the display menu has default settings', () => {
 });
 
 Then('the graph reflects default settings', () => {
-  cy.waitForReact(1000, '#root');
+  cy.waitForReact();
   cy.getReact('CytoscapeGraph')
+    .should('have.length', '1')
     .getCurrentState()
     .then(state => {
       // no nonDefault edge label info
@@ -148,7 +149,7 @@ Then('the graph reflects default settings', () => {
 
       // a variety of not-found tests
       numNodes = state.cy.nodes(
-        `[isBox = "cluster"],[?isIdle],[?rank],[?hasMissingSC],[?hasVS],[nodeType = "operation"]`
+        `[isBox = "cluster"],[?isIdle],[?rank],[nodeType = "operation"]`
       ).length;
       assert.isTrue(numNodes === 0);
     });
@@ -169,8 +170,9 @@ Then('user sees {string} edge labels', el => {
       rate = el;
   }
 
-  cy.waitForReact(1000, '#root');
+  cy.waitForReact();
   cy.getReact('CytoscapeGraph')
+    .should('have.length', '1')
     .getCurrentState()
     .then(state => {
       const numEdges = state.cy.edges(`[${rate}" > 0]`).length;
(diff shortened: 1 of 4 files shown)


frontend/cypress/integration/common/graph_display.ts
```diff
@@ -257,8 +264,9 @@ Then('user does not see service nodes', () => {
 Then('security {string} in the graph', action => {
   validateInput('filterSecurity', action);
 
-  cy.waitForReact(1000, '#root');
+  cy.waitForReact();
```
nrfox, line 267: Is the timeout param no longer needed?

frontend/package.json
```diff
@@ -37,7 +37,7 @@
     "build": "if [ \"${KIALI_ENV}\" = \"production\" ]; then npm run build:prod; else npm run build:dev; fi",
     "build-css": "node-sass src/ --output-style compressed --include-path $npm_package_sassIncludes_src -o src/",
     "build:dev": "sh -ac '. ./.env.upstream; npm run lint && npm run build:kiali'",
-    "build:kiali": "npm run build-css && REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) GENERATE_SOURCEMAP=false EXTEND_ESLINT=true react-scripts build",
+    "build:kiali": "npm run build-css && REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) GENERATE_SOURCEMAP=false EXTEND_ESLINT=true react-scripts build --profile",
```
nrfox, line 40: Is there any way to add this to `build:dev` without adding to `build:prod`? Not if the CI build is done with `build:dev` or `build:prod`.

nrfox (approved):
One suggestion but it can probably be a optimization follow-up. Good find with the `--profile` flag.

The 1 failure in CI is unrelated to these changes and looks like a flake.

frontend/cypress/integration/common/graph_display.ts
jshaughn: it was only needed because the second param was being provided.  Now the root definition has moved to the config.  The default timeout is 10s, I think it's fine but if you want me to lower it, I can.

frontend/package.json
jshaughn: the CI is a prod build.  I think if it were a dev build this may not have been an issue, but I'm not sure we want a dev build.  If we are not comfortable adding this option to a prod build, then I'd probably suggest we add a CI build option.  From what I read, there is not a lot of additional overhead.

frontend/cypress/integration/common/graph_display.ts
jshaughn: > The 1 failure in CI is unrelated to these changes and looks like a flake

Yes, that test succceded on some runs.  It is not failing due to these changes.

frontend/package.json
nrfox: A separate CI build opt sounds fine. It can either be included in this PR or in a followup since the graph tests are causing the rest of the CI runs to fail at the moment.

jshaughn: let's do it in a follow up (and maybe someone else that isn't as much of a dork at that stuff as me) :-)  Would like to get this in.

jshaughn: so, if you are ok, feel free to approve.

jshaughn: oh, it's already approved  :-)  I'll merge, thanks.  And I'll create the follow-up issue.

jshaughn:
@nrfox see https://github.com/kiali/kiali/issues/5067 for CI build issue.

Issue #4912, opened 2022-04-06
Add cypress UI tests around the Workload List page


aljesusg:
Subtask for https://github.com/kiali/kiali/issues/4854 Epic

PR #5006, opened 2022-04-26
[cypress] Tests for istio config list page

nrfox:
Fixes #4969

PR #5064, opened 2022-05-11
[cypress] Workloads page testing

leandroberetta:
Resolves https://github.com/kiali/kiali/issues/4912.

Adds basic Cypress testing to the Workloads page.

Epic: https://github.com/kiali/kiali/issues/4854

### DIFF
--- a/frontend/cypress/integration/featureFiles/workloads.feature
+++ b/frontend/cypress/integration/featureFiles/workloads.feature
@@ -1,6 +1,6 @@
 Feature: Kiali Workloads page
   
-  User opens the Workloads page and sees the bookinfo namespaces
+  User opens the Workloads page and sees the bookinfo workloads.
 
   Background:
     Given user is at administrator perspective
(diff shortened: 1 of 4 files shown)


frontend/cypress/integration/featureFiles/workloads.feature
```diff
@@ -0,0 +1,60 @@
...
+    Then user sees a table with headings
+      | Health | Name | Namespace | Type | Labels | Details
+    And the "details-v1" row is visible
+    And the health column on the "details-v1" row has a health icon
+    And the "Name" column on the "details-v1" row has a link ending in "/namespaces/bookinfo/workloads/details-v1"
+    And the "Namespace" column on the "details-v1" row has the text "bookinfo"
+    And the "Labels" column on the "details-v1" row has the text "app=details"
+    And the "Labels" column on the "details-v1" row has the text "version=v1"
+    And the "Type" column on the "details-v1" row has the text "Deployment"
+    And the "Details" column on the "details-v1" row is empty
```
nrfox, line 21: These are more "procedural" in style. The preferred syntax is probably less procedural: https://github.com/kiali/kiali/pull/5006#discussion_r859226118. That being said, I don't think it's a blocker and we can consolidate later.

nrfox: The Istio config page is probably a better example that is a "list" page: https://github.com/kiali/kiali/blob/master/frontend/cypress/integration/featureFiles/istio_config.feature

frontend/cypress/integration/featureFiles/workloads.feature
```diff
@@ -0,0 +1,60 @@
...
+  Scenario: Filter workloads table by health
+    When user selects filter "Health"
+    And user filters for health "Healthy"
+    Then user sees "workloads" in workloads table
+    And user should only see healthy workloads in workloads table
+
+  @workloads-page
+  Scenario: Filter workloads table by label
+    When user selects filter "Label"
+    And user filters for label "app=details-v1"
```
nrfox, line 58: This actually yields a false positive. I believe it should be `"app=details"`. The filtering happens after the `Then user sees "details-v1"` assertion` so the assertion is checking the table pre-filter. The table does have a length of 1 but it has the "No workloads" element in there instead of `details-v1`. It would be better to switch the two assertions to ensure that the check for `details-v1` happens _after_ the filtering.

frontend/cypress/integration/featureFiles/workloads.feature
```diff
@@ -0,0 +1,60 @@
...
+
+  Background:
+    Given user is at administrator perspective
+    And user is at the "workloads" page
+    And user selects the "bookinfo" namespace
+
+  @workloads-page
+  Scenario: See a table with correct info
+    Then user sees a table with headings
+      | Health | Name | Namespace | Type | Labels | Details
```
nrfox, line 13:
```suggestion
      | Health | Name | Namespace | Type | Labels | Details |
```

frontend/cypress/integration/featureFiles/workloads.feature
nrfox: @leandroberetta to be clear, you do not need to re-write these tests in the more procedural syntax. Just fix the false positive then these tests LGTM.

leandroberetta: ok, thanks for the clarification, I was looking in the example to pasted.

nrfox (approved):

PR #5066, opened 2022-05-13
Fix some trace-detail related issues

jshaughn:
Epic #4260

- fix the Trace Detail header
  - use standard CardHeader components, slightly different look, hopefully
    as good or better.
- Change "Spans Details" tab to "Span Details", double-plural is typically
  avoided in English. Note that "Spans Detail" would be another option.
- Remove cellWidths on the span detail table.  This seems to layout better
  at the widths and data I tested; column headers are all fully visible
  and the heatmap renders well.
- Fix issue with Kebab not opening. This was actually a bad problem
  due to our DIY approach to adding the kebab menu to table rows. It
  resulted in the updated PF firing multiple events for a single kebab
  click.  In the end I ended up refactoring the SpanTable impl to
  both simplify things and use standard action support for PF tables.
  SpanTable is now consolidated in one component, the SpanTableItem
  helper has been removed, as has our single-use Expandable utility.


lucasponce:
image

jshaughn:
@lucasponce There is something odd about this table structure, I think.  I need to look more closely.  It's rebuilding the table on every render, I'm not sure that's correct, but I have to revisit tables a little bit.  So, still looking at this one.

lucasponce (approved):
This work looks good.
Thanks, @jshaughn for refactoring the DIY table and adapting to the regular Table structure.

I don't have more comments, I saw a couple of CSS minor issues, but nothing to prevent delay more this work.

The CI failures are unrelated to this PR, probably due lack of rebasing, but I'll proceed with the merging.

Any additional step can be done in a follow-up commit.

PR #5071, opened 2022-05-13
[cypress] Add tests for App details page

### LINKED ISSUE #5008: Add cypress UI tests around the Application details page
Tests for this page and all tabs:

Screenshot from 2022-04-26 16-20-34

Epic: #4854

nrfox:
Basic cypress tests for the App details page.

Fixes #5008

### DIFF
--- a/frontend/cypress/integration/common/app_details.ts
+++ b/frontend/cypress/integration/common/app_details.ts
@@ -92,7 +92,7 @@ And('user sees span details after selecting a trace', () => {
   // Blue dot on scatterplot is a trace so find one and click on it.
   const tracingDotQuery = '[style*="fill: var(--pf-global--palette--blue-200)"][style*="stroke: transparent"]';
   cy.getBySel('jaeger-scatterplot').find(`path${tracingDotQuery}`).first().should('be.visible').click({ force: true });
-  cy.getBySel('trace-details-tabs').should('be.visible').contains('Spans Details').click();
+  cy.getBySel('trace-details-tabs').should('be.visible').contains('Span Details').click();
 });
 
 And('user can filter spans by app', () => {
--- a/frontend/cypress/integration/common/app_details.ts
+++ b/frontend/cypress/integration/common/app_details.ts
@@ -1,4 +1,5 @@
-import { Then } from 'cypress-cucumber-preprocessor/steps';
+import { And, Then } from 'cypress-cucumber-preprocessor/steps';
+import { getCellsForCol } from './table';
 
 const APP = 'details';
 const NAMESPACE = 'bookinfo';
@@ -61,6 +62,8 @@ Then('user sees trace information', () => {
   openTab('Traces');
   cy.wait('@fetchTraces');
   cy.waitForReact(1000, '#root');
+  cy.getBySel('jaeger-scatterplot');
+  cy.contains('No traces').should('not.exist');
   cy.getReact('TracesComponent')
     .nthNode(1)
     .getCurrentState()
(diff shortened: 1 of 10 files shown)


frontend/cypress/integration/common/app_details.ts
```diff
@@ -0,0 +1,71 @@
...
+Then('user sees inbound metrics information', () => {
+  cy.intercept(Cypress.config('baseUrl') + `/api/namespaces/${NAMESPACE}/apps/${APP}/dashboard*`).as('fetchMetrics');
+  openTab('Inbound Metrics');
+  cy.wait('@fetchMetrics');
+  cy.waitForReact(1000, '#root');
+  cy.getReact('IstioMetrics', { props: { 'data-test': 'inbound-metrics-component' } })
+    // Theres be multiple components of the same name providing state through redux/hooks.
+    // This gets us the one we we want but this tests will fail if the context providers
+    // change names or are removed.
+    .nthNode(2)
```
leandroberetta, line 39: Just curious, why the second node?.

frontend/cypress/integration/common/app_details.ts
```diff
@@ -0,0 +1,71 @@
...
+      cy.wrap(state.dashboard).should('not.be.empty');
+    });
+});
+
+Then('user sees outbound metrics information', () => {
+  cy.intercept(Cypress.config('baseUrl') + `/api/namespaces/${NAMESPACE}/apps/${APP}/dashboard*`).as('fetchMetrics');
+  openTab('Outbound Metrics');
+  cy.wait('@fetchMetrics');
+  cy.waitForReact(1000, '#root');
+  cy.getReact('IstioMetrics', { props: { 'data-test': 'outbound-metrics-component' } })
```
leandroberetta, line 51: Just to confirm, you are grabbing a component, getting its state and then asserting the values of for example, the dashboard.

Maybe is for a followup PR but do we need to test this component checking of the information is correctly displayed right?.

leandroberetta (commented):
First quick review. Nice examples on how to test the state of components!. 

Some small questions for now!.

frontend/cypress/integration/common/app_details.ts
nrfox: There's two parent components from the router and from redux that both match the `cy.getReact('IstioMetrics')` query. We want the actual component that we added which is the bottom one at index 2. Wish there was a way to specify the "last" one but doesn't look like it.

Screenshot from 2022-05-13 14-12-52

frontend/cypress/integration/common/app_details.ts
nrfox: Yes I can add some assertions about a chart being rendered but since the actual elements in that chart are just svgs, I couldn't think of a good way to tell if these had displayed correctly. So I just test that the component has the correct state which at least tests that the jaeger integration is working properly. If you can think of a way to test that the charts have rendered correctly I'm happy to update. The tests are pretty sparse though. I can at least add tests for interacting with the "display" menu.

leandroberetta: Yes, I understand, and it's true. Well, this comment only applies for other tests where it's more suitable, but of course you won't be testing the state in those.

frontend/cypress/integration/common/app_details.ts
leandroberetta: I see, thanks.

leandroberetta (approved):
LGTM.

nrfox:
Added new tests for the "trace" and "span" details tabs. Rebased with master.

nrfox:
Some test failures in CI

israel-hdez (dismissed):
All good. Just minor questions and comments.

frontend/src/components/JaegerIntegration/TracesComponent.tsx
```diff
@@ -266,6 +266,7 @@ class TracesComponent extends React.Component<TracesProps, TracesState> {
             >
               <Tabs
                 id="trace-details"
+                data-test="trace-details-tabs"
```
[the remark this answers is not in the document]
nrfox: Yeah I'm just following the cypress recommendation for selectors: https://docs.cypress.io/guides/references/best-practices#Selecting-Elements. In practice, I've seen patternfly increment the id automatically for some elements. Although that's not why I added it here.

frontend/cypress/integration/common/app_details.ts
```diff
@@ -0,0 +1,107 @@
...
+  cy.contains('Outbound Traffic');
+});
+
+Then('user sees inbound metrics information', () => {
+  cy.intercept(Cypress.config('baseUrl') + `/api/namespaces/${NAMESPACE}/apps/${APP}/dashboard*`).as('fetchMetrics');
+  openTab('Inbound Metrics');
+  cy.wait('@fetchMetrics');
+  cy.waitForReact(1000, '#root');
+  cy.getReact('IstioMetrics', { props: { 'data-test': 'inbound-metrics-component' } })
+    // HOCs can match the component name. This filters the HOCs for just the bare component.
```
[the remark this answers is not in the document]
nrfox: Higher Order Component. It's a component that wraps another component with some functionality and seems to be how some of the state handling is implemented like `withRouter` or `connect` for redux. Kind of a like a python decorator.

frontend/cypress/integration/common/app_details.ts
```diff
@@ -0,0 +1,107 @@
...
+    .then((metricsComponents: any) => metricsComponents.filter(component => component.name === 'IstioMetrics')[0])
+    .getCurrentState()
+    .then(state => {
+      cy.wrap(state.dashboard).should('not.be.empty');
```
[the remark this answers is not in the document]
nrfox: To be honest, wasn't totally sure what the best way to test for the existence of metrics is. I can check for those components but I don't think it'll check that there's actually lines drawn on them from the existence of metrics. Those components could be there but just be empty.

leandroberetta (dismissed):
LGTM too.

frontend/cypress/integration/common/app_details.ts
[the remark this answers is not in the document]
israel-hdez: Yes, I understand.
I may think we may need to better define what should be in scope for cypress/cucumber and what not. IMO, a few of the tests that we already have in `master` would be more robust in units.

But OK, then no change is needed here :+1:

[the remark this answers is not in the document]
nrfox: Yes some of these tests are probably better served as unit tests. In this case, I wanted to test that the charts actually showed you something when there's metrics present without testing what the chart was actually rendering e.g. the svg components. What I ended up testing was that the component has some non-empty state for the dashboards and presuming that it translates directly to "lines a graph".

nrfox:
Rebased and added the `at administrator perspective` statement

israel-hdez (approved):
I retried and it is still all green.

PR #5079, opened 2022-05-16
[cypress] add health indicator checks for workloads and services lists

israel-hdez:
Add cypress tests for health indicators in workloads and services list pages.

Looks like tests in the overview page does not apply, since the health indicator on overview is focused only on apps.

### DIFF
--- a/frontend/cypress/integration/featureFiles/services.feature
+++ b/frontend/cypress/integration/featureFiles/services.feature
@@ -60,3 +60,34 @@ Feature: Kiali Services page
     And user filters for label "app=productpage"
     Then user sees "productpage" in the table
     And table length should be 1
+
+  @services-page
+  Scenario: The healthy status of a service is reported in the list of services
+    Given a service in the cluster with a healthy amount of traffic
+    When I fetch the list of services
+    And user selects the "bookinfo" namespace
+    Then the service should be listed as "healthy"
+
+  @services-page
+  Scenario: The idle status of a service is reported in the list of services
+    Given a service in the cluster with no traffic
+    When I fetch the list of services
+    And user selects the "default" namespace
+    Then the service should be listed as "na"
+    And the health status of the service should be "No health information"
+
+  @services-page
+  Scenario: The failing status of a service is reported in the list of services
+    Given a service in the mesh with a failing amount of traffic
+    When I fetch the list of services
+    And user selects the "alpha" namespace
+    Then the service should be listed as "failure"
+    And the health status of the service should be "Failure"
+
+  @services-page
+  Scenario: The degraded status of a service is reported in the list of services
+    Given a service in the mesh with a degraded amount of traffic
+    When I fetch the list of services
+    And user selects the "alpha" namespace
+    Then the service should be listed as "degraded"
+    And the health status of the service should be "Degraded"
--- a/frontend/cypress/integration/featureFiles/services.feature
+++ b/frontend/cypress/integration/featureFiles/services.feature
@@ -5,10 +5,10 @@ Feature: Kiali Services page
   Background:
     Given user is at administrator perspective
     And user is at the "services" page
-    And user selects the "bookinfo" namespace
 
   @services-page
   Scenario: See a table with correct info
+    When user selects the "bookinfo" namespace
     Then user sees a table with headings
       | Health | Name | Namespace | Labels | Configuration | Details |
     And the "productpage" row is visible
@@ -22,41 +22,47 @@ Feature: Kiali Services page
     And the "Details" column on the "productpage" row has a link ending in "/namespaces/bookinfo/istio/gateways/bookinfo-gateway"
 
   @services-page
-  Scenario: Filter services table by Service Name 
-    When user selects filter "Service Name"
+  Scenario: Filter services table by Service Name
+    When user selects the "bookinfo" namespace
+    And user selects filter "Service Name"
     And user filters for name "productpage"
     Then user sees "productpage" in the table
     And table length should be 1
   
   @services-page
   Scenario: Filter services table by Service Type
-    When user selects filter "Service Type"
+    When user selects the "bookinfo" namespace
+    And user selects filter "Service Type"
     And user filters for service type "External"
     Then user sees "nothing" in the table
   
   @services-page
   Scenario: Filter services table by sidecar
-    When user selects filter "Istio Sidecar"
+    When user selects the "bookinfo" namespace
+    And user selects filter "Istio Sidecar"
     And user filters for sidecar "Present"
     Then user sees "something" in the table
 
   @services-page
   Scenario: Filter services table by Istio Type
-    When user selects filter "Istio Type"
+    When user selects the "bookinfo" namespace
+    And user selects filter "Istio Type"
     And user filters for istio type "VirtualService"
     Then user sees "productpage" in the table
     And table length should be 1
   
   @services-page
   Scenario: Filter services table by health
-    When user selects filter "Health"
+    When user selects the "bookinfo" namespace
+    And user selects filter "Health"
     And user filters for health "Healthy"
     Then user sees "something" in the table
     And user should only see healthy services in the table
 
   @services-page
   Scenario: Filter services table by label
-    When user selects filter "Label"
+    When user selects the "bookinfo" namespace
+    And user selects filter "Label"
     And user filters for label "app=productpage"
     Then user sees "productpage" in the table
     And table length should be 1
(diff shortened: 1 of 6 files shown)


frontend/cypress/integration/featureFiles/services.feature
```diff
@@ -60,3 +60,34 @@ Feature: Kiali Services page
     And user filters for label "app=productpage"
     Then user sees "productpage" in the table
     And table length should be 1
+
+  @services-page
+  Scenario: The healthy status of a service is reported in the list of services
+    Given a service in the cluster with a healthy amount of traffic
+    When I fetch the list of services
```
nrfox, line 67: You don't need to include this check here since the Background already navigates to the services page with the check `And user is at the "services" page`. Moreover, the user isn't really fetching the list of services themselves, the app is doing that. The user is just navigating to a particular page and expecting to see a list of services show up after the page loads.

frontend/cypress/integration/featureFiles/services.feature
```diff
@@ -60,3 +60,34 @@ Feature: Kiali Services page
...
+    And user selects the "bookinfo" namespace
+    Then the service should be listed as "healthy"
+
+  @services-page
+  Scenario: The idle status of a service is reported in the list of services
+    Given a service in the cluster with no traffic
+    When I fetch the list of services
+    And user selects the "default" namespace
```
nrfox, line 75: This will require updating the rest of the tests and not just the ones you've added but if we're testing things in namespaces outside of `bookinfo` we should probably move the `And user selects the "bookinfo" namespace` statement out of Background and into the individual tests.

frontend/cypress/integration/featureFiles/workloads.feature
```diff
@@ -58,3 +58,34 @@ Feature: Kiali Workloads page
     And user filters for label "app=details"
     Then user sees "details-v1" in the table
     And table length should be 1
+
+  @workloads-page
+  Scenario: The healthy status of a workload is reported in the list of workloads
+    Given a healthy workload in the cluster
+    When I fetch the list of workloads
+    And user selects the "bookinfo" namespace
+    Then the workload should be listed as "healthy"
```
nrfox, line 67: Same comments here as on the services page

frontend/cypress/integration/featureFiles/services.feature
israel-hdez: In general, I think it will be very hard to test a single Kiali page using the preset resources of only one namespace...

israel-hdez: And well... I'll move that statement outside the background.

frontend/cypress/integration/featureFiles/services.feature
israel-hdez: > You don't need to include this check here since the Background already navigates to the services page with the check `And user is at the "services" page`.

I'm not thinking in \<the services page\> as initial context, but more as an action that the user would want to do :confused: 

> The user is just navigating to a particular page and expecting to see a list of services show up after the page loads.

That would be entering in implementation details, no? :confused: 

I'm trying to follow what is mentioned in this small page: https://cucumber.io/docs/bdd/examples/

I know it is very hard to assume no computers in our case. But we can make the assumption that Kiali does not exist and you still need to "see" the healthy status of a service. Perhaps, somebody will need to prepare a list on paper and would place colored sticky markers reflecting health status of each item in the list...

Well... I'll try to see if there is some way to fix what you are commenting here.

israel-hdez: > the Background already navigates to the services page

BTW, I'm writing a few new tests for the Istio Config page. In these tests I need to apply changes to the env (similar to the sidecar tests).

There is the common pattern that the `Background` is pre-navigating to the Istio Config page. However, since env changes come after the `Background`, I anyway need to refresh the page via a `When the user fetches the list of Istio CRDs` sentence, similarly to how I was doing here. So, that makes me doubt if the `Background` is really a good place to do pre-navigation if you may need to do more preparation in `Given`s on each scenario.

israel-hdez: I just created #5085, a small draft PR of the other tests I'm starting to write, to let you see the need for reloading the page even if the same page is already loaded in the `Background`.

It's just to reach an agreement about how it should be done and use the same pattern here and there, as I think all tests should be consistent.

nrfox: > I'm not thinking in <the services page> as initial context, but more as an action that the user would want to do confused
> 
> > The user is just navigating to a particular page and expecting to see a list of services show up after the page loads.
> 
> That would be entering in implementation details, no? confused

That's fair the wording could probably be updated to be something less implementation specific like `Given I am viewing my "services"` or `Given I am viewing my "bookinfo/details" "service"` but, for me, "fetching" doesn't really describe what the user is doing here. They're not opening up a network request to go get each service or as in your sticky note example writing out each note and arranging it themselves, rather they are just viewing all their services. Maybe this isn't strictly adhering to the Gherkin syntax but I also think it's more understandable to describe the tests in terms of what we're actually testing i.e. we're testing the services list page.

> There is the common pattern that the Background is pre-navigating to the Istio Config page. However, since env changes come after the Background, I anyway need to refresh the page via a When the user fetches the list of Istio CRDs sentence, similarly to how I was doing here. So, that makes me doubt if the Background is really a good place to do pre-navigation if you may need to do more preparation in Givens on each scenario.

Couldn't you also hit the refresh button after patching your env? That's what a user would do to have Kiali pickup new changes to their env when they've paused the autorefresh like we do in the tests.

From the docs on the Background section:
> Occasionally you’ll find yourself repeating the same Given steps in all of the scenarios in a Feature.
>
> Since it is repeated in every scenario, this is an indication that those steps are not essential to describe the scenarios; they are incidental details. You can literally move such Given steps to the background, by grouping them under a Background section.

The fact that we're on the services page is an incidental detail here since we have to be on the services page in order to test the services page. Every test otherwise would have that `Given` step. If the page loading doesn't belong in the `Background` section, I'm not really sure what does.

frontend/cypress/integration/featureFiles/services.feature
nrfox: > In general, I think it will be very hard to test a single Kiali page using the preset resources of only one namespace...

That's probably true eventually but so far many of the tests have only relied on a single namespace. If all scenarios in a test require a single namespace to be selected then it's `Background`. Once the namespace differs between scenarios then it's no longer `Background`.

frontend/cypress/integration/featureFiles/services.feature
israel-hdez: > they are just viewing all their services.

Well, I could use `When the user looks at the services list`, if that sounds better. The implementation would remain, though.

> Maybe this isn't strictly adhering to the Gherkin syntax but I also think it's more understandable to describe the tests in terms of what we're actually testing 

I agree. But IMHO that route is not aligned with cucumber vision.

> Couldn't you also hit the refresh button after patching your env?

That's an option, I think.

> From the docs on the Background section:

Well, I agree with the docs. But `Background` is limited to `Given` steps, while this is a `When`.

Not sure, but I'm so-so reading that you want this `When` converted to a `Given`. I know that in typescript all these things are aliases. But, conceptually they are different. So, I'm not sure that we should be changing/using the terms indistinctly.

> The fact that we're on the services page is an incidental detail here since we have to be on the services page in order to test the services page. Every test otherwise would have that Given step. If the page loading doesn't belong in the Background section, I'm not really sure what does.

IMHO, that is too focused on implementation. And about the background, well, the only thing I see valid without entering in implementation details is the sentence telling that the logged in user is an administrator. Change that background to a non-admin user and you can then start thinking in very different scenarios.

All that said... well... I'm not leading this effort. So, meh, I'll rework this PR to your liking.

nrfox: > Not sure, but I'm so-so reading that you want this When converted to a Given. I know that in typescript all these things are aliases. But, conceptually they are different. So, I'm not sure that we should be changing/using the terms indistinctly.

Yeah `When` statements don't seem to make much sense in `Background`. Maybe something like `Given user is looking at their services list`?

> All that said... well... I'm not leading this effort. So, meh, I'll rework this PR to your liking.

It's a group effort and you've spent much more time reading the cucumber docs/recommendations. I appreciate your input and perspective on how we're implementing these tests. Adherence to cucumber has not been as much of a priority as getting coverage and encouraging folks to write tests. Ultimately whichever syntax makes the tests easier to read, write, and maintain for contributors is the way I prefer.

frontend/cypress/integration/featureFiles/services.feature
israel-hdez: It is moved.

frontend/cypress/integration/featureFiles/workloads.feature
israel-hdez: Hopefully, I also fixed this.

nrfox (approved):

frontend/cypress/integration/featureFiles/services.feature
nrfox: > It makes me wonder why we chose it, if perhaps something else would have been a better fit (like the regular specs).

That's a good question. I believe it was primarily chosen so that the testing requirements for a feature could be formally defined separately from the feature itself. QE could write/review the feature file independently of the implementation. I think it still serves that purpose even if we're not following the recommended syntax exactly.

PR #5080, opened 2022-05-17
[e2e] Added a support of new env variable OC_COMMAND

hhovsepy:

Added support of providing oc or kubectl command name to execute.

So now, for executing e2e integration tests on environment where `oc` command is not available, but `kubectl` is there, the `-e OC_COMMAND="kubectl"` env variable should be provided.

### DIFF
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -10,7 +10,7 @@
 ## Environment Setup
 
 * System dependencies that will need to be available
-    *  `oc`
+    *  `oc` or `kubectl`
     *  `go`
     *  `make`
     *  `npm`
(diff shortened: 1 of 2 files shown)


Xunzhuo (approved):
LGTM

lucasponce (commented):
Mostly ok, I think something is missing in the tests/integration/README.md and/or in the expected installation steps as the URL steps fail for me

image

I'm going to repeat my steps in case I had something wrong in my env.

lucasponce (approved):
This PR is good, I think perhaps we need to keep iterating to be sure that the instructions for the local test environment are good, but it's a minor comment.

PR #5087, opened 2022-05-18
Wait for given namespace to be removed when uninstalling bookinfo app

FilipB:
In cases when the CI is removing existing demo app and installing new one, it often fails because the removal operation is not blocking and the namespace is still being terminated when trying to install new app there.
This will wait for namespace to be removed.

There is no timeout defined but if it ever gets stuck on this operation, it will require manual interaction with the cluster anyway.

### DIFF
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -200,6 +200,15 @@ if [ "${DELETE_BOOKINFO}" == "true" ]; then
   else
     $CLIENT_EXE delete namespace ${NAMESPACE}
   fi
+  # It takes some time to delete the namespace, let's wait for that
+  echo -n "Waiting for ${NAMESPACE} to be deleted..."
+  while $CLIENT_EXE get namespace ${NAMESPACE} &> /dev/null
+  do
+    echo -n '.'
+    sleep 10
+  done
+  echo ""
+
   echo "====== BOOKINFO UNINSTALLED ====="
   exit 0
 fi
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -197,18 +197,11 @@ if [ "${DELETE_BOOKINFO}" == "true" ]; then
     fi
     $CLIENT_EXE delete scc bookinfo-scc
     $CLIENT_EXE delete project ${NAMESPACE}
+    # oc delete project does not wait for a namespace to be removed, we need to also call 'oc delete namespace'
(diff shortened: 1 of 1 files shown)


hack/istio/install-bookinfo-demo.sh
```diff
@@ -200,6 +200,15 @@ if [ "${DELETE_BOOKINFO}" == "true" ]; then
...
   fi
+  # It takes some time to delete the namespace, let's wait for that
+  echo -n "Waiting for ${NAMESPACE} to be deleted..."
+  while $CLIENT_EXE get namespace ${NAMESPACE} &> /dev/null
+  do
+    echo -n '.'
+    sleep 10
+  done
+  echo ""
+
```
jmazzitelli, line 211:
```suggestion
    $CLIENT_EXE delete namespace ${NAMESPACE} --wait=true
  fi
```

jmazzitelli (changes requested):
I think you did a little overkill here :) You can use the `--wait=true` option with the delete command

hack/istio/install-bookinfo-demo.sh
nrfox: wait should be `true` by default. Does `oc project delete` not wait for the namespace to be deleted before returning?

jmazzitelli:
```
$ oc delete --help | grep wait=true
      --wait=true: If true, wait for resources to be gone before returning. This waits for finalizers.
```

jmazzitelli: But I don't know if that is the default or not.

nrfox: Whatever is after the `=` is the default so `true` is the default. Does the `project` resource stick around until the `namespace` gets deleted or does `project` get removed without waiting for the `namespace` to be deleted?

nrfox: Either way adding `$CLIENT_EXE delete namespace` should fix it but I think you also need to add it to the `IS_OPENSHIFT=true` case.

FilipB:
Hmm, so 'oc delete project' does not wait for namespace to be removed. 'oc delete namespace' does.

jmazzitelli (approved):
very interesting. I'll have to look at my other hack scripts that might also do a `delete project` -- I think our istio install hack script might also do this (I'll have to check)

hack/istio/install-bookinfo-demo.sh
FilipB: Added new commit using 'delete namespace'

Issue #4864, opened 2022-03-28
Kiali and Istio validation messages should be placed together


lucasponce:
This is an improvement of the sidepanel in the Istio Config details.

I think that the "KIA0301" message should be also be referenced in the "Configuration Analysis" section on the side panel.

The current inline warning is good, this issue is an enhancement to place them together in the side panel.

![image](https://user-images.githubusercontent.com/1662329/160378651-7747ac3e-3a8a-41e6-b67d-40b79ce92f80.png)

Perhaps with this change, the validation icon on the resource name:
![image](https://user-images.githubusercontent.com/1662329/160378944-c5dee8b4-c442-45d9-b3af-1eac9eaaa50c.png)

is not necessary if we are going to list the validation issues.

hhovsepy:
Hi @lucasponce, can you still see the "Configuration Analysis" section in side panel?

lucasponce:
The configuration analysis messages are not enabled by default, those are defined at the mesh level at the installation phase (I think). 

I typically setup my cluster with these flags:

```
# --set values.global.istiod.enableAnalysis=true
# --set values.pilot.env.PILOT_ENABLE_STATUS=true
```

to enable them

PR #5090, opened 2022-05-18
Added Kiali validation messages alongside with Istio Analysis.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4864

When configuration analysis messages are enabled on Istio, Kiali shows valiation messages alongside with these Istio configuration analysis messages. Also removed object status icon in that cases.

Screenshot from 2022-05-18 19-50-21

### DIFF
--- a/frontend/cypress/integration/common/wizard_request_routing.ts
+++ b/frontend/cypress/integration/common/wizard_request_routing.ts
@@ -121,6 +121,16 @@ And('user creates the configuration', () => {
         .should('not.exist');
 });
 
+And('user updates the configuration', () => {
+    cy.get('button[data-test="update"]')
+        .click();
+
+    cy.get('button[data-test="confirm-update"]')
+        .click()
+        .get('#loading_kiali_spinner')
+        .should('not.exist');
+});
+
 And('user confirms delete the configuration', () => {
     cy.get('button[data-test="confirm-delete"]')
         .click()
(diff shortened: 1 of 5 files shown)


lucasponce:
This PR is great. I wonder if you could use it to introduce a simple UI test around the "Istio Config details page" with a validation.

It doesn't need to focus on the mesh validations, as those are not enabled by default in the CIs, but just a test that checks that the validation is present in the side panel.

Then we can have that test prepared for future use cases.

(Again, sorry to ask, but as we are starting to work in features/bug fixes, I'm starting to introduce these testing changes in the regular PR work)

Thanks for starting this effort @hhovsepy

lucasponce:
I think you could re-use some of the tests done in the wizard:
image

I mean, that test navigates from the Service details page to the Istio Config details to validate the generated config, but for this case, it could be simpler, just to add a wrong configuration and check that the validation happens in the sidepanel (in the ace editor perhaps it can be to search a single selector, but if it's not clear, we can add it later, I know there are some pending issues in that area from the PF4 migration).

hhovsepy:
@lucasponce added two more tests, including verification of warning in config details.

frontend/cypress/integration/common/wizard_request_routing.ts
```diff
@@ -158,4 +168,23 @@ And('user sees the {string} regex in the editor', (regexContent) => {
     cy.get('.ace_content')
         .invoke('text')
         .should('match', re);
+});
+
+And('user clicks on Show Advanced Options', () => {
+    cy.get('.pf-c-expandable-section__toggle-text')
```
lucasponce, line 174: Sorry for my delay in this review.
I was also checking if this selector could be changed for some custom data-test attribute, but it seems that the component is wrapped in a way that there is no obvious way to add data- attributes or reuse an existing aria- attribute.

lucasponce: I was wondering if "contentId" could make the trick here, to avoid to query just the style:
image

But not a blocker.

lucasponce (approved):
I have a minor comment about one selector in the cypress test.
But not a blocker.

frontend/cypress/integration/common/wizard_request_routing.ts
hhovsepy: +1, thanks for the hint and review

lucasponce (approved):

PR #5095, opened 2022-05-19
Remove storybook tests as part of the PF-React update

jshaughn:
The extensive transitive deps are making other dep updates difficult, as well as triggering
several dependabot security warnings.  Moreover, we don't pay attention to
these tests.

@jmazzitelli to test I think just pull, rebuild node_modules, run `yarn test` and `yarn run build`.  If it aworks I think it's fine.


jmazzitelli:
`make clean-ui build-ui` builds successfully.
`cd frontend && yarn test` is successful

jmazzitelli (approved):

aljesusg:
Well we did the storybooks to verify that charts metrics render correctly so if we have this in cypress I don't see any reason to not remove this :)

PR #5096, opened 2022-05-19
Remove support for `yarn run analyze`.  The command is broken and the

jshaughn:
Epic https://github.com/kiali/kiali/issues/4260

source-map-explorer dependency has dependabot issues.  I don't think
anyone runs this, if so it can be added back by someone, in a working
fashion.


jmazzitelli (approved):
i'm ok with this only because I don't even know what `analyze` gave us. I never used it myself, that's for sure. I don't know anyone that has used it recently either.

PR #5105, opened 2022-05-23
Fix package.json version and release pipeline

leandroberetta:
In the last release, the package.json version did not match the requirements of the version checker, it has an extra "v" prefixed. 

This PR fix the version and also the pipeline that modifies that.

Next time the version checker should pass.


aljesusg (approved):
LGFM.

Issue #5073, opened 2022-05-13
UI tests around the Graph page Find/Hide


nrfox:
Add cypress tests for Graph find/hide feature.

Epic: https://github.com/kiali/kiali/issues/4854

PR #5108, opened 2022-05-24
Add graph find/hide cypress tests

nrfox:
Adds cypress tests for graph find/hide.

Tests:
- Finding unhealthy nodes
- Hiding unhealthy nodes
- Using find/hide presets
- Opening help menu

Fixes: #5073


jshaughn (approved):
Probably good test practice, but slower and a little tedious to reset every test to the initial graphing of namespaces.   Could have tested a few more things but this is sufficient for a sanity check of the feature.

jshaughn (approved):
Re-approving, CI Passes after latest change.

Issue #4896, opened 2022-04-04
Add outboundTrafficPolicy value to overview istio-system card 


ceastman-r7:
This would be helpful to see what the current outboundTrafficPolicy is set to.

<img width="516" alt="Screen Shot 2022-04-04 at 8 05 16 AM" src="https://user-images.githubusercontent.com/97614291/161550532-ee073408-d8f8-451c-a82f-5e6cd9135378.png">

lucasponce:
Thanks! Yes, +1, I agree.
This namespace overview summary is focused on the "inbound traffic" but it makes all sense to offer both.

hhovsepy:
One idea of UX is to add a Traffic direction dropdown in Overview toolbar to show 'Outbound' or "Inbound" traffics:
![Screenshot from 2022-05-24 17-20-03](https://user-images.githubusercontent.com/604313/170073026-77f310c1-f5fe-4059-95db-68f4bf0597e2.png)

lucasponce:
cc @andrew-ronaldson

andrew-ronaldson:
There needs to be some padding following the Apps dropdown before traffic direction text. Could I see a screenshot with the full screen, please. Thanks

hhovsepy:
padding sure will be added, screenshot of the suggested solution:
![Screenshot from 2022-05-24 22-53-12](https://user-images.githubusercontent.com/604313/170130284-6e6052f3-97de-45c9-a774-397c030f219d.png)

andrew-ronaldson:
what is the "ops" below applications? If it is an axis label for the chart there should be some padding between the list and the chart

leandroberetta:
> what is the "ops" below applications? If it is an axis label for the chart there should be some padding between the list and the chart

Yes, it is the axis label. Do you mean bring the "ops" closer to the chart, separating it a little bit of the other section?. 

Another comment not related direclty with this issue but, would be better to change ops for req (requests)?.

andrew-ronaldson:
If there was a vertical line on the chart it would be clearer that ops is the label on that axis. With the ops and outbound traffic on the same if you add some space above it would help break up the content in the card.

hhovsepy:
The OutBoundTrafficPolicy.Mode will be shown in Overview page when configured:
![Screenshot from 2022-06-24 12-29-04](https://user-images.githubusercontent.com/604313/175648701-8b91782b-caa6-4b3e-9e44-e91c6e2f2b89.png)

[...]

PR #5117, opened 2022-05-25
Outbound traffic in Overview cards. Added dropdown.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4896

Added Traffic Direction dropdown in Overview toolbar to select Outbound and Inbound traffic to show. Selection triggers reload of Overview items and the traffic with particular direction.

Added cypress tests.

Screenshot from 2022-05-25 18-04-08

### DIFF
--- a/frontend/cypress/integration/common/overview.ts
+++ b/frontend/cypress/integration/common/overview.ts
@@ -125,9 +125,9 @@ When(`user selects {string} time range`, (interval) => {
         .should('not.exist');
 });
 
-When(`user selects {string} traffic duration`, (duration) => {
+When(`user selects {string} traffic direction`, (direction) => {
     let innerId = '';
-    switch (duration) {
+    switch (direction) {
         case 'Outbound':
             innerId = 'outbound';
             break;
--- a/frontend/cypress/integration/common/overview.ts
+++ b/frontend/cypress/integration/common/overview.ts
@@ -125,6 +125,26 @@ When(`user selects {string} time range`, (interval) => {
         .should('not.exist');
 });
 
+When(`user selects {string} traffic duration`, (duration) => {
+    let innerId = '';
+    switch (duration) {
+        case 'Outbound':
+            innerId = 'outbound';
+            break;
+        case 'Inbound':
+            innerId = 'inbound';
+            break;
+    }
+    cy.get('button[aria-labelledby^="direction-type"]')
+        .click()
+        .get('#loading_kiali_spinner')
+        .should('not.exist');
+    cy.get(`li[id="${innerId}"]`).children('button')
+        .click()
+        .get('#loading_kiali_spinner')
+        .should('not.exist');
+});
+
 When('I fetch the overview of the cluster', function () {
     cy.visit('/console/overview?refresh=0');
 });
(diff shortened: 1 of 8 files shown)


frontend/cypress/integration/common/overview.ts
```diff
@@ -125,6 +125,26 @@ When(`user selects {string} time range`, (interval) => {
         .should('not.exist');
 });
 
+When(`user selects {string} traffic duration`, (duration) => {
```
leandroberetta, line 128: Typo, I guess it should be direction.

I guess we can add the type to the parameter? (direction: string)

leandroberetta (commented):
@hhovsepy very nice addition to the overview page. 

Everything LGTM, only some minor comments, one is below and the other is an aesthetic comment:

overview_in_out_traffic

I guess we need more space between the chart and the "ops" axis label, but also following Andrew's comment on the issue, more space separating the section above.

lucasponce:
Minor:
image

"Traffic direction" -> "Traffic"

I think that in the chart we use the "inbound/outbound traffic" then the text can be shorter, which would help to not have a very large populated toolbar

lucasponce:
Some note, I'm reviewing the graph vs the overview to check that information is synced:
image

Queries used are correct:
image

But I saw some gap in the Graph sidebar, "total" is populated:
image


But, "inbound" and "outbound" are empty:
image
image

It sounds a regression for @jshaughn but it's out of the scope of this PR, just leaving the comment here.

lucasponce (changes requested):
LGTM
A minor suggestion to "shorten" the name to avoid a populated toolbar, but the work looks good.

hhovsepy:
Have updated the paddings and the toolbar text.

hhovsepy:
"MeshConfig.OutboundTrafficPolicy.Mode" flag will be added in a scope of different PR.

hhovsepy:
Screenshot from 2022-05-27 12-45-37

lucasponce (approved):

jshaughn:
@lucasponce , I'll look at that graph side panel.  One comment for you and @hhovsepy re: this PR:  Perhaps the new dropdown should be disabled when using small cards, because they have no graph?

jshaughn:
> It sounds a regression for @jshaughn but it's out of the scope of this PR, just leaving the comment here.

Actually this is working as expected.  There is no `Incoming` traffic into the selected namespace, and no outgoing traffic from the namespace.  The total reflects the traffic within the namespace.

lucasponce:
Then this picture:
image

The "total" is the sum of the inbound of the workload and the outbound of the workloads, but it doesn't necessarily mean out of the namespace.

am I right?

jshaughn:
I believe the numbers and sparklines are all as intended.  The reason it's confusing is that "Inbound" and "Outbound" have some context-specific semantics, and there is currently no tooltip or other help to actually explain what it means in the different places.  

For example, for the full graph side-panel, "Inbound" is defined as traffic originating from a traffic source, i.e. the traffic flowing into the graph.  This makes sense.   For the namespace side-panel "Inbound" means traffic entering the namespace from outside the namesapce.  This also makes sense.  "Total" in all cases (I think) means all of the traffic for the object under observation.

Note that the community has never really complained, so I'm not sure they care, or look carefully at some of this stuff.  And I think we should prioritize changes accordingly (meaning I'm not in favor of a lot of work here).

lucasponce:
> Actually this is working as expected. There is no Incoming traffic into the selected namespace and no Outgoing traffic from the namespace. The total reflects the traffic within the namespace.

I was thinking more of a minor tooltip to explain this, here and potentially on the overview page. The queries look right but I think that contextual information should deserve some notes.

Issue #4782, opened 2022-03-03
Add validation for checking Service port and protocol consistency


hhovsepy:
Currently in Services when port and protocol are configured in an inconsistent way, for instance port is set '443' but the protocol is set 'http', there is not Validation message shown.

RFE: Add a validation to check that all the supported protocols "{"grpc", "http", "http2", "https", "mongo", "redis", "tcp", "tls", "udp", "mysql"}" and configured ports in Services are consistent.

jshaughn:
Closing for inactivity.  Can be re-opened if a problem is encountered.

PR #4765, opened 2022-02-28
Skip KIA0701 Port missmatch validation check for Services in istio-system namespace.

hhovsepy:
Fixing https://github.com/kiali/kiali/issues/4764.

For the Services in istio-system control plane namespace, skip "KIA0701 Deployment exposing same port as Service not found" validation.
But keep it for other namespaces.

For QE:
In "istiod", "tracing" and other Services from istio-system namespace, verify that validation warning "Deployment exposing same port as Service not found" is not shown.
Before:
Screenshot from 2022-02-28 10-56-57

After:
Screenshot from 2022-02-28 12-57-41

Regression test if Services from other nemaspaces continue getting validation warning: https://raw.githubusercontent.com/Kiali-QE/kiali-qe-python/master/data/resources/istio_objects/validation/ratings_java_svc.yaml
Screenshot from 2022-02-28 12-58-28

PR #5119, opened 2022-05-26
Added new validation check for Service port protocol matching.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4782

Added new validation message for Service port and protocol consistency matching.

Screenshot from 2022-05-26 12-28-45
Screenshot from 2022-05-26 12-29-07

### DIFF
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -30,7 +30,9 @@ const (
 
 var (
 	portNameMatcher = regexp.MustCompile(`^[\-].*`)
-	portProtocols   = [...]string{"grpc", "http", "http2", "https", "mongo", "redis", "tcp", "tls", "udp", "mysql"}
+	portProtocols   = map[string][]int32{"grpc": {443}, "http": {80}, "http2": {443, 80}, "https": {443},
+		"mongo": {27017, 27018, 27019}, "redis": {6379}, "tcp": {}, "tls": {}, "udp": {},
+		"mysql": {3306, 33060, 33061}}
 )
 
 type IstioClientInterface interface {
@@ -526,7 +528,7 @@ func MatchPortNameRule(portName, protocol string) bool {
 }
 
 func MatchPortNameWithValidProtocols(portName string) bool {
-	for _, protocol := range portProtocols {
+	for protocol := range portProtocols {
 		if strings.HasPrefix(portName, protocol) &&
 			(strings.ToLower(portName) == protocol || portNameMatcher.MatchString(portName[len(protocol):])) {
 			return true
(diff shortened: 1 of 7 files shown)


hhovsepy:
Documentation PR https://github.com/kiali/kiali.io/pull/536

kubernetes/istio.go
```diff
@@ -30,7 +30,9 @@ const (
 
 var (
 	portNameMatcher = regexp.MustCompile(`^[\-].*`)
-	portProtocols   = [...]string{"grpc", "http", "http2", "https", "mongo", "redis", "tcp", "tls", "udp", "mysql"}
+	portProtocols   = map[string][]int32{"grpc": {443}, "http": {80}, "http2": {443, 80}, "https": {443},
+		"mongo": {27017, 27018, 27019}, "redis": {6379}, "tcp": {}, "tls": {}, "udp": {},
```
lucasponce, line 34: I'm reviewing this PR. @hhovsepy do you remember how Istio defines these mappings.
I remember this was something defined at the Envoy level, but those names/numbers were defined in some variables?

lucasponce:
I'm reviewing this but I have to catch up a little bit more with the Istio standards, because it's pretty normal that services may define ports like HTTP with different numbers of "80" like

image

Then, this new validation is going to create a lot of new warnings, making this a little bit noise.

I guess that if Istio is managing a "table" of names/port, perhaps we could refer to that.

Anyway, I don't touch this area, I'm going to investigate this (to refresh my memories and I'll be back to this work).

lucasponce:
@hhovsepy I think we should review this RFE, I don't think that Istio enforces to use of particular ports for particular services

https://istio.io/latest/docs/ops/deployment/requirements/#pod-requirements
https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

If I understand this feature, the "explicit protocol selection" helps Istio to determine if a port is of one type (to not treat that traffic as just TCP one), but it doesn't force me to use any specific port number (unless I missed something).

So, this new validation would introduce a lot of noise in general, I think the KIA0701 validation was ok because it can help if pod and service aren't in sync. But this KIA0603 would introduce more noise in normal scenarios (i.e. an HTTP port is not in the 80).

I'd recommend that we can validate the RFE to see if we missed some info, or what was the original request.

hhovsepy:
The original issue was created after working on https://github.com/kiali/kiali/pull/4765 , could not find the suggestion which was made in the discussion thread,

lucasponce:
@hhovsepy I'd propose to close this PR and re-open it when the investigation is progressed.
We could update the progress in the original issue https://github.com/kiali/kiali/issues/4782

PR #5120, opened 2022-05-26
Improve PF Badging

jshaughn:
Adopt PF Badge styling that is more consistent with OpenShift.  The badge styling is somewhat arbitrary in Kiali, and in practice Openshift styling seems nicer.  Adopting the OS styling:
- is inline with PF4.
- improves badge text legibility.
- is more space-efficient.
- introduces color classification, in addition to text.

Additionally, from a Kiosk perspective, for a Kiali page embedded in OS Console, it provides more conformity.

Some before-after shots:

Before:
image
After:
image

Before:
image
After:
image


jshaughn:
> Not a bad idea but it would trigger to reconsider/review the font sizes.

Well it's interesting because the font size in the badge is intentionally larger than the fonts used for the information around them.  I think it's for clarity.  What you see in the screenshots is the same as what you see in OpenShift.  For example:

image

andrew-ronaldson:
This looks good. I have seen some mockups for newer badges but not sure when that will be implemented.

jshaughn:
> This looks good. I have seen some mockups for newer badges but not sure when that will be implemented.

@andrew-ronaldson Thanks for taking a look.  If console was to update their styling we could follow suit.  One concern I had was for the box labels (see the graph screenshot above).  I wonder if the dark background maybe doesn't work with the new badge backgroundColor, especially for Namespaces.

andrew-ronaldson:
> I wonder if the dark background maybe doesn't work with the new badge backgroundColor, especially for Namespaces.

This has definitely come up especially with dark theme. The new versions work better in this case. I'll check on the status

lucasponce:
My comment about the font sizes is more about the side panel, where the differences with the new badges look strange.

The other fonts look ok but worth checking if the styles are the same, I have the feeling there are some differences that could be good to align (mostly minor style changes).

lucasponce:
The main issue I see is the differences in style/size, especially on the graph side panel.
If we want to update the badges (I'm +1 on that) but we need to align the same style sizing (align to size 14px, perhaps check if there is any other style affecting).

Kiali style:
image

OpenShift style:
image

Perhaps there are more places to review, but those examples are pretty visible and the final effect is not good IMO, it's like we are importing a third-party style but without proper adjusting.

Again, not proposing big refactoring changes, but good to open this effort as most of the migration to PF4 is done.

jshaughn:
Did some more work here:

- change graph box label background from "dark" to "light" theme, as our whole UI is light theme.  This plays better with the light-themed console badge backgrounds.  The new background is also chosen to work with the grayscale graph theming.
    - assign a non-default color to the cluster badge.
    - introduce PFBadge 'size' property.  Can now set size="sm" where applicable
      - now using "sm" for
        - graph side-panel
        - graph context menu
        - workload detail pod card

image

image

lucasponce:
> +1, it needs some review and I see what you are saying about the side panel. Fortunately I think the side panel is failry unique in its use of a smaller font. We would need to look around for similar inconsistencies. I did already add some handling for the graph labels, so hopefully that sizing is correct. I will update the PR with the change for side-panel.

Font sizes in the graph are ok as they can zoom in/out and it's expected that the user can see different sizes.
But we have a set of font size inconsistencies that would be good to align in this effort.

Font 12px in the side panel is not good, also the 16px in the value for lists and some cards, but I think with those couple of changes, the embedded experience will be good.

lucasponce:
> Are you saying you think we should increase the font size throughout the side panel? 

Yes, but not only the side panel, we have some inconsistent use of the font size, not any kind of blocker :), but in the context of the embedded mode is more visible and certainly it should require some minor address.

Font sizes like 12px (side panel) and 16px (list and other components) are not used in OpenShift Console, then when adding a new tab, those changes are pretty visible.

I think those could be reviewed easily, and just fixing this alignment would help.

jshaughn:
> Font sizes like 12px (side panel) and 16px (list and other components) are not used in OpenShift Console, then when adding a new tab, those changes are pretty visible.

@lucasponce , see what you think, I made the use of 14px more consistent, the graph side panel and also the list views:

image

image

lucasponce:
@jshaughn would you mind rebasing this PR with the master branch? Then I could get the latest changes and compare better in embedded mode

lucasponce:
Thanks, @jshaughn, the work on this PR looks good.
I think is close.

The main pending change I see is in the "selectors/dropdown/filters", where Kiali still uses 16px and OpenShift Console uses 14px:
image

Another example where the filters are still in 16px vs the 14px:
image
image

jshaughn:
@lucasponce Another round of updates.  I think the fontSize should now be consistent, with the execption of some places where I'm pretty sure we want it to be different (like log text).  Spent time fixing up labels, especially in the list views.  And fixed a regression in label filtering.

lucasponce:
Navigation menu is moved to font 14px, but it should keep in 16px:
image
image

lucasponce:
Font size 14px in the controls look good, i.e. "Actions" menu looks similar to the Graph controls in the embedded mode:
image

lucasponce:
App/Workload/Service/Span filters look good:
image

If I'd have some comments, I'd say that it would require some minor margin/padding adjustment in both situation (no filters, and filters):
image

lucasponce:
Tabs have moved to font 14px, but for the App/Workload/Service and other I think they can keep in font 16px:
image

image

Then, it would work better both in standalone and embedded:
image

As it would move from 18px -> 16px in the tabs rows, vs 18px -> 14px

lucasponce:
I don't see more aspects to address in this work, maybe there are some minor margin/padding in the content of the cards, but I think that iterating on those last comments would be enough to mark this PR as stable and do the final checks to consolidate.

jshaughn:
OK, my brush stroke was to broad.  I'll have to go back and be more precise.  Thanks for the continued review, you have a good eye for these things.

jshaughn:
@lucasponce Another round of change given your feedback.  Also rebased and squashed.  I think the PR could probably be promoted from Draft.  I'm going to add some cypress testing to cover the regression I found (and fixed) in the area of label filtering.

lucasponce:
I haven't seen this before, when the window is in a narrow width, the namespace filter is collapse to an icon that doesn't work well:
image

lucasponce:
Minor change, but I think I can't avoid fixing this:
image

The 16px left padding when the filters have values could be turned to 0px and be left-aligned for clarity

lucasponce:
Good work @jshaughn, thanks for addressing all the comments.
I've moved to Ready and I'll approve.

lucasponce (approved):
A lot of minor - but important - styling changes will help to harden the PF4 update.

lucasponce:
@jshaughn If you are ok, I'd like to consolidate this in master; we could address any follow-up work; let me know I wasn't sure if you have some pending WIP that you'd like to add.

Issue #5097, opened 2022-05-19
Fix "info" icons in the yaml config editor


jshaughn:
After the Patternfly update we lost the "Info" icons in the yaml editor.  We need to fix this regression.

![image](https://user-images.githubusercontent.com/1662329/166652383-2560e2d5-2be3-42b1-b986-667dafa14d0b.png)

josunect:
@jshaughn , I think this is because in firefox had a different style:
![Screenshot from 2022-05-25 11-20-55](https://user-images.githubusercontent.com/49480155/170247444-ecdf938f-c02a-48f3-b684-43982d449660.png)

I've changed to show in Firefox and Chrome: 

![Screenshot from 2022-05-25 11-56-12](https://user-images.githubusercontent.com/49480155/170247557-a21d5fdb-1b29-4ed7-a1e7-02ce48708f78.png)

If that looks good, I will create the PR. 

Regards,

lucasponce:
@leandroberetta can help on this as he worked on that feature.
He can guide you if you'd need more details about the design.

josunect:
Thanks, @lucasponce , I'll ask him.

josunect:
The original icons, thanks @leandroberetta 

![image](https://user-images.githubusercontent.com/49480155/170486845-61f42072-8a64-4f0d-8176-7e817f135596.png)

PR #5121, opened 2022-05-26
Update info icon in the yaml config editor

josunect:
Update info icon in the yaml config editor, regression after Patternfly update 
It looks like the icon code changed. 

Fixes https://github.com/kiali/kiali/issues/5097


josunect:
cc @leandroberetta

PR #5124, opened 2022-05-26
Update confirmation button placement and styling

jshaughn:
Epic #4260 

The initial PF update left the confirmation box buttons in an intermediate
state.  PF styling moved the buttons from bottom-right to bottom-left,
and as such the button order needed to change such that "Cancel" is the
inner, not the leading, button.  Furthermore PF styling calls for the
"Cancel" button to be the "link" variant.

See https://www.patternfly.org/v4/components/modal/design-guidelines/#confirmation-dialogs

I think I found all of the instances for the change.  Here is one example:

Before:
image
After:
image

### DIFF
--- a/frontend/src/app/InitializingScreen.tsx
+++ b/frontend/src/app/InitializingScreen.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Alert, Button } from '@patternfly/react-core';
+import { Alert, Button, ButtonVariant } from '@patternfly/react-core';
 import { style } from 'typestyle';
 import { isKioskMode } from '../utils/SearchParamUtils';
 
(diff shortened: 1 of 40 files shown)


lucasponce:
I'm not keen about moving the "Cancel" button to a link, even if PF4 applies, I can't see that pattern in other OpenShift features, where the "Cancel" button is still there.

So, I'm -1 to perform that change, which would introduce more different styles
image
image
image
image

jshaughn:
I'm fine either way, we just can't leave it as-is.  Let's ask @andrew-ronaldson for an opinion...

lucasponce:
Thanks, I'm giving a local test and I'll merge if no blockers.

lucasponce (approved):
Tested locally.

PR #5126, opened 2022-05-27
Show Traffic dropdown only in mode when traffic graph is shown.

hhovsepy:
Follow up change request: https://github.com/kiali/kiali/pull/5117#issuecomment-1139601934


jshaughn (commented):
LG but I'm not sure whether our policy is to hide or disable in this scenario.  @lucasponce , do you have an opinion?

hhovsepy:
For Overview Toolbar sort it is hidden.

jshaughn (approved):

Issue #5129, opened 2022-05-30
Review conditional rendering in the kiosk mode


lucasponce:
Kiali can be embedded to be used and integrated into other platforms [1].

This mechanism is implemented using conditional style classes when Kiali is in an embedded state.

Some pages (i.e. details pages) may not render some common elements when embedded state.

The scope of this issue is to introduce this conditional logic in the render phase of the details pages as a first step to enhance the rendering in embedded scenarios.

[1] https://github.com/kiali/kiali#embedding-kiali

PR #5130, opened 2022-05-30
Add a standalone/embedded global state from kiosk variable

lucasponce:
Fixes https://github.com/kiali/kiali/issues/5129

PR for the following tasks:
- Add the "kiosk" variable into the global state.
- Use the global state for conditional rendering on the detail pages.
- Adjust the height calculation on embedded mode for the detail pages.
- Make these changes transparent to the standalone mode.

This work can be visualized in the Workload and Services where there are integration points with Kiali:
image

image

Note that the time range, time selector, and actions controllers are hidden on purpose in the context of this PR.

It's part of the iterations with @andrew-ronaldson to discuss how to visualize those controllers in the embedded mode.


jmazzitelli (approved):
Looks good within the console plugin:
image

leandroberetta (dismissed):
This looks good.

jshaughn (commented):
It seems OK.  Even as a first iteration I wonder whether it should be a little more robust, anyway, see the comments.

lucasponce:
I've updated the changes with @jshaughn suggestion, but the integration to test shouldn't change:
image

The conditional rendering logic is open, as soon as we identify more use cases, these changes could be reverted and added in a different location, but I think it's good to start adding them into the code.

lucasponce:
I've updated the changes with @jshaughn suggestion, but the integration to test shouldn't change:
image

The conditional rendering logic is open, as soon as we identify more use cases, these changes could be reverted and added in a different location, but I think it's good to start adding them into the code.

jshaughn (approved):
LGTM

PR #5133, opened 2022-05-30
OpenShift auth: Allow sessions from 3rd parties

### LINKED ISSUE #5125: Fix find/hide toolbar alignment issues
Epic #4260 

The recent Patternfly update left graph find/hide with a minor alignment issue (at least on chrome, not sure about other browsers).  See the minor offset with the input boxes and the dropdowns:

image

### LINKED ISSUE #5127: Add more mechanisms to provide OpenShift tokens to Kiali
One of the authentication mechanisms supported by Kiali is OpenShift [1].

In the context of the integration of Kiali with other OpenShift platforms, it may happen that the user is already authenticated and a valid token is already stored in the browser.

Kiali could extend the mechanism that this token can be received (HTTP headers or web parameters) to support more rich integration scenarios.

[1] https://kiali.io/docs/configuration/authentication/openshift/

israel-hdez:
The third party must modify every request and inject an OpenShift token via the HTTP `Authorization` Header (preferred) or via the URL `oauth_token` parameter.

Fixes #5127.

### DIFF
--- a/business/authentication/openshift_auth_controller.go
+++ b/business/authentication/openshift_auth_controller.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"k8s.io/client-go/tools/clientcmd/api"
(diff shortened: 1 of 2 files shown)


israel-hdez:
> Then, with a session token, a "bookmarked" URL like this, should also be authenticated and not require the redirect to the login page:

That would require the UI taking an active role on each request.... I think it is possible. Let me try it.

israel-hdez:
@lucasponce It should now work as desired.

lucasponce (approved):
Thanks, @israel-hdez.
Tested standalone and embedded mode.
It works as expected.

Issue #5131, opened 2022-05-30
"This subset's labels are not found in any matching host" - DestinationRule and ServiceEntry


artursitarski:
Kiali configs validator shows `This subset's labels are not found in any matching host` error for DestinationRule in combination with ServiceEntry and VirtualService:

<img width="1134" alt="Screenshot 2022-05-30 at 14 00 32" src="https://user-images.githubusercontent.com/2012391/170988589-02d3fdeb-d3c0-4c78-8128-83354e3baede.png">

ServiceEntry:
```
apiVersion: networking.istio.io/v1beta1
kind: ServiceEntry
metadata:
  name: app-rpc.app.cluster-1a
  namespace: app
spec:
  endpoints:
  - address: XXX.us-east-1.elb.amazonaws.com
    labels:
      cluster: cluster-1a
    ports:
      data: 443
  hosts:
  - app-rpc.app.svc.cluster-1a.global
  location: MESH_INTERNAL
  ports:
  - name: data
    number: 8765
    protocol: TCP
  resolution: DNS
```

VirtualService:
```
apiVersion: networking.istio.io/v1beta1
kind: VirtualService
metadata:
  name: app-rpc-icc
  namespace: app
spec:
  hosts:
  - app-rpc
  - app-rpc.app
  gateways:
  - app/ingress-app
  - mesh
  http:
  - route:
    - destination:
        host: app-rpc.app.svc.cluster-1b.global
        subset: cluster-1b
      weight: 0
    - destination:
        host: app-rpc.app.svc.cluster-1a.global
        subset: cluster-1a
      weight: 0
    - destination:
        host: app-rpc.app.svc.cluster.local
      weight: 100
```

DestinationRule:
```
apiVersion: networking.istio.io/v1beta1
kind: DestinationRule
metadata:
  name: app-rpc.app.cluster-1a
  namespace: app
spec:
  host: app-rpc.app.svc.cluster-1a.global
  subsets:
  - labels:
      cluster: cluster-1a
    name: cluster-1a
  trafficPolicy:
    tls:
      mode: SIMPLE
      sni: app-rpc.app
```

Gateway:
```
apiVersion: networking.istio.io/v1alpha3
kind: Gateway
metadata:
  name: ingress-app
  namespace: app
spec:
  selector:
    app: istio-ingressgateway
  servers:
  - hosts:
    - app/app-rpc.app
    port:
      name: grpc
      number: 443
      protocol: HTTPS
    tls:
[...]

PR #5134, opened 2022-05-31
Added support of considering ServiceEntry endpoint labels in DestinationRule subset labels validation

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5131

DestinationRule subset labels has been checked to match workload labels.
In case when no Workload/Service exists, the ServiceEntry could be created which points to external resource and match the DestinationRule subset labels.
With this PR, the validation of DR subset labels is extended to consider ServiceEntry endpoint labels when no Workload is found with labels.

Screenshot from 2022-05-31 15-15-22

### DIFF
--- a/business/checkers/destinationrules/no_dest_checker.go
+++ b/business/checkers/destinationrules/no_dest_checker.go
@@ -16,7 +16,7 @@ type NoDestinationChecker struct {
 	WorkloadsPerNamespace map[string]models.WorkloadList
 	DestinationRule       networking_v1beta1.DestinationRule
 	VirtualServices       []networking_v1beta1.VirtualService
-	ServiceEntries        map[string][]string
+	ServiceEntries        []networking_v1beta1.ServiceEntry
 	RegistryServices      []*kubernetes.RegistryService
 }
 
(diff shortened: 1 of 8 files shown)


business/checkers/destinationrules/no_dest_checker.go
```diff
@@ -77,21 +77,29 @@ func (n NoDestinationChecker) hasMatchingWorkload(host kubernetes.Host, subsetLa
...
-
 	subsetLabelSet := labels.Set(subsetLabels)
 	subsetSelector := labels.SelectorFromSet(subsetLabelSet)
 
-	for _, wl := range n.WorkloadsPerNamespace[localNs].Workloads {
-		wlLabelSet := labels.Set(wl.Labels)
-		if selector.Matches(wlLabelSet) {
-			if subsetSelector.Matches(wlLabelSet) {
-				return true
+	// Check workloads
```
lucasponce, line 83: The code looks good.
About naming, perhaps could be good to rename the upper meeting "hasMatchingWorkload" to "hasMatchingEndpoint" in the meaning of Istio:
image

It's just to not forget that the DR could point to any service and not necessarily a workload (we don't have that supported in other places, like the wizards but it can be a start).

lucasponce:
I've set up an example with ServiceEntry, DestinationRule and VirtualService pointing, and validations work as expected:
image

lucasponce (approved):
LGTM
Thanks for adding the tests!

PR #5135, opened 2022-06-01
Added istio port-forward to run-kiali script

josunect:
Added istio port forward to run kiali script when istio_url is passed as a parameter. 

Changes from https://github.com/kiali/kiali/pull/5061/files#diff-4835d54dd5ef3a768c44727c31447cdc82e8fe4cb7be22858493b38ca2d239c4

For more details, see https://github.com/kiali/kiali.io/pull/537#issuecomment-1142047243 

cc @jmazzitelli

### DIFF
--- a/hack/run-kiali.sh
+++ b/hack/run-kiali.sh
@@ -75,7 +75,7 @@ DEFAULT_API_PROXY_PORT="8001"
 DEFAULT_CLIENT_EXE="kubectl"
 DEFAULT_ENABLE_SERVER="true"
 DEFAULT_ISTIO_NAMESPACE="istio-system"
-DEFAULT_ISTIO_URL="http://127.0.0.1:15014/version"
+DEFAULT_ISTIOD_URL="http://127.0.0.1:15014/version"
 DEFAULT_KIALI_CONFIG_TEMPLATE_FILE="${SCRIPT_DIR}/run-kiali-config-template.yaml"
 DEFAULT_KIALI_EXE="${GOPATH:-.}/bin/kiali"
 DEFAULT_KUBE_CONTEXT="kiali-developer"
@@ -155,7 +155,7 @@ Valid options:
       Default: ${DEFAULT_ISTIO_NAMESPACE}
   -iu|--istiod-url
       The URL of the istiod endpoint.
-      Default: ${DEFAULT_ISTIO_URL}
+      Default: ${DEFAULT_ISTIOD_URL}
   -kah|--kubernetes-api-host
       The hostname of the Kubernetes API Endpoint.
       Default: <will be auto-discovered>
@@ -288,9 +288,7 @@ fi
 # Port forward data for Istiod, used for the Istiod URL
 PORT_FORWARD_SERVICE_ISTIOD="service/istiod"
 LOCAL_REMOTE_PORTS_ISTIOD="15014:15014"
-if [ -z "${ISTIOD_URL:-}" ]; then
-  ISTIOD_URL= "${DEFAULT_ISTIO_URL}"
-fi
+ISTIOD_URL="${ISTIOD_URL:-${DEFAULT_ISTIOD_URL}}"
 
 
 # If the user didn't tell us what the Prometheus URL is, try to auto-discover it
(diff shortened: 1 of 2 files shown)


hack/run-kiali.sh
```diff
@@ -96,6 +96,7 @@ while [[ $# -gt 0 ]]; do
     -ce|--client-exe)            CLIENT_EXE="$2";                    shift;shift ;;
     -es|--enable-server)         ENABLE_SERVER="$2";                 shift;shift ;;
     -gu|--grafana-url)           GRAFANA_URL="$2";                   shift;shift ;;
+    -iu|--istiod-url)            ISTIOD_URL="$2";                    shift;shift ;;
     -in|--istio-namespace)       ISTIO_NAMESPACE="$2";               shift;shift ;;
```
jmazzitelli, line 100: Options are in alphabetical order - so swap these around:
```suggestion
    -in|--istio-namespace)       ISTIO_NAMESPACE="$2";               shift;shift ;;
    -iu|--istiod-url)            ISTIOD_URL="$2";                    shift;shift ;;
```

jmazzitelli: Note that you need to add a help description for this new option down on line 155. Something like:
```
  -iu|--istiod-url
      The URL of the istiod endpoint.
      Default: http://127.0.0.1:15014/version
```

Note that the defaults are normally defined via a "DEFAULTS" env var so you don't duplicate the values in multiple areas in the script. So you should consider adding that and then just using it here (`Default: ${DEFAULT_ISTIOD_URL}`) and down below where that default value is set.

jmazzitelli: You also need to add this new setting to the configuration file, otherwise, it isn't really being used. So add this section

```
istio:
  url_service_version: "${ISTIOD_URL}"
```

to this file after the grafana section (to retain alphabetic sorting of the settings): https://github.com/kiali/kiali/blob/master/hack/run-kiali-config-template.yaml#L32

Here's the diff that I just used to test this and it works:

```diff
diff --git a/hack/run-kiali-config-template.yaml b/hack/run-kiali-config-template.yaml
index 4fa6bf601..242d9dedd 100644
--- a/hack/run-kiali-config-template.yaml
+++ b/hack/run-kiali-config-template.yaml
@@ -30,6 +30,9 @@ external_services:
     in_cluster_url: "${GRAFANA_URL}"
     url: "${GRAFANA_URL}"
 
+  istio:
+    url_service_version: "${ISTIOD_URL}"
+
   prometheus:
     url: "${PROMETHEUS_URL}"
```

jmazzitelli (changes requested):
I just ran this and found a few things that need to be fixed. A couple are just minor nitpicks because that's what I do :) (alphabetical sorting - that's a disease I caught from @jshaughn)

But you do need to add the help text and definitely need to add that section to the config file in order to pick up the url.

hack/run-kiali.sh
josunect: Thanks, @jmazzitelli 
I didn't know about that disease ^^
But is pretty helpful, especially when you have so many options.

hack/run-kiali.sh
```diff
@@ -75,6 +75,7 @@ DEFAULT_API_PROXY_PORT="8001"
 DEFAULT_CLIENT_EXE="kubectl"
 DEFAULT_ENABLE_SERVER="true"
 DEFAULT_ISTIO_NAMESPACE="istio-system"
+DEFAULT_ISTIO_URL="http://127.0.0.1:15014/version"
```
jmazzitelli, line 78: I believe you are missing the "D" in this variable ... "ISTIOD", not "ISTIO":
```suggestion
DEFAULT_ISTIOD_URL="http://127.0.0.1:15014/version"
```

hack/run-kiali.sh
```diff
@@ -151,6 +153,9 @@ Valid options:
   -in|--istio-namespace
       The name of the control plane namespace - this is where Istio components are installed.
       Default: ${DEFAULT_ISTIO_NAMESPACE}
+  -iu|--istiod-url
+      The URL of the istiod endpoint.
+      Default: ${DEFAULT_ISTIO_URL}
```
jmazzitelli, line 158:
```suggestion
      Default: ${DEFAULT_ISTIOD_URL}
```

hack/run-kiali.sh
```diff
@@ -280,6 +285,14 @@ else
   infomsg "You are connecting to a (non-OpenShift) Kubernetes cluster"
 fi
 
+# Port forward data for Istiod, used for the Istiod URL
+PORT_FORWARD_SERVICE_ISTIOD="service/istiod"
+LOCAL_REMOTE_PORTS_ISTIOD="15014:15014"
+if [ -z "${ISTIOD_URL:-}" ]; then
+  ISTIOD_URL= "${DEFAULT_ISTIO_URL}"
+fi
```
jmazzitelli, line 293: You could replace this entire if-statement with a simpler single line:

```suggestion
ISTIOD_URL="${ISTIOD_URL:-${DEFAULT_ISTIOD_URL}}"
```

But definitely do change the variable name - needs a "D" in it.

jmazzitelli (changes requested):

PR #5138, opened 2022-06-01
To avoid CVE, force resolution of indirect dependencies to later version

jshaughn:
@nrfox I found what seems to be the gomod equivalent of a yarn 'resolution'.

`go mod edit -replace gopkg.in/yaml.v3=gopkg.in/yaml.v3@v3.0.1`

Additionally, the PR bumps a test dep, which eliminated one of the original, bad, indirect deps


nrfox:
Backend integration tests failed but it doesn't seem related to this so I'm guessing it's a flake.

go.mod
```diff
@@ -75,3 +75,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.1.2 // indirect
 	sigs.k8s.io/yaml v1.2.0 // indirect
 )
+
+replace gopkg.in/yaml.v3 => gopkg.in/yaml.v3 v3.0.1
```
nrfox, line 79: I think we could also use `exclude` here so that future patch bumps are picked up instead of pinning to `v3.0.1` but either way gets the job done.

jshaughn:
> Backend integration tests failed but it doesn't seem related to this so I'm guessing it's a flake.

Hmm, not sure, I'll retry and look into it.

PR #5140, opened 2022-06-01
Remove KIA1003 and KIA1103 checkers

israel-hdez:
These checkers are now obsolete, because Istio is actively ensuring that the offending configurations verified by KIA1003 and KIA1103 cannot make it into the cluster (i.e. the cluster will reject a resource with the offending configs). This means that KIA1003 and KIA1103 are now dead checkers and can be removed safely.

Found while working on #5085.

### DIFF
--- a/business/checkers/sidecars/egress_listener_checker.go
+++ b/business/checkers/sidecars/egress_listener_checker.go
@@ -61,10 +61,7 @@ func (elc EgressHostChecker) validateHost(host string, egrIdx, hostIdx int) ([]*
 	checks := make([]*models.IstioCheck, 0)
 	sns := elc.Sidecar.Namespace
 
-	hostNs, dnsName, valid := getHostComponents(host)
-	if !valid {
-		return append(checks, buildCheck("sidecar.egress.invalidhostformat", egrIdx, hostIdx)), false
-	}
+	hostNs, dnsName := getHostComponents(host)
 
 	// Don't show any validation for common scenarios like */*, ~/* and ./*
 	if (hostNs == "*" || hostNs == "~" || hostNs == ".") && dnsName == "*" {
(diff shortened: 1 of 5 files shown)


hhovsepy:
LGTM, is it still draft?

hhovsepy (approved):
Verified, did some smoke regression testing on VS and Sidecars.

israel-hdez:
OK, after doing some testing, I don't see any side errors.

For the record, this is what happens when trying to create the "bad sample" mentioned in docs for KIA1003:

```
cat <<EOF | kubectl apply -f -
pipe heredoc> apiVersion: networking.istio.io/v1alpha3
kind: Sidecar
metadata:
  name: notmatching
  namespace: bookinfo                                                                                                                                                                                             
spec:
  workloadSelector:
    labels:
      app: reviews
  egress:
  - port:
      number: 3306
      protocol: MYSQL
      name: egressmysql
    captureMode: NONE
    bind: 127.0.0.1
    hosts:
    - "*/mysql.foo.com"
    - "noslashsymbolpresent" # unsupported format
pipe heredoc> EOF
Error from server: error when creating "STDIN": admission webhook "validation.istio.io" denied the request: configuration is invalid: host must be of form namespace/dnsName
```

and this is for KIA1103:

```
cat <<EOF | kubectl apply -f -
pipe heredoc> apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: details
spec:
  hosts:
  - details
pipe heredoc> EOF
Error from server: error when creating "STDIN": admission webhook "validation.istio.io" denied the request: configuration is invalid: http, tcp or tls must be provided in virtual service
```

Thus, the bad resources can no longer be created.

israel-hdez:
Thanks for your review, @hhovsepy .

PR #5154, opened 2022-06-06
Default to podman in CI molecule tests

nrfox:
Molecule tests no longer support the docker driver.

### DIFF
--- a/make/Makefile.molecule.mk
+++ b/make/Makefile.molecule.mk
@@ -90,13 +90,9 @@ MOLECULE_KUBECONFIG ?= ${HOME}/.kube/config
 # Set up some additional things needed in order to run the molecule tests against a minikube installation
 ifeq ($(CLUSTER_TYPE),minikube)
 MOLECULE_MINIKUBE_DIR ?= ${HOME}/.minikube
-ifeq ($(DORP),podman)
 # Mounting each file under .minikube individually. podman 3 has more restrictive volume permissions and the :Z label is required
 # but there are two files which are owned by the qemu user and these cannot be labeled since the current user doesn't own them.
 MOLECULE_MINIKUBE_VOL_ARG ?= $(shell MYVAR=""; while read -r file ; do MYVAR="$${MYVAR} -v $${file}:$${file}:Z" ; done <<< $$(find ${MOLECULE_MINIKUBE_DIR} -type f | grep -v ".iso" | grep -v ".rawdisk") && echo $$MYVAR)
-else
-MOLECULE_MINIKUBE_VOL_ARG ?= -v ${MOLECULE_MINIKUBE_DIR}:${MOLECULE_MINIKUBE_DIR}
-endif
 MOLECULE_MINIKUBE_IP ?= $(shell ${MINIKUBE} ip -p ${MINIKUBE_PROFILE})
 MOLECULE_MINIKUBE_ENV_ARGS ?= --env MOLECULE_MINIKUBE_IP=$(shell echo -n ${MOLECULE_MINIKUBE_IP})
 # if there are no hosts the user wants, explicitly set this to empty string to avoid errors later
@@ -107,7 +103,7 @@ endif
 
 # Set up some additional things needed in order to run the molecule tests against a kind installation
 ifeq ($(CLUSTER_TYPE),kind)
-MOLECULE_KIND_IP ?= $(shell ${DORP} inspect ${KIND_NAME}-control-plane --format "{{ .NetworkSettings.Networks.kind.IPAddress }}")
+MOLECULE_KIND_IP ?= $(shell docker inspect ${KIND_NAME}-control-plane --format "{{ .NetworkSettings.Networks.kind.IPAddress }}")
 MOLECULE_KIND_ENV_ARGS ?= --env MOLECULE_KIND_IP=$(shell echo -n ${MOLECULE_KIND_IP})
 # if there are no hosts the user wants, explicitly set this to empty string to avoid errors later
 ifndef MOLECULE_ADD_HOST_ARGS
@@ -130,19 +126,11 @@ endif
 MOLECULE_WAIT_RETRIES_ARG ?= --env MOLECULE_WAIT_RETRIES=${MOLECULE_WAIT_RETRIES}
 
 .prepare-force-molecule-build:
-ifeq ($(DORP),docker)
-	@$(eval FORCE_MOLECULE_BUILD ?= $(shell docker inspect kiali-molecule:latest > /dev/null 2>&1 || echo "true"))
-else
 	@$(eval FORCE_MOLECULE_BUILD ?= $(shell podman inspect kiali-molecule:latest > /dev/null 2>&1 || echo "true"))
-endif
 
 ## molecule-build: Builds an image to run Molecule without requiring the host to have python/pip installed. If it already exists, and you want to build it again, set env var FORCE_MOLECULE_BUILD to "true".
 molecule-build: .ensure-operator-repo-exists .prepare-force-molecule-build
-ifeq ($(DORP),docker)
-	@if [ "${FORCE_MOLECULE_BUILD}" == "true" ]; then docker build --no-cache -t kiali-molecule:latest ${ROOTDIR}/operator/molecule/docker; else echo "Will not rebuild kiali-molecule image."; fi
-else
 	@if [ "${FORCE_MOLECULE_BUILD}" == "true" ]; then podman build --no-cache -t kiali-molecule:latest ${ROOTDIR}/operator/molecule/docker; else echo "Will not rebuild kiali-molecule image."; fi
-endif
 
 ifndef MOLECULE_ADD_HOST_ARGS
 .prepare-add-host-args: .prepare-cluster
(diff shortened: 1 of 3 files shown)


jmazzitelli:
@nrfox all CI tests pass - does this need a review/merge or should it remain in draft?

nrfox:
I was struggling to get these to run locally in kind. The current CI tests that run per PR do not run the molecule tests. I don't know how to test the molecules other than merging and waiting to see if the nightly run pass/fails which is why I wanted to get these to run locally first.

Picking this back up.

nrfox:
@jmazzitelli was able to get this running locally in kind with some molecule tests passing.

make/Makefile.molecule.mk
```diff
@@ -130,19 +126,11 @@ endif
 MOLECULE_WAIT_RETRIES_ARG ?= --env MOLECULE_WAIT_RETRIES=${MOLECULE_WAIT_RETRIES}
 
 .prepare-force-molecule-build:
-ifeq ($(DORP),docker)
```
jmazzitelli, line 133: Can't I build using docker? This doesn't have anything to do with running molecule. This only affects builds, which may be done on a machine that does not have podman. 

I would keep DORP support for things like this.

make/Makefile.molecule.mk
```diff
@@ -130,19 +126,11 @@ endif
...
-	@$(eval FORCE_MOLECULE_BUILD ?= $(shell docker inspect kiali-molecule:latest > /dev/null 2>&1 || echo "true"))
-else
 	@$(eval FORCE_MOLECULE_BUILD ?= $(shell podman inspect kiali-molecule:latest > /dev/null 2>&1 || echo "true"))
-endif
 
 ## molecule-build: Builds an image to run Molecule without requiring the host to have python/pip installed. If it already exists, and you want to build it again, set env var FORCE_MOLECULE_BUILD to "true".
 molecule-build: .ensure-operator-repo-exists .prepare-force-molecule-build
-ifeq ($(DORP),docker)
```
jmazzitelli, line 141: same as previous comment - building can still be done using docker, no?

make/Makefile.molecule.mk
```diff
@@ -155,18 +143,14 @@ else
 endif
 
 .prepare-molecule-data-volume:
-	$(DORP) volume create molecule-tests-volume
```
jmazzitelli, line 158: same comment - building using docker should be ok

make/Makefile.molecule.mk
```diff
@@ -155,18 +143,14 @@ else
...
+	podman volume create molecule-tests-volume
+	podman create -v molecule-tests-volume:/data --name molecule-volume-helper docker.io/busybox true
+	podman cp "${HELM_CHARTS_REPO}" molecule-volume-helper:/data/helm-charts-repo
+	podman cp "${ROOTDIR}/operator/" molecule-volume-helper:/data/operator
+	podman cp "${MOLECULE_KUBECONFIG}" molecule-volume-helper:/data/kubeconfig
+	podman rm molecule-volume-helper
 
 ## molecule-test: Runs Molecule tests using the Molecule docker image
 molecule-test: .ensure-operator-repo-exists .ensure-operator-helm-chart-exists .prepare-add-host-args molecule-build .prepare-molecule-data-volume .create-operator-pull-secret
-ifeq ($(DORP),docker)
```
jmazzitelli, line 167: are we sure we can't run these using docker? I used to have only docker at one point and this all worked. This was during a time we used the old , now deleted, molecule base images so perhaps that is the difference?

jmazzitelli (changes requested):

make/Makefile.molecule.mk
nrfox: Sure you could build with docker and run with podman... but you'd still need to run with podman and therefore have podman installed on the same system you are building on so why not just build with podman? Part of the motivation for removing the `DORP` option was to make it more clear that you can't specify `DORP=docker` and expect the molecule tests to run with docker since they don't support docker anymore and need to be run with podman.

make/Makefile.molecule.mk
nrfox: Yeah this was part of the fallout from those old images getting deleted. We _have_ to run the tests with podman now since the ansible/molecule tests no longer support docker.

make/Makefile.molecule.mk
jmazzitelli: not if I'm building on one machine and running on another.... I could build the image and push to the other machine.

make/Makefile.molecule.mk
nrfox: When you try to run with docker you get:
Screenshot from 2022-06-23 14-20-03

from: https://github.com/kiali/kiali/runs/7017770872?check_suite_focus=true

make/Makefile.molecule.mk
jmazzitelli: my motivation: I'm trying to figure out if this breaks my mazz basement setup  :)

nrfox: Yeah I guess you could build your dev/test image on one machine that only has docker, push it to quay, then pull it on another machine that has podman and run it there but does anyone do that?

> my motivation: I'm trying to figure out if this breaks my mazz basement setup :)

If mazz basement doesn't have both docker and podman installed then it won't ever be able to run the tests but I guess you could build them on mazz basement.

jmazzitelli: > If mazz basement doesn't have both docker and podman

It does not. I do not have docker.

make/Makefile.molecule.mk
jmazzitelli: OK, I'm looking at all of this a little closer, and I think you can leave the Makefile.molecule.mk file alone. I run the full molecule test suite fine on a machine that ONLY has podman (see the test results from last night: https://github.com/jmazzitelli/kiali-molecule-test-logs/blob/openshift/molecule-tests-2022-06-23_02-00-02/results.log)

So there is no need to remove DORP from the Molecule makefile.

It seems the bulk of the changes are to support KinD. So make those changes (you can require "docker" for the Kind specific things if need be). But don't require the machine to have docker since there is no need to for some use cases (i.e. mine :). Having just podman works fine in my case.

jmazzitelli: conversely, if you NEED podman and you don't have it, you can fail at that point.

nrfox: > So there is no need to remove DORP from the Molecule makefile.

The main motivation was to make it clear that you can't pass `DORP=docker` and expect the molecule tests to pass. We could keep the `DORP` option and just make it so that the molecule tests always run podman even when passing `DORP=docker` but imo that's confusing.

> If you NEED docker and it is missing, then you can fail as is the case for the Kind stuff like:

If you are running the molecule tests in kind (which is how CI runs them), then you need both docker and podman. Because of this, the changes here were two fold: 1. Ensure molecule tests are always run with podman 2. Ensure kind is always started with docker (since podman support is experimental). If you're running the molecule tests against minikube or openshift then you'd only need podman.

> finally, leave the DORP , --dorp options in place - don't delete them. I have a feeling in the future the molecule peeps will add back docker support and we can just re-use it.

I don't think they will :). If you're going to go through the trouble of removing support and breaking people that way, I'd be surprised if they added it back after ripping off the bandaid. But if we're going to leave the DORP options, what should be the behavior then? Either we can ignore `DORP` even when `DORP=docker` is passed or allow `DORP=docker` but then the molecule tests will always fail with that option. The same is true of the kind script since that should always run with docker and will need to ignore `DORP=podman` or fail. In the future we could support rootless podman with kind. I got this to work locally but I had to tweak some kernel settings to do it which doesn't give me confidence in this working for CI.

jmazzitelli: From what I see then, it seems only KinD needs this special environment that requires both podman and docker, correct? (That has to be - because in my case, I don't even have docker and all works well - I am not using KinD). So for the KinD specific script and make code, hardcode docker or podman where it makes sense.

For everything else, use DORP. If DORP must be "podman" (like in the "molecule-run" make target) then simply fail fast.
So...

```
ifeq($(DORP),docker)
   echo "Must use podman" && exit 1
endif
```

```bash
[ "$DORP" == "docker" ] && echo "Must use podman" && exit 1
```

I do NOT need docker and things work for me. I want to keep it that way. And I don't really want to have to remember all the places where I define DORP env var and where I pass -dorp to the scripts in on my different machines to now change it (becuse now in this PR the `-dorp` option is unknown/invalid). I want to keep what I have. Not to mention people out in the community using these things - we don't want to break them (I know there is at least one other person out in the community running these molecule tests).

nrfox: > From what I see then, it seems only KinD needs this special environment that requires both podman and docker, correct? (That has to be - because in my case, I don't even have docker and all works well - I am not using KinD). So for the KinD specific script and make code, hardcode docker or podman where it makes sense.

Yes but I just got rootless podman working with kind on my local environment so you could run kind with either docker or podman and both should work provided you can already run rootless podman on your machine. CI will use docker for kind and podman for the molecule tests because in order to get rootless podman working, you may need to modify some kernel settings and I have strong doubts about getting that working in CI.

> I do NOT need docker and things work for me. I want to keep it that way. And I don't really want to have to remember all the places where I define DORP env var and where I pass -dorp to the scripts in on my different machines to now change it (becuse now in this PR the -dorp option is unknown/invalid). I want to keep what I have. Not to mention people out in the community using these things - we don't want to break them (I know there is at least one other person out in the community running these molecule tests).

That's fine I'll add back in the dorp opt and just have the script fail if you pass docker as you suggested. If other folks out there are passing `dorp=docker` to the molecule tests they are already broken.

nrfox: @jmazzitelli I added back all the dorp settings but now I can no longer run things locally for some reason so I'm not sure if this will work or not in CI. The molecule tests are complaining about the helm chart not being built. I've got no clue why since I can see the helm chart being built in the correct location and it's very difficult to track settings being plumbed through 3 different scripts and makefiles.

nrfox: If the nightly molecule run passes then that's good enough for me personally.

jmazzitelli (approved):
runs on my podman-only machine as well as on my laptop that has both.

PR #5156, opened 2022-06-06
Move to v1/CronJob Api

### LINKED ISSUE #4519: Update beta interfaces for CronJob workloads
As noted in #4042, on Kubernetes clusters with versions 1.21+ clusters Kiali logs emit these warnings:

```
W0524 13:55:34.638105  209600 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
```

It wasn't possible to fix those warnings, because Kiali still needs to support Kubernetes clusters older than v1.21. This should be scheduled to be done on or before August 2022.

israel-hdez:
This is because the v1beta1/CronJob Api is going to be removed in the next Kubernetes release. Also, since Istio 1.14 no longer supports Kubernetes versions prior to v1.20, it no longer makes sense to support the older Api in Kiali.

OpenShift 4.9+ has Kubernetes API 1.22+. So, also on this platform is safe to leave behind the older api.

 This removes warning messages from Kiali logs.

 Fixes #4519.

### DIFF
--- a/business/apps_test.go
+++ b/business/apps_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/stretchr/testify/mock"
 	apps_v1 "k8s.io/api/apps/v1"
 	batch_v1 "k8s.io/api/batch/v1"
-	batch_v1beta1 "k8s.io/api/batch/v1beta1"
 	core_v1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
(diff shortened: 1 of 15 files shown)


lucasponce (changes requested):
This requires more investigation.
Using minikube v1.24.0 with kubernetes v1.20.2 there are multiple errors:
image

Note that these controllers are excluded by default, to enable them you'd need to update the config:

image

israel-hdez:
> Using minikube v1.24.0 with kubernetes v1.20.2 and Istio 1.14 there are multiple errors:

This is expected.

This PR is adding support for Kubernetes versions 1.25+ (going to be released very soon) with the side effect of dropping support for versions prior to 1.21 (non-inclusive), which are already at End-of-Life. This is depicted quite fine here: https://endoflife.date/kubernetes.

This is quite aligned to supported k8s versions in Istio and Maistra, I think.

israel-hdez:
> I wanted to be sure that we won't create a problem.

No worries! It is good to double, triple or nth check to be sure!

jshaughn:
Just to confirm for OpenShift, V4.8 is k8s 1.21, and so it should be fine for upstream.   OSSM 2.2 minimum OpenShift is 4.8, and so the next version, 2.3, would be no lower than 48, more likely 4.9, so also OK.

lucasponce:
Yes, Istio has some outdated platform requirements pointing to Kubernetes 1.20 and that created some noise on my side.
I'll update to 1.21 and keep testing.

PR #5163, opened 2022-06-07
Fix for Envoy filters on Workloads details

### LINKED ISSUE #5161: Envoy filter broken
I've noticed this bug in the filter of the Envoy tabs:

image
image

leandroberetta:
Fixes #5161 

The Cypress test for this scenario is being worked at: #5141.


lucasponce (approved):
It works well, thanks!
image

Issue #5164, opened 2022-06-07
[e2e] TestAuthPolicyPrincipalsError test flaking


nrfox:
It looks like this backend e2e test is flaking. The panic makes it hard to debug what the actual error is here.

See: https://github.com/kiali/kiali/runs/6727104653?check_suite_focus=true for an example.

PR #5165, opened 2022-06-07
Fixed potential flaky tests.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/5164

Loading Istio config by poll.

### DIFF
--- a/tests/integration/tests/config_validations_test.go
+++ b/tests/integration/tests/config_validations_test.go
@@ -82,9 +82,7 @@ func getConfigDetails(namespace, name, configType string, assert *assert.Asserti
 	}
 	pollErr := wait.Poll(time.Second, time.Minute, func() (bool, error) {
 		config, _, err := utils.IstioConfigDetails(namespace, name, configType)
-		assert.Nil(err)
-		assert.NotNil(config)
-		if config.IstioValidation != nil {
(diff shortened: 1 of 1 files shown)


tests/integration/tests/config_validations_test.go
```diff
@@ -63,11 +66,29 @@ func TestServiceEntryLabelsNotMatch(t *testing.T) {
...
+
+func getConfigDetails(namespace, name, configType string, assert *assert.Assertions) (*models.IstioConfigDetails, error) {
+	config, _, err := utils.IstioConfigDetails(namespace, name, configType)
+	if err == nil && config != nil && config.IstioValidation != nil {
+		return config, nil
+	}
+	pollErr := wait.Poll(time.Second, time.Minute, func() (bool, error) {
+		config, _, err := utils.IstioConfigDetails(namespace, name, configType)
+		assert.Nil(err)
+		assert.NotNil(config)
```
nrfox, line 86: I think including these here would still cause the test to fail even if the polling returned successfully eventually. If we got a 404 back then `err != nil`. We want to poll until the object exists right?

nrfox (approved):
:+1: thanks for the quick fix!

hhovsepy:
thanks for the review, merging

Issue #5160, opened 2022-06-07
Adjust CustomTooltip in charts


lucasponce:
After the migration to PF4 the "CustomTooltip" used in charts seems displaced and the "arrow/callout" point a little bit far from the point that it represents:

![image](https://user-images.githubusercontent.com/1662329/172318673-8f67bdff-4acc-47f9-9f0f-a2d758598b7b.png)
![image](https://user-images.githubusercontent.com/1662329/172318710-7c8adc24-838d-4573-899e-42ac7578b068.png)

The "CustomTooltip" is using some position coordinates that need some review and adjustment.

PR #5174, opened 2022-06-08
Update customTooltip to place tooltip in the correct point in the graph

josunect:
The property constraintToVisibleArea seems to collide somehow with the with and heigh modified by the tooltip. 

https://formidable.com/open-source/victory/docs/victory-tooltip/#constraintovisiblearea

But, maybe it is expected to render the tooltip just inside the graph?

Fixes https://github.com/kiali/kiali/issues/5160


lucasponce (changes requested):
To me, it works well on these metrics cases:
image
image

I don't see a problem letting the tooltip render out of the chart box, as it's clear and it doesn't introduce noise.

I see it would require to update this change also in the "traces" component:
image

As in that component, there are some tooltips that don't point to the proper data.

Thanks for working on this!

Issue #4701, opened 2022-02-02
Failing to Display Larger Number of Log Lines


s7an-it:
Requirement:
We use Kiali as a log correlation tool in our development stack aside the regular kiali use cases. I know this is not the main feature of Kiali and is not fully supported but it what our developers preference to be using it compared to other tools like datadog/kibana.

Installed with kiali-operator helm chart.
kiali version: 1.44
istio version: 1.11.4 (installed through helm manifests)
EKS 1.20.7

When enabling trace in logs last thing I see before freeze or before I refresh whole browser is similar to, happens to any container:
2022-02-02T11:27:06Z DBG Skipping empty log line [2022-02-02T11:22:44.708153712Z ]
2022-02-02T11:27:06Z DBG Skipping empty log line [2022-02-02T11:22:44.708676239Z ]
2022-02-02T11:27:06Z DBG Skipping empty log line [2022-02-02T11:22:44.718115411Z ]
2022-02-02T11:27:06Z DBG Skipping empty log line [2022-02-02T11:22:44.718534737Z ]
2022-02-02T11:27:07Z DBG Skipping unexpected log line []

Symptom:
The problem we generate hunder of millions of events daily so the default 10MB which is equal to 20k lines in our log format was not enough for us. Kiali has no issues handling this, it may be a bit slower on 20k lines, I have to wit 10-20 sec but it eventually fetch it same when copy, wait a bit and display.
Now since default rotation gives is 1 min of log in our case when we troubleshoot is useless, even if we have injest is kibana if we want to use kiali to correlate traces and envoy data we have to modify logdriver of docker daemon, so eventually we have set 300MB logs on contaner level.

However as soon as log becomes larger than 20k (I only managed to display 25k once), everything in kiali freezes (no error in console, no error in trace aside from some empty log line messsage), ui freeze and whole kiali freezes/dies, once it got OOM killed on 1-2 Gi, but eventually I tried higher resources but it looks when this happens it doesn't use anything that much, maybe at some point it leaked but the usual after I increase over the leak limit I get something like 300MB and 50m usage... Ultimately tried qps/burst 2000/2000, no gzip compression, but nothing seems to influence it.
When I process from kubectl logs I get 500k lines response in 4 seconds and all is good. It looks like something is breaking the buffer maybe in Kiali?

lucasponce:
Thanks @s7an-it for creating the issue

jshaughn:
[...]

PR #5178, opened 2022-06-08
Reduce memory usage for the pod logs endpoint

israel-hdez:
The pod logs endpoint was loading to memory the whole results from the cluster API, then parsing/processing them (also on memory) and, finally sending the results to the browser/client. While this works OK for small data sets, this is leading to high memory usage for large logs.

This is changing the how the input from the cluster API is read and how the results are sent:
1. There is the specific case when the users chooses a time range specifying both a start and end date AND also chooses to show only the last _N_ lines. For this case, the original code is changed only slightly:
    * Instead of loading the whole response of the cluster API into memory to thereafter process it, the response from the cluster API is read as processing is possible. So, only one unprocessed/raw log line is loaded at memory.
    * Instead of loading into memory an undetermined amount of processed log lines to thereafter pick the requested tail, a circular list is used to discard the unneeded lines as soon as possible, keeping memory usage constrained to the amount needed for the requested tail lines.
2. For the other cases that the front-end allows, it is possible to read a line from the cluster API, process it, and immediately send the results to the client/browser, keeping memory usage to a minimum.

---
Results:

**Notice:** These are dumb tests. I simply changed parameters until I had large results. I wasn't counting how many lines are coming from the cluster, nor how many resulting lines are coming from Kiali.

At past state, after a few runs this is Kiali's memory usage:
image

So, Kiali uses 438MiB (in the terminal) when the returned response from Kiali is about 44MiB (seen in browser's developer console, _Size_ column). Also, in the developer console can be seen that the browser spends most of its time "waiting" for a response rather than receiving data.

With these changes applied, this is for case (2):

image

It can be seen that memory usage is down to 46MiB. In the browser console, it is possible to see that Kiali is _not_ faster nor slower, but it can be seen that the browser is now spending more time receiving data than waiting. This clearly shows that Kiali is now "streaming" data to the browser as is being processed.

For case (1) I don't have screenshots, but similarly memory usage remains at the ~50MiB mark even when using the maximum amount of "tail lines" allowed from the front-end (which is 5,000).

---

It is worth to note that, although the back-end is more optimized regarding memory usage reducing the chances of being killed by the cluster, the front-end cannot handle these big responses because it freezes on rendering. I didn't wait until the UI unfroze to know how much time it takes. I'll deal with the front-end on another PR.

---

There is also a bug-fix: When specifying both a start and end date, the shown logs did not belong to the chosen time window under some circumstances.

---
Related to #4701

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -445,7 +445,7 @@ func (in *WorkloadService) BuildLogOptionsCriteria(container, duration, isProxy,
 				opts.MaxLines = &numLines
 			}
 		} else {
-			return nil, fmt.Errorf("Invalid max-lines [%s]: %v", maxLines, err)
+			return nil, fmt.Errorf("Invalid maxLines [%s]: %v", maxLines, err)
 		}
 	}
 
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -67,6 +67,7 @@ type AccessLogEntry struct {
 type LogEntry struct {
 	Message       string            `json:"message,omitempty"`
 	Severity      string            `json:"severity,omitempty"`
+	OriginalTime  time.Time         `json:"-"`
 	Timestamp     string            `json:"timestamp,omitempty"`
 	TimestampUnix int64             `json:"timestampUnix,omitempty"`
 	AccessLog     *parser.AccessLog `json:"accessLog,omitempty"`
@@ -449,6 +450,89 @@ func (in *WorkloadService) BuildLogOptionsCriteria(container, duration, isProxy,
 	return opts, nil
 }
 
+func parseLogLine(line string, isProxy bool, engardeParser *parser.Parser) *LogEntry {
+	entry := LogEntry{
+		Message:       "",
+		Timestamp:     "",
+		TimestampUnix: 0,
+		Severity:      "INFO",
+	}
+
+	splitted := strings.SplitN(line, " ", 2)
+	if len(splitted) != 2 {
+		log.Debugf("Skipping unexpected log line [%s]", line)
+		return nil
+	}
+
+	// k8s promises RFC3339 or RFC3339Nano timestamp, ensure RFC3339
+	splittedTimestamp := strings.Split(splitted[0], ".")
+	if len(splittedTimestamp) == 1 {
+		entry.Timestamp = splittedTimestamp[0]
+	} else {
+		entry.Timestamp = fmt.Sprintf("%sZ", splittedTimestamp[0])
+	}
+
+	entry.Message = strings.TrimSpace(splitted[1])
+	if entry.Message == "" {
+		log.Debugf("Skipping empty log line [%s]", line)
+		return nil
+	}
+
+	// If we are past the requested time window then stop processing
+	parsedTimestamp, err := time.Parse(time.RFC3339, entry.Timestamp)
+	entry.OriginalTime = parsedTimestamp
+	if err != nil {
+		log.Debugf("Failed to parse log timestamp (skipping) [%s], %s", entry.Timestamp, err.Error())
+		return nil
+	}
+
+	severity := severityRegexp.FindString(line)
+	if severity != "" {
+		entry.Severity = strings.ToUpper(severity)
+	}
+
+	// If this is an istio access log, then parse it out. Prefer the access log time over the k8s time
+	// as it is the actual time as opposed to the k8s store time.
+	if isProxy {
+		al, err := engardeParser.Parse(entry.Message)
+		// engardeParser.Parse will not throw errors even if no fields
+		// were parsed out. Checking here that some fields were actually
+		// set before setting the AccessLog to an empty object. See issue #4346.
+		if err != nil || isAccessLogEmpty(al) {
+			if err != nil {
+				log.Debugf("AccessLog parse failure: %s", err.Error())
+			}
+			// try to parse out the time manually
+			tokens := strings.SplitN(entry.Message, " ", 2)
+			timestampToken := strings.Trim(tokens[0], "[]")
+			t, err := time.Parse(time.RFC3339, timestampToken)
+			if err == nil {
+				parsedTimestamp = t
+			}
+		} else {
+			entry.AccessLog = al
+			t, err := time.Parse(time.RFC3339, al.Timestamp)
+			if err == nil {
+				parsedTimestamp = t
+			}
+
+			// clear accessLog fields we don't need in the returned JSON
+			entry.AccessLog.MixerStatus = ""
+			entry.AccessLog.OriginalMessage = ""
+			entry.AccessLog.ParseError = ""
+		}
+	}
+
+	// override the timestamp with a simpler format
+	timestamp := fmt.Sprintf("%d-%02d-%02d %02d:%02d:%02d",
+		parsedTimestamp.Year(), parsedTimestamp.Month(), parsedTimestamp.Day(),
+		parsedTimestamp.Hour(), parsedTimestamp.Minute(), parsedTimestamp.Second())
+	entry.Timestamp = timestamp
+	entry.TimestampUnix = parsedTimestamp.Unix()
+
+	return &entry
+}
+
 // getParsedLogs covers the case when opts has Duration *and* TailLines set.
 func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOptions) (*PodLog, error) {
 	k8sOpts := opts.PodLogOptions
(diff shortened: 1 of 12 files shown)


business/workloads.go
```diff
@@ -445,35 +449,53 @@ func (in *WorkloadService) BuildLogOptionsCriteria(container, duration, isProxy,
...
+			log.Errorf("Error when closing the connection streaming logs of a pod: %s", e.Error())
+		}
+	}()
+
+	bufferedReader := bufio.NewReader(logsReader)
+
+	// It is known in advance that "tailLines" is the maximum number of
+	// log entries that will be returned. So, it is possible to pre-allocate
+	// memory. To keep memory usage constrained, the "entries" will be used
+	// as a circular list: once it is full, we "rewind" and start overwriting
```
jshaughn, line 478: Good idea.

business/workloads.go
```diff
@@ -508,18 +530,16 @@ func (in *WorkloadService) getParsedLogs(namespace, name string, opts *LogOption
...
 
-				if parsedTimestamp.After(*endTime) {
-					break
-				}
+			if parsedTimestamp.After(*endTime) {
+				break
 			}
 		} else {
-			log.Debugf("Failed to parse log timestamp (skipping) [%s], %s", entry.Timestamp, err.Error())
+			log.Errorf("Failed to parse log timestamp (skipping) [%s], %s", entry.Timestamp, err.Error())
```
jshaughn, line 542: Not sure this is a good idea, I think I'd leave this at Debug.

business/workloads.go
```diff
@@ -1852,3 +1870,191 @@ func (in *WorkloadService) GetWorkloadAppName(ctx context.Context, namespace, wo
 	app := wkd.Labels[appLabelName]
 	return app, nil
 }
+
+// streamParsedLogs only works correctly when opts does not have Duration set || does not have TailLines set.
```
jshaughn, line 1874:
```suggestion
// streamParsedLogs covers the case when opts does not have both Duration and TailLines set.
```

business/workloads.go
```diff
@@ -1852,3 +1870,191 @@ func (in *WorkloadService) GetWorkloadAppName(ctx context.Context, namespace, wo
...
+
+	// To avoid high memory usage, the JSON will be written
+	// to the HTTP Response as it's received from the cluster API.
+	// That is, each log line is parsed, decorated with Kiali's metadata,
+	// marshalled to JSON and immediately written to the HTTP Response.
+	// This means that it is needed to push HTTP headers and start writing
+	// the response body right now and any errors at the middle of the log
+	// processing can no longer be informed to the client. So, starting
+	// these lines, the best we can do if some error happens is to simply
+	// log the error and strop/truncate the response, which will have the
```
jshaughn, line 1915:
```suggestion
	// log the error and stop/truncate the response, which will have the
```

business/workloads.go
```diff
@@ -1852,3 +1870,191 @@ func (in *WorkloadService) GetWorkloadAppName(ctx context.Context, namespace, wo
...
+	// effect of sending an incomplete JSON document that the browser will fail
+	// to parse. Hopefully, the client/UI can catch the parsing error and
+	// properly show an error message about that there was a failure retrieving logs.
```
jshaughn, line 1918:
```suggestion
	// properly show an error message about the failure retrieving logs.
```

business/workloads.go
```diff
@@ -1852,3 +1870,191 @@ func (in *WorkloadService) GetWorkloadAppName(ctx context.Context, namespace, wo
...
+	w.Header().Set("Content-Type", "application/json")
+	_, writeErr := w.Write([]byte("{\"entries\":[")) // This starts the JSON document
+	if writeErr != nil {
+		return writeErr
+	}
+
+	firstEntry := true
+	line, readErr := bufferedReader.ReadString('\n')
+	for ; readErr == nil || (readErr == io.EOF && len(line) > 0); line, readErr = bufferedReader.ReadString('\n') {
+		entry := LogEntry{
```
jshaughn, line 1928: I think probably a lot of the duplicated code for processing the individual log entries could be consolidated into a single function, leaving `streamParsedLogs` and `getParsedLogs` to more clearly focus on the different retrieval approaches.

jshaughn (commented):
@israel-hdez Definitely looks like a promising and needed enhancement.  A few comments for your consideration.

business/workloads.go
israel-hdez: I think it's a leftover from a debugging session. I'll revert.

business/workloads.go
israel-hdez: I agree. I'll check how to de-duplicate the code.

israel-hdez: OK, I extracted some code into a function.
I think it is still some room for code de-duplication, but that would require more of a refactoring.

I hope it is better.

business/workloads.go
```diff
@@ -445,35 +449,53 @@ func (in *WorkloadService) BuildLogOptionsCriteria(container, duration, isProxy,
...
+	defer func() {
+		e := logsReader.Close()
+		if e != nil {
+			log.Errorf("Error when closing the connection streaming logs of a pod: %s", e.Error())
+		}
+	}()
+
+	bufferedReader := bufio.NewReader(logsReader)
+
+	// It is known in advance that "tailLines" is the maximum number of
```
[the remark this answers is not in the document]
jshaughn: Hmmm, interesting thought.  I'm not sure.  For situations where there is heavy logging, even a short time window can result in a large number of log entries, and tailLines prevents unnecessary lines from being sent back to the client.

If we remove tailLines we could always perform streaming, right?  If we did that and added maxLines (where ML is maybe something easily consumable, like 500 or maybe 1000, and make it configurable?), it may be OK.  But with streaming you would get the first ML lines, not the tailing ML lines, right?  Wouldit make any sense to use the circular buffer of ML size, in conjunction with streaming?

jshaughn (commented):
De-dup of code looks good.  Let's resolve the tailLines conversation before doing testing...

business/workloads.go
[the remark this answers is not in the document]
israel-hdez: > If we remove tailLines we could always perform streaming, right?

Yes! :-)

> If we did that and added maxLines (where ML is maybe something easily consumable, like 500 or maybe 1000, and make it configurable?), it may be OK. But with streaming you would get the first ML lines, not the tailing ML lines, right?

Yes! :-(

About the _maxLines_ I think it depends on who are you thinking by "easily consumable". I opened yesterday #5201 which is a draft-PR for the front-end. Currently, hide/show is broken (I'm working on it and it is almost done). With the virtual scrolling, the browser can easily digest 250k lines of logs (which is half of the size in the tests mentioned in the main comment). However, with 250k lines of logs, clearly a user won't read/inspect it all, but I do think it is nice to have those large results because of the find/hide/regex feature.

If we do want to limit _maxLines_ to a few hundreds, I think find/hide filters will need to move to server side to keep those features valuable. I think the server would return up to _maxLines_ of already filtered log lines rather than stopping after processing _maxLines_ raw logs.

> Wouldit make any sense to use the circular buffer of ML size, in conjunction with streaming?

I think it is possible. I added "streaming" at two points:
1. From cluster API feeding raw logs to Kiali
2. From Kiali feeding the processed logs to the client/browser.

The (1) is happening always, regardless of chosen options and that would already be a good improvement. The (2), sadly, cannot happen when using the circular list (or I can't think a way to do it).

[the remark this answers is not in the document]
israel-hdez: > Currently, hide/show is broken (I'm working on it and it is almost done).

It works again on the PR :-)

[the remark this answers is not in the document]
jshaughn: OK, I understand your point about allowing a larger number of log lines, because Show/Hide is a powerful mechanism that can reduce the set easily, and the browser can deal with a fairly large set.  I'm OK if we want to try your more radical change here, and remove `tailLines`, as part of this attempt to increase our scale.  It would remove the need for two different approaches in the backend.

I'm thinking though, that maybe instead of `maxLines` being a configured limit, that `maxLines` could replace `tailLines` as a client option.  This would give the user control.  The dropdown tooltip could change to indicate it's the *first* N lines.  The entries could be much bigger, like [100, 1000 (default), 10000,  All Lines]. or soemthing like that.

If a user does hit the `maxLines` limit, we may need to indicate it somewhere on-screen, or maybe generate an informational toast

Do you think "Last 1m" is still good for the smallest duration?  Or would it make sense to offer something even smaller by default for logs, like "Last 30s" (note that currently a custom time range has a 1m minimum but that can't easily be changed)?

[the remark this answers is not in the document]
israel-hdez: > I'm OK if we want to try your more radical change here, and remove `tailLines`, as part of this attempt to increase our scale. It would remove the need for two different approaches in the backend.
> 
> I'm thinking though, that maybe instead of `maxLines` being a configured limit, that `maxLines` could replace `tailLines` as a client option.

Good! Let's try it. I'll change this PR.

> Do you think "Last 1m" is still good for the smallest duration? Or would it make sense to offer something even smaller

After removing tailLines, I do think something smaller makes sense. I would even add 15 and 10 secs, which are the smaller auto-refresh intervals. Although, perhaps 30secs as default would be OK.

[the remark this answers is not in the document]
israel-hdez: OK, I did the changes.
The only change that I didn't do is adding the smaller durations. I understand that would affect all pages using the duration drop-down. So, I'm unsure about it.

I used a default of 3K max lines. Looks like at the default duration of 10m, the bookinfo apps reach the 1K limit easily. If this is not OK, I can change it to 1K.

jshaughn (commented):
Tested and seems to be working.  See comments but close to approval.

frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
```diff
@@ -115,16 +115,17 @@ enum LogLevel {
...
-  '1000': '1000 lines',
-  '5000': '5000 lines'
+  '100': 'Max: 100 lines',
+  '500': 'Max: 500 lines',
+  '1000': 'Max: 1,000 lines',
+  '3000': 'Max: 3,000 lines',
+  '5000': 'Max: 5,000 lines',
+  '10000': 'Max: 10,000 lines',
+  '20000': 'Max: 20,000 lines',
+  '50000': 'Max: 50,000 lines',
```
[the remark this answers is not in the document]
israel-hdez: OK, it is changed.

frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
```diff
@@ -929,6 +930,7 @@ export class WorkloadPodLogs extends React.Component<WorkloadPodLogsProps, Workl
           responses.shift();
         }
 
+        let linesSurpassedContainers: string[] = [];
```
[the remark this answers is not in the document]
israel-hdez: I indeed do agree that the toast is sub-optimal for this case. 

I put the warning between the logs pane and the container selection. I'm not sure how to do it better:

image

jshaughn (approved):
This PR seems to be working and the code looks good to me.

Issue #5092, opened 2022-05-19
Update font color on green/red labels for trace details


lucasponce:
After the PF4 update, some styles would need an adjustment, this issue is to track a minor change in the trace details labels:

![image](https://user-images.githubusercontent.com/1662329/169293901-4ddfad07-765c-4ecb-8006-2999bab1ec60.png)
![image](https://user-images.githubusercontent.com/1662329/169293955-4c1d8dae-18c1-47e2-b7b1-54b9a3cb7717.png)

On green/red labels, the font color should have better contrast (white?).

This issue is also filled for new comers :-), as it can be a good one to start adding minor changes.

MThabsheer7:
Hey @lucasponce, can I do this?

lucasponce:
@MThabsheer7 sure, feel free to add a PR ! Thanks

jshaughn:
@MThabsheer7 Is there a PR coming for this?

jshaughn:
@MThabsheer7, you seem busy.  I'm going to reassign this one.  We still hope you can find time for a contribution going forward!

josunect:
It seems pf adds a wrapper span in the label:

https://github.com/patternfly/patternfly-react/issues/2320

PR #5180, opened 2022-06-09
Update font color on green red labels for trace details

josunect:
Update font color on green red labels for trace details

I've though in adding a custom Label style with a new class whiteColorLaberContent with nested css. Because PF4 adds a wrapper in the label content, and apply the style to the label, is not enough. 

Comment: Should we add this into warning? (I will revert if not)

Screenshot from 2022-06-09 09-57-41
Screenshot from 2022-06-09 09-59-14
Screenshot from 2022-06-09 10-00-40

Fixes https://github.com/kiali/kiali/issues/5092


lucasponce (approved):
image
image

There are no "warning" traces as they represent typically error conditions, but it's ok to add the warning logic there.

PR #5155, opened 2022-06-06
Toolbar help and bell icons misaligned

josunect:
Align toolbar icons Bell and Help from Helpdropdown and MessageCenter.

Fixes #5148 

cc @jshaughn

PR #5182, opened 2022-06-09
Toolbar icons missaligment. Adjust IstioStatus alert icon

josunect:
Toolbar icons missaligment. Adjust IstioStatus alert icon, missed to check from merged PR.

Fixes #5155


lucasponce (changes requested):
This one requires an update on the snapshot:

```
yarn test -> update snapshots
```

Then commit to the PR the new snapshot

lucasponce (approved):
image

Issue #5081, opened 2022-05-17
[cypress] Move sleep app to its own namespace


nrfox:
Currently the sleep app is deployed to the `default` namespace by the hack script that deploys the demo apps for setting up an e2e testing environment. This can cause issues on environments that treat the default namespace differently than other namespaces. The sleep app can either be moved to one of the existing testing namespaces such as `bookinfo` or a separate namespace can be created specifically for it. The tests that interact with the sleep app should also be updated.

nrfox:
Fixed in #5188

PR #5188, opened 2022-06-10
Add script for setting up a perf env in IBM Cloud

### LINKED ISSUE #5142: Reproducible performance testing environment
The Kiali project needs a way to create a reproducible perf testing environment so that performance tests can run against a consistent environment.

The end result of this issue should be a script or command that creates the infrastructure necessary to run the perf tests. Upstream Istio should be installed along with Kiali and the testing demo apps.

This script will be used to both create and teardown the infrastructure used for perf testing when the testing is completed.

nrfox:
Adds a script for creating a perf env on OpenShfit in IBM Cloud. Also updates `hack/istio/install-testing-demos.sh` to install the sleep app in its own namespace rather than using the default. This fixes issues with installing sleep app on OpenShift.

To test:
```
./hack/perf-ibmcloud-openshift.sh -akf <your_ibmcloud_api_key file>
```

Fixes #5142 #5081


jmazzitelli:
What's the quickest way to test this?

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,244 @@
...
+HELPMSG
+      exit 1
+      ;;
+    *)
+      echo "Unknown option: [$key]. Aborting. Use '$0 --help' for proper usage."
+      exit 1
+      ;;
+  esac
+done
+
```
jmazzitelli, line 95:
```suggestion

if [ ! -f ${API_KEY_FILE} ]; then
  echo "The API Key file is invalid: ${API_KEY_FILE}"
  exit 1
fi

```

jmazzitelli (changes requested):
I don't like being forced to put a password-like api key on a command line at least without an option to pass it in as a file.

See my suggestions below... I replace the api-key option with api-key-file - the file can be generated by the ibmcloud-openshift.sh hack.

Feel free to retain the -ak|--api-key option if you want by adding it back in, but don't make it the only way to pass in the api key - give the user the option to pass it in within a file.

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,244 @@
...
+DEFAULT_NODE_FLAVOR="bx2.8x32"
+DEFAULT_NODES="3"
+DEFAULT_OPENSHIFT_VERSION="4.10_openshift"
+DEFAULT_REGION="us-south"
+
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case $key in
+    # OPTIONS
+    -ak|--api-key)            API_KEY="$2";            shift;shift ;;
```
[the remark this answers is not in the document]
nrfox: Passing it as a file sounds good.

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,250 @@
...
+
+install_kiali() {
+  infomsg "Installing kiali operator and kiali"
+
+  # cluster-status will expose the cluster's internal image registry route that podman needs to push images to.
+  make -C "${SCRIPT_DIR}/.." -e DORP=podman cluster-status
+  local cluster_repo
+  cluster_repo=$(oc get image.config.openshift.io/cluster -o custom-columns=EXT:.status.externalRegistryHostnames[0] --no-headers 2>/dev/null)
+  local master_url
+  master_url=$(ibmcloud oc cluster get -c kiali-perf-cluster --output json | jq -r '.masterURL')
```
jmazzitelli, line 133: I think we should do this so we don't have N clusters called the same thing created by a bunch of different devs
```suggestion
  master_url=$(ibmcloud oc cluster get -c ${USER}-kiali-perf-cluster --output json | jq -r '.masterURL')
```

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,250 @@
...
+  echo -e "\U0001F6A8 ERROR: ${1}"
+  exit 1
+}
+
+infomsg() {
+  echo -e "\U0001F4C4 ${1}"
+}
+
+install_istio() {
+	hack/istio/download-istio.sh -iv "1.13.0"
```
jmazzitelli, line 24: TAB!
```suggestion
  hack/istio/download-istio.sh -iv "1.13.0"
```

jmazzitelli (changes requested):
remove tab.

Change name of cluster so we know the user who created it

download latest istio version (no hardcoded version)

hack/perf-ibmcloud-openshift.sh
jmazzitelli: finally, do we want this fixed on istio 1.13? 1.14 is out.

Perhaps don't provide any version - the download script will download the latest release, so no need to hardcode a value here.
```suggestion
  hack/istio/download-istio.sh
```

nrfox: > There are also lines with spaces before EOL. Ewwww...

I've become reliant on my IDE to spot/fix these things for me :)

> finally, do we want this fixed on istio 1.13? 1.14 is out.

Yeah good catch. That was just leftover from me testing things.

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,250 @@
...
+      tracing:
+        enabled: true
+EOF
+
+}
+
+REGION="us-south"
+
+install_openshift() {
+  if hack/ibmcloud-openshift.sh status -np kiali-perf | grep -q 'Cluster is deployed'; then
```
jmazzitelli, line 220:
```suggestion
  if hack/ibmcloud-openshift.sh status -np ${USER}-kiali-perf | grep -q 'Cluster is deployed'; then
```

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,250 @@
...
+    infomsg "Openshift cluster is already deployed. Waiting for it to be ready to use..."
+    hack/ibmcloud-openshift.sh finish -r "${REGION}" -np kiali-perf
+    infomsg "Openshift cluster is ready to use."
+    return 0
+  fi
+
+  hack/ibmcloud-openshift.sh create \
+    --name-prefix kiali-perf \
```
jmazzitelli, line 228:
```suggestion
    --name-prefix ${USER}-kiali-perf \
```

jmazzitelli (changes requested):
need to prefix with user everywhere.

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,250 @@
...
+EOF
+
+}
+
+REGION="us-south"
+
+install_openshift() {
+  if hack/ibmcloud-openshift.sh status -np kiali-perf | grep -q 'Cluster is deployed'; then
+    infomsg "Openshift cluster is already deployed. Waiting for it to be ready to use..."
+    hack/ibmcloud-openshift.sh finish -r "${REGION}" -np kiali-perf
```
nrfox, line 222: Need one here too:
```suggestion
    hack/ibmcloud-openshift.sh finish -r "${REGION}" -np ${USER}-kiali-perf
```

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,250 @@
...
+DEFAULT_OPENSHIFT_VERSION="4.10_openshift"
+DEFAULT_REGION="us-south"
+
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case $key in
+    # OPTIONS
+    -akf|--api-key-file)      API_KEY_FILE="$2";       shift;shift ;;
+    -hr|--helm-charts-repo)   HELM_CHARTS_REPO="$2";   shift;shift ;;
+    -iv|--istio-version)      ISTIO_VERSION="$2";      shift;shift ;;
```
jmazzitelli, line 44: ISTIO_VERSION doesn't appear to be used anywhere

jmazzitelli: Even if you pass it to download, the install istio script isn't guaranteed to use it. We don't have a "-iv" option to install-istio-via-istioctl. That will install the highest version of istio you have downloaded (unless you point to a specific one via `--istio-dir`).

So, you could probably just get rid of this option, or tweek the istio install hack script to add a `--istio-version`

hack/perf-ibmcloud-openshift.sh
```diff
@@ -122,6 +117,16 @@ REGION=${REGION}
...
 infomsg "==END CONFIG=="
 
+install_istio() {
+  if [ -z "${ISTIO_VERSION}" ]; then
+    hack/istio/download-istio.sh -iv ${ISTIO_VERSION}
+  else
+    # Download latest
+    hack/istio/download-istio.sh
+  fi
+  hack/istio/install-istio-via-istioctl.sh -iee true -cp openshift
```
jmazzitelli, line 127: Just FYI: this isn't guaranteed to use the version passed in and set in ISTIO_VERSION.
Presumably, this allows you to test with an earlier version of Istio... but if you already had installed 1.14 and you pass in `-iv 1.13` this won't do what you think it will do.

hack/perf-ibmcloud-openshift.sh
nrfox: Alright I'll get rid of it for now and we can add it later if need be.

hack/perf-ibmcloud-openshift.sh
jmazzitelli: https://github.com/kiali/kiali/blob/v1.52.0/hack/istio/install-istio-via-istioctl.sh#L244-L245

jmazzitelli: a hacky workaround would be:

`touch _output/istio-${ISTIO_VERSION}`

but then everytime the user runs install-istio-via-istioctl, it will install THAT version.

I think we should update install-istio-via-istictl so it can take a `--istio-version` option.

jmazzitelli (changes requested):
Now that -iv is a valid option to the istio install script, use it.

Also, I noticed a bad line in here -- need to wrap the file name in quotes, otherwise the `if [ -f ...]` test was broken. I fixed that in this diff, as well as adding a test to make sure the file option was provided, since it is required.

this diff is also in PR https://github.com/nrfox/kiali/pull/3

```diff
diff --git a/hack/perf-ibmcloud-openshift.sh b/hack/perf-ibmcloud-openshift.sh
index 143adffa1..e88c9bda6 100755
--- a/hack/perf-ibmcloud-openshift.sh
+++ b/hack/perf-ibmcloud-openshift.sh
@@ -88,7 +88,12 @@ HELPMSG
   esac
 done
 
-if [ ! -f ${API_KEY_FILE} ]; then
+if [ -z "${API_KEY_FILE:-}" ]; then
+  echo "The API Key file is required. Please specify --api-key-file"
+  exit 1
+fi
+
+if [ ! -f "${API_KEY_FILE}" ]; then
   echo "The API Key file is invalid: ${API_KEY_FILE}"
   exit 1
 fi
@@ -118,13 +123,11 @@ EOM
 infomsg "==END CONFIG=="
 
 install_istio() {
-  if [ -z "${ISTIO_VERSION}" ]; then
-    hack/istio/download-istio.sh -iv ${ISTIO_VERSION}
-  else
-    # Download latest
-    hack/istio/download-istio.sh
+  local iv_option=""
+  if [ -n "${ISTIO_VERSION}" ]; then
+    iv_option="-iv ${ISTIO_VERSION}"
   fi
-  hack/istio/install-istio-via-istioctl.sh -iee true -cp openshift
+  hack/istio/install-istio-via-istioctl.sh -iee true -cp openshift ${iv_option}
 }
 
 install_kiali() {
```

jmazzitelli (changes requested):

hack/perf-ibmcloud-openshift.sh
```diff
@@ -0,0 +1,258 @@
...
+  exit 1
+}
+
+infomsg() {
+  echo -e "\U0001F4C4 ${1}"
+}
+
+SCRIPT_DIR=$( dirname -- "$0"; )
+
+DEFAULT_HELM_CHARTS_REPO="${SCRIPT_DIR}/../helm-charts"
```
jmazzitelli, line 25: I think this default is incorrect. Assuming one has a kiali dev environment where you clone all the kiali repos in peer directories, such as:

/source/kiali
/source/kiali-operator
/source/helm-charts

Then this should be:

```suggestion
DEFAULT_HELM_CHARTS_REPO="${SCRIPT_DIR}/../../helm-charts"
```

The SCRIPT_DIR will be `/source/kiali/hack`, thus you need to go up TWO directories (`../..`) in order to get to the helm-chart repo.

jmazzitelli (changes requested):

jmazzitelli (approved):
all looks good on my end

jmazzitelli (approved):
works and I confirm all cypress tests pass on OpenShift 4.10 using anonymous auth strategy.
```
    ✔  All specs passed!                        06:46      174      152        -       22        -  
```

PR #5193, opened 2022-06-13
Disabling (for now) integrations test due to stability issues

leandroberetta:


jmazzitelli (commented):
I'm not comfortable approving this. I would want someone who is familiar with these tests to look at them and confirm these are truly false positives. We don't want to release something that is truly breaking our tests.

lucasponce:
the last run seems progressing https://github.com/kiali/kiali/runs/6864567151?check_suite_focus=true

lucasponce:
Perhaps the action item could be to have a backup of release from a local laptop?
I mean if for some reason the CI is unstable for reasons that we can't control, a manual built and push from any laptop could be a backup?

lucasponce:
Put on hold, the merge, as it seems the latest run passed. So this is not necssary

PR #5199, opened 2022-06-13
Try fixing the cypress failure

### LINKED ISSUE #5198: [CI] Test Flake - Cypress sidecar injection
See: https://github.com/kiali/kiali/runs/6865180576?check_suite_focus=true for an example

jshaughn:
Try fixing the cypress failure by filtering the overview list to a single entry, on the assumption that maybe the kebab menu item is sometimes getting clipped and not in the dom, if the row happens to be too close
to the bottom.

Fixes #5198

### DIFF
--- a/frontend/cypress/integration/featureFiles/sidecar_injection.feature
+++ b/frontend/cypress/integration/featureFiles/sidecar_injection.feature
@@ -11,24 +11,32 @@ Feature: Controlling sidecar injection
 
     Scenario: Override the default policy for automatic sidecar injection by enabling it in a namespace
         Given a namespace without override configuration for automatic sidecar injection
-        When I override the default automatic sidecar injection policy in the namespace to enabled
+        When I visit the overview page
+        And user filters "default" namespace
+        And I override the default automatic sidecar injection policy in the namespace to enabled
         Then I should see the override annotation for sidecar injection in the namespace as "enabled"
 
     Scenario: Switch the override configuration for automatic sidecar injection in a namespace to disabled
         Given a namespace which has override configuration for automatic sidecar injection
         And the override configuration for sidecar injection is "enabled"
-        When I change the override configuration for automatic sidecar injection policy in the namespace to "disable" it
+        When I visit the overview page
+        And user filters "default" namespace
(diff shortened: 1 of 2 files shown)


jshaughn:
@nrfox This works for me repeatedly locally and twice here in CI.  Any thoughts?

frontend/cypress/integration/featureFiles/sidecar_injection.feature
```diff
@@ -11,24 +11,32 @@ Feature: Controlling sidecar injection
 
     Scenario: Override the default policy for automatic sidecar injection by enabling it in a namespace
         Given a namespace without override configuration for automatic sidecar injection
-        When I override the default automatic sidecar injection policy in the namespace to enabled
+        When I visit the overview page
+        And user filters "default" namespace
```
nrfox, line 15: I wonder if the fix here is that this line is waiting for kiali to finish loading before attempting to "click" something in the step definition afterward. Basically anytime the tests visit a page we need to make sure that Kiali is done loading the page before trying to interact with it. Otherwise you can end up clicking dom elements that don't have the listeners attached yet or else clicking things and having the page re-render essentially undoing the previous click if you've opened a menu or something. Maybe adding an `ensureKialiFinishedLoading` inside the existing step definitions after the page visit would've fit in more nicely with the way these feature files are written but we haven't been imposing a particular style across the tests and I'm happy that this fixes the flakes.

nrfox (approved):

frontend/cypress/integration/featureFiles/sidecar_injection.feature
jshaughn: I'm not sure, that may have been a cleaner fix if it works.  Edgar had added instances of `ensureKialiFinishedLoading` in the previous attempt to fix the flakiness, but perhaps we neded them in additional places.  Anyway, since I wasn't looking at that I was tryng to fix my hunch for what the problem may be.  Like you, if the tests pass I'm good.  If we see more failures lets add some more strategic use of `ensureKialiFinishedLoading`.

PR #5200, opened 2022-06-13
be able to install a specific version of istio

jmazzitelli:
To test, download two versions of Istio in order of their release numbers:

```
hack/istio/download-istio.sh -iv 1.13.0
hack/istio/download-istio.sh -iv 1.14.0
```

Now try to install the earlier version:

```
hack/istio/install-istio-via-istioctl.sh -iv 1.13.0
```

The way this used to work, the last version you downloaded would have been installed. So `install-istio-via-istioctl.sh` (without any options) would have installed 1.14. But now you can tell it to install an earlier version, even if you already downloaded another version recently.

Note another test - if you have NOT downloaded the version yet, it will download it for you, then install... so try this to test that that works:

```
hack/istio/install-istio-via-istioctl.sh -iv 1.12.0
```


nrfox (approved):
:+1: works well

PR #5201, opened 2022-06-13
Use virtual scrolling on the workload logs page

israel-hdez:
This is to avoid the UI to freez when a large amount of logs are loaded.

Related back-end PR: #5178 

Related to #4701


nrfox:
Does Kiali already implement similar functionality with the `VirtualList`/`VirtualItem` components?

israel-hdez:
> Does Kiali already implement similar functionality with the `VirtualList`/`VirtualItem` components?

I though it. But I looked at the code and it wasn't clear how it would implement it, as I only saw simple renders. If it somehow is doing it, it is not reflected in cypress tests, because you can select any row without worrying about scrolling.

Probably, `VirtualList` is anyway not very suitable, because its focus is tabular data.

jshaughn:
Don't try to leverage the VirtualList stuff, it's not suitable.

lucasponce:
Side note, but important.
The same feature is already implemented in https://github.com/openshift/console/tree/master/frontend.

Remember that now third-party dependencies with Patternfly and others should be in sync, and this would be even more demanding when we progress in our dynamic plugin efforts.

Before, this effort progresses too much, I'd take a look if we could have a similar implementation, or at least that we don't pull different dependencies that may create a conflict when we need to integrate in the future.

israel-hdez:
@lucasponce I wasn't able to hit the bug you mentioned.
I'm not sure if the changes in this PR may have affected/fixed the bad behavior.

lucasponce:
Thanks for checking it @israel-hdez.
This is what I see, not sure if a bug, but something that looked non-synced, the istio-proxy vs app log
logs-issue

israel-hdez:
> This is what I see, not sure if a bug, but something that looked non-synced, the istio-proxy vs app log

I think it is a consequence about how logs are fetched and what's the timestamp that Kiali attaches to each log entry.

So, we rely on the cluster API to fetch logs of each container and giving the start timestamp. Kubernetes provides a timestamp for each log line. However, specifically for istio-proxy logs, the timestamp is replaced by the one provided by the Istio proxy, which may be a little in the past when compared with the time that kubernetes is recording. So, because of time offsets, I think it is expected that in the "head" logs from the app are missing, while in the "tail" logs from the istio-proxy are missing.

jshaughn:
> the timestamp is replaced by the one provided by the Istio proxy, which may be a little in the past when compared with the time that kubernetes is recording. 

The access-log timestamp is parsed out of the access log, and it reflects the proper time.  The k8s timestamp on the access-log is incorrect, it's the time it was stored to k8s.  The app logs are storing directly to k8s and so the k8s timestamp is correct.  Together, the app and access logs should be intertwined correctly.  I don't think there should be a behavioral change with the unification, and so the old mechanism and new mechanism, I think, should show the same results for the unified logs.  If that's not true we need to understand why there is a change, and whether the change is more or less accurate.

israel-hdez:
> I don't think there should be a behavioral change with the unification, and so the old mechanism and new mechanism

I don't think so.
I think Lucas is showing what's happening in `master` and simply would like to check if it is possible to fix that bad behavior under these PRs.

> The access-log timestamp is parsed out of the access log, and it reflects the proper time. The k8s timestamp on the access-log is incorrect, it's the time it was stored to k8s.

This is why I'm saying that _it is expected that in the "head" logs from the app are missing, while in the "tail" logs from the istio-proxy are missing._ (or vice-versa). Specifically, for the `istio-proxy` logs, we can only query by the incorrect timestamp because that's what allows the cluster API. When parsing each etry, the right timestamp is taken. But then, I may think it is expected to have some offset that is evident at the beginning and end of the unified view if the form of additional or missing entries for some of the containers.

jshaughn:
> This is why I'm saying that it is expected that in the "head" logs from the app are missing, while in the "tail" logs from the istio-proxy are missing. (or vice-versa). Specifically, 

Yes, I wasn't implying there was a problem, I was in fact saying that I think the existing code was correct in its handling, and that then new code should be returining the same (and I'm not saying it's not).

lucasponce:
> If we want to be strict on the queried start/end time range, for the istio-proxy we may need to drop entries that are before the startTime after resolving the right timestamp, and also query a little bit past the requested duration/time to not miss any entry that falls in the range (of course, this is assuming that the Istio timestamp is always a little behind the kuberentes timestamp). I'm not sure if this is worth to fix.

I don't think it's worth refactoring this too much, but at least it requires some explanation in the way of a tooltip or similar.

As a user, I didn't expect that out-of-sync logs, because I know that it should be a 1-to-1 mapping between the app log and the istio-proxy one, and like me, other users will expect something similar if they try to figure out why.

But the problem is beyond the Kiali domain, but at least some information about the differences in the timestamp can help the user not identify this as a potential bug.

In the context of this work in this area, I guess it's worth providing those enhancements.

lucasponce:
I looked closer to prepare a better example that may help, and I think in this area there is something not working well.
Not discussing if the user knows where the timestamp comes from, but here it shows where the lines are not synced specially when scrolling is happening:
image

I haven't investigated but I have some gut feeling that's related to the UI side.

So hope it helps, as commented, if you need more details I can show this in a live example.

lucasponce:
Yes, but please, note that I'd like to see this comment https://github.com/kiali/kiali/pull/5201#issuecomment-1162903204 addressed, on this work or in a different issue if you don't see fit.

I consider that example something that is not working well.

israel-hdez:
@lucasponce I checked more deeply the sorting issue. The issue is in the back-end. It works OK given the decisions made in the past. See #5246.

I do not want to do that change here because it is uncertain for me why it was chosen to normalize up to the seconds. So, I think it deserves a :+1: from somebody who knows better why timestamps are being truncated.

frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
```diff
@@ -730,65 +763,60 @@ export class WorkloadPodLogs extends React.Component<WorkloadPodLogsProps, Workl
...
     if ('' === val) {
       this.clearShow();
-    } else {
-      this.setState({ showLogValue: val });
     }
   };
 
-  private filterEntries = (entries: Entry[], showValue: string, hideValue: string): void => {
-    entries.forEach(e => (e.isHidden = undefined));
+  private filteredEntries = memoize((entries: Entry[], showValue: string, hideValue: string, useRegex: boolean) => {
```
jshaughn, line 783: I think a comment about what `memoize` is doing could be helpfule here.  I had to go look up what this library was doing and I still may not be sure.  I guess what this is about is caching the the filteredEntries for a combination of entries/showValue/hideValue/useRegex ?  That cache could get large if the user makes a lot of changes to the params and the set is large, no?

jshaughn (changes requested):
For `yarn start` I see errors in the terminal.  It seems mico-memoize is expecting something that is not generated in our dev builds.  We need to get rid of those errors.

israel-hdez:
@jshaughn I fixed the errors in the terminal by using the previous release of `micro-memoize`. I reported the issue to the author(s) here: planttheidea/micro-memoize#79

frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
israel-hdez: I somehow though that memoization was a widespread concept (there is, even, a [Wikipedia article about it](https://en.wikipedia.org/wiki/Memoization)). In typescript/javascript I think is somewhat common.

Anyway, you are right that it is caching the result given a set of arguments. This specific library has a single element cache by default (see: https://www.npmjs.com/package/micro-memoize/v/4.0.9#maxsize), so I think this should be OK.

jshaughn (changes requested):
@israel-hdez The speed seems good, definitely snappier.  Although the browser search is limited to the on-screen logs, I don't think we need to add a new "Find"  feature.  For me it's confusing given we already have the "Show" and "Hide".  Let's remove it for now and see if users complain or generate an RFE.  The scroll bar looks strange, let's try a little more to fix it, if possible.

image

israel-hdez:
@jshaughn I pushed changes to remove the Find box and also fixed the padding to make sure scrollbars are on the right place.

jshaughn (dismissed):
Tested and seems to work well on both Chrome and FF.  All requested changes have been made.  @israel-hdez , I think this can close #4701 but we can still ask for feedback from the community user.  As a follow-up I think the millisecond sort improvement would be good to look at, to see if it is feasible.  Approving, as the CI failure looks like a blip, tests passed for me locally.

israel-hdez:
@jshaughn A `console.log` statement slipped and, apparently, that was causing the build error.
I hope you can re-approve.

jshaughn (approved):
I always search for console logging, I can't believe I missed that, sorry!  Re-approving.

israel-hdez:
@jshaughn No problem. It is my fault in the first place because I introduced it.

israel-hdez:
@jshaughn BTW, thanks for the reviews!

PR #5202, opened 2022-06-13
Attempt to ensure colors are always defined in advance of use in the graph

jshaughn:
@lucasponce I can't reproduce the issue but perhaps this will fix https://github.com/kiali/openshift-servicemesh-plugin/issues/22.  Maybe give it a try.


lucasponce (approved):
It seems it worked.
I tried to reproduce the previous problems with local/remote development and I don't see them in my env.

PR #5203, opened 2022-06-14
[v1.44] Fix race while querying the Istio registry

lucasponce:
Per discussion https://github.com/kiali/kiali/pull/4572 this fix is backported to the v1.44 to make it available for Istio 1.12 users.

cc @zijianz

### DIFF
--- a/business/apps_test.go
+++ b/business/apps_test.go
@@ -21,6 +21,7 @@ import (
 func setupAppService(k8s *kubetest.K8SClientMock) AppService {
 	prom := new(prometheustest.PromClientMock)
 	layer := NewWithBackends(k8s, prom, nil)
+	setupGlobalMeshConfig()
 	return AppService{k8s: k8s, prom: prom, businessLayer: layer}
 }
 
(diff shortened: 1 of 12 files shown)


lucasponce:
This would require some manual build I think.

lucasponce:
@hhovsepy @aljesusg could you give a final test?

>  I just wanted to make sure v1.44 is the intended target.

It was requested here https://github.com/kiali/kiali/pull/4572#issuecomment-1135321435 I think they can't change the Istio version.

Also, note that 1.44 doesn't have the UI in the same repo, so probably the release process should be manual cc @leandroberetta

aljesusg (approved):
LGFM

leandroberetta:
> Also, note that 1.44 doesn't have the UI in the same repo, so probably the release process should be manual cc @leandroberetta

Ok, yes, new pipelines are designed to work with the UI in the same repo where the server is. Manual steps are required, I will review the manual process for this one.

lucasponce:
@leandroberetta @jshaughn @jmazzitelli once merged this may need a release in the community for the 1.44.x.
Good thing is that the UI hasn't changed.

Issue #5143, opened 2022-06-02
[cypress] Test to measure graph loading time


nrfox:
A cypress test to visit the graph page and wait for the graph to fully load.

This is separate from the other cypress testing efforts and more relates to scalability. As such, it does not need a feature file and can be written in the standard cypress syntax.

josunect:
tests/perf

ScriptingShrimp:
On client side, we can use [performance.getEntries()](https://developer.mozilla.org/en-US/docs/Web/API/Performance/getEntries). I can imagine some tests on top of that in Cypress - this kind of snapshot can give us base line if we start tune API or application code. I do expect various environments to perform differently - but could be really easy to implement.

If we want to take things more seriously:
- Lighthouse can be used (even in console/headless mode) https://web.dev/lighthouse-ci/
- Playwright could be weapon of choice [with some library ](https://dev-tester.com/check-your-sites-performance-with-artillery-and-playwright/) (need to take a look more in depth)

Issue #5144, opened 2022-06-02
Tunable dimensions for performance tests


nrfox:
Performance testing should be run across a number of dimensions. Namely we want to run performance tests with:
- Varying number of namespaces
- Varying number of workloads
- Varying cluster age
- Varying traffic rates between workloads

lucasponce:
I guess it's implicit in the description, but as part of this task, it would require a follow-up of https://github.com/kiali/kiali/issues/5142 to deploy the test applications (based on https://github.com/kiali/demos) to simulate heterogeneous and rich scenarios.

Worth to use most of the demo apps to combine all types of traffic (gRPC, HTTP), errors, and potentially ServiceEntry traffic (but this one I'm less sure about as it can introduce undesired elements in the perf environment).

nrfox:
This has been partially completed. The perf cluster script includes creation of a variable number of workloads but not all of the demos are being used and the auto-generated workloads are limited to a single namespace.

To finish this up, I'd say we:
1. Deploy additional demo apps as part of the perf script
2. Add additional options to the perf script to change the number of namespaces/traffic rates.

nrfox:
Going to close this issue as done. Not all of the demo apps have been deployed but many of the parameters are tunable using the hack perf script.

PR #5205, opened 2022-06-14
Adding performance tests

josunect:
cypress Test to measure graph loading time 

Fixes https://github.com/kiali/kiali/issues/5143


tests/perf/cypress/e2e/Graphload.cy.js
```diff
@@ -0,0 +1,58 @@
...
+            .then(function (data) {
+                graphUrl = encodeURI("/console/graph/namespaces?traffic="+data.traffic
+                    +"&graphType="+data.graphType
+                    +"&namespaces="+data.namespaces
+                    +"&duration="+data.duration
+                    +"&refresh="+data.refresh
+                    +"&layout="+data.layout
+                    +"&namespaceLayout="+data.namespaceLayout);
+            })
+            .as('data');
```
nrfox, line 25: We won't use a fixture when running the perf tests against a real environment.

tests/perf/cypress/e2e/Graphload.cy.js
```diff
@@ -0,0 +1,58 @@
...
+            cy.get("#cy", { timeout: 10000 }).should('be.visible')
+                .then(() => {
+                    performance.mark("end")
+                    performance.measure("pageLoad", "start", "end");
+                    const measure = performance.getEntriesByName("pageLoad")[0];
+                    const duration = measure.duration;
+                    assert.isAtMost(duration, Cypress.config('threshold'));
+
+                    cy.writeFile('logs/performance.txt', `[PERFORMANCE] Graph load time for ${graphUrl}: \n ${duration / 1000} seconds`)
+                })
```
nrfox, line 54: Looks good to me!

tests/perf/cypress/e2e/Graphload.cy.js
```diff
@@ -0,0 +1,58 @@
...
+    it('Measure Graph load time', {
+            defaultCommandTimeout: Cypress.config('timeout')
+    }, () => {
+        cy.visit(graphUrl, {
+            onBeforeLoad(win) {
+                win.performance.mark("start");
+            }
+        })
+        .its("performance").then((performance) => {
+            cy.intercept(`**/api/namespaces/graph*`).as('graphNamespaces');
```
nrfox, line 41: You may need to move this call so that it is setup before the `cy.visit` call in case you try to visit the page before the intercept is setup.

tests/perf/cypress.config.js
```diff
@@ -0,0 +1,9 @@
+const { defineConfig } = require("cypress");
+
+module.exports = defineConfig({
+  e2e: {
+    baseUrl: 'http://localhost:3000',
+    timeout: 10000,
+    threshold: 100000
+  }
+});
```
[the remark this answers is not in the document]
josunect: Added support for typescript.

tests/perf/cypress/e2e/Graphload.cy.js
josunect: Moved. Thanks!

tests/perf/cypress/e2e/Graphload.cy.js
nrfox: Ah I missed the fact that the fixture was being used to parametrize the test. That's good. We probably won't need to configure all the request params and some of these can be hardcoded for now. The options that will need to be tunable are: https://github.com/kiali/kiali/issues/5144. We'll probably need to do some cleanup between tests and it makes sense to keep that within cypress. Otherwise the person running the tests will need to execute the tests one at a time and manually cleanup resources between them.

josunect: Ok. 
Is there any other test to add in order to measure any other page load?

nrfox: For now, no, that's the main thing to test but you can also add measuring the overview page in a similar way.

josunect: Thanks, @nrfox ! Changes added.

tests/perf/README.md
```diff
@@ -0,0 +1,15 @@
+# baseUrl default baseurl localhost:3000
+## Can be updated in cypress.config.js or:
+
+yarn cypress run --config baseUrl=http://mybaseurl
```
nrfox, line 4: These instructions don't work for me anymore now that there's no `package.json` here. Do you need to set a yarn setting to link to the `frontend/package.json`?

I tried
```
yarn --cwd /kiali/frontend cypress run --config-file .
```
Which seemed like it should work but now cypress is throwing some weird errors.

tests/perf/cypress/support/e2e.ts
```diff
@@ -0,0 +1,20 @@
...
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
```
nrfox, line 17: We'll also need to support logging into a cluster with auth when running these tests. There's already a login command that you can hopefully just import here and everything will work: https://github.com/kiali/kiali/blob/master/frontend/cypress/support/commands.ts#L43-L92? :crossed_fingers:

tests/perf/cypress/e2e/Graphload.cy.ts
```diff
@@ -0,0 +1,74 @@
+describe('graphload', () => {
```
nrfox, line 1: nitpick: can this file be named `Graphload.spec.ts`? Makes it more consistent with other frontend test files.

tests/perf/cypress.config.js
nrfox, line 9: Does cypress 9.x support this syntax or should this be a `.json` file? cypress 10.x seems to have switched to the `js` config but we're still on 9.x.

tests/perf/cypress/support/e2e.ts
josunect: I will add this change.

tests/perf/cypress/e2e/Graphload.cy.ts
josunect: I will add the change.

tests/perf/README.md
josunect: I have tested some options:

- Using --cwd, and it gave some errors. Even creating a new yarn setting, doesn't seems to work. Maybe because is in other path?
- Adding kiali-ui as a depency: Won't install the devDependies, just dependencies. So trying to run the tests, it will raise the error: cypress not found
- Adding a .yarnrc file with:
--modules-folder ../../frontend/node_modules
But fails running cypress, missing typescript dependency. Probably cypress uses a different configuration for its own dependencies? 

I can think of: 
- Keep separating the tests in frontend and tests/perf: Add a package.json in perf dir, with just the two dependencies required (cypress and typescript). It will run faster, but the risk of mismatched dependencies exist (Or look for an alternative like package.json inheritance to get the versions from the frontend package).
- Include perf test into the frontend and create a separate configuration for the /perf tests. Like, use workspaces, https://classic.yarnpkg.com/lang/en/docs/workspaces/ --> It would require to change the /perf location

josunect: Added a script to install kiali-ui dependencies, updated README.

nrfox: All the different config settings are a pain to get working together. Between cypress, typescript, and yarn there's a lot of different configs to manage. You mind if I open a PR against your branch/repo and you can review it and accept/reject any changes?

josunect: Please, go ahead!

frontend/cypress/perf/Graphload.spec.ts
```diff
@@ -0,0 +1,74 @@
+describe('graphload', () => {
+
+    http://localhost:3000/console/graph/namespaces/
+        // ?traffic=grpc%2CgrpcRequest%2Chttp%2ChttpRequest%2Ctcp%2CtcpSent
+        // &graphType=versionedApp
+        // &namespaces=bookinfo%2Ccert-manager
+        // &duration=60
+        // &refresh=15000
+        // &layout=kiali-dagre
+        // &namespaceLayout=kiali-dagre
```
nrfox, line 10: nitpick: Can this be cleaned up? Not sure what it's documenting.

frontend/cypress/perf/Graphload.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    }, () => {
+        cy.intercept(`**/api/namespaces/graph*`).as('graphNamespaces');
+        cy.visit(graphUrl, {
+            onBeforeLoad(win) {
+                win.performance.mark("start");
+            }
+        })
+        .its("performance").then((performance) => {
+            cy.wait('@graphNamespaces')
+            //cy.get('#loading_kiali_spinner').should('not.exist')
```
nrfox, line 60: Same here

nrfox (commented):
Two small nitpicks otherwise LGTM

frontend/cypress/perf/Graphload.spec.ts
josunect: Cleaned!

frontend/cypress/perf/Graphload.spec.ts
josunect: Cleaned!

nrfox (approved):
LGTM

PR #5209, opened 2022-06-14
[hack] ask that you be added as an admin on the cluster

jmazzitelli:
If you run this and point to a running cluster, you can be granted admin rights:

```
hack/ibmcloud-openshift.sh admin
```

This is what it looks like when the cluster is running and the operation succeeds:

```
$ hack/ibmcloud-openshift.sh -np jmazzite-kiali-perf admin
...
📄 Adding you as a cluster admin user
OK
The configuration for jmazzite-kiali-perf-cluster was downloaded successfully.

Added context for jmazzite-kiali-perf-cluster to the current kubeconfig file.
You can now execute 'kubectl' commands against your cluster. For example, run 'kubectl get nodes'.
If you are accessing the cluster for the first time, 'kubectl' commands might fail for a few seconds while RBAC synchronizes.
```

If the cluster is not running, this is the error you will get:

```
$ hack/ibmcloud-openshift.sh admin
...
📄 Adding you as a cluster admin user
FAILED
The specified cluster could not be found. If applicable, make sure that you target the correct account and resource group. (G0004)

To list the clusters you have access to, run 'ibmcloud ks cluster ls'. To list the resource groups that you have access to, run 'ibmcloud resource groups'. To target the resource group, run 'ibmcloud target -g <resource_group>'.

Incident ID: 1e....
```

### DIFF
--- a/hack/ibmcloud-openshift.sh
+++ b/hack/ibmcloud-openshift.sh
@@ -116,6 +116,12 @@ status() {
   [ "$(is_cluster_deployed)" == "true" ] && infomsg "Cluster is deployed" || infomsg "Cluster is NOT deployed!"
 }
 
+# FUNCTION: set_admin - Grants to the user full admin rights on the cluster.
+set_admin() {
+  infomsg "Adding you as a cluster admin user"
+  ibmcloud oc cluster config --cluster ${CLUSTER_NAME} --admin
+}
+
 # FUNCTION: finish - Waits for the cluster to be deployed and then finishes the deployment.
 finish() {
   infomsg "Waiting for the cluster to be fully deployed"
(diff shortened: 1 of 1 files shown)


jmazzitelli:
cc @josunect @nrfox you may want to review this.

jmazzitelli:
> extend the feature to provision users - Not just the current one.

If you can find out how to do that, that would be great! But! When I tried to do it (when I first wrote that ibmcloud-openshift.sh script), I was told by the IBM folks that you could not provision new users in a cloud cluster.

I wanted to do what our [`make cluster-add-users` target](https://github.com/kiali/kiali/blob/master/make/Makefile.cluster.mk#L129-L179) does - but it doesn't work with the IBMCloud cluster.

So if you can find a way to provision other users, I would be all for that, but I was told it is not possible to do it.

nrfox (approved):
LGTM

PR #5210, opened 2022-06-14
Fix issue with lock-icon sometimes not being available

jshaughn:
- make the pficon webfont available locally, preload it,
  and immediately apply it to styling. This  seems to finally
  solve the issue of the lock icon  (and possibly icons used in
  the ace editor, also webfont-based), not always being
  intially rendered,

@lucasponce , After many tries I think I finaly resolved this annoying webfont issue re the lock icon.


frontend/public/index.html
```diff
@@ -22,6 +22,14 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>Kiali</title>
+    <!-- Preload the pficon font to ensure it is available immediately for Graph edges or ace editors -->
+    <link rel="preload" href="%PUBLIC_URL%/pficon.woff2" as="font" type="font/woff2" crossorigin />
```
lucasponce, line 26: Did you check if this link works without the "cross-origin" tag?

They shouldn't be open as it's a source of high vulnerabilities, a couple of similar links on that page don't use it:
image

lucasponce (approved):

Issue #5207, opened 2022-06-14
Customizable links in Kiosk mode


lucasponce:
As part of the improvements of the Kiosk mode, Kiali should have some mechanism to capture external links that can be used for integration with external Kubernetes consoles.

I.e. User may want to link to an existing details page for a workload of a service.

This issue will prototype some early steps in that direction.

This feature should be only enabled on kiosk mode, standalone use cases should keep working as usual.

PR #5211, opened 2022-06-15
Introduce conditional links for Kiosk mode on Overview page

lucasponce:
Related to https://github.com/kiali/kiali/issues/5207

This PR adds a new logic to the Kiosk mode:
- It would provide a mechanism to send messages to the parent of the kiosk.
- These messages delegate the navigation on certain links on the parent instead of Kiali.
- The first step is scoped mostly in the Overview links introducing a conditional rendering on kiosk presence.

This work is transparent to the standalone UI and no logic/feature should be affected.

### DIFF
--- a/frontend/src/components/Nav/Page/RenderHeader.tsx
+++ b/frontend/src/components/Nav/Page/RenderHeader.tsx
@@ -28,8 +28,11 @@ const actionsToolbarStyle = style({
   borderBottom: '1px solid #d2d2d2'
 });
 
-interface RenderHeaderProps {
+type ReduxProps = {
   isKiosk: boolean;
+}
+
+type RenderHeaderProps = ReduxProps & {
   location?: {
     pathname: string;
     search: string;
--- a/frontend/src/components/Kiosk/KioskActions.ts
+++ b/frontend/src/components/Kiosk/KioskActions.ts
@@ -51,7 +51,8 @@ export const kioskOverviewAction = (showType: Show, namespace: string, duration:
 
 // Message has no format, parent should parse it for its needs
 const sendParentMessage = (msg: string): void => {
-  // Kiosk parameter will capture the parent target; this will enable iframe -> parent communication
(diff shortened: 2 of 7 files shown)


jshaughn (commented):
The kiosk harness is definitely more powerful with the changes.  The conditional kiosk logic is a little messy but unavoidable.  I think probably the `isKiosk` needs to change from a boolean to a string and the conditionals should likely be limited to os-console.  We don't want to negatively impact community users already using kiosk mode.

lucasponce:
>  The conditional kiosk logic is a little messy but unavoidable. I think probably the isKiosk needs to change from a boolean to a string and the conditionals should likely be limited to os-console. We don't want to negatively impact community users already using kiosk mode.

Good point and +1.

I kept the "kiosk=true" working as designed, only when "kiosk=<parent>" is when the conditional links will send messages to the parent of the Kiosk.

Then, this logic can be generic, and no need to involve any platform detail in Kiali logic.

I've updated the PR, let me know how it looks.

jmazzitelli (approved):
built and tested with the plugin PR. It works

frontend/src/components/Nav/Page/RenderHeader.tsx
```diff
@@ -28,8 +28,11 @@ const actionsToolbarStyle = style({
   borderBottom: '1px solid #d2d2d2'
 });
 
-interface RenderHeaderProps {
+type ReduxProps = {
   isKiosk: boolean;
+}
```
jshaughn, line 33: I don't think we want both `kiosk` and `isKiosk` in global state.  Is this intentional or maybe an oversight?

jshaughn (commented):
@lucasponce Thanks for the changes, looking good, one comment.

lucasponce:
Merging. The kiosk state is not definitive and there will be a couple of more follow-up work until these features consolidate.

Issue #5206, opened 2022-06-14
Not found messages may have a better message in the body page


lucasponce:
Today, when an URL has a resource that is not found or it fails, there is a message in the message center but a "blank" body:
![image](https://user-images.githubusercontent.com/1662329/173563102-9141e615-4df3-4224-b62a-b73f6b2c08ab.png)

This can be improved to have a message similar to the Graph page:
![image](https://user-images.githubusercontent.com/1662329/173563719-4e4dce90-be73-4ab1-9af2-4b236fd9c159.png)

That would help as well in embedded scenarios:
![image](https://user-images.githubusercontent.com/1662329/173563901-e82ce91c-a1f1-43b2-ad13-996c03383282.png)

So, the message could be something like "<resource> is not found in the Mesh" and adding the detail of the warning/error message.

A blank page looks like something unexpected/broken.

PR #5218, opened 2022-06-16
Not found messages in body page for workloads and Apps

josunect:
Not found messages may have a better message in the body page.
Done for apps and workloads. 

Screenshot from 2022-06-16 13-56-33

Fixes https://github.com/kiali/kiali/issues/5206

### DIFF
--- a/frontend/src/components/ErrorSection/ErrorSection.tsx
+++ b/frontend/src/components/ErrorSection/ErrorSection.tsx
@@ -1,0 +1,28 @@
+import * as React from 'react';
+import {style} from "typestyle";
+import {EmptyState, EmptyStateBody, EmptyStateVariant, Title, TitleSizes} from "@patternfly/react-core";
+import {ErrorMsg} from "../../types/ErrorMsg";
+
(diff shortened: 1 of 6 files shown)


hhovsepy:
Hi @josunect, a suggestion, can the Service details page also be changed to show this message? Because current empty props in confusing.

hhovsepy:
Thank you for the fixes @josunect 
Screenshot from 2022-06-16 17-17-24

lucasponce:
I wonder if we could put something similar in the IstioConfig details:
image

Then all the details will look similar, I forgot to mention in the original issue but after testing it locally I think it makes sense.

josunect:
> I wonder if we could put something similar in the IstioConfig details: image
> 
> Then all the details will look similar, I forgot to mention in the original issue but after testing it locally I think it makes sense.

Thanks, @lucasponce ! Changes added.

hhovsepy:
Verified with config details.

hhovsepy (approved):

PR #5223, opened 2022-06-17
Graph customizable links

lucasponce:
Related https://github.com/kiali/kiali/issues/5207


frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "Instead update 'web_root' in Kaili config map in OpenShift console."
   ],
   "homepage": "./",
+  "proxy": "http://localhost:8000",
```
lucasponce, line 18: Missed to remove the proxy.

leandroberetta (dismissed):
LGTM. Tested alongside with https://github.com/kiali/openshift-servicemesh-plugin/pull/27

jmazzitelli (dismissed):
Tested the links and all looks good.

frontend/src/components/Link/KialiPageLink.tsx
```diff
@@ -2,38 +2,61 @@ import * as React from 'react';
...
+class KialiPageLink extends React.Component<KialiPageLinkProps> {
 
-  // If cluster is specified, and it's the home cluster, render a local link.
-  if (!serverConfig.clusterInfo?.name || props.cluster === serverConfig.clusterInfo.name) {
-    return <Link to={props.href}>{props.children}</Link>;
-  }
+  render() {
+    // Without a cluster, simply render a local link
+    // If cluster is specified, and it's the home cluster, render a local link.
+    if (this.props.cluster === undefined || !serverConfig.clusterInfo?.name || this.props.cluster === serverConfig.clusterInfo.name) {
```
jshaughn, line 24:
```suggestion
    if (!this.props.cluster || this.props.cluster === serverConfig.clusterInfo?.name) {
```

frontend/src/components/Link/KialiPageLink.tsx
```diff
@@ -2,38 +2,61 @@ import * as React from 'react';
...
+      if (this.props.kiosk.length > 0 && this.props.kiosk !== 'true') {
```
jshaughn, line 25: This test is becoming so frequent that perhaps a simple utility would be nice, `isParentKiosk(this.props.kiosk)`.

frontend/src/pages/Graph/SummaryPanelNodeTraces.tsx
```diff
@@ -19,12 +19,17 @@ import { TraceListItem } from 'components/JaegerIntegration/TraceListItem';
...
 import { DecoratedGraphNodeData } from 'types/Graph';
 import transformTraceData from 'utils/tracing/TraceTransform';
+import {kioskContextMenuAction} from "../../components/Kiosk/KioskActions";
 
-type Props = {
+type ReduxProps = {
+  selectedTrace?: JaegerTrace;
+  setTraceId: (traceId?: string) => void;
+  kiosk: string;
+};
```
jshaughn, line 28: alphabetize

jshaughn (approved):
As much as I don't like all the conditional logic, I still don't see a better way, and at least it's clear and explicit.  A couple of minor comments for consideration, but nothing blocking, so I will approve.

frontend/src/components/Link/KialiPageLink.tsx
lucasponce: It's a good one, but I'm reserving this change for follow-up refactoring.
I think we need to keep the kiosk parameter for "true/false" values and also introduce another set of kiosk aren't, kioskXYZ for additional parameters that it may need.

The next tasks is to work in the links from the "Service Mesh" tab to Kiali, and in that context, I'd apply this recommendation.

jshaughn (approved):
Jut re-approving after some invalidating changes.

PR #5225, opened 2022-06-17
Perf environment - deploy separate prometheus

nrfox:
Instead of using the addon prometheus, deploy a separate instance with dedicated volumes and resources.

### DIFF
--- a/hack/perf-ibmcloud-openshift.sh
+++ b/hack/perf-ibmcloud-openshift.sh
@@ -132,7 +132,7 @@ install_istio() {
 
 install_prometheus() {
   # Need to first setup the security context
-  oc apply -f - <<SCC
+  cat <<SCC | oc apply -f -
 apiVersion: security.openshift.io/v1
 kind: SecurityContextConstraints
 metadata:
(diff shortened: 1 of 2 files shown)


hack/perf-ibmcloud-openshift.sh
```diff
@@ -127,7 +127,48 @@ install_istio() {
   if [ -n "${ISTIO_VERSION}" ]; then
     iv_option="-iv ${ISTIO_VERSION}"
   fi
-  hack/istio/install-istio-via-istioctl.sh -iee true -cp openshift ${iv_option}
+  hack/istio/install-istio-via-istioctl.sh -iee true -cp openshift ${iv_option} --addons "grafana jaeger"
+}
+
+install_prometheus() {
+  # Need to first setup the security context
+  oc apply -f - <<SCC
```
jmazzitelli, line 135: are you sure this syntax works? I just tried a simple test and it did not:

```bash
$  oc apply -f - <<SCC
> apiVersion: v1
> kind: Namespace
> metadata:
>   name: foo
> SCC
error: Missing or incomplete configuration info.  Please point to an existing, complete config file:
...
```

jmazzitelli: This should work: 

```suggestion
  cat <<SCC | oc apply -f -
```

jmazzitelli (approved):
I didn't run this, but all the code looks good.

Issue #5187, opened 2022-06-10
Update the information about mTLS data in details pages


lucasponce:
The Inbound/Outbound metrics offer a rich way to group metrics by different criteria:
![image](https://user-images.githubusercontent.com/1662329/173056115-3133a668-3262-48c5-9d1c-26081772dfbd.png)

But it doesn't offer to group traffic by "connection_security_policy" which is used to see what traffic is under mTLS vs plain text.

Grafana has these criteria defined in an explicit way:
![image](https://user-images.githubusercontent.com/1662329/173056357-01cc13b4-63ca-4830-b19e-249564412796.png)

And Kiali should provide this information as well, the Graph shows this information on the edges using locks:
![image](https://user-images.githubusercontent.com/1662329/173056680-43d1e174-7c26-44e7-a687-f42622eadff7.png)

What it's not clear in the traffic tab:
![image](https://user-images.githubusercontent.com/1662329/173056769-c65169d1-d302-48f3-99fc-0f624f54b5a3.png)

In short, this issue would provide:
- A explicit way to tell in the "Traffic" tab which traffic is mTLS vs which one is not.
- A new Inbound/Outbound grouping criteria in the Istio charts to see traffic split into mTLS vs no mTLS.

PR #5243, opened 2022-06-23
Add mTLS filter in details pages

josunect:
Added Connection Security Policy filter to show traffic using mTLS in Services and App details for Inbound and Outbound metrics.
 
image

Fixes https://github.com/kiali/kiali/issues/5187


josunect:
I will update a test to include the validation of the select options, TestPrepareIstioDashboard.

Issue #4511, opened 2021-11-19
Revisit DestinationRule no labels warning


lucasponce:
I wonder if the DR validation of no labels should be applied when there are more than one subset or to keep it as it is defined today.

Perhaps on this case it sounds redundant:
![image](https://user-images.githubusercontent.com/1662329/142606219-67ce96b0-df82-493e-88af-36a6199ece8d.png)

hhovsepy:
If we change it, then IMHO it should be either:
a) Keep the message level as warning and show it in 2 cases:
1 There is only one subset and it is without labels, display warning on that subset.
2 There are more than one subsets, and all of them without labels, display warning per each subset.
This will make the validation logic over-complicated, also confusing users when some warning might appear or gone.

b) Change the level of KIA0209 from warning to info. And introduce one more message with warning only in case when no any subset has labels and show it on top of DR.
But this also makes the logic a bit over-complicated.

lucasponce:
@hhovsepy putting on this as you worked recently in these validations.
I wonder if the scope is still valid.

PR #5253, opened 2022-06-27
KIA0209 message severity is decreased to Info when more subsets with valid labels.

hhovsepy:
RFE https://github.com/kiali/kiali/issues/4511

When more subsets with valid labels, only Info is shown:
Screenshot from 2022-06-29 17-44-52

When all subsets have missing labels, the Warning severity of KIA0209 is kept.
Screenshot from 2022-06-29 17-45-43

### DIFF
--- a/business/checkers/destinationrules/no_dest_checker_test.go
+++ b/business/checkers/destinationrules/no_dest_checker_test.go
@@ -664,7 +664,6 @@ func TestSubsetWithoutLabels(t *testing.T) {
 	assert := assert.New(t)
 
 	vals, valid := NoDestinationChecker{
-		Namespace: "test-namespace",
 		WorkloadsPerNamespace: map[string]models.WorkloadList{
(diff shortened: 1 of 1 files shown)


lucasponce:
Please, would you mind rebasing this PR with the master? The big work around the refactoring is consolidating and it would be nice to see it all together. Thx!

hhovsepy:
rebased

israel-hdez:
@hhovsepy Just to confirm: I see that the warning for subsets without labels is kept if all other subsets have invalid labels. However, as soon as one subset has valid labels, the warning is decreased to info. Is this the intended behavior?

hhovsepy:
> @hhovsepy Just to confirm: I see that the warning for subsets without labels is kept if all other subsets have invalid labels. However, as soon as one subset has valid labels, the warning is decreased to info. Is this the intended behavior?

@israel-hdez right, this is intended behavior, it shows warning when there is no subset label match at all, if there is at least one, then info is shown.

Issue #5151, opened 2022-06-03
[scalability] How to thin metrics to those required only by Kiali


jshaughn:
Investigate providing a configuration that limits Istio to generating only the Prometheus metrics required by Kiali.   For users that may only need the metrics to run Kiali, this would potentially make it easy to significantly reduce metric storage, and to avoid unnecessary processing of unused metrics.  Users could of course modify the configuration to support custom dashboards etc...  This would also avoid users removing required metrics or attributes that would break Kiali.

This would also allow us to analyze the impact of such changes on performance.

Xunzhuo:
This is a good idea, one of the problems we meet in production is that the metrics storage  is very large, istio provides many metrics in default besides istio_xxxx metrics. Cutting off the extra unused metrics can save a lot of storage in Promethues, and reduce its pressure.

lucasponce:
Side comment, Kiali use also some native envoy metrics for the "Workload -> Envoy -> Metric" dashboard.
As part of this work, perhaps that dashboard could be revisited and reviewed if the default ones are good/relevant ones or it can be revisited.

jshaughn:
> Side comment, Kiali use also some native envoy metrics for the "Workload -> Envoy -> Metric" dashboard.

Yes, I recently captured this in my update to kiali.io [required metrics FAQ.](https://staging.kiali.io/docs/faq/general/#which-istio-metrics-and-attributes-are-required-by-kiali)  I did the update as a first step to this effort.  Feel free to review that new content!

jshaughn:
There may be other things we can do in this area, but work is at a reasonable first milestone, and so closing this infavor of new issues as needed.

PR #5254, opened 2022-06-27
Initial set of metric and label drops

jshaughn:
Related to #5151 

This PR adds handling for when Histogram telemetry is dropped, thus crippling the UI.  It disables invalid options in the graph display and metric settings dropdowns,  and adds some info tooltip text.

So, if you, for example, drop `istio_request_duration_milliseconds_bucket`, as well as istio_request_bytes and istio_response_bytes (as documented in https://github.com/kiali/kiali.io/pull/565), you should see something like this:

image

Note the disabled option in the graph display options above.  And below, the disabled settings as well as the empty charts:

image

**To TEST**
1. install istio via the istioctl hackscript
2. install kiali via the PR
3. install some demo app to get some traffic
4. verify that your graph Display options for edge labels are all available, as well as detail page metric chart histogram options
5. copy <istio_home>/samples/addons/prometheus.yaml to some local location
6. edit it according to https://staging.kiali.io/docs/configuration/p8s-jaeger-grafana/prometheus/#metric-thinning
7. `kubectl -n istio-system apply -f <new-prom-config>`
8. restart the prometheus pod
9. this should reduce metrics but should not have any noticeable effect on kiali
10. further edit the config, this time dropping some histogram metrics, see https://staging.kiali.io/docs/configuration/p8s-jaeger-grafana/prometheus/#metric-thinning-with-crippling
11. again apply config and restart prometheus
12. this time you should see disabled options in graph display menu and in metrics histogram options, in acordance to what you dropped.
13. finally, verify the enhaced info tooltips in the relevant graph display menu options, and the new info tooltip for metrics histogram options.

Updates:
[08/05/22]
Retargeting this PR as code that handles the crippling of telemetry when users decide to drop some or all histogram metrics.

Below is the original description/notes, but it's applicable now to the kiali.io PR: https://github.com/kiali/kiali.io/pull/565:

- avoid "keep", it's too unpredictable,
  only "drop" things we know we don't want
- make sure not to drop prom labels mapped from pod labels
  - these are needed by custom dashboards
  - note that, for now at least, this allows us to drop some canonical
    labels

[06/27/22]
I'm thinking now that the Telemetry API may not be needed, or may only be needed in a smaller capacity.  I think the TAPI is maybe better for adding metrics, as it avoids a lot of messiness in the filters and such.  But for dropping, it seems simpler to use the native Prometheus config and filter things away at ingestion time.  This takes a nominal amount of time and accomplishes the main goal of reducing storage.

The initial commit removes a large set of metrics, as well as labels unused by kiali or the built-in dashboards.  There are still more metrics, especially envoy metrics, that could be removed, it's more a trade-off with the ultimate complexity of the filtering regex.  It's not horrible to leave metrics with relatively low cardinality.  But, I think there are some histograms that likely should still be removed.  As for histograms, that's the next hurdle (see more below).

There are a variety of useful Prometheus queries to analyze "top" metrics from a cardinality perspective.  Here is a good one:

Top metrics by count:
`topk(20, count by (__name__)({__name__=~"istio.+"}))`
`topk(20, count by (__name__)({__name__=~"envoy.+"}))`

Histograms are the major offenders for storage.  The irony is that in many cases, in particular when left to default buckets, the actual quantiles (percentiles) produced have large degrees of error.  Unless there are specific SLAs to meet, or if there is a major lack of uniformity in values (be it response time, response size, etc), it may not be worth the overhead, because values are not particularly valid or valuable.

[06/29/22]
Note that the Istio histograms are defaulting to the [Envoy bootstrap buckets for Histograms](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/metrics/v3/stats.proto#envoy-v3-api-msg-config-metrics-v3-statsconfig):

[
  0.5,
  1,
  5,
  10,
  25,
  50,
  100,
  250,
  500,
  1000,
  2500,
  5000,
  10000,
  30000,
  60000,
  300000,
  600000,
  1800000,
  3600000
]

This is not great behavior.  Note that there are 20 buckets (the 19 above and the 1 implicit bucket for +inf).  Decent quantiles are usually possible with 10 [reasonably chosen] buckets.  Note also that the same buckets are used regardless of units, meaning milliseconds for response times, and bytes for request and response sizes.  The result may be a larger set of time-series (each label-set will be repeated for every bucket it falls into), as well as increased bucket updates (as each bucket is cumulative.  Note that for consistently behaving services this is less of a problem.

I had a [virtual] conversation with Doug Reid (Istio telem lead) about the histograms and he agrees that Istio likely needs a [backward compatible] way to better handle things.  Minimally, a chance to maybe apply better defaults, or even to allow customization.  There is already at least one [related issue](https://github.com/istio/istio/issues/38277).  He is going to try and start a design document, but it should be noted that there is no one on the google side to pursue this work, impl would likely need to be a contribution.

### DIFF
--- a/prometheus/client.go
+++ b/prometheus/client.go
@@ -341,7 +341,7 @@ func (in *Client) GetExistingMetricNames(metricNames []string) ([]string, error)
 		metricsWeAreLookingFor[string(metricNames[i])] = true
 	}
 
-	metricsWeFound := make([]string, 0, 5)
+	metricsWeFound := make([]string, 0, len(metricNames))
 	for _, item := range results {
 		name := string(item)
 		if metricsWeAreLookingFor[name] {


tools/prometheus.yaml
```diff
@@ -0,0 +1,507 @@
+---
```
jmazzitelli, line 1: I suggest putting a comment(s) in here pointing the reader to the places you modified (I assume these are the standard addon yamls that you modified somewhere to drop additional metrics?). Right now I can't see what the user needs to pay attention to. A user may have his own yaml or installation mechanism to install Prometheus and only needs to know the few lines of code here that drop the metrics.

jmazzitelli (changes requested):

tools/prometheus.yaml
jshaughn: sure, but this is WIP, not asking for review yet...

jshaughn:
**Scrape Interval**

This comment is to capture information regarding the configured Prom globalScrapeInterval.   The scrape interval can have a significant effect on metrics collection overhead as it takes effort to pull all of those configured metrics and update the relevant time-series.  And although it doesn't affect time-series cardinality, it does affect storage for the data-points, as well as potential impact when aggregating query results, computing quantiles, etc.

Users should think carefully about their configured scrape interval.  Note that the Istio addon for prometheus configures it to 15s.  This is great for demos but may be too frequent for production scenarios.  The prometheus helm charts set a default of 1m, which is more reasonable for most installations, but may not be the desired frequency for any particular setup.

I don't think for Kiali there is a recommended setting, but the impact on Kiali should be understood.  First, it is important to realize that request rates (or byte rates, message rates, etc) require a minumum of two data points:

`rate=(dp2-dp1)/timePeriod`

That means for Kiali to show anything useful in the graph, or anywhere rates are used (many places), the minimum duration must be `>= 2 x globalScrapeInterval`. 

- note that prom can be configured such that individual scrape points can override the global setting, but Kiali is not currently concerned with this corner case.

Kiali will eliminate invalid Duration options given the globalScrapeInterval.   Also, Kiali does a lot of aggregation and querying over time periods.  As such, the number of data points will affect query performance, especially as the time periods grow.

The bootom line is that the customer needs to choose a value that works for them, our recommendation is that be the largest interval that meets their needs.

jshaughn:
With the help of @nrfox we are going to run a storage comparison test using the ibm cloud perf env, one without thinning and one with thinning.   Very basic testing on my local box show about a 28% reduction in storage size when thinning.  We'll see if the larger scale test matches these results.

To get more significant savings it will likely require users to either drop or redefine some of the histograms.  To put this into perspective, given the default bucketing defined by istio/envoy (as shown above), there are more than 60 time-series created for every time series in istio_requests_total.  That's because you get 20 buckets for each of the histograms: istio_request_bytes_bucket, istio_response_bytes_bucket and istio_request_duration_milliseconds_bucket (as well as one for sum and one for count).

Note that I don't yet know how to get these histograms to be redefined.  But dropping a histogram would be a large savings.  Dropping the buckets (leaving only sum and count), would allow an average but no percentiles.  And redefining the buckets from 20 to a well-chosen-10 would still be a huge savings and would likely give percentile results as good, or better than the defaults.

jshaughn:
Just noting a few more useful things for examining the db:

The Prom Console shows a lot of TSDB info/stats under the `Status` dropdown.   Or, some of that info can be gathered with the API, such as:

TSDB stats:  `http://localhost:9090/api/v1/status/tsdb`

Prom query to see histogram buckets: `sum(istio_request_duration_milliseconds_bucket) by (le)`

Prom console query to graph used space: `kubelet_volume_stats_available_bytes{persistentvolumeclaim="your-pvc"}`

How to measure storage size given the data mount point: `du -sh /data`
- `kubectl exec -it -n istio-system -c prometheus-server $(kubectl get pods -o name -n istio-system -l app=prometheus) -- du -sh /data`

jshaughn:
@nrfox or anyone,
Looking for feedback on how Kiali UI should handle not finding particular metrics it typically assumes are there.  This is in regard to users that want to drop, or partially drop, the histograms that back request/response size, and request duration (aka as request/response throughput and response time, in the graph).

The histograms are hogs and for users that don't need them, or are willing to lose a feature to save storage, it may be worth dropping.   Our recommendation will be that they either drop the entire histogram, or just the `_bucket` metric for the histogram.  I checked, and we don't break, the ui simply doesn't show data it can't get.  It may be attractive to drop just the `_bucket` metric for certain histos, because you lose the percentiles but you can still get rates and averages.

My question is whether we need to be proactive and actually disable the options that have been crippled, or, just leave it as-is, and not show the data they don't have (even if they explicitly ask for it).  My take is that we could iterate, and add disabling later.  We're not sure if folks will choose to do this crippling, or if they would care.  But, if folks think it's too poor of an experience, we could add the support now.

jmazzitelli:
You know what's going to happen - people are going to say, `My goodness... prometheus is a hog, I'm going to remove some metrics I don't care about.....` (1 day later) `HEY! Why doesn't Kiali show my Response times!!!`

I think if you detect missing metrics, you should at minimum post a Info log message, if not an Info message in the message center .. "Some metrics have been removed from telemetry - you will not see some data in Kiali due to the missing metrics"

jshaughn:
> I think if you detect missing metrics, you should at minimum post a Info log message, if not an Info message in the message center .. "Some metrics have been removed from telemetry - you will not see some data in Kiali due to the missing metrics"

Unfortunately I think you're probably right, and it's likely worth going through the hassle now instead of later.  I just hate doing pre-emptive work when I'm not sure it's necessary.

nrfox:
> I think if you detect missing metrics, you should at minimum post a Info log message, if not an Info message in the message center .. "Some metrics have been removed from telemetry - you will not see some data in Kiali due to the missing metrics"

How much work is it to display an info message vs. just doing the work to disable the option? To display the info message Kiali will need to first detect that the buckets are missing which is most of the work to disable the options right?

I lean more towards not showing the missing data initially. If the user is going to go through the trouble of intentionally dropping the buckets then it seems good enough in the short term to document in the same place Kiali's behavior when the buckets are missing. It's possible though that whoever drops the buckets will not be the same person that is using Kiali which can lead to confusion.

Xunzhuo (approved):
I have reviewed the codes of kiali server, noticed that we added a new api to check if metrics are existed, this part looks good to me, need others go through the kiali ui part. Thanks @jshaughn

jshaughn:
Thanks @Xunzhuo , if you can find time, please look at the kiali.io PR, and the new sections about dropping metrics.  I am very interested to see if this aould be useful to your installation. https://github.com/kiali/kiali.io/pull/565.

frontend/src/components/Metrics/CustomMetrics.tsx
```diff
@@ -264,6 +264,8 @@ class CustomMetrics extends React.Component<Props, MetricsState> {
                   direction={this.state.dashboard?.title || 'dashboard'}
                   labelsSettings={this.state.labelsSettings}
                   hasHistograms={hasHistograms}
+                  hasHistogramsAverage={hasHistograms}
+                  hasHistogramsPercentiles={hasHistograms}
```
jmazzitelli, line 268: I see this looks different than frontend/src/components/Metrics/IstioMetrics.tsx on lines 333-4.. which has:
```
                hasHistogramsAverage={hasHistogramsAverage}
                hasHistogramsPercentiles={hasHistogramsPercentiles}
```
But the code here uses `hasHistograms` for all the keys `hasHistograms`, `hasHistogramsAverage`, and `hasHistogramsPercentiles`... is this expected? Or should u use the same variable names as I show above from the other place in the code that does this?

prometheus/client.go
```diff
@@ -320,6 +320,39 @@ func (in *Client) GetMetricsForLabels(metricNames []string, labelQueryString str
...
+	if err != nil {
+		return nil, errors.NewServiceUnavailable(err.Error())
+	}
+
+	metricsWeAreLookingFor := make(map[string]bool, len(metricNames))
+	for i := 0; i < len(metricNames); i++ {
+		metricsWeAreLookingFor[string(metricNames[i])] = true
+	}
+
+	metricsWeFound := make([]string, 0, 5)
```
jmazzitelli, line 344: I would set the size of this array to the same as the total # we are looking for. Under normal circumstances we are going to get everything we are looking for (only if someone pruned prometheus will it be different, correct?) ... in this case, I would avoid having to reallocate and copy the array as it grows - just set it to the maximum size we expect it could be.
```suggestion
	metricsWeFound := make([]string, 0, len(metricNames))
```

jmazzitelli (commented):
I have two comments - one is a suggested change - not required, but I think it should be done.

frontend/src/components/Metrics/CustomMetrics.tsx
jshaughn: Right, it's correct as-is.  CustomMetrics.tsx is just satisfying the need for new properties.  I could have made them optional properties and defaulted to true, but I wanted to make sure they got set where appropriate.  The metrics we support for crippling are the standard istio histograms, not anything custom.  So, if a custom histogram exists we assume it's in tact, and so we can use the same flag for all three props.

jmazzitelli (commented):
Missing a "." in the docs (`Appending |istio_request_bytes_* to the drop regex`) but once I added that, the graph options are disabled as expected.

image


However, the metrics options are not.

```diff
diff --color _output/istio-1.14.1/samples/addons/prometheus.yaml /tmp/DELETEME/prometheus.yaml 
223a224,229
>       metric_relabel_configs:
>       - action: drop
>         source_labels: [__name__]
>         regex: istio_agent_.*|istiod_.*|istio_build|citadel_.*|galley_.*|pilot_.*|envoy_cluster_[^u].*|envoy_cluster_update.*|envoy_listener_[^dh].*|envoy_server_[^mu].*|envoy_wasm_.*|istio_request_bytes_.*|istio_request_duration_milliseconds_bucket
>       - action: labeldrop
>         regex: chart|destination_app|destination_version|heritage|.*operator.*|istio.*|release|security_istio_io_.*|service_istio_io_.*|sidecar_istio_io_inject|source_app|source_version
```

I still see:

image

jshaughn:
> However, the metrics options are not.

It's still working as expected.  The charts can not disable the options if even one of the applicable histograms still supports the metric.  So unless you drop all of the `_bucket` metrics you will still see the options enabled.  But not every chart will be able to show the info.  Click the info icon... _"No data available" is displayed for a histogram that does not have telemetry supporting the selected option. If no histograms support the necessary telemetry, the option will be disabled._

jmazzitelli (approved):

PR #5256, opened 2022-06-28
Kiosk links from detail pages

lucasponce:
Last step for https://github.com/kiali/kiali/issues/5207

This work adds support for the "parent" kiosk mode to the links used from the details pages.
This PR allows the "parent" of the kiosk to redirect the links to their own console.

Note, all links are supported but the "App" details links are not added on purpose.

The "App" concept is not unified in Kubernetes, so, I've decided to tackle those links in the next iteration.

There is also some tidy in the code, adding some previous feedback from @jshaughn on this.
Now the structure of these links is more clear and we could reorganize the code a little bit better, but this area will be active and it may evolve as soon as we need to introduce new changes.

The main test/review task for this PR is to be sure that there is no regression in the standalone mode, all of these changes should only happen on kiosk mode.


jmazzitelli (approved):
lgtm

israel-hdez:
I did code review and I can't spot anything wrong. I'll test the changes.

aljesusg (dismissed):
LGFM, nothing weird and tested it, need rebase.

lucasponce:
Thanks for the feedback.
I rebased the PR with the master branch.

jmazzitelli (approved):
re-approve after rebase

israel-hdez:
I did some smoke testing and I could find only one instance where apparently it doesn't work as expected:

The Application detail page -> main card, when clicking on a workload while on `kiosk=parent` mode, it does an internal redirection rather than "passing the ball" to the parent.

 Although I'm not sure if that case is still WIP. It looks buggy because for the other kinds and also in the other pages it works OK.

lucasponce:
> The Application detail page -> main card, when clicking on a workload while on kiosk=parent mode, it does an internal redirection rather than "passing the ball" to the parent.

Good point. Yes, the "application" is not yet part of the plugin scope, but as this work should be generic to enable anyone to build a similar integration, so I'll review the missing links, it should be located in the app details card.

Thanks

lucasponce:
Added a commit with the fix that @israel-hdez has spotted.

The workload link in the App Details missed a kiosk link, and now parents can capture the link and redirect to their own logic.
image

This was missed because "apps" are not a unified concept and "parents" can define what to do in these cases.

But now, the kiali side code of the kiosk should be completed.

israel-hdez (approved):
LGTM!

PR #5270, opened 2022-06-29
Login programatically for cypress tests

nrfox:
This replaces the manual flow that required cypress to actually navigate to the login page and type fields into boxes. Not only is this faster, but it eliminates the need for all the hacks around the login page. It is also the only way to login to the perf clusters. Note that the manual login flow is still being tested by the `kiali_login` feature.

### DIFF
--- a/frontend/cypress/README.md
+++ b/frontend/cypress/README.md
@@ -11,11 +11,12 @@ Installed all dev dependencies from frontend folder. Ensure the `baseUrl` field 
 Before you start using Cypress suite, you might need export some environment variables - depending on environment where tests are executed. If your authentication method defaults to `anonymous` **(i.e. dev env), no actions are needed.**
 
 ```bash
-export CYPRESS_BASE_URL=<value>        # defaults to http://localhost:3000
-export CYPRESS_USERNAME=<value>        # defaults to jenkins, opt. kubeadmin
-export CYPRESS_PASSWD=<value>          # no defaults 
-export CYPRESS_AUTH_PROVIDER=<value>   # defaults to my_htpasswd_provider, 
-                                       # or optionally openshift for AWS
+export CYPRESS_BASE_URL=<value>           # defaults to http://localhost:3000
+export CYPRESS_USERNAME=<value>           # defaults to jenkins, opt. kubeadmin
+export CYPRESS_PASSWD=<value>             # no defaults
+export CYPRESS_AUTH_PROVIDER=<value>      # defaults to my_htpasswd_provider,
+export CYPRESS_AUTH_PROVIDER_NAME=<value> # defaults to my_htpasswd_provider. Can pass a custom name for the htp provider.
+                                          # or optionally openshift for AWS
 ```
 
 Tests can be run with the cypress browser:
@@ -93,6 +94,14 @@ You can adjust some inputs of the performance tests by changing the [fixture fil
 ### Results:
 
 Results are logged here: `logs/performance.txt`
+
+### Running on IBM Cloud:
+
+You can use the [perf hack script](../../hack/perf-ibmcloud-openshift.sh) to spin up an openshift cluster on IBMCloud with Kiali + Istio + Kiali demos installed. To run the perf tests against the cluster, you must generate an IBM Cloud API Key and pass that in as the `CYPRESS_PASSWD`.
+
+```
+make -e CYPRESS_BASE_URL="https://<kiali-openshift-route>" -e CYPRESS_PASSWD="<IBMCloud API Key>" -e CYPRESS_USERNAME="IAM#<SSO-EMAIL>" -e CYPRESS_AUTH_PROVIDER="ibmcloud" perf-tests-gui
+```
 
 ## Testing Strategies
 
--- a/frontend/cypress/README.md
+++ b/frontend/cypress/README.md
@@ -11,12 +11,11 @@ Installed all dev dependencies from frontend folder. Ensure the `baseUrl` field 
 Before you start using Cypress suite, you might need export some environment variables - depending on environment where tests are executed. If your authentication method defaults to `anonymous` **(i.e. dev env), no actions are needed.**
 
 ```bash
-export CYPRESS_BASE_URL=<value>           # defaults to http://localhost:3000
-export CYPRESS_USERNAME=<value>           # defaults to jenkins, opt. kubeadmin
-export CYPRESS_PASSWD=<value>             # no defaults
-export CYPRESS_AUTH_PROVIDER=<value>      # defaults to my_htpasswd_provider,
-export CYPRESS_AUTH_PROVIDER_NAME=<value> # defaults to my_htpasswd_provider. Can pass a custom name for the htp provider.
-                                          # or optionally openshift for AWS
+export CYPRESS_BASE_URL=<value>               # defaults to http://localhost:3000
+export CYPRESS_USERNAME=<value>               # defaults to jenkins, opt. kubeadmin
+export CYPRESS_PASSWD=<value>                 # no defaults
+export CYPRESS_AUTH_PROVIDER=<value>          # defaults to my_htpasswd_provider
+export CYPRESS_AUTH_HTP_PROVIDER_NAME=<value> # defaults to my_htpasswd_provider. Can pass a custom name for the htp provider.
 ```
 
 Tests can be run with the cypress browser:
--- a/frontend/cypress/README.md
+++ b/frontend/cypress/README.md
@@ -15,7 +15,6 @@ export CYPRESS_BASE_URL=<value>               # defaults to http://localhost:300
 export CYPRESS_USERNAME=<value>               # defaults to jenkins, opt. kubeadmin
 export CYPRESS_PASSWD=<value>                 # no defaults
 export CYPRESS_AUTH_PROVIDER=<value>          # defaults to my_htpasswd_provider
-export CYPRESS_AUTH_HTP_PROVIDER_NAME=<value> # defaults to my_htpasswd_provider. Can pass a custom name for the htp provider.
 ```
 
 Tests can be run with the cypress browser:
--- a/frontend/cypress/plugins/index.ts
+++ b/frontend/cypress/plugins/index.ts
@@ -19,6 +19,9 @@ const axios = require('axios');
 // eslint-disable-next-line no-unused-vars
 module.exports = (on, config) => {
   config.env.cookie = false;
+  // This name is non-standard and might change based on your environment hence the separate
+  // env variable.
+  config.env.AUTH_HTTP_PROVIDER_NAME = config.env.AUTH_HTTP_PROVIDER_NAME || 'my_htpasswd_provider';
   config.env.AUTH_PROVIDER = config.env.AUTH_PROVIDER || 'my_htpasswd_provider';
 
   async function exportConfig() {
--- a/frontend/cypress/plugins/index.ts
+++ b/frontend/cypress/plugins/index.ts
@@ -21,7 +21,6 @@ module.exports = (on, config) => {
   config.env.cookie = false;
   // This name is non-standard and might change based on your environment hence the separate
   // env variable.
-  config.env.AUTH_HTP_PROVIDER_NAME = config.env.AUTH_HTP_PROVIDER_NAME || 'my_htpasswd_provider';
   config.env.AUTH_PROVIDER = config.env.AUTH_PROVIDER || 'my_htpasswd_provider';
 
   async function exportConfig() {
--- a/frontend/cypress/plugins/index.ts
+++ b/frontend/cypress/plugins/index.ts
@@ -21,7 +21,7 @@ module.exports = (on, config) => {
   config.env.cookie = false;
   // This name is non-standard and might change based on your environment hence the separate
   // env variable.
-  config.env.AUTH_HTTP_PROVIDER_NAME = config.env.AUTH_HTTP_PROVIDER_NAME || 'my_htpasswd_provider';
+  config.env.AUTH_HTP_PROVIDER_NAME = config.env.AUTH_HTP_PROVIDER_NAME || 'my_htpasswd_provider';
   config.env.AUTH_PROVIDER = config.env.AUTH_PROVIDER || 'my_htpasswd_provider';
 
   async function exportConfig() {
--- a/frontend/cypress/plugins/index.ts
+++ b/frontend/cypress/plugins/index.ts
@@ -19,6 +19,8 @@ const axios = require('axios');
 // eslint-disable-next-line no-unused-vars
 module.exports = (on, config) => {
   config.env.cookie = false;
+  config.env.AUTH_PROVIDER = config.env.AUTH_PROVIDER || 'my_htpasswd_provider';
+
   async function exportConfig() {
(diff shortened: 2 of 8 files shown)


frontend/cypress/support/commands.ts
```diff
@@ -59,31 +61,94 @@ Cypress.Commands.add('login', (provider: string, username: string, password: str
...
-      cy.visit('/console/overview?duration=60&refresh=0');
 
       if (auth_strategy === 'openshift') {
-        cy.get('.pf-c-form').contains(auth_strategy, { matchCase: false }).click();
-        cy.get('.pf-c-button').contains(provider, { matchCase: false }).click();
-        cy.get('#inputUsername').clear().type(username);
-        cy.get('#inputPassword').clear().type(password);
-        cy.get('button[type="submit"]').click();
-        cy.wait('@authorized').its('response.statusCode').should('eq', 200);
+        if (provider === 'my_htpasswd_provider') {
```
jmazzitelli, line 66: Use the env var rather than hardcode the IdP name since it could be different:

```suggestion
        if (provider === config.env.KUBEADMIIN_IDP) {
```

nrfox: :+1: I think it should be `AUTH_PROVIDER` from: https://github.com/kiali/kiali/pull/5270/files#diff-8a2033002a0cacda548d12a931922a40b9e4142b8ce4430a6711c92c550c33ccR5.

jmazzitelli: My suggestion is stupid. I was not paying attention.

Somethign like this is what I think we need:

`if (provider contains 'htpasswd') {`

 that's pseudocode - hopefully you get what I mean

nrfox: Added a separate env var so one can provide a custom name if need be.

frontend/cypress/plugins/index.ts
```diff
@@ -11,12 +11,33 @@
...
  */
 // eslint-disable-next-line no-unused-vars
 module.exports = (on, config) => {
-  // `on` is used to hook into various events Cypress emits
-  // `config` is the resolved Cypress config
-}
+  config.env.cookie = false;
+  // This name is non-standard and might change based on your environment hence the separate
+  // env variable.
+  config.env.AUTH_HTTP_PROVIDER_NAME = config.env.AUTH_HTTP_PROVIDER_NAME || 'my_htpasswd_provider';
```
jmazzitelli, line 24: This var name has "HTTP" in it... this doesn't match the doc ... see the [new line 18 of the readme ](https://github.com/kiali/kiali/pull/5270/files#diff-5544425694e0ea24a620263de8dda1c76e42d213916683ff35788135761b3922R18):
```
export CYPRESS_AUTH_PROVIDER_NAME=<value> # defaults to my_htpasswd_provider. Can pass a custom name for the htp provider.
                                          # or optionally openshift for AWS
```

Also - this isn't "HTTP" related.. its "HTPASSWD"

nrfox:
For testing you can follow the instructions here: https://github.com/kiali/kiali/pull/5270/files#diff-5544425694e0ea24a620263de8dda1c76e42d213916683ff35788135761b3922R98-R105

Or if running on CRC you can provide a custom htp provider name with `AUTH_HTTP_PROVIDER_NAME=htprovider`.

frontend/cypress/README.md
```diff
@@ -11,11 +11,12 @@ Installed all dev dependencies from frontend folder. Ensure the `baseUrl` field
...
-export CYPRESS_BASE_URL=<value>        # defaults to http://localhost:3000
-export CYPRESS_USERNAME=<value>        # defaults to jenkins, opt. kubeadmin
-export CYPRESS_PASSWD=<value>          # no defaults 
-export CYPRESS_AUTH_PROVIDER=<value>   # defaults to my_htpasswd_provider, 
-                                       # or optionally openshift for AWS
+export CYPRESS_BASE_URL=<value>           # defaults to http://localhost:3000
+export CYPRESS_USERNAME=<value>           # defaults to jenkins, opt. kubeadmin
+export CYPRESS_PASSWD=<value>             # no defaults
+export CYPRESS_AUTH_PROVIDER=<value>      # defaults to my_htpasswd_provider,
+export CYPRESS_AUTH_PROVIDER_NAME=<value> # defaults to my_htpasswd_provider. Can pass a custom name for the htp provider.
```
jmazzitelli, line 18: See my previous comment. This name doesn't match the code.

jmazzitelli (changes requested):

jmazzitelli:
I can't get this to work:

```make -e CYPRESS_AUTH_PROVIDER=htpasswd -e CYPRESS_HTTP_AUTH_PROVIDER_NAME=htpasswd -e CYPRESS_PASSWD=kiali -e CYPRESS_USERNAME=kiali -e CYPRESS_BASE_URL=https://kiali-istio-system.apps.kialici.ABC.com cypress-gui
```

frontend/cypress/plugins/index.ts
nrfox: Whoops I think my brain must've mixed the two up while typing this. I'll update to be `HTP`.

frontend/cypress/README.md
nrfox: > I can't get this to work:
> 
> ```
> make -e CYPRESS_AUTH_PROVIDER=htpasswd -e CYPRESS_AUTH_HTTP_PROVIDER_NAME=htpasswd -e CYPRESS_PASSWD=kiali -e CYPRESS_USERNAME=kiali -e CYPRESS_BASE_URL=https://kiali-istio-system.apps.kialici.ABC.com cypress-gui
> ```

It should now be:
```
make -e CYPRESS_BASE_URL="https://kiali-istio-system.apps-crc.testing" -e CYPRESS_PASSWD="PW" CYPRESS_USERNAME="kubeadmin" CYPRESS_AUTH_PROVIDER="anyauth" CYPRESS_AUTH_HTP_PROVIDER_NAME="anyauth" cypress-gui
```
I tested this on my CRC cluster and it worked fine.

nrfox: > Also, I'm confused. Why are both "my_htpasswd_provider"? Shouldn't it be:
> `CYPRESS_AUTH_PROVIDER=(openshift|aws|htpasswd)`
> and if it is htpasswd THEN you use CYPRESS_AUTH_PROVIDER_NAME which defaults to my_htpasswd_provider but can be any custom name you happen to use for your htpasswd provider???

Yeah... thinking about this more it might make more sense just to have a boolean here and not even worry about the actual name. `CYPRESS_HTP_PROVIDER=true` or something. Because the name is non standard for the provider it _can_ be anything, right?

ScriptingShrimp:
tested following scenarios - ordered:
1. tested on OCP with default CYPRESS_AUTH_PROVIDER ( my_htpasswd_provider + default user) :heavy_check_mark: 
2. tested on OCP with `export CYPRESS_AUTH_PROVIDER=kube:admin`  :heavy_check_mark: 
3. tested on OCP with `export CYPRESS_AUTH_HTP_PROVIDER_NAME=my_htpasswd_provide` - I was hoping this will overwrite `CYPRESS_AUTH_PROVIDER` which I set befor, but I got same result as in test 2 (OCP logged in with kube:admin)

frontend/cypress/README.md
ScriptingShrimp: can we fetch this provider name somwhere from API?

jmazzitelli:
cc @ScriptingShrimp ... @nrfox is going to change this a bit. The CYPRESS_AUTH_PROVIDER is going to be one of "ibmcloud" or "htpasswd" (or "aws" in the future). with the CYPRESS_AUTH_HTP_PROVIDER_NAME being the name of the htpasswd provider for when you set CYPRESS_AUTH_PROVIDER to "htpasswd".

The "kube:admin" provider is something I don't think we expected to work. @nrfox are you aware of the use case that uses the provider called "kube:admin" ??? That's a new one to me.

frontend/cypress/README.md
jmazzitelli: You can have multiple providers in OpenShift. So you could look at the OAuth object at the identityProviders, but there could in theory be more than one (and that is why we see buttons in the login form - you need to tell OpenShift which provider you are authenticating with). In this case, we need to tell the Cypress stuff which provider needs to authenticate the credentials we are passing in.

ScriptingShrimp:
Thinking out loud, if you need some sort of new login routine, introduce new env variable as a switch for request based autohization and override that way default settings. It's not braking change and we don't need to update our private pipelines

jmazzitelli:
We aren't breaking. In fact, Nick and I just had a discussion earlier and one point we made to each other was we didn't want to make QE change anything in the pipeline.

Can you show me anywhere in the Cypress code where we actually look at "kube:admin" for PROVIDER? There is no code that I can see what does anything with a value that is kube:admin. I suspect it just "accidentially" works. And probably still works - but I can say that kube:admin value seems to be completely ignored .

ScriptingShrimp:
@jmazzitelli 
image

jmazzitelli:
> @jmazzitelli image

Right - that's the name of the button, not the provider name. So this is using the htpasswd provider.

This is news to me and I bet news to Nick that you are passing in "kube:admin" as the provider.

cc @nrfox here's something interesting we need to take into account - perhaps this is just another PROVIDER type (in addition to "htpasswd" we'll want "kube:admin" and if the provider type is "kube:admin" then we press the button with that name.

ScriptingShrimp:
If you set i.e. `export CYPRESS_AUTH_PROVIDER=kube:admin` it gets propagated into code below as `KUBEADMIN_IDP`. Cypress will search for a button with this value and will click that button 
https://github.com/kiali/kiali/blob/4953758c752a806a4803db8e48a7e1b485742c6f/frontend/cypress/integration/common/kiali_login.ts#L32

jmazzitelli:
@nrfox here's some additional details as to how OpenShift IdPs work.

First, the "kube:admin" is a basic, built-in provider. It is NOT specified in the OAuth object. Rather, by the mere existence of a secret called `kubeadmin` in the `kube-system` namespace, then the kube:admin "provider" is enabled. If there is no other IdPs installed, then you will just get a username/password form in which you enter the kubeadmin credentials. If there is another IdP installed, you will see a button labeled "kube:admin". Thus, I think the Cypress stuff can have a value "kube:admin" for the CYPRESS_AUTH_PROVIDER - if specified, we'll assume the username is "kubeadmin" and we'll assume the button is called "kube:admin"

Next, for other IdPs, they are inside the OAuth cluster object. Here's mine in which I have an htpasswd IdP installed:

```
oc get OAuth cluster -o jsonpath={.spec.identityProviders} | jq
[
  {
    "htpasswd": {
      "fileData": {
        "name": "htpasswd"
      }
    },
    "mappingMethod": "claim",
    "name": "mazzhtpasswd",
    "type": "HTPasswd"
  }
]
```

This will show a button called "mazzhtpasswd" - you enter credentials for any user in that htpasswd data file.

If there is no other IdP installed (and no kube:admin secret) you won't see a button page - just the username/password form.

jmazzitelli:
We need to know if the OpenShift login process will show you a page of buttons (to select an IdP) or not. The page of buttons will show if you have 2 or more IdPs installed (including kube:admin). This is how you know if a button needs to be pressed:

Run this command `oc get OAuth cluster -o jsonpath={.spec.identityProviders}` and count the number of items in the list that it returns. Then, to that number, add 1 if `oc get secret kubeadmin -n kube-system` returns with exit code of 0 (i.e. the secret exists).

If the resulting value is > 1, you know there will be a page with buttons that the Cypress tests need to press. If the value is 1, then you know you will not have buttons and you will go directly to the username/password form.

If the value is 0, then you are screwed and you can't log in :) I don't think that will ever happen. There has to be at least one IdP installed to be able to log in right? Anyway, that won't happen in any of our test environments.

jmazzitelli:
> If you set i.e. `export CYPRESS_AUTH_PROVIDER=kube:admin` it gets propagated into code below as `KUBEADMIN_IDP`. Cypress will search for a button with this value and will click that button

Right. So there is no special handling for "kube:admin" - its just "the name of the button" like anything else. But of course if you ONLY have kube:admin and no other IdP installed, that won't work. This is what we are trying to fix here. Plus, we also need to support ibmcloud clusters and CRC clusters - which is the other thing we are trying to fix.

jmazzitelli:
> `oc get OAuth cluster -o jsonpath={.spec.identityProviders}` can we get similar data from REST API? I don't want to hard wire Cypress to `oc` dependency

Right. I only mentioned "oc" as kind of "pseudo code" -- however we can get this information is fine, as long as we can. We just need to get the "OAuth" object (it is a cluster-scoped object) and look at its `spec.identityProviders` list - count the items in the list. Then we just need to see if there is a secret called "kubeadmin" in kube-system namespace - however we do that.

ScriptingShrimp:
Fetching this sort of data form any API would be the best solution in my opinion. Lets try to refactor/address this it in different PR, since I don't want to block this PR. 

Right now we run whole regression suite on each platform. `cy.login()` should be working the _UI way_ one every platform, because it is smoke test. I don't have finite list of all platforms and providers - shall we create ticket/issue about that?

nrfox:
> Fetching this sort of data form any API would be the best solution in my opinion. Lets try to refactor/address this it in different PR, since I don't want to block this PR.

That would be great if we could figure out how to do that but it doesn't seem like openshift supports this for the htpasswd provider at least. The docs say that you can pass a `WWW-Authenticate` header and the login flow should be different but I was never able to get this working: https://docs.openshift.com/container-platform/4.9/authentication/understanding-authentication.html#oauth-token-requests_understanding-authentication.

> Right now we run whole regression suite on each platform. cy.login() should be working the UI way one every platform, because it is smoke test. I don't have finite list of all platforms and providers - shall we create ticket/issue about that?

The `kiali_login` test still runs on every platform/provider but with this change the UI way will only be tested by the `kiali_login` test instead of before each test.

In short, everything that was working before this change should keep working after this change. Passing `CYPRESS_AUTH_PROVIDER=kube:admin` should still work and the default will remain `my_htpasswd_provider` so the upstream pipelines shouldn't need to be changed either.

frontend/cypress/support/commands.ts
nrfox: Removed the separate env var. Now `CYPRESS_AUTH_PROVIDER` other than `ibmcloud` is assumed to be `htpasswd` and the button that is searched for during login will be whatever `CYPRESS_AUTH_PROVIDER` is set to e.g. `my_htpasswd_provider` or `kube:admin`.

frontend/cypress/README.md
nrfox: Now it just uses whatever you passed in to `CYPRESS_AUTH_PROVIDER` and assumes it follows the `htpasswd` flow. Works even when multiple providers are present.

jmazzitelli (approved):
I'm going to approve this with a caveat.

Running cypress-gui I can see the tests are passing for me EXCEPT the kiali_login test. I can see it "press the IdP button" and type in my user/pass and press the login button, but it fails to login . I think there is just something wrong with my cluster (I had this kind of problem yesterday when I was playing around with IdPs). I can login fine now, so I don't know why this test can't, but I will assume it is just flakiness in my cluster rather than a problem with this PR.

Issue #5269, opened 2022-06-29
Remove invalid durations based on prometheus scrape interval


jshaughn:
When the prometheus scrape interval is greater than 30s, selecting the "1m" duration option will result in an empty graph (unless the Idle Edges Display option is enabled).  The reason is that the graph is based on traffic, and it takes at least two data points to generate a positive request rate.  If the selected duration is less than twice the scrape interval, there will be at most one data point for the time period, and therefore no positive traffic request rate.

In this case Kiali should remove the invalid duration option.

A FAQ entry should also be part of this issue resolution.

PR #5271, opened 2022-06-29
Improve Duration/Refresh handling

jshaughn:
- remove invalid durations based on globalScrapeInterval (duration < 2 x scrape)
- add 2m duration to choices.  This is good when 1m is invalid, and may
  be useful even when 1m is available.
- fix bug in toValidDuration util
- change default refreshInterval from 15s to 1m, 15s is too frequent, especially
  in non-trivial installs.  1m may also be too small, but it's better.

Part-of https://github.com/kiali/kiali/issues/5269

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -687,7 +687,7 @@ func NewConfig() (c *Config) {
 				MetricsOutbound:   MetricsDefaults{},
 				MetricsPerRefresh: "1m",
 				Namespaces:        make([]string, 0),
-				RefreshInterval:   "15s",
+				RefreshInterval:   "60s",
 			},
 			Validations: Validations{
 				Ignore: make([]string, 0),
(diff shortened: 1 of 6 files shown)


config/config.go
```diff
@@ -687,7 +687,7 @@ func NewConfig() (c *Config) {
 				MetricsOutbound:   MetricsDefaults{},
 				MetricsPerRefresh: "1m",
 				Namespaces:        make([]string, 0),
-				RefreshInterval:   "15s",
+				RefreshInterval:   "60s",
```
jmazzitelli, line 690: I will assume an operator PR is coming that will update the default there:
https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L274

jshaughn: yes, and also a FAQ....

jmazzitelli:
When I bring up Kiali, this is the defaults on the graph page:

image

I see this in the ConfigMap:

```
    ui_defaults:
      graph:
        metrics_per_refresh: 1m
        namespaces: []
        refresh_interval: 60s
```

jmazzitelli (approved):

Xunzhuo (approved):
Looks good, thanks!

jmazzitelli (approved):
i will approve on the condition the cypress tests go green.

Issue #5275, opened 2022-06-30
mTLS traffic should be included in the Traffic tab


lucasponce:
This task should continue the enhancements on the mTLS information in the details pages.

Now, Istio metrics have a "Connection Security Policy" filter that allow to easily visualize that info:
![image](https://user-images.githubusercontent.com/1662329/176672686-b430d2f3-87d9-424f-a6a6-7b74b3cc4007.png)

Minigraph has information about the mTLS edges:
![image](https://user-images.githubusercontent.com/1662329/176672515-bd74e459-5630-4e11-8575-dd7dea703eed.png)

But the Traffic tab doesn't show that information, and mTLS edges are one of the first aspects that users would like to see
![image](https://user-images.githubusercontent.com/1662329/176672553-fa513ba9-5368-4fcf-93a1-8d1444e53e7e.png)

As the backend data for minigraph and the traffic data is the same, this information should be available.

josunect:
How can this information be shown? I can think in:

- Add a lock icon if mTLS is enabled || Add a new column, mTLS, with the percentage of mTLS traffic (But if will be blank if mTLS is disabled)
- The table has enough space to add, near the icon, the percentage of mTLS traffic || Can be shown when hover over the icon

PR #5280, opened 2022-07-01
Add lock icon if mTLS is enabled in traffic tab table

josunect:
Added mTLS information in the Traffic tab:

- Added lock icon if mTLS is enabled. This is less intrusive than use another column. 
- Added percentage of mTLS traffic in a tooltip

Screenshot from 2022-07-01 13-36-42

Fixes https://github.com/kiali/kiali/issues/5275

### DIFF
--- a/frontend/src/components/TrafficList/TrafficDetails.tsx
+++ b/frontend/src/components/TrafficList/TrafficDetails.tsx
@@ -60,6 +60,7 @@ export interface TrafficItem {
   node: TrafficNode;
   proxy?: TrafficItem;
   traffic: ProtocolTraffic;
+  mTLS?: number;
 }
 
 type TrafficDetailsProps = {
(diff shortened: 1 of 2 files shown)


lucasponce:
I'm testing locally, but I don't see the new icon:
image

But in the traffic tab:
image

As it happens in the "minigraph" I'd add the lock icon always, showing the two states, open (no mTLS) closed (mTLS).

As this info is very common and it would help the users to easily identify which traffic is what.

I'm going to re-build the branch to see if it's on my side why I don't see the expected behavior.

josunect:
Sorry, @lucasponce , I was too fast and forgot to add the mTLS parameters to also Workloads and services. 

The Traffic component needs to add the securityPolicy parameter in the request in order to have this information available - And this is done for each type, separately. 

Added the open lock icon when mTLS is disabled.

israel-hdez (approved):
It works OK.

I saw one issue, but I'm quite sure it's patternfly, so I'm still approving. The issue is that when the icon transitions from the closed lock to the open lock (or vice-versa) the tooltip is lost (it will no longer appear). However, from the code, I cannot understand why that would happen.

lucasponce (approved):

PR #5284, opened 2022-07-04
Metrics tab: Ability to show both reporters

### LINKED ISSUE #2887: Show network latencies in graph

israel-hdez:
In the metrics tabs, in the _Reporter_ drop-down a third _"Both"_ option is added. This allows to see data from both reporters in the charts. This is how it will look:

image

It can be appreciated that when both reporters are requested, the generated charts will be _area charts_. The solid lines still represent the actual data coming from the reporters and series are now arranged in pairs: lines that belong to the same serie where only the reporter varies, are associated with a filled area and are given the same color. This should make easier to see data differences between reporters which may tell, for example network latency or could be understood as triggering points for circuit breakers.

Fixes #2887

### DIFF
--- a/business/metrics.go
+++ b/business/metrics.go
@@ -28,7 +28,9 @@ func (in *MetricsService) GetMetrics(q models.IstioMetricsQuery, scaler func(n s
 
 func createMetricsLabelsBuilder(q *models.IstioMetricsQuery) *MetricsLabelsBuilder {
 	lb := NewMetricsLabelsBuilder(q.Direction)
-	lb.Reporter(q.Reporter)
+	if q.Reporter != "both" {
+		lb.Reporter(q.Reporter)
+	}
 
 	namespaceSet := false
 
--- a/frontend/src/components/Charts/KChart.tsx
+++ b/frontend/src/components/Charts/KChart.tsx
@@ -156,7 +156,13 @@ class KChart<T extends LineInfo> extends React.Component<KChartProps<T>, State> 
 
   private determineChartType() {
     if (this.props.chart.chartType === undefined) {
-      return this.props.chart.xAxis === 'series' ? barInfo : lineInfo;
+      if (this.props.chart.xAxis === 'series') {
+        return barInfo;
+      } else if (this.props.chart.metrics.some(m => m.datapoints.some(dp => dp[2]))) {
+        return areaInfo;
+      } else {
+        return lineInfo;
+      }
     }
     const chartType = this.props.chart.chartType;
     switch (chartType) {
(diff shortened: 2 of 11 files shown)


israel-hdez:
@lucasponce It's almost done...

I have the open question about what to do with the trendlines?
Right now, when both reporters are selected, trendlines are generated from data of the "source" but I think this is not fair. Not sure if trendlines should be disabled, or perhaps using the average value between both related reporters? 

We could also generate trendlines for each reporter, but I think the charts will become very crowded.

lucasponce:
Thanks, @israel-hdez, the work looks great :).

> Not sure if trendlines should be disabled, or perhaps using the average value between both related reporters?

About trendlines, I think using the average, in this case, sounds useful. Trendlines is a "helper" tool that may add some information on cases where perhaps the user doesn't notice a "trend" in the data series, but as it's not an accurate tool, so, it's expected that some gap is present.

I see very useful the trendlines, especially when I zoom a metric, and it's not fully clear the movement of the series.
(In some cases, it's just redundant, but that's optional for the user).

lucasponce:
I took a quick look and, from the feature perspective, it works as expected:
image

And it opens more future possibilities to not show the data but give more insights, i.e. with these couple of captures:
image
image

The user can see the "impact" of the envoy algorithms in terms of latency.

So having these sets of toolings helps to see in a single step all the info I'd need.

I haven't yet reviewed code or regression, feel free to ping me when you'd like to run a formal review.
Thanks

israel-hdez:
@lucasponce This is now ready for proper review.

lucasponce:
Thanks, @israel-hdez, it works fine and the code looks good to me.

This feature is adding a nice new capability to Kiali.

In healthy scenarios like this one:
image

The combined reporters can give visibility to the user about the latency between the observed telemetry in source/destination, typically that may offer insights about a global "overhead" of the proxies:
image

(The less latency/overhead, the differences should be minimal).

But when there are error conditions on the traffic, like this example:
image

The observed telemetry between source and destination are different because envoy algorithms are working and marking a destination as "unhealthy" by applying a circuit breaker pattern. The details about how the "source" proxy responds in that situation are more clear in this capture:
image

lucasponce (approved):

Issue #5276, opened 2022-06-30
Adjust mouse pointer on areas that user can navigate/jump to


lucasponce:
Kiali has some areas in the graph (navigation or zoom on a node/edge) and charts (jump to a trace) that would require a minor change in the mouse pointer, to use the same icon to indicate to users that there is a "clickable" action.

In integration scenarios, this lack of pointer help is more relevant as it's the default behavior with other areas.

jshaughn:
I have submitted a PR for the graph portion of this issue.

PR #5285, opened 2022-07-05
Adjust mouse pointer in charts for traces

lucasponce:
Fixes https://github.com/kiali/kiali/issues/5276

Follow-up work to change the mouse pointer when:
- "Show spans" is enabled in metrics.
- "Traces" main chart.


israel-hdez (approved):
I tested and it looks fine here. The code is very straightforward!.

Issue #5263, opened 2022-06-29
Mesh wide settings should take the MeshConfig defaults


lucasponce:
Kiali detects a mesh wide TLS setup:
![image](https://user-images.githubusercontent.com/1662329/176436949-da46ffb6-43ac-4e1e-bae2-0e129a1248c8.png)

But only when there is an explicit DestinationRule in the "istio-system" (or any other control plane namespace).

![image](https://user-images.githubusercontent.com/1662329/176437414-7ea2dc78-5f97-464b-8178-48c2d43a5d1e.png)

This logic should be revisited and considered to get the Mesh defaults and inform the user about the default mode in addition to any particular configuration that a user may perform on a specific namespace.

This issue may involve a review of some TLS validations, the current logic is valid but it should be completed with the mesh defaults.

Related to this issue, the KIA0204, KIA0205, KIA0206, KIA0401, KIA0501 validations don't seem to work well, but I'd try to address the whole TLS logic is related here:

https://kiali.io/docs/features/validations/#kia0204---mtls-settings-of-a-non-local-destination-rule-are-overridden
https://kiali.io/docs/features/validations/#kia0205---peerauthentication-enabling-mtls-at-mesh-level-is-missing
https://kiali.io/docs/features/validations/#kia0206---peerauthentication-enabling-namespace-wide-mtls-is-missing
https://kiali.io/docs/features/validations/#kia0401---mesh-wide-destination-rule-enabling-mtls-is-missing
https://kiali.io/docs/features/validations/#kia0501---destination-rule-enabling-namespace-wide-mtls-is-missing

As part of this work, the error level of these validations can be revisited (if it makes sense to keep them on "error" or just to relax the level to "warning").

Other validations are outdated, related to https://github.com/kiali/kiali/issues/5262 but it may need some review in the backend logic as well:
https://kiali.io/docs/features/validations/#kia0207---peerauthentication-with-tls-strict-mode-found-it-should-be-permissive

Subtasks:
- [x] Show info about extended mtls from mesh level in namespaces.
- [x] Show "autoMTLS true" mode info in toolbar lock icon tooltip.
- [x] Revisit existing mTLS validations while combining them with default mtls mode.

hhovsepy:
Hi @lucasponce , can you please clarify what is wrong with the validations mentioned above? Are they not aligned with Istio?

lucasponce:
I think this issue can be used to revisit this logic.
The goal here is that the user should not assume any default.

[...]

PR #5296, opened 2022-07-08
Added new option for namespace's mTLS enabled extended from Mesh-wide

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5263

In Overview page, while showing mTLS lock icon on namespaces, where mTLS is enabled not for namespace but for Mesh-wide, it will show now a message in lock's tooltip that mTLS status is extended from Mesh-wide.

Screenshot from 2022-07-08 17-00-22


leandroberetta:
> Hi @leandroberetta can you take a look into this PR, it seems we work in same area?

Yes, sorry, I totally missed this.

leandroberetta (approved):
This LGTM.

PR #5306, opened 2022-07-15
npm cypress scripts fix

ScriptingShrimp:
reverting changes in npm scripts so cypress test suite is triggered correctly in CI


mattmahoneyrh (approved):

nrfox:
@ScriptingShrimp @mattmahoneyrh iirc the reason the cypress run script is setup this way is so that the validations are run _after_ all the other tests because the some of the validation tests cause cluster wide issues that make other tests unstable. It looks like CI did not pass for this PR which could be caused by this change.

PR #5314, opened 2022-07-19
Add padding to charts equal to largest data size

### LINKED ISSUE #5222: [CI] Test flake - timeout looking for traces tab
Originally found here: https://github.com/kiali/kiali/pull/5141#discussion_r891638062.
Screenshot from 2022-06-07 15-16-20

> You can see from the screenshot that the traces have loaded and it's clicking the right thing but it's still failing. I think the trace is outside of the chart and clicking anywhere outside the chart doesn't do anything. I can reproduce this outside this test as well. In other words, it's kind of a bug since all the data on the chart should be clickable and rendered inside the chart. Not sure if this is a problem with Kiali or with the chart lib.

nrfox:
This ensures that the clickable area of the chart encompasses the largest point. Otherwise the points on the edge of the graph can't be clicked.

fixes #5222

### DIFF
--- a/frontend/src/components/Charts/ChartWithLegend.tsx
+++ b/frontend/src/components/Charts/ChartWithLegend.tsx
@@ -144,18 +144,10 @@ class ChartWithLegend<T extends RichDataPoint, O extends LineInfo> extends React
     const overlayIdx = this.props.data.length;
     const showOverlay = (this.props.overlay && this.props.showSpans) || false;
     const overlayRightPadding = showOverlay ? 15 : 0;
-    let largestSize = 0;
-    this.props.data.forEach(dataSet => {
-      dataSet.datapoints.forEach(dp => {
-        if (dp.size !== undefined) {
-          largestSize = Math.max(largestSize, dp.size);
-        }
-      });
-    });
     const padding: Padding = {
-      top: largestSize,
+      top: 0,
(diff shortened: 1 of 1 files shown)


israel-hdez (approved):
I can notice the added padding by looking closely.
I still see that dots at the very left are difficult to click. So, I guess only time will tell if this fixes the Cypress issue (i.e. if it doesn't fail again)

nrfox:
@israel-hdez this is a slightly different approach that uses `domainPadding` instead of padding for the chart itself. Basically it pads the data points on the chart so that they get squeezed into the chart and away from the edges. The edges are still not clickable but the points don't reach the edges anymore. This changes where the points lie on the chart but doesn't mess with the outside padding.

Screenshot from 2022-07-25 14-20-05

As a side note, the charts should be taking the padding into account and resizing accordingly instead of having the edges unclickable but after some investigation I wasn't able to figure out exactly why the charts aren't resizing appropriately. Instead just opting for a hack here to get around the edges being unclickable.

israel-hdez (approved):
LG!
You added a comment that this is a hack, but I find this much more elegant than the previous approach

nrfox:
> LG! You added a comment that this is a hack, but I find this much more elegant than the previous approach

Well the real fix would've been to figure out why the graphs are not resizing properly even in the presence of padding but I agree that this workaround is better than the previous :)

Issue #5310, opened 2022-07-15
Sidecar with no workloadSelector in two separate namespaces are marked as conflicting...


diranged:
We have a`istio-system/default` Sidecar with no workloadSelector, and then the majority of our other apps use `Sidecar` resources that are scopd to their workload (and also launched in their namespace). We have one other namespace though where the `Sidecar` has no workload selector in it and that is causing Kiali to throw an error `KIA0002` about two conflicting resources ... but they don't conflict, because they are in different namespaces:

![image](https://user-images.githubusercontent.com/768067/179305580-8a70c1df-345e-4e5d-b4c2-e97679d2ea7d.png)
![image](https://user-images.githubusercontent.com/768067/179305598-60c1ccf0-a095-43af-a783-474530cf447b.png)

Slack Thread: https://istio.slack.com/archives/CJVRXB09F/p1657895790021149

hhovsepy:
Thank you @diranged for logging this issue, the fix will be included in upcoming Kiali v1.55 release.

diranged:
Thank you!

PR #5316, opened 2022-07-19
Consider objects from own namespace only in workload selector/less checkers.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/5310
When checking "KIA0002 More than one selector-less object in the same namespace" and "KIA0003 More than one object applied to the same workload" for Sidecars, PeerAuthentications and RequestAuthentiations, it should show validation reference objects to only local namespace objects. Cross namespace validations are not supported for these object types.

Before: Sidecars, PeerAuths and RequestAuths validations of workload selectors "KIA0003" and "KIA0002" were considering objects from all namespaces.
Screenshot from 2022-07-19 14-43-24
Screenshot from 2022-07-19 14-40-49

Now: As those objects are not supporting cross-namespace validations, consider objects only from own namespace.
Screenshot from 2022-07-19 14-41-35
Screenshot from 2022-07-19 14-43-48

### DIFF
--- a/business/checkers/common/multi_match_selector_checker_test.go
+++ b/business/checkers/common/multi_match_selector_checker_test.go
@@ -36,12 +36,28 @@ func TestTwoSidecarsWithoutSelector(t *testing.T) {
 		[]*networking_v1beta1.Sidecar{
 			data.CreateSidecar("sidecar1", "bookinfo"),
 			data.CreateSidecar("sidecar2", "bookinfo"),
+			data.CreateSidecar("sidecar3", "bookinfo2"),
 		},
 		workloadList(),
 	).Check()
 
 	assertMultimatchFailure(t, "generic.multimatch.selectorless", validations, "sidecar1", []string{"sidecar2"})
 	assertMultimatchFailure(t, "generic.multimatch.selectorless", validations, "sidecar2", []string{"sidecar1"})
+}
+
+func TestTwoSidecarsWithoutSelectorDifferentNamespaces(t *testing.T) {
+	assert := assert.New(t)
+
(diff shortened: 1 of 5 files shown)


jshaughn:
cc @israel-hdez looks like this one is still pending a review.

business/checkers/common/multi_match_selector_checker_test.go
```diff
@@ -111,15 +127,15 @@ func TestSidecarsDifferentNamespaces(t *testing.T) {
...
 		}, data.CreateSidecar("sidecar2", "bookinfo2")),
 		data.AddSelectorToSidecar(map[string]string{
 			"app":     "details",
-			"version": "v1",
+			"version": "v3",
 		}, data.CreateSidecar("sidecar3", "bookinfo3")),
 		data.AddSelectorToSidecar(map[string]string{
 			"app":     "details",
-			"version": "v1",
+			"version": "v4",
```
israel-hdez, line 138: Nitpick: if I understand correctly, `v4` does not exist in the list returned by `workloadList()`...
It is a very small nit. It does not affect the tests. So, take this as a simple comment.

So, no action needed.

If you do want to fix, perhaps do it on another PR as a side change (even if it is an unrelated change). Let's just merge this as is.

israel-hdez (approved):
Sorry for the delay.

Looks good!

business/checkers/common/multi_match_selector_checker_test.go
hhovsepy: actually I did it on purpose, to make sure that tests will fail if future changes cause bug

hhovsepy:
thanks @israel-hdez

Issue #5179, opened 2022-06-09
Automate the generation of the Istio model in the frontend code


lucasponce:
The frontend layer maintains an IstioObjects.ts file to map the Istio backend with the model used in the frontend (mainly wizards and Istio Config editor).

As we moved to the istio.io/api in the backend, all these types have a formal proto buffer definition; it could be nice to perform a minimal refactoring here, then every time that we change the istio.io/api, we could recreate this mapping automatically.

What this issue expects is a helper hack script that could parse the types and generate the *.ts files in the frontend used for the IstioObject.ts. Perhaps that would trigger a minimal reorganization of those files.

Related to https://github.com/kiali/kiali/issues/5159

Issue #4783, opened 2022-03-03
Improve the API. Add types for queryparams and promises


aljesusg:
In the UI the are a lot of any types for params and options to  request to the backend. Also the promises returned are not defined.

Make the types for these options in /api/services

lucasponce:
Related to https://github.com/kiali/kiali/issues/5179

jshaughn:
We have added a lot of type safety since this was created.  Closing as complete.

PR #5319, opened 2022-07-20
Cluster-scoped cache

### LINKED ISSUE #4678: Optimize the Kiali Cache design under cluster rights presence

nrfox:
Utilizes a cluster scoped cache if the Kiali service account has cluster-scoped permissions for each of the cache types.

Fixes #4678

### DIFF
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -6,11 +6,16 @@ import (
 	"sync"
 	"time"
 
-	kube "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
-	"k8s.io/client-go/tools/cache"
-
 	istio "istio.io/client-go/pkg/clientset/versioned"
+	istioext_v1alpha1_listers "istio.io/client-go/pkg/listers/extensions/v1alpha1"
+	istionet_v1alpha3_listers "istio.io/client-go/pkg/listers/networking/v1alpha3"
+	istionet_v1beta1_listers "istio.io/client-go/pkg/listers/networking/v1beta1"
+	istiosec_v1beta1_listers "istio.io/client-go/pkg/listers/security/v1beta1"
+	istiotelem_v1alpha1_listers "istio.io/client-go/pkg/listers/telemetry/v1alpha1"
+	kube "k8s.io/client-go/kubernetes"
+	apps_v1_listers "k8s.io/client-go/listers/apps/v1"
+	core_v1_listers "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/rest"
 
 	kialiConfig "github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
@@ -28,8 +33,10 @@ type (
 		// Check if a namespace is listed to be cached; if yes, creates a cache for that namespace
 		CheckNamespace(namespace string) bool
 
-		// Clear a namespace's cache
-		RefreshNamespace(namespace string)
+		// Refresh will recreate the necessary cache. If the cache is cluster-scoped the "namespace" argument
+		// is ignored and the whole cache is recreated, otherwise only the namespace-specific cache is updated.
+		Refresh(namespace string)
+
 		// Stop all caches
 		Stop()
 
@@ -43,10 +50,6 @@ type (
 		RegistryStatusCache
 	}
 
-	// This map will store Informers per specific types
-	// i.e. map["Deployment"], map["Service"]
-	typeCache map[string]cache.SharedIndexInformer
-
 	namespaceCache struct {
 		created       time.Time
 		namespaces    []models.Namespace
@@ -59,15 +62,46 @@ type (
 		proxyStatus *kubernetes.ProxyStatus
 	}
 
+	// cacheLister combines a bunch of lister types into one.
+	// This can probably be simplified or turned into an interface
+	// with go generics.
+	cacheLister struct {
+		// Kube listers
+		configMapLister   core_v1_listers.ConfigMapLister
+		daemonSetLister   apps_v1_listers.DaemonSetLister
+		deploymentLister  apps_v1_listers.DeploymentLister
+		endpointLister    core_v1_listers.EndpointsLister
+		podLister         core_v1_listers.PodLister
+		replicaSetLister  apps_v1_listers.ReplicaSetLister
+		serviceLister     core_v1_listers.ServiceLister
+		statefulSetLister apps_v1_listers.StatefulSetLister
+
+		// Istio listers
+		authzLister           istiosec_v1beta1_listers.AuthorizationPolicyLister
+		destinationRuleLister istionet_v1beta1_listers.DestinationRuleLister
+		envoyFilterLister     istionet_v1alpha3_listers.EnvoyFilterLister
+		gatewayLister         istionet_v1beta1_listers.GatewayLister
+		peerAuthnLister       istiosec_v1beta1_listers.PeerAuthenticationLister
+		requestAuthnLister    istiosec_v1beta1_listers.RequestAuthenticationLister
+		serviceEntryLister    istionet_v1beta1_listers.ServiceEntryLister
+		sidecarLister         istionet_v1beta1_listers.SidecarLister
+		telemetryLister       istiotelem_v1alpha1_listers.TelemetryLister
+		virtualServiceLister  istionet_v1beta1_listers.VirtualServiceLister
+		wasmPluginLister      istioext_v1alpha1_listers.WasmPluginLister
+		workloadEntryLister   istionet_v1beta1_listers.WorkloadEntryLister
+		workloadGroupLister   istionet_v1beta1_listers.WorkloadGroupLister
+	}
+
 	kialiCacheImpl struct {
+		clusterScoped          bool // Creates either cluster-scoped or namespace-scoped informers
 		istioClient            kubernetes.K8SClient
 		k8sApi                 kube.Interface
 		istioApi               istio.Interface
 		refreshDuration        time.Duration
 		cacheNamespacesRegexps []regexp.Regexp
 		cacheIstioTypes        map[string]bool
-		stopChan               map[string]chan struct{}
-		nsCache                map[string]typeCache
+		stopNSChans            map[string]chan struct{}
+		stopClusterScopedChan  chan struct{} // Close this channel to stop the cluster-scoped informers.
 		cacheLock              sync.RWMutex
 		tokenLock              sync.RWMutex
 		tokenNamespaces        map[string]namespaceCache
@@ -79,11 +113,16 @@ type (
 		registryStatusLock     sync.RWMutex
 		registryStatusCreated  *time.Time
 		registryStatus         *kubernetes.RegistryStatus
-		stopCacheChan          chan bool
+		// Stops the background goroutine which refreshes the cache's
+		// service account token.
+		stopCacheChan chan bool
+
+		clusterCacheLister *cacheLister
+		nsCacheLister      map[string]*cacheLister
 	}
 )
 
-func NewKialiCache() (KialiCache, error) {
+func NewKialiCache(namespaceSeedList ...string) (KialiCache, error) {
 	config, err := kubernetes.ConfigClient()
 	if err != nil {
 		return nil, err
@@ -103,6 +142,7 @@ func NewKialiCache() (KialiCache, error) {
 			cacheToken = saToken
 		}
 	}
+
 	istioConfig := rest.Config{
 		Host:            config.Host,
 		TLSClientConfig: config.TLSClientConfig,
@@ -125,12 +165,6 @@ func NewKialiCache() (KialiCache, error) {
 	}
 	log.Tracef("[Kiali Cache] cacheIstioTypes %v", cacheIstioTypes)
 
-	stopChan := make(map[string]chan struct{})
-
-	for _, ns := range cacheNamespaces {
-		stopChan[ns] = make(chan struct{})
-	}
-
 	cacheNamespacesRegexps := make([]regexp.Regexp, len(cacheNamespaces))
 	for i, ns := range cacheNamespaces {
 		cacheNamespacesRegexps[i] = *regexp.MustCompile(strings.TrimSpace(ns))
@@ -141,18 +175,34 @@ func NewKialiCache() (KialiCache, error) {
 		refreshDuration:        refreshDuration,
 		cacheNamespacesRegexps: cacheNamespacesRegexps,
 		cacheIstioTypes:        cacheIstioTypes,
-		stopChan:               stopChan,
-		nsCache:                make(map[string]typeCache),
+		// Only when all namespaces are accessible should the cache be cluster scoped.
+		// Otherwise, kiali may not have access to all namespaces since
+		// the operator only grants clusterroles when all namespaces are accessible.
+		clusterScoped:          kConfig.AllNamespacesAccessible(),
 		tokenNamespaces:        make(map[string]namespaceCache),
 		tokenNamespaceDuration: tokenNamespaceDuration,
 		proxyStatusNamespaces:  make(map[string]map[string]podProxyStatus),
 	}
-
 	kialiCacheImpl.k8sApi = istioClient.GetK8sApi()
 	kialiCacheImpl.istioApi = istioClient.Istio()
 
 	// Update SA Token
 	kialiCacheImpl.stopCacheChan = kialiCacheImpl.refreshCache(istioConfig)
+	kialiCacheImpl.registryRefreshHandler = NewRegistryHandler(kialiCacheImpl.RefreshRegistryStatus)
+
+	if kialiCacheImpl.clusterScoped {
+		log.Debug("[Kiali Cache] Using 'cluster' scoped Kiali Cache")
+		kialiCacheImpl.createClusterScopedCache()
+	} else {
+		log.Debug("[Kiali Cache] Using 'namespace' scoped Kiali Cache")
+		kialiCacheImpl.nsCacheLister = make(map[string]*cacheLister)
+		kialiCacheImpl.stopNSChans = make(map[string]chan struct{})
+
+		for _, ns := range namespaceSeedList {
+			kialiCacheImpl.CheckNamespace(ns)
+		}
+	}
+
 	log.Infof("Kiali Cache is active for namespaces %v", cacheNamespaces)
 	return &kialiCacheImpl, nil
 }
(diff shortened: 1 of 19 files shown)


nrfox:
Opening as a draft. Still needs some work but when you get a chance @lucasponce if you could take a look and see if this is headed in the right direction that would be much appreciated.

lucasponce:
Thanks, @nrfox, I'm catching up with this PR.

kubernetes/cache/cache.go
```diff
@@ -123,10 +131,35 @@ func NewKialiCache() (KialiCache, error) {
 	}
 	log.Tracef("[Kiali Cache] cacheIstioTypes %v", cacheIstioTypes)
 
-	stopChan := make(map[string]chan struct{})
+	// To determine if the cache is cluster-scoped or namespace-scoped,
+	// we check if Kiali has cluster-scoped permissions for each of the
+	// cache types. If Kiali does not have cluster-scoped permissions
+	// for any of the cache types, it is considered namespace-scoped.
+	// TODO: Can probably optimize this by paralellizing the cani checks.
```
lucasponce, line 138: I'm inclined to assume this comes from the configuration and avoid determining it dynamically.
Then, if some situation fails (i.e. no right permissions, or for some reason, some API is not present), I think it's better to have a clean error message about an expected situation than having this.

I guess it will get fewer moving parts that, in the long term, may help to maintain better.

But the logic is correct, I think it's used in other places as well.

kubernetes/cache/cache_handlers.go
```diff
@@ -1,149 +1,50 @@
...
 func (sh RegistryRefreshHandler) OnUpdate(oldObj, newObj interface{}) {
-	resourceVersion1, resourceVersion2 := sh.parseResourceVersion(oldObj, newObj)
-	if resourceVersion1 != resourceVersion2 {
+	var (
+		oldMeta v1.Object
+		newMeta v1.Object
+		err error
+	)
+	
+	if oldMeta, err = meta.Accessor(oldObj); err != nil {
```
lucasponce, line 34: Oh, nice one, but I remember that there was some "kinda" "bug/feature" in the API machinery, and some fields in the returned object are not populated.

But it makes sense; just a note to test this with detail to avoid side effects.

lucasponce:
> if you could take a look and see if this is headed in the right direction, that would be much appreciated.

Hey @nrfox, yes, I did the first pass, and the changes are pretty straightforward, I leave a couple of comments about the dynamic logic of detecting the cluster vs explicit from the config, but I think the PR is good.

I haven't tried to test it locally, just looking at the code; feel free to ping me when you'd need more reviews.

As "handlers" are used to "invalidate" the cache coming from the Istio Registry, I cc @hhovsepy as I think it could be good if he also jumps on these changes and gives a look as well when ready.

Thanks for the work.

hhovsepy:
I did review yesterday, changes looks good to me, will need this PR merged for my https://github.com/kiali/kiali/pull/5345, looking forward to run on OCP cluster for final testing.

kubernetes/cache/cache.go
nrfox: > Then, if some situation fails (i.e. no right permissions, or for some reason, some API is not present), I think it's better to have a clean error message about an expected situation than having this.

That's a good point. Someone may want the cluster scoped cache but not realize that they dropped to a namespace scoped cache because of lack of cluster permissions for a particular type. A large motivation for determining it dynamically was to avoid adding more cache related config options. My assumption was that users will just want the caching option that is faster and more efficient as long as they have the correct permissions for it. Kiali can choose this for them without needing an extra config option but to your point they may want the cluster scoped cache but be unaware that they are running with the namespace scoped cache if kiali transparently drops to the namespaced cache. I think that's more of a problem of informing the user what is happening within Kiali though.

Would a user ever _choose_ to have a namespace scoped cache when they could have a cluster scoped cache that is more efficient? I personally don't think so and Kiali could just the better option when it's available. Adding more config options adds complexity too it just pushes it more onto the user and into the deployment/operator/docs.

We could include the cache scope in the kiali status to make it more visible to users what was chosen? (namespace vs. cluster scoped)

kubernetes/cache/cache_handlers.go
nrfox: This shouldn't be affected by that issue since it's just using `resourceVersion` and not any of the type meta but +1 on testing to avoid side effects.

kubernetes/cache/cache.go
jshaughn: I doubt users would care about scoping (just repsonse time), but if we think it would be useful for debugging then yes.

lucasponce: I think that the operator should have the responsibility to populate that value.

Dynamic configuration as this one typically presents more edges that create issues in the long term; as an example, Kiali also had some "auto-discovery" logic for the "tracing" service, and that was a source of a lot of combinations that users presented.

If you are comfortable with the discovery logic, to me, it's ok; just sharing my thoughts here.

nrfox: > I think that the operator should have the responsibility to populate that value.
>
> Dynamic configuration as this one typically presents more edges that create issues in the long term; as an example, Kiali also had some "auto-discovery" logic for the "tracing" service, and that was a source of a lot of combinations that users presented.

That's a fair point. It could very well add some edge cases in the future. Having a config option that defaults to `clusterScope: enabled` would accomplish much the same thing in terms of ease of use. I'll remove the autodiscovery code in favor of a config option defaulting the cache to cluster scoped.

nrfox:
Moving this out of draft. This will now require operator/helm changes for the new config value.

nrfox:
This probably needs to be rebased with master now.

nrfox:
Tests are failing because I overlooked filtering at the informer level. In other words, with a cluster scoped cache all namespaces are being returned. I assumed that the filtering was happening at a lower level but it's actually dependent on the namespace scoped cache. IMO it'd be a good opportunity to refactor to use listers here instead of indexers inside of the cache implementation as this does the namespace filtering for you and would reduce boilerplate instead of adding it. The cache interface can stay the same.

nrfox:
Going to include removal of the kube cache config option and make the cache mandatory as part of this change. It was originally added as a fallback option before the caching was stable but it is no longer needed. In fact the Istio Regsitry relies on the cache being enabled in order to refresh on add/update/delete events.

nrfox:
> Going to include removal of the kube cache config option and make the cache mandatory as part of this change. It was originally added as a fallback option before the caching was stable but it is no longer needed. In fact the Istio Regsitry relies on the cache being enabled in order to refresh on add/update/delete events.

Perhaps we can make removal of the caching config option a separate change? The diff on this PR is becoming quite large.

kubernetes/cache/cache.go
```diff
@@ -59,15 +58,44 @@ type (
 		proxyStatus *kubernetes.ProxyStatus
 	}
 
+	// cacheLister combines a bunch of lister types into one.
```
jshaughn, line 61:
```suggestion
	// cacheLister combines a bunch of listener types into one.
```

nrfox:
Cypress tests have been failing consistently but they're catching a real discrepancy and I hopefully pushed a fix to address it. One thing I've noticed is that the frontend assumes that some API response fields are required but they are actually optional on the backend. When these fields aren't present, the frontend usually fails with a "Sorry, there was a problem. Please refresh the page" message.

jshaughn:
> that some API response fields are required but they are actually optional on the backend

I have seen this problem and made point fixes in some other PRs.  I'm not surprised the issue still exists in places.  Probably the client should be more robust in that situation as it's not always possible for the server to provide a value.

lucasponce:
Now that there are:
- More robust clients for Kubernetes and Istio backend.
- Unified single repo for backend/frontend code.

These issues https://github.com/kiali/kiali/issues/5179 and https://github.com/kiali/kiali/issues/4783 could be revisited.

lucasponce:
@nrfox a side comment for working on this area: if you enable the development mode out of the cluster, then you can easily intercept all the client's traffic (using wireshark, or tcpdump):

image

This is a quick way to check if implementing the "list+watch" on the kubernetes/istio packages doesn't have any issue.

Now they are pretty stable, but if the version of those dependencies is not 100% synced, the watchers typically may show some "oddities" (extra warnings, or something that doesn't work as you expect.

Using this tool saves time to check if something is missing.

HTH

lucasponce:
I think it could be time to put all k8s using the list+watch:
image

The optional "Jobs", and "Cronjobs" seem that are not yet there, and "wireshark" shows non-expected "namespace" queries.

Those are not visible in smaller environments, but typically have an impact on these large scenarios that users are reporting.

kubernetes/cache/istio.go
```diff
@@ -11,807 +10,536 @@ import (
...
-		}
-	}
-	return nil, nil
+
+	dr, err := c.getCacheLister(namespace).destinationRuleLister.DestinationRules(namespace).Get(name)
+	if err != nil {
+		return nil, err
+	}
+
+	dr.Kind = kubernetes.DestinationRuleType
```
lucasponce, line 115: Ah, nice.
Yes, k8s apimachinery returns "empty" and "Kind" and those are used when parsing objects into YAML into the frontend.

lucasponce (commented):
I'd recommend having a detailed review in the development mode first, watching the real queries that are performed against the k8s API, that may show if some bump on the k8s/istio clients-go is necessary.

The code looks ok, and smoke tests worked for me, but I couldn't perform deep testing; I think this one would require one.

nrfox:
> I think it could be time to put all k8s using the list+watch: image
> 
> The optional "Jobs", and "Cronjobs" seem that are not yet there, and "wireshark" shows non-expected "namespace" queries.
> 
> Those are not visible in smaller environments but typically have an impact on these large scenarios that users are reporting.
> 
> Note that OpenShift "DeploymentConfig" controllers do not have a "list+watch"; perhaps those ones would need support, I am aware some users are still using them.
> 
> All these controllers are marked as optional in the config, as they were usually "corner cases," but if users want to use them, I think they should be supported, otherwise some oddities related to performance may come from this angle.

I'll take a look there and see what requests are happening. I agree it makes sense to add those to the cache as well when those workloads aren't excluded. I'm planning on a followup change to remove the cache_enabled config option and have this be the default so perhaps it'd be better to add it then along with the DeploymentConfig caching.

nrfox:
> I've added a couple of comments. I'd recommend having a detailed review in the development mode first, watching the real queries that are performed against the k8s API, that may show if some bump on the k8s/istio clients-go is necessary.
> 
> The code looks ok, and smoke tests worked for me, but I couldn't perform deep testing; I think this one would require one.

Thanks for taking a look! Agreed before merging we'll want to make sure everything is stable.

kubernetes/cache/cache.go
```diff
@@ -163,90 +207,126 @@ func (c *kialiCacheImpl) isCached(namespace string) bool {
...
 
+func (c *kialiCacheImpl) createClusterScopedCache() bool {
+	kubeInformerFactory := c.createKubernetesInformers("")
+	istioInformerFactory := c.createIstioInformers("")
+
+	c.stopClusterScopedChan = make(chan struct{})
+	kubeInformerFactory.Start(c.stopClusterScopedChan)
+	istioInformerFactory.Start(c.stopClusterScopedChan)
+
+	log.Info("Waiting for Kiali cache for to sync")
```
jshaughn, line 249:
```suggestion
	log.Info("[Kiali Cache] Waiting for cluster-scoped cache to sync")
```

kubernetes/cache/cache.go
```diff
@@ -163,90 +207,126 @@ func (c *kialiCacheImpl) isCached(namespace string) bool {
...
+	for _, synced := range kubeInformerFactory.WaitForCacheSync(c.stopClusterScopedChan) {
+		if !synced {
+			log.Error("[Kiali Cache] failed to sync")
```
jshaughn, line 252:
```suggestion
			log.Error("[Kiali Cache] failed to sync kube cluster-scoped cache")
```

kubernetes/cache/cache.go
```diff
@@ -163,90 +207,126 @@ func (c *kialiCacheImpl) isCached(namespace string) bool {
...
+			return false
+		}
+	}
+	for _, synced := range istioInformerFactory.WaitForCacheSync(c.stopClusterScopedChan) {
+		if !synced {
+			log.Error("[Kiali Cache] failed to sync")
```
jshaughn, line 258:
```suggestion
			log.Error("[Kiali Cache] failed to sync Istio cluster-scoped cache")
```

kubernetes/cache/cache.go
```diff
@@ -163,90 +207,126 @@ func (c *kialiCacheImpl) isCached(namespace string) bool {
...
-		log.Errorf("Kiali cache for [namespace: %s] sync failure", namespace)
-		return false
+	for _, synced := range istioInformerFactory.WaitForCacheSync(c.stopNSChans[namespace]) {
+		if !synced {
+			log.Error("[Kiali Cache] failed to sync")
+			return false
+		}
 	}
+
 	log.Infof("Kiali cache for [namespace: %s] started", namespace)
```
jshaughn, line 236:
```suggestion
	log.Infof("[Kiali Cache] started [namespace: %s] cache", namespace)
```

kubernetes/cache/cache.go
```diff
@@ -163,90 +207,126 @@ func (c *kialiCacheImpl) isCached(namespace string) bool {
...
-		for _, informer := range c.nsCache[namespace] {
-			go informer.Run(stopCh)
-		}
-		<-stopCh
-		log.Infof("Kiali cache for [namespace: %s] stopped", namespace)
-	}(c.stopChan[namespace])
+	kubeInformerFactory.Start(c.stopNSChans[namespace])
+	istioInformerFactory.Start(c.stopNSChans[namespace])
 
 	log.Infof("Waiting for Kiali cache for [namespace: %s] to sync", namespace)
```
jshaughn, line 221:
```suggestion
	log.Infof("[Kiali Cache] waiting for [namespace: %s] cache to sync", namespace)
```

kubernetes/cache/cache.go
```diff
@@ -163,90 +207,126 @@ func (c *kialiCacheImpl) isCached(namespace string) bool {
...
-		return c.createCache(namespace)
+		return c.createNSCache(namespace)
 	}
-	return c.isKubernetesSynced(namespace) && c.isIstioSynced(namespace)
+
+	return true
 }
 
 // RefreshNamespace will delete the specific namespace's cache and create a new one.
+// If a cluster-scoped namespace is used, it will recreate the cluster-scoped cache.
```
jshaughn, line 292: This seems strange, would it not lead to refreshing a cluster-scoped cache too often, in some loop for namespaces?  I think I'd rather see a dedicated function for each, and if the namespace version is called when cluster-scoped cache is enabled, it should be an error (and vice-versa).

kubernetes/cache/cache.go
```diff
@@ -59,15 +58,44 @@ type (
...
+	cacheLister struct {
+		// Kube listers
+		configMapLister   core_v1_listers.ConfigMapLister
+		daemonSetLister   apps_v1_listers.DaemonSetLister
+		deploymentLister  apps_v1_listers.DeploymentLister
+		endpointLister    core_v1_listers.EndpointsLister
+		podLister         core_v1_listers.PodLister
+		replicaSetLister  apps_v1_listers.ReplicaSetLister
+		serviceLister     core_v1_listers.ServiceLister
+		statefulSetLister apps_v1_listers.StatefulSetLister
```
jshaughn, line 73: This doesn't seem to cover all of the workload types we support, is that intended?

jshaughn (commented):
A first pass, couple of questions/comments/suggestions.  Definitely looking good.

kubernetes/cache/cache.go
nrfox: It's all the types that are currently cached today. The "optional" types such as `DeploymentConfig` and `CronJob` are not cached. I can add those as part of this change but I was planning on adding them in a followup where we remove the `cache_enabled` option and make caching permanently enabled.

kubernetes/cache/cache.go
nrfox: Yes the cluster scoped cache could get refreshed in a for loop but the refresh only happens when we do a write operation on some resource: https://github.com/kiali/kiali/blob/master/business/services.go#L595. Unless Kiali is updating resources across many namespaces at once, many different refreshes at once are not very likely to happen. There may be other places but the main spot that Kiali performs write operations is through the wizards.

My guess is that the primary purpose of the refresh is that there's no way, that I know of, to fast forward the cache to a particular ResourceVersion. You could update the object then do another read from the cache and not see the updated object. Recreating the cache entirely force a fast forward in a way to the latest ResourceVersion.

I'll look into adding separate functions for the refresh but it'd be nice for clients to not have to know that the cache is/isn't cluster scoped.

nrfox:
Going to squash and rebase this with master to address conflicts

kubernetes/cache/cache.go
jshaughn: No problem adding as a follow-up, was just trying to understand.

kubernetes/cache/cache.go
jshaughn: If you feel like the direct callers of this function should not care about the cache-type, and that there is no real concern around the "unnecessary looping" comment, I'm fine with things as-is.  My comment was mainly around catching issues in the calling code where the cache-type was being mishandled.  I might just suggest, for clarity, a change to the function name, but totally up to you:

```
// RefreshCache will recreate the necessary cache. If the cache is cluster-scoped the "namespace" argument 
// is ignored, otherwise only the namespace-specific cache is updated.
func (c *kialiCacheImpl) RefreshCache(namespace string)
```

nrfox: Renaming sounds good. Direct callers would still need to call refresh in all the same places (writes to kube objects) regardless of whether the cache is namespace or cluster scoped so it would only change where the check for cluster vs. namespace scope is happening. Cluster scoped refreshes could be slower especially if there's a lot of kube objects to list. It's a problem for namespace scoped as well but might be worse with cluster scope since there's more objects. As far as I can tell writes are fairly infrequent operations though as they happen only when a user uses a wizard or adds injection etc. This is something that needs to be tested out with this change.

nrfox:
Looking at wireshark requests, in addition to requests to the api for the non-cached types, there's some calls to the kube api for pods around getting the debug status. Looking at the code these don't use the cache: https://github.com/kiali/kiali/blob/master/kubernetes/istio.go#L56. There could be good reasons for doing that but all that to say the results with this change are as expected.

nrfox:
Need to rebase this with master now.

nrfox:
Rebased and tested with `accessible_namespaces == **` and `accessible_namespaces != **`

jshaughn:
@nrfox Is this ready again for a review?

nrfox:
@jshaughn yes!

jshaughn (dismissed):
This looks ready to me.

jshaughn (approved):
Re-approval after minor changes.

PR #5322, opened 2022-07-20
Redesign of Control Plane card in the Overview page

### LINKED ISSUE #5167: Reorganize the Overview page to better show data plane vs control plane status
Today Kiali's overview page is focused on the data-plane status:
image

It shows pretty much the same information provided by the "Mesh Dashboard" which is also focused on the data plane:
image

Global ratio, % of errors, and health per app/services/workloads. It also allows to operate per namespace level:
image

The "control plane" is handled like another regular namespace on the page, but it's also special as it has some specific "control plane" options that are less visible:
image

I think this epic could be used to review how to make a better distinction on this page for the control plane namespace, and probably try to better add in the "istio-system" card, a status that belongs to the control plane itself:
image

So, a proposal could be that the "istio-system" (or any other namespace that works like a control plane) may have a slightly different design and show a combined status with the relevant info, but it's still kept integrated into the overview view (compact, expand, list).

leandroberetta:
Resolves #5167.

Control plane card with more meaningful metrics and information.


leandroberetta:
Some screenshots showing the latest changes:

1. The control plane card is the first card no matter what sort the user is applying and also the indicator that is the control plane namespace is added

cp_overview_1

2. I decided (for now, I'm testing other metrics as well) to display the memory and cpu for the istiod process. Also, a new section on the right will be used to display some configuration, for example, the registry only configuration (I will add tooltip to help the user understanding the information better)

cp_overview_2

leandroberetta:
Some progress related to this issue:

I added tooltips to better explain the metrics:

cp_o_tooltip

I also introduced this changes to the List view:

cp_o_list

I'm investigating more control plane common configurations to create more badges apart from the registry_only one.

ceastman-r7:
How about Istio version?

leandroberetta:
> How about Istio version?

This could be added to the about section, that already contains other versions like Prometheus, Grafana, etc.

leandroberetta:
A new badge is added to indicate (if there is information during the timeframe) the average proxy push time:

overview_cp_new

This will indicate if there is some kind of saturation, when the time increase, it will be noticeable.

lucasponce:
Thanks, @leandroberetta, I'm catching up with this work.
One of my comments would be that the control plane chart may have a "different" style to reinforce the control-plane/data-plane differentiation.

Combining the metrics or exploring any other option may help here, but I don't have any specific proposal :).

I think it could be a good area to experiment with different visualization that helps the correlation (i.e., inspired by what we have with metrics+traces+heatmaps).

hhovsepy (dismissed):
looks good to me
Screenshot from 2022-08-03 08-15-28

lucasponce:
Hi @leandroberetta, some final feedback from my side about this work, in case that helps:
- +1 to add mem and cpu
- Instead of using two separate charts, I'd combine them into a single one, using the y-axis left and right; this will have some benefits: less space, and also you can "correlate" the charts (i.e. if a peak of mem is directly related to a peak of CPU).
- I'd use a completely different color scheme style for the control plane to make it more relevant to the data-plane ones.
- I'd introduce some "dash" line indicating some threshold; as discussed, the absolute values are not interesting, but users may want to know when a peak is not "normal" (you can start with some "canned" threshold i.e. >50% of CPU or similar for memory, the goal here is to identify something unusual to "click" into the details).

leandroberetta:
The following image represents some improvements based on suggestions:

image

- I'm trying with removing the y axis from every sparkline chart on this page, but including the unit and what represent the curve on the tooltip. I think it helps to save space
- I'm changing the color schme to green, even the border top just to diffferentiate more the card and also to be more consistent with the graph color

Next step is to add the thresholds and maybe scale cpu a little bit up.

leandroberetta:
The following is an image where the proxy update time appears (it will appear when it's available, meaning that if on the time range there are proxy updates, the highets value will appear). 

image

leandroberetta:
> The following is an image where the proxy update time appears (it will appear when it's available, meaning that if on the time range there are proxy updates, the highets value will appear).
> 
> image

@andrew-ronaldson do you have any comments about this and the past UX meeting? something related to the color schemes and what not. Thanks!.

ceastman-r7:
@leandroberetta Love the proxy update delay information.

leandroberetta:
> @leandroberetta Love the proxy update delay information.

Thanks for the feedback!.

leandroberetta:
The following is an update based on recent UX discussions:

image

israel-hdez:
> The following is an update based on recent UX discussions:
> 
> image

Look at it... It's beautiful!

leandroberetta:
> > The following is an update based on recent UX discussions:
> > image
> 
> Look at it... It's beautiful!

Thanks @israel-hdez :)

leandroberetta:
@jshaughn this effort is ready for review. Cypress test are running, hopefully they will pass now.

An screenshot of the last updates (minimal updates really):

image

business/metrics.go
```diff
@@ -226,3 +226,31 @@ func createStatsMetricsLabelsBuilder(q *models.MetricsStatsQuery) *MetricsLabels
 	}
 	return lb
 }
+
+func (in *MetricsService) GetControlPlaneMetrics(q models.IstioMetricsQuery, scaler func(n string) float64) (models.MetricsMap, error) {
+	metrics := make(models.MetricsMap)
+
+	h := in.prom.FetchHistogramRange("pilot_proxy_convergence_time", "", "", &q.RangeQuery)
```
jshaughn, line 233: The `pilot_proxy_convergence_time`  metric is not on the [required metric list](https://kiali.io/docs/faq/general/#requiredmetrics) and needs to be added.  As such it is removed in our [thinning config](https://kiali.io/docs/configuration/p8s-jaeger-grafana/prometheus/#metric-thinning).  We will need to fix the recommended thinning config and make the appropriate kiali.io changes.

business/metrics.go
```diff
@@ -226,3 +226,31 @@ func createStatsMetricsLabelsBuilder(q *models.MetricsStatsQuery) *MetricsLabels
...
+	var err error
+	converted, err := models.ConvertHistogram("pilot_proxy_convergence_time", h, models.ConversionParams{Scale: 1})
+	if err != nil {
+		return nil, err
+	}
+	metrics["pilot_proxy_convergence_time"] = append(metrics["pilot_proxy_convergence_time"], converted...)
+
+	metric := in.prom.FetchRateRange("process_cpu_seconds_total", []string{`{app="istiod"}`}, "", &q.RangeQuery)
```
jshaughn, line 241: The `process_cpu_seconds_total` metric is not on the [required metric list](https://kiali.io/docs/faq/general/#requiredmetrics) and needs to be added.   It is not in our [thinning config](https://kiali.io/docs/configuration/p8s-jaeger-grafana/prometheus/#metric-thinning), and so no changes are required there.

business/metrics.go
```diff
@@ -226,3 +226,31 @@ func createStatsMetricsLabelsBuilder(q *models.MetricsStatsQuery) *MetricsLabels
...
+	converted, err = models.ConvertMetric("process_cpu_seconds_total", metric, models.ConversionParams{Scale: 1})
+	if err != nil {
+		return nil, err
+	}
+	metrics["process_cpu_seconds_total"] = append(metrics["process_cpu_seconds_total"], converted...)
+
+	metric = in.prom.FetchRange("container_memory_working_set_bytes", `{pod=~"istiod-.*|istio-pilot-.*"}`, "", "", &q.RangeQuery)
```
jshaughn, line 248: The `container_memory_working_set_bytes` metric is not on the [required metric list](https://kiali.io/docs/faq/general/#requiredmetrics) and needs to be added.  It is not in our [thinning config](https://kiali.io/docs/configuration/p8s-jaeger-grafana/prometheus/#metric-thinning), and so no changes are required there.

frontend/src/components/Charts/SparklineChart.tsx
```diff
@@ -20,6 +23,9 @@ type Props = ChartProps & {
   showLegend?: boolean;
   showYAxis?: boolean;
   tooltipFormat?: (dp: VCDataPoint) => string;
+  labelName: string;
+  thresholds?: VCLines<RichDataPoint>;
+  showXAxisValues?: boolean;
```
jshaughn, line 28: the new props break the alphabetical ordering.

frontend/src/pages/Overview/OverviewCardControlPlaneNamespace.tsx
```diff
@@ -0,0 +1,176 @@
...
+import { Card, CardBody, Flex, FlexItem, Grid, GridItem, Tooltip, TooltipPosition } from '@patternfly/react-core';
+import { KialiIcon } from 'config/KialiIcon';
+import { style } from 'typestyle';
+import { IstiodResourceThresholds } from 'types/IstioStatus';
+
+export const infoStyle = style({
+    margin: '0px 0px -1px 4px'
+});
+
+
```
jshaughn, line 21: extra line, maybe file needs formatting?

routing/routes.go
```diff
@@ -1420,6 +1420,24 @@ func NewRoutes() (r *Routes) {
 			handlers.OutboundTrafficPolicyMode,
 			true,
 		},
+		// GET /api/mesh/resources/thresholds
```
jshaughn, line 1423: Should this be ` swagger:route GET /api/mesh/resources/thresholds`  ?

jshaughn (commented):
@leandroberetta Looks good, a few comments.

routing/routes.go
leandroberetta: Yes, I forgot to add it, although we are not using Swagger anymore,  I added in the same fashion as others just in case we will use it again.

business/metrics.go
leandroberetta: Thanks, this is addressed on: https://github.com/kiali/kiali.io/pull/582

PR #5323, opened 2022-07-20
Launch wizards from graph's side panel

### LINKED ISSUE #4505: Launch Kiali wizard scenarios from graph nodes
Based on part of [this discussion](https://github.com/kiali/kiali/discussions/4084#discussioncomment-937881), users can launch the Kiali wizards directly from the graph nodes.

I propose that right clicking a graph node brings up the appropriate wizards in the context menu based on the node type e.g. right clicking on a service node would bring up the context menu with the "create request routing" option. Clicking on "create request routing" would bring up the request routing wizard. In addition, clicking the kebab on the node summary panel would also have a wizard options appropriate to the node type.

Current context menu:
Screenshot from 2021-11-17 13-24-49

Node summary panel kebab:
Screenshot from 2021-11-17 13-25-08

Some additional considerations:
- Passthrough/Blackhole cluster should probably be exceptions
- What happens if the graph data refreshes and the node the wizard was being applied to disappears? Perhaps the graph refresh should be paused while the wizard is open and resumed to it's previous value after being closed?

israel-hdez:
Wizards are available under the side-panel's kebab menu and also on the contextual menu of the Cy component:

image

Fixes #4505


israel-hdez:
@nrfox This will cover your RFE, so I'm adding you as a reviewer.
Also, pinging @lucasponce in case you also want to review.

lucasponce:
Thanks, @israel-hdez, the screenshots look good.

Better than me, I cc @hhovsepy, which has most of the knowledge finding potential problems in the wizards, and @aljesusg, which added the latest preview features.

Wizards have a lot of non-trivial "state" management, so I guess they would be able also to test and review this work.

Thanks,

hhovsepy:
Looks good to me:
Screenshot from 2022-08-03 07-13-11

Only I see one regression on Wizard itself, can this be because of this changes? As on master I could not reproduce.
When choosing Advanced options -> Traffic Policy -> Add LoadBalancer -> Choose Simple:
Screenshot from 2022-08-03 07-01-17
It previews correctly.
Screenshot from 2022-08-03 07-01-17
Create it.
Try to edit traffic via wizard or view the DR/VS content:
Screenshot from 2022-08-03 07-03-05
It does not keep ROUND_ROBIN
Screenshot from 2022-08-03 07-18-42
Screenshot from 2022-08-03 07-18-23

israel-hdez:
@hhovsepy I could replicate.
I simply did a rebase (mainly to fix the conflict) and the issue went away.

I'm unsure what was the issue, but most likely it was present in `master` and somebody already fixed it.

frontend/src/hooks/services.ts
```diff
@@ -0,0 +1,91 @@
...
+  const [peerAuthentications, setPeerAuthentications] = React.useState<PeerAuthentication[] | null>(null);
+
+  React.useEffect(() => {
+    if (namespace.length === 0 || serviceName.length === 0) {
+      return;
+    }
+
+    setIsLoading(true); // Mark as loading
+    let getDetailPromise = API.getServiceDetail(namespace, serviceName, false, duration);
+    let getGwPromise = API.getIstioConfig('', ['gateways'], false, '', '');
```
hhovsepy, line 27: There was a change in loading of configs, it fails here:
Screenshot from 2022-08-09 12-30-05
Should call: "allIstioConfigs"

hhovsepy: @israel-hdez it seems another rebase is needed and fix of loading gateways, as Service dropdown is not loading.

israel-hdez: I just pushed the fix.
Hopefully, I did it right this time.

hhovsepy (approved):
Verified

israel-hdez:
I'm merging. Thanks @hhovsepy for your review.

nrfox:
I'm late to comment here but it'd be good to add some cypress tests covering some basic aspects of the new feature. Excited to see this feature become a reality.

israel-hdez:
Somehow I understood that cypress tests no longer were important on the dev side.

In any case, my thought is that unit tests would be a much better fit for this case. It is not really a brand-new feature, this is just porting an already existent feature into the graph. So, I think testing should be limited to verifying the right behavior of the menus and stop there. IMO, that would be better covered in units. That said, we really haven't agreed about what to do with units. We are, simply, working in best-effort mode to update what gets broken and not really caring much past that.

I would agree that testing the wizards is better in cypress, but that would be out of scope for this specific PR.

jshaughn:
> Somehow I understood that cypress tests no longer were important on the dev side.

Totally the opposite :-)  We are responsible for increasing test coverage incrementally as part of feature or fix work, as appropriate.  The "as appropriate" clause is totally up to the devs and reviewers.  If unit tests are more appropriate that is fine too (although I'd discourage snapshot tests).  And, of course, if it is agreed between devs and reviewers that current coverage is sufficient, that's fine too. 

Bottom line is simple, we want to be able to say that test coverage is robust.

israel-hdez:
>  If unit tests are more appropriate that is fine too (although I'd discourage snapshot tests). And, of course, if it is agreed between devs and reviewers that current coverage is sufficient, that's fine too.

Well, as I said, I personally think that units are better for this case. Although, I that's debatable and if you guys think cypress is better, I can try it. I'm not sure how to proceed. As I said, units are mostly abandoned and are just being patched as needed. We should probably spend a little time in tomorrow's planning call to make a decision about if it is still worth to maintain the units (in it's broad sense; i.e. not touching the specifics about if they are snapshots or regular specs).

About the snapshots, if we choose to keep the units, I think I can prepare a small talk about how *I think* they are supposed to be used. Thereafter, we can choose whether to keep and actively use them, or if it is better to remove them all. Personally, I'm not in favor not against snapshots.

nrfox:
> Well, as I said, I personally think that units are better for this case. Although, I that's debatable and if you guys think cypress is better, I can try it.

A cypress test that:
1. clicks on a node
2. validates context menu pops up with "create xxx" option
3. clicks on the link and verifies that the wizard pops up
would be valuable here imo.

The [istio config page](https://github.com/kiali/kiali/blob/master/frontend/cypress/integration/featureFiles/istio_config.feature#L33-L49) does something similar where it just tests that the wizard opens but doesn't test the wizard itself.

> I'm not sure how to proceed. As I said, units are mostly abandoned and are just being patched as needed. We should probably spend a little time in tomorrow's planning call to make a decision about if it is still worth to maintain the units (in it's broad sense; i.e. not touching the specifics about if they are snapshots or regular specs).

Personally I think the unit tests are very good for testing "pure logic" like the health calculations that happen on the frontend. I even think there's some value in the "it renders" tests because everything can compile with typescript but the component may still not render. Probably redux logic is another good use for unit tests. Functionally testing the UI however is better suited for a cypress test. When to add one is debatable but imo if the feature is adding some new functionality to the UI or fixing a regression then a cypress test should be added along with the code changes. That's just my opinion and we can discuss and decide more formally when/what tests to add.

israel-hdez:
> Personally I think the unit tests are very good for testing "pure logic" like the health calculations that happen on the front-end.

Yeah. I think you are right. Indeed, the logic is what is in my head. What I'm wanting to add are several units to check that the right items are enabled, disabled and rendered or not rendered, given the several possibilities of data that can come from the back-end (including errors), and also that the right events are fired. As usual in units, I was thinking in doing it in an isolated way, if possible (i.e. just the menus). 

So, that's why I was mainly recommending units here. Simply, because this is what is in my head.

> Functionally testing the UI however is better suited for a cypress test. 

Yeah, I think you are right again. The test you suggested is fine, but I see it more as a complement of the units. It is just to ensure that everything is wired correctly. It is valuable, but IMHO only the combo (units+cy) would offer a robust suite.

That said, I would leave out cucumber from the equation because your suggested test is not a use-case. I still think that cucumber is more geared towards testing use cases rather than functionality, so I'm not sure why cucumber is a "must use" when we can have a mixed suite of cucumber and non-cucumber tests depending on what's better.

All in all, I don't disagree with anything. And as I said in the past, I think we can use the tooling to our liking. I have the idea that the way to use the testing tooling has already been settled down in a *de-facto* way, but nothing is formalized in written form (in the [READMEs of the repos we only have vague mentions]((https://github.com/kiali/kiali/blob/master/frontend/README.adoc#integration-testing))) nor well communicated. I do see that trend to prefer cypress, but that same trend is causing us to immediately rule out units. Moreover, I have seen (very) few PRs where units are simply deleted without a replacement if they become hard to fix, which is somewhat disappointing. So, why not make it formal that front-end units are deprecated? We could remove `yarn test` from our workflow and save some time. Note that all this is specially true for the UI side, while in the back-end units are still well maintained.

Anyway, all good. I'll add the cypress test you suggested (using cucumber).

nrfox:
> Yeah, I think you are right again. The test you suggested is fine, but I see it more as a complement of the units. It is just to ensure that everything is wired correctly. It is valuable, but IMHO only the combo (units+cy) would offer a robust suite.

Agreed that having both is ideal but if I was to choose one over the other I'd choose having a cypress test.

> That said, I would leave out cucumber from the equation because your suggested test is not a use-case. I still think that cucumber is more geared towards testing use cases rather than functionality, so I'm not sure why cucumber is a "must use" when we can have a mixed suite of cucumber and non-cucumber tests depending on what's better.

I don't disagree with you here but that is a tool/format that the testers have chosen for the cypress tests and given that it's a shared testing suite, they have a large input into what the tests should look like. If were going to only use cucumber sometimes then we need to be in agreement on when that is.

> I do see that trend to prefer cypress, but that same trend is causing us to immediately rule out units. 

I don't like that trend. Can we not have both as we discussed above? I think we've been less stringent on requiring automated testing (cypress or units) with code changes in the past but imo we should be more proactive about that going forward. At least because the automated testing (selenium testing) lived in another repo it was less obviously connected to particular code changes.

israel-hdez:
> if I was to choose one over the other I'd choose having a cypress test.

We should not have that luxury :-P

> Can we not have both as we discussed above?

I can do it. Yet, I don't think the team is in sync about this :-/
For now, I'll do only the cypress one, which is what you originally asked for and is what the team is focusing on.

If later is made clear that also units are a worthwhile, I'll add the units.

jshaughn:
It looks like we need to meet about this.  But until we meet I'll offer up a few more thoughts...  The bottom line is that we want tests that make us confident that what we release has good quality.   Test code must itself be good code, meaning it is understandable and can be maintained, runs efficiently enough to serve our purposes, has obvious benefit, and the effort to write or maintain it is worth the coverage it provides. Unit testing and automation/cypress testing are both valuable.  I think the acid test for cypress is whether the alternative is manual interactive testing of the UI.  We don't want people having to sit there and manually interact. Unit testing is certainly valuable, and we have a bunch of useful unit tests that are not going anywhere.   They are particularly well-suited to testing utilities, logic, parsing, etc.  Where it is a toss-up, I personally would go with Cypress, as this testing is fairly easy to understand and can be maintained by the most people.  There has been some disagreement about the snapshots, and I think that has to do with them being seen by some (like me) as not meeting of all the criteria above.  

As for using the test approaches "the right way", like the semantics of using cucumber for use case testing, it's something for us to hash out.  But a test that just verifies the mechanics of a feature is usually better than no test at all, and so I'm personally not going to stop someone from writing a test that perhaps doesn't leverage the tool to its fullest extent.

Automated testing is a hard topic.  You have to find a balance where you have a good amount of useful testing while spending a reasonable amount of time working on test code.

PR #5327, opened 2022-07-21
[hack] Delete install-sm11.sh

jmazzitelli:
ossm 1.1 is no longer supported - we don't need this hack script anymore


leandroberetta (approved):
LGTM

PR #5333, opened 2022-07-25
Overlay trace onclick event doesn't work

### LINKED ISSUE #5286: Overlay trace onclick event doesn't work in the metrics charts
It seems that Kiali may have a regression on the "click" event used in the overlay traces
image

In theory, when clicking on a trace, it should jump the user to the traces tab with that selected trace.

This problem is intermittent but after the merge of the latest changes, when I click in the trace, I see the URL changed
image

But not the jump to the traces tab.

There were some issues in the past with this event, and mostly not directly related to Kiali code but the VictoryCharts libraries used, perhaps this bug issue can be used to see if this event can be fixed (or refactored) in a solid way.

josunect:
Fixes #5286


lucasponce:
Thanks, @josunect, I'm catching up with this review.

lucasponce (approved):

PR #5337, opened 2022-07-25
Pipeline to release patches with no integration testing

leandroberetta:
To release a patch of v1.48 with the current release pipeline I needed to skip the jobs related to integration testing (because there are missing scripts and cypress tests on that version). 

The problem is that the last job called "release", needs that steps, because, for the versions that we have integration tests in place, we want the release process to happen after all checks.

If I skip the integration jobs needed by the release job, the release job is skipped too, it makes sense because the dependencies are not met.

So, to avoid changing our current functional and tested pipeline much, I decided to create a similar pipeline that does not have the integration testing steps.

### DIFF
--- a/.github/workflows/integration-tests-backend.yml
+++ b/.github/workflows/integration-tests-backend.yml
@@ -15,7 +15,6 @@ env:
 
 jobs:
   integration_tests_backend:
-    if: ${{ !contains(inputs.build_branch, 'v1.48') }}
     name: Backend API integration tests
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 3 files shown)


jmazzitelli (approved):
ok. kinda stinks we need to just copy everything from the other workflow but, oh well.

leandroberetta:
> ok. kinda stinks we need to just copy everything from the other workflow but, oh well.

I know, I will keep investigating a little more to avoid this, but I thought I investigated this on the past with no luck, anyway, if I couldn't find anything for tomorror I will merge this.

Thanks.

leandroberetta:
@jmazzitelli I'm closing this pull request in favor of https://github.com/kiali/kiali/pull/5338 that seems to be a better solution.

PR #5338, opened 2022-07-25
Adding release scripts to this version without integration testing

leandroberetta:
When starting pipelines, the source where the pipeline is obtained is configurable, usually is master, using the lastest pipeline, but someone can point to another branch/tag.

To fix the situation of duplicating a pipeline just to avoid integration testing, I'm introducing GitHub Actions pipelines to v1.48, a version that didn't have this kind of automations in the past.

The pipeline that I'm adding is the same as the current but without the automations.

This is just for v1.48 as is the only version with no integration tests which we might need patches, for the subsequents, we will continue using the lastest version of the release workflow.

I will document this on the document in the master branch.

### DIFF
--- a/.github/workflows/build-backend.yml
+++ b/.github/workflows/build-backend.yml
@@ -1,0 +1,56 @@
+name: Build Backend
+
+on:
+  workflow_call:
+    inputs:
+      build_branch:
+        required: true
+        type: string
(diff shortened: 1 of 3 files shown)


jmazzitelli (approved):
lgtm

PR #5339, opened 2022-07-26
Trace details heatmap vertical labels are truncated 

### LINKED ISSUE #5291: trace details heatmap vertical labels are truncated
Applications Traces tab - look at the subtab Trace Details - specifically look at the right-most heat map... and look at the labels of the vertical axis. "ull duration" and "ans avera" - those labels look truncated. We should fix so the labels are fully visible.

image

josunect:
Remove the whitespace for label cells for the heatmap component in Trace Details .

image

Fixes #5291


jmazzitelli (approved):
just tested... much better :)
image

Issue #5026, opened 2022-04-29
Support for Istio Gateway API


nrfox:
Kiali should support the new [Gateway API](https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/) and other new networking resources such as `HTTPRoute`.

Support would entail:
- [x] https://github.com/kiali/kiali/issues/5485
- [x] https://github.com/kiali/kiali/issues/5490
- [x] https://github.com/kiali/kiali/issues/5501
- [x] https://github.com/kiali/kiali/issues/5502
- [x] Gateway API in Console Plugin

lucasponce:
This issue is big.
It should start with supporting these new objects but it can't replace the current functionality invalidations and wizards.
Also, co-existing both implementations will have side effects as these new API is under evolution and it's not a 1-to-1 related to the Istio Gateway/VirtualService.

Another aspect is this API is optional and it requires the admin to install it separately, this would affect and it would require a "is it enabled" logic in Kiali that would add extra complexity.

So, I think we need to review this, but I don't see how Kiali can extend wizards on validations on these new objects just yet.

My recommendation is to limit the scope of this work to listing/editing these objects as a first step.

hhovsepy:
It seems that Gateway API is not supported yet in https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go

jshaughn:
Taking off the sprint backlog until there is an API we can query for this kind of config.

Issue #5213, opened 2022-06-15
More than one Gateway, but cannot find duplicate.


simonpie:
Kiali reports for a few Gateway Objects that we have a duplicate host and port conflict.  
![Capture d’écran, le 2022-06-15 à 09 01 50](https://user-images.githubusercontent.com/1424864/173869664-fc4d8e04-1a91-423a-ab81-b6cef963739d.png)

![Capture d’écran, le 2022-06-15 à 09 11 10](https://user-images.githubusercontent.com/1424864/173869776-2cf96a68-352e-43c8-9407-dcdc8ea5e5f1.png)

But, I cannot find the conflict.  `Istioctl analyze -A` does not report anything on more than one gateway : 

```
istioctl analyze -A | grep -i Error
Error: Analyzers found issues when analyzing all namespaces.
See https://istio.io/v1.13/docs/reference/config/analysis for more information about causes and resolutions.
Error [IST0101] (Gateway istio-gateways/gateway-for-all-apis) Referenced selector not found: "istio=ingressgateway"
Error [IST0101] (Gateway istio-gateways/gateway-for-services) Referenced selector not found: "istio=ingressgateway"
Error [IST0101] (Gateway istio-squelette/httpbin-gateway) Referenced selector not found: "istio=ingressgateway"
Error [IST0101] (Gateway test-automatique/test-automatique-gateway) Referenced selector not found: "app=istio-ingressgateway"
```

As can be seen in the screen capture above, there are a few Validation References, but if one clicks on the links, kiali cannot find the ressource : 

![Capture d’écran, le 2022-06-15 à 09 01 59](https://user-images.githubusercontent.com/1424864/173868006-a513ea1f-bf12-4775-afe3-aac94c86c912.png)

Warnings are not more helpful.  `kubectl `does not help either : 

```
> kc get gateways.networking.istio.io -n istio-system
No resources found in istio-system namespace.
> kc get ingress -n istio-system                                                                                   
No resources found in istio-system namespace.
``` 

Nor in the namespace : 

```
> kc get gw -n istio-squelette
NAME              AGE
httpbin-gateway   26d
> kc get virtualservices.networking.istio.io -n istio-squelette 
NAME       GATEWAYS                                  HOSTS                                 AGE
httpbin    ["istio-gateways/gateway-for-all-apis"]   ["api.kube.lab.ile.montreal.qc.ca"]   28d
httpbin2   ["httpbin-gateway"]                       ["httpbin.squl"]                      14d
```

[...]

PR #5345, opened 2022-07-27
Added support of K8s Gateway API objects

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5026

Added support of Gateway API objects: Gateway and HTTPRoute, for now only in Istio Config lists and details pages.
Separate filter types are added for both of them, accordingly "K8sGateway" and "K8sHTTPRoute".

Further PR for Graph.

Screenshot from 2022-09-13 13-18-08
Screenshot from 2022-09-13 13-17-49
Screenshot from 2022-09-13 13-17-04

For QE: Steps to add Gateway API into cluster https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/


josunect:
I guess this will also require changes in the Operator and Helm charts in order to give permissions for this API?

hhovsepy:
Need to add check of Gateway API CRD before registering object handlers in cache.

kubernetes/cache/istio.go
```diff
@@ -761,6 +786,128 @@ func (c *kialiCacheImpl) GetTelemetries(namespace, labelSelector string) ([]*v1a
...
+		obj, exist, err := nsCache[kubernetes.K8sGatewayType].GetStore().GetByKey(key)
+		if err != nil {
+			return nil, err
+		}
+		if exist {
+			l, ok := obj.(*gatewayapi.Gateway)
+			if !ok {
+				return nil, errors.New("bad K8sGateway type found in cache")
+			}
+			l.Kind = kubernetes.K8sGatewayType
```
nrfox, line 805: This is a bit of a problem since the actual Kind is `Gateway` and not `K8sGateway`. This is more of a problem with how Kiali handles Kind types in general. Kiali assumes that the Kind(s) are unique but as seen here they are not. The GVKs _are_ unique (Group + Version + Kind) but everywhere in Kiali we assume that the Kinds alone are unique including in the UI.

From the Istio Config page:
Screenshot from 2022-09-14 16-15-50

`K8sGateway` is not an actual Kind but we need some way of distinguishing it from the `networking.istio.io` `Gateway`. I'm not entirely sure how users perceive what the `Type` column represents there but I'm leery of setting this to a random `Kind` that doesn't actually exist.

The Istio config editor is a good example of this:
Screenshot from 2022-09-14 13-53-31

Compare this to the output of `kubectl`:
```
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"gateway.networking.k8s.io/v1alpha2","kind":"Gateway","metadata":{"annotations":{},"name":"gateway","namespace":"istio-ingress"},"spec":{"gatewayClassName":"istio","listeners":[{"allowedRoutes":{"namespaces":{"from":"All"}},"hostname":"*.example.com","name":"default","port":80,"protocol":"HTTP"}]}}
  creationTimestamp: "2022-09-14T20:47:04Z"
  generation: 1
  name: gateway
  namespace: istio-ingress
  resourceVersion: "4080"
```

It's confusing to have Kiali tell you one thing but `kubectl` and other tooling tell you another. Not sure what a good solution is here without doing a bunch of refactoring to include GVK in places where we always assume Kind. Perhaps we can display the correct `Gateway` kind in the istio config yaml editor at least?

hhovsepy: Yeah there is  a confusion between two Gateways on Kiali backend, also on UI side.
I provided a solution to differentiate two types of objects.
That YAML editor value was hard-coded to show G8sGateway then it changes to Gateway while pushing.
But at least I can change how it is displayed in UI. Then question is the filters, which now have K8s Types of filters..

hhovsepy:
Hi @jshaughn @nrfox it is rebased with latest cache changes. Ready for review. @josunect you recently added new object types support, so it will be fresh you to review.
This PR contains support of Gateway API in Istio config list and details pages.

frontend/src/pages/IstioConfigList/FiltersAndSorts.ts
```diff
@@ -93,6 +93,14 @@ export const istioTypeFilter: FilterType = {
...
       title: 'Gateway'
     },
+    {
+      id: 'K8sGateway',
+      title: 'K8sGateway'
+    },
+    {
+      id: 'K8sHTTPRoute',
+      title: 'K8sHTTPRoute'
+    },
```
nrfox, line 103: Under the hood, Istio creates a `VirtualService` and `istio.Gateway` when a `K8sGateway` and `K8sHTTPRoute` is created. I wonder if we want to filter out those underlying objects?

Screenshot from 2022-09-19 12-00-43

Screenshot from 2022-09-19 12-00-15

If so, we could filter out anything with an `internal.istio.io/parent` annotation.

kubernetes/cache/istio.go
```diff
@@ -118,6 +125,23 @@ func (c *kialiCacheImpl) createIstioInformers(namespace string) istio.SharedInfo
 	return sharedInformers
 }
 
+func (c *kialiCacheImpl) createGatewayInformers(namespace string) gateway.SharedInformerFactory {
+	sharedInformers := gateway.NewSharedInformerFactory(c.gatewayApi, c.refreshDuration)
+	lister := c.getCacheLister(namespace)
+
+	if c.istioClient.IsGatewayAPI() {
```
nrfox, line 132: Something I wasn't expecting is that Kiali is automatically picking up the gateway API objects without a restart. It's also working fine without having the gateway API CRDs but I was expecting to have to restart the kiali pod before seeing the new Gateway objects.

kubernetes/cache/cache.go
```diff
@@ -242,6 +250,12 @@ func (c *kialiCacheImpl) createNSCache(namespace string) bool {
 			return false
 		}
 	}
+	for _, synced := range gatewayInformerFactory.WaitForCacheSync(c.stopNSChans[namespace]) {
+		if !synced {
+			log.Errorf("[Kiali Cache] failed to sync [namespace: %s] Istio cache", namespace)
```
nrfox, line 255:
```suggestion
			log.Errorf("[Kiali Cache] failed to sync [namespace: %s] gateway-api cache", namespace)
```

nrfox (commented):
Some minor comments but otherwise the code LGTM. Want to hear your thoughts on hiding the underlying Istio objects for the `K8sGateway` and `K8sHTTPRoute` objects.

business/istio_validations.go
```diff
@@ -318,7 +322,7 @@ func (in *IstioValidationsService) fetchAllWorkloads(ctx context.Context, rValue
 
 		allWorkloads := map[string]models.WorkloadList{}
 		for _, ns := range nss {
-			criteria := WorkloadCriteria{Namespace: ns.Name, IncludeIstioResources: true, IncludeHealth: false}
+			criteria := WorkloadCriteria{Namespace: ns.Name, IncludeIstioResources: false, IncludeHealth: false}
```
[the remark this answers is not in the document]
hhovsepy: Actually this was not related to Gateway API, it is optimization. We do not need Workload validations itself while validating configs and config validations require only workloads itself.

frontend/src/pages/IstioConfigList/FiltersAndSorts.ts
hhovsepy: Showing those autogenerated objects was added recently based on community issue https://github.com/kiali/kiali/issues/5213, as there was a confusion with registry only objects. So they are listed and shown as read-only..

kubernetes/cache/istio.go
hhovsepy: Restart of Kiali pod is required after adding (or removing) Gateway API CRDs only. Any further restart is not required while creating Gateway objects, as they are already registered in cache listeners.

kubernetes/kubernetes.go
```diff
@@ -206,6 +206,22 @@ func (in *K8SClient) IsOpenShift() bool {
...
 }
 
+func (in *K8SClient) IsGatewayAPI() bool {
+	if in.GatewayAPI() == nil {
+		return false
+	}
+	if in.isGatewayAPI == nil {
+		isGatewayAPI := false
+		// GatewayClass should exist got K8s Gateway API to work.
+		_, err := in.GatewayAPI().GatewayV1alpha2().GatewayClasses().List(in.ctx, emptyListOptions)
```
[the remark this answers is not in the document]
hhovsepy: Here any error reason means Gateway API is not supported. So probably better to log any error different from NotFound.

hhovsepy:
Thanks for the review @nrfox.

kubernetes/cache/istio.go
nrfox: > Restart of Kiali pod is required after adding (or removing) Gateway API CRDs only. 

Yes! I was expecting to have to restart after adding the Gateway API CRDs but in fact Kiali picked them up automatically without needing to restart. I'm not sure if that's a good or bad thing to be honest. Something may be recreating the cache unexpectedly.

business/istio_config.go
```diff
@@ -273,6 +290,40 @@ func (in *IstioConfigService) GetIstioConfigList(ctx context.Context, criteria I
...
+	}(ctx, errChan)
+
+	go func(ctx context.Context, errChan chan error) {
+		defer wg.Done()
+		if in.k8s.IsGatewayAPI() && criteria.Include(kubernetes.K8sHTTPRoutes) {
+			// ignore an error as system could not be configured to support K8s Gateway API
+			// Check if namespace is cached
+			if IsResourceCached(criteria.Namespace, kubernetes.K8sHTTPRoutes) {
+				istioConfigList.K8sHTTPRoutes, _ = kialiCache.GetK8sHTTPRoutes(criteria.Namespace, criteria.LabelSelector)
+			}
```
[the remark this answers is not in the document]
hhovsepy: In upper "in.k8s.IsGatewayAPI()" chec filters out the cases when Gateway API is not configured, so good idea to add all errors into errChan.

hhovsepy:
>  hiding the underlying Istio objects for the `K8sGateway` and `K8sHTTPRoute` objects.
I think at this step we need to show both Gateways objects, as Gateway API integration in Istio is still in progress.
But ideally filters would be automatically changed based on CRD, for now this  will require some refactoring, so leaving it for later.

kubernetes/cache/istio.go
```diff
@@ -118,6 +125,23 @@ func (c *kialiCacheImpl) createIstioInformers(namespace string) istio.SharedInfo
...
+		if c.CheckIstioResource(kubernetes.K8sGateways) {
+			lister.k8sgatewayLister = sharedInformers.Gateway().V1alpha2().Gateways().Lister()
+			sharedInformers.Gateway().V1alpha2().Gateways().Informer().AddEventHandler(c.registryRefreshHandler)
+		}
+		if c.CheckIstioResource(kubernetes.K8sHTTPRoutes) {
+			lister.k8shttprouteLister = sharedInformers.Gateway().V1alpha2().HTTPRoutes().Lister()
+			sharedInformers.Gateway().V1alpha2().Gateways().Informer().AddEventHandler(c.registryRefreshHandler)
```
nrfox, line 139:
```suggestion
			sharedInformers.Gateway().V1alpha2().HTTPRoutes().Informer().AddEventHandler(c.registryRefreshHandler)
```

kubernetes/kubernetes.go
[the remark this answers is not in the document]
nrfox: Well it could be that Kiali is just having temporary problems talking to the kube API. We could retry for some subset of errors like `IsTimeout` or `IsTooManyRequests`. Another scenario is that someone could've deployed only _some_ of the CRDs but not all e.g. `GatewayClasses` were deployed but not `Gateways` or `HTTPRoutes`. I don't think we need to overcomplicate things though and probably just checking for one or two errors is good enough.

I actually get `Forbidden` errors when the Gateway API CRDs are not present which could either mean that the CRDs do not exist _or_ that the CRDs do exist and Kiali doesn't have access to them. Since we can't differentiate between the two scenarios, what do you think about just logging a `Warning` level log in case of any error?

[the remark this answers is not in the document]
nrfox: Have you tried this when Kiali doesn't have cluster scoped permissions? i.e. `accessible_namespaces != **`? The cluster scoped list may always fail in this scenario.

kubernetes/cache/istio.go
nrfox: Ah I figured out why these objects were showing up even with the Gateway APIs disabled. That info is coming from the registry.

kubernetes/kubernetes.go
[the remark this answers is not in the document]
nrfox: Seeing a bunch of:
```
2022-09-20T14:38:47Z ERR Error checking Kubernetes Gateway API configuration: gatewayclasses.gateway.networking.k8s.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "gatewayclasses" in API group "gateway.networking.k8s.io" at the cluster scope
```
when you visit the apps list page. This looks like the only place where we list `GatewayClasses`.

josunect:
I was testing with minikube and crc. 
The cache doesn't throw errors if the GW API is not installed. 
Once installed it is possible to see the new types: 
image

The new resources can be modified, filtered and removed.

image

kubernetes/kubernetes.go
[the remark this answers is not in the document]
nrfox: I think the reason I'm seeing multiple error messages is that Kiali creates multiple `K8sClient`. The layer creates one, the cache creates another, and the client factory creates them as needed. Having the `IsGatewayAPI` check on the `K8sClient` means that check will happen each time a new client is initialized which happens more than once in Kiali. That's probably fine and restarting kiali once you deploy the CRDs into the cluster will still fix everything.

The bigger issue is the cluster scoped list call to GatewayClasses. With `spec.deployment.accessible_namespaces = bookinfo`:
```
kubectl auth can-i list gatewayclasses --as system:serviceaccount:istio-system:kiali-service-account
> Warning: resource 'gatewayclasses' is not namespace scoped in group 'gateway.networking.k8s.io'
> no
kubectl auth can-i list gatewayclasses --as system:serviceaccount:istio-system:kiali-service-account -n bookinfo
> Warning: resource 'gatewayclasses' is not namespace scoped in group 'gateway.networking.k8s.io'
> yes
```

One potential workaround is to do a namespaced scoped call using the `Config.IstioNamespace`. This will work for operator deployments since the operator automatically gives Kiali a role for this namespace but I don't think that happens automatically for helm deployments. Another option is to use one of the namespaces in `spec.deployment.accessible_namespaces`. These can be regexs though so you might need to first use the `NamespaceService` to get a list of name of accessible namespaces. Unless there's another way test to see if the CRDs have been deployed?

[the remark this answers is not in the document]
hhovsepy: I had tested both cases accessible_namespaces="**" and accessible_namespaces = "bookinfo" on OCP installed kiali via operator and deployed on local machine. When CRD is installed or not. In all cases I did not see any error "Error checking Kubernetes Gateway API configuration....", even after listing apps. Not sure how it could happen on your env.
Only case when bunch of errors logged is the case when CRD is removed and before Kiali pod is restarted.
So at this point I will suggest to leave Gateway API recognition as it is,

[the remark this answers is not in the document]
nrfox: > So at this point I will suggest to leave Gateway API recognition as it is, only change the Error log level to Warn, not to make confusion, then wait for community feedback. As Gateway API itself is still in WIP.

My only concern is that for users who set `accessible_namespaces` to something other than `**`, Kiali will be logging error messages about gateway APIs even if they deploy the CRDs. With most of the info coming from the registry anyways, I don't think it'll impact the user experience much for right now with the gateway api info only being listed in the config.

nrfox (approved):

PR #5348, opened 2022-07-28
[Cypress] istio config editor test

ScriptingShrimp:
Adding new testcase for istio config editor, for now its only simple test.
Also there is new method for handling anchors in tables.

### DIFF
--- a/frontend/cypress/integration/common/istio_config_editor.ts
+++ b/frontend/cypress/integration/common/istio_config_editor.ts
@@ -1,0 +1,5 @@
+import { Then } from 'cypress-cucumber-preprocessor/steps';
+
+Then('user can see istio config editor', () => {
+	cy.get('#ace-editor').should('be.visible');
+})
--- a/frontend/cypress/integration/common/table.ts
+++ b/frontend/cypress/integration/common/table.ts
@@ -30,6 +30,10 @@ And('the {string} column on the {string} row is empty', (column: string, rowText
   getColWithRowText(rowText, column).children().should('be.empty');
 });
 
+And('user clicks in {string} column on the {string} text', (column: string, rowText: string) => {
+  getColWithRowText(rowText, column).find('a').click();
+});
+
 Then('user sees {string} in the table', (service: string) => {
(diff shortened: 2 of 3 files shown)


frontend/cypress/integration/common/istio_config_editor.ts
```diff
@@ -0,0 +1,8 @@
+import { Then } from 'cypress-cucumber-preprocessor/steps';
+
+Then('user can see isto config editor', () => {
+    let editor = cy.get('.pf-c-drawer__body')
+    editor.should('be.visible')
+    editor.find('#ace-editor').should('be.visible')
```
nrfox, line 6: A few small comments but otherwise LGTM:
Assigning from return values is [generally frowned upon in Cypress](https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Return-Values). You can chain here instead `cy.get('.pf-c-drawer').should('be.visible').find('#ace-editor')...`.

The class selector `.pf-c-drawer__body` is not great to use as it is subject to change when/if the patternfly library changes. If you just want the `ace-editor`, you can do `cy.get('#ace-editor')` or if you want to ensure the ace editor is within the card body, you can add a `data-test` tag to the react component that renders the body and search for that: https://github.com/kiali/kiali/blob/master/frontend/src/pages/IstioConfigDetails/IstioConfigDetailsPage.tsx#L515.

The formatting is lacking some semi colons and a newline at the end of the file. If you use vscode you can install the prettier addon and enable it to run "on save". That should automatically take care of all the formatting issues for you.

ScriptingShrimp: Yes, you are right about chaining, also I changed selector for the test :+1: 

Added proper indentation and useless semicolon :grin: 

What kind of prettier do you use?

nrfox: I use [this](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) vscode plugin and it just runs on save.

nrfox (approved):

PR #5351, opened 2022-08-01
Upgrade go version to 1.18

### LINKED ISSUE #5538: CVEs in latest distroless image due to 1.17.10 of go
REMOVED details of CVEs
Looks like upgrading to go 1.17.13 would fix these.  Is this a change that can be made?

### LINKED ISSUE #5561: Bump go version to 1.18 (or 1.19?)
For a security standpoint, we need to bump Go to a recent version.

aljesusg:
Closes #5561 
Fixes #5538

Upgrade go version 1.18.4


jshaughn (commented):
Technically I don't think there would be an issue moving to Go 1.18, which should be backward compatible.   The question I have is only around compatibility with upstream console, and the plugin effort.  Also, to some degree, whether there is any build impact against older supported versions.

jmazzitelli:
> The question I have is only around compatibility with ... the plugin effort

The [plugin](https://github.com/kiali/openshift-servicemesh-plugin/tree/main/plugin) doesn't use Go - strictly javascript.
The [plugin operator](https://github.com/kiali/openshift-servicemesh-plugin/tree/main/operator) doesn't use Go - strictly Ansible.

lucasponce:
I wonder if:
- Keep the latest go 1.17 to be aligned with Istio 1.14.x https://github.com/istio/istio/blob/1.14.3/go.mod (if the issue is addressed there).
- Bump to go 1.18 when Kiali bumps to Istio 1.15 https://github.com/istio/istio/blob/master/go.mod#L3

In theory, it should work but using the same version of the dependencies https://github.com/istio/client-go/blob/v1.14.3/go.mod#L3 always helps to avoid potential issues.

But if the testing is solid, I imagine all would be good.

aljesusg:
Mmm in this case I prefer to wait until istio 1.15 to be aligned.

What do you think?

nrfox:
@aljesusg with istio 1.15 released can this be worked on again?

jshaughn:
From Alberto:

> I wonder if:
Keep the latest go 1.17 to be aligned with Istio 1.14.x https://github.com/istio/istio/blob/1.14.3/go.mod (if the issue is addressed there).
Bump to go 1.18 when Kiali bumps to Istio 1.15 https://github.com/istio/istio/blob/master/go.mod#L3
In theory, it should work but using the same version of the dependencies https://github.com/istio/client-go/blob/v1.14.3/go.mod#L3 always helps to avoid potential issues.
But if the testing is solid, I imagine all would be good.
I don't know implications for 1.19 but again I need to check downstream @leandroberetta
Well this is a tricky question when upstream and downstream build with the same Go version. 1.57 is already targeted to 1.15 upstream. 1.59 will be targeted to 1.16 upstream. Yet 1.57 will be used with 1.14 downstream. We don't have to backport to previous versions, but we have already had requests to bump Go in the upstream.

I think we should proceed with updating to v1.18 for this sprint (1.59). We probably should backport the change to 1.58 becaue it will be the highest recommended version for use with 1.15. We do not have to backport to 1.57 because it will be superseded by 1.58 for the upstream use, and then it will remain at 1.17 for downstream 2.3/1.14.

jshaughn:
@aljesusg Can we remove the Do Not Merge label?  I know the PR is not ready but it is active for this sprint, right?

aljesusg:
Yes, I am checkin that all is running correctly

aljesusg:
I disabled sa1029 while I check hwo to fix it.

handlers/metrics_test.go:348:57: SA1029: should not use built-in type string as key for value; define your own type to avoid collisions (staticcheck)
	req = req.WithContext(context.WithValue(req.Context(), "authInfo", &api.AuthInfo{Token: "test"}))


If I use a constant to do that or another way the tests fail I need to check the latest changes

graph/config/cytoscape/cytoscape.go
```diff
@@ -251,13 +255,7 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
 
 		addNodeTelemetry(n, nd)
 
-		// set annotations, if available
-		if val, ok := n.Metadata[graph.HealthData]; ok {
-			nd.HealthData = val.(interface{})
-		}
-		if val, ok := n.Metadata[graph.HealthDataApp]; ok {
-			nd.HealthDataApp = val.(interface{})
-		}
```
nrfox, line 260: @aljesusg this could be the cause of the frontend failures from CI. You'll likely need to keep the assignment but maybe remove the type cast to `interface{}` to make the linter happy.

graph/config/cytoscape/cytoscape.go
```diff
@@ -251,12 +255,12 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
 
 		addNodeTelemetry(n, nd)
 
-		// set annotations, if available
+		// HealthData && HealthDataApp S1040: type assertion to the same type: val already has type interface{}
```
nrfox, line 258: nit: you don't need to add the linting rule here as a comment.

config/authentication.go
```diff
@@ -0,0 +1,17 @@
+package config
```
nrfox, line 1: I don't think this belongs in the `config` package. Can it be moved to `handlers`?

.github/workflows/build-backend.yml
```diff
@@ -19,18 +19,11 @@ jobs:
...
-            ~/go/pkg/mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-          restore-keys: |
-            ${{ runner.os }}-go-
-
+          go-version-file: go.mod
+          # The builtin cache feature ensures that installing golangci-lint
+          # is consistently fast.
+          cache: true
+          cache-dependency-path: go.sum
```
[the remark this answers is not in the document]
jshaughn: +1, this looks like a nicer approach.

Makefile
```diff
@@ -29,7 +29,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
-GO_VERSION_KIALI = 1.17.10
+GO_VERSION_KIALI = 1.18.7
```
jshaughn, line 32: I see there is now a 1.18.8, but I think we need to use what is available in Brew, right?

config/authentication.go
jshaughn: I think there is already an `authentication.go` in `handlers`.  Perhaps this can be merged in?   Or it can stay here if you think it makes more sense, but I agree that I don't currently see why this is config.

go.mod
```diff
@@ -1,87 +1,91 @@
...
+	go.opentelemetry.io/otel v1.11.1
+	go.opentelemetry.io/otel/exporters/jaeger v1.11.1
+	go.opentelemetry.io/otel/sdk v1.11.1
+	go.opentelemetry.io/otel/trace v1.11.1
+	go.uber.org/automaxprocs v1.5.1
+	golang.org/x/sync v0.1.0
+	google.golang.org/genproto v0.0.0-20221018160656-63c7b68cfc55
+	google.golang.org/grpc v1.50.1
+	google.golang.org/protobuf v1.28.1
+	gopkg.in/square/go-jose.v2 v2.6.0
```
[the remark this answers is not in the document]
jshaughn: The prometheus updates make me a little nervous, just need to make sure the updated versions are compatible with the Prom versions we support for upstream and downstream.

graph/config/cytoscape/cytoscape.go
```diff
@@ -1,17 +1,19 @@
...
+Main page:   http://js.cytoscape.org/
+JSON config: http://js.cytoscape.org/#notation/elements-json
+Demos:       http://js.cytoscape.org/#demos
+
+Algorithm: Process the graph structure adding nodes and edges, decorating each
+with information provided.  An optional second pass generates compound
+nodes for requested boxing.
+
+The package provides the Cytoscape implementation of graph/ConfigVendor.
+*/
```
jshaughn, line 16: Was this the linter or just personal preference?

graph/protocol.go
```diff
@@ -38,9 +38,9 @@ type Protocol struct {
 // Each supported protocol is defined below.  Each rate provided as node or edge metadata must be defined.
 // Each method below should have a section handling each supported protocol.
 
-//
-// GRPC Protocol
-//
+/*
+GRPC Protocol
+*/
```
jshaughn, line 43: same as above, I don't think we should be changing the comment style of the original author, unless there is some good reason.  Files are being updated in this PR just to change comment style or indentation.

jshaughn (commented):
Very close, a couple of comments

.github/workflows/build-backend.yml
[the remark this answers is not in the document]
aljesusg: Yes is read it from the go.mod

Makefile
aljesusg: Yep in brew we have only until 1.18.7

config/authentication.go
aljesusg: Well the point is that we need to have this in the same file I don't care about where I thought in the config by the "authInfo" variable let me merge this in handlers

aljesusg: mmmm we'll have import cycle if merge this in handlers/authentication

graph/telemetry/istio/istio.go
```diff
@@ -910,7 +910,8 @@ func promQuery(query string, queryTime time.Time, api prom_v1.API) model.Vector
 
 	promtimer := internalmetrics.GetPrometheusProcessingTimePrometheusTimer("Graph-Generation")
 	value, warnings, err := api.Query(ctx, query, queryTime)
-	if warnings != nil && len(warnings) > 0 {
+	// Any slice that has a len greater than zero is a non-nil slice
```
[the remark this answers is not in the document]
aljesusg: You are right the warning was

S1009: should omit nil check; len() for github.com/prometheus/client_golang/api/prometheus/v1.Warnings is defined as zero (gosimple)

models/health.go
```diff
@@ -71,8 +71,8 @@ type WorkloadHealth struct {
 // In healthy scenarios all variables should point same value.
 // When something wrong happens the different values can indicate an unhealthy situation.
 // i.e.
-// 	desired = 1, current = 10, available = 0 would means that a user scaled down a workload from 10 to 1
-//  but in the operaton 10 pods showed problems, so no pod is available/ready but user will see 10 pods under a workload
+// - desired = 1, current = 10, available = 0 would means that a user scaled down a workload from 10 to 1
+// - but in the operaton 10 pods showed problems, so no pod is available/ready but user will see 10 pods under a workload
```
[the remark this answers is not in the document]
aljesusg: Yes... :(

prometheus/metrics.go
```diff
@@ -52,7 +52,8 @@ func fetchHistogramValues(ctx context.Context, api prom_v1.API, metricName, labe
 	for k, query := range queries {
 		log.Tracef("[Prom] fetchHistogramValues: %s", query)
 		result, warnings, err := api.Query(ctx, query, queryTime)
-		if warnings != nil && len(warnings) > 0 {
+		// Any slice that has a len greater than zero is a non-nil slice
+		if warnings != nil {
```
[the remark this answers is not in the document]
aljesusg: Yes, the warning was S1009: should omit nil check; len() for github.com/prometheus/client_golang/api/prometheus/v1.Warnings is defined as zero (gosimple)

aljesusg:
well maybe I did a go mod tidy .

Testing with:
go mod edit -go=1.18
 go get istio.io/client-go@v1.15.2

prometheus/metrics.go
```diff
@@ -94,7 +95,8 @@ func buildHistogramQueries(metricName, labels, grouping, rateInterval string, av
 func fetchRange(ctx context.Context, api prom_v1.API, query string, bounds prom_v1.Range) Metric {
 	log.Tracef("[Prom] fetchRange: %s", query)
 	result, warnings, err := api.QueryRange(ctx, query, bounds)
-	if warnings != nil && len(warnings) > 0 {
+	// Any slice that has a len greater than zero is a non-nil slice
+	if warnings != nil {
```
jshaughn, line 99:
```suggestion
	if len(warnings) > 0 {
```

jshaughn (changes requested):
You need to be careful to catch all instances of the same change.  Please accept the suggested changes I added in the comments.  I suggested a different package for that code-move that I think would work and may make more sense that `config`.  See what you think.  I'm not sure about all of the comment changes, is that all due to the new linter, or is you IDE?  I just want to make sure that we don't get into an IDE formatting war.

prometheus/metrics.go
```diff
@@ -175,7 +175,7 @@ func getRequestRatesForLabel(ctx context.Context, api prom_v1.API, time time.Tim
 	log.Tracef("[Prom] getRequestRatesForLabel: %s", query)
 	promtimer := internalmetrics.GetPrometheusProcessingTimePrometheusTimer("Metrics-GetRequestRates")
 	result, warnings, err := api.Query(ctx, query, time)
-	if warnings != nil && len(warnings) > 0 {
+	if len(warnings) > 0 {
 		log.Warningf("fetchHistogramValues. Prometheus Warnings: [%s]", strings.Join(warnings, ","))
```
jshaughn, line 179:
```suggestion
		log.Warningf("getRequestRatesForLabel. Prometheus Warnings: [%s]", strings.Join(warnings, ","))
```

jshaughn (changes requested):
Almost there.  The CI is now failing.  I suggested one change for a small issue in an error message, not something you actually changed.  Were you able to ignore those jaeger pb.go files from the linter (I'mreferring to an older comment from Nick)?

jshaughn (approved):
Code changes look good to me.  You didn't really answer my questions about the changes to comments, as to whether they were required by the new linter or just things you wanted to change.  I'll assume they were forced by the linter.  Approved!

aljesusg:
Yes sorry . Of course I hate spend time change unnecessary code like comments 😅😅 I tried gofmt and others and linter throw a lot of warnings due spaces, tabs, indentation 😔

nrfox (approved):
LGTM

PR #5357, opened 2022-08-01
Allow custom names in perf script and install wklds

nrfox:
Adds option to specify a custom cluster name and installs demo workloads to generate load by default.


jmazzitelli (approved):
should we do any of these changes in the original ibmcloud-openshift.sh hack script? Or is this new stuff only needed for larger clusters for performance testing?

nrfox:
> should we do any of these changes in the original ibmcloud-openshift.sh hack script? Or is this new stuff only needed for larger clusters for performance testing?

The cluster name is just passing through an option that already exists in the ibm script and the workloads are not ibmcloud related but rather part of the post-infra setup so neither is needed there.

Issue #5289, opened 2022-07-06
logs tab page has wrong container when navigating from trace tab 


jmazzitelli:
This is a hard one to explain how to replicate, but it is easy to see the bug once you do it.

To replicate:

1. Install Kiali and Istio in your cluster
2. Install the travel demo (`./hack/istio/install-travel-agency-demo.sh`).
3. Open an incognito window (you want a clean cache) and navigate to the Kiali UI.
4. Click "Workloads" left-hand menu, select "travel-portal" in the namespace dropdown menu, and click to navigate to the "travels" workload (`https://kiali-istio-system.apps.../console/namespaces/travel-portal/workloads/travels`)
5. Select the "Traces" tab to view the traces timeline
6. Click any bubble to view the traces
7. Select the "Span Details" sub-tab to view all the traces
8. Look in the list, under the "Summary" column and find a row with "Operation" that starts with "cars." If you can't find one, go back to step 6 and try another bubble.
9. In that row with "Operation" of "cars", click the kabob menu and select "Logs"
10. You are now in the "Logs" tab of the workload. Notice the bug.

Weirdness you will see:

First, you will see in the log window a single line with the error message `Failed to fetch workload logs: pods "travels-6c8f5b5f5c-5qdj5" not found`. Well, that is not the name of the selected pod - the dropdown menu for the pod shows `travels-v1-<hash>` but this error message indicates it wants to load logs for `travels-<hash>`. Note also I have seen this where the selected pod is "cars-v1-<hash>" - I can't figure out where its getting the initial value of this dropdown, but it is incorrect.

Wait for the auto-refresh to happen (or just click the refresh button to manually refresh) and notice the error message changes to be: `Failed to fetch workload logs: container control is not valid for pod travels-v1-77b6c94545-m7krt` (sometimes I see it with a cars-v1 pod name in the message like `Failed to fetch workload logs: container control is not valid for pod cars-v1-85d6f94474-52cft`).

Notice also the containers that you can select with checkboxes. One container is called "control". Now select the pod from the drop down menu. It already has selected the "travels-v1" pod (note! "-v1"), but the drop down now only shows "travels" (NO -v1 in the name). Select it. Notice now the checkbox "control" changes its name - it is now "travels"... and now, with the "travels" checkbox checked, you can see all the logs.

So, there is a lot wrong going on here. I can't figure out what it is supposed to be doing, but what it is doing is broken.

jshaughn:
cc @israel-hdez

[...]

PR #5363, opened 2022-08-03
Logs tab page has wrong container when navigating from trace tab

josunect:
The error seems to be reproduced when selecting a trace that belongs to a pod from another namespace. 

The WorkloadPodLogs component is initialized using: 

```
namespace={this.props.match.params.namespace}
workload={this.props.match.params.workload}
pods={this.state.workload!.pods}
```
The workload and the namespace are correct, because they come from the current call, but, the pod is retrieved from the current state, which is not yet updated, because the API call to fetch the current workload has not finished. 

This fix uses await to make the fetchWorkload async and update the state once the fetch has finished.  

Fixes https://github.com/kiali/kiali/issues/5289


jmazzitelli (approved):
code looks good and I ran this following the replication procedures documented in the issue and I no longer see the bug.

PR #5367, opened 2022-08-04
k8s autogenerated objects are treated differently.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/5213.

When Istio is configured to auto generate k8s ingress Gateways, those Gateways are available in Istio cache registry, but in fact do not exist in k8s API, so retrieving details of them is failing.
But they are causing troubles in config list and in validations, making not existing references.

This PR is a ignoring those Gateways in Validations.

Also, the Istio config details which was failing before, shows read-only result from Registry.
Screenshot from 2022-08-05 09-59-34

Screenshot from 2022-08-05 10-00-03


lucasponce:
@hhovsepy mmm I'm not 100% sure of filtering these.

I remember that the federation implementation in OSSM relies on "memory" resources that are available in the Istio Registry but not as a k8s resource.

Perhaps it may need something in the middle (I think) like to give that information to the user "this is a memory-based gateway," but don't let them navigate to the detail or another option (like allow them to navigate but not to operate, i.e. update/delete).

In short, IMHO memory-only resources should be visible for users, but I agree that they should be treated in a different way.

But thanks for catching this @hhovsepy ! It's a very valid scenario.

hhovsepy:
@lucasponce the initial [issue](https://github.com/kiali/kiali/issues/5213) was, that Gateway validations were considering objects from registry, thus having cache only objects from registry and making not existing references to them. I was not able to reproduce that before.
Now, after Istio config lists optimizations, those non existing (k8s generated) configs started to appear in Istio config list page of Kiali, because now it reads from registry (AllNamespaces=true).

lucasponce:
Yes, I know, but my point is, if those "memory configurations" are part of the Istio resources, those should be valid as well.
I understand that this may trigger a bigger "Epic," but I think Kiali should also handle those configs somehow (indicating the difference, or if it's source only).

I mean, it's similar to the Federation logic we have, I don't mean to block the PR, but perhaps we can describe these cases well and fill Epic to plan a major change in the short term.

I understand that an auto-generated config of a Gateway can also affect other parts, which is important information that users should have.

hhovsepy:
Thanks @lucasponce  I agree to split this task, for now it ignores autogenerated Gateways in validations.
The further step will be making Config details read only. This will require additional feature to be added to retrieve single config from registry.

hhovsepy:
Now this PR contains both changes:
- Autogenerated Gateways are not considered in validations.
- Registry only Istio objects details are shown as read-only.

business/istio_config.go
```diff
@@ -542,6 +542,158 @@ func (in *IstioConfigService) GetIstioConfigDetails(ctx context.Context, namespa
...
+		return istioConfigDetail, errors.New("RegistryConfiguration is nil. This is an unexpected case. Is the Kiali cache disabled ?")
+	}
+
+	switch objectType {
+	case kubernetes.DestinationRules:
+		configs := registryConfiguration.DestinationRules
+		for _, cfg := range configs {
+			if cfg.Name == object && cfg.Namespace == namespace {
+				istioConfigDetail.DestinationRule = cfg
+				istioConfigDetail.Gateway.Kind = kubernetes.DestinationRuleType
```
lucasponce, line 581: This logic is not 100% clear to external reviewers.
I understand that you iterate on the Istio registry to pick the memory config, but is the auto-generated Gateway linked with the other objects?

Perhaps is it being re-used the Gateway model just for convenience?

lucasponce (commented):
I'm still trying to create an end-to-end flow for testing this.
I understand that these "autogenerated gateways" are created in the context of the integration of Istio with the "cert-manager" and "Kubernetes Ingress" objects, isn't it? [1]

https://istio.io/latest/docs/ops/integrations/certmanager/#kubernetes-ingress [1]

In those cases, yeah, the auto-generated config would add noise to the Istio users, and this PR should help.

I see it as a useful corner case for those cases, but it won't hit regular flow.

I put a couple of comments, but I'm fine with the work.

I'd like to have some local tests before final approval, @hhovsepy, did you have some hack script that could simulate part of the flow?

business/istio_config.go
hhovsepy: oops, that is a typo

hhovsepy:
Right, those objects are auto generated for Kubernetes ingress and internal use.
Besides Gateway I also see VS auto generated, so I am going to ignore validations for VS as well:
Screenshot from 2022-08-08 13-58-04
Screenshot from 2022-08-08 13-57-41

Currently there is not a constant steps to reproduce the environment, I did applied manifest "istioctl manifest apply --set k8sIngress.enabled=true".

lucasponce:
Thanks, @hhovsepy, yes I can reproduce the scenario to test this.
Basically, following the "Kubernetes Gateway API" https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/ will help.

As it's described in the task, the mapping between the Kubernetes API and the Istio API is not 1-to-1:

```
$ kubectl get crds | grep k8s.io
gatewayclasses.gateway.networking.k8s.io      2022-08-09T10:06:09Z
gateways.gateway.networking.k8s.io            2022-08-09T10:06:09Z
httproutes.gateway.networking.k8s.io          2022-08-09T10:06:09Z
referencepolicies.gateway.networking.k8s.io   2022-08-09T10:06:09Z
tcproutes.gateway.networking.k8s.io           2022-08-09T10:06:09Z
tlsroutes.gateway.networking.k8s.io           2022-08-09T10:06:09Z
udproutes.gateway.networking.k8s.io           2022-08-09T10:06:09Z
```

Then, Istio seems to detect the Kubernetes Gateway and create the proper Istio ones in MEMORY in the Istio Registry.

Now I can see the failures and I'm ready to test the code.

lucasponce:
Ok, now more useful feedback.
With the [1] https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/ example using "minikube" I can see some auto-generated resources, but it seems there are some validations out-of-sync, between pages and details.

[a] Overview (look into the errors in the "default" one)
image

[b] Istio Config list:
image

[c] Details:
image

So, probably user may feel confused with a false negative, or if we ignore these resources for validations, those namespaces should look "green" in these cases.

lucasponce:
Another comment, I think that the parsing of the Istio registry configuration should take into consideration the apis.

For example, this "gateway" one:
image
image

It shouldn't appear:
image

As it's an implementation of the "gateway.networking.k8s.io" one, that's the reason that it has an empty "spec".

So, if we don't show those objects yet, that's good; I wouldn't introduce them.

Or another option is that we can wait until the "networking.k8s.io" is managed first; all the resources are also available from the registry. But It would require more work to implement the CRUD operations.

lucasponce (changes requested):
Work is progressing now the auto-generated objects are processed, and those errors are gone, but validations are not in sync, plus some side comments on the parsing.

lucasponce:
Another comment that I think belongs to this work.
I don't have a suggestion, but I think that the user should have a way to know why a "memory" resource (autogenerated) is not editable:

image

Like a tooltip or just a better/clear message in the side panel that indicates that it's an "Istio Registry" memory entry.

Or perhaps the annotations are clear enough?
image

lucasponce:
Another comment, these "autogenerated" objects are important, as is the active config that is sent into Envoy:
image

hhovsepy:
The appearance of "gateway.networking.k8s.io" Gateways is because of loading resources from Registry. It happened before this PR.
Fixed the validation inconsistency.
Added help message for read only.
Screenshot from 2022-08-09 18-26-59

lucasponce:
> The appearance of "gateway.networking.k8s.io" Gateways is because of loading resources from Registry. It happened before this PR.

Correct. My comment is that it creates a duplicate in the first one:

image

I think this one may be skipped if Kiali doesn't support them yet, as it's a "networking.k8s.io" resource.

Thanks for the changes, I'll prepare the environment tomorrow, and I'll re-test.

lucasponce:
I'm still seeing the problem of lack of sync validations.

[1] Overview of the "default" namespace:
image

[2] Istio Config for the "default" namespace:
image

[3] Details for the offending VirtualService:
image

So, these three info should be consistent, we can choose to skip these objects completely, or include them in the validations at all levels, but having a mix of messages would create more perception that something is not synced, I think.

lucasponce:
I did a fresh PR, my previous comment is gone, but I still see the wrong parsing of a gateway under the "istio-ingress" namespace:

image

The first Gateway with "gateway" name is not an Istio resource:
image

That creates an empty "gateway" when getting into the details:
image

I think this is because when parsing the "Gateway" objects from the Istio Registry, there is now a situation where the same "kind" "Gateway" is reused for "istio.io" and "k8s.io", so I think this one should be skipped.

Issue #5070, opened 2022-05-13
k8s api token not auto refreshing (aws eks warning)


ldemailly:
Here is what we got from amazon with arn:aws:eks:XXXXX|istio-system:kiali (xxx is our account/region/clustername details) being the kiali 1.48 pod - a colleague upgraded to 1.50 and still see the warnings

What is surprising is that the k8s api client I see in go.mod seems above the version they mention as min required... 

----
Hello,

We have identified applications running in one or more of your Amazon EKS clusters that are not refreshing service account tokens. Applications making requests to Kubernetes API server with expired tokens will fail. You can resolve the issue by updating your application and its dependencies to use newer versions of Kubernetes client SDK that automatically refreshes the tokens.

What is the problem?

Kubernetes version 1.21 graduated BoundServiceAccountTokenVolume feature [1] to beta and enabled it by default. This feature improves security of service account tokens by requiring a one hour expiry time, over the previous default of no expiration. This means that applications that do not refetch service account tokens periodically will receive an HTTP 401 unauthorized error response on requests to Kubernetes API server with expired tokens. You can learn more about the BoundServiceAccountToken feature in EKS Kubernetes 1.21 release notes [2].

To enable a smooth migration of applications to the newer time-bound service account tokens, EKS v1.21+ extends the lifetime of service account tokens to 90 days. Applications on EKS v1.21+ clusters that make API server requests with tokens that are older than 90 days will receive an HTTP 401 unauthorized  error response.

How can you resolve the issue?

To make the transition to time bound service account tokens easier, Kubernetes has updated the below official versions of client SDKs to automatically refetch tokens before the one hour expiration:

* Go v0.15.7 and later
* Python v12.0.0 and later
* Java v9.0.0 and later
* Javascript v0.10.3 and later
* Ruby master branch
* Haskell v0.3.0.0

We recommend that you update your application and its dependencies to use one of the above client SDK versions if you are on an older version.

While not an exhaustive list, the below AWS components have been updated to use the newer Kubernetes client SDKs that automatically refetches the token :

* Amazon VPC CNI: v1.8.0 and later
* CoreDNS: v1.8.4 and later
* AWS Load Balancer Controller: v2.0.0 and later
* kube-proxy: v1.21.2-eksbuild.2 and later 

[...]

PR #5383, opened 2022-08-16
Refresh K8s Service Account Token

josunect:
k8s service token is never refreshed, so this PR is adding some logic to re-read the token:

- No more than once every minute
- If the SA token file has been modified

The auth strategy "anonymous" uses this token also as the user auth token. As the user is anonymous, get the SA to inherit its permissions. 

TODO: Still need to verify if the token is properly refreshed in the cache to be used when other auth strategies are set. 

Fixes https://github.com/kiali/kiali/issues/5070

### DIFF
--- a/kubernetes/token.go
+++ b/kubernetes/token.go
@@ -13,7 +13,7 @@ import (
 const DefaultServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 
 var KialiToken string
-var LastRead time.Time
+var LastRead, timer time.Time
 
 func GetKialiToken() (string, error) {
 	expired, err := IsTokenExpired()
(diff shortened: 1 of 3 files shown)


kubernetes/token.go
```diff
@@ -1,25 +1,91 @@
...
+
+	"github.com/kiali/kiali/log"
+)
 
 // Be careful with how you use this token. This is the Kiali Service Account token, not the user token.
 // We need the Service Account token to access third-party in-cluster services (e.g. Grafana).
 
-const DefaultServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+var DefaultServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+var refreshTime = int64(60)
```
Xunzhuo, line 15: Need we to make this configurable, since the refreshTime is different in different platforms?

Xunzhuo (commented):
I know this is WIP, just a side comment.

kubernetes/token.go
josunect: Thanks for your comment, @Xunzhuo 
I'm reviewing this.

josunect:
I'm closing this PR as I created another different approach in https://github.com/kiali/kiali/pull/5386 
Setting expire at in the k8s configuration.

Issue #5384, opened 2022-08-16
Debug information autoscroll when copying from the text field


matejnesuta:
Found this one during the "Help" dropdown testing.

When I open the Help dropdown and choose the "View Debug Info", I am presented with a modal, which includes two text fields. It's entire content gets copied after a single click or with the "Copy" button. 

The issue is that it scrolls automatically all the way to the bottom when the content is copied either with the button, or with the click. It's confusing visually and if I wanted to read it's content, one accidental click could ruin it. 

Before copying:
![image](https://user-images.githubusercontent.com/45061792/184929337-d31c4c61-6a98-4eaa-b9b4-2655bf145a3c.png)

After copying:
![image](https://user-images.githubusercontent.com/45061792/184929529-4a26768b-35b9-468d-a748-5ed36c3f0efa.png)

The "Health config" behaves differently even though it looks exactly the same as the bottom field, which is a little confusing as well. 

Not sure, if bug or a feature.

jmazzitelli:
I have always hated that separate "health config" text box at the top. Why is "health config" special? It is already in the debug yaml. This just duplicates the output and makes this page more complicated and confusing than it needs to be.

I say we take this opportunity to collapse this dialog window so it only shows the full "Please include this information" text area and removes that health yaml which, as I understand, is nothing more than a subset of what you see in the full yaml text area.

jshaughn:
I agree about those two sections, I never really understood that.  Also, don't we use a nice YAML editor in other places, why not here?  Having said all that, this is pretty low priority.  I'll add to the backlog but also a "Good First issue", maybe we can get a contribution.

jmazzitelli:
> Also, don't we use a nice YAML editor in other places, why not here?

+1... that's a good idea.

jshaughn:
Closing as this page has now been redesigned.

PR #5385, opened 2022-08-16
[Cypress] "Help" dropdown test migration

matejnesuta:
Some tests from the old [Selenium](https://github.com/Kiali-QE/kiali-qe-python/blob/master/kiali_qe/tests/test_menu.py) suite revolved around the "Help" dropdown, now migrated into the Cypress framework 

It does everything the old suite did and a little more:
- Clicks on the "Help" button
- Checks if all the 4 options are present
- Checks if the "Documentation" link exists
- Opens modals for debug and certificates

Due to some inconsistencies mentioned in #5384 , I am not able to properly test the text fields in the "View Debug Info" modal. Right now, it only checks if the modal has opened.

### DIFF
--- a/frontend/cypress/integration/common/kiali_help.ts
+++ b/frontend/cypress/integration/common/kiali_help.ts
@@ -1,0 +1,26 @@
+import { When, Then, And } from "cypress-cucumber-preprocessor/steps";
+import { TableDefinition } from 'cypress-cucumber-preprocessor';
+
+
+When('user clicks on Help Button', () => {
+    cy.getBySel('about-help-button').click();
+});
+
+Then("user can see all of the Help dropdown options", (options: TableDefinition) => {
+  const names = options.raw()[0];
+  names.forEach(function (value) {
+    cy.get('li[role="menuitem"]').contains(value).should('be.visible');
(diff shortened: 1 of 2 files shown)


frontend/cypress/integration/common/kiali_help.ts
```diff
@@ -0,0 +1,26 @@
...
+And("the {string} button has a link", (title:string) => {
+  cy.get('li[role="menuitem"]').contains(title).should('have.attr', 'href');
+});
+
+When('user clicks on the {string} button', (title:string) => {
+  cy.get('li[role="menuitem"]').contains(title).click();
+});
+
+Then('user sees the {string} modal', (title:string) => {
+  cy.contains(title).should('be.visible')
```
ScriptingShrimp, line 25: You are searching `title` globally in whole DOM, please use more specific selector for better performance of test `cy.get('h1.pf-c-modal-box__title').contains(title).should('be.visible')`

ScriptingShrimp (changes requested):
please address requested changes

ScriptingShrimp (approved):
LGTM!

PR #5386, opened 2022-08-17
Refresh K8s Service Account Token (a)

josunect:
k8s service token is never refreshed, so this PR is adding some logic to re-read the token:

- Setting the expire token time  in token_expire_duration, in seconds, defaults to 1 minute: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md 

The auth strategy "anonymous" uses this token also as the user auth token. As the user is anonymous, get the SA to inherit its permissions.

- the cache uses a timer that runs every tokenExpireDuration because it needs to create a new client with the new token for each restClient. 
- GetKialiToken also uses a refresh timestamp when it is called outside the cache

Fixes https://github.com/kiali/kiali/issues/5070

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -705,7 +705,7 @@ func NewConfig() (c *Config) {
 			CacheIstioTypes:             []string{"AuthorizationPolicy", "DestinationRule", "EnvoyFilter", "Gateway", "PeerAuthentication", "RequestAuthentication", "ServiceEntry", "Sidecar", "VirtualService", "WorkloadEntry", "WorkloadGroup"},
 			CacheNamespaces:             []string{".*"},
 			CacheTokenNamespaceDuration: 10,
-			TokenExpireDuration:         90 * 86400, // Default to 90 days
+			TokenExpireDuration:         60, // Default to 60 seconds
 			ExcludeWorkloads:            []string{"CronJob", "DeploymentConfig", "Job", "ReplicationController"},
 			QPS:                         175,
 		},
--- a/config/config.go
+++ b/config/config.go
@@ -281,6 +281,9 @@ type KubernetesConfig struct {
 	// Kiali cache list of namespaces per user, this is typically short lived cache compared with the duration of the
 	// namespace cache defined by previous CacheDuration parameter
 	CacheTokenNamespaceDuration int `yaml:"cache_token_namespace_duration,omitempty"`
+	// Cache token duration (seconds)
+	// Refresh the SA token
+	TokenExpireDuration int `yaml:"token_expire_duration,omitempty"`
 	// List of controllers that won't be used for Workload calculation
 	// Kiali queries Deployment,ReplicaSet,ReplicationController,DeploymentConfig,StatefulSet,Job and CronJob controllers
 	// Deployment and ReplicaSet will be always queried, but ReplicationController,DeploymentConfig,StatefulSet,Job and CronJobs
@@ -702,6 +705,7 @@ func NewConfig() (c *Config) {
 			CacheIstioTypes:             []string{"AuthorizationPolicy", "DestinationRule", "EnvoyFilter", "Gateway", "PeerAuthentication", "RequestAuthentication", "ServiceEntry", "Sidecar", "VirtualService", "WorkloadEntry", "WorkloadGroup"},
 			CacheNamespaces:             []string{".*"},
 			CacheTokenNamespaceDuration: 10,
+			TokenExpireDuration:         90 * 86400, // Default to 30 days
 			ExcludeWorkloads:            []string{"CronJob", "DeploymentConfig", "Job", "ReplicationController"},
 			QPS:                         175,
 		},
--- a/config/config.go
+++ b/config/config.go
@@ -705,7 +705,7 @@ func NewConfig() (c *Config) {
 			CacheIstioTypes:             []string{"AuthorizationPolicy", "DestinationRule", "EnvoyFilter", "Gateway", "PeerAuthentication", "RequestAuthentication", "ServiceEntry", "Sidecar", "VirtualService", "WorkloadEntry", "WorkloadGroup"},
 			CacheNamespaces:             []string{".*"},
 			CacheTokenNamespaceDuration: 10,
-			TokenExpireDuration:         90 * 86400, // Default to 30 days
+			TokenExpireDuration:         90 * 86400, // Default to 90 days
 			ExcludeWorkloads:            []string{"CronJob", "DeploymentConfig", "Job", "ReplicationController"},
 			QPS:                         175,
 		},
--- a/config/config.go
+++ b/config/config.go
@@ -281,9 +281,6 @@ type KubernetesConfig struct {
 	// Kiali cache list of namespaces per user, this is typically short lived cache compared with the duration of the
 	// namespace cache defined by previous CacheDuration parameter
 	CacheTokenNamespaceDuration int `yaml:"cache_token_namespace_duration,omitempty"`
-	// Cache token duration (seconds)
-	// Refresh the SA token
-	TokenExpireDuration int `yaml:"token_expire_duration,omitempty"`
 	// List of controllers that won't be used for Workload calculation
 	// Kiali queries Deployment,ReplicaSet,ReplicationController,DeploymentConfig,StatefulSet,Job and CronJob controllers
 	// Deployment and ReplicaSet will be always queried, but ReplicationController,DeploymentConfig,StatefulSet,Job and CronJobs
@@ -705,7 +702,6 @@ func NewConfig() (c *Config) {
 			CacheIstioTypes:             []string{"AuthorizationPolicy", "DestinationRule", "EnvoyFilter", "Gateway", "PeerAuthentication", "RequestAuthentication", "ServiceEntry", "Sidecar", "VirtualService", "WorkloadEntry", "WorkloadGroup"},
 			CacheNamespaces:             []string{".*"},
 			CacheTokenNamespaceDuration: 10,
-			TokenExpireDuration:         60, // Default to 60 seconds
 			ExcludeWorkloads:            []string{"CronJob", "DeploymentConfig", "Job", "ReplicationController"},
 			QPS:                         175,
 		},
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -159,7 +159,7 @@ func NewKialiCache() (KialiCache, error) {
 			select {
 			case <-ticker.C:
 				if newToken, err := kubernetes.GetKialiToken(); err != nil {
-					log.Errorf("Error updating Kiali Token")
+					log.Errorf("Error updating Kiali Token %v", err)
 				} else {
 					log.Debug("Kiali Cache: Updating token")
 					istioConfig.BearerToken = newToken
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -163,7 +163,11 @@ func NewKialiCache() (KialiCache, error) {
 				} else {
 					log.Debug("Kiali Cache: Updating token")
 					istioConfig.BearerToken = newToken
-					istioClient, err = kubernetes.NewClientFromConfig(&istioConfig)
+					var errorInitClient error
+					istioClient, errorInitClient = kubernetes.NewClientFromConfig(&istioConfig)
+					if errorInitClient != nil {
+						log.Errorf("Error creating new Client From Config")
+					}
 					kialiCacheImpl.istioClient = *istioClient
 					kialiCacheImpl.k8sApi = istioClient.GetK8sApi()
 					kialiCacheImpl.istioApi = istioClient.Istio()
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -117,7 +117,6 @@ func NewKialiCache() (KialiCache, error) {
 
 	refreshDuration := time.Duration(kConfig.KubernetesConfig.CacheDuration) * time.Second
 	tokenNamespaceDuration := time.Duration(kConfig.KubernetesConfig.CacheTokenNamespaceDuration) * time.Second
-	tokenExpireDuration := time.Duration(kConfig.KubernetesConfig.TokenExpireDuration) * time.Second
 
 	cacheNamespaces := kConfig.KubernetesConfig.CacheNamespaces
 	cacheIstioTypes := make(map[string]bool)
@@ -153,7 +152,7 @@ func NewKialiCache() (KialiCache, error) {
 	kialiCacheImpl.istioApi = istioClient.Istio()
 
 	// Update SA Token
-	kialiCacheImpl.stopCacheChan = kialiCacheImpl.refreshCache(tokenExpireDuration, istioConfig)
+	kialiCacheImpl.stopCacheChan = kialiCacheImpl.refreshCache(istioConfig)
 	log.Infof("Kiali Cache is active for namespaces %v", cacheNamespaces)
 	return &kialiCacheImpl, nil
 }
@@ -242,8 +241,8 @@ func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
 }
 
 // RefreshNamespace will delete the specific namespace's cache and create a new one.
-func (c *kialiCacheImpl) refreshCache(tokenExpireDuration time.Duration, istioConfig rest.Config) chan bool {
-	ticker := time.NewTicker(tokenExpireDuration)
+func (c *kialiCacheImpl) refreshCache(istioConfig rest.Config) chan bool {
+	ticker := time.NewTicker(60 * time.Second)
 	quit := make(chan bool)
 	go func() {
 		for {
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -702,6 +705,7 @@ func NewConfig() (c *Config) {
 			CacheIstioTypes:             []string{"AuthorizationPolicy", "DestinationRule", "EnvoyFilter", "Gateway", "PeerAuthentication", "RequestAuthentication", "ServiceEntry", "Sidecar", "VirtualService", "WorkloadEntry", "WorkloadGroup"},
 			CacheNamespaces:             []string{".*"},
 			CacheTokenNamespaceDuration: 10,
+			TokenExpireDuration:         90 * 86400, // Default to 30 days
```
Xunzhuo, line 708: It is more like 90 days in default?

kubernetes/cache/cache.go
```diff
@@ -149,6 +151,34 @@ func NewKialiCache() (KialiCache, error) {
...
+			case <-ticker.C:
+				if newToken, err := kubernetes.GetKialiToken(); err != nil {
+					log.Errorf("Error updating Kiali Token")
+				} else {
+					log.Debug("Kiali Cache: Updating token")
+					istioConfig.BearerToken = newToken
+					var errorInitClient error
+					istioClient, errorInitClient = kubernetes.NewClientFromConfig(&istioConfig)
+					if errorInitClient != nil {
+						log.Errorf("Error creating new Client From Config")
```
Xunzhuo, line 169: How about also logging with error itself, since you got the return error? Then the error logs make more sense, it is better for us to find the reason why it failed.

config/config.go
josunect: That's correct, committed change

kubernetes/cache/cache.go
josunect: Good idea! Thanks

kubernetes/cache/cache.go
```diff
@@ -149,6 +151,34 @@ func NewKialiCache() (KialiCache, error) {
...
 
+	// Update SA Token
+	ticker := time.NewTicker(tokenExpireDuration)
+	quit := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				if newToken, err := kubernetes.GetKialiToken(); err != nil {
+					log.Errorf("Error updating Kiali Token")
```
Xunzhuo, line 162: Maybe we need to add error details here too:)

Xunzhuo (commented):
BTW, it looks like we added a new field in configuration, maybe we need to add related changes to kiali operator too?

kubernetes/cache/cache.go
josunect: Thanks, @Xunzhuo ! I missed this one

josunect:
> BTW, it looks like we added a new field in configuration, maybe we need to add related changes to kiali operator too?
> 
> Refer to PR like [kiali/kiali-operator#558](https://github.com/kiali/kiali-operator/pull/558)

I will review this change, thanks.

config/config.go
```diff
@@ -702,6 +705,7 @@ func NewConfig() (c *Config) {
 			CacheIstioTypes:             []string{"AuthorizationPolicy", "DestinationRule", "EnvoyFilter", "Gateway", "PeerAuthentication", "RequestAuthentication", "ServiceEntry", "Sidecar", "VirtualService", "WorkloadEntry", "WorkloadGroup"},
 			CacheNamespaces:             []string{".*"},
 			CacheTokenNamespaceDuration: 10,
+			TokenExpireDuration:         90 * 86400, // Default to 90 days
```
israel-hdez, line 708: A little note:

So, the token will be re-read after 90 days. In the {k8s changelog for v1.21](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.21.md) I see this:

```
The BoundServiceAccountTokenVolume feature has been promoted to beta, and enabled by default.
    Clients should reload the token from disk periodically (once per minute is recommended) to ensure they continue to use a valid token. k8s.io/client-go version v11.0.0+ and v0.15.0+ reload tokens automatically.
```

So, the recommendation is to re-read every minute.
I'm not wrong with keeping 90 days. I just want to be sure that you really want 90 days as the default value, because my thought is that 90days may be a default for EKS, while other flavors of k8s may have something different (although I may be wrong).

I think you also may need an additional PR on the [kiali/kiali-operator repo](https://github.com/kiali/kiali-operator/) to, and probably also the [kiali/helm-charts repo](https://github.com/kiali/helm-charts/) to document the new setting and let people change this value.

josunect: Thanks for the review, @israel-hdez ! I will have a look

josunect:
Operator PR: https://github.com/kiali/kiali-operator/pull/563

config/config.go
israel-hdez: New configs sometimes trigger a change in the `kiali-server` chart: https://github.com/kiali/helm-charts/tree/master/kiali-server

But most often, no changes are required. So, you may simply try the `kiali-server` chart and try to add your new config. If it gets configured correctly, then no changes are needed there. There is also the [`kiali-operator` chart](https://github.com/kiali/helm-charts/tree/master/kiali-operator) but that one very rarely needs a change.

kubernetes/cache/cache.go
```diff
@@ -149,6 +151,34 @@ func NewKialiCache() (KialiCache, error) {
...
+					if errorInitClient != nil {
+						log.Errorf("Error creating new Client From Config %v", errorInitClient)
+					}
+					kialiCacheImpl.istioClient = *istioClient
+					kialiCacheImpl.k8sApi = istioClient.GetK8sApi()
+					kialiCacheImpl.istioApi = istioClient.Istio()
+				}
+			case <-quit:
+				ticker.Stop()
+				return
```
israel-hdez, line 177: I'm wondering if this is dead code, as I don't understand where the `quit` channel is fed.

israel-hdez (commented):
Hmmm... I see the token is refreshing correctly. Despite that, once the token rotates, I start seeing a lot of `Unauthorized` messages in the logs:

image

Something is still holding the old token, but I cannot see where...

EDIT: Note that it's only the logs showing that something is wrong. Surprisingly, the UI works OK. So, I may think there is some memory leak.

israel-hdez (changes requested):
I'm "escalating" to "request changes" as I see that, although Kiali works ~~correctly~~ without emitting errors in the UI, and despite you are refreshing the token of the cache, there still seems to be issues with the cache.

Despite deleting pods and workloads on the cluster, the deleted resources are still listed in Kiali. So, I think those errors in the logs are because something is broken in the cache.

kubernetes/cache/cache.go
nrfox: Yes you want to close the `quit` channel in the `Stop` method. That's the only way this goroutine will exit.

kubernetes/token.go
```diff
@@ -1,16 +1,23 @@
...
 
 // Be careful with how you use this token. This is the Kiali Service Account token, not the user token.
 // We need the Service Account token to access third-party in-cluster services (e.g. Grafana).
 
-const DefaultServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+var DefaultServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 
 var KialiToken string
+var tokenRead time.Time
+var tokenExpireDuration time.Duration
```
nrfox, line 17: Do we need to add more global variables?

kubernetes/token.go
```diff
@@ -20,6 +27,32 @@ func GetKialiToken() (string, error) {
...
+
+// Set ServiceAccountPath
+func SetDefaultServiceAccountPath(path string) {
+	DefaultServiceAccountPath = path
+}
+
+// Set Token Expiration
+func SetTokenExpireDuration(expireTime time.Duration) {
+	tokenExpireDuration = expireTime
+}
```
nrfox, line 58: Are these being used anywhere?

kubernetes/token_test.go
```diff
@@ -0,0 +1,59 @@
...
+	data := []byte(content)
+	err := os.WriteFile(name, data, 0644)
+	return err
+}
+
+// Aux func to remove file
+func removeFile(name string) error {
+	err := os.Remove(name)
+	return err
+}
```
nrfox, line 59: Generally it's neater to pass in `t *testing.T` and fail instead of returning an error.
```
func removeFile(name string, t *testing.T) {
	if err := os.Remove(name); err != nil {
	  t.Fatal(err)
	}
}
```

You can also call [t.Helper()](https://pkg.go.dev/testing#T.Helper) at the beginning of helper functions to make the test failure output nicer.

kubernetes/cache/cache.go
```diff
@@ -149,6 +151,34 @@ func NewKialiCache() (KialiCache, error) {
...
+					log.Debug("Kiali Cache: Updating token")
+					istioConfig.BearerToken = newToken
+					var errorInitClient error
+					istioClient, errorInitClient = kubernetes.NewClientFromConfig(&istioConfig)
+					if errorInitClient != nil {
+						log.Errorf("Error creating new Client From Config %v", errorInitClient)
+					}
+					kialiCacheImpl.istioClient = *istioClient
+					kialiCacheImpl.k8sApi = istioClient.GetK8sApi()
+					kialiCacheImpl.istioApi = istioClient.Istio()
```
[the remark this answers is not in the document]
nrfox: We shouldn't try to recreate the cache every 60 seconds. Can we read the token every 60 seconds to check for updates but only expire if there's been some change?

josunect:
Thanks for the last review, @nrfox and @israel-hdez 
Very useful comments! 
I'm working on the changes

kubernetes/token.go
israel-hdez: My eyes see they are being used in the tests.

kubernetes/cache/cache.go
josunect: Added the change in the Stop method.

kubernetes/cache/cache.go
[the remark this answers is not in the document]
josunect: - I still need to test the first point. 
- Great idea. I just refresh the cache in case the token has changed.

kubernetes/token.go
josunect: tokenRead would be needed to keep track of when the token was read in the next time. 
Probably tokenExpireDuration doesn't need to be global. Just to avoid read kConfig again.

kubernetes/token_test.go
josunect: Updated.

kubernetes/token.go
josunect: Yes, that is correct

nrfox: Ah well can you set those directly in the tests instead of adding methods to set them since these setters are only used by the tests?
```
DefaultServiceAccountPath = tmpFileTokenExpired
...
```

kubernetes/cache/cache.go
```diff
@@ -82,6 +82,8 @@ type (
 	}
 )
 
+var stopCacheChan chan bool
```
nrfox, line 85: This shouldn't be a global var. It's state that should be scoped to the `kialiCacheImpl` since nothing outside of that needs to use it. Can you make this a field on the struct?

kubernetes/cache/cache.go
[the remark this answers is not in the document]
josunect: I was able to reproduce the error with the steps provided by @israel-hdez, thanks! 

As @nrfox it seem it was an issue with namespace cached. After the last change, we can see the logs: 

```
022-08-31T15:28:17Z INF Kiali Cache: Updating cache with new token
2022-08-31T15:28:17Z INF Adding a RegistryRefreshHandler for [default]
2022-08-31T15:28:17Z INF Waiting for Kiali cache for [namespace: default] to sync
2022-08-31T15:28:17Z INF Kiali cache for [namespace: default] stopped
2022-08-31T15:28:17Z INF Kiali cache for [namespace: default] started
2022-08-31T15:28:17Z INF Adding a RegistryRefreshHandler for [kiali]
2022-08-31T15:28:17Z INF Waiting for Kiali cache for [namespace: kiali] to sync
2022-08-31T15:28:17Z INF Kiali cache for [namespace: kiali] stopped
2022-08-31T15:28:18Z INF Kiali cache for [namespace: kiali] started
```

Instead of the errors

kubernetes/token.go
josunect: Updated.

kubernetes/cache/cache.go
josunect: Updated.

kubernetes/token.go
```diff
@@ -1,16 +1,23 @@
...
-const DefaultServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
+var DefaultServiceAccountPath = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 
 var KialiToken string
+var tokenRead time.Time
+var tokenExpireDuration time.Duration
 
 func GetKialiToken() (string, error) {
-	if KialiToken == "" {
+	if KialiToken == "" || IsTokenExpired() {
```
nrfox, line 20: It'd be nice to refresh before the token expires by watching for when the token changes and updating `KialiToken` then. We'd need a separate watch thread that continuously reads the token from the service account path and diff the existing token with the token it reads. The cache (or other places we use KialiToken) can then watch for when KialiToken changes. That may require some refactoring though since KialiToken is a global var and it doesn't seem like a good idea to startup a background goroutine simply by importing this package. That can be an optimization we address later. Could you add a `TODO:` here to refresh the token when it changes rather than after it expires?

kubernetes/token.go
```diff
@@ -20,6 +27,22 @@ func GetKialiToken() (string, error) {
...
+		kConfig := kialiConfig.Get()
+		tokenExpireDuration = time.Duration(kConfig.KubernetesConfig.TokenExpireDuration) * time.Second
+	}
+
+	if time.Since(tokenRead) > tokenExpireDuration {
+		return true
+	} else {
+		return false
+	}
+}
```
nrfox, line 48: nit: `IsTokenExpired` is a bit of a misnomer since this is just checking how long it's been since the token was last read directly from disk rather than returning the cached token. This doesn't need to be exported by this package and I'd even say you don't need to add another config option to adjust this interval. The config option here currently adjusts how often we read the token directly from disk vs. returning the cached option and users shouldn't need to adjust that beyond the default 60 seconds. You can set `var tokenExpireDuration = time.Second * 60` and remove the config option. This also saves the trouble of having to make operator/helm changes.

kubernetes/cache/cache.go
```diff
@@ -236,6 +241,47 @@ func (c *kialiCacheImpl) RefreshNamespace(namespace string) {
...
+				if newToken, err := kubernetes.GetKialiToken(); err != nil {
+					log.Errorf("Error updating Kiali Token %v", err)
+				} else {
+					if istioConfig.BearerToken != newToken {
+						log.Info("Kiali Cache: Updating cache with new token")
+						istioConfig.BearerToken = newToken
+
+						istioClient, errorInitClient := kubernetes.NewClientFromConfig(&istioConfig)
+						if errorInitClient != nil {
+							log.Errorf("Error creating new Client From Config %v", errorInitClient)
```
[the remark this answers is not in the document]
josunect: But, this call, NewClientFromConfig, is using the BearerToken in the Config to create the k8SClient:

```
client := K8SClient{
		token: config.BearerToken,
	}
```

So, if the token is not updated, the K8SClient will be created with the old config. 

But, If this call fails, we can restore the old token in the istioConfig.

kubernetes/token.go
josunect: Added TODO comment

kubernetes/token.go
josunect: I have renamed the function name. 

Note this config option is also used in cache.go to indicate the timer periodicity. 
Do we want to change to a fixed value? The read token periodicity is a recommendation, but not sure if that could change? 
I think it could be more flexible as a configuration option. But if no required, the issue https://github.com/kiali/kiali-operator/pull/563 can be closed.

kubernetes/cache/cache.go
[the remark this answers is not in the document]
nrfox: Ah you're right I didn't notice that. Looks good :+1:

kubernetes/token.go
nrfox: Yep that'd be good

kubernetes/token.go
nrfox: I don't see the need for users to configure it but if we're going to keep the config option I'd suggest renaming it. `TokenExpireDuration` to me makes it sound like we will expire the token in 60 seconds.

josunect: I have finally removed the config item. 
I think it would make sense if this is an expiration time, but less if this is a timer.

nrfox (approved):

israel-hdez (approved):
So, after having some headache with the operator, you no longer need changes on that side :-)

Nick has done a good review. Anyway, code is fine to me.
I tried your changes and all looks fine. Errors are gone!

josunect:
> So, after having some headache with the operator, you no longer need changes on that side :-)

I know! :) But it was a good learning, at least :D
> 
> Nick has done a good review. Anyway, code is fine to me. I tried your changes and all looks fine. Errors are gone!

Thanks for the review!

PR #5387, opened 2022-08-17
[Cypress] Navbar toggle test

matejnesuta:
Another migrated test from the old [Selenium](https://github.com/Kiali-QE/kiali-qe-python/blob/6acc35c8e2f38bf9b7daa5db5ee14114e867772a/kiali_qe/tests/test_menu.py#L47) suite.

Does nothing less than toggling the sidebar and verifying if it is or is not visible.

### DIFF
--- a/frontend/cypress/integration/common/kiali_sidebar_toggle.ts
+++ b/frontend/cypress/integration/common/kiali_sidebar_toggle.ts
@@ -1,0 +1,26 @@
+import {And, When, Then} from "cypress-cucumber-preprocessor/steps";
+
+When('the sidebar is open', () => {
+  cy.get('#page-sidebar').should('be.visible');
+});
+
+And('user presses the navigation toggle button', () => {
+  cy.get('#nav-toggle').click()
+});
+
+Then('user cannot see the sidebar', () => {
+  cy.get('#page-sidebar').should('not.be.visible');
+});
+
+When('the sidebar is closed', () => {
+  cy.get('#page-sidebar').should('exist').then(($sidebar) => {
+    if($sidebar.attr('aria-hidden') === 'false') {
+      cy.get('#nav-toggle').click()
+    }
+  })
+  cy.get('#page-sidebar').should('not.be.visible');
+});
+
+Then('user sees the sidebar', () => {
+  cy.get('#page-sidebar').should('be.visible');
+});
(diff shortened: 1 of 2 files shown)


frontend/cypress/integration/common/kiali_sidebar_toggle.ts
```diff
@@ -0,0 +1,31 @@
...
+
+function sidebarVisibility(){
+  cy.get('#page-sidebar').should('be.visible');
+  return;
+};
+
+function buttonClick(){
+  cy.get('#nav-toggle').click()
+  return;
+};
```
ScriptingShrimp, line 11: can we abstract this somewhere else, WDYT @nrfox ?

frontend/cypress/integration/featureFiles/kiali_sidebar_toggle.feature
```diff
@@ -0,0 +1,18 @@
...
+  @sidebar-toggle @smoke
+  Scenario: Default visibility of sidebar
+    Then user sees the sidebar 
+
+  @sidebar-toggle @smoke
+  Scenario: Tooggle visibility of the sidebar
+    When user presses the navigation toggle button
+    And user cannot see the sidebar
+    And user presses the navigation toggle button
+    Then user sees the sidebar
```
nrfox, line 18: Since these are two different actions I'd split this up into two different scenarios and then I think it'll read better too. Something like:
```
Given the sidebar is closed
When user presses the navigation toggle button
Then user sees the sidebar
```
and
```
Given the sidebar is open
When user presses the navigation toggle button
Then user cannot see the sidebar
```
That'll also help ensure these tests won't fail if some other test left the sidebar open or closed.

matejnesuta: I don't think this is a big issue. There is a setup defined on the top of the Gherkin and it basically "logs in" and loads the page all over again. I refactored the code, hope everything will be fine this time.

nrfox:
@matejnesuta the cypress test is failing in CI. You just need a conditional in the `When the sidebar is closed/open` step definition that looks at some open/close attr on either the button or the navbar. Something like:
```
When('the sidebar is closed', () => {
  # Conditional needs to be inside the `then` block because of how cypress queues and executes commands async.
  cy.get('#page-sidebar').should('exist').then(($sidebar) => {
    if($sidebar.attr('aria-hidden') === 'false') {
      cy.get('#nav-toggle').click()
    }
  })
  ...
});
```

nrfox:
CI failure is unrelated to these changes. Looks like just a slow API response caused the failure.

PR #5389, opened 2022-08-18
Show time controls in a modal from mini-graph card

israel-hdez:
**This only affects the kiosk=parent mode**

The _refresh interval_ and _duration_ controls are currently not available in the kiosk mode, because toolbars are not visible.

When Kiali is running under in kiosk mode, the title of the mini-graph card (i.e. the time range) will become clickable. When clicking, a modal dialog will appear offering the missing time controls.

Since this is done in the mini-graph card component, these changes are reflected in all detail pages (apps, workloads, services).

Related to kiali/openshift-servicemesh-plugin#42

### DIFF
--- a/frontend/src/components/CytoscapeGraph/MiniGraphCard.tsx
+++ b/frontend/src/components/CytoscapeGraph/MiniGraphCard.tsx
@@ -7,7 +7,8 @@ import {
   CardTitle,
   Dropdown,
   DropdownItem,
-  KebabToggle
+  KebabToggle,
+  Button
 } from '@patternfly/react-core';
 import history from '../../app/History';
 import GraphDataSource from '../../services/GraphDataSource';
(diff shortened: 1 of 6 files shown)


lucasponce:
Thanks, @israel-hdez; I'm going to test it.

One comment; changes that are sent to the kiali repo are considered generic to the "kiosk" mode is whatever mode a user can use (the "simple" default one or the "advanced" with the communication enabled between parent/iframe.

In other words, let's avoid adding "OpenShift Plugin" or any other reference in the main kiali repo. Otherwise, that may bring confusion to other users. Just use the "kiosk" mode instead.

(I see comments needing to be changed in the commit message and PR).

lucasponce:
I have a first comment about the style: in the kiosk mode is introducing some padding that breaks the line alignment:
image
vs 
image

Another comment about style, if you look to the other "icons" all of them are aligned to the "right," so it looks "weird" to align the "time" icon to the left in this context.

Also, the font sizes:
image

I would keep that long text in "font 14px".

The modal is ok from a style point of view; I understand the tradeoff, and probably this location may change if the "parent" of the iframe provides any helper for this scenario.

I think the other tabs may show just the icon (aligned to the right, similar to the graph offers) to invoke the modal similarly.

But that can be performed in a different PR if it makes this work longer.

israel-hdez:
@lucasponce Thanks for the review!
I pushed changes to fix the font and paddings. It should be good now.

About the other tabs, yes I'm planning to open new PRs to add the dialog on them.

israel-hdez:
> let's avoid adding "OpenShift Plugin

OK, I changed the main description :-)

lucasponce (approved):
From the kiosk=parent:
image

Modal:
image

From standalone:
image

israel-hdez:
Thanks for the review, @lucasponce.
Merging!

Issue #4635, opened 2021-12-29
Support single cluster traces view when using Jaeger with multi-cluster storage backend


yeya24:
# Use Case
We have over 80+ edge clusters and each cluster deploys an Istio and Kiali. At the beginning, we deploy one Jaeger instance with Badger backend at each cluster so that Kiali can query Jaeger directly and get the single cluster trace view.

Now we plan to switch our Jaeger storage backend to a central Elasticsearch deployment, which means all the traces from 80+ clusters will go into ES now. However, in this case, Kiali won't work properly because now it will get traces from all the clusters that matched the requested service.

To solve this, right now we have to deploy another jaeger instance for single-cluster-level query only. This is not sufficient.

## Possible solution
Right now, all the traces contain a tag to indicate the cluster name, for example `cluster.name=us-east`. I hope Kiali can support this to do some traces filtering for such multi-cluster deployment.

## Alternative solution 
https://github.com/jaegertracing/jaeger/pull/3412 is going on the Jaeger side so once that is ready this issue can be solved as well.

lucasponce:
The "query.proto" API allows to filter via "tags" so this parameter could be added to the query used to fetch the data:

![image](https://user-images.githubusercontent.com/1662329/148036800-fb79c6f0-8698-4f5b-bd60-23d4d9ac2b31.png)

One aspect to sync is that the "cluster.name" used in Jaeger matches with the "cluster ID" used in the mesh as this one is relevant in other Kiali areas.

jshaughn:
@lucasponce , We recently added `queryScope` to the prometheus configuration, to handle unified metrics.  It sounds like we need the same thing as part of the Jaeger/Tracing config.  See https://github.com/kiali/kiali/issues/4664.

jshaughn:
@yeya24 Is this still a problem for you?  It sounds like we could do something similar to the `queryScope` support we have for querying a unified Prometheus, and leverage the query.proto API mentioned above.

yeya24:
@jshaughn I haven't followed the same topic for a while now. And now Jaeger seems to support storage tenancy already. I need to double check.

jshaughn:
@yeya24 Thanks for the reply.  If you find out any more information please let me know if this is still a problem for your Kiali deployment model.

jshaughn:
I think it makes sense to add the same support here as we did for Prometheus queries.  If traces are unified it seems this could be a problem, and the solution seems straightforward if we support adding static tags to each query.

PR #5395, opened 2022-08-19
For Tracing queries, add QueryScope support (analogous to Prometheus)

jshaughn:
Closes https://github.com/kiali/kiali/issues/4635

**Operator PR: ** https://github.com/kiali/kiali-operator/pull/558

As an example, By adding this to the Kiali CR:

```
spec:
  external_services:
    tracing:
      query_scope:
        test: test-value
```

I then get this in the Jaeger query:

```
2022-08-19T15:04:38Z INF Prepared Jaeger query: query:{service_name:"x-server.alpha" tags:{key:"test" value:"test-value"} start_time_min:{seconds:1660921234} start_time_max:{seconds:1660921478 nanos:862887247} duration_min:{} search_depth:15}
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -187,14 +187,15 @@ type GrafanaVariablesConfig struct {
 
 // TracingConfig describes configuration used for tracing links
 type TracingConfig struct {
-	Auth                 Auth     `yaml:"auth"`
-	Enabled              bool     `yaml:"enabled"` // Enable Jaeger in Kiali
-	InClusterURL         string   `yaml:"in_cluster_url"`
-	IsCore               bool     `yaml:"is_core,omitempty"`
-	NamespaceSelector    bool     `yaml:"namespace_selector"`
-	URL                  string   `yaml:"url"`
-	UseGRPC              bool     `yaml:"use_grpc"`
-	WhiteListIstioSystem []string `yaml:"whitelist_istio_system"`
+	Auth                 Auth              `yaml:"auth"`
+	Enabled              bool              `yaml:"enabled"` // Enable Jaeger in Kiali
+	InClusterURL         string            `yaml:"in_cluster_url"`
+	IsCore               bool              `yaml:"is_core,omitempty"`
+	NamespaceSelector    bool              `yaml:"namespace_selector"`
+	QueryScope           map[string]string `yaml:"query_scope,omitempty"`
+	URL                  string            `yaml:"url"`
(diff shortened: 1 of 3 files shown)


Xunzhuo (commented):
Looks like we already passed tags from request Query, why do we need another configuration fields to get query tags?

jshaughn:
> Looks like we already passed tags from request Query, why do we need another configuration fields to get query tags?

My understanding is that we support tags in queries to the jaeger API, but that the tags we supply are fixed (in particular we look for the 'error' tag).  I'm not aware of a way to provide ad-hoc tag key-value pairs, or to configure a set of tags in the kiali CR.  That is why I thought we could extend to Jaeger the same "queryScope" approach we use for narrowing Prom queries.  The PR uses the existing support for tags.

I may have misinterpreted something, and so I'd like to hear from @lucasponce.

config/config.go
```diff
@@ -187,14 +187,15 @@ type GrafanaVariablesConfig struct {
...
-	NamespaceSelector    bool     `yaml:"namespace_selector"`
-	URL                  string   `yaml:"url"`
-	UseGRPC              bool     `yaml:"use_grpc"`
-	WhiteListIstioSystem []string `yaml:"whitelist_istio_system"`
+	Auth                 Auth              `yaml:"auth"`
+	Enabled              bool              `yaml:"enabled"` // Enable Jaeger in Kiali
+	InClusterURL         string            `yaml:"in_cluster_url"`
+	IsCore               bool              `yaml:"is_core,omitempty"`
+	NamespaceSelector    bool              `yaml:"namespace_selector"`
+	QueryScope           map[string]string `yaml:"query_scope,omitempty"`
```
lucasponce, line 195: "QueryScope" sounds strange to me in this context.
I think it can create confusion, but something like "AdditionalTags" would be clear for maintaining this in the future.

lucasponce:
The propagation of the additional tags into the jaeger query looks correct.

config/config.go
jshaughn: I understand what you are saying but I like the synergy between the naming for both this and the Prometheus external service.  The use case being solved is the same.  If we ever had to, we could also add `AdditionalTags` and it would simply do the same thing.  But for the user trying to solve the problem of unified storage, I think it's good to have the same naming here and for prom.

jshaughn:
@Xunzhuo Does the explanation above answer your doubt?

PR #5400, opened 2022-08-22
Support Telemetry and WasmPlugin Istio objects

### LINKED ISSUE #5274: Support Telemetry and WasmPlugin Istio objects
Kiali supports Istio configuration, but two new resources are not yet added:

```
  {
    group: 'extensions.istio.io',
    version: 'v1alpha1',
    kind: 'WasmPlugin',
  },
  {
    group: 'telemetry.istio.io',
    version: 'v1alpha1',
    kind: 'Telemetry',
  },
```

This issue would add them to the regular Istio Config visualization, but the basic editing features.

Kiali Validations are not part of the scope of this task.

josunect:
New Istio Types support, excluding validations at this stage. 

image

Telemetry example: 

image

wasm plugin:

image

The types have been tested with the following yaml in crc: 

oc apply -f telemetry.yaml

```
kind: Telemetry
metadata:
  name: mesh-default
  namespace: bookinfo
spec:
  tracing:
  - randomSamplingPercentage: 10.00
    customTags:
      my_new_foo_tag:
        literal:
          value: "foo"
```

oc apply -f wasmplugin.yaml

```
apiVersion: extensions.istio.io/v1alpha1
kind: WasmPlugin
metadata:
  name: openid-connect
  namespace: bookinfo
spec:
  selector:
    matchLabels:
      istio: ingressgateway
  url: oci://private-registry:5000/openid-connect/openid:latest
  imagePullPolicy: IfNotPresent
  imagePullSecret: private-registry-pull-secret
  phase: AUTHN
  pluginConfig:
    openid_server: authn
    openid_realm: ingress
```

Fixes #5274

### DIFF
--- a/business/checkers/telemetries_checker.go
+++ b/business/checkers/telemetries_checker.go
@@ -1,0 +1,23 @@
+package checkers
+
+import (
+	"github.com/kiali/kiali/models"
+	"istio.io/client-go/pkg/apis/telemetry/v1alpha1"
+)
+
+const TelemetryCheckerType = "telemetry"
+
+type TelemetryChecker struct {
+	Namespaces  models.Namespaces
+	Telemetries []*v1alpha1.Telemetry
+}
+
+// An Object Checker runs all checkers for an specific object type (i.e.: pod, route rule,...)
+// It run two kinds of checkers:
+// 1. Individual checks: validating individual objects.
+// 2. Group checks: validating behaviour between configurations.
+func (in TelemetryChecker) Check() models.IstioValidations {
+	validations := models.IstioValidations{}
+
+	return validations
+}
--- a/business/checkers/telemetries_checker.go
+++ b/business/checkers/telemetries_checker.go
@@ -1,8 +1,9 @@
 package checkers
 
 import (
-	"github.com/kiali/kiali/models"
 	"istio.io/client-go/pkg/apis/telemetry/v1alpha1"
+
+	"github.com/kiali/kiali/models"
 )
 
 const TelemetryCheckerType = "telemetry"
--- a/business/checkers/wasm_plugin_checker.go
+++ b/business/checkers/wasm_plugin_checker.go
@@ -1,0 +1,23 @@
+package checkers
+
+import (
+	"github.com/kiali/kiali/models"
+	extentions_v1alpha1 "istio.io/client-go/pkg/apis/extensions/v1alpha1"
+)
+
+const WasmPluginCheckerType = "wasmplugin"
+
+type WasmPluginChecker struct {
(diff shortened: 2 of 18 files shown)


josunect:
Tested in OpenShift Service Mesh plugin:

image
Screenshot from 2022-08-23 12-20-45

lucasponce:
Note that the implementation of the Istio Config in the OSSM plugin is native :)
So, these two new objects were updated here:
https://github.com/kiali/openshift-servicemesh-plugin/blob/main/plugin/src/k8s/resources.ts#L70

But thanks for testing them!

josunect:
> Note that the implementation of the Istio Config in the OSSM plugin is native :) So, these two new objects were updated here: https://github.com/kiali/openshift-servicemesh-plugin/blob/main/plugin/src/k8s/resources.ts#L70
> 
> But thanks for testing them!

That's right! Thanks, @lucasponce

josunect:
So the error happens in the following call when using the hack script run-kiali with -kc current: 
`
{"error":"wasmplugins.extensions.istio.io is forbidden: User \"system:serviceaccount:istio-system:kiali-service-account\" cannot list resource \"wasmplugins\" in API group \"extensions.istio.io\" in the namespace \"bookinfo\""}`

In the following call: 

http://localhost:8000/api/namespaces/bookinfo/istio?validate=true

jmazzitelli:
> cannot list resource \"wasmplugins\" in API group \"extensions.istio.io\"

I'm pretty sure we do NOT have that permission in our code. We need to grant that in the operator and the server templates.

For the things we need to do when adding new permissions, see https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission-that-applies-to-resources-in-all-accessible-namespaces

josunect:
> > cannot list resource "wasmplugins" in API group "extensions.istio.io"
> 
> I'm pretty sure we do NOT have that permission in our code. We need to grant that in the operator and the server templates.
> 
> For the things we need to do when adding new permissions, see https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-role-permission-that-applies-to-resources-in-all-accessible-namespaces

Thanks! I was testing the role and role-viewer changes, but I was missing other checks. Very useful!

josunect:
Operator changes: https://github.com/kiali/kiali-operator/pull/561

hhovsepy:
Screenshot from 2022-08-25 15-55-01
Screenshot from 2022-08-25 15-54-48
Screenshot from 2022-08-25 15-54-37

hhovsepy (approved):

PR #5402, opened 2022-08-22
[make] fix pull secret

jmazzitelli:
OpenShift 4.11 no longer provides a secret with SAs. We need to fix our dev tools to work with 4.11 (but don't break when on 4.10 and before).

### DIFF
--- a/make/Makefile.cluster.mk
+++ b/make/Makefile.cluster.mk
@@ -5,7 +5,7 @@
 # If using OpenShift, you must expose the image registry externally.
 #
 
-.prepare-ocp-image-registry: .ensure-oc-exists
+.prepare-ocp-image-registry: .ensure-oc-login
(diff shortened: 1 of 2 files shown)


jmazzitelli:
The make targets no longer require the secret to exist in the SA. Instead, the make targets will get the token for the logged in user and use that.

jmazzitelli:
To test this is very easy - just `oc login` to an OpenShift 4.11 cluster, and push and deploy kiali. If it works, you know this PR is good.

Here's the make command to execute after you `oc login` to the cluster:

`make cluster-status build build-ui cluster-push operator-create kiali-create`

If you see a Kiali server running after you give everything time to install and come up, things are good.

nrfox (approved):
Tested on 4.11 :+1:

jmazzitelli:
some docs just as fyi: https://access.redhat.com/solutions/6964710

PR #5403, opened 2022-08-23
[make] clean up after no longer using default SA secret

jmazzitelli:
Now that OpenShift 4.11 (and related k8s versions) no longer have secrets in SAs by default,
we no longer rely on them. This cleans up some stuff a bit due to that.


jmazzitelli:
if you can `make cluster-push operator-create` and it works, this PR is good

leandroberetta (approved):
LGTM.

jmazzitelli:
i confirm this works on 4.11

jmazzitelli:
oh, there is one thing I wanted to do with this. The `.remove-operator-pull-secret` make target needs to be implemented. I'll do that here.

Issue #5258, opened 2022-06-28
[CI] Molecule tests no longer work with kube 1.24+


nrfox:
Some molecule tests rely on the default service account token secret to be present when dealing with token auth. As of kube 1.24+, these secrets are no longer created by default and have to be created manually.

See: https://github.com/kiali/kiali/runs/7086748946?check_suite_focus=true for an example of the failing molecule tests.

jmazzitelli:
While fixing this, we should look into fixing a couple hack scripts, too, which also rely on secrets being in SAs:

* https://github.com/kiali/kiali/blob/v1.52.0/hack/run-kiali.sh#L566
* https://github.com/kiali/kiali/blob/v1.52.0/hack/istio/multicluster/install-everything.sh#L59

jmazzitelli:
I did this for the make targets so they no longer rely on the SA secret - but I don't think its solution is relevant. I am just documenting it here as an fyi: https://github.com/kiali/kiali/pull/5402

jmazzitelli:
docs that may help: https://access.redhat.com/solutions/6964710

jmazzitelli:
For this issue, I'm not going to do the https://github.com/kiali/kiali/blob/v1.52.0/hack/istio/multicluster/install-everything.sh#L59 script - we'll save that for another day if someone wants it. AFAIK, no one is using that multicluster scripting anyway.

jmazzitelli:
PRs for this issue:
* https://github.com/kiali/kiali/pull/5405
* https://github.com/kiali/kiali-operator/pull/560

PR #5405, opened 2022-08-23
[hack] run-kiali.sh no longer expects default SA secret for obtaining tokens/certs

jmazzitelli:
This is to support k8s versions that do not by default provide SA secrets.

part of https://github.com/kiali/kiali/issues/5258

To test, just run `hack/run-kiali.sh` and if it works, this PR is good. Do NOT use `-kc current` option - that will not test this change. You must run this without that option.


josunect (approved):
Tested and works as expected

PR #5408, opened 2022-08-25
[hack] Add option to specify number of namespaces to perf cluster

nrfox:
Adds options to specify the number of namespaces and traffic rate for demo workloads when creating a perf cluster.

Relates to #5144

### DIFF
--- a/hack/perf-ibmcloud-openshift.sh
+++ b/hack/perf-ibmcloud-openshift.sh
@@ -32,22 +32,24 @@ DEFAULT_OPENSHIFT_VERSION="4.10_openshift"
 DEFAULT_REGION="us-south"
 DEFAULT_WORKLOADS="50"
 DEFAULT_WORKLOAD_NAMESPACES="1"
+DEFAULT_WORKLOAD_REQUEST_RATE="100"
 
 while [[ $# -gt 0 ]]; do
   key="$1"
   case $key in
     # OPTIONS
-    -akf|--api-key-file)       API_KEY_FILE="$2";        shift;shift ;;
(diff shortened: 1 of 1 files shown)


hack/perf-ibmcloud-openshift.sh
```diff
@@ -314,6 +333,13 @@ install_demo_apps() {
 }
 
 install_workloads() {
+  SCC_SERVICE_ACCOUNTS=""
+  for (( c=0; c<WORKLOAD_NAMESPACES; c++ ))
+  do
+    SCC_SERVICE_ACCOUNTS="${SCC_SERVICE_ACCOUNTS}- \"system:serviceaccount:depth-$c:default\"
+"
```
jmazzitelli, line 340: this looks weird. Is this syntax what you meant? (quote all by its lonesome on a separate line?)

jmazzitelli, line 340: I think you want to put the quote on the line above, no? That seems to do the same thing and won't look weird.
```suggestion
    SCC_SERVICE_ACCOUNTS="${SCC_SERVICE_ACCOUNTS}- \"system:serviceaccount:depth-$c:default\""
```

jmazzitelli (changes requested):
move quote to end of line. other than that, looks ok

hack/perf-ibmcloud-openshift.sh
jmazzitelli: oh, jsut saw this.

Add a comment in here then, to explain. otherwise, people (me) will see it and get a twitch in their eye..it just  looks wrong.

jmazzitelli (approved):

PR #5411, opened 2022-08-25
[make] Force the operator image to always be pulled when debugging via OLM

jmazzitelli:
Without this fix, you'll never get the latest image to be used
even after doing a make cluster-push-operator. And it will drive
you crazy wondering why your changes aren't getting picked up.


jshaughn (approved):
Make it so.

PR #5412, opened 2022-08-25
Time controls in details pages for Kiosk mode

israel-hdez:
Time controls in metrics dashboards, traces, workload pod logs, and traffic tabs.

image


israel-hdez:
Hey @lucasponce 

This is still a draft. I would like some feedback.
I've added the time controls modal to the metrics dashboards, traces and workload pod logs. A button is added in the tabs to launch the modal dialog.

I'm not sure I'm liking this approach, because most tabs have a toolbar that looks suitable to directly put the drop-downs instead of a button launching the dialog. Yet, I'm not sure if a consistent experience is what is wanted.

So, if possible, please share your feedback. It would be great if also @andrew-ronaldson can comment.

lucasponce:
@israel-hdez I think this "pattern" is already implemented in the Graph:
image

Where I can see some icons aligned top-right. I think in this case, as a user, I'd expect something similar:
image

or
image

Your approach is too intrusive with the current design; I think using the "top-right" alignment in those tabs would help.

jshaughn:
> I'm not sure I'm liking this approach, because most tabs have a toolbar that looks suitable to directly put the drop-downs instead of a button launching the dialog. Yet, I'm not sure if a consistent experience is what is wanted.

I agree that the actual dropdowns could fit in various places, and it is convenient to see the current settings onscreen.  But because they don't fit everywhere (like the minigraph) I think a consistent experience is best.  Looking at the screenshot, I agree with Lucas that the left-hand-side button is a little much.  I would suggest that we stick with Andrew's suggestion of the clock icon, and position it consistently to the far right side (left of a kebab, if there is a kebab).  It should probably show a tooltip with current settings, and open the modal on click.

Another option that we could pass by @andrew-ronaldson would combine your button idea with the icon idea. The clock icon could be combined with a short form of the current settings as the button text.  Something like:

image

israel-hdez:
BTW, I did is some "typing cleanup". That's what probably introduced most changes in this PR.

andrew-ronaldson:
Screen Shot 2022-08-30 at 3 38 43 PM
Made these two options @israel-hdez @jshaughn

israel-hdez:
I applied the suggestions.

This is how it looks without the border:

image

and with the border:

image

To me it is slightly better with the border, although without the border looks more neat. So, no strong preference, really.

israel-hdez:
After thinking about it a little, I'll pick the non-bordered option.

jshaughn (changes requested):
I like the cleaner look using KioskElement and KialiDispatch.  The new button looks good.  The modal looks good but has one problem. Because it is centered you need a tall window to see all of the time period menu items.  There is no scrolling in the modal, so if it's not tall enough the bottom choices are clipped.  Not sure how you want to resolve that issue but I think it should be fixed.

Also, I think in the UX meeting the decision was to, in the minigraph, move the clock icon out of the time-range chip and relocate it left of the kebab.

The time button on the Logs tab looks like it may be a little too far to the right.  Not a blocker, but take a look if you like.

israel-hdez:
@jshaughn I solved the issue of the dropdowns by moving the modal to the top of the page. I see this as a workaround for now and given this is mainly for the plugin, the modal should still be perceived so-so in the middle. The right fix (when needed) is to replace the `Dropdown` with a `Select` which has a `maxHeight` prop and shows scrollbars when the maxHeight is reached. But I think that change should go on its own PR.

About the Workload Pod Logs, I added a small margin at the right..

jshaughn (approved):
Everything looks good!

PR #5413, opened 2022-08-26
Moving lucasponce to Emeritus contributors

lucasponce:
It's been a pleasure working with the Kiali and Istio teams. 
They keep working on awesome stuff that will help shape how big architectures will work in the cloud and how users can have a rich experience in managing it.

I will focus on other projects and will not be able to follow or contribute to Kiali/Istio in the short/mid-term, so I think it makes sense to update the Governance.

Best,
Lucas

cc @kiali/maintainers


Xunzhuo (approved):
Thanks @lucasponce! It's a great pleasure to work with you and very appreciated for your helps to guide me into Kiali Projects. 

Best wishes, Xunzhuo.

Issue #5390, opened 2022-08-19
[Cypress] kiali_qe.tests.test_graph_page.test_layout selenium migration


ScriptingShrimp:
Please migrate this test from selenium. @matejnesuta 

Verify all elements in `/console/graph` are displayed. For reference please see screenshot 

![Screenshot from 2022-08-19 14-44-53](https://user-images.githubusercontent.com/86114326/185621636-485390d0-a955-4ca7-8d11-7e1f97629d61.png)

ScriptingShrimp:
Closing issue in favor of https://github.com/kiali/kiali/pull/5415

PR #5415, opened 2022-08-26
[Cypress] Migration from Selenium (graph buttons and legend sidebar test)

matejnesuta:
Another migrated [suite from Selenium](https://github.com/Kiali-QE/kiali-qe-python/blob/75f60eed7cb290e7e3c94f71cf870d1445154468/kiali_qe/tests/test_graph_page.py#L124).

The old suite most likely switched between the 4 available graph layouts. The new suite tests all of the buttons, as well as button groups (for example you should not be able to have active more than one Graph Layout at once). It also briefly checks the opening and closing of the Legend sidebar. 

The tests mess only with the buttons, not with the graph itself. We might do that sometimes in a future, however it is not a priority now.

Related issue: #5390

### DIFF
--- a/frontend/cypress/integration/common/graph_toolbar_legend.ts
+++ b/frontend/cypress/integration/common/graph_toolbar_legend.ts
@@ -1,0 +1,69 @@
+import { Then, And, When } from "cypress-cucumber-preprocessor/steps";
+
+function buttonClick(label:string){
+    cy.get(`button[aria-label="${label}"]`).click();
+};
+
+function buttonPrepare(label:string, active:boolean){
+    cy.waitForReact();
+    cy.getReact("Button",{ props: { 'aria-label': `${label}`}})
+    .should('have.length', '1')
+    .getProps()
+    .then((props => {
+        if (props.isActive !== active){
+            buttonClick(label);
+        }
+    }));
+};
+
+function buttonState(label:string, active:boolean){
+    cy.waitForReact();
+    cy.getReact("Button",{ props: { 'aria-label': `${label}`}})
+    .should('have.length', '1')
+    .getProps('isActive')
+    .should('eq', active);
+}
+
+Then("the toggle button {string} is enabled", (label: string) => {
+    cy.get(`button[aria-label="${label}"]`).should('have.attr', 'aria-disabled', 'false');
+});
+
+And("the toggle button {string} is enabled", (label: string) => {
+    cy.get(`button[aria-label="${label}"]`).should('have.attr', 'aria-disabled', 'false');
+});
+
+And("the button {string} is clicked", (label: string) => {
+    buttonClick(label);
+});
+
+Then("the button {string} is active", (label: string) => {
+   buttonState(label, true); 
+});
+
+Then("the button {string} is not active", (label: string) => {
+   buttonState(label, false); 
+});
+
+And("the {string} is turned on", (label: string) => {
+   buttonPrepare(label, true); 
+});
+
+And("the {string} is turned off", (label: string) => {
+   buttonPrepare(label, false); 
+});
+
+Then("user can see the legend section", () => {
+    cy.get("[data-test='graph-legend']").should("be.visible");
+});
+
+When("the Legend section is visible", () => {
+    buttonClick("Show Legend");
+});
+
+And ("the cross is clicked", () => {
+    cy.get("#legend_close").click()
+});
+
+Then("user cannot see the legend section", () => {
+    cy.get("[data-test='graph-legend']").should("not.exist");
+});
--- a/frontend/cypress/integration/featureFiles/graph_toolbar_legend.feature
+++ b/frontend/cypress/integration/featureFiles/graph_toolbar_legend.feature
@@ -1,0 +1,143 @@
+Feature: Kiali Graph page - Graph toolbar and legend sidebar 
+
+  User opens the Graph page and manipulates the "alpha", "beta" namespace graph with buttons
+    located at the bottom of the page 
+
+  Background:
+    Given user is at administrator perspective
+    And user graphs "alpha,beta" namespaces
+
+@graph-page-manipulation
+Scenario Outline: Check if the <label> button is usable
+  Then the toggle button "<label>" is enabled
+  Examples:
+  | label |
+  | Toggle Drag |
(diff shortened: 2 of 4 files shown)


frontend/cypress/integration/featureFiles/graph_manipulation.feature
```diff
@@ -0,0 +1,140 @@
...
+Scenario: Check that Hide Healthy Edges and Hide All Edges cannot be active at the same time
+  When the "Hide Healthy Edges" is turned off 
+  And the "Hide All Edges" is turned off 
+  And the button "Hide Healthy Edges" is clicked
+  And the button "Hide All Edges" is clicked
+  Then the button "Hide Healthy Edges" is not active
+  And the button "Hide All Edges" is active
+
+@graph-page-manipulation
+Scenario: Check that Hide Healthy Edges and Hide All Edges cannot be active at the same time #2
```
nrfox, line 68: nit: instead of `#2` can you add something more descriptive like to distinguish it from the first test?

frontend/cypress/integration/featureFiles/graph_manipulation.feature
```diff
@@ -0,0 +1,140 @@
...
+Scenario: Check that Namespace Layout Style 1 and Namespace Layout Style 2 cannot be active at the same time
+  When the "Namespace Layout Style 1" is turned off 
+  And the "Namespace Layout Style 2" is turned off 
+  And the button "Namespace Layout Style 1" is clicked
+  And the button "Namespace Layout Style 2" is clicked  
+  Then the button "Namespace Layout Style 1" is not active
+  And the button "Namespace Layout Style 2" is active
+
+@graph-page-manipulation
+Scenario: Check that Namespace Layout Style 1 and Namespace Layout Style 2 cannot be active at the same time #2
```
nrfox, line 86: Same here, something more descriptive than `#2`

frontend/cypress/integration/featureFiles/graph_manipulation.feature
```diff
@@ -0,0 +1,140 @@
...
+  And the button "Graph Layout Style 1" is clicked
+  And the button "Graph Layout Style 2" is clicked
+  And the button "Graph Layout Style 3" is clicked
+  Then the button "Graph Layout Default Style" is not active
+  And the button "Graph Layout Style 1" is not active
+  And the button "Graph Layout Style 2" is not active
+  And the button "Graph Layout Style 3" is active
+
+@graph-page-manipulation
+Scenario: Check that only one graph layout style can be active at the same time in the opposite order
```
nrfox, line 109: > Same here, something more descriptive than #2

Like this here!

> in the opposite order

frontend/cypress/integration/featureFiles/graph_manipulation.feature
```diff
@@ -0,0 +1,140 @@
...
+  And the toggle button "Zoom to Fit" is enabled
+  And the toggle button "Hide Healthy Edges" is enabled
+  And the toggle button "Hide All Edges" is enabled
+  And the toggle button "Graph Layout Default Style" is enabled
+  And the toggle button "Graph Layout Style 1" is enabled
+  And the toggle button "Graph Layout Style 2" is enabled
+  And the toggle button "Graph Layout Style 3" is enabled
+  And the toggle button "Namespace Layout Style 1" is enabled
+  And the toggle button "Namespace Layout Style 2" is enabled
+  And the toggle button "Show Legend" is enabled
```
nrfox, line 22: This seems like it would be a good table test similar to the `Scenario Outline` below.

frontend/cypress/integration/featureFiles/graph_manipulation.feature
```diff
@@ -0,0 +1,140 @@
+Feature: Kiali Graph page - Graph manipulation
```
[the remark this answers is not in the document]
nrfox: Can you rename the feature file name as well along with the accompanying `.ts` file?

frontend/cypress/integration/common/graph_manipulation.ts
```diff
@@ -0,0 +1,69 @@
...
+And("the {string} is turned on", (label: string) => {
+   buttonPrepare(label, true); 
+});
+
+And("the {string} is turned off", (label: string) => {
+   buttonPrepare(label, false); 
+});
+
+Then("user can see the legend section", () => {
+    cy.get("span").contains("Legend").should("be.visible");
```
nrfox, line 56: It'd be better to use something more specific than looking at all the span elements for the text "Legend". If there's multiple legends on the screen this could fail later or be flaky. This is a good case for adding a `data-test` attribute to the legend and searching for that. There's a `GraphLegend` component you can add it to: https://github.com/kiali/kiali/blob/master/frontend/src/pages/Graph/GraphLegend.tsx#L44.

nrfox (commented):
A few minor things otherwise looks good

matejnesuta:
Hopefully everything is fixed by now. If not, I'll fix it on Monday.

frontend/cypress/integration/common/graph_manipulation.ts
matejnesuta: I tried it and I am not sure if I did it the right way, because it broke one of the CI pipelines.

nrfox: It looks like just a snapshot test that needs to be updated. When you run the ui unit tests locally you should see an option for updating those.

ScriptingShrimp:
note to myself: disable this testsuite in selenium

ScriptingShrimp:
@nrfox all tests fail for me locally, no prometheus available in my local instance of CRC
image

wondering how this passed in CI? :thinking:

nrfox:
> @nrfox all tests fail for me locally, no prometheus available in my local instance of CRC

@ScriptingShrimp it sounds more like there's a problem with your environment. I'll try testing this against mine.

nrfox:
> @ScriptingShrimp it sounds more like there's a problem with your environment. I'll try testing this against mine.

@ScriptingShrimp tests all pass in my environment as well. The tests are good to go from my perspective.

nrfox (approved):

ScriptingShrimp (approved):
LGTM :+1:

PR #5422, opened 2022-08-30
Use rw locks and remove globals

### LINKED ISSUE #4849: unexpected ci error in TestClientExpiry

nrfox:
Replaces read lock with read/write locks and moves some global variables into the factory. I'm not totally sure what was causing the race condition but I suspect something to do with either the read locks themselves or when the lock/unlocking was happening.

I ran this locally for ~50 iterations with `make test-race` and encountered 0 failures.

Fixes #4849

### DIFF
--- a/kubernetes/client_factory.go
+++ b/kubernetes/client_factory.go
@@ -15,12 +15,6 @@ import (
 
 var factory *clientFactory
 
-// mutex for when modifying the stored clients
-var mutex = &sync.RWMutex{}
-
-// when a client is expired, a signal with its tokenHash will be sent to recycleChan
-var recycleChan = make(chan string, 10)
-
 // defaultExpirationTime set the default expired time of a client
 const defaultExpirationTime = time.Minute * 15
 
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
This looks good.  It could be useful to maybe enhance the test-only functions to mention that they are test-hooks and establish their own locks (so don't nest inside outer locking functions), but not a blocker.  Not sure the RW lock is the source of the fix, may very well be the cleaner, defer-based unlocking.  Either way, it's working and it was a real problem.

Xunzhuo (approved):
Looks good, nice catch! Thanks @nrfox

nrfox:
@jshaughn I added some comments to the helper functions.

jshaughn (approved):
re-approving after trivial changes.  the test failure is a known issue that is being fixed independently.

Issue #5393, opened 2022-08-19
[Cypress] Ensure all validations in Istio Config have references


ScriptingShrimp:
@matejnesuta
In /console/istio verify that all values in drop downs are:

- clickable (able to add)
- removable (cross in each card)
- clear all is working

for more details see screenshot:

![image](https://user-images.githubusercontent.com/86114326/185626538-4e23ef8f-8f5d-473b-a907-80d4ccce3672.png)

Issue #5392, opened 2022-08-19
[Cypress] Ensure all validations in Istio Type have references


ScriptingShrimp:
@matejnesuta  
In `/console/istio` verify that all values in drop downs are:
- clickable (able to add)
- removable (cross in each card)
- clear all is working

for more details see screenshot:
![image](https://user-images.githubusercontent.com/86114326/185625861-97c30059-646c-44e1-8eb9-6abd8821ead2.png)

PR #5424, opened 2022-08-31
[Cypress] Istio Config filters

matejnesuta:
Another migration from the Selenium. These suites are supposed to check whether are all of the filters usable or not.

Most of the implementation is in the istio_config_type_filters.ts, because some of the steps in both gherkins are identical. 

This PR is related to #5392 and #5393.

### DIFF
--- a/frontend/cypress/integration/featureFiles/istio_config_type_filters.feature
+++ b/frontend/cypress/integration/featureFiles/istio_config_type_filters.feature
@@ -1,0 +1,95 @@
+@istio-page
+Feature: Kiali Istio Config page
+
+  On the Istio Config page, an admin should be able to choose all of the filters present. 
+  The admin should also be able close a single filter either with it's cross, or with the Close All feature.
+
+  Background:
+    Given user is at administrator perspective
+    And user is at the "istio" page
+    And user selects the "istio-system" namespace
+    And user filters by "Istio Type" 
+    And no filters are active
+
+    Scenario Outline: All of the filters should be available for turning on
+    When user chooses a "<category>"
+    Then user can see only the "<category>"  
+    Examples:
+    | category |
+    | AuthorizationPolicy |
+    | DestinationRule |
+    | EnvoyFilter |
+    | Gateway |
+    | PeerAuthentication |
+    | RequestAuthentication |
+    | ServiceEntry |
+    | Sidecar |
+    | VirtualService |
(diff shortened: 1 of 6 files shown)


matejnesuta:
Looks like the frontend CI failed, but not in the suites I added, so I will keep here for now without changes.

mattmahoneyrh:
@matejnesuta Nice work! Good to see the test passing in the CI test job. I will leave the PR approval to others who know the code much better then me.

frontend/cypress/integration/featureFiles/istio_config_type_filters.feature
```diff
@@ -0,0 +1,95 @@
...
+  The admin should also be able close a single filter either with it's cross, or with the Close All feature.
+
+  Background:
+    Given user is at administrator perspective
+    And user is at the "istio" page
+    And user selects the "istio-system" namespace
+    And user filters by "Istio Type" 
+    And no filters are active
+
+    Scenario Outline: All of the filters should be available for turning on
```
jshaughn, line 14: This test works but I feel the run time is maybe not worth the test.  In essence it tests that the menu has all of the expected options, but it does it one at a time, relogging in, reselecting the Istio Config menu, and then selecting the filter.  I feel like all of the the options could be tested at the same time.   We are not really trying to test patternfly menus, that's for PF to do, we're trying to test Kiali.  As such, I think we would be better just making sure that all of the expected options are there and then activate a filter (or two) check that the filtering actually works, and that deactivation works.

The same comment applies for the tests below that are repeating the same thing for every menu option.  This test is the second longest running cypress feature file, and the two new feature files combined take over 3 minutes, which is over 10% of the total runtime.

  │ ✔  featureFiles/istio_config_type_filt      02:13       39       39        -        -        - │
  │    ers.feature                                                                                 │
  
  │ ✔  featureFiles/istio_config_validatio      00:55       17       17        -        -        - │
  │    n_filters.feature

jshaughn (changes requested):
Although working I think these tests need to be revised with a better balance of what the test does with how long it takes to run.

frontend/cypress/integration/featureFiles/istio_config_type_filters.feature
matejnesuta: Decided to rework this entirely. Closing the PR for now. I am going to make a new one once it's done.

PR #5427, opened 2022-09-01
Fix cypress graph toggle flake

nrfox:
Updates the assertion in the test to `be.visible`. Using `exists` would cause the `then` statement to fail because the element can exist in the dom but not be visible yet.

Example of flake: https://github.com/kiali/kiali/runs/8140114181?check_suite_focus=true


nrfox:
@matejnesuta I can't assign you directly but tagging you as well for a review.

jshaughn (approved):
yup. be.visible is the way go in general

PR #5433, opened 2022-09-03
Fix refresh interval in Kiosk mode

israel-hdez:
The Refresh.tsx component (and redux) was being used for refresh logic. Since this is the component that is added in pages that need refresh data periodically (and thus, need the duration/time-interval and refresh interval controls), it made sense that this component also fired the refreshes. But in Kiosk mode this component is not rendered, thus refreshes are not being fired.

To fix the refresh for Kiosk mode, the timer has to be removed from the Refresh.tsx component, and also Redux is no longer involved for triggering refreshes (although Redux is still involved for storing/updating the desired interval). A useRefreshInterval custom react-hook is created to host the timer and this is the hook that function components should use to watch/subscribe for refreshes. For class components, the new RefreshNotifier.tsx component exposes similar functionality.

The refresh timer will be active while there are active "subscriptions" and the auto-refresh is not "paused". Since Redux is no longer involved in refreshes, a triggerRefresh function is also created to programmatically trigger a refresh (i.e. for refresh buttons).

Since several components were relying on Redux to watch/listen when a refresh is triggered, a connectRefresh.tsx HOC is also created to provide a similar interface, and mainly to reduce the churn of the changes required for the fix. However, it is probably better to use the useRefreshInterval hook or the RefreshNotifier.tsx component because they make explicit to "clients" when a refresh is happening, rather than trying to determine the refresh event in the componentDidUpdate method.

Related kiali/openshift-servicemesh-plugin#42

### DIFF
--- a/frontend/src/components/Refresh/RefreshNotifier.tsx
+++ b/frontend/src/components/Refresh/RefreshNotifier.tsx
@@ -6,6 +6,12 @@ interface Props {
   onTick: (timestamp: TimeInMilliseconds) => void;
 }
 
+// RefreshNotifier won't render any visual element. Its work is limited to calling
+// the onTick function passed in the props. The onTick function will be called
+// each time a global refresh event is emitted (i.e. when the user specified
+// refresh interval has elapsed). Class components wanting to watch for these
+// global refresh events should use this <RefreshNotifier onTick={yourCallBack}> component.
+// For function components, use the useRefreshInterval() hook.
 export default function RefreshNotifier({ onTick }: Props) {
   const refreshing = useRefreshInterval();
 
--- a/frontend/src/components/Refresh/RefreshNotifier.tsx
+++ b/frontend/src/components/Refresh/RefreshNotifier.tsx
@@ -1,0 +1,24 @@
+import { useEffect } from "react";
+import { TimeInMilliseconds } from "../../types/Common";
+import useRefreshInterval from "../../hooks/refresh";
+
+interface Props {
+  onTick: (timestamp: TimeInMilliseconds) => void;
+}
+
+export default function RefreshNotifier({ onTick }: Props) {
+  const refreshing = useRefreshInterval();
+
+  useEffect(() => {
+    if (refreshing.previousRefreshAt !== refreshing.lastRefreshAt) {
+      // We only want to notify when a refresh happens. At mount, previousRefreshAt == lastRefreshAt.
+      // So, we notify only when both values are different.
+      onTick(refreshing.lastRefreshAt);
+
+      // NOTE: This won't handle well the case when props.onTick changes. If that happens,
+      // this will immediately call props.onTick, even if a refresh hasn't been fired.
+    }
+  }, [onTick, refreshing.previousRefreshAt, refreshing.lastRefreshAt]);
+
+  return null;
+}
--- a/frontend/src/hooks/refresh.ts
+++ b/frontend/src/hooks/refresh.ts
@@ -1,0 +1,61 @@
+import { useSelector } from "react-redux";
+import { KialiAppState } from "../store/Store";
+import { IntervalInMilliseconds, TimeInMilliseconds } from "../types/Common";
+import { useEffect, useState } from "react";
+
+let numSubscribers = 0;
+let intervalId: null | number = null;
+
+function doTick(time?: TimeInMilliseconds) {
+  const refreshTick = new CustomEvent('refreshTick', {detail: time ?? Date.now()});
+  document.dispatchEvent(refreshTick);
+}
+
+export function triggerRefresh(time?: TimeInMilliseconds) {
+  doTick(time);
+}
+
+export default function useRefreshInterval() {
+  const refreshInterval = useSelector<KialiAppState, IntervalInMilliseconds>((state) => state.userSettings.refreshInterval);
+  const [lastRefreshAt, setLastRefreshAt] = useState<TimeInMilliseconds>(Date.now());
+  const [previousRefreshAt, setPreviousRefreshAt] = useState<TimeInMilliseconds>(lastRefreshAt);
+
+  useEffect(() => {
+    function handleTick(e: CustomEventInit<TimeInMilliseconds>) {
+      setPreviousRefreshAt(lastRefreshAt);
+      setLastRefreshAt(e.detail!);
+    }
+
+    // Subscribe
+    document.addEventListener('refreshTick', handleTick);
+    numSubscribers++;
+
+    return function () {
+      // Unsubscribe;
+      document.removeEventListener('refreshTick', handleTick);
+      numSubscribers--;
+    };
+  }, [lastRefreshAt]);
+
+  useEffect(() => {
+    if (intervalId !== null) {
+      // When mounting, reset the timer.
+      // Also, if refreshInterval changed, set a new timer.
+      window.clearInterval(intervalId);
+      intervalId = null;
+    }
+
+    if (numSubscribers !== 0 && refreshInterval !== 0) {
+      intervalId = window.setInterval(triggerRefresh, refreshInterval);
+    }
+
+    return function () {
+      if (intervalId !== null && numSubscribers === 0) {
+        window.clearInterval(intervalId);
+        intervalId = null;
+      }
+    }
+  }, [refreshInterval]);
+
+  return { lastRefreshAt, previousRefreshAt, refreshInterval };
+}
--- a/frontend/src/actions/GlobalActions.ts
+++ b/frontend/src/actions/GlobalActions.ts
@@ -1,7 +1,7 @@
 // Action Creators allow us to create typesafe utilities for dispatching actions
 import { ActionType, createAction, createStandardAction } from 'typesafe-actions';
 import { ActionKeys } from './ActionKeys';
-import { KioskMode, TimeInMilliseconds  } from '../types/Common';
+import { KioskMode } from '../types/Common';
 
 export const GlobalActions = {
   unknown: createAction('KIALI_UNKNOWN'), // helper for testing
@@ -9,7 +9,6 @@ export const GlobalActions = {
   decrementLoadingCounter: createAction(ActionKeys.DECREMENT_LOADING_COUNTER),
   setPageVisibilityHidden: createAction(ActionKeys.SET_PAGE_VISIBILITY_HIDDEN),
   setPageVisibilityVisible: createAction(ActionKeys.SET_PAGE_VISIBILITY_VISIBLE),
-  setLastRefreshAt: createStandardAction(ActionKeys.SET_LAST_REFRESH)<TimeInMilliseconds>(),
   setKiosk: createStandardAction(ActionKeys.SET_KIOSK)<KioskMode>(),
 };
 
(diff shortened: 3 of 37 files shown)


frontend/src/hooks/refresh.ts
```diff
@@ -0,0 +1,61 @@
...
+let intervalId: null | number = null;
+
+function doTick(time?: TimeInMilliseconds) {
+  const refreshTick = new CustomEvent('refreshTick', {detail: time ?? Date.now()});
+  document.dispatchEvent(refreshTick);
+}
+
+export function triggerRefresh(time?: TimeInMilliseconds) {
+  doTick(time);
+}
```
nrfox, line 16: Do we need `triggerRefresh` if we also have the `connectRefresh` HOC? Could we pass `setLastRefreshAt` through there?

frontend/src/hooks/refresh.ts
```diff
@@ -0,0 +1,61 @@
...
+    // Subscribe
+    document.addEventListener('refreshTick', handleTick);
+    numSubscribers++;
+
+    return function () {
+      // Unsubscribe;
+      document.removeEventListener('refreshTick', handleTick);
+      numSubscribers--;
+    };
+  }, [lastRefreshAt]);
```
nrfox, line 38: I'm confused as to why we're adding dom listeners here and tracking the number of subscribers?

Is the refresh interval shared across all components or does each component have its own refresh interval?

nrfox (commented):
The behavior of the button changed a bit. Now when you click the refresh button it stays "selected" whereas before the button would disable until the refresh completed.

Before:
buttonfrommain

After (nevermind the flashing. something wrong with my screencap software):
stickybutton

wrt the hooks implementation I'm happy to see us adopting hooks more but I wonder if trying to keep the two state management techniques makes things less clear. If we're going to use hooks for state management for a component, then components that rely on that state should be converted to functions so they can use the hook directly. As it is here I'm confused here as to what is managing the state since redux, hooks, and direct dom listeners are all involved with setting the refresh interval.

israel-hdez:
> The behavior of the button changed a bit. Now when you click the refresh button it stays "selected" whereas before the button would disable until the refresh completed.

I've been comparing with `master` and looks like it's only the graph page that shows that behavior. All other pages won't disable the controls. So, after looking closely, to my eye behavior is the same before and after.

frontend/src/components/Refresh/RefreshNotifier.tsx
```diff
@@ -0,0 +1,24 @@
+import { useEffect } from "react";
+import { TimeInMilliseconds } from "../../types/Common";
+import useRefreshInterval from "../../hooks/refresh";
+
+interface Props {
+  onTick: (timestamp: TimeInMilliseconds) => void;
+}
+
+export default function RefreshNotifier({ onTick }: Props) {
```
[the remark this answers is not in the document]
israel-hdez: It will call the `onTick` function that is passed in the arguments when a refresh is triggered. It is its only goal :-)

As you commented in your review comment, this `RefreshNotifier` is intended to be used in class components, just to prevent a refactor to a function component. It won't render nothing, but it still invokes that `onTick` callback.

frontend/src/hooks/refresh.ts
israel-hdez: Well.... setLastRefreshAt is no longer in Redux. So, in that `connectRefresh` we can pass a `setLastRefreshAt` prop, but it would be only a reference to `triggerRefresh`.

I'm hoping some day we would remove that `connectRefresh` HOC. So, I think we still would need this.

frontend/src/hooks/refresh.ts
israel-hdez: > Is the refresh interval shared across all components or does each component have its own refresh interval?

Yes, it is shared.

Previously, the `Refresh.tsx` component owned the `interval` and propagated the refresh through Redux which other components "listened" passively. Across Kiali pages you may notice that there is always a single set of duration-refreshInterval-refreshButton components, which hopefully gives clarity that the refresh interval is shared. If that `Refresh.tsx` component is not rendered, auto-refresh does not happen and this is why the plugin is not refreshing data right now (because in Kiosk mode, toolbars aren't rendered, thus `Refresh.tsx` is also not rendered).

I wanted to keep the shared nature of the refresh interval. So, this refresh.ts hook is what I'm trying to use for it. Without redux in play to propagate refresh events (i.e. without that `setLastRefreshAt` dispatch) I couldn't think a way to listen/fire the refresh event other than using DOM listeners. Also, since this hook can be invoked by several components on a single page, I thought it was worth to track the number of subscribers to know when to start/stop the interval - somehow I wanted to have a way to stop the interval when needed; i.e. something similar to not having the interval if the `Refresh.tsx` is not rendered - it is still not the same, but perhaps somewhat close. Sadly, I found that a few components in the masthead are watching for refreshes, meaning that once you login, auto-refresh remains active all the time :-/

israel-hdez:
> @nrfox As it is here I'm confused here as to what is managing the state since redux, hooks, and direct dom listeners are all involved with setting the refresh interval.

About the dom listeners, I tried to hide that with the custom `useRefreshInterval` hook. So, I hope we can see that as implementation detail.

About redux vs hooks vs (something else?) I think both approaches are and will be maintained. AFAIK, no one is recommended over the other. So, I'm not sure that we really should chose a single approach. The way I see it is that it is still react or redux state and you choose how to access that state (function->hooks vs class-->attributes).

The only reason I've been using functions (and thus hooks) is because I like its simplicity for small components. However, as the component grows, sometimes that simplicity goes away. Also, interestingly, custom hooks make it more straightforward to re-use code (at least for me).

frontend/src/hooks/refresh.ts
nrfox: So we want to share the same state across components (the refresh interval) and have multiple components subscribe to that state but we can't use Redux because the plugin does not support redux and using hooks would give each component its own copy of the state rather than having a single shared copy? Is that right? Why does the plugin not use redux? Is this a factor of using iframes and kiosk mode?

I'm not sure this is any better but another option is to set the refresh interval in the URL and have the components watch for changes to the url. iirc some components are already doing something similar.

I'm fine with using the dom listeners for now but it feels like what we want to use (and were using) was Redux for sharing state across components. Is that it an inherent limitation of the plugin that we can't use Redux or just the way it's currently implemented with iframes and kiosk mode?

nrfox:
> The only reason I've been using functions (and thus hooks) is because I like its simplicity for small components. However, as the component grows, sometimes that simplicity goes away. Also, interestingly, custom hooks make it more straightforward to re-use code (at least for me).

That seems like exactly what hooks are for and good at and I'm also in favor of using them more especially for managing local state and reusing code. You can also use redux with hooks. Trying to use hooks inside of classes ends up making things less clear to me but I realize that converting everything that uses the refresh logic to a functional component would be a larger refactor.

nrfox:
> I've been comparing with `master` and looks like it's only the graph page that shows that behavior. All other pages won't disable the controls. So, after looking closely, to my eye behavior is the same before and after.

Whoops I just assumed that both pages would be the same. Sorry for the noise.

frontend/src/components/Refresh/RefreshNotifier.tsx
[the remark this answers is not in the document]
nrfox: With this, do we need the `connectRefresh` HOC at all then? Aren't they serving the same purpose?

You mind adding some comments here explaining this component's purpose?

frontend/src/hooks/refresh.ts
israel-hdez: > we can't use Redux because the plugin does not support redux and using hooks would give each component its own copy of the state rather than having a single shared copy? Is that right? Why does the plugin not use redux? Is this a factor of using iframes and kiosk mode?

At the moment I can only give answers that represent just my opinion, because currently I think everything it is somewhat vague. I cannot say that we cannot use Redux. Indeed, we _can_ because it is just an `iframe` and we can do whatever we want there. However, I do think we _may need_ to reduce Redux usage if we eventually fall on the need of re-using components between Kiali and the plugin (either by creating a new repo of shared components, or by doing refactoring within the existent repos). The Redux store is what is giving high coupling across components in the front-end. I do think we can still use Redux in a native OS plugin, but we may need to find ways to reduce the coupling and for this I think we may need to "pull up" redux state to have just a few top level components connected to Redux and, then, pass state down using standard props rather than having everything connected to Redux.

That said, I don't think this "subscription" model is what motivated the usage of DOM listeners. (read the following)

> I'm not sure this is any better but another option is to set the refresh interval in the URL and have the components watch for changes to the url. iirc some components are already doing something similar.

The refresh interval is indeed already available through the URL. Indeed, this number is still in Redux. However, when I refer to the `interval` I'm talking about the javascript timer that is setup by invoking the `setInterval` function. I don't think that can go in the URL. I couldn't think on another way of managing this timer, but suggestions are very welcome as I also don't like this approach that much.

frontend/src/components/Refresh/RefreshNotifier.tsx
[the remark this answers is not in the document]
israel-hdez: All three - the `useRefreshInterval` hook, the `RefreshNotifier` and the `connectRefresh` HOC - serve the same purpose. The hook and the `RefreshNotifier` would have been enough. I added the HOC just to avoid a major refactor in the class components, which have large logic in the `componentDidUpdate` to find when to refresh data.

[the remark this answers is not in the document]
israel-hdez: > You mind adding some comments here explaining this component's purpose?

I'll do that :-)

frontend/src/hooks/refresh.ts
nrfox: Thanks for answering my questions. We don't need to figure all this out now in this PR I'm just trying to get more clarity on how the different state management techniques are supposed to work together.

> I do think we can still use Redux in a native OS plugin, but we may need to find ways to reduce the coupling and for this I think we may need to "pull up" redux state to have just a few top level components connected to Redux and, then, pass state down using standard props rather than having everything connected to Redux.

Yep!

> The refresh interval is indeed already available through the URL. Indeed, this number is still managed in Redux. However, when I refer to the interval I'm talking about the javascript timer that is setup by invoking the setInterval function. I don't think that can go in the URL. I couldn't think on another way of managing this timer, but suggestions are very welcome as I also don't like this approach that much.

Managing the interval could be part of the redux action I think. The timer would just start ticking when the store gets initialized. The current approach is fine for now I'd say and is working as expected. When we go to refactor more things related to state management in the future perhaps we can revisit then.

frontend/src/components/Refresh/RefreshNotifier.tsx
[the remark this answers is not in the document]
jshaughn: Having read through the other comments as well, please be generous with in-code comments for this PR, so that future devs can have a head start understanding why things are like they will be after this is merged.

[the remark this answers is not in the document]
israel-hdez: I added a comment. Not sure if that will be enough. You tell :-)

nrfox (approved):
Tested both Kiali and the plugin and everything works as expected for both

PR #5434, opened 2022-09-05
[Cypress] Mismatched node type graph fix testing

matejnesuta:
This is a little test case related [to this issue fix](https://github.com/kiali/kiali-ui/pull/2328) from 1.48.  

It toggles the node graph on the services/<detail>/ page and then checks that the graph type dropdown on the top of the page is disabled.

### DIFF
--- a/frontend/cypress/integration/common/service_details.ts
+++ b/frontend/cypress/integration/common/service_details.ts
@@ -1,4 +1,4 @@
-import { Then } from '@badeball/cypress-cucumber-preprocessor';
+import { Then, And } from '@badeball/cypress-cucumber-preprocessor';
 
 function openTab(tab: string) {
   cy.get('.pf-c-tabs__list').should('be.visible').contains(tab).click();
@@ -70,4 +70,15 @@ Then('sd::user does not see No data message in the {string} graph', (graph: stri
     .children()
     .contains(graph)
     .should('not.contain', 'No data available');
+});
+
+And('user chooses the {string} option', (title: string) => {
+  cy.wait('@waitForCall');
+  cy.get('button[aria-label="Actions"]').click();
+  cy.contains(title).should('be.visible');
+  cy.contains(title).click();
+});
+
+Then('the graph type is disabled', () => {
+  cy.get('button[aria-label="Options menu"]').should('be.disabled');
 });
--- a/frontend/cypress/integration/common/navigation.ts
+++ b/frontend/cypress/integration/common/navigation.ts
@@ -24,6 +24,12 @@ And('user is at the details page for the {string} {string}', (detail: detailType
       break;
     case detailType.Service:
       pageDetail = 'services';
+      cy.intercept({
+        pathname: '**/api/namespaces/bookinfo/services/productpage',
+        query: {
+          objects: ''
+        }
+      }).as('waitForCall');
       break;
     case detailType.Workload:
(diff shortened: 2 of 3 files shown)


josunect (approved):
Tested and works as expected. 
Code looks good. 

Thanks!

frontend/cypress/integration/common/service_details.ts
```diff
@@ -101,3 +101,13 @@ And('user sees table details after selecting a trace', () => {
         cy.get('ul.pf-c-dropdown__menu').should('be.visible').contains('Inbound Metrics');
     });
 });
+
+When('user chooses the {string} option', (title:string) => {
+    cy.get('button[aria-label="Actions"]').click();
+    cy.contains(title).should('exist');
```
matejnesuta, line 107: I replaced the `cy.wait(1000);` line with something hopefully more sensible.

nrfox:
@matejnesuta there's a CI failure for the newly added test if you wouldn't mind looking into it: https://github.com/kiali/kiali/runs/8243955353?check_suite_focus=true

matejnesuta:
Looking at the screenshot right now. The page seems fine. My first guess is that the `'exist'` should be replaced with `'be.visible'`. @nrfox

matejnesuta:
And it failed again. I will investigate it tomorrow I think.

josunect:
If you see the image from the failing test: 
https://github.com/kiali/kiali/suites/8198803432/artifacts/357005649

The node link is disabled because there is no service selected (No namespace is selected). 
When I run this test locally I don't have that error.

matejnesuta:
> 

You are right. Didn't see that yesterday. The namespace is selected, but it takes time to load the graph. That's why it failed probably.
The thing is, the tests are passing locally for me as well, so it's hard for me to predict what's gonna flake.

josunect:
> Added a codition that the mini-graph should be visible before the click happens. I hope everything will be good now.

I think this is more related to the order in which the steps are performed. 
If you notice, the file is expecting to have an ordered list of steps (Given, when, then, and, ...). But when is breaking the sequence.

matejnesuta:
>I think this is more related to the order in which the steps are performed. If you notice, the file is expecting to have an ordered list of steps (Given, when, then, and, ...). But when is breaking the sequence.

I don't think so. The preproccessor doesn't care about the gherkin keywords. For example, this gherkin line ( `But user sees a minigraph`) can be translated with this typescript `Then('user sees a minigraph', () => { ... });`, even though the keywords are different and it doesn't make much sense visually.

josunect (changes requested):
So, looking at the code, the "Show Node Graph" link action is using data from the xhr call to create the url params. It looks like it haven't finished (From the screenshot). So, I will add a condition to wait for the xhr call before perform the click action in that link.
The xhr call: 

/namespaces/{namespace}/services/{service} services serviceDetails

As an example: 

http://localhost:3000/api/namespaces/bookinfo/services/productpage?validate=true

image

image

That could explain why this test is working in my environment (Is pretty quick), but not in CI.

image

matejnesuta:
I think I fixed the test and it looks like it's passing in the CI.

josunect (approved):
LGTM

ScriptingShrimp (approved):
Tested on CRC, all passing, LGTM!

PR #5436, opened 2022-09-06
mTLS lock icon's tooltip contains hint about default "enableAutoMtls:true"

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5263

The idea of this change it to show Kiali users more information about mTLS configuration icon shown in toolbar.
When enableAutoMtls:true then mTLS status is taken from default config.

Screenshot from 2022-09-06 13-44-30

QE: Regression testing is required for "enableAutoMtls:true/false". Namespace wide TLS.

### DIFF
--- a/business/tls.go
+++ b/business/tls.go
@@ -56,7 +56,8 @@ func (in *TLSService) MeshWidemTLSStatus(ctx context.Context, namespaces []strin
 	}
 
 	return models.MTLSStatus{
-		Status: mtlsStatus.MeshMtlsStatus().OverallStatus,
+		Status:          mtlsStatus.MeshMtlsStatus().OverallStatus,
+		AutoMTLSEnabled: *in.enabledAutoMtls,
 	}, nil
 }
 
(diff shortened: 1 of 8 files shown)


jshaughn (commented):
Hi Hayk,
The change to better indicate on the masthead the default MTLS setting looks fine to me.  But I'm unsure if this completely addresses what Lucas was saying about the related validations.  From reading the issue I think he was indicating that the referenced validations may not consider the combined "final" mtls status, only the mtls determined from an existing DR.  Can you comment on the validations because I don't see any change in that logic.

I also had a question about a DR being able to "downgrade" the MTLS.  I see that we combine the status and the meshWideStatus, and if either is true we assume true.  But is is possible that the meshWide is true, but is then disabled by the DR?

Finally, I have a [trivial sub-pr](https://github.com/hhovsepy/kiali/pull/2) that cleans up a couple of this related to the PR, but not touched by the PR.  Just two things I noticed when looking at the code.

hhovsepy:
Thank you @jshaughn , I will make the fixes in further PRs.

hhovsepy:
Hi @jshaughn , have created the tasks for the initial epic. This PR is ready for the second task to be reviewed.

jshaughn (approved):
Looks good, related work will be handled in other epic subtasks.

PR #5438, opened 2022-09-06
Skip Login Page for OpenId strategy

### LINKED ISSUE #5232: openid strategy should not show login page: where are kiali's autologin options?
Hi, when `server.auth.strategy: openid` is chosen, it only allows for one openid provider, so why show a login screen? Marketing? It is unnecessary from the users perspective, and at the minimum it should be made configurable to automatically login imo.

josunect:
The idea is to use a configuration item to enable the feature, SkipLoginScreen, just valid for OpenId at the first version (If required, can be added to others). 

Still need to verify if there is any CORS issue. 

Fixes #5232

### DIFF
--- a/business/authentication/openid_auth_controller.go
+++ b/business/authentication/openid_auth_controller.go
@@ -612,7 +612,6 @@ func (p *openidFlowHelper) checkOpenIdAuthorizationCodeFlowParams() *openidFlowH
 
 	if p.Code == "" {
 		p.Error = &badOidcRequest{Detail: "no authorization code is present"}
-		log.Errorf("Error with Open Id authorization %v", p)
 	}
 
 	return p
@@ -657,22 +656,22 @@ func (p *openidFlowHelper) checkUserPrivileges() *openidFlowHelper {
 		return p
 	}
 
-	//conf := config.Get()
+	conf := config.Get()
 	p.UseAccessToken = false
-	//if conf.Auth.OpenId.DisableRBAC {
-	// When RBAC is on, we delegate some validations to the Kubernetes cluster. However, if RBAC is off
-	// the token must be fully validated, as we no longer pass the OpenId token to the cluster API server.
-	// Since the configuration indicates RBAC is off, we do the validations:
-	err := validateOpenIdTokenInHouse(p)
-	if err != nil {
-		p.Error = &AuthenticationFailureError{
-			HttpStatus: http.StatusForbidden,
-			Reason:     "the OpenID token was rejected",
-			Detail:     err,
+	if conf.Auth.OpenId.DisableRBAC {
+		// When RBAC is on, we delegate some validations to the Kubernetes cluster. However, if RBAC is off
+		// the token must be fully validated, as we no longer pass the OpenId token to the cluster API server.
+		// Since the configuration indicates RBAC is off, we do the validations:
+		err := validateOpenIdTokenInHouse(p)
+		if err != nil {
+			p.Error = &AuthenticationFailureError{
+				HttpStatus: http.StatusForbidden,
+				Reason:     "the OpenID token was rejected",
+				Detail:     err,
+			}
+			return p
 		}
-		return p
-	}
-	/*	} else {
+	} else {
 		// Check if user trying to login has enough privileges to login. This check is only done if
 		// config indicates that RBAC is on. For cases where RBAC is off, we simply assume that the
 		// Kiali ServiceAccount token should have enough privileges and skip this privilege check.
@@ -690,7 +689,7 @@ func (p *openidFlowHelper) checkUserPrivileges() *openidFlowHelper {
 			}
 			return p
 		}
-	}*/
+	}
 
 	return p
 }
@@ -867,8 +866,6 @@ func (p *openidFlowHelper) requestOpenIdToken(redirect_uri string) *openidFlowHe
 
 	if len(cfg.ClientSecret) > 0 {
 		tokenRequest.SetBasicAuth(url.QueryEscape(cfg.ClientId), url.QueryEscape(cfg.ClientSecret))
-		log.Infof("Adding basic auth client ID %v", cfg.ClientId)
-		log.Infof("Client Secret %v", cfg.ClientSecret)
 	}
 
 	tokenRequest.Header.Add("Content-Type", "application/x-www-form-urlencoded")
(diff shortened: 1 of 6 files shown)


jmazzitelli:
Question: if the auth strategy is `anonymous`, should this setting always be assumed "true"? In other words, regardless of what the user sets this new setting to, should we just always skip the login screen when using anonymous auth strategy? Because IIRC, even with anonymous, I seem to recall seeing a "flash" on the screen where Kiali draws the login screen but immediately takes you to the home page when using anonymous.

jshaughn:
I thought the issue was to skip the login screen in any scenario where it would just flash temporarily.  So I'm not sure about why we need a configuration setting.  I would think you should never skip login based on a config setting, if it's needed then it must show?

jmazzitelli:
> I thought the issue was to skip the login screen in any scenario where it would just flash temporarily. So I'm not sure about why we need a configuration setting. I would think you should never skip login based on a config setting, if it's needed then it must show?

I agree. We have hundreds of config settings, let's not add more if we don't need them. :)

josunect:
The reason to do this in the client instead of sending a redirect is to avoid CORS.

jshaughn:
> Remove the configuration item so this behavior is the default.
Keep the login screen for when it is needed...

The only problem with the login screen is that it shows up briefly when the login is unnecessary.  It shows up (and then goes away) when the auth-strategy for `server.auth.strategy; anonymous` and for `server.auth.strategy: openid`.

Don't be confused by the issue writer's request for _at the minimum it should be made configurable to automatically login imo_.  He is saying that it should minimally be configurable to not flash the login screen, but that is written under the assumption that we do it on purpose.  We don't do it on purpose, we just never figured out how to hide it when it is not necessary.  This issue is for hiding it when it's not necessary (because no auth is required, or the authorization has been done externally).

josunect:
Ok, so probably I misunderstand the issue :) 

I thought that the issue was to avoid this screen: 

image

You need to click in order to be redirected to the openId provider screen. 

Or the real issue is the flash in the login screen?

jshaughn:
@josunect , perhaps I was confused, I was under the impression that it was a flash, but an unnecessary click is the same problem, I think.  @israel-hdez Can you comment here?  Is there a reason why we stop here?

israel-hdez:
Well, from the title of the RFE, I understand that it is being requested that Kiali immediately redirects to the 3rd-party system rather than having to click on the "Log In With OpenID" button. The flashing is probably something else.

I don't remember if we offer a "Logout" option for `openid`. If we do, just be careful to not introduce a logout-login loop.

>  Is there a reason why we stop here?

I think it was something discussed in the past, but I no longer remember the answer :-/

israel-hdez:
BTW, I'm in favor of not adding a new config option, despite this would change the behavior of the `openid` auth because IMO, it is a better UX if users are immediately redirected to the 3rd-party login page.

josunect:
So, we should create another ticket for the flash login screen?

josunect:
> > So, we should create another ticket for the flash login screen?
> 
> Yes I'd say they're separate issues. This issue is just about skipping the login page to select a provider since openid can only have one provider, right?

Yes, that's correct. 
Also, I have set the autologin for OpenShift. Or, should be applied to just OpenId? I though it would be more consistent.

josunect:
> > Also, I have set the autologin for OpenShift. Or, should be applied to just OpenId? I though it would be more consistent.
> 
> Try to double-check, because I remember OpenShift auth is already doing the redirection by default. But I may be remembering incorrectly.

I've made the test. It looks like it has the same behaviour: You need to click to be redirected to OpenShift:

josunect:
BTW. If you log out, you are redirected again to the 3rd party login page. 

So, as a summary, this change setup autologin for OpenShift and OpenId auth strategies.

israel-hdez (changes requested):
The auto-redirection works, but looks like it is not possible to redirect unconditionally.

The case that I hit: I logged in with a valid user, but such user didn't have any privileges. Kiali shows an error message in the login page, but since the redirection happens almost instantly, it is very hard to read it. So, for login errors the redirection should not happen.

There is another case which I didn't try, but most likely is impacted: If the session expires, the user is kicked to the login page with a message stating that the session terminated. With these changes, most likely the user will be kicked out and immediately redirected to login. I'm not sure if the behavior would be right for this case. For some people may be confusing, for others it could be OK.

josunect:
> The auto-redirection works, but looks like it is not possible to redirect unconditionally.
> 
> The case that I hit: I logged in with a valid user, but such user didn't have any privileges. Kiali shows an error message in the login page, but since the redirection happens almost instantly, it is very hard to read it. So, for login errors the redirection should not happen.
> 
> There is another case which I didn't try, but most likely is impacted: If the session expires, the user is kicked to the login page with a message stating that the session terminated. With these changes, most likely the user will be kicked out and immediately redirected to login. I'm not sure if the behavior would be right for this case. For some people may be confusing, for others it could be OK.

Thanks, @israel-hdez ! I haven't though in that use case. 

So, there is an structure, LoginStatus, that I think it would be good to use for that: 

  logging,
  loggedIn,
  loggedOut,
  error,
  expired

So, I guess just in the case of logging (And maybe expired?) we can redirect here. 

But, It think the problem is that is not updated. I'm doing some tests.

israel-hdez:
> But, It think the problem is that it doesn't seem to be correctly updated. I'm doing some tests.

Quite possible, as errors come from different sources. Probably, we may need to unify everything into the LoginStatus and the redux store.

josunect:
> > But, It think the problem is that it doesn't seem to be correctly updated. I'm doing some tests.
> 
> Quite possible, as errors come from different sources. Probably, we may need to unify everything into the LoginStatus and the redux store.

josunect:
I think with these two additional checks we can ensure that the redirect will be made: 

- For oauth authentication
- If there is not an error in the authentication (LogginStage.LoggedOut)
- Avoid redirecting loops verifying if the token for oauth has been created

israel-hdez:
I'm still seeing a loop if the user does not have privileges. This is a screen recording showing the issue:

oidc-loop

In the first trial, you may manage to see the message and the redirection happening quite fast. On the second try it's not visible on the recording despite it was visible briefly (so, it's more because of the frame rate of the recording) but what I want to show is that it is redirecting without giving the chance to see the error message.

israel-hdez:
@josunect I opened a PR targeting your fork-branch: https://github.com/josunect/kiali/pull/3

I simply added more conditions to the `if` and now it seems to stop properly when needed. Let me know if that helps.

israel-hdez (approved):
Now, it looks good!

LaikaN57:
@israel-hdez Heads up that this redirect code causes issues when trying to debug login issues. See the issue linked above.

PR #5439, opened 2022-09-06
add 2.3 compat information.

jmazzitelli:
Right now it looks like it will be 1.57, but this could change.

See: https://issues.redhat.com/browse/OSSM-1964


jmazzitelli:
If we are keeping the compat matrix in the code, and if 1.57 is the one we are going with for 2.3, then this PR needs to go in

jshaughn (approved):
If it is taken out it won't be before the 1.57 release.  So LGTM.

Issue #5407, opened 2022-08-24
[CI] Test flake - workload details


nrfox:
Scenario: See Envoy metrics for a workload

Example: https://github.com/kiali/kiali/runs/7997292946?check_suite_focus=true

Screenshot from failure:
![Kiali Workload Details page -- See Envoy metrics for a workload (failed)](https://user-images.githubusercontent.com/6226732/186470166-bbd23bc7-8299-473c-b1a6-28acb5a3bbcb.png)

hhovsepy:
It seems failure happens when "API.getCustomDashboard" is returned, but the sate.dashboard is not set yet, thus dashboards are not drawn: but tests wait until "API.getCustomDashboard" response and expects dashboard to be not empty.
For instance for traces tests, there is a "No traces" div  which disappears when traces response comes.

hhovsepy:
Should be fixed now.

PR #5449, opened 2022-09-08
Added message while loading metrics for custom dashboards

hhovsepy:
Issue https://github.com/kiali/kiali/issues/5407

Added message "Loading metrics" while custom dashboards ate not loaded, waiting for that message to disappear in cypress tests.

### DIFF
--- a/frontend/src/components/Metrics/CustomMetrics.tsx
+++ b/frontend/src/components/Metrics/CustomMetrics.tsx
@@ -9,7 +9,7 @@ import {
   CardBody,
   Checkbox,
   EmptyState,
-  EmptyStateVariant, Title, TitleSizes, EmptyStateBody
+  EmptyStateVariant, Title, TitleSizes
 } from '@patternfly/react-core';
 import { style } from 'typestyle';
 import { serverConfig } from '../../config/ServerConfig';
@@ -217,14 +217,13 @@ class CustomMetrics extends React.Component<Props, MetricsState> {
     }
   }
 
-  renderFetchMetrics = (title, msg) => {
+  renderFetchMetrics = (title) => {
     return (
       <div className={emptyStyle}>
         <EmptyState variant={EmptyStateVariant.small}>
           <Title headingLevel="h5" size={TitleSizes.lg}>
             {title}
           </Title>
-          <EmptyStateBody>{msg}</EmptyStateBody>
         </EmptyState>
       </div>
     );
@@ -258,7 +257,7 @@ class CustomMetrics extends React.Component<Props, MetricsState> {
     const content = (
       <>
         {this.renderOptionsBar()}
-        {this.state.dashboard !== undefined ? dashboard : this.renderFetchMetrics('Loading metrics', 'Metrics will be loaded in a few seconds...')}
+        {this.state.dashboard !== undefined ? dashboard : this.renderFetchMetrics('Loading metrics')}
       </>
     );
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/Metrics/CustomMetrics.tsx
```diff
@@ -226,7 +258,7 @@ class CustomMetrics extends React.Component<Props, MetricsState> {
     const content = (
       <>
         {this.renderOptionsBar()}
-        {this.state.dashboard && dashboard}
+        {this.state.dashboard !== undefined ? dashboard : this.renderFetchMetrics('Loading metrics', 'Metrics will be loaded in a few seconds...')}
```
nrfox, line 261: Kiali seems to be somewhat inconsistent on how it handles loading states. The graph page shows "Loading Graph" text but the other metrics charts and the list pages (istio config, workload, etc) either show nothing or show text saying something like "No Traces" until the data loads. Personally I prefer the loading text as it at least gives you the expectation that something is happening and will load eventually.

Can you have an empty custom tab with no dashboard? Would you then see "Loading metrics" permanently? If you can have an empty custom tab with no dashboard then there would be three states: dashboard is undefined (no dashboard), loading metrics (dashboard is loading could still return empty and go back to no dashboard), dashboard is defined.

frontend/src/components/Metrics/CustomMetrics.tsx
```diff
@@ -198,6 +217,19 @@ class CustomMetrics extends React.Component<Props, MetricsState> {
...
   }
 
+  renderFetchMetrics = (title, msg) => {
+    return (
+      <div className={emptyStyle}>
+        <EmptyState variant={EmptyStateVariant.small}>
+          <Title headingLevel="h5" size={TitleSizes.lg}>
+            {title}
+          </Title>
+          <EmptyStateBody>{msg}</EmptyStateBody>
```
nrfox, line 227: imo you don't need the message. Just "Loading metrics" is good enough and there's no guarantee that the metrics will load in a few seconds :). The graph page just has "Loading Graph" so it'd make it more consistent.

nrfox:
Thanks for tackling one of the flakes!

frontend/src/components/Metrics/CustomMetrics.tsx
jmazzitelli: > there's no guarantee that the metrics will load in a few seconds :)

I agree. Let's not give the user any kind of guarantee or even expectation of load times. What is a few seconds? 2 seconds? 5? 10? 30? 120? :) I would say just tell them its Loading and that's it.

hhovsepy: I first had the title only, but visually it seemed strange to have only Header without body. But agree that "few seconds" is odd.

frontend/src/components/Metrics/CustomMetrics.tsx
hhovsepy: Imho the message "No metrics" could be confusing, as it shows metrics right after.
When there was a problem it loading or showing custom dashboards, there is a error message in notifications:
Screenshot from 2022-09-09 09-14-07
So it s clear, that metrics were loading but error happened.

hhovsepy:
Thanks for the review. have removed the "few seconds..." message.

frontend/src/components/Metrics/CustomMetrics.tsx
nrfox: There's potentially 4 states then!

1. Loading - request for dashboard
2. Dashboard(s) exist - finished loading successfully
3. No dashboard - finished loading successfully but empty dashboard
4. Error in request - finished loading but had an error in the request

It looks like you actually can have the custom tab with no dashboard but this is probably an invalid state and even in the main branch the page is currently blank so state 3 we don't have to worry about and state 4 is handled by the message center and more clear now by having the "Loading message".
Screenshot from 2022-09-09 11-01-22

nrfox (approved):

Issue #5358, opened 2022-08-01
[CI] Test flake - Service details page


nrfox:
Failure in the service details test during the `See span info after selecting a span` test.

Output from CI failure:
```
   Kiali Service Details page
    ✓ See details for productpage (2670ms)
    ✓ See details for service (1065ms)
    ✓ See minigraph for details app. (604ms)
    ✓ See Traffic information (1038ms)
    ✓ See Inbound Metrics for productspage service details (9020ms)
    ✓ See Graph data for productspage service details Inbound Metrics graphs (1599ms)
    ✓ See graph traces for productspage service details (2635ms)
    1) See span info after selecting a span


  7 passing (1m)
  1 failing

  1) Kiali Service Details page
       See span info after selecting a span:
     CypressError: Timed out retrying after 40050ms: `cy.click()` failed because this element is detached from the DOM
```

Link to CI run: https://github.com/kiali/kiali/runs/7617242496?check_suite_focus=true

Screenshot from test failure:
![Kiali Service Details page -- See span info after selecting a span (failed)](https://user-images.githubusercontent.com/6226732/182219165-767c6d6f-4f34-4d34-86b8-b10375994218.png)

josunect:
What about adding a retry configuration to prevent failing the CI due to any flaky test? 

This is suggested in the cypress documentation: 
https://docs.cypress.io/guides/dashboard/flaky-test-management

Aside from reviewing these tests, but it could help.

PR #5450, opened 2022-09-08
Fix Test Flake [hopefully]

jshaughn:
Fixes https://github.com/kiali/kiali/issues/5358

- ensure kebab is visible before click
- only test span select span one time
  - reduces scrolling which may contribute to the problem
  - reduce unnecessary redundancy/speed up test
- remove some unused imports and variables

### DIFF
--- a/frontend/cypress/integration/common/service_details.ts
+++ b/frontend/cypress/integration/common/service_details.ts
@@ -13,16 +13,16 @@ Given('a service in the cluster with a healthy amount of traffic', function () {
     this.targetService = SERVICE;
 });
 
-Then('user sees a list with content {string}', (tab: string) => {
+Then('sd::user sees a list with content {string}', (tab: string) => {
     cy.get('.pf-c-tabs__list').contains(tab);
 });
 
-Then('user sees the actions button', () => {
+Then('sd::user sees the actions button', () => {
     cy.getBySel('wizard-actions').should('be.visible').click();
     cy.getBySel('wizard-actions').siblings().contains("Request Routing");
 });
 
-Then('user sees {string} details information for service {string}', (name: string, version: string) => {
+Then('sd::user sees {string} details information for service {string}', (name: string, version: string) => {
     cy.get('#ServiceDescriptionCard').within(() => {
         cy.get('#pfbadge-S').parent().parent().contains(name); // Service
         cy.get('#pfbadge-A').parent().parent().contains(name); // App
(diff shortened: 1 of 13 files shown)


jshaughn:
@nrfox the main test-related changes are just the addition of a `be.visible` and a small optimization of the tests for two tabs for a selected trace.   But there were also a ton of console errors about duplicate keys and I didn't know if maybe that could contribute to a problem, so the other changes are all related to those, adding/changing keys to many places.  I think also some unintended reformatting, just to make it more messy for you.  In short, I think if the tests pass for you locally we can give it a try.  I've seen 3 or 4 straight successes in the CI.

nrfox:
@jshaughn the tests were failing for me locally and it seems to be because there are two different step definitions that are named the same thing:
https://github.com/kiali/kiali/blob/8f36dbbad2332c358370ccb894c311714367d028/frontend/cypress/integration/common/service_details.ts#L84
https://github.com/kiali/kiali/blob/d823380f784ce9649a4ff90d3d06a1b2d18f96aa/frontend/cypress/integration/common/app_details.ts#L75

Probably it chooses one at random if they're both named the same thing. This has been an issue before. Unfortunately there's not a good way of namespacing these or more importantly know when two names overlap.

If both of these step definitions are the same can you consolidate them into one?

nrfox:
Having feature coupled step definitions seems to be an [anti-pattern](https://cucumber.io/docs/guides/anti-patterns/?lang=java#feature-coupled-step-definitions) that we've unfortunately been following.

frontend/cypress/integration/common/service_details.ts
```diff
@@ -13,85 +13,86 @@ Given('a service in the cluster with a healthy amount of traffic', function () {
...
-    cy.get('#pf-tab-section-1-trace-details').within(() => {
-        cy.get('.pf-c-dropdown__toggle.pf-m-plain').first().should('be.visible').click();
-        cy.get('ul.pf-c-dropdown__menu').should('be.visible').contains('Inbound Metrics');
-    });
+    cy.get('table').should('be.visible')
+      .find('tbody tr')                // ignore thead rows
+      .should('have.length.above', 1)  // retries above cy.find() until we have a non head-row
+      .eq(1)                           // take 1st  row
+      .find('td').eq(4)                // take 5th cell (kebab)
+      .click()
```
nrfox, line 96: There actually are some helper functions for interacting with tables in [table.ts](https://github.com/kiali/kiali/blob/master/frontend/cypress/integration/common/table.ts#L12-L47) but if this works too or is specific to this test then that's fine.

frontend/cypress/integration/common/service_details.ts
```diff
@@ -13,85 +13,86 @@ Given('a service in the cluster with a healthy amount of traffic', function () {
     this.targetService = SERVICE;
 });
 
-Then('user sees a list with content {string}', (tab: string) => {
+Then('sd::user sees a list with content {string}', (tab: string) => {
```
nrfox, line 16: We talked some offline about this. Newer versions of cypress cucumber detect and throw an error for multiple definitions of the same step definition but until we upgrade namespacing will work for keeping them separate.

nrfox (approved):
LGTM

frontend/cypress/integration/common/service_details.ts
jshaughn: Ah, good to know. Thanks.

frontend/cypress/integration/common/service_details.ts
jshaughn: This poor-man's namespacing caught several name-clash issues.  The upgrade is still good, although taking this sort of explicit approach makes it very clear that the steps in the associated .ts file are expected to be used.

jshaughn: I'll merge assuming the relevant CI passes.

PR #5456, opened 2022-09-13
Add kiosk actions for refresh and duration

josunect:
Add kiosk actions when update refresh or duration to send a message to the parent.

Fixes [53](https://github.com/kiali/openshift-servicemesh-plugin/issues/53)

PR #5458, opened 2022-09-13
Refresh Duration in Details page for Overview and propagate parameter…

josunect:
Refresh Duration in Details page for Overview tab and propagate parameters in Traffic tab

Fixes https://github.com/kiali/openshift-servicemesh-plugin/issues/94 

- If customDuration is set to false, the Duration is not refreshed. This is because the duration is not refreshed in the TimeDuration Modal component, it is expecting a time range. 
- Based on this, Is this the expected behavior? Or the component should be fixed?

### DIFF
--- a/frontend/src/components/CytoscapeGraph/MiniGraphCard.tsx
+++ b/frontend/src/components/CytoscapeGraph/MiniGraphCard.tsx
@@ -177,7 +177,7 @@ class MiniGraphCard extends React.Component<MiniGraphCardProps, MiniGraphCardSta
           </CardBody>
         </Card>
         <TimeDurationModal
-          customDuration={false}
+          customDuration={true}
           isOpen={this.state.isTimeOptionsOpen}
           onConfirm={this.toggleTimeOptionsVisibility}
           onCancel={this.toggleTimeOptionsVisibility} />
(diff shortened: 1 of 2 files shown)


israel-hdez (changes requested):
> Based on this, Is this the expected behavior? Or the component should be fixed?

To my eye what is wrong is the indicator.
At the top left of the mini-graph card it's shown the range of the chosen _Duration_. Before the changes you would see the range at the left changing as you change the options in the dialog and the indicator at the right is showing something else. Thus what is showing bad data is the indicator.

After the changes, the indicator becomes in sync with the dialog, yet the fetched data (i.e. the range shown in the mini-graph card) won't match the shown options.

So, it's a matter of passing the right args to the indicator (the `TimeDurationIndicatorButton` component).

I don't know what I was thinking... I should have used same names for the args of the `TimeDurationIndicatorButton` and `TimeDurationModal` components to be more straightforward... Duh! How brilliant!

josunect:
> > Based on this, Is this the expected behavior? Or the component should be fixed?
> 
> To my eye what is wrong is the indicator. At the top left of the mini-graph card it's shown the range of the chosen _Duration_. Before the changes you would see the range at the left changing as you change the options in the dialog and the indicator at the right is showing something else. Thus what is showing bad data is the indicator.
> 
> After the changes, the indicator becomes in sync with the dialog, yet the fetched data (i.e. the range shown in the mini-graph card) won't match the shown options.
> 
> So, it's a matter of passing the right args to the indicator (the `TimeDurationIndicatorButton` component).
> 
> I don't know what I was thinking... I should have used same names for the args of the `TimeDurationIndicatorButton` and `TimeDurationModal` components to be more straightforward... Duh! How brilliant!

Thanks for your comments! I'm reviewing this today.

josunect:
I've added this change in this issue: https://github.com/kiali/kiali/pull/5456
I think it would be better to tests as a whole

josunect:
I've added this change in this issue: https://github.com/kiali/kiali/pull/5456
I think it would be better to tests as a whole

PR #5463, opened 2022-09-14
Bump cypress version and cucumber plugin version

nrfox:
Bumps cypress to v10.0 and updates cucumber cypress plugin to the latest maintained package (ownership of the package transferred somewhat recently). This was done mostly to get checking for duplicate step definitions. This has caused hard to detect errors and flakes in the past but now will be detected and an error will be thrown by the cucumber plugin.

The largest changes are in the way that cypress handles configuration. Instead of a json file, cypress configs are now typescript files. Some duplicate step definitions have been removed and formatting changes by my linter have been applied by whatever files I touched.

**Note:** with the dependency update you'll need to run `yarn` on your local environment while testing this change or after this change has been merged.


ScriptingShrimp:
Big thanks for the effort and update :+1: 
Since this version is critical for our downstream testing, could be this PR merged in next sprint? I don't want to introduce possible braking changes in upstream tests just before end of sprint / release. 
Please let me know what you think.

nrfox:
> Since this version is critical for our downstream testing, could be this PR merged in next sprint? I don't want to introduce possible braking changes in upstream tests just before end of sprint / release.

Yes would like to get it merged as soon as possible. There are a couple of flakes that have surfaced after bumping to cypress 10. I'm not sure why that is other than maybe the cypress runner got slower or faster with the update. Perhaps maybe something changed with the cucumber update as well. Either way I'm actively working on fixing the flakes in this PR and we should be able to get this merged by the end of the sprint.

nrfox:
Looks like cypress is happy with the latest changes when you get a chance to take another look @jshaughn

jshaughn (approved):
Lots of good changes, tests passing.  May need to let people know what they need to do about the lib update.

Issue #5432, opened 2022-09-02
Revisit "round timeseries on client-side with significant decimals"


jshaughn:
This relates to very old work done in https://issues.redhat.com/browse/KIALI-2217, https://issues.redhat.com/browse/KIALI-2207 and PR https://github.com/kiali/kiali/pull/773.

@leandroberetta and I were looking at this code because of some questions Leandro had when working on https://github.com/kiali/kiali/issues/5167, and trying to understand the way the metrics code was dealing with significant digits.

Looking back at the original PR, and the way Prometheus works, it seemed to us that we may be asking Prometheus to evaluate two query segments (two side of an OR), when only one is necessary.

We should take a closer look here, because at larger scale and larger time periods, this could unnecessarily lengthen queries and stress the metrics backend.

jshaughn:
The problem with the current approach is that in exchange for some convenience we're essentially asking Prometheus to perform the same query twice, once while applying rounding, and the second time without.  For example, let's say we're interested in the following query:

`sum(rate(istio_requests_total{source_workload_namespace="bookinfo"} [7h])) by (source_workload, destination_workload) > 0` 

<html>
<body>


{destination_workload="details-v1", source_workload="productpage-v1"} | 0.01612070612981115
-- | --
{destination_workload="reviews-v1", source_workload="productpage-v1"} | 0.005955925941560355
{destination_workload="reviews-v2", source_workload="productpage-v1"} | 0.005439745692871368
{destination_workload="reviews-v3", source_workload="productpage-v1"} | 0.004725034555289475
{destination_workload="unknown", source_workload="reviews-v3"} | 0.0007147111092034501
{destination_workload="unknown", source_workload="reviews-v2"} | 0.0007147111375818939


</body>
</html>

But actually we want to perform rounding, so the natural thing to try would be:

`round(sum(rate(istio_requests_total{source_workload_namespace="bookinfo"} [7h])) by (source_workload, destination_workload), 0.001) > 0`

<html>
<body>


{destination_workload="details-v1", source_workload="productpage-v1"} | 0.016
-- | --
{destination_workload="reviews-v1", source_workload="productpage-v1"} | 0.006
{destination_workload="reviews-v2", source_workload="productpage-v1"} | 0.005
{destination_workload="reviews-v3", source_workload="productpage-v1"} | 0.005
{destination_workload="unknown", source_workload="reviews-v3"} | 0.001
{destination_workload="unknown", source_workload="reviews-v2"} | 0.001


[...]

PR #5466, opened 2022-09-14
Remove the prometheus-performed rounding on metrics queries.

jshaughn:
Fixes https://github.com/kiali/kiali/issues/5432

The OR-based query requires prometheus to basically query the data twice, once with rounding and once without.  The data may still have raw numbers, and so it's questionable whether the rounding provides much value.  It does return a more compact response, but the metrics responses are not particularly large, so the bytes-savings is not worth the extra load on prom or the extra time to generate the response.  Clients can perform any required rounding of the results.

I tested the UI and could not see a place where returning un-rounded numbers changed what was rendered to the user (like showing numbers with many more significant digits).  I think in general our charts, tooltips and such already apply desired rounding.

You can see the impact of the lighter query in this before/after shot of dashboard requests.  Note that the Response Size goes up because more digits are returned, but the Response Time goes down significantly:

Before:
pre-Screenshot-requests

After:
post-Screenshot-requests

### DIFF
--- a/business/metrics_test.go
+++ b/business/metrics_test.go
@@ -43,16 +43,16 @@ func TestGetServiceMetrics(t *testing.T) {
 	q.Quantiles = []string{"0.99"}
 
 	labels := `reporter="source",destination_service_name="productpage",destination_service_namespace="bookinfo"`
-	api.MockRangeRounded("sum(rate(istio_requests_total{"+labels+"}[5m]))", 2.5)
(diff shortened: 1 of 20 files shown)


leandroberetta:
The only thing I'm noticing is the values on the sparkline charts: 

image

But I'm fixing this on #5322 (it's the same component).

jshaughn:
The CI failures are unrelated.  backend test is being fixed in https://github.com/kiali/kiali/pull/5475.  frontent failure looks like a flake.

PR #5467, opened 2022-09-14
Use "token" auth strategy for front-end integration tests

### LINKED ISSUE #5082: Enable github pipeline to run integration tests with token auth enabled
Auth should be enabled by default in the e2e testing environment to make the tests more realistic to a production deployment of Kiali. Probably the most straightforward form of auth to enable for e2e environment is token auth.

israel-hdez:
Fixes #5082

### DIFF
--- a/frontend/cypress/support/commands.ts
+++ b/frontend/cypress/support/commands.ts
@@ -188,10 +188,8 @@ Cypress.Commands.add('login', (username: string, password: string) => {
           .then(result => {
             cy.visit('/');
             cy.get('#token').type(result.stdout, { log: false });
-
-            cy.intercept(Cypress.config('baseUrl') + 'api/authenticate').as('auth')
             cy.get('button[type=submit]').click();
-            cy.wait('@auth')
+            cy.get('#page-sidebar').should('be.visible');
             haveCookie = true;
           });
       }
--- a/frontend/cypress/support/commands.ts
+++ b/frontend/cypress/support/commands.ts
@@ -186,10 +186,14 @@ Cypress.Commands.add('login', (username: string, password: string) => {
       } else if (auth_strategy === 'token') {
         cy.exec('kubectl exec deploy/kiali -n istio-system -- cat /var/run/secrets/kubernetes.io/serviceaccount/token')
           .then(result => {
-            cy.visit('/');
-            cy.get('#token').type(result.stdout, { log: false });
-            cy.get('button[type=submit]').click();
-            cy.get('#page-sidebar').should('be.visible');
+            cy.request({
+              method: 'POST',
+              url: 'api/authenticate',
+              form: true,
+              body: {
+                token: result.stdout
+              }
+            });
             haveCookie = true;
           });
       }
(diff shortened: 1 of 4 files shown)


jshaughn (changes requested):
Code changes look OK but CI is very unhappy.  Note that @nrfox just merged https://github.com/kiali/kiali/pull/5463, I suggest rebasing before looking at the issues.

israel-hdez:
> Code changes look OK but CI is very unhappy.

Programmer's day just was this month, just 2 days ago. This is a good opportunity to say our beloved phrase:

_It works on my machine_ :smile: 

> I suggest rebasing before looking at the issues.

OK, jokes aside... It's OK, I've already read his advice on Slack. I'm on it!

jshaughn (changes requested):
Strill some CI issues, looks like a lot of first-test failures, maybe there is a timing issue/delay due to login, that hurts the first test and then gets resolved.

nrfox:
> Strill some CI issues, looks like a lot of first-test failures, maybe there is a timing issue/delay due to login, that hurts the first test and then gets resolved.

Yes! So we had this exact some problem with the openshift login and we hacked around it for awhile but then moved to logging in through the API and bypassing the UI altogether. The issue is that after you hit the "login" button, kiali redirects you to the overview page. But the tests don't wait for the redirect to happen and call `cy.visit(...)` sometimes _before_ the redirect. You end up in a state where:
1. Login completes
2. Test starts calling `cy.visit(/apps)`
3. Redirect to overview page after login
4. Test assumes that you are on the "/apps" page but you are actually on the overview page because of the redirect.

You can see this in the screenshots of the failures that Kiali is on the overview page instead of the page that should've been visited. All the rest of the tests pass after the first one because the token is saved in the browser session so the login doesn't happen again until a new spec is run which wipes the session.

frontend/cypress/support/commands.ts
```diff
@@ -183,6 +183,15 @@ Cypress.Commands.add('login', (username: string, password: string) => {
...
             haveCookie = true;
           });
+      } else if (auth_strategy === 'token') {
+        cy.exec('kubectl exec deploy/kiali -n istio-system -- cat /var/run/secrets/kubernetes.io/serviceaccount/token')
+          .then(result => {
+            cy.visit('/');
+            cy.get('#token').type(result.stdout, { log: false });
+            cy.get('button[type=submit]').click();
+            cy.get('#page-sidebar').should('be.visible');
+            haveCookie = true;
```
nrfox, line 193: If you login programmatically through the API using `cy.request` like in the openshift login above, you should end up with a valid kiali token cookie. iirc `cy.request` will automatically save the cookie when it gets a `Set Cookie` header. The steps should be MUCH simpler than the openshift steps though for token auth. Hopefully just 1 request to get the cookie given a valid token.

nrfox (changes requested):

frontend/cypress/support/commands.ts
jshaughn: ooh, awesome info Nick, Edgar is on PTO today but should be able to proceed on Monday.

israel-hdez: OK, I did auth via the API and now it looks stable.

jshaughn (approved):
LG2M

nrfox (approved):

Issue #5248, opened 2022-06-24
Relax cluster names in Graph badges and cluster boxes defaults


lucasponce:
Commented with @jshaughn about the possibility to show the "cluster names" in graph badges only when the "Cluster boxes" option is enabled.

Also, the default value is "enabled" and perhaps it can start "disabled" until the user would like to change it.

![image](https://user-images.githubusercontent.com/1662329/175494626-c6aa9052-ac1b-4f92-aa21-89cf7d6712e1.png)

jshaughn:
@lucasponce I could not reproduce this in some brief testing.  It's weird, I'm not sure why you see what you're seeing in the above screenshot.  It would indicate that maybe, somehow, your cluster name, "cluster-default", did not properly get set in the serverConfig passed back to the ui.  The screenshot doesn't show it, but do you see the cluster name in the masthead, if so is it "cluster-default"?  Do you see it in the clusterInfo in the masthead's debug info ?

josunect:
I'm no sure what should be the behavior for "Cluster Boxes", it makes no difference in the graph in my environment (I was expecting to remove (cluster-default): 

![image](https://user-images.githubusercontent.com/49480155/181024399-d4e91314-11e4-4464-bb81-d51eefa4ba68.png)

jshaughn:
@josunect The reason the cluster boxing option had no effect is that you only have one cluster, and we don't add boxes when there is only one option.  Having said that, I'm not sure why `(cluster-default)` is present on your namespace box.  Like the original screenshot from Lucas, I don't see the same thing.  I'm not sure why you are seeing the cluster name unless you truly have multiple clusters in play.

josunect:
Sure, we can have a video call to share this view.
 
Thanks, @jshaughn

jshaughn:
OK, thanks @josunect , that was a useful call because it confirms that we do in fact have different behavior. In your, and presumably Lucas'  env, you are missing the server config indicating the configured home cluster.  As such, the client code assumes your cluster name is the Istio default, "Kubernetes".  But the Istio telemetry is correctly using the configured cluster name, "cluster-default", which is the default cluster name when installing via the hack script.  As such, in your env the graph code basically thinks all of the traffic is on a different, non-home cluster, and therefore is appending the cluster classifier on the labels.

You should see something like this in the debug info:

```
"computedServerConfig": {
      "clusters": {
        "cluster-default": {
          "apiEndpoint": "https://10.96.0.1:443",
[...]

PR #5474, opened 2022-09-15
Change defaults for the istio install hack script

jshaughn:
Change defaults for the istio install hack script such that it doesn't look like a multi-cluster install.  This lets devs run kiali outside of the cluster (via run-kiali.sh) without seeing extra cluster-labeling in the graph. Of course, if you are testing multi-cluster, you can specify args to the hack script to set a non-default cluster-name.

Closes https://github.com/kiali/kiali/issues/5248

### DIFF
--- a/hack/istio/install-istio-via-istioctl.sh
+++ b/hack/istio/install-istio-via-istioctl.sh
@@ -16,7 +16,7 @@
 # CLIENT_EXE_NAME must be either "oc" or "kubectl"
 ADDONS="prometheus grafana jaeger"
 CLIENT_EXE_NAME="oc"
-CLUSTER_NAME="cluster-default"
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):

josunect:
image

Looks good to me. 

Just a though, I guess that is normal that we don't see the cluster name in the top right of the page when we start the hack scripts when we start the hack scripts without specifying --kube-context - As it was set o empty string.

jshaughn:
That's correct, when taking the new defaults we don't set cluster name, and as such, we assume it's single-cluster install.  And so you will not see cluster name in the masthead, you will not see the additional cluster-discriminator in the graph labels, and you will not see the additional Mesh menu item in the left-hand-side.

PR #5479, opened 2022-09-19
Include three additional options in Graph's factory reset

### LINKED ISSUE #4817: Include more settings to clear when resetting to factory settings
Today the factory reset button only affects on the Graph toolbar (the top part) but is not resetting the layout, namespace layout and the new edge options.

This issue is for add this capabilities to the factory reset button.

The configuration to add are in:

graph_reset

israel-hdez:
Add the following options for factory reset:
* Edge mode (hide all/healthy edges)
* Graph layout
* Namespace layout

Fixes #4817

### DIFF
--- a/frontend/src/pages/Graph/GraphToolbar/GraphFind.tsx
+++ b/frontend/src/pages/Graph/GraphToolbar/GraphFind.tsx
@@ -1,16 +1,16 @@
 import * as React from 'react';
-import { Button, Tooltip, ButtonVariant, TextInput, Form } from '@patternfly/react-core';
+import { Button, ButtonVariant, Form, TextInput, Tooltip } from '@patternfly/react-core';
 import { connect } from 'react-redux';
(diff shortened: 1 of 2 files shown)


israel-hdez:
@jshaughn I didn't want to drop another review in your bucket. However, I'm dubious about the changes in `GraphFind` which you crafted carefully. So, I hope you can review. It is small.

jshaughn (changes requested):
@israel-hdez , nothing is easy when it touches Find/Hide :-)  I'm not sure your approach to detecting a reset is the way to go because there is no guarantee that  either of thise layout options changes during the reset.  I was able to crash the PR.  Instead, I think maybe we can do something like what I suggest in this possible sub-pr:

https://github.com/israel-hdez/swscore/pull/28

Note that if you rebase your PR on master this should reduce down to two files.

PR #5487, opened 2022-09-22
[hack] run-kiali to support running in OSSM environment

jmazzitelli:
The istiod service name is different in OSSM. Be able to tell run-kiali.sh what the service name is (e.g. `-isn istiod-custom`).
This is also useful for upstream canary deployments where the istiod service name is also different from "istiod".


jshaughn (approved):
LOK

Issue #5261, opened 2022-06-29
Add Argocd Rollout as workload type to Kiali.


tomahkvt:
We are using Argocd rollout instead Kubernetes Deployment because Argocd rollout has more features.
But we have a problem with Workload visualization in Kiali. The picture you can see below.
![image](https://user-images.githubusercontent.com/13186324/176418107-47388b33-4233-46f8-965b-f78e3a063254.png)
We are using Kiali virsion v1.50.0.
We analyzed the Kiali code and find out that there is no workload type.
Please add the flowing Workload type to kiali:
apiVersion: argoproj.io/v1alpha1  
kind: Rollout 
You can read more about Argocd by using this reference: https://argoproj.github.io/argo-rollouts/migrating.
Thank you in advance.

lucasponce:
I guess you should have minimal support for this, at least in the telemetry.
Perhaps some label at pod generation is missed.

Is this scenario easily deployable in a demo cluster?

The policy for "custom controllers" (i.e. different from Deployments, etc) is to have only basic features from the pods, so we don't plan to support them all or 100% features (contributions are welcome!) but at least using the default pods I think the workloads logic should capture more information than you show here.

Worth to take a look.

leandroberetta:
Hi, I'm also investigating the following issue related to Argo rollouts (https://github.com/kiali/kiali/issues/4210), it could be related but not sure. 

I will take a look too.

jshaughn:
Adding to the backlog as we've now had multiple reports of bad behavior for ArgoCD rollouts, and ArgoCD seems to be a popular approach.

jshaughn:
@leandroberetta The issue here seems to be confusion about what should be a workload from Kiali's perspective.  The workload list seems to be capturing the rollout itself, which I think is the workload controller and it can be responsible for one or more replicasets.  The istio telemetry reports the replicaset as the workload, and so there is a mismatch.  If Kiali were to list the replicasets as the workload then the graph would probably be happy.  We should talk more about how Kiali should handle this situation.

PR #5491, opened 2022-09-23
For Argo rollout, use the controller's replica sets as workloads

### LINKED ISSUE #4210: "KIA0203 This subset's labels are not found in any matching host" For Argo Rollout canary scenario
During canary rollout with Argocd with subset-traffic splitting: https://argoproj.github.io/argo-rollouts/features/traffic-management/istio/#subset-level-traffic-splitting

One of the destination is marked as: "KIA0203 This subset's labels are not found in any matching host", however its live and istio routes traffic correctly.
Screen Shot 2021-07-23 at 3 03 45 PM

Istio routes traffic fine to both destinations and pods with labels from destination rule are available as well
```
$kubectl get pods -l app=example,rollouts-pod-template-hash=6fd44bcc96
NAME                       READY   STATUS    RESTARTS   AGE
example-6fd44bcc96-csqrf   2/2     Running   0          6h2m

 $kubectl get pods -l app=example,rollouts-pod-template-hash=64db898946
NAME                       READY   STATUS    RESTARTS   AGE
example-64db898946-hxq22   2/2     Running   0          23m
```
**Versions used**
Kiali: 1.37.0
Istio: 1.8.1
Kubernetes flavour and version: AKS 1.20

**To Reproduce**
1. Set up Canary Release with subset-level traffic splitting: https://argoproj.github.io/argo-rollouts/features/traffic-management/istio/#subset-level-traffic-splitting
2. Trigger Canary rollout
3. Check destination rule in Kiali

**Expected behavior**
Subset should be available in Kiali.

output of: `curl 'http://<kiali>/kiali/api/namespaces/example-dev/istio/destinationrules/example?validate=true'`
[out.txt](https://github.com/kiali/kiali/files/6871414/out.txt)

jshaughn:
For a custom controller that manages replica sets (like Argo Rollout), defer to the replica sets as the workloads, and not the custom controller. The controller may control multiple replica sets, it may not live in the same namespace as the replica sets, and it may have unusual characteristics. But maybe most importantly, Istio telemetry will report the RS as the workload, and so it is critical to the graph, and linking, to use the RS as the workload.

Closes https://github.com/kiali/kiali/issues/5261
Closes #4210

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -95,6 +95,11 @@ func isWorkloadIncluded(workload string) bool {
 	return !excludedWorkloads[workload]
 }
 
+func isWorkloadValid(workloadType string) bool {
+	_, knownWorkloadType := controllerOrder[workloadType]
+	return knownWorkloadType && isWorkloadIncluded(workloadType)
+}
+
 func (in *WorkloadService) getWorkloadValidations(authpolicies []*security_v1beta1.AuthorizationPolicy, workloadsPerNamespace map[string]models.WorkloadList) models.IstioValidations {
 	validations := checkers.WorkloadChecker{
 		AuthorizationPolicies: authpolicies,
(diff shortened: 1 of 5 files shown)


business/workloads.go
```diff
@@ -1370,12 +1380,12 @@ func fetchWorkload(ctx context.Context, layer *Layer, criteria WorkloadCriteria)
 	// Resolve ReplicaSets from Deployments
 	// Resolve ReplicationControllers from DeploymentConfigs
 	// Resolve Jobs from CronJobs
-	for cname, ctype := range controllers {
-		if ctype == kubernetes.ReplicaSetType {
+	for controllerName, controllerType := range controllers {
+		if controllerType == kubernetes.ReplicaSetType {
 			found := false
 			iFound := -1
 			for i, rs := range repset {
```
leandroberetta, line 1387: I didn't test this yet, but when debugging with the Argo Rollout installed, with two ReplicaSets belonging to it, one of the RS was discarded because it is missing in the repset variable. This is because the GetReplicaSet method only returns one of the ReplicaSets belonging to the Rollout. 

Are you seeing both RS on the Workloads List?.

leandroberetta (approved):
LGTM. Tested locally and solves #4210

business/workloads.go
jshaughn: This was resolved.

PR #5494, opened 2022-09-23
[Cypress] Minor improvements regarding visual regression testing

matejnesuta:
Some minor additions to existing tests. All of these are visual regression tests. This PR contains small test on the Apps page and also filtering on the Workloads page by app and version labels, which was not present before. 

I'll turn this as "Ready for review" if the CI won't fail.


ScriptingShrimp (approved):
Tested on OCP, all passing. Very nice PR, good job!

Issue #5490, opened 2022-09-22
Badging gateways correctly on the graph


nrfox:
Gateways get a special icon on the graph and show their associated hostnames in the node summary panel. The gateway API gateways do not have this icon nor do they show their associated hostnames.

Compare the Istio ingress gateway `istio-ingressgateway` with the Gateway API gateway `gateway` nodes on the graph:
![Screenshot from 2022-09-22 13-56-50](https://user-images.githubusercontent.com/6226732/191818777-823b3cfe-1866-4465-bacc-5bef7f61586a.png)

Node summary panels:
![Screenshot from 2022-09-22 14-01-30](https://user-images.githubusercontent.com/6226732/191819109-da7194bd-8bfb-4d9f-8bc7-b32dbaf4b0ca.png)

![Screenshot from 2022-09-22 14-01-38](https://user-images.githubusercontent.com/6226732/191819140-da9f02fa-15e6-4882-b673-66bd253b5791.png)

PR #5496, opened 2022-09-26
Added Gateway API support in Graph

hhovsepy:
Issue https://github.com/kiali/kiali/issues/5490

Now the Gateway API applications in Graph page are shown with similar icons as Ingress gateway ones.

Screenshot from 2022-09-26 14-57-43

### DIFF
--- a/frontend/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/frontend/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -266,11 +266,11 @@ export class GraphStyles {
       if (node.hasWorkloadEntry) {
         badges = `<span class="${NodeIconWorkloadEntry} ${badgeMargin(badges)}"></span> ${badges}`;
       }
-      if (node.isGateway?.gatewayAPIInfo?.hostnames?.length !== undefined) {
-        badges = `<span class="${NodeIconGateway} ${badgeMargin(badges)}"></span> ${badges}`;
-      }
       if (node.isRoot) {
-        if (node.isGateway?.ingressInfo?.hostnames?.length !== undefined) {
+        if (
+          node.isGateway?.ingressInfo?.hostnames?.length !== undefined ||
+          node.isGateway?.gatewayAPIInfo?.hostnames?.length !== undefined
+        ) {
           badges = `<span class="${NodeIconGateway} ${badgeMargin(badges)}"></span> ${badges}`;
         }
         badges = `<span class="${NodeIconRoot} ${badgeMargin(badges)}"></span> ${badges}`;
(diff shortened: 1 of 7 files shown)


frontend/src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -266,6 +266,9 @@ export class GraphStyles {
       if (node.hasWorkloadEntry) {
         badges = `<span class="${NodeIconWorkloadEntry} ${badgeMargin(badges)}"></span> ${badges}`;
       }
+      if (node.isGateway?.gatewayAPIInfo?.hostnames?.length !== undefined) {
+        badges = `<span class="${NodeIconGateway} ${badgeMargin(badges)}"></span> ${badges}`;
+      }
```
jshaughn, line 271: This is probably OK but to totally avoid the chance of multiple gateway badges, maybe this should be incorporated into the `if else` directly below.  Something like:

```
     if (node.isRoot) {
        if (
          node.isGateway?.ingressInfo?.hostnames?.length !== undefined ||
          node.isGateway?.gatewayAPIInfo?.hostnames?.length !== undefined
        ) {
          badges = `<span class="${NodeIconGateway} ${badgeMargin(badges)}"></span> ${badges}`;
        }
        badges = `<span class="${NodeIconRoot} ${badgeMargin(badges)}"></span> ${badges}`;
      } else {
        if (
          node.isGateway?.egressInfo?.hostnames?.length !== undefined ||
          node.isGateway?.gatewayAPIInfo?.hostnames?.length !== undefined
        ) {
          badges = `<span class="${NodeIconGateway} ${badgeMargin(badges)}"></span> ${badges}`;
        }
      }
```

Actually, Are GatewayAPI gateways always ingress?  In that case eliminate the change in `else`.

jshaughn (commented):
Looks good.  Tested and it's working, although it took me a little while to remember that we need to add app/version labels after following the instructions in https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/.

One suggested code change but not a blocker.

frontend/src/components/CytoscapeGraph/graphs/GraphStyles.ts
hhovsepy: In my initial commit there was Gateway API under isRoot, then I changed to have it separate and independent of being isRoot edge, because Gateway API is always Ingress. Not sure if Gateway API should be always root edge or not.

jshaughn: I think it's safe to assume that an ingress gateway will always also be a root (aka traffic source).  But if there is any doubt it can also be added to the `else` clause (now or later or never).

jshaughn (approved):
One non-blocking comment, but LGTM!

Issue #5239, opened 2022-06-23
Provide Kiali inside distroless container


mblaschke-daimlertruck:
Please provide Kiali also with a distroless container (like Istio is also doing it) https://github.com/GoogleContainerTools/distroless
This would be a big security win as no additional tools (eg shell) are installed.

israel-hdez:
Closing as done in #5505

PR #5505, opened 2022-09-28
Add distroless image

zhech2:
Issue #5239

Add distroless image definitions following istio's pattern

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -105,3 +105,4 @@ jobs:
         docker login -u ${{ secrets.QUAY_USER }} -p ${{ secrets.QUAY_PASSWORD }} quay.io
 
         make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-push-kiali-quay
+        make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-distroless-push-kiali-quay
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -104,5 +104,5 @@ jobs:
       run: |
         docker login -u ${{ secrets.QUAY_USER }} -p ${{ secrets.QUAY_PASSWORD }} quay.io
 
-        make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-push-kiali-quay
-        make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-distroless-push-kiali-quay
+        make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-all-push-kiali-quay
+        
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
-GO_VERSION_KIALI = 1.17.10
+GO_VERSION_KIALI = 1.17.13
 
 # Identifies the Kiali container image that will be built.
 IMAGE_ORG ?= kiali
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
-GO_VERSION_KIALI = 1.17.13
+GO_VERSION_KIALI = 1.17.10
 
 # Identifies the Kiali container image that will be built.
 IMAGE_ORG ?= kiali
--- a/deploy/docker/Dockerfile-multi-arch-distroless
+++ b/deploy/docker/Dockerfile-multi-arch-distroless
@@ -1,0 +1,36 @@
+# source https://github.com/istio/istio/blob/master/docker/Dockerfile.distroless
+FROM gcr.io/distroless/static-debian11:latest-${TARGETARCH:-amd64} AS distroless_source
+
+LABEL maintainer="kiali-dev@googlegroups.com"
+
+# prepare a base dev to modify file contents
+FROM ubuntu:focal as ubuntu_source
+
+# Modify contents of container
+COPY --from=distroless_source /etc/ /home/etc
+COPY --from=distroless_source /home/nonroot /home/nonroot
+RUN echo kiali:x:1000: >> /home/etc/group
+RUN echo kiali:x:1000:1000:/home/kiali:/sbin/nologin >> /home/etc/passwd
+
+# Customize distroless with the following:
+# - password file
+# - groups file
+# - /home/nonroot directory
+FROM distroless_source
+COPY --from=ubuntu_source /home/etc/passwd /etc/passwd
+COPY --from=ubuntu_source /home/etc/group /etc/group
+COPY --from=ubuntu_source /home/nonroot /home/nonroot
+
+ENV KIALI_HOME=/opt/kiali \
+    PATH=$KIALI_HOME:$PATH
+
+WORKDIR $KIALI_HOME
+
+ARG TARGETARCH
+COPY kiali-${TARGETARCH} $KIALI_HOME/kiali
+
+COPY --chown=kiali:kiali console/ $KIALI_HOME/console/
+
+USER 1000
+
+ENTRYPOINT ["/opt/kiali/kiali"]
--- a/make/Makefile.container.mk
+++ b/make/Makefile.container.mk
@@ -117,3 +117,11 @@ container-multi-arch-push-kiali-operator-quay: .ensure-operator-repo-exists .ens
 container-multi-arch-push-kiali-quay: .ensure-buildx-builder .prepare-kiali-image-files
 	@echo Pushing Kiali multi-arch image to ${QUAY_TAG} using docker buildx
 	docker buildx build --push --pull --no-cache --builder=kiali-builder $(foreach arch,${TARGET_ARCHS},--platform=linux/${arch}) $(foreach tag,${QUAY_TAG},--tag=${tag}) -f ${OUTDIR}/docker/Dockerfile-multi-arch ${OUTDIR}/docker
+
+## container-multi-arch-distroless-push-kiali-quay: Pushes the Kiali multi-arch distroless image to quay.
+container-multi-arch-distroless-push-kiali-quay: .ensure-buildx-builder .prepare-kiali-image-files
+	@echo Pushing Kiali multi-arch distroless image to ${QUAY_TAG}-distroless using docker buildx
+	docker buildx build --load --pull --builder=kiali-builder --platform=linux/amd64 $(foreach tag,${QUAY_TAG},--tag=${tag}-distroless) -f ${OUTDIR}/docker/Dockerfile-multi-arch-distroless ${OUTDIR}/docker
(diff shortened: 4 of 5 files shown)


jmazzitelli:
Thanks for the contribution. This looks additive (meaning, it doesn't affect how things work or are built today) so this is a good thing.

However, I don't see any instructions on how to use the distroless install. The helm charts or operator won't create Deployments by default that pull in the distroless containers - so there needs to be some kind of instructions on how to use these distroless images. Can you provide a comment in here as to how you see these distroless images being installed? From that we can decide how we can document this and where to document it.

Makefile
```diff
@@ -29,7 +29,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
-GO_VERSION_KIALI = 1.17.10
+GO_VERSION_KIALI = 1.17.13
```
jmazzitelli, line 32: Is this required for this PR? Why is it necessary to bump up the Go version?

If necessary, then before we merge this, we'll need to make an announcement to the dev community that a bump in Go is being introduced ... we don't like forcing infrastructure/dev tooling updates on developers without giving notice.

make/Makefile.container.mk
```diff
@@ -117,3 +117,11 @@ container-multi-arch-push-kiali-operator-quay: .ensure-operator-repo-exists .ens
 container-multi-arch-push-kiali-quay: .ensure-buildx-builder .prepare-kiali-image-files
 	@echo Pushing Kiali multi-arch image to ${QUAY_TAG} using docker buildx
 	docker buildx build --push --pull --no-cache --builder=kiali-builder $(foreach arch,${TARGET_ARCHS},--platform=linux/${arch}) $(foreach tag,${QUAY_TAG},--tag=${tag}) -f ${OUTDIR}/docker/Dockerfile-multi-arch ${OUTDIR}/docker
+
+## container-multi-arch-distroless-push-kiali-quay: Pushes the Kiali multi-arch distroless image to quay.
+container-multi-arch-distroless-push-kiali-quay: .ensure-buildx-builder .prepare-kiali-image-files
+	@echo Pushing Kiali multi-arch distroless image to ${QUAY_TAG}-distroless using docker buildx
+	docker buildx build --load --pull --builder=kiali-builder --platform=linux/amd64 $(foreach tag,${QUAY_TAG},--tag=${tag}-distroless) -f ${OUTDIR}/docker/Dockerfile-multi-arch-distroless ${OUTDIR}/docker
+	# docker buildx build --push --pull --builder=kiali-builder $(foreach arch,${TARGET_ARCHS},--platform=linux/${arch}) $(foreach tag,${QUAY_TAG},--tag=${tag}-distroless) -f ${OUTDIR}/docker/Dockerfile-multi-arch-distroless ${OUTDIR}/docker
```
jmazzitelli, line 125: Since this is commented out, please remove it.

NOTE: are you sure you want to do the `--load` and not the `--push` -- Please confirm -- it seems to me you want to do this??

```suggestion
	docker buildx build --push --pull --builder=kiali-builder $(foreach arch,${TARGET_ARCHS},--platform=linux/${arch}) $(foreach tag,${QUAY_TAG},--tag=${tag}-distroless) -f ${OUTDIR}/docker/Dockerfile-multi-arch-distroless ${OUTDIR}/docker
```

jmazzitelli (changes requested):
Do not merge until the previous questions are addressed.

Also, please see the build failure - bumping up Go requires other changes. Please make sure bumping up Go is required for this PR - otherwise, it should be handled in another PR separately from this:

```
fatal: go version '1.17.[10](https://github.com/kiali/kiali/actions/runs/3146580633/jobs/5115395767#step:7:11)' should be at least '1.17.13'
```

zhech2:
Wow, thanks for all the feedback so quickly.  I'll look to adjust the PR as per your feedback.

jmazzitelli:
> Wow, thanks for all the feedback so quickly. I'll look to adjust the PR as per your feedback.

We've had a few people ask for this - so, thanks :)

zhech2:
The images are named as follows: v1.56.1-distroless.  Is this naming acceptable?

I tested it by switching our kiali deployment's image tag to include "-distroless" and it worked. 

If there is anything else I can do to help with documentation or the helm chart let me know.

jmazzitelli:
> The images are named as follows: v1.56.1-distroless. Is this naming acceptable?

Yes, I think that naming is OK. In the future (not in this PR), what I think we'll want to do is eventually switch it so the "distroless" is the default tag ("vX.Y") and the full distro container (the current one we have today) gets named something like "vX.Y-full" or something. This would allow people to install a full container (in case they need to debug things and require some tools that come with the full distro) but by default they will get a smaller, more secure distroless container.

But I want this to be released first like it is so people in the community can play and test with the distroless first. If it works OK and no one finds problems with it, then we can make a switch. We definitely want to give the distroless container time to soak in the community before making it the default image.

As for documentation, not sure where to put this yet. Doesn't have to be in this PR. I'll think about it. But here's what I think people need to know in order to use this distroless container.

The way you change the image to use distroless depends on the install mechanism. If using the server helm chart, you'll pass this option in to the helm install command: `--set deployment.image_version=v1.57.0-distroless`. If using the operator, you will need to first configure the operator to accept ad-hoc images (see [this](https://github.com/kiali/helm-charts/blob/v1.57.0/kiali-operator/values.yaml#L65-L69)) and then the user defines the `spec.deployment.image_version` in the Kiali CR.

jmazzitelli:
I ran some tests on my own fork and quay repo. I was able to run the nightly github action and it successfully built and pushed the full and the distroless images. I then installed the distroless:

```
helm install -n istio-system --set deployment.image_name=quay.io/jmazzitelli/kiali --set deployment.image_version=latest-distroless kiali-server kiali/kiali-server
```

I checked the kiali server is using the distroless image:
```
$ oc get pods -n istio-system -l app=kiali -o jsonpath={.items[0].spec.containers[0].image}
quay.io/jmazzitelli/kiali:latest-distroless
```

The logs looks normal:
```
$ oc logs -n istio-system -l app=kiali
2022-09-29T11:59:16Z INF Some validation errors will be ignored [KIA1201]. If these errors do occur, they will still be logged. If you think the validation errors you see are incorrect, please report them to the Kiali team if you have not done so already and provide the details of your scenario. This will keep Kiali validations strong for the whole community.
2022-09-29T11:59:16Z INF Mesh Name: [Istio], Mesh Version: [1.15.0]
2022-09-29T11:59:16Z INF Initializing Kiali Cache
2022-09-29T11:59:16Z INF Adding a RegistryRefreshHandler
2022-09-29T11:59:16Z INF [Kiali Cache] Waiting for cluster-scoped cache to sync
2022-09-29T11:59:16Z INF [Kiali Cache] started
2022-09-29T11:59:16Z INF Kiali Cache is active for namespaces [.*]
2022-09-29T11:59:16Z INF Server endpoint will start at [:20001/kiali]
2022-09-29T11:59:16Z INF Server endpoint will serve static content from [/opt/kiali/console]
2022-09-29T11:59:16Z INF Starting Metrics Server on [:9090]
```

And I confirmed some things are missing (to be expected in a distroless environment):

```
$ kubectl exec -n istio-system deploy/kiali -- ls
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "ls": executable file not found in $PATH: unknown
command terminated with exit code 126
```

I quickly poked around the Kiali UI and did not see any errors. It was just a smoke test, but it looks OK.

I then installed the full image built by the nightly workflow (just to make sure that didn't break) and I can see it runs fine and I can access some executables that come with the distro that kiali runs in:

```
$ oc get pods -n istio-system -l app=kiali -o jsonpath={.items[0].spec.containers[0].image}
quay.io/jmazzitelli/kiali:latest

$ kubectl exec -n istio-system deploy/kiali -- ls
console
kiali
```

deploy/docker/Dockerfile-multi-arch-distroless
```diff
@@ -0,0 +1,36 @@
...
+COPY --from=ubuntu_source /home/etc/passwd /etc/passwd
+COPY --from=ubuntu_source /home/etc/group /etc/group
+COPY --from=ubuntu_source /home/nonroot /home/nonroot
+
+ENV KIALI_HOME=/opt/kiali \
+    PATH=$KIALI_HOME:$PATH
+
+WORKDIR $KIALI_HOME
+
+ARG TARGETARCH
```
jmazzitelli, line 29: This needs to be moved up to line 1 (before the first time TARGETARCH is used in this Dockerfile).  The way this is currently, only amd64 archs will work. Please see the ARG scope documentation to understand why you need to do this: https://docs.docker.com/engine/reference/builder/#scope

deploy/docker/Dockerfile-multi-arch-distroless
```diff
@@ -0,0 +1,36 @@
+# source https://github.com/istio/istio/blob/master/docker/Dockerfile.distroless
+FROM gcr.io/distroless/static-debian11:latest-${TARGETARCH:-amd64} AS distroless_source
```
jmazzitelli, line 2: Remove the default value. We expect and require the TARGETARCH arg to be passed in - if not, we expect and require errors to occur. We do not want to blindly use "amd64" as the default here because if this is building other arches, this will be wrong. This currently hid the bug that I reported in my previous comment - the ARG is defined further down below, so when this line 1 is evaluated, the TARGETARCH isn't the one passed and is always defaulting to amd64, which is wrong for the other arch builds.

So we want this:
```suggestion
FROM gcr.io/distroless/static-debian11:latest-${TARGETARCH} AS distroless_source
```

jmazzitelli (changes requested):
All my tests show this working, however, I can only test on amd64 hardware. I noticed a problem in this stuff for the other archs by reading the Dockerfile. See my comments for what needs to be fixed.

.github/workflows/nightly.yml
```diff
@@ -104,4 +104,5 @@ jobs:
       run: |
         docker login -u ${{ secrets.QUAY_USER }} -p ${{ secrets.QUAY_PASSWORD }} quay.io
 
-        make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-push-kiali-quay
\ No newline at end of file
+        make -e DOCKER_CLI_EXPERIMENTAL=enabled build-linux-multi-arch container-multi-arch-all-push-kiali-quay
+
```
jmazzitelli, line 108: While we are here, please remove this line with the spaces :) It is trivial, but lines with spaces in them bother me for some reason :) 
```suggestion
```

jmazzitelli (changes requested):
another trivial change request

deploy/docker/Dockerfile-multi-arch-distroless
jmazzitelli: hmmm... I just ran a test and it failed. The TARGETARCH wasn't set when I expected it to be. I'm looking into this myself to find out why moving the ARG up to line 1 doesn't work

jmazzitelli: Ah... https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
```
The following ARG variables are set automatically:
...
TARGETARCH - architecture component of TARGETPLATFORM
```

Based on this, I guess we never have to specify "ARG TARGETARCH"

jmazzitelli:
Wow, OK, I just misunderstand how TARGETARCH is defined. Because when I took out the ARG line completely, I got:

`ERROR: failed to solve: failed to compute cache key: "/kiali-" not found: not found`

So you need to define it even though it is a built-in arg? I don't get it. But, I'm going to merge this as-is. Since it works as-is and anything I think needs to change will break it :)

jmazzitelli (approved):

Issue #5508, opened 2022-09-29
switch the published images to be distroless


jmazzitelli:
We are now producing distroless containers: https://github.com/kiali/kiali/pull/5505, https://github.com/kiali/kiali/pull/5511

Currently, the distroless containers will have `-distroless` after the version tag (e.g. "latest-distroless" or "v1.58.0-distroless").

We will want to switch it so the "distroless" is the default image ("latest" or "vX.Y.Z") and the containers with the full distro (the current ones we have today) get named something like "vX.Y.Z-full" or something like that.

This would allow people to install a container with a distro (in case they need to debug things which require some tools that come with the full distro) but by default they will get a smaller, more secure distroless container.

We don't want to do this immediately. We only just merged that PR so we haven't had a chance to soak the distroless containers out in the community for feedback. If the distroless is OK and no one finds problems with it, then we can make a switch. We definitely want to give the distroless container time to soak in the community before making it the default image.

Today, the way you change the image to use distroless depends on the install mechanism. If using the server helm chart, you'll pass this option in to the helm install command: `--set deployment.image_version=latest-distroless`. If using the operator, you will need to first configure the operator to accept ad-hoc images (see [this](https://github.com/kiali/helm-charts/blob/v1.57.0/kiali-operator/values.yaml#L65-L69)) and then the user defines the `spec.deployment.image_version` in the Kiali CR.

PR #5509, opened 2022-09-29
switch images so the distroless is the default image

jmazzitelli:
The full distro image will be flagged with "-distro" in the tag.
fixes: https://github.com/kiali/kiali/issues/5508

### DIFF
--- a/make/Makefile.container.mk
+++ b/make/Makefile.container.mk
@@ -115,13 +115,13 @@ container-multi-arch-push-kiali-operator-quay: .ensure-operator-repo-exists .ens
 
 ## container-multi-arch-push-kiali-quay: Pushes the Kiali multi-arch image to quay.
 container-multi-arch-push-kiali-quay: .ensure-buildx-builder .prepare-kiali-image-files
-	@echo Pushing Kiali multi-arch image to ${QUAY_TAG} using docker buildx
-	docker buildx build --push --pull --no-cache --builder=kiali-builder $(foreach arch,${TARGET_ARCHS},--platform=linux/${arch}) $(foreach tag,${QUAY_TAG},--tag=${tag}) -f ${OUTDIR}/docker/Dockerfile-multi-arch ${OUTDIR}/docker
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Easiest way to test if you are running with a distroless image, is to try to invoke a command like `ls` in the pod and it should fail:

```
$ kubectl exec -n istio-system deploy/kiali -- ls
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "ls": executable file not found in $PATH: unknown
command terminated with exit code 126
```

If you get a list of files (i.e. if `ls` succeeds) then you know you are running with a full distro image:
```
$ kubectl exec -n istio-system deploy/kiali -- ls
console
kiali
```

jmazzitelli:
@zhech2 @mblaschke-daimlertruck

You both were interested in [distroless kiali containers](https://github.com/kiali/kiali/issues/5239). Thanks to the work contributed by @zhech2 we are able to produce them now. We will eventually make these the default image (barring any problems), but we need the distroless container to be used/tested out in the community to make sure there are no hidden problems with it.

So if you are able to run with "latest-distroless" and can report back after using it for a while, that would be helpful in giving us confidence that the distroless image can be used as the default image that people will pull.

The "latest-distroless" in [quay.io/kiali/kiali](https://quay.io/repository/kiali/kiali?tab=tags) is the latest from master branch (basically it is 1.57.0 with whatever else was added to master since the beginning of this week).

To deploy that distroless image today depends on the install mechanism. If using the server helm chart, you'll pass this option in to the helm install command: `--set deployment.image_version=latest-distroless`. If using the operator, you will need to first configure the operator to accept ad-hoc images (see [this](https://github.com/kiali/helm-charts/blob/v1.57.0/kiali-operator/values.yaml#L65-L69)) and then the user defines the `spec.deployment.image_version` in the Kiali CR.

jshaughn (approved):
LGTM

PR #5511, opened 2022-09-29
WIP use ubi9-micro as the distroless base image

jmazzitelli:
I want to see how easy/hard it would be to convert the distroless builds to use ubi9-micro. For some reason, github CI can't run this (but it works on my box).

The idea is that you want to manually create a user and group "kiali" with UI of 1000 (this is how it is done today with the distroless build today). Not sure if there is another way to do this when using ubi9-micro.

```
 > [linux/s390x stage-4 1/5] RUN echo kiali:x:1000: >> /etc/group:
#10 0.259 Fatal glibc error: CPU lacks VXE support (z14 or later required)
------
Dockerfile-multi-arch-distroless:11
--------------------
   9 |     
  10 |     # Modify contents of container
  11 | >>> RUN echo kiali:x:1000: >> /etc/group
  12 |     RUN echo kiali:x:1000:1000:/home/kiali:/sbin/nologin >> /etc/passwd
  13 |     
--------------------
ERROR: failed to solve: executor failed running [/bin/sh -c echo kiali:x:1000: >> /etc/group]: exit code: 127
```

### DIFF
--- a/deploy/docker/Dockerfile-multi-arch-distroless
+++ b/deploy/docker/Dockerfile-multi-arch-distroless
@@ -1,25 +1,15 @@
-# source https://github.com/istio/istio/blob/master/docker/Dockerfile.distroless
-FROM gcr.io/distroless/static-debian11:latest-${TARGETARCH:-amd64} AS distroless_source
+FROM registry.access.redhat.com/ubi9-micro AS base-amd64
+FROM registry.access.redhat.com/ubi9-micro AS base-arm64
+FROM registry.access.redhat.com/ubi9-micro AS base-s390x
+FROM registry.access.redhat.com/ubi9-micro AS base-ppc64le
+
+FROM base-${TARGETARCH}
 
 LABEL maintainer="kiali-dev@googlegroups.com"
 
-# prepare a base dev to modify file contents
-FROM ubuntu:focal as ubuntu_source
-
 # Modify contents of container
-COPY --from=distroless_source /etc/ /home/etc
-COPY --from=distroless_source /home/nonroot /home/nonroot
-RUN echo kiali:x:1000: >> /home/etc/group
-RUN echo kiali:x:1000:1000:/home/kiali:/sbin/nologin >> /home/etc/passwd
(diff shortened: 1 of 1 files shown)


jmazzitelli:
similar error when github ci tries to build on ppc arch:
```
Fatal glibc error: CPU lacks float128 support (POWER 9 or later required)
```

When I test on my local amd64 machine, this stuff seems to be working - I can even run the image and see the files chowned to the kiali user that is created by this line that fails. So I don't know why github has a problem with this.

jmazzitelli:
The problem was ubi9 - doesn't look like the github infrastructure can run the new 9 version. Going down to ubi8-micro works.

jmazzitelli:
I tested building the dev image on my box and it works. I also tested in my github fork and saw the CI can build the multi-arch image with the ubi8-micro base image. So all looks good.

jmazzitelli:
Here is how I tested.

1. Start minikube and install istio
2. Build and deploy distroless kiali: `make -e CLUSTER_TYPE=minikube cluster-push operator-create kiali-create`
3. Wait for kiali to start up
4. Try to run curl in the kiali pod: `kubectl exec -n istio-system deploy/kiali -- curl` and you should get something like `OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "curl": executable file not found in $PATH: unknown`
5. But now access the Kiali UI and see that it works (I run `hack/kiali-port-forward.sh` and go to http://localhost:20001/kiali/console)
6. Confirm you can see the logs from the pod: `kubectl logs -n istio-system deploy/kiali`

I did similar testing on OpenShift 4.11 and it works there, too.

jmazzitelli:
As for the size of the images:

```
latest-distroless 105 MB
latest            193 MB
```

jshaughn (approved):
I didn't test this but I am +1 for moving to RH ubi images.  Despite being a larger than the debian image it keeps all of our containers based on RH images, giving us consistency and a little more control (I think).

PR #5518, opened 2022-10-03
We need to update deployment annotations after we create it

FilipB:
** Describe the change **
Fixing err demo installation script on maistra


jmazzitelli (approved):
lgtm

PR #5526, opened 2022-10-04
[Cypress] added @smoke annotations to related feature files

ScriptingShrimp:
added `@smoke` annotations to related feature files

### DIFF
--- a/frontend/cypress/integration/featureFiles/kiali_about.feature
+++ b/frontend/cypress/integration/featureFiles/kiali_about.feature
@@ -6,6 +6,7 @@ Feature: Kiali help about verify
     Given user is at administrator perspective
     And user is at the "overivew" page
 
+  @smoke
   Scenario: Open Kiali about page
 
     And user clicks on Help Button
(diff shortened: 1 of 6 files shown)


jmazzitelli:
I don't think you meant to call out user "@"smoke in the description :)

Perhaps you should say it with backticks -  `@smoke`

ScriptingShrimp:
whoops, sorry :sweat_smile:

jmazzitelli (approved):
I don't see any issues here.

But, can I ask a question - what does the `@smoke` annotation do? :}

ScriptingShrimp:
Annotations like this will help us separate testing suite into more specific layers; i.e. if we deploy on third party infrastructure, we want to check only basic functionality and do not trigger full testing (functional & regression). Its cheaper and its up to us to decide what are relevant tests for our use. :point_right:  https://en.wikipedia.org/wiki/Smoke_testing_(software)

ScriptingShrimp:
we already use that here - in scripts via `-e TAGS=` 
https://github.com/kiali/kiali/blob/d11b634883cb099d74e9579971f6a62e20fec311/frontend/package.json#L44

smoke:
// offtopic - @ScriptingShrimp don't worry calling me out like that happens all the time, so I don't mind :) @jmazzitelli thank you for caring about that :beers:

PR #5529, opened 2022-10-04
Adding support for maistra and custom istio namespace

FilipB:
Adding support for maistra so the sidecars are enabled for all used demo applications. Also adding support for custom istio namespace.

### DIFF
--- a/hack/istio/install-testing-demos.sh
+++ b/hack/istio/install-testing-demos.sh
@@ -9,6 +9,33 @@
 
 set -eu
   
+# Given a namepace, prepare it for inclusion in Maistra's control plane
+# This means:
+# 1. Create a SMM
+# 2. Annotate all of the namespace's Deployments with the sidecar injection annotation if enabled
+prepare_maistra() {
+  local ns="${1}"
+
+  cat <<EOM | ${CLIENT_EXE} apply -f -
+apiVersion: maistra.io/v1
+kind: ServiceMeshMember
+metadata:
+  name: default
+  namespace: ${ns}
+spec:
+  controlPlaneRef:
+    namespace: ${ISTIO_NAMESPACE}
+    name: "$(${CLIENT_EXE} get smcp -n ${ISTIO_NAMESPACE} -o jsonpath='{.items[0].metadata.name}' )"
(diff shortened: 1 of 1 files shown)


hack/istio/install-testing-demos.sh
```diff
@@ -100,12 +136,15 @@ HELPMSG
 done
 
 IS_OPENSHIFT="false"
+IS_MAISTRA="false"
 if [[ "${CLIENT_EXE}" = *"oc" ]]; then
   IS_OPENSHIFT="true"
+  IS_MAISTRA=$([ "$(oc get crd | grep servicemesh | wc -l)" -gt "0" ] && echo "true" || echo "false")
```
jmazzitelli, line 142:
```suggestion
  IS_MAISTRA=$([ "$(${CLIENT_EXE} get crd | grep servicemesh | wc -l)" -gt "0" ] && echo "true" || echo "false")
```

hack/istio/install-testing-demos.sh
```diff
@@ -152,6 +191,7 @@ else
 
   echo "Deleting the 'sleep' app in the 'sleep' namespace..."
   ${CLIENT_EXE} delete -n sleep -f ${ISTIO_DIR}/samples/sleep/sleep.yaml
+  ${CLIENT_EXE} delete smm default -n "sleep" --ignore-not-found=true
```
jmazzitelli, line 194: Because line 10 has `set -eu` - this line is going to cause this script to abort right here on a non-Maistra environment. The `--ignore-not-found=true` doesn't help - this will still exit with non-zero if the type ServiceMeshMember doesn't exist:

```
$ kubectl delete smm default -n "sleep" --ignore-not-found=true
error: the server doesn't have a resource type "smm"
$ echo $?
1
```
That means it will never get to the lines below this and thus fail to clean up non-Maistra environments.
You should wrap this around IS_MAISTRA if-stmt so this only gets calls in Maistra environments.

jmazzitelli (changes requested):

jmazzitelli (approved):
i didn't run this, but it looks ok

PR #5530, opened 2022-10-04
[v1.57] Adding support for maistra and custom istio namespace

FilipB:
Adding support for maistra so the sidecars are enabled for all used demo applications. Also adding support for custom istio namespace.


jmazzitelli (changes requested):
fix the same things that I mention in the associated master PR.

jmazzitelli (approved):
i didn't run this, but it looks ok

PR #5532, opened 2022-10-04
show last message with same content

### LINKED ISSUE #5340: Istio Config editor - error reporting in messaging center
Given I open `Istio Config > Namespace: bookinfo > VirtualService > bookinfo`
And I have `0` notifications in Message Center 
image

When I change in YAML editor port nuber on non existent i.e. `90800` 
image

And I hit `Save`
Then I receive notification in UI (red banner)
image

And I receive event in Message Center
image

And on line no. 1 in YAML editor I can see propagated error
image

And I can see network event in dev console 
```error	"admission webhook \"validation.istio.io\" denied the request: configuration is invalid: port number 90800 must be in the range 1..65535"```

so far so good
now I change port back to `9080`
and create another error, i.e. deleting host in destination route (line no. 64): 
image

I hit save
I do receive notification in Messaging Center 
image
Error in YAML editor looks OK for now 
image
network tab showing only one error received, thats also OK
```error	"admission webhook \"validation.istio.io\" denied the request: configuration is invalid: empty domain name not allowed"``` 

**possible problems:**
1. I do not receive notification in UI (red popup, see screenshot above) on second try (I guess its because of 2)
2. The message in Message Center is same as before, second (different event) was aggregated wrongly under previous message see (2 events in 10 mins)
image
3. when 2 happens, context of last message (empty domain name not allowed) is lost in messaging center

### PR DESCRIPTION
**Pull request by leandroberetta**
Fixes #5340

### CODE REVIEW

**Review by israel-hdez — CHANGES_REQUESTED** [review]
Interesting use of `_.remove`. Sadly, lodash docs tell that it will mutate `group.messages`, which goes against the [rules of reducers](https://redux.js.org/tutorials/fundamentals/part-3-state-actions-reducers#rules-of-reducers):

>  They are not allowed to modify the existing `state`. Instead, they must make _immutable_ updates, by copying the existing `state` and making changes to the copied values.

**Comment by leandroberetta** [comment]
> Interesting use of `_.remove`. Sadly, lodash docs tell that it will mutate `group.messages`, which goes against the [rules of reducers](https://redux.js.org/tutorials/fundamentals/part-3-state-actions-reducers#rules-of-reducers):
> 
> > They are not allowed to modify the existing `state`. Instead, they must make _immutable_ updates, by copying the existing `state` and making changes to the copied values.

@israel-hdez well, honestly I didn't check that, mainly because I saw it is used in some of the lines I removed. So basically I just moved that line a little bit up.

So, discarding this PR, according to that we have an existing issue?.

**Comment by israel-hdez** [comment]
> So, discarding this PR, according to that we have an existing issue?.

Hmm... I didn't look to the old code. But now that I review the old code, yes it has the issue.
Let's fix it, if possible.

**Comment by leandroberetta** [comment]
> > So, discarding this PR, according to that we have an existing issue?.
> 
> Hmm... I didn't look to the old code. But now that I review the old code, yes it has the issue. Let's fix it, if possible.

@israel-hdez, right, I agree, I changed the remove method by the filter one, that does not mutate the array, just creates a new one and that is compatible with redux.

Thanks for noticing this.

File: frontend/src/reducers/MessageCenter.ts [hunk 1]
```diff
@@ -83,54 +83,28 @@ const Messages = (
...
+          newMessage = createMessage(
+            state.nextId,
+            content,
+            detail,
+            messageType,
+            1,
+            showNotification,
+            new Date(),
+            false,
+            undefined
```
israel-hdez, line 103: So, in contrast with the old code, this new revision is always using default values for some params, while the old code copied values from the old replica.

OK, let's look at the screen shot of the MessageCenter for the case when there is more than one replica:

image

If when invoking `createMessage` we always:
* pass `1` to the `count` arg, we will never see the `X in Y minutes` text, because it won't be considered as a replica.
* pass `undefined` to the `firstTriggered` arg, we would lose the date of the first replica and the `X in Y minutes` text may become `X in a few seconds ago minutes` if it happens to be visible.
* pass `false` instead of `existingMessage.showDetail` to the `showDetail` arg, we may end-up collapsing details of the message (they gray textarea) programatically.
* pass `showNotification` rather than `existingMessage.show_notification` to the `showNotification` arg ... well... this one apparently does not impact the old behavior.

That was the detailed technical explanation.

But in short, what I want to say is that this changes are fixing #5340, although we are losing the replica counter. It sounds negligible, so I open the question about if we want to lose the replica counter.

Also, there is the (also negligible) side effect that the for the replicated message, the user may see its details being collapsed automatically, which may be annoying for replicas triggered by the auto-refresh.

Both changes in behavior are negligible to me. So, I can approve if you are sure that you want them.

leandroberetta: @israel-hdez yes, I see what you mean and it's true, I prefer not losing that features. Thanks for the heads up. The message center is something I learning now and I've never noticed that.

So, I updated the code, honoring the count and the firstTriggered field now. Also, this makes the test to pass withouth any modification to them. I also tested locally, the last message is kept.

israel-hdez (approved):
LGTM!

Issue #5404, opened 2022-08-23
Kiali does not work well with Google Anthos Service Mesh


gidesh:
Hello team,

We are on Google's GKE Kubernetes cluster. And we are also using Google's Istio offering called Anthos Service Mesh. This is a managed offering of istio where the control plane is managed by Google.

We deployed Kiali through the Kiali operator and CRD and we have seen that when it's just a handful of workloads, it can show us the required information about graphs and other visualization.

However, when we added all our workloads into the mesh, Kali became unusable and we were not able to see any traffic visualization

We tried bumping the resources but that did not help either.

All we see in the kiali pods are the following warning

```
kiali-7c79cb774b-hc8jr kiali {"level":"warn","time":"2022-08-23T12:09:16Z","message":"GetPodProxyStatus is failing for [namespace: platform-metrics] [pod: vmstorage-jldp-14]: unable to find any healthy Pilot instance "}
kiali-7c79cb774b-4rn9h kiali {"level":"warn","time":"2022-08-23T12:09:16Z","message":"GetPodProxyStatus is failing for [namespace: mobile-apps] [pod: stubs-things-we-love-64b7d98866-945tl]: unable to find any healthy Pilot instance "}
kiali-7c79cb774b-hc8jr kiali {"level":"warn","time":"2022-08-23T12:09:16Z","message":"GetPodProxyStatus is failing for [namespace: enablement] [pod: stubs-greetings-ui-7f4fffd7f-rfvhl]: unable to find any healthy Pilot instance "}
kiali-7c79cb774b-4rn9h kiali {"level":"warn","time":"2022-08-23T12:09:16Z","message":"GetPodProxyStatus is failing for [namespace: platform-metrics] [pod: vmselect-jldp-7]: unable to find any healthy Pilot instance "}
kiali-7c79cb774b-4rn9h kiali {"level":"warn","time":"2022-08-23T12:09:16Z","message":"GetPodProxyStatus is failing for [namespace: mobile-apps] [pod: stubs-things-we-love-64b7d98866-vw82w]: unable to find any healthy Pilot instance "}
kiali-7c79cb774b-hc8jr kiali {"level":"warn","time":"2022-08-23T12:09:16Z","message":"GetPodProxyStatus is failing for [namespace: catalogue-feeds] [pod: stubs-interface-01p-collator-5969f45c7-4rw69]: unable to find any healthy Pilot instance "}
[...]

PR #5533, opened 2022-10-05
Connect to istio registry remotely

nrfox:
Adds a config option to connect to the istio registry (istiod debug endpoints) through an endpoint rather than the port forwarding into the istiod pods. This enables the istiod pods to live outside of the cluster where kiali is located while giving kiali a way to access the registry. It's assumed that this will be used with a proxy of some kind for now but additional options such as auth can be added in the future.

You can test this by adding a proxy pod to the istiod deployment to proxy requests to the istiod debug port. Note that this is just for testing purposes and not recommended. If the istiod debug endpoints are accessible by the kiali pod, the proxy would most likely be deployed in the kiali pod and would require: https://github.com/kiali/kiali-operator/pull/524.

Testing procedures. Assuming you have istio + kiali deployed:

1. Make the istio debug endpoints externally accessible by patching the istiod deployment to include an nginx proxy
```
kubectl patch deployments istiod -n istio-system --patch-file=/dev/stdin <<EOF
spec:
  template:
    spec:
      containers:
      - command:
        - /bin/bash
        - -c
        - |
          cat <<EOF > /etc/nginx/nginx.conf
          events { }
          http {
            server {
              listen 9240;
              location / {
                proxy_pass http://localhost:15014;
              }
            }
          }
          EOF
          nginx -g "daemon off;"
        image: nginx:stable
        name: nginx
EOF
```

2. Create a service for the proxy
```
apiVersion: v1
kind: Service
metadata:
  name: istiod-debug
  namespace: istio-system
spec:
  ports:
  - name: http
    port: 9240
    protocol: TCP
    targetPort: 9240
  selector:
    app: istiod 
  type: ClusterIP
```

3. Update kiali config to use that service
```
kubectl patch kialis kiali -n kiali-operator -p '{"spec": {"external_services": {"istio": {"registry": {"istiod_url": "http://istiod-debug.istio-system:9240"}}}}}' --type=merge
```

Relates to: #5404 and #4359


jshaughn (dismissed):
Code looks fine.  ran through test procedures and everything seems to be working in the UI that is querying istiod.

```
 deployment:
  ....
  external_services:
    istio:
      registry:
        istiod_url: http://istiod-debug.istio-system:9240
  istio_namespace: istio-system
  version: default
```

Xunzhuo (dismissed):
Thanks for adding this support! @nrfox

jshaughn:
Holding merge until next release to get soak time in master.  Can merge after release of v1.58.

nrfox:
@jshaughn @Xunzhuo added an e2e test for this if you wouldn't mind taking another look.

jshaughn (commented):
The new test failed in the latest integration run.

nrfox:
> The new test failed in the latest integration run.

CI doesn't use the operator so I needed to add support for updating the kiali configmap directly.

jshaughn (commented):
@nrfox Tests ran through with your change from [late] last night.  I am re-running the jobs now as a sanity check.  If all runs through I think it's good to go.

jshaughn (approved):
Sanity check run of tests passed, removing DNM label, looks good!

jmazzitelli:
did this follow the checklist: https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/defaults/main.yml#L157-L186

I don't see this new Registry option declared in the operator.

PR #5541, opened 2022-10-06
How Kiali Identifies Ambient

### LINKED ISSUE #5535: How Kiali identifies Ambient
How Kiali can identify if Istio is running in ambient mode. 

- It is a istio-proxy ambient pod? It has the vars ISTIO_META_AMBIENT_TYPE = ztunnel
- If a pod is part of an ambient mesh, it should be labeled with ambient-type=Component (Ex. workload, etc)

josunect:
This is a proposal to show a L4 label if the pod is labeled with the Ambient annotations. 

image

- Istio checks if the pod has an Istio Container running, but this is not a valid check for Ambient. 
- Istio Ambient checks if the pod has an Ambient label (ambient-type=Component)

In the UI: 
- Now it doesn't show the "MIssing sidecar" if the pod has istio sidecar OR if the pod is labeled with istio-ambient label
- It shows an L4 badge if is Istio Ambient. 

Fixes #5535


josunect:
I should be replaced in details pages: 

image

josunect:
Indicate Istio Ambient global state: 

image

josunect:
- Add it Ambient label in the CP card:
image

- Show "Not in the mesh" warning: 
image

- Show mode in details (ztunnel as example): 
image

jshaughn (commented):
Looking good, I like the additional badge to signal `Ambient` at the control-plane card.  The workload mode badging is also good, I think, if it can indicate whether its handled by z-tunnel, waypoint, sidewcar, etc.  Only minor comment is that the badge styling looks too different.  I'd maybe suggest that all of these "mesh metadata" use the same styling as we already use for that `control plane` badge, so that a user can get used to knowing that that styling indicates mesh infrastructure info.

josunect:
I've done a change in the styling: 

image

ScriptingShrimp:
HI, I' adding label test-add-coverage and will request some more tests bundled into this PR. From what I understand, there are only updated snapshots of components. 

I think this PR is really big and will require some extensive testing in Cypress due to visual changes in Kiali UI. 

Not sure about how big changes are in API (business folder), but my assumption is that we will need integration tests coverage as well. 

Please open discussion about test coverage, thank you :+1:

josunect:
> HI, I' adding label test-add-coverage and will request some more tests bundled into this PR. From what I understand, there are only updated snapshots of components.
> 
> I think this PR is really big and will require some extensive testing in Cypress due to visual changes in Kiali UI.
> 
> Not sure about how big changes are in API (business folder), but my assumption is that we will need integration tests coverage as well.
> 
> Please open discussion about test coverage, thank you +1

Thanks for the update, @ScriptingShrimp 

At the moment, this is just a draft PR because Istio Ambient was not released yet and it was tested in an early stage. This is just a first approximation about how it could look like but I think it is too early to define the tests. But, it would require extensive testing, I think. 

I think next steps here, will be to test a more recent version of Ambient and rebase this branch with master.

jshaughn:
> At the moment, this is just a draft PR because Istio Ambient was not released yet and it was tested in an early stage. This is just a first approximation about how it could look like but I think it is too early to define the tests. But, it would require extensive testing, I think.
> I think next steps here, will be to test a more recent version of Ambient and rebase this branch with master.

@ScriptingShrimp As mentioned above, it's probably best to ignore Draft PRs.  They are really just a developer playground.  We will probably hold on much dedicated work for Ambient until it matures a bit more and maybe after we have more multi-cluster  UX changes in place, as that work will likely impact UI.  In both cases we will try to avoid giant PRs and favor smaller iteration, with necessary testing included.

One more thing about Ambient, I think in the beginning we were thinking of it as Istio [classic] OR Istio Ambient, like 2 different things.   But actually, I think it's looking more like Ambient is a capability being built into Istio.  Almost an "Interface" that says Istio supports HBONE interoperbility and ztunnel/Waypoint-based sidecarless deployment.  Istio's Ambient support will get baked in, people can use it or feasibly make no changes at all to what they have currently.  So I'm not sure it even makes sense to flag "Ambient" in Kiali, but where necessary make it explicit what's being used (sidecar, waypoint, z-tunnel only, etc).

josunect:
Closed for https://github.com/kiali/kiali/pull/5952

PR #5554, opened 2022-10-12
Remove "isMaistra" conditional code regarding workload-level autoInjection

jshaughn:
Remove "isMaistra" conditional code regarding workload-level label/annotation support for autoInjection.  Maistra 2.3 is still opt-in, but supports the same as upstream (both label and annotation).

https://issues.redhat.com/browse/OSSM-2105


nrfox (approved):
LGTM. Perhaps the cypress test could update both the label and annotation but that's not a blocker.

jshaughn:
I didn't want to play with the test in order to ensure the same test now passes on OSSM.  But yes, we could make that change going forward.

PR #5562, opened 2022-10-14
Bump Go version to 1.19.1

### LINKED ISSUE #5561: Bump go version to 1.18 (or 1.19?)

leandroberetta:
Closes #5561

### DIFF
--- a/.github/workflows/build-backend.yml
+++ b/.github/workflows/build-backend.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.17.10
+          go-version: 1.19.1
 
       - name: Cache Go deps
         uses: actions/cache@v3
--- a/.github/workflows/integration-tests-backend.yml
+++ b/.github/workflows/integration-tests-backend.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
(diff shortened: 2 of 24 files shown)


aljesusg:
Related with https://github.com/kiali/kiali/pull/5351

Note:

I wonder if:

Keep the latest go 1.17 to be aligned with Istio 1.14.x https://github.com/istio/istio/blob/1.14.3/go.mod (if the issue is addressed there).
Bump to go 1.18 when Kiali bumps to Istio 1.15 https://github.com/istio/istio/blob/master/go.mod#L3
In theory, it should work but using the same version of the dependencies https://github.com/istio/client-go/blob/v1.14.3/go.mod#L3 always helps to avoid potential issues.

But if the testing is solid, I imagine all would be good.

I don't know implications for 1.19 but again I need to check downstream @leandroberetta

jshaughn:
Well this is a tricky question when upstream and downstream build with the same Go version.  1.57 is already targeted to 1.15 upstream.  1.59 will be targeted to 1.16 upstream.  Yet 1.57 will be used with 1.14 downstream.  We don't have to backport to previous versions, but we have already had requests to bump Go in the upstream.

I think we should proceed with v1.18 for our next release and can hold on any backports.

leandroberetta:
Ok, so I will close this issue to keep the following issue active: https://github.com/kiali/kiali/pull/5351 (bumping to v1.18).

PR #5563, opened 2022-10-14
[Cypress] Fix sidecar injection cypress tests that assumed only label

nrfox:
https://issues.redhat.com/browse/OSSM-2105

Clears the annotation if it's present and only uses the label. Adds a note for platform specific tags.


jshaughn (approved):
This looks fine.  If and when we have conditional testing for OSSM I think all tests just need to run against both.

PR #5566, opened 2022-10-17
Hide memory or CPU charts if there is no values to show

leandroberetta:
For example, when the memory series has no values: 

image


jshaughn (approved):
LGTM

PR #5573, opened 2022-10-18
[v1.57] Ignore cluster info checking, as Kiali ignores it in case of any problem

hhovsepy:
https://issues.redhat.com/browse/OSSM-2138

Not critical check of Cluster name is removed.
On OCP clusters retrieving Cluster ID from istiod pod fails and is ignored on Kiali side.
https://github.com/kiali/kiali/blob/master/handlers/config.go#L106

### DIFF
--- a/tests/integration/tests/kiali_urls_test.go
+++ b/tests/integration/tests/kiali_urls_test.go
@@ -26,8 +26,6 @@ func TestKialiConfig(t *testing.T) {
 	assert.Nil(err)
 	assert.Equal(200, statusCode)
 	assert.NotEmpty(response)
-	assert.NotEmpty(response.ClusterInfo)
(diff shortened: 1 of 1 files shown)


jshaughn:
@hhovsepy Nothing wrong with this PR but setting to DNM because we are not currently making changes to the 1.57 release branch.  The failure does not have to be resolved until OSSM 2.3.1.  Having said that, if we need another build for 2.3.0 then we may consider adding this one as well.

hhovsepy:
Yeah sure @jshaughn , just to note that this change is test only.

FilipB:
@jshaughn If possible we would like to have this merged as this is test change only and it will make 2.3 testing much easier when having green tests...

jshaughn (approved):
LGTM, merging due to QE request

PR #5575, opened 2022-10-18
An attempt to fix timing issue resulting in missing side car

FilipB:
In some environments some bookinfo pods are started without a side car even though everything is configured correctly. This is an attempt to fix this behavior so the pods are not patched before smmr is Ready.

### DIFF
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -32,6 +32,8 @@ spec:
 EOM
 
   if [ "${AUTO_INJECTION}" == "true" ]; then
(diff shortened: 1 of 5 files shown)


FilipB:
This change is already in v1.57 and tested. It resolves the missing side car problem.

jmazzitelli (dismissed):

hhovsepy:
It seems failure is not related, but strangely it indicates that newly created service is not recognized yet by Kiali.

FilipB:
Updated also other install scripts

jmazzitelli (approved):

PR #5579, opened 2022-10-21
Preview shows old state in Istio Config Wizard Preview

### LINKED ISSUE #5546: Config Create Wizard - Preview shows old state
Happens on Kiali master.

In any Istio Config object creation Wizard, for instance new Gateway.
When adding several ports and then removing one port.
Click on Preview.
It shows all ports in YAML content, including the removed one. Here is the bug.
When cancel the Preview and then again Preview, it shown correctly the YAML content.
Screenshot from 2022-10-07 14-21-44
Screenshot from 2022-10-07 14-21-29

josunect:
Add an additional state item so IstioConfigPreview can control when the modal is Open. 

Fix #5546


hhovsepy (approved):
Verified, works correctly now.

PR #5583, opened 2022-10-25
Add min TLS Version in CP Card

### LINKED ISSUE #5528: Add TLS min version to the Control Plane card in the Overview Page
According to [this](https://istio.io/latest/docs/tasks/security/tls-configuration/workload-min-tls-version/#:~:text=This%20task%20shows%20how%20to,for%20Istio%20workloads%20is%201.3.) documentation, we can indicate the minimal tls version in the Control Plane card.

Also, we can add the certificates information in the card as well. Today, certificates information are in the dropdown menu near the help section. 

A proposal could be to add a little icon related to security and then display a tooltip with all these information.

Epic https://github.com/kiali/kiali/issues/4435

josunect:
Fixes #5528 

image

When this is not set: 

image

Some notes: 

- This should be shown when this.state.displayMode === OverviewDisplayMode.EXPAND? 
- I've added a new component into components/Overview
- Tested with "--set values.meshConfig.meshMTLS.minProtocolVersion=TLSV1_3" : Should we set up this as a default in the hack script install-istio-via-istioctl?

### DIFF
--- a/business/istio_certs.go
+++ b/business/istio_certs.go
@@ -26,6 +26,10 @@ type mtlsMinV struct {
 	MTLSMinVersion string `yaml:"minProtocolVersion"`
 }
 
+type meshMTLS struct {
+	MeshMTLS mtlsMinV `yaml:"meshMTLS"`
+}
+
 type istioConfig struct {
 	Certificates []certConfig `yaml:"certificates"`
 }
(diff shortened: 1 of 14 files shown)


leandroberetta:
> * This should be shown when this.state.displayMode === OverviewDisplayMode.EXPAND?

I think so, yes, this section is only in this view (outbound policy, proxy push time and tls min version).

What do you think on adding the same style than for example, outbound policy mode, that has the value inside a label?. The proxy push time has a similar design too.

josunect:
> > * This should be shown when this.state.displayMode === OverviewDisplayMode.EXPAND?
> 
> I think so, yes, this section is only in this view (outbound policy, proxy push time and tls min version).
> 
> What do you think on adding the same style than for example, outbound policy mode, that has the value inside a label?. The proxy push time has a similar design too.

Thanks for the revision, @leandroberetta! I'll send the changes.

leandroberetta (approved):
LGTM.

PR #5585, opened 2022-10-26
Adding support for IMB's p/z arch to bookinfo install script

FilipB:
This is required by IBM p/z teams as there are no official ppc64le and s390x bookinfo images. This change allows to specify architecture to use matching bookinfo images.


FilipB:
Thank you for checking.

PR #5586, opened 2022-10-26
Adding support for IMB's p/z arch to error rates install script

FilipB:
This is required by IBM p/z teams as there are no official ppc64le and s390x error rates images. This change allows to specify architecture to use matching error rates images. Also updating install-testing-demos.sh to work with new arch.


FilipB:
I tested default arch and ppc64le

jmazzitelli (approved):
I didn't run this but the code looks like the other changes and seems ok. approving.

PR #5589, opened 2022-10-28
[1.57] Cherrypick of additional state item preview fix

nrfox:


israel-hdez (approved):
Same code changes as in #5579. I verified that it fixes the issue.

LG!

Issue #5591, opened 2022-10-28
mount secret data from files, not environment variables


jmazzitelli:
[Here the k8s docs](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/) say:
```
This page shows how to securely inject sensitive data, such as passwords and encryption keys, into Pods.
```
Oh, goodie. I need a way to securely inject sensitive data. "How can I do this?", I ask. [A section on that k8s docs page](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-container-environment-variables-using-secret-data) says:
```
Define container environment variables using Secret data 
```
Perfect. I'll do that. So the Kiali pod has secrets "securely" injected via environment variables.

Er... not so fast. [This doc](https://www.stigviewer.com/stig/kubernetes/2021-04-14/finding/V-242415) says:
```
Secrets in Kubernetes must not be stored as environment variables.

Secrets, such as passwords, keys, tokens, and certificates should not be stored
as environment variables. ... Secrets must be mounted from files ...
```

Some users are asking that Kiali not inject secret data via env vars for NIST compliance. This enhancement request is to change the Kiali implementation so secrets will be "mounted from files" instead (See [k8s docs here](https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod))

jmazzitelli:
server PR: https://github.com/kiali/kiali/pull/5592
operator PR: https://github.com/kiali/kiali-operator/pull/585

jmazzitelli:
1.57 cherry pick PRs:
* https://github.com/kiali/kiali/pull/5604
* https://github.com/kiali/kiali-operator/pull/587

PR #5592, opened 2022-10-28
secrets now injected into pod file system, not as env vars

jmazzitelli:
part of https://github.com/kiali/kiali/issues/5591
server PR: https://github.com/kiali/kiali-operator/pull/585

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -77,6 +77,9 @@ func (fn FeatureName) IsValid() error {
 var configuration Config
 var rwMutex sync.RWMutex
 
+// Defines where the files are located that contain the secrets content
+var overrideSecretsDir = "/kiali-override-secrets"
+
 // Metrics provides metrics configuration for the Kiali server.
 type Metrics struct {
 	Enabled bool `yaml:"enabled,omitempty"`
@@ -911,7 +914,7 @@ func Unmarshal(yamlString string) (conf *Config, err error) {
 	}
 
 	for _, override := range overrides {
-		fullFileName := "/kiali-override-secrets/" + override.fileName + "/value.txt"
+		fullFileName := overrideSecretsDir + "/" + override.fileName + "/value.txt"
 		b, err := os.ReadFile(fullFileName)
 		if err == nil {
 			fileContents := string(b)
--- a/config/config.go
+++ b/config/config.go
@@ -2,6 +2,7 @@ package config
 
 import (
 	"embed"
+	"errors"
 	"fmt"
 	"io/fs"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


jmazzitelli:
How to test. Make sure you have both this PR and the operator PR checked out locally.

1. Install minikube and install Istio in it (`hack/k8s-minikube.sh start && hack/k8s-minikube.sh istio`)
2. Install the operator and kiali server using make (`make -e CLUSTER_TYPE="minikube" build cluster-push operator-create kiali-create`)
3. Check that you do not have any environment variables associated with the signing key secret. This command should return environment variables, but none related to the login token signing key:
```
kubectl get pods -n istio-system -l app.kubernetes.io/name=kiali -o jsonpath='{.items[0].spec.containers[0].env}' | jq
```
4. Check that you have volumeMount/volume for the signing key:

```
kubectl get pods -n istio-system -l app.kubernetes.io/name=kiali -o jsonpath='{.items[0].spec.containers[0].volumeMounts'} | jq ;
kubectl get pods -n istio-system -l app.kubernetes.io/name=kiali -o jsonpath='{.items[0].spec.volumes'} | jq
```

Confirm you see this in the results:

volumeMount:
```json
  {
    "mountPath": "/kiali-override-secrets/login-token-signing-key",
    "name": "login-token-signing-key",
    "readOnly": true
  },
```

and volume:

```json
  {
    "name": "login-token-signing-key",
    "secret": {
      "defaultMode": 420,
      "items": [
        {
          "key": "key",
          "path": "value.txt"
        }
      ],
      "optional": false,
      "secretName": "kiali-signing-key"
    }
  },
```

5. Create a test secret:
```
kubectl create secret generic test-secret --from-literal=username=testuser --from-literal=password=testpw --from-literal=token=testtoken -n istio-system
```
6.

config/config.go
```diff
@@ -17,18 +18,18 @@ import (
 	"github.com/kiali/kiali/log"
 )
 
-// Environment variables that can override the ConfigMap yaml values
+// Files found in /kiali-override-secrets that override the ConfigMap yaml values
```
nrfox, line 21: We don't want to support both? Only reading these from a file?

config/config_test.go
```diff
@@ -23,33 +23,14 @@ func TestEnvVarOverrides(t *testing.T) {
...
-	conf, _ = Unmarshal(yamlString)
-
-	// env vars are now set- values should be overridden
-	assert.Equal(t, conf.ExternalServices.Grafana.Auth.Password, "grafanapasswordENV")
-	assert.Equal(t, conf.ExternalServices.Grafana.Auth.Token, "grafanatokenENV")
-	assert.Equal(t, conf.ExternalServices.Prometheus.Auth.Password, "prometheuspasswordENV")
-	assert.Equal(t, conf.ExternalServices.Prometheus.Auth.Token, "prometheustokenENV")
-	assert.Equal(t, conf.ExternalServices.Tracing.Auth.Password, "tracingpasswordENV")
-	assert.Equal(t, conf.ExternalServices.Tracing.Auth.Token, "tracingtokenENV")
-	assert.Equal(t, conf.LoginToken.SigningKey, "signingkeyENV")
```
nrfox, line 52: This test is not testing anything anymore. If you want to keep coverage, you can make the base file path internally configurable https://github.com/kiali/kiali/pull/5592/files#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R914 then create a temp testing folder/files with some dummy data in it.

jmazzitelli: I thought about that - but thought it would be adding complexity for the sake of unit testing. Do we do this kind of "internal configuration of file paths for unit tests" anywhere else? I tried to see if there is some Go magic to mock out a file system for tests, but didn't find anything easily consumable.

nrfox: > I thought about that - but thought it would be adding complexity for the sake of unit testing.

This is true but IMO it's worth it. The complexity is small. You'd basically just be putting `/kiali-override-secrets` into a var like `defaultSecretsBasePath` or something and then setting that in the unit test. But another option is to just remove the test which is better than having a test that no longer tests what it says it does.

> I tried to see if there is some Go magic to mock out a file system for tests, but didn't find anything easily consumable.

[t.TempDir()](https://pkg.go.dev/testing#T.TempDir) is what you want. I know we do this elsewhere in the tests but I'd need to track it down.

jmazzitelli: I'll work on this now.

jmazzitelli: @nrfox done. unit tests updated and passing.

nrfox (approved):
LGTM

Issue #5584, opened 2022-10-26
Kiali dashboard freezes when checking traces information


Myeongseok-Kim:
After enabling Jaeger, I confirmed that I can check traces information in kiali.
![image](https://user-images.githubusercontent.com/54100684/197927234-091e8510-70eb-4566-a97e-7356f80e782b.png)

But when I try to check traces information
Something seems to be happening in the Kiali dashboard.
(Loading animation plays)
![image](https://user-images.githubusercontent.com/54100684/197927173-54b6b428-8ef8-485a-8ef8-df42557990b3.png)
This puts a load on my prometheus server, and eventually the Prometheus server went down due to OOM.
This issue is still occurring.
Does it only happen in my environment?

**Version information**
istio: 1.15.2
kiali: 1.58.0
jaeger: 1.38.0
prometheus: 2.37.0

jshaughn:
Hi @Myeongseok-Kim,  a few questions:
- Is the problem repeatable?
- When you say, `But when I try to check traces information` do you mean that you clicked on one of the traces in the chart? Or was it some other action, like moving the mouse over the chart to look at tooltip information? 
- Did you click the "Distributed Tracing" option in the main menu, or are you on the "Traces" tab of a workload detail? 
- Does the rest of Kiali function well, for example, can you generate a graph with no problem?
- Is there any reason to think the issue may be with you prometheus pod resource limits or connectivity?

jshaughn:
@Myeongseok-Kim I think maybe I see a problem but let me ask, after the traces chart is displayed, did you at any point drag the pointer across a large number of trace nodes, or look at many of the tooltips?  Also, did this problem happen quickly or had you been interacting with the chart over a longer period of time?

Myeongseok-Kim:
> Hi @Myeongseok-Kim, a few questions:
> 
> * Is the problem repeatable?
> * When you say, `But when I try to check traces information` do you mean that you clicked on one of the traces in the chart? Or was it some other action, like moving the mouse over the chart to look at tooltip information?
> * Did you click the "Distributed Tracing" option in the main menu, or are you on the "Traces" tab of a workload detail?
> * Does the rest of Kiali function well, for example, can you generate a graph with no problem?
> * Is there any reason to think the issue may be with you prometheus pod resource limits or connectivity?


- It is possible to reproduce the problem.
- Occurs when the mouse hovers over a point on the tracking chart.
(This simply happens without any other action.)
[...]

PR #5600, opened 2022-11-01
Prevent flooding backend with concurrent calls as the user drags across trace chart

jshaughn:
Trace scatter-chart tooltips pull stats from the backend to show heatmaps. Prevent flooding the backend with concurrent calls as the user drags across the chart:
- move launch of the backend request from the N tooltip components to the single JaegerScatter component, to be able to throttle the api calls. Add some callback props for this support.
- ensure the thunk action returns a promise and use it to preven the concurrent calls.
- ignore making calls for points that are just moused-over while moving across the chart.
- also, remove an unnecessary deprecated arg in Dashboard.tsx

Fixes http://github.com/kiali/kiali/issues/5584

### DIFF
--- a/frontend/src/actions/MetricsStatsThunkActions.ts
+++ b/frontend/src/actions/MetricsStatsThunkActions.ts
@@ -1,6 +1,6 @@
 import { KialiAppState } from '../store/Store';
 import * as API from '../services/Api';
-import { KialiDispatch } from "../types/Redux";
+import { KialiDispatch } from '../types/Redux';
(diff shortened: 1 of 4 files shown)


jshaughn:
This network activity waterfall shows the change clearly when dragging across a Trace scatter chart like this one:

image

Before the change you can see the concurrent calls launching, and the round-trip times being longer as things get loaded on the prometheus end of things.  After the change you can see the sequential calls, less in number and shorter round-trips.

Before:
image

After:
image

For me the tooltip response time was fine.

nrfox (approved):

Issue #5594, opened 2022-10-31
Filter "istio" GatewayClassName Gateway API gateways


hhovsepy:
Currently Kiali shows all Gateway API gateways, but it needs to filter and show only mesh related Gateways, which has "gatewayClassName: istio".

PR #5603, opened 2022-11-02
Filter K8sGateways by GatewayClass Names, which by default is 'istio'

hhovsepy:
Issue https://github.com/kiali/kiali/issues/5594

Filtering K8s API Gateways to show only those ones which are pointing to GatewayClass created by Istio.

Using 'istio' as default GatewayClass Name to filter K8sGatgeways from Gateway API.

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -176,7 +176,7 @@ func (in *IstioConfigService) GetIstioConfigList(ctx context.Context, criteria I
 			istioConfigList.Gateways = kubernetes.FilterSupportedGateways(registryConfiguration.Gateways)
 		}
 		if criteria.Include(kubernetes.K8sGateways) {
-			istioConfigList.K8sGateways = registryConfiguration.K8sGateways
+			istioConfigList.K8sGateways = kubernetes.FilterSupportedK8sGateways(registryConfiguration.K8sGateways)
 		}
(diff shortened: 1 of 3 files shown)


josunect (commented):
So, if I understood correctly, when we search for a istio name, if there is a match with the GW api class name - Istio in this case - We will have as a result the auto generated istio gateways? 

image

hhovsepy:
It seems I did not describe correctly.
When creating a Gateway, it requires gatewayClassName to be specified, https://istio.io/latest/docs/tasks/traffic-management/ingress/gateway-api/#configuring-a-gateway.
And only 'istio' GetewayClass belongs to Istio and Kiali. 
There can be other GatewayClasses than 'istio' ones and Gateways configured to that classes in cluster, which should be ignored.
This fix is to ignore any other Gateway which is not configured to 'istio' GatewayClass, or 'istio' name is set as configurabe so for multi-cluster muti-tenant  or OSSM cases they can have different values that default hardcoded 'istio'.

josunect:
> 

Thanks for the explanation, @hhovsepy ! Sorry I have had not this clear, now I was able to tests this feature. 

The first gateway was created with a different class name:

In master: 
image

With this PR, the gateway is no longer available: 
image

Code LGTM.

josunect (approved):

PR #5616, opened 2022-11-08
Show Kiali configuration 

### LINKED ISSUE #5588: Show Kiali configuration in the application
The motivation of this is to have a section to display the Kiali config map in a nice and simple way.

It's ok to see the configmap using CLI but it should be easier to see it in Kiali directly.

josunect:
The changes include, update the debug information Modal: 

image

And show the kiali configuration: 

image

This is just a draft as a proposal about how to show this information. 

I wasn't sure about the information to be shown. I've thought in having a list of headers to filter items from the already available config. Another possibility is to create an specific API returning just the specific information (Get the spec from the kiali CR?). Depending on the information, add the security checks. 

Fixes #5588

### DIFF
--- a/frontend/src/components/DebugInformation/KialiConfiguration.tsx
+++ b/frontend/src/components/DebugInformation/KialiConfiguration.tsx
@@ -50,7 +50,7 @@ const copyToClipboardOptions = {
   message: 'We failed to automatically copy the text, please use: #{key}, Enter\t'
 };
 
-export class DebugInformation extends React.PureComponent<DebugInformationProps, DebugInformationState> {
+export class KialiConfiguration extends React.PureComponent<DebugInformationProps, DebugInformationState> {
   aceEditorRef: React.RefObject<AceEditor>;
 
   constructor(props: DebugInformationProps) {
@@ -135,7 +135,7 @@ export class DebugInformation extends React.PureComponent<DebugInformationProps,
         variant={ModalVariant.small}
         isOpen={this.state.show}
         onClose={this.close}
-        title="Debug information"
+        title="Kiali Configuration"
         actions={[
           <Button onClick={this.close}>Close</Button>,
           <CopyToClipboard onCopy={this.copyCallback} text={renderDebugInformation()} options={copyToClipboardOptions}>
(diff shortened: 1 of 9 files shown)


jmazzitelli:
> Another possibility is to create an specific API returning just the specific information (Get the spec from the kiali CR?).

I don't think we need to spit out the entire Kial CR spec. The user can get that very easily via simple (`kubectl get kiali kiali -o jsonpath='{spec}'`).

Perhaps we just retrieve some common Kiali CR settings that people would normally care about? I don't know which ones, just mentioning this as a possible way to do it.

jshaughn:
>Perhaps we just retrieve some common Kiali CR settings that people would normally care about? I don't know which ones, just mentioning this as a possible way to do it.

Which `people` are we talking about?  I think the people may be us, as this could be provided back with bug reports.  If you don't know which parts are the ones, `people would normally care about` then I don't think anyone knows.  So maybe best to provide everything that isn't a security risk.

josunect:
So, this would be the information that is already returned by api/config?  
https://github.com/kiali/kiali/blob/master/handlers/config.go#L70

This is meant to return the public config. It gets part of the config - That defines de full YAML https://github.com/kiali/kiali/blob/master/config/config.go - and enriches it with additional information. 
It discards: 

- AdditionalDisplayDetails
- API
- Auth
- CustomDashboards
- Deployment: Just viewOnly mode
- Identity
- InCluster
- KubernetesConfig
- LoginToken
- Server

And it adds: 

- Clusters
- Prometheus 
- GatewayAPIEnabled

So maybe, we can enrich the publicConfig a bit, like: 
- AccesibleNamespaces
- AuthStrategy
- LogLevel

config contains: 

image

jshaughn:
It looks like we stalled on this effort.  After reading the comments, I suggest we move forward in the following way:
- No new `?` menu entries, just keep `View Debug Info`.
- Two sections for that screen, like today, but change:
  - `Health Config` -> `Kiali Config`
     - This section will show formatted `Spec` yaml, stripping any security-sensitive information.
  - `Please include this information when opening a bug.` -> `Additional Kiali State`
     - This section will show selected `State` information (just like today), like the latest graph config, as formatted yaml. It should not duplicate information in the `Kiali Config` section.

josunect:
Created https://github.com/kiali/kiali/pull/5680

Issue #5197, opened 2022-06-13
[CI] Test Flake - TestCreateSessionNoChunks


nrfox:
See: https://github.com/kiali/kiali/runs/6861149294?check_suite_focus=true for an example

>  --- FAIL: TestCreateSessionNoChunks (0.02s)
    session_persistor_test.go:58: 
        	Error Trace:	session_persistor_test.go:58
        	Error:      	"5WVbSMsyrn1SABmGm+k3P0t+PHlCDPfxt6SbVmNlyH9BXcjPZKZfOn1v/cxr1PHFoozyLgAteMxb9QTq8DXeCT8hI6vR3Q2hnEBjr1KQjGg86obOQ/R65/i3oD/aer5tqsNCk4nFCWslzYkaT37RayIJ5DCknbo=" should not contain "Foo"
        	Test:       	TestCreateSessionNoChunks

amantayal5:
Hi @nrfox 
I want to work on this issue! Can you explain the bug to me?

nrfox:
@amantayal5 sure! From here: https://github.com/kiali/kiali/blob/master/business/authentication/session_persistor_test.go#L53-L59, it looks like the test is checking that the cookie doesn't contain "Foo" to ensure that text actually got encrypted however you can randomly get "Foo" as part of the encrypted text hence the test flake/failure here. @israel-hdez can confirm but perhaps you can just update the assertion from an `assert.NotContains` to an `assert.NotEquals`?

israel-hdez:
It is just a little bit more complicated that just doing `assert.NotEquals`. In case it happens to be plain text, it would be a string of a JSON document. Something like `{ "firstField": "Foo" }`. I was trying to avoid that "dance" of parsing the JSON string. Perhaps if parsing, succeeds may that would be enough to conclude it is in plain text and finish with failure? (i.e. no need to check the contents of the json doc?)

jshaughn:
@amantayal5, are you working this issue?

amantayal5:
@jshaughn Not exactly! I am unable to do setup Kiali in my system right now!

nrfox:
> @nrfox CI seems stable at the moment, did this go away?

No I think it's just a relatively rare flake since the encrypted text is unpredictable and has to contain `Foo` in order to trigger a failure.

jshaughn:
@matejnesuta , would you be able to take this one?  It's not urgent but I'd like to see it get cleaned up.

[...]

PR #5617, opened 2022-11-10
Fixing flaky session test.

hhovsepy:
Fix for https://github.com/kiali/kiali/issues/5197

Instead of checking Contains cookie value, now it checks equals cookie value + checks contains the cookie json.


israel-hdez (approved):
I cannot think of a better approach. So, LGTM.

PR #5625, opened 2022-11-11
Added smoke test for the new Overview page brought in 1.57.

matejnesuta:
One of the tests for the new features added in the OSSM 2.3 version of Kiali

The test checks the new Control plane for Istio in the overview page. It also checks some additional links in its kebab menu, which are not available in the other namespaces.

### DIFF
--- a/frontend/cypress/integration/common/overview.ts
+++ b/frontend/cypress/integration/common/overview.ts
@@ -1,4 +1,5 @@
-import { Before, Given, Then, When } from '@badeball/cypress-cucumber-preprocessor';
+import { Before, Given, Then, When, And } from '@badeball/cypress-cucumber-preprocessor';
+import { ensureKialiFinishedLoading } from './transition';
 
 Before(() => {
   // Focing to not stop cypress on unexpected errors not related to the tests.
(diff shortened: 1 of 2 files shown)


ScriptingShrimp:
Hello @matejnesuta , I tried to reproduce CI error, this is what I got

from CI
```
1 failing

  1) Kiali Overview page
       The Istio panel should be visible in the control panel:
     AssertionError: Timed out retrying after 40000ms: Expected to find content: 'Istio Mesh Dashboard' within the element: <a> but never did.
      at Context.eval (http://172.18.255.71:20001/__cypress/tests?p=cypress/integration/featureFiles/overview.feature:1786:32)
```

from my local OCP cluster with upstream Istio and demo apps installed :
```
36
containsIstio Mesh Dashboard
0
AssertionError
Timed out retrying after 40000ms: Expected to find content: 'Istio Mesh Dashboard' within the element: <a> but never did.
cypress/integration/common/overview.ts:227:28
  225 |     }
  226 |     cy.wait(1000);
> 227 |     cy.wrap($el).find('a').contains('Istio Mesh Dashboard');
```
assertion error is the same, its not in CI

leandroberetta:
> On second try, I was able to reproduce different results on upstream OCP (jenkins) and GH actions.
> 
> Question is: what are the differences in upstream environment (KIND) and downstream (OCP) setup? Looks like different names of control plane. Can someone elaborate on this please? cc @nrfox @josunect pray
> 
> Problem with this test is that green badge "Control Plane" (next to istio-system) have different name in different environments. image

@ScriptingShrimp it should always display "Control plane" (notice the lower case "plane"), originally it was "Control Plane", then we moved it to "Control plane".

Is this the difference you're noticing in the values? or, what is the other value you're seeing there?. It's strange because it's a hardcoded name in the code. I will check if it is duplicate somehow with different values but I've never seen differente values there.

matejnesuta:
> 

Just installed a new Kiali instance on OCP cluster and the label is exactly as you are describing. I think this is Pavel's issue regarding his environment.

matejnesuta:
Finally it passed. Thank God.

ScriptingShrimp (approved):
Tested on OCP, all passing. LGTM :+1:

PR #5628, opened 2022-11-14
compat matrix for istio 1.16

jmazzitelli:
need to backport this to v1.59 and build a new v1.59.1 so we can support Istio 1.16.1 when it is release (we missed the train for istio 1.16.0).


jshaughn (approved):
LOK

Issue #5502, opened 2022-09-27
Gateway API Objects - Include in wizards


hhovsepy:
The Gateway API objects (HTTPRoute, Gateway) needs to be included in:

- [x] #5578
- [x] Istio Config create Gateway
- [x] Traffic shifting support between multiple versions
- [x] Route Filter
- [x] RequestMirror
- [x] RequestRedirect

PR #5630, opened 2022-11-14
Traffic Shifting in K8s Routing Wizard

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5502

K8s Routing wizard supports Traffic Shifting.
In an contrary with regular Istio HTTP routing, the K8s Gateway API HTTP Routing traffic chifting overall weights sum should not be 100. https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.HTTPBackendRef

For testing: you need to have multiple versions of a service, like here: https://github.com/istio/istio/blob/master/samples/bookinfo/platform/kube/bookinfo-versions.yaml

Screenshot from 2022-11-24 12-40-36

Screenshot from 2022-11-24 12-40-51


hhovsepy:
Here one thing remain to complete: In a case of several versions of Service (multiple workloads), the K8s Gateway API HTTPRoute requires a K8s Service per version. So those multiple version services should be created in advance and loaded in Traffic Wizard for traffic shifting between them.

josunect (commented):
Great work, @Hayk! 

I've tested the view in several services, and it works as expected. 
I would add a couple of comments: 

- If weight is not a percentage (From the docs) - Maybe the % can be removed? 
image
- When You click in "Add Route Rute" Is taking the information specified in both "Request Matching" and "Route to" tabs. It woulnd't make more sense if they are both in the same tab? When I see both objects in different tabs it looks like there are not so closely related.

hhovsepy:
Removed the '%' from slider @josunect , thanks for the review. The second point in your comment: it is by the design of wizards itself, and also more tabs will be added later, so I would leave it as it is now because that will cause whole wizard redesign.

hhovsepy:
Thank you @josunect , all the comments were valid and approached. The "showMirror" feature will be added later.

josunect (approved):
Thanks, @hhovsepy !
LGTM

Issue #5611, opened 2022-11-07
Kiali SA should use view-only role unless using anonymous strategy


technotaff-nbs:
## Feature Request
Currently the following setting in the kiali-server helm chart:

```
deployment:
  view_only_mode: true
```
performs the following actions:
1. Caused the kiali-viewer ClusterRole to be bound the the kiali service-account.
2. Causes the kiali-server front-end to operate in view-only mode (even when logged in with a token that has edit privileges)

Would it be possible to split this functionality, so that the ClusterRole assignment is done on a *different* values-file, e.g.:

```
deployment:
  view_only_mode: false
  sa_view_only: true
```
In this scenario:
1. Caused the kiali-viewer ClusterRole to be bound the the kiali service-account.
2. Causes the kiali-server front-end to operate in edit mode if the logged-in user has the correct rights (say, in a cluster-admin role).

## Reasoning
We operate in a secure environment and try to provide least-privilege to all service-accounts. However when we do this with kiali , properly-authenticated users are unable to edit istio resources from the UI.

jmazzitelli:
I do not believe that will not be possible because some of under-the-cover work done by Kiali is done via the SA (not everything is done via the user credentials). I suspect this kind of mixing of permissions might break things. I say this but I am not 100% sure in my answer though. Feel free to perform some investigation if its possible to turn on "view-only" in the SA but have users be able to edit resources through the Kiali frontend (would need some hacking of the frontend code to do things like show the "Edit" buttons in the editors even if view-only it enabled).

technotaff-nbs:
We have patched the ClusterRoleBinding and restarted Kiali, everything *seems* okay. I'll double check for errors in the logs.

Jinn22:
I don't see any errors in kiali pods when kiali service account is bound to kiali-viewer ClusterRole while I log into Kiali using a token with privileged access and I have admin access in Kiali UI.

jmazzitelli:
This is valid use case for the operator; not just helm chart. Updated the title.

Now that I think about it - assuming it all works - there is no need for the Kiali SA to have read-write privileges if Kiali will use the user's credentials/token (and that is the case when using auth.strategy of "token", "openid", and "openshift" (and I believe "header" too). The only time Kiali SA would need read-write is if using auth.strategy of anonymous and view_only_mode is off.

[...]

PR #5641, opened 2022-11-23
the CI script used to run the molecule tests needs to set up the header and openid tests using new role

jmazzitelli:
part of https://github.com/kiali/kiali/issues/5611


leandroberetta (approved):
Looks good.

PR #5642, opened 2022-11-24
Kiali working with no Istiod access

### LINKED ISSUE #5626: Investigate the idea of Kiali working with no istiod connection
Investigate and define how Kiali should work - in a limited way - when there is no connection with Istiod. 

At the moment there is a hard dependency with Istiod, used to retrieve data that Istio has in memory, and also the Istio configuration. 

This is an example (With anonymous auth - special case as it uses the Istio SA to access the resources)

image
image
image

josunect:
Proposal as a first version to have Kiali running with no Istiod. 

Details in: https://docs.google.com/document/d/1XjkodS5Uu14C_1l3-j7iNaWfLufQFycY-cKMlagIxyc

Changes: 

- Not perform some calls if there is no access to istiod. They were making the system really slow (When istiod was not available): 
  - get pod proxy status for workloads
  - Get registry services for services
  - Fetch registry services
- Istio Config: Show a message indicating that Istiod is not available

Some tests to review performance: Frontend tests were showing a degraded performance so a new commit was sent to internally cache the istioAccess value

- [ ] How this could be tested
- [ ] Documentation PR: https://github.com/kiali/kiali.io/pull/613
- [ ] Operator PR: https://github.com/kiali/kiali-operator/pull/595

Fixes #5626

### DIFF
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -78,7 +78,7 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 	var rbacDetails kubernetes.RBACDetails
 	var registryServices []*kubernetes.RegistryService
 
-	var istioApiEnabled = config.Get().ExternalServices.Istio.IstioApiEnabled
+	var istioApiEnabled = config.Get().ExternalServices.Istio.IstioAPIEnabled
 
 	wg.Add(3) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
 	if service != "" {
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -78,7 +78,7 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 	var rbacDetails kubernetes.RBACDetails
 	var registryServices []*kubernetes.RegistryService
 
-	var istioApiEnabled = config.Get().IstioApiEnabled
+	var istioApiEnabled = config.Get().ExternalServices.Istio.IstioApiEnabled
 
 	wg.Add(2) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
 	if istioApiEnabled {
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -78,10 +78,10 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 	var rbacDetails kubernetes.RBACDetails
 	var registryServices []*kubernetes.RegistryService
 
-	var istioAccess = in.k8s.IstioAccess()
+	var istioApiEnabled = config.Get().IstioApiEnabled
 
 	wg.Add(2) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
-	if istioAccess {
+	if istioApiEnabled {
 		wg.Add(2)
 		if service != "" {
 			wg.Add(1)
@@ -98,7 +98,7 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 		go in.fetchAllWorkloads(ctx, &workloadsPerNamespace, &namespaces, errChan, &wg)
 	}
 
-	if istioAccess {
+	if istioApiEnabled {
 		go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
 		if service != "" {
 			go in.fetchServices(ctx, &services, namespace, errChan, &wg)
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -78,9 +78,14 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 	var rbacDetails kubernetes.RBACDetails
 	var registryServices []*kubernetes.RegistryService
 
-	wg.Add(4) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
-	if service != "" {
-		wg.Add(1)
+	var istioAccess = in.k8s.IstioAccess()
+
+	wg.Add(2) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
+	if istioAccess {
+		wg.Add(2)
+		if service != "" {
+			wg.Add(1)
+		}
 	}
 
 	// We fetch without target service as some validations will require full-namespace details
@@ -92,11 +97,14 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 	} else {
 		go in.fetchAllWorkloads(ctx, &workloadsPerNamespace, &namespaces, errChan, &wg)
 	}
-	go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
-	if service != "" {
-		go in.fetchServices(ctx, &services, namespace, errChan, &wg)
+
+	if istioAccess {
+		go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
+		if service != "" {
+			go in.fetchServices(ctx, &services, namespace, errChan, &wg)
+		}
+		go in.fetchRegistryServices(&registryServices, errChan, &wg)
 	}
-	go in.fetchRegistryServices(&registryServices, errChan, &wg)
 
 	wg.Wait()
 	close(errChan)
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -80,12 +80,13 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 
 	var istioApiEnabled = config.Get().ExternalServices.Istio.IstioApiEnabled
 
-	wg.Add(2) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
+	wg.Add(3) // We need to add these here to make sure we don't execute wg.Wait() before scheduler has started goroutines
+	if service != "" {
+		wg.Add(1)
+	}
+
 	if istioApiEnabled {
-		wg.Add(2)
-		if service != "" {
-			wg.Add(1)
-		}
+		wg.Add(1)
 	}
 
 	// We fetch without target service as some validations will require full-namespace details
@@ -98,11 +99,12 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
 		go in.fetchAllWorkloads(ctx, &workloadsPerNamespace, &namespaces, errChan, &wg)
 	}
 
+	go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
+	if service != "" {
+		go in.fetchServices(ctx, &services, namespace, errChan, &wg)
+	}
+
 	if istioApiEnabled {
-		go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
-		if service != "" {
-			go in.fetchServices(ctx, &services, namespace, errChan, &wg)
-		}
 		go in.fetchRegistryServices(&registryServices, errChan, &wg)
 	}
 
--- a/business/istio_validations.go
+++ b/business/istio_validations.go
@@ -175,6 +175,8 @@ func (in *IstioValidationsService) GetIstioObjectValidations(ctx context.Context
 	var referenceChecker ReferenceChecker
 	istioReferences := models.IstioReferencesMap{}
 
+	var istioApiEnabled = config.Get().ExternalServices.Istio.IstioAPIEnabled
+
 	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
 	// if namespace is accessible from Kiali (Deployment.AccessibleNamespaces)
 	if _, err = in.businessLayer.Namespace.GetNamespace(ctx, namespace); err != nil {
@@ -189,11 +191,20 @@ func (in *IstioValidationsService) GetIstioObjectValidations(ctx context.Context
 	errChan := make(chan error, 1)
 
 	// Get all the Istio objects from a Namespace and all gateways from every namespace
-	wg.Add(4)
+	wg.Add(3)
+
+	if istioApiEnabled {
+		wg.Add(1)
+	}
+
 	go in.fetchIstioConfigList(ctx, &istioConfigList, &mtlsDetails, &rbacDetails, namespace, errChan, &wg)
 	go in.fetchAllWorkloads(ctx, &workloadsPerNamespace, &namespaces, errChan, &wg)
 	go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
-	go in.fetchRegistryServices(&registryServices, errChan, &wg)
+
+	if istioApiEnabled {
+		go in.fetchRegistryServices(&registryServices, errChan, &wg)
+	}
+
 	wg.Wait()
 
 	noServiceChecker := checkers.NoServiceChecker{Namespaces: namespaces, IstioConfigList: &istioConfigList, WorkloadsPerNamespace: workloadsPerNamespace, AuthorizationDetails: &rbacDetails, RegistryServices: registryServices, PolicyAllowAny: in.isPolicyAllowAny()}
--- a/business/services.go
+++ b/business/services.go
@@ -99,7 +99,7 @@ func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriter
 			Namespace:       criteria.Namespace,
 			ServiceSelector: criteria.ServiceSelector,
 		}
-		if in.businessLayer.k8s.IstioAccess() {
+		if config.Get().IstioApiEnabled {
 			rSvcs, err2 = in.businessLayer.RegistryStatus.GetRegistryServices(registryCriteria)
 			if err2 != nil {
 				log.Errorf("Error fetching Registry Services per namespace %s: %s", criteria.Namespace, err2)
@@ -493,7 +493,7 @@ func (in *SvcService) GetServiceDetails(ctx context.Context, namespace, service,
 			Namespace:   namespace,
 			ServiceName: service,
 		}
-		if in.businessLayer.k8s.IstioAccess() {
+		if config.Get().IstioApiEnabled {
 			rEps, err2 = in.businessLayer.RegistryStatus.GetRegistryEndpoints(criteria)
 			if err2 != nil {
 				log.Errorf("Error fetching Registry Endpoints namespace %s and service %s: %s", namespace, service, err2)
--- a/business/services.go
+++ b/business/services.go
@@ -63,6 +63,9 @@ func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriter
 	if criteria.IncludeIstioResources {
 		nFetches = 5
 	}
+	if !config.Get().ExternalServices.Istio.IstioApiEnabled {
+		nFetches--
+	}
 
 	wg := sync.WaitGroup{}
 	wg.Add(nFetches)
@@ -92,21 +95,21 @@ func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriter
 		}
 	}()
 
-	go func() {
-		defer wg.Done()
-		var err2 error
-		registryCriteria := RegistryCriteria{
-			Namespace:       criteria.Namespace,
-			ServiceSelector: criteria.ServiceSelector,
-		}
-		if config.Get().ExternalServices.Istio.IstioApiEnabled {
+	if config.Get().ExternalServices.Istio.IstioApiEnabled {
+		go func() {
+			defer wg.Done()
+			var err2 error
+			registryCriteria := RegistryCriteria{
+				Namespace:       criteria.Namespace,
+				ServiceSelector: criteria.ServiceSelector,
+			}
 			rSvcs, err2 = in.businessLayer.RegistryStatus.GetRegistryServices(registryCriteria)
 			if err2 != nil {
 				log.Errorf("Error fetching Registry Services per namespace %s: %s", criteria.Namespace, err2)
 				errChan <- err2
 			}
-		}
-	}()
+		}()
+	}
 
 	go func() {
 		defer wg.Done()
@@ -439,8 +442,12 @@ func (in *SvcService) GetServiceDetails(ctx context.Context, namespace, service,
 	var nsmtls models.MTLSStatus
 
 	wg := sync.WaitGroup{}
-	wg.Add(6)
-	errChan := make(chan error, 6)
+	nwg := 6
+	if !config.Get().ExternalServices.Istio.IstioApiEnabled {
+		nwg = 4
+	}
+	wg.Add(nwg)
+	errChan := make(chan error, nwg)
 
 	labelsSelector := labels.Set(svc.Selectors).String()
 	// If service doesn't have any selector, we can't know which are the pods and workloads applying.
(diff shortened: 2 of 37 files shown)


Xunzhuo:
Great! If you want some reviews, ping me after this is ready! @josunect

josunect:
> Great! If you want some reviews, ping me after this is ready! @josunect

Thanks, @Xunzhuo ! It would be great! I'll let you know.

josunect:
Changes when istiod is not available:

**Istio no available messages**

This remains the same. Should we change it to just a warning instead of error? 
image

**Service mesh version unknown**

image

**Workload status warning because no proxy pod sync**

image

**Istio Config not available**

Show a message instead
image

kubernetes/istio.go
```diff
@@ -175,6 +177,21 @@ func (in *K8SClient) CanConnectToIstiod() (IstioComponentStatus, error) {
...
+func (in *K8SClient) HasIstioAccess() bool {
+	componentStatus, err := in.CanConnectToIstiod()
+	if err != nil {
+		return false
+	}
+	if len(componentStatus) == 0 {
+		return false
+	}
+	return true
+}
```
nrfox, line 189: I think this would be better as a config option for the registry where you can enable/disable the registry altogether if you'd like rather than having it just be a fallback option: https://github.com/kiali/kiali/blob/master/config/config.go#L208.

@hhovsepy may be looking into this but you should be able to get the istio config from the kube api rather than the registry when there's no istiod access. 

What will be the impact on the controlplane namespace card once there is no direct istiod access? It might still be possible to get some details, like controlplane revision, about the controlplane from the injection webhook since this may be present in the cluster even if the controlplane is not.

nrfox (commented):
Thanks for picking up this investigation!

kubernetes/istio.go
jshaughn: Funny, I was just looking at the same thing.  I also think it makes sense to configure Kiali for this scenario.  By default Kiali would expect to be able to connect to istiod's debug endpoint and proceed normally.  If it can't connect then that would be a FAILURE scenario for the control plane card.  If configured to ignore istiod then it would be a DEGRADED scenario for the control plane card (even if it is accessible we won't connect, this could be for performance reasons, or whatever).

jshaughn (commented):
Good progress @josunect , I think I agree with @nrfox about making this a more explicit configuration option for Kiali (see comments).  What do you think?

kubernetes/istio.go
hhovsepy: draft solution of retrieving configs when istiod is not available https://github.com/kiali/kiali/pull/5649

josunect:
> Good progress @josunect , I think I agree with @nrfox about making this a more explicit configuration option for Kiali (see comments). What do you think?

Thanks, @jshaughn and @nrfox for the review! I'll approach the changes and submit. 

I've though in do it this way as it can be more robust in the sense that it can work even if Istiod is or is not available, but it is true that it is not possible to know the reason, can be intentional or not - a failure. So it make sense to do it more explicit as part of the configuration.

kubernetes/istio.go
josunect: > draft solution of retrieving configs when istiod is not available #5649

Thanks, @hhovsepy !

josunect:
I've also added the following change: 

image

Xunzhuo (commented):
I have reviewed the backend codes, did not find any issues. LGTM. Wait for other reviews in front part :)

hhovsepy:
Is it supposed, that the case when 'istiod' is not running/accessible and the config "istio_api_enabled: true" will be handled?  Because it throws "open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory" error message while selecting "istiod" workload.

josunect:
> Is it supposed, that the case when 'istiod' is not running/accessible and the config "istio_api_enabled: true" will be handled? Because it throws "open /var/run/secrets/kubernetes.io/serviceaccount/token: no such file or directory" error message while selecting "istiod" workload.

In that case, it won't be handled - That should be the default behavior. But, I will reproduce it.

config/config.go
```diff
@@ -214,6 +214,7 @@ type IstioConfig struct {
 	ConfigMapName                     string              `yaml:"config_map_name,omitempty"`
 	EnvoyAdminLocalPort               int                 `yaml:"envoy_admin_local_port,omitempty"`
 	GatewayAPIClassName               string              `yaml:"gateway_api_class_name,omitempty"`
+	IstioApiEnabled                   bool                `yaml:"istio_api_enabled,omitempty"`
```
nrfox, line 217: This option is just related to the registry right? You still get controlplane information even with this disabled? If so, I think it should go under the `Registry` config and then be a bool `Enabled` to be consistent with the other options like `Tracing.Enabled`.

business/workloads.go
```diff
@@ -1718,11 +1718,15 @@ func fetchWorkload(ctx context.Context, layer *Layer, criteria WorkloadCriteria)
...
-				if err != nil {
-					log.Warningf("GetPodProxyStatus is failing for [namespace: %s] [pod: %s]: %s ", criteria.Namespace, pod.Name, err.Error())
+				if config.Get().ExternalServices.Istio.IstioApiEnabled {
+					ps, err := layer.ProxyStatus.GetPodProxyStatus(criteria.Namespace, pod.Name)
+					if err != nil {
+						log.Warningf("GetPodProxyStatus is failing for [namespace: %s] [pod: %s]: %s ", criteria.Namespace, pod.Name, err.Error())
+					}
+					pod.ProxyStatus = castProxyStatus(ps)
+				} else {
+					pod.ProxyStatus = &models.ProxyStatus{}
```
nrfox, line 1728: This might actually port forward or exec directly into the pod to get the proxy status instead of talking with the istiod. Have you tried this without istiod access and it doesn't work?

business/istio_validations.go
```diff
@@ -92,11 +97,14 @@ func (in *IstioValidationsService) GetValidations(ctx context.Context, namespace
...
-	go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
-	if service != "" {
-		go in.fetchServices(ctx, &services, namespace, errChan, &wg)
+
+	if istioApiEnabled {
+		go in.fetchNonLocalmTLSConfigs(&mtlsDetails, errChan, &wg)
+		if service != "" {
+			go in.fetchServices(ctx, &services, namespace, errChan, &wg)
+		}
+		go in.fetchRegistryServices(&registryServices, errChan, &wg)
```
nrfox, line 106: We don't want to fetch services or non local mtls configs when the registry is disabled? Seems like we only want to exclude the registry services?

config/config.go
nrfox: Even if istiod access is disabled, I think we can at least infer some controlplane info from different objects in the kube cluster if istiod is deployed there such as the istiod deployment itself or the sidecar injection webhook.

jmazzitelli:
> Many errors in Kiali pod log: `1202 12:56:50.277723 1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: failed to list *v1.Pod: pods is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "pods" in API group "" at the cluster scope

@hhovsepy What is your `deployment.accessible_namespaces` set to? Is it `['**']` or a list of namespaces?

That error message says it wants permission to list pods AT THE CLUSTER SCOPE. Which means it expects (for some reason) a ClusterRole. The Kiali SA only gets bound to a ClusterRole when accessible_namespaces is set to `['**']`. Otherwise, there are just Roles for each namespace listed in accessible_namespaces.

We never before _required_ ClusterRole access to look at pods. So this error message seems strange. We should avoid this requirement if possible.

hhovsepy (dismissed):
Verified now. After correct installation and configuration on OCP cluster.

business/workloads.go
josunect: Thanks, @nrfox, this is a good suggestion. I would try to get the information like this.

config/config.go
josunect: I think it would make sense if we are getting some more controlplane info

business/services.go
```diff
@@ -476,10 +478,12 @@ func (in *SvcService) GetServiceDetails(ctx context.Context, namespace, service,
...
-			rSvcs, err2 = in.businessLayer.RegistryStatus.GetRegistryServices(registryCriteria)
-			if err2 != nil {
-				log.Errorf("Error fetching Registry Services per namespace %s: %s", registryCriteria.Namespace, err2)
-				errChan <- err2
+			if config.Get().ExternalServices.Istio.IstioApiEnabled {
+				rSvcs, err2 = in.businessLayer.RegistryStatus.GetRegistryServices(registryCriteria)
+				if err2 != nil {
+					log.Errorf("Error fetching Registry Services per namespace %s: %s", registryCriteria.Namespace, err2)
+					errChan <- err2
+				}
```
jshaughn, line 486: Same comment here, maybe just skip the whole go func() block?

business/workloads.go
jshaughn: Although this may work, I'm a little worried about perf when scaling to a lot of pods.  Is that something we should worry about?  Do we already exec into the pod for some other reason?  For me, if the users elect to block istiod access perhaps they can expect to lose this  feature.

jshaughn (commented):
A couple of comments.  Also, is there an operator PR for the new config?

business/services.go
```diff
@@ -99,10 +99,12 @@ func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriter
...
-		rSvcs, err2 = in.businessLayer.RegistryStatus.GetRegistryServices(registryCriteria)
-		if err2 != nil {
-			log.Errorf("Error fetching Registry Services per namespace %s: %s", criteria.Namespace, err2)
-			errChan <- err2
+		if config.Get().ExternalServices.Istio.IstioApiEnabled {
+			rSvcs, err2 = in.businessLayer.RegistryStatus.GetRegistryServices(registryCriteria)
+			if err2 != nil {
+				log.Errorf("Error fetching Registry Services per namespace %s: %s", criteria.Namespace, err2)
+				errChan <- err2
+			}
```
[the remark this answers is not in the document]
josunect: Updated

business/services.go
josunect: Updated

business/services.go
```diff
@@ -491,11 +495,14 @@ func (in *SvcService) GetServiceDetails(ctx context.Context, namespace, service,
 			Namespace:   namespace,
 			ServiceName: service,
 		}
-		rEps, err2 = in.businessLayer.RegistryStatus.GetRegistryEndpoints(criteria)
-		if err2 != nil {
-			log.Errorf("Error fetching Registry Endpoints namespace %s and service %s: %s", namespace, service, err2)
-			errChan <- err2
+		if config.Get().ExternalServices.Istio.IstioApiEnabled {
```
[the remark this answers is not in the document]
josunect: Updated

business/istio_validations.go
josunect: Updated

josunect:
> A couple of comments. Also, is there an operator PR for the new config?

Regarding adding the option into the Registry. This is a comment from the code: 

// RegistryConfig contains configuration for connecting to an external istiod.
// This is used when Kiali should connect to the istiod via a url instead of port forwarding.

https://github.com/kiali/kiali/blob/master/config/config.go#L204 

But with this option, we will be disabling the connection to an external or local Istio. 
Also, some of the options in IstioConfig (The outside configuration) won't take effect if we are disabling the Istiod access, so probably it would make sense to be outside that section?

josunect:
> A couple of comments. Also, is there an operator PR for the new config?

I was trying to figure out the possibility to obtain the proxy status information but: 

- istioctl doesn't seem to be available in the istio-proxy container: 

```
kubectl exec -i -t details-v1-5fdd69c6f9-wrn5b -n bookinfo -c istio-proxy -- /bin/bash

istio-proxy@details-v1-5fdd69c6f9-wrn5b:/$ istioctl  
bash: istioctl: command not found
istio-proxy@details-v1-5fdd69c6f9-wrn5b:/$ find . -name istioctl
find: './etc/ssl/private': Permission denied
find: './var/cache/apt/archives/partial': Permission denied
find: './var/cache/ldconfig': Permission denied
find: './proc/tty/driver': Permission denied
find: './root': Permission denied
istio-proxy@details-v1-5fdd69c6f9-wrn5b:/$ 
```
Because I guess we are not supposed to have access to the remote istiod pod?
But I'm not sure if it is possible to get this information - It looks like this is calculated using some logic in istiod.

jshaughn:
> But with this option, we will be disabling the connection to an external or local Istio.
Also, some of the options in IstioConfig (The outer configuration) won't take effect if we are disabling the Istiod access, so probably it would make sense to be outside that section?

Ah, OK, I didn't have the complete picture.  I am fine with the current setting.

business/workloads.go
nrfox: > For me, if the users elect to block istiod access perhaps they can expect to lose this feature.

@jshaughn this is what I'm wondering. I wasn't sure if the proxy status came from the pod's proxy or from the istiod pod but it looks like it comes from the istiod pod: https://github.com/kiali/kiali/blob/2165abf70c4cfae7d4a08665af6a008329a35ab3/kubernetes/istio.go#L254-L275. In which case this feature _should_ be disabled when istiod access is disabled. Sorry for the noise!

config/config.go
jshaughn: Given the explanation below, I am now fine with this as-is.  Not sure if @nrfox agrees, perhaps he can comment.

business/workloads.go
jshaughn: no worries, it was a good question.

config/config.go
nrfox: I was suggesting `ExternalServices.IstioConfig.Registry.Enabled` since then it'd be consistent with the other external service configs like `ExternalServices.Tracing.Enabled`. Some fields in `IstioConfig` are still relevant even with the registry disabled such as `GatewayAPIClassName` so having a top level `Enabled: false` field would be confusing if other fields in that config are still relevant. If consistency with the other external services is not important since the Istio external service is more unique anyways then I'm good with keeping `IstioApiEnabled`. What do y'all think?

One minor nit if we're keeping `IstioApiEnabled` it should be `IstioAPIEnabled`.

hhovsepy (approved):
Tested with "istio_api_enabled: false" and istiod pod not running:
Screenshot from 2022-12-15 14-37-17

josunect:
> Tested with "istio_api_enabled: false" and istiod pod not running: Screenshot from 2022-12-15 14-37-17

Thanks for reviewing, @hhovsepy !

frontend/src/pages/IstioConfigList/IstioConfigListPage.tsx
```diff
@@ -153,27 +156,49 @@ class IstioConfigListPageComponent extends FilterComponent.Component<
...
+          </>)}
+        {this.props.istioAPIEnabled === false && (
+          <>
+            <div style={{ backgroundColor: '#fff' }}>
+              <DefaultSecondaryMasthead
+              />
+            </div>
+            <RenderContent>
+              <div style={{ padding: '10px' }}>
+                <WarningTriangleIcon color={PFColors.Warning} style={{marginRight: '5px'}} /> Istio API is disabled and Istio Config is read only.
```
nrfox, line 184: Why is editing istio config disabled when the registry is disabled?

It looks like you can still use the wizards to create/update/delete istio config.

frontend/src/components/IstioStatus/IstioStatus.tsx
```diff
@@ -94,6 +100,10 @@ export class IstioStatus extends React.Component<Props> {
     let addonUnhealthy: boolean = false;
     let notReady: boolean = false;
 
+    if (this.props.istioAPIEnabled === false) {
+      return PFColors.Warning
+    }
```
nrfox, line 105: I wonder if this should really be a permanent warning if users have purposefully disabled istiod access by setting the config option?

business/workloads.go
```diff
@@ -1718,11 +1718,15 @@ func fetchWorkload(ctx context.Context, layer *Layer, criteria WorkloadCriteria)
...
-				if err != nil {
-					log.Warningf("GetPodProxyStatus is failing for [namespace: %s] [pod: %s]: %s ", criteria.Namespace, pod.Name, err.Error())
+				if config.Get().ExternalServices.Istio.IstioAPIEnabled {
+					ps, err := layer.ProxyStatus.GetPodProxyStatus(criteria.Namespace, pod.Name)
+					if err != nil {
+						log.Warningf("GetPodProxyStatus is failing for [namespace: %s] [pod: %s]: %s ", criteria.Namespace, pod.Name, err.Error())
+					}
+					pod.ProxyStatus = castProxyStatus(ps)
+				} else {
+					pod.ProxyStatus = &models.ProxyStatus{}
```
nrfox, line 1728: Having an empty proxy status will cause workloads to show a warning on their health details:

Screenshot from 2022-12-15 09-47-30

This is more to do with how the frontend is calculating the workload health. I'm not sure we should show the workload as being unhealthy or show a warning when the istiod access has been disabled.

hhovsepy:
For making istio configs edit/update enabled, 'validations.istio.io' webhook needs to be disabled, which is called on k8s side when Kiali calls directly k8s client for creating configs.

nrfox:
> For making istio configs edit/update enabled, 'validations.istio.io' webhook needs to be disabled, which is called on k8s side when Kiali calls directly k8s client for creating configs. Otherwise any k8s api edit/create calls ends "Internal error occurred: failed calling webhook "validation.istio.io": failed to call webhook:"

What webhook is that?

josunect:
> > For making istio configs edit/update enabled, 'validations.istio.io' webhook needs to be disabled, which is called on k8s side when Kiali calls directly k8s client for creating configs. Otherwise any k8s api edit/create calls ends "Internal error occurred: failed calling webhook "validation.istio.io": failed to call webhook:"
> 
> What webhook is that?

I think is ValidatingWebhookConfiguration: 

```
kubectl get ValidatingWebhookConfiguration
NAME                           WEBHOOKS   AGE
istio-validator-istio-system   1          3d2h
istiod-default-validator       1          3d2h
```

But, i should say this issue did not happen in my local environment when creating an object. I will try to reproduce

business/workloads.go
nrfox: @josunect I still don't think we should show a warning label here for the empty proxy status when istiod access is disabled. As it is, pods will always be in a "warning" state but if you've purposefully disabled registry access then imo the frontend shouldn't try to use proxy status for the health calculation if it has no way of ever getting it.

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -885,75 +889,83 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+                                  <ControlPlaneVersionBadge version={this.state.canaryUpgradeStatus.upgradeVersion}
+                                                            isCanary={true}></ControlPlaneVersionBadge>
                                 }
                                 {ns.name !== serverConfig.istioNamespace && this.hasCanaryUpgradeConfigured() && this.state.canaryUpgradeStatus?.pendingNamespaces.includes(ns.name) &&
-                                  <ControlPlaneVersionBadge version={this.state.canaryUpgradeStatus.currentVersion} isCanary={false}></ControlPlaneVersionBadge>
+                                  <ControlPlaneVersionBadge version={this.state.canaryUpgradeStatus.currentVersion}
+                                                            isCanary={false}></ControlPlaneVersionBadge>
+                                }
+                                {ns.name === serverConfig.istioNamespace && !this.props.istioAPIEnabled &&
+                                  <Label style={{ marginLeft: 5 }} color={"orange"} isCompact>Istio API disabled</Label>
```
nrfox, line 900: I think this label looks great!

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -885,75 +889,83 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
...
+                                        {this.state.canaryUpgradeStatus && this.hasCanaryUpgradeConfigured() &&
+                                          <GridItem md={4}>
+                                            <CanaryUpgradeProgress
+                                              canaryUpgradeStatus={this.state.canaryUpgradeStatus}/>
+                                          </GridItem>
+                                        }
+                                        {this.props.istioAPIEnabled === true &&
+                                          <GridItem md={(this.hasCanaryUpgradeConfigured()) ? 8 : 12}>
+                                            {this.renderCharts(ns)}
+                                          </GridItem>}
```
nrfox, line 943: Minor: You don't need to resolve this as part of this PR but there's now a bunch of whitespace on the controlplane card without the metrics:

Screenshot from 2022-12-19 21-02-48

I'm not sure what to fill that space with.

hhovsepy:
So when istiod is not available, this means it is not installed, thus "ValidatingWebhookConfiguration istiod-default-validator" is also not created.
So the solution to bring edit back for Istion configs can be merged: https://github.com/josunect/kiali/pull/6

josunect:
> So when istiod is not available, this means it is not installed, thus "ValidatingWebhookConfiguration istiod-default-validator" is also not created. So the solution to bring edit back for Istion configs can be merged: [josunect#6](https://github.com/josunect/kiali/pull/6)

Thanks, @hhovsepy ! It has been merged

nrfox (commented):
The only remaining issue I can see is that validations appear to be somewhat degraded with the istio api disabled. As an example, edit a destination rule's subset to match some non-existent label and you'll see a validation error with registry enabled but see no error with the registry disabled. Perhaps some validations need info from the registry? 

@hhovsepy not sure if validations should be degraded or disabled entirely with the registry disabled? If they can still provide some value without the registry imo it would be worth keeping but perhaps there should be a note saying that validations are degraded without registry access? What is your opinion @hhovsepy @josunect?

hhovsepy:
@nrfox the idea of making configs read only was, that Kiali validations will be degraded, because they require registry services. Also, when istiod does not exist, the Istio validations are not performed, so the user can break the configs even without noticing.
There can be a message notifying to user in Istio config YAML edit page that validations are degraded because istiod api is disabled, edit the config on your risk.

nrfox:
> @nrfox the idea of making configs read only was, that Kiali validations will be degraded, because they require registry services. Also, when istiod does not exist, the Istio validations are not performed, so the user can break the configs even without noticing.

Ah I'm sorry then for I didn't realize that the read only mode for istio config was to workaround the validations being degraded when I suggested adding back editing the configs.

> There can be a message notifying to user in Istio config YAML edit page that validations are degraded because istiod api is disabled, edit the config on your risk.

This sounds fine to me. I'm also good with making the istio configs read only again for now if that's what you'd prefer for the validations. Perhaps some of the validations that rely on the registry can still be performed in the future without it. For example, it seems like Kiali should be able to validate Destination Rules matching workloads without needing the registry but this doesn't work today. That's a separate issue though and will require its own investigation.

josunect:
I think we can put a message to let the user know, and still have the possibility to be able to create/edit/remove Istio Configs. Because, this still will be possible to do it manually, right?

hhovsepy:
Yeah agree to put a message on editable configs, and have a separate task for validations: https://github.com/kiali/kiali/issues/5676

josunect:
A message like this? 

image

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -864,18 +867,20 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
                 {filteredNamespaces.map((ns, i) => {
                   const isLongNs = ns.name.length > NS_LONG;
                   return (
-                    <GridItem sm={(ns.name === serverConfig.istioNamespace && this.state.displayMode === OverviewDisplayMode.EXPAND) ? lg : sm}
```
nrfox, line 867: Minor: it looks like the controlplane card's height doesn't match the other cards now with the istio api disabled:
Screenshot from 2023-01-09 12-41-30

nrfox (commented):
A few minor comments otherwise LGTM. Looks like you'll need to rebase with the main branch before this can be merged.

jshaughn:
@josunect Please rebase against and then let's proceed with a final review!

josunect:
It will need to test further after the merge. I see some errors like: 

```
2023-01-10T15:01:54Z WRN Error getting proxy status from istiod. Proxy status may be stale. Err: unable to proxy Istiod pods. Make sure your Kubernetes API server has access to the Istio control plane through 8080 port

2023-01-10T15:03:54Z ERR Failed to call Istiod endpoint /debug/syncz error: unable to proxy Istiod pods. Make sure your Kubernetes API server has access to the Istio control plane through 8080 port
```

frontend/src/components/IstioWizards/ServiceWizard.tsx
```diff
@@ -725,6 +735,10 @@ class ServiceWizard extends React.Component<ServiceWizardProps, ServiceWizardSta
               this.onConfirmPreview(items);
             }}
           />
+          {!this.props.istioAPIEnabled &&
+            (<div style={{padding: " 0 0 20px 0"}}>
+              <KialiIcon.Warning /> <b>Istio API is disabled.</b> Be careful when editing the configuration as changes may not be compliant (The validations are disabled)
```
nrfox, line 740:
```suggestion
              <KialiIcon.Warning /> <b>Istio API is disabled.</b> Be careful when editing the configuration as the Istio config validations are disabled when the Istio API is disabled.
```

nrfox: Missed this in my last review but not a blocker to make it consistent with the other messages.

nrfox (approved):

kubernetes/cache/proxy_status.go
```diff
@@ -18,46 +19,51 @@ type ProxyStatusCache interface {
...
-			case <-ctx.Done():
-				log.Debug("[Kiali Cache] Stopping polling for istiod proxy status")
-				return
-			case <-time.After(c.tokenNamespaceDuration):
-				// Get the proxy status from istiod with some retries.
-				// Wrapping this so we can defer cancel.
-				func() {
-					ctx, cancel := context.WithTimeout(ctx, c.tokenNamespaceDuration)
-					defer cancel()
+	if config.Get().ExternalServices.Istio.IstioAPIEnabled {
```
nrfox, line 22: Can you move this if statement up to where the function is called in `NewKialiCache` https://github.com/kiali/kiali/blob/master/kubernetes/cache/cache.go#L210-L211? Calling a function that actually does nothing if you've enabled a config option is an odd side effect and could be confusing later on.

nrfox (approved):
LGTM

PR #5654, opened 2022-11-29
[tests] create a SA whose token will be used to log into kiali

jmazzitelli:


jmazzitelli:
FYI: frontend tests now pass. This PR looks correct.

hhovsepy (approved):

josunect:
Tested and works as expected!

Issue #5599, opened 2022-11-01
Remove option to disable kube caching


nrfox:
Caching has been enabled by default for some time now. This issue is to remove the option to disable kube caching. Kiali may still query the kube api directly in some cases but will largely rely on the local cache. Some supported kube types are not currently cached like `CronJob` but these should be added as well.

This is a followup to #5319 and #4678.

nrfox:
Rather than blocking making the cache mandatory on refactoring all the backend unit tests, @israel-hdez had the idea of removing the external config option while keeping the internal cfg option that the unit tests can still set. I think this is a better way forward and we should do this while the test refactoring is on-going.

@israel-hdez I'm assigning you to this task but feel free to unassign yourself or assign someone else.

israel-hdez:
Transferring this issue to @leandroberetta as the new Ωwner.

PR #5664, opened 2022-12-08
Use cache only for workload service

nrfox:
Relates to #5599. Splitting out the changes into multiple PRs to make the more reviewable.

This PR also fixes some race conditions present in the cache.

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -386,7 +386,10 @@ func (in *WorkloadService) UpdateWorkload(ctx context.Context, namespace string,
 	// Cache is stopped after a Create/Update/Delete operation to force a refresh.
 	// Refresh once after all the updates have gone through since Update Workload will update
 	// every single workload type of that matches name/namespace and we only want to refresh once.
-	in.cache.Refresh(namespace)
+	// TODO: Remove conditional once cache is mandatory
+	if in.cache != nil {
+		in.cache.Refresh(namespace)
+	}
 
 	// After the update we fetch the whole workload
 	return in.GetWorkload(ctx, WorkloadCriteria{Namespace: namespace, WorkloadName: workloadName, WorkloadType: workloadType, IncludeServices: includeServices})
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -27,17 +27,33 @@ import (
 	"github.com/kiali/kiali/business/checkers"
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/kubernetes/cache"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/observability"
 	"github.com/kiali/kiali/prometheus"
 )
 
+func NewWorkloadService(k8s kubernetes.ClientInterface, prom prometheus.ClientInterface, cache cache.KialiCache, layer *Layer, config *config.Config) *WorkloadService {
+	return &WorkloadService{
+		k8s:           k8s,
+		prom:          prom,
+		cache:         cache,
+		businessLayer: layer,
+		config:        config,
+	}
+}
+
 // WorkloadService deals with fetching istio/kubernetes workloads related content and convert to kiali model
 type WorkloadService struct {
-	prom          prometheus.ClientInterface
-	k8s           kubernetes.ClientInterface
+	prom prometheus.ClientInterface
+	k8s  kubernetes.ClientInterface
+	// Careful not to call the workload service from here as that would be a infinite loop.
 	businessLayer *Layer
+	// The global kiali cache. This should be passed into the workload service rather than created inside of it.
+	cache cache.KialiCache
+	// The global kiali config.
+	config *config.Config
 }
 
 type WorkloadCriteria struct {
@@ -320,7 +336,7 @@ func (in *WorkloadService) GetWorkload(ctx context.Context, criteria WorkloadCri
 	wg.Add(1)
 	go func() {
 		defer wg.Done()
-		conf := config.Get()
+		conf := in.config
 		app := workload.Labels[conf.IstioLabels.AppLabelName]
 		version := workload.Labels[conf.IstioLabels.VersionLabelName]
 		runtimes = NewDashboardsService(ns, workload).GetCustomDashboardRefs(criteria.Namespace, app, version, workload.Pods)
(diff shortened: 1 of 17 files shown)


kubernetes/cache/cache.go
```diff
@@ -172,6 +179,11 @@ func NewKialiCache(namespaceSeedList ...string) (KialiCache, error) {
 
 	cacheNamespacesRegexps := make([]regexp.Regexp, len(cacheNamespaces))
 	for i, ns := range cacheNamespaces {
+		// '**' is a special character for selecting all namespaces
+		// but needs to be converted to an actual regexp.
+		if ns == "**" {
+			ns = ".*"
+		}
```
israel-hdez, line 186: Is this fixing some bug? I cannot figure out how it was working in the past.

kubernetes/cache/kubernetes_test.go
```diff
@@ -33,3 +34,34 @@ func TestKubeGetAndListReturnKindInfo(t *testing.T) {
...
+
+	stop := make(chan bool)
+	go func() {
+		for {
+			select {
+			case <-stop:
+				return
+			default:
+				_, err := kialiCache.GetDeployment(d.Namespace, d.Name)
+				require.NoError(err)
```
israel-hdez, line 60: Are you sure this branch will actually run?
And why do you need the loop?

israel-hdez (commented):
A few initial comments and questions.

business/workloads.go
```diff
@@ -362,15 +378,15 @@ func (in *WorkloadService) UpdateWorkload(ctx context.Context, namespace string,
...
 	}
 
-	// Cache is stopped after a Create/Update/Delete operation to force a refresh
-	if kialiCache != nil && err == nil {
-		kialiCache.Refresh(namespace)
-	}
+	// Cache is stopped after a Create/Update/Delete operation to force a refresh.
+	// Refresh once after all the updates have gone through since Update Workload will update
+	// every single workload type of that matches name/namespace and we only want to refresh once.
+	in.cache.Refresh(namespace)
```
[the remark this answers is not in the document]
nrfox: That's right. As a consequence of splitting this change out over multiple PRs, the cache can still be disabled in this PR and therefore `in.cache` can be `nil` and we should guard against that here.

business/workloads.go
```diff
@@ -604,15 +616,8 @@ func fetchWorkloads(ctx context.Context, layer *Layer, namespace string, labelSe
...
-
-		// Check if namespace is cached
-		// Namespace access is checked in the upper caller
 		var err error
-		if IsNamespaceCached(namespace) {
-			pods, err = kialiCache.GetPods(namespace, labelSelector)
-		} else {
-			pods, err = layer.k8s.GetPods(namespace, labelSelector)
-		}
+		pods, err = layer.k8s.GetPods(namespace, labelSelector)
```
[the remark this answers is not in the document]
nrfox: So `layer.k8s` will use the `cachingClient` when the kube cache is enabled which is just a different implementation of the client interface that uses the cache when the cache is supported for that method call like `GetPods`. We don't need the `cachingClient` and can instead call the kiali cache directly in the business layer as it's done here but imo it makes it easier to transition these services over to just using the cache.

kubernetes/cache/cache.go
nrfox: I added this awhile back and I honestly can't remember if it was a bug fix or not. I'll double check and remove if it works fine without it.

kubernetes/cache/kubernetes_test.go
nrfox: You can turn on debug logs for the tests and see that `GetDeployment` is being called: `make GO_TEST_FLAGS="-race -v -run=\"TestConcurrentAccessDuringRefresh\"" test`. The loop is to help ensure that we're calling `GetDeployment` at the same time that we are recreating the cache with `kialiCache.Refresh("test")`. It just calls `GetDeployment` a whole bunch of times continuously to catch race conditions where the cache is being refreshed while something is reading from the cache.

business/workloads.go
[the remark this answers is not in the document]
israel-hdez: OK, now I got it.
Probably I was confused and I thought that `layer.k8s` was always the usual kubeclient giving me the idea that the cache was going to be bypassed. But I see that in `layer.go` it will be either the `K8sClient` or the `cachingClient`. So, all good.

kubernetes/cache/kubernetes_test.go
israel-hdez: Again, I think something confused me, but for this one I cannot remember what it was.
Perhaps, what I don't like too much is that this is kinda of doing brute forcing to catch the race conditions. So, it is non-deterministic. However, I cannot come up with a proper way of doing it.

business/workloads_test.go
```diff
@@ -978,34 +893,20 @@ func TestGetWorkloadListRSOwnedByCustom(t *testing.T) {
...
+	}
+	for _, obj := range replicaSets {
+		o := obj
+		kubeObjs = append(kubeObjs, &o)
+	}
+	for _, obj := range pods {
+		o := obj
+		kubeObjs = append(kubeObjs, &o)
+	}
+	k8s := kubetest.NewFakeK8sClient(kubeObjs...)
```
israel-hdez, line 907: Good simplification of the tests!

israel-hdez (approved):
I clicked through the UI and I wasn't able to see any issues.

Approving!

business/workloads_test.go
```diff
@@ -139,23 +120,19 @@ func TestGetWorkloadListFromReplicaSets(t *testing.T) {
...
-
-	svc := setupWorkloadService(k8s)
+	kubeObjs := []runtime.Object{
+		&osproject_v1.Project{ObjectMeta: v1.ObjectMeta{Name: "Namespace"}},
+	}
+	for _, obj := range FakeReplicationControllers() {
+		o := obj
+		kubeObjs = append(kubeObjs, &o)
+	}
+	k8s := kubetest.NewFakeK8sClient(kubeObjs...)
```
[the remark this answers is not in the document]
nrfox: I tried doing this a number of ways actually and wasn't able to get anything working. It'd be nice to be able to more easily convert a slice of impls of an interface to the interface type but I ran into this design issue: https://go.googlesource.com/proposal/+/HEAD/design/43651-type-parameters.md#no-way-to-express-convertibility. That being said, I have very little experience with go generics so it's very possible that there's a way to simplify these that I'm not aware of.

Issue #5663, opened 2022-12-08
Redirect Loop on OpenID Connect Failures


LaikaN57:
I have recently upgraded from v1.53.0 to v1.59.0. However it seems there have been some OIDC changes that are causing login errors. I would expect to be able to see these error messages on the login screen, however due to a recent code change (https://github.com/kiali/kiali/pull/5438) I just constantly get redirected back to my IdP (and back to kiali, and back to my IdP, ...).

Here are the relevant URLs that are getting looped:

```
https://[REDACTED].okta.com/oauth2/v1/authorize?
    client_id=0o[REDACTED]d8
    response_type=id_token
    redirect_uri=https://kiali.[REDACTED].com/kiali
    scope=openid+profile+email
    nonce=dc[REDACTED]ea
    state=96[REDACTED]69-221207231242

https://kiali.[REDACTED].com/kiali#
    state=96[REDACTED]69-221207231242
    error=unsupported_response_type
    error_description=The+response+type+is+not+supported+by+the+authorization+server.+Configured+response+types%3A+%5Bcode%5D.
```

I am not 100% sure, but this may be related to `this.props.message` not being set before the redirect code runs.

israel-hdez:
Hi @LaikaN57 

I've confirmed the redirect loop is happening, although the issue is only hit when using the _implicit flow_. We've been wanting to remove support for the _implicit flow_ (see #4705). I'll fix the issue if it is easy. Regardless, I would suggest you to move to the _authorization code_ flow.

israel-hdez:
Fix goes to Kiali v1.61.
However, take note that support for _implicit flow_ is going to be deprecated and removed in the short term.

LaikaN57:
Yup, I am actively trying to migrate. This change at least just helps me troubleshoot if I have things misconfigured. Thanks again.

PR #5666, opened 2022-12-09
Catch OAuth/OpenId implicit flow error params

israel-hdez:
In OAuth/OpenId implicit flow, errors are passed as hash params rather than query params. Errors passed as hash params weren't being read potentially causing a login/redirect loop, because the front-end assumes that the user was landing for the first time in Kiali and doing the automatic redirection to the 3rd party provider.

Fixes kiali/kiali#5663


LaikaN57 (approved):

josunect (approved):
Tested with implicit flow: 

image


No redirect loops on error. 

Code LGTM. 

Thanks, @israel-hdez !

israel-hdez:
Thanks for the review @LaikaN57 and @josunect. I'm merging.

LaikaN57:
Thanks for the quick turn-around time @israel-hdez!

Issue #5618, opened 2022-11-10
Multicluster


nrfox:
Kiali should support a mesh that spans multiple clusters providing a single unified view of the mesh.

There have been some past discussions around this: https://github.com/kiali/kiali/discussions/categories/multicluster.

israel-hdez:
Removing my assignment...

jmazzitelli:
> I'm really hoping to see this finished/merged, the whole point having such a view is to be able to see exactly where the traffic flows, right? :)

The Kiali team and community have been spending months on this epic with some of these features already in existing Kiali releases, and more to come.

But we figured... nah, forget it, and we abandoned it. ;)

nrfox:
@Aur0nd you can already use Kiali with multi-cluster istio deployments. See https://kiali.io/docs/configuration/multi-cluster/ for more info. What's left in this issue as of today is mostly cleanup or enabling specific deployments (openshift).

nrfox:
Closing this as completed since support for all deployment models has been added which was the goal of this epic. Future multi-cluster related work and enhancements can be tracked separately.

PR #5673, opened 2022-12-15
Multicluster proposal

nrfox:
Draft proposal for multicluster support in Kiali.

Relates to: #5618


design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Custom aggregation of all Kiali’s datasources across clusters. There are other tools and initiatives for aggregating metrics/logs/traces such as otel and this does not need to be handled by Kiali.
+- Ambient is not going to be touched in this effort.
+- Multi-mesh support - multi-cluster is focused on single mesh support only but shouldn’t prevent Kiali from supporting multi-mesh in the future.
+
+# Solution
+
+A single Kiali configured to access each Kubernetes cluster that is part of the mesh. Kiali will require Metrics/traces to be aggregated in a centralized datastore that Kiali can query.
+
+In a multicluster deployment, Kiali will only support the OIDC auth strategy for now.
```
jmazzitelli, line 37: Does this mean OpenShift will not be supported (i.e. `openshift` auth strategy)?

jmazzitelli: nm.. I see it below... `The openshift strategy does not support multiple clusters.`

nrfox: There might be a way to configure your openshift clusters to use the same OIDC provider and for Kiali in turn to just use the `openshift` strategy but I'm not familiar enough with the `openshift` strategy to know if this would work without any changes to Kiali itself to support this. OIDC works out of the box today once you've configured Kiali and your clusters properly.

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Testing
+
+All automated e2e testing, API and cypress, is currently performed against a single cluster in CI. There will be a large number of Kiali features related to multi-cluster so there needs to be a way to test multi-cluster deployments frequently and in an automated fashion. Any e2e tests that require a multi-cluster environment will need to be run separately from the rest of the suite.
+
+## Other solutions
+
+### Kiali Agent
+
+A process that runs on each kube cluster that is a part of the mesh and exports Kiali’s view of that mesh to a centralized Kiali. This option offers potential performance gains and ease of configuration since Kiali no longer has to communicate with each individual kube cluster. However, it complicates Kiali’s ability to perform authorization with RBAC since the agents would be exporting data to a centralized Kiali and the centralized Kiali would not have access to the individual clusters. The agent itself adds an additional component to Kiali that would need to be maintained and configured across clusters.
```
jmazzitelli, line 147: The more I think about this Kiali agent idea, the more it scares me. It will completely circumvent k8s RBAC security. Yes, the agent itself will have k8s RBAC, but it would expose data to the central Kiali without any k8s RBAC behind it. This will cause angst with many people, and surely would be ripe for CVEs as security holes are punched into the implementation. We should rely on k8s itself for our RBAC/permissions everywhere.

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+![Multi-Primary](Multi-Primary.png "Multi Primary")
+
+### Primary-Remote deployment
+
+![Primary-Remote](Primary-Remote.png "Primary Remote")
+
+## Aggregated metrics and traces
+
+Kiali will require metrics/traces to be aggregated into a centralized datastore rather than aggregating this information itself. It is the user’s responsibility to ensure that metrics/traces are centralized by some means. Metrics and traces will need to include a cluster identifier as an attribute/label on each metric so that Kiali can distinguish between workloads/services running on separate clusters. [Istio standard metrics](https://istio.io/latest/docs/reference/config/metrics/#labels) includes `destination_cluster` and `source_cluster`. Kiali already uses these values for cluster boxing. Traces do no include a cluster identifier but one can be added by creating a telemetry api object. Example:
```
TommasoBagassi, line 64: "Traces do no include..." missing a "t"

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+The istiods provide istio config and some information about the mesh workloads. However, the endpoints are not always exposed outside of the istiod pod. Kiali today uses port forwarding through the kubernetes API to connect to the istiod pods and this would require access to the kube API anyways. Some other features rely on the kube API like streaming logs and any create/update operations need to use the Kube API ultimately making this an un-viable option to keep most of the existing Kiali features.
+
+# Roadmap
+
+- [ ] Multicluster Kubernetes support for Kiali.
+  - [ ] Cache can be configured for multiple kubernetes clusters.
+  - [ ] Client can be configured for multiple kubernetes clusters.
+- [ ] Business services query across configured clusters for resources.
+- [ ] Investigate automated testing setup: how kiali should test multi-cluster features in CI/CD.
+- [ ] Multi-cluster demos for the different deployment models. These can also be used for testing.
```
josunect, line 160: Maybe add a new point regarding the UI changes? Something like: 
[] Define UI/UX changes to support MC

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+1. Creating a service account in the target cluster.
+2. Granting proper permissions for this service account.
+3. Creating a secret with the credentials for this service account in the cluster/namespace where Kiali is deployed.
+
+This process will need to be repeated for every new cluster added. Adding or removing clusters will also require restarting the Kiali pod in order for the cache watches to pick up the new kube config. Adding/removing kube clusters without having to restart Kiali can be an improvement made later.
+
+## Operator
+
+The Kiali operator will still be used for configuring Kiali itself however, it's assumed that the operator won't have access to other kube clusters other than where Kiali is deployed and therefore configuring multicluster access for Kiali will be a manual operation. The operator will remain unchanged in a multi-cluster deployment.
```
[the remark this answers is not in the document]
nrfox: The credentials will be a kubeconfig file that fully describes how to connect to the cluster. That will contain details on how to connect to the cluster and metadata about the cluster. We shouldn't need to tie the credential file name to the cluster name. We can provide a script to help generate this.

You will need to mount them into the kiali pod but the naming shouldn't matter so much for identifying the cluster.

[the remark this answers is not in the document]
jmazzitelli: Which type of secret are we assuming the user will create? Types are: https://kubernetes.io/docs/concepts/configuration/secret/#secret-types

I am not sure how it will work if we don't identify which credentials belong to which cluster. For example, if the graph is showing 2 nodes, one from cluster-1 and one from cluster-2, and the user wants to see details for the node for cluster-2, how does the Kiali server know which credentials to use to connect to cluster-2 to get data for that node?

design/KEPS/multicluster/proposal.md
nrfox: Yeah it would either be less secure or add a bunch of complexity that probably isn't even feasible for Kiali to manage. There's probably some users for which having a central Kiali without RBAC enforced on the remote cluster resources is ok but this won't work today and we'd need additional changes to support this less secure model. I'd rather start with a model where RBAC is still enforced on the remote clusters.

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+A single Kiali configured to access each Kubernetes cluster that is part of the mesh. Kiali will require Metrics/traces to be aggregated in a centralized datastore that Kiali can query.
+
+In a multicluster deployment, Kiali will only support the OIDC auth strategy for now.
+
+## Deployment models
+
+In each of the deployment models, these are the requirements for a centralized Kiali
+
+1. Access to each kube API server that is part of the mesh
+2. Istiod access\*
```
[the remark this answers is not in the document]
nrfox: Yes, you're right there should at least be an asterisk to the "Regarding istiod access..." section.

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Anonymous
+- Header
+- OpenID Connect (OIDC)
+- OpenShift
+- Token
+
+of these only OpenID Connect and Header currently provide the ability to perform cross cluster authentication. The Token scheme is tied to a single cluster since it uses service account tokens tied to service accounts in a single cluster. The openshift strategy does not support multiple clusters.
+
+The security model Kiali has today assumes a namespace boundary scoped to a single cluster. Of the different authentication options that Kiali supports, RBAC would work cross cluster if RBAC was managed by an external provider that spanned across the clusters. This requires an admin to properly setup the OIDC provider across clusters.
```
jshaughn, line 94:
```suggestion
The security model Kiali has today assumes a namespace boundary scoped to a single cluster. Of the different authentication options that Kiali supports, RBAC would work cross-cluster if RBAC was managed by an external provider that spanned across the clusters. This requires an admin to properly setup the OIDC provider across clusters.
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+The namespace service is used to determine if a user has access to resources in that namespace: https://github.com/kiali/kiali/blob/master/business/namespaces.go#L64. For multi-cluster, Kiali will check the user's permissions for that namespace across all clusters. As an optimization, the namespace service can accept a list of specific clusters to query against.
+
+## API
+
+Kiali’s current model of a Service/Workload/App (s/w/a) uses a namespace/name to uniquely identify any particular s/w/a e.g. a “reviews” service in namespace “bookinfo” is uniquely identified as “reviews/bookinfo”. This model works fine for a single cluster deployment but does not uniquely identify a service in multi-cluster. Services and workloads can span multiple clusters.
```
jshaughn, line 100:
```suggestion
Kiali’s current model of a Service/Workload/App (s/w/a) uses namespace/name to uniquely identify a particular s/w/a.  For example, a “reviews” service in namespace “bookinfo” is uniquely identified as “reviews/bookinfo”. This model works fine for a single cluster deployment but does not uniquely identify a service in multi-cluster. Services and workloads can span multiple clusters.
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+Exceptions to this are:
+
+1. [Cluster local services](https://istio.io/latest/docs/ops/configuration/traffic-management/multicluster/#keeping-traffic-in-cluster) configured either locally or globally for a particular service.
+2. [Workloads partitioned](https://istio.io/latest/docs/ops/configuration/traffic-management/multicluster/#partitioning-services) by destination rule / Virtual service.
+
+The API routes to a particular s/w/a based on the namespace/name e.g. `/api/namespaces/{namespace}/workloads/{workload}`. These namespace specific routes will stay the same but they will include results from all clusters. e.g. given a `bookinfo/reivews` service in cluster A and cluster B, `/api/namespaces/bookinfo/services/reviews` would return info about a reviews service from both cluster A and B.
```
jshaughn, line 109:
```suggestion
The API routes to a particular s/w/a based on the namespace/name e.g. `/api/namespaces/{namespace}/workloads/{workload}`. These namespace specific routes will stay the same but they will include results from all clusters. e.g. given a `bookinfo/reviews` service in cluster A and cluster B, `/api/namespaces/bookinfo/services/reviews` would return info about a reviews service from both cluster A and B.
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+Some services/workloads may span across clusters but some may not. Business layer services can optionally accept a list of specific clusters to query against for a given s/w/a as an optimization. The business services will need to handle this aggregation properly and cannot assume that s/w/a will exist in all clusters or only in a single cluster.
+
+## UI/UX
+
+This proposal focuses mostly on the backend changes necessary to enable a multi-cluster deployment of Kiali however some UI elements can and should change with a multi-cluster deployment. Exact details of UI changes will be left to future issues.
+
+## Configuration
+
+To deploy a multi-cluster Kiali, users are expected to configure Kiali with access to:
```
jshaughn, line 121:
```suggestion
To deploy Kiali MC, users are expected to configure Kiali with access to:
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+- Centralized prometheus
```
jshaughn, line 123:
```suggestion
- Unified Prometheus
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+- Centralized Jaeger
```
jshaughn, line 124:
```suggestion
- Unified Jaeger
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+- 1..N Kube API servers where N is the number of kube clusters in the mesh. Kiali will require kube api credentials to communicate with each of these kube api servers.
```
jshaughn, line 125:
```suggestion
- 1..N Kube API servers where N is the number of Kube clusters in the mesh. Kiali will require Kube API credentials to communicate with each of these Kube API servers.
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+This process will need to be repeated for every new cluster added. Adding or removing clusters will also require restarting the Kiali pod in order for the cache watches to pick up the new kube config. Adding/removing kube clusters without having to restart Kiali can be an improvement made later.
+
+## Operator
+
+The Kiali operator will still be used for configuring Kiali itself however, it's assumed that the operator won't have access to other kube clusters other than where Kiali is deployed and therefore configuring multicluster access for Kiali will be a manual operation. The operator will remain unchanged in a multi-cluster deployment.
+
+## Testing
+
+All automated e2e testing, API and cypress, is currently performed against a single cluster in CI. There will be a large number of Kiali features related to multi-cluster so there needs to be a way to test multi-cluster deployments frequently and in an automated fashion. Any e2e tests that require a multi-cluster environment will need to be run separately from the rest of the suite.
```
jshaughn, line 141:
```suggestion
All automated e2e testing, API and Cypress, is currently performed against a single cluster in CI. There will be a large number of Kiali features related to multi-cluster so there needs to be a way to test multi-cluster deployments frequently and in an automated fashion. Any e2e tests that require a multi-cluster environment will need to be run separately from the rest of the suite.
```

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Kiali Agent
+
+A process that runs on each kube cluster that is a part of the mesh and exports Kiali’s view of that mesh to a centralized Kiali. This option offers potential performance gains and ease of configuration since Kiali no longer has to communicate with each individual kube cluster. However, it complicates Kiali’s ability to perform authorization with RBAC since the agents would be exporting data to a centralized Kiali and the centralized Kiali would not have access to the individual clusters. The agent itself adds an additional component to Kiali that would need to be maintained and configured across clusters.
+
+### Connecting directly to istiod
+
+The istiods provide istio config and some information about the mesh workloads. However, the endpoints are not always exposed outside of the istiod pod. Kiali today uses port forwarding through the kubernetes API to connect to the istiod pods and this would require access to the kube API anyways. Some other features rely on the kube API like streaming logs and any create/update operations need to use the Kube API ultimately making this an un-viable option to keep most of the existing Kiali features.
```
jshaughn, line 151:
```suggestion
The istiods provide istio config and some information about the mesh workloads. However, the endpoints are not always exposed outside of the istiod pod. Kiali today uses port forwarding through the Kubernetes API to connect to the istiod pods and thus requires access to the kube API . Some other features rely on the Kube API like streaming logs and any create/update operations need to use the Kube API ultimately making this an un-viable option to keep most of the existing Kiali features.
```

jshaughn (commented):
Some comments...

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+# Solution
+
+A single Kiali configured to access each Kubernetes cluster that is part of the mesh. Kiali will require Metrics/traces to be aggregated in a centralized datastore that Kiali can query.
+
+In a multicluster deployment, Kiali will only support the OIDC auth strategy for now.
+
+## Deployment models
+
+In each of the deployment models, these are the requirements for a centralized Kiali
```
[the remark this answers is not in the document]
nrfox: Establishing a more definitive name would be helpful. Istio has something similar with the "deployments models" e.g. "primary-remote". I went with "centralized" because I thought it somewhat described how Kiali would be deployed with one centralized Kiali. It is a vague term though. Maybe there's something less vague than "centralized" but more descriptive of the deployment model than "Multi-Cluster"?

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+
+Regarding istiod access, if the istiod pods are running within the kube clusters that are part of the multi-cluster mesh, Kiali will access these by portforwarding using the kube API. Kiali won't support controlplanes that are completely external to the mesh but if the istiod requirement becomes optional in a future version of Kiali then Kiali could support external controlplanes as well.
+
+A service account in each cluster will provide Kiali with access to that cluster's kube API. Credentials for that service account will be stored in a secret and used by the Kiali pod to communicate with the external Kube apis.
+
+Note that a single or different network topology won't affect Kiali's configuration and deployment since Kiali will only need access to the Kubernetes API.
+
+### Multi Primary deployment
+
+![Multi-Primary](Multi-Primary.png "Multi Primary")
```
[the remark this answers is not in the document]
nrfox: Yes perhaps it's even worth having a section dedicated to what's possible today with multi-cluster where Kiali can link to another Kiali. Without aggregated metrics you'd end up with the same setup that's possible today where one Kiali links to the other.

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,160 @@
...
+  namespace: istio-system
+spec:
+  tracing:
+  - customTags:
+      cluster:
+        literal:
+          value: Kubernetes
+```
+
+[query_scope](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.prometheus.query_scope) can be used to limit what Kiali sees from the prometheus and jaeger queries to include only workloads/services within the mesh if these datastores are shared with non-mesh workloads.
```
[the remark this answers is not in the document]
nrfox: Limiting the scope of prom/jaeger will work for multiple meshes but limiting the kube objects that Kiali pulls isn't supported yet. Would it be confusing to mention multiple meshes here since that isn't part of this proposal?

design/KEPS/multicluster/proposal.md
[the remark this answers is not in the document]
jshaughn: I don't have a better name.  I had thought maybe we would mimic Istio and could use Kiali Multi-Primary and Kiali Primary-Remote, but that may be confusing because we're not realy talking about the same thing, as we may apply this same deployment model in both of those istio models.  We really are talking about a single Kiali communicating with multiple Kube clusters, and so Kiali Multi-Cluster was the best thing I could think of.

I don't like "centralized" because it's vague and it possibly infers some sort of physical location, when that's not really the case.

design/KEPS/multicluster/proposal.md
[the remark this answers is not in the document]
jshaughn: Maybe what I wrote wasn't clear.  The proposal is for Kiali to have a view into an entire [single] mesh, where that mesh may be deployed across multiple clusters.  But we are also requiring unified metrics and traces.  What I don't think we should do is require that the unified metric and/or trace DB be limited to the one mesh for that Kiali.  Rather, a user could unify metrics or traces from multiple meshes, and possibly have a Kiali for each mesh.  But those Kiali's should not query cross-mesh.  In that case the users would have to label each metric or trace with a mesh_id, and use query_scope to limit it's queries to the mesh for that Kiali instance.

design/KEPS/multicluster/proposal.md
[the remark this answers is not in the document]
nrfox: Someone could deploy multiple Kialis for greater redundancy across clusters but they would be independent of one another and would need to be configured separately. This is also not a requirement for Kiali MC. 

> I don't like "centralized" because it's vague and it possibly infers some sort of physical location, when that's not really the case.

"Single pane of glass" comes to mind but that's not very succinct and is probably more vague than "centralized".

Outside of this proposal, I'd prefer the more verbose Kiali Multi-Cluster over Kiali MC.

design/KEPS/multicluster/proposal.md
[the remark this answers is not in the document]
nrfox: >  In that case the users would have to label each metric or trace with a mesh_id, and use query_scope to limit it's queries to the mesh for that Kiali instance.

This will work for metrics/traces where you have data from multiple meshes in your unified store but if those meshes exist on the same Kube clusters that Kiali is configured to talk to, Kiali currently has no way to distinguish between objects that belong in one mesh vs. the other and adding the support isn't part of this proposal. I was just wondering if mentioning multiple meshes would be confusing if Kiali only partially (through query scope) supports multiple meshes?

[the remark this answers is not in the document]
jshaughn: Does this multi-cluster scenario differ from today when querying a Kube API?  We always include namespace in the query and two meshes don't share a namespace, right?  Maybe I'm getting confused.  Anyway, you can leave out multi-mesh if you like.

nrfox:
@Xunzhuo you might be interested in this proposal if you could please take a look when you get a chance, thanks!

Xunzhuo:
Great job @nrfox : ) I will take a review next morning, thanks for starting to raise this!

Xunzhuo (approved):
LGTM, thanks for raising this epic, which can widely broaden the scenarios of Kiali.

A follow-up suggestion, plz separate this into issues, and label the issue with frontend or backend, I can help on its implementations in backend.

Sorry for lack of attendance for the recent releases, I am busy on working other things. But I will come back soon for this epic implementations.

design/KEPS/multicluster/proposal.md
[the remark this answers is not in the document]
jmazzitelli: See PRs related to: https://github.com/kiali/kiali/issues/5738

design/KEPS/multicluster/proposal.md
```diff
@@ -0,0 +1,161 @@
...
+
+1. Creating a service account in the target cluster.
+2. Granting proper permissions for this service account.
+3. Creating a secret with the credentials for this service account in the cluster/namespace where Kiali is deployed.
+
+This process will need to be repeated for every new cluster added. Adding or removing clusters will also require restarting the Kiali pod in order for the cache watches to pick up the new kube config. Adding/removing kube clusters without having to restart Kiali can be an improvement made later.
+
+## Operator
+
+The Kiali operator will still be used for configuring Kiali itself however, it's assumed that the operator won't have access to other kube clusters other than where Kiali is deployed and therefore configuring multicluster access for Kiali will be a manual operation. The operator will remain unchanged in a multi-cluster deployment.
```
jmazzitelli, line 137:
```suggestion
## Operator

The Kiali operator will still be used for configuring Kiali itself however, it's assumed that the operator won't have access to other kube clusters other than where Kiali is deployed and therefore configuring multicluster access for Kiali will be a manual operation. The operator will remain unchanged in a multi-cluster deployment.

In order to configure the Kiali Server for multicluster access, you must manually create remote cluster secrets in the same namepace where the Kiali Server will be deployed (normally this is the control plane namespace, aka `istio-system`, but that is not required). As a convenience, you can create these remote cluster secrets by using `istioctl`. For example, suppose you have all your remote cluster credentials defined in local kube contexts (e.g. contexts named remote1 and remote2). You can create these remote cluster secrets via:

`istioctl x create-remote-secret --name remote-cluster-1-name --context remote1 --namespace istio-system`
and
`istioctl x create-remote-secret --name remote-cluster-2-name --context remote2 --namespace istio-system`

By default, the Kiali Operator will autodiscover all remote cluster secrets found in the Kiali deployment namespace and will mount them to the Kiali Server pod at install time. When the Kiali Server starts up, it will obtain all the necessary remote cluster connection information from the mounted secrets.

jmazzitelli (changes requested):

nrfox:
> A follow-up suggestion, plz separate this into issues, and label the issue with frontend or backend, I can help on its implementations in backend.

Thanks for taking a look @Xunzhuo! The work to implement is being tracked [here](https://github.com/kiali/kiali/issues/5618). Most of the focus is on the backend right now.

jshaughn (approved):
Looks good as a checkpoint.  I think we should merge and then we can make more additions.

jmazzitelli:
we can merge this and I'll then create a new PR with additions to talk about the requirement for the remote cluster secret stuff.

jmazzitelli (approved):

Issue #5669, opened 2022-12-13
Kiali may prevent istiod from becoming ready on initial startup of istiod pod


nrfox:
Kiali port forwards to the istiod pod in order to get info from the debug endpoints. When Kiali is already running and an istiod pod first comes online, Kiali may flood the istiod pod with requests before the istiod pod can become ready.

luksa:
This is the Kiali version that display this issue: Kiali: Version: 1.48.3, Commit: 269fa3d1352c7e39432fbfd9322df7f769d7396f

jshaughn:
I would expect that this problem extends to all recent versions.  Any ideas of a way to mitigate the problem?  A way to throttle requests, or wait for readiness?

nrfox:
> Any ideas of a way to mitigate the problem? A way to throttle requests, or wait for readiness?

There needs to be some form of backoff for port forwarding requests. Since the `KialiCache` is the only resource that is shared across requests, I think it needs to happen there in the registry cache and in the proxy cache since those are the two services that do the port forwarding.

luksa:
FYI: the mesh where this issue happened had ~30 member namespaces, with 101 pods. I see that Kiali is performing a separate port-forward for every pod. That explains the extraordinarily high number of port-forward requests.

I've also noticed a problem in the `GetPodProxyStatus()` function (https://github.com/kiali/kiali/blob/34202811d72b0a6a0579cd7e7476a3355626cced/business/proxy_status.go#L17). If the cache is stale, it tries to get the status via the port-forward, but if that fails, it doesn't update the cache in any way, which means that the next call to the function will again bypass the cache. So, if the cache is stale, each invocation of `GetPodProxyStatus` results in a request to istiod.

I see multiple such requests for the same pod in the same second. If there are a large number of pods in the system, this leads to a very large of requests being performed.

nrfox:
> If the cache is stale, it tries to get the status via the port-forward, but if that fails, it doesn't update the cache in any way, which means that the next call to the function will again bypass the cache. So, if the cache is stale, each invocation of `GetPodProxyStatus` results in a request to istiod.

[...]

PR #5674, opened 2022-12-16
Poll for proxy status continually rather than fetching lazily

nrfox:
Helps mitigate #5669 by fetching proxy status continually every 10 seconds rather than fetching it only when it's asked for. This will limit the amount of times the proxy status service will attempt to port forward to istiod.

### DIFF
--- a/kubernetes/cache/proxy_status.go
+++ b/kubernetes/cache/proxy_status.go
@@ -17,18 +17,19 @@ type ProxyStatusCache interface {
 
 // pollIstiodForProxyStatus is a long running goroutine that will periodically poll istiod for proxy status.
 // Polling stops when the stopCacheChan is closed.
-func (c *kialiCacheImpl) pollIstiodForProxyStatus() {
-	log.Debug("[Kiali Cache] Scraping istiod for proxy status")
+func (c *kialiCacheImpl) pollIstiodForProxyStatus(ctx context.Context) {
+	log.Debug("[Kiali Cache] Starting polling istiod for proxy status")
 	go func() {
 		for {
 			select {
-			case <-c.stopCacheChan:
+			case <-ctx.Done():
+				log.Debug("[Kiali Cache] Stopping polling for istiod proxy status")
 				return
 			case <-time.After(c.tokenNamespaceDuration):
 				// Get the proxy status from istiod with some retries.
 				// Wrapping this so we can defer cancel.
 				func() {
-					ctx, cancel := context.WithTimeout(context.Background(), c.tokenNamespaceDuration)
+					ctx, cancel := context.WithTimeout(ctx, c.tokenNamespaceDuration)
 					defer cancel()
 
 					var (
@@ -36,7 +37,8 @@ func (c *kialiCacheImpl) pollIstiodForProxyStatus() {
 						err         error
 					)
 
-					retryErr := wait.PollImmediateUntilWithContext(ctx, time.Second*4, func(ctx context.Context) (bool, error) {
+					interval := c.tokenNamespaceDuration / 2
+					retryErr := wait.PollImmediateUntilWithContext(ctx, interval, func(ctx context.Context) (bool, error) {
 						log.Trace("Getting proxy status from istiod")
 						proxyStatus, err = c.istioClient.GetProxyStatus()
 						if err != nil {
@@ -47,7 +49,7 @@ func (c *kialiCacheImpl) pollIstiodForProxyStatus() {
 						return true, nil
 					})
 					if retryErr != nil {
-						log.Errorf("Error getting proxy status from istiod: %v", err)
+						log.Warningf("Error getting proxy status from istiod. Proxy status may be stale. Err: %v", err)
 						return
 					}
 
--- a/kubernetes/cache/proxy_status.go
+++ b/kubernetes/cache/proxy_status.go
@@ -1,31 +1,61 @@
 package cache
 
 import (
+	"context"
 	"strings"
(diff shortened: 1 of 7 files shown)


kubernetes/cache/proxy_status.go
```diff
@@ -1,31 +1,61 @@
...
+				func() {
+					ctx, cancel := context.WithTimeout(context.Background(), c.tokenNamespaceDuration)
+					defer cancel()
+
+					var (
+						proxyStatus []*kubernetes.ProxyStatus
+						err         error
+					)
+
+					retryErr := wait.PollImmediateUntilWithContext(ctx, time.Second*4, func(ctx context.Context) (bool, error) {
```
jshaughn, line 39: The fixed 4s is probably OK because the 10s setting for tokenNamespaceDuration is probably not going to change, but if we wanted to ensure at least 1 retry, or say limit to 2 retries, then the retry time should be based on the config setting.

jshaughn (commented):
This is an OK approach, I guess.  Although now I suppose it's possible we hit istiod unnecessarily when no one is even using Kiali?   Wondering if maybe there is some in-between where we guarantee some time between requests, but don't query for no reason.

nrfox:
> This is an OK approach, I guess. Although now I suppose it's possible we hit istiod unnecessarily when no one is even using Kiali? Wondering if maybe there is some in-between where we guarantee some time between requests, but don't query for no reason.

Yes Kiali will now scrape the istiod endpoint even if nobody is actively using Kiali. This is also how Kiali communicates with the Kube API through the lists/watches from the cache although that's probably more optimized for this scenario than the istiod debug endpoints. It has the benefit of the first load being faster rather than having to wait on the data to be populated before a response is returned.

> Wondering if maybe there is some in-between where we guarantee some time between requests, but don't query for no reason.

Couple of different approaches:

1. Keeping a timer in the cache on the last request made and only firing a new request when `last request > token namespace duration`. We'd probably need a separate timer for each endpoint kiali scrapes.
2. Request pool. Basically the same approach as the first with the timer state or some request counter moving into a the request pool go routine.

Open to other ideas. Whatever is gating the requests can't live in the business layer since a new business layer is created per request/token.

kubernetes/cache/proxy_status.go
nrfox: Updated this to be 1/2 `tokenNamespaceDuration`

hack/istio/download-istio.sh
```diff
@@ -65,9 +65,16 @@ echo "Output Directory: ${OUTPUT_DIR}"
 if [ -z "${ISTIO_VERSION}" ]; then
   if [ -z "${DEV_ISTIO_VERSION}" ]; then
     # get the latest released version
-    VERSION_WE_WANT=$(curl -L -s https://api.github.com/repos/istio/istio/releases | \
-          grep tag_name | sed -e 's/.*://' -e 's/ *"//' -e 's/",//' | \
-          grep -v -E "(snapshot|alpha|beta|rc)\.[0-9]$" | sort -t"." -k 1.2g,1 -k 2g,2 -k 3g | tail -n 1)
+    LATEST_RESPONSE=$(curl -L -s https://api.github.com/repos/istio/istio/releases/latest)
```
jshaughn, line 68: Not sure why this file is updated in this PR.  Did you mean for it to be updated here?  Also, are you sure "latest" is what we want?  I thought we wanted "highest".  Can't "latest" be some z-stream release for an older version?

nrfox: > Not sure why this file is updated in this PR. Did you mean for it to be updated here?

There were a number of opaque CI failures that originated from this file and talking to the github API. The current implementation hides any error that occurs and just results in `VERSION_WE_WANT` being empty which causes errors later in the script.

> Also, are you sure "latest" is what we want? I thought we wanted "highest". Can't "latest" be some z-stream release for an older version?

Yes you're right we don't want "latest" but rather "highest". It looks like the intermittent github api issues have passed and the frontend tests are running again. I'll revert this change and open up a separate PR for it.

jshaughn (commented):
One typo change requested, and I had a comment about including a hack script change in this PR.  Other than that this looks good.   I'm not sure this is the final solution with respect to how we interact with the istiod API, but it should relieve the proxy-status "flood" scenario.  It's possible now that a client may get back a fast error, or a slightly stale value, but I think both are acceptable.  We will poll the Istio API even when Kiali has no users, but it likely won't be an issue, and we can always enhance this logic if necessary.

nrfox:
I want to say the failure is a flake. I can't reproduce it locally and from the screenshot it seems like it's trying to find a link within a kebab menu which is notoriously flaky.

Kiali Overview page -- The Istio panel should be visible in the control panel (failed)

I'll re-run the tests again to see if it passes.

jshaughn (approved):
The CI failure is odd in that it wasn't really failing prior to this PR but that doesn't mean it's not a flake.  It does look tangential so let's create a new issue for the test failure.  We may need to prioritize that failure because we have stabilized the cypress CI and don't want it to regress.

nrfox:
> The CI failure is odd in that it wasn't really failing prior to this PR but that doesn't mean it's not a flake. It does look tangential so let's create a new issue for the test failure. We may need to prioritize that failure because we have stabilized the cypress CI and don't want it to regress.

I found at least one other instance of it on a recent run in a different change: https://github.com/kiali/kiali/actions/runs/3702399185/jobs/6272750167 so I'm comfortable saying it's a flake. Will create a separate issue for it.

PR #5677, opened 2022-12-22
Request Mirror in K8s Traffic routing

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5502

Added new filter type "requestMirror" which selects service name and port to where mirror the requests without waiting for response.

Screenshot from 2022-12-23 16-38-29
Screenshot from 2022-12-23 16-38-50


israel-hdez:
Just FYI: there seems to be a problem with the _Add K8s API Gateway_ option in the advanced options pane:
* Switch it on and click "Preview". All OK here.
* Cancel the preview. Then Switch it off and click again "Preview". This time Kiali crashes.

israel-hdez:
Should form fields being validated?

I'm asking because when choosing `requestMirror` it is possible to click _Add filter_ and then _Add Route Rule_ without selecting the service/port and this leads to an invalid resource having, for example (the YAML is partial):

```
spec:
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: reviews
      port: 9080
      weight: 100
    filters:
    - requestMirror:
        backendRef:
          group: ""
          kind: Service
          name: requestMirror
      type: RequestMirror
```

where `spec.rules[0].filters[0].requestMirror.backendRef.name` is `requestMirror` which does not exist. However, I'm seeing the other two options also let you add rules with empty values. So, I'm not sure if correctness of data is left to the user, or if it should be validated by the wizards.

hhovsepy:
@israel-hdez thanks for the review, it is ready now.

israel-hdez (approved):
It is working OK, and code is good.
I approve!

PR #5679, opened 2022-12-23
Remove support for OpenID implicit flow

### LINKED ISSUE #4705: Deprecate and remove "implicit" oauth flow
Kiali currently supports both ["implicit"](https://openid.net/specs/openid-connect-core-1_0.html#ImplicitFlowAuth) and ["authorization code"](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth) oauth flows. There are security benefits to using the authorization code flow vs. the implicit flow. In addition, moving to the auth code flow would allow Kiali to take advantage of the widely used [go-oidc](https://github.com/coreos/go-oidc) lib.

Pros:
- Improved security
- Less to maintain
- Less to test

Cons:
- Would break providers who only support the "implicit" flow. (I'm not sure these exist)
- May require additional configuration for existing open id users

israel-hdez:
There are security benefits to using the authorization code flow vs. the implicit flow. This is removing the weaker authentication method.

Fixes #4705


jmazzitelli:
I suspect this may break the openid molecule test. Run this to see if the test passes:

```
hack/k8s-minikube.sh start --dex-enabled true
hack/k8s-minikube.sh istio
hack/run-molecule-tests.sh -dorp podman -ce $(which kubectl) -ct minikube -at "openid-test"
```

israel-hdez:
> I suspect this may break the openid molecule test.

OK, it is broken. Although, I think it is broken "OK" with this message:

```
TASK [Assert that we can access Kiali console login screen that does not need authentication] ***
[32mok: [localhost][0m

TASK [Try to access Kiali api endpoint that requires authentication (should return 401)] ***
[32mok: [localhost][0m

TASK [Attempt login with invalid credentials] **********************************
[1;35m[WARNING]: Module did not set no_log for password[0m
[31mfatal: [localhost]: FAILED! => {"changed": false, "connection": "close", "content": "{\"error\":\"support for OpenID's implicit flow has been removed\"}", "content_length": "63", "content_type": "application/json", "date": "Tue, 10 Jan 2023 20:02:04 GMT", "elapsed": 0, "json": {"error": "support for OpenID's implicit flow has been removed"}, "msg": "Status code was 500 and not [400]: HTTP Error 500: Internal Server Error", "redirected": false, "status": 500, "url": "https://192.168.59.107/kiali/api/authenticate", "vary": "Accept-Encoding"}[0m
```

I'll check how to fix it.

israel-hdez:
Molecule fix: https://github.com/kiali/kiali-operator/pull/606

israel-hdez:
I've tested and been using this changes for a little bit and I haven't seen any issue.
Units are passing, molecules are fixed and passing, cypress doesn't seem to have an issue.

So, moving out of draft.

jmazzitelli:
openid-test molecule test was successful. Here's how I ran the test:

1. Pull this PR and the [operator PR](https://github.com/kiali/kiali-operator/pull/606)
2. Setup the cluster: `./hack/k8s-minikube.sh -mp ci start --dex-enabled true && ./hack/k8s-minikube.sh -mp ci istio && make -e CLUSTER_TYPE=minikube -e MINIKUBE_PROFILE=ci clean build build-ui cluster-push`
3. Run the molecule test: `hack/run-molecule-tests.sh -dorp podman -ce $(which kubectl) -mp ci -ct minikube -udi true -at "openid-test"`

See that the results are successful:

```
=====================
=== TEST RESULTS: ===
=====================

                             openid-test... success [2m 48s]
```

jmazzitelli (approved):

israel-hdez:
Mazz has approved. @nrfox Do you want to review, or can I go ahead and merge?

business/authentication/openid_auth_controller.go
```diff
@@ -137,41 +134,15 @@ func NewOpenIdAuthController(persistor SessionPersistor, businessInstantiator fu
...
-// Authenticate is the entry point to handle OpenId authentication using the implicit flow. The HTTP request
-// should contain "id_token" and "state" as URL parameters. If RBAC is enabled, the id_token should be
-// valid to be used in the Kubernetes API (thus, privileges are verified to allow login); else, only token
-// validity is checked and users will share the same privileges.
-// An AuthenticationFailureError is returned if the authentication failed. Any
-// other kind of error means that something unexpected happened.
+// Authenticate was the entry point to handle OpenId authentication using the implicit flow. Support
+// for the implicit flow has been removed. This is left here, because the "Authenticate" function is required
+// by the AuthController interface which must be implemented by all auth controllers. So, this simply
+// returns an error.
```
nrfox, line 140: Auth is now just handled by providing a redirect url?

business/authentication/openid_auth_controller.go
```diff
@@ -335,6 +306,7 @@ func (c OpenIdAuthController) authenticateWithAuthorizationCodeFlow(r *http.Requ
 			if flow.ShouldTerminateSession {
 				c.SessionStore.TerminateSession(r, w)
 			}
+			log.Warningf("Authentication rejected: %s", flow.Error.Error())
```
nrfox, line 309: Maybe this should be a debug log instead of a warning? Users may not want their logs flooded by login requests and I'm not sure what details or credentials the `flow.Error.Error()` message shows.

business/authentication/openid_auth_controller.go
```diff
@@ -361,6 +333,14 @@ func (c OpenIdAuthController) redirectToAuthServerHandler(w http.ResponseWriter,
 		return
 	}
 
+	// Kiali only supports the authorization code flow.
+	if !isOpenIdCodeFlowPossible() {
+		w.Header().Set("Content-Type", "text/plain")
+		w.WriteHeader(http.StatusConflict)
```
nrfox, line 339: I'm not sure this is a 409. There's no way in which the user can change the request in order to fix this. In order to fix this they would either need to change their auth strategy or move to a provider that supports the code flow right? Maybe a `500` error or 501: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501?

nrfox (commented):
Just a few small comments but no blockers

business/authentication/openid_auth_controller.go
israel-hdez: I would say "no"?
Although, I think I don't fully understand the question...

business/authentication/openid_auth_controller.go
israel-hdez: Because of #5638 you no longer will see the details of auth errors in the UI.

I realized that when an error happens, the error does not go to the logs. You also won't see it in the UI. It happened to me that I had to look at the URL to see the `Cannot view any namespaces. Please read Kiali's RBAC documentation for more details.`, which meant that my user didn't have enough privileges to login.

So, I added this to have the error in the logs. I do agree that for some cases it's better a `debug` rather than a `warning`, but there is no way to make distinction here.

Honestly, I think the specific error that made me put this should be shown at the UI, but because of #5638 and it's motivated issue we can't just be passing strings through the URL and showing them at the UI as is. So, we need a way to pass to the UI these kind of errors. That may be as simple as having in the UI a whitelist about what is safe to show... or pass error codes rather than strings.

business/authentication/openid_auth_controller.go
israel-hdez: I think I agree with changing this to `501`

israel-hdez: Fixed.

business/authentication/openid_auth_controller.go
nrfox: I was wondering why the `Authenticate` function is no longer used with just the code flow?

business/authentication/openid_auth_controller.go
nrfox: > I realized that when an error happens, the error does not go to the logs. You also won't see it in the UI. It happened to me that I had to look at the URL to see the Cannot view any namespaces. Please read Kiali's RBAC documentation for more details., which meant that my user didn't have enough privileges to login.

I've run into this issue as well. Maybe the frontend can point users to the backend logs for more details? Anyways keeping it as a warning is fine, esp. since the errors from the frontend can be cryptic.

nrfox (approved):

business/authentication/openid_auth_controller.go
israel-hdez: Well, in OpenID, we tell in the docs to configure the IdP to callback to the Kiali "web root". IIRC, this is because we started with the implicit flow.

In the implicit flow, when the IdP calls back to the "web root" the front-end can boot and dispatch the login flow (i.e. it collects callback params and fires a request to `api/authenticate` using those params). The `Authenticate` func runs when the `api/authenticate` endpoint is hit. This is done this way because the IdP will callback with _hash parameters_ which can only be read by the front-end and aren't accessible in the back-end. So, the help of the front-end was needed to pass parameters to the back-end through the right method.

Then, we implemented the code flow. I no longer remember correctly, but perhaps for consistency (and simpler docs) we chose to keep the need to configure the IdP to callback the "web root". Otherwise, people would need to know which flow will become into effect to set the right callback endpoint in the IdP. However, in the code flow there is no need to rely in the UI and the back-end can handle everything alone. So, auth is handled as soon as the IdP calls back but since it is not a callback to the `api/authenticate` the go-router won't call the `Authenticate` function. For this, the router is patched here: https://github.com/kiali/kiali/blob/f3ef5f41b9a4c45aba27a143d0173b781d4980fe/routing/router.go#L102-L107.

Honestly, nowadays  I would prefer to have a dedicated callback endpoint for OpenID (and also for OpenShift). However, that would be a breaking change as everybody would need to re-configure their IdPs. So, I'm simply keeping things backwards compatible.

Not sure if this makes sense, or if I'm confusing you more :smile:

business/authentication/openid_auth_controller.go
israel-hdez: I agree! Pointing the user to the logs will give more clarity about where to look.

PR #5684, opened 2023-01-03
Specify minimun version for indirect dependency json5 

josunect:
Set to 2.2.2


josunect:
Closing as Won't Do because the affected package is not a dependency for the production runtime, it is only for test and build

PR #5690, opened 2023-01-03
Use caching client for services business layer service.

nrfox:
Part of https://github.com/kiali/kiali/issues/5599


nrfox:
This is missing some commits from the original PR. Closing until those are copied over.

Issue #5501, opened 2022-09-27
Validations for Gateway API objects


hhovsepy:
Epic: #5026 

The Gateway API objects require validations in Kiali side:
HTTPRoute 
Gateway.

- Validations for k8s Gateways:
  - [X]  No more than one GW per host and port combination [5687](https://github.com/kiali/kiali/pull/5687)
  - [X]  No more than one GW per address [5687](https://github.com/kiali/kiali/pull/5687)
  - [X]  Highlight status rules [5687](https://github.com/kiali/kiali/pull/5687)
  - [x] No more than one GW per host, protocol and port combination per GW. https://github.com/kiali/kiali/issues/5710 
- Validations for k8s httpRoute [#5093 ](https://github.com/kiali/kiali/pull/5693)
  - [x] k8s route referring to non existent gateway https://github.com/kiali/kiali/pull/5693
  - [x] k8s route referring to non existent service https://github.com/kiali/kiali/pull/5706
- [x] References for k8s Gateway: https://github.com/kiali/kiali/issues/5708
- [x] #5709

josunect:
Docs: 

* GW API: https://gateway-api.sigs.k8s.io/concepts/api-overview/#gateway
* Istio Config analysis: https://istio.io/latest/docs/reference/config/analysis/
* Kiali validations: https://kiali.io/docs/features/validations/

### K8sGateway 
---

* More than one GW for the same addresses/Listeners combination
* No matching route found for the listener 


### HTTPRoute
---

* Not matching service found
* Not matching k8sGateway

PR #5693, opened 2023-01-05
Added K8sGateway existance checker in K8sHTTPRoute

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5501

Added new Error message "KIA1401 HTTPRoute is pointing to a non-existent K8s gateway"

Screenshot from 2023-01-05 13-36-22

Testing: Create K8sHTTPRoute with Parent Ref to non existing K8sGateway.

### DIFF
--- a/business/checkers/k8shttproutes/no_k8sgateway_checker.go
+++ b/business/checkers/k8shttproutes/no_k8sgateway_checker.go
@@ -34,7 +34,7 @@ func (s NoK8sGatewayChecker) ValidateHTTPRouteGateways(validations *[]*models.Is
 				if parentRef.Namespace != nil && string(*parentRef.Namespace) != "" {
 					namespace = string(*parentRef.Namespace)
 				}
-				valid = s.checkGateway(string(parentRef.Name), namespace, validations, fmt.Sprintf("spec/parentRefs[%d]", index)) && valid
+				valid = s.checkGateway(string(parentRef.Name), namespace, validations, fmt.Sprintf("spec/parentRefs[%d]/name/%s", index, string(parentRef.Name))) && valid
 			}
 		}
 	}


business/checkers/k8shttproutes/no_k8sgateway_checker.go
```diff
@@ -0,0 +1,55 @@
...
+
+// ValidateHTTPRouteGateways checks all HTTPRoute gateways and checks that they're found from the given list of gatewayNames. Also return index of missing gatways to show clearer error path in editor
+func (s NoK8sGatewayChecker) ValidateHTTPRouteGateways(validations *[]*models.IstioCheck) bool {
+	valid := true
+
+	if len(s.K8sHTTPRoute.Spec.ParentRefs) > 0 {
+		for index, parentRef := range s.K8sHTTPRoute.Spec.ParentRefs {
+			if string(parentRef.Name) != "" && string(*parentRef.Kind) == kubernetes.K8sActualGatewayType && string(*parentRef.Group) == kubernetes.K8sNetworkingGroupVersionV1Beta1.Group {
+				namespace := s.K8sHTTPRoute.Namespace
+				if string(*parentRef.Namespace) != "" {
```
josunect, line 34: It is possible to add an additional check? 

`if parentRef.Namespace != nil && string(*parentRef.Namespace) != "" {`

To avoid error: 

josunect (commented):
I see there is also a status validator, like the Gateway. Maybe a suggestion would be to also highlight the syntax, the same as the Gateway? But it can also be done in an additional issue. 

image

business/checkers/k8shttproutes/no_k8sgateway_checker.go
hhovsepy: good catch, thank you :+1:

business/checkers/k8shttproutes/no_k8sgateway_checker.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (s NoK8sGatewayChecker) checkGateway(name, namespace string, validations *[]*models.IstioCheck, location string) bool {
+	hostname := kubernetes.ParseGatewayAsHost(name, namespace)
+	for gw := range s.GatewayNames {
+		gwHostname := kubernetes.ParseHost(gw, namespace)
+		if found := kubernetes.FilterByHost(hostname.String(), hostname.Namespace, gw, gwHostname.Namespace); found {
+			return true
+		}
+	}
+	validation := models.Build("k8shttproutes.nok8sgateway", location)
```
josunect, line 52: Maybe we can add the Path here to highlight just one line: 
`
validation.Path = "spec/parentRefs/name/" + name`

image

(This is just a suggestion)

hhovsepy: yes, much better

josunect (approved):
LGTM

Thanks, @hhovsepy !

hhovsepy:
Thank you @josunect

PR #5698, opened 2023-01-06
Minor post-merge follow-up for kiali-5588 PR

jshaughn:
- apply standard formatting to DebugInformation.tsx
- alphabetize DebugInformationState
- ensure modal always opens with default tab
- simplify change to Tabs component, instead of adding new "skipHistory" property, make "tabName" an optional property


leandroberetta (approved):
Looks good. Tested locally and works fine.

Issue #5696, opened 2023-01-06
 Kiali errors out loading workload graph in 'default' namespace 


Anand25021998:
It’s able to load the workload graph for couple of deployments I checked in different namespaces but errors out while loading workload graph of any workload in the 'default' namespace. The only difference is that I don't have istio injected on a namespace level in the 'default' namespace and it's on workload level but to my understanding it shouldn't matter to Kiali. It's able to load all the app graphs in the 'default' namespace and the issue is seen only with workload graphs. I can’t derive much from the stack trace in pod logs which are:

```
2023-01-06T08:54:20Z ERR Failed ID gen2: cluster=[Kubernetes] namespace=[default] workload=[unknown] app=[test-app] version=[29917846] service=[] graphType=[workload]: goroutine 2741890 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.17.10/x64/src/runtime/debug/stack.go:24 +0x65
github.com/kiali/kiali/handlers.handlePanic({0x2097ea0, 0xc004ee8690})
        /home/runner/work/kiali/kiali/handlers/graph.go:86 +0x219
panic({0x1a243e0, 0xc0040fddf0})
        /opt/hostedtoolcache/go/1.17.10/x64/src/runtime/panic.go:1038 +0x215
github.com/kiali/kiali/graph.Id({0xc006fddb10, 0xc003cdb2d8}, {0xc006fddb40, 0x7f60a06cf538}, {0x0, 0x0}, {0xc006fddb40, 0x23}, {0xc006fddb07, 0x7}, ...)
        /home/runner/work/kiali/kiali/graph/types.go:213 +0xde9
github.com/kiali/kiali/graph/telemetry/istio/appender.ResponseTimeAppender.addResponseTime({{0xc00274f559, 0x8}, 0x1, 0xc00530f4d0, 0x3fee666666666666, 0x63b7e1bc, {{0x1dd45a2, 0x8}, {0x1dd45a2, 0x8}, ...}}, ...)
        /home/runner/work/kiali/kiali/graph/telemetry/istio/appender/response_time.go:229 +0xe5
github.com/kiali/kiali/graph/telemetry/istio/appender.ResponseTimeAppender.populateResponseTimeMap({{0xc00274f559, 0x8}, 0x1, 0xc00530f4d0, 0x3fee666666666666, 0x63b7e1bc, {{0x1dd45a2, 0x8}, {0x1dd45a2, 0x8}, ...}}, ...)
[...]

PR #5699, opened 2023-01-06
Avoid graph generation panic scenario

jshaughn:
It seems there is a legitimate case (possibly with a newer version of Istio, I'm not sure) where the telemetry may have app set, but not workload or service.  Possibly when the workload is not injected but app/version labeling is still applied to the workload.

In this case, avoid a graph-gen panic when trying to generate a workload graph node, and instead default to an Unknown service node.  Panics don't really help anyone, so anything we can do to avoid that situation is, I think, better.

Closes https://github.com/kiali/kiali/issues/5696

### DIFF
--- a/graph/types.go
+++ b/graph/types.go
@@ -209,11 +209,12 @@ func Id(cluster, serviceNamespace, service, workloadNamespace, workload, app, ve
 	// handle workload graph nodes (service graphs are initially processed as workload graphs)
 	if graphType == GraphTypeWorkload || graphType == GraphTypeService {
 		// workload graph nodes are type workload or service
-		if !workloadOk && !serviceOk {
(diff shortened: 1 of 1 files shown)


israel-hdez (approved):
This seems to be hard to test, as conditions to replicate are not clear.

In any case, the code is OK to me. I checked the callers, and I don't think there is any impact other than generating an "unknown" triangle, which is much better than a panic.

I didn't review in deep the callers because the code is large. So perhaps, what is unknown is what the dependent code would do if the service is blank?.... probably the graph would just display a blank label (I just hope it won't crash because of the blank service name).

Anyway, approving!

jshaughn:
> This seems to be hard to test, as conditions to replicate are not clear.

Yes, it's not easy to test; it's a code path that requires specific setup in Prometheus, and has up to now been assumed to not happen, which is why we had the panic.  It will be easier for the affected user to validate the change.

Issue #5689, opened 2023-01-03
[CI] Cypress flake - Overview istio panel visible


nrfox:
A cypress test is flaking for the overview page: "The Istio panel should be visible in the control panel". It has failed on a couple different unrelated test runs: https://github.com/kiali/kiali/actions/runs/3760564732/jobs/6391515970 and https://github.com/kiali/kiali/actions/runs/3702399185/jobs/6272750167.

Screenshot from failure:
![Kiali Overview page -- The Istio panel should be visible in the control panel (failed)](https://user-images.githubusercontent.com/6226732/210405566-a05baeee-a4f3-4fb4-8e57-15fccfcc0312.png)

The test appears to be searching for a link within a kebab menu which is a notoriously flaky operation as the menus sometimes disappear after they are clicked but before the element can be found.

PR #5703, opened 2023-01-09
Fix flaky test in the Overview page

leandroberetta:
Fixes https://github.com/kiali/kiali/issues/5689

### DIFF
--- a/frontend/cypress/integration/featureFiles/overview.feature
+++ b/frontend/cypress/integration/featureFiles/overview.feature
@@ -120,5 +120,5 @@ Feature: Kiali Overview page
     Then user sees the "istio-system" namespace card
     And user sees the "Control plane" label in the "istio-system" namespace card
     And user sees the "Outbound policy" label in the "istio-system" namespace card
-    When user clicks the toggle on the right side of the "istio-system" namespace card
-    Then user can see links to external services
+    Then the toggle on the right side of the "istio-system" namespace card exists
+    
--- a/frontend/cypress/integration/featureFiles/overview.feature
+++ b/frontend/cypress/integration/featureFiles/overview.feature
@@ -121,4 +121,3 @@ Feature: Kiali Overview page
     And user sees the "Control plane" label in the "istio-system" namespace card
     And user sees the "Outbound policy" label in the "istio-system" namespace card
     Then the toggle on the right side of the "istio-system" namespace card exists
-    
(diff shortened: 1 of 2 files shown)


frontend/cypress/integration/featureFiles/overview.feature
```diff
@@ -120,5 +120,5 @@ Feature: Kiali Overview page
     Then user sees the "istio-system" namespace card
     And user sees the "Control plane" label in the "istio-system" namespace card
     And user sees the "Outbound policy" label in the "istio-system" namespace card
-    When user clicks the toggle on the right side of the "istio-system" namespace card
-    Then user can see links to external services
```
nrfox, line 124: This fix is now just testing the kebab exists instead of clicking on it and ensuring the link exists? Fine by me but tagging @matejnesuta since I think he initially added the test. Is the `Then user can see links to external services` definition used anywhere else or should that be removed as well?

frontend/cypress/integration/featureFiles/overview.feature
```diff
@@ -120,5 +120,5 @@ Feature: Kiali Overview page
...
+    Then the toggle on the right side of the "istio-system" namespace card exists
+
```
nrfox, line 124:
```suggestion
    Then the toggle on the right side of the "istio-system" namespace card exists

```

frontend/cypress/integration/featureFiles/overview.feature
leandroberetta: It simplifies the test to avoid the flaky. I talked with @jshaughn and this approach was done in other test I guess. Having said that, sure, I also waits for @matejnesuta's opinion. No problem.

leandroberetta: Hi, @matejnesuta do you have any comments about this?.

matejnesuta: Hi @leandroberetta. Sorry for the late reply, I was out due to my final exams.

Anyways. As Nick mentioned, the original idea was to test that the menu has some additional links, compared to other menus. That seemed to work originally, but looking at it now, it might've not been the best idea. I am honestly not sure, if it's necessary to check the existence of the kebab only. It seems kind of pointless to me, but I suppose you might see some benefits in doing that and it is only one assertion anyway.

nrfox: In the past we've had trouble getting kebab tests to work. My guess is that after you click the kebab there's a refresh that happens _before_ the links are found on the screen which causes the kebab menu to close and the links to not be found.

Anyways some tests that are reliable are better than more tests that are unreliable so I'd prefer we remove the kebab link assertions for now until we can figure out how to test those reliably.

PR #5706, opened 2023-01-10
Added new check for HTTPRoute backend Ref Service existence.

hhovsepy:
RFE: https://github.com/kiali/kiali/issues/5501
Added new validation check and message when K8s HTTPRoute is referring to a non existing Service.

Screenshot from 2023-01-10 15-43-30

### DIFF
--- a/business/checkers/k8shttproutes/no_host_checker.go
+++ b/business/checkers/k8shttproutes/no_host_checker.go
@@ -1,0 +1,45 @@
+package k8shttproutes
+
+import (
+	"fmt"
+
+	k8s_networking_v1alpha2 "sigs.k8s.io/gateway-api/apis/v1alpha2"
+
+	"github.com/kiali/kiali/kubernetes"
+	"github.com/kiali/kiali/models"
+)
+
+type NoHostChecker struct {
+	Namespaces       models.Namespaces
+	K8sHTTPRoute     *k8s_networking_v1alpha2.HTTPRoute
+	RegistryServices []*kubernetes.RegistryService
+}
+
+func (n NoHostChecker) Check() ([]*models.IstioCheck, bool) {
+	validations := make([]*models.IstioCheck, 0)
+	valid := true
+	namespace := n.K8sHTTPRoute.Namespace
+
+	for k, httpRoute := range n.K8sHTTPRoute.Spec.Rules {
+		for i, ref := range httpRoute.BackendRefs {
+			if ref.Kind == nil || string(*ref.Kind) != "Service" {
+				continue
+			}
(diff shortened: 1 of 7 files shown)


business/checkers/k8shttproutes/no_host_checker.go
```diff
@@ -0,0 +1,45 @@
...
+	validations := make([]*models.IstioCheck, 0)
+	valid := true
+	namespace := n.K8sHTTPRoute.Namespace
+
+	for k, httpRoute := range n.K8sHTTPRoute.Spec.Rules {
+		for i, ref := range httpRoute.BackendRefs {
+			if ref.Kind == nil || string(*ref.Kind) != "Service" {
+				continue
+			}
+			fqdn := kubernetes.GetHost(string(ref.Name), namespace, n.Namespaces.GetNames())
```
josunect, line 28: For what I can see in the Object reference (https://gateway-api.sigs.k8s.io/references/spec/#gateway.networking.k8s.io/v1beta1.BackendObjectReference) It is possible to specify a namespace for the service, as an example: 

image


So in case it exist, it should be grabbed from the spec (If not, namespace should be the default): 

image

hhovsepy: Good catch, have fixed.

business/checkers/k8shttproutes/no_host_checker.go
```diff
@@ -0,0 +1,48 @@
...
+	valid := true
+	namespace := n.K8sHTTPRoute.Namespace
+
+	for k, httpRoute := range n.K8sHTTPRoute.Spec.Rules {
+		for i, ref := range httpRoute.BackendRefs {
+			if ref.Kind == nil || string(*ref.Kind) != "Service" {
+				continue
+			}
+			if ref.Namespace != nil && string(*ref.Namespace) != "" {
+				namespace = string(*ref.Namespace)
```
josunect, line 29: I think it would be better to create a local variable inside the for loop for namespace. 
If we set namespace, and for the next iterations the namespace is not compared with the default one. See example (details service do exist in bookinfo namespace): 

image

josunect (commented):
I think it would be better to create a local variable inside the for loop for namespace.
If we set namespace, and for the next iterations the namespace is not set, it will be compared with the default one. See example (details service do exist in bookinfo namespace):

hhovsepy:
Thanks for the catch with namespace variable.

josunect (commented):
Thanks, @hhovsepy ! It looks to be fixed now.

I see a couple of frontend tests failing, I will approve when CI passes.

hhovsepy:
@josunect seems the ci failures were flaky, now they pass.

josunect (approved):
LGTM

PR #5721, opened 2023-01-14
try to hide the namespace injection menu option on OSSM

jmazzitelli:
namespace label is not currently honored in ossm environments

fixes https://github.com/kiali/kiali/issues/5725

### DIFF
--- a/frontend/src/pages/Overview/OverviewPage.tsx
+++ b/frontend/src/pages/Overview/OverviewPage.tsx
@@ -9,7 +9,8 @@ import {
   EmptyStateBody,
   EmptyStateVariant,
   Grid,
-  GridItem, Label,
+  GridItem,
+  Label,
   Title,
   TitleSizes,
   Tooltip,
@@ -32,7 +33,12 @@ import {
 } from '../../types/Health';
 import { SortField } from '../../types/SortFilters';
 import { PromisesRegistry } from '../../utils/CancelablePromises';
-import OverviewToolbarContainer, { OverviewDisplayMode, OverviewToolbar, OverviewType, DirectionType } from './OverviewToolbar';
+import OverviewToolbarContainer, {
+  OverviewDisplayMode,
+  OverviewToolbar,
+  OverviewType,
+  DirectionType
+} from './OverviewToolbar';
 import NamespaceInfo, { NamespaceStatus } from './NamespaceInfo';
 import NamespaceMTLSStatusContainer from '../../components/MTls/NamespaceMTLSStatus';
 import { RenderComponentScroll } from '../../components/Nav/Page';
@@ -65,8 +71,8 @@ import { MessageType } from '../../types/MessageCenter';
 import { CanaryUpgradeStatus, OutboundTrafficPolicy, ValidationStatus } from '../../types/IstioObjects';
 import { GrafanaInfo, ISTIO_DASHBOARDS } from '../../types/GrafanaInfo';
 import { ExternalLink } from '../../types/Dashboards';
-import { isParentKiosk, kioskOverviewAction } from "../../components/Kiosk/KioskActions";
-import ValidationSummaryLinkContainer from "../../components/Link/ValidationSummaryLink";
+import { isParentKiosk, kioskOverviewAction } from '../../components/Kiosk/KioskActions';
+import ValidationSummaryLinkContainer from '../../components/Link/ValidationSummaryLink';
 import _ from 'lodash';
 import ControlPlaneBadge from './ControlPlaneBadge';
 import OverviewStatusContainer from './OverviewStatus';
@@ -263,7 +269,7 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
               showTrafficPoliciesModal: prevState.showTrafficPoliciesModal,
               kind: prevState.kind,
               nsTarget: prevState.nsTarget,
-              opTarget: prevState.opTarget,
+              opTarget: prevState.opTarget
             };
           },
           () => {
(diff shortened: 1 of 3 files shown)


jmazzitelli:
Tested on OSSM 2.3 and this PR - this is what it looks like (notice the option to create the injection label is now hidden):
image

jmazzitelli:
I just ran this on upstream Istio 1.16 and minikube, notice this is what it looks like with the Overview page menu option looks like when it is available to enable auto injection on the namespace.

image

jmazzitelli:
This makes a change to status/version.go whereby this function no longer returns an error if the istio cluster type/version is unknown. We need to test this thoroughly to make sure this doesn't break something. We'll want to test this on known versions of Istio and OSSM as well as unknown versions of both (say, an RC).

status/versions.go
```diff
@@ -267,7 +267,7 @@ func istioVersion() (*ExternalServiceInfo, error) {
 		Put(MeshVersion, meshVersion)
 		Put(MeshName, meshName)
 		AddWarningMessages(compatibleWarnings)
-		return nil, fmt.Errorf(compatibleWarnings)
+		//return nil, fmt.Errorf(compatibleWarnings)
```
jshaughn, line 270: remove this

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -159,6 +159,7 @@ type ReduxProps = {
   refreshInterval: IntervalInMilliseconds;
   minTLS: string;
   istioAPIEnabled: boolean;
+  isMaistra: boolean;
```
jshaughn, line 162: Take this opportunity to fix the alphabetization of these props.

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -1168,6 +1169,7 @@ const mapStateToProps = (state: KialiAppState): ReduxProps => ({
   refreshInterval: refreshIntervalSelector(state),
   minTLS: minTLSVersionSelector(state),
   istioAPIEnabled: state.statusState.istioEnvironment.istioAPIEnabled,
+  isMaistra: state.statusState.istioEnvironment.isMaistra,
```
jshaughn, line 1172: Take this opportunity to fix the alphabetization of these props.

frontend/src/pages/Overview/__tests__/OverviewPage.test.tsx
```diff
@@ -101,7 +101,7 @@ describe('Overview page', () => {
 
   it('renders initial layout', () => {
     const wrapper = shallow(
-      <OverviewPage meshStatus={MTLSStatuses.NOT_ENABLED} navCollapse={false} duration={600} refreshInterval={10000} kiosk={''} minTLS={''} istioAPIEnabled={false}/>
+      <OverviewPage meshStatus={MTLSStatuses.NOT_ENABLED} navCollapse={false} duration={600} refreshInterval={10000} kiosk={''} minTLS={''} istioAPIEnabled={false} isMaistra={false}/>
```
jshaughn, line 104: You lost the separator above `Remove Traffic Policies` in the Maistra use case.

jshaughn (changes requested):

status/versions.go
jmazzitelli: well, I left that in there for a reason. This seems to me to be a big change.  This is what I was talking about when I said, `This makes a change to status/version.go whereby this function no longer returns an error if the istio cluster type/version is unknown. We need to test this thoroughly to make sure this doesn't break something.`

I left that commented out because in the future I predicate a 50/50 chance we will want to resurrect that line quickly to fix a bug we can't see right now.

frontend/src/pages/Overview/OverviewPage.tsx
jmazzitelli: done

frontend/src/pages/Overview/OverviewPage.tsx
jmazzitelli: done

frontend/src/pages/Overview/__tests__/OverviewPage.test.tsx
jmazzitelli: I don't understand what this is commenting on. This is a Cypress test and it passed when GitHub Actions ran the tests. Can you be more specific what this code here on line 104 needs to do?

jmazzitelli:
will update to get the separator line in the menu. Here's on non-OSSM:
image

here is on OSSM:
image

frontend/src/pages/Overview/__tests__/OverviewPage.test.tsx
jmazzitelli: I submitted a new commit that retains the separator line. See the latest screen captures in my latest comment.

status/versions.go
jmazzitelli: I added a comment to the comment in commit https://github.com/kiali/kiali/pull/5721/commits/e12b9b166b5d43ce2666c77657d4ea259e5feff4 that points this out

frontend/src/pages/Overview/__tests__/OverviewPage.test.tsx
jshaughn: I just added the comment in the wrong place.  As long as the separator is back that's good.

jshaughn (changes requested):
The separators still aren't right.  When isMaistra is true you end up with a separator with no options underneath:

image

I think you just want to move the `!this.props.isMaistra` test to the existing conditional at line 662

jshaughn (approved):
LG

PR #5732, opened 2023-01-18
Multicluster support for kiali cache

### LINKED ISSUE #5724: Cache supports multiple kube backends
For multicluster scenarios, the cache should support multiple kubernetes backends. Aggregation of objects across the backends should be handled in the business layer but the kiali cache needs to manage those backends and provide them to the business layer. 

#5618

nrfox:
Adds support for multiple kube backends to the kiali cache. Splits out the kubernetes part of the cache into its own implementation. Keeps a single global kiali cache. Adds a cluster client manager to create/manage clients from different clusters.

Fixes #5724


kubernetes/cache/proxy_status.go
```diff
@@ -41,7 +40,7 @@ func (c *kialiCacheImpl) pollIstiodForProxyStatus(ctx context.Context) {
 					interval := c.tokenNamespaceDuration / 2
 					retryErr := wait.PollImmediateUntilWithContext(ctx, interval, func(ctx context.Context) (bool, error) {
 						log.Trace("Getting proxy status from istiod")
-						proxyStatus, err = c.istioClient.GetProxyStatus()
+						proxyStatus, err = c.clientFactory.GetSAHomeClusterClient().GetProxyStatus()
```
leandroberetta, line 43: Just a thought, in some other places I read that eventually we will need to treat all clusters equally with no home cluster distinction but in some other places I see this method, I think in some way, we will need to define that one of them is the "home", right?. Not sure why yet but sounds logical.

leandroberetta (approved):
Nick, great work, looks good to me. I have just some comments but to me it's ready for merge.

business/workloads.go
```diff
@@ -46,14 +47,15 @@ func NewWorkloadService(k8s kubernetes.ClientInterface, prom prometheus.ClientIn
...
-	k8s  kubernetes.ClientInterface
 	// Careful not to call the workload service from here as that would be a infinite loop.
 	businessLayer *Layer
 	// The global kiali cache. This should be passed into the workload service rather than created inside of it.
-	cache cache.KialiCache
+	cache   cache.KialiCache
+	clients map[string]kubernetes.ClientInterface
 	// The global kiali config.
 	config *config.Config
+	k8s    kubernetes.ClientInterface
```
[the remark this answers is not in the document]
nrfox: Yes whenever the workload service gets updated to support multicluster, then the individual k8s client can get removed from this service. I had included some of those changes with this PR originally but removed them so that this change would mostly be about updating the cache.

kubernetes/cache/proxy_status.go
nrfox: Yeah there's a couple different services that do autodiscovery that will need to know the "home" cluster at least in addition to this case where you want to use the client of the cluster kiali is deployed in to do the polling of istiod.

nrfox:
@leandroberetta I noticed that the cache was not being refreshed when the Kiali service account token changed which is a regression. I fixed that and added a few unit tests for that scenario if you could PTAL again.

kubernetes/cache/proxy_status.go
israel-hdez, line 43: I think this is OK as we are starting with a single control plane. But as we add support to other models, this may need to evolve, as I think we won't be able to tie this proxy status cache to a cluster cache, nor the home cluster will be the only one having a control plane.

israel-hdez (commented):
Just flushing some comments, as I think I need to give a closer look to the code.

BTW, I wonder if it would be better to merge this PR after the release? It looks a little risky to merge this week because of the refactoring.

business/workloads.go
```diff
@@ -126,7 +128,6 @@ func (in *WorkloadService) getWorkloadValidations(authpolicies []*security_v1bet
 	return validations
 }
 
-// GetWorkloadList is the API handler to fetch the list of workloads in a given namespace.
```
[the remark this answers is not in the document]
nrfox: Whoops I must've removed this while refactoring by mistake. Good catch, I'll add it back in.

business/layer.go
```diff
@@ -189,17 +187,22 @@ func NewWithBackends(k8s kubernetes.ClientInterface, prom prometheus.ClientInter
 	temporaryLayer.TLS = TLSService{k8s: k8s, businessLayer: temporaryLayer}
 	temporaryLayer.TokenReview = NewTokenReview(k8s)
 	temporaryLayer.Validations = IstioValidationsService{k8s: k8s, businessLayer: temporaryLayer}
+
+	// TODO: Use client manager or passed in clients rather than hardcoding the home cluster.
+	clusterClients := map[string]kubernetes.ClientInterface{"home": k8s}
```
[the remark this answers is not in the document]
nrfox: Ideally it wouldn't be `"home"` but come from the istio config that Kiali is deployed alongside of. The other cluster names will come from the remote secret kubeconfig files. I'm unsure if these need to match with their name in their cluster's istio config but they at least need to be unique. At any rate, the `home` cluster is just the cluster that Kiali is deployed in. It's at least helpful while we are iteratively adding multi cluster support and I think it will be useful for some of the business services that gather info about the mesh and the cluster but it may not make sense to keep in the future. I'll add a global const for it.

kubernetes/cache/proxy_status.go
nrfox: Yes we'll need to scrape all of them and most likely merge the results together.

nrfox:
> BTW, I wonder if it would be better to merge this PR after the release? It looks a little risky to merge this week because of the refactoring.

Yes it'd probably be better to let this soak at the beginning of a release cycle given the large refactor. Someone else may need to merge this and carry the torch then if it will make it into the next release since I will most likely be unable to work on this during that time period.

nrfox:
I removed the workload svc changes from this PR. This change now only updates the cache. The workload svc changes couldn't be tested yet without the ability to load multiple clusters in and those changes were more half baked. I originally included those as an example for how the different business services could be changed to support multicluster but I believe they'd be better served as their own PRs.

kubernetes/cache/cache.go
```diff
@@ -30,369 +16,148 @@ import (
...
 }
 
-// refreshCache watches for changes to the cache's service account token
-// and recreates the cache(s) when the token changes.
-func (c *kialiCacheImpl) refreshCache(ctx context.Context, istioConfig rest.Config) {
-	ticker := time.NewTicker(60 * time.Second)
+// watchForClientChanges watches for changes to the cache's service account client
+// and recreates the cache(s) when the client changes. The client is updated when
+// the token for the client changes.
+func (c *kialiCacheImpl) watchForClientChanges(ctx context.Context, token string) {
```
israel-hdez, line 143: So, only the "Home cluster" client is updated. Perhaps it isn't clear in this PR, but I guess this is because the bound SA token only applies to the "local" cluster, while for remote ones we will be using standard non-expiring tokens.

Nothing to fix, just pointing out this.

kubernetes/client_factory.go
```diff
@@ -144,22 +166,35 @@ func (cf *clientFactory) newClient(authInfo *api.AuthInfo, expirationTime time.D
 	return newClient, err
 }
 
+func (cf *clientFactory) newSAClient(cluster string) (*K8SClient, error) {
+	// TODO: Need a way to load in cluster configuration from the kube secret.
```
israel-hdez, line 170: I guess this `TODO` means that this `newSAClient` func isn't really capable of creating a client for remote clusters, right?.

If so, I guess that, although we can say that the cache would be ready for MC after merging this PR, the client factory is not ready yet.

kubernetes/client_factory.go
```diff
@@ -253,3 +288,50 @@ func getTokenHash(authInfo *api.AuthInfo) string {
...
+}
+
+// Check for kiali token changes and refresh the client when it does.
+func (cf *clientFactory) refreshClientIfTokenChanged(cluster string) error {
+	kialiSAToken, err := GetKialiToken()
+	if err != nil {
+		return err
+	}
+
+	if cf.saClientEntries[cluster].GetToken() != kialiSAToken {
```
israel-hdez, line 313: I think a `TODO` is missing somewhere around this func, no?

If I understand correctly, it tries to compare the token of a remote cluster client, with the Kiali SA token (which would be for the local cluster). That's always going to be a mismatch, no? So, it will always want to recreate the client.

(this is one of the reasons why I think `clientfactory` is not ready for MC yet... although I see you have a unit test for this, so it is confusing).

israel-hdez (commented):
Just a few comments after my 2nd review round.

Most likely, nothing major. Probably, just questions that need an answer.

kubernetes/cache/cache.go
nrfox: Yeah for remote clusters, I don't think the token will expire and even if it did the auto-renew mechanism that exists for the in-cluster token wouldn't auto-renew the out of cluster tokens. That being said, it could be good to watch for client changes for the other clients anyways in case the secrets change. I think that's something we can add later though when more of the pieces have come together and we can test this in earnest and have a clearer picture of the requirements around that. For now, just watching the client changes for the home cluster keeps the existing functionality of token renewal in single cluster which is why I added this.

kubernetes/client_factory.go
nrfox: That's right. The cache will start a kube cache for each kiali service account client from the client factory but the client factory only supports a single service account client for now. It will need to be updated to read in the remote secrets from the filesystem and create a kiali service account client for each remote secret. This can happen after #5752 is merged.

kubernetes/client_factory.go
nrfox: Yeah at the very least it needs a `TODO`. It should not be comparing to the kiali token for each cluster. Rather it should be comparing the client's token with the token read from the filesystem for that cluster. The unit for the home cluster is valid but the one for multiple clients is a bit misleading since we won't want to test with kiali token but rather with the token for that cluster. I'll remove the multiple cluster test since that will change anyways once the client factory supports reading the secrets from the fs.

nrfox: Well looking at the test again it's actually testing two different methods. One test is for `GetSAHomeClusterClient` and the other test is for `GetSAClients`. In both cases we want to test that you get an updated client back in return but I'll update the test for `GetSAClients` so it's more clear.

israel-hdez (approved):
OK, I _think_ that I understand all changes. So, I'm going to approve, although I haven't tried the code.

I'll try it tomorrow. If I find something wrong, I'll remove my approval. Otherwise, you already have my green tick mark :-)

leandroberetta (approved):
Second review, still looks good to me.

matejnesuta:
Hello. Can this possibly be tested with the integration suite as well? I am trying to find spots where the coverage can be increased and this looks promising, but I am not 100% sure.

jshaughn:
> Hello. Can this possibly be tested with the integration suite as well? I am trying to find spots where the coverage can be increased and this looks promising, but I am not 100% sure.

The cache use is accessed via almost all code-paths, so there should be good coverage given existing and future tests.  I'm not sure that anything needs to be added here explicitly, but I'll let @nrfox comment when he returns from leave.

PR #5740, opened 2023-01-20
Add date to chart tooltips

### LINKED ISSUE #5320: Add dates in chart tooltips
Charts have tooltips showing the time of each datapoint:

image

However, like in the previous image, when the data of the resulting chart covers more than one day, it is not possible to distinguish where each data point lives, making it harder to diagnose problems (i.e. it is not possible to manually adjust the duration to "zoom-in").

Perhaps, in the _Dashboards_ the x-axis labels can be used to find the day. However, all Kiali charts are affected. So, for example, the graph side-panel and the overview page charts don't offer x-axis legends.

In the given example the chosen duration guarantees that the resulting charts will span several days. However, even shorter durations can potentially lead to charts covering more than one day. Probably, dates should be shown *only* if the data points span over more than one day.

israel-hdez:
Fixes #5320

### DIFF
--- a/frontend/src/components/Charts/CustomTooltip.tsx
+++ b/frontend/src/components/Charts/CustomTooltip.tsx
@@ -46,7 +46,7 @@ const getHeader = (activePoints?: VCDataPoint[]): string | undefined => {
     const x = activePoints[0].x;
     if (typeof x === 'object') {
       // Assume date
-      return x.toLocaleString();
+      return x.toLocaleStringWithConditionalDate();
     }
   }
   return undefined;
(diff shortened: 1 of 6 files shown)


israel-hdez:
The only problematic component I see are the sparklines of the graph's side panels:

image

It is a problem even in `master` although adding the date is making it worse.
Perhaps a naive workaround is to have the date on its own line, then the time under it. Trying to do the right positioning so that the tooltip does not go out of the viewport is, apparently, not trivial :-/

jshaughn (commented):
I think because the date makes the tooltip so wide, that we should consider only adding the date when it differs from the current date.

israel-hdez:
@jshaughn There you go. I just added a commit that should be doing what you suggested.

jshaughn (approved):
LG2M!

Issue #5738, opened 2023-01-19
Multicluster - cluster configuration


nrfox:
Kiali needs to configure connections to multiple kubernetes clusters. This issue covers how Kiali will do that.

Epic: #5618 

PRs:
* server: https://github.com/kiali/kiali/pull/5752
* operator: https://github.com/kiali/kiali-operator/pull/609
* helm charts: https://github.com/kiali/helm-charts/pull/189

jmazzitelli:
I'm going to think up something here to see how we can notify Kiali about which clusters are configured with their credentials and perhaps some extension points for future needs (like adding additional config/metadata around each cluster definition)

jmazzitelli:
Just wanted to document this... here's how these cluster secrets (to be used by Kiali in order to authenticate to remote clusters):

```
istioctl x create-remote-secret --name remote-cluster-name --context my-context --namespace istio-system
```

Results in something like this:

```yaml
# This file is autogenerated, do not edit.
apiVersion: v1
kind: Secret
metadata:
  annotations:
    networking.istio.io/cluster: remote-cluster-name
  creationTimestamp: null
  labels:
    istio/multiCluster: "true"
  name: istio-remote-secret-remote-cluster-name
  namespace: istio-system
stringData:
  remote-cluster-name: |
    apiVersion: v1
    clusters:
    - cluster:
        certificate-authority-data: LS0tLS1CR...QotLS0tLUVORCBDRVJUSUZJQ0FURS0tLS0tCg==
        server: https://192.168.59.254:8443
      name: remote-cluster-name
    contexts:
    - context:
        cluster: remote-cluster-name
        user: remote-cluster-name
      name: remote-cluster-name
    current-context: remote-cluster-name
    kind: Config
    preferences: {}
    users:
    - name: remote-cluster-name
      user:
        token: eyJhbGciOiJSU...Yzpv5Q3GmsN-JeVg166A
---
```

nrfox:
Regardless of what tool is used to generate the credentials, what we need per cluster is:

1. Secret in the Kiali cluster with a kubeconfig that has credentials/connection info for the kiali user
2. Service account in the remote cluster for Kiali
3. Role/rolebinding for the Kiali service account in the remote cluster

jmazzitelli:
FYI: when creating the secret using istioctl, notice two pieces of data you can get - the cluster name and the label that identifies it as a secret for a remote cluster:

```
[...]

PR #5752, opened 2023-01-23
read the remote cluster secrets from the file system

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/5738

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -3,6 +3,7 @@ package business
 import (
 	"context"
 	"encoding/json"
+	stderrors "errors"
 	"fmt"
 	"net/http"
 	"os"
@@ -60,9 +61,6 @@ type Cluster struct {
 
 	// Network specifies the logical NETWORK_ID as known by the Control Plane
 	Network string `json:"network"`
-
-	// SecretName is the name of the kubernetes "remote secret" where data of this cluster was resolved
-	SecretName string `json:"secretName"`
 }
 
 // KialiInstance represents a Kiali installation. It holds some data about
@@ -293,7 +291,6 @@ func (in *MeshService) ResolveKialiControlPlaneCluster(r *http.Request) (*Cluste
 		KialiInstances:       kialiInstances,
 		Name:                 myClusterName,
 		Network:              kialiNetwork,
-		SecretName:           "",
 	}
 
 	return kialiControlPlaneCluster, nil
@@ -458,12 +455,13 @@ func (in *MeshService) resolveKialiNetwork() (string, error) {
 	return typedNetworkConfig, nil
 }
 
+// Defines where the files are located that contain the remote cluster secrets
+var remoteClusterSecretsDir = "/kiali-remote-cluster-secrets"
+
 // resolveRemoteClustersFromSecrets resolves the metadata about "other" clusters that are
 // visible to the adjacent mesh control plane. This assumes that the Istio namespace is
 // named the same as in Kiali's Cluster.
 func (in *MeshService) resolveRemoteClustersFromSecrets() ([]Cluster, error) {
-	conf := config.Get()
-
 	// For the ControlPlane to be able to "see" remote clusters, some "remote secrets" need to be in
 	// place. These remote secrets contain <kubeconfig files> that the ControlPlane uses to
 	// query the remote clusters. Without them, the control plane is not capable of pushing traffic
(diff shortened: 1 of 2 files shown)


jmazzitelli:
This is working, in conjunction with the operator PR. I just have to fix up the unit tests in mesh_test.go.

business/mesh.go
```diff
@@ -462,7 +459,6 @@ func (in *MeshService) resolveKialiNetwork() (string, error) {
 // visible to the adjacent mesh control plane. This assumes that the Istio namespace is
 // named the same as in Kiali's Cluster.
 func (in *MeshService) resolveRemoteClustersFromSecrets() ([]Cluster, error) {
```
nrfox, line 461: Perhaps this can be a separate PR but imo we should move this logic to the client factory and attach the cluster info to the kiali `kubernetes.ClientInterface`. Right now there's a few different places that call `ClientConfig()` to recreate the Kiali Service Account client but we only need to create this client once and then pass off the client to the various business services. Managing the client creation in a single place will also greatly simplify using clients for mutlicluster.

So if all the logic for getting the cluster info from the secrets and turning those into `kubernetes.ClientInterface` goes into the client factory and if something like a `GetClusterInfo` method gets added to `kubernetes.ClientInterface`, then the mesh service would just call `GetClusterInfo()` for each one of the Kiali Service Account clients.

But again maybe this change should just focus on reading the secrets from fs rather than from the kube API and another change can refactor this logic into the client factory. Either way is fine.

business/mesh.go
```diff
@@ -474,54 +470,70 @@ func (in *MeshService) resolveRemoteClustersFromSecrets() ([]Cluster, error) {
...
+	// Each cluster has its own subdirectory, usually named the same as the cluster (though not necessarily). e.g. "/kiali-remote-cluster-secrets/<cluster name>"
+	// Cluster configs are found in a file whose name is the cluster name; e.g. "/kiali-remote-cluster-secrets/<cluster name>/<cluster name>"
+	// So walk the directory "/kiali-remote-cluster-secrets" to obtain all known remote cluster configs.
+
+	// Defines where the files are located that contain the remote cluster secrets
+	const remoteClusterSecretsDir = "/kiali-remote-cluster-secrets"
+	remoteClusterSecrets := make(map[string][]byte, 0)
+
+	secretDirs, err := os.ReadDir(remoteClusterSecretsDir)
+	if err == nil {
```
nrfox, line 483: You can reverse the `err` check to `err != nil` and return early. You may end up with more lines of code but it'll be much more readable than a deeply nested `if/else` tree and I'd prioritize readability.

nrfox (commented):
I know this is still draft but offering my 2c early in case it's useful

business/mesh.go
jmazzitelli: > But again maybe this change should just focus on reading the secrets from fs rather than from the kube API and another change can refactor this logic into the client factory. Either way is fine.

Right. That is the intent. We'll iterate over this is different PRs.

business/mesh.go
jmazzitelli: > You can reverse the err check to err != nil and return early.

We don't return after that if-stmt. It keeps going to do more processing, even if there are no secrets (or we fail to read secrets). So there really isn't a true short-circuit we could do here.

jmazzitelli:
To test, you need to install Istio with a mesh ID (in our hack install-istio-via-istioctl script, pass in "--mesh-id mymesh" to set it when installing). Then install a remote cluster secret (doesn't have to point to a real remote cluster) and then install Kiali (either with make operator-create or with the kiali server helm chart - make sure you have the PRs pulled for those repos). Then in the Kiali UI look at the Mesh page and you should see things still work.

business/mesh.go
nrfox: > So there really isn't a true short-circuit we could do here

Can you use `continue` instead of `return` like you are already doing on line 493.
```
for _, sd := range secretDirs {
	secretAbsDir := remoteClusterSecretsDir + "/" + sd.Name()
	secretFiles, err := os.ReadDir(secretAbsDir)
	if err != nil {
            log.(...)
            continue
       }
       for ... {
```

The outermost I think can return empty remote secrets?
```
	secretDirs, err := os.ReadDir(remoteClusterSecretsDir)
	if err == nil {
            log.Error(...)
            return []Cluster{}, nil
        }
```

This might just be a personal preference but it's just hard for me to read what's going on when it's nested like this. Sometimes it can't be helped.

business/mesh.go
```diff
@@ -474,54 +472,68 @@ func (in *MeshService) resolveRemoteClustersFromSecrets() ([]Cluster, error) {
...
+	// Cluster configs are found in a file whose name is the cluster name; e.g. "/kiali-remote-cluster-secrets/<cluster name>/<cluster name>"
+	// So walk the directory "/kiali-remote-cluster-secrets" to obtain all known remote cluster configs.
+
+	remoteClusterSecrets := make(map[string][]byte, 0)
+
+	secretDirs, err := os.ReadDir(remoteClusterSecretsDir)
+	if err == nil {
+		for _, sd := range secretDirs {
+			secretAbsDir := remoteClusterSecretsDir + "/" + sd.Name()
+			secretFiles, err := os.ReadDir(secretAbsDir)
```
nrfox, line 486: We're expecting a separate remote secret dir per remote secret? Won't there just be 1 remote secret per cluster?

jmazzitelli: This has to do with the way the secrets are mounted on the file system by k8s. It's weird. I think its because there can be multiple configs in the secret itself. So the way the file system will be laid out is:

`/kiali-remote-cluster-secrets/<name of the cluster as defined in the pod mountVolume>/<name of the secret data as found in the Secret itself, which is the name of the cluster if created by istioctl>`

The way I coded this up, it should support 1 cluster per secret, or N clusters per secret (the idea being that perhaps I want a single Secret to contain all the information on all my clusters). I did not test it, but that's what this code should support today. This came for free just because of the way k8s lays out the file system by the way the volumes are declared - we just loop through all the files in the directory. Right now, you'll probably only see 1 secret data file per cluster subdirectory.

leandroberetta (approved):
Looks good, Tested locally with the operator.

image

jmazzitelli:
@nrfox this is all going to get refactored soon into the ClientFactory stuff - we can change the err==nil to err!=nil stuff during that refactor.

business/mesh.go
nrfox: Hmmm can the operator/helm chart combine all the secrets into a single dir with a projected volume? https://kubernetes.io/docs/concepts/storage/projected-volumes/. This is not very important though so if it's much work to change then don't bother.

nrfox:
> @nrfox this is all going to get refactored soon into the ClientFactory stuff - we can change the err==nil to err!=nil stuff during that refactor.

That works too. Might be worth adding a `TODO` note but I'll approve otherwise and leave that up to you.

nrfox (approved):

Issue #5753, opened 2023-01-24
[cypress] Gateway API in Cypress tests.


hhovsepy:
Add cypress tests for:

- [x] Service Details Wizard K8s Gateway API Routing
- [x] Config List Wizard create K8s Gateway

Issue #5748, opened 2023-01-23
[cypress flake] istio_config.feature KIA1102


hhovsepy:
Kiali Istio Config page
KIA1102 validation:
AssertionError: Timed out retrying after 40000ms: expected 'color: rgb(62, 134, 53); vertical-align: -0.125em;' to contain 'danger-color'

Should be added waiting in a loop for the newly created VS to be loaded.

https://github.com/kiali/kiali/actions/runs/3967483927/jobs/6805450547

hhovsepy:
Resolved in a scope of PR: https://github.com/kiali/kiali/pull/5733

PR #5754, opened 2023-01-24
[cypress] Added Service Details wizard check for K8s Gateway API Routing

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5753


josunect (commented):
image

Code looks good and it works as expected. My only concern would be, should we need to change the script that generates the CI environment to enable GW API?

hhovsepy:
thank you @josunect , we have changed the script in another PR https://github.com/kiali/kiali/issues/5748

josunect (approved):
So LGTM

PR #5758, opened 2023-01-25
Update Forms to add elements

### LINKED ISSUE #5741: Form to add port should start empty
When creating objects such as Service Entries that have a port field, the port field is already present and prepopulated with some values. It seems like completing the port form will cause it to be added to the form submission however you must first click the `+` button in order to add the port. Clicking the `+` button does add the port to the form submission but it also adds a new port field that is not part of the preview. This UX is confusing because the `+` button does 2 things: adds the existing port field to the form _and_ adds a new port field that is not part of the form.

Instead the port field should start empty and any necessary ports added by first hitting the `+` button. Then any port fields that are present will be considered part of the form submission and should show up in the preview. The `+` button will only do 1 function which is to add new port fields to the form.

Current behavior:
port3

josunect:
Fixes #5741 

image

### DIFF
--- a/frontend/src/pages/IstioConfigNew/ServiceEntryForm.tsx
+++ b/frontend/src/pages/IstioConfigNew/ServiceEntryForm.tsx
@@ -5,7 +5,7 @@ import { TextInputBase as TextInput } from '@patternfly/react-core/dist/js/compo
 import { isGatewayHostValid } from '../../utils/IstioConfigUtils';
 import { cellWidth, ICell, Table, TableBody, TableHeader } from '@patternfly/react-table';
 import { PlusCircleIcon, TrashIcon } from '@patternfly/react-icons';
-import {cssRaw, style} from 'typestyle';
+import { style} from 'typestyle';
 import { PFColors } from '../../components/Pf/PfColors';
 import { isValid } from 'utils/Common';
 import { FormEvent } from "react";
@@ -63,9 +63,17 @@ type Props = {
   onChange: (serviceEntry: ServiceEntryState) => void;
 };
 
+type FormPort = {
+  number: string;
+  name: string;
+  protocol: string;
+  targetPort: string;
+}
+
 export type ServiceEntryState = {
   serviceEntry: ServiceEntrySpec;
   validHosts: boolean;
+  formPorts: FormPort[];
 };
 
 export const initServiceEntry = (): ServiceEntryState => ({
@@ -75,31 +83,36 @@ export const initServiceEntry = (): ServiceEntryState => ({
     ports: []
   },
   validHosts: false,
+  formPorts: Array<FormPort>()
 });
 
-const isValidPort = (ports: Port[]) => {
+const isValidPort = (ports: FormPort[]) => {
+
   return ports.every((p, i) =>
-    p.name.length > 0 && p.number !== undefined && p.protocol.length > 0 && noDuplicatePortNames(p.name, i, ports))
+    isValidName(p.name) && isValidPortNumber(p.number)
+      && isValidTargetPort(p.targetPort)
+      && noDuplicatePortNames(p.name, i, ports))
 };
 
-const noDuplicatePortNames = (name: string, index: number, ports: Port[]) => {
+const noDuplicatePortNames = (name: string, index: number, ports: FormPort[]) => {
   return ports.every((p, i) => i !== index ? p.name !== name : true )
 }
 
 export const isServiceEntryValid = (se: ServiceEntryState): boolean => {
-  return se.validHosts && se.serviceEntry.ports !== undefined && se.serviceEntry.ports.length > 0
-    && isValidPort(se.serviceEntry.ports);
+  return se.validHosts && se.serviceEntry.ports !== undefined && isValidPort(se.formPorts);
 };
 
-cssRaw(`
-  .noArrows input::-webkit-outer-spin-button, input::-webkit-inner-spin-button {
-    -webkit-appearance: none;
-    margin: 0;
-  }
-  input[type=number] {
-    -moz-appearance: textfield;
-  }
-`)
+const isValidName = (name: string): boolean => {
+  return name.length > 0
+}
+
+const isValidPortNumber = (portNumber: string): boolean => {
+  return portNumber.length > 0 && !isNaN(Number(portNumber))
+}
+
+const isValidTargetPort = (targetPort: string): boolean => {
+  return targetPort.length === 0 || (targetPort.length > 0 && !isNaN(Number(targetPort)))
+}
 
 class ServiceEntryForm extends React.Component<Props, ServiceEntryState> {
   constructor(props: Props) {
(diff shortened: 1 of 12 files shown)


nrfox (changes requested):
From the screenshot, if the ports field is required then it needs a red asterisk next to it indicating that.

frontend/src/pages/IstioConfigNew/ServiceEntryForm.tsx
```diff
@@ -8,6 +8,7 @@ import { PlusCircleIcon } from '@patternfly/react-icons';
 import { style } from 'typestyle';
```
nrfox, line 8: Ignore this line. Just adding this here to capture multiple review comments in one place.

nrfox: Can you get rid of the up arrow/down arrow that increments/decrements the ports? These don't seem useful since you'd likely need to click these thousands of times to get to the desired port. They also let you enter invalid port values like negative numbers.

Screenshot from 2023-01-25 10-35-42

nrfox: I'm not this is worth changing but the Port Number box is marked green when it is empty. Since the user never entered any value here, it seems strange to mark this with a green check mark even that it's technically correct. Not sure how Kiali handles this for other fields that are similar but the `Protocol` and the `Location` fields both have default values and are not marked as green. What do you think?

Screenshot from 2023-01-25 10-37-23

Protocol field:
Screenshot from 2023-01-25 10-37-31

nrfox: It looks like you can have an invalid port field and still see the "preview" button. I think the preview button should only appear when the form is totally valid. Otherwise you could create invalid SE objects.

Screenshot from 2023-01-25 10-38-03

nrfox: Can you replace the kebab with a minus button? The only option in the kebab is to "remove port" and with the kebab it takes two clicks to remove the port. Having a minus button reduces the clicks to one. The kebab also somewhat hides removing the port whereas the minus button makes it more clear how to remove a port from the form.

josunect: > Can you get rid of the up arrow/down arrow that increments/decrements the ports? These don't seem useful since you'd likely need to click these thousands of times to get to the desired port. They also let you enter invalid port values like negative numbers.
> 
> Screenshot from 2023-01-25 10-35-42

This is changed know

josunect: > I'm not this is worth changing but the Port Number box is marked green when it is empty. Since the user never entered any value here, it seems strange to mark this with a green check mark even that it's technically correct. Not sure how Kiali handles this for other fields that are similar but the `Protocol` and the `Location` fields both have default values and are not marked as green. What do you think?
> 
> Screenshot from 2023-01-25 10-37-23
> 
> Protocol field: Screenshot from 2023-01-25 10-37-31

Probably this one is a bit more tricky. Because this is the default behavior of the validation, and also the default behavior in the other components of the wizards. In the case of the Protocol, this is not validated as the values are predefined (That is why the green check is not shown)

josunect: > It looks like you can have an invalid port field and still see the "preview" button. I think the preview button should only appear when the form is totally valid. Otherwise you could create invalid SE objects.
> 
> Screenshot from 2023-01-25 10-38-03

Well seen! I've changed.

josunect: > Can you replace the kebab with a minus button? The only option in the kebab is to "remove port" and with the kebab it takes two clicks to remove the port. Having a minus button reduces the clicks to one. The kebab also somewhat hides removing the port whereas the minus button makes it more clear how to remove a port from the form.

Done.

josunect:
Thanks for the review, @nrfox ! 
Once the solution is validated in the serviceEntry it should be applied to: 

- Gateway
- k8sGateway

frontend/src/pages/IstioConfigNew/ServiceEntryForm.tsx
nrfox: It looks great now! Just a couple small issues.

1. I suspect this bug existed before your changes but the validation lets you specify multiple ports with the same name. When you go to create them you'll get an error back from the API.
Screenshot from 2023-01-26 14-59-51

2. There's no validation for Target Port. You can enter letters there and invalid numbers and it'll still appear valid. Sometimes kiali crashes when you do this and sometimes it'll exclude the target port field entirely from the form data. The target port validation should be the same as the port number validation. As an edge case I'm not sure if you can have multiple target ports that are the same or not.
Screenshot from 2023-01-26 15-01-51

Last thing, should there be cypress tests with these changes? There may already be coverage?

josunect: Thanks, @nrfox ! Changes addressed. 

There are no changes required in testing, I haven't seen test coverage for this. I'll suggest to create a separate issue to improve test coverage for this.

nrfox (commented):
Looks like target port is still not being validated but everything else works great. I dismissed the old "requested changes review" so that shouldn't be a blocker anymore.

targetport

josunect:
> Looks like target port is still not being validated but everything else works great. I dismissed the old "requested changes review" so that shouldn't be a blocker anymore.
> 
> targetport [ targetport ](https://user-images.githubusercontent.com/6226732/215173005-03688971-3bfd-4087-b108-20b767b1d755.gif) [ ](https://user-images.githubusercontent.com/6226732/215173005-03688971-3bfd-4087-b108-20b767b1d755.gif)

Yes, it looks like still has an unexpected behavior. It shouldn't let you write letters, but after write a number, a letter can be written. Anyway, I've tried this solution: 

```
type="text" 
inputMode="numeric" 
pattern="[0-9]*"
```
And I can write letters anyway. So, I will be going to the previous solution of having a separate structure with text fields and then, convert to numeric fields.

hhovsepy:
Should the negative port numbers be accepted?
Screenshot from 2023-01-30 17-19-34
Otherwise lgtm.

hhovsepy (approved):
Verified:
Screenshot from 2023-01-30 17-56-45
Screenshot from 2023-01-30 17-55-47

josunect:
> Verified: Screenshot from 2023-01-30 17-56-45 Screenshot from 2023-01-30 17-55-47

Thanks, @hhovsepy ! I will apply to other Istio Configs where applicable.

jshaughn:
@josunect , I think you fixed the issue requested by @nrfox , correct?  If so I think this is ready for merge.

josunect:
> @josunect , I think you fixed the issue requested by @nrfox , correct? If so I think this is ready for merge.

Yes, but just in the service entry. The same usability issue is also in other wizards, like k8sGateway. Should we open a new issue? I was working on this today but I haven't finished yet.

jshaughn:
Ah, OK!  Maybe just change the title for clarity, thanks!

josunect:
Hi, @hhovsepy , still need to work on improving the GW Form for adding a Port, but, if you want to review what it is already done, it is fine

hhovsepy:
Hi @josunect , early testing comment while WIP, the Listeners are fine, but the Addresses does not accept values when typing:
Screenshot from 2023-02-02 11-04-30

hhovsepy:
K8sGateway works fine now:
Screenshot from 2023-02-03 11-40-42

josunect:
> K8sGateway works fine now: Screenshot from 2023-02-03 11-40-42

Thank you, @hhovsepy ! I'm reworking the Gateway and Servers field ATM.

hhovsepy:
@josunect thanks for the changes it works fine, just 2 minor things:
in Service Entry form, "Add Port" label better to be clickable alongside with "+" icon, to be consistent:
Screenshot from 2023-02-06 09-28-59
Port field's icon is not red when wrong port is set, even though the Preview button is disabled:
Screenshot from 2023-02-06 09-27-30

josunect:
> @josunect thanks for the changes it works fine, just 2 minor things: in Service Entry form, "Add Port" label better to be clickable alongside with "+" icon, to be consistent: Screenshot from 2023-02-06 09-28-59 Port field's icon is not red when wrong port is set, even though the Preview button is disabled: Screenshot from 2023-02-06 09-27-30

Well caught, thank you @hhovsepy !

hhovsepy:
@josunect thanks for the fixes, they work now as expected.
Only one thing I just noticed now:
Screenshot from 2023-02-06 13-24-37
In TLS SIMPLE mode, when certficates are not provided, but the Preview is enabled.

hhovsepy:
@josunect another PR was merged causing conflicts, sorry about that.

josunect:
> @josunect another PR was merged causing conflicts, sorry about that.

Thank you, @hhovsepy ! Changes addressed.

hhovsepy (approved):
Verified now, thank you for the fix @josunect

josunect:
> Verified now, thank you for the fix @josunect

Thanks for the review!

PR #5768, opened 2023-01-30
Fetch OpenShift version using Kiali ServiceAccount

### LINKED ISSUE #5755: Don't use the anonymous account to fetch cluster version (OpenShift authentication)
When using OpenShift authentication, there is this part of the code that fetches the cluster version: https://github.com/kiali/kiali/blob/9c98bbbbf5141eedd0f0b67b05271b4b33b072a9/business/openshift_oauth.go#L76-L79

To fetch the cluster version, the `system:anonymous` account is used. With the cluster's default bundled ClusterRoles and ClusterRoleBindings, the anonymous account can fetch the version correctly. However, these roles/bindings can be modified and privileges can be removed for security hardening. If the `system:anonymous` account loses its privileges to fetch the cluster version, OpenShift authentication becomes unusable.

It is better to use the Kiali SA to fetch the cluster version and allow for improved security on the cluster. The Kiali SA would need to gain privileges to query the cluster version.

israel-hdez:
This change applies only to the OpenShift authentication mechanism.

Previously, the cluster version was retrieved using an unauthenticated query to the cluster API. This is ensuring that such query is authenticated using the Kiali ServiceAccount. This is to overcome situations when non-authenticated users cannot query the cluster version.

**Describe the change**

There are cases when users choose to remove the privileges to fetch the cluster version to unauthenticated queries to the cluster API. Presumably, this is because of security reasons. Since Kiali was using an unauthenticated query to fetch the cluster version, OpenShift authentication becomes unusable in such situation.

I was thinking in also modifying the kiali-operator, so that the Kiali SA explicitly gets privileges to query the version of the cluster. However, I stepped back.

Turns out, the privilege to query the version of the cluster is given to _all_ authenticated queries via the ClusterRole `system:discovery`.  Additionally, in OpenShift the ClusterRoles `cluster-status` and `system:openshift:discovery` also give this privilege to authenticated users. After cleaning the privilege from all these ClusterRoles, I successfully had a failing Kiali. However, since all _authenticated_ queries will no longer have this privilege, the operator also started failing as it queries the version of the cluster. That would mean that also the operator helm charts and the operator OLM manifests would be impacted as they would need to include the privilege to fetch the version of the cluster. This would be a bigger change.

So, for now and in order to keep things simple, I'm just relying in that _authenticated_ queries will, by default, be able to get the version of the cluster. If in the future we hear that somebody is removing "fetch cluster version" privileges even to authenticated queries, we may consider that bigger change.

**Issue reference**

Fixes #5755

### DIFF
--- a/handlers/authentication.go
+++ b/handlers/authentication.go
@@ -7,6 +7,7 @@ import (
 
 	"k8s.io/client-go/tools/clientcmd/api"
 
+	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/business/authentication"
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):
lgtm

jshaughn:
Thanks for documenting the thought process here, @israel-hdez !

PR #5771, opened 2023-01-30
Show NA for istio config if istio api disabled

### LINKED ISSUE #5762: No istiod: Disable validations also in the Overview page
When istio_api_enabled is false, validations are disabled but they can still be seen in the Overview page: 

image
image

They should be disabled in the Overview page to keep consistency.

josunect:
Fixes #5762 

image


leandroberetta (approved):
Looks good.

PR #5779, opened 2023-01-31
No longer read kubernetes_config.cache_enabled config

israel-hdez:
This small change will prevent reading (unmarshalling) the kubernetes_config.cache_enabled configuration from a YAML file (i.e. from a Kubernetes ConfigMap).

This effectively makes it impossible to disable Kiali's kube cache via the usual configuration mechanism. This will allow to write new code under the assumption that the kube cache is always enabled (i.e. no more if/else conditionals around this configuration variable).

The kube cache can still be disabled programmatically. This will allow to gradually update existing code to stop relying on the removed configuration (most notably, unit tests need to be adapted).

Related #5599

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -286,7 +286,7 @@ type KubernetesConfig struct {
 	CacheDuration int `yaml:"cache_duration,omitempty"`
 	// Enable cache for kubernetes and istio resources
 	// TODO: Remove once all services are migrated to use the cache.
-	CacheEnabled bool `yaml:"cache_enabled,omitempty"`
+	CacheEnabled bool `yaml:"-,omitempty"`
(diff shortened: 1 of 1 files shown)


nrfox (approved):
LGTM

jshaughn:
Can you point to an example of how to disable it programmatically (for tests, I assume)?

israel-hdez:
Yes, basically you get the default configs, turn off the cache in the copy you get, and then set it to take effect. Here is a chunk of code from some of the tests: https://github.com/kiali/kiali/blob/298daa9b9e8267024a60eaa83577a20930b5bad7/handlers/apps_test.go#L145-L148

PR #5783, opened 2023-02-01
Use downloaded Helm instead of the packed in the VM

leandroberetta:
Always use the Helm chart downloaded from Kiali scripts.

Helm charts PR: https://github.com/kiali/helm-charts/pull/193

### DIFF
--- a/hack/setup-kind-in-ci.sh
+++ b/hack/setup-kind-in-ci.sh
@@ -147,11 +147,16 @@ infomsg "Cloning kiali helm-charts..."
 git clone --single-branch --branch "${TARGET_BRANCH}" https://github.com/kiali/helm-charts.git helm-charts
 make -C helm-charts build-helm-charts
 
+HELM="helm-charts/_output/helm-install/helm"
+
+infomsg "Using helm: $(ls -l ${HELM})"
+infomsg "$(${HELM} version)"
+
 infomsg "Installing kiali server via Helm"
 # The grafana and tracing urls need to be set for backend e2e tests
 # but they don't need to be accessible outside the cluster.
 # Need a single dashboard set for grafana.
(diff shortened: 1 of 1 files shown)


jmazzitelli:
This looks to be a dup of https://github.com/kiali/kiali/pull/5781

See my comments in that PR for something we need to think about.

jmazzitelli (commented):
see my previous comment about the existing PR. Feel free to close that one and we can work in this one.

But my comment in the other PR still stands - I think we need to make sure "helm" isn't in PATH already - because if it is (in the CI environment) then this change will be a no-op because the make target will not download helm if there is already a helm in PATH (it presumes the user wants to use his own helm).

So we could change this such that we should remove helm from PATH if it exists, prior to building the helm chart. That will assure we are building with a helm version that we want and helm will be downloaded.

jmazzitelli:
> So we could change this such that we should remove helm from PATH if it exists

 Or the alternative would be to change the helm chart Makefile so it always downloads helm if it is missing from the _output directory (i.e. ignore what is in PATH and always use what is in _output)

leandroberetta:
> This looks to be a dup of #5781 - we can close that one and use this one if you want. But see my comments in that PR for something we need to think about.

Sorry @jmazzitelli I didn't realize of the other PR. As you say, we can continue on this one and also improve what you're indicating.

To me, the best option is to always use the downloaded one, and avoid future errors with versions incompatbilities. If you're ok, I will add this change.

jmazzitelli:
> If you're ok, I will add this change.

sounds good

jmazzitelli (approved):
LGTM.

This should always do what we expect now that the helm Makefile will always download helm regardless of what is in PATH

PR #5784, opened 2023-02-01
Fix caching when fetching workloads

### LINKED ISSUE #5743: Graphs loading very slowly
Hi,
I've just deployed istio and now adding kiali, but the graphs are loading very slowly:
1 namespace with 2 services 2 apps and 2 edges takes 2-3 secs to load,
2 namespaces with 3 services 3 apps and 6 edges takes 5-6 secs to load,
3 namespaces with 3 services 4 apps and 7 edges takes 8-9 secs and so forth, until it reaches a point where it times out and displays an error instead of graph.

My current state:
Kiali version: 1.62.0
Istio version: 1.16.1
Running on AKS

I have only configured prometheus for istio, no grafana, nor jaeger, also no ingress or egress gateways.
image

I have checked prometheus/istio/kiali pods - no unusual restarts, no cpu throttling:
image
image
image

K8s node resources look OK to me:
image

metrics in prometheus seem to be loading normally (though I am not sure I am testing it properly):
image

I have no more ideas where to look for problems.

nrfox:
The fetchWorkload(s) funcs used the business layer k8s rather than the workload service's k8s. This resulted in the fetchWorkload(s) funcs not using the caching client and causing a perf regression since fetchWorkloads called the API server directly rather than pulling from the kube cache.

Fixes #5743

### DIFF
--- a/business/apps_test.go
+++ b/business/apps_test.go
@@ -8,6 +8,7 @@ import (
 	osproject_v1 "github.com/openshift/api/project/v1"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
 	apps_v1 "k8s.io/api/apps/v1"
 	batch_v1 "k8s.io/api/batch/v1"
 	core_v1 "k8s.io/api/core/v1"
@@ -27,7 +28,10 @@ func setupAppService(k8s *kubetest.K8SClientMock) *AppService {
 
 func TestGetAppListFromDeployments(t *testing.T) {
 	assert := assert.New(t)
+	require := require.New(t)
+
 	conf := config.NewConfig()
+	conf.KubernetesConfig.CacheEnabled = false
 	config.Set(conf)
 
 	// Setup mocks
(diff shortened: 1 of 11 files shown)


jshaughn:
This PR will need a backport to v1.63 to ensure we don't have this performance regression in the Istio v1.17 addon version.  We'll then need to release 1.63.2.

jshaughn (approved):
Looks good to me, good eyes Nick for seeing this problem, and thanks for helping me with this issue.  Good example here of how we now have to programatically disable the cache for tests.

israel-hdez:
I think we may need to improve the variable naming. I can see that `k8s` variable sitting in `WorkloadService` is tempting to use. And there is no clue that it would be a "direct" k8s client rather than a caching client.

jshaughn:
True, it's not easy to know whether what you're using is cache-aware.

nrfox:
> True, it's not easy to know whether what you're using is cache-aware.

Yes introducing the `CachingClient` obscures when Kiali is using the cache. If everything was using the `CachingClient` this wouldn't really be an issue but since the move to a mandatory cache is happening iteratively it is confusing. I'm ok with removing the `CachingClient` since it's just another layer of abstraction that in this case hides what actually should be explicit: cache reads. Another reason to be more explicit about when we read from the cache vs. directly talk to kube is that they use two different client tokens. Cache reads use the kiali service account token and talking directly to the kube api usually uses the user's token. Kiali should be explicit on when one is being used vs. the other. In other words, if you have a `k8s` you can assume that it is the user's client.

israel-hdez:
> I'm ok with removing the `CachingClient` since it's just another layer of abstraction that in this case hides what actually should be explicit: cache reads.

Hmmm, nope that wasn't my suggestion.
I was meaning _variable naming_ rather than _type naming_.

To me, the `CachingClient` is fine. And I think removing it won't solve anything, because most of the time the code is referencing the `ClientInterface`, which is already hiding the implementation anyways.

I was talking specifically about the `k8s` variable in this struct: https://github.com/kiali/kiali/blob/e842306ba1844aeefcb7b5a03bffa1d6f20edf3f/business/workloads.go#L48-L57

In `business`, most of the time we are using with a "cached" ClientInterface. So, you may think that the `k8s` variable can be a cached client (depending on configs), but it isn't. If I understand the code correctly, hat specific `k8s` variable is _always_ a non-cached client, but there is no comment communicating that, nor the variable name gives a clue. So, it is error prone.

So, my suggestion was to rename that one to make it clear that that client is bypassing the cache when you want to make use of it. Or, perhaps, at least add a comment.

jshaughn:
+1, I was only talking about var naming as well.

Issue #5751, opened 2023-01-23
Configure Kiali timeouts while fetching traces from jaeger


RaiAnandKr:
The kiali instance is hitting timeouts while fetching few hundreds traces over a period of 30mins:
```
Get "http://jaeger-query.jaeger.svc.cluster.local.:80/api/traces?
end=1674495432000000&limit=500&service=ratelimit.default&start=1674493632000000": 
context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```
I tried the query directly from Jaeger UI and it takes ~10s to complete there, so I know it's slow but is there any way to configure the timeouts for jaeger query in Kiali so that if the user is willing to wait, they can see the traces ultimately?

The other part I am curious about is, if I select `100 limits per query` from the `Display` dropdown, we actually query 500 traces in the backends (basically always 5X, if I select `500 limits per query`, we query 2500 traces). Attaching a snapshot for reference:

![Screenshot 2023-01-23 at 11 07 36 PM](https://user-images.githubusercontent.com/21973868/214113592-3d49a42a-02ea-467f-88c4-7aca19c1b0e5.png)

Query 2500 traces is really expensive (even Jaeger UI allows a max limit of 1500) so there is not much point in trying `500 limits per query` or anything higher than that. What's the motive behind querying 5X traces in the backend? Can we make this configurable too or do anything to work around this?

I am using Kiali v1.61.0

jshaughn:
Hi @RaiAnandKr,  I'm not an expert in the tracing part of Kiali, but I did a little bit of looking at the code...  First, about querying 5X  more traces than specified in the UI.  It seems that's due to Kiali trying to find workload-level traces given an API that deals at the "app" level.  Because only a subset of the app traces may involve the specific workload, Kiali seems to be increasing the number of traces to compensate. At least I think this is what is happening.  There is a similar 2X when looking for traces in service details.  In the UI, for App details it should be the expected 1X as specified in the UI.

Although well-intentioned I agree that this is confusing and can cause querying impact beyone what the UI user would expect.  Probably better would be to honor the UI limits and add an info icon in the UI explaining that the user may want to increase the limits when querying for workload (or service) traces.

As for the timeouts, also not great news, it looks like a fixed 5s timeout for HTTP and 4s timeout for GRPC.  I agree it would be better to make this configurable.

So, I don't think what you are seeing is a bug, but I think perhaps we need an enhancement here.

jshaughn:
[...]

PR #5793, opened 2023-02-02
Honor configured timeout for calls to tracing API

jshaughn:
- Do not inflate the query limit for workload and service trace queries, it is misleading on the client and can cause unexpected timeout.
- Add info tooltip in Traces Display dropdown to better explain query limit
  - fix tooltip styling and position for the existing menu

Part-of  https://github.com/kiali/kiali/issues/5751

**OPERATOR PR**: https://github.com/kiali/kiali-operator/pull/610


jmazzitelli (approved):
the code looks correct. I did not run this.

PR #5798, opened 2023-02-03
[Cypress] Fix for test flakes introduced by the presence of the Gateway API

matejnesuta:
This PR is related to 2 flaky tests on OCP.
image
image

These tests expected exactly one item to be present after filtering, but there are more items present due to Gateway API being used. Now the tests expect variable number of items.

### DIFF
--- a/frontend/cypress/integration/common/apps.ts
+++ b/frontend/cypress/integration/common/apps.ts
@@ -85,10 +85,14 @@ And('user sees Details information for Apps', () => {
   });
 });
 
-Then('user only sees {int} apps', (sees: number) => {
+And('user only sees the apps with the {string} name in the {string} namespace', (name:string,ns:string) => {
+  let count:number;
+  cy.request('GET', `/api/namespaces/${ns}/apps`).should((response) =>{
+    count = response.body.applications.filter((item) => item.name.includes(name)).length;
+  }); 
   cy.get('tbody').within(() => {
     cy.contains('No apps found').should('not.exist');
-    cy.get('tr').should('have.length', sees);
+    cy.get('tr').should('have.length', count);
   });
 });
 
--- a/frontend/cypress/integration/common/istio_config.ts
+++ b/frontend/cypress/integration/common/istio_config.ts
@@ -18,6 +18,25 @@ function labelsStringToJson(labelsString: string) {
   return `{${labelsJson}}`;
 }
 
+// I included this, because the URL parameter is in plural, but the the Type itself in Kiali is singular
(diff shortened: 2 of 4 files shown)


frontend/cypress/integration/common/apps.ts
```diff
@@ -85,10 +85,14 @@ And('user sees Details information for Apps', () => {
   });
 });
 
-Then('user only sees {int} apps', (sees: number) => {
+Then('user only sees the filtered apps', () => {
+  let count:number;
+  cy.request('GET', "/api/namespaces/bookinfo/apps").should((response) =>{
```
hhovsepy, line 90: The method should not contain hardcoded values such as "bookinfo" and "reviews", they need to be as param.
What if you just change the conflicting k8s tests to use "ratings" instead of "reviews"? They should not collide.

frontend/cypress/integration/common/istio_config.ts
```diff
@@ -301,6 +301,17 @@ Then('user only sees {string}', (sees: string) => {
   });
 });
 
+Then('user only sees items related to {string}', (sees: string) => {
+  let count:number;
+  cy.request('GET', "/api/istio/config?objects=gateways&validate=true").should((response) =>{
```
hhovsepy, line 306: the same goes here, should not be hardcoded "gateways"

hhovsepy: the problem is that Istio generates Apps and Gateways for k8s gateways, which is causing conflicts.

hhovsepy (changes requested):

frontend/cypress/integration/common/istio_config.ts
```diff
@@ -18,6 +18,25 @@ function labelsStringToJson(labelsString: string) {
...
+      zes: 'ze',
+      ses: 's',
+      es: 'e',
+      s: ''
+  };
+  return word.replace(
+      new RegExp(`(${Object.keys(endings).join('|')})$`), 
+      r => endings[r]
+  );
+}
```
matejnesuta, line 38: Hopefully I refactored everything correctly. I am not sure about this function though due to reasons mentioned in the code. On the other hand, I feel like installing a JS library or having tens of lines just to support this one test would be too much.

hhovsepy:
@matejnesuta did this PR passed on OCP test environment ?

matejnesuta:
@hhovsepy Yes. I use OCP most of the time when writing tests.

hhovsepy (approved):
LGTM

Issue #5805, opened 2023-02-06
Remove @webcomponents/custom-elements from package.json since Firefox ESR supports it natively


ferhoyos:
Library @webcomponents/custom-elements was added to polyfill window.customElements since it was required for Firefox ESR (Extended Support Release) some time ago. According to https://developer.mozilla.org/en-US/docs/Web/API/Window/customElements, window.customElements is supported from Firefox version 62 and current Firefox ESR version is 102. Then this library is no longer needed since Firefox ESR supports this feature natively.

jshaughn:
This work is actually being done by a comunity member, https://github.com/MRafaydev.  @ferhoyos , I assigned you to help make sure this contribution moves forward.

PR #5800, opened 2023-02-05
Update Kiali to build with Go 1.19

MRafaydev:
** Describe the change **
This change updates the Kiali codebase to use Go 1.19 instead of Go 1.18 as the build environment. The reason for this change is that Go 1.18 is considered deprecated and is no longer receiving patches, making it less stable and compatible than more recent versions of the Go language.

_explanation of what it does_
This pull request updates the Kiali codebase to build with Go 1.19 instead of Go 1.18. Go 1.18 is should be considered as deprecated and no longer receives patches, so updating to a newer and maintained version will ensure stability and compatibility for Kiali.

Changes:
- Update the build scripts to specify Go 1.19 as the target version.
- Test the build process and application functionality with Go 1.19.
- have tested these changes on multiple platforms and confirmed that Kiali builds and works as expected with Go 1.19.

** Issue reference **
this issue is reference to https://github.com/kiali/kiali/issues/5799

PR #5806, opened 2023-02-06
Remove unnecessary @webcomponents/custom-elements package

MRafaydev:
** Describe the change **
This pull request removes the library "@webcomponents/custom-elements", which was added as a polyfill for the `window.customElements` API for older versions of Firefox ESR. According to https://developer.mozilla.org/en-US/docs/Web/API/Window/customElements, the current version of Firefox ESR (102) now supports this API natively, so the library is no longer needed and can be removed.

** Issue reference **
Fixes https://github.com/kiali/kiali/issues/5805


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
jmazzitelli, line 1: This file should not be here.

node_modules/.yarn-integrity
```diff
@@ -0,0 +1,10 @@
+{
```
jmazzitelli, line 1: What is this file? I do not think this should be here.

jmazzitelli:
@MRafaydev The frontend code of Kiali is in the [frontend/ directory](https://github.com/kiali/kiali/tree/master/frontend)  -- the fact that `node_modules` and `yarn.lock` are in your root Kiali directory that you pushed in this PR means your dev set up is not correct (those files are in `frontend/` and in Kiali project we do not check in `node_modules` at all - those should be git ignored as you see [here](https://github.com/kiali/kiali/blob/master/frontend/.gitignore#L4)).

Before submitting PRs against the Kiali code base, it is best to become familiar with it and its development environment. I suggest reading through and understanding these files:

- https://github.com/kiali/kiali/blob/master/README.adoc#developer-setup
- https://github.com/kiali/kiali/blob/master/frontend/README.adoc

If you see an issue that you would like to try to fix, feel free to submit a [Q&A discussion on the topic](https://github.com/kiali/kiali/discussions) and the Kiali team can help walk you through it.

yarn.lock
jshaughn: These files show up if you accidentally execute `yarn install` at the top level `kiali` directory.

MRafaydev:
As a beginner, I apologize for making this mistake and will make sure to
follow the proper development setup as outlined in the provided links. I
appreciate your guidance and will make sure to review the Kiali development
environment before submitting any future PRs. Thank you for taking the time
to point out this error and help me improve my contributions to the project.

Makefile
```diff
@@ -29,7 +29,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
-GO_VERSION_KIALI = 1.18.7
+GO_VERSION_KIALI = 1.19.5
```
ferhoyos, line 32: This change does not correspond to the purpose of the PR, but to #5800 . Please try to keep separate branches when working on multiple PRs at the same time.

ferhoyos (changes requested):
Also you need to remove @webcomponents/custom-elements reference in polyfills.ts file

frontend/src/polyfills.ts
```diff
@@ -1,3 +1,2 @@
 // Polyfills window.customElements; currently required for Firefox ESR
```
ferhoyos, line 1: Please remove the comments too.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/kiali/kiali
 
-go 1.18
+go 1.19
```
jmazzitelli, line 3: this change does not belong in this PR. It is in the other PR branch.

jmazzitelli (changes requested):
`frontend/yarn.lock` is deleted but should not be. This needs to remain.

This PR really needs to be cleaned up before review/acceptance.

MRafaydev:
Can i raise the new clean PR ?

jmazzitelli:
> Can i raise the new clean PR ?

yes.

PR #5814, opened 2023-02-06
[hack] support dev images in mc script

jmazzitelli:
be able to build your own kiali dev image and have that be used in the MC set up.

See the `--kiali-use-dev-image` option. You need to make sure you build a dev image on your local box (`make build build-ui`) and then ensure its container is built and tagged as `quay.io/kiali/kiali:dev` (which is the default, anytime you do `make container-build` or `make cluster-push` this happens, you can even just do `make container-build-kiali` to just build the server image).

### DIFF
--- a/hack/istio/multicluster/deploy-kiali.sh
+++ b/hack/istio/multicluster/deploy-kiali.sh
@@ -14,7 +14,11 @@ if [ "${KIALI_ENABLED}" != "true" ]; then
   echo "Will not install kiali"
   return 0
 else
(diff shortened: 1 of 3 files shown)


jmazzitelli:
to use this:

`make build build-ui`

Then you can create the container:

`make container-build-kiali`

At this point, you can run the `install-everything.sh` script with the `--kiali-use-dev-image true` (or `-kudi true`).

leandroberetta (approved):
Looks good, tested locally.

Issue #5815, opened 2023-02-07
Format frontend code with prettier standards


ferhoyos:
Due to #5803 , currently most of the frontend files do not follow prettier format code standards. 

One-time manual code formatting is needed. Afterwards the pre-commit hook fixed in #5804 will format the files properly before commit.

PR #5816, opened 2023-02-07
Format frontend code with prettier

ferhoyos:
** Describe the change **

Format frontend code with prettier standards

** Issue reference **

#5815


jshaughn (approved):
This looks good.  I spot checked a few files against what happens if I reformat within vscode and I see the same results.  So this is good, it should reduce future diffs to just code changes and not prettier changes.

Issue #5429, opened 2022-09-01
[hack] use the official istioctl uninstall command now


jmazzitelli:
Istioctl uninstall is now officially supported: https://istio.io/latest/news/releases/1.15.x/announcing-1.15/#istioctl-uninstall

We should update our hack script so it always uses it now. For where this needs to be done, see: https://github.com/kiali/kiali/blob/v1.55.1/hack/istio/install-istio-via-istioctl.sh#L391-L399

PR #5819, opened 2023-02-07
Remove experimental scope for istio uninstall command

ferhoyos:
** Describe the change **

Since Istioctl uninstall is now officially supported, it is possible to remove experimental scope (x) from istio uninstall command.

** Issue reference **

[hack] use the official istioctl uninstall command now (#5429) 

** Testing **

Execute './hack/istio/install-istio-via-istioctl.sh -di true' to uninstall istio with stable command.

### DIFF
--- a/hack/istio/install-istio-via-istioctl.sh
+++ b/hack/istio/install-istio-via-istioctl.sh
@@ -410,12 +410,10 @@ if [ "${DELETE_ISTIO}" == "true" ]; then
   echo Deleting Core Istio
   if [ "${PURGE_UNINSTALL}" == "true" ]; then
     echo "Purging all Istio resources"
-    # Although the 'uninstall' command has been available since Istio 1.7 and is listed in the official istio doc,
-    # it's still available in 'experimental' (x) scope, so it might change in the future.
     # The optional --purge flag will remove all Istio resources, including cluster-scoped resources that may be shared with other Istio control planes.
-    ${ISTIOCTL} x uninstall --purge -y
+    ${ISTIOCTL} uninstall --purge -y
   else
-    ${ISTIOCTL} manifest generate --set profile=${CONFIG_PROFILE} ${MANIFEST_CONFIG_SETTINGS_TO_APPLY} | ${CLIENT_EXE} delete -f -
+    ${ISTIOCTL} manifest generate --set profile=${CONFIG_PROFILE} ${MANIFEST_CONFIG_SETTINGS_TO_APPLY} | ${CLIENT_EXE} delete --ignore-not-found=true -f -
   fi
   if [[ "${CLIENT_EXE}" = *"oc" ]]; then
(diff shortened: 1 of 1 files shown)


hack/istio/install-istio-via-istioctl.sh
```diff
@@ -410,12 +410,10 @@ if [ "${DELETE_ISTIO}" == "true" ]; then
...
   if [ "${PURGE_UNINSTALL}" == "true" ]; then
     echo "Purging all Istio resources"
-    # Although the 'uninstall' command has been available since Istio 1.7 and is listed in the official istio doc,
-    # it's still available in 'experimental' (x) scope, so it might change in the future.
     # The optional --purge flag will remove all Istio resources, including cluster-scoped resources that may be shared with other Istio control planes.
-    ${ISTIOCTL} x uninstall --purge -y
+    ${ISTIOCTL} uninstall --purge -y
   else
-    ${ISTIOCTL} manifest generate --set profile=${CONFIG_PROFILE} ${MANIFEST_CONFIG_SETTINGS_TO_APPLY} | ${CLIENT_EXE} delete -f -
+    ${ISTIOCTL} manifest generate --set profile=${CONFIG_PROFILE} ${MANIFEST_CONFIG_SETTINGS_TO_APPLY} | ${CLIENT_EXE} delete --ignore-not-found=true -f -
```
jmazzitelli, line 416: I think we still want the ignore-not-found... this could mean something got deleted already.  There is no need to fail this command if a resource doesn't exist - that's what we want anyway :) As long as the resources don't exist, that is all we care about (whether this command deleted it or something else did, it does not matter).

jmazzitelli (changes requested):

hack/istio/install-istio-via-istioctl.sh
ferhoyos: Yes, --ignore-not-found property is specified in istio documentation (https://istio.io/latest/docs/setup/install/istioctl/#uninstall-istio) and I thought the same as you (no need to fail the command if a resources does not exist).

jmazzitelli: Ha! I think I read that wrong. I thought you took it OUT.. lol... I'm not paying attention. I didn't realize you added that.

I approve this PR :)

jmazzitelli: For the record, the uninstall went into "supported" mode in v1.15 of Istio:

notice the 1.14 docs vs. 1.15 docs:
* https://istio.io/v1.14/docs/setup/install/istioctl/#uninstall-istio
* https://istio.io/v1.15/docs/setup/install/istioctl/#uninstall-istio

so this uninstall will only work with 1.15 and up. Just an FYI.

jmazzitelli (approved):

Issue #5799, opened 2023-02-03
Build Kiali with a newer minor version of Golang


israel-hdez:
Kiali is currently being built with Golang 1.18. However, Go 1.20 was released recently, which means that Go1.18 should be considered as deprecated and would not receive patches anymore.

Kiali should switch to be built using a newer/maintained Golang version. Perhaps, using golang v1.19 rather than the newest is a fair target for stability and compatibility reasons.

jshaughn:
@aljesusg , I assigned you to help https://github.com/MRafaydev with his contribution, as needed.

PR #5820, opened 2023-02-07
[1.48] Upgrade go version to 1.19.4

aljesusg:
Fix https://github.com/kiali/kiali/issues/5799


leandroberetta (approved):
LGTM

jmazzitelli (approved):
this looks correct, though I am surprised that there is no differences to the go.mod dependencies. I did not run the build to check.

aljesusg:
I tested it with build...the point is that from 1.48 to 1.57 we did some upgrades without execute go mod tidy...yes...we need the upgrade doc

PR #5821, opened 2023-02-07
[1.57] Upgrade go version to 1.19.4

aljesusg:
Fix https://github.com/kiali/kiali/issues/5799


aljesusg:
> I need to check this here the go.mod is different

It seems that no one made a go tidy here There are multiple versions

jmazzitelli (approved):
this looks correct

Issue #5720, opened 2023-01-13
Rework RBAC documentation


israel-hdez:
Current RBAC documentation is here: https://github.com/kiali/kiali.io/blob/6311dcf0fe8fe9aa8d39b48ae470d67935a4c607/content/en/docs/Configuration/rbac.md

This documentation was fine before the Kiali cache was introduced. When the Kiali cache was introduced, it became inaccurate as users just need read privileges to namespaces to gain visibility to several other resources through Kiali (i.e. people may just need what we have documented as the _minimum required privileges to login_ and Kiali will let users read everything else).

Probably, this was overlooked because the Kiali cache was opt-in at first. But when the cache transitioned to be turned on by default, the RBAC became "loose by default" and this isn't documented. So, docs are outdated already and are due for an update, although the documented behavior can still be obtained by turning off the Kiali cache.

However, since #5599 is already in progress and people won't be able to turn off the Kiali cache, the documented RBAC capabilities no longer stand and docs need to be updated.

PR #5824, opened 2023-02-07
RBAC-related fixes

israel-hdez:
I'm not sure if this is fair. I'm anyway suggesting these.

This is doing two fixes:
* In istio.go some messages are moved from ERR level to TRACE level. Apparently, when these funcs fail because of the user not having enough privileges, Kiali will retry using its ServiceAccount. So, these messages are not that useful unless something weird happened.
* In proxy_status.go config dumps are now retrieved using the Kiali ServiceAccount rather than the token of the logged-in user. This is for the "Envoy" tab of Workloads to work properly if the user has "less" privileges.

Related #5720

### DIFF
--- a/handlers/proxy_status.go
+++ b/handlers/proxy_status.go
@@ -4,13 +4,21 @@ import (
 	"net/http"
 
 	"github.com/gorilla/mux"
+	kiali_business "github.com/kiali/kiali/business"
+	"github.com/kiali/kiali/kubernetes"
+	"k8s.io/client-go/tools/clientcmd/api"
 )
 
 func ConfigDump(w http.ResponseWriter, r *http.Request) {
 	params := mux.Vars(r)
 
 	// Get business layer
-	business, err := getBusiness(r)
+	kialiToken, err := kubernetes.GetKialiToken()
+	if err != nil {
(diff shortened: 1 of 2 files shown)


handlers/proxy_status.go
```diff
@@ -4,13 +4,22 @@ import (
...
+	kiali_business "github.com/kiali/kiali/business"
+	"github.com/kiali/kiali/kubernetes"
 )
 
 func ConfigDump(w http.ResponseWriter, r *http.Request) {
 	params := mux.Vars(r)
 
 	// Get business layer
-	business, err := getBusiness(r)
+	kialiToken, err := kubernetes.GetKialiToken()
```
jmazzitelli, line 17: I'm just documenting this here for me (and this is why I want to get my PRs merged soon)... this `GetKialiToken() `method will be renamed in the future to `GetKialiTokenForHomeCluster()`.

hhovsepy (approved):
Did smoke test on OCP cluster, lgtm.

jmazzitelli:
I'm OK to merge this now. @israel-hdez merge this if you are ready and then I'll update my PR to fix a build error that it will have after this goes in.

jmazzitelli (approved):

PR #5825, opened 2023-02-08
Change compatible node version in package.json

### LINKED ISSUE #5208: Wrong engine node version in package.json
After upgrade node to the latest one I got the message that node engine must be >=12.22.0 and < 16. this is great because we set in the [package.json](https://github.com/kiali/kiali/blob/master/frontend/package.json#L143):

  **"node": ">=12.22.0 && <16"**

So I installed the latest 15.X 15.14.0  (latest one of 15) and I got :

`error npmlog@6.0.2: The engine "node" is incompatible with this module. Expected version "^12.13.0 || ^14.15.0 || >=16.0.0". Got "15.14.0"`

So I installed 14.19.3 (latest one of 14) and worked ...So we need to change the node engine to  ">=12.22.0 && <15"

ferhoyos:
** Describe the change **

Change compatible node version in package.json to avoid libraries incompatible messages of node version.

** Issue reference **

Fixes #5208 

** Test **

Install node from version v14.17 to check that there are not incompatible node version errors on 'yarn install' execution in frontend folder. Any other node version (13, 15 or less than 14.17) will bring incompatible node version errors.


aljesusg (approved):
LGFM

PR #5826, opened 2023-02-08
Change log level to debug in openid auth controller

josunect:
openid authentication was raising an error message even when the authentication is successful. The message is useful for diagnose purposes just in case the authentication failed, so changing the message log level to debug.


israel-hdez (approved):
All good!

Issue #5811, opened 2023-02-06
[Cypress] Improve test coverage for Istio Config Wizards


josunect:
Improve coverage to validate changes in https://github.com/kiali/kiali/issues/5741

- There is already coverage for K8s Gateway 
- Improve coverage including Gateway and Service Entry wizard tests

matejnesuta:
Working on it

PR #5835, opened 2023-02-10
[Cypress] Istio Config page wizard coverage

matejnesuta:
Tests related to the #5811 issue. This suite increases coverage of the wizard menu for the ServiceEntries and Gateways in the Istio config section of Kiali. The tests cover creation of said objects, tests of the text fields for valid and invalid values and also tests that object of same name cannot be created twice. Currently not sure if I should implement a teardown of some sort to clean the created objects after the suite is completed.


hhovsepy (approved):
Impressive work @matejnesuta thank you

PR #5837, opened 2023-02-11
Optimize the code for better performance and clarity

MRafaydev:
This pull request makes the following optimizations to the code:

**Optimization of the Date.toLocaleStringWithConditionalDate function:**

- The code has been optimized to store the result of new Date().toLocaleDateString() in a variable, nowDate, and use it both times, reducing the number of redundant operations and making the code more efficient.

**Optimization of the ReactDOM.render function:**

- The code now first retrieves the root element using document.getElementById('root'). If root exists, the ReactDOM.render function is then called with <App /> as the first argument and root as the second argument. This change ensures that the ReactDOM.render function is only called if root exists, avoiding an error in case the element does not exist in the document.

**Benefits:**
_**The optimizations made in this pull request result in the following benefits:**_

- Increased efficiency in the Date.toLocaleStringWithConditionalDate function by reducing redundant operations.
- Avoidance of errors in the ReactDOM.render function by only calling it if the root element exists.

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -3,18 +3,29 @@ import * as ReactDOM from 'react-dom';
 import App from './app/App';
 import './styles/index.css';
 
+// Declare a new interface for the Date object to add the new method toLocaleStringWithConditionalDate
 declare global {
   interface Date {
     toLocaleStringWithConditionalDate(): string;
   }
 }
 
+// Extend the Date object to add the new method toLocaleStringWithConditionalDate
 // eslint-disable-next-line no-extend-native
 Date.prototype.toLocaleStringWithConditionalDate = function () {
-  const nowDate = (new Date()).toLocaleDateString();
+  // Get the locale date string for the current date
+  const nowDate = new Date().toLocaleDateString();
+  // Get the locale date string for the date this method is being called on
   const thisDate = this.toLocaleDateString();
 
+  // Return the locale time string if the dates are the same, otherwise return the locale string
   return nowDate === thisDate ? this.toLocaleTimeString() : this.toLocaleString();
 };
 
-ReactDOM.render(<App />, document.getElementById('root') as HTMLElement);
+// Get the element with id "root"
+const rootElement = document.getElementById('root');
+
+// If the root element exists, render the App component to it
+if (rootElement) {
+  ReactDOM.render(<App />, rootElement);
+}


frontend/src/index.tsx
```diff
@@ -3,18 +3,29 @@ import * as ReactDOM from 'react-dom';
...
 // eslint-disable-next-line no-extend-native
 Date.prototype.toLocaleStringWithConditionalDate = function () {
-  const nowDate = (new Date()).toLocaleDateString();
+  // Get the locale date string for the current date
+  const nowDate = new Date().toLocaleDateString();
+  // Get the locale date string for the date this method is being called on
   const thisDate = this.toLocaleDateString();
 
+  // Return the locale time string if the dates are the same, otherwise return the locale string
   return nowDate === thisDate ? this.toLocaleTimeString() : this.toLocaleString();
```
jmazzitelli, line 22: > The code has been optimized to store the result of new Date().toLocaleDateString() in a variable, nowDate, and use it both times, reducing the number of redundant operations and making the code more efficient.
 
This code change is effectively a no-op, it doesn't seem to do anything - I don't see any of the optimization that is claimed in the description of the PR. All I see are added comments (lines 13, 16, 18, 21) and you removed the parens on the change to line 17 - but that has already been done in the [previous commit ](https://github.com/kiali/kiali/pull/5816) that ran `prettier` to clean that up (see master branch of this file: https://github.com/kiali/kiali/blob/master/frontend/src/index.tsx). There is nothing materially different otherwise that I can see.

frontend/src/index.tsx
```diff
@@ -3,18 +3,29 @@ import * as ReactDOM from 'react-dom';
...
 };
 
-ReactDOM.render(<App />, document.getElementById('root') as HTMLElement);
```
jmazzitelli, line 20: > Avoidance of errors in the ReactDOM.render function by only calling it if the root element exists.

But don't we WANT errors in this case (and specifically NOT avoid errors)?

I'm pretty sure if the `root` element doesn't exist, we want this to immediately fail. Without the `App/` rendered, the Kiali UI will fail. If there is a missing `root` element, that's a major failure and we'd most likely want this to fail very loudly and quickly. Therefore, I see nothing advantageous to this code change - in fact, it might even make debugging much harder in the case of a missing `root` element because if a `root` element doesn't exist, this code will silently continue on even though I'm fairly sure Kiali won't work at all without that `App/` being rendered.

jmazzitelli (changes requested):
I don't see any qualities in this PR that would make me think we would want to merge this. See my comments on the 2 main changes to this PR.

In addition, this PR is not mergeable at all anyway because it is not rebased on the current master branch and there are git conflicts that need to be resolved.

jshaughn:
@MRafaydev, thank you for your contribution but in general the Kiali maintainers don't merge optimizations on working code unless it addresses a quantified performance problem, documented as an accepted issue on the project backlog, and shows proof that it solves the problem.   We do appreciate your interest in the project but for future contributions please look for issues already identified in the github issues, especially things labeled as "good first issue" or "help wanted".  If you have something you'd like to see, please add a new github issue, and then we can work with you on whether it's appropriate before you start work.  Please see the [Contributing](https://github.com/kiali/kiali/blob/master/CONTRIBUTING.md) document, which explains the above in a bit more detail.  I am closing this PR as it is does not solve an accepted issue, but we do welcome contributions from you in the future!  Thanks again for this effort.

PR #5838, opened 2023-02-13
Rename kiali_api.md to kiali_internal_api.md

### LINKED ISSUE #5544: remove kiali_api.md
The Kiali API is not for public use, is not supported for public use, and can change from version to version with no guarantee or promise of backward compatibility.

So why do we have [kiali_api.md](https://github.com/kiali/kiali/blob/v1.57.1/kiali_api.md) in the repo?

We even ripped out the auto-generation of that file via swagger a while ago. There is no way we are going to manage to keep that file up to date if we have to manually do it, and people might see it and think (a) it is up to date and accurate and (b) they can use the API outside of the Kiali UI.

Thus, we should delete kiali_api.md entirely.

 Comments?

ferhoyos:
** Describe the change **

Rename kiali_api.md to kiali_internal_api.md to be clear that kiali api is not public and can change from version to version.

** Issue reference **

Closes #5544.


jshaughn (changes requested):
Hi @ferhoyos, this turned out to be a little more complicated :-).  Instead of just renaming the old file we should delete it and regenerate it with the new name.  But, as it turns out we no longer have a `make` command to do it.  So instead, let's just document how to regenerate the file inside the API md itself.    I have submitted a PR to your PR: https://github.com/ferhoyos/kiali/pull/2, please accept that and then follow the directions in the file to regenerate the api markdown.  Then we'll update this PR with the new file,

jshaughn (approved):
LG2M!

PR #5839, opened 2023-02-14
Remove check for response type id_token

### LINKED ISSUE #5833: reponse type id_token check in Open ID authenticacion is no longer needed
With the removal of support for the implicit code flow for Open ID authentication, the check for the response type id_token is no longer needed. At the moment it is checked and it throws a warning when it is not supported:

https://github.com/kiali/kiali/blob/master/business/authentication/openid_auth_controller.go#L1159

josunect:
Fixes #5833 

Removed support for implicit flow in OpenID


israel-hdez (approved):
Looks good!

josunect:
> Looks good!

Thanks for the review, @israel-hdez !

PR #5840, opened 2023-02-14
Fix port number and add unique identifiers for in some wizards

### LINKED ISSUE #5834: Duplicate IDs for the text inputs + high port value breaks the UI (Istio Config wizard page)
Some things I noticed while I was working today on the #5811 issue.

On the wizard page for ServiceEntry, I wanted to do a test on a 2 port specification with a same port name on the Istio Config wizard page. The issue here is that when multiple port specifications are present, the IDs in the HTML code for the text inputs and selects of the same category are identical. 
image

On the wizard page for the Gateway, a port number little higher than 65535 will break the UI.
image

josunect:
Fixes #5834 

- Use constant for MAX and MIN port numbers
- Use unique identifiers by row in form wizards
- Fix React warning

### DIFF
--- a/frontend/cypress/integration/common/wizard_istio_config.ts
+++ b/frontend/cypress/integration/common/wizard_istio_config.ts
@@ -23,7 +23,7 @@ And('user adds listener', () => {
 });
 
 And('user types {string} in the name input', (name) => {
-  cy.get('input[id="name0"]').type(name);
+  cy.get('input[id="name"]').type(name);
 });
 
 And('user types {string} in the add listener name input', (name) => {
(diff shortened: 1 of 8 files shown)


hhovsepy:
It seems ServiceEntryForm.tsx also needs the constant MAX_PORT to be consistent:

hhovsepy (approved):
Verified now.

josunect:
> Verified now.

Thanks for your review, @hhovsepy !

hhovsepy (approved):

matejnesuta:
LGTM

PR #5841, opened 2023-02-14
Download button for debug info

### LINKED ISSUE #4276: Add 'Download" option for Debug information
It would be helpful to add a 'Download' (to file) option to the Kiali Debug information view, instead of having to copy/paste into a file.

ferhoyos:
** Describe the change **

Add 'Download" button for Debug information

image

** Issue reference **

Closes #4276

### DIFF
--- a/frontend/src/components/DebugInformation/DebugInformation.tsx
+++ b/frontend/src/components/DebugInformation/DebugInformation.tsx
@@ -118,6 +118,15 @@ export class DebugInformation extends React.PureComponent<DebugInformationProps,
     this.setState({ copyStatus: result ? CopyStatus.COPIED : CopyStatus.NOT_COPIED });
   };
 
+  download = () => {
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
Looks good, working on Chrome and FF.  I think the buttons on this dialog and the other `?` dropdowns aren't quite styled correctly, or at least consistently, with other dialogs.  Not a blocker, just a comment.

ferhoyos:
You are right @jshaughn. For example in IstioConfigPreview modal, the copy and download buttons are icons shown at the top of the modal:

image

All modals within the application should be consistent (or icon buttons at the top or normal button at the footer). IMO the best option is to place all action buttons in the footer, the icon buttons seem very isolated and do not look good to me. I will create an issue to track this.

Another thing I do not understand is why we have a Close button in the footer when the modal already provides close icon at the top...

PR #5842, opened 2023-02-14
Adapt apps_test.go to not rely on turning off the cache

israel-hdez:
The adaptation moves from using mocks to using a fake client. This allows for starting a cache where the underlying client is a fake, making it possible to allow the code to use "caching" as with the real client.

Related #5599

### DIFF
--- a/business/apps_test.go
+++ b/business/apps_test.go
@@ -4,54 +4,72 @@ import (
 	"context"
 	"testing"
 
-	osapps_v1 "github.com/openshift/api/apps/v1"
 	osproject_v1 "github.com/openshift/api/project/v1"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
-	apps_v1 "k8s.io/api/apps/v1"
-	batch_v1 "k8s.io/api/batch/v1"
 	core_v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


business/apps_test.go
```diff
@@ -4,54 +4,72 @@ import (
...
 
-func setupAppService(k8s *kubetest.K8SClientMock) *AppService {
+func setupAppService(k8s *kubetest.FakeK8sClient) *AppService {
 	prom := new(prometheustest.PromClientMock)
 	layer := NewWithBackends(k8s, prom, nil)
 	setupGlobalMeshConfig()
 	return &AppService{k8s: k8s, prom: prom, businessLayer: layer}
 }
 
+func setupTestingKialiCache(k8s *kubetest.FakeK8sClient, conf *config.Config, require *require.Assertions) func() {
```
josunect, line 28: Not sure, if this can be reused for other tests and might be placed as a common utility?

israel-hdez: Quite possible. I was thinking in moving this to a shared place in a follow-up PR, if needed. But I think I won't do more work on the tests. Let me check where I can move this and change this PR.

josunect: I think it is Ok to change it in a follow up PR

josunect (approved):

israel-hdez:
Thanks @josunect for the review!

PR #5845, opened 2023-02-15
[1.48] disable linter

jmazzitelli:
the linter used in this branch does not work with go 1.19. If we upgrade, we'll get a bunch of errors from the linter that we do not want to fix in this branch. So this PR disables the linter.


jshaughn (approved):
Yup

PR #5851, opened 2023-02-16
Update the Namespace cache to support multicluster

### LINKED ISSUE #5766: Update the namespace cache which caches the user token per namespace to support multicluster
The namespaces cache saves a user's token and the namespaces they can access so that multiple requests to `GetNamespaces` don't need to be repeatedly made. This cache should support caching namespaces by cluster/namespace.

https://github.com/kiali/kiali/blob/b3310862c6fa8ce90e0f0a6e37dac72545d1dddf/kubernetes/cache/cache.go#L62-L67

I will link the proposal:
https://github.com/kiali/kiali/blob/master/design/KEPS/multicluster/proposal.md#api

josunect:
Fixes #5766 

The first approach was to change the cache: 

`tokenNamespaces        map[string]namespaceCache`

For the following: 

`tokenNamespaces        map[string]map[string]namespaceCache // By token, by cluster`

The purpose of this would be to access the information easily of quickly in case we need to reload just one cluster information. And the information will be mixed before returning. 

The other approach it is to just save the information as it was. 

Another change will be to include the cluster inside the namespace model. It can help to filter the information in the GUI: 

`Cluster string json:"cluster"`

Still need to figure out how to return the information in the API. ATM, the API 7namespaces returns: 
`
[{"name":"alpha","cluster":"_kiali_home","labels":{"istio-injection":"enabled","kubernetes.io/metadata.name":"alpha"},"annotations":{}},{"name":"beta","cluster":"_kiali_home","labels":{"istio-injection":"enabled","kubernetes.io/metadata.name":"beta"},"annotations":{}}, ...`

It can merge all the information: 

```
{"name":"NSNAME", 
  "clusters": [ "clustername", "clustername2"],
{labels, annotations}}
```

`{"name":"alpha","clusters":["_kiali_home", "east"],"labels":{"istio-injection":"enabled","kubernetes.io/metadata.name":"alpha", "kubernetes.io/metadata.name":"new_metadata"},"annotations":{}}`

Or be returned by cluster: 

```
{"name":"NSNAME", 
  "clusters": [ {"clustername" : {"abels, annotations"}}]}
```

```
{"name":"alpha",
"clusters":[{"_kiali_home": {"labels":{"istio-injection":"enabled","kubernetes.io/metadata.name":"alpha"},"annotations":{}}}, 
{"east": {"labels":{"istio-injection":"enabled","kubernetes.io/metadata.name":"alpha", "kubernetes.io/metadata.name":"new_metadata"},"annotations":{}}}]
```

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -443,11 +443,16 @@ func (in *NamespaceService) GetNamespace(ctx context.Context, namespace string) 
 		result = models.CastProject(*project)
 	} else {
 		var ns *core_v1.Namespace
-		ns, err = in.k8s.GetNamespace(namespace)
-		if err != nil {
-			return nil, err
+		var cl string
+		for _, cluster := range clientFactory.GetClusterNames() {
+			kialiClient := clientFactory.GetSAClient(cluster)
+			ns, err = kialiClient.GetNamespace(namespace)
+			if err == nil {
+				cl = cluster
+				break
+			}
 		}
-		result = models.CastNamespace(*ns, "")
+		result = models.CastNamespace(*ns, cl)
 	}
 	// Refresh cache in case of cache expiration
 	if kialiCache != nil {
@@ -526,7 +531,7 @@ func (in *NamespaceService) getNamespacesUsingKialiSA(cluster string, failedClie
 }
 
 func (in *NamespaceService) getNamespacesForKialiSA(cluster string, labelSelector string) ([]core_v1.Namespace, error) {
-	if cluster == clientFactory.GetHomeClusterName() {
+	if cluster == clientFactory.GetHomeClusterName() || cluster == "" {
 		clientFactory, err := kubernetes.GetClientFactory()
 		if err != nil {
 			return nil, err
@@ -549,7 +554,19 @@ func (in *NamespaceService) getNamespacesForKialiSA(cluster string, labelSelecto
 
 		return nss, nil
 	} else {
-		return nil, nil
+		client := clientFactory.GetSAClient(cluster)
+
+		k8s, err := clientFactory.GetClient(&api.AuthInfo{Token: client.GetToken()})
+		if err != nil {
+			return nil, err
+		}
+
+		nss, err := k8s.GetNamespaces(labelSelector)
+		if err != nil {
+			return nil, err
+		}
+
+		return nss, nil
 	}
 
 }
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -114,8 +114,6 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 	// (Side note 3: The control plane namespace is always included via api.namespaces.include and
 	// never excluded via api.namespaces.exclude or api.namespaces.label_selector_exclude.)
 
-	labelSelectorInclude := configObject.API.Namespaces.LabelSelectorInclude
-
 	// determine if we are to exclude namespaces by label - if so, set the label name and value for use later
 	labelSelectorExclude := configObject.API.Namespaces.LabelSelectorExclude
 	var labelSelectorExcludeName string
@@ -130,118 +128,14 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 	}
 
 	namespaces := []models.Namespace{}
-	_, queryAllNamespaces := in.isAccessibleNamespaces["**"]
 
-	// If we are running in OpenShift, we will use the project names since these are the list of accessible namespaces
-	if in.hasProjects {
-		projects, err2 := in.k8s.GetProjects(labelSelectorInclude)
-		if err2 == nil {
-			// Everything is good, return the projects we got from OpenShift
-			if queryAllNamespaces {
-				namespaces = models.CastProjectCollection(projects)
-
-				// add the namespaces explicitly included in the include list.
-				includes := configObject.API.Namespaces.Include
-				if len(includes) > 0 {
-					var allNamespaces []models.Namespace
-					var seedNamespaces []models.Namespace
-
-					if labelSelectorInclude == "" {
-						// we have already retrieved all the namespaces, but we want only those in the Include list
-						allNamespaces = namespaces
-						seedNamespaces = make([]models.Namespace, 0)
-					} else {
-						// we have already got those namespaces that match the LabelSelectorInclude - that is our seed list.
-						// but we need ALL namespaces so we can look for more that match the Include list.
-						if allProjects, err := in.k8s.GetProjects(""); err != nil {
-							return nil, err
-						} else {
-							allNamespaces = models.CastProjectCollection(allProjects)
-							seedNamespaces = namespaces
-						}
-					}
-					namespaces = in.addIncludedNamespaces(allNamespaces, seedNamespaces)
-				}
-			} else {
-				filteredProjects := make([]osproject_v1.Project, 0)
-				for _, project := range projects {
-					if _, isAccessible := in.isAccessibleNamespaces[project.Name]; isAccessible {
-						filteredProjects = append(filteredProjects, project)
-					}
-				}
-				namespaces = models.CastProjectCollection(filteredProjects)
+	for _, cluster := range clientFactory.GetClusterNames() {
+		kialiClient := clientFactory.GetSAClient(cluster)
+		nsList, error := in.GetNamespacesByClient(kialiClient, cluster)
+		if error == nil {
+			for _, ns := range nsList {
+				namespaces = append(namespaces, ns)
 			}
-		}
-	} else {
-		// if the accessible namespaces define a distinct list of namespaces, use only those.
-		// If accessible namespaces include the special "**" (meaning all namespaces) ask k8s for them.
-		// Note that "**" requires cluster role permission to list all namespaces.
-		accessibleNamespaces := configObject.Deployment.AccessibleNamespaces
-
-		if queryAllNamespaces {
-			nss, err := in.k8s.GetNamespaces(labelSelectorInclude)
-			if err != nil {
-				// Fallback to using the Kiali service account, if needed
-				if errors.IsForbidden(err) {
-					if nss, err = in.getNamespacesUsingKialiSA(labelSelectorInclude, err); err != nil {
-						return nil, err
-					}
-				} else {
-					return nil, err
-				}
-			}
-
-			namespaces = models.CastNamespaceCollection(nss)
-
-			// add the namespaces explicitly included in the includes list.
-			includes := configObject.API.Namespaces.Include
-			if len(includes) > 0 {
-				var allNamespaces []models.Namespace
-				var seedNamespaces []models.Namespace
-
-				if labelSelectorInclude == "" {
-					// we have already retrieved all the namespaces, but we want only those in the Include list
-					allNamespaces = namespaces
-					seedNamespaces = make([]models.Namespace, 0)
-				} else {
-					// we have already got those namespaces that match the LabelSelectorInclude - that is our seed list.
-					// but we need ALL namespaces so we can look for more that match the Include list.
-					allK8sNamespaces, err := in.k8s.GetNamespaces("")
-					if err != nil {
-						// Fallback to using the Kiali service account, if needed
-						if errors.IsForbidden(err) {
-							if allK8sNamespaces, err = in.getNamespacesUsingKialiSA("", err); err != nil {
-								return nil, err
-							}
-						} else {
-							return nil, err
-						}
-					}
-					allNamespaces = models.CastNamespaceCollection(allK8sNamespaces)
-					seedNamespaces = namespaces
-				}
-				namespaces = in.addIncludedNamespaces(allNamespaces, seedNamespaces)
-			}
-		} else {
-			k8sNamespaces := make([]core_v1.Namespace, 0)
-			for _, ans := range accessibleNamespaces {
-				k8sNs, err := in.k8s.GetNamespace(ans)
-				if err != nil {
-					if errors.IsNotFound(err) {
-						// If a namespace is not found, then we skip it from the list of namespaces
-						log.Warningf("Kiali has an accessible namespace [%s] which doesn't exist", ans)
-					} else if errors.IsForbidden(err) {
-						// Also, if namespace isn't readable, skip it.
-						log.Warningf("Kiali has an accessible namespace [%s] which is forbidden", ans)
-					} else {
-						// On any other error, abort and return the error.
-						return nil, err
-					}
-				} else {
-					k8sNamespaces = append(k8sNamespaces, *k8sNs)
-				}
-			}
-			namespaces = models.CastNamespaceCollection(k8sNamespaces)
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


jshaughn:
> The purpose of this would be to access the information easily of quickly in case we need to reload just one cluster information. And the information will be mixed before returning.

I think this data structure will make sense for the reasons you say; it would make it easy to update cache for a specific cluster, or to perform backend filtering on cluster.  I'm not sure we will do backend filtering on cluster, but if it's needed it would be easy.   Also, we may not have to put the cluster name on all of the cached objects, because the cluster name has been normalized out.

As for returning information in the API, it may not become clear until we start to see it, but I think maybe we want to de-normalize further.  You had the option of:

```
{"name":"NSNAME", 
  "clusters": [ {"clustername" : {"abels, annotations"}}]}

```
But maybe it should be:

`{"name":"NSNAME",  "cluster": "CLUSTERNAME" : {"labels, annotations"}}`

I could envision a separate card for each namespace-cluster combination.  But that may only make sense if we expect real differences between a namespace on different clusters, or a different workload, service, etc.  It's a question of whether we want to treat the entities on different clusters in a very distinct way, or a more unified way.  Right now I'm thinking distinct, but I'm really not sure.

business/namespaces.go
```diff
@@ -129,12 +127,93 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+		}
+	}
+	for _, cluster := range clientFactory.GetClusterNames() {
+		kialiClient := clientFactory.GetSAClient(cluster)
+		nsList, error := in.GetNamespacesByClient(kialiClient, cluster)
+		if error == nil {
+			for _, clusterNamespace := range nsList {
+				namespacesAcum := nsmap[clusterNamespace.Name]
+				if namespacesAcum.Name != "" {
+					// Merge data
```
jmazzitelli, line 149: we should log some information here if the data already exists from another cluster's namespace but is different from the current cluster's namespace.

For example, what happens if cluster A namespace has sidecar injection annotation set to true but cluster B namespace has the same sidecar annotation set to false? The last one "wins" but we'll never know of this difference. We should at least log this condition so we will have SOME indication of a mismatch. It would be important to know if the labels or annotations are different across clusters.

That said, I can also envision annotations and even labels having different values across clusters (think of things like "last-updated"  or "last-applied-configuration" type of annotations. Logging messages for those will get noisy. Perhaps log these at TRACE level?

My only concern is behavior is going to be different because the user didn't annotate and label the namespaces the same across clusters and we have no way of letting the user know that.

josunect: Maybe we can send a warning in the case the annotations are not the same? It would be an added value

jmazzitelli: Yes, I would say, in general, if the labels / annotations are not the same, then we should "say something" because it would be weird if we are treating the namespaces as "the same" but the user has labeled/annotated them differently.

I'm sure there will be cases where it is expected some labels/annotations will be different (just based on the fact they are on different clusters) but in general, I would think they should be basically the same and if they are not, we should log something about those differences.

leandroberetta: I agree that labels and annotations may vary because users possibly will label namespaces on different clusters with for example a "zone" label (east, west).

Maybe we can just gather those labels and annotations that are the same, and if in the UI, eventually we let users to zoom into any of these merged namespaces, we should fetch all the labels and annotations of it.

business/namespaces.go
```diff
@@ -129,12 +127,110 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+		if err != nil {
+			log.Errorf("Error getting client factory: %v", err.Error())
+			return nil, fmt.Errorf("Error getting client factory")
+		}
+	}
+
+	var clusterError error
+	for _, cluster := range clientFactory.GetClusterNames() {
+		kialiClient := clientFactory.GetSAClient(cluster)
+		nsList, error := in.GetNamespacesByClient(kialiClient, cluster)
```
leandroberetta, line 147: Here, maybe in the future we need to fetch the information in go routines and then do the merging.

leandroberetta (dismissed):
Hi Josune, nice work, I like the idea of having 1 namespaces with the cluster list inside. This approach is what we discussed in the UI meeting and I think is the direction we want to go.

I have left minor comments that can be addresses in the near future.

business/namespaces.go
josunect: I sent a last push. If we are not going to perform any logic in the frontend, maybe we can just concat the values when we have a different value for different clusters. This is an example: 

image

In the case of `topology.istio.io/network=[value1, value2]`

Not sure if this is a valid solution, though. 
The other approach would be to change the data structure returned for: 

- value of labels can be an array. Disadvantage: Maybe a bit more difficult to manipulate
- create a map for each cluster containing their annotations and labels. 

The disadvantage of this, it could be a little bit more difficult to manipulate

business/namespaces.go
josunect: Thanks for the suggestion! I think is a good idea, I have made a commit with the changes.

business/namespaces.go
jshaughn: I think I'd prefer a more verbose format that allows the UI more flexibility.  The UI is the presentation layer, and I think the API should probably make it easy to be flexible in how the data is visualized.

josunect: I was thinking about it and created this new branch: https://github.com/kiali/kiali/pull/5883 As it has some more code changes and I'm not sure if it goes in the right direction. I think it is a good approach, but I would like to have an approval first.

matejnesuta:
Hello. Does this PR need some kind of test coverage (integration suite, cypress suite)? My assumption is yes, but I am asking to be sure.

business/namespaces.go
```diff
@@ -129,12 +128,133 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+
+	wg := &sync.WaitGroup{}
+	type result struct {
+		cluster string
+		ns      []models.Namespace
+		err     error
+	}
+	resultsCh := make(chan result)
+
+	go func(ctx context.Context) {
```
hhovsepy, line 152: is the ctx used in function?

jshaughn:
> Hello. Does this PR need some kind of test coverage (integration suite, cypress suite)? My assumption is yes, but I am asking to be sure.

This sort of infrastructural change may seem so dedicated tests but ot may not, as it will be in the code paths of basically every integration test, I think.

business/namespaces.go
josunect: Thanks, @hhovsepy 
I've added the change

kubernetes/cache/namespaces.go
```diff
@@ -18,10 +18,16 @@ type (
...
 	c.tokenLock.Lock()
-	nameNamespace := make(map[string]models.Namespace, len(namespaces))
+	nameNamespace := make(map[string]map[string]models.Namespace)
 	for _, ns := range namespaces {
-		nameNamespace[ns.Name] = ns
+		for _, cluster := range ns.Clusters {
+			if nameNamespace[ns.Name] == nil {
+				nameNamespace[ns.Name] = make(map[string]models.Namespace)
+			}
+			nameNamespace[ns.Name][cluster] = ns
```
leandroberetta, line 27: For what I understand here, maybe I'm missing something, but for multiple clusters, the same namespace is stored in the map?.

I guess we need to store the namespace from the cluster because probably labels and annotations may vary between namespaces (with same name) across different clusters.

leandroberetta (commented):
Hi Josune, nice work. I have some comments, one is minor and the other one is related to namespaces and clusters.

kubernetes/cache/namespaces.go
josunect: This was something I forgot to change, but in a first approach I combined the namespace information and the namespace model was saving an array of namespaces where it lived, so this just reflects that situation. But you made a good point, as this needs to be updated.

business/namespaces.go
```diff
@@ -493,26 +584,23 @@ func (in *NamespaceService) getNamespacesUsingKialiSA(labelSelector string, forw
...
-	clientFactory, err := kubernetes.GetClientFactory()
-	if err != nil {
-		return nil, err
-	}
-
-	kialiToken, err := kubernetes.GetKialiTokenForHomeCluster()
-	if err != nil {
-		return nil, err
-	}
+func (in *NamespaceService) getNamespacesForKialiSA(cluster string, labelSelector string) ([]core_v1.Namespace, error) {
```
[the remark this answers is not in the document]
nrfox: After looking more into what's going on inside `GetNamespaces`, it looks like the kiali service account token is being used when `accessible_namespaces == **`. Kiali tries to do a `GetNamespaces` with the user client: https://github.com/kiali/kiali/blob/f3f431063e067dd58c0906008e9a4c915e75a6e4/business/namespaces.go#L182-L192.

If that fails, then Kiali does a `GetNamespaces` with the Kiali Service Account client to get the whole list of namespaces and then does `GetNamespace(...)` call against each individually to check the user's access to each one.
https://github.com/kiali/kiali/blob/f3f431063e067dd58c0906008e9a4c915e75a6e4/business/namespaces.go#L474-L490

When `accessible_namespaces == **` _and_ the logged in user doesn't have `List Namespaces` access to a cluster, the Namespace Service uses the Kiali service account client to retrieve the list of namespaces and check the user's access to each one. So the Namespace Service needs the Kiali service account client for this. IMO it's still best to pass the kiali service account clients into the namespace service rather than using the client factory from inside the Namespace Service. That'd look something like:
```
type NamespaceService struct {
  userClients map[string]ClientInterface
  kialiSAClients map[string]ClientInterface
}
```
that might require a change in `NewWithBackends`. Either passing in the clientfactory rather than `map[string]ClientInterface` or passing in two different client maps, one for the user clients and one for the sa clients.

business/namespaces.go
```diff
@@ -129,12 +127,103 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+		wg.Wait()
+		close(resultsCh)
+	}()
+
+	// Combine namespace data
+	for result := range resultsCh {
+		if result.err != nil {
+			log.Errorf("Error fetching Namespaces per cluster %s: %s", result.cluster, result.err)
+			// TODO: Return error if one failure?
+			return nil, result.err
```
nrfox, line 174: This is a little tricky for multicluster. The current single cluster implementation returns an error if `GetNamespaces` fails but for multicluster we don't necessarily want the user to know if there's a cluster they can't access that is part of the mesh. Perhaps when we initially create the user clients in the client factory we then do some kind of access check and if that fails then we don't include that cluster in the map of user clients?

business/apps.go
```diff
@@ -20,7 +20,7 @@ import (
 // AppService deals with fetching Workloads group by "app" label, which will be identified as an "application"
 type AppService struct {
 	prom          prometheus.ClientInterface
-	k8s           kubernetes.ClientInterface
+	k8s           map[string]kubernetes.ClientInterface
```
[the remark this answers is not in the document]
josunect: You are right, removed from AppService

business/layer.go
```diff
@@ -84,8 +87,30 @@ func initKialiCache() {
...
+							log.Errorf("Error getting client for cluster %s: %s", c, errClient)
+						}
+
+					} else {
+						log.Errorf("Error getting config for remote cluster info: %s", err2)
+					}
+				}
+			} else {
+				log.Errorf("Error getting cluster remote info: %s", errClusterInfo)
+			}
```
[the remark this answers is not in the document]
josunect: Created new method GetSAClients in clientFactory. The code looks much more cleaner now.

business/namespaces.go
```diff
@@ -129,12 +127,103 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+	if clientFactory == nil {
+		var err error
+		clientFactory, err = kubernetes.GetClientFactory()
+		if err != nil {
+			log.Errorf("Error getting client factory: %v", err.Error())
+			return nil, fmt.Errorf("Error getting client factory")
+		}
+	}
+
+	clusterNames := clientFactory.GetClusterNames()
```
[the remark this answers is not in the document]
josunect: That's right. Now we can use `in.userClients`

business/namespaces.go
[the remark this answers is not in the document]
josunect: Yes, change addressed. Also, code looks much more clear now.

business/namespaces.go
josunect: +1, I'll add a new test.

business/namespaces.go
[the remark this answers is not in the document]
josunect: +1, I'll add a new test.

business/mesh.go
```diff
@@ -334,7 +334,7 @@ func findKialiInNamespace(ctx context.Context, namespace string, clusterName str
 				services = kubernetes.FilterServicesByLabels(selector, tmpSvc)
 			}
 		} else {
-			services, getSvcErr = layer.k8s.GetServicesByLabels(kialiNs.Name, "app.kubernetes.io/part-of=kiali")
+			services, getSvcErr = layer.k8sClients[clusterName].GetServicesByLabels(kialiNs.Name, "app.kubernetes.io/part-of=kiali")
```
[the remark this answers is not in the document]
josunect: Not sure if we would need to add a k8sClients in MeshService. At the moment, it is created using the HomeCluster k8s. But, as findKialiInNamespace is passing the layer as parameter, layer needs the k8sClients - So that is why this change.

business/namespaces.go
nrfox: > So in case the error came from the local cluster it should be an error, it will keep the compatibility.

Returning an error to keep existing functionality for the home cluster (single cluster) sounds good to me. I don't think we want to make a bunch of exceptions for single cluster vs. multicluster but this seems like a good one to make to preserve the existing behavior for single cluster. We may need to make this configurable in the future but that can be added later.

> In case of MC, we assume the user always has to have permissions in the cluster, right?

This is true for single cluster but won't necessarily be true for multi cluster. The user could have no permissions for the "home cluster" but have access instead to the "remote cluster(s)".

It'd be good to document behavior like this for when multicluster is ready for use. It's probably too premature to have a multicluster section in the docs right now but I can create a separate MC issue for docs and make a note of this there.

business/layer.go
```diff
@@ -165,29 +156,29 @@ func SetWithBackends(cf kubernetes.ClientFactory, prom prometheus.ClientInterfac
...
 
-// NewWithBackends creates the business layer using the passed k8s and prom clients.
+// NewWithBackends creates the business layer using the passed k8sClients and prom clients.
 // Note that the client passed here should *not* be the Kiali ServiceAccount client.
 // It should be the user client based on the logged in user's token.
-func NewWithBackends(k8s kubernetes.ClientInterface, prom prometheus.ClientInterface, jaegerClient JaegerLoader) *Layer {
+func NewWithBackends(k8s map[string]kubernetes.ClientInterface, kialiSAClients map[string]kubernetes.ClientInterface, prom prometheus.ClientInterface, jaegerClient JaegerLoader) *Layer {
 	temporaryLayer := &Layer{}
-	temporaryLayer.App = AppService{prom: prom, k8s: k8s, businessLayer: temporaryLayer}
+	temporaryLayer.App = AppService{prom: prom, k8s: k8s[kubernetes.HomeClusterName], businessLayer: temporaryLayer}
```
jmazzitelli, line 164: maybe its just me not understanding this function, but... why is the home cluster hardcoded in all of these? All the clients will be for the home cluster...should there be a TODO here to update this later for multicluster (I would think we'd need to pass a cluster name to this function)?

business/namespaces.go
```diff
@@ -38,10 +39,10 @@ func IsAccessibleError(err error) bool {
 	return isAccessibleError
 }
 
-func NewNamespaceService(k8s kubernetes.ClientInterface) NamespaceService {
+func NewNamespaceService(k8s map[string]kubernetes.ClientInterface, kialiSAClients map[string]kubernetes.ClientInterface) NamespaceService {
 	var hasProjects bool
 
-	if k8s != nil && k8s.IsOpenShift() {
+	if k8s != nil && k8s[kubernetes.HomeClusterName].IsOpenShift() {
```
jmazzitelli, line 45: I think we want to use the SA here - we should not assume the user will have the permissions to determine if the cluster is openshift. We can assume the SA will have those permissions. (I actually don't remember what IsOpenShift queries, but I thought it looks at some internal resource; in any event, I don't see a reason why we should use the user client here - I think it should be the SA client).

business/namespaces.go
```diff
@@ -129,12 +129,98 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+			}(cluster)
+		}
+		wg.Wait()
+		close(resultsCh)
+	}()
+
+	// Combine namespace data
+	for result := range resultsCh {
+		if result.err != nil {
+			log.Errorf("Error fetching Namespaces per cluster %s: %s", result.cluster, result.err)
```
jmazzitelli, line 168: the user may not have permission to see the cluster right? In that case this should not be an error. At most it should be a warning (but even then I'm not sure - maybe this should be Info or even Debug?)

business/namespaces.go
```diff
@@ -129,12 +129,98 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+					}
+				}
+			}
+			resultns = append(resultns, namespace)
+		}
+	}
+
+	if kialiCache != nil {
+		// Saved per user token
+		kialiCache.SetNamespaces(in.userClients[kubernetes.HomeClusterName].GetToken(), resultns)
```
jmazzitelli, line 207: are the tokens common across all clusters? I thought maybe the tokens would be cluster-specific (in which case hardcoding this to the home cluster might not work).

business/namespaces.go
```diff
@@ -241,44 +327,19 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
-				}
-			}
-			result = append(result, namespace)
-		}
-	}
-
 	if kialiCache != nil {
-		kialiCache.SetNamespaces(in.k8s.GetToken(), result)
+		// Set namespace by user token
+		kialiCache.SetNamespaces(in.userClients[kubernetes.HomeClusterName].GetToken(), result)
```
jmazzitelli, line 338: see my previous comment - i'm not sure tokens are the same across all clusters. Hardcoding to the home cluster might not work.

kubernetes/client_factory.go
```diff
@@ -210,26 +240,53 @@ func (cf *clientFactory) newSAClient(clusterInfo *RemoteClusterInfo) (*K8SClient
 	}
 }
 
+// getClient returns a client for the specified token. Creating one if necessary.
+func (cf *clientFactory) GetSAClients() map[string]ClientInterface {
+	return cf.saClientEntries
```
jmazzitelli, line 245: This does not look safe. This is using the saClientEntries map without the mutex. Everywhere else in here "saClientEntries" access is protected around the mutex lock. I suspect this function will need to use the mutex also.

business/namespaces.go
leandroberetta: I believe tokens are the same because all clusters are using the same OIDC provider. One token for all. @nrfox am I right?.

jmazzitelli (commented):
I left some comments. I don't have any requested changes, though some things in here look weird - missing the mutex around saClientEntries is the main thing I am questioning.

business/namespaces.go
nrfox: This raises an interesting question on whether it is valid to have both openshift and non-openshift clusters in the same mesh? The same is probably true of gateway-api that some clusters may have the CRDs installed and some may not.

business/namespaces.go
nrfox: Given we're assuming a common OIDC provider, the user tokens will be the same, but the service account tokens will be unique for each cluster.

business/layer.go
josunect: You are right, we need a TODO here. This was to delimit this task. Not sure if other services will need all the k8s structure, because e.g. for Jaeger, there is going to be just the local service, even though, probably it will need access to all the k8s client for the clusters.

business/namespaces.go
josunect: Right. We can shown an error when it is from the home cluster. For the remote cluster, maybe we can use an info: It could be more important as this cannot be just from permissions issue, maybe a network issue, or other reason - But yes, there could be many messages.

business/namespaces.go
josunect: I was testing in OpenShift and it looks like it is able to get the information. Probably because it uses the Kiali SA when it is required. 
We can change it for kialiSAClients to make it more explicit, anyway.

jmazzitelli (changes requested):

jmazzitelli (approved):

jmazzitelli (commented):
I added an approval because all my suggested changes have been incorporated.

I'll let others chime in with their own suggestions - my approval is only related to my own suggested changes.

josunect:
> I added an approval because all my suggested changes have been incorporated.
> 
> I'll let others chime in with their own suggestions - my approval is only related to my own suggested changes.

Thanks for your review! I will have a look at the others comments before merge.

kubernetes/cache/namespaces.go
```diff
@@ -18,14 +18,25 @@ type (
...
 	}
+
 	c.tokenNamespaces[token] = namespaceCache{
-		created:       time.Now(),
-		namespaces:    namespaces,
-		nameNamespace: nameNamespace,
+		created:          time.Now(),
+		namespaces:       namespaces,
+		nameNamespace:    nameNamespace,
+		clusterNamespace: clusterNamespace,
```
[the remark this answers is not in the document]
josunect: At the moment just the clusterNamespace is used in GetNamespace. 
Depending on how we request the information it would be more efficient have one structure or the other, but, at the moment, we can just leave the one that is used - or is more efficient.

business/namespaces.go
josunect: I will resolve this comment. The initial approach was changed and each namespace will be unique by name/ cluster

business/namespaces.go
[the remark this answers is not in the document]
josunect: Test is added

business/namespaces.go
josunect: Now GetNamespace uses the cluster

business/layer.go
josunect: Added TODO

business/namespaces.go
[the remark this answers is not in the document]
josunect: Changes done.

business/namespaces.go
```diff
@@ -129,12 +129,100 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+		err     error
+	}
+	resultsCh := make(chan result)
+
+	go func() {
+		for _, cluster := range clusterNames {
+			wg.Add(1)
+			go func(c string) {
+				defer wg.Done()
+				list, error := in.GetNamespacesByCluster(c)
```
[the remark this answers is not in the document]
josunect: https://github.com/kiali/kiali/issues/5941

nrfox (approved):

nrfox:
Tested this manually in multiple auth modes on single and multicluster deployments and all worked as expected.

@matejnesuta unit tests were added along with this change but there's no integration/cypress suite for multi-cluster scenarios right now. For single cluster as @jshaughn mentioned, the existing suite should cover the changes here to ensure there's not a regression.

josunect:
Tested with OpenShift auth: 

Permissions looks as expected: 
image
image

PR #5852, opened 2023-02-16
Fix validation for K8s gateway hosts

### LINKED ISSUE #5802: K8s Gateway wizard hostname is valid in frontend but rejected in Kiali server
If you set a hostname value in K8s gateway wizard with pattern "string/string.string", the value is set as valid in frontend wizard:

image

But Kiali server returns error with following message:
{"error":"Gateway.gateway.networking.k8s.io \"test\" is invalid: spec.listeners.hostname: Invalid value: \"test/test.test\": spec.listeners.hostname in body should match '^(\\*\\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$'"}

image

The same validation should be both in frontend wizard and server backend.

ferhoyos:
** Describe the change **

Applies different hostname validations for Istio Gateways and K8s Gateways:
- Istio Gateway hosts have namespace/dnsName with namespace optional (https://istio.io/latest/docs/reference/config/networking/gateway/)
- K8s Gateway hosts have only dnsName (https://gateway-api.sigs.k8s.io/v1alpha2/references/spec/#gateway.networking.k8s.io/v1beta1.Hostname)

K8s gateway host validations obtained from K8s Gateway API library (shared_types.go):

// Hostname is the fully qualified domain name of a network host. This matches
// the RFC 1123 definition of a hostname with 2 notable exceptions:
//
// 1. IPs are not allowed.
// 2. A hostname may be prefixed with a wildcard label (`*.`). The wildcard
//    label must appear by itself as the first label.
//
// Hostname can be "precise" which is a domain name without the terminating
// dot of a network host (e.g. "foo.example.com") or "wildcard", which is a
// domain name prefixed with a single wildcard label (e.g. `*.example.com`).
//
// Note that as per RFC1035 and RFC1123, a *label* must consist of lower case
// alphanumeric characters or '-', and must start and end with an alphanumeric
// character. No other punctuation is allowed.
//
// +kubebuilder:validation:MinLength=1
// +kubebuilder:validation:MaxLength=253
// +kubebuilder:validation:Pattern=`^(\*\.)?[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
type Hostname string

** Issue reference **

Fixes #5802 

** Testing **

Checks that valid hostnames in Istio Gateways are no longer valid in K8s Gateways, like for example:
- \*
- test/test.test
- test/*
- TEST.test

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ TARGET_ARCHS ?= amd64 arm64 s390x ppc64le
 
 # Identifies the current build.
 # These will be embedded in the app and displayed when it starts.
-VERSION ?= v1.64.0-SNAPSHOT
+VERSION ?= v1.65.0-SNAPSHOT
 COMMIT_HASH ?= $(shell git rev-parse HEAD)
 
 # The path where the UI project has been git cloned. The UI should
--- a/frontend/cypress/README.md
+++ b/frontend/cypress/README.md
@@ -60,7 +60,7 @@ cypress/
 4) Failed tests could mean that:
     * Reuse of step definitions is not suitable or gets broken by different testcase or your modifications
     * We want to refactor broken code and if its heavily used, move it into a custom command file (cypress/support/commands.ts) - i.e. `cy.login()`, `cy.kiali_apply_config()` lives there
-5) Test case execution should be all green, you are ready to commit your test case. You might want verify whole regression run locally - so you did not introduce any braking changes in your PR
+5) Test case execution should be all green, you are ready to commit your test case. You might want verify whole regression run locally - so you did not introduce any breaking changes in your PR
 
 ## Performance Tests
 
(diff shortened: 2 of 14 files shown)


ferhoyos:
mmmm that means the istio validation regex in the API is not restricted enough for IP addresses and names without fqdn. Thanks @hhovsepy, I will double check this and the K8sRouteHost wizard too.

ferhoyos:
Same validation must be applied to K8sHTTPRouteHosts too (https://gateway-api.sigs.k8s.io/api-types/httproute/)

image

hhovsepy:
Code lgtm and works as expected. Waiting for cypress tests for final approve.

ferhoyos:
@hhovsepy Input validation has been added to Cypress test. Since it is the same validation I have only added to K8s Gateway creation test. If you think this validation need to be added to Service K8s Gateway wizard advanced settings too please let me know.

ferhoyos:
Closing PR due to bad resolution of master merge conflicts. Created new one with correct changes (#5863).

PR #5855, opened 2023-02-16
[hack] ca data might be specified directly (as with CRC)

jmazzitelli:
I noticed that the kubeconfig may have the CA data directly defined (as opposed to referring to a file that we then have to read in). This will happen if you install CRC/OpenShift, for example. If the certificate-authority-data is already in the kubeconfig, use it as-is. Otherwise, try to get the CA file and read it like before.


jshaughn (approved):
LOK

Issue #5864, opened 2023-02-21
Application crash when accessing service from K8s Gateway


ferhoyos:
After creating a K8s Gateway (e.g. the one created by Cypress test), the application crashes when the user accesses the service details associated to K8s Gateway.

![image](https://user-images.githubusercontent.com/122779323/220376400-914dfb36-1f87-4712-98b4-30cda0fb0e39.png)

![image](https://user-images.githubusercontent.com/122779323/220376497-30808da8-fcd7-4a2d-922e-56cca1a86bd3.png)

![image](https://user-images.githubusercontent.com/122779323/220376746-eae962d2-4322-4c08-9417-ad396532d40d.png)

I think it is related with new annotations feature added by @aljesusg. The error happens because server returns null for service annotations and the following code cannot map a null value throwing the error (WizardAnnotations.tsx):

```
convertAnnotationsToMap = () => {
    const m = new Map();
    Object.keys(this.props.annotations).map((value, index) => m.set(index, [value, this.props.annotations[value]]));
    return m;
  };  
```

aljesusg:
O.o the value should not be null, must be empty, this error must be fixed in the backend

PR #5865, opened 2023-02-21
Fix annotations if value is null

aljesusg:
Fix https://github.com/kiali/kiali/issues/5864

### DIFF
--- a/frontend/src/components/IstioWizards/ServiceWizardDropdown.tsx
+++ b/frontend/src/components/IstioWizards/ServiceWizardDropdown.tsx
@@ -24,7 +24,7 @@ import {
   WIZARD_EDIT_ANNOTATIONS
 } from './WizardActions';
 import { MessageType } from '../../types/MessageCenter';
-import WizardAnnotations from "./WizardAnnotations";
+import WizardAnnotations from './WizardAnnotations';
 import ServiceWizard from './ServiceWizard';
 import { canCreate, canUpdate, ResourcePermissions } from '../../types/Permissions';
 import ServiceWizardActionsDropdownGroup, { DELETE_TRAFFIC_ROUTING } from './ServiceWizardActionsDropdownGroup';
@@ -130,7 +130,7 @@ class ServiceWizardDropdownComponent extends React.Component<Props, State> {
         break;
       }
       case WIZARD_EDIT_ANNOTATIONS: {
-        this.setState({showAnnotationsWizard: true})
+        this.setState({ showAnnotationsWizard: true });
         break;
       }
       default:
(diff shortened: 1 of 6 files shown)


aljesusg:
> Now application works but server still returns null value on annotations field. Should not return empty [] when there are not annotations on the service?

should... but let's merge this one to fix the error in the front, then we can discuss about this because I expect more "return null values" in other sides

ferhoyos:
ok, in that case frontend code lgtm. Approved.

ferhoyos (approved):

ferhoyos:
@aljesusg Upss I do not have write access yet. You will need a maintainer approval to be able to merge the PR.

aljesusg:
> @aljesusg Upss I do not have write access yet. You will need a maintainer approval to be able to merge the PR.

step by step but at least you are reviewing ;)

hhovsepy:
It seems when user edits and removes all annotations, then it is not possible to add them anymore as the "Edit Annotations" action item is removed. Which causes inconsistency:
Screenshot from 2023-02-21 18-46-01

@aljesusg can you include also backend changes into this PR to have consistency?

frontend/src/components/IstioWizards/ServiceWizardDropdown.tsx
```diff
@@ -255,13 +256,15 @@ class ServiceWizardDropdownComponent extends React.Component<Props, State> {
...
           ? this.renderTooltip('tooltip_wizard_actions', TooltipPosition.top, toolTipMsgActions, dropdown)
           : dropdown}
-        <WizardAnnotations 
-          showAnotationsWizard={this.state.showAnnotationsWizard} 
-          onChange={(annotations) => this.onChangeAnnotations(annotations)} 
-          onClose={() => this.setState({showAnnotationsWizard: false})} 
-          annotations={this.props.annotations} 
-          canEdit={serverConfig.kialiFeatureFlags.istioAnnotationAction && !serverConfig.deployment.viewOnlyMode}
-        />
+        {this.props.annotations && (
```
hhovsepy, line 259: When there is not annotation, then it causes dead end.

aljesusg:
> It seems when user edits and removes all annotations, then it is not possible to add them anymore as the "Edit Annotations" action item is removed. Which causes inconsistency: Screenshot from 2023-02-21 18-46-01
> 
> @aljesusg can you include also backend changes into this PR to have consistency?

Yes...it makes sense...that is why it should return the backend and empty array let me make the change

hhovsepy:
@aljesusg when backend is fixed, can the frontend null checks be removed on annotations for Service and Workloads? So the Edit Annotations are back.

hhovsepy:
Here I will suggest this change on line https://github.com/kiali/kiali/blob/master/frontend/src/components/IstioWizards/WizardAnnotations.tsx#L47 because the original issue with page crashing is back.
`Object.keys(this.props.annotations || {}).map((value, index) => m.set(index, [value, this.props.annotations[value]]));`

hhovsepy (approved):
Verified now. The failing cypress tests are flakes.

ferhoyos (approved):
Verified from my end too.

PR #5877, opened 2023-02-28
Beautify render error message

### LINKED ISSUE #5862: Error page could be more than just a text message
When there is an error, just unstyled small text message is shown at the top:

image

A custom error component could be created with some warning icon or image, styled error text message and even a refresh button.

ferhoyos:
** Describe the change **

Convert simple render error message to a more visual one (adding error icon and return to last page button).

image

** Issue reference **

Closes #5862 

** Testing **

Just adding `throw new Error()` in componentDidMount() method of any rendered page (e.g., OverviewPage) to cause an intentional error and see the error message.


jshaughn:
p.s. @ferhoyos remember to set assignee and sprint.

ferhoyos:
sorry about the assignee, on the other hand I think only maintainers can set sprint :-(

PR #5883, opened 2023-03-01
Update the Namespace cache to support multicluster (2 approach)

josunect:
Ref. https://github.com/kiali/kiali/issues/5766

This is a second approach to reflect the structure of a namespace VS a combined Namespace. 
It keeps both because: 
- The combination of namespaces is saved in the cache.
- In case we need a namespace details by cluster we still keep the uniqueness for this. We can analyze if the data from one namespace in a specific cluster is incorrect, and point to that cluster.

CombinedNamespaces can return: 

- A slice of clusters 
- For each namespace label, an array of values 
- For each namespace annotation, an array of values

It was done in a separate branch as it includes many code changes.


josunect:
No combination of information in the API. At least for now.

PR #5885, opened 2023-03-02
Don't let the graph generation panic on unexpected/corrupt telemetry.

jshaughn:
Fixes https://github.com/kiali/kiali/issues/5890

Don't let the graph generation panic on unexpected/corrupt telemetry.  Instead, add error return values as needed, and use then to skip over the bad telem.  The change here is fairly straightforward but it touches a lot of files, many test files, because the call signatures change for a few commonly used functions in graph generation.


jmazzitelli (approved):
lots of changes, but most are just due to the new return value and pass-by-reference.

LGTM

Issue #5882, opened 2023-03-01
Istio Config doesn't show correct yaml for Sidecar OutboundTrafficPolicy.Mode


jmazzitelli:
For background, see these:
* https://github.com/istio/istio/issues/43657
* https://github.com/istio/istio/issues/43643

From one of the comments in that first issue:

> `outboundTrafficPolicy: {}`: REGISTRY_ONLY (due to proto)
> `outboundTrafficPolicy: null` (or more likely, not present at all): ALLOW_ANY, due to Istio specific code

The problem is that proto doesn't give us any way that I can see to determine if the value was explicitly specified but stripped from the object because it was the default value (that's the `{}`) OR if the value was simply not defined (that's the null). So... there is really no way to know how to display this to the user once proto has serialized/deserialized the object.

Here's a quick way to see the problem.

1. Create main.go:
```go
package main

import (
	"fmt"

	api_networking_v1beta1 "istio.io/api/networking/v1beta1"
	networking_v1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
)

func main() {
	sc1 := &networking_v1beta1.Sidecar{
		Spec: api_networking_v1beta1.Sidecar{
			WorkloadSelector: &api_networking_v1beta1.WorkloadSelector{
				Labels: map[string]string{
					"modeSpecified": "yes",
				},
			},
			OutboundTrafficPolicy: &api_networking_v1beta1.OutboundTrafficPolicy{
				Mode: api_networking_v1beta1.OutboundTrafficPolicy_REGISTRY_ONLY,
			},
		},
	}
	sc2 := &networking_v1beta1.Sidecar{
		Spec: api_networking_v1beta1.Sidecar{
			WorkloadSelector: &api_networking_v1beta1.WorkloadSelector{
				Labels: map[string]string{
					"modeSpecified": "no",
				},
			},
			OutboundTrafficPolicy: &api_networking_v1beta1.OutboundTrafficPolicy{
			},
		},
	}
	b, _ := sc1.Spec.MarshalJSON()
	fmt.Printf("sc1 json  --> %v\n", string(b))
	b, _ = sc1.Spec.GetOutboundTrafficPolicy().MarshalJSON()
	fmt.Printf("otp1 json --> %v\n", string(b))
	fmt.Printf("sc1.Mode  --> %v\n", sc1.Spec.GetOutboundTrafficPolicy().Mode)

	b, _ = sc2.Spec.MarshalJSON()
[...]

PR #5894, opened 2023-03-03
validation: flag sidecar outbound traffic policy as ambiguous

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5882

### DIFF
--- a/business/checkers/sidecars/outbound_traffic_policy_mode.go
+++ b/business/checkers/sidecars/outbound_traffic_policy_mode.go
@@ -1,0 +1,25 @@
+package sidecars
+
+import (
+	networking_v1beta1 "istio.io/client-go/pkg/apis/networking/v1beta1"
+
+	"github.com/kiali/kiali/models"
+)
+
+type OutboundTrafficPolicyModeChecker struct {
+	Sidecar *networking_v1beta1.Sidecar
+}
+
+func (c OutboundTrafficPolicyModeChecker) Check() ([]*models.IstioCheck, bool) {
+	checks := make([]*models.IstioCheck, 0)
+
+	// read this issue to find out why we do this check: https://github.com/kiali/kiali/issues/5882
+	if c.Sidecar.Spec.OutboundTrafficPolicy != nil {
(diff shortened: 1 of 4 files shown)


jmazzitelli:
This is what it currently looks like. I don't think I like it 100%. Notice that because we only have Unknown, Warning, and Error severity, I set this to unknown (because it isn't an error and not even really a warning. It is only "this is ambiguous, and you can't know what the value is just looking at it"). So this is more informational. But there is no "information" severity level. And because its "unknown" it seems like it doesn't flag the line with the path that is mentioned in the validation (which is spec.outboundTrafficPolicy). 

I would like an icon on line 35 here - how can I put a `(i)` icon here? This isn't "help text" because it only should show up when the value is ambiguous (and that is what this PR's validation checker now does).

unknown-sev-validation

FYI: the `models.IstioCheck` that is passed back by the new validation checker is this:

```yaml
Code: "KIA1007"
Message: "OutboundTrafficPolicy with empty mode value is ambiguous due to an Istio limitation. This may indicate ALLOW_ANY or REGISTRY_ONLY. Inspect the value using other means."
Severity: "unknown"
Path: "spec.outboundTrafficPolicy"
```

jmazzitelli:
Here is how you can test this (assume you have Istio installed and this PR built and Kiali deployed)

1. Install a Sidecar with the outbound traffic policy set to REGISTRY_ONLY:
```bash
cat <<EOM | kubectl apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: mazz
---
kind: Sidecar
metadata:
  name: default
  namespace: mazz
spec:
  egress:
  - hosts:
    - ./*
    - istio-system/*
  outboundTrafficPolicy:
    mode: REGISTRY_ONLY
EOM
```
2. Go to the Kiali UI and inspect that Sidecar resource in the Istio Config section of the UI.  You should see the info message this PR added.
3. Edit the sidecar resource you created in step 1 - change mode to ALLOW_ANY
4. Go back to the Kiali UI page and notice the message has went away.

jshaughn (approved):
Code looks fine.  KIA1007 matches the associated doc PR.  Working for me:

image

Issue #5764, opened 2023-01-27
Update each business layer to read from each configured backend kube cluster


nrfox:
Each business layer service needs to aggregate kube config across the different clusters kiali is connected to. Reads should come from the kiali cache ~and writes should be done using the user's token client.~ writes can be handled in #5765 or a separate issue.

~This issue can probably be broken down even further to update each business service individually.~

The following business layers should be modified:

- [ ] Services
- [ ] Workloads
- [ ] Configs
- [ ] Applications
- [ ] Graph
- [ ] Filter by Cluster name

matejnesuta:
Adding the test-add-coverage label just to mark this issue for a re-visit later

nrfox:
Moving this to #5919 which encompasses both frontend and backend changes for the list view. This issue only focused on the backend.

PR #5903, opened 2023-03-10
[WIP] Workload list show workloads from all clusters

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5764

Workloads are listed for namespaces across multiple clusters.
Screenshot from 2023-03-10 18-32-15

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -57,6 +57,7 @@ type WorkloadService struct {
 }
 
 type WorkloadCriteria struct {
+	Cluster               string
 	Namespace             string
 	WorkloadName          string
 	WorkloadType          string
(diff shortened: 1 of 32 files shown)


business/workloads.go
```diff
@@ -613,6 +635,8 @@ func fetchWorkloads(ctx context.Context, layer *Layer, namespace string, labelSe
 		return nil, err
 	}
 
+	kialiClient := clientFactory.GetSAClient(cluster)
```
leandroberetta, line 638: I think we need to use the user client here?, to just get what the user is allowed to see?.

leandroberetta (commented):
First iteration, so far so good, I just have one question below.

business/workloads.go
hhovsepy: @leandroberetta thanks, right, will use the user client once https://github.com/kiali/kiali/pull/5851 get merged, for now the main point of this PR is to rework workloads list and details to add "cluster" property.

hhovsepy:
Working on tests.

josunect (commented):
I know this is still a draft but I will add some comments (Maybe they were not addressed yet): 

image

- Should we show the cluster column if we are in single cluster? 
- I think we can discuss what name to show for the home cluster (_kiali_home)
- I think we can also add the cluster filter:
image


In the workload details, I have an issue with the graph: 

image

hhovsepy:
Facing some local git issues, closing this PR in a favor to https://github.com/kiali/kiali/pull/5944

PR #5914, opened 2023-03-15
[WIP] Kiali model to support multicluster

### LINKED ISSUE #5736: Multicluster - data models support multiple clusters
Kiali's data models assume a single cluster for workloads/services/apps. A [workload](https://github.com/kiali/kiali/blob/master/models/workload.go) for example has no way to track the cluster(s) it resides in. These need to be revisited for multicluster.

Part of this issue will be determining how or if these models need to change.

Part of: #5618

leandroberetta:
Resolves #5736 

This PR is draft and the purpose is to have some early reviews on the code.

Note: I added a cluster field to the IstioConfigList but I'm not sure this is needed, or at least I'm not sure if it will have a value different than the Home cluster. This doubt comes because when having just one control plane (as we were discussing on support just this scenario) for example in a primary-remote scenario, the Istio custom resources will be in the cluster where the control plane is installed (Istio does not create the CRDs on remote cluster so we won't expect to find any Istio configurations on them). Finally, in the case of multi primary control planes, we were discussing about having 1 Kiali per control plane, where the Istio configurations will belong to the Home clusters as well (each Kiali will be in a Home cluster where a control plane is also deployed).

### DIFF
--- a/models/app.go
+++ b/models/app.go
@@ -53,6 +53,9 @@ type WorkloadItem struct {
 }
 
 type App struct {
+	// Cluster where the app lives in
+	Cluster string
+
 	// Namespace where the app lives in
 	// required: true
 	// example: bookinfo
--- a/models/app.go
+++ b/models/app.go
@@ -31,6 +31,9 @@ type AppListItem struct {
 
 	// Health
 	Health AppHealth `json:"health,omitempty"`
+
+	// Cluster where the app lives in
+	Cluster string `json:"cluster"`
 }
 
 type WorkloadItem struct {
@@ -50,11 +53,14 @@ type WorkloadItem struct {
 	// List of service accounts involved in this application
 	// required: true
 	ServiceAccountNames []string `json:"serviceAccountNames"`
+
+	// Cluster where the app lives in
+	Cluster string `json:"cluster"`
 }
 
 type App struct {
 	// Cluster where the app lives in
-	Cluster string
+	Cluster string `json:"cluster"`
 
 	// Namespace where the app lives in
 	// required: true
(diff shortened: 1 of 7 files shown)


hhovsepy:
Hi @leandroberetta , it seems we are doing the same effort https://github.com/kiali/kiali/pull/5903

leandroberetta:
> Hi @leandroberetta , it seems we are doing the same effort #5903

Hello @hhovsepy, I guess it's not exactly the same effort but for sure in the models we will have some coincidences. Basically, the issue I was working on was related to the model changes, so this PR includes not just the cluster field for workloads but also for other resources. 

It seems that you're only touching the workload model, I can revert that file because yours already have the change. 

But for the rest of the model, I guess this PR stills being valid. If you don't mind I will add you as reviewer too.

nrfox:
@hhovsepy do you think the "list" models should also have a cluster field in them? https://github.com/kiali/kiali/blob/885b0400273e44df529f3b2dc5f6c5d576cfa979/models/service.go#L57-L61

The list models currently have namespace, validations, and each one of the service details for that namespace. Inside of service details we are adding a cluster field. So presumably the List models are cluster agnostic and the validations they show are for the entire mesh and not just one specific cluster. The same is true of the namespace field on the List models, it is every namespace with that name across all the cluster and not just one specific cluster. I'm not sure if adding a cluster field to the list models is better or not. If we do add a cluster field to the list models, you can then imagine the api returning a list of list models, one for each cluster.

hhovsepy:
@nrfox I think 'list' models do not need a Cluster field, List Item or Overview items field should have Cluster inside and this is enough.

PR #5924, opened 2023-03-19
Script to install the primary-remote mesh scenario

leandroberetta:
Resolves https://github.com/kiali/kiali/issues/5925.

The following is an script to install the primary-remote scenario using two minikube clusters within a single mesh.

The script follows [this](https://istio.io/latest/docs/setup/install/multicluster/primary-remote_multi-network/) intructions from the Istio docs. The multi network option was chosen because although both clusters' api servers are on the same subnet, the pods and services from the clusters are in a different network (one pod from east cluster cannot reach a service from west cluster directly, for that reason this script installs the east-west gatewatys for indirect communication)

It just installs Istio and configures the mesh. 

To install an split bookinfo version, the following command can be used:

`./hack/istio/multicluster/split-bookinfo.sh`

Note: When installing some Istio custom resources like Gateways or VirtualServices on the remote cluster, is expected to fail because the remote cluster does not have the CRDs installed, this is expected from the primary-remote scenario.

Same for Kiali, as this is an script (for now) for development, Kiali is intended to be deployed building it locally with the lastest changes.

There are two workarounds due to some problems with:

- Metallb's Minikube addon: Images from Docker seems to be gone and now they are replacing it with Quay in the next Minikube version (see https://github.com/kubernetes/minikube/issues/16053), this is also affecting the other script (install-everything.sh)
- Remote cluster's MutatingWebhookConfigurations failing due to missing caBundle to communicate with the primary cluster (see https://github.com/istio/istio/issues/42327)

After installing Kiali and the split bookinfo version, the graph will look like:

image

Note: There is a missing part in the graph, where reviews v2 and v3 calls the ratings from the same remote cluster, this might be something related to the fact that this communication happens enterirely on the remote cluster and should be  revisit.


leandroberetta:
> The base env script does provide for the `--bookinfo-enabled` flag (which by default is true)
> 
> Is there a reason why this script couldn't just do the following at the end (just like [install-everything.sh does](https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/install-everything.sh#L161-L162)):
> 
> ```shell
> # Install bookinfo across cluster if enabled
> source ${SCRIPT_DIR}/split-bookinfo.sh
> ```

It can, I will add it.

jmazzitelli:
ran the split-bookinfo script, Installed Kiali in east, and I see this - so this looks correct.
image

jmazzitelli:
I would say once you add the `source split-bookinfo.sh` line as mentioned earlier, this LGTM

jmazzitelli (approved):
I haven't run the latest changes, but everything looks good (bookinfo is added now). I see the remote cluster secrets and prom are also added, too.

jmazzitelli (changes requested):
With the rename of install-everything and the addition of the other scripts, can you update the README with the descriptions.

https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/README.adoc Might be helpful in the README to explain what `install-multi-primary` and `install-primary-remote` do differently compared to each other.

jmazzitelli (dismissed):

PR #5930, opened 2023-03-20
Use cache only for workload service

nrfox:
Part of #5599. Most of this change is updates to the unit tests that previously did not work with the cache enabled. This will be affected by #5851 and should probably be merged after and conflicts resolved.


hhovsepy (commented):
Did some regression testing, works for me.

nrfox:
> Just one comment, kialiCache is just passed to the service business service because is the only test that didn't work with the disabled cache?

No before it was using the global var `kialiCache`. Having it as a global var meant I had to set that global var everywhere but the only way to know which tests were using it was having it crash at _runtime_. This really obscured how large this change was. If the various business services did not use global vars but required all their dependencies to be passed in, we should know at _compile_ time if something was missing or not.

All of the tests that were updated relied on the cache being disabled and once the cache was required for the business service, each one of the tests now had to setup the cache.

josunect (approved):

PR #5940, opened 2023-03-22
Check if Istio API (CRDs) are defined to cache them

### LINKED ISSUE #5939: Modify the Kube Cache to allow whether to cache Istio types or not
This is related to multicluster. 

In the Primary-Remote configuration, on remote clusters, the Istio CRDs may not be present, in this case, the actual Kube Cache will fail to initialize because of this (in fact, this is happening right now with our install-primary-remote.sh script).

This issue aims to analyze the situation and dicuss the approach on doing this. 

One proposal is to check if the cluster has any CRD from Istio and then if there aren't present, create the Kube Cache with an empty list of Istio resources to cache:

`c := &kubeCache{cacheIstioTypes:  map[string]bool{},`

Epic: #5618

leandroberetta:
Resolves #5939.

Similarly to the GatewayAPI logic, this PR adds the same for Istio APIs. This will be necessary for the remote clusters where is not mandatory to have the CRDs installed.

The following is an example of a Primary-Remote setup with Kiali working ok:

image

### DIFF
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -1066,6 +1066,11 @@ func (c *kubeCache) GetVirtualServices(namespace, labelSelector string) ([]*netw
 	// but it won't prevent other routines from reading from the lister.
 	defer c.cacheLock.RUnlock()
 	c.cacheLock.RLock()
+	// Check if informers are started for this type
+	if c.getCacheLister(namespace).virtualServiceLister == nil {
+		// If disabled, return an empty list
+		return []*networking_v1beta1.VirtualService{}, nil
+	}
 	vs, err := c.getCacheLister(namespace).virtualServiceLister.VirtualServices(namespace).List(selector)
 	if err != nil {
 		return nil, err
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -365,70 +365,72 @@ func (c *kubeCache) createIstioInformers(namespace string) istio.SharedInformerF
 	sharedInformers := istio.NewSharedInformerFactoryWithOptions(c.client.Istio(), c.refreshDuration, opts...)
 	lister := c.getCacheLister(namespace)
 
-	if c.CheckIstioResource(kubernetes.AuthorizationPolicies) {
-		lister.authzLister = sharedInformers.Security().V1beta1().AuthorizationPolicies().Lister()
-		lister.cachesSynced = append(lister.cachesSynced, sharedInformers.Security().V1beta1().AuthorizationPolicies().Informer().HasSynced)
-		sharedInformers.Security().V1beta1().AuthorizationPolicies().Informer().AddEventHandler(c.registryRefreshHandler)
-	}
-	if c.CheckIstioResource(kubernetes.DestinationRules) {
-		lister.destinationRuleLister = sharedInformers.Networking().V1beta1().DestinationRules().Lister()
-		lister.cachesSynced = append(lister.cachesSynced, sharedInformers.Networking().V1beta1().DestinationRules().Informer().HasSynced)
-		sharedInformers.Networking().V1beta1().DestinationRules().Informer().AddEventHandler(c.registryRefreshHandler)
-	}
-	if c.CheckIstioResource(kubernetes.EnvoyFilters) {
-		lister.envoyFilterLister = sharedInformers.Networking().V1alpha3().EnvoyFilters().Lister()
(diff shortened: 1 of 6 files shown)


kubernetes/kubernetes.go
```diff
@@ -228,6 +228,24 @@ func (in *K8SClient) IsGatewayAPI() bool {
...
+	if in.isIstioAPI == nil {
+		isIstioAPI := false
+		_, err := in.k8s.Discovery().RESTClient().Get().AbsPath("/apis/networking.istio.io").Do(in.ctx).Raw()
+		if err == nil {
+			isIstioAPI = true
+		} else if !errors.IsNotFound(err) {
+			log.Warningf("Error checking Istio API configuration: %v", err)
+		}
+		in.isIstioAPI = &isIstioAPI
+	}
```
nrfox, line 245: This is how the other checks are done too so it's fine to keep it this way but I wonder if it'd be better to have finer grained checks per api group/resource rather than just checking a single group and assuming the rest are present.

kubernetes/cache/kube_cache.go
```diff
@@ -365,70 +365,72 @@ func (c *kubeCache) createIstioInformers(namespace string) istio.SharedInformerF
...
-	if c.CheckIstioResource(kubernetes.WorkloadEntries) {
-		lister.workloadEntryLister = sharedInformers.Networking().V1beta1().WorkloadEntries().Lister()
-		lister.cachesSynced = append(lister.cachesSynced, sharedInformers.Networking().V1beta1().WorkloadEntries().Informer().HasSynced)
-		sharedInformers.Networking().V1beta1().WorkloadEntries().Informer().AddEventHandler(c.registryRefreshHandler)
-	}
-	if c.CheckIstioResource(kubernetes.WorkloadGroups) {
-		lister.workloadGroupLister = sharedInformers.Networking().V1beta1().WorkloadGroups().Lister()
-		lister.cachesSynced = append(lister.cachesSynced, sharedInformers.Networking().V1beta1().WorkloadGroups().Informer().HasSynced)
-		sharedInformers.Networking().V1beta1().WorkloadGroups().Informer().AddEventHandler(c.registryRefreshHandler)
+	if c.client.IsIstioAPI() {
```
nrfox, line 368: Can you add a test for this in `kube_cache_test.go`?

nrfox:
Will the UI/API still let you attempt to create istio objects in a cluster with `IsIstioAPI = false`?

kubernetes/kubernetes.go
leandroberetta: I would continue doing this at least for now, maybe in a follow up issue we can change this to a more granular design (including the gateway api too). For multicluster I think this is enough. What do you think?.

leandroberetta:
> Will the UI/API still let you attempt to create istio objects in a cluster with `IsIstioAPI = false`?

I don't think we should let that, checky my other comment. I think we need to allow the creation of this resources where IsIstioAPI is true.

josunect (commented):
Tested and works as expected: 
image

Kiali was not starting on the environment without these changes. 
I will take a look at the code.

kubernetes/cache/kube_cache_test.go
```diff
@@ -19,11 +19,15 @@ import (
 	"github.com/kiali/kiali/kubernetes/kubetest"
 )
 
-func newTestingKubeCache(t *testing.T, cfg *config.Config, objects ...runtime.Object) *kubeCache {
+const IstioAPIEnabled = true
+
+func newTestingKubeCache(t *testing.T, cfg *config.Config, istioAPIEnabled bool, objects ...runtime.Object) *kubeCache {
```
nrfox, line 24: It probably would've been better when I wrote this to accept a `k8s kubernetes.ClientInterface` here rather than the `objects` so that you could set some options for the fake client. You don't need to refactor but can you copy the setup code into that test that needs to edit the client rather than adding more opts to this setup func? Otherwise these things quickly grow without bound and then it's hard to use and isn't much of a "helper" method.

kubernetes/cache/kube_cache.go
```diff
@@ -1066,6 +1066,11 @@ func (c *kubeCache) GetVirtualServices(namespace, labelSelector string) ([]*netw
 	// but it won't prevent other routines from reading from the lister.
 	defer c.cacheLock.RUnlock()
 	c.cacheLock.RLock()
+	// Check if informers are started for this type
+	if c.getCacheLister(namespace).virtualServiceLister == nil {
+		// If disabled, return an empty list
+		return []*networking_v1beta1.VirtualService{}, nil
+	}
```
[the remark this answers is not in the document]
leandroberetta: The error was the other option around. Yeah, I agree on changing this.

PR #5942, opened 2023-03-24
Add retries to downloading istio

nrfox:
Some transient errors in CI seem to be caused by failure to download Istio. Adding retries to try and alleviate the issue.

See: https://github.com/kiali/kiali/actions/runs/4504544840/jobs/7929276953 for an example of a failed CI job due to the failure to download Istio.


nrfox:
@jmazzitelli what do you think about this change? I've seen test flakes related to the download script for awhile now and hopefully this will make it better.

jmazzitelli (approved):
muy excelente

PR #5944, opened 2023-03-27
Workload list show workloads from all clusters

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5959

Workloads are listed for namespaces across multiple clusters.
Screenshot from 2023-03-29 08-41-54

[update]
Workload details page loads the overview info data from multiple clusters. Graph and other tabs are not loading any info for now from remote clusters.
In single cluster mode it works are before, the 'Cluster' column is not shown.


hhovsepy:
Comment from @josunect 
I know this is still a draft but I will add some comments (Maybe they were not addressed yet): 

image

- Should we show the cluster column if we are in single cluster? 
- I think we can discuss what name to show for the home cluster (_kiali_home)
- I think we can also add the cluster filter:
image


In the workload details, I have an issue with the graph: 

image

josunect:
Just a note here: I've found on ACM demo that the Cluster is identified as a badge like this: 

Screenshot from 2023-03-27 13-43-21

Probably it would be a good idea to follow the same format?

jshaughn:
> Just a note here: I've found on ACM demo that the Cluster is identified as a badge like this:

That's different than what we have defined in PfBadges for Cluster.  I'm not exactly sure where we got our current definition, may have to ask Andrew to confirm exactly what we should use.

business/workloads.go
```diff
@@ -613,6 +635,12 @@ func fetchWorkloads(ctx context.Context, layer *Layer, namespace string, labelSe
 		return nil, err
 	}
 
+	if cluster == "" {
+		cluster = kubernetes.HomeClusterName
+	}
```
nrfox, line 640: I don't think you need to have a special case for the empty string `""`, you can just pass the `kubernetes.HomeClusterName` where it hasn't been updated for multi-cluster yet.

hhovsepy:
@nrfox thanks for the review, the hiding single cluster column is WIP now.

business/workloads.go
hhovsepy: As the 'cluster' is a param coming from REST API, it can be empty or missing, for instance in called from E2E tests or any other REST Client. So adding default value here secures the backend itself.
Do you mean to add that default "HomeClusterName" value in UI calls?

hhovsepy:
@josunect the issue with graph is resolved.
@nrfox the 'Cluster' column is not shown for single cluster installation.

hhovsepy:
ready for review

business/workloads.go
```diff
@@ -326,7 +328,13 @@ func (in *WorkloadService) GetWorkload(ctx context.Context, criteria WorkloadCri
 		return nil, err
 	}
 
-	workload, err2 := fetchWorkload(ctx, in.businessLayer, criteria)
+	cluster := criteria.Cluster
+	if cluster == "" {
```
josunect, line 332: If we don't send any parameter at all (http://localhost:3000/console/namespaces/bookinfo/workloads/details-v1?duration=60&refresh=0), I get a nullPointer. It looks like this is converted to "null": 

image

josunect (commented):
Nice work, Hayk! 

Maybe we can add a new test to validate that there's no cluster column? As the frontend tests are done for single cluster at the moment. Or maybe it can be addressed in a separate issue.

business/workloads.go
hhovsepy: thanks for the catch Josune, I really missed that

hhovsepy:
@josunect thank you for the review, have approached you comments, also in workloads test there was a check of all columns.

josunect (approved):
LGTM, 

The cluster filter is missing but it will be addressed in a separate issue to keep consistency with all the list views.

hhovsepy:
Cypress tests are not starting: 'Error: Can't walk dependency graph: Cannot find module 'utils/ResponseFlags' from '/home/runner/work/kiali/kiali/frontend/src/components/Metrics/Helper.ts''

frontend/cypress/support/e2e.ts
nrfox, line 1: Is it necessary to add these cypress files? Is the the cause of the frontend CI failures?

frontend/src/types/Common.ts
```diff
@@ -8,6 +8,9 @@ export type RawDate = string;
 
 export type KioskMode = string;
 
+// cluster name to denote the cluster where Kiali is deployed
+export const HomeClusterName = '_kiali_home';
+
```
nrfox, line 13: This is kind of confusing to display to users since their actual cluster name probably won't be `_kiali_home`. This `_kiali_home` var wasn't really meant to be an external facing detail. We just needed something to put in the clients map that wouldn't overlap with another valid cluster name. On the backend we should probably get the "real" cluster name from something like the istio configmap but regardless of how the backend populates this value the cluster names should come from the backend. Why is it necessary to have this in the frontend as well?

hhovsepy: if we have this value in backend (not hardcoded), then that can be reused in frontend, but in backend it is hardcoded as well.

frontend/cypress/support/e2e.ts
hhovsepy: agree, these were odd due to a mistake

frontend/src/types/Common.ts
nrfox: Can the frontend optionally send cluster as a query param to the backend and when there's no cluster query param, the backend can assume that you want the home cluster? This way single cluster stays the same and we don't need to copy this hard coded value to the frontend that will probably be going away soon. If you are on a list page then the backend should populate the cluster names already and if you are on a details page then (I think) the url will contain the cluster name?

hhovsepy:
the failed cypress test is a K8s GW tests flake not related to this PR

frontend/src/types/Common.ts
hhovsepy: the frontend now has empty default cluster name in params, and the backend in this case takes from the default value it has

nrfox (approved):

andrew-ronaldson:
Looks like ACM took the badge out of the latest designs. We can use the blue C badge for now. I'd like to do an audit of the resource badges for future reference and document it as a convention with OCP.

Issue #5945, opened 2023-03-27
[make]Error to create operator pull secret


ferhoyos:
When I try to create operator with `make operator-create` command on Openshift CRC cluster, I receive following error at the end of the process:

```
error: failed to create secret Secret "kiali-operator-pull-secret" is invalid: data[.dockerconfigjson]: Invalid value: "<secret contents redacted>": invalid character 'S' looking for beginning of value
```

This is the version of my oc client:
```
Client Version: 4.12.4
Kustomize Version: v4.5.7
Server Version: 4.12.0
```

jmazzitelli:
This needs to be fixed in the kiali repo and openshift-servicemesh-plugin repo:
* https://github.com/kiali/kiali/blob/v1.65.0/make/Makefile.operator.mk#L18
* https://github.com/kiali/openshift-servicemesh-plugin/blob/20453c1f7e81bc6e3133f39298497a24d6cb6f9b/make/Makefile.cluster.mk#L33

PR #5946, opened 2023-03-27
[make] the --to option no longer writes to stdout so we need to refactor this.

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5945


ferhoyos (approved):
Tested and code LGTM.

Issue #5950, opened 2023-03-28
[make] run-operator should enable the profiler


jmazzitelli:
When a dev is testing the operator via the make target `run-operator` the ansible profiler should be enabled so you can quickly see the effects of any changes being made.

This should be done in both the main repo and the plugin repo.

the change is simply adding these two env vars when running `ansible-operator`

```
ANSIBLE_CALLBACK_WHITELIST="profile_tasks" \
ANSIBLE_CALLBACKS_ENABLED="profile_tasks" \
```

We want to put both in, even though the current version of the operator we are using doesn't support ansible that uses the ENABLED one. This is because when we do upgrade to an ansible-operator version that upgrades ansible, that WHITELIST one will be deprecated/obsolete and we'll need to use the ENABLED this. So having both will allow us to work with both versions of the operator. For more details on this, see https://github.com/kiali/kiali/issues/4338

These env vars should be added to these two places:

* https://github.com/kiali/kiali/blob/v1.65.0/make/Makefile.operator.mk#L260
* https://github.com/kiali/openshift-servicemesh-plugin/blob/v0.3.0/make/Makefile.operator.mk#L140

PR #5951, opened 2023-03-28
[make] turn on profiler when running the operator via run-operator

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/5950

When you run `make run-operator` and you see the operator finish reconciling, this kind of report should now appear in the output logs:

```
Tuesday 28 March 2023  08:23:53 -0400 (0:00:00.941)       0:00:00.990 ********* 
=============================================================================== 
Get all Kiali CRs ------------------------------------------------------- 0.94s
[.... and a whole bunch of other tasks and their times will be here...]
/home/jmazzite/source/kiali/kiali/operator/playbooks/kiali-new-namespace-detected.yml:18 
```


ferhoyos (approved):
Tested with OSSMC. It is the same change so no need to test with kiali operator too.

jshaughn (approved):
LOK

Issue #5921, opened 2023-03-16
Multicluster view for overview page


josunect:
- Define how to visualize the Overview page in multicluster: 
  - Show a single card for the same namespace name in each cluster
  - Add a Cluster filter
  - Include the Multicluster badge in the card? (If a card shows a combined namespace view, they could be many badges, if a namespace is in many clusters) 
  - Remove the CP card (Move to the Mesh page #5913)

Example with independent cards: 
![Overview](https://user-images.githubusercontent.com/49480155/226981821-63b9a2b0-aaa3-4ebf-9bc2-6d688d371a90.jpg)

Ref. #5620

josunect:
@andrew-ronaldson we would like to add the cluster name in the namespace cards in the Overview page. In multi cluster, it is possible to have the same namespace name, from different clusters, and each one will be represented in a different card. 
This proposal is adding the cluster name at the top of each card, with a similar style to the cluster column in the List views: https://github.com/kiali/kiali/issues/5919

It would be great if you can give us some feedback? Thank you!

PR #5952, opened 2023-03-28
How Kiali Identifies Ambient

josunect:
This is a proposal to define how Kiali identifies Ambient and its components for Istio Alpha (1.18). The behavior won't change if Istio is not installed with Ambient profile. 

- Added a Control Plane Ambient label:
image

- Added a Data Plane Ambient label if the namespace is labeled with Ambient annotations:
image

-Replace the "Sidecar missing" message if the workload/app/service doesn't have an Ambient OR Sidecar annotation: 
image
  
- If the workload/app/service has an Ambient annotation, it will show a ztunnel label in the details view. 

image

- If the workload/app/service has a waypoint proxy, it will show a waypoint label in the details view. 

image

- show the waypoint in the services/workloads list views and indicate this is part of the infrastructure: 
image

### Internal details

| Description  | Identifier |
| ------------- | ------------- |
| Ambient Enabled: use ambient profile | checks in the Istio config map `cni_network_config` for ambient_enabled = true  |
| Ambient namespace  | labeled as `istio.io/dataplane-mode=ambient` |
| Workload  |  the workload has the annotation `ambient.istio.io/redirection=enabled` |
| Identify a waypoint proxy  | labeled as `gateway.istio.io/managed=istio.io-mesh-controller`  |
| waypoint proxy per namespace | labeled as `istio.io/gateway-name=namespace`  |
| waypoint proxy per service account | Annotated as `istio.io/for-service-account`  |

### To be defined

- TODO: Show error in the labels if there is a ztunnel and a sidecar at the same time.
- Another proposal to show the waypoint proxy, as it is [Envoy-based](https://github.com/kiali/kiali/pull/5952), could be shown in a new tab, like the Envoy tab (A new Waypoint tab). e.x.
image

- Details view: Show sidecar badge to indicate that the workload has a sidecar? 

Ref. https://github.com/kiali/kiali/pull/5541 (Reviewed and integrated with the latest istio changes/Kiali code). 

Fixes https://github.com/kiali/kiali/issues/5535


josunect:
Some ideas from the demo: 

- Use the Ambient label in the namespace cards
- Detect errors if there is a sidecar and a ztunnel in the same namespace
- In the waypoint proxy view: Show the workloads attached to it (The service account)

josunect:
Short demo: 

[Ambient.webm](https://user-images.githubusercontent.com/49480155/229852598-9585a3e3-1f94-411e-9b15-6c81bb409441.webm)

jmazzitelli:
FYI: I was able to install minikube and run through the instructions here https://preliminary.istio.io/latest/docs/ops/ambient/getting-started/ and got it to work. Note that my GW got an error trying to externally expose the service (error: `Assigned to service(s) bookinfo-gateway-istio.istio-system.svc.cluster.local:80, but failed to assign to all requested addresses: address pending for hostname "bookinfo-gateway-istio.istio-system.svc.cluster.local"`) but I ignored this and kept going. Was able to install everything and get things to work (including the demos).

jmazzitelli:
From the UI perspective, I do see what is expected regarding the labels. The namespace where ambient is enabled is showed appropriately:

image
image

---
And where there is no ambient, I do not see the label:
image

---

I see the ztunnel labels:

image

image

---

So this all looks good. I'll look at the code now and do a review.

business/apps.go
```diff
@@ -219,6 +224,7 @@ func (in *AppService) GetAppDetails(ctx context.Context, criteria AppCriteria) (
 	(*appInstance).Workloads = make([]models.WorkloadItem, len(appDetails.Workloads))
 	for i, wkd := range appDetails.Workloads {
 		(*appInstance).Workloads[i] = models.WorkloadItem{WorkloadName: wkd.Name, IstioSidecar: wkd.IstioSidecar, Labels: wkd.Labels, ServiceAccountNames: wkd.Pods.ServiceAccounts()}
+		(*appInstance).Workloads[i] = models.WorkloadItem{WorkloadName: wkd.Name, IstioSidecar: wkd.IstioSidecar, IstioAmbient: wkd.IstioAmbient, ServiceAccountNames: wkd.Pods.ServiceAccounts()}
```
jmazzitelli, line 227: I'm confused here. Is this line duplicated, with the additional istioAmbient in the second line? Shouldn't line 226 be deleted?

business/istio_config.go
```diff
@@ -1123,6 +1124,24 @@ func (in *IstioConfigService) IsGatewayAPI() bool {
 	return in.k8s.IsGatewayAPI()
 }
 
+func (in *IstioConfigService) IsAmbientProfile() bool {
```
jmazzitelli, line 1127: The name of this function infers that it will return true if Istio was installed with the ambient *profile* - but that's technically not what we care about, right? I mean, in the future there could be other profiles that also enable Ambient (not just the `ambient` profile). I would call this something like `IsAmbientEnabled` or `IsAmbientSupported`. This really isn't necessarily related to the profile - it is related to the fact that Ambient is enabled in the cluster.

business/workloads.go
```diff
@@ -1678,13 +1679,42 @@ func fetchWorkload(ctx context.Context, layer *Layer, criteria WorkloadCriteria)
...
 		if cnFound {
 			return &w, nil
 		}
 	}
 	return wl, kubernetes.NewNotFound(criteria.WorkloadName, "Kiali", "Workload")
 }
 
+// Return the list of waypoint proxy binded to a workload service account
+func listWaypoint(ctx context.Context, layer *Layer, namespace string, w models.Workload) []string {
+	// Get List of waypoint proxies from the namespace and compare the
```
jmazzitelli, line 1696: `...and compare the` .... what does it compare? :) comment is incomplete.

frontend/src/components/VirtualList/Config.ts
```diff
@@ -31,7 +31,11 @@ export function hasHealth(r: RenderResource): r is SortResource {
 }
 
 export const hasMissingSidecar = (r: SortResource): boolean => {
-  return !isIstioNamespace(r.namespace) && !r.istioSidecar && !isGateway(r.labels);
+  return !isIstioNamespace(r.namespace) && !r.istioSidecar && !isGateway(r.labels) && !isWaypoint(r.labels);
+};
+
+export const hasMissingAmbient = (r: SortResource): boolean => {
```
jmazzitelli, line 37: what does "missing ambient" mean?

frontend/src/config/ServerConfig.ts
```diff
@@ -51,6 +51,7 @@ const computeValidDurations = (cfg: ComputedServerConfig) => {
 // than may not be providedby/set on the server.
 const defaultServerConfig: ComputedServerConfig = {
   accessibleNamespaces: [],
+  ambientProfile: false,
```
jmazzitelli, line 54: I think this is another one where "profile" isn't an accurate term to be used. Ambient could be installed in other istio profiles... I would prefer something like "ambientEnabled" or "ambientSupported".

frontend/src/types/ServerConfig.ts
```diff
@@ -111,6 +111,7 @@ export interface ToleranceConfig {
 
 export interface ServerConfig {
   accessibleNamespaces: Array<string>;
+  ambientProfile: boolean;
```
jmazzitelli, line 114: another one of these "profile" names that I think needs to change -- "enabled" or "supported" are more preferable to me.

handlers/config.go
```diff
@@ -51,6 +51,7 @@ type DeploymentConfig struct {
 type PublicConfig struct {
 	AccessibleNamespaces []string                    `json:"accesibleNamespaces,omitempty"`
 	AuthStrategy         string                      `json:"authStrategy,omitempty"`
+	AmbientProfile       bool                        `json:"ambientProfile,omitempty"`
```
jmazzitelli, line 54: another "profile" (see my earlier comments about this word and my alternatives).

models/app.go
```diff
@@ -23,6 +23,11 @@ type AppListItem struct {
 	// example: true
 	IstioSidecar bool `json:"istioSidecar"`
 
+	// Define if belongs to a namespace labeled as ambient
```
jmazzitelli, line 26: I see in another place where "IstioAmbient" is defined as the pod having an ambient annotation (see below from pod.go in this file) but this comment says this flag is true if the namespace is labeled (no mention of pod annotations). Please double check and just make sure this comment is correct and this really only involves namespace labels. This may be correct already - it could be different semantics between app and workload; I just want to make sure.

jmazzitelli (changes requested):
nothing major, but some suggestions and I had some questions, too.

business/apps.go
josunect: You are right! Line 226 is deleted now.

business/istio_config.go
josunect: I have renamed it for `IsAmbientEnabled`, it is more clear, and you are right, it doesn't need to reflect how it is set, just if the feature is enabled or not. But I think this is one of the things that could change in the future, I mean, maybe this check won't be needed as it could be the default working mode in the mesh. Let's see how it evolves, but I think it should be revisited in future versions.

business/workloads.go
josunect: That's right, thanks for pointing at this! I have reworked that function (And completed the comment). I think that would be more useful to check, for a waypoint proxy, the list of applications with the same service account that the Waypoint proxy is attached to. So, we can show a bit more of information in the Waypoint Details view: 

image

frontend/src/components/VirtualList/Config.ts
josunect: The first idea was to have a similar meaning that `missingSidecar`. But the implementation is not the same, so probably can be more meaningful with something like `noAmbientLabels`

frontend/src/config/ServerConfig.ts
josunect: Changed for AmbientEnabled

models/app.go
josunect: Yes, you are right. The check is done by looking at the pod and the ambient annotation, so I've updated the comment.

frontend/src/types/ServerConfig.ts
josunect: This was also renamed to "enabled"

jmazzitelli (dismissed):

leandroberetta:
Hi @josunect nice work!. One thought/suggestion, what do you think on adding a badge for the namespaces that has waypoint proxies installed?.

Probably another badge (another color?) similar to the "Ambient" one, but just saying "Waypoint". I think it will be useful at a glance in the overview to see what namespaces have a Waypoint.

models/pod.go
```diff
@@ -187,6 +196,28 @@ func (pod Pod) HasIstioSidecar() bool {
...
+	}
+	return false
+}
+
+// HasAmbient returns true if the pod is labeled as ambient-type
+func (pod *Pod) HasAmbient() bool {
+	return pod.Annotations[ambientAnnotation] == "enabled"
+}
+
+// IsWaypoint returns true if the pod is labeled as ambient-type
```
Xunzhuo, line 216: Comment here is the same as line 211, a bit confusing.

business/istio_config.go
```diff
@@ -1129,6 +1130,24 @@ func (in *IstioConfigService) IsGatewayAPI() bool {
...
+	var cniNetwork map[string]any
+	istioConfigMap, err := in.k8s.GetConfigMap(config.Get().IstioNamespace, "istio-cni-config")
+	if err != nil {
+		log.Errorf("Error getting istio-cni-config configmap: %s ", err.Error())
+	} else {
+		err = yaml.Unmarshal([]byte(istioConfigMap.Data["cni_network_config"]), &cniNetwork)
+		if err != nil {
+			log.Errorf("Error reading istio-cni-config configmap: %s ", err.Error())
+		}
+		if cniNetwork["ambient_enabled"] == true {
```
Xunzhuo, line 1144: Should we have type check here, since I believe cniNetwork["ambient_enabled"] is a `any`?

Like:

```
enabledAmbient, ok := cniNetwork["ambient_enabled"].(bool)

if ok && enabledAmbient {
    return true
}

```

Xunzhuo (commented):
Thanks for working on this @josunect, I have a few comments.

models/pod.go
josunect: That's right, @Xunzhuo , updated  comment

business/istio_config.go
josunect: Thanks for the catch!

josunect:
> Thanks for working on this @josunect, I have a few comments.

Thanks for the review, @Xunzhuo , the changes have been addressed.

josunect:
> Hi @josunect nice work!. One thought/suggestion, what do you think on adding a badge for the namespaces that has waypoint proxies installed?.
> 
> Probably another badge (another color?) similar to the "Ambient" one, but just saying "Waypoint". I think it will be useful at a glance in the overview to see what namespaces have a Waypoint.

Thanks for the suggestion, @leandroberetta ! Change has been done: 

image

business/istio_config.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"sync"
 
+	"gopkg.in/yaml.v2"
```
Xunzhuo, line 11: I just see many places we use different yaml packages to deal with yamls.. 😄

josunect: This package is also used in https://github.com/kiali/kiali/blob/master/business/mesh.go#L11 or https://github.com/kiali/kiali/blob/master/handlers/config.go#L9, not sure what other packages are used but, I was thinking that maybe that can be addressed in a separate issue?

jmazzitelli: AFAIK, we almost exclusively use this gopkg.in... I don't know where we use anything else. However, I can say we only recently are about to add the k8s yaml utility in one place, but that is only because it we can't use this gopkg.in yaml package -- only the k8s yaml was able to get the yaml to demarshal into the auto-generated protobuf code (that was done recently [here](https://github.com/kiali/kiali/pull/6027/files#diff-445738cf30752e18544dc219fe6aaeb2f7b10bc9be22a38a2ed5089783f05dc3R681)).

Xunzhuo: That is right, I saw this [change](https://github.com/kiali/kiali/pull/6027/files#diff-445738cf30752e18544dc219fe6aaeb2f7b10bc9be22a38a2ed5089783f05dc3R681) recently, so I point this out, no blocker on this.

josunect:
The last commit was to change the blue labels for Ambient: 

image

Ref. https://github.com/kiali/kiali/issues/5921#issuecomment-1514850443

mdfaizsiddiqui:
Hey @josunect 

I am not using Ambient mesh or istio-cni but I constantly see this error in kiali pod -

```
Error getting istio-cni-config configmap: configmaps "istio-cni-config" not found
```

This looks to be introduced by this PR, do you mind having a look at this?

Let me know if you want anything from me.

Thanks

jmazzitelli:
> Let me know if you want anything from me.

Please submit a github issue with all the relevant details. See: https://kiali.io/docs/faq/general/#kiali-does-not-work---what-do-i-do

mdfaizsiddiqui:
> > Let me know if you want anything from me.
> 
> Please submit a github issue with all the relevant details. See: https://kiali.io/docs/faq/general/#kiali-does-not-work---what-do-i-do

Here you go - https://github.com/kiali/kiali/issues/6111

PR #5969, opened 2023-03-30
Update demo apps hack scripts to support OSSM cluster wide mode

FilipB:
OSSM cluster wide mode has a different approach for member namespace management. This change is updating demo apps hack scripts to follow that.
Also the prepare_maistra functions has been moved to external script.


jmazzitelli (approved):
LGTM

Issue #5961, opened 2023-03-29
Multi-cluster list view - Applications


nrfox:
Applications list page pulls applications from all configured clusters.

josunect:
Probably the concept of application is a bit different than a workload or a service. As described in [apps.go](https://github.com/kiali/kiali/blob/master/business/apps.go):

> AppService deals with fetching Workloads group by "app" label, which will be identified as an "application"

For what we can see now, probably application doesn't need to be tied to a particular cluster: 

![image](https://user-images.githubusercontent.com/49480155/228812138-485fbc49-0afb-416d-a1fe-a6343e4dde5d.png)

An application can be the logical view of a subset of workloads from different clusters.
In that case, probably we would want to remove the cluster column from that view, and include the cluster name in the application details. 

Does it makes more sense, or do we want to keep splitting also this view by cluster?

jshaughn:
Hmm, it is an interesting question.
  In the graph the app and service nodes are still delimited by cluster.  Reviews-East is distinct from Reviews-West.  I think probably we should maintain that distinction in the list and detail views, but I'm not 100% sure.

 The thing is, for a detail page mini-graph you would always get a graph showing the app on all clusters (unless we make a change)
.
 My general guidance is that a single app or service that is deployed on different clusters may behave differently on a specific cluster, and a user may want to see it specifically for that cluster.   Meaning, the workloads for that App, for example, would be limited to the specific cluster.

PR #5990, opened 2023-04-04
Applications list for Multicluster

josunect:
Fixes https://github.com/kiali/kiali/issues/5961

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -161,7 +161,6 @@ func (in *AppService) GetAppList(ctx context.Context, criteria AppCriteria) (mod
 		for keyApp, valueApp := range clusterApps {
 			appItem := &models.AppListItem{
 				Name:         keyApp,
-				Cluster:      valueApp.cluster,
 				IstioSidecar: true,
 				Health:       models.EmptyAppHealth(),
 			}
@@ -212,6 +211,7 @@ func (in *AppService) GetAppList(ctx context.Context, criteria AppCriteria) (mod
 					log.Errorf("Error fetching Health in namespace %s for app %s: %s", criteria.Namespace, appItem.Name, err)
 				}
 			}
+			appItem.Cluster = valueApp.cluster
 			(*appList).Apps = append((*appList).Apps, *appItem)
 		}
 	}
--- a/business/apps.go
+++ b/business/apps.go
@@ -20,12 +20,13 @@ import (
 // AppService deals with fetching Workloads group by "app" label, which will be identified as an "application"
 type AppService struct {
 	prom          prometheus.ClientInterface
-	k8s           kubernetes.ClientInterface
+	userClients   map[string]kubernetes.ClientInterface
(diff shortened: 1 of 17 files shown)


hhovsepy:
It seems to be an existing issue in UI, when selecting all Namespaces, it requests workloads and applications twice, so the data is duplicated.
Screenshot from 2023-04-05 20-20-46

hhovsepy:
+1 in listing separate app per cluster for the duplicate namespace

business/apps.go
```diff
@@ -109,78 +134,86 @@ func (in *AppService) GetAppList(ctx context.Context, criteria AppCriteria) (mod
...
+				if appItem.IstioSidecar = w.IstioSidecar; !appItem.IstioSidecar {
+					break
+				}
 			}
-		}
-		if criteria.IncludeHealth {
-			appItem.Health, err = in.businessLayer.Health.GetAppHealth(ctx, criteria.Namespace, appItem.Name, criteria.RateInterval, criteria.QueryTime, valueApp)
-			if err != nil {
-				log.Errorf("Error fetching Health in namespace %s for app %s: %s", criteria.Namespace, appItem.Name, err)
+			if criteria.IncludeHealth {
```
leandroberetta, line 208: I was thinking if there is any work related to mc in the health computation, I noticed that there some queries to get the traffic rate that is used in the health, but I don't think we need to include the cluster value in the queries, but I'm not sure. I think that we need to check any inbound and outbound connection to the specific application (from the application's cluster or other clusters too).

leandroberetta (commented):
Code looks good to me. I have just a question related to the health.

leandroberetta (dismissed):
Code looks good and also verified that works fine. Regarding my health question, health for mc can be done in a follow up PR.

josunect:
> Code looks good and also verified that works fine. Regarding my health question, health for mc can be done in a follow up PR.

Thanks, @leandroberetta ! I can create the follow up PR and linked to this one.
I have rebased.

leandroberetta (dismissed):

leandroberetta (dismissed):
LGTM.

hhovsepy:
It seems the changes https://github.com/kiali/kiali/pull/5944/files#diff-d9a135c62e989128a43034ac55d64a837f21fedc602dbeadb6709874fa32ce3bR110 are somehow lost, so on single cluster case the columns are mixed, thus the test fails.

josunect:
> It seems the changes https://github.com/kiali/kiali/pull/5944/files#diff-d9a135c62e989128a43034ac55d64a837f21fedc602dbeadb6709874fa32ce3bR110 are somehow lost, so on single cluster case the columns are mixed, thus the test fails.

Thanks, Hayk! I've just sent a commit.

hhovsepy (approved):
Verified:
Screenshot from 2023-04-13 18-38-08

Issue #5962, opened 2023-03-29
Multi-cluster list view - Istio Config


nrfox:
Istio Config list page pulls Istio Config files from all configured clusters.

PR #5994, opened 2023-04-05
Istio Configs from multiple clusters

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5962

Screenshot from 2023-04-20 08-20-12

Config validations for remote clusters is a separate task https://github.com/kiali/kiali/issues/6056


hhovsepy:
The failure test is a flaky which is back.

business/istio_config.go
```diff
@@ -123,6 +124,49 @@ var newSecurityConfigTypes = []string{
...
+	for cluster := range in.userClients {
+		singleClusterConfigList, err := in.GetIstioConfigListPerCluster(ctx, criteria, cluster)
+		if err != nil {
+			if api_errors.IsNotFound(err) || api_errors.IsForbidden(err) {
+				// If a cluster is not found or not accessible, then we skip it
+				log.Debugf("Error while accessing to cluster [%s]: %s", cluster, err.Error())
+				continue
+			}
+
+			if cluster == kubernetes.HomeClusterName && len(in.userClients) == 1 {
```
josunect, line 139: Shouldn't be using the configuration name after this PR https://github.com/kiali/kiali/pull/6019 was merged?

`homeClusterName := config.Get().KubernetesConfig.ClusterName`

(I see some other uses like this in the code, also)

handlers/workloads.go
```diff
@@ -93,6 +96,11 @@ func WorkloadDetails(w http.ResponseWriter, r *http.Request) {
 
 	criteria := business.WorkloadCriteria{Namespace: p.Namespace, WorkloadName: p.WorkloadName, WorkloadType: p.WorkloadType, IncludeIstioResources: true, IncludeServices: true, IncludeHealth: p.IncludeHealth, RateInterval: p.RateInterval, QueryTime: p.QueryTime, Cluster: p.Cluster}
 
+	cluster := criteria.Cluster
+	if cluster == "null" || cluster == "" {
```
josunect, line 100: Maybe we could create a generic funcion, like isValidCluster, so it can be used by other handlers and also validates that the cluster exist and avoid to propagate the error to other layers. But this also be addressed later

josunect (commented):
Great job, @hhovsepy ! I am able to see the Istio configurations from all clusters: 

image


I have added a couple of comments. 

One thing I see is that I'm not able to create Istio objects, even in single cluster, I wonder if this might a regression? 

image

business/istio_config.go
hhovsepy: I think we need a separate PR to remove all deprecated HomeClusterName, I added this line in a parallel with #6019.

handlers/workloads.go
hhovsepy: Agree, but let's have a separate PR to cleanup this kind of things.

hhovsepy:
@josunect thank you for the review, can you confirm that on master there are permissions in wizards? I see the same problem in master.

business/istio_config.go
```diff
@@ -244,10 +296,11 @@ func (in *IstioConfigService) GetIstioConfigList(ctx context.Context, criteria I
 		if criteria.Include(kubernetes.DestinationRules) {
 			var err error
 			// Check if namespace is cached
+			// @TODO better way to check IsResourceCached per kubeCache
 			if IsResourceCached(criteria.Namespace, kubernetes.DestinationRules) {
-				istioConfigList.DestinationRules, err = in.cache.GetDestinationRules(criteria.Namespace, criteria.LabelSelector)
+				istioConfigList.DestinationRules, err = kubeCache.GetDestinationRules(criteria.Namespace, criteria.LabelSelector)
 			} else {
-				drl, e := in.k8s.Istio().NetworkingV1beta1().DestinationRules(criteria.Namespace).List(ctx, listOpts)
+				drl, e := in.userClients[cluster].Istio().NetworkingV1beta1().DestinationRules(criteria.Namespace).List(ctx, listOpts)
```
leandroberetta, line 303: Here, I'm not sure we need to do this, because as we are using informers/listers to fetch kubernetes resources, we need to use them to fetch things, the GetDestinationRules method from the cache, use the destionationRuleLister, and if I understand them correctly, they are subscribe to events that kubernetes send after there is an update/creation,etc, and there is an internal cache store every resource (the client-go library keeps them in sync with kubernetes).

leandroberetta (commented):
Hello @hhovsepy, nice PR! I have some questions after an initial review, I haven't tested yet.

business/istio_config.go
hhovsepy: Will not it be just an unreachable code in the when cache is always available?
But if by some reason cache is not available, it still can call directly the k8s client.
This is more historical , keeping the existing logic.

nrfox: IsResourceCached also checks that the type is cached. You can specify which istio types you want cached here: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.kubernetes_config.cache_istio_types. Not sure why you would ever not cache it and IMO we should remove that config option but it's still valid today.

business/istio_config.go
```diff
@@ -588,91 +645,91 @@ func (in *IstioConfigService) GetIstioConfigDetails(ctx context.Context, namespa
 
 	switch objectType {
 	case kubernetes.DestinationRules:
-		istioConfigDetail.DestinationRule, err = in.k8s.Istio().NetworkingV1beta1().DestinationRules(namespace).Get(ctx, object, getOpts)
+		istioConfigDetail.DestinationRule, err = in.userClients[cluster].Istio().NetworkingV1beta1().DestinationRules(namespace).Get(ctx, object, getOpts)
```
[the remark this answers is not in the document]
nrfox: Seems like there should at least be the same `if ResourceCached(...)` that uses the cache. Switching to using the cache could provide a significant perf improvement even in single cluster.

business/istio_config.go
nrfox: After seeing this in a few other places I'm wondering if instead of checking every occurrence in the business layer if it'd be better to check this in the handlers. It's probably worth also checking in the business layers since we really don't want to cause a panic.

business/istio_config.go
```diff
@@ -1126,16 +1187,16 @@ func (in *IstioConfigService) CreateIstioConfigDetail(namespace, resourceType st
...
 
-func (in *IstioConfigService) IsGatewayAPI() bool {
-	return in.k8s.IsGatewayAPI()
+func (in *IstioConfigService) IsGatewayAPI(cluster string) bool {
+	return in.userClients[cluster].IsGatewayAPI()
 }
 
 // Check if istio Ambient profile was enabled
 // ATM it is defined in the istio-cni-config configmap
 func (in *IstioConfigService) IsAmbientEnabled() bool {
```
nrfox, line 1196: Should this take `cluster` as an arg as well?

business/istio_validations.go
```diff
@@ -35,27 +35,33 @@ type ReferenceChecker interface {
...
+		observability.Attribute("cluster", cluster),
 		observability.Attribute("namespace", namespace),
 		observability.Attribute("service", service),
 		observability.Attribute("workload", workload),
 	)
 	defer end()
+
+	if cluster == "null" || cluster == "" {
+		cluster = kubernetes.HomeClusterName
+	}
```
nrfox, line 51: IMO the handlers should do this check

frontend/src/config/Config.ts
```diff
@@ -130,7 +130,11 @@ const conf = {
       istioConfig: (namespace: string) => `api/namespaces/${namespace}/istio`,
       allIstioConfigs: () => `api/istio/config`,
       istioConfigCreate: (namespace: string, objectType: string) => `api/namespaces/${namespace}/istio/${objectType}`,
-      istioConfigDetail: (namespace: string, objectType: string, object: string) =>
+      istioConfigDetail: (cluster: string, namespace: string, objectType: string, object: string) =>
+        `api/namespaces/${namespace}/istio/${objectType}/${object}?cluster=${cluster}`,
```
nrfox, line 134: I don't think query params belong here. These are just about constructing the required path. Query params are treated differently since they're optional and adding them here could cause issues with other things trying to construct the URL with its own query params.

frontend/src/helpers/EnvoyHelpers.tsx
```diff
@@ -15,7 +16,7 @@ export const istioConfigLink = (halfQDN: string, objectType: string): JSX.Elemen
   if (nameParts.length === 2) {
     return (
       <React.Fragment>
-        <IstioObjectLink name={nameParts[0]} namespace={nameParts[1]} type={objectType}>
+        <IstioObjectLink name={nameParts[0]} namespace={nameParts[1]} cluster={HomeClusterName} type={objectType}>
```
nrfox, line 19: Should this be `HomeClusterName`?

frontend/src/pages/IstioConfigList/IstioConfigListPage.tsx
```diff
@@ -99,20 +101,42 @@ class IstioConfigListPageComponent extends FilterComponent.Component<
     );
 
     if (namespacesSelected.length !== 0) {
-      this.fetchConfigs(namespacesSelected, istioTypeFilters, istioNameFilters, configValidationFilters, activeToggles);
+      this.setState({ listItems: [] });
+      if (isMultiCluster()) {
```
nrfox, line 105: You shouldn't need this if/else since for single cluster the loop will only iterate once for the one cluster.

nrfox: Actually since it looks like the istio config is aggregated across all clusters on the backend, do you need to make a request per cluster?

frontend/src/pages/ServiceDetails/ServiceDetailsPage.tsx
```diff
@@ -66,10 +66,10 @@ class ServiceDetails extends React.Component<ServiceDetailsProps, ServiceDetails
   constructor(props: ServiceDetailsProps) {
     super(props);
     const urlParams = new URLSearchParams(this.props.location.search);
-    const cluster = urlParams.get('cluster');
+    const cluster = urlParams.get('cluster') || HomeClusterName;
```
nrfox, line 69: `HomeClusterName` is not valid anymore. If the frontend does not provide cluster as a query param, then the backend can check for this and assume that you want the home cluster to keep backward compatability. Perhaps the backend should even throw an error if you're in "multicluster" mode and don't provide cluster as a query param but singlecluster should work the same in that you can use the same backend api calls without the cluster param.

handlers/namespaces.go
```diff
@@ -33,9 +33,17 @@ func NamespaceList(w http.ResponseWriter, r *http.Request) {
 // NamespaceValidationSummary is the API handler to fetch validations summary to be displayed.
 // It is related to all the Istio Objects within the namespace
 func NamespaceValidationSummary(w http.ResponseWriter, r *http.Request) {
+	query := r.URL.Query()
 	vars := mux.Vars(r)
 	namespace := vars["namespace"]
 
+	cluster := ""
+	if query.Has("cluster") && query.Get("cluster") != "" {
```
nrfox, line 41: I think you can simplify these checks to just be `query.Get("cluster") != ""`. `cluster==""` means the same thing as not providing cluster as a query param. Technically I think `""` could be a valid name for a cluster but if that's true then you should just check for `query.Has("cluster")`.

handlers/services.go
```diff
@@ -111,7 +111,7 @@ func ServiceDetails(w http.ResponseWriter, r *http.Request) {
 	namespace := params["namespace"]
 	service := params["service"]
 	queryTime := util.Clock.Now()
-	rateInterval, err = adjustRateInterval(r.Context(), business, namespace, rateInterval, queryTime)
+	//rateInterval, err = adjustRateInterval(r.Context(), business, namespace, rateInterval, queryTime)
```
nrfox, line 114: Does this need to be uncommented?

josunect:
> @josunect thank you for the review, can you confirm that on master there are permissions in wizards? I see the same problem in master.

Thanks for pointing at this. There is the same behavior in master.

business/istio_config.go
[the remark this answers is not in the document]
hhovsepy: Details of Istio config showing and as well as editing is done directly by K8s api, not from the cache. As here not a bunch of configs is loaded but a single one which does not impact on performance as much as lists.

frontend/src/helpers/EnvoyHelpers.tsx
hhovsepy: having this default value as a TODO to resolve later

frontend/src/pages/IstioConfigList/IstioConfigListPage.tsx
hhovsepy: this is for the case when serverConfig.clusters is empty which was happening on testing environment

hhovsepy: the istio config is returned from backend per cluster (called per cache/client for cluster), it is not aggregated across all clusters, otherwise that would require a lot of API JSON changed

frontend/src/pages/ServiceDetails/ServiceDetailsPage.tsx
hhovsepy: this is done as 'cluster' is a mandatory attribute, similar to other places we have, agree that this is not the best solution, and we need a separate PR for resolving them at all and 'HomeClusterName' is a keyword where to look at.
When the value is empty, the links generated does not have a 'cluster' param.

frontend/src/pages/IstioConfigList/IstioConfigListPage.tsx
leandroberetta: Took me some time to see too this, now the handler calls the `GetIstioConfigListPerCluster` function and not the former one which is the one that is aggregated now.

I saw that the aggregated function, `GetIstioConfigList` is called by for example the `GetServiceList` service, which can include Istio configs. In that service, now the Istio configs is aggregated, and the service list is also aggregated. If we are appending Istio configs (in the GetIstioConfigList) with no cluster information, then a service from a cluster will have, let's say, destination rules for all of the clusters?.

leandroberetta (commented):
Overall code looks good, but I have a question below.

frontend/src/pages/IstioConfigList/IstioConfigListPage.tsx
hhovsepy: Good catch, I missed that, for GetServiceList it will call GetIstioConfigListPerCluster

leandroberetta: Sure, no problem, there are also another references to the list function, so I think in all places will be required to used the one with the cluster. 

Will be required the aggregated function?.

hhovsepy: The aggregated function GetIstioConfigList is left for those areas when multicluster is not applied yet or is in still progress and is out of scope from this PR.
For the rest of places where GetIstioConfigList is called we will be back later. Otherwise the scope of this PR becomes endless.

hhovsepy: there are several subtasks created to cleanup the code https://github.com/kiali/kiali/issues/5618

leandroberetta (approved):
LGTM. Thanks for addressing the feedback.

business/istio_config.go
josunect: I think so. But, the only thing to take into account is that we should have access to the Istio Control plane - We won't have one in a primary-remote

PR #5998, opened 2023-04-05
Multicluster services list page

### LINKED ISSUE #5943: Multicluster - Services view
Services list page shows services from multiple clusters. Some work on the detail page may need to be included so that links are not broken but getting everything to work on the details page will be a separate issue.

#5919

### LINKED ISSUE #5960: Multi-cluster list view - Services
Workloads list page pulls workloads from all configured clusters.

nrfox:
Implements multi cluster services list.

Fixes #5960

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -55,7 +55,9 @@ func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriter
 	)
 	defer end()
 
-	serviceList := models.ServiceList{}
+	serviceList := models.ServiceList{
+		Validations: models.IstioValidations{},
+	}
 	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
 	// if namespace is accessible from Kiali (Deployment.AccessibleNamespaces)
 	for cluster := range in.userClients {
@@ -88,8 +90,8 @@ func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriter
 		}
 
 		serviceList.Services = append(serviceList.Services, singleClusterSVCList.Services...)
-		// TODO: Do we need to aggregate namespaces?
 		serviceList.Namespace = singleClusterSVCList.Namespace
+		serviceList.Validations = serviceList.Validations.MergeValidations(singleClusterSVCList.Validations)
 	}
 
 	return &serviceList, nil
--- a/business/services.go
+++ b/business/services.go
@@ -221,7 +221,8 @@ func (in *SvcService) getServiceList(ctx context.Context, criteria ServiceCriter
 
 	if criteria.IncludeHealth {
 		for i, sv := range services.Services {
-			services.Services[i].Health, err = in.businessLayer.Health.GetServiceHealth(ctx, cluster, criteria.Namespace, sv.Name, criteria.RateInterval, criteria.QueryTime, sv.ParseToService())
+			// TODO: Fix health for multi-cluster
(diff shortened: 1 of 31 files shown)


kubernetes/cache/namespaces.go
nrfox, line 1: I updated this to include `cluster` in the key because without this the cache could return a namespace for a different cluster and most callers of the cache are just checking if `ns != nil`. See [this test](https://github.com/kiali/kiali/pull/5998/files#diff-b4f5651eca4cec992ad548c9625937d5cc407a526449db05e3ac0f6184a57db2R237-R268) for more details.

hhovsepy:
The code looks good to me, have run locally:
Screenshot from 2023-04-06 11-12-58

hhovsepy:
Regression tested on single OCP cluster.

prometheus/metrics.go
```diff
@@ -144,8 +144,8 @@ func getNamespaceServicesRequestRates(ctx context.Context, api prom_v1.API, name
 // getServiceRequestRates retrieves traffic rates for requests entering, or internal to the namespace, for a specific service name
 // Note that it does not discriminate on "reporter", so rates can be inflated due to duplication, and therefore
 // should be used mainly for calculating ratios (e.g total rates / error rates)
-func getServiceRequestRates(ctx context.Context, api prom_v1.API, namespace, service string, queryTime time.Time, ratesInterval string) (model.Vector, error) {
-	lbl := fmt.Sprintf(`destination_service_name="%s",destination_service_namespace="%s"`, service, namespace)
+func getServiceRequestRates(ctx context.Context, api prom_v1.API, cluster, namespace, service string, queryTime time.Time, ratesInterval string) (model.Vector, error) {
+	lbl := fmt.Sprintf(`destination_service_name="%s",destination_service_namespace="%s",destination_cluster="%s"`, service, namespace, cluster)
```
leandroberetta, line 148: Do we want just traffic from the cluster that the service lives in, or from any cluster the traffic comes from?.

business/services.go
```diff
@@ -48,24 +49,70 @@ func (in *SvcService) GetServiceList(ctx context.Context, criteria ServiceCriter
...
-	var pods []core_v1.Pod
-	var deployments []apps_v1.Deployment
-	var istioConfigList models.IstioConfigList
-	var err error
-
+	serviceList := models.ServiceList{}
 	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
 	// if namespace is accessible from Kiali (Deployment.AccessibleNamespaces)
-	if _, err = in.businessLayer.Namespace.GetNamespace(ctx, criteria.Namespace); err != nil {
+	for cluster := range in.userClients {
```
leandroberetta, line 55: Performance wise, can we implement this fetches with go routines?.

leandroberetta (commented):
Code looks good, I have only two comments below.

prometheus/metrics.go
leandroberetta: I think you're right. We are interested in the traffic rates from any service/cluster to the destination service, which lives in a cluster.

nrfox:
Putting "do no merge" label until #6007 is fixed since that will affect this PR too.

josunect (commented):
Service list with multicluster: 

image

image

But shouldn't the configuration be applicable for the local cluster? This is an example (from master branch) 

image

nrfox:
> But shouldn't the configuration be applicable for the local cluster? This is an example (from master branch)

You're right! I forgot to add the validations to the response. Good catch.

hhovsepy:
Tested locally:
Screenshot from 2023-04-13 20-34-09

hhovsepy:
The missing '?cluster' params in service detils references could be added in a further PRs.

hhovsepy (approved):

PR #5999, opened 2023-04-06
[OSSM-3496]Adding a Dockerfile for integration tests to be used with OpenShiftCI

FilipB:
Image produced by this Dockerfile will be used by the interop pipeline in OpenShift CI.
We can add targets for building and pushing to the makefile later. This is just to test integration with openShiftCI.

### DIFF
--- a/tests/integration/Dockerfile
+++ b/tests/integration/Dockerfile
@@ -1,0 +1,29 @@
+FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
+
+ENV GOPATH=/go
+ENV PATH=/usr/local/go/bin:$GOPATH/bin:$PATH
+
+# install required packages and prepare go dirs
+WORKDIR /bin
+RUN microdnf install --nodocs tar gzip make \
+    && curl -Lo ./oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz \
+    && tar -xf oc.tar.gz \
+    && rm -f oc.tar.gz \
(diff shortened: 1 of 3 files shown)


FilipB:
@jmazzitelli only risk is the new '.dockerignore' file but I think it should not influence other image builds.

tests/integration/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+ENV GOPATH=/go
+ENV PATH=/usr/local/go/bin:$GOPATH/bin:$PATH
+
+# install required packages and prepare go dirs
+WORKDIR /bin
+RUN microdnf install --nodocs tar gzip make \
+    && curl -Lo ./oc.tar.gz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux.tar.gz \
+    && tar -xf oc.tar.gz \
+    && rm -f oc.tar.gz \
+    && curl -Lo ./golang.tar.gz https://go.dev/dl/go1.20.2.linux-amd64.tar.gz \
```
nrfox, line 12: This all looks good. My only suggestion would be to add a `make` command to generate this dockerfile. Then you can also plumb through the same go version that kiali uses here either as a build arg or something: https://github.com/kiali/kiali/blob/bfbf8c76ff399e24eb9eff232054361744e415e3/Makefile#L32

jmazzitelli (approved):

tests/integration/Dockerfile
FilipB: Thank you for the suggestion. I can't work on that now so I created a follow up task - https://issues.redhat.com/browse/OSSM-3710

Issue #6000, opened 2023-04-06
Perf regression when fetching workloads


nrfox:
Related to #5784. The perf regression was re-introduced as part of other changes. This keeps happening because of the `CachingClient` abstraction which obscures when the cache is being used. 
https://github.com/kiali/kiali/blob/bfbf8c76ff399e24eb9eff232054361744e415e3/kubernetes/cache/cachingclient.go#L10-L17

That abstraction should be removed and the workloads service should be updated to use the kube cache directly.

PR #6002, opened 2023-04-06
Revert multi-cluster workload changes for v1.66

nrfox:
Because of #6000


hhovsepy (approved):

nrfox:
Just removed the CI yaml change from this patch since the tests don't work anyway even when they run.

PR #6005, opened 2023-04-06
Remove caching client and update workload svc to use cache

### LINKED ISSUE #6000: Perf regression when fetching workloads

nrfox:
Removes the caching client and updates the workload service to use the cache. Includes adding cluster to the Namespace cache key as part of this change. Note that after this change, there are duplicate entries for workloads on the workload list page in multi-cluster but that will go away once the namespace dropdown selector correctly aggregates the duplicate namespace.

Fixes #6000


leandroberetta (commented):
Looks good, I tested it and saw the duplicated workloads in the list as you mentioned in the description, for example in the bookinfo namespace.

I have a comment related to the CacheEnabled field in the configuration, searching in the code, seems that there are still occurrences of this field in some tests and also in 

https://github.com/kiali/kiali/blob/667c8e294110cc628afdfbd2ca36838e7d78144a/business/layer.go#L67

Should this PR remove also the above check or a following PR should do it?.

nrfox:
> Should this PR remove also the above check or a following PR should do it?

That check is purely internal now since the ability to disable the cache externally has already been removed but the checks still appear because some unit tests still rely on them. I haven't tried removing the line above but removing it could break other tests. There's already a separate issue to disable caching: #5599 so I'd say removing these references can be part of that issue.

leandroberetta (approved):
Looks good.

PR #6025, opened 2023-04-13
Multicluster - Filter duplicate namespaces in namespace dropdown

### LINKED ISSUE #6007: Multicluster - Combine duplicated namespaces in dropdown selector
Namespaces that are duplicated across clusters should be combined into a single namespace name.

Related to #5932 

#5618

nrfox:
This combines all namespaces of the same name in the namespace dropdown. This should also fix the problem of duplicate entries showing up on the list pages. This works because the Kiali API makes a list call per namespace selected in the dropdown. On the backend, the business services will reach out to every connected cluster returning the aggregated results across each. Now that there are no longer duplicate namespaces in the dropdown and there's only one list call per namespace regardless of how many clusters that namespace appears in, we no longer see duplicate entries for workloads.

Fixes #6007


leandroberetta (approved):
LGTM.

Issue #3914, opened 2021-04-20
Support for Istio's discoverySelectors


israel-hdez:
**Is your feature request related to a problem? Please describe.**
In Istio 1.10, a new `discoverySelectors` configuration was introduced. (See [Istio 1.10 docs](https://preliminary.istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig), [design document](https://docs.google.com/document/d/1S3GPCUBmTmvh09NmpuMCG3vwDTNkVeRhGT3lbLyA4qs/edit?resourcekey=0-a8iMutRwjQElYN5dlgEDdw#heading=h.gryauiau89tc), and the [merged PR implementing the feature](https://github.com/istio/istio/pull/29802)).

The `discoverySelectors` option limits the set of namespaces that Istio processes. If I understand correctly, this means that only the _selected_ namespaces will be part of the mesh.

**Describe the solution you'd like**
Currently, Kiali assumes that the Istio ControlPlane is given access to all namespaces. Despite this, Kiali can be configured with a restricted set of namespaces (aka `accessible_namespaces`). 

However, if Istio's `discoverySelectors` configuration is set, it's **probably** desired that Kiali takes into accounts this setting to also properly limit namespace visibility, to make scoped validations, to treat any service outside the configured set as external to the mesh, etc.

**Describe alternatives you've considered**
We could leave Kiali as is and ask users to use the `accessible_namespaces` option.

lucasponce:
It's interesting.
I think we need to explore how to integrate in the future.

Note that in a previous Istio Meetup one user described a real scenario where he would like to see how a migration into a ServiceMesh is going, meaning that showing namespaces and workloads that are part of the mesh and which not is also a business goal that Kiali addresses today.

So, I guess we'll have users that may want to see only namespaces/workloads in the mesh, or users that may see both.

jshaughn:
As I understand it it's not necessarily for visibility but rather for efficiency.  The control plane can't distinguish namespaces that are and are not in the mesh (I guess), and so it just assumes everything is in the mesh.  And that increases what it needs to watch.  In OSSM I believe we are explicitly told which namespace are in the mesh, but upstream did not have anything similar.

As for what we do with DiscoverSelectors, I guess we could minimally display them, and potentially configure them.  As for our accessible_namespaces, it would seem that that could be automatically narrowed to be a proper subset of Istio's discoveredNamespaces.

[...]

PR #6027, opened 2023-04-13
use discovery selectors to filter namespaces

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/3914

operator PR: https://github.com/kiali/kiali-operator/pull/628

The Istio ConfigMap has discoverySelectors that Kiali will now use to filter namespaces.

If no discoverySelectors are specified, or discoverySelectors is specified as an empty list in the ConfigMap, then Kiali works as before.

If discoverySelectors are specified, then they will filter out any namespaces that do not match. After that filter is performed, Kiali will further exclude namespaces defined in the excludes config settings in the Kiali CR (this works as before).

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -10,6 +10,8 @@ import (
 	osproject_v1 "github.com/openshift/api/project/v1"
 	core_v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
@@ -80,6 +82,26 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 	}
 
 	configObject := config.Get()
+
+	// determine what the discoverySelectors are by examining the Istio ConfigMap
+	var discoverySelectors []*meta_v1.LabelSelector
+	if kialiCache != nil {
+		if icm, err := kialiCache.GetConfigMap(configObject.IstioNamespace, configObject.ExternalServices.Istio.ConfigMapName); err == nil {
+			if ic, err2 := kubernetes.GetIstioConfigMap(icm); err2 == nil {
+				discoverySelectors = ic.DiscoverySelectors
+			} else {
+				log.Errorf("Will not process discoverySelectors due to a failure to get the Istio ConfigMap: %v", err2)
+			}
+		} else {
+			log.Errorf("Will not process discoverySelectors due to a failure to parse the Istio ConfigMap: %v", err)
+		}
+	}
+
+	if len(discoverySelectors) > 0 {
+		log.Tracef("Istio discovery selectors: %+v", discoverySelectors)
+	} else {
+		log.Tracef("No Istio discovery selectors defined.")
+	}
 
 	// Let's explain the four different filters along with accessible namespaces (aka AN).
 	//
@@ -169,10 +191,10 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 	for resultCh := range resultsCh {
 		if resultCh.err != nil {
 			if resultCh.cluster == kubernetes.HomeClusterName {
-				log.Errorf("Error fetching Namespaces for local cluster %s: %s", resultCh.cluster, resultCh.err)
+				log.Errorf("Error fetching Namespaces for local cluster [%s]: %s", resultCh.cluster, resultCh.err)
 				return nil, resultCh.err
 			} else {
-				log.Infof("Error fetching Namespaces for cluster %s: %s", resultCh.cluster, resultCh.err)
+				log.Infof("Error fetching Namespaces for cluster [%s]: %s", resultCh.cluster, resultCh.err)
 				continue
 			}
 		}
@@ -180,6 +202,40 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 	}
 
 	resultns := namespaces
+
+	// Filter out those namespaces that do not match discoverySelectors.
+	// Follow the semantics that Istio follows, which is:
+	//   If there is no discoverySelectors section in the config, skip this entirely.
+	//   If there is an empty discoverySelectors section, that means all namespaces are to be used.
+	//   If there are one or more discoverySelectors specified, the filter namespaces based on what they select.
+	if len(discoverySelectors) > 0 {
+		// 1. convert LabelSelectors to Selectors
+		selectors := make([]labels.Selector, 0)
+		for _, selector := range discoverySelectors {
+			ls, err := meta_v1.LabelSelectorAsSelector(selector)
+			if err != nil {
+				return nil, fmt.Errorf("error initializing discovery selectors filter, invalid discovery selector: %v", err)
+			}
+			selectors = append(selectors, ls)
+		}
+
+		// 2. range over all namespaces to get discovery namespaces, notice each selector result is ORed (as per Istio convention)
+		selectedNamespaces := make([]models.Namespace, 0)
+		for _, ns := range resultns {
+			if ns.Name == configObject.IstioNamespace {
+				selectedNamespaces = append(selectedNamespaces, ns) // we always want to return the control plane namespace
+			} else {
+				for _, selector := range selectors {
+					if selector.Matches(labels.Set(ns.Labels)) {
+						selectedNamespaces = append(selectedNamespaces, ns)
+						break
+					}
+				}
+			}
+		}
+		namespaces = selectedNamespaces
+		resultns = namespaces
+	}
 
 	// exclude namespaces that are:
 	// 1. to be filtered out via the exclude list
--- a/config/config.go
+++ b/config/config.go
@@ -541,6 +541,7 @@ func NewConfig() (c *Config) {
 		CustomDashboards: dashboards.GetBuiltInMonitoringDashboards(),
 		Deployment: DeploymentConfig{
 			AccessibleNamespaces: []string{"**"},
+			ClusterWideAccess:    true,
 			InstanceName:         "kiali",
 			Namespace:            "istio-system",
 			ViewOnlyMode:         false,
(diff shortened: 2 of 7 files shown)


business/namespaces.go
```diff
@@ -78,6 +80,26 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
+			return nil, fmt.Errorf("Failed to get the Istio ConfigMap: %v", err)
+		}
+		ic, err := kubernetes.GetIstioConfigMap(icm)
+		if err != nil {
+			return nil, fmt.Errorf("Failed to parse the Istio ConfigMap: %v", err)
+		}
+		istioConfig = ic
+	}
+
+	if istioConfig.DiscoverySelectors != nil && len(istioConfig.DiscoverySelectors) > 0 {
```
nrfox, line 97: Just trying to get your tests passing... you likely need an additional check for `istioConfig != nil`

jmazzitelli: Well, this is what I'm struggling with. In production, the Kiali server should always have access to the istio configmap - that value should never be nil (or, technically, you should never get here if its nil). This function aborts with an error if the configmap cannot be processed - so you should never get here.

Obviously, I could just not abort and keep going with a nil ConfigMap but:

(a) that would solely be to support unit testing and is not valid in production - and that sucks to have production code written in a way solely to support unit tests, especially when it really isn't "right"

(b) IMO, if we can't get the DS, then we need to abort. This could be considered a security issue. Because suppose Kiali is configured with ** but the user wants to only make available those namespaces filtered by DS - if we do not have access to what the DSs are - we should NOT default to all namespace (i.e. we should not default to a list of namespaces that were not filtered by DS). To me, that is a security flaw. If we can't get to the DS, then we need to abort and tell the user we are broken until Kiali has access to see the Istio ConfigMap so it can correctly display only those namespaces we should be.

nrfox: > Well, this is what I'm struggling with. In production, the Kiali server should always have access to the istio configmap - that value should never be nil (or, technically, you should never get here if its nil). This function aborts with an error if the configmap cannot be processed - so you should never get here.

Kiali may never have access to external control planes or istio may temporarily be down in which case Kiali should still keep functioning.

> This could be considered a security issue.

I thought DS was not about access? Isn't that what `cluster_wide_access` is for? DS is not a security boundary for istio either I don't think.

jmazzitelli: If we all don't think this is a security-issue to worry about, I can relax that code. I'm security-paranoid because I know all the times people have complained to us that we expose more than we should, which is why I was hesitant.

jmazzitelli: This has been reworked so the server continues on even if the Istio ConfigMap is not found or for some reason can't be parsed. It will still log an error but it keeps going.

jmazzitelli: lo-and-behold almost all the unit tests pass again :)

There is one failure that I don't understand:
```
--- FAIL: TestTlsPerfNsDr (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x18022ae]

goroutine 8785 [running]:
testing.tRunner.func1.2({0x1c0ef60, 0x311c200})
	/home/jmazzite/bin/go-install/go/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/home/jmazzite/bin/go-install/go/src/testing/testing.go:1399 +0x39f
panic({0x1c0ef60, 0x311c200})
	/home/jmazzite/bin/go-install/go/src/runtime/panic.go:884 +0x212
github.com/kiali/kiali/kubernetes/cache.(*kialiCacheImpl).GetConfigMap(0xc000818790?, {0x1ef387a?, 0x1ef1e16?}, {0x1ee9965?, 0xc000b42a00?})
	<autogenerated>:1 +0x2e
github.com/kiali/kiali/business.(*NamespaceService).GetNamespaces(0xc000a30e70, {0x21fa8f0, 0xc0001ac008})
	/home/jmazzite/source/kiali/kiali/business/namespaces.go:86 +0x278
github.com/kiali/kiali/business.(*TLSService).getNamespaces(0x21fa8f0?, {0x21fa8f0?, 0xc0001ac008?})
```


I don't get it because this is wrapped in the kialiCache != nil test, so something else must be blowing up

jmazzitelli: this test has been fixed with a not-so-obvious change.

jmazzitelli:
@nrfox I need suggestions to know how to use the new "cluster wide access" setting - the cache will need to use it. (in fact I forgot to add that to the config.go in this PR - I'll do that next). But I need to know where/how the cache will make use of it.

nrfox:
> @nrfox I need suggestions to know how to use the new "cluster wide access" setting - the cache will need to use it. (in fact I forgot to add that to the config.go in this PR - I'll do that next). But I need to know where/how the cache will make use of it.

I believe you just need to change this method to return the new setting instead of looking for `**` in AN: https://github.com/kiali/kiali/blob/3b4ecc9c5153bd7c12f7178146f3d8592db3ae4b/config/config.go#L786-L796

jmazzitelli:
OK, latest commit updates that function and sets the defaults in a way that unit tests continue to pass.

jmazzitelli:
all operator molecule tests pass with this PR applied to the server image

jmazzitelli:
One note that needs to be mentioned. The main difference the user will notice between the implementation where the operator processes discoverySelectors, aka DS (the previous WIP implementation) and now where the server processes DS, the user will have to know to create an AN (accessible_namespaces list) that at minimum includes the namespaces selected by the DS. Is this a burden on the user?

jshaughn:
> The main difference the user will notice between the implementation where the operator processes discoverySelectors, aka DS (the previous WIP implementation) and now where the server processes DS, the user will have to know to create an AN (accessible_namespaces list) that at minimum includes the namespaces selected by the DS. Is this a burden on the user?

I think the new approach may just require the same from users as we did in the past.  Existing users already using an explicit AN should be fine or using ** should also be fine, right?  So this is mainly for new users, I think.  And I think the guidance is the same, If using namespace-security then you need the specific AN list.  If cluster-wide you can have an explicit AN or not, up to you.  Either way, the DS definitions, if any, can only narrow the accessible namepaces.

nrfox (commented):
One question otherwise LGTM

business/namespaces.go
```diff
@@ -78,6 +80,26 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 
 	configObject := config.Get()
 
+	// determine what the discoverySelectors are by examining the Istio ConfigMap
+	var discoverySelectors []*meta_v1.LabelSelector
+	if kialiCache != nil {
+		if icm, err := kialiCache.GetConfigMap(configObject.IstioNamespace, configObject.ExternalServices.Istio.ConfigMapName); err == nil {
```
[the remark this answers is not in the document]
jmazzitelli: Good point.. do we have a function that can be used to isolate the configmap name? Something like "getIstioConfigMap" and have it deal with the versioning/canary details? I believe with canary deployments, it also uses a differently named config map (??)

@leandroberetta might know, he did some canary work in the past ^^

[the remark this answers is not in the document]
nrfox: > Good point.. do we have a function that can be used to isolate the configmap name? Something like "getIstioConfigMap" and have it deal with the versioning/canary details?

That would be useful. Probably can be added to the config object.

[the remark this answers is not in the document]
leandroberetta: It's a good point, every time you install a canary, you'll get a new config map for that revision. Particulary in the canary work I've done, I'm not using the config map, but there are places that the config map is used. 

So, after a canary, eventually, Kiali will need to grab the new configmap (has a different name). I will create an issue to document this better, and to see if it is just that.

nrfox (approved):
LGTM. I think this needs an e2e test but I'll add the label and it can be included with the PR or after.

Xunzhuo (approved):
Overall looks good to me, thanks

jshaughn (approved):
This looks OK to me.  The '**' use is an internal impl detail from what I can see.  We want to remove '**' from any public exposure as it's very confusing and should be replaced by `cluster_wide_access: true`.

jshaughn:
Added DNM label because the Operator PR may need some work.

jshaughn:
@Xunzhuo Just a question, do you use istio Discovery Selectors in your environment?

Xunzhuo:
@jshaughn yes, we use it for multi Istio control plane in a single cluster

jmazzitelli:
fixed a conflict that was caused by a major refactoring if istio_test.go. I squashed the commits in this PR so conflicts won't be difficult next time. But I confirmed the code is the same as it was before the squash.

config/config.go
```diff
@@ -361,6 +361,7 @@ type OpenIdConfig struct {
 // DeploymentConfig provides details on how Kiali was deployed.
 type DeploymentConfig struct {
 	AccessibleNamespaces []string `yaml:"accessible_namespaces"`
+	ClusterWideAccess    bool     `yaml:"cluster_wide_access,omitempty"`
```
jshaughn, line 364: I believe with `omitempty` a bool set to `false` will be omitted.   As long as unset would always reflect `false` I guess this is OK, but given the semantics of CWA I'm wondering if it's best to not omit.

jmazzitelli: `omitempty` is meaningless in our context and not used. In fact, we could remove every mention of `omitempty` in this file and it would do nothing to us because it is only used for marshaling into YAML and AFAIK we do not marshal into yaml (we only unmarshal from the ConfigMap to the internal Go structs - we don't go the other way around and go from the structs to a YAML string - we certainly don't write back to the ConfigMap nor write to a yaml file anywhere).

See: https://pkg.go.dev/gopkg.in/yaml.v2#Marshal

```
omitempty    Only include the field if it's not set to the zero
             value for the type or to empty slices or maps.
             Zero valued structs will be omitted if all their public
             fields are zero, unless they implement an IsZero
             method (see the IsZeroer interface type), in which
             case the field will be excluded if IsZero returns true.
```

jshaughn: It's fine if we're never going to marshal to YAML.  But in general, for booleans, I've been burned when using omitempty, so it was just a mention.

jshaughn (approved):
Code looks fine.  DS are narrowing AN as expected.  LG2M.

PR #6029, opened 2023-04-13
[OSSM-1216] cypress container

ScriptingShrimp:
First iteration of docker file for cypress execution in container

### DIFF
--- a/deploy/docker/Dockerfile-cypress
+++ b/deploy/docker/Dockerfile-cypress
@@ -1,0 +1,48 @@
+# don't forget to pass environmet variables to docker run like this:
+# podman run -e CYPRESS_BASE_URL=<kiali_url> -e CYPRESS_PASSWD=<passwd> docker-cypress:latest
+# more details here https://github.com/kiali/kiali/tree/master/frontend/cypress
+
+# using same baseimage of node as its defined in package.json
+FROM cypress/base:14.17.0
+
+# Run as non-root user for MP+/OpenShift
+USER node
+
+ENV KIALI_HOME=/opt/kiali \
+    PATH=$KIALI_HOME:$PATH
+
+WORKDIR $KIALI_HOME
+
+# "Fake" dbus address to prevent errors. (is this still needed?)
+# https://github.com/SeleniumHQ/docker-selenium/issues/87
+ENV DBUS_SESSION_BUS_ADDRESS=/dev/null
+
+# A few environment variables to make NPM installs easier.
+# Good colors for most applications.
+ENV TERM xterm
+# Avoid million NPM install messages.
+ENV npm_config_loglevel warn
+# Allow installing when the main user is root.
+ENV npm_config_unsafe_perm true
+
+# copy frontend folder to container
+COPY --chown=node:node frontend/ $KIALI_HOME/
+# Install Cypress dependencies.
+RUN /usr/local/bin/yarn install --frozen-lockfile
+
+# Run Cypress tests.
+CMD ["yarn", "cypress:run:junit"]
+
+
+# TODO:
+# add oc and kubectl client binaries
+# log in clients above - all test using kubectl failed on `/bin/bash: kubectl: command not found`
+#
+# TODO:
+# shared folder with host for 
+# 1. test results (XML)
+# 2. videos and screenshots
+# 
+# TODO:
+# makefile target for build?
+# makefile target for run?


deploy/docker/Dockerfile-cypress
```diff
@@ -0,0 +1,48 @@
...
+# log in clients above - all test using kubectl failed on `/bin/bash: kubectl: command not found`
+#
+# TODO:
+# shared folder with host for 
+# 1. test results (XML)
+# 2. videos and screenshots
+# 
+# TODO:
+# makefile target for build?
+# makefile target for run?
```
ScriptingShrimp, line 48: @FilipB @jmazzitelli any suggestions please?

ScriptingShrimp: base image is [Debian Buster](https://github.com/cypress-io/cypress-docker-images/blob/master/base/14.17.0/Dockerfile)

deploy/docker/Dockerfile-cypress
```diff
@@ -0,0 +1,48 @@
...
+# Run as non-root user for MP+/OpenShift
+USER node
+
+ENV KIALI_HOME=/opt/kiali \
+    PATH=$KIALI_HOME:$PATH
+
+WORKDIR $KIALI_HOME
+
+# "Fake" dbus address to prevent errors. (is this still needed?)
+# https://github.com/SeleniumHQ/docker-selenium/issues/87
```
FilipB, line 17: How is selenium involved?

deploy/docker/Dockerfile-cypress
```diff
@@ -0,0 +1,48 @@
...
+# Allow installing when the main user is root.
+ENV npm_config_unsafe_perm true
+
+# copy frontend folder to container
+COPY --chown=node:node frontend/ $KIALI_HOME/
+# Install Cypress dependencies.
+RUN /usr/local/bin/yarn install --frozen-lockfile
+
+# Run Cypress tests.
+CMD ["yarn", "cypress:run:junit"]
```
FilipB, line 34: We will also need 'yarn cypress:combine:reports' I guess

deploy/docker/Dockerfile-cypress
```diff
@@ -0,0 +1,48 @@
...
+# add oc and kubectl client binaries
+# log in clients above - all test using kubectl failed on `/bin/bash: kubectl: command not found`
+#
+# TODO:
+# shared folder with host for 
+# 1. test results (XML)
+# 2. videos and screenshots
+# 
+# TODO:
+# makefile target for build?
```
FilipB, line 47: I think we can do this later as a follow up task. Now we just need a basic Dockerfile to onboard to openShift CI

deploy/docker/Dockerfile-cypress
```diff
@@ -0,0 +1,48 @@
...
+# Run Cypress tests.
+CMD ["yarn", "cypress:run:junit"]
+
+
+# TODO:
+# add oc and kubectl client binaries
+# log in clients above - all test using kubectl failed on `/bin/bash: kubectl: command not found`
+#
+# TODO:
+# shared folder with host for
```
FilipB, line 42: I believe test results are expected in SHARED_DIR folder. e.g. ${SHARED_DIR}/junit_output.xml

FilipB (changes requested):
Not sure about this but it might make better sense to put the Dockerfile into cypress folder? Instead of 'deploy/docker' folder?

deploy/docker/Dockerfile-cypress
ScriptingShrimp: I can add `cypress:combine:reports` into the npm scripts or extend this CMD
but
I'm still not sure how do I get XML file outside of container

deploy/docker/Dockerfile-cypress
ScriptingShrimp: :+1:

deploy/docker/Dockerfile-cypress
ScriptingShrimp: Do we have some example which I can reuse?

ScriptingShrimp:
> Not sure about this but it might make better sense to put the Dockerfile into cypress folder? Instead of 'deploy/docker' folder?

I asked team where to put it, looks like everyone is fine with `deploy/dockerfile` folder. I also saw make targets navigate here for other docker related things.

deploy/docker/Dockerfile-cypress
FilipB: We don't have to worry about that now. We just need to put the xml to correct location and it will be used later in the pipeline. The ${SHARED_DIR} should be correct location and it should be available for later pipeline steps.

deploy/docker/Dockerfile-cypress
FilipB: I don't have any example yet. SHARED_DIR env variable should be available in openShiftCI so we can just put xml there. I don't know yet how to make it work both in openShiftCI and when running as a standalone container outside of openshiftCI

deploy/docker/Dockerfile-cypress
ScriptingShrimp: It's old fix for standalone chrome. Verified that it have no effects. Removing

.dockerignore
```diff
@@ -0,0 +1,4 @@
+_output
+operator
+deploy
+frontend/node_modules
```
ScriptingShrimp, line 4: this will affect other Dockerfiles in this folder, is that safe @FilipB

FilipB: It should be safe as other images are build using CONTEXT ${OUTDIR}/docker, e.g. https://github.com/kiali/kiali/blob/master/make/Makefile.container.mk#L17

FilipB (approved):
LGTM

.dockerignore
ScriptingShrimp: I don't understand it precisely, but I believe you  :smile:

Issue #6028, opened 2023-04-13
Multicluster - Update the namespace health service


leandroberetta:
The namespace health service should be updated to return health for applications/services/workloads considering the cluster.

Epic: #5618

PR #6030, opened 2023-04-14
Multicluster - Namespace health

leandroberetta:
Resolves https://github.com/kiali/kiali/issues/6028

Belongs to epic #5618 

This PR adds namespace health for multicluster using the cluster as a query parameter to fetch health for a namespace.

This PR also addresses the Prometheus queries to include the cluster label.

The namespace health can be get for applications, services and workloads, and all of these are addressed in this PR.

image

To test this, you'll need:

1. Checkout the code:

`gh pr checkout 6030
`

2. Prepare the images

`make clean build build-ui container-build
`

3. A multicluster setup, a primary-remote will work:

`./hack/istio/multicluster/install-primary-remote.sh -kudi true
`

4. Deploy errors rate demo (distributed into the clusters):

`./hack/istio/install-error-rates-demo.sh -c kubectl -dd true`

4. Permissions

```
kubectl create rolebinding bookinfo-admin --clusterrole=admin  --user=oidc:kiali -n bookinfo --context west
kubectl create rolebinding beta-admin --clusterrole=admin  --user=oidc:kiali -n beta --context west
```

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -8,7 +8,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/kiali/kiali/kubernetes"
-	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/observability"
 	"github.com/kiali/kiali/prometheus"
@@ -24,6 +23,7 @@ type HealthService struct {
 type NamespaceHealthCriteria struct {
 	IncludeMetrics bool
 	Namespace      string
+	Cluster        string
 	QueryTime      time.Time
 	RateInterval   string
 }
@@ -117,45 +117,31 @@ func (in *HealthService) GetWorkloadHealth(ctx context.Context, namespace, workl
 }
 
 // GetNamespaceAppHealth returns a health for all apps in given Namespace (thus, it fetches data from K8S and Prometheus)
-func (in *HealthService) GetNamespaceAppHealth(ctx context.Context, criteria NamespaceHealthCriteria) (models.NamespaceAppsHealth, error) {
+func (in *HealthService) GetNamespaceAppHealth(ctx context.Context, criteria NamespaceHealthCriteria) (models.NamespaceAppHealth, error) {
 	var end observability.EndFunc
 	ctx, end = observability.StartSpan(ctx, "GetNamespaceAppHealth",
 		observability.Attribute("package", "business"),
 		observability.Attribute("namespace", criteria.Namespace),
+		observability.Attribute("cluster", criteria.Cluster),
 		observability.Attribute("rateInterval", criteria.RateInterval),
 		observability.Attribute("queryTime", criteria.QueryTime),
 	)
 	defer end()
 
-	namespaceApps := models.NamespaceAppsHealth{}
-	for cluster := range in.userClients {
-		appEntities, err := fetchNamespaceApps(ctx, in.businessLayer, criteria.Namespace, cluster, "")
-
-		// If the cluster is the home cluster and we have just one cluster (single cluster mode), we want to return an error if we can't fetch the apps
-		if err != nil && cluster == kubernetes.HomeClusterName && len(in.userClients) == 1 {
-			return nil, err
-		}
-		if err != nil {
-			log.Debugf("Error fetching Applications for cluster %s: %s", cluster, err)
-			continue
-		}
-
-		namespaceAppsCluster, err := in.getNamespaceAppHealth(appEntities, criteria, cluster)
-		if err != nil {
-			return nil, err
-		}
-
-		namespaceApps = append(namespaceApps, namespaceAppsCluster...)
+	appEntities, err := fetchNamespaceApps(ctx, in.businessLayer, criteria.Namespace, criteria.Cluster, "")
+	if err != nil {
+		return nil, err
 	}
 
-	return namespaceApps, nil
+	return in.getNamespaceAppHealth(appEntities, criteria)
 }
 
-func (in *HealthService) getNamespaceAppHealth(appEntities namespaceApps, criteria NamespaceHealthCriteria, cluster string) (models.NamespaceAppsHealth, error) {
+func (in *HealthService) getNamespaceAppHealth(appEntities namespaceApps, criteria NamespaceHealthCriteria) (models.NamespaceAppHealth, error) {
 	namespace := criteria.Namespace
 	queryTime := criteria.QueryTime
 	rateInterval := criteria.RateInterval
-	allHealth := models.NamespaceAppsHealth{}
+	cluster := criteria.Cluster
+	allHealth := make(models.NamespaceAppHealth)
 
 	// Perf: do not bother fetching request rate if no workloads or no workload has sidecar
 	sidecarPresent := false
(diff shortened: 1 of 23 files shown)


Xunzhuo:
Can we link to the specific issue?

leandroberetta:
@Xunzhuo hi, sure, the isue is https://github.com/kiali/kiali/issues/6028. 

Remember this is a draft PR and is not ready for review yet, for that reason there is no description yet.

hhovsepy:
Did run locally on multiclster and on single OCP cluster.

business/health.go
```diff
@@ -114,7 +117,7 @@ func (in *HealthService) GetWorkloadHealth(ctx context.Context, namespace, workl
 }
 
 // GetNamespaceAppHealth returns a health for all apps in given Namespace (thus, it fetches data from K8S and Prometheus)
-func (in *HealthService) GetNamespaceAppHealth(ctx context.Context, criteria NamespaceHealthCriteria) (models.NamespaceAppHealth, error) {
+func (in *HealthService) GetNamespaceAppHealth(ctx context.Context, criteria NamespaceHealthCriteria) (models.NamespaceAppsHealth, error) {
```
nrfox, line 120: I know we're not doing this on other pages because those are driven off the aggregated namespace selector but for the overview page and the namespace health, can you just send cluster as a query param through `NamespaceHealthCriteria`. Then, I think, we can keep all the models the same `models.NamespaceAppHealth` vs. `models.NamespaceAppsHealth` and we wouldn't need to loop over all the clusters. For single cluster, since cluster is a query param, if it's not provided or `""` then we can assume you want the home cluster.

business/health.go
```diff
@@ -189,11 +206,6 @@ func (in *HealthService) GetNamespaceServiceHealth(ctx context.Context, criteria
 
 	var services *models.ServiceList
 	var err error
-	// Check if user has access to the namespace (RBAC) in cache scenarios and/or
-	// if namespace is accessible from Kiali (Deployment.AccessibleNamespaces)
-	if _, err := in.businessLayer.Namespace.GetNamespace(ctx, namespace); err != nil {
-		return nil, err
-	}
```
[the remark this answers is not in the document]
leandroberetta: I don't think so as then it is check when getting the services list.

business/health.go
leandroberetta: I thought of that before, it can be done I guess. 

When I decided to go this path, was to be consistent with the same logic we are using in the list pages (this is not a list page but it's similar, you have namespaces and health for n applications, so we are listing health among other things, just that the UI show them in a different aggregated way). There is no namespace selector because we are usually fetching all the aggregated namespaces.

I need to check if there is always the cluster information available in the UI, I guess so, because I added the field to the Namespace model, but I need to check in terms of the state, if I won't have a problem there. 

Having said this, maybe we need to do this (cluster as query param) in the lists too?.

nrfox: > Having said this, maybe we need to do this (cluster as query param) in the lists too?.

Yes we should but it's more of a pain there. Those pages make api requests per namespace based off of the namespace selector which currently aggregates all the namespaces across all the clusters. Since they're aggregated you lose the cluster info that goes along with the namespace object. In addition, those pages sometimes save their state in the URL where namespace is only saved as its name. We should update those to pass cluster as a query param but it's a bigger lift right now.

With the overview page, I don't think you have those problems. It just gets the namespaces from the `/api/namespaces` api call instead of the namespace selector which should return a `Namespace` object with the cluster since `/api/namespaces` returns the namespaces for every cluster. Also I don't think there's state saved in the URL for the namespaces page. It would be good to be consistent across pages but imo we should consistently pass cluster as a query param rather than trying to talk to every cluster on the backend for each API call.

leandroberetta: I also checked in the appenders, because the namespace health is used there too, and we should have the cluster name because the Node object has it according to:

https://github.com/kiali/kiali/commit/6ec37a53ddbcc88ee55959aeeac7114b33f893aa

So, I will refactor this PR to use the cluster as the query param, no problem.

leandroberetta: Hi @nrfox, I refactored the code to use the cluter as query parameter.

hhovsepy:
Should the overview card contain cluster name?

leandroberetta:
> Should the overview card contain cluster name?

No, that is something I'm doing as part of: https://github.com/kiali/kiali/issues/5921

frontend/src/services/Api.ts
```diff
@@ -328,7 +329,11 @@ export const getNamespaceAppHealth = (
   if (queryTime) {
     params.queryTime = String(queryTime);
   }
-  return newRequest<NamespaceAppHealth>(HTTP_VERBS.GET, urls.namespaceHealth(namespace), params, {}).then(response => {
+  var nsHealthUrl = urls.namespaceHealth(namespace);
+  if (cluster !== '') {
+    nsHealthUrl = nsHealthUrl + '?cluster=' + cluster;
+  }
```
nrfox, line 335: nit: this might break if other params are passed. Instead of modifying the URL you can add cluster to `params` and it'll get handled correctly.

frontend/src/services/Api.ts
```diff
@@ -355,23 +361,26 @@ export const getNamespaceServiceHealth = (
...
-      Object.keys(response.data).forEach(k => {
-        ret[k] = ServiceHealth.fromJson(namespace, k, response.data[k], {
-          rateInterval: duration,
-          hasSidecar: true,
-          hasAmbient: false
-        });
+  var nsHealthUrl = urls.namespaceHealth(namespace);
+  if (cluster !== '') {
+    nsHealthUrl = nsHealthUrl + '?cluster=' + cluster;
+  }
```
nrfox, line 367: Same comment about params

frontend/src/services/Api.ts
```diff
@@ -384,19 +393,21 @@ export const getNamespaceWorkloadHealth = (
...
-      Object.keys(response.data).forEach(k => {
-        ret[k] = WorkloadHealth.fromJson(namespace, k, response.data[k], {
-          rateInterval: duration,
-          hasSidecar: true,
-          hasAmbient: false
-        });
+  var nsHealthUrl = urls.namespaceHealth(namespace);
+  if (cluster !== '') {
+    nsHealthUrl = nsHealthUrl + '?cluster=' + cluster;
+  }
```
nrfox, line 399: Same comment about params

graph/telemetry/istio/appender/health.go
```diff
@@ -233,29 +236,29 @@ func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *g
 		for namespace, req := range healthReqs {
 			if req.app {
 				wg.Add(1)
-				go func(ctx context.Context, namespace string) {
+				go func(ctx context.Context, namespace, cluster string) {
 					defer wg.Done()
-					h, err := bs.Health.GetNamespaceAppHealth(ctx, business.NamespaceHealthCriteria{Namespace: namespace, IncludeMetrics: false})
+					h, err := bs.Health.GetNamespaceAppHealth(ctx, business.NamespaceHealthCriteria{Namespace: namespace, Cluster: cluster, IncludeMetrics: false})
 					resultsCh <- result{appNSHealth: h, namespace: namespace, err: err}
```
nrfox, line 242: Do you also need to add cluster to the `resultsCh`? Based off the comment it's key'd off `namespace+name` so if you had two `details.bookinfo` in two different clusters results from one may override the other.

nrfox (commented):
A few minor comments otherwise LGTM

graph/telemetry/istio/appender/health.go
leandroberetta: Yeah, good catch, thanks!

leandroberetta: Yeah, good catch. It is required, and I think I will need to add the cluster in the key too... I need to review this.

leandroberetta: Or change to something like `map[namespace][cluster][name]`....

frontend/src/services/Api.ts
leandroberetta: Sure, it's true... my bad

leandroberetta:
> A few minor comments otherwise LGTM

All issues addressed.

nrfox (approved):

josunect (approved):
Nice work, @leandroberetta !

Tested and LGMT

image

In the actual version we can see the health just from the local cluster: 

image

PR #6045, opened 2023-04-19
For apps list return error just if we are in single cluster

### LINKED ISSUE #6044: Error fetching app health when a namespace only exist in the remote clusters
image

This is expected, so the app service should check if we are in single cluster and throw the message just in that case, to keep compatibility

josunect:
Fixes #6044


hhovsepy (approved):
lgtm

Issue #5989, opened 2023-04-04
Missing horizontal scrollbar when seeing logs


im-Amitto:
Kiali Info: 
<img width="601" alt="Screenshot 2023-04-04 at 5 44 51 PM" src="https://user-images.githubusercontent.com/29653282/229788431-72814d65-94e1-4e77-a8c4-cbb5ccd4d6d3.png">

There is no horizontal scrollbar when checking workload logs so there is no way to see long logs

<img width="601" alt="Screenshot 2023-04-04 at 5 45 00 PM" src="https://user-images.githubusercontent.com/29653282/229788796-0a7ddaf9-947a-402f-b23b-252afc78c195.png">

im-Amitto:
FYI:
If you have more than a page of logs in current page then the horizontal bar does appear as well along with vertical one.

hhovsepy:
Thanks for logging @im-Amitto , I confirm the issue.

jshaughn:
That;s definitely some sort of regression. @im-Amitto, can you confirm your Browser type and version?

im-Amitto:
I forgot to drop this last time but i think the culprit is the `hidden` overflow on `x-axis` for the log grid.
```
<div aria-label="grid" aria-readonly="true" class="ReactVirtualized__Grid ReactVirtualized__List" role="grid" tabindex="0" style="box-sizing: border-box; direction: ltr; height: 10px; position: relative; width: 1209px; will-change: transform; overflow: hidden auto;">
```

im-Amitto:
@hhovsepy @jshaughn 
I was just checking out the codebase and realised that `List` from `react-virtualized` is being used to render logs which is causing this issue. `List` component does not have a `horizontal scrollbar` and is a implementation of `Grid` component which actually have the scrollbar we need.

One thing we can do is simply replace the `List` Component with a `Grid` Component.

Let me know if i should raise a PR for this.

hhovsepy:
@im-Amitto thank you for the research, feel free to send a PR with the solution, just in this case assign this issue to you.

PR #6048, opened 2023-04-20
Adjust scroll for Workload Pod Logs

josunect:
Fixes https://github.com/kiali/kiali/issues/5989 

I was able to reproduce this issue in Firefox (111.0.1), but not in Chrome (112.0.5615.49). 
Firefox view: 

image

Chrome: 

image

I've tried to use the Grid (https://github.com/bvaughn/react-virtualized/blob/master/docs/Grid.md) instead of the List, but I had the same issue in Firefox. 

The Container style has the following: 

`overflow: auto`

An for firefox: 

`overflow: hidden auto`

Setting the overflow to auto seems to do the trick: 

image


hhovsepy (approved):
Verified.

jshaughn (approved):
I see the same behavior, working on Chrome (whish is why we probably never noticed) and not working on FF (102.8.0esr).  WIth change working on both!

im-Amitto:
Thanks @jshaughn for the fix.
Sorry for the hiatus, was stuck with a couple of things.

Issue #6056, opened 2023-04-24
Multicluster - Config validation from remote cluster


hhovsepy:
Currently Istio Config validation and references are done for Home cluster only.
Extend it to support remote cluster configs as well.
This requires refactoring in retrieving Workloads per cluster.

Issue #5967, opened 2023-03-29
Multicluster - Istio config details view


nrfox:

hhovsepy:
Considering this subtask as done in a scope of List Configs task.

PR #6058, opened 2023-04-25
[DNM] Multicluster - Istio Config validations

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5967
Subtask: https://github.com/kiali/kiali/issues/6056

Istio config list and details shows validations from remote clusters.

Added missing cluster param in several places, put default values for metrics.

One issue I noticed, when opening config pages, then going to Overview page, there were failures in loading metrics, and not all namespaces are loaded. 
For this I did put 'cluster' param's value to home cluster in some places with TODO as I added more subtasks in https://github.com/kiali/kiali/issues/5618
Also disabled loading namespaces from cache, because it was loading only home cluster cache after some point.

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -364,7 +364,7 @@ func (in *AppService) fetchNamespaceApps(ctx context.Context, namespace string, 
 			IncludeOnlyDefinitions: true,
 			ServiceSelector:        labels.Set(w.Labels).String(),
 		}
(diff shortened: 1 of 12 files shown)


hhovsepy:
Moving this PR outside of the project sprint, as the task goes to 'beta' version of multicluster support.

Issue #5964, opened 2023-03-29
Multicluster - Services details view


nrfox:
TODO: What is missing is to propagate the cluster name in the tabs. This is always using the local cluster.

PR #6066, opened 2023-04-26
Services details view

josunect:
Fixes https://github.com/kiali/kiali/issues/5964

### DIFF
--- a/frontend/src/pages/Overview/OverviewPage.tsx
+++ b/frontend/src/pages/Overview/OverviewPage.tsx
@@ -495,8 +495,8 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
       nss.push(ns.name);
     });
 
-    // @TODO add cluster
-    return Promise.all([API.getConfigValidations(nss), API.getAllIstioConfigs('', nss, [], false, '', '')])
+    // TODO: Do we need the cluster here?
+    return Promise.all([API.getConfigValidations(nss), API.getAllIstioConfigs(nss, [], false, '', '')])
       .then(results => {
         chunk.forEach(nsInfo => {
           nsInfo.validations = results[0].data[nsInfo.name];
--- a/business/jaeger.go
+++ b/business/jaeger.go
@@ -9,7 +9,6 @@ import (
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/jaeger"
 	jaegerModels "github.com/kiali/kiali/jaeger/model/json"
-	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/models"
 	"github.com/kiali/kiali/observability"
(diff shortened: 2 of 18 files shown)


frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -495,8 +495,8 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
       nss.push(ns.name);
     });
 
-    // @TODO add cluster
-    return Promise.all([API.getConfigValidations(nss), API.getAllIstioConfigs('', nss, [], false, '', '')])
+    // TODO: Do we need the cluster here?
```
hhovsepy, line 498: We need a cluster here, Istio configs loading require `cluster` param, otherwise it will take home cluster as default.

hhovsepy:
@josunect thanks for the changes, looks good to me,
Tried locally, I only see one `metrics` request without cluster param, but cannot find when it happens:
Screenshot from 2023-04-27 10-32-58

frontend/src/pages/Overview/OverviewPage.tsx
josunect: Right. I though it was a "bulk" call... The problem I see is, the chunk is an array of mixed namespaces, so then, we should probably make a loop to make a different call per namespace. Now we are sending a call per namespace but different clusters, so it is not possible to associate them like the call it is right now, because we are also using just the namespace name, and not the whole namespace structure:

`http://localhost:3000/api/namespaces/metallb-system/metrics?filters[]=request_count&filters[]=request_error_count&duration=1800&step=180&rateInterval=180s&direction=inbound&reporter=destination`

hhovsepy: I think we need a separate sub-task for this, s it is out of scope of this PR.

josunect:
> @josunect thanks for the changes, looks good to me, Tried locally, I only see one `metrics` request without cluster param, but cannot find when it happens: Screenshot from 2023-04-27 10-32-58

Thanks, @hhovsepy 
I think it only happens in the local cluster, right? Because the remote cluster traces are still not loading. This is something that probably require a separate issue - see comment also for workloads details https://github.com/kiali/kiali/pull/6049 - I think it is not trivial to do, because I think the trace details would probably need to include the cluster information, at the moment it is not available

frontend/src/pages/Overview/OverviewPage.tsx
josunect: https://github.com/kiali/kiali/issues/6070

hhovsepy:
> I think it only happens in the local cluster, right? Because the remote cluster traces are still not loading. This is something that probably require a separate issue - see comment also for workloads details #6049 - I think it is not trivial to do, because I think the trace details would probably need to include the cluster information, at the moment it is not available
It happens to remote cluster 'west' as well, But I agree to leave it for different issue.

hhovsepy (approved):
Lgtm, tested locally, thanks @josunect

Issue #5868, opened 2023-02-22
[Cypress] Test procedure: Add validation references to K8S Gateways validations


matejnesuta:
The reason for creating this issue is to not lose this test procedure, which might be worth automating using Cypress.

Pre-requisites: 

Install Gateway API CRDs:

```
kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
  { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.1" | kubectl apply -f -; }
```

Install bookinfo demo on bookinfo namespace

Then, create the following K8s Gateway objects that has some colliding hostnames and host port combinations:

```
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gatewayapi-1
  labels:
    app: gatewayapi
spec:
  gatewayClassName: istio
  listeners:
  - name: default
    hostname: "bookinfo-istio-system.apps.ocp4-kqe1.maistra.upshift.redhat.com"
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
  addresses:
    - type: Hostname
      value: "google.com"
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gatewayapi-2
  labels:
    app: gatewayapi
spec:
  gatewayClassName: istio
  listeners:
  - name: default
    hostname: "bookinfo-istio-system.apps.ocp4-kqe1.maistra.upshift.redhat.com"
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
  addresses:
    - type: Hostname
      value: "google.com"
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: Gateway
metadata:
  name: gatewayapi-3
  labels:
    app: gatewayapi
spec:
  gatewayClassName: istio
  listeners:
  - name: default
    hostname: "bookinfo-istio-system.apps.ocp4-kqe1.maistra.upshift.redhat.com"
    port: 80
    protocol: HTTP
    allowedRoutes:
      namespaces:
        from: All
  addresses:
    - type: Hostname
      value: "google.com"
```

Expect to see the same validations references as shown in the PR description.

_Originally posted by @leandroberetta in https://github.com/kiali/kiali/issues/5853#issuecomment-1433575712_

PR #5853, opened 2023-02-16
Add validation references to K8S Gateways validations

leandroberetta:
Resolves https://github.com/kiali/kiali/issues/5769

image

PR #6067, opened 2023-04-26
Automation of a test procedure related to multiple K8 Gateways with same ports and hosts at the same time

matejnesuta:
This PR is an automation for #5868. The PR contains 2 tests. One of them is checking if the validations are present when 2 K8s Gateways with same ports and hosts are created. The other test then deletes one of the gateways and checks that the validations are no longer present. 

Related to #5853.

(works for the upstream OCP, not sure about the CI yet)


hhovsepy (approved):
lgtm, a good scenario, thanks @matejnesuta

ScriptingShrimp:
tested on OCP, well done

ScriptingShrimp (commented):
lgtm

PR #6069, opened 2023-04-27
Added missing cluster params

hhovsepy:
Follow-up PR for fixing some missing `cluster` params.


business/apps.go
```diff
@@ -364,7 +364,7 @@ func (in *AppService) fetchNamespaceApps(ctx context.Context, namespace string,
 			IncludeOnlyDefinitions: true,
 			ServiceSelector:        labels.Set(w.Labels).String(),
 		}
-		ss, err = in.businessLayer.Svc.GetServiceList(ctx, criteria)
+		ss, err = in.businessLayer.Svc.getServiceList(ctx, criteria, cluster)
```
josunect, line 367: Should this be a public method, then? Using upper case?

handlers/services.go
```diff
@@ -111,7 +111,7 @@ func ServiceDetails(w http.ResponseWriter, r *http.Request) {
 	namespace := params["namespace"]
 	service := params["service"]
 	queryTime := util.Clock.Now()
-	rateInterval, err = adjustRateInterval(r.Context(), business, namespace, rateInterval, queryTime)
+	rateInterval, err = adjustRateInterval(r.Context(), business, cluster, namespace, rateInterval, queryTime)
```
josunect, line 114: Should we use the cluster here? 
When we choose a list of namespaces from different cluster, I receive a 404 error: 

image

josunect (changes requested):
Thanks for completing the cluster parameters calls, @hhovsepy ! 

I have a couple of comments, otherwise LGTM

handlers/services.go
hhovsepy: I see the issue now. Agree, this was wrong asumption, because with empty cluster it looks namespaces through all clusters, reverting this.

business/apps.go
hhovsepy: agree

leandroberetta (approved):
Looks good.

josunect (approved):
LGMT! Thanks, @hhovsepy 

The issue is not reproducible anymore: 

image

PR #6074, opened 2023-04-27
[1.57] OSSM-3710: adding make targets to build and push kiali int tests image

FilipB:


jshaughn (approved):
LGTM, @jmazzitelli please take a look when you get the chance.  This will be evaluated on the 1.57 branch.  When that is all good we may "forwardport".

PR #6075, opened 2023-04-27
[v1.57] We also need hack scripts and 'which' package for bookinfo install

FilipB:
We will use also hack scripts to install bookinfo via this image. Also 'which' package was missing and it is needed for the bookinfo install script.


aljesusg (approved):
LGFM

PR #6078, opened 2023-04-28
removing flaky login test [OSSM-3772]

ScriptingShrimp:
just removing one annoying test that fails on OCP all the time

this will be reworked later to comply with interop pipeline, where it fails all the time


jshaughn (approved):
No problem for me.

PR #6079, opened 2023-04-28
[1.57] removing flaky login test [OSSM-3772]

ScriptingShrimp:
same as https://github.com/kiali/kiali/pull/6078


jshaughn (approved):
No problem.  Do we also need a backport to v1.65 for 2.4 testing?

Issue #6084, opened 2023-04-29
Crash in graph appender due to excluded namespace


jshaughn:
There is an occasional crash occuring in Kiali's graph generation.  The stack will show something like:

```
"error":"Namespace [kube-system] is excluded for Kiali",
"detail":"goroutine 631588 [running]:
runtime/debug.Stack()
  /home/jshaughn/App/go/src/runtime/debug/stack.go:24 +0x65
github.com/kiali/kiali/handlers.handlePanic({0x209edc0, 0xc0022fa330})
  /home/jshaughn/go/src/github.com/kiali/kiali/handlers/graph.go:86 +0x219
panic({0x19fe3c0, 0xc00286b9e0})
  /home/jshaughn/App/go/src/runtime/panic.go:884 +0x212
github.com/kiali/kiali/graph.CheckError(...)
  /home/jshaughn/go/src/github.com/kiali/kiali/graph/util.go:38
github.com/kiali/kiali/graph/telemetry/istio/appender.ServiceEntryAppender.getServiceEntry({0xc0019ca5a0?, {0xc0018b2e20?, 0x1dc6139?}}, {0xc0018b2dd4, 0xd}, {0xc001540ec9, 0x7}, 0xc00209ab00)
  /home/jshaughn/go/src/github.com/kiali/kiali/graph/telemetry/istio/appender/service_entry.go:173 +0xefb
github.com/kiali/kiali/graph/telemetry/istio/appender.ServiceEntryAppender.applyServiceEntries({0xc0019ca5a0?, {0xc0018b2e20?, 0xc000550d30?}}, 0x40d927?, 0xc00209ab00, 0xc0003ce2a0)
  /home/jshaughn/go/src/github.com/kiali/kiali/graph/telemetry/istio/appender/service_entry.go:94 +0x1e5
github.com/kiali/kiali/graph/telemetry/istio/appender.ServiceEntryAppender.AppendGraph({0xc0019ca5a0?, {0xc0018b2e20?, 0xc001a28e10?}}, 0xc0019cab40?, 0xc00209ab00?, 0xc001210390?)
  /home/jshaughn/go/src/github.com/kiali/kiali/graph/telemetry/istio/appender/service_entry.go:73 +0x117
[...]

PR #6085, opened 2023-04-29
Prevent crash scenario in namespace management

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6084

Prevent scenario where a token's namespaces cache can be set to the unfiltered namespaces for a single cluster.

I think this problem emerged gradually, due to a couple of things.

1) We introduced business.namespaces.go#GetNamespacesByCluster()

This method supports GetNamespaces() and is called for each cluster such that the parent can get the cross-cluster list.  I think there were two problems here.  First, it was declared public but really should have been private because it really should only be consumed by GetNamespaces to be launched as a gofunc.  Second, and more importantly, it sets the token's namespace cache to the partial, unfiltered list of namespaces for one cluster.  I'm pretty sure this was just a mistake.  Although unnecessary it's mainly harmless unless this routine is called from outside of GetNamespaces(), which is expecting to be, and should be, the only setter of the cache (I think).

Note that this method was made private by Nick along the way, but still maintained the cache setting.

2) We made it public again and called it from outside of GetNamespaces()

This was the danger scenario.  When called from outside of GetNamespaces the token's cache can get set to the unfiltered set of namespaces for just one cluster.  Given unlucky timing that wrong set of namespaces can then be provided to callers of GetNamespaces.

The PR again makes this method private, removes the unnecessary and dangerous cache setting, and offers a new public method for getting the namespaces of a single cluster.

I'm asking for a lot of reviewers because I'm not originally familiar with this code.

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -175,7 +175,7 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 			wg.Add(1)
 			go func(c string) {
 				defer wg.Done()
-				list, error := in.GetNamespacesByCluster(c)
+				list, error := in.getNamespacesByCluster(c)
 				if error != nil {
 					resultsCh <- result{cluster: c, ns: nil, err: error}
 				} else {
@@ -264,6 +264,7 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 		}
 	}
 
+	// store only the filtered set of namespaces in cache for the token
 	if kialiCache != nil && in.homeClusterUserClient != nil {
 		// just get the home cluster token because it is assumed tokens are identical across all clusters
 		kialiCache.SetNamespaces(in.homeClusterUserClient.GetToken(), resultns)
@@ -272,7 +273,7 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 	return resultns, nil
 }
 
-func (in *NamespaceService) GetNamespacesByCluster(cluster string) ([]models.Namespace, error) {
+func (in *NamespaceService) getNamespacesByCluster(cluster string) ([]models.Namespace, error) {
 	configObject := config.Get()
 
 	labelSelectorInclude := configObject.API.Namespaces.LabelSelectorInclude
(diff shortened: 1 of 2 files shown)


jshaughn:
I think the resulting fix should be backported to v1.67 because that version has the bug and is tied to Istio 1.18.

business/namespaces.go
```diff
@@ -391,14 +392,24 @@ func (in *NamespaceService) GetNamespacesByCluster(cluster string) ([]models.Nam
 		}
 	}
 
-	result := namespaces
+	return namespaces, nil
+}
 
-	if kialiCache != nil && in.homeClusterUserClient != nil {
-		// Set namespace by user token
-		kialiCache.SetNamespaces(in.homeClusterUserClient.GetToken(), result)
```
jmazzitelli, line 398: I see it - this is what was wrong. We shouldn't update the cache here - this list of namespaces is incorrect because it isn't filtered.

jmazzitelli (approved):
I didn't run this (since it sounds like its random and hard to reproduce, probably wouldn't help anyway) but I did look at the code and it makes sense to me.

If there is a fool-proof way to reproduce the error, I can try to actually run it, but from the sounds of it, there isn't a way. Perhaps we could write a unit test to make sure GetNamespaces() never screws up the cache???  I dunno.

business/namespaces.go
jshaughn: The only way I found to make it happen fairly consistently was to drill into node graphs, usually for service nodes.  When we drilled in we ended up making a call for istio config, which caused that sideways update, and when the timing was bad... boom.

josunect:
This is the error that I've got in multicluster, but I was not able to reproduce it in single cluster, so I've thought this was just affecting multi cluster.  

I've managed to reproduced with these steps (In multicluster): 

- Select namespaces bookinfo
- Select service graph
- Create "Request routing"
- Once the graph is reloaded, I'm able to reproduce the issue

I think it is a good catch, testing with the code of this PR I cannot reproduce the issue.

josunect (approved):

hhovsepy (approved):
Code looks good, will test after merge when my env is fixed.

hhovsepy:
I still see an error 500 in 'api/namespaces/kiali-access-ns/validations' response "{"error":"namespaces \"kiali-access-ns\" not found"}" while opening Graph page.

hhovsepy:
The error above is because 'cluster' param is not set in request. So this will be fixed in another PR.

PR #6086, opened 2023-04-30
[make] fix olm-operator-create target so it works again.

jmazzitelli:
The CSV was changed last year but we didn't realize it broke the make target to test on OLM. See: https://github.com/kiali/kiali-operator/pull/577


jshaughn (approved):
LOK

Issue #5966, opened 2023-03-29
Multicluster - Applications details view


nrfox:
TODO: What is missing it propagate the cluster name in the tabs. This is always using the local cluster.

PR #6088, opened 2023-05-01
App details view multicluster

hhovsepy:
Subtask https://github.com/kiali/kiali/issues/5966

### DIFF
--- a/frontend/src/pages/AppDetails/AppDetailsPage.tsx
+++ b/frontend/src/pages/AppDetails/AppDetailsPage.tsx
@@ -150,12 +150,7 @@ class AppDetails extends React.Component<AppDetailsProps, AppDetailsState> {
   private staticTabs() {
     const overTab = (
       <Tab title="Overview" eventKey={0} key={'Overview'}>
-        <AppInfo
-          app={this.state.app}
-          duration={this.props.duration}
-          health={this.state.health}
-          cluster={this.state.cluster ? this.state.cluster : ''}
-        />
+        <AppInfo app={this.state.app} duration={this.props.duration} health={this.state.health} />
       </Tab>
     );
 
--- a/frontend/src/pages/AppDetails/AppDetailsPage.tsx
+++ b/frontend/src/pages/AppDetails/AppDetailsPage.tsx
@@ -11,7 +11,7 @@ import * as AlertUtils from '../../utils/AlertUtils';
 import IstioMetricsContainer from '../../components/Metrics/IstioMetrics';
 import { MetricsObjectTypes } from '../../types/Metrics';
 import CustomMetricsContainer from '../../components/Metrics/CustomMetrics';
-import { DurationInSeconds, HomeClusterName, TimeInMilliseconds, TimeRange } from '../../types/Common';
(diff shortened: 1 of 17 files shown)


frontend/src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -62,7 +62,7 @@ class AppDetails extends React.Component<AppDetailsProps, AppDetailsState> {
   constructor(props: AppDetailsProps) {
     super(props);
     const urlParams = new URLSearchParams(this.props.location.search);
-    const cluster = urlParams.get('cluster') || HomeClusterName;
+    const cluster = urlParams.get('cluster') || undefined;
```
leandroberetta, line 65: Shouldn't be the home cluster by default if the cluter query parameter is missing?.

hhovsepy: Originally the HomeClusterName in frontend was empty '' string and supposed to be removed,
we have it now by default in backend, when no cluster param is send, the backend takes the default conf.KubernetesConfig.ClusterName

frontend/src/pages/AppDetails/AppDetailsPage.tsx
```diff
@@ -150,7 +150,12 @@ class AppDetails extends React.Component<AppDetailsProps, AppDetailsState> {
   private staticTabs() {
     const overTab = (
       <Tab title="Overview" eventKey={0} key={'Overview'}>
-        <AppInfo app={this.state.app} duration={this.props.duration} health={this.state.health} />
+        <AppInfo
+          app={this.state.app}
+          duration={this.props.duration}
+          health={this.state.health}
+          cluster={this.state.cluster ? this.state.cluster : ''}
```
josunect, line 157: If cluster is part of app, probably the cluster parameter is not needed here?

frontend/src/pages/AppDetails/AppInfo.tsx
```diff
@@ -51,7 +52,12 @@ class AppInfo extends React.Component<AppInfoProps, AppInfoState> {
     if (!this.props.app) {
       return;
     }
-    this.graphDataSource.fetchForVersionedApp(this.props.duration, this.props.app.namespace.name, this.props.app.name);
+    this.graphDataSource.fetchForVersionedApp(
+      this.props.duration,
+      this.props.app.namespace.name,
+      this.props.app.name,
+      this.props.cluster
```
josunect, line 59: So it should change here: 

`this.props.app.cluster ? this.props.app.cluster : ''`

josunect (commented):
Tested and works as expected. I've just added a couple of comments

frontend/src/pages/AppDetails/AppDetailsPage.tsx
hhovsepy: could be optimized, it is done similar to workloads and services, then we need to sync them as well

leandroberetta (approved):
LGTM

josunect (approved):
Thanks, @hhovsepy.
I haven't tested again but the changes looks good.

PR #6090, opened 2023-05-02
removing flaky smoke test

ScriptingShrimp:
Removing flaky test case (flakes on OCP platform).

It just one of the first tests we made, does not have any real value.


jshaughn (approved):
Goodbye files.

Issue #6064, opened 2023-04-26
Multicluster - add cluster param for TLS


hhovsepy:
RFE: To add support for `cluster` param sending to backend for `tls` request.
For instance iin Overview page, all `/api/namespaces/{namespace}/tls` requests do not have `cluster` params.
![Screenshot from 2023-04-26 13-07-36](https://user-images.githubusercontent.com/604313/234557368-158ce88c-41f6-42ad-a8e7-c97d6ae7519f.png)

Issue #6063, opened 2023-04-26
Multicluster - Istio Configs duplicated


hhovsepy:
Setup a multicluster env.
Create an Istio config in one of clusters, for instance PeerAuth.

```
apiVersion: security.istio.io/v1beta1
kind: PeerAuthentication
metadata:
  name: default
  namespace: bookinfo
spec:
  mtls:
    mode: STRICT
``` 
The configs will show duplicated for both clusters.
The issue is coming from Istio Registry status, where both PeerAuth objects are returned for both clusters. https://github.com/kiali/kiali/blob/master/business/istio_config.go#L210

![Screenshot from 2023-04-26 11-39-05](https://user-images.githubusercontent.com/604313/234536456-69bee935-576c-40a9-bbae-ec30b3fa7671.png)

Reported by cc @josunect

PR #6092, opened 2023-05-02
Cluster support for mTLS

hhovsepy:
Subtask https://github.com/kiali/kiali/issues/6064

All 'namespace/tls' requests into backend are containing 'cluster' param.
Can be seen in Overview page:
Screenshot from 2023-05-03 13-18-16

mTLS request for the home cluster does not send 'cluster' param and the default home cluster name is chosen backend.

Consider this [logged issue](https://github.com/kiali/kiali/issues/6063) when verifying mTLS for namespace only, as the created PeerAuth will be shown for same named namespaces in all clusters, thus all same named namespaces will be considered as enabled mTLS.


josunect (commented):
LGTM, the cluster is applied as a parameter in the tls namespace call. 
image


The other comment is, mtls was applied just to cluster east, but the locker is in both. This is related to this issue? https://github.com/kiali/kiali/issues/6063 

image

Thanks!

hhovsepy:
> The other comment is, mtls was applied just to cluster east, but the locker is in both. This is related to this issue? #6063
Yes, this is known issue with registry,

josunect (approved):
LGTM!

hhovsepy:
Thank you @josunect

PR #6095, opened 2023-05-03
[v1.67] Prevent bad namespaces cache scenario (#6085)

jshaughn:
**Backport** of https://github.com/kiali/kiali/pull/6085

Prevent scenario where a token's namespaces cache can be set to the unfiltered namespaces for a single cluster.

I'm not 100% sure v1.67 suffers from a problem because the call from istio_config is not yet made, but this should ensure the problem does not happen in an unanticipated scenario.

https://github.com/kiali/kiali/issues/6084

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -264,6 +264,7 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 		}
 	}
 
+	// store only the filtered set of namespaces in cache for the token
 	if kialiCache != nil && in.homeClusterUserClient != nil {
 		// just get the home cluster token because it is assumed tokens are identical across all clusters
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):
lgtm - just a cherry-pick of the master PR

jmazzitelli:
Actually, there are 2 files changed in the master PR that this is based off of - is that right @jshaughn ?

jshaughn:
> Actually, there are 2 files changed in the master PR that this is based off of - is that right @jshaughn ?

Yes, that's correct.  Only the actual fix to the GetNamespaces logic is needed in 1.67.  I'm not 100% sure 1.67 suffers from the scenario that I hit, but there may be others, the core fix should still be good.  This is a better-safe-than-sorry patch for 1.67, because it is tied to Istio 1.18 release.

Issue #6097, opened 2023-05-03
[integration] Fix istio_config_test.TestIstioConfigs on OCP single cluster


hhovsepy:
Jenkins nightly job on Kiali master and OCP is failing:

```
{"level":"debug","time":"2023-05-03T02:08:35Z","message":"gateway.gateway.networking.k8s.io/gatewayapi created\nhttproute.gateway.networking.k8s.io/httproute created\n"}
    istio_config_test.go:33: 
    istio_config_test.go:34: 
{"level":"debug","time":"2023-05-03T02:08:35Z","message":"gateway.gateway.networking.k8s.io \"gatewayapi\" deleted\nhttproute.gateway.networking.k8s.io \"httproute\" deleted\n"}
panic: runtime error: invalid memory address or nil pointer dereference
/usr/local/go/src/testing/testing.go:1396 +0x24e
/usr/local/go/src/testing/testing.go:1399 +0x39f
/usr/local/go/src/runtime/panic.go:884 +0x212
/home/fedora/kiali/tests/integration/tests/istio_config_test.go:35 +0x1b0
/usr/local/go/src/testing/testing.go:1446 +0x10b
/usr/local/go/src/testing/testing.go:1493 +0x35f
```

Issue #6051, opened 2023-04-21
Multicluster - Remove kubernetes.HomeClusterName


hhovsepy:
Remove `kubernetes.HomeClusterName` from backend usage, replace by `config.Get().KubernetesConfig.ClusterName` where necessary.

Issue #6052, opened 2023-04-21
Multicluster - Remove GetIstioConfigList


hhovsepy:
Remove the usage of `GetIstioConfigList` and replace it by `GetIstioConfigListPerCluster` from places in backend where multicluster support is not fully implemented yet.

EDIT: Move cluster param into Criteria.

PR #6100, opened 2023-05-04
Fix integration test failure

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6097
Fixes https://github.com/kiali/kiali/issues/6051

Handle the fact that we currently need to equate "" with "Kubernetes" when querying for Namespaces for the home cluster.  This should change as we better handle the home cluster name.

---

@hhovsepy,  I'm not exactly sure why it affects the OCP intergration tests and not other situations, but the root cause is what we had talked about.  The client call 
is using the configured home cluster name, which defaults to "Kubernetes", but the cache is using "".  It probably has to do with our current home cluster handling; I can see 3 different ways of determining the home cluster name: `kubernetes.HomeClusterName`, `business.DefaultClusterID`, and `kialiConfig.Get().KubernetesConfig.ClusterName`.  

For now, I think this resolves the problems, but I think the config is probably the thing that should always be used for the hoce cluster (and it should default to "Kubernetes").

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -564,8 +564,16 @@ func (in *NamespaceService) GetNamespaceByCluster(ctx context.Context, namespace
 			if err2 != nil {
 				return nil, err2
 			}
+		} else {
+			if _, ok := in.userClients[cluster]; !ok {
+				return nil, fmt.Errorf("OCP Cluster [%s] is not found or is not accessible for Kiali", cluster)
+			}
+			project, errC := in.userClients[cluster].GetProject(namespace)
+			if errC != nil {
+				return nil, errC
+			}
+			result = models.CastProject(*project, cluster)
 		}
-
 	} else {
 		// TODO: MC
 		var ns *core_v1.Namespace
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -44,15 +44,16 @@ func IsAccessibleError(err error) bool {
 
 func NewNamespaceService(userClients map[string]kubernetes.ClientInterface, kialiSAClients map[string]kubernetes.ClientInterface) NamespaceService {
 	var hasProjects bool
+	conf := config.Get()
 
-	homeClusterName := config.Get().KubernetesConfig.ClusterName
+	homeClusterName := conf.KubernetesConfig.ClusterName
 	if saClient, ok := kialiSAClients[homeClusterName]; ok && saClient.IsOpenShift() {
 		hasProjects = true
 	} else {
 		hasProjects = false
 	}
 
-	ans := config.Get().Deployment.AccessibleNamespaces
+	ans := conf.Deployment.AccessibleNamespaces
 	isAccessibleNamespaces := make(map[string]bool, len(ans))
 	for _, ns := range ans {
 		isAccessibleNamespaces[ns] = true
@@ -81,12 +82,12 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 		}
 	}
 
-	configObject := config.Get()
+	conf := config.Get()
 
 	// determine what the discoverySelectors are by examining the Istio ConfigMap
 	var discoverySelectors []*meta_v1.LabelSelector
 	if kialiCache != nil {
-		if icm, err := kialiCache.GetConfigMap(configObject.IstioNamespace, configObject.ExternalServices.Istio.ConfigMapName); err == nil {
+		if icm, err := kialiCache.GetConfigMap(conf.IstioNamespace, conf.ExternalServices.Istio.ConfigMapName); err == nil {
 			if ic, err2 := kubernetes.GetIstioConfigMap(icm); err2 == nil {
 				discoverySelectors = ic.DiscoverySelectors
 			} else {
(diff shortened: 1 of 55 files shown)


business/namespaces.go
```diff
@@ -404,7 +404,11 @@ func (in *NamespaceService) GetNamespacesForCluster(ctx context.Context, cluster
 
 	clusterNamespaces := []models.Namespace{}
 	for _, ns := range tokenNamespaces {
-		if ns.Cluster == cluster {
+		// TODO: The RHS of this expression shouldn't be necessary. I don't think we shold have a DefaultClusterID
+		// or a HomeClusterName constant.  We should always be using the configured cluster name (which should be
+		// set the proper "Kubernetes" default as needed).  The RHS is here because there are times when the client
+		// is using one value and the server is using the other.
+		if ns.Cluster == cluster || (cluster == DefaultClusterID && ns.Cluster == "") {
```
hhovsepy, line 411: DefaultClusterID is hardcoded to "Kubernetes", but in fact "install-istio-via-istioctl.sh" can take any cluster name, and in QE env it is set "cluster-default", so the tests are still failing.
Suggestion to change 'DefaultClusterID' here to 'config.Get().KubernetesConfig.ClusterName'

hhovsepy: making changes as experimental in https://github.com/kiali/kiali/pull/6101 and checking on OCP cluster, now more failures, investigating

hhovsepy: @jshaughn if you make the changes then we can merge this PR, it blocks other PRs.

jshaughn: Yes, the tests were passing on my system because DefaultClusterID was correct, because I was using the default.  But the main issue here is bigger, we are going to get into trouble with these three different things going on.

jshaughn: I'm going to look at this a bit more.  I still think the solution is to get the proper home cluster name into the cache.  It should not be "".

jshaughn:
Hi @hhovsepy , So after diving into this some more I found the root cause of the failures on OCP.  It had to do with openshift projects not being assigned cluster information in the same way that we already had for vanilla namespaces.  I've updated this PR to remove my original band-aid fix and it now has the proper fix.  I went further, though, because I wanted to help make the namespace management a little more clean.  The `kubernetes.HomeClusterName` variable was deprecated and I think that was a good idea.  So I removed it in favor of relying on the config.  I found a variety of other small issues that I cleaned up along the way, some in the code and some in the tests.

I am pretty sure we want to merge this stuff in and layer any other changes on top of it.  Don't be scared by the number of files, most of them are just touched by the removal of HomeClusterName.  I'll be online tomorrow morning for a while, so we can discuss it further if necessary.

**Update** For some reason the CI is failing, but I'm not sure why, I think maybe there is actually a CI env issue because I'm seeing timeouts and things like that.  It's not failing for me locally.  Or maybe it now fails on non-OCP :-(  Anyway, please take a look...

hhovsepy:
Thank you Jay, I will take a look,
we had a separate task for optimizing kubernetes.HomeClusterName https://github.com/kiali/kiali/issues/6051, (I was postponing it's solution :) ) so this PR resolves that.

hhovsepy:
Graph page is failing to load, when `/graph` request does not contain `cluster` param, it returns '"Cluster [cluster-default] is not found or is not accessible for Kiali"' error response.
But `/graph?cluster=Kubernetes` requests works fine.

business/namespaces.go
```diff
@@ -556,15 +557,23 @@ func (in *NamespaceService) GetNamespaceByCluster(ctx context.Context, namespace
...
 				if err2 == nil {
-					result = models.CastProject(*project)
+					result = models.CastProject(*project, cluster)
 					break
 				}
 			}
 			if err2 != nil {
 				return nil, err2
 			}
+		} else {
```
hhovsepy, line 567: here is one of the key changes, the issue described here https://github.com/kiali/kiali/pull/6106  is resolved now

business/tls.go
```diff
@@ -44,13 +44,15 @@ func (in *TLSService) MeshWidemTLSStatus(ctx context.Context, namespaces []strin
 		IncludePeerAuthentications: true,
 		Cluster:                    cluster,
 	}
+	conf := config.Get()
 
-	istioConfigList, err := in.businessLayer.IstioConfig.GetIstioConfigListPerCluster(ctx, criteria, cluster)
+	// @TODO hardcoded HomeClusterName
+	istioConfigList, err := in.businessLayer.IstioConfig.GetIstioConfigListPerCluster(ctx, criteria, conf.KubernetesConfig.ClusterName)
```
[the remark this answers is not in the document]
jshaughn: I changed this as you suggested.  But I am wondering what is the difference between what is supplied in the `criteria.cluster` and `cluster` arguments?  Also, why wouldn't this `cluster` argument be in the criteria?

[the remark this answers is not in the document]
hhovsepy: Thank you Jay, it seems separate param remains because the Criteria.cluster was added later, but there is a task to cleanup this code sections: https://github.com/kiali/kiali/issues/6052

hhovsepy (approved):
Thank you for your work @jshaughn  
On OCP QE single cluster env there is only 1 failures (probably not related).
Tested on local multicluster.

Issue #6107, opened 2023-05-04
Cluster badges in details pages should be only visible in Multi cluster 


josunect:
The multi cluster check should be done using `import { isMultiCluster } from 'config';` and not if the details has any cluster property.

Now it is possible to see Kubernetes in details pages: 

![image](https://user-images.githubusercontent.com/49480155/236221711-7b964573-e1c3-45a7-85a6-613c4f4b95ca.png)

Ref. #5618

PR #6108, opened 2023-05-04
Check number of clusters to show the cluster badge in details pages

josunect:
Fixes https://github.com/kiali/kiali/issues/6107


hhovsepy:
@josunect can the check be done by 'isMultiCluster' method? `import { isMultiCluster } from 'config';`

josunect:
Thanks for the review, @hhovsepy. I tried to use that method first, but for some reason it was always returned true - I figured out that the data was not loaded for some reason. But, let me try it again.

josunect:
> Thanks for the review, @hhovsepy. I tried to use that method first, but for some reason it was always returned true - I figured out that the data was not loaded for some reason. But, let me try it again.

Maybe some issues in my env, it is working now. 
Thank you!

hhovsepy (approved):
@josunect thanks, verified on a single cluster.

PR #6109, opened 2023-05-04
Test objects cleanup related to the Istio Config Wizard

### LINKED ISSUE #6073: Fix tests in Istio Config Wizard failing on subsequent runs
This issue is related [this featureFile](https://github.com/kiali/kiali/blob/master/frontend/cypress/integration/featureFiles/wizard_istio_config.feature) only, as far as I know. The issue is that there are multiple gateway and service objects created and these do not get deleted after the test run. This does not affect the CI or the OCP pipelines, because the demoapps are usually deleted altogether right after the whole suite finishes. However when you don't reinstall the demoapps, half of the test might break and the only solution right now is to delete the objects manually, either in the Kiali UI or in the terminal. This is especially annoying when debugging/creating new tests locally.

Affected tests:
[Scenario: Create a K8s Gateway scenario](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L14)
[Scenario: Create a Gateway scenario](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L44)
[Scenario: Create a Gateway with TLS](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L124)
[Scenario: Create a ServiceEntry without ports specified](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L162)
[Scenario: Create a ServiceEntry with ports specified](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L188)
[Scenario: Create multiple K8s Gateways with colliding hostnames and port combinations and check for a reference](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L219)

Some of these tests only need a quick teardown, however there are also 2 tests, which are already dependent on these tests. These 2 tests are expecting the gateway object to be already created. They don't fail because of the issue already mentioned, but it would be nice to reworked them a little in the process as well. 

These are:
[Scenario: Create a Gateway with duplicate name](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L94), which has dependency on the "Create a Gateway scenario" test
[Scenario: Delete one of the K8s Gateways and check that the reference is removed](https://github.com/kiali/kiali/blob/53a52ea68bafcaef827b8e3b0d077fd0ca335429/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L251), which has dependency on the "Create multiple K8s Gateways" test

This means that most of the tests need a simple teardown, while the 2 tests with dependencies need more complex approach.

matejnesuta:
Fixes #6073.

I added one step before all affected tests to delete the object. 

I also merged the dependent scenarios together. Do we want it this way or is there a more suitable way? The gherkins are huge. @ScriptingShrimp

### DIFF
--- a/frontend/cypress/integration/common/wizard_istio_config.ts
+++ b/frontend/cypress/integration/common/wizard_istio_config.ts
@@ -20,6 +20,21 @@ When('viewing the detail for {string}', (object: string) => {
   cy.get('a').contains(object).should('be.visible').click();
 });
 
+When ('k8sgateway {string} is not created',(name:string) =>{
+  cy.exec(`kubectl delete gateways.gateway.networking.k8s.io ${name} -n bookinfo`, { failOnNonZeroExit: false });
+  ensureKialiFinishedLoading();
+});
+
+When ('gateway {string} is not created',(name:string) =>{
+  cy.exec(`kubectl delete gateway.networking.istio.io ${name} -n bookinfo`, { failOnNonZeroExit: false });
+  ensureKialiFinishedLoading();
+});
+
+When ('service {string} is not created',(name:string) =>{
+  cy.exec(`kubectl delete serviceEntries ${name} -n bookinfo`, { failOnNonZeroExit: false });
+  ensureKialiFinishedLoading();
+});
+
 And('user sees the {string} config wizard', (title: string) => {
   cy.get('h1').should('contain.text', title);
 });
--- a/frontend/cypress/integration/common/wizard_istio_config.ts
+++ b/frontend/cypress/integration/common/wizard_istio_config.ts
@@ -20,17 +20,18 @@ When('viewing the detail for {string}', (object: string) => {
   cy.get('a').contains(object).should('be.visible').click();
 });
 
-When ('k8sgateway {string} is not created',(name:string) =>{
+
+When ('user deletes k8sgateway named {string} and the resource is no longer available',(name:string) =>{
   cy.exec(`kubectl delete gateways.gateway.networking.k8s.io ${name} -n bookinfo`, { failOnNonZeroExit: false });
   ensureKialiFinishedLoading();
 });
 
-When ('gateway {string} is not created',(name:string) =>{
+When ('user deletes gateway named {string} and the resource is no longer available',(name:string) =>{
   cy.exec(`kubectl delete gateway.networking.istio.io ${name} -n bookinfo`, { failOnNonZeroExit: false });
   ensureKialiFinishedLoading();
 });
 
-When ('service {string} is not created',(name:string) =>{
+When ('user deletes service named {string} and the resource is no longer available',(name:string) =>{
   cy.exec(`kubectl delete serviceEntries ${name} -n bookinfo`, { failOnNonZeroExit: false });
   ensureKialiFinishedLoading();
 });
@@ -96,6 +97,10 @@ And('choosing to delete it', () => {
   cy.get('#actions').should('be.visible').click();
   cy.get('#actions').contains('Delete').should('be.visible').click();
   cy.get('#pf-modal-part-2').find('button').contains('Delete').should('be.visible').click();
+});
+
+And('user closes the success notification',()=>{
+  cy.get('[aria-label="Close Success alert: alert: Istio Gateway created"]').click();
 });
 
 Then('the {string} {string} should be listed in {string} namespace', function (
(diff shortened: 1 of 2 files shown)


ScriptingShrimp:
@matejnesuta please test this PR against upstream OCP. Meanwhile I'll try OSSM 2.4 so we can cherry pick it into release branch right away

frontend/cypress/integration/common/wizard_istio_config.ts
```diff
@@ -19,6 +19,20 @@ When('viewing the detail for {string}', (object: string) => {
   cy.get('a').contains(object).click();
 });
 
+When ('k8sgateway {string} is not created',(name:string) =>{
```
ScriptingShrimp, line 22: please adjust this Gherkin and step definition to something a bit more comprehensible

`
user deletes k8sgateway named {string} and resource no longer available
`

frontend/cypress/integration/common/wizard_istio_config.ts
```diff
@@ -19,6 +19,20 @@ When('viewing the detail for {string}', (object: string) => {
...
+  cy.exec(`kubectl delete gateways.gateway.networking.k8s.io ${name} -n bookinfo`, { failOnNonZeroExit: false });
+  ensureKialiFinishedLoading();
+});
+
+When ('gateway {string} is not created',(name:string) =>{
```
ScriptingShrimp, line 27: please adjust this Gherkin and step definition to something a bit more comprehensible

`user deletes gateway named {string} and resource no longer available`

frontend/cypress/integration/common/wizard_istio_config.ts
```diff
@@ -19,6 +19,20 @@ When('viewing the detail for {string}', (object: string) => {
...
+  cy.exec(`kubectl delete gateway.networking.istio.io ${name} -n bookinfo`, { failOnNonZeroExit: false });
+  ensureKialiFinishedLoading();
+});
+
+When ('service {string} is not created',(name:string) =>{
```
ScriptingShrimp, line 32: please adjust this Gherkin and step definition to something a bit more comprehensible

`user deletes service named {string} and resource no longer available`

frontend/cypress/integration/featureFiles/wizard_istio_config.feature
```diff
@@ -12,7 +12,8 @@ Feature: Kiali Istio Config page
 
   @wizard-istio-config
   Scenario: Create a K8s Gateway scenario
-    When user clicks in the "K8sGateway" Istio config actions
+    When k8sgateway "k8sapigateway" is not created
```
ScriptingShrimp, line 15: see above in wizard_istio_config.ts

frontend/cypress/integration/featureFiles/wizard_istio_config.feature
```diff
@@ -216,8 +220,9 @@ Feature: Kiali Istio Config page
     Then the preview button should be disabled
 
 @wizard-istio-config
-  Scenario: Create multiple K8s Gateways with colliding hostnames and port combinations and check for a reference
-    When user clicks in the "K8sGateway" Istio config actions
+  Scenario: Create multiple K8s Gateways with colliding hostnames and port combinations and check for a reference. Then delete one of them and the reference should be gone.
```
ScriptingShrimp, line 223: this scenario is big and might be a bit fragile. I have no problem with merging it right now. If will be stable, we can keep it, if not we will refactor that later.

ScriptingShrimp (commented):
please address changes and test this against upstream OCP if stable
Thank you

matejnesuta:
image

works in the upstream OCP

will adjust the gherkins

ScriptingShrimp (approved):
LGTM!

Issue #6114, opened 2023-05-08
Cache tries to list cluster scoped resource but it is namespace scoped


josunect:
There is a particular scenario where the Kiali cache is trying to list resources cluster scoped but the cache is namespace-scope. The errors in the log: 

```
2023-05-08T12:52:25Z DBG [Kiali Cache] Starting namespace-scoped for namespace: informers
2023-05-08T12:52:25Z INF [Kiali Cache] Waiting for namespace-scoped for namespace: cache to sync
W0508 12:52:25.429122 1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: failed to list *v1.StatefulSet: statefulsets.apps is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "statefulsets" in API group "apps" at the cluster scope
E0508 12:52:25.429173 1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: Failed to watch *v1.StatefulSet: failed to list *v1.StatefulSet: statefulsets.apps is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "statefulsets" in API group "apps" at the cluster scope
W0508 12:52:25.429135 1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: failed to list *v1beta1.WorkloadGroup: workloadgroups.networking.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "workloadgroups" in API group "networking.istio.io" at the cluster scope
E0508 12:52:25.429189 1 reflector.go:138] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: Failed to watch *v1beta1.WorkloadGroup: failed to list *v1beta1.WorkloadGroup: workloadgroups.networking.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "workloadgroups" in API group "networking.istio.io" at the cluster scope
W0508 12:52:25.429136 1 reflector.go:324] pkg/mod/k8s.io/client-go@v0.24.2/tools/cache/reflector.go:167: failed to list *v1.Service: services is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "services" in API group "" at the cluster scope
[...]

PR #6115, opened 2023-05-08
fix bug when doing cluster-scoped lookups when not authorized to do so [OSSM-3868]

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6114

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -173,7 +173,9 @@ func (in *IstioConfigService) GetIstioConfigListPerCluster(ctx context.Context, 
 	)
 	defer end()
 
-	if criteria.Namespace == "" && !criteria.AllNamespaces {
+	clusterWideAccess := config.Get().Deployment.ClusterWideAccess
+
+	if criteria.Namespace == "" && (!criteria.AllNamespaces || !clusterWideAccess) {
 		return models.IstioConfigList{}, errors.New("GetIstioConfigList needs a non empty Namespace")
 	}
 	istioConfigList := models.IstioConfigList{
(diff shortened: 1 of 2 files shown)


jmazzitelli:
This PR is now in a state where it looks like the errors are averted. However, you will now see things like this:

```
2023-05-08T13:44:25-04:00 ERR GetIstioConfigList needs a non empty Namespace
```

This is because the code is written in such a way that it doesn't look valid when we do not have cluster wide access permissions. When cluster_wide_access is false, it means we cannot query for resources unless a non-empty string namespace is specified - and in some cases that rule is violated.

jmazzitelli:
I would like a review of this. With `deployment.accessible_namespaces` set to something like `['istio-system']` (i.e. NOT `**`), and `deployment.cluster_wide_access` set to `false` (which is the default if AN is not `**`), and `external_services.istio.istio_api_enabled` set to `false` this will no longer spew out errors in the logs or the UI. It will log a debug message saying you are losing capabilities with that setup.

One thing I don't understand is why, when I go to a specific namespace's Istio Config page in the UI, it still wants to access "AllNamespaces" when querying resources.

josunect:
Thanks, @jmazzitelli ! I will have a look

josunect:
@jmazzitelli , I have applied these changes in this PR https://github.com/kiali/kiali/pull/6116/files to merge with master and test in OCP pipeline. I have to merge in order to have the no_istiod tests running, as the code before the last commit in master had no working istio config tests that prevented to run the no_istiod tests. 

I'm afraid the tests are still failing with not a lot of information about the error. But I do think your changes makes sense, it would probably be another call with that parameter. I think this change wouldn't be needed after this PR but I think it will be required: https://github.com/kiali/kiali/blob/master/kubernetes/cache/kube_cache.go#L358

josunect:
I'm re running the tests. I think this must work and there was just an error in the validations test. The validations should be available now with your changes.

josunect (approved):
Test no_istiod passed in OCP

PR #6119, opened 2023-05-09
Fix OCP tests

josunect:
Fixes https://github.com/kiali/kiali/issues/6121

There was a couple of failures: 

- **no_istiod**: It is not enough for wait for the pod to be ready, we need to wait to the old pod is terminated. Otherwise, doing a request to the application we receive an error `The application is currently not serving requests at this endpoint. It may not have been started or is still starting.`
-  **remote_istiod**: There was a couple of things: 
   - One of the calls was using wrong namespace (kialiNS vs KialiDeploymentNS)
   - It is safer to don't delete the pod it a Kiali CRD exists - The operator will handle it.  
   - Time out per test has to be increased

### DIFF
--- a/tests/integration/tests/no_istiod_test.go
+++ b/tests/integration/tests/no_istiod_test.go
@@ -12,7 +12,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
-	
+
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/tests/integration/utils"
 )
--- a/tests/integration/tests/no_istiod_test.go
+++ b/tests/integration/tests/no_istiod_test.go
@@ -1,13 +1,13 @@
 package tests
 
 import (
-	"os/exec"
-	"strconv"
-	"strings"
-	"testing"
-	"time"
 
-	"github.com/stretchr/testify/assert"
+"github.com/stretchr/testify/assert"
+"os/exec"
+"strconv"
+"strings"
+"testing""time"
+
 
 	"github.com/kiali/kiali/log"
 	"github.com/kiali/kiali/tests/integration/utils"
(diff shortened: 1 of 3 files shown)


josunect:
After PRs https://github.com/kiali/kiali/pull/6115 and https://github.com/kiali/kiali/pull/6100 there is still one test failure in OCP, TestLogsContainerIstioProxy. 
The error:

`The application is currently not serving requests at this endpoint. It may not have been started or is still starting.`

It might by because, when no_istiod test finalizes, it restarts the Kiali pod again to set istio_api_enabled to true. If this test is running after no_istiod it could be affected.

tests/integration/tests/no_istiod_test.go
```diff
@@ -1,24 +1,37 @@
...
+func update_istio_api_enabled(value bool, kubeClientSet kubernetes.Interface, ctx context.Context) {
 	original := !value
 
-	cmdGetProp := ocCommand + " get cm kiali -n istio-system -o yaml | grep 'istio_api_enabled'"
+	// Restart kiali pod
+	// Get kiali pod name
+	cmdGetPodName := ocCommand + " get pods -o name -n " + kialiNamespace + " | egrep kiali | sed 's|pod/||'"
+	kialiPodName, err2 := exec.Command("bash", "-c", cmdGetPodName).Output()
+	podName := strings.Replace(string(kialiPodName), "\n", "", -1)
+	log.Debugf("Kiali pod name: %s", podName)
```
nrfox, line 32: Why are we using shell commands to get/update kube objects? This wasn't part of your changes, I'm just curious.

tests/integration/tests/no_istiod_test.go
```diff
@@ -30,43 +43,59 @@ func update_istio_api_enabled(value bool) {
...
+					for _, condition := range pod.Status.Conditions {
+						if condition.Type == "Ready" && condition.Status == "False" {
+							log.Debugf("New kiali pod is not ready.")
+							log.Debugf("Condition type %s status %s pod name %s", condition.Type, condition.Status, pod.Name)
+							return false, nil
+						}
+					}
+				}
+				return true, nil
+			})
```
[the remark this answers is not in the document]
josunect: Yes, using the one from remote_istio.

tests/integration/tests/remote_istiod_test.go
```diff
@@ -115,10 +115,19 @@ func TestRemoteIstiod(t *testing.T) {
 	// Register clean up before creating resources in case of failure.
 	t.Cleanup(func() {
 		log.Debug("Cleaning up resources from RemoteIstiod test")
+
+		var currentKialiPod string
+		pods, err := kubeClient.CoreV1().Pods(kialiDeploymentNamespace).List(ctx, metav1.ListOptions{LabelSelector: "app=kiali"})
+		if err != nil {
+			log.Errorf("Error getting the pods list %s", err)
```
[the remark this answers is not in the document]
josunect: Yes, I have added the require No error. 
Regarding the pods length, it should not be just 1? In all the environments I've just seen 1 pod.

tests/integration/tests/no_istiod_test.go
josunect: Well, nothing specific, as we are not testing the kube client and the shell is also available, is other possibility.

josunect: But, is there is any issue I can change it.

nrfox: It'll be harder to check error output with running shell commands and harder to reuse code but save that for another change.

tests/integration/tests/remote_istiod_test.go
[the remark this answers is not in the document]
nrfox: In this case yes, but you could also have a generic
```
getKialiPods()
```
and it's probably fine for it to be more than 1 there.

nrfox (approved):
LGTM

Issue #6065, opened 2023-04-26
Multicluster - add cluster param for Metrics


hhovsepy:
RFE: Add support of `cluster` param in `api/namespaces/{naemspace}/metrics` query send to backend, for instance from Overview page:
![Screenshot from 2023-04-26 13-10-42](https://user-images.githubusercontent.com/604313/234558176-37ae9f7f-c29f-4666-b520-e3c4250f9794.png)

PR #6131, opened 2023-05-10
Multicluster - Add cluster params for metrics 

josunect:
Fix https://github.com/kiali/kiali/issues/6065

### DIFF
--- a/frontend/src/pages/Graph/SummaryPanelCommon.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelCommon.tsx
@@ -121,7 +121,7 @@ export const getNodeMetrics = (
     case NodeMetricType.APP:
       return API.getAppMetrics(nodeData.namespace, nodeData.app!, options, nodeData.cluster);
     case NodeMetricType.SERVICE:
-      return API.getServiceMetrics(nodeData.namespace, nodeData.service!, options, nodeData.cluster);
+      return API.getServiceMetrics(nodeData.namespace, nodeData.service!, options);
     case NodeMetricType.WORKLOAD:
       return API.getWorkloadMetrics(nodeData.namespace, nodeData.workload!, options, nodeData.cluster);
     default:
--- a/frontend/src/pages/Graph/SummaryPanelCommon.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelCommon.tsx
@@ -121,7 +121,7 @@ export const getNodeMetrics = (
     case NodeMetricType.APP:
       return API.getAppMetrics(nodeData.namespace, nodeData.app!, options, nodeData.cluster);
     case NodeMetricType.SERVICE:
-      return API.getServiceMetrics(nodeData.namespace, nodeData.service!, options);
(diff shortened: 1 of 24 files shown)


josunect:
Param is not handle in the backend

josunect:
When we are in multi cluster, send the cluster parameter in the query:

image

josunect:
That should include the cluster as part of the metrics query: 

image

image

hhovsepy:
@josunect can you please add which areas are touched, so to be tested?
Because in Overview page, the metrics queries do not have cluster param:
Screenshot from 2023-05-16 16-19-50

hhovsepy (dismissed):
Looks good, @josunect thank you for the changes, tested locally.

frontend/src/pages/Graph/SummaryPanelCommon.tsx
```diff
@@ -121,7 +121,7 @@ export const getNodeMetrics = (
     case NodeMetricType.APP:
       return API.getAppMetrics(nodeData.namespace, nodeData.app!, options, nodeData.cluster);
     case NodeMetricType.SERVICE:
-      return API.getServiceMetrics(nodeData.namespace, nodeData.service!, options, nodeData.cluster);
+      return API.getServiceMetrics(nodeData.namespace, nodeData.service!, options);
```
jshaughn, line 124: Are you sure we want to drop the 'cluster' here?  When you have selected a service node it is within a cluster (and within a namespace).  I think we may need to keep this discriminator.

frontend/src/services/Api.ts
```diff
@@ -771,12 +771,8 @@ export const getIstioPermissions = (namespaces: string[], cluster?: string) => {
...
 };
 
-export const getMetricsStats = (queries: MetricsStatsQuery[], cluster?: string) => {
-  const queryParams: any = {};
-  if (cluster) {
-    queryParams.cluster = cluster;
-  }
-  return newRequest<MetricsStatsResult>(HTTP_VERBS.POST, urls.metricsStats, queryParams, { queries: queries });
+export const getMetricsStats = (queries: MetricsStatsQuery[]) => {
+  return newRequest<MetricsStatsResult>(HTTP_VERBS.POST, urls.metricsStats, {}, { queries: queries });
```
jshaughn, line 775: +1 on this change.

jshaughn (changes requested):
Initially I thought there was a missing 'cluster' param for the getNamespaceMetrics requests from the full-graph side-panel.  But after thinking about I think it's correct, the full-graph traffic totals ahousl not narrow to a single cluster on those queries, they should show total traffic across clusters, for the namespace (note, this is only relevant for single-namespace graphs, we don't make those queries for multi-namespace graphs).

But, I don't think the same applies to the NamespaceBox side panel.  A NamespaceBox is constrained to a parent cluster, and as such I think the 'cluster' param needs to be added (see SummaryPanelNamespaceBox:553)

matejnesuta:
Hello. Is this something that could use some back-end testing?

PR #6135, opened 2023-05-11
Remove the check for number of valid Kiali instances in remote clusters

### LINKED ISSUE #6102: Multicluster - Remove warning "Not all remote clusters have reachable Kiali instances"
As we are only expecting to have one Kiali instance, probably this message could be removed (Just for multicluster): 

image

josunect:
Fixes #6102


hhovsepy (approved):
Verified locally, no message is shown.

Issue #6136, opened 2023-05-11
Multicluster Workload list - Wrong details column content


hhovsepy:
The 'Details' column shown content from different cluster.
In multicluster setup when creating a config in one cluster, that config is shown for workload in other cluster as well in 'Details' column:
![Screenshot from 2023-05-11 10-44-52](https://github.com/kiali/kiali/assets/604313/511adcb0-f5b1-4d4c-ac98-6cc9849bbc89)

Issue #6020, opened 2023-04-12
Workload List gets Validations from backend but do not use them


hhovsepy:
Happens on Kiali master.
Outcome of PR https://github.com/kiali/kiali/pull/5953#issuecomment-1503928894

![Screenshot from 2023-04-11 18-09-37](https://user-images.githubusercontent.com/604313/231362159-7eb6e06d-2712-4536-83b7-5a292d6a3dd8.png)

In Workloads page of Kiali, when checking the responses backend, it returns Validaions for WorkloadList, but this info is not used in Workloads page.
We need to figure out where that Validations are used in Kiali and leave them as it is.
And in Workloads list page, either start showing Validations column and add a toggle to show/hide it, OR do not retrieve Validations from backend at all by adding a new "includeValiations" in WorkloadCriteria and set it to 'false' for WorkloadList page.

PR #6141, opened 2023-05-15
Workload and App lists are loading Istio configs per cluster.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6136

App and Workload lists pages are sending request to backend to load items per namespace and for all clusters,
and they require also loading Istio configs per Workload and App to be shown in Details column.
Before: Istio configs were loaded for all clusters, which was causing a mess.
Now: Istio configs are loaded per cluster.

New method 'GetIstioConfigMap' is introduced, this method will replace 'GetIstioConfigList' in further task: https://github.com/kiali/kiali/issues/6052
Loading workload validations is disabled, because it is never used, we have a task for this: https://github.com/kiali/kiali/issues/6020

How to test:
Create a PeerAuth in 'west' cluster, it should not be shown in 'east' clsuter's Workloads and Applications.

```
kind: PeerAuthentication
metadata:
  name: west
  namespace: bookinfo
spec:
  mtls:
    mode: STRICT
``` 

Screenshot from 2023-05-15 17-11-18
Screenshot from 2023-05-15 17-10-52

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -134,7 +134,7 @@ func (in *AppService) GetAppList(ctx context.Context, criteria AppCriteria) (mod
 		IncludeSidecars:               true,
 		IncludeVirtualServices:        true,
 	}
-	var istioConfigList models.IstioConfigList
+	var istioConfigMap models.IstioConfigMap
 
 	// TODO: MC
 	if criteria.IncludeIstioResources {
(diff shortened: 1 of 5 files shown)


hhovsepy:
fixing tests

josunect:
It works as expected: 

image

If the CRDs are not installed you can see just a message in the logs, but no errors. 

Just a couple of comments: 
- While the validations per cluster are not implement yet, we will be loosing the validation per single cluster? If so, I think they should probably go in the same PR?
- There is an error regarding validations, when the namespace is not found, there is a 500 error (Maybe this is related to the TODOs in the code?): 

image

josunect (approved):
Thanks for clarify, @hhovsepy ! LGTM

Issue #5922, opened 2023-03-16
Multicluster view for graph page


josunect:
- Include a filter by cluster? 
- Cluster boxes are enabled by default (Actual behavior)

Cluster boxes: 
![image](https://user-images.githubusercontent.com/49480155/225615759-1ba4d6c9-d80a-4049-b67f-116af86076e0.png)

Example with cluster filter: 
![image (4)](https://user-images.githubusercontent.com/49480155/226981954-8c59dd47-1fa6-4df0-a8c8-0ccd2a1981d8.jpg)

Ref. #5620

jshaughn:
I'm not sure there is work to do here.  We already have the `cluster` operand for Find/Hide, which should satisfy the filter requirement.  We will need to ensure that links are correct, they may need to add cluster as a query param.

jshaughn:
> I think we need to show the same cluster box for cluster west as we are showing for cluster east

The clusters are directly related to the traffic for the requested namespaces.  I don't think we want the cluster selector as seen in the the screenshot from @josunect, instead just the usual Namespace Selector and Find/Hide.   You should see East and West Clusters as much as necessary given the traffic into, out of, and within the requested namespaces.  This should happen automatically as long as we have unified metrics for the two clusters.

> When right clicking on some service, there is no menu today

This is expected until we add Multi-Cluster support to the graph.  Today it likely ignores clusters other than the "home" cluster.  But it will need to realize that items for any of the configured clusters are available.

leandroberetta:
> > I think we need to show the same cluster box for cluster west as we are showing for cluster east
> 
> The clusters are directly related to the traffic for the requested namespaces. I don't think we want the cluster selector as seen in the the screenshot from @josunect, instead just the usual Namespace Selector and Find/Hide. You should see East and West Clusters as much as necessary given the traffic into, out of, and within the requested namespaces. This should happen automatically as long as we have unified metrics for the two clusters.

I was referring just to the box, meaning, how cluster east looks like with all namespaces inside. 

I agree that probably the cluster selector is something we don't want.

jshaughn:
> Replace the parenthesis for the cluster badge:

I am working on this one in https://github.com/kiali/kiali/issues/6123

josunect:
This task involves several task. 

Closed by:

- https://github.com/kiali/kiali/issues/6123 
- https://github.com/kiali/kiali/issues/6150 
- https://github.com/kiali/kiali/issues/6062

PR #6142, opened 2023-05-15
Improve graph labels with PF badges instead of parentheses

### LINKED ISSUE #6123: Improve graph node labels - remove parentheses
To add necessary discriminators to Graph node labels we append namespace and/or cluster names in parentheses.  This has proven to be confusing and can also be hard to read.  It originated before we used PF badges and can now be improved by using fully-badged information.

This is related to https://github.com/kiali/kiali/issues/5922, which suggests that we add back single-node parent-boxes.  But that approach has proven to be cumbersome for larger graphs.  Hopefully this will be sufficient.  If not, we could add single-node boxing back, as a configurable ui-option (but I don't really want to).

Before:
image

After:
TBD

jshaughn:
Closes #6123 
Relates To #5922

In the graph, replace parenthesized label information with properly PF badged content.  This is for when labels add namespace and/or cluster disciminators to ensure label uniqueness.  

To test you need to generate situations where labels would previously use the parenthesized information for namespace/cluster discriminator.  Some ways of doing this:
- multi-cluster
- select multiple namespaces for the graph
- remove namespace and/or cluster boxing

Negative testing may also be useful, to ensure unnecessary badging doesn't now appear.  Different graph types, enable boxing, single namespace, single cluster, etc.

### DIFF
--- a/frontend/src/components/CytoscapeGraph/graphs/GraphStyles.ts
+++ b/frontend/src/components/CytoscapeGraph/graphs/GraphStyles.ts
@@ -107,22 +107,14 @@ const contentDefault = style({
   borderRadius: '3px',
   borderWidth: '1px',
   color: NodeTextColor,
-  padding: '3px 5px'
+  padding: '1px 5px'
 });
 
 const contentBox = style({
   backgroundColor: NodeTextBackgroundColorBox,
   color: NodeTextColorBox
 });
-/*
-const contentWithBadges = style({
-  borderBottomLeftRadius: 'unset',
-  borderColor: NodeBadgeBackgroundColor,
-  borderStyle: 'solid',
-  borderTopLeftRadius: 'unset',
-  borderLeft: '0'
-});
-*/
+
 const hostsClass = style({
   $nest: {
     '& div:last-child': {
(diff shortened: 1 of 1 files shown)


leandroberetta (commented):
I've been testing this and looks good but I noticed that there is not badge for the home cluster:

image

I think we need to add it.

jshaughn:
> I've been testing this and looks good but I noticed that there is not badge for the home cluster:

I understand your point, because it means the user has to be aware of his context (i.e. he needs to realize the home-cluster is the default),  but we do everything we can to minimize what's on-screen because of the issues with larger graphs.  I wouldn't make the change unless users say they find it confusing.  Or if there is a mutiny of engineers :-)

josunect:
Testing the PR and it looks good to me: 

image

I think not including the home cluster name could be confusing at the beginning. But once you know, I think it is more clear, as it makes a differentiation about which one is the home cluster. Also, if the cluster boxes are added, it is possible to see all the cluster names:

image

leandroberetta (approved):
LGTM.

jshaughn:
> I still think adding it won't hurt but I'm ok if we document this, or at least document it on the first question we receive in the future.

It's analogous to not showing the namespace-name for the single namespace selected in the namespace-selector dropdown.  In that case the user is expected to infer the namespace from the dropdown.  In this case the user is expected to infer the cluster from the masthead.  Anyway, if people report that they are confused I will support the change.

PR #6143, opened 2023-05-15
Try to fix flakes by ensuring button visibility before click

jshaughn:
- also, reformatting automatically applied, sorry, but I guess it's necessary.

https://issues.redhat.com/browse/OSSM-3940

If this passes upstream CI we need to backport to 2.4 (v1.65 branch) to try and get OCP CI to pass.


jshaughn:
> @jshaughn looks like there is no difference against OCP

From our Slack chat, this run was not actually executing the PR, so it can be ignored.

jmazzitelli (approved):

PR #6146, opened 2023-05-16
[v1.57] Adding missing packages to cypress image

FilipB:
Adding missing packages to cypress image


ScriptingShrimp (approved):
tested, OK

PR #6149, opened 2023-05-16
[v1.57] Adding readme for cypress tests in a container

FilipB:
Adding readme section which describes how to run cypress tests in a container


ScriptingShrimp (approved):
works locally

PR #6151, opened 2023-05-16
Add contextual menu for read links for remote clusters 

### LINKED ISSUE #6150: Add contextual menu in the nodes for the remote cluster
the same as in the local cluster, but just with read actions:

image

Ref. https://github.com/kiali/kiali/issues/5618

josunect:
image

There was some code that was removed, because it was expecting a link to a remote kiali instance, but in this case, we want the link to be in this Kiali instance (We are just expecting one)

Fixes #6150


leandroberetta (approved):
Looks good!

PR #6153, opened 2023-05-16
Showing cluster in the traffic details

leandroberetta:
The cluster column is added:

Screenshot from 2023-05-16 14-54-10


jshaughn (approved):
No problemas aqui!

Issue #6155, opened 2023-05-17
molecule test failure - changing security context not working


jmazzitelli:
Noticed a failure in the molecule test that tests the security context can be customized. Need to figure out if this is just a side-effect of k8s or we really do have a bug.

Customizing the security context [here](https://github.com/kiali/kiali-operator/blob/v1.67.0/molecule/config-values-test/converge.yml#L156-L167) and testing it was changed [here](https://github.com/kiali/kiali-operator/blob/v1.67.0/molecule/config-values-test/converge.yml#L284-L293)

Notice `spec.containers[0].securityContext.readOnlyRootFilesystem` wants to be customized to false but ends up being true. This is what I saw in the deployment's template, which does show it is true (but should be false):

```yaml
kind: Deployment
spec:
  template:
    spec:
      containers:
        securityContext:
          allowPrivilegeEscalation: false
          capabilities:
            drop:
            - ALL
          privileged: false
          readOnlyRootFilesystem: true
          runAsNonRoot: true

```

PR #6158, opened 2023-05-17
fix the config-values-test molecule test when testing with OLM

jmazzitelli:
This test requires the operator to allow security context override; when installing the operator via OLM we need to enable that feature.

fixes: https://github.com/kiali/kiali/issues/6155


jshaughn (approved):
LOK

Issue #6161, opened 2023-05-18
Multicluster support Namespace Validations


hhovsepy:
Currently from the Graph page the requests "/api/namespaces/{namespace}/validations" does not contain `cluster` param and it returns 500 when namespace it not found in home cluster.

![Screenshot from 2023-05-18 12-04-42](https://github.com/kiali/kiali/assets/604313/316dab5b-3941-4fd2-a494-8e186383cbed)

PR #6164, opened 2023-05-18
unified validations for summary graph

josunect:
In a multi cluster environment, if we don't send the cluster parameter, it means we want the validations from all the clusters that include that namespace. This is a unified view for a particular case, the summary panel for the graph: 

 
image


hhovsepy:
https://github.com/kiali/kiali/issues/6161

josunect:
As it is right now, it is not accurate because validations use the name, the object type and the namespace as the key. But in this case, the cluster should also be used as the key so the validations summary show the real number of objects analyzed: 

image

image

hhovsepy (approved):
Verified on multicluster and OCP, lgtm,
Screenshot from 2023-05-19 18-35-32

Issue #6169, opened 2023-05-22
Multicluster - Overview page istio status


hhovsepy:
Currently for `/api/istio/status/` requests per 'istiod' from Overview page there is not `cluster` param, so for all clusters it returns the home cluster 'istiod' status.

PR #6170, opened 2023-05-22
Added cluster param in Istiod status request, Overview page and Masthead

hhovsepy:
RFE https://github.com/kiali/kiali/issues/6169

### DIFF
--- a/frontend/src/pages/Overview/OverviewPage.tsx
+++ b/frontend/src/pages/Overview/OverviewPage.tsx
@@ -957,7 +957,9 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
                                 title={ns.name}
                               >
                                 {ns.name}
-                                {ns.name === serverConfig.istioNamespace && <ControlPlaneBadge></ControlPlaneBadge>}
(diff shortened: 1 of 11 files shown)


frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -957,7 +957,9 @@ export class OverviewPage extends React.Component<OverviewProps, State> {
                                 title={ns.name}
                               >
                                 {ns.name}
-                                {ns.name === serverConfig.istioNamespace && <ControlPlaneBadge></ControlPlaneBadge>}
+                                {ns.name === serverConfig.istioNamespace && (
```
josunect, line 960: Now in multi cluster, we have the possibility to have a remote cluster with the istioNamespace but no Istiod, what do you think to extend this check to this condition? Or maybe we can think about this when the Overview page is re worked?

hhovsepy: I would leave that for next iteration with Overview rework. To keep this PR only for adding cluster param.

josunect (approved):
LGTM

PR #6172, opened 2023-05-22
Kiali does not handle namespace-level injection enable/disable for Maistra

jshaughn:
Update Cypress to handle that use-case.

https://issues.redhat.com/browse/OSSM-3989

### DIFF
--- a/frontend/cypress/integration/common/sidecar_injection.ts
+++ b/frontend/cypress/integration/common/sidecar_injection.ts
@@ -239,32 +239,64 @@ When('I visit the overview page', function () {
 });
 
 When('I override the default automatic sidecar injection policy in the namespace to enabled', function () {
-  cy.get('[data-test=overview-type-LIST]').should('be.visible').click();
-  cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).should('be.visible').click();
-  cy.get(`[data-test=enable-${this.targetNamespace}-namespace-sidecar-injection]`).should('be.visible').click();
-  cy.get('[data-test=confirm-traffic-policies]').should('be.visible').click();
-  ensureKialiFinishedLoading();
+  cy.request('GET', '/api/status').should(response => {
+    expect(response.status).to.equal(200);
+    const isMaistra = response.body.istioEnvironment.isMaistra;
+
+    cy.get('[data-test=overview-type-LIST]').should('be.visible').click();
+    cy.get('[data-test=overview-type-LIST]').should('be.visible').click();
+    cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).should('be.visible').click();
+    // Maistra does not support namespace-level injection
+    if (isMaistra) {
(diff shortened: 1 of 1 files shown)


frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -239,32 +239,64 @@ When('I visit the overview page', function () {
...
+    expect(response.status).to.equal(200);
+    const isMaistra = response.body.istioEnvironment.isMaistra;
+
+    cy.get('[data-test=overview-type-LIST]').should('be.visible').click();
+    cy.get('[data-test=overview-type-LIST]').should('be.visible').click();
+    cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).should('be.visible').click();
+    // Maistra does not support namespace-level injection
+    if (isMaistra) {
+      cy.get(`[data-test=enable-${this.targetNamespace}-namespace-sidecar-injection]`).should('not.exist');
+    } else {
```
nrfox, line 252: These actually seem like two separate tests. You don't need to include this with this change but it'd be nice to start using tags to include/exclude certain tests based on the env. We'll need to do this for multi-cluster tests and probably other things as well.

leandroberetta (approved):
Looks good.

frontend/cypress/integration/common/sidecar_injection.ts
```diff
@@ -239,32 +239,64 @@ When('I visit the overview page', function () {
...
 
 When('I override the default automatic sidecar injection policy in the namespace to enabled', function () {
-  cy.get('[data-test=overview-type-LIST]').should('be.visible').click();
-  cy.get(`[data-test=VirtualItem_${this.targetNamespace}] button`).should('be.visible').click();
-  cy.get(`[data-test=enable-${this.targetNamespace}-namespace-sidecar-injection]`).should('be.visible').click();
-  cy.get('[data-test=confirm-traffic-policies]').should('be.visible').click();
-  ensureKialiFinishedLoading();
+  cy.request('GET', '/api/status').should(response => {
+    expect(response.status).to.equal(200);
+    const isMaistra = response.body.istioEnvironment.isMaistra;
```
nrfox, line 244: It'd be nice to get this kind of info once before any of the tests are run but I'm not sure the best way of doing that within cypress.

nrfox (approved):
Just some comments for the future but otherwise LGTM

frontend/cypress/integration/common/sidecar_injection.ts
jshaughn: Yes, at the moment just trying to get the stuff passing so it's not blocking 2.4.  With 3.0 it's possible things will align with upstream and we won't have the disparity.

frontend/cypress/integration/common/sidecar_injection.ts
jshaughn: me either...

PR #6178, opened 2023-05-23
Check kube cache namespace

josunect:
fix istio_api disabled issues


hhovsepy:
@josunect thank you for the fix, it fixed the problem of loading Workloads and failing integration test.
But test 'emptyValidations' is failing case and needs to be disabled, as it is disabled in master.

josunect:
Thank you, @hhovsepy 

That is right. I'll send the change.

hhovsepy (approved):
thank you @josunect

Issue #6098, opened 2023-05-03
After navigating to the Service graph, all other graphs are displayed with injectServiceNodes=false


rsalmond:
To repro, visit the Graph tab, click the "reset to factory settings" button, then use the drop down to choose "service graph", then use the drop down to select any other graph type and you should see app/workload displayed without the service icon.

---
Before:

<img width="689" alt="image" src="https://user-images.githubusercontent.com/4367945/235962480-0f9d9116-c616-4faa-a18e-8a2429b5e412.png">

URI: /kiali/console/graph/namespaces/?traffic=grpc%2CgrpcRequest%2Chttp%2ChttpRequest%2Ctcp%2CtcpSent&graphType=versionedApp&namespaces=blue%2Ccert-manager%2Cdefault%2Cistio-system%2Cprism%2Cred&duration=60&refresh=60000&layout=kiali-dagre&namespaceLayout=kiali-dagre

---

During:

<img width="674" alt="image" src="https://user-images.githubusercontent.com/4367945/235962612-e89373be-94b7-477a-8bb6-ef6f0ff518ed.png">

URI: /kiali/console/graph/namespaces/?traffic=grpc%2CgrpcRequest%2Chttp%2ChttpRequest%2Ctcp%2CtcpSent&graphType=service&namespaces=blue%2Ccert-manager%2Cdefault%2Cistio-system%2Cprism%2Cred&duration=60&refresh=60000&layout=kiali-dagre&namespaceLayout=kiali-dagre&injectServiceNodes=false

---

After:

<img width="684" alt="image" src="https://user-images.githubusercontent.com/4367945/235962784-2a55313f-5d3c-44aa-aa51-f09b25cdea4f.png">

URI: /kiali/console/graph/namespaces/?traffic=grpc%2CgrpcRequest%2Chttp%2ChttpRequest%2Ctcp%2CtcpSent&graphType=versionedApp&namespaces=blue%2Ccert-manager%2Cdefault%2Cistio-system%2Cprism%2Cred&duration=60&refresh=60000&layout=kiali-dagre&namespaceLayout=kiali-dagre&injectServiceNodes=false

jshaughn:
You are right, it's sort of a known quirk. I'd have to look again to see if it's something we can get around.  Thanks.

jshaughn:
This has confused enough people over the years, and looks reasonably easy to change, that we will do it.

PR #6188, opened 2023-05-23
When generating a Service graph don't change user display settings

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6098

When generating a Service graph don't change the users' settings for showServiceNodes and showOperationNodes.  The user shouldn't care that those being enabled does not make sense for Service graphs, and shouldn't get surprised by them suddenly getting disabled. Instead, let the backend ignore those settings as needed.


jshaughn:
To Test:  Just ensure that the settings for Display->Service Nodes and Display->Operation Nodes are maintained despite changing to/from Service Graph type.

PR #6189, opened 2023-05-23
[master] Fix apps CI failure in downstream CI

jshaughn:
make apps test happy even if bookinfo has the traffic generator app

forward-port of ossm-3940 change in v1.65

### DIFF
--- a/frontend/cypress/integration/common/apps.ts
+++ b/frontend/cypress/integration/common/apps.ts
@@ -113,10 +113,6 @@ Then('the health status of the application should be {string}', function (health
   checkHealthStatusInTable(this.targetNamespace, null, this.targetApp, healthStatus);
 });
 
-Then('user cannot see any apps in the table', () => {
-  cy.get('h5').contains('No applications found').should('exist');
-});
-
 Then('user sees all the Apps toggles', () => {
   cy.get('[data-test="toggle-health"]').should('be.checked');
   cy.get('[data-test="toggle-istioResources"]').should('be.checked');
--- a/frontend/cypress/integration/common/apps.ts
+++ b/frontend/cypress/integration/common/apps.ts
@@ -135,3 +135,16 @@ When('user {string} toggle {string}', function (action: 'checks' | 'unchecks', t
 Then('the {string} column {string}', (col: string, action: 'appears' | 'disappears') => {
   colExists(col, action === 'appears');
 });
+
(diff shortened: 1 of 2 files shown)


frontend/cypress/integration/common/apps.ts
ScriptingShrimp, line 1: `Then('user cannot see any apps in the table', () => {
  cy.get('h5').contains('No applications found').should('exist');
});`

this code should be removed since its not used anywhere else

frontend/cypress/integration/common/apps.ts
```diff
@@ -135,3 +135,16 @@ When('user {string} toggle {string}', function (action: 'checks' | 'unchecks', t
...
 });
+
+Then('user may only see {string}', (sees: string) => {
+  cy.get('tbody').within(() => {
+    cy.get('tr').should('have.length', 1);
+    cy.get('td').then(td => {
+      if (td.length === 1) {
+        cy.get('h5').contains('No applications found');
+      } else {
+        cy.contains('tr', sees);
```
ScriptingShrimp, line 146: I'm not completely sure how this works.
Does `if else` condition handles the presence of traffic generator?

ScriptingShrimp (changes requested):
Passed on OCP, all green expect `Controlling sidecar injection.Switch to using the default policy for automatic sidecar injection in a namespace` - which is not related in this case

frontend/cypress/integration/common/apps.ts
jshaughn: OK, I'll remove.

frontend/cypress/integration/common/apps.ts
jshaughn: Well this code is generic but yes, this handles the fact that in different CI environments we may or may not have the traffic generator or it may or may not have a sidecar injected.  It just allows for it to be there or not be there in the apps list.  It's a concession to our different needs in different places.

So now in the feature you see `Then user may only see "kiali-traffic-generator"` because in some CI (downstream) it's expected to be there and in some places it isn't (upstream).

ScriptingShrimp: Thanks for explanation. :+1: 
Noting for myself - it will be interesting to see if we come across more exceptions like this in future

ScriptingShrimp (approved):
LGTM!

Issue #6198, opened 2023-05-24
Multi-cluster graph - update AccessibleNamespaces to be cluster-aware


jshaughn:
Epic: #5618 

Namespace graphs are independent of clusters, only caring about the traffic for the requested namespaces.  The resulting graph renders the traffic for namespaces regardless of whether that traffic is cross-cluster.  But because we append k8s-based config information to the noes, that node-specific code needs to be cluster-aware.  In other words, the graph itself is not, but the nodes are, cluster-aware.   The graph appenders decorate nodes using k8s configuration information.  They need to know whether the user has access to the necessary namespaces.  Because namespace access for a user can differ for different clusters, we need to make the `AccessibleNamespaces` structure cluster-aware, and then apply that as needed to the affected appenders.

PR #6200, opened 2023-05-25
Include cluster in health appender request key

### LINKED ISSUE #6185: Multi-cluster graph - error when user lacks access to remote cluster
Given two clusters, `east` and `west`, with a `bookinfo` namespace in both and a user that has access to `bookinfo.east` but not `bookinfo.west`, Kiali sometimes shows the graph and sometimes shows an access error. There seems to be some access check that is performed but the exact cluster that it is performed against varies. The graph should only show namespaces to the user that they have access to but should not throw an error when that namespace exists on both clusters.

namespace_access_error

nrfox:
Key did not include cluster and when multiple namespaces exist across clusters the request was being overwritten.

Fixes #6185

### DIFF
--- a/graph/telemetry/istio/appender/health.go
+++ b/graph/telemetry/istio/appender/health.go
@@ -156,10 +156,11 @@ func (a *HealthAppender) attachHealthConfig(trafficMap graph.TrafficMap, globalI
 func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo) {
 	var nodesWithHealth []*graph.Node
 	type healthRequest struct {
-		app      bool
-		service  bool
-		workload bool
-		cluster  string
+		app       bool
+		service   bool
+		workload  bool
+		cluster   string
+		namespace string
 	}
 
 	// Health requests are per namespace meaning if a single node in the namespace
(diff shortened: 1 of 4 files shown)


nrfox:
I should clarify that this doesn't actually fix the forbidden issue. That is caused by #6198. But it does fix the issue where sometimes you could see a graph for namespaces in a remote cluster that you didn't have access to.

nrfox:
@leandroberetta when you get a chance would you mind taking a look at this?

leandroberetta (approved):
Looks good.

graph/telemetry/istio/appender/health.go
```diff
@@ -182,7 +183,7 @@ func (a *HealthAppender) attachHealth(trafficMap graph.TrafficMap, globalInfo *g
 
 		var req healthRequest
 		var ok bool
-		if req, ok = healthReqs[n.Namespace]; !ok {
+		if req, ok = healthReqs[n.Namespace+n.Cluster]; !ok {
```
jshaughn, line 186: Minor nit, across the graph appender code, when needing a key like this, the format used has been `cluster:namespace`.  In general, it's safer to use a delimiter when combining strings, and conventionally the cluster has been left of the namespace in the graph-related keys.

Issue #6196, opened 2023-05-24
Istio validations wihen istio api is disabled


josunect:
After the some changes introducing the kube cache, istio_api disabled was not working properly. 
After the fix, the test related to validations was commented because the validations were returned when istio_api was disabled. 
This will need to view reviewed so: 

- Test can be uncommented
- If validations are working, test if they are available with no Istiod 
- If are not working, they should not be returned
- Update documentation if required

josunect:
- Istio configs are not available in OCP.

PR #6204, opened 2023-05-25
get config list by namespace when not all namespaces are accessible

josunect:
Fixes https://github.com/kiali/kiali/issues/6196

Validations test in no_istiod can be uncommented now. 
The behavior will keeps the same - validations not available, Istio configs available.


josunect:
> LGTM, tested locally, also when istio api is disabled: Screenshot from 2023-05-25 18-55-56

Thanks for the review!

Issue #6203, opened 2023-05-25
Refactor some integration tests code


josunect:
This is a follow up task for https://github.com/kiali/kiali/issues/6121 to refactor some code and create generic functions to manage pods (list pods, restart). The tests that can be refactor are: 

- no_istiod_test
- remote_istiod_test

josunect:
Testing in OCP

PR #6205, opened 2023-05-25
Create kube client with common funcions

josunect:
Refactor some integration tests. 

Fixes https://github.com/kiali/kiali/issues/6203 

- Use Istio Deployment name for patching istio deployment
- Use nginx unprivileged for proxy patch so no permission issues in OCP.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -220,7 +220,7 @@ type IstioConfig struct {
 	ConfigMapName                     string              `yaml:"config_map_name,omitempty"`
 	EnvoyAdminLocalPort               int                 `yaml:"envoy_admin_local_port,omitempty"`
 	GatewayAPIClassName               string              `yaml:"gateway_api_class_name,omitempty"`
-	IstioAPIEnabled                   bool                `yaml:"istio_api_enabled"`
+	IstioAPIEnabled                   bool                `yaml:"istio_api_enabled,omitempty"`
 	IstioCanaryRevision               IstioCanaryRevision `yaml:"istio_canary_revision,omitempty"`
 	IstioIdentityDomain               string              `yaml:"istio_identity_domain,omitempty"`
 	IstioInjectionAnnotation          string              `yaml:"istio_injection_annotation,omitempty"`
--- a/config/config.go
+++ b/config/config.go
@@ -220,7 +220,7 @@ type IstioConfig struct {
 	ConfigMapName                     string              `yaml:"config_map_name,omitempty"`
 	EnvoyAdminLocalPort               int                 `yaml:"envoy_admin_local_port,omitempty"`
 	GatewayAPIClassName               string              `yaml:"gateway_api_class_name,omitempty"`
-	IstioAPIEnabled                   bool                `yaml:"istio_api_enabled,omitempty"`
+	IstioAPIEnabled                   bool                `yaml:"istio_api_enabled"`
 	IstioCanaryRevision               IstioCanaryRevision `yaml:"istio_canary_revision,omitempty"`
 	IstioIdentityDomain               string              `yaml:"istio_identity_domain,omitempty"`
 	IstioInjectionAnnotation          string              `yaml:"istio_injection_annotation,omitempty"`
--- a/tests/integration/utils/kube/kube_client.go
+++ b/tests/integration/utils/kube/kube_client.go
@@ -2,8 +2,6 @@ package kube
 
 import (
 	"context"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/types"
 	"strings"
 	"testing"
 	"time"
@@ -12,6 +10,8 @@ import (
 	"gopkg.in/yaml.v2"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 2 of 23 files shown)


tests/integration/utils/kube_client.go
```diff
@@ -0,0 +1,137 @@
...
+				}
+			}
+		}
+		return true, nil
+	})
+}
+
+// Returns the name of the Kiali pod
+// It expects to find just one Kiali pod
+func GetKialiPodName(kubeClient kubernetes.Interface, kialiNamespace string, ctx context.Context, t *testing.T) string {
```
nrfox, line 101: nit: context is usually passed as the first argument to the function
```suggestion
func GetKialiPodName(ctx context.Context, kubeClient kubernetes.Interface, kialiNamespace string, t *testing.T) string {
```

tests/integration/utils/kube_client.go
```diff
@@ -0,0 +1,137 @@
...
+	require := require.New(t)
+	pods, err := kubeClient.CoreV1().Pods(kialiNamespace).List(ctx, metav1.ListOptions{LabelSelector: "app=kiali"})
+	require.NoError(err)
+	require.Len(pods.Items, 1)
+
+	return pods.Items[0].Name
+}
+
+// Get Kiali config map
+func GetKialiConfigMap(kubeClient kubernetes.Interface, kialiNamespace string, kialiName string, ctx context.Context, t *testing.T) (*config.Config, *v1.ConfigMap) {
```
nrfox, line 112:
```suggestion
func GetKialiConfigMap(ctx context.Context, kubeClient kubernetes.Interface, kialiNamespace string, kialiName string,  t *testing.T) (*config.Config, *v1.ConfigMap) {
```

tests/integration/utils/kube_client.go
```diff
@@ -0,0 +1,137 @@
...
+	require.NoError(err)
+
+	currentConfig := config.NewConfig()
+	require.NoError(yaml.Unmarshal([]byte(cm.Data["config.yaml"]), currentConfig))
+
+	return currentConfig, cm
+}
+
+// Update Kiali config map
+func UpdateKialiConfigMap(kubeClient kubernetes.Interface, kialiNamespace string, currentConfig *config.Config, cm *v1.ConfigMap, ctx context.Context, t *testing.T) {
```
nrfox, line 127:
```suggestion
func UpdateKialiConfigMap(ctx context.Context, kubeClient kubernetes.Interface, kialiNamespace string, currentConfig *config.Config, cm *v1.ConfigMap, t *testing.T) {
```

tests/integration/utils/kube_client.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Restart Kiali pod when kiali CRD does not exist (Otherwise, operator will delete the old one)
+	if !keepOldPod {
+		log.Debugf("Deleting Kiali pod %s", currentKialiPod)
+		err := kubeClient.CoreV1().Pods(namespace).Delete(ctx, currentKialiPod, metav1.DeleteOptions{})
+		if err != nil {
+			log.Errorf("Error deleting Kiali pod %s", err)
+			return err
+		}
+	}
```
nrfox, line 70: When `keepOldPod == true` then this func just waits for the Kiali pod to be ready? Maybe can split out the "waiting for kiali to be healthy" into its own func and remove the need for the bool arg?

tests/integration/utils/kube_client.go
```diff
@@ -0,0 +1,137 @@
+package utils
```
nrfox, line 1: nit: utils packages don't really describe what's in the package. This is just personal preference so feel free to ignore this but imo it's better to have more packages that describe what's in them than less packages that are more generic. Could maybe be a `utils/kiali` package and a `utils/kube` package? Or `utils/kube/kiali`?

nrfox (commented):
Just some minor comments but nothing blocking. Nice refactor!

tests/integration/utils/kube_client.go
josunect: updated

tests/integration/utils/kube_client.go
josunect: updated

config/config.go
```diff
@@ -220,7 +220,7 @@ type IstioConfig struct {
 	ConfigMapName                     string              `yaml:"config_map_name,omitempty"`
 	EnvoyAdminLocalPort               int                 `yaml:"envoy_admin_local_port,omitempty"`
 	GatewayAPIClassName               string              `yaml:"gateway_api_class_name,omitempty"`
-	IstioAPIEnabled                   bool                `yaml:"istio_api_enabled,omitempty"`
+	IstioAPIEnabled                   bool                `yaml:"istio_api_enabled"`
```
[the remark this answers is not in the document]
josunect: That is a good question. I've realized that, when doing the marshal with the config this value was not updated. When it was false, the config map was not updated, and by default, it was true. 
But, maybe there is another way to change it..

tests/integration/utils/kube_client.go
josunect: Created packages.

josunect:
CI passed (Also in OCP)

tests/integration/utils/kube/kube_client.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		log.Errorf("Error deleting Kiali pod %s", err)
+		return err
+	}
+	return nil
+}
+
+// Deletes the existing kiali Pod and waits for the new one to be ready.
+func RestartKialiPod(ctx context.Context, kubeClient kubernetes.Interface, namespace string, currentKialiPod string) error {
+	log.Debugf("Restarting kiali pod %s %s", namespace, currentKialiPod)
```
nrfox, line 68: nit: since this doesn't delete anymore can we call it something else and update or remove the comment/log statement?

nrfox (approved):
1 small nit but not blocking. Thanks for cleaning this up!

tests/integration/utils/kube/kube_client.go
josunect: Good catch! Updated

nrfox (approved):

PR #6206, opened 2023-05-26
Update time out so the pod restart not affects other tests

josunect:
pod_logs_test OCP failure


hhovsepy (approved):
Tests passed on OSSM 2.4.

PR #6207, opened 2023-05-26
Print int tests output both to the console and the log file

FilipB:


josunect (approved):
I can see the tests output in the pipeline. Thank you!

PR #6217, opened 2023-05-30
Disable integration and cypress testing on release

leandroberetta:


jshaughn:
> Is this related to an issue? Why are we disabling testing releases?

Because we've already run CI on everything in the release, and we;ve had too many releases fail due to some unexpected failure that seems to bite the release CI.  Do you have a strong objection?  We can always vote on it.

nrfox:
> > Is this related to an issue? Why are we disabling testing releases?
> 
> Because we've already run CI on everything in the release, and we;ve had too many releases fail due to some unexpected failure that seems to bite the release CI. Do you have a strong objection? We can always vote on it.

No objections! We could do some sanity checks on the published image after release but that doesn't require running the full test suite against it and the current release testing builds the image from scratch so that's even less useful.

jshaughn (approved):
LG!

Issue #6218, opened 2023-05-31
Update Kiali to build with Go 1.20


aljesusg:
Changes:

- Update the build scripts to specify Go 1.20 as the target version.
- Test the build process and application functionality with Go 1.20.
- have tested these changes on multiple platforms and confirmed that Kiali builds and works as expected with Go 1.20

Related issues:

- https://go.dev/issue/57855
- https://pkg.go.dev/vuln/GO-2023-1571
- https://groups.google.com/g/golang-announce/c/V0aBFqaFs_E

jshaughn:
All PRs are merged.

PR #6221, opened 2023-05-31
[1.65] Upgrade go to 1.20

aljesusg:
Fix https://github.com/kiali/kiali/issues/6218

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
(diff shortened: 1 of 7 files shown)


aljesusg:
Why don't tests run on these branches?

jmazzitelli:
> Why don't tests run on these branches?

Change the `branches` setting in `kiali-ci.yml` -- lines 6 and 12 here:

https://github.com/kiali/kiali/blob/master/.github/workflows/kiali-ci.yml

I think those regexes are bad (see the github docs on this [here](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-branches-or-tags-for-push-events)).

I would try this:

```
    branches:
    - master
    - v*.*
```

jmazzitelli:
@aljesusg Do you think this is OK to merge?

PR #6222, opened 2023-05-31
[1.57] Upgrade go to 1.20

aljesusg:
Fix https://github.com/kiali/kiali/issues/6218

I just remembered all the lints that had to be fixed...

### DIFF
--- a/.github/workflows/build-backend.yml
+++ b/.github/workflows/build-backend.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.4
+          go-version: 1.20.3
 
       - name: Cache Go deps
         uses: actions/cache@v3
--- a/.github/workflows/config/.golangci.yml
+++ b/.github/workflows/config/.golangci.yml
@@ -1,3 +1,11 @@
+run:
+  skip-files:
+    - doc.go
+    - ".*\\.pb\\.go$"
+  skip-dirs:
+    - frontend  
+  timeout: 5m
+  tests: true  
 linters:
   enable:
     - goimports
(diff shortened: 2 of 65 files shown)


aljesusg:
Why don't tests run on these branches?

jmazzitelli:
> Why don't tests run on these branches?

Change the `branches` setting in `kiali-ci.yml` -- lines 6 and 12 here:

https://github.com/kiali/kiali/blob/master/.github/workflows/kiali-ci.yml

I think those regexes are bad (see the github docs on this [here](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-branches-or-tags-for-push-events)).

I would try this:

```
    branches:
    - master
    - v*.*
```

jmazzitelli:
@aljesusg rebase this PR on the latest branch and the tests should now run

jmazzitelli:
@aljesusg Do you think this is OK to merge?

aljesusg:
Do we want to add the go-restful too?

jmazzitelli:
> Do we want to add the go-restful too?

Since this is for OSSM 2.3, it might be a good idea to also add go-restful update, too. So I would say `yes`.

jmazzitelli:
@jshaughn what do you think?

jshaughn (approved):
@jmazzitelli , I think we can merge this.  The CI failure is due to a problem in the CI config (see https://github.com/kiali/kiali/pull/6282 for the solution to that problem).  I don't think we need to backport the go-restful update as there is no formal issue.

jmazzitelli:
the backend int. tests pass, the front end can't start for some reason. Looks like a CI issue. I think this is mergeable

PR #6223, opened 2023-05-31
[github workflow] fix regex in branches setting

jmazzitelli:
see https://docs.github.com/en/actions/using-workflows/triggering-a-workflow#using-filters-to-target-specific-branches-or-tags-for-push-events


aljesusg (approved):
LGFM

Issue #6070, opened 2023-04-27
Multicluster - The Overview page makes a Validations call for all namespaces with no cluster parameter 


josunect:
It is not possible to identify the namespace. The call is made like: 

`http://localhost:3000/api/istio/validations?namespaces=istio-system,istio-system,bookinfo,bookinfo,default,default,ingress-nginx,ingress-nginx,keycloak,kiali-access-ns`

We can do N calls, where N = number of clusters, like:

`http://localhost:3000/api/istio/validations?namespaces=istio-system,bookinfo,default,ingress-nginx,keycloak&cluster=east`
`http://localhost:3000/api/istio/validations?namespaces=istio-system,bookinfo,default,ingress-nginx,kiali-access-ns&cluster=west`

https://github.com/kiali/kiali/blob/master/frontend/src/pages/Overview/OverviewPage.tsx#L499

REF: https://github.com/kiali/kiali/pull/6066#discussion_r1178767846

josunect:
cc @hhovsepy

hhovsepy:
We need to resolve this issue, as validations are working only for Home cluster.
The further improvements of validations will be done later with caching validation results and applying remote cluster validations.

PR #6228, opened 2023-06-01
Skip validations for OverviewPage remote cluster's namespaces

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6070

Before: 
There was request to backend with a chunks of namespaces, and without cluster params, the namespaces were duplicated. So in Overview page for remote cluster's namespaces it was showing the home cluster validations.

Now:
Single validations and configs request to backend without namespaces param, will return validations and configs by default for home cluster.
Only home cluster's config validations are shown, for remote ones N/A.

TODO: Separate sub-epic for remote cluster's configs validations.

Screenshot from 2023-06-01 13-26-19


josunect (approved):
Thanks for the fix, @hhovsepy 

image

Just one validations call is done: 

image

Issue #5028, opened 2022-04-29
Sidecar usage extension for the kiali operator.


libesz:
Hi!
 I would like to discuss a generic extension possibility to the Kiali operator. The PromQL API that I would like to integrate Kiali to, is requesting a API bearer token that is expiring very often. Since Kiali is not able (and I guess also not planning to be able 😄 ) to renew tokens or exchange API keys to tokens, I am thinking on another solution. (see previous ticket from our colleague: https://github.com/kiali/kiali/issues/4677). 
 So the idea would be that:
* the integrator (us) is implementing the token renewal magic and a (transparent) proxy as part of a sidecar container in the kiali pod
* prometheus endpoint would point to this sidecar (i.e. localhost)
* the integrator configures the sidecar container in the kiali cr

Having the sidecar in a standalone pod would need it's own authentication and TLS on it's HTTP interface to avoid misuse. Implementing it as a classic proxy sidecar seems to be cleaner.

This could be done via adding a new kiali operator config object which let's the integrator to add a complete container to the kiali pod spec. This would be very similar to what [additional_service_yaml](https://github.com/kiali/kiali-operator/blob/master/crd-docs/crd/kiali.io_kialis.yaml#L220) does with the kiali service object.

jmazzitelli:
I don't fully grok the proposed solution, but... wouldn't it be simpler to just update the Kiali CR with the new prometheus credentials? The operator will see the change and immediately propagate the new credentials down to the Kiali ConfigMap and automatically restart the Kiali pod so it picks up the new credentials. This is already supported. You would just need to patch the Kiali CR with the new credentials in the `spec.external_services.prometheus.auth` section. Or am I missing something?

israel-hdez:
> wouldn't it be simpler to just update the Kiali CR with the new prometheus credentials?

My guess is that it comes down to efficiency.

I note that @libesz says that the token expires "very often". The typical expiration I've seen for these cases is 5 minutes. You will want to roll out a fresh token before the old one expires to avoid the situation when Kiali can no longer reach Prometheus, so let's assume that you refresh every 4 minutes. That's 15 updates in an hour. The operator will use way much more resources for all these reconciliations (which may also flood the kubeapi), than the resources a sidecar may use by doing a simple request to refresh a token. Also, a sidecar can refresh on-demand (i.e. if nobody uses Kiali for some time, no need to refresh the token).

[...]

PR #6232, opened 2023-06-05
Support exec credentials in remote secret

### LINKED ISSUE #6193: Use ExecPlugin for remote cluster secret
Can I use ExecPlugin for remote cluster secrets in Kiali?

I have tried using remote cluster secret in a Kiali multi cluster configuration and it does not work properly.
I am using AWS EKS and would like to use the ExecPlugin provided by AWS to access the remote cluster.
The commands required for ExecPlugin are installed via Kiali's init container.

```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: <certificate-authority-data>
    server: <server-url>
  name: test
contexts:
- context:
    cluster: test
    user: test
  name: test
current-context: test
kind: Config
preferences: {}
users:
- name: test
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      args:
        - token
        - -i
        - test
      command: /var/custom-commands/aws-iam-authenticator
      env:
        - name: AWS_STS_REGIONAL_ENDPOINTS
          value: regional
        - name: AWS_DEFAULT_REGION
          value: ap-northeast-1
      interactiveMode: IfAvailable
      provideClusterInfo: false
```

I checked the Kiali source code and it looks like the remote secret parse ignores methods other than user token.
https://github.com/kiali/kiali/blob/9ce7013534aee7df2cd66345b2e6161ad3719dd5/kubernetes/secret.go#L20-L46

Is it possible to support ExecPlugin in Kiali's remote cluster secrets?

frauniki:
**Describe the change** 

Currently, kiali only supports SA bearer tokens for connections to the remote cluster when the auth strategy is Anonymous.
This PR allows the use of [ExecProvider](https://kubernetes.io/docs/reference/access-authn-authz/authentication/#client-go-credential-plugins) supported by kubernetes in addition to SA bearer tokens.

Please note that while there is a discussion of sidecar support by kiali in #6193, these features are not necessarily required in this case. This is because at this time users are free to add sidecar and other definitions by using the json patch in kustomize, etc.

 **Issue reference** 

- Fixed #6193

### DIFF
--- a/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
+++ b/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
@@ -36,7 +36,7 @@ DEFAULT_REMOTE_CLUSTER_CONTEXT="west"
 DEFAULT_REMOTE_CLUSTER_NAME=""
 DEFAULT_REMOTE_CLUSTER_NAMESPACE="kiali-access-ns"
 DEFAULT_VIEW_ONLY="true"
-DEFAULT_EXEC_AUTH="false"
+DEFAULT_EXEC_AUTH_JSON=""
 
 : ${ALLOW_SKIP_TLS_VERIFY:=${DEFAULT_ALLOW_SKIP_TLS_VERIFY}}
 : ${CLIENT_EXE:=${DEFAULT_CLIENT_EXE}}
@@ -52,7 +52,7 @@ DEFAULT_EXEC_AUTH="false"
 : ${REMOTE_CLUSTER_NAMESPACE:=${DEFAULT_REMOTE_CLUSTER_NAMESPACE}}
 : ${REMOTE_CLUSTER_NAME:=${DEFAULT_REMOTE_CLUSTER_NAME}}
 : ${VIEW_ONLY:=${DEFAULT_VIEW_ONLY}}
-: ${EXEC_AUTH:=${DEFAULT_EXEC_AUTH}}
+: ${EXEC_AUTH_JSON:=${DEFAULT_EXEC_AUTH_JSON}}
 
 DRY_RUN_ARG="--dry-run=none"
 
@@ -246,20 +246,12 @@ create_kiali_remote_cluster_secret() {
     error "The remote cluster name [${REMOTE_CLUSTER_NAME}] does not conform to Kubernetes rules for secret key data. Use --remote-cluster-name to specify a name that matches the regex '^[-._a-zA-Z0-9]+$'"
   fi
 
-  if [ "${EXEC_AUTH}" == "true" ]; then
-    local user_auth="exec:
-          apiVersion: client.authentication.k8s.io/v1beta1
-          command: echo
-          args:
-          - |
-            {
-              \"apiVersion\": \"client.authentication.k8s.io/v1beta1\",
-              \"kind\": \"ExecCredential\",
-              \"status\": {
-                \"token\": "\"$TOKEN\""
-              }
-            }
-          interactiveMode: IfAvailable"
+  if [ "${EXEC_AUTH_JSON}" != "" ]; then
+    local user_auth=$(cat <<EOF
+env:
+$(echo ${EXEC_AUTH_JSON} | yq -P | sed "s/^/  /g")
+EOF
+)
   else
     local user_auth="token: ${TOKEN}"
   fi
@@ -289,7 +281,7 @@ stringData:
     users:
     - name: ${REMOTE_CLUSTER_NAME}
       user:
-        ${user_auth}
+$(echo "${user_auth}" | sed "s/^/        /g")
     clusters:
     - name: ${REMOTE_CLUSTER_NAME}
       cluster:
@@ -382,9 +374,8 @@ while [ $# -gt 0 ]; do
       VIEW_ONLY="$2"
       shift;shift
       ;;
-    -ea|--exec-auth)
-      [ "${2:-}" != "true" -a "${2:-}" != "false" ] && error "--exec-auth must be 'true' or 'false'"
-      EXEC_AUTH="$2"
+    -eaj|--exec-auth-json)
+      EXEC_AUTH_JSON="$2"
       shift;shift
       ;;
     -h|--help)
@@ -481,11 +472,13 @@ Valid command line arguments:
   -vo|--view-only: if 'true' then the created service account/remote secret
                    will only provide a read-only view of the remote cluster.
                    Default: "${DEFAULT_VIEW_ONLY}"
-  -ea|--exec-auth: If 'true', use the mocked exec auth for remote secret authentication.
-                   To use this option, kiali's 'auth.strategy' must be changed to 'anonymous'.
-                   (e.g. helm upgrade -n istio-system --set auth.strategy="anonymous" \\
-                         kiali-server kiali/kiali-server)
-                   Default: "${DEFAULT_EXEC_AUTH}"
+  -eaj|--exec-auth-json: If you want to use exec auth for authentication, 
+                         specify ExecConfig in clientcmd/v1 in json format
+                         To use this option, kiali's 'auth.strategy' must be 
+                         changed to 'anonymous'. 'yq' command is required.
+                         (e.g. helm upgrade -n istio-system \\
+                         --set auth.strategy="anonymous" kiali-server kiali/kiali-server)
+                         Default: "${DEFAULT_EXEC_AUTH_JSON}"
   -h|--help: this text.
 HELPMSG
       exit 1
@@ -514,7 +507,7 @@ info REMOTE_CLUSTER_CONTEXT=${REMOTE_CLUSTER_CONTEXT}
 info REMOTE_CLUSTER_NAME=${REMOTE_CLUSTER_NAME}
 info REMOTE_CLUSTER_NAMESPACE=${REMOTE_CLUSTER_NAMESPACE}
 info VIEW_ONLY=${VIEW_ONLY}
-info EXEC_AUTH=${EXEC_AUTH}
+info EXEC_AUTH_JSON=${EXEC_AUTH_JSON}
 
 #
 # Main processing - get some additional information we need and then start creating (or deleting) resources.
--- a/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
+++ b/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
@@ -36,6 +36,7 @@ DEFAULT_REMOTE_CLUSTER_CONTEXT="west"
 DEFAULT_REMOTE_CLUSTER_NAME=""
 DEFAULT_REMOTE_CLUSTER_NAMESPACE="kiali-access-ns"
 DEFAULT_VIEW_ONLY="true"
+DEFAULT_EXEC_AUTH="false"
 
 : ${ALLOW_SKIP_TLS_VERIFY:=${DEFAULT_ALLOW_SKIP_TLS_VERIFY}}
 : ${CLIENT_EXE:=${DEFAULT_CLIENT_EXE}}
@@ -51,6 +52,7 @@ DEFAULT_VIEW_ONLY="true"
 : ${REMOTE_CLUSTER_NAMESPACE:=${DEFAULT_REMOTE_CLUSTER_NAMESPACE}}
 : ${REMOTE_CLUSTER_NAME:=${DEFAULT_REMOTE_CLUSTER_NAME}}
 : ${VIEW_ONLY:=${DEFAULT_VIEW_ONLY}}
+: ${EXEC_AUTH:=${DEFAULT_EXEC_AUTH}}
 
 DRY_RUN_ARG="--dry-run=none"
 
@@ -244,6 +246,24 @@ create_kiali_remote_cluster_secret() {
     error "The remote cluster name [${REMOTE_CLUSTER_NAME}] does not conform to Kubernetes rules for secret key data. Use --remote-cluster-name to specify a name that matches the regex '^[-._a-zA-Z0-9]+$'"
   fi
 
+  if [ "${EXEC_AUTH}" == "true" ]; then
+    local user_auth="exec:
+          apiVersion: client.authentication.k8s.io/v1beta1
+          command: echo
+          args:
+          - |
+            {
+              \"apiVersion\": \"client.authentication.k8s.io/v1beta1\",
+              \"kind\": \"ExecCredential\",
+              \"status\": {
+                \"token\": "\"$TOKEN\""
+              }
+            }
+          interactiveMode: IfAvailable"
+  else
+    local user_auth="token: ${TOKEN}"
+  fi
+
   KIALI_SECRET_YAML=$(cat <<EOF
 ---
 apiVersion: v1
@@ -269,7 +289,7 @@ stringData:
     users:
     - name: ${REMOTE_CLUSTER_NAME}
       user:
-        token: ${TOKEN}
+        ${user_auth}
     clusters:
     - name: ${REMOTE_CLUSTER_NAME}
       cluster:
@@ -360,6 +380,11 @@ while [ $# -gt 0 ]; do
     -vo|--view-only)
       [ "${2:-}" != "true" -a "${2:-}" != "false" ] && error "--view-only must be 'true' or 'false'"
       VIEW_ONLY="$2"
+      shift;shift
+      ;;
+    -ea|--exec-auth)
+      [ "${2:-}" != "true" -a "${2:-}" != "false" ] && error "--exec-auth must be 'true' or 'false'"
+      EXEC_AUTH="$2"
       shift;shift
       ;;
     -h|--help)
@@ -456,6 +481,11 @@ Valid command line arguments:
   -vo|--view-only: if 'true' then the created service account/remote secret
                    will only provide a read-only view of the remote cluster.
                    Default: "${DEFAULT_VIEW_ONLY}"
+  -ea|--exec-auth: If 'true', use the mocked exec auth for remote secret authentication.
+                   To use this option, kiali's 'auth.strategy' must be changed to 'anonymous'.
+                   (e.g. helm upgrade -n istio-system --set auth.strategy="anonymous" \\
+                         kiali-server kiali/kiali-server)
+                   Default: "${DEFAULT_EXEC_AUTH}"
   -h|--help: this text.
 HELPMSG
       exit 1
@@ -484,6 +514,7 @@ info REMOTE_CLUSTER_CONTEXT=${REMOTE_CLUSTER_CONTEXT}
 info REMOTE_CLUSTER_NAME=${REMOTE_CLUSTER_NAME}
 info REMOTE_CLUSTER_NAMESPACE=${REMOTE_CLUSTER_NAMESPACE}
 info VIEW_ONLY=${VIEW_ONLY}
+info EXEC_AUTH=${EXEC_AUTH}
 
 #
 # Main processing - get some additional information we need and then start creating (or deleting) resources.
--- a/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
+++ b/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
@@ -248,7 +248,7 @@ create_kiali_remote_cluster_secret() {
 
   if [ "${EXEC_AUTH_JSON}" != "" ]; then
     local user_auth=$(cat <<EOF
-env:
+exec:
 $(echo ${EXEC_AUTH_JSON} | yq -P | sed "s/^/  /g")
 EOF
 )
--- a/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
+++ b/hack/istio/multicluster/kiali-prepare-remote-cluster.sh
@@ -249,7 +249,7 @@ create_kiali_remote_cluster_secret() {
   if [ "${EXEC_AUTH_JSON}" != "" ]; then
     local user_auth=$(cat <<EOF
 exec:
-$(echo ${EXEC_AUTH_JSON} | yq -P | sed "s/^/  /g")
+$(echo "${EXEC_AUTH_JSON}" | yq -P | sed "s/^/  /g")
 EOF
 )
   else
@@ -507,7 +507,7 @@ info REMOTE_CLUSTER_CONTEXT=${REMOTE_CLUSTER_CONTEXT}
 info REMOTE_CLUSTER_NAME=${REMOTE_CLUSTER_NAME}
 info REMOTE_CLUSTER_NAMESPACE=${REMOTE_CLUSTER_NAMESPACE}
 info VIEW_ONLY=${VIEW_ONLY}
-info EXEC_AUTH_JSON=${EXEC_AUTH_JSON}
+info EXEC_AUTH_JSON="${EXEC_AUTH_JSON}"
 
 #
 # Main processing - get some additional information we need and then start creating (or deleting) resources.
--- a/kubernetes/client.go
+++ b/kubernetes/client.go
@@ -77,10 +77,34 @@ func (client *K8SClient) GetToken() string {
 	return client.token
 }
 
+func SetUserIdentificationFromRemoteSecretUser(config *rest.Config, user *RemoteSecretUser) {
+	if len(user.User.Token) > 0 {
+		config.BearerToken = user.User.Token
+	}
+	exec := user.User.Exec
+	if exec != nil {
+		config.ExecProvider = &api.ExecConfig{
+			Command:            exec.Command,
+			Args:               exec.Args,
+			Env:                exec.Env,
+			APIVersion:         exec.APIVersion,
+			InstallHint:        cleanANSIEscapeCodes(exec.InstallHint),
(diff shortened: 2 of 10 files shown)


jmazzitelli:
@frauniki will this still require these? (well, maybe not "require", but perhaps "would be better")

The sidecar injection config: https://github.com/kiali/kiali/issues/6193#issuecomment-1570274350

frauniki:
In Istio, if you want to use an authentication method other than oidc and token in the remote secret of a multi-cluster, you need to permit the authentication method you want to use in advance for security reasons through the [INSECUER option](https://istio.io/latest/news/releases/1.13.x/announcing-1.13/upgrade-notes/#multicluster-secret-authentication-changes).
My concern here is whether we should have a similar option for kiali as well.
Could you give me your opinion on this?

jmazzitelli:
> My concern here is whether we should have a similar option for kiali as well.

Off the top of my head, I don't think we need that unless, for some reason, allowing custom auth methods would introduce a security loophole somewhere that someone could exploit. I don't know enough about this to know what an exec auth would entail.

frauniki:
When using exec auth with Kubernetes, the go-client executes the specified external Command and attempts to retrieve credentials from its standard output.
Since kiali can auto-detect remote secret, it is possible to have kiali execute arbitrary commands on kiali by adding a remote secret.
I am not a security expert, so I do not know if this arbitrary command execution poses a security threat, but at least blocking exec auth in the default configuration may improve security by preventing kiali from executing arbitrary commands unless the user explicitly specifies the option This may improve security by preventing kiali from executing arbitrary commands unless the user explicitly specifies the option.

However, in order to execute arbitrary commands, the binary you want to execute must exist in the pod beforehand, so if the pod's file system is set to read-only, the security threat may be lower.
And if there were hypothetically a case where a malicious user could add a remote secret at will, I think we would already have a serious security incident, so it may be nonsense to consider such a case only for KIRIALI.

If there are no additional comments, I will skip adding the INSECURE option.

jmazzitelli:
> However, in order to execute arbitrary commands, the binary you want to execute must exist in the pod beforehand, so if the pod's file system is set to read-only, the security threat may be lower.
> And if there were hypothetically a case where a malicious user could add a remote secret at will, I think we would already have a serious security incident, so it may be nonsense to consider such a case only for kiali.

File system is read-only by default and the only way to configure it not that way is to flip the operator switch to allow custom security context - see https://github.com/kiali/kiali-operator/blob/master/roles/default/kiali-deploy/templates/kubernetes/deployment.yaml#L62-L65

And Kiali server's base container is the ubi-micro "distroless" container- there are very little binaries that ship in there: see https://github.com/kiali/kiali/blob/master/deploy/docker/Dockerfile-multi-arch-distroless#L1-L4

frauniki:
@jmazzitelli 
Thank you very much.

I have verified that this PR exec auth works fine using AWS EKS.
Can you please review this PR as a next step?

jmazzitelli:
To test this doesn't break token-based OpenID auth, I ran a test using our multicluster hack scripts.

1. Pull this PR and build the dev image: `make clean build build-ui container-build-kiali`
2. Install the multicluster setup using the dev images: `hack/istio/multicluster/install-multi-primary.sh --client-exe kubectl -dorp podman --kiali-enabled true --single-kiali false --kiali-create-remote-cluster-secrets true --kiali-use-dev-image true`
3. After everything is up and running, log into both Kialis (using credentials `kiali`/`kiali`) that are running in the two minikube clusters:

To open the east cluster Kiali UI:
```
xdg-open https://$(minikube ip --profile east)/kiali/console
```
to open the west cluster Kiali UI:
```
xdg-open https://$(minikube ip --profile west)/kiali/console
```

kubernetes/client.go
```diff
@@ -77,10 +77,34 @@ func (client *K8SClient) GetToken() string {
...
+	if exec != nil {
+		config.ExecProvider = &api.ExecConfig{
+			Command:            exec.Command,
+			Args:               exec.Args,
+			Env:                exec.Env,
+			APIVersion:         exec.APIVersion,
+			InstallHint:        cleanANSIEscapeCodes(exec.InstallHint),
+			ProvideClusterInfo: exec.ProvideClusterInfo,
+			InteractiveMode:    exec.InteractiveMode,
+		}
```
jmazzitelli, line 94: We are going to want a log message to indicate an ExecProvider has been detected, at minimum for auditing purposes.

```suggestion
		}
		log.Debugf("Auth ExecProvider has been detected: cmd=[%v], args=%v", exec.Command, exec.Args)
```

jmazzitelli (changes requested):

frauniki:
@jmazzitelli 
> Can I ask that you provide some quick documentation or "instructions" on how this is expected to be configured and used (even if its just a comment here in this PR)?

First, setup a multi cluster
```shell
$ cd hack/istio/multicluster/

$ ./install-multi-primary.sh
```

Since exec auth is only supported when the auth strategy is anonymous, use helm upgrade to change the auth strategy.
Then you need to change the kiali image to the one that contains this PR.
I uploaded the image of this PR to my dockerhub repository. 
You can use 'frauniki/kiali:dev' if you want.
```shell
$ helm upgrade -n istio-system --set auth.strategy="anonymous" --set deployment.image_name=frauniki/kiali --set deployment.image_version=dev kiali-server kiali/kiali-server
```

Finally, change kiali's remote secret to exec auth.
I have added support for exec auth to 'kiali-prepare-remote-cluster.sh' for this purpose.
The following command will change the remote secret and use the mocked exec auth in the echo command.
Originally the exec auth command retrieves authentication information via other external systems, but this time the echo command is used to return static token information.
```shell
$ ./kiali-prepare-remote-cluster.sh --exec-auth true

$ kubectl get secret kiali-remote-cluster-secret-west -n istio-system -o jsonpath='{.data.west}' | base64 -d | grep users: -A 15
users:
- name: west
  user:
    exec:
      apiVersion: client.authentication.k8s.io/v1beta1
      command: echo
      args:
      - |
        {
          "apiVersion": "client.authentication.k8s.io/v1beta1",
          "kind": "ExecCredential",
          "status": {
            "token": "token"
          }
        }
      interactiveMode: IfAvailable
```

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
```diff
@@ -456,6 +481,11 @@ Valid command line arguments:
   -vo|--view-only: if 'true' then the created service account/remote secret
                    will only provide a read-only view of the remote cluster.
                    Default: "${DEFAULT_VIEW_ONLY}"
+  -ea|--exec-auth: If 'true', use the mocked exec auth for remote secret authentication.
```
jmazzitelli, line 484: this hack script is going to be used (documented as a helper script) by people that will be using the new multi-cluster features in Kiali (a blog post will be available shortly introducing this stuff). So even though this is in a "hack" directory, it really is going to be used by people other than just kiali devs (think of this as the poor man's "istioctl create-remote-secret").

So, let's not put "mock" things in here. Let's change this option to something like this (unless you can think of a better way).... Change it to `--exec-auth-json` which takes a file path - that file contains the json you want to inject into the secret's kubeconfig. This way, people can use this "for real" to inject their actual config.

jmazzitelli (changes requested):

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
frauniki: Removed mock from hack script and added --exec-auth-json option instead.
Slightly different from the comments, the --exec-auth-json option does not require a file path, but an actual json payload.
This allows more flexibility when using scripts.
If you want to use file paths, you can use scripts like this.
`./kiali-prepare-remote-cluster.sh -eaj "$(cat . /exec-auth.json)"`

jmazzitelli:
(just an FYI: I will be reviewing this; I'm still in discussions with some security experts. It doesn't look like this would increase security vulnerability, however, to assuage my paranoia, I may want to ask that some kind of "disable" flag be incorporated here somewhere just as a safety valve. I'm still not sure it is needed, but just as a heads up, we should consider how we could disable this.

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
```diff
@@ -244,6 +246,16 @@ create_kiali_remote_cluster_secret() {
     error "The remote cluster name [${REMOTE_CLUSTER_NAME}] does not conform to Kubernetes rules for secret key data. Use --remote-cluster-name to specify a name that matches the regex '^[-._a-zA-Z0-9]+$'"
   fi
 
+  if [ "${EXEC_AUTH_JSON}" != "" ]; then
+    local user_auth=$(cat <<EOF
+env:
+$(echo ${EXEC_AUTH_JSON} | yq -P | sed "s/^/  /g")
```
jmazzitelli, line 252: Without wrapping in quotes, you get this:

`Error: bad file '-': yaml: mapping values are not allowed in this context`

To see this, create a `/tmp/exec-auth.yaml` with:

```
exec:
  apiVersion: client.authentication.k8s.io/v1beta1
  command: echo
  args:
  - |
    {
      "apiVersion": "client.authentication.k8s.io/v1beta1",
      "kind": "ExecCredential",
      "status": {
        "token": "token"
      }
    }
  interactiveMode: IfAvailable
```

Then run
```
hack/istio/multicluster/kiali-prepare-remote-cluster.sh -dr true -eaj "$(cat /tmp/exec-auth.yaml)"
```

To fix:

```suggestion
$(echo "${EXEC_AUTH_JSON}" | yq -P | sed "s/^/  /g")
```

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
```diff
@@ -484,6 +507,7 @@ info REMOTE_CLUSTER_CONTEXT=${REMOTE_CLUSTER_CONTEXT}
 info REMOTE_CLUSTER_NAME=${REMOTE_CLUSTER_NAME}
 info REMOTE_CLUSTER_NAMESPACE=${REMOTE_CLUSTER_NAMESPACE}
 info VIEW_ONLY=${VIEW_ONLY}
+info EXEC_AUTH_JSON=${EXEC_AUTH_JSON}
```
jmazzitelli, line 510: Wrap in quotes, otherwise only the first line is logged.

```suggestion
info EXEC_AUTH_JSON="${EXEC_AUTH_JSON}"
```

jmazzitelli (changes requested):
fix the util script to support passing in file contents

jmazzitelli:
I played around with using the util script to add a exec-auth json section to the secret and then have it create a secret without one. When mocking the exec-auth section, I see the cache fails to be able to look up resources because it really isn't authenticated (because my mock isn't really valid - that echo "token"). When I removed it, it started working again. So at least this shows  Kiali is doing "something" with that exec auth stuff :) I don't have a system set up that properly authenticates using that stuff, but @frauniki does and has verified this works properly.

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
jmazzitelli: @frauniki don't do this if the "EXEC_AUTH_JSON" content is supposed to have real credentials. We don't want things like passwords or keys being printed here if they are sensitive. But if this exec auth section isn't anything more than setting up a cmd line and its arguments, then we should want to output it all, not just the first line.

jmazzitelli:
I submitted a PR to this PR: https://github.com/frauniki/kiali/pull/1 -- that adds a new hidden feature flag that allows you to turn on support for ExecProvider. It is disabled by default for security purposes.

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
jmazzitelli: this fix is included in my [PR](https://github.com/frauniki/kiali/pull/1)

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
jmazzitelli: this fix is included in my [PR](https://github.com/frauniki/kiali/pull/1)

nrfox:
> I don't have a system set up that properly authenticates using that stuff, but @frauniki does and has verified this works properly.

Is there a way we can test this in an automated way? It'd be nice to have an e2e test or a molecule test for this to ensure it's still working.

jmazzitelli:
> Is there a way we can test this in an automated way? It'd be nice to have an e2e test or a molecule test for this to ensure it's still working.

The difficulty here is (a) this involves multi-cluster setup (which we do not have automated tests for) and (b) the ExecProvider defined in the remote secret needs to output actual valid credentials for a remote cluster (so we would need to produce a remote cluster secret that has an ExecProvider that at least "echo"s proper credentials for the test remote cluster.

Right now I don't know of how to mock any of that.

@nrfox as part of all this multi-cluster work we've been doing, has anyone done any development of e2e tests? I know we don't have any molecule tests for it.

jmazzitelli:
@frauniki I need some more clarification re: your statement from earlier: `Since exec auth is only supported when the auth strategy is anonymous`

The remote cluster secrets directly provide credentials through the ExecProvider. So I assume this means we do not forward users' credentials when they navigate to portions of the Kiali UI that need to access remote clusters.

But what about the local cluster? The one where Kiali is hosted? There is no remote cluster secret for the local cluster - Kiali uses the user's credentials directly when accessing the local cluster k8s master API. With `anonymous` auth strategy being _required_ that means any user (authenticated or not) will be able to access the local cluster details in Kiali.

Are we sure `anonymous` is required in order to support this PR's ExecProvider feature?

This is something we definitely do not want. So we are going to need to remove this requirement `exec auth is only supported when the auth strategy is anonymous` - we need this to work with other auth strategies, particularly `openid` (because right now - and @nrfox, please correct me here - Kiali supports both `openid` and `anonymous` when in a multi-cluster setup, with `openshift` to be supported in the near future).

frauniki:
> @frauniki I need some more clarification re: your statement from earlier: `Since exec auth is only supported when the auth strategy is anonymous`
> 
> The remote cluster secrets directly provide credentials through the ExecProvider. So I assume this means we do not forward users' credentials when they navigate to portions of the Kiali UI that need to access remote clusters.
> 
> But what about the local cluster? The one where Kiali is hosted? There is no remote cluster secret for the local cluster - Kiali uses the user's credentials directly when accessing the local cluster k8s master API. With `anonymous` auth strategy being _required_ that means any user (authenticated or not) will be able to access the local cluster details in Kiali.
> 
> Are we sure `anonymous` is required in order to support this PR's ExecProvider feature?
> 
> This is something we definitely do not want. So we are going to need to remove this requirement `exec auth is only supported when the auth strategy is anonymous` - we need this to work with other auth strategies, particularly `openid` (because right now - and @nrfox, please correct me here - Kiali supports both `openid` and `anonymous` when in a multi-cluster setup, with `openshift` to be supported in the near future).

I do not fully understand Kiali, so I apologize if I have made any crazy statements.

The reason I said that anonymous is required for ExecProvider is because, as you said, Kiali currently supports openid and anonymous in a multi-cluster environment, and if openid is used, it is expected that ExecProvider will be ignored.

Looking at the part where the remote cluster client is created, it appears that if you use an auth strategy other than anonymous, the user's token is used in the ClientConfig.
https://github.com/kiali/kiali/blob/34b889b52775c303051852026066abb57488a6a6/kubernetes/client_factory.go#L213-L216
In such a case, if both token and ExecProvider are specified in the kubeconfig, kubernetes/go-client will use the token first. (This is a kubernetes/go-client specification.)
So, in effect, the auth strategy in openid will always ignore the ExecProvider setting.
Thus, I thought that ExecProvider in kiali would only work correctly with anonymous strategy.
I can add the ExecProvider setting in openid as well, but it is unnecessary and never really works.

This is my understanding.
I would be happy to point out if I am wrong in any way.

nrfox:
> Looking at the part where the remote cluster client is created, it appears that if you use an auth strategy other than anonymous, the user's token is used in the ClientConfig.

That's correct. There's two different sets of credentials within a Kiali request: the user's credentials and the Kiali Service Account's credentials. The Kiali server uses the Kiali Service Account credentials to talk to the kube API in nearly all *read* operations. Then it uses the user's credentials to check that the user has access to a namespace before showing them anything from that namespace and it uses the user's credentials when performing any *write* operations. Anonymous mode uses the Kiali Service Account credentials for everything since there is no login and no user credentials.

This `ExecProvider` option would just be updating how the Kiali Service Account credentials are obtained correct? I think you could still have other auth modes for obtaining user credentials but use the `ExecProvider` for the Kiali Service Account credentials?

nrfox:
> The difficulty here is (a) this involves multi-cluster setup (which we do not have automated tests for) and (b) the ExecProvider defined in the remote secret needs to output actual valid credentials for a remote cluster (so we would need to produce a remote cluster secret that has an ExecProvider that at least "echo"s proper credentials for the test remote cluster.

Is this just for multi-cluster use cases? Couldn't you also use the `ExecProvider` in single cluster mode?

jmazzitelli:
> Is this just for multi-cluster use cases? Couldn't you also use the `ExecProvider` in single cluster mode?

Well, the ExecProvider details only comes from remote cluster secrets, and we only have remote cluster secrets for .... remote clusters :) We don't have a secret containing kubeconfig for the local cluster so there is no ExecProvider details for the local cluster because there is no place to specify it.

jmazzitelli:
@frauniki maybe you can describe your AWS use-case in more detail? Because @nrfox brings up a good point.

Suppose you have a mesh on only a single AWS cluster ... how does Kiali work in your use-case where you need ExecProvider to provide credentials for that local cluster?

jmazzitelli:
> it appears that if you use an auth strategy other than anonymous, the user's token is used in the ClientConfig.

> In such a case, if both token and ExecProvider are specified in the kubeconfig, kubernetes/go-client will use the token first.

I believe that is what we want. So all that is good.

frauniki:
> Suppose you have a mesh on only a single AWS cluster ... how does Kiali work in your use-case where you need ExecProvider to provide credentials for that local cluster?

If you want to use ExecProvider on a single EKS Cluster, there is no particular difference from the multi-cluster case.
For example, if you want to use ExecProvider on EKS, configure the following in Kubeconfig.
The only information required here is the EKS Cluster name.
By passing only the EKS Cluster name to the aws command, you can issue a token that can be used by the kubernetes client.
````
users: name: arn:aws:aws:users
- name: eks-user
  user: $region_code:$account_id:cluster/$cluster_name
    exec: $account_id:cluster/$cluster_name
      apiVersion: client.authentication.k8s.io/v1beta1
      command: aws
      args: --region
        --region
        - ap-northeast-1
        - eks
        - get-token
        --cluster-name
        - <EKS Cluster Name>
```

So, if you want to use this with a single EKS Cluster, simply specify the local cluster name in the aws command.

Since ExecProvider is only interested in the token contained in the command response, it should work in any environment as long as the token and server url are correct.

jmazzitelli:
> By passing only the EKS Cluster name to the aws command, you can issue a token that can be used by the kubernetes client. (Of course, the token will only be issued if the kiali execution pod also has the appropriate aws role.)

I still do not 100% understand.

When you say, `if the kiali execution pod also has the appropriate aws role` are you manually modifying the Kiali Service Account and its Role Bindings/Roles?

And how does a user log into Kiali in this case? Kiali does have a `token` auth strategy (as opposed to `anonymous` or `openid`) where you can enter a token string (not a username/password) and Kiali will use that token to connect to the k8s api. But I don't think that is what you are using. So how does your users log into Kiali (i.e. when they encounter the Kiali login page, what do they do?)

Or do you just use `anonymous` auth strategy and simply rely on Kiali to use that manually modified Kiali Service Account to connect to the k8s api using that single AWS generated token?

nrfox:
> And how does a user log into Kiali in this case? Kiali does have a `token` auth strategy (as opposed to `anonymous` or `openid`) where you can enter a token string (not a username/password) and Kiali will use that token to connect to the k8s api. But I don't think that is what you are using. So how does your users log into Kiali (i.e. when they encounter the Kiali login page, what do they do?)

My take on this is that this exec provider config would just be for the `Kiali Service Account's credentials` and not for the user's credentials. There's currently no way for the UI to interact with this exec plugin and I don't think you'd want that anyways. This is basically independent of the "auth" strategy since the "auth" strategy is for user's credentials. This would add an additional authentication mechanism for the Kiali Service Account. Right now we only support service account tokens so there's no configuration for this. For the home cluster it doesn't matter much since we just use the in cluster config but for the remote clusters I guess the token is explicitly copied and everything else discarded from the kubeconfig.

jmazzitelli:
> this exec provider config would just be for the `Kiali Service Account's credentials` and not for the user's credentials.

OK, that would make sense to me.

jmazzitelli:
I performed the same MC test as [before](https://github.com/kiali/kiali/pull/6232#issuecomment-1584988968) - still OK.

frauniki:
> this exec provider config would just be for the Kiali Service Account's credentials and not for the user's credentials.

Yes, this ExecProvider is not intended to be used as user credentials.

frauniki:
There was a partial problem in the process of generating remote secret in kiali-prepare-remote-cluster.sh, which did not generate the correct ExecProvider config.
This problem has been fixed in this commit.
https://github.com/kiali/kiali/pull/6232/commits/abe6e95e2dd2622dd0960e10998175ba254209e3

I built the latest commit of this PR and updated my DockerHub image.
With the fixed hack script, you can try the mocked ExecProvider with the following command.
```bash
$ ./kiali-prepare-remote-cluster.sh -eaj "{
    \"apiVersion\":\"client.authentication.k8s.io/v1beta1\",
    \"command\":\"echo\",
    \"args\":[
       \"{\\\"apiVersion\\\":\\\"client.authentication.k8s.io/v1beta1\\\",\\\"kind\\\":\\\"ExecCredential\\\",\\\"status\\\":{\\\"token\\\":\\\"$(kubectl --context=west get secrets -n kiali-access-ns kiali-remote-access -o jsonpath='{.data.token}' | base64 -d)\\\"}}\"
    ],
    \"interactiveMode\":\"IfAvailable\"
 }"
```

And I have also incorporated @jmazzitelli's suggestion (https://github.com/frauniki/kiali/pull/1) into this PR.`
I tried enable_exec_provider and it seemed to work fine.

```bash
# enable_exec_provider=true
$ helm upgrade -n istio-system --set auth.strategy="anonymous" --set deployment.image_name=frauniki/kiali --set deployment.image_version=dev --set kiali_feature_flags.clustering.enable_exec_provider=true kiali-server kiali/kiali-server

# enable_exec_provider=false
$ helm upgrade -n istio-system --set auth.strategy="anonymous" --set deployment.image_name=frauniki/kiali --set deployment.image_version=dev --set kiali_feature_flags.clustering.enable_exec_provider=false kiali-server kiali/kiali-server
```

jmazzitelli:
@frauniki the CI test failure does not seem to have anything to do with this PR, but can you rebase this on master and let's see if we can get "all green" tests? This test failure might be a flaky test, but I seem to think something was changed in a recent commit to master that might fix it.

kubernetes/secret.go
```diff
@@ -18,12 +19,23 @@ type RemoteSecretClusterListItem struct {
...
+
+type RemoteSecretUserExec struct {
+	Command            string                  `yaml:"command"`
+	Args               []string                `yaml:"args"`
+	Env                []api.ExecEnvVar        `yaml:"env"`
+	APIVersion         string                  `yaml:"apiVersion"`
+	InstallHint        string                  `yaml:"installHint"`
+	ProvideClusterInfo bool                    `yaml:"provideClusterInfo"`
+	InteractiveMode    api.ExecInteractiveMode `yaml:"interactiveMode"`
 }
```
nrfox, line 39: Do we need this struct? Can we just use `k8s.io/client-go/tools/clientcmd/api.ExecConfig` directly instead of having to copy over all the struct fields and some defaulting code?

kubernetes/client_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/kiali/kiali/config"
+)
+
+func TestSetUserIdentificationFromRemoteSecretUser(t *testing.T) {
```
nrfox, line 13: Thanks for providing good tests for this!

nrfox (commented):
One question but not blocking otherwise LGTM

kubernetes/secret.go
nrfox: I think it's fine to include this as is and we can look at cleaning it up in a later change.

nrfox (approved):

jmazzitelli:
Everything is green. Discussed this with security folks. While this could, obviously, be exploited given high enough privileges, it seems if someone has those privileges, they would be able to do worse things already even without this feature.

Merging so it gets in the next release.

Thanks, @frauniki for the help.

We may want to look to include [that sidecar feature](https://github.com/kiali/kiali/issues/5028) in the future - but let's see how this feature works out first.

isarns:
Could you please provide an example of using this feature with two AWS EKS clusters?

Issue #6233, opened 2023-06-05
Test single cluster mode with external links


nrfox:
Ensure that the [existing experimental multi-cluster single control plane support](https://kiali.io/docs/features/multi-cluster/) still works as intended with the addition of the unified deployment.

jmazzitelli:
@nrfox just so I'm clear - when you say "multi-cluster single control plane support" - this is referring to one Istio control plane with one Kiali installed, with the data plane across multiple clusters? (i.e. primery-remote)?

This issue with the Kiali links showing in the mesh page was really to show links to multiple Kialis in the multi-primary deployment scenario (so you can jump from one Kiali/control plane to another by clinking the link).

In short, I'm confused what this is asking for. :)

nrfox:
> This issue with the Kiali links showing in the mesh page was really to show links to multiple Kialis in the multi-primary deployment scenario (so you can jump from one Kiali/control plane to another by clinking the link).

This issue is to test that this setup still works: multiple kialis with one per controlplane as opposed a single "unified" kiali.

jmazzitelli:
> This issue is to test that this setup still works: multiple kialis with one per controlplane as opposed a single "unified" kiali.

OK, so testing with multi-primary mode, with kiali installed in each control plane.

I'm going to test with this:

```
hack/istio/multicluster/install-multi-primary.sh --client-exe kubectl -dorp podman --kiali-enabled true --single-kiali false
```
My predication is it won't work until I manually create the kiali remote secrets (that hack script with `--single-kiali` set to false won't create the remote secrets, if I read the code correct). I'm not sure if that is how we want it to work - we should not need the kiali remote secrets when installing in multi-primary/multi-kiali mode, but we also should not require permissions to read the istio secrets either (but that is probably going to need to be if we want these links to show up - the URLs come from data read from the secrets).

But let me play with this to see what really happens.

nrfox:
> I'm not sure if that is how we want it to work - we should not need the kiali remote secrets when installing in multi-primary/multi-kiali mode, but we also should not require permissions to read the istio secrets either (but that is probably going to need to be if we want these links to show up - the URLs come from data read from the secrets).

[...]

PR #6235, opened 2023-06-06
[hack] hack script provides option to install remote cluster secrets

jmazzitelli:
this allows you to install the secrets regardless of whether you are installing one or two kiali instances

This is for the multi-cluster hack scripts.


jmazzitelli:
I needed this to test https://github.com/kiali/kiali/issues/6233

hack/istio/multicluster/deploy-kiali.sh
```diff
@@ -75,6 +81,7 @@ deploy_kiali() {
     --set auth.strategy="openid"                  \
     --set auth.openid.client_id="kube"            \
     --set auth.openid.issuer_uri="https://${kube_hostname}/realms/kube" \
+    --set auth.openid.insecure_skip_verify_tls="true" \
```
nrfox, line 84: This doesn't seem related to this change?

jmazzitelli: I needed that otherwise I was having other keycloak issues.

jmazzitelli: ok, I set that back explicitly to false (which is the default).

nrfox (approved):

PR #6238, opened 2023-06-07
Add both clusters to realm config

nrfox:
Adds both clusters to the redirect uris for the Kiali app so that both east and west kialis can be accessed.

### DIFF
--- a/hack/istio/multicluster/start-minikube.sh
+++ b/hack/istio/multicluster/start-minikube.sh
@@ -140,24 +140,10 @@ ${CLIENT_EXE} create secret tls -n keycloak keycloak.kind.cluster-tls \
   --cert="${KEYCLOAK_CERTS_DIR}"/cert.pem \
   --key="${KEYCLOAK_CERTS_DIR}"/key.pem
 
-# Give the ingress some time to be ready
-# ${CLIENT_EXE} wait ingress/keycloak -n keycloak --context east --for=jsonpath='{.status.loadBalancer.ingress[*].hostname}'=localhost
-sleep 20
+# Before proceeding with the rest of the keycloak setup, we need to start the second cluster so that we can get the IP
+# and add it to the redirect URI of the kube client in keycloak.
 
-# Get a token from keycloak to use the admin api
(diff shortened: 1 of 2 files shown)


hack/istio/multicluster/start-minikube.sh
```diff
@@ -153,25 +172,13 @@ TOKEN_KEY=$(curl -k -X POST https://"${KUBE_HOSTNAME}"/realms/master/protocol/op
             -d scope=openid \
             -d response_type=id_token | jq -r '.access_token')
 
+# "https://_KIALI_KUBE_HOSTNAME/kiali/*`
 # Replace the redirect URI with the minikube ip. Create the realm.
-sed "s/_KIALI_KUBE_HOSTNAME/$(minikube ip -p "${CLUSTER1_NAME}")/g" "${SCRIPT_DIR}"/realm-export-template.json | curl -k -L https://"${KUBE_HOSTNAME}"/admin/realms -H "Authorization: Bearer $TOKEN_KEY" -H "Content-Type: application/json" -X POST -d @-
+jq ".clients[] |= if .clientId == \"kube\" then .redirectUris = [\"https://${CLUSTER1_IP}/kiali/*\", \"https://${CLUSTER2_IP}/kiali/*\"] else . end" < "${SCRIPT_DIR}"/realm-export-template.json | curl -k -L https://"${KUBE_HOSTNAME}"/admin/realms -H "Authorization: Bearer $TOKEN_KEY" -H "Content-Type: application/json" -X POST -d @-
```
jmazzitelli, line 177: I'm confused what KUBE_HOSTNAME is... we have two minikubes here - which one does this variable represent? Should we rename this variable to MINIKUBE_HOSTNAME1 to make it clear this is the hostname of the first minikube  (which I assume it is?)

nrfox: Yeah it made more sense when there was just one. It's really the `KEYCLOAK_KUBE_HOSTNAME`. I'll rename it to that.

jmazzitelli (approved):

PR #6242, opened 2023-06-07
Bookinfo gherkin annotations for upcoming hook

matejnesuta:
This is related to the OSSM-4049 ticket. 

In order to make more robust setup, we want Cypress to detect if certain demoapps are installed before related tests. For example, if the test uses bookinfo, then all of the 6 bookinfo pods and traffic generator must be installed. If they are not or if they are malformed at some way (for example missing pod), then Cypress installs the specific demo app via the hack scripts. 

To test this:
- Deploy Kiali without the bookinfo demo app
- Run any Cypres feature file, which has the `@bookino-app` tag.
- Bookinfo demo app should be installed prior to the first test and all tests related to bookinfo from that FeatureFile should pass
- This should work both for the OCP and non-OCP.

(Drafting this with bookinfo app only. I will turn it into a regular PR once I add other demos as well.

### DIFF
--- a/frontend/cypress/integration/common/hooks.ts
+++ b/frontend/cypress/integration/common/hooks.ts
@@ -9,39 +9,3 @@ Before({tags: '@gateway-api'}, async function () {
     }
   })
 });
-
-Before({tags: '@bookinfo-app'}, async function () {
-  cy.exec(`kubectl get pods -n bookinfo -o=custom-columns=NAME:.metadata.name,Status:.status.phase --no-headers=true --sort-by='.metadata.name' |
-    grep -E -w '(reviews-v(1|2|3))|ratings-v1|details|productpage-v1|kiali-traffic-generator' |
-    awk -F ' ' '{print $2}' |
-    grep Running | 
-    wc -l`,{failOnNonZeroExit: false}).then((result) => {
-    cy.log(result.stdout);
-    if (result.stdout == "7" && result.code == 0){
-      cy.log("Bookinfo demo app is up and running.");
-    } 
-    else{
-      cy.log("Bookinfo demo app is either broken or not present. Installing now.").log("Removing old bookinfo installations.");
-      // is the suite running on openshift?
-      cy.exec('kubectl api-versions | grep --quiet "route.openshift.io";', {failOnNonZeroExit:false}).then((result) =>{
-        if (result.code == 0){
-          cy.exec('../hack/istio/install-bookinfo-demo.sh --delete-bookinfo true').then(()=>{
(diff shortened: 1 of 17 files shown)


jshaughn (commented):
Hi @matejnesuta , This looks good.  Would it make sense to make this something that is done only once for a feature file?  Does adding the annotation on every Scenario slow things down?

ScriptingShrimp:
> done only once for a feature file

the hooks for a "feature file" are in `Scenario:` paragraph of feature file. Definitely we can put some of the feature specific codes inhere. We decided to use conditional hook in this case because i.e. bookinfo is not related to the feature, its just a test precondition shared across multiple feature files.

  

> Does adding the annotation on every Scenario slow things down?

Definitively. But the `cy.exec` is executed outside of web browser context, so the slowdown should not be significant (I believe)

frontend/cypress/integration/common/hooks.ts
```diff
@@ -9,3 +9,39 @@ Before({tags: '@gateway-api'}, async function () {
...
+                    `kubectl wait --for=condition=Ready pods --all -n bookinfo --timeout 60s || true;` +
+                    `kubectl wait --for=condition=Ready pods --all -n bookinfo --timeout 60s;` +
+                    `sleep 80;`,{timeout:400*1000});
+              });
+            })
+          })
+        }
+        else{
+          cy.exec('../hack/istio/install-bookinfo-demo.sh --delete-bookinfo true -c kubectl').then(()=>{
+            cy.exec('../hack/istio/install-bookinfo-demo.sh -c kubectl -tg -in istio-system -a amd64');
```
ScriptingShrimp, line 41: As we discussed, refactoring of shell scripts will be better approach.

ScriptingShrimp (changes requested):
we can merge ` @bookinfo-app` annotations in feature files. 

lets refactor bookinfo installation in another PR if you want

matejnesuta:
I commented out the hook for now, so the annotations can be merged

frontend/cypress/integration/common/hooks.ts
```diff
@@ -9,3 +9,41 @@ Before({tags: '@gateway-api'}, async function () {
     }
   })
 });
+
+// Will be refactored soon
```
ScriptingShrimp, line 13: Please roll back the file to its original state. It's not a good practice to commit TODOs and unused code.

ScriptingShrimp (changes requested):
Just remove the comments from `hooks.ts` and LGTM!
Thank you

matejnesuta:
I removed the comments as you requested

PR #6245, opened 2023-06-07
Add `IsInaccessible` to non-outsider nodes

### LINKED ISSUE #6244: Graph outsider appender - nodes not marked as "IsInaccessible"
Nodes that are not outsiders but are also `IsInaccessible` are not marked as `IsInaccessible`. I think it stems from here which only adds `IsInaccessible` when the node is an outsider: https://github.com/kiali/kiali/blob/ecd24d32d10a3da2747a70997afd8ba409c92ce0/graph/telemetry/istio/appender/outsider.go#L59-L65.

This is causing failures in the health appender which excludes `IsInaccesible` nodes. If inaccesible nodes are included, the health appender will fail when trying to fetch health info for that node.

nrfox:
Labels all nodes that are inaccessible as `IsInaccessible`.

Fixes #6244


nrfox:
When you have access to the "east" cluster but not the "west" cluster your graph will look like this:
Screenshot from 2023-06-07 15-18-10

When you have access to both your graph will look like this:
Screenshot from 2023-06-07 15-19-07

This seems correct to me but maybe the inaccessible nodes should actually be pruned from the graph altogether @jshaughn?

graph/telemetry/istio/appender/outsider.go
```diff
@@ -50,17 +50,16 @@ func (a *OutsiderAppender) markOutsideOrInaccessible(trafficMap graph.TrafficMap
 				if isOutside(n, a.Namespaces) {
 					n.Metadata[graph.IsOutside] = true
 				}
+				if isInaccessible(n, a.AccessibleNamespaces) {
+					n.Metadata[graph.IsInaccessible] = true
+				}
```
jshaughn, line 55: I think you can remove this addition (see the other comment).

graph/telemetry/istio/appender/outsider.go
```diff
@@ -50,17 +50,16 @@ func (a *OutsiderAppender) markOutsideOrInaccessible(trafficMap graph.TrafficMap
...
 				n.Metadata[graph.IsOutside] = true
 			}
-		}
-		if isOutsider, ok := n.Metadata[graph.IsOutside]; ok && isOutsider.(bool) {
-			if _, ok2 := n.Metadata[graph.IsInaccessible]; !ok2 {
-				if isInaccessible(n, a.AccessibleNamespaces) {
-					n.Metadata[graph.IsInaccessible] = true
-				}
+			if isInaccessible(n, a.AccessibleNamespaces) {
+				n.Metadata[graph.IsInaccessible] = true
```
jshaughn, line 62: I think you want to leave the `default` clause as-is and move the isInaccessible check outside, unfortunately it seems we now need to check every node.  I think:
```
default:
	if isOutside(n, a.Namespaces) {
		n.Metadata[graph.IsOutside] = true
	}
}
if _, ok := n.Metadata[graph.IsInaccessible]; !ok {
	if isInaccessible(n, a.AccessibleNamespaces) {
		n.Metadata[graph.IsInaccessible] = true
	}
}
```

jshaughn (changes requested):
I didn't see this PR before responding in the Slack thread.  But to reiterate, I think we may want to leave it as-is and not prune nodes after the first inaccessible node.  Users may see value here and may be ok as long as we don't give them access to details.  Also, it's feasible traffic may go through something inaccessible (like a gateway) and then into accessible nodes.  At worst we may have to make this a configurable option.

graph/api/api_test.go
```diff
@@ -203,12 +151,15 @@ func mockNamespaceGraph(t *testing.T) (*prometheus.Client, *prometheustest.PromA
...
 		"grpc_response_status":           "0",
-		"response_flags":                 "-"}
+		"response_flags":                 "-",
+	}
 	q1m1 := model.Metric{
 		"source_workload_namespace":      "unknown",
 		"source_workload":                "unknown",
 		"source_canonical_service":       "unknown",
 		"source_canonical_revision":      "unknown",
+		"source_cluster":                 "east",
```
jshaughn, line 161: This should be `unknown`.   It's unlikely a cluster would be set on an unknown source.

graph/api/api_test.go
```diff
@@ -219,12 +170,15 @@ func mockNamespaceGraph(t *testing.T) (*prometheus.Client, *prometheustest.PromA
...
 		"grpc_response_status":           "0",
-		"response_flags":                 "-"}
+		"response_flags":                 "-",
+	}
 	q1m2 := model.Metric{
 		"source_workload_namespace":      "unknown",
 		"source_workload":                "unknown",
 		"source_canonical_service":       "unknown",
 		"source_canonical_revision":      "unknown",
+		"source_cluster":                 "east",
```
jshaughn, line 180: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -934,69 +1036,87 @@ func mockNamespaceRatesGraph(t *testing.T) (*prometheus.Client, *prometheustest.
...
 		"destination_canonical_revision": "v1",
-		"response_flags":                 "-"}
+		"response_flags":                 "-",
+	}
 	q7m1 := model.Metric{
 		"source_workload_namespace":      "unknown",
 		"source_workload":                "unknown",
 		"source_canonical_service":       "unknown",
 		"source_canonical_revision":      "unknown",
+		"source_cluster":                 "east",
```
jshaughn, line 1055: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -1071,6 +1201,40 @@ func respond(w http.ResponseWriter, code int, payload interface{}) {
 	_, _ = w.Write(response)
 }
 
+// Helper method that tests the objects are equal and if they aren't will
+// unmarshal them into a json object and diff them. This way the output of the failure
+// is actually useful. Otherwise printing the byte slice results in completely incomprehensible output.
```
jshaughn, line 1206:
```suggestion
// is actually useful. Otherwise printing the byte slice results is incomprehensible.
```

jshaughn: Agreed it was incomprehensible, I always just ignored that part, clipped the whole differing json, and then diff'd it in git gui.  Thanks for taking time to improve.

graph/api/api_test.go
```diff
@@ -1390,6 +1539,8 @@ func TestWorkloadNodeGraph(t *testing.T) {
 		"source_workload":                "unknown",
 		"source_canonical_service":       "unknown",
 		"source_canonical_revision":      "unknown",
+		"source_cluster":                 "east",
```
jshaughn, line 1542: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -1550,12 +1731,15 @@ func TestWorkloadNodeGraph(t *testing.T) {
...
-		"response_flags":                 "-"}
+		"response_flags":                 "-",
+	}
 	q1m8 := model.Metric{
 		"source_workload_namespace":      "bookinfo",
 		"source_workload":                "productpage-v1",
 		"source_canonical_service":       "productpage",
 		"source_canonical_revision":      "v1",
+		"source_cluster":                 "east",
+		"destination_cluster":            "east",
```
jshaughn, line 1742: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -1666,6 +1864,8 @@ func TestAppNodeGraph(t *testing.T) {
 		"source_workload":                "unknown",
 		"source_canonical_service":       "unknown",
 		"source_canonical_revision":      "unknown",
+		"source_cluster":                 "east",
```
jshaughn, line 1867: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -1826,12 +2056,15 @@ func TestAppNodeGraph(t *testing.T) {
...
-		"response_flags":                 "-"}
+		"response_flags":                 "-",
+	}
 	q1m8 := model.Metric{
 		"source_workload_namespace":      "bookinfo",
 		"source_workload":                "productpage-v1",
 		"source_canonical_service":       "productpage",
 		"source_canonical_revision":      "v1",
+		"source_cluster":                 "east",
+		"destination_cluster":            "east",
```
jshaughn, line 2067: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -1942,6 +2189,8 @@ func TestVersionedAppNodeGraph(t *testing.T) {
 		"source_workload":                "unknown",
 		"source_canonical_service":       "unknown",
 		"source_canonical_revision":      "unknown",
+		"source_cluster":                 "east",
```
jshaughn, line 2192: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -2102,12 +2381,15 @@ func TestVersionedAppNodeGraph(t *testing.T) {
...
-		"response_flags":                 "-"}
+		"response_flags":                 "-",
+	}
 	q1m8 := model.Metric{
 		"source_workload_namespace":      "bookinfo",
 		"source_workload":                "productpage-v1",
 		"source_canonical_service":       "productpage",
 		"source_canonical_revision":      "v1",
+		"source_cluster":                 "east",
+		"destination_cluster":            "east",
```
jshaughn, line 2392: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -2304,6 +2608,8 @@ func TestRatesNodeGraphTotal(t *testing.T) {
 		"source_workload":                "unknown",
 		"source_canonical_service":       "unknown",
 		"source_canonical_revision":      "unknown",
+		"source_cluster":                 "east",
```
jshaughn, line 2611: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -2464,12 +2800,15 @@ func TestRatesNodeGraphTotal(t *testing.T) {
...
-		"response_flags":                 "-"}
+		"response_flags":                 "-",
+	}
 	q1m8 := model.Metric{
 		"source_workload_namespace":      "bookinfo",
 		"source_workload":                "productpage-v1",
 		"source_canonical_service":       "productpage",
 		"source_canonical_revision":      "v1",
+		"source_cluster":                 "east",
+		"destination_cluster":            "east",
```
jshaughn, line 2811: This should be `unknown`.

graph/api/api_test.go
```diff
@@ -87,84 +89,28 @@ func setupMocked() (*prometheus.Client, *prometheustest.PromAPIMock, *kubetest.K
 	return client, api, k8s, nil
 }
 
-func setupMockedWithIstioComponentNamespaces(meshId string, userClients map[string]kubernetes.ClientInterface) (*prometheus.Client, *prometheustest.PromAPIMock, *kubetest.K8SClientMock, error) {
+func setupMockedWithIstioComponentNamespaces(meshId string, userClients map[string]kubernetes.ClientInterface) (*prometheus.Client, *prometheustest.PromAPIMock, error) {
```
jshaughn, line 92: Good cleanup of the clients stuff...

jshaughn (changes requested):
Looks good except that it's not realistic to set the cluster on the unknown nodes.  I think you need to set the cluster="unknown" to reflect what Istio would actually set in that situation.  That unfortunately may affect some of the expectations...

graph/api/api_test.go
nrfox: Ah I could've done that as well. I mostly didn't want to scroll through the large output. Hopefully won't need to update these expectation files much.

nrfox:
@jshaughn updated the "unknown" references

jshaughn (approved):
Big effort, looks good!

PR #6254, opened 2023-06-13
Sleep hack script refactor to a separate file

matejnesuta:
Due to the upcoming Cypress hooks, I needed to move the installation of the Sleep app into a separate file. This file is now intended to be called from Cypress when Sleep demo app is not detected.

Tested on upstream OCP and it works.


jmazzitelli (approved):

FilipB (approved):
LGTM

Issue #6253, opened 2023-06-12
K8Gateway object not visible in the Istio Config List. 


matejnesuta:
I encountered this on the upstream OCP cluster with Istio. When creating a K8Gateway object (just like in [this test](https://github.com/kiali/kiali/blob/26cfb9e05446498cb5aa02aa89045556ec886046/frontend/cypress/integration/featureFiles/wizard_istio_config.feature#L15)), the object gets created and I was able to query it using the OC command:
![image](https://github.com/kiali/kiali/assets/45061792/06c3fd10-8217-47c3-85b4-6ffa3e28c145)

However, the objects I created using this method are not visible in the Istio Config List and are not present in the request payload either. 
![image](https://github.com/kiali/kiali/assets/45061792/5e23802b-9348-4044-8304-e9578b48d206)

The first 2 test in this screenshot fail because of this and the last 2 tests might as well, but I am not 100% sure. 
![image](https://github.com/kiali/kiali/assets/45061792/4dce4e33-f8a3-4c4f-b7b8-ab3157ed086a)

PR #6256, opened 2023-06-14
Added log

hhovsepy:
For debug purposes to reproduce https://github.com/kiali/kiali/issues/6253 on OCP

### DIFF
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -482,6 +482,7 @@ func ParseRegistryConfig(config map[string][]byte) (*RegistryConfiguration, erro
 			for _, iItem := range ajRegistry {
 				if mItem, ok := iItem.(map[string]interface{}); ok {
 					kind := mItem["kind"].(string)
+					log.Infof("Kind %s", kind)
 					switch kind {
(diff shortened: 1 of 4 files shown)


kubernetes/istio.go
```diff
@@ -482,7 +482,7 @@ func ParseRegistryConfig(config map[string][]byte) (*RegistryConfiguration, erro
 			for _, iItem := range ajRegistry {
 				if mItem, ok := iItem.(map[string]interface{}); ok {
 					kind := mItem["kind"].(string)
-					log.Infof("Kind %s", kind)
+					log.Infof("Kind %s %s", kind, iItem)
```
jmazzitelli, line 485: just FYI.. not wrapping dynamic values in log messages will be confusing/hard to read especially if one happens to be an empty string or if they have spaces in them or if you have multiple values and you don't demarcate them. Our usual way to do this is wrap in brackets to make the log message easier to read. Also, I don't think we want Infof here, Debugf would be better so as not to make the logs noisy.
```suggestion
					log.Debugf("Kind [%s] [%s]", kind, iItem)
```

PR #6257, opened 2023-06-14
Demoapps cypress hooks

matejnesuta:
This is what I originally intended to do in #6242. I moved the shell commands into separate .sh files to make the hook more clear, just like @ScriptingShrimp requested. I also changed the way of the pod check because of an incident on one of the downstream pipelines. In the original PR, if for example productpage was not in "Running" mode and there were duplicate details-v1 pods running, the original assertion would evaluate this as healthy bookinfo app. This is no longer the case. 

I don't think the commands in `wait-for-demoapp.sh` are used anywhere else other than our Jenkins jobs though.

I added detection of the node architecture, so the installation should theoretically work on non-amd64 clusters. The hook does not do anything when worker nodes with multiple architectures inside one cluster are detected. I am honestly not even sure, if it is a possible scenario, because I don't think hack scripts allow for such setup at all. 

If the constant checking before every test slows down the whole suite, I can add only one hook per featureFile.


matejnesuta:
I added Error Rates app detection and I am thinking about adding the rest of the hooks here as well.

hack/istio/install-sleep-demo.sh
```diff
@@ -24,6 +24,9 @@ DELETE_SLEEP="false"
 while [[ $# -gt 0 ]]; do
   key="$1"
   case $key in
+    -a|--arch)
```
matejnesuta, line 27: This argument doesn't do anything at the moment, but it will be needed sooner or later, because there was a requirement to update sleep app for non-amd64 architectures. This seemed like the easiest solution, rather than messing with the hooks.ts file and than reverting to original state once the mentioned requirement is fulfilled

matejnesuta:
To test this:
- Deploy Kiali without any demoapps
- Run any test with either @bookinfo-app, @error-rates-app or @sleep-app
- The presence of the specific app should be detected before every test. If app is not detected, then Cypress should run hack scripts to install it. If the app is detected, Cypress should proceed to the test itself.

ScriptingShrimp:
> I didn't run this but it looks OK. I think I'd prefer if the demo-app requirements could be specified one time for the feature file, and not be checked on each test. But, I defer to QE to make the final decision as to the best approach.

This adds better granularity and control over each scenario dependency. Currently in some `.feature` files we do utilize different demo apps, so we just check that they are present and valid. It does add roughly 50miliseconds to execution of each scenario (tested, calculated). 

When we do develop new `.feature` files, we can definitely utilize `Background` pre-condition better, but honestly I feel this is a bit better / faster solution, see:

- If we trigger i.e. all tests utilizing sleep app, we trigger them using tags `@sleep-app` so the app gets installed, all regression executed 
vs 
- If we trigger specific '.feature' files and we have different dependencies for each feature, meaning that the different demo apps could get installed during the execution of different feature files, which is more time complex

ScriptingShrimp (dismissed):
LGTM

hack/istio/cypress/bookinfo-status.sh
```diff
@@ -0,0 +1,32 @@
...
+      "reviews-v1"
+      "reviews-v2"
+      "reviews-v3")
+
+input=$(kubectl get pods -n bookinfo -o=custom-columns=NAME:.metadata.name,Status:.status.phase --no-headers=true)
+
+for pod in ${!apps[@]}; do
+  count=$(echo "$input" | grep "${apps[$pod]}" | awk -F ' ' '{print $1}' | wc -l)
+  status=$(echo "$input" | grep "${apps[$pod]}" | awk -F ' ' '{print $2}')
+  if [ $count -ne 1 ] || [ $status != "Running" ]
```
FilipB, line 24: Pod could be also Pending I believe. We should not fail if the pod is still starting and wait for Running status.

FilipB (changes requested):
Does it make sense to move sh scripts out of cypress folder? Those could be used not only by cypress tests.

ScriptingShrimp:
> Does it make sense to move sh scripts out of cypress folder? Those could be used not only by cypress tests.

Eventually we will migrate scripts out of cypress folder. It would be nice to have them also utilized for Mastra 3.0 testing so we do not have things all over the place

matejnesuta:
I used oc wait command, which should be enough I believe ([see the k8s docs related](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase)). The pod can also get stuck in the Pending state indefinitely, so I added a timeout of 60 seconds.

I moved the wait_for_namespace.sh script out of the Cypress folder, because I expect it to be used in our OCP pipelines soon. Not sure about the rest yet.

FilipB (approved):
LGTM

ScriptingShrimp (approved):
retested on OCP, all passing. Merging

PR #6259, opened 2023-06-15
Control plane metrics improvements

### LINKED ISSUE #6177: show more control plane info in OverView page
We are using the openshift service mesh and kiali, and now I have encountered a feature to add some monitoring information from the control surface to the overview page. The current page is as follows, and only a chart of istiod cpu information seems to be displayed.
image
I went to kiali's code and found that the other metric code was displayed in front of kiali. Use 'ShowMetrics' to determine if there is any data to show.
image

So is there a configuration that allows the overview page to display more information about the control panel? Such as memory information?

leandroberetta:
Resolves #6177.

Docs PR: https://github.com/kiali/kiali.io/pull/664

This PR adds some logic to fetch the "process" metrics (CPU and memory) if "container" metrics are not available in Prometheus (this is the case for OSSM when it's using it's own Prometheus, which hasn't configured cAdvisor scraping metrics).

Kiali will indicate the source of the metric like:

image

### DIFF
--- a/business/metrics.go
+++ b/business/metrics.go
@@ -262,19 +262,19 @@ func (in *MetricsService) GetControlPlaneMetrics(q models.IstioMetricsQuery, sca
 	}
 	metrics["pilot_proxy_convergence_time"] = append(metrics["pilot_proxy_convergence_time"], converted...)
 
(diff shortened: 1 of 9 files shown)


jshaughn (approved):
Looks good!  Code looks fine and in my test I now see a memory chart when using openshift, where before I did not.  The tooltip reflects the proper source:

image

PR #6260, opened 2023-06-15
Fixed not loading of several config objects which are not cached.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6253
Istio objects which are not in a Cache (hence K8s Gateways) were not loaded in a case when accessible_namespaces are not ["**"], this happens after merging a PR https://github.com/kiali/kiali/pull/6204.
The [initial issue](https://github.com/kiali/kiali/issues/6196) was about IstioAPI disabled case.

The fix it to add additional check when IstioAPI is not enabled, only then skip loading configs from registry.

### DIFF
--- a/handlers/istio_config.go
+++ b/handlers/istio_config.go
@@ -96,8 +96,9 @@ func IstioConfigList(w http.ResponseWriter, r *http.Request) {
 
 	istioConfig := models.IstioConfigList{}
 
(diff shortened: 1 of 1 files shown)


hhovsepy:
Tests pass on OCP.

jshaughn (approved):
Looks good!

PR #6272, opened 2023-06-19
[v1.65] Cypress image

FilipB:
Adding cypress image dockerfile and updating readme. This is needed for interop job to use OSSM 2.4


ScriptingShrimp (approved):
LGTM

PR #6273, opened 2023-06-19
[v1.65] OSSM-4214: Adding a Dockerfile for integration tests to be used with …

FilipB:
Adding int tests image dockerfile and updating readme. This is needed for interop job to use OSSM 2.4


hhovsepy (approved):
code lgtm, not tested

PR #6275, opened 2023-06-19
Virtual list type does not depend on url value

### LINKED ISSUE #6262: Virtual List type should not depend on url value
The way to obtain which Virtual List type has to be instantiated is to check the url with a regex, but this is not possible in OSSMC. As an alternative the type can be a React property and each Virtual List defines the type to implement.

ferhoyos:
** Describe the change **

Virtual list type is not a dynamic value, so it does not need to depend on url value. To make it compatible with OSSMC plugin, type is converted to React input property.

** Issue reference **

Closes #6262

### DIFF
--- a/frontend/src/components/VirtualList/Config.ts
+++ b/frontend/src/components/VirtualList/Config.ts
@@ -280,7 +280,6 @@ const istio: Resource = {
 };
 
 const conf = {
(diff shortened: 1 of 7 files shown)


hhovsepy (approved):
Code lgtm, tested locally.

PR #6276, opened 2023-06-19
Add Kiali proxy to API url

### LINKED ISSUE #6265: Add Kiali proxy url to API services
Since OSSMC uses a proxy to call the Kiali API, the proxy endpoint must be added as an environment variable in OSSMC configuration and taken into account for API endpoint definition.

Axios Request:
url: proxyUrl + '/' + url

ferhoyos:
** Describe the change **

Add Kiali proxy value from environment variable (optional) to API url.

** Issue reference **

Closes #6265

### DIFF
--- a/frontend/src/services/Api.ts
+++ b/frontend/src/services/Api.ts
@@ -49,6 +49,10 @@ export interface Response<T> {
   data: T;
 }
 
+const getKialiProxy = (): string | null => {
+  return process.env.KIALI_PROXY ?? null;
+};
+
 /** API URLs */
 
 const urls = config.api.urls;
@@ -59,13 +63,17 @@ const loginHeaders = config.login.headers;
 
 /**  Helpers to Requests */
 
-const getHeaders = () => {
-  return { ...loginHeaders };
+const getHeaders = (proxyUrl: string | null) => {
+  if (proxyUrl) {
+    return { 'Content-Type': 'application/x-www-form-urlencoded' };
+  } else {
+    return { ...loginHeaders };
+  }
 };
 
 /** Create content type correctly for a given request type */
-const getHeadersWithMethod = method => {
-  var allHeaders = getHeaders();
+const getHeadersWithMethod = (method: HTTP_VERBS, proxyUrl: string | null) => {
+  let allHeaders = getHeaders(proxyUrl);
   if (method === HTTP_VERBS.PATCH) {
     allHeaders['Content-Type'] = 'application/json';
   }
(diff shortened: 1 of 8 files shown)


jshaughn (commented):
Is this proposed Environment variable something that should be defined in the OSSMC CRD? I guess it can not be defined in the Kiali CRD because you need the proxy to be able to initially contact the API.

ferhoyos:
No, there is no need to define it in OSSMC CRD because it is always the same value (/api/proxy/plugin/ossmconsole/kiali). It only changes when you deploy the plugin locally (http://localhost:3001).

ferhoyos:
This is how Kiali proxy was defined in OSSMC plugin code:

```
// The ConsolePlugin resource defines a proxy to the internal Kiali Service.
// This is the main mechanism provided by the OpenShift Console backend to proxy internal requests.
// This "proxy" is defined under the same domain of the browser and it's mapped to:
// https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins
export const CONSOLE_PROXY = '/api/proxy/plugin/ossmconsole/kiali';

// Direct requests from the Plugin to Kiali API should use the KialiProxy host.
// This can be relative to the same domain in production environments but for development it requires a different config.
export const getKialiProxy = (): string => {
    if (process.env.NODE_ENV === 'development') {
        return process.env.KIALI_API_HOST;
    }
    return CONSOLE_PROXY;
}
```

jshaughn:
I'm assigning review to @jmazzitelli because I think he is well suited for this one and I am overloaded.

frontend/src/services/Api.ts
```diff
@@ -49,6 +49,11 @@ export interface Response<T> {
   data: T;
 }
 
+/** Get internal Kiali proxy if exists */
+const getKialiProxy = (): string | null => {
```
jmazzitelli, line 53: When I see the word "proxy" I think of an HTTP proxy that sits in front of some other server. But I don't think that's what this means.

Can you add some comments here about what this proxy is? Because right now it isn't clear. Is this an HTTP proxy or something else?

Also, can you explain WHERE this environment variable is set? If I want to set "KIALI_PROXY" environment variable, where is that done? Is that only for, say, a development environment (i.e. "yarn start" where I set the environment variable in my own command line window?) or is this also going to be set when deployed inside the cluster? If so, who sets this environment variable?

In short, this is confusing to people not familiar with this stuff. Having a few lines of comments here would help people to understand what this proxy is and how it is meant to be set.

jmazzitelli (changes requested):
the code seems OK, but I'm not sure what this proxy is :) See my request below for what I would like to see documented.

frontend/src/services/Api.ts
ferhoyos: I have renamed to API proxy and added a comment that this is only defined in some platforms like Openshift console and not application for standalone Kiali application

ferhoyos:
I have also added comments to clarify other Kiali changes done only because of OSSMC (history basename and routing wrappers). I hope these comments reduce the confusion regarding these 'weird' changes.

jmazzitelli (approved):
the comments are helpful

PR #6279, opened 2023-06-20
Return cluster info despite mesh id and refactor svc

### LINKED ISSUE #6249: [Multi-cluster] Some pages are broken when meshID is not set
When the meshID is not set, the istio api does not populate the `clusters` field on the `/config` response. The frontend looks for this clusters field to determine whether or not Kiali is in "multi-cluster mode".

nrfox:
Shows mesh page now even when the mesh id is not set. Also refactors the mesh service since it wasn't using the cache in some places as part of #5599. Now the cluster info is stored in the kiali cache rather than inside the mesh service. This should also lay some of the ground work for multi-primary support.

Fixes #6249


kubernetes/client_factory.go
```diff
@@ -430,7 +437,7 @@ func (cf *clientFactory) refreshClientIfTokenChanged(cluster string) error {
 		remoteRci, ok := cf.remoteClusterInfos[cluster]
 		cf.mutex.RUnlock()
 		if !ok {
-			return fmt.Errorf("Cannot refresh token for unknown cluster [%s]", cluster)
+			return fmt.Errorf("cannot refresh token for unknown cluster [%s]", cluster)
```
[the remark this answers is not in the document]
nrfox: It's something my editor's linter complained about but I think it comes from [here](https://github.com/golang/go/wiki/CodeReviewComments#error-strings).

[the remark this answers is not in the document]
jshaughn: Interesting.  OK, ignore this comment.

handlers/config.go
```diff
@@ -49,25 +47,25 @@ type DeploymentConfig struct {
...
-	IstioLabels          config.IstioLabels          `json:"istioLabels,omitempty"`
-	IstioConfigMap       string                      `json:"istioConfigMap"`
-	KialiFeatureFlags    config.KialiFeatureFlags    `json:"kialiFeatureFlags,omitempty"`
-	LogLevel             string                      `json:"logLevel,omitempty"`
-	Prometheus           PrometheusConfig            `json:"prometheus,omitempty"`
+	AccessibleNamespaces []string                      `json:"accessibleNamespaces,omitempty"`
+	AuthStrategy         string                        `json:"authStrategy,omitempty"`
+	AmbientEnabled       bool                          `json:"ambientEnabled,omitempty"`
+	ClusterInfo          ClusterInfo                   `json:"clusterInfo,omitempty"`
+	Clusters             map[string]kubernetes.Cluster `json:"clusters,omitempty"`
```
jshaughn, line 54: This is sort of confusing.  Not sure why we need `ClusterInfo` just to replicate the home cluster info when there are multiple clusters.  Instead of being backward compatible maybe we should clean this up.   We could maybe defer until we do more work on the `Mesh` page, because I think in the future we want to show that menu option at all times, even for single-cluster, and migrate the detailed control plane info to there from `Overview`.

jshaughn (approved):
Things are working, I see the new `config.clusters`:

```
  "authStrategy": "anonymous",
  "clusterInfo": {},
  "clusters": {
    "Kubernetes": {
      "apiEndpoint": "https://10.217.4.1:443",
      "isKialiHome": true,
      "isGatewayToNamespace": false,
      "kialiInstances": [
        {
          "namespace": "istio-system",
          "operatorResource": "kiali-operator/kiali",
          "serviceName": "kiali",
          "url": "https://kiali-istio-system.apps-crc.testing:20001",
          "version": "dev"
        }
      ],
      "name": "Kubernetes",
      "network": "",
      "secretName": ""
    }
  },
  "deployment": {},
```

Approving but see my comment about keeping around 'ClusterInfo`...

handlers/config.go
nrfox: That's a good point. I just kept the `ClusterInfo` field around but that same info is exposed by the `clusters` field. I'll look at consolidating them as part of this change.

jshaughn:
@nrfox , Let's get this one updated and in...

handlers/config.go
nrfox: The rebase on this was kind of hairy and this PR is already large. If it's alright I'll create a separate issue for removing the `ClusterInfo`.

jshaughn (approved):
Looks good to me!  Let's follow up with that "ClusterInfo" change.

PR #6284, opened 2023-06-20
[ci] don't expect SNAPSHOT to be in helm chart name

jmazzitelli:
we no longer look for SNAPSHOT in the name - this is so we can run this script in version branches (like `v1.65`).


jshaughn (approved):
LOK

PR #6286, opened 2023-06-21
Delete DetailObject component

### LINKED ISSUE #6281: DetailObject component is not longer used
I have discovered that frontend component DetailObject **(components/Details/DetailObject)** is not used in Kiali application.

Is there any historical reason to maintain such component? If not, I think we could delete it.

ferhoyos:
** Describe the change **

Delete DetailObject component since it is not used in Kiali application

** Issue reference **

Closes #6281


jshaughn (approved):
Looks good

PR #6288, opened 2023-06-21
Remove routing properties from Kiali pages

### LINKED ISSUE #6263: Remove routing properties from Kiali pages
Kiali pages make use of react routing properties (values obtained from url), but OSSMC does not use react-router and it is not possible to obtain such values from url. The solution is to isolate routing properties from pages (convert them into normal React properties) and create small wrappers acting as a bridge between routing and kiali pages (obtain the routing properties and pass them to the Kiali page).

This issue only affects to pages implementd in OSSMC:
- Graph Page
- Istio Config Details Page
- Service Details Page
- Workload Details Page

ferhoyos:
** Describe the change **

Remove routing properties from Kiali pages to be able to use them without react-router. Route wrappers have been created to extract routing properties and add to Kiali pages as React input properties

** Issue reference **

Closes #6263

### DIFF
--- a/frontend/src/routes.tsx
+++ b/frontend/src/routes.tsx
@@ -1,17 +1,18 @@
+import { MenuItem, Path } from './types/Routes';
+import { Paths } from './config';
 import WorkloadListPage from './pages/WorkloadList/WorkloadListPage';
 import ServiceListPage from './pages/ServiceList/ServiceListPage';
 import IstioConfigPage from './pages/IstioConfigList/IstioConfigListPage';
-import IstioConfigDetailsPage from './pages/IstioConfigDetails/IstioConfigDetailsPage';
-import WorkloadDetailsPage from './pages/WorkloadDetails/WorkloadDetailsPage';
 import AppListPage from './pages/AppList/AppListPage';
-import AppDetailsPage from './pages/AppDetails/AppDetailsPage';
-import OverviewPageContainer from './pages/Overview/OverviewPage';
-import { MenuItem, Path } from './types/Routes';
-import GraphPageContainer from './pages/Graph/GraphPage';
-import { Paths } from './config';
-import ServiceDetailsPageContainer from './pages/ServiceDetails/ServiceDetailsPage';
-import IstioConfigNewPageContainer from './pages/IstioConfigNew/IstioConfigNewPage';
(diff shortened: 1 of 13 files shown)


hhovsepy:
After the second commit, there is broken IstioConfigNew page
Screenshot from 2023-06-21 13-29-47

ferhoyos:
Fixed

hhovsepy (approved):
LGTM, the issue is fixed now, tests are passing.

frontend/src/routes.tsx
```diff
@@ -63,39 +64,39 @@ const defaultRoute = '/overview';
 const pathRoutes: Path[] = [
   {
     path: '/overview',
-    component: OverviewPageContainer
+    component: OverviewPage
   },
   {
     path: '/graph/node/namespaces/:namespace/' + Paths.AGGREGATES + '/:aggregate/:aggregateValue',
-    component: GraphPageContainer
+    component: GraphRoute
```
jshaughn, line 71: @ferhoyos I know this is already merged, but what I don't understand is how Redux is still working when neither the routes nor the wrappers are referring to the redux containers but rather the react component directly.   Here the component is changed from GraphPageContainer to GraphRoute, but in GraphRoute I see `<GraphPage>` not `<GraphPageContainer>`, for example.

ferhoyos: When you import a default export you can set whatever name you want, it is not mandatory to use the same name of the export (because there is only one default export per file). Only with normal export you have to import the same variable name (because there can be many of them per file).

An example:
```
export default GraphPageContainer;
import GraphPage from 'xxx/GraphPage'  //correct 
import AAA from 'xxx/GraphPage'  //correct 
```
```
export GraphPageContainer;
import { GraphPage } from 'xxx/GraphPage' //wrong
import { GraphPageContainer } from 'xxx/GraphPage' //correct
```

There are other examples in Kiali of different import names with default exports:
```
reducers/TourState:
export default tour;

reducers/index:
import TourState from './TourState';
```

jshaughn: @ferhoyos Thanks for that explanation!  I was confused.  And now I know why, it's confusing!   I totally agree, we should remove the use of default export, I don't think there is a huge benefit to being able to rename the component being exported by default, and it can lead to confusion.

Issue #5932, opened 2023-03-21
Revisit namespace selector


nrfox:

leandroberetta:
Another option can be to use the namespace sameness principle and just show unique entries, no matter in which clusters the namespaces are. 

Example: when selecting the bookinfo, two cards will appear, one for east and other one for west. 

Of course we would need to add a filter for cluster (I guess we agreed on this already also for list pages).

Issue #5923, opened 2023-03-16
Multicluster view for Istio config wizards


josunect:
The Istio config wizards can include a multi cluster selector: 
- Not visible for single cluster
- What should be selected by default? 

Example of selector: 
![image (5)](https://user-images.githubusercontent.com/49480155/225616870-fa5a34c8-d4c0-4e06-9b6d-a6fa91e5f617.jpg)

Ref. #5620

josunect:
That would probably be modified when writes are enabled in remote clusters? 
For multicluster maybe we can add a warning or something similar, to make explicit that the writes are only allowed for the home cluster.

PR #6289, opened 2023-06-21
Istio Config creation: select cluster

hhovsepy:
RFE https://github.com/kiali/kiali/issues/5923

In multi-cluster installation mode, while creating new Istio config file, it appears "Select Clusters" dropdown to select cluster where configs will be created (by default selected home cluster).
Creation will create number_of_selected_clusters X number_of_selected_namespaces times requests to backend.

EDIT: After talking to UX, the Cluster selection is moved after Namespace selector. When selected Namespace does not exist in selected Cluster, Warning message is shown and the 'Create' button is disabled.
Screenshot from 2023-07-13 15-54-36


hhovsepy:
cc  Hi @andrew-ronaldson , we have added a Cluster selector similar to the Namespace selector in config creation wizard. Is this fine from UX perspective?

josunect:
Tested creating a PeerAuthentication in the remote cluster: 

image

PA were created:

image

Also validation: 

image

Tested edit/delete, also works as expected.

josunect:
This is a behavior that I've found: 

image

It creates the Istio Config, but still it is not listed in the Istio config list.

hhovsepy:
Thanks for the review @josunect , actually this behavior of creating configs where possible and failing where not but still showing the edit page happened always, will try to find a solution.

> 
> Maybe it is a bit difficult to define the behavior in multiple editions, but, I would expect here:
> 
>     * To redirect to the istio configs page if at least one of the configs was created
> 
>     * Show in the Istio config list any errors during the creation of the object
> 
>     * Not fail on error (I think the PA could be created also in the cluster east and NS alpha, but just cluster west and NS beta was created)

hhovsepy:
@josunect I have fixed the failing tests and the messy behavior with notifications.

josunect (commented):
Thanks, @hhovsepy ! That was a great work. Now you can see the created Istio config in the config list after creation. 

Just a comment. When you create multiple Istio objects, and some are created, some fail, you are not redirected to the istio config list (When a config creation action is successful, you are redirected there), and you can see the error messages. So, I think it is a bit confusing, and, If some configs where created, it would make sense to redirect to the Istio Config list. What do you think? 
image

hhovsepy:
@josunect thank you for the review, the multiple creation is designed in a way that even in a single cluster mode when some configs were not created but some were (multiple namespaces), it was still keeping the edit page, so I left that logic as it was (even if it is confusing). If it redirects to list page but shows error message, that could also be somehow confusion. If we rework the logic of error handling there, I think this is a separate task to agree.

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -202,24 +226,44 @@ class IstioConfigNewPageComponent extends React.Component<Props, State> {
...
+    }
+  };
+
+  onIstioResourceCreateForCluster = (cluster?: string) => {
     const jsonIstioObjects: { namespace: string; json: string }[] = this.state.itemsPreview.map(item => ({
       namespace: item.items[0].metadata.namespace || '',
       json: JSON.stringify(item.items[0])
     }));
 
     this.promises
```
josunect, line 244: What about something like this? So at least, in case of any failure, the user receives a success message for the Istio configs that where created successfully: 
```
    let errors = 0
    await Promise.all(jsonIstioObjects.map(o => API.createIstioConfigDetail(o.namespace, DIC[this.props.objectType], o.json, cluster))
    .map(p => p.catch(error => {
      AlertUtils.addError(
        'Could not create Istio ' +
        this.props.objectType +
        ' objects' +
        (cluster ? ' in cluster ' + cluster + '.' : '.'),
        error
      )
      errors++
    }))).then(() => {
      AlertUtils.add(
        'Istio ' + this.props.objectType + ' created' + (cluster ? ' in cluster ' + cluster : ''),
        'default',
        MessageType.SUCCESS
      );
    });

    if (errors == 0) {
      this.backToList();
    }
```

image

hhovsepy: Thank you Josune for the hints, will try them.

nrfox:
Nice to have this functionality! Regarding the implementation, I'd defer to @andrew-ronaldson but IMO we shouldn't add a cluster selector just for the wizards. We're still undecided on the best approach for selecting clusters (see https://github.com/kiali/kiali/issues/5932). Until we've got more clarity there IMO cluster should just be a field on the form itself. I don't think we need to support creating objects across multiple clusters at once. That greatly complicates error handling and I'm not sure that's a feature that most people will use.

hhovsepy:
Users can always select one cluster and one namespace to create the config objects, batch creation is an option and to avoid copy config objects across clusters. Error handling is done similarly as in multi-namespaces.
Cluster selector next to the namespace selector now is used only in Istio config new page, but can be shared later anywhere like selected namespaces, and if necessary can be merged with namespace selector.

nrfox:
> Cluster selector next to the namespace selector now is used only in Istio config new page, but can be shared later anywhere like selected namespaces, and if necessary can be merged with namespace selector.

IMO until we resolve #5932 we shouldn't add the cluster selector in some places but not others. That makes the UX inconsistent. If we want to support bulk creation of objects across multiple clusters, the cluster selector could go into the form itself for now? IMO that's how we should do bulk creation across namespaces as well instead of using the namespace selector but that is another issue.

andrew-ronaldson:
Following up on the conversation we had yesterday I'd rather see the Cluster & namespace fields as part of the creation form instead of above the page header. If there is an error (eg namespace didn't exist) it would be good to have all those interactive elements contained in one place.

hhovsepy:
Namespaces and Clusters selectors are moved into form:
Screenshot from 2023-07-14 09-49-53

josunect (changes requested):
Thanks, @hhovsepy 
The changes looks good!

I have some suggestions/small issues: 

I think the error message would be better next to the field it refers to, like the other error messages: 

image

Once the cluster is created it redirects to the config list but, I cannot see any configuration for the remote cluster: 

image

It looks like some state is saved. If you create one config in cluster east, and then you click in create another config, it looks like the cluster is pre - selected (There is no error for the cluster missing): 

image

But it is not: 

image

hhovsepy:
@josunect thank you for the review and suggestions, now the Clusters and Namespaces validations results are under their fields.
Yes there is a state kept for Clusters. similar to Namespaces, for the future, and it is fixed now that previously selected Clusters will be pre-selected next time (the bug is fixed).
And in my env all configs are created, probably it is a cache config issue on your env, let's look into it together.
Screenshot from 2023-07-17 19-16-17
Screenshot from 2023-07-17 19-15-55

frontend/src/pages/Graph/GraphHelpTour.tsx
```diff
@@ -57,6 +57,11 @@ export const GraphTourStops: { [name: string]: TourStopInfo } = {
     description: 'Select the namespaces you want to see in the graph.',
     position: PopoverPosition.bottomStart
   },
+  Clusters: {
+    name: 'Clusters',
+    description: 'Select the clusters.',
+    position: PopoverPosition.bottomStart
+  },
```
jshaughn, line 64: I don't think this is used.  Maybe it is left from an initial design?

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -381,16 +434,38 @@ class IstioConfigNewPageComponent extends React.Component<Props, State> {
...
-          <DefaultSecondaryMasthead />
+          <DefaultSecondaryMasthead showClusterSelector={false} hideNamespaceSelector={true} />
         </div>
         <RenderContent>
           <Form className={formPadding} isHorizontal={true}>
+            <FormGroup
+              label="Namespaces"
+              isRequired={true}
+              fieldId="namespaces"
+              helperTextInvalid={'An Istio Config resource needs at least a namespace selected'}
```
jshaughn, line 455:
```suggestion
              helperTextInvalid={'An Istio Config resource needs at least one namespace selected'}
```

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -381,16 +434,38 @@ class IstioConfigNewPageComponent extends React.Component<Props, State> {
...
+              validated={isValid(isNamespacesValid)}
+            >
+              <NamespaceDropdown disabled={false} />
+            </FormGroup>
+            <FormGroup
+              label="Clusters"
+              isRequired={true}
+              fieldId="clusters"
+              helperTextInvalid={'An Istio Config resource needs at least a cluster selected'}
```
jshaughn, line 464:
```suggestion
              helperTextInvalid={'An Istio Config resource needs at least one cluster selected'}
```

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -476,7 +548,9 @@ class IstioConfigNewPageComponent extends React.Component<Props, State> {
 
 const mapStateToProps = (state: KialiAppState) => {
   return {
-    activeNamespaces: activeNamespacesSelector(state)
+    activeNamespaces: activeNamespacesSelector(state),
+    activeClusters: activeClustersSelector(state),
```
jshaughn, line 552:
```suggestion
    activeClusters: activeClustersSelector(state),
    activeNamespaces: activeNamespacesSelector(state),
```

frontend/src/reducers/NamespaceState.ts
```diff
@@ -57,6 +69,7 @@ export const NamespaceStateReducer = (
       return updateState(state, {
         isFetching: false,
         items: filterDuplicateNamespaces(action.payload.list),
+        namespacesPerCluster: namespacesPerCluster(action.payload.list),
         lastUpdated: action.payload.receivedAt,
```
jshaughn, line 73:
```suggestion
        lastUpdated: action.payload.receivedAt,
        namespacesPerCluster: namespacesPerCluster(action.payload.list),
```

frontend/src/reducers/index.ts
```diff
@@ -22,6 +23,7 @@ export const rootReducer = combineReducers<KialiAppState, KialiAppAction>({
   graph: GraphDataStateReducer,
   messageCenter: MessageCenterReducer,
   namespaces: NamespaceStateReducer,
+  clusters: ClusterStateReducer,
```
jshaughn, line 26: alphabetize

frontend/src/actions/ClusterAction.ts
```diff
@@ -0,0 +1,11 @@
+import { ActionType, createStandardAction } from 'typesafe-actions';
+import { ActionKeys } from './ActionKeys';
+import { MeshCluster } from '../types/Mesh';
+
+export const ClusterActions = {
+  toggleActiveCluster: createStandardAction(ActionKeys.TOGGLE_ACTIVE_CLUSTER)<MeshCluster>(),
+  setActiveClusters: createStandardAction(ActionKeys.SET_ACTIVE_CLUSTERS)<MeshCluster[]>(),
+  setFilter: createStandardAction(ActionKeys.CLUSTER_SET_FILTER)<string>()
```
jshaughn, line 8:
```suggestion
  setActiveClusters: createStandardAction(ActionKeys.SET_ACTIVE_CLUSTERS)<MeshCluster[]>(),
  setFilter: createStandardAction(ActionKeys.CLUSTER_SET_FILTER)<string>()
  toggleActiveCluster: createStandardAction(ActionKeys.TOGGLE_ACTIVE_CLUSTER)<MeshCluster>(),
```

frontend/src/actions/KialiAppAction.ts
```diff
@@ -21,6 +22,7 @@ export type KialiAppAction =
   | LoginAction
   | MessageCenterAction
   | NamespaceAction
+  | ClusterAction
```
jshaughn, line 25: alphabetize, please fix the other ones here as well.

frontend/src/components/DefaultSecondaryMasthead/DefaultSecondaryMasthead.tsx
```diff
@@ -15,6 +15,7 @@ type ReduxProps = {
 type Props = ReduxProps & {
   actionsToolbar?: JSX.Element;
   hideNamespaceSelector?: boolean;
+  showClusterSelector?: boolean;
   rightToolbar?: JSX.Element;
```
jshaughn, line 19:
```suggestion
  rightToolbar?: JSX.Element;
  showClusterSelector?: boolean;
```

frontend/src/pages/IstioConfigNew/ClusterDropdown.tsx
```diff
@@ -0,0 +1,269 @@
...
+} from '../../components/BoundingClientAwareComponent/BoundingClientAwareComponent';
+import { KialiIcon } from 'config/KialiIcon';
+import { KialiAppState } from '../../store/Store';
+import { KialiDispatch } from '../../types/Redux';
+import { serverConfig } from '../../config';
+
+type ReduxProps = {
+  activeClusters: MeshCluster[];
+  filter: string;
+  clusters: MeshCluster[];
```
jshaughn, line 29:
```suggestion
  clusters: MeshCluster[];
  filter: string;
```

frontend/src/reducers/__tests__/NamespacesReducer.test.ts
```diff
@@ -129,6 +129,7 @@ describe('Namespaces reducer', () => {
     const currentDate = new Date();
     const currentState = {
       activeNamespaces: [{ name: 'my-namespace' }],
+      namespacesPerCluster: new Map<string, string[]>(),
```
jshaughn, line 132: alphabetize, also the `filter` attribute

frontend/src/reducers/__tests__/NamespacesReducer.test.ts
```diff
@@ -140,6 +141,7 @@ describe('Namespaces reducer', () => {
     );
     const expectedState = {
       activeNamespaces: [],
+      namespacesPerCluster: new Map<string, string[]>(),
```
jshaughn, line 144: alphabetize, also the `filter` attribute

frontend/src/store/Store.ts
```diff
@@ -36,8 +37,14 @@ export interface GlobalState {
...
+  readonly activeClusters: MeshCluster[];
+  readonly filter: string;
+}
+
 export interface NamespaceState {
   readonly activeNamespaces: Namespace[];
+  readonly namespacesPerCluster?: Map<string, string[]>;
   readonly items?: Namespace[];
   readonly isFetching: boolean;
   readonly lastUpdated?: Date;
```
jshaughn, line 50:
```suggestion
  readonly items?: Namespace[];
  readonly isFetching: boolean;
  readonly lastUpdated?: Date;
  readonly namespacesPerCluster?: Map<string, string[]>;
```

frontend/src/store/Store.ts
```diff
@@ -142,6 +149,7 @@ export interface KialiAppState {
   authentication: LoginState;
   messageCenter: MessageCenterState;
   namespaces: NamespaceState;
+  clusters: ClusterState;
```
jshaughn, line 152: alphabetize

jshaughn (changes requested):
This is looking good to me and is in line with Andrew's design input.  I agree with all of @josunect's suggestions and am happy that you made all of those suggested changes.  

> [@nrfox] IMO until we resolve https://github.com/kiali/kiali/issues/5932 we shouldn't add the cluster selector in some places but not others.

I'm not sure I agree with this because I think the cluster selector may only apply in certain use cases, like this one.  In other places, like the graph, or possibly other places, it may not make sense to aggressively narrow the clusters.  If nothing else I'd like the community to be a guide.  I actually think we could close 5932 given the de-duplication we've already applied.  I like the separate selectors we see here, local to the wizard.

I'm not sure we should retain the previous multi-namespace behavior around error-reporting and post-create navigation.  I feel like it's not the best experience but I'm OK keeping it for now, to limit the scope of change in this PR.  I still don't like that we generate errors to the Message Center for entries with invalid cluster-namespace combinations.  I think this should all be handled in the Preview.  The user should be able to confirm that he's OK with the entries that will not be created.  The only errors should be true, unexpected errors about failures to the config we are trying to create.  In general there should be no errors, and the user should be able to see the newly created config.

I think that in single-cluster mode the Cluster selector should be completely removed instead of showing the entry with no dropdown.
image

Sorry for all of the alphabetizing suggestions/comments but when something is already alphabetical I think it should stay that way.

jshaughn (approved):
Looks good from my side!

nrfox:
> I'm not sure I agree with this because I think the cluster selector may only apply in certain use cases, like this one. In other places, like the graph, or possibly other places, it may not make sense to aggressively narrow the clusters.

Having the cluster selector inside of the form seems good to me too. Before the cluster selector was next to the namespace selector outside of the form. 

> I like the separate selectors we see here, local to the wizard.

@jshaughn Agreed I think having the selectors as part of the form is a better UX. In fact, I wish we would extend this further and add a "namespace selector" to all forms instead of using what is selected in the namespace selector that is outside of the form. That has always confused me. My only gripe with two separate selectors, one for namespace and one for cluster, is that it is not fine grained enough. You somehow need cluster with namespace. For example:

Given cluster: [namespaces]
Cluster A: [bookinfo, alpha, beta]
Cluster B: [bookinfo]

if you wanted to create an istio config object(s) in cluster A: bookinfo but not cluster B: bookinfo there's no way to specify that by separating the cluster from the namespace. If it's a primary-remote deployment where cluster A is the primary and cluster B is the remote, then there aren't even any istio config CRDs on the remote cluster. On the backend, all kube objects, including namespaces, live in a single specific cluster.

@hhovsepy feel free to ignore all this ^^ for this change.

josunect:
If there are no errors it works as expected: 

image

The issue that I see is that, if the Istio config cannot be created in the remote cluster, there are no info or error messages: 
image

The user is redirected to the Istio config list but, it is empty.
image

I think in this case it would be nice to have an error/message to inform the user? 
(To reproduce it, I uninstalled the Istio CRDs from the cluster west)

hhovsepy:
@josunect thanks for the review, now it will show missing CRDs errors to user.

hhovsepy:
@josunect now it shows error to user when CRD is missing on remote cluster.

jshaughn (approved):
Still looks good to me!

josunect (approved):
Thanks, @hhovsepy ! Now you can see the errors when the Config cannot be created: 

image

Issue #6294, opened 2023-06-22
Service details of ServiceEntry fails


hhovsepy:
Create ServiceEntry in Istio Configs page.
Go to Service List.
![Screenshot from 2023-06-22 09-01-05](https://github.com/kiali/kiali/assets/604313/8aadf639-a8bf-4fd3-974b-681d72cd7eeb)

Click on ES ServiceEntry Service name.
The details page loading fails with error "not found":
![Screenshot from 2023-06-22 09-01-47](https://github.com/kiali/kiali/assets/604313/73887b13-4b22-4a95-ae7b-a8aa5a0ce1b7)

PR #6295, opened 2023-06-22
Fixed showing error when Service is found in registry.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6294

Even though ServiceEntry was found in Registry, the initial error of not found in cache was returning.
Fixed to override the initial error.


jshaughn (approved):
LG, tested and working.

The CI is failing but it seems environmental.  I think this can be merged.

PR #6305, opened 2023-06-26
Cypress Kiali Login test fix

matejnesuta:
The test expects buttons with auth providers when on Openshift. When no users are present on the cluster and the only way to log in is using the Kubeadmin, Openshift login page skips the buttons and goes straight to login form. However, the test still expects buttons, which causes it to fail. 

I've added a condition to prevent this from happening.

Thanks @FilipB for discovering this.


FilipB:
LGTM

ScriptingShrimp:
do we want to cherry pick this into v1.65?

ScriptingShrimp (approved):
LGTM

matejnesuta:
> do we want to cherry pick this into v1.65?

Yes, I think so. I will do it.

PR #6316, opened 2023-06-28
Auto-enable find/hide filters via Kiali CR

### LINKED ISSUE #6012: Auto-enable find/hide filters via Kiali CR
The `Kiali` CR allows us to create pre-defined hide/find filters via `cr.spec.kiali_feature_flags.ui_defaults.graph`.
A feature I feel is missing, is the ability to autmatically enable these filters.

In our Kiali instance, we really want to automatically hide traffic from outside of the selected namespaces.
So we do the following:

```
kiali-operator:
  cr:
    spec:
      kiali_feature_flags:
        ui_defaults:
          metrics_per_refresh: "10m"
          graph:
            hide_options:
              - description: "Hide unselected namespaces"
                expression: "outside"
          namespaces:
            - istio-system
            - backend-apps
            - frontend-apps
```

The user then has to select the hide options manually.
What I would love to see, is a third key under `hide_options` that allows us to toggle auto-enabling. Eg:

```
- description: "Hide unselected namespaces"
  expression: "outside"
  auto-enable: true
```

josunect:
Fixes #6012

Operator PR: https://github.com/kiali/kiali-operator/pull/665 
kiali.io PR: https://github.com/kiali/kiali.io/pull/671

Example for testing: 

```
  ui_defaults: 
    graph:
      hide_options:
         - description: "Hide unselected namespaces"
           expression: "outside"
         - description: "Hide nodes with no label region"
           expression: "!label:region"
      find_options:
         - description: "name not contains rev"
           expression: "name not contains rev"
         - description: "tcpserver"
           expression: "name = tcpserver"
           auto_enable: true
```

image

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -376,6 +376,7 @@ type DeploymentConfig struct {
 type GraphFindOption struct {
 	Description string `yaml:"description,omitempty" json:"description,omitempty"`
 	Expression  string `yaml:"expression,omitempty" json:"expression,omitempty"`
+	AutoEnable  bool   `yaml:"auto_enable,omitempty" json:"autoEnable,omitempty"`
 }
 
 // GraphSettings affect the graph visualization.
--- a/config/config.go
+++ b/config/config.go
@@ -376,7 +376,7 @@ type DeploymentConfig struct {
 type GraphFindOption struct {
 	Description string `yaml:"description,omitempty" json:"description,omitempty"`
 	Expression  string `yaml:"expression,omitempty" json:"expression,omitempty"`
-	AutoEnable  bool   `yaml:"auto_enable,omitempty" json:"autoEnable,omitempty"`
+	AutoSelect  bool   `yaml:"auto_select,omitempty" json:"autoSelect,omitempty"`
 }
 
 // GraphSettings affect the graph visualization.
--- a/config/config.go
+++ b/config/config.go
@@ -374,9 +374,9 @@ type DeploymentConfig struct {
 
 // GraphFindOption defines a single Graph Find/Hide Option
 type GraphFindOption struct {
+	AutoSelect  bool   `yaml:"auto_select,omitempty" json:"autoSelect,omitempty"`
 	Description string `yaml:"description,omitempty" json:"description,omitempty"`
 	Expression  string `yaml:"expression,omitempty" json:"expression,omitempty"`
-	AutoSelect  bool   `yaml:"auto_select,omitempty" json:"autoSelect,omitempty"`
 }
 
 // GraphSettings affect the graph visualization.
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -376,6 +376,7 @@ type DeploymentConfig struct {
 type GraphFindOption struct {
 	Description string `yaml:"description,omitempty" json:"description,omitempty"`
 	Expression  string `yaml:"expression,omitempty" json:"expression,omitempty"`
+	AutoEnable  bool   `yaml:"auto_enable,omitempty" json:"autoEnable,omitempty"`
```
jshaughn, line 379: See the operator PR review, propose to change this to "AutoSelect".

frontend/src/pages/Graph/GraphToolbar/GraphFind.tsx
```diff
@@ -162,6 +163,12 @@ export class GraphFind extends React.Component<GraphFindProps, GraphFindState> {
       }
     } else if (!!findValue) {
       HistoryManager.setParam(URLParam.GRAPH_FIND, findValue);
+    } else {
+      serverConfig.kialiFeatureFlags.uiDefaults.graph.findOptions.forEach(opt => {
+        if (opt.autoEnable === true) {
+          props.setFindValue(opt.expression);
+        }
+      });
```
jshaughn, line 171:
```suggestion
      const autoSelect = serverConfig.kialiFeatureFlags.uiDefaults.graph.findOptions.find(opt => opt.autoSelect);
      if (autoSelect) {
        props.setFindValue(opt.expression);
      }
```

This is minor, but in general I suggest not using Array.ForEach when you don't really want to perform an action on each array element.  In this case you only want to perform autoSelect one time, even if it is defined more than once.

jshaughn (changes requested):
Looks good, it correctly prefers anything in the URL.  It just needs a change to the new config attribute name when we decide on the name we want (see the Operator PR).  Also made one style suggestion.

config/config.go
josunect: Thanks for the review, @jshaughn ! I have updated the changes.

config/config.go
```diff
@@ -376,6 +376,7 @@ type DeploymentConfig struct {
 type GraphFindOption struct {
 	Description string `yaml:"description,omitempty" json:"description,omitempty"`
 	Expression  string `yaml:"expression,omitempty" json:"expression,omitempty"`
+	AutoSelect  bool   `yaml:"auto_select,omitempty" json:"autoSelect,omitempty"`
```
jshaughn, line 379: alphabetize

frontend/src/types/ServerConfig.ts
```diff
@@ -20,6 +20,7 @@ interface IstioAnnotations {
 interface GraphFindOption {
   description: string;
   expression: string;
+  autoSelect: boolean;
```
jshaughn, line 23: alphabetize

jshaughn (changes requested):
Looks good but one trivial change requested.

config/config.go
josunect: Well catched! Thank you, It has been updated.

jshaughn (approved):
Looks good!

josunect:
> Looks good!

Thanks for the review, @jshaughn !

PR #6318, opened 2023-06-28
Replace default exports by named ones

### LINKED ISSUE #6302: Avoid default exports in frontend
Currently both default and named exports are used in Kiali frontend application. This is confusing for developers because they do not know which kind of import need to use for each item exported. To avoid confusion, only named exports should be used.

A small refactorization is required to get rid of default exports within Kiali application

Small explanation of default vs named exports: https://dev.to/shacodes/default-vs-named-exports-19hj

ferhoyos:
** Describe the change **

Export defaults have been replaced by named ones and all the default imports have been refactored.

I have followed these rules for export naming convention:
For redux components:
-  The export default variable (redux container) has been renamed to the file name (e.g., `MiniGraphCard` instead of `MiniGraphContainer`) since it is usually the name used outside the file. 
- I have added `Component` suffix to base component (e.g., `MiniGraphCardComponent`). 
- If the file name already has `Component` at the end (e.g., `TimeRangeComponent`), the base component is renamed with `Comp`suffix (e.g.,`TimeRangeComp`)

For reducers:
- All reducers have the suffix `Reducer`to avoid conflicts with state objects (e.g., `GlobalStateReducer`)

For the rest of items the export default removal is straightforward (just remove `default` or type `export` just before the variable name to be exported).

To avoid the `export default` in the future, I have added an eslint rule.

There is no specific PR testing since no code logic has been modified. if the application compiles and all test are passed  it should work correctly.

** Issue reference **

Closes #6302


jshaughn (commented):
Looks good but still a cypress failure to take care of...

jshaughn (approved):
A huge PR but a consistent change and better, more consistent impl.  I'm sure it wasn't fun, thanks @ferhoyos !

ferhoyos:
> A huge PR but a consistent change and better, more consistent impl. I'm sure it wasn't fun, thanks @ferhoyos !

On the contrary, I like this kind of refactorizations! It is really worth when you see the code cleaner and more consistent :-)

PR #6292, opened 2023-06-21
Empty ServiceEntry Ports disables Preview button

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6287
 
Go To Istio Config page.
Actions -> Create ServiceEntry.
Fill the 'Name' and 'Hosts' fields.

Empty 'Ports' were showing error message but the 'Preview' button was enabled and confusing.
Now the 'Preview' button will be disabled when Ports are not set at all.

Before:
Screenshot from 2023-06-21 09-51-30

Now:
Screenshot from 2023-06-21 19-25-23

PR #6321, opened 2023-06-29
Empty ServiceEntry Ports disables Preview button fix (Backport to 1.65)

matejnesuta:
Backport of #6292, because the same failure has appeared on our downstream pipeline


matejnesuta:
I grabbed wrong changes in the featureFile. It should be ok now, but I would still personally wait for the CI to pass.

PR #6323, opened 2023-06-30
L4 Telemetry is backwards

### LINKED ISSUE #5822: L4 Telemetry is "backwards"
Arguably, istio metrics are backwards (https://github.com/istio/istio/pull/32549), but :shrug: 

What I did:

Shell ---> Echo. 
I sent a few bytes from shell to echo and got 100s back

In the graph, on the edge from shell --> Echo I see: Sent: 14. Recv: 0.1

I expect: Sent: 0.1, Recv: 14

josunect:
Fixes #5822


jshaughn:
I took a quick look and nice job, I think you found the two places I was talking about when I said, "... in the graph itself we may be able to make a very local change. On the server side, we may also be able to make a local change".  I'll follow up with you on Monday...

josunect:
> I took a quick look and nice job, I think you found the two places I was talking about when I said, "... in the graph itself we may be able to make a very local change. On the server side, we may also be able to make a local change". I'll follow up with you on Monday...

Sure, thanks @jshaughn . 
I will be reviewing the tests failures.

josunect:
Example with mTLS enabled: 

image

graph/api/api_test.go
```diff
@@ -2938,7 +2938,10 @@ func TestRatesNodeGraphTotal(t *testing.T) {
...
 	v6 := model.Vector{}
 
-	q7 := `round(sum(rate(istio_tcp_sent_bytes_total{reporter="source",source_workload_namespace="bookinfo",source_workload="productpage-v1"} [600s])) by (source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
+	q6a := `round(sum(rate(istio_tcp_received_bytes_total{reporter="destination",destination_workload_namespace="bookinfo",destination_workload="productpage-v1"} [600s])) by (source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
```
jshaughn, line 2941: Why did we need to add an extra query?

business/metrics_definitions.go
```diff
@@ -56,12 +56,12 @@ var istioMetrics = []istioMetric{
 	},
 	{
 		kialiName: "tcp_received",
-		istioName: "istio_tcp_received_bytes_total",
+		istioName: "istio_tcp_sent_bytes_total", // L4 telemetry is backwards
```
jshaughn, line 59:
```suggestion
		istioName: "istio_tcp_sent_bytes_total", // L4 telemetry is backwards, see https://github.com/istio/istio/issues/32399
```

business/metrics_definitions.go
```diff
@@ -56,12 +56,12 @@ var istioMetrics = []istioMetric{
...
 		isHisto:   false,
 	},
 	{
 		kialiName: "tcp_sent",
-		istioName: "istio_tcp_sent_bytes_total",
+		istioName: "istio_tcp_received_bytes_total", // L4 telemetry is backwards
```
jshaughn, line 64:
```suggestion
		istioName: "istio_tcp_received_bytes_total", // L4 telemetry is backwards, see https://github.com/istio/istio/issues/32399
```

graph/telemetry/istio/istio.go
```diff
@@ -742,11 +742,11 @@ func buildNodeTrafficMap(cluster, namespace string, n *graph.Node, o graph.Telem
 
 		switch o.Rates.Tcp {
 		case graph.RateReceived:
-			metrics = []string{"istio_tcp_received_bytes_total"}
```
jshaughn, line 745: Above the switch stmt can we add:

// L4 telemetry is backwards, see https://github.com/istio/istio/issues/32399

jshaughn (changes requested):
It looks strange to see the numbers and metrics reversed, but I think it is correct.  I had one question and am requesting just a few changes in comments.

jshaughn (approved):
Looks good, nice job!  Assuming the CI passes we can merge.

PR #6325, opened 2023-06-30
Rename cluster param for clusterMesh

### LINKED ISSUE #6322: Cluster param is invalid in Openshift Plugin
When Kiali is deployed as Openshift plugin ([kiali/openshift-servicemesh-plugin](https://github.com/kiali/openshift-servicemesh-plugin)), `cluster` param in HTTP request used for Multicluster feature is considered invalid:

For example:
`api/istio/status?cluster=Kubernetes` throws 400 Bad Request. Invalid cluster Kubernetes

image

This error does not happen when Kiali is used as standalone application within Openshift. 

image

According to https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins, Openshift uses an internal proxy to redirect plugin HTTP requests to corresponding service (in this case Kiali backend service). It seems that `cluster` param is reserved for internal use within Openshift Console, which validates the cluster name (in a different way that Kiali does). 

For example `cluster` param is used below in internal Openshift API (in this case `cluster=local-cluster`)

image

If we use same value `cluster=local-cluster` for Kiali, the HTTP request is not rejected by Openshift but the cluster is not found in Kiali (`Kubernetes` is the default cluster name):

image

For this reason `cluster` param should be renamed to `meshcluster` or similar to avoid conflicts with Openshift.

josunect:
This is changed just in the parameter, the variable name "cluster" was not renamed. 

I wonder if I should also change it to keep consistency, but still at some point (Like in the queries), clusterMesh should be changed by cluster.

Fixes #6322

### DIFF
--- a/frontend/src/app/History.ts
+++ b/frontend/src/app/History.ts
@@ -30,7 +30,7 @@ export { history };
 export enum URLParam {
   AGGREGATOR = 'aggregator',
   BY_LABELS = 'bylbl',
-  CLUSTER = 'cluster',
+  CLUSTERMESH = 'clusterMesh',
   DIRECTION = 'direction',
   DISPLAY_MODE = 'displayMode',
   DURATION = 'duration',
--- a/frontend/src/app/History.ts
+++ b/frontend/src/app/History.ts
@@ -30,7 +30,7 @@ export { history };
 export enum URLParam {
   AGGREGATOR = 'aggregator',
   BY_LABELS = 'bylbl',
-  CLUSTERMESH = 'clusterMesh',
+  CLUSTERNAME = 'clusterName',
   DIRECTION = 'direction',
   DISPLAY_MODE = 'displayMode',
   DURATION = 'duration',
--- a/frontend/src/components/BreadcrumbView/BreadcrumbView.tsx
+++ b/frontend/src/components/BreadcrumbView/BreadcrumbView.tsx
@@ -57,7 +57,7 @@ export class BreadcrumbView extends React.Component<BreadCumbViewProps, BreadCum
     let itemName = page !== 'istio' ? match[3] : match[5];
     return {
       namespace: ns,
-      cluster: urlParams.get('clusterMesh') || undefined,
+      cluster: urlParams.get('clusterName') || undefined,
       pathItem: page,
       item: itemName,
       itemName: ItemNames[page],
--- a/frontend/src/components/BreadcrumbView/BreadcrumbView.tsx
+++ b/frontend/src/components/BreadcrumbView/BreadcrumbView.tsx
@@ -57,7 +57,7 @@ export class BreadcrumbView extends React.Component<BreadCumbViewProps, BreadCum
     let itemName = page !== 'istio' ? match[3] : match[5];
     return {
       namespace: ns,
-      cluster: urlParams.get('cluster') || undefined,
+      cluster: urlParams.get('clusterMesh') || undefined,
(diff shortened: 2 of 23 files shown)


jmazzitelli:
> I wonder if I should also change it to keep consistency

I can tell you that this will cause massive confusion and should be changed so they are consistently the same:

```
	MeshCluster string `json:"meshCluster"`
```
```
		RateInterval: p.RateInterval, QueryTime: p.QueryTime, Cluster: p.ClusterMesh}
```

I can see people asking all the time, "Wait, should I use `MeshCluster` or `ClusterMesh` ??"

So, I'm +1 that you should change these to be consistent everywhere. I don't have an opinion one way or the other if it should be cluster-mesh or mesh-cluster but I say pick one and use it consistently everywhere.

josunect:
> meshCluster

Thanks for pointing at this. 

I realized there is other "MeshCluster" already: 

https://github.com/kiali/kiali/blob/master/frontend/src/types/Mesh.ts#L1

But I don't think they are the same, so maybe renaming to MeshCluster can be confusing?

jmazzitelli:
> I realized there is other "MeshCluster" already:
>
> But I don't think they are the same, so maybe renaming to MeshCluster can be confusing?

OK, we definitely have to refactor these. These names are too close, and if they really are two different things, it will be even more confusing :) 

So, let's figure out what to name both of these and change one or the other (or both?) so they aren't so close.

Maybe rename the new one literally with "Param" in the name??? `MeshClusterNameParam` ?? I'm just thinking out loud. But for sure we should not have `MeshCluster` and `ClusterMesh` especially if they are for two different things.

josunect:
> > I realized there is other "MeshCluster" already:
> > But I don't think they are the same, so maybe renaming to MeshCluster can be confusing?
> 
> OK, we definitely have to refactor these. These names are too close, and if they really are two different things, it will be even more confusing :)
> 
> So, let's figure out what to name both of these and change one or the other (or both?) so they aren't so close.
> 
> Maybe rename the new one literally with "Param" in the name??? `MeshClusterNameParam` ?? I'm just thinking out loud. But for sure we should not have `MeshCluster` and `ClusterMesh` especially if they are for two different things.

What about ClusterName? Is more simple. But, if you also want to include "Mesh" in the name, MeshClusterName sounds good (I think MeshClusterNameParam looks too long)

jmazzitelli:
> What about ClusterName?

That seems fine for the new parameter. At least it doesn't have "Mesh" in the name :)

josunect:
I have done some testing in a multi cluster environment

jmazzitelli (approved):
testing this way

1. Pull this PR
2. `make build build-ui container-build`
3. `hack/istio/multicluster/install-primary-remote.sh -c kubectl -kudi true`
4. `hack/istio/install-error-rates-demo.sh -c kubectl -dd true`
5. Open Kiali UI on east: `xdg-open https://$(minikube ip --profile east)/kiali/console`
6. Click on `Workloads -> (alpha namespace) -> a-client` and you see the URL as `https://192.168.59.253/kiali/console/namespaces/alpha/workloads/a-client?clusterName=east&duration=60&refresh=60000`

`clusterName` is what we are looking for. Change that in the URL to `cluster=foobar` and it will not affect anything. Change `clusterName=east` to `clusterName=foobar` in the URL and you get an error about a-client not found.

I clicked around Applications and Services and the same thing applies.

PR #6326, opened 2023-06-30
security scan notice

jmazzitelli:
We have been getting some security scans from the community where they are running reports and blindly sending them to us without doing any kind of validation or verification of those reports. We want to ask people to at least make sure (a) they are running them on the latest version and (b) that they aren't old security issues that don't affect Kiali in any way.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,6 +1,8 @@
 # Security Policy
 Kiali takes security seriously and encourages users to report security concerns.
 
+If you run a security scan on Kiali software and would like to report a security scan report to the Kiali team, we only ask that you first verify that your scan is correctly validating the latest release and that the results are valid. Because the Kiali team takes security reports seriously, they often take priority over current work being done, and it takes the Kiali team a long time to research and validate them. So please make sure you verify that any reports you submit are reporting an accurate reflection of the Kiali software being scanned and that the security issues being reported actually affect Kiali or one of its dependencies.
+
 For current security bulletins see https://kiali.io/news/security-bulletins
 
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -1,6 +1,8 @@
 # Security Policy
 Kiali takes security seriously and encourages users to report security concerns.
 
+If you run a security scan on Kiali software and would like to report a security scan report to the Kiali team, we only ask that you first verify that your scan is correctly validating the latest release and that the results are valid. Because the Kiali team takes security reports seriously, they often take priority over current work being done, and it takes the Kiali team a long time to research and validate them. So please make sure you verify that any reports you submit are reporting an accurate reflection of the Kiali software being scanned and that the security issues being reported actually affect Kiali or one of its dependencies.
```
jshaughn, line 4: If you run a security scan on Kiali software and would like to report a security scan report to the Kiali team, we only ask that you first verify that your scan is correctly validating the latest release and that the results are valid. Because the Kiali team takes security reports seriously, they often take priority over current work being done, and it takes the Kiali team a long time to research and validate them. So please make sure you verify that any reports you submit are reporting an accurate reflection of the Kiali software being scanned and that the security issues being reported actually affect Kiali or one of its dependencies.

jmazzitelli: There is no diff between your suggestion and what the original text is.

jmazzitelli: i committed the changes suggested over in the kiali.io pr

jshaughn (approved):
Goodest!

PR #6327, opened 2023-06-30
Actions improvements

### LINKED ISSUE #6324: Fix the Go version in Actions
Actions are not using the correct version of Go when building and releasing. 

This is also a good opportunity to add the "go check" step (found in the Makefile) to check for the version and fail if it is not the right one.

leandroberetta:
Fixes #6324.

In this PR:

- Setup Go honoring the variable GO_VERSION_KIALI found in the Makefile
- Simplify the options when building:

image

When the workflow is set from master, means we are building a minor version and when the workflow comes from other branches (usually prior versions like v1.70, etc) will are building a patch release from that branch's code.

If this PR is merged, this will be the new way of launching builds for master and also for the branches that has this definition backported, meaning, ideally we should backport this changes to the branches we usually release (OSSM versions). 

What happens if we don't backport this change? When choosing from the combo box where the workflow comes from, let's say we want to release a patch for v1.69, we are going to look for that workflow's branch and the old set of parameters will appear before releasing (will be like before). 

One important thing to notice is that after this change, we won't be able to release patch version using master's workflow (because we won't find anymore the options like branch to release and release type).

### DIFF
--- a/.github/workflows/build-backend.yml
+++ b/.github/workflows/build-backend.yml
@@ -15,13 +15,21 @@ jobs:
         uses: actions/checkout@v3
         with:
           ref: ${{ inputs.build_branch }}
+    
+      - name: Determine go version
+        id: go_version
+        run: |
+          GO_VERSION=$(sed -rn 's/^GO_VERSION_KIALI \= (.*)/\1/p' Makefile)
+
+          # Remove any pre release identifier (ie: "-SNAPSHOT")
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
This looks good.  If necessary, when merging please update any relevant build docs on the shared Kiali drive.

jmazzitelli (approved):
I'm Ok with this change.

Issue #6333, opened 2023-07-04
Review usages of GetNamespace with no cluster parameter 


josunect:
This TODO: https://github.com/kiali/kiali/blob/master/business/namespaces.go#L514 was created while not all the layers used the cluster parameter but shouldn't be needed anymore. Review usages.

PR #6334, opened 2023-07-04
Use cluster param when it is available

josunect:
Fixes https://github.com/kiali/kiali/issues/6333

There was some calls to getNamespace that didn't include the cluster parameter. But in a multicluster environment, a namespace with the same name can exist in different clusters, which can result in more than one namespace. getNamespace mehod was returning just one namespace, the first namespace found, which is a behavior non deterministic. 

This behavior was changed to use, when we want a namespace and we don't know the cluster (So we can expect to have more than one) GetNamespaceClusters, implemented as part as https://github.com/kiali/kiali/pull/6336/files.

The majority of the calls include the cluster parameter. Some metrics calls doesn't include it, because in the right panel summary for the graphs, we want the metrics for the namespaces for all the clusters, see https://github.com/kiali/kiali/pull/6131#pullrequestreview-1429313391

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -506,13 +506,6 @@ func (in *NamespaceService) isIncludedNamespace(namespace string) bool {
 	return false
 }
 
-// GetNamespace returns the definition of the specified namespace.
-// TODO: Multicluster: We are going to need something else to identify the namespace, the cluster (OR Return a list/array/map)
-func (in *NamespaceService) GetNamespace(ctx context.Context, namespace string) (*models.Namespace, error) {
-	// TODO: Wrapper for MC while other services are not updated to propagate the cluster
-	return in.GetNamespaceByCluster(ctx, namespace, "")
-}
-
 // GetNamespaceClusters is a convenience routine that filters GetNamespaces for a particular namespace
 func (in *NamespaceService) GetNamespaceClusters(ctx context.Context, namespace string) ([]models.Namespace, error) {
 	namespaces, err := in.GetNamespaces(ctx)
@@ -531,7 +524,6 @@ func (in *NamespaceService) GetNamespaceClusters(ctx context.Context, namespace 
 }
 
 // GetNamespace returns the definition of the specified namespace.
-// TODO: Multicluster: We are going to need something else to identify the namespace, the cluster (OR Return a list/array/map)
 func (in *NamespaceService) GetNamespaceByCluster(ctx context.Context, namespace string, cluster string) (*models.Namespace, error) {
 	var end observability.EndFunc
 	ctx, end = observability.StartSpan(ctx, "GetNamespaceByCluster",
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -315,6 +315,8 @@ func (in *NamespaceService) getNamespacesByCluster(cluster string) ([]models.Nam
 				}
 				namespaces = models.CastProjectCollection(filteredProjects, cluster)
 			}
+		} else {
+			return nil, err2
 		}
 	} else {
 		// if the accessible namespaces define a distinct list of namespaces, use only those.
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -392,8 +392,8 @@ func (in *NamespaceService) getNamespacesByCluster(cluster string) ([]models.Nam
 	return namespaces, nil
 }
 
-// GetNamespacesForCluster is just a convenience routine that filters GetNamespaces for a particular cluster
-func (in *NamespaceService) GetNamespacesForCluster(ctx context.Context, cluster string) ([]models.Namespace, error) {
+// GetClusterNamespaces is just a convenience routine that filters GetNamespaces for a particular cluster
+func (in *NamespaceService) GetClusterNamespaces(ctx context.Context, cluster string) ([]models.Namespace, error) {
 	tokenNamespaces, err := in.GetNamespaces(ctx)
 	if err != nil {
 		return nil, err
@@ -523,10 +523,10 @@ func (in *NamespaceService) GetNamespaceClusters(ctx context.Context, namespace 
 	return result, nil
 }
 
-// GetNamespace returns the definition of the specified namespace.
-func (in *NamespaceService) GetNamespaceByCluster(ctx context.Context, namespace string, cluster string) (*models.Namespace, error) {
+// GetClusterNamespace returns the definition of the specified namespace.
+func (in *NamespaceService) GetClusterNamespace(ctx context.Context, namespace string, cluster string) (*models.Namespace, error) {
 	var end observability.EndFunc
-	ctx, end = observability.StartSpan(ctx, "GetNamespaceByCluster",
+	ctx, end = observability.StartSpan(ctx, "GetClusterNamespace",
 		observability.Attribute("package", "business"),
 		observability.Attribute("namespace", namespace),
 		observability.Attribute("cluster", cluster),
@@ -593,7 +593,7 @@ func (in *NamespaceService) UpdateNamespace(ctx context.Context, namespace strin
 	defer end()
 
 	// A first check to run the accessible/excluded logic and not run the Update operation on filtered namespaces
-	_, err := in.GetNamespaceByCluster(ctx, namespace, cluster)
+	_, err := in.GetClusterNamespace(ctx, namespace, cluster)
 	if err != nil {
 		return nil, err
 	}
@@ -608,7 +608,7 @@ func (in *NamespaceService) UpdateNamespace(ctx context.Context, namespace strin
 	in.kialiCache.RefreshTokenNamespaces()
 
 	// Call GetNamespace to update the caching
-	return in.GetNamespaceByCluster(ctx, namespace, cluster)
+	return in.GetClusterNamespace(ctx, namespace, cluster)
 }
 
 func (in *NamespaceService) getNamespacesUsingKialiSA(cluster string, labelSelector string, forwardedError error) ([]core_v1.Namespace, error) {
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -75,7 +75,7 @@ func NewNamespaceService(userClients map[string]kubernetes.ClientInterface, kial
 // Returns a list of the given namespaces / projects
 func (in *NamespaceService) GetClusterList() []string {
 	var clusterList []string
-	for cluster, _ := range in.userClients {
+	for cluster := range in.userClients {
 		clusterList = append(clusterList, cluster)
 	}
 	return clusterList
@@ -284,7 +284,7 @@ func (in *NamespaceService) getNamespacesByCluster(cluster string) ([]models.Nam
 
 	labelSelectorInclude := configObject.API.Namespaces.LabelSelectorInclude
 
-	namespaces := []models.Namespace{}
+	var namespaces []models.Namespace
 	_, queryAllNamespaces := in.isAccessibleNamespaces["**"]
 	// If we are running in OpenShift, we will use the project names since these are the list of accessible namespaces
 	if in.hasProjects {
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -524,7 +524,6 @@ func (in *NamespaceService) GetNamespaceClusters(ctx context.Context, namespace 
 }
 
 // GetNamespace returns the definition of the specified namespace.
-// TODO: When cluster is "" it returns the first occurrence
 func (in *NamespaceService) GetNamespaceByCluster(ctx context.Context, namespace string, cluster string) (*models.Namespace, error) {
 	var end observability.EndFunc
 	ctx, end = observability.StartSpan(ctx, "GetNamespaceByCluster",
@@ -556,53 +555,23 @@ func (in *NamespaceService) GetNamespaceByCluster(ctx context.Context, namespace
 	var result models.Namespace
 	if in.hasProjects {
 		var project *osproject_v1.Project
-		// TODO: MC
-		if cluster == "" {
-			var err2 error
-			for cl := range in.userClients {
-				project, err2 = in.userClients[cl].GetProject(namespace)
-				if err2 == nil {
-					result = models.CastProject(*project, cluster)
-					break
-				}
-			}
-			if err2 != nil {
-				return nil, err2
-			}
-		} else {
-			if _, ok := in.userClients[cluster]; !ok {
-				return nil, fmt.Errorf("OCP Cluster [%s] is not found or is not accessible for Kiali", cluster)
-			}
-			project, errC := in.userClients[cluster].GetProject(namespace)
-			if errC != nil {
-				return nil, errC
-			}
-			result = models.CastProject(*project, cluster)
+		if _, ok := in.userClients[cluster]; !ok {
+			return nil, fmt.Errorf("OCP Cluster [%s] is not found or is not accessible for Kiali", cluster)
 		}
+		project, errC := in.userClients[cluster].GetProject(namespace)
+		if errC != nil {
+			return nil, errC
+		}
+		result = models.CastProject(*project, cluster)
 	} else {
-		// TODO: MC
 		var ns *core_v1.Namespace
 		var errC error
-		if cluster == "" {
-			for cl := range in.userClients {
-				ns, errC = in.userClients[cl].GetNamespace(namespace)
-				if errC == nil {
-					// Namespace found, assign that cluster
-					cluster = cl
-					break
-				}
-			}
-			if errC != nil {
-				return nil, errC
-			}
-		} else {
-			if _, ok := in.userClients[cluster]; !ok {
-				return nil, fmt.Errorf("Cluster [%s] is not found or is not accessible for Kiali", cluster)
-			}
-			ns, errC = in.userClients[cluster].GetNamespace(namespace)
-			if errC != nil {
-				return nil, errC
-			}
+		if _, ok := in.userClients[cluster]; !ok {
+			return nil, fmt.Errorf("Cluster [%s] is not found or is not accessible for Kiali", cluster)
+		}
+		ns, errC = in.userClients[cluster].GetNamespace(namespace)
+		if errC != nil {
+			return nil, errC
 		}
 
 		result = models.CastNamespace(*ns, cluster)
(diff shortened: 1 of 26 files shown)


handlers/utils.go
```diff
@@ -22,13 +22,59 @@ const defaultPatchType = "merge"
...
+		log.Error(err)
+		RespondWithError(w, http.StatusServiceUnavailable, "Prometheus client error: "+err.Error())
+		return nil, nil
+	}
+	var nsInfo nsInfoError
+
+	nsLists, err := layer.Namespace.GetNamespaceClusters(r.Context(), ns.Name)
+	if err != nil || len(nsLists) == 0 {
+		log.Infof("No clusters found.")
+		// TWe cant find the cluster list, we asume it is local
```
nrfox, line 48: Shouldn't the local cluster be included in `GetNamespaceClusters`? `err` could be nil for a variety of reasons so I'm not sure we can assume this?

handlers/utils.go
```diff
@@ -22,13 +22,59 @@ const defaultPatchType = "merge"
...
+			nsInfo = nsInfoError{info: info, err: nil}
+		}
+	}
+	metrics := business.NewMetricsService(prom)
+	if nsInfo.err != nil {
+		RespondWithError(w, http.StatusForbidden, "Cannot access namespace data: "+nsInfo.err.Error())
+		return nil, nil
+	}
+
+	return metrics, nsInfo.info
```
nrfox, line 71: For callers of this function, I don't think it's deterministic which namespace you'll get? Should this return a list of namespaces now instead of a single namespace? The namespace model is scoped to a specific cluster rather than spanning across all clusters even though that's how they're currently represented in the frontend.

handlers/workloads.go
```diff
@@ -68,7 +70,13 @@ func WorkloadList(w http.ResponseWriter, r *http.Request) {
...
 
 	if criteria.IncludeHealth {
-		rateInterval, err := adjustRateInterval(r.Context(), businessLayer, p.Namespace, p.RateInterval, p.QueryTime)
+		// When the cluster is not specified, we need to get it. If there are more than one, get the first one
+		clusters, _ := businessLayer.Namespace.GetNamespaceClusters(context.TODO(), p.Namespace)
+		if len(clusters) == 0 {
+			handleErrorResponse(w, err, "Error looking for cluster: "+err.Error())
+			return
+		}
+		rateInterval, err := adjustRateInterval(r.Context(), businessLayer, p.Namespace, p.RateInterval, p.QueryTime, clusters[0].Cluster)
```
nrfox, line 79: Kind of the same question here: is rate interval tied to a specific cluster or no? It seems like it would be random which cluster you got in `clusters[0].Cluster` which would make results inconsistent?

handlers/utils.go
```diff
@@ -22,13 +22,59 @@ const defaultPatchType = "merge"
...
-	return nsServ.GetNamespace(ctx, namespace)
+func checkNamespaceAccess(ctx context.Context, nsServ business.NamespaceService, namespace string, cluster string) (*models.Namespace, error) {
+	return nsServ.GetNamespaceByCluster(ctx, namespace, cluster)
 }
 
-func createMetricsServiceForNamespace(w http.ResponseWriter, r *http.Request, promSupplier promClientSupplier, namespace string) (*business.MetricsService, *models.Namespace) {
-	metrics, infoMap := createMetricsServiceForNamespaces(w, r, promSupplier, []string{namespace})
-	if result, ok := infoMap[namespace]; ok {
+// Create Metrics Service for a single namespace and no cluster parameter. It will check permissions for all the clusters, and return the service
+// if it has permissions at least one of the clusters
```
nrfox, line 30: Can you add to this comment _why_ we're doing permissions checks this way?

handlers/utils.go
josunect: Well, the local cluster should be included in `GetNamespaceCluster`. I found this problem in the tests, so I figured out that it could be any scenario when it is not possible to retrieve these list so we could make this assumption?

handlers/utils.go
josunect: Yes, you made a good point! 

I guess the question here would be, from the namespace info, the creation timestamp will be used, which one could we used for the metrics query? The newest/oldest? This would be for the query start time, https://github.com/kiali/kiali/blob/master/handlers/metrics.go#L300

handlers/utils.go
josunect: If we change to return a list of namespaces, maybe this is not needed and return error just in case of multicluster (Or if we don't have any permissions?)

handlers/utils.go
josunect: I will ping @jshaughn in case he has some ideas here

handlers/utils.go
josunect: I will ping @jshaughn in case he has some ideas here

jshaughn: I agree, a non-deterministic return should be avoided but I'm not sure a list makes sense here.  The caller only needs one metrics service.  The way metrics queries work, correct me if I'm wrong, it will either be against a single cluster (i.e. the request includes the `cluster` param), or across all clusters.   I think the caller is probably responsible for knowing what it's going to do.  And so I think this function probably should take a cluster argument.  If the namespace exists on the cluster and the user has access then return the metricsService.  If the cluster arg is "" then assume it's a cross-cluster metrics query and make sure the user has access to the namespace on all its clusters.  If so, return the metricsService.

As for the creation timestamp, that's sort of an odd one.  It's used to mainly to make sure we're not querying farther back than makes sense.  So for a single cluster query it should be set to the namespace on that cluster.  For all clusters ("") its hould be set to the oldest create time of the namespace clusters.

handlers/workloads.go
```diff
@@ -68,7 +70,13 @@ func WorkloadList(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if criteria.IncludeHealth {
-		rateInterval, err := adjustRateInterval(r.Context(), businessLayer, p.Namespace, p.RateInterval, p.QueryTime)
+		// When the cluster is not specified, we need to get it. If there are more than one, get the first one
+		clusters, _ := businessLayer.Namespace.GetNamespaceClusters(context.TODO(), p.Namespace)
+		if len(clusters) == 0 {
+			handleErrorResponse(w, err, "Error looking for cluster: "+err.Error())
```
jshaughn, line 76: Same issue again.

handlers/workloads.go
jshaughn: See above, the createTimestamp should be set to the Namespace on the specific cluster, or the oldest time for all of the namespace clusters.

jshaughn (changes requested):

handlers/utils.go
josunect: I just found this issue in testing scenario. I added a TODO to review further.

handlers/utils.go
josunect: Thanks for clarifying. I have changed it to return the oldest one, and return an error if there are not permissions in one of the namespaces. (Maybe we should just get into account the namespaces that are coming from the clusters that the user has permissions from?)

Regarding to use the cluster argument, there is a separate function that has the cluster argument and returns the metrics service. I've thought it would be more clear to have a separate function as it can have different logic to just handle this specific situation. Maybe it makes more sense to return a single metrics service but a list of namespace info?

handlers/workloads.go
josunect: Updated, thanks

business/namespaces.go
```diff
@@ -535,7 +528,6 @@ func (in *NamespaceService) GetNamespaceClusters(ctx context.Context, namespace
 }
 
 // GetNamespace returns the definition of the specified namespace.
-// TODO: Multicluster: We are going to need something else to identify the namespace, the cluster (OR Return a list/array/map)
```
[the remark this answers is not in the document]
josunect: Renamed

handlers/apps.go
```diff
@@ -59,7 +60,13 @@ func AppList(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if criteria.IncludeHealth {
-		rateInterval, err := adjustRateInterval(r.Context(), business, p.Namespace, p.RateInterval, p.QueryTime)
+		// When the cluster is not specified, we need to get it. If there are more than one, the home cluster will be used
```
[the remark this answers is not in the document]
josunect: Updated to get the one with oldest creation timestamp

handlers/apps.go
```diff
@@ -59,7 +60,13 @@ func AppList(w http.ResponseWriter, r *http.Request) {
...
+		clusters, _ := business.Namespace.GetNamespaceClusters(context.TODO(), p.Namespace)
+		if len(clusters) == 0 {
+			handleErrorResponse(w, err, "Error looking for cluster: "+err.Error())
```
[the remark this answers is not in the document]
josunect: Thanks. Updated.

handlers/services.go
```diff
@@ -63,7 +65,13 @@ func ServiceList(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if criteria.IncludeHealth {
-		rateInterval, err := adjustRateInterval(r.Context(), business, p.Namespace, p.RateInterval, p.QueryTime)
+		// When the cluster is not specified, we need to get it. If there are more than one, get the first one
+		clusters, _ := business.Namespace.GetNamespaceClusters(context.TODO(), p.Namespace)
+		if len(clusters) == 0 {
+			handleErrorResponse(w, err, "Error looking for cluster: "+err.Error())
```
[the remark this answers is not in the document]
josunect: Updated.

handlers/utils.go
```diff
@@ -22,13 +22,59 @@ const defaultPatchType = "merge"
...
+	}
+	prom, err := promSupplier()
+	if err != nil {
+		log.Error(err)
+		RespondWithError(w, http.StatusServiceUnavailable, "Prometheus client error: "+err.Error())
+		return nil, nil
+	}
+	var nsInfo nsInfoError
+
+	nsLists, err := layer.Namespace.GetNamespaceClusters(r.Context(), ns.Name)
```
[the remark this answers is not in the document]
josunect: Updated.

handlers/utils.go
```diff
@@ -22,13 +22,59 @@ const defaultPatchType = "merge"
...
+		nsInfo = nsInfoError{info: info, err: err}
+		if err != nil {
+			nsNoErrors--
+		}
+		// At least, one ns with permissions
+		if nsNoErrors > 0 {
+			nsInfo = nsInfoError{info: info, err: nil}
+		}
+	}
+	metrics := business.NewMetricsService(prom)
```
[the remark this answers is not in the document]
josunect: Rewritten.

handlers/workloads.go
josunect: updated.

jshaughn (changes requested):

handlers/utils.go
josunect: I was going to add the list o namespaces to the extractIstioMetricsQueryParams but, there are other methods that are using it and they just have a single namespace. So, in order to not to change them, I have created an aux method to have just this functionality.

handlers/utils.go
```diff
@@ -22,13 +22,56 @@ const defaultPatchType = "merge"
...
+	}
+	var nsInfo nsInfoError
+
+	nsClusters, err := layer.Namespace.GetNamespaceClusters(r.Context(), ns.Name)
+	if err != nil || len(nsClusters) == 0 {
+		log.Infof("No clusters found.")
+		//TODO: We cant find the cluster list, we assume it is local
+		nsClusters = make([]models.Namespace, 1)
+		nsClusters[0] = models.Namespace{Cluster: config.Get().KubernetesConfig.ClusterName, Name: ns.Name}
+	}
```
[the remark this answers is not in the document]
josunect: Well, the tests are failing without this logic, just the tests, and there is just one single cluster. But it looks to be working for single cluster, so I thought there was something not well initialized but maybe this is just in the tests. I'm reviewing.

handlers/services.go
```diff
@@ -63,7 +66,22 @@ func ServiceList(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		var cluster string
+		var creationTimestamp time.Time
+		for i, cl := range clusters {
+			if i == 0 || cl.CreationTimestamp.Before(creationTimestamp) {
+				cluster = cl.Cluster
+			}
+		}
```
jshaughn, line 83: Can this be replaced with your new utility?

handlers/utils.go
```diff
@@ -22,13 +22,65 @@ const defaultPatchType = "merge"
...
+	var nsInfo []models.Namespace
+
+	nsClusters, err := layer.Namespace.GetNamespaceClusters(r.Context(), nsName)
+	if err != nil || len(nsClusters) == 0 {
+		log.Infof("No clusters found.")
+		//TODO: We cant find the cluster list, we assume it is local
+		nsClusters = make([]models.Namespace, 1)
+		nsClusters[0] = models.Namespace{Cluster: config.Get().KubernetesConfig.ClusterName, Name: nsName}
+	}
```
jshaughn, line 51: Unless you can explain in detail why we need this behavior I still think this should just `RespondWithError`.

handlers/utils.go
```diff
@@ -22,13 +22,65 @@ const defaultPatchType = "merge"
...
+
+	}
+	metrics := business.NewMetricsService(prom)
+
+	return metrics, nsInfo
+}
+
+// Get namespespace with oldest creation time
+// Used to choose between one namespace from all the clusters and use the creation timestamp for the metrics service
+func GetNsWithOldestCreationDate(namespaces []models.Namespace) *models.Namespace {
```
jshaughn, line 70:
```suggestion
// GetOldestNamespace is a convenience function that takes a list of Namespaces and returns the
// Namespace with the oldest CreationTimestamp.  In a tie, preference is towards the head of the list.
func GetOldestNamespace(namespaces []models.Namespace) *models.Namespace {
```

Function comments in Go typically start with the function name.  In this case I suggest we also shorten the name and remove the abbreviation, but you can decide what you like best.

jshaughn (changes requested):

handlers/apps.go
```diff
@@ -59,7 +61,22 @@ func AppList(w http.ResponseWriter, r *http.Request) {
...
+		clusters, _ := business.Namespace.GetNamespaceClusters(r.Context(), p.Namespace)
+		if len(clusters) == 0 {
+			err = fmt.Errorf("No clusters found for namespace  [%s]", p.Namespace)
+			handleErrorResponse(w, err, err.Error())
+			return
+		}
+		var cluster string
+		var creationTimestamp time.Time
+		for i, cl := range clusters {
+			if i == 0 || cl.CreationTimestamp.Before(creationTimestamp) {
```
nrfox, line 75: Where does `creationTimestamp` get initialized? If it's uninitialized then I think you'll always get `cluster = clusters[0]` because of the first condition `i == 0`. This would be a good thing to write a test for where there's more than one namespace and the rate interval gets adjusted according to the oldest.

handlers/apps.go
```diff
@@ -59,7 +61,22 @@ func AppList(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if criteria.IncludeHealth {
-		rateInterval, err := adjustRateInterval(r.Context(), business, p.Namespace, p.RateInterval, p.QueryTime)
+		// When the cluster is not specified, we need to get it. If there are more than one,
+		// get the one for which the namespace creation time is oldest
+		clusters, _ := business.Namespace.GetNamespaceClusters(r.Context(), p.Namespace)
```
nrfox, line 66: nit: these are `namespaces` not `clusters` right?

handlers/apps.go
nrfox: Actually it looks like there's a helper func `GetOldestNamespace` that is implemented differently. Perhaps you can just use that here instead?

handlers/workloads.go
```diff
@@ -68,7 +70,23 @@ func WorkloadList(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		var cluster string
+		var creationTimestamp time.Time
+		for i, cl := range clusters {
+			if i == 0 || cl.CreationTimestamp.Before(creationTimestamp) {
+				cluster = cl.Cluster
+			}
+		}
```
nrfox, line 87: Same comment here about `creationTimestamp` and instead using the `GetOldestNamespace` helper func.

handlers/services.go
josunect: Yes, makes sense

handlers/utils.go
josunect: Update to return the clusters name list using the user clients map index. 

Using GetNamespaceClusters was not working in this case because we were returning an error, but we want to get all the whole list and then check the access for each particular namespace.

handlers/utils.go
josunect: Updated

handlers/apps.go
josunect: Sure, update to use GetOldestNamespace

handlers/workloads.go
josunect: Updated.

handlers/apps.go
josunect: That's right! Updated.

business/namespaces.go
```diff
@@ -68,6 +68,15 @@ func NewNamespaceService(userClients map[string]kubernetes.ClientInterface, kial
 	}
 }
 
+// Returns a list of the given namespaces / projects
```
jshaughn, line 71: GetClusterList returns a list of the clusters (actually the userClients for the NamespaceService)

jshaughn (changes requested):
Looks good, one trivial change requested and then can approve!

business/namespaces.go
josunect: Updated

jshaughn (approved):
Looks good to me!

Update: Well, now it has conflicts...

josunect:
> Looks good to me!
> 
> Update: Well, now it has conflicts...

Ops :) Now rebased.

jshaughn (approved):
OK, conflicts are resolve, looks good to me!

Issue #6331, opened 2023-07-03
Graph IdleNodes option is broken


jshaughn:
The graph's IdleNode Display option broke at some point during the multi-cluster work.

jmazzitelli:
This may still be broken. Turn off "service nodes" and you get the "empty graph" message again:

![image](https://github.com/kiali/kiali/assets/2029470/3e705a03-63c3-4e8e-b310-12287b5baf25)

turn on "service nodes" and now I see the idle nodes:

![image](https://github.com/kiali/kiali/assets/2029470/3a65da73-4e5b-49c7-9848-842c92288bb3)

jshaughn:
> This may still be broken. Turn off "service nodes" and you get the "empty graph" message again:

This is unrelated to this bug and this bug fix.   For better or for worse, we don't look for unused services unless it's a service graph or the Show Service Nodes option is selected.  It's been this way for 4 years, so not going to make a change unless a real user complains.

PR #6336, opened 2023-07-05
Fix issue with IdleNodes detection

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6331

Fix issue with IdleNodes detection. The multi-cluster changes resulted in only looking at clusters WITH traffic, but for idle nodes we need to look at all clusters for a namespaces, regardless of traffic.

- Adds new NamespaceService.GetNamespaceClusters() utility function

https://github.com/kiali/kiali/issues/6331

Test Note:
To recreate this bug you need to have NO traffic for any services/workloads in the target namespace, for the time period.

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -517,6 +517,23 @@ func (in *NamespaceService) GetNamespace(ctx context.Context, namespace string) 
 	return in.GetNamespaceByCluster(ctx, namespace, "")
 }
 
+// GetNamespaceClusters is a convenience routine that filters GetNamespaces for a particular namespace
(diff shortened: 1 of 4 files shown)


josunect (approved):
LGTM

It works as expected: 

image

image


And code LG

jmazzitelli (approved):
My simple test now works.

Just install Istio, Kiali with nothing else in the cluster. Ask to see the graph with idle nodes. Before when I noticed the problem, the graph just showed the "no nodes to display" message. Now it shows what it originally did before the bug was introduced - all the idle nodes (without edges).

Issue #6270, opened 2023-06-18
Align Patternfly version with Openshift 4.13


ferhoyos:
Update Patternfly version to the one used in latest Openshift version 4.13

jshaughn:
Looking at this branch on June 28, 2023, I see:

https://github.com/openshift/console/tree/release-4.13
https://github.com/openshift/console/blob/release-4.13/frontend/package.json

```
    "@patternfly/patternfly": "4.224.2",
    "@patternfly/quickstarts": "2.3.1",
    "@patternfly/react-catalog-view-extension": "4.95.1",
    "@patternfly/react-charts": "6.94.18",
    "@patternfly/react-core": "4.276.6",
    "@patternfly/react-log-viewer": "4.87.100",
    "@patternfly/react-table": "4.112.39",
    "@patternfly/react-tokens": "4.94.6",
    "@patternfly/react-topology": "4.91.27",
    "@patternfly/react-virtualized-extension": "4.88.113",

```

Also, we likely want to minimally match the relevant react-related versions:

```
    "react": "^17.0.1",
    "react-copy-to-clipboard": "5.x",
    "react-dnd": "^9.4.0",
    "react-dnd-html5-backend": "^9.4.0",
    "react-dom": "^17.0.1",
    "react-draggable": "4.x",
    "react-helmet": "^6.1.0",
    "react-i18next": "^11.7.3",
    "react-linkify": "^0.2.2",
    "react-measure": "^2.2.6",
    "react-modal": "^3.12.1",
    "react-monaco-editor": "0.41.x",
    "react-redux": "7.2.2",
    "react-rnd": "^10.3.4",
    "react-router": "5.2.0",
    "react-router-dom": "5.2.0",
    "react-router-hash-link": "^2.0.0",
    "react-tagsinput": "3.19.x",
    "react-transition-group": "2.3.x",
    "react-virtualized": "9.x",

```

PR #6337, opened 2023-07-05
Update PF and React dependencies to match Console in OpenShift 4.13.

jshaughn:
Closes https://github.com/kiali/kiali/issues/6270

A useful update for upstream but also to support console plugin work.


jshaughn:
@ferhoyos , @josunect , The tests are passing and I have clicked around in many places and I don't see any issues with this update.  Maybe you can sanity check and see if you can see anything unusual.  Thanks!

frontend/package.json
ferhoyos, line 1: Maybe it is not the intention of this issue (only patternfly and react libraries), but I have extended the analysis to whole package.json of openshift 4.13 and there are more differences that we should take into account for our package.json:
- Use of lodash-es instead of lodash: lodash-es uses ES-6 modules, allowing tree shaking and therefore having a smaller size bundle (https://til.hashrocket.com/posts/dnzttruljf-prefer-lodash-es-when-using-webpack). Also I have noticed that we are using lodash in 3 different ways within Kiali:
```
import _ from 'lodash';
import { isArray } from 'lodash';
import _isEqual from 'lodash/isEqual';
```
Library use should be consistent through the whole Kiali application, but that's for another PR. 

- Add caret to js-yaml to reuse same dependency version with other libraries:
```
"js-yaml": "^3.13.1"
```

- Update typesafe-actions to version 4.2.1:
```
"typesafe-actions": "^4.2.1",
```

- Add @types of react-virtualized:
```
"@types/react-virtualized": "9.x",
```

Bonus: 
`seedrandom` library is only used in tests and therefore it should go to devDependencies.

ferhoyos (approved):
I have done some smoke tests and I do not see any issue. If the scope of this PR is just patternfly and react libraries, I approve as-is. @jshaughn, take a look to my comment, I would like to know your opinion if we should extend the alignment to whole package.json.

josunect (approved):
I did some tests in a multi cluster environment, no issues found. LG!

frontend/package.json
jshaughn: I like the idea of extending things a bit more to match console, it seems like it would reduce any chance of a strange interaction in the plugin and reduce the chance of CVEs.

I reviewed and agree with the suggested updates.  I also added caret to `axios": "^0.21.4"`

Pushed changes, setting @ferhoyos for re-review, thanks!

ferhoyos (approved):
LGTM

Issue #6266, opened 2023-06-17
Add prefix to CSS classnames


ferhoyos:
In order to know which CSS classnames belong to a specific Openshift plugin, it is required to add a prefix. The library we use to create dynamic CSS classnames (typestyle) allows the use of that prefix, so we will have to create some util to add that prefix to all CSS classes generated by typestyle.

ferhoyos:
Closes via #6340.

PR #6340, opened 2023-07-07
Add prefix to generated typestyle classNames

ferhoyos:
** Describe the change **

Add prefix to classNames generated by typestyle library.

** Issue reference **

Part of #6266.

Testing PR:
Check that all classNames generated by typestyle library contains 'kiali' prefix (hardcoded value for Kiali standalone, environment variable for OSSMC).

For example:
```
<div class="kiali_f1r2nm0v">
```


jshaughn (approved):
Touches a lot of files but the root of the change is small, and it looks good.  Approving and merging.

PR #6348, opened 2023-07-11
v1.65 - remote istiod obtain Istio deployment name 

josunect:
Reported failure for interop tests: 

`deployments.apps "istiod-install-istio-system" not found`

The Istio deployment name was hard coded in the tests.


FilipB (approved):
LGTM

PR #6350, opened 2023-07-13
Convert custom static CSS to typestyle

### LINKED ISSUE #6349: Convert all custom static CSS to typestyle
Currently CSS style in Kiali is implemented via two different ways:
- Dynamic CSS classnames created with typestyle library
- Static CSS classes in multiple SCSS files

Static CSS classes must be converted to typestyle to be able to add prefix classname as specified in #6266. 

Besides, most of the static CSS classes modifies top HTML elements (html, body, etc.) which is not allowed in OSSMC (only style of the plugin stuff can be modified). The same happens for CSS variables definition. Converting them to dynamic CSS classes with typestyle and apply them to top level of each application (root at Kiali standalone, plugin page at OSSMC) would solve the issue.

ferhoyos:
** Describe the change **

In order to create a dynamic style that can be attached to specific element (body in standalone Kiali, OSSMC section page in Openshift console), I have replaced Kiali global CSS files into typestyle classnames. Since CSS variables cannot be converted to typestyle (see https://github.com/typestyle/typestyle/issues/285), I have created a CSS module that also generates a prefixed className (more information about CSS modules in https://github.com/css-modules/css-modules).

Both typestyle and CSS modules are scoped locally, not interfering in global CSS classnames and variables if Kiali code is added as plugin of bigger application (e.g., OSSMC)

The same have done for specific components that still contain custom static CSS.

** Issue reference **

Closes #6349 

** Testing PR **

Since this is just a refactorization, the PR testing is simple. Deploy two kiali instances (one of them with PR changes) and check that both kiali applications have the same look and feel in all pages.


jshaughn (approved):
@ferhoyos I sanity checked on Chrome and FF and could not visually detect any differences in styling.  To me this looks like another nice cleanup/refactor, where the main code base again benefits from the changes needed for the console plugin. Nice!

josunect (approved):
I also tested locally and it LGTM!

Just one question, why `npm run build-css` is not longer needed?

ferhoyos:
> I also tested locally and it LGTM!
> 
> Just one question, why `npm run build-css` is not longer needed?

`npm run build-css` were used to convert SCSS files into CSS and then imported from Javascript file. CSS modules work in a different way and they are managed entirely by webpack, so SCSS-to-CSS conversion is not longer needed.

PR #6352, opened 2023-07-13
[hack] support ambient by telling bookinfo install script if you want injection and if so what kind

jmazzitelli:
This supports ambient mode. The default behavior stays the same (which is auto-injecting the sidecar), however, if you turn off auto-injection (`-ai false`) then no sidecars are injected. If you do this, a message will spit out telling you what you have to do to enable ambient mode (you are told what command to use to label the namespace).

Note that this still retains the ability to manually inject the sidecars via istioctl (you do this using the `-mi` option )

### DIFF
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -26,6 +26,7 @@ MANUAL_INJECTION="false"
 DELETE_BOOKINFO="false"
 MINIKUBE_PROFILE="minikube"
 ARCH="amd64"
+WAIT_TIMEOUT="0" # can be things like "60s" or "30m"
 
 AMBIENT_ENABLED="false" # the script will set this to true only if Ambient is enabled and no sidecars are injected
 
@@ -89,6 +90,10 @@ while [[ $# -gt 0 ]]; do
       TRAFFIC_GENERATOR_ENABLED="true"
       shift;
       ;;
+    -wt|--wait-timeout)
+      WAIT_TIMEOUT="$2"
+      shift;shift
+      ;;
     -h|--help)
       cat <<HELPMSG
 Valid command line arguments:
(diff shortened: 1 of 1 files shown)


leandroberetta (dismissed):
Looks good

jmazzitelli:
@josunect when I run this now with `-ai false`, and then I label the bookinfo namespace :
```
$ kubectl get ns bookinfo --show-labels 
NAME       STATUS   AGE   LABELS
bookinfo   Active   15m   istio.io/dataplane-mode=ambient,kubernetes.io/metadata.name=bookinfo
```
I see in Kiali UI that it detects Ambient, but I'm not seeing any graph. Do you know if there is something else that is needed in order to get the graph to show traffic?

I do have the traffic generator running and it seems to be getting productpage successfully, but I'm also seeing some 503s, too:

```
$ kubectl logs -n bookinfo -l app=kiali-traffic-generator

Requests      [total, rate, throughput]         1031, 1.00, 0.48
Duration      [total, attack, wait]             17m10s, 17m10s, 45.633ms
Latencies     [min, mean, 50, 90, 95, 99, max]  954.076µs, 31.743ms, 24.644ms, 56.848ms, 63.582ms, 72.538ms, 870.28ms
Bytes In      [total, mean]                     2366539, 2295.38
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           47.72%
Status Codes  [code:count]                      200:492  503:539  
Error Set:
503 Service Unavailable
```

So I'm wondering what else needs to be done to get ambient to work with Kiali?

josunect:
When I tested, I was following these instructions: 

https://istio.io/latest/docs/ops/ambient/getting-started/

They apply an Authorization policy to allow send traffic to an application with a different service account. 

Another thing is that in the example, in kind, there are 3 nodes created. I think it is supposed to be L4 processing in the local node, but, maybe it is another difference.

jmazzitelli:
> They apply an Authorization policy to allow send traffic to an application with a different service account.

OK, let me see if it would be appropriate if I added the Auth Policy creation to this hack script, too. I think I'll add an `--enable-ambient` option here so it can be told to also do things like label the namespace.

jmazzitelli:
There is now a `--wait-timeout` option. By default it is "0" which means it won't wait (thus this is backward compat with previous behavior). But if you set it like `-wt 5m` or `--wait-timeout 60s` the script will wait for that amount of time, waiting for the pods in the new bookinfo namespace to come up.

jmazzitelli:
OK, this works, but ambient doesn't work on minikube so you can't test this on minikube.
I'll leave this open for review and we can merge at any time.

josunect:
Tested with kind and latest 1.19-dev

```
istio/download-istio.sh -div 1.19-dev
istio/install-istio-via-istioctl.sh -c kubectl -cp ambient -it default -ih default
istio/install-bookinfo-demo.sh -c kubectl -ai false -tg
```

image

josunect (approved):
LGTM
Tested with kind, it works as expected: 

image

PR #6364, opened 2023-07-17
Make cache mandatory and remove some cache config opts

### LINKED ISSUE #5599: Remove option to disable kube caching

nrfox:
Cleanup `cache != nil` checks since cache has been mandatory for awhile. Also removes some caching options from the kiali config.

Fixes #5599

### DIFF
--- a/business/authentication/openid_auth_controller_test.go
+++ b/business/authentication/openid_auth_controller_test.go
@@ -18,6 +18,7 @@ import (
 	osproject_v1 "github.com/openshift/api/project/v1"
 	"github.com/stretchr/testify/assert"
 	"gopkg.in/square/go-jose.v2"
+	core_v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd/api"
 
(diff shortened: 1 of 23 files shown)


nrfox:
Since this changes the kiali config it will also require separate changes to the docs and helm/operator repos.

jmazzitelli (approved):
* unit tests pass
* I installed and then in the ConfigMap I set:
```
    kubernetes_config:
      cache_istio_types: "foo"
      cache_namespaces: "foo"
```
and no errors occurred and cache started fine:
```
$ kubectl logs -n istio-system -l app=kiali
...
2023-07-18T11:34:41Z INF [Kiali Cache] Waiting for cluster-scoped cache to sync
2023-07-18T11:34:41Z INF [Kiali Cache] Started
2023-07-18T11:34:41Z INF [Kiali Cache] Kube cache is active for cluster: [Kubernetes]
```

jmazzitelli:
This won't require any changes to the helm charts. The three settings being completely removed here (`cache_istio_types`, `cache_namespaces`, `cache_enabled`) are not directly referenced in the helm charts.

Issue #6366, opened 2023-07-18
Multicluster - unbound variable in kiali-deploy


josunect:
When running the `istio/multicluster/install-primary-remote.sh` script with the following arguments: 

`istio/multicluster/install-primary-remote.sh -c kubectl -kudi true -kshc ~/dev/kiali_sources/helm-charts/_output/charts/kiali-server-1.72.0-SNAPSHOT.tgz`

It fails with the following error: 

```
Preparing remote cluster secret for single Kiali install in multicluster mode.
/home/jcordoba/dev/kiali_sources/kiali/hack/istio/multicluster/deploy-kiali.sh: line 88: remote_url_flag: unbound variable
```

jmazzitelli:
@josunect try this: https://github.com/kiali/kiali/pull/6367

PR #6367, opened 2023-07-18
ensure var is alway set

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/6366


josunect (approved):
LGTM, fixes the error

PR #6371, opened 2023-07-18
Use remote secret in hack run script

nrfox:
Fixes `hack/run-kiali.sh` script by creating a temp kubeconfig and using that as a [remote secret](https://github.com/kiali/kiali#observing-a-remote-cluster) rather than trying to use the "in cluster config".

### DIFF
--- a/hack/run-kiali.sh
+++ b/hack/run-kiali.sh
@@ -649,7 +649,8 @@ EOM
     chmod 'u=r,go=' "${TOKEN_FILE}" "${CA_FILE}"
 
     infomsg "Setting up kubeconfig at [${kubeconfig}]"
-    ${CLIENT_EXE} config set-cluster ${KUBE_CONTEXT} --kubeconfig="${kubeconfig}" "--server=https://${KUBERNETES_API_HOST}:${KUBERNETES_API_PORT}" "--certificate-authority=${CA_FILE}"
+    # Embedding the ca cert so we don't have pathing issues.
+    ${CLIENT_EXE} config set-cluster ${KUBE_CONTEXT} --kubeconfig="${kubeconfig}" "--server=https://${KUBERNETES_API_HOST}:${KUBERNETES_API_PORT}" --certificate-authority="${CA_FILE}" --embed-certs
     ${CLIENT_EXE} config set-credentials ${KUBE_CONTEXT} --kubeconfig="${kubeconfig}" --token="$(cat ${TOKEN_FILE})"
     ${CLIENT_EXE} config set-context ${KUBE_CONTEXT} --kubeconfig="${kubeconfig}" --user=${KUBE_CONTEXT} --cluster=${KUBE_CONTEXT} --namespace=${ISTIO_NAMESPACE}
     ${CLIENT_EXE} config use-context --kubeconfig="${kubeconfig}" ${KUBE_CONTEXT}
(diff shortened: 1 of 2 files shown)


hack/run-kiali.sh
```diff
@@ -824,25 +825,11 @@ cleanup_and_exit() {
   kill_port_forward_prometheus
   kill_port_forward_grafana
   kill_port_forward_tracing
-  restore_original_context
 
   exitmsg "Exiting"
   exit ${1:-0}
 }
 
-restore_original_context() {
```
jmazzitelli, line 833: I'm looking through this PR now and trying to figure out what its doing and why :)

But I wonder why this is being removed - if you do not use the current context, that means this script is going to create one and then switch to it. But when the script exits, you want to put the user back to what his context originally was... otherwise, the user will be confused why he isn't back to the way things worked before he started the script.

jmazzitelli: I think i see it - you are using a completely different kubeconfig file.. using the user's own and its current context won't change.

jmazzitelli (changes requested):
I haven't tested this yet, but I have a minor suggestion to clean up that unused var

hack/run-kiali.sh
nrfox: > I think i see it - you are using a completely different kubeconfig file.. using the user's own and its current context won't change.

Yes rather than changing the user's kubeconfig, this change either uses their existing kubeconfig as it is or creates a new one using the kiali SA creds but either way the existing kubeconfig isn't changed. Since the "remote secret" takes a kubeconfig file as input, this seemed easier and doesn't require messing with the user's kubeconfig at all.

hack/run-kiali.sh
```diff
@@ -637,39 +638,39 @@ metadata:
...
-  fi
-  infomsg "Setting up context [${KUBE_CONTEXT}]"
-  ${CLIENT_EXE} config set-cluster ${KUBE_CONTEXT} "--server=https://${KUBERNETES_API_HOST}:${KUBERNETES_API_PORT}" "--certificate-authority=${CA_FILE}"
-  ${CLIENT_EXE} config set-credentials ${KUBE_CONTEXT} "--token=$(cat ${TOKEN_FILE})"
-  ${CLIENT_EXE} config set-context ${KUBE_CONTEXT} --user=${KUBE_CONTEXT} --cluster=${KUBE_CONTEXT} --namespace=${ISTIO_NAMESPACE}
-  ${CLIENT_EXE} config use-context ${KUBE_CONTEXT}
-fi
+  rm -f "${REMOTE_SECRET_FILE}"
+
+  ln -s "${kubeconfig}" "${REMOTE_SECRET_FILE}"
```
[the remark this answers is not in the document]
nrfox: So this is confusing but _this_ remote secret is different from the multi-cluster remote secrets. Really the two should be combined but there's currently a few different places in the kiali server that try to get the "local client" so we need still need to distinguish between the "local" remote secret and the multi-cluster remote secrets. The local remote secret deployment has been an experimental feature for awhile. It'd be good to graduate that if we are going to continue support for it.

josunect:
I got a nil pointer error when running this. 
I set up the multi cluster environment, run the hack script and got this: 

```
secret/runkiali-secret unchanged
📄 Setting up kubeconfig at [/tmp/run-kiali/kubeconfig]
Cluster "kiali-developer" set.
User "kiali-developer" set.
Context "kiali-developer" created.
Switched to context "kiali-developer".
```

But, when trying to access the UI I have a nil pointer: 

```
2023/07/19 08:30:07 http: panic serving 127.0.0.1:42084: runtime error: invalid memory address or nil pointer dereference
goroutine 63 [running]:
net/http.(*conn).serve.func1()
        /usr/local/go/src/net/http/server.go:1854 +0x148
panic({0x295e080, 0x3d18310})
        /usr/local/go/src/runtime/panic.go:890 +0x262
github.com/kiali/kiali/business.Get(0xc00050c800)
        /home/jcordoba/dev/kiali_sources/kiali/business/layer.go:125 +0x59
github.com/kiali/kiali/handlers.getBusiness(0xc00050c900)
        /home/jcordoba/dev/kiali_sources/kiali/handlers/utils.go:89 +0xfc
```

The same environment, in master and with the old `GetConfigForLocalCluster` works fine: 

image

josunect (approved):
I've tested and it works as expected!

nrfox:
Thanks for testing @josunect!

PR #6376, opened 2023-07-19
Added options to set each of the cluster's context name to the Error Rates demo install script

matejnesuta:
I wanted to run the Cypress regression on my Kind MC setup, but the script didn't have a way to setup custom contexts. This would be super useful, because the context names in kind are different to the default `east` and `west`. I added the options and also added some echoes related.


hack/istio/install-error-rates-demo.sh
```diff
@@ -81,6 +91,9 @@ echo NAMESPACE_ALPHA=${NAMESPACE_ALPHA}
 echo NAMESPACE_BETA=${NAMESPACE_BETA}
 echo NAMESPACE_GAMMA=${NAMESPACE_GAMMA}
 echo SOURCE=${SOURCE}
+echo DISTRIBUTE_DEMO=${DISTRIBUTE_DEMO}
+echo CLUSTER1_CONTEXT=${CLUSTER1_CONTEXT}
+echo CLUSTER2_CONTEXT=${CLUSTER2_CONTEXT}
```
jshaughn, line 96: This list is alphabetical, please add these alphabetically (and also fix CLIENT_EXE because it is out of order).

jshaughn (changes requested):
Looks good, one small change needed.

hack/istio/install-error-rates-demo.sh
matejnesuta: Sorted as requested

PR #6231, opened 2023-06-05
Upgrade go-restful to latest one

aljesusg:
update k8s client-go

	k8s.io/api v0.27.2
	k8s.io/apimachinery v0.27.2
	k8s.io/client-go v0.27.2

This updated the requirement go-restful library to 3.9

PR #6377, opened 2023-07-19
[v1.66] Backport error checks for registry handler

nrfox:
Backporting error checks from #6231 since the linter complains about this during release.

### DIFF
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -383,67 +383,106 @@ func (c *kubeCache) createIstioInformers(namespace string) istio.SharedInformerF
 		if c.CheckIstioResource(kubernetes.AuthorizationPolicies) {
 			lister.authzLister = sharedInformers.Security().V1beta1().AuthorizationPolicies().Lister()
 			lister.cachesSynced = append(lister.cachesSynced, sharedInformers.Security().V1beta1().AuthorizationPolicies().Informer().HasSynced)
-			sharedInformers.Security().V1beta1().AuthorizationPolicies().Informer().AddEventHandler(c.registryRefreshHandler)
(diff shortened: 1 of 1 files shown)


kubernetes/cache/kube_cache.go
```diff
@@ -458,12 +497,18 @@ func (c *kubeCache) createGatewayInformers(namespace string) gateway.SharedInfor
...
+			_, err := sharedInformers.Gateway().V1beta1().Gateways().Informer().AddEventHandler(c.registryRefreshHandler)
+			if err != nil {
+				log.Errorf("Error adding Handler to Informer Gateways: %s", err.Error())
+			}
 		}
 		if c.CheckIstioResource(kubernetes.K8sHTTPRoutes) {
 			lister.k8shttprouteLister = sharedInformers.Gateway().V1beta1().HTTPRoutes().Lister()
 			lister.cachesSynced = append(lister.cachesSynced, sharedInformers.Gateway().V1beta1().HTTPRoutes().Informer().HasSynced)
-			sharedInformers.Gateway().V1beta1().Gateways().Informer().AddEventHandler(c.registryRefreshHandler)
+			_, err := sharedInformers.Gateway().V1beta1().Gateways().Informer().AddEventHandler(c.registryRefreshHandler)
```
jmazzitelli, line 508: shouldn't this be HTTPRoutes?

Also, make sure the error messages are different (even if slightly different) so when log messages are being examined we know where they were logged (having identical error messages logged in multiple places sometimes makes debugging more difficult)

nrfox: Yes it should. Looks like this was also present in the orginal PR #6231 and I just ported these lines directly from there. I can update it though.

jmazzitelli (approved):
i approve for merge, if we submit a separate PR to fix that copy-n-paste error (fix in master, then cherry pick that fix).

Issue #6372, opened 2023-07-18
App health - Inconsistency in Overview page


hhovsepy:
Happens on Kiali master, there is a regression in showing Apps health in Overview page, it is different than in Apps list and details pages. A Failure app is not shown anywhere else besides Overview page.

![Screenshot from 2023-07-18 19-50-12](https://github.com/kiali/kiali/assets/604313/98922b1c-b0fc-496e-b622-b658f10ab1cd)
![Screenshot from 2023-07-18 19-50-37](https://github.com/kiali/kiali/assets/604313/38a808b6-9855-4acb-82e0-0233fd5b0a5b)

PR #6379, opened 2023-07-19
Skip loading requests health for apps and workloads without sidecar

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6372

The issue can be reproduced on multicluster env, for 'east' cluster's 'istio-system' namespace (easiest way).
It shows 1 Error on the 'istio-system' namespace card for 'east' cluster in Overview page for Apps and Workloads, for the workload 'istio-ingressgateway', but in fact when checking Apps list or details, all are healthy.
The failure is because there is some 'outbound' request with '403' for that workload 'istio-ingressgateway'.
The issue is in inconsistency when calculating health for App/Workloads and for Namespace in Overview page.
For individual or list of Apps/Workloads, those workloads which does not have sidecars, the health does not include Requests.
But in Overview page there was a bug, if at least one workload has a sidecar, then load Requests for all Apps/Workloads in that namespace.
And in 'istio-system' namespace there is one workload 'istio-eastwestgateway' which has a sidecar. This causes loading Requests for 'istio-ingressgateway' workload as well, and in fact I am not sure why there is that Request for this workload, but as it does not have a sidecar, the Requests should be ignored.
The fix is to introduce a map of App/Workload names and sidecars, so for each App or Workload while calculating Namespace health in Overview page, keep loading Requests only for those which has sidecars.

### DIFF
--- a/business/health.go
+++ b/business/health.go
@@ -153,6 +153,7 @@ func (in *HealthService) getNamespaceAppHealth(appEntities namespaceApps, criter
 
 	// Perf: do not bother fetching request rate if no workloads or no workload has sidecar
 	sidecarPresent := false
+	var appSidecars = make(map[string]bool)
 
 	// Prepare all data
 	for app, entities := range appEntities {
@@ -164,6 +165,7 @@ func (in *HealthService) getNamespaceAppHealth(appEntities namespaceApps, criter
 				for _, w := range entities.Workloads {
 					if w.IstioSidecar {
 						sidecarPresent = true
+						appSidecars[app] = true
 						break
 					}
 				}
@@ -178,7 +180,7 @@ func (in *HealthService) getNamespaceAppHealth(appEntities namespaceApps, criter
 			return allHealth, errors.NewServiceUnavailable(err.Error())
 		}
 		// Fill with collected request rates
-		fillAppRequestRates(allHealth, rates)
+		fillAppRequestRates(allHealth, rates, appSidecars)
 	}
 
 	return allHealth, nil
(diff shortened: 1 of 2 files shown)


leandroberetta (approved):
After a clean installation, I couldn't reproduce this, that 403 request it's missing here for sure, it's rare you have it, because I don't know where it comes from but this code looks good and makes sense to me.

hhovsepy:
w.IstioSidecar for istio-ingressgateway is false when I debug locally.

jshaughn:
> w.IstioSidecar for istio-ingressgateway is false when I debug locally.

I can believe that's correct, because it does not have a sidecar.  But, it IS a sidecar :-)

jshaughn (commented):
@hhovsepy I understand there is an inconsistency in the way the health is performed for the overview namespace and the list/detail.  My concern here is that I think the solution makes both the overview card and the app show green, when actually I think they should both show red.  Is that true or am I confused?  I'm not able to quite reproduce it as I don't have the MC setup.

So, it's possible this PR is correct as-is.  But my concern was that the handling for istio-ingressgateway (or istio-egressgateway) may not be correct.  Those gateways don't have a sidecar but are themselves proxies, and they do generate request telemetry.  So even though currently they will show `IstioSidecar: false`, I think we may need to update the `HasIstioSidecar()` functions to set to `true` for these istio gateways.

hhovsepy:
@jshaughn yes you are are right, now I have added a new method 'IsGateway', and used in health calculation for App/Workload Details and  in Overview pages. Just that check is used for health calculations only, not in general 'HasIstioSidecar' where it touches a lot of areas.

jshaughn (approved):
@hhovsepy Looks good, I like the solution to my previous feedback.

Issue #6380, opened 2023-07-20
Kiali pod fails to start because of "failed to list *v1beta1.Gateway: gateways.gateway.networking.k8s.io is forbidden"


FilipB:
Kiali: Version: v1.72.0-SNAPSHOT, Commit: f1b0bc8a1e070942229e70c5a544269d14a90778, Go: 1.20.5
Istio version: 1.18.1
OCP version: tried 4.13.4 and 4.12.23

Repro steps (available via jenkins job):
1. install-istio-via-istioctl.sh -iee true -m true -cn cluster-default -cp openshift -gae true
2. build the latest helm charts (make build-helm-charts)
3. install kiali operator via helm (details in jenkins job)
4. create kiali CR (details in jenkins job)

Result:
kiali pod is not starting with following error in a log:
```
W0720 08:22:14.742543 1 reflector.go:533] pkg/mod/k8s.io/client-go@v0.27.2/tools/cache/reflector.go:231: failed to list *v1beta1.HTTPRoute: httproutes.gateway.networking.k8s.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "httproutes" in API group "gateway.networking.k8s.io" at the cluster scope
E0720 08:22:14.742568 1 reflector.go:148] pkg/mod/k8s.io/client-go@v0.27.2/tools/cache/reflector.go:231: Failed to watch *v1beta1.HTTPRoute: failed to list *v1beta1.HTTPRoute: httproutes.gateway.networking.k8s.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "httproutes" in API group "gateway.networking.k8s.io" at the cluster scope
W0720 08:22:14.746429 1 reflector.go:533] pkg/mod/k8s.io/client-go@v0.27.2/tools/cache/reflector.go:231: failed to list *v1beta1.Gateway: gateways.gateway.networking.k8s.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "gateways" in API group "gateway.networking.k8s.io" at the cluster scope
[...]

PR #6384, opened 2023-07-20
support namespace scoped gateway informers

jmazzitelli:
fixes https://github.com/kiali/kiali/issues/6380


jmazzitelli:
I ran a quick test and I no longer get those errors mentioned in the issue.  This looks to have fixed the problem.

nrfox (approved):
LGTM

We probably need some basic e2e test for this specific combination as well. That can be a separate change.

hhovsepy (approved):
Tested on OCP cluster, Kiali installs and starts fine. K8s Gateway API objects are CRUD fine.

PR #6385, opened 2023-07-20
Pass kiali SA client to OS biz svc

### LINKED ISSUE #6381: openshift auth logout not behaving
Molecule test is failing since July 18, 2023. It was passing before that.

To see the failure, log into a OpenShift cluster and then run: `./hack/run-molecule-tests.sh -dorp podman -ce $(which oc) -ct openshift -at openshift-auth-test`

error:
```
"Status code was 502 and not [204]: HTTP Error 502: Bad Gateway", "redirected": false, "status": 502, "url": "https://kiali-istio-system.apps.ocp4.local/api/logout"}
```

It seems before we were returning status code 204, but now a 502 is resulting from the call to Kiali's `api/logout`.

nrfox:
The Openshift OAuth business service now relies on the kiali service account client but it wasn't being passed in by the business layer.

Fixes #6381


jmazzitelli (approved):
Check out this PR, run `make build build-ui cluster-push` and then run the molecule test:

```
./hack/run-molecule-tests.sh -dorp podman -ce $(which oc) -ct openshift -at openshift-auth-test -udi true

```

results in:

```
                     openshift-auth-test... success [4m 27s]
```

PR #6389, opened 2023-07-20
[CI] e2e tests - Fail on error

nrfox:
Replaces  `assert` with `require` in e2e tests to ensure that tests immediately fail and stop executing if unexpected input is encountered. This is necessary to avoid nil pointer access. For example
```
resp, err := request(...)
assert.Nil(err)
assert.NotNil(resp)
resp.Name
```
would result in a panic if there's an error and resp is nil because `assert` does not stop executing the test. Whereas using `require` would stop execution of the test making it more clear where/how the test failed.

Also replaces `Nil(err)` checks with `NoError(err)` because the latter will log the error instead of printing the error object which does not show you the error text.

Note this doesn't fix any CI flake but it should make it more visible what the problem is when flakes or e2e test failures happen in the future.

### DIFF
--- a/tests/integration/tests/applications_test.go
+++ b/tests/integration/tests/applications_test.go
@@ -4,64 +4,64 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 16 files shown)


jmazzitelli (approved):
+1

Let's see the CI tests go green and get this merged!

PR #6392, opened 2023-07-21
An approach to handling the "mutiple badge" issue for Nodes

### LINKED ISSUE #6370: [PFT graph] Solve problem with node badging
Our Cy graph supports multiple badges on Nodes (the purple badge icons showing things like VS, FI, TrafficSource, etc).  PFT seems to support only one optional badge, and currently even the one badge does not look very good:

image

image

We need to come up with a solution to support something roughly equivalent.

jshaughn:
Closes #6370 

Because PFT allows for only a single labelIcon (i.e badge), we will reserve that specifically to indicate Traffic Sources (or Gateways) because it helps users identify where traffic is entering the mesh.  The other badges we see in Cytoscape will be presented as node attachments.  BUT, since there are only a limited number of attachments (4), and we want to present badges in a consistent location, all VS features will be indicated by the VS icon, and the tooltip will indicate the specific VS features in play.

Screenshot from 2023-07-20 16-52-53

### DIFF
--- a/frontend/src/config/Icons.ts
+++ b/frontend/src/config/Icons.ts
@@ -1,56 +1,153 @@
 import deepFreeze from 'deep-freeze';
-
 import solidPinIcon from '../assets/img/solid-pin.png';
 import hollowPinIcon from '../assets/img/hollow-pin.png';
-import { BlueprintIcon, WrenchIcon, SecurityIcon } from '@patternfly/react-icons';
+import {
+  ArrowAltCircleRightIcon,
+  BanIcon,
+  BlueprintIcon,
+  BoltIcon,
+  ClockIcon,
+  CodeBranchIcon,
+  GlobeRouteIcon,
+  LockedIcon,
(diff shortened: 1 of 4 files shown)


leandroberetta (commented):
@jshaughn I tested and works. Looks good and I agree with the approach. Just minor comment, can we change the color to the violet that we used to use for these kinds of things?. Apart from mantaining the colour, will be highlighted better. I don't know if it will be patternfly compliance though in terms of styles...

leandroberetta (approved):
I will approve, the color of the badge can be discussed later and added in a follow up PR.

jshaughn:
> can we change the color to the violet...

@leandroberetta I agree that the current styling does not have much impact.  The problem is that the rendering of the labelIcon is "under the covers" and from what I can see there is no support for additional styling (other than a padding value).  So, maybe we can figure something out, but I'm +1 to pushing this to the future.

PR #6400, opened 2023-07-24
Fix merge issue

jshaughn:


nrfox (approved):
FE tests look like they're passing now

Issue #6401, opened 2023-07-24
Validate Kiali working with Ambient for Istio 1.19 dev


josunect:
ref. https://github.com/kiali/kiali/issues/5476 

Validate how Kiali works with Istio Ambient 1.19. 

- Download Istio 1.19.0-alpha
`curl -L https://istio.io/downloadIstio | ISTIO_VERSION=1.19.0-alpha.1 TARGET_ARCH=x86_64 sh -`

- Create kind cluster 
- Install Istio: 
`istio/install-istio-via-istioctl.sh -id Downloads/istio-1.19.0-alpha.1 -cp ambient`

- Use the hack script to install istio:
`istio/install-bookinfo-demo.sh -ai false -id /home/jcordoba/Downloads/istio-1.19.0-alpha.1 -c kubectl`

Tasks: 

- [x] Improve Istio Ambient global check https://github.com/kiali/kiali/pull/6406
- [x] No Services nodes in the graph https://github.com/kiali/kiali/issues/5978
- [x] The nodes in the graph don't show "Has missing sidecar" https://github.com/kiali/kiali/pull/6420 

![image](https://github.com/kiali/kiali/assets/49480155/9b2cd140-0365-4d2c-b9ea-5f8fa3d3a3b4)

VS the graph generated with telemetry generated by sidecars: 

![image](https://github.com/kiali/kiali/assets/49480155/575e7ff2-a20d-4688-88c9-dd459a93eab4)

josunect:
It is not detecting Ambient in this version: 

![image](https://github.com/kiali/kiali/assets/49480155/d3f57380-01c3-4d51-a3d0-14aaf76da33d)

josunect:
With pr and bookinfo example: 

![image](https://github.com/kiali/kiali/assets/49480155/96ec38eb-d7a0-4bdb-b194-af9716248148)

jmazzitelli:
@josunect can you approve this PR? This has the bookinfo demo script with the `-ai` that is needed for ambient testing. https://github.com/kiali/kiali/pull/6352

PR #6406, opened 2023-07-25
Update the way kiali identifies Ambient

josunect:
Fixes https://github.com/kiali/kiali/issues/6401

Create a more robust way to detect Istio Ambient


jmazzitelli (approved):
this looks correct to me - ztunnel installed in istio namespace implies that ambient is enabled.

PR #6408, opened 2023-07-25
[hack] tell k8s minikube what cni impl to use

jmazzitelli:
This is potentially needed to get Ambient to work on minikube. You need `kindnet` as the CNI option.
However, it still doesn't work for me. But this new option is probably going to be needed eventually and is nice to have.


josunect (approved):
I tested and in works as expected. 

I was trying to validate if the CNI argument was applied correctly but I've seen the same output as the default: 

```
./k8s-minikube.sh start -kcni kindnet

Configuring CNI (Container Networking Interface) ...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
```

But with calico the output is different: 

```
./k8s-minikube.sh start -kcni calico

Configuring Calico (Container Networking Interface) ...
    ▪ Using image gcr.io/k8s-minikube/storage-provisioner:v5
```

It looks like kindnet is the default CNI for minikube: https://minikube.sigs.k8s.io/docs/handbook/network_policy/

jmazzitelli:
> It looks like kindnet

That would explain why it still doesn't work for me :) I was told kindnet is required for ambient, but even with `-kcni kindnet` it still breaks (which makes sense if that is the default anyway) - this tells me there is something else that is wrong.

But I'll merge this anyway, since it could be useful.

PR #6411, opened 2023-07-25
Gherkin tags for single cluster Cypress tests

matejnesuta:
I added some tags for the current tests. From now on, we want to label every test related to a single cluster mode with the `@single-cluster` tag and every test related to Multicluster testing with the `@multicluster` tag. The reason for this is that we want to have more control over the test execution. For example, there is a high chance that we might need to do 2 test runs (one with @single-cluster tag and one with @multicluster) due to the Github CI limitations.

I also tried to fix the indentation of some files.

### DIFF
--- a/frontend/cypress/integration/featureFiles/app_details.feature
+++ b/frontend/cypress/integration/featureFiles/app_details.feature
@@ -12,31 +12,37 @@ Feature: Kiali App Details page
 
   @app-details-page
   @bookinfo-app
+  @single-cluster
   Scenario: See details for app.
     Then user sees details information for app
 
   @app-details-page
   @bookinfo-app
+  @single-cluster
   Scenario: See app minigraph for details app.
     Then user sees a minigraph
 
   @app-details-page
   @bookinfo-app
+  @single-cluster
   Scenario: See app Traffic information
     Then user sees inbound and outbound traffic information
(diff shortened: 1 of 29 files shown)


jshaughn (commented):
A couple of questions:I don't see anything labeled with `@multicluster` (actually, can we make this `@multi-cluster` for consistency?).   Is this a first step, to label everything with `@single-cluster`?  Also, can a test be labeled with both?

ScriptingShrimp:
> A couple of questions:I don't see anything labeled with `@multicluster` (actually, can we make this `@multi-cluster` for consistency?).

+1 for `@multi-cluster`

>  Is this a first step, to label everything with `@single-cluster`? Also, can a test be labeled with both?
@matejnesuta I would suggest to use `@multi-cluster` only. The `@single-cluster` tag is a bit redundant here.

The multicluster regression will be triggered [by the tag](https://github.com/badeball/cypress-cucumber-preprocessor/blob/master/docs/tags.md#running-a-subset-of-scenarios) coming form the package.json.

nrfox:
> Do we actually need @single-cluster? 

+1, can we avoid `@single-cluster` and just have `@multi-cluster`. The default with no tag should just be single-cluster.

PR #6412, opened 2023-07-25
[hack] ci kind molecule script to use newer loadbalancer

jmazzitelli:
follow the pattern in setup-kind-in-ci.sh


jshaughn (approved):
I KinD of like it.

hack/ci-kind-molecule-tests.sh
```diff
@@ -90,7 +90,7 @@ Options:
 
 -lb|--logs-branch <branch name>
     The logs branch to clone.
-    Default: master
+    Default: kind
```
nrfox, line 93: Should this still be `kind`?

hack/ci-kind-molecule-tests.sh
```diff
@@ -114,7 +114,7 @@ Options:
 
 -sd|--src-dir <directory>
     Where the git source repositories will be cloned.
-    Default: /tmp/KIALI-GIT
+    Default: /tmp/KIALI-GIT-KIND
```
nrfox, line 117: Same with this?

jmazzitelli: no. branch name and file name are irrelevant to each other

hack/ci-kind-molecule-tests.sh
jmazzitelli: yes, this is the cluster kind branch for the log repo.. See: https://github.com/jmazzitelli/kiali-molecule-test-logs/branches

You can change the branch name if you want to put logs in a different repo and you want a different branch.

jmazzitelli: Because this is running molecule tests on KinD, I put the logs in the "kind" branch. You will see I also have minikube and openshift branches for when I run nightly tests on those clusters

hack/ci-kind-molecule-tests.sh
jmazzitelli: Oh, I see what you are asking... because this is only running with KinD, I put this is a KIND specific directory so as not to conflict with the other CI hack script that uses KIALI-GIT -- I don't want the test logs stepping on each other so I set this default name to be different.

PR #6416, opened 2023-07-26
[hack] be able to rebuild cluster to start tests fresh.

jmazzitelli:
This is useful when running multiple tests and you want to start each with a fresh cluster (like when testing with OLM and then with helm).


jshaughn (approved):
Fine.

Issue #6307, opened 2023-06-26
Automate this scenario around Graph Display in Cypress 


matejnesuta:
I turned on operation nodes and turned off service nodes in app-graph:
![image](https://github.com/kiali/kiali/assets/2029470/742956ef-2cb9-4128-b81a-2962a0aa4a90)
I then switched to service graph and those options are the same, but grayed out to indicate they are unused for this graph type:
![image](https://github.com/kiali/kiali/assets/2029470/c8267f1c-8d7e-49c4-ba33-09c471b6f38b)

_Originally posted by @jmazzitelli in https://github.com/kiali/kiali/issues/6188#issuecomment-1562916608_

matejnesuta:
This could be most likely automated in Cypress.

PR #6419, opened 2023-07-26
Cypress automation for a Graph page test scenario

matejnesuta:
This is a Cypress automation of a test scenario (#6307) and it relates to certain Graph page controls being disabled when using the Service graph (#6188).


jshaughn (approved):
Looks OK to me.

PR #6420, opened 2023-07-26
Add Ambient logic in graphs with a graph ambient appender

josunect:
Ref. https://github.com/kiali/kiali/issues/6401 

When an application has the Ambient labels, we can still see in the graphs the "Missing Sidecar" icon. Also, in the summary panel there is the "No sidecar" message. 

image

The graph should not show these icons when the nodes are on an Ambient namespace.


jshaughn:
> Add Ambient logic in graphs with a graph ambient appender

I'm not exactly sure we need another appender for ambient, maybe, but before we do that let's think a little more on this.

> The graph should not show these icons when the nodes are on an Ambient namespace.

I agree that we need to update our handling here.   Maybe we can expand more on our previous introduction of "Out of Mesh".  That same terminology can be applied in multiple situations:

Ambient Mesh Namespace:
- when a workload opts out (I think if it does not have the Ambient annotation `ambient.istio.io/redirection`?)
  - question: what happens if the pod has a sidecar in an ambient namespace?

Classic Mesh Namespace:
- when there is no sidecar

I think we can keep the current icon but instead of it meaning "Missing Sidecar", it could be re-purposed to mean, "Out of Mesh".  We could update our implementation of this flag to be more comprehensive.  For example:

- Rename `HasMissingSidecar` to `IsOutOfMesh`
- Change the type from `boolean` to `string` with some values like [`nosidecar`, `noambient`]
  - Or, keep it as `boolean` and add an `IsAmbient` flag on every Node

The existing `sidecarsCheck` appender could be updated to perform different checks for an Ambient or Classic namespace.  We could even rename the appender.

josunect:
> > Add Ambient logic in graphs with a graph ambient appender
> 
> I'm not exactly sure we need another appender for ambient, maybe, but before we do that let's think a little more on this.
> 
> > The graph should not show these icons when the nodes are on an Ambient namespace.
> 
> I agree that we need to update our handling here. Maybe we can expand more on our previous introduction of "Out of Mesh". That same terminology can be applied in multiple situations:
> 
> Ambient Mesh Namespace:
> 
>     * when a workload opts out (I think if it does not have the Ambient annotation `ambient.istio.io/redirection`?)
>       
Yes
>       * question: what happens if the pod has a sidecar in an ambient namespace?

For what I understand, the sidecar takes preference over Ambient annotations to facilitate the migration from Sidecars to Ambient. We can display a message, though, or even a warning.

> 
> 
> Classic Mesh Namespace:
> 
>     * when there is no sidecar
> 
> 
> I think we can keep the current icon but instead of it meaning "Missing Sidecar", it could be re-purposed to mean, "Out of Mesh". We could update our implementation of this flag to be more comprehensive. For example:
> 
>     * Rename `HasMissingSidecar` to `IsOutOfMesh`
> 
>     * Change the type from `boolean` to `string` with some values like [`nosidecar`, `noambient`]
>       
>       * Or, keep it as `boolean` and add an `IsAmbient` flag on every Node
> 

I'm agree with keeping the current icon. But, the problem I was trying to solve in this issue is that we display the "Missing sidecar" in an Ambient namespace, which should not be a problem as the node is in the Ambient Mesh. 

I think we could keep it as boolean, because it could have both at the same time (And we could identity those situations and show a warning, to help detect issues in the mesh, for example). 

> 
> The existing `sidecarsCheck` appender could be updated to perform different checks for an Ambient or Classic namespace. We could even rename the appender.

I think is a good idea, but I would rename it to clarify. I can think in something like mesh_checker. What do you think?

jshaughn:
> I think is a good idea, but I would rename it to clarify. I can think in something like mesh_checker. What do you think?

I will start commenting in the code for clarity...

graph/telemetry/istio/appender/appender.go
```diff
@@ -34,14 +34,14 @@ func ParseAppenders(o graph.TelemetryOptions) (appenders []graph.Appender, final
 				requestedAppenders[IdleNodeAppenderName] = true
 			case IstioAppenderName:
 				requestedAppenders[IstioAppenderName] = true
+			case MeshCheckAppenderName:
```
jshaughn, line 37: If we change the appender name and don't add  backward-compatible test for "sidecarsCheck" we need to make sure this backend will only be called by compatible clients.  I'm not 100% sure if we could have an older OSSMC calling this backend, probably not, but if we have any doubt we could add the additional check.

frontend/src/components/CytoscapeGraph/graphs/GraphStyles.ts
```diff
@@ -224,7 +224,10 @@ export class GraphStyles {
     // Badges portion of label...
 
     let badges = '';
-    if (cyGlobal.showMissingSidecars && node.hasMissingSC) {
+    if (
+      (cyGlobal.showMissingSidecars && node.outOfMesh && !serverConfig.ambientEnabled) ||
+      (cyGlobal.showMissingSidecars && node.outOfMesh && serverConfig.ambientEnabled)
+    ) {
```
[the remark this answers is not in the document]
josunect: It makes sense. Updated.

frontend/src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -321,14 +322,14 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
...
-      hasMissingSC,
       hasRequestRouting,
       hasRequestTimeout,
       hasTCPTrafficShifting,
       hasTrafficShifting,
       hasVS,
       isDead,
-      isGateway
+      isGateway,
+      outOfMesh
```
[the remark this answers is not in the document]
josunect: Updated.

frontend/src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -369,12 +370,18 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
             <span style={{ paddingLeft: '4px' }}>Has Mirroring</span>
           </div>
         )}
-        {hasMissingSC && (
+        {outOfMesh && !serverConfig.ambientEnabled && (
```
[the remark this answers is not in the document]
josunect: Updated with isAmbient

frontend/src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -369,12 +370,18 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
...
           <div>
             <KialiIcon.MissingSidecar />
             <span style={{ paddingLeft: '4px' }}>Has Missing Sidecar</span>
           </div>
         )}
+        {outOfMesh && serverConfig.ambientEnabled && (
+          <div>
+            <KialiIcon.MissingSidecar />
```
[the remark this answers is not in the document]
josunect: Updated. Do you think it would make sense to also update all other MissingSidecar occurrences in the code?

frontend/src/pages/GraphPF/components/stylesComponentFactory.tsx
```diff
@@ -98,12 +99,21 @@ const handleDoubleTap = (doubleTapNode: GraphElement) => {
...
   }
 
-  if (dtNodeData.hasMissingSC) {
-    AlertUtils.add(
-      `A node with a missing sidecar provides no node-specific telemetry and can not provide a node detail graph.`,
-      undefined,
-      MessageType.WARNING
-    );
+  if (dtNodeData.outOfMesh) {
+    if (serverConfig.ambientEnabled) {
```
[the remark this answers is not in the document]
josunect: Updated

graph/telemetry/istio/appender/appender.go
josunect: It makes sense to just in case be backwards compatible. I have included the sidecarsCheck.

josunect:
> > I think is a good idea, but I would rename it to clarify. I can think in something like mesh_checker. What do you think?
> 
> I will start commenting in the code for clarity...

Thanks for the review, @jshaughn ! 

I still need to review the `isAmbient` logic (It is not yet implemented)

frontend/src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -114,7 +114,8 @@ export interface GraphNodeTapEvent {
   aggregateValue?: string;
   app: string;
   cluster?: string;
-  hasMissingSC: boolean;
+  isAmbient: boolean;
+  isOutOfMesh: boolean;
```
jshaughn, line 118: alphabetize :-)

frontend/src/pages/GraphPF/GraphPFElems.tsx
```diff
@@ -170,7 +170,7 @@ export const setNodeAttachments = (node: Node<NodeModel>, settings: GraphPFSetti
   const data = node.getData() as NodeData;
   const attachments = [] as React.ReactNode[];
 
-  if (settings.showMissingSidecars && data.hasMissingSC) {
+  if (settings.showMissingSidecars && data.isOutOfMesh) {
```
jshaughn, line 173: We should probably also change the setting name.

frontend/src/types/Graph.ts
```diff
@@ -571,12 +571,14 @@ export const NodeAttr = {
   httpInNoResponse: 'httpInNoResponse',
   httpOut: 'httpOut',
   id: 'id',
+  isAmbientNamespace: 'isAmbientNamespace',
```
jshaughn, line 574: Is there a reason why this isn't also `isAmbient`?

graph/telemetry/istio/appender/appender.go
```diff
@@ -34,14 +34,16 @@ func ParseAppenders(o graph.TelemetryOptions) (appenders []graph.Appender, final
...
```
jshaughn, line 37:
```suggestion
			case MeshCheckAppenderName:
			case SidecarsCheckAppenderName: // For backward compatibility
```

I know, it's not alphabetical :-)  But, I think it is better to combine the cases that are handled in the same way.  You can leave as-is if you disagree.

graph/telemetry/istio/appender/mesh_check.go
```diff
@@ -1,44 +1,46 @@
...
 
 import (
+	"context"
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/graph"
 )
 
-const SidecarsCheckAppenderName = "sidecarsCheck"
+const MeshCheckAppenderName = "meshCheck"
+const SidecarsCheckAppenderName = "sidecarCheck"
```
jshaughn, line 10:
```suggestion
const SidecarsCheckAppenderName = "sidecarsCheck"
```

Typo.  If we want back compat we need to keep the exact same name constant.

graph/telemetry/istio/appender/mesh_check.go
```diff
@@ -98,3 +109,15 @@ func (a *SidecarsCheckAppender) namespaceOK(namespace string, namespaceInfo *gra
...
+func (a *MeshCheckAppender) isAmbientNs(namespace string, cluster string, globalInfo *graph.AppenderGlobalInfo) bool {
+	ns, _ := globalInfo.Business.Namespace.GetClusterNamespace(context.TODO(), namespace, cluster)
+
+	for _, label := range ns.Labels {
+		if label == "istio.io/dataplane-mode" {
+			return true
+		}
+	}
+	return false
+}
```
jshaughn, line 123: This is fine but I think it could be better if `models.Namespace` included an `IsAmbient` flag, which was set one time, in a centralized location.  Then the consumers of `models.Namespace` don't need to check repeatedly, and don't need to know how to check if a namespace isAmbient.  What do you think?

jshaughn (commented):
Looking good.  A few more comments and suggestions.  Also, where does the Node actually get a value for`isAmbient`?  I'm not sure I see where this gets set.

frontend/src/pages/Graph/SummaryPanelNode.tsx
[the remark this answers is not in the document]
jshaughn: Yes!  `MissingSidecar` is now really just a reason that a Node is OutOfMesh.

frontend/src/components/CytoscapeGraph/CytoscapeGraph.tsx
josunect: Ops, I forgot this one :)

frontend/src/pages/GraphPF/GraphPFElems.tsx
josunect: That's right. I updated all the occurrences.

frontend/src/types/Graph.ts
josunect: I was thinking in a way to differentiate when a Node isAmbient when there is a check in the pod annotation (It includes the ztunnel redirection annotation) or because it is an Ambient namespace. I think as you said in the next comment, it would be better to add it to models.Namespace, when it is a Namespace check

graph/telemetry/istio/appender/appender.go
josunect: It is more clear like this

frontend/src/pages/Graph/GraphToolbar/GraphFind.tsx
```diff
@@ -1001,7 +1001,7 @@ export class GraphFindComponent extends React.Component<GraphFindProps, GraphFin
         };
       case 'sc':
       case 'sidecar':
-        return { target: 'node', selector: isNegation ? `[?${NodeAttr.hasMissingSC}]` : `[^${NodeAttr.hasMissingSC}]` };
+        return { target: 'node', selector: isNegation ? `[?${NodeAttr.isOutOfMesh}]` : `[^${NodeAttr.isOutOfMesh}]` };
```
jshaughn, line 1004: This needs to be updated in the graphhelpfind.tsx as well.  The `sidecar` and `sc` should be replaced with `outofmesh` and `om`, although we can keep `sidecar` and `sc` for back compat (undocumented), just in case someone is using them in a stored  expression.

graph/telemetry/istio/appender/mesh_check.go
josunect: That's a good idea. I have added the logic to models.Namespace

josunect:
> Looking good. A few more comments and suggestions. Also, where does the Node actually get a value for`isAmbient`? I'm not sure I see where this gets set.

Right, I have updated it in Mesh_check https://github.com/kiali/kiali/pull/6420/files#diff-df4d4895c0c4f6016f6f0766ccf817d45cc73b295d7ad3d75d5d331d75c4e7a2R95

frontend/src/pages/Graph/GraphToolbar/GraphFind.tsx
josunect: Thanks for pointing at this! I didn't realized about it.

josunect:
> > Just a question, in this version, will we still be keeping the global check to see if we are in Ambient profile?
> 
> To summarize our off-issue discussion... I think it will be several releases before Ambient is ever enabled by default. Until then the user will need to set the profile when installing (or maybe re-starting) Istio. So the global check is still very useful and the "Ambient" badge on the control plane card is also useful.
> 
> Users will still also need to opt-in for ambient on a namespace, via a label. And so it's important to show that at the namespace level as well, in Overview. And finally a particular workload can opt-out, right, so that is why we need the more granular, `isAmbient` check at the node level.
> 
> One thing we could check for is for an Ambient namespace label but when the ambient profile is not enabled on the control plane. Another would be to show `HasSidecar` for an Ambient workload.

It would be Ok to implement those checks in additional issues?

josunect:
> Looking good. A few more comments and suggestions. Also, where does the Node actually get a value for`isAmbient`? I'm not sure I see where this gets set.

Just a note about using global serverconfig here: https://github.com/kiali/kiali/pull/6420/files#diff-d68058b42d3440f3154832ff89569e4695e0145620440fd6038293e18e4527e4R373

This is to whether we should show the usual message "Has missing sidecards" and keep the same message VS "Out of Mesh" when we are in an Ambient Mesh. Does it makes sense?

frontend/src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -369,12 +370,18 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
             <span style={{ paddingLeft: '4px' }}>Has Mirroring</span>
           </div>
         )}
-        {hasMissingSC && (
+        {isOutOfMesh && !serverConfig.ambientEnabled && (
```
jshaughn, line 373: I don't think you need the `serverConfig.ambientEnabled`.  Instead, I think you can use the `nodeData.isAmbient` setting which you already added to GraphNodeData.

josunect: But in that case, we can just check `isOutOfMesh`, because `isOutOfMesh` checks that the node meets both requirements -`!isAmbient` and !sidecars -, so in this case this comparison is not needed. 

The use of the global ambientEnabled is whether we change the message for nodes out of the mesh, from "Missing sidecars" to "Out of Mesh"

jshaughn: > The use of the global ambientEnabled is whether we change the message for nodes out of the mesh, from "Missing sidecars" to "Out of Mesh"

The global setting is for Istio overall, all namespaces, right?  But the node's `isAmbient` is specific to the node, so why not use `node.isAmbient` to make this decision.  Maybe I am confused :-)

josunect: Yes, that's right. 

I can put an example, 

image
 
This namespace doesn't have any annotation. Is this the message that we want to show even when we start the Mesh with Ambient profile? I think if we are in Ambient profile we should show "Out of Mesh". But if we check node.isAmbient it would be false in this case, and the message would be "Has missing sidecar". 

Checking globally if Ambient is enabled, when the node is out of mesh (Because the node has no sidecars or Ambient annotations) we can show the message "Out of Mesh". 

But maybe, this is not the desired behavior?

jshaughn: Installing with the Ambient profile means that namespaces in the mesh may be, but are not required to be, enabled for Ambient, right?   In the example above, the `local-path-storage` namespace is not annotated for Ambient.  The `local-path-provisioner` I think would have `node.isOutOfMesh: true` and `node.isAmbient: false`.  And so I think it is correct to show `Has Missing Sidecar` for the OutOfMesh situation.  If `local-path-storage` were annotated for Ambient,  then I think would have `node.isOutOfMesh: true` and `node.isAmbient: true`.  And so the message would be something like, "Out of Ambient Mesh".  Maybe this is the confusion.  I was expecting that `node.isAmbient` would be true for all nodes in an Ambient namespace.  But perhaps you expect, that, if the workload somehow opted out of the ambient mesh, it would be `node.isAmbient: false`.   That may be a fair assumption, but in that case I don't know how `isAmbient` is helpful, it is just duplicating `isOutOfMesh`.   In the way I am thinking `isAmbient` is just a convenience, because you could also remove that attribute and check the same thing with a utility function like, `isAmbient(node.namespace)`.

josunect: Ah so, I think the confusion was about the expected behavior. In the case of `local-path-store` namespace if it is not annotated for Ambient, but does not have the injection enabled annotation, I wouldn't expect to be warned about sidecars, because you have different elements to include the node in the mesh, so why just the sidecars are a warning? (In case you are in Ambient profile)

It is how it works at the moment in other pages: 

image

image

In case you want to show the `missing sidecars` message, we could use `node.isAmbient` in the way you have mentioned, then.

jshaughn: > I wouldn't expect to be warned about sidecars, because you have different elements to include the node in the mesh, so why just the sidecars are a warning? (In case you are in Ambient profile)

I think I am just making this more confusing.  I think we both want the same thing.  My only question was whether it was needed to use the global ambient profile setting in this logic.  If you think it is necessary then it is fine for me.  I do wonder whether the new `node.isAmbient` is needed.  Maybe it could be removed, and just have `node.isOutOfMesh`.  If we need to check whether the node is in an ambient namespace then we could have a utility function like `isAmbient(node.namespace)`.

josunect: Yes, the node isAmbient property was used in some other place but I have changed to use the global ambient profile to keep consistency with the rest of the application Ambient messages.

frontend/src/pages/Graph/GraphPage.tsx
```diff
@@ -632,7 +632,7 @@ class GraphPageComponent extends React.Component<GraphPageProps, GraphPageState>
       return;
     }
 
-    if (event.hasMissingSC) {
+    if (event.isOutOfMesh && !event.isAmbient) {
       AlertUtils.add(
         `A node with a missing sidecar provides no node-specific telemetry and can not provide a node detail graph.`,
         undefined,
```
jshaughn, line 638: Do we need the `isAmbient` check here?  If the node is outOfMesh then wouldn't this apply to ambient or non-ambient?

jshaughn (commented):
A couple of more things we may be able to remove for clarity.

frontend/src/components/CytoscapeGraph/CytoscapeGraph.tsx
```diff
@@ -114,9 +114,10 @@ export interface GraphNodeTapEvent {
   aggregateValue?: string;
   app: string;
   cluster?: string;
-  hasMissingSC: boolean;
+  isAmbient: boolean;
```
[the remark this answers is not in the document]
josunect: no, that is right.

frontend/src/pages/Graph/GraphPage.tsx
josunect: Yes, that is right. Removed isAmbient check

josunect:
> A couple of more things we may be able to remove for clarity.

Do you think it would be better to move from the checker, also? 

https://github.com/josunect/kiali/blob/issue6401_graph/graph/telemetry/istio/appender/mesh_check.go#L96C1-L96C1

It is not used right now., not sure if it could be useful.

graph/meta.go
```diff
@@ -23,13 +23,13 @@ const (
...
 	HasHealthConfig       MetadataKey = "hasHealthConfig"
 	HasMirroring          MetadataKey = "hasMirroring"
-	HasMissingSC          MetadataKey = "hasMissingSC"
 	HasTCPTrafficShifting MetadataKey = "hasTCPTrafficShifting"
 	HasTrafficShifting    MetadataKey = "hasTrafficShifting"
 	HasRequestRouting     MetadataKey = "hasRequestRouting"
 	HasRequestTimeout     MetadataKey = "hasRequestTimeout"
 	HasVS                 MetadataKey = "hasVS"
 	HasWorkloadEntry      MetadataKey = "hasWorkloadEntry"
+	IsAmbient             MetadataKey = "isAmbient"
```
jshaughn, line 32: If we remove from the Node then I guess we would not need this, either.

jshaughn (commented):
Looking good, now just wondering if we need the node-level `isAmbient` on the server-side, or whether just tracking on the Namespace is sufficient.

frontend/src/pages/Graph/GraphToolbar/GraphFind.tsx
```diff
@@ -1001,7 +1001,10 @@ export class GraphFindComponent extends React.Component<GraphFindProps, GraphFin
...
```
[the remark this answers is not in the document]
josunect: I think they mean the opposite, right? sidecar nodes are highlighted when out of the Mesh are not, that is why I put them in different cases.

graph/config/cytoscape/cytoscape.go
```diff
@@ -105,19 +105,20 @@ type NodeData struct {
...
 	HasMirroring          bool                `json:"hasMirroring,omitempty"`          // true (has mirroring) | false
-	HasMissingSC          bool                `json:"hasMissingSC,omitempty"`          // true (has missing sidecar) | false
 	HasRequestRouting     bool                `json:"hasRequestRouting,omitempty"`     // true (vs has request routing) | false
 	HasRequestTimeout     bool                `json:"hasRequestTimeout,omitempty"`     // true (vs has request timeout) | false
 	HasTCPTrafficShifting bool                `json:"hasTCPTrafficShifting,omitempty"` // true (vs has tcp traffic shifting) | false
 	HasTrafficShifting    bool                `json:"hasTrafficShifting,omitempty"`    // true (vs has traffic shifting) | false
 	HasVS                 *VSInfo             `json:"hasVS,omitempty"`                 // it can be empty if there is a VS without hostnames
 	HasWorkloadEntry      []graph.WEInfo      `json:"hasWorkloadEntry,omitempty"`      // static workload entry information | empty if there are no workload entries
+	IsAmbient             bool                `json:"isAmbient,omitempty"`             // If the node is in Ambient namespace
```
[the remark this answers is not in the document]
josunect: yes. I think it could be useful in future checks, but at the moment it is not used.

graph/meta.go
josunect: right!

frontend/src/pages/Graph/GraphToolbar/GraphFind.tsx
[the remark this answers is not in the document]
jshaughn: Yes, you are right!

Issue #5476, opened 2022-09-16
[ambient] initial istio Ambient support


jshaughn:
This is to start tracking notes, observations, behavior of Kiali with Istio Ambient.  Any identified sub-tasks can be added below:

Some useful things:
- [Introducing Ambient blog](https://istio.io/latest/blog/2022/introducing-ambient-mesh/)
- [Ambient limitations/readme](https://github.com/istio/istio/blob/experimental-ambient/README.md#limitations)
- [#Ambient channel on Istio Slack](https://istio.slack.com/archives/C041EQL1XMY)
- [Ambient weekly meeting contributor notes](https://docs.google.com/document/d/1SMlwliEnthgq7r2PjpLl1kCq3t8rAMbgu6r_lDAXJ0w/edit#heading=h.o8pz6aqnzzgk)

- Ambient Alpha (Istio 1.18): 
  - [x]  #5504
  - [x] Install bookinfo via our hackscript and Sanity check Kiali pages (set AUTO_INJECTION to false)
  - [x] Install testing demos and #5507 
  - [x] Verify Kiali in Istio 1.18 dev #5910 
  - [x] Verify Kiali using Istio 1.18 Ambient alpha https://github.com/kiali/kiali/issues/6241
  - [x] #5581 
  - [x] Alpha support #5535

- Ambient Alpha with some improvements (Istio 1.19 - 6th september):
  - [x] Beta support  https://github.com/kiali/kiali/issues/6401
  - [x] https://github.com/kiali/kiali/issues/6423
  - [x] #5539 
  - [x] https://github.com/kiali/kiali/pull/6352
  - [x] https://github.com/kiali/kiali/issues/6522
  - [x] https://github.com/kiali/kiali/issues/6579
 
- Istio 1.20: 
  - [x] https://github.com/kiali/kiali/issues/6757
  - [x] #5978  

- Istio 1.21 (Kiali 1.79): 
  - [x] Validate Kiali and Istio Ambient 1.21 dev https://github.com/kiali/kiali/issues/6981
  - [x] Include Waypoint proxy as a default: https://github.com/kiali/kiali/issues/6878
  - [x] Hide waypoint proxies from graph: https://github.com/kiali/kiali/pull/7015 
  - [x] Kiali and Istio Ambient in OpenShift https://github.com/kiali/kiali/issues/7159
  - [x] https://github.com/kiali/kiali/issues/7259

- Istio 1.22 - Ambient L4 beta (Kiali 1.84?): 
[...]

PR #6424, opened 2023-07-28
Include waypoint creation in bookinfo demo hack script

josunect:
Include waypoint option to create a waypoint proxy in bookinfo demo hack script. Just in case ambient is enabled

Ref. https://github.com/kiali/kiali/issues/5476

### DIFF
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -369,16 +369,16 @@ $CLIENT_EXE get pods -n ${NAMESPACE}
 if [ "${AMBIENT_ENABLED}" == "true" ]; then
   echo "Sidecar injection was not performed. Ambient support will be enabled."
   ${CLIENT_EXE} label namespace ${NAMESPACE} istio.io/dataplane-mode=ambient
-   # It could also be applied to service account
-   if [ "${WAYPOINT}" == "true" ]; then
-     # Verify Gateway API
-     echo "Verifying that Gateway API is installed; if it is not then it will be installed now."
-     $CLIENT_EXE get crd gateways.gateway.networking.k8s.io &> /dev/null || \
-       { $CLIENT_EXE kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2" | $CLIENT_EXE apply -f -; }
-     # Create Waypoint proxy
-     echo "Create Waypoint proxy"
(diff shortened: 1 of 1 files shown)


hack/istio/install-bookinfo-demo.sh
```diff
@@ -359,6 +369,16 @@ $CLIENT_EXE get pods -n ${NAMESPACE}
...
+    # It could also be applied to service account
+    if [ "${WAYPOINT}" == "true" ] && [ "${AMBIENT_ENABLED}" == "true" ]; then
+      # Verify Gateway API
+      echo "Verify if Istio API is installed"
+      $CLIENT_EXE get crd gateways.gateway.networking.k8s.io &> /dev/null || \
+        { $CLIENT_EXE kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2" | $CLIENT_EXE apply -f -; }
+      # Create Waypoint proxy
+      echo "Create Waypoint proxy"
+      ${ISTIOCTL} x waypoint apply -n ${NAMESPACE}
+    fi
```
jmazzitelli, line 381: I'll test this shortly, but...

1. You are in an if-stmt that already determined ambient is enabled - no need to check again.
2. Indentation is off by 1.
3. Regarding the message `"Verify if Istio API is installed"` - the code looks like it is looking for the Gateway API and, if it is not, it will install it. So I suggest changing the message to be more clear.

```suggestion
   # It could also be applied to service account
   if [ "${WAYPOINT}" == "true" ]; then
     # Verify Gateway API
     echo "Verifying that Gateway API is installed; if it is not then it will be installed now."
     $CLIENT_EXE get crd gateways.gateway.networking.k8s.io &> /dev/null || \
       { $CLIENT_EXE kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v0.6.2" | $CLIENT_EXE apply -f -; }
     # Create Waypoint proxy
     echo "Create Waypoint proxy"
     ${ISTIOCTL} x waypoint apply -n ${NAMESPACE}
   fi
```

jmazzitelli (changes requested):

hack/istio/install-bookinfo-demo.sh
josunect: Thanks for the review, @jmazzitelli ! 
I've make a commit regarding indentation - I guess it looks right now?

jmazzitelli (approved):
Tested... This is good.

PR #6425, opened 2023-07-28
Remove extra parenthesis in PF Graph

josunect:
There was an extra parenthesis in the PF Graph, which was visible in the left middle of the graph: 

image

To reproduce, we should set impl: "pf"


jshaughn (approved):
Good find!  I just noticed that yesterday and was wondering why it was there :-)

PR #6426, opened 2023-07-28
Restrict Bootstrap CSS only to Slider scope

### LINKED ISSUE #6417: Slider CSS affects Openshift console look and feel (OSSMC plugin)
To implement Kiali slider component we are using library `bootstrap-slider-without-jquery` that uses Bootstrap CSS (Slider/styles/default.css). When Bootstrap CSS is applied in OSSMC plugin gets in conflict with Openshift console (especially in dark mode). 

Service view in Openshift console before OSSMC is loaded:
image

Service view in Openshift console after OSSMC is loaded:
image

According to Openshift plugin guidelines (https://docs.openshift.com/container-platform/4.13/web_console/dynamic-plugin/overview-dynamic-plugin.html):

```
Avoid using other CSS libraries such as Bootstrap or Tailwind. 
They can conflict with PatternFly and will not match the console look and feel.
```

A solution would be to isolate Bootstrap CSS to affect only OSSMC plugin and not the whole Openshift console.

ferhoyos:
** Describe the change **

Currently all Bootstrap CSS is imported when slider component is loaded. Since Bootstrap CSS contains global CSS classnames, it affects whole Kiali application. In this PR I have extracted from Bootstrap CSS only classnames really used in Kiali application, simplifying styles applied and reducing any conflict between Patternfly and Bootstrap.

Bootstrap CSS classnames can be split in two groups:
- Global: CSS classnames that apply to whole Kiali application (moved to GlobalStyle). Some examples are `.table` and `.panel`classnames.
- Slider: CSS classnames that apply only to Slider component (moved to slider CSS module). Here we find `.slider` and `.tooltip` classnames.

Some components style have been modified a bit since Bootstrap CSS was affecting them in some degree (e.g., AboutUIModal, MetricSettings and SessionTimeout).

Finally styling section of README file has been modified to show new style structure (Typestyle + CSS module instead of Typestyle + CSS files).

** Issue reference **

Fixes #6417 

** PR Testing **

Since this issue only affects to OSSMC, look and feel of Kiali application should be the same as without applying the PR. If you copy Kiali code into OSSMC repository you will notice that Openshift console is not affected anymore by Bootstrap CSS applied in Kiali (especially in dark mode)


ScriptingShrimp:
@ferhoyos would it make sense to write a test scenario for that slider? I think we would be sure nothing breaks until we remove bootstrap specific CSS.

frontend/src/components/IstioWizards/Slider/slider.module.scss
```diff
@@ -0,0 +1,376 @@
+.style {
+  :global {
+    /* Extracted from Bootstrap CSS. Don't modify these styles */
```
jshaughn, line 3: Should we indicate which version this was taken from?

frontend/src/styles/GlobalStyle.ts
```diff
@@ -116,6 +121,37 @@ export const globalStyle = kialiStyle({
       overflow: 'visible !important'
     },
 
+    // Table styles from bootstrap
```
jshaughn, line 124: Again, if this is a copy do we need to know the bootstrap version?

frontend/.gitignore
```diff
@@ -26,19 +26,4 @@ npm-debug.log*
...
-src/img/*
-!src/img/.gitkeep
-src/fonts/*
-!src/fonts/.gitkeep
-src/sass/*
-!src/sass/index.sass
-
-# Commit Slider css
-!src/components/IstioWizards/Slider/styles/default.css
+package-lock.json
```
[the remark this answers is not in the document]
ScriptingShrimp: Why `package-lock.json` is ignored? [Not a good idea.](https://dev.to/adamklein/package-lock-json-in-git-or-not-50l5)

jshaughn (commented):
@ferhoyos , you can teach a CSS mater class.  Tested various places that looked most likley to be affected by the changes, everything looks good to me.  A couple of small comments about adding a version where CSS is copied from bootstrap.  But overall looks good to me.

frontend/.gitignore
ScriptingShrimp: Why `package-lock.json` is ignored? [Not a good idea.](https://dev.to/adamklein/package-lock-json-in-git-or-not-50l5)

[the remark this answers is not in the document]
jshaughn: @ScriptingShrimp , note that `package-lock.json` is not really used by us, and I'm not sure the entry is even necessary.  It may be a legacy entry in the .gitignore.  We use `yarn` and it's associated `yarn.lock`.  We do keep `yarn.lock` in git.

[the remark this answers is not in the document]
ScriptingShrimp: Cool! thanks for clarifying!

ScriptingShrimp: answered here https://github.com/kiali/kiali/pull/6426#discussion_r1279663323

ScriptingShrimp:
> @ScriptingShrimp since this PR is just a refactorization, no additional test scenario is needed in Kiali. 

I was thinking about having regression on that slider - once we remove (refacor) bootstrap CSS the regression will break. Its up to you to decide. 

> Maybe next step is to create an issue in OSSMC repository to create specific cypress tests for OSSMC that checks Kiali plugin works successfully and its load does not break Openshift console look and feel. Do you think it is feasible to do this with Github actions?

Honestly no idea how to deploy OSSMC in upstream / GH actions. We can test that in private pipelines = nightly builds are running on OCP. 

OCP console have its own regression. If you can take a note somewhere that we wan to deploy it, load the plugin and run OCP console regression suite to be sure nothing is broken. This one time check could reassure us we did not broke anything by our CSS.

IIRC for OCP console plugin we need to create PR in their repo to ensure _"plugin works successfully and its load"_ed OK

frontend/src/components/IstioWizards/Slider/slider.module.scss
ferhoyos: According to https://github.com/kiali/kiali-ui/pull/1673, these styles come from PF 3 where Bootstrap slider (v 9.9.0) and Bootstrap (v 3.4.1) libraries were used. I have compared style values with the CSS files that comes with the libraries and the values are a bit different (probably customized by PF3). I will update the comment.

ferhoyos:
> I was thinking about having regression on that slider - once we remove (refacor) bootstrap CSS the regression will break. Its up to you to decide.

Not necessarily this PR will break any regression test because I have only cleaned CSS and restrict the scope to the slider component so that it does not affect the look and feel of the rest of the application.

If you mean to perform a manual regression test, reviewers can check that sliders present in the Kiali application have exactly the same look and feel (e.g., Replay feature within graph, Traffic routing within Service detail).

If you mean to create unit tests of this components or more cypress tests about slider, I would create a new issue (I think it is not in the scope of this PR)

jshaughn (approved):
@ferhoyos Thanks, Looks good!

Issue #6397, opened 2023-07-24
K8s Gateway API - Appenders in Graph


hhovsepy:
Add detection of Graph appenders for traffic-routing on K8s Gateway and HTTPRoute objects.

jshaughn:
On istio.io some features now show how to configure it using Istio's Virtual Service AND also K8s Gateway API and HTTPRoute objects.  The expectation would be for more and more people to opt for the more standard Gateway API approach.  Kiali graph badging needs to take this into account.

I'm not exactly sure how we want to handle things.  My first impression is that we can keep the same set of visual badges but that the badge may represent different implementations, and be able to make it clear which are being used.  In that way users get a consistent visual experience as then introduce or migrate from VS to K8s GA.

jshaughn:
@hhovsepy Right, this is a good example.   My concern here was that we would not detect request routing implemented via HTTPRoute.   The basic idea is that when we see an [Istio feature that could be implemented via VS or GA](https://istio.io/latest/docs/tasks/traffic-management/request-routing/#route-to-version-1), that we can detect it either way.  For example, we can still show the same Request Routing badge on the graph, and the side panel should also still show RR information, regardless of whether it is configured by VS or GA.

PR #6427, opened 2023-07-28
Show K8s Gateway API info on Graph side panel.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6397

In a case of K8s Gateway API objects, such as HTTPRoute and Gateway, the system auto generates Gateway and VirtualService (this is how K8s Gateways are working, under them regular objects are generated and working). So those VS and Gateways are loaded and shown in Graph side panel.

The change of this PR is to show user that the traffic configuration is coming from K8s Gateway API.

In case od traffic routing:
Screenshot from 2023-07-28 17-53-28

In case of Gateway itself:
Screenshot from 2023-07-28 17-53-11


hhovsepy:
@jshaughn wdyt about this kind of solution diferentiating K8s Gateway objects?

frontend/src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -420,7 +421,7 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
           <>
             <div>
               <KialiIcon.Gateway />
-              <span style={{ paddingLeft: '4px' }}>Is Gateway</span>
+              <span style={{ paddingLeft: '4px' }}>Is Gateway{isAutogenerated ? ' (K8s API)' : ''}</span>
```
jshaughn, line 424:
```suggestion
              <span style={{ paddingLeft: '4px' }}>Is Gateway{isAutogenerated ? ' (K8s GW API)' : ''}</span>
```

frontend/src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -433,6 +434,12 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
             </span>
           </div>
         )}
+        {!isGateway && isAutogenerated && (
+          <div>
+            <KialiIcon.Gateway />
+            <span style={{ paddingLeft: '4px' }}>K8s Gateway API</span>
+          </div>
+        )}
```
jshaughn, line 442: I think I would prefer it if this info is appended to the right.  So, similar to what is done for `Is Gateway (K8s GW API)` we could have `Has Request Routing (K8s GW API)`.    This would be more consistent and improve locality of the information.

graph/config/cytoscape/cytoscape.go
```diff
@@ -112,6 +112,7 @@ type NodeData struct {
 	HasTrafficShifting    bool                `json:"hasTrafficShifting,omitempty"`    // true (vs has traffic shifting) | false
 	HasVS                 *VSInfo             `json:"hasVS,omitempty"`                 // it can be empty if there is a VS without hostnames
 	HasWorkloadEntry      []graph.WEInfo      `json:"hasWorkloadEntry,omitempty"`      // static workload entry information | empty if there are no workload entries
+	IsAutogenerated       bool                `json:"isAutogenerated,omitempty"`       // true (object is auto-generated from K8s API Gateway) | false
```
jshaughn, line 115:
```suggestion
	IsK8sGatewayAPI       bool                `json:"isK8sGatewayAPI,omitempty"`       // true (object is auto-generated from K8s API Gateway) | false
```

I think `IsAutogenerated` is too much of an implementation detail.  Currently I guess Istio must be autogenerating K8s GW API definitions into the existing Virtual Service definition and assigning a special name.  I'm not sure this will always be true and I think it would just be more clear with a name change.

A name change would require a little bit of re-alphabetizing...

jshaughn (changes requested):
@hhovsepy I think this is good extra information to add, until some day in the far future when K8s GW API will be the only accepted config.  A couple of suggested changes...

frontend/src/pages/Graph/SummaryPanelNode.tsx
hhovsepy: This was my initial implementation, but I thought, that from UX perspective too long sentence could be not nice, but I am fine to change it.

hhovsepy: Then we need to have it on all, even not K8s GW supported routing types.

hhovsepy:
Screenshot from 2023-08-01 08-54-44
Screenshot from 2023-08-01 08-54-21

PR #6428, opened 2023-07-28
Istio Status test ensuring there is no warning or danger status upon a fresh install

matejnesuta:
There is an Istio status in the header of the Kiali page. This status has various states (success, info, warning and danger). This scenario tests that there should be no warning or error status upon new Kiali installation. The reason for adding this test is that there was a faulty Prometheus deployment, but all of the Cypress test passed and the failure was only visible either through Kiali status, or the Prometheus UI. We also think Kiali should not have any warnings or errors upon new installation. 

I've also added some data tests in the React code.

### DIFF
--- a/frontend/cypress/integration/common/kiali_alert.ts
+++ b/frontend/cypress/integration/common/kiali_alert.ts
@@ -1,0 +1,7 @@
+import { Then } from '@badeball/cypress-cucumber-preprocessor';
+import { ensureKialiFinishedLoading } from './transition';
+
+Then(`user should see no Istio Components Status`, () => {
+  ensureKialiFinishedLoading(); 
+  cy.get('#istio-status-danger').should('not.be.visible');
+});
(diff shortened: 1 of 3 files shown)


jshaughn:
Hi @matejnesuta , I think you hit a flake in the integration tests that is already fixed.  Can you please rebase this PR to see if it goes away?

matejnesuta:
oops, one wrong git command and the whole PR is ruined...

jshaughn:
> oops, one wrong git command and the whole PR is ruined...

I hate when that happens :-)  Anyway, I'll take a look when the PR/CI issues are resolved...

hhovsepy:
Thank @matejnesuta for the initiative, lgtm in general, only question, is the warning status icon acceptable?

matejnesuta:
The force push is not updating for some reason. I'll close this one and open a new one

PR #6433, opened 2023-08-01
Enable Istio configs for services in Ambient namespace

### LINKED ISSUE #6423: Not possible to create Istio configs for services in Ambient Namespace
In a Service detail page of a service that is part of Ambient Mesh, it is not possible to create service configurations because there are no sidecars: 

image

I would make sense to make them available for Ambient also.

josunect:
Fixes #6423 

image

### DIFF
--- a/frontend/src/components/IstioWizards/ServiceWizardDropdown.tsx
+++ b/frontend/src/components/IstioWizards/ServiceWizardDropdown.tsx
@@ -87,16 +87,21 @@ class ServiceWizardDropdownComponent extends React.Component<Props, State> {
   private appLabelName = serverConfig.istioLabels.appLabelName;
   private versionLabelName = serverConfig.istioLabels.versionLabelName;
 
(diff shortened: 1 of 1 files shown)


josunect:
Steps to verify: 

- Create kind cluster: 

```
kind create cluster --config=- <<EOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: ambient
nodes:
- role: control-plane
- role: worker
- role: worker
EOF
```
- Install Istio with profile Ambient and bookinfo with auto-injection false 

```
istio/install-istio-via-istioctl.sh -c kubectl -cp ambient 
istio/install-bookinfo-demo.sh -c kubectl -ai false -tg
```

- Verify that bookinfo and the control plane have the Ambient badge 
- Go to a Service in bookinfo namespace, and click in Actions. Button should be enabled.

hhovsepy (approved):
Verified, tested locally:
Screenshot from 2023-08-02 16-08-45

PR #6434, opened 2023-08-01
Enable smoke testing in Cypress

matejnesuta:
** Describe the change **

_explanation of what it does_

** Issue reference **

### DIFF
--- a/frontend/cypress/integration/featureFiles/app_details.feature
+++ b/frontend/cypress/integration/featureFiles/app_details.feature
@@ -10,13 +10,11 @@ Feature: Kiali App Details page
     Given user is at administrator perspective
     And user is at the details page for the "app" "bookinfo/details"
 
-  @smoke
   @app-details-page
   @bookinfo-app
   Scenario: See details for app.
     Then user sees details information for app
 
-  @smoke
   @app-details-page
   @bookinfo-app
   Scenario: See app minigraph for details app.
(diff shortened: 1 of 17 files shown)


ScriptingShrimp:
Taking look into changed files, I see some scenarios tagged `@smoke` and also with `@bookinfo-app` or `@error-rates-app`. The smoke testing principle is [defined here](https://www.techtarget.com/searchsoftwarequality/definition/smoke-testing) - so based on that I would recommend to have smoke testing scenarios without other dependencies. Speculating a little bit further, the optimal workflow might be something like:

1. deploy istio 
2. smoke test istio
3. deploy kiali 
4. smoke test kiali
5. if all above passes, execute other layers of tests (functional, nonfunctional etc..)

What do you think?

matejnesuta:
I removed tags with dependencies to demo applications, which shrinked the possible candidates to roughly 10 scenarios. I wonder if the number is not too low. 

>deploy istio
>smoke test istio

I am not sure, what do you mean by this.

ScriptingShrimp:
> I am not sure, what do you mean by this.

Just highlighting importance of smoke testing. For an example in our private pipelines this could serve as gating, we don't want to test Kiali on broken setup of Istio. I like [this definition](https://www.javatpoint.com/smoke-testing)

ScriptingShrimp:
I already annotated some smokes in https://github.com/kiali/kiali/pull/6493, so if its get merged, you can rebase and add some more

matejnesuta:
Closed in favor of #6684.

PR #6435, opened 2023-08-01
Istio Status test ensuring there is no warning or danger status upon a fresh install 

matejnesuta:
There is an Istio status in the header of the Kiali page. This status has various states (success, info, warning and danger). This scenario tests that there should be no warning or error status upon new Kiali installation. The reason for adding this test is that there was a faulty Prometheus deployment, but all of the Cypress test passed and the failure was only visible either through Kiali status, or the Prometheus UI. We also think Kiali should not have any warnings or errors upon new installation.

I've also added some data tests in the React code.

Reopened after some issues with #6428 

To test this:
- deploy Kiali
- run the test and it should pass by default
- scale down replicas of something like Prometheus to 0 to invoke the Istio Status icon ` kubectl scale deployment details-v1 --replicas=0`
- run the test again, this time it should fail

### DIFF
--- a/frontend/cypress/integration/common/kiali_alert.ts
+++ b/frontend/cypress/integration/common/kiali_alert.ts
@@ -1,0 +1,16 @@
+import { Then } from '@badeball/cypress-cucumber-preprocessor';
+
+Then(`user should see no Istio Components Status`, () => {
+  cy.intercept({
+    pathname: '**/api/istio/status*',
+    query: {
+      objects: ''
+    }
+  }).as('istioStatus');
+
+  cy.request('GET', '/api/istio/status')
+  cy.wait('@istioStatus');
+  cy.waitForReact();
+  cy.get('[data-test="istio-status-danger"]',{ timeout: 1000 }).should('not.exist');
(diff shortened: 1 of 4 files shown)


frontend/cypress/integration/common/kiali_alert.ts
```diff
@@ -0,0 +1,16 @@
...
+
+Then(`user should see no Istio Components Status`, () => {
+  cy.intercept({
+    pathname: '**/api/istio/status*',
+    query: {
+      objects: ''
+    }
+  }).as('istioStatus');
+
+  cy.request('GET', '/api/istio/status')
```
hhovsepy, line 11: IMO this request will not happen in Kiali UI side, but from cypress itself, so the next wait method will not be satisfied.

hhovsepy: you might need to click refresh button here instead of cy.request

matejnesuta: Yep. I checked the Cypress docs and you are right. The cy.request is now replaced.

hhovsepy (approved):
lgtm

jshaughn (approved):
Lgtm!

PR #6440, opened 2023-08-04
Cypress automation for 'Service details of ServiceEntry fails' fix

matejnesuta:
Related to #6295.

Do we want to backport this test to 1.65 as well?


frontend/cypress/integration/common/wizard_istio_config.ts
```diff
@@ -141,3 +141,12 @@ Then('{string} should not be referenced anymore', (gateway: string) => {
   ensureKialiFinishedLoading();
   cy.get(`a[data-test="k8sgateway-bookinfo-${gateway}"]`).should('not.exist');
 });
+
+Then('{string} details information for service {string} can be seen',(host: string, name: string)=>{
```
hhovsepy, line 145: As the #id's below are hardcoded with 'ES' and 'SE', I would suggest here to add 'service entry' instead of 'service' in name.

hhovsepy (approved):
lgtm, thank you @matejnesuta

PR #6443, opened 2023-08-04
PFT graph enhancements

jshaughn:
Epic https://github.com/kiali/openshift-servicemesh-plugin/issues/99

This PR makes two enhancements:
- small changes to the layout to reduce distances between nodes and lengths of edges.  This makes the graph slightly more compact to better fit larger graphs.
- add support for bringing labels to the top on hover, to be able to see labels that may be partially overlapped.
  - note that this adds a D3 dependency into package.json.  This is the same version already required by PFT, so I don't think it adds any size or risk, but I did want to call it out.

Before:
- Note the nodes are more zoomed out because the graph takes more space
image

After:
- The graph is easier to read because it takes less space and requires less zoom
- Note that the "bookinfo" label has ben brought forward by a hover
image

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -67,6 +67,7 @@
     "cytoscape": "3.15.5",
     "cytoscape-canvas": "3.0.1",
     "cytoscape-popper": "1.0.7",
+    "d3": "^5.16.0",
     "dagre": "0.8.5",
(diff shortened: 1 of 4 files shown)


leandroberetta (approved):
Looks good!

PR #6445, opened 2023-08-04
Update tls value for keycloak helm

nrfox:
Pinning to the last known working version of keycloak.

Workaround but not a complete solution for #6455


hhovsepy (approved):
The workaround is fine on my local.

Issue #6243, opened 2023-06-07
[Multicluster] provide links to external Kialis without requiring istio secrets


nrfox:
Today Kialis in "single cluster" mode require secrets to create links to external kialis. This is effectively putting the Kiali instance in "multi cluster" mode. There needs to be a way to for Kiali to generate links to external Kialis without requiring access to external clusters. This can be a simple as adding a config value that Kiali would use to generate these links. Something like `externalKialis: {"name": ..., "url": ...}`.

jshaughn:
That seems reasonable.  Did the former secrets do anything for us other than establish the link?  Users would still have to log into the remote Kiali, etc, right?

jmazzitelli:
> Did the former secrets do anything for us other than establish the link?

AFAIK, that's all the old secrets did for us. It told us the name of the remote cluster and helped us determine the remote kiali URL.

jshaughn:
So how would we handle the situation that someone was already using the experimental feature and has secrets already, but isn't trying to set up our new MC ability?  Just documentation/release notes to delete the secrets and optionally add this proposed config?

jmazzitelli:
They still need the istio secrets - those are required for istio MC to work. They just need the kiali secrets so kiali can work.

nrfox:
This isn't working like I would expect. I want to use this feature for Kiali to show multiple clusters on the "mesh" page without giving access to those clusters. My kiali config in the "east" cluster looks like:
```
kiali_feature_flags:
  clustering:
    kiali_urls:
      - cluster_name: west
        instance_name: kiali
        namespace: istio-system
        url: kiali-west.io
```

but still I just see "east"
![Screenshot from 2023-12-05 21-01-35](https://github.com/kiali/kiali/assets/6226732/44211538-31ce-4bbe-919a-3a853db34186)

hhovsepy:
The `url` param here expects a protocol as well.

hhovsepy:
@nrfox right, this feature was implemented partly, I saw your PR, so assigning this to you, thank you.

PR #6452, opened 2023-08-07
Added new config for kiali URL per cluster name mapping

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6243

Operator PR https://github.com/kiali/kiali-operator/pull/672

A new configuration is added in Kiali configmap, called 'kiali_urls', which is a mapping between cluster name and Kiali URL.
This value will be used if Kiali Service annotation 'kiali.io/external-url' does not have value.

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -330,16 +330,14 @@ func (in *MeshService) discoverKiali(ctx context.Context, clusterName string, r 
 	for _, d := range services {
 		kiali := convertKialiServiceToInstance(&d)
 		// If URL is already populated (because of an annotation), trust that because it's user configuration.
-		// If empty, check Kiali URL configured per cluster name.
-		if kiali.Url == "" {
-			for _, cfgurl := range config.Get().KialiFeatureFlags.Clustering.KialiURLs {
-				if cfgurl.ClusterName == clusterName {
-					kiali.Url = cfgurl.URL
-					break
-				}
+		// But if Kiali URL configured per cluster name, instance name and namespace, then use that URL.
+		for _, cfgurl := range config.Get().KialiFeatureFlags.Clustering.KialiURLs {
+			if cfgurl.ClusterName == clusterName && cfgurl.InstanceName == kiali.ServiceName && cfgurl.Namespace == kiali.Namespace {
+				kiali.Url = cfgurl.URL
+				break
 			}
 		}
-		// If still empty, only guess ourselves on our own cluster.
+		// If URL is still empty, only guess ourselves on our own cluster.
 		if kiali.Url == "" && clusterName == in.conf.KubernetesConfig.ClusterName {
 			kiali.Url = httputil.GuessKialiURL(r)
 		}
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -330,7 +330,15 @@ func (in *MeshService) discoverKiali(ctx context.Context, clusterName string, r 
 	for _, d := range services {
 		kiali := convertKialiServiceToInstance(&d)
 		// If URL is already populated (because of an annotation), trust that because it's user configuration.
-		// Only guess ourselves on our own cluster.
+		// If empty, check Kiali URL configured per cluster name.
+		if kiali.Url == "" {
+			for _, cfgurl := range config.Get().KialiFeatureFlags.Clustering.KialiURLs {
+				if cfgurl.ClusterName == clusterName {
+					kiali.Url = cfgurl.URL
+				}
+			}
+		}
+		// If still empty, only guess ourselves on our own cluster.
 		if kiali.Url == "" && clusterName == in.conf.KubernetesConfig.ClusterName {
 			kiali.Url = httputil.GuessKialiURL(r)
 		}
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -335,6 +335,7 @@ func (in *MeshService) discoverKiali(ctx context.Context, clusterName string, r 
 			for _, cfgurl := range config.Get().KialiFeatureFlags.Clustering.KialiURLs {
 				if cfgurl.ClusterName == clusterName {
 					kiali.Url = cfgurl.URL
+					break
 				}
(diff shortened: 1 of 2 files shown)


jmazzitelli (dismissed):
this looks good, but you'll need to rebase on master (a new requirement for our PRs - the CI tests must pass on the latest master code)

aljesusg (approved):
LGFM

business/mesh.go
```diff
@@ -133,7 +133,16 @@ func (in *MeshService) discoverKiali(ctx context.Context, clusterName string, r
 	for _, d := range services.Items {
 		kiali := convertKialiServiceToInstance(&d)
 		// If URL is already populated (because of an annotation), trust that because it's user configuration.
-		// Only guess ourselves on our own cluster.
+		// If empty, check Kiali URL configured per cluster name.
```
nrfox, line 136: IMO the kiali config should trump the annotation. They are both user configuration and, as user, it would be surprising if something I set on the kiali config was applied for any reason.

business/mesh.go
```diff
@@ -133,7 +133,16 @@ func (in *MeshService) discoverKiali(ctx context.Context, clusterName string, r
...
+		if kiali.Url == "" {
+			for _, cfgurl := range config.Get().KialiFeatureFlags.Clustering.KialiURLs {
+				if cfgurl.ClusterName == clusterName {
+					kiali.Url = cfgurl.URL
+					break
+				}
```
nrfox, line 142: It seems like having `KialiURLs` be an array enables there to be multiple Kiali instances with distinct urls on a single cluster but when looping over them we are just choosing the first URL. How would we even associated the correct URL with the correct kiali instance? Note that `services` here all comes from a single cluster but `var instances []kubernetes.KialiInstance` is an array so presumably there could be multiple kiali instances on a single cluster.

nrfox (changes requested):

business/mesh.go
nrfox: I don't see why we need to support multiple kiali instances, but if we're keeping the response the same then it seems like you'll need to add service name/namespace to the `KialiURL` struct in order to associate the correct url with the correct service.

config/config.go
```diff
@@ -448,7 +448,14 @@ type CertificatesInformationIndicators struct {
 
 // Clustering defines configuration around multi-cluster functionality.
 type Clustering struct {
-	EnableExecProvider bool `yaml:"enable_exec_provider,omitempty" json:"enable_exec_provider"`
+	EnableExecProvider bool       `yaml:"enable_exec_provider,omitempty" json:"enable_exec_provider"`
+	KialiURLs          []KialiURL `yaml:"kiali_urls,omitempty" json:"kialiURLs,omitempty"`
```
nrfox, line 452: Should this be `kiali_urls` for the JSON tag as well? I can't seem to get these settings to apply when testing.

business/mesh.go
nrfox: @hhovsepy sorry for all the noise. We want to provide the URL config option when the user _doesn't_ have access to the other clusters but still wants a link to an external kiali. I think we need to create the `KialiInstance` from the config URL regardless of whether or not the `GetServices` call fails.

business/mesh.go
```diff
@@ -133,7 +133,14 @@ func (in *MeshService) discoverKiali(ctx context.Context, clusterName string, r
...
 		kiali := convertKialiServiceToInstance(&d)
 		// If URL is already populated (because of an annotation), trust that because it's user configuration.
-		// Only guess ourselves on our own cluster.
+		// But if Kiali URL configured per cluster name, instance name and namespace, then use that URL.
+		for _, cfgurl := range config.Get().KialiFeatureFlags.Clustering.KialiURLs {
+			if cfgurl.ClusterName == clusterName && cfgurl.InstanceName == kiali.ServiceName && cfgurl.Namespace == kiali.Namespace {
+				kiali.Url = cfgurl.URL
+				break
+			}
+		}
```
nrfox, line 142: I don't think this enables the main use case for this feature which is: having a url to an external kiali instance _without_ having to give kiali access to the cluster that external kiali instance is running on. If Kiali is only creating a `KialiInstance` for every service that it reads from the external cluster, when Kiali does not have access to the external cluster it will not create a `KialiInstance` for it even though it has the url field set on the config. I think this will require more changes than adding the url here.

hhovsepy: The feature request said  "_There needs to be a way to for Kiali to generate links to external Kialis without requiring access to external clusters. This can be a simple as adding a config value that Kiali would use to generate these links._ " 
If the Kiali does not have access to remote cluster, but the configuration contains a URL, then it can be shown to user, but will it be OK from UX and security perspective?

hhovsepy: @nrfox you mean basically show another lines in Mesh with a 'Cluster Name' and 'Kiali' link columns (the rest of columns empty if unable to access) for those clusters which are not accessible but are set in config?

nrfox: > If the Kiali does not have access to remote cluster, but the configuration contains a URL, then it can be shown to user, but will it be OK from UX and security perspective?

Presumably the user would have access even though Kiali doesn't. It's a separate Kiali so the user would have to login to that Kiali separately.

> @nrfox you mean basically show another lines in Mesh with a 'Cluster Name' and 'Kiali' link columns (the rest of columns empty if unable to access) for those clusters which are not accessible but are set in config?

Yes and in addition to the cluster column on the mesh page there are links that appear in the side panel when you click on a node on the graph. These should link to the external kiali if one is set.

nrfox (approved):

business/mesh.go
hhovsepy: > 
> Yes and in addition to the cluster column on the mesh page there are links that appear in the side panel when you click on a node on the graph. These should link to the external kiali if one is set.

@nrfox which node's side panel in Graph contains links to Kiali?

PR #6459, opened 2023-08-08
Consider d3 target to have multiple classes in pf graph

### LINKED ISSUE #6458: PFGraph throws console error when hovering an application label
An error message appears in the console when an application label on PFGraph is hovered:

image

This is the error message in the console:
```
tr.js:48 Uncaught TypeError: node.getAttribute is not a function
    at Selection.__WEBPACK_DEFAULT_EXPORT__ [as attr] (attr.js:48:1)
    at raise (groupExpanded.tsx:220:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994:1)
    at invokeGuardedCallback (react-dom.development.js:4056:1)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4070:1)
    at executeDispatch (react-dom.development.js:8243:1)
    at processDispatchQueueItemsInOrder (react-dom.development.js:8275:1)
    at processDispatchQueue (react-dom.development.js:8288:1)
    at dispatchEventsForPlugins (react-dom.development.js:8299:1)
```

ferhoyos:
** Describe the change **

When user hovers an application label in PF graph, application considers that d3 target can have only one class, but this is not correct and the error is produced. Instead of comparing the whole string, just check the first class is `pf-topology__group`

** Issue reference **

Fixes #6458

** PR testing **

After the fix is applied, no error message appears in the console.

### DIFF
--- a/frontend/src/pages/GraphPF/components/groupExpanded.tsx
+++ b/frontend/src/pages/GraphPF/components/groupExpanded.tsx
@@ -217,10 +217,14 @@ const BaseGroupExpandedComponent: React.FunctionComponent<BaseGroupExpandedProps
   // "knows" it needs to go one level higher to reach the proper grouping.
   const raise = e => {
     let target = e.target;
-    while (!d3.select(target).attr('class')?.startsWith('pf-topology__group ')) {
-      target = target.parentNode;
+    try {
+      while (d3.select(target).attr('class') !== 'pf-topology__group') {
+        target = target.parentNode;
+      }
+      d3.select(target.parentNode).raise();
+    } catch (err) {
+      // ignore when this logic doesn't work and travels too far up the chain
     }
-    d3.select(target.parentNode).raise();
   };
 
   const data = element.getData();
--- a/frontend/src/pages/GraphPF/components/groupExpanded.tsx
+++ b/frontend/src/pages/GraphPF/components/groupExpanded.tsx
@@ -217,7 +217,7 @@ const BaseGroupExpandedComponent: React.FunctionComponent<BaseGroupExpandedProps
   // "knows" it needs to go one level higher to reach the proper grouping.
   const raise = e => {
     let target = e.target;
-    while (d3.select(target).attr('class') !== 'pf-topology__group') {
+    while (!d3.select(target).attr('class')?.startsWith('pf-topology__group')) {
       target = target.parentNode;
     }
     d3.select(target.parentNode).raise();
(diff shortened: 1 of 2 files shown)


frontend/src/pages/GraphPF/components/groupExpanded.tsx
```diff
@@ -217,7 +217,7 @@ const BaseGroupExpandedComponent: React.FunctionComponent<BaseGroupExpandedProps
   // "knows" it needs to go one level higher to reach the proper grouping.
   const raise = e => {
     let target = e.target;
-    while (d3.select(target).attr('class') !== 'pf-topology__group') {
+    while (!d3.select(target).attr('class')?.startsWith('pf-topology__group ')) {
       target = target.parentNode;
     }
     d3.select(target.parentNode).raise();
```
jshaughn, line 223:
```suggestion
    try {
      while (d3.select(target).attr('class') !== 'pf-topology__group') {
        target = target.parentNode;
      }
      d3.select(target.parentNode).raise();
    } catch (err) {
      // ignore when this logic doesn't work and travels too far up the chain
    }
```

Unfortunately the change stops the mentioned console error, but also breaks the "raise" feature.  I think maybe we can keep the original code but catch and ignore errors when it doesn't find what it wants.

ferhoyos: Thanks for the suggestion @jshaughn. As you say, this is a hacky piece of code and I don't understand deeply what the 'raise' feature. In my case I just take `node.tsx`code as reference:

(https://github.com/kiali/kiali/blob/d6f9027e56125a3a031d0fd916f160820fa32152/frontend/src/pages/GraphPF/components/node.tsx#L323C1-L333C5)

jshaughn: We should probably add a try/catch in the other location as well.  It took me some time to get the node and group impls to work as expected, and you are right, they are slightly different.  But both are hacks.  The `raise` is basically re-ordering the d3 layers so that the user can see a partially occluded label.  This would not be necessary if PFT didn't overlap labels...

ferhoyos:
@jshaughn I have committed your suggestion and added try/catch clause to similar raise feature in node.tsx file. Ready for your review.

jshaughn (approved):
This looks good, thanks.  Hopefully in the future PFT will handle this on its own and we won't need this logic.  Until then this should suppress the console errors.

Issue #6460, opened 2023-08-08
"Kiali Istio Config page -- KIA0104 validation (failed)" failure


matejnesuta:
This test has been failing consistently both on our upstream OCP and on my local setup when using KinD. I think this might indicate a bug in Kiali, because it doesn't look like a test flake to me, but I am not able to investigate further.
![image](https://github.com/kiali/kiali/assets/45061792/ef47734f-f9f1-4cc0-9a1f-1cfecc62a08d)

hhovsepy:
The validation KIA0104 also is checking hosts in VS, in your testing env you have a 'bookinfo' VS with host '*', so this host is considered as a match for a AP's host 'missing.hostname'.
This needs to be documented in kiali.io.
The test should remove any VS before execution.

PR #6461, opened 2023-08-08
Remove wildcard VS bookinfo before checking KIA0104

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6460

'bookinfo' VS is a wildcard, which covers KIA0104 validation, so make sure it is removed before.


matejnesuta:
LGTM!

matejnesuta (approved):

Issue #6464, opened 2023-08-09
Wrong cluster badge shown in the page for reviews app detail when accessing it from a graph


matejnesuta:
Steps to reproduce the issue: 
- Go to the` /console/namespaces/bookinfo/applications/productpage?clusterName=east` 
- Access the reviews app from the west cluster via the minigraph
- The details for the app should contain a badge for `east` cluster, even though the URL says otherwise
- Refresh to see the badge changed to `west`
![image](https://github.com/kiali/kiali/assets/45061792/3ee164fc-783c-4d94-ab69-343a1d773048)

It looks like the badge is not re-rendering correctly when going through different apps via the mini graph. I have not encountered any similiar issues regarding workload or service mini graphs.

PR #6466, opened 2023-08-10
Fixed clusterName param.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6464
Problem comes from getting an older name of param "cluster".
TODO: For using 'clusterNameFromQuery' a refactoring is needed in Graph Options parsing.


nrfox (commented):
Now when I click on the "west" reviews application I see the"east" reviews application in the minigraph. I think it should be showing the "west" reviews application instead.

reviews_example

josunect (approved):
LGTM. 

Tested in single and multi cluster. 

The issue reported by @nrfox is addressed. 

I think this could be a good testing scenario for https://github.com/kiali/kiali/issues/6475

Also, I think we could create an issue with the TODO comment in this issue description, so it is not lost.

hhovsepy:
thank you @josunect for the review and scenarios, the new issue is here https://github.com/kiali/kiali/issues/6486

Issue #5848, opened 2023-02-16
Check Tempo works with kiali with jaeger-query


aljesusg:
Check Tempo works with kiali with jaeger-query

Related with https://github.com/kiali/kiali/issues/4278#issuecomment-980638393

iblancasa:
I can help with this

jshaughn:
@iblancasa Are you still planning to help here?

iblancasa:
Yes Jay. I’ll add some documentation this week.

Sorry for the delay.
Thanks!

jshaughn:
Thanks @iblancasa!  @aljesusg, how should we proceed here, documentation on the steps necessary to use Tempo, or something more formal?

jshaughn:
Hi @iblancasa , I think what we will want to do is turn this information into content for kiali.io.  Probably in this section: https://kiali.io/docs/configuration/p8s-jaeger-grafana/jaeger/.

iblancasa:
Awesome. In that case, I'll create a PR to the [kiali.io repository](https://github.com/kiali/kiali.io) as soon as I can :)

jshaughn:
Sounds great, if you need any guidance let us know.

Hronom:
Please can we add it?

Hronom:
Finally manage it to work https://github.com/kiali/kiali/issues/5849#issuecomment-1624311897

caerulescens:
I managed to get this all working with the latest version of the `tempo-distributed` chart by working from @Hronom and @iblancasa instructions. Thanks guys

aljesusg:
Thabks to verify.
Anyway @josunect did a demo in kiali.io you will see it soon

You can check the preview here https://deploy-preview-686--kiali.netlify.app/docs/tutorials/tempo/

Please feel free to add more information about it 🙏

caerulescens:
Sure thing. I looked over the docs, and that's about the same thing I had to do. Create a PVC for MinIO and connect Tempo. Although, I'm doing it all through Helm managed by ArgoCD.

aljesusg:
Great !!! If you find some problem with argocd please let us know it or if you prefer we can help you to contribute in the project with a tutorial about it 🤗

aljesusg:
Added tutorial to kiali.io

PR #6468, opened 2023-08-10
Adapt run-kiali script to use other tracing backend than Jaeger

josunect:
Ref. #5848 

To use it with the environment set in https://github.com/kiali/kiali.io/pull/686:

`./run-kiali.sh -pg 13000:3000 -pp 19090:9090 -app 8080 -es false -iu http://127.0.0.1:15014 -tr tempo-smm-query-frontend -ts tempo-smm-query-frontend -tn tempo`

### DIFF
--- a/hack/run-kiali.sh
+++ b/hack/run-kiali.sh
@@ -87,6 +87,8 @@ DEFAULT_LOCAL_REMOTE_PORTS_TRACING="16686:16686"
 DEFAULT_LOG_LEVEL="info"
 DEFAULT_REBOOTABLE="true"
 DEFAULT_TMP_ROOT_DIR="/tmp"
+DEFAULT_TRACING_APP="jaeger"
+DEFAULT_TRACING_SERVICE="tracing"
 
 # Process command line options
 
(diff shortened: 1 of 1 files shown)


hack/run-kiali.sh
```diff
@@ -680,7 +704,12 @@ start_port_forward_component() {
     else
       startmsg "The port-forward to ${COMPONENT_NAME} is being started on [${EXPECTED_URL}]"
       set -m
-      (while true; do ${CLIENT_EXE} port-forward -n ${ISTIO_NAMESPACE} ${PORT_FORWARD_DEPLOYMENT} --address=127.0.0.1 ${LOCAL_REMOTE_PORTS} > /dev/null; warnmsg "${COMPONENT_NAME} port-forward died - restarting on [${EXPECTED_URL}]"; sleep 1; done) &
+      if [ ${COMPONENT_NAME} == "Tracing" ]; then
```
jmazzitelli, line 707: You should always wrap env vars in quotes  when inside if stmts to ensure you always get an explicit argument
```suggestion
      if [ "${COMPONENT_NAME}" == "Tracing" ]; then
```

jmazzitelli (changes requested):
minor change. other than that, the code looks good.

josunect:
> minor change. other than that, the code looks good.

Thanks for the review, @jmazzitelli !

jmazzitelli (approved):

PR #6470, opened 2023-08-11
[hack] update to ci-kind-molecule hack script so you can specify the kind executable

jmazzitelli:
helps people run this via cron and not require to have kind in path


aljesusg (dismissed):
LGFM

jmazzitelli:
I think this finally gets it to work on my own laptop via cron

nrfox (approved):

Issue #6418, opened 2023-07-26
[CI] - Test flake applications list


nrfox:
The applications list e2e test is failing somewhat consistently. I think this is just the first test to run and it has nothing to do with that particular test itself. I believe the real issue is that the Kiali API is taking too long to respond and the request from the client (e2e test) is timing out.

From the logs:
```
 === RUN   TestApplicationsList
    applications_test.go:16: 
        	Error Trace:	/home/runner/work/kiali/kiali/tests/integration/tests/applications_test.go:16
        	Error:      	Received unexpected error:
        	            	Get "http://172.18.255.71:20001/kiali/api/namespaces/bookinfo/apps": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
        	Test:       	TestApplicationsList
```

Example of failed test run: https://github.com/kiali/kiali/actions/runs/5669354886/job/15362182500

Some retries in the Kiali e2e tests could help with this.

nrfox:
Haven't seen this particular error in a week since merging the fix. Closing this.

PR #6472, opened 2023-08-11
Do external health check before running tests

nrfox:
Add an external health check before running tests. My guess is that it takes a minute for the pods to be added to the service's load balancer pool even after the pods are ready so we need to do an external health check before running the tests.

Fixes (hopefully) https://github.com/kiali/kiali/issues/6418


jmazzitelli (dismissed):
LGTM
We will know it is good when this PR goes all green.

jmazzitelli (approved):
looks good now

PR #6473, opened 2023-08-11
Add GetMesh method to gather info about controlplanes

### LINKED ISSUE #6430: Multicluster - Update backend mesh model for multi-primary.
The mesh model needs to support multiple primaries with each configuration option (mtls, etc.) associated with a specific primary.

nrfox:
Creates a couple new models to store controlplane configuration per controlplane since with multi-primary there can be multiple controlplanes. Even with single cluster there can be multiple controlplanes when revisions are used. This will allow Kiali to better support control plane revisions in the future even for single cluster.

This adds the models but stops short of updating various methods that gather controlplane info to use them yet.

Fixes #6430

### DIFF
--- a/models/namespace.go
+++ b/models/namespace.go
@@ -5,12 +5,12 @@ import (
 
 	osproject_v1 "github.com/openshift/api/project/v1"
 	core_v1 "k8s.io/api/core/v1"
-
-	"github.com/kiali/kiali/config/dashboards"
 )
 
-const AmbientLabel = "istio.io/dataplane-mode"
-const AmbientValue = "ambient"
+const (
+	AmbientLabel = "istio.io/dataplane-mode"
+	AmbientValue = "ambient"
+)
 
 // A Namespace provide a scope for names
 // This type is used to describe a set of objects.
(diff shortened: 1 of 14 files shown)


nrfox:
This needs some more tests but opening it up in draft to get feedback for models and the rest of the implementation.

models/namespace.go
```diff
@@ -94,11 +91,8 @@ func CastProject(p osproject_v1.Project, cluster string) Namespace {
 	namespace.Cluster = cluster
 	namespace.CreationTimestamp = p.CreationTimestamp.Time
 	namespace.Labels = p.Labels
-	namespace.Annotations = make(map[string]string)
-	// Parse only annotations used by Kiali
-	if da, ok := p.Annotations[dashboards.DashboardTemplateAnnotation]; ok {
-		namespace.Annotations[dashboards.DashboardTemplateAnnotation] = da
-	}
+	namespace.Annotations = p.Annotations
```
leandroberetta, line 94: I have the same change for my work of indicating which namespace is primary and which are remote. I think this is necessary for my work of hiding some information in the control plane card. I'm detecting remotes with the annotation `topology.io/controlPlaneClusters`.

hhovsepy (commented):
lgtm, tested locally

leandroberetta (approved):
LGTM

PR #6479, opened 2023-08-15
New package.json target for test developement

matejnesuta:
The Cypress GUI runner currently opens all of the tests, no matter if used against single or multi-cluster setup. This can get annoying at best, because you must wait for all of the unrelated tests to finish, and contraproductive at worst, because the bookinfo detection only works for single cluster at the moment and it is able to ruin your cluster-wide bookinfo installation. 

The change in this PR is related to that. Now the `yarn cypress` command opens tests related to single cluster only, while the `yarn cypress:mc` opens tests related to multi-cluster.


ScriptingShrimp (approved):
This setting also cleared the results table, so there is no longer record of skipped tests caused by `not @crd-validation` setting and etc... Also retested locally, all passes. LGTM

[for the record](https://github.com/badeball/cypress-cucumber-preprocessor/blob/master/docs/tags.md):

> **Omit filtered tests**
> 
> By default, all filtered tests are made *pending* using `it.skip` method. If you want to completely omit them, set `omitFiltered` to `true`.
> 
> **Running a subset of scenarios**
> 
> Normally when running a subset of scenarios using `cypress run --env tags=@foo`, you could potentially encounter files containing no matching scenarios. These can be pre-filtered away by setting `filterSpecs` to `true`, thus saving you execution time.
>

PR #6480, opened 2023-08-15
Add sysctl info to README and hack script

nrfox:
Adds a section to the README and help message about increasing some kernel settings when setting up a multi-cluster environment locally with kind.

### DIFF
--- a/hack/README.adoc
+++ b/hack/README.adoc
@@ -53,11 +53,21 @@ Use this to access the Kiali UI that is installed locally in `minikube` or `kind
 
 * `link:purge-kiali-from-cluster.sh[]`
 
-Sets up the environment for the integration tests and runs them. This is the same setup that is used by the CI system.
+Removes all remnants of Kiali and Kiali Operator from a cluster.
 
 * `link:run-integration-tests.sh[]`
 
-Removes all remnants of Kiali and Kiali Operator from a cluster.
+Sets up the environment for the integration tests and runs them. This is the same setup that is used by the CI system.
+
+NOTE: When running the multi-cluster tests locally, it might be necessary to
+edit some kernel settings to allow for the kind clusters to be created.
+
+The following settings added to your `/etc/sysctl.d/local.conf` file
+should work but YMMV depending on your system:
+```bash
+fs.inotify.max_user_watches=524288
+fs.inotify.max_user_instances=512
+```
 
 * `link:run-kiali.sh[]`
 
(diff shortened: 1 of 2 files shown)


hack/README.adoc
```diff
@@ -53,11 +53,21 @@ Use this to access the Kiali UI that is installed locally in `minikube` or `kind
...
+
+NOTE: When running the multi-cluster tests locally, it might be necessary to
+edit some kernel settings to allow for the kind clusters to be created.
+
+The following settings added to your `/etc/sysctl.d/local.conf` file
+should work but YMMV depending on your system:
+```bash
+fs.inotify.max_user_watches=524288
+fs.inotify.max_user_instances=512
+```
```
jmazzitelli, line 70: You could tell the user to just change the kernel settings for the current runtime (this will revert back after a reboot). This (a) doesn't require the user to change the persistent config of the machine and (b) this command should be more cross-platform than the name of the file (e.g. my filename on Fedora 38 is different).

````suggestion
These commands will set the kernel settings temporarily. After a reboot, your system will revert back to its original settings:
```bash
sudo sysctl fs.inotify.max_user_watches=524288
sudo sysctl fs.inotify.max_user_instances=512
```
If you wish to make these settings permanent, the following lines added to your sysctl configuration file should work (depending on your operating system, this configuration file can be something like `/etc/sysctl.d/local.conf`, `/etc/sysctl.conf`, or something else. Refer to the `man sysctl` docs for details):
```bash
fs.inotify.max_user_watches=524288
fs.inotify.max_user_instances=512
```
````

File: hack/run-integration-tests.sh [hunk 2]
```diff
@@ -38,6 +38,14 @@ Valid command line arguments:
...
   -h|--help:
        this message
+
+NOTE: When running the multi-cluster tests locally, it might be necessary to
+edit some kernel settings to allow for the kind clusters to be created.
+
+The following settings added to your /etc/sysctl.d/local.conf file
+should work but YMMV depending on your system:
+fs.inotify.max_user_watches=524288
+fs.inotify.max_user_instances=512
```
jmazzitelli, line 48:
```suggestion
The following settings added to your sysctl config file should work (the filename will be something like '/etc/sysctl.d/local.conf' - refer to your operating system 'man sysctl' docs to determine which file should be changed):
fs.inotify.max_user_watches=524288
fs.inotify.max_user_instances=512
```

jmazzitelli (changes requested):
some changes to help those on different OSes

hack/README.adoc
nrfox: That's more helpful than "YMMV" :)

jmazzitelli (approved):

PR #6482, opened 2023-08-16
[make] fix molecule test make target

jmazzitelli:
This auto-detection of hosts should never occur unless the cluster is openshift. Without this fix, the getent commands hang until they timeout causing the test times to go from 2 minutes to 16 minutes. lol


hhovsepy (approved):
lgtm

PR #6493, opened 2023-08-17
Cypress: refactoring of feature files tags

ScriptingShrimp:
I'm suggesting using the rule: single tag for a file at first line. 

things changed: 

- added unique tags for each .feature file on first line
- added comment not to remove the first line :)
- redundant tags removed from the scenarios
- fixed to same indentation
- linted with BDD linter

We can tag whole file at the beginning of the file instead tagging every scenario separately. This is important for parallelization - lets say we can use something like `head -n 1 -q ../frontend/cypress/integration/featureFiles/*.feature` and that will return tag for an each file.


frontend/cypress/integration/featureFiles/wizard_k8sgw_api_routing.feature
```diff
@@ -52,14 +52,12 @@ Feature: Service Details Wizard: K8s Gateway API Routing
...
 
   @gateway-api
-  @wizard-k8s-routing
   @bookinfo-app
   Scenario: See a K8s Gateway generated with warning
     When user clicks in the "Istio Config" table "G" badge "reviews-gateway" name row link
     Then user sees the "kind: Gateway" regex in the editor
 
   @gateway-api
-  ## @wizard-k8s-routing
```
ScriptingShrimp, line 62: @matejnesuta do you know why this was commented?

matejnesuta: Possibly because you don't have to open the wizard itself to delete the object. The scenario was written by @hhovsepy, maybe he knows better.

matejnesuta (approved):
Looks good to me!

Issue #5569, opened 2022-10-17
Use Opentelementry Exporter instead of Jaeger


nrfox:
For debugging Kiali itself, Kiali currently uses [the Jaeger exporter](https://github.com/kiali/kiali/blob/master/observability/tracing.go#L39-L42). We should migrate to the [Opentelemetry Exporter](https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace#Exporter) to be compatibile with any otel backend.

krishnaindani:
hey, what changes are supposed to be made in order to migrate to otel? Would like to lean more and maybe attempt to solve this issue

israel-hdez:
The OTEL collector supports ingesting Jaeger traces and sending them to any OTEL compatible back-end.

What would be the benefit of migrating to the OTEL exporter if we still would need a Jaeger (or Jaeger compatible) back-end?

Xunzhuo:
I think this can make kiali more compatible to other tracing systems, right? @nrfox

nrfox:
@krishnaindani thanks for the interest. The work involved will be replacing the jaeger client here: https://github.com/kiali/kiali/blob/master/observability/tracing.go#L39-L42 with the otelhttp client: https://pkg.go.dev/go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp. You may also need to update or add some options to the kiali config: https://github.com/kiali/kiali/blob/master/config/config.go#L85-L91 depending on the format of the url that the otlptracehttp lib expects.

> What would be the benefit of migrating to the OTEL exporter if we still would need a Jaeger (or Jaeger compatible) back-end?
>
> I think this can make kiali more compatible to other tracing systems, right? @nrfox

Yes we can export traces from the kiali server to more backends without needing the collector and the Jaeger clients are deprecated: https://www.jaegertracing.io/docs/1.38/client-libraries/#deprecating-jaeger-clients. Although I'm not sure if that includes the otel jaeger exporter that Kiali currently uses.

jshaughn:
@krishnaindani We're still looking for a contribution if you're interested.

jshaughn:
cc @aljesusg , does this play into the Tempo work at all?

aljesusg:
Yes @jshaughn I am going to assign this myself until we check

PR #6499, opened 2023-08-21
Support for OpenTelemetry exporter

josunect:
Fixes https://github.com/kiali/kiali/issues/5569

### Configuration

Maybe it is a bit confusing, but, the idea is to just set up or the `jaeger_collector` OR the `otel_collector`
- The Jaeger collector by default sends traces to `http://0.0.0.0:14268/api/traces`. It needs the protocol in the URL. 
- The OTel collector can use http, https or grpc. There is a TODO to improve grpc config using TLS (Need to set up the cert parameters). The URL does not use the protocol. By default, the http exporter sends the traces to `/v1/traces`

Configuration example: 

```
  observability: 
    tracing: 
      collector_url: "localhost:4318"
      enabled: true
      otel_collector: 
        enabled: true
        protocol: "http"
      jaeger_collector: 
        enabled: false
```

### Compatibility

Not sure how this configuration could work with backwards compatibility, as now the jaeger_collector would need to be enabled (If not, it will throw an error)

### Test

The OTel collectors were tested using Grafana Tempo and Tanka. 
- Visit the workloads list page
- curl (It should return the traces generated by Kiali): `curl -G -s http://localhost:16686/api/search --data-urlencode 'tags=package=business'`

### Operator 

Operator PR: https://github.com/kiali/kiali-operator/pull/673


jmazzitelli:
You should do the same kind of thing we do in the `auth` section (see https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.auth). We have "auth.strategy" and whatever value you set in there, there is an optional section with the same name that contains configuration specific for that strategy.

For example:

```
auth:
  strategy: openshift
  openshift:
    client_id_prefix: foo
    token_inactivity_timeout: 5
    ...and others...
```

But if auth.strategy is "openid", then there is a separate section for that:

```
auth:
  strategy: openid
  openid:
    client_id: foo
    disable_rbac: true
    ... and others...
```

So, what this means is, for your new setting, you would add a new `collector_impl` or `collector_type` setting (whatever name you think is best), and each type would have its own separate section like you have (but remove each "enabled" flag that you added - no need for that since this new "collector_impl" or "collector_type" setting would tell you which is enabled)... like this:

```
observability: 
    tracing: 
      collector_url: "localhost:4318"
      enabled: true
      collector_type: "otel" # each value ("otel" or "jaeger") would have a matching section below.
      otel: 
        protocol: "http"
      jaeger: 
        something_here: foo
```

Note that if, say, jaeger doesn't need any extra config, then you don't have to worry about its own section (just like for "auth.strategy: anonymous" we do not have a "auth.anonymous" section because there is nothing extra to configure for anonymous mode).

josunect:
> You should do the same kind of thing we do in the `auth` section (see https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.auth). We have "auth.strategy" and whatever value you set in there, there is an optional section with the same name that contains configuration specific for that strategy.
> 
> For example:
> 
> ```
> auth:
>   strategy: openshift
>   openshift:
>     client_id_prefix: foo
>     token_inactivity_timeout: 5
>     ...and others...
> ```
> 
> But if auth.strategy is "openid", then there is a separate section for that:
> 
> ```
> auth:
>   strategy: openid
>   openid:
>     client_id: foo
>     disable_rbac: true
>     ... and others...
> ```
> 
> So, what this means is, for your new setting, you would add a new `collector_impl` or `collector_type` setting (whatever name you think is best), and each type would have its own separate section like you have (but remove each "enabled" flag that you added - no need for that since this new "collector_impl" or "collector_type" setting would tell you which is enabled)... like this:
> 
> ```
> observability: 
>     tracing: 
>       collector_url: "localhost:4318"
>       enabled: true
>       collector_type: "otel" # each value ("otel" or "jaeger") would have a matching section below.
>       otel: 
>         protocol: "http"
>       jaeger: 
>         something_here: foo
> ```
> 
> Note that if, say, jaeger doesn't need any extra config, then you don't have to worry about its own section (just like for "auth.strategy: anonymous" we do not have a "auth.anonymous" section because there is nothing extra to configure for anonymous mode).

Thank you, @jmazzitelli ! This is very useful.

aljesusg (approved):
LGFM code no tested, I'l try it later

observability/tracing.go
```diff
@@ -117,3 +134,72 @@ func StartSpan(ctx context.Context, funcName string, attrs ...attribute.KeyValue
...
+
+					// TODO: Support TLS
+					exporter, err := otlptracegrpc.New(ctx, otlptracegrpc.WithInsecure(),
+						otlptracegrpc.WithEndpoint(collectorURL),
+						otlptracegrpc.WithDialOption(grpc.WithBlock()),
+					)
+
+					return exporter, err
+				} else {
+					return nil, errors.New("Error in configuration options getting the observability exporter")
```
jmazzitelli, line 197: Actually, make these error messages more specified. This error and the one on line 201 are identical - when you see this, you have no idea if its the protocol that is bad or the collector type that is bad.
```suggestion
					return nil, errors.New("Error in configuration options getting the observability exporter - invalid otel.protocol.")
```

observability/tracing.go
```diff
@@ -117,3 +134,72 @@ func StartSpan(ctx context.Context, funcName string, attrs ...attribute.KeyValue
...
+				}
+			}
+		} else {
+			return nil, errors.New("Error in configuration options getting the observability exporter")
```
jmazzitelli, line 201:
```suggestion
			return nil, errors.New("Error in configuration options getting the observability exporter. Invalid collector type.")
```

jmazzitelli (changes requested):
a couple minor changes needed

jmazzitelli (dismissed):
lgtm

jmazzitelli (approved):

Issue #6502, opened 2023-08-22
Fixed help icon in Envoy tab


josunect:
Fix the position of the help icon (In the right corner of the table) in the Envoy tab. 

The position of the icon is fixed, so it shouldn't move when we scroll down. 

![image](https://github.com/kiali/kiali/assets/49480155/daaeab9f-2c9d-410f-9d6f-7a7a338cb9bb)

PR #6503, opened 2023-08-22
Change fixed position for help icon in Envoy tab

josunect:
Fixes: #6502

This is a very small fix to not fix the position of the help icon (In the right corner of the table) in the Envoy tab. 
Previous to this fix: 

image

The position of the icon was fixed, so it didn't move when we scroll down. 

After the fix: 

image


hhovsepy (approved):
verified

PR #6508, opened 2023-08-22
[v1.65] Backport of https://github.com/kiali/kiali/pull/6389

nrfox:
Backport of #6389 to v1.65. Just a find/replace of `assert` --> `require` in e2e tests so that tests will fail without panicing.


aljesusg (approved):
LGFM

PR #6509, opened 2023-08-22
Graph legend for Patternfly topology

leandroberetta:
Resolves https://github.com/kiali/openshift-servicemesh-plugin/issues/183

The following PR adds the legend to the Patternfly graph:

image

It's very similar to the old one but there are some changes due to differences in Patternfly Topology vs Cytoscape Topology.

In the near future, there is a possibility to contribute to PF's code to add the old service, service entry and operation nodes svc for us to use. At the moment, there is not way to change the default shapes.

The rest of the legend:

image
image


josunect (changes requested):
Nice work, @leandroberetta ! 

It looks good:

image

Just a minor thing: In the dark mode, the Workload looks different to the other Node Shapes: 

image

Maybe it shouldn't have a background? Also, for the Node colors, I think they would be more visible when they don't have a background.

leandroberetta:
@josunect I fixed the issue, this was also happening for the old graph.

josunect (approved):
LG! 

image

jmazzitelli (commented):
I have a question with the node color of red:
image

Is that how we labeled it before.. as `Danger`?

"Danger" typically means something "likely to cause harm or injury."

I'm not sure if "Danger" is the right word to use here. I'm curious if UX has suggested that word? Can we think of another word? It just doesn't seem to fit, IMO.

What about "Unhealthy" or something of that sort?

leandroberetta:
> I have a question with the node color of red: image
> 
> Is that how we labeled it before.. as `Danger`?
> 
> "Danger" typically means something "likely to cause harm or injury."
> 
> I'm not sure if "Danger" is the right word to use here. I'm curious if UX has suggested that word? Can we think of another word? It just doesn't seem to fit, IMO.
> 
> What about "Unhealthy" or something of that sort?

Yes, that is how it was before, I can change it to Unhealthy...

jmazzitelli (approved):
I ran it looking at both cy and pf graph impls.. looks good. "Danger" has been replaced with "Unhealthy" which is a better word to use IMO.

jmazzitelli:
This might not be something to worry about.. but notice the legend in the cy graph shows "Normal" whereas the pf graph shows "Healthy".

cy:
image

pf:
image

leandroberetta:
> This might not be something to worry about.. but notice the legend in the cy graph shows "Normal" (gray color) whereas the pf graph shows "Healthy" (green color).
> 
> cy: image
> 
> pf: image

Yeah, this is a changed I've made, and I did it because in the new graph, there are some minor differences... I noticed that in the CY graph, the normal (grey), now in the PF graph, is used for idle. We can change the CY graph to say "Healthy" instead of normal, but I think it's ok to leave as it is.

Issue #6437, opened 2023-08-02
Remote cluster istio-system namespace card show data from primary control plane


leandroberetta:
When using the primary-remote multicluster setup, the istio-system card from the remote, should not show data related to metrics (in fact, is showing information from the primary).

Epic: https://github.com/kiali/kiali/issues/5618

hhovsepy:
Probably the `clusterName` param is send wrongly or lost somewhere.

leandroberetta:
> Probably the `clusterName` param is send wrongly or lost somewhere.

Not really, this is happening because in the code, the control plane metrics are fetch when the namespaces matches de "Istio" namespace, but we have two of them, resulting on the same metrics being fetched for both of them.

I can distinguish between primary and remote istio-system namespaces with an annotation called topology.istio.io/controlPlaneClusters (today, Kiali is not adding all the annotations to the Namespace object, it is trimming them to have just the Kiali's dashboards ones, we might need to allow other annotations too).

I might also need to add a new field in the frontend to distinguish between a primary control plane namespace and a remote one.

I'm doing some testing.

PR #6514, opened 2023-08-23
Skip metrics for remote cluster control plane.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6437


leandroberetta:
I think it would be nice to also hide the "Control plane metrics" text, and I think the canary information is not right there for the remote cluster as there is no istiod process there.

Same goes for the outbound policy I think.

Looking at the issue, I mentioned metrics, but I was thinking in these other information too. If you want we can split this issue into two.

One thing I believe if that we remove the control plane metrics, the remote cluster card can be smaller, like the other ones... what do you think about it?

hhovsepy:
@leandroberetta I have added another badge "Remote Cluster" for remote control planes.
Screenshot from 2023-08-24 18-12-32

I think your ideas are correct and we would need a UX discussion.

leandroberetta:
> @leandroberetta I have added another badge "Remote Cluster" for remote control planes. Screenshot from 2023-08-24 18-12-32
> 
> I think your ideas are correct and we would need a UX discussion.

Nice addition! Yes, I will add a bullet for our next meeting on tuesday.

hhovsepy:
@leandroberetta I removed the metrics, canary upgrade and outbound traffic fields from remote control plan card.
However the card length is remaining the same not to break the design.

hhovsepy:
Screenshot from 2023-09-13 14-19-52

andrew-ronaldson:
It can be smaller. I was first thinking to add an Empty state component under control plane but probably fine to just get rid of it altogether.

hhovsepy:
Remote control plane card size is smaller (half of the regular long one), so easy to differentiate against the small data plane one:
Screenshot from 2023-09-14 13-16-07

hhovsepy:
@leandroberetta @andrew-ronaldson this PR is ready to review.

andrew-ronaldson (approved):
I think this works. Can't really get around that empty space created by the missing metrics so I think this LGTM.

leandroberetta:
> I think this works. Can't really get around that empty space created by the missing metrics so I think this LGTM.

Hi, and what about making the istio-system card (remote, with no control plane), the same size as the regular ones, and avoid ordering or positioning first, in that way, the cards with a control plan will be bigger and will stack vertically, and the rest (regular namespaces and remote namespaces card) will just stack as usual.

leandroberetta:
@hhovsepy code looks good, so if @andrew-ronaldson is ok with the UX side, I can approve, but I wanted to hear any opinion about my last comment, so I'll wait a little bit for the approval.

andrew-ronaldson:
@leandroberetta I like that suggestion.

hhovsepy:
Screenshot from 2023-09-22 20-45-21
@leandroberetta now it has the same size as data plane card, but it still shows above the control plane card.

leandroberetta:
> Screenshot from 2023-09-22 20-45-21 @leandroberetta now it has the same size as data plane card, but it still shows above the control plane card.

So, my idea is to avoid the gray are with no cards, my proposal is that the control plane namespaces that really does not have any control plane, just are remote istio-system namespaces, can be treated as a normal card in terms or ordering or now that I think, I would order the cards putting first the ones that are full length, then the other control plane cards, then the data plane ones.

What do you think about this?. My goal really is to avoid a blank (gray) space on the top.

hhovsepy:
@leandroberetta I agree, now it is much better to have sorted the primary first:
Screenshot from 2023-09-25 08-38-32

leandroberetta (approved):
Very nice! LGTM

Issue #5849, opened 2023-02-16
Support tempo reading traces


aljesusg:
# Adapt Kiali to query Tempo directly

The better solution (imo) would be to add a client specific to Tempo. Looking at the code this would require a couple of changes:

The current client uses the Jaeger data model (for instance [JaegerSingleTrace](https://github.com/kiali/kiali/blob/f432cf15ffdb23c02fc04555145caa17a4c3fb66/jaeger/types.go#L23-L26)) but Tempo returns traces in the OpenTelemetry format (JSON or Protobuf). The proto definition: [opentelemetry-proto - opentelemetry/proto/trace/v1/trace.proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/trace/v1/trace.proto).
Luckily there is already a lot of code out there to convert between OpenTelemetry and Jaeger, Zipkin. So you can convert OpenTelemetry into Jaeger in the client or opt to switch Kiali to use the OpenTelemetry format all together and do the conversion in the Jaeger client.

The search results of Jaeger and Tempo are different: while Jaeger returns a list of full traces, Tempo only returns metadata of the traces. See https://grafana.com/docs/tempo/latest/api_docs/#search
This might be a problem if you need the full trace data for visualisations like the workload detail: https://kiali.io/docs/features/tracing/#workload-detail

-------------------

If not make a switch to 

Jaeger => Read from GRPC
Tempo => Read from TraceQL

jshaughn:
It sounds right, @aljesusg.  I think we want to hide as much as possible in the lower layers so that the client/ui code doesn't care if it's Tempo, Jaeger or something else.  In other words,  Our API should probably return a `Trace` object defined for our needs, and populated via some adapter code based on the configured tracing client.  Is that what you are proposing?    We could let the client specify whether it needs only metadata, (the default) or the full trace.    And then we'd make sure to return the full trace and hide the complexity.  Or, we could just return to the client a flag like `isFullTrace` and the client would need to query again if it needs more detail.

aljesusg:
[...]

PR #6518, opened 2023-08-24
Kiali client for Tempo API 

josunect:
Fixes https://github.com/kiali/kiali/issues/5849

There are some main changes: 

- The Jaeger package is renamed to Tracing so it can include different tracing providers. Inside the Tracing package we can find the Jaeger package, but also includes a new Tempo package and OTEL package. Tempo returns traces in OpenTelemetry format, so OTEL package includes specific data types and converters between OTEL and Jaeger traces (The documentation for reference: https://opentelemetry.io/docs/specs/otel/trace/sdk_exporters/jaeger/#attributes).
- The client creates the http client based on the tracing provider. At the moment, Jaeger an Tempo are supported. 
- The tempo http client is slow because it queries the search endpoint that returns just a few data but it is also needed all the information from the trace, like the span list. It need to do a request for each trace. 

There are still some TODOs: 

- [x] Review the Tempo http - still is a draft. 
- [x] At the moment just one span is converted
- [x] Review the conversion (Some fields are still empty) 
- [x] Probably more refactoring 
- [x] There are not error traces. Probably they are not captured correctly? 
- [x] The trace details says: "Incomplete data, check Span Details" (Investigate): There are some spans with no spanStats ()

Related PRs:
- [ ] Operator for new tracing options (Trace provider) https://github.com/kiali/kiali-operator/pull/679/files
- [ ] Docs: https://github.com/kiali/kiali.io/pull/692

image

image

image


josunect:
Steps for testing (Locally): 

- Create a Tempo environment. The tested environment was created with Tempo Tanka. 
- Kiali configuration: 
```
  tracing:
    enabled: true
    provider: "tempo"
    in_cluster_url: http://localhost:16685
    url: http://localhost:16686
```

And run the run-kiali hack script with args: 

`/run-kiali.sh -pg 13000:3000 -pp 19090:9090 -app 8080 -es false -iu http://127.0.0.1:15014 -tr query-frontend -ts query-frontend -tn tempo`

jmazzitelli:
> * Create a Tempo environment. The tested environment was created with Tempo Tanka.

Can you provide the steps you did this? Because I just read this at [Tanka : Deploy on Kubernetes with Tanka](https://grafana.com/docs/tempo/latest/setup/tanka/#deploy-on-kubernetes-with-tanka)

`A Kubernetes cluster with at least 40 CPUs and 46GB of memory for the default configuration. Small ingest or query volumes could use a far smaller configuration.`

Clearly, we don't have 40 CPUs and 46 GB on our laptops :) So I'm assuming you did something else to test with.

jmazzitelli:
Here is how I set it up, step by step - I used minikube for my cluster:

1. Start the cluster: `hack/k8s-minikube start`
2. Store this in a file `minio.yaml`:
```yaml
apiVersion: v1
kind: Namespace
metadata:
  Name: minio
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: minio
  namespace: minio
spec:
  selector:
    matchLabels:
      app: minio
  strategy:
    type: Recreate
  template:
    metadata:
      labels:
        # Label is used as selector in the service.
        app: minio
    spec:
      # Refer to the PVC created earlier
      volumes:
        - name: storage
          emptyDir:
            sizeLimit: 256Mi
      initContainers:
        - name: create-buckets
          image: busybox:1.28
          command:
            - "sh"
            - "-c"
            - "mkdir -p /storage/tempo-data"
          volumeMounts:
            - name: storage # must match the volume name, above
              mountPath: "/storage"
      containers:
        - name: minio
          # Pulls the default Minio image from Docker Hub
          image: minio/minio:latest
          args:
            - server
            - /storage
            - --console-address
            - ":9001"
          env:
            # Minio access key and secret key
            - name: MINIO_ROOT_USER
              value: "minio"
            - name: MINIO_ROOT_PASSWORD
              value: "minio123"
          ports:
            - containerPort: 9000
            - containerPort: 9001
          volumeMounts:
            - name: storage # must match the volume name, above
              mountPath: "/storage"
---
apiVersion: v1
kind: Service
metadata:
  name: minio
  namespace: minio
spec:
  type: ClusterIP
  ports:
    - port: 9000
      targetPort: 9000
      protocol: TCP
      name: api
    - port: 9001
      targetPort: 9001
      protocol: TCP
      name: console
  selector:
    app: minio
```
3. Deploy Minio: `kubectl apply -f minio.yaml`
4. Test that it installed successfully:
4a. `kubectl port-forward --namespace minio service/minio 9001:9001`
4b. Point browser to http://localhost:9001/
4c. Login as `minio` / `minio123`
4d. Verify that the Buckets page lists `tempo-data`.
5. Create a secret in a new "tempo" namespace that points to the Minio bucket:
```bash
kubectl create namespace tempo &&
kubectl create -n tempo secret generic tempostack-dev-minio --from-literal=bucket="tempo-data" --from-literal=endpoint="minio.minio:9000" --from-literal=access_key_id="minio" --from-literal=access_key_secret="minio123"
```
6. Install OLM: `make -e CLUSTER_TYPE=minikube olm-install`
7. Install Tempo operator: `kubectl create -f https://operatorhub.io/install/tempo-operator.yaml`
8. Create a file `tempostack.yaml` containing the TempoStack CR (here is [the CRD](https://github.com/grafana/tempo-operator/blob/main/bundle/community/manifests/tempo.grafana.com_tempostacks.yaml)):
```yaml
apiVersion: tempo.grafana.com/v1alpha1
kind: TempoStack
metadata:
  name: tempo-demo
  namespace: tempo
spec:
  storageSize: 1Gi
  storage:
    secret:
      type: s3
      name: tempostack-dev-minio
  resources:
    total:
      limits:
        memory: "1Gi"
        cpu: "500m"
  template:
    queryFrontend:
      jaegerQuery:
        enabled: true
        ingress:
          type: "ingress"
```
9. Deploy Tempo: `kubectl apply -f tempostack.yaml`
10. Install Istio, configuring it for Tempo: `hack/istio/install-istio-via-istioctl.sh -c kubectl --set values.meshConfig.defaultConfig.tracing.zipkin.address=tempo-tempo-demo-distributor.tempo.svc.cluster.local:9441`
12. Deploy bookinfo demo: `hack/istio/install-bookinfo-demo.sh -c kubectl -tg -wt 2m`
13. In a Kiali CR, you need to have this:
```yaml
 external_services:
   tracing:
     enabled: true
     provider: "tempo"
     use_grpc: true
     in_cluster_url: http://tempo-tempo-demo-query-frontend.tempo.svc.cluster.local:16685
     url: $(minikube ip):8080
```
_note: I don't really know what 'url' should be_

jmazzitelli:
I can see in the Kiali log
```
2023-09-01T21:32:13Z INF Create tempo GRPC client tempo-tempo-demo-query-frontend.tempo.svc.cluster.local:16685
```

I can get to the Jaeger UI from the Kiali left-hand "Distributed Tracing" link. From that Jaeger UI I can see bookinfo traces. So it looks like everything is setup and configured correctly for Minio, Tempo, and Istio (muy bueno!).

However, the Kiali logs show:
```
2023-09-01T22:39:10Z ERR Error fetching availability of the tracing service: rpc error: code = Internal desc = failed to fetch services: plugin error: rpc error: code = Unknown desc = failed GET to tempo Get "http://127.0.0.1:3200/api/search/tag/service.name/values": dial tcp 127.0.0.1:3200: connect: connection refused
```

jmazzitelli:
Just a side note: we should document somewhere if you change the tracing impl that you should also change this, so Kiali can store its own tracing data:

```
   server:
        tracing:
          collector_type: jaeger
          collector_url: http://jaeger-collector.istio-system:14268/api/traces
          enabled: false
          otel:
            protocol: http
```

That `collector_url` needs to change if you ever enable this. I don't know what that URL would be for this setup.

jmazzitelli (changes requested):
I think I finally got the testing procedures figured it out. However, I'm seeing an error which looks like it might be a bug or a misconfiguration. In either case, something is broken that I could not fix using the [kiali.io docs PR](https://github.com/kiali/kiali.io/pull/692/files) as a reference.

See my earlier comment for the error I'm seeing. It's the one about `failed GET to tempo Get "http://127.0.0.1:3200/api/search/tag/service.name/values":
dial tcp 127.0.0.1:3200: connect: connection refused`

josunect:
It looks that with traceQL you can do queries like: 

`{.service.name="productpage.bookinfo" } | select(span.http.method, span.http.target)`

The problem is that the select it is just available from Tempo 2.2 and the latest operator version 0.4 uses Tempo 2.1.

jshaughn:
> The problem is that the select it is just available from Tempo 2.2 and the latest operator version 0.4 uses Tempo 2.1.

Maybe they will release 0.5 if you ask :-)

josunect:
> > The problem is that the select it is just available from Tempo 2.2 and the latest operator version 0.4 uses Tempo 2.1.
> 
> Maybe they will release 0.5 if you ask :-)

Thanks, @jshaughn ! So it is already included for the next version. 

At the meantime, we can use TraceQL and have in just one query with "some" that we need, maybe we don't get all the spans at a time (the dot's size would be the same) 

This is an example with jq: 
`
curl -G -s http://localhost:3200/api/search --data-urlencode 'q={ resource.service.name = "productpage.bookinfo" && status = unset }' | jq`

And the response: 

```
{
  "traces": [
    {
      "traceID": "663eee761cad858128dc5897efd23d4",
      "rootServiceName": "istio-ingressgateway.istio-system",
      "rootTraceName": "productpage.bookinfo.svc.cluster.local:9080/productpage",
      "startTimeUnixNano": "1694095089041245000",
      "durationMs": 12,
      "spanSet": {
        "spans": [
          {
            "spanID": "9871fbb6eb85012f",
            "startTimeUnixNano": "1694095089041540000",
            "durationNanos": "11352000",
            "attributes": [
              {
                "key": "service.name",
                "value": {
                  "stringValue": "productpage.bookinfo"
                }
              },
              {
                "key": "status",
                "value": {
                  "stringValue": "unset"
                }
              }
            ]
          }
        ],
        "matched": 1
      }
    },
    {

```

jshaughn:
When we query today e don't actually get all of the traces and all of the spans, right?  That would be a large query.  I thought maybe we got the traces and that the *number of spans* was maybe provided, and that was how we determined the relative dot size.   I know that I did some work fairly recently that made that chart more efficient, but if I recall it was more about ensuring that we limited concurrent querying on mouse-drag, not necessarily anything to do with the initial query.

josunect:
I've changed the approach to use TraceQL. It is possible to return all the matching traces in just one query, but it does not return all the spans. 

So, in the graph, I've changed the graph elements to have a fixed size. If not, once you hover one element, it will fetch the trace and the element will be resized.

jshaughn:
> So, in the graph, I've changed the graph elements to have a fixed size. If not, once you hover one element, it will fetch the trace and the element will be resized.

That sounds like it should be a more performant approach, have you been able to determine whether there is in fact a significant performance gain?  If there is a gain I think we may be OK with the behavior change, but if, unexpectedly, there is only minimal benefit, we may want to reconsider.

If we do leave it this way, do you think we should resize the node on hover or just leave it alone?

josunect:
> > So, in the graph, I've changed the graph elements to have a fixed size. If not, once you hover one element, it will fetch the trace and the element will be resized.
> 
> That sounds like it should be a more performant approach, have you been able to determine whether there is in fact a significant performance gain? If there is a gain I think we may be OK with the behavior change, but if, unexpectedly, there is only minimal benefit, we may want to reconsider.
> 
> If we do leave it this way, do you think we should resize the node on hover or just leave it alone?

I think it is a big performant approach because, with the approach to create a request per trace, the trace request query time goes from ms to seconds. Also, it is not even working with the operator, because from more than ~= 25 concurrent requests the pod dies (I have also this error https://github.com/kiali/kiali/pull/6518#issuecomment-1703400353) 

Maybe we could investigate to improve the tempo performance when installing with the operator, but I think, doing one request per trace has a poor performance. 

I would investigate how to return all the spans from one request with TraceQL, I wonder if this could be done as an improvement task or if we should wait for it. At the moment I was investigating but I'm not sure how to do it, I've asked in the Tempo slack channel.

josunect:
> > I would investigate how to return all the spans from one request with TraceQL, I wonder if this could be done as an improvement task or if we should wait for it. At the moment I was investigating but I'm not sure how to do it, I've asked in the Tempo slack channel.
> 
> Definitely your current approach looks like a big win over the individual queries. If you do find a way to gather the span-count, or the complete span info like we did in the past, it would be interesting to know how the perf compares.

So, I've found a way to get the span count, but it is not working for the Tempo operator at the moment (As it uses Tempo 2.1 and this looks to be supported from 2.2):

`{ .service.name == "srv" } && { } `

Now it uses a new property `matched` with the number of spans. The only thing is that, in the tooltip, we won't have the heat map until we hover over it, and also the avg duration of the trace will change (As we don't have all the traces)  

The behavior at the moment for Tempo 2.1 is that we have a small dot element in the graph that will be resized when we hover it. Or, if it looks better, we can set the matched value for Tempo to a fixed value when this information in not available.

jshaughn:
> I've found a way to get the span count, but it is not working for the Tempo operator at the moment (As it uses Tempo 2.1 and this looks to be supported from 2.2):

That's great.  We can release first with what we have for Tempo 2.1 and create a new issue to come back and improve things for Tempo 2.2.

> Now it uses a new property matched with the number of spans. The only thing is that, in the tooltip, we won't have the heat map until we hover over it

If I recall correctly The heatmaps are not generated until mouseover happens, and we make a call to the backend for each tooltip we generate.  

> ...and also the avg duration of the trace will change (As we don't have all the traces)

I though we did have all of the traces, and the we just didn't know the number of spans for each trace.  I'm a little confused...

> The behavior at the moment for Tempo 2.1 is that we have a small dot element in the graph that will be resized when we hover it. 

I don't think we need to worry about a resize using tempo 2.1.  That seems a little strange seeing some dots resize as the mouse hovers.  But if it looks good to you then that's fine, keep it. 

>  Or, if it looks better, we can set the matched value for Tempo to a fixed value when this information in not available.

If we don't know the number of spans-per-trace then a fixed dot-size seems to make sense.  With Tempo 2.2 it seems we may be able to go back to the relative sizing, if it is performant.

josunect:
> 
> > I've found a way to get the span count, but it is not working for the Tempo operator at the moment (As it uses Tempo 2.1 and this looks to be supported from 2.2):
> 
> That's great. We can release first with what we have for Tempo 2.1 and create a new issue to come back and improve things for Tempo 2.2.
> 

I've made a change so it works for 2.1 and 2.2, depending on the data that we have. (As it can be tested installing Tempo with Tanka, instead of the operator). 

> > Now it uses a new property matched with the number of spans. The only thing is that, in the tooltip, we won't have the heat map until we hover over it
> 
> If I recall correctly The heatmaps are not generated until mouseover happens, and we make a call to the backend for each tooltip we generate.
> 

I was checking this behavior and they are generated before doing the call for the trace. You can see the heatmap on mouseover, but the backend request to get all the trace info is done on the click event. (I mean, this heatmap in the Tooltip):

image


> > ...and also the avg duration of the trace will change (As we don't have all the traces)
> 
> I though we did have all of the traces, and the we just didn't know the number of spans for each trace. I'm a little confused...
> 

Sorry, I explained it wrong. Your are right, we have all the traces, but the avg is calculated as the average for each span, and we don't have all the spans for the trace. 

> > The behavior at the moment for Tempo 2.1 is that we have a small dot element in the graph that will be resized when we hover it.
> 
> I don't think we need to worry about a resize using tempo 2.1. That seems a little strange seeing some dots resize as the mouse hovers. But if it looks good to you then that's fine, keep it.

I think it looks better without the resizing.

> 
> > Or, if it looks better, we can set the matched value for Tempo to a fixed value when this information in not available.
> 
> If we don't know the number of spans-per-trace then a fixed dot-size seems to make sense. With Tempo 2.2 it seems we may be able to go back to the relative sizing, if it is performant.

This is done now, using the "matched" value for the traces.

jshaughn:
> I was checking this behavior and they are generated before doing the call for the trace. You can see the heatmap on mouseover, but the backend request to get all the trace info is done on the click event. (I mean, this heatmap in the Tooltip):

Ah, OK.  All I remember is that I had to do some sort of tricky work to limit the number of concurrent backend calls when a user would drag the mouse across the chart, because it was launching concurrent backend calls on each dot the mouse was passing over.  We need to ensure that that logic is unchanged.

> I've made a change so it works for 2.1 and 2.2 ... I think it looks better without the resizing

Great work, I don't think I have any more input.  I guess you can proceed with the formal reviews!

josunect:
> > I was checking this behavior and they are generated before doing the call for the trace. You can see the heatmap on mouseover, but the backend request to get all the trace info is done on the click event. (I mean, this heatmap in the Tooltip):
> 
> Ah, OK. All I remember is that I had to do some sort of tricky work to limit the number of concurrent backend calls when a user would drag the mouse across the chart, because it was launching concurrent backend calls on each dot the mouse was passing over. We need to ensure that that logic is unchanged.
> 
> > I've made a change so it works for 2.1 and 2.2 ... I think it looks better without the resizing
> 
> Great work, I don't think I have any more input. I guess you can proceed with the formal reviews!

Thank you, @jshaughn !

josunect:
@jmazzitelli It should be working now, and easier to test with the hack script, in case you want to try. 

Testing procedures: 

- Run the script `hack/istio/tempo/install-tempo-env.sh` (With no params, it will install Tempo, also istio, Kiali and bookinfo)
- Run the hack script:` ./run-kiali.sh -pg 13000:3000 -pp 19090:9090 -app 8080 -es false -iu http://127.0.0.1:15014 -tr tempo-cr-query-frontend -ts tempo-cr-query-frontend -tn tempo -pt 3200:3200`
- Set up Kiali to:
```
  tracing:
    enabled: true
    provider: "tempo"
    in_cluster_url: http://localhost:3200
    url: http://localhost:3200
    use_grpc: false
```

- Access any "Traces" tab, should be there. 
- Click on any trace, the trace info should be there. 
- In the graph page, click in a workload. In the summary panel view, in the traces Tab, there should be traces.

hhovsepy:
Testing locally, a massive work @josunect, still checking it.
What I saw first, when mouse-over on trace, it shows N/A on a root span, then loads the root span only after clicking on a trace, is it optimized by design:
Screenshot from 2023-09-18 19-41-36

josunect:
> Testing locally, a massive work @josunect, still checking it. What I saw first, when mouse-over on trace, it shows N/A on a root span, then loads the root span only after clicking on a trace, is it optimized by design: Screenshot from 2023-09-18 19-41-36

Thank you, @hhovsepy ! 

Yes, this is an optimization, but from previous work. All the trace information is not loaded until you click on that trace. But, let me check the possibility to get the root trace information (I think it should be possible)

josunect:
> Testing locally, a massive work @josunect, still checking it. What I saw first, when mouse-over on trace, it shows N/A on a root span, then loads the root span only after clicking on a trace, is it optimized by design: Screenshot from 2023-09-18 19-41-36

Added the root span. Thanks!

hhovsepy:
@josunect thank you for fixing, one more small comment, not urgent, when clicking on span it loads the info in tooltip details, but it shows only when changing mouse position because the span is also reloaded and changed the location. Can the info be shown when loading is completed without mouse move trigger?
Also wdyt to change 'n/a' message with 'Click for loading' or something similar?

hhovsepy:
The tooltip is fine now:
Screenshot from 2023-09-22 08-48-46

josunect:
I've just uploaded one change to refresh the tooltip onClick. 

image

hhovsepy:
Much better now:
Screenshot from 2023-09-22 10-17-46

And clicking makes data loading without need to mousemove.

hhovsepy (dismissed):
Works fine for me on local setup, regression tested on OCP cluster with Jaeger.

josunect:
Thank you, @hhovsepy ! 

It looks like I still need the review from @jmazzitelli

jmazzitelli:
i will review today.

josunect:
> i will review today.

Thanks, @jmazzitelli ! I've just rebased.

frontend/src/components/JaegerIntegration/JaegerScatter.tsx
```diff
@@ -89,6 +93,21 @@ class JaegerScatterComponent extends React.Component<JaegerScatterProps> {
...
+      var size;
+      if (this.props.provider === TEMPO) {
+        // The query to get the spans in Tempo does not return all of them,
+        // This is to avoid the resize of the circle
+        if (this.props.selectedTrace?.traceID === trace.traceID) {
+          size = this.seletedTraceMatched;
+          trace.matched = this.seletedTraceMatched;
+        } else {
+          size = trace.matched;
+        }
```
jmazzitelli, line 106: Can you explain this in more detail?

I do not understand the setting of `size` here. I see when `provider` is not `TEMPO`, the `size` is going to be set between `MINIMAL_SIZE` and `MAXIMAL_SIZE` (which are values 2 and 30) - that is done in the line below, line 108. Here, however, I do not see any limitation on the value of `size`. How do we know this `size` does not get set to a value less than 2 or greater than 30?

Also, `size` is being set to `trace.matched` or `this.selectedTraceMatched`. I assume these are integers since they are being used to set `size`, but what do those values represent? At first I thought they were a boolean - "yes, it matched, or no it didn't match" - but that doesn't make sense since they are being set to `size`.

tracing/jaeger/model/json/model.go
```diff
@@ -54,6 +54,7 @@ type Trace struct {
 	Spans     []Span                `json:"spans"`
 	Processes map[ProcessID]Process `json:"processes"`
 	Warnings  []string              `json:"warnings"`
+	Matched   int                   `json:"matched"`
```
jmazzitelli, line 57: Perhaps the name of this variable is causing my confusion. "Matched" seems like it would be a boolean: something "matched".. but I see this is of type `int` so I think this is referring to the number of matches (??) ... if that is correct, I would consider changing this to be more meaningful... say, "NumMatched" or "MatchCount" or something like that. This would avoid confusion for people unfamiliar with the code and just seeing the name of the variable as "matched".

tracing/otel/model/converter/converter.go
```diff
@@ -0,0 +1,127 @@
...
+
+		startTime, err := strconv.ParseUint(span.StartTimeUnixNano, 10, 64)
+		if err != nil {
+			log.Errorf("Error converting start time. Skipping trace")
+			continue
+		}
+
+		duration, err := getDuration(span.EndTimeUnixNano, span.StartTimeUnixNano)
+		if err != nil {
+			log.Errorf("Error converting start time. Skipping trace")
```
jmazzitelli, line 36: This error message is identical to the one on line 30. I usually ask that people generate unique error messages everywhere (otherwise you won't know where in the code the error came from - did the error come from this line 36 or line 30?)... even if you just add a period at the end of one but not the other.

However, in this case, I think this is just a copy-n-paste mistake. I think you meant for this to be:
```suggestion
			log.Errorf("Error converting duration. Skipping trace")
```

Either way, this error message should be changed.

tracing/otel/model/converter/converter.go
```diff
@@ -0,0 +1,127 @@
...
+
+func getDuration(end string, start string) (uint64, error) {
+	endInt, err := strconv.ParseUint(end, 10, 64)
+	if err != nil {
+		log.Error("Error converting end date %s", err.Error())
+		return 0, err
+	}
+	startInt, err := strconv.ParseUint(start, 10, 64)
+	if err != nil {
+		log.Error("Error converting end date %s", err.Error())
```
jmazzitelli, line 103: Copy-n-paste typo, and need to use "f" function:

```suggestion
		log.Errorf("Error converting start date: %s", err.Error())
```

tracing/otel/model/converter/converter.go
```diff
@@ -0,0 +1,127 @@
...
+
+	toRet = append(toRet, jaegerSpan)
+
+	return toRet
+}
+
+func getDuration(end string, start string) (uint64, error) {
+	endInt, err := strconv.ParseUint(end, 10, 64)
+	if err != nil {
+		log.Error("Error converting end date %s", err.Error())
```
jmazzitelli, line 98: Since using parameters, need to use the "f" function:

```suggestion
		log.Errorf("Error converting end date: %s", err.Error())
```

tracing/otel/model/types.go
```diff
@@ -0,0 +1,46 @@
...
+	SpanID            string           `json:"spanID"`
+	StartTimeUnixNano string           `json:"startTimeUnixNano"`
+	DurationNanos     string           `json:"durationNanos"`
+	Attributes        []json.Attribute `json:"attributes"`
+	Status            json.Status      `json:"status"`
+}
+
+type SpanSet struct {
+	Spans   []Span `json:"spans"`
+	Matched int    `json:"matched"`
```
jmazzitelli, line 31: Here's that "Matched" variable name again. Should this be "Matches" or "NumMatched" or something else?

tracing/tempo/http_client.go
```diff
@@ -0,0 +1,261 @@
...
+
+	q.Set("q", queryQL)
+	if query.MinDuration > 0 {
+		q.Set("minDuration", fmt.Sprintf("%dms", query.MinDuration.Milliseconds()))
+	}
+	if query.Limit > 0 {
+		q.Set("limit", strconv.Itoa(query.Limit))
+	}
+	u.RawQuery = q.Encode()
+	log.Debugf("Prepared Tempo API query: %v", u)
```
jmazzitelli, line 209: How noisy will this be? Should be consider only logging this at Trace level?
```suggestion
	log.Tracef("Prepared Tempo API query: %v", u)
```

jmazzitelli (changes requested):
Some of these are just questions based on confusion, others are requested changes.

As for the  "matched" variable name - I may be confused, so if it makes sense to leave it as "matched", that's OK. But see if I am making sense - should it be renamed it to something else?

frontend/src/components/JaegerIntegration/JaegerScatter.tsx
josunect: The Tempo API does not return all the matched spans. But, in Tempo 2.2 it returns the `matched` field with the number of matched, like: 

```
      "spanSet": {
        "spans": [
          {
            "spanID": "2fc76fe3ac9a9ecb",
            "startTimeUnixNano": "1695630141086172000",
            "durationNanos": "782000",
            "attributes": [
              {
                "key": "service.name",
                "value": {
                  "stringValue": "details.bookinfo"
                }
              },
              {
                "key": "node_id",
                "value": {
                  "stringValue": "sidecar~10.244.1.16~details-v1-5f4d584748-2stb6.bookinfo~bookinfo.svc.cluster.local"
                }
              },
              {
                "key": "status",
                "value": {
                  "stringValue": "unset"
                }
              }
            ]
          },
          {
            "spanID": "74eeeb50a8e23208",
            "startTimeUnixNano": "1695630141083075000",
            "durationNanos": "13459000",
            "attributes": [
              {
                "key": "status",
                "value": {
                  "stringValue": "unset"
                }
              },
              {
                "key": "node_id",
                "value": {
                  "stringValue": "sidecar~10.244.2.11~productpage-v1-564d4686f-886dn.bookinfo~bookinfo.svc.cluster.local"
                }
              }
            ]
          },
          {
            "spanID": "0a09881263ee9c76",
            "startTimeUnixNano": "1695630141090410000",
            "durationNanos": "4449000",
            "attributes": [
              {
                "key": "node_id",
                "value": {
                  "stringValue": "sidecar~10.244.2.11~productpage-v1-564d4686f-886dn.bookinfo~bookinfo.svc.cluster.local"
                }
              },
              {
                "key": "status",
                "value": {
                  "stringValue": "unset"
                }
              }
            ]
          }
        ],
        "matched": 8
   }
```

So I named the same as this parameter. If it is confusing I can change it, but, not sure if it can introduce other confusions about where this value is coming from (Or we can also add a comment here to clarify) 

But you are right, I missed the MIN and MAX logic, I will add it here.

tracing/tempo/http_client.go
josunect: I see that this is the same level as the Jaeger prepare query, https://github.com/josunect/kiali/blob/master/jaeger/http_client.go#L108
I thought it would make sense to have the same level in both?

tracing/jaeger/model/json/model.go
josunect: See previous comment

josunect:
Thanks for the review, @jmazzitelli. I have committed some changes requested, and added a couple of comments.

jmazzitelli:
> and added a couple of comments.

OK, I understand about "matched". I don't see any new comments in the latest commits. I could like "matched" commented in the few places it is declared, so it is clear what this is.

josunect:
> > and added a couple of comments.
> 
> OK, I understand about "matched". I don't see any new comments in the latest commits. I could like "matched" commented in the few places it is declared, so it is clear what this is.

Sure! I will add the comments there

jshaughn (commented):
Just a side comment to make sure that the additional `queryScope` tags are handled in this impl.  Probably it works fine, it just crossed my mind as something to check.

josunect:
> Just a side comment to make sure that the additional `queryScope` tags are handled in this impl. Probably it works fine, it just crossed my mind as something to check.

I didn't realized about it. Thanks for catching this!  I will include it.

josunect:
> Just a side comment to make sure that the additional `queryScope` tags are handled in this impl. Probably it works fine, it just crossed my mind as something to check.

Added the queryScope logic.

jmazzitelli (approved):
i have no more suggestions... lgtm

Issue #6523, opened 2023-08-25
Improve check for detection of workload in Ambient Mesh


josunect:
It was reported that Kiali was marking as "Out of Mesh" workloads included in an Ambient Namespace. 

Tested with Rollouts, Deployments and DaemonSets from 1 to N replicas. On an EKS cluster. And Istio 1.19.0-beta.0, 1.18

Maybe it would be enough to just check the namespace label? But for sidecars, there are more than one check, it actually checks for the existence of the sidecar. What happen if the namespace is annotated but for some reason the workload was not redirecting the traffic? I guess it would be useful to have some way to detect and alert for this situation. 

[istio.io/dataplane-mode=ambient](http://istio.io/dataplane-mode=ambient) -> namespace label, will include or exclude a namespace.
[ambient.istio.io/redirection=enabled](http://ambient.istio.io/redirection=enabled) = pod label, indicates the pod "is captured" - adding a namespace label instructs the node agent to capture all pods in the labeled namespace, and label each "captured" pod with this.
Unlabeleing the namespace reverses this process, and (after the agent processes/uninjects a given pod) unlabels the pods.
[ambient.istio.io/redirection=disabled](http://ambient.istio.io/redirection=disabled)  = can be manually applied to a pod to opt it out

Ref. https://github.com/kiali/kiali/issues/5476

josunect:
That might be the reason: https://github.com/istio/istio/blob/master/pilot/pkg/serviceregistry/kube/controller/ambientindex_external.go#L409 

`For VMs we use Labels instead of an Annotations since we don't have access to the top level WorkloadEntry object.`

jshaughn:
One thing I notice in the current [Getting Started With Ambient Mesh](https://istio.io/latest/docs/ops/ambient/getting-started/#addtoambient) Istio docs is that they only talk about the namespace labeling.  There is no discussion about how to label/annotate individual pods to somehow override the namespace label for a particular pod.  So, maybe for now Kiali should just defer to the namespace label and assume all pods in the namespace are in-mesh if there is no label/annotation?

josunect:
[...]

PR #6528, opened 2023-08-28
Check for Ambient labels at workload level

josunect:
Fixes https://github.com/kiali/kiali/issues/6523


josunect:
This is not going to fix the issue as the labels are in the workload entry, not in the workload.

PR #6532, opened 2023-08-29
Change login func to not send Auth header when token is present

rohitsakala:
** Describe the change **

When a token is present, we send it through params and when a username and password is present, we send it through Auth header. 

There is no need to send Auth header when the token is not empty.

** Issue reference **

https://github.com/kiali/kiali/issues/6533

### DIFF
--- a/frontend/src/services/Api.ts
+++ b/frontend/src/services/Api.ts
@@ -122,13 +122,18 @@ export const login = async (
 
   const proxyUrl = getAPIProxy();
 
-  return axios({
+  const axiosRequest = {
     method: HTTP_VERBS.POST,
     url: proxyUrl ? `${proxyUrl}/${urls.authenticate}` : urls.authenticate,
     headers: getHeaders(proxyUrl),
-    auth: basicAuth(request.username, request.password),
     data: params
-  });
+  };
+
+  if (request.username !== '' || request.password !== '') {
+    axiosRequest['auth'] = basicAuth(request.username, request.password);
+  }
+
+  return axios(axiosRequest);
 };
 
 export const logout = () => {


jmazzitelli:
We would need to see replication procedures and an illustration of the problem you are trying to fix. Changing auth code in Kiali is very tricky and fragile because it touches many different use cases and auth strategies (such as openid, openshift, and token strategies). I have a feeling this may break something, particularly on the openid side of things, but I am not sure.

> There is no need to send Auth header when the token is not empty.

This I am not sure about. The CI tests are running - let's see what they say. We will also want to test this for other auth strategies like openid and openshift in addition to token.

rohitsakala:
@jmazzitelli The tests are all passing. To be less disruptive, can I change the PR to fix the exact bug like this 

```
export const login = async (
  request: LoginRequest = { username: ANONYMOUS_USER, password: 'anonymous', token: '' }
): Promise<Response<LoginSession>> => {
   const params = new URLSearchParams();
   params.append('token', request.token);
  
   const proxyUrl = getAPIProxy();

  const axiosRequest = {
    method: HTTP_VERBS.POST,
    url: proxyUrl ? `${proxyUrl}/${urls.authenticate}` : urls.authenticate,
    headers: getHeaders(proxyUrl),
    data: params,
  }

  if (request.username != '') || (request.password != '')  {
    axiosRequest['auth'] = basicAuth(request.username, request.password);
  }

  return axios(axiosRequest);
};


```

rohitsakala:
@jmazzitelli I have updated the codebase to be less disruptive and fixes the exact bug. Can you please review this ? Thanks

frontend/src/services/Api.ts
```diff
@@ -122,13 +122,18 @@ export const login = async (
 
   const proxyUrl = getAPIProxy();
 
-  return axios({
+  const axiosRequest = {
     method: HTTP_VERBS.POST,
     url: proxyUrl ? `${proxyUrl}/${urls.authenticate}` : urls.authenticate,
     headers: getHeaders(proxyUrl),
-    auth: basicAuth(request.username, request.password),
     data: params
```
jshaughn, line 129: Should we also do something similar for `token` and don't append `params` if token is unset?

jshaughn: I'm fine minimizing the change.  Just noticed this.

rohitsakala: @jshaughn @jmazzitelli Yes I agree to minimize the change as it would be less disruptive.

jmazzitelli: I just doing some manual testing right now. But on the face of it, it looks good. I'll be able to approve today some time if all looks good.

jmazzitelli (approved):
All CI testing looks good. I also tested with a basic openid setup and things still work.

jshaughn (approved):
This looks safe to me.  I can't see any good reason to set this header to a silly value when using token auth.

PR #6536, opened 2023-08-30
Refactor PFColors

aljesusg:
Refactor PFColors. Only replace colors in one site

### DIFF
--- a/frontend/src/components/Pf/PfColors.tsx
+++ b/frontend/src/components/Pf/PfColors.tsx
@@ -21,6 +21,7 @@ export enum PFColors {
   Black800 = 'var(--pf-v5-global--palette--black-800)',
   Black900 = 'var(--pf-v5-global--palette--black-900)',
   Black1000 = 'var(--pf-v5-global--palette--black-1000)',
+  Blue50 = 'var(--pf-v5-global--palette--blue-50)',
   Blue200 = 'var(--pf-v5-global--palette--blue-200)',
   Blue300 = 'var(--pf-v5-global--palette--blue-300)',
   Blue400 = 'var(--pf-v5-global--palette--blue-400)',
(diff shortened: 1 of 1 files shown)


frontend/src/components/Pf/PfColors.tsx
```diff
@@ -127,43 +131,54 @@ export type PFColorValues = {
 
 export let PFColorVals: PFColorValues;
 
+/*
+  Subtract color from var
```
jshaughn, line 135:
```suggestion
  Extract color from var
```

aljesusg: Shit I marked this like resolved I'll do the change thanks

ferhoyos (dismissed):
LGTM, I like the idea to have only one source of truth in PFColors. 

@aljesusg just a suggestion if you can remove `BackgroundColor300` variable because it is not used in Kiali and this color is only applicable in dark mode:

From PF light theme palette:
image

aljesusg:
> LGTM, I like the idea to have only one source of truth in PFColors.
> 
> @aljesusg just a suggestion if you can remove `BackgroundColor300` variable because it is not used in Kiali and this color is only applicable in dark mode:
> 
> From PF light theme palette: image

Let keep it because I did this PR to keep things clear and aligned with backstage and i am going to use that color

ferhoyos:
Then it is ok if you are going to use it in backstage. Just take into account that this color does not have any effect in light theme ;-)

jshaughn:
> i am going to use that color...

If there is a color you want to keep bit it's not used in the Kiali code base (and won't be found in a search) then I suggest a comment, otherwise someone (like me) may remove it.

ferhoyos (approved):
LGTM

PR #6542, opened 2023-08-30
[WIP] Fix for double tap on a node in MC redirecting to wrong cluster

leandroberetta:
Fixes https://github.com/kiali/kiali/issues/6548

### DIFF
--- a/frontend/src/components/Nav/NavUtils.tsx
+++ b/frontend/src/components/Nav/NavUtils.tsx
@@ -37,7 +37,7 @@ const buildCommonQueryParams = (params: GraphUrlParams): string => {
   q += `&${URLParam.REFRESH_INTERVAL}=${params.refreshInterval}`;
 
   if (params.node && params.node.cluster && isMultiCluster) {
-    q += `&${URLParam.CLUSTERNAME}=${params.node?.cluster}`;
+    q += `&${URLParam.CLUSTERNAME}=${encodeURIComponent(params.node?.cluster)}`;
   }
 
   return q;
--- a/frontend/src/components/Nav/NavUtils.tsx
+++ b/frontend/src/components/Nav/NavUtils.tsx
@@ -34,6 +34,11 @@ const buildCommonQueryParams = (params: GraphUrlParams): string => {
   q += `&${URLParam.DURATION}=${params.duration}`;
   q += `&${URLParam.GRAPH_OPERATION_NODES}=${params.showOperationNodes}`;
   q += `&${URLParam.REFRESH_INTERVAL}=${params.refreshInterval}`;
+
+  if (params.node && params.node.cluster) {
+    q += `&${URLParam.CLUSTERNAME}=${params.node?.cluster}`;
+  }
+
   return q;
 };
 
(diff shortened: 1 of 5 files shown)


frontend/src/components/Nav/NavUtils.tsx
```diff
@@ -34,6 +34,11 @@ const buildCommonQueryParams = (params: GraphUrlParams): string => {
   q += `&${URLParam.DURATION}=${params.duration}`;
   q += `&${URLParam.GRAPH_OPERATION_NODES}=${params.showOperationNodes}`;
   q += `&${URLParam.REFRESH_INTERVAL}=${params.refreshInterval}`;
+
+  if (params.node && params.node.cluster) {
+    q += `&${URLParam.CLUSTERNAME}=${params.node?.cluster}`;
+  }
+
```
jshaughn, line 41: I wonder if we should add it only if it's not the default/home cluster?

leandroberetta: Yeah, and also we should check if we are on multicluster too.

jshaughn: I guess another option would be to stop using a query param and accept the fact that `cluster` is part of the node definition.  And so just change the paths to include cluster.  For example:

From:
path: '/graph/node/namespaces/:namespace/' + Paths.WORKLOADS + '/:workload',

To:
path: '/graph/node/clusters/:cluster/namespaces/:namespace/' + Paths.WORKLOADS + '/:workload',

frontend/src/pages/Graph/GraphPage.tsx
```diff
@@ -281,6 +281,7 @@ class GraphPageComponent extends React.Component<GraphPageProps, GraphPageState>
       aggregate: aggregate!,
       aggregateValue: aggregateValue!,
       app: app!,
+      cluster: getClusterName(),
```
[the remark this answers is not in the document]
leandroberetta: Yeah, I'm just focusing on one graph for now, but true.

frontend/src/components/Nav/NavUtils.tsx
leandroberetta: Maybe, but will require several changes into Kiali. For example, in list pages, today we are querying once and fetching all from all clusters, with these new approach, we should fetch and aggregate on the UI.

jshaughn: I think I was not clear.  I was not suggesting we add `:cluster` everywhere, just when we are talking about a specific object, in this case the node for a node graph.  A specific node is basically defined as `cluster:namespace:name`.  And I am only talking about the UI routes at the moment.

In our current approach we are trying to manage the `cluster` portion as a query param, and the namespace and name in the path.

jshaughn: Although it may be more clear, it's probably not worth the effort at this time, as the parameter approach is working.

frontend/src/components/Nav/NavUtils.tsx
```diff
@@ -34,6 +35,11 @@ const buildCommonQueryParams = (params: GraphUrlParams): string => {
   q += `&${URLParam.DURATION}=${params.duration}`;
   q += `&${URLParam.GRAPH_OPERATION_NODES}=${params.showOperationNodes}`;
   q += `&${URLParam.REFRESH_INTERVAL}=${params.refreshInterval}`;
+
+  if (params.node && params.node.cluster && isMultiCluster) {
+    q += `&${URLParam.CLUSTERNAME}=${params.node?.cluster}`;
```
jshaughn, line 40: Do we need to `encodeURIComponent` the cluster name?

jshaughn (approved):
@leandroberetta looks good, I'll approve but see my comment in case you think we need to make that change...

jshaughn (approved):
LG!

Issue #6545, opened 2023-08-31
[make] work around opm render bug when building for OLM


jmazzitelli:
There is a bug in `opm render` ... see https://github.com/operator-framework/operator-registry/issues/935 and https://github.com/operator-framework/operator-registry/issues/919 ... both myself and @ferhoyos hit this when running `make  olm-operator-create` (really, the problem is in the target `build-olm-index`).

There is an interesting workaround as you see [here](https://github.com/kubev2v/forklift/pull/113/files#diff-48b8da821505014013b98f59eabdbf034dee562ad32be73f055e2d3a40327668R152-R171) which seems based on this simple workaround [here](https://github.com/operator-framework/operator-registry/issues/919#issuecomment-1382675497).

I think we might be able to do something simpler, but the idea is we should tell `opm render` what its DOCKER_CONFIG is (don't rely on the default environment).

I think if DORP=podman (meaning you are relying on podman for the image registry access), then the Makefile should simply create a softlink to the podman auth.json to a file called `config.json` (which is docker's name, and apparently what opm render looks at) and just point opm to the podman directory. So something as simple as this should work:

```
ln $XDG_RUNTIME_DIR/containers/auth.json $XDG_RUNTIME_DIR/containers/config.json
export DOCKER_CONFIG="$XDG_RUNTIME_DIR/containers"
opm render ...
unset DOCKER_CONFIG
```

Alternatively, copy the auth.json to the build output directory and point DOCKER_CONFIG there.

PR #6546, opened 2023-08-31
workaround opm render bug by providing custom docker_config location

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6545

We will want to do a similar fix for the OSSMC make target that is similar: https://github.com/kiali/openshift-servicemesh-plugin/blob/main/make/Makefile.olm.mk#L69 (_update_: this was done in this PR: https://github.com/kiali/openshift-servicemesh-plugin/pull/201)


ferhoyos (approved):
LGTM. Tested locally and works like a charm. No need to delete docker config.json for workaround.

Issue #6549, opened 2023-08-31
Multicluster - Negative testing against single cluster


matejnesuta:
Tests that there are no unnecessary badges, columns, and other multi-cluster related elements visible in single cluster setup.

PR #6557, opened 2023-09-01
Multi-cluster - negative testing against single cluster setup

matejnesuta:
Related to #6549. 

This PR is adding additional tests and checks to ensure Multi-cluster related features are not causing problems for the single cluster setup. This includes checking that no clusterName parameter is used in links, no cluster badges and columns are visible and also no cluster dropdown should be available in the Istio config wizard.


hhovsepy:
code lgtm, testing it on OCP cluster

matejnesuta:
Passed on OCP. Merging.

PR #6558, opened 2023-09-01
Fix some tour stops in PF graph

leandroberetta:
Resolves https://github.com/kiali/openshift-servicemesh-plugin/issues/202

This PR fixes two stops that weren't working in PF graph (Layout and Legend) and also updates a little bit some wording in the Graph and ContextualMenu stops.


jshaughn (approved):
I didn't test it but I can see that we may need some different wording in a few places, where the behavior differs.  Looks Good!

PR #6560, opened 2023-09-04
Update victory library to version 36.6.11

### LINKED ISSUE #6559: Align victory library with PF5
Currently Kiali uses `victory` library with fixed version `36.3.0` while PF5 uses version `^36.6.11`. Since `victory`library is used only to support PF5 charts, I think this library should be aligned with PF5 to avoid inconsistencies and reduce application size (both versions are loaded at runtime).

ferhoyos:
** Describe the change **

`victory`library has been updated to version 36.6.11 to be aligned with PF5.

After update, application file size has been reduced in 58 kB:

image

** Issue reference **

Closes #6559

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -99,7 +99,7 @@
     "tippy.js": "3.4.1",
     "typesafe-actions": "^4.2.1",
     "typestyle": "^2.4.0",
-    "victory": "36.3.0",
+    "victory": "^36.6.11",
     "visibilityjs": "2.0.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


aljesusg:
PF5 use 36.6.11 (https://github.com/kiali/kiali/blob/00196413f5c542a8d4ffbe2fbcd5e049c7f9c072/frontend/yarn.lock#L2679)  note that the library is used in another charts. please add some screenshots with these charts to check that they render correctly:

- Chartwithlegend
- Container of Chart

ferhoyos:
@aljesusg here is a screenshot of ChartWithLegend example and its container that uses `victory`library directly:

image

aljesusg (approved):
LGFM

PR #6568, opened 2023-09-05
Update react-scripts dependencies

### LINKED ISSUE #6563: Build fails with Node 18
Kiali build fails with Node 18 and this is the console error:

image

According to https://sebhastian.com/error-0308010c-digital-envelope-routines-unsupported/, the issue is related with Webpack that uses MD4 hash algorithm but this is not supported in OpenSSL 3 (used by default from Node 17).

ferhoyos:
** Describe the change **

To support Node 18, I have forced the update of react-scripts dependencies to get latest versions of webpack and rest of build libraries (babel, postcss, eslint, etc.).

To force `react-scripts` dependencies update:

1. Remove `react-scripts`line from package.json
2. Execute `yarn install` to remove `react-scripts`dependencies from yarn.lock
3. Add `react-scripts`line again to package.json
4. Execute `yarn install` again to add `react-scripts`dependdencies to yarn.lock with latest versions

Other actions:

- Since `postcss` is updated the fixed version in resolutions section can be removed
- Align `@types/node` to minimum support node version (14.17.0)
- Add some `@babel` libraries to dev dependencies to remove some warning messages:

**@babel/plugin-proposal-private-property-in-object:**

image

**@babel/core:**

image

**@babel/plugin-syntax-flow`and `@babel/plugin-transform-react-jsx:**

image

 Tested successfully with Node versions 14, 16 and 18

** Issue reference **

Fixes #6563.


aljesusg (changes requested):
I vote NO to add babel to the project. IMO We not need it 

Me proposal is force to use node 16 and not node 18 for now. This is not an issue that we need to address right?

aljesusg:
> Babel library is already added by react-scripts: image
> 
> What I have done is just define them on package.json to remove warning messages (I can remove them, but same babel libraries are going to be downloaded anyway by react-script dependencies).
> 
> About `react-scripts`, this library is not maintained anymore, and I think the dependencies update have to be done sooner or later to support Node 18 and fix frontend vulnerabilities. I know that this PR can impact other areas and more testing is needed before we merge it, so I am ok to postpone this change if it is not the correct moment to do it (and force to use node 16 instead).

We should try to move to nodejs18 if we are going to lose the support but this is going to require more changes and we need to test so things here. 

About babel I wonder if add them like resolutions fix the problem instead devDependencies.

ferhoyos:
Set babel libraries in `resolutions`section does the trick and warning messages are removed too. Thanks @aljesusg!

ferhoyos:
Tested with Node 20 (v20.6.0), it works fine.

aljesusg:
Tested downstream with nodejs 18 and the actual master works with commit 260ae195ac63556416fe5b2a1bf083dcbb37c72a 

Yarn 1.22.10
Node 18.16.1
Npm 9.5.1

I need to test this locally

aljesusg:
> Tested downstream with nodejs 18 and the actual master works with commit [260ae19](https://github.com/kiali/kiali/commit/260ae195ac63556416fe5b2a1bf083dcbb37c72a)
> 
> Yarn 1.22.10 Node 18.16.1 Npm 9.5.1
> 
> I need to test this locally

Read https://github.com/kiali/kiali/issues/6563#issuecomment-1706672534

Work adding this option, it seems that is added in  the build container in downstream so we don't need to make changes in libraries @ferhoyos

ferhoyos:
Closing since issue #6563 has been fixed by #6582

ferhoyos:
Re-open PR to get latest versions of react-scripts dependencies and make Kiali more robust and secure.

jshaughn (approved):
From what I can tell nothing is blocking this change, approving but also would like to wait for approval from @aljesusg .

aljesusg (approved):

aljesusg:
We need to backport this to 1.73 to remove the openssl flag to make the build

PR #6582, opened 2023-09-06
Add openssl-legacy-provider flag to support Node 18

### LINKED ISSUE #6563: Build fails with Node 18

ferhoyos:
** Describe the change **

Following suggestion done by @aljesusg in https://github.com/kiali/kiali/issues/6563#issuecomment-1706672534, openssl-legacy-provider flag has been added to support Node 18. Since this flag is not recognized by Node < 17, conditional script must be done in package.json (add the flag only if Node version >= 17). For this reason `if-node-version`library has been added here (https://github.com/mysticatea/if-node-version).

Tested with Node 14, 16, 18 and 20.

** Issue reference **

Fixes #6563

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -34,10 +34,11 @@
     "Some build environments we use do not include npm access, and installing yarn is not possible."
   ],
   "scripts": {
+    "check:node": "if-node-version '>= 17' && npm run $npm_config_command --param=--openssl-legacy-provider || npm run $npm_config_command",
     "build": "if [ \"${KIALI_ENV}\" = \"production\" ]; then npm run build:prod; else npm run build:dev; fi",
-    "build:dev": "sh -ac '. ./.env.upstream; npm run lint && npm run build:kiali'",
-    "build:kiali": "CI=false REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) GENERATE_SOURCEMAP=false EXTEND_ESLINT=true react-scripts build --profile",
-    "build:prod": "sh -ac '. ./.env.downstream; npm run build:kiali'",
+    "build:dev": "sh -ac '. ./.env.upstream; npm run lint && npm run check:node --command=build:kiali'",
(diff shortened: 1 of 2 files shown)


aljesusg:
Thabks for this @ferhoyos in fact we are going to move to nodejs18 so please can you make the migration?

Set engine to >=18
Change scripts in workflow from 16 to 18

I tested the build with nodejs18

aljesusg:
Thanks @ferhoyos but if we move to nodejs18 we don't need to check the version right? The flag is available

frontend/package.json
aljesusg, line 1: we must change https://github.com/kiali/kiali/blob/553415e87840c091d289dfb8be23c2cd07990115/frontend/package.json#L151

```yaml
  "engines": {
    "node": ">= 18.0.0",
```

frontend/package.json
```diff
@@ -125,6 +127,7 @@
     "enzyme": "3.11.0",
     "enzyme-to-json": "3.4.4",
     "husky": "1.3.1",
+    "if-node-version": "^1.1.1",
     "jest-canvas-mock": "2.2.0",
```
aljesusg, line 131: We don't need if-node-version, we are goign to force to use 18 or greater and flag is supported.

Redo scripts-tasks

aljesusg (changes requested):

frontend/package.json
```diff
@@ -145,9 +145,8 @@
     "postcss": "^8.4.12"
   },
   "engines": {
-    "node": "^14.17.0 || >= 16.0.0",
-    "npm": ">=6.0.0 ",
```
ferhoyos, line 149: Node 18.0.0 already comes with NPM 8.6.0 installed, so this condition is not needed anymore (fulfilled when you install node 18 or greater version).

aljesusg (approved):
LGFM, great job

Issue #6512, opened 2023-08-23
IstioType in istioconfigList is propagated to other views


aljesusg:
## Behaviour

This filter shoudn't be applied in other views.
Filter related with the resource should not applied, examples:

- istio config name is not applied in other views
- workload type neither 
- ....

## Reproduce

1- Filter **istioConfig list** list by some `istio config type`
2- Move to workloads list, this filter is applied.

## Expected behaviour

Not filter workloads by the istioconfigtype selected in istioconfiglist. It makes no sense, there are not worklaods with attribute istio config type.

jshaughn:
@aljesusg I was going to start on this but now I'm not sure I agree we have to make a change.  It's because the Details columns can show Istio Config that is related to a Workload or Service or App.  A Workload for example can have an Authorization Policy.  And so each of the list views does support the `Istio Type` filter.  Because the mechanism just carries over filters that are common the Istio Type filter will be applied as you travel around.  I agree that it's not the greatest behavior, because an Istio Config type that can be applied to a workload may not make sense for a Service.

If we want to prevent this filter from carrying over from the Istio Config list view to the other list views I think we could change the filter name from "Istio Type" to just "Type".  And maybe for the other views change it from "Istio Type" to "Istio Config Type", just to make it more clear.  What do you think?

PR #6583, opened 2023-09-06
Prevent "istio type" filter carryover

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6512

Prevent "istio type" filter carryover between config list page and other list pages.

- Change "Istio Type" filter to be "Type" for Istio Config list page, the context is implied.
- Change "Istio Type" filter to be "Istio Config Type" for other list pages. the filter should be more explicit.
- The change in filter naming prevents the previous "Istio Type" filter carryover

### DIFF
--- a/frontend/cypress/integration/common/istio_config.ts
+++ b/frontend/cypress/integration/common/istio_config.ts
@@ -217,12 +217,9 @@ Given(
   }
 );
 
-Given(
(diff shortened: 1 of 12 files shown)


aljesusg (approved):
LGFM

Screenshot from 2023-09-07 08-50-56

tested and not carryover

PR #6584, opened 2023-09-07
Remove PF direct dependencies from package.json

### LINKED ISSUE #6564: Align d3 library with PF5
Currently Kiali uses `d3` library with version ^5.16.0 while PF5 react-topology library uses version ^7.8.0. Since `d3` is used only to support PF5 topology component, I think this library should be aligned with PF5 to avoid inconsistencies and reduce application size (both versions are loaded at runtime).

ferhoyos:
** Describe the change **

There are some dependencies in package.json that are direct dependencies of PF libraries and tightly related (only used to support PF components):
- d3
- victory
- @patternfly/react-tokens (this one is not used directly by Kiali)

All these libraries can be removed from package.json since they are direct dependencies of PF and downloaded with PF libraries. Then we do not have to update version of such libraries when we update PF. 

** Issue reference **

Closes #6564

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -69,7 +69,6 @@
     "cytoscape": "3.15.5",
     "cytoscape-canvas": "3.0.1",
     "cytoscape-popper": "1.0.7",
-    "d3": "^5.16.0",
     "dagre": "0.8.5",
     "deep-freeze": "0.0.1",
     "eventemitter3": "4.0.7",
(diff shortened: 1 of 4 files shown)


ferhoyos:
Screenshots of PF react-tolopogy and PF react-charts to show that they work fine with direct dependency removal:

image

image

jshaughn (approved):
Interesting, I'm not sure why I ended up adding `d3` to package.json because the import I added didn't change at all.  Anyway, excellent change, it will help us avoid package mismatch, and reduces our direct deps from a CVE perspective.

Tested the build and ensured the d3 feature worked as expected in PFT graph.

Issue #6429, opened 2023-07-31
GW API Multiple implementations


hhovsepy:
There can be multiple implementations of K8s Gateway API running on an OCP cluster.
https://gateway-api.sigs.k8s.io/implementations/

- [ ] Kiali at least should be able to run correctly.

- [ ] Next step to read-only support of objects.

PR #6585, opened 2023-09-07
Gateway API class selector in K8s Gateway Istio config wizard

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6429

When there are more than 2 Gateway classes configured in Kiali CR
```
        gateway_api_classes:
          - name: Istio
            class_name: istio
          - name: Contour
            class_name: contour
```

In both wizards, Kiali Service -> Create K8s Gateway API Routing and Istio Config -> Create K8s Gateway, it appears dropdown to select the Gateway API implementation:

Screenshot from 2023-09-07 18-37-57
Screenshot from 2023-09-08 19-36-53

### DIFF
--- a/frontend/src/components/DebugInformation/DebugInformation.tsx
+++ b/frontend/src/components/DebugInformation/DebugInformation.tsx
@@ -64,6 +64,7 @@ const propsToShow = [
   'authStrategy',
   'clusters',
   'gatewayAPIEnabled',
+  'gatewayAPIClasses',
   'istioConfigMap',
   'istioIdentityDomain',
   'istioNamespace',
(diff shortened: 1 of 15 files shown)


frontend/src/components/DebugInformation/DebugInformation.tsx
```diff
@@ -64,6 +64,7 @@ const propsToShow = [
   'authStrategy',
   'clusters',
   'gatewayAPIEnabled',
+  'gatewayAPIClasses',
```
jshaughn, line 67: trivial - alphabetized out of order in a few places (not a blocker, can fix in some other PR)

frontend/src/components/IstioWizards/K8sGatewaySelector.tsx
```diff
@@ -174,6 +186,20 @@ export class K8sGatewaySelector extends React.Component<Props, K8sGatewaySelecto
             )}
             {this.state.newGateway && (
               <>
+                {serverConfig.gatewayAPIClasses.length > 1 && (
```
jshaughn, line 189: I guess another option would be to always display this so that the user can see the set value.  But this approach is fine as well, not a blocker.

jshaughn (approved):
Looks good to me!

frontend/src/components/IstioWizards/K8sGatewaySelector.tsx
hhovsepy: I made the Gateway Classes available in Debug info.

PR #6586, opened 2023-09-07
[hack] fix install-bookinfo-demo and other hack scripts for SCC priority

jmazzitelli:
https://docs.openshift.com/container-platform/4.13/authentication/managing-security-context-constraints.html#scc-prioritization_configuring-internal-oauth

The SCC created by the script didn't have a priority, so it defaults to 0, which is the same as all the other restrictive sccs that come out of box. So...

```
If the priorities are equal, the SCCs are sorted from most restrictive to least restrictive.
```

### DIFF
--- a/hack/istio/create-mesh-namespace.sh
+++ b/hack/istio/create-mesh-namespace.sh
@@ -154,6 +154,7 @@ seLinuxContext:
   type: RunAsAny
 supplementalGroups:
   type: RunAsAny
+priority: 9
 users:${USERS}
 SCC
 fi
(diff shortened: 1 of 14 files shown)


jshaughn (commented):
This change fixed the issue but I think this PR needs to update all of our hack scripts, no?

jmazzitelli:
> This change fixed the issue but I think this PR needs to update all of our hack scripts, no?

Yeah, I can do that.

jmazzitelli:
This priority field was supported at least as far back as OpenShift 4.8 so this should be good for any OpenShift version we care about: https://docs.openshift.com/container-platform/4.8/authentication/managing-security-context-constraints.html#scc-prioritization_configuring-internal-oauth

jshaughn (approved):
Tested against bookinfo and it workee!  Assuming others will benefit in the same way.

Issue #6587, opened 2023-09-07
Upgrade to minor patternfly 5.0.4


aljesusg:
Parent https://github.com/kiali/kiali/issues/6438

PR #6588, opened 2023-09-07
Upgrade patternfly to 5.0.4

aljesusg:
Fix https://github.com/kiali/kiali/issues/6587

Remove CI=false
Fix some issues relted with patternfly 5.0.2

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -36,7 +36,7 @@
   "scripts": {
     "build": "if [ \"${KIALI_ENV}\" = \"production\" ]; then npm run build:prod; else npm run build:dev; fi",
     "build:dev": "sh -ac '. ./.env.upstream; npm run lint && npm run build:kiali'",
-    "build:kiali": "CI=false REACT_APP_VERSION=$npm_package_version REACT_APP_NAME=$npm_package_name REACT_APP_GIT_HASH=$(git rev-parse HEAD) GENERATE_SOURCEMAP=false EXTEND_ESLINT=true react-scripts --openssl-legacy-provider build --profile",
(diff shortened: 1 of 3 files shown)


jshaughn (changes requested):
It builds and runs for me locally but it won't build in CI because it treats warnings like errors:

```
Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.

Failed to compile.

Warning
(1[50](https://github.com/kiali/kiali/actions/runs/6113553385/job/16594175197?pr=6588#step:4:51):3) autoprefixer: start value has mixed support, consider using flex-start instead
```

Looks like we need to either get rid of the warning or ignore it.

aljesusg:
> It builds and runs for me locally but it won't build in CI because it treats warnings like errors:
> 
> 
> 
> ```
> 
> Treating warnings as errors because process.env.CI = true.
> 
> Most CI servers set it automatically.
> 
> 
> 
> Failed to compile.
> 
> 
> 
> Warning
> 
> (1[50](https://github.com/kiali/kiali/actions/runs/6113553385/job/16594175197?pr=6588#step:4:51):3) autoprefixer: start value has mixed support, consider using flex-start instead
> 
> ```
> 
> 
> 
> Looks like we need to either get rid of the warning or ignore it.

I opened an issue, they didn't update the deps in react-topology library. I added the issue to the epic so let's wait until they make the release

aljesusg:
Found the problem, they need to update the deps but we had in the yarn.lock both react-styles installed. yarn management is not upgrading correctly this. So ...FIXED !

ferhoyos (approved):
LGTM, tested locally. Glad to check that autoprefixer warnings have been removed.

Issue #6504, opened 2023-08-22
Sorting by cluster does not work in the list view located in the Overview page


matejnesuta:
Go to /kiali/console/overview?duration=60&refresh=0&displayMode=2

When choosing to sort by Cluster, the list gets sorted by Name instead. 
![image](https://github.com/kiali/kiali/assets/45061792/3fddf161-3cf5-4f04-a197-9a9502781595)
![image](https://github.com/kiali/kiali/assets/45061792/58e498b9-fec7-448d-8699-68479e227f95)

PR #6590, opened 2023-09-07
Fix Cluster column sort on Overview list view

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6504


matejnesuta (approved):
Works as expected. Thank you!

PR #6607, opened 2023-09-13
Make the scrollbar in RenderComponentScroll optional

### LINKED ISSUE #6601: Avoid two scrollbars in certain pages of OSSMC plugin
Two scrollbars appear in certain pages of Openshift Service Mesh Console plugin (OSSMC), making the user experience worse since the user sometimes have to play with both scrollbars to check some OSSMC information.

Check if it is possible to remove one of the scrollbars maintaining the same user experience in Kiali standalone application.

YAML Editor of Istio config object:
image

Trace details:
image

Envoy details:
image

ferhoyos:
**Describe the change**

By default, Kiali hides the global scrollbar (`overflow: hidden` in global style) and fixes the height for pages that use RenderComponentScroll to force the scrollbar to appear. The issues is that hiding the global scrollbar is not possible when Kiali is embedded in other application (like Openshift Console). For this reason an environment variable `RENDER_SCROLLBAR` has been added to make the scrollbar in RenderComponentScroll optional. When `RENDER_SCROLLBAR` is false, the height is not fixed and the component does not force a new scrollbar to appear (the user will use the global scrollbar in this case).

This environment variable is not defined in standalone Kiali application (the value is always true).

Besides, heights of some components have been adjusted in Kiosk mode to fit better in OSSMC withouth breaking the user experience in Kiosk mode:
- IstioConfigDetailsPage:
image

- WorkloadPodLogs: 
image

**PR Testing**
This PR only affects to Kiali standalone in kiosk mode and OSSMC. Check that scrollbar works as expected in standalone application and the pages with heights adjusted look fine in kiosk mode.

**Issue reference**

Closes #6601

### DIFF
--- a/frontend/src/components/IstioConfigPreview/EditorPreview.tsx
+++ b/frontend/src/components/IstioConfigPreview/EditorPreview.tsx
@@ -72,7 +72,7 @@ export class EditorPreviewComponent extends React.Component<Props, State> {
         mode="yaml"
         theme={this.props.theme === Theme.DARK ? 'twilight' : 'eclipse'}
         onChange={value => this.onChange(value)}
-        height={'275px  '}
+        height={'275px'}
         width={'100%'}
         className={istioAceEditorStyle}
         wrapEnabled={true}
--- a/frontend/src/components/Kiosk/KioskActions.ts
+++ b/frontend/src/components/Kiosk/KioskActions.ts
@@ -90,7 +90,7 @@ export const isParentKiosk = (kiosk: string): boolean => {
 // Message has no format, parent should parse it for its needs
 const sendParentMessage = (msg: string): void => {
   // Kiosk parameter will capture the parent target when kiosk !== "true"
-  // this will enable iframe -> parent communication
+  // this will enable parent communication
   const targetOrigin = store.getState().globalState.kiosk;
   window.top?.postMessage(msg, targetOrigin);
 };
--- a/frontend/src/components/Nav/Page/RenderComponentScroll.tsx
+++ b/frontend/src/components/Nav/Page/RenderComponentScroll.tsx
@@ -15,9 +15,9 @@ const EMBEDDED_PADDING = 42;
  * By default, Kiali hides the global scrollbar and fixes the height for some pages to force the scrollbar to appear
  * Hiding global scrollbar is not possible when Kiali is embedded in other application (like Openshift Console)
  * In these cases height is not fixed to avoid multiple scrollbars (https://github.com/kiali/kiali/issues/6601)
- * This environment variable is not defined in standalone Kiali application (default value is true)
+ * GLOBAL_SCROLLBAR environment variable is not defined in standalone Kiali application (value is always false)
  */
-const renderScrollbar = process.env.RENDER_SCROLLBAR ?? 'true';
+const globalScrollbar = process.env.GLOBAL_SCROLLBAR ?? 'false';
 
 const componentStyle = kialiStyle({
   padding: '20px'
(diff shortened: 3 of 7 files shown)


jshaughn (commented):
I haven't tested this yet but the code looks good.  My first comment is just that I find the RENDER_SCROLLBAR environment var a little confusing.  It doesn't have much context.  I think maybe GLOBAL_SCROLLBAR or something that more describes the actual environment may make more sense.  When set to 'true' it would then deactivate the normal behavior of RenderComponentScroll.  I don't know, maybe it's just me.

ferhoyos:
Yes, I don't like the variable name either but I could not find a better one. I prefer your suggestion, thanks @jshaughn!

josunect:
@ferhoyos how can I test it in OSSMC? Apart from copy the related files, is there any related PR for the plugin?

josunect:
> Hi @josunect, yes you have to add GLOBAL_SCROLLBAR environment variable in webpack and .env files. I have just created PR [kiali/openshift-servicemesh-plugin#209](https://github.com/kiali/openshift-servicemesh-plugin/pull/209) in OSSMC repository to test it. Thanks for the review!

Thank you!

josunect (approved):
Verified in OSSMC. 
Before the change: 

image

No additional scroll after this PR: 

image


Istio config: 

image

josunect (commented):
Is this the expected behavior? 

image

josunect (approved):
Double scroll in some pages before the PR: 

Istio config details: 

image

Span details: 

image

Envoy: 

image

. After the PR no double scrolls: 

Traces: 

image

Envoy: 
image

Istio config: 

image

ferhoyos:
In some pages there is double scroll to avoid very small components, but the user experience is not bad because the component with the second scroll fits in the screen (don't need to play with both scrolls at the same time):
- WorkloadPodLogs:
image

- Envoy -> Bootstrap and Config
image

jshaughn:
>  In some pages there is double scroll to avoid very small components, but the user experience is not bad because the component with the second scroll fits in the screen (don't need to play with both scrolls at the same time):

+1, it makes sense to still have scroll in the logs display box, for example, as well as for the whole logs page.

ferhoyos:
Checking environment variables I have figured out that all environment variables used in Kiali are stored in local variables except API_PROXY that is obtained through function. Since this value does not change at runtime I think it is better to be stored in a variable and we save one function call per API request. Small refactor performed to fix this.

jshaughn (approved):
Code looks good, thanks for changing the EnvVar name.  Tested against Standalone Kiali for regressions and everything looks good.  I couldn't break it.  There was one small strange behavior on a config detail.  On a short screen the yaml editor resized to not use all of the available space.  But with more height it expanded to the full space.  Definitely not a blocker.

image

Also tested some stuff adding `&kiosk=true` and it looked fine.

UPDATE:  I notice that the behavior above for the yaml window is not due to this PR, it was like that in master as well!

ferhoyos:
@jshaughn the yaml editor height issue on resizing is a consequence of how height is calculated:
```
height={`calc(var(--kiali-yaml-editor-height)`}
```

The value of kiali-yaml-editor-height is:
```
--kiali-yaml-editor-height': 'calc(var(--kiali-details-pages-tab-content-height) - 170px)'
```

image

The problem is that when you make the screen smaller, there is one moment when the bottom part size is less than 170px, but the kiali yaml editor height still substracting that amount of pixels. For this reason you have that unavailable space.

Best practice is to calculate dynamically widths and heights, but not easy thing to do in most cases. Anyway IMO it is not an important issue in Kiali since it is a desktop application and it does not need to support mobile or tablets.

PR #6622, opened 2023-09-15
[v1.73] Fix for double tap on a node in MC redirecting to wrong cluster (#6542)

ferhoyos:
Backport of https://github.com/kiali/kiali/pull/6542 to Kiali version v1.73


josunect (approved):
I've tested and if we first click on a workload from a remote cluster, the redirection is correct: 

image

Double click in reviews > v2: 

image

But if we click again in V2 it looks like it is not saving the cluster: 

image

I've just tested that is the same behavior than in master, so I will approve and probably will be better to address in a new PR?

ferhoyos:
Thanks @josunect! @leandroberetta can you take a look and create new PR if needed?

Issue #6615, opened 2023-09-14
Multicluster - Traffic routings created via Graph page are always located in the local cluster


matejnesuta:
If you try to create a Traffic Routing via the main graph, it always creates it in the local cluster, even though the service is in the remote cluster. This happens both with the node's context menu and the side panel menu.

Steps to reproduce:
- go to the Graph page and choose any service node located in the remote cluster
- create any Traffic Routing
- now the routing cannot be delete through the context menu nor the side panel
- Go to the Istio page and the traffic routing should be located in the local cluster

![image](https://github.com/kiali/kiali/assets/45061792/969f2e33-c6cb-4289-9b46-70d3509e1c6a)
![image](https://github.com/kiali/kiali/assets/45061792/40979228-e5e2-4ef9-80a8-7af0030378a1)

matejnesuta:
I am also not sure what the expected outcome is in this case. Please let me know, so I can write a Gherkin scenario.

matejnesuta:
Yes. I created a Traffic Shifting object and it was placed in the right cluster. 
![image](https://github.com/kiali/kiali/assets/45061792/9ff8d63f-e766-4cce-9045-3045ea630505)
![image](https://github.com/kiali/kiali/assets/45061792/81a1473a-18a5-4d7e-b33e-183a660093d2)

PR #6625, opened 2023-09-15
Added missing cluster params in graph wizards.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6615


matejnesuta:
Tested this both without and with the Istio CRDs and objects now are created in the correct cluster when creating them from the Graph node wizard/side panel. Service details page remains uneffected by this change. LGTM.

matejnesuta (approved):

Issue #6510, opened 2023-08-22
"Cannot load the graph: json: cannot unmarshal object into Go value of type []*kubernetes.RegistryEndpoint"


howardjohn:
UI fails to load with this error: `Cannot load the graph: json: cannot unmarshal object into Go value of type []*kubernetes.RegistryEndpoint`

Image `quay.io/kiali/kiali:v1.72`

Cluster is GKE 1.27

Running OSS Istio from master with ambient enabled

```
2023-08-22T23:34:46Z ERR json: cannot unmarshal object into Go value of type []*kubernetes.RegistryEndpoint: goroutine 6341 [running]:
runtime/debug.Stack()
        /opt/hostedtoolcache/go/1.20.5/x64/src/runtime/debug/stack.go:24 +0x65
github.com/kiali/kiali/handlers.handlePanic({0x232aee0, 0xc000cb3b90})
        /home/runner/work/kiali/kiali/handlers/graph.go:86 +0x219
panic({0x1c286e0, 0xc0003ad8f0})
        /opt/hostedtoolcache/go/1.20.5/x64/src/runtime/panic.go:884 +0x213
github.com/kiali/kiali/graph.CheckError(...)
        /home/runner/work/kiali/kiali/graph/util.go:38
github.com/kiali/kiali/graph/telemetry/istio/appender.getServiceList({0xc000f35986, 0xa}, {0xc000540c10, 0x6}, 0xc00154bda0)
        /home/runner/work/kiali/kiali/graph/telemetry/istio/appender/appender.go:291 +0x314
github.com/kiali/kiali/graph/telemetry/istio/appender.getServiceLists(0x8?, {0xc000540c10, 0x6}, 0xc00079d050?)
        /home/runner/work/kiali/kiali/graph/telemetry/istio/appender/appender.go:264 +0xac
github.com/kiali/kiali/graph/telemetry/istio/appender.IstioAppender.AppendGraph({0x202afe8?}, 0xc00154be00, 0xc00059c240?, 0xc000c85908)
        /home/runner/work/kiali/kiali/graph/telemetry/istio/appender/istio_details.go:47 +0x65
github.com/kiali/kiali/graph/telemetry/istio.BuildNamespacesTrafficMap({_, _}, {0xc00154bc50, {0x0, {0xc00150b440, 0x6, 0x6}}, 0x0, 0x1, 0xc00154bb90, ...}, ...)
[...]

PR #6629, opened 2023-09-15
refactor for the /debug/endpointz changes in Istio 1.20

jmazzitelli:
fixes: #6510

### DIFF
--- a/business/registry_status.go
+++ b/business/registry_status.go
@@ -174,10 +174,16 @@ func filterRegistryEndpoints(registryStatus *kubernetes.RegistryStatus, criteria
 	if criteria.Namespace != "" && criteria.ServiceName != "" {
 		for _, rEndpoint := range registryStatus.Endpoints {
 			// External ServiceEntries don't have endpoints in the Istio Registry
-			if rEndpoint.Endpoints != nil {
-				for _, innerEndpoint := range rEndpoint.Endpoints {
-					if innerEndpoint.Service.Attributes.Namespace == criteria.Namespace && innerEndpoint.Service.Attributes.Name == criteria.ServiceName {
-						filteredRegistryEndpoints = append(filteredRegistryEndpoints, rEndpoint)
+			if rEndpoint.IstioServiceEndpointShards != nil {
+				for serviceName, innerEndpoint := range rEndpoint.IstioServiceEndpointShards {
+					for _, shards := range innerEndpoint {
+						for _, shardArray := range shards.Shards {
+							for _, shard := range shardArray {
+								if shard.Namespace == criteria.Namespace && serviceName == criteria.ServiceName {
+									filteredRegistryEndpoints = append(filteredRegistryEndpoints, rEndpoint)
+								}
+							}
+						}
 					}
 				}
 			}
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -426,16 +426,17 @@ func ParseRegistryEndpoints(endpoints map[string][]byte) ([]*RegistryEndpoint, e
 		if isRegistryLoaded {
 			break
 		}
-		var eps []*RegistryEndpoint
+		var eps IstioServiceEndpointShards
 		err := json.Unmarshal(endpoint, &eps)
 		if err != nil {
 			log.Errorf("Error parsing RegistryEndpoints results: %s", err)
 			return nil, err
 		}
-		for _, ep := range eps {
-			ep.pilot = pilot
+		re := RegistryEndpoint{
+			pilot:                      pilot,
+			IstioServiceEndpointShards: eps,
 		}
-		fullRegistryEndpoints = append(fullRegistryEndpoints, eps...)
+		fullRegistryEndpoints = append(fullRegistryEndpoints, &re)
 		if len(eps) > 0 {
 			isRegistryLoaded = true
 		}
--- a/kubernetes/istio_internal_test.go
+++ b/kubernetes/istio_internal_test.go
@@ -287,8 +287,8 @@ func TestParseRegistryEndpoints(t *testing.T) {
 	assert.NoError(err2)
 	assert.NotNil(registry)
 
-	assert.Equal(101, len(registry))
-	assert.Equal("*.msn.com:http-port", registry[0].Service)
+	assert.Equal(69, len(registry[0].IstioServiceEndpointShards))
+	assert.Equal("http", registry[0].IstioServiceEndpointShards["music.electronic-shop.svc.cluster.local"]["electronic-shop"].Shards[ShardKey{"Kubernetes", "Kubernetes"}][0].ServicePortName)
 }
 
 func TestRegistryServices(t *testing.T) {
--- a/kubernetes/types.go
+++ b/kubernetes/types.go
@@ -250,11 +250,10 @@ type RegistryEndpoint struct {
 	IstioServiceEndpointShards
 }
 
-// IstioEndpointShards is a helper struct to fetch the /debug/endpointz results before parsing it to the Kiali's service model.
+// IstioServiceEndpointShards is a helper struct to fetch the /debug/endpointz results before parsing it to the Kiali's service model.
 // Not all fields from /debug/endpointz are mapped, only those needed by Kiali.
 // There may be differences between Istio versions to be addressed case by case in the mapping.
 // See: https://github.com/istio/istio/blob/be3ca0bbb3f06f4151d6353a37bb91e20cfd811c/pilot/pkg/model/endpointshards.go#L39-L47
-
 type IstioServiceEndpointShards map[string]EndpointShards
 type EndpointShards map[string]struct {
 	Shards map[ShardKey][]EndpointShard `json:"Shards,omitempty"`
--- a/kubernetes/types.go
+++ b/kubernetes/types.go
@@ -1,7 +1,8 @@
 package kubernetes
 
 import (
-	"time"
+	"fmt"
+	"strings"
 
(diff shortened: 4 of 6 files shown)


jmazzitelli:
This is how I got the test data file that is included in this PR:

```
# Install Istio 1.20 (only available as a dev build right now):
hack/istio/download-istio.sh -div 1.20-dev
hack/istio/install-istio-via-istioctl.sh -ih default -it default
# Install demo:
kubectl create namespace electronic-shop
kubectl label namespace electronic-shop istio-injection=enabled
kubectl apply -n electronic-shop -f <(curl -L https://raw.githubusercontent.com/kiali/demos/master/electronic-shop/electronic-shop.yaml)
# Get istiod data:
kubectl exec -it -n istio-system deploy/istiod -- curl http://localhost:15014/debug/endpointz | jq 
```

jmazzitelli:
This already looks like it is close. I can see everything I would expect to see in the UI (a good graph, istio config objects, apps/services/workloads lists, etc).

But the integration tests are failing, so there looks to be some things that are still broken.

jmazzitelli:
In order for the CI tests to pass, this diff must be done (in other words, the CI tests will fail - you have to have these changes so Istio 1.20 is used during the tests).

```diff
$ git diff
diff --git a/hack/istio/multicluster/install-primary-remote.sh b/hack/istio/multicluster/install-primary-remote.sh
index cb53962ac..8a316818e 100755
--- a/hack/istio/multicluster/install-primary-remote.sh
+++ b/hack/istio/multicluster/install-primary-remote.sh
@@ -57,7 +57,7 @@ switch_cluster "${CLUSTER1_CONTEXT}" "${CLUSTER1_USER}" "${CLUSTER1_PASS}"
 ${CLIENT_EXE} label namespace ${ISTIO_NAMESPACE} topology.istio.io/network=${NETWORK1_ID}
 
 ISTIO_INSTALL_SCRIPT="${SCRIPT_DIR}/../install-istio-via-istioctl.sh"
-${ISTIO_INSTALL_SCRIPT} --client-exe-path ${CLIENT_EXE} --cluster-name ${CLUSTER1_NAME} --istioctl ${ISTIOCTL} --istio-dir ${ISTIO_DIR} --mesh-id ${MESH_ID} --namespace ${ISTIO_NAMESPACE} --network ${NETWORK1_ID} --set values.pilot.env.EXTERNAL_ISTIOD=true --k8s-gateway-api-enabled true
+${ISTIO_INSTALL_SCRIPT} --client-exe-path ${CLIENT_EXE} --cluster-name ${CLUSTER1_NAME} --istioctl ${ISTIOCTL} --istio-dir ${ISTIO_DIR} --mesh-id ${MESH_ID} --namespace ${ISTIO_NAMESPACE} --network ${NETWORK1_ID} --set values.pilot.env.EXTERNAL_ISTIOD=true --k8s-gateway-api-enabled true -it default -ih default
 
 GEN_GATEWAY_SCRIPT="${ISTIO_DIR}/samples/multicluster/gen-eastwest-gateway.sh"
 ${GEN_GATEWAY_SCRIPT} --mesh ${MESH_ID} --cluster ${CLUSTER1_NAME} --network ${NETWORK1_ID} | ${ISTIOCTL} --context=${CLUSTER1_CONTEXT} install -y -f -
diff --git a/hack/setup-kind-in-ci.sh b/hack/setup-kind-in-ci.sh
index e85b1d337..aa4ba5676 100755
--- a/hack/setup-kind-in-ci.sh
+++ b/hack/setup-kind-in-ci.sh
@@ -93,14 +93,14 @@ if [ -n "${ISTIO_VERSION}" ]; then
 fi
 
 infomsg "Downloading istio"
-hack/istio/download-istio.sh ${DOWNLOAD_ISTIO_VERSION_ARG}
+hack/istio/download-istio.sh -div 1.20-dev
 
 setup_kind_singlecluster() {
   "${SCRIPT_DIR}"/start-kind.sh --name ci --image "${KIND_NODE_IMAGE}" 
 
   infomsg "Installing istio"
   # Apparently you can't set the requests to zero for the proxy so just setting them to some really low number.
-  hack/istio/install-istio-via-istioctl.sh --reduce-resources true --client-exe-path "$(which kubectl)" -cn "cluster-default" -mid "mesh-default" -net "network-default" -gae "true"
+  hack/istio/install-istio-via-istioctl.sh --reduce-resources true --client-exe-path "$(which kubectl)" -cn "cluster-default" -mid "mesh-default" -net "network-default" -gae "true" -ih default -it default
 
   infomsg "Pushing the images into the cluster..."
   make -e DORP="${DORP}" -e CLUSTER_TYPE="kind" -e KIND_NAME="ci" cluster-push-kiali
@@ -172,7 +172,7 @@ setup_kind_multicluster() {
   fi
 
   infomsg "Downloading istio"
-  hack/istio/download-istio.sh ${DOWNLOAD_ISTIO_VERSION_ARG}
+  hack/istio/download-istio.sh -div 1.20-dev
 
   infomsg "Cloning kiali helm-charts..."
   git clone --single-branch --branch "${TARGET_BRANCH}" https://github.com/kiali/helm-charts.git "${HELM_CHARTS_DIR}"
```

jmazzitelli:
With the above diff in place, `hack/run-integration-tests.sh --test-suite backend` passes on my local machine.

jmazzitelli:
With the above diff in place, `hack/run-integration-tests.sh --test-suite frontend` also passes on my local machine.

jmazzitelli:
Leaving this in draft because we cannot merge this into master due to the fact this code will only work with Istio 1.20.

jmazzitelli:
I'm considering removing all of this code, removing all Kiali access to istiod /debug/endpointz, and trying to obtain the needed data elsewhere. This saves us from maintaining code talking to a private/unsupported istiod API, and it saves us from loading and unmarshalling a huge amount of data for what amounts to 2 small pieces of information - a service port's "istio protocol" and a service port's "TLSMode".

What Kiali Needs Today from /debug/endpointz
===

These are the only 2 pieces of [/debug/endpointz data](https://github.com/kiali/kiali/blob/v1.73/models/port.go#L10-L21) - no other piece of data is even referenced in Kiali:
```go
	IstioProtocol string  `json:"istioProtocol"`
	TLSMode string `json:"tlsMode"`
```

Istio  Protocol
---

Note there are three "protocols" actually defined within this Port object:
```go
	Protocol      string  `json:"protocol"`
	AppProtocol   *string `json:"appProtocol,omitempty"`
	IstioProtocol string  `json:"istioProtocol"`
```

Protocol and AppProtocol are probably the same as those defined in [a k8s Service](https://kubernetes.io/docs/reference/networking/service-protocols/), but I do not know what "IstioProtocol" refers to. There are no code comments to explain this field; it is a mystery. :)

TLSMode
---

There are comments to explain TLSMode - which is a string that apparently has only 2 valid values:
* `disabled` = this endpoint should receive traffic as-is (mostly plaintext)
* `istio` = the endpoint is ready to receive Istio mTLS connections.

Where This Data Is Used in Kiali UI
===

These two pieces of data are only referenced/visualized in the Kiali UI in one place - [ServiceNetwork.tsx](https://github.com/kiali/kiali/blob/v1.73/frontend/src/pages/ServiceDetails/ServiceNetwork.tsx#L170-L178). Navigate to the "Services" left-hand nav menu, and pick some Service and go to its Overview tab. In the Network card you will see something like:
image
There you can see "mazz-protocol" (which was dummy data I added so I can visualize `IstioProtocol == 'mazz-protocol'`) and you can see the mTLS icon to visualize `TLSMode == 'istio'`

That's it. Those two pieces of string text in this ServiceNetwork card is the only place where /debug/endpointz data is used (so far as I can find). IMO, we should either (a) try to find these two pieces of data elsewhere in a public API or (b) just remove it here. Either way, it will free us from /debug/endpointz bondage - no longer would we need to maintain huge json (and thus we'd save all the dev work it would take to maintain it, as well as saving all the runtime resources that are required to retrieve the data from istiod, unmarshall it, then store the data in memory). Seems overkill to do all that just to obtain 2 pieces of data that only show up in this small corner of the Kiali UI.

jmazzitelli:
I asked John Howard for his insight on this. He said that /debug/endpointz TLSMode is not what we think it is. He said, `TLSMode almost means "the pod has a sidecar. But in any event, can be (is, in Istio) derived from the pod/WorkloadEntry object labels.`

As for Kiali showing the lock on the service, he opines, `"Its a bit odd to use that data to give a lock icon next to a service."`

In his thinking, a better name for TLSMode would be `"This pod should handle TLS". Its the signal we use for auto-mtls to decide which endpoints are capable of handling TLS. Its one input to "will mTLS be used for this connection"`

In short, Kiali should not be exposing this internal data to Istio end-users - it is confusing at best, and no one will understand its true semantics. For sure it does NOT mean "connections to this service are always secured via mTLS".

So, because of this, and because protocol is (usually) easily derived from port name, I'm going to rip out of Kiali all its usage of this endpointz API. There is nothing in here we use or should use.

jshaughn:
> So, because of this, and because protocol is (usually) easily derived from port name, I'm going to rip out of Kiali all its usage of this endpointz API. There is nothing in here we use or should use.

alright, so I agree it's misleading as we have it at the moment.  It's not not-valuable, it means that the endpoint is tls-capable, but it doesn't mean the traffic is definitely using mtls.  So I am +1 to eliminating the confusion and the use of this unsupported API.

jmazzitelli:
I'm going to close this PR so I can keep my branch alive JUST in case we want to revive the ability to request and unmarshall this new /debug/endpointz API. But then I will create another PR that rips out all of this.

jmazzitelli:
replacement PR: https://github.com/kiali/kiali/pull/6673

Issue #6609, opened 2023-09-13
Istio warning/error status for situations where eastwestgateway is not healthy.


matejnesuta:
Semi-related to #6606 and #6626. 

![image](https://github.com/kiali/kiali/assets/45061792/ace0a302-574a-4943-b27c-b2237087866f)

There is the Istio Status indicator in the main navigation bar, which turns yellow or red in situations where key pods like Prometheus or Grafana are not healthy. Wouldn't it make sense to add this for the gateway as well when running in the Multicluster scenario? It feels a little strange when Kiali does not show any warning, even though you are not able to see anything from the remote cluster.

hhovsepy:
Agree to include 'eastwestgateway's status into 'Istio Component Status' in multicluster mode.

PR #6647, opened 2023-09-20
Istio Status test for Multicluster

matejnesuta:
Currently, there are 2 different CI jobs for front-end testing and there is also an enhancement in the backlog (#6609), which might extend the Status for external services related to multicluster. Because of this, I think it is suitable to duplicate this test.


ScriptingShrimp (approved):
Okay, will merge

PR #6651, opened 2023-09-20
[CI] [hack] be able to run integration tests using Istio dev builds

jmazzitelli:
this will help with testing PRs like this one that works only with the upstream Istio master branch, but not with any releases yet - https://github.com/kiali/kiali/pull/6629


jshaughn (approved):
LOK

PR #6646, opened 2023-09-20
Fix little disalignments in Kiali UI

ferhoyos:
**Describe the change**

Fix little disalignments in Kiali UI:
- 100% in slider does not fit into numeric box:

Before:
image

After:
image

- Kiali icon in Mesh page not adapted to dark mode:

Before:
image

After:
image

- Dark mode switch with round borders:

Proposal:
image

Before:
image

After:
image

- Envoy tabs with white background:

Before:
image

After: 
image

- (Graph) Info icon too far from input box:

Before:
image

After: 
image

Bonus:
- I have removed defaultIconStyle from KialiIcon component (it does not make sense to export a global defaultIconStyle with no content when you are applying default icon styles inside the component)
- Remove checkboxes and radio button margins (not needed with PF5)

About UI modal issue is fixed in #6654 

**Issue reference**

Part of #6639.

Issue #5843, opened 2023-02-15
Modal style is not consistent (copy and download buttons)


ferhoyos:
Istio config preview modal and debug information modal are not consistent:

Istio Config Preview:
![image](https://user-images.githubusercontent.com/122779323/218974626-2c1113d8-addc-49fa-8085-a12cbea37a9d.png)

Debug information:
![image](https://user-images.githubusercontent.com/122779323/218979212-197aafea-51de-46ea-adff-b45d38318484.png)

It can be noticed that copy and download buttons are different in both modals:
- Icon buttons with tooltips at the top in istio config preview
- Normal button at the footer in debug information

All modals within the application should be consistent (or icon buttons at the top or normal button at the footer). IMO the best option is to place all action buttons in the footer, the icon buttons seem very isolated and do not look good to me.

Besides, I don't know why we have a Close button at the footer in debug information when the modal already provides close icon at the top...

ferhoyos:
Fixed by https://github.com/kiali/kiali/pull/6654

PR #6654, opened 2023-09-21
Refactor AboutModal and sibling items

### LINKED ISSUE #6639: Little disalignments in Kiali UI
I have noticed some little disalignments in Kiali UI due to recent changes (Dark mode and PF5 upgrade). Since they are minor ones I gather them all in one issue:

- 100% in slider does not fit into numeric box:

Before:
image

After:
image

- Envoy tabs with white background:

Before:
image

After: 
image

- About UI image is not shown and non-sense scrollbar appears:

Before:
image

After:
image

- (Graph) Info icon too far from input box:

Before:
image

After: 
image

- Kiali icon in Mesh page not adapted to dark mode:

Light theme:
image

Dark theme:
image

- Dark mode switch with round borders:

Proposal:
image

Current switch:
image

ferhoyos:
**Describe the change**

Working on #6639 I have fixed About Modal little disalignments and added Kiali icon in the background:
Before:
image

After:
image

Apart from little UI fix, I have easily converted AboutModal component to React Hook (thinking in future React 18 upgrade, better to migrate such small components little by little). The problem is that to perform this conversion I have had to convert its parent (HelpDropdown) and siblings components too (DebugInformation and IstioCertsInfo). 

I prefer to separate this PR from the other (#6646) since it contains 4 conversions to React Hooks and maybe some extra review is needed.

**Issue reference**

Closes #6639.


jshaughn (dismissed):
Tested and everything seems to work and alignments look good.  I notice the `View Debug Info` tabs shifted left, but I think that's actually a fix.  This is a bit more of a refactor than I would have expected, it's not clear to me that even a move to React 18 would requires a chnage to Hooks.  But, I'm not really against it as I do think we will be in a better position with everone getting more comfortable with Hooks.

A couple of non-blocking comments (I don't think we need to spend huge amount of time on `View Debug Info`)...  This is unchanged from before, but I'm not sure why in `View Debug Info` the modal size changes when we change tabs.  Also, we may want to start being more consistent with our buttons.  For example, in PF the Close/Cancel is typically on the inside and styles differently.  For example, from https://www.patternfly.org/components/modal:

image

From `View Debug Info`:

image

For what it's worth, we could also use icons for copy/download...

ferhoyos:
Hi @jshaughn, thanks for the review. Here are some comments:
- You are right, it is not a requirement to move classes to Hooks in React 18. However, I think that it is worth to migrate small components to Hooks (easier maintenance and readibility). For big ones, maybe it won't never be migrated due to big effort, but that is another story...

- I have set the height of Kiali Config table so the modal size is not modified when you change tabs in `View Debug Info`.

- I agree with you that buttons are not consistent (actually I created an issue some months ago (https://github.com/kiali/kiali/issues/5843). For this case I have removed the close button (not needed having a close button at the top) and convert Copy and Download buttons to icon + name links:
image

- Close button of View Certificate Info modal footer also removed (not needed having a close button at the top):
image

- To be consistent I have done the same for other copy and download icons in other parts of the application:
WorkloadPodLogs:
image

IstioConfigPreview:
image

Envoy:
image

This is my proposal, if you don't like any of the changes please let me know.

aljesusg:
> Hi @jshaughn, thanks for the review. Here are some comments:
> 
> - You are right, it is not a requirement to move classes to Hooks in React 18. However, I think that it is worth to migrate small components to Hooks (easier maintenance and readibility). For big ones, maybe it won't never be migrated due to big effort, but that is another story...
> 
> 
> 
> - I have set the height of Kiali Config table so the modal size is not modified when you change tabs in `View Debug Info`.
> 
> 
> 
> - I agree with you that buttons are not consistent (actually I created an issue some months ago (https://github.com/kiali/kiali/issues/5843). For this case I have removed the close button (not needed having a close button at the top) and convert Copy and Download buttons to icon + name links:
> 
> image
> 
> 
> 
> - Close button of View Certificate Info modal footer also removed (not needed having a close button at the top):
> 
> image
> 
> 
> 
> - To be consistent I have done the same for other copy and download icons in other parts of the application:
> 
> WorkloadPodLogs:
> 
> image
> 
> 
> 
> IstioConfigPreview:
> 
> image
> 
> 
> 
> Envoy:
> 
> image
> 
> 
> 
> This is my proposal, if you don't like any of the changes please let me know.
> 
> 

-1 in icons plus "word" I didn't see that pattern in other interfaces. IMO we should ask about this to UX before remove icon+tooltip

jshaughn:
> -1 in icons plus "word" I didn't see that pattern in other interfaces. IMO we should ask about this to UX before remove icon+tooltip

+1 to asking Andrew about this one.   In OCP (on my 4.12 CRC) I can see that they do use the icon+word (which I found surprising), but in PF I don't see a clear example, so I'm not sure.

+ to the other changes.

aljesusg:
> > -1 in icons plus "word" I didn't see that pattern in other interfaces. IMO we should ask about this to UX before remove icon+tooltip
> 
> 
> 
> +1 to asking Andrew about this one.   In OCP (on my 4.12 CRC) I can see that they do use the icon+word (which I found surprising), but in PF I don't see a clear example, so I'm not sure.
> 
> 
> 
> + to the other changes.

Mmm ok so if the idea is to be aligned with OCP so ok. But ask Andrew because as you said not PF examples and other interfaces ...only ocp

ferhoyos:
+1 to asking @andrew-ronaldson, both approaches are valid and this is more related with user experience and UI design trends. IMO there are some icons that they are not clear what exactly do (copy, download, reset, replay) and adding some text can help the user to understand the button function at a glance without having to hover for a tooltip explanation. 

@jshaughn in ocp I have seen both approaches (even for the same icon in the same page):

image

aljesusg:
> +1 to asking @andrew-ronaldson, both approaches are valid and this is more related with user experience and UI design trends. IMO there are some icons that they are not clear what exactly do (copy, download, reset, replay) and adding some text can help the user to understand the button function at a glance without having to hover for a tooltip explanation.
> 
> @jshaughn in ocp I have seen both approaches (even for the same icon in the same page):
> 
> image

Ask andrew because IMO since some years all people usually see these icons to perform these actions (Yes copy, download, edit - pencil, reset - arrow circle include ) that's why there is no need to put the word, to make it clear, as you indicate, it was proposed to put the tooltip. 

For example in backstage they have the icon with tooltips.
image

As far as I remenber from some classes I received about user interface design The idea of ​​this proposal was to reduce the size of the "control bar" and give the possibility to add more options without having to redesign everything given that the user already knew those buttons from day to day...but of course Andrew should clarify these for us and tell us what is best. _I just wanted to share my opinion_.

ferhoyos:
@aljesusg thanks for your opinion. Maybe we do not have to apply the same approach for all icons in the application, and it depends on the context and how much space you have for that use case. In the backstage table above there is no space to add any text so it is ok to add only icon with tooltip. In other examples if you just put only one icon alone without text it seems that you are not taking advantage of the space you have (it is just a feeling, hard to explain):

For example in View Debug Info:
Only icons:
image

Icons with text:
image

IMO in this case it looks better if you add some text to the icon, but maybe not in other cases where you have lots of icons and adding text will overwhelm the screen layout.

ferhoyos:
From conversation with @andrew-ronaldson:
- If there is enough room space, better to add text to link icons.
- DebugInformation component: Copy and Download should be buttons (secondary style) since they copy/download what the user sees. Close button with primary style.

image

I have reviewed all link icons that we have in Kiali (in the review process I have fixed some margins and CSS stuff) and add text where applicable:
- WorkloadPodLogs:

image

- IstioConfigPreview:

image

- Envoy:

image

- Graph Toolbar:

image

I think info icons are well-known by users to add 'Info' text (the same for delete button in Annotations wizard).

image

jshaughn (approved):
Looks good and matches what I took away from the UX meeting!

aljesusg (approved):
LGFM

Issue #6449, opened 2023-08-05
Move deprecated Select


aljesusg:
## Requirement 

https://github.com/kiali/kiali/issues/6446

## Detect with

![image](https://github.com/kiali/kiali/assets/3019213/535a9e94-f999-401b-84f6-a0b7008117f9)

Documentation: https://www.patternfly.org/components/menus/select

Parent: https://github.com/kiali/kiali/issues/6438

PR #6656, opened 2023-09-22
Migrate PF5 deprecated Selects 

ferhoyos:
**Describe the change**

Migrate deprecated Select components to new PF5 ones.

Select locations:
- Summary trace details (Graph):

image

- Toolbar dropdown:

image

- Stateful Filters (typeahead):

image

Besides, I have fixed small issue on Kiosk Actions found when I tested TimeDurationModal. If kiosk value is true (kiosk activated but no parent is defined), the postMessage is sent to invalid targetOrigin and throws a console error.  

**Issue reference**

Closes https://github.com/kiali/kiali/issues/6449.


hhovsepy (approved):
Looks good to me, tested locally, however was not able to replicate the use-case with the dropdown in graph traces.

jshaughn (approved):
Code looks good, I tested everything I could think of, all good!

PR #6659, opened 2023-09-26
Component status check in multicluster mode

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6609

Now it is possible to check Istio component status on multiple clusters,
just add to any component's config the new `is_multicluster: true` (by default 'false') and the component status check will check the existence of the particular component in all clusters.
```
external_services:
  istio:
    component_status:
      components:
      - app_label: istio-eastwestgateway
        is_core: true
        is_proxy: true
        is_multicluster: true
        namespace: ''
```

Screenshot from 2023-09-26 14-17-00

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -16,6 +16,8 @@ import (
 	"github.com/kiali/kiali/util/httputil"
 )
 
+const istioEastWestGateway = "istio-eastwestgateway"
+
 // SvcService deals with fetching istio/kubernetes services related content and convert to kiali model
 type IstioStatusService struct {
 	userClients   map[string]kubernetes.ClientInterface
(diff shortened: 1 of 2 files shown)


jshaughn (commented):
There is nothing wrong with this check but are you sure `istio-eastwestgateway` is a name you can rely on?  I would think the gateway could be called anything, I think this naming is just part of a sample?

jmazzitelli:
> There is nothing wrong with this check but are you sure `istio-eastwestgateway` is a name you can rely on? I would think the gateway could be called anything, I think this naming is just part of a sample?

Yeah. That's only the name we give it in our demos. That name can be anything.

nrfox:
Not all multi-cluster setups will have an east-west gateway. I believe multi-cluster setups with a flat network don't necessarily need one. This check would be useful to have but maybe it should be off by default unless we can detect when it should/shouldn't be enabled?

jshaughn (approved):
Code looks fine to me, defering to @matejnesuta for live test.

matejnesuta (approved):
LGTM

PR #6661, opened 2023-09-28
Fix double tap on same node

leandroberetta:
Fixes https://github.com/kiali/kiali/issues/6657


josunect (approved):
Thanks, @leandroberetta 
It works as expected now! 

image

Code LG

Issue #6042, opened 2023-04-18
[code] provide function to get versioned/canary istio configmap name


jmazzitelli:
See: https://github.com/kiali/kiali/pull/6027#pullrequestreview-1388615090

We have code like this:

```
icm, err := kialiCache.GetConfigMap(configObject.IstioNamespace, configObject.ExternalServices.Istio.ConfigMapName)
```

We would like to work when using Istio revisions (e.g. canary)

We should have a function that can be used to isolate the configmap name... Something like "getIstioConfigMapName" and have it deal with the versioning/canary details.

Xunzhuo:
xref: https://github.com/kiali/kiali/issues/6043

jmazzitelli:
someone in discussions has asked for the ability to auto-detect canary configmap: https://github.com/kiali/kiali/discussions/6644

yogeek:
don't know if it is the right place but I am using istio canary revisions (1.18 and 1.20 are present in the cluster at the same time, in the future it will be 1.20 and 1.22, etc) in order to allow users to choose when they want to upgrade to the latest revision.
How is it supposed to work with kiali please ? does kiali knows how to handle multiple istio revisions ?
Apparently, the `config_map_name` has to be specified in the CR but do I have to choose only 1 revision ?
Does this issue meant to cover both revisions or only one ?

jmazzitelli:
> if it is the right place

For generation questions like this, please post a new discussion in our github Discussions - https://github.com/kiali/kiali/discussions/new?category=q-a

For details see:

PR #6675, opened 2023-10-05
Fix double revision in istio configmap name

### LINKED ISSUE #6669: Double istio rev in configmap name
Hello.

istio installed via helm

```
  values:
    revision: "1-18-0"
```

kiali

```
  values:
    allowAdHocKialiImage: true
    resources:
      limits:
        cpu: 1
    cr:
      create: true
      namespace: istio-system
      spec:
        istio_namespace: istio-system
        external_services:
          istio:
            config_map_name: "istio-1-18-0"
            istiod_deployment_name: "istiod-1-18-0"
            istio_sidecar_injector_config_map_name: "istio-sidecar-injector-1-18-0"
```

i receive this error from kiali log.

`ERR Error getting mesh config: configmap "istio-1-18-0-1-18-0" not found`

i do not understand why rev is doubled in configmap name?

nrfox:
Uses: `conf.ExternalServices.Istio.ConfigMapName` without adding the rev tag when it is set and not the default `istio`.

Fixes #6669

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -155,10 +155,7 @@ func (in *MeshService) GetMesh(ctx context.Context) (*Mesh, error) {
 					Revision: istiod.Labels[IstioRevisionLabel],
 				}
 
-				configMapName := in.conf.ExternalServices.Istio.ConfigMapName
-				if revLabel := istiod.Labels[IstioRevisionLabel]; revLabel != "default" && revLabel != "" {
-					configMapName = configMapName + "-" + revLabel
-				}
+				configMapName := IstioConfigMapName(in.conf, controlPlane.Revision)
 
 				controlPlaneConfig, err := getControlPlaneConfiguration(kubeCache, istiod.Namespace, configMapName)
 				if err != nil {
@@ -224,6 +221,26 @@ func (in *MeshService) GetMesh(ctx context.Context) (*Mesh, error) {
 	}
 
 	return mesh, nil
+}
+
+// IstioConfigMapName guesses the istio configmap name.
+func IstioConfigMapName(conf config.Config, revision string) string {
+	// If the config map name is explicitly set and it's not the default value, we should always use that.
+	// Note that this means that the revision is ignored and every controlplane
+	// will use this configmap regardless of which configmap actually corresponds
+	// to the revision.
+	if conf.ExternalServices.Istio.ConfigMapName != "" && conf.ExternalServices.Istio.ConfigMapName != "istio" {
+		return conf.ExternalServices.Istio.ConfigMapName
+	}
+
+	// If the revision is set, we should use the revisioned configmap name
+	// otherwise the hardcoded 'istio' value is used.
+	configMapName := "istio" // As of 1.19 this is hardcoded in the helm charts.
+	if revision != "default" && revision != "" {
+		configMapName = configMapName + "-" + revision
+	}
+
+	return configMapName
 }
 
 func envVarIsSet(key string, env core_v1.EnvVar) bool {
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -485,7 +485,7 @@ func (in *MeshService) resolveNetwork(clusterName string) string {
 
 func (in *MeshService) OutboundTrafficPolicy() (*models.OutboundPolicy, error) {
 	otp := models.OutboundPolicy{Mode: "ALLOW_ANY"}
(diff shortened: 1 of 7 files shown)


nrfox:
@jmazzitelli please take a look at this when you get a chance

jmazzitelli (changes requested):
This code change makes sense. We need to make the call to this new IstioConfigMap name in all the places where we need the name.

These places: https://github.com/search?q=repo%3Akiali%2Fkiali+%2FGetConfigMap.*ExternalServices.Istio.ConfigMapName%2F&type=code

Here's the files that probably need to be changed in this PR:

```
$ grep -Rl ExternalServices.Istio.ConfigMapName *
business/istio_certs.go
business/namespaces.go
business/tls.go
business/istio_validations.go
business/mesh.go
handlers/config.go
```

Note that this PR directly addresses [this issue](https://github.com/kiali/kiali/issues/6042) too - so we can close this issue when this PR is merged.

yohanb:
@nrfox thanks for this PR!  We just ran into this issue and traced it back to the same faulty code.
When can we expect this to be complete?  Only missing a few refactors ;)

nrfox:
> @nrfox thanks for this PR! We just ran into this issue and traced it back to the same faulty code.
> When can we expect this to be complete? Only missing a few refactors ;)

@yohanb hoping to have this complete by the end of the week so it can be included with the next release.

yohanb:
> > @nrfox thanks for this PR! We just ran into this issue and traced it back to the same faulty code.
> > When can we expect this to be complete? Only missing a few refactors ;)
> 
> @yohanb hoping to have this complete by the end of the week so it can be included with the next release.

Awesome, thanks so much 👍

nrfox:
@jmazzitelli it seemed like the e2e tests were failing because of the tracing flakes. I rebased with main to pickup the latest fix. I also added some additional test coverage to the area that was causing this error to be logged which was in gateway validations.

business/mesh.go
```diff
@@ -155,10 +155,7 @@ func (in *MeshService) GetMesh(ctx context.Context) (*Mesh, error) {
 					Revision: istiod.Labels[IstioRevisionLabel],
 				}
 
-				configMapName := in.conf.ExternalServices.Istio.ConfigMapName
-				if revLabel := istiod.Labels[IstioRevisionLabel]; revLabel != "default" && revLabel != "" {
-					configMapName = configMapName + "-" + revLabel
-				}
+				configMapName := IstioConfigMapName(in.conf, controlPlane.Revision)
```
jmazzitelli, line 158: the thing I don't see explained anywhere here is why does this pass in the Revision but everywhere else it passes in empty string for revision. How does this work when sometimes you need the revision but other times that is not required to get the configmap?

jmazzitelli (approved):
this does the refactor so everywhere in the code gets the configmap name using the common function. I have a feeling our support for canary upgrades still isn't seamless but we can worry about that later.

business/mesh.go
nrfox: Because basically this is the only place that is handling there being multiple controlplanes. Everywhere else just assumes there's one. When set `config.external_services.istio.config_map_name`, `IstioConfigMapName` will always return what you've set there regardless of what revision you pass in. It's a bit of a hack to get around the fact that, right now we expect users to set this manually when using control plane revisions.

nrfox:
> I have a feeling our support for canary upgrades still isn't seamless but we can worry about that later.

Yes this change doesn't make the support for canary upgrades any better but it at least gets rid of the error log when you've set `config.external_services.istio.config_map_name`.

amazingandyyy:
good to see this get merged, what's the ETA for new helm chart release? thanks for the hard work!

yohanb:
probably removing the logic to append the revision all together would be a more cohesive solution.  I think we can simply expect users to provide the correct values through `external_services.istio`

jmazzitelli:
> probably removing the logic to append the revision all together would be a more cohesive solution. I think we can simply expect users to provide the correct values through `external_services.istio`

I'm OK doing either suggestion. if Kiali users like @yohanb think it would be simpler and expected of users to set the correct value in the Kiali config, then let's do that.

cc @nrfox

nrfox:
> I'm OK doing either suggestion. if Kiali users like @yohanb think it would be simpler and expected of users to set the correct value in the Kiali config, then let's do that.

That sounds good to me. The only place that is currently using the offending code from `GetMesh` does not use the config map anyways and I'd rather be sure to not break existing functionality or cause noise for users. I appreciate the feedback here and it's very useful knowing that this is config that users rely on.

jmazzitelli:
we need to backport this to the 1.73 branch.

Issue #6677, opened 2023-10-05
Bug after v1.72.0 release with oAuth2 strategy when DisableRBAC is true


AlexDexsoon:
Look like there was a mistake in condition for this logic:

https://github.com/kiali/kiali/blob/master/kubernetes/client_factory.go#L224

The second part of condition, looks wrong:
```
		if (cfg.Auth.Strategy != kialiConfig.AuthStrategyAnonymous) ||
			(cfg.Auth.Strategy == kialiConfig.AuthStrategyOpenId && cfg.Auth.OpenId.DisableRBAC) {
			remoteConfig.BearerToken = authInfo.Token
		}
```
Should be something like:
```
		if (cfg.Auth.Strategy != kialiConfig.AuthStrategyAnonymous) &&
			(cfg.Auth.Strategy == kialiConfig.AuthStrategyOpenId &&  !cfg.Auth.OpenId.DisableRBAC) {
			remoteConfig.BearerToken = authInfo.Token
		}
``` 
No Goland developer, sorry.

AlexDexsoon:
@nrfox Can you help please? 
[@[nrfox](https://github.com/kiali/kiali/commits?author=nrfox)
nrfox](https://github.com/kiali/kiali/pull/6279)

nrfox:
You're right this was a bug introduced by that change. Thanks for opening the issue and providing a fix.

jshaughn:
@nrfox A fix should likely be backported to all subsequent versions?  Minimally, 1.73.

jmazzitelli:
Looks like this was already done. See PRs:
* https://github.com/kiali/kiali/pull/6679
* https://github.com/kiali/kiali/pull/6779

@nrfox Can you confirm if we can close this issue now?

nrfox:
> @nrfox Can you confirm if we can close this issue now?

Yes, closing

PR #6679, opened 2023-10-05
fix condition when kiali uses user token instead kiali SA

AlexDexsoon:
fix condition when kiali uses user token instead kiali SA

** Describe the change **

We should use user token only when both condition is true:
1) Auth strategy isn't anonymous
2) OpenID RBAC isn't disabled

** Issue reference **

[Bug after v1.72.0 release with oAuth2 strategy when DisableRBAC is true](https://github.com/kiali/kiali/issues/6677)

### DIFF
--- a/kubernetes/client_factory.go
+++ b/kubernetes/client_factory.go
@@ -219,10 +219,10 @@ func (cf *clientFactory) newClient(authInfo *api.AuthInfo, expirationTime time.D
 		}
 
 		// Replace the Kiali SA token with the user's auth token.
-		// In anonymous mode and when OpenID RBAC is disabled use the Kiali SA token
-		// since these modes don't have a user token.
-		if (cfg.Auth.Strategy != kialiConfig.AuthStrategyAnonymous) ||
-			(cfg.Auth.Strategy == kialiConfig.AuthStrategyOpenId && cfg.Auth.OpenId.DisableRBAC) {
+		// Only if we are not in an anonymous mode
+		// and if we don't use OpenID with RBAC is disable.
+		if !(cfg.Auth.Strategy == kialiConfig.AuthStrategyAnonymous) &&
+			!(cfg.Auth.Strategy == kialiConfig.AuthStrategyOpenId && cfg.Auth.OpenId.DisableRBAC) {
 			remoteConfig.BearerToken = authInfo.Token
 		}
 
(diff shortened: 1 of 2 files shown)


kubernetes/client_factory.go
```diff
@@ -219,10 +219,10 @@ func (cf *clientFactory) newClient(authInfo *api.AuthInfo, expirationTime time.D
...
 
 		// Replace the Kiali SA token with the user's auth token.
-		// In anonymous mode and when OpenID RBAC is disabled use the Kiali SA token
-		// since these modes don't have a user token.
-		if (cfg.Auth.Strategy != kialiConfig.AuthStrategyAnonymous) ||
-			(cfg.Auth.Strategy == kialiConfig.AuthStrategyOpenId && cfg.Auth.OpenId.DisableRBAC) {
+		// Only if we are not in an anonymous mode
+		// and in case we use OpenID with RBAC is enables.
+		if (cfg.Auth.Strategy != kialiConfig.AuthStrategyAnonymous) &&
+			(cfg.Auth.Strategy == kialiConfig.AuthStrategyOpenId && !cfg.Auth.OpenId.DisableRBAC) {
```
nrfox, line 225: The conditional (that I wrote) is kind of confusing here. We only want to use the Kiali SA token when 1. anonymous mode is enabled or 2. openid + rbac disabled. However the config returned from `cf.getConfig` has the SA client token already set so we need to replace with the user token in all but those two cases. That said, this now looks correct since we don't need to evaluate the right side if we're in anonymous mode and then only use the user token when openid is set and rbac is not disabled.

nrfox (commented):
Thanks for the contribution! Tested the fix and it works. Will you add a test for this as well? You can look at `TestClientCreatedWithClusterInfo` in `client_factory_test.go` for an example that's close to this. Basically you'd need to set `conf.Auth.Strategy = kialiConfig.AuthStrategyOpenId` and `cfg.Auth.OpenId.DisableRBAC = true` and then ensure that the user clients return the service account token.

kubernetes/client_factory.go
AlexDexsoon: Added tests

nrfox (approved):
Everything looks great! Thanks for both opening this issue and contributing a fix.

PR #6685, opened 2023-10-05
adding smoke test group

ScriptingShrimp:
Few scenarios and feature files are now annotated with `@smoke` identifier. Such tests does not require any additional apps deployed, so its a good idea to let them run in first place - to check overall health of application. 

Second modification is in package.json scripts, I added `@smoke` group to be triggered first. During this test the CI environment might get a bit more responsive since those tests are not demanding.

Most of those test also pass with Jaeger and Prometheus disabled


hhovsepy (approved):
agree on having smoke test set, could be added more tests ondemand

PR #6690, opened 2023-10-06
Fixed config names duplication to avoid potential flake

hhovsepy:
Names of configs created to be different.


josunect (approved):
LGTM 

Regarding failed tests, maybe you can rebase and probably they passed? CI looks more stable now.

PR #6691, opened 2023-10-06
support for OSSMC in the Kiali Operator

jmazzitelli:
- Kiali: https://github.com/kiali/kiali/pull/6691
- Kiali Operator: https://github.com/kiali/kiali-operator/pull/691
- Helm Charts: https://github.com/kiali/helm-charts/pull/226
- Kiali.io docs: https://github.com/kiali/kiali.io/pull/705
- OSSMC: https://github.com/kiali/openshift-servicemesh-plugin/pull/211

see: https://issues.redhat.com/browse/OSSM-4944

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,8 +49,10 @@ OPERATOR_QUAY_TAG ?= ${OPERATOR_QUAY_NAME}:${OPERATOR_CONTAINER_VERSION}
 
 # Where the control plane is
 ISTIO_NAMESPACE ?= istio-system
-# Declares the namespace/project where the objects are to be deployed.
+# Declares the namespace/project where the Kiali objects are to be deployed.
 NAMESPACE ?= ${ISTIO_NAMESPACE}
+# Declares the namespace/project where the OSSM Console objects are to be deployed.
+OSSMCONSOLE_NAMESPACE ?= ossmconsole
 
 # Local arch details needed when downloading tools
 OS := $(shell uname -s | tr '[:upper:]' '[:lower:]')
@@ -120,6 +122,7 @@ OPERATOR_WATCH_NAMESPACE ?= \"\"
 OPERATOR_INSTALL_KIALI ?= false
 OPERATOR_ALLOW_AD_HOC_KIALI_NAMESPACE ?= true
 OPERATOR_ALLOW_AD_HOC_KIALI_IMAGE ?= true
+OPERATOR_ALLOW_AD_HOC_OSSMCONSOLE_IMAGE ?= true
 ifeq ($(OPERATOR_WATCH_NAMESPACE),\"\")
 OPERATOR_INSTALL_KIALI_CR_NAMESPACE ?= ${OPERATOR_NAMESPACE}
 else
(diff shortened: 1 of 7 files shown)


jmazzitelli:
Here's the test procedures. This requires all the PRs to be checked out locally (not just this PR).

Setup:

- `hack/crc-openshift.sh start`
- `hack/istio/install-istio-via-istioctl.sh -c oc`
- From the openshift-servicemesh-plugin repo, run `make cluster-push` to get the plugin image into the cluster
- From the kiali/kiali repo, run `make build build-ui cluster-push` to get the server and operator images into the cluster
- If testing with the PRs, you need to `export HELM_CHARTS_REPO_PULL=false` before running some of the make targets described below.

Test procedures:

1. Install the operator and kiali server via `make operator-create kiali-create`
2. Confirm the Kiali server is running in istio-system.
3. Point browser to the OS Console: https://console-openshift-console.apps-crc.testing/
4. You should not see Service Mesh option on lefthand nav
5. Create OSSMConsole CR via `make ossmconsole-create`
6. Wait for the OS Console to pop up the message saying you need to refresh
7. After you refresh, you should see Service Mesh option on lefthand nav
8. Look at the OSSMConsole CR and the status field should show all the information (version #, kiali namespace and service port #, etc): `oc get ossmconsole -n ossmconsole -oyaml`
9. Confirm the plugin pod is running via `oc get pods -n ossmconsole -l app.kubernetes.io/name=ossmconsole`
10. Delete the OSSMConsole CR via `make ossmconsole-delete` and see that the OS Console asked to refresh and the Service Mesh lefthand nav goes away, and the pod in ossmconsole ns terminates.
11. Run `make ossmconsole-purge` just to make sure it finishes successfully (there is nothing to purge because the `make ossmconsole-delete` above removed everything).
12. Run `make run-operator-playbook-ossmconsole` and make sure it finishes successfully. This will install and then immediately uninstall the plugin. NOTE: this will only work if you have ansible installed locally.
13. Uninstall Kiali and the operator to clean out the cluster: `make operator-delete`
14. Start the operator locally via `make run-operator` - let it process all the existing namespaces (you’ll see some logs scroll by). NOTE: this will only work if you have ansible prerequisites installed.
15. In another command window, run `make kiali-create` to create a Kiali CR which installs Kiali. See that `oc get pods -n istio-system -l app.kubernetes.io/name=kiali` shows Kiali pod is running
16. In another command window, run `make ossmconsole-create` to create a OSSMConsole CR which installs the plugin. See that `oc get pods -n ossmconsole -l app.kubernetes.io/name=ossmconsole` shows the plugin pod is running. See that the OS Console shows the Service Mesh option in the lefthand nav.
17. Now uninstall the plugin and Kiali by removing the OSSMConsole CR and the Kiali CR via `make ossmconsole-delete kiali-delete`. You should see both Kiali and plugin pods terminate and the Service Mesh lefthand nav disappear from the OS Console.
18. You can kill the “make run-operator” process that was started earlier.
19. Run the molecule test and ensure it passes: `make MOLECULE_SCENARIO="ossmconsole-config-values-test" -e MOLECULE_USE_DEV_IMAGES="true" molecule-test`. You know it passed if the make exit code is 0 (`echo $?`). There also will not be an error message at the end.
20. In your kiali/operator local git repo, validate the OLM metadata via `make validate`
21. In your kiali/operator local git repo, validate the correctness of the Kiali CRD/example CR and the OSSMConsole CRD/example CR via `make validate-cr` - the make command should succeed and you should see two “is valid” log messages spit out to indicate both CRs are correct and the CRD schemas were valid.
23. Back in the kiali/kiali local repo, install the operator by OLM via `make olm-operator-create`
24. In the OS Console, go to Operators > Installed Operators and you should see the Kiali operator is installed.
25. See that the custom catalog pod and operator pod are both running in openshift-operators namespace via `oc get pods -n openshift-operators | grep kiali`
26. Just like before, run `make kiali-create`, see that Kiali pod starts in istio-system, and then `make ossmconsole-create` to install the plugin and confirm it shows up in the OS Console.
27. Confirm you have both CRDs installed via `oc get crds | grep kiali`
28. Uninstall the CRs and the operator via `make olm-operator-delete` - everything should be deleted - you should not have any Kiali CRDs and operator and catalog pods should be terminated from the openshift-operators namespace.

jshaughn (approved):
This took me a while but it also must have taken a while to write up all of those test instructions.  Good job, I was able to walk through it and it all worked.

Issue #6682, opened 2023-10-05
Service of a remote app/workload is not reported in the detail view


matejnesuta:
When in the primary remote setup, no service link is found both in the remote app detail and in the remote workload detail.

![image](https://github.com/kiali/kiali/assets/45061792/f374e0f9-7379-4156-bf1e-ad0fe13cd9f8)
![image](https://github.com/kiali/kiali/assets/45061792/b9136d2d-6709-47bf-afca-4c81f10d84cd)

However, both app and workload are reported from the detail page of a remote service. 

![image](https://github.com/kiali/kiali/assets/45061792/fa359410-d266-4c3a-9b9b-90fa74c81450)

PR #6700, opened 2023-10-10
Apply filter by service selector after loading from kube cache

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6682

For services loading from kube cache, does not mater local or remote, the label selector was filtering out the services which did not have `version` label selector.
In an contrary, the registry loading was loading all services then filtering by selector.
This was causing an issue for remote cluster services loaded for workload. The selector which contains `version` for workload was filtering out all services from cache. The local cluster's workload's services were loading from registry and were fine. Which was correct but still an Inconsistency with remote ones. 

Now when selector is provided for loading services from kube cache, it lists all services then filters out the selector matched ones, similar to loading from registry.
When no selector is provided, it loads all services as it was before.

Screenshot from 2023-10-10 14-21-32
Screenshot from 2023-10-10 14-21-21
Screenshot from 2023-10-10 14-21-06

### DIFF
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -720,6 +720,7 @@ func (c *kubeCache) GetStatefulSet(namespace, name string) (*apps_v1.StatefulSet
 	return retSet, nil
 }
 
+// GetServices returns list of services filtered by Spec.Selector instead of Metadata.Labels
 func (c *kubeCache) GetServices(namespace string, selectorLabels map[string]string) ([]core_v1.Service, error) {
 	// Read lock will prevent the cache from being refreshed while we are reading from the lister
 	// but it won't prevent other routines from reading from the lister.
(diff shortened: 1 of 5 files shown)


kubernetes/cache/kube_cache.go
```diff
@@ -729,21 +729,27 @@ func (c *kubeCache) GetServices(namespace string, selectorLabels map[string]stri
...
-		retServices = append(retServices, *svc)
+		svcSelector := labels.Set(service.Spec.Selector).AsSelector()
+		// selector match is done after listing all services, similar to registry reading
+		// empty selector is loading all services, or match the service selector
+		if selector.AsSelector().Empty() || (!svcSelector.Empty() && svcSelector.Matches(selector)) {
+			// Do not modify what is returned by the lister since that is shared and will cause data races.
+			svc := service.DeepCopy()
+			svc.Kind = kubernetes.ServiceType
+			retServices = append(retServices, *svc)
+		}
```
nrfox, line 752: This is basically a "field selector" now that selects the `service.Spec.Selector` which is not what you'd expect when you call `kubeCache.GetServices(...)`. It might have weird effects elsewhere. I think it'd be better to keep the `kubeCache.GetServices` method as is and have a separate function for filtering a slice of `[]core_v1.Service` by their `spec.Selector` matching a `map[string]string`.

hhovsepy: `kubeCache.GetServices` method will stay working wrongly for cases when `selectorLabels` is provided.

business/services.go
```diff
@@ -153,6 +153,14 @@ func (in *SvcService) getServiceListForCluster(ctx context.Context, criteria Ser
 			log.Errorf("Error fetching Services per namespace %s: %s", criteria.Namespace, err2)
 			errChan <- err2
 		}
+		// when services were not loaded, try once more for remote clusters only by filtering services by selector
+		if len(svcs) == 0 && len(selectorLabels) != 0 && cluster != in.config.KubernetesConfig.ClusterName {
+			svcs, err2 = kubeCache.GetAllServicesFilteredBySelector(criteria.Namespace, selectorLabels)
```
nrfox, line 158: Sorry what's confusing me here is what does `ServiceCriteria.ServiceSelector` mean? On every other `Get<Objects>` method the selector selects the labels of the object and not the `Spec.Selector` like `GetPods` which returns all pods that match the passed in label selector. It seems like all callers of `GetServices` actually want the `service.Spec.Selector` from where I can tell it's used in the workloads and apps services. Given that, I think what you had in the previous commit was fine and it's better than doing multiple calls here. Can you simply revert to the previous commit and add a comment to the `GetServices` method that when you pass `selectorLabels` you are actually filtering on `Spec.Selector` rather than `Metadata.Labels`?

hhovsepy: ok it's clear now, have reverted back

hhovsepy:
@nrfox PR is ready to review

nrfox (commented):
LGTM. Can you also provide a cypress test for this? It looks like there's already a feature file for `app_details_multicluster` and `workload_details_multicluster`. @matejnesuta can confirm but I think you can add a scenario to each that is something like "see service link". If the other scenarios aren't implemented yet then you may need to add the `@skip` annotation to them.

hhovsepy:
@nrfox have added a cypress checks.

josunect (approved):
LGTM

image

cypress test were added for multicluster.

nrfox (commented):
@hhovsepy sorry I missed this. Thanks for adding the cypress tests. Are they actually run though since they have this `@skip` annotation on them? https://github.com/kiali/kiali/blob/master/frontend/cypress/integration/featureFiles/service_details_multicluster.feature#L3. For example it looks like all of the tests have been skipped in this run: https://github.com/kiali/kiali/actions/runs/6510155092/job/17684923138.

I think you'll need to remove that `@skip` annotation at the top level and add it to each scenario block except the one you just implemented in this PR.

Issue #6699, opened 2023-10-10
Tempo - Update Trace data on hover 


josunect:
With Jaeger, all the trace information is in the traces Graph, so when hovering on one of the traces it is possible to see the heat map.
With Tempo, not all this information is there for a trace. At the moment, it is needed to click on one trace to fetch all the information for that. 
![image](https://github.com/kiali/kiali/assets/49480155/65150967-9538-42e3-b717-e6eaae1adadc)

Change the event to hover to avoid additional actions for the user.

PR #6707, opened 2023-10-11
Update trace data on hover 

josunect:
Fixes https://github.com/kiali/kiali/issues/6699


hhovsepy (commented):
@josunect code lgtm, the feature is very useful. Just found one issue, when hover over for the first time without clicking, it tries to load trace, very obvious in incognito window when empty Trace Details is shown:
Screenshot from 2023-10-17 20-22-04

josunect:
> @josunect code lgtm, the feature is very useful. Just found one issue, when hover over for the first time without clicking, it tries to load trace, very obvious in incognito window when empty Trace Details is shown: Screenshot from 2023-10-17 20-22-04

Thanks for catching this, @hhovsepy ! Updated now.

hhovsepy (approved):
Now all works fine.
The hover-over on a trace loaded the data in tooltip.
Trace tab does not load Trace Details by default.
Trace details are shown only when clicking on a trace.

jshaughn:
@josunect Can you make sure of one more thing here, please verify that as a user moves the mouse across the chart, that we do not launch concurrent requests to the backend.  Probably it is fine, but just to make sure...

josunect:
> @josunect Can you make sure of one more thing here, please verify that as a user moves the mouse across the chart, that we do not launch concurrent requests to the backend. Probably it is fine, but just to make sure...

With Tempo we don't have the all the trace data, if we want to load it on hover, we need to do a request to the backed - one per trace - in order to get all the trace information and be able to show that in the Tooltip. Not sure if concurrent, but , if there are many traces, it can perform many requests to the backend. 

The other alternative, the previous behavior, would be to keep the "Click to load" Message and just perform this request if the user request it...  But, it will look different.

Issue #6010, opened 2023-04-11
Kiali Crashing in sidecar validation (without a sidecar)


shardulsrivastava:
I have a kiali instance that crashes during sidecar validations for a service that doesn't even have a sidecar resource. From the code, it seems to be crashing at this [line](https://github.com/kiali/kiali/blob/master/business/checkers/sidecars/egress_listener_checker.go#L99) .

**Istio Version :** 1.17.2
**Kiali Version:** 1.61.2

## Screenshot
<img width="1512" alt="Screenshot 2023-04-11 at 02 28 10" src="https://user-images.githubusercontent.com/9951135/231024798-feeb91de-22df-48c2-911f-06f1741ea4d5.png">

## Logs
<details>
  <summary>Error Logs</summary>
Error logs:

```go
panic: runtime error: index out of range [1] with length 1

goroutine 378474 [running]:
github.com/kiali/kiali/business/checkers/sidecars.getHostComponents(...)
        /home/runner/work/kiali/kiali/business/checkers/sidecars/egress_listener_checker.go:99
github.com/kiali/kiali/business/checkers/sidecars.EgressHostChecker.validateHost({0xc00debd4a0, 0xc094f95cb0, {0xc022af2000, 0xb1d, 0xc00}}, {0xc033399f40, 0x38}, 0xc03b7c9f20?, 0x1dcc6ba?)
        /home/runner/work/kiali/kiali/business/checkers/sidecars/egress_listener_checker.go:64 +0x2d4
github.com/kiali/kiali/business/checkers/sidecars.EgressHostChecker.Check({0xc00debd4a0, 0xc094f95cb0, {0xc022af2000, 0xb1d, 0xc00}})
        /home/runner/work/kiali/kiali/business/checkers/sidecars/egress_listener_checker.go:33 +0x36e
github.com/kiali/kiali/business/checkers.SidecarChecker.runChecks({{0xc036db4000, 0x634, 0x6a0}, {0xc08f1cf2d0, 0xe, 0xe}, {0xc00ed21b00, 0x1e, 0x20}, 0xc03b7c9f20, ...}, ...)
        /home/runner/work/kiali/kiali/business/checkers/sidecars_checker.go:72 +0x415
[...]

PR #6708, opened 2023-10-12
Avoid potential runtime error

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6010

Unable to reproduce, the fix is just to avoid potential runtime error with panic.
Probably this will never happen.


jshaughn (approved):
LG!

Issue #6515, opened 2023-08-23
remove all usage / mention of kiali-bot


jmazzitelli:
There is a kiali-bot user that will soon be rendered unusable because 2FA is not enabled on it (and github is requiring users to all have 2FA enabled).

I don't think this is used anyway so we should get rid of it.

There is also a kiali-ui-bot that also should be removed.

jmazzitelli:
all mention of kiali-bot can be seen here: https://github.com/search?q=org%3Akiali%20kiali-bot&type=code

yuluo-yx:
Hi, team. I want to try. if it need to solve, pls assign to me.

jmazzitelli:
@yuluo-yx this task is already complete. There is one final [PR](https://github.com/kiali/kiali/pull/7037) that needs some additional research, but the work has been completed already.

jshaughn:
https://github.com/FreelancingTwin, Will you be able to finish the work in https://github.com/kiali/kiali/pull/7037?

yuluo-yx:
> @yuluo-yx这个任务已经完成了。最终的[PR](https://github.com/kiali/kiali/pull/7037)需要一些额外的研究，但这项工作已经完成。

got it.

yuluo-yx:
> @yuluo-yx this task is already complete. There is one final [PR](https://github.com/kiali/kiali/pull/7037) that needs some additional research, but the work has been completed already.

got it.

jmazzitelli:
this is done

jmazzitelli:
I am wrong - this isn't completely done. There is still some reviewing to be done here: https://github.com/kiali/kiali/pull/7037

PR #6709, opened 2023-10-12
maintainence: erased all mentions of kiali bot

FreelancingTwin:
** Describe the change **
-Erased all mentions of kiali bot.

_explanation of what it does_
-Just removed mentions of kiali-bot in the code

** Issue reference **
#6515

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -243,8 +243,6 @@ jobs:
       run: |
         git config user.email 'kiali-dev@googlegroups.com'
 
-        git config user.name 'kiali-bot'
-
     - name: Create tag
       run: |
         git add Makefile
--- a/frontend/getLabels.py
+++ b/frontend/getLabels.py
@@ -3,7 +3,6 @@ import sys, json;
 labels=''
 rawLabels=json.loads(sys.argv[1])['labels']
 for rLabel in rawLabels:
-  if rLabel['name'].startswith("kiali-bot"):
       if len(labels) > 0:
         labels+=','
       labels+=rLabel['name'].encode("utf-8")
--- a/.github/troubleshooting.md
+++ b/.github/troubleshooting.md
@@ -29,7 +29,7 @@ In case of errors in any of these steps, the log should be inspected to
 check what was the last action the pipeline executed. The pipelines are not idempotent so to retry them if there was a failure in the release step, manual amends should be done to prepare the next run:
 
 * Delete the tags and the release on GitHub (if exists)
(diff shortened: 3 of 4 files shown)


frontend/getLabels.py
```diff
@@ -3,7 +3,6 @@
 labels=''
 rawLabels=json.loads(sys.argv[1])['labels']
 for rLabel in rawLabels:
-  if rLabel['name'].startswith("kiali-bot"):
```
jmazzitelli, line 6: Do we even use `getLabels.py` anymore?  I thought we could just remove this. This technical debt may be from an old build system, but I'm not sure. I don't think it even makes sense just to remove this line, because in that case I'm not sure what the purpose of this script would then be.

.github/workflows/release.yml
```diff
@@ -243,8 +243,6 @@ jobs:
       run: |
         git config user.email 'kiali-dev@googlegroups.com'
 
-        git config user.name 'kiali-bot'
```
jmazzitelli, line 246: What would happen to the workflows now that git is not configured with a user.name? We need to know if GitHub Actions already define a generic username or if a git config user.name can be left undefined.

jmazzitelli: If it is OK to remove user.name config, then additional PRs need to be submitted for the other repos that have similar workflows with similar lines:
* https://github.com/kiali/kiali-operator/blob/72cee3f7903990ef26cfaba5e4454c51561fc9f7/.github/workflows/release.yml#L207
* https://github.com/kiali/kiali.io/blob/c5ff023a7528f49a0cd3280c70e4f97178efbbf6/.github/workflows/release.yaml#L113
* https://github.com/kiali/helm-charts/blob/544472b3e247801602ba49eaf01601e0e58e25d0/.github/workflows/release.yml#L168
* https://github.com/kiali/openshift-servicemesh-plugin/blob/033a693b4b5be5cb39ae73788dff2e8844df73d8/.github/workflows/release.yaml#L234

jmazzitelli (commented):
I have some questions.

jshaughn:
@FreelancingTwin , Any plan to bring this to completion?

.github/workflows/release.yml
FreelancingTwin: Sorry for the late reply.

 Apparently, user.name config is necessary to push to git usually. Otherwise the terminal would prompt an error to fill it in. But as kiali-bot is not in use, it'll never be staging or pushing to git. Seems it's safe to remove. 

I only just started with opensource, so, please correct me if I'm wrong or approaching the problem poorly. Thanks.

jmazzitelli:
this PR is superceded by https://github.com/kiali/kiali/pull/7037 so closing

PR #6713, opened 2023-10-13
Show Kiali links per cluster in graph returned from backend

hhovsepy:
Second commit for https://github.com/kiali/kiali/issues/6243 , not the last one, inaccessible Kiali is still TBD.

In Graph page, when selecting cluster box, on right panel the Kiali link is now shown in "namespace / name" format, similar to Mesh page:
Screenshot from 2023-10-13 13-33-21

### DIFF
--- a/frontend/src/pages/Graph/SummaryPanelClusterBox.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelClusterBox.tsx
@@ -14,9 +14,12 @@ import {
   getAccumulatedTrafficRateTcp
 } from 'utils/TrafficRate';
 import { PFBadge, PFBadges } from 'components/Pf/PfBadges';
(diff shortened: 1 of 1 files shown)


jshaughn (commented):
It looks fine, I do have one question...  The `KialiPageLink` component was used in only two places, here in `SummaryPanelClusterBox` and then also in `SummaryLink`.  I wonder if the logic you introduced should have been built into `KialiPageLink`?  If not, then `KialiPageLink` would be a single-use utility and could probably be deleted, and its logic moved into `SummaryLink`.

hhovsepy:
@jshaughn I left the same logic in 2 places for now because we will be going to redesign the Mesh page in future. Then we will come to a single solution if needed.

jshaughn (approved):
OK, we may still refactor a bit going forward, see the Q&A in the comments.  But this looks good for now.

PR #6724, opened 2023-10-16
[v1.65] Fix for wizard request routing test

nrfox:
Backport of https://github.com/kiali/kiali/commit/39ba5ba9b4a8d7e088ae3228f6f227729824a298#diff-4ae784ad0be35a26c3cedb2593703569ba1060a9a05344760d2919f28b0adb22


hhovsepy (approved):
This particular test could not cover KIA0301 validations, as it rely on another Gateway which exists by default and has 8080 port, so removing this check, will add another test for KIA0301 in wildcards.

Issue #6711, opened 2023-10-12
Envoy is duplicated across both clusters


matejnesuta:
Envoy tab is duplicated for both workloads (east cluster workload is in the left browser, west in the right one). However, there is no details workload in the west cluster. 
![image](https://github.com/kiali/kiali/assets/45061792/88db3e86-123d-49f4-b81c-440d433e5602)

hhovsepy:
Hi @matejnesuta, I was not able to reproduce this error, can you provide more info how did you setup your environment?

matejnesuta:
Hello Hayk. I used the `./hack/run-integration-tests.sh --test-suite frontend-multi-cluster -so true` to setup the KinD multicluster environment.

PR #6732, opened 2023-10-17
Read Namespaces per cluster

hhovsepy:
Found missing a `cluster` in reading namespaces while checking https://github.com/kiali/kiali/issues/6711

However the original issue could not be reproduced.

### DIFF
--- a/business/proxy_status.go
+++ b/business/proxy_status.go
@@ -68,7 +68,7 @@ func (in *ProxyStatusService) GetConfigDumpResourceEntries(cluster, namespace, p
 		return nil, err
 	}
 
-	namespaces, err := in.businessLayer.Namespace.GetNamespaces(context.TODO())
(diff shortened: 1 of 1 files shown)


hhovsepy:
@matejnesuta can you please check if the original issue still exists?

josunect (approved):
LGTM. 

Regarding envoy tab, I cannot see the tab when there is no proxy status: 

Easst: 
image

West: 
image


Not sure if the failing test is related, I was able to create HTTP routest (k8s) in both clusters.

matejnesuta:
The envoy tab is shown when I delete the cluster parameter from the URL and load the page. Maybe that was the reason why the issue happend in the first place? 

However the tab is now hidden when the workload is idle, which solves the issue above.

matejnesuta (approved):

Issue #6731, opened 2023-10-17
when I get workload  envoy metrics in multi cluster enviroment, I got err like this: "Could not fetch custom dashboard."


EdisonFu:
BUG describe：
curl  “https://*******.com/kiali/api/namespaces/travel-control/customdashboard/envoy?labelsFilters=app:control,version:v1&avg=true&duration=1800&step=36&rateInterval=36s&workload=control&cluster=cluster-test
Got response:  "error": "Cannot access namespace data: namespaces \"travel-control\" not found"

I have found  solution to  fix this bug, and hope to submit pre to fix this problem；
<img width="1737" alt="image" src="https://github.com/kiali/kiali/assets/22769471/b349e366-5359-422b-a7c3-3b1e99020c70">

jshaughn:
Hi @EdisonFu , We look forward to your PR!

jmazzitelli:
We are missing version of Kiali this happens on. Because this cluster param rename was done recently ... see https://github.com/kiali/kiali/issues/6328

aljesusg:
Needed the backport to 1.73 then we can close this one

ferhoyos:
Backport to 1.73 completed, closing the issue.

Issue #6328, opened 2023-07-03
Cluster and ClusterName params in the same API url


ferhoyos:
On workload miniGraph, both cluster and clusterName rest params appears in the API url to get the node graph elements:

![image](https://github.com/kiali/kiali/assets/122779323/5b8c9b14-901d-497f-a1ea-c1a3a7605102)

Cluster param need to be removed according to https://github.com/kiali/kiali/issues/6322.

PR #6737, opened 2023-10-18
Fixed getting cluster name from url

EdisonFu:
** Issue reference **
https://github.com/kiali/kiali/issues/6731

In multi-cluster mode, when browser wants to get workload's Envoy Metrics, it will send  request of "https://kiali.kiali.com/kiali/api/namespaces/travel-agency/customdashboard/envoy?labelsFilters=app:hotels,version:v1&avg=true&duration=1800&step=36&rateInterval=36s&workload=hotels-v1&cluster=cluster-test", 
In this URL, the cluster name parameter here is "cluster", but when the service handles the request to get cluster name, it 
 will use parameter "clusterName" in the URL .
Therefore, We need to add the following logic to obtain the value of cluster name when "clusterName" cannot be obtained. If take less than, to use KubernetesConfig.ClusterName.

### DIFF
--- a/frontend/src/components/TrafficList/TrafficDetails.tsx
+++ b/frontend/src/components/TrafficList/TrafficDetails.tsx
@@ -163,7 +163,8 @@ class TrafficDetailsComponent extends React.Component<TrafficDetailsProps, Traff
         const params = this.graphDataSource.fetchForServiceParams(
           this.props.duration,
           this.props.namespace,
-          this.props.itemName
+          this.props.itemName,
+          this.props.cluster
         );
         params.includeHealth = false;
         params.showSecurity = true;
@@ -187,7 +188,8 @@ class TrafficDetailsComponent extends React.Component<TrafficDetailsProps, Traff
         const params = this.graphDataSource.fetchForAppParams(
           this.props.duration,
           this.props.namespace,
-          this.props.itemName
+          this.props.itemName,
+          this.props.cluster
         );
         params.includeHealth = false;
         params.injectServiceNodes = false;
--- a/frontend/src/services/GraphDataSource.ts
+++ b/frontend/src/services/GraphDataSource.ts
@@ -310,17 +310,25 @@ export class GraphDataSource {
 
   // Some helpers
 
-  public fetchForApp = (duration: DurationInSeconds, namespace: string, app: string) => {
-    const params = this.fetchForAppParams(duration, namespace, app);
+  public fetchForApp = (duration: DurationInSeconds, namespace: string, app: string, cluster?: string) => {
+    const params = this.fetchForAppParams(duration, namespace, app, cluster);
     params.showSecurity = true;
     this.fetchGraphData(params);
   };
 
-  public fetchForAppParams = (duration: DurationInSeconds, namespace: string, app: string): FetchParams => {
+  public fetchForAppParams = (
+    duration: DurationInSeconds,
+    namespace: string,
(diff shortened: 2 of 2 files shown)


handlers/utils.go
```diff
@@ -140,6 +140,9 @@ func getBusiness(r *http.Request) (*business.Layer, error) {
 // and provides a default value if it's not present.
 func clusterNameFromQuery(queryParams url.Values) string {
 	cluster := queryParams.Get("clusterName")
+	if cluster == "" {
+		cluster = queryParams.Get("cluster")
```
jmazzitelli, line 144: This sounds familiar. We renamed this parameter a while ago because some environments (like OpenShift) have cluster as a hardcoded param that collide.

@nrfox i think this might mean we forgot other places to rename the parameter.

jmazzitelli:
Do not merge this yet. We renamed the cluster parameter a while ago - IIRC, we should never have a param called "cluster" now... if we do, something else might be broke and we need to fix it a different way.

jmazzitelli:
Yeah, see this: https://github.com/kiali/kiali/issues/6328

@ferhoyos did we miss more places where cluster needs to be renamed?

ferhoyos:
@josunect renamed all cluster params to clusterName in https://github.com/kiali/kiali/pull/6325. @EdisonFu what is the Kiali version you are using? It should be fixed from Kiali v1.71, unless we missed some place in the code to do the renaming.

EdisonFu:
@ferhoyos my kiali version is v1.74

josunect:
There was a PR that could be related? https://github.com/kiali/kiali/pull/6732

jmazzitelli:
> instead of changing the backend to support 'cluster' (if we do this we will have problems with OSSMC plugin in openshift)

Right. That's why I put a DNM label on this PR. I think this PR is the wrong solution. As @ferhoyos says, we need to fix the frontend - we should not be using "cluster" param anywhere so the server-side should never see it.

jmazzitelli:
However, that said, this PR could be useful if we change it. If the server side DOES see "cluster" - we can have this code log a noisy message (say with a WARN level message). Because if we do see "cluster", it means something is wrong on the front end.

EdisonFu:
I know!  Thanks!
So you will fix this bug on the front end, and what can I do for you?

jmazzitelli:
> and what can I do for you?

I think we can close this PR. I believe @nrfox or @josunect will be fixing this in the front-end (they can correct me if I'm wrong).

nrfox:
@EdisonFu if you'd like to update this PR or open a new one to fix on the front-end instead that'd be great! I can reproduce the issue in a multi-cluster deployment as well. This is related to what @ferhoyos mentioned above:

> Thanks @EdisonFu, I will try to reproduce the issue and check what is adding the 'cluster' param here. Anyway I think the solution is to fix the frontend to send 'clusterName' param instead of changing the backend to support 'cluster' (if we do this we will have problems with OSSMC plugin in openshift)

So if you take a look at where the custom dashboard request url is being created (probably somewhere in here: https://github.com/kiali/kiali/blob/master/frontend/src/services/Api.ts), you probably just need to change `cluster` --> `clusterName` in the query params. There should be other examples of this around the same place as well.

EdisonFu:
Thank you for your suggestions.
Now I have resubmitted the PR on [https://github.com/kiali/kiali/pull/6748](url).
Please review.
Thank you very much!

Issue #6738, opened 2023-10-18
[ci] need to fix CI script for running molecule tests on openshift


jmazzitelli:
I found a place I missed when integrating OSSMC into the Kiali operator - there is a CI hack script used to run nightly molecule tests on OpenShift that I forgot to change.

An option name changed from `--use-default-server-image` to `--use-default-images`, but the ci-openshift-molecule-tests.sh didn't get that change.

There is also a change that needs to be made to allow the OSSMConsole CR to be given a customizable spec.version. I will make that change here, also. This allows testing to be done with different spec.version values.

serverPR: https://github.com/kiali/kiali/pull/6739
operator PR: https://github.com/kiali/kiali-operator/pull/695

PR #6739, opened 2023-10-18
use the new --use-default-images option.

jmazzitelli:
also allow for custom spec.version to be set in created ossmconsole cr
fixes: https://github.com/kiali/kiali/issues/6738


jmazzitelli:
I tested this on my openshift cluster and all works.

jshaughn (approved):
LOK

PR #6741, opened 2023-10-18
Cypress Details page implementation for multicluster

matejnesuta:
Cypress implementation for skipped tests in `apps_details_multicluster.feature`, `service_details_multicluster.feature`  and `workloads_details_multicluster.feature`. I also added some additional scenarios and waiting for Prometheus.


nrfox (changes requested):
Thanks for adding the implementation. It's great to get more multi-cluster coverage. Most of my comments are around re-using step definitions across feature files. If we can re-use these rather than re-implement them for every feature file it'll make them much easier to maintain.

matejnesuta:
I moved the datatest ID to a different place.

Issue #6742, opened 2023-10-18
Add cluster_name support for run-kiali.sh hack script


jshaughn:
`run-kiali.sh` is not working well if Istio is installed with an unset cluster name.  In this scenario I had to locally update `hack/run-kiali-config-template.yaml` with:

```
kubernetes_config:
  cluster_name: "Kubernetes"
```

`run-kiali.sh` should handle the default cluster name scenario, or be configurable.

jshaughn:
I think what is happening is we may be querying Istio itself and it is probably returning an empty cluster_name.  In general we, Kiali, then default to "Kubernetes" (as does Istio for things like telemetry), but for this run-kiali scenario we were not inserting the default.  Setting the variable is working, though.

nrfox:
> I think what is happening is we may be querying Istio itself and it is probably returning an empty cluster_name. In general we, Kiali, then default to "Kubernetes" (as does Istio for things like telemetry), but for this run-kiali scenario we were not inserting the default. Setting the variable is working, though.

That sounds like a bug then. We should still be able to detect the cluster name even when running outside the cluster and we should still default to `Kubernetes` when we can't. Kiali "detects the cluster name" by looking at the istiod deployment and looking for a `CLUSTER_ID` env var. From what I've seen, this still gets set even when you do not provide a value for cluster id in the mesh config.

PR #6743, opened 2023-10-18
run-kiali.sh now works if Istio is installed with an unset cluster name

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6742

### DIFF
--- a/hack/run-kiali-config-template.yaml
+++ b/hack/run-kiali-config-template.yaml
@@ -42,6 +42,9 @@ external_services:
     in_cluster_url: "${TRACING_URL}"
     url: "${TRACING_URL}"
 
+kubernetes_config:
+  cluster_name: "${CLUSTER_NAME}"
+
 login_token:
   signing_key: "notsecure"
 
--- a/hack/run-kiali-config-template.yaml
+++ b/hack/run-kiali-config-template.yaml
@@ -42,9 +42,6 @@ external_services:
     in_cluster_url: "${TRACING_URL}"
     url: "${TRACING_URL}"
 
-kubernetes_config:
-  cluster_name: "${CLUSTER_NAME}"
-
 login_token:
   signing_key: "notsecure"
 
--- a/hack/run-kiali.sh
+++ b/hack/run-kiali.sh
@@ -74,6 +74,7 @@ DEFAULT_API_PROXY_HOST="127.0.0.1"
 DEFAULT_API_PROXY_PORT="8001"
 DEFAULT_CLIENT_EXE="kubectl"
 DEFAULT_COPY_CLUSTER_SECRETS="true"
+DEFAULT_CLUSTER_NAME="Kubernetes"
 DEFAULT_ENABLE_SERVER="true"
 DEFAULT_ISTIO_NAMESPACE="istio-system"
 DEFAULT_ISTIOD_URL="http://127.0.0.1:15014/version"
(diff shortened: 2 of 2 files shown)


jmazzitelli:
@josunect you use this script alot - can you test this and see if it breaks how you use it? Jay needed this set, but I wonder if this new default will break other environments?

jshaughn (dismissed):
This works for my setup where Istio is installed without a cluster name, and therefore defaults to "Kubernetes".  For users of `run-kiali.sh` that have a non-default cluster name, and wher (I assum) `run-kiali.sh` was working, they may need to now set the new `-cn` option.

jmazzitelli:
> I want to make sure that weren't not covering up a problem with cluster name handling here. I don't think you should need to set `cluster_name` when using `run-kiali.sh`.

I'll put a DNM label on here until you give it a thumbs up.

josunect:
> I want to make sure that weren't not covering up a problem with cluster name handling here. I don't think you should need to set `cluster_name` when using `run-kiali.sh`.

I think the issue was related that the istio cluster name was "" (In the istio install hack script). 
I will also test it.

josunect:
I have tested it and in the graph page I still have the same error message: 
image

When cluster_name is not set in the config. 
In the run-kiali I can see the cluster_name is set: 

image

If I set the cluster_name in the config I'm able to see the graph: 

image

EdisonFu:
I have met same question. Cluster name should be set into istio operator. 
Kiali read it from istio config.

josunect:
I've installed istio with the istio hack script: 

`istio/install-istio-via-istioctl.sh -c kubectl`

`MANIFEST_CONFIG_SETTINGS_TO_APPLY= --set hub=gcr.io/istio-release  --set values.meshConfig.enableAutoMtls=true --set values.pilot.env.ENABLE_NATIVE_SIDECARS=false  --set values.gateways.istio-egressgateway.enabled=true --set values.gateways.istio-ingressgateway.enabled=true --set values.global.meshID= --set values.global.multiCluster.clusterName= --set values.global.network= --set values.meshConfig.accessLogFile=/dev/stdout`

EdisonFu:
@josunect This error is due to kiali reading  metrics from prometheus, and prometheus get cluster name from istio. If istio's cluster name is not set, prometheus will set a default value "Kubernetes" for its metrics label. If the "Kubernetes" is not set in kiali, then the error will appear like what you got. 
All we need to do is match kiali's cluster name to istio's.

nrfox:
@josunect can you try with: https://github.com/kiali/kiali/pull/6746?

@jmazzitelli FWIW we can still add the cluster name opt to that run-kiali hack script but we shouldn't _need_ to do that. It might be nice to leave it off by default to test that the defaults work.


> @josunect This error is due to kiali reading metrics from prometheus, and prometheus get cluster name from istio. If istio's cluster name is not set, prometheus will set a default value "Kubernetes" for its metrics label. If the "Kubernetes" is not set in kiali, then the error will appear like what you got.
All we need to do is match kiali's cluster name to istio's.

@EdisonFu that's right only I believe that if you don't set `values.global.multiCluster.clusterName` then istio, not prometheus, will set a default value of `Kubernetes`. Kiali gets this by looking at the istiod deployment and looking for an env var named `CLUSTER_ID`. The steps are:

1. If `config.KuberentesConfig.ClusterName` has been set, use that since that is set by the user.
2. Try to auto-detect the cluster name from the istiod deployment's env var.
3. Fallback to `Kubernetes` if auto-detect fails.

Are you seeing problems with the cluster id not being set properly? If so, we can open a separate issue and discuss there to better track the issue/discussion. It could be that step 3 is too lenient and Kiali should fail to start if 1 and 2 fails.

jmazzitelli:
let me figure out how get this to only show up if that cmd line argument isn't specified

jmazzitelli:
> we can still add the cluster name opt to that run-kiali hack script but we shouldn't _need_ to do that. It might be nice to leave it off by default to test that the defaults work.

I changed this PR so that now the cluster_name setting is only put in the config IF the user sets a cluster name to run-kiali.sh via `--cluster-name`. It leaves it undefined in the config (like it was before) if you do not specify that option.

nrfox (approved):

PR #6744, opened 2023-10-18
[hack] When installing via istioctl, emit all traces

### LINKED ISSUE #6671: [Test Flake] See tracing info after selecting a trace
An example of action with this flake: https://github.com/kiali/kiali/actions/runs/6390379296/job/17343843690

There are no traces in the screenshot: 
image

nrfox:
Increases the sampling rate for tracing to 100. This will hopefully fix #6671 where traces are not being seen because the env is very slow. The kiali traffic generator that creates traffic for the e2e tests probably doesn't produce enough requests to overwhelm the jaeger deployment in the testing environment but the sampling rate can be lowered in the future if that happens.


josunect (approved):
LGTM!

image

Let's see what the CI says :)

PR #6746, opened 2023-10-18
Fix auto-detect of istio cluster name for "remote kiali" setup

nrfox:
Previously the auto-detect of the cluster name was skipped entirely when `config.InCluster = false`. Kiali can still auto-detect the cluster name in this case so there's no reason to skip it.

Related to #6742

### DIFF
--- a/kiali.go
+++ b/kiali.go
@@ -260,6 +260,7 @@ func updateConfigWithIstioInfo() {
 		homeCluster = config.DefaultClusterID
 	}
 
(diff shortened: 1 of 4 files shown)


josunect:
This fixed my environment when cluster_name was ""

jmazzitelli (dismissed):
I approve.
go linter does not. :)

jmazzitelli (approved):

jshaughn (approved):
This fixed the run-kiali.sh situation for me, when using unset/defauit cluster for my istio install.

PR #6209, opened 2023-05-26
[DRAFT] Kiali6198

jshaughn:

PR #6747, opened 2023-10-18
Fix issues with gateway-badging in graph

### LINKED ISSUE #6740: Gateway badge is not being applied to gateways in the graph
It looks like the server is not reporting the Gateway hosts, which makes the UI think it's not a gateway.  Looks like this has been an issue for several releases...

jshaughn:
Fixes #6740 

As far as I can tell this has been broken since Kiali v1.69, due to the multi-cluster work in https://github.com/kiali/kiali/pull/6209.  That is a server-side problem.  But then there was a second problem introduced with the PF5 update, which changed the name of the icon we are using for our gateway badge.

We should backport this to 1.73 (careful of the icon name changes, those may not be necessary).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -55,6 +55,8 @@ operator
 # Do not commit helm-charts code to this repo
 helm-charts
 
+# Ignore ansible if a dev has installed it under kiali for convenience
+ansible
 
 ############################################
 #                FRONTEND                  #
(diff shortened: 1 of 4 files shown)


ferhoyos:
@jshaughn this has nothing to do with this issue but the rest of icons in `frontend/src/config/Icons.ts` have not been migrated to PF5.

For example:
```
mirroring: {
      className: 'pf-icon pf-icon-migration',
      icon: MigrationIcon,
      name: 'migration',
      text: 'Mirroring',
      type: 'pf'
    } as IconType,
```

jshaughn:
> @jshaughn this has nothing to do with this issue but the rest of icons in frontend/src/config/Icons.ts have not been migrated to PF5.

@ferhoyos As it turns out, three other badge icons have been lost or have changed class names for PF5.  I will see if I can fix those as well...

hhovsepy (approved):
lgtm, tested locally:
before:
Screenshot from 2023-10-20 18-13-34

after:
Screenshot from 2023-10-20 18-14-30

PR #6748, opened 2023-10-19
Fixed getting cluster name from url on frontend

EdisonFu:
** Issue reference **
https://github.com/kiali/kiali/issues/6731

** First commit  **
https://github.com/kiali/kiali/pull/6737

Thank you for your suggestions. 
Now I have resubmitted the PR to fix this problem, please review my code! 
I fix this issue on frontend now.

### DIFF
--- a/frontend/src/types/MetricsOptions.ts
+++ b/frontend/src/types/MetricsOptions.ts
@@ -17,7 +17,6 @@ export interface DashboardQuery extends MetricsQuery {
   additionalLabels?: string;
   workload?: string;
   workloadType?: string;
-  clusterName?: string;
 }
 
 export type Aggregator = 'sum' | 'avg' | 'min' | 'max' | 'stddev' | 'stdvar';
--- a/frontend/src/types/MetricsOptions.ts
+++ b/frontend/src/types/MetricsOptions.ts
@@ -17,7 +17,7 @@ export interface DashboardQuery extends MetricsQuery {
   additionalLabels?: string;
   workload?: string;
   workloadType?: string;
-  cluster?: string;
+  clusterName?: string;
 }
 
 export type Aggregator = 'sum' | 'avg' | 'min' | 'max' | 'stddev' | 'stdvar';
--- a/frontend/src/components/Metrics/CustomMetrics.tsx
+++ b/frontend/src/components/Metrics/CustomMetrics.tsx
@@ -162,8 +162,7 @@ class CustomMetricsComponent extends React.Component<Props, MetricsState> {
     // Workload name can be used to find personalized dashboards defined at workload level
     this.options.workload = this.props.workload;
     this.options.workloadType = this.props.workloadType;
-    this.options.cluster = this.state.cluster;
-    API.getCustomDashboard(this.props.namespace, this.props.template, this.options)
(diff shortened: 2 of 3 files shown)


ferhoyos:
Thanks for the PR @EdisonFu! Now I see the issue. `params` of `getCustomDashboard` is of type `DashboardQuery` which contains attribute `cluster` instead of `clusterName`.

There are two ways to fix this issue:
- Rename `cluster` to `clusterName`in `DashboardQuery` type
- Remove `cluster` in `DashboardQuery`type and add the cluster in API `getCustomDashboard` method (done in this PR).

To be consistent with other Kiali APIs and avoid future errors I prefer the second option (for example we have the same issue in `SummaryPanelNamespaceBox`:
```
promiseIn = API.getNamespaceMetrics(namespace, {
        byLabels: ['request_protocol'], // ignored by prom if it doesn't exist
        cluster: cluster,
        direction: 'inbound',
        duration: props.duration,
        filters: filters,
        queryTime: props.queryTime,
        rateInterval: props.rateInterval,
        reporter: 'destination',
        step: props.step
      } as IstioMetricsOptions);
```

@EdisonFu, your PR is incomplete. Pending things:
- Remove `cluster` from `DashboardQuery` type
- Add cluster param in all `getCustomDashboard` API calls (there is only one in `CustomMetrics.tsx`):
```
 private fetchMetrics = () => {
    // Time range needs to be reevaluated everytime fetching
    MetricsHelper.timeRangeToOptions(this.props.timeRange, this.options);
    // Workload name can be used to find personalized dashboards defined at workload level
    this.options.workload = this.props.workload;
    this.options.workloadType = this.props.workloadType;
    this.options.cluster = this.state.cluster;
    API.getCustomDashboard(this.props.namespace, this.props.template, this.options)
```

ferhoyos:
@EdisonFu you have half work done in this PR. I think it is better to modify this one instead of creating new PR.

EdisonFu:
@ferhoyos Yeah, I thought so

EdisonFu:
@ferhoyos I have committed the rest of the changes. Please review my commits. 
Thank you very much!

frontend/src/types/MetricsOptions.ts
```diff
@@ -17,7 +17,7 @@ export interface DashboardQuery extends MetricsQuery {
   additionalLabels?: string;
   workload?: string;
   workloadType?: string;
-  cluster?: string;
+  clusterName?: string;
```
ferhoyos, line 20: Don't change MetricsOptions file, this is another issue (https://github.com/kiali/kiali/issues/6753). What you have to do is to remove clusterName attribute from DashboardQuery type.

ferhoyos (changes requested):

frontend/src/types/MetricsOptions.ts
EdisonFu: Got it.
Done.

ferhoyos (approved):
LGTM. Tested locally, the Metrics dashboard is shown and clusterName query param is sent to backend.

image

image

EdisonFu:
You're welcome! It's my pleasure to contribute to the Kiali project. I'm glad that my contributions have been helpful.

Issue #6710, opened 2023-10-12
Traces tab is duplicated across both clusters 


matejnesuta:
Traces are duplicated for both workloads (east cluster workload is in the left browser, west in the right one). However, there is no details workload in the west cluster. Same behaviour can be seen for apps and services as well.
![image](https://github.com/kiali/kiali/assets/45061792/8112aa0a-5cff-4103-953a-199168c8a51a)

jshaughn:
Maybe same issue as https://github.com/kiali/kiali/issues/6711

josunect:
Working in Tempo fix

PR #6749, opened 2023-10-19
Hide Traces for remote cluster's Workload/Service/App object details

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6710

As the 'Traces' for multicluster is not implemented yet, and it is not possible to filter in tracing service by cluster, hiding the 'Traces' tab for Service/Workload/Application details pages for remote clusters, as they were duplicating the info from home cluster.

### DIFF
--- a/frontend/src/pages/AppDetails/AppDetailsPage.tsx
+++ b/frontend/src/pages/AppDetails/AppDetailsPage.tsx
@@ -24,6 +24,7 @@ import { ErrorSection } from '../../components/ErrorSection/ErrorSection';
 import { connectRefresh } from '../../components/Refresh/connectRefresh';
 import { history, HistoryManager } from 'app/History';
 import { basicTabStyle } from 'styles/TabStyles';
+import { isHomeCluster } from '../../config/ServerConfig';
 
 type AppDetailsState = {
   app?: App;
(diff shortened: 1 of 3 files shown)


josunect:
Should we hide the traces for remote clusters? It is possible to see traces from the remote clusters (Example from error-rates-demo with option -dd true, where beta is installed in the remote cluster):  

image

I think the problem would be, when there is the same application/workload in both clusters and namespaces. But the "incorrect" traces would be seen in both cluster (Local and remote). It could seem like duplicate information, but that could be because the trace is involved in both clusters.

leandroberetta:
I agree with @josunect, I think we can have traces from both clusters, and probably a trace will span multiple clusters. How are we checking that are duplicates?. 

Acoording to the picture, traces are the "same", but, maybe we have an issue where we gathere traces from all clusters and we need to start adding a cluster filter. I'm not sure our hack scripts managed that scenario with tracing.

leandroberetta:
Checking, we have something at https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/setup-tracing.sh

That scripts combines both cluster's traces

josunect:
> Checking, we have something at https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/setup-tracing.sh
> 
> That scripts combines both cluster's traces

Ahh right, so that script combines them, but, not sure, if it would be possible to add some additional tag? 
Because, in Jaeger, we can see the traces, but I don't think if they have any attribute to be filtered by the cluster. 
This is an example of a Jaeger trace in a MC environment: 

image

And this is from a remote cluster:

leandroberetta:
I think we can, with open telemetry:

https://istio.io/latest/docs/reference/config/telemetry/

```
kind: Telemetry
metadata:
  name: mesh-default
  namespace: istio-system
spec:
  tracing:
  - randomSamplingPercentage: 10.00
    customTags:
      my_new_foo_tag:
        literal:
          value: "foo"

```

jshaughn:
If we have a unified trace store from multiple clusters then I would think those traces would need a `cluster` tag added so that we could discriminate based on cluster.  Without that tag I don't think we can support a unified trace store.  This may need to be documented.

Note that there are two scenarios here, there is the scenario where we WANT Kiali to query cross-cluster, and the tenancy scenario where we don't.  In the tenancy scenario you still need some sort of `cluster` tag so that we can use it with a `query_scope` config.

hhovsepy:
Closing this PR, the correct solution is in WIP https://github.com/kiali/kiali/pull/6775

Issue #6751, opened 2023-10-19
istio hack script: cluster name should not be set to empty string


josunect:
If no cluster name is specified, this is set to "".

`--set values.global.multiCluster.clusterName=""`

The cluster parameter should not be set if no parameters are specified (Or, should be set with another default).

PR #6752, opened 2023-10-19
istio hack script: cluster name should not be set to empty string

josunect:
Fixes https://github.com/kiali/kiali/issues/6751

### DIFF
--- a/hack/istio/install-istio-via-istioctl.sh
+++ b/hack/istio/install-istio-via-istioctl.sh
@@ -425,16 +425,21 @@ if [ "${REDUCE_RESOURCES}" == "true" ]; then
     --set components.pilot.k8s.resources.requests.memory=1Mi"
 fi
 
+if [ "${CLUSTER_NAME}" != "" ]; then
+     CLUSTER_NAME_OPTION=" \
+       --set values.global.multiCluster.clusterName=${CLUSTER_NAME}"
+   fi
+
 for s in \
    "${IMAGE_HUB_OPTION}" \
    "${IMAGE_TAG_OPTION}" \
    "${MTLS_OPTIONS}" \
    "${NATIVE_SIDECARS_OPTIONS}" \
+   "${CLUSTER_NAME_OPTION}" \
    "${CUSTOM_NAMESPACE_OPTIONS}" \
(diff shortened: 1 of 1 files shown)


hack/istio/install-istio-via-istioctl.sh
```diff
@@ -425,16 +425,21 @@ if [ "${REDUCE_RESOURCES}" == "true" ]; then
     --set components.pilot.k8s.resources.requests.memory=1Mi"
 fi
 
+if [ "${CLUSTER_NAME}" != "" ]; then
+     CLUSTER_NAME_OPTION=" \
+       --set values.global.multiCluster.clusterName=${CLUSTER_NAME}"
+   fi
```
jmazzitelli, line 431: Just some reformatting.. The "fi" is indented incorrectly and no need for a new line in the cluster name option value.
```suggestion
if [ "${CLUSTER_NAME}" != "" ]; then
  CLUSTER_NAME_OPTION="--set values.global.multiCluster.clusterName=${CLUSTER_NAME}"
fi
```

jmazzitelli (changes requested):
some formatting things, but the code looks ok

jmazzitelli:
I took a look at when we changed this (in the past, we always had a cluster name, I wondered when it changed to default to empty string)... the change was here: https://github.com/kiali/kiali/commit/5733e3e675d9e71c68821e209dece7815257d822

@josunect I think you should do the same fix you did here for the other two that are now defaulting to empty string - specifically NETWORK and MESH_ID. Both used to always be set, but now are empty strings.

josunect:
Thanks for pointing at this, @jmazzitelli 
I've applied the same to MESH_ID and NETWORK. 

I was doing some test and doesn't look like there is an additional cluster indicators for local cluster: 

image

I still have the problem not setting the cluster_name in the config. This seems to be fixed by: https://github.com/kiali/kiali/pull/6746

So, not sure if this change have a lot of impact. Said that, I guess it would prevent to set an empty string instead of using the defaults.

jmazzitelli (approved):
I like this PR better than how it is working today. Rather than explicitly setting things to empty string, I would rather we not set anything and fallback to default Istio behavior.

So I'm +1 on this. approving.

jshaughn (approved):
+1 to this change, not explcitly setting to `""` makes sense, even if it doesn't change the behavior.

Issue #6736, opened 2023-10-18
Replace strings with correct concatenation


aljesusg:
There are multiple sentences in code with old concatenation, replace them with the new one \`\<string\>\` (**Require template literals instead of string concatenation**):

## Rule:

In ES2015 (ES6), we can use template literals instead of string concatenation.

```ts
var str = "Hello, " + name + "!";
```

```ts
var str = `Hello, ${name}!`;
```

## Example in **kiali** code:

https://github.com/kiali/kiali/blob/efd78c68454efdca5e0baca87d5166766f254163/frontend/src/types/Health.ts#L276

that would be great to replace with
```ts
text: `${prefix}:${thresholdStatus.status === NA ? 'No requests' : `${thresholdStatus.value.toFixed(2)}%`}`
```

More information: http://eslint.org/docs/rules/prefer-template

jshaughn:
I agree that the new way is better, and I would encourage the change as we work on the code, but I'm -1 for a straight refactor unless there is some benefit other than just nicer looking code.  I do already change this when I see it, but you have to be careful not to make a mistake.

aljesusg:
Well...the point is that I am doing these changes for backstage effort, I must do it by repo rules (logically methods and some components with material-ui instead pf are identical - copy/paste).

I agree that it makes no sense make all the changes and I opened the issue looking for a contributor. If you want we can close it.

jmazzitelli:
I believe this is getting taken care of with the linter stuff we now have in  place. I'm seeing PRs that are getting created with these changes. So this issue may be getting implemented as time marches forward and the linter forces devs to make these changes.

PR #6756, opened 2023-10-20
Avoid cluster in API query params, only clusterName

### LINKED ISSUE #6753: getNamespaceMetrics includes cluster in query params
Reviewing https://github.com/kiali/kiali/issues/6731, I have noticed that API getNamespaceMetrics sends cluster as query param, which breaks the `clusterName` query param convention to avoid https://github.com/kiali/kiali/issues/6322.

```
promiseIn = API.getNamespaceMetrics(namespace, {
        byLabels: ['request_protocol'], // ignored by prom if it doesn't exist
        cluster: cluster,
        direction: 'inbound',
        duration: props.duration,
        filters: filters,
        queryTime: props.queryTime,
        rateInterval: props.rateInterval,
        reporter: 'destination',
        step: props.step
      } as IstioMetricsOptions);
```

Effectively, if you hit the namespace label in the graph, cluster query param is sent:

image

image

ferhoyos:
**Describe the change**

API getNamespaceMetrics is modified to send clusterName query param instead of cluster:

image

To avoid the use of cluster param in the future, I have added types to query params of all API methods (too many anys in Api.ts). Besides, cluster should be a parameter in the API method, not an attribute in the query params type.

For this reason I have removed cluster attribute from IstioMetricsOptions interface:
```
export interface IstioMetricsOptions extends MetricsQuery {
  direction: Direction;
  filters?: string[];
  requestProtocol?: string;
  reporter: Reporter;
}

export const getNamespaceMetrics = (namespace: string, params: IstioMetricsOptions, cluster?: string) => {
  const queryParams: QueryParams<IstioMetricsOptions> = { ...params };
  if (cluster) {
    queryParams.clusterName = cluster;
  }
  return newRequest<Readonly<IstioMetricsMap>>(HTTP_VERBS.GET, urls.namespaceMetrics(namespace), queryParams, {});
};
```

Bonus:
- Alphabetized props lists to modified files
- Replace strings with correct concatenation to files modified (https://github.com/kiali/kiali/issues/6736)

**Issue reference**

Fixes #6753

### DIFF
--- a/frontend/src/pages/Graph/SummaryPanelNamespaceBox.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelNamespaceBox.tsx
@@ -192,26 +192,30 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
           <br />
           {this.renderTopologySummary(numSvc, numWorkloads, numApps, numVersions, numEdges)}
         </div>
+
         <div className={summaryBodyTabs}>
-          <SimpleTabs id="graph_summary_tabs" defaultTab={0} style={{ paddingBottom: '10px' }}>
+          <SimpleTabs id="graph_summary_tabs" defaultTab={0} style={{ paddingBottom: '0.5rem' }}>
             <Tooltip
               id="tooltip-inbound"
               content="Traffic entering from another namespace."
               entryDelay={1250}
               triggerRef={tooltipInboundRef}
             />
+
             <Tooltip
               id="tooltip-outbound"
               content="Traffic exiting to another namespace."
               entryDelay={1250}
               triggerRef={tooltipOutboundRef}
             />
+
             <Tooltip
               id="tooltip-total"
               content="All inbound, outbound and internal namespace traffic."
               entryDelay={1250}
               triggerRef={tooltipTotalRef}
             />
+
             <Tab style={summaryFont} title="Inbound" eventKey={0} ref={tooltipInboundRef}>
               <div style={summaryFont}>
                 {grpcIn.rate === 0 && httpIn.rate === 0 && tcpIn.rate === 0 && (
@@ -219,6 +223,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                     <KialiIcon.Info /> No inbound traffic.
                   </>
                 )}
+
                 {grpcIn.rate > 0 && (
                   <RateTableGrpc
                     isRequests={isGrpcRequests}
@@ -227,6 +232,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                     rateNR={grpcIn.rateNoResponse}
                   />
                 )}
+
                 {httpIn.rate > 0 && (
                   <RateTableHttp
                     title="HTTP (requests per second):"
@@ -237,6 +243,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                     rateNR={httpIn.rateNoResponse}
                   />
                 )}
+
                 {tcpIn.rate > 0 && <RateTableTcp rate={tcpIn.rate} />}
                 {
                   // We don't show a sparkline here because we need to aggregate the traffic of an
@@ -244,6 +251,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                 }
               </div>
             </Tab>
+
             <Tab style={summaryFont} title="Outbound" eventKey={1} ref={tooltipOutboundRef}>
               <div style={summaryFont}>
                 {grpcOut.rate === 0 && httpOut.rate === 0 && tcpOut.rate === 0 && (
@@ -251,6 +259,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                     <KialiIcon.Info /> No outbound traffic.
                   </>
                 )}
+
                 {grpcOut.rate > 0 && (
                   <RateTableGrpc
                     isRequests={isGrpcRequests}
@@ -259,6 +268,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                     rateNR={grpcOut.rateNoResponse}
                   />
                 )}
+
                 {httpOut.rate > 0 && (
                   <RateTableHttp
                     title="HTTP (requests per second):"
@@ -269,6 +279,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                     rateNR={httpOut.rateNoResponse}
                   />
                 )}
+
                 {tcpOut.rate > 0 && <RateTableTcp rate={tcpOut.rate} />}
                 {
                   // We don't show a sparkline here because we need to aggregate the traffic of an
@@ -276,6 +287,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                 }
               </div>
             </Tab>
+
             <Tab style={summaryFont} title="Total" eventKey={2} ref={tooltipTotalRef}>
               <div style={summaryFont}>
                 {grpcTotal.rate === 0 && httpTotal.rate === 0 && tcpTotal.rate === 0 && (
@@ -283,6 +295,7 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
                     <KialiIcon.Info /> No traffic.
                   </>
                 )}
+
                 {grpcTotal.rate > 0 && (
                   <RateTableGrpc
                     isRequests={isGrpcRequests}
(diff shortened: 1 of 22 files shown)


ferhoyos:
I have noticed that `getAggregateMetrics` and `getNamespaceMetrics` (this one in `SummaryPanelGraph`) do not send cluster param in API request. @josunect , @hhovsepy  could you confirm that this is correct? Thanks

getAggregateMetrics:
```
case NodeMetricType.AGGREGATE:
      return API.getAggregateMetrics(nodeData.namespace, nodeData.aggregate!, nodeData.aggregateValue!, options);
    case NodeMetricType.APP:
      return API.getAppMetrics(nodeData.namespace, nodeData.app!, options, nodeData.cluster);
    case NodeMetricType.SERVICE:
      return API.getServiceMetrics(nodeData.namespace, nodeData.service!, options, nodeData.cluster);
    case NodeMetricType.WORKLOAD:
      return API.getWorkloadMetrics(nodeData.namespace, nodeData.workload!, options, nodeData.cluster);
```

getNamespaceMetrics in SummaryPanelGraph:
```
promiseIn = API.getNamespaceMetrics(namespace, {
        byLabels: ['request_protocol'], // ignored by prom if it doesn't exist
        direction: 'inbound',
        duration: props.duration,
        filters: filters,
        queryTime: props.queryTime,
        rateInterval: props.rateInterval,
        reporter: 'destination',
        step: props.step
      } as IstioMetricsOptions);
```

frontend/src/services/Api.ts
```diff
@@ -589,49 +640,50 @@ export const getNodeGraphElements = (node: NodeParamsType, params: any) => {
...
         HTTP_VERBS.GET,
         urls.workloadGraphElements(node.namespace.name, node.workload),
-        params,
+        queryParams,
         {}
       );
     default:
       // default to namespace graph
-      return getGraphElements({ namespaces: node.namespace.name, ...params });
+      return getGraphElements({ ...params, namespaces: node.namespace.name });
```
hhovsepy, line 670: getGraphElements does not accept 'cluster' as a second param?

jshaughn: The graph does not limit to a specific cluster.  Only namespaces.  Graph will always query across cluster, to limit to a specific cluster the user can narrow things in the UI.

hhovsepy (commented):
One small issue in Overview page, when changing the health interval:
Screenshot from 2023-10-23 14-37-11

ferhoyos:
> One small issue in Overview page, when changing the health interval: Screenshot from 2023-10-23 14-37-11

Thanks @hhovsepy, small typo in Api.ts. I have fixed it so you can continue the review.

hhovsepy (approved):
LGTM, tested locally.

frontend/src/services/Api.ts
```diff
@@ -494,24 +534,26 @@ export const getNamespaceServiceHealth = (
   );
 };
 
-export const getNamespaceWorkloadHealth = (
+export const getNamespaceWorkloadHealth = async (
```
jshaughn, line 537: I guess adding `async` here is mainly declarative?  I suppose we could add `async` to almost all of the functions in here, as even the ones that don't have explicit Promise return values are still returning promises (via `return newRequest`)

frontend/src/services/Api.ts
```diff
@@ -567,18 +609,27 @@ export const getTrace = (idTrace: string) => {
   return newRequest<TracingSingleResponse>(HTTP_VERBS.GET, urls.tracingTrace(idTrace), {}, {});
 };
 
-export const getGraphElements = (params: any) => {
-  return newRequest<GraphDefinition>(HTTP_VERBS.GET, urls.namespacesGraphElements, params, {});
+export const getGraphElements = (params: GraphElementsQuery, cluster?: string) => {
```
jshaughn, line 612: I don't think this signature should be exported.  The `cluster?` param should not exist here, I think it's here only because you want to call this from `getNodeGraphElements`.  I'd suggest you make this a private function and introduce a thin exported function that does not take the `cluster?` param, and just then calls the private function.

frontend/src/services/GraphDataSource.ts
```diff
@@ -293,10 +302,11 @@ export class GraphDataSource {
     }
 
     this.emit('loadStart', isPreviousDataInvalid, fetchParams);
+
     if (fetchParams.node) {
-      this.fetchDataForNode(restParams);
+      this.fetchDataForNode(restParams, cluster);
     } else {
-      this.fetchDataForNamespaces(restParams);
+      this.fetchDataForNamespaces(restParams, cluster);
```
jshaughn, line 309: This should never take `cluster` see my comment above about removing this public signature.

frontend/src/services/GraphDataSource.ts
```diff
@@ -468,16 +492,18 @@ export class GraphDataSource {
     this.eventEmitter.emit(eventName, ...args);
   };
 
-  private fetchDataForNamespaces = (restParams: any) => {
+  private fetchDataForNamespaces = (restParams: GraphElementsQuery, cluster?: string) => {
```
jshaughn, line 495:
```suggestion
  private fetchDataForNamespaces = (restParams: GraphElementsQuery) => {
```

jshaughn (changes requested):
Good stuff.  I love the large amount of increased type safety and consistent handling of the `clusterName` param.  Also love the code cleanup and alphabetizing.  Only one small requested change before merge.

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -348,15 +348,15 @@ export class OverviewPageComponent extends React.Component<OverviewProps, State>
       });
   }
 
-  fetchHealthChunk(chunk: NamespaceInfo[], duration: DurationInSeconds, type: OverviewType) {
+  async fetchHealthChunk(chunk: NamespaceInfo[], duration: DurationInSeconds, type: OverviewType) {
```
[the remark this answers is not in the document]
ferhoyos: No, it is not necessary. I forgot to mention it in the refactoring bonus section of the PR. Code editor did the suggestion since it returns a promise and I just add the word for cleaner syntax.

frontend/src/services/Api.ts
ferhoyos: The code editor suggested to add async when the method returns `Promise<XX>`but not when the method returns `Promise<Response<XX>>`. I don't know why. You are right, this async should be done in all methods since they are returning promises. I also do not like that we use our custom `Response` type in some methods and `AxiosResponse` in others. The same happens with `AxiosError`, we should have our own custom `Error` type like we have with `Response`. Doing all this stuff is out of scope of this PR (I have refactored enough I think), so I will create an issue to handle this and finish the refactorization of Kiali API in the frontend.

frontend/src/pages/Graph/SummaryPanelNamespaceBox.tsx
```diff
@@ -610,28 +610,34 @@ export class SummaryPanelNamespaceBox extends React.Component<SummaryPanelPropTy
     }
 
     if (filters.length > 0) {
-      promiseIn = API.getNamespaceMetrics(namespace, {
-        byLabels: ['request_protocol'], // ignored by prom if it doesn't exist
-        cluster: cluster,
```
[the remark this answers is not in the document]
ferhoyos: It seems that when you set `{ ... } as <Type>`, you are forcing the object to be that type, disabling the property validations of that type. If you just remove the `as <Type>` (in this case `as IstioMetrics` in the parameter) the warning would have appeared because the cluster property does not exist. Something new to learn about Typescript :-).

frontend/src/services/Api.ts
ferhoyos: Ok, so the default getGraphElements method is cluster-wide. In that case I will remove the cluster param from the method.

frontend/src/pages/Graph/SummaryPanelNamespaceBox.tsx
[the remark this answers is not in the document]
jshaughn: Interesting.  I would have thought the opposite, that `as <Type>` would actually ensure the property validation!

frontend/src/pages/Overview/OverviewPage.tsx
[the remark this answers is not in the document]
jshaughn: It was OK, I questioned this one mainly because you also had to update the body of the function.  In other places you could add `async` and the function was otherwise untouched.

frontend/src/services/Api.ts
jshaughn: I agree.  I think I tried doing something like this suggested refactor at one point, and it turned into a can or worms. The resulting type checking issues became a little overwhelming, so I abandoned the effort.  Perhaps with all of our good refactoring and more strict typing, it would work better now, but be prepared for a headache :-)

jshaughn (approved):
Looks Good to Me!

hhovsepy (approved):
Works fine, checked on FF and chrome.

PR #6762, opened 2023-10-23
[make] don't allow ossmc cr to be installed before kiali is installed

jmazzitelli:
trivial enhancement to the ossmconsole-create target - don't allow the dev to run this make target if Kiali isn't installed yet.


jshaughn (approved):
Logical.

PR #6774, opened 2023-10-25
OSSM-4624: trying to workaround missing side car in demo apps problem

FilipB:
Trying to workaround missing side car in demo apps problem which is causing failures in interop pipeline.


FilipB:
Interop pipeline was green with this change. If you agree I will also create PR for master and v1.73 which will be used in ossm 2.5

nrfox (approved):

PR #6775, opened 2023-10-25
Filter traces by cluster, show warning otherwise

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6710

Follow up after discussion on https://github.com/kiali/kiali/pull/6749

In a multi-cluster primary-remote mode, when Jaeger is configured to have Cluster ID [(Multi-cluster Setup)](https://kiali.io/docs/configuration/multi-cluster/#setup) and the Istio installation includes the [workaround for Telemetry setup](https://github.com/istio/istio/issues/35750#issuecomment-1165988761).
The Jaeger traces are including a `cluster` tag.
As in the multi-cluster env, the Tracing service supposed to be setup at once and during the installation, then it will be configured to support traces per cluster. This is the expected behavior, otherwise all Traces will be duplicated per service name which exists in all clusters.

What is done: 
In Multi-cluster mode, the Traces are first filtered by the `cluster` tag. This is the expected behavior.
However if no traces are found for particular cluster, then all traces are loaded and a warning message is shown to user saying that 'Loading traces for all clusters. Tracing is not configured to store traces per cluster.'. This is a sign to user to setup Jaeger to support traces per cluster.

Screenshot from 2023-10-31 18-30-00

Currently done for the Jaeger only.

### DIFF
--- a/frontend/cypress/integration/featureFiles/app_details_multicluster.feature
+++ b/frontend/cypress/integration/featureFiles/app_details_multicluster.feature
@@ -35,6 +35,7 @@ Feature: Kiali App Details page for multicluster
 
   Scenario: See tracing info after selecting a trace
     And user sees trace information
+    And user sees tracing warning
     When user selects a trace
     Then user sees trace details
 
--- a/frontend/cypress/integration/featureFiles/service_details_multicluster.feature
+++ b/frontend/cypress/integration/featureFiles/service_details_multicluster.feature
@@ -48,6 +48,7 @@ Feature: Kiali Service Details page for remote cluster
 
   Scenario: See graph traces for ratings service details
     And user sees trace information
+    And user sees tracing warning
     When user selects a trace
     Then user sees trace details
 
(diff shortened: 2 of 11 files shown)


josunect (commented):
It works as expected: 

When the cluster is included: 

image

But, I'm not able to see traces from the remote cluster: 
image

There are no tags. Maybe this is some issue related to my configuration. 

When the cluster tag is not included: 

image

I think we should create a follow up task to include the same behavior for Tempo.

josunect (approved):
Tested and LGTM. Thank you, @hhovsepy !

I think we should create a follow up task to include the same behavior for Tempo.

leandroberetta (approved):
Looks good!, I like the implementation. Thanks @hhovsepy !

hhovsepy:
Thank you @josunect and @leandroberetta , have added 2 checkboxes for https://github.com/kiali/kiali/issues/6710, second for Tempo.

Issue #6772, opened 2023-10-24
Multi mesh setup results in an error when fetching workloads


jshaughn:
### Discussed in https://github.com/kiali/kiali/discussions/6761

<div type='discussions-op-text'>

<sup>Originally posted by **InverseIntegral** October 23, 2023</sup>
I'm running a multi-mesh / multi-cluster setup with two instances of Kiali. Whenever I access Kiali on cluster A, I get the following error:

```
2023-10-23T16:32:21Z INF Kiali: Version: v1.75.0, Commit: 45d72534b0556c0425f52a1488811eb372c92e43, Go: 1.20.5
2023-10-23T16:32:21Z INF Using authentication strategy [anonymous]
2023-10-23T16:32:21Z WRN Kiali auth strategy is configured for anonymous access - users will not be authenticated.
2023-10-23T16:32:21Z INF Some validation errors will be ignored [KIA1301]. If these errors do occur, they will still be logged. If you think the validation errors you see are incorrect, please report them to the Kiali team if you have not done so already and provide the details of your scenario. This will keep Kiali validations strong for the whole community.
2023-10-23T16:32:21Z INF Initializing Kiali Cache
2023-10-23T16:32:21Z INF Adding a RegistryRefreshHandler
2023-10-23T16:32:21Z INF [Kiali Cache] Waiting for cluster-scoped cache to sync
2023-10-23T16:32:21Z INF [Kiali Cache] Started
2023-10-23T16:32:21Z INF [Kiali Cache] Kube cache is active for cluster: [A]
2023-10-23T16:32:21Z INF Server endpoint will start at [:20001/kiali]
2023-10-23T16:32:21Z INF Server endpoint will serve static content from [/opt/kiali/console]
2023-10-23T16:32:21Z INF Starting Metrics Server on [:9090]
2023-10-23T16:32:26Z INF [Prom Cache] Enabled
2023-10-23T16:32:26Z ERR grafana URL is not set in Kiali configuration
2023-10-23T16:32:26Z INF Create jaeger GRPC client tracing.istio-system:16685
2023-10-23T16:32:26Z ERR This controller DeploymentRevision is assigned in a Pod and it's not properly managed
2023-10-23T16:32:26Z ERR This controller DeploymentRevision is assigned in a Pod and it's not properly managed
2023-10-23T16:32:28Z ERR This controller DeploymentRevision is assigned in a Pod and it's not properly managed
[...]

PR #6782, opened 2023-10-25
Make sure mesh_check appender only tries to check nodes in accessible namespaces

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6772

Make sure mesh_check appender only tries to check nodes in accessible namespaces. Otherwise the fetch of workload information will fail.

### DIFF
--- a/graph/telemetry/istio/appender/mesh_check.go
+++ b/graph/telemetry/istio/appender/mesh_check.go
@@ -1,8 +1,6 @@
 package appender
 
 import (
-	"fmt"
-
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/graph"
 )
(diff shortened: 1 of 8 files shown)


graph/telemetry/istio/appender/mesh_check.go
```diff
@@ -95,15 +97,9 @@ func (a *MeshCheckAppender) applyMeshChecks(trafficMap graph.TrafficMap, globalI
...
-	}
-	for _, ns := range a.AccessibleNamespaces {
-		if namespace == ns.Name {
-			return true
-		}
-	}
-	return false
+// nodeOK returns true if we have access to its workload info
+func (a *MeshCheckAppender) nodeOK(node *graph.Node, namespaceInfo *graph.AppenderNamespaceInfo) bool {
+	key := fmt.Sprintf("%s:%s", node.Cluster, node.Namespace)
```
jmazzitelli, line 102: alarm bells go off when I see the same Sprintf everywhere building some formatted string (cluster, followed by colon, followed by namespace).

Consider creating a common utility function that everyone uses so you don't have these Sprintfs all over the place (especially if in the future that syntax needs to change for some reason).

jmazzitelli (approved):
i approve, however, the code can be simplified IMO by extracting out those Sprintfs into a single common utility so that future devs dont' have to worry about building the correct formatted string.

jmazzitelli (approved):
much better. I approve.

PR #6784, opened 2023-10-26
[v1.73] OSSM-4624: trying to workaround missing side car in demo apps problem…

FilipB:
Trying to workaround missing side car in demo apps problem.


FilipB:
Closing as it's probably no longer an issue.

PR #6755, opened 2023-10-19
Add absolute script dir path in setup-kind-ci.sh script

ferhoyos:
**Describe the change**

Add absolute script dir path in setup-kind-ci.sh script

**Issue reference**

Fixes #6750

PR #6789, opened 2023-10-27
[1.73] Add absolute script dir path in setup-kind-ci.sh script (#6755)

### LINKED ISSUE #6750: setup-kind-ci.sh script fails if it is not executed from root folder
The script setup-kind-ci.sh fails if it is executed from any folder except kiali root:

image

Path to hack/istio/download-istio.sh should be absolute and not relative.

ferhoyos:
**Describe the change**

Backport of #6755 to version 1.73. There are more changes than in referenced PR because there are previous PRs modifying this file that have not been backported. For this reason I have pushed the latest version of the hack script into v1.73.

**Issue reference**

Closes #6750

### DIFF
--- a/hack/istio/multicluster/deploy-kiali.sh
+++ b/hack/istio/multicluster/deploy-kiali.sh
@@ -27,11 +27,11 @@ if ! which helm; then
 fi
 
 deploy_kiali() {
-  local helm_args=""
+  local helm_args=()
   if [ "${IS_OPENSHIFT}" == "true" ]; then
     helm_args="--disable-openapi-validation"
     if [ "${KIALI_USE_DEV_IMAGE}" == "true" ]; then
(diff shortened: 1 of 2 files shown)


jmazzitelli (approved):

aljesusg:
Sorry but Why we need to backport this? It's not a bug in the product. Do we need it to test something?

jmazzitelli:
> Sorry but Why we need to backport this? It's not a bug in the product. Do we need it to test something?

I assumed it was to be able to run the tests within this branch (??)

ferhoyos:
Actually the bug does not break the tests, it is just only for developers. I set the backport label to this bug by mistake. Anyway, since it is done and reviewed, there is no harm to backport this change, right?

jmazzitelli:
no harm.

Issue #6035, opened 2023-04-17
Multicluster - Primary - Primary support


nrfox:
Support the Primary - Primary deployment model including communication with istiod(s).

In no particular order, here's a rough outline of what needs to be done for this:
- [x] #6430
- [x] #6432
- [x] #6846
- [x] #6937

Xunzhuo:
Assign myself on this sub-epic.

nrfox:
This is supported as of 1.78

PR #6791, opened 2023-10-27
Revert common cfgmap func and protect against failure

nrfox:
This reverts the callers of `IstioConfigMapName` back to using `cfg.external_services.istio.config_map_name` directly. It now logs failures in `GetMesh` to auto-detect the configmap instead of returning an error. Nothing is currently using that configuration so an error shouldn't be returned there. Related to: https://github.com/kiali/kiali/issues/6669.

### DIFF
--- a/business/istio_certs.go
+++ b/business/istio_certs.go
@@ -103,7 +103,7 @@ func (ics *IstioCertsService) getCertificateFromSecret(secretName, certName stri
 func (ics *IstioCertsService) getCertsConfigFromIstioConfigMap() ([]certConfig, error) {
 	cfg := config.Get()
 
-	istioConfigMap, err := ics.k8s.GetConfigMap(cfg.IstioNamespace, IstioConfigMapName(*cfg, ""))
+	istioConfigMap, err := ics.k8s.GetConfigMap(cfg.IstioNamespace, cfg.ExternalServices.Istio.ConfigMapName)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


nrfox:
cc @yohanb if you want to take a look. Basically this change does 2 things:

1. Removes the use of `external_services.istio.config_map_name` altogether when trying to guess the configmap name.
2. Makes any failure to get the "guessed" configmap a debug log that won't cause failures in gateway validations like before.

Another option is to just rip out fetching the guessed configmap entirely. I kept it since it will be used for in progress future work for #6035 but really it doesn't need to be there until it is being used directly.

business/istio_certs.go
```diff
@@ -103,7 +103,7 @@ func (ics *IstioCertsService) getCertificateFromSecret(secretName, certName stri
 func (ics *IstioCertsService) getCertsConfigFromIstioConfigMap() ([]certConfig, error) {
 	cfg := config.Get()
 
-	istioConfigMap, err := ics.k8s.GetConfigMap(cfg.IstioNamespace, IstioConfigMapName(*cfg, ""))
+	istioConfigMap, err := ics.k8s.GetConfigMap(cfg.IstioNamespace, cfg.ExternalServices.Istio.ConfigMapName)
```
jmazzitelli, line 106: Why wouldn't we want to keep this common function here and in the common function just return `cfg.ExternalServices.Istio.ConfigMapName`... in the future, if anything needs to change (like trying to get the name automagically) you only have to change this function and everywhere the name needs to be obtained will just work.

With this revert, if we ever have to change this, we need to go through all these places again and change it back to calling a common function.

I would leave the common function here - and for now just have it simply `return config.Get().ExternalServices.Istio.ConfigMapName`

jmazzitelli (dismissed):
I say keep a common function unless there is a reason you think ripping it back out is better.

business/istio_certs.go
nrfox: > Why wouldn't we want to keep this common function here and in the common function just return cfg.ExternalServices.Istio.ConfigMapName... in the future, if anything needs to change (like trying to get the name automagically) you only have to change this function and everywhere the name needs to be obtained will just work.

In the future, I would expect callers to call `GetMesh` to get the config info from there instead of fetching and unmarshaling the configmap every the config is needed.

jmazzitelli (approved):
i think the ci error is a flake, so approving in case you want this in for the release

nrfox:
Closing this since there are many conflicts and changes since this was opened.

Issue #6790, opened 2023-10-27
[operator] do not watch for OSSMConsole CRs on non-OpenShift environments


jmazzitelli:
It would be nice to not have to deploy the OSSMConsole CRD on non-OpenShift environments since you can't install the OSSMC plugin anywhere except within OpenShift. Right now [this problem occurs](https://github.com/kiali/kiali/issues/6788) without the OSSMConsole CRD.

This will require the operator to maintain two watches.yaml files and have the Deployment start with the proper one (passing the [`--watches-file`](https://github.com/operator-framework/ansible-operator-plugins/blame/release-v0.1/internal/ansible/flags/flag.go#L66-L69) to the operator container as a command line argument.).

master PRs:
* https://github.com/kiali/kiali-operator/pull/710
* https://github.com/kiali/helm-charts/pull/228

backport PR to 1.73
* https://github.com/kiali/kiali-operator/pull/711

PR #6793, opened 2023-10-27
[make] pass in correct watches file when using make target run-operator

jmazzitelli:
I missed this spot when I did https://github.com/kiali/kiali/issues/6790

This make target runs the operator locally - useful for dev debugging. It now needs to pass in the correct watches file.

This doesn't affect anything other than a make target that is used during development.


jmazzitelli:
Just run `make run-operator` and you'll see the problem :) This PR fixes the problem

nrfox (approved):

PR #6809, opened 2023-11-02
Add refresh delay to avoid multiple refresh in a very short period of time

### LINKED ISSUE #6714: Repeatedly refreshing causes the UI to crash
Kiali version: 1.75
Istio version: 1.19

Steps to reproduce:
- Repeatedly and rapidly click the refresh button
- After ~5 seconds the UI crashes

In the web console there is this message: `Too many calls to Location or History APIs within a short timeframe.`

Probably some component is trying to update the URL inside of a react lifecycle hook that is tied to the refresh button.

ferhoyos:
**Describe the change**

Add refresh delay when clicking Refresh Button to avoid multiple refresh in a very short period of time.

Besides, I have converted the Refresh button component in a React Hook.

**Issue reference**

Fixes #6714.

### DIFF
--- a/frontend/src/components/Refresh/RefreshButton.tsx
+++ b/frontend/src/components/Refresh/RefreshButton.tsx
@@ -2,39 +2,37 @@ import * as React from 'react';
 import { Button, ButtonVariant, Tooltip } from '@patternfly/react-core';
 import { SyncAltIcon } from '@patternfly/react-icons';
 
-type ComponentProps = {
-  id?: string;
+type RefreshButtonProps = {
   disabled?: boolean;
   handleRefresh: () => void;
+  id?: string;
 };
 
-export class RefreshButton extends React.Component<ComponentProps> {
-  getElementId() {
-    return this.props.id || 'refresh_button';
-  }
+export const RefreshButton: React.FC<RefreshButtonProps> = (props: RefreshButtonProps) => {
+  const [enableRefresh, setEnableRefresh] = React.useState<boolean>(true);
 
-  getDisabled() {
-    return this.props.disabled || false;
-  }
+  const handleRefresh = (): void => {
+    if (enableRefresh) {
+      props.handleRefresh();
 
-  render() {
-    return (
-      <Tooltip position="bottom" content={<>Refresh</>}>
-        <Button
-          id={this.getElementId()}
-          data-test="refresh-button"
-          onClick={this.handleRefresh}
-          isDisabled={this.getDisabled()}
-          aria-label="Action"
-          variant={ButtonVariant.primary}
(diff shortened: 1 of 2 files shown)


nrfox:
This might make it better but it doesn't seem to fix the issue:

refresh_error

In fact, there was already some delay before since the refresh button was disabled while the UI waits for the `/api/graph` request to respond.

Based on the console error logs, the issue seems to lie Kiali's handling of the URL and updating the History or Location too many times. If that's true then there might even be other ways to trigger this same error. It just so happens that clicking the refresh button repeatedly does it.

Screenshot from 2023-11-02 10-17-20

ferhoyos:
The problem occurs on `componentDidUpdate` method of  `GraphToolbar.tsx`:

```
componentDidUpdate() {
    // ensure redux state and URL are aligned
    if (this.props.edgeLabels?.length === 0) {
      HistoryManager.deleteParam(URLParam.GRAPH_EDGE_LABEL, true);
    } else {
      HistoryManager.setParam(URLParam.GRAPH_EDGE_LABEL, String(this.props.edgeLabels));
    }

    if (this.props.rankBy?.length === 0) {
      HistoryManager.deleteParam(URLParam.GRAPH_RANK_BY, true);
    } else {
      HistoryManager.setParam(URLParam.GRAPH_RANK_BY, String(this.props.rankBy));
    }

    if (this.props.activeNamespaces?.length === 0) {
      HistoryManager.deleteParam(URLParam.NAMESPACES, true);
    } else {
      HistoryManager.setParam(URLParam.NAMESPACES, namespacesToString(this.props.activeNamespaces));
    }

    if (this.props.replayActive === INITIAL_USER_SETTINGS_STATE.replayActive) {
      HistoryManager.deleteParam(URLParam.GRAPH_REPLAY_ACTIVE, true);
    } else {
      HistoryManager.setParam(URLParam.GRAPH_REPLAY_ACTIVE, String(this.props.replayActive));
    }

    if (this.props.trafficRates?.length === 0) {
      HistoryManager.deleteParam(URLParam.GRAPH_TRAFFIC, true);
    } else {
      HistoryManager.setParam(URLParam.GRAPH_TRAFFIC, String(this.props.trafficRates));
    }

    HistoryManager.setParam(URLParam.GRAPH_TYPE, String(this.props.graphType));
  }
```

History is modified multiple times everytime the graph is refreshed. I have added if statements to check if props have changed to perform such history modification. I hope this works.

I will maintain refresh button delay because it avoids error when you press repeatedly Refresh button in Overview page:

image

nrfox (approved):
Much better now! Haven't been able to reproduce the issue with this patch.

PR #6812, opened 2023-11-02
Remove padding from navigation list

### LINKED ISSUE #6810: Extra space between left nav and top nav
There's extra space between the left nav and top nav that shouldn't be there.

image

image

From an earlier screenshot it should look like this where there's no space between the two:
image

Kiali version: v1.76

ferhoyos:
**Describe the change**

Remove extra space between left nav and top nav.

image

**Issue reference**

Fixes #6810.

### DIFF
--- a/frontend/src/components/Nav/Menu.tsx
+++ b/frontend/src/components/Nav/Menu.tsx
@@ -24,7 +24,11 @@ const externalLinkStyle = kialiStyle({
   }
 });
 
-const ExternalLink = ({ href, name }) => (
+const navListStyle = kialiStyle({
+  padding: 0
+});
+
+const ExternalLink = ({ href, name }: { href: string; name: string }): React.ReactElement => (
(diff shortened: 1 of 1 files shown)


nrfox (approved):
LGTM

nrfox:
I think this needs a backport to 1.73 as well

ferhoyos:
The solution it is a bit different because I have to addapt it to PF4, but not very complicated.

ferhoyos:
Also I think that according to documentation guidelines, the `backport needed` labels are to be added to issues, not to PRs

nrfox:
> The solution it is a bit different because I have to addapt it to PF4, but not very complicated.

Ah ok well then is this still an issue if 1.73 is PF4?

> Also I think that according to documentation guidelines, the backport needed labels are to be added to issues, not to PRs

Whoops I'll adjust it to be correct. Thanks!

Issue #6816, opened 2023-11-02
purge maistra related things


jmazzitelli:
Now that Maistra will soon transition more towards upstream Istio (aka Project Sail), we should remove all things Maistra related, unless needed for backward compatibility reasons.

There is a lot of extra cruft in our hack scripts, Ansible operator roles, README type docs, etc. that is related to Maistra that will no longer be applicable. We should clean all of that stuff out of master branch (again, we can leave it in other branches for backwards compatibility if necessary).

server PR: https://github.com/kiali/kiali/pull/6819
operator PR: https://github.com/kiali/kiali-operator/pull/723
docs PR: https://github.com/kiali/kiali.io/pull/719

jmazzitelli:
Note: DO NOT backport ... this maistra stuff is needed in past branches.

PR #6819, opened 2023-11-03
purge maistra related stuff

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6816


jmazzitelli:
How I'm testing (I'll update this comment as my tests continue).

1. You need the operator PR, too: https://github.com/kiali/kiali-operator/pull/723
2. Get an OpenShift cluster and log into it via `oc` and log into the image registry via `podman`. Run `make cluster-status` to set up the image registry and get the image registry login command.
3. `make build build-ui test`
4. Install Istio via hack script: `hack/istio/install-istio-via-istioctl.sh -c oc`
5.

jshaughn (approved):
Wow, we had a lot of maistra-specific code, this is a nice simplifier.   Are we sure that none of this `is-maistra` is actually equivalent to `is-openshift`?   Looks good to me, I suppose we'll find out as we go if we need to add anything back for Sail.

jmazzitelli:
I never used is_maistra to mean is_openshift. There are is_openshift things in here that are for that. is_maistra was always for OSSM specific things. I looked at the is_maistra usages here and none look like "is_openshift" (in fact, many are inside if blocks that say if is_openshift..

PR #6821, opened 2023-11-03
[hack] flag for enabling monitoring so we can have prometheus in the cluster

jmazzitelli:
This option is necessary if you want CRC to install OpenShift with its own Prometheus server.

This PR is needed before you can use the new hack script to configure everything else (see https://github.com/kiali/kiali/pull/6822)


aljesusg (approved):
LGFM

Issue #6806, opened 2023-10-31
Add labels and annotations in wizards


ferhoyos:
Currently it is not possible to add labels or annotations within the Istio object creation wizard. If you want to add any label or annotation must be done in the preview YAML editor. 

![image](https://github.com/kiali/kiali/assets/122779323/acc2f629-a080-4693-9642-025fd5b58416)

![image](https://github.com/kiali/kiali/assets/122779323/a968c770-f758-4bec-9809-f3a150e6c82f)

I think adding the possibility to add labels and/or annotations from the wizard would be useful for Kiali users. 

EDIT:
Dividing the task into two subtasks to keep it simple and smaller:

- [ ] Add Labels
- [ ] Add Annotations

jshaughn:
First, I didn't know the Preview window was an editor :-)   Second, I'll ask @nrfox for input because he has recently been playing with adding Auth Policy.

nrfox:
Yeah I think it'd be generally useful to have the ability to add both annotations and labels to the istio objects when you create them through Kiali. They're optional fields so perhaps they could be minimized on the form somehow to reduce visual noise.

For Auth Policy specifically, `dry-run` is enabled through an annotation so it'd be nice to have the option to add it in the form itself instead of just in "preview". The API is alpha so the annotation might eventually move to a field on the spec but like I said above the ability to add annotations/labels is still useful and IMO wouldn't add much noise to the form.

PR #6823, opened 2023-11-06
Added Labels in Istio config creation wizard.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6806

Added a new switch "Labels" in all Istio config creation pages, when checked, a new Input appears to set 'labels' as a comma separated 'key=value' input values.

Screenshot from 2023-11-06 17-00-38

Annotations will be in a separate PR.


hhovsepy:
The failing tests are flaky with timeouts, and are in Envoy tab.

ferhoyos:
Hi @hhovsepy, your approach is ok but I suggest another thing to add labels... Since labels and annotations are key-value, and we already have the annotations wizard, I think that we could reuse it to add the labels instead of a long string with key-value labels separated by commas:

image

Another approach is the Openshift way of adding labels and annotations:
- Labels:

image

image

- Annotations:

image

image


My opinion: I would select the Openshift way to show labels (we coud do with existing Kiali `Labels` component) and reuse annotations wizard to edit the labels:

Here some mock screenshots:

image

image


In this case I recommend to ask @andrew-ronaldson for UX guidance in this issue.

hhovsepy:
thanks @ferhoyos good idea to follow the OCP styles

hhovsepy:
@ferhoyos have applied the styles:
Screenshot from 2023-11-27 10-36-28
Screenshot from 2023-11-27 10-36-10
Screenshot from 2023-11-27 10-35-55

hhovsepy:
Failures are in Workload details envoy tab.

frontend/src/components/IstioWizards/WizardAnnotations.tsx
```diff
@@ -188,14 +189,15 @@ export class WizardAnnotations extends React.Component<Props, State> {
     const header = (
       <>
         <Title id="modal-custom-header-label" headingLevel="h1" size={TitleSizes['2xl']}>
-          {this.props.canEdit ? 'Edit ' : 'View '}annotations
+          {this.props.canEdit ? 'Edit ' : 'View '}
+          {this.props.header ? this.props.header : 'annotations'}
```
ferhoyos, line 193: I don't think annotations should be default value (see my comment above). I would make header to be mandatory and don't set default value.

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -105,6 +112,14 @@ type State = {
 
 const formPadding = kialiStyle({ padding: '30px 20px 30px 20px' });
 
+const editIcon = kialiStyle({ marginLeft: '0.35em', marginBottom: '0.2em' });
```
ferhoyos, line 115: In my opinion I think rem measures are more robust because it depends on global font size instead of parent one, but if you like em is valid too ;-)

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -240,6 +259,32 @@ class IstioConfigNewPageComponent extends React.Component<Props, State> {
     });
   };
 
+  onLabelsWizardToggle = (value: boolean) => {
```
ferhoyos, line 262: Add void output

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -240,6 +259,32 @@ class IstioConfigNewPageComponent extends React.Component<Props, State> {
...
+  };
+
+  onAddLabels = (value: { [key: string]: string }): void => {
+    this.setState({
+      labels: value,
+      showLabelsWizard: false
+    });
+  };
+
+  onAnnotationsWizardToggle = (value: boolean) => {
```
ferhoyos, line 275: Add void output

ferhoyos (commented):
Good work @hhovsepy! The label and annotation creation work fine in the wizards.

Just some comments:
- If there are no labels/annotations in the wizards I would add an empty one by defualt when the user opens the wizard:

image

- Annotations could be very large (for example lastAppliedConfiguration), so labels components should be truncated from maximum number of characters (64?) and add tooltip in this case. This does not happens now:

frontend/src/components/Label/Labels.tsx
```diff
@@ -8,6 +8,7 @@ const SHOW_MORE_TRESHOLD = 2;
 
 interface LabelsProps {
   expanded?: boolean;
+  header?: string;
```
ferhoyos, line 11: I don't think header is a correct name since we do not have header in Labels component... Maybe type or similar?

hhovsepy:
Screenshot from 2023-11-28 11-29-49

ferhoyos:
I have seen in the snapshot that if you have more than one label, they are grouped vertically, leaving a lot of space to the right. 

You can fix this modifying `labelsContainerStyle` in Labels component:
```
const labelsContainerStyle = kialiStyle({
  display: 'flex',
  alignItems: 'center',
  flexWrap: 'wrap',
  overflow: 'hidden'
});
```

andrew-ronaldson:
I believe they use this pencil icon in OCP for Edit labels
image

hhovsepy:
right, changing the icon, the alignment is similar to OCP one:
Screenshot from 2023-11-28 16-31-44

ferhoyos:
@hhovsepy Much better now, edit icon updated, tooltip only on truncate labels, labels added horizontally...

One thing that does not look good to me is the 'More labels...' limit when the number of labels is 3 or more. It is ok to have it in service or workload description because of the limited space, but here we have plenty space to show all labels and annotations:

image

I would add new optional property to `Labels` component to enable/disable threshold.

ferhoyos:
Also I am not sure if `Labels` and `Annotations` should be below `Name` field or at the end of the form. Maybe the generic fields should be first (like `metadata` in yaml) and then specific fields of each Istio Config object afterwards. What do you think @hhovsepy?

frontend/src/components/Label/Labels.tsx
ferhoyos, line 11: Replace `header` by `type` like in `WizardLabels` component

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -15,6 +15,7 @@ import {
   HelperTextItem,
   TextInput
 } from '@patternfly/react-core';
+import { PencilAltIcon } from '@patternfly/react-icons/dist/esm/icons/pencil-alt-icon';
```
ferhoyos, line 18: Since we have KialiIcon wrapper component, I think we should use our own component instead of patternfly one directly.

hhovsepy:
> 
> I would add new optional property to `Labels` component to enable/disable threshold.

There is an optional property 'expanded', this can be used.
And regarding the location, I would leave it on the end, as of having a plenty of space to add more labels and annotations. Otherwise we need to have collapsed the labels not to push the rest of inputs to low and break the UX.

hhovsepy:
Thanks for the review @ferhoyos 
Expanded it looks like this:
Screenshot from 2023-11-29 09-14-16

ferhoyos (changes requested):

frontend/src/components/Label/Labels.tsx
hhovsepy: ok, but keeping it as optional with default value, as this component is mostly used in other places for labels

frontend/src/config/KialiIcon.tsx
```diff
@@ -118,6 +119,7 @@ export const KialiIcon: { [name: string]: React.FunctionComponent<IconProps> } =
   OutOfMesh: (props: IconProps) => conversorIconProps(props, <BlueprintIcon />),
   Pause: (props: IconProps) => conversorIconProps(props, <PauseIcon />),
   PauseCircle: (props: IconProps) => conversorIconProps(props, <PauseCircleIcon />),
+  PencilAlt: (props: IconProps) => conversorIconProps(props, <PencilAltIcon />),
```
ferhoyos, line 122: Maybe `Edit` has more meaning for Kiali application than `PencilAlt` (actually it is the edit icon for Kiali application that will be applied in future features)

jshaughn:
Nice development and review collaboration on this one, @hhovsepy and @ferhoyos .

PR #6832, opened 2023-11-08
Use cluster tag for Tempo

### LINKED ISSUE #6710: Traces tab is duplicated across both clusters

josunect:
Fixes #6710 (Tempo)

For some reason, telemetry is not adding the cluster tag. But without this fix, Tempo search is not working, as the cluster tag is included and the search is not returning any result.

### DIFF
--- a/tracing/client.go
+++ b/tracing/client.go
@@ -141,6 +141,7 @@ func NewClient(token string) (*Client, error) {
 				httpTracingClient, err = tempo.NewOtelClient(client, u)
 				if err != nil {
 					log.Errorf("Error creating HTTP client %s", err.Error())
+					return nil, err
 				}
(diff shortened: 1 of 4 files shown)


hhovsepy (approved):
verified
before:
Screenshot from 2023-11-20 14-05-59

after:
Screenshot from 2023-11-20 14-05-15

josunect:
Thanks for the review, @hhovsepy ! Sorry I requested again after fixing some conflicts

hhovsepy (approved):

PR #6833, opened 2023-11-08
Cypress implementation of the Request Routing tests for the Multicluster setup 

matejnesuta:
A few of the implemented gherkin scenarios for MC which have been previously marked with the @skip tag. 

I have also added a hook to install the necessary CRDs if not detected on the remote cluster. 

I am not sure, if we want to test with enabled or disabled CRDs, but we can always alter the execution order later

### DIFF
--- a/frontend/cypress/integration/common/hooks.ts
+++ b/frontend/cypress/integration/common/hooks.ts
@@ -1,5 +1,8 @@
 import { Before } from '@badeball/cypress-cucumber-preprocessor';
 
+const CLUSTER1_CONTEXT = Cypress.env('CLUSTER1_CONTEXT');
+const CLUSTER2_CONTEXT = Cypress.env('CLUSTER2_CONTEXT');
+
 function install_demoapp(demoapp: string) {
   var namespaces: string = 'bookinfo';
   var deletion: string = `--delete-${demoapp}`;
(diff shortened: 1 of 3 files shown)


frontend/cypress/integration/common/hooks.ts
```diff
@@ -85,3 +88,17 @@ Before({ tags: '@error-rates-app' }, async function () {
 Before({ tags: '@sleep-app' }, async function () {
   install_demoapp('sleep');
 });
+
```
ScriptingShrimp, line 91: @nrfox how do we install Istio CRD's in multicluster setup? Do we want to implement logic below in cypress only? Just thinking about having extra script to do that for better re usability. (non blocking)

ScriptingShrimp (approved):
If there is no need to refactor installation of CRD's, LGTM!

frontend/cypress/integration/common/hooks.ts
nrfox: For primary-remote, CRDs are installed on the primary cluster but not the remote cluster like in what you see here: https://istio.io/latest/docs/setup/install/multicluster/primary-remote/. "config" in the diagram means the istio CRDs.

The primary-remote installation happens in the kiali hack scripts that run before any of tests to setup the environment. Here's where the remote cluster gets setup: https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/install-primary-remote.sh#L88. The `--profile=remote` flag is what installs the necessary components for the remote cluster but not the CRDs.

In general the Kiali e2e and cypress tests right now assume that the environment is all setup ahead of time by the hack scripts before they run but sometimes they do things like check that the necessary demo apps are installed iirc. I think we should keep the environment setup outside the scope of the tests so that the tests can be run against a variety of environments (locally on kind or against a cloud env etc.) but installing istio itself could be part of the test framework since that will probably be environment agnostic and it would be nice to ensure things about the istio installation before running specific tests like for multi cluster or testing against multiple revisions. That's a larger change though as today installing istio is part of the environment setup.

nrfox: @ScriptingShrimp @matejnesuta ^^

Issue #6773, opened 2023-10-24
Info icon in yaml editor's overview panel is not aligned properly


nrfox:
In the istio config yaml editor, the info icon on the right hand side overview panel is not vertically aligned properly with the rest of the elements.

![Screenshot from 2023-10-24 16-48-25](https://github.com/kiali/kiali/assets/6226732/cd204a9b-d374-4149-8b69-5ad1c71248c6)

IMO this info icon can be removed entirely because it's only purpose is showing a tooltip when you hover over it that merely duplicates the information that can already be seen from the istio config editor namely "creationTimestamp" and "resourceVersion":

![Screenshot from 2023-10-24 16-48-04](https://github.com/kiali/kiali/assets/6226732/7c20abe2-e7cd-4930-9cf2-c8f568827afa)

If Kiali should duplicate this info on the overview panel, then IMO it should just be plaintext on the panel rather hidden in an "info" icon/tooltip.

ferhoyos:
Not only this icon, almost all info icons are not aligned properly after PF5 upgrade. I have fixed some of them in Table upgrade PR (https://github.com/kiali/kiali/pull/6695) and I will check the rest after upgrading PF to version 5.1 (https://github.com/kiali/kiali/issues/6768).

ferhoyos:
Fixed with PR #6836

PR #6836, opened 2023-11-10
Update PF to version 5.1 and fix some alignment UI issues

### LINKED ISSUE #6768: Update Patternfly library to version 5.1
To finish PF5 upgrade tasks in Kiali it would be great to update PF to latest version (5.1.1) and make a final UI review.

ferhoyos:
**Describe the change**

This PR upgrades Patternfly to version 5.1 and fix some UI issues:
- Info icons not aligned properly (https://github.com/kiali/kiali/issues/6773)
- Formatting and fix some CSS style disalignments (margin and paddings)
- The Close button does not work on Debug Information.
- Type column sorting does not seem to work for Envoy -> Clusters subtab
- Traffic tab sorting seems to only sort one way
- Same health icon size in all Kiali tables

image

- Change some colors so that it looks better on Dark mode:
  - Graph Replay controls
  - Metrics chart background stroke
  - 'Edit in Kiali' link in Istio config page (only for kiosk)

- More legend overlay disalignment

image

- Since ServiceValidationSummary component is very similar to ValidationSummary component and Class inheritance is not allowed with React Hooks, I have added new type property to ValidationSummary (`istio` or `service`) and delete ServiceValidationSummary component.

- NamespaceInfo.ts moved to types folder
- HealthStyle moved to styles folder
- Remove some console.log in Renderers component 

**Refactorization Bonus (in files modified):**
- Small components migrated to hooks (plain components, straightforward conversions).
- Alphabetized props lists
- Add method types (params and return)
- Replace strings with correct concatenation (https://github.com/kiali/kiali/issues/6736)
- Replace || by ?? to check nullish values (|| operator can lead to errors since it checks falsy values, not nullish)
- Convert function to arrow function when it is possible
- Convert var to let or constant when it is possible

** Issue reference **

Closes #6768


ferhoyos:
@jshaughn One of the issues you found in one of previous PR reviews is confusing me...
- **Traffic tab sorting seems to only sort one way**

I have fixed it but now both inbound and outbound tables are connected in the sorting handling. I mean that if you sort inbound traffic table by name, the outbound traffic table is also sorted. Besides, if you see the in the url there are only one `sort` and `direction` URL params. Is this the expected behaviour or both tables should have separate sorting handlers?

image

image

jshaughn:
> Is this the expected behaviour or both tables should have separate sorting handlers?

Right, I think I remember this as always being an issue for this page.  I think the decision was that it is OK if the sort affects both tables as it was not worth the added complexity to make it two different sorts.  So, I think this is fine as is!

jshaughn (changes requested):
Almost everything looks good to me!  Tested the oddities that I knew about and everything seems fixed.  Looked at some other fixes and everything was good/better.  One new issue, in the PF graph the legend now has icon+text, which makes the legend too verbose:

image

ferhoyos:
I've been investigating the issue and the text should not appear.

The text element within the legend is:
```
<span class="pf-v5-u-screen-reader">Zoom Out</span>
```

According to https://www.patternfly.org/utility-classes/accessibility/#usage, CSS class `pf-v5-u-screen-reader` should hide the text but leaves accessible to assistive technologies. That means we are not loading all Patternfly CSS. I will check how to load PF accessibility CSS.

aljesusg:
> Almost everything looks good to me! Tested the oddities that I knew about and everything seems fixed. Looked at some other fixes and everything was good/better. One new issue, in the PF graph the legend now has icon+text, which makes the legend too verbose:
> 
> image

Could we add some padding betweenthe icon and text ?

ferhoyos:
Yes, we could add padding but I think the best approach here (and the intention of `react-topology` library) is to show only the icons (then tooltip gives more information about the toolbar option). This is the approach we follow in Cystoscape graph and patternfly examples (https://www.patternfly.org/topology/control-bar)

aljesusg:
> Yes, we could add padding but I think the best approach here (and the intention of `react-topology` library) is to show only the icons (then tooltip gives more information about the toolbar option). This is the approach we follow in Cystoscape graph and patternfly examples (https://www.patternfly.org/topology/control-bar)

We have so buttons with the same icon...we'll need help here from UX for this if we want to have only icons

ferhoyos:
Sure, we can ask UX for guiding in the topology legend, only thing is that we would have to follow same approach for both graphs (cytoscape and PF).

ferhoyos:
I have loaded PF utilities CSS and now the text is hidden in PF graph legend:

image

jshaughn:
> Sure, we can ask UX for guiding in the topology legend, only thing is that we would have to follow same approach for both graphs (cytoscape and PF).

We can ask for guidance but I will veto anything that has on-screen text, it needs to be handled with icons and tooltips.  The alternative is too verbose.

jshaughn (approved):
Now that the ControlBar (on-graph icons) is back to having no text, just icons, I approve!  Nice work.

hhovsepy (approved):
Regression tested the pages, compared the alignment with older one, now it looks better and all small issues with icons alignments are resolved. Thank you @ferhoyos for the massive work.

PR #6853, opened 2023-11-16
gRPC client for Tempo (II)

### LINKED ISSUE #6578: Tempo support for gRPC Tempo client
With the following settings: 

```
  tracing:
    enabled: true
    provider: "tempo"
    use_grpc: true
```

The Tempo client is not working. 
The client uses the .proto specific to Jaeger.

josunect:
I was not able to make it work the grpc frontend deploying Tempo with Tanka or the tempo-operator, but it is working fine in the deployment with helm charts. 

image
	
Fixes #6578


josunect:
Note about the implementation: 

- The StreamClient just support the Search endpoint. The traceId doesn't look to be available in the search endpoint, that's why it uses the http client for traceID. It seems to be the way used by Grafana. https://github.com/grafana/grafana/blob/main/pkg/tsdb/tempo/search_stream.go#L63
- So Tempo would need the httpClient also for use grpc. That's why this new option (grpc_port) was added in the config.

jmazzitelli:
Looking at the `tempopb` directory, it appears you copied some code from "some place" to support the Tempo client?

You are going to want to create a README in here (or something like that) explaining:

1. Where you got this source code (so when we need to upgrade it or pull in security/CVE fixes, we know where to go to get the new code)
2. What versions of Tempo this supports.

josunect:
> Looking at the `tempopb` directory, it appears you copied some code from "some place" to support the Tempo client?
> 
> You are going to want to create a README in here (or something like that) explaining:
> 
>     1. Where you got this source code (so when we need to upgrade it or pull in security/CVE fixes, we know where to go to get the new code)
> 
>     2. What versions of Tempo this supports.

Thanks, @jmazzitelli ! It makes sense. I will add it.

hhovsepy (approved):
Code lgtm, tested and debugged locally grpc client, regression tested http client.

jmazzitelli (approved):
Tested, I can see the traces in the UI (which I assume is all we need to look for).

(FYI: I did update the testing instructions in the previous comment to be more specific and clear).

josunect:
> Tested, I can see the traces in the UI (which I assume is all we need to look for).
> 
> (FYI: I did update the testing instructions in the previous comment to be more specific and clear).

Yes that's right. Thanks for the review!

Issue #6815, opened 2023-11-02
Multicluster - Missing cluster param in Show traces


josunect:
When we click on "Show Traces" on a workload from a namespace that just exist in remote: 

![image](https://github.com/kiali/kiali/assets/49480155/8af014ed-c76a-4dda-87f1-54ccf28dc3c6)

The following error occurs: 

![image](https://github.com/kiali/kiali/assets/49480155/d72afa7e-21e4-4670-83e7-7027d8507166)

From Kiali logs: 
`
2023-11-02T19:42:15Z ERR namespaces "beta" not found`

As it is not including the cluster parameter in the URL: 
`
http://localhost:3000/console/namespaces/beta/workloads/x-server?tab=traces&rangeDuration=600&refresh=0`

PR #6854, opened 2023-11-16
Multi-cluster test implementation related mainly to Istio Config page

matejnesuta:
I have decided to change the order of test execution a little. We can now test for remote Istio CRDs both being and not being present. I have also installed a sleep app for one of the tests to also cover situation, where namespace is present only on 1 of the clusters. I think the demo app will be useful when testing this issue as well: https://github.com/kiali/kiali/issues/6815.

### DIFF
--- a/frontend/cypress/integration/featureFiles/istio_config.feature
+++ b/frontend/cypress/integration/featureFiles/istio_config.feature
@@ -78,14 +78,13 @@ Feature: Kiali Istio Config page
     Then the user can create a "Sidecar" Istio object
 
   @multi-cluster
-  @skip
+  @remote-istio-crds
   Scenario: See all Istio Config objects in the bookinfo namespace in the multi-cluster environment.
+    And there are Istio config objects created in both clusters
     Then the "Cluster" column "appears"
-    And user sees all the Istio Config objects from both clusters in the bookinfo namespace
-    And user sees Name information for Istio objects
-    And user sees Namespace information for Istio objects
-    And user sees Type information for Istio objects
-    And user sees Configuration information for Istio objects
+    And user sees the Istio Config objects from both clusters in the bookinfo namespace
+    And user sees "Name" information for Istio objects from both clusters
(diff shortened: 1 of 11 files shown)


frontend/cypress/integration/featureFiles/istio_config.feature
```diff
@@ -78,14 +78,13 @@ Feature: Kiali Istio Config page
...
+    And there are Istio config objects created in both clusters
     Then the "Cluster" column "appears"
-    And user sees all the Istio Config objects from both clusters in the bookinfo namespace
-    And user sees Name information for Istio objects
-    And user sees Namespace information for Istio objects
-    And user sees Type information for Istio objects
-    And user sees Configuration information for Istio objects
+    And user sees the Istio Config objects from both clusters in the bookinfo namespace
+    And user sees "Name" information for Istio objects from both clusters
+    And user sees "Configuration" information for Istio objects from both clusters
```
nrfox, line 87: This test will fail unless it's a multi-primary deployment since by definition there are no CRDs present in a primary-remote deployment. Personally I'd prefer we use a `@multi-primary` and/or `@primary-remote` rather than `@remote-istio-crds`.

Let me know what you think but I see there being three different tags to use here and two different ways to use them:

Option 1: Everything tagged with multi-cluster. Need to filter out primary-remote and multi-primary in test execution.
`@multi-cluster`: Every multi-cluster test should be tagged with this.
`@primary-remote`: These only run with a primary-remote deployment.
`@multi-primary`: These only run with a multi-primary deployment.

Option 2: Only tests that can run on _both_ multi-primary _and_ primary-remote are tagged with `@multi-cluster`. Tests tagged with `@multi-primary` and `@primary-remote` are not tagged with `@multi-cluster`.
`@multi-cluster`: Only tests that can run on _both_ multi-primary _and_ primary-remote should be tagged with this.
`@primary-remote`: These only run with a primary-remote deployment.
`@multi-primary`: These only run with a multi-primary deployment.

Issue #6858, opened 2023-11-20
Integration tests - Kiali 1.73 is not compatible with Istio 1.20


ferhoyos:
Integration tests performed within github actions for target branch v1.73 are failing because latest version of Istio (1.20) is installed. For the moment Kiali 1.73 is not compatible with Istio 1.20, so installation scripts have to be modified to install a compatible Istio version.

The same should be done for previous supported Kiali versions (1.48, 1.57 and 1.65)

PR #6860, opened 2023-11-20
Install compatible Istio versions on integration tests

ferhoyos:
**Describe the change**

Install compatible Istio versions depending on target branch when performing integration tests.

Compatible Istio versions are obtained from:
https://kiali.io/docs/installation/installation-guide/prerequisites/#version-compatibility

Compatible Kubernetes versions are obtained from:
https://istio.io/latest/docs/releases/supported-releases/

**Issue reference**

#6858

### DIFF
--- a/hack/setup-kind-in-ci.sh
+++ b/hack/setup-kind-in-ci.sh
@@ -61,11 +61,17 @@ TARGET_BRANCH="${TARGET_BRANCH:-master}"
 if [ -z "${ISTIO_VERSION}" ]; then
   if [ "${TARGET_BRANCH}" == "v1.48" ]; then
     ISTIO_VERSION="1.13.0"
+  elif [ "${TARGET_BRANCH}" == "v1.57" ]; then
+    ISTIO_VERSION="1.15.0"
+  elif [ "${TARGET_BRANCH}" == "v1.65" ]; then
+    ISTIO_VERSION="1.17.0"
+  elif [ "${TARGET_BRANCH}" == "v1.73" ]; then
+    ISTIO_VERSION="1.18.0"
   fi
 fi
 
 KIND_NODE_IMAGE=""
-if [ "${ISTIO_VERSION}" == "1.13.0" ]; then
+if [ "${ISTIO_VERSION}" == "1.13.0" || "${ISTIO_VERSION}" == "1.15.0" || "${ISTIO_VERSION}" == "1.17.0" ]; then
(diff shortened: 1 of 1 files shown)


hack/setup-kind-in-ci.sh
```diff
@@ -61,11 +61,17 @@ TARGET_BRANCH="${TARGET_BRANCH:-master}"
...
+  elif [ "${TARGET_BRANCH}" == "v1.65" ]; then
+    ISTIO_VERSION="1.17.0"
+  elif [ "${TARGET_BRANCH}" == "v1.73" ]; then
+    ISTIO_VERSION="1.18.0"
   fi
 fi
 
 KIND_NODE_IMAGE=""
-if [ "${ISTIO_VERSION}" == "1.13.0" ]; then
+if [[ "${ISTIO_VERSION}" == "1.13.0" || "${ISTIO_VERSION}" == "1.15.0" || "${ISTIO_VERSION}" == "1.17.0" ]]; then
```
jmazzitelli, line 74: this is fine, but FYI, you dont need to use the 2-bracket solution if you use the `-o` syntax like:
`if [ "${ISTIO_VERSION}" == "1.13.0" -o "${ISTIO_VERSION}" == "1.15.0" -o "${ISTIO_VERSION}" == "1.17.0" ]; then`

ferhoyos: ok, for the next time what is better to use? 2-bracket or `-o` option?

jmazzitelli:
I try to avoid the 2-bracket only because I believe that is only valid for some implementations of bash (not POSIX compliant). But I could be several years out of date :) It's a habit for me now to use `-o` (OR) and `-a` (AND).

hack/setup-kind-in-ci.sh
```diff
@@ -61,11 +61,17 @@ TARGET_BRANCH="${TARGET_BRANCH:-master}"
 if [ -z "${ISTIO_VERSION}" ]; then
   if [ "${TARGET_BRANCH}" == "v1.48" ]; then
     ISTIO_VERSION="1.13.0"
+  elif [ "${TARGET_BRANCH}" == "v1.57" ]; then
+    ISTIO_VERSION="1.15.0"
+  elif [ "${TARGET_BRANCH}" == "v1.65" ]; then
+    ISTIO_VERSION="1.17.0"
+  elif [ "${TARGET_BRANCH}" == "v1.73" ]; then
+    ISTIO_VERSION="1.18.0"
```
jshaughn, line 69: This probably should have been aligned with the OSSM versions or Istio:  1.12, 1.14, 1.16, 1.18.

Issue #6856, opened 2023-11-17
Istio 1.20 incompatibility


nbjohnson:
After upgrading istio to 1.20, kiali has now started to fail to load data. I am running kiali 1.76 which seemed to work fine on istio 1.19, however after the istio upgrade kiali is unable to load any configs and just starts spewing `json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool` and `Error parsing RegistryServices results: json: cannot unmashal object into Go struct field .Attributes.ExportTo of type bool` errors repeatedly. I assume this is related to our istio configs that have the `exportTo` option set, however don't klnow why it would be referencing type bool when that should be a list of strings. Any help in determining why this error is occurring would be greatly appreciated, it might be a config issue but since this worked on istio 1.19 I feel like it must have something to do with the upgrade to istio 1.20 and possibly an incompatibility.

jmazzitelli:
We need the full logs (and hopefully the stack trace if there is one) where that exception happens.

nbjohnson:
So I attempted to set the log to trace level and there are no additional logs, it is just repeated ones of the logs I already posted. I even tried with the new 1.77.0 version but nothing has changed, still the same errors

nbjohnson:
Sorry thought you just meant debug logs from kiali.

Here is the full log output from kiali:
```
2023-11-20T15:06:08Z ERR Error parsing RegistryServices results: json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool
2023-11-20T15:06:08Z ERR Error fetching Registry Services per namespace <namespace>: json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool
2023-11-20T15:06:08Z ERR Error while fetching app health: json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool
2023-11-20T15:06:08Z ERR Error parsing RegistryServices results: json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool
2023-11-20T15:06:09Z ERR Error parsing RegistryServices results: json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool
2023-11-20T15:06:09Z ERR json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool
2023-11-20T15:06:09Z ERR Error parsing RegistryServices results: json: cannot unmarshal object into Go struct field .Attributes.ExportTo of type bool
[...]

PR #6866, opened 2023-11-20
Registry Service ExportTo is now map[string]struct{}.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6856

The RegistryService since Istio 1.20, is returned from Istio reigstry with "map[string]struct{}" value for 'ExportTo' field, instead of 'map[string]bool' previously.

This was causing errors when we had a ServiceEntry with "exportTo" attribute, this was generating a Service which was set in ServicesRegistry.

All the other Istio Config object types are not touched with this issue, as the issue is in Service Registry results parsing.

Added exportTo value int ServiceEntry test data for e2e tests.

### DIFF
--- a/kubernetes/filters_test.go
+++ b/kubernetes/filters_test.go
@@ -231,8 +231,8 @@ func CreateFakeRegistryService(host string, namespace string, exportToNamespace 
 	registryService.Hostname = host
 	registryService.IstioService.Attributes.Namespace = namespace
 	registryService.IstioService.Attributes.Labels = labels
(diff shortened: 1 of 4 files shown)


jmazzitelli (approved):
1. Running `make build build-ui test` passes all unit tests.
2. Running replication procedures documented [here](https://github.com/kiali/kiali/issues/6856#issuecomment-1819541014), but using a dev build produced by building this PR's source code no longer shows the error in the UI nor the pod logs.

PR #6871, opened 2023-11-21
Update min duration unit for Jaeger traces filter

### LINKED ISSUE #6870: Jaeger traces: Filter by percentile no returning any trace
In the traces tab, if we filter by percentile, there are no results returned: 

image

image

The query: 

end=1700570248000000&limit=500&minDuration=14031&service=productpage.bookinfo&start=1700570188000000

It looks like it need the duration unit now: 

image

josunect:
Fixes #6870


hhovsepy (approved):
tested locally

PR #6874, opened 2023-11-21
Removed edge customizations in favor of adding more CSS logic to styleEdge component

leandroberetta:
Resolves https://github.com/kiali/openshift-servicemesh-plugin/issues/223

This PR aims to remove the customizations done to the edge component.

This PR uses the DefaultEdge from PF and it's modified by CSS in the StyleEdge component managed by Kiali (can be renamed to KialiEdge really).


jshaughn (commented):
This is promising, it seems like you may be able to eliminate the need for us to copy DefaultEdge.  But there are some things that have changed or have been lost.

- Edge width has changed.  I'm actually OK thinning the edges a bit, but I don't think it's intentional.  I think this is because if `pathStyle` is supplied we are not applying the entire style, only color.  I think we should probably be honoring the entire style (note - the terminator needs only the color).

- We lost the unhighlighting support
  - Before:
image

  - After:
image

- We lost the `add showTag prop and show scaled tag on hover (when showTag is false)`  customization.

leandroberetta:
> This is promising, it seems like you may be able to eliminate the need for us to copy DefaultEdge. But there are some things that have changed or have been lost.
> 
> * Edge width has changed.  I'm actually OK thinning the edges a bit, but I don't think it's intentional.  I think this is because if `pathStyle` is supplied we are not applying the entire style, only color.  I think we should probably be honoring the entire style (note - the terminator needs only the color).
> * We lost the unhighlighting support
>   
>   * Before:
>     image
>   * After:
>     image
> * We lost the `add showTag prop and show scaled tag on hover (when showTag is false)`  customization.

The edge width is intentional, I'm using 2 I guess now, I think it's cleaner. 

The rest, yes, is missing, I forgot why this was a draft :). I will add it back.

jshaughn (changes requested):
The pathStyle issue is fixed, and I like the thinner edges.  The unhighlighting is better but there is still a problem with unhighlighting the tags:

image

leandroberetta:
> The pathStyle issue is fixed, and I like the thinner edges. The unhighlighting is better but there is still a problem with unhighlighting the tags:
> 
> image

Good catch, now is fixed!

image

jshaughn (changes requested):
Edge tab unhighlighting is fixed, thanks!  But, two more things.  Hovering over an edge loses the edge coloring, and also does not trigger the unhighlighting:

image

leandroberetta:
> Edge tab unhighlighting is fixed, thanks! But, two more things. Hovering over an edge loses the edge coloring, and also does not trigger the unhighlighting:
> 
> image

Interesting, it seems to be a problem when hovering, this should be related to the override of css properties. Thanks.

frontend/src/pages/GraphPF/styles/styleEdge.tsx
```diff
@@ -1,23 +1,116 @@
...
-});
-
 const StyleEdgeComponent: React.FC<StyleEdgeProps> = ({ element, ...rest }) => {
   const data = element.getData();
   const detailsLevel = useDetailsLevel();
 
+  let classes: string[] = [];
+
+  // Change edge color according to the pathStyle
+  const edgeClass = style({
```
ferhoyos, line 30: Please use kialiStyle wrapper instead of style. It adds a prefix to CSS class name to fulfill Openshift plugin CSS requirement for OSSMC

frontend/src/pages/GraphPF/styles/styleEdge.tsx
```diff
@@ -31,7 +124,7 @@ const StyleEdgeComponent: React.FC<StyleEdgeProps> = ({ element, ...rest }) => {
     return newData;
   }, [data, detailsLevel]);
 
-  return <BaseEdge element={element} tagClass={tagClass} {...rest} {...passedData} />;
+  return <DefaultEdge className={classes?.join(' ')} element={element} {...rest} {...passedData} />;
```
ferhoyos, line 127: Instead of `join(' ')` I would use `classes` method of `typestyle` library, a utility to join CSS classes conditionally
```
<DefaultEdge className={classes(...cssClasses)} element={element} {...rest} {...passedData} />;
```

ferhoyos (changes requested):

leandroberetta:
> > Edge tab unhighlighting is fixed, thanks! But, two more things. Hovering over an edge loses the edge coloring, and also does not trigger the unhighlighting:
> > image
> 
> Interesting, it seems to be a problem when hovering, this should be related to the override of css properties. Thanks.

@jshaughn I found the issue, there is a missing part due to the removal of the custom edge code, but sadly, I don't see yet how I can do it without touching the edge. At some point, in the edge, when hover is true, we are calling a Kiali function:

https://github.com/kiali/kiali/blob/master/frontend/src/pages/GraphPF/components/edge.tsx#L105

jshaughn:
> but sadly, I don't see yet how I can do it without touching the edge. At some point, in the edge, when hover is true, we are calling a Kiali function:

Right, this is what I meant in the file comment:

`//   add showTag prop and show scaled tag on hover (when showTag is false)`

We need to decide what to do for this PR.  We can:
1. Eliminate our custom `Edge.tsx` and regress with respect to this tag feature.
2. Re-introduce a more minimal `Edge.tsx`, that maintains the feature while we work with PFT for a supported alternative.  

My preference would probably be option 2 because in addition to `Edge.tsx` we still have out customized versions for Node and Groups, so it's not the end of our customization.  I think we should go through the same exercise for Nodes and Groups, seeing what we can move to the `Style` classes, and work with PFT for anything remaining.

leandroberetta:
@jshaughn I finally could make the unhighlighting part:

image

I also fixed the edge style when hovering.

@ferhoyos you're feedback is addressed too. Thanks!.

frontend/src/pages/GraphPF/styles/styleEdge.tsx
```diff
@@ -1,23 +1,128 @@
...
+  };
+
+  const onMouseLeave = () => {
+    data.onHover(element, false);
+  };
+
+  // Change edge color according to the pathStyle
+  const edgeClass = kialiStyle({
+    $nest: {
+      '.pf-topology__edge__link': data.pathStyle
```
ferhoyos, line 41: According to `typestyle` documentation (https://typestyle.github.io/#/core), it is best practice to add symbol `$` to overwrite third party nested CSS classes:

```
const container = style({
  $nest: {
    '& .third-party-class': {
      color: 'red'
    }
  }
});
```
Most of the times it works fine without the `&` symbol, but not in all cases (if the third party CSS class has more specifity than yours). More information here: https://github.com/kiali/kiali/pull/6562

ferhoyos (changes requested):

frontend/src/pages/GraphPF/styles/styleEdge.tsx
```diff
@@ -1,23 +1,128 @@
...
+      '& .pf-topology__edge__link': data.pathStyle
+    }
+  });
+  cssClasses.push(edgeClass);
+
+  const edgeHoverClass = kialiStyle({
+    $nest: {
+      '& .pf-topology__edge.pf-m-hover': {
+        $nest: {
+          '& .pf-topology__edge__link, .pf-topology-connector-arrow': data.pathStyle
```
ferhoyos, line 50: Each selector is independent, so you have to add '&' for each selector:
`'& .pf-topology__edge__link, & .pf-topology-connector-arrow': data.pathStyle`

ferhoyos (dismissed):
LGTM from the CSS point of view, you have learnt a lot in the process and done a great job @leandroberetta ;-). Anyway I would wait for @jshaughn review from topology features point of view.

frontend/src/pages/GraphPF/styles/styleEdge.tsx
```diff
@@ -1,23 +1,128 @@
...
+const OverlayOpacity = 0.3;
+const OverlayWidth = 30;
 
 type StyleEdgeProps = {
   element: Edge;
 } & WithSelectionProps;
 
-const tagClass = kialiStyle({
-  fontFamily: 'Verdana,Arial,Helvetica,sans-serif,pficon'
-});
```
jshaughn, line 15: This is needed to support the lock icons on edge tags.

frontend/src/pages/GraphPF/styles/styleEdge.tsx
```diff
@@ -31,7 +136,11 @@ const StyleEdgeComponent: React.FC<StyleEdgeProps> = ({ element, ...rest }) => {
     return newData;
   }, [data, detailsLevel]);
 
-  return <BaseEdge element={element} tagClass={tagClass} {...rest} {...passedData} />;
+  return (
+    <g style={{ opacity: opacity }} onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
+      <DefaultEdge className={classes(...cssClasses)} element={element} {...rest} {...passedData} />
```
jshaughn, line 141:
```suggestion
      <DefaultEdge className={classes(...cssClasses)} element={element} tagClass={tagClass} {...rest} {...passedData} />
```
This is needed to support the lock icons on edge tags.

jshaughn (changes requested):
I think all of the prior comments are resolved!  And we have decided to temporarily remove the tag hide/hover customization and contribute that feature to PFT itself.  So that feature is expected to not be there.

There is a regression in showing lock icons on edge labels, this can be fixed by accepting the recommended changes in my comments.

Also, I'm seeing maybe an issue in trace overlay, but I don't think it has to do with this PR, it may be nothing, but it seems sometimes the overlay stops before the selected node.  Anyway, not a blocker for this.

I think if you fix the lock icon issue we can approve.

leandroberetta:
> I think all of the prior comments are resolved! And we have decided to temporarily remove the tag hide/hover customization and contribute that feature to PFT itself. So that feature is expected to not be there.
> 
> There is a regression in showing lock icons on edge labels, this can be fixed by accepting the recommended changes in my comments.
> 
> Also, I'm seeing maybe an issue in trace overlay, but I don't think it has to do with this PR, it may be nothing, but it seems sometimes the overlay stops before the selected node. Anyway, not a blocker for this.
> 
> I think if you fix the lock icon issue we can approve.

Thanks for your review @jshaughn.

I vote for remove for now the hide/hover feature (there is a PR already for that in PF repo: https://github.com/patternfly/react-topology/pull/123, this issue is mentioned in the epic: https://github.com/kiali/openshift-servicemesh-plugin/issues/99). 

I will address the feedback.

Issue #2923, opened 2020-06-22
Internationalization (I18N) support for Kiali UI


gpscode:

jshaughn:
@gpscode I'm not aware of any plan for I18N in Kiali, sorry.   If there were anyone interested in contributing we'd be open to it, but it would be a large effort.

lucasponce:
There is no plan to support I18N.
Probably this should be triggered by some downstream/product requeriment first.

dannyxu2015:
yeah, we need it to support Chinese, and my colleagues have interest to do the work, how can we start?

jshaughn:
Hi @dannyxu2015 , there is no support in the code for I18N, and so it is a much larger effort than just providing a language translation.

dannyxu2015:
@jshaughn yes, we are working on it

jshaughn:
Hi @dannyxu2015 , Well that is interesting :-)   I expect this would be a large effort.  If you are considering offering this effort to the project then I'd recommend letting us know more details about the approach and design, and working with us to ensure it would be a feasible addition.  Of course, if your effort is fork then you are free to do what you need!  It's cool that you like Kiali enough to try and make it available to native Chinese users.

dannyxu2015:
@jshaughn It's great to have your help, my colleague is working on it, we'll let you know the details later

BlueBiuBiu:
@jshaughn @dannyxu2015 Hello! Based on this project, I attempted to implement multilingual support for the frontend section. I extracted the non-dynamic words and phrases from the project (though I'm not sure if all of them were completely extracted), and created a separate configuration file for each language. Additionally, I added a switch button to facilitate language switching. This can serve as a template or reference for adapting to more languages. To extend support to other languages, simply copy the en.json file and translate the content into the desired language. Currently, a zh.json configuration file is included, enabling switching between Chinese and English.This is the GitHub repository link for the project.[link](https://github.com/BlueBiuBiu/kiali_frontend_multilingual_configuration)

jmazzitelli:
> Based on this project, I attempted to implement multilingual support for the frontend section.  ... This is the GitHub repository link for the project.[link](https://github.com/BlueBiuBiu/kiali_frontend_multilingual_configuration)

[...]

PR #6843, opened 2023-11-14
I18n support

ferhoyos:
### Describe the change

This PR adds i18n support to Kiali application using `i18next` library and some examples of Chinese translation in different kind of React components (classes, hooks, etc.). Translation of Kiali application in any language is out of scope of this PR.

Translation examples provided are:.
- Navigation menu (external function)
- Home cluster name tooltip in the top navigation menu (React hook)
- Number of labels per namespace (React class)

image

A language dropdown has been created to switch between English and Chinese. New locale value is stored in redux store so the value is persisted on refresh)

image

For the moment supported languages are English (default) and Chinese (in progress).

Translations are stored in json files in `public/locales/${LOCALE}` folder and generated automatically by `i18next-parser` library on yarn command `yarn i18n`. This command is executed automatically on build process to add new statements to translation files (or removed unused ones). English translation file must not be modified since code text is the source of truth for English statements.

I have added two i18n settings in the Kiali server configuration under `kiali_feature_flags/ui_defaults/i18n`:
- language: Default language (if no value is stored in redux store)
- showSelector: If true shows selector language icon

Since Kiali code is the base of OSSMC plugin, I have tried to follow OpenShift dynamic plugin requirements for i18n labels (https://github.com/openshift/console-plugin-template/tree/main?tab=readme-ov-file#i18n):
- Added I18N_NAMESPACE variable as i18n translation namespace (empty for Kiali, proper plugin value for OSSMC).

```
const { t } = useTranslation(I18N_NAMESPACE);
```

- Key should be English value since key separators are not allowed.

```
t('Traffic Graph')
```

### How to translate different React components
**React class**: 
Use `withTranslation` high order component (https://react.i18next.com/latest/withtranslation-hoc)

You have to embed React class with `withTranslation` function (similar to redux `connect` function):
```
export const Menu = withTranslation(I18N_NAMESPACE)(MenuComponent);
```

Besides, `WithTranslation` props must be added to React class properties (like Redux properties)

```
type MenuProps = WithTranslation & {
   ...
};
```

**React hook**
Use `useTranslation` hook (https://react.i18next.com/latest/usetranslation-hook)

```
const { t } = useTranslation(I18N_NAMESPACE);
```

**External function**
Include `t` function as input parameter:

```
const navMenuItems = (t: TFunction): MenuItem[] => {
  return [
    {
      id: 'overview',
      title: t('Overview'),
      to: '/overview',
      pathsActive: [/^\/overview\/(.*)/]
    },


[...]

PR #6877, opened 2023-11-22
Kiali frontend multilingual configuration

BlueBiuBiu:
Closes: #2923

** Describe the change **

The frontend module is multi-language compatible, and is set to English by default.
```ts
const $t: (key: string, defaultValue?: string) => string;
```
For example  
```tsx  
{$t('tip160', 'An error occured while fetching this metric')}
```
If the translation configuration file corresponding to tip160 can be found, the translation corresponding to the configuration file will be displayed. Otherwise, the default value of the second parameter will be used as the translation for display.

** Issue reference **


frontend/package.json
```diff
@@ -15,6 +15,7 @@
     "Instead update 'web_root' in Kaili config map in OpenShift console."
   ],
   "homepage": "./",
+  "proxy": "http://192.168.2.140:20001/kiali",
```
aljesusg, line 18: You need to remove this line

frontend/.eslintrc
```diff
@@ -7,6 +7,6 @@
     "no-unused-variable": "off",
     "no-restricted-globals": ["error", "event"],
     "no-undef": "off",
-    "import/no-default-export": "error"
+    "import/no-default-export": "off"
```
aljesusg, line 10: Why  we need to set this to off ?

aljesusg:
Hi @BlueBiuBiu , Thanks for this contribution I'll take a look

frontend/src/actions/LoginThunkActions.ts
```diff
@@ -36,7 +36,10 @@ const performLogin = (dispatch: KialiDispatch, state: KialiAppState, data?: any)
         error => bail(error)
       );
     } else {
-      bail({ status: AuthResult.FAILURE, error: 'Preparation for login failed, try again.' });
+      bail({
+        status: AuthResult.FAILURE,
+        error: $t('tip153', 'Preparation for login failed, try again.')
```
aljesusg, line 41: Can we set another name instead of tip? Something with more sense like 'login_failed_msg'

frontend/src/actions/MetricsStatsThunkActions.ts
```diff
@@ -34,7 +34,7 @@ export const MetricsStatsThunkActions = {
             }
           })
           .catch(err => {
-            addError('Could not fetch metrics stats.', err);
+            addError($t('tip154', 'Could not fetch metrics stats.'), err);
```
aljesusg, line 37: Same replacement of tip

frontend/src/actions/TracingThunkActions.ts
```diff
@@ -32,7 +32,7 @@ export const TracingThunkActions = {
             dispatch(TracingActions.setTrace(undefined));
             dispatch(TracingActions.tabTrace(undefined));
             AlertUtils.addMessage({
-              ...AlertUtils.extractAxiosError('Could not fetch trace', error),
+              ...AlertUtils.extractAxiosError($t('tip155', 'Could not fetch trace'), error),
```
aljesusg, line 35: Same replacement of tip

frontend/src/app/AuthenticationController.tsx
```diff
@@ -68,8 +68,10 @@ class AuthenticationControllerComponent extends React.Component<
   AuthenticationControllerProps,
   AuthenticationControllerState
 > {
-  static readonly PostLoginErrorMsg = `Kiali failed to initialize. Please ensure that services
-    Kiali depends on, such as Prometheus, are healthy and reachable by Kiali then refresh your browser.`;
+  static readonly PostLoginErrorMsg = $t(
+    'tip156',
```
aljesusg, line 72: Same replacement of tip

frontend/src/components/About/AboutUIModal.tsx
```diff
@@ -75,7 +74,7 @@ export const AboutUIModal: React.FC<AboutUIModalProps> = (props: AboutUIModalPro
     const grafanaService = externalServices.find(service => service.name === 'Grafana');
 
     if (tempoService && grafanaService && grafanaService.url) {
-      tempoService.url = GetTracingURL(externalServices);
+      tempoService.url = `${grafanaService.url}/explore?left={"queries":[{"datasource":{"type":"tempo"},"queryType":"nativeSearch"}]}`;
```
aljesusg, line 77: Please let GetTracingUrl method

frontend/src/components/DebugInformation/DebugInformation.tsx
```diff
@@ -189,68 +188,79 @@ const DebugInformationComponent: React.FC<DebugInformationProps> = (props: Debug
...
+    return conf;
+  };
 
-  const renderTabs = (): React.ReactNode[] => {
+  const renderTabs = () => {
     const kialiConfig = (
-      <Tab eventKey={0} title="Kiali Config" key="kialiConfig">
-        <CopyToClipboard onCopy={copyCallback} text={rows} options={copyToClipboardOptions}>
-          <SimpleTable label="Debug Information" className={tableStyle} columns={columns} rows={rows} />
+      <Tab eventKey={0} title={$t('KialiConfig', 'Kiali Config')} key="kialiConfig">
```
aljesusg, line 231: Please let SimpleTable component I don't know why we need this change

aljesusg (changes requested):
Ok It seems that we need to make some changes like no replace components and change the tips naming

jshaughn (commented):
@BlueBiuBiu Thanks for creating the fork and generating a PR!   I agree with @aljesusg  , and I think @ferhoyos made a similar comment in the past, I think it would be good to avoid the numerical`tipNUM` keys.   The numerics don't really have any meaning and you have to reference the locale file to know what is the next number.   Instead I think we could use keys that have some semantic meaning.  So, for example, instead of:

` error: $t('tip153', 'Preparation for login failed, try again.')`

Maybe use something like:

` error: $t('login.fail.prep', 'Preparation for login failed, try again.')`

In this approach you can use a series of key segments that:
1. have meaning
2. can be grouped in the locale file
3. are easily searchable in the IDE to see what exists.

In the example above, you have grouping1="login", grouping2="fail" and then the final discriminator.   You don't always need 3 groups, you could have 1,2,3 or more, whatever makes sense.  For one-word translations the key could just be the word.

frontend/package.json
BlueBiuBiu: OK, I will remove this line when I organize it later.

frontend/.eslintrc
BlueBiuBiu: The default export was used when exporting i18n.

frontend/src/actions/LoginThunkActions.ts
BlueBiuBiu: I will sort out the translation later and try to make sure that I can understand the meaning of the corresponding expression when I see the name.

BlueBiuBiu:
> @BlueBiuBiu Thanks for creating the fork and generating a PR! I agree with @aljesusg , and I think @ferhoyos made a similar comment in the past, I think it would be good to avoid the numerical`tipNUM` keys. The numerics don't really have any meaning and you have to reference the locale file to know what is the next number. Instead I think we could use keys that have some semantic meaning. So, for example, instead of:
> 
> ` error: $t('tip153', 'Preparation for login failed, try again.')`
> 
> Maybe use something like:
> 
> ` error: $t('login.fail.prep', 'Preparation for login failed, try again.')`
> 
> In this approach you can use a series of key segments that:
> 
> 1. have meaning
> 2. can be grouped in the locale file
> 3. are easily searchable in the IDE to see what exists.
> 
> In the example above, you have grouping1="login", grouping2="fail" and then the final discriminator. You don't always need 3 groups, you could have 1,2,3 or more, whatever makes sense. For one-word translations the key could just be the word.

At first, extracting these languages may be a bit rough, but later I will organize the translations and try to make sure that I can understand the meaning of the corresponding expressions just by looking at the name.

frontend/src/components/DebugInformation/DebugInformation.tsx
BlueBiuBiu: I just looked at the code, and it seems like I just changed the title, and the rest of the code is still there. Is this a bug in the review?

ferhoyos (changes requested):
@BlueBiuBiu Thanks for your PR! You have done huge effort to translate Kiali application :-). The translation works fine although there are some points to be discussed/enhanced:
- Keys with semantic meaning (mentioned in this PR by @jshaughn and @aljesusg)
- Dynamic values within the translation
- Plurals
- Use WithTranslation and useTranslation to avoid page reload when you change locale
- I18n parser to create locale files

I have created PR to address these issues (https://github.com/kiali/kiali/pull/6843). Just a few labels have been translated to check the enhancements. Please take a look and let me know your opinion.

frontend/src/locales/i18n.ts
```diff
@@ -0,0 +1,25 @@
...
+    }
+  },
+  lng: localStorage.getItem('locale') || 'en',
+  interpolation: {
+    escapeValue: false
+  }
+});
+
+// 将全局翻译方法绑定到window对象上
+(window as any).$t = i18n.t
```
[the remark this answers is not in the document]
BlueBiuBiu: > The issue using i18n.t is that page needs to be reloaded to perform translation. With React applications it is recommended to use WithTranslation function (Class components) and useTranslation hooks (Function components)

The original purpose of using a globally bound method is to separate translations from code as much as possible. I think using the useTranslation hooks is also a good approach, but for some static files, it may increase the difficulty of translation. With a large number of project files, using a global approach eliminates the need to consider where it is being used specifically, what do you think?

BlueBiuBiu:
@aljesusg @jshaughn @ferhoyos I have reorganized my previous rough translation template file to make it as clear as possible. What do you think?
[sorted_en.json](https://github.com/kiali/kiali/files/13486701/sorted_en.json)

frontend/src/locales/i18n.ts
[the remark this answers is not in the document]
ferhoyos: I understand your reason and it does make sense. The issue of using i18n.t is that translations won't be updated when you change the locale (for this reason you reload the whole page when you change the locale). On the other hand, useTranslation hook and WithTranslation function update the translations because they are part of the rendering process (see https://react.i18next.com/guides/the-drawbacks-of-other-i18n-solutions).

ferhoyos:
@BlueBiuBiu Much better with sorted_en.json, the keys are sorted and have semantic meaning. What about dynamic values with interpolation? Are you taking into account for translations?

BlueBiuBiu:
> @BlueBiuBiu Much better with sorted_en.json, the keys are sorted and have semantic meaning. What about dynamic values with interpolation? Are you taking into account for translations?

The interpolation problem can be solved by 
```ts
const $t: (key: string, defaultValue?: Record<string, string> | string) => string;
```
for example
```ts
$t('ShowDetail', { detail: 'detail' || 'defaultDetail' })

{
  "ShowDetail": "Show {{detail}}"
}
```

jshaughn:
Ho @BlueBiuBiu , @ferhoyos , any updates?

ferhoyos:
@jshaughn no update from my side, I like sorted template provided by @BlueBiuBiu with semantic meaning keys but I am still thinking that it is better to use useTranslation hook and WithTranslation function (more integrated with React rendering process) than global bound variable (see comment https://github.com/kiali/kiali/pull/6877#discussion_r1407558925)

BlueBiuBiu:
> > @BlueBiuBiu Much better with sorted_en.json, the keys are sorted and have semantic meaning. What about dynamic values with interpolation? Are you taking into account for translations?
> 
> The interpolation problem can be solved by
> 
> ```ts
> const $t: (key: string, defaultValue?: Record<string, string> | string) => string;
> ```
> 
> for example
> 
> ```ts
> $t('ShowDetail', { detail: 'detail' || 'defaultDetail' })
> 
> {
>   "ShowDetail": "Show {{detail}}"
> }
> ```
I will start internationalization operations in this way. If you need to use the useTranslation hook and WithTranslation function later for iteration, what do you think? @jshaughn @ferhoyos

BlueBiuBiu:
Hello, I have revised and submitted a new update according to the above requirements @ferhoyos @jshaughn @aljesusg

ferhoyos:
Thanks for the update @BlueBiuBiu! Now it looks much better with more semantic i18n message codes. I have noticed that this PR has some conflicts to be resolved. Could you resolve them please? If you have any issue or doubt in any conflict resolution please let us know.

BlueBiuBiu:
> Thanks for the update @BlueBiuBiu! Now it looks much better with more semantic i18n message codes. I have noticed that this PR has some conflicts to be resolved. Could you resolve them please? If you have any issue or doubt in any conflict resolution please let us know.

I just checked and found that the code and some of the structure on the page have undergone significant changes. Relying on git's code conflict analysis is not accurate. If I want to internationalize, I may need to pull the latest branch of the code and start writing translations from scratch.

ferhoyos:
Sorry to hear that @BlueBiuBiu. Recently we have upgraded to Patternfly version 5 and most part of the frontend code has been refactorized. I hope the internationalization does not take too long.

BlueBiuBiu:
> Sorry to hear that @BlueBiuBiu. Recently we have upgraded to Patternfly version 5 and most part of the frontend code has been refactorized. I hope the internationalization does not take too long.

I will start to re-translate the new version for internationalization tomorrow, which may take a little time

BlueBiuBiu:
this is the latest internationalization configuration for the front-end code after the upgrade of Kiali version. https://github.com/kiali/kiali/pull/7014

PR #6094, opened 2023-05-03
Cluster badge on the Overview page

leandroberetta:
Resolves https://github.com/kiali/kiali/issues/5921.

This PR adds the cluster badge to all views. It also fixes a problem in list view that wasn't showing the control plane metrics.

image
image

PR #6882, opened 2023-11-23
Test coverage for MC badges on the Overview page

matejnesuta:
Some additional checks related to #6094.


josunect (approved):
LGTM

Just a note, the statistic for the remote cluster have been removed as they are not available (See here: https://github.com/kiali/kiali/pull/6514#issuecomment-1733147536), but I don't think this is validated for the remote cluster.

Issue #6881, opened 2023-11-22
tracing UI - hover over trace dots "flickers" the heat map


jmazzitelli:
Following the test instructions [here](https://github.com/kiali/kiali/pull/6853#issuecomment-1818855128), go to a Traces tab view such as this:

![image](https://github.com/kiali/kiali/assets/2029470/1e6b9d25-a081-4cfa-8396-7d112e1673c3)

And  try to hover over a trace dot.. some of them flicker, going from "Loading trace details" to showing the heat map (it flickers back and forth as you slightly move the mouse).

See this video (unfortunately, you can't see the mouse pointer, but I am hovering over a trace dot and every so slightly moving the mouse around that dot):
[trace-hover.webm](https://github.com/kiali/kiali/assets/2029470/5a242fac-0e8b-442f-a9bb-f23c9c2382b0)

PR #6883, opened 2023-11-23
Hover trace update - Tempo

josunect:
Fixes https://github.com/kiali/kiali/issues/6881

This PR simplifies the complexity of the code by removing the "tabTrace" property. This property was introduced to reload the trace data onhover for Tempo, because it does not include all the trace information returned by the search query. 

Now the http client uses the ".component", ".upstream_cluster" properties in the select query because they are used to create the heat map. 

There is still a small "issue" with Tempo: once we click on the trace, all the trace information will be complete and the heat map will be recalculated and the number of spans will be updated. Maybe it would be better to add a message for Tempo like "Click to load the complete Trace details"?


hhovsepy:
I agree to show to user some info "Click to load the complete Trace details "

jshaughn:
I'm -1 on ever having to click anything in a tooltip.  If we can't figure out how to make hover handle the heat-map seamlessly, then I'd vote to remove the heat-map completely from the tooltip.

jmazzitelli:
The flickering has stopped, but this doesn't look good. Sometimes the heatmap shows up, other times not, and I can't tell what causes it to load. I think if I click the dot, it loads after a second. OR sometimes I just hover and it might load the heatmap (but other times it doesn't).

[Screencast from 2023-11-27 12-11-52.webm](https://github.com/kiali/kiali/assets/2029470/f7c4a8db-b8bf-4c51-b843-770c21ce486a)

I would say if this is a performance issue, I would just take out the heatmap from the tooltip. Otherwise, this behavior will confuse and frustrate the user.

jmazzitelli (commented):
Well, it doesn't flicker anymore... but I'm not sure this works like we want. I don't want to approve this the way it is, because I don't think we want it like this.

josunect:
So, I was thinking the best way to solve this, I've some ideas: 

1) I think the ideal would be to have the same data as in Jaeger query, so I will open an issue in Grafana Tempo to see if it is possible. 
2) At the meantime, we can hide the heatmap. 
3) Another idea would be to do a query from all the traces in the backend, but this will have a poor performance, it could be many queries. 
4) Load the trace info on hover. But this could have many requests at the same time, it should be "controlled" somehow.

josunect:
I've changed the Jaeger query so it get all the data needed. 

There are a slightly differences in some of the traces, it looks this is because how TraceQL works, as it evaluates blocks of traces which could be different than the complete trace.

josunect:
@hhovsepy @jmazzitelli maybe you could have a look at these changes? Now it should work almost the same as with Jaeger. Thank you!

jmazzitelli:
OK, so I can say it works better. No flickering and the heatmaps appear more consistently in a manner that I can figure out. However, it is 100% there. If we can to keep it like it is for now, I'm OK with it. But if there is an easy solution to this, we should fix it.

When you hover over a dot that hasn't had its heatmap loaded yet, you see `n/a` where the heatmap would be. If you then click the dot, the heatmap data is loaded, but it is NOT displayed in the tooltip. But if I move the mouse just a single pixel, then the heatmap is displayed.

So in order to see the heatmap in the tooltip, the user needs to know to do two things: (1) click the dot and (2) move the mouse slightly (but do not leave the dot). It would be nice if (2) wasn't required, but I do not know how easy that would be.

[Screencast from 2023-12-01 08-55-51.webm](https://github.com/kiali/kiali/assets/2029470/a40a31fe-f19d-44f6-8093-9fed8eeccb9c)

hhovsepy (approved):
Tested locally, not it works fine, Tempo traces are loaded, and spans are loaded while clicking.
For local test env applied the memory settings:
```
kubectl apply -n tempo -f - <<EOF
apiVersion: [tempo.grafana.com/v1alpha1](http://tempo.grafana.com/v1alpha1)
kind: TempoStack
metadata:
  name: cr
spec:
  storageSize: 1Gi
  storage:
    secret:
      type: s3
      name: tempostack-dev-minio
  resources:
    total:
      limits:
        memory: 4Gi
        cpu: 8000m
  template:
    queryFrontend:
      jaegerQuery:
        enabled: false
EOF
```

josunect:
> OK, so I can say it works better. No flickering and the heatmaps appear more consistently in a manner that I can figure out. However, it is not 100% there. If we want to keep it like it is for now, I'm OK with it. But if there is an easy solution to this, we should fix it.
> 
> When you hover over a dot that hasn't had its heatmap loaded yet, you see `n/a` where the heatmap would be. If you then click the dot, the heatmap data is loaded, but it is NOT displayed in the tooltip. But if I move the mouse just a single pixel, then the heatmap is displayed.
> 
> So in order to see the heatmap in the tooltip, the user needs to know to do two things: (1) click the dot and (2) move the mouse slightly (but do not leave the dot). It would be nice if (2) wasn't required, but I do not know how easy that would be.
> Screencast.from.2023-12-01.08-55-51.webm

@jmazzitelli is the backend updated? I would need to be refreshed as well, because now it should send all the trace information (And we shouldn't see the n/a anymore)

[Screencast from 01-12-23 15:16:00.webm](https://github.com/kiali/kiali/assets/49480155/18a82e08-72ff-4692-aa18-8ba578678e55)

jmazzitelli:
I still see the same behavior as my previous screencast video. I see `n/a`, I click the dot, move the mouse, then I see the heatmap.

I confirmed in the About box that I am looking at my latest build that should include all of the backend changes in this PR

josunect:
> I still see the same behavior as my previous screencast video. I see `n/a`, I click the dot, move the mouse, then I see the heatmap.
> 
> I confirmed in the About box that I am looking at my latest build that should include all of the backend changes in this PR.
> 
> I see logs like this:
> 
> ```
> 2023-12-01T11:28:26-05:00 DBG QueryQL { .service.name = "details.bookinfo"  && .node_id =~ ".*"   } && {  } | select(status, .service_name, .node_id)
> ```

@jmazzitelli are you using gRPC? I'm sorry, I forgot to update the gRPC client, so that would explain it... I have updated the query in the gRPC client too.

jmazzitelli:
> you using gRPC? I'm sorry, I forgot to update the gRPC client, so that would explain it... I have updated the query in the gRPC client too.

I was following the [test instructions linked here](https://github.com/kiali/kiali/pull/6883#issuecomment-1836359398).

jmazzitelli (approved):
I just ran everything with the latest from this PR and now I always see the heatmap in the tooltip and I don't see any flickering and no need to click anything. The only thing I noticed (which I believe is expected) is it takes about a second or two to load the traces tab view.

josunect:
> See my followup issue here:
> 
>     * [trace hover tooltip performance boost #6915](https://github.com/kiali/kiali/issues/6915)
> 
> 
> You can do that work here in this PR, or we can do that in a followup PR (assuming people agree with my idea of getting rid of the heatmaps in the tooltips).

I think that if we remove the heatmap from the tooltip we can try with a query more simple and improve the performance in the call to Tempo. 
I think we can work on a follow up PR.

PR #6897, opened 2023-11-29
New eslint config for precommit linting

### LINKED ISSUE #6893: Apply new eslint rules only to edited files
Due to Kiali application size it is very hard to introduce new eslint rules like https://github.com/kiali/kiali/issues/6736 (the number of errors after applying the eslint rule to whole Kiali application is so high that is impossible to handle correctly). For this reason I think we could apply new eslint rules only to edited files where developer is working (`yarn lint:precommit` command). 

With this approach the new eslint rule is applied little by little to Kiali applicatiion. After some time these eslint rules could be incorporated to general eslint configuration when the number of errors is manageable.

ferhoyos:
**Describe the change**

I have configured `yarn eslint:precommit` command with different eslint config file to apply new rules only to committed files.

For the moment the new rules I have applied are:
- "no-var": Use const or let instead of var
- "prefer-template": Use template interpolation instead of string literals (https://github.com/kiali/kiali/issues/6736)
- "prefer-arrow-callback": Use arrow functions instead of normal functions
- "@typescript-eslint/no-inferrable-types": Avoid typing the type if a default value is set (to reduce verbosity)
- "@typescript-eslint/explicit-function-return-type": Return types have to be specified in functions
- "@typescript-eslint/explicit-module-boundary-types": Method types have to be specified in functions

Besides, I have updated Husky (library that performs pre-commit linting) to latest version 8.0.3

**Issue reference**

Closes #6893

### DIFF
--- a/frontend/.eslintrc_precommit
+++ b/frontend/.eslintrc_precommit
@@ -1,0 +1,20 @@
+{
+  "extends": "react-app",
+  "rules": {
+    "jsx-a11y/anchor-is-valid": "off",
+    "jsx-a11y/anchor-has-content": "off",
+    "no-used-before-declare": "off",
+    "no-unused-variable": "off",
+    "no-restricted-globals": ["error", "event"],
+    "no-undef": "off",
+    "import/no-default-export": "error",
+
+    // New eslint rules
+    "no-var": "error",
+    "prefer-template": "error",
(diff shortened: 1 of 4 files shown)


jshaughn (approved):
This looks OK to me, we can always make changes if it becomes too much or is too little.

PR #6638, opened 2023-09-19
Added missing cluster in K8s Gateway validations.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6633

Now the validation results will be consistent in lists and details of K8s Gateway objects.

Create 2 K8s Gateways with the same host and port combination. Warning will be shown for both of them.

PR #6901, opened 2023-11-30
Additional checks related to an inconsistency in an Istio config list

matejnesuta:
Related to #6638.


hhovsepy (approved):
thank you @matejnesuta

PR #6904, opened 2023-12-01
Improve appearance of workload log checkboxes

### LINKED ISSUE #6811: Workload logs - improve appearance of checkboxes
This might just be personal preference but the appearance of the checkboxes doesn't look great and doesn't fit the style of other checkboxes in the UI:

Screenshot from 2023-11-02 09-07-59

Compare to the metrics page checkboxes:

Screenshot from 2023-11-02 09-10-42

This is probably just a matter of getting rid of the background highlighting.

ferhoyos:
**Describe the change**

I have removed black background of workload log checkboxes that makes the look and feel a bit weird. In order to maintain the legend of integrated logs, I have change the color of the checkbox itself with `accent-color` CSS class. Also I have adapted log colors to work fine both in light and dark themes and adjust some margins and paddings of log lines.

image

For first application container I have selected blue color. These are the other colors selected for multiple application containers:
image

image

image

Besides, I have added new eslint rule `@typescript-eslint/member-ordering` to force type and interface properties to be sorted alphabetically and modify `@typescript-eslint/explicit-function-return-type` to allow expressions (probably more rules will have to be adjusted in the next PRs until we find an optimal set of rules to apply in edited files).

**Issue reference**

Closes #6811

### DIFF
--- a/frontend/.eslintrc_precommit
+++ b/frontend/.eslintrc_precommit
@@ -12,8 +12,13 @@
     "no-var": "error",
     "prefer-template": "error",
     "prefer-arrow-callback": "error",
+    "@typescript-eslint/member-ordering": ["error", {
+      "default": {},
+      "interfaces": {"order": "alphabetically"},
+      "typeLiterals": {"order": "alphabetically"}
+    }],
     "@typescript-eslint/no-inferrable-types": "error",
-    "@typescript-eslint/explicit-function-return-type": "error",
+    "@typescript-eslint/explicit-function-return-type": ["error", { "allowExpressions": true }],
     "@typescript-eslint/explicit-module-boundary-types": ["error", { "allowArgumentsExplicitlyTypedAsAny": true }]
   }
 }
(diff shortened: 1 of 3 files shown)


ferhoyos:
I have noticed that Show/Hide filters do not filter span log messages. Is this the expected behaviour?

jshaughn (approved):
This looks fine for me.  As for the Show/Hide behavior with spans, I'm actually not sure.  But I can see the benefit of leaving it as-is.  The logs are intended to help the user identify spans and traces of interest.  I think the spans themselves are mainly for clicking into.  I would leave it alone unless an actual user has a compelling reason to make a change.

nrfox:
> I have noticed that Show/Hide filters do not filter span log messages. Is this the expected behaviour?

I think it's expected. The find/hide filter is just a text search but spans have attributes and aren't text so filtering over them with find/hide would be strange.

nrfox (approved):
It looks much better!

PR #6925, opened 2023-12-05
Check CRD for Gateway API

hhovsepy:
** Describe the change **

_explanation of what it does_

** Issue reference **


nrfox (commented):
@hhovsepy pulling in the CRD client might also require giving Kiali permissions to get/list CRDs. Instead of pulling in another dependency, can we use the `Discovery()` interface for the gatewayapiclient to check if the resources exist? https://pkg.go.dev/sigs.k8s.io/gateway-api@v0.5.1/pkg/client/clientset/versioned#Interface

hhovsepy:
> @hhovsepy pulling in the CRD client might also require giving Kiali permissions to get/list CRDs. Instead of pulling in another dependency, can we use the `Discovery()` interface for the gatewayapiclient to check if the resources exist? https://pkg.go.dev/sigs.k8s.io/gateway-api@v0.5.1/pkg/client/clientset/versioned#Interface

@nrfox good idea with possible missing access to CRDs, wdyt how else we can check that the CRDs are installed without retrieving them? because in original check `k8s.Discovery().RESTClient().Get().AbsPath("/apis/networking.istio.io").Do(in.ctx).Raw()`  we can check the installed versions not separate CRD existence.

nrfox:
> > @hhovsepy pulling in the CRD client might also require giving Kiali permissions to get/list CRDs. Instead of pulling in another dependency, can we use the `Discovery()` interface for the gatewayapiclient to check if the resources exist? https://pkg.go.dev/sigs.k8s.io/gateway-api@v0.5.1/pkg/client/clientset/versioned#Interface
> 
> @nrfox good idea with possible missing access to CRDs, wdyt how else we can check that the CRDs are installed without retrieving them? because in original check `k8s.Discovery().RESTClient().Get().AbsPath("/apis/networking.istio.io").Do(in.ctx).Raw()` we can check the installed versions not separate CRD existence.

I think you can do something like this: `client.Discovery().ServerResourcesForGroupVersion(groupVersion)` to list all of the resources and determine if `ReferenceGrant` is among them without needing to get/list CRDs.

hhovsepy:
@nrfox the comments approached, ready for review.

kubernetes/kubernetes.go
```diff
@@ -208,18 +208,36 @@ func (in *K8SClient) IsGatewayAPI() bool {
...
-		}
+		isGatewayAPI := checkGatewayAPIs(in)
 		in.isGatewayAPI = &isGatewayAPI
 	}
 	return *in.isGatewayAPI
 }
 
+func checkGatewayAPIs(in *K8SClient) bool {
+	found := 0
+	res, err := in.k8s.Discovery().ServerResourcesForGroupVersion(K8sNetworkingGroupVersionV1.String())
```
jshaughn, line 219: This looks good, it incorporates Nick's feedback.

kubernetes/kubernetes.go
```diff
@@ -208,18 +208,36 @@ func (in *K8SClient) IsGatewayAPI() bool {
...
+	if err != nil {
+		log.Debugf("Error while checking K8s Gateway API CRDs: %s. K8s Gateway API will not be used.", err.Error())
```
jshaughn, line 221: Should this error message indicate which version we are expecting to find?  Or is that going to be in the Error string?

kubernetes/kubernetes.go
```diff
@@ -208,18 +208,36 @@ func (in *K8SClient) IsGatewayAPI() bool {
...
+		K8sGatewayClassType:    K8sActualGatewayClasses,
+		K8sActualHTTPRouteType: K8sActualHTTPRoutes,
+	}
+	for _, r := range res.APIResources {
+		if name, foundKind := types[r.Kind]; foundKind && r.Name == name {
+			found++
+		}
+	}
+	if found > 0 && found < len(types) {
+		log.Warningf("Not all required K8s Gateway API CRDs are installed.")
```
jshaughn, line 236: Should this message specify which are missing?

jshaughn (approved):
So with this PR it means we require v1 CRDs to enable Gateway API features, right?  This should be fine, it just means that we'd require V1.20 or higher for Istio, right?

I made a few minor comments, take a look and decide if you want to make any changes before merge, but I will approve.

hhovsepy:
Thanks @jshaughn for the review, yes we will require "v1" of K8s Gateway API as it is released and Istio supports that.
Your suggestions are very valid, I will include them in my ongoing PR https://github.com/kiali/kiali/pull/6957.
As the 'K8sReferenceGrants' is still in 'v1beta1' (not sure why), then I will do a bit refactoring of CRD checking and include a better logging of what is missing.

Issue #6912, opened 2023-12-04
Set Secure Attribute on Session Cookie 


stanenbaum-va:
WASA scans reveal a low security finding when Kiali is served over HTTPS due to the `kiali-token-aes` only having `HTTPOnly`, without the `Secure` attribute.

jshaughn:
See https://github.com/kiali/kiali/discussions/6907 for more on what is being requested here.

jmazzitelli:
Here's the three locations where this enhancement will probably need to touch:
* https://github.com/kiali/kiali/blob/v1.77.0/business/authentication/session_persistor.go#L135-L141
* https://github.com/kiali/kiali/blob/v1.77.0/business/authentication/openid_auth_controller.go#L372-L378
* https://github.com/kiali/kiali/blob/v1.77.0/business/authentication/openid_auth_controller.go#L491-L497

jmazzitelli:
And here is a brief overview of why this is being asked. https://resources.infosecinstitute.com/topics/general-security/securing-cookies-httponly-secure-flags

Note: Because Kiali's endpoint can be configured to use http only (as opposed to https) we'll want to be able to have the cookie sent over http. Thus the Secure flag should be set on the cookie only if Kiali's endpoint is specifically configured for https.

jmazzitelli:
This has been implemented - https://github.com/kiali/kiali/pull/6930 - that PR is just waiting on review; if that happens in the next day or two, this might get into the upcoming release.

PR #6930, opened 2023-12-06
set the Secure flag on auth cookies when appropriate

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6912


config/config.go
```diff
@@ -873,6 +873,11 @@ func (conf *Config) AllNamespacesAccessible() bool {
 	return conf.Deployment.ClusterWideAccess
 }
 
+// IsServerHttps returns true if the server endpoint should use HTTPS. If false, only plaintext HTTP is supported.
+func (conf *Config) IsServerHttps() bool {
```
nrfox, line 877: Very minor nit that you can ignore if you like: but note that if you accept this suggestion you'll probably have to update it everywhere else.
```suggestion
func (conf *Config) IsServerHTTPS() bool {
```

nrfox (approved):

config/config.go
jmazzitelli: I originally had it that way, and the the Code editor made the suggestion to change it to camel case. What is our typical code convention here? I was just following the editor blindly :)

business/authentication/openid_auth_controller.go
```diff
@@ -372,6 +372,7 @@ func (c OpenIdAuthController) redirectToAuthServerHandler(w http.ResponseWriter,
 	nonceCookie := http.Cookie{
 		Expires:  expirationTime,
 		HttpOnly: true,
+		Secure:   conf.IsServerHttps(),
```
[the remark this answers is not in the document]
jmazzitelli: Yes, I almost think it has to be because the "Secure" flag is just that -- a flag in the cookie. So the set-cookie header doesn't have "Secure=true" or "Secure=false" its just "Secure" or not specified.

And yes the default is false.

e.g. https://http.dev/set-cookie#secure

config/config.go
nrfox: I think we're inconsistent but it's a generally accepted style for abbreviations in go: https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

jmazzitelli: done. See latest commit

business/authentication/openid_auth_controller.go
```diff
@@ -367,12 +367,14 @@ func (c OpenIdAuthController) redirectToAuthServerHandler(w http.ResponseWriter,
 		return
 	}
 
+	guessedKialiURL := httputil.GuessKialiURL(r)
+	secureFlag := conf.IsServerHTTPS() || strings.HasPrefix(guessedKialiURL, "https:")
```
nrfox, line 371: Since this is repeated a bunch of places now can it just be its own function? Something like `IsSecureCookie(...)`?

nrfox (approved):
1 comment but non-blocking

PR #6933, opened 2023-12-07
Fixed graph node decorators in PF5

### LINKED ISSUE #6926: patternfly graph not showing node decorators
The patternfly graph is no longer showing node decorators, it seems like a regression in PFT with PF5.  The issue in our code is [here](https://github.com/kiali/kiali/blob/master/frontend/src/pages/GraphPF/GraphPFElems.tsx#L162-L164): 

This used to work, but surrounding the decorator with a Tooltip now stops the decorator from rendering:
```
    <Tooltip content={!!tooltip ? tooltip : icon.text}>
      <Decorator x={x} y={y} radius={DEFAULT_DECORATOR_RADIUS} showBackground icon={React.createElement(icon.icon)} />
    </Tooltip>
```

We can do this, but we lose the tooltip, and the tooltip is important in order to explain what the decorator means.
```
      <Decorator x={x} y={y} radius={DEFAULT_DECORATOR_RADIUS} showBackground icon={React.createElement(icon.icon)} />
```

@leandroberetta , should we ask Jeff whether this is expected or if there is a workaround?

leandroberetta:
Fixes #6926 

A fix is introduced because decorators were failing due to a change introduced in PF5.

### DIFF
--- a/frontend/src/pages/GraphPF/GraphPFElems.tsx
+++ b/frontend/src/pages/GraphPF/GraphPFElems.tsx
@@ -38,7 +38,7 @@ import { PFColors } from 'components/Pf/PfColors';
 import { getEdgeHealth } from 'types/ErrorRate/GraphEdgeStatus';
 import { Span } from 'types/TracingInfo';
 import { IconType } from 'config/Icons';
-import { TrafficDecorator } from './components/TrafficDecorator';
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
LG!

PR #6934, opened 2023-12-07
emit selected trace from url

### LINKED ISSUE #5635: Link to trace does not always open trace details
Links to traces/span details are included in the logs viewer and on the metric charts. When you click on this link, you are redirected to the traces tab and the trace along with the trace details should be highlighted. This does not always work and sometimes you are redirected to the traces tab but with nothing open. Happens anywhere there's a link to an individual trace. Could be a disconnect between the traces/spans shown in the link vs. what is fetched from Kiali when you navigate to the traces tab.

josunect:
Fixes #5635 

I was able to reproduce it when the component had not been started, it looks like it was not getting the parameters from the url. 

The other changes are from the pre commit lint.

### DIFF
--- a/frontend/src/actions/TracingThunkActions.ts
+++ b/frontend/src/actions/TracingThunkActions.ts
@@ -8,7 +8,7 @@ import { setTraceId as setURLTraceId } from 'utils/SearchParamUtils';
 import { transformTraceData } from 'utils/tracing/TraceTransform';
 
 export const TracingThunkActions = {
-  setTraceId: (cluster?: string, traceId?: string) => {
+  setTraceId: (cluster?: string, traceId?: string): ((dispatch: KialiDispatch) => void) => {
     setURLTraceId(traceId);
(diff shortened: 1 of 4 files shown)


frontend/src/actions/TracingThunkActions.ts
```diff
@@ -17,7 +17,6 @@ export const TracingThunkActions = {
             if (response.data.data) {
               const trace = transformTraceData(response.data.data, cluster);
               if (trace) {
-                trace.loaded = true;
```
nrfox, line 20: Is this no longer needed?

nrfox (approved):

frontend/src/actions/TracingThunkActions.ts
josunect: No, this is not. It was used when not all the spans info was in the trace, so it was an indicator for the on-hover event. But now, all the info is in the first search call.

PR #6935, opened 2023-12-07
Initial Mesh Page Topology POC

### LINKED ISSUE #6840: Initial Mesh Page Topology POC
Epic #5913 

This issue is just to generate an initial POC for a Mesh topology page, to generate discussion and further task breakdown.  The general approach will be to steal from, and then adapt the traffic graph approach/code.  This will be natively a Patternfly Topology graph so as to be immediately usable in both Kiali, OSSMC and maybe Backstage.

jshaughn:
Closes #6840 

The new page is hidden behind a feature flag:

```
spec.kiali_feature_flags.ui_defaults.mesh.impl: classic | topo | topo-as-overview
```

The default is 'classic', which keeps today's [list-based] Mesh page and completely hides this POC Mesh Graph.  Set to 'topo' to replace the list Mesh with the new Graph Mesh.

The main purpose of this PR is to just get it into the master branch to be available to other maintainers, if there is a chance for them to work on some of the subtasks.  The next milestone in the Epic would be to replace the current Mesh page.


jmazzitelli:
changed title of PR - annoying just seeing an issue number in the title - can't tell what this PR is for!

prometheus/internalmetrics/internal_metrics.go
```diff
@@ -245,6 +248,34 @@ func GetGraphMarshalTimePrometheusTimer(graphKind string, graphType string, with
 	return timer
 }
 
+// GetMesgGraphGenerationTimePrometheusTimer returns a timer that can be used to store
```
ferhoyos, line 251:
```suggestion
// GetMeshGraphGenerationTimePrometheusTimer returns a timer that can be used to store
```
Little typo

mesh/options.go
```diff
@@ -0,0 +1,165 @@
...
+	"k8s.io/client-go/tools/clientcmd/api"
+
+	"github.com/kiali/kiali/business"
+	"github.com/kiali/kiali/business/authentication"
+)
+
+// The supported vendors
+const (
+	VendorCytoscape     string = "cytoscape"
+	defaultConfigVendor string = VendorCytoscape
```
ferhoyos, line 23: Mesh is only implemented with Patternfly, right? Why the `cytoscape` reference?

ferhoyos (changes requested):
As starting point, the Mesh graph looks promising. I have played with it and it works fine, except when you click on Grafana badge:
image

This is the console error:

image

I think I have addressed the issue in one of my code comments, just check that the issue is solved.

There are multiple things that still not working (e.g., find and hide feature), but they will be addressed in following PRs.

frontend/src/pages/Mesh/Mesh.tsx
```diff
@@ -0,0 +1,620 @@
...
+
+    // When the initial layoutName property is set it is premature to perform a layout
+    if (initialLayout) {
+      initialLayout = false;
+      return;
+    }
+
+    controller.getGraph().setLayout(layoutName);
+    controller.getGraph().layout();
+    if (requestFit) {
```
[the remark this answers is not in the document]
jshaughn: No good reason :-)   There are problems with the fit logic and it all needs to be cleaned up. The Epic was missing this, so I just added https://github.com/kiali/kiali/issues/7122.

mesh/options.go
jshaughn: You are right, but this is sort of a legacy that can be cleaned up if/when we remove Cytoscape.  The fact is that even for the Patternfly graph, the JSON we return to the client is in Cytoscape format.  Meaning, the JSON can be passed directly into Cytoscape to form the graph.  But for Patternfly, we take the same cytoscape-compatible JSON and just parse it ourselves to define the PFT elements.  So, in effect, the config vendor actually is Cytoscape, for both graph renderings.

frontend/src/components/Nav/Menu.tsx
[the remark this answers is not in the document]
jshaughn: Makes sense.

jshaughn:
@ferhoyos Thanks for the review and suggestions (some good eyes you have :-)).  I committed all of the suggestions and added replies to the comments.

frontend/src/pages/Mesh/target/TargetPanelNode.tsx
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  private renderNode = (infraName: string, infraType: MeshInfraType): React.ReactNode => {
+    let pfBadge;
+
+    switch (infraType) {
+      case MeshInfraType.CLUSTER:
+        pfBadge = PFBadges.Cluster;
+        break;
+      case MeshInfraType.ISTIOD:
```
ferhoyos, line 77:
```suggestion
      case MeshInfraType.GRAFANA:
        pfBadge = PFBadges.Grafana;
        break;
      case MeshInfraType.ISTIOD:
```
Grafana badge is missing. This is causing the error I commented in my review above.

ferhoyos (approved):
LGTM

PR #6943, opened 2023-12-08
Update of go dependencies

hhovsepy:
Further Gateway API implementation support tasks require gateway-api client version to be updated.
It causes update of several go library dependencies.


aljesusg (approved):
LGFM

PR #6944, opened 2023-12-08
[hack] on OpenShift when no ingress gateway is installed, wait for productpage route

jmazzitelli:
Note that this fixes a bug in the hack script - it was trying to expose the openshift routes before the services even existed. This also fixes that problem.


jshaughn (approved):
LOK

PR #6949, opened 2023-12-12
Include istio-waypoint api class as default

### LINKED ISSUE #6878: Include istio-waypoint gateway api class as a default
Kiali validation error KIA1504 for the istio-waypoint gatewayClassName.

image

After creating a waypoint proxy:

`istioctl x waypoint apply -n bookinfo`

Should be added as a default, considering that Istio waypoint is still experimental? 

```
external_services:
  istio:
    gateway_api_classes:
     - name: Waypoint
       class_name: istio-waypoint
     - name: Istio
       class_name: istio  
```

Ref. https://github.com/kiali/kiali/issues/6757

josunect:
Fixes #6878

image

### DIFF
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -374,7 +374,7 @@ func ClusterInfoFromIstiod(conf config.Config, k8s ClientInterface) (string, boo
 	return clusterName, gatewayToNamespace, nil
 }
 
-func GatewayAPIClasses() []config.GatewayAPIClass {
+func GatewayAPIClasses(ambientEnabled bool) []config.GatewayAPIClass {
 	result := []config.GatewayAPIClass{}
 	for _, gwClass := range config.Get().ExternalServices.Istio.GatewayAPIClasses {
 		if gwClass.ClassName != "" && gwClass.Name != "" {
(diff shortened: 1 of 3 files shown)


josunect:
> Hi @josunect , I would add an 'if' condition for checking `ambient` mode to include another default value in gatewayClasses, as in regular Gateway API installation there is only one gatewayClass:
> 
> ```
> $ oc get GatewayClasses --all-namespaces
> NAME           CONTROLLER                    ACCEPTED   AGE
> istio          istio.io/gateway-controller   True       4h
> istio-remote   istio.io/unmanaged-gateway    True       4h
> ```

Thanks for the suggestion, @hhovsepy , I've done some changes.

kubernetes/istio.go
```diff
@@ -384,5 +384,8 @@ func GatewayAPIClasses() []config.GatewayAPIClass {
 	if len(result) == 0 {
 		result = append(result, config.GatewayAPIClass{Name: "Istio", ClassName: "istio"})
 	}
+	if ambientEnabled {
```
hhovsepy, line 387: We need to put this condition inside the above `if len(result) == 0 {` as when configuration is already set in 'ExternalServices.Istio.GatewayAPIClasses' then we cannot add any default value. Configured values are taken, default values are set when no config provided.

hhovsepy (dismissed):
LGTM, Thank you @josunect

hhovsepy (approved):
looks good

Issue #6948, opened 2023-12-11
Istio config bug


Zorgji:
### Bug brief

I'm testing Istio with gateway API and httproutes. It's all accessible and working well. But in Kiali saying as below img.
<img width="1512" alt="Screenshot 2023-12-12 at 12 20 32" src="https://github.com/kiali/kiali/assets/36134132/4830a683-0532-446d-a46f-9b7c4df79b8a">

If I put full name like `dev-grafana.monitoring.svc.cluster.local`, Istio config is fine but can't route to the App anymore.
<img width="1512" alt="Screenshot 2023-12-12 at 12 27 23" src="https://github.com/kiali/kiali/assets/36134132/94c84c9e-99f0-4de9-bf16-6536a7e16b5c">
<img width="1512" alt="Screenshot 2023-12-12 at 12 29 50" src="https://github.com/kiali/kiali/assets/36134132/f5f97fd4-845d-408a-80b2-a14e4bea43b7">

### Environment
*Currently I'm running*

- **Kiali version: v1.78.0**
- **Istio version: 1.20.0** 
- **Kubernetes version: 1.26** 
- **Other notable environmental factors: Gateway API, HTTPROUTES**

jshaughn:
@hhovsepy Perhaps the validation here is too aggressive?

hhovsepy:
Hi @Zorgji , in this screenshot it still shows an error message over 'dev-grafana.monitoring.svc.cluster.local' service in FQDN, does this service exist with this FQDN?

[...]

PR #6954, opened 2023-12-13
Include FQDN in error list for K8s HTTPRoute backend ref service name.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6948

Before: Kiali was accepting FQDN in K8sHTTPRoute config's 'backendRef.name' as a Service name, but Gateway API expects it to be 'name' only. A bug!

Now: Error `KIA1402` will be extended to be shown also in FQDN formatted Service name (even if FQDN is correct). Error message is changed to be more clear that not a `host` is expected but `Service name`.

However, Kiali requires `istio_identity_domain: svc.cluster.local` set in `kiali` configmap, or not set that property at all to take the default.

### DIFF
--- a/business/checkers/k8shttproutes/no_host_checker.go
+++ b/business/checkers/k8shttproutes/no_host_checker.go
@@ -2,6 +2,7 @@ package k8shttproutes
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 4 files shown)


jshaughn (approved):
CI looks better now.  Code looks good.

PR #6956, opened 2023-12-13
Use kiali svc token to create shared tracing and prom clients

### LINKED ISSUE #6955: Tracing client must use the Kiali SA Token (Not the user token)
Based on the https://github.com/kiali/kiali/blob/2de84a65541ce24a0f78fa455a639e99291c7576/tracing/client.go#L88-L90 and https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.tracing.auth.use_kiali_token, the Tracing client must use the Kiali SA Token and not the users token, as the client is reused between sessions and users could be using token's from another users.

nrfox:
** Describe the change **

Creates prom and tracing clients at the top level that are then passed into the business layer. Tracing client now properly uses the kiali auth token when the [config opt](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.tracing.auth.use_kiali_token) is set instead of the user token.

** Issue reference **

Fixes #6955


josunect (commented):
I know this is not part of this issue, but doing some tests, I've realized a runtime error when we set everything for Jaeger but the provider is incorrect, it does not return any error but the tracing client is empty. E.g. 

```
  tracing:
    enabled: true
    provider: "jeager"
```

I think that could be solved by adding in kiali.go > validateconfig.go: 

```
	// Check the tracing section
	cfgexTracing := cfg.ExternalServices.Tracing
	if cfgexTracing.Enabled && cfgexTracing.Provider != tracing.JAEGER && cfgexTracing.Provider != tracing.TEMPO {
		return fmt.Errorf("error in configuration options for the external services tracing provider. Invalid provider type [%s]", cfgexTracing.Provider)
	}
```

But, said that, your changes looks good and this can be done in another PR.

nrfox:
> I think that could be solved by adding in kiali.go > validateconfig.go:

I can add that into this change.

nrfox:
@josunect well I made more changes than you asked for but I think it makes sense. Moved `validateConfig` from main --> `config.Validate(...)`. I had to move some of the jaeger/tempo constants to the `config` package as well because of that. I think this makes sense still since these get validated in `config` and they come from the `Config` struct as well.

josunect:
> @josunect well I made more changes than you asked for but I think it makes sense. Moved `validateConfig` from main --> `config.Validate(...)`. I had to move some of the jaeger/tempo constants to the `config` package as well because of that. I think this makes sense still since these get validated in `config` and they come from the `Config` struct as well.

Thanks for the refactor, @nrfox ! 
It looks good. 

Just one question: It is expected that, when starting the server, if the grpc client cannot be started (E.x. it has an incorrect port), the server is unable to start? The same if the provider type is incorrect?

nrfox:
> Just one question: It is expected that, when starting the server, if the grpc client cannot be started (E.x. it has an incorrect port), the server is unable to start? The same if the provider type is incorrect?

Only if tracing is enabled. If tracing is enabled and tracing is not configured properly but the server starts anyway then that is basically a broken state and could lead to runtime errors which would make it more confusing what the actual issue is.

josunect (approved):
Tested with Jaeger, Tempo, different config options, and works as expected. 

Code LGTM.

PR #6938, opened 2023-12-08
[OSSM-5306][OSSM-5307] Add building tests images workflow into pipeline

mkralik3:
**Describe the change**

Adding a workflow (`tests-images-creator.yml`) for building and pushing integration and cypress test images (both at once or a particular one). The workflow can be run from a different workflow or manually (e.g. for manually rebuilding a particular tag).
image

1. The workflow is called from `tests-images-update-latest.yml` job, which is watching folders related to tests. If something is pushed into those folders, the job detects in which folders the changes were done and rebuilds the cypress, integration, or both test images accordingly. So the `latest` tag will always contain the latest changes in test suites.

2 In addition, the workflow is called whenever a new tag is created (e.g. after release) by `tests-images-release.yml`. In that case, the new integration and cypress test images are created in Quay under the released tag version, e.g. `1.65.12`. Also, the images under the tag `1.65` are updated to point to the latest Z version (other words, the X.Y tag will contain always the latest X.Y.Z released version).

**Issue reference**
Ref:
https://issues.redhat.com/browse/OSSM-5306
https://issues.redhat.com/browse/OSSM-5307
https://issues.redhat.com/browse/OSSM-5356

**Examples**

- Manual trigger 
  - both: https://github.com/mkralik3/kiali/actions/runs/7183479032
  - integration tests only: https://github.com/mkralik3/kiali/actions/runs/7183520056
  - cypress tests only: https://github.com/mkralik3/kiali/actions/runs/7183525511
- Update master with update in
  - hack folder, which is related for both: https://github.com/mkralik3/kiali/actions/runs/7183511121
  - readme in integration, only int is rebuilt: https://github.com/mkralik3/kiali/actions/runs/7183607891
  - frontend/package.json, only cypress is rebuilt: https://github.com/mkralik3/kiali/actions/runs/7183629336
- After tag creation: https://github.com/mkralik3/kiali/actions/runs/7183109197 
  - I have created three tags, `v1.65.30`, `v1.65.31` and `v1.65.32`. So after that, the Quay repo contains `v1.65.30`, `v1.65.31`, `v1.65.32` and `v1.65` [same as v1.65.32] images

**Additional info**
I will backport this to the master and other necessary branches[which are maintained]. [This](https://github.com/kiali/kiali/commit/887d4100a4ef5a37664823e346ee84593009e9bb) needs to be backported as well (+ update cypress docker file), that is why firstly I want to add all these changes to 1.65 which is ready for it.

PR #6959, opened 2023-12-15
[OSSM-5564][hack] Fix install testing demo script to support OSSM 2.5

mkralik3:
**Describe the change**
Fix autoinjection for `bookinfo` and `sleep` demo projects when using with OSSM

Due to this change https://github.com/kiali/kiali/commit/997f5440113462d0d6fc41e1fbde66e72ad75b31 , the bookinfo script doesn't use `ENABLE_INJECTION` anymore, however, it is needed in `prepare_maistra` function.

Also, `ISTIO_NAMESPACE` env is needed in `prepare_mainstra` function so it was added to `install-sleep-demo.sh` script. (in version 1.65, the sleep installation was a part of `install-testing-demos.sh` where the ENV was available) 

**Issue reference**
https://issues.redhat.com/browse/OSSM-5564

### DIFF
--- a/hack/istio/functions.sh
+++ b/hack/istio/functions.sh
@@ -8,6 +8,7 @@
 # 1. Create a SMM (this is skipped for the cluster wide mode)
 # 2. Label the member namespace (only for the cluster wide mode)
 # 3. Annotate all of the namespace's Deployments with the sidecar injection annotation if enabled
+#! Be sure, that ISTIO_NAMESPACE and (ENABLE_INJECTION or AUTO_INJECTION) are properly set in the script where the function is called!
 prepare_maistra() {
   local ns="${1}"
   if $(is_cluster_wide);
@@ -30,7 +31,7 @@ EOM
     ${CLIENT_EXE} wait --for condition=Ready -n ${ISTIO_NAMESPACE} smmr/default --timeout 300s
   fi
 
-  if [ "${ENABLE_INJECTION}" == "true" ]; then
+  if [ "${ENABLE_INJECTION}" == "true" ] || [ "${AUTO_INJECTION}" == "true" ]; then
     for d in $(${CLIENT_EXE} get deployments -n ${ns} -o name)
     do
       echo "Enabling sidecar injection for deployment: ${d}"
--- a/hack/istio/install-sleep-demo.sh
+++ b/hack/istio/install-sleep-demo.sh
@@ -19,6 +19,8 @@ SCRIPT_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
 ISTIO_DIR=
 CLIENT_EXE="oc"
 DELETE_SLEEP="false"
+: ${ISTIO_NAMESPACE:=istio-system}
+: ${ENABLE_INJECTION:=true}
 
 # process command line args
 while [[ $# -gt 0 ]]; do
@@ -33,6 +35,10 @@ while [[ $# -gt 0 ]]; do
       ;;
     -id|--istio-dir)
       ISTIO_DIR="$2"
+      shift;shift
+      ;;
+    -in|--istio-namespace)
+      ISTIO_NAMESPACE="$2"
       shift;shift
       ;;
     -c|--client-exe)
--- a/hack/istio/install-testing-demos.sh
+++ b/hack/istio/install-testing-demos.sh
@@ -112,7 +112,7 @@ if [ "${DELETE_DEMOS}" != "true" ]; then
     echo "Deploying error rates demo ..."
     "${SCRIPT_DIR}/install-error-rates-demo.sh" -in ${ISTIO_NAMESPACE} -a ${ARCH}
     echo "Deploying sleep demo ..."
-    "${SCRIPT_DIR}/install-sleep-demo.sh" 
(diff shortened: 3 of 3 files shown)


jmazzitelli (approved):
This looks good to me, but I don't use these scripts so I would like someone else to also review this as well.

Also, this PR is targeting the v1.73 branch. We will need another PR to cherry pick this over to the master branch as well to keep everything in sync.

mkralik3:
@jmazzitelli @ScriptingShrimp 
Do we need to backport into master? If I understand correctly, all Maistra stuff were removed from a master branch by #6819, weren't they? This PR is also related only to Maistra. (the root cause is in prepare_maistra function)

hack/istio/functions.sh
```diff
@@ -8,6 +8,7 @@
 # 1. Create a SMM (this is skipped for the cluster wide mode)
 # 2. Label the member namespace (only for the cluster wide mode)
 # 3. Annotate all of the namespace's Deployments with the sidecar injection annotation if enabled
+#! Be sure, that ISTIO_NAMESPACE and (ENABLE_INJECTION or AUTO_INJECTION) are properly set in the script where the function is called!
```
jmazzitelli, line 11: This comment seems relevant even for non-Maistra environments (injections is still "a thing" in upstream Istio).

hack/istio/functions.sh
```diff
@@ -30,7 +31,7 @@ EOM
     ${CLIENT_EXE} wait --for condition=Ready -n ${ISTIO_NAMESPACE} smmr/default --timeout 300s
   fi
 
-  if [ "${ENABLE_INJECTION}" == "true" ]; then
+  if [ "${ENABLE_INJECTION}" == "true" ] || [ "${AUTO_INJECTION}" == "true" ]; then
```
jmazzitelli, line 34: This seems relevant even for non-Maistra environments (I don't see anything here that restricts this to only maistra environments)

hack/istio/install-sleep-demo.sh
```diff
@@ -19,6 +19,8 @@ SCRIPT_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
 ISTIO_DIR=
 CLIENT_EXE="oc"
 DELETE_SLEEP="false"
+: ${ISTIO_NAMESPACE:=istio-system}
+: ${ENABLE_INJECTION:=true}
```
jmazzitelli, line 23: This seems relevant even for non-Maistra environments. I don't see anything restricting this to maistra-only. Same with the changes in the rest of the script.

hack/istio/install-testing-demos.sh
```diff
@@ -179,7 +179,7 @@ EOF
     ${CLIENT_EXE} patch VirtualService bookinfo -n bookinfo --type json -p "[{\"op\": \"replace\", \"path\": \"/spec/hosts/0\", \"value\": \"${ISTIO_INGRESS_HOST}\"}]"
   fi
 
-  for namespace in bookinfo alpha beta
+  for namespace in bookinfo alpha beta gamma
```
jmazzitelli, line 182: This seems relevant even for non-Maistra environments (waiting for the workloads in gamma seems unrelated to maistra).

jmazzitelli (commented):
I don't see anything here that is specifically related to Maistra (at least nothing explicitly indicating maistra is required here).

If that is true, we would want this in master if only to keep these files in sync as best we can, and thus avoid potential git conflicts in the future if we ever need to make changes that need backporting.

But with that said, if ALL of this is simply to support Maistra, and this code isn't really used outside of, say, some "IF Maistra THEN" code block somewhere, then we can avoid putting it in master.

hack/istio/functions.sh
mkralik3: the function here is named `prepare_maistra` , so I expected that it is related only to `maistra`. (This function is not in the main branch anymore )

hack/istio/install-sleep-demo.sh
mkralik3: explicitly not, but implicitly yes, since those ENVs are used only in `prepare_maistra` function which `install-sleep-demo` script uses ,, but maybe it can be used for something else in the future, so, I can backport it into the main branch if needed

hack/istio/install-testing-demos.sh
mkralik3: yes, you are right, I forgot on this minor change. this is not only for maistra, so this can also be backported into main

hack/istio/functions.sh
mkralik3: the function here is named `prepare_maistra` , so I expected that it is related only to `maistra`. Also, this function is not in the main branch anymore.

jmazzitelli:
Merging this. @mkralik3 cherry pick what you think is best to the master branch, if anything.

ScriptingShrimp:
demo apps installed on OCP with OSSM2.5 OK :+1:

mkralik3:
@jmazzitelli  I think only [this line](https://github.com/mkralik3/kiali/blob/v1.73/hack/istio/install-testing-demos.sh#L182) is worth to backport. I will include it in backport for this https://github.com/kiali/kiali/pull/6938

Issue #6844, opened 2023-11-14
Tempo - Review performance


josunect:
It looks like, when doing some queries that returns a big amount of results, the pod frontend-query dies - Kiali is returning a 503 error. 

There are not many of this calls - maybe 2 or 3.

jshaughn:
Any comparison with Jaeger?

josunect:
I made this table so we can compare different calls. This is the bookinfo application. 

| -  | Jaeger | Tempo |
| ------------- | ------------- | ------------- |
| Traces 1m | 104ms  | 370 ms |
| Traces 10m  | 195ms | 1,26s |
| Traces 7d | 155ms | 712 ms |
| Get Trace | 10 ms | 16ms |
| Traces (Graph) | 47 ms | 458ms |

- Traces call: http://localhost:3000/api/namespaces/bookinfo/workloads/reviews-v1/traces?startMicros=1700737749946000&tags=&limit=100
- Trace: http://localhost:3000/api/traces/76cea5a020203ffebb1ef23e04f26282

jshaughn:
Wow, Tempo is significantly slower.  Thanks for the table comparison!

josunect:
> Wow, Tempo is significantly slower. Thanks for the table comparison!

There is also true, that we are also processing the traces and transform them into Jaeger format to be used internally. So, in the future maybe we can change internally to use the OTel format. I think this will require an effort.

josunect:
I was reviewing the performance, and I see (From the graph view) there are many concurrent calls to the Tracing API with just one click on the graph: 

```
2023-12-13T18:13:50Z INF [2023-12-13 18:12:46 - 2023-12-13 18:13:50] http://localhost:16686/jaeger/api/traces?end=1702491230000000&limit=15&service=reviews.bookinfo&start=1702491166000000&tags=%7B%22cluster%22%3A%22Kubernetes%22%7D
2023-12-13T18:13:50Z INF [2023-12-13 18:12:46 - 2023-12-13 18:13:50] http://localhost:16686/jaeger/api/traces?end=1702491230000000&limit=15&service=reviews.bookinfo&start=1702491166000000
2023-12-13T18:13:50Z INF [2023-12-13 18:12:58 - 2023-12-13 18:13:05] http://localhost:16686/jaeger/api/traces?end=1702491185000000&limit=1&service=reviews.bookinfo&start=1702491178000000&tags=%7B%22cluster%22%3A%22Kubernetes%22%7D
[...]

PR #6960, opened 2023-12-15
Use ignorecluster global for tracing

josunect:
Ref. #6844

https://github.com/kiali/kiali/issues/6844#issuecomment-1854504377

In order to get a sample of the traces from different times and a limited number of results, in some calls (Like the node graph for traces), splits the search query in 10, this is, will perform up to 10 concurrent requests to the tracing platform. 

When there are no cluster tags, each query is repeated, which in a busy environment can be translated in a worst performance. This PR tries to get this information when the tracing client is created and avoid the duplication of calls to the backend.

### DIFF
--- a/tracing/client.go
+++ b/tracing/client.go
@@ -99,10 +99,6 @@ func NewClient(token string) (*Client, error) {
 			return nil, errParse
 		}
 
-		if cfg.ExternalServices.Tracing.Provider == JAEGER {
-			httpTracingClient = jaeger.JaegerHTTPClient{}
-		}
-
 		port := u.Port()
 		if port == "" {
 			p, _ := net.LookupPort("tcp", u.Scheme)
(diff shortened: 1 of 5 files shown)


hhovsepy:
@josunect the solution looks good, should it be applied for Jaeger GRPC client as well?

josunect:
Tested with MC. It uses the cluster tag and no duplicated queries in the backend. 

Testing procedures: 

- `istio/multicluster/install-primary-remote.sh -c kubectl`
- `istio/multicluster/deploy-kiali.sh -c kubectl -kas anonymous`
- Update the istio cm: 

```
data:
  mesh: |-
    accessLogFile: /dev/stdout
    defaultConfig:
      discoveryAddress: istiod.istio-system.svc:15012
      meshId: mesh-hack
      proxyMetadata: {}
      tracing:
        sampling: 100
    defaultProviders:
      metrics:
      - prometheus
      tracing:
      - localtrace
    enableAutoMtls: true
    enablePrometheusMerge: true
    extensionProviders:
    - name: "localtrace"
      zipkin:
        service: zipkin.istio-system.svc.cluster.local
        port: 9411
```

And create a Telemetry to add the cluster tag: 

```
kubectl apply -f - <<EOF
kind: Telemetry
metadata:
  name: mesh-default
  namespace: istio-system
spec:
  tracing:
  -  customTags:
      cluster:
        environment:
          name: "ISTIO_META_CLUSTER_ID"
EOF
```

josunect:
> @josunect the solution looks good, should it be applied for Jaeger GRPC client as well?

Good catch, @hhovsepy ! I will add it also there, thanks.

josunect:
> > @josunect the solution looks good, should it be applied for Jaeger GRPC client as well?
> 
> Good catch, @hhovsepy ! I will add it also there, thanks.

@hhovsepy, I've included it in the Jaeger GRPC client. Thanks!

hhovsepy (approved):
@josunect looks good , tested locally.

Issue #6966, opened 2023-12-18
use the Prometheus `/-/healthy` endpoint for the default value for health_check_url


jmazzitelli:
by default, use the [`/-/healthy` endpoint that Prometheus has](https://prometheus.io/docs/prometheus/latest/management_api/), rather than the main prometheus root url.

This is because sometimes Prometheus doesn't have the UI installed and this causes the root URL to tell the client to redirect to an invalid `/graph` endpoint, causing a error code to be returned, thus Kiali UI will report Prometheus being unreachable which isn't true.

For example, in one setup we've seen in testing, this happens.

If I go into the prometheus pod itself, and try to submit a request to check its health (using curl), I get this:
```
curl -k -u internal:DiQkSCH6Y/DManOj8HI7S..[chomp]..qQ6AsJxTEI3 https://prometheus.istio-system.svc:9090/
<a href="/graph">Found</a>.
```
Which looks fine, but that is actually returning a 302 - it is redirecting the client to `/graph` and if I go to the `/graph` path I see this:
```
curl -k -q -u internal:DiQkSCH6Y/DManOj8HI7S..[chomp]..qQ6AsJxTEI3 https://prometheus.istio-system.svc:9090/graph
Error opening React index.html: open web/ui/static/react/index.html: no such file or directory
```
and that returns a error code of 500, which results in the Kiali UI to report Prometheus as "unreachable". But that is not true. Kiali UI can show the graph and metrics fine. This is because the Prometheus backend is working fine, it is just the Prometheus UI isn't installed or configured properly. But Kiali doesn't care about that - it just cares that Prometheus can respond to our queries (which it can in this particular case).

PR #6969, opened 2023-12-18
for health check, use prom healthy endpoint by default

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6966

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -227,6 +227,12 @@ func (iss *IstioStatusService) getAddonComponentStatus() kubernetes.IstioCompone
 
 	staChan := make(chan kubernetes.IstioComponentStatus, 4)
 	extServices := config.Get().ExternalServices
+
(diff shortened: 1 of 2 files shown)


jmazzitelli:
to test, just deploy everything normally and get into a good state. Then scale down the prometheus deployment to 0 pods:

```
kubectl scale --replicas=0 deployment -n istio-system prometheus
```

and then look at the Kiali UI and wait for it to refresh. You should get the "unreachable" masthead message (along with some other bad errors because prometheus is down, but ignore those, that's not part of this issue).

Then scale up the prom deployment back to 1:

```
kubectl scale --replicas=1 deployment -n istio-system prometheus
```

and the Kiali UI should go back to normal - you should no longer see the "unreachable" masthead message.

jshaughn (approved):
Ran the test instructions successfully. LG

PR #6976, opened 2023-12-19
Retrieve log level from correct source

### LINKED ISSUE #6964: debug info shows incorrect log level
Set the `spec.deployment.logger.log_level` to something other than "trace" in the Kiali CR. Then go to the Debug Info in the Kiali UI and notice it has the wrong log level - it always says trace:
loglevel

ferhoyos:
**Describe the change**

Correct log level information can be found in zerolog.GlobalLevel(), not from log.Logger.GetLevel()

** Issue reference **

Fixes #6964

### DIFF
--- a/log/log.go
+++ b/log/log.go
@@ -121,7 +121,7 @@ func Fatalf(format string, args ...interface{}) {
 // Return log level
 // Used to get debug info
 func GetLogLevel() string {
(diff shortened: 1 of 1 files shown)


jmazzitelli (approved):
lgtm
image

Issue #6918, opened 2023-12-05
K8s GW API - Support of ReferenceGrant


hhovsepy:
K8s Gateway API routing supports cross namespace referencing of objects by using a ReferenceGrant: https://gateway-api.sigs.k8s.io/api-types/referencegrant
Kiali needs to support ReferenceGrant objects in:

- [x] Istio objects list and filters.
- [x] Object references to HTTPRoute and Service
- [x] Validations of 'from:' and 'to:' fields.
- [x] Include in wizards.

PR #6983, opened 2023-12-20
Consider ReferenceGrant in HTTPRoute Service validation KIA1402

hhovsepy:

In K8s HTTPRoute Backend reference (https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.BackendObjectReference), when the Namespace where Service is referenced differs from the local namespace, where the HTTPRoute is created, then a ReferenceGrant object is required to reference the Service namespace and local namespace.

Now: KIA1402 validation is modified to consider ReferenceGrant objects when those 2 namespaces are different. 

How to test it:
Apply K8s Gateway API CRD v1.0.0
Create HTTPRoute and ReferenceGrant.
```
kind: HTTPRoute
metadata:
  name: inbookinfo
  namespace: bookinfo
spec:
  rules:
  - matches:
    - path:
        type: Exact
        value: "/bar"
    backendRefs:
      - name: kiali
        namespace: istio-system
        port: 9080
---
apiVersion: gateway.networking.k8s.io/v1beta1
kind: ReferenceGrant
metadata:
  name: toistio
  namespace: istio-system
spec:
  from:
  - group: gateway.networking.k8s.io
    kind: HTTPRoute
    namespace: bookinfo
  to:
  - group: ""
    kind: Service
```
Verify that HTTPRoute is referring to a Service from remote namespace, but no error as ReferenceGrant is referencing those namespaces.
Screenshot from 2023-12-20 19-54-30
Screenshot from 2023-12-20 19-52-32

When ReferenceGrant does not exist:
Screenshot from 2023-12-20 19-54-51

### DIFF
--- a/business/checkers/k8shttproute_checker.go
+++ b/business/checkers/k8shttproute_checker.go
@@ -2,6 +2,7 @@ package checkers
 
 import (
 	k8s_networking_v1 "sigs.k8s.io/gateway-api/apis/v1"
+	k8s_networking_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/kiali/kiali/business/checkers/k8shttproutes"
 	"github.com/kiali/kiali/kubernetes"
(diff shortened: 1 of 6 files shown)


business/checkers/k8shttproute_checker.go
```diff
@@ -11,11 +12,12 @@ import (
 const K8sHTTPRouteCheckerType = "k8shttproute"
 
 type K8sHTTPRouteChecker struct {
-	K8sHTTPRoutes    []*k8s_networking_v1.HTTPRoute
-	K8sGateways      []*k8s_networking_v1.Gateway
-	Namespaces       models.Namespaces
-	RegistryServices []*kubernetes.RegistryService
-	Cluster          string
+	K8sHTTPRoutes      []*k8s_networking_v1.HTTPRoute
```
josunect, line 15: Just a minor thing, it would be possible to alphabetize?

josunect (commented):
Verified: 
image

With the reference grant: 
image

Just a minor comment.

business/checkers/k8shttproute_checker.go
hhovsepy: ah right

josunect (approved):
LGTM 

Thanks, @hhovsepy !

Issue #6962, opened 2023-12-15
Kiali render hostnames as individual service instead of serviceentry as whole


im-Amitto:
### Describe the bug
Loom Video :-> https://www.loom.com/share/2c1906c2ffe042dcbf6d96ca9e4f7ef4
A serviceentry is supposed to be visualised as single icon in graph does not matter how many number of hostnames it has in it.
```yaml
apiVersion: networking.istio.io/v1alpha3
kind: ServiceEntry
metadata:
  name: api.headout.com
spec:
  hosts:
  - api.example.com
  - api.example2.com
  ports:
  - number: 443
    name: https
    protocol: HTTPS
  resolution: DNS
  location: MESH_EXTERNAL
```

It should be shown as a single icon but kiali is showing 1 service icon per hostname. It does not happen always. It does show the service entry as a single icon sometime but goes back to showing one icon per hostname on refresh.

### Expected Behavior
Single icon for a single service entry

### What are the steps to reproduce this bug?

1. Create a serviceentry with two hostname
2. Visualise the traffic

### Environment

- **Kiali version:** v1.76.1 (0f96891475ab41b10821ae8e1850572fbcd1505f)
- **Istio version:** 1.18.2
- **Kubernetes version:**  v1.28.4-eks-8cb36c9
- **Prometheus** 2.48.0

hhovsepy:
I confirm seeing SE per host in the graph, this is because Istio is generating a Service per SE host in Service Registry.

im-Amitto:
@hhovsepy but it does not happen everytime, sometime i get the visualisation that i want where it's shown as a single entity.
Here `mb-domain` contain more than 400 hostname.
<img width="428" alt="Screenshot 2023-12-18 at 9 40 49 AM" src="https://github.com/kiali/kiali/assets/29653282/66ae4b8e-559c-4bec-b13d-23ca517f1f62">

hhovsepy:
Seems like Kiali would need a new functionality in a Graph. A checkbox to merge all URLs of the ServiceEntry into one or show it separately.

jshaughn:
> Seems like Kiali would need a new functionality in a Graph. A checkbox to merge all URLs of the ServiceEntry into one or show it separately.

@hhovsepy After thinking about this I think this should not be an option, I think the requested behavior is the intended behavior, and only a single ServiceEntry node should exist for all of the hosts.  And so, this does seem like a bug, and actually a regression, as I believe the code intends to consolidate into a single node.

[...]

PR #6984, opened 2023-12-20
Fix issue in ServiceEntry conversion

jshaughn:
Fixes https://github.com/kiali/kiali/issues/6962

- I'm not sure if telemetry changed recently but when making requests
  of a service entry the telemetry may not include any cluster or namespace
  information, just the destination_service_name (i.e. the requested host)
- The subsequent "unknown" values for cluster and namespace can not be used
  to help fetch the ServiceEntry config for matching. Depending on
  random node ordering, sometimes we would "get lucky" and fetch the needed
  information, but the whole approach was inherently flawed.
- mesh_external ServiceEntry definitions are really used by the source node,
  to properly route out of the mesh, and so the SE definition needs to be
  fetched using the source node's cluster+namespace information. This PR makes
  that change. It also better handles the mesh_internal scenario, which
  *does* use the cluster+namespace provided.
- The PR also adds a useful optimization that can speed up this appender;
  for namespaces that only have injected service nodes we will no longer
  bother with loading SE definitions or doing the matching, because injected
  service nodes are not going to be converted to a ServiceEntry.
- The tests are updated to not use an "unknown" node as the SE requestor,
  that is not a valid scenario.  Also, fixed an issue in the mocking of one
  test.


jshaughn:
@hhovsepy I think you said you had recreated the problem locally.  But if you are looking for a way to test this you can install the electronic-shop demo just up to the point before the "Egress Traffic" section.  The service entries should never flip-flop to service nodes on repeated refreshes.  For me it only took a few refreshes to cause the problem.

hhovsepy (approved):
lgtm, I can see the issue is fixed:
Before:
Screenshot from 2023-12-21 21-21-28
Screenshot from 2023-12-21 21-21-05

After:
Screenshot from 2023-12-21 21-23-15

PR #6986, opened 2023-12-21
Fix Workload Logs scrollbar issue

### LINKED ISSUE #6982: Scrollbar in Workloads Logs view
### Describe the bug

There is no horizontal scrollbar in Firefox: 

image

Which makes impossible to read large lines. 

The behavior in Chrome is different. There are vertical/horizontal scrollbar, but if we zoom in, a double scrollbar appears, which makes it confusing (It looks like the scrollbar is not working because the second one is hidden): 

image

### Expected Behavior

- I think there should be horizontal scrollbar in Firefox, and no double scrollbars in Chrome. 

1. Go to a workload, and click in the log tabs.
2. Zoom in with Chrome

### Environment

- **Kiali version:** Running Kiali from master
- **Istio version:** Istio 1.20
- **Kubernetes impl:** 
- **Kubernetes version:** 

ferhoyos:
**Describe the change**

Fix of two scrollbar issues in workload logs:
- Horizontal scrollbar does not appear in Firefox:
 
  React Virtualized List component has different overflow style depending on browser:
  - Chrome: `overflow: auto`
  - Firefox: `overflow: hidden auto`

  Setting the oveflow style to `auto !important` makes the trick.

- Double scrollbar in Chrome when user zooms in:

  Extra `overflow` style present in the List component. 

**Issue reference**

Fixes #6982

### DIFF
--- a/frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
+++ b/frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
@@ -158,7 +158,6 @@ const logsDiv = kialiStyle({
 const logsDisplay = kialiStyle({
   fontFamily: 'monospace',
   margin: 0,
-  overflow: 'auto',
   padding: 0,
   resize: 'none',
   whiteSpace: 'pre',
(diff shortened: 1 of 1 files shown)


josunect (approved):
Fixed! 

LGTM, thank you!

PR #6988, opened 2023-12-22
Update trace logic for Tempo lo limit the number of results

josunect:
For some reason, it looks like Tempo returns more than the limit results. 
It limits the number of results and splits the query when necessary. 
It looks like with increasing the number of results supports the load. 

With this change, with Tempo: 
- No more results than the limit are returned. 
- Traces are distributed between the graph. 

Ref. #6844


hhovsepy (approved):
The solution looks good, debugged locally. However was not able to reproduce the initial issue on my local env. Anyway the limit will be applied.

PR #6991, opened 2024-01-02
Remove Jaeger Exporter

### LINKED ISSUE #6908: Deprecate and remove Jaeger exporter
Kiali exports traces to help debug/monitor Kiali itself: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.server.observability.tracing.

Kiali currently supports both the jaeger and the otel exporter. Since jaeger itself can ingest otel traces and exporting otel grants more flexibility for where those traces can be sent, Kiali should just remove the jaeger exporter and use the otel one.

Note that this will probably also involve removing: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.server.observability.tracing.collector_url in favor of different config opts where you provide the `hostname:port` and/or `path` since that is the format that the otel sdk expects. Kiali should by default send traces to the otel endpoint rather than the jaeger one: `http://jaeger-collector.istio-system:4318/v1/traces`.

josunect:
**Describe the change**

Remove Jaeger exporter.
Example to test: 
```
server:
  observability: 
   tracing: 
    collector_url: "localhost:4318"
    enabled: true
    protocol: "http"
    tls_enabled: false
    tls_config: 
     skip_verify: true
     ca_name: "/tmp/tls.crt"
```

I've included some changes in the observability section from the configuration. As there is just one collector type, the option `collector_type` was removed. Because of this, the `otel` configuration section was renamed to `tls_config`, and the `protocol` and `tls_enabled` was moved to the `tracing` section. 
I have in mind to minimize this CR changes, but, I think this reorganization makes more sense and probably the new settings (With the collector_type) were not used a lot yet. 

**Issue reference**

Fixes #6908

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -76,6 +76,14 @@ const (
 	TempoProvider  TracingProvider = "tempo"
 )
 
+// TracingCollectorType is the type of collector that Kiali will export traces to.
+// These are traces that kiali generates for itself.
+type TracingCollectorType string
+
+const (
+	OTELCollectorType TracingCollectorType = "otel"
+)
+
 var validPathRegEx = regexp.MustCompile(`^\/[a-zA-Z0-9\-\._~!\$&\'()\*\+\,;=:@%/]*$`)
 
 // FeatureName is the enum type used for named features that can be disabled via KialiFeatureFlags.DisabledFeatures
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -119,21 +109,20 @@ type Metrics struct {
...
+type TLSConfig struct {
 	CAName     string `yaml:"ca_name,omitempty"`
-	Protocol   string `yaml:"protocol,omitempty"` // http or https or grpc
 	SkipVerify bool   `yaml:"skip_verify,omitempty"`
-	TLSEnabled bool   `yaml:"tls_enabled,omitempty"`
 }
 
 // Tracing provides tracing configuration for the Kiali server.
 type Tracing struct {
-	CollectorType TracingCollectorType `yaml:"collector_type,omitempty"` // Possible values "otel" or "jaeger"
```
nrfox, line 131: I'm not opposed to removing this since everything in the foreseeable future will support otel but iirc there was a discussion (that I can no longer find) around keeping this in case in the future we again add more than one type. iirc @jmazzitelli wanted to keep the collector type for this reason?

config/config.go
```diff
@@ -119,21 +109,20 @@ type Metrics struct {
...
-	CollectorURL  string               `yaml:"collector_url,omitempty"`  // Endpoint for Kiali server traces
-	Enabled       bool                 `yaml:"enabled,omitempty"`
-	Otel          OtelCollector        `yaml:"otel,omitempty"`
+	CollectorURL string `yaml:"collector_url,omitempty"` // Endpoint for Kiali server traces
+	Enabled      bool   `yaml:"enabled,omitempty"`
+	Protocol     string `yaml:"protocol,omitempty"` // http or https or grpc
 	// Sampling rate for Kiali server traces. >= 1.0 always samples and <= 0 never samples.
-	SamplingRate float64 `yaml:"sampling_rate,omitempty"`
+	SamplingRate float64   `yaml:"sampling_rate,omitempty"`
+	TLSEnabled   bool      `yaml:"tls_enabled,omitempty"`
```
nrfox, line 124: To be consistent with other config options in Kiali, I think this should go inside `TLSConfig` as `TLSConfig.Enabled`. The pattern seems to be:

If it's a singular flag, it is named `<Flag>Enabled` e.g. `GzipEnabled` https://github.com/kiali/kiali/blob/c8210e66953dd09824b61f006403358fdead3f42/config/config.go#L150

If it's a struct with multiple options, `Enabled` goes inside the struct e.g. `Tracing.Enabled`: https://github.com/kiali/kiali/blob/c8210e66953dd09824b61f006403358fdead3f42/config/config.go#L133

A different way to do this is to omit the `Enabled` field altogether and instead have `Tracing.TLSConfig` be a pointer and if the pointer is non-nil then it is enabled implicitly. But Kiali doesn't follow this pattern for the rest of the config so I'd just stick with `TLSConfig.Enabled`.

config/config.go
josunect: Right, I remember, now that you mention it... I think it might be in slack. Maybe we can wait to see what @jmazzitelli thinks about it?

josunect:
@nrfox based on the discussion on slack, I've updated the config as it was before in order to: 

- minimize the changes in the CR. 
- The possibility to add other providers in the future

nrfox (approved):

PR #6994, opened 2024-01-02
Added Reference from K8sHTTPRoute to K8sReferenceGrant

hhovsepy:

Reference from K8sHTTPRoute to K8sReferenceGrant.
Also the reference from Service to K8sHTTPRoute considers the existence of K8sReferenceGrant in case when they are in different namespaces.

Screenshot from 2024-01-03 09-07-22
Screenshot from 2024-01-03 09-07-08
Screenshot from 2024-01-04 10-55-00


josunect (commented):
Thanks,  @hhovsepy! It works as expected:

image

I wonder if it could make sense to do it in the opposite way, if from the Reference Grant the K8sHTTPRoute should be added as a reference? 

Code looks good, tests added, just adding a couple of minor comments.

business/references/k8shttproute_references.go
```diff
@@ -2,14 +2,16 @@ package references
...
+	k8s_networking_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/models"
 )
 
 type K8sHTTPRouteReferences struct {
-	Namespaces    models.Namespaces
-	K8sHTTPRoutes []*k8s_networking_v1.HTTPRoute
+	Namespaces         models.Namespaces
```
josunect, line 12: It is possible to alphabetize here?

models/service.go
```diff
@@ -72,16 +73,17 @@ type ServiceDefinitionList struct {
...
-	Cluster          string                                `json:"cluster"`
-	IstioSidecar     bool                                  `json:"istioSidecar"`
-	Endpoints        Endpoints                             `json:"endpoints"`
-	VirtualServices  []*networking_v1beta1.VirtualService  `json:"virtualServices"`
-	DestinationRules []*networking_v1beta1.DestinationRule `json:"destinationRules"`
-	K8sHTTPRoutes    []*k8s_networking_v1.HTTPRoute        `json:"k8sHTTPRoutes"`
-	ServiceEntries   []*networking_v1beta1.ServiceEntry    `json:"serviceEntries"`
-	IstioPermissions ResourcePermissions                   `json:"istioPermissions"`
-	Workloads        WorkloadOverviews                     `json:"workloads"`
+	Service            Service                                  `json:"service"`
```
josunect, line 76: It is possible to alphabetize here?

hhovsepy:
Thank you @josunect , I was thinking to add the back references, but in case of ReferenceGrant, it will show all HTTPRoutes from the particular namespace, so could be a bunch of references.
Then that can be a link to a Istio Configs page with applied filter "K8sHTTPRoute", but this will require additional feature for References only. So left that as it is now, user can search objects in a namespace manually.
> 
> I wonder if it could make sense to do it in the opposite way, if from the Reference Grant the K8sHTTPRoute should be added as a reference?
>

josunect:
> Thank you @josunect , I was thinking to add the back references, but in case of ReferenceGrant, it will show all HTTPRoutes from the particular namespace, so could be a bunch of references. Then that can be a link to a Istio Configs page with applied filter "K8sHTTPRoute", but this will require additional feature for References only. So left that as it is now, user can search objects in a namespace manually.
> 
> > I wonder if it could make sense to do it in the opposite way, if from the Reference Grant the K8sHTTPRoute should be added as a reference?

Yeah, that makes sense, thanks for clarifying!

josunect (approved):
LGTM

PR #6997, opened 2024-01-03
[1.73][OSSM-5599] Add support for OSSM for wait-to-namespace script

mkralik3:
**Describe the change**

Test case `The idle status of a service is reported in the list of services` is failing with OSSM 2.5 because `wait-for-namespace.sh` script cannot find `istio` namespace

**Issue reference**

ref: https://issues.redhat.com/browse/OSSM-5599

### DIFF
--- a/hack/istio/wait-for-namespace.sh
+++ b/hack/istio/wait-for-namespace.sh
@@ -35,11 +35,14 @@ HELPMSG
   esac
 done
 
-for NAMESPACE in ${NAMESPACES[@]}; do
-  oc patch kiali kiali -n kiali-operator --type=json '-p=[{"op": "add", "path": "/spec/deployment/accessible_namespaces/0", "value":"'$NAMESPACE'"}]'
-done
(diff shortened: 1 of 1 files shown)


hack/istio/wait-for-namespace.sh
```diff
@@ -35,11 +35,14 @@ HELPMSG
   esac
 done
 
-for NAMESPACE in ${NAMESPACES[@]}; do
-  oc patch kiali kiali -n kiali-operator --type=json '-p=[{"op": "add", "path": "/spec/deployment/accessible_namespaces/0", "value":"'$NAMESPACE'"}]'
-done
+IS_MAISTRA=$([ "$(oc get crd | grep servicemesh | wc -l)" -gt "0" ] && echo "true" || echo "false")
```
ScriptingShrimp, line 38: I think its safe to add MAISTRA related things into 1.73 since it will be last version for OSSM 2.x

ScriptingShrimp (approved):
Passing on OCP, additionally I discovered one [issue](https://github.com/kiali/kiali/issues/7004) related to this scripts file but no blocker for this PR.

LGTM

PR #6999, opened 2024-01-03
Output failure info in CI to a file instead of stdout/err

nrfox:
** Describe the change **

The output of the `Get debug info when integration tests fail` step in CI is really large and usually when there's a failure you just want to see the test output which is different than this info. This change will write that info to a file that can be downloaded rather than to stdout/err.


leandroberetta (approved):
Looks good.

PR #7003, opened 2024-01-04
Node and group customizations moved to Kiali style components

leandroberetta:
Resolves https://github.com/kiali/openshift-servicemesh-plugin/issues/224.
Resolves https://github.com/kiali/openshift-servicemesh-plugin/issues/225.

This PR removes the node.tsx and groups related .tsx files which were a copy from PF's DefaultNode and Default groups.

There is a small feature that can't be moved and an issue was created on PF repo to contribute it there as we have been doing it in the past with other features: https://github.com/patternfly/react-topology/issues/127

Also, this PR introduces a new way of focusing a node or a group when going back from the Node graph, it just simply selects the node or group, the selected node or group will remain until the user does another selection. I think this is even better than the original animation as a slow user can miss that animation.


frontend/src/pages/GraphPF/NodeDecorator.tsx
```diff
@@ -0,0 +1,38 @@
...
+  quadrant: TopologyQuadrant;
+  tooltip?: string;
+}
+
+const NodeDecoratorInner: React.FC<Props> = ({ element, quadrant, icon, tooltip }) => {
+  const { x, y } = getDefaultShapeDecoratorCenter(quadrant, element);
+  const decoratorRef = React.useRef<SVGAElement | null>(null);
+
+  return (
+    <Tooltip triggerRef={decoratorRef} content={!!tooltip ? tooltip : icon.text}>
```
ferhoyos, line 25: Content value can be simplified to: `content={tooltip ?? icon.text}`

leandroberetta: Thanks @ferhoyos !.

jshaughn (commented):
The work you've done is good, but I think we may be sacrificing too much to make this change:

-  In my opinion pre-select is not a reasonable replacement for the focus feature.  I can not locate the selected node, especially in a larger graph.  To be honest, PFT does not make it very obvious which node is selected.  I sometimes lose track of my selected node just when looking away from it.  
- The "raise" feature is very important because PFT loves to overlap labels.  If there were a way to keep labels inside of boxes it would help a lot, but I couldn't get Jeff's suggestion to work.  FWIW,  I think it should be PFT built-in to raise the hovered information.
- The scaled-label-on-hover feature has also been lost.  This is very important when trying to look at a larger graph.
- As an aside, I think there are problems with the trace overlay , although I'm not sure it's due to the PR.

I think we need to wait for PFT to support some of our customizations before we make this change.

leandroberetta:
Thanks for the review @jshaughn.

> The work you've done is good, but I think we may be sacrificing too much to make this change:
> 
> * In my opinion pre-select is not a reasonable replacement for the focus feature.  I can not locate the selected node, especially in a larger graph.  To be honest, PFT does not make it very obvious which node is selected.  I sometimes lose track of my selected node just when looking away from it.

Well, select a node or group is a kind of focus, although not animated. One think that I like for the select vs the focus is that the selected node and group remains selected. The focus animation goes away and the user might miss it. I do understand the issue with larger graphs, an interesting idea should be to zoom in the graph to the selected node/group. 

> * The "raise" feature is very important because PFT loves to overlap labels.  If there were a way to keep labels inside of boxes it would help a lot, but I couldn't get Jeff's suggestion to work.  FWIW,  I think it should be PFT built-in to raise the hovered information.

I created the following issue to address the overlapping labels: https://github.com/patternfly/react-topology/issues/127

> * The scaled-label-on-hover feature has also been lost.  This is very important when trying to look at a larger graph.

I missed this one, but we can create another issue in PF to address this.


> * As an aside, I think there are problems with the trace overlay , although I'm not sure it's due to the PR.
> 

I see, don't think it's related but I will fix it.

> I think we need to wait for PFT to support some of our customizations before we make this change.

Well, I do understand your concerns about the missing features in the new graph. My rational in this changes is that I prefer to go with what PF can do today, and then add these features through contributions in Patternfly topology.
Having said that, if you prefer, I can keep these customizations in the tsx files borrowed from PF, mostly for the groups.

leandroberetta:
> > One think that I like for the select vs the focus is that the selected node and group remains selected. The focus animation goes away and the user might miss it
> 
> Note that with Cytoscape we have focus animation and the node remains selected. The zoom idea may be an alternative, if we could center on the focus node.
> 
> If we merge without solving the raise/label-overlap issue, as well as the scaled-label-on-hover issue, I feel like the PFT solution will be quite degraded from cy for non-trivial graphs. This would only affect OSSMC and Backstage users, as we will not change from Cy for Kiali until PFT is as good or better. IF we can get Jeff to better explain his suggestion for getting labels inside the box, that would improve the situation A LOT.

When using the rectangular boxing (not yet released...) we can achieve the labels inside the box, take a look at: https://github.com/patternfly/react-topology/issues/126. 

This only fixes the overlapping labels, but not the scaling label, that still should be done. 

I will revert the commits that address the group customizations, let's keep the node changes, I think the group customizations will take longer due to all the requirements.

jshaughn:
I've been thinking about how to proceed here.  I do want to merge in the work you have done, because it is good to eliminate our private, modified copy of the internal PFT code.  That is a real danger with respect to being able to upgrade PFT.  But, we do lose important customizations that you were not able to port.  I think what we can do is merge you changes while also aggressively creating issues, and pushing for feature additions in PFT.  This approach may be OK because PFT is not the default graph impl for Kiali, so Kiali users will not really be affected.  OSSMC is pinned to Kiali version 1.73, and will be until OSSM 3 is release, giving us several months before changes to master will be pushed on OSSMC users.  In the best-case scenario this will be enough time for us to improve the PFT impl to the point that it is as good as, or better, than the Cyto impl.  If that happens we will be able to remove Cyto in the OSSM 3 timeframe.  Or, minimally, make PFT the default and deprecate Cyto.

jshaughn (dismissed):
OK, approving so we can avoid the copied+customized internal PFT code.  I've updated the Epic with follow-up tasks for the customizations we've lost.

I'm assuming that the PR is now targeted just to Nodes and we will deal with groups in another PR (may want to update the PR title).

jshaughn (approved):
OK, this looks more complete.  Forward!

Issue #7006, opened 2024-01-05
Potential runtime error in kube_cache.GetK8sGateways


hhovsepy:
Happens on Kiali master.
Start the Kiali when K8s GW CRDs are not created.
Apply K8s GW CRDs.
Immediately refresh the Kiali page.
Runtime panic nil pointer error in logs, on any of those methods `kube_cache.GetK8sGateways`.

This is because the `userClient.IsGatewayAPI()` return 'true' when the CRDs exist, but in fact the `k8sgatewayLister` was not created while starting the Kiali first time.
https://github.com/kiali/kiali/blob/master/kubernetes/cache/kube_cache.go#L397
![Screenshot from 2024-01-05 09-52-45](https://github.com/kiali/kiali/assets/604313/a7621bb9-cd07-4043-9829-79cf8e975632)

And any call to `k8sgatewayLister` causes nil error.
https://github.com/kiali/kiali/blob/master/kubernetes/cache/kube_cache.go#L1504
![Screenshot from 2024-01-05 09-55-57](https://github.com/kiali/kiali/assets/604313/5187cc66-1637-459d-a61f-699e4934c804)

jshaughn:
@hhovsepy , I think this issue should have been closed by https://github.com/kiali/kiali/pull/7008, no?  I'll close, please re-open if I am wrong.

PR #6932, opened 2023-12-06
Fix race condition in kube cache by locking sooner

nrfox:
Fixes #6641

PR #7008, opened 2024-01-05
Check K8sGatewayLister init.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/7006

Make sure that `k8sgatewayLister` exists before calling it, to skip 'nil' errors.
This can happen when K8s GW CRDs are creatd, but the cache it not reloaded.

### DIFF
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -1464,7 +1464,7 @@ func (c *kubeCache) GetTelemetries(namespace, labelSelector string) ([]*v1alpha1
 	return retTelemetries, nil
 }
 
-func (c *kubeCache) IsK8sGatewayListerInit(namespace string) bool {
+func (c *kubeCache) isK8sGatewayListerInit(namespace string) bool {
 	// K8s GW has several cache listers, those can be namespace or cluster based
 	// if one of them is not initialized, then all others are not init as well
 	// this can happen when CRDs are created after Kiali start
@@ -1483,7 +1483,7 @@ func (c *kubeCache) GetK8sGateway(namespace, name string) (*gatewayapi_v1.Gatewa
 	// but it won't prevent other routines from reading from the lister.
 	defer c.cacheLock.RUnlock()
 	c.cacheLock.RLock()
-	if !c.IsK8sGatewayListerInit(namespace) {
+	if !c.isK8sGatewayListerInit(namespace) {
 		return nil, nil
 	}
 	g, err := c.getCacheLister(namespace).k8sgatewayLister.Gateways(namespace).Get(name)
@@ -1510,7 +1510,7 @@ func (c *kubeCache) GetK8sGateways(namespace, labelSelector string) ([]*gatewaya
 	// but it won't prevent other routines from reading from the lister.
 	defer c.cacheLock.RUnlock()
 	c.cacheLock.RLock()
-	if !c.IsK8sGatewayListerInit(namespace) {
+	if !c.isK8sGatewayListerInit(namespace) {
 		return nil, nil
 	}
 	k8sGateways := []*gatewayapi_v1.Gateway{}
@@ -1554,7 +1554,7 @@ func (c *kubeCache) GetK8sHTTPRoute(namespace, name string) (*gatewayapi_v1.HTTP
 	// but it won't prevent other routines from reading from the lister.
 	defer c.cacheLock.RUnlock()
 	c.cacheLock.RLock()
-	if !c.IsK8sGatewayListerInit(namespace) {
+	if !c.isK8sGatewayListerInit(namespace) {
 		return nil, nil
 	}
 	g, err := c.getCacheLister(namespace).k8shttprouteLister.HTTPRoutes(namespace).Get(name)
@@ -1581,7 +1581,7 @@ func (c *kubeCache) GetK8sHTTPRoutes(namespace, labelSelector string) ([]*gatewa
 	// but it won't prevent other routines from reading from the lister.
 	defer c.cacheLock.RUnlock()
 	c.cacheLock.RLock()
-	if !c.IsK8sGatewayListerInit(namespace) {
+	if !c.isK8sGatewayListerInit(namespace) {
 		return nil, nil
 	}
 	k8sHTTPRoutes := []*gatewayapi_v1.HTTPRoute{}
@@ -1625,7 +1625,7 @@ func (c *kubeCache) GetK8sReferenceGrant(namespace, name string) (*gatewayapi_v1
 	// but it won't prevent other routines from reading from the lister.
 	defer c.cacheLock.RUnlock()
 	c.cacheLock.RLock()
-	if !c.IsK8sGatewayListerInit(namespace) {
+	if !c.isK8sGatewayListerInit(namespace) {
 		return nil, nil
 	}
 	g, err := c.getCacheLister(namespace).k8sreferencegrantLister.ReferenceGrants(namespace).Get(name)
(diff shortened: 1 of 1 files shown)


kubernetes/cache/kube_cache.go
```diff
@@ -1464,11 +1464,21 @@ func (c *kubeCache) GetTelemetries(namespace, labelSelector string) ([]*v1alpha1
 	return retTelemetries, nil
 }
 
+func (c *kubeCache) IsK8sGatewayListerInit(namespace string) bool {
+	// Read lock will prevent the cache from being refreshed while we are reading from the lister
+	// but it won't prevent other routines from reading from the lister.
+	defer c.cacheLock.RUnlock()
+	c.cacheLock.RLock()
+	return c.getCacheLister(namespace).k8sgatewayLister != nil
```
nrfox, line 1472: This is a larger change that you don't need to implement but `getCacheLister` should probably just return an error to handle scenarios like this. More verbose but less risky.
```
func (c *kubeCache) getCacheLister(namespace string) (*cacheLister, error)
```

kubernetes/cache/kube_cache.go
```diff
@@ -1464,11 +1464,21 @@ func (c *kubeCache) GetTelemetries(namespace, labelSelector string) ([]*v1alpha1
...
+}
+
 func (c *kubeCache) GetK8sGateway(namespace, name string) (*gatewayapi_v1.Gateway, error) {
 	if err := checkIstioAPIsExist(c.client); err != nil {
 		return nil, err
 	}
-
+	if !c.IsK8sGatewayListerInit(namespace) {
+		return nil, nil
```
nrfox, line 1480: I think we should always return an error here instead of `nil, nil`. Whatever is calling `GetK8sGateway` can handle the case where the CRDs are installed after the server is started and either log a warning/error or send an error response to the user telling them to restart their kiali server to work with the new CRDs.

It also seems like we should just check this once at startup. Maybe that info can be saved on the client. Not necessary to change for this fix.

nrfox:
> I think this is OK although I'm not sure if we actually need the new `IsK8sGatewayListerInit` function. Each time it's called the check could have been done inside the existing locked section below, prior to the access, which would save an additional read lock. Something like:
> 
> ```
> defer c.cacheLock.RUnlock()
> c.cacheLock.RLock()
> if kbsgatewayLister := c.getCacheLister(namespace).k8sgatewayLister; kbsgatewayList != nil {
>         g, err := k8sgatewayLister.Gateways(namespace).Get(name)
> 	if err != nil {
> 		return nil, err
> 	}
> }
> ```

Yeah you can get rid of the lock in `IsK8sGatewayListerInit` and move the read lock to the very beginning of the function since `checkIstioAPIsExist` should also be covered under the read lock. See: https://github.com/kiali/kiali/pull/6932.

kubernetes/cache/kube_cache.go
hhovsepy: Afaik when CRDs are installed, it is not mandatory that the cluster is configured to use GW API, so showing user an error/warning message in UI can be confusing, so better to log a message. Also as it is not an error/warning, it can be an info message, as we have in documentations that Kiali needs to be restarted.
The issue happens after the Kiali is started, when CRDa are created, we cannot detect it while starting.
I thought that would be odd to add one more variable in `kubeCache` describing whether listers are inited or not.

kubernetes/cache/kube_cache.go
hhovsepy: The `getCacheLister` is just one of the calls alongside with `c.clusterCacheLister` and `c.nsCacheLister`

```
		if c.clusterScoped {
			k8sGateways, err = c.clusterCacheLister.k8sgatewayLister.List(selector)
			if err != nil {
				return nil, err
			}
		} else {
			for _, nsCacheLister := range c.nsCacheLister {
				gatewaysNamespaced, err := nsCacheLister.k8sgatewayLister.List(selector)
				if err != nil {
					return nil, err
				}
				k8sGateways = append(k8sGateways, gatewaysNamespaced...)
			}
		}
	} else {
		k8sGateways, err = c.getCacheLister(namespace).k8sgatewayLister.Gateways(namespace).List(selector)
		if err != nil {
			return nil, err
		}
	}
```
I still think we need to have `lister` existence check outside of the `getCacheLister` method, as the method should not check GW API related listers it has, so keeping the `IsK8sGatewayListerInit` but this time withing the upper caller cache lock:
```
defer c.cacheLock.RUnlock()
	c.cacheLock.RLock()
	if !c.IsK8sGatewayListerInit(namespace) {
		return nil, nil
	}
	g, err := c.getCacheLister(namespace).k8sgatewayLister.Gateways(namespace).Get(name)
	if err != nil {
		return nil, err
	}
```

hhovsepy:
@jshaughn , @nrfox thanks for the review, I removed the odd cache lock from the `IsK8sGatewayListerInit` method and having this call inside a single cache lock block. Also making a `Info` log that the listers are not started. Not sure if it is worthy to have additional variable in `kubeCache` to determine whether listers are loaded or not, so making an info logs each time GW API is called.

kubernetes/cache/kube_cache.go
```diff
@@ -1464,15 +1464,27 @@ func (c *kubeCache) GetTelemetries(namespace, labelSelector string) ([]*v1alpha1
 	return retTelemetries, nil
 }
 
+func (c *kubeCache) IsK8sGatewayListerInit(namespace string) bool {
```
jshaughn, line 1467: Can this be made private?  Not sure it needs to be Capitalized?

kubernetes/cache/kube_cache.go
```diff
@@ -1464,15 +1464,27 @@ func (c *kubeCache) GetTelemetries(namespace, labelSelector string) ([]*v1alpha1
...
+	// K8s GW has several cache listers, those can be namespace or cluster based
+	// if one of them is not initialized, then all others are not init as well
+	// this can happen when CRDs are created after Kiali start
+	if c.getCacheLister(namespace).k8sgatewayLister == nil {
+		log.Info("K8s Gateway API CRDs are installed, Kiali needs to be restarted to apply.")
+	}
+	return c.getCacheLister(namespace).k8sgatewayLister != nil
```
jshaughn, line 1474:
```suggestion
	if c.getCacheLister(namespace).k8sgatewayLister == nil {
		log.Info("K8s Gateway API CRDs are installed, Kiali needs to be restarted to apply.")
		return false
	}
	return true
```

No need to do the lookup/test twice.

jshaughn (changes requested):
I'm good with your new approach, just a minor suggestion.

kubernetes/cache/kube_cache.go
nrfox: You're right the issue here is the `.k8sgatewayLister` part and not the `getCacheLister(namespace)` part.

kubernetes/cache/kube_cache.go
nrfox: This method still needs to return an error when it is returning a `nil` value. Returning `nil, nil` will lead to runtime panics for anything that calls this and only checks `err != nil` before trying to use the return value. Actually taking a second look these methods aren't being used anywhere right now which I think is why there is no runtime error right now but if they were used they'd be used here: https://github.com/kiali/kiali/blob/95e4b4975a81ae486aa2badefc102237cd737b30/business/istio_config.go#L539-L543

Since just the `err` is checked before accessing `istioConfigDetail.K8sGateway.Kind`, which is a perfectly reasonable thing to do, you'd be trying to access a field on a nil pointer and that'd be a runtime error.

kubernetes/cache/kube_cache.go
hhovsepy: agree, I was trying to use that method from `func (in *IstioConfigService) getIstioConfigListForCluster` but it was a bad idea.

hhovsepy:
@nrfox Now changed to return empty list in the methods when multiple objects are requested. And return error when object details are requested.

hhovsepy:
@jshaughn you approval is required to merge this PR.

jshaughn (approved):
Nice revisions, thanks @hhovsepy .

PR #7009, opened 2024-01-05
multi-cluster list page e2e test implementations

### LINKED ISSUE #6474: Multicluster - Automated testing for list pages
Test for presence of `cluster` column in multi-cluster mode on the different list pages:

- Applications
- Services
- Istio config
- Workloads

nrfox:
** Describe the change **

Implements remaining multi-cluster list page cypress tests.

** Issue reference **

Fixes #6474

### DIFF
--- a/frontend/src/components/Validations/Validation.tsx
+++ b/frontend/src/components/Validations/Validation.tsx
@@ -10,7 +10,6 @@ import { Text, TextVariants } from '@patternfly/react-core';
 import { PFColors } from 'components/Pf/PfColors';
 import { kialiStyle } from 'styles/StyleUtils';
 import { IconProps, createIcon } from 'config/KialiIcon';
-import { classes } from 'typestyle';
 
 const validationStyle = kialiStyle({
   textAlign: 'left',
@@ -45,7 +44,7 @@ const InfoValidation: IconProps = {
 const CorrectValidation: IconProps = {
   color: PFColors.Success,
   icon: CheckCircleIcon,
-  className: 'icon-correct-validation'
+  dataTest: 'icon-correct-validation'
 };
 
 const severityToValidation: { [severity: string]: IconProps } = {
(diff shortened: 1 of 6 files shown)


frontend/src/components/Validations/Validation.tsx
```diff
@@ -60,7 +62,7 @@ export const Validation: React.FC<ValidationProps> = (props: ValidationProps) =>
 
   // Set styles
   const textStyle = props.messageColor ? severityColor : {};
-  const iconStyle = kialiStyle(severityColor);
+  const iconStyle = classes(kialiStyle(severityColor), validation.className);
```
nrfox, line 65: @ferhoyos it looks like this keeps the `kialiStyle` prefix as the first classname since it is the first argument to `classes` so I think this will still work with OSSMC?

ferhoyos: good question @nrfox. Two different things to comment here:
- classes is just a util method to concatenate different CSS classes with comma, but they are still independent. That means `validation.className` does not follow OSSMC recommendation to add prefix (this won't make OSSMC to fail, the prefix is just to avoid a possible conflict with Openshift or another plugin that could have same CSS name).
- I think that you have added classname only to be able to filter the element in the test, which is not a good practice. `IconProps` has `dataTest` property, so I recommend you to replace className by dataTest (check `IconProps` interface in `KialiIcon`)

```
dataTest: 'icon-correct-validation'

.find('span[data-test*="icon-correct-validation"]')
```

ferhoyos (changes requested):

frontend/src/components/Validations/Validation.tsx
nrfox: Not sure how I missed the `dataTest` prop that was already there! I'll just use that, thanks.

ferhoyos (dismissed):
LGTM

nrfox:
Had to rebase

ferhoyos (approved):

ferhoyos:
Approved, I thought that rebases didn't dismiss approvals

PR #7011, opened 2024-01-08
Format all staged go and yaml files

ferhoyos:
### Describe the change

Currently we format staged yaml files with `pretty-quick` library only to files inside `frontend/src` folder. I think we should apply it to all yaml files across Kiali application to maintain format consistency (for example I have noticed different yaml indentations in .github workflows).

### Steps to test the PR
1. Modify any go or yaml file with wrong format
2. Commit the change
3. Check that pre-commit hook format the file correctly


jshaughn (dismissed):
This is OK by me.  I'm also going to ask Mazz.

jmazzitelli (approved):
i'm not sure what this does. I edited a yaml file (`hack/istio/prometheus-ingress.yaml`) and added some bad yaml, then did a "git commit -am test" and I don't see it do anything.

I'll approve only because I don't see anything bad happening, but I also don't know what this does :)

ferhoyos:
> i'm not sure what this does. I edited a yaml file (`hack/istio/prometheus-ingress.yaml`) and added some bad yaml, then did a "git commit -am test" and I don't see it do anything.

No, this does not detect bad yaml, only format good yaml the same way (indentation, spaces, this kind of stuff). If you want to see this change in action, do a valid change to `.github/workflows/nightly.yml` and commit the change. The file will be formatted according to prettier rules (.e.g, array items have indent). In summary this change is just to be consistent wrt yaml format across Kiali application.

jmazzitelli (changes requested):
OK, now I see it doing things. And it does too much :)

The indentation for lists is now changed - an extra level is created, which goes against our de-facto standard in virtually all our yaml files. Everything I've written (and, for example, the entire Kiali operator is essentially just a large set of yaml files) has the array lists starting on the level of the parent, not an extra level... i.e.:

```yaml
lists:
- we indent
- like this
```

But this pretty-quick stuff now adds a new level so if I just change a single line in a yaml file, every list item will be shifted like this:
```yaml
lists:
  - we do not
  - indent like this
```

I like the way we do it - I hate that extra indentation level on yaml files personally. And since this is how almost all our yaml files are (at least all the yaml files I seem to come across in the kiali codebase), we should keep it that way.

For example of the major changes this thing does, I added a simple comment `# MAZZ` to this file, and it changed all the lists:

```diff
diff --git a/hack/istio/bookinfo-traffic/bookinfo-service-entry.yaml b/hack/istio/bookinfo-traffic/bookinfo-service-entry.yaml
index 09bedd4bd..341313d17 100644
--- a/hack/istio/bookinfo-traffic/bookinfo-service-entry.yaml
+++ b/hack/istio/bookinfo-traffic/bookinfo-service-entry.yaml
@@ -29,6 +29,7 @@
 # reviews and ratings services to see this new traffic going to the external service.
 #
 # last tested on Istio 1.10 (but could not get traffic to flow - all requests result in an error)
+# MAZZ
 ---
 apiVersion: networking.istio.io/v1beta1
 kind: ServiceEntry
@@ -36,17 +37,17 @@ metadata:
   name: external-service-foo-se
 spec:
   hosts:
-  - foo.bookinfo.ext # this is an arbitrary string; only used in VS
+    - foo.bookinfo.ext # this is an arbitrary string; only used in VS
   location: MESH_EXTERNAL
   ports:
-  - number: 9080
-    name: http-ext
-    protocol: HTTP
+    - number: 9080
+      name: http-ext
+      protocol: HTTP
   resolution: STATIC
   endpoints:
-  - address: 192.168.1.16
-    ports:
-      http-ext: 9080
+    - address: 192.168.1.16
+      ports:
+        http-ext: 9080
 ---
 apiVersion: networking.istio.io/v1beta1
 kind: VirtualService
@@ -54,16 +55,16 @@ metadata:
   name: external-service-foo-vs
 spec:
   hosts:
-  - ratings
+    - ratings
   http:
-  - route:
-    - destination:
-        host: ratings
-        subset: external-service-foo-dr-subset-v1
-      weight: 0
-        host: foo.bookinfo.ext
-      weight: 100
+    - route:
+        - destination:
+            host: ratings
+            subset: external-service-foo-dr-subset-v1
+          weight: 0
+        - destination:
+            host: foo.bookinfo.ext
+          weight: 100
 ---
 apiVersion: networking.istio.io/v1beta1
 kind: DestinationRule
@@ -72,6 +73,6 @@ metadata:
 spec:
   host: ratings
   subsets:
-  - name: external-service-foo-dr-subset-v1
-    labels:
-      version: v1
+    - name: external-service-foo-dr-subset-v1
+      labels:
+        version: v1
```

That's too much churn for such a basic thing like indentation.

We should change the pretty-quick configuration to keep the list indentation the way the files already are. I don't know what else this linter pretties up, so there may be other benefits for using this.

jmazzitelli:
note: I see all the github actions workflows use that extra indentation level. So it looks like we have a mix in here. I can say all the operator ansible scripts do NOT look like that, and we won't be changing those files. And it looks like alot of the hack yamls also have the same indentation as the operator yaml (but not all).

all I can say is, I use the single-level indentation. I won't be using the extra level indentation.

ferhoyos:
I don't usually work with yaml files so I don't have an opinion if it is better to have extra level indentation or not. My main concern here is to be consistent and use same indentation across Kiali application (as you have noticed there is a mix in here). I agree that if most of yaml files in Kiali are with single level indentation the plan should be to convert those ones with extra level indentation to single level.

Bad news is that prettier linter is very opinionated and it is not possible to modify yaml indentation (although there are lots of requests to include it - https://www.github.com/prettier/prettier/issues/12385).

I have tried to search an alternative for prettier but it seems de facto standard in npm world. The only alternative I see is to use pre-commit hook in the go part of the application and format yaml files from there (there is google yaml format in go which allows the non-indent format - https://www.github.com/google/yamlfmt).

The question is: do we have some pre-commit hooks to format/lint go files in the backend similar to the ones we have for typescript files in the frontend?

jshaughn:
I hate to do it, but I do agree with @jmazzitelli.   It is typical for k8s to not indent.  If you look at the Istio example yamls, they are also not indented.  I think the prettier folks are being "pretty" stubborn.  If we can't find tooling to standardize on the no-indent then we will have to leave this alone.

ferhoyos:
I have noticed that there was some pre-commit hook in hack folder to format go files (not used unless the developer executes `git-init` command, which is not documented on README). I have moved that bash code to already established pre-commit hook in frontend folder and added yaml formatter using https://www.github.com/google/yamlfmt. Unlike prettier, with this tool it is possible to configure the array indent, so I think it will work with our yaml standard requirements. 

After this PR, all staged backend (.go), frontend (.tsx and .scss) and yaml files will be formatted with precommit hook.
For go and yaml files there are 2 commit approaches:
1.  Following what it was defined in previous go pre-commit hook, if any file is bad formatted the format is done but the commit is not performed, having the developer to check and add manually format changes.
2. Following what we do with prettier, format changes are added and committed automatically along with developer's changes.

Not sure what is the best approach. For the moment I have followed option 1 to maintain as much as possible previous go pre-commit hook code, but I am open for feedback.

jshaughn (dismissed):
This seems to be ok, will need mazz to agree.  One thing was that my vscode IDE was still formatting using prettier.  But after some wrangling I was able to get it working better:
1. Install the yamlfmt vscode extension
2. Match the config by specifying the following in the settings.json:
```
  "[yaml]": {
    "editor.defaultFormatter": "bluebrown.yamlfmt"
  },
  "yamlfmt.args": [
    "-formatter=type=basic,retain_line_breaks=true,indentless_arrays=true,scan_folded_as_literal=true"
  ]
}
```
The first part may already be there from the extension, but the `yamlfmt.args` needs to be added.

jmazzitelli:
Just so I'm clear on how to test this, is this supposed to gofmt .go files as well as things like yaml files?

I ask because I only see some precommit things commited in the frontend/ folder.  So what if someone just does backend coding and only changes server side go code? Is this supposed PR supposed to kick in too? Or is this just to run a pre-commit script in frontend code?

In short, how is this PR to be tested?

ferhoyos:
> Or is this just to run a pre-commit script in frontend code?

No, the precommit hook applies to any commit and affect the whole code in the repository (there are no .go files in frontend folder). The `precommit.sh` script is stored in frontend folder because the precommit hook is installed through node `husky` library when the developer executes `yarn install` (or `make build-ui`). 

I know that it could be a bit counter intuitive (formatting .go files in frontend?). There is a `husky` library for go, but it is not as mature as node one. Besides, I prefer to have only one pre-commit hook. To avoid confusion maybe I can move `precommit.sh` to root folder, indicating that it is something that applies to whole repo.

In summary, execute `make build-ui`, modify any yaml or go file from anywhere (with wrong format) and commit that change to see the precommit hook in action.

jmazzitelli (approved):
OK, I just tried it. Formatted go and yaml fine. When it detects a file it needs to reformat, it will tell you and not finish the commit - you just have to go in and confirm the changes and do a git add when confirmed it all looks good.

ferhoyos:
Two last minute small changes:
- I have moved pre-commit.sh to previous folder `hack/hooks`. I think this path is more suitable since this pre-commit code affects whole repository, not only frontend folder

- To be consistent with `yaml` and `go` formatters, I have added some flags to pretty-quick to abort the commit if any frontend `.tsx` file is formatted. Then all formatters behave the same. If you think it is better the other option (adding formatted code to the commit automatically) please let me know.

jmazzitelli (dismissed):
ran quick check and it still works like my last review. LGTM

jmazzitelli:
Opps... somehow I EDITED a comment above when I meant to just add this comment:

"I like the abort. Gives me a chance to confirm the changes."

ferhoyos:
Sorry for last minute change, but modifying a cypress test I have noticed that formatter pre-commit hook do not include such files, only .ts and .tsx files from application. I have modified `pretty-quick` to format all `.ts`, `.tsx` and `.scss` from whole application (not only `frontend/src`)

jmazzitelli (approved):
lgtm

PR #7012, opened 2024-01-08
Added new check KIA1601 for ReferenceGrant

hhovsepy:

An Error severity message is added for "Spec.From.Namespace" field of "K8sReferenceGrant" object checking if namespace exists: "KIA1601 Namespace is not found or is not accessible"

Screenshot from 2024-01-08 11-12-03

### DIFF
--- a/business/checkers/k8sreferencegrants/namespace_checker.go
+++ b/business/checkers/k8sreferencegrants/namespace_checker.go
@@ -2,6 +2,7 @@ package k8sreferencegrants
 
 import (
 	"fmt"
+
 	k8s_networking_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
 
 	"github.com/kiali/kiali/models"
--- a/business/checkers/k8sreferencegrants/namespace_checker.go
+++ b/business/checkers/k8sreferencegrants/namespace_checker.go
@@ -9,8 +9,8 @@ import (
 )
 
 type NamespaceChecker struct {
-	ReferenceGrant k8s_networking_v1beta1.ReferenceGrant
 	Namespaces     models.Namespaces
+	ReferenceGrant k8s_networking_v1beta1.ReferenceGrant
 }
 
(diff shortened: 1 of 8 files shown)


business/checkers/k8sreferencegrants/namespace_checker.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"fmt"
+
+	k8s_networking_v1beta1 "sigs.k8s.io/gateway-api/apis/v1beta1"
+
+	"github.com/kiali/kiali/models"
+)
+
+type NamespaceChecker struct {
+	ReferenceGrant k8s_networking_v1beta1.ReferenceGrant
```
josunect, line 12: That could be alphabetized?

josunect (commented):
image
 It works as expected and code looks good, just a small comment. 

Also, when saving the yaml creating an incorrect namespace, this is the message received: 

image

I think it could be a bit more clear with something like: 

"foo Namespace is not found or is not accessible"

hhovsepy:
I think this is working by design of validation notifications, the messages does not contain `{}` variables to replace with strings. They are shown in a format `type:name message`, I agree this message format is a bit outdated. But this can be changed for all validations in a scope of a separate epic.
> 
> Also, when saving the yaml creating an incorrect namespace, this is the message received:
> 
> image
> 
> I think it could be a bit more clear with something like:
> 
> "foo Namespace is not found or is not accessible"

josunect:
> I think this is working by design of validation notifications, the messages does not contain `{}` variables to replace with strings. They are shown in a format `type:name message`, I agree this message format is a bit outdated. But this can be changed for all validations in a scope of a separate epic.
> 
> > Also, when saving the yaml creating an incorrect namespace, this is the message received:
> > image
> > I think it could be a bit more clear with something like:
> > "foo Namespace is not found or is not accessible"

Thanks for clarifying, @hhovsepy !

PR #7016, opened 2024-01-10
K8sReferenceGrant: new Istio config creation wizard item

hhovsepy:

Screenshot from 2024-01-10 14-35-59

### DIFF
--- a/frontend/cypress/integration/featureFiles/wizard_istio_config.feature
+++ b/frontend/cypress/integration/featureFiles/wizard_istio_config.feature
@@ -51,6 +51,17 @@ Feature: Kiali Istio Config wizard
     And user creates the istio config
     Then the "K8sGateway" "k8sapigateway" should be listed in "bookinfo" namespace
 
+  @gateway-api
+  @bookinfo-app
+  Scenario: Create a K8s Reference Grant scenario
+    When user deletes k8sreferencegrant named "k8srefgrant" and the resource is no longer available
+    And user clicks in the "K8sReferenceGrant" Istio config actions
+    And user sees the "Create K8sReferenceGrant" config wizard
+    And user types "k8srefgrant" in the "name" input
+    And user previews the configuration
+    And user creates the istio config
+    Then the "K8sReferenceGrant" "k8srefgrant" should be listed in "bookinfo" namespace
+
   @bookinfo-app
   Scenario: Try to create a Gateway with no name
     When user clicks in the "Gateway" Istio config actions
--- a/frontend/cypress/integration/featureFiles/wizard_istio_config.feature
+++ b/frontend/cypress/integration/featureFiles/wizard_istio_config.feature
@@ -58,6 +58,9 @@ Feature: Kiali Istio Config wizard
     And user clicks in the "K8sReferenceGrant" Istio config actions
     And user sees the "Create K8sReferenceGrant" config wizard
     And user types "k8srefgrant" in the "name" input
+    And user chooses "Gateway" mode from the "ReferenceGrantFromKind" select
+    And user chooses "Secret" mode from the "ReferenceGrantToKind" select
+    And user chooses "istio-system" mode from the "ReferenceGrantFromNamespace" select
     And user previews the configuration
     And user creates the istio config
     Then the "K8sReferenceGrant" "k8srefgrant" should be listed in "bookinfo" namespace
--- a/frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
+++ b/frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
@@ -94,10 +94,13 @@ import { NamespaceDropdown } from '../../components/NamespaceDropdown';
 import { Labels } from '../../components/Label/Labels';
 import { WizardLabels } from '../../components/IstioWizards/WizardLabels';
 
-type Props = {
+type ReduxProps = {
   activeClusters: MeshCluster[];
   activeNamespaces: Namespace[];
   namespacesPerCluster?: Map<string, string[]>;
+};
+
+type Props = ReduxProps & {
   objectType: string;
 };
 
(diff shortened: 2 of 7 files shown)


frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -86,27 +95,28 @@ import { Labels } from '../../components/Label/Labels';
 import { WizardLabels } from '../../components/IstioWizards/WizardLabels';
 
 type Props = {
```
ferhoyos, line 97: Since some of the component props are Redux ones, you should split them in two types:
```
type ReduxProps = {
   activeClusters: MeshCluster[];
  activeNamespaces: Namespace[];
  namespacesPerCluster?: Map<string, string[]>;
}

type Props = ReduxProps & {
  objectType: string;
}
```
Then set `Redux` props as return type of `mapStateToProps` methods.

```
const mapStateToProps = (state: KialiAppState): ReduxProps => {
  return {
    activeClusters: activeClustersSelector(state),
    activeNamespaces: activeNamespacesSelector(state),
    namespacesPerCluster: namespacesPerClusterSelector(state)
  };
};
```

ferhoyos (changes requested):

hhovsepy:
@ferhoyos have approached the comments, thanks, ready to review. Here a point to pay attention is "From Namespace" selector which is not reusing existing `<NamespaceDropdown` as a single namespace is required to stick.

frontend/cypress/integration/featureFiles/wizard_istio_config.feature
```diff
@@ -51,6 +51,17 @@ Feature: Kiali Istio Config wizard
...
     Then the "K8sGateway" "k8sapigateway" should be listed in "bookinfo" namespace
 
+  @gateway-api
+  @bookinfo-app
+  Scenario: Create a K8s Reference Grant scenario
+    When user deletes k8sreferencegrant named "k8srefgrant" and the resource is no longer available
+    And user clicks in the "K8sReferenceGrant" Istio config actions
+    And user sees the "Create K8sReferenceGrant" config wizard
+    And user types "k8srefgrant" in the "name" input
+    And user previews the configuration
```
ferhoyos, line 61: I think you should test `From Namespace`, `From Kind` and `To Kind` selects, although default values are valid for K8sReferenceGrant creation.

There is following action in `wizard_istio_config.feature`:
```
And('user chooses {string} mode from the {string} select', (option: string, id: string) => {
  cy.get(`select[id="${id}"]`).select(option);
});
```
Then it should be easy to add such actions into Reference Grant test scenario

frontend/src/pages/IstioConfigNew/IstioConfigNewPage.tsx
```diff
@@ -143,6 +157,7 @@ export const NEW_ISTIO_RESOURCE = [
   { value: AUTHORIZACION_POLICY, label: AUTHORIZACION_POLICY, disabled: false },
   { value: GATEWAY, label: GATEWAY, disabled: false },
   { value: K8SGATEWAY, label: K8SGATEWAY, disabled: false },
+  { value: K8SREFERENCEGRANT, label: K8SREFERENCEGRANT, disabled: false },
```
ferhoyos, line 160: I think it is better to write variable name in snake case: K8S_REFERENCE_GRANT. The same should be applied to K8S_GATEWAY, but it is not in the scope of this PR.

frontend/src/pages/IstioConfigNew/K8sReferenceGrantForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+  HTTPRoute: 'gateway.networking.k8s.io',
+  Gateway: 'gateway.networking.k8s.io'
+};
+
+export const TO_KINDS = {
+  Service: '',
+  Secret: ''
+};
+
+type ReduxProps = {
```
ferhoyos, line 24: Rename to `ReduxStateProps`, since `ReduxDispatchProps` are also Redux props

frontend/src/pages/IstioConfigNew/K8sReferenceGrantForm.tsx
```diff
@@ -0,0 +1,154 @@
...
+  };
+
+  render(): React.ReactNode {
+    return (
+      <>
+        <FormGroup label="From Namespace" fieldId="FromNamespace">
+          <FormSelect
+            value={this.state.from[0].namespace}
+            onChange={this.onChangeReferenceGrantFromNamespace}
+            id="ReferenceGrantClass"
```
ferhoyos, line 103: Why id and name value is `ReferenceGrantClass` and not `ReferenceGrantFromNamespace`? I think it this id and name represents better the select

ferhoyos (changes requested):
Thanks for the PR @hhovsepy! The form works fine, just small changes and two comments: 
- According to ReferenceGrant specification, this object is used to grant `Gateways` to use `Secrets` or grant `HTTPRoutes` to use `Services` in other namespaces, but no other possibilities (grant `HTTPRoutes` to use `Secrets` or grant `Gateways` to use `Services`), so to avoid wrong configurations maybe it is better to have only one select (`From Kind`) and convert `To Kind` into a label that is changed according to the select:
  - HTTPRoute -> Service
  - Gateway -> Secret

- I guess it is because we do not support them yet, but ReferenceGrant can also be used with other `xRoute` objects (`TLSRoute`, `TCPRoute`, `GRPCRoute` and `UDPRoute`)

hhovsepy:
Thanks for the review @ferhoyos have approached the change requests.
But leaving the 'FromKind' and 'ToKind' fields as they are, as Istio itself ignores a wrong configurations. Also soon will add the rest of routing types.

ferhoyos (approved):
LGTM, the form works fine and comments have been addressed. If Istio ignore wrong configurations, then it is ok to leave from and to kind as they are. Just one small suggestion but it does not block the PR. Well done @hhovsepy!

hhovsepy:
@ferhoyos I miscatched your previous suggestion about 'ReduxStateProps', thanks for suggestion.

Issue #6924, opened 2023-12-05
Use client-go's service account token client refresh


nrfox:
### What do you want to improve?

This is a "tech debt" issue. To address: https://github.com/kiali/kiali/issues/5070, the ability to watch the service account token file and refresh Kiali's kube client/cache was added. It turns out this functionality already exists in the kubernetes go client. In order for this to work however, when you create the `rest.Config` you need to set the `TokenFile` field per: https://github.com/kubernetes/client-go/blob/84a6fe7e4032ae1b8bc03b5208e771c5f7103549/transport/round_trippers.go#L283-L302

The way that Kiali is constructing the `rest.Config`, Kiali does not set this. 
https://github.com/kiali/kiali/blob/9a11e24d993ef34be0de7fbad3d6df946a530bfa/kubernetes/client_factory.go#L88-L93

Kiali should only set this for the Kiali Service Account Client but once that field is set, the client should be able to refresh itself which would simplify the client and cache.

### What is the new behavior?

Token refresh works by relying on the underlying client-go to do the refresh.

jmazzitelli:
> In order for this to work however, when you create the `rest.Config` you need to set the `TokenFile` field per:

I believe you mean `BearerTokenFile` as declared here: https://github.com/kubernetes/client-go/blob/v12.0.0/rest/config.go#L74-L77

PR #7023, opened 2024-01-12
use built-in SA token refresh functionality within client-go

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/6924

### DIFF
--- a/kubernetes/client_factory.go
+++ b/kubernetes/client_factory.go
@@ -160,11 +160,12 @@ func newClientFactory(restConfig *rest.Config) (*clientFactory, error) {
 	return f, nil
 }
 
-// newClient creates a new ClientInterface based on a users k8s token
+// newClient creates a new ClientInterface based on a users k8s token. It is assumed users do not have a token file in authInfo.
 func (cf *clientFactory) newClient(authInfo *api.AuthInfo, expirationTime time.Duration, cluster string) (ClientInterface, error) {
 	config := *cf.baseRestConfig
 
 	config.BearerToken = authInfo.Token
+	config.BearerTokenFile = ""
 
 	// There is a feature when using OpenID strategy to allow using a proxy
 	// for the cluster API.  People may want to place a proxy in
@@ -185,7 +186,7 @@ func (cf *clientFactory) newClient(authInfo *api.AuthInfo, expirationTime time.D
 		var err error
 
 		if cluster == cf.homeCluster {
-			kialiToken, err = GetKialiTokenForHomeCluster()
+			kialiToken, _, err = GetKialiTokenForHomeCluster()
 		} else {
 			kialiToken, err = cf.GetSAClient(cluster).GetToken(), nil
 		}
@@ -250,6 +251,7 @@ func (cf *clientFactory) newClient(authInfo *api.AuthInfo, expirationTime time.D
 		if !(cfg.Auth.Strategy == kialiConfig.AuthStrategyAnonymous) &&
 			!(cfg.Auth.Strategy == kialiConfig.AuthStrategyOpenId && cfg.Auth.OpenId.DisableRBAC) {
 			remoteConfig.BearerToken = authInfo.Token
+			remoteConfig.BearerTokenFile = ""
 		}
 
 		newClient, err = newClientWithRemoteClusterInfo(remoteConfig, &clusterInfo)
@@ -417,64 +419,9 @@ func getTokenHash(authInfo *api.AuthInfo) string {
 
 // KialiSAClients returns all clients associated with the Kiali service account across clusters.
 func (cf *clientFactory) GetSAClient(cluster string) ClientInterface {
-	// while we are here, refresh the client
-	if err := cf.refreshClientIfTokenChanged(cluster); err != nil {
-		log.Errorf("Unable to refresh Kiali SA client for cluster [%s]: %v", cluster, err)
-	}
-
 	cf.mutex.RLock()
 	defer cf.mutex.RUnlock()
 	return cf.saClientEntries[cluster]
-}
-
-// Check for kiali token changes and refresh the client when it does.
-func (cf *clientFactory) refreshClientIfTokenChanged(cluster string) error {
-	var refreshTheClient bool // will be true if the client needs to be refreshed
-	var rci *RemoteClusterInfo
-
-	if cluster == cf.homeCluster {
-		// LOCAL CLUSTER
-		if newTokenToCheck, err := GetKialiTokenForHomeCluster(); err != nil {
-			return err
-		} else {
-			cf.mutex.RLock()
-			client, ok := cf.saClientEntries[cluster]
-			cf.mutex.RUnlock()
-			if !ok {
-				return errors.New("there is no home cluster SA client to refresh")
-			}
-			refreshTheClient = client.GetToken() != newTokenToCheck
-			rci = nil
-		}
-	} else {
-		// REMOTE CLUSTER
-		cf.mutex.RLock()
-		remoteRci, ok := cf.remoteClusterInfos[cluster]
-		cf.mutex.RUnlock()
-		if !ok {
-			return fmt.Errorf("cannot refresh token for unknown cluster [%s]", cluster)
-		} else {
-			if reloadedRci, err := reloadRemoteClusterInfoFromFile(remoteRci); err != nil {
-				return err
-			} else {
-				refreshTheClient = (reloadedRci != nil) // note that anything (not just the token) that changed will trigger the client to be refreshed
-				rci = reloadedRci
-			}
-		}
-	}
-
-	if refreshTheClient {
-		log.Debugf("Kiali SA token has changed for cluster [%s], refreshing the client", cluster)
-		newClient, err := cf.newSAClient(rci)
-		if err != nil {
-			return err
-		}
-		cf.mutex.Lock()
-		cf.saClientEntries[cluster] = newClient
-		cf.mutex.Unlock()
-	}
-
-	return nil
 }
 
 // KialiSAHomeClusterClient returns the Kiali service account client for the cluster where Kiali is running.
@@ -498,13 +445,14 @@ func (cf *clientFactory) getConfig(clusterInfo *RemoteClusterInfo) (*rest.Config
 	} else {
 		// Just read the token and then use the base config.
 		// We're an in cluster client. Read the kiali service account token.
-		kialiToken, err := GetKialiTokenForHomeCluster()
+		kialiToken, kialiTokenFile, err := GetKialiTokenForHomeCluster()
 		if err != nil {
 			return nil, fmt.Errorf("unable to get Kiali service account token: %s", err)
 		}
 
 		// Copy over the base rest config and the token
 		clientConfig.BearerToken = kialiToken
+		clientConfig.BearerTokenFile = kialiTokenFile
 	}
 
 	if !kialiConfig.KialiFeatureFlags.Clustering.EnableExecProvider {
--- a/business/authentication/header_auth_controller.go
+++ b/business/authentication/header_auth_controller.go
@@ -72,7 +72,7 @@ func (c headerAuthController) Authenticate(r *http.Request, w http.ResponseWrite
 		}
 	}
 
-	kialiToken, err := kubernetes.GetKialiTokenForHomeCluster()
+	kialiToken, _, err := kubernetes.GetKialiTokenForHomeCluster()
 	if err != nil {
 		return nil, err
 	}
--- a/business/authentication/openid_auth_controller.go
+++ b/business/authentication/openid_auth_controller.go
@@ -247,7 +247,7 @@ func (c OpenIdAuthController) ValidateSession(r *http.Request, w http.ResponseWr
 		// If RBAC is off, it's assumed that the kubernetes cluster will reject the OpenId token.
 		// Instead, we use the Kiali token and this has the side effect that all users will share the
 		// same privileges.
-		token, err = kubernetes.GetKialiTokenForHomeCluster()
+		token, _, err = kubernetes.GetKialiTokenForHomeCluster()
 		if err != nil {
 			return nil, fmt.Errorf("error reading the Kiali ServiceAccount token: %w", err)
 		}
--- a/business/dashboards.go
+++ b/business/dashboards.go
@@ -6,8 +6,6 @@ import (
 	"strings"
 	"sync"
 
-	"k8s.io/client-go/tools/clientcmd/api"
-
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/config/dashboards"
 	"github.com/kiali/kiali/log"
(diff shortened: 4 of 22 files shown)


jmazzitelli:
@nrfox I think I need to write at least one unit test here (if not more) but I'm still figuring what they will look like. However, I want you to take a look at what I have now. All unit tests pass (indeed, all CI tests are green) but I would like a second pair of eyes on this to make sure I touched all the places (correctly) and that I didn't miss others.

jmazzitelli:
I'm not sure how to test this. The SA token that gets auto-mounted at `/var/run/secrets/kubernetes.io/serviceaccount/token` has a long expiration time. I think we'd need to customize the cluster (at startup) to change that.

Reading some info on the internets, I found this:
* https://medium.com/@th3b3ginn3r/understanding-service-accounts-in-kubernetes-e9d2abe19df8
* https://stackoverflow.com/questions/69375195/kubernetes-projected-service-account-token-expiry-time-issue
* https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/

But I can't find anything that seems like it will make this easy in, say, a unit test (or even a molecule test).

To see the expiration date default, I have Kiali installed and looked at the SA token:

```sh
$ kubectl exec -it -n istio-system deployments/kiali -- cat /var/run/secrets/kubernetes.io/serviceaccount/token | hack/jwt-decode.sh 
Header
{
  "alg": "RS256",
  "kid": "WaO1w3fabzz0RKyNh5Lp9-8ycXIzFDtZOAQ6KXS1gtk"
}
Payload
{
  "aud": [
    "https://kubernetes.default.svc.cluster.local"
  ],
  "exp": 1736946190,
  "iat": 1705410190,
  "iss": "https://kubernetes.default.svc.cluster.local",
  "kubernetes.io": {
    "namespace": "istio-system",
    "pod": {
      "name": "kiali-68b89947f-jfrpn",
      "uid": "c4777af2-cb0b-4fb9-beb1-5ab2ca7dd298"
    },
    "serviceaccount": {
      "name": "kiali-service-account",
      "uid": "66977dbb-1689-4291-8b14-a2e0c45ea2c2"
    },
    "warnafter": 1705413797
  },
  "nbf": 1705410190,
  "sub": "system:serviceaccount:istio-system:kiali-service-account"
}
Signature is NOT valid
```

That "exp" expiry date (`1736946190`) is a 1-year expiration date (Wednesday, 15 January 2025).

nrfox:
> I'm not sure how to test this. The SA token that gets auto-mounted at `/var/run/secrets/kubernetes.io/serviceaccount/token` has a long expiration time. I think we'd need to customize the cluster (at startup) to change that.

@jmazzitelli iirc the default is 1 hour. If the token gets updated automatically after the expiry time then it should be working.

> But I can't find anything that seems like it will make this easy in, say, a unit test (or even a molecule test).

I don't think we need to test the rotation specifically. We rely on the client-go library to do the rotation as long as we set the `BearerTokenFile` field so we can test that we set the field appropriately. Also iirc you can't set the expiry time lower than some threshold (15 mins mabye) so writing an automated real world test for this would require a long running test.

kubernetes/client.go
```diff
@@ -130,7 +130,13 @@ func getConfigForLocalCluster() (*rest.Config, error) {
 	remoteSecretPath := kialiconfig.Get().Deployment.RemoteSecretPath
 	if remoteSecret, readErr := GetRemoteSecret(remoteSecretPath); readErr == nil {
 		log.Debugf("Using remote secret for local cluster config found at: [%s]. Kiali must be running outside the kube cluster.", remoteSecretPath)
-		return clientcmd.NewDefaultClientConfig(*remoteSecret, nil).ClientConfig()
+		cc, err := clientcmd.NewDefaultClientConfig(*remoteSecret, nil).ClientConfig()
+		if err != nil {
+			return cc, err
+		}
+		cc.BearerToken = "" // force use of BearerTokenFile so client can refresh the token
+		cc.BearerTokenFile = remoteSecretPath
```
nrfox, line 138: I don't think you need to do this for the remote secrets. They will use long lived tokens: https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/kiali-prepare-remote-cluster.sh#L127-L141 and I don't think those can be managed or refreshed by client-go since it is for a service account in another cluster.

kubernetes/client.go
```diff
@@ -139,6 +145,7 @@ func getConfigForLocalCluster() (*rest.Config, error) {
 			log.Errorf("Error '%v' getting config for local cluster", err.Error())
 			return nil, err
 		}
+		incluster.BearerToken = "" // force use of BearerTokenFile so client can refresh the token
```
nrfox, line 148: I don't think you need to set this explicitly here either since it gets set already in `rest.InClusterConfig`: https://github.com/kubernetes/client-go/blob/fb8b7346aacefea5ee2ab2e234afc4451c90c435/rest/config.go#L511-L541.

nrfox (commented):
Just glancing over everything and it looks good to me. I think you touched all the areas that needed to get cleaned up.

kubernetes/client.go
jmazzitelli: This code now sets BearerTokenFile everywhere that it is known, and it also sets BearerToken to "" everywhere that the BearerTokenFile is known. I did this for two reasons:

1. This makes everything consistent - where a file is known, then set it; and blank the token in those instances. Otherwise we'll have code that will look strange to future devs ("why is the file set here but not here?"). Just always set the bearer file everywhere that it is known. The file is used under the covers by the go-client and the token is loaded at that time (at least, that's how I am seeing it work). If, for some reason, setting the BearerTokenFile is a no-op and that file is not read to obtain the token, that we can look to adjust this (but if that really happens and the file is ignored, that sounds like a bug - which gets me to my second reason why I did this...)
2. I want to ensure that BearerTokenFile is actually being used - if we set both, we actually cannot be assured all underlying code is using the File if it also has the Token. And we'll never know we have a bug unless some one tests a token past its expiration time (and that would only happen when deployed out in the wild since we don't have any tests that literally test an expiring token).

My latest commit has a unit test that now looks at the bearer token and bearer token file fields to make sure they are set for the home cluster and a remote cluster.

jmazzitelli:
ok, I think I found a problem in this entire thing. Unit test failures are bringing this up.

func (cf *clientFactory) GetClients(authInfo *api.AuthInfo) (map[string]ClientInterface, error) {

expects a token in the authInfo. And I think we use this sometimes with SA auth.

jmazzitelli:
> expects a token in the authInfo. And I think we use this sometimes with SA auth.

Ah.. no, I think it is because these unit tests are using some functions as a convenience to create some user tokens, but the functions aren't behaving like they did before. Gonna need to rethink some of the test code here.

kubernetes/client.go
nrfox: > I want to ensure that BearerTokenFile is actually being used - if we set both, we actually cannot be assured all underlying code is using the File if it also has the Token.

I think you can legitimately set one but not the other i.e. you can set `BearerToken` but not `BearerTokenFile`. Right now for the remote config, we only override `qps` and `burst` but all other settings come straight from user configuration and I don't want that to change by overriding more fields: https://github.com/kiali/kiali/blob/7eb710750713a2ee70a9a6f6d1252b3da2ac9f4a/kubernetes/client_factory.go#L491-L497

The only thing that needs to change is for the kiali SA client for the home cluster to also set the `BearerTokenFile`. If we were to just use `rest.InClusterConfig` then this would be happening already. For instance, I'm not sure why we copy over the token here instead of using `rest.InClusterConfig`: https://github.com/kiali/kiali/blob/7eb710750713a2ee70a9a6f6d1252b3da2ac9f4a/kubernetes/client_factory.go#L499-L507

jmazzitelli: > I think you can legitimately set one but not the other i.e. you can set `BearerToken` but not `BearerTokenFile`.

Docs imply you can set both, but a non-empty File means the Token will be refreshed (except for when using OAuth2, which is why I wasn't sure if I should leave Token as-is or empty it out.. I'm thinking I'm going to revert that change and just leave them both set.). Here's the docs:

https://pkg.go.dev/k8s.io/client-go@v0.28.3/rest#Config.BearerToken

```go
// Server requires Bearer authentication. This client will not attempt to use
// refresh tokens for an OAuth2 flow.
// TODO: demonstrate an OAuth2 compatible client.
BearerToken [string](https://pkg.go.dev/builtin#string) `datapolicy:"token"`

// Path to a file containing a BearerToken.
// If set, the contents are periodically read.
// The last successfully read value takes precedence over BearerToken.
BearerTokenFile [string](https://pkg.go.dev/builtin#string)
```

jmazzitelli: > For instance, I'm not sure why we copy over the token here instead of using `rest.InClusterConfig`

Yeah, there was a couple things in the code that made me scratch my head wondering why we were doing things. I think changes happened rapidly in here and we just missed a few places.

jmazzitelli: I reverted some of the code I originally pushed. I no longer blank the token. The docs imply that the token will be overridden if File is provided and the File is refreshed. My paranoia just caused me to do more work, so I reverted that. I'll trust the go-client does the right thing.

jmazzitelli: > For instance, I'm not sure why we copy over the token here instead of using `rest.InClusterConfig`:

Well, I found out why we do this. It uses the base config that was passed to the factory, not the InClusterConfig. And it does this because it is mock-friendly (we must pass in mock configs when creating factories in tests - indeed, I see errors in the tests when I changed this code to use InClusterConfig).

See: https://github.com/kiali/kiali/blob/v1.78.0/kubernetes/client_factory.go#L119-L121

```
// newClientFactory allows for specifying the config and expiry duration
// Mock friendly for testing purposes
func newClientFactory(restConfig *rest.Config) (*clientFactory, error) {
```

kubernetes/client_factory.go
```diff
@@ -493,18 +440,22 @@ func (cf *clientFactory) getConfig(clusterInfo *RemoteClusterInfo) (*rest.Config
 			return nil, err
 		}
 
+		// Set the token file so the underlying client can refresh it when needed.
+		remoteConfig.BearerTokenFile = clusterInfo.SecretFile
```
nrfox, line 444: I'm still leery of manually overriding these settings and I think this already gets set by `client-go` if you have a `SecretFile` in your kubeconfig?

kubernetes/client.go
```diff
@@ -130,7 +130,12 @@ func getConfigForLocalCluster() (*rest.Config, error) {
 	remoteSecretPath := kialiconfig.Get().Deployment.RemoteSecretPath
 	if remoteSecret, readErr := GetRemoteSecret(remoteSecretPath); readErr == nil {
 		log.Debugf("Using remote secret for local cluster config found at: [%s]. Kiali must be running outside the kube cluster.", remoteSecretPath)
-		return clientcmd.NewDefaultClientConfig(*remoteSecret, nil).ClientConfig()
+		cc, err := clientcmd.NewDefaultClientConfig(*remoteSecret, nil).ClientConfig()
+		if err != nil {
+			return cc, err
+		}
+		cc.BearerTokenFile = remoteSecretPath
```
nrfox, line 137: Same comment here about not overriding what the user provides in the remote kube config. We should be able to verify that if you set `SecretFile` in your kubeconfig that `client-go` will in turn set `BearerTokenFile`.

nrfox (commented):
Some minor comments but otherwise looks good. I'll test this out to make sure that when the token gets rotated, access still works.

jmazzitelli:
int tests showing failures like this:
```
E0116 20:51:37.072155       1 reflector.go:147] pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: Failed to watch *v1.Deployment: failed to list *v1.Deployment: Get "https://172.18.0.5:6443/apis/apps/v1/deployments?limit=500&resourceVersion=0": net/http: invalid header field value for "Authorization"
```
Looks like I nicked something and causing that header to have an invalid value. The logs don't tell me the value, though.

kubernetes/client.go
jmazzitelli: let me check that. I was assuming it wasn't going to set that under the covers. But I'll check. This could be why I am getting errors in CI.

kubernetes/client_factory.go
jmazzitelli: not set. I change this line 444 to this:
```go
		//MAZZ DELETE remoteConfig.BearerTokenFile = clusterInfo.SecretFile
		log.Errorf("MAZZ2 btf=[%v];sf=[%v]", remoteConfig.BearerTokenFile, clusterInfo.SecretFile)
```
and in the logs I see:
```
2024-01-17T19:09:02Z ERR MAZZ2 btf=[];sf=[/kiali-remote-cluster-secrets/kiali-remote-cluster-secret-west/west]
```

jmazzitelli: But yet when I leave line 444 the way it is here, I get this error:
```
2024-01-17T19:12:17Z ERR MAZZ2 btf=[/kiali-remote-cluster-secrets/kiali-remote-cluster-secret-west/west];sf=[/kiali-remote-cluster-secrets/kiali-remote-cluster-secret-west/west]
...
W0117 19:12:27.423526       1 reflector.go:535] pkg/mod/k8s.io/client-go@v0.28.3/tools/cache/reflector.go:229: failed to list *v1.ReplicaSet: Get "https://172.18.0.5:6443/apis/apps/v1/replicasets?limit=500&resourceVersion=0": net/http: invalid header field value for "Authorization"
```
I suspect I don't understand what value the BearerTokenFile should be.

jmazzitelli: I was assuming "BearerTokenFile" is the same as the kubeconfig file, but apparently it is not. The BearerTokenFile can be specified in the kubeconfig, and I believe that is what should get set here. We do not have BearerTokenFiles defined in our generated remote cluster secrets.

So I have a feeling this client-go feature is for something slightly different that what we are expecting to be able to use it. I don't think this is something we can use (i.e. I think we should keep our code that refreshes).

@nrfox See if you understand this differently than me, and let me know what I'm missing.

nrfox: > I was assuming "BearerTokenFile" is the same as the kubeconfig file, but apparently it is not.

`BearerTokenFile` is not the same as the kubeconfig. It is a file that contains the `BearerToken`. In the case of the `InClusterConfig` it is just `/var/run/secrets/kubernetes.io/serviceaccount/token`.

jmazzitelli: OK, I think I get it. I was thinking this was going to refresh everything, not just the in-cluster client.

kubernetes/client.go
jmazzitelli: I removed this

kubernetes/client_factory.go
jmazzitelli: I removed this.

jmazzitelli:
OK, now I only have the bearer token file explicitly being set for when using the in-cluster SA token.

nrfox (approved):

Issue #7022, opened 2024-01-12
Graph: Link to App which does not exist


hhovsepy:
Happen on Kiali master.
Setup an env with waypoint proxy.
```
kind create cluster --config=- <<EOF
kind: Cluster
apiVersion: kind.x-k8s.io/v1alpha4
name: ambient
nodes:
- role: control-plane
- role: worker
- role: worker
EOF
./hack/istio/install-istio-via-istioctl.sh -c kubectl -cp ambient 
./hack/istio/install-bookinfo-demo.sh -c kubectl -ai false -tg
kubectl apply -f _output/istio-1.20.0/samples/addons/kiali.yaml
kubectl create clusterrolebinding kiali-binding --clusterrole=cluster-admin --serviceaccount=istio-system:kiali
kubectl apply -k "[github.com/kubernetes-sigs/gateway-api/config/crd?ref=v1.0.0](http://github.com/kubernetes-sigs/gateway-api/config/crd?ref=v1.0.0)"
./_output/istio-1.20.1/bin/istioctl x waypoint apply -n bookinfo
```

Go to Graph page and select the waypoing proxy `namespace-istio-waypoint`, on the right side panel of Graph you will see a reference to 'Application' which does not exist:
![Screenshot from 2024-01-12 15-07-37](https://github.com/kiali/kiali/assets/604313/30fb7969-513b-45ff-83cf-69197aea7275)
![Screenshot from 2024-01-12 15-07-52](https://github.com/kiali/kiali/assets/604313/53a285e6-d1c5-4e92-8180-44b752455069)

The workload of that waypoint proxy does not contain 'app' label:
![Screenshot from 2024-01-12 15-11-34](https://github.com/kiali/kiali/assets/604313/30ee64c3-4cbd-4ce9-bd24-fcd4aaa97c53)

Expected behavior: No Application link in graph, only the link to Workload.

PR #7026, opened 2024-01-15
Hide application link for waypoint in graph

josunect:
**Describe the change**

It looks like the graph generation is not looking at the labels of the workload, it is looking at the [source_canonical_service](https://github.com/kiali/kiali/blob/master/graph/telemetry/istio/istio.go#L280) to fill the fill the app field. 

That's the reason to include the waypoint check in the frontend code. 

**Issue reference**

Ref. https://github.com/kiali/kiali/issues/7022

### DIFF
--- a/frontend/src/pages/Graph/SummaryPanelNode.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelNode.tsx
@@ -48,8 +48,6 @@ const defaultState: SummaryPanelNodeState = {
   isActionOpen: false
 };
 
-const waypointPrefix = 'istio-waypoint';
-
 type ReduxProps = {
   kiosk: string;
   rankResult: RankResult;
@@ -226,7 +224,7 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
               )}
 
               {secondBadge}
-              {!this.isWaypoint(nodeData) && (
+              {!nodeData.isWaypoint && (
                 <div className={nodeInfoStyle}>
                   {renderBadgedLink(nodeData)}
                   {renderHealth(nodeData.health)}
@@ -249,10 +247,6 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
       </div>
     );
   }
-
-  private isWaypoint = (nodeData: DecoratedGraphNodeData): boolean => {
-    return nodeData.app?.endsWith(waypointPrefix) ? true : false;
-  };
 
   private renderWorkloadSection = (nodeData: DecoratedGraphNodeData): React.ReactNode => {
     if (!nodeData.hasWorkloadEntry) {
--- a/frontend/src/pages/Graph/SummaryPanelNode.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelNode.tsx
@@ -129,7 +129,7 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
     const shouldRenderDestsList = destsList && destsList.length > 0;
     const shouldRenderSvcList = servicesList && servicesList.length > 0;
     const shouldRenderService = service && ![NodeType.SERVICE, NodeType.UNKNOWN].includes(nodeType);
-    const shouldRenderApp = app && ![NodeType.APP, NodeType.UNKNOWN].includes(nodeType);
+    const shouldRenderApp = app && ![NodeType.APP, NodeType.UNKNOWN].includes(nodeType) && !nodeData.isWaypoint;
     const shouldRenderWorkload = workload && ![NodeType.WORKLOAD, NodeType.UNKNOWN].includes(nodeType);
     const shouldRenderTraces =
       !isServiceEntry &&
--- a/frontend/src/pages/Graph/SummaryPanelNode.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelNode.tsx
@@ -48,6 +48,8 @@ const defaultState: SummaryPanelNodeState = {
   isActionOpen: false
 };
 
+const waypointPrefix = 'istio-waypoint';
+
 type ReduxProps = {
   kiosk: string;
   rankResult: RankResult;
@@ -224,11 +226,12 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
               )}
 
               {secondBadge}
-
-              <div className={nodeInfoStyle}>
-                {renderBadgedLink(nodeData)}
-                {renderHealth(nodeData.health)}
-              </div>
+              {!this.isWaypoint(nodeData) && (
+                <div className={nodeInfoStyle}>
+                  {renderBadgedLink(nodeData)}
+                  {renderHealth(nodeData.health)}
+                </div>
+              )}
             </span>
           </div>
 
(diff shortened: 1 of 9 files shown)


jshaughn (commented):
The fundamental problem here is that we're trying to introduce Istio infrastructure into the application graph. To some degree we do this already with sidecar proxies, but because they are bundled with application pods it's not as explicit and obtrusive.    But with waypoints Istio is leaking this infra telemetry into the application namespaces. I'm not even sure Istio should be generating this telemetry, and maybe it will go away.  But for now it does exist and so I guess we need to deal with it.  We can't just treat these nodes like standard application nodes, and I think there may be multiple places where we will find ourselves wanting to do some special-case handling.

I don't really love this approach.  Instead, I'd recommend that we make this determination on the server side, as part of what happens in the new "ambient" appender.   We can add an `IsWaypoint` metadata and just pass it up as node data.  Then the UI can do what it wants to do whenever there is special handling involved.

josunect:
> The fundamental problem here is that we're trying to introduce Istio infrastructure into the application graph. To some degree we do this already with sidecar proxies, but because they are bundled with application pods it's not as explicit and obtrusive. But with waypoints Istio is leaking this infra telemetry into the application namespaces. I'm not even sure Istio should be generating this telemetry, and maybe it will go away. But for now it does exist and so I guess we need to deal with it. We can't just treat these nodes like standard application nodes, and I think there may be multiple places where we will find ourselves wanting to do some special-case handling.
> 
> I don't really love this approach. Instead, I'd recommend that we make this determination on the server side, as part of what happens in the new "ambient" appender. We can add an `IsWaypoint` metadata and just pass it up as node data. Then the UI can do what it wants to do whenever there is special handling involved.

I like this approach, it can also be really useful for other issues.
In the last commit I've included the _isOutOfMesh_ to fix this issue: https://github.com/kiali/kiali/issues/7027 as I think it could also be part of this appender.

jshaughn:
> I like this approach, it can also be really useful for other issues.
In the last commit I've included the isOutOfMesh to fix this issue: https://github.com/kiali/kiali/issues/7027 as I think it could also be part of this appender.

+1.  In general, I think it is best if the client-side is as dumb as possible, and worried specifically about "presentation", minimizing any other sort of logic.  With this approach we defer the logic to the back-end and the front-end just reacts to the information.

frontend/src/types/Graph.ts
```diff
@@ -474,6 +474,8 @@ export interface DecoratedGraphNodeData extends GraphNodeData {
   httpOut: number;
   // true if has istio namespace
   isIstio?: boolean;
+  // true if is an Ambient Istio waypoint
+  isWaypoint?: boolean;
```
jshaughn, line 478: This should actually be moved to the base type,`GraphNodeData`.    You may also want to add this in the `NodeAttr` definition.

graph/telemetry/istio/appender/ambient.go
```diff
@@ -34,11 +34,13 @@ func (a AmbientAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *gr
 	log.Trace("Running ambient appender")
 
 	if !a.Waypoints {
-		a.removeWaypointEntries(trafficMap, globalInfo, namespaceInfo)
+		a.isWaypoint(trafficMap, globalInfo, true)
+	} else {
+		a.isWaypoint(trafficMap, globalInfo, false)
 	}
```
jshaughn, line 40: Fine as is, but could reduce to just one line:

`a.isWaypoint(trafficMap, globalInfo, !a.Waypoints)`

jshaughn (commented):
@josunect Looks good, this is the approach we talked about.  A couple of comments.  Also, if you wanted to you could add waypoints to graph find/hide, but maybe it's not necessary given that we already have the Display option.

frontend/src/types/Graph.ts
josunect: Moved and added also to NodeAttr

graph/telemetry/istio/appender/ambient.go
josunect: oh! Good catch

frontend/src/pages/Graph/SummaryPanelNode.tsx
```diff
@@ -224,11 +224,12 @@ export class SummaryPanelNodeComponent extends React.Component<SummaryPanelNodeC
...
-              <div className={nodeInfoStyle}>
-                {renderBadgedLink(nodeData)}
-                {renderHealth(nodeData.health)}
-              </div>
+              {!nodeData.isWaypoint && (
+                <div className={nodeInfoStyle}>
+                  {renderBadgedLink(nodeData)}
+                  {renderHealth(nodeData.health)}
+                </div>
+              )}
```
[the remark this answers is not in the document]
josunect: But this is really the problem: 

image

The waypoint proxy doesn't have an app label, so the app link or text should not be there, and also the health (Probably is using the label to select the workload to get the health so it does not exist and appears as unknown)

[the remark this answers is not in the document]
hhovsepy: @josunect @jshaughn wdyt to show some label or info near the '**Application**', like '(Waypoint proxy)' or something similar, as the application link or text does not exist anymore.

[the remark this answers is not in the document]
josunect: Maybe we can add something like this? 

image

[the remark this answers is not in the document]
hhovsepy: +1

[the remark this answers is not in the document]
jshaughn: I'm also +1.  One suggested change, please update the PfBadge from `W` to `WP`.  I think it's a little confusing as `W` is the PfBadge for Workload, which is used a lot.  `WP` is already used for WasmPlugin, but that is much less of an overlap, and never shows up in the graph side panel.

frontend/src/types/Graph.ts
```diff
@@ -399,6 +399,8 @@ export interface GraphNodeData {
   isOutside?: boolean;
   isRoot?: boolean;
   isServiceEntry?: SEInfo;
+  // true if is an Ambient Istio waypoint
```
jshaughn, line 402:
```suggestion
```

graph/telemetry/istio/appender/ambient.go
```diff
@@ -33,12 +33,10 @@ func (a AmbientAppender) AppendGraph(trafficMap graph.TrafficMap, globalInfo *gr
...
 	log.Trace("Running ambient appender")
 
-	if !a.Waypoints {
-		a.removeWaypointEntries(trafficMap, globalInfo, namespaceInfo)
-	}
+	a.isWaypoint(trafficMap, globalInfo, !a.Waypoints)
 }
 
-func (a AmbientAppender) removeWaypointEntries(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo, namespaceInfo *graph.AppenderNamespaceInfo) {
+func (a AmbientAppender) isWaypoint(trafficMap graph.TrafficMap, globalInfo *graph.AppenderGlobalInfo, remove bool) {
```
jshaughn, line 39: Very minor, but `isWaypoint` seems like a function that would test something and return a bool.  Maybe rename to something like `processWaypoints` or `handleWaypoints` ?

jshaughn (commented):
Looking good, some final comments...

graph/telemetry/istio/appender/ambient.go
josunect: Changed to handleWaypoints

jshaughn (approved):
Perfecto!  With the work we are doing to handle it, I hope Istio continues to generate the waypoint telem :-)

hhovsepy (approved):
thanks @josunect looks good:
Screenshot from 2024-01-18 17-34-42

PR #7028, opened 2024-01-15
added the arm arch

pbajjuri20:
Update the ARM arch to support for the ARM cluster

### DIFF
--- a/hack/istio/install-testing-demos.sh
+++ b/hack/istio/install-testing-demos.sh
@@ -74,7 +74,7 @@ HELPMSG
 done
 
 # check arch values
-if [ "${ARCH}" != "ppc64le" ] && [ "${ARCH}" != "s390x" ] && [ "${ARCH}" != "amd64" ]; then
+if [ "${ARCH}" != "ppc64le" ] && [ "${ARCH}" != "s390x" ] && [ "${ARCH}" != "amd64" ] && [ "${ARCH}" != "arm" ]; then
   echo "${ARCH} is not supported. Exiting."
   exit 1
 fi


hack/istio/install-testing-demos.sh
```diff
@@ -74,7 +74,7 @@ HELPMSG
 done
 
 # check arch values
-if [ "${ARCH}" != "ppc64le" ] && [ "${ARCH}" != "s390x" ] && [ "${ARCH}" != "amd64" ]; then
+if [ "${ARCH}" != "ppc64le" ] && [ "${ARCH}" != "s390x" ] && [ "${ARCH}" != "amd64" ] && [ "${ARCH}" != "arm" ]; then
```
ScriptingShrimp, line 77: please add mention of arm on [line 59](https://github.com/kiali/kiali/blob/2bd3c211592905e84bb78507a2c796b0b6b4f89a/hack/istio/install-testing-demos.sh#L59C8-L59C8) 

on other lines of this file there is reference to `${ARCH}` when calling other scripts, please ensure that other scripts are compatible as well 

- [ ] [line 108](https://github.com/kiali/kiali/blob/2bd3c211592905e84bb78507a2c796b0b6b4f89a/hack/istio/install-testing-demos.sh#L108) /install-bookinfo-demo.sh
- [ ] [line 110](https://github.com/kiali/kiali/blob/2bd3c211592905e84bb78507a2c796b0b6b4f89a/hack/istio/install-testing-demos.sh#L110) install-error-rates-demo.sh

currently install-sleep-demo.sh does not support arch variable, it might cause issues

Additionally we want to merge this into 1.73 branch for OSSM 2.5

ScriptingShrimp (changes requested):

Issue #7030, opened 2024-01-15
Remove graph "Compress-On_hide" Display option


jshaughn:
This Display option is probably not used, and likely is not well-understood.  It has historically been flaky and now, with the addition of the new "Show Waypoints" option, the Display menu is getting very long, often occluding the on-screen graph icons.  Finally, the feature can not be offered in the Patternfly version of the graph, only Cytoscape.

When `Compress-on-hide` is enabled, which is the default, it means that when the user hides graph elements (via find/hide) there will be a graph re-layout in order to compress the graph into a smaller form, after the nodes are removed.  When not enabled the graph elements remain in place, with the missing elements simply being removed.  This is rarely, if ever, desired, and so for simplicity, this option will be removed.

PR #7031, opened 2024-01-15
Remove graph CompressOnHide option.

jshaughn:
closes https://github.com/kiali/kiali/issues/7030

The option should no longer show up in graph Display menu and the behavior should now always be to compress the graph (i.e. re-layout as if the hidden elements don't exist).


josunect (approved):
Works as expected: 

image

No option in the menu, it works as if compressOnHide was enabled. 

Code LGTM!

PR #7032, opened 2024-01-15
[cypress] Add graph side panel multi-cluster test implementation

### LINKED ISSUE #6476: Multicluster - Automated testing for multi-cluster graph scenarios
Test for multi-cluster specific features of the graph page:

- Check that the graph has two regions.
- Clicking on a node in the "remote" region links you to the details page for that node.

nrfox:
** Describe the change **

Adds remaining graph multi-cluster test implementations.

** Issue reference **

Fixes #6476

### DIFF
--- a/frontend/cypress/integration/common/graph_context_menu.ts
+++ b/frontend/cypress/integration/common/graph_context_menu.ts
@@ -1,4 +1,4 @@
-import { After, Then, When } from '@badeball/cypress-cucumber-preprocessor';
+import { Then, When } from '@badeball/cypress-cucumber-preprocessor';
 import { clusterParameterExists } from './navigation';
 import { ensureKialiFinishedLoading } from './transition';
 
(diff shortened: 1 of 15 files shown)


frontend/cypress/integration/featureFiles/graph_context_menu.feature
```diff
@@ -9,14 +9,12 @@ Feature: Kiali Graph page - Context menu actions
     Given user is at administrator perspective
     When user graphs "bookinfo" namespaces
 
-  @bookinfo-app
```
matejnesuta, line 12: Do we really want to delete this? This tag is for the demo-app detection and the test is for single cluster.

frontend/cypress/integration/featureFiles/graph_context_menu.feature
```diff
@@ -9,14 +9,12 @@ Feature: Kiali Graph page - Context menu actions
...
   Scenario: Actions in context menu for service node with existing traffic routing
     And user opens the context menu of the "productpage" service node
     And user should see "no" cluster parameter in links in the context menu
     And user clicks the "delete-traffic-routing" item of the context menu
     Then user should see the confirmation dialog to delete all traffic routing
 
-  @bookinfo-app
```
matejnesuta, line 19: Same as the tag before.

frontend/cypress/integration/common/graph_context_menu.ts
```diff
@@ -44,3 +84,45 @@ Then('user should see {string} cluster parameter in links in the context menu',
...
+  'user should see the {string} cluster parameter in the {string} link in the context menu',
+  (cluster: string, linkText: string) => {
+    cy.get(`[data-test="graph-node-context-menu"]`).within(() => {
+      cy.get('a').contains(linkText).should('have.attr', 'href').and('include', `clusterName=${cluster}`);
+    });
+  }
+);
+
+// @istio-config-cleanup
+After({ tags: '@istio-config-cleanup' }, () => {
```
matejnesuta, line 98: I love the idea, but I have a few suggestions:

It is usually better to perform cleanups optionally before the execution of a specific test and not after. The remaining objects can be still handy for debugging purposes.

The name of the tag seems too general, as it does not cover all of the Istio configs. 

I would also move this hook to the hooks.ts file.

matejnesuta (changes requested):

frontend/cypress/integration/featureFiles/graph_context_menu.feature
nrfox: Ah no I had some issues with this hook when running locally so I removed it but forgot to add it back.

frontend/cypress/integration/common/graph_context_menu.ts
nrfox: > It is usually better to perform cleanups optionally before the execution of a specific test and not after. The remaining objects can be still handy for debugging purposes.

This is true but keeping these objects around affects actual traffic on the graph which affects the rest of the tests. With multi-primary, you actually need to setup the request routing for each cluster that your service is located in otherwise traffic won't be sent to the service located in the cluster without the request routing objects created there. At first I tried duplicating the objects to both clusters but this was more awkward than just deleting the objects right after the test completed. I can give that another try.

nrfox: Alright now the configuration gets duplicated to both clusters as part of the scenario rather than removing the istio config at the end.

matejnesuta:
LGTM

matejnesuta (approved):

PR #7037, opened 2024-01-17
removed all mentions of kiali-bot

FreelancingTwin:
** Describe the change **
removed all mentions of kiali-bot

** Issue reference **
#6515


frontend/getLabels.py
```diff
@@ -3,7 +3,6 @@
 labels=''
 rawLabels=json.loads(sys.argv[1])['labels']
 for rLabel in rawLabels:
-  if rLabel['name'].startswith("kiali-bot"):
```
jmazzitelli, line 6: seems odd we would just remove this outer if-stmt conditional and yet leave the remaining inner if-stmts as-is. Seems like this would change behavior.

@aljesusg you created this "getLabels.py" file way back in 2018 and it hasn't been touched since. Do you know if it is even used anymore? I don't see where it is used. My question is: can we just delete this frontend/getLabels.py file entirely?

.github/troubleshooting.md
```diff
@@ -29,8 +29,7 @@ In case of errors in any of these steps, the log should be inspected to
 check what was the last action the pipeline executed. The pipelines are not idempotent so to retry them if there was a failure in the release step, manual amends should be done to prepare the next run:
 
 * Delete the tags and the release on GitHub (if exists)
-* Revert the commits made by the pipeline on the version branch (if there are commits made by kiali-bot on the same time around the pipeline run)
-* Delete the PR generated if exists
```
jmazzitelli, line 33: @leandroberetta you wrote this troubleshooting doc last year. Can you review this? I don't think these lines should be outright deleted. I think just mention of kiali-bot should be removed. However, we do still generate commits and PRs from GitHub Actions that need to be reverted in the case of manual intervention, which is what this part of the docs are talking about.

Also, please review the other two changes in here.

.github/troubleshooting.md
```diff
@@ -39,7 +38,7 @@ The Helm charts pipeline has a "release" job that in case of errors, the log sho
 The manual actions to do in case of a failure are:
 
 * Delete the $RELEASE_VERSION-main branch (if exists)
-* Revert the last commit in the $RELEASE_VERSION branch (if there is one made by kiali-bot on the same time around the pipeline run)
```
jmazzitelli, line 42: @leandroberetta please review this line. I don't think we should delete this entirely. Perhaps just remove mention of kiali-bot specifically (since kiali-bot user isn't involved now). Perhaps something like this?

```
* Revert the last commit in the $RELEASE_VERSION branch (if there is one made by the GitHub action at the same time around the pipeline run)
```

.github/troubleshooting.md
```diff
@@ -50,5 +49,5 @@ The Kiali.io pipeline has a "release" job that in case of errors, the log should
 The manual actions to do in case of a failure are:
 
 * Delete the $RELEASE_VERSION branch (if exists)
-* Revert the last commits made by kiali-bot on the staging branch
```
jmazzitelli, line 53: @leandroberetta again, review this line and see if we should say something different? Right now this is being deleted and I don't think it should be deleted.
```
* Revert the last commits made by GitHub Actions on the staging branch
```

jmazzitelli (changes requested):
I am asking @leandroberetta and @aljesusg to review some of the changes here, as I think some things do not look correct.

frontend/getLabels.py
aljesusg: Well this was a requirement from qe to check the labels of the PR. I think that is deprecated because we are not using it so we can remove it.

jmazzitelli: @FreelancingTwin as per discussion above, we can delete this `frontend/getLabels.py` file entirely.

.github/troubleshooting.md
leandroberetta: It's true that the kiali-bot account is gone, but, the gh actions requires to configure the user and email (check https://github.com/kiali/kiali/blob/master/.github/workflows/release.yml#L246), I'm not sure if today there is a better way to do this, in some post back in that time, this was needed. And the kiali-bot works (really, whatever there will work) so there is still a commit by kiali-bot:

image

So, right, it doesn't need to be removed, the github-actions[bot] user also is part of the commit, so I think this requires an investigation to see if we can remove the kiali-bot configuration in our actions, but whoever does the commit, if there is any error, the commit should be reverted.

.github/troubleshooting.md
leandroberetta: Same here.

jshaughn:
@jmazzitelli , can we move this forward, are we still waiting for review from @leandroberetta and/or @aljesusg ?

jmazzitelli:
we can't change the email stuff (this PR removes it).. we tried that and the CI actions break. git requires those emails.

This is so minor an issue and most of this removals in this PR are not going to work, so I say just close it.

jshaughn:
Closing as most of the work is done, and this one probably not worth moving forward.

PR #7038, opened 2024-01-17
Fix nil error in OCP

hhovsepy:
Kiali pod is failing with nil error on OCP cluster in `.isK8sGatewayListerInit`.

Seems checking only cache lister `c.getCacheLister(namespace).k8sgatewayLister == nil` is causing `nil` in case of cluster or namespace scope of caches. 
Added a `nil` check similar to `GetK8sGateways` method structure.

### DIFF
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -1467,12 +1467,27 @@ func (c *kubeCache) GetTelemetries(namespace, labelSelector string) ([]*v1alpha1
 }
 
 func (c *kubeCache) isK8sGatewayListerInit(namespace string) bool {
-	// K8s GW has several cache listers, those can be namespace or cluster based
-	// if one of them is not initialized, then all others are not init as well
-	// this can happen when CRDs are created after Kiali start
-	if c.getCacheLister(namespace).k8sgatewayLister == nil {
-		log.Info(K8sGatewayAPIMessage)
-		return false
+	// potential issue can happen when CRDs are created after Kiali start
+	if namespace == metav1.NamespaceAll {
+		if c.clusterScoped {
+			if c.clusterCacheLister.k8sgatewayLister == nil {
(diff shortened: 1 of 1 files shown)


hhovsepy:
Kiali run fine now on OCP.

kubernetes/cache/kube_cache.go
```diff
@@ -1467,12 +1467,27 @@ func (c *kubeCache) GetTelemetries(namespace, labelSelector string) ([]*v1alpha1
 }
 
 func (c *kubeCache) isK8sGatewayListerInit(namespace string) bool {
-	// K8s GW has several cache listers, those can be namespace or cluster based
-	// if one of them is not initialized, then all others are not init as well
-	// this can happen when CRDs are created after Kiali start
-	if c.getCacheLister(namespace).k8sgatewayLister == nil {
```
nrfox, line 1473: Looks like the issue is that `c.getCacheLister(namespace)` can return `nil`? Would probably be simpler to just add a conditional here then that checks for that before accessing `k8sgatewayLister`. Something like:
```
if cacheLister := c.getCacheLister(namespace); cacheLister == nil || cacheLister.k8sgatewayLister == nil {
  return false
}
```

nrfox (changes requested):

kubernetes/cache/kube_cache.go
hhovsepy: then it will always return false on OCP cluster and the Gateway functionality could be missed

hhovsepy: yeah agree, I had that idea on my previous PR, but avoided adding a new variable, now it sound better idea than many checks

hhovsepy:
@nrfox ready for a review

nrfox (approved):

jshaughn (commented):
Sorry for late review, but this approach looks good.

Issue #6478, opened 2023-08-14
Multicluster - Automated testing for wizards in multi-cluster scenarios


nrfox:
Test that you can create/edit/update objects in the "primary" cluster in multi-cluster mode. Ensure you can't edit istio config for "remote" clusters in multi-cluster mode.

matejnesuta:
The PR above doesn't test clusterrolebindings in this specific PR (#6060), because it needs more setup (if we are going to test for it in the first place).

nrfox:
Closing since #7045 has been merged

PR #6240, opened 2023-06-07
Remote cluster - Istio Configs Editable

hhovsepy:
RFE https://github.com/kiali/kiali/issues/6239

Remote clusters Istio Configs can be edited in 3 places now:
1. Config Details YAML editor. Edit and Delete.
2. Overview Page Namespace card. Create, Update, Delete traffic policies.
3. Service Details page Traffic Wizard. Crete, Update, Delete traffic.

Screenshot from 2023-06-09 19-33-00
Screenshot from 2023-06-09 19-32-43
Screenshot from 2023-06-09 19-32-12

TODO: Istio Config creation wizard is a separate epic.

PR #7045, opened 2024-01-22
Tests for Primary Remote and Multi-Primary related to the Istio objects creation

matejnesuta:
related to #6478 and #6240, as the gherkin steps will be pretty similar, even though the deployments are different


frontend/cypress/integration/common/traffic_policies_multicluster.ts
```diff
@@ -0,0 +1,57 @@
...
+  cy.get('button').contains(`${action[0].toUpperCase() + action.slice(1)} Traffic Policies`).should('be.visible');
+  cy.get('button').contains(`${action[0].toUpperCase() + action.slice(1)} Traffic Policies`).click();
+  ensureKialiFinishedLoading();
+});
+
+When('user confirms to {string} the Traffic Policy',(action:string) => {
+  if (action === 'create' || action === 'update'){
+    cy.get(`button[data-test="${action}"]`).click();
+  }
+  cy.get('button[data-test="confirm-create"]').click();
```
hhovsepy, line 37: is it hardcoded 'create' for update as well?

hhovsepy (approved):
LGTM, thanks for your work @matejnesuta

frontend/cypress/integration/common/traffic_policies_multicluster.ts
matejnesuta: Hello. Yes it is, but it only applies for this specific modal on the Overview page I think.

PR #7046, opened 2024-01-22
PF topology new version

leandroberetta:
Update to the new version of PF topology which adds the rectangular boxing.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -69,7 +69,7 @@
     "@patternfly/react-charts": "^7.1.0",
     "@patternfly/react-core": "^5.1.0",
     "@patternfly/react-table": "^5.1.0",
-    "@patternfly/react-topology": "^5.2.0-prerelease.3",
+    "@patternfly/react-topology": "^5.2.0",
     "axios": "^1.6.2",
     "bootstrap-slider-without-jquery": "10.0.0",
     "cy-node-html-label": "2.0.0",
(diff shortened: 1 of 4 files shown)


jshaughn (dismissed):
Rectangular boxing looks good. Label spacing looks better, not everything entirely within the boxes but label overlap is minimal in my test.

Box support is still really only OK in Dagre, the other layouts we use (grid, concentric and breadth first) don't seem to support boxing.  We should probably disable those for any boxed graph, or maybe add back cola, which is the only other one I know of that supports boxes in PFT.  Anyway, let's get this update installed.

jshaughn:
> What do yo mean by supported?

I mean they are rendered but I don't think they are actually taken into account by the layout.  That's why they overlap and look randomly placed.

jshaughn:
I'm also noticing that collapsing and expanding groups isn't always very optimal.

leandroberetta:
> > What do yo mean by supported?
> 
> I mean they are rendered but I don't think they are actually taken into account by the layout. That's why they overlap and look randomly placed.

I think both rectangular and hulled outlines are the same only that rectangular takes more space of boxing, but the overlap is the same. 

I've never used collapsible groups yet.

leandroberetta:
I created the following issue for the layouts problems: https://github.com/patternfly/react-topology/issues/131

jshaughn (approved):
Same changes as the previous PR, should be good!

Update:  I just noticed there are CI issues, I'm re-running but there is a good chance it's not a flake.  Probably a problem with the PF update affecting something.

Issue #6879, opened 2023-11-22
Make support of "ExportTo" feature of Istio config configurable


hhovsepy:
With Ambient mesh, the 'exportTo' of Istio configs will not be supported.
So Kiali should be able to ignore 'exportTo' in Ambient mode.
Meanwhile in a sidecar proxy mode the 'exportTo' support should stay.

PR #7047, opened 2024-01-22
Ignore ExportTo for Ambient mesh

hhovsepy:
Issue https://github.com/kiali/kiali/issues/6879

In a case of Ambient mesh, the `ExportTo` field of several Istio config objects are ignored, and the Object is validated only in own namespace scope. Affected configs: VirtualService, DestinationRule, ServiceEntry.

Added tests, but in fact made tests disabled because the `layer.IstioConfig` is not reset from test to test.


josunect (dismissed):
Code LG

hhovsepy:
@josunect have rebased this PR.

josunect (approved):

PR #7049, opened 2024-01-22
Fix clusters e2e test for operator deployments

nrfox:
** Describe the change **

Fix clusters test for operator deployments and cleanup e2e tests by using `kiali.Instance` to manage config.


jshaughn (approved):
Looks good, thanks for fixing this up.

PR #7056, opened 2024-01-24
Try to resolve cypress fails in sidecar_injection feature

jshaughn:
Hopefully Fixes OSSM 4460

Add page refresh to allow time for sidecar injection to happen and be reflected in the UI.  Note that this introduces a delay of a few seconds because the refresh button is temporarily hidden by a message center toast.

Also, rename one action to be more generic, for re-use.


nrfox (approved):

ScriptingShrimp:
note to myself: retest on OCP

Issue #7077, opened 2024-02-01
PFT Graph not handling graph background clicks


jshaughn:
PFT Graph not handling graph background clicks.  As such, a selected element stays selected, but what we want is to clear the selection and revert back to the full-graph side panel.

PR #7078, opened 2024-02-01
Add handling of graph [background] selection

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7077

Add handling of graph [background] selection to ensure we properly unselect any currently selected element.


leandroberetta (approved):
Changes looks good, tested locally and works.

PR #7080, opened 2024-02-02
Skip service validations for potential flaky remaining services

hhovsepy:
'tests.TestServicesList' is failing because of flaky 'TestK8sGatewaysAddressesError' not deleting auto-generated Service without 'app' label because of K8sGateway.

Skip validations for services besides 'productpage' to avoid potential flakes.

### DIFF
--- a/tests/integration/tests/services_test.go
+++ b/tests/integration/tests/services_test.go
@@ -24,12 +24,14 @@ func TestServicesList(t *testing.T) {
 	require.True(len(serviceList.Services) >= 4)
 	for _, service := range serviceList.Services {
 		require.NotEmpty(service.Name)
-		require.True(service.IstioSidecar)
-		require.True(service.AppLabel)
-		require.NotNil(service.Health)
-		require.NotNil(service.Health.Requests)
-		require.NotNil(service.Health.Requests.Outbound)
-		require.NotNil(service.Health.Requests.Inbound)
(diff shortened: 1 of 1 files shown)


tests/integration/tests/services_test.go
```diff
@@ -24,12 +24,14 @@ func TestServicesList(t *testing.T) {
...
-		require.NotNil(service.Health.Requests)
-		require.NotNil(service.Health.Requests.Outbound)
-		require.NotNil(service.Health.Requests.Inbound)
+		if service.Name == "productpage" {
+			require.True(service.IstioSidecar)
+			require.True(service.AppLabel)
+			require.NotNil(service.Health)
+			require.NotNil(service.Health.Requests)
+			require.NotNil(service.Health.Requests.Outbound)
+			require.NotNil(service.Health.Requests.Inbound)
```
ferhoyos, line 33:
```suggestion
			require.NotNil(service.Health.Requests.Inbound)
			break
```

It would be more performant with a break but the service list is not very large so it is non-blocking issue.

ferhoyos (approved):
I am ok reducing the scope to just one service we have under control to check the requires.

tests/integration/tests/services_test.go
hhovsepy: we still need to check other services at least the name

Issue #7076, opened 2024-02-01
Kiali Performance improvements.


hhovsepy:
This is a sub-epic to track Kiali performance improvements as a subtasks.

- [x] #4832 
- [x] https://github.com/kiali/kiali/pull/7084
- [x] https://github.com/kiali/kiali/issues/7017
- [x] Optimize fetching 'tls' in Overview page.
- [x] Optimize fetching 'metrics' in Overview page.
- [x] Optimize fetching 'health' in Overview page.

nrfox:
You can see that number of namespaces is a big issue when it comes to perf simply due to the amount of requests the frontend makes to the backend. This can be tested even on minikube by creating ~200 namespaces and then trying to load the graph or overview pages. Solving https://github.com/kiali/kiali/issues/4832 would make this much better. Adding it to the list here.

hhovsepy:
Kiali performance improvements are finished at this stage, in a scope of the epic.
More improvements could be done are separate tasks in the future.

jshaughn:
Nice job managing this epic @hhovsepy!

PR #7084, opened 2024-02-03
Optimize fetching validations on graph page

nrfox:
** Describe the change **

Fetches validations once per cluster instead of once per namespace.

** Issue reference **

ref: #7076 

Test with:
1. Creating ~200 namespaces with something like:
```bash
```
2. Visit the graph page.
3. Compare previous time it takes to load all the requests and the page with this PR.

Cleanup with:
```bash
for i in {1..200}; do kubectl delete ns test-$i --wait=false --ignore-not-found=true; done
```

In my environment with 208 namespaces selected:
Before:
211 requests ~10.2s

After:
4 requests ~1.8s

### DIFF
--- a/frontend/src/pages/Overview/OverviewPage.tsx
+++ b/frontend/src/pages/Overview/OverviewPage.tsx
@@ -536,13 +536,25 @@ export class OverviewPageComponent extends React.Component<OverviewProps, State>
   async fetchValidationResultForCluster(namespaces: NamespaceInfo[], cluster: string): Promise<void> {
     return Promise.all([API.getConfigValidations(cluster), API.getAllIstioConfigs([], [], false, '', '', cluster)])
       .then(results => {
+        const validations = results[0].data;
+        const istioConfig = results[1].data;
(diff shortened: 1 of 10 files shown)


nrfox:
This change is already covered by the `Graph page - See Istio config validations from both clusters` scenario. See an old commit: https://github.com/kiali/kiali/actions/runs/7763643512/job/21176115102 for an example.

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -536,13 +536,25 @@ export class OverviewPageComponent extends React.Component<OverviewProps, State>
   async fetchValidationResultForCluster(namespaces: NamespaceInfo[], cluster: string): Promise<void> {
     return Promise.all([API.getConfigValidations(cluster), API.getAllIstioConfigs([], [], false, '', '', cluster)])
```
hhovsepy, line 537: Here a namespaces param is required:
API.getConfigValidations(namespacesToString(namespaces), cluster)

hhovsepy:
Tested locally, only one fix to apply, with this fix all is working fine.

Issue #7088, opened 2024-02-05
Cypress testsuite for performance tests. All list and details pages including all table. Read only tests.


hhovsepy:
The following pages needs to be covered:
- [x] Apps list - selected namespaces, all namespaces
- [x] App Details - all tabs
- [x] Workloads list - selected namespaces, all namespaces
- [x] Workload Details - all tabs
- [x] Services list - selected namespaces, all namespaces
- [x] Service Details - all tabs
- [x] Istio Configs list  - selected namespaces, all namespaces
- [x] Istio Config Details

PR #7091, opened 2024-02-05
Added more list page tests for Apps, Workloads and Services.

hhovsepy:
Issue https://github.com/kiali/kiali/issues/7088

Added tests for

- [ ] Apps list - selected namespaces, all namespaces
- [ ] Workloads list - selected namespaces, all namespaces
- [ ] Services list - selected namespaces, all namespaces

More tests to come for Istio Config and Details pages.

### DIFF
--- a/frontend/cypress/perf/Graphload.spec.ts
+++ b/frontend/cypress/perf/Graphload.spec.ts
@@ -1,4 +1,4 @@
-import { reportFilePath } from './constants';
+import { reportFilePath } from './common';
 
 describe('Graph performance tests', () => {
   beforeEach(() => {
@@ -43,9 +43,7 @@ describe('Graph performance tests', () => {
 
               assert.isAtMost(duration, Cypress.env('threshold'));
 
-              const contents = `Graph load time for ${graphUrl}: ${(duration / 1000).toPrecision(5)} seconds
-
-  `;
+              const contents = `Graph load time for ${graphUrl}: ${(duration / 1000).toPrecision(5)} seconds\n`;
               cy.writeFile(reportFilePath, contents, { flag: 'a+' });
             });
         });
--- a/frontend/cypress/perf/common.ts
+++ b/frontend/cypress/perf/common.ts
@@ -1,0 +1,41 @@
+export const reportFilePath = 'logs/performance.txt';
+
+export const visits = 5;
+
+export function measureListsLoadTime(name: string, visits: number, listUrl: string): void {
+  // Getting an average to smooth out the results.
+  let sum = 0;
+  const visitsArray = Array.from({ length: visits });
+
+  cy.wrap(visitsArray)
+    .each(() => {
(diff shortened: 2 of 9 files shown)


ferhoyos:
@hhovsepy How performance tests are executed? I see that they have different configuration file `cypress-perf.config` but I don't find any yarn command to execute them.

hhovsepy:
For running without UI: 'make perf-tests-run', and for running with cypress UI: 'make perf-tests-gui', it's in 'frontend/cypress/README'.

ferhoyos:
Now I can execute them. Anyway, since all test commands are in package.json I think it would be good to have performance tests command too.

```
"cypress:perf": "cypress open --config-file cypress-perf.config.ts",
....
"cypress:run:perf": "cypress run --config-file cypress-perf.config.ts",
```

frontend/cypress/perf/common.ts
```diff
@@ -0,0 +1,41 @@
+export const reportFilePath = 'logs/performance.txt';
```
ferhoyos, line 1:
```suggestion
export const reportFilePath = 'cypress/results/performance.txt';
```
I would include performance results within cypress folder

frontend/cypress/perf/Graphload.spec.ts
ferhoyos, line 1: Rename to `graphload.spec.ts` (first letter in lowercase) to be consistent with the rest of the files

ferhoyos (commented):
Two comments:
- Not sure if it is my environment but I have an error with istioconfigloads.spec.ts

image

image

I can reproduce the issue in the Kiali application if I modify the namespaces from URL and refresh the page.

- I don't understand the difference between the namespace and allNamespaces (just bookinfo and istio-system) measurement. IMO what you should measure is the load time with different number of workloads, services and istio config items (similar like you do with overview perf test) to check load performance degradation according to number of items.

ferhoyos (approved):
LGTM. Good starting point to load times of Config, Apps, Workload and Services. For future PRs I think it would be great to measure load times for specific amounts of items (e.g., 50, 100, 150, etc.) to check limits of load times.

Another small comment. Now that we have perf commands in `package.json` we could simplify `make/Makefile.ui.mk` file:
```
## perf-tests-run: Runs the frontend perf tests locally without the GUI.
perf-tests-run:
	@cd ${ROOTDIR}/frontend && yarn cypress:run:perf --headless --config numTestsKeptInMemory=0,video=false

## perf-tests-gui: Runs the frontend perf tests locally with the GUI.
perf-tests-gui:
	@cd ${ROOTDIR}/frontend && yarn cypress:perf
```

hhovsepy:
@ferhoyos thanks for the review, the Makefile changes are included in https://github.com/kiali/kiali/pull/7103
The load time tests by the amount of items is planned in another subtask of the main epic.

Issue #4597, opened 2021-12-15
Add pprof endpoints for debugging perf issues


nrfox:
Taking inspiration from https://github.com/maxbischoff/kiali/commit/69865f05f7b5247e256b5dc82d145e6bbd708f1f, having pprof endpoints will help in diagnosing perf issues.

The debug handlers can be optionally added at runtime when a config option is enabled, something like `config.Server.Debug = true`, but disabled by default.

jmazzitelli:
I'm going to play with this today to see how easy and un-intrusive we can implement this.

jshaughn:
@hhovsepy , @jmazzitelli I added this to the perf Epic to tie in with our current priorities.

PR #7105, opened 2024-02-07
add profiler to kiali server

jmazzitelli:
To test, set Kiali CR `spec.server.profiler.enabled` to `true`. The server logs should then have a line like this at startup:

```
2024-02-07T18:09:40Z INF Profiler is enabled
```

Go to the Kiali URL, specifically the `<root context>/debug/pprof/` endpoint and you get the html index page of the profiler. Click all the links to see the data.

For example, if you port-forward to the Kiali server that is running on minikube (say, via `hack/kiali-port-forward.sh`), you should be able to see it at http://localhost:20001/kiali/debug/pprof/

Then use the "go tool pprof" utility to view it in a web browser. For example:

```
go tool pprof -http :8080 http://localhost:20001/kiali/debug/pprof/heap?debug=1
```

will bring up a browser tab pointing to the heap profiler. You can access the other profiles by just changing `heap` in the above URL to another profile name (such as `goroutine`, `allocs`, `profile` (which takes a while to load), et. al.).

fixes: https://github.com/kiali/kiali/issues/4597

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -159,6 +159,7 @@ type Server struct {
 	GzipEnabled                bool          `yaml:"gzip_enabled,omitempty"`
 	Observability              Observability `yaml:"observability,omitempty"`
 	Port                       int           `yaml:",omitempty"`
+	Profiler                   Profiler      `yaml:"profiler,omitempty"`
 	StaticContentRootDirectory string        `yaml:"static_content_root_directory,omitempty"`
 	WebFQDN                    string        `yaml:"web_fqdn,omitempty"`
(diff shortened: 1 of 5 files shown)


jmazzitelli:
Note: the /debug/pprof endpoint can be behind https (so it is encrypted) but it is NOT behind authentication... so if you have access to the Kiali URL, and the profiler is enabled, you can get the profiler data without needing to enter any credentials.

leandroberetta (approved):
Looks good.

Tried locally and works as expected:

image

PR #7108, opened 2024-02-07
put /debug/pprof behind auth

jmazzitelli:
To assuage those with security paranoia, this PR puts the pprom endpoint behind the Kiali auth. If Kiali is in anonymous mode, you can still get to pprom without credentials, but if Kiali has a non-anonymous auth-strategy set, then pprom will require credentials just as the Kiali console does.

### DIFF
--- a/routing/router.go
+++ b/routing/router.go
@@ -78,7 +78,63 @@ func NewRouter(conf *config.Config, kialiCache cache.KialiCache, clientFactory k
 	// Build our API server routes and install them.
 	apiRoutes := NewRoutes(conf, kialiCache, clientFactory, prom, traceClientLoader, cpm)
 	authenticationHandler, _ := handlers.NewAuthenticationHandler()
-	for _, route := range apiRoutes.Routes {
+
+	allRoutes := apiRoutes.Routes
+
+	// Add the Profiler handlers if enabled
+	if conf.Server.Profiler.Enabled {
+		log.Infof("Profiler is enabled")
+		allRoutes = append(allRoutes,
+			Route{
+				Method:        "GET",
+				Name:          "PProf Index",
+				Pattern:       "/debug/pprof/", // the ending slash is important
+				HandlerFunc:   hpprof.Index,
+				Authenticated: true,
+			},
+			Route{
+				Method:        "GET",
+				Name:          "PProf Cmdline",
+				Pattern:       "/debug/pprof/cmdline",
+				HandlerFunc:   hpprof.Cmdline,
+				Authenticated: true,
+			},
+			Route{
+				Method:        "GET",
+				Name:          "PProf Profile",
+				Pattern:       "/debug/pprof/profile",
+				HandlerFunc:   hpprof.Profile,
+				Authenticated: true,
+			},
+			Route{
+				Method:        "GET",
+				Name:          "PProf Symbol",
+				Pattern:       "/debug/pprof/symbol",
+				HandlerFunc:   hpprof.Symbol,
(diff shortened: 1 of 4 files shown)


jmazzitelli:
This worked when I set auth strategy to `token` and `anonymous`, but I can't get the unit test to pass.

jmazzitelli:
Note that accessing any profile (allocs, profile, goroutines, etc) with a request of more than 60 seconds will result in an error: `profile duration exceeds server's WriteTimeout`

So if you try something like this: `http://localhost:20001/kiali/debug/pprof/trace?debug=1&seconds=61` that error will result.

This is because we do not extend the server's writetimeout past 1 minute (the default is 30 seconds without the profiler enabled, but we extend it to 1 minute with profiler enabled - see server.server.go function NewServer where we do this:

```go
	writeTimeout := 30 * time.Second
	if conf.Server.Profiler.Enabled {
		writeTimeout = 1 * time.Minute
	}
```

If this is a problem in the future, we could make this configurable in the new spec.server.profiler section, something like:

```yaml
spec:
  server:
    profiler:
      enabled: true
      write_timeout: 300
```

where the write_timeout is the number of seconds we will set the server's writeTimeout setting.

For now, I'm leaving it hardcoded at 1 minute. We can add this capability later if we deem it necessary.

jmazzitelli:
To test this. First enable the profiler (`spec.server.profiler.enabled=true`).

Then turn on non-anonymous auth (either deploy to OpenShift or use `token` for `spec.auth.strategy` on minikube for the easiest ways to do it).

Then try to go to the `/kiali/debug/pprom/` endpoint (if on OpenShift, its just `/debug/pprom/`). You need to log in to see that now. That should be true for all the endpoints.

You could also just try to do a "curl" on all the different profile endpoints and see that curl reports the error "Unauthorized" when not using `anonymous`, but does return results when using `anonymous`.

For example, try these when anonymous and non-anonymous and see what you get (obviously, use the correct hostname/port/root path for your environment):

```
curl http://localhost:20001/kiali/debug/pprof/
curl http://localhost:20001/kiali/debug/pprof/allocs?debug=1
curl http://localhost:20001/kiali/debug/pprof/goroutine?debug=1
curl http://localhost:20001/kiali/debug/pprof/heap?debug=1
curl http://localhost:20001/kiali/debug/pprof/threadcreate?debug=1
curl http://localhost:20001/kiali/debug/pprof/block?debug=1
curl http://localhost:20001/kiali/debug/pprof/mutex?debug=1
curl http://localhost:20001/kiali/debug/pprof/allocs?debug=1
curl http://localhost:20001/kiali/debug/pprof/symbol?debug=1
curl http://localhost:20001/kiali/debug/pprof/trace?debug=1
curl http://localhost:20001/kiali/debug/pprof/profile?debug=1
```

jmazzitelli:
It's a real pain now trying to get the profiler data from the command line. I can't figure out the magic to `curl` when using `token` auth strategy. Need to do something like send a POST request with form data `token=<token>` (I think). But I'm not sure if you have to go through the `/api/authenticate` endpoint first, then get the kiali token, then pass that into the /debug/pprom endpoint

You can download the binary data the browser - pass in `?debug=0` and a file will be downloaded that you then pass to `go tool pprof` (e.g. `go tool pprof -http :8080 ~/Downloads/allocs`)

nrfox (approved):

jshaughn:
@jmazzitelli Should anything regarding to this stuff be included in a README or some sort of doco?

PR #7112, opened 2024-02-09
Eslint rule to use arrow functions

ferhoyos:
### Describe the change

This PR adds new precommit eslint rule to force use of arrow functions instead of function declarations in order to be consistent through Kiali application. 

This rule is not applied for those functions that cannot be converted to arrow functions: 
- generator methods (constructors)
- functions that contain `this` reference (e.g., some cypress functions).

Also, this eslint rule does not apply to function expressions since they are quite used in classes (e.g., static functions)

Besides, I have extended scope of eslint to whole `frontend` folder (except `node_modules` and `build` folders).

### Steps to test the PR

Modify a file that contains a function declaration (or add new one in any file) and try to commit the change. The error `"Function declarations are not allowed. Please use arrow functions instead"` should appear in the console log.

### Automation testing

Not needed


frontend/src/services/Api.ts
```diff
@@ -1100,6 +1100,7 @@ export function deleteServiceTrafficRouting(
 
 export function deleteServiceTrafficRouting(serviceDetail: ServiceDetailsInfo): Promise<ApiResponse<string>[]>;
 
+//eslint-disable-next-line no-restricted-syntax
```
ferhoyos, line 1103: Function overloading cannot be converted to arrow function, but I don't know how to add such use case in the eslint rule. Since there is only one case in Kiali application, eslint disable line is added.

nrfox:
It doesn't matter to me one way or another but why prefer the arrow function over a regular function? Seems like:

This is a function
```
function doSomething() {

}
```
And this is a variable whose value is a function
```
const doSomething = () => {

}
```

And you should use whichever one makes the most sense based on the context?

ferhoyos:
Yes, although they are not the same thing, in most cases both of them can be used. In my opinion arrow functions are often preferred for their readability and shorter syntax. I have added exceptions for those cases where functions are the only way (constructor methods, use of `this`, function expressions within classes). 

IMHO the code should be consistent through the application and define a set of rules where two or more options can be indistinctly used (e.g, trailing comma, indentation, etc.). This provides a cleaner code and better readability. I know that sometimes it is hard to define where is the limit of opinionated vs flexibility and people have different opinions about this. 

This is just a suggestion to use arrow functions over regular functions in those cases where both can be indistinctly used to make the code more consistent through the application. No problem if most of the team prefer the flexibility to use whatever kind of function consider in each case :-)

ferhoyos:
Closing the issue, maybe it is better to be flexible and let developers use what they need in each use case (although it is preferred to use arrow functions according to our style guide https://github.com/kiali/kiali/blob/master/STYLE_GUIDE.adoc#frontend-typescript)

Issue #4832, opened 2022-03-22
Cross-namespace endpoints


nrfox:
From discussion: https://github.com/kiali/kiali/discussions/4801

Kiali should provide cross-namespace endpoints for any resource that will need to be queried per namespace by the UI. For example:

- [x] Workloads: `/api/workloads`
- [x] Services: `/api/services`
- [x] Apps: `/api/apps`
- [x] Validations: `/api/validations`
- [x] Health: `/api/health`
- [x] Metrics: `/api/metrics`

Queries can be limited to a specific namespace by query params e.g. `/api/workloads/?namespaces=bookinfo,default,istio-system`. Here requested namespaces are specified as a comma separated list.

The `*` param will be interpreted as "all namespaces that the requesting user has access to". This is useful in cases where the user "selects all" namespaces on the front-end. For example, if 3 namespaces exist named `ns1`, `ns2`, and `ns3` and the user has `get` permission rights on `ns1` and `ns2` but not `ns3`, information for `ns1` and `ns2` will be returned by not `ns3`.

There could be thousands of namespaces, each with very long names, but URLs are limited in size so the UI will need to make multiple requests to fetch data for all the namespaces in these cases. It will still be cheaper than the cost of the UI making an individual request _per namespace_.

Optionally, regex support can be added for namespace filtering e.g. `/api/workloads/?namespaces="^development.*"`. This is more of a optimization that can be added later if support is added on the front-end in the namespace selector.

lucasponce:
> There could be thousands of namespaces, each with very long names, but URLs are limited in size so the UI will need to make multiple requests to fetch data for all the namespaces in these cases. It will still be cheaper than the cost of the UI making an individual request per namespace.

Well, for this situation when we have a large population of namespaces I think that the main pain point would be the design of the backend.

https://github.com/kiali/kiali/issues/4678

The endpoint design would help, but it won't have a real impact compared to enabling cluster-rights and taking that assumption on some circumstances (nothing strange, most of the Kubernetes admin consoles take this).

Multiplexing URLs are good, and I think handlers can be improved in that way, but the real work would be to use the "--all-namespace" semantic in the backend, which we couldn't use due to the cluster-less restriction.

To me is doable to point a user to give Kiali cluster rights under high namespaces population.

hhovsepy:
It seems this task is already done, in a scope of perf optmization epic.

PR #7113, opened 2024-02-09
Optimized Overview page mTLS requests per namespace into a single request

hhovsepy:
Issue https://github.com/kiali/kiali/issues/7076

Performance improvements for Overview page.
'mTLS' request now it send for a bunch of namespaces in a cluster instead of a request per namespace.
Tried for 200 namespaces, performance on my local is improved 2 times.

Still the delays are mostly caused by the 'health' requests per namespace. Will be fixed in separate PR.


frontend/cypress/results/performance.txt
ferhoyos, line 1: This should not be in the repo. Please add `frontend/cypress/results` path into `.gitignore` file.

hhovsepy: oops

business/tls.go
```diff
@@ -115,9 +114,62 @@ func (in *TLSService) NamespaceWidemTLSStatus(ctx context.Context, namespace, cl
 	return models.MTLSStatus{
 		Status:          mtlsStatus.NamespaceMtlsStatus(namespace).OverallStatus,
 		AutoMTLSEnabled: mtlsStatus.AutoMtlsEnabled,
+		Cluster:         cluster,
+		Namespace:       namespace,
 	}, nil
 }
 
+func (in *TLSService) ClusterWideNSmTLSStatus(ctx context.Context, nss []string, cluster string) ([]models.MTLSStatus, error) {
```
nrfox, line 122: Is this limited to the mTLS status of the passed in namespaces? If so then this might be better named something like `CrossNamespacemTLSStatus` since it's not scoped to the whole cluster just the provided namespaces.

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -470,9 +470,17 @@ export class OverviewPageComponent extends React.Component<OverviewProps, State>
...
 
   fetchTLS(isAscending: boolean, sortField: SortField<NamespaceInfo>): void {
-    _.chunk(this.state.namespaces, 10).forEach(chunk => {
+    const uniqueClusters = new Set<string>();
+
+    this.state.namespaces.forEach(namespace => {
+      if (namespace.cluster) {
+        uniqueClusters.add(namespace.cluster);
+      }
+    });
```
nrfox, line 479: Does this work in multi-cluster mode? Unfortunately depending on where the namespaces come from in `this.state.namespaces` sometimes you have the cluster in `namespace.cluster` and sometimes you don't. This is partially because the namespace names can come from the URL and we lose the cluster when that happens.

nrfox (commented):
Minor comments otherwise LGTM. Nice speedup!

frontend/src/pages/Overview/OverviewPage.tsx
hhovsepy: yes it works with multi/single clusters, also added a filter here and in validations request to send namespaces for that cluster only, otherwise all namespaces were sent comma separated.
Screenshot from 2024-02-12 12-17-36
Screenshot from 2024-02-12 12-17-16

routing/routes.go
```diff
@@ -1049,6 +1049,27 @@ func NewRoutes(conf *config.Config, kialiCache cache.KialiCache, clientFactory k
...
+		//      400: badRequestError
+		//      500: internalError
+		//
+		{
+			"ClusterTls",
+			"GET",
+			"/api/cluster/tls",
+			handlers.ClusterTls,
+			true,
+		},
```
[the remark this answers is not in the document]
hhovsepy: I will leave them separately not to complicate the logic, and the `/namespaces/tls` is not used in UI anymore, but we can leave it as REST API call for the community users.

business/tls.go
hhovsepy: 'CrossNamespacemTLSStatus' seems to me also a bit confusing, as when namespaces are not provided the method loads all namespaces from provided cluster by default. I will leave it as it is before some better idea will come. The idea of having `ClusterWide...` in names is having this task https://github.com/kiali/kiali/issues/4832 in a mind, to switch Namespace related fetches to Cluster related ones.

routing/routes.go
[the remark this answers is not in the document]
nrfox: I would remove it so as not to cause confusion later but that can be done separately. The API is considered internal only to kiali. If there's e2e tests that cover the `/namespaces/tls` endpoint we should get those moved over to `/cluster/tls`. I think we can do some cleanup in general of the API routes. There's already an `/api/clusters` endpoint and now were adding `/api/cluster`. We could even add more to `/api/cluster` as we continue to optimize the frontend calls. Because the api is considered internal only, it's not as much of a concern but it'd be nice to be consistent and clean things up at some point.

nrfox (commented):
Should we move this e2e test over to use the new route since the old namespaces route is no longer being used or else add a new test for cluster tls? https://github.com/kiali/kiali/blob/33a1fee6f090d11de1d3a8958f2f31f96dc9df85/tests/integration/tests/kiali_urls_test.go#L83-L92

routing/routes.go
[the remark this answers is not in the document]
hhovsepy: good point to have single `/clusters**`

nrfox (approved):

Issue #7042, opened 2024-01-18
Multicluster - Create a demo of a multi-cluster Kiali deployment on Openshift


nrfox:
### What do you want to improve?

Provide a demo of Kiali working on Openshift with the `openshift` auth strategy.

nrfox:
After some investigation, this is blocked by #6360 and #6037. Moving this back to backlog and out of beta.

nrfox:
This is blocked by #6360

nrfox:
This exists now at https://docs.redhat.com/en/documentation/red_hat_openshift_service_mesh/3.2/html/installing/ossm-multi-cluster-topologies#ossm-installing-kiali-multi-cluster-mesh_ossm-multi-cluster-topologies

PR #7114, opened 2024-02-09
Add hack script for setting up an "SSO" idp on openshift across two clusters

nrfox:
Note that this script is for testing only and demos/examples.

### Describe the change

Adds a hack script to setup an "SSO" idp across two openshift clusters for testing multi-cluster kiali setups on openshift.

### Steps to test the PR

1. Create two openshift clusters.
2. Login via CLI and rename your kube contexts to "east" and "west".
3. Run this script:
  `./hack/istio/multicluster/setup-keycloak.sh --manage-minikube false --manage-kind false --cluster1-context "east"  --cluster2-context "west"`
4. Navigate to the first cluster's console and login to the openid idp with the kiali user.
5. Navigate to the second cluster's console and the login should be skipped.

### Automation testing

N/A

### Issue reference

Related to #7042

### DIFF
--- a/hack/istio/multicluster/setup-keycloak.sh
+++ b/hack/istio/multicluster/setup-keycloak.sh
@@ -15,15 +15,22 @@ set -e
 CLUSTER1_OPENSHIFT_OAUTH_ROUTE=https://$(kubectl get routes --context "${CLUSTER1_CONTEXT}" -n openshift-authentication oauth-openshift -o jsonpath='{.spec.host}')
 CLUSTER2_OPENSHIFT_OAUTH_ROUTE=https://$(kubectl get routes --context "${CLUSTER2_CONTEXT}" -n openshift-authentication oauth-openshift -o jsonpath='{.spec.host}')
 
+# Need to create the namespace first so we can apply the openshift scc magic numbers.
+kubectl --context "${CLUSTER1_CONTEXT}" get ns keycloak || kubectl --context "${CLUSTER1_CONTEXT}" create ns keycloak
+SE_LINUX_LEVEL="$(kubectl --context "${CLUSTER1_CONTEXT}" get ns keycloak -o jsonpath='{.metadata.annotations.openshift\.io/sa\.scc\.mcs}')"
+
 # Postgres is required for keycloak and it's easiest to set it up separately.
-helm upgrade --kube-context "${CLUSTER1_CONTEXT}" --create-namespace --install --wait keycloak-postgresql oci://registry-1.docker.io/bitnamicharts/postgresql -n keycloak --reuse-values --values - <<EOF
+echo "Creating postgres deployment"
+helm upgrade --kube-context "${CLUSTER1_CONTEXT}" --install --wait \
+  --repo https://charts.bitnami.com/bitnami postgresql postgresql \
+  -n keycloak --reuse-values --values - <<EOF
 primary:
   persistence:
     enabled: false
   containerSecurityContext:
     runAsUser: null
     seLinuxOptions:
-      level: "s0:c26,c15"
+      level: "${SE_LINUX_LEVEL}"
   podSecurityContext:
     fsGroup: null 
 auth:
@@ -35,8 +42,9 @@ EOF
 APPS_DOMAIN=$(echo "${CLUSTER1_OPENSHIFT_OAUTH_ROUTE}" | cut -d '.' -f2-)
 KEYCLOAK_HOSTNAME="keycloak-keycloak.${APPS_DOMAIN}"
 
+echo "Creating keycloak deployment"
 helm upgrade --kube-context "${CLUSTER1_CONTEXT}" --install --wait --timeout 15m \
-  --namespace keycloak --create-namespace \
+  --namespace keycloak \
   --repo https://charts.bitnami.com/bitnami keycloak keycloak \
   --reuse-values --values - <<EOF
 auth:
@@ -50,12 +58,12 @@ image:
 proxyAddressForwarding: true
 containerSecurityContext:
   seLinuxOptions:
-    level: s0:c26,c15
+    level: "${SE_LINUX_LEVEL}"
   runAsUser: null
 podSecurityContext:
   fsGroup: null
 externalDatabase:
-  host: keycloak-postgresql
+  host: postgresql
   database: keycloak
   user: keycloak
   password: "${KEYCLOAK_DB_PASSWORD}"
@@ -64,6 +72,7 @@ postgresql:
 EOF
 
 # Create keycloak route.
+echo "Creating keycloak route"
 kubectl --context "${CLUSTER1_CONTEXT}" apply -f - <<EOF
 apiVersion: route.openshift.io/v1
 kind: Route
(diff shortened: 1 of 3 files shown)


jmazzitelli:
> Login via CLI and rename your kube contexts to "east" and "west".

How do you do that? I have never been able to get kubeconfig contexts to keep working when switching between OpenShift clusters. I always have to do a "oc login" when switching (e.g. I can't do "oc --context XYZ" if I'm not already logged into XYZ - otherwise, I get an auth error).

nrfox:
> How do you do that? I have never been able to get kubeconfig contexts to keep working when switching between OpenShift clusters. I always have to do a "oc login" when switching (e.g. I can't do "oc --context XYZ" if I'm not already logged into XYZ - otherwise, I get an auth error).

1. `oc login <url> -u <user> -p <pass>`
2. `kubectl config get-contexts`
3. `kubectl config rename-context <long-east-cluster-url-for-admin-user> east`

There might be a one liner to provide the context name when logging in with `oc login` but I couldn't find one.

jmazzitelli:
> There might be a one liner to provide the context name when logging in with `oc login` but I couldn't find one.

I did a quick look just to get this clear in my head. I think I get it now. You do need to login using the normal `oc login` command we are used to (you don't use `--context`). But once you log in, apparently the token stays valid for at least some amount of time even if you log into another OpenShift Cluster. So, for example, I got this to work. In short, I `oc login` to cluster 1 first, then log into cluster 2, then I use `--context` to switch back to cluster 1 when I need to `oc get` something from cluster 1.

```
# login to cluster 1
oc login --server https://api.dbtest.lab.xxx.com:6443 -u kiali -p kiali

# login to cluster 2
oc login --server https://api.istiotest.lab.xxx.com:6443 -u kiali -p kiali

# need to switch to cluster 1 using --context to get its namespaces
oc --context='/api-dbtest-lab-xxx-com:6443/kiali' get ns

# without --context, get info from the last cluster I logged into (cluster 2)
oc get ns

# you can still use --context to get info from the last cluster I logged into
 oc --context='/api-istiotest-lab-xxx-com:6443/kiali' get ns
```

jmazzitelli (changes requested):
This isn't working for me. When keycloak statefulset is installed, it fails to start:

`oc --context='/api-dbtest-xxx-com:6443/kiali' -n keycloak describe statefulset/keycloak-postgresql`

shows this event happening:

`Warning  FailedCreate  3m26s (x17 over 7m32s)  statefulset-controller  create Pod keycloak-postgresql-0 in StatefulSet keycloak-postgresql failed error: pods "keycloak-postgresql-0" is forbidden: unable to validate against any security context constraint: [provider "anyuid": Forbidden: not usable by user or serviceaccount, provider restricted-v2: .containers[0].seLinuxOptions.level: Invalid value: "s0:c26,c15": must be s0:c26,c20, provider "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2": Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden: not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden: not usable by user or serviceaccount, provider "machine-api-termination-handler": Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2": Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden: not usable by user or serviceaccount, provider "hostaccess": Forbidden: not usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable by user or serviceaccount, provider "privileged": Forbidden: not usable by user or serviceaccount]`

I am on OpenShift 4.13.27.

jmazzitelli:
This is getting me past that one error:

```diff
diff --git a/hack/istio/multicluster/setup-keycloak.sh b/hack/istio/multicluster/setup-keycloak.sh
index 8a28dddec..cd40ee2e2 100755
--- a/hack/istio/multicluster/setup-keycloak.sh
+++ b/hack/istio/multicluster/setup-keycloak.sh
@@ -23,7 +23,7 @@ primary:
   containerSecurityContext:
     runAsUser: null
     seLinuxOptions:
-      level: "s0:c26,c15"
+      level: "s0:c26,c20"
   podSecurityContext:
     fsGroup: null 
 auth:
@@ -50,7 +50,7 @@ image:
 proxyAddressForwarding: true
 containerSecurityContext:
   seLinuxOptions:
-    level: s0:c26,c15
+    level: s0:c26,c20
   runAsUser: null
 podSecurityContext:
   fsGroup: null
```

jmazzitelli:
This script now finishes successfully after that change, however, I'm still not seeing the OID login... I still only see my two normal login providers:
image

jmazzitelli:
OK, I see the `openid` button in the OS login screen. I click it, enter the username `kiali` with password `kiali` and I get this:

image

In the keycloak logs I see this:

`2024-02-13 13:50:20,040 WARN [org.keycloak.events] (executor-thread-0) type=LOGIN_ERROR, realmId=e0ac20c5-93f2-4e17-9372-174c837a2090, clientId=kube, userId=null, ipAddress=10.0.91.81, error=user_not_found, auth_method=openid-connect, grant_type=password, client_auth_method=client-secret, username=test`

hack/istio/multicluster/setup-keycloak.sh
```diff
@@ -0,0 +1,197 @@
...
+kubectl --context "${CLUSTER2_CONTEXT}" create configmap keycloak-oidc-client-ca-cert --from-file=ca.crt="${INGRESS_ROUTER_CA_FILE}" -n openshift-config
+
+function update_cluster_idp {
+  local cluster_context=$1
+  # Patch the OAuth config with the keycloak idp for the west cluster.
+  if [ -z "$(kubectl --context "${cluster_context}" get oauths cluster -o jsonpath='{.spec.identityProviders[?(@.name == "openid")]}')" ]; then
+    echo "openid provider doesn't exist on the cluster's oauth config. Adding it..."
+    # Need to check if the spec.identityProviders exists and if it doesn't then ensure it does
+    # before attempting to add an item to it with the patch below otherwise the patch will fail.
+    if [ -z "$(kubectl --context "${cluster_context}" get oauths cluster -o jsonpath='{.spec.identityProviders}' --context west)" ]; then
```
jmazzitelli, line 151: I'm getting this error spit out a couple times (note, my context name is NOT "west"):
```
Error in configuration: context was not found for specified context: west
```

I think its due to `west` context name being hardcoded when it should be the context name passed in:

```suggestion
    if [ -z "$(kubectl --context "${cluster_context}" get oauths cluster -o jsonpath='{.spec.identityProviders}' --context ${cluster_context})" ]; then
```

jmazzitelli (changes requested):

hack/istio/multicluster/setup-keycloak.sh
```diff
@@ -0,0 +1,197 @@
...
+kind: Secret
+metadata:
+  name: openid-client-secret
+  namespace: openshift-config
+type: Opaque
+EOF
+
+KEYCLOAK_ISSUER_URI="https://keycloak-keycloak.${APPS_DOMAIN}"
+
+# Create the configmap with the CA in it for the west cluster.
```
jmazzitelli, line 141: probably should change "west" in this comment to something like "context #2"

hack/istio/multicluster/setup-keycloak.sh
jmazzitelli: wait a minute, I just noticed that `--context` is passed in twice - the last one "winning". Just remove that last one:
```suggestion
    if [ -z "$(kubectl --context "${cluster_context}" get oauths cluster -o jsonpath='{.spec.identityProviders}')" ]; then
```

jmazzitelli:
FWIW: it didn't work, my identityProvider lost the array item that was there originally and was overridden again. I suspect it might have to do with the error that I just reported in my latest review.

jmazzitelli:
and fwiw: we really should have a "delete" command or option in this script - I don't know how you test this when things fail - because you then have to tear down and restart both clusters (unless you know the magic to remove everything this script did). It would be nice to have an idempotent "delete" option so it can clean up anything and everywhere no matter where in the process it might have failed.

jmazzitelli:
I was able to run the script again all the way through without errors by:
(a) removing that second `--context west` option that I mentioned in my last review 
(b) oc --context <context 2> delete ns keycloak
(c) oc --context <context 2> delete cm keycloak-oidc-client-ca-cert -n openshift-config

Not sure what else a "delete" option would entail, but if it simply does (b) and (c), that looks like most of it (looks like there might be others things like deleting openid-client-secret from openshift-config namespace.

jmazzitelli:
I've waited several minutes (it needs that long for the new identity provider to deploy), but cluster 1 still isn't ready, though cluster 2 is. Not sure if it is because I didn't clean things up properly in my attempt to re-try again after the initial failure. So I guess there might be more things that are needed to completely revert what the script did. But here's what I see:

cluster 2:
image

cluster 1:
image

jmazzitelli:
BTW: it is very easy to do what I'm testing. Just stand up two OpenShift clusters, then log into one and run "make cluster-add-users" - you now have an htpasswd provider with "kiali" user (password "kiali"). Log into the second cluster, and run the same make command to get htpasswd provider over in that cluster.

Now run the script in this PR.

jmazzitelli:
I'm standing up 2 new clean openshift clusters and I'll try again (with removing that "--cluster west" in my local script) and I'll see if it works. If it still doesn't work, then we can ignore this and not worry about trying to support multiple identityProviders - maybe there is more to this than I realize and it isn't as simple as I thought to support both htpasswd and openid at the same time.

jmazzitelli:
with this PR, I got things working: https://github.com/nrfox/kiali/pull/4

jmazzitelli (dismissed):
this all works - install and delete

jmazzitelli (approved):
LGTM

Issue #7127, opened 2024-02-19
Warning in workload 'istio-ingressgateway' in non control-plane namespace


hhovsepy:
### Describe the bug

`istio-ingressgateway` created in different than control-plane namespace has a warning in proxy status check.

![Screenshot from 2024-02-19 07-48-14](https://github.com/kiali/kiali/assets/604313/6547737a-42ab-4ea7-ad66-c088b20fb236)

This happens because when `istio-ingressgateway` is deployed in different namespace that 'istio-system', in this case it is istio-ingress, deployed according to https://docs.openshift.com/container-platform/4.14/service_mesh/v2x/ossm-traffic-manage.html#ossm-automatic-gateway-injection_traffic-management, for automatic geteway injection the Deployment contains [sidecar.istio.io/inject=](http://sidecar.istio.io/inject=)"true"  as required in the doc. But in that case, when sidecar is injected into istio-ingressgateway pod, Kiali starts checking proxy status for that pod and shows warning on workload RDS: NOT_SENT, which is fine, just that warning is not shown for istio-system  gateway istio-ingressgateway generated by SMCP because the sidecars are not injected on that pod.

### Expected Behavior
For those Workloads and Pods which are Gateway, Kiali should skip the proxy status check.

### What are the steps to reproduce this bug?
(https://docs.openshift.com/container-platform/4.14/service_mesh/v2x/ossm-traffic-manage.html#ossm-automatic-gateway-injection_traffic-management)

### Environment
Happens on Kiali upstream and OSSM.
Should be backported to Kiali v1.73.

PR #7128, opened 2024-02-19
Skip Gateway workload pods proxy status check.

hhovsepy:
### Describe the change

Skip proxy status checks for Workloads which are Gateways.

### Steps to test the PR

Steps to deploy a Gateway 'istio-ingressgateway' into namespace different than 'istio-system' https://docs.openshift.com/container-platform/4.14/service_mesh/v2x/ossm-traffic-manage.html#ossm-automatic-gateway-injection_traffic-management
Here for this Workload it was showing a Proxy Status warning message, but it fact it should not. For Gateways in 'istio-system' it used to ignore the check.
With this PR it will not show Proxy check warning message on Gateways at all.

Regression test the 'istio-ingressgateway' in 'istio-system' namespace and other Worklaods from 'bookinfo' namespace with sidecars.

### Automation testing

Covered in unit and integration tests.

### Issue reference

Issue https://github.com/kiali/kiali/issues/7127

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -1161,7 +1161,7 @@ func (in *WorkloadService) fetchWorkloadsFromCluster(ctx context.Context, cluste
 
 		// Add the Proxy Status to the workload
 		for _, pod := range w.Pods {
-			if pod.HasIstioSidecar() && config.Get().ExternalServices.Istio.IstioAPIEnabled {
+			if pod.HasIstioSidecar() && !w.IsGateway() && config.Get().ExternalServices.Istio.IstioAPIEnabled {
 				pod.ProxyStatus = in.businessLayer.ProxyStatus.GetPodProxyStatus(cluster, namespace, pod.Name)
 			}
 		}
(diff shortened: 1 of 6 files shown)


hhovsepy:
Ready for review, the failing and commented out test assertion is not related and happens on master running in CI https://github.com/kiali/kiali/pull/7135

business/workloads.go
```diff
@@ -1161,7 +1161,7 @@ func (in *WorkloadService) fetchWorkloadsFromCluster(ctx context.Context, cluste
 
 		// Add the Proxy Status to the workload
 		for _, pod := range w.Pods {
-			if pod.HasIstioSidecar() && config.Get().ExternalServices.Istio.IstioAPIEnabled {
+			if pod.HasIstioSidecar() && !w.IsGateway() && config.Get().ExternalServices.Istio.IstioAPIEnabled {
```
jshaughn, line 1164: To be honest, I don't know why `pod.HasIstioSidecar()` is true for a Gateway workload, because a Gateway doesn't really have a sidecar, it is itself a proxy.  If `pod.HasIstioSidecar()` returned false for this case we would not need to remember to add this additional test in various places.

jshaughn (approved):
This seems to do what we need, although see my non-blocking comment.

business/workloads.go
hhovsepy: `pod.HasIstioSidecar()` is true for the Gateway deployed outside of the 'istio-system' namespace and has 'sidecar.istio.io/inject: "true"'.

Issue #7117, opened 2024-02-12
[mesh page] Initial Find/Hide support


jshaughn:
The POC has Find/Hide in place but it is not working.  Make sure it has a decent set of initial operands and is working as expected.

jshaughn:
Closed by https://github.com/kiali/kiali/pull/7130

PR #7130, opened 2024-02-19
Fix issues in Mesh Find/Hide

jshaughn:
Fix issues in Mesh Find/Hide
    - fix some type casting
    - fix some redux decls
    - improve debug logging

https://github.com/kiali/kiali/issues/7117

### DIFF
--- a/frontend/src/pages/Graph/GraphHelpFind.tsx
+++ b/frontend/src/pages/Graph/GraphHelpFind.tsx
@@ -12,22 +12,24 @@ export interface GraphHelpFindProps {
   className?: string;
   onClose: () => void;
 }
-
-const width = '600px';
-const maxWidth = '604px';
-const contentWidth = '540px';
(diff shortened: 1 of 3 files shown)


ferhoyos (changes requested):
The PR fixes the issue where find and hide value selectors were shared between `Traffic` and `Mesh` graphs.

I have seen two more possible issues wrt mesh find/hide support:
- All mesh graph nodes are considered not healthy (they are highlighted if I enter find value `! healthy`) 
image

- Find/Hide help info looks weird if there is not enough information to show:
image

That's probably because the height of the popover is fixed.

jshaughn:
@ferhoyos I will fix the help dialog issue, good catch.  As for the health F/H, let's defer that to be part of the epic sub-task for health support.

ferhoyos (approved):
LGTM

Issue #7118, opened 2024-02-12
[mesh page] improve non-infra-namespaces handling


jshaughn:
In the POC there is something wrong with the collapsible box handling for the non-infra-namespaces (i.e. data-plane) namespaces.  This is to investigate and improve handling.

PR #7131, opened 2024-02-19
Fix the dataplane namespaces box

jshaughn:
The namespace box was not properly attached to its parent cluster box.
- change "non-infra" terminology to "data planes"
- simplify boxing strategy
- change node shapes to be more distinct from the traffic graph
- increase node sizes for easier visibility and more distinction from the traffic graph
- add dataplane namespaces to test mesh

Closes https://github.com/kiali/kiali/issues/7118

### DIFF
--- a/frontend/src/pages/Mesh/Mesh.tsx
+++ b/frontend/src/pages/Mesh/Mesh.tsx
@@ -248,7 +248,6 @@ const TopologyContent: React.FC<{
 
       function addNode(data: NodeData): NodeModel {
         data.onHover = onHover;
-        console.log(`infraType=${data.infraType}`);
         const size = data.infraType === MeshInfraType.NAMESPACE ? NAMESPACE_NODE_SIZE : DEFAULT_NODE_SIZE;
         const node: NodeModel = {
           data: data,
--- a/frontend/src/pages/Mesh/Mesh.tsx
+++ b/frontend/src/pages/Mesh/Mesh.tsx
@@ -37,7 +37,7 @@ import { TourStop } from 'components/Tour/TourStop';
 import { getFocusSelector, unsetFocusSelector } from 'utils/SearchParamUtils';
 import { meshComponentFactory } from './components/meshComponentFactory';
 import { MeshData } from './MeshPage';
-import { MeshNodeData, MeshTarget } from 'types/Mesh';
+import { MeshInfraType, MeshTarget } from 'types/Mesh';
 import { MeshHighlighter } from './MeshHighlighter';
 import {
   EdgeData,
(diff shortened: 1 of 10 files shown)


jshaughn:
@ferhoyos With this PR you should now see the non-infra namespace box properly inside the cluster box. Also, moved to "dataplane" terminology.

frontend/src/pages/Mesh/Mesh.tsx
```diff
@@ -247,15 +248,17 @@ const TopologyContent: React.FC<{
 
       function addNode(data: NodeData): NodeModel {
         data.onHover = onHover;
+        console.log(`infraType=${data.infraType}`);
```
ferhoyos, line 251: `console.log` is not recommended. If this message is for debugging purposes, please change it to `console.debug`

ferhoyos (changes requested):
The PR works fine:
image

I like the hexagon shape to differentiate from traffic graph nodes, but I see a problem with node size. Labels are not shown on first sight (you have to zoom in to see them), so you don't know cluster, namespace or node names (only Kiali and Istio nodes have logo images). You have to click or zoom in to figure out what is in your mesh.

To solve this I would set `DEFAULT_NODE_SIZE` value a bit smaller (e.g., 100) or if you want to maintain that size it is needed to add icons for the rest of the nodes (grafana, jaeger and prometheus). Then user will know what is in the mesh on a first sight.

Another small issue is that Kiali icon is not centered in the node. You can fix this by adding some offset to KialiLogo (`MeshLegendData.tsx`:
```
  yOffset: 150,
  xOffset: -100
```

Out of scope of this PR, if I change to `kiali-cola` layout and refresh the page, the layout is set to default `kiali-dagre` layout although the url has suffix `&meshLayout=kiali-cola`

frontend/src/pages/Mesh/Mesh.tsx
jshaughn: That shouldn't be there at all, I'll remove it.  It's already gone in my next PR...

jshaughn:
> I like the hexagon shape to differentiate from traffic graph nodes, but I see a problem with node size. Labels are not shown on first sight (you have to zoom in to see them), so you don't know cluster, namespace or node names (only Kiali and Istio nodes have logo images). You have to click or zoom in to figure out what is in your mesh.

I'm glad you like the Hexagon.  There is a lot more to do on the visualization, and initial improvements are the focus of the next PR (https://github.com/kiali/kiali/pull/7133).  This one is really just to fix the boxing.

> To solve this I would set DEFAULT_NODE_SIZE value a bit smaller (e.g., 100) or if you want to maintain that size it is needed to add icons for the rest of the nodes (grafana, jaeger and prometheus). Then user will know what is in the mesh on a first sight.

+1, I want to have background logos for all of the Infra nodes.  It's just that they are a pain to create (or I don't know the easy way).

> Another small issue is that Kiali icon is not centered in the node. You can fix this by adding some offset to KialiLogo (MeshLegendData.tsx:

+1, already fixed in the next PR

> Out of scope of this PR, if I change to kiali-cola layout and refresh the page, the layout is set to default kiali-dagre layout although the url has suffix &meshLayout=kiali-cola

The PFT layouts are horrible.  So (have I said this enough :-)), in the next PR I have started on a custom layout.

ferhoyos (approved):
Since all my comments are going to be addressed in following PRs, I am ok with this one :-)

Issue #7120, opened 2024-02-13
[mesh page] Initial improvements to the graph visualization


jshaughn:
This is a catch-all issue for immediate UX improvements to the graph itself, up to the point where we will address other visualization concerns on a case-by-case basis.   It can include things like and :
- layout
- node shape, size, colors, backgrounds, etc
- edge styles, etc
- anything else that impact the visuals or ability to manipulate or interpret the graph itself.

PR #7133, opened 2024-02-19
Work on the Mesh graph usability

jshaughn:
Closes https://github.com/kiali/kiali/issues/7120

- First pass at dedicated custom layout (MeshLayout)
  - basically a copy of ColaGroupsLayout that lays out the non-group nodes in an alphabetical grid.
- Update node sizes and complete initial set of logos
- Fix issue with unwantd re-layout on updates
- Improve side panel header sections


jshaughn:
@ferhoyos , this PR was mainly to provide initial, general improvements in the page.  It primarily introduces the custom layout.  After this PR we will move on more targeted tasks.  So, for review of this one I would suggest that you just look for any major issues in the code or function, because we will follow up with specific issues.

ferhoyos (approved):
Good job @jshaughn, I really like the new mesh layout you have created:

image

Since this mesh topology is not going to be very dynamic (as opposed to traffic graph), I agree to have only one possible layout.

I don't see any major code issue, approving the PR.

PR #7136, opened 2024-02-19
Create controller framework and move validations to offline processing

nrfox:
### Describe the change

Moves calculating namespace/cluster-wide validations into a "controller" for offline processing. Individual object validations are still done "on demand" within the request. Calculates validations every 30s. Does not yet watch istio objects to re-calculate validations when they change.

### Steps to test the PR

1. Visit istio config page and ensure validations are present (configuration is green/yellow/red).
2. Update an individual istio object to be invalid.
3. Return to list page and ensure that validations on the list page match what was reported on the individual page.

### Automation testing

Covered by newly added "controller" tests.

### Issue reference

Relates to #7076


frontend/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "proxy": "http://localhost:20001",
```
hhovsepy, line 2: remove this

hhovsepy:
This PR becomes very important for Perf optimization effort.

hhovsepy:
The failing 2 validation tests are because of cache is not reloaded yet, just require timeout and refresh the page.

hhovsepy:
I think we can close this PR, as it is rebased and merged in another PR.

PR #7137, opened 2024-02-19
[hack] skupper hack script: be able to install a specific kiali version

jmazzitelli:
Pass in `-kv` (aka `--kiali-version`) with the version you want to install (e.g. `v1.79.0`). If you pass in `dev`, a local build will be done and the dev build pushed into your cluster and installed. `dev` is the default.


jshaughn (approved):
Thanks!

PR #7144, opened 2024-02-21
Move `AmbientEnabled` to the `KialiCache`

nrfox:
### Describe the change

Moves `AmbientEnabled` check to the `KialiCache` and make the check cluster specific. This gets rid of the global variables that were breaking the unit tests.

Split out from #7136

### DIFF
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -15,6 +15,8 @@ import (
 	"github.com/kiali/kiali/store"
 )
 
+const ambientCheckExpirationTime = 1 * time.Minute
+
 // Istio uses caches for pods and controllers.
 // Kiali will use caches for specific namespaces and types
 // https://github.com/istio/istio/blob/master/mixer/adapter/kubernetesenv/cache.go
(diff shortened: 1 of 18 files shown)


kubernetes/cache/cache.go
```diff
@@ -151,5 +161,39 @@ func (c *kialiCacheImpl) SetClusters(clusters []kubernetes.Cluster) {
...
 }
 
+// Check if istio Ambient profile was enabled
+// ATM it is defined in the istio-cni-config configmap
+func (in *kialiCacheImpl) IsAmbientEnabled(cluster string) bool {
+	currentTime := time.Now()
+	if in.ambientLastUpdateTime == nil {
+		in.ambientLastUpdateTime = new(time.Time)
+		in.ambientLastUpdateTime = &currentTime
+	}
```
nrfox, line 171: Just realized that these need to be cluster specific now

josunect (dismissed):
Tested (In single cluster) and looks good: 

image

Code LGTM

nrfox:
@josunect I made this both more complicated and more simple. I added an expiring cache which is generally useful and we can reuse it elsewhere. It made the `IsAmbientEnabled` implementation a bit simpler.

kubernetes/cache/cache_test.go
```diff
@@ -64,3 +65,59 @@ func TestKubeCacheCreatedPerClient(t *testing.T) {
 	_, err = kialiCache.GetKubeCache("cluster3")
 	require.Error(err)
 }
+
+func TestIsAmbientEnabled(t *testing.T) {
```
josunect, line 69: What about adding another test to check that Ambient is not enabled?

josunect (commented):
Tested and looks good (In single & multi cluster). 
image


I guess that should apply this changes before merge? 
https://github.com/kiali/kiali/pull/7162

nrfox:
> I guess that should apply this changes before merge? #7162

Yeah let's get that one in first and I'll rebase.

josunect (approved):
LGTM!

Tested in single cluster.

Issue #7147, opened 2024-02-21
The duration label is confusing in the Overview control plane charts


jshaughn:
Usability bug, people find the duration confusing in the overview control plane charts.  In the card below, the "10m".

![image](https://github.com/kiali/kiali/assets/2104052/3aafa72a-ba35-4dd0-9232-88ba6267e2de)

This can be removed as the duration is reflected in the global duration component, and also in the x axis.

PR #7148, opened 2024-02-21
Remove duration labeling for Overview control plan charts

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7147

Remove duration labeling for Overview control plan charts
- move info icon up when space allows
- improve info icon tooltip text

Now looks like:

image


jmazzitelli (approved):
much better, less confusing

Issue #7153, opened 2024-02-27
[graph] ServiceEntry ExportTo is not handled correctly


jshaughn:
This is a follow-on problem to #6962.  In that issue we solved a problem with ServiceEntry handling in the way it retrieved SE definitions.  This has revealed a second issue, where we have a problem determining which ServiceEntries are applicable to which namespaces via ExportTo.  See the post-issue-close comments in the previous issue for more information.

jshaughn:
@im-Amitto, I *think* this will finally resolve your service entry rendering issue.  If you would like to test this fix before the next Kiali release the change will be in or next nightly dev build, published on quay.  See the `latest` tag at https://quay.io/repository/kiali/kiali?tab=tags.  It should be available around 12 hours from now...  

Otherwise, the change will be in the forthcoming v1.81 release, but if you can test before the release it would be great.

im-Amitto:
@jshaughn This has finally fixed the issue, Thanks!!

Just a question, shouldn't we graph a SE as single entity if it is exported to multiple namespaces.

For ex: Coralogix should have been better graphed as 1 entity instead of two.
[LOGAF=low]

<img width="543" alt="Screenshot 2024-03-05 at 6 29 01 PM" src="https://github.com/kiali/kiali/assets/29653282/711fcb6f-afff-46b6-a16d-84b2418967e8">

jshaughn:
> @jshaughn This has finally fixed the issue, Thanks!!

Hooray! Thanks for testing with the new release.

> Just a question, shouldn't we graph a SE as single entity if it is exported to multiple namespaces.

This has been asked for before, and if you'd like you can create an Enhancement request issue.  I guess what you would like to see is one SE node in the namespace where the SE is defined?   We handle it today with the idea that exporting the definition is like a copy of the definition to each exported namespace.  I think different users would prefer to see it different ways, so it would likely need to be an Display menu option, if offered.

PR #7154, opened 2024-02-27
[graph] Fix serviceEntry handling when the definition is exported (not local) to the source node namespace

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7153

Yet another pass as ServiceEntry handling
- Update approach to ensure that we consider ALL of the service entry definitions that are local to, or exported to, the cluster+namespace of the source node requesting the serviceEntry host.
  - And also all of those serviceEntry's hosts
- Remove duplicates when determining "se-service" candidates
  - replace slice with map
- Revise service_entry appender comments, for clarity
- Add test specific to the bug's use-case

### DIFF
--- a/graph/telemetry/istio/appender/service_entry.go
+++ b/graph/telemetry/istio/appender/service_entry.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


jmazzitelli (approved):
i didn't run this. but the unit tests pass.

Issue #7071, opened 2024-02-01
envoy access log entry doc links are broken


jmazzitelli:
### Describe the bug

Links in the envoy logs dialog window are broken.

### Expected Behavior

Links to the actual docs.

### What are the steps to reproduce this bug?

1. Install some mesh apps
2. Navigate to a workload in the mesh that has a sidecar
3. Go to the "Logs" tab
4. Click on one of the (i) icons on one of the envoy log messages
5. In the popup dialog, click "response flags"
6. Look at the right side and click on the doc links - they are all broken. (e.g. I'm using port-forwarding and all the links have localhost:20001 as host: `http://localhost:20001/docs/envoy/latest/intro/arch_overview/upstream/circuit_breaking#arch-overview-circuit-break`)

![image](https://github.com/kiali/kiali/assets/2029470/4af9d25f-fa6b-4aa1-a04a-6631a43f2629)

![image](https://github.com/kiali/kiali/assets/2029470/508538ab-61b1-49f7-a045-747ba9866e82)

### Environment

minikube

- **Kiali version:**

master build

- **Istio version:**

1.20.2

jshaughn:
Broken links are bad, if we can't easily fix them we should try and just remove the links.

jmazzitelli:
this should be quick - I'm going to look at fixing this.

jmazzitelli:
It looks like we are missing `https://www.envoyproxy.io/` in front of all these doc links. If I replace `http://localhost:20001/` with `https://www.envoyproxy.io/` the links are ok.

PR #7156, opened 2024-02-28
fix envoyproxy.io doc links

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7071


jmazzitelli:
see the issue for the replication procedures - just clink the links and see that they now go to the envoy docs

jshaughn (approved):
Yay!

PR #7158, opened 2024-02-28
[hack] use kiali token for access to openshift user monitoring prometheus

jmazzitelli:
We changed the way this script works because the secret token we were using is [no longer valid in OpenShift 4.15](https://issues.redhat.com/browse/OSSM-6008). This new way will work in 4.12+ and is also simpler.

### DIFF
--- a/hack/use-openshift-prometheus.sh
+++ b/hack/use-openshift-prometheus.sh
@@ -68,6 +68,7 @@ echo "=== SETTINGS ==="
 # * Istio installed in ISTIO_NAMESPACE
 # * Kial Operator installed
 # * A Kiali CR installed
+# * A Kiali Server installed (we need the service account for it)
 #
 if ! ${OC} whoami &> /dev/null; then
   echo "Either you are not logged into OpenShift or [${OC}] is not a valid client executable"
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
This is required for OCP 4.15 right?  And a backward compatible simplifier for other supported OCP versions?  Is this something that needs to be documented somewhere for OSSM or OSSMC?

jshaughn:
The CI fail is that same flake.  We need to fix that or get rid of that test.

jmazzitelli:
> This is required for OCP 4.15 right? And a backward compatible simplifier for other supported OCP versions? Is this something that needs to be documented somewhere for OSSM or OSSMC?

Yes, required for this to work on OCP 4.15. And yes it is backward compatible (I tried on 4.14 and it worked; Jacek tried this similar setup on 4.12 and it worked for him).

Yes, it needs to be documented - that's what the JIRA is for that  I linked in this PR's description - see https://issues.redhat.com/browse/OSSM-6008

jmazzitelli:
There are no unit tests for this - this is an independent hack script that is not used anywhere in our CI tests. Ignoring the test results and merging...

PR #7162, opened 2024-03-01
change detection method for Istio Ambient

### LINKED ISSUE #7159: Kiali & Istio Ambient in OpenShift
### Describe the bug

A new Istio profile will be added in order to run Ambient in OpenShift. As OpenShift requires the privileged SCC, in order to simplify the installation, cni and ztunnel deployents will be installed in the `kube-system` namespace. 
Kiali is looking for the ztunnel deployment in the istio-system namespace, so is not detecting Ambient Mesh: 

image

Istio-system namespace: 

image

kube-system namespace: 

image

### Expected Behavior

Kiali detects Ambient in OpenShift. 
For this reason: 

- Set the ztunnel namespace as a parameter with 'istio-system' as a default
- Look for other indicators

1.Install Istio with the following parameter: 

`istioctl install --set hub=gcr.io/istio-testing --set tag=1.22-alpha.f1898ff2fc1406b7899dcf2409876f94b0e76dca --set profile=openshift-ambient`

### Environment

- **Kiali version:**
- **Istio version:**
- **Kubernetes impl:** 
- **Kubernetes version:** 

josunect:
### Describe the change

Add istioAmbientNamespace as a config parameter and add Ambient Namespace logic to projects (For OpenShift). 

Kiali running in OpenShift and Istio 1.22.alpha: 

image

image

image

image

### Steps to test the PR

(Note: I was not able to run it from the hack script because I was received `Error: generate config: failed to read profile openshift-ambient from : open profiles/openshift-ambient.yaml: file does not exist`, tested with: `istio/install-istio-via-istioctl.sh -cp openshift-ambient -ih gcr.io/istio-testing -it 1.22-alpha.f1898ff2fc1406b7899dcf2409876f94b0e76dca -ic ~/Downloads/istio-1.21.0-rc.0/bin/istioctl`

`wget -O - https://storage.googleapis.com/istio-build/dev/1.22-alpha.f1898ff2fc1406b7899dcf2409876f94b0e76dca/istioctl-1.22-alpha.f1898ff2fc1406b7899dcf2409876f94b0e76dca-linux-amd64.tar.gz | tar zxvf -`

./istioctl install --set hub=gcr.io/istio-testing --set tag=1.22-alpha.f1898ff2fc1406b7899dcf2409876f94b0e76dca --set values.meshConfig.enableAutoMtls=true --set values.pilot.env.ENABLE_NATIVE_SIDECARS=false   --set values.gateways.istio-egressgateway.enabled=true --set values.gateways.istio-ingressgateway.enabled=true --set values.meshConfig.defaultConfig.tracing.sampling=100.00 --set values.meshConfig.accessLogFile=/dev/stdout --set components.cni.enabled=true --set components.cni.namespace=kube-system --set values.cni.cniBinDir=/var/lib/cni/bin --set values.cni.cniConfDir=/etc/cni/multus/net.d --set values.cni.chained=false --set values.cni.cniConfFileName=istio-cni.conf --set values.sidecarInjectorWebhook.injectedAnnotations.k8s\.v1\.cni\.cncf\.io/networks=istio-cni --set profile=openshift-ambient

Deploy Kiali in the cluster. As the operator PR is not done, in order to test this, I just updated this line: https://github.com/kiali/kiali/compare/master...josunect:kiali:issue7159?expand=1#diff-fe44f09c4d5977b5f5eaea29170b6a0748819c9d02271746a20d81a5f3efca17R706 with `kube_system` value. 

Install bookinfo: 

`istio/install-bookinfo-demo.sh -c kubectl -ai false -tg`

### Automation testing

Added unit testing coverage

### Issue reference

Fixes #7159

### DIFF
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -1020,10 +1020,8 @@ func (in *IstioConfigService) IsAmbientEnabled() bool {
 	}
 	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > (10*time.Minute) {
 		ambientEnabled = new(bool)
-		selector := map[string]string{
-			"app": "ztunnel",
-		}
-		daemonset, err := in.kialiCache.GetDaemonSetsWithSelector(meta_v1.NamespaceAll, selector)
+
+		daemonset, err := in.kialiCache.GetDaemonSetsWithSelector(meta_v1.NamespaceAll, "app=ztunnel")
 		if err != nil {
 			log.Debugf("No ztunnel daemonset found in Kiali accessible namespaces: %s ", err.Error())
 		} else {
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -1020,9 +1020,9 @@ func (in *IstioConfigService) IsAmbientEnabled() bool {
 	}
 	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > time.Minute {
 		ambientEnabled = new(bool)
-		daemonset, err := in.kialiCache.GetDaemonSet(in.config.IstioNamespace, "ztunnel")
+		daemonset, err := in.kialiCache.GetDaemonSet(in.config.ExternalServices.Istio.IstioAmbientNamespace, "ztunnel")
 		if err != nil {
-			log.Debugf("No ztunnel found in istio namespace: %s ", err.Error())
+			log.Debugf("No ztunnel found in Istio Ambient namespace (%s): %s ", in.config.ExternalServices.Istio.IstioAmbientNamespace, err.Error())
 		} else {
 			if daemonset != nil {
 				*ambientEnabled = true
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -1018,7 +1018,7 @@ func (in *IstioConfigService) IsAmbientEnabled() bool {
 		lastUpdateTime = new(time.Time)
 		lastUpdateTime = &currentTime
 	}
-	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > time.Minute {
+	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > (10*time.Minute) {
 		ambientEnabled = new(bool)
 		selector := map[string]string{
 			"app": "ztunnel",
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -1011,7 +1011,7 @@ func (in *IstioConfigService) GatewayAPIClasses() []config.GatewayAPIClass {
 }
 
 // Check if istio Ambient profile was enabled
-// ATM it is defined in the istio-cni-config configmap
+// ATM it will look for app=ztunnel selector from a configmap
 func (in *IstioConfigService) IsAmbientEnabled() bool {
 	currentTime := time.Now()
 	if lastUpdateTime == nil {
@@ -1020,11 +1020,14 @@ func (in *IstioConfigService) IsAmbientEnabled() bool {
 	}
 	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > time.Minute {
 		ambientEnabled = new(bool)
-		daemonset, err := in.kialiCache.GetDaemonSet(in.config.ExternalServices.Istio.IstioAmbientNamespace, "ztunnel")
+		selector := map[string]string{
+			"app": "ztunnel",
+		}
+		daemonset, err := in.kialiCache.GetDaemonSetsWithSelector(meta_v1.NamespaceAll, selector)
 		if err != nil {
 			log.Debugf("No ztunnel found in Istio Ambient namespace (%s): %s ", in.config.ExternalServices.Istio.IstioAmbientNamespace, err.Error())
 		} else {
-			if daemonset != nil {
+			if len(daemonset) > 0 {
 				*ambientEnabled = true
 				return true
 			} else {
--- a/business/istio_config.go
+++ b/business/istio_config.go
@@ -1020,8 +1020,10 @@ func (in *IstioConfigService) IsAmbientEnabled() bool {
 	}
 	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > (10*time.Minute) {
 		ambientEnabled = new(bool)
-
-		daemonset, err := in.kialiCache.GetDaemonSetsWithSelector(meta_v1.NamespaceAll, "app=ztunnel")
+		selector := map[string]string{
+			"app": "ztunnel",
+		}
+		daemonset, err := in.kialiCache.GetDaemonSetsWithSelector(meta_v1.NamespaceAll, selector)
 		if err != nil {
 			log.Debugf("No ztunnel daemonset found in Kiali accessible namespaces: %s ", err.Error())
 		} else {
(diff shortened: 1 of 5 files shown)


business/istio_config.go
```diff
@@ -1020,9 +1020,9 @@ func (in *IstioConfigService) IsAmbientEnabled() bool {
 	}
 	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > time.Minute {
 		ambientEnabled = new(bool)
-		daemonset, err := in.kialiCache.GetDaemonSet(in.config.IstioNamespace, "ztunnel")
+		daemonset, err := in.kialiCache.GetDaemonSet(in.config.ExternalServices.Istio.IstioAmbientNamespace, "ztunnel")
```
nrfox, line 1023: Is there a label that the ztunnel daemonsets have that we can use to look for them? Rather than doing a `GET` in a specific namespace, can we list across all namespaces using label and just check if `len(result) > 0`? This wouldn't require adding another config option and for the user to set the right config option. You might need to adapt the `kialiCache.GetDaemonSets()` method to work when `accessible_namespaces != **` but just look at the istio config list methods for an example.

nrfox (changes requested):

business/istio_config.go
josunect: I see that the istio pod has this annotation:

`Annotations:      ambient.istio.io/redirection: disabled`

At least, at the moment, it looks like the ztunnel namespace is predefined by istio, so, if Kiali knows somehow if we are in OpenShift we can look for the ztunnel in the kube-system namespace. Or even have this 3rd option to change in case ztunnel cannot be found in the defaults namespaces.

nrfox: If this label is stable: `app=ztunnel` I would do something like this: `GetDaemonSets(meta_v1.NamespaceAll, app=ztunnel)`. You'll probably need to adjust the implementation of `GetDaemonSets` or add a new method that accepts a selector label but see one of the istio cache methods for an example: https://github.com/kiali/kiali/blob/757c6b47f41933147e678b7c98f9a0d74ad4f372/kubernetes/cache/kube_cache.go#L1183-L1228

This should work even if `accessible_namespaces != **`.

> At least, at the moment, it looks like the ztunnel namespace is predefined by istio, so, if Kiali knows somehow if we are in OpenShift we can look for the ztunnel in the kube-system namespace. 

I think we should avoid having special cases for Openshift if we can and instead try to find an implementation that works for both.

The `ambient.istio.io/redirection` annotation isn't specific to ztunnel though right? Other daemonsets could also use that annotation? I also don't think you can use annotations in `ListOptions`.

josunect: Thanks, @nrfox ! This approach works as expected. 

Probably it doesn't need to be checked so often so perhaps I would also considered changing the refresh time: 
https://github.com/kiali/kiali/pull/7162/files#diff-163a601f65fbd474e86f1928ac31ba72659f4fd009c067c6f3226ce066f8861fR1021

What do you think? 

Tested in crc: 
image

nrfox: Yeah 1 minute seems low. 10 minutes?

business/istio_config.go
```diff
@@ -1020,11 +1020,14 @@ func (in *IstioConfigService) IsAmbientEnabled() bool {
...
 	if ambientEnabled == nil || currentTime.Sub(*lastUpdateTime) > time.Minute {
 		ambientEnabled = new(bool)
-		daemonset, err := in.kialiCache.GetDaemonSet(in.config.IstioNamespace, "ztunnel")
+		selector := map[string]string{
+			"app": "ztunnel",
+		}
+		daemonset, err := in.kialiCache.GetDaemonSetsWithSelector(meta_v1.NamespaceAll, selector)
 		if err != nil {
-			log.Debugf("No ztunnel found in istio namespace: %s ", err.Error())
+			log.Debugf("No ztunnel found in Istio Ambient namespace (%s): %s ", in.config.ExternalServices.Istio.IstioAmbientNamespace, err.Error())
```
nrfox, line 1028: This log statement isn't accurate anymore since kiali is checking all namespaces.

kubernetes/cache/kube_cache.go
```diff
@@ -537,6 +538,38 @@ func (c *kubeCache) GetDaemonSet(namespace, name string) (*apps_v1.DaemonSet, er
...
 }
 
+func (c *kubeCache) GetDaemonSetsWithSelector(namespace string, selectorLabels map[string]string) ([]apps_v1.DaemonSet, error) {
+	defer c.cacheLock.RUnlock()
+	c.cacheLock.RLock()
+
+	daemonSets, err := c.GetDaemonSets(namespace)
+	if err != nil {
+		return nil, err
+	}
```
nrfox, line 548: Have you tested this with `accessible_namespace != **`? I don't think this will work until you update the `GetDaemonSets` method to be like the istio config list methods.

josunect: Thanks, @nrfox, I've thought that with metav1.NamespaceAll you just got the namespaces that you had permissions to. I've made the changes and some testing.

image

image

Not sure where, but probably we should add to the docs that, Kiali should have access to the namespace where the ztunnel is located.

business/istio_config.go
josunect: Changed the message

kubernetes/cache/kube_cache.go
```diff
@@ -537,6 +538,64 @@ func (c *kubeCache) GetDaemonSet(namespace, name string) (*apps_v1.DaemonSet, er
...
+
+	if namespace == metav1.NamespaceAll {
+		if c.clusterScoped {
+			daemonSets, err = c.GetDaemonSets(namespace)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			for _, nsCacheLister := range c.nsCacheLister {
+				dsNS, err2 := nsCacheLister.daemonSetLister.List(labels.Everything())
```
nrfox, line 557: If you pass the label selector into the `List()` method rather than `labels.Everything()` then you can remove all of the manual filtering below. The label selector is meant to be passed into the list methods this way and the filtering will be handled by the underlying client.

See the one of the Istio methods like `GetDestinationRules` as an example: https://github.com/kiali/kiali/blob/90ceb89066c2fb82a06dd3fdcc5c875f8c8c30a7/kubernetes/cache/kube_cache.go#L869

nrfox (changes requested):

kubernetes/cache/kube_cache.go
nrfox: That's a good point. Maybe the ambient docs on kiali.io?

kubernetes/cache/kube_cache.go
josunect: Yes, I don't know why when I tried something similar it was not working, that's why I added all that filtering logic. It looks to be working now

kubernetes/cache/kube_cache.go
```diff
@@ -537,6 +538,53 @@ func (c *kubeCache) GetDaemonSet(namespace, name string) (*apps_v1.DaemonSet, er
...
+	if err != nil {
+		return nil, err
+	}
+
+	if namespace == metav1.NamespaceAll {
+		if c.clusterScoped {
+			daemonSets, err = c.GetDaemonSets(namespace)
+			if err != nil {
+				return nil, err
+			}
```
nrfox, line 558: I think I see why the filtering isn't working. Calling `c.GetDaemonSets` here won't apply the `selector` since `c.GetDaemonSets` just passes `labels.Everything()`. You probably need to just copy these lines from that function so that you can pass the selector into the `List` call:
```
	daemonSets, err := c.getCacheLister(namespace).daemonSetLister.DaemonSets(namespace).List(labels.Everything())
	if err != nil {
		return nil, err
	}
```
rather than trying to call `c.GetDaemonSets` and then add the type/deep copy at the bottom.

Issue #6036, opened 2023-04-17
Multicluster - External controlplane support


nrfox:
Support the external controlplane deployment model.

Xunzhuo:
I can work on it : )

nrfox:
Technically this is supported already as an "experimental" feature but afaik it is only documented on [Kiali's Github page](https://github.com/kiali/kiali#observing-a-remote-cluster) and there's no automated e2e testing for this deployment model.

To move this feature out of experimental, imo we at least need some automated testing around this and documentation on Kiali.io. Testing should be fairly simple since the Kiali binary just needs to be run outside the cluster and some basic validation performed. Running kiali outside the cluster is already supported for dev envs (see https://github.com/kiali/kiali#running-standalone) but iirc that uses a different code path than the "external cluster" setup for whatever reason. That dev env setup should be merged with this one.

jshaughn:
@Xunzhuo , will you do this or should I assign to someone else?

nrfox:
Support has been added by #7168

PR #7168, opened 2024-03-04
Add external controlplane support

### LINKED ISSUE #4729: Support Istio external control plane
When istio is deployed as external control plane, Istio workload will be running in a different cluster from the istio control plane, since kiali can only access one cluster, there is no way currently for kiali to access istiod and at the same time to load workload status.

If kiali can be enhanced to do the following, then this problem can be resolved.
1. Add detection when /kiali-remote-secret/kiali exists, this indicates that istio workload is remote and so is the control plane.
2. When istio control plane is not found, kiali should look for the istiod running incluster which basically follow the path when /kiali-remote-secret/kiali does not exist.

Doing this while kiali try to access istiod will have very small impact to Kiali but truly support Istio external control plane.

nrfox:
### Describe the change

Adds support for the external controlplane deployment model. Also does some cleanup of the istio config service.

### Steps to test the PR

Manual testing:
1. `hack/run-integration-tests.sh --test-suite backend-external-controlplane --setup-only true`
2. Visit "mesh" page.
3. Verify functionality of overview / list pages.

### Automation testing

Automated testing: `hack/run-integration-tests.sh --test-suite backend-external-controlplane`

Added a hack script to setup an external controlplane environment. Added a github workflow to run e2e tests against the external controlplane environment.

Added backend external controlplane tests to verify that the `/mesh` endpoint returns the proper info for the external-controplane deployment.

### Issue reference

Fixes #4729 #6036


hhovsepy (approved):
Tested the Istio config list locally, regression tested. Code LGTM.

Issue #7122, opened 2024-02-15
[mesh page] Fix "Fit" behavior


jshaughn:
The mesh graph is not properly re-fitting to the page dimensions on various actions.  It should re-fit when the side-panel collapses or expands, when a box is expanded or collapsed, etc...

PR #7169, opened 2024-03-04
Fix Mesh "fit" issues

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7122

Fix Mesh "fit" issues via graph reset and/or layout in response to:
- group box collapse/expand
- graph canvas resize

### Steps to test the PR

Check to see if the graph resets/relayouts reasonably(*) when you:
- Collapse/Expand the data-plane box.
- Collapse/Expand the side panel.
- Resize the browser

(*) By "reasonably" I mean it doesn't have to be perfect.  There are still unusual layout issues, especially with the collapsed "dataplane" box, and sometimes labels may not always be completely visible.


ferhoyos (approved):
LGTM. It works "reasonably" well. Like you mention in the PR the re-layout is not perfect but this is the work of `react-topology`  (we cannot do much here except executing `group.getGraph()?.layout()`)

PR #7170, opened 2024-03-05
Remove unused patternfly background images

ferhoyos:
### Describe the change

Remove patternfly background images (path `frontend/src/img`) since they are not used in Kiali application (at least I have not seen any reference in the code, all images are stored in `frontend/src/asssets/img`).

### Steps to test the PR

Just check that Kiali works as expected


jshaughn (approved):
I guess this is fine.  I didn't even know they were there.

PR #7175, opened 2024-03-06
[Perf] Added performance baselines

hhovsepy:
### Describe the change

Added 'baselines.json' file with some test baselines.
Using the baselines to compare load time for each test.
Currently only logging the compared results without failing.

### Steps to test the PR
make perf-tests-run

### Issue reference
https://github.com/kiali/kiali/issues/7174

### DIFF
--- a/frontend/cypress/perf/common.ts
+++ b/frontend/cypress/perf/common.ts
@@ -65,17 +65,17 @@ export const measureDetailsLoadTime = (name: string, baseline: number, detailsUr
 export const compareToBaseline = (resultMS: number, baseline: number): string => {
   // to seconds
   const result = resultMS / 1000;
-  const resultSeconds = result.toPrecision(5);
-  const difference = baseline - result;
+  const resultOutput = result.toPrecision(5);
 
-  return `${resultSeconds} sec, baseline: ${baseline} sec, difference: ${formatNumberWithSign(difference)} sec\n`;
+  return `${resultOutput} sec, baseline: ${baseline} sec, difference: ${getDifference(result, baseline)} sec\n`;
 };
 
-const formatNumberWithSign = (num: number): string => {
-  if (num > 0) {
-    return `+${num.toPrecision(precision)}`;
-  } else if (num < 0) {
-    return `-${Math.abs(num).toPrecision(precision)}`;
+const getDifference = (result, baseline: number): string => {
+  const difference = result - baseline;
+  if (difference > 0) {
+    return `+${difference.toPrecision(precision)}`;
+  } else if (difference < 0) {
+    return `-${Math.abs(difference).toPrecision(precision)}`;
   } else {
     return '0';
   }
--- a/frontend/cypress/perf/common.ts
+++ b/frontend/cypress/perf/common.ts
@@ -1,6 +1,7 @@
 export const reportFilePath = 'cypress/results/performance.txt';
 
 export const visits = 5;
+const precision = 5;
 export const baselines = 'baselines';
 
 before(() => {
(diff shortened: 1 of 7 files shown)


jshaughn (dismissed):
This visually looks fine.  It didn't really run for me but that's probably an env issue on my side, because I also can't seem to execute the existing `make cypress-run` command.  Both fail in a similar way.

frontend/cypress/perf/common.ts
```diff
@@ -39,15 +50,23 @@ const measureLoadTime = (name: string, visits: number, loadUrl: string, loadElem
...
 
-export const measureDetailsLoadTime = (name: string, visits: number, detailsUrl: string): void => {
-  measureLoadTime(name, visits, detailsUrl, '.pf-v5-c-tabs');
-}
+export const compareToBaseline = (resultMS: number, baseline: number): string => {
+  // to seconds
+  const result = resultMS / 1000;
+  const resultSeconds = result.toPrecision(5);
+  const difference = baseline - result;
+  return `${resultSeconds} sec, baseline: ${baseline} sec, difference: ${difference.toPrecision(5)} sec\n`;
```
nrfox, line 71: nit: it would be nice to have `+` or `-` in the difference like:
`App Outbound Metrics load time: 0.45238 sec, baseline: 1 sec, difference: -0.54762 sec`
`App Outbound Metrics load time: 2 sec, baseline: 1 sec, difference: +1 sec`

nrfox (commented):
One small nit but not blocking.

Also these tests have undefined/NAN for me:
```
[Empty Namespaces]                                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                                          
Namespaces: 10                                                                                                                                                                                                                                                                            
Init page load time: 0.59296 sec, baseline: undefined sec, difference: NaN sec

Namespaces: 50                                                        
Init page load time: 0.97492 sec, baseline: undefined sec, difference: NaN sec

Namespaces: 300                                                       
Init page load time: 21.913 sec, baseline: undefined sec, difference: NaN sec
```

frontend/cypress/perf/common.ts
hhovsepy: yeah good point,

hhovsepy:
@nrfox fixed the NaN and the formatting, @jshaughn you can try by running 'yarn cypress:run:perf' in frontend directory.

jshaughn (dismissed):
Still unable to run this because of non-pr issues, but it still looks fine to me.

nrfox:
I still see `undefined` for one of the baselines:
```
[Overview page]                                                       

Namespaces: 10                                                        
Init page load time: 0.75760 sec, baseline: 2 sec, difference: +1.2424 sec                                                                                                                                                                                                                

Namespaces: 50                                                        
Init page load time: 1.2810 sec, baseline: 2 sec, difference: +0.71900 sec                                                                                                                                                                                                                

Namespaces: 300                                                       
Init page load time: 22.915 sec, baseline: undefined sec, difference: 0 sec                                                                                                                                                                                                               
```

Also I think the `+/-` are inverted. Shouldn't it be `-` when the time goes down compared to the baseline and not `+` e.g. from the example above: `-0.71900 sec` ? In this case `-` is good.

hhovsepy:
Fixed the typo.
> I still see `undefined` for one of the baselines:

Ok, seems better to show `+` when result exceeds baseline.
> Also I think the `+/-` are inverted. Shouldn't it be `-` when the time goes down compared to the baseline and not `+` e.g. from the example above: `-0.71900 sec` ? In this case `-` is good.

nrfox (approved):

Issue #7179, opened 2024-03-07
Multi mesh setup results in an error when fetching workloads


minuyim:
### Describe the bug
https://github.com/kiali/kiali/discussions/6761
As in the above environment, I am running multi-mesh / cluster. And each mesh has its own kiali and prometheus.
When there is very little traffic with other clusters, there is an issue where graphs cannot be drawn intermittently.
```
2024-03-06T02:19:01Z ERR Error fetching Workloads per namespace xxx: Cluster [b] is not found or is not accessible for Kiali
2024-03-06T02:19:01Z ERR Cluster [b] is not found or is not accessible for Kiali: goroutine 51603 [running]:
runtime/debug.Stack()
	/opt/hostedtoolcache/go/1.20.10/x64/src/runtime/debug/stack.go:24 +0x65
github.com/kiali/kiali/handlers.handlePanic({0x240d580, 0xc0b7d94df8})
	/home/runner/work/kiali/kiali/handlers/graph.go:86 +0x219
panic({0x1cd6360, 0xc0d57dc900})
	/opt/hostedtoolcache/go/1.20.10/x64/src/runtime/panic.go:884 +0x213
github.com/kiali/kiali/graph.CheckError(...)
	/home/runner/work/kiali/kiali/graph/util.go:38
github.com/kiali/kiali/graph/telemetry/istio/appender.getWorkloadList({0xc089f7fcf6, 0x2}, {0xc0cc8d8b16, 0xd}, 0xc0fd395800)
	/home/runner/work/kiali/kiali/graph/telemetry/istio/appender/appender.go:345 +0x40a
github.com/kiali/kiali/graph/telemetry/istio/appender.getWorkload({0xc089f7fcf6?, 0xc0fd43b800?}, {0xc0cc8d8b16?, 0x6?}, {0xc0f66e0e28, 0x14}, 0x0?)
	/home/runner/work/kiali/kiali/graph/telemetry/istio/appender/appender.go:356 +0x65
github.com/kiali/kiali/graph/telemetry/istio/appender.DeadNodeAppender.applyDeadNodes({}, 0x28?, 0x1ea5780?, 0xc0b945f5a8)
	/home/runner/work/kiali/kiali/graph/telemetry/istio/appender/dead_node.go:106 +0x2bb
[...]

PR #7181, opened 2024-03-07
Fix DeadNode appender when dealing with inaccessible workload

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7179

Fix DeadNode appender when dealing with inaccessible workload.  Don't try to fetch it, just assume it's valid.

To Test:
A little bit tricky, covered now with a unit test.  You need to set up requests from an accessible to an inaccessible workload (like from an accessible namespace to an inaccessible namespace, or cluster).  Then stop the flow of traffic.  And then it would blow up because DeadNodeAppender would try to fetch the inaccessible workload, but only happens when the traffic stops, which lest DNA get to that point.


jmazzitelli (approved):
test passes. good to go

PR #7184, opened 2024-03-11
Services list per cluster

hhovsepy:
### Describe the change

In Service list page, there was sending a REST API request to retrieve services per namespace, which was causing to filter per cluster which contains the provided namespace.
Now Optimized to send Service list request to REST API per cluster with providing selected namespaces list.

### Steps to test the PR

Create a set of namespaces:
`for i in {1..200}; do kubectl create ns test-$i; done`
Open the Service list page and select all services.
Check if Health and Configuration fields are loading correctly: There was a bug here which was mirroring one cluster's Configuration to another one when two Services with same name/namespace.

### Automation testing

Added unit tests. Added integration tests.

### Issue reference

### DIFF
--- a/business/services.go
+++ b/business/services.go
@@ -10,6 +10,7 @@ import (
 	apps_v1 "k8s.io/api/apps/v1"
 	core_v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/kiali/kiali/business/checkers"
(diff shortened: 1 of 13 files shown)


tests/integration/tests/services_test.go
```diff
@@ -37,6 +37,30 @@ func TestServicesList(t *testing.T) {
 	require.Equal(kiali.BOOKINFO, serviceList.Namespace.Name)
 }
 
+func TestClusterServicesList(t *testing.T) {
+	require := require.New(t)
+	serviceList, err := kiali.ClusterServicesList(kiali.BOOKINFO)
+
+	require.NoError(err)
+	require.NotEmpty(serviceList)
+	require.True(len(serviceList.Services) >= 4)
```
nrfox, line 46: It would probably be good to first query the kube api for services and then query the kiali api for services and make sure the two list lengths match. I'm not sure the significance of `>= 4` here.

tests/integration/utils/kiali/kiali_client.go
```diff
@@ -311,6 +311,21 @@ func ServicesList(namespace string) (*ServiceListJson, error) {
...
+		err = json.Unmarshal(body, &serviceList)
+		if err == nil {
+			return serviceList, nil
+		} else {
+			return nil, err
+		}
+	} else {
+		return nil, err
+	}
+}
```
nrfox, line 327: You can use `getRequestAndUnmarshalInto(url, serviceList)` to save on typing a bit. It will also check the status code before unmarshaling the json so that the actual api error doesn't get swallowed by the unmarshal error when the response is not 200.

handlers/services.go
```diff
@@ -101,6 +106,78 @@ func ServiceList(w http.ResponseWriter, r *http.Request) {
 	RespondWithJSON(w, http.StatusOK, serviceList)
 }
 
+// ClustersServices is the API handler to fetch the list of services from a given cluster
+func ClustersServices(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
hhovsepy: The `ServiceList` handler is retrieving a single namespace's services across all clusters, while the `ClustersServices` is retrieving services from given namespaces or from all namespaces for a given cluster only.
The `ServiceList` handler should be removed, it is staying only for backward compatibility in Backstage.

tests/integration/tests/services_test.go
hhovsepy: this is the copy paste of previous test but including the new API call

hhovsepy: there should be at least 4 default bookinfo services

business/services.go
```diff
@@ -399,6 +407,7 @@ func (in *SvcService) buildRegistryServices(rSvcs []*kubernetes.RegistryService,
 		svcReferences = FilterUniqueIstioReferences(svcReferences)
 		// External Istio registries may have references to ServiceEntry and/or Federation
 		service := models.ServiceOverview{
+			Cluster:           clusterId,
```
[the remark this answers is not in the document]
hhovsepy: yes, this is commented in the beginning of the method, and registry services filtered by 'clusterId ' by calling `filterIstioServiceByClusterId`

handlers/services.go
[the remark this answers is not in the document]
jshaughn: Is there a comment/TODO for that future removal?

hhovsepy:
closing this PR in favor of https://github.com/kiali/kiali/pull/7229

Issue #7180, opened 2024-03-07
Improve Cypress test related in kiali_help.feature


matejnesuta:
### Describe the bug
This popped up to me on a cluster with no Istio running. 

Test related to "View Certificates Info" needs to be improved. I don't think it should pass while the menu displays an error message. 
![image](https://github.com/kiali/kiali/assets/45061792/03d9be5f-a70a-46b0-9818-ace8eaec2240)

PR #7191, opened 2024-03-13
Negative check added for the Certificates information test

matejnesuta:
### Issue reference

#7180

### DIFF
--- a/frontend/cypress/integration/common/kiali_help.ts
+++ b/frontend/cypress/integration/common/kiali_help.ts
@@ -14,6 +14,9 @@ When('user clicks on the {string} button', (title: string) => {
 
 Then('user sees the {string} modal', (title: string) => {
   cy.get('h1.pf-v5-c-modal-box__title').contains(title).should('be.visible');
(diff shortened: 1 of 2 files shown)


frontend/cypress/integration/common/kiali_help.ts
```diff
@@ -14,6 +14,9 @@ When('user clicks on the {string} button', (title: string) => {
 
 Then('user sees the {string} modal', (title: string) => {
   cy.get('h1.pf-v5-c-modal-box__title').contains(title).should('be.visible');
+  if (title === 'Certificates information'){
+    cy.contains('An error occurred getting certificates information').should('not.exist');
+  }
```
nrfox, line 19: I would put this as another step in whatever feature file is testing the certs modal rather than putting a special case in this definition for the `Certificates information` title.

Something like:
```
Then user sees the "Certificates information" modal
And there's no error getting certificate information
```

nrfox (changes requested):

nrfox (approved):

PR #7192, opened 2024-03-14
[hack] hack scripts that allow you to install Kiali and Istio via operators (including the Sail operator)

jmazzitelli:
This allows you to quickly install Kiali and Istio using the Kiali Operator and Sail (aka ServiceMesh3) operator via OLM.

Right now it only works on OpenShift. See followup comment for support on non-OpenShift clusters.


jmazzitelli:
These scripts will eventually be able to support installing everything on minikube or other non-OpenShift clusters. We just need to wait for [the new sail operator to get published on operatorhub.io](https://issues.redhat.com/browse/OSSM-4829). When that happens we need to tweak func-sm.sh and its install_servicemesh_operators function (see the comments in there on what needs to be done).

jshaughn (approved):
At first I wasn't sure about using the `Sail` project name in our folders and such, but I guess it's alright, it's just an opensource project.  And these are just hack scripts.

Visual inspection looks ok to me.

jmazzitelli:
we can always rename the directory.. but yeah, its just hack scripts.

PR #7195, opened 2024-03-18
Update protobuf dependency

jshaughn:
https://issues.redhat.com/browse/OSSM-6037

This same change should be applied to 1.73 branch.


ferhoyos (approved):
Not tested, but CI passes and they are small upgrades, so lgtm.

Issue #7198, opened 2024-03-18
Business.Svc.GetService does not supply cluster in the returned svc.NamespaceInfo structure


jshaughn:
Something seems wrong when fetching a Service from the Business layer.  In my situation, single cluster, it seems the returned service definition's `NamespaceInfo` does not set `Cluster`.  I'm not sure about the MC scenario, nor am I sure what other information may be unset or whether this affects other "Get" functions.

PR #7199, opened 2024-03-18
Make Service objects more dependably cluster-aware

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7198

Make the Service model cluster-aware so that calling layers can assume the Namespace object is set with both cluster and namespace names.  Extend the change to the UI as needed.  Also uncovered an existing type mapping problem for Service objects in the UI, which is now fixed.

In general, even though we support cross-cluster query, the resulting objects are cluster-specific.  We should make sure that when possible we deal with cluster-aware object definitions.  This is just one use case, there may be others.


models/service.go
```diff
@@ -73,7 +73,6 @@ type ServiceDefinitionList struct {
 }
 
 type ServiceDetails struct {
-	Cluster            string                                   `json:"cluster"`
```
nrfox, line 76: I'm not really sure why we embed the whole `Namespace` object in the `Service` model since it looks like just the name was being parsed anyways and you'd lose all the rest of the fields if you relied on e.g. `svc.Namespace.Labels`. Can we change `Namespace` to be a string and move this `Cluster` field to the `Service` model so it's more consistent with app/workload?

jshaughn: The `Namespace` object was already there, and was the original source of confusion as I had assumed it would minimally have the cluster and namespace names set.  But the entire object had only the namespace name.  I left it in place because I was unsure if some code path may make use of the additional fields, but I can make the suggested change.  It's functionally equivalent, and if its more clear and consistent then we should do it.

nrfox (approved):

PR #7200, opened 2024-03-19
Cache namespaces per cluster and per token

### LINKED ISSUE #6037: Multicluster - Token per cluster
Associate a token per user per cluster. Kiali can assume that the user credentials will remain the same across clusters but the backend will still need to authenticate with those credentials to each kube cluster to offer a SSO experience with multicluster.

nrfox:
### Describe the change

Caches namespaces per token and per cluster rather than simply per token.

### Steps to test the PR

Regression testing.

### Automation testing

Unit tests added. Unable to fully cover with e2e tests since token per cluster is not a supported deployment yet.

### Issue reference

Fixes #6037


business/istio_validations.go
```diff
@@ -553,18 +553,14 @@ func (in *IstioValidationsService) fetchNonLocalmTLSConfigs(mtlsDetails *kuberne
...
 
-	kubeCache := kialiCache.GetKubeCaches()[cluster]
-	if kubeCache == nil {
-		return
-	}
-
-	userClient := in.userClients[cluster]
-	if userClient == nil {
+	cfg := config.Get()
+	// TODO: Handle multi-primary instead of only using home cluster.
```
jshaughn, line 557: Is there an issue for this?  We do claim MP support at this point.

kubernetes/cache/cache.go
```diff
@@ -202,5 +204,78 @@ func (in *kialiCacheImpl) IsAmbientEnabled(cluster string) bool {
...
+	c.namespacesLock.Lock()
+	defer c.namespacesLock.Unlock()
+
+	namespacesByCluster := make(map[string][]models.Namespace)
+	for _, namespace := range namespaces {
+		namespacesByCluster[namespace.Cluster] = append(namespacesByCluster[namespace.Cluster], namespace)
+	}
+
+	for cluster, clusterNamespaces := range namespacesByCluster {
+		key := namespacesKey{cluster: cluster, token: token}
```
jshaughn, line 257: Any need to validate the token here, check for empty, before it gets used in a map key?

jshaughn (approved):
Visual inspection of logic looks good.  Good, distinct error messages.  A couple of minor, non-blocking, comments

business/istio_validations.go
nrfox: This is one of those things that should get cleaned up with moving things to the mesh page. I'll create a separate issue to capture that list.

nrfox: Created: https://github.com/kiali/kiali/issues/7203

PR #7201, opened 2024-03-19
[v1.73] Add make ossmconsole-create option

ferhoyos:
### Describe the change

Backport of make option `ossmconsole-create` to be able to deploy local OSSMConsole plugin in remote clusters for 1.73 release (OSSM 2.5)

### Steps to test the PR

Run `make clean build build-ui cluster-push operator-create kiali-create ossmconsole-create` in OCP cluster to check that OSSMConsole plugin is installed correctly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,8 +61,10 @@ CYPRESS_TESTS_QUAY_TAG ?= ${CYPRESS_TESTS_QUAY_NAME}:${CYPRESS_TESTS_CONTAINER_V
 
 # Where the control plane is
 ISTIO_NAMESPACE ?= istio-system
-# Declares the namespace/project where the objects are to be deployed.
+# Declares the namespace/project where the Kiali objects are to be deployed.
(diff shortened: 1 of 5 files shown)


jmazzitelli (approved):
some of this is likely going to break other things (for example, 1.73 is still meant to work witih Maistra, but yet we removed the IS_MAISTRA env var set up here). But these makefiles were also broken trying to work with ossmc so, it is just broken in different places. :)

But we rarely use the make files in release branches to do things so if this works for what fernando needs, we can merge this. If anyone runs into the broken parts, we can fix it later. As I say, we rarely need the make files in the branches (other than the ones that actually do the builds, which this isn't touching).

ferhoyos:
Yes, this is only needed to deploy local OSSM Console code from v1.73 in remote cluster which contains any special configuration (e.g. cluster with IPv6). Maybe it is not used anymore (normal debugging in local development does not use these make options), but just in case.

PR #7204, opened 2024-03-19
[hack] make sure we get a version from github

jmazzitelli:


nrfox (approved):

jmazzitelli:
I can see this portion of the hack script running successfully in the CI tests... ignoring the failure as it is not related. will merge now.

PR #7211, opened 2024-03-21
Apps lists page optimized to query per cluster

hhovsepy:
### Describe the change

Apps page was querying backend per namespace, now it queries per cluster by providing selected namespaces.
New API endpoing "api/clusters/apps" is added.

### Steps to test the PR

Create a set of namespaces:
Open the Apps list page and select all apps.
Check if all fields are loaded correctly.

### Automation testing

Modified e2e and unit tests.

### Issue reference

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -89,6 +89,10 @@ func (in *AppService) GetClusterAppList(ctx context.Context, criteria AppCriteri
 	}
 
 	allApps, err := in.businessLayer.App.fetchNamespaceApps(ctx, namespace, cluster, "")
+	if err != nil {
+		log.Errorf("Error fetching Applications for cluster %s per namespace %s: %s", cluster, namespace, err)
+		return *appList, err
+	}
 
 	icCriteria := IstioConfigCriteria{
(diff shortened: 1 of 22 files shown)


ferhoyos (approved):
Good performance improvement @hhovsepy!

Test with 200 namespaces
Before: 200 requests that last around 500ms

image


Now: 1 request of 90 ms (5x improvement)

image

Just one question... what is the difference between `GetClusterAppList` and `GetAppList`? Both have similar code and same input criteria. Could they be unified in one method to simplify?

hhovsepy:
Thanks @ferhoyos , `GetAppList` is the old method which loops over all clusters, this method it not removed because it is used in Graph, which still uses info from all clusters to generate.
The `GetClusterAppList` is the optimized method which does the same as before but for only one cluster.
They cannot be unified for now because the `GetAppList` should be gone once Graph loading is reworked.

ferhoyos:
Thanks for the clarification @hhovsepy. Ok for me if the `GetAppList` method is going to be removed in a future PR.

PR #7212, opened 2024-03-21
[hack] silly copy-paste typos in hack script

jmazzitelli:


jshaughn (approved):
C+P is both God and Devil

PR #7214, opened 2024-03-21
Start moving component info from About Box to Mesh Page

### LINKED ISSUE #6431: Show mesh configuration per controlplane
Kiali has assumed that there's just one controlplane in the mesh. Things like the "About page" which shows "Istio version", the  "Mesh mTLS lock" which shows if your mesh is secured with mTLS based on some mesh config settings from the controlplane, etc. all assume that there's a single controlplane. However even with single cluster, if you use istio revisions, there can be multiple controlplanes that Kiali should take into account for these various features. This is especially true for multi-primary.

This coincides with #5913 as much of the info mentioned above may move to that page and be shown per controlplane.

jshaughn:
- replace components in About with link to mesh page
- move Mesh TLS Status from masthead to mesh page side panel
- add more proper external service handling/url auto-discovery in mesh generation
- move Kiali instance info to Cluster side panel
- fix headers for some side panels
- add formatted config json as side panel body for nodes
    
Closes #6431

### DIFF
--- a/frontend/src/components/About/AboutUIModal.tsx
+++ b/frontend/src/components/About/AboutUIModal.tsx
@@ -11,16 +11,13 @@ import {
   Alert
 } from '@patternfly/react-core';
 import kialiIconAbout from '../../assets/img/icon-aboutbkg.svg';
-import { ExternalServiceInfo, Status, StatusKey } from '../../types/StatusState';
+import { Status, StatusKey } from '../../types/StatusState';
 import { config, kialiLogoDark } from '../../config';
 import { kialiStyle } from 'styles/StyleUtils';
 import { KialiIcon } from 'config/KialiIcon';
-import { TEMPO } from '../../types/Tracing';
-
-import { GetTracingUrlProvider } from '../../utils/tracing/UrlProviders';
+import { IstioLogo } from 'pages/Mesh/MeshLegendData';
 
 type AboutUIModalProps = {
-  externalServices: ExternalServiceInfo[];
   isOpen: boolean;
   onClose: () => void;
   status: Status;
@@ -37,10 +34,6 @@ const iconStyle = kialiStyle({
   marginRight: '0.5rem'
 });
 
-const websiteStyle = kialiStyle({
-  marginRight: '2rem'
-});
-
 const textContentStyle = kialiStyle({
   $nest: {
     '& dt, & dd': {
@@ -49,8 +42,13 @@ const textContentStyle = kialiStyle({
   }
 });
 
+const websiteStyle = kialiStyle({
+  marginRight: '2rem'
+});
+
 export const AboutUIModal: React.FC<AboutUIModalProps> = (props: AboutUIModalProps) => {
-  const additionalComponentInfoContent = (externalService: ExternalServiceInfo): string | JSX.Element => {
+  /*
+  const additionalComponentInfoContent = (externalService: ExternalServiceInfo) => {
     if (!externalService.version && !externalService.url) {
       return 'N/A';
     }
@@ -70,8 +68,10 @@ export const AboutUIModal: React.FC<AboutUIModalProps> = (props: AboutUIModalPro
       </>
     );
   };
+  */
 
-  const renderTempo = (externalServices: ExternalServiceInfo[]): JSX.Element => {
+  /*
+  const renderTempo = (externalServices: ExternalServiceInfo[]) => {
     const tempoService = externalServices.find(service => service.name === TEMPO);
 
     if (tempoService) {
@@ -81,8 +81,10 @@ export const AboutUIModal: React.FC<AboutUIModalProps> = (props: AboutUIModalPro
       return <></>;
     }
   };
+  */
 
-  const renderComponent = (externalService: ExternalServiceInfo): JSX.Element => {
+  /*
+  const renderComponent = (externalService: ExternalServiceInfo) => {
     const name = externalService.version ? externalService.name : `${externalService.name} URL`;
     const additionalInfo = additionalComponentInfoContent(externalService);
     return (
@@ -91,6 +93,33 @@ export const AboutUIModal: React.FC<AboutUIModalProps> = (props: AboutUIModalPro
         <TextListItem component="dd">{additionalInfo}</TextListItem>
       </React.Fragment>
     );
+  };
+  */
+
+  const renderMeshLink = () => {
+    if (config?.about?.mesh) {
+      return (
+        <Button component="a" href={config.about.mesh.url} variant={ButtonVariant.link} isInline>
+          <IstioLogo className={iconStyle} />
+          {config.about.mesh.linkText}
+        </Button>
+      );
+    }
+
+    return null;
+  };
+
+  const renderProjectLink = () => {
+    if (config?.about?.project) {
+      return (
+        <Button component="a" href={config.about.project.url} variant={ButtonVariant.link} target="_blank" isInline>
+          <KialiIcon.Repository className={iconStyle} />
+          {config.about.project.linkText}
+        </Button>
+      );
+    }
+
+    return null;
   };
 
   const renderWebsiteLink = (): JSX.Element | null => {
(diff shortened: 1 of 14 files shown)


ferhoyos (commented):
This PR looks promising @jshaughn! Just some small comments.

Related to the Components section of the About page, I think it would be good to display external services version (at least the Istio and Kiali versions) at the same place along with the Mesh link. That's what we document to determine the version for bugs: https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running.

If this is not accurate in multi-cluster scenarios, do we have to redefine the externalServices API to allow multiple Istio instances?

frontend/src/components/About/AboutUIModal.tsx
```diff
@@ -49,48 +42,35 @@ const textContentStyle = kialiStyle({
...
-    if (tempoService) {
-      tempoService.url = GetTracingUrlProvider(externalServices)?.HomeUrl();
-      return renderComponent(tempoService);
-    } else {
-      return <></>;
+  const renderProjectLink = () => {
+    if (config?.about?.project) {
+      return (
+        <Button component="a" href={config.about.project.url} variant={ButtonVariant.link} target="_blank" isInline>
+          <KialiIcon.Repository className={iconStyle} />
```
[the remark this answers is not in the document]
jshaughn: I see it's available... Done.

frontend/src/pages/Mesh/Mesh.tsx
```diff
@@ -492,40 +485,38 @@ const TopologyContent: React.FC<{
...
-                    icon: <TopologyIcon />,
-                    tooltip: 'Layout - dagre',
-                    callback: () => {
-                      setLayoutName(LayoutName.Dagre);
+                  customButtons: [
+                    {
+                      ariaLabel: 'Layout - Dagre',
+                      id: 'toolbar_layout_dagre',
+                      disabled: LayoutName.Dagre === layoutName,
+                      icon: <TopologyIcon />,
```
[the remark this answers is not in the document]
jshaughn: In this case we seem limited to the react-icon, as that is what PFT seems to want here.  Changing to the KialiIcon fails to render.

jshaughn:
@ferhoyos So, I think this is stable enough to be considered for merge.  It's still hidden behind the feature flag.  I am working on a follow-up [Issue](https://github.com/kiali/kiali/issues/7298)/PR to remove the feature flag and make this the default mesh page.

ferhoyos (changes requested):
Some things I have found:
- Kiali mesh cola layout does not look good on initial render:

image

The layout looks fine when you click on `Reset View` button:

image

Is it possible to execute a `Reset View` in the initial rendering? 

- Mesh layout is not applied correctly from url. For example if I enter `/console/mesh?meshLayout=kiali-mesh-cola` the default layout `Dagre` is applied
- Mesh topology is not adapted to canvas when I show/hide the side panel:

image

Maybe it is needed to reset the view after the side panel is shown/hidden

Minor ones (can be done in a different PR):
- Legend is not working. We should hide the button if the legend is not ready yet.
- Mesh title `TODO: Mesh Name Here` should be modified or hidden
- Can we hide the three dots button if there is no context menu to show in the graph node?

image

jshaughn:
@ferhoyos With the exception of the "fit" issue, which I will defer, I pushed changes for your other comments.  I feel like PFT has issues with resize that may need to be handled in the PFT code itself, because it is not updating the bounds of the canvas in a timely way.

ferhoyos:
@jshaughn I still have issues with the url issue commented above. If I enter `/console/mesh?meshLayout=kiali-mesh-dagre` the default layout `kiali-dagre` is applied. If you agree, this one can be deferred to a different PR because it affects to the PF Traffic graph too. The following code in Cytoscape graph is missing in the PF and mesh graph:

```
// Let URL override current redux state at construction time. Update URL with unset params.
    const urlLayout = HistoryManager.getParam(URLParam.GRAPH_LAYOUT);

    if (urlLayout) {
      if (urlLayout !== props.layout.name) {
        props.setLayout(LayoutDictionary.getLayoutByName(urlLayout));
      }
    } else {
      HistoryManager.setParam(URLParam.GRAPH_LAYOUT, props.layout.name);
    }
```

The rest of the issues are addressed (except the "fit" one, which I am ok to defer to future PR).

One question... Is it possible to check the version of mesh components from the Mesh page? You have removed them from About modal because they are supposed to be included in the Mesh page, but I do not see the version info when you click on Istio or Prometheus node. 

image

frontend/src/components/About/AboutUIModal.tsx
```diff
@@ -166,12 +132,16 @@ export const AboutUIModal: React.FC<AboutUIModalProps> = (props: AboutUIModalPro
           <TextListItem key="kiali-container-version" component="dd">
             {containerVersion!}
           </TextListItem>
-          <TextListItem key="service-mesh-name" component="dt">
-            Service Mesh
-          </TextListItem>
-          <TextListItem key="service-mesh-version" component="dd">
-            {meshVersion!}
-          </TextListItem>
+          {false && (
```
ferhoyos, line 135: Code section to be removed along with the `meshVersion` variable

frontend/src/config/Config.ts
```diff
@@ -68,15 +68,17 @@ const conf = {
   },
   /** About dialog configuration */
   about: {
+    mesh: {
+      url: '/mesh',
+      linkText: 'Visit the Mesh page'
```
ferhoyos, line 73:
```suggestion
      linkText: i18n.t('Visit the Mesh page')
```
Adding i18n support

frontend/src/pages/Mesh/layouts/MeshColaLayout.ts
```diff
@@ -37,7 +37,9 @@ class MeshLayout extends ColaLayout implements Layout {
 
   private childLayouts: BaseLayout[] = [];
 
-  protected initializeLayout(): void {}
+  protected initializeLayout(): void {
+    console.log('init mesh layout');
```
ferhoyos, line 41: Is this comment needed?

jshaughn:
> I still have issues with the url issue commented above. 

@ferhoyos You are right, the URL handling was more broken that I thought.  I have added the suggested URL handling for both mesh and PF graph.

jshaughn:
> One question... Is it possible to check the version of mesh components from the Mesh page

@ferhoyos Oh my gosh, good catch!  In all of the activity I forgot one of the primary pieces of data.  I will add support...

frontend/cypress/integration/featureFiles/overview.feature
```diff
@@ -101,12 +101,6 @@ Feature: Kiali Overview page
     Then there should be a "degraded" application indicator in the namespace
     And the "degraded" application indicator should list the application
 
-  @error-rates-app
```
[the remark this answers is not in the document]
jshaughn: Yes, it will need to migrate.  But I needed to remove it here because the information is no longer on the overview card, it is moved to the control plane node side panel.

frontend/src/components/About/AboutUIModal.tsx
jshaughn: Code removed, and the mesh name (aka mesh version) is now displayed in the default mesh side panel.

frontend/src/pages/Mesh/layouts/MeshColaLayout.ts
jshaughn: Removed.  And currently this whole layout needs work and is not being used/exposed.

jshaughn:
@ferhoyos I think I've taken care of all of the feedback up to this point...

frontend/src/services/MeshDataSource.ts
```diff
@@ -144,12 +154,13 @@ export class MeshDataSource {
     this.promiseRegistry.register(PROMISE_KEY, API.getMeshGraph(restParams)).then(
       response => {
         const responseData: any = response.data;
-        this.meshElements = responseData && responseData.elements ? responseData.elements : EMPTY_MESH_DATA;
-        this.meshTimestamp = responseData && responseData.timestamp ? responseData.timestamp : 0;
+        this.meshElements = responseData?.elements ?? EMPTY_MESH_DATA;
```
ferhoyos, line 157: Much cleaner and more readable code! :-)

ferhoyos (approved):
Code LGTM and all the issues have been addressed. Really good work @jshaughn!

This PR represents a significant improvement in the functionality of the new Mesh page. I believe we're almost ready to showcase the Mesh graph without relying on the Kiali feature flag.

Some suggestions for UI improvements in future PRs:
- Replace `unknown` cluster with the name `External Services`, without any cluster badge.
image

- The same applies to specific services when deployed as external services.
image

- Displaying the configuration as plain JSON doesn't look very good. I suggest presenting it in a table format, similar to what we do in the `View Debug Info` modal:
image

We could also hide the configuration within an expandable table row:
image

- Improvements in the target panel namespaces:
  - IMO card borders inside expandable rows do not look good
  - There is duplication of information: the namespace name is repeated, and we already know that we are in cluster X (shown in the data plane header).
image

Issue #6777, opened 2023-10-25
[e2e] Tests should check status code before attempting to unmarshal into json


nrfox:
Many e2e tests attempt to unmarshal the kiali api response before checking the status code resulting in an unmarshalling error. This can mask the response from the server making it harder to debug the tests. The status code should first be checked before unmarshalling.

See: https://github.com/kiali/kiali/actions/runs/6413559944/job/18045431679 for an example of the error.

PrinsonF77:
Hey, is this issue solved yet?

jshaughn:
hi @PrinsonF77 , I'm going to say no just based on the fact that this is Open, still in the Backlog there are no PRs...   Did you want to contribute here?

PrinsonF77:
Yes. Can you assign it to me?
Also according to what I've seen at https://github.com/kiali/kiali/blob/master/tests/integration/tests/kiali_urls_test.go , 
`func TestGrafana(t *testing.T)` already implements this. Something similar has to be done for the tests that do not do this check right?

nrfox:
Sure the issue is here inside of the `Grafana()` call (and everywhere else this happens): https://github.com/kiali/kiali/blob/9cb961f10a7e7c9ce98765b4094402ee95abaa7b/tests/integration/utils/kiali/kiali_client.go#L605-L616

Until we check the status code, we can't safely unmarshal into `status := new(models.GrafanaInfo)` since the server response for 404s for non 200 is usually just a string. When the unmarshal fails, it masks the actual error message from the server that resides in the response.Body.

What we need is to stop unmarshaling non-200 responses inside of the client calls. It looks like some tests still need to check the status code so I wouldn't return an error here for non-200 responses. It doesn't appear that anything uses the response body directly though so you can simply log the string of `response.Body` for non-200 responses.

PrinsonF77:
Thanks for the explanation. Will start working on it

nrfox:
The kiali client methods can now use this method: https://github.com/kiali/kiali/blob/470b68b0909efc0e1adf75b1675efc20b39f1f42/tests/integration/utils/kiali/kiali_client.go#L647-L663 which will check the status code before unmarshaling. We just need to update the kiali client methods to use this new method.

PrinsonF77:
Hey @jshaughn @nrfox . Will be taking this up. Sorry for the delay.
[...]

PR #7215, opened 2024-03-24
Modifying Kiali Client methods to use getRequestAndUnmarshalInto to avoid masking of errors returned by the server

PrinsonF77:
### Describe the change

Kiali client methods would try to unmarshall response from the server without checking the status code. This would mask the actual error sent by the server and make debugging difficult
This PR modifies Kiali Client methods to use getRequestAndUnmarshalInto to avoid masking of errors returned by the server.

### Steps to test the PR

Run Integration tests for the backend. Changes in this PR are made to the methods used by these tests.

### Automation testing

N/A

### Issue reference

https://github.com/kiali/kiali/issues/6777


PrinsonF77:
@nrfox One thing I noticed while making these changes is that there are some tests, that check for the invalid condition.
For example, 
https://github.com/kiali/kiali/blob/076f52e7a7474d8af7b0e2c8c3a3b5addc60242b/tests/integration/tests/namespaces_test.go#L37C1-L46C2

These tests require that the error returned by the client method is NIL. But modifying the client method to use `getRequestandUnMarshalInto` returns an error on a non-200 error code. Currently, I have modified the kiali client methods that are called by such tests to use the old code, as that does not return an error. But ideally, it looks like a better idea to me, to modify these tests to expect an error to be returned (with a specific code/ message). What are your thoughts on this?

nrfox:
> But ideally, it looks like a better idea to me, to modify these tests to expect an error to be returned (with a specific code/ message). What are your thoughts on this?

That sounds good. You could go a few different ways with that. 1. Return some kind of custom error type from `getRequestAndUnmarshalInto` that has the code/message that the tests can use if they want to do assertions on the response code. 2. Have some lower level client method like `DoRequest` where the tests are more verbose but you have more control over the request and assertions on the response. Either way (or something else that accomplishes the same thing) is fine with me. The ultimate goal of this change is to not hide the underlying error by unmarshaling into a bad response.

nrfox:
Looks like you'll need to rebase as well to fix conflicts

PrinsonF77:
@nrfox Can you please review the PR.

nrfox (approved):
Looks good, thanks for the contribution!

PR #7217, opened 2024-03-25
Workloads lists page optimized to query per cluster

hhovsepy:
### Describe the change

Workloads page was querying backend per namespace, now it queries per cluster by providing selected namespaces.
New API endpoing "api/clusters/workloads" is added.

### Steps to test the PR

Create a set of namespaces:
Open the Workloads list page and select all workloads.
Check if all fields are loaded correctly.

### Automation testing

Modified e2e and unit tests.

### Issue reference

### DIFF
--- a/business/checkers/gateways/selector_checker_test.go
+++ b/business/checkers/gateways/selector_checker_test.go
@@ -157,9 +157,7 @@ func TestValidIstioNamespaceSelector(t *testing.T) {
 		Gateway: gw,
 		WorkloadsPerNamespace: map[string]models.WorkloadList{
 			"test": {
-				Namespace: models.Namespace{
-					Name: "test",
-				},
+				Namespace: "test",
 				Workloads: []models.WorkloadListItem{},
 			},
 		},
--- a/tests/data/workload_data.go
+++ b/tests/data/workload_data.go
@@ -12,7 +12,7 @@ func CreateWorkloadsPerNamespace(namespaces []string, items ...models.WorkloadLi
 
 func CreateWorkloadList(namespace string, items ...models.WorkloadListItem) models.WorkloadList {
 	return models.WorkloadList{
-		Namespace: models.Namespace{Name: namespace},
+		Namespace: namespace,
 		Workloads: items,
 	}
 }


business/workloads.go
```diff
@@ -979,7 +980,7 @@ func (in *WorkloadService) fetchWorkloadsFromCluster(ctx context.Context, cluste
 			Pods:     models.Pods{},
 			Services: []models.ServiceOverview{},
 		}
-		w.Cluster = cluster
+		w.Namespace = models.Namespace{Cluster: cluster, Name: namespace}
```
nrfox, line 983: @jshaughn and I recently touched on this same thing for services. The problem parsing a `Namespace` this way is that you remove all the other `Namespace` fields that exist on the object: https://github.com/kiali/kiali/blob/76c4ec1806180ac7dc5a7874f7a8cb8f1103b0ef/models/namespace.go#L29-L45

For example `namespace.IsAmbient` would return different results than `workload.namespace.IsAmbient` even though they represent the same namespace. Most of the other models just have `namespace` as a string and I think that's all you need in most cases if `cluster` is also on the model. We should be consistent either way.

frontend/src/config/Config.ts
```diff
@@ -124,6 +124,7 @@ const conf = {
       clustersHealth: () => `api/clusters/health`,
       clustersMetrics: () => `api/clusters/metrics`,
       clustersTls: () => `api/clusters/tls`,
+      clustersWorkloads: () => `api/clusters/workloads`,
```
nrfox, line 127: If we're going to continue adding the cluster routes and removing the namespace routes altogether then imo we should include cluster in the route instead of passing it as a query param e.g. `api/clusters/{clusterName}/workloads`. We also need to be sure that whatever is returning the list of clusters that the frontend iterates over to do these requests is stable and accurate and will always return at least something since it is a now a pre-req for doing any of these other requests. Basically this has ended up being a large change to the API and it's one that I support but it does affect the whole app.

nrfox (changes requested):

business/workloads.go
hhovsepy: right, will rework these parts also in apps

frontend/src/config/Config.ts
hhovsepy: The reason of having 'clusterName' as a param is historical, from 'clusterNameFromQuery' method, where the `clusterName` param was not mandatory and had a default value. https://github.com/kiali/kiali/blob/master/handlers/utils.go#L173
But I agree that in future we need to switch to `api/clusters/{clusterName}/[workloads][apps][services]` requests.

business/workloads.go
```diff
@@ -145,11 +145,23 @@ func (in *WorkloadService) GetWorkloadList(ctx context.Context, criteria Workloa
 	)
 	defer end()
 
+	namespace := criteria.Namespace
+	cluster := criteria.Cluster
+
 	workloadList := &models.WorkloadList{
-		Namespace:   models.Namespace{Name: criteria.Namespace, CreationTimestamp: time.Time{}},
+		Namespace:   models.Namespace{Cluster: cluster, Name: namespace, CreationTimestamp: time.Time{}},
```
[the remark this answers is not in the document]
hhovsepy: What if we remove `Namespace` from here at all, it does not make sense anymore as 'WorkloadListItem' contains 'Namespace'?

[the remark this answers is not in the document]
hhovsepy: It seems used in too many places.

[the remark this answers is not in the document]
hhovsepy: comment approached

hhovsepy:
PR is ready for review @nrfox

models/workload.go
```diff
@@ -13,11 +13,24 @@ import (
 	"github.com/kiali/kiali/config"
 )
 
+type ClusterWorkloads struct {
+	// Cluster where the apps live in
+	// required: true
+	// example: east
+	Cluster string `json:"cluster"`
```
nrfox, line 20: Would it make sense to add `Cluster` to `WorkloadList` instead of creating a new model without `Namespace`?

nrfox (approved):

frontend/src/pages/WorkloadList/WorkloadListPage.tsx
```diff
@@ -95,36 +95,40 @@ class WorkloadListPageComponent extends FilterComponent.Component<
...
     const activeFilters: ActiveFiltersInfo = FilterSelected.getSelected();
     const activeToggles: ActiveTogglesInfo = Toggles.getToggles();
-    const namespacesSelected = this.props.activeNamespaces.map(item => item.name);
+    const uniqueClusters = new Set<string>();
 
-    if (namespacesSelected.length !== 0) {
-      this.fetchWorkloads(namespacesSelected, activeFilters, activeToggles, this.props.duration);
+    Object.keys(serverConfig.clusters).forEach(cluster => {
+      uniqueClusters.add(cluster);
+    });
```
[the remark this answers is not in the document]
ferhoyos: And what should Kiali do in that case? Show some kind of warning message?

models/workload.go
hhovsepy: Ideally `WorkloadList` should be gone at some point. Because the idea is to tight Workloads into Cluster.

PR #7224, opened 2024-03-26
Adapt Kiali code to make OSSMC compatible with OCP 4.15

ferhoyos:
### Describe the change

This PR adapts Kiali code to make OSSMC compatible with OCP 4.15:
- Add kiosk redirection to all history modifications (graph context menu, trace tab, etc.)
- Stringify body data when content-type is `application/x-www-form-urlencoded`
- According to [console-dynamic-plugin-sdk](https://github.com/openshift/console/blob/master/frontend/packages/console-dynamic-plugin-sdk/README.md#patternfly-dynamic-modules) README, we should avoid non-index imports for any PatternFly packages. Some of the imports are still non-index since they have to be exported correctly from `patternfly-react` library (https://www.github.com/patternfly/patternfly-react/issues/10155)
- Lint issues

### Steps to test the PR

This PR only affects OSSMC, so there should not be any change from Kiali perspective.

### Automation testing

N/A

### Issue reference

Kiali part of https://github.com/kiali/openshift-servicemesh-plugin/issues/284

### DIFF
--- a/frontend/src/services/Api.ts
+++ b/frontend/src/services/Api.ts
@@ -106,19 +106,8 @@ const getHeaders = (): Partial<AxiosHeaders> => {
   if (apiProxy) {
     return { 'Content-Type': 'application/x-www-form-urlencoded' };
   } else {
-    return { ...loginHeaders };
+    return { 'Content-Type': 'application/json', ...loginHeaders };
   }
-};
-
-/** Create content type correctly for a given request type */
-const getHeadersWithMethod = (method: HTTP_VERBS): Partial<AxiosHeaders> => {
-  let allHeaders = getHeaders();
-
-  if (method === HTTP_VERBS.PATCH) {
-    allHeaders['Content-Type'] = 'application/json';
-  }
-
-  return allHeaders;
 };
 
 const basicAuth = (username: UserName, password: Password): BasicAuth => {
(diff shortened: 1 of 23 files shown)


frontend/src/services/Api.ts
```diff
@@ -106,21 +106,10 @@ const getHeaders = (): Partial<AxiosHeaders> => {
...
-    return { ...loginHeaders };
+    return { 'Content-Type': 'application/json', ...loginHeaders };
   }
 };
 
-/** Create content type correctly for a given request type */
-const getHeadersWithMethod = (method: HTTP_VERBS): Partial<AxiosHeaders> => {
-  let allHeaders = getHeaders();
-
-  if (method === HTTP_VERBS.PATCH) {
```
ferhoyos, line 117: This is not valid because there are POST methods that send JSON body and they should have header `Content-Type: application/json`. One example is `getMetricsStats`. 

```
export const getMetricsStats = (queries: MetricsStatsQuery[]): Promise<ApiResponse<MetricsStatsResult>> => {
  return newRequest<MetricsStatsResult>(HTTP_VERBS.POST, urls.metricsStats, {}, { queries: queries });
};
```

Since all API requests send JSON body (`{}` for empty requests), we can add this header to all requests and avoid HTTP method distinction.

PR #7225, opened 2024-03-26
[v1.73] Add kiosk support to all Kiali links

ferhoyos:
Partial backport of https://github.com/kiali/kiali/pull/7224 (only those changes that affect v1.73 and PF4):

- Add kiosk redirection to all history modifications (graph context menu, trace tab, etc.)
- Stringify body data when content-type is application/x-www-form-urlencoded

### Issue reference

https://github.com/kiali/openshift-servicemesh-plugin/issues/284


jshaughn (approved):
Visual check looks good to me.

PR #7229, opened 2024-03-26
Service list backend request per cluster

hhovsepy:
### Describe the change

In Service list page, there was sending a REST API request to retrieve services per namespace, which was causing to filter per cluster which contains the provided namespace.
Now Optimized to send Service list request to REST API per cluster with providing selected namespaces list.

### Steps to test the PR

Create a set of namespaces:
`for i in {1..200}; do kubectl create ns test-$i; done`
Open the Service list page and select all services.
Check if Health and Configuration fields are loading correctly: There was a bug here which was mirroring one cluster's Configuration to another one when two Services with same name/namespace.

### Automation testing

Added unit tests. Added integration tests.

### Issue reference

### DIFF
--- a/models/service.go
+++ b/models/service.go
@@ -61,6 +61,17 @@ type ServiceOverview struct {
 	Health ServiceHealth `json:"health,omitempty"`
 }
 
+type ClusterServices struct {
+	// Cluster where the services live in
+	// required: true
+	// example: east
+	Cluster string `json:"cluster"`
+	// Services list for namespaces of a single cluster
+	// required: true
+	Services    []ServiceOverview `json:"services"`
+	Validations IstioValidations  `json:"validations"`
+}
+
 type ServiceList struct {
 	Namespace   Namespace         `json:"namespace"`
 	Services    []ServiceOverview `json:"services"`
--- a/models/service.go
+++ b/models/service.go
@@ -73,7 +73,7 @@ type ClusterServices struct {
 }
 
 type ServiceList struct {
-	Namespace   Namespace         `json:"namespace"`
+	Namespace   string            `json:"namespace"`
 	Services    []ServiceOverview `json:"services"`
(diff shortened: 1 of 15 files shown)


hhovsepy:
Ready for review @nrfox

hhovsepy:
@nrfox this PR is rebased and ready for review.

models/service.go
```diff
@@ -61,8 +61,19 @@ type ServiceOverview struct {
...
 }
 
+type ClusterServices struct {
+	// Cluster where the services live in
+	// required: true
+	// example: east
+	Cluster string `json:"cluster"`
+	// Services list for namespaces of a single cluster
+	// required: true
+	Services    []ServiceOverview `json:"services"`
```
ferhoyos, line 71: I see that you have renamed `ServiceOverview` type to `ServiceListItem` in the frontend, but not in the backend. I think it is easier for developers that type names are aligned to avoid confusion.

ferhoyos (approved):
One more great optimization, @hhovsepy!  

In my environment with 200 namespaces:
- Previously, loading the service list required 200 requests, which took around 600 ms:

image

- Now, it is simplified to just 1 request, which takes 180 ms:

image

As a nice to have, in case the user has numerous services and namespaces, I think it would be better to initially display a `Loading services` message instead of `No services found`. Despite the optimization, if the request takes a couple of seconds, the user can think there is an issue with Kiali.

models/service.go
hhovsepy: In frontend `ServiceListItem extends ServiceOverview` by adding few frontend list view specific properties for better ordering.

hhovsepy:
Thanks for the review @ferhoyos , that `No services found` is from VirtualList.tsx by design, and yes a bit confusing. I would leave that for another UX improvements.

PR #7230, opened 2024-03-26
Check Kiali accessible namespaces in wait-for-namespace.sh script

ferhoyos:
### Describe the change

Check the `accessible_namespaces` value in `wait-for-namespace.sh` script to determine if it has to patch the Kiali CR or not to include the test namespace.

### Steps to test the PR

PR is good if the CI tests pass


ScriptingShrimp (approved):
LGTM!

PR #7234, opened 2024-03-27
[hack] skupper script can now install the east-west demo

jmazzitelli:
installing this demo is useful because it shows http metrics
east-west demo from here: https://skupper.io/start/


jshaughn (approved):
LG

PR #7241, opened 2024-03-29
[hack] skupper - put west end of pipe in the mesh

jmazzitelli:
The mesh graph now looks like this (notice: http traffic from frontend to the skupper router; the skupper router is also in the mesh - but only the west end... the east end of the demo is outside of the mesh):

image

### DIFF
--- a/hack/istio/skupper/install-skupper-demo.sh
+++ b/hack/istio/skupper/install-skupper-demo.sh
@@ -193,7 +193,7 @@ minikube_install_basic_demo() {
     if [ "${KIALI_DEV_BUILD}" == "true" ]; then
       local make_build_targets="build build-ui"
     fi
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Here's how you can take out portions of skupper itself from the mesh. I don't know if I want to make this optional as part of the script but for now, for example, take out the skupper prometheus:

```
kubectl -n west patch deployment skupper-prometheus --type=json -p='[{"op": "add", "path": "/spec/template/metadata/labels/sidecar.istio.io~1inject", "value":"false"}]'
```

and the graph will end up like this:

image

There is no incoming requests because even though I am poking the UI with the browser, the sidecar is missing so there is no incoming requests telemetry.

jshaughn (commented):
Skupperific.

jmazzitelli:
To take out all skupper components from the mesh, but leaving the frontend app in the mesh:

```
for d in skupper-prometheus skupper-router skupper-service-controller; do kubectl -n west patch deployment $d --type=json -p='[{"op": "add", "path": "/spec/template/metadata/labels/sidecar.istio.io~1inject", "value":"false"}]'; done
```

image

PR #7244, opened 2024-04-01
[hack] support unique domain names for CRC clusters to enable multiple clusters on the same network

jmazzitelli:
_tl;dr This allows you to use CRC to run multiple OpenShift clusters on a local network. Very useful for multi-cluster work._

This is a very useful enhancement to the CRC hack script.

We already had a command "expose" and "unexpose" which exposes the CRC cluster to remote machines on the local network. This allows you to access the OpenShift cluster via a browser or an oc/kubectl client running anywhere on the network (not just from the same box where CRC is running). This uses firewall rules.

However, the problem is you can only have one CRC cluster on your network because CRC fixes the domain name to "crc.testing". So if I want to start two CRC clusters, one on machine "foo" and one on machine "bar" they both bind to "crc.testing" and therefore only one can be exposed to remote machines on the network.

This PR introduces a new command "changedomain" which changes the domain of a running CRC cluster from the default `crc.testing` to an nip.io hostname unique to the IP address of the host machine (so, for example, `192.168.1.10.nip.io`). This command will also configure HAProxy so remote machines can access the CRC cluster using that new domain name.

If you have the firewalld service already running, this new command will add some firewall rules to open up the ports necessary to access the CRC cluster (technically, it just runs the "expose" command for you at the end).

==

To test:

1. Start CRC normally: `hack/crc-openshift.sh start`
2. Change the domain: `hack/crc-openshift.sh changedomain`
3. Get the IP address of your host machine (e.g. `IP_ADDR=$(hostname -I | awk '{print $1}')`).
4. Log into the OpenShift server using the new domain: `oc login -u kiali -p kiali https://api.${IP_ADDR}.nip.io:6443`
5. Confirm the server you connected to is using that nip.io domain: `oc whoami --show-server`
6. Confirm you can access the cluster: `oc get ns`
7. Now go to another machine on your local network and run that same login command from step 4. Run steps 5 and 6 to confirm it all works.

### DIFF
--- a/hack/crc-openshift.sh
+++ b/hack/crc-openshift.sh
@@ -16,6 +16,7 @@
 #     services: outputs all known service endpoints (excluding internal openshift services)
 #       expose: creates firewalld rules so remote clients can access the cluster
 #     unexpose: removes firewalld rules so remote clients cannot access the cluster
+# changedomain: change the cluster to use an nip.io domain rather than "crc.testing".
 #
 # This script accepts several options - see --help for details.
 #
@@ -680,7 +681,7 @@ The command must be one of:
   * sshoc: Provides a command line prompt with root access inside the CRC VM. Logs in via oc debug.
   * routes: Outputs URLs for all known routes.
   * services: Outputs URLs for all known service endpoints (excluding internal openshift services).
-  * expose: Creates firewalld rules so remote clients can access the cluster.
+  * expose: Creates firewalld rules so remote clients can access the cluster. Do not use in conjunction with changedomain.
   * unexpose: Removes firewalld rules so remote clients cannot access the cluster.
   * changedomain: Changes the CRC cluster base domain name from 'crc.testing' to a unique 'nip.io' name.
 
--- a/hack/crc-openshift.sh
+++ b/hack/crc-openshift.sh
@@ -157,7 +157,7 @@ get_status() {
     fi
     echo "====================================================================="
     echo "kubeadmin password: $(cat ${CRC_KUBEADMIN_PASSWORD_FILE})"
-    echo "$(${CRC_COMMAND} console --credentials)"
+    echo "$(${CRC_COMMAND} console --credentials | sed 's/crc.testing/'${OPENSHIFT_BASE_DOMAIN_NAME}'/')"
     echo "====================================================================="
     echo "To push images to the image repo you need to log in."
     echo "You can use docker or podman, and you can use kubeadmin or kiali user."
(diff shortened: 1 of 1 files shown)


jmazzitelli:
Note that if you use "changedomain", the firewalld service is stopped on the machine (if there was one running).

This also means the "expose" and "unexpose" commands should not be used when the domain is changed. There is no need to use them since "changedomain" will expose the cluster to remote clients.

jmazzitelli:
Some of this work (particularly the part that configured OpenShift with the new base domain name) originated here: https://github.com/iLLeniumStudios/remote-crc-setup/blob/main/install.sh

jmazzitelli:
To test:

1. Start CRC normally: `hack/crc-openshift.sh start`
2. Change the domain: `hack/crc-openshift.sh changedomain`
3. Get the IP address of your host machine (e.g. `IP_ADDR=$(hostname -I | awk '{print $1}')`).
4. Log into the OpenShift server using the new domain: `oc login -u kiali -p kiali https://api.${IP_ADDR}.nip.io:6443`
5. Confirm the server you connected to is using that nip.io domain: `oc whoami --show-server`
6. Confirm you can access the cluster: `oc get ns`
7. Now go to another machine on your local network and run that same login command from step 4. Run steps 5 and 6 to confirm it all works.

jmazzitelli:
One thing I have not tested yet is now this works with our make targets used to push and use dev images. Things like `make cluster-push operator-create kiali-create` need to be tested.

jmazzitelli:
This doesn't fully work. oc CLI can log in and get resources, but it looks like oauth isn't working and you cannot log into the Console UI.

jmazzitelli:
Worst case, I can just have the script shutdown the firewall when changing the domain. But I would like this to work with the firewall enabled... so I need to find what rules to create to get this to work with the firewall running.

jmazzitelli:
Well, after that last commit, now I can't see it fail. It may be due to the removal of the passthroughs, but I don't see why that would be the issue. These were removed:

```
firewall-cmd --direct --passthrough ipv4 -I FORWARD -i ${virt_interface} -j ACCEPT
firewall-cmd --direct --passthrough ipv4 -I FORWARD -o ${virt_interface} -j ACCEPT
```

jmazzitelli:
can't figure out the firewall. I was close, but not there. So the script will disable firewalld if it is running when you use the changedomain command.

Other than that, this all works.

jshaughn (approved):
Fine.

Issue #7058, opened 2024-01-25
Kiali Performance and Scalability testing


hhovsepy:
This is an Epic for all Performance and Scalability  tasks and sub-epics to do for Kiali project.

One of the requirements from Kiali customers is the performance of the Kiali from UI/UX perspective, with a load of many namespaces/workloads/services and a metrics loaded for several days.

TODO: Add requirements and subtasks.

- [x] #7088
- [x] Testing environment preparation.
- [x] #7174
- [ ] #7699
- [x] #7076
- [x] #7097
- [x] #7397
- [ ] #7479

nrfox:
Does this epic cover scalability improvements or is it only for testing?

hhovsepy:
I think we need to include all the improvements in this epic as sub-epics, so right both 1. and 2. points are valid subtasks here.

jmazzitelli:
Now that we have pprof integrated in the Kiali server (to turn it on, set `spec.server.profiler.enabled` to `true` in Kiali CR), it would be a good thing to learn how to use and analyze the pprof data. Here's a blog post on a quick tutorial: https://go.dev/blog/pprof

hhovsepy:
Most tasks of this epic are done. The remaining subtasks are logged as separate issues and can be tracked outside of this epic.

PR #7245, opened 2024-04-02
Script for perf tests run and test data creation

hhovsepy:
### Describe the change

Added a new hack script 'run-performance-tests.sh' which prepares the test data and executes Kiali performance tests.

To run Kiali cypress perf tests now can be done within a single script: `./hack/run-performance-tests.sh -td 10`
where '-td' parameter indicates how many empty 'perf-test-$i' namespaces should be created before testsuite run.

The 'make perf-tests-gui' and 'yarn cypress:perf' endpoints are remaining untouched.

### Issue reference
https://github.com/kiali/kiali/issues/7058

### DIFF
--- a/hack/run-performance-tests.sh
+++ b/hack/run-performance-tests.sh
@@ -114,12 +114,13 @@ deleteData() {
   for ((i = 1; i <= $TEST_DATA; i++)); do
     kubectl delete --ignore-not-found=true -l kiali.io=perf-test ns
   done
+  jq '.allNamespaces = .namespaces' "$COMMON_PARAMS" > "$COMMON_PARAMS.tmp" && mv "$COMMON_PARAMS.tmp" "$COMMON_PARAMS"
 }
 
 ensureCypressInstalled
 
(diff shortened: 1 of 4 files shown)


hack/run-performance-tests.sh
```diff
@@ -0,0 +1,147 @@
...
+    -h|--help)
+      cat <<HELPMSG
+Valid command line arguments:
+  -so|--setup-only <true|false>
+    If true, only setup the test environment and exit without running the tests.
+    Default: false
+  -to|--tests-only <true|false>
+    If true, only run the tests and skip the setup.
+    Default: false
+  -td|--test-data <number>
```
jmazzitelli, line 48: should this be called "test-namespaces" because I can envision in the future we'll have additional test "data' to configure (like "test-apps" or "test-services" or "test-users" or "test-clusters" or...). Calling this "test-data" is not specific enough, especially if we expand this script in the future.
```suggestion
  -tn|--test-namespaces <number>
```

jmazzitelli (approved):

PR #7246, opened 2024-04-02
Setup keycloak and openid auth in multi-primary CI pipeline

### LINKED ISSUE #7021: Add auth to multi-cluster tests in CI pipeline

Frontend tests in CI run in anonymous auth mode. This doesn't allow for automated testing of scenarios where a user has access to cluster A but not cluster B. See this scenario for a concrete example: https://github.com/kiali/kiali/blob/95e4b4975a81ae486aa2badefc102237cd737b30/frontend/cypress/integration/featureFiles/app_details_multicluster_graph.feature#L29-L36

The hack scripts support installing minikube clusters with keycloak acting as an OIDC provider across both clusters. Support can be extended for kind and then run as part of CI.

nrfox:
### Describe the change

Adds keycloak and deploys kiali with `openid` auth in the CI pipeline so that multi-cluster access tests can be performed.

### Steps to test the PR

1. Setup environment
   ```hack/run-integration-tests.sh --test-suite frontend-multi-primary --setup-only true```
2. Visit kiali url and login to keycloak with `kiali:kiali`

### Automation testing

If current automation passes then all is well.

### Issue reference

Fixes #7021

### DIFF
--- a/frontend/cypress/support/commands.ts
+++ b/frontend/cypress/support/commands.ts
@@ -121,7 +121,7 @@ Cypress.Commands.add('login', (username: string, password: string) => {
   const provider = Cypress.env('AUTH_PROVIDER');
 
   cy.window().then((win: any) => {
-    if (auth_strategy !== 'openshift' && auth_strategy !== 'token') {
+    if (auth_strategy === 'anonymous') {
       cy.log('Skipping login, Kiali is running with auth disabled');
       return;
     }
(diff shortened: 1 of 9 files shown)


ScriptingShrimp:
passing on OCP

jmazzitelli (approved):
If pavel says it's good, then it's good :)

PR #7248, opened 2024-04-03
Use latest v2 of go-jose

nrfox:
### Describe the change

Replaces `gopkg.in/square/go-jose.v2` with `"github.com/go-jose/go-jose"` since the package moved to a new org where the latest v2 branch lives.

### Steps to test the PR

Test with an auth strategy that uses go jose like `openid`. You can use this change to easily deploy that setup: https://github.com/kiali/kiali/pull/7246.

### Automation testing

N/A


jshaughn (approved):
Looks OK to me, CI passing.

PR #7251, opened 2024-04-03
[hack] for extension metrics R&D

jmazzitelli:
This is a small demo app (docker container and some k8s resources) that will be useful when doing the extensions metrics work in Kiali graph generator/appenders.

The Makefile provides some basically commands for dev workflow. The idea is we can define a series of metrics timeseries and labels, see how Kiali graphs them. Then we can change the metrics with different names or labels and see how well Kiali graphs those. In other words, this allows us to do quick dev cycles as we try to determine what good extension metrics would look like.

This is going to be useful for the skupper integration work.

Dev cycle would look like this:

1. Edit `k8s.yaml`, specifically the metric definitions in the ConfigMap.
2. Run `make deploy` to get the simple app installed. Prometheus will immediately scrape the endpoints. We can put this in or out of the mesh depending on what we are testing.
3. Look at the metrics within Prometheus UI or the graph within Kiali UI.
4. Make adjustments to the Kiali code.
5. Make adjustments to the metrics in the k8s.yaml ConfigMap
6. Redeploy the simple demo app via `make redeploy` so it picks up the new metrics and starts emitting them to Prometheus.
7. Rinse, repeat.

### DIFF
--- a/hack/istio/skupper/demo-emit-metrics-deployment/Dockerfile
+++ b/hack/istio/skupper/demo-emit-metrics-deployment/Dockerfile
@@ -1,0 +1,9 @@
+# To build:  podman build -t demo-extension-metrics-server:latest .
+# To run:    podman run -it --rm -p 9090:9090 demo-extension-metrics-server:latest
+# To access: curl http://localhost:9090
+
+FROM registry.access.redhat.com/ubi9/ubi:latest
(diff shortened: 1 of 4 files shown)


jshaughn (approved):
Should be fine.

Issue #7252, opened 2024-04-04
token & OpenShift authentication not working 


josunect:
### Describe the bug

When setting token or OpenShift strategy, there is an error in the login: 
Invalid login: Token is empty or invalid. 

### Expected Behavior

User can login with a valid token or username/password

### What are the steps to reproduce this bug?

1. auth strategy should be set to token/OpenShift
2. Try to login with a valid token or username/pwd

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:** latest master
- **Istio version:** 1.21
- **Kubernetes impl:** 
- **Kubernetes version:** 
- **Other notable environmental factors:**

PR #7253, opened 2024-04-04
Update Content-Type for login request

josunect:
Fixes https://github.com/kiali/kiali/issues/7252

### DIFF
--- a/frontend/src/services/Api.ts
+++ b/frontend/src/services/Api.ts
@@ -103,9 +103,12 @@ const loginHeaders = config.login.headers;
 /**  Helpers to Requests */
 
 const getHeaders = (urlEncoded?: boolean): Partial<AxiosHeaders> => {
-  if (apiProxy || urlEncoded) {
+  if (apiProxy) {
     return { 'Content-Type': 'application/x-www-form-urlencoded' };
   } else {
+    if (urlEncoded) {
+      return { 'Content-Type': 'application/x-www-form-urlencoded', ...loginHeaders };
+    }
     return { 'Content-Type': 'application/json', ...loginHeaders };
   }
 };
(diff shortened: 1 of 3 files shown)


frontend/src/services/Api.ts
```diff
@@ -102,8 +102,8 @@ const loginHeaders = config.login.headers;
 
 /**  Helpers to Requests */
 
-const getHeaders = (): Partial<AxiosHeaders> => {
-  if (apiProxy) {
+const getHeaders = (urlEncoded?: boolean): Partial<AxiosHeaders> => {
+  if (apiProxy || urlEncoded) {
```
nrfox, line 106: It looks like this leaves off `...loginHeaders`. Should those be included for both content types?

josunect: I've thought that loginHeaders contain the login information (Once logged in), but let me verify

josunect: Updated

nrfox: I'm wondering if OSSMC should also have those login headers? @ferhoyos?

ferhoyos: I don't think so, as OSSMC leverages of Openshift login. Where `'X-Auth-Type-Kiali-UI': '1'` is used?

Said that, including loginHeaders probably do not affect OSSMC. Maybe you can simplify the code:
```
    if (apiProxy || urlEncoded) {
      return { 'Content-Type': 'application/x-www-form-urlencoded', ...loginHeaders };
    } else {
      return { 'Content-Type': 'application/json', ...loginHeaders };
   }
```

nrfox (approved):

nrfox (approved):

PR #7255, opened 2024-04-04
add automaxprocs to determine gomaxprocs

### LINKED ISSUE #7254: automaxprocs removed from kiali
### Describe the bug

In clusters with large nodes, Kiali creates many more processes than set by our limits.cpu and causes high CPU throttling.

### Expected Behavior

Given #4579 was added, I'd expect Kiali to adjust GOMAXPROCS automatically according to the CPU limits.

It looks like automaxprocs was removed in #6100 without explanation.

antoinedeschenes:
### Describe the change

This adds [automaxprocs](https://github.com/uber-go/automaxprocs) to determine `GOMAXPROCS` automatically in a container according to `resources.limits.cpu`. We're currently seeing excessive throttling when setting a CPU limit.

Originally implemented in #4579 
Removed in #6100 though that doesn't look intentional.

### Steps to test the PR

Setting limits lower than the node CPU count on the container shouldn't cause excessive CPU throttling

Fixes #7254

### DIFF
--- a/kiali.go
+++ b/kiali.go
@@ -35,6 +35,8 @@ import (
 	"strings"
 	"syscall"
 
+	_ "go.uber.org/automaxprocs"
+
 	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/business/authentication"
 	"github.com/kiali/kiali/config"
--- a/go.mod
+++ b/go.mod
@@ -73,6 +73,7 @@ require (
 	github.com/vjeantet/grok v1.0.0 // indirect
 	go.opentelemetry.io/otel/metric v1.22.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.0.0 // indirect
+	go.uber.org/automaxprocs v1.5.3 // indirect
 	golang.org/x/crypto v0.15.0 // indirect
 	golang.org/x/oauth2 v0.13.0 // indirect
 	golang.org/x/sys v0.16.0 // indirect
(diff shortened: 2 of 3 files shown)


jmazzitelli (approved):
I was just going to submit a PR that looks identical to this one :)

LGTM

kiali.go
```diff
@@ -47,6 +47,8 @@ import (
 	"github.com/kiali/kiali/status"
 	"github.com/kiali/kiali/tracing"
 	"github.com/kiali/kiali/util"
+
+	_ "go.uber.org/automaxprocs"
```
jmazzitelli, line 51: actually, as per our coding guidelines, @antoinedeschenes please move this new import inbetween the two sections above.

jmazzitelli (changes requested):

kiali.go
jmazzitelli: Should be this:

```go
...
	"strings"
	"syscall"
	"github.com/kiali/kiali/business"
	"github.com/kiali/kiali/business/authentication"

	_ "go.uber.org/automaxprocs"

	"github.com/kiali/kiali/business"
	"github.com/kiali/kiali/business/authentication"
...
```

antoinedeschenes: done ✅

jmazzitelli:
I'm investigating these massive test failures. I don't think this has anything to do with this PR...

antoinedeschenes:
> @antoinedeschenes please rebase this on the latest master (I'm not sure if that's the problem, I see you are close to the head of master already but.. just in case, get this up to the latest master and we'll try it again) _UPDATE: actually, I think that is the problem -- rebase on latest master should hopefully pass the tests_

update pushed

jmazzitelli (approved):

Issue #7152, opened 2024-02-27
[multi-Cluster AuthorizationPolicy] (KIA0106) when namespace SPIFFY is on remote cluster


michaelalang:
### Describe the bug

with multi-Cluster I noticed one thing in Kiali (NOTE: happens only in multi-Cluster setup)

* one defines it's service (bookinfo app)
* one defines an AuthorizationPolicy granting a SPIFFY of a not configured namespace on the cluster of the Kiali instance
* Kiali will always complain with `KIA0106` even though Istio will grant access accordingly as SPIFFY is handed to the destination namespace

![Kiali-RBAC-SPIFFE-multi-cluster](https://github.com/kiali/kiali/assets/94735640/93fe7ab3-3685-4e15-af3c-f26e3ebb072b)

### Expected Behavior

Kiali could report a warning instead that the SPIFFE cannot be verified due to not being configured locally.

### What are the steps to reproduce this bug?

1. deploy OSSM 2.4.5
2. deploy an AuthorizationPolicy
    ```
    apiVersion: security.istio.io/v1beta1
    kind: AuthorizationPolicy
    metadata:
      name: productpage-viewer
      namespace: bookinfo
    spec:
      rules:
      - from:
        - source:
            principals:
            - cluster.local/ns/west/sa/default
            - cluster.local/ns/east/sa/default
      selector:
        matchLabels:
          app: productpage
    ```
4. Istio will grant permission to access based upon the forwarded SPIFFY headers
    ```
    X-Forwarded-Client-Cert": "By=spiffe://cluster.local/ns/west/sa/default;Hash=c51856eb61f92f6f9225ae2babdde6568a3a995d6f9e0799c26ccbc68e8f84c0;Subject=\"\";URI=spiffe://cluster.local/ns/east/sa/default"
    ```
### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version: 1.65.11 (1ec3a2cb167144d803b7f621fe6eb5ff5d79739f)**
- **Istio version: 2.4.5**
- **Kubernetes impl: OCP4.14** 
- **Kubernetes version: v1.27.10+28ed2d7** 
- **Other notable environmental factors: multi-Cluster setup**

hhovsepy:
Replaced Error message with a Warning message 'KIA0107 - Service Account for this principal found on a remote cluster'. This is a temporary solution and will be released with Kiali v1.84.
[...]

PR #7258, opened 2024-04-05
New KIA0107 for ServiceAccounts on remote cluster

hhovsepy:
### Describe the change
Actual for multicluster mode only. Kiali does not have validations for multicluster configs.
Currently, when ServiceAccount is created on a workload on remote cluster and the access is configured by spiffe: https://spiffe.io/docs/latest/try/getting-started-k8s/
Kiali is complaining on AuthPolicy that the ServiceAccount for the principal is not found. But in fact the SA exists.
Added a support of checking SA existence on remote clusters, and if exists, show warning message only on principals that "Service Account for this principal is on remote cluster"
Screenshot from 2024-04-05 10-47-57

### Steps to test the PR

Create a multicluster environment.

On `east` cluster create a Deployment with 'serviceAccountName', like in a SPIRE described in steps here: https://istio.io/latest/docs/ops/integrations/spire/

On `west` cluster create  AuthPolicy which points to a ServiceAccount on `east` cluster:
```
kind: AuthorizationPolicy
metadata:
  name: productpage-viewer
  namespace: bookinfo
spec:
  rules:
  - from:
    - source:
        principals:
        - cluster.local/ns/east/sa/default
  selector:
    matchLabels:
      app: productpage
```
The AuthorizationPolicy will show a Warning message instead of Error that the ServiceAccount is found on remote cluster.
When referring to a ServiceAccount which does not exit, the original Error message is still shown.

TBD An Epic to support SPIRE Integration to Istio in Kiali as well: https://github.com/kiali/kiali/issues/7261

### Automation testing

Unit tests added.

### Issue reference
https://github.com/kiali/kiali/issues/7152

### DIFF
--- a/business/checkers/authorization/principals_checker.go
+++ b/business/checkers/authorization/principals_checker.go
@@ -12,8 +12,9 @@ import (
 )
 
 type PrincipalsChecker struct {
+	Cluster             string
 	AuthorizationPolicy *security_v1beta.AuthorizationPolicy
-	ServiceAccounts     []string
+	ServiceAccounts     map[string][]string
 }
 
 const (
@@ -56,11 +57,17 @@ func (pc PrincipalsChecker) validateFromField(ruleIdx int, from []*api_security_
 		}
 
 		for i, p := range f.Source.Principals {
-			if !pc.hasMatchingServiceAccount(p) {
-				valid = false
+			if !pc.hasMatchingServiceAccount(pc.ServiceAccounts[pc.Cluster], p) {
 				path := fmt.Sprintf("spec/rules[%d]/from[%d]/source/principals[%d]", ruleIdx, fromIdx, i)
-				validation := models.Build("authorizationpolicy.source.principalnotfound", path)
-				checks = append(checks, &validation)
+				if pc.hasMatchingRemoteServiceAccount(p) {
+					valid = true
+					validation := models.Build("authorizationpolicy.source.principalremote", path)
+					checks = append(checks, &validation)
+				} else {
+					valid = false
+					validation := models.Build("authorizationpolicy.source.principalnotfound", path)
+					checks = append(checks, &validation)
+				}
 			}
 		}
 	}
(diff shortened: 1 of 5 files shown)


josunect (approved):
Without this PR: 
image

With the PR: 
image

Code looks good. 
I think it good be great to clarify why if this is a warning in the docs.

josunect (changes requested):
When the PA is in local cluster but the sa is just in the remote, the PA for the local cluster is not showing any warning: 

Local cluster: 
image

Remote cluster: 
image

sa from east: 
image

sa from west: 
image

The request send the cluster parameter: 
image

If we change it for bookinfo-ratings (In both clusters), the warning is just shown for the local cluster: 
image

hhovsepy:
The issue happens by design, when ServiceAccount Names are loaded from Pod's, not from ServiceAccount objects (which in fact is not integrated into Kiali): Historical code: https://github.com/kiali/kiali/pull/5057/files#diff-d3d5bb00d54bcd1a9a0596fda3687c2019e0135f81a3c9e4f56856b57545bd11R76
Does anybody remember why we avoided loading ServiceAccounts directly from cluster? Some security reasons?

hhovsepy:
@josunect so the Kiali is designed to load SA not directly from K8s but from pods 'ServiceAccountName'.
The correct testing scenario will be to refer AuthPolicy to service account which is mentioned in pods.

hhovsepy:
@jshaughn wdyt to keep this newly added validation "Service Account for this principal is on remote cluster" as a `Warning` or make it an `Info` ? Before we implement Kiali integration with SPIFFY.

josunect (approved):
LGTM, tests added, work as expected

Thanks for clarifying, @hhovsepy !

jshaughn:
> wdyt to keep this newly added validation "Service Account for this principal is on remote cluster" as a Warning or make it an Info ? Before we implement Kiali integration with SPIFFY.

@hhovsepy, sorry, I missed this question.  It seems you chose Warning, and that seems good to me.

Issue #5913, opened 2023-03-14
New Mesh Topology page


jmazzitelli:
Last update [jshaughn, Feb 9, 24]  See original description at bottom of comment

We have decided on an approach and it will be to implement a topology-based Mesh Page, showing the mesh deployment in a graphical way.   It will be natively implemented using Patternfly Topology so that it will be compatible with both Kiali and OSSMC.

**Milestone 1: POC [COMPLETE]**
- [x] [Initial POC](https://github.com/kiali/kiali/issues/6840)
- [x] Initial PR behind feature flag (as part of POC)

**Milestone 2: Replace the current Mesh Page and make available to users**
- [x] Add side-panel for Cluster
- [x] Add side-panel for Istiod
- [x] Add side-panel for Prom // start with common "InfraNode" panel
- [x] Add side-panel for Grafana // start with common "InfraNode" panel
- [x] Add side-panel for Tracing // start with common "InfraNode" panel
- [x] [Fix Fit](https://github.com/kiali/kiali/issues/7122)
- [x] Initial Cypress support (done as part of 7298)
- [x] [Finish initial Find/Hide](https://github.com/kiali/kiali/issues/7117)
- [x] [Improve non-infra namespace box handling](https://github.com/kiali/kiali/issues/7118)
- [x] [Add Initial Health](https://github.com/kiali/kiali/issues/7139)
- [x] [Improve layout/look](https://github.com/kiali/kiali/issues/7120)
- [x] [Remove legacy Mesh Page](https://github.com/kiali/kiali/issues/7298)
- [x] Remove GetClusters API, needed only by legacy Mesh Page (done as part of 7298)
- [x] https://github.com/kiali/kiali/issues/6431
- [x] Update https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running.
- [x] Add/Check Graph Tour (done as part of 7298)

**NOTE - I'm going to close this Epic at this point, given that the page is in decent shape.  We will proceed iteratively with individual issue enhancements.  Some issue are already created, and other ideas are below for reference.**

**Milestone 3: Necessary improvements**
- [control-plane specific data](https://github.com/kiali/kiali/issues/7203)
- [correct multi-cluster istio and cluster versions](https://github.com/kiali/kiali/issues/7361)
- More work on PFT issues like Layout and Resize
- [Add Legend](https://github.com/kiali/kiali/issues/7377)
[...]

PR #7260, opened 2024-04-05
New mesh icons imported from svg files

ferhoyos:
### Describe the change

Existing mesh icons support only one colour. This PR allows the use of svg files as mesh icons, being able to use multi-colour icons for the mesh topology.

Light mode:
image

Dark mode:
image

I have also added two new icons:
- Tempo for tracing.

image

- Thanos for metrics storage. I think this icon needs extra work to be available since prometheus external service does not contain thanos configuration.

image 

### Steps to test the PR

1- Enable Mesh page in the server config
2 - Navigate to the Mesh page
3 - Verify the new icons

### Automation testing

Not applicable as it only affects look and feel

### Issue reference

https://github.com/kiali/kiali/issues/5913


jshaughn (approved):
This is great! So much better!  Maybe we can fix up the Kiali logo in a follow-up, the SVG is not very good.

PR #7268, opened 2024-04-09
Visualize ztunnel logs for pod

### LINKED ISSUE #6898: Ambient - Envoy logs
How Envoy logs will be visualized in Kiali? 

Example of ztunnel logs from the istio-system namespaces: 
image

- It is possible to associate these logs with a specific workload? It is possible to filter? 
- The performance will be good? 
- Why is shown as istio-proxy? 

Waypoint proxy logs: 

image

josunect:
### Describe the change

Visualize the ztunnel logs, filtering the ztunnel logs for the IP of the pod service: 

image

image

ztunnel workload logs: 

image

Updated the istio-proxy name: 

image

### Steps to test the PR

Create a Ambient cluster: 
- `istio/download-istio.sh -div 1.22-dev`
- Install Istio: `istio/install-istio-via-istioctl.sh -c kubectl -cp ambient -it default -ih default`
- Install Kiali
- Install bookinfo: `istio/install-bookinfo-demo.sh -c kubectl -ai false -tg` 

### Automation testing

Added backend test coverage

### Issue reference

Fixes #6898 
Ref. https://github.com/istio/ztunnel/pull/874

Note: The logs format has changed in Istio 1.22. This PR is compatible with the new version. Should be tested with Istio 1.22.


business/test_util.go
```diff
@@ -722,6 +722,14 @@ func FakePodLogsProxy() *kubernetes.PodLogs {
...
+2024-04-12T10:17:44.081519Z	info	access	connection complete	src.addr=10.244.0.5:45896 src.workload="istio-ingressgateway-cbdcbb75c-9q9nb" src.namespace="istio-system" src.identity="spiffe://cluster.local/ns/istio-system/sa/istio-ingressgateway-service-account" dst.addr=10.244.0.16:9080 dst.hbone_addr="10.244.0.16:9080" dst.service="productpage.bookinfo.svc.cluster.local" dst.workload="productpage-v1-87d54dd59-fzflt" dst.namespace="productpage" dst.identity="spiffe://cluster.local/ns/bookinfo/sa/bookinfo-productpage" direction="inbound" bytes_sent=5483 bytes_recv=1551 duration="8ms"
+2024-04-12T10:31:51.078103Z	info	access	connection complete	src.addr=10.244.0.16:51748 src.workload="productpage-v1-87d54dd59-fzflt" src.namespace="bookinfo" src.identity="spiffe://cluster.local/ns/bookinfo/sa/bookinfo-productpage" dst.addr=10.244.0.11:15008 dst.service="details.bookinfo.svc.cluster.local" dst.workload="details-v1-cf74bb974-wg44w" dst.namespace="details" dst.identity="spiffe://cluster.local/ns/bookinfo/sa/bookinfo-details" direction="outbound" bytes_sent=200 bytes_recv=358 duration="1ms"`,
```
nrfox, line 729: This is probably better as a test file that also includes log lines from other workloads as well.

business/workloads.go
```diff
@@ -99,10 +99,12 @@ type LogEntry struct {
 
 // LogOptions holds query parameter values
 type LogOptions struct {
-	Duration *time.Duration
-	IsProxy  bool // fetching logs for Istio Proxy (Envoy access log)
-	MaxLines *int
+	Duration  *time.Duration
+	IsProxy   bool // fetching logs for Istio Proxy (Envoy access log)
+	IsZtunnel bool
```
nrfox, line 104: This might be better as something like an "enum" now e.g. `LogType = "Proxy"` or something.

business/workloads.go
```diff
@@ -2034,11 +2124,21 @@ func (in *WorkloadService) streamParsedLogs(cluster, namespace, name string, opt
 			break
 		}
 
-		entry := parseLogLine(line, opts.IsProxy, engardeParser)
```
nrfox, line 2037: I think maybe this PR could be greatly simplified by providing a new regex pattern to this line:
```
engardeParser := parser.New(parser.IstioProxyAccessLogsPattern)
```
Something like: `parser.New(parser.ZTunnelLogsPattern)`?

business/workloads.go
```diff
@@ -2034,11 +2124,21 @@ func (in *WorkloadService) streamParsedLogs(cluster, namespace, name string, opt
...
+			entry = parseLogLine(line, opts.IsProxy, engardeParser)
+		}
+
 		if entry == nil {
 			continue
 		}
 
+		if opts.filter != nil && !filterMatches(entry.Message, opts.filter) {
+			continue
+		}
```
nrfox, line 2140: We've already parsed out the entry. It might be better to filter by entry fields now like `entry.AccessLog.UpstreamCluster`. I worry we'll get false positives doing simple string matching.

You could have one app named `productpage` and another named `longerproductpage` and they both would match doing simple string matching on the whole line.

business/workloads.go
josunect: Well, this is really from an external library so, we should create a parser and a parse function in our code, as this library doesn't include a ztunnel parser yet.

nrfox: Ah for some reason I didn't realize this was from an external library.

josunect: But right maybe would be better to try to use a regex or similar

business/workloads.go
josunect: Updated

business/test_util.go
josunect: Updated

business/test_util.go
```diff
@@ -722,6 +726,16 @@ func FakePodLogsProxy() *kubernetes.PodLogs {
 	}
 }
 
+func FakePodLogsZtunnel() *kubernetes.PodLogs {
+	content, err := readFile("../tests/data/logs/ztunnel.log")
+	if err != nil {
+		log.Errorf("Error reading logs file: %s", err.Error())
+	}
```
nrfox, line 733: nit: you should be able to just embed the file like this: https://github.com/kiali/kiali/blob/6bbef003ec097d104a9120588841eada9937fd75/kubernetes/client_factory_test.go#L28-L29

business/workloads.go
```diff
@@ -2104,5 +2216,35 @@ func (in *WorkloadService) streamParsedLogs(cluster, namespace, name string, opt
...
+	// First, get ztunnel namespace and containers
+	pods := in.cache.GetZtunnel(cluster)
+	opts.PodLogOptions.Container = models.IstioProxy
+
+	// The ztunnel line should include the pod and the namespace
+	fs := filterOpts{
+		destWk: fmt.Sprintf("dst.workload=\"%s\"", name),
+		destNs: fmt.Sprintf("dst.namespace=\"%s\"", namespace),
+		srcWk:  fmt.Sprintf("src.workload=\"%s\"", name),
+		srcNs:  fmt.Sprintf("src.namespace=\"%s\"", namespace),
```
nrfox, line 2234: This might need to include `dst.cluster` or `src.cluster` if ambient beings to support multi-cluster but I think this is fine for now for single cluster.

kubernetes/cache/cache.go
```diff
@@ -204,6 +207,48 @@ func (in *kialiCacheImpl) IsAmbientEnabled(cluster string) bool {
 	return check
 }
 
+// IsAmbientEnabled checks if the istio Ambient profile was enabled
+// by checking if the ztunnel daemonset exists on the cluster.
+func (in *kialiCacheImpl) GetZtunnel(cluster string) models.PodsList {
```
nrfox, line 212: nit: the comment is for a different function here.

Why not just return the `ztunnel` pods themselves?

nrfox (approved):
A couple minor comments but non-blocking

nrfox:
@josunect tests are failing because you need to include the add test file: https://github.com/kiali/kiali/pull/7268/files#diff-d9dc0e72779a0531aaf6644f0410ce3f7c6c58f1ad934b52983c6354bac66ed8R730

nrfox (dismissed):

business/workloads.go
```diff
@@ -599,26 +587,78 @@ func parseZtunnelLine(line string) *LogEntry {
 		Severity:      "INFO",
 	}
 
+	regex := regexp.MustCompile(models.ZtunnelRegexLine)
```
nrfox, line 590: You should pre-compile regexes by just having them be globals like this: https://github.com/kiali/kiali/blob/6bbef003ec097d104a9120588841eada9937fd75/status/versions.go#L28-L32.

business/workloads.go
```diff
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"reflect"
```
nrfox, line 10: Generally using `reflect` is known to be a performance hit and considering we're parsing logs here it could have an impact. Is there another way to accomplish the same thing without reflect?

business/workloads.go
```diff
@@ -599,26 +587,78 @@ func parseZtunnelLine(line string) *LogEntry {
...
+		prop = models.Ztunnel
+		if len(matches) < 18 {
+			regex = regexp.MustCompile(models.ZtunnelError)
+			matches = regex.FindStringSubmatch(line)
+			prop = models.ErrorLine
+			if len(matches) < 16 {
+				regex = regexp.MustCompile(models.ZtunnelSimpleError)
+				matches = regex.FindStringSubmatch(line)
+				prop = models.SimpleError
+				if len(matches) < 4 {
```
nrfox, line 607: I wonder if you need 4 different regexs? I don't have much experience with writing regexps but it seems like you could have a capture group for each prop that is optional in case it doesn't exist?

models/workload.go
```diff
@@ -39,6 +39,47 @@ type WorkloadList struct {
...
+const ZtunnelInboundRegexLine = `(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6,9}Z(?:\s+\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6,9}Z)*)\t+(\w+)\t+(\w+)\t+(\w+\s*\w+)\t+src\.addr=(\S+)\s+src\.workload="([^"]+)"\s+src\.namespace="([^"]+)"\s+src\.identity="([^"]+)"\s+dst\.addr=(\S+)\s+dst\.service="([^"]+)"\s+dst\.workload="([^"]+)"\s+dst\.namespace="([^"]+)"\s+dst\.identity="([^"]+)"\s+direction="([^"]+)"\s+bytes_sent=(\d+)\s+bytes_recv=(\d+)\s+duration="([^"]+)"`
+const ZtunnelError = `(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6,9}Z(?:\s+\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6,9}Z)*)\t+(\w+)\t+(\w+)\t+(\w+\s*\w+)\t+src\.addr=(\S+)\s+src\.workload="([^"]+)"\s+src\.namespace="([^"]+)"\s+src\.identity="([^"]+)"\s+dst\.addr=(\S+)\s+dst\.service="([^"]+)"\s+dst\.workload="([^"]+)"\s+dst\.namespace="([^"]+)"\s+dst\.identity="([^"]+)"\s+direction="([^"]+)"\s+duration="([^"]+)"\s+error="([^"]+)"`
+const ZtunnelSimpleError = `(\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6,9}Z(?:\s+\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}\.\d{6,9}Z)*)\t+(\w+)\t+([\w:{}=-]+)\t+([\w:{}=,()\s]+)`
```
nrfox, line 67: I'm not the right person to review this :)

nrfox (commented):
What's the speed difference between the two approaches? It would probably be pretty straightforward to write a go benchmark test for this.

business/workloads.go
```diff
@@ -556,6 +572,126 @@ func parseLogLine(line string, isProxy bool, engardeParser *parser.Parser) *LogE
...
 
 	// override the timestamp with a simpler format
+	timestamp := parseTimestamp(parsedTimestamp)
+	entry.Timestamp = timestamp
+	entry.TimestampUnix = parsedTimestamp.UnixMilli()
+
+	return &entry
+}
+
+func parseZtunnelLine(line string) *LogEntry {
```
jshaughn, line 582: This is just a fundamental question... It looks like we are trying to treat these ztunnel logs to be handled like an Envoy access log.  Do we really need to do this parsing?  Or should be just treat them like "app" logs, just text lines?  Or maybe just try and parse out the timestamp?

business/workloads.go
```diff
@@ -2104,5 +2247,35 @@ func (in *WorkloadService) streamParsedLogs(cluster, namespace, name string, opt
...
-	return in.streamParsedLogs(cluster, namespace, name, opts, w)
+	return in.streamParsedLogs(cluster, namespace, name, opts, w, false)
+}
+
+// StreamZtunnelLogs streams pod logs to an HTTP Response given the provided options
+func (in *WorkloadService) StreamZtunnelLogs(cluster, namespace, name string, opts *LogOptions, w http.ResponseWriter) error {
+
+	// First, get ztunnel namespace and containers
+	pods := in.cache.GetZtunnelPods(cluster)
+	opts.PodLogOptions.Container = models.IstioProxy
```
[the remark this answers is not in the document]
josunect: This is what the rest client accepts, when setting other container it gives the following error: 
`container ztunnel is not valid for pod ztunnel-l6chs`

https://github.com/josunect/kiali/blob/master/kubernetes/kubernetes.go#L391

business/workloads.go
josunect: I've change the line processing, hopefully this approach would be more efficient

business/workloads.go
josunect: Added a new filter

business/workloads.go
josunect: We can create a follow up issue so we don't forget? Or maybe just a TODO

models/workload.go
josunect: I've simplified all of this

business/workloads.go
```diff
@@ -97,12 +98,27 @@ type LogEntry struct {
 	AccessLog     *parser.AccessLog `json:"accessLog,omitempty"`
 }
 
+type LogType string
+
+const (
```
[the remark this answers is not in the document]
josunect: That was updated.

handlers/workloads.go
```diff
@@ -282,13 +282,25 @@ func PodLogs(w http.ResponseWriter, r *http.Request) {
 		queryParams.Get("container"),
 		queryParams.Get("duration"),
 		queryParams.Get("isProxy"),
+		queryParams.Get("isZtunnel"),
```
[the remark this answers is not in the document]
josunect: That was updated.

frontend/src/types/IstioObjects.ts
```diff
@@ -253,6 +254,7 @@ export interface PodLogsQuery {
   container?: string;
   duration?: string;
   isProxy?: boolean;
+  isZtunnel?: boolean;
```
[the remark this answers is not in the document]
josunect: That was updated.

handlers/workloads.go
```diff
@@ -282,13 +282,25 @@ func PodLogs(w http.ResponseWriter, r *http.Request) {
...
+	if ztunnel == "true" {
+		// Fetch pod logs
+		err = business.Workload.StreamZtunnelLogs(cluster, namespace, pod, opts, w)
+		if err != nil {
+			handleErrorResponse(w, err)
+			return
+		}
+		return
+	}
+
```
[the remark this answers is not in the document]
josunect: That was updated.

kubernetes/cache/cache.go
```diff
@@ -62,6 +64,8 @@ type KialiCache interface {
 	// IsAmbientEnabled checks if the istio Ambient profile was enabled
 	// by checking if the ztunnel daemonset exists on the cluster.
 	IsAmbientEnabled(cluster string) bool
+
+	GetZtunnelPods(cluster string) []v1.Pod
```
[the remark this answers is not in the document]
josunect: That was updated.

business/workloads.go
```diff
@@ -1970,7 +2103,7 @@ func (in *WorkloadService) GetWorkloadAppName(ctx context.Context, cluster, name
 // streamParsedLogs fetches logs from a container in a pod, parses and decorates each log line with some metadata (of possible) and
 // sends the processed lines to the client in JSON format. Results are sent as processing is performed, so in case of any error when
 // doing processing the JSON document will be truncated.
-func (in *WorkloadService) streamParsedLogs(cluster, namespace, name string, opts *LogOptions, w http.ResponseWriter) error {
+func (in *WorkloadService) streamParsedLogs(cluster, namespace, name string, opts *LogOptions, w http.ResponseWriter, ambient bool) error {
```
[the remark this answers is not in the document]
josunect: That was updated.

business/workloads.go
josunect: As an improvement, we could create a custom zTunnel logs modal with all the parsed information specific to the ztunnel

jshaughn: I'm OK with the parsing as long as it's efficient enough, and reliable.  I was mainly just asking.

business/workloads.go
```diff
@@ -556,6 +564,103 @@ func parseLogLine(line string, isProxy bool, engardeParser *parser.Parser) *LogE
...
+
+	entry.Message = msgSplit[4]
+	if entry.Message == "" {
+		log.Debugf("Skipping empty log line [%s]", line)
+		entry.Message = line
+		return &entry
+	}
+
+	// k8s promises RFC3339 or RFC3339Nano timestamp, ensure RFC3339
+	// Split by blanks, to get the miliseconds for sorting, try RFC3339Nano
```
jshaughn, line 604:
```suggestion
	// Split by blanks, to get the milliseconds for sorting, try RFC3339Nano
```

josunect:
Running locally it doesn't seem inefficient, looking a some times for the request, it takes just some ms: 

image

models/workload.go
```diff
@@ -39,6 +39,15 @@ type WorkloadList struct {
 	Validations IstioValidations `json:"validations"`
 }
 
+type LogType string
+
+const (
+	AppLog      LogType = "app"
+	ProxyLog    LogType = "proxy"
+	WaypointLog LogType = "waypoint"
+	ZtunnelLog  LogType = "ztunnel"
```
jshaughn, line 48: This is a totally subjective comment, so please
 feel free to ignore.  I like to name these more like an enum, so LogTypeApp, LogTypeProxy, etc...

business/workloads.go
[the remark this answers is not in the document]
jshaughn: OK, thanks for the explanation!

business/workloads.go
```diff
@@ -556,6 +564,103 @@ func parseLogLine(line string, isProxy bool, engardeParser *parser.Parser) *LogE
...
+	splitted := strings.SplitN(line, " ", 2)
+	if len(splitted) != 2 {
+		log.Debugf("Skipping unexpected log line [%s]", line)
+		return nil
+	}
+
+	msgSplit := strings.Split(line, "\t")
+
+	if len(msgSplit) < 4 {
+		log.Debugf("Error splitting line line [%s]", line)
```
nrfox, line 591:
```suggestion
		log.Debugf("Error splitting log line [%s]", line)
```

business/workloads.go
```diff
@@ -556,6 +564,103 @@ func parseLogLine(line string, isProxy bool, engardeParser *parser.Parser) *LogE
...
+
+	// Process some access log data
+	// More validations can be done. Data is in format direction=outbound
+	// Also, more data could be added?
+	al := parser.AccessLog{}
+	al.Timestamp = timestamp
+	if len(msgSplit) > 4 {
+		accessLog := strings.Split(msgSplit[4], " ")
+		for _, field := range accessLog {
+			parsed := strings.Split(field, "=")
```
nrfox, line 633: If one of the values can have an `=` in it (possibly `error`?) then this might be problematic.

nrfox (approved):

business/workloads.go
josunect: I've changed for `strings.SplitN(field, "=", 2)`. It is less probable that = is in the left side, so at least just the first "=" is split.

nrfox (approved):

Issue #7237, opened 2024-03-28
False positive assertion in Cypress test related to Istio config validation


matejnesuta:
### Describe the bug

https://github.com/kiali/kiali/blob/d0b8a331c4620f0178eb3ba5706e7c6a1813f727/frontend/cypress/integration/common/istio_config.ts#L522

This specific command passes even though the health of the object is different then asserted. This command is also used in multiple tests. 

![image](https://github.com/kiali/kiali/assets/45061792/f558343a-e1c4-44a2-b0b1-fe096f5b5083)

matejnesuta:
I tried to reproduce the issue today, but no luck

matejnesuta:
I am not sure why, but on some occasions, the assertion completely skips the inner block of code starting at line 526 and passes even though it should not pass.

PR #7269, opened 2024-04-10
Flaky assertion fix

matejnesuta:
### Describe the change

Related to #7237.

I have not found a reliable way to reproduce the failure, but I have encountered it multiple times both on upstream OCP and Kind. Occasionally, the code inside the Mocha hook was not executed at all and the test passed even though it should detect a failure. This piece of code is tied to most of the CRD validation tests in Cypress.

### Steps to test the PR

Create a minimal Istio gateway in Kiali and insert this scenario in the `istio_config.feature` file:
```
  @bookinfo-app
  @selected
  Scenario: delete me later
    When user selects the "bookinfo" namespace
    Then the "foo" "Gateway" of the "bookinfo" namespace should have a "danger"
```
Run the `yarn cypress:selected` command and the test should fail.

When doing the same procedure against the master branch, there is a chance that the scenario will pass instead of failing.

### Issue reference
#7237

### DIFF
--- a/frontend/cypress/integration/common/istio_config.ts
+++ b/frontend/cypress/integration/common/istio_config.ts
@@ -522,19 +522,19 @@ Then('the AuthorizationPolicy should have a {string}', function (healthStatus: s
 Then(
   'the {string} {string} of the {string} namespace should have a {string}',
   (crdInstanceName: string, crdName: string, namespace: string, healthStatus: string) => {
-      cy.request('GET', `${Cypress.config('baseUrl')}/api/istio/config?refresh=0`);
-      cy.get('[data-test="refresh-button"]').click();
-      ensureKialiFinishedLoading();
+    cy.request('GET', `${Cypress.config('baseUrl')}/api/istio/config?refresh=0`);
+    cy.get('[data-test="refresh-button"]').click();
+    ensureKialiFinishedLoading();
 
-      cy.get(`[data-test=VirtualItem_Ns${namespace}_${crdName.toLowerCase()}_${crdInstanceName}] span.pf-v5-c-icon`, {
-        timeout: 40000
-      })
-        .should('be.visible')
-        .hasCssVar('color', `--pf-v5-global--${healthStatus}-color--100`);
+    cy.get(`[data-test=VirtualItem_Ns${namespace}_${crdName.toLowerCase()}_${crdInstanceName}] span.pf-v5-c-icon`, {
+      timeout: 40000
+    })
+      .should('be.visible')
+      .hasCssVar('color', `--pf-v5-global--${healthStatus}-color--100`);
   }
 );
 
-After({ tags: '@istio-page and @crd-validation' }, () => {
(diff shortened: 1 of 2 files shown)


matejnesuta:
KIA0505 validation is a flaky test, but the other two scenarios are failing consistently on my local Kind.

matejnesuta:
The cleanup hook was not working, so a fix of that will hopefully resolve some of the flakes. KIA1102 is probably a bug (see #7275).

frontend/cypress/integration/common/istio_config.ts
```diff
@@ -522,21 +522,19 @@ Then('the AuthorizationPolicy should have a {string}', function (healthStatus: s
 Then(
   'the {string} {string} of the {string} namespace should have a {string}',
   (crdInstanceName: string, crdName: string, namespace: string, healthStatus: string) => {
-    it('loading config list', { retries: 3 }, () => {
```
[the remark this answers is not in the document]
matejnesuta: I replaced the original code with my own mechanism using recursion. The cy.wait() interval can be tweaked further.

frontend/cypress/integration/common/istio_config.ts
```diff
@@ -519,31 +519,56 @@ Then('the AuthorizationPolicy should have a {string}', function (healthStatus: s
...
+      if (dataTestAttr !== null) {
+        if (dataTestAttr.value === `VirtualItem_Ns${namespace}_${crdName.toLowerCase()}_${crdInstanceName}`) {
+          found = true;
+        }
+      }
+    })
+    .then(() => {
+      if (!found) {
+        cy.wait(40000);
+        waitUntilConfigIsVisible(attempt - 1, crdInstanceName, crdName, namespace);
```
ScriptingShrimp, line 543: @matejnesuta does the tests eventually stop looping if there is error?

frontend/cypress/integration/common/istio_config.ts
```diff
@@ -519,31 +519,56 @@ Then('the AuthorizationPolicy should have a {string}', function (healthStatus: s
...
+  if (attempt === 0) {
+    return;
+  }
+  cy.request('GET', `${Cypress.config('baseUrl')}/api/istio/config?refresh=0`);
+  cy.get('[data-test="refresh-button"]').click();
+  ensureKialiFinishedLoading();
+  let found = false;
+  cy.get('tr')
+    .each($row => {
+      console.log($row);
```
hhovsepy, line 532: just remove the log, otherwise lgtm, thank you @matejnesuta

hhovsepy (approved):

frontend/cypress/integration/common/istio_config.ts
matejnesuta: Hello. Yes, there is a condition at the beginning of the function.

ScriptingShrimp: oh I see now, thanks!

ScriptingShrimp (commented):
LGTM

PR #7271, opened 2024-04-11
Edit existing Telemetry resource when found instead of creating new one

FilipB:
### Describe the change
Updating a hack script to edit existing Telemetry resource when found instead of creating new one

### DIFF
--- a/hack/use-openshift-prometheus.sh
+++ b/hack/use-openshift-prometheus.sh
@@ -216,6 +216,12 @@ EOM
   fi
 
   echo "Apply Telemetry resource to [${ISTIO_NAMESPACE}] to ensure Prometheus is a metrics provider"
+  tel_name=$(${OC} get telemetry -n ${ISTIO_NAMESPACE} -o name)
+  if [ ! -z "$tel_name" ]
+  then
+    echo "Found existing ${tel_name} in ${ISTIO_NAMESPACE}, patching..."
+    ${OC} patch ${tel_name} -n ${ISTIO_NAMESPACE} --type=merge -p '{"spec":{"metrics":[{"providers":[{"name":"prometheus"}]}]}}'
+  else
   cat <<EOM | ${OC} apply -n ${ISTIO_NAMESPACE} -f -
 apiVersion: telemetry.istio.io/v1alpha1
 kind: Telemetry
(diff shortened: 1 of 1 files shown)


hack/use-openshift-prometheus.sh
```diff
@@ -216,6 +216,12 @@ EOM
   fi
 
   echo "Apply Telemetry resource to [${ISTIO_NAMESPACE}] to ensure Prometheus is a metrics provider"
+  tel_name=$(${OC} get telemetry -n ${ISTIO_NAMESPACE} -o name)
+  if [ ! -z "$tel_name" ]
+  then
+    echo "Found existing ${tel_name} in ${ISTIO_NAMESPACE}, patching..."
+    ${OC} patch ${tel_name} -n ${ISTIO_NAMESPACE} --type=merge -p '{"spec":{"metrics":[{"providers":[{"name":"prometheus"}]}]}}'
+  else
   cat <<EOM | ${OC} apply -n ${ISTIO_NAMESPACE} -f -
```
jmazzitelli, line 225: This is to be part of the `else` clause, correct? Needs to be indented.

```suggestion
    cat <<EOM | ${OC} apply -n ${ISTIO_NAMESPACE} -f -
```

hack/use-openshift-prometheus.sh
```diff
@@ -216,6 +216,12 @@ EOM
   fi
 
   echo "Apply Telemetry resource to [${ISTIO_NAMESPACE}] to ensure Prometheus is a metrics provider"
+  tel_name=$(${OC} get telemetry -n ${ISTIO_NAMESPACE} -o name)
+  if [ ! -z "$tel_name" ]
+  then
+    echo "Found existing ${tel_name} in ${ISTIO_NAMESPACE}, patching..."
```
jmazzitelli, line 222: Wrap in brackets the dynamic text to be displayed so it is easy to see in the output what is static text and what is dynamic.
```suggestion
    echo "Found existing [${tel_name}] in [${ISTIO_NAMESPACE}], patching..."
```

jmazzitelli (changes requested):
some minor things. otherwise looks good.

jmazzitelli (approved):
lgtm

PR #7283, opened 2024-04-18
Fix sleep yaml for s390x

matejnesuta:
I added the sed command before applying the YAML and I also added a regex check to make sure the script can be run multiple times.

This needs to be backported to 1.73.


jshaughn (approved):
LG

PR #7286, opened 2024-04-18
[hack] use :z when mounting the skupper metrics file

jmazzitelli:
I want this to work with podman as well as docker. Podman gives me a permission denied when trying to read the mounted file. Use :z so it can work around this permissions issue.

test: `make -e DORP=podman run`

see that you do not get the error `cat: /tmp/metrics.txt: Permission denied`


jshaughn (approved):
ok

Issue #7287, opened 2024-04-19
False KIA1102 alert


im-Amitto:
### Describe the bug
Kiali is flagging VS with KIA1102 even when the gateway do exist. 
FYI, the virtual service shown in SS is working completely fine in the cluster. 

<img width="1292" alt="Screenshot 2024-04-19 at 2 20 13 PM" src="https://github.com/kiali/kiali/assets/29653282/1127a9da-78b4-4d1c-b3d9-caf0920ba0b4">

### Expected Behavior
1. Gateway do exist so the error flagged is wrong in nature.

### What are the steps to reproduce this bug?

1. Create a gateway
3. Create a virtual service in a different namespace and map it that gateway

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:** v1.82.0 (935f83feb847c38f4f7de7110c7ee441a3754618)
- **Istio version:** 1.17.3
- **Kubernetes impl:**  EKS
- **Kubernetes version:**  v1.23.17-eks-7a52527
- **Other notable environmental factors:**

im-Amitto:
TBH i am still investigating it and feels like i might have done something wrong on permission side when moving to kiali operator from normal helm chart. 
Anything that i should look for?

jshaughn:
cc @hhovsepy

hhovsepy:
The issue is fixed and will be available with Kiali v1.84 release.

jshaughn:
@hhovsepy I added the backport-needed label here, and moved the backport-completed from the PR https://github.com/kiali/kiali/pull/7296.

hhovsepy:
Backported to Kiali v1.82

PR #7296, opened 2024-04-23
Fixed validations for details, load configs for all namespaces

hhovsepy:
### Describe the change

There was a regression in Istio config details validations that only local namespace objects are loaded, now it loads config objects for all namespaces.

### Steps to test the PR

1. Create a gateway
2. Create a virtual service in a different namespace and map it that gateway

### Automation testing

Added unit and cypress tests.

### Issue reference
https://github.com/kiali/kiali/issues/7287


josunect (approved):
LG!

Tested and work as expected: 

image

PR #7300, opened 2024-04-24
Don't use kiali login headers with api proxy (OSSMC)

ferhoyos:
### Describe the change

OSSMC shows weird CORS issues because Kiali login header is present in the HTTP requests to the Kiali server. This Kiali login header can be removed for those requests performed with an API proxy like OSSMC because it has a different login process. 

### Steps to test the PR

This PR does not affect Kiali standalone, only OSSMC. Kiali should work as expected.


josunect (commented):
Tested and LG!

Maybe we can add a small comment to prevent issues in the future? Something like this?

// apiProxy is used by OSSMC. loginHeaders cause CORS issues

josunect (approved):
LG!

PR #7308, opened 2024-04-25
Show the health status of mesh infra nodes

### LINKED ISSUE #7139: [mesh page] Add Infra health status
Add Kiali infrastructure health status to Mesh page to be able to show in this graph problems with any of the components (Istio, Grafana, Jaeger, etc.).

ferhoyos:
### Describe the change

Show the health status of mesh infra nodes (istio, grafana, jaeger, prometheus), indicating if they are healthy or not.

If the infra node is unhealthy, it will be highlighted in red and linked to the Kiali instance with dash red edge.

image

Status of all infra nodes (except Kiali) is shown in the target panel:

image

Besides, a link to the mesh page has been added in the Istio Component Status tooltip so the user can investigate the infra status in detail.

image

**Note**: Now the istio status api returns status of all infra nodes (both healthy and unhealthy). Currently istio status api returns only those nodes that have issues, but IMO status should confirm the status of all infra nodes. In the frontend we could assume that `no_health_status` == `healthy node` but this assumption looks weird to me.

### Steps to test the PR

1. Break any infra node (modifying the deployment, removing port-forwarding, etc.)
2. Check the health status of infra nodes

### Issue reference

Closes #7139


mesh/appender.go
```diff
@@ -14,10 +15,11 @@ type AppenderVendorInfo map[string]interface{}
...
 // is initially empty.
 type AppenderGlobalInfo struct {
-	Business   *business.Layer
-	Context    context.Context
-	PromClient *prometheus.Client
-	Vendor     AppenderVendorInfo // telemetry vendor's global info
+	Business          *business.Layer
+	Context           context.Context
+	PromClient        *prometheus.Client
+	IstioStatusGetter IstioStatusGetter
```
jshaughn, line 21:
```suggestion
	IstioStatusGetter IstioStatusGetter
	PromClient        *prometheus.Client
```

mesh/appender.go
```diff
@@ -56,3 +58,8 @@ type Appender interface {
 	// Name returns a unique appender name and which is the name used to identify the appender (e.g in 'appenders' query param)
 	Name() string
 }
+
+// IstioStatusGetter gets the status of the istio graph components
```
jshaughn, line 62: This `IstioStatusGetter` name is a little confusing to me.  It's not getting the status of just Istio, right, it's getting status for the various mesh infra components.  Would `MeshStatusGetter` or `InfraStatusGetter` make sense?

jshaughn (commented):
This is looking really good.  The status colors combined with the broken edge are easy to see.  A couple of things I noticed:
- The health status does not change in the side panel even when it changes in the graph.  This may be unrelated to the PR, it could be a side-panel refresh issue.  I can look at this tomorrow, as well.
- Is there anything we can show when hovering on health in the side panel?  In general our health icons give something on hover, so it's a little weird not to have something pop up.
- One thing we may want consider is making the health checking optional and done in an appender.  But it's so critical that it probably doesn't make sense, unless it somehow ends up having perf impact.

I did recreate the test fails locally.  I looked quickly but I didn't see the cause...

business/istio_status.go
```diff
@@ -277,17 +276,23 @@ func getAddonStatus(name string, enabled bool, isCore bool, auth *config.Auth, u
 		auth.Token = token
 	}
 
+	var status string
```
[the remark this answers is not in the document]
ferhoyos: Code updated

mesh/appender.go
ferhoyos: Yes, I think `MeshStatusGetter` is more appropiate in this case.

ferhoyos:
Thanks for the review @jshaughn! I think I have addressed all the issues/comments:

> The health status does not change in the side panel even when it changes in the graph.

It was not related by this PR, but I have fixed it anyway. The target panel was not being updated when the mesh data changed. Since I was looking that part of the code, I have also fixed the resize issue when you open/close the target side panel. We should check if these issues are present in PF graph and fix them there as well.

> Is there anything we can show when hovering on health in the side panel? In general our health icons give something on hover, so it's a little weird not to have something pop up.

Good catch! To be aligned with the rest of UX, I have placed the health icon next to the node text and with the tooltip, like we have in detail pages.

image

> One thing we may want consider is making the health checking optional and done in an appender. But it's so critical that it probably doesn't make sense, unless it somehow ends up having perf impact.

From the moment I would not make it optional because the health status is critical information and IMO it is difficult to have some performance issues (you have small number of infra nodes, different from namespaces or workload health request).

frontend/src/pages/Mesh/Mesh.tsx
```diff
@@ -188,7 +190,13 @@ const TopologyContent: React.FC<{
       requestFit = true;
       controller.getGraph().reset();
       controller.getGraph().layout();
+
+      // Fit padding after resize
+      setTimeout(() => {
+        controller.getGraph().fit(FIT_PADDING);
+      }, 0);
     }
```
jshaughn, line 198: As you know, I spent a while trying to solve this issue and I had considered a timeout but thought that really this was an issue in PFT, and that it needed to be more responsive to resize.  I don't really know why adding this to the end of the queue works, but I'm glad it does!  We may be able to remove the "onResize" infra I added while trying to solve the problem, I don't think it's necessary.  I can do that in my current PR, as well as apply this to GraphPF.

jshaughn (approved):
Really awesome job with the new feature and also resolving some existing problems!  Let's get this merged!

PR #7312, opened 2024-04-25
[hack] bring back kiali in MC scripts

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7313

I do not know why these were removed. We have all the kiali options for these hack scripts but they are now ignored - kiali is never deployed even if enabled.

These were removed in the following commits:

* https://github.com/kiali/kiali/commit/804f21ebe3d93be9f167353b81895ce7044c7e44#diff-ae968abe0209eefd78f858f137a2e418a056ec6e02965d2214bcb46c8a978efeL77-L78
* https://github.com/kiali/kiali/commit/d8631a1d8814f2c025ce6ca460de147aa810368b#diff-b53029271ec37bbeb86a1eb29db8fd5e77a74f0fda8dac3d7d22109fcb11fdf3L203-L204

### DIFF
--- a/hack/istio/multicluster/env.sh
+++ b/hack/istio/multicluster/env.sh
@@ -52,7 +52,7 @@ ISTIO_DIR=""
 DORP="${DORP:-podman}"
 
 # The namespace where Istio will be found - this namespace must be the same on both clusters
-ISTIO_NAMESPACE="istio-system"
+ISTIO_NAMESPACE="${ISTIO_NAMESPACE:-istio-system}"
 
 # If you want to pull Istio images from a different image repository than what the hack script
 # will tell Istio to pull from, then set that hub name here. If you set this to "default",
@@ -75,11 +75,11 @@ NETWORK1_ID="network-east"
 NETWORK2_ID="network-west"
 
 # Deploy a single kiali or a kiali per cluster
-SINGLE_KIALI="true"
+SINGLE_KIALI="${SINGLE_KIALI:-true}"
 
 # Create kiali remote secrets so kiali can access the different clusters
 # When left empty, this will be true if SINGLE_KIALI is true or false otherwise.
-KIALI_CREATE_REMOTE_CLUSTER_SECRETS=""
+KIALI_CREATE_REMOTE_CLUSTER_SECRETS="${KIALI_CREATE_REMOTE_CLUSTER_SECRETS:-}"
 
 # If a gateway is required to cross the networks, set this to true and one will be created
 # See: https://istio.io/latest/docs/setup/install/multicluster/multi-primary_multi-network/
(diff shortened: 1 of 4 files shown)


jmazzitelli:
I'm testing this now. I'm curious why these are removed... there must have been a reason :) But with them removed, all the kiali options are useless and you never get kiali installed if you use these hack scripts to start a MC demo.

jmazzitelli:
After running `hack/istio/multicluster/install-primary-remote.sh` I now see kiali:

```
$ kubectl --context east get pods -n istio-system -l app=kiali
NAME                     READY   STATUS    RESTARTS   AGE
kiali-7b556f99cf-gjhsv   1/1     Running   0          3m37s
```

jmazzitelli:
Same thing with the other script - only this time it exits immediately when the error occurs:

```
INFO: A remote cluster secret named [kiali-remote-cluster-secret-west] has been created in the Kiali cluster namespace [istio-system]. It can be used by Kiali to access the remote cluster.
Release "kiali-server" does not exist. Installing it now.
NAME: kiali-server
LAST DEPLOYED: Thu Apr 25 14:03:24 2024
NAMESPACE: istio-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
Welcome to Kiali! For more details on Kiali, see: https://kiali.io

The Kiali Server [v1.83.0] has been installed in namespace [istio-system]. It will be ready soon.

(Helm: Chart=[kiali-server], Release=[kiali-server], Version=[1.83.0])
service/kiali patched
service/kiali condition met
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:10 --:--:--     0curl: (6) Could not resolve host: realms
curl: (6) Could not resolve host: admin
```

jmazzitelli:
All CI checks are passing, so the changes to this PR do not affect the CI setup / scripts. That's a good thing. I will try to figure out how to clear up those errors and get this to work so people can run the MC hack scripts and get a MC demo running locally with Kiali.

jmazzitelli:
KEYCLOAK_ADDRESS is an empty string when the deploy-kiali.sh is being called... I believe that's the problem. We need that set correctly and hopefully that fixes the problem

jmazzitelli:
@nrfox What should KEYCLOAK_ADDRESS be when on minikube?

Based on what I see, our `deploy-kiali.sh` is not even being used anymore anywhere. I put the code back to invoke it again, but it can't run because KEYCLOAK_ADDRESS is an empty string, which I think is an env var that is only set when these scripts are using Kind.

I think if I find out what KEYCLOAK_ADDRESS is supposed to be, we can set it and it should be back to working.

I'm curious how people are actually using these MC hack scripts to use a MC environment without having Kiali even deployed?

nrfox:
> I'm testing this now. I'm curious why these are removed... there must have been a reason

I think if you add back the `deploy-kiali.sh` call then you will end up trying to deploy Kiali twice when running kind since that script is called here:

https://github.com/kiali/kiali/blob/a55c5c53bdf19d8fbe1e5a8f249d04d1d371c458/hack/setup-kind-in-ci.sh#L278-L283

And later kiali is deployed here: https://github.com/kiali/kiali/blob/a55c5c53bdf19d8fbe1e5a8f249d04d1d371c458/hack/setup-kind-in-ci.sh#L309-L326

> I'm curious how people are actually using these MC hack scripts to use a MC environment without having Kiali even deployed?

Personally I just use kind because that is what CI uses and it is the easiest to setup and consumes the least amount of resources. Which might mean that minikube isn't well supported if other folks are not running it.

> @nrfox What should KEYCLOAK_ADDRESS be when on minikube?

This is where keycloak address comes from in the kind setup: https://github.com/kiali/kiali/blob/a55c5c53bdf19d8fbe1e5a8f249d04d1d371c458/hack/setup-kind-in-ci.sh#L311-L313

nrfox:
@jmazzitelli I think if you run these separately it might still work on minikube:
```
hack/istio/multicluster/install-multi-primary.sh -mm true -kudi true
hack/istio/multicluster/deploy-kiali.sh --single-kiali true --cluster1-context east --cluster2-context west --kiali-auth-strategy openid --keycloak-address keycloak-192-168-39-130.nip.io
```

jmazzitelli:
`./hack/istio/multicluster/install-multi-primary.sh` results in:

image

jmazzitelli:
`hack/istio/multicluster/install-primary-remote.sh` results in:
image

jshaughn (approved):
Visually approving this.

PR #7316, opened 2024-04-26
[hack] bump crc to 2.34.1 / openshift 4.15

jmazzitelli:
There was a problem that the CRC folks figured out over here: https://github.com/crc-org/crc/issues/4110 

@ferhoyos can confirm the problem I am seeing - he sees the same error.

This new PR performs the workaround that the CRC folks recommended.


jshaughn (approved):
Fine by me.

PR #7321, opened 2024-04-29
Replace old Mesh page with New Mesh page

### LINKED ISSUE #6431: Show mesh configuration per controlplane

### LINKED ISSUE #7298: Remove legacy Mesh page
This is part of the new Mesh page epic.  This replaces the current Mesh page with the new Mesh page, although it has additional impact, as this ties into https://github.com/kiali/kiali/issues/6431.  As such, there are changes to the masthead and the control plane card in Overview.

Closes #6431

jshaughn:
This PR replaces the old Mesh page with the new Mesh page, removing the undocumented feature flag.  It should contain a releasable version.

Closes #7298
Closes #6431

TESTING
The PR removes the new mesh page feature flag, and so it should appear as the default Mesh Page.  The goal of the first release of the page is not perfection, but it should be solid, usable, and useful.  It replaces the old mesh page, the non-Kiali parts of the About page, and demotes the Overview Control plane card to a standard card (although it will still be labeled as a control plane).   It should be able to handle standard deployments, multi-cluster deployments, canary upgrades, and externally deployed infra (you don't necessarily have to test all of these things).


frontend/src/pages/Mesh/target/TargetPanelCluster.tsx
```diff
@@ -88,42 +92,57 @@ export class TargetPanelCluster extends React.Component<TargetPanelCommonProps,
...
+              .map(n => {
+                return renderNodeHeader(n.getData() as MeshNodeData, this.props.t, true, TitleSizes.md);
+              })}
+          </div>
+        ) : (
+          <div className={targetPanelBody}>
+            {clusterData.accessible && this.renderKialiLinks(clusterData.kialiInstances)}
+            {version && (
+              <>
+                {`${this.props.t('Version')}: `}
```
ferhoyos, line 131:
```suggestion
                {`${this.props.t('Version')}: ${version}`}
```

frontend/src/pages/Mesh/target/TargetPanelCluster.tsx
```diff
@@ -88,42 +92,57 @@ export class TargetPanelCluster extends React.Component<TargetPanelCommonProps,
...
+                {version}
+                <br />
+              </>
+            )}
+            {`${this.props.t('Network')}: `}
```
ferhoyos, line 136:
```suggestion
            {`${this.props.t('Network')}: ${clusterData.network ?? this.props.t('n/a')}`}
```

ferhoyos:
It is not related with this PR, but I think istio status should take into account both istiod and istio pod status. Currently it only takes into account istiod, show it shows a `healthy` status if there is something wrong with istio pod:

image

image

@jshaughn I can create a different issue or fix it in this PR, as you consider.

jshaughn:
> It is not related with this PR, but I think istio status should take into account both istiod and istio pod status. Currently it only takes into account istiod, show it shows a healthy status if there is something wrong with istio pod:

@ferhoyos +1, a different issue or a PR against this PR is fine.  This PR is mostly stable with changes only resulting from adding test code or making fixes.  I made your suggested change about treating Data Plane and Kiali as Healthy (basically, if there is no health assigned then it defaults to Healthy).  I also applied your suggestion about embedding the JSON config into the Data Plane expandable sections.

frontend/src/pages/Mesh/MeshPage.tsx
```diff
@@ -82,8 +82,9 @@ export type MeshData = {
 };
 
 type MeshPageState = {
+  controller?: Controller;
   meshData: MeshData;
-  lastResizeTime: TimeInMilliseconds; // just a way to force a top-down re-render on a mesh-level resize (e..f targetPanelCollapse)
+  setSelectedIds?: (values: string[]) => void;
```
ferhoyos, line 87: Why do you add `setSelectedIds` in the MeshPageState? If it is just for cypress tests I think we should add a comment indicating it (bit weird to see this kind of methods in a component state)

jshaughn: yes, and it is weird.  and yes, I will add something here, I only got this in place last night and then stopped for the day...

hhovsepy:
Tested on Multicluster, single cluster and in Ambient mode, works fine, currently checking on OCP cluster (will take longer due to env issues):
Screenshot from 2024-05-15 09-40-18
Screenshot from 2024-05-15 09-22-03
Screenshot from 2024-05-15 09-15-16
Screenshot from 2024-05-15 09-14-17

hhovsepy:
Tested on OCP, working fine, just one question, the component's versions are undefined, does any config missing:
Screenshot from 2024-05-15 12-53-53

at least Grafana has a version:
Screenshot from 2024-05-15 12-55-23

leandroberetta:
Did a first run on Minikube and works well and I really like how it looks and how the information is grouped.

image

The collapsible namespaces for the data plane is awesome.

One minor thing that I guess can be improve is that when you have a namespace with no inbount and outbound traffic, it can be a small indication of this situation (to avoid the size of the empty graphs):

image

Related to versions, I do notice that Jaeger and Grafana versions are unknown.

I will start the proper code review and probably a canary/mc testing too now, but I'm very happy with this new mesh page!.

jshaughn:
> Tested on OCP, working fine, just one question, the component's versions are undefined,

@hhovsepy Thanks for the testing!  Good question, the version is not set in the place we are currently looking, but maybe there is another way to get the version for the "unknown" component versions.  We will look into it.

jshaughn:
> One minor thing that I guess can be improve is that when you have a namespace with no inbount and outbound traffic, it can be a small indication of this situation (to avoid the size of the empty graphs):

@leandroberetta Thanks for testing!  I agree about the wasted space.  We will try and improve that situation.

>  I do notice that Jaeger and Grafana versions are unknown.

Yes, Hayk mentioned the same.  We will look to see if we can determine the information.

jshaughn:
@hhovsepy , @leandroberetta , @ferhoyos , I pushed changes that should now show Grafana version, if available (tracing version is still not solved).  Also, I reduced the vertical whitespace in an expanded data plane namespace, when it lacks traffic for the charts.

jshaughn:
Hi @hhovsepy , @leandroberetta ,  We've incorporated your feedback.  If public URLs can be determined then we can show the Grafana and Jaeger version info:
image

And we've fixed the excessive whitespace when no metrics are available for a data plane namespace:
image

The CI failures are limited to the know issue with multi-primary.   If you are OK with the state of things, please approve and we'll get this in for some soak time prior to sprint end.  If you have more suggestions we can follow up with other issues for the Epic.  Of course if you see a blocking issue we'll fix it in this PR.  Thanks!

hhovsepy:
Hi @jshaughn , the components versions are fixed, thanks.
However the cluster version is Unknown, tested on OCP:
Screenshot from 2024-05-16 21-32-43

hhovsepy (approved):

PR #7329, opened 2024-05-02
Fix cypress login on openshift

nrfox:
### Describe the change

Validates cookie is present in `cy.session`. Fixes `my_htpasswd_provider` environments.

### Steps to test the PR

Deploy kiali on openshift. Run cypress with:
```
CYPRESS_BASE_URL=<kiali_route>
CYPRESS_USERNAME=<username>
CYPRESS_PASSWD=<pass>
```

### Automation testing

Fixes existing tests.

### DIFF
--- a/frontend/cypress.config.ts
+++ b/frontend/cypress.config.ts
@@ -42,7 +42,6 @@ export default defineConfig({
 
       // This name is non-standard and might change based on your environment hence the separate
       // env variable.
-      config.env.AUTH_PROVIDER = config.env.AUTH_PROVIDER || 'my_htpasswd_provider';
       config.env.AUTH_STRATEGY = await getAuthStrategy(config.baseUrl!, config.env.ALLOW_INSECURE_KIALI_API);
 
       return config;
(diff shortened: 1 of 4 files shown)


matejnesuta:
Tested on upstream OCP: 

the last test fails with this, but I am not sure, if the failure is related to this PR:
image

matejnesuta:
Oh, there is a missing gherkin step in the last scenario. That's why it failed.

Also the 4th scenario should have a demo app tag like `@bookinfo-app`. I tried to run the test against an env without any apps and the test failed because of that. 

Not sure if these changes should be addressed here, or in a separate PR.

nrfox:
> Oh, there is a missing gherkin step in the last scenario. That's why it failed.

Good catch! I'll update.

> Also the 4th scenario should have a demo app tag like @bookinfo-app. I tried to run the test against an env without any apps and the test failed because of that.

It should still work even without apps deployed and all of these tests are smokes so should run in envs without apps. You should still be able to go to the overview page. Maybe the assertion about the overview page is busted?

nrfox:
@matejnesuta would you try running the tests again?

matejnesuta (approved):
Passing. LGTM.

PR #7333, opened 2024-05-03
[hack] update with new sail stuff

jmazzitelli:
Some things have changed with Sail operator. This updates the hack scripts to work with it.

To test:

1. Start an OpenShift cluster (e.g. `hack/crc-openshift.sh start`)
2. Use `oc` to log into the cluster as a kubeadmin user
3. Install the operators for Sail, Kiali, Tempo, etc.: `hack/istio/sail/install-ossm-release.sh install-operators`
4. Install a Istio, Kiali, Prometheus, etc. etc.: `hack/istio/sail/install-ossm-release.sh install-istio`
5. See that everything eventually installs successfully: `hack/istio/sail/install-ossm-release.sh status`
6. Bring up the Kiali UI in a browser: `hack/istio/sail/install-ossm-release.sh kiali-ui`
7. See that you can get to the UI and things look OK

When you are done, you can delete everything except the operators with: `hack/istio/sail/install-ossm-release.sh delete-istio`

After Istio and all other components are removed, you can then delete the operators which takes you back to a clean slate:  `hack/istio/sail/install-ossm-release.sh delete-operators`

### DIFF
--- a/hack/istio/sail/func-kiali.sh
+++ b/hack/istio/sail/func-kiali.sh
@@ -75,6 +75,16 @@ install_kiali_cr() {
     fi
   fi
 
+  # determine the control plane's Istio version - we need it because it is part of the names of the Istio config maps/deployments
+  local istio_version="unknown"
+  for r in $(${OC} get istio -o name);
+  do
+    local ns="$(${OC} get $r -o jsonpath='{.spec.namespace}')"
+    if [ "${ns}" == "${control_plane_namespace}" ]; then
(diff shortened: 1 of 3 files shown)


jmazzitelli:
Note the one change, replacing `quay.io/official-images/busybox:1.28` with `quay.io/jitesoft/alpine:latest` was because the original one is no longer accessible, and all we really need is some small image with the ability to run the small sh command.

jshaughn (approved):
Ran through the test steps, works like a chaahm!

Issue #7238, opened 2024-03-28
make sure Kiali can observe important istiod metrics


jmazzitelli:
Istio has a bunch of good metrics that we should graph in Kiali so people can observe performance of their control  plane directly in the Kiali UI.

This is a good write up and documents these pilot metrics - we should create an istio dashboard with these metrics so when someone looks at the istiod application in Kiali, they can see metric graphs for these metrics:

https://github.com/jewertow/istio-playground/tree/ossm-performance-tuning/ossm-performance-tuning

Custom dashboards are documented here: https://kiali.io/docs/configuration/custom-dashboard/

We would want to add this as a built-in dashboard, which are defined here: https://github.com/kiali/kiali/blob/master/config/dashboards/dashboards.go

jmazzitelli:
I'll look to see what this would involve to get a basic dashboard in place. If it involves what I think it involves, it shouldn't be too hard.

PR #7347, opened 2024-05-07
add Istiod dashboard

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7238

To test, navigate to the istiod Application, and click the "Istiod Metrics" tab:

Screenshot from 2024-05-07 16-12-41

See https://istio.io/latest/docs/reference/commands/pilot-discovery/#metrics

### DIFF
--- a/config/dashboards/dashboards.go
+++ b/config/dashboards/dashboards.go
@@ -1,6 +1,43 @@
 package dashboards
 
 const DEFAULT_DASHBOARDS_YAML = `
+- name: istiod
+  title: Istiod Metrics
+  discoverOn: "pilot_info"
+  items:
+  - chart:
+      name: "Proxy Push Time"
+      spans: 3
+      metricName: "pilot_proxy_convergence_time_count"
+      unit: "s"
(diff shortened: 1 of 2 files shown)


jshaughn (approved):
I first tried this against an env I had running, which was the Sail operator etc, and it didn't seem to work.  It did work in a clean env off of master.

One issue is that some of the metrics get dropped if you follow our Prom Tuning docs.  Also, we don't list these metrics as used by Kiali.  I'd suggest adding a kiali.io PR to add these to the list of used metrics, and also update the regex we suggest for dropping metrics (see me if you need details).

jmazzitelli:
doc changes:
1. Change the regex to pilot_[^px].* in this line: https://github.com/kiali/kiali.io/blob/staging/content/en/docs/Configuration/p8s-jaeger-grafana/prometheus.md?plain=1#L94
2. add the metrics to the list here: https://github.com/kiali/kiali.io/blob/staging/content/en/docs/FAQ/general.md?plain=1#L159

Issue #7266, opened 2024-04-08
Custom http headers for tracing


krmcbride:
### What do you want to improve?
Similar to https://github.com/kiali/kiali/issues/4323 but for the tracing config -- I'm using a multitenant Grafana Tempo deployment which needs a `X-Scope-OrgID` header set with a tenant ID.  I can connect to Grafana Mimir with the `prometheus.custom_headers` config:

```yaml
spec:
  # ...
  external_services:
    prometheus:
      custom_headers:
        "X-Scope-OrgID": "<the tenant id>"
```
But there doesn't appear to be an equivalent for tracing/tempo.

### What is the current behavior?
Requests to Tempo without the `X-Scope-OrgID` set result in a
```
HTTP/1.1 401 Unauthorized
Content-Type: text/plain; charset=utf-8

no org id
```
response, which causes a JSON unmarshalling error.

As a temp workaround I'm proxying requests from Kiali to Tempo through an nginx pod with
```
location ^~ /api {
  proxy_pass        https://my-grafana-tempo-url$request_uri;
  proxy_set_header  X-Scope-OrgID "<the tenant id>";
}
```
to set the correct tenant header, and trace info appears in Kiali as expected.

### What is the new behavior?
Logically I'd expect there to be an optional `spec.external_services.tracing.custom_headers` map, same as for the prometheus client.

jshaughn:
I think this makes sense.  We support `query_scope`, but that adds tags to queries in order to narrow the queried data, when trace data is unified.  This is different as it is adding headers.

I will add this to the backlog, @krmcbride, would you be interested in contributing this feature?

jmazzitelli:
I did the original impl for the prom headers, so I can try to do this since its very similar

jmazzitelli:
@krmcbride Do you want to test this locally to see if it addresses your issue? I uploaded the dev images on my quay repo:

* Server: `quay.io/jmazzitelli/kiali:issue7266` -> https://quay.io/repository/jmazzitelli/kiali?tab=tags
* Operator (if you are using the operator to install): `quay.io/jmazzitelli/kiali-operator:issue7266` -> https://quay.io/repository/jmazzitelli/kiali-operator?tab=tags

The new configuration should look like:

```yaml
external_services:
  tracing:
    custom_headers:
      HeaderName1: HeaderValue1
```

jmazzitelli:
The two PRs are ready for review. In my testing, I can see the custom headers get passed using kubeshark to examine the requests.

[...]

PR #7353, opened 2024-05-13
send custom headers in tracing backend requests

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7266

### DIFF
--- a/tracing/client.go
+++ b/tracing/client.go
@@ -198,7 +198,7 @@ func newClient(ctx context.Context, cfg *config.Config, token string) (*Client, 
 					dialOps = append(dialOps, grpc.WithTransportCredentials(insecure.NewCredentials()))
 				}
 				grpcAddress := fmt.Sprintf("%s:%d", u.Hostname(), cfg.ExternalServices.Tracing.GrpcPort)
-				clientConn, _ := grpc.Dial(grpcAddress, dialOps...)
+				clientConn, _ := grpc.DialContext(ctx, grpcAddress, dialOps...)
 				streamClient, err := tempo.NewgRPCClient(client, u, clientConn)
 				if err != nil {
 					log.Errorf("Error creating gRPC Tempo Client %s", err.Error())
--- a/tracing/client.go
+++ b/tracing/client.go
@@ -15,6 +15,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/metadata"
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/kiali/kiali/config"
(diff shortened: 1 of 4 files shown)


jmazzitelli:
1. With this PR checked out, build the server: `make build build-ui test`
2. Start a cluster with Kiali (using what you built in step 1) w/Tracing enabled: `hack/run-integration-tests.sh -so true`
3. Add a custom header to the Tracing config: `kubectl edit cm kiali -n istio-system`
```yaml
external_services:
  tracing:
    url: http://tracing.istio-system:16685/jaeger
    custom_headers:
      X-MazzWasHere_Foo: mazz_WasHere
```
4. Restart the Kiali pod to pick up the change: `kubectl delete pod -n istio-system -l app.kubernetes.io/name=kiali`
5. Confirm the header was added to the client
```
$ kubectl logs -n istio-system kiali-8ff578f47-hv6c4 --tail=-1 | grep "custom headers"
2024-05-17T14:27:30Z TRC Adding [1] custom headers to Tracing client
```
6. At this point it would be nice to actually see the header in the HTTP request, but at the moment I don't know a way to see it. The request is going from the server to the tracing backend - seeing the browser's network traffic via the browser debugging tools is therefore useless. But that's how you really can know if the header was sent.

jmazzitelli:
latest commit has it working with grpc:
grpc-kubeshark

jmazzitelli:
This is all ready. I was able to confirm the header is passed with using gRPC and non-gRPC.

jmazzitelli:
@josunect I added you as a reviewer to take a look at the code since you are most familiar with it. The person who asked for this feature tested it in his cluster and it works for him, and I tested it (as you see above) and it works for me, too.  I just want to make sure the code doesn't break something else that I cannot see.

jshaughn (approved):
I didn't run it because the community originator  of the RFE already tested it successfully.  Code looks good for me.

tracing/client.go
```diff
@@ -165,7 +171,7 @@ func newClient(ctx context.Context, cfg *config.Config, token string) (*Client,
 		// Legacy HTTP client
 		log.Tracef("Using legacy HTTP client for Tracing: url=%v, auth.type=%s", u, auth.Type)
 		timeout := time.Duration(config.Get().ExternalServices.Tracing.QueryTimeout) * time.Second
-		transport, err := httputil.CreateTransport(&auth, &http.Transport{}, timeout, nil)
+		transport, err := httputil.CreateTransport(&auth, &http.Transport{}, timeout, cfgTracing.CustomHeaders)
 		if err != nil {
 			return nil, err
 		}
```
josunect, line 177: I can't add the comment in the specific line, but I think the ctx should also need to be added in the Tempo grpc client? This would be in line 195: 
`
				clientConn, _ := grpc.Dial(grpcAddress, dialOps...)`

I think that could be replaced with something like: 

`clientConn, _ := grpc.DialContext(ctx, grpcAddress, dialOps...)`

josunect (commented):
I haven't tested it yet, I've just added one comment after looking at the code. Except for that, it looks good!

tracing/client.go
jmazzitelli: I made this change - see commit https://github.com/kiali/kiali/pull/7353/commits/abb5e7a8472e35784294dd9d3b00cf0eeac6ed14

jmazzitelli:
The failure is the known flake unrelated to this PR

jshaughn (approved):
Re-approving, +1 to Josune's suggested change.

PR #7356, opened 2024-05-15
Explicitly enable tracing in ci istio setup

nrfox:
Starting in 1.22, zipkin tracing is no longer enabled by default. Therefore, in our CI setup we need to explicitly enable it.


jshaughn (approved):
Thanks, Nick, let's get this in and then worry about the Tempo issue.

PR #7358, opened 2024-05-15
Fix tracing for tempo CI job

nrfox:
Re-enables tracing for tempo CI job.

Follow up to #7356


jshaughn (approved):
Thanks again, @nrfox !  Did the trick.

jshaughn (approved):

PR #7362, opened 2024-05-16
Remove login actions for OAuth

nrfox:
### Describe the change

These login actions are (probably) no longer needed and they mess with the URL post login which is causing the cypress tests to fail and probably unnecessary redirects for users.

### Steps to test the PR

Perform a successful login/logout with openshift/openid auth strategies. Session timeout should also work as expected.

### Automation testing

Fixes current automation.


nrfox:
We probably don't want to remove this entirely and this should get addressed by #7366

PR #7364, opened 2024-05-16
Fix deploy-kiali.sh testing user roles

nrfox:
### Describe the change

Due to recent helm chart changes, the `kiali` clusterrole no longer gets created when using the oauth strategy. The CI tests relied on this role to be present to give the testing user permissions to login and interact with kiali. This change generates the old `kiali` clusterrole from helm directly rather than relying on it being present in the cluster.

### Steps to test the PR

`./hack/run-integration-tests.sh --test-suite frontend-multi-primary`

### Automation testing

Multi-primary tests pass.


jmazzitelli (approved):
Bah! I knew [that change](https://github.com/kiali/helm-charts/pull/261) was going to break something. I did look around but obviously missed this.

PR #7365, opened 2024-05-17
[hack] clean up the SCRIPT_DIR declaration and usage

jmazzitelli:
to test -- cd to some directory (like `/hack`) and run the script. It should still work.


ferhoyos (approved):
LGTM

PR #7366, opened 2024-05-17
Implement multi-session for openshift auth strategy

### LINKED ISSUE #6360: Multicluster - Support openshift auth strategy
Openshift clusters each have their own internal oauth server and tokens are specific to each cluster.

Required:
https://github.com/kiali/kiali/issues/6037

nrfox:
### Describe the change

Enables multiple logins through openshift auth strategy.

Some consequences of this change:
1. Current sessions will be invalidated and users will need to login again.
2. Callback url for openshift has changed to `/api/auth/callback/<cluster>` or `/api/auth/callback` for the home cluster. These should get updated automatically when using the `openshift` auth strategy by the operator or helm chart.

### Steps to test the PR

Regression test each auth strategy, including openid for multicluster and anonymous.

Testing multiple clusters on openshift:

1. Login to each cluster and set your contexts to `east`/`west` 
```
oc login <east cluster>
kubectl config rename-context $(kubectl config current-context) east

oc login <west cluster>
kubectl config rename-context $(kubectl config current-context) west
```

2. Checkout helm chart pr and build helm charts

    https://github.com/kiali/helm-charts/pull/262
    ```
    make build-helm-charts
    ```

3. Deploy istio and kiali
```
hack/istio/multicluster/install-multi-primary.sh --manage-minikube false --manage-kind false --cluster1-context east --cluster2-context west --kiali-auth-strategy openshift --kiali-use-dev-image true --kiali-server-helm-charts <path-to-helm-charts> --kiali-build-dev-image true
```

4. Open kiali and login to "east" cluster
5. Login to remote cluster by clicking on the user profile dropdown and clicking the "west" cluster.
6. Logout

### Automation testing

Included cypress tests for logging into multiple openshift clusters.

### Issue reference

fixes #6360 

There's still some behavior that can be addressed later.
1. You **must have access to the cluster kiali is deployed on** to login. With this change the default is to login to the cluster where kiali is deployed and only after you have logged in there can you login to remote clusters with the user profile dropdown. If you only have access to a "remote" cluster then you won't be able to login. We will likely need some kind of "cluster selector" page so users can choose which cluster to login to initially. We most likely will want some kind of warning or notification as well when the remote cluster session expires.
2. **Implementing refresh tokens** on the backend should greatly reduce the amount of times the user has to login to the various clusters.


hack/istio/multicluster/deploy-kiali.sh
```diff
@@ -70,26 +76,36 @@ deploy_kiali() {
       echo "Pushing the images into the cluster..."
       make -e DORP="${DORP}" -e CLUSTER_TYPE="kind" -e KIND_NAME="${cluster_name}" CLUSTER_REPO=localhost cluster-push-kiali
       helm_args+=('--set deployment.image_pull_policy="Never"')
+      helm_args+=("--set deployment.image_name=localhost/kiali/kiali --set deployment.image_version=dev")
+    elif [ "${KIALI_AUTH_STRATEGY}" == "openshift" ]; then
+      local image_name
+      image_name="$(oc get image.config.openshift.io/cluster -o custom-columns=INT:.status.internalRegistryHostname --no-headers 2>/dev/null)/kiali/kiali"
+      make -e DORP="${DORP}" CLUSTER_TYPE="openshift" build build-ui cluster-push-kiali
```
jmazzitelli, line 83: To avoid what happened earlier, make sure this script does a "cd" to the expected location thus allowing people to run this script from anywhere. Line 10 sets the SCRIPT_DIR but I do not see it "cd"'ing anywhere so this script only assumes it knows what directory it is in, but it may not be where this script thinks it is.

hack/istio/multicluster/deploy-kiali.sh
```diff
@@ -68,29 +74,49 @@ deploy_kiali() {
   if [ "${KIALI_USE_DEV_IMAGE}" == "true" ]; then
     if [ "${MANAGE_KIND}" == "true" ]; then
       echo "Pushing the images into the cluster..."
-      make -e DORP="${DORP}" -e CLUSTER_TYPE="kind" -e KIND_NAME="${cluster_name}" CLUSTER_REPO=localhost cluster-push-kiali
-      helm_args+=('--set deployment.image_pull_policy="Never"')
-      helm_args+=("--set deployment.image_name=localhost/kiali/kiali --set deployment.image_version=dev")
+      make -e -C "${KIALI_REPO_ROOT}" DORP="${DORP}" CLUSTER_TYPE="kind" KIND_NAME="${cluster_name}" cluster-push-kiali
```
jmazzitelli, line 77: looks like the -e is in the wrong spot .  "-e" is for env var setting.. which is the DORP var.. so you want this (note that some make impls the -e is optional.. either get rid of all the `-e` or put them for everything. I honestly don't know why make has this -e because I've always seen it optional.

```suggestion
      make -C "${KIALI_REPO_ROOT}" -e DORP="${DORP}" -e CLUSTER_TYPE="kind" -e KIND_NAME="${cluster_name}" cluster-push-kiali
```

hack/istio/multicluster/deploy-kiali.sh
```diff
@@ -68,29 +74,49 @@ deploy_kiali() {
...
+      helm_args+=(
+        '--set deployment.image_pull_policy="Never"'
+        "--set deployment.image_name=localhost/kiali/kiali"
+        "--set deployment.image_version=dev"
+      )
+    elif [ "${KIALI_AUTH_STRATEGY}" == "openshift" ]; then
+      local image_name
+      podman login --tls-verify=false -u "$(oc whoami | tr -d ':')" -p "$(oc whoami -t)" "$(oc get image.config.openshift.io/cluster -o custom-columns=EXT:.status.externalRegistryHostnames[0] --no-headers 2>/dev/null)"
+      image_name="$(oc get image.config.openshift.io/cluster -o custom-columns=INT:.status.internalRegistryHostname --no-headers 2>/dev/null)/kiali/kiali"
+      make -e -C "${KIALI_REPO_ROOT}" DORP="${DORP}" CLUSTER_TYPE="openshift" cluster-push-kiali
```
jmazzitelli, line 87:
```suggestion
      make -C "${KIALI_REPO_ROOT}" -e DORP="${DORP}" -e CLUSTER_TYPE="openshift" cluster-push-kiali
```

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
```diff
@@ -14,7 +14,7 @@ set -u
 ##############################################################################
 
 # Used to name all the resources on the remote cluster.
-KIALI_RESOURCE_NAME="kiali-remote-access"
+KIALI_RESOURCE_NAME="kiali"
```
jmazzitelli, line 17: you sure you want to rename this? IIRC it was specifically not named "kiali" to avoid potential confusion with a real kiali install (say from the helm chart or operator).

jmazzitelli (changes requested):
just a quick review of a couple of the files. this is not an indepth review yet

hack/istio/multicluster/deploy-kiali.sh
jmazzitelli: Come to think of it, it could be if you do not specify `-e` maybe bash interprets that as env vars that it passes to the process. So it could be `make DORP=docker` is the same as `DORP=docker make ...` (that's just a guess).

Anyway, I would use `-e` to make it clear what it is going on, rather than omit it.

nrfox: I don't think you need to specify `-e` multiple times.
>        -e, --environment-overrides
>           Give variables taken from the environment precedence over variables from makefiles.

It's just a flag for the `make` command and not a variable argument?

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
nrfox: Eh no I probably edited this while testing out changes and forgot to remove it. Good catch.

hack/istio/multicluster/deploy-kiali.sh
```diff
@@ -121,17 +157,22 @@ deploy_kiali() {
...
     kiali-server
     ${KIALI_SERVER_HELM_CHARTS}'
 
   eval $helm_command
+  if [ "${KIALI_AUTH_STRATEGY}" == "openshift" ]; then
+    local kiali_route_url
+    kiali_route_url=$(kubectl get route kiali -n "${ISTIO_NAMESPACE}" -o jsonpath='{.spec.host}')
+    ${CLIENT_EXE} get oauthclients --context "${CLUSTER1_CONTEXT}" -n "${ISTIO_NAMESPACE}" kiali-"${ISTIO_NAMESPACE}" -o json | jq ".redirectURIs = [\"https://${kiali_route_url}/api/auth/callback/${CLUSTER2_CONTEXT}\"]" | ${CLIENT_EXE} --force=true --context "${CLUSTER2_CONTEXT}" apply -f -
+  else
   # Helm chart doesn't support passing in service opts so patch them after the helm deploy.
```
jmazzitelli, line 172: Is this useful? This helm value does set the service port: https://github.com/kiali/helm-charts/blob/v1.84.0/kiali-server/templates/service.yaml#L36

hack/istio/multicluster/install-multi-primary.sh
```diff
@@ -242,8 +242,8 @@ source ${SCRIPT_DIR}/split-bookinfo.sh
 # Install Kiali if enabled
 if [ "${KIALI_ENABLED}" == "true" ]; then
   if [ -z "${KEYCLOAK_ADDRESS}" ]; then
-    echo "Keycloak is not available for this cluster setup. Switching Kial to 'anonymous' mode."
-    export KIALI_AUTH_STRATEGY="anonymous"
+    echo "Keycloak is not available for this cluster setup. Switching Kiali to 'anonymous' mode."
+    # export KIALI_AUTH_STRATEGY="anonymous"
```
jmazzitelli, line 246: I remember having to add this very recently. Are you sure we do not need this anymore?

hack/istio/multicluster/kiali-prepare-remote-cluster.sh
jmazzitelli: Note also - this is an customer-facing script that people in the community are using now. Is this going to cause a problem if we change the name on them?

handlers/authentication.go
```diff
@@ -1,72 +1,106 @@
...
 
 type AuthInfo struct {
-	Strategy              string      `json:"strategy"`
-	AuthorizationEndpoint string      `json:"authorizationEndpoint,omitempty"`
-	LogoutEndpoint        string      `json:"logoutEndpoint,omitempty"`
-	LogoutRedirect        string      `json:"logoutRedirect,omitempty"`
-	SessionInfo           sessionInfo `json:"sessionInfo"`
-	SecretMissing         bool        `json:"secretMissing,omitempty"`
+	Strategy                        string            `json:"strategy"`
+	AuthorizationEndpoint           string            `json:"authorizationEndpoint,omitempty"`
```
jmazzitelli, line 28: Is AuthorizationEndpoint still needed now that we have AuthorizationEndpointPerCluster? Perhaps a comment here to explain what this is now (I assume it is the endpoint for the "home cluster" but that's just a guess).

hack/istio/multicluster/deploy-kiali.sh
nrfox: What this is doing is creating the `OAuthClient` on the _remote_ cluster which the helm chart can't do since it only deals with the cluster kiali is deployed onto. There needs to be an `OAuthClient` on each openshift cluster since that's what registers the kiali app as an oauth client of that cluster's oauth server. This requirement needs to be documented somewhere and it might be worth providing a script that does this or else modifying `kiali-prepare-remote-cluster` to also do this when the target cluster is openshift.

hack/istio/multicluster/install-multi-primary.sh
nrfox: Yes because we don't want `anonymous` we want the auth strategy we passed in which is `openshift`. This should be commented out so I'll look at cleaning it up. Why was it changed to `anonymous` instead of the passed in auth strategy?

handlers/authentication.go
nrfox: I guess technically no but it's basically the auth endpoint for the cluster kiali is deployed on and one of the current requirements of this feature is that you login to that cluster so it's nice to have it separate as it requires less changes on the frontend that way.

hack/istio/multicluster/deploy-kiali.sh
jmazzitelli: Ah yes... that explains things... https://www.gnu.org/software/make/manual/html_node/Environment.html

hack/istio/multicluster/deploy-kiali.sh
```diff
@@ -81,6 +81,8 @@ deploy_kiali() {
         "--set deployment.image_version=dev"
       )
     elif [ "${KIALI_AUTH_STRATEGY}" == "openshift" ]; then
+      echo "Pushing the images into the cluster..."
+      make -e DORP="${DORP}" CLUSTER_TYPE="openshift" cluster-status
```
jmazzitelli, line 85: Don't forget the `-C` like what is being done up on line 77 and down on line 89:

```suggestion
      make -e -C "${KIALI_REPO_ROOT}" DORP="${DORP}" CLUSTER_TYPE="openshift" cluster-status
```

hack/istio/multicluster/deploy-kiali.sh
jmazzitelli: Yeah, we should create a github issue on that so we remember to do this. Or, make that change in this PR. Up to you if you want to implement that in this PR or defer that and create a github issue so we don't forget

handlers/authentication.go
jmazzitelli: I would suggest adding a comment here so future devs reading this will understand. Otherwise, the name `AuthorizationEndpoint` doesn't explain this is the endpoint for the cluster that Kiali is deployed in. For example:

```suggestion
	// this is the endpoint for the cluster in which Kiali is deployed
	AuthorizationEndpoint           string            `json:"authorizationEndpoint,omitempty"`
```

leandroberetta:
Hi, @nrfox I started playing with the code and I see that when I'm not logged in into west, the Mesh page throws an error:

image

Shouldn't I see east's mesh page even with west's cluster unavailable?

jmazzitelli:
I'm trying to run this using the test procedures. I see the option `--kiali-use-dev-image true` is used. I assume this means we need to "make build build-ui cluster-push` to BOTH clusters right? Because Kiali is deployed to both so we need to push the kiali image to both (??)

jmazzitelli:
Just noticed this error in the first script output:
```
error: strconv.Atoi: parsing "http2": invalid syntax
```

jmazzitelli:
And I got a failure and the script exits - this is during step 3 of the test procedures:

```
==== CREATE CROSSNETWORK GATEWAY ON CLUSTER #1 [east] - east
Switched to context "east".
✘ Ingress gateways encountered an error: failed to wait for resource: resources not ready after 5m0s: context deadline exceeded                               
                               
- Pruning removed resources                                                                                                                                   Error: failed to install manifests: errors occurred during operation
Failed to install crossnetwork gateway on cluster [east]
```

nrfox:
> And I got a failure and the script exits - this is during step 3 of the test procedures:
> 
> ```
> ==== CREATE CROSSNETWORK GATEWAY ON CLUSTER #1 [east] - east
> Switched to context "east".
> ✘ Ingress gateways encountered an error: failed to wait for resource: resources not ready after 5m0s: context deadline exceeded                               
>                                
> - Pruning removed resources                                                                                                                                   Error: failed to install manifests: errors occurred during operation
> Failed to install crossnetwork gateway on cluster [east]
> ```
> 
> Failure due to:
> 
> ```
>     message: 'pods "istio-eastwestgateway-7d9c8cb77b-" is forbidden: unable to validate
>       against any security context constraint: [provider "anyuid": Forbidden: not
>       usable by user or serviceaccount, provider "istio-addons-openshift-scc": Forbidden:
>       not usable by user or serviceaccount, provider restricted-v2: .containers[0].runAsUser:
>       Invalid value: 1337: must be in the ranges: [1000700000, 1000709999], provider
>       "restricted": Forbidden: not usable by user or serviceaccount, provider "nonroot-v2":
>       Forbidden: not usable by user or serviceaccount, provider "nonroot": Forbidden:
>       not usable by user or serviceaccount, provider "hostmount-anyuid": Forbidden:
>       not usable by user or serviceaccount, provider "machine-api-termination-handler":
>       Forbidden: not usable by user or serviceaccount, provider "hostnetwork-v2":
>       Forbidden: not usable by user or serviceaccount, provider "hostnetwork": Forbidden:
>       not usable by user or serviceaccount, provider "hostaccess": Forbidden: not
>       usable by user or serviceaccount, provider "node-exporter": Forbidden: not usable
>       by user or serviceaccount, provider "privileged": Forbidden: not usable by user
>       or serviceaccount]'
> ```
> 
> OpenShift version:
> 
> ```
>  oc version
> Client Version: 4.15.0-202404260736.p0.gfa6afba.assembly.stream.el8-fa6afba
> Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
> Server Version: 4.15.9
> Kubernetes Version: v1.28.7+c1f5b34
> ```

Yeah this is something I forgot to address. On openshift the generated east-west gateway yaml doesn't work without removing the `runAs` fields from the pod security context. I need to update this PR to add an additional step to remove those fields from the manifest before applying them.

nrfox:
> I'm trying to run this using the test procedures. I see the option `--kiali-use-dev-image true` is used. I assume this means we need to `make build build-ui` and then `make cluster-push` to BOTH clusters right? Because Kiali is deployed to both so we need to push the kiali image to both (??)

No kiali is just deployed to one not both but I think the scripts should still do a `make build build-ui` before pushing to the east cluster.

nrfox:
> Hi, @nrfox I started playing with the code and I see that when I'm not logged in into west, the Mesh page throws an error:
> 
> image
> 
> Shouldn't I see east's mesh page even with west's cluster unavailable? (I know this is not probably related to this PR, probably it's something to fix in a follow up PR, just commenting).

Hmmm that looks like a bug. You should still see `east` without having to login to `west`.

nrfox:
@jmazzitelli I fixed the install issues and updated the instructions. It should properly install and build now. I added building as a separate flag because in CI the frontend and backend assets are built in a separate step / script than the deploy script.

nrfox:
> Hi, @nrfox I started playing with the code and I see that when I'm not logged in into west, the Mesh page throws an error:

@leandroberetta I think this is unrelated to this PR and occurs on master in non-openshift environments as well. I'll open a separate issue for this.

jmazzitelli:
After logging into West, I see this:
image

My one question is - the name of the east cluster is showing as "Kubernetes". Not sure if that is expected.

jmazzitelli:
Also getting some errors about how it cannot connect to prometheus. e.g.
image

image

Again, not sure if this is expected under certain conditions (I'm running two OpenShift clusters that might not be able to talk to each other :) .

nrfox:
> My one question is - the name of the east cluster is showing as "Kubernetes". Not sure if that is expected.

What's the name of the cluster in the config and on the istiod deployment?

jmazzitelli:
> What's the name of the cluster in the config and on the istiod deployment?

I didn't configure anything (outside of what the scripts already do) so I assume "Kubernetes" is the name (since its the default).

handlers/authentication.go
jmazzitelli: I'd still recommend this comment or something like it so people aren't confused and wonder what it is for

jmazzitelli (commented):
This doesn't seem like it should happen. After I ran the script, I found a change in a file:

```diff
$ git diff
diff --git a/hack/istio/multicluster/prometheus.yaml b/hack/istio/multicluster/prometheus.yaml
index f9d5cffa9..4c5d3249c 100644
--- a/hack/istio/multicluster/prometheus.yaml
+++ b/hack/istio/multicluster/prometheus.yaml
@@ -27,7 +27,7 @@ data:
 
       static_configs:
         - targets:
-          - 'WEST_PROMETHEUS_ADDRESS:9090'
+          - ':9090'
           labels:
             cluster: 'CLUSTER_NAME'
     - job_name: prometheus
```

We don't want these scripts to modify another script (unless it later reverts that change so people aren't caught off guard when git tells them a file has changed).

jmazzitelli (changes requested):
Molecule test failures.

First, the good news - I did test `token-test` on OpenShift and minikube, and that passes both times.

However, `openid-test` on minikube now failed:

```
TASK [Assert that the auth endpoint returned valid data] ***********************^M
fatal: [localhost]: FAILED! => {
   "assertion": "kiali_output.set_cookie is search(\"kiali-token-openid-nonce\")",
   "changed": false,
```

`header-auth-test` also failed on minikube:
```
TASK [Make a request that requires authentication (should now return 200)] *****
fatal: [localhost]: FAILED! => {"changed": false, "connection": "close", "content": "Unauthorized\n", "content_length": "13", "content_type": "text/plain; charset=utf-8", "date": "Thu, 20 Jun 2024 20:58:16 GMT", "elapsed": 0, "msg": "Status code was 401 and not [200]: HTTP Error 401: Unauthorized", "redirected": false, "status": 401, "url": "https://192.168.39.14/kiali/api/namespaces", "vary": "Accept-Encoding", "x_content_type_options": "nosniff"}
```

How to test:
1. Make sure the helm-chart repo has the PR checked out and the helm chart repo is set up to pull from the remote (use `git branch --set-upstream-to=...`)
2. Set things up 
```
./hack/k8s-minikube.sh --dex-enabled true -mp ci start && ./hack/k8s-minikube.sh -mp ci istio && make -e CLUSTER_TYPE=minikube -e MINIKUBE_PROFILE=ci build build-ui cluster-push
```
3. Run tests:
```
./hack/run-molecule-tests.sh --client-exe "$(which kubectl)"  --minikube-exe $(which minikube) --minikube-profile ci --cluster-type minikube --use-dev-images true -at "openid-test header-auth-test token-test"
```

jmazzitelli (dismissed):
with [this molecule test update](https://github.com/kiali/kiali-operator/pull/781), the openid-test, header-auth-test and token-test all pass.
LGTM
```
=====================
=== TEST RESULTS: ===
=====================

                             openid-test... success [1m 27s]
                        header-auth-test... success [1m 28s]
                              token-test... success [1m 48s]
```

jmazzitelli:
when this PR gets merged, we also want this merged since it changes a molecule test related to this PR:
* https://github.com/kiali/kiali-operator/pull/781

leandroberetta (dismissed):
Looks good and works.

jmazzitelli (approved):

PR #7368, opened 2024-05-20
Initial load graph or mesh on componentDidMount

ferhoyos:
### Describe the change

The initial loading of the mesh and graph data is performed in the `componentDidUpdate` method to ensure that everything is set before the mesh loads. The issue is that the `componentDidUpdate` method is not called in OSSMC during the initial mesh load, so the page displays the `Loading` message until the user performs an action that triggers the componentDidUpdate method.

image

This PR also sets some missing redux properties in the mesh page and fix lint issues.

### Steps to test the PR

This PR only affects OSSMC. Kiali standalone should work as usual.

### Automation testing

N/A

### Issue reference

https://github.com/kiali/openshift-servicemesh-plugin/issues/308

### DIFF
--- a/frontend/src/pages/Graph/GraphPage.tsx
+++ b/frontend/src/pages/Graph/GraphPage.tsx
@@ -375,15 +375,15 @@ class GraphPageComponent extends React.Component<GraphPageProps, GraphPageState>
     if (urlTrace !== this.props.trace?.traceID) {
       this.props.setTraceId(urlTrace);
     }
+
+    // Ensure we initialize the graph. We wait for the toolbar to render and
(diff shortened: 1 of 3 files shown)


jshaughn (dismissed):
The old `setTimeout() to 0` trick again!  I wish I had known/thought of that in the past, clean and solves lots of various problems :-)  Tested, seems to work well without the need for the `isInitialLoad` hack.

ferhoyos:
Yes, it is a very useful trick to have in your UI development toolset :-)

jshaughn (approved):
Re-approving

Issue #7203, opened 2024-03-19
Per controlplane istio settings


nrfox:
### What do you want to improve?

There's a number of istio settings that Kiali exposes like mTLS for the namespace/cluster/mesh. These need to take into account multiple controlplanes whether that's single cluster when multiple revisions are present or multi-cluster with a multi-primary deployment.

### What is the current behavior?

Multiple controlplanes are not used when reporting istio settings.

### What is the new behavior?

All relevant controlplanes are used when reporting istio settings.

jshaughn:
I think this is really "Determine correct Istio versions for multi-primary scenarios" as I'm not aware of what other Istio settings we're talking about in the description, or not already handled by the new mesh page.

nrfox:
I'd say it's bigger than that. It's many of the settings under [external_services.istio](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.istio) like `config_map_name`.  Some of these settings are used in various places like validations: https://github.com/kiali/kiali/blob/976461bbf9099f4f6228442f918c106ae8d0a9dc/business/istio_validations.go#L407 and probably would lead to incorrect results if they differ between primaries and/or revisions.

jshaughn:
Thanks for the clarification, @nrfox.   It sounds almost like this should be an epic, with different information being handled as individual issues.  I had currently been treating this one as a follow-up for istio version.

jshaughn:
@nrfox You've been working in this area already, I've moved this to the sprint backlog and assigned you, does that make sense?

nrfox:
> There's a lot of TODO: multi-primary support littered around the backend right now.

Actually it'd probably be better to create a new issue for this. This issue is about properly displaying per controlplane settings which we currently do so it can be closed.

PR #7370, opened 2024-05-21
Correct Kubernetes versions for multi-primary scenarios

### LINKED ISSUE #7361: [mesh page] show correct istio and cluster versions in multi-cluster deployments
Epic: https://github.com/kiali/kiali/issues/5913

Currently the mesh page is showing the version for the home Kiali's istio and cluster.  This may not be correct and versions should reflect reality.

ferhoyos:
### Describe the change

This PR adds support for multi-primary scenarios, providing correct versions for Kubernetes in the mesh page. 

### Steps to test the PR

1. Deploy a multi-primary scenario 
2. Go to Mesh page
3. Verify that Kubernetes versions are the correct ones

image

image

### Automation testing

N/A

### Issue reference

Fixes #7361

### DIFF
--- a/frontend/public/locales/zh/translation.json
+++ b/frontend/public/locales/zh/translation.json
@@ -65,6 +65,7 @@
   "Kiali home cluster": "Kiali home cluster",
   "Kiali home cluster: {{name}}": "Kiali主集群: {{name}}",
   "Kiali on GitHub": "Kiali on GitHub",
+  "Kubernetes": "Kubernetes",
   "Last": "最近的",
   "List view": "列表视图",
   "Loading...": "Loading...",
--- a/frontend/src/pages/Mesh/target/TargetPanelMesh.tsx
+++ b/frontend/src/pages/Mesh/target/TargetPanelMesh.tsx
@@ -36,21 +36,24 @@ export const TargetPanelMesh: React.FC<TargetPanelMeshProps> = (props: TargetPan
     const clusterDataPlanes = dataPlaneNodes.filter(node => node.getData().cluster === clusterData.cluster);
 
     return (
-      <div style={{ marginBottom: '1rem' }}>
+      <div key={clusterData.id} style={{ marginBottom: '1rem' }}>
         {renderNodeHeader(clusterData, { nameOnly: true, smallSize: false, hideBadge: clusterData.isExternal })}
         <div className={infoStyle}>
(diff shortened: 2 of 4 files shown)


ferhoyos:
Currently Istio version from remote cluster cannot be obtained because Istio settings are only supported for single-cluster (related to issue https://github.com/kiali/kiali/issues/7203). I think that for the moment we can fix only Kubernetes version and assume that remote Istio version is the same as home cluster one.

frontend/public/locales/zh/translation.json
```diff
@@ -65,6 +65,7 @@
   "Kiali home cluster": "Kiali home cluster",
   "Kiali home cluster: {{name}}": "Kiali主集群: {{name}}",
   "Kiali on GitHub": "Kiali on GitHub",
+  "Kubernetes": "Kubernetes",
```
jshaughn, line 68: I guess this maybe has a translation into other character sets?  I don't know, not a problem to include it even if they leave it alone.

frontend/src/pages/Mesh/target/TargetPanelMesh.tsx
```diff
@@ -62,7 +65,7 @@ export const TargetPanelMesh: React.FC<TargetPanelMeshProps> = (props: TargetPan
 
   const renderInfraNodeSummary = (nodeData: MeshNodeData): React.ReactNode => {
     return (
-      <div className={summaryStyle}>
+      <div key={nodeData.id} className={summaryStyle}>
```
jshaughn, line 68: +1, avoid key conflict.

jshaughn (approved):
Tested on my single OCP 4.14 CRC cluster and working fine in that scenario.  I'll leave the changes to version determination to Nick as I know you two were going back on forth on it.  +1 to limiting to kube version.  Also, the CI fails are all know flakes so no problem there.  LGTM.

image

frontend/public/locales/zh/translation.json
ferhoyos: I don't know, probably there is no translation. I just changed `Version` to `Kubernetes` and I did not care about the translation. Anyway, if that is the case we can remove it afterwards.

jshaughn:
Merging as CI failures are known flakes and @ferhoyos  has said it is good to merge.

Issue #7355, opened 2024-05-14
K8s GW API 1.1 Support


hhovsepy:
### What do you want to improve?
K8s Gateway API v1.1.0 is released https://github.com/kubernetes-sigs/gateway-api/releases/tag/v1.1.0
GRPCRoute is GA now.

- [x] GRPCRRoute to v1.1
- [x] Add support of GRPCRoute in Wizards. https://github.com/kiali/kiali/issues/7223
- [x] HTTPRoute can now have a Service as a parentRef. Update validations.

jshaughn:
@hhovsepy I force-merged PRs for this, but was unsure whether this issue itself was complete.  Please update as needed.

Issue #7223, opened 2024-03-26
[GW] GRPCRoute support


hhovsepy:
K8s Gateway API v1.1.0 will come with GRPCRoute support.

More info here https://gateway-api.sigs.k8s.io/guides/grpc-routing/

Kiali should support it as well in:

- [x] Istio Config List
- [x] Services wizards

PR #7373, opened 2024-05-21
Adding HTTPS and TLS into K8s Gateway creation wizard

hhovsepy:
### Describe the change
In K8s Gateway creation wizard, added support of 'HTTPS'  protocol and 'TLS Certificate' input.
Done in a scope of GRPC Routing https://gateway-api.sigs.k8s.io/guides/grpc-routing/

Screenshot from 2024-05-23 14-47-13

### Steps to test the PR

Istio config list -> Actions -> Create K8sGateway.
Add listener.
Choose HTTPS.
New row appears for TLS params. Currently only 'Terminate' is supported.
Add TLS Certificate.
Preview and save.
In create config object should be 'tls:' field with Kind and name. The 'Terminate' default mode value is not shown, considered as default for 'HTTPS'.
Regression testing: Make sure that "HTTP" protocol mode creation is also working fine.

### Automation testing

cypress test added

### Issue reference
https://github.com/kiali/kiali/issues/7223
Parent epic https://github.com/kiali/kiali/issues/7355

### DIFF
--- a/frontend/src/components/IstioWizards/WizardActions.ts
+++ b/frontend/src/components/IstioWizards/WizardActions.ts
@@ -983,7 +983,8 @@ export const buildIstioConfig = (wProps: ServiceWizardProps, wState: ServiceWiza
                         matchLabels: {}
                       }
                     }
-                  }
+                  },
+                  tls: null
                 }
               ]
             }
@@ -2154,9 +2155,9 @@ export const addAnnotations = (istioObject: IstioObject, value: { [key: string]:
   istioObject.metadata.annotations = value;
 };
 
-export const getTLS = (tls?: K8sGatewayTLS): K8sGatewayTLS | undefined => {
+export const getTLS = (tls?: K8sGatewayTLS | null): K8sGatewayTLS | null => {
   if (!tls) {
-    return undefined;
+    return null;
   }
   return {
     certificateRefs: tls.certificateRefs.map(c => ({
--- a/frontend/src/components/IstioWizards/WizardActions.ts
+++ b/frontend/src/components/IstioWizards/WizardActions.ts
@@ -25,6 +25,7 @@ import {
   K8sHTTPRouteMatch,
   K8sHTTPRouteRequestRedirect,
   K8sReferenceGrant,
+  Listener,
   LoadBalancerSettings,
   Operation,
   OutlierDetection,
@@ -1868,27 +1869,43 @@ export const buildK8sGateway = (
     spec: {
       // Default for istio scenarios, user may change it editing YAML
       gatewayClassName: state.gatewayClass,
-      listeners: state.listeners.map(s => ({
-        name: s.name,
-        port: s.port,
-        protocol: s.protocol,
-        hostname: s.hostname,
-        allowedRoutes: {
-          namespaces: {
-            from: s.allowedRoutes.namespaces.from,
-            selector: {
-              matchLabels: s.allowedRoutes.namespaces.selector?.matchLabels
+      listeners: state.listeners.map(s => {
+        const listener: Listener = {
+          name: s.name,
+          port: s.port,
+          protocol: s.protocol,
+          hostname: s.hostname,
+          allowedRoutes: {
+            namespaces: {
+              from: s.allowedRoutes.namespaces.from
             }
           }
-        },
-        tls: getTLS(s.tls)
-      })),
-      addresses: state.addresses.map(s => ({
-        type: s.type,
-        value: s.value
-      }))
+        };
+
+        if (s.allowedRoutes.namespaces.from === 'Selector' && s.allowedRoutes.namespaces.selector?.matchLabels) {
+          listener.allowedRoutes.namespaces.selector = {
+            matchLabels: s.allowedRoutes!.namespaces.selector?.matchLabels
+          };
+        }
+
+        const tls = getTLS(s.tls);
+
+        if (tls) {
+          listener.tls = tls;
+        }
+
+        return listener;
+      })
     }
   };
+
+  if (state.addresses.length > 0) {
+    k8sGateway.spec.addresses = state.addresses.map(s => ({
+      type: s.type,
+      value: s.value
+    }));
+  }
+
   addLabels(k8sGateway, labels);
   addAnnotations(k8sGateway, annotations);
 
--- a/frontend/src/components/IstioWizards/WizardActions.ts
+++ b/frontend/src/components/IstioWizards/WizardActions.ts
@@ -16,6 +16,7 @@ import {
   HTTPRouteDestination,
   IstioObject,
   K8sGateway,
+  K8sGatewayTLS,
   K8sHTTPHeaderFilter,
   K8sHTTPMatch,
   K8sHTTPRequestMirrorFilter,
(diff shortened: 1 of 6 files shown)


ferhoyos (changes requested):
Good job @hhovsepy adding the HTTPS and TLS certificates to K8s Gateway wizard!
Overall works fine, I am able to create, edit and delete K8s gateway with HTTPS port listener and TLS certificate.

Some comments related to this PR:
- `tls` field is optional and it should not be added to the preview editor in case this field is empty:

image

Comments not related to this PR (possible new Kiali issues)
- The validation of the label selector within the listener does not work, it should have the format `label=value,label2=value2` but it allows any string. IMO it would be better to add the label wizard here

image

- Label selector text input should be enabled only for value `Selector` in `From Namespaces` field (https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.RouteNamespaces) 

- Empty optional values should be removed from preview yaml (e.g., addresses, selector when fromNamespaces =! selector, allowedRoutes if fromNamespaces == All, etc). I think it only adds noise, although in most cases they will be finally removed by Kubernetes when the preview yaml is applied.

- Close preview editor only when the creation process is successful. If there is a creation error because of a validation spec not covered by the wizard, the preview editor is closed and the user does not know why it failed. Besides, it is not possible to fix the validation issue and the user has to create the object again with the wizard. One validation example is try to create two listeners with same hostname, port and protocol within the gateway (https://gateway-api.sigs.k8s.io/reference/spec/#gateway.networking.k8s.io/v1.GatewaySpec)

frontend/src/components/IstioWizards/WizardActions.ts
```diff
@@ -982,7 +983,8 @@ export const buildIstioConfig = (wProps: ServiceWizardProps, wState: ServiceWiza
                         matchLabels: {}
                       }
                     }
-                  }
+                  },
+                  tls: null
```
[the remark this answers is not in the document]
hhovsepy: there was some design issue with tls field, when optional and undefined, json parser issues were happening

hhovsepy:
Thank you @ferhoyos for the review, approached comments,
only the form submit/close preview is outside of this scope,,
the labels validation is done inline way as in other forms
Screenshot from 2024-05-29 15-06-35

frontend/src/pages/IstioConfigNew/GatewayForm/ListenerBuilder.tsx
```diff
@@ -78,83 +99,140 @@ export const ListenerBuilder: React.FC<ListenerBuilderProps> = (props: ListenerB
...
+          </FormSelect>
+        </Td>
+
+        <Td>
+          <TextInput
+            id={`addSelectorLabels_${props.index}`}
+            name="addSelectorLabels"
+            isDisabled={props.listener.from !== SELECTOR}
+            onChange={onAddSelectorLabels}
+            validated={isValid(isValidSelector(props.listener.sSelectorLabels))}
```
ferhoyos, line 184:
```suggestion
            validated={isValid(
              props.listener.from === SELECTOR ? isValidSelector(props.listener.sSelectorLabels) : undefined
            )}
```
Apply validation only if the text input is enabled

frontend/src/pages/IstioConfigNew/GatewayForm/ListenerBuilder.tsx
```diff
@@ -27,12 +32,28 @@ export const isValidHostname = (hostname: string): boolean => {
...
+  return true;
+};
+
 export const isValidPort = (port: string): boolean => {
   return port.length > 0 && !isNaN(Number(port)) && Number(port) >= MIN_PORT && Number(port) <= MAX_PORT;
 };
 
 export const isValidSelector = (selector: string): boolean => {
-  return selector.length === 0 || typeof addSelectorLabels(selector) !== 'undefined';
+  return selector.length === 0 || addSelectorLabels(selector)[0];
```
ferhoyos, line 56:
```suggestion
  return addSelectorLabels(selector)[0];
```
Now that the label input is only enabled for `Selector` option, empty value should be invalid.

frontend/src/pages/IstioConfigNew/GatewayForm/ListenerList.tsx
```diff
@@ -55,11 +55,12 @@ const columns: ThProps[] = [
   }
 ];
 
-export const addSelectorLabels = (value: string) => {
+export const addSelectorLabels = (value: string): [boolean, {}] => {
   if (value.length === 0) {
-    return;
+    return [true, {}];
```
ferhoyos, line 60:
```suggestion
    return [false, {}];
```
Now that the label input is only enabled for `Selector` option, empty value should be invalid.

frontend/src/pages/IstioConfigNew/GatewayForm/ListenerList.tsx
```diff
@@ -108,7 +114,8 @@ export const ListenerList: React.FC<ListenerListProps> = (props: ListenerListPro
       port: 70000,
       name: '',
       protocol: protocols[0],
-      allowedRoutes: { namespaces: { from: allowedRoutes[0], selector: { matchLabels: {} } } }
+      allowedRoutes: { namespaces: { from: allowedRoutes[0], selector: { matchLabels: {} } } },
+      tls: null
```
ferhoyos, line 118:
```suggestion
      allowedRoutes: { namespaces: { from: allowedRoutes[0] } }
```
Remove empty values by default

frontend/src/pages/IstioConfigNew/GatewayForm/ListenerList.tsx
```diff
@@ -144,16 +151,28 @@ export const ListenerList: React.FC<ListenerListProps> = (props: ListenerListPro
     if (listenerForm.port.length === 0 || isNaN(Number(listenerForm.port))) return;
     if (listenerForm.hostname.length === 0) return;
 
-    const selector = addSelectorLabels(listenerForm.sSelectorLabels) || {};
+    const selector = addSelectorLabels(listenerForm.sSelectorLabels)[1] || {};
 
     const listener: Listener = {
```
ferhoyos, line 156: I don't understand this part of the code... It seems that we are creating twice the listener yaml... once here and the same in `WizardActions.buildK8sGateway` method.

IMO I would create the yaml here and then in WizardActions just map the listeners as is:
```
    spec: {
      // Default for istio scenarios, user may change it editing YAML
      gatewayClassName: state.gatewayClass,
      listeners: state.listeners
    }
  };

  if (state.addresses.length > 0) {
    k8sGateway.spec.addresses = state.addresses;
  }
```

hhovsepy: good point

frontend/src/components/IstioWizards/WizardActions.ts
```diff
@@ -982,7 +984,8 @@ export const buildIstioConfig = (wProps: ServiceWizardProps, wState: ServiceWiza
...
```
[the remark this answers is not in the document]
hhovsepy: right, missed that

frontend/src/types/IstioObjects.ts
```diff
@@ -783,6 +783,7 @@ export interface Listener {
   name: string;
   port: number;
   protocol: string;
+  tls?: K8sGatewayTLS | null;
```
hhovsepy, line 786: @ferhoyos it seems null is odd here?

ferhoyos: yes, now that it is optional you can remove the null here

hhovsepy:
Thank you @ferhoyos , it seems comments are approached, and the issue with K8sGateway in wizard is fixed.

PR #7381, opened 2024-05-22
maybe namespace is omitting the bookinfo traces and setting to 1 percent

jshaughn:


nrfox (commented):
I think this just sets the telem for the `default` namespace rather than setting the root configuration. It seems to have made it worse based on the number of failures which is useful to know because, to me, it implies that the sampling percentage is being set correctly.

jshaughn:
experiment failed

PR #7393, opened 2024-05-28
GRPC Route wizard in Service Actions

hhovsepy:
### Describe the change

In Service details Action menu, added "Create/Update K8s GRPC Routing" items.
Updated "K8s Gateway API Routing" to "K8s HTTP Routing".
Added 'GRPC Routing' wizard.

Screenshot from 2024-06-11 18-36-52

GRPCRoute object has validations and references.

Screenshot from 2024-06-11 18-38-53

Screenshot from 2024-06-11 18-37-18

Service list with gRPC route configs:
Screenshot from 2024-06-11 18-37-06

### Steps to test the PR

Check the functionality of Service wizard "K8s GRPC Routing" in create/update/delete.
Verify that GRPCRoute has validations, 'KIA1402' and 'KIA1401', like in HTTPRoute.

### Automation testing

added e2e, unit and cypress tests

### Issue reference
https://github.com/kiali/kiali/issues/7223


josunect (commented):
Testing the PR, I've found the following issue: 
When I try to add a GRPC Routing with a header match, I've got an error: 

image

And an error in the logs:

`2024-06-13T16:17:19+01:00 ERR GRPCRoute.gateway.networking.k8s.io "details" is invalid: spec.rules[0].matches[0].headers[0].value: Invalid value: "": spec.rules[0].matches[0].headers[0].value in body should be at least 1 chars long`

It looks like the problem is that is adding the value as an empty string (Or maybe I'm not adding the correct value, but I think it should be validated at least in that case): 

image

Once created, I cannot see the GRPC route: 

image

And for some reason, it is not detecting it: 

image

But it was created: 

image

josunect:
Testing the latest changes, it works as expected: 

image
image
image

It is not possible to create the route if there is no header value: 
image

I think it should be more accurate if the button "Add Match" is disabled until you don't have valid values in the fields. So it would be easier for the user to identify what is the issue, if the wizard doesn't let them add a Match if the match is not valid. What do you think, @hhovsepy ? (And not just the "Add Route Rule")

hhovsepy:
Thanks @josunect , good catch, there was a bug in code, now "Add Match" is disabled when invalid. Also fixed the "method" match key/values.

business/references/k8s_gateway_references.go
```diff
@@ -39,5 +40,16 @@ func (g K8sGatewayReferences) getConfigReferences(gw *k8s_networking_v1.Gateway)
 		}
 	}
 
+	for _, rt := range g.K8sGRPCRoutes {
+		if len(rt.Spec.ParentRefs) > 0 {
```
jshaughn, line 44: Is this check necessary?  Wouldn't the range just not execute if len is 0?  I'm fairly sure that range on a nil slice is fine, as internally it would just be zero iterations.

business/references/k8sgrpcroute_references.go
```diff
@@ -0,0 +1,88 @@
...
+			fqdn := kubernetes.GetHost(string(ref.Name), namespace, n.Namespaces.GetNames())
+			if !fqdn.IsWildcard() {
+				allServices = append(allServices, models.ServiceReference{Name: fqdn.Service, Namespace: fqdn.Namespace})
+			}
+		}
+	}
+
+	// filter unique references
+	for _, sv := range allServices {
+		if !keys[sv.Name+"."+sv.Namespace] {
```
jshaughn, line 54: A small function to generate these various keys may be safer than all of the manual concats.

business/references/k8shttproute_references.go
```diff
@@ -74,13 +74,13 @@ func (n K8sHTTPRouteReferences) getConfigReferences(rt *k8s_networking_v1.HTTPRo
 	return result
 }
 
-func getAllK8sGateways(rt *k8s_networking_v1.HTTPRoute) []models.IstioReference {
+func getAllK8sGateways(prs []k8s_networking_v1.ParentReference, ns string) []models.IstioReference {
 	allGateways := make([]models.IstioReference, 0)
 
-	if len(rt.Spec.ParentRefs) > 0 {
-		for _, parentRef := range rt.Spec.ParentRefs {
+	if len(prs) > 0 {
```
jshaughn, line 80: Same comment, I don't think you need this `if`.

frontend/src/components/IstioWizards/ConfirmDeleteTrafficRoutingModal.tsx
```diff
@@ -53,6 +54,14 @@ export const ConfirmDeleteTrafficRoutingModal: React.FunctionComponent<Props> =
         : '';
     deleteItems.push(<div key={`delete_item_${++i}`}>{k8sHTTPRouteMessage}</div>);
 
+    let k8sGRPCRouteMessage =
```
jshaughn, line 57:
```suggestion
    const k8sGRPCRouteMessage =
```

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -575,9 +575,9 @@ export class OverviewPageComponent extends React.Component<OverviewProps, State>
                   destinationRules: [],
                   envoyFilters: [],
                   gateways: [],
-                  k8sGRPCRoutes: [],
                   k8sGateways: [],
                   k8sHTTPRoutes: [],
+                  k8sGRPCRoutes: [],
```
jshaughn, line 580:
```suggestion
                  k8sGRPCRoutes: [],
                  k8sHTTPRoutes: [],
```

jshaughn (commented):
Code looks good, a lot of work here!   A couple of minor comments.  Did some light testing and did not find any problems. 

I have one question, probably due to me not knowing enough of GW API...   Working with bookinfo, I go into into the routing wizard (grpc or http) for `Reviews` the service.  The `Route To` tab looks like:
image
Is it expected that I don't see the three versioned workloads here?  I see only:
image

This is a little confusing, because it says `Destination Service`, but shows a `Workload` badge with the service name.  It seems like this should maybe be a `Service` badge, or I should see the different workloads.

business/references/k8sgrpcroute_references.go
hhovsepy: right

business/references/k8shttproute_references.go
hhovsepy: right

business/references/k8s_gateway_references.go
hhovsepy: right

hhovsepy:
Hi @jshaughn , thank you for your review, have applied all requested changes with keys, loops and `Service` badge.
Regarding multiple services, sorry for confusion, did not mention in testing steps, for Gateway API there should be a Service per Workload, so a special configuration should be applied for bookinfo in GW api: https://github.com/istio/istio/blob/master/samples/bookinfo/platform/kube/bookinfo-versions.yaml
Then you can see all workloads while making GW API routing to any of those created 3 services.
Screenshot from 2024-06-18 09-43-33

When you create the routing, the route config will be shown as a reference from all 3 services and vice-versa.
Screenshot from 2024-06-18 09-38-24

I know this is not the perfect solution to apply current wizzard to GW API needs, when several services are tight together.
There was a task to make research and see if we can make better solution, but it will require a bigger refactoring.

ferhoyos (commented):
Good job @hhovsepy! A lot of work has been done here! I have tested the GRPCRoute wizard and I have not found any blocking issues.

Just a few non-blocking comments about all traffic routing wizards (not only GRPC): 
- The selected values for Matching and Filtering do not look good from a UX perspective. When you click them, they are removed, but there is no `X` indicating that they can be removed. Additionally, the cursor changes to `not allowed` when you hover over the label. I suggest adding an `X` to the right, similar to the filter labels on list pages, and changing the cursor type.

image

- Once you add a Route rule to the list, all matching, mirroring and filtering values of the rule should reset to their default values.

image


- The tabs at the top are a bit confusing to me. Maybe it is just me, but when I see tabs at the top, I usually think that each tab represents a global tab. So, in this case, it seems like you can create three kind of different `Route rules` instead of different aspects of the same rule. I think we should discuss this with @fkargbo to determine the best way to display the wizard.

One question: is it unusual to create Istio config objects in application namespaces like `bookinfo`? When I see the following modal, I wonder if I am doing something wrong or if I need to be very careful with it.

image

If it is quite common, I think we should remove that modal and create the Istio config object without confirmation.

Regarding the differences between Istio and Gateway API concerning services when multiple versions exist, maybe we should attach all these wizards to Applications instead of Services. For now, it is okay, but I agree that it is strange to have to go to one of the services to create a GRPC route that affects all services of the application.

jshaughn:
> Just a few non-blocking comments about all traffic routing wizards (not only GRPC):

Good UX comments @ferhoyos , I have always found some of this wizard config creation confusing.  But let's not burden this large PR with optional UX work.  Let's move those comments to a separate issue.

jshaughn (approved):
@hhovsepy Thanks for the explanation about the Services fronting Workloads with the GW API routing.  I'm +1 on your choice to take the current approach and to avoid a larger refactor.  Unless we get negative feedback, I think this is fine.

All good from my side!

hhovsepy:
Thanks for comments @jshaughn  and @ferhoyos.
I will go to merge this PR as it is, as the UX changes Wizard specific, they can be done in a separate PR.
The minor comments will be addressed in https://github.com/kiali/kiali/pull/7426 which is blocked by this PR.

PR #7395, opened 2024-05-29
Fix release test suite images workflow

mkralik3:
### Describe the change

When a new Kiali version is released, this [workflow](https://github.com/kiali/kiali/blob/master/.github/workflows/test-images-release.yml#L6) should build and publish test images as well however, it didn't work ( because it is triggered by tag pushing, not by release, fixed ).

### Example how that workflow works
Let's imagine we release the following kiali versions in a particular time frame
- Kiali v1.85.0 is released
  -  The workflow creates int and cypress test images with `v1.85.0` and `v1.85` tags.
- Kiali v1.85.1 is released
  -  The workflow creates int and cypress test images with `v1.85.1` tag and update images under `v1.85` tag
- Kiali v1.85.2 is released
  -  The workflow creates int and cypress test images with `v1.85.2` tag and update images under `v1.85` tag

So these test images would exist:
v1.85.0
v1.85.1
v1.85.2
v1.85 (same as v1.85.2, so that tag will contain the latest changes for a major version 1.85, [that one is used in lpinterop pipeline])

### DIFF
--- a/.github/workflows/test-images-release.yml
+++ b/.github/workflows/test-images-release.yml
@@ -1,9 +1,8 @@
 name: Release test suite images
 
 on:
-  push:
-    tags:
-      - 'v*'
+  release:
+    types: [published]
 jobs:
   initialize:
     name: Initialize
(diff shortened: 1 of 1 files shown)


ScriptingShrimp (approved):
LGTM

PR #7399, opened 2024-05-30
Support TP releases for OSSM versions

hhovsepy:
OSSM releases with 'TP' inside version were ignoring 'TP'. Now it is also included.

Issue: OSSM-6425


jmazzitelli (approved):
the new unit tests pass - that is saying this is all good.

jshaughn (approved):
Regex looks good.  We need to try the about page and mesh page against OSSM3 just to verify, but that can be done as part of the Jira validation.

PR #7401, opened 2024-05-31
Simplify i18n support

### LINKED ISSUE #7394: Simplify i18n support

Recently I discovered a simple way to perform i18n translations in React classes and external functions. This approach uses the function `getI18n().t` from `react-i18next` directly, instead of adding the `t` property to React classes with the `withTranslation` HOC function. I successfully applied this approach on the mesh page, so it should be propagated to the rest of the Kiali application and the corresponding documentation should be modified..

### What is the current documentation?

https://github.com/kiali/kiali/blob/master/frontend/README.adoc#internationalization-i18n

Currently there are 4 different use cases for i18n translation (React hooks, React classes, external functions and variables)

### What is the new documentation?

With the new approach, we simplify the i18n translation to only 2 different use cases (React hooks and everything else).

ferhoyos:
### Describe the change

To facilitate community contributions for translating Kiali into multiple languages, I have simplified the i18n support in the Kiali application.

Previously, there were multiple methods for translating a label depending on the use case (React hook, React class, external function, and external variables). Now, everything is translated using the `t` function, although initialization differs slightly between React hooks (`useKialiTranslation`) and other cases (directly importing the `t` function from `utils/I18nUtils`). In some React classes, it is necessary to add the language as a Redux property to force the component to re-render when the language changes.

`i18n` section of frontend README has been updated with the simplified process.

Additionally, I have introduced a new utility function, `tMap`, to translate key-value pair objects.

**Bonus:**
- Delete jest.config.ts since we already have the jest testing setup in `setupTests.ts`

### Steps to test the PR

Verify that Chinese language is working as expected (special attention to initial load and switch language).

### Automation testing

N/A

### Issue reference

Closes #7394


jmazzitelli:
> Verify that Chinese language is working as expected (special attention to initial load and switch language).

I don't even the UI control to switch languages. How do you switch languages in the UI now?

ferhoyos:
You're right @jmazzitelli, I have the i18n feature enabled all the time in my local so I always forget to mention it in the test procedure :sweat_smile:. Thanks for reviewing :slightly_smiling_face:

PR #7404, opened 2024-06-04
Go version 1.22.1

josunect:
### Describe the change

- Update go.mod version
- Add a check in the Makefile to make sure both go versions are de same

### Steps to test the PR

- Update GO_VERSION_KIALI = 1.22.2
- Run make 
- It should be an error

### Automation testing
GH Actions

### Issue reference

Failing Actions, ex. https://github.com/kiali/kiali/actions/runs/9361191323/job/25767881628

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,7 +29,7 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
-GO_VERSION_KIALI = $(shell sed -En 's/^go[[:space:]]+([[:digit:].]+)/\1/p' go.mod)
+GO_VERSION_KIALI = $(shell sed -rn 's/^go[[:space:]]+([[:digit:].]+)/\1/p' go.mod)
 GO_TEST_FLAGS ?=
 
 # Identifies the Kiali container image that will be built.
--- a/Makefile
+++ b/Makefile
@@ -242,9 +242,3 @@ help: Makefile
 	  echo "======================================="; \
 	  exit 1; \
 	fi
-
-.check_go_version:
-	@if [ "$(GO_MOD_VERSION)" != "$(GO_VERSION_KIALI)" ]; then \
-		echo "Kiali Go version ${GO_VERSION_KIALI} different than go.mod ${GO_MOD_VERSION}"; \
-		exit 1; \
-	fi
--- a/Makefile
+++ b/Makefile
@@ -29,8 +29,16 @@ VERSION_LABEL ?= ${VERSION}
 # The go commands and the minimum Go version that must be used to build the app.
 GO ?= go
 GOFMT ?= $(shell ${GO} env GOROOT)/bin/gofmt
+GO_MOD_VERSION = $(shell sed -En 's/^go[[:space:]]+([[:digit:].]+)/\1/p' go.mod)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -29,8 +29,16 @@ VERSION_LABEL ?= ${VERSION}
...
+GO_MOD_VERSION = $(shell sed -En 's/^go[[:space:]]+([[:digit:].]+)/\1/p' go.mod)
 GO_VERSION_KIALI = 1.22.1
 GO_TEST_FLAGS ?=
+all: check_version
+
+check_version:
+	@if [ "$(GO_MOD_VERSION)" != "$(GO_VERSION_KIALI)" ]; then \
+		echo "Kiali Go version ${GO_VERSION_KIALI} different than go.mod ${GO_MOD_VERSION}"; \
+		exit 1; \
+	fi
```
jmazzitelli, line 41: You will want to get rid of that `all` target, and move the check code into the `go-check` target found in `make/Makefile.build.mk` here: https://github.com/kiali/kiali/blob/master/make/Makefile.build.mk#L24-L28

ferhoyos:
Why do we need to define the go version in the Makefile manually? Couldn't we use the go version defined in go.mod as only source of truth and avoid to check if both go versions are the same?

jmazzitelli:
> Why do we need to define the go version in the Makefile manually? Couldn't we use the go version defined in go.mod as only source of truth and avoid checking if both go versions are the same?

Good point. We only have it in Makefile for legacy reasons - it was there long before we were using go modules. Now that we have go.mod and it, itself, defines the go version that we want, we should just use that.

Just make sure no other make targets, Dockerfiles, hack scripts, or things like that use the Makefile GO_VERSION.

josunect:
> > Why do we need to define the go version in the Makefile manually? Couldn't we use the go version defined in go.mod as only source of truth and avoid checking if both go versions are the same?
> 
> Good point. We only have it in Makefile for legacy reasons - it was there long before we were using go modules. Now that we have go.mod and it, itself, defines the go version that we want, we should just use that.
> 
> Just make sure no other make targets, Dockerfiles, hack scripts, or things like that use the Makefile GO_VERSION.

Make we can just set `GO_VERSION_KIALI = $(shell sed -En 's/^go[[:space:]]+([[:digit:].]+)/\1/p' go.mod)`?

aljesusg:
Yes and we need to make changes in GH actions

ferhoyos:
@josunect there is a simple way to get the go version from `go.mod` in GH actions. Please check `nightly.yml` workflow:
```
- name: Set up Go
      uses: actions/setup-go@v5
      with:
        go-version-file: go.mod
        # The builtin cache feature ensures that installing golangci-lint
        # is consistently fast.
        cache: true
        cache-dependency-path: go.sum
```
The same can be applied to `build-backend` and `release` GH workflows.

josunect:
> @josunect there is a simple way to get the go version from `go.mod` in GH actions. Please check `nightly.yml` workflow:
> 
> ```
> - name: Set up Go
>       uses: actions/setup-go@v5
>       with:
>         go-version-file: go.mod
>         # The builtin cache feature ensures that installing golangci-lint
>         # is consistently fast.
>         cache: true
>         cache-dependency-path: go.sum
> ```
> 
> The same can be applied to `build-backend` and `release` GH workflows.

@ferhoyos we saved one step this way!

ferhoyos (approved):
LGTM

PR #7405, opened 2024-06-04
Promote Istio API to v1

### LINKED ISSUE #7398: Kiali 1.85.0 - Istio 1.22 : display istio resources V1 as V1Beta1
### Describe the bug

With Istio 1.22.0 a lot of resources as VS, DR, SE have been released in version V1. If we install the Kiali as addons with the helm chart in version 1.85.0, the istio resources are displayed in the Kiali UI with the version V1Beta1.
If we get the istio resources with the kubectl get vs, dr, se -A the version of all resources are V1.

### Expected Behavior
The version of the Istio resources displayed in the Kiali UI should be V1 and not V1Beta1

1. Install Istio in version 1.22.0
2. Install Kiali with the helm chart in version 1.85.0
3. Deploy a SE as below in your cluster:
```yaml
kind: ServiceEntry
metadata:
  name: keycloak
  namespace: my-ns
spec:
  hosts:
    - identity.my-ns.xxxxxxx
  ports:
    - number: 443
      protocol: HTTPS
      name: https
  resolution: DNS
  exportTo:
    - .
```
4. If you get the resource with kubectl get service-entry keycloak -n my-ns -oyaml
You see the good version V1 for the resources.
5. In the Kiali, in the Istio Config tab, if you select the same resource, the version displayed to the top will be v1beta1
image

### Environment

- **Kiali version: 1.85.0 **
- **Istio version: 1.22.0 **
- **Kubernetes impl: EKS** 
- **Kubernetes version: 1.29**

ferhoyos:
### Describe the change

With Istio 1.22.0 a lot of resources as VS, DR, SE have been released in version V1 (https://medium.com/google-cloud/everything-new-with-istio-1-22-9cded0ccd759). This PR promotes networking, security and telemetry Istio API to v1 in Kiali application.

**Bonus**: Kiali internal API `kiali_internal_api.md` file has been re-generated to update API version

### Steps to test the PR

1. Try to create any Virtual Service, Destination Rule or Service Entry in Kiali
2. Verify that apiVersion is `networking.istio.io/v1`

image

3. The same applies for AuthorizationPolicy

image

### Automation testing

N/A

### Issue reference

Closes #7398


kubernetes/types.go
```diff
@@ -113,6 +113,18 @@ var (
 	}
 	ApiNetworkingVersionV1Alpha3 = NetworkingGroupVersionV1Alpha3.Group + "/" + NetworkingGroupVersionV1Alpha3.Version
 
+	NetworkingGroupVersionV1Beta1 = schema.GroupVersion{
+		Group:   "networking.istio.io",
+		Version: "v1beta1",
+	}
+	ApiNetworkingVersionV1Beta1 = NetworkingGroupVersionV1Beta1.Group + "/" + NetworkingGroupVersionV1Beta1.Version
```
nrfox, line 120: Do the older versions need to stay here?

tests/data/references/multiple-gateways.yaml
```diff
@@ -1,60 +1,60 @@
...
   name: reviews-well
   namespace: bookinfo
 spec:
   hosts:
-    - 'reviews.prod.svc.cluster.local'
+  - 'reviews.prod.svc.cluster.local'
   gateways:
-    - gateway1
-    - bookinfo2/gateway2
-    - mesh
```
nrfox, line 12: Some yaml linter?

nrfox (commented):
One question but otherwise LGTM. You'll also need to update the compat matrix on kiali.io since after this gets merged the next release will no longer be compatible with versions of istio < 1.22.

aljesusg:
Care with this. 
We don't know if we are going to have a 2.6.

Ar this moment 2.6 will ship istio 1.20 so if we merge this will have an old kiali version and not the latest one for 2.6

I'll wait to discuss this

ferhoyos:
Agree with you, @aljesusg. We need to ensure that this change does not break our plans for OSSM.

As you mentioned, theoretically, there should be no problem since OSSM 2.6 will use kiali 1.73, and OSSM 3.0 will have Istio 1.22.

In any case, I will wait to discuss this with the rest of the team before merging this PR.

tests/data/references/multiple-gateways.yaml
ferhoyos: Yes, we added yaml linter some time ago (https://github.com/kiali/kiali/pull/7011) to ensure consistency across Kiali application.

kubernetes/types.go
ferhoyos: It is used in `kubernetes/filters.go`:

```
func FilterSupportedGateways(gateways []*networking_v1.Gateway) []*networking_v1.Gateway {
	filtered := []*networking_v1.Gateway{}
	for _, gw := range gateways {
		if gw.APIVersion == ApiNetworkingVersionV1 || gw.APIVersion == ApiNetworkingVersionV1Beta1 || gw.APIVersion == ApiNetworkingVersionV1Alpha3 {
			filtered = append(filtered, gw)
		}
	}
	return filtered
}
```
Actually I don't know why we are checking the apiVersion of the gateway. Do you think we could directly remove that if statement?
```
for _, gw := range gateways {
	filtered = append(filtered, gw)
}
```

hhovsepy: It seems `FilterSupportedGateways` is not used anymore, so it can be removed. 
Another place where `NetworkingGroupVersionV1Beta1` is used is https://github.com/kiali/kiali/blob/master/business/istio_config.go#L978 but it needs to be edited to `NetworkingGroupVersionV1`.
And you can remove 'NetworkingGroupVersionV1Beta1'.

hhovsepy:
Tested locally and on OCP cluster, works fine.

kubernetes/types.go
ferhoyos: Thanks @hhovsepy, I have removed `NetworkingGroupVersionV1Beta1` with your comments.

kubernetes/types.go
```diff
@@ -131,12 +137,6 @@ var (
...
 
-	NetworkingGroupVersionV1Beta1 = schema.GroupVersion{
-		Group:   "networking.istio.io",
-		Version: "v1beta1",
-	}
-	ApiNetworkingVersionV1Beta1 = NetworkingGroupVersionV1Beta1.Group + "/" + NetworkingGroupVersionV1Beta1.Version
-
 	SecurityGroupVersion = schema.GroupVersion{
 		Group:   "security.istio.io",
 		Version: "v1beta1",
```
hhovsepy, line 142: Shouldn't this be 'v1'?

hhovsepy, line 142:
```suggestion
		Version: "v1",
```

kubernetes/types.go
```diff
@@ -149,11 +149,11 @@ var (
 	}
 	ApiExtensionV1Alpha1 = ExtensionGroupVersionV1Alpha1.Group + "/" + ExtensionGroupVersionV1Alpha1.Version
 
-	TelemetryGroupV1Alpha1 = schema.GroupVersion{
+	TelemetryGroupV1 = schema.GroupVersion{
 		Group:   "telemetry.istio.io",
 		Version: "v1alpha1",
```
hhovsepy, line 154:
```suggestion
		Version: "v1",
```

hhovsepy: same here. to be 'v1'

hhovsepy:
Hi @ferhoyos , have suggested 2 changes, otherwise lgtm.

doc.go
```diff
@@ -796,6 +806,34 @@ type MetricsStatsResponse struct {
...
+type ClustersNamespaceHealthResponse struct {
+	// in: body
+	Body models.ClustersNamespaceHealth
+}
+
+// Response of the mesh query
+// swagger:response meshResponse
+type MeshResponse struct {
+	// in: body
+	Body business.Mesh
```
hhovsepy, line 827: It seems some doc is auto-generated.

hhovsepy (approved):
Tested on OCP cluster.

PR #7406, opened 2024-06-05
Remove ambient tests from jenkins pipeline tests

josunect:
### Describe the change

Remove ambient tests from jenkins pipeline tests, as Ambient requires another environment.

### Steps to test the PR

Run yarn cypress:run:junit, `@ambient `tests should not run.

### Automation testing

N/A

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -40,6 +40,7 @@
     "build:prod": "sh -ac '. ./.env.downstream; npm run build:kiali'",
     "precypress:run:junit": "npm run cypress:delete:reports",
     "cypress": "cypress open -e TAGS=\"not @multi-cluster and not @ambient\"",
+    "cypress:ambient": "cypress open -e TAGS=\"@ambient\"",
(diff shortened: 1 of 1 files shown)


matejnesuta (commented):
I think we should add the same condition for the target on the line 42 and add a separate `cypress:ambient` target as well. Not that it's important for Jenkins, but it might be handy. Otherwise lgtm.

josunect:
> I think we should add the same condition for the target on the line 42 and add a separate `cypress:ambient` target as well. Not that it's important for Jenkins, but it might be handy. Otherwise lgtm.

You mean, this one, right, @matejnesuta ? https://github.com/kiali/kiali/blob/master/frontend/package.json#L42
There is already one for Ambient here: https://github.com/kiali/kiali/blob/master/frontend/package.json#L50

matejnesuta:
> You mean, this one, right,
Yes
>There is already one for Ambient here
Yes, but this one is for the headless mode. I just thought it might be useful to add one for the GUI runner too.

matejnesuta (approved):
LGTM

PR #7411, opened 2024-06-06
Configuration table for mesh target panel nodes

### LINKED ISSUE #7379: [mesh page] display side-panel JSON in a table format
Displaying the configuration as plain JSON doesn't look very good. I suggest presenting it in a table format, similar to what we do in the View Debug Info modal:

image

ferhoyos:
### Describe the change

This PR presents mesh node configuration in a table format instead of plain JSON, adding copy and download buttons.

image

**Bonus**:
- Update `json-beautify` library to add Typescript types.
- Add typescript types for `react-copy-to-clipboard` library.
- Refactor i18n labels that use variables with the interpolation feature.
- Refactor `DebugInformation` to use the shared `JsonTable` component.
- Reorganize some components for better classificiation (all components belonging to one folder):
  - Create a `Table` folder to group `SimpleTable` and `JsonTable`.
  - Create a `Dropdown` folder to group all dropdown components.
  - Create a `Select` folder and move the `SimpleSelect` component there.
  - Move `PfSpinner` to the `Pf` folder.
  - Move `TextOfLink` to the `Link` folder.
- Move the `download` function to `utils` as it is used by at least 3 different components.
- Refactor `IstioConfigPreview` since it contained some `trafficPolicies` hardcodes and it should be a generic component. Modify the copy tooltip to `Copied` when the user clicks to copy button to indicate the user that the text has been successfully copied to the clipboard.
- Convert `GraphSecondaryMasthead` component to React hook
- Fix `HealthIndicator` unit test
- Lint fixes

### Steps to test the PR

1. Go to mesh page
2. Click on any mesh node (e.g., Jaeger, Grafana, Prometheus, etc.)
3. Check that the mesh node configuration is displayed in a table format
4. Check that the `Copy` button copies the configuration json in the clipboard
5. Check that the `Download` button downloads the configuration json file
6. For the bonus, check that the DebugInformation in the toolbar and the Istio config object preview work as expected (only refactor work)

image

image

### Automation testing

N/A

### Issue reference

Closes #7379


hhovsepy (dismissed):
Works fine for me:
Screenshot from 2024-06-10 09-29-35

Also config yaml and debug info are fine.

frontend/src/types/Health.ts
```diff
@@ -19,11 +19,17 @@ interface HealthConfig {
   statusConfig?: HealthItemConfig;
 }
 
+export const enum HealthItemType {
+  TRAFFIC_STATUS = 'traffic_status',
+  POD_STATUS = 'pod_status'
```
jshaughn, line 24: Could be alphabetized.

jshaughn (approved):
The minor code-reorg looks good and I see your usual improvements around class->FC, i18n, typing, px>rem, etc :-)   The new JsonTable component is nice. 
TargetPanelConfigTable seems to be working fine.   Having Copy/Download everywhere is great!  Looks good!

nrfox:
Is it possible to render the configuration as yaml? It could be more readable especially for deeply nested fields and it would also match the format that your istio and kiali configs are in: `kubectl get configmaps -n istio-system istio -o yaml`

ferhoyos:
@nrfox yes, it is quite straightforward to render the configuration as yaml.

Kiali configuration
- JSON:
image

- YAML:
image

Debug information:
- JSON:
image

- YAML:
image

Additional State:
- JSON:
image

- YAML:
image

The YAML format seems to offer better readability in tables than JSON. Regarding additional state, IMO, it is quite similar. However, in the Kubernetes world, YAML is preferred over JSON.

What do you think @jshaughn @hhovsepy?

hhovsepy:
IMO this new table format is more readable, additionally there can be `Download as JSON` and `Download as YAML` both buttons.

nrfox:
Sorry I'm coming in late looking at this so I don't have much context but why have the table at all? Could you just use something like the ace editor to show the yaml unbroken? With the table the top level keys are all separated which imo makes it harder to read. The table is great for labels and annotations which are always going to be `map[string]string`. You can have annotations that are json strings but that's more for one-off configuration and it's still just a string. The kiali and istio configs are always going to be nested data.

Also do we want to filter out empty values? There's a lot of config values whose value is just `{}` or `""` which makes the config more noisy.

ferhoyos:
No problem @nrfox, you are always welcome to the party :wink: 

IMO I would use ace-editor for editable yamls or for complex objects like `Additional State`. Kiali and Istio config are quite simple, mainly `map[string]string` except for some fields, and the table format is a more beautiful way to display this information instead of plain yaml or json:

image

If the top level keys are very separated we can play with the table column widths...

Regarding the empty values... maybe it is good the user knows that this config value is empty. I'm not sure here.

@hhovsepy I would keep it simple and download only in one format (e.g., yaml). The user can always use yaml<->json conversion tool (as said Kiali and Istio configs are quite simple objects).

jshaughn:
@ferhoyos , I like the YAML suggestion, looks better.  @nrfox , I think the table approach was just to match the approach previously taken by our `Debug Info` presentation.  I don't know whether that was UX-influenced or our own doing.  My suggestion is to keep the new table format, I think it looks good in most cases.  But, make the switch to YAML.  For simplicity, I'd leave single `copy` and `download` buttons that deliver the raw YAML.  That lets a user easily capture it and inspect it in their editor of choice, if needed.

jshaughn:
Oh, I'd leave the empty values, I prefer seeing the more complete picture.

ferhoyos:
I have converted tables from JSON to YAML. I had to upgrade `js-yaml` to the latest version to be able to add a `replacer` function in the `Additional State` tab to remove sensitive data like username. Additionally, I have removed the `json-beautify` library as there is no JSON parsing in Kiali now.

hhovsepy (approved):
YAML tables looks good:

Screenshot from 2024-06-12 11-11-08
Screenshot from 2024-06-12 11-10-47

jshaughn (approved):
LG!

PR #7412, opened 2024-06-07
[hack] skupper metrics update

jmazzitelli:
cc @jshaughn


jshaughn (approved):
LG, this reflects data from skupper team.

PR #7414, opened 2024-06-10
Add support for external deployment cluster in mesh cypress test

ferhoyos:
### Describe the change

In some environments, external services (such as tracing, metric store, etc.) can be deployed outside of the mesh (external deployments). This PR adds support for executing mesh cypress tests in such environments.

### Automation testing

Mesh page cypress test


ScriptingShrimp (approved):
LGTM!

PR #7419, opened 2024-06-11
Don't force parallel edges to be straight

### LINKED ISSUE #7418: [graph] PFT graph does not show parallel edges with different protocols
The PFT graph does not show parallel edges.  This is typically when we have traffic with different protocols (both with telemetry) between the same source and destination node.  This is seen commonly in an ambient graph, but can happen
in non-ambient graphs as well.

Currently, PFT graph for ambient bookinfo demo looks like this:
image

It looks like only TCP edges, but actually the two edges are superimposed, note the correct edge count in the side panel, which shows more edges than you can see onscreen.

jshaughn:
Fixes #7418 

In the early days of the PFT graph dev we overrode PFT's bendpoints to force straight edges, like we have in Cytoscape.  But in the case that the same source and dest nodes (a -> b) has parallel edges (typically edges with different protocols), the removal of bendpoints caused the edges to bothe be straight, thus overlaying. This adds a check to allow bendpoints in this situation.

before:
image

after:
image

### DIFF
--- a/frontend/src/pages/GraphPF/elements/extendedBaseEdge.ts
+++ b/frontend/src/pages/GraphPF/elements/extendedBaseEdge.ts
@@ -1,6 +1,8 @@
 import { BaseEdge, Point } from '@patternfly/react-topology';
 
-// This extends BaseEdge to eliminate Bendpoints (mainly because Dagre layout does not provide an option to use or not use bendpoints).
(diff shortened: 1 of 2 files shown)


josunect (approved):
Works as expected: 
image

At the beginning, I didn't understood why there was different end terminal types, then I've seen in the code it depends on the protocol.

PR #7421, opened 2024-06-12
[CI] set ipFamily in kind config

nrfox:


jmazzitelli (approved):
I'm approving this now in case this fixes the issue, you can merge immediately
(update: nm :) still ipv6 failure)

jmazzitelli:
we fixed this in another PR

PR #7423, opened 2024-06-12
[v1.65] Bump go jose

nrfox:
Backport of #7248


nrfox:
@jshaughn I just rebased with #7428 so hopefully tests will run now

PR #7435, opened 2024-06-13
[molecule] make sure to use ipv4 in molecule tests

jmazzitelli:
similar to what we did in PR https://github.com/kiali/kiali/pull/7427 for other CI tests

If you really want to see it work, run:

```
./hack/ci-kind-molecule-tests.sh --client-exe $(which kubectl) --kind-exe $(which kind) --all-tests "config-values-test" --git-clone-protocol https --irc-room "" --upload-logs false --rebuild-cluster true -ci true --operator-installer helm --olm-enabled false
```

(if you run that, you should see it end successfully. After it finishes, clean up via `kind delete cluster --name=ci`


josunect (approved):
LGTM

PR #7433, opened 2024-06-13
Add Ambient label in service and app details

josunect:
### Describe the change
Add Ambient label in service and app details

New labeling: 

image
image

image
image

Included Label in Mesh page: 
image

### Steps to test the PR

- Install istio ambient
- Add namespace to Ambient
- Check Services and App details, should include the Ambient Label

Note: Waypoint labels should be fixed here: https://github.com/kiali/kiali/pull/7351 so, L7 should not be visible for workloads

### Automation testing

N/A

### Issue reference

Fixes #7432

PR #7436, opened 2024-06-13
First pass at supporting ambient/waypoint reporting in metrics queries

### LINKED ISSUE #7429: [ambient] support http rmetric charts
In Ambient the HTTP metrics are generated from the Waypoint, with `reporter="waypoint"`.  We need to make changes to support HTTP metric charts.

jshaughn:
Closes #7429 

- add 'includeAmbient' istio metric query param. When true (and reporter != 'both') modify the query to use a regex match on 'reporter' (e.g. reporter=~"source|waypoint")
  - default is false and so the client has to "opt in".  We could make it true (only if istio is ambient enabled), so that we always include waypoint reporting unless the client opts out. This could be better as it won't miss anything, but also could be slightly less efficient.
- adds this as a required option in the client, but that could change if we were to change the server-side default.
  - currently setting to true client-side when we don't have a simple "isAmbient" flag to use for the component.
- adds some anticipated infrastructire for supplying "IsAmbient" on all graph nodes.  This should be doable given upcoming server-side detection for apps and services.  At that time I think we can fix the TODOs in the client (or if we decide to change the "IncludeAmbient" default from "false" to"true".


frontend/src/components/Metrics/__tests__/IstioMetrics.test.tsx
```diff
@@ -110,6 +110,7 @@ describe('Metrics for a service', () => {
                 object="svc"
                 objectType={MetricsObjectTypes.SERVICE}
                 direction={'inbound'}
+                includeAmbient={false}
```
josunect, line 113: Maybe we can add some tests with includeAmbient={true}?

josunect (approved):
Tested and LG: 

image
image

Code Looks good. 
Just a comment, but it can be merged like it is so it can be included in this release. 

The TODOs can be done in https://github.com/kiali/kiali/pull/7433

Issue #7256, opened 2024-04-04
[CI] Flake - graph_context_menu nodes undefined


nrfox:
### Describe the bug

CI flake involving the following test:
![Screenshot from 2024-04-04 18-59-13](https://github.com/kiali/kiali/assets/6226732/b0ccc631-082e-4e17-bb2b-662ee8916eee)

Screenshot of cypress at time of failure:
![Kiali Graph page - Context menu actions -- Actions in context menu for a service node with existing traffic routing (failed)](https://github.com/kiali/kiali/assets/6226732/e69ec831-c2ad-48a9-8f6d-3486dda17f65)

From run: https://github.com/kiali/kiali/actions/runs/8560404196/job/23459494170#step:8:861

nrfox:
I can't say for sure but I think this is the line that is failing: https://github.com/kiali/kiali/blob/5f59bc2c2f56dc70065b948657892b9cf925154e/frontend/cypress/integration/common/graph_context_menu.ts#L35-L37

jshaughn:
Will monitor for one more sprint.

PR #7442, opened 2024-06-17
apply same flake protection as we did for issue 7256

jshaughn:
In an effort to prevent graph-page-related flakes, add the same wrapping we did to avoid the content-menu flake (#7256), around remaining instances.

Note that this does not apply the same logic to the mini-graph tests, because that code is a bit different.  If we suspect a similar flake in that area, we'll have to do some additional work to apply this kind of logic.

@nrfox Should be a straightforward review.  I ran the tests a few times already, feel free to try again if you like.

### DIFF
--- a/frontend/cypress/integration/common/graph.ts
+++ b/frontend/cypress/integration/common/graph.ts
@@ -14,33 +14,41 @@ Then('user sees a minigraph', () => {
 
 Then('user sees the {string} namespace deployed across the east and west clusters', (namespace: string) => {
   cy.waitForReact();
-  cy.getReact('CytoscapeGraph')
+  cy.getReact('GraphPageComponent', { state: { graphData: { isLoading: false } } })
     .should('have.length', '1')
-    .getCurrentState()
-    .then(state => {
-      const namespaceBoxes = state.cy
-        .nodes()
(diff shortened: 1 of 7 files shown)


jshaughn:
@nrfox nudge

frontend/cypress/integration/common/graph.ts
```diff
@@ -14,61 +14,73 @@ Then('user sees a minigraph', () => {
 
 Then('user sees the {string} namespace deployed across the east and west clusters', (namespace: string) => {
   cy.waitForReact();
-  cy.getReact('CytoscapeGraph')
+  cy.getReact('GraphPageComponent', { state: { graphData: { isLoading: false } } })
```
nrfox, line 17: Is there a way to make a custom command or something for this to make it easier to call? Like `cy.ensureGraphLoaded` or something?

nrfox (approved):
One comment but non-blocking. Thanks for copying this to the other tests.

frontend/cypress/integration/common/graph.ts
jshaughn: Probably.  We can look at that when the tests migrate from Cytoscape to PFT, sometime in the next couple of months.

Issue #7444, opened 2024-06-18
[pft] improve "focus node'


jshaughn:
Currently for the PFT graph the focus node is only "selected".  This is not easy to see when visiting a larger graph.  We need something more visually striking.  In Cytoscape we have the "zooming circles", we don't need to do the same thing, just something that lets the user locate the focus node.

PR #7446, opened 2024-06-18
[pft] Improve focus node handling

jshaughn:
To help the user locate the focus node add an animation, in addition to selecting the node.
- change things so that the focusNode is passed into the graph (so that the graph code does not itself process URL params)
- remove focusNode code from MeshPage, can bring it back in its correct form if/when needed.
- also, remove some commented code

closes https://github.com/kiali/kiali/issues/7444

To test:
- node graph a regular node and return to main graph
- repeat for box node

### DIFF
--- a/frontend/src/pages/Graph/GraphPage.tsx
+++ b/frontend/src/pages/Graph/GraphPage.tsx
@@ -421,7 +421,7 @@ class GraphPageComponent extends React.Component<GraphPageProps, GraphPageState>
       this.loadGraphDataFromBackend();
     }
 
-    if (!!this.focusSelector) {
+    if (this.focusSelector) {
       this.focusSelector = undefined;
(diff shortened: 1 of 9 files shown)


leandroberetta (commented):
Hi @jshaughn, first round, looks good except for:

image

It seems some dark mode styles are being applied?.

jshaughn:
> It seems some dark mode styles are being applied?.

Yeah, that doesn't look too great.  It doesn't leave it that way, just while it was flashing, right?  Anyway, I think I was able to make it better.  Let me know!

leandroberetta (approved):
The issue is fixed, thanks @jshaughn. 

I'm approving but just in case, there is a minor difference between the node focus and the group focus, the group focus seems to be in a higher z position (overlaps the group) while the node focus does not.

Issue #7448, opened 2024-06-19
[Ambient] Cannot load the graph: Namespace is excluded


josunect:
### Describe the bug

When the graph has any nodes from a namespace excluded by Kiali, the Graph cannot be visualized:

![image](https://github.com/kiali/kiali/assets/49480155/aff552c7-f736-4d8d-ad58-68a947f7887a)

This is because, the handleWaypoint from the Ambient appended gets the workloads list from each namespace, to check whether a node is a Waypoint proxy. Ref. https://github.com/kiali/kiali/blob/master/graph/telemetry/istio/appender/ambient.go#L55

This method `getWorkload` is defined in the appender, https://github.com/kiali/kiali/blob/master/graph/telemetry/istio/appender/appender.go#L380, so probably this could happen in other situations as well?

I think the main point is that just the control of allowed namespaces is done when one of the appenders is using the 

### Expected Behavior

- Maybe exclude check if the namespace is excluded by Kiali? (In that case, we won't know if is a waypoint proxy)

### What are the steps to reproduce this bug?

1. Add istio-system to the ambient mesh
2. See the graph

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:** 1.86
- **Istio version:** 1.22
- **Kubernetes impl:** 
- **Kubernetes version:** 
- **Other notable environmental factors:**

jshaughn:
The relevant PRs need to be backported to v1.86 for a patch release, to avoid possible graph crash for ambient users.

PR #7449, opened 2024-06-19
Fix graph issue with excluded namespaces

josunect:
### Describe the change

Doesn't fail if node is not accessible for graph

### Steps to test the PR

- Install istio 1.22 with Ambient profile: `istio/install-istio-via-istioctl.sh -c kubectl -cp ambient`
- Install Kiali
- Install bookinfo and include in Ambient: `istio/install-bookinfo-demo.sh -c kubectl -ai false -tg -w true`
- Add some workload in one namespace excluded for Kiali with traffic for a selected namespace. An easy way is to label the `istio-system` namespace with the Ambient label (`kubectl label ns istio-sytem istio.io/dataplane-mode=ambient`)
- Go to the Traffic graph and select bookinfo an istio-system namespaces. The graph should be seen, no errors in the log. In master, there should be a stack trace error, and sometimes, the "Cannot generate graph" message (It doesn't appear always)

### Automation testing
Added unit test.

### Issue reference
Fixes https://github.com/kiali/kiali/issues/7448

### DIFF
--- a/frontend/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/frontend/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -25,8 +25,8 @@ import { getServiceDetailsUpdateLabel, hasServiceDetailsTrafficRouting } from '.
 
 type ReduxProps = {
   duration: DurationInSeconds;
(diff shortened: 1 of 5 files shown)


hhovsepy (approved):
Verified:
Screenshot from 2024-06-20 14-28-35

PR #7450, opened 2024-06-19
Skipp KIA0701 check for waypoint proxy generated service

hhovsepy:
### Describe the change

Skips 'KIA0701 Deployment exposing same port as Service not found' for generated 'waypoint' service.

image

### Automation testing

Unit testing.


hhovsepy:
> 
> I was thinking that probably it would be useful to have de `isWaypoint` function in other place to be reused by other checkers (If required) but I don't see there is any other common place for this?

I also did not find any suitable place for `isWaypoint` function, it should be used inside 'PortMappingChecker', and as we do not have there `models.Service` to put into, then decided to have it in 'PortMappingChecker'.

josunect (approved):
LGTM

josunect (approved):
Thanks for the refactor, @hhovsepy ! LG

Issue #7447, opened 2024-06-19
Routing Wizard - Empty Matching fail


hhovsepy:
### Describe the bug

Happens on all Service Routing wizard types.
![Screenshot from 2024-06-19 13-29-22](https://github.com/kiali/kiali/assets/604313/273e790c-9ed6-473e-b641-f016991c921c)

### Expected Behavior
Disable `Add Match` button when values are empty.

### What are the steps to reproduce this bug?
Go to Service details.
Actions => Create K8s HTTP Route.
Add empty Match.
Add empty Route. 
Preview and Save.
Error creating objects.

### Environment
Kiali master

PR #7462, opened 2024-06-21
Avoid empty matches adding

hhovsepy:
### Describe the change

For Routing Wizards:

- Request Routing
- K8s HTTP Routing
- K8s GRPC Routing 

Made impossible to create Rule on empty "Header Name/Value" combination by default.

### Steps to test the PR

Go To above mentioned routing wizards.
It should be impossible to add "Add Rule" on empty Header.

### Automation testing

n/a

### Issue reference

There is a bigger task of refactoring the wizards, so this solution is temporary for fixing issue:  https://github.com/kiali/kiali/issues/7447

### DIFF
--- a/frontend/src/components/IstioWizards/K8sRequestRouting.tsx
+++ b/frontend/src/components/IstioWizards/K8sRequestRouting.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import { K8sRules, MOVE_TYPE, K8sRule } from './K8sRequestRouting/K8sRules';
 import { K8sRuleBuilder } from './K8sRequestRouting/K8sRuleBuilder';
 import { K8sRouteBackendRef } from './K8sTrafficShifting';
-import { EXACT, PATH, METHOD, GET, HEADERS, QUERY_PARAMS, GRPC } from './K8sRequestRouting/K8sMatchBuilder';
+import { EXACT, PATH, METHOD, GET, HEADERS, QUERY_PARAMS } from './K8sRequestRouting/K8sMatchBuilder';
 import { getDefaultBackendRefs, getDefaultService } from './WizardActions';
 import { ServiceOverview } from '../../types/ServiceList';
 import { REMOVE, REQ_MOD, RESP_MOD, SET, HTTP, SC301, REQ_RED, REQ_MIR } from './K8sRequestRouting/K8sFilterBuilder';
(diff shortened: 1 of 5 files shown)


ferhoyos (changes requested):
@hhovsepy the `Add Match` button is disabled when there is a header validation error message in the wizard:

image

But the validation is not working when you fill and then empty the header value textbox:

image

Additional comments:
- I think you should disable the `Add Match` button even for empty values. A button that can be clicked but does nothing is very confusing.

- I don't agree with removing the `is present` operator for matching headers. Maybe move it down in the operator dropdown and make the `exact` operator the default option. This use case appears in the istio docs (https://istio.io/latest/docs/reference/config/networking/virtual-service/#HTTPMatchRequest). What we need to do is change the anything regex value to an empty value:

```
 - match:
   - headers:
       myheader: {}
```

hhovsepy:
Hi @ferhoyos , the idea behind making "Add Match" button doing nothing was not to overcomplicate the code not to make confusion in wizard when just opened it and all is disabled and errors shown. 
Will try to find a better solution with 'is present'.
Anyway there is a task to redesign the whole wizard.

ferhoyos:
@hhovsepy we can discuss with @fkargbo what is the best approach in this case. If it is not very complicated, I would show the validation error message if the user clicks on the `Add Match` button with empty values. Or disable the button if the other approach is very complicated until we implement the wizard redesign. My limited UX knowledge says that a button that does nothing when it is clicked is not a good UX approach. The user always expect something after clicking a button :slightly_smiling_face:

hhovsepy:
Thanks for the feedback @ferhoyos , returned back the 'presence', also error is shown when clicked on "Add Match" while values are empty.

hhovsepy:
When first opened the wizard:
Screenshot from 2024-06-26 13-38-36


When click on "Add Match" while values are  empty:
Screenshot from 2024-06-26 13-37-43

hhovsepy:
The validation mechanism is kept, the thing here is, that both buttons "Add Match" and "Add Route Rule" have a single validation, they both are enabled or disabled. And "Add Route Rule" should be available to add "Any Match" when filters and matches are not added.

josunect:
I think this is a better solution as you cannot add invalid values. 

I would probably expect that "Add match" has always the same behavior (If there are no values, you can click in Add Match, but after you click, the button is disabled for the same data), but, at least it adds this error value new to the "Add Router Rule". 

image

image


image

ferhoyos (approved):
Thanks for the changes @hhovsepy. Now the wizard avoids adding empty match values. 

A non-blocking comment: I think it is more user-friendly to say `header [xxx] is present` than `header[xxx] presents {}`:

image

hhovsepy:
> Thanks for the changes @hhovsepy. Now the wizard avoids adding empty match values.
> 
> A non-blocking comment: I think it is more user-friendly to say `header [xxx] is present` than `header[xxx] presents {}`:
> 
> image

Thanks for the reviews @josunect and @ferhoyos.
Initially it was 'is present', but this breaks the parsing function of string, so made it as one word without spaces in between.
And having the notifications next to the "Add Route Match" button is by design of wizard.

PR #7464, opened 2024-06-24
Fix the apiVersion of ReferenceGrant in the Kiali wizard

ferhoyos:
### Describe the change

This PR fixes the `apiVersion` value of the Gateway API Reference Grant object in the Kiali wizard.

### Steps to test the PR

1. Navigate to the Istio Config page
2. Click on Actions -> K8sReferenceGrant
3. Fulfill the wizard form and click on the `Preview` button
4. Check that the `apiVersion` value is the correct one `gateway.networking.k8s.io/v1beta1`

image

### Automation testing

N/A

### Issue reference

Fixes #7463

PR #7466, opened 2024-06-24
[v1.86] Promote Istio API to V1 + Fix the apiVersion of ReferenceGrant

ferhoyos:
### Describe the change

PR backports of #7405 and #7464 for version v1.86. They are grouped together because the second PR fixes a small issue introduced by the first one.


hhovsepy (approved):
Tested locally and on OCP cluster.

PR #7467, opened 2024-06-25
A follow-up for https://github.com/kiali/kiali/issues/7448

jshaughn:
Related to #7448 

A follow-up for https://github.com/kiali/kiali/issues/7448.  The original PR may be sufficient but I think could be improved by using some of the existing conventions and/or flags available.
- In Ambient appender, use AccessibleNamespaces to check whether it is safe to call the back-end.
- For the cytoscape context menu, use the `isInaccessible` flag to block the user (and the code) from requesting unavailable information.
- Also block context menu options in PFT graph for inaccessible nodes.

### DIFF
--- a/frontend/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
+++ b/frontend/src/components/CytoscapeGraph/ContextMenu/NodeContextMenu.tsx
@@ -103,13 +103,13 @@ function getLinkParamsForNode(node: DecoratedGraphNodeData): LinkParams | undefi
       break;
   }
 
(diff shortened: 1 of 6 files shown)


josunect (commented):
It works as expected: 
image

No issues with non accessible namespaces (see `kube-system`): 

image

I think there should be something left to change. When you click on an unknown node, you can see the summary panel node and the kebab menu. When you click on that menu, it throws an error: 

image

Because it is trying to perform a request with unknown data. 

image

Maybe in the contextual menu can be also checked if the node is accessible?

PR #7482, opened 2024-06-27
Upgrade React Router to v6

### LINKED ISSUE #7207: React Router migration from v5 to v6
Currently Kiali uses `react-router` library version 5, which works fine but has some issues with the routing algorithm, complex API, big bundle size, etc.. New version 6 is adapted to React hooks and provide new features like relative path links, smaller bundle size, nested routes, etc.

Besides, OpenShift Console has recently migrated from version 5 to 6 and it will remove version 5 support in the medium term (maybe version 4.17). Since OSSMC shares `react-router` library with OpenShift console, when it removes react-router v5 support it won't be possible to update OSSMC code until Kiali completes the migration.

ferhoyos:
### Describe the change

Upgrade React Router to v6:
- `history` is not supported in React Router v6 and has been replaced by the `router` object.
- Creation of a `location` object to retrieve pathname and search values from the `router` object.
- Adapted path routes.
- Some main and navigation simple classes converted to hooks: App, BreadcrumbView, Menu, Navigation, RenderPage.
- Error boundary adapted to React Router v6.

Small refactors:
- Removed unused components `PfTitle`, `ErrorBoundaryWithMessage` and `SwitchErrorBoundary`.
- Moved `ServiceId` interface to the `ServiceInfo.ts` file.
- Lint fixes and some translations.

Although the usual approach for routing in React Router v6 is using hooks (`useNavigate` instead of `router.navigate`), I have followed the `router` approach because the hooks method would involve a significant effort and risk for the Kiali application:
- Converting around 30 big and complex classes (including graphs) into hooks.
- Refactoring the entire `HistoryManager` class and how urls are handled in Kiali.

I have also temporarily disabled the unsaved changes prompt on the Istio config details page because the current `Prompt` component is not supported in React Router v6. I will create a separate PR to enable it again (this change is not straightforward and this PR is big and complex enough).

### Steps to test the PR

Test that Kiali works as expected, paying special attention to routing.

### Automation testing

Unit and cypress tests adapted to React Router v6

### Issue reference

Closes #7207


hhovsepy:
Tested locally and on OCP cluster with running cypress tests, works fine, just the failing test happens again. But graph context menu items works fine.

ferhoyos:
I have fixed the issue that was causing the Cypress tests to fail. The PR is now ready for review :slightly_smiling_face:

hhovsepy (approved):
Works fine now, tested on OCP cluster.

josunect (approved):
LGTM

I did some tests in the application, using the history, creating objects, with anonymous/token authentication, external links, it looks like everything works as expected.

frontend/src/index.tsx
```diff
@@ -39,4 +42,16 @@ Date.prototype.toLocaleStringWithConditionalDate = function () {
...
+  {
+    element: <App />,
+    children: pathRoutes
+  }
+]);
+
+// redirect to the router basename if the pathname does not include it
+if (!window.location.pathname.includes(rootBasename)) {
+  router.navigate('/', { replace: true });
+}
```
jshaughn, line 55: voodoo.

frontend/src/pages/AppList/AppListPage.tsx
```diff
@@ -74,6 +74,11 @@ class AppListPageComponent extends FilterComponent.Component<AppListPageProps, A
     this.promises.cancelAll();
   }
 
+  onSort = (): void => {
+    // force list update on sorting
+    this.setState({});
+  };
```
jshaughn, line 80: I guess this is this a bug fix?

frontend/src/components/CytoscapeGraph/ContextMenu/EdgeContextMenu.tsx
```diff
@@ -12,14 +12,12 @@ const contextMenu = kialiStyle({
...
-    return (
-      <div className={contextMenu}>
-        {getTitle(`Edge (${prettyProtocol(this.props.protocol)})`)}
-        {renderBadgedName(decoratedNodeData((this.props.element as EdgeSingular).source()), 'From:  ')}
-        {renderBadgedName(decoratedNodeData((this.props.element as EdgeSingular).target()), 'To:        ')}
-      </div>
-    );
-  }
-}
+export const EdgeContextMenu: React.FC<EdgeContextMenuProps> = (props: EdgeContextMenuProps) => {
```
jshaughn, line 15: I'm fairly sure Edge context menus never worked or aren't used.  But this if fine.

jshaughn (approved):
Code Looks good, @ferhoyos!  I couldn't break anything, and I tried hard :-)  How long will this v5-compat approach hold us?

frontend/src/index.tsx
ferhoyos: Here you define the router application:
- the parent element will be the `App` component (the one that will always be loaded). 
- the children paths are the components loaded depending on the URL (within the `<Outlet />` component used in `RenderPage.tsx`).

Lines 52-55 add a feature provided by React Router v5, but not by v6. The scenario is that if you add a basename for your router (in our case `/console`), this is not added automatically when you navigate to the root url (e.g. http://localhost:20001/kiali). Therefore, we have to force the navigation to `/console` if this basename is not included in the url pathname.

More info here: https://www.github.com/remix-run/react-router/issues/8427

frontend/src/pages/AppList/AppListPage.tsx
ferhoyos: Yes. For list sorting, we modified the URL parameter in the pathname. React Router v5 reloads the component if the pathname is modified, but React Router v6 does not. Therefore, I have to force the reload of the list component with a setState to see the sorting.

ferhoyos:
> How long will this v5-compat approach hold us?

We have to maintain v5-compat library until OpenShift decides to remove it due to OSSMC plugin.

Issue #7232, opened 2024-03-26
Hardcoded ingressgateway labels in code


hhovsepy:
In Kiali code there are hardcoded "istio-ingressgateway" workload labels:
https://github.com/kiali/kiali/blob/master/frontend/src/pages/IstioConfigNew/GatewayForm.tsx#L29
https://github.com/kiali/kiali/blob/master/models/workload.go#L437

They should be configurable with default values.

PR #7492, opened 2024-06-28
Made istio=ingressgateway and istio=egressgateway selector labels configurable

hhovsepy:
### Describe the change

There were hardcoded `istio=ingressgateway` and `istio=egressgateway` label selectors in the code, moved them into "istio_labels" configuration with default values.

### Steps to test the PR

Make sure that config values `ingress_gateway_label` and `egress_gateway_label` are read correctly.
Regression testing of workloads lists and details pages for gateway workloads.
Screenshot from 2024-06-28 16-52-07

Istio config wizard -> creation of Gateway:
Screenshot from 2024-06-28 16-32-27

### Automation testing

n/a

### Issue reference

https://github.com/kiali/kiali/issues/7232

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -345,7 +345,9 @@ type IstioLabels struct {
 	AmbientWaypointLabelValue  string `yaml:"ambient_waypoint_label_value,omitempty" json:"ambientWaypointLabelValue"`
 	AmbientWaypointUseLabel    string `yaml:"ambient_waypoint_use_label,omitempty" json:"ambientWaypointUseLabel"`
 	AppLabelName               string `yaml:"app_label_name,omitempty" json:"appLabelName"`
-	InjectionLabelName         string `yaml:"injection_label,omitempty" json:"injectionLabelName"`
+	EgressGatewayLabel         string `yaml:"egress_gateway_label,omitempty" json:"egressGatewayLabel"`
+	IngressGatewayLabel        string `yaml:"ingress_gateway_label,omitempty" json:"ingressGatewayLabel"`
+	InjectionLabelName         string `yaml:"injection_label_name,omitempty" json:"injectionLabelName"`
 	InjectionLabelRev          string `yaml:"injection_label_rev,omitempty" json:"injectionLabelRev"`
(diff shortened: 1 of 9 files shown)


config/config.go
```diff
@@ -1000,7 +1012,7 @@ func (conf Config) String() (str string) {
 	obf := conf.Obfuscate()
 	str, err := Marshal(&obf)
 	if err != nil {
-		str = fmt.Sprintf("Failed to marshal config to string. err=%v", err)
+		str = fmt.Sprintf("Failed to marsx`hal config to string. err=%v", err)
```
josunect, line 1015: Is that correct?

frontend/src/helpers/LabelFilterHelper.ts
```diff
@@ -83,24 +84,38 @@ const andLabelOperation = (labels: { [key: string]: string }, filters: string[])
...
+const getKeyAndValues = (filters: string[]): { keyValues: string[]; keys: string[] } => {
   // keys => List of filters with only Label Presence
   // keyValues => List of filters with Label and value
   const keys = filters.filter(f => !f.includes('='));
   const keyValues = filters.filter(f => f.includes('='));
-  return { keys, keyValues };
+  return { keyValues, keys };
+};
+
+export const getGatewayLabels = (labelConfig: string): string[] => {
```
josunect, line 103: Maybe here it can throw some warn/error if the format of the label is not right? If an incorrect value is specified and Kiali is not picking the value because is incorrect, the use is not going to have any warning about it. 

E.g.
image

image

But, then, the label is removed: 
image

Maybe here or while validating the config in https://github.com/kiali/kiali/blob/master/config/config.go#L1187

josunect (commented):
The config is read correctly.

frontend/src/helpers/LabelFilterHelper.ts
hhovsepy: good catch @josunect have added it

hhovsepy:
failed test has separate issue https://github.com/kiali/kiali/issues/7522

josunect (approved):
LGTM

PR #7496, opened 2024-07-01
Autodetect sidecar injector configmap

### LINKED ISSUE #7177: be able to auto-detect more istio things for ease-of-configuration

Right now there are lots of individual configuration settings in the Kiali CR that need to be set if Istio is not installed in the default way (and by default I mean the istioctl installation method with addons installed via the addon yaml that ships with istio).

It is hard for users to know all the touch points to change if they have a non-default install such as Istio is in namespace other than istio-system or that they are using canary upgrades where the istio config map has a version number in the name.

The user needs to read the Kiali CR reference doc and figure out what settings need to be explicitly set so their non-default install works with Kiali. This is tedious, confusing, and error prone and often results in long discussion threads in Slack to help the person get Kiali working.

### What is the new behavior?

At minimum, we need a doc page in kiali.io that gives the details of all the things that the user might need to change. Title the doc page something like, "How to configure Kiali to work with non-default Istio installations".

But I would like to see if we could add some auto-detection features to the operator where the operator can do more to figure out what the settings should be. The operator could, for example, scan the spec.istio_namespace and look for config maps that appear to be the Isto ConfigMap (it would do this only if the config_map_name setting does not refer to an actual configmap). There could be other auto-detection mechanisms we could use (for example, look for an istiod pod in a namespace if there is no istiod pod with the default name or in the default namespace).

- [ ] https://github.com/kiali/kiali/issues/7514
- [x] https://github.com/kiali/kiali/issues/7515

nrfox:
### Describe the change

Autodetects the istio sidecar injector configmap. Moves `network` from the cluster object to the controlplane since this is where the info comes from.

### Steps to test the PR

1. Deploy kiali and istio
2. Set `external_services.istio.istio_sidecar_injector_config_map_name = ""`
3. Ensure kiali still works and there's no errors logs in the kiali server about the sidecar injector configmap
4. Ensure you can see the `network` in the mesh target panel for the istiod:
Screenshot from 2024-07-01 13-50-34

### Automation testing

Included unit tests

### Issue reference

Fixes #7177


jmazzitelli:
I didn't change anything in the CM yet, but I see an empty Network:
image

Replication procedure:

1. Start minikube: `hack/k8s-minikube.sh start`
2. Deploy Istio: `hack/istio/install-istio-via-istioctl.sh -c kubectl`
3. Build this PR and deploy operator and kiali: `make -e CLUSTER_TYPE=minikube build build-ui test cluster-push operator-create kiali-create`
4. Point browser to the Mesh page and click the Istiod icon
5. Look at Network row in the right-side table.

Here's what my setting looks like:
```
$ kubectl get cm kiali -n istio-system -oyaml |grep sidecar_injector
        istio_sidecar_injector_config_map_name: istio-sidecar-injector
```
and that exists:
```
$ kubectl get cm istio-sidecar-injector -n istio-system
NAME                     DATA   AGE
istio-sidecar-injector   2      9m47s
```

jmazzitelli:
OK, the issue was my sidecar configmap had network set to `""`. I set it explicitly to network-default and I see it in the mesh table. So change the test procedure step 2:

2. Deploy Istio: `hack/istio/install-istio-via-istioctl.sh -c kubectl -net network-default`

The sidecar configmap will now have that network setting:

```
$ kubectl get cm istio-sidecar-injector -n istio-system -oyaml | grep \"network\"
        "network": "network-default",
```

Which is what I see now, and matches the PR description screenshot:

image

jmazzitelli (approved):

PR #7498, opened 2024-07-02
Stringify the query data only if it is not already stringified

ferhoyos:
### Describe the change

Currently, Kiali can send body data in requests as JSON or string since its `Content-Type` header has the value `application/json`. However, OSSMC only supports string data because its `Content-Type` header is `application/x-www-form-urlencoded` to avoid CORS issues. 

Previously, all request data for OSSMC was stringified, but the problem is that some request data is already stringified in the business logic, which causes failures (e.g., updateNamespace). Now the application only stringifies data that is not already stringified.

### Steps to test the PR

This PR fixes an OSSMC issue. From Kiali perspective, just check that everything works as expected.

### Automation testing

N/A

### Issue reference

Fixes https://github.com/kiali/openshift-servicemesh-plugin/issues/330

### DIFF
--- a/frontend/src/services/Api.ts
+++ b/frontend/src/services/Api.ts
@@ -128,7 +128,7 @@ const newRequest = <P>(
   return axios.request<P>({
     method: method,
     url: apiProxy ? `${apiProxy}/${url}` : url,
-    data: apiProxy ? JSON.stringify(data) : data,
+    data: data,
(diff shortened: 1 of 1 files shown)


josunect (approved):
Doing some test with different methods, in crc, also with Backstage, and looks good.

PR #7502, opened 2024-07-02
Update filter match for ztunnel logs

### LINKED ISSUE #7500: [Ambient] ztunnel logs are using pod name (And not workload)
### Describe the bug

To filter the ztunnel logs we use " " for the pod name, but it is not always like this (Or it was not in the previous version). The ztunnel logs cannot be visualized.

### Expected Behavior
Visualize ztunnel logs for a workload log feature.

1. Install Istio with Ambient profile
2. Install Kiali
3. Install bookinfo and include into Ambient
4. See the workload logs (No logs are found).

### Environment

- **Kiali version:** master
- **Istio version:** 1.22.2
- **Kubernetes impl:** 
- **Kubernetes version:** 

josunect:
### Describe the change

Use a regex to match for the pod name using quotes or without quotes. This has changed a couple of times, not sure if related to the data or to the version.

### Steps to test the PR

- Install istio Ambient: `istio/install-istio-via-istioctl.sh -c kubectl -cp ambient`
- Install Kiali
- Install bookinfo and include it in Ambient Mesh: `istio/install-bookinfo-demo.sh -c kubectl -ai false -tg` 

The performance is compared and if very similar (Using 

### Automation testing

Added integration tests including logs with " and without.

### Issue reference

Fixes #7500


hhovsepy (approved):
Verified,
before:
Screenshot from 2024-07-03 11-56-09


after:
Screenshot from 2024-07-03 11-57-14

Issue #7504, opened 2024-07-02
kiali may hang when asking for masthead's Debug Info while graph page is displayed


jshaughn:
This only happens with the cytoscape graph and I suspect it has to do with recent changes when changing to table format/yaml. 
I won't be surprised if it is due to trying to stringify something from cytoscape, which typically will not stringify due to circular paths.

jshaughn:
The problem is as suspected, introduced with the move to yaml formatting. The js-yaml `replacer` approach is flawed because it stringifies the values BEFORE it gives the replacer func a chance to ignore the key.  We'll need to change this a bit...  I'll assign it to myself for now...

jshaughn:
I think this should probably be backported to v1.86 for both an upstream fix to the istio add-on, and an ossm z-stream release.

PR #7505, opened 2024-07-02
Fix issue with yaml debug-info generation

jshaughn:
Fix issue with yaml debug-info generation when we try to dump cytoscape graph info.  Make sure to protect against cy's duplicate refs, and also stringification of any cy structures. 

Fixes https://github.com/kiali/kiali/issues/7504

See issue for more detail about the problem.

To test:
Before the PR visiting a Cy graph and then asking for debug info would hang the browser tab.  After the fix it should work as expected.

### DIFF
--- a/frontend/src/components/DebugInformation/DebugInformation.tsx
+++ b/frontend/src/components/DebugInformation/DebugInformation.tsx
@@ -157,7 +157,7 @@ const DebugInformationComponent: React.FC<DebugInformationProps> = (props: Debug
     setCopyStatus(CopyStatus.NOT_COPIED);
   };
 
-  const parseConfig = (key: string, value: any): any | null => {
+  const parseConfig = (key: string, value: unkown): unkown | null => {
     // We have to patch some runtime properties  we don't want to serialize
     if (propsToPatch.includes(key)) {
       return null;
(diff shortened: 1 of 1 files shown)


ferhoyos (dismissed):
Good catch @jshaughn! A tricky issue with a simple solution, the perfect fix PR :slightly_smiling_face: 

I have tested, and the Debug Info does not hang after the PR change.

jshaughn:
@ferhoyos I applied your suggestion, thanks!  Feel free to merge this assuming CI passes.

frontend/src/components/DebugInformation/DebugInformation.tsx
```diff
@@ -157,7 +157,7 @@ const DebugInformationComponent: React.FC<DebugInformationProps> = (props: Debug
     setCopyStatus(CopyStatus.NOT_COPIED);
   };
 
-  const parseConfig = (key: string, value: string): string | null => {
+  const parseConfig = (key: string, value: unkown): unkown | null => {
```
ferhoyos, line 160:
```suggestion
  const parseConfig = (key: string, value: unknown): unknown | null => {
```
There was a typo in my first suggestion

ferhoyos (approved):

Issue #7332, opened 2024-05-03
The tracing service should not try to connect indefinitely 


josunect:
### What do you want to improve?

The tracing service should not try to connect indefinitely 

### What is the current behavior?

The tracing service is trying to connect indefinitely 

### What is the new behavior?

The tracing service should try to connect a limited number of times after some period of time and after some retries, it should stop trying to connect. 
The log messages should be info messages until it finishes the retries, then there should be one error message if it is not able to connect.

jshaughn:
I'm reclassifying this as a bug.  It's come up a couple of times and I want to increase the priority.

josunect:
I was reviewing the behavior of the tracing service, and I think it could be difficult to distinguish when the error should be silent or not (The error could be expected when the tracing service is enabled and it has an outage for some reason).

Kiali creates the tracing service but it does not check it's availability. It does it, every time that one of the pages is refreshed to show the status for the enabled services: 

![image](https://github.com/kiali/kiali/assets/49480155/749b3b9a-f165-4467-9f38-6906d58ec168)

Every time the tracing service is not available, it shows an error in the Kiali logs: 

`2024-07-03T09:26:59+01:00 ERR Error fetching availability of the tracing service: Get "http://localhost:3200/api/services": dial tcp 127.0.0.1:3200: connect: connection refused`

Which is expected is the tracing service is really enabled but is not available. 

An alternative could be that the tracing client does some retries during the startup and set an internal variable like auto_disabled, the behavior of Kiali should be then the same as if the client was disabled by the user (Not checking the status of tracing, hiding the trace tab). Maybe it could be doing some query to get some traces (Like 3 queries every minute). If the 3 of them doesn't return a 200, the set the auto_disabled setting to true.

jmazzitelli:
If `external_services.tracing.enabled` is `false`, I assume Kiali doesn't do anything with the tracing client. At least, it shouldn't. Can we change the code such that if that setting is false, Kiali will never check for tracing availability and that component status should never look for tracing status?

josunect:
> If `external_services.tracing.enabled` is `false`, I assume Kiali doesn't do anything with the tracing client. At least, it shouldn't. Can we change the code such that if that setting is false, Kiali will never check for tracing availability and that component status should never look for tracing status?

Yes, this is how is actually working: if tracing is disabled, you won't see any error messages, and Kiali won't look at the tracing status:

[...]

PR #7508, opened 2024-07-03
Tracing is disabled by default

josunect:
### Describe the change

Tracing is disabled by default now

### Steps to test the PR

Run Kiali without any additional config (And without the operator).
Tracing should be disabled. 

### Automation testing
Update setup for tests

### Issue reference
#7332

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -741,7 +741,7 @@ func NewConfig() (c *Config) {
 					Type: AuthTypeNone,
 				},
 				CustomHeaders:        map[string]string{},
-				Enabled:              true,
+				Enabled:              false,
 				GrpcPort:             9095,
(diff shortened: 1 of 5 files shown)


hhovsepy:
QE Jenkins pipeline installs with default traces disabled. It needs to be adjusted to consider default param.

ScriptingShrimp:
do we need to enable the traces in the Jenkins job? Its related only to integration tests? Or more things could go wrong?

josunect:
> Do we need to enable the traces in the Jenkins job? Its related only to integration tests? Or more things could go wrong?

It will affect all the jobs that are installing Kiali and traces are needed...

hhovsepy (approved):
works fine now on Jenkins pipeline, kudos to @mkralik3 for the fix

PR #7509, opened 2024-07-03
[v1.86] Stringify the query data only if it is not already stringified (#7498)

ferhoyos:
Backport of PR #7498 to version 1.86

Issue: https://github.com/kiali/openshift-servicemesh-plugin/issues/330


josunect (approved):
LGTM

Issue #7522, opened 2024-07-04
[cypress] KIA1102 validation fails - Issue in Kiali


hhovsepy:
### Describe the bug

Cypress test "KIA1102 validation" fails due to issue in Kiali side.
For VS "foo" in "sleep" namespace, the configuration in Istio config list is showing "Danger" instead of "Warning" sometimes after refreshing the page.
VS Details page shown correct "Warning" message.

### Expected Behavior
Warning message constantly on Istio Config list page.

### What are the steps to reproduce this bug?

1. Run "KIA1102 validation" cypress test, it will fail.
2. Keep the env.
3. Go to Istio Config page and try to refresh the page several times, the validation for VS "foo" in "sleep" namespaces created by test will change it's status randomly. BUG is here.
4. Go to that VS details page. The "warning" message is there constantly.

### Environment
Happens on Kiali master.

PR #7530, opened 2024-07-09
Fixed validation message position, fixed failing KIA1102 validation.

hhovsepy:
### Describe the change

Fixed failing cypress test 'KIA1102 validation'
Also fixed the validation messages position in YAML editor. Which was introduced after ' - manager:' fields are added in YAML editor inside collapsed "managedFields:".
Before:
Screenshot from 2024-07-09 11-04-32

After:
Screenshot from 2024-07-09 11-06-03

Added missing "cluster" param in ValidationKey and Validations.

### Steps to test the PR
Make sure you have a VS or any other objects with validation messages or warnings.
Check that all messages in YAML editor are shown in correct lines.

### Automation testing

fixed failing test

### Issue reference
https://github.com/kiali/kiali/issues/7522


josunect (commented):
image

That should not be validating the spec.hosts? foo is a not existent one. 

I think it would be interesting that the multi clusters e2e tests cover some of this validations.

hhovsepy:
> 
> That should not be validating the spec.hosts? foo is a not existent one.
> 

KIA1101 and KIA1107 are for `spec.route.hosts`.
For `spec.hosts` there is 'KIA1106 More than one Virtual Service for same host'

josunect:
> > That should not be validating the spec.hosts? foo is a not existent one.
> 
> KIA1101 and KIA1107 are for `spec.route.hosts`. For `spec.hosts` there is 'KIA1106 More than one Virtual Service for same host'

So, the validations looks correct, then.

josunect (approved):
LGTM

PR #7534, opened 2024-07-09
Namespace discovery proposal

nrfox:
### Describe the change

Provide more details for namespace discovery proposal.


design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,17 +12,18 @@ Provide a better way for namespace discovery within both server and operator.
 
 # Motivation
 
-Currently, the Kiali Server and Operator have many different ways to configure which namespaces should be accessible and viewable by the Kiali UI. The configurations are complex and confusing to users. We want to make things more easily configurable and understandable while still maintaining the ability for Kiali to have either cluster-wide access or limited access to specific namespaces.
+Currently, the Kiali Server and Operator have many different ways to configure which namespaces should be accessible and viewable by the Kiali UI. The configurations are complex and confusing to users. We want to make things more easily configurable and understandable while still maintaining the ability for Kiali to have either cluster-wide access or limited access to specific namespaces. We also want kiali to support soft multitenancy using the same mechanism as istio to make configuration more consistent with istio and easier for users.
```
jmazzitelli, line 15: We need to define what we mean by "soft" multitenancy somewhere in here. Maybe even define "multitenancy" to differentiate what we mean.

Oh, and missing capitalization causes my OCD to flare up :)
```suggestion
Currently, the Kiali Server and Operator have many different ways to configure which namespaces should be accessible and viewable by the Kiali UI. The configurations are complex and confusing to users. We want to make things more easily configurable and understandable while still maintaining the ability for Kiali to have either cluster-wide access or limited access to specific namespaces. We also want Kiali to support soft multitenancy using the same mechanism as Istio to make configuration more consistent with Istio and easier for users.
```

nrfox: "soft" in this case means kiali still has cluster wide access but limits what it shows / operates on to the namespaces selected by discovery selectors.

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -99,6 +99,30 @@ TODO: external controlplane
 
 Since the discovery selectors act as a way to separate multiple Kiali deployments from one another, when Kiali relies on autodiscovering the Istio discovery selectors and it cannot auto discover them, the Kiali server will fail to start and/or panic rather than potentially show out of scope data. In a multi-primary deployment, if one primary cannot be reached then that cluster will be considered inaccessible.
 
+## Discovery Selector Kiali Config Option
+
+A new configuration option will be added that will mimic the Istio discovery selectors configuration option. Using the configuration option, users will be able to specify which namespaces Kiali should be scoped to. Having a config option within Kiali has the advantage of not relying on autodiscovery of the Istio controlplanes. In cases where Kiali temporarily cannot access the control plane configuration, Kiali does not need to fail since it will rely on its own discovery selectors to scope itself. The configuration option needs to provide enough flexibility for users to specify different discovery selectors for each cluster in multi-primary scenarios.
```
jshaughn, line 104:
```suggestion
A new configuration option will be added that will mimic the Istio `discoverySelectors` configuration option. Using the configuration option, users will be able to similarly scope Kiali to the defined set of namespaces. Having a config option within Kiali has the advantage of not relying on auto-discovery via the Istio control plane. Kiali does not need to fail if it cannot, even temporarily, access the control plane configuration, because it can rely on its own discovery selectors configuration. The configuration option needs to provide enough flexibility for users to specify different discovery selectors for each cluster in multi-primary scenarios.
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -99,6 +99,30 @@ TODO: external controlplane
...
+
+Kiali would add a top level `discoverySelectors` config option where you can specify a `map[string]labelselector` where the key is the cluster name that the discovery selector will apply to.
```
jshaughn, line 106: I think perhaps we should have a global discoverySelector and only cluster overrides.  This would allow a simple definition that can apply to all clusters, which makes things easy for single-cluster and scenarios where all clusters use the same thing.

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,116 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+## Discovery Selector Kiali Config Option
+
+A new configuration option will be added that will mimic the Istio discovery selectors configuration option. Using the configuration option, users will be able to specify which namespaces Kiali should be scoped to. Having a config option within Kiali has the advantage of not relying on autodiscovery of the Istio controlplanes. In cases where Kiali temporarily cannot access the control plane configuration, Kiali does not need to fail since it will rely on its own discovery selectors to scope itself. The configuration option needs to provide enough flexibility for users to specify different discovery selectors for each cluster in multi-primary scenarios.
+
+Kiali would add a top level `discoverySelectors` config option where you can specify a `map[string]labelselector` where the key is the cluster name that the discovery selector will apply to.
```
jmazzitelli, line 106: I don't think it should be top level - seems appropriate under "deployment" - where "accessible_namespaces" are today.

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,116 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+TODO: external controlplane
+
+### Failure scenarios
+
+Since the discovery selectors act as a way to separate multiple Kiali deployments from one another, when Kiali relies on autodiscovering the Istio discovery selectors and it cannot auto discover them, the Kiali server will fail to start and/or panic rather than potentially show out of scope data. In a multi-primary deployment, if one primary cannot be reached then that cluster will be considered inaccessible.
+
+## Discovery Selector Kiali Config Option
+
+A new configuration option will be added that will mimic the Istio discovery selectors configuration option. Using the configuration option, users will be able to specify which namespaces Kiali should be scoped to. Having a config option within Kiali has the advantage of not relying on autodiscovery of the Istio controlplanes. In cases where Kiali temporarily cannot access the control plane configuration, Kiali does not need to fail since it will rely on its own discovery selectors to scope itself. The configuration option needs to provide enough flexibility for users to specify different discovery selectors for each cluster in multi-primary scenarios.
```
jmazzitelli, line 104: Question is default behavior - IF the user does not set the  Kiali DS, what is the default behavior? I assume it is, "Use the Istio DS." If Istio DS cannot be found, then do what? Abort with error, I assume.

If the user DOES specify Kiali DS, then I assume Istio DS is ignored entirely.

jshaughn: Do we need an option to use Istio DS *if possible*?

design/KEPS/namespace-discovery/proposal.md
nrfox: I like that option.

design/KEPS/namespace-discovery/proposal.md
nrfox: I wasn't sure the best place for it. `Deployment` seems to be config options related to how Kiali is deployed. The other options in there are stuff like `InstanceName`, `Namespace`, etc. all things related to the actual Kiali deployment. Whereas this option doesn't have anything to do with how Kiali itself is deployed. I think `accessible_namespaces` made sense there because the operator would change how Kiali was deployed based on that field but that is what `cluster_wide_access` is now for. I'm happy to change this though.

design/KEPS/namespace-discovery/proposal.md
jmazzitelli: Is it a valid edge case that clusters can have different DS? I'm sure Istio allows it (because, one control plane probably doesn't care about another control plane's DS) but is it a valid thing to do (do people do that?). 

And I thought Kiali assumes "namespace sameness" in multi-cluster?

design/KEPS/namespace-discovery/proposal.md
jmazzitelli: I don't think so, unless a nil Kiali DS can have more than one semantic. I assume you either set Kiali DS and ignore Istio DS, or you do NOT set Kiali DS and you use Istio DS.

When would you set Kiali DS AND STILL want to use Istio DS?
When would you NOT set Kiali DS AND NOT want to use Istio DS either?

nrfox: > Question is default behavior - IF the user does not set the Kiali DS, what is the default behavior? I assume it is, "Use the Istio DS." If Istio DS cannot be found, then do what? Abort with error, I assume.

Yes.

> If the user DOES specify Kiali DS, then I assume Istio DS is ignored entirely.

Yes.

> Do we need an option to use Istio DS if possible?

Something like "fallback to Kiali DS but try to use Istio DS first"? That would be useful but I also don't want to add more config options than we have to.

jmazzitelli: > Something like "fallback to Kiali DS but try to use Istio DS first"? 

Oh, that's an interesting one. Er... I agree I don't want more config than we have to. Is there a real use-case for this "try Istio DS first, even if Kiali DS is set"?

design/KEPS/namespace-discovery/proposal.md
jmazzitelli: Could go under "istio" since this is supposed to declaring the mesh namespaces, but I don't want to add one more thing there :)

`deployment` is still applicable in the sense that it will determine where the Roles will be deployed (just like accessible namespaces).

This just doesn't seem a large enough "thing" to warrant being its own top-level setting. And since it kinda replaces accessible_namespaces, seems like it will be less confusing to put it under the  same category.

I am not strongly of opinion either way though :)

Make sure its called "discovery_selectors" though (snake_case, not camelCase) to remain consistent with everything else. :)

design/KEPS/namespace-discovery/proposal.md
nrfox: > Is it a valid edge case that clusters can have different DS? I'm sure Istio allows it (because, one control plane probably doesn't care about another control plane's DS) but is it a valid thing to do (do people do that?).

I'd say yes. It's perfectly valid to have namespaces that exist in one cluster but don't exist on another.

> And I thought Kiali assumes "namespace sameness" in multi-cluster?

You can't assume that the same namespace will exist on both clusters and that the two clusters will be exactly the same. The actual namespace object only exists inside a single cluster even if istio logically treats an app deployed in the same namespace across both clusters as the same app.

nrfox: > We could probably hold off on the cluster-specific impl until it's proven we need it.

I'd rather give users the ability to specify this now or at the very least keep the structure the same where we have a `global:` setting so that we can easily add cluster specific overrides later.

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+```yaml
+spec:
+  deployment:
+    cluster_wide_access: false
+    discovery_selectors:
+      - matchLabels:
+          kubernetes.io/metadata.name: backend-app1
+      - matchLabels:
+          kubernetes.io/metadata.name: backend-app2
```
jmazzitelli, line 151: I actually like this better than the "keep accessible_namespaces" idea - even though the latter would be easier to implement than the former :) We'd completely get rid of accessible_namespaces which avoids any confusion about whether the "old" accessible_namespaces still works or not and just keeps one way in the server to get the list of namespaces - through discovery selectors (and we already have code in the server that does that).

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+discoverySelectors:
+  - matchLabels:
+      team: backend
+```
+
+The same rules for multiple revisions in a multi-cluster deployment would still apply.
+
+Primary-remote deployments would behave the same as single cluster where the remote cluster is scoped to the primary controlplane's discovery selectors.
+
+TODO: external controlplane
```
jshaughn, line 96:
```suggestion
TODO: external control plane
```

jshaughn (commented):
No content changes, just some wording and mostly `controlplane` -> `control plane`.

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
 - Provide a way for Kiali Server to have either cluster-wide access to all namespaces or limited access to a subset of namespaces.
 - Remove the deprecated functionality that provided the old way of filtering and limiting namespace access.
+- Enable soft multi-tenancy Kiali deployments where multiple Kialis with cluster wide access can be deployed to the same cluster and scoped to specific namespaces.
 
 # Solution
 
-Provide a new list of "discovery selectors" in the same way that Istio itself has discovery selectors. In fact, the solution should include Kiali Server and/or Operator using the Istio Discovery Selectors where appropriate, and using Kiali's own discovery selectors as an optional mechanism to override Istio discovery selectors.
+Provide a new list of "discovery selectors" in the same way that Istio itself has discovery selectors. Kiali will re-use the Istio Discovery Selectors it retreive's from the control plane's `istio` configmap. Kiali will also have a new config option `deployment.discovery_selectors` that, when specified, will be used instead of the control plane's Istio Discovery Selectors.
```
jmazzitelli, line 26:
```suggestion
Provide a new list of "discovery selectors" in the same way that Istio itself has discovery selectors. Kiali will re-use the Istio Discovery Selectors it retreives from the control plane's `istio` configmap. Kiali will also have a new config option `deployment.discovery_selectors` that, when specified, will be used instead of the control plane's Istio Discovery Selectors.
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
 
-Kiali must prepare for the situation where Istio has no discovery selectors defined or Istio's discovery selectors configuration cannot be found (either Kiali does not have access to see the Istio ConfigMap or Istio simply isn't installed on the cluster where Kiali is located).
+## Istio Discovery Selectors
+
+Kiali already autodiscovers the istio control planes. Discovery selectors for the istio control plane can be found by looking at the control plane's `istio` configmap under `mesh.discoverySelectors`. Kiali will read this configmap and if discovery selectors have been set, Kiali will scope itself to the namespaces selected by the discovery selectors. In cases where there are multiple control planes, for example with revisioned upgrades, Kiali will read both discovery selectors and OR the two together scoping itself to the disjoint of the two sets. For example:
```
jmazzitelli, line 30:
```suggestion
Kiali already autodiscovers the Istio control planes. Discovery selectors for the Istio control plane can be found by looking at the control plane's `istio` configmap under `mesh.discoverySelectors`. Kiali will read this configmap and if discovery selectors have been set, Kiali will scope itself to the namespaces selected by the discovery selectors. In cases where there are multiple control planes, for example with revisioned upgrades, Kiali will read both discovery selectors and OR the two together scoping itself to the disjoint of the two sets. For example:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+```yaml
+discoverySelectors:
+  - matchLabels:
+      team: backend
+  - matchLabels:
+      team: api
+```
+
+For multi-primary cluster deployments, Kiali will scope the namespaces according to the discovery selector's defined on that cluster.
```
jmazzitelli, line 58:
```suggestion
For multi-primary cluster deployments, Kiali will scope the namespaces according to the discovery selectors defined on that cluster.
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+Kiali would add a `deployment.discovery_selectors` config option where you can specify global discovery selectors or cluster specific overrides `map[string]labelselector` where the key is the cluster name that the discovery selector will apply to. Cluster specific overrides will take precedence over the global option for that cluster.
+
+```yaml
+discoverySelectors:
+  global:
+    - labelSelector:
+  cluster_name:
+    - labelSelector:
```
jmazzitelli, line 113: Follow operator convention - snake_case:
```suggestion
discovery_selectors:
  global:
    - label_selector:
  cluster_name:
    - label_selector:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+For multi-primary with different discovery selectors:
+
+```yaml
+discoverySelectors:
+  east:
+    - matchLabels:
+        team: backend
+  west:
+    - matchLabels:
+        team: api
```
jmazzitelli, line 125: Follow operator convention - snake_case:

```suggestion
discovery_selectors:
  east:
    - match_labels:
        team: backend
  west:
    - match_labels:
        team: api
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+## No cluster wide access
+
+Kiali relies on `deployment.accessible_namespaces` when `deployment.cluster_wide_access = false` to know which namespaces the server should create a kube cache for. The operator also uses `deployment.accessible_namespaces` to create a `Role/RoleBinding` for the Kiali server's Service Account giving the Kiali server access to that namespace. In order for the operator to create RBAC rules for the Kiali server, the operator will need to reconcile the discovery selectors with namespaces that exist in the cluster. The operator can take what is set in `deployment.discovery_selectors` and set the Kiali server's `deployment.discovery_selectors` config field to be the result of this reconcilation. For example:
+
+```yaml
+spec:
+  deployment:
+    cluster_wide_access: false
+    discovery_selectors:
+      - matchLabels:
```
jmazzitelli, line 137: Follow operator convention - snake_case:

```suggestion
      - match_labels:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+would be translated into the Kiali config as:
+
+```yaml
+spec:
+  deployment:
+    cluster_wide_access: false
+    discovery_selectors:
+      - matchLabels:
+          kubernetes.io/metadata.name: backend-app1
+      - matchLabels:
```
jmazzitelli, line 150: Follow operator convention - snake_case:

```suggestion
      - match_labels:
          kubernetes.io/metadata.name: backend-app1
      - match_labels:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+```yaml
+spec:
+  deployment:
+    accessible_namespaces:
+      - backend-app1
+      - backend-app2
+    cluster_wide_access: false
+    discovery_selectors:
+      - matchLabels:
```
jmazzitelli, line 164: Follow operator convention - snake_case:

```suggestion
      - match_labels:
```

jmazzitelli (changes requested):
trivial spelling fixes, capitalize Istio, use the kiali operator naming convention of snake_case

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
+  - matchLabels:
+      team: api
+```
+
+Kiali discovery selector cluster B:
+
+```yaml
+discoverySelectors:
+  - matchLabels:
+      team: backend
```
jmazzitelli, line 89: Follow operator convention - snake_case:

```suggestion
discovery_selectors:
  - match_labels:
      team: backend
```

jmazzitelli (changes requested):
more operator convention name changes

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,162 @@ Provide a better way for namespace discovery within both server and operator.
...
 - Provide a way for Kiali Server to have either cluster-wide access to all namespaces or limited access to a subset of namespaces.
 - Remove the deprecated functionality that provided the old way of filtering and limiting namespace access.
+- Enable soft multi-tenancy Kiali deployments where multiple Kialis with cluster wide access can be deployed to the same cluster and scoped to specific namespaces.
 
 # Solution
 
-Provide a new list of "discovery selectors" in the same way that Istio itself has discovery selectors. In fact, the solution should include Kiali Server and/or Operator using the Istio Discovery Selectors where appropriate, and using Kiali's own discovery selectors as an optional mechanism to override Istio discovery selectors.
+Provide a new list of "discovery selectors" in the same way that Istio itself has discovery selectors. Kiali will re-use the Istio Discovery Selectors it retreives from the control plane's `istio` configmap. Kiali will also have a new config option `deployment.discovery_selectors` that, when specified, will be used instead of the control plane's Istio Discovery Selectors.
```
jmazzitelli, line 26:
````suggestion
Provide a new list of "discovery selectors" in the same way that Istio itself has discovery selectors. Kiali will re-use the Istio Discovery Selectors it retrieves from the control plane's `istio` configmap. Kiali will also have a new config option `deployment.discovery_selectors` that, when specified, will be used instead of the control plane's Istio Discovery Selectors. It will support [Kubernetes set-based selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements).  For example:
```yaml
spec:
  deployment:
    discovery_selectors:
      match_labels:
        abc: def
        uvw: xyz
      match_expressions:
      - key: tier
        operator: In
        values: ["production"]
      - key: region
        operator: NotIn
        values: ["east"]
```
````

**Review by jmazzitelli — CHANGES_REQUESTED** [review]

File: design/KEPS/namespace-discovery/proposal.md (hunk 3 above)
Comment (jmazzitelli) [reply design/KEPS/namespace-discovery/proposal.md]: I'll add "global:" to all the examples where there is none here in this document. The one issue I'm wrestling with is how will the operator know the home cluster name so it can do its own override when processing the selectors (because it needs to do this in order to create the Roles in the proper places).

For example, suppose we have this, for a hard multi-tenancy setup:

```yaml
spec:
  deployment:
    cluster_wide_access: false
    discovery_selectors:
      global:
      - match_labels:
          foo: howdy
      cluster-a:
      - match_labels:
          foo: hello
      cluster-b:
      - match_labels:
          foo: wassup
```

Suppose Kiali's home cluster is `cluster-b` -- how does the operator know that? Because the operator will need to know that, since it will need to know that `cluster-b` is overriding `global`.

(BTW: I'm still not clear on what it means for one cluster's selectors to override the global selectors - I assume it means the cluster-specific one just completely overrides the global, where I could have a dozen more clusters, but because they aren't specified in `discovery_selectors`, then just the `global` ones are used, but for cluster-a and cluster-b, the entire array of selectors under global are ignored.)

jmazzitelli: Hmm.. I think I might have an answer to my question above about "how does the operator know the home cluster?". The operator can look at [spec.clustering.clusters[*].name](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.clustering.clusters[*].name) - those are all the remote cluster names. If there is an override cluster section under "discovery_selectors" that does NOT appear in that list of remote cluster names, it must be the home cluster. This assumes every remote cluster is specified in that remote cluster section (even those not accessible).

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,183 @@ Provide a better way for namespace discovery within both server and operator.
...
+```
+
+Kiali discovery selector:
+
+```yaml
+discovery_selectors:
+  - match_labels:
+      team: backend
+  - match_labels:
+      team: api
```
jmazzitelli, line 76:
```suggestion
```yaml
discovery_selectors:
  global:
  - match_labels:
      team: backend
  - match_labels:
      team: api
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,183 @@ Provide a better way for namespace discovery within both server and operator.
...
+```
+
+Kiali discovery selector cluster B:
+
+```yaml
+discovery_selectors:
+  - match_labels:
+      team: backend
+```
```
jmazzitelli, line 111:
````suggestion
Kiali discovery selector cluster A:

```yaml
discovery_selectors:
  global:
  - match_labels:
      team: api
```

Kiali discovery selector cluster B:

```yaml
discovery_selectors:
  global:
  - match_labels:
      team: backend
```
````

File: design/KEPS/namespace-discovery/proposal.md [hunk 20]
```diff
@@ -12,19 +12,183 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+Kiali relies on `deployment.accessible_namespaces` when `deployment.cluster_wide_access = false` to know which namespaces the server should create a kube cache for. The operator also uses `deployment.accessible_namespaces` to create a `Role/RoleBinding` for the Kiali server's Service Account giving the Kiali server access to that namespace. In order for the operator to create RBAC rules for the Kiali server, the operator will need to reconcile the discovery selectors with namespaces that exist in the cluster. The operator can take what is set in `deployment.discovery_selectors` and set the Kiali server's `deployment.discovery_selectors` config field to be the result of this reconcilation. For example:
+
+```yaml
+spec:
+  deployment:
+    cluster_wide_access: false
+    discovery_selectors:
+      - match_labels:
+          team: backend
```
jmazzitelli, line 159:
```suggestion
    discovery_selectors:
      global:
      - match_labels:
          team: backend
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,183 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+```yaml
+spec:
+  deployment:
+    cluster_wide_access: false
+    discovery_selectors:
+      - match_labels:
+          kubernetes.io/metadata.name: backend-app1
+      - match_labels:
+          kubernetes.io/metadata.name: backend-app2
```
jmazzitelli, line 172:
```suggestion
    discovery_selectors:
      global:
      - match_labels:
          kubernetes.io/metadata.name: backend-app1
      - match_labels:
          kubernetes.io/metadata.name: backend-app2
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,183 @@ Provide a better way for namespace discovery within both server and operator.
...
+```yaml
+spec:
+  deployment:
+    accessible_namespaces:
+      - backend-app1
+      - backend-app2
+    cluster_wide_access: false
+    discovery_selectors:
+      - match_labels:
+          team: backend
```
jmazzitelli, line 186:
```suggestion
    discovery_selectors:
      global:
      - match_labels:
          team: backend
```

jmazzitelli (changes requested):

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,189 @@ Provide a better way for namespace discovery within both server and operator.
...
+      - key: tier
+        operator: In
+        values: ["production"]
+      - key: region
+        operator: NotIn
+        values: ["east"]
+    - match_expressions:
+      - key: country
+        operator: In
+        values: ["us", "ca"]
```
jmazzitelli, line 46:
```suggestion
    discovery_selectors:
      global:
      - match_labels:
          abc: def
          uvw: xyz
      - match_labels:
          org: accounting
        match_expressions:
        - key: tier
          operator: In
          values: ["production"]
        - key: region
          operator: NotIn
          values: ["east"]
      - match_expressions:
        - key: country
          operator: In
          values: ["us", "ca"]
```

jmazzitelli (changes requested):
missed a spot for global section

design/KEPS/namespace-discovery/proposal.md
jshaughn: Instead of `global` maybe it should be `default`.  I would expect if there is a cluster-override it would not be additive, it would be a replacement.

jmazzitelli: We could also say the "global" (or "default") simply IS the home cluster. So maybe instead of global or default we call it "home" or "kiali-home" or "local" (we want it unique enough for it not to conflict with any cluster name a user might want to use - so "kiali-home" seems like something people won't use, but "local" and "home" seem more useful cluster names for users to want to use).

So you would define the kiali home cluster and it is used as the default (which, when you think about it, is likely going to be the common global selectors anyway - at least for the majority of use cases. So that seems like a viable "default" for "global" selectors).

So instead of:

```yaml
deployment_selectors:
  global:
    foo: hello
  cluster-a:
    foo: howdy
```

it would be:
```yaml
deployment_selectors:
  kiali-home:
    foo: hello
  cluster-a:
    foo: howdy
```

That seems to make sense to me

jshaughn (approved):
I think we can merge this updated KEP, it's in a good state.

design/KEPS/namespace-discovery/proposal.md
jmazzitelli: I also think that we should change "global" for no other reason than that seems like it could potentially be a cluster name that a user might want ("global" as in "worldwide", then you have different regions like "emea", "na", etc).

I think it should be "kiali-home" because (a) it is more descriptive and (b) likely won't clash with any name a user might want to use.

jmazzitelli (approved):
I'm going to approve this. Someone look at my comments here: https://github.com/kiali/kiali/pull/7534#discussion_r1676385510

and see if we should change anything. If no, I'm OK just approving this as is. I'm still not 100% sold on the name of that "global" section, but I also dont' know a good name to use :)

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
 
 # Solution
 
-Provide a new list of "discovery selectors" in the same way that Istio itself has discovery selectors. In fact, the solution should include Kiali Server and/or Operator using the Istio Discovery Selectors where appropriate, and using Kiali's own discovery selectors as an optional mechanism to override Istio discovery selectors.
+Provide a new list of "discovery selectors" in the same way that [Istio itself has discovery selectors](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig). Kiali will re-use the Istio Discovery Selectors it retrieves from the control plane's `istio` configmap. Kiali will also have a new config option `deployment.discovery_selectors` that, when specified, will be used instead of the control plane's Istio Discovery Selectors. It will support an array of [Kubernetes set-based selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements) just like Istio Discovery Selectors.  For example:
+```yaml
+spec:
+  deployment:
+    discovery_selectors:
+      global:
```
jmazzitelli, line 31:
```suggestion
      default:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+discoverySelectors:
+  - matchLabels:
+      team: backend
+```
+
+Kiali discovery selector:
+
+```yaml
+discovery_selectors:
+  global:
```
jmazzitelli, line 74:
```suggestion
  default:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+discoverySelectors:
+  - matchLabels:
+      team: backend
+```
+
+Kiali discovery selector cluster A:
+
+```yaml
+discovery_selectors:
+  global:
```
jmazzitelli, line 103:
```suggestion
  default:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+  - match_labels:
+      team: api
+```
+
+Kiali discovery selector cluster B:
+
+```yaml
+discovery_selectors:
+  global:
```
jmazzitelli, line 112:
```suggestion
  default:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+## Discovery Selector Kiali Config Option
+
+A new configuration option will be added that will mimic the Istio `discoverySelectors` configuration option. Using the configuration option, users will be able to similarly scope Kiali to the defined set of namespaces. This configuration option would take precedence over the autodiscovered control plane discovery selectors. Having a config option within Kiali has the advantage of not relying on auto-discovery via the Istio control plane. Kiali does not need to fail if it cannot, even temporarily, access the control plane configuration, because it can rely on its own discovery selectors configuration. The configuration option needs to provide enough flexibility for users to specify different discovery selectors for each cluster in multi-primary scenarios.
+
+Kiali would add a `deployment.discovery_selectors` config option where you can specify global discovery selectors or cluster specific overrides `map[string]labelselector` where the key is the cluster name that the discovery selector will apply to. Cluster specific overrides will take precedence over the global option for that cluster.
```
jmazzitelli, line 131:
```suggestion
Kiali would add a `deployment.discovery_selectors` config option where you can specify a default set of discovery selectors and additional cluster-specific overrides `map[string]labelselector` where the key is the cluster name that the discovery selector will apply to. Cluster specific overrides will take precedence over the defaults.
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+```yaml
+discovery_selectors:
+  global:
```
jmazzitelli, line 135:
```suggestion
  default:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+## No cluster wide access
+
+Kiali relies on `deployment.accessible_namespaces` when `deployment.cluster_wide_access = false` to know which namespaces the server should create a kube cache for. The operator also uses `deployment.accessible_namespaces` to create a `Role/RoleBinding` for the Kiali server's Service Account giving the Kiali server access to that namespace. In order for the operator to create RBAC rules for the Kiali server, the operator will need to reconcile the discovery selectors with namespaces that exist in the cluster. The operator can take what is set in `deployment.discovery_selectors` and set the Kiali server's `deployment.discovery_selectors` config field to be the result of this reconcilation. For example:
+
+```yaml
+spec:
+  deployment:
+    cluster_wide_access: false
+    discovery_selectors:
+      global:
```
jmazzitelli, line 162:
```suggestion
      default:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+```
+
+would be translated into the Kiali config as:
+
+```yaml
+spec:
+  deployment:
+    cluster_wide_access: false
+    discovery_selectors:
+      global:
```
jmazzitelli, line 174:
```suggestion
      default:
```

design/KEPS/namespace-discovery/proposal.md
```diff
@@ -12,19 +12,190 @@ Provide a better way for namespace discovery within both server and operator.
...
+
+```yaml
+spec:
+  deployment:
+    accessible_namespaces:
+      - backend-app1
+      - backend-app2
+    cluster_wide_access: false
+    discovery_selectors:
+      global:
```
jmazzitelli, line 191:
```suggestion
      default:
```

jmazzitelli (changes requested):

jmazzitelli (approved):

design/KEPS/namespace-discovery/proposal.md
jmazzitelli: we ended up just going with "default" for now.

jshaughn (approved):
This looks good.

design/KEPS/namespace-discovery/proposal.md
nrfox: > The one issue I'm wrestling with is how will the operator know the home cluster name so it can do its own override when processing the selectors (because it needs to do this in order to create the Roles in the proper places)

The operator should not be cluster aware. Today iirc the "hard multi-tenancy mode" of `accessible_namespaces` does not work for multi-cluster. Unless that's changing the operator would just configure the roles for the cluster that the operator is deployed on and not try to do anything cluster related.

nrfox: > I also think that we should change "global" for no other reason than that seems like it could potentially be a cluster name that a user might want ("global" as in "worldwide", then you have different regions like "emea", "na", etc).

Hmmm that's a good point and any name will eventually conflict so imo we should just change the structure to something like:
```
deployment_selectors:
  global (or default):
    foo: hello
  overrides:
    <cluster-name>:
      foo: howdy
```
Merging global selectors with override selectors would be more complicated so I think the simplest heuristic would be to only use what is specified in `overrides` if you have specified anything.

jshaughn: I have no problem with adding an `overrides`.  We already preferred `default`, and `overrides` works well with `default`.   I'm -1, though on ignoring `default` if

nrfox: > I think you mean that if there is a cluster-specific override that you would ignore default for that cluster, and any cluster not included in overrides would get the defaults.

Yes exactly

jmazzitelli: OK. let's change this KEP to refer to this extra level "overrides".

PR #7536, opened 2024-07-11
Fix url matching for openshift login negative test

nrfox:
### Describe the change

Fix url matching for openshift login negative test. Uses new auth callback url.

### Steps to test the PR

Run `kiali_login` test on openshift with the changes.

### Automation testing

Fixes e2e test.


ScriptingShrimp (approved):
passing on OCP

Issue #7524, opened 2024-07-05
Rework - Duplicate labels in Kiali CR and code


hhovsepy:
Currently Kiali CR has duplicate labels in `istio_labels:` and in `external_services.istio.component_status.components`.
Also there are sill hardcoded "app=istiod" label in Kiali code, but this label is duplicated in `component_status.components`.

The task is to rework the component status config and Kiali code of checking components to:
Have single place of labels in Kiali CR without duplicates
Reuse same labels in Kiali code.

The issue was initiated in a discussion over a PR: https://github.com/kiali/kiali-operator/pull/788

hhovsepy:
Related to this one https://github.com/kiali/kiali/issues/7501

Issue #7501, opened 2024-07-02
Autodiscover gateways


nrfox:
### What do you want to improve?

Autodetect istio gateways.

### What is the current behavior?

Currently Kiali uses the configuration in [external_services.istio.component_status.components](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.istio.component_status.components) to monitor and report in the masthead the health status of istio gateways. The default configuration assumes that you have deployed istio ingress and egress gateways with certain app labels that may not be present. And if the manual configuration doesn't match what you've deployed, Kiali will show errors in the masthead:
![Screenshot from 2024-07-02 10-30-06](https://github.com/kiali/kiali/assets/6226732/5a264121-7431-43b2-a1df-50f47b8ab80a)

### What is the new behavior?

Gateways are autodetected and manual configuration under `external_services.istio.component_status` is deprecated.

PR #7538, opened 2024-07-15
Removed 'external_services.istio.component_status.components' duplicate labels.

hhovsepy:
### Describe the change

`external_services.istio.component_status.components` was expecting to have values for `istiod` and `ingress/egressgateway`, which was causing duplication for Istio component statuses, also duplication of labels in configuration and in code.
Now `external_services.istio.component_status.components` by default made empty, and `istiod` and `ingress/egressgateway` component statuses are read in code.
Functionality of  `external_services.istio.component_status.components` is left as deprecated for now, until will be removed, or will left as empty for possible custom components such as `eastwestgateway`.

### Steps to test the PR
Make sure you have 'istiod' without pods sunning.
Before:
Screenshot from 2024-07-15 14-08-00

After:
Screenshot from 2024-07-15 13-34-34

### Automation testing
Tests modified.

### Issue reference
https://github.com/kiali/kiali/issues/7501
https://github.com/kiali/kiali/issues/7524
Operator PR https://github.com/kiali/kiali-operator/pull/795

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -279,6 +279,7 @@ type IstioConfig struct {
 	ConfigMapName                     string              `yaml:"config_map_name,omitempty"`
 	EnvoyAdminLocalPort               int                 `yaml:"envoy_admin_local_port,omitempty"`
 	GatewayAPIClasses                 []GatewayAPIClass   `yaml:"gateway_api_classes,omitempty"`
+	GatewayNamespace                  string              `yaml:"gateway_namespace,omitempty"`
 	IstioAPIEnabled                   bool                `yaml:"istio_api_enabled"`
 	IstioCanaryRevision               IstioCanaryRevision `yaml:"istio_canary_revision,omitempty"`
 	IstioIdentityDomain               string              `yaml:"istio_identity_domain,omitempty"`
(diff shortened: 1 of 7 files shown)


jshaughn (commented):
@hhovsepy I see a difference in status, but for me I still see a duplicate, is that expected?  I didn't start with a clean CR, would that make a difference?

Before:
image
After:
image

hhovsepy:
Hi @jshaughn , is there a 'external_services.istio.component_status.components' remaining in your local configuration file?

jshaughn:
@hhovsepy I think I must have an issue with my test env.  Let me try again...

config/config.go
```diff
@@ -688,23 +689,6 @@ func NewConfig() (c *Config) {
...
-							AppLabel: "istio-ingressgateway",
-							IsCore:   true,
-							IsProxy:  true,
-						},
-						{
-							AppLabel: "istiod",
-							IsCore:   true,
-							IsProxy:  false,
-						},
-					},
```
jshaughn, line 707: > Functionality of external_services.istio.component_status.components is left as deprecated for now, until will be removed, or will left as empty for possible custom components such as eastwestgateway.

Do we really want to leave this unused Config?  If we think it's going to be useful in the near future, then maybe, but otherwise I'm not sure it's worth keeping.  If nothing else I think a comment may be appropriate in the type definition.

mesh/api/testdata/test_mesh_graph.expected
```diff
@@ -230,33 +230,12 @@
...
-                },
-                {
-                  "AppLabel": "istiod",
-                  "IsCore": true,
-                  "IsProxy": false,
-                  "IsMultiCluster": false,
-                  "Namespace": ""
-                }
-              ]
+              "Components": null
```
jshaughn, line 233: This doesn't look right.  I would have expected it to be omitted, or be an empty array.

jshaughn (commented):
OK, sorry, on re-test it looks as expected: 
image

A couple of comments...

config/config.go
hhovsepy: The reason it is left is, that there can be components such as "istio-eastwestgateway", or any other custom component users might want to be checked. So far we can mark it as deprecated.

mesh/api/testdata/test_mesh_graph.expected
hhovsepy: yes, thanks, should be an empty array

jshaughn (approved):
Looks good!

PR #7540, opened 2024-07-15
[hack] allow skupper demo script be able to log into OpenShift cluster via token

jmazzitelli:


jshaughn (dismissed):
It's fine but should this target my POC PR or master?

PR #7544, opened 2024-07-17
[hack-scripts] Add waiting for bookinfo app before the traffic genera…

mkralik3:
### Describe the change

The traffic generator is installed right after bookinfo app, however, it doesn't wait till bookinfo app is ready. So, if it takes longer (e.g. cluster is still pulling the images), the traffic generator starts sending requests to the app that is not ready. So results in the traffic generator contain 503 errors

### DIFF
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -430,6 +430,14 @@ if [ "${TRAFFIC_GENERATOR_ENABLED}" == "true" ]; then
       fi
     fi
   fi
+  # # Do not create traffic generator until all pods are ready
+  echo "Waiting till all pods are ready before installing the traffic generator"
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/details-v1 --timeout=5m
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/productpage-v1 --timeout=5m
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/ratings-v1 --timeout=5m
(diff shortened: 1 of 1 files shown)


hack/istio/install-bookinfo-demo.sh
```diff
@@ -430,6 +430,14 @@ if [ "${TRAFFIC_GENERATOR_ENABLED}" == "true" ]; then
...
     fi
   fi
+  # # Do not create traffic generator until all pods are ready
+  echo "Waiting till all pods are ready before installing the traffic generator"
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/details-v1 --timeout=5m
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/productpage-v1 --timeout=5m
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/ratings-v1 --timeout=5m
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/reviews-v1 --timeout=5m
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/reviews-v2 --timeout=5m
+  $CLIENT_EXE -n ${NAMESPACE} wait --for condition=Available deployment/reviews-v3 --timeout=5m
```
jmazzitelli, line 440: There is already a `-wt` option - see if you can utilize that. It sets WAIT_TIMEOUT env var.

Sometimes we don't want this script waiting (locally, I often do not, I just want the script to end and I'm ok with traffic gen getting some errors). So we don't want this always blocking. But utilize the `-wt` option if you want to wait. See https://github.com/kiali/kiali/pull/7544/files#diff-7bca579da46f87d5c671523bb79d7146eee8f84e94114d8fa4aeeacb520c126aR119

PR #7547, opened 2024-07-18
Hide mesh page for non istio-system users

### LINKED ISSUE #7527: Mesh page - Hide mesh page for non istio-system users

The new mesh page shows information about the istio controlplane but should be hidden for users who don't have read access to the `istio-system` namespace.

Mesh page can be seen by all users.

### What is the new behavior?

Mesh page is hidden for users without read access to `istio-system`.

`/api/mesh/graph` checks for read access to `istio-system`.

josunect:
### Describe the change

- Hide mesh page for non istio-system users
- Don't perform istio/status call if Kiali doesn't have access to the control plane
- Remove the link to the Mesh page from the About window

### Steps to test the PR

It can be tested with token strategy. 
Use a kiali token: 
`
k create token kiali -n istio-system`

The user should see the mesh page, and the Mesh page link in the Kiali about window. 

Create a user with restricted permissions: 
`kubectl create serviceaccount test`

Create role: 
`kubectl apply -f https://raw.githubusercontent.com/josunect/yamls/main/roleBookinfo.yaml`
Create roleBinding: 
`kubectl apply -f https://raw.githubusercontent.com/josunect/yamls/main/roleBindingBookinfo.yaml`

Get a token: 
`kubectl create token test`

Logging, and the user should not see the mesh page: 

image

No link to Mesh in the about window: 

image

### Automation testing

 e2e tests: Added a new "authorization" feature file to test features with a user (**bookinfouser**) that has access to:

- east cluster
- bookinfo namespace (No control plane access)

This is meant to work for the multi primary scripts that set up this user and the permissions with openId.

Some initial tests: 

- The user doesn't see the Mesh menu
- The user doesn't see the istio-system card
- The user see the bookinfo card
- The user doesn't see the Mesh link in the about window

In the mesh page, a couple of tests were added: 

- The user see the Mesh menu
- The user see the Mesh link in the about window

### Issue reference

Fixes #7527

### DIFF
--- a/frontend/src/utils/MeshUtils.ts
+++ b/frontend/src/utils/MeshUtils.ts
@@ -1,0 +1,17 @@
+import { store } from '../store/ConfigStore';
+import { serverConfig } from '../config';
+
+// isControlPlaneAccessible returns true if
+export const isControlPlaneAccessible = (cluster?: string): boolean => {
+  const ns = store.getState().namespaces;
+
+  return (
+    ns.items !== undefined &&
+    ns.items.some(nsItem => {
+      if (cluster) {
+        return nsItem.name === serverConfig.istioNamespace && nsItem.cluster === cluster;
+      }
+      return nsItem.name === serverConfig.istioNamespace;
+    })
+  );
+};
--- a/handlers/config.go
+++ b/handlers/config.go
@@ -1,7 +1,6 @@
 package handlers
 
 import (
-	"context"
 	"fmt"
 	"net/http"
 	"strings"
@@ -51,7 +50,6 @@ type PublicConfig struct {
 	AuthStrategy         string                        `json:"authStrategy,omitempty"`
 	AmbientEnabled       bool                          `json:"ambientEnabled,omitempty"`
 	Clusters             map[string]models.KubeCluster `json:"clusters,omitempty"`
-	ControlPlaneClusters []string                      `json:"controlPlaneClusters,omitempty"`
 	Deployment           DeploymentConfig              `json:"deployment,omitempty"`
 	GatewayAPIClasses    []config.GatewayAPIClass      `json:"gatewayAPIClasses,omitempty"`
 	GatewayAPIEnabled    bool                          `json:"gatewayAPIEnabled,omitempty"`
@@ -82,7 +80,6 @@ func Config(conf *config.Config, discovery *istio.Discovery) http.HandlerFunc {
 			AccessibleNamespaces: conf.Deployment.AccessibleNamespaces,
 			AuthStrategy:         conf.Auth.Strategy,
 			Clusters:             make(map[string]models.KubeCluster),
-			ControlPlaneClusters: []string{},
 			Deployment: DeploymentConfig{
 				ViewOnlyMode: conf.Deployment.ViewOnlyMode,
 			},
@@ -133,12 +130,6 @@ func Config(conf *config.Config, discovery *istio.Discovery) http.HandlerFunc {
 
 			for _, cluster := range clusters {
 				publicConfig.Clusters[cluster.Name] = cluster
-				// Check if there is access to the cluster control plane for Mesh features
-				// TODO: conf.IstioNamespace is valid for all the clusters?
-				ns, err := layer.Namespace.GetClusterNamespace(context.TODO(), conf.IstioNamespace, cluster.Name)
-				if err == nil && ns != nil {
-					publicConfig.ControlPlaneClusters = append(publicConfig.ControlPlaneClusters, cluster.Name)
-				}
 			}
 
 			return nil
--- a/handlers/config.go
+++ b/handlers/config.go
@@ -1,6 +1,7 @@
 package handlers
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"strings"
@@ -50,6 +51,7 @@ type PublicConfig struct {
 	AuthStrategy         string                        `json:"authStrategy,omitempty"`
 	AmbientEnabled       bool                          `json:"ambientEnabled,omitempty"`
 	Clusters             map[string]models.KubeCluster `json:"clusters,omitempty"`
+	ControlPlaneClusters []string                      `json:"controlPlaneClusters,omitempty"`
 	Deployment           DeploymentConfig              `json:"deployment,omitempty"`
 	GatewayAPIClasses    []config.GatewayAPIClass      `json:"gatewayAPIClasses,omitempty"`
 	GatewayAPIEnabled    bool                          `json:"gatewayAPIEnabled,omitempty"`
@@ -80,6 +82,7 @@ func Config(conf *config.Config, discovery *istio.Discovery) http.HandlerFunc {
 			AccessibleNamespaces: conf.Deployment.AccessibleNamespaces,
 			AuthStrategy:         conf.Auth.Strategy,
 			Clusters:             make(map[string]models.KubeCluster),
+			ControlPlaneClusters: []string{},
 			Deployment: DeploymentConfig{
 				ViewOnlyMode: conf.Deployment.ViewOnlyMode,
 			},
@@ -130,6 +133,12 @@ func Config(conf *config.Config, discovery *istio.Discovery) http.HandlerFunc {
 
 			for _, cluster := range clusters {
 				publicConfig.Clusters[cluster.Name] = cluster
+				// Check if there is access to the cluster control plane for Mesh features
+				// TODO: conf.IstioNamespace is valid for all the clusters?
+				ns, err := layer.Namespace.GetClusterNamespace(context.TODO(), conf.IstioNamespace, cluster.Name)
+				if err == nil && ns != nil {
+					publicConfig.ControlPlaneClusters = append(publicConfig.ControlPlaneClusters, cluster.Name)
+				}
 			}
 
 			return nil
--- a/handlers/mesh.go
+++ b/handlers/mesh.go
@@ -3,7 +3,9 @@ package handlers
 import (
 	"context"
 	"net/http"
-	"strings"
+	"slices"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/config"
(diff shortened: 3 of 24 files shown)


jshaughn (commented):
A couple of comments, and also would be good to wait for @nrfox to review.

handlers/config.go
```diff
@@ -50,6 +51,7 @@ type PublicConfig struct {
 	AuthStrategy         string                        `json:"authStrategy,omitempty"`
 	AmbientEnabled       bool                          `json:"ambientEnabled,omitempty"`
 	Clusters             map[string]models.KubeCluster `json:"clusters,omitempty"`
+	ControlPlaneClusters []string                      `json:"controlPlaneClusters,omitempty"`
```
nrfox, line 54: I don't think this belongs on the config because it's not a configuration setting. We just need a way to know if the user has access to the istio system namespace without necessarily calling `/api/mesh`? Maybe it can be added to the session info redux state? Actually as part of the post login actions I think we already have a list of all the namespaces. This can be some redux state derived from that list that just checks for the presence of the istio system namespace for each cluster.

nrfox:
> It can be tested with token strategy.
> Use a kiali token:
> k create token kiali -n istio-system
 
 These steps can be turned into a cypress test that checks for the mesh page being present or not. Ideally it'd also test for the multi-cluster case where the user has access to one istio system namespace but not another.

josunect:
@nrfox @jshaughn I have simplified this frontend logic so: 
- There is a common function to validate the access to the control plane
- The publicConfig item is removed. This function checks if the namespace is available in the namespace list. 

I have added a function to validate the access to the Mesh Control Planes. 

Regarding the tests, I think that's a good idea. This might be included in another PR? I think it could have several implications, depending on which authentication method is used, who configures the user (If this is done by external testing scripts it might be also needed to be updated)

frontend/src/utils/MeshUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+  return (
+    ns.items !== undefined &&
+    ns.items.some(nsItem => {
+      if (cluster) {
+        return nsItem.name === serverConfig.istioNamespace && nsItem.cluster === cluster;
+      }
+      return nsItem.name === serverConfig.istioNamespace;
+    })
+  );
+};
```
nrfox, line 17: This could also just be a redux selector I think instead of a separate function. I'm not sure if it's common to import `store` directly or if that has other side effects.

handlers/mesh.go
```diff
@@ -63,3 +67,13 @@ func MeshGraph(
 		respond(w, code, payload)
 	}
 }
+
+func filterAccessibleControlPlanes(ctx context.Context, namespaceService business.NamespaceService, mesh *models.Mesh) {
+	for i, cp := range mesh.ControlPlanes {
+		// Check if the user is able to access to the control plane
+		_, err := namespaceService.GetClusterNamespace(ctx, cp.IstiodNamespace, cp.Cluster.Name)
+		if err != nil && strings.Contains(err.Error(), "forbidden") {
```
nrfox, line 75: iirc you can just check kuberrors.IsForbidden(err) here: https://pkg.go.dev/k8s.io/apimachinery@v0.30.3/pkg/api/errors#IsForbidden.

handlers/mesh.go
```diff
@@ -63,3 +67,13 @@ func MeshGraph(
...
+			mesh.ControlPlanes = append(mesh.ControlPlanes[:i], mesh.ControlPlanes[i+1:]...)
```
nrfox, line 76: I think there's a nifty func for this now if you want to use it: https://pkg.go.dev/slices#Delete

handlers/mesh.go
```diff
@@ -46,6 +49,7 @@ func MeshGraph(
 		o := mesh.NewOptions(r, &business.Namespace)
 
 		meshInfo, err := discovery.Mesh(r.Context())
+		filterAccessibleControlPlanes(r.Context(), business.Namespace, meshInfo)
```
nrfox, line 52: I'm not sure if we should filter out the controlplanes the user doesn't have access to entirely or if we should show them as inaccessible nodes like what we do with the traffic graph.

nrfox:
> Regarding the tests, I think that's a good idea. This might be included in another PR? I think it could have several implications, depending on which authentication method is used, who configures the user (If this is done by external testing scripts it might be also needed to be updated)

Yes you'd need to use the multi-primary setup which uses openid auth. We need similar tests elsewhere so perhaps a user with access to one cluster but not another could get created as part of the setup scripts.

handlers/mesh.go
josunect: > I'm not sure if we should filter out the controlplanes the user doesn't have access to entirely or if we should show them as inaccessible nodes like what we do with the traffic graph.

Maybe for security reasons, if the user doesn't have access to the cluster it would be better to just hide it?

frontend/src/utils/MeshUtils.ts
josunect: You can use it to access to the redux state outside of the React context. The only thing is that it wouldn't refresh automatically with a state change (But I don't think it is required in this context)

handlers/mesh.go
josunect: I see in the Traffic graph the user can see the inaccessible nodes: 

image

But not sure if that should be the same for the Mesh.

frontend/src/utils/MeshUtils.ts
ferhoyos: Although using a Redux selector is the recommended approach, I think there is no problem with getting the state directly (there are multiple cases in Kiali). The function is called on component render, so there is no problem with automatic refresh.

The redux selector approach that @nrfox mentions would be something like this (not tested):
```
export const isControlPlaneAccessible = (cluster?: string): boolean => {
    return useKialiSelector(state => {
        const ns = state.namespaces;

        return (
          ns.items !== undefined &&
          ns.items.some(nsItem => {
            if (cluster) {
              return nsItem.name === serverConfig.istioNamespace && nsItem.cluster === cluster;
            }

            return nsItem.name === serverConfig.istioNamespace;
          })
        );
      })
```

I don't think so, but if we have any performance issue because of number of namespaces we can use memoized Redux selectors (https://redux.js.org/usage/deriving-data-selectors#writing-memoized-selectors-with-reselect).

josunect:
> > Regarding the tests, I think that's a good idea. This might be included in another PR? I think it could have several implications, depending on which authentication method is used, who configures the user (If this is done by external testing scripts it might be also needed to be updated)
> 
> Yes you'd need to use the multi-primary setup which uses openid auth. We need similar tests elsewhere so perhaps a user with access to one cluster but not another could get created as part of the setup scripts.

@nrfox I've included some e2e tests for this (See the description with the details)

handlers/mesh.go
```diff
@@ -63,3 +69,14 @@ func MeshGraph(
...
+
+func filterAccessibleControlPlanes(ctx context.Context, namespaceService business.NamespaceService, mesh *models.Mesh) {
+
+	for i, cp := range mesh.ControlPlanes {
+		// Check if the user is able to access to the control plane
+		_, err := namespaceService.GetClusterNamespace(ctx, cp.IstiodNamespace, cp.Cluster.Name)
+		if err != nil && errors.IsForbidden(err) {
+			mesh.ControlPlanes = slices.Delete(mesh.ControlPlanes, i, i+1)
+		}
+	}
```
nrfox, line 81: The `mesh.ControlPlanes` slice is being ranged over here and then inside the loop it is also being modified. I'm not sure but this might cause weird side effects. To be safe, you could either build a new slice with just the filtered elements or else keep track of all the indexes to be deleted inside the `range mesh.ControlPlanes` loop and then call `slices.Delete` on them afterward outside the `range mesh.ControlPlanes` loop.

handlers/mesh.go
nrfox: I'm fine with either one. I think if the user didn't have access they would still see the cluster node wouldn't they? It just wouldn't have the controlplane node in there?

josunect:
> > Regarding the tests, I think that's a good idea. This might be included in another PR? I think it could have several implications, depending on which authentication method is used, who configures the user (If this is done by external testing scripts it might be also needed to be updated)
> 
> Yes you'd need to use the multi-primary setup which uses openid auth. We need similar tests elsewhere so perhaps a user with access to one cluster but not another could get created as part of the setup scripts.

But then, maybe we should not hide the mesh page for non istio-system users and just show the Data Plane?

handlers/mesh.go
josunect: Right, maybe that can cause issues

handlers/mesh.go
```diff
@@ -46,6 +50,7 @@ func MeshGraph(
 		o := mesh.NewOptions(r, &business.Namespace)
 
 		meshInfo, err := discovery.Mesh(r.Context())
+		filterAccessibleControlPlanes(r.Context(), business.Namespace, meshInfo)
```
nrfox, line 53: If you try to hit this endpoint as the `bookinfo` user you get a `500` response. Probably need to return a 403 response when we get back a 403 response from every `GetClusterNamespace` check. Reproduce with:

1. Setup multi-primary

    ```
    hack/run-integration-tests.sh --test-suite frontend-multi-primary --setup-only true
    ```
2. Login as bookinfo user and grab cookie from web inspector
3. Make api request with cookie to graph the mesh page
    ```
    curl -v --cookie kiali-token-east=<cookie> <kiali_url/api/mesh/graph
    ```
This is just an edge case though since we hide the mesh page if you don't have access to any controlplane namespaces. Right now the only way to hit this error is to use the api directly outside the UI.

nrfox (dismissed):
Thanks for adding tests for this as well! The one issue is non-blocking since you would only see it if you were using the API directly outside the UI.

nrfox:
> But then, maybe we should not hide the mesh page for non istio-system users and just show the Data Plane?

No the mesh page is more for mesh admins. Users can get the same info about the data plane namespaces from the overview page.

frontend/src/utils/MeshUtils.ts
nrfox: Thanks for weighing in here @ferhoyos !

handlers/mesh.go
josunect: Yes, it can be reproduced also with the token strategy (See description) with a token for the test SA. 
I have updated the logic, to return 403 in this case.

nrfox (approved):

Issue #7413, opened 2024-06-10
K8s GW API Cross-Namespace routing


hhovsepy:
K8s Gateway API supports cross-namespace routing based on labels on Gateway and Namespace.
More info https://gateway-api.sigs.k8s.io/guides/multiple-ns/

- [x] Add support of cross-namespace validations in Gateway API objects.
- [x] Tests added updated.

PR #7552, opened 2024-07-22
Added cross-namespace validations into K8s Routes.

hhovsepy:
K8s HTTPRoutes and GRPCRoutes can be created in a different namespace than K8sGateway exists and is referred in a parentRef.
For having access to K8sGateway from different namespace:
1. Gateway should be marked by label `shared-gateway-access: "true"`.
2. Namespace where HTTPRoute is created should also be labeled `shared-gateway-access: "true"`.
Then the HTTPRoute with a 'parentRefs' to a K8sGateways from different namespace will be recognized. Otherwise Error `KIA1401 Route is pointing to a non-existent K8s gateway` will be shown.

Fixed GRPCRoutes showing. Added tests.

More info at https://gateway-api.sigs.k8s.io/guides/multiple-ns/

### Describe the change

Added cross namespace validation for GRPCRoute and HTTPRoute objects which point to K8sGateways from different namespace.
GRPCRoute is GA, not experimental.

### Steps to test the PR
Make sure K8s CRDs are installed:
`kubectl get crd gateways.gateway.networking.k8s.io || { kubectl kustomize "github.com/kubernetes-sigs/gateway-api/config/crd?ref=v1.1.0" | kubectl apply -f -; }
`
Create the following objects.

```
kind: Gateway
metadata:
  name: shared-gateway
  namespace: infra-ns
spec:
  gatewayClassName: istio
  listeners:
  - name: https
    hostname: "foo.example.com"
    protocol: HTTPS
    port: 443
    allowedRoutes:
      namespaces:
        from: Selector
        selector:
          matchLabels:
            shared-gateway-access: "true"
---
apiVersion: v1
kind: Namespace
metadata:
  name: infra-ns
  labels:
    shared-gateway-access: "true"
---
apiVersion: v1
kind: Namespace
metadata:
  name: site-ns
  labels:
    shared-gateway-access: "true"
---
apiVersion: v1
kind: Namespace
metadata:
  name: store-ns
  labels:
    shared-gateway-access: "true"
---
apiVersion: v1
kind: Namespace
metadata:
  name: no-external-access
---
kind: HTTPRoute
metadata:
  name: store
  namespace: store-ns
spec:
  parentRefs:
  - name: shared-gateway
    namespace: infra-ns
  rules:
  - matches:
    - path:
        value: /store
---
kind: HTTPRoute
metadata:
  name: home
  namespace: site-ns
spec:
  parentRefs:
  - name: shared-gateway
    namespace: infra-ns
---
kind: HTTPRoute
metadata:
  name: login
  namespace: site-ns
spec:
  parentRefs:
  - name: shared-gateway
    namespace: infra-ns
  rules:
  - matches:
    - path:
        value: /login
---
kind: GRPCRoute
metadata:
  name: example-route
  namespace: store-ns
spec:
  parentRefs:
  - name: shared-gateway
    namespace: infra-ns
  hostnames:
  - "example.com"

```
HTTPRoutes and GRPCRoutes should not display errors on 'parentRefs:'.
Change and spoil the Gateway's `shared-gateway-access: "true"`.
Now errors are displayed on all routes.

### Automation testing

Unit tests are created.

### Issue reference

https://github.com/kiali/kiali/issues/7413

### DIFF
--- a/business/checkers/k8shttproutes/no_k8sgateway_checker.go
+++ b/business/checkers/k8shttproutes/no_k8sgateway_checker.go
@@ -3,6 +3,7 @@ package k8shttproutes
 import (
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/labels"
 	k8s_networking_v1 "sigs.k8s.io/gateway-api/apis/v1"
 
 	"github.com/kiali/kiali/kubernetes"
@@ -10,8 +11,10 @@ import (
 )
 
 type NoK8sGatewayChecker struct {
+	Cluster      string
+	GatewayNames map[string]k8s_networking_v1.Gateway
 	K8sHTTPRoute *k8s_networking_v1.HTTPRoute
-	GatewayNames map[string]struct{}
+	Namespaces   models.Namespaces
 }
 
 // Check validates that the HTTPRoute is pointing to an existing Gateway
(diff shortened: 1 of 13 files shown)


ferhoyos (dismissed):
The validations work. I have modified the namespace label and the gateway selector, and the validation is applied correctly:

image

This is not blocking, but I think we could modify the validation error message to include this use case (the gateway already exists but the route is not allowed to use it).

My suggestion: **'Route is pointing to a non-existent K8s gateway or is not allowed to use it'**

hhovsepy:
Thanks @ferhoyos for the testing, yes updated the message.

ferhoyos (approved):
Better validation message @hhovsepy :slightly_smiling_face:

business/checkers/k8shttproutes/no_k8sgateway_checker.go
```diff
@@ -30,26 +33,46 @@ func (s NoK8sGatewayChecker) ValidateHTTPRouteGateways(validations *[]*models.Is
...
 }
+
+// If K8sGateway's allowedRoutes selector matches the labels of given namespace
+func IsGatewaySharedWithNS(namespace string, cluster string, gw k8s_networking_v1.Gateway, nss models.Namespaces) bool {
+	ns := nss.GetNamespace(namespace, cluster)
+	if ns == nil {
+		return false
+	}
+	for _, l := range gw.Spec.Listeners {
+		if *l.AllowedRoutes.Namespaces.From == "Selector" &&
```
ferhoyos, line 71: @hhovsepy I have just discovered another scenario that conflicts with the KIA1401 validation. According to https://gateway-api.sigs.k8s.io/concepts/api-overview/#restricting-route-attachment, the `namespaces.from` field supports the value `All`, which allow Routes from all Namespaces to be attached.

I used this value in the shared gateway, and the validation error arises when it shouldn't. We need to take this into account for the validation.

hhovsepy: good catch

ferhoyos (approved):
LGTM. Now the validation works with `All` value. Thanks for the quick fix @hhovsepy!

PR #7554, opened 2024-07-22
Implement notification list with AlertGroup PF component

ferhoyos:
### Describe the change

Replace the custom `AlertToast` component with the `AlertGroup` PF component to fix an OSSMC issue and display alert notifications according to PF default standards (demo here: [PatternFly Alert Group Toast with Notification Drawer](https://www.patternfly.org/components/alert/react-demos/alert-group-toast-with-notification-drawer/)).

Note that the alert notification placement defined in PF is not exactly as it was previously in Kiali:

**Previously:**

image

**Now:**

image

If required, the alert notification placement position can be modified with CSS (cc @fkargbo).

### Steps to test the PR

1. Perform any action that triggers an alert notification message (e.g., create/update any Istio object, create traffic policies, enable injection, etc.).
2. Check the alert notification message.

### Automation testing

N/A

### Issue reference

Closes https://github.com/kiali/openshift-servicemesh-plugin/issues/335


hhovsepy (approved):
Verified:
Screenshot from 2024-07-22 15-05-56

ferhoyos:
Confirmed that alert notifications must follow the PF default position so that OSSMC alerts are placed in the same position as OpenShift alerts:

OpenShift console notification:

image

OSSMC notification:

image

jshaughn:
Interesting, I liked our position better.  Also, I wonder what happens when you have a toast come in when one is already displayed, we stacked them vertically.

ferhoyos:
Yes, they are stacked vertically.

image

There is an option in PF to limit the number of notifications shown on the screen (a link to the message center appears to check the rest of the notifications).

image

fkargbo:
@ferhoyos Looks good to me!

PR #7556, opened 2024-07-23
[KEP] [namespace discovery] add the new "overrides" section to the proposal

jmazzitelli:


nrfox (changes requested):
Can you also specify the behavior of `overrides` when `default` is also specified?

jmazzitelli:
> Can you also specify the behavior of `overrides` when `default` is also specified?

Line 131 has this : `Cluster specific overrides will take precedence over the defaults`

Is there something more you want to add to that?

nrfox:
> Line 131 has this : Cluster specific overrides will take precedence over the defaults

Ah I didn't see that line. As long as it's unambiguous what happens when you specify both `defaults` and `overrides` i.e. `defaults` is ignored entirely if you specify any `overrides` and not just the keys that overlap are ignored.

nrfox (approved):

Issue #7515, opened 2024-07-03
Remove or rework istio canary settings


nrfox:
### What do you want to improve?

The canary functionality is incorporated into the mesh page without having to manually specify the current/update settings: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.istio.istio_canary_revision.

### What is the new behavior?

The canary settings are deprecated but you can still use the mesh page to see which namespaces are using each revision without them.

nrfox:
One part of the existing functionality that I think should be kept was a donut chart showing you which namespaces were in current vs. upgrade. This chart was previously on the namespace card but IMO it needs to move to the mesh card so that you see it in the target panel when the mesh is selected. There could technically be more than two controlplanes in the donut chart so it should handle that.

PR #7562, opened 2024-07-25
Canary versions autodiscover

hhovsepy:
### Describe the change

Removed `ExternalServices.Istio.IstioCanaryRevision` configuration usage, now Kiali checks if there are Istiod pods with  `istio.io/rev=canary` and `istio.io/rev=default` labels.

### Steps to test the PR

Install Istio default.
Install Istio 1.22.2.
Install bookinfo.
Install Istio 1.22.3.
In Kiali UI you should be able to canary upgrade downgrade your data plane namespaces.
In Mesh page, the piechart is removed for dataplane.
In Mesh page, the piechart is changed to support multiple revisions.

Screenshot from 2024-08-08 11-39-12
Screenshot from 2024-08-08 11-39-25
Screenshot from 2024-08-08 11-41-00
Screenshot from 2024-08-13 12-43-36
Screenshot from 2024-08-13 12-43-17

### Automation testing

Unit tests are modified to support listing istiod pods.

### Issue reference

https://github.com/kiali/kiali/issues/7515

### DIFF
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/kiali/kiali/config"
+	"github.com/kiali/kiali/istio"
 	"github.com/kiali/kiali/kubernetes"
 	"github.com/kiali/kiali/kubernetes/cache"
 	"github.com/kiali/kiali/models"
@@ -54,8 +55,29 @@ func NewMeshService(
 }
 
 func (in *MeshService) CanaryUpgradeStatus() (*models.CanaryUpgradeStatus, error) {
-	upgrade := in.conf.ExternalServices.Istio.IstioCanaryRevision.Upgrade
-	current := in.conf.ExternalServices.Istio.IstioCanaryRevision.Current
+	upgrade := ""
+	current := ""
+
+	kubeCache, err := in.kialiCache.GetKubeCache(in.homeClusterSAClient.ClusterInfo().Name)
+	if err != nil {
+		return nil, err
+	}
+
+	canaryPods, err := istio.GetHealthyIstiodPods(kubeCache, istio.CanaryRevisionLabel, in.conf.IstioNamespace)
+	if err != nil {
+		return nil, err
+	}
+	if len(canaryPods) > 0 {
+		upgrade = istio.CanaryRevisionLabel
+	}
+	currentPods, err := istio.GetHealthyIstiodPods(kubeCache, istio.DefaultRevisionLabel, in.conf.IstioNamespace)
+	if err != nil {
+		return nil, err
+	}
+	if len(currentPods) > 0 {
+		current = istio.DefaultRevisionLabel
+	}
+
 	migratedNsList := []string{}
 	pendingNsList := []string{}
 
--- a/business/mesh.go
+++ b/business/mesh.go
@@ -2,6 +2,7 @@ package business
 
 import (
 	"fmt"
+	"slices"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/istio"
@@ -88,7 +89,7 @@ func (in *MeshService) CanaryUpgradeStatus() (*models.CanaryUpgradeStatus, error
 
 	// Get migrated and pending namespaces
 	// TODO: Support multi-primary
-	migratedNss, err := in.homeClusterSAClient.GetNamespaces(fmt.Sprintf("istio.io/rev=%s", upgrade))
+	migratedNss, err := in.homeClusterSAClient.GetNamespaces(fmt.Sprintf("%s=%s", in.conf.IstioLabels.InjectionLabelRev, upgrade))
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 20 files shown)


istio/discovery.go
```diff
@@ -36,6 +36,8 @@ const (
 const (
 	// DefaultRevisionLabel is the value for the default revision.
 	DefaultRevisionLabel = "default"
+	// CanaryRevisionLabel is the value for the canary revision.
+	CanaryRevisionLabel = "canary"
```
nrfox, line 40: This is a non-standard label.

frontend/src/pages/Overview/OverviewPage.tsx
```diff
@@ -848,18 +866,13 @@ export class OverviewPageComponent extends React.Component<OverviewProps, State>
 
       if (
         serverConfig.kialiFeatureFlags.istioUpgradeAction &&
-        serverConfig.istioCanaryRevision.upgrade &&
-        serverConfig.istioCanaryRevision.current
+        this.state.canaryUpgradeStatus?.upgradeVersion &&
+        this.state.canaryUpgradeStatus.currentVersion
```
nrfox, line 870: Instead of picking between two options "upgrade" / "downgrade", IMO this should change to pick between any one of the revisions that are managing this cluster. A picker more like:
```
1-21-0
1-22-0
experimental
```

nrfox: You can find the clusters that a controlplane manages by looking at `ControlPlane.ManagedClusters` https://github.com/kiali/kiali/blob/2a0f9deda6b93369363e04ea1c381113c8b9de18/models/mesh.go#L42-L44

nrfox (changes requested):

frontend/src/pages/Overview/OverviewPage.tsx
nrfox, line 1: I don't see the option to choose a particular revision:
Screenshot from 2024-08-09 11-10-11

hhovsepy:
Here one confusion thing is, that in Mesh page, the context is not used while retrieving namespaces per control plane, so in theory users can see more namespaces in mesh page data plane cards compared to Overview page cards.

hhovsepy:
Hi @nrfox , comments approached, ready for another review round.

business/mesh.go
```diff
@@ -54,47 +57,49 @@ func NewMeshService(
...
+	kubeCache, err := in.kialiCache.GetKubeCache(in.conf.KubernetesConfig.ClusterName)
 	if err != nil {
 		return nil, err
 	}
-	for _, ns := range migratedNss {
-		migratedNsList = append(migratedNsList, ns.Name)
-	}
 
-	pendingNss, err := in.homeClusterSAClient.GetNamespaces(fmt.Sprintf("%s=enabled", in.conf.IstioLabels.InjectionLabelName))
+	revisions, err := istio.GetHealthyIstiodRevisions(kubeCache, in.conf.IstioNamespace)
```
nrfox, line 65: Getting the revisions from the healthy pods has a side effect of returning zero revisions when the istiod has zero healthy pods. You can test this by creating a revision and then scaling the istiod replicas down to zero. This is probably an edge case but instead of getting the revisions from the pods, you can also use `in.discovery.Mesh().ControlPlanes`. This will return controlplanes across all clusters so you may need to then filter for a specific cluster if multi-primary isn't supported.

Screenshot from 2024-08-26 15-07-35

business/mesh.go
```diff
@@ -54,47 +57,49 @@ func NewMeshService(
...
+		return &models.CanaryUpgradeStatus{
+			NamespacesPerRevision: namespacesPerRevision,
+		}, nil
 	}
 
+	for _, revision := range revisions {
+		nsList := make(map[string]bool)
+		// Get namespaces for revision
+		// TODO: Support multi-primary
+		nss, err := in.homeClusterSAClient.GetNamespaces(fmt.Sprintf("%s=%s", in.conf.IstioLabels.InjectionLabelRev, revision))
```
nrfox, line 81: This probably has perf implications and I don't think it will work when `cluster_wide_access = false` but we can update this once we add support for multi-primary.

business/mesh_test.go
```diff
@@ -9,78 +9,128 @@ import (
...
+		ObjectMeta: v1.ObjectMeta{Name: "travel-portal", Labels: map[string]string{"istio-injection": "enabled"}},
+	}
 
 	// Create a MeshService and invoke IsMeshConfigured
-	k8sclients := make(map[string]kubernetes.ClientInterface)
-	k8sclients[conf.KubernetesConfig.ClusterName] = k8s
-	layer := business.NewWithBackends(k8sclients, k8sclients, nil, nil)
-	meshSvc := layer.Mesh
+	k8sclients := map[string]kubernetes.ClientInterface{
+		conf.KubernetesConfig.ClusterName: kubetest.NewFakeK8sClient(
```
nrfox, line 37: Thanks for updating these!

nrfox (approved):
A couple of comments but non-blocking. The changes look nice! It even fixes a bug where namespaces that were previously counted as managed but they had no istio injection label.

Screenshot from 2024-08-26 15-00-27

business/mesh.go
hhovsepy: I had in mind to ignore unhealthy Istios from upgrading to, so retrieving healthy istiod pods were on purpose.
Maybe we could return back to this later based on business needs.

hhovsepy:
Thanks for the review @nrfox , I will merge it for now, so we can return back to this while performing multy-cluster updates.

jmazzitelli:
resurrecting - we need to update kiali.io docs to reflect the change made in this PR

PR #7563, opened 2024-07-26
use camelCase for set-based discovery selectors

jmazzitelli:
This is consistent with how Istio and Kubernetes both specify them:
* Istio selectors: https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#k8s-io-apimachinery-pkg-apis-meta-v1-LabelSelector
* Kubernetes selectors: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

This also will make things easier to code up in the server since we need to look at both Kiali AND Istio discovery selectors. It will be nice to have the same yaml parsing for both.

### DIFF
--- a/design/KEPS/namespace-discovery/proposal.md
+++ b/design/KEPS/namespace-discovery/proposal.md
@@ -29,19 +29,19 @@ spec:
   deployment:
     discovery_selectors:
       default:
-      - match_labels:
+      - matchLabels:
           abc: def
           uvw: xyz
-      - match_labels:
+      - matchLabels:
           org: accounting
-        match_expressions:
+        matchExpressions:
         - key: tier
           operator: In
           values: ["production"]
         - key: region
           operator: NotIn
           values: ["east"]
-      - match_expressions:
+      - matchExpressions:
         - key: country
           operator: In
           values: ["us", "ca"]
@@ -72,9 +72,9 @@ Kiali discovery selector:
 ```yaml
 discovery_selectors:
   default:
-  - match_labels:
+  - matchLabels:
       team: backend
-  - match_labels:
+  - matchLabels:
       team: api
 ```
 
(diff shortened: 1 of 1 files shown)


nrfox:
This makes this particular config value different casing from all the rest. Do we want to save this change for when (or if) we change the casing of the rest of the config values?

jmazzitelli:
> This makes this particular config value different casing from all the rest. Do we want to save this change for when (or if) we change the casing of the rest of the config values?

Well, here's my thinking on this. These are set-based selectors that are used in other places (Istio is one, but also think of the match selectors for Deployments in core Kubernetes). So these can be considered settings that are standard Kubernetes (I'm not saying they are standard by the mere fact they are camelCase - but rather these are particular, literal settings `matchLabels` and `matchExpressions` that are common in both Istio and Kubernetes).

The REAL reason why I am proposing this, though, is I think it will make the server-side coding more simple. Because we already have (and will still need) parsing for Istio discovery selectors (see [here](https://github.com/kiali/kiali/blob/v1.87.0/business/namespaces.go#L94-L103)). If we use the standard names (camelCase), I think we can reuse the same yaml parsing code we get from the Istio client API. Otherwise we have to add special handling for the snake_case.

We still will have our own snake_case parent config (`spec.deployment.discovery_selectors`) - it's just we put these under that parent using the camelCase standard names. (This is actually similar to how we handle other kubernetes resource yaml in things like our [spec.deployment.hpa.spec](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.hpa), [spec.deployment.ingress.override_yaml.spec](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.deployment.ingress.override_yaml), and other places (look at the affinity settings as another example).

nrfox (approved):

PR #7570, opened 2024-07-30
Add React import to .tsx files

ferhoyos:
### Describe the change

Due to different webpack bundling options, OSSMC requires that all .tsx files include the React import statement:

```
import * as React from 'react';
```

In the React router upgrade to v6, two files have been renamed from `.ts` to `.tsx`. This PR adds the React import to those files.

**Bonus**:
- To be consistent across Kiali application, I have replaced `import React from 'react'` with the above import statement. This import is valid but not recommended and might not be supported in future React versions.

From https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports:

```
The default imports (i.e. import React from "react") will keep working in React 17, but in the longer term we encourage moving away from them.
```

- Fix lint issues

### Steps to test the PR

All code changes are related to import and linting, there is no impact in Kiali application behaviour.

### Automation testing

N/A


hhovsepy (approved):
lgtm, checked locally

PR #7574, opened 2024-07-30
Always return service account clients when rbac disabled

### LINKED ISSUE #7542: k8s api token not auto refreshing for calls to fetch cacerts
### Describe the bug

With audit logs turned on I see that some Kubernetes API calls use a stale service account token. Especially all calls to fetch the secret cacerts use a token from when Kiali started up.

Looking at the code it seem like cacerts is fetched in https://github.com/kiali/kiali/blob/master/business/istio_certs.go, but I haven't found the root cause of the problem.

PR #7023 should be related.

### Expected Behavior

A current token is used as for other Kubernetes API calls.

### Environment

- **Kiali version:** 1.87.0
- **Istio version:** 1.22.2
- **Kubernetes impl:** EKS
- **Kubernetes version:** 1.30.2

nrfox:
### Describe the change

When the auth strategy is `anonymous` or `openid` with rbac disabled, Kiali uses its own service account client to talk to the kube API. This commit ensures for these auth strategies that the client factory will always use the service account client rather than creating a new client with the kiali service account token.

### Automation testing

Unit tests created.

### Issue reference

Fixes #7542

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -968,6 +968,11 @@ func (conf *Config) GatewayLabel(labelConfig string) []string {
 	return []string{}
 }
 
+func (conf *Config) RBACDisabled() bool {
+	return conf.Auth.Strategy == AuthStrategyAnonymous ||
+		(conf.Auth.Strategy == AuthStrategyOpenId && conf.Auth.OpenId.DisableRBAC)
+}
+
 // Get the global Config
 func Get() (conf *Config) {
 	rwMutex.RLock()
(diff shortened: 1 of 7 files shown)


config/config.go
```diff
@@ -968,6 +968,11 @@ func (conf *Config) GatewayLabel(labelConfig string) []string {
 	return []string{}
 }
 
+func (conf *Config) RBACDisabled() bool {
```
jmazzitelli, line 971: nit - don't our functions that return things like this have names that start with  "Is"? Like:
```suggestion
func (conf *Config) IsRBACDisabled() bool {
```

jmazzitelli (approved):
LGTM, unit tests pass

Issue #7445, opened 2024-06-18
[Ambient] Improve Ambient Graph


josunect:
### What do you want to improve?
Visualize waypoint proxies in the graph

### What is the current behavior?
Waypoint proxies are included in the graph, but some edges are missing, no double edges because of problems highlighting the node graph, and maybe they shouldn't be represented as a node. 
Ref. https://github.com/istio/ztunnel/issues/1128 

### What is the new behavior?
Think the best way to represent these elements in the graph. 
Ref. https://github.com/kiali/kiali/pull/7560

Related PRs:

- Waypoints: 
  - [x] Waypoint node icons: https://github.com/kiali/kiali/pull/7575
  - [x] Double edges for waypoints: https://github.com/kiali/kiali/pull/7593
  - [x] Simplify Ambient Graph: https://github.com/kiali/kiali/pull/7742
  - [x] Summary panel for double edges: https://github.com/kiali/kiali/pull/7719
  - [x] Waypoint nodes should be workload type, not app: https://github.com/kiali/kiali/issues/7702 
  - [x] Combined edges: https://github.com/kiali/kiali/pull/7693
  - [x] https://github.com/kiali/kiali/pull/7766
  - [x] https://github.com/kiali/kiali/issues/7774
  - [x] Combined edges? (for now we will just go with custom bendpoints, see #7780).
- zTunnel: 
  - [x] Visualize ztunnel in the Graph: https://github.com/kiali/kiali/issues/6900

jshaughn:
I'm reclassifying as a research topic, I don't think we need to think about this and what the best approach may be.

josunect:
![image](https://github.com/user-attachments/assets/45bc76db-f11f-4e43-aff7-014b9ba36bd3)

The problem when there is a waypoint proxy seems to be that the waypoint "hides" the source App, because it is going to be waypoint. So all the edges are hidden when we hide the waypoint proxy for the tcp graph: 

![image](https://github.com/user-attachments/assets/7c6d83c0-4563-485c-a914-77e2bfb93601)

https://github.com/kiali/kiali/blob/master/graph/telemetry/istio/istio.go#L363

jshaughn:
@josunect , I've been looking at the ambient bookinfo install and thinking about what we may want to do to improve the ambient graphs.  As a note,  I went back to looking at things without even running the current ambient appender, just to kind of get a baseline.  A couple of thoughts:

[...]

PR #7575, opened 2024-07-31
Add badge icon to waypoints

josunect:
### Describe the change

Add badge icon to waypoints in the Traffic graph: 

image

- Included a legend (Just in Ambient) 

Added to PF graph also: 

image

### Steps to test the PR

Install Kiali and Ambient: 

- `minikube start`
- Install kiali

### Automation testing

e2e tests will be added in a follow up PR

### Issue reference
https://github.com/kiali/kiali/issues/7445


leandroberetta:
Hi @josunect, I did a quick review and works fine.

One thought I have is that we also have the PF graph, which I think needs to be updated too, the next picture is an example of how it shows badges:

image

I know this graph is not the default yet, but we want to promote it at some point this year, for this reason, would be cool to add the same features to both, so we save time later trying to catch up with PF graph differences. What do you think?.

josunect:
> Hi @josunect, I did a quick review and works fine.
> 
> One thought I have is that we also have the PF graph, which I think needs to be updated too, the next picture is an example of how it shows badges:
> 
> image
> 
> I know this graph is not the default yet, but we want to promote it at some point this year, for this reason, would be cool to add the same features to both, so we save time later trying to catch up with PF graph differences. What do you think?.

Thanks for the feedback, @leandroberetta ! Sure, makes sense, I will also add the changes into the PF graph.

leandroberetta (approved):
LGTM

frontend/src/pages/GraphPF/GraphPFElems.tsx
```diff
@@ -254,6 +255,11 @@ export const setNodeLabel = (node: NodeModel, nodeMap: NodeMap, settings: GraphP
     if (data.isGateway?.egressInfo?.hostnames?.length !== undefined) {
       data.labelIcon = <span className={`${badgeMap.get('GW')?.className}`}></span>;
     }
+    if (data.isWaypoint) {
+      data.labelIcon = (
+        <span className={`${badgeMap.get('WA')?.className}`} style={{ marginBottom: '1px', marginLeft: '2px' }}></span>
+      );
+    }
```
jshaughn, line 262: This is OK because we being a Waypoint should be mutually exclusive with being a traffic source. But we may want to update the comment above (in a different PR) to reflect this case.

PR #7576, opened 2024-07-31
[v1.73] Fix missing sidecars in some demo applications

mkralik3:
### Describe the change

Sometimes (on some platforms), not every demo application has a sidecar which causes test failures. 
I suspect that the install script starts patching deployment CRs with sidecar annotation, however, the deployments are not ready yet, or, starts restarting the pods before the annotation is there ( so no sidecar after the restart ).

- So the script must wait until all pods are ready in the namespace (without sidecars) before enabling sidecar infections.
- Added also get all pods after the script is done to identify this type of issue in the future from the log.

Backport is not needed since this is only maistra-related (which is removed in the next branches).

### DIFF
--- a/hack/istio/functions.sh
+++ b/hack/istio/functions.sh
@@ -32,6 +32,7 @@ EOM
   fi
 
   if [ "${ENABLE_INJECTION}" == "true" ] || [ "${AUTO_INJECTION}" == "true" ]; then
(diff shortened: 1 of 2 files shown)


pbajjuri20 (approved):
LGTM..!

Issue #7017, opened 2024-01-10
Investigate using informer transform to save on memory caching costs


nrfox:
### What do you want to improve?

Kiali only uses a subset of fields from the workload (pods, deployments, etc.) and service objects. v1.29 of client-go adds a [WithTransform](https://pkg.go.dev/k8s.io/client-go@v0.29.0/informers#WithTransform) informer factory option that could be used to potentially save on memory costs when caching these kube objects. Anything that is not needed to transform a kube object into kiali workload: https://github.com/kiali/kiali/blob/35e9e088f52106404278b64f7c92130a689b0351/models/workload.go#L120-L157

could potentially be dropped. Same goes for `Service`.

nrfox:
For more context, Kiali gets objects from kube and these are cached in memory but Kiali only uses a small subset of fields on those objects. We could potentially save a lot of resources by trimming down the objects we save in memory with the WithTransform option.

The new pprof endpoints should be used to measure the impact of this at some reasonable scale.

PR #7577, opened 2024-07-31
Added WithTransform for Pods and Services.

hhovsepy:
### Describe the change

Retrieving Pods and Services comes with a lot of additional info which is not used in Kiali.
Added WithTransform to trim those objects and return only used fields.
Removed deprecated 'fetchWorkloads' method and replaced by 'fetchWorkloadsFromCluster'.

### Steps to test the PR

Full Regression test, including health checks of Workloads, Services.

### Automation testing

n/a, existing tests should pass.

Running Kiali perf tests does show only slightly improvements.

### Issue reference

https://github.com/kiali/kiali/issues/7017

### DIFF
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -446,14 +446,7 @@ func (c *kubeCache) createKubernetesInformers(namespace string) informers.Shared
 	opts = append(
 		opts,
 		informers.WithTransform(func(obj interface{}) (interface{}, error) {
-			// similarly is trimmed in istio
-			t, ok := obj.(metav1.ObjectMetaAccessor)
-			if !ok {
-				// shouldn't happen
-				return obj, nil
-			}
-			// ManagedFields is large and we never use it
-			t.GetObjectMeta().SetManagedFields(nil)
+			StripUnusedFields(obj)
 
 			switch obj := obj.(type) {
 			case *core_v1.Pod:
@@ -2299,4 +2292,17 @@ func (c *kubeCache) GetRequestAuthentications(namespace, labelSelector string) (
 		retRequestAuthentications = append(retRequestAuthentications, raCopy)
 	}
 	return retRequestAuthentications, nil
+}
+
+// StripUnusedFields is the transform function for shared informers,
+// it removes unused fields from objects before they are stored in the cache to save memory.
+func StripUnusedFields(obj any) (any, error) {
+	t, ok := obj.(metav1.ObjectMetaAccessor)
+	if !ok {
+		// shouldn't happen
+		return obj, nil
+	}
+	// ManagedFields is large and we never use it
+	t.GetObjectMeta().SetManagedFields(nil)
+	return obj, nil
 }
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -443,6 +443,59 @@ func (c *kubeCache) createKubernetesInformers(namespace string) informers.Shared
 	if namespace != "" {
 		opts = append(opts, informers.WithNamespace(namespace))
 	}
+	opts = append(
+		opts,
+		informers.WithTransform(func(obj interface{}) (interface{}, error) {
+			if pod, ok := obj.(*core_v1.Pod); ok {
+				trimmedPod := &core_v1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:            pod.Name,
+						Namespace:       pod.Namespace,
+						Labels:          pod.Labels,
+						Annotations:     pod.Annotations,
+						OwnerReferences: pod.OwnerReferences,
+					},
+					Spec: core_v1.PodSpec{
+						Containers:         pod.Spec.Containers,
+						InitContainers:     pod.Spec.InitContainers,
+						ServiceAccountName: pod.Spec.ServiceAccountName,
+						Hostname:           pod.Spec.Hostname,
+					},
+					Status: core_v1.PodStatus{
+						Phase:                 pod.Status.Phase,
+						Message:               pod.Status.Message,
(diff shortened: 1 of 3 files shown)


hhovsepy:
Did run pprof before and after changes, both 'cpu' and 'heap'. With the 'heap' one results are very very slightly improvements.
But for the CPU, consumption time improvements are visible, but not a crucial.
CPU before:
```
         0     0%   100%       10ms  1.59%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster
         0     0%   100%       30ms  4.76%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func1
         0     0%   100%       40ms  6.35%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func3
         0     0%   100%       50ms  7.94%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func4
         0     0%   100%       10ms  1.59%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func6
         0     0%   100%       20ms  3.17%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func7
         0     0%   100%       20ms  3.17%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func8
```
CPU after:
```
         0     0%   100%       10ms  1.72%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster
         0     0%   100%       20ms  3.45%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func1
         0     0%   100%       40ms  6.90%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func2
         0     0%   100%       10ms  1.72%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func4
         0     0%   100%       20ms  3.45%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func7
         0     0%   100%       10ms  1.72%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func8
         0     0%   100%       10ms  1.72%  github.com/kiali/kiali/business.(*WorkloadService).fetchWorkloadsFromCluster.func9
```
So the conclusion is, that however there are some improvements on CPU.

kubernetes/cache/kube_cache.go
```diff
@@ -443,6 +443,59 @@ func (c *kubeCache) createKubernetesInformers(namespace string) informers.Shared
 	if namespace != "" {
 		opts = append(opts, informers.WithNamespace(namespace))
 	}
+	opts = append(
+		opts,
+		informers.WithTransform(func(obj interface{}) (interface{}, error) {
+			if pod, ok := obj.(*core_v1.Pod); ok {
```
nrfox, line 449: In the future we can add one of these transform funcs for each of the types that kiali watches e.g. deployments, statefulsets, configmaps etc.

Turning this into a `switch` statement would make it easier to add those:
```
switch o := obj.(type) {
case *core_v1.pod:
  trimmedPod: &core_v1.Pod{
    ObjectMeta: metav1.ObjectMeta{
      name: o.Name
   ...
case *core_v1.Service:
...

```

nrfox: You might be able to use something like this to trim metadata for all objects at the top or bottom of this function if it's common to all objects: https://github.com/openshift/ovn-kubernetes/blob/8f1dcb6a878fcd22b4cf2727abeba7f7c69ba64b/go-controller/pkg/factory/factory.go#L212-L214

nrfox (commented):
Memory savings might depend on how many of these fields are set in your cluster and/or how large your cluster is. If you have a bunch of pods that set `.metadata.managedFields` then you may see higher savings. Either way I think this is a good thing to add even if the improvements are small.

kubernetes/cache/kube_cache.go
hhovsepy: yeah good ideas, have applied

kubernetes/cache/kube_cache.go
```diff
@@ -443,6 +443,69 @@ func (c *kubeCache) createKubernetesInformers(namespace string) informers.Shared
...
+		opts,
+		informers.WithTransform(func(obj interface{}) (interface{}, error) {
+			// similarly is trimmed in istio
+			t, ok := obj.(metav1.ObjectMetaAccessor)
+			if !ok {
+				// shouldn't happen
+				return obj, nil
+			}
+			// ManagedFields is large and we never use it
+			t.GetObjectMeta().SetManagedFields(nil)
```
nrfox, line 456: `t` is not actually used anywhere since down below different objects are returned so I don't think this does anything. If it's not easy to restructure this such that you can do `obj.GetObjectMeta().SetManagedFields(nil)` on the actual object that gets returned then I'd say don't bother with it. It save a few lines of code but it's basically being duplicated by returning a new `ObjectMeta` for the objects inside the case statements anyways.

hhovsepy: I think this is reference type, as similarly is done in Istio code:
https://github.com/istio/istio/blob/master/pkg/kube/util.go#L380
https://github.com/istio/istio/blob/master/pkg/kube/util.go#L355

nrfox: The difference is that they are always returning `obj` at the bottom whereas we create and return a new object inside the case statements, either `trimmedPod` or `trimmedService`.

Can you include some unit tests with the changes? You'll probably need to move this into its own function.

nrfox (approved):

PR #7578, opened 2024-07-31
Add multi-session auth proposal

nrfox:
### Describe the change

Adds a multi-session auth proposal. This is mostly to document the multi-cluster login flow for the `openshift` auth strategy.

### Issue reference

#6360

### DIFF
--- a/design/KEPS/multi-session-auth/proposal.md
+++ b/design/KEPS/multi-session-auth/proposal.md
@@ -1,0 +1,82 @@
+# Kiali Enhancement Proposal
+
+1. [Summary](#summary)
+2. [Motivation](#motivation)
+   1. [Goals](#goals)
+   2. [Non-Goals](#non-goals)
+3. [Solution](#solution)
+   1. [UI Changes](#ui-changes)
+   2. [Limitations](#limitations)
+   3. [Other solutions](#other-solutions)
+4. [Roadmap](#roadmap)
+
+# Summary
+
+Add support for handling multiple login sessions. Having a login session per cluster would allow users without centralized authentication to use Kiali across multiple clusters.
+
+# Motivation
+
+When a user logs into Kiali what they can see and do is restricted, depending on the authentication strategy, to what that they can access inside the Kubernetes cluster. During the authentication process for most authentication stratgies, the Kiali backend is sent the user's actual Kubernetes API token. Kiali then saves the Kubernetes API token inside of an encrypted cookie in the user's browser. The browser subsequently sends this cookie to the backend on further requests and the backend can then decrypt the cookie and use it to make requests to the Kubernetes API server on behalf of the user. This is a single login session.
+
+With some exceptions, the user's API token is only used 1. to get the list of namespaces the user has access to and 2. for all write operations. Almost all read operations in the Kiali backend are done using a different token: Kiali's Service Account token. The Kubernetes cache inside Kiali uses Kiali's Service Account token(s) to list/watch all Kubernetes resources that Kiali consumes (`deployments`, `virtualservices`, `pods`, etc.). The Kiali backend performs access checks by caching the list of namespaces the user has access to and then checking the requested resources' namespace against that list and either filtering the results or failing the request.
+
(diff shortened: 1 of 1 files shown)


design/KEPS/multi-session-auth/proposal.md
```diff
@@ -0,0 +1,82 @@
...
+
+With some exceptions, the user's API token is only used 1. to get the list of namespaces the user has access to and 2. for all write operations. Almost all read operations in the Kiali backend are done using a different token: Kiali's Service Account token. The Kubernetes cache inside Kiali uses Kiali's Service Account token(s) to list/watch all Kubernetes resources that Kiali consumes (`deployments`, `virtualservices`, `pods`, etc.). The Kiali backend performs access checks by caching the list of namespaces the user has access to and then checking the requested resources' namespace against that list and either filtering the results or failing the request.
```
jshaughn, line 21: Trying to see if there are ways we can make the user's life as easy as possible... Could we default to a view-only mode (assuming kiali itself is no set to view-only), such that a user does not need to log into every cluster if he is not performing any write operation? Most users, I think, do not perform writes. In essence, apply a lazy approach to write-auth, maintaining the Kiali SA approach with filtering.

Could we check login of cluster2..N using the same creds as cluster1? In effect, making life easier for users chaining their creds?

design/KEPS/multi-session-auth/proposal.md
```diff
@@ -0,0 +1,82 @@
...
+
+Kiali has five different [authentication strategies](https://kiali.io/docs/configuration/authentication/). For multi-cluster deployments, Kiali initially had support for only the `anonymous` and `openid` authentication strategies. `anonymous` is supported since this simply uses the underlying Kiali Service Accounts for write operations and skips all access checks. `openid` is supported because it allows Kiali to use a single Kubernetes API token across clusters. This kept the login flow the same since users would only need to login once and the Kiali backend could use the same token across clusters. Other authentication strategies like `openshift` and `token` do not support re-using Kubernetes API tokens across clusters. For these strategies, the user must login to each cluster individually and the Kiali backend must save the Kubernetes API token for each cluster.
```
jshaughn, line 23: Would users be limited to the same username on each cluster?

design/KEPS/multi-session-auth/proposal.md
```diff
@@ -0,0 +1,82 @@
...
+Since login is performed per cluster, Kiali will need separate auth callback endpoints per cluster for the `openid` and `openshift` auth strategies. In a typical OAuth login flow, when the `kiali-token` session cookie is not present, the user is redirected to the OAuth server with the callback URL back to the Kiali server embedded inside the redirect URL so that after the user logs into the OAuth server, the OAuth server redirects the user back to the callback URL. When the Kiali server receives a request on the callback endpoint from the OAuth server, the Kiali server needs to know which cluster the user was logging in to. By including the cluster name in the callback URL, the Kiali server can know which cluster the user is logging in to and then use the nonce matching that cluster. For example:
+
+```
+1. Visit Kiali without session cookie
+   user --> Kiali server
+2. Redirected to OAuth server of the "east" cluster.
+   user <-302 to OAuth server with callback- Kiali server
+3. Login to OAuth server
+   user --> OAuth server
+3. OAuth server redirects to callback URL included in URL
```
jshaughn: nit. "3" repeated

design/KEPS/multi-session-auth/proposal.md
nrfox: No it is assumed that the login systems for each cluster are entirely separate from one another. User `foo` can exist on cluster A but not cluster B. Even if user `foo` exists on both clusters, `foo` on A is entirely separate from `foo` on B i.e. they are not the same user.

design/KEPS/multi-session-auth/proposal.md
nrfox: We still need the user to login to each cluster because it's assumed that the login system are entirely separate from one another. Even without WRITE access, in order to check the user's credentials for READ access to the namespace, we need the user's credentials. Otherwise if we don't check the user's creds we need to use the Kiali SA and that is effectively anonymous mode.

Local mode makes this easier since your access to the cluster is already setup through your kubeconfig or if it's not you can `oc login` to the cluster.

nrfox:
@jshaughn circling back around to open PRs. Can this be merged?

Issue #7546, opened 2024-07-17
Enhance namespace accessibility based on Namespace Discovery KEP


jshaughn:
Over the years Kiali's "accessible namespace" approach has confused users (and maintainers), and has struggled to address all of the accessibility needs.  The [Namespace Discovery KEP](https://github.com/kiali/kiali/blob/master/design/KEPS/namespace-discovery/proposal.md) presents a design to hopefully simplify our approach as a natural extension of how upstream Istio does things with Discovery Selectors.

Related: https://github.com/kiali/kiali/issues/7528

jmazzitelli:
I will be working on the operator-side.

The helm-chart-side is probably minimal work since the server helm chart doesn't support any of the stuff the operator does with accessible namespace processing - the server helm chart just needs to remove the accessible namespaces setting and maybe set some default for discovery selectors (though likely it will be the server helm chart leaves them unset letting the server default to the istio discovery selectors).

jmazzitelli:
One thing we need to make sure to put in the release notes - when this feature gets published, you will not be able to upgrade from an older version to a newer one (things won't get cleaned up nicely). So users should delete any Kiali CRs that exist first, before upgrading the operator.

jshaughn:
I've actually been wondering whether this should coincide with a move to Kiali 2.0.  Potentially at the same time we make the PFT graph the default.

jmazzitelli:
Just had a meeting with @nrfox to discuss the progress I'm making on this issue. Some points that came out of that meeting were interesting, so I need to document them here so I don't forget (Nick - feel free to add your own comments here).

* When using Kiali in multi-cluster mode, Kiali MUST have cluster wide access enabled (`spec.deployment.cluster_wide_access` in the Kiali CR):
  * This is something we should document clearly if it is not already.
  * The remote clusters do not have Kiali Roles installed (the operator can't do it - it doesn't have access to the remote clusters; the [remote secret script we provide](https://github.com/kiali/kiali/blob/v1.88.0/hack/istio/multicluster/kiali-prepare-remote-cluster.sh#L107-L118) doesn't do it either, it just creates a ClusterRole).
[...]

PR #7592, opened 2024-08-02
accessible namespaces to discovery selectors refactoring

jmazzitelli:
Implement the new discovery selector support in the server.

part of: kiali/kiali#7546
part of KEP: https://github.com/kiali/kiali/blob/master/design/KEPS/namespace-discovery/proposal.md

server PR: https://github.com/kiali/kiali/pull/7592
operator PR: https://github.com/kiali/kiali-operator/pull/797
helm chart PR: kiali/helm-charts#274
docs PR: https://github.com/kiali/kiali.io/pull/807

This also fixes: https://github.com/kiali/kiali/issues/7655

Some ways to test:
* Unit tests
* Molecule tests
* [manual tests](https://github.com/kiali/kiali/pull/7592#issuecomment-2276388510)
* [multicluster tests](https://github.com/kiali/kiali/pull/7592#issuecomment-2278416500)

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -3,15 +3,10 @@ package business
 import (
 	"context"
 	"fmt"
-	"regexp"
-	"strings"
 	"sync"
 
-	osproject_v1 "github.com/openshift/api/project/v1"
 	core_v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/kiali/kiali/config"
 	"github.com/kiali/kiali/kubernetes"
@@ -23,14 +18,13 @@ import (
 
 // NamespaceService deals with fetching k8sClients namespaces / OpenShift projects and convert to kiali model
 type NamespaceService struct {
-	conf                   *config.Config
-	discovery              meshDiscovery
-	hasProjects            bool
-	homeClusterUserClient  kubernetes.ClientInterface
-	isAccessibleNamespaces map[string]bool
-	kialiCache             cache.KialiCache
-	kialiSAClients         map[string]kubernetes.ClientInterface
-	userClients            map[string]kubernetes.ClientInterface
+	conf                  *config.Config
+	discovery             meshDiscovery
+	hasProjects           bool
+	homeClusterUserClient kubernetes.ClientInterface
+	kialiCache            cache.KialiCache
+	kialiSAClients        map[string]kubernetes.ClientInterface
+	userClients           map[string]kubernetes.ClientInterface
 }
 
 type AccessibleNamespaceError struct {
@@ -62,21 +56,14 @@ func NewNamespaceService(
 		hasProjects = false
 	}
 
-	ans := conf.Deployment.AccessibleNamespaces
-	isAccessibleNamespaces := make(map[string]bool, len(ans))
-	for _, ns := range ans {
-		isAccessibleNamespaces[ns] = true
-	}
-
 	return NamespaceService{
-		conf:                   conf,
-		discovery:              discovery,
-		hasProjects:            hasProjects,
-		homeClusterUserClient:  userClients[homeClusterName],
-		isAccessibleNamespaces: isAccessibleNamespaces,
-		kialiCache:             cache,
-		kialiSAClients:         kialiSAClients,
-		userClients:            userClients,
+		conf:                  conf,
+		discovery:             discovery,
+		hasProjects:           hasProjects,
+		homeClusterUserClient: userClients[homeClusterName],
+		kialiCache:            cache,
+		kialiSAClients:        kialiSAClients,
+		userClients:           userClients,
 	}
 }
 
@@ -87,29 +74,6 @@ func (in *NamespaceService) GetClusterList() []string {
 		clusterList = append(clusterList, cluster)
 	}
 	return clusterList
-}
-
-// Swallows all errors.
-func (in *NamespaceService) getDiscoverySelectors(ctx context.Context) []*meta_v1.LabelSelector {
-	mesh, err := in.discovery.Mesh(ctx)
-	if err != nil {
-		log.Errorf("Will not process discoverySelectors due to a failure to get mesh controlplane information: %v", err)
-		return nil
-	}
-
-	// Add all the discovery selectors together for all control planes.
-	var discoverySelectors []*meta_v1.LabelSelector
-	for _, controlPlane := range mesh.ControlPlanes {
-		discoverySelectors = append(discoverySelectors, controlPlane.Config.DiscoverySelectors...)
-	}
-
-	if len(discoverySelectors) > 0 {
-		log.Tracef("Istio discovery selectors: %+v", discoverySelectors)
-	} else {
-		log.Tracef("No Istio discovery selectors defined.")
-	}
-
-	return discoverySelectors
 }
 
 // Returns a list of the given namespaces / projects
@@ -129,6 +93,7 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 		if !found {
 			clustersToCheck[cluster] = client
 		} else {
+			// namespaces should already be filtered by discovery selectors, no need to do it here
 			namespaces = append(namespaces, cachedNamespaces...)
 		}
 	}
@@ -136,60 +101,6 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 	// Cache hit for all namespaces.
 	if len(clustersToCheck) == 0 {
 		return namespaces, nil
-	}
-
-	discoverySelectors := in.getDiscoverySelectors(ctx)
-	// Need to do per revision discovery selectors.
-
-	// Let's explain the four different filters along with accessible namespaces (aka AN).
-	//
-	// First, we look at AN. AN is either ["**"] or it is not.
-	//
-	// If AN is ["**"], then the entire cluster of namespaces is accessible to Kiali.
-	// In this case, the user can further filter what namespaces this function should return using both includes and excludes.
-	// 1. LabelSelectorInclude is used to obtain an initial set of namespaces, if specified.
-	// 2. Added to that initial list will be the namespaces named in the Include list, if those namespaces actually exist.
-	// 3. If no LabelSelectorInclude or Include list is specified, then all namespaces are in the list.
-	// 4. Remove from that list those namespaces that match LabelSelectorExclude, as well as those namespaces found in the Exclude list.
-	// (Side note: You might ask: Why have an Include list when we already have the AN list?
-	// The difference is if you specify AN (not ["**"]), only those namespaces that exist __at install time__ will get a Role
-	// and hence are accessible to Kiali. The Include list is evaluated at the time this function is called, thus it
-	// allows Kiali to see those namespaces even if they are created after Kiali is installed).
-	//
-	// If AN is not ["**"], then only a subset of namespaces in the cluster is accessible to Kiali.
-	// When installed by the operator, Kiali will be given access to a set of namespaces (as defined in AN) via Roles that
-	// are created by the operator. Those namespaces that Kiali has access to (as defined in AN) will be labeled with the label
-	// selector defined in LabelSelectorInclude (Kiali CR "spec.api.namespaces.label_selector_include").
-	// 1. All of those namespaces are retrieved with the LabelSelectorInclude to obtain a set of namespaces.
-	// 2. Remove from that list those namespaces that match LabelSelectorExclude, as well as those namespaces found in the Exclude list.
-	// The Include option is ignored in this case - you cannot Include more namespaces over and above what AN specifies.
-	// (Side note 1: It probably doesn't make sense to set LabelSelectorExclude and Excludes when AN is not ["**"]. This is because
-	// you already have defined what namespaces you want to give Kiali access to (the AN list itself). However, for consistency,
-	// this function will still use those additional filters to filter out namespaces. So it is possible this function returns
-	// a subset of namespaces that are listed in AN.)
-	// (Side note 2: Notice the difference here between when AN is set to ["**"] and when it is not. When AN is not set to ["**"],
-	// LabelSelectorInclude does not tell the operator which namespaces are included - AN does that. Instead, the operator will
-	// create that label as defined by LabelSelectorInclude on each namespace defined in AN. Thus, after the operator installs
-	// Kiali, the Kiali Server can then use LabelSelectorInclude in this function in order to select all namespaces as defined in AN.
-	// If installed via the server helm chart, none of that is done, and it is up to the user to ensure
-	// LabelSelectorInclude (if defined) selects all namespaces in AN. It is a user-error if they do not configure that correctly.
-	// The server helm chart will not assume they did it correctly. The user therefore normally should not set LabelSelectorInclude
-	// if they also set AN to something not ["**"]. This is one reason why we recommend using the Kiali operator, and why we say
-	// the server helm chart is only provided as a convenience.)
-	// (Side note 3: The control plane namespace is always included via api.namespaces.include and
-	// never excluded via api.namespaces.exclude or api.namespaces.label_selector_exclude.)
-
-	// determine if we are to exclude namespaces by label - if so, set the label name and value for use later
-	labelSelectorExclude := in.conf.API.Namespaces.LabelSelectorExclude
-	var labelSelectorExcludeName string
-	var labelSelectorExcludeValue string
-	if labelSelectorExclude != "" {
-		excludeLabelList := strings.Split(labelSelectorExclude, "=")
-		if len(excludeLabelList) != 2 {
-			return nil, fmt.Errorf("api.namespaces.label_selector_exclude is invalid: %v", labelSelectorExclude)
-		}
-		labelSelectorExcludeName = excludeLabelList[0]
-		labelSelectorExcludeValue = excludeLabelList[1]
 	}
 
 	wg := &sync.WaitGroup{}
@@ -210,6 +121,7 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 				if error != nil {
 					resultsCh <- result{cluster: c, ns: nil, err: error}
 				} else {
+					// getNamespacesByCluster filters namespaces using discovery selectors; we don't have to do it here
 					resultsCh <- result{cluster: c, ns: list, err: nil}
 				}
 			}(cluster)
@@ -232,140 +144,40 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
 		namespaces = append(namespaces, resultCh.ns...)
 	}
 
-	resultns := namespaces
-
-	// Filter out those namespaces that do not match discoverySelectors.
-	// Follow the semantics that Istio follows, which is:
-	//   If there is no discoverySelectors section in the config, skip this entirely.
-	//   If there is an empty discoverySelectors section, that means all namespaces are to be used.
-	//   If there are one or more discoverySelectors specified, the filter namespaces based on what they select.
-	if len(discoverySelectors) > 0 {
-		// 1. convert LabelSelectors to Selectors
-		selectors := make([]labels.Selector, 0)
-		for _, selector := range discoverySelectors {
-			ls, err := meta_v1.LabelSelectorAsSelector(selector)
-			if err != nil {
-				return nil, fmt.Errorf("error initializing discovery selectors filter, invalid discovery selector: %v", err)
-			}
-			selectors = append(selectors, ls)
-		}
-
-		// 2. range over all namespaces to get discovery namespaces, notice each selector result is ORed (as per Istio convention)
-		selectedNamespaces := make([]models.Namespace, 0)
-		for _, ns := range resultns {
-			if ns.Name == in.conf.IstioNamespace {
-				selectedNamespaces = append(selectedNamespaces, ns) // we always want to return the control plane namespace
-			} else {
-				for _, selector := range selectors {
-					if selector.Matches(labels.Set(ns.Labels)) {
-						selectedNamespaces = append(selectedNamespaces, ns)
-						break
-					}
-				}
-			}
-		}
-		namespaces = selectedNamespaces
-		resultns = namespaces
-	}
-
-	// exclude namespaces that are:
-	// 1. to be filtered out via the exclude list
-	// 2. to be filtered out via the label selector
-	// Note that the control plane namespace is never excluded
-	excludes := in.conf.API.Namespaces.Exclude
-	if len(excludes) > 0 || labelSelectorExclude != "" {
-		resultns = []models.Namespace{}
-	NAMESPACES:
-		for _, namespace := range namespaces {
-			if namespace.Name != in.conf.IstioNamespace {
-				if len(excludes) > 0 {
-					for _, excludePattern := range excludes {
-						if match, _ := regexp.MatchString(excludePattern, namespace.Name); match {
-							continue NAMESPACES
-						}
-					}
-				}
-				if labelSelectorExclude != "" {
-					if namespace.Labels[labelSelectorExcludeName] == labelSelectorExcludeValue {
-						continue NAMESPACES
-					}
-				}
-			}
-			resultns = append(resultns, namespace)
-		}
-	}
-
 	// store only the filtered set of namespaces in cache for the token
 	namespacesPerCluster := make(map[string][]models.Namespace)
-	for _, ns := range resultns {
+	for _, ns := range namespaces {
 		namespacesPerCluster[ns.Cluster] = append(namespacesPerCluster[ns.Cluster], ns)
 	}
 	for cluster, ns := range namespacesPerCluster {
 		in.kialiCache.SetNamespaces(in.userClients[cluster].GetToken(), ns)
 	}
 
-	return resultns, nil
+	return namespaces, nil
 }
 
+// getNamespacesByCluster returns the namespaces for the given cluster. The namespaces are filtered such
+// that only those namespaces that match discovery selectors are returned.
 func (in *NamespaceService) getNamespacesByCluster(ctx context.Context, cluster string) ([]models.Namespace, error) {
-	configObject := in.conf
-
-	labelSelectorInclude := configObject.API.Namespaces.LabelSelectorInclude
 
 	var namespaces []models.Namespace
-	_, queryAllNamespaces := in.isAccessibleNamespaces["**"]
+
 	// If we are running in OpenShift, we will use the project names since these are the list of accessible namespaces
 	if in.hasProjects {
-		projects, err := in.userClients[cluster].GetProjects(ctx, labelSelectorInclude)
+		projects, err := in.userClients[cluster].GetProjects(ctx, "")
 		if err != nil {
 			return nil, err
 		}
-		if queryAllNamespaces {
-			namespaces = models.CastProjectCollection(projects, cluster)
-			// add the namespaces explicitly included in the include list.
-			includes := configObject.API.Namespaces.Include
-			if len(includes) > 0 {
-				var allNamespaces []models.Namespace
-				var seedNamespaces []models.Namespace
-
-				if labelSelectorInclude == "" {
-					// we have already retrieved all the namespaces, but we want only those in the Include list
-					allNamespaces = namespaces
-					seedNamespaces = make([]models.Namespace, 0)
-				} else {
-					// we have already got those namespaces that match the LabelSelectorInclude - that is our seed list.
-					// but we need ALL namespaces so we can look for more that match the Include list.
-					allProjectList, err := in.userClients[cluster].GetProjects(ctx, "")
-					if err != nil {
-						return nil, err
-					}
-
-					allNamespaces = models.CastProjectCollection(allProjectList, cluster)
-					seedNamespaces = namespaces
-				}
-				namespaces = in.addIncludedNamespaces(allNamespaces, seedNamespaces)
-			}
-		} else {
-			filteredProjects := make([]osproject_v1.Project, 0)
-			for _, project := range projects {
-				if _, isAccessible := in.isAccessibleNamespaces[project.Name]; isAccessible {
-					filteredProjects = append(filteredProjects, project)
-				}
-			}
-			namespaces = models.CastProjectCollection(filteredProjects, cluster)
-		}
+		namespaces = models.CastProjectCollection(projects, cluster)
 	} else {
-		// if the accessible namespaces define a distinct list of namespaces, use only those.
-		// If accessible namespaces include the special "**" (meaning all namespaces) ask k8sClients for them.
-		// Note that "**" requires cluster role permission to list all namespaces.
-		accessibleNamespaces := configObject.Deployment.AccessibleNamespaces
-		if queryAllNamespaces {
+		// Note that cluster-wide-access mode requires cluster role permission to list all namespaces.
+		if in.conf.Deployment.ClusterWideAccess {
 
-			nss, err := in.userClients[cluster].GetNamespaces(labelSelectorInclude)
+			nss, err := in.userClients[cluster].GetNamespaces("")
 			if err != nil {
 				// Fallback to using the Kiali service account, if needed
 				if errors.IsForbidden(err) {
-					if nss, err = in.getNamespacesUsingKialiSA(cluster, labelSelectorInclude, err); err != nil {
+					if nss, err = in.getNamespacesUsingKialiSA(cluster, "", err); err != nil {
 						return nil, err
 					}
 				} else {
@@ -374,37 +186,13 @@ func (in *NamespaceService) getNamespacesByCluster(ctx context.Context, cluster 
 			}
 
 			namespaces = models.CastNamespaceCollection(nss, cluster)
-
-			// add the namespaces explicitly included in the includes list.
-			includes := configObject.API.Namespaces.Include
-			if len(includes) > 0 {
-				var allNamespaces []models.Namespace
-				var seedNamespaces []models.Namespace
-
-				if labelSelectorInclude == "" {
-					// we have already retrieved all the namespaces, but we want only those in the Include list
-					allNamespaces = namespaces
-					seedNamespaces = make([]models.Namespace, 0)
-				} else {
-					// we have already got those namespaces that match the LabelSelectorInclude - that is our seed list.
-					// but we need ALL namespaces so we can look for more that match the Include list.
-					allK8sNamespaces, errGetNs := in.userClients[cluster].GetNamespaces("")
-					if errGetNs != nil {
-						// Fallback to using the Kiali service account, if needed
-						if errors.IsForbidden(errGetNs) {
-							if allK8sNamespaces, errGetNs = in.getNamespacesUsingKialiSA(cluster, "", errGetNs); errGetNs != nil {
-								return nil, errGetNs
-							}
-						} else {
-							return nil, errGetNs
-						}
-					}
-					allNamespaces = models.CastNamespaceCollection(allK8sNamespaces, cluster)
-					seedNamespaces = namespaces
-				}
-				namespaces = in.addIncludedNamespaces(allNamespaces, seedNamespaces)
-			}
 		} else {
+			// We do not have cluster wide access, so we do not have permission to list namespaces.
+			// Therefore, we assume we can extract the list of accessible namespaces from the discovery selectors configuration.
+			// That list of accessible namespaces will be used as our base list which we then filter with discovery selectors down below.
+			// Note if this is a remote cluster, that remote cluster must have the same namespaces as those in our own local
+			// cluster's accessible namespaces. This is one reason why we suggest enabling CWA for multi-cluster environments.
+			accessibleNamespaces := in.conf.Deployment.AccessibleNamespaces
 			k8sNamespaces := make([]core_v1.Namespace, 0)
 			for _, ans := range accessibleNamespaces {
 				k8sNs, err := in.userClients[cluster].GetNamespace(ans)
@@ -427,6 +215,8 @@ func (in *NamespaceService) getNamespacesByCluster(ctx context.Context, cluster 
 		}
 	}
 
+	namespaces = filterNamespacesWithDiscoverySelectors(namespaces, getDiscoverySelectorsForCluster(cluster, in.conf))
+
 	return namespaces, nil
 }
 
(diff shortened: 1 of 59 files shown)


jmazzitelli:
As of the latest commit, everything is working. Unit tests have been updated and are passing. Integration tests run on github actions are passing. Molecule tests have also been updated and are also passing on minikube:
```
=====================
=== TEST RESULTS: ===
=====================

              accessible-namespaces-test... success [6m 58s]
     affinity-tolerations-resources-test... success [3m 10s]
                cluster-wide-access-test... success [7m 41s]
                      config-values-test... success [2m 26s]
                  default-namespace-test... success [1m 46s]
                            grafana-test... success [3m 11s]
                        header-auth-test... success [1m 37s]
                      instance-name-test... success [2m 11s]
                             jaeger-test... success [2m 24s]
                            metrics-test... success [3m 30s]
                     null-cr-values-test... success [1m 32s]
                only-view-only-mode-test... success [2m 17s]
                             openid-test... success [1m 36s]
                   os-console-links-test... skipped
          ossmconsole-config-values-test... skipped
                 read-certs-secrets-test... success [1m 31s]
                              roles-test... success [4m 47s]
                    rolling-restart-test... success [2m 13s]
                              token-test... success [1m 52s]
```

business/namespaces.go
```diff
@@ -613,6 +300,10 @@ func (in *NamespaceService) GetClusterNamespace(ctx context.Context, namespace s
 		log.Errorf("Unable to refresh cache for cluster [%s]: %s", cluster, err)
 	}
 
+	if !in.isAccessibleNamespace(result) {
+		return nil, &AccessibleNamespaceError{msg: "Namespace [" + namespace + "] is not accessible for Kiali"}
+	}
```
nrfox, line 305: Don't you want to do this access check above where the previous ones were done? `GetClusterNamespaces` might include it in the cache which would then bypass the access check the next time?

kubernetes/cache/cache.go
```diff
@@ -145,7 +146,17 @@ func NewKialiCache(clientFactory kubernetes.ClientFactory, cfg config.Config) (K
...
-		cache, err := NewKubeCache(client, cfg)
+		var accessibleNamespaces []string
+		if !cfg.Deployment.ClusterWideAccess {
+			// Determine what namespaces are to be accessed for the cluster.
+			// Since we do not have cluster wide access, we do not have permission to list namespaces.
+			// Therefore, we assume we can extract the list of accessible namespaces from the discovery selectors configuration.
+			// That list of accessible namespaces will be used as our base list which we then filter with discovery selectors later.
+			// Note if this is a remote cluster, that remote cluster must have the same namespaces as those in our own local
+			// cluster's accessible namespaces. This is one reason why we suggest enabling CWA for multi-cluster environments.
+			accessibleNamespaces = cfg.ExtractAccessibleNamespaceList()
```
nrfox, line 157: Should this be part of config validation?: https://github.com/kiali/kiali/blob/61e8b992bc8be877b53017a61983d866ec9d323a/config/config.go#L1195

That way if the list is malformed the server can fail fast rather than silently drop errors. You could extract once there and just save the list on the config object for other places to consume, similar to `cfg.Deployment.AccessibleNamespaces` except it shouldn't be a config opt that the user can set.

nrfox: Either way, since the list comes from the config object and we're already passing the config object into the cache constructor, do you need to add the `accessibleNamespaces` as a separate argument to the cache constructor or just pull it from the config inside `NewKubeCache`?

business/namespaces.go
```diff
@@ -232,140 +144,46 @@ func (in *NamespaceService) GetNamespaces(ctx context.Context) ([]models.Namespa
...
-
-	labelSelectorInclude := configObject.API.Namespaces.LabelSelectorInclude
 
 	var namespaces []models.Namespace
-	_, queryAllNamespaces := in.isAccessibleNamespaces["**"]
+
+	mesh, ok := in.kialiCache.GetMesh()
+	if !ok {
+		log.Debugf("Cannot obtain mesh details for cluster [%v]; will not be able to filter on Istio discovery selectors if they exist", cluster)
+	}
```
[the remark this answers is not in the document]
jmazzitelli: > We need to fail at some point if there's no discovery selectors configured for kiali and we can't autodiscover the istio discovery selectors.

Only if `cluster_wide_access` == `false`, no? I don't think we want to force/require users to define discovery selectors if Kiali is given cluster-wide access (even [Istio's docs](https://istio.io/latest/docs/reference/config/istio.mesh.v1alpha1/#MeshConfig) say that if Istio discovery selectors aren't defined, all namespaces are processed: `If omitted, Istio will use the default behavior of processing all namespaces in the cluster.`)

business/namespaces.go
```diff
@@ -424,6 +219,7 @@ func (in *NamespaceService) getNamespacesByCluster(ctx context.Context, cluster
 				}
 			}
 			namespaces = models.CastNamespaceCollection(k8sNamespaces, cluster)
+			namespaces = filterNamespacesWithDiscoverySelectors(namespaces, getDiscoverySelectorsForCluster(cluster, in.conf, mesh))
```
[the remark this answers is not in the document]
jmazzitelli: Yeah, I can look to see if that filter call can be moved outside of all the if-else statements and just put one at the end.

business/namespaces.go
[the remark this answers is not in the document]
nrfox: Well since they are used as a form of access control / scoping for Kiali, we don't want Kiali to suddenly start showing namespaces outside of the istio discovery selectors because Kiali can't get the list of discovery selectors from istio for whatever reason.

I'm wondering if we drop the autodiscovery of the istio discovery selectors altogether and default the kiali discovery selectors in the config to process all namespaces?

jmazzitelli:
I want to document this here because this was a question we had as to what the behavior is.

Scenario:
* cluster wide access is false
* I have "foo" being a namespace selected by a discovery selector (e.g. `matchLabels: {"kubernetes.io/metadata.name": "foo"}`)
* Things work fine. But now DELETE the "foo" namespace (e.g. `kubectl delete ns foo`)

Here's what happens (I just tried it).

You will get a warning message in the kiali pod logs:

```
WRN Kiali has an accessible namespace [foo] which is forbidden
```

Things appear to work nicely to the user. If I go to the graph page, I even just see the existing namespaces in the list (I do not see "foo"). When I go to `http://localhost:20001/kiali/api/namespaces` I see only the existing namespaces that match the selectors (I do not see foo in the list).

After a little bit of time (presumably the refresh timer), I start seeing the cache spit out warnings and errors like this:

```
W0806 15:57:11.153978       1 reflector.go:547] pkg/mod/k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: failed to list *v1.DaemonSet: daemonsets.apps is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "daemonsets" in API group "apps" in the namespace "foo"
E0806 15:57:11.154085       1 reflector.go:150] pkg/mod/k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: Failed to watch *v1.DaemonSet: failed to list *v1.DaemonSet: daemonsets.apps is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "daemonsets" in API group "apps" in the namespace "foo"
E0806 15:57:17.393490       1 reflector.go:150] pkg/mod/k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: Failed to watch *v1.ServiceEntry: unknown (get serviceentries.networking.istio.io)
W0806 15:57:18.290348       1 reflector.go:547] pkg/mod/k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: failed to list *v1.ServiceEntry: serviceentries.networking.istio.io is forbidden: User "system:serviceaccount:istio-system:kiali-service-account" cannot list resource "serviceentries" in API group "networking.istio.io" in the namespace "foo"
```

Still, from a UI user perspective, things look to be working fine.

So it isn't catastrophic. But the errors/warnings are annoying due to a now-deleted namespace.

I think rather than expecting the operator to restart the server, we should have the cache self-repair itself... if it gets errors like we see above, it should check "does that namespace exist?" -- if it doesn't, it should just remove it, or flag it or something so it doesn't keep trying to get things from it. Maybe put it on a blacklist and have it retry later but stop checking for awhile.  I dunno, but I think we should try to get the cache to ignore deleted namespaces rather than expect to have to restart the pod just because a namespace was deleted.

business/namespaces.go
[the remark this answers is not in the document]
jmazzitelli: That's a possibility. IIRC, the Istio Discovery Selectors isn't really technically defining what's considered in the mesh anyway.. I thought that was just a way for Istio to limit what it looks for to, say, process auto injection. But it is possible to manually inject sidecars in namespaces that aren't in the discovery selectors and that would work (in other words, DS in Istio was just a feature to increase performance; it wasn't a feature to literally define the mesh namespaces).

Let's think about this some more. This would make Kiali simpler because it would mean both server and operator do not have to worry about auto-discoverying the one (or more) istiod control planes to determine what namespaces to look at. If the user doesn't want Kiali to make all-namespaces visible, they would need to explicitly configure Kiali in that case, rather than rely on Istio DS as a backup. Off the top of my head, I'm thinking that may not be a bad compromise.

nrfox:
> I dunno, but I think we should try to get the cache to ignore deleted namespaces rather than expect to have to restart the pod just because a namespace was deleted.

The Kiali pod is getting restarted anyways because the operator will update the Kiali CR's discovery selector field which will update the Kiali configmap and trigger a restart of the Kiali pod. This only affects `cluster_wide_access = false` and the only benefit would be the window between the namespace getting deleted and the operator restarting the pod. If you have a lot of churn in your namespaces that are selected by Kiali then you'd get a lot of restarts. IMO in that case you should either use `cluster_wide_access = true` or else try and scope down the namespaces selected by the discovery selector when `cluster_wide_access = false`.

jmazzitelli:
> The Kiali pod is getting restarted anyways because the operator will update the Kiali CR's discovery selector field which will update the Kiali configmap and trigger a restart of the Kiali pod.

Not when namespaces are deleted. Operator does nothing about that today.

nrfox:
> Not when namespaces are deleted. Operator does nothing about that today.

Ah ok I missed this. So then the operator wouldn't do anything until the Kiali CR got reconciled again by manually editing it or by some resync period?

In that case we probably want to make that better but I think that's a separate issue from the discovery selector implementation here.

jmazzitelli:
> In that case we probably want to make that better but I think that's a separate issue from the discovery selector implementation here.

Yes, I was thinking the same. Another PR for that effort, probably even a separate github issue to track that.

> So then the operator wouldn't do anything until the Kiali CR got reconciled again by manually editing it or by some resync period?

Yes, exactly. I talk about that very thing in the kiali.io docs PR I wrote for this issue .. see here: https://deploy-preview-807--kiali.netlify.app/docs/configuration/namespace-management/#operator-namespace-watching ... it doesn't mention the "delete namespace" specifically, but talks about how to get the Kiali CR to reconcile manually.

business/namespaces.go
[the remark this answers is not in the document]
jmazzitelli: lol... molecule test failed when i did it at the end... I need to look at this closer... seems trivial to move the filtering at the end but for some reason it failed.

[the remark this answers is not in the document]
jmazzitelli: this  is done. molecule test shows it working

kubernetes/cache/cache.go
jmazzitelli: This is now done. It is done as part of the unmarshalling; it isn't done during validation. There is no modification of the config in the validate function - that strictly looks at the config as it is and gives it a thumbs-up or a thumbs-down. The Unmarshal method is where additional work is done to massage the config, so i added this stuff there.

business/namespaces.go
jmazzitelli: After looking at this deeper, yeah, move it back up. Best to do the isAccessibleNamespace check (which is fast) before bothering to do all that client/cache lookup stuff.  This is done.

business/namespaces.go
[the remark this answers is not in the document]
jmazzitelli: I just reviewed the semantics of Istio discovery selectors, and those do seem to be the way users can control which namespaces Istio considers "in the mesh" ... see blog: https://istio.io/v1.13/blog/2021/discovery-selectors/

Note they make the same distinction we do with our "cluster_wide_access=true" : `Note: discoverySelectors is not a security boundary. Istiod will continue to have access to all namespaces even when you have configured your discoverySelectors.` The difference is we provide `cluster_wide_access=false` for our paranoid users.

Anyway - what should we do here? Right now, if there are NO `deployment.discovery_selectors.default` selectors defined in the Kiali CR, AND we get an error trying to obtain the control plane's discovery selectors (or we just can't get them for whatever reason), we keep going. We should probably fail.

But there are some other things we need to be concerned about if we are to use Istio discovery selectors:
1. What happens if we have multiple control planes (think: canary upgrades of Istio)?
2. What happens if there is an external control plane(s) in remote clusters and it/they have Istio discovery selectors? We don't attempt to look for those at all.

I'm starting to think we need to rip out anything to do with Kiali attempting to find and use Istio discovery selectors and we need to document that we do not use them. If the Istio service mesh has discovery selectors, the Kiali admin needs to configure Kiali with the same discovery selectors. I don't think we can effectively auto-discovery them for all the different use cases (multiple control planes, external control planes, istiod that isn't accessible to Kiali or we get an error trying to get these istio selectors).

jmazzitelli:
All unit tests pass and all molecule tests pass (I confirmed on minikube just now).

All suggestions have been incorporated and the Istio discovery selectors are no longer used.

All PRs have been updated to reflect the changes, including the [docs PR](https://github.com/kiali/kiali.io/pull/807).

KEP is updated in this PR to reflect the change in the usage of Istio discovery selectors.

jmazzitelli:
The unit tests and molecule tests are very good tests for this functionality. But here are some manual test procedures you can run so you can actually see it work, both in the UI and with API responses:

1. Check out the three PRs in the three repos (kiali, kiali-operator, helm-charts) into your dev environment

2. Install minikube, Istio, bookinfo
```sh
hack/k8s-minikube.sh start && \
hack/k8s-minikube.sh istio && \
hack/istio/install-bookinfo-demo.sh -c kubectl -tg
```

3. Build, push, and install dev builds of server and operator:
```sh
make -e CLUSTER_TYPE=minikube build build-ui test cluster-push operator-create kiali-create
```
Wait for the Kiali Server pod to start before continuing (`kubectl get pods -n istio-system -l app=kiali`)

4. You now have Kiali with cluster-wide-access enabled and no discovery selectors. Look at the Kiali graph and see that you can select all namespaces in the graph.
```sh
xdg-open http://$(minikube ip):80/kiali
```

5. Look at the results of the namespaces API - you should get back all of your namespaces:
```sh
curl -s http://$(minikube ip):80/kiali/api/namespaces | jq '.[] | .name'
```

6. Now add discovery selectors so it only selects "bookinfo" (via the istio inject label) and "default" namespaces, but leave Kiali cluster_mode_access set to true (that's its default):
```sh
kubectl edit kiali kiali -n kiali-operator
```
```yaml
  deployment:
    cluster_wide_access: true
    discovery_selectors:
      default:
      - matchLabels:
          istio-injection: enabled
      - matchLabels:
          kubernetes.io/metadata.name: default
```

7. Wait for the Kiali server pod to restart and then do steps 4 and 5 again, but notice you should only see default, bookinfo, and istio-system namespaces (you didn't specify istio-system in the selectors, but we are testing that kiali operator will always give you access to the control plane whether it matches a selector or not):

Check the UI:
```sh
xdg-open http://$(minikube ip):80/kiali
```
Check the namespaces API:
```
$ curl -s http://$(minikube ip):80/kiali/api/namespaces | jq '.[] | .name'
"bookinfo"
"default"
"istio-system"
```

8. You are in cluster wide access mode - there should be no Roles (other than the controlplane one), only the ClusterRole; confirm this:
```
$ kubectl get clusterrole -l app.kubernetes.io/name=kiali 
NAME    CREATED AT
kiali   2024-08-08T17:10:09Z

$ kubectl get role -l app.kubernetes.io/name=kiali --all-namespaces
NAMESPACE      NAME                 CREATED AT
istio-system   kiali-controlplane   2024-08-08T17:10:09Z
```

9. Add a new namespace, and label it for istio injection:
```
kubectl create ns another-ns && kubectl label ns another-ns istio-injection=enabled
```

10. Because we are in cluster-wide access mode, that new namespace can be picked up by the server immediately due to the discovery selector we provided earlier:
```sh
$ curl -s http://$(minikube ip):80/kiali/api/namespaces | jq '.[] | .name'
"another-ns"
"bookinfo"
"default"
"istio-system"
```

11. Edit the Kiali CR again, this time change cluster_wide_access to false; leave the discovery selectors as-is:
```sh
kubectl edit kiali kiali -n kiali-operator
```
```yaml
  deployment:
    cluster_wide_access: false
    discovery_selectors:
      default:
      - matchLabels:
          istio-injection: enabled
      - matchLabels:
          kubernetes.io/metadata.name: default
```

12. Wait for the server to restart and then check that you still only see the another-ns, bookinfo, default, and istio-system namespaces, both in the UI and the namespaces API results:

Check the UI:
```sh
xdg-open http://$(minikube ip):80/kiali
```

Check the namespaces API:
```
$ curl -s http://$(minikube ip):80/kiali/api/namespaces | jq '.[] | .name'
"another-ns"
"bookinfo"
"default"
"istio-system"
```

13. You should no longer have that ClusterRole, but you should now have a Role for each accessible namespace:
```
$ kubectl get clusterrole -l app.kubernetes.io/name=kiali 
No resources found

$ kubectl get role -l app.kubernetes.io/name=kiali --all-namespaces
NAMESPACE      NAME                 CREATED AT
another-ns     kiali                2024-08-08T18:05:29Z
bookinfo       kiali                2024-08-08T18:05:29Z
default        kiali                2024-08-08T18:05:29Z
istio-system   kiali                2024-08-08T18:05:28Z
istio-system   kiali-controlplane   2024-08-08T17:54:30Z
```

jmazzitelli:
Here's another good manual test I just tried - multi-cluster test.

1. Install primary-remote. This one I had to do some manual things to set it up because our hack scripts don't install via the operator. Let's change the hack script to install kiali via operator by applying this diff:
```diff
diff --git a/hack/istio/multicluster/deploy-kiali.sh b/hack/istio/multicluster/deploy-kiali.sh
index 9b9061bd2..93b7b5bc2 100755
--- a/hack/istio/multicluster/deploy-kiali.sh
+++ b/hack/istio/multicluster/deploy-kiali.sh
@@ -111,20 +111,31 @@ deploy_kiali() {
         "--set deployment.image_version=dev"
       )
     else
-      local image_to_tag="quay.io/kiali/kiali:dev"
-      local image_to_push="$(minikube -p ${cluster_name} ip):5000/kiali/kiali:dev"
+      local image_to_tag="quay.io/kiali/kiali-operator:dev"
+      local image_to_push="$(minikube -p ${cluster_name} ip):5000/kiali/kiali-operator:dev"
       echo "Tagging the dev image [${image_to_tag}] -> [${image_to_push}]..."
       ${DORP} tag ${image_to_tag} ${image_to_push}
       echo "Pushing the dev image [${image_to_push}] to the cluster [${cluster_name}]..."
       ${DORP} push --tls-verify=false ${image_to_push}
       helm_args+=(
-        "--set deployment.image_name=localhost:5000/kiali/kiali"
-        "--set deployment.image_version=dev"
+        "--set allowAdHocKialiImage=true"
+        "--set image.repo=localhost:5000/kiali/kiali-operator"
+        "--set image.tag=dev"
+      )
+
+      image_to_tag="quay.io/kiali/kiali:dev"
+      image_to_push="$(minikube -p ${cluster_name} ip):5000/kiali/kiali:dev"
+      echo "Tagging the dev image [${image_to_tag}] -> [${image_to_push}]..."
+      ${DORP} tag ${image_to_tag} ${image_to_push}
+      echo "Pushing the dev image [${image_to_push}] to the cluster [${cluster_name}]..."
+      ${DORP} push --tls-verify=false ${image_to_push}
+      helm_args+=(
+        "--set cr.spec.deployment.image_name=localhost:5000/kiali/kiali"
+        "--set cr.spec.deployment.image_version=dev"
       )
     fi
   fi
 
-
   if [ "${KIALI_CREATE_REMOTE_CLUSTER_SECRETS}" == "true" ]; then
     local openshift_flags=""
     if [ "${KIALI_AUTH_STRATEGY}" == "openshift" ]; then
@@ -166,23 +177,30 @@ deploy_kiali() {
   helm_command='helm upgrade --install
     ${helm_args[@]}
     --namespace ${ISTIO_NAMESPACE}
-    --set deployment.logger.log_level="debug"
-    --set external_services.grafana.url="http://grafana.istio-system:3000"
-    --set external_services.grafana.dashboards[0].name="Istio Mesh Dashboard"
-    --set external_services.tracing.url="http://tracing.istio-system:16685/jaeger"
-    --set health_config.rate[0].kind="service"
-    --set health_config.rate[0].name="y-server"
-    --set health_config.rate[0].namespace="alpha"
-    --set health_config.rate[0].tolerance[0].code="5xx"
-    --set health_config.rate[0].tolerance[0].degraded=2
-    --set health_config.rate[0].tolerance[0].failure=100
-    --set deployment.ingress.enabled=${ingress_enabled}
-    --set deployment.service_type=${service_type}
-    --set server.web_port=${web_port}
-    kiali-server
+    --set cr.create=true
+    --set cr.spec.deployment.logger.log_level="debug"
+    --set cr.spec.deployment.cluster_wide_access=true
+    --set cr.spec.external_services.grafana.url="http://grafana.istio-system:3000"
+    --set cr.spec.external_services.grafana.dashboards[0].name="Istio Mesh Dashboard"
+    --set cr.spec.external_services.tracing.url="http://tracing.istio-system:16685/jaeger"
+    --set cr.spec.health_config.rate[0].kind="service"
+    --set cr.spec.health_config.rate[0].name="y-server"
+    --set cr.spec.health_config.rate[0].namespace="alpha"
+    --set cr.spec.health_config.rate[0].tolerance[0].code="5xx"
+    --set cr.spec.health_config.rate[0].tolerance[0].degraded=2
+    --set cr.spec.health_config.rate[0].tolerance[0].failure=100
+    --set cr.spec.deployment.ingress.enabled=${ingress_enabled}
+    --set cr.spec.deployment.service_type=${service_type}
+    --set cr.spec.server.web_port=${web_port}
+    --set cr.spec.auth.strategy=anonymous
+    kiali-operator
     ${KIALI_SERVER_HELM_CHARTS}'
 
   eval $helm_command
+
+  echo "Waiting for Kiali to be installed by the operator"
+  ${CLIENT_EXE} --context east wait --timeout=300s --for=condition=Successful kiali kiali -n ${ISTIO_NAMESPACE}
+
   if [ "${KIALI_AUTH_STRATEGY}" == "openshift" ]; then
     local kiali_route_url
     kiali_route_url=$(kubectl get route kiali -n "${ISTIO_NAMESPACE}" -o jsonpath='{.spec.host}')
```
Make sure you build everything via `make build build-ui container-buid` and then run the hack script to install: `hack/istio/multicluster/install-primary-remote.sh -mm true -c kubectl -kudi true -kshc ../helm-charts/_output/charts/kiali-operator-*-SNAPSHOT.tgz`

2. Make sure your current context is the east cluster: `kubectl config use-context east`

**The next few steps will edit the Kiali CR - you do this via `kubectl edit kiali kiali -n istio-system` , after you edit it, port-forward to the Kiali server so you can examine the UI: `hack/kiali-port-forward.sh` **

3. Edit the Kiali CR so it contains:
```yaml
spec:
  auth:
    strategy: anonymous
  deployment:
    cluster_wide_access: true
    discovery_selectors:
      default:
      - matchLabels:
          istio-injection: enabled
```

4. Look at the UI and see that the namespace drop down just has "bookinfo" and "istio-system"
5. You can set cluster_wide_access to false if you want - just to see it still works (check that it has a Role in bookinfo to know cluster wide access was turned off -- turn cluster_wide_access back to true and see the Role in bookinfo goes away and that a ClusterRole named kiali comes back)
6. Edit the Kiali CR so it contains override selectors for the west cluster - add default namespace:
```yaml
```yaml
spec:
  auth:
    strategy: anonymous
  deployment:
    cluster_wide_access: true
    discovery_selectors:
      default:
      - matchLabels:
          istio-injection: enabled
      overrides:
        west:
        - matchLabels:
            istio-injection: enabled
        - matchExpressions:
          - key:  kubernetes.io/metadata.name
            operator: In 
            values:
            - "default"
```
7. Look at the UI and see that the namespace drop down now has "default" as well as "bookinfo" and "istio-system"
8. Look at the namespaces API and see that you get "default" namespace for "west" cluster:
```sh
$ curl -s http://localhost:20001/kiali/api/namespaces | jq
```
results in:
```json
[
  {
    "name": "bookinfo",
    "cluster": "east",
    "isAmbient": false,
    "labels": {
      "istio-injection": "enabled",
      "kubernetes.io/metadata.name": "bookinfo"
    },
    "annotations": null
  },
  {
    "name": "istio-system",
    "cluster": "east",
    "isAmbient": false,
    "labels": {
      "kubernetes.io/metadata.name": "istio-system",
      "topology.istio.io/network": "network-east"
    },
    "annotations": null
  },
  {
    "name": "bookinfo",
    "cluster": "west",
    "isAmbient": false,
    "labels": {
      "istio-injection": "enabled",
      "kubernetes.io/metadata.name": "bookinfo"
    },
    "annotations": null
  },
  {
    "name": "default",
    "cluster": "west",
    "isAmbient": false,
    "labels": {
      "kubernetes.io/metadata.name": "default"
    },
    "annotations": null
  },
  {
    "name": "istio-system",
    "cluster": "west",
    "isAmbient": false,
    "labels": {
      "kubernetes.io/metadata.name": "istio-system",
      "topology.istio.io/network": "network-west"
    },
    "annotations": {
      "topology.istio.io/controlPlaneClusters": "east"
    }
  }
]
```

jmazzitelli:
I ran the molecule tests one more time on OpenShift ... all pass:

```
=====================
=== TEST RESULTS: ===
=====================

              accessible-namespaces-test... success [7m 37s]
     affinity-tolerations-resources-test... success [3m 38s]
                cluster-wide-access-test... success [8m 30s]
                      config-values-test... success [2m 34s]
                  default-namespace-test... success [1m 47s]
                            grafana-test... success [3m 20s]
                        header-auth-test... skipped
                      instance-name-test... success [2m 14s]
                             jaeger-test... success [2m 33s]
                            metrics-test... success [3m 40s]
                     null-cr-values-test... success [1m 39s]
                only-view-only-mode-test... success [2m 30s]
                             openid-test... skipped
                   os-console-links-test... success [4m 35s]
          ossmconsole-config-values-test... success [3m 31s]
                 read-certs-secrets-test... success [1m 40s]
                              roles-test... success [5m 14s]
                    rolling-restart-test... success [2m 25s]
                              token-test... success [1m 57s]
```

jmazzitelli:
Ran molecule tests on minikube - same result - all successfully pass.

Due to all unit, molecule, and manual tests passing, this issue's PRs should be considered ready to go. No additional changes are expected in this PR (or the other PRs related to this issue) except at most a rebase on master branch if conflicts need to be addressed before merging

jmazzitelli:
I re-ran all molecule tests on minikube after rebasing on master and fixing conflicts. everything still passes.

tests/testutils/prepare_config.go
nrfox, line 1: This is very much a nit but anything `utils` usually becomes a dumping ground where it's hard to know what's in there or what the purpose of the package is unless it's scoped somehow like `kubetest` or `net/http/httptest`. Maybe this can go in `config/configtest`? I can think of at least one other helper method that should go in there. Having a `configtest.Set` that resets the global variable after the test is over for all the places that rely on the global `config.Config` variable. There's at least one implementation of this in the biz package but it should probably be in something like `configtest`.

Package names are also generally lower case e.g. `testutils` --> `testutil`: https://google.github.io/styleguide/go/decisions.html#package-names.

kubernetes/kubetest/mock.go
```diff
@@ -279,9 +279,19 @@ func FakeIstioAmbientAnnotations() map[string]string {
 }
 
 func FakeNamespace(name string) *core_v1.Namespace {
+	return FakeNamespaceWithLabels(name, nil)
+}
+
+func FakeNamespaceWithLabels(name string, labels map[string]string) *core_v1.Namespace {
```
nrfox, line 285: nit: you don't need to remove this but this can often end up with having many different methods for setting all the fields in a namespace e.g. `FakeNamespace(name string, labels, annotations, hasInjection, revisionLabel... etc.)` where it's simple enough to just do this where it's needed:
```
ns := FakeNamespace("")
ns.Labels = map[string]string{...}
ns.Annotations
```
or define the whole struct
```
corev1.Namespace{{ObjectMeta: ...}}
```

kubernetes/cache/kube_cache.go
```diff
@@ -298,8 +331,17 @@ func (c *kubeCache) refresh(namespace string) error {
 		namespace = ""
 	}
 
-	c.stop(namespace)
-	return c.startInformers(namespace)
+	// We need to be thread-safe here so refreshing doesn't deadlock after a deleted namespace is re-created.
+	// Locking here ensures concurrent calls to getCacheLister doesn't deadlock. There tends to be a flurry
+	// of concurrent calls to getCacheLister and if at that time the namespace cache needs to be refreshed, we
+	// only want to create informers once here. This might not always create informers one time, but it helps limit
+	// the number of times informers are created and then immediately thrown away and created again.
```
nrfox, line 338: So we could end up starting multiple informers for the same namespace/resource?

nrfox: Can we put a `refreshLock` on `getCacheLister` that does a 
```
c.refreshLock.RLock()
defer c.refreshLock.RUnlock()
```
and inside `refresh` we do:
```
c.refreshLock.Lock()
defer c.refreshLock.Unlock()
```
I something like that should help ensure that we don't try to read from the listers map while they are refreshing since callers won't be able to get a readlock while a refresh is happening.

kubernetes/cache/cache.go
```diff
@@ -145,7 +147,7 @@ func NewKialiCache(clientFactory kubernetes.ClientFactory, cfg config.Config) (K
 	}
 
 	for cluster, client := range clientFactory.GetSAClients() {
-		cache, err := NewKubeCache(client, cfg)
+		cache, err := NewKubeCache(client, cfg, kialiCacheImpl.deleteNamespace)
```
nrfox, line 150: Can we pass `nil` when it's a cluster scoped cache? I really don't want to call this error handler or have any side effects when there's a cluster scoped cache.

design/KEPS/namespace-discovery/proposal.md
nrfox, line 1: We should also document the behavior for the external control plane scenario.

tests/testutils/prepare_config.go
```diff
@@ -0,0 +1,23 @@
...
+
+// GetConfigFromYaml provides a way for tests to write config as yaml. This gives a chance for tests to exercise the Unmarshal functionality.
+// This also provides an easy way for tests to set up the config object correctly. This comes into play when, for example,
+// you want cluster_wide_access set to false with discovery selectors set - the unmarshalling code this function utilizes will prepare the
+// accessible namespaces thus freeing the test authors from having to remember to do that explicitly in the test code.
+// This will provide easy-to-read failure log messages if the given yaml is malformed in some way (for easier debugging of test code).
+func GetConfigFromYaml(t *testing.T, yaml string) *config.Config {
+	if cfg, err := config.Unmarshal(yaml); err != nil {
+		// mark the test as a failure; we'll keep going so the log will show what test failed
+		t.Helper()
```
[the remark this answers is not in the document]
jmazzitelli: So you mean move t.Helper below line 14?

jmazzitelli, line 17: Do you mean this?
```suggestion
func GetConfigFromYaml(t *testing.T, yaml string) *config.Config {
	t.Helper()
	if cfg, err := config.Unmarshal(yaml); err != nil {
		// mark the test as a failure; we'll keep going so the log will show what test failed
```

tests/testutils/prepare_config.go
jmazzitelli: @nrfox write up a github issue on this - "refactor test utils" - we can do that separately in another PR after this monstrosity is merged ;)

kubernetes/kubetest/mock.go
jmazzitelli: I added this specifically to make things still one line of code AND avoid devs having to remember to always add the metadata.name label (which all k8s namespaces have and which our discovery selector stuff utilize - I suspect devs writing code in the future won't even know that sometimes namespaces need to have that label so they won't add it).

The issue here is we had a bunch of corev1.Namespaces all over the place, and we always want this `kubernetes.io/metadata.name` label in there - yes, I could have added that label in there without a new Fake function, but I didn't want people to have to remember to do that every time in the future. So FakeNamespace() is the way to do it..

BUT! Then you have tests that actually need additional labels. But we don't want to (a) lose the metadata.name and (b) require the dev to remember to add it. Hence, FakeNamespaceWithLabels.

So this won't grow with additional parameters - because the only reason "WithLabels" exists is because we want to support adding custom labels but NOT override the default label we always want in there. Because we don't have, say, default annotations or default whatever in our fake namespaces, the developer is free to do what you say with the returned namespace to add that stuff and no need to add parameters to the function.

ns := FakeNamespaceWithLabels(...labels in addition to the metadata.name that I get for free...)
ns.Annotations ...
ns...whatever

So, again, this is only to help the devs always get that "metadata.name" label in namespaces that are created - the WithLabels is merely to avoid the case where the dev needs a custom label, too. And you can do this with one line of code with the new functions.

kubernetes/cache/kube_cache.go
jmazzitelli: > So we could end up starting multiple informers for the same namespace/resource?

That's me doing a CYA :) I have not see multiple informers get created. The do-once stuff appears to be working as intended. But, because I was not 100% confident in my understanding of the thread-safety stuff surrounding this code, I made that statement.

I actually DID do what you suggested with the refresh lock - and was getting deadlocks consistently (hence why my confidence isn't 100% - I couldn't figure out why that deadlock was happening - I think there are some simultenous go func executions happening that I could not find). That's why there is this OnceWrapper - I went to the design to avoid the deadlocks BUT still ensure we only run it one time.

kubernetes/cache/cache.go
jmazzitelli: We know we are cluster scoped at this point in the code right? I guess "cfg.Deployment.ClusterWideAccess" ??

design/KEPS/namespace-discovery/proposal.md
jmazzitelli: I  don't even know what would happen :)

kubernetes/cache/cache.go
jmazzitelli: This change is coming when I push my next commit:
```diff
diff --git a/kubernetes/cache/cache.go b/kubernetes/cache/cache.go
index eb5edf1d7..e306546de 100644
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -147,7 +147,12 @@ func NewKialiCache(clientFactory kubernetes.ClientFactory, cfg config.Config) (K
        }
 
        for cluster, client := range clientFactory.GetSAClients() {
-               cache, err := NewKubeCache(client, cfg, kialiCacheImpl.deleteNamespace)
+               // we only need our deleteNamespace function called when an error occurs in a namespace-scoped cache
+               var errHandler ErrorHandler
+               if !cfg.Deployment.ClusterWideAccess {
+                       errHandler = kialiCacheImpl.deleteNamespace
+               }
+               cache, err := NewKubeCache(client, cfg, errHandler)
```

tests/testutils/prepare_config.go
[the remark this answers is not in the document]
jmazzitelli: This change is coming when I push my next commit:
```diff
--- a/tests/testutils/prepare_config.go
+++ b/tests/testutils/prepare_config.go
@@ -12,9 +12,9 @@ import (
 // accessible namespaces thus freeing the test authors from having to remember to do that explicitly in the test code.
 // This will provide easy-to-read failure log messages if the given yaml is malformed in some way (for easier debugging of test code).
 func GetConfigFromYaml(t *testing.T, yaml string) *config.Config {
+       t.Helper()
        if cfg, err := config.Unmarshal(yaml); err != nil {
                // mark the test as a failure; we'll keep going so the log will show what test failed
-               t.Helper()
                t.Fatalf("Test provided invalid config YAML.\nERROR: %v\nYAML:%v", err, yaml)
```

config/config.go
```diff
@@ -1275,3 +1292,57 @@ func validateSigningKey(signingKey string, authStrategy string) error {
...
+			if len(expr.Values) == 1 {
+				if expr.Key == "kubernetes.io/metadata.name" && expr.Operator == metav1.LabelSelectorOpIn {
+					namespaceNames = append(namespaceNames, expr.Values[0])
+				} else {
+					errs = append(errs, fmt.Sprintf("invalid accessible namespace discovery selectors: matchExpressions selector must match the label named kubernetes.io/metadata.name using the IN operator: %v", selector))
+				}
+			} else {
+				errs = append(errs, fmt.Sprintf("invalid accessible namespace discovery selectors: matchExpressions selector must match one and only one value for label named kubernetes.io/metadata.name using the IN operator: %v", selector))
+			}
+		}
```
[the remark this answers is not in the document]
jmazzitelli: done - unit test shows it working https://github.com/kiali/kiali/pull/7592/commits/cdce937d633615f796b5a22f855c3dbe31c986e1

config/config.go
```diff
@@ -423,17 +411,58 @@ type OpenIdConfig struct {
...
+	Namespace            string                   `yaml:"namespace,omitempty"` // Kiali deployment namespace
+	ViewOnlyMode         bool                     `yaml:"view_only_mode,omitempty"`
 	// RemoteSecretPath is used to identify the remote cluster Kiali will connect to as its "local cluster".
 	// This is to support installing Kiali in the control plane, but observing only the data plane in the remote cluster.
 	// Experimental feature. See: https://github.com/kiali/kiali/issues/3002
 	RemoteSecretPath string `yaml:"remote_secret_path,omitempty"`
 }
 
+// we need to play games with a custom unmarshaller/marshaller for metav1.LabelSelector because it has no yaml struct tags so
+// it is not processing it the way we want by default (it isn't using camelCase; the fields are lowercase - e.g. matchlabels/matchexpressions)
```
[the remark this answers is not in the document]
jmazzitelli: you can write up a github enhancement request so we don't lose that idea.

tests/testutils/prepare_config.go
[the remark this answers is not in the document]
jmazzitelli: done.

kubernetes/cache/cache.go
jmazzitelli: done.

tests/testutils/prepare_config.go
jmazzitelli: done

kubernetes/cache/kube_cache.go
jmazzitelli: I'm going to revisit the refresh lock thing, just to see if I see that deadlock again. This is the last thing left for me to look at - all other suggestions have been implemented in the latest commit.

jmazzitelli: OK, I see the deadlocks like I saw before. Currently using this:
```diff
diff --git a/kubernetes/cache/kube_cache.go b/kubernetes/cache/kube_cache.go
index 5bf59ef93..326aaf5cc 100644
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -177,7 +177,8 @@ type ErrorHandler func(kc *kubeCache, namespace string, err error)
 
 // kubeCache is a local cache of kube objects. Manages informers and listers.
 type kubeCache struct {
-       cacheLock sync.RWMutex
+       cacheLock   sync.RWMutex
+       refreshLock sync.RWMutex
        // refreshOnce enforces thread safety around the re-starting of informers
        refreshOnce        *OnceWrapper
        cfg                config.Config
@@ -318,6 +319,10 @@ func (c *kubeCache) stop(namespace string) {
 // Refresh will recreate the necessary cache. If the cache is cluster-scoped the "namespace" argument
 // is ignored and the whole cache is recreated, otherwise only the namespace-specific cache is updated.
 func (c *kubeCache) Refresh(namespace string) {
+       // COMMENT OUT BELOW AND IT STILL DEADLOCKS
+       c.refreshLock.Lock()
+       defer c.refreshLock.Unlock()
+
        c.cacheLock.Lock()
        defer c.cacheLock.Unlock()
 
@@ -327,6 +332,9 @@ func (c *kubeCache) Refresh(namespace string) {
 }
 
 func (c *kubeCache) refresh(namespace string) error {
+       c.refreshLock.Lock()
+       defer c.refreshLock.Unlock()
+
        if c.clusterScoped {
                namespace = ""
        }
@@ -612,6 +620,9 @@ func (c *kubeCache) getCacheLister(namespace string) *cacheLister {
                return c.clusterCacheLister
        }
 
+       c.refreshLock.RLock()
+       defer c.refreshLock.RUnlock()
+
        lister, ok := c.nsCacheLister[namespace]
        if !ok {
 ```

Comment (jmazzitelli) [reply kubernetes/cache/kube_cache.go:338]: This is the unit test that deadlocks after making that change: https://github.com/kiali/kiali/blob/v1.89.1/kubernetes/cache/kube_cache_test.go#L112

Comment (jmazzitelli) [reply kubernetes/cache/kube_cache.go:338]: Tried locking directly in the do-once function, still deadlocks:
```diff
diff --git a/kubernetes/cache/kube_cache.go b/kubernetes/cache/kube_cache.go
index 5bf59ef93..8d2bbcfc1 100644
--- a/kubernetes/cache/kube_cache.go
+++ b/kubernetes/cache/kube_cache.go
@@ -177,7 +177,8 @@ type ErrorHandler func(kc *kubeCache, namespace string, err error)
 
 // kubeCache is a local cache of kube objects. Manages informers and listers.
 type kubeCache struct {
-       cacheLock sync.RWMutex
+       cacheLock   sync.RWMutex
+       refreshLock sync.RWMutex
        // refreshOnce enforces thread safety around the re-starting of informers
        refreshOnce        *OnceWrapper
        cfg                config.Config
@@ -338,6 +339,8 @@ func (c *kubeCache) refresh(namespace string) error {
        // the number of times informers are created and then immediately thrown away and created again.
        defer c.refreshOnce.Reset()
        return c.refreshOnce.Do(func() error {
+               c.refreshLock.Lock()
+               defer c.refreshLock.Unlock()
                log.Debugf("Restarting cache informers for namespace [%v]", namespace)
                c.stop(namespace)
                return c.startInformers(namespace)
@@ -612,6 +615,9 @@ func (c *kubeCache) getCacheLister(namespace string) *cacheLister {
                return c.clusterCacheLister
        }
 
+       c.refreshLock.RLock()
+       defer c.refreshLock.RUnlock()
+
```

jmazzitelli: I think it's happening here: https://github.com/kiali/kiali/blob/v1.89.1/kubernetes/cache/kube_cache.go#L337

jmazzitelli: OK, if I release the lock after calling c.stop but before c.startInformers inside the "once.do" function, the deadlock goes away, but I don't think that is thread safe at that point.

```diff
        return c.refreshOnce.Do(func() error {
+               c.refreshLock.Lock()
                log.Debugf("Restarting cache informers for namespace [%v]", namespace)
                c.stop(namespace)
+               c.refreshLock.Unlock()
                return c.startInformers(namespace)
```

I see something odd about that startInformers function anyway. I need to stare at this some more.

jmazzitelli: > I see something odd about that startInformers function anyway. I need to stare at this some more.

I will briefly explain here...

1. [startInformers](https://github.com/kiali/kiali/blob/v1.89.1/kubernetes/cache/kube_cache.go#L312-L338) runs go funcs to create the three sets of informers (kubernetes, istio, gateways)
2. [create istio informers](https://github.com/kiali/kiali/blob/v1.89.1/kubernetes/cache/kube_cache.go#L350-L357) and [create gateway informers](https://github.com/kiali/kiali/blob/v1.89.1/kubernetes/cache/kube_cache.go#L403-L410) both call [getCacheLister](https://github.com/kiali/kiali/blob/v1.89.1/kubernetes/cache/kube_cache.go#L535-L540) to get the listers they will work on
3. BUT the [createKubernetesInformers](https://github.com/kiali/kiali/blob/v1.89.1/kubernetes/cache/kube_cache.go#L441-L530) is the one that actually creates the listers (the ones that getCacheListers returns).

So how can the code expect (1) to work (i.e. simultaneously run the three create informers functions)  when one of them creates the listers the other 2 retrieve (??). That doesn't seem right. This might be working just by luck... the createKubernetesInformers needs to be guaranteed to run first; otherwise, the createIstio and createGateway functions calling getCacheLister aren't guaranteed to get listers back -- they are in a race with createKubernetes who is creating them.

jmazzitelli: With the OnceWrapper in place, I don't think we have to worry about this refreshLock. We get thread-safety with OnceWrapper - I've seen it working properly and I've been checking this on purpose.

nrfox (approved):

PR #7593, opened 2024-08-05
[ambient] waypoint arrows and PF fix

josunect:
### Describe the change

This PR: 
- Adds a double arrow in the same edge for the waypoint proxies. This aims to simplify the number of edges in the graph
 
image

- Fixes patternfly graph to avoid infinite loops on hover: 
image

This error: 
image

The specific error cannot be reproduced in master, just when the waypoint appender does not delete the double edges.

### Steps to test the PR

Install Kiali and Ambient: 

- `minikube start`
- Install kiali

### Automation testing

New tests will be added in a separate PR.

### Issue reference

Ref. #7445 

TODO: 

This is just a proposal, but in order to be merged, it would need some improvements like: 
- Traffic animation
- Traffic rate: (It is needed to add both send/received?) 
- Summary panel

### DIFF
--- a/graph/telemetry/istio/appender/ambient.go
+++ b/graph/telemetry/istio/appender/ambient.go
@@ -83,15 +83,13 @@ func (a AmbientAppender) handleWaypoints(trafficMap graph.TrafficMap, globalInfo
 				n.Edges = sliceutil.Filter(n.Edges, func(edge *graph.Edge) bool {
 					return !waypointsNodes[edge.Dest.ID]
 				})
+				continue
 			}
 
 			// Find duplicates
 			for _, edge := range n.Edges {
-				// If not show the waypoint, mark the nodes
-				if a.ShowWaypoints {
-					if waypointsNodes[edge.Dest.ID] {
-						edge.Metadata[graph.Display] = "reverse"
-					}
+				if waypointsNodes[edge.Dest.ID] {
+					edge.Metadata[graph.Display] = "reverse"
 				}
 			}
 		}
--- a/graph/telemetry/istio/appender/ambient.go
+++ b/graph/telemetry/istio/appender/ambient.go
@@ -90,7 +90,7 @@ func (a AmbientAppender) handleWaypoints(trafficMap graph.TrafficMap, globalInfo
 				// If not show the waypoint, mark the nodes
 				if a.ShowWaypoints {
 					if waypointsNodes[edge.Dest.ID] {
-						edge.Metadata[graph.Direction] = "reverse"
+						edge.Metadata[graph.Display] = "reverse"
 					}
 				}
 			}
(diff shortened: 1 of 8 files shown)


graph/telemetry/istio/appender/ambient.go
```diff
@@ -50,7 +50,7 @@ func (a AmbientAppender) handleWaypoints(trafficMap graph.TrafficMap, globalInfo
 	// Fetch the waypoint workloads
 	waypoints := globalInfo.Business.Workload.GetWaypoints(context.Background())
 
-	deletedWaypoints := make(map[string]bool)
+	waypointsNodes := make(map[string]bool)
```
jshaughn, line 53:
```suggestion
	waypointNodes := make(map[string]bool)
```

Just removing the double-plural.

graph/telemetry/istio/appender/ambient.go
```diff
@@ -62,36 +62,42 @@ func (a AmbientAppender) handleWaypoints(trafficMap graph.TrafficMap, globalInfo
...
+
+			// Find duplicates
+			for _, edge := range n.Edges {
+				// If not show the waypoint, mark the nodes
+				if a.ShowWaypoints {
+					if waypointsNodes[edge.Dest.ID] {
+						edge.Metadata[graph.Display] = "reverse"
+					}
+				}
+			}
```
jshaughn, line 96:
```suggestion
			if !a.ShowWaypoints {
				n.Edges = sliceutil.Filter(n.Edges, func(edge *graph.Edge) bool {
					return !waypointsNodes[edge.Dest.ID]
				})
				continue
			}

			// Find duplicates
			for _, edge := range n.Edges {
				if waypointsNodes[edge.Dest.ID] {
					edge.Metadata[graph.Display] = "reverse"
				}
			}
```

I think this can be simplified, we already know `a.showWaypoints` is true.

jshaughn (changes requested):
Finally got a chance to look at this, @josunect, sorry for the delay.  Your idea of using the bidirectional edge definitely simplifies the current waypoint visualization in the graph.  As you mentioned in the comments, the PR does still have work to do as it does not manipulate rates and doesn't do anything for the side panel.

But after just a couple of minor changes, we can probably merge it as an immediate simplification of the graph.

Despite the fact that this makes the graph more clear, I'm still not really happy with waypoint visualization, and I'm starting to think a more drastic change may be necessary.  I have a few ideas that I will share with you outside of this PR.

josunect:
> Finally got a chance to look at this, @josunect, sorry for the delay. Your idea of using the bidirectional edge definitely simplifies the current waypoint visualization in the graph. As you mentioned in the comments, the PR does still have work to do as it does not manipulate rates and doesn't do anything for the side panel.
> 
> But after just a couple of minor changes, we can probably merge it as an immediate simplification of the graph.
> 
> Despite the fact that this makes the graph more clear, I'm still not really happy with waypoint visualization, and I'm starting to think a more drastic change may be necessary. I have a few ideas that I will share with you outside of this PR.

Thanks for the review, @jshaughn ! Changes addressed. 
I'm working on the summary panel now (But we can merge and continue in another PR). 
I was working on this Draft PR https://github.com/kiali/kiali/pull/7693 regarding the Ambient graph, but we can review separately, as you mentioned.

jshaughn (approved):
I think this is fine for now, and should be backported to 1.89.  Things may change in 2.0, but we may very well still incorporate bidirectional edges when focusing on a waypoint's node or detail graph.

PR #7595, opened 2024-08-06
Simplify Ambient graph

josunect:
### Describe the change

For Ambient Graph, with the Ambient graph appender, when there are 2 edges from/to the same node, mark the tcp edge to be hidden and the http to show both traffic data (tcp and http). 

The motivation is:
- To simplify the graph. 
- Don't lose information. We know there are tcp+http Telemetry. This means it is an Ambient graph.

This is a graph in a namespace with sidecars: 
image

The actual Ambient graph: 
image

Te graph simplified: 
image

This is just a PoC, if we decided to go with this approach there should be some TODOs:

- Show in the summary panel the information about TCP/HTTP. 
- Work in the Display options like response time, protocol on hover, etc..

### Cytoscape graph 

This would be an example about how it should be displayed with the Cytoscape graph: 
image

It could use the properties: 
- line-outline-width
- line-outline-color

- [x] It needs a Cytoscape typescript PR (Done in 3.21.7) to include these properties and a Cytoscape update. 

### Steps to test the PR

- `minikube start`
- Install kiali

### Automation testing
e2e tests will be added in an additional PR

### Issue reference

ref. #7445 

TODO: 

This is just a proposal, but in order to be merged, it would need some improvements like: 
- Traffic animation
- Traffic rate: (It is needed to add both send/received?) 
- Summary panel


josunect:
@fkargbo

fkargbo:
Looks good to me. The simplified graph is easier to follow. Thanks!

fkargbo:
Hi @josunect  I've attached an image showing Edge connection statuses and connection types as we discussed last week. 
The idea is to show our various connection statuses and the types of connections.

I've created an idea for the "Tunnel" connection we talked about, which also shows the various connection statuses. I hope this helps. Let me know what you think. Thanks! 
kiali network connections

josunect:
I will close this issue for the version 2 based on @fkargbo feedback: https://github.com/kiali/kiali/pull/7693

PR #7601, opened 2024-08-07
Verify uncommitted new i18n statements

ferhoyos:
### Describe the change

When a developer adds a new i18n label (e.g., t('label')), translation files must be re-generated with the new i18n statements. To avoid committing without the new i18n statements, this PR updates the pre-commit hook to re-generate the translation files and ensure that no new i18n statements remain uncommitted.

### Steps to test the PR

1. Add a new i18n label (e.g., change `Log In` to `t('Log In')` in [LoginPage.tsx](https://github.com/kiali/kiali/blob/86d5778217d5784aeb8bfbe4f6d712980378a91f/frontend/src/pages/Login/LoginPage.tsx#L191)).
2. Attempt to create a commit.
3. Verify that the commit is aborted because new i18n statements have not been committed and `translation.json` is modified with the new i18n statement.

### Automation testing

N/A

### DIFF
--- a/hack/hooks/pre-commit.sh
+++ b/hack/hooks/pre-commit.sh
@@ -53,6 +53,6 @@ if [ -n "$yaml_unformatted" ] || [ -n "$go_unformatted" ]; then
 fi
 
 if [ -n "$i18n_files" ]; then
-  echo "New i18n statements have not been committed - the git commit is aborted."
+  echo "New i18n statements are generated but not committed - the git commit is aborted. Please include the updated translation files in the commit."
   exit 1
 fi
--- a/hack/hooks/pre-commit.sh
+++ b/hack/hooks/pre-commit.sh
@@ -39,8 +39,20 @@ if [ -n "$yaml_unformatted" ]; then
   yamlfmt $yaml_unformatted
 fi
 
+#### I18N missing statements ####
+cd frontend
+
+yarn i18n
+
+i18n_files=$(git diff --name-only HEAD --diff-filter=M | grep -E 'translation.json')
+
 #### Git commit check ####
 if [ -n "$yaml_unformatted" ] || [ -n "$go_unformatted" ]; then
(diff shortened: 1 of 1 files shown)


hack/hooks/pre-commit.sh
```diff
@@ -39,8 +39,20 @@ if [ -n "$yaml_unformatted" ]; then
...
+i18n_files=$(git diff --name-only HEAD --diff-filter=M | grep -E 'translation.json')
+
 #### Git commit check ####
 if [ -n "$yaml_unformatted" ] || [ -n "$go_unformatted" ]; then
   echo "Some files have been formatted - the git commit is aborted."
   exit 1
 fi
+
+if [ -n "$i18n_files" ]; then
+  echo "New i18n statements have not been committed - the git commit is aborted."
```
josunect, line 56: Could we add instructions about how to fix it? Now I know, but maybe someone doesn't have an idea about how to fix it. We could add (Ex): "Run 'make build-ui' and include the generated translation files" (Or a link with the instructions)?

josunect (commented):
Thanks, that's very useful and it works as expected. 
image

I've just added a small comment.

hack/hooks/pre-commit.sh
ferhoyos: With the verification the new i18n statements are added to the translation files, but not committed.

Is this message clearer for you?:

```
echo "New i18n statements are generated but not committed - the git commit is aborted. Please include the updated translation files in the commit"
```

josunect: Ah, i didn't realized it already was generated! So it looks good.

josunect (approved):

josunect (approved):
lgtm!

jmazzitelli (approved):
I approve. It works, and the new pre-commit check is fast, so it shouldn't bother anyone because the commit doesn't slow up much.

PR #7608, opened 2024-08-08
Include namespace in service versions yaml from Ambient hack bookinfo script

josunect:
### Describe the change

Apply the correct namespace as this is creating the additional services in the default namespace.

### Steps to test the PR

- `minikube start`
- Install kiali
- Verify no additional services are created in the default namespace.
- Verify the graph works as expected: 

image

### Automation testing

Waypoint e2e added in a follow up PR.

### Issue reference

N/A


hhovsepy:
I think this was added because in Ambient mode the Gateway API is enabled, and for GW API, each version of service should be a separate Service with own version instead of a Workload with version.

josunect:
> I think this was added because in Ambient mode the Gateway API is enabled, and for GW API, each version of service should be a separate Service with own version instead of a Workload with version.

@hhovsepy in that case, I've included the namespace, as it was incorrectly applied in the default namespace - Should be the bookinfo namespace.

hhovsepy (approved):
right @josunect , thanks for keeping that

PR #7615, opened 2024-08-12
Remove unsupported Kiali versions from tag creator workflow

ferhoyos:
### Describe the change

Remove the extra code from the `tag-creator` workflow added for Kiali versions `v1.24` and `v1.36`, which are currently unsupported. Yaml formatted with pre-commit hook.

Tested in my fork (https://github.com/ferhoyos/kiali/actions/runs/10347091079)


jshaughn (approved):
This seems fine to me.

PR #7625, opened 2024-08-14
[v1.73][OSSM-6835][OSSM-6843] Fix for cypress flakiness on OCP 4.16

mkralik3:
### Describe the change
The `sleep` application is scaled down in one test but it does not scale up after the test, which causes failures in the next tests.

The test `The idle status of a logical mesh application is reported in the list of applications` in the `apps.feature` scale down `sleep` application, so before the `services.feature` tests start, the `sleep` application is installed again (because there is no sleep pod ready) by the hack script which causes restarting of Kiali pod. 

However, the session had already been saved before that, so after the Kiali restart, the tests could not continue because they contained old session info.

ref.: https://issues.redhat.com/browse/OSSM-6835 and also https://issues.redhat.com/browse/OSSM-6843 ( the big number of restarts were already fixed by https://github.com/kiali/kiali/pull/7576 )

~~So I have introduced new `SKIP_INSTALL_DEMOS` env to disable reinstalling demo apps if needed (by default, it is enabled as it is working now) ( e.g. running on OCP ) and a fix to scale sleep application up after the test.~~

All tests passed with this fix: https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/kiali/job/test-jobs/job/kiali-cypress-tests/2876/ 

Backports:

- https://github.com/kiali/kiali/pull/7626
- https://github.com/kiali/kiali/pull/7627

### DIFF
--- a/frontend/cypress/integration/common/hooks.ts
+++ b/frontend/cypress/integration/common/hooks.ts
@@ -1,4 +1,4 @@
-import { Before } from '@badeball/cypress-cucumber-preprocessor';
+import { Before, After } from '@badeball/cypress-cucumber-preprocessor';
 
 function install_demoapp(demoapp: string) {
   var namespaces: string = 'bookinfo';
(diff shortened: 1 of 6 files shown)


ScriptingShrimp (approved):
@mkralik3 thanks for fix! This also highlight why we need spend time refactoring and stabilizing  cypress hooks and hack scripts.

Issue #7457, opened 2024-06-20
Properly use Group + Version + Kind for kube resources


nrfox:
Kiali relies on the `Kind` field to be unique across all kube resources but only `Group` + `Version` + `Kind` is unique. This has already been an issue with the `Gateway` kind when adding gateway api support. We keep track of the "internal" kind and the "real" kind separately when we want to display an object. Kiali should instead just use the full GVK (Group,Version,Kind) internally and externally when displaying resources to users. If externally we don't want to use the full GVK when displaying resources, Kiali can choose how to disambiguate between two resources with the same `Kind` but at the very least internally objects should use the full GVK.

- [x] Istio Config object CRUD API requests in `group/version/kind` format.
- [x] UI changes to have all URL references to Istio Configs in `group/version/kind` format.
- [x] Istio Config validations response json change to GVK format.
- [x] Istio Config List response to list configs in GVK keys not hardcoded plural properties like 'k8sgateways'. TODO

PR #7628, opened 2024-08-15
Added GroupVersionKind support for Istio Config objects lists.

hhovsepy:
### Describe the change

ResourceTypesToAPI changed to refer to GroupVersionKind.
Removed K8s Actual Types.
Istio Config CRUD and Validations requests to backend are now in `namespaces/namespace/istio/group/version/kind/name` format.

### Steps to test the PR

Regression testig of Kiali UI and validations.
Check K8s Gateway API validations.

### Automation testing

Modified existing unit tests.

### Issue reference

https://github.com/kiali/kiali/issues/7457

TODO in a next iteration, change REST API response JSON format to have all config types in a map instead of hardcoded properties. Remove the `dicIstioType` in frontend.


kubernetes/types.go
```diff
@@ -57,40 +57,25 @@ const (
...
-	K8sGatewayType = "K8sGateway"
-	// K8sActualGatewayType There is a naming conflict between Istio and K8s Gateways, keeping here an actual type to show in YAML editor
-	K8sActualGatewayType = "Gateway"
-	K8sActualGateways    = "gateways"
+	K8sGatewayType = "Gateway"
 
-	K8sGatewayClasses       = "k8sgatewayclasses"
-	K8sGatewayClassType     = "GatewayClass"
-	K8sActualGatewayClasses = "gatewayclasses"
+	K8sGatewayClasses   = "k8sgatewayclasses"
```
nrfox, line 62: It looks like we're still keeping some of the `k8s...` types. I was expecting this change to remove those and replace them with the real GVK e.g. `gateway.networking.k8s.io/v1.Gateway`. I haven't looked at it very closely but it doesn't seem like we need to define any of the types/mappings here in `kubernete/types.go` since those are already defined in the api packages themselves.

Are we going to show the full GVK on the frontend as well? For example this:
Screenshot from 2024-08-16 09-45-20

would become something like `Gateway (gateway.networking.k8s.io/v1)` or `gateway.networking.k8s.io/v1.Gateway` etc. since that is the real type.

hhovsepy: yeah we still have them, because of istio config details page, when the url is '/namespaces/namespace/configtype/configname'

hhovsepy: If we want to completely remove `K8sGateways    = "k8sgateways"` and the rest, then we need to rework `handlers.IstioConfigDetails /api/namespaces/{namespace}/istio/{object_type}/{object}` and similar REST endpoints for Istoi config.

nrfox: Yes the endpoints would need to be updated in order to fully remove `k8sgateways`. If we wanted to just display the real type to the user we could maintain the mapping on the backend along with the routes. The frontend should always have the full object so it can use `.apiVersion` and `.kind`. If we're going to close #7457 with this change can we create a separate issue to track displaying the real type on the frontend and possibly reworking the routes?

hhovsepy:
Hi @nrfox @ferhoyos this PR is ready now to be reviewed, this is too big, sorry about that, but I made as small as possible. There will be another PR to include the TODO comments I put which will bring another huge changes better to have separately.

frontend/src/components/VirtualList/Config.ts
```diff
@@ -188,79 +189,85 @@ const istioType: ResourceType<IstioConfigItem> = {
...
+  GVKToBadge[gvkToString(dicIstioTypeToGVK[key])] = PFBadges[key];
+});
+
+// export const IstioTypes: { [type: string]: IstioConfigType } = {
+//   adapter: { name: 'Adapter', url: 'adapters', badge: PFBadges.Adapter },
+//   attributemanifest: {
+//     name: 'AttributeManifest',
+//     url: 'attributemanifests',
+//     badge: PFBadges.AttributeManifest
+//   },
```
nrfox, line 210: Are these commented out for a reason?

nrfox:
> TODO in a next iteration, change REST API response JSON format to have all config types in a map instead of hardcoded properties. Remove the dicIstioType in frontend.

If there's some follow up TODOs, can we add those as checklist items on the linked issue before we close it? https://github.com/kiali/kiali/issues/7457

frontend/src/types/IstioConfigList.ts
```diff
@@ -21,11 +21,13 @@ import {
...
   WorkloadEntry,
   WorkloadGroup,
-  IstioObject
+  IstioObject,
+  GroupVersionKind
 } from './IstioObjects';
 import { ResourcePermissions } from './Permissions';
 
 export interface IstioConfigItem {
+  apiVersion?: string;
```
nrfox, line 30: `apiVersion` and `kind` should be required. They can probably be their own interface `TypeMeta` that this one `extends`.

```
export interface TypeMeta {
  apiVersion: string;
  kind: string;
}
...
export interface IstioConfigItem extends TypeMeta { ... }
```
I can submit a small PR for this if it'd be helpful?

nrfox (commented):
By convention, the plural Resource name (GVR) is used in the URL e.g. 
```
apis/networking.istio.io/v1/namespaces/default/virtualservices
```
And that then gets parsed into the GVK but I'm not sure it matters since the Kiali API does not follow the same conventions as the kube api.

frontend/src/components/VirtualList/Config.ts
nrfox: @hhovsepy there's a few e2e tests failing across the suites and it looks like it's related to the parent comment's wizard creation problem.

frontend/src/types/IstioConfigList.ts
hhovsepy: yes please

nrfox: Opened: https://github.com/hhovsepy/kiali/pull/5

hhovsepy: thanks @nrfox merged

nrfox (approved):

PR #7629, opened 2024-08-15
stop kubecache on namespace delete

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7655

_note: (this PR is based on my [discovery_selectors PR]( https://github.com/kiali/kiali/pull/7592))_

the important commit is this one: 
https://github.com/kiali/kiali/pull/7629/commits/bc6f38cf096ae8c5f99f5b13f6e69d122dbcff80

### DIFF
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -10,6 +10,7 @@ import (
 	"golang.org/x/exp/maps"
 
 	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/kiali/kiali/config"
@@ -87,9 +88,6 @@ type KialiCache interface {
 	// SetNamespace caches a specific namespace by cluster + token.
 	SetNamespace(token string, namespace models.Namespace)
 
-	// DeleteNamespace cleans up some internals when it has been detected that a namespace has been deleted.
-	DeleteNamespace(cluster string, namespace string, token string)
-
 	// Stop stops the cache and all its kube caches.
 	Stop()
 }
@@ -149,7 +147,7 @@ func NewKialiCache(clientFactory kubernetes.ClientFactory, cfg config.Config) (K
 	}
 
 	for cluster, client := range clientFactory.GetSAClients() {
-		cache, err := NewKubeCache(client, cfg, &kialiCacheImpl)
+		cache, err := NewKubeCache(client, cfg, kialiCacheImpl.deleteNamespace)
 		if err != nil {
 			log.Errorf("[Kiali Cache] Error creating kube cache for cluster: [%s]. Err: %v", cluster, err)
 			return nil, err
@@ -385,17 +383,15 @@ func (c *kialiCacheImpl) SetNamespace(token string, namespace models.Namespace) 
 	c.namespaceStore.Set(key, ns)
 }
 
-// DeleteNamespace should be called when it has been detected that a namespace was deleted from a particular cluster.
-// This cleans up some internals related to the namespace that is now gone.
-func (c *kialiCacheImpl) DeleteNamespace(cluster string, namespace string, token string) {
-	c.namespacesLock.Lock()
-	defer c.namespacesLock.Unlock()
-
-	key := namespacesKey{cluster: cluster, token: token}
-	c.namespaceStore.Remove(key)
-
-	if kubeCache, err := c.GetKubeCache(cluster); err == nil {
-		kubeCache.StopNamespace(namespace)
+// deleteNamespace is an error handler that will clean up some internals related to the given namespace
+// if that namespace has been detected to have been deleted. This function is called by the kube cache
+// informers when they encounter an error.
+func (c *kialiCacheImpl) deleteNamespace(kc *kubeCache, namespace string, err error) {
+	if apierrors.IsForbidden(err) || apierrors.IsUnauthorized(err) {
+		cluster := kc.client.ClusterInfo().Name
+		log.Errorf("Namespace [%v] in cluster [%v] appears to have been deleted or Kiali is forbidden from seeing it [err=%v]. Shutting down namespace cache.", namespace, cluster, err)
+		c.RefreshTokenNamespaces(cluster)
+		kc.StopNamespace(namespace)
 	}
 }
 
--- a/kubernetes/cache/cache.go
+++ b/kubernetes/cache/cache.go
@@ -87,6 +87,9 @@ type KialiCache interface {
 	// SetNamespace caches a specific namespace by cluster + token.
 	SetNamespace(token string, namespace models.Namespace)
 
+	// DeleteNamespace cleans up some internals when it has been detected that a namespace has been deleted.
+	DeleteNamespace(cluster string, namespace string, token string)
+
 	// Stop stops the cache and all its kube caches.
 	Stop()
 }
@@ -146,7 +149,7 @@ func NewKialiCache(clientFactory kubernetes.ClientFactory, cfg config.Config) (K
 	}
 
 	for cluster, client := range clientFactory.GetSAClients() {
-		cache, err := NewKubeCache(client, cfg)
+		cache, err := NewKubeCache(client, cfg, &kialiCacheImpl)
 		if err != nil {
 			log.Errorf("[Kiali Cache] Error creating kube cache for cluster: [%s]. Err: %v", cluster, err)
 			return nil, err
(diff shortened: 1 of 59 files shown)


jmazzitelli:
I'm looking closely and stepping through this code to see what's happening under the covers.

This code is properly shutting down the informers. `c.stop(namespace)` ultimately closes the channel that triggers the informer to stop watching. That function eventually hits this code in `kube_cache.go`; notice the `close(nsChan` - that is the important thing:

```go
		if nsChan, exist := c.stopNSChans[namespace]; exist {
			close(nsChan)
			delete(c.stopNSChans, namespace)
			delete(c.nsCacheLister, namespace)
```

Note also we purge the namespace-specific objects from the nsCacheLister so all of those listers should be garbage collected.

So, this all works.

A couple things I think need to be addressed still.
1. We probably need to clean up some things in the KialiCache (for example, remove the kubecache associated with the namespace; clean up other namespace-specific things like prehaps `namespaceStore` which is of type `store.Store[namespacesKey, map[string]models.Namespace]`.
2. Determine what to do if a namespace that was deleted is later re-created. We cannot restart informers. So we would need to re-create the informers/listers for the namespace (we first need to know the namespace was re-created, and then we need to re-initialize it just as we do when the server starts up). This might not be an issue, because the operator might already solve this problem for us (the operator sees the new namespace and should restart the server pod; I'm not sure if that will actually happen; we need to test the operator behavior when a namespace is deleted and then re-created).

jmazzitelli:
The two things I mentioned in the previous comment are addressed.

1. Namespace store is cleaned up. See new DeleteNamespace function.
2. When a namespace is recreated, the server can survive this and refresh (recreate) the informers so it can begin watching the namespace again. See change to getCacheLister function.

jmazzitelli:
Here's how to quickly test this.

1. Get a cluster up with kiali operator installed with namespace watching enabled (in its env var WATCHES_FILE, set it to `watches-k8s-ns.yaml` as an example).
2. In the Kiali CR you need to have `deployment.cluster_wide_access` set to `false` and set `deployment.discovery_selectors.default` to some namespace:
```yaml
deployment:
  cluster_wide_access: false
  discovery_selectors:
    default:
    - matchLabels: {"kubernetes.io/metadata.name": "foo"}
```
3. The Kiali UI should work normally. You should see "foo" namespace as well as the "istio-system" namespace.
4. Delete the foo namespace (`kubectl delete ns foo`).
5. For some reason, it takes a long time for the k8s client informers to see the namespace is gone. So wait until you see a messages like this in the kiali server logs (the first is the error logs from the k8s informer code; the second is the kiali log message)
```
E0822 21:31:21.950719       1 reflector.go:150] pkg/mod/k8s.io/client-go@v0.30.0/tools/cache/reflector.go:232: Failed to watch *v1alpha1.WasmPlugin: unknown (get wasmplugins.extensions.istio.io)
2024-08-22T21:31:21Z ERR Namespace [foo] in cluster [Kubernetes] appears to have been deleted or Kiali is forbidden from seeing it [err=unknown (get pods)]. Shutting down namespace cache.
```
6. Go to the Kiali UI and see that you only see istio-system namespace. There is no `foo` namespace.
7. Notice in the server logs you no longer see a stream of logs coming from the k8s client informer saying it failed to watch something (see the first error message above for an example; it would come from `reflector.go` and say it can't watch some kind of object)
8. Now re-create the namespace (`kubectl create ns foo`). At this point the operator will see the new namespace, re-create the role for kiali to be able to see it. But it won't restart the kiali (for reasons I don't want to explain; suffice it to say, since the original list of accessible namespaces didn't change, and nothing else in the configmap changed, the operator doesn't restart the kiali pod).
9. Go to the Kiali UI and look at the Istio Config page, then go to some other pages like Applications, Workloads or the Graph. Notice you can see the "foo" namespace again.
10. Look in the server logs and notice that you will see the namespace-scoped cache for the foo namespace is restarted:
```
2024-08-22T21:38:54Z DBG Kiali Cache for namespace [foo] in cluster [Kubernetes]: attempting to restart informers
2024-08-22T21:38:54Z DBG [Kiali Cache] Starting namespace-scoped for namespace: foo informers
2024-08-22T21:38:54Z INF [Kiali Cache] Waiting for namespace-scoped for namespace: foo cache to sync
2024-08-22T21:38:54Z INF [Kiali Cache] Started
2024-08-22T21:38:54Z DBG Kiali Cache for namespace [foo] in cluster [Kubernetes]: successfully obtained listers
```

kubernetes/cache/cache.go
```diff
@@ -146,7 +149,7 @@ func NewKialiCache(clientFactory kubernetes.ClientFactory, cfg config.Config) (K
 	}
 
 	for cluster, client := range clientFactory.GetSAClients() {
-		cache, err := NewKubeCache(client, cfg)
+		cache, err := NewKubeCache(client, cfg, &kialiCacheImpl)
```
nrfox, line 152: If the `KubeCache` and the `KialiCache` have dependencies on each other, they might as well be the same struct. Can we just pass a `func(r *cache.Reflector, err error)` into the `KubeCache`? That way the two can remain more separate and you also don't have to export `DeleteNamespace`.

kubernetes/cache/cache.go
```diff
@@ -382,6 +385,20 @@ func (c *kialiCacheImpl) SetNamespace(token string, namespace models.Namespace)
...
+	defer c.namespacesLock.Unlock()
+
+	key := namespacesKey{cluster: cluster, token: token}
+	c.namespaceStore.Remove(key)
+
+	if kubeCache, err := c.GetKubeCache(cluster); err == nil {
+		kubeCache.StopNamespace(namespace)
+	}
+}
+
```
nrfox, line 401: Might as well delete all the tokens for that namespace/cluster right? Something like `RefreshTokenNamespace(cluster, namespace)`.

kubernetes/cache/kube_cache.go
```diff
@@ -541,7 +598,23 @@ func (c *kubeCache) getCacheLister(namespace string) *cacheLister {
...
+		}
+		lister, ok = c.nsCacheLister[namespace]
+		if ok {
+			log.Debugf("%v: successfully obtained listers", logPrefix)
+		} else {
+			// this likely means the caller is going to segfault with nil pointer; TODO return error and let callers gracefully fail
+			log.Fatalf("%v: failed to obtain listers", logPrefix)
+		}
+	}
+	return lister
```
nrfox, line 622: iirc this used to work by calling `CheckNamespace` in every `Get<Resource>` method. Way more lines of code but one less lock. This is probably a better place to do it for now.

kubernetes/cache/kube_cache.go
```diff
@@ -507,6 +531,39 @@ func (c *kubeCache) createKubernetesInformers(namespace string) informers.Shared
...
+			sharedInformers.Apps().V1().DaemonSets().Informer(),
+			sharedInformers.Core().V1().Services().Informer(),
+			sharedInformers.Core().V1().Endpoints().Informer(),
+			sharedInformers.Core().V1().Pods().Informer(),
+			sharedInformers.Apps().V1().ReplicaSets().Informer(),
+			sharedInformers.Core().V1().ConfigMaps().Informer(),
+		}
+
+		watchErrorHandler := func(r *cache.Reflector, err error) {
+			if apierrors.IsForbidden(err) || apierrors.IsUnauthorized(err) {
```
nrfox, line 557: We're only checking for these two errors because I'm guessing the operator deletes the role?

nrfox (commented):
I'm still not very keen on putting this all in the server but I don't have any better suggestions if ansible operator doesn't support our use case.

kubernetes/cache/cache.go
jmazzitelli: Yes. Good idea. I implemented "ErrorHandler" now which avoids the dependency - see commit https://github.com/kiali/kiali/pull/7629/commits/3ce366677cc10a4bd35882f4f6973dab5b979ff3

kubernetes/cache/cache.go
jmazzitelli: OK. I put that in the implementation along with the new ErrorHandler stuff. See https://github.com/kiali/kiali/commit/3ce366677cc10a4bd35882f4f6973dab5b979ff3#diff-cfd09f004cc6d39d422ad55dba06932081887b2323b3854596e06a27a6563d74R393

kubernetes/cache/kube_cache.go
jmazzitelli: Right. So I tried to pinpoint a place in the code that would catch the problem without adding a bunch of lines of code. This appeared to be that place.

jmazzitelli:
After the latest commit, I ran some quick tests to make sure it doesn't break anything, and it is working fine.

jmazzitelli:
Starting to get conflicts because master is moving forward and this PR (with discovery selectors PRs) are so large but not getting merged. So I'm closing this PR, cherry picking its contents over to https://github.com/kiali/kiali/pull/7592 so conflict resolutions will at least be a tiny bit easier. Right now, it's a pita.

PR #7650, opened 2024-08-20
Allow specifying prometheus version explicitly if endpoint is not available

JDuskey:
### Describe the change

Kiali attempts to load version/revision information from prometheus as part of the status checks. When using Grafana Mimir as the metrics datasource, this endpoint isn't available. There is a separate build info endpoint for Mimir, but the data is not structured in the way Kiali is expecting. This PR extends the functionality created to address https://github.com/kiali/kiali/pull/4386 to allow specifying a version manually to allow metrics queries to execute when using non-vanilla-prometheus metrics stores.

### Issue reference

https://github.com/kiali/kiali/pull/4386

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -192,6 +192,7 @@ type ThanosProxy struct {
 	Enabled         bool   `yaml:"enabled,omitempty"`
 	RetentionPeriod string `yaml:"retention_period,omitempty"`
 	ScrapeInterval  string `yaml:"scrape_interval,omitempty"`
+	Version         string `yaml:"version,omitempty"`
 }
 
 // PrometheusConfig describes configuration of the Prometheus component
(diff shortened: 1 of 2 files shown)


jmazzitelli:
Seems kinda hacky to require a fake version string to be configured to be able to workaround a missing version endpoint in a Prometheus proxy.

Also, I thought there was a way to tell Kiali to not check Prom status (I can't remember if this is true, but I seem to remember something like this. But perhaps that's the  better way to solve this (add a way to skip version retrieval, rather than to require the user to provide some fake version string that could be completely meaningless or, worse, wrong.)

For the record, before we accept PRs from contributors, we like to see a github issue created first, so we can dialogue about what the problem is and discuss potential solutions. For example, we should have been able to discuss potential solutions before this "fake-version" setting was implemented. Please see the contributor's guidelines here: https://github.com/kiali/kiali/blob/master/CONTRIBUTING.md#making-a-change

JDuskey:
I've closed this out and created an issue for this.

Issue #7590, opened 2024-08-02
Inconsistent ServiceEntry Display in Multi-Namespace Environment


zodiac12k:
### Describe the bug

We are experiencing an issue with the Kiali graph displaying ServiceEntry connections incorrectly in a multi-namespace environment. Our Istio global mesh config is set with `defaultServiceExportTo` as `"."`. We have the same host's ServiceEntry registered in two different namespaces. However, intermittently, the Kiali graph shows traffic routing to the ServiceEntry in the other namespace. When we explicitly set `exportTo` as `"."` in the ServiceEntry, it displays correctly, routing traffic within the same namespace.

### Expected Behavior

Each namespace's workloads should route traffic to their respective ServiceEntry resources, as specified by the global mesh config setting.

### What are the steps to reproduce this bug?

1. Configure Istio global mesh config with `defaultServiceExportTo: "."`.
2. Create two namespaces.
3. In each namespace, create a ServiceEntry for the same host.
4. Deploy workloads in each namespace that use the ServiceEntry.
5. Observe the traffic flow in Kiali.

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:** 1.85.0
- **Istio version:** 1.21.3
- **Kubernetes impl:** Openshift
- **Kubernetes version:** 1.23.17
- **Other notable environmental factors:**

jmazzitelli:
This sounds like this might be a duplicate. Please see these issues and see if they are related:

* https://github.com/kiali/kiali/issues/7537
* https://github.com/kiali/kiali/issues/7153

This comment is one I think may be related: https://github.com/kiali/kiali/issues/7537#issuecomment-2234052694

zodiac12k:
> This sounds like this might be a duplicate. Please see these issues and see if they are related:
> 
> * [Kiali Graph showing incorrectly when we create Multiple Service entry with same hosts and different port #7537](https://github.com/kiali/kiali/issues/7537)
> * [[graph] ServiceEntry ExportTo is not handled correctly #7153](https://github.com/kiali/kiali/issues/7153)
> 
> This comment is one I think may be related: [#7537 (comment)](https://github.com/kiali/kiali/issues/7537#issuecomment-2234052694)

@jmazzitelli Thank you for your reply
It seems in your comment the ports are different, but in our case, the ports are the same.

jshaughn:
[...]

PR #7652, opened 2024-08-20
Added support of Mesh config defaultExportTo fields.

hhovsepy:
### Describe the change
Mesh config has defaultExportTo values for DR, VS and SE (including Services) which should be considered when Istio config object does not have 'ExportTo' fields.

### Steps to test the PR

Make sure Istio's Global MeshConfig has values set for `defaultServiceExportTo`, `defaultVirtualServiceExportTo` and `defaultDestinationRuleExportTo`, for instance to `.`. By default the value is `*`.
Screenshot from 2024-08-28 09-36-48

Promote VS, DR and SE validation scenarios. On those objects which does not have 'ExportTo' themselves. Verify that these objects behave like they are not available in other namespaces.
Also verify that when objects have 'ExportTo' to `*`, it overrides the default `.`

Original issue scenario with SE in graph page should also work: https://github.com/kiali/kiali/issues/7590
Was not able to find the steps, will ask the issue owner to check.

### Automation testing

Existing unit tests modified, added more ones.

### Issue reference

https://github.com/kiali/kiali/issues/7590

Please note that we have an issue in validations which makes inconsistency between list and details pages when 'exportTo' field exists: https://github.com/kiali/kiali/issues/7685


hhovsepy:
Thank you @jshaughn for the review, all comments approached.

jshaughn (approved):
Looks good, Hayk.  I also added the backport label to the issue for v1.89. Thanks!

PR #7656, opened 2024-08-21
Adapt cypress tests to OSSMC

ferhoyos:
### Describe the change

Adapt cypress tests to run with OSSMC:
- Add wildcard to the URLs intercepted to include the OSSMC proxy URL.
- Set the input parameters of the `visit` and `intercept` cypress commands as objects so they can be overwritten by OSSMC.
- Replace the relative paths of Kiali code imports with absolute paths since the folder structure in OSSMC is different. 
- Make some Cypress tests more specific to ensure they work in OSSMC.
- Add `@ossmc` and `@skip-ossmc` cypress tags to selectively execute tests with OSSMC.

### Steps to test the PR

Verify that CI tests pass

### Issue reference

https://github.com/kiali/openshift-servicemesh-plugin/issues/356

### DIFF
--- a/frontend/cypress/integration/common/app_details.ts
+++ b/frontend/cypress/integration/common/app_details.ts
@@ -83,11 +83,11 @@ Then('no cluster badge for the {string} should be visible', (type: string) => {
   if (type === 'Istio config') {
     cy.get('#pfbadge-C').should('not.exist');
   } else if (type === 'graph side panel') {
-    cy.get('#graph-side-panel').within($div => {
+    cy.get('#graph-side-panel').within(() => {
       cy.get('#pfbadge-C').should('not.exist');
     });
   } else {
-    cy.get(`#${type[0].toUpperCase()}${type.slice(1)}DescriptionCard`).within($article => {
+    cy.get(`#${type[0].toUpperCase()}${type.slice(1)}DescriptionCard`).within(() => {
       cy.get('#pfbadge-C').should('not.exist');
     });
   }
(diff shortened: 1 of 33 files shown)


ScriptingShrimp:
testing via upstream-istio-pipeline/2163

frontend/cypress/integration/common/app_details.ts
```diff
@@ -83,11 +83,11 @@ Then('no cluster badge for the {string} should be visible', (type: string) => {
   if (type === 'Istio config') {
     cy.get('#pfbadge-C').should('not.exist');
   } else if (type === 'graph side panel') {
-    cy.get('#graph-side-panel').within($div => {
+    cy.get('#graph-side-panel').within(() => {
       cy.get('#pfbadge-C').should('not.exist');
     });
   } else {
-    cy.get(`#${type[0].toUpperCase()}${type.slice(1)}DescriptionCard`).within($article => {
+    cy.get(`#${type[0].toUpperCase()}${type.slice(1)}DescriptionCard`).within(() => {
```
ScriptingShrimp, line 90: I find `.within($article => { ... })` valid here

ferhoyos: Yes, it is valid, but the $article input variable is not used within the arrow function. VS code linting tool encourage to remove unused variables and function parameters.

frontend/cypress/integration/common/graph_toolbar.ts
```diff
@@ -187,7 +186,7 @@ Then('user does not see graph duration menu', () => {
 
 Then('user sees selected graph duration {string}', (duration: string) => {
   cy.get('button#time_range_duration-toggle')
-    .find('span[class*="pf-v5-c-menu-toggle__text"]')
+    .find('span[class="pf-v5-c-menu-toggle__text"]')
```
ScriptingShrimp, line 189: looks like removing wildcard did not affected test run

ferhoyos: No, because the CSS class is exact. In that case, I prefer to remove the wildcard to be more precise in the cypress test.

frontend/cypress/integration/common/navigation.ts
```diff
@@ -57,7 +57,7 @@ Given(
       }).as('waitForCall');
     }
 
-    cy.visit(`${Cypress.config('baseUrl')}/console/namespaces/${namespace}/${pageDetail}/${name}`, { qs });
+    cy.visit({ url: `${Cypress.config('baseUrl')}/console/namespaces/${namespace}/${pageDetail}/${name}`, qs });
```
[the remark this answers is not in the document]
ferhoyos: Yes, I only made the necessary changes and kept the URL the same as it was before.

PR #7659, opened 2024-08-21
Check if a Helm chart tarbal is defined when using a dev image in a MC script

leandroberetta:
image


jmazzitelli (approved):
this looks good!

PR #7669, opened 2024-08-26
[v1.73] Handle structs or bools when unmarshaling service entries from the istio registry

nrfox:
### Describe the change

In istio 1.20 these are `map[string]struct{}` but in earlier versions they are `map[string]bool` so this change implements some custom json unmarshaling to handle both.

### Steps to test the PR

Deploy Kiali 1.73 with istio 1.20. Create a `ServiceEntry` with `exportTo` set:

```
kind: ServiceEntry
metadata:
    name: se-kiali-test
    namespace: kialitest
spec:
    exportTo:
    - .
    - istio-system
    hosts:
      - www.example.com
    location: MESH_EXTERNAL
    ports:
      - name: https
        number: 443
        protocol: HTTPS
        targetPort: 443
    resolution: DNS
```

### Automation testing

Not covered.

### Issue reference

Relates to https://github.com/kiali/kiali/issues/6856

### DIFF
--- a/kubernetes/filters_test.go
+++ b/kubernetes/filters_test.go
@@ -231,8 +231,8 @@ func CreateFakeRegistryService(host string, namespace string, exportToNamespace 
 	registryService.Hostname = host
 	registryService.IstioService.Attributes.Namespace = namespace
 	registryService.IstioService.Attributes.Labels = labels
-	registryService.IstioService.Attributes.ExportTo = make(map[string]bool)
-	registryService.IstioService.Attributes.ExportTo[exportToNamespace] = true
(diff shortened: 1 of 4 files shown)


jshaughn (approved):
Nice turnaround, @nrfox.  Code looks good.  I installed 1.20 and created the test SE, and it broke as expected.  I updated to the PR and it worked.  I then installed 1.18 and created the test SE. It worked as expected.  I updated to the PR and it continued to work.  So it seems good to me, and a nice example of how to add custom marshaling.

ScriptingShrimp:
I see this is fix for OSSM-8006 which is pending ON_QA. 
How to verify this? Creation of `ServiceEntry` does not throw error?

PR #7673, opened 2024-08-27
Fix multicluster service entry scenario

### LINKED ISSUE #7589: Multi-cluster ServiceEntry Routing
### Describe the bug

We are experiencing an issue with Kiali's multi-cluster functionality. Specifically, it appears that ServiceEntry routing is not functioning as expected. In our setup, we have two clusters, e.g., `a1` and `a2`, with identical ServiceEntry configurations for the same host. However, the traffic seems to be incorrectly routed to the ServiceEntry of the other cluster, as depicted in the attached Kiali graph.

### Expected Behavior

Each cluster's workloads should route traffic to their respective ServiceEntry resources.

1. Set up two clusters (`a1` and `a2`).
2. In both clusters, create identical ServiceEntry resources for the same host, e.g., `a.example.com`.
3. Deploy workloads in both clusters that should use their respective ServiceEntry resources.
4. Observe the traffic flow in Kiali.

### Environment

- **Kiali version:** 1.85.0
- **Istio version:** 1.21.3
- **Kubernetes impl:** Openshift
- **Kubernetes version:** 1.23.17

jshaughn:
Fixes #7589

Ensure that when an se-service node is shared by (i.e. has edges from) nodes on different clusters, that each cluster gets its own service entry nodes (using the cluster-specific SE definitions).

I think this is best tested via the new unit test.  And hopefully by the issue creator in his env.

### DIFF
--- a/graph/telemetry/istio/appender/service_entry.go
+++ b/graph/telemetry/istio/appender/service_entry.go
@@ -75,12 +75,13 @@ func (a ServiceEntryAppender) AppendGraph(trafficMap graph.TrafficMap, globalInf
 		}
 
 		// for non-service nodes, look for edges to non-injected service nodes that may represent a mesh_external ServicEntry.
-		// It probably will not have cluster and namespace set, either way, we need to get this information from the source node
-		// because it is the requesting service that needs access to the ServiceEntry.
+		// It probably will not have cluster and namespace set (because it's external to the mesh, the dest_service is just a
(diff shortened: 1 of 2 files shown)


hhovsepy (approved):
Verified locally by test debug for before and after.
I can see the issue before, wrong Dest.ID, from another cluster, messed up:
Screenshot from 2024-08-28 09-17-48

And after, Edge.Dest.ID is correct:
Screenshot from 2024-08-28 09-19-49

PR #7675, opened 2024-08-28
Update webpack to version 5.94

ferhoyos:
### Describe the change

Update webpack to version 5.94

### Steps to test the PR

Verify that Kiali works as expected and CI tests pass

### Automation testing

N/A


jshaughn (approved):
This looks fine to me, I'm trusting CI here as it already does a build and runs tests.   How did you update just the yarn.lock and not have to touch package.json?

ferhoyos:
The key is understanding how yarn works. When a dependency has a version with caret ^, yarn will download the latest minor version. In this case, webpack is a dependency of react-scripts:

```
webpack "^5.64.4"
```

So, if you want the latest minor version of the webpack library (v5.94.0), you need to remove the entry in yarn.lock and execute `yarn install`. For a major version update, you would need to specify the version in the resolutions section of package.json

Issue #7685, opened 2024-08-30
Validations - Inconsistency between List and Details pages


hhovsepy:
### Describe the bug

Istio Config Validations have inconsistency between List and Details pages when Istio Config objects have 'exportTo'.
Or when default mesh exported to local namespace.
This issue happens because Istio Config list calls validations for all namespaces, so the 'exportTo' per config is ignored.

### Expected Behavior
Validations should be consistent in all places.

### What are the steps to reproduce this bug?

1. Create 'bookinfo' and 'bookinfo2' namespaces.
2. In each of namespaces create any object with 'exportTo: .', for instance DR 
```
apiVersion: networking.istio.io/v1
kind: DestinationRule
metadata:
  name: all.googleapis.com
spec:
  host: '*.googleapis.com'
  trafficPolicy:
    loadBalancer:
      simple: ROUND_ROBIN
    portLevelSettings:
      - port:
          number: 443
        tls:
          mode: SIMPLE # initiates HTTPS when accessing www.googleapis.com
  exportTo:
  - '.'
```
3. See that List page shows validation warning (which is wrong, because those objects should not interact as both exported to own namespace), but the details page does not (which is correct).
![Screenshot from 2024-08-30 11-29-07](https://github.com/user-attachments/assets/a1b08873-9e69-4b58-b4ff-890c0ebc5229)
![Screenshot from 2024-08-30 11-28-47](https://github.com/user-attachments/assets/1e5304e3-7580-40ba-ab28-9ace5a445d05)

The same inconsistency is in Overview page.

### Environment

- **Kiali version: 1.90 snapshot, and v1.89, probably even older versions**
- **Istio version: 1.23**

jshaughn:
@hhovsepy I added the backport-needed label, I assume this bug fix should go into 1.89.  If not, remove the label.

Issue #4713, opened 2022-02-07
Review and plan a migration of Kiali validations to Istio Analyze


lucasponce:
This Epic is a follow-up discussion started in https://github.com/kiali/kiali/discussions/4080#discussioncomment-1210770.

One of the Kiali's goals is to help Istio users to help to manage Istio configuration providing a validation service built-in in the YAML editor.
https://kiali.io/docs/features/validations/

Istio community echoed this service and introduced a similar service that runs similar validations as proposed by Kiali, also the Kiali team contributed to some of these commits [1] (i.e. Kudos to @xeviknal).

https://istio.io/latest/docs/ops/diagnostic-tools/istioctl-analyze/

As discussed in the main discussion, it's a good moment to review the Kiali validations and study which logic could be moved to the Istio upstream.

In that way, Kiali could simplify the validations services delegating into the Istio feature and adding value in the manipulation of these validations across other features.

So, the tasks of these epics would be:
- Identify which validations of Kiali are already covered in Istioctl Analyze.
- Study which Kiali validations can be migrated as a new Istio analyzer and be part of the Istio upstream.
- Check the potential changes in the Kiali UI to support these validations as part of the "Istio Status" instead of the "Kiali Validation", there would be changes in the markers in the YAML editor.
- Propose a tentative plan of tasks for this epic.

[1] https://github.com/istio/istio/commits?author=xeviknal

jshaughn:
This is not a bad idea but is likely to be out of scope for the current maintainers.  Setting to Future and Help Wanted.

nrfox:
There's a few potential paths forward for this one I think that are not mutually exclusive:

1. Rely on the [configuration status](https://istio.io/latest/docs/reference/config/config-status/)
2. Run istioctl analyze from Kiali. We'd need to include `istioctl` in the container image.
3. Import analyzers directly into Kiali: https://github.com/istio/istio/blob/master/pkg/config/analysis/README.md

jshaughn:
Closing as duplicate of https://github.com/kiali/kiali/issues/8225

PR #7688, opened 2024-08-30
Istio config list: loop call validations per namespace in backend

hhovsepy:
### Describe the change

We had a regression at some point when calling validations per all namespaces, this caused an issue that Istio config objects are not filtered by `exportTo` field, because there is not destination namespace to check whether particular Istio config is exported to that namespace or not.

The solution here is to return back the loop per namespace to call validations.

There is another issue to be fixed later https://github.com/kiali/kiali/issues/7690
Istio config validations needs to be refactored and optimized, but this is out of scope of this PR.
This PR only returns back this loop of code missing of which was caused this validation inconsistency: https://github.com/kiali/kiali/pull/7168/files#diff-ce9a87f33646f1c782b605531625267299fc515a5754ecf697ef6e11c011ceaeL88-L99

The performance degradation for Istio Config list is 2.5 - 3.5 times.
Before:
All Namespaces Istio Configs load time: 1.0479 sec, baseline: 2 sec, difference: -0.95210 sec
Selected Namespaces Istio Configs load time: 0.49130 sec, baseline: 2 sec, difference: -1.5087 sec
After:
All Namespaces Istio Configs load time: 2.4857 sec, baseline: 2 sec, difference: +0.48570 sec
Selected Namespaces Istio Configs load time: 1.8972 sec, baseline: 2 sec, difference: -0.10280 sec

But still is ~ within 2s of baselines.

Overview page performance comparison before and after:
Before:
Init page load time: 1.4626 sec, baseline: 2 sec, difference: -0.53738 sec
After:
Init page load time: 2.4374 sec, baseline: 2 sec, difference: +0.43742 sec

All tests were run for 50 namespaces, 2 configs per namespace.
### Steps to test the PR

Steps from https://github.com/kiali/kiali/issues/7685

### Automation testing

Integration test created for this PR.

### Issue reference
https://github.com/kiali/kiali/issues/7685


handlers/istio_config.go
```diff
@@ -73,11 +73,25 @@ func IstioConfigList(w http.ResponseWriter, r *http.Request) {
...
+			loadedNamespaces, _ := business.Namespace.GetClusterNamespaces(r.Context(), cluster)
+			istioConfig.IstioValidations = models.IstioValidations{}
+			for _, ns := range loadedNamespaces {
+				nsValidations, nsErr := business.Validations.GetValidations(r.Context(), cluster, ns.Name, "", "")
+				if nsErr != nil {
+					handleErrorResponse(w, nsErr)
+					return
+				}
+				istioConfig.IstioValidations.MergeValidations(nsValidations)
+			}
```
nrfox, line 87: Why is the being done in the handler and not in the validations? It seems like when namespace is empty, `business.Validations.GetValidations` should validate `exportTo` across namespaces without having to loop through every namespace. Otherwise calling `GetValidations` with an empty namespace is never valid?

hhovsepy: Yes, calling `GetValidations` with empty `namespace` is wrong, even though it does not throw an exception, it will not consider `exportTo` of configs and will loop through all configs of all namespaces and do bulk list validation.
When having `namespace` value set, the validation will be done for configs which are `exportedTo` provided `naemspace`, and results will be correct.

jshaughn (approved):
If I understand this correctly, this PR is just a "quick fix" to revert back to per-namespace processing, even when asking to validate all namespaces (i.e. handler is passed `namespace == ""`).  The handler is now forcing an explicit namespace to be sent to `GetValidations`, bypassing the current support for `namespace == ""` in that function.  It seems like we should still be able to fetch all namespaces and then apply exportTo filtering, as needed at the business level.  Is this what we may do in a follow-up PR (maybe for 2.0?).  I'm conditionally approving this as long as my understanding is correct, and we have an issue created for a near-term "real fix".

handlers/istio_config.go
nrfox: > it will not consider exportTo of configs and will loop through all configs of all namespaces and do bulk list validation.

So if you pass `""` for namespace then `exportTo` will be ignored? This seems like an issue inside `GetValidations` then. Can we fix the problem there? Other places that call `GetValidations` outside of this handler may suffer from the same issue. Doesn't the `services` page have an `IncludeValidations` option?

hhovsepy: Yes, when you pass `""` for namespace then `exportTo` will be ignored, which is an issue.
Only other place where `GetValidations` are called is `ConfigValidationSummary` for Overview page. And here is inconsistency: Overview page calls validations then count Istio config objects by the validation results, so the `obj1` which is in `ns1` but exported to `ns2` might cause inconsistency: When namespace is empty in GetValidations call, than it will ignore `exportTo` and will show validations for `obj1` in`ns1` Overview card. When for Overview page we pass namespace param in "GetValidations", then `obj1` validations will be ignored in `ns1` Overview card, but then you go to 'ns1' Istio config list, then "obj1" is there with validations. Because Istio config list lists objects then adds validations to them.
Services page does not have config validations, but a service validations.
For the rest of the calls namespace param has a value.

hhovsepy:
Added the similar loop for Overview page's namespace cards validation to avoid inconsistency. @nrfox @jshaughn ready for re-review.

jshaughn:
> If I understand this correctly, this PR is just a "quick fix" to revert back to per-namespace processing, even when asking to validate all namespaces (i.e. handler is passed namespace == ""). The handler is now forcing an explicit namespace to be sent to GetValidations, bypassing the current support for namespace == "" in that function. It seems like we should still be able to fetch all namespaces and then apply exportTo filtering, as needed at the business level. Is this what we may do in a follow-up PR (maybe for 2.0?). I'm conditionally approving this as long as my understanding is correct, and we have an issue created for a near-term "real fix".

@hhovsepy Before re-review, can you respond to this prior comment? I'm trying to understand whether there is follow-on work, or whether this is the final fix.

hhovsepy:
> > If I understand this correctly, this PR is just a "quick fix" to revert back to per-namespace processing, even when asking to validate all namespaces (i.e. handler is passed namespace == ""). The handler is now forcing an explicit namespace to be sent to GetValidations, bypassing the current support for namespace == "" in that function. It seems like we should still be able to fetch all namespaces and then apply exportTo filtering, as needed at the business level. Is this what we may do in a follow-up PR (maybe for 2.0?). I'm conditionally approving this as long as my understanding is correct, and we have an issue created for a near-term "real fix".
> 
> @hhovsepy Before re-review, can you respond to this prior comment? I'm trying to understand whether there is follow-on work, or whether this is the final fix.

@jshaughn oh sorry, yes right, this is a quick fix of inconsistency, which in fact is reverting the call with namespace="" for performance improvements which causes inconsistency in corner cases when object is "exportedTo" another namespace. The necessity of this fix is that Validation are designed in the way, that they prepare Istio config list in the beginning then pass this list to loop validations, this is done fore performance reasons. And the weakness of that approach is that "exportTo" filtering is also done in preparation phase, which means that without "namespace" param given it cannot filter "exportTo", because it does not know against which "namespace" to filter "exportTo". 
The better solution is eigther complete rework all validations and add "exportTo" filtering for each of them, but this is risky and requires more efforts before 2.0. Or to cache Validation results which will resolve performance issues in loop calls per  namespace.

jshaughn (approved):
> The better solution is eigther complete rework all validations and add "exportTo" filtering for each of them, but this is risky and requires more efforts before 2.0. Or to cache Validation results which will resolve performance issues in loop calls per namespace.

Rework all validations seems to risky, and a lot of work.  Caching validations seems like the correct approach to me.  As a note, ambient does not support exportTo.

I am approving this workaround, but you should also wait for Nick.

nrfox (approved):

jshaughn:
> IMO before we do any major reworking of validations we need to do: https://github.com/kiali/kiali/issues/4713. Users do not want two different validation tools that potentially differ from one another.

I guess another option would be to somehow pull the `analyze` results into Kiali.  Not sure if that's feasible.  I'm ambivalent about migrating our stuff to Istio Analyze.

> Just to note that caching the validations won't actually speed up the time it takes to validate. It will only improve the user's perception of how long it takes to validate since we'll move validations outside of the request's life cycle. With caching, if validations take a long time to perform then users will likely see stale data and the Kiali server will consume more resources in performing the validations.

Right, I get that.  But I think config changes rarely.  I think I'd be fine , actually, performing validations at startup , caching it, and never doing it again unless someone either  makes a config change via Kiali, or possible requests Kiali to re-validate all config.

nrfox:
> I guess another option would be to somehow pull the analyze results into Kiali. Not sure if that's feasible. I'm ambivalent about migrating our stuff to Istio Analyze.

I think this is gated behind a feature flag in istio but the results can get written to the `status` field of every istio object.

Issue #7672, opened 2024-08-27
Cannot load the graph: cluster [unknown] is not found or is not accessible for Kiali


dtwilliamsWork:
### Describe the bug

We have a fresh install of Kiali v1.88.0 using the Operator and CR but the graphs will not load.

We have Istio v1.23 installed in Ambient mode with the Kubernetes Gateway API instead of the Istio Ingress and Egress gateways. Istio base, cni, istiod and ztunnel are all installed using the Helm chart with default values, apart from setting profile=ambient.

```
kc -n istio-system logs deployment.apps/kiali
2024/08/27 13:52:27 maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
2024-08-27T13:52:27Z INF Kiali: Version: v1.88.0, Commit: 8b2946fee5f39c6f26fde4599a274b80cc75d04b, Go: 1.22.4
2024-08-27T13:52:27Z INF Using authentication strategy [openid]
2024-08-27T13:52:27Z INF Some validation errors will be ignored [KIA1301]. If these errors do occur, they will still be logged. If you think the validation errors you see are incorrect, please report them to the Kiali team if you have not done so already and provide the details of your scenario. This will keep Kiali validations strong for the whole community.
2024-08-27T13:52:27Z INF Initializing Kiali Cache
2024-08-27T13:52:27Z INF [Kiali Cache] Waiting for cluster-scoped cache to sync
2024-08-27T13:52:27Z INF [Kiali Cache] Started
2024-08-27T13:52:27Z INF [Kiali Cache] Kube cache is active for cluster: [Kubernetes]
2024-08-27T13:52:27Z INF [Prom Cache] Enabled
2024-08-27T13:52:27Z INF Server endpoint will start at [:20001/]
2024-08-27T13:52:27Z INF Server endpoint will serve static content from [/opt/kiali/console]
2024-08-27T13:52:27Z INF Starting Metrics Server on [:9090]
2024-08-27T13:52:27Z INF Create jaeger GRPC client jaeger-instance-query.jaeger:16685
2024-08-27T13:56:12Z WRN Could not read the session: session not found: cookie kiali-token-Kubernetes does not exist in request
2024-08-27T13:56:13Z WRN No Grafana dashboard found for pattern 'Istio Control Plane Dashboard'
2024-08-27T13:56:13Z WRN No Grafana dashboard found for pattern 'Istio Wasm Extension Dashboard'
[...]

PR #7689, opened 2024-08-30
Fix graph appender problem

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7672

Fix a subtle problem in Appender processing in the appender.getTrafficClusters() utility function.  This function depended on the "isInaccessible" node metadata being set. But "isInaccessible" is set in the "Outsider" appender, a Finalizer which is run late in the appender processing, so appenders using this utility prior to the Outsider Finalizer being run, were at risk.

Update appender.getTrafficClusters() to not depend on that node metadata being set.

Updated test code reproduced the issue.


josunect (approved):
Instead of failure, now the cluster is recognized as a external cluster
image
If the name in the config is different. 
Code looks good, more easy to read also.

Tested also in MC: 

image

Issue #7687, opened 2024-08-30
filter system namespaces from namespace list


jmazzitelli:
### What do you want to improve?

I do not want to see all the system namespaces (such as `(kube-.*|openshift.*)`) in the Kiali UI (e.g. the Namespace drop down menus).

### What is the current behavior?

With the new discovery selectors (and the removal of the different filtering settings such as accessible namespaces and api.namespaces.exclude settings), those namespaces are not getting filtered out by default.

Workaround is to add them to the discovery selectors in a "NotIn" matchExpression.

### What is the new behavior?

Add some kind of setting to the Kiali CR so the system namespaces are hidden by default.

Before, these were filtered out by default: https://github.com/kiali/kiali-operator/blob/v1.89.2/crd-docs/cr/kiali.io_v1alpha1_kiali.yaml#L22-L26
```
      - "^istio-operator"
      - "^kube-.*"
      - "^openshift.*"
      - "^ibm.*"
      - "^kiali-operator"
```

We want them hidden by default now.

One way I think we can do it is to create a default set of discovery selectors such that all namespaces are selected except those above. Now, discovery selectors do not support regex. But we could add a setting `spec.deployment.discovery_selectors.ignore_system_namespaces` when, if true, the operator could add this to the discovery selectors:

```yaml
discovery_selectors:
  default:
  - matchExpressions:
    - key: kubernetes.io/metadata.name
      operator: NotIn
      values:
      - kiali-operator
      - kube-public
      - kube-system
      - ibm-system
      - istio-operator
      - openshift
      - openshift-apiserver
      - openshift-apiserver-operator
      (...and the rest)
```

The internal regex would be hardcoded and the operator just builds the list of namespaces itself and writes out that matchExpressions with the full list.

If we REALLY want to resurrect the idea of regex, we could just make `spec.deployment.discovery_selectors.ignore_system_namespaces` a string value containing a regex people could override. But I am hesitant to do that because it just resurrects a feature we wanted to get away with and just obsoleted. People should be using discovery selectors to filter out namespaces. The system namespaces won't change (its always kube-.* and openshift.* and ibm.*) so there is no reason to force people to use or know that regex since it will always be the same. If, for some oddball reason it isn't the same, well, they can just use discovery selectors normally and filter that way.

nrfox:
Can we just filter the list of namespaces that gets returned by the namespace service for these regexs inside the kiali server? Do we need to expose this as a config value?

jshaughn:
[...]

PR #7691, opened 2024-08-30
exclude system namespaces if no discovery selectors were defined

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7687

### DIFF
--- a/business/discovery_selectors.go
+++ b/business/discovery_selectors.go
@@ -1,6 +1,8 @@
 package business
 
 import (
+	"regexp"
+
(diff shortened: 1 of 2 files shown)


jmazzitelli:
before:
image

after:
image

jshaughn (approved):
This looks good. I started on minikube with just the usual things + bookinfo and all of the namespaces were visible. I added openshift-foo and openshft-foo and only the misspelled one appeared. I think created a DS that allowed only bookinfo and openshift-foo, and after reconciling that's all I saw, plus istio-system. Bully!

PR #7692, opened 2024-09-02
Use curl image from quay for P and Z

mkralik3:
### Describe the change

IBM clusters are not able to pull from the DockerHub, use quay instead


ScriptingShrimp (approved):
LGTM

PR #7696, opened 2024-09-04
[hack] disable ambient cni ipv6

jmazzitelli:
We need to disable ipv6 when installing Ambient because it breaks on minikube when not using docker driver (you get errors that say something like "Command error output: xtables parameter problem: ip6tables-restore: unable to initialize table 'nat'"

To test this:

1. Start minikube with some driver other than docker (our hack script defaults to kvm2, so this is a good test)
```
hack/k8s-minikube.sh start
```
2. Install Istio with Ambient using the hack script
```
hack/istio/install-istio-via-istioctl.sh -c kubectl -cp ambient
```

Istio with Ambient should install successfully now.

### DIFF
--- a/hack/istio/install-istio-via-istioctl.sh
+++ b/hack/istio/install-istio-via-istioctl.sh
@@ -359,6 +359,12 @@ if [ "${IS_OPENSHIFT}" == "true" ]; then
   fi
 fi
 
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
This is fine for now, although we may need to allow for ipv6 in the near-ish future and instead have an option to use the docker-driver, which also seems to solve the problem.

PR #7698, opened 2024-09-04
[1.73] Fix flaky int tests with tempo tracing

mkralik3:
### Describe the change
The Kiali Integration tests related to `Tracing` are sometimes flaky when Tempo tracing is used (since the older jaeger collector is used for ossm 2.6 / Kiali 1.73, so query to tempo tracing can take longer )
In the Kiali CR, I increased the query timeout to 60s, however, the test uses hardcoded 10s.

This PR increases client timeout to 60s for Tracing and Spans tests. ( to be aligned with downstream pipeline where integration with tempo is tested )

The backport is not needed since the newer Kali version supports Tempo querier, which is faster than the old and, in a little while, deprecated Jaeger querier.

Tested this change 3x against clusters: OCP on OpenStack, ARM, IPv6, ARO, ROSA, IBM P and all integration tests passed. ( so they are not flaky anymore )

### DIFF
--- a/tests/integration/utils/kiali/kiali_client.go
+++ b/tests/integration/utils/kiali/kiali_client.go
@@ -93,9 +93,10 @@ type MetricsJson struct {
 var client = *NewKialiClient()
 
 const (
-	BOOKINFO = "bookinfo"
-	ASSETS   = "tests/integration/assets"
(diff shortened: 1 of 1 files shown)


josunect (approved):
LGTM

Issue #7651, opened 2024-08-20
Issue querying Grafana Mimir for metrics


JDuskey:
### Describe the bug

When using Mimir as the backing datasource, we get lots of logs like this:
```
Error getting Prometheus version: invalid character 'p' after top-level value
```

It looks like this is caused by Kiali checking a `/version` endpoint that is as far as I can find scarcely documented. Mimir offers the same `/api/v1/status/buildinfo` that prometheus has documented, but there seems to be no way to force this to be used. Furthermore, if I could get it to be used, the format of the JSON is different than what is provided by `/version`, so it wouldn't be parseable. I originally added a PR to add this to the thanos_proxy configs similar to the other overrides that are already provided, but closed that PR out to have a discussion here instead.

Ideally this `/version` endpoint also isn't harcoded, but I'd like to at least use what is documented in the Prometheus project:
https://prometheus.io/docs/prometheus/latest/querying/api/#:~:text=GET%20/api/v1/status/buildinfo

### What are the steps to reproduce this bug?

1. Configure Kiali to use Mimir for metrics
2. Look at logs for Prometheus version errors

jmazzitelli:
Sounds like we could add some kind of setting in the Prometheus external_services to tell it where the version endpoint is. We have something similar for Istio -- see https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.istio.url_service_version

We could add something for prometheus... something similar to the health_check_url we already have: https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.prometheus.health_check_url

(BTW: @JDuskey did you set the xternal_services.prometheus.health_check_url? Maybe that does what you need already?)

nrfox:
I suspect that `/version` is an older endpoint and `/api/v1/status/buildinfo`  is "newer". I don't even see `/version` documented anywhere in the current docs. IMO we should just use `/api/v1/status/buildinfo`. 

https://prometheus.io/docs/prometheus/latest/querying/api/#build-information

> It looks like this is caused by Kiali checking a /version endpoint that is as far as I can find scarcely documented.

There's some info on what endpoints Kiali requires from prom [here](https://kiali.io/docs/configuration/p8s-jaeger-grafana/prometheus/#compatibility-with-prometheus-like-servers) but you're right it doesn't include `/version`  and I don't think it's exhaustive. We should update to the docs to spell out what all the endpoints are exactly.

jmazzitelli:
[...]

PR #7703, opened 2024-09-04
use prometheus buildinfo endpoint to get its version

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7651


jmazzitelli:
How to test:

Install everything, and make sure Kiali can talk to Prometheus.

Look at the Mesh page, click on Prometheus node, and check that you can see the version of your Prometheus - it should be a valid version string; it should not be `unknown` or empty string (`""`).

jshaughn (approved):
Seems to work:
image

image

Issue #7706, opened 2024-09-05
[Ambient] Add rates in double waypoint edges


josunect:
### What do you want to improve?
Add rates in double Waypoint edges

### What is the current behavior?
Combined edges for Waypoints but just one edge data is shown

### What is the new behavior?
Show the rates from both edges in the combined one. 
Show additional data in the summary panel.

Ref. https://github.com/kiali/kiali/pull/7593
Ref. https://github.com/kiali/kiali/issues/5476

PR #7719, opened 2024-09-09
 [Ambient] Add rates in double waypoint edges

josunect:
### Describe the change

- Add rates in double waypoint edges
- Include both edges information in the summary panel node
- Change TCP for arrows terminators. 
- Update legends. 
- Update waypoint node type for graph. 

image

(Just Patternfly Graph)

### Steps to test the PR

- `minikube start`
- Install kiali

### Automation testing

TODO

### Issue reference

Fixes https://github.com/kiali/kiali/issues/7706

TODO: 

- Add Traffic min/max for second edge


josunect:
Maybe it doesn't make sense to show hosts for each because they are reported the same: 

image

image

josunect:
Hello @fkargbo 

We are working on this approach of combining one edge. 
Here we would have an arrow from **ratings** to **waypoint** and another from **waypoint** to **ratings**. 
In order to clear the graph and avoid overlapping, we have this change where we combine the two arrows in just one, adding an edge termination in both ways: 

image

The side panel in the right, when you click on the edge, has information about both edges:

- ratings tab has the edge information from ratings to waypoint
- waypoint tab has the edge information from waypoint to ratings. 

I wonder if do you think it is clear or would you have any suggestion about how to improve it? 

Also, the edge label indicates a summary from the bytes sent/received, and includes a * to highlight it is a summary. Maybe we could just both values? I'm not sure if this is clear like it is. 

This is an example with a single edge (So you can compare): 
image

Also, is the edge termination a clear indicator about the direction? It would be a way to differentiate tcp/http traffic, this is an example with both types of traffic, but we wonder if the direction is clear:

image

frontend/src/pages/GraphPF/GraphPFElems.tsx
```diff
@@ -451,7 +457,7 @@ const getEdgeLabel = (edge: EdgeModel, nodeMap: NodeMap, settings: GraphPFSettin
     }
   }
 
-  let label = labels.join('\n');
+  let label = labels.join(' | ');
```
jshaughn, line 460: this is a decent alternative.

Note that the cytoscape graph doesn't have the double-labels for the bidirectional edges.  I guess this is OK, we are starting to diverge between the two graph impls.  I also noticed a highlighting issue in the cyto graph, related to waypoints (not a crash).

frontend/src/pages/GraphPF/GraphPFElems.tsx
```diff
@@ -508,14 +514,15 @@ const toFixedErrRate = (num: number): string => {
   return `${trimFixed(num.toFixed(num < 1 ? 1 : 0))}%err`;
 };
 
-const toFixedByteRate = (num: number, includeUnits: boolean): string => {
+const toFixedByteRate = (num: number, includeUnits: boolean, includeMarker?: string): string => {
```
jshaughn, line 517: This `includeMarker` param is unused given the commented code.  But it's fine if you think you may use it later.

frontend/src/types/Graph.ts
```diff
@@ -513,6 +512,10 @@ export interface DecoratedGraphEdgeData extends GraphEdgeData {
   tcp: number;
   // Default value NaN
   throughput: number;
+  // Waypoint from/to
+  waypoint?: string;
+  // Waypoint edge
+  waypointEdge?: DecoratedGraphEdgeData;
```
jshaughn, line 518: This seems like a circular definition, but I guess it works.  What is happening with this "waypointEdge" is pretty tricky, maybe we could use some comments somewhere about how we're doing bidirectional edges.

graph/config/cytoscape/cytoscape.go
```diff
@@ -471,6 +473,19 @@ func buildConfig(trafficMap graph.TrafficMap, nodes *[]*NodeWrapper, edges *[]*E
...
+		if e.Data.Waypoint == appender.WaypointTo {
+			for _, otherEdge := range *edges {
+				if e.Data.ID != otherEdge.Data.ID && e.Data.Source == otherEdge.Data.Target && e.Data.Target == otherEdge.Data.Source {
+					e.Data.WaypointEdge = otherEdge.Data
+					// Delete the other edge?
+				}
+			}
+		}
+	}
+
```
jshaughn, line 488: I'm not sure this is the best place for this logic.  In Cytoscape.go we are really just trying to translate the traffic map into the cytoscape json format (we use the same format for PFT, it's just that for Cy this format can be loaded automatically as a blob, and for PFT we have to parse it out ourselves).

Should this maybe be done in the ambient appender?

Is it necessary to have two separate Waypoint fields in the EdgeData?  Should we combine `EdgeData.Waypoint` and `EdgeData.WaypointEdge` into a single structure?  Something like:

```
type WaypointEdgeData struct {
	Direction string // to | from
	FromEdge *EdgeData // for a 'to' waypoint edge, this is the return 'from' edge
}

EdgeData.Waypoint: WaypointEdgeData

File: graph/telemetry/istio/istio.go [hunk 5]
```diff
@@ -386,11 +386,17 @@ func populateTrafficMap(trafficMap graph.TrafficMap, vector *model.Vector, metri
 }
 
 func addTraffic(trafficMap graph.TrafficMap, metric string, inject bool, val float64, protocol, code, flags, host, sourceCluster, sourceNs, sourceSvc, sourceWl, sourceApp, sourceVer, destCluster, destSvcNs, destSvcName, destWlNs, destWl, destApp, destVer string, sourceIsWaypoint, destIsWaypoint bool, o graph.TelemetryOptions) {
+	if sourceIsWaypoint {
+		sourceApp = ""
+	}
```
jshaughn, line 391:
```suggestion
        // waypoints are not apps, force it to be a workload regardless of graph type
	if sourceIsWaypoint {
		sourceApp = ""
	}
```

jshaughn (commented):
Josune, tested and things look good, but I do have some comments/questions.

frontend/src/pages/Graph/SummaryPanelCommon.tsx
```diff
@@ -199,7 +206,7 @@ export const renderNoTraffic = (protocol?: string): React.ReactNode => {
   );
 };
 
-export const getTitle = (title: string): React.ReactNode => {
+export const getTitle = (title: string, ambient?: React.ReactElement): React.ReactNode => {
```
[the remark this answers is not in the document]
josunect: Right, I will remove it.

frontend/src/pages/Graph/SummaryPanelEdge.tsx
```diff
@@ -169,95 +174,121 @@ export class SummaryPanelEdge extends React.Component<SummaryPanelPropType, Summ
...
       );
     };
+    const ambientBadge = waypoint ? <AmbientBadge tooltip={'reported from Ambient'}></AmbientBadge> : undefined;
+    const MainSummary = ({ waypointEdge }: { waypointEdge?: boolean }): React.ReactElement => {
+      const source = waypointEdge ? destData : sourceData;
+      const dest = waypointEdge ? sourceData : destData;
+      return (
+        <div>
+          <div className={panelHeadingStyle}>
+            {getTitle(`Edge (${prettyProtocol(protocol)})`, ambientBadge)}
```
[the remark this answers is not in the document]
josunect: Yes, I was thinking this was not very consistent, I will remove it. 
But, I will think it would be useful to add somehow information about where the telemetry was reported (Like ztunnel or waypoint?)

frontend/src/pages/Graph/SummaryPanelCommon.tsx
```diff
@@ -214,5 +221,22 @@ export const getTitle = (title: string): React.ReactNode => {
...
+  switch (node.nodeType) {
+    case NodeType.APP:
+      return node.app ? node.app : '';
+    case NodeType.SERVICE:
+      return node.service ? node.service : '';
+    case NodeType.WORKLOAD:
+      return node.workload ? node.workload : '';
+    default:
+      return '';
+  }
```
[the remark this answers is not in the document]
josunect: Right, I committed your suggestions, removed this one

frontend/src/pages/GraphPF/GraphPF.tsx
```diff
@@ -377,7 +377,11 @@ const TopologyContent: React.FC<{
           type: 'edge'
         };
         setEdgeOptions(edge, nodeMap, graphSettings);
-        edges.push(edge);
+        // If display is hide, don't push the node
+        // To avoid click in the hidden one
```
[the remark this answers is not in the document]
josunect: Updated

frontend/src/pages/GraphPF/GraphPFElems.tsx
```diff
@@ -414,6 +414,12 @@ const getEdgeLabel = (edge: EdgeModel, nodeMap: NodeMap, settings: GraphPFSettin
             break;
           case Protocol.TCP:
             labels.push(toFixedByteRate(rate, includeUnits));
+            if (data.waypoint === 'to' && data.waypointEdge) {
+              //labels.push(toFixedByteRate(rate + data.waypointEdge.tcp, includeUnits, '*'));
+              labels.push(toFixedByteRate(data.waypointEdge.tcp, includeUnits));
+            } //else {
+            //labels.push(toFixedByteRate(rate, includeUnits));
+            //}
```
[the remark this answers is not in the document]
josunect: Right, I was testing both options and I forgot. Removed now.

frontend/src/pages/GraphPF/GraphPFElems.tsx
josunect: Right, I just focused this changes in the PF graph. 

Regarding the highlighting issue, I think this is because now in PF the edge is not pushed, so it is easier to follow the path. But when there are two edges, there are more related nodes, and sometimes all the graph is highlighted, which might be confusing.

frontend/src/pages/GraphPF/GraphPFElems.tsx
josunect: This was used to show this * when the rates was the sum of the source and destination rates, as the rates is a marker, I added this additional parameter, but is not used anymore, removed.

graph/config/cytoscape/cytoscape.go
josunect: I have added this structure in the last commit. 

I will review about adding the logic in the Ambient appender. I agree is not the best place, but I've found some issues for adding it in the appender (I think it was related to that in the beginning I was using the ID to just link the edge, and it was not available until this point, but now that the node is attached, it might be possible).

josunect: Moved this logic to the Appender, there was the need to redo the Cytoscape edge conversion.

frontend/src/pages/Graph/SummaryPanelEdge.tsx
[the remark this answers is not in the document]
jshaughn: It may be useful, especially in a mixed proxy/ambient impl.  But if we are going to add that reporting we probably will have to add it all in one PR, and I think we can wait on it.

frontend/src/pages/GraphPF/GraphPFElems.tsx
jshaughn: I'm not going to worry too much about the Cy highlighting issue.

jshaughn (commented):
This is working and the changes are good, I think the logic is in the right places, etc.  Reviewing the code I thought it could be a simplified a little by using edge metadata that was closer to what we want to send back to the client.  I submitted a sub-pr because I thought it was easier than describing in comments.  See what you think.

jshaughn (approved):
Everything looks good, great improvements for ambient!  Thanks for the hard work on this one!

PR #7735, opened 2024-09-11
Certificate information per control plane in the Mesh page

### LINKED ISSUE #7456: Show certificate information per controlplane

Certificate information should be shown per controlplane instead from a single controlplane.

The certificates info dropdown only shows certificate information from a single controlplane. This isn't entirely accurate when there's multiple controlplanes either with revision based deployments and/or with multi-primary.

Screenshot from 2024-06-20 11-49-24

### What is the new behavior?

Certificates info is moved to the mesh page. When you click on a controlplane node the certificate information is shown for that particular controlplane.

leandroberetta:
Resolves #7456 

image


jmazzitelli:
I can see the cert info in the mesh page (note: I had to resolve the conflicts in the operator PR, but after that, this worked):

image

frontend/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "proxy": "http://localhost:20001",
```
jmazzitelli, line 2: IIRC, we don't commit changes to the proxy setting here

models/mesh.go
```diff
@@ -1,8 +1,13 @@
 package models
 
 import (
+	"time"
+
 	corev1 "k8s.io/api/core/v1"
 
+	"crypto/x509"
+	"encoding/pem"
```
jmazzitelli, line 9: trivial: the go core libs should go up top in the first group

```suggestion
	"crypto/x509"
	"encoding/pem"
	"time"

	corev1 "k8s.io/api/core/v1"
```

models/mesh.go
```diff
@@ -120,8 +125,34 @@ type ControlPlaneConfiguration struct {
...
-	DNSNames   []string `yaml:"dnsNames"`
-	SecretName string   `yaml:"secretName"`
+	DNSNames      []string  `yaml:"dnsNames"`
+	ConfigMapName string    `json:"configMapName"`
+	Issuer        string    `json:"issuer"`
+	NotBefore     time.Time `json:"notBefore"`
+	NotAfter      time.Time `json:"notAfter"`
+	Error         string    `json:"error"`
+	Accessible    bool      `json:"accessible"`
+	ClusterName   string    `json:"cluster"`
```
jmazzitelli, line 135: I noticed the existing settings have "yaml" in the struct tags. However, this new code uses "json". Seems odd that sometimes we have yaml and sometimes its json - we should be consistent and use the ones we need (or specify both).

jmazzitelli (changes requested):

frontend/package.json
leandroberetta: My bad...

models/mesh.go
leandroberetta: Well, I think the right one is json in this case, I fixed one tag what was still using yaml.

istio/discovery.go
```diff
@@ -83,6 +85,16 @@ func (in *Discovery) getControlPlaneConfiguration(kubeCache cache.KubeCache, con
 		return nil, err
 	}
 
+	certConfigMap, err := kubeCache.GetConfigMap(controlPlane.IstiodNamespace, certificatesConfigMapName)
+	if err != nil {
+		return nil, err
+	}
```
nrfox, line 91: I don't think a failure here should result in the whole function returning an error. Can we log an error or warning if `err != nil` and then parse if there's no error?

frontend/src/pages/Mesh/target/TargetPanelControlPlane.tsx
```diff
@@ -260,6 +264,20 @@ export class TargetPanelControlPlane extends React.Component<
...
+    return API.getIstioCertsInfo()
+      .then(response => {
+        this.setState({
+          certificates: response.data
+        });
+      })
+      .catch(error => {
+        AlertUtils.addError('Error fetching namespace certificates.', error, 'default', MessageType.ERROR);
+      });
+  };
```
nrfox, line 279: Since the certificates are already coming from `data.infraData.config.certificates`, do we need this API call / endpoint at all?

mesh/api/api_test.go
```diff
@@ -253,11 +279,13 @@ func TestMeshGraph(t *testing.T) {
...
+
 	expected, _ := os.ReadFile("testdata/test_mesh_graph.expected")
-	if runtime.GOOS == "windows" {
-		expected = bytes.Replace(expected, []byte("\r\n"), []byte("\n"), -1)
-	}
-	expected = expected[:len(expected)-1] // remove EOF byte
+	//if runtime.GOOS == "windows" {
+	//	expected = bytes.Replace(expected, []byte("\r\n"), []byte("\n"), -1)
+	//}
+	//expected = expected[:len(expected)-1] // remove EOF byte
```
nrfox, line 288: Does this need to be removed?

leandroberetta: Probably not, it was something I added at the beginning, then I found why it was failing, forgot to remove it. Thanks.

jmazzitelli (dismissed):

istio/discovery.go
```diff
@@ -83,6 +92,14 @@ func (in *Discovery) getControlPlaneConfiguration(kubeCache cache.KubeCache, con
 		return nil, err
 	}
 
+	certConfigMap, err := kubeCache.GetConfigMap(controlPlane.IstiodNamespace, certificatesConfigMapName)
+	if err != nil {
+		log.Warningf("Unable to get certificate configmap [%s/%s]. Err: %s", controlPlane.IstiodNamespace, certificatesConfigMapName, err)
+	} else {
+		cert, _ := parseIstioControlPlaneCertificate(certConfigMap)
```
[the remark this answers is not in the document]
leandroberetta: The potential error is set on a "error" field, which is then show in the UI if it happens (nicely, unable to decode or parse). I will remove the returned error, because is not necessary.

nrfox (approved):

PR #7736, opened 2024-09-11
Update bookinfo install hack script to have db services opt out of waypoint

jshaughn:
Update bookinfo install hack script such that for ambient the two "db" services opt out of the namespace waypoint.  This helps test L4-only traffic.


josunect (approved):
LGTM

image

PR #7738, opened 2024-09-12
Update express dependency (again)

ferhoyos:
### Describe the change

Update the express dependency to version 4.21.0 to remove version 0.18.0 of the `send` dependency.

Also, I have updated `path-to-regexp` to the latest 1.x release (version 1.9.0), instead of 8.0.0. Both versions meet our requirements, but I believe it is a bit risky to force an upgrade from version 1.x to 8.x.

### Steps to test the PR

Check that Kiali works as expected and CI tests pass.

### Automation testing

N/A

### Issue reference

https://issues.redhat.com/browse/OSSM-8124

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -159,10 +159,8 @@
     "typescript": "^4.9.5"
   },
   "resolutions": {
-    "express/path-to-regexp": "^0.1.10",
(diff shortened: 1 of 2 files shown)


aljesusg (approved):
LGFM

PR #7742, opened 2024-09-13
Prevent service node injection for waypoint edges

jshaughn:
Relates to #7445 

The main change is to prevent service injection for ztunnel waypoint edges (i.e. edges where the source or dest workload is the waypoint. For the reasoning see the github issue discussion.

This moves waypoint node identification into the base graph-gen code, performed only when the `ambient` appender is requested from the client.  That means the ambient appender does not need to do waypoint detection, it can rely on nodes already being identified.  And so the appender is concerned with handling the showWaypoint option and any other special handling that we may want to do (like bidirectional edge detection, etc).

Some other stuff in this PR:
- Fix: treat ambient as a finalizer appender, it was declared a finalizer but treated as a namespace appender
- Fix: Health appender was being used as both a namespace and a finalizer appender.  Should only execute once, as a finalizer.
- Fix: TrafficGeneratorAppender.Name() returned the wrong appender name
- Add: sliceutil.Some()
- Refactors
  - Move PromClient into AppenderGlobalInfo
    - I don't see a reason why we can't use the same prom client across the graph generation, this is more tidy/efficient
  - Rename AppenderGlobalInfo to just GlobalInfo, AppenderVendorInfo to just VendorInfo
    - This information can be used by appenders but also in the main graph gen.
    - More tidy and allows for wider use of Vendor info (like for storing waypoint workloads)
    - Apply the same change in MeshPage code, for consistency

TODO: Follow-up PR may be needed for other "injection-aware" appenders...

### DIFF
--- a/graph/telemetry/istio/appender/ambient_test.go
+++ b/graph/telemetry/istio/appender/ambient_test.go
@@ -5,8 +5,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	core_v1 "k8s.io/api/core/v1"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/kiali/kiali/business"
 	"github.com/kiali/kiali/config"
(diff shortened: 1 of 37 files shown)


graph/telemetry/istio/istio.go
```diff
@@ -64,9 +65,20 @@ func BuildNamespacesTrafficMap(ctx context.Context, o graph.TelemetryOptions, cl
...
 	trafficMap := graph.NewTrafficMap()
 
+	// To handle ambient telemetry we need to know about the configured waypoints. We typically avoid mixing
+	// config with telemetry because config is current and telemetry can be dated.  But for the moment
+	// it's unavoidable, as we need to do our best to identify source or dest waypoint workloads.  If in
+	// the future we can make that determination via the telem, we should change to that approach.
+	if sliceutil.Some(finalizers, func(f graph.Appender) bool {
+		return f.Name() == appender.AmbientAppenderName
+	}) {
+		waypoints := globalInfo.Business.Workload.GetWaypoints(context.Background())
```
josunect, line 75: The ctx context cannot be reused here?

graph/telemetry/istio/istio.go
```diff
@@ -966,3 +1008,36 @@ func promQuery(query string, queryTime time.Time, api prom_v1.API) model.Vector
...
+// hasWaypoint returns true if the source or dest workload is determined to be a waypoint workload.  Note that this logic can
+// go away if and when https://github.com/istio/ztunnel/issues/1128 is implemented, and then we can make this determination
+// directly from the telemetry
+func hasWaypoint(ztunnel bool, sourceCluster, sourceWlNs, srcWl, destCluster, destWlNs, destWl string, globalInfo *graph.GlobalInfo) (sourceIsWaypoint bool, destIsWaypoint bool) {
+	if !ztunnel {
+		return false, false
+	}
+
+	// first, try a simple substring check on the name, for a common waypoint giveaway
+	sourceIsWaypoint = strings.Contains(strings.ToLower(srcWl), "waypoint")
```
josunect, line 1021: I know this is an optimization but there is a small possibility of having a node containing waypoint but not being a waypoint?

graph/telemetry/istio/appender/ambient_test.go
```diff
@@ -192,50 +134,13 @@ func TestRemoveWaypoint(t *testing.T) {
 	assert.False(found)
 }
 
-func TestIsWaypoint(t *testing.T) {
```
josunect, line 195: I wonder if it would be possible to add some test coverage for the code that determines if is a waypoint (https://github.com/kiali/kiali/pull/7742/files#diff-611af3df8bfed322d3a0a785476482260509aa84540c2f8d70230b065ddf4820R1015)

josunect (commented):
I have made different tests and the code looks good: 

image

image

(It looks like for some reason there are some missing lockers here)

image

With idle edges: 

image

image

Service graph: 
image

App graph: 
image

Workload graph: 
image

I have added 3 comments, just to clarify, in the code.

josunect (approved):
LG!

graph/telemetry/istio/istio.go
jshaughn: Good catch, I will change this now.

graph/telemetry/istio/istio.go
jshaughn: I agree there is a very small risk.  Originally, I thought it was worth the quick test, but I'll just remove it and use the slightly longer and safer approach.

graph/telemetry/istio/appender/ambient_test.go
jshaughn: This is tested in an indirect way, in the new ambient test case in api_test.go.  That test will fail if for some reason we can't detect the waypoint workloads.

graph/telemetry/istio/istio.go
jshaughn: Funny, removing it revealed a problem in my new test setup :-)

josunect (approved):
Re tested again, LGTM!

PR #7746, opened 2024-09-13
obtain the grafana and tracing version from the internal in-cluster url

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7758

### DIFF
--- a/status/versions.go
+++ b/status/versions.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
@@ -80,16 +81,36 @@ func tracingVersion(conf *config.Config, homeClusterSAClient kubernetes.ClientIn
 	product.Name = string(tracingConfig.Provider)
 	product.Url = tracingConfig.URL
 
-	if product.Url != "" {
+	// we want to go to inClusterURL to obtain the version. If it isn't specified, fallback to the external URL.
+	versionUrl := tracingConfig.InClusterURL
+	if versionUrl == "" {
+		versionUrl = tracingConfig.URL
+	}
+
+	// there is no known way to get the version from GRPC. So we'll try to change the URL to go over HTTP,
+	// but this is not guaranteed to work. But it is worth a try.
+	if tracingConfig.UseGRPC {
+		parsedUrl, err := url.Parse(versionUrl)
+		if err == nil {
+			// strip the port - if the URL is http, it'll go over 80, if https, it'll go over 443
+			if host := parsedUrl.Hostname(); host != "" {
+				parsedUrl.Host = host
+				versionUrl = parsedUrl.String()
+				log.Debugf("Cannot get tracing version via GRPC; will try over HTTP: [%v]", versionUrl)
+			}
+		}
+	}
+
+	if versionUrl != "" {
 		// try to determine version by querying
 		if tracingConfig.Provider == config.JaegerProvider {
 			auth := tracingConfig.Auth
 			if auth.UseKialiToken {
 				auth.Token = homeClusterSAClient.GetToken()
 			}
-			body, statusCode, _, err := httputil.HttpGet(product.Url, &auth, 10*time.Second, nil, nil)
+			body, statusCode, _, err := httputil.HttpGet(versionUrl, &auth, 10*time.Second, nil, nil)
 			if err != nil || statusCode > 399 {
-				log.Infof("jaeger version check failed: url=[%v], code=[%v]", product.Url, statusCode)
+				log.Infof("jaeger version check failed: url=[%v], code=[%v], err=[%v]", versionUrl, statusCode, err)
 			} else {
 				// Jaeger does not provide api to get version, so it is obtained from js function inside html main page
 				// const JAEGER_VERSION = {"gitCommit: xxx, gitVersion: yyy, buildDate: zzz"}
@@ -108,9 +129,9 @@ func tracingVersion(conf *config.Config, homeClusterSAClient kubernetes.ClientIn
 		} else {
 			// Tempo
 			if tracingConfig.Provider == config.TempoProvider {
-				body, statusCode, _, err := httputil.HttpGet(fmt.Sprintf("%s/api/status/buildinfo", product.Url), nil, 10*time.Second, nil, nil)
+				body, statusCode, _, err := httputil.HttpGet(fmt.Sprintf("%s/api/status/buildinfo", versionUrl), nil, 10*time.Second, nil, nil)
 				if err != nil || statusCode > 399 {
-					log.Infof("tempo version check failed: url=[%v], code=[%v]", product.Url, statusCode)
+					log.Infof("tempo version check failed: url=[%v], code=[%v], err=[%v]", versionUrl, statusCode, err)
 				} else {
 					tempoV := new(tempoResponseVersion)
 					err = json.Unmarshal(body, &tempoV)
--- a/status/versions.go
+++ b/status/versions.go
@@ -139,18 +139,18 @@ func grafanaVersion(ctx context.Context, grafana *grafana.Service, grafanaConfig
 	product := models.ExternalServiceInfo{}
 	product.Name = "Grafana"
 	product.Url = grafana.URL(ctx)
-	if product.Url != "" {
-		// try to determine version by querying
-		url := fmt.Sprintf("%s/api/frontend/settings", product.Url)
+
+	versionUrl := grafana.VersionURL(ctx)
+	if versionUrl != "" {
 		// Be sure to copy config.Auth and not modify the existing
 		auth := grafanaConfig.Auth
 		if auth.UseKialiToken {
 			auth.Token = homeClusterSAClient.GetToken()
 		}
-		body, statusCode, _, err := httputil.HttpGet(url, &auth, 10*time.Second, nil, nil)
+		body, statusCode, _, err := httputil.HttpGet(versionUrl, &auth, 10*time.Second, nil, nil)
 
 		if err != nil || statusCode > 399 {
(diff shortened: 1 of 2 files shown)


jmazzitelli:
to test, install Istio and Grafana (just start up a minikube cluster and run `hack/istio/install_istio_via_istioctl.sh`).

Then build and push the dev images and install kiali (`make -e CLUSTER_TYPE=minikube build build-ui test cluster-push operator-create kiali-create`).

If you go to the Kiali UI Mesh page, and click the Grafana node, you will see "unknown" Grafana version - that's because it isn't configured yet.

Now edit the Kiali CR and set `external_services.grafana.in_cluster_url` to `http://grafana.istio-system:3000`. Set the url (`external_services.grafana.url`) to some bogus URL like `http://foo`. After you save the Kiali CR, wait for Kiali to restart and then go look at the Mesh page again in the UI and you should see the version now.

This shows that the version will be obtained by going over the in cluster URL. If you don't set that (but if you set the `url` to the valid endpoint) then it should still work -- the `url` is the fallback if `in_cluster_url` isn't set).

josunect (commented):
I was testing this in different scenarios, and I think it works as expected expect for the last one: 

in_cluster_url is set and valid, url incorrect. The version is set and there is access to Grafana: 
image

in_cluster_url is set and invalid, url is valid. The version is not set and there is no access to Grafana: 
image

in_cluster_url is not set, url is valid. The version is set and there is no access to Grafana: 
image

in_cluster_url is set to '', and url is set to a valid URL. There is access to Grafana but there is no version in Grafana. Shouldn't this fall back to use the url? 
image

Another question that I have is, should this behavior be the same for tracing and Prometheus urls?

jmazzitelli:
> Another question that I have is, should this behavior be the same for tracing and Prometheus urls?

Prometheus doesn't have an "in_cluster_url" - there's [only one URL setting](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.prometheus.url)

Yes, should be the same as tracing. Does this bug exist for tracing too?

jmazzitelli:
> in_cluster_url is set to '', and url is set to a valid URL. There is access to Grafana but there is no version in Grafana. Shouldn't this fall back to use the url?

I just tried it.

Set it using your test scenario:
```
$ kubectl get kiali kiali -n kiali-operator -ojsonpath='{.spec.external_services}'|jq
```
```json
{
  "grafana": {
    "in_cluster_url": "",
    "url": "http://grafana.istio-system:3000"
  }
}
```

I see Grafana version in the Mesh page, and the api/status shows it:

```
$ curl -k http://localhost:20001/kiali/api
{
...
  "externalServices": [
...
    {
      "name": "Grafana",
      "version": "11.0.0",
      "url": "http://grafana.istio-system:3000",
      "tempoConfig": {}
    }
  ],
```

Mesh page shows this:

image

jmazzitelli:
Looks like tracing has the same issue:
```
2024-09-18T14:13:40Z INF jaeger version check failed: url=[http://foo], code=[0]
```
That's with this:

```
$ kubectl get kiali kiali -n kiali-operator -ojsonpath='{.spec.external_services.tracing}'|jq
{
  "enabled": true,
  "in_cluster_url": "http://tracing.telemetry:16685/jaeger",
  "url": "http://foo",
  "use_grpc": true
}
```

jmazzitelli:
I created a new issue to include both tracing and grafana https://github.com/kiali/kiali/issues/7758
This PR will solve both issues. I need to update this PR to fix tracing, too.

josunect:
> > in_cluster_url is set to '', and url is set to a valid URL. There is access to Grafana but there is no version in Grafana. Shouldn't this fall back to use the url?
> 
> I just tried it.
> 
> Set it using your test scenario:
> 
> ```
> $ kubectl get kiali kiali -n kiali-operator -ojsonpath='{.spec.external_services}'|jq
> ```
> 
> ```json
> {
>   "grafana": {
>     "in_cluster_url": "",
>     "url": "http://grafana.istio-system:3000"
>   }
> }
> ```
> 
> I see Grafana version in the Mesh page, and the api/status shows it:
> 
> ```
> $ curl -k http://localhost:20001/kiali/api
> {
> ...
>   "externalServices": [
> ...
>     {
>       "name": "Grafana",
>       "version": "11.0.0",
>       "url": "http://grafana.istio-system:3000",
>       "tempoConfig": {}
>     }
>   ],
> ```
> 
> Mesh page shows this:
> 
> image

Sorry, @jmazzitelli , I meant to say the opposite (The screenshot is fine). `in_cluster_url` is valid, and `url` is set to "". Shouldn't this get the `in_cluster_url`? I understand it has precedence:

image

jmazzitelli:
> `in_cluster_url` is valid, and `url` is set to "". Shouldn't this get the `in_cluster_url`?

I think it is because of this code: https://github.com/kiali/kiali/blob/master/grafana/grafana.go#L210

This actually makes sense to me because in Kiali's Grafana integration (unlike our Tracing integration), you need an external public URL defined. Otherwise, it doesn't make sense. Because all the Grafana integration does is put external links in the Metrics pages for the user to click and go directly to the Grafana UI. So without this public `url` setting, the integration is basically a no-op -- it doesn't do anything.

This is different with our Tracing integration - we might not show the "Distributed Tracing" link in the left-hand nav, but we still provide tracing data directly within the Kiali UI (think of the bubble charts, or the purple highlights in the topology graph). So it is possible to not want to define a `url` for tracing, but still set `in_cluster_url`.

But without a `url` in Grafana, what does the Kiali UI provide?

Am I misunderstanding the Kiali Grafana integration? Do we actually have functionality in the Kiali UI where we simply query Grafana for data but do not need to show the user external links via that `url` ?

josunect:
> > `in_cluster_url` is valid, and `url` is set to "". Shouldn't this get the `in_cluster_url`?
> 
> I think it is because of this code: https://github.com/kiali/kiali/blob/master/grafana/grafana.go#L210
> 
> This actually makes sense to me because in Kiali's Grafana integration (unlike our Tracing integration), you need an external public URL defined. Otherwise, it doesn't make sense. Because all the Grafana integration does is put external links in the Metrics pages for the user to click and go directly to the Grafana UI. So without this public `url` setting, the integration is basically a no-op -- it doesn't do anything.
> 
> This is different with our Tracing integration - we might not show the "Distributed Tracing" link in the left-hand nav, but we still provide tracing data directly within the Kiali UI (think of the bubble charts, or the purple highlights in the topology graph). So it is possible to not want to define a `url` for tracing, but still set `in_cluster_url`.
> 
> But without a `url` in Grafana, what does the Kiali UI provide?
> 
> Am I misunderstanding the Kiali Grafana integration? Do we actually have functionality in the Kiali UI where we simply query Grafana for data but do not need to show the user external links via that `url` ?

@jmazzitelli yes, I think you are right about the Grafana and Kiali integration functionality. 

But, IMO, If checking the Grafana version will use the `in_cluster_url` first, I would expect this to be consistent and see the Grafana version if `in_cluster_url` is correct, independently of the `url` value. Because even if the version is "unknown", I don't think it warns the user that the `url` is wrong (I would expect to maybe see the Grafana component in red)

So, here, I don't see the Grafana version, and I see that Grafana is in green color, but there is an info message indicating that the links to Grafana are off:

image

It could be difficult to know what is wrong?

jmazzitelli:
OK, @josunect ... check out the latest commit. That should fix that use case where "url" is empty string but in_cluster_url is set correctly.

jmazzitelli:
Latest commit now obtains tracing version over `in_cluster_url` (falls back to `url` if in_cluster_url is not specified).

Note that also we can't get version if GRPC is enabled... today, when GRPC is enabled, you will see "unknown" for version. This new code now attempts to get the version over HTTP, even if GRPC is enabled. It just does a best guess for the URL to use (it simply removes the port number from the URL, with the hopes the tracing server has an HTTP endpoint over the default port (80 for http, 443 for https) and that the HTTP API is even enabled in the tracing server).

status/versions.go
```diff
@@ -80,16 +81,36 @@ func tracingVersion(conf *config.Config, homeClusterSAClient kubernetes.ClientIn
...
+		versionUrl = tracingConfig.URL
+	}
+
+	// there is no known way to get the version from GRPC. So we'll try to change the URL to go over HTTP,
+	// but this is not guaranteed to work. But it is worth a try.
+	if tracingConfig.UseGRPC {
+		parsedUrl, err := url.Parse(versionUrl)
+		if err == nil {
+			// strip the port - if the URL is http, it'll go over 80, if https, it'll go over 443
+			if host := parsedUrl.Hostname(); host != "" {
```
josunect, line 96: I would strip the port just for Jaeger, because Tempo is going to set the URL for the http client with the port also when using grpc (And then it wouldn't make sense to remove the port, IMO). Also, the Tempo operator has an option to setup grpc in the same port as http, `stream_over_http_enabled`, so http and GRPC would be using the same port and the version would be available. 
For example, in this configuration: 
```
      tracing:
        enabled: true
        grpc_port: 9095
        in_cluster_url: http://tempo-cr-query-frontend.tempo:3200
        provider: tempo
        url: 'http://tempo-cr-query-frontend.tempo:3200'
        use_grpc: true
```
With this is not possible to get the version: 
image

But the code that we have now makes it possible: 
image

josunect (commented):
I have done some tests and it looks good, I've just added one comment for tracing

status/versions.go
jmazzitelli: OK, check the latest commit. I just moved that block of code down into the "if-jaeger" section.

jmazzitelli:
> I have done some tests and it looks good, I've just added one comment for tracing

@josunect I modified the code based on that last comment you posted. See the latest commit.

josunect (approved):
LGTM!

image

image

PR #7747, opened 2024-09-16
[WIP] Support deploying dualStack Kind Clusters

sridhargaddam:
Part-of #7902 

This PR enables support for deploying KIND and Istio in dualStack mode using the flag IP_FAMILY. For a dualStack cluster, export IP_FAMILY=dual (default is ipv4).

### DIFF
--- a/hack/configs/default.yaml
+++ b/hack/configs/default.yaml
@@ -1,0 +1,32 @@
+# This configs KinD to spin up a k8s cluster with mixed protocol LB support and GRPCContainerProbe enabled
+# This should be used to create K8s clusters with versions >= 1.23
+# Source: https://github.com/istio/istio/blob/release-1.23/prow/config/default.yaml
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+kubeadmConfigPatches:
+- |
+  apiVersion: kubeadm.k8s.io/v1beta3
+  kind: ClusterConfiguration
+  metadata:
+    name: config
+  etcd:
+    local:
+      # Run etcd in a tmpfs (in RAM) for performance improvements
+      dataDir: /tmp/kind-cluster-etcd
+  # We run single node, drop leader election to reduce overhead
+  controllerManager:
+    extraArgs:
+      leader-elect: "false"
+  scheduler:
+    extraArgs:
+      leader-elect: "false"
+  apiServer:
+    extraArgs:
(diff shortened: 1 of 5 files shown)


nrfox:
Hey @sridhargaddam thanks for looking at this. I think it makes a lot of sense to use the same `kind-provisioner` script that istio and sail use. It would be nice to deploy both clusters into the same network since that would simplify the multi-cluster setup. iirc the `start-kind.sh` script doesn't support that today.

There are a couple of requirements that Kiali has which I'm not sure if that script supports:
1. oidc opts: https://github.com/kiali/kiali/blob/6a9061b6492eb439c5f7d64ef51021a2e01fd3c1/hack/start-kind.sh#L147C1-L154C62. Kiali supports login with oidc but in order to test this, you have to configure the kube API server on startup with these oidc flags. This is kind of a chicken and egg problem because the oidc server, keycloak, runs inside of one of the kind clusters meaning that the kind cluster needs to know the public ip of the keycloak server running inside of it when the kind cluster starts. To get around this we just assume that the first public ip allocated in the range we setup in metallb will be the keycloak server's public ip. The oidc server doesn't actually have to be online or resolvable when you start the kind cluster so normally you'd just use an FQDN instead of an ip address but we didn't want to use some external service like `nip.io` and now I forget why we didn't use something like `dnsmasq`. At any rate, we need to be able to pass the oidc flags to the kind cluster setup.
2. Local registry. We need the [local kind registry](https://kind.sigs.k8s.io/docs/user/local-registry/) setup for CI and local testing.

jmazzitelli:
`hack/configs` -- are these  KinD only configuration files? Please rename this to be more explicit what this is for. Right now someone sees "configs" under hack directory and it isn't clear what configuration files these are for.

i.e. `hack/kind-config-files` would be something more explicit.

jshaughn:
@jmazzitelli @nrfox , can we get together to figure out what we need to do here?  This PR has just sort of languished and I think it needs an assignee to take it over.

jmazzitelli:
I think we can close this - this PR gives us ipv6 in our kind setup script https://github.com/kiali/kiali/pull/8008

jshaughn:
> I think we can close this - this PR gives us ipv6 in our kind setup script

Agreed.  We may still want to reference some things here, but the PR does not need to move forward.

PR #7750, opened 2024-09-17
Update serialize-javascript dependency

ferhoyos:
### Describe the change

Update `serialize-javascript` dependency to version 6.0.2.

Instead of adding this version to `resolutions` section, I have updated the parent dependencies that import `serialize-javascript` as they were a bit outdated. This will improve the application's security."

Libraries updated:
- `@badeball/cypress-cucumber-preprocessor` to version `20.1.2`.  
- `workbox-webpack-plugin` to version `7.1.0`.

Breaking changes in the major version of both dependencies do not affect us:
- https://github.com/badeball/cypress-cucumber-preprocessor/releases?page=2
- https://github.com/GoogleChrome/workbox/releases

**Bonus:** 
- Align `seed-random` version in `package.json` to download only one version in the application.

### Steps to test the PR

Verify that CI tests pass.

### Issue reference

https://issues.redhat.com/browse/OSSM-8148

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -1,4 +1,5 @@
 {
+  "proxy": "http://localhost:20001",
   "name": "@kiali/kiali-ui",
   "version": "1.90.0",
   "description": "React UI for [Kiali](https://github.com/kiali/kiali).",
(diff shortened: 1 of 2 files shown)


aljesusg (approved):
LGfM

PR #7752, opened 2024-09-17
Show workload action list in kiosk mode

ferhoyos:
### Describe the change

In kiosk mode (used in OSSMC), the service and workload actions should be integrated into the minigraph toggle menu because the actions toggle button is not displayed. Currently, the service actions are integrated into the minigraph toggle menu, but the workload actions are not. This PR displays the workload actions list in kiosk mode.

image

**Bonus:**:
- Adapt cypress tests related to workload and service actions to OSSMC
- Add sucess message notification when a user deletes the Istio config of a service.

### Steps to test the PR

1. Navigate to any workload in the service mesh
1. Add `&kiosk=true` to the Kiali URL to access kiosk mode
2. Click on the minigraph toggle menu
3. Verify that workload actions are displayed (e.g., enable/disable workload injection) 

### Automation testing

Sidecar injections cypress tests have been adapted to OSSMC, which runs Kiali in kiosk mode under the hood

### Issue reference

https://github.com/kiali/openshift-servicemesh-plugin/issues/375


ferhoyos:
Thanks @jshaughn for the review. Let me try to answer all your questions:

> I don't understand is why we don't show the Actions menu in kiosk mode?
As far as I know, kiosk mode was mainly created because of OSSMC, and this OpenShift plugin can only display the Kiali tabs for workload/service detail pages. The standard OpenShift detail pages already have an Actions menu, which cannot be modified by the plugin. Therefore, we include the specific Service Mesh actions in a different location, like the minigraph menu toggle.

image

> I suspect that maybe Edit Annotations is not supposed to show up in Kiosk mode? 
I think `Edit Annotations` doesn't show up for the same reason as above. This option is displayed in the standard OpenShift console Actions menu, and it would be redundant to have a duplicate option.

> The Actions section showed up, but with no options
Good catch. I will hide the Actions section if there are no options to show.

jshaughn (approved):
> As far as I know, kiosk mode was mainly created because of OSSMC, and this OpenShift plugin can only display the Kiali tabs for workload/service detail pages. 

This is not entirely true.  Kiosk mode pre-dates the OSSM console plugin.  While OSSMC is certainly the biggest use case, we have had, and may still have unknown embedders of the UI.  For example, I know Rancher had us embedded in the past.  This is why there is the ability to special-case kiosk mode behavior based on the kiosk value (why `getKioskMode` returns a string).  To be honest, we may have messed this up over the last few years, assuming that kiosk = OCP console.  But really, I **think** that for a general kiosk user we shouldn't necessarily be taking away features just because it conflicts with OCP, and we should special case OSSMC kiosking, when necssary.  I'll approve this PR, but we may want to create an issue to revisit our general support for kiosk mode.

ferhoyos:
Thanks @jshaughn for approving this PR despite disagreeing with the current state of the kiosk mode. This PR just completes the work started [here](https://www.github.com/kiali/openshift-servicemesh-plugin/issues/43) and [here](https://github.com/kiali/openshift-servicemesh-plugin/issues/42).

The kiosk mode does not determine the layout to be displayed but rather the targetOrigin of the routing messages to be captured by the parent container (except when the kiosk value is `true`). For example is OSSMC all routing messages are intercepted by the plugin and modified to be redirected to other plugin pages.

```
export const isParentKiosk = (kiosk: string): boolean => {
  return kiosk.length > 0 && kiosk !== 'true';
};

// Message has no format, parent should parse it for its needs
const sendParentMessage = (msg: string): void => {
  // Kiosk parameter will send the parent target when kiosk !== "true"
  // this will enable parent communication
  const targetOrigin = store.getState().globalState.kiosk;
  if (isParentKiosk(targetOrigin)) {
    window.top?.postMessage(msg, targetOrigin);
  }
};
```

I did not know that Kiosk mode was created before the OSSM console plugin. In that case, maybe the Kiosk mode could be customized to show different layouts (e.g., hide only navigation menu, show only cards, etc.). Since the kiosk mode is used for the `targetOrigin` of the routing messages, the kiosk layout could be an optional UI feature in the Kiali configuration.

jshaughn:
>  In that case, maybe the Kiosk mode could be customized to show different layouts (e.g., hide only navigation menu, show only cards, etc.). 

I don't have any evidence that kiosk mode is still in use by external entities, or if it is that what we offer is a problem.  So, I don't think we need to do anything extra until someone comes to us with an RFE.

Issue #7745, opened 2024-09-13
change names of url / in_cluster_url to better reflect what they are


jmazzitelli:
### What do you want to improve?

The `url` and `in_cluster_url` setting for Grafana and Tracing are a bit confusing now (e.g. [here](https://github.com/kiali/kiali/issues/7717))

We should rename them for 2.0 release, so it makes more sense.

### What is the current behavior?

`url` was supposed to be only for displaying the links to the user so they can direct their browser to the Grafana and Jaeger UIs.
`in_cluster_url` was supposed to be the URL for Kiali to use when querying Grafana and Jaeger.

But in the case when your Grafana or Jaeger is outside of the cluster, the names don't make sense.

### What is the new behavior?

Rename `url` to something like `external_url`
Rename `in_cluster_url` to something like `internal_url`.

### PRs

* server: https://github.com/kiali/kiali/pull/7759
* operator: https://github.com/kiali/kiali-operator/pull/820
* kiali.io docs: https://github.com/kiali/kiali.io/pull/820

Test procedures can be found here: https://github.com/kiali/kiali/pull/7759#issuecomment-2359469926

jshaughn:
New names are OK although we should still support the old names as deprecated, right? At least for a while.  And what you described in the thread sounds like what is supposed to happen; Kiali uses in_cluster_url (internal_url) to do its work, and the url (external_url) to provide user links to the external UI.  If the in_cluster_url is not set we could try to use the url for our work, but I would expect it would not necessarily function well for API querying.  I think in general the in_cluster_url is supposed to be set, even if it's not in cluster :slightly_smiling_face: .

I'm not opposed to `internal_url and external_url`,  they make a nice set because the words go together, but just for completeness I'll throw out one other combo that came to mind: `api_url and ui_url`

jmazzitelli:
BTW: as part of this fix, we need to fix the version retrieval code for Grafana.. it is definitely using the `url` when it should be using `in_cluster_url` to get the version of Grafana.

jmazzitelli:
PRs:
* server: https://github.com/kiali/kiali/pull/7759
* operator: https://github.com/kiali/kiali-operator/pull/820
* kiali.io docs: https://github.com/kiali/kiali.io/pull/820

PR #7759, opened 2024-09-18
rename in_cluster_url to internal_url; rename url to external_url

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7745

this PR is rebased on the [version url fix PR](https://github.com/kiali/kiali/pull/7746) -- so that PR needs to be merged first before this one.


jmazzitelli:
To test, install the grafana and jaeger addons with istio (`hack/istio/install-istio-via-istioctl.sh` will do this for you).

Then make sure the Kiali CR has this:

```yaml
  external_services:
    grafana:
      internal_url: http://grafana.istio-system:3000
      external_url: http://foo.com
    tracing:
      internal_url: http://tracing.istio-system:16685/jaeger
      external_url: http://foo.com
      use_grpc: true
      enabled: true
```

hack/istio/sail/func-kiali.sh
```diff
@@ -126,11 +126,15 @@ spec:
   external_services:
     grafana:
       enabled: true
+      internal_url: "http://grafana.${CONTROL_PLANE_NAMESPACE}:3000"
+      external_url: "${grafana_external_url}"
       in_cluster_url: "http://grafana.${CONTROL_PLANE_NAMESPACE}:3000"
```
josunect, line 131: I wonder if these shouldn't be also removed?

jmazzitelli: I left those in there on purpose.. because right now this installs Kiali that comes from OperatorHub.io, which today is 1.89 and won't support this new renamed settings. So it still needs in_cluster_url/url in order to work.

When we release the next minor release of Kiali in OperatorHub.io, we can remove in_cluster_url / url here since that new version will support the renamed settings.

josunect (commented):
Thanks for the testing procedures, it works as expected: 

image

image

With the old settings: 
image

Also with Tempo: 
image

LG! I've just added a question in the code, mostly to confirm

josunect (approved):
LGTM!

PR #7763, opened 2024-09-20
[cypress] Add coverage for the summary panel when click on an edge in the graph

josunect:
### Describe the change

Add coverage for the summary panel when click on an edge in the graph

### Steps to test the PR

Run cypress test graph_side_panel. Also it is running in CI

### Automation testing

e2e

### Issue reference

Fixes #7762

PR #7765, opened 2024-09-23
[Ambiet] Update separator 

josunect:
### Describe the change
The separator was changed for all the labels in this PR: https://github.com/kiali/kiali/pull/7719 But it should just be changed for the waypoint edges

### Steps to test the PR

- `minikube start`
- Install kiali

### Automation testing

N/A (New e2e test added in https://github.com/kiali/kiali/pull/7763)

### Issue reference
Ref. #7706

### DIFF
--- a/frontend/src/pages/GraphPF/GraphPFElems.tsx
+++ b/frontend/src/pages/GraphPF/GraphPFElems.tsx
@@ -413,9 +413,14 @@ const getEdgeLabel = (edge: EdgeModel, nodeMap: NodeMap, settings: GraphPFSettin
             }
             break;
           case Protocol.TCP:
-            labels.push(toFixedByteRate(rate, includeUnits));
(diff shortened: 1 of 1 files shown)


jshaughn (approved):
@josunect , this is working and I will approve.  But as we talked about, the | separators actually don't look too bad.  If we want to keep them it's OK by me, but we need to find a way to separate the bidirectional edges.  I tried for a while to load the FA webfont, but can't get it to render.  I'm not sure if I'm doing something wrong of if the woff2 file I grabbed from patternfly has an issue.  Anyway, I'll leave it up to you.  One simple option is just to use a double-bar || to separate the two edges' labels.

jshaughn:
I think this PR is replaced by https://github.com/kiali/kiali/pull/7766?

josunect:
Right, I will close this one.

PR #7766, opened 2024-09-23
[ambient] Updated styles for waypoint edges summary panel

josunect:
### Describe the change

Use right left arrows to indicate the direction:
image

- Remove "Feature Under Development" For the Graph Display option "Waypoint"

### Steps to test the PR

- `minikube start`
- Install kiali
- Display `waypoint`

### Automation testing

### Issue reference

Ref. #7706

### DIFF
--- a/frontend/src/config/Icons.ts
+++ b/frontend/src/config/Icons.ts
@@ -32,6 +32,18 @@ export type IconType = {
 // or from the font awesome site: https://fontawesome.com/icons
 const mutIcons = {
   istio: {
+    arrowLeftRight: {
+      ascii: '\u21c4 ',
+      name: 'arrowLeftRight',
+      text: 'Arrow Left Right',
+      type: 'fa'
+    },
+    arrowLeftRightV: {
+      ascii: '\u21c5 ',
+      name: 'arrowLeftRightV',
+      text: 'Arrow Left Right Vertical',
+      type: 'fa'
+    },
     circuitBreaker: {
       ascii: '\uf0e7 ',
       className: 'fa fa-bolt',
--- a/frontend/src/config/Icons.ts
+++ b/frontend/src/config/Icons.ts
@@ -32,18 +32,6 @@ export type IconType = {
 // or from the font awesome site: https://fontawesome.com/icons
 const mutIcons = {
   istio: {
-    arrowLeftRight: {
-      ascii: '\u21c4 ',
-      name: 'arrowLeftRight',
-      text: 'Arrow Left Right',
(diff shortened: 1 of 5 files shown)


frontend/src/config/Icons.ts
```diff
@@ -32,6 +32,18 @@ export type IconType = {
...
+      name: 'arrowLeftRight',
+      text: 'Arrow Left Right',
+      type: 'fa'
+    },
+    arrowLeftRightV: {
+      ascii: '\u21c5 ',
+      name: 'arrowLeftRightV',
+      text: 'Arrow Left Right Vertical',
+      type: 'fa'
+    },
```
jshaughn, line 46: Since these aren't really icons and don't actually represent an Istio feature/component, maybe we can add a new section below, under this `istio` set of icons, that's dedicated to just unicode chars we want to use, that don't have an associated icon.  Also, let's use the exact naming as used on the site:

```
  },
  unicode: {
    // these are not fully defined icons, just unicode characters.
    // see https://symbl.cc/en/unicode-table/
    arrowRightOverLeft: {
      ascii: '\u21c4 ',
      name: 'RightwardsArrowOverLeftwardsArrow',
      text: 'Rightwards Arrow Over Leftwards Arrow'
    },
    arrowUpLeftofDown: {
      ascii: '\u21c5 ',
      name: 'UpwardsArrowLeftwardsofDownwardsArrow',
      text: 'Upwards Arrow Leftwards of Downwards Arrow'
    }
  }
```

jshaughn (commented):
@josunect This looks really good, nice finding all those working unicode characters!  I have one recommendation.  Also, the new button for toggling the side panel is nicer than the tabs.   Have you tried it in front of the From/To?  It may not look great but maybe as a comparison?

frontend/src/config/Icons.ts
josunect: I've done some tests and maybe is not really nice on this side: 

image

Maybe a bit closer?

image

I also added the arrows in the up/down direction so the icon is switched when you click on it, it makes more apparent to the user that things are changed:


[Screencast from 24-09-24 16:56:37.webm](https://github.com/user-attachments/assets/ce2fa486-4da5-4d93-ba5c-a27313935915)

jshaughn: I actually like it on the left :-)  Let's show it to Foday tomorrow and see if he has an opinion.  Either way, I like it better than tabs!

jshaughn (approved):
Looks Great!

PR #7761, opened 2024-09-19
[hack] support Sail on vanilla Kubernetes now that OperatorHub.io has the Sail operator

jmazzitelli:
These set of scripts will be able to install a full Istio-Kiali setup locally on a vanilla Kubernetes (not OpenShift).

You will get Sail operator and its Istio installation, Kiali operator and Kiali UI, along with Tempo operator and Tempo (for tracing backend and JaegerUI), Grafana, and Prometheus. The external services will have their UIs available assuming your k8s has a LoadBalancer (the steps below will give you that).

Here's how you can try it out.

For those that already have the kiali source repo git cloned locally and this PR branch checked out, here's a quick summary of the commands you can run to see everything... following this will be a more detailed set of instructions that even those that do not have the full source git cloned can run this stuff.

### Quick summary of the test steps

1. `hack/k8s-minikube.sh --load-balancer-addrs "70-84" start`
2. `hack/istio/sail/install-ossm-release.sh -c kubectl install-operators`
3. `sleep 30` # _(wait a little bit for the operators to start installing - just wait 30 seconds or so for OLM to start installing things)_
4. `hack/istio/sail/install-ossm-release.sh -c kubectl install-istio`
5. `hack/istio/install-bookinfo-demo.sh -tg -c kubectl -ail istio.io/rev=default-v1-23-0`
6. `hack/kiali-port-forward.sh`
7. Point your browser to Kiali UI at http://localhost:20001/kiali/console

Give it a minute or two for everything to start up and requests start flowing through the bookinfo traffic generator. Eventually you should see everything - bookinfo demo graph, you should be able to see traces, Jaeger UI , and Grafana UI links.

### More details on the test steps

1. Start minikube. You can put a loadBalancer in it if you want to try to access the Jaeger UI (but I haven't been able to get that to work yet.) You can use our hack script for this (not part of this PR) or just start minikube on your own. You can use KinD if you want, rather than minikube.
> If you want minikube (which is suggested for testing this stuff because this gives you a load balancer) - call k8s-minikube.sh with the arg ` --load-balancer-addrs "70-84"`. If you do not already have the entire kiali source repo git cloned locally, you can run this command so you just run the script without the need to clone everything):
```sh
bash <(curl -sk https://raw.githubusercontent.com/kiali/kiali/master/hack/k8s-minikube.sh) --load-balancer-addrs "70-84" start
```
> If you really want KinD (not sure how this works without a load balancer - might not be able to access grafana UI or jaeger UI, but other than that things should work)
[...]

PR #7768, opened 2024-09-24
Update tempo resources for Sail hack script

josunect:
### Describe the change

Update tempo resources for Sail hack script.

I think this resources specification are not enough for Tempo operator. This resources specification splits the resources between all the components, which is not enough for the Tempo query-frontend component. 

The tests can be done with the url and a simple query (With master code), for one minute: 

http://localhost:3200/api/search?end=1727106314&q={+.service.name+=+%22reviews.bookinfo%22}&start=1727106301

image

It takes around ~30 request to be OOM killed. 

In the query that we do, takes ~10 requests to be OOM killed (There is more data requested in order to create the chart): 

`http://localhost:3200/api/search?end=1727106314&limit=100&q={+.service.name+=+"reviews.bookinfo"++&&+.istio.mesh_id+=+"cluster.local"+++}+&&+{++}+|+select(status,+.service_name,+.node_id,+.component,+.upstream_cluster,+.http.method,+.response_flags,+resource.hostname)&spss=10&start=1727106301`

For reference: https://github.com/grafana/tempo-operator/issues/726 

* update readme based on this: https://github.com/kiali/kiali/pull/7761

### Steps to test the PR

* Tests using the updated README instructions (But in a kind cluster). 
* This is a first PR to have tracing stable. Second PR will be investigate Kiali and the use of Tempo (See https://github.com/kiali/kiali/issues/7769)

### Automation testing

### Issue reference

Ref. https://github.com/kiali/kiali/issues/7769

### DIFF
--- a/hack/istio/sail/README-RELEASE.md
+++ b/hack/istio/sail/README-RELEASE.md
@@ -1,5 +1,7 @@
 # Installing Istio with the Latest Sail and Kiali Operators
 
+## Quick installation guide
+
 To install Istio (including Kiali), use the script `install-ossm-release.sh`. This will utilize the Sail operator and Kiali operator to install the latest released images via OLM using either the Red Hat repository or the public Community repository or the public OperatorHub.io.
 
 Here's what you need to do in order to install everything.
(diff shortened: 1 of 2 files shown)


jmazzitelli (approved):
LGTM. I just want to make sure I understand what this PR is doing. Other than just updating the readme, this is removing the overrides to the resource limits simply to get back to a default Tempo installation. From here, another PR will be created later that will customize the Tempo installation with appropriate configuration for better performance.

josunect:
> LGTM. I just want to make sure I understand what this PR is doing. Other than just updating the readme, this is removing the overrides to the resource limits simply to get back to a default Tempo installation. 

Right, this is removing the overrides as I think this is not the optimal way to change the resource limits. 

>From here, another PR will be created later that will customize the Tempo installation with appropriate configuration for better performance.

I think It might not be needed to apply an specific configuration, you could have a test and see if you still see this flakiness (I haven't seen with the defaults). 

I think the next step should be focused in optimize the Kiali code, if needed.

josunect:
So it looks like the bottleneck in this config is the CPU. Here are some stats when Kiali runs this query:
`
http://localhost:3200/api/search?end=1727347481&limit=10&q=%7B+.service.name+%3D+%22reviews.bookinfo%22++%26%26+.istio.mesh_id+%3D+%22cluster.local%22+++%7D+%26%26+%7B++%7D+%7C+select%28status%2C+.service_name%2C+.node_id%2C+.component%2C+.upstream_cluster%2C+.http.method%2C+.response_flags%2C+resource.hostname%29&spss=10&start=1727346401`
x10 times

When trying to get the last 3h traces.

image

image

Issue #7751, opened 2024-09-17
Discovery selectors use case 


josunect:
### What do you want to improve?
The Kiali behavior when the discovery selectors doesn't include a namespace that is part of the mesh?

For example:
```
service mesh has namespaces A, B, C, istio-system.
I want Kiali instance X to only see A, istio-system.
I want Kiali instance Y to only see B, istio-system.
I want Kiali instance Z to only see C, istio-system.
```

### What is the current behavior?
In this example, bookinfo is part of the Mesh, but discovery selectors are not finding the namespace: 
```
deployment:
  discovery_selectors:
    default:
    - matchLabels:
        abc: def
        uvw: xyz
  view_only_mode: false
  cluster_wide_access: true
```
![image](https://github.com/user-attachments/assets/29c3f903-6d2f-4ad8-8a40-14c15b4de7d5)

![image](https://github.com/user-attachments/assets/d20eb06c-8f3d-450e-937c-197a3d7bb91b)

### What is the new behavior?
I think both views should be consistent, if user is able to see the bookinfo namespace in the Mesh view, probably also in the Overview, and not throwing an error.

jmazzitelli:
Ultimately, it comes down to this (we should come to a consensus for the answer to this question):

`What should the Kiali UI display if a namespace is part of the mesh but that namespace does not match a discovery selector?`

I was under the assumption the answer to that question is "Nothing". If a namespace doesn't match a discovery selector, it should look as if the namespace doesn't even exist. Nowhere in the Kiali UI should it be showing a namespace that doesn't match a discovery selector.

Is that how it really should be? Or are there edge cases where we still want to show the namespace (like here in that Mesh page)? I don't know the answer. Without thinking about it deeply, my guess is that my original assumption is correct - we should not show a namespace anywhere in the UI if it doesn't match a discovery selector. I'm curious to hear what other people think.

(BTW: ignore the control plane namespace (e.g. "istio-system") and the Kiali deployment namespace - those are always shown whether they have a discovery selector or not)

nrfox:
IMO this is a bug. If the namespace is excluded from the discovery selectors, why is it then shown on the mesh page? You'll end up with weird behavior like the screenshot above. Maybe we can somehow signal to the user that discovery selectors might be filtering the namespaces they are seeing on the mesh page though?

jmazzitelli:
> why is it then shown on the mesh page?

[...]

PR #7771, opened 2024-09-24
Mesh page: do not show namespaces that fail to match any discovery selector

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/7751


josunect (approved):
LGTM!
image

Namespace test is not in the mesh: 
image

Adding the ns: 
image

After removing the custom config: 

image

Issue #7713, opened 2024-09-06
Cypress tests for the Patternfly graph


ferhoyos:
### What do you want to improve?

e2e testing for the Patternfly graph

### What is the current behavior?

All Cypress tests for the graph are currently focused on the Cytoscape graph, not the PatternFly graph.

### What is the new behavior?

New Cypress tests should be created to verify that all test scenarios working with the Cytoscape graph also work with the PatternFly graph.

PR #7772, opened 2024-09-25
Adapt Cypress tests to the patternfly graph

ferhoyos:
### Describe the change

This PR updates all Cypress tests related to the graph so that they are tested for both `cytoscape` and `patternfly` graphs.

### Steps to test the PR

Verify that all CI tests pass.

### Issue reference

Closes https://github.com/kiali/kiali/issues/7713


jshaughn:
Thanks @ferhoyos for this initial work.  I don't have a problem with the "graphType" (maybe "graphImpl"?) approach, although I'm not sure it's the easiest path forward.  The separate files (like a big c+p) may be easier.  We will deprecate the cytoscape graph in 2.0 but we do want to ensure it is working, and so we do want the cypress testing.  But after we gain acceptance with the PFT graph, we will want to remove the cytoscape impl from the code base.  Knowing that, would separate files make it easier for removal?

On a different technical note, in the Mesh page cypress testing I eventually found a way to access the PFT graph controller, allowing us to inspect the graph's nodes and edges.  We do the same with accessing the CyRoot object in the cytoscape cypress tests (in this PR, you can see this in `graph_context_menu.ts`) .  In both places it works by accessing the currentstate of the page component.  For the PFT graph that means keeping it as a class impl, not a hook.  I could not find a way to access the hook state.  Anyway, you can see an example of accessing the PFT controller in `cypress/integration/common/mesh.ts`.

ferhoyos:
Thanks @jshaughn for your review! Having the cypress tests in the same file or duplicated ones depends on the requirements:
- Same file: Both implementations in the same method - easier to maintain because you only have to change one method but more difficult to dettach them
- Duplicated files: Different methods - Easier to remove one of them but more difficult to maintain (for each scenario you have to modify two files)

If the plan is to remove the cytoscape graph and tests in the medium term, then I would go with duplicated files. Also I will review the Mesh cypress test to figure out how to use the PFT graph controller

leandroberetta:
Hi, sorry for the delayed review, but I do agree on duplicate as we will eventually remove Cytoscape.

jshaughn:
I gave a thumbs-up above, but thinking about it I agree with the duplicated file approach.  I don't think we need to enhance the cytoscape tests after deprecation, and it will make it easier to remove.

ferhoyos:
@leandroberetta @jshaughn I have duplicated the Cytoscape and PatternFly graph tests. Since this PR is becoming large, I suggest splitting the task into two phases:
1. **Duplication of cypress tests:** `*-cy` for cytoscape and `*-pf` for patternfly. In this phase, the PatternFly tests will contain the same code as the Cytoscape tests. They will be temporarily tagged with `@pft` to prevent them from running in CI/CD pipelines.
2. **Fix patternfly graph cypress tests:** Here we can adopt a divide-and-conquer approach, creating multiple PRs to fix different Cypress tests. This will allow multiple contributors to work on it simultaneously.

I have completed Phase 1 in this PR. If you agree with my approach, please review and merge it. After that, additional PRs can be created to enable specific PatternFly Cypress tests by removing the `@pft` tag.

leandroberetta (approved):
Looks good to me, I would say to merge and start fixing tests.

Issue #7781, opened 2024-10-01
[pft graph]  Fix issue with bookmarked find/hide not being evaluated


jshaughn:
PFT graph. Define a graph find or hide expression. It should work fine.  Bookmark the page (or just do a full refresh).  The expression is populated in the toolbar, but does not get applied until the first refresh.

PR #7782, opened 2024-10-02
Fix issue with bookmarked find/hide not being evaluated in PFT graph or Mesh page

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7781

- Fix issue with bookmarked find/hide not being evaluated in PFT graph or Mesh page

@leandroberetta - This fix also provides the infrastructure for PFT cypress tests to access the graph in the same way as the Mesh Page cypress tests.

To Test:
1. Define a graph find or hide expression. It should work fine.
2. Do a full refresh of the Bookmarked page

Before: The expression is populated in the toolbar but does not get applied until the first refresh.
After: The expression is populated in the toolbar and gets applied immediately.


leandroberetta (approved):
Tested and works fine. Code looks good too.

PR #7783, opened 2024-10-02
Apply our own bendpoint logic to parallel edges

### LINKED ISSUE #7780: [pft graph] solve bendpoint issues with parallel edges
partof https://github.com/kiali/openshift-servicemesh-plugin/issues/99

With ambient we had an issue where there was a lot of parallel edge rendering.  Much of that has been resolved by the ambient work in #7445.  Although greater with ambient ztunnel reporting, there is always a chance to have multiple edges between the same source and dest node, when there is telemetry for different protocols.

Although this rendered fairly well with cytoscape (using arcs), PFT layouts [somewhat inexplicably] have a lot of trouble here. Most layouts don't handle it (just overlap the lines), and Dagre handles it with "bendpoints" (a series of connected lines).  The problem with bendpoints is that they have "drag" issues, where the bendpoints don't update.  I can do this using the PFT "surge demo, by dragging the group:

image

We already force straight edges (no bendpoints) for single edges.  But for multiple edges we still have this blocking problem.  We have reported the problem to PFT but likely need a workaround given our current timing.

One option is the idea of "combined edges", where we render a single edge to represent multiple edges.   You can see some new edge rendering ideas [here](https://github.com/kiali/kiali/pull/7693).  But past the rendering this has additional issues to solve, like edge selection and/or side panel rendering. edge labeling, animation behavior, etc.  This does have a benefit of being applicable to all of the layouts.

A second option is to try and workaround the bendpoint problem.  This would only solve the problem for Dagre but, if possible, would likely be less disuptive.

jshaughn:
Fixes #7780

Apply our own bendpoint logic to parallel edges.  Use simple bendpoints and update them when needed (after drag).

@josunect This is what I came up with to solve the parallel edge issue.  Read the [issue description](https://github.com/kiali/kiali/issues/7780) and it talks about the problem and two options.  This is the second option.  We could still look at option 1 in the future, but I ended up going with something that was a smaller overall change.  The coding was a little tricky, but it's not as big as trying to merge edges, both in implementation and in UX impact

### DIFF
--- a/frontend/src/pages/GraphPF/elements/extendedBaseEdge.ts
+++ b/frontend/src/pages/GraphPF/elements/extendedBaseEdge.ts
@@ -1,4 +1,4 @@
-import { BaseEdge, Point } from '@patternfly/react-topology';
+import { BaseEdge, Edge, Point } from '@patternfly/react-topology';
 
 // This extends BaseEdge to eliminate Bendpoints when possible (mainly because Dagre layout
 // does not provide an option to use or not use bendpoints). Bendpoints must be honored when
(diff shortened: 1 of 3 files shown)


josunect (approved):
I think it looks much better now! 

image

I see that when you move the Graph you still can see the sticky bendpoints, but they are adjusted again: 

[Screencast from 08-10-24 14:56:23.webm](https://github.com/user-attachments/assets/717e535e-4c6d-4741-be50-15f2afeb3456)

I see a minor thing, when different Display options are set, some of the labels can be overlapped: 
image

But I think it could be adresed later, if we want to improve it.

jshaughn:
> I see a minor thing, when different Display options are set, some of the labels can be overlapped:

Issue #7774, opened 2024-09-25
[Ambient] Support metrics for waypoint and ztunnel edges


josunect:
### What do you want to improve?
TCP charts not visualized because not enough data.
This happens with telemetry from ztunnel, (I think) specifically for the waypoint proxies. 

### What is the current behavior?
In some ztunnel edges, the TCP charts show "Not enough traffic to generate chart".
![image](https://github.com/user-attachments/assets/252cc3d2-b3ef-4a00-bf9c-3a84c70d2a8d)

The metrics query is looking for reporter=source:

`http://localhost:3000/api/namespaces/bookinfo/workloads/ratings-v1/metrics?byLabels[]=source_workload&direction=inbound&duration=300&filters[]=tcp_sent&filters[]=tcp_received&includeAmbient=false&quantiles[]=0.5&quantiles[]=0.95&quantiles[]=0.99&queryTime=1727273962&rateInterval=30s&reporter=source&step=30&clusterName=Kubernetes`

Returning: 
{"tcp_received":null,"tcp_sent":null}

![image](https://github.com/user-attachments/assets/24557300-8ee2-4c22-a45d-11e398031ae9)

![image](https://github.com/user-attachments/assets/506a5954-103d-40f9-a515-920af77ed36b)

### What is the new behavior?
Have data in the charts: 
![image](https://github.com/user-attachments/assets/794a27b4-5d56-4772-a91b-4a94166fa9b6)

For the double edges, some logic have been added here: https://github.com/kiali/kiali/pull/7719/files#diff-828be31fea9323742d8b78ef2e2e4254d5d38f4f4dc48c2ad962221a881c4e34R522

PR #7784, opened 2024-10-02
Fix sparkline issue for waypoint-edge side panel

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7774

- The problem here was the assumption that a 'from' waypoint edge was always the 'fromEdge' of a bidirectional edge. But there can be uni-directional edges both to and from waypoints.
- Part of the confusion may have been some overloading of 'waypoint' variable names.  This PR updates some var/type names to hopefully make things more clear. It also adds some stricter typing.
- Note that 'to' waypoint edges are 'source' reported, and 'from' waypoint edges are 'destination' reported.

### DIFF
--- a/frontend/src/pages/Graph/SummaryPanelEdge.tsx
+++ b/frontend/src/pages/Graph/SummaryPanelEdge.tsx
@@ -3,6 +3,7 @@ import { RateTableGrpc, RateTableHttp } from '../../components/SummaryPanel/Rate
 import { RequestChart, StreamChart } from '../../components/SummaryPanel/RpsChart';
 import { ResponseTimeChart, ResponseTimeUnit } from '../../components/SummaryPanel/ResponseTimeChart';
 import {
+  DecoratedGraphEdgeData,
   DecoratedGraphNodeData,
   GraphType,
   NodeType,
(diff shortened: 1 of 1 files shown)


josunect (approved):
LGTM! 

TCP metrics is available now: 
image
image

This is just a comment but, I've thought that the rates are the AVG for TCP sent, but, it doesn't look like the AVG is calculated here. So, maybe it is not the AVG or, are these rates calculated differently?

image

image

jshaughn:
> This is just a comment but, I've thought that the rates are the AVG for TCP sent, but, it doesn't look like the AVG is calculated here. So, maybe it is not the AVG or, are these rates calculated differently?

Thanks for the review, Josune!  For rates, whether it's HTTP or TCP, you are getting the range of actual rate data points, where each step is I think 30s.  So in that chart for "last 1m" you have 3 datapoints (t0, t30, t60).  The values are rough for shorter intervals, and smooth out for longer intervals.  But it's basically Prometheus' rate (or irate) function, applied to the available sample for the time period, trying to determine a per-second rate.  So, it's not necessarily average, bit rather a series of points.  Note that for response time we do calculate an avg, which is possible given the histogram.  But request rates are calculated based on a counter.

PR #7786, opened 2024-10-02
Set default graph impl to PFT

### LINKED ISSUE #7785: Make PFT Graph the default
Cytoscape will still be available by setting  `spec.kiali_feature_flags.ui_defaults.graph.impl` to either 'cy' or 'both'.

jshaughn:
Closes #7785 

- Set default graph impl to PFT
- also remove some old, hidden mesh page options
- maintains cytoscape in CI until we want to change the tests to PFT (WIP)
- fix an issue in the cy/pft test split
- also, this fixes test flakiness introduced in the recent PR to fix find/hide, as well as add support for PFT cypress testing.  That change slightly delayed when the PFT controller was made available to the test code, which could cause a flake when the test tried to access it prematurely.  This should ensure that the controller is available prior to the test trying to use it.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -823,7 +823,7 @@ func NewConfig() (c *Config) {
 							Expression:  "rank > 2",
 						},
 					},
-					Impl: "cy",
+					Impl: "pf",
 					Settings: GraphSettings{
 						FontLabel:    13,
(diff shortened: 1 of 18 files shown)


leandroberetta (approved):
LGTM.

Issue #7598, opened 2024-08-06
Mesh page - controlplanes not mapped to their dataplanes when using stable revision labels


nrfox:
### What do you want to improve?

Kiali's support for [stable revision labels](https://istio.io/latest/docs/setup/upgrade/canary/#stable-revision-labels).

### What is the current behavior?

When you use stable revision labels, on the mesh page Kiali does not show dataplanes associated with their respective controlplanes.

### What is the new behavior?

On the mesh page, dataplanes are mapped correctly to their controlplanes even when using stable revision labels.

PR #7787, opened 2024-10-03
Improve tag support on the mesh page

### LINKED ISSUE #7726: Canary upgrade status only uses home cluster

Canary upgrade status only uses the home cluster:
https://github.com/kiali/kiali/blob/153dd90d4a56a5c463c59720680efa2804d13aef/business/mesh.go#L82

### What is the new behavior?

Control plane status should be show across clusters.

nrfox:
### Describe the change

Improves tag support on the mesh page by:

1. Adding support for multi-primary status
2. Showing tags as a secondary label on the controlplane node
3. Improving the wording / styling of the donut chart on the controlplane namespace target panel
4. Removing the metrics chart and adding controlplane info to the controlplane namespace target panel
5. Reducing number of requests. `/api/namespaces` was being called repeatedly when you interact with different elements of the mesh page but this can come from the redux store.
6. Removing dead/commented out code in parts of the mesh page components.

### Steps to test the PR

1. Deploy Kiali

    **Deploying with helm/kind:**
  
    Build kiali server and UI then run setup.
    ```
    make build build-ui
    ./hack/run-integration-tests.sh --test-suite backend --setup-only true
    ```
    **Deploying with minikube/operator:**
    ```
    hack/k8s-minikube.sh start 
    hack/k8s-minikube.sh istio
    hack/k8s-minikube.sh bookinfo
    make CLUSTER_TYPE=minikube build build-ui cluster-push operator-create kiali-create
    hack/istio/install-testing-demos.sh -c kubectl
    ```
    
2. Create two additional controlplanes
    ```
    istioctl install --revision=1-21-1 --set profile=minimal --set values.global.multiCluster.clusterName="cluster-default" --skip-confirmation                                                           
    istioctl install --revision=1-22-3 --set profile=minimal --set values.global.multiCluster.clusterName="cluster-default" --skip-confirmation
    ```
3. Create a tag for each
    ```
    istioctl tag set prod-stable --revision 1-21-1
    istioctl tag set prod-canary --revision 1-22-3
    ```
4. Create test namespaces and label each
    ```
    kubectl create ns app-ns-1
    kubectl label ns app-ns-1 istio.io/rev=prod-stable
    kubectl create ns app-ns-2
    kubectl label ns app-ns-2 istio.io/rev=prod-stable
    kubectl create ns app-ns-3
    kubectl label ns app-ns-3 istio.io/rev=prod-canary
    ```
5. Create an app in each namespace
    ```
    kubectl apply -n app-ns-1 -f https://raw.githubusercontent.com/istio/istio/master/samples/sleep/sleep.yaml
    kubectl apply -n app-ns-2 -f https://raw.githubusercontent.com/istio/istio/master/samples/sleep/sleep.yaml
    kubectl apply -n app-ns-3 -f https://raw.githubusercontent.com/istio/istio/master/samples/sleep/sleep.yaml
    ```
7. Visit the mesh page and verify that each controlplane has a tag label underneath it and the side panel has controlplane information in it including tags.

    Screenshot from 2024-10-02 09-35-00
    Screenshot from 2024-10-02 09-38-51

8. Overview page should have a badge for every namespace that has an injection label of any kind including `istio-injection: enabled`:
    Screenshot from 2024-10-10 14-25-24

### Automation testing

N/A

### Issue reference
 
Relates to https://github.com/kiali/kiali/issues/7598

Fixes #7726


frontend/src/store/Store.ts
```diff
@@ -48,9 +48,15 @@ export interface NamespaceState {
...
   readonly isFetching: boolean;
+  // TODO: Can this be non-optional since we default to empty list?
   readonly items?: Namespace[];
   readonly lastUpdated?: Date;
   readonly namespacesPerCluster: Map<string, string[]>;
+  // The items list filters out namespaces with the same name that exist
+  // on multiple clusters because that is how the namespace selector
+  // currently works but these are actually separate namespaces and
+  // some parts of the UI need the full list unfiltered.
+  readonly unfilteredItems?: Namespace[];
```
jshaughn, line 59: I'm not sure I love this `unfilteredItems` addition.  Not because you don't need it, but because this whole NamespaceState thing is getting pretty confusing.  Minimally, the fields could use some renaming and comments.  Without studying code how are we supposed to know what 'items' is, or that it's some special filtered list.  Oh, and it's not actually filtered by the `filter` field above :-)  We'd be best off by removing the derived state, like `items` and `namespacesPerCluster`, but the danger is that we may have components off of those state changes.  So, I think we should live with them.  But, what do you think of doing some renaming and commenting here?  Maybe something like:
```

Side note:  While looking at this, I think maybe we have a bug in `EnvoyDetails`.  It seems to be using `items` but I think it may actually want `unfilteredItems`
activeNamespaces  // the namespaces currently selected in NamespaceDropdown
filter  // the filter currently defined in NamespaceDropdown
isFetching
namespaces  // formerly 'unfilteredItems', all namespaces returned from API.getNamespaces
namespaceNames // formerly 'items', Set of unique namespace names, for convenience
lastUpdated
namespacesPerCluster: // map clusterName -> namespaces on cluster
```

frontend/src/types/Namespace.ts
```diff
@@ -4,6 +4,7 @@ export interface Namespace {
   isAmbient?: boolean;
   labels?: { [key: string]: string };
   annotations?: { [key: string]: string };
+  revision?: string;
```
jshaughn, line 7: Let's alphabetize this touched type.

mesh/generator/generator.go
```diff
@@ -98,6 +98,12 @@ func BuildMeshMap(ctx context.Context, o mesh.Options, gi *mesh.GlobalInfo) (mes
 			"revision":   cp.Revision,
 			"thresholds": cp.Thresholds,
 		}
+		// Assuming there's just one tag per controlplane.
+		// TODO: Is this correct? Especially for primary-remote?
```
jshaughn, line 102: I don't know :-)

models/namespace.go
```diff
@@ -69,6 +76,18 @@ func CastNamespace(ns core_v1.Namespace, cluster string) Namespace {
 	namespace.Labels = ns.Labels
 	namespace.Annotations = ns.Annotations
 
+	if ns.Labels != nil {
+		// Injection label takes precedence.
+		// See rules at: https://istio.io/latest/docs/setup/additional-setup/sidecar-injection/#controlling-the-injection-policy
```
jshaughn, line 81: note sure if there is anything ambient-related here.

frontend/src/pages/Mesh/target/TargetPanelControlPlane.tsx
```diff
@@ -39,14 +36,22 @@ import { load, dump } from 'js-yaml';
...
+import { namespaceItemsSelector } from 'store/Selectors';
+import { Namespace } from 'types/Namespace';
+import { connect } from 'react-redux';
+import { KialiAppState } from '../../../store/Store';
 
-type TargetPanelControlPlaneProps = TargetPanelCommonProps & {
-  meshStatus: string;
-  minTLS: string;
+type ReduxProps = {
+  namespaces: Namespace[];
```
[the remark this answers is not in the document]
nrfox: It does actually stay in sync but somewhat tangentially. There's a component in the masthead (`IstioStatus`) that is hooked into the refresh button so when a new namespace is added with the target namespace panel open and you hit the refresh button or the refresh timer ticks, the `IstioStatus` component will update the redux namespace store causing this panel to be updated.

The namespace count that you see on the dataplane node comes from the graph info so the two can get temporarily out of sync. Since this component is tied in with the graph, the namespace info should probably come from the graph also. I'll either plumb it through as props or pull it from the graph response and add it to the redux store.

frontend/src/store/Store.ts
nrfox: I'll change where the namespace info comes from so these changes won't be needed anymore.

models/namespace.go
nrfox: This is just the revision so I don't think so

frontend/src/pages/Mesh/target/TargetPanelControlPlane.tsx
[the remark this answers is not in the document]
jshaughn: OK, that's interesting.  The Namespace Dropdown has to be up to date, so in that case we ensure an update outside of the standard refresh/status plumbing.  The graph updates at refresh time, so the graph should basically be in sync with refresh, so not sure there needs to be any further sync work but do what you think is needed to ensure namespace changes are reflected.

frontend/src/store/Store.ts
jshaughn: OK, but whatever is still here, please add a few comments to capture the understanding.

frontend/src/pages/Mesh/target/TargetPanelControlPlane.tsx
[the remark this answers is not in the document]
nrfox: Alright the namespaces actually come from the graph response. They are included in the `infraData` of the controlplane nodes. When the graph request returns, the frontend saves the controlplanes in the redux state. This way the namespaces that get displayed on the graph will always stay in sync with the namespaces on the target panel.

frontend/src/types/Mesh.ts
```diff
@@ -92,6 +92,29 @@ export interface MeshDefinition {
   timestamp: number;
 }
 
+export interface Tag {
+  name: string;
+}
+
+export interface ControlPlaneConfig {
+  OutboundTrafficPolicy?: any;
+  configMap?: any;
```
jshaughn, line 101:
```suggestion
  configMap?: any;
  outboundTrafficPolicy?: any;
```
alphabetization and capitalization

frontend/src/store/Store.ts
```diff
@@ -134,6 +139,11 @@ export interface MeshToolbarState {
 }
 
 export interface MeshState {
+  // controlPlanes are pulled from the graph response and are used
+  // by other elements on the graph page. They are in the redux
+  // store to so that the other elements can stay in sync with
+  // with graph and we don't have to duplicate requests to get this info.
+  controlPlanes?: ControlPlane[];
```
jshaughn, line 146: I'm not 100% sure it's necessary to increase redux state with the controlPlanes.  Are they really needed to trigger a state change?  Could they not be parsed out of the Mesh elements when needed?

mesh/api/testdata/test_mesh_graph.expected
```diff
@@ -268,45 +270,142 @@
...
+                  "version": ""
+                }
+              ],
+              "name": "cluster-primary",
+              "secretName": "",
+              "accessible": true
+            },
             "config": {
-              "ConfigMap": {
+              "configMap": {
```
jshaughn, line 290: intentional case change?

frontend/src/pages/Mesh/MeshElems.tsx
```diff
@@ -146,6 +146,9 @@ export const setNodeLabel = (node: NodeModel, _nodeMap: NodeMap): void => {
         break;
       case MeshInfraType.ISTIOD:
         pfBadge = PFBadges.Istio;
+        if (data.infraData && data.infraData.tag) {
```
jshaughn, line 149:
```suggestion
        if (data.infraData?.tag) {
```

models/mesh.go
```diff
@@ -40,77 +40,75 @@ type Mesh struct {
...
 
 	// Status is the status of the controlplane as reported by kiali.
 	// It includes the deployment status and whether kiali can connect
 	// to the controlplane or not.
-	Status string
+	Status string `json:"status"`
 
 	// Tags are the tags associated with the controlplane.
-	Tags []Tag
+	Tag *Tag `json:"tag"`
```
jshaughn, line 99:
```suggestion
	Tag *Tag `json:"tag,omitempty"`
```

I think this and several others could probably have have omitempty, and save us from nulls in the json.

mesh/api/testdata/test_mesh_graph.expected
nrfox: Yes because I added a tag for this field on the backend to lowercase it. Probably need to do the same for the rest of them to be consistent.

frontend/src/store/Store.ts
jshaughn: Not really, from an element (the selected control plane node) you can always work "up" to the controller or graph.   You could do something like:
```
const controlPlanes: ControlPlane[] | undefined = 
    this.state.controlPlaneNode?.getGraph().getNodes()?
        .filter(node => node.getData().infraType === MeshInfraType.ISTIOD)
        .map(node => node.getData().infraData);
 ```

**Review by jshaughn — COMMENTED** [review]
To summarize our offline conversation while I was testing.  Things worked well when I configured things correctly.  But when I set the wrong cluster name on a control plane revision, it wasn't flagged in any way, but did mess things up.  A log error would be good if we could detect the issue.  Maybe a degraded health would even make sense.
If defined, I'd like to see the Tag in the controlplane side panel, probably just above `Version`.
Finally, if we can get away with not putting the control planes in redux I think that would be good.  But if it makes things hard then we'll leave it.  See my latest inline comment.  Should be on track for a merge tomorrow.

File: frontend/src/store/Store.ts (hunk 7 above)
Comment (nrfox) [reply frontend/src/store/Store.ts:146]: Ah that's good to know. I'll give that a shot.

**Review by jshaughn — APPROVED** [review]
Code looks good.  Thanks for removing the redux stuff, cleaner now I think.  Now seeing the log warning, nice!
```
2024-10-18T18:32:41Z WRN The controlplane [istio-system/istiod-bad-cluster] cluster name ['cluster-default'] does not match the cluster ['Kubernetes'] where it is deployed. This is likely a misconfiguration. Check your values.global.multiCluster.clusterName' setting on your controlplane or check your Kiali configuration setting 'kubernetes_config.cluster_name'.
```

Mesh page and CPNamespace side panel donut looks good:
image

CP Side panel looks good with new Tag being displayed.
image


Good stuff!

Issue #7770, opened 2024-09-24
PF node graph side panel snaps back to target node on refresh


jshaughn:
1. PF Graph
2. Go to node graph
3. see side panel is set to selected target node
4. select an edge or different node (click on it)
5. see side panel change as expected
6. refresh, or wait for auto-refresh, side panel snaps back

It should remain showing the selected element

PR #7788, opened 2024-10-03
[pft graph] fix node-graph selection snap-back issue

jshaughn:
Fixes https://github.com/kiali/kiali/issues/7770

Only pre-select node-graph node when the graph data has actually changed (i.e. for a new node graph, or when there is a structural change on refresh).

- also, downgrade some logging from debug to trace

Test:
1. Go to node graph
2. see side panel is set to selected target node
3. select an edge or different node (click on it)
4. see side panel change as expected
5. refresh, or wait for auto-refresh

- Before:
  - side panel snaps back to select the original node-graph node
- After:
  - It should remain showing the selected element


leandroberetta (approved):
Looks good and works as expected.

PR #7790, opened 2024-10-03
[hack] update dex

jmazzitelli:
Update to Dex 2.41.1.

The reason for this is because currently, and for some unknown reason, the oauth proxy that gets started now hits this error (it did NOT get this error two days ago):

```
$ kubectl logs -n oauth2-proxy  deployments/oauth2-proxy 
[2024/10/03 20:09:10] [provider.go:55] Performing OIDC Discovery...
[2024/10/03 20:09:10] [main.go:59] ERROR: Failed to initialise OAuth2 Proxy: error initialising provider: could not create provider data: error building OIDC ProviderVerifier: could not get verifier builder: error while discovery OIDC configuration: failed to discover OIDC configuration: error performing request: Get "https://192-168-39-217.nip.io:32000/.well-known/openid-configuration": tls: failed to verify certificate: x509: certificate signed by unknown authority
```

Not sure what happened, but it looks like the root cert that is used is no longer valid.

This causes the header-auth-test molecule test to fail (simply because the oauth proxy in front of Dex is failing to start).

_UPDATE: Something broke in oauth2-proxy - known issue as of today. See https://github.com/oauth2-proxy/oauth2-proxy/issues/2802_

### DIFF
--- a/hack/k8s-minikube.sh
+++ b/hack/k8s-minikube.sh
@@ -24,7 +24,7 @@ set -u
 DEFAULT_CLIENT_EXE="kubectl"
 DEFAULT_DEX_ENABLED="false"
 DEFAULT_DEX_REPO="https://github.com/dexidp/dex"
-DEFAULT_DEX_VERSION="v2.30.2"
+DEFAULT_DEX_VERSION="v2.41.1"
 DEFAULT_DEX_USER_NAMESPACES="bookinfo"
 DEFAULT_INSECURE_REGISTRY_IP=""
 DEFAULT_K8S_CNI="auto"
@@ -165,10 +165,6 @@ EOF
 <   replicas: 3
 ---
 >   replicas: 1
-26c26
-<       - image: dexidp/dex:v2.27.0 #or quay.io/dexidp/dex:v2.26.0
----
->       - image: docker.io/dexidp/dex:${DEX_VERSION}
 41c41
 <         - name: GITHUB_CLIENT_ID
 ---
@@ -200,7 +196,7 @@ EOF
 <         org: kubernetes
 94a81
 >       responseTypes: ["code", "id_token"]
-96a84,89
+101a89,94
 >     - id: kiali-app
 >       redirectURIs:
 >       - 'http://${MINIKUBE_IP}/kiali'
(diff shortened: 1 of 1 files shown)


jmazzitelli:
The code in this PR as it sits right now still doesn't work; oauth proxy still fails to start.

jmazzitelli:
`image: quay.io/oauth2-proxy/oauth2-proxy:latest`

Here is something that is possibly the source of the problem. The OAuthProxy pod is pulling in "latest" - I suspect (though I haven't checked yet) that some change went into the proxy recently that changed behavior.

Note that I see in the oauth proxy ConfigMap that this setting should be ignoring the error: `ssl_insecure_skip_verify = true` -- perhaps the setting name changed or it is no longer supported?

jmazzitelli:
https://quay.io/repository/oauth2-proxy/oauth2-proxy?tab=tags

"latest" was updated on Oct 2 (right about the time when our errors started occurring). Last update was 8 months ago. So, I suspect this latest oauth proxy update is what broke us.

jmazzitelli:
If we want to take the easy way out, we can simply change the oauth proxy image to be v7.6.0 rather than latest - I just tried that and it works again (which makes sense, because that was the previous version - "latest" which was just released and is v7.7.0 seems to break us for whatever reason).

```diff
diff --git a/hack/k8s-minikube.sh b/hack/k8s-minikube.sh
index acb0517e8..6a3721af8 100755
--- a/hack/k8s-minikube.sh
+++ b/hack/k8s-minikube.sh
@@ -260,7 +260,7 @@ spec:
         - --config
         - /etc/oauthproxy/oauth2-proxy.conf
         env: []
-        image: quay.io/oauth2-proxy/oauth2-proxy:latest
+        image: quay.io/oauth2-proxy/oauth2-proxy:v7.6.0
         imagePullPolicy: Always
         livenessProbe:
```

jmazzitelli:
I just found out that something in the latest oauth2-proxy broke (this confirms why we just saw this happening very recently). See https://github.com/oauth2-proxy/oauth2-proxy/issues/2802

We could keep this PR since it is upgrading Dex, too, but as it is right now, we do not need to do that. We just have to wait for a fix to go into oauth2-proxy.

jmazzitelli:
I'm closing this. The new Dex is behaving differently so rather than dork with getting all that to work again, I'll create another PR that simply pins on the known-working oauth2-proxy.

PR #7791, opened 2024-10-03
[pft graph] various look+feel tweaks to improve UX

jshaughn:


ferhoyos:
Additional thing to improve in PFT UX: the edge state is not changed from hover to selected when it is clicked (the focus still in grey). The user does not really know if the edge has been clicked correctly (well, the side panel is refreshed accordingly, but it should be noticed in the graph itself).

Hovered / Hovered and selected:
image

Selected:
image

In summary, the selected state should be prioritized over hover state.

ferhoyos:
I have noticed that we have lost the feature to show icons inside services:

Before:
image

After:
image

Probably it is caused by the use of custom triangle shapes. I wonder if it is possible to add such images in custom shapes as well.

jshaughn:
> I have noticed that we have lost the feature to show icons inside services:

Ooh, good catch.  I will investgate!

jshaughn:
OK, so in addition to the original items, this PR now also has:
- Fix/Improve node background images (bigger, better placement)
- Fix to selected node highlighting, now uses the same highlighting as hover (highlight the paths involving the hovered node/edge)
- Fix node shape issue for "find" nodes (from @ferhoyos)
- Improve group selection styling (to just style edges like cytoscape), and some node/edge styling (from @ferhoyos)
  - Apply the same group styling to mesh page
  
Also, I have merged all other PFT/Ambient changes and rebased.  So, I think this is ready for final review, and hopefully merge!

leandroberetta (commented):
Nice additions!. 

I did a test and looks good, the only thing I noticed is that for the zoom node when scaling, sometimes there are persistent zoomed nodes and labels:

image

Also, it seems that you need to stay for a second in the node for then to go an click on the kebab menu, otherwise it goes away pretty fast.

jshaughn:
Thanks for the review, @leandroberetta!

> I did a test and looks good, the only thing I noticed is that for the zoom node when scaling, sometimes there are persistent zoomed nodes and labels:

I haven't seen the duplicate zoom node thing, but it looks like it can happen.  This feature is completely controlled by PFT so not sure we can do anything about it.  And I agree that it's not built well to get to the zoomed kebab, as oon as you leave the node boundary it goes away.  Better to right-click on the node.  I still sort of like this feature because it makes it easy to zoom on a small node (like if it's red in a large graph), especially since we don't yet have select-zoom.

If you are OK with the PR as-is, please approve and then we'll just wait on Fernando's review.  But if you want that feature removed let me know.

jshaughn:
One more push: Fix the graph tour's Shortcuts for pf graph

leandroberetta:
> Thanks for the review, @leandroberetta!
> 
> > I did a test and looks good, the only thing I noticed is that for the zoom node when scaling, sometimes there are persistent zoomed nodes and labels:
> 
> I haven't seen the duplicate zoom node thing, but it looks like it can happen. This feature is completely controlled by PFT so not sure we can do anything about it. And I agree that it's not built well to get to the zoomed kebab, as oon as you leave the node boundary it goes away. Better to right-click on the node. I still sort of like this feature because it makes it easy to zoom on a small node (like if it's red in a large graph), especially since we don't yet have select-zoom.
> 
> If you are OK with the PR as-is, please approve and then we'll just wait on Fernando's review. But if you want that feature removed let me know.

No, I don't want to remove it, to me it's fine, once you know about it, is not an issue.

ferhoyos (approved):
Great improvements @jshaughn! I think we are ok with this PR.

I also experience the persistent zoom node issue that @leandroberetta mentioned, but as @jshaughn said, it's a PFT topology issue, and the feature is very useful for large topologies.

My question about kebab menu is wether we need it if we already have the right-click action to open the context menu...

Regarding Mesh topology, I think that CSS classes could be improved a little (e.g., hovering edges indicates that we can interact with them, which is not correct). However, this is out of scope of this PR.

jshaughn:
> My question about kebab menu is wether we need it if we already have the right-click action to open the context menu...

@ferhoyos good observation.  After talking with Foday on slack, let's remove the Kebab.  I can deal with that in a follow-up PR.  Let's get this one merged.

PR #7794, opened 2024-10-04
[hack] k8s-minikube.sh to pin to oauth v7.6.0 

jmazzitelli:
due to https://github.com/oauth2-proxy/oauth2-proxy/issues/2802

Without this fix, some of the molecule tests fail


jshaughn (approved):
Looks fine, should we switch back when "latest" is fixed?

Issue #7709, opened 2024-09-05
[Perf] Add performance tests for PF Graph


hhovsepy:
Currently there are performance measurements on cytscape Graph.
Need to make similar measurements for PF Graph as well to compare the results.

First the cypress integration tests should be created for PF Graph.

PR #7797, opened 2024-10-07
PF Graph performance tests.

hhovsepy:
### Describe the change

Adjusted existing Graph performance tests to PF Graph.
Taking 3 measurements:

1. All Namespaces.
2. All Namespaces with Idle nodes.
3. Selected (few) Namespaces.

### Automation testing

Perf tests added

### Issue reference
https://github.com/kiali/kiali/issues/7709

### DIFF
--- a/frontend/cypress/perf/common.ts
+++ b/frontend/cypress/perf/common.ts
@@ -23,8 +23,7 @@ const measureLoadTime = (
 ): void => {
   // Getting an average to smooth out the results.
   let sum = 0;
-  // for graph page load only once, otherwise braking on jenkins
-  const visitsArray = Array.from({ length: isGraph ? 1 : visits });
+  const visitsArray = Array.from({ length: visits });
 
   cy.wrap(visitsArray)
     .each(() => {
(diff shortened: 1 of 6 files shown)


frontend/cypress/perf/common.ts
```diff
@@ -33,7 +39,12 @@ const measureLoadTime = (name: string, baseline: number, loadUrl: string, loadEl
           cy.on('uncaught:exception', (err, runnable) => {
             return false;
           });
-          cy.get(loadElementToCheck).should('be.visible');
+          if (isGraph) {
+            cy.waitForReact();
+            cy.getReact('GraphPF').should('have.length', '1');
```
ferhoyos, line 44: I think we should double check here that the Graph is really loaded (maybe the component `GraphPF` exists when the graph is being loaded). I would add an extra assertion here, e.g., the graph nodes length is more than 0 or the message `Graph Loading` is not displayed. Look at https://github.com/kiali/kiali/pull/7799 for some examples.

ferhoyos (approved):
LGTM. Thanks @hhovsepy for adding more assertions to the graph load. Checking that nodes and edges exist will ensure the graph has been loaded correctly.

frontend/cypress/perf/common.ts
hhovsepy: have changed to `cy.getReact('GraphPagePFComponent', { state: { isReady: true } }).should('have.length', '1');`, the double check with `graphRefs` causes Jenkins job to fail with CPU/Memory issues.
for now having this solution with TODO

ferhoyos (approved):
LGTM

PR #7798, opened 2024-10-08
Update the config for the open telemetry collector

josunect:
### Describe the change
Update the config for the open telemetry collector (For multi cluster)

### Steps to test the PR

And check if it has traces. 

cy should pass.

### Automation testing

N/A

### Issue reference
N/A


jshaughn (approved):
Thank you @josunect for tracking this down!  When merging please make sure the commit message indicates why we had to make this change, so looking back we understand what broke.

PR #7805, opened 2024-10-10
Hide node kebab context menus from pft graph

### LINKED ISSUE #7803: [pft graph] remove node kebab menus in favor of secondary-click contextual menu
There is no need for the kebab now that we have secondary-click feature to pop up the context menu.  This simplifies and shortens the node labels, and will be a straight carryover from the cytoscape behavior.

ferhoyos:
### Describe the change

This PR hides the node kebab context menus in the PFT graph using CSS, as it is not possible to disable them within patternfly topology

image

I have been able to center the text in the label, but I couldn't reduce the size of the label's rectangle since it is defined programmatically in the `patternfly-topology` library. However, the labels still look nice.

**Bonus:** Fine-tune the CSS of the gateway icon within the gateway label node.

### Steps to test the PR

Look that the kebab menus are hiding in the pft graph and only the right-click action can display the context menu.

### Issue reference

Closes #7803

### DIFF
--- a/frontend/src/styles/GlobalStyle.ts
+++ b/frontend/src/styles/GlobalStyle.ts
@@ -89,7 +89,7 @@ export const globalStyle = kialiStyle({
     },
 
     /**
-     * Hide kebab menu from Patternfly topology groups
+     * Hide the kebab menu of Patternfly topology groups
      */
     '& .pf-topology__group__label': {
       $nest: {
--- a/frontend/src/styles/GlobalStyle.ts
+++ b/frontend/src/styles/GlobalStyle.ts
@@ -86,6 +86,23 @@ export const globalStyle = kialiStyle({
           color: 'var(--pf-v5-global--link--Color--light--hover)'
         }
       }
+    },
+
+    /**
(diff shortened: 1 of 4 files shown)


frontend/src/styles/GlobalStyle.ts
```diff
@@ -86,6 +86,23 @@ export const globalStyle = kialiStyle({
           color: 'var(--pf-v5-global--link--Color--light--hover)'
         }
       }
+    },
+
+    /**
+     * Hide kebab menu from Patternfly topology groups
+     */
+    '& .pf-topology__group__label': {
```
ferhoyos, line 94: PFT groups don't have a `labelClassName` property, so I have to modify the group label CSS in the global style.

ferhoyos:
Good catch about the "Shortcuts" in the graph tour. I think it is a good idea to create an issue in PFT and, whenever possible, submit a PR to let the developer decides how the context menu is opened (only kebab, only right-click or both).

jeff-phillips-18:
Why do you want to hide the kebab menu? This makes kiali different from most (all?) other products' topology views.

PR #7808, opened 2024-10-10
[pft graph] Add dynamic "wizard" options to service node context menus

### LINKED ISSUE #7807: [pft graph] add back dynamic context menu actions
When the PFT graph was first created PFT required that all context menu items be returned immediately.  But some Kiali context menu actions are determined given the result of back-end calls.  These dynamically determined items had to be excluded.  We requested the ability to return items async, via a promise in https://github.com/patternfly/react-topology/issues/15.  This was delivered in PFT v5, so now we should be able to add back those dynamically determined action items.

jshaughn:
Closes #7807 

Add dynamic "wizard" options to service node context menus. This is to resolve a feature gap with cytoscape graph.  I'm not sure if there is a better way to do this, one that shares more code with the side panel kebab, but it seems to be working and I was struggling just to get something working. The context menu code is really hairy.
Also:
- compress the context menu, by default the options take up too much space
  - this is a PF standard, but it's cumbersome on the graph.
- fix a small problem in the graph tour shortcuts, the node labels no longer have a kebab menu.


frontend/src/pages/GraphPF/components/stylesComponentFactory.tsx
```diff
@@ -68,22 +75,94 @@ const nodeContextMenu = (node: GraphElement, kiosk: string): React.ReactElement[
...
+    nodeData.namespace,
+    ['peerauthentications'],
+    false,
+    '',
+    '',
+    nodeData.cluster
+  );
+
+  return new Promise<React.ReactElement[]>((resolve, reject) => {
+    const allPromise = new CancelablePromise(Promise.all([getDetailPromise, getGwPromise, getPeerAuthsPromise]));
```
ferhoyos, line 118: I don't understand why you are using a CancelablePromise here when there is no possibility to cancel it (you are not using the `PromiseRegistry` or canceling directly the promise anywhere)

ferhoyos (commented):
Nice improvement on the PF context menu @jshaughn! While reviewing your PR, I noticed that you can add any component to the context menu, which could help us make it closer to the Cytoscape one. I have created a PR to your PR to include all default context menu options in a `Show` dropdown group

image

We could even add the header from the Cytoscape context menu:

image

However, I don’t recommend doing that for the moment because dynamic context menus don't update their direction to open upward if there isn't enough space downward.. I have created an issue in `react-topology` - https://github.com/patternfly/react-topology/issues/238

frontend/src/pages/GraphPF/components/stylesComponentFactory.tsx
jshaughn: Right, there was no need.  That's just an artifact of stealing code and not properly cleaning it up.

jshaughn:
>  While reviewing your PR, I noticed that you can add any component to the context menu, 

Right, when I noticed that it helped just add the "dynamic" options in one "block".

> However, I don’t recommend doing that for the moment because dynamic context menus don't update their direction

+1.  The long context menus are chopped at the bottom.  This was true in cytoscape as well, so it's not a regression.  But if we can get PFT to fix it that would be nice.

ferhoyos (approved):
LGTM

PR #7809, opened 2024-10-11
fix issues with control plane metrics

### LINKED ISSUE #7441: Memory consumption chart for istiod container isn't getting created
### Describe the bug
 - When I use the default prometheus addon and link it with the kiali from kiai-operator I can see the cpu and memory consumption charts in the mesh page
image

- But when I link it with teh prometheus as an external deployment (deployed using kube-prometheus-stack) I can see the cpu consumption chart but I couldn't see memory consumption chart
image

- I can see the required metrics to create those charts in my prometheus UI
image

Why is that happening?
### Environment

- **Kiali version:** 1.86.0
- **Istio version:** 1.22.1
- **Kubernetes impl:** prometheus-community/kube-prometheus-satck
- **Kubernetes version:** 1.29.2

jshaughn:
Fixes #7441 

When control plane information was shown in a namespace card on the Overview page, it was included as part of namespace metrics. But now that we have more fine-grained control plane nodes in the Mesh page, and there may be multiple control planes, we need to move the logic from namespace metrics and ensure that the retrieved information is specific to the specific control plane workload.
    
The new logic creates dedicated control plane metric support. It determines the pods for the specific control plane deployment, and uses the pod names to narrow metrics to just the relevant time series.
    
Also, I think there was a flaw in the prior logic for determining the proxy sync time. Those metrics are unusual as they are infrequent and reported more like a step function. You can't really capture it well as a "rate" because it reports all at once when there is a config push, and then goes flat.  Rates need a series of data points. So, this applies some non-standard logic to capture any sync that takes place during the requested time period, by capturing deltas for the time period, and then calculating the avg sync time (deltaSyncTime / deltaNumberOfSyncs).   This turned out to be painful, as we didn't really have the underlying support for the type of prom queries needed. So, this adds some support, but it seems to work pretty well.

There is also a little bit of refactoring/cleanup done, which explains some of the file touches:
- Rename and move OverviewCardControlPlaneNamespace to TargetPanelControlPlaneMetrics.
- Rename and move ControlPlaneNamespaceStatus to TargetPanelControlPlaneStatus.Mesh page side panel.
  - These were only used in Mesh page side panel.
- Update "Proxy Push Time" to "Proxy Sync Time", and show it always for consistency. Update tooltip text.
- Fix a side issue with multiple exports of "infoStyle".
- Linter fixes on some touched files

To test you can navigate to the Mesh Page and select a control plane.  You can try comparing the charts to the raw graphs provided in Prometheus, or just sort of eyeball it for sanity against the prior charts.  Because this no longer applies the `app="discovery"` label check (which I didn't understand and nobody explained), it's quite likely that memory numbers will come from `container_memory_working_set_bytes` as opposed to `process_resident_memory_bytes`.

As for the proxy sync time, you will likely see `N/A` to start with.  But if you perform a config change you should see a number.

I wasn't sure who to assign for review, @leandroberetta was more involved in this code, but @nrfox has been spending more time in the Mesh code.  You can both review, or one of you can just volunteer.


business/metrics.go
```diff
@@ -251,39 +252,62 @@ func createStatsMetricsLabelsBuilder(q *models.MetricsStatsQuery) *MetricsLabels
...
+	// and so there is typically no range, because any range requires at least two data points.  What we really want here
+	// is delta(sum) / delta(count) for the time period. Note, this is pretty non-standard manipulation of a histogram,
+	// don't try this at home.
+	deltaDuration := q.End.Sub(q.Start)
+	deltaSumMetric := in.prom.FetchDelta("pilot_proxy_convergence_time_sum", podLabel, "", q.End, deltaDuration)
+	deltaSumConverted, err := models.ConvertMetric("pilot_proxy_convergence_time_sum", deltaSumMetric, models.ConversionParams{Scale: 1})
+	if err != nil {
+		return nil, err
+	}
+	deltaSum := deltaSumConverted[0].Datapoints[0].Value
```
nrfox, line 278: These are always going to have 1 value?

handlers/metrics.go
```diff
@@ -199,14 +195,34 @@ func NamespaceMetrics(promSupplier promClientSupplier, discovery remoteClusterId
 			return
 		}
 
-		metrics, err := metricsService.GetMetrics(params, nil)
+		if discovery.IsRemoteCluster(r.Context(), cluster) {
+			RespondWithError(w, http.StatusServiceUnavailable, fmt.Sprintf("cannot fetch control plane metrics for remote cluster [%s]", cluster))
+			return
+		}
```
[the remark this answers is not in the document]
jshaughn: Right, this is actually a completely new route, introduced in this PR.  And it is called only from the control place side panel on the mesh page.  I carried over this check from the original namespace metrics route, and wasn't exactly sure what it was about.  So thanks for the explanation, I will remove it here.

business/metrics.go
jshaughn: That's correct.  It's the delta for the counter for the time period. Basically val(tEnd) - val(tStart).

jshaughn:
Thanks for the review, @nrfox .  I made the recommended changes and pushed the update.

handlers/metrics.go
```diff
@@ -166,6 +166,69 @@ func getAggregateMetrics(w http.ResponseWriter, r *http.Request, promSupplier pr
 	RespondWithJSON(w, http.StatusOK, metrics)
 }
 
+// ControlPlaneMetrics is the API handler to fetch metrics to be displayed, related to a single control plane revision
+func ControlPlaneMetrics(promSupplier promClientSupplier, discovery remoteClusterIdentifier) http.HandlerFunc {
```
nrfox, line 170: I think you can now get rid of `discovery` from here and from the rest of the <...Metrics> functions as it was only for checking for the remote cluster and we no longer need to do that. ` remoteClusterIdentifier` can go too it looks like?

jshaughn: You are correct, sir!  Weird, I'm surprised the IDE didn't flag that unused parameter.  Changes pushed.

Issue #7491, opened 2024-06-28
Include 2e2 tests for Waypoint proxies


josunect:
### What do you want to improve?
Include 2e2 Tests for Waypoint proxies

### What is the current behavior?
All the Ambient tests are related to ztunnel, but there are some logic related to waypoint that should be also included. 
Note: Some of this features could change in the future Istio releases. Tests should help us detect those changes related. 

### What is the new behavior?
No e2e waypoint tests

josunect:
Additional tests should include: 

- [ ] Click on a waypoint edge and verify the double arrow button exists and works
- [ ] The waypoint proxy should not have "Missing sidecar" messages

josunect:
Fixed by https://github.com/kiali/kiali/pull/7813

PR #7813, opened 2024-10-14
[ambient] e2e test - Initial waypoint tests

josunect:
### Describe the change

Add initial test for Ambient with waypoints.
This PR uses a different tag, instead of @ambient, uses @waypoint. This way, the environment is set with an Ambient Mesh and an Ambient namespace, bookinfo. A waypoint proxy is created but the namespace is not enrolled. 
Once the tests run for ztunnel, the bookinfo namespace is enrolled labeling the namespace with the waypoint label, and the @waypoint tests run. 

waypoint tests: 

- [X] Add waypoint label
- [X] See the waypoint workload in the workloads list with the right labels and details 
- [x] See the workload has the L4 and L7 label in details
- [x] See the ztunnel traffic
- [x] See the waypoint traffic
- [x] Don't see Ambient traffic
- [x] See all the Ambient traffic
- [x] User doesn't see the waypoint node 
- [x] User sees the waypoint proxy node when "Display waypoint" is selected
- [x] Waypoint doesn't have validation errors in the Istio Config
- [x] Namespace is labeled with the waypoint labels

### Steps to test the PR

`yarn cypress run -e TAGS=@waypoint`

### Automation testing

e2e Ambient job runs @ambient and then @waypoint

### Issue reference

#7491


hhovsepy (approved):
lgtm, all pass locally

PR #7816, opened 2024-10-15
Graph toolbar: Replace disabled with active style

ferhoyos:
### Describe the change

Currently, active options in the patternfly graph toolbar appear disabled. This is not ideal from a UX perspective, as they can be mistaken for actually disabled toolbar options. Therefore, this PR changes the toolbar icons to blue when they are active and slightly increases their size to better differentiate them from similar inactive icons. 

Before:

image

After: 

image

To be consistent with the Cytoscape graph, options to hide edges can be activated and deactivated by clicking the same toolbar option (which was not possible with the disabled approach).

### Steps to test the PR

1. Go to graph page
2. Verify that the toolbar looks fine with the new active icon styles.

### Automation testing

N/A

### DIFF
--- a/frontend/src/pages/GraphPF/GraphPF.tsx
+++ b/frontend/src/pages/GraphPF/GraphPF.tsx
@@ -71,9 +71,8 @@ import { GraphRefs } from './GraphPagePF';
 import { WizardAction, WizardMode } from 'components/IstioWizards/WizardActions';
 import { ServiceDetailsInfo } from 'types/ServiceInfo';
 import { PeerAuthentication } from 'types/IstioObjects';
-import { kialiStyle } from 'styles/StyleUtils';
-import { PFColors } from 'components/Pf/PfColors';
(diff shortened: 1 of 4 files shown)


jshaughn (commented):
This is great!  I didn't actually think we could do this without some changes to PFT.  Please also apply to the Mesh page and then we can merge.

ferhoyos:
I have implemented the active toolbar option for the mesh page as well.

I didn't realize this was a "Nice to have" in the epic. I just did it because it took less time for me to adapt the PFT graph code to the Cytoscape `graph_toolbar` Cypress test than the other way around :slightly_smiling_face:

jshaughn (approved):
>  didn't know this was a "Nice to have" in the epic. I just did it because it took less time for me to adapt the PFT graph code to the Cytoscape graph_toolbar Cypress test than the other way around 🙂

Whatever it takes :-)  We'll say we're doing "test first development".

LG!

PR #7834, opened 2024-10-17
Istio Configs JSON to GVK

hhovsepy:
### Describe the change

Istio Config List and Details REST API calls JSON results are now returning `resources` with `key=GVK` and values as `[]IstioConfig` instead of hardcoded 'k8sGateways' like before.

### Steps to test the PR

Regression testing on Istio Config List/Details pages, Graph pages.

### Automation testing

Existing tests are modified to support the new format.

### Issue reference

https://github.com/kiali/kiali/issues/7457


ferhoyos (changes requested):
Nice job @hhovsepy. With this PR, `k8s...` prefixes are gone from HTTP requests/responses.

Some comments:
- To me, the string format {$group/$version, Kind=$kind} is a bit weird for describing a key. For example, OpenShift uses the format {$group~$version~$kind} to uniquely identify a GVK in console URLs. This format is also easier to convert into GVK object.

- The function `gvkToString(dicIstioTypeToGVK[...])` is used multiple times across the application and can be hard to understand what this code is doing.

In my opinion, I’d suggest replacing `gvkToString` with a util function like this:
- `getGVKReference(gvk: GroupVersionKind | istioType: string): string`: Returns the GVK reference in the format `$group~$version~$kind`. The input parameter can be either the istio type or a GVK object.

hhovsepy:
Thank you @ferhoyos  for the review, have approached the comments. However the `{$group/$version, Kind=$kind}` format is remaining as this came from k8s Golang libraries.

frontend/cypress/integration/common/istio_config.ts
```diff
@@ -1,6 +1,7 @@
 import { After, Given, Then, When } from '@badeball/cypress-cucumber-preprocessor';
 import { colExists, getColWithRowText } from './table';
 import { ensureKialiFinishedLoading } from './transition';
+import { getGVKTypeString } from '../../../src/utils/IstioConfigUtils';
```
ferhoyos, line 4: Could you use absolute 'utils' folder path in cypress test imports? Relative paths do not work well with OSSMC.

```suggestion
import { getGVKTypeString } from 'utils/IstioConfigUtils';
```

ferhoyos (changes requested):
Thanks for addressing my comments @hhovsepy! If the string format is defined by the K8s golang libraries, then I have nothing else to say :slightly_smiling_face: 

Just one small change in a cypress test before approval.

frontend/cypress/integration/common/istio_config.ts
hhovsepy: sure, done :)

ferhoyos (approved):
LGTM

Issue #7835, opened 2024-10-18
Mesh and Traffic Graph are not changing layout on an element change


jshaughn:
Paraphrased from Slack:

From @nrfox : I have noticed this on the patternfly graph that the layout is a bit weird when you hit the refresh button and it redraws elements vs. when you refresh the whole page

![WeirdLayout](https://github.com/user-attachments/assets/46763bba-d4cf-468e-9c88-ff43c6b38592)

It's like it doesn't get the zoom level right when it needs to add/remove elements

If you look closely the httpbin node when it appears gets placed on top of another node. It's not until I refresh the whole page that the whole layout changes to properly accommodate the new node.

You can reproduce by:
```
kubectl apply -f https://raw.githubusercontent.com/maistra/istio/maistra-2.6/samples/httpbin/httpbin.yaml -n bookinfo
kubectl exec -it -n bookinfo deployments/productpage-v1 -c istio-proxy -- curl -I httpbin:8000

```
And then hitting the refresh button (edited) 

The mesh page has the same problem

![hiddennode](https://github.com/user-attachments/assets/6777294d-dbb5-4f5d-9d58-a47abb49c152)

It took me a second to find where the node was. It got placed behind the grafana node. Basically if another node gets added from the graph response, the actual layout doesn't seem to change.

jshaughn:  ok, that's pretty bad.

PR #7836, opened 2024-10-18
[mesh page] [graph] Fix issue with laying out on an element change

jshaughn:
Fix issue with laying out on an element change in both the traffic graph and the mesh page.

Also, fix a resize issue for mesh page.

Fixes https://github.com/kiali/kiali/issues/7835

To test:
- try the repro steps in the issue.  
- for the resize issue on the mesh page, try collapse and expand of the side panel
- also try resizing, graph hide/unhide (refresh with hide), eliminating/adding things via the Traffic dropdown, refreshing when you have changed the default layout (like zooming in/out, dragging etc) and make sure your custom view is maintained (unless there is an element change).

### DIFF
--- a/frontend/src/pages/Graph/GraphToolbar/GraphFindPF.tsx
+++ b/frontend/src/pages/Graph/GraphToolbar/GraphFindPF.tsx
@@ -43,8 +43,8 @@ import {
   selectOr,
   SelectOr
 } from 'pages/GraphPF/GraphPFElems';
-import { FIT_PADDING } from 'pages/GraphPF/GraphPF';
 import { isArray } from 'lodash';
+import { graphLayout, LayoutType } from 'pages/GraphPF/GraphPF';
 
 type ReduxStateProps = {
   edgeLabels: EdgeLabelMode[];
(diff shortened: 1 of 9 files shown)


nrfox:
This fixes the problem for the mesh page but it looks like the graph page still doesn't adjust the layout when a new node is added:
layout

jmazzitelli:
I don't know how this happened but --- an edge going to nowhere out of productpage:
image

jmazzitelli:
Crazy edge going to and from nowhere - I click it but the sidepanel has no indication of what its for
image

jmazzitelli:
Finally, set hide and find expressions (I used this:)
image

Zoom in or out, hit refresh - that works fine - the zoom level stays.
But move some of the nodes and boxes around for a custom layout, hit refresh - the graph snaps back to the way it was originally. From the description of this PR, it seems like that should not happen.

jshaughn:
I have not seen that sort of weird edge behavior on hide. That would be independent of this PR, I think.  Not sure how you'd get a ghost edge.  

Anyway, your other thing, the snap back on "node drag with hide" was a good find.  This code is so tricky to get right.  But, I think I have it fixed.  I pushed the changes.  Also, as part of this I fixed the mesh page label issue you mentioned (the clipped labels and the hard to read badges in dark mode).  That's also pushed.

jmazzitelli (approved):
I'm going to approve this now. I didn't re-test since the latest commit, but it isn't going to be made worse, so let's get this in the release now if possible. If we miss the release with this PR, we can get this in for the 2.1 release.

PR #7799, opened 2024-10-08
Cypress tests for PatternFly Graph

leandroberetta:
Cypress tests for PatternFly Graph.

Coverage:

- [x] frontend/cypress/integration/featureFiles/app_details_graph-pf.feature
- [x] frontend/cypress/integration/featureFiles/app_details_multicluster_graph-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_context_menu-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_display_user-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_display-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_find_hide-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_replay-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_side_panel-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_toolbar_legend-pf.feature
- [x] frontend/cypress/integration/featureFiles/graph_toolbar-pf.feature
- [x] frontend/cypress/integration/featureFiles/service_details_graph-pf.feature
- [x] frontend/cypress/integration/featureFiles/service_details_multicluster_graph-pf.feature
- [x] frontend/cypress/integration/featureFiles/workloads_details _graph-pf.feature
- [x] frontend/cypress/integration/featureFiles/workloads_details_multicluster_graph-pf.feature

PR #7847, opened 2024-10-21
Uncomment waypoint tests

josunect:
### Describe the change
Wait for the traffic to be created.

### Steps to test the PR

`yarn cypress open -e TAGS=@waypoint`

### Automation testing

cypress tests update

### Issue reference
Ref. https://github.com/kiali/kiali/pull/7799

### DIFF
--- a/frontend/cypress/integration/common/graph_display-pf.ts
+++ b/frontend/cypress/integration/common/graph_display-pf.ts
@@ -528,7 +528,7 @@ Then('the {string} node {string} exists', (nodeName: string, action: string) => 
       const foundNode = nodes.filter(node => node.getData().workload === nodeName);
 
       if (action === 'does') {
-        assert.isAtLeast(foundNode.length, 1);
+        assert.equal(foundNode.length, 1);
       } else {
         assert.equal(foundNode.length, 0);
       }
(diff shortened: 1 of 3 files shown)


frontend/cypress/integration/common/graph_display-pf.ts
```diff
@@ -497,6 +497,8 @@ When('user {string} {string} traffic option', (action: string, option: string) =
 
 Then('{int} edges appear in the graph', (graphEdges: number) => {
   cy.waitForReact();
+  ensureKialiFinishedLoading();
```
leandroberetta, line 500: Is this necessary as this is also in:

https://github.com/kiali/kiali/blob/3e9b2a2588384e1bd371696834fc3fa02736c055/frontend/cypress/integration/common/graph_display-pf.ts#L33

leandroberetta (commented):
I haven't test it yet but looks good, just one question.

frontend/cypress/integration/common/graph_display-pf.ts
ferhoyos: I don't think this command is necessary because it already waits for React to finish. However, it doesn't cause any harm either.

frontend/cypress/integration/featureFiles/waypoint.feature
```diff
@@ -13,6 +13,7 @@ Feature: Kiali Waypoint related features
   @waypoint
```
ferhoyos, line 13: Not related to this PR, but if the whole feature is already tagged with @waypoint, then it's not necessary to tag the scenarios as well, right?

ferhoyos (approved):
LGTM. Just a pair of comments, but they don't block the PR. Thanks for fixing the ambient cypress tests @josunect!

PR #7852, opened 2024-10-21
Fine-tune Mesh topology style

ferhoyos:
### Describe the change

This PR changes some mesh topology styles for a better appearance.
- Keep other nodes highlighted when a mesh node is selected or hovered. This feature is useful in the traffic graph, where many nodes are present, but it's less helpful in the mesh graph when there are only a few nodes.

- Hovering over an edge does not change its style. Since edges on the mesh page are not clickable, it's better to maintain a consistent CSS style to avoid user confusion.

- Don't hide the icons at low detail levels. The icons remain recognizable even at the lowest zoom levels.

image

- Scale the nodes at low detail levels:

image

- The selection stroke is wider and blue like in the traffic graph:

image

- Edges connected to failure nodes are thicker, making the failures more visible:

image

**Bonus:**
- Adjusted the margin for the Waypoint proxy icon in the traffic graph.

Before:

image

After:

image
 

### Steps to test the PR

1. Go to the mesh page
2. Check that style changes look good

### Automation testing

N/A

### DIFF
--- a/frontend/src/pages/GraphPF/GraphPFElems.tsx
+++ b/frontend/src/pages/GraphPF/GraphPFElems.tsx
@@ -208,7 +208,7 @@ export const setNodeAttachments = (node: Node<NodeModel>, settings: GraphPFSetti
   }
 };
 
-const labelIconStyle = kialiStyle({
+const rootIconStyle = kialiStyle({
   color: PFColors.White,
   display: 'flex',
   marginLeft: '0.125rem',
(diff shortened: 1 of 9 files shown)


jshaughn (approved):
I agree with everything here. Nice changes, @ferhoyos .  We could remove the mesh highlighter support completely, I guess.  But, I'm not opposed to leaving it in place for now, just in case we decide  to do some sort of highlighting in the future.

Issue #7769, opened 2024-09-24
[Tempo] Review performance


josunect:
### What do you want to improve?
Review Tempo performance

### What is the current behavior?
With the sail hack script, the behavior is flaky. 

### What is the new behavior?
Performance improved in Tempo

Tasks:

- [x] https://github.com/kiali/kiali/pull/7768
- [x] Review best approaches (Like from Grafana, etc), default Tempo config, ... Document "Tempo tuning" https://github.com/kiali/kiali.io/pull/830
  - https://grafana.com/docs/tempo/latest/operations/backend_search/
  - https://grafana.com/docs/tempo/latest/operations/dedicated_columns/
  - https://grafana.com/docs/tempo/latest/operations/caching/
- [x] Improve Tempo client in Kiali https://github.com/kiali/kiali/pull/7854 
- [x] Test Tempo with multi tenancy in OpenShift (Updated blog post for reference)
- [x] Allow users to set trace query limits https://github.com/kiali/kiali/pull/7895
- [x] Optimize Tempo query: https://github.com/kiali/kiali/issues/7903
- [x] https://github.com/kiali/kiali/pull/7923

josunect:
I think the main issue with Tempo is, that in the Tempo operator, they specify this resources as the default (https://docs.redhat.com/en/documentation/openshift_container_platform/4.15/html/distributed_tracing/distributed-tracing-platform-tempo#distr-tracing-tempo-install-tempostack-web-console_dist-tracing-tempo-installing):
 ``` resources:
    total:
      limits:
        memory: 2Gi
        cpu: 2000m 
```

These resources are for all the tempo components, and I think they are not enough to run just simple queries. If you run this query just 5 times from the grafana UI, you got a OOMKilled:
![image](https://github.com/user-attachments/assets/635a9575-9509-4879-92c5-21bbd480de20)
![image](https://github.com/user-attachments/assets/126637a1-e700-4c6a-9934-42fa3ea1bf57)

Without specifying those limits, the query that we do: 
![image](https://github.com/user-attachments/assets/381ac3df-ff2a-4952-b193-295bb5a8ad53)

Takes between 300 ~= 400 ms.

jshaughn:
That's good data.  Without the limits can you tell how much memory gets consumed? That seems like a pretty quick OOM for 2gb of mem, but I guess that's what happens.

josunect:
[...]

PR #7854, opened 2024-10-22
[Tempo] Add Tempo cache

josunect:
### Describe the change

Adds a TTL cache for Tempo to minimize the access to the Tempo backend. 

It creates a couple of structures: 
- One that saves the results for Tempo queries
- Other that saves the result for Tempo individual traces

The cache for queries save the results by minutes granularity, trying to increase the cache effectiveness. 

As a TTL cache, every entry has a expiration time that removes the entry once is expired. 

This expiration time is defined globally for all the elements (CacheExpiration) and it is set to 300s by default. It will be checked by a watcher every minute, as the watcher needs to check all the entries so decrease the watcher time, will probably have a higher cost. 

Every time the watcher runs, it will display the hit rate (Since the cache was created) and the cache size. So it might help to adjust the expiration time. 

This cache can be useful, for example, when the user is looking at the traces, and click on the same trace twice (It can be done for comparison). Also, when it clicks on the traffic graph node, in the traces tab, and click back on the same node. Here it can save 10 calls to the backend.

### Steps to test the PR

Run the environment with: 
- `hack/istio/tempo/install-tempo-env.sh -c kubectl`
- Install Kiali
- Tempo cache should be enabled 

### Automation testing
Covered in Tempo tracing CI pipeline

### Issue reference

Ref. #7769


jshaughn:
I like that the caching is optional, I can imagine this cache could get large quickly.  I guess the prom caching would be similar.  We don't hold onto that information very long, it's mainly to handle when the same query is executed repeatedly in a small time window.

josunect:
> I like that the caching is optional, I can imagine this cache could get large quickly. I guess the prom caching would be similar. We don't hold onto that information very long, it's mainly to handle when the same query is executed repeatedly in a small time window.

Right. So maybe the default value that I've added was very large (600 seconds). Maybe set to 300 seconds or less?
I have added these stats so we can see the size/hit rate:

```
2024-10-22T15:50:49+01:00 INF [Tempo Cache] Cache hit rate: 41.94%
2024-10-22T15:50:49+01:00 INF [Tempo Cache] Cache size: 7611 bytes
```

Probably better to be as a debug info

jshaughn:
If I understand the code, it looks like we use the cached information if we re-query during the TTL period.  So the longer the TTL, the longer we go without fetching for new traces.  And then we throw out the cache and reload.  So the pattern would be: fullFetch, staleCachedFetches, fullFetch.  Considering that we have a "sliding window" when talking about the scatterplot graph, I wonder if this is the best approach?  Wouldn't this leave gaps in the sliding window?  If TTL was 5m then would we see a 5m period with no traces?  What happens if the time window is <= 5m, could it go completely empty?

I wonder if, on each query, we should try to re-use applicable cached data, and fetch for new "head" traces?  In other words, treat the cache more like a fifo list, with old entries falling off the tail, and new entries being added to the head, while much of the data could be re-used?  For example, what if we are fetching every minute for 15 minutes of traces?  The first fetch could pull all 15m, then afterwards it would pull only 1m and re-use the prior 14m from the cache.

josunect:
> e, it looks like we use the cached information if we re-query dur

- Right, the cached information is used if we re query during the TTL
- Every minute, there is a clean up for the expired records (Not the full cache). There is not a fullFetch, just we save in the cache the queries done.
- It cannot go completely empty because when the data is not in the cache, it is checked in the backend.

The cache is not very smart, because if you look for start 1.12 to 1.18 it saves one result and if you look for 1.14 to 1.20 y will save another record, with the full list of resources. So, you mean by saving all the separated traces? I guess the problem is that we use another query data like the cluster, the min duration, tags, that could make also this list to increase. 
 
This helps on this case: The user checks the traces on a node in the graphs and see the traces, then clicks in another node, then clicks again in that node --> It saves 10 queries to the backend. 
Also if the use clicks in the same trace again in the graph, it saves one query in the backend.

nrfox:
How does this compare to the caching support tempo has: https://grafana.com/docs/tempo/latest/operations/caching/? If someone is already using caching through tempo, would they need this?

I don't think there's much we can do in Kiali to be more efficient with how we are querying tempo but maybe users should focus on tuning their tempo instance or adding caching there before we add it to Kiali. We can even provide some guidance for what knobs they can tune similar to the metric thinning documentation we have for prom.

jshaughn (commented):
I agree with Nick that we should investigate what the Tempo-specific caching can do, and whether RH Observability already has docs or recommended configuration.  But I think some relatively simple caching on our end could also be beneficial, both when the server-side caching isn't enabled, and because we could potentially cut down on data retrievals.  The trace cache probably wouldn't add much if the tempo caching was enabled, but it's simple and wouldn't be particularly costly for us to do and it's not going to be a resource-hog. So, I think I like the trace cache.  I can see how the same trace might be accessed repeatedly in a short time window.  Because this is a low-volume cache, you could probably ignore the timed reaping in favor of just holding some small N of traces.

My problem is more with the query/app cache. Maybe I don't see the use case, but when do we expect a cache hit for the app cache? As I understand it, we would have to repeat the same query for the same time period. Is that right?  I think our use case is typically the same query variables but for a sliding time window. If I sit on the Trace tab and the scatterplot refreshes, we don't get a cache hit, even though we are overlapping `(duration - refreshTime)` results. For example, if I fetch 30m of traces every 10s, only the latest 10s is new, the rest we already had.  (Actually, due to sampling rates we may actually see different traces, but we don't actually have to do that, in fact it could be confusing).  Reducing those subsequent queries to just the time-period "head" seems like the best opportunity for client-side caching. In the example above, you could maybe have a fifo-style cache that dropped the oldest 10s from the tail and added the newest 10s to the head. If the query tags changed you could drop the whole thing and start over.  Again, I don't think you'd necessarily have to reap this cache, you'd have at most one query's worth of data at any time.  Thoughts?

josunect:
> How does this compare to the caching support tempo has: https://grafana.com/docs/tempo/latest/operations/caching/? If someone is already using caching through tempo, would they need this?
> 
> I don't think there's much we can do in Kiali to be more efficient with how we are querying tempo but maybe users should focus on tuning their tempo instance or adding caching there before we add it to Kiali. We can even provide some guidance for what knobs they can tune similar to the metric thinning documentation we have for prom.

Right. I think from Kiali, if we just minimize the number of queries that hit Tempo, it could help minimizing the resources used by Tempo - Even when using caching, we would save that request. 
I think this cache could be more helpful in the scenarios where there is no tune in the backend, and the user just thinks that Kiali is not working fine. 
I'm agree with the tune section also.

josunect:
> I agree with Nick that we should investigate what the Tempo-specific caching can do, and whether RH Observability already has docs or recommended configuration. But I think some relatively simple caching on our end could also be beneficial, both when the server-side caching isn't enabled, and because we could potentially cut down on data retrievals. The trace cache probably wouldn't add much if the tempo caching was enabled, but it's simple and wouldn't be particularly costly for us to do and it's not going to be a resource-hog. So, I think I like the trace cache. I can see how the same trace might be accessed repeatedly in a short time window. Because this is a low-volume cache, you could probably ignore the timed reaping in favor of just holding some small N of traces.
> 
> My problem is more with the query/app cache. Maybe I don't see the use case, but when do we expect a cache hit for the app cache? As I understand it, we would have to repeat the same query for the same time period. Is that right? I think our use case is typically the same query variables but for a sliding time window. If I sit on the Trace tab and the scatterplot refreshes, we don't get a cache hit, even though we are overlapping `(duration - refreshTime)` results. For example, if I fetch 30m of traces every 10s, only the latest 10s is new, the rest we already had. (Actually, due to sampling rates we may actually see different traces, but we don't actually have to do that, in fact it could be confusing). Reducing those subsequent queries to just the time-period "head" seems like the best opportunity for client-side caching. In the example above, you could maybe have a fifo-style cache that dropped the oldest 10s from the tail and added the newest 10s to the head. If the query tags changed you could drop the whole thing and start over. Again, I don't think you'd necessarily have to reap this cache, you'd have at most one query's worth of data at any time. Thoughts?

So, specially for this use case:

image

When you click on a node, it usually generates 10 queries in the backend: 

image

It is mainly the same query but it splits the time to get results spread over a period of time, and not just the last X ones. 
This is done here, independent of the client: 
https://github.com/kiali/kiali/blob/master/business/tracing.go#L146

If you click in one then, and you click back on the previous, the 10 queries will be repeated. And, I think that would be the best scenario for this cache...

jshaughn:
> If you click in one then, and you click back on the previous, the 10 queries will be repeated. And, I think that would be the best scenario for this cache...

Nothing in tracing, or querying temp makes sense to me.  I would never guess that clicking one node in the graph would result in 10 queries to the back-end. Before we continue talking about caching maybe we need to ensure we can't improve the querying.

josunect:
> > If you click in one then, and you click back on the previous, the 10 queries will be repeated. And, I think that would be the best scenario for this cache...
> 
> Nothing in tracing, or querying temp makes sense to me. I would never guess that clicking one node in the graph would result in 10 queries to the back-end. Before we continue talking about caching maybe we need to ensure we can't improve the querying.

If you want to see 100 traces for the last hour, and the search has 1000 traces without the limit, the query will return the last 100 results. If you want to display the last 1 hour graph, you will be probably visualizing 100 traces in the last 10 minutes. 

That's the reason why it makes 10 queries, each one querying for smaller ranges of time for a limit of 10, to get results spread over the last hour. 

That is what happens in the Jaeger UI, for example. Setting a limit of 1500 and requesting the "Last hour" we get traces from 3.30pm to 3.56pm:
image

If we change the limit to 100, we get results from 3.55pm to 3.56pm:

image

We would optimize the number of queries doing just 1 request. But do we want this? This way, we will request the last 1 hour and we will probably see just the last 5 minutes.

This is the current behavior in Kiali:

image

Without this code, requesting 1 hour, returns the last 2 minutes:
image

jshaughn:
> If you want to see 100 traces for the last hour, and the search has 1000 traces without the limit, the query will return the last 100 results. If you want to display the last 1 hour graph, you will be probably visualizing 100 traces in the last 10 minutes.

Wow, OK.  This is something I never knew, and it's super important.  I had always assumed that the returned traces would be distributed randomly over the time period.  And now after reading a bit more on the Grafana site, I think it's not even guaranteed which results you will get back. In fact it seems you may even [get the oldest traces](https://community.grafana.com/t/tempo-in-grafana-search-limited-to-n-rows-returns-first-n-rows-rather-than-last-n-rows/113617)!

> We would optimize the number of queries doing just 1 request. But do we want this? 

This all makes me wonder whether we should pivot... We are trying to force an API to do something it's not designed to do, at the expense of poor performance. So, I'd say yes, I propose we STOP performing multiple queries to try and fill out a time period.  I think a single query is sufficient and we can let users shrink the time period and/or increase the limit.  I think we could probably add some "info" icons in applicable places, briefly explaining this.  I don't think Kiali needs to be the end-all for tracing, we do a good job of correlation with metrics/logs/graph.  We'd still work well with short time periods, and I think OK with larger time periods.  And we can always link to the full tracing UI, when configured.

We may need to figure out how to set a limit in all of the places we offer traces. The graph side panel, as well as the log and metric tab's `spans` option don't have this.

To limit our exposure we may even want to introduce a new config option for tracing, something like,`max_duration`. If the user's time period is < max, use the time period, otherwise the max. It would protect against large time period queries that could be slow, even with a low limit I think Tempo will still look at the whole time period.

Note that in the link above, a later comment mentions that Tempo *may* look into an option to retrieve newest spans, as opposed to any (or typically older) traces.

josunect:
> > If you want to see 100 traces for the last hour, and the search has 1000 traces without the limit, the query will return the last 100 results. If you want to display the last 1 hour graph, you will be probably visualizing 100 traces in the last 10 minutes.
> 
> Wow, OK. This is something I never knew, and it's super important. I had always assumed that the returned traces would be distributed randomly over the time period. And now after reading a bit more on the Grafana site, I think it's not even guaranteed which results you will get back. In fact it seems you may even [get the oldest traces](https://community.grafana.com/t/tempo-in-grafana-search-limited-to-n-rows-returns-first-n-rows-rather-than-last-n-rows/113617)!
> 
> > We would optimize the number of queries doing just 1 request. But do we want this?
> 
> This all makes me wonder whether we should pivot... We are trying to force an API to do something it's not designed to do, at the expense of poor performance. So, I'd say yes, I propose we STOP performing multiple queries to try and fill out a time period. I think a single query is sufficient and we can let users shrink the time period and/or increase the limit. I think we could probably add some "info" icons in applicable places, briefly explaining this. I don't think Kiali needs to be the end-all for tracing, we do a good job of correlation with metrics/logs/graph. We'd still work well with short time periods, and I think OK with larger time periods. And we can always link to the full tracing UI, when configured.
> 
> We may need to figure out how to set a limit in all of the places we offer traces. The graph side panel, as well as the log and metric tab's `spans` option don't have this.
> 
> To limit our exposure we may even want to introduce a new config option for tracing, something like,`max_duration`. If the user's time period is < max, use the time period, otherwise the max. It would protect against large time period queries that could be slow, even with a low limit I think Tempo will still look at the whole time period.
> 
> Note that in the link above, a later comment mentions that Tempo _may_ look into an option to retrieve newest spans, as opposed to any (or typically older) traces.

Could we still leave the current behavior as an option (Spread the results over the period of time)? This was like this for a long time, and I don't think no one complained about the performance. 

Also, we still want the cache? If we don't perform many search calls probably the query cache is not needed, but I think we can save some calls to the backend with the traceID cache. 

A limit is always specified. In the graph is 15, in the traces tab by default is 100.

Right, for what I've read the results returned in Tempo are not deterministic (https://github.com/grafana/tempo/issues/2659#issuecomment-1638794014), when I've tried, it looks to be returning the latest ones (Maybe because they are more accessible, not sure)

tracing/tempo/cache.go
```diff
@@ -0,0 +1,181 @@
...
+
+func NewTempoCache() TempoCache {
+	kConfig := kialiConfig.Get()
+
+	cacheExpiration := time.Duration(kConfig.ExternalServices.Tracing.TempoConfig.CacheExpiration) * time.Second
+
+	tempoCacheImpl := tempoCacheImpl{
+		cacheExpiration:   cacheExpiration,
+		cacheAppTraces:    make(map[string]*queryResult),
+		cacheTraceDetails: make(map[string]*traceResult),
```
nrfox, line 56: Have you seen the [ExpirationStore](https://github.com/kiali/kiali/blob/master/store/expiration_store.go). It already handles expiring keys after a certain time period and it's threadsafe. The KialiCache uses it for a few different things: https://github.com/kiali/kiali/blob/31bcc777ea265c4ec971f25608ecbbf52c291ca8/kubernetes/cache/cache.go#L146 It doesn't show you stats but you can add that if needed.

tracing/tempo/http_client.go
```diff
@@ -26,9 +27,26 @@ type OtelHTTPClient struct {
 	ClusterTag bool
 }
 
+var (
+	once       sync.Once
+	tempoCache TempoCache
+)
```
nrfox, line 33: Can we avoid global variables and instead pass the cache into the client? This would make it easier to test as well and will avoid side effects for anything that is testing this package.

jshaughn:
> Could we still leave the current behavior as an option (Spread the results over the period of time), even if this is not default? This was like this for a long time, and I don't think no one complained about the performance. IMO, it improves the user experience providing results over the specified period of time.

My feeling is that we should not keep this behavior as an option. To me it increases risk of a performance issue and still doesn't really guarantee behavior that is easy to understand.  When I see those "stovepipe" scatter-plot charts, I really don't understand what I'm looking at.  I think this reduces risk, simplifies both the code and UX, and reduces load on the Tempo back-end.

I don't really have a problem with users seeing a 2m period of traces, even when the current time period may be 1h.  We offer two ways to solve this problem: change the limit and/or change the time period.   I do think we should have a maximum limit, to protect against fetching too many spans. If the max limit doesn't solve the problem, then the user should change the time period.  **Remember**, we do support custom time ranges!  And so if a user really wants to see older traces, they can set the time period to be shorter and in the past.

There are some things we can do to help users understand what they are looking at. Some ideas:
* We can add more info icons.
* We can suggest increasing Limit when retrieved traces cover only a small portion of the current time period.
* We can enhance the "Spans" checkbox on Logs and Metrics pages to show a Limit dropdown when enabled.
* And as always, we can launch the native UI if they configure the external_url.

This is my opinion, but this is a team effort and if you feel strongly, let's open this up to a wider audience for a discussion and decision.

> Also, we still want the cache? If we don't perform many search calls probably the query cache is not needed, but I think we can save some calls to the backend with the traceID cache.

After we arrive on an approach we can revisit what sort of caching, if any, will benefit the users and/or reduce load on the Tempo back-end.  I'm not opposed to client-side caching, even if it is in addition to server-side tuning.

> A limit is always specified. In the graph is 15, in the traces tab by default is 100.

Definitely we always want a default limit. As I said above, I think we also want a max limit, and I think we should make it user-selectable.

josunect:
> > Could we still leave the current behavior as an option (Spread the results over the period of time), even if this is not default? This was like this for a long time, and I don't think no one complained about the performance. IMO, it improves the user experience providing results over the specified period of time.
> 
> My feeling is that we should not keep this behavior as an option. To me it increases risk of a performance issue and still doesn't really guarantee behavior that is easy to understand. When I see those "stovepipe" scatter-plot charts, I really don't understand what I'm looking at. I think this reduces risk, simplifies both the code and UX, and reduces load on the Tempo back-end.
> 
> I don't really have a problem with users seeing a 2m period of traces, even when the current time period may be 1h. We offer two ways to solve this problem: change the limit and/or change the time period. I do think we should have a maximum limit, to protect against fetching too many spans. If the max limit doesn't solve the problem, then the user should change the time period. **Remember**, we do support custom time ranges! And so if a user really wants to see older traces, they can set the time period to be shorter and in the past.
> 
> There are some things we can do to help users understand what they are looking at. Some ideas:
> 
>     * We can add more info icons.
> 
>     * We can suggest increasing Limit when retrieved traces cover only a small portion of the current time period.
> 
>     * We can enhance the "Spans" checkbox on Logs and Metrics pages to show a Limit dropdown when enabled.
> 
>     * And as always, we can launch the native UI if they configure the external_url.
> 
> 
> This is my opinion, but this is a team effort and if you feel strongly, let's open this up to a wider audience for a discussion and decision.
> 
> > Also, we still want the cache? If we don't perform many search calls probably the query cache is not needed, but I think we can save some calls to the backend with the traceID cache.
> 
> After we arrive on an approach we can revisit what sort of caching, if any, will benefit the users and/or reduce load on the Tempo back-end. I'm not opposed to client-side caching, even if it is in addition to server-side tuning.
> 
> > A limit is always specified. In the graph is 15, in the traces tab by default is 100.
> 
> Definitely we always want a default limit. As I said above, I think we also want a max limit, and I think we should make it user-selectable.

@jshaughn I've created this related PR for this change. 
Should we close this PR or revisit after [the linked](https://github.com/kiali/kiali/pull/7891) PR?

jshaughn:
> Should we close this PR or revisit after https://github.com/kiali/kiali/pull/7891 PR?

Let's leave this open and revisit the need for caching when we are done with the other changes.

jshaughn:
> Although I have some doubts if the span data might be added progressively, we might lost some spans?

Hmm, that's interesting.  I suppose it's possible that we could cache a trace and then not pick up spans added after that initial fetch. I guess if you wanted to add a TTL you still could, but instead of having a separate reaper, you could just clear dead entries when checking the cache.

store/fifo_store.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+type StoreStats struct {
+	Hits          int
+	TotalRequests int
+	Size          int
+}
+
+// FifoStore uses a FIFO approach storage and is safe for concurrent use.
+type FifoStore[K comparable, V any] struct {
```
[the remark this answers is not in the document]
nrfox: `FIFOStore` should implement the `Store` interface but like I said if you embed the `Store` interface in the `FIFOStore` struct then you will get all the methods you don't explicitly implement "for free".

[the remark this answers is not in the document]
nrfox: Actually you'll probably want the expirationStore last since when keys expire you want the order to be updated/preserved:
```
vanillaStore := store.New[string, *model.TracingSingleTrace]()
fifoStore := store.NewFIFOStore(vanillaStore, capacity)
tempoCache.Store = store.NewExpirationStore(ctx, fifoStore, nil, nil)
```

josunect:
Thanks for the suggestion, @nrfox !
I've made the changes

tracing/tempo/http_client.go
```diff
@@ -50,13 +61,24 @@ func NewOtelClient(client http.Client, baseURL *url.URL) (otelClient *OtelHTTPCl
 		}
 	}
 
-	return &OtelHTTPClient{ClusterTag: tags}, nil
+	otelHTTPClient := &OtelHTTPClient{ClusterTag: tags}
+	if config.Get().ExternalServices.Tracing.TempoConfig.CacheEnabled {
+		vanillaStore := store.New[string, *model.TracingSingleTrace]()
```
nrfox, line 66: I gave you a bad example earlier. Probably should call this something other than `vanilla` :). Even just `s`  I think is fine since you are using it immediately after. You also should only need to give the type params once here since they can be inferred on the rest of the store.New... methods:
```
s := store.New[string, *model.TracingSingleTrace]()
fifoStore := store.NewFIFOStore(s, config.Get().ExternalServices.Tracing.TempoConfig.CacheCapacity)
otelHTTPClient.TempoCache = store.NewExpirationStore((context.Background(), fifoStore, util.AsPtr(TTL), util.AsPtr(expirationCheckInterval))
```

tracing/tempo/http_client.go
```diff
@@ -50,13 +61,24 @@ func NewOtelClient(client http.Client, baseURL *url.URL) (otelClient *OtelHTTPCl
...
+		fifoStore := store.NewFIFOStore[string, *model.TracingSingleTrace](vanillaStore, config.Get().ExternalServices.Tracing.TempoConfig.CacheCapacity)
+		otelHTTPClient.FIFOStore = fifoStore
+		otelHTTPClient.TempoCache = store.NewExpirationStore[string, *model.TracingSingleTrace](context.Background(), fifoStore, util.AsPtr(TTL), util.AsPtr(expirationCheckInterval))
```
nrfox, line 69: It'd be nice to pass the context into `NewOtelClient` and down through here. Otherwise the goroutine that expires keys will never stop.

frontend/src/pages/Mesh/target/TargetPanelNode.tsx
```diff
@@ -23,15 +26,31 @@ export const TargetPanelNode: React.FC<TargetPanelNodeProps<MeshNodeData>> = (
...
+              {props.tracingStats.size && (
+                <div style={{ marginLeft: '20px' }}>
+                  <span style={{ color: 'rgb(25, 118, 116)', fontWeight: 'bold' }}>Size:</span>{' '}
+                  {props.tracingStats.size}
+                </div>
+              )}
+            </div>
+            {targetPanelHR}
+          </>
+        )}
```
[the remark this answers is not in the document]
josunect: I didn't know Kiali also emits its own prom metrics. So, probably can be done in the same way and store this value as a metric. 
I though the Mesh page because that is intended for admins, which can be also a good place to made them know that he cache is active and if it is useful or not.

tracing/tempo/http_client.go
```diff
@@ -22,13 +25,21 @@ import (
...
 
+const (
+	expirationCheckInterval = 1 * time.Minute
+	TTL                     = 5 * time.Minute
+)
+
 type OtelHTTPClient struct {
 	ClusterTag bool
+	TempoCache store.Store[string, *model.TracingSingleTrace]
+	FIFOStore  *store.FIFOStore[string, *model.TracingSingleTrace]
```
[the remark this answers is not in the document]
josunect: It was keeping a reference to access to the stats, but I've changed for metrics, it is not needed now

tracing/tempo/http_client.go
josunect: For some reason I have a warning in my IDE, that's why I added it: 
image

josunect:
@nrfox I've changed the approach to use 2 different metrics, hits/total requests, so the rate can be obtained: 

`(sum(kiali_cache_hits_total{name="tempo"}) / sum(kiali_cache_requests_total{name="tempo"})) * 100`

image

store/fifo_store.go
```diff
@@ -0,0 +1,97 @@
...
+	log.Tracef("[FIFO store] Returning from cache: %v", key)
+	f.stats.Hits++
+	return elem, true
+}
+
+// Set
+func (f *FIFOStore[K, V]) Set(key K, value V) {
+	if _, exists := f.Store.Get(key); exists {
+		return
+	}
```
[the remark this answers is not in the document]
josunect: Ahh ok, so I changed to update the method when calling set and the key already exists

store/fifo_store.go
```diff
@@ -0,0 +1,97 @@
...
+	Hits          int
+	TotalRequests int
+	Size          int
+}
+
+// FIFOStore uses a FIFO approach storage and is safe for concurrent use.
+type FIFOStore[K comparable, V any] struct {
+	capacity int
+	Store[K, V]
+	order *list.List
```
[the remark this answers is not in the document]
josunect: Added a lock/unlock and Replace

tracing/tempo/http_client.go
josunect: Updated to pass the context

store/fifo_store.go
```diff
@@ -43,30 +45,52 @@ func (f *FIFOStore[K, V]) Get(key K) (V, bool) {
...
+			f.lock.Unlock()
 			f.Store.Remove(oldest.Value.(K))
 		}
 	}
 
+	f.lock.Lock()
 	f.order.PushBack(key)
+	f.lock.Unlock()
 	f.Store.Set(key, value)
 }
```
nrfox: A question on the behavior of `Set`:

If an element exists and you call `Set()` on that element, what should happen to the `order` of that element? Right now it looks like it will save the order twice in `f.order`. e.g.
```
fifo.Set("color", "white")
fifo.Set("color", "white")
fifo.order.Len() <-- 2
len(fifo.Store.Keys()) <-- 1
```
I don't think that's correct because then you'll have a mismatch between the elements in `f.Store` and `f.order`. If the same element is accessed many times `f.order` could be full of just one element.

To hopefully keep things simple, I think what we want when the key is found is to update the value of the key in `f.Store` but leave the order unchanged. You'll probably want to first check the capacity and remove the oldest key before checking if the key exists or not. Also to make things simpler, you should be able to `lock/unlock` once at the beginning of this method with `defer`:
```
f.lock.Lock()
defer f.lock.Unlock()
```
That might also help prevent race conditions where `f.Store` gets updated between calling `f.Store.Get()` and `f.Store.Set()`.

josunect: That makes sense!

store/fifo_store.go
```diff
@@ -0,0 +1,96 @@
...
+func (f *FIFOStore[K, V]) Remove(key K) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	for e := f.order.Front(); e != nil; e = e.Next() {
+		if e.Value == key {
+			f.order.Remove(e)
+			break
+		}
+	}
```
nrfox, line 78: Internally `f.order.Remove` should handle iterating to find the element so maybe this can just be
```suggestion
	f.order.Remove(list.Element{Value: key})
```

store/fifo_store.go
```diff
@@ -0,0 +1,96 @@
...
+
+	if !exists {
+		f.order.PushBack(key)
+	}
+	f.Store.Set(key, value)
+}
+
+// Set
+func (f *FIFOStore[K, V]) Remove(key K) {
```
nrfox, line 69: nit: The comment doesn't match the function name.

store/fifo_store.go
```diff
@@ -0,0 +1,96 @@
...
+		return value, false
+	}
+
+	log.Tracef("[FIFO store] Returning from cache: %v", key)
+	internalmetrics.GetCacheHitsTotalMetric(f.name).Inc()
+	return elem, true
+}
+
+// Set
+func (f *FIFOStore[K, V]) Set(key K, value V) {
```
nrfox, line 47: nit: Can you fill out the comment here with the behavior of `Set`?

store/fifo_store.go
```diff
@@ -0,0 +1,96 @@
...
+// Replace replaces the contents of the store with the given map and updates the order list
+func (f *FIFOStore[K, V]) Replace(items map[K]V) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	f.Store.Replace(items)
+	if items == nil {
+		f.order = list.New()
+		return
+	}
```
nrfox, line 91: What is `items != nil`? `f.order` will not be reset and we will keep pushing elements new elements into it?

store/fifo_store.go
```diff
@@ -0,0 +1,96 @@
...
+	f.Store.Remove(key)
+}
+
+// Replace replaces the contents of the store with the given map and updates the order list
+func (f *FIFOStore[K, V]) Replace(items map[K]V) {
+	f.lock.Lock()
+	defer f.lock.Unlock()
+
+	f.Store.Replace(items)
```
nrfox, line 87: What if `len(items) > f.capacity`? Do we need to truncate `items` to the "first `capacity` elements" or something?

store/fifo_store.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"container/list"
+	"sync"
+
+	"github.com/kiali/kiali/log"
+	"github.com/kiali/kiali/prometheus/internalmetrics"
+)
+
+// FIFOStore uses a FIFO approach storage and is safe for concurrent use.
+type FIFOStore[K comparable, V any] struct {
```
nrfox, line 12: nit: I think this type no longer needs to be exported?

store/fifo_store.go
josunect: Added

store/fifo_store.go
josunect: initialized list always

store/fifo_store.go
josunect: Ah, right. Truncating items

store/fifo_store_test.go
```diff
@@ -0,0 +1,160 @@
...
+	_, found1 := expirationStore.Get("foo")
+	require.False(found1)
+	_, found2 := expirationStore.Get("foo2")
+	require.False(found2)
+	_, found3 := expirationStore.Get("foo3")
+	require.False(found3)
+
+}
+
+func TestReplace(t *testing.T) {
```
nrfox, line 127: Can you also test the truncation works? I don't see that in the tests below.

nrfox (commented):
One nit but otherwise LGTM. Thanks for all the work on this!

nrfox (approved):

jshaughn (approved):
This looks really good, @josunect .  I didn't know about the existing ExpirationStore stuff, so it was nice that created a new fifo store, and hooked it into that infra.  And also hooking into the existing prom metrics.  I think it's OK to have the TTL settings to be hardcoded for simplicity.  The cache could always be disabled if there were an issue, and we could add more config later, if needed.

Issue #7839, opened 2024-10-18
Unable to navigate between workloads using the minigraph


ferhoyos:
### Describe the bug

From any workload details page, the user cannot navigate to another workload. Although the user is briefly routed to the new workload, it immediately returns to the previous one.

### Expected Behavior

The user should be able to navigate from one workload to another through the minigraph.

### What are the steps to reproduce this bug?

1. Go to any workload details page (e.g., productpage)
2. In the minigraph, selects another workload node (e.g., reviews-v3) 
3. Verify that the application does not navigate to the selected workload node.

![image](https://github.com/user-attachments/assets/f0b80e8e-c71d-4170-b33a-490170e0b9c0)

### Environment

- **Kiali version:** master
- **Istio version:** 1.23.2
- **Kubernetes impl:** kind
- **Kubernetes version:** 1.29

jshaughn:
This is a tricky issue. It has to do with the fact that when we navigate from one workload to another, the WorkloadDetailsPage component does not get recreated/remounted etc.  Note that we have the same problem navigating from a service to a service.  That in itself should probably not be a problem, but it seems to surface an issue in the way we are dealing with the graph datasource in the detail pages.  I am continuing to look at it, I think we will need to change how things are done and ensure we have more discrete GraphData objects.  I think what is happening is that we re-use the same GraphDataSource, and it updates at unexpected times, causing a mismatch of property data.

PR #7857, opened 2024-10-23
Fix issue with minigraph node navigation

jshaughn:
Protect against minigraph problems when the parent detail page component is changing its target element.

This is a semi-hack to prevent a major refactoring of how we work with the GraphDatasource object in the detail pages. Because we re-use the GraphDatasource, and it's an asynch fetch, it's possible that if the component gets re-used for a new workload/service/app, that things get out of sync.

In short, detect a mismatch between the target component and the datasource, and avoid updating the graph until things get back in sync.

Also:
- remove an unwanted dependency on "graphData" in effect that responds to a change in selectedIds
- slightly shorten the focus node flashing.
- suppress a bunch of console warnings.

Fixes https://github.com/kiali/kiali/issues/7839

### DIFF
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -21,6 +21,9 @@ import 'ace-builds/src-noconflict/theme-twilight';
 // Enables the search box for the ACE editor
 import 'ace-builds/src-noconflict/ext-searchbox';
 
+// A PFT dependency that is noisy console logging (see below)
+import { configure } from 'mobx';
+
 // i18n
 import './i18n';
 
@@ -53,5 +56,10 @@ setRouter([
 if (!window.location.pathname.includes(rootBasename)) {
   router.navigate(`/${window.location.search}`, { replace: true });
 }
+
+// Suppress thousands of console warnings when scrolling over PFT graphs
+configure({
+  enforceActions: 'never'
+});
 
 ReactDOM.render(<RouterProvider router={router} />, document.getElementById('root') as HTMLElement);
--- a/frontend/src/index.tsx
+++ b/frontend/src/index.tsx
@@ -21,9 +21,6 @@ import 'ace-builds/src-noconflict/theme-twilight';
 // Enables the search box for the ACE editor
 import 'ace-builds/src-noconflict/ext-searchbox';
 
-// A PFT dependency that is noisy console logging (see below)
-import { configure } from 'mobx';
-
 // i18n
 import './i18n';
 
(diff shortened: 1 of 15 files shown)


frontend/src/index.tsx
```diff
@@ -54,4 +57,9 @@ if (!window.location.pathname.includes(rootBasename)) {
   router.navigate(`/${window.location.search}`, { replace: true });
 }
 
+// Suppress thousands of console warnings when scrolling over PFT graphs
+configure({
+  enforceActions: 'never'
+});
```
nrfox, line 63: Should we be suppressing these or are the warnings indicative of a problem with our implementation?

jshaughn: I think the issue is rooted in PFT.  We get thousands of warnings as we move the cursor over the graph.  It seems to be a PFT thing, not something we are doing.

nrfox: I can open an issue with PFT. I've noticed these warnings too and wondered about them but didn't do anything about it.

jshaughn: Before opening an issue, I guess I should ensure it's not somehow our problem.  I think we can leave it suppressed while we look into it.  I'm slight;y suspicious now, because when mousing around in their "surge" demo, I don't see the warnings.

nrfox: I only see the warnings in dev mode I don't see them with the production build of the UI.

jshaughn: Well, now I'm not sure.  This happens when we call `elem.setData({some new set of data})`.  But I'm not sure why that's a problem.  the message says `Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed`.  So, maybe it's an issue in PFT's impl of `setData()`, because under the covers the `data` field is tied to a mobX Observable.  Maybe I'll ask about this in PFT Slack...

jshaughn: OK, well that was a pita, but I believe we've gotten rid of the warnings and now perform element updates in the way PFT wants us to do it.  This may give us a small boost in responsiveness as the user mouses across the graph, not sure, but now we batch together a bunch of highlighting updates.

code is pushed.

ferhoyos (changes requested):
Thanks @jshaughn for fixing the Cypress tests! Glad to see the PFT warning has been removed as well. While testing the PR again, I found another place where the `setObserved` function needs to be called.

**GraphPF.tsx - line 545** and **Mesh.tsx - line 402**
```
if (e.getType() !== 'graph') {
        setObserved(() => {
           controller.removeElement(e);
        });
}
```

jshaughn:
@ferhoyos good catch!  pushed...

ferhoyos (approved):
LGTM

Issue #7844, opened 2024-10-18
failed to get istio deployment status when using IP in external services URLs


jmazzitelli:
Things appear to work with cluster_wide_access set to true. But set it to false with discovery selectors of:

```yaml
    discovery_selectors:
      default:
      - matchExpressions:
        - key: istio-injection
          operator: Exists
```

and I see this error in the Kiali UI message center:

![image](https://github.com/user-attachments/assets/34546fde-d160-4a6b-8e08-096daa2bc056)

Pod logs:

```
$ kubectl logs -l app=kiali -n istio-system --tail=-1
2024/10/18 14:07:58 maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
2024-10-18T14:07:58Z INF Kiali: Version: v1.90.0-SNAPSHOT, Commit: 7337eb87334518c09f28433611fca7d5031f3d88, Go: 1.22.5
2024-10-18T14:07:58Z DBG Kiali: Command line: [/opt/kiali/kiali -config /kiali-configuration/config.yaml]
2024-10-18T14:07:58Z DBG Reading YAML config from [/kiali-configuration/config.yaml]
2024-10-18T14:07:58Z DBG Custom dashboards [count=21, enabled=true]: envoy,go,istiod,kiali,micrometer-1.0.6-jvm,micrometer-1.0.6-jvm-pool,micrometer-1.1-jvm,microprofile-1.1,microprofile-x.y,nodejs,quarkus,springboot-jvm,springboot-jvm-pool,springboot-tomcat,thorntail,tomcat,vertx-client,vertx-eventbus,vertx-jvm,vertx-pool,vertx-server
2024-10-18T14:07:58Z DBG Credentials loaded from secret file [/kiali-override-secrets/login-token-signing-key/value.txt]
2024-10-18T14:07:58Z DBG Cluster name is not set. Attempting to auto-detect the cluster name from the home cluster environment.
2024-10-18T14:07:58Z DBG Unable to read remote secret. It may or may not exist. Error: open /kiali-remote-secret/kiali: no such file or directory. Falling back to in cluster config
2024-10-18T14:07:58Z DBG Creating new Kiali Service Account client
2024-10-18T14:07:58Z DBG Rest perf config QPS: 175.000000 Burst: 200
2024-10-18T14:07:58Z DBG Auto-detected the istio cluster name to be [Kubernetes]. Updating the kiali config
2024-10-18T14:07:58Z INF Using authentication strategy [anonymous]
[...]

PR #7865, opened 2024-10-25
protect against IP internal_url in mesh service discovery

jshaughn:
In mesh discovery of external_services, if the internal_url is IP-based and not in one of the expected "namespace.service" formats then don't attempt auto-discovery of the service.

Fixes https://github.com/kiali/kiali/issues/7844


jmazzitelli (approved):
pretty ingenious... have a regex actually match ##.##.##.##  but without matching groups so the length check will fail even though the regex matches, hence its skipped.

BTW: just a side note in case anyone asks.. ipv6 won't match anything so it will be skipped too (which is what we want).

Issue #7690, opened 2024-08-30
Istio validations inconsisteny - exported to other namespace


hhovsepy:
### Describe the bug

Istio config details page validations are not executed for 2 objects which are exported to different namespace.

### Expected Behavior
Even if Istio configs are exported to another namespaces, the validations should be done.

### What are the steps to reproduce this bug?

1. Create 'bookinfo' and 'bookinfo2' namespaces.
2. In 'bookinfo' create 2 DRs with different names but similar content both exporting to 'bookinfo2'
```
apiVersion: networking.istio.io/v1
kind: DestinationRule
metadata:
  name: test
spec:
  host: '*.googleapis.com'
  trafficPolicy:
    loadBalancer:
      simple: ROUND_ROBIN
    portLevelSettings:
      - port:
          number: 443
        tls:
          mode: SIMPLE # initiates HTTPS when accessing www.googleapis.com
  exportTo:
  - 'bookinfo2'
---
apiVersion: networking.istio.io/v1
kind: DestinationRule
metadata:
  name: test2
spec:
  host: '*.googleapis.com'
  trafficPolicy:
    loadBalancer:
      simple: ROUND_ROBIN
    portLevelSettings:
      - port:
          number: 443
        tls:
          mode: SIMPLE # initiates HTTPS when accessing www.googleapis.com
  exportTo:
  - 'bookinfo2'
  ```
3. You will see validation warning in list page, which is correct.
![Screenshot from 2024-08-30 18-23-14](https://github.com/user-attachments/assets/fe6c48fa-1853-442a-9ce9-34458c499490)
But not in a details page, which is bug.

### Environment

- **Kiali version: v1.90 snapshot, and v1.89, probably even older versions**
- **Istio version: 1.23, probably even older versions**

PR #7867, opened 2024-10-25
Config details validations for exportTo namespaces

hhovsepy:
### Describe the change

There was inconsistency in Validations. When 2 Istio Objects are both exportedTo another namespace, their validations in details page was ignored, the details page considers validations only to object's namespace.
But in Istio Config list validations are fine as there are done for all namespaces.

So the fix is: In cases when there is 'exportTo' to another namespace than the current one, validate these namespaces configs as well. Ans still validate the own namespace.

This has no performance impact in cases when not 'exportedTo'.

### Steps to test the PR

Create this DRs in 'bookinfo':
```
kind: DestinationRule
metadata:
  name: test
spec:
  host: '*.googleapis.com'
  trafficPolicy:
    loadBalancer:
      simple: ROUND_ROBIN
    portLevelSettings:
      - port:
          number: 443
        tls:
          mode: SIMPLE # initiates HTTPS when accessing www.googleapis.com
  exportTo:
  - 'bookinfo2'
---
kind: DestinationRule
metadata:
  name: test2
spec:
  host: '*.googleapis.com'
  trafficPolicy:
    loadBalancer:
      simple: ROUND_ROBIN
    portLevelSettings:
      - port:
          number: 443
        tls:
          mode: SIMPLE # initiates HTTPS when accessing www.googleapis.com
  exportTo:
  - 'bookinfo2'
```

In both configs there should be the validations warning in List and Details pages refering to each other.
Screenshot from 2024-10-29 12-21-35

Before there was no validation in Details page.

### Automation testing
integration tests

### Issue reference
https://github.com/kiali/kiali/issues/7690


leandroberetta (approved):
Looks good and works fine.

PR #7873, opened 2024-10-29
[ci] fallback to olm v0.28.0 due to bug in v0.29.0

jmazzitelli:
There is a bug in OLM v0.29.0 that causes OLM not to start. Use v0.28.0 for now until OLM is fixed.

See: https://github.com/operator-framework/operator-lifecycle-manager/issues/3419

(note: we need this PR because as of now our CI molecule test runs are failing because of this issue)


jshaughn (approved):
This is the thing dgrimm found as well, right? LGTM

PR #7874, opened 2024-10-30
Add more types to mesh page components

nrfox:
### Describe the change

Adds more typing to the mesh page components. This gets rid of many of the `as ...` casting and makes it easier to work with the various components because now the IDE can autocomplete what `getData()` actually returns in many places. Mostly it adds [discriminated union types](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#discriminated-unions) to `types/mesh.ts`.

### Steps to test the PR

Regression testing the mesh page.

### Automation testing

N/A


frontend/src/pages/Mesh/target/TargetPanelNode.tsx
```diff
@@ -1,25 +1,28 @@
...
+};
 
-export const TargetPanelNode: React.FC<TargetPanelNodeProps> = (props: TargetPanelNodeProps) => {
+export const TargetPanelNode: React.FC<TargetPanelNodeProps<MeshNodeData>> = (
+  props: TargetPanelNodeProps<MeshNodeData>
+) => {
   const { t } = useKialiTranslation();
 
-  const node = props.target.elem as Node<NodeModel, any>;
+  const node = props.target.elem;
```
ferhoyos, line 19: To be consistent with the `TargetPanelDataPlane` component:

```
const node = props.target;

  if (!node) {
    return null;
  }

  const data = node.elem.getData()!;
```

frontend/src/types/Mesh.ts
```diff
@@ -35,22 +37,81 @@ export enum MeshInfraType {
...
+
+export interface DataPlaneNodeData extends BaseNodeData {
+  infraData: NamespaceInfo[];
+  infraType: MeshInfraType.DATAPLANE;
+}
+
+export interface GrafanaNodeData extends BaseNodeData {
+  // Grafana node data is the raw grafana config. We don't actually care about what
+  // each field is since we just display the whole config in the side panel as is.
+  infraData: any;
```
ferhoyos, line 68: If possible, it is better to use `unknown`, which is the type-safe counterpart of any.

```suggestion
  infraData: unknown;
```

frontend/src/pages/Mesh/MeshElems.tsx
```diff
@@ -253,7 +254,7 @@ export const elems = (c: Controller): { edges: Edge[]; nodes: Node[] } => {
 
 // TODO: When/if it is fixed this can be replaced with a straight call to node.getAllNodeChildren();
 // https://github.com/patternfly/patternfly-react/issues/8350
-export const descendents = (node: Node): Node[] => {
+export const descendents = <E extends NodeModel = NodeModel, D = any>(node: Node<E, D>): Node<E, D>[] => {
```
ferhoyos, line 257: This method is duplicated for the Patternfly graph. We should apply the same type, or, even better, use the same Patternfly helpers for both graphs (traffic and mesh) :slightly_smiling_face:

frontend/src/types/Mesh.ts
```diff
@@ -142,13 +204,41 @@ export interface DecoratedMeshElements {
...
+  type: MeshType.Node;
 }
 
+export interface MeshControllerTarget {
+  elem: Controller | undefined;
+  type: MeshType.Mesh;
+}
+
+export interface EdgeTarget {
+  elem: any;
```
ferhoyos, line 227: If possible, it is better to use unknown, which is the type-safe counterpart of any.

```suggestion
  elem: unknown;
```

ferhoyos (changes requested):
Wow, nice type refactoring, @nrfox! The code is now type-safer and easier to read. I've added a few nitpicking comments :slightly_smiling_face:

frontend/src/pages/Mesh/MeshElems.tsx
nrfox: I moved this method to `helpers/GraphHelpers.tsx`. Not sure if that's the best place for it but at least it's now shared between the two graph pages. There's other methods that look like they are shared between the two graphs that could move there as well but I left those untouched in this PR.

jshaughn: The history here is just that the PFT traffic graph was still in development when we started the mesh page, so I just copied the whole thing as a starting point for Mesh.  I don't know if we want to go down that path of unifying PFT utilities, but I'm sort of -1 on just partially moving one util. In general I'm not too aggressive about avoiding code duplication because I like compartmentalized feature code.  It's easier to tweak and locate as needed, and I'm typically willing to deal with the fix-in-multiple-places as a trade-off.  Having said that, there are definitely helpers in MeshElems and ElemsPF that are purely at the PFT level, and could be candidates for sharing.

nrfox: @jshaughn so what would you prefer here? Move this method back into the individual graphs and copy the type signature update to both _or_ copy the rest of the helper methods to `graph/helpers`? I'm fine with either.

jshaughn: OK, since you started, let's just start consolidating.  Take the entire "// PFT Helpers section" from GraphPFElems.tsx and put it in the new GraphHelpers. This version of the code has a fix needed for traffic graph that should be irrelevant to Mesh graph.  Delete that whole section from MeshElems.tsx.

ferhoyos: Yes, this is a trade-off between isolation and avoiding code duplication. I understand that code duplicated across both meshes could change in the future, but I’m +1 on unifying helpers and utilities that are exclusively related to Patternfly topology and will always apply to both graphs. We could create a new issue/PR, moving all helpers is out of scope of this PR.

nrfox: Alright I moved all the helpers.

ferhoyos (approved):
LGTM

Issue #7830, opened 2024-10-17
[CI] Use the sail operator instead of istioctl to deploy istio


nrfox:
Use the Sail Operator to deploy istio in Kiali's testing pipelines.

ScriptingShrimp:
we need to put this into to our hack scripts when using OCP and sail operator
```
./hack/use-openshift-prometheus.sh -in istio-system -n alpha -ml ossm-3 -kcns istio-system -np false
./hack/use-openshift-prometheus.sh -in istio-system -n beta -ml ossm-3 -kcns istio-system -np false
./hack/use-openshift-prometheus.sh -in istio-system -n gamma -ml ossm-3 -kcns istio-system -np false
./hack/use-openshift-prometheus.sh -in istio-system -n default -ml ossm-3 -kcns istio-system -np false
./hack/use-openshift-prometheus.sh -in istio-system -n bookinfo -ml ossm-3 -kcns istio-system -np false
./hack/use-openshift-prometheus.sh -in istio-system -n sleep -ml ossm-3 -kcns istio-system -np false
```
now its only in Jenkins pipelines

mkralik3:
@ScriptingShrimp IMHO, we don't need this in upstream. Upstream uses istio addons: https://github.com/kiali/kiali/blob/master/hack/istio/install-istio-via-sail.sh#L113-L121 

we need that only in downstream ( where productized, ocp build in, Prometheus is used )

ScriptingShrimp:
Okay, with some further clarification with Matej, we push the fix only into 2.4 branch. Later it will be most likely cherry picked into master 👍

PR #7876, opened 2024-10-31
[CI] Install Istio with the Sail Operator

nrfox:
### Describe the change

Installs Istio with the Sail Operator rather than `istioctl` for the backend integration tests. The other test suites can be updated later.

### Steps to test the PR

```bash
hack/run-integration-tests.sh --test-suite backend
```

Ensure tests complete successfully.

### Automation testing

N/A

### Issue reference

#7830

### DIFF
--- a/hack/istio/install-istio-via-sail.sh
+++ b/hack/istio/install-istio-via-sail.sh
@@ -1,0 +1,109 @@
+#!/bin/bash
+
+set -e
+
+if ! command -v yq &> /dev/null; then
+  echo "yq is required but it is either not installed or not in the PATH."
+  exit 1
+fi
+
+CUSTOM_INSTALL_SETTINGS=""
+PATCH_FILE=""
+
+# process command line args
+while [[ $# -gt 0 ]]; do
+  key="$1"
+  case $key in
+    -pf|--patch-file)
+      PATCH_FILE="$2"
+      shift;shift
+      ;;
+    -s|--set)
+      if [ -n "${CUSTOM_INSTALL_SETTINGS}" ]; then
+        CUSTOM_INSTALL_SETTINGS="${CUSTOM_INSTALL_SETTINGS} | $2"
+      else
+        CUSTOM_INSTALL_SETTINGS="$2"
+      fi
+      shift;shift
+      ;;
+    -h|--help)
+      cat <<HELPMSG
+Valid command line arguments:
+  -pf|--patch-file <name=value>:
+       filepath to a yaml file of an 'Istio' resource that will overlay the default 'Istio' resource. 
+       --patch-file /path/to/patch-file.yaml
+  -s|--set <name=value>:
+       Sets a name/value pair for a custom install setting. Some examples you may want to use:
+       --set installPackagePath=/git/clone/istio.io/installer
+  -h|--help:
+       this message
+HELPMSG
+      exit 1
+      ;;
+    *)
+      echo "ERROR: Unknown argument [$key]. Aborting."
+      exit 1
+      ;;
+  esac
+done
+
+helm upgrade sail-operator sail-operator \
+  --install \
+  --create-namespace \
+  --namespace sail-operator \
+  --repo https://istio-ecosystem.github.io/sail-operator
+
+K8S_GATEWAY_API_VERSION=$(curl --head --silent "https://github.com/kubernetes-sigs/gateway-api/releases/latest" | grep "location: " | awk '{print $2}' | sed "s/.*tag\///g" | cat -v | sed "s/\^M//g")
+echo "Installing Gateway API version ${K8S_GATEWAY_API_VERSION}"
+kubectl apply -k "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=${K8S_GATEWAY_API_VERSION}"
+
+ISTIO_YAML=$(
+cat <<EOF
+apiVersion: sailoperator.io/v1alpha1
+kind: Istio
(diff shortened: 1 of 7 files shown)


hack/istio/install-istio-via-sail.sh
```diff
@@ -0,0 +1,109 @@
+#!/bin/bash
+
+set -e
+
+if ! command -v yq &> /dev/null; then
```
jmazzitelli, line 5: helm may also not be installed or in PATH, I would check that, too.

hack/istio/install-istio-via-sail.sh
```diff
@@ -0,0 +1,109 @@
...
+      exit 1
+      ;;
+  esac
+done
+
+helm upgrade sail-operator sail-operator \
+  --install \
+  --create-namespace \
+  --namespace sail-operator \
+  --repo https://istio-ecosystem.github.io/sail-operator
```
jmazzitelli, line 54: This assumes the sail-operator repo is already added to helm, which likely isn't going to be the case (almost certainly for the first time someone runs this). I would also update the helm repo just as a convienence, in order to pick up the latest updates for the user if they already added the repo earlier:

So above this `helm upgrade` I would do this:

```bash
if ! helm repo list | grep -q "sail-operator"; then
    echo "Adding Helm repo 'sail-operator'"
    echo helm repo add "sail-operator" "https://istio-ecosystem.github.io/sail-operator"
else
    echo helm repo update "sail-operator"
fi
```

hack/istio/install-istio-via-sail.sh
```diff
@@ -0,0 +1,109 @@
...
+kubectl apply -k "github.com/kubernetes-sigs/gateway-api/config/crd/experimental?ref=${K8S_GATEWAY_API_VERSION}"
+
+ISTIO_YAML=$(
+cat <<EOF
+apiVersion: sailoperator.io/v1alpha1
+kind: Istio
+metadata:
+  name: default
+spec:
+  version: v1.23.0
```
jmazzitelli, line 67: I would make this configurable via a command line option `--istio-version` (it can default to `v1.23.0` but we are certainly going to want to be able to install different versions in the future).

hack/istio/install-istio-via-sail.sh
```diff
@@ -0,0 +1,109 @@
...
+  namespace: istio-system
+  updateStrategy:
+    type: InPlace
```
jmazzitelli, line 70: I would also consider making this configurable. I know in the past I wanted to switch between InPlace and RevisionBased just to test the different behavior.

hack/istio/install-istio-via-sail.sh
```diff
@@ -0,0 +1,109 @@
...
+fi
+
+kubectl get ns istio-system || kubectl create ns istio-system
+kubectl apply -f - <<<"$ISTIO_YAML"
+kubectl wait --for=condition=Ready istios/default -n istio-system
+
+# Install addons
+addons=("prometheus" "grafana" "jaeger")
+for addon in "${addons[@]}"; do
+  kubectl apply -n istio-system -f "https://raw.githubusercontent.com/istio/istio/refs/heads/release-1.23/samples/addons/$addon.yaml"
```
jmazzitelli, line 108: Now that if we make the istio revision configurable, this would need to be changed also.. to something like:
```suggestion
  kubectl apply -n istio-system -f "https://raw.githubusercontent.com/istio/istio/refs/heads/release-${ISTIO_VERSION}/samples/addons/$addon.yaml"
```

jmazzitelli (changes requested):

hack/istio/install-istio-via-sail.sh
nrfox: The `--repo https://istio-ecosystem.github.io/sail-operator` flag will use the url instead of looking at your local charts repo.

hack/istio/install-istio-via-sail.sh
nrfox: Actually so that's the point of these two options:
```
  -pf|--patch-file <name=value>:
       filepath to a yaml file of an 'Istio' resource that will overlay the default 'Istio' resource. 
       --patch-file /path/to/patch-file.yaml
  -s|--set <name=value>:
       Sets a name/value pair for a custom install setting. Some examples you may want to use:
       --set 
```

The YAML in this file is just the default YAML if you don't apply any override but you can override any part of the YAML using either of these two options. With set you use it like this: `--set '.spec.version="v1.22.4"'` to specify a different version. If you're specifying many options you can use the patch file like this:
```
    local patch_file
    patch_file=$(mktemp)
    cat <<EOF > "$patch_file"
spec:
  values:
    global:
      meshID: mesh-default
      network: network-default
      multiCluster:
        clusterName: cluster-default
EOF
    "${SCRIPT_DIR}"/istio/install-istio-via-sail.sh --patch-file "$patch_file"
```

In both cases it just overrides the default YAML in this file and you can override any option you want.

hack/istio/install-istio-via-sail.sh
nrfox: Same comment here: https://github.com/kiali/kiali/pull/7876#discussion_r1836651741

hack/istio/install-istio-via-sail.sh
jmazzitelli: Interesting. So `--repo` will pull the latest charts. TIL...

jmazzitelli (approved):
lgtm

PR #7756, opened 2024-09-18
Appender work related to waypoint service injection

jshaughn:
partof https://github.com/kiali/kiali/issues/7445

Multiple appenders perform service node injection and could have been affected by the new code preventing injection for ztunnel waypoint edges.  Fortunately, only one required changes, the security appender, and in there we were able to make a fairly simple change.

Before: Note that some of the waypoint edges lack a lock icon and show no mTLS wuth graph find:
image

After: The waypoint edges show secure mTLS:
image

PR #7879, opened 2024-11-01
Handle Security appender Ambient corner case

### LINKED ISSUE #7811: kiali doesn't show connection from ambient to sidecar injected namespace as mTLS
### Describe the bug

the scenario is as follows:

- two fresh namespaces - no peer authentication object defined, no network policy defined
- namespace A labeled with istio.io/dataplane-mode=ambient
- namespace B labeled with istio-injection=enabled
- sleep pod with curl installed in namespace A
- ratings pod installed in namespace B
- curl issued from sleep pod in namespace A to ratings pod in namespace B
- response returned, ztunnel logs show the connection as mTLS
- traffic graph in kiali does not show the connection as mTLS
- the same scenario has been tested in different settings: ambient -> sidecar, sidecar -> ambient, ambient -> ambient, sidecar -> sidecar. the only setting where kiali doesn't show the connection as mTLS is in an ambient -> sidecar situation.

curl command:
image

ztunnel logs:
image

traffic graph in kiali:
image

image

### Expected Behavior

assuming the ztunnel logs confirm the connection is mTLS, the traffic graph in kiali should show the same.

follow along with the scenario above to reproduce.

### Environment

- **Kiali version:** 1.89.7
- **Istio version:** 1.23.1
- **Kubernetes impl:** gardener
- **Kubernetes version:** 1.30.5

jshaughn:
Fixes #7811 

Handle Security appender Ambient corner case. For ztunnel-to-sidecar traffic, we get source reporting only.  Historically we always use destination reporting to determine mtls.
- add necessary queries to security_appender
  - unfortunate but I don't see an alternative
- make sure we disable ambient traffic when ambient is not enabled
  - this saves unnecessary queries
- update unit test 

Test:
@josunect , you already have the environment because you were the one to figure out the problem :-)  I think I updated the queries as you suggested, to look for ztunnel source telem when needed.  Let me know if this solves the issue for you, and if the solution looks good.  It does add queries, and those queries may often return duplicate reporting (because ztunnel typically reports both source and dest telem when source and dest are both in the ambient mesh). Despite adding queries and pulling more date, any duplicates should basically overwrite the original with the same result.

### DIFF
--- a/frontend/src/config/ServerConfig.ts
+++ b/frontend/src/config/ServerConfig.ts
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import { ServerConfig } from '../types/ServerConfig';
 import { parseHealthConfig } from './HealthConfig';
 import { MeshCluster } from '../types/Mesh';
+import { TrafficRate } from 'types/Graph';
 
 export type Durations = { [key: number]: string };
 
@@ -169,6 +170,9 @@ export const setServerConfig = (cfg: ServerConfig): void => {
 
   homeCluster = getHomeCluster(serverConfig);
   isMultiCluster = isMC();
+  if (!serverConfig.ambientEnabled) {
+    serverConfig.kialiFeatureFlags.uiDefaults.graph.traffic.ambient = 'none';
+  }
 };
 
 export const isIstioNamespace = (namespace: string): boolean => {
--- a/frontend/src/config/ServerConfig.ts
+++ b/frontend/src/config/ServerConfig.ts
@@ -2,7 +2,6 @@ import _ from 'lodash';
 import { ServerConfig } from '../types/ServerConfig';
 import { parseHealthConfig } from './HealthConfig';
 import { MeshCluster } from '../types/Mesh';
-import { TrafficRate } from 'types/Graph';
 
 export type Durations = { [key: number]: string };
 
--- a/graph/telemetry/istio/appender/security_policy.go
+++ b/graph/telemetry/istio/appender/security_policy.go
@@ -195,6 +195,35 @@ func (a SecurityPolicyAppender) appendGraph(trafficMap graph.TrafficMap, namespa
 			query = fmt.Sprintf(`%s OR (%s)`, query, tcpReceivedQuery)
 		}
 	}
+	// If we are including ztunnel traffic we may need more TCP queries, because for ztunnel-to-sidecar traffic ztunnel
+	// will report as source telemetry. If this results in duplicate source and dest ztunnel telem, it should be OK, as
+	// the rate should be the same, and will just overwrite itself.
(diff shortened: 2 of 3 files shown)


frontend/src/config/ServerConfig.ts
```diff
@@ -169,6 +169,9 @@ export const setServerConfig = (cfg: ServerConfig): void => {
 
   homeCluster = getHomeCluster(serverConfig);
   isMultiCluster = isMC();
+  if (!serverConfig.ambientEnabled) {
+    serverConfig.kialiFeatureFlags.uiDefaults.graph.traffic.ambient = 'none';
```
josunect, line 173: Is this used somewhere else?

graph/telemetry/istio/appender/security_policy.go
```diff
@@ -195,6 +195,35 @@ func (a SecurityPolicyAppender) appendGraph(trafficMap graph.TrafficMap, namespa
 			query = fmt.Sprintf(`%s OR (%s)`, query, tcpReceivedQuery)
 		}
 	}
+	// If we are including ztunnel traffic we may need more TCP queries, because for ztunnel-to-sidecar traffic ztunnel
+	// will report as source telemetry. If this results in duplicate source and dest ztunnel telem, it should be OK, as
+	// the rate should be the same, and will just overwrite itself.
+	if a.Rates.Ambient == graph.AmbientTrafficTotal || a.Rates.Ambient == graph.AmbientTrafficZtunnel {
```
josunect, line 201: I've though it would be useful for this waypoint code: https://github.com/kiali/kiali/pull/7756/files#diff-55173479d7c6bec43bb8b3b60d6b5d67bd408e9d3b8c70ea176c6b75e4eff0cfR317 
But it looks like it is not applicable here.

frontend/src/config/ServerConfig.ts
jshaughn: The value ends up being used when we are setting the traffic options for graph generation requests.  Without this change we will, by default, ask the graph-gen code to include ambient traffic, even if the ambient profile in not enabled.

I considered changing this server-side but decided that it was better to return the real server config from the server, and let clients modify it as they see fit.  We also do some massaging of the health config, so updating the server config does have a precedent.

graph/telemetry/istio/appender/security_policy.go
jshaughn: I don't think so, but I do think we will see more corner cases. We always want to minimize the number of queries, so we have to be careful about what we add. One thing to notice in this appender is that we actually minimize the number of prometheus round trips, because we use OR expressions.  This is sort of a trick, but works as long as each sub-query has the same `by (...)` clause.  When that is true you can actually query different metrics all within the same prom query.

PR #7880, opened 2024-11-01
Be able to tell our test scripts what version of OLM to use.

jmazzitelli:
OLM has a bug that causes it to be uninstallable so we need a way to install an older version to continue being able to test.

Right now the github action molecule tests are failing - that's the reason why I'm doing this now.

See https://github.com/operator-framework/operator-lifecycle-manager/issues/3419#issuecomment-2451589801

(when that bug is fixed and a new OLM release is published, we will follow up and reset to the "latest" olm release in our tests by merging this second PR: https://github.com/kiali/kiali/pull/7881)

### DIFF
--- a/.github/workflows/molecules.yml
+++ b/.github/workflows/molecules.yml
@@ -10,6 +10,12 @@ on:
         required: false
         default: ""
         type: string
+      olm_version:
+        description: "Version of OLM to install or 'latest'. e.g. v0.28.0"
+        required: false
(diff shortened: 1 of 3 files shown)


jmazzitelli:
I ran this on my own fork and it all works - see molecule github action results here: https://github.com/jmazzitelli/kiali/actions/runs/11628596362/job/32384026990#step:5:144

leandroberetta (approved):

Issue #7861, opened 2024-10-24
Improve deployment of remote cluster resources


nrfox:
### What do you want to improve?

When you are deploying Kiali to read from a remote cluster, you need to deploy some resources to give it access. The way that you do that today is by using the [kiali-prepare-remote-cluster.sh](https://github.com/kiali/kiali/blob/master/hack/istio/multicluster/kiali-prepare-remote-cluster.sh) script which in turn calls `helm template` for those resources.

This works well for getting started but it would be nice to have a way of tracking these resources outside of this script either within the helm chart or within the Kiali CR. A concrete example of this is when the RBAC roles in those templates change across versions. Users would need to re-run the script to regenerate the templates. Instead if there was a helm template or the Kiali CR tracked these resources, they would get automatically updated when you update your version of Kiali in the remote cluster.

### What is the new behavior?

Provide a way of deploying/tracking remote resources outside of the hack script. Could be either a separate helm chart/CR or an additional `remoteDeployment` option on the existing chart/CR to only deploy the remote templates.

jmazzitelli:
After thinking about this, I think it would be best to do this with a helm chart. We normally wouldn't want users to install the operator in their remote cluster if all they need is those remote resources installed. The operator is really to manage a full running Kiali server instance... the helm chart seems more natural to use to simply deploy some resources.

I'll think about this some more.... I want to see how easily it can be done in the already existing helm charts we have (as a first attempt at this).. maybe a separate helm chart would be needed. I don't know yet.

jmazzitelli:
@nrfox the hack script creates resources in both clusters (the kiali cluster and the remote cluster). Is this enhancement just for the operator in the remote cluster to create the remote cluster resources (basically, [the SA and its secret, the clusterRole, and the clusterRoleBinding](https://github.com/kiali/kiali/blob/v2.0.0/hack/istio/multicluster/kiali-prepare-remote-cluster.sh#L107-L150))??

How does the user then know how to create the remote cluster secret on the Kiali cluster side (and also to configure Kiali CR for it - this is the [spec.clustering.clusters.secret_name setting](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.clustering.clusters[*].secret_name))?? They could create the secret with the correct label so auto-discovery would work and no change needed in the Kiali CR, but the user would need to know to do that, too. My guess is - the user needs to RTFM. (?)

nrfox:
[...]

PR #7883, opened 2024-11-01
multi-cluster setup - no need for script to show only specific templates - the helm chart has a flag for this now

jmazzitelli:
part of: https://github.com/kiali/kiali/issues/7861


nrfox (dismissed):

jmazzitelli:
@nrfox I think I need to update the script... if on OpenShift, we need it to pass `--set isOpenShift=true` to the `helm template` command.

jmazzitelli:
> @jmazzitelli won't it default to autodiscover?

No, "helm template" doesn't process `.Capabilities` - it always returns `false`. Those `.Capabilities` expressions are always evaluated by "helm install" (NOT "helm template").

And this script uses "helm template"

nrfox (approved):

Issue #7878, opened 2024-10-31
Graph hide can leave orphan edges


jshaughn:
It seems using graph hide can leave orphan edges under certain circumstances.  I currently see it:

1) installing bookinfo in an ambient mesh
2) add a hide of app=prometheus
3) change the graph type

![image](https://github.com/user-attachments/assets/e92dbfca-260b-490e-9afc-1414068c9c28)

note: @jmazzitelli originally saw this in https://github.com/kiali/kiali/pull/7836#issuecomment-2424905214, but at the time I didn't know how to reproduce.

PR #7885, opened 2024-11-02
Fix orphan edge hide issue in graph and mesh page

jshaughn:
PFT Graph and Mesh hide had needed to call e.remove() to detach hidden edges from their parent group. I had inline comments stating that it really shouldn't be necessary. Now, either do to a PFT fix, and/or our recent change to wrap element updates in mobx actions, it seems unnecessary, and seems to be causing this orphan edge issue. Which makes sense because we're detaching it from its parent.

Fixes https://github.com/kiali/kiali/issues/7878

@jmazzitelli note: repro steps are in the issue.  Also, you reported this originally in https://github.com/kiali/kiali/pull/7836#issuecomment-2424905214.


jmazzitelli (approved):
LGTM

PR #7887, opened 2024-11-04
[Ambient] get reporter for tcp telemetry

### LINKED ISSUE #7875: [ambient] External service shown as unknown
### Describe the bug
Using latest Istio code base (as of yesterday) and Kiali , I found my external service (httpbin.org) is shown as unknown.

### Expected Behavior

External service shown as the service defined in SE.

How I installed Istio using istioctl of the dev build below:

```
TAG=1.24-alpha.6a458274241dd51fbd6e015e7f439a29e8beb07f

istioctl install --set tag=$TAG --set hub=gcr.io/istio-testing --set profile=ambient --skip-confirmation  --set meshConfig.accessLogFile=/dev/stdout --set values.pilot.env.PILOT_ENABLE_IP_AUTOALLOCATE=true --set  values.cni.ambient.dnsCapture=true

kubectl get crd gateways.gateway.networking.k8s.io &> /dev/null || \
  { kubectl apply -f https://github.com/kubernetes-sigs/gateway-api/releases/download/v1.2.0/standard-install.yaml }

ISTIO_DOWNLOAD_DIR=~/src/github.com/istio/istio/
kubectl apply -f $ISTIO_DOWNLOAD_DIR/samples/addons/prometheus.yaml
kubectl apply -f $ISTIO_DOWNLOAD_DIR/samples/addons/kiali.yaml
kubectl apply -f $ISTIO_DOWNLOAD_DIR/samples/addons/grafana.yaml
```

Steps to recreate is using John's blog:

https://www.solo.io/blog/egress-gateways-made-easy/

Generate a bunch of traffic from client to httpbin.org.  Launch the kiali dashboard and see the external traffic going through egress waypoint as unknown.

### Environment

- **Kiali version:**
- **Istio version:**
- **Kubernetes impl:** 
- **Kubernetes version:** 

josunect:
### Describe the change

Get the reporter based on ambient rates for tcp telemetry 

### Steps to test the PR

- Start istio with: 
```--set values.pilot.env.PILOT_ENABLE_IP_AUTOALLOCATE=true" --set values.cni.ambient.dnsCapture=true" ```and Ambient profile.

- Add default ns to ambient: `kubectl label ns default istio.io/dataplane-mode=ambient`
- Create a pod in default namespace to sent http and https requests: 
```apiVersion: v1
kind: Pod
metadata:
  name: curl-test
spec:
  containers:
  - name: curl
    image: curlimages/curl
    command: ["sh", "-c", "sleep infinity"]
```

- Create ns istio-egress  
- Add to ambient: `kubectl label ns istio-egress istio.io/dataplane-mode=ambient`
- Create and enroll to waypoint: istioctl waypoint apply --enroll-namespace --namespace istio-egress
- Create a service entry in the istio-egress namespace:
```apiVersion: networking.istio.io/v1
kind: ServiceEntry
metadata:
  name: httpbin.org
spec:
  hosts:
  - httpbin.org
  ports:
  - number: 80
    name: http
    protocol: HTTP
  - number: 443
    name: https
    protocol: HTTPS
  resolution: DNS
```
- Send http and https traffic:
`k exec -it curl-test -- curl http://httpbin.org/get -v`
`k exec -it curl-test -- curl https://httpbin.org/get -v`

Kiali graph: 

image

Example with sidecars: 

image

Note: If the https port is not added to the service entry, the result for the Ambient ns is the following: 
```
k exec -it curl-test -- curl https://httpbin.org/get -v
* Host httpbin.org:443 was resolved.
* IPv6: 2001:2::3
* IPv4: 240.240.0.3
*   Trying [2001:2::3]:443...
* Immediate connect fail for 2001:2::3: Address not available
*   Trying 240.240.0.3:443...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
*  CAfile: /cacert.pem
*  CApath: /etc/ssl/certs
* OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to httpbin.org:443 
* closing connection #0
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to httpbin.org:443 
command terminated with exit code 35 
```
It doesn't happen with the sidecar example.

### Automation testing

### Issue reference

Fixes #7875

### DIFF
--- a/graph/api/api_test.go
+++ b/graph/api/api_test.go
@@ -908,10 +908,10 @@ func mockNamespaceGraph(t *testing.T) (*prometheus.Client, *prometheustest.PromA
 		},
 	}
 
-	q3 := `round(sum(rate(istio_tcp_sent_bytes_total{reporter="source",source_workload_namespace!="bookinfo",destination_workload_namespace="unknown",destination_workload="unknown",destination_service=~"^.+\\.bookinfo\\..+$"} [600s])) by (app,source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
+	q3 := `round(sum(rate(istio_tcp_sent_bytes_total{reporter=~"source|waypoint",source_workload_namespace!="bookinfo",destination_workload_namespace="unknown",destination_workload="unknown",destination_service=~"^.+\\.bookinfo\\..+$"} [600s])) by (app,source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
 	v3 := model.Vector{}
 
-	q4 := `round(sum(rate(istio_tcp_sent_bytes_total{reporter="destination",destination_workload_namespace="bookinfo"} [600s])) by (app,source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
+	q4 := `round(sum(rate(istio_tcp_sent_bytes_total{reporter=~"destination|waypoint",destination_workload_namespace="bookinfo"} [600s])) by (app,source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
 	q4m0 := model.Metric{
 		"source_workload_namespace":      "istio-system",
 		"source_workload":                "ingressgateway-unknown",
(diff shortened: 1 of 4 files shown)


graph/api/api_test.go
```diff
@@ -908,10 +908,10 @@ func mockNamespaceGraph(t *testing.T) (*prometheus.Client, *prometheustest.PromA
 		},
 	}
 
-	q3 := `round(sum(rate(istio_tcp_sent_bytes_total{reporter="source",source_workload_namespace!="bookinfo",destination_workload_namespace="unknown",destination_workload="unknown",destination_service=~"^.+\\.bookinfo\\..+$"} [600s])) by (app,source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
+	q3 := `round(sum(rate(istio_tcp_sent_bytes_total{reporter=~"source|waypoint",source_workload_namespace!="bookinfo",destination_workload_namespace="unknown",destination_workload="unknown",destination_service=~"^.+\\.bookinfo\\..+$"} [600s])) by (app,source_cluster,source_workload_namespace,source_workload,source_canonical_service,source_canonical_revision,destination_cluster,destination_service_namespace,destination_service,destination_service_name,destination_workload_namespace,destination_workload,destination_canonical_service,destination_canonical_revision,response_flags) > 0,0.001)`
```
josunect, line 911: @jshaughn I was wondering why I had to change some of the tests queries but I've realized it uses the default Ambient rates. Maybe some of these tests should use Ambient: none? (There are some specific tests for Ambient, but not all of them)

jshaughn: The other possibility is we reverse our approach, and we change the default to be "none", and enable it only when ambient is enabled.  But if we think that the migration will be to ambient, then maybe it's bestto leave it alone.

jshaughn (changes requested):
I think these changes look good, but I'm not sure this has enough coverage, because we have other places that query against the tcp metrics.  I think you may need to make similar changes in:
- istio.go#buildNodeTrafficMap
  - also in here I see an outdated check for ambient workload 
    - `if strings.HasPrefix(n.Workload, "waypoint") {`
    -  we probably want to replace that with our `isWaypoint` logic.
- security_policy.go

josunect:
> I think these changes look good, but I'm not sure this has enough coverage, because we have other places that query against the tcp metrics. I think you may need to make similar changes in:
> 
>     * istio.go#buildNodeTrafficMap
>       
>       * also in here I see an outdated check for ambient workload
>         
>         * `if strings.HasPrefix(n.Workload, "waypoint") {`
>         * we probably want to replace that with our `isWaypoint` logic.
> 
>     * security_policy.go

Thanks, @jshaughn 
I think my latest commits are addressing these missing updates. 

Regarding that outdated check, I had to add a new condition to skip the cluster comparison if this was `unknown`: 
https://github.com/kiali/kiali/pull/7887/files#diff-611af3df8bfed322d3a0a785476482260509aa84540c2f8d70230b065ddf4820R1084

Otherwise, the graph node for waypoint was empty.

I've noticed (Not from this PR) that when you highlight a http edge it turns blue:
image

Issue #7623, opened 2024-08-13
Consistent display format for configuration values on the Mesh page


ferhoyos:
### What do you want to improve?

UI alignment of the configuration values on the Mesh page.

### What is the current behavior?

Currently, different configurations are displayed in different formats:

- YAML format for Istio config or Kiali status

![image](https://github.com/user-attachments/assets/f4ab64d4-88c0-4a94-b5c9-39f660b30721)

- Table format for Kiali config and other components (Jaeger, Prometheus, Grafana, etc.)

![image](https://github.com/user-attachments/assets/a4b16b8a-45fc-4851-a8c2-81662819e2c2)

### What is the new behavior?

To avoid user confusion and improve UI alignment, all configurations should be displayed in the same format. The YAML format is the best approach because it is already familiar to users and better displays complex configurations (https://github.com/kiali/kiali/issues/7459).

PR #7888, opened 2024-11-04
[mesh page] Always show yaml format config in the Mesh side panels

jshaughn:
Always show yaml format config in the Mesh side panels
- remove table format

Closes https://github.com/kiali/kiali/issues/7623


ferhoyos (changes requested):
Thanks @jshaughn, I wanted that feature in the mesh some time ago :slightly_smiling_face:.

The PR is working fine. However, I don’t like having two different input parameters in the `TargetPanelEditor` component (`configMap` and `configJson`). It’s better to accept only one format, with the parent component handling any necessary formatting if the value is invalid. I’ve refactored the PR a bit in https://github.com/jshaughn/kiali/pull/42.

ferhoyos (approved):
LGTM

PR #7891, opened 2024-11-05
[Tempo] Don't split calls to backend

josunect:
### Describe the change

When the tempo/jaeger search returned more results than the specified in the limit, it was used a split & join mode to spread traces over the requested interval. Don't use this method anymore, and show a warning if the requested start time is at least one minute older than the oldest trace returned from the results:

image

6 hours requested > Showing the last 30 minutes

### Steps to test the PR
We can use Tempo or Jaeger for testing. With Tempo: 

- `minikube start`
- `hack/istio/tempo/install-tempo-env.sh -c kubectl`
- Install Kiali and set up as instructed
-  Install bookinfo to generate traces
- Go to the traces tab for a workload > Select different time ranges

### Automation testing

### Issue reference

Discussion PR Ref. https://github.com/kiali/kiali/pull/7854 
Issue Ref. #7769


jshaughn:
@josunect as an addition to this core work, I put together a PR that allows users to set the trace limit in the places they couldn't before.  I think that's the missing piece to taking away the multiple-queries approach.  You can check it out here: https://github.com/kiali/kiali/pull/7895

frontend/src/components/TracingIntegration/TracesFetcher.ts
```diff
@@ -72,6 +72,15 @@ export class TracesFetcher {
             'Loading traces for all clusters. Tracing is not configured to store traces per cluster.'
           );
         }
+        const firstTraceTimestamp = Math.min(...traces.map(s => s.startTime));
+        // If the first trace starts more than one minute later than the start date that the user selects, show an info message
+        // The order in which traces are returned is not deterministic and
```
jshaughn, line 77: Maybe a better test would be to compare the time period with the delta time of the returned traces?  And if coverage was bad then warn the user.  For example, let's say the requested time period is `Last 30m` and it's bad if the returned traces cover < 90% of the time period.  So, if `t(newestTrace)-t(oldestTrace) < 0.9*30` then show warning.

This basically is saying, if our returned traces cover enough the requested range, then good. It wouldn't matter if they were the oldest or the newest, just that they didn't all lump to one time slice.

josunect:
Inline alert example: 

image

frontend/src/components/Alerts/InlineAlert.tsx
```diff
@@ -0,0 +1,42 @@
...
+  WARNING = 'warning',
+  DANGER = 'danger'
+}
+
+type InlineAlertProps = {
+  message: string;
+  type: alertType;
+};
+
+export const InlineAlert: React.FC<InlineAlertProps> = (props: InlineAlertProps) => {
```
jshaughn, line 16: Why did you decide to use this impl instead of the standard PF inline alert? 

https://www.patternfly.org/components/alert#inline-alerts-variants

frontend/src/components/TracingIntegration/TracesComponent.tsx
```diff
@@ -89,14 +91,20 @@ class TracesComp extends React.Component<TracesProps, TracesState> {
...
-    this.fetcher = new TracesFetcher(this.onTracesUpdated, errors => {
-      // If there was traces displayed already, do not hide them so that the user can still interact with them
-      // (consider it's probably a temporary failure)
-      // Note that the error message is anyway displayed in the notifications component, so it's not going unnoticed
-      if (this.state.traces.length === 0) {
-        this.setState({ tracingErrors: errors, toolbarDisabled: true });
+    this.fetcher = new TracesFetcher(
+      this.onTracesUpdated,
+      errors => {
+        // If there was traces displayed already, do not hide them so that the user can still interact with them
```
jshaughn, line 97:
```suggestion
        // If there were already traces displayed, do not hide them so that the user can still interact with them
```

frontend/src/components/TracingIntegration/TracesFetcher.ts
```diff
@@ -72,6 +73,20 @@ export class TracesFetcher {
             'Loading traces for all clusters. Tracing is not configured to store traces per cluster.'
           );
         }
+        const firstTraceTimestamp = Math.min(...traces.map(s => s.startTime));
+        // If the traces cover less than 90% of the selected period, show an info message
```
jshaughn, line 77: By the way, I just used 90% as an example, it may be a good value but if you think a different value is better then it is fine with me to change.

frontend/src/components/TracingIntegration/TracesFetcher.ts
```diff
@@ -72,6 +73,20 @@ export class TracesFetcher {
...
+        // The order in which traces are returned is not deterministic
+        const endDate = q.endMicros ? q.endMicros : Date.now() * 1000;
+        const timeRange = (endDate - q.startMicros) * 0.9;
+
+        if (firstTraceTimestamp && this.lastFetchMicros && this.lastFetchMicros - firstTraceTimestamp < timeRange) {
+          this.onInfo(
+            `Last ${q.limit} traces shown. To search for traces in a different period, select a custom time range or adjust the traces limit.`
```
jshaughn, line 84:
```suggestion
            `The ${q.limit} traces do not fully cover the time period. You may want to increase the limit, reduce the time range, or define a custom time range.`
```

jshaughn (changes requested):
I like the inline alert more than the toast.  I do have a couple of suggestions.

frontend/src/components/Alerts/InlineAlert.tsx
josunect: That was to use just the pf5 ones, right? https://v5-archive.patternfly.org/components/alert/html
(The one in your link is pf6)

frontend/src/components/TracingIntegration/TracesFetcher.ts
josunect: Right, well, I think it is difficult to adjust. I would say between 80-90% I think

jshaughn: How about 85% :-)

frontend/src/components/TracingIntegration/TracesFetcher.ts
```diff
@@ -72,6 +73,20 @@ export class TracesFetcher {
...
         }
+        const firstTraceTimestamp = Math.min(...traces.map(s => s.startTime));
+        // If the traces cover less than 90% of the selected period, show an info message
+        // The order in which traces are returned is not deterministic
+        const endDate = q.endMicros ? q.endMicros : Date.now() * 1000;
+        const timeRange = (endDate - q.startMicros) * 0.9;
+
+        if (firstTraceTimestamp && this.lastFetchMicros && this.lastFetchMicros - firstTraceTimestamp < timeRange) {
+          this.onInfo(
+            `The ${q.limit} traces do not fully cover the time period. You may want to increase the limit, reduce the time range, or define a custom time range.`
```
jshaughn, line 84: After using this, I think it would be nice if the message indicated the %.  Could you change things so that we could say

`The ${q.limit} traces cover only ${tracePercent}% of the time period. You may want to increase the fetch limit, reduce the time range, or define a custom time range.`

jshaughn (commented):
It all looks good. After playing with it I think a few tweaks will make it even better.

frontend/src/components/TracingIntegration/TracesFetcher.ts
josunect: Updated: 
image

jshaughn (approved):
Looks good to me!

PR #7893, opened 2024-11-05
Cache validations rebase

hhovsepy:
### Describe the change

Explanation of what this PR does

### Steps to test the PR

Recommendations for how to test this PR. Reminder that each PR should also include a "Test" label.

### Automation testing

### Issue reference

If this PR is related to a github issue, please link it here ([more info](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue))


business/istio_validations.go
```diff
@@ -15,135 +15,168 @@ import (
...
-
-	// We fetch without target service as some validations will require full-namespace details
-	go in.fetchIstioConfigList(ctx, &istioConfigList, &mtlsDetails, &rbacDetails, cluster, namespace, errChan, &wg)
-
-	if workload != "" {
-		// load only requested workload
-		go in.fetchWorkload(ctx, &workloadsPerNamespace, cluster, workload, namespace, errChan, &wg)
-	} else {
-		go in.fetchAllWorkloads(ctx, &workloadsPerNamespace, cluster, &namespaces, errChan, &wg)
+	err := in.fetchAllWorkloadsDirect(ctx, &workloadsPerNamespace, cluster, &namespaces)
```
nrfox, line 142: Do we need two different `fetchAllWorkloads` methods?

In the other PR, it uses `GetWorkloadList`: https://github.com/kiali/kiali/pull/7136/files#diff-4afe91ed13da4c9e145fdccf0596bee0b6c0fc2e29d3da22f112584428159dd3R159. Is there a difference between the two?

nrfox: It looks like inside of `fetchAllWorkloadsDirect` we set the value of `&workloadsPerNamespace` and `&namespaces` but we also need `namespaces` to be set further down. Why not just set `namepaces` in this method and pass those down into `fetchAllWorkloads(ctx, cluster, namespaces)`? `fetchAllWorkloads` can return the `map[string]models.WorkloadList` and then you can get around needing to pass a pointer to a map into those methods which is a little awkward to grok.

nrfox (commented):
Thanks for pushing this forward! A couple of minor comments. I haven't tested it out yet but it looks good.

business/istio_validations.go
```diff
@@ -15,135 +15,168 @@ import (
...
-			return nil, fmt.Errorf("Service [namespace: %s] [name: %s] doesn't exist for Validations.", namespace, service)
+			log.Warningf("Error invoking GetService %s", err)
 		}
+		return nil, fmt.Errorf("Service [namespace: %s] [name: %s] doesn't exist for Validations.", namespace, service)
 	}
 
-	// time this function execution so we can capture how long it takes to fully validate this namespace/service
-	timer := internalmetrics.GetValidationProcessingTimePrometheusTimer(namespace, service)
-	defer timer.ObserveDuration()
+	return models.IstioValidations(validationsForCluster(in.kialiCache.Validations().Items(), cluster)).FilterBySingleType(schema.GroupVersionKind{Group: "", Version: "", Kind: "service"}, service), nil
```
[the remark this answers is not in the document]
nrfox: Actually this is how it is currently so just leave it for now.

business/istio_validations.go
hhovsepy: `fetchAllWorkloads` is a wrapper method for `fetchAllWorkloadsDirect` passing all params and designed for async call, while `fetchAllWorkloadsDirect` is for sync call.
For `GetIstioObjectValidations` we use async calls of fetch methods, and I kept those async calls order and method params for 'CreateValidations' sync calls to be consistent and easy to read the code. For this have created 'Direct' methods for sync calls, and the existing methods are now still async but calling newly created direct sync versions of methods and pass all params + WaitGroup, error.

nrfox: I think it's ok that `GetIstioObjectValidations` and `CreateValidations` are different. Having this wrapper method makes things harder to follow. If we want both methods to be the same then we can update both rather than having the additional wrapper on top. I don't remember why I updated `GetValidations` which turned into `CreateValidations` in the first place. Probably async fetching of different things is not necessary since everything is in memory already.

hhovsepy: Yes async fetching objects for `CreateValidations` and cache was causing some backend errors after some time, so the sync call is necessary there.
However  the `GetIstioObjectValidations` still needs async call for performance reasons.

hhovsepy:
The performance improvements are obvious in list pages, up to 50%.

<h4>[Apps List Page]</h4>
Metric | Baseline | Before | After | Improvement

All Namespaces Apps load time | 2.0 sec | 2.8499 sec | 1.5893 sec | ✓ Reduced by 1.2606 sec
Selected Namespaces Apps load time | 2.0 sec | 2.7373 sec | 1.3376 sec | ✓ Reduced by 1.3997 sec

<h4>[Istio Configs List Page]</h4>
Metric | Baseline | Before | After | Improvement

All Namespaces Istio Configs | 2.0 sec | 2.9479 sec | 1.0165 sec | ✓ Reduced by 1.9314 sec
Selected Namespaces Istio Configs | 2.0 sec | 2.5933 sec | 0.4001 sec | ✓ Reduced by 2.1932 sec

<h4>[Services List Page]</h4>
Metric | Baseline | Before | After | Improvement

All Namespaces Services | 2.0 sec | 3.1151 sec | 1.6040 sec | ✓ Reduced by 1.5111 sec
Selected Namespaces Services | 2.0 sec | 2.8106 sec | 1.0684 sec | ✓ Reduced by 1.7422 sec

<h4>[Workloads List Page]</h4>
Metric | Baseline | Before | After | Improvement

All Namespaces Workloads | 2.0 sec | 2.9952 sec | 1.6074 sec | ✓ Reduced by 1.3878 sec
Selected Namespaces Workloads | 2.0 sec | 2.7508 sec | 1.1108 sec | ✓ Reduced by 1.64 sec

business/istio_validations.go
nrfox: > Yes async fetching objects for CreateValidations and cache was causing some backend errors after some time, so the sync call is necessary there.
However the GetIstioObjectValidations still needs async call for performance reasons.

Can we keep the two separate then without the wrappers?

nrfox:
@hhovsepy I think you'll need to rebase this to pickup https://github.com/kiali/kiali/pull/7908 which has e2e test fixes.

frontend/cypress/integration/common/istio_config.ts
```diff
@@ -620,27 +630,34 @@ function waitUntilConfigIsVisible(attempt: number, crdInstanceName: string, crdN
...
 }
 
 Then(
   'the {string} {string} of the {string} namespace should have a {string}',
   (crdInstanceName: string, crdName: string, namespace: string, healthStatus: string) => {
-    waitUntilConfigIsVisible(3, crdInstanceName, crdName, namespace);
-    cy.get(`[data-test=VirtualItem_Ns${namespace}_${crdName}_${crdInstanceName}] span.pf-v5-c-icon`)
-      .should('be.visible')
-      .hasCssVar('color', `--pf-v5-global--${healthStatus}-color--100`);
+    waitUntilConfigIsVisible(3, crdInstanceName, crdName, namespace, healthStatus);
```
ferhoyos, line 660: What happens if the health status is incorrect? Does the function run indefinitely? I'm not sure if Cypress has a time limit per scenario. If not, I think we should add a `retries` parameter to the function (waiting until the config is visible for up to X seconds) and perform a final check.

hhovsepy: The function `waitUntilConfigIsVisible` has the retry parameter, which is set 3 here, this function was modified to not only check the config visibility but also the healthStatus icon of config. So it should fail after 3rd attempt (x10000ms).

business/istio_validations.go
hhovsepy: There is one new function `fetchAllWorkloadsDirect` which does the main job and is called sync only, it has the core of the original function `fetchAllWorkloads`. And `fetchAllWorkloads` now is a wrapper for async call for  `fetchAllWorkloadsDirect`.

frontend/cypress/integration/common/istio_config.ts
ferhoyos: You are right, I did not see the `attempt` input parameter. However, when `attempt === 0` the function returns and the test scenario is ok because there is no final validation

business/istio_validations.go
nrfox: Can we keep `fetchAllWorkloads` the same as it was and remove the `fetchAll...Direct` methods? If we need to keep `CreateValidations` sync, we don't need to call `fetchAllWorkloads` at all in `CreateValidations` and can just put that logic in directly like how it is here: https://github.com/kiali/kiali/pull/7136/files#diff-4afe91ed13da4c9e145fdccf0596bee0b6c0fc2e29d3da22f112584428159dd3R151-R164. Maybe there's another way of organizing things that also is clearer.

The two things that are hard for me to follow are 1. passing the return values _into_ the function as opposed to just returning them _from_ the function. This also means you need to pass pointers to maps/slices etc. which generally isn't needed and also if you change the order of the function calls then it will affect the return value args but you'll get no compiler error. 2. having multiple wrappers for functions. This makes it hard for me to follow where things are being set.

I actually think if you pull all the async bits out of `fetchAllWorkloads` then you can use it in both `CreateValidations` and `GetIstioObjectValidations` without needing the wrappers. Something like this:
```
go func() {
		defer wg.Done()

		if len(errChan) > 0 {
			return
		}
		var fetchErr error
		if istioConfigList, mtlsDetails, rbacDetails, fetchErr = in.fetchIstioConfigList(ctx, cluster, namespace); err != nil {
			errChan<-err
		} 
	}()
```
Then you can call `fetchIstioConfigList` the same way in `CreateValidations`

```
	for _, namespace := range namespaces {
		istioConfigs, mtlsDetails, rbacDetails, err := in.fetchIstioConfigList(ctx, cluster, namespace.Name)
		if err != nil {
			return nil, err
		}
        ...
```

frontend/cypress/integration/common/istio_config.ts
hhovsepy: Good catch, have added error when retries reach 0.

business/istio_validations.go
hhovsepy: I see now, thanks for the suggestion, have removed the wrapper methods and now only one method is left for direct calls. However the reference variables seems to be the part of sync calls logic in `GetIstioObjectValidations` methods.

ferhoyos:
The frontend tests are good to me. Waiting for @nrfox approval in the backend side.

nrfox (approved):
LGTM. We'll also want to merge in the design doc: https://github.com/kiali/kiali/pull/7142. I need to update that PR with the suggestions/comments that were left.

store/threadsafe_store.go
```diff
@@ -62,5 +66,9 @@ func (s *threadSafeStore[K, V]) Keys() []K {
 	return maps.Keys(s.data)
 }
 
+func (s *threadSafeStore[K, V]) Version() uint {
+	return s.version
+}
+
 // Interface guard to ensure threadSafeStore implements the Store.
-var _ Store[any, any] = &threadSafeStore[any, any]{}
+var _ Store[string, any] = &threadSafeStore[string, any]{}
```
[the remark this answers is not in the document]
hhovsepy: This comes from original PR
https://github.com/kiali/kiali/pull/7136/files#diff-6a38cf90314ad517bd2ee63d650266ffde68ff6a06c775492b465a0192a23429R62

Issue #7882, opened 2024-11-01
Make gateways optional in istio status


nrfox:
### What do you want to improve?

Kiali has an Istio health check in the masthead. As part of that health check, by default, Kiali checks if there's an ingress and egress gateway installed in the `istio-system` namespace. If you haven't installed a gateway or if you installed your gateways in a different namespace, Kiali will show an error in the masthead until you [manually configure](https://kiali.io/docs/configuration/kialis.kiali.io/#.spec.external_services.istio.component_status) Kiali to ignore these errors or look for the gateways in a different namespace.

### What is the new behavior?

By default, Kiali shouldn't show an error in the masthead if the gateways are not installed or are not installed in the `istio-system` namespace. Instead, Kiali should look for gateways across all accessible namespaces and only show an error if the gateway is unhealthy. If no gateways are found, this should not be an error.

jshaughn:
Adding to the sprint backlog, this has been questioned multiple times by users.

xSAVIKx:
Just wanted to add that it is also common to have more than one ingress in a separate namespace multiple ingresses in different namespaces

PR #7894, opened 2024-11-05
Autodetect gateways across namespaces

nrfox:
### Describe the change

If a user specifies `istio.component_status`, then that will be used exclusively for the istio status checks. If a user doesn't specify `istio.component_status`, then Kiali will look for gateway workloads across all namespaces. Since we look for gateway workloads across namespaces, this PR also removes the gateway related Kiali config options. It also moves the logic for determining if a workload is a gateway or not to the backend whereas before it was duplicated in both the frontend and the backend. This PR also changes ingress gateways to be "not core" in the istio status health checks. Now when you deploy Kiali without creating an ingress/egress gateway, it will no longer complain at you in the masthead.

#### Test gateway autodetection
1. Deploy Istio + Kiali
2. Create a gateway with gateway-injection:
    ```
    kubectl create namespace istio-ingress
    helm install istio-ingressgateway istio/gateway -n istio-ingress
    ```
3. Ensure Kiali doesn't have a "missing sidecar" badge for the gateway workload.
4. Create a gateway with gateway-api
    ```
    kubectl create namespace istio-ingress
    kubectl apply -f - <<EOF
    kind: Gateway
    metadata:
      name: gateway
      namespace: istio-ingress
    spec:
      gatewayClassName: istio
      listeners:
      - name: default
        hostname: "*.example.com"
        port: 80
        protocol: HTTP
        allowedRoutes:
          namespaces:
            from: All
    EOF
    ```
5. Ensure Kiali doesn't have a "missing sidecar" badge for the gateway workload.

#### Test specifying `component_status`
1. Deploy Istio + Kiali
2. Set istio component status to something wrong:
    ```
      istio:
        component_status
          components:
          - app_label: nonexistantgateway
            is_core: false
            is_proxy: true
            namespace: doesnotexist
    ```
3. Masthead should show a warning just for what you've set.

#### Test deploying Istio without gateways doesn't show warning

1. Deploy Sail + Kiali. Use this PR to easily deploy Kiali + sail: https://github.com/kiali/kiali/pull/7876. Then rebuild/reload kiali with this PR.
2. Ensure no warnings in masthead

### Automation testing

Included unit tests.

### Issue reference

#7882

### DIFF
--- a/business/istio_status.go
+++ b/business/istio_status.go
@@ -67,12 +67,18 @@ func (iss *IstioStatusService) GetStatus(ctx context.Context) (kubernetes.IstioC
 
 func (iss *IstioStatusService) getIstioComponentStatus(ctx context.Context, cluster string) (kubernetes.IstioComponentStatus, error) {
 	// Fetching workloads from component namespaces
-	workloads, err := iss.getComponentNamespacesWorkloads(ctx, cluster)
-	if err != nil {
-		return kubernetes.IstioComponentStatus{}, err
+	// If there's some explicit config then use that. Otherwise autodiscover.
+	if len(iss.conf.ExternalServices.Istio.ComponentStatuses.Components) > 0 {
+		log.Trace("Istio components config set. Using this instead of autodetecting components.")
+		workloads, err := iss.getComponentNamespacesWorkloads(ctx, cluster)
+		if err != nil {
+			return kubernetes.IstioComponentStatus{}, err
+		}
+
+		return iss.getStatusOf(workloads, cluster), nil
 	}
 
-	deploymentStatus := iss.getStatusOf(workloads, cluster)
+	log.Trace("Istio components config not set. Autodetecting components.")
 
 	mesh, err := iss.discovery.Mesh(ctx)
 	if err != nil {
@@ -100,46 +106,28 @@ func (iss *IstioStatusService) getIstioComponentStatus(ctx context.Context, clus
 			IsCore:    true,
 		})
 	}
-	deploymentStatus = deploymentStatus.Merge(istiodStatus)
-	deploymentStatus = deploymentStatus.Merge(iss.getGatewaysStatus(ctx, cluster))
 
-	return deploymentStatus, nil
-}
+	// Autodiscover gateways.
+	workloads, err := iss.workloads.GetAllWorkloads(ctx, cluster)
+	if err != nil {
+		// Don't error on gateways since they are non-essential.
+		log.Debugf("Unable to get gateway workloads when building istio component status. Cluster: %s. Err: %s", cluster, err)
+		return istiodStatus, nil
+	}
 
-func (iss *IstioStatusService) getGatewaysStatus(ctx context.Context, cluster string) kubernetes.IstioComponentStatus {
-	var gatewaysStatus kubernetes.IstioComponentStatus
-	updateGatewayStatus := func(label string, isCore bool, namespace string) {
-		// Fetching gateway workloads from gateway namespace
-		if namespace == "" {
-			namespace = iss.conf.IstioNamespace
-		}
-		// keep the label in status name, to align with ComponentStatuses
-		wls, err := iss.workloads.fetchWorkloadsFromCluster(ctx, cluster, namespace, label)
-		if err == nil && len(wls) != 0 {
-			for _, wl := range wls {
-				status := GetWorkloadStatus(*wl)
-				gatewaysStatus = append(gatewaysStatus, kubernetes.ComponentStatus{
-					Cluster:   cluster,
-					Name:      label,
-					Namespace: namespace,
-					Status:    status,
-					IsCore:    isCore,
-				})
-			}
-		} else {
-			gatewaysStatus = append(gatewaysStatus, kubernetes.ComponentStatus{
-				Cluster:   cluster,
-				Name:      label,
-				Namespace: namespace,
-				Status:    kubernetes.ComponentNotFound,
-				IsCore:    isCore,
+	for _, workload := range workloads {
+		if workload.IsGateway() {
+			istiodStatus = append(istiodStatus, kubernetes.ComponentStatus{
+				Cluster:   workload.Cluster,
+				Name:      workload.Name,
+				Namespace: workload.Namespace,
+				Status:    GetWorkloadStatus(*workload),
+				IsCore:    false,
 			})
 		}
 	}
-	updateGatewayStatus(iss.conf.IstioLabels.IngressGatewayLabel, true, iss.conf.ExternalServices.Istio.IngressGatewayNamespace)
-	updateGatewayStatus(iss.conf.IstioLabels.EgressGatewayLabel, false, iss.conf.ExternalServices.Istio.EgressGatewayNamespace)
 
-	return gatewaysStatus
+	return istiodStatus, nil
 }
 
 func (iss *IstioStatusService) getComponentNamespacesWorkloads(ctx context.Context, cluster string) ([]*models.Workload, error) {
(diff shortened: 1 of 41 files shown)


frontend/src/components/VirtualList/Config.ts
```diff
@@ -28,8 +29,17 @@ export const hasHealth = (r: RenderResource): r is SortResource => {
   return (r as SortResource).health !== undefined;
 };
 
-export const hasMissingSidecar = (r: SortResource): boolean => {
-  return !isIstioNamespace(r.namespace) && !r.istioSidecar && !isGateway(r.labels) && !isWaypoint(r.labels);
+export const hasMissingSidecar = (workload: Workload | WorkloadListItem | AppWorkload | AppListItem): boolean => {
+  return (
+    // TODO:
+    // Are we missing an ambient case here where ambient is enabled AND we are missing the ambient labels AND we are missing the sidecar?
+    // hasMissingSC && hasMissingA && serverConfig.ambientEnabled
```
josunect, line 36: But, in this case, workload.isAmbient should be false

josunect: Forgot about my last comment, in the case of Ambient, that check would be needed also: 
image

josunect (commented):
Going throw the testing procedure: 

**Test gateway autodetection**
No missing sidecar for the gateway:
image

In Ambient mesh, Gateway API shows missing sidecar (Out of mesh for ambient): 
image

**Test deploying Istio without gateways doesn't show warning**

No errors: 
image



**Test specifying component_status**:
is_core=false 
I can see an error instead of a warning, and the mesh page:
image

(The mesh page error happens also in master)

Just this:
image

Maybe it should be excluded by default in the sail hack script config?

frontend/package.json
```diff
@@ -1,4 +1,5 @@
 {
+  "proxy": "http://172.18.255.71/kiali",
```
josunect, line 2: I think that shouldn't be here :=)

business/istio_status.go
```diff
@@ -100,46 +106,28 @@ func (iss *IstioStatusService) getIstioComponentStatus(ctx context.Context, clus
 			IsCore:    true,
 		})
 	}
-	deploymentStatus = deploymentStatus.Merge(istiodStatus)
-	deploymentStatus = deploymentStatus.Merge(iss.getGatewaysStatus(ctx, cluster))
 
-	return deploymentStatus, nil
-}
+	// Autodiscover gateways.
+	workloads, err := iss.workloads.GetAllWorkloads(ctx, cluster)
```
josunect, line 111: I'm not sure is this could be more efficient but just asking: 
All the workloads are fetched just to see if they are gateways, if the gateways are identified by labels, it would be better to create a GetAllGateways method that fetches all the gateway labels?

models/workload.go
```diff
@@ -473,23 +485,49 @@ func (workload *Workload) HasIstioSidecar() bool {
...
+	if workload.TemplateAnnotations != nil && workload.TemplateAnnotations["inject.istio.io/templates"] == "gateway" {
+		return true
+	}
+
+	// gateway-api
+	// If this label exists then it's a gateway
+	if _, ok := workload.Labels["istio.io/gateway-name"]; ok {
+		return true
+	}
+
```
josunect, line 530: Should we include here some other labels? When I create the Gateway in Ambient mesh for the istio-ingress namespace I get the missing sidecar. It has another labels: 

image

Maybe the `gateway.istio.io/managed=istio.io-gateway-controller`?

nrfox: This looks like it changed in 1.24 according to the release notes:
> Removed the deprecated istio.io/gateway-name label, please use gateway.networking.k8s.io/gateway-name label instead.

nrfox: > gateway.istio.io/managed

Just using this might be enough. If we use `gateway.networking.k8s.io/gateway-name` I think we'll want to use `gateway.istio.io/managed` as well anyways because there could be other gateway-api implementations in the cluster that aren't istio and we don't want to get those workloads mixed up with the istio gateways.

nrfox: I actually wonder if we're doing this backwards. Shouldn't a `Gateway` be whatever workload is selected by either an Istio `Gateway` resource or a gateway-api `Gateway` resource? We can use that same selector to find the pods/deployments for the gateways rather than having to look at all these different labels.

frontend/src/components/VirtualList/Config.ts
nrfox: I think this has more to do with the new gateway label. Can you test again with the latest changes?

models/workload.go
nrfox: > I actually wonder if we're doing this backwards. Shouldn't a Gateway be whatever workload is selected by either an Istio Gateway resource or a gateway-api Gateway resource? We can use that same selector to find the pods/deployments for the gateways rather than having to look at all these different labels.

I ended up not going with this approach because it would require a lot of refactoring since every workload needs to know if it is a gateway or not due to how the missing sidecar checks work. I wonder if this check in general is somewhat backwards... maybe it should be a positive check like a green check mark if the workload is in the mesh and not a negative check like what it is today that a workload is out of the mesh.

jshaughn: I think I'm -1 on that approach.  I think users usually expect us to flag the outlier.  As an aside, aren't all gateway workloads "kind: gateway" ?

nrfox: This was just a shower thought. Not making any changes for this.

>  I think users usually expect us to flag the outlier.

Workloads without a sidecar are only an outlier in namespaces that are part of the mesh. Kiali shows you workloads across your entire cluster. More than likely some parts of cluster is not part of the mesh and you don't want them to be. A good example of this is the Sail Operator. When you deploy the Sail Operator into its own namespace, Kiali complains that it is not part of the mesh. But you don't want the Sail Operator to ever be part of your mesh so this shouldn't be an error. On the other hand, if you have a workload in a namespace that has injection enabled or is part of ambient and that workload isn't part of the mesh, then you definitely want an error/warning.

>  As an aside, aren't all gateway workloads "kind: gateway" ?

Kind of... there's two different `kind: Gateway`s. There's the classic Istio `kind: Gateway` and then there's the Gateway API `kind: Gateway`. When you create a classic Istio `kind: Gateway`, you still have to create the deployment for the Istio proxy yourself. In the Istio `kind: Gateway`, you just specify a selector to reference the gateway Istio proxy pods in the deployment. For Gateway API `kind: Gateway`, the Istio gateway proxy deployment gets created for you when you create the `kind: Gateway` object. From Kiali's perspective, we only care about the actual Istio proxy deployment in both cases since that is what we consider as the "workload" so we don't care about the `kind: Gateway` at all.

business/workloads.go
```diff
@@ -148,6 +148,33 @@ func (in *WorkloadService) getWorkloadValidations(authpolicies []*security_v1.Au
 	return validations
 }
 
+// GetAllWorkloads fetches all workloads across every namespace in the cluster.
+func (in *WorkloadService) GetAllWorkloads(ctx context.Context, cluster string) (models.Workloads, error) {
```
josunect, line 152: If this method is just used from `istio_status` to check for the gateways, maybe it could be renamed to "GetAllGateways" and return just the workloads which are gateways? Maybe is a small improvement but, in case there are lots of workloads, we could save a loop here.

josunect (dismissed):
Tested (Ambient/Sail) and all looks good. 

I've added a comment, I think it could be a small improvement, but approving anyway.

business/workloads.go
nrfox: There's at least one other place where we do the same thing as we are here https://github.com/kiali/kiali/blob/818cb89384dd46a240db8070a3e42a4e305bce50/handlers/workloads.go#L93-L116 so I thought this method could be used more widely but that being said I didn't actually update `handlers/workloads`.

josunect (approved):
LGMT!

PR #7895, opened 2024-11-06
Allow the user to set trace limits for logs and metrics pages

jshaughn:
Issue 7769 has to do with trace performance, and the main idea is that we should stop trying to do more than what the [tempo and jaeger] api/ui actually do. Meaning, don't try and perform multiple queries, under the covers, in order to fill out a time period. Instead let the user adjust time periods and/or trace limits. The main PR will adjust that logic, but for it to really work, the user needs to be able to adjust the query limit as needed. This PR aims to add that flexibility where needed.

Allow the user to set trace limits for logs and metric detail pages, as they already can in trace detail page.

Part-of https://github.com/kiali/kiali/issues/7769


frontend/src/components/Dropdown/ToolbarDropdown.tsx
```diff
@@ -90,7 +91,11 @@ export const ToolbarDropdown: React.FC<ToolbarDropdownProps> = (props: ToolbarDr
 
   return (
     <>
-      {props.nameDropdown && <span className={dropdownTitle}>{props.nameDropdown}</span>}
+      {props.nameDropdown && (
+        <span id={`${props.id}-name`} className={`${dropdownTitle} ${props.nameDropdownClassName}`}>
```
ferhoyos, line 95: When you have more than a className it is better to use the utility method `classes` from 'typestyle'. This will avoid `undefined` classNames.

frontend/src/components/Metrics/TraceLimit.tsx
```diff
@@ -0,0 +1,106 @@
...
+    100: '100',
+    500: '500',
+    1000: '1000'
+  };
+
+  const traceLimitComponent = (
+    <span id="trace-limit">
+      <div style={{ marginTop: '0.5rem' }}>
+        <span className={props.titleClassName} style={{ paddingRight: 0 }}>
+          {props.title ? props.title : t('Limit per query')}
```
ferhoyos, line 65: Shorter
```suggestion
          {props.title ?? t('Limit per query')}
```

frontend/src/components/Metrics/TraceLimit.tsx
```diff
@@ -0,0 +1,106 @@
...
+      })}
+    </span>
+  );
+
+  const traceLimitDropdownComponent = (
+    <span>
+      <ToolbarDropdown
+        id="trace-limit-dropdown"
+        handleSelect={onLimitChange}
+        nameDropdown={props.title ? props.title : t('limit')}
```
ferhoyos, line 95:
```suggestion
        nameDropdown={props.title ?? t('limit')}
```

frontend/src/pages/WorkloadDetails/WorkloadPodLogs.tsx
```diff
@@ -1279,7 +1293,8 @@ const formatDate = (timestamp: string): string => {
 const mapStateToProps = (state: KialiAppState): ReduxProps => {
   return {
     kiosk: state.globalState.kiosk,
-    timeRange: timeRangeSelector(state)
+    timeRange: timeRangeSelector(state),
+    tracingIntegration: state.tracingState.info ? state.tracingState.info.integration : false
```
ferhoyos, line 1297: Shorter. Not tested but this should work.
```suggestion
    tracingIntegration: state.tracingState.info?.integration ?? false
```

frontend/src/components/Metrics/TraceSpansLimit.tsx
```diff
@@ -0,0 +1,60 @@
...
+    currentLimit = limit;
+    props.onSpansChange(showSpans, currentLimit);
+  };
+
+  const onSpansChange = (_event, checked): void => {
+    setShowSpans(checked);
+    props.onSpansChange(checked, currentLimit);
+  };
+
+  const label = props.spansLabel ? props.spansLabel : t('spans');
```
ferhoyos, line 32:
```suggestion
  const label = props.spansLabel ?? t('spans');
```

ferhoyos (changes requested):
The PR works fine. When I change the limit size, I can see fewer or more traces depending on the value. I'm sure that @josunect can better judge if the limit on tracing is working as expected.

It seems odd to me that the `spans` label is replaced by `limit`, as it loses focus on what is actually being limited. I think the dropdown itself indicates the number of spans/traces to be fetched without needing the `limit` label. Additionally, there's an info icon to describe the dropdown.

image

josunect:
> The PR works fine. When I change the limit size, I can see fewer or more traces depending on the value. I'm sure that @josunect can better judge if the limit on tracing is working as expected.
> 
> It seems odd to me that the `spans` label is replaced by `limit`, as it loses focus on what is actually being limited. I think the dropdown itself indicates the number of spans/traces to be fetched without needing the `limit` label. Additionally, there's an info icon to describe the dropdown.
> 
> image

Yes, the limit looks good, but the limit applies to the number of traces, and we are showing spans. That might be confusing?

ferhoyos:
> Yes, the limit looks good, but the limit applies to the number of traces, and we are showing spans. That might be confusing?

You could add the word `traces` in the dropdown to clarify that the limit applies to the number of traces, not spans. This is similar to what we do with log lines:

image

frontend/src/components/Dropdown/ToolbarDropdown.tsx
jshaughn: gracias, I didn't know about this.  Will change.

frontend/src/components/Metrics/CustomMetrics.tsx
```diff
@@ -135,17 +140,19 @@ class CustomMetricsComponent extends React.Component<Props, MetricsState> {
...
       this.props.workload !== prevProps.workload ||
       this.props.version !== prevProps.version ||
       this.props.template !== prevProps.template ||
       this.props.lastRefreshAt !== prevProps.lastRefreshAt ||
+      this.state.showSpans !== prevState.showSpans ||
+      this.state.showSpansLimit !== prevState.showSpansLimit ||
       !isEqualTimeRange(this.props.timeRange, prevProps.timeRange)
     ) {
-      const settings = MetricsHelper.retrieveMetricsSettings();
+      const settings = MetricsHelper.retrieveMetricsSettings(spansLimitDefault);
```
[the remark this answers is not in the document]
jshaughn: Hmm. I think you are right.  I will change and test.

jshaughn:
> Yes, the limit looks good, but the limit applies to the number of traces, and we are showing spans. That might be confusing?

The query limit is only on Traces, right?  You are correct, we are showing spans in the metric charts, and embedded in the logs, because we are focused only on the spans involving the detail page's app/service/workload. It is a little confusing, but I'm not sure what to do about it.  One possibility is to change the `spans` checkbox to just be `traces`, and stop using the spans terminology. We could still just show the relevant trace spans in the metrics or logs.  Thoughts?

jshaughn:
@josunect , @ferhoyos , I have made some changes based on your feedback, let me know if you like it better.

josunect:
> > Yes, the limit looks good, but the limit applies to the number of traces, and we are showing spans. That might be confusing?
> 
> The query limit is only on Traces, right? You are correct, we are showing spans in the metric charts, and embedded in the logs, because we are focused only on the spans involving the detail page's app/service/workload. It is a little confusing. How about if I change the `spans` checkbox to just be `traces`, and stop using the spans terminology. We could still just show the relevant trace spans in the metrics or logs.

Right, the limit is just for spans. 

I think it would be less confusing letting the "spans" in the checkbox title, because it is showing spans, and not traces, and don't hide it when the new select is shown - because it will show a different data. But I think it represents better what it is showing - spans, limited by the number of traces fetched.

jshaughn:
> I think it would be less confusing letting the "spans" in the checkbox title, because it is showing spans, and not traces, and don't hide it when the new select is shown

OK, I changed the checkbox label back to "Spans" because you are correct, we are really showing spans in the logs and metric charts.  But, I am still hiding the label when the box is checked.  I think it looks too "busy" to show both the checkbox label and the limit dropdown. To me I think it's clear because the dropdown entries are now more verbose (e.g. "20 traces"), there is the info tooltip, and I did add a line to the tooltip to mention spans.

Let's also wait for @ferhoyos to comment.

josunect:
> I'm still not sure that converting a label next to a checkbox into a dropdown is the right approach from a UX perspective. I think we'll need the expert opinion of @fkargbo here.
> 
> Two other comments:
> 
>     * ~I think we need more consistency with the info tooltips on this page. The info tooltip for trace limit has its own info icon, while the info tooltip for lines limit on the right side appears when hovering over the dropdown. Either approach is fine, but both should use the same style for consistency.~ After thinking it over, maybe for longer info descriptions we have to use the info icon, and for shorter ones, the hover approach...
> 
>     * The trace limit parameter is currently defined in the URL as `limit`. The issue is that we also have the limit lines dropdown on the right side, which could confuse users. I suggest renaming it to `traceLimit` or something similar to make it clearer.

I think it would be better to rename the lines dropdown limit, as this (I think) is only used here. The traces limit is used in other different pages.

ferhoyos:
> I think it would be better to rename the lines dropdown limit, as this (I think) is only used here. The traces limit is used in other different pages.

The lines limit does not have a URL parameter, so there’s no issue in that regard. If the trace limit is used on other pages and would require significant effort to rename, we can leave it as-is.

jshaughn:
> I'm still not sure that converting a label next to a checkbox into a dropdown is the right approach from a UX perspective

I think I misinterpreted your original comment: "...I think the dropdown itself indicates the number of spans/traces to be fetched without needing the limit label".  I took this to mean that we should keep the dropdown but eliminate the label, giving what we see today.  But I think you meant, keep "Spans" but don't label the dropdown.  Josune also commented that we should not remove the "Spans" label.  I will change it.

>  I suggest renaming it to traceLimit

We have a lot of url param names I don't like, but I prefer not to rename them because it invalidates bookmarked urls.

ferhoyos:
> But I think you meant, keep "Spans" but don't label the dropdown. Josune also commented that we should not remove the "Spans" label. I will change it.

Yes, that's exactly what I meant. Keep the "spans" label and just add the dropdown to the right when the checkbox is checked.

> We have a lot of url param names I don't like, but I prefer not to rename them because it invalidates bookmarked urls.

At first, I thought this URL parameter was new. When @josunect mentioned that this parameter already existed and affected other pages, I understood that it’s better not to rename it.

jshaughn:
I updated the PR once again, and now it keeps the "Spans" label.  Sorry for not understanding the initial review feedback.  This also adds some minor toolbar tweaking to better alignment. A follow up PR will be created to further apply consistent margin styling for info tooltips.

PR #7908, opened 2024-11-08
[CI] Fix cypress test failures with Istio 1.24

josunect:
### Describe the change

**Ambient tests**
Remove the check looking for the label "istio.io/gateway-name=waypoint" as it was removed in istio 1.24.
This label is not been used in other checks internally, only this change is required.

**Frontend tests**
The image for the bookinfo deployment has been updated. The logs had changed and the just log INFO messages. The "Last 30m" was updated in the logs to be able to see the messages (They are no so frequent as the DEBUG), and also filter by INFO text. (Trying to patch the deployment changing the log level but it doesn't have any effect). 

**Multi primary tests**
The failures are because the tests are trying to create a remote istio config and they expect failure, but now the Istio CRDs are created in the remote cluster so it successful creates the istio config in the remote. 
Talk to @nrfox and this tests are not needed.

### Steps to test the PR

CI passes.
Also, install istio ambient and a waypoint proxy, and check the labels. No "istio.io/gateway-name=waypoint" should be there.

### Automation testing

Update cypress teste

### Issue reference

If this PR is related to a github issue, please link it here ([more info](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue))

### DIFF
--- a/frontend/cypress/integration/common/workload_logs.ts
+++ b/frontend/cypress/integration/common/workload_logs.ts
@@ -8,6 +8,8 @@ Given(
   'I am on the logs tab of the {string} workload detail page of the {string} namespace',
   (workload: string, namespace: string) => {
     cy.visit({ url: `/console/namespaces/${namespace}/workloads/${workload}?tab=logs&refresh=0` });
+    cy.get('#metrics_filter_interval_duration-toggle').click();
+    cy.get('#1800').click();
   }
 );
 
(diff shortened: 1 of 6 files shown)


nrfox (approved):

ferhoyos:
@josunect I think this PR needs to be backported to v1.89, PR https://github.com/kiali/kiali/pull/7911 is failing

PR #7914, opened 2024-11-11
get release to build with updated buildkit and binfmt

jmazzitelli:
This is to fix the multi-arch build error:

Fatal glibc error: CPU lacks float128 support (POWER 9 or later required)


jmazzitelli:
this seems to be the same issue we had before in OSSMC:

- https://github.com/kiali/openshift-servicemesh-plugin/issues/312
- https://github.com/kiali/openshift-servicemesh-plugin/pull/316

Issue #7820, opened 2024-10-15
Alternative Workload controllers support


apolovov:
### What do you want to improve?

I want to use alternative workload controllers like [kruise](https://openkruise.io/docs/user-manuals/advanceddaemonset/) for handling workloads.

### What is the current behavior?
We use [kruise](https://openkruise.io/docs/user-manuals/advanceddaemonset/) for handling DaemonSets, but Kiali [uses the standard k8s api to fetch the original DS](https://github.com/kiali/kiali/blob/654d836513d220452de7ad3fa8d272887fb3382b/business/workloads.go#L868) during workload discovery.

### What is the new behavior?

Kiali could discover the api version from Pod's ownerReference to use it for original controller discovery.

```
  ownerReferences:
  - apiVersion: apps.kruise.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: DaemonSet
    name: controller-main
    uid: bc83975f-28e0-4dae-b588-1d08b7954861
```

jshaughn:
Hi @apolovov , I'm a little surprised your DaemonSets are not found, given that the kind is still daemonset.  Although this is tricky code and I'm not very familiar.

apolovov:
The kubeCache libe [uses](https://github.com/kiali/kiali/blob/b3c4323708930d7c55e83ca8632721315f50d5d3/kubernetes/cache/kube_cache.go#L583) the `apps/v1` Lister to fetch DaemonSets. But our DaemonSets has `apps.kruise.io/v1alpha1` apiVersion. So, kiali can't fetch ours with error `Workload xxx is not found as DaemonSet`. I agree that the code is tricky and the redesign will require serious reworking. But the kruise controller is quite popular (and i'm sure there are some other implementations for DS/STS/etc)...

nrfox:
> Kiali could discover the api version from Pod's ownerReference to use it for original controller discovery.

It's a good idea. I have had a similar thought that this would simplify the workload discovery and make it more flexible. The workload discovery kind of does this today by gathering all the pods, looking at their owner refs, and associating them with that "controller" but it compares the owner refs to a predefined list of types (`Daemonset`, `Deployment`, etc.) and has special handling for each specific type. Kiali also caches all objects for these predefined types and it probably needs to because this workload discovery gets called a lot.

To support this, I think we'd need to:

[...]

PR #7918, opened 2024-11-12
Workloads GVK support + Custom Workloads

hhovsepy:
### Describe the change

Workloads were tight to their `Kind`, which was making impossible to read those ones which are Custom and has the same Kind but different 'apiVersion'.
Now Workloads support GroupVersionKind  instead of 'type' field.

### Steps to test the PR
Install Kruise, wait until it is up.
`helm install kruise openkruise/kruise --version 1.7.2`
Install this custom DaemonSet.
```
apiVersion: apps.kruise.io/v1alpha1
kind: DaemonSet
metadata:
  name: fluentd-elasticsearch
  labels:
    k8s-app: fluentd-logging
spec:
  selector:
    matchLabels:
      name: fluentd-elasticsearch
  updateStrategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: 1
  template:
    metadata:
      labels:
        name: fluentd-elasticsearch
    spec:
      tolerations:
      # these tolerations are to have the daemonset runnable on control plane nodes
      # remove them if your control plane nodes should not run pods
      - key: node-role.kubernetes.io/control-plane
        operator: Exists
        effect: NoSchedule
      - key: node-role.kubernetes.io/master
        operator: Exists
        effect: NoSchedule
      containers:
      - name: fluentd-elasticsearch
        image: quay.io/fluentd_elasticsearch/fluentd:v2.5.2
        resources:
          limits:
            memory: 200Mi
          requests:
            cpu: 100m
            memory: 200Mi
        volumeMounts:
        - name: varlog
          mountPath: /var/log
        - name: varlibdockercontainers
          mountPath: /var/lib/docker/containers
          readOnly: true
      terminationGracePeriodSeconds: 30
      volumes:
      - name: varlog
        hostPath:
          path: /var/log
      - name: varlibdockercontainers
        hostPath:
          path: /var/lib/docker/containers
```

List Workloads, the newly added DaemonSet should be listed.
Click on it.
Details page comes with disabled Action:
Screenshot from 2024-11-15 11-33-32

### Automation testing

Existing tests modified to support GVK.

### Issue reference
https://github.com/kiali/kiali/issues/7820


frontend/src/components/IstioWizards/WorkloadWizardDropdown.tsx
```diff
@@ -144,6 +147,13 @@ export const WorkloadWizardDropdown: React.FC<Props> = (props: Props) => {
             t('User does not have permission on this Workload'),
             dropdown
           )
+        : !supportedWorkload
```
ferhoyos, line 150: In the dropdown you have to add `supportedWorkload` as disabled option in L111:

```
isDisabled={!validActions || !supportedWorkload}
```

frontend/src/pages/WorkloadDetails/WorkloadDescription.tsx
```diff
@@ -98,7 +103,7 @@ export const WorkloadDescription: React.FC<WorkloadDescriptionProps> = (props: W
 
           <li>
             <span>Type</span>
-            {workload.type ? workload.type : 'N/A'}
+            {workload.gvk.Kind ? workload.gvk.Kind : 'N/A'}
```
ferhoyos, line 106:
```suggestion
            {workload.gvk.Kind || 'N/A'}
```

ferhoyos (changes requested):
Another big GVK refactorization @hhovsepy! And this one brings support for custom workloads :slightly_smiling_face: 

Overall the PR is good, I have some minor comments (the enum one is optional and up to you to implement it). 

Related to custom workloads, except for the disabled dropdown I would not know that this kind of workload is not supported. The type in the info icon just says `DaemonSet`, maybe we should add also the apiVersion.

image

What about the rest of the info (graph, rest of tabs)? Are still valid for not supported workloads? In case the answer is no, I think we should hide invalid tabs and show proper message in the graph card:

```

**Comment by ferhoyos** [comment]
Thanks @hhovsepy for the quick fixes. In the Type tooltip I would set only the GVK, no comment about it is read-only. 

What about an info message similar to the one we have in traces? The user will know at one sight why this workload has not all the information available
image

The message coould be something like: `This type of workload is not fully supported by Kiali, only limited information is available for display`

**Review by ferhoyos — CHANGES_REQUESTED** [review]
@hhovsepy I like your approach of adding an alert message.

Last suggestion is to split the kind and group version when the gvk is not supported. The statement is too long and it does not fit well in the tooltip:

image

Try the following code and `resourceListStyle` CSS class modification:
```
const resourceListStyle = kialiStyle({
  display: 'flex',
  $nest: {
    '& > ul > li': {
      display: 'flex',
      $nest: {
        '& span': {
          minWidth: '125px',
          fontWeight: 700
        }
      }
    }
  }
});
```

```
      {!isGVKSupported(workload.gvk) && (
            <li>
              <span>API Version</span>
              {`${workload.gvk.Group}.${workload.gvk.Version}`}
            </li>
          )}

          <li>
            <span>Type</span>
            {workload.gvk.Kind || 'N/A'}
          </li>
```
This is the resulting tooltip with the modifications above:

image

**Comment by hhovsepy** [comment]
thank you  @ferhoyos , much better now

**Review by ferhoyos — DISMISSED** [review]
With the latest changes it looks good to me. Nice job @hhovsepy!

File: business/workloads.go [hunk 3]
```diff
@@ -918,36 +920,41 @@ func (in *WorkloadService) fetchWorkloadsFromCluster(ctx context.Context, cluste
...
+					continue
+				}
 				if ref.Controller != nil && *ref.Controller && in.isWorkloadIncluded(ref.Kind) {
 					if _, exist := controllers[ref.Name]; !exist {
-						controllers[ref.Name] = ref.Kind
+						controllers[ref.Name] = refGV.WithKind(ref.Kind)
 					} else {
-						if controllers[ref.Name] != ref.Kind {
-							controllers[ref.Name] = controllerPriority(controllers[ref.Name], ref.Kind)
+						if controllers[ref.Name].String() != refGV.WithKind(ref.Kind).String() {
```
nrfox, line 939: You may not need to convert to `.String()` before comparison here and below.

business/workloads_test.go
nrfox, line 1: Can you include this test: https://github.com/kiali/kiali/pull/7930 with this change?

kubernetes/filters.go
```diff
@@ -217,12 +219,16 @@ func FilterK8sGatewaysByRoutes(allGws []*k8s_networking_v1.Gateway, httpRoutes [
...
 	for _, pod := range allPods {
 		for _, ref := range pod.OwnerReferences {
-			// TODO: Kind is not a definitive reference. Need to include check for API version as well.
-			if ref.Controller != nil && *ref.Controller && ref.Name == controllerName && ref.Kind == controllerType {
+			refGV, err := schema.ParseGroupVersion(ref.APIVersion)
+			if err != nil {
+				log.Errorf("could not parse OwnerReference api version %q: %v", ref.APIVersion, err)
+				continue
+			}
+			if ref.Controller != nil && *ref.Controller && ref.Name == controllerName && refGV.WithKind(ref.Kind).String() == controllerGVK.String() {
```
nrfox, line 231: Same comment here... you may not need the conversion to `String()` before doing a comparison.

hhovsepy: right

business/workloads_test.go
hhovsepy: added

hhovsepy:
@nrfox thanks for the review, have approached all suggestions, removed odd GVK.String() in comparisons, added your test method.

ferhoyos (approved):
LGTM

PR #7923, opened 2024-11-13
Allow setting trace limit in graph node-side-panel Trace tab

jshaughn:
For completeness, we should also allow users the ability to set the trace fetch limit in the graph's node-side-panel Traces tab.

As part of this, remove the "use graph refresh" option and just have users manually refresh the traces list, as needed. This simplifies the tab header, given the new dropdown, and I doubt anyone will miss this option. If they do, we can always bring it back.

part-of kiali7769

Test
@josunect , I overlooked this in the original PR, but I think we need the option here as well. You can look at the network requests to ensure the initial request has limit 20, and that new requests are made if the limit changes.

### DIFF
--- a/frontend/src/components/Metrics/TraceLimit.tsx
+++ b/frontend/src/components/Metrics/TraceLimit.tsx
@@ -5,8 +5,6 @@ import { itemStyleWithoutInfo } from 'styles/DropdownStyles';
 import { ToolbarDropdown } from 'components/Dropdown/ToolbarDropdown';
 import { infoStyle } from 'styles/InfoStyle';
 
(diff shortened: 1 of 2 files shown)


josunect:
I'm wondering if the user would need to increase the trace limit in this window? I see this trace table as a trace preview, because there is not a lot of space to show information. Also, it is not evident if the traces only cover a small portion of what the user has requested, because this is shown in a table and not in a chart. There is also a button "Show traces" to redirect the user to the traces tab for the selected app for a broader view.

josunect (approved):
LG! 

image

PR #7924, opened 2024-11-14
Optimization loop logic with serviceentry-hostname

### LINKED ISSUE #7917: wokeload page became slow when the vs continued growth
### Describe the bug

When there are many service entry resources in the cluster and a large number of virtual services, the display of the workload and istioconfig pages in kiali will be very slow，More than 30 seconds，Finally, we found that the problem was that the ServiceEntryHostnames function had unreasonable for range

img_v3_02g3_31b762b4-9601-463e-b7cb-a5a1e6609dbg
img_v3_02g3_f047eaf4-c5a0-4ffc-ae82-04b79242258g

### Expected Behavior

/kiali/api/namespaces/ns/workloads/xxx
it should be return in seconds

we have 1400 se and 800 vs

### Environment

- **Kiali version:**
v 1.80
- **Istio version:**
v 1.65
- **Kubernetes impl:** 
- **Kubernetes version:** 
v 1.24

wth12138:
Fixes #7917

### DIFF
--- a/kubernetes/istio.go
+++ b/kubernetes/istio.go
@@ -146,17 +146,17 @@ func (in *K8SClient) SetProxyLogLevel(namespace, pod, level string) error {
 func ServiceEntryHostnames(serviceEntries []*networking_v1.ServiceEntry) map[string][]string {
 	hostnames := make(map[string][]string)
 
-	for _, v := range serviceEntries {
-		for _, host := range v.Spec.Hosts {
-			hostnames[host] = make([]string, 0, 1)
-		}
-		for _, port := range v.Spec.Ports {
-			protocol := mapPortToVirtualServiceProtocol(port.Protocol)
-			for host := range hostnames {
-				hostnames[host] = append(hostnames[host], protocol)
+	for _, serviceEntry := range serviceEntries {
+		for _, host := range serviceEntry.Spec.Hosts {
+			protocols := make([]string, 0, len(serviceEntry.Spec.Ports))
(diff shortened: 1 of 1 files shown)


kubernetes/istio.go
```diff
@@ -146,17 +146,17 @@ func (in *K8SClient) SetProxyLogLevel(namespace, pod, level string) error {
...
-			for host := range hostnames {
-				hostnames[host] = append(hostnames[host], protocol)
+	for _, serviceEntry := range serviceEntries {
+		for _, host := range serviceEntry.Spec.Hosts {
+			protocols := make([]string, 0, len(serviceEntry.Spec.Ports))
+			for _, port := range serviceEntry.Spec.Ports {
+				protocol := mapPortToVirtualServiceProtocol(port.Protocol)
+				protocols = append(protocols, protocol)
 			}
+			hostnames[host] = protocols
```
hhovsepy, line 156:
```suggestion
			hostnames[host] = append(protocols, hostnames[host]...)
```

hhovsepy: Thanks for the PR, it seems more optimal, just suggested one change to keep the previous protocols from other hosts and SEs rather than overriding it with the last SE in a loop.

hhovsepy (approved):
Looks good, tested locally and on OCP cluster.

wth12138:
> @wth12138 Thank you for this contribution to Kiali!

:laughing:

PR #7927, opened 2024-11-15
OSSM-8304: fixing ocp prometheus hack script for multiple kiali CRs

FilipB:
It's now possible to specify a namespace where the script will look for the Kiali CR.

### DIFF
--- a/hack/use-openshift-prometheus.sh
+++ b/hack/use-openshift-prometheus.sh
@@ -12,6 +12,7 @@ set -ue
 
 DELETE="false"
 ISTIO_NAMESPACE="istio-system"
+KIALI_CR_NAMESPACE=""
 MESH_LABEL="mymesh"
 NAMESPACES=""
 NETWORK_POLICIES="false"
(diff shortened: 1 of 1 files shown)


hack/use-openshift-prometheus.sh
```diff
@@ -96,35 +100,38 @@ RESOURCE_LABEL_VALUE="kiali"
...
-  else
-    KIALI_SA_NAMESPACE="$(echo $sa | cut -d: -f1)"
-    KIALI_SA_NAME="$(echo $sa | cut -d: -f2)"
-  fi
-done
+echo "Using following Kiali CR [${KIALI_CR_NAMESPACE}:${KIALI_CR_NAME}]"
 
+# Get a Kiali server namespace, if not defined in .spec.deployment.namespace, use Kiali CR's namespace
+KIALI_SA_NAMESPACE=$(${OC} get kiali ${KIALI_CR_NAME} -n ${KIALI_CR_NAMESPACE} -o jsonpath='{.spec.deployment.namespace}{"\n"}{.metadata.namespace}' | awk 'NF{print $1; exit}')
+# Get a service account requied to generate Kiali Cluster Role Binding for OpenShift Prometheus access
```
jmazzitelli, line 131:
```suggestion
# Get a service account required to generate Kiali Cluster Role Binding for OpenShift Prometheus access
```

hack/use-openshift-prometheus.sh
```diff
@@ -96,35 +100,38 @@ RESOURCE_LABEL_VALUE="kiali"
...
+KIALI_SA_NAME=$(${OC} get sa ${KIALI_CR_NAME}-service-account -n ${KIALI_SA_NAMESPACE} -o custom-columns=N:.metadata.name --no-headers)
```
jmazzitelli, line 132: The Kiali CR name has nothing to do with the names of the resources. It is determined from the Kiali CR's `.spec.deployment.instance_name` - if that is not defined, the default is `kiali`.

jmazzitelli (changes requested):

hack/use-openshift-prometheus.sh
jmazzitelli: That just returns the string of the expected name... that line doesn't actually attempt to see if it exists. I think that is something you are also trying to do here, so that's something you would need to add back in.

jmazzitelli (approved):
LGTM

PR #7938, opened 2024-11-19
Use the same kind node image criteria for all hack scripts

ferhoyos:
### Describe the Change

Currently, the Kind node image is defined in multiple hack scripts according to different criteria (sometimes it is hardcoded, sometimes the latest version is used, and sometimes it is not). This PR centralizes the criteria for determining the Kind node image to be used based on the Istio version. It also ensures that the latest Kind node image is used for the latest Istio version (e.g., multicluster scripts were installing kind 1.27 which is not supported by the latest Istio version 1.24.0)

### Steps to Test the PR

- Verify that the hack scripts use the latest Kind version with the `master` branch.
- Check that specific Kind versions are used for previous Istio versions.

### Automation Testing

N/A

### DIFF
--- a/hack/istio/multicluster/env.sh
+++ b/hack/istio/multicluster/env.sh
@@ -18,6 +18,8 @@ if [ "${HACK_ENV_DONE:-}" == "true" ]; then
   return 0
 fi
 
+set -u
+
 SCRIPT_DIR="$(cd $(dirname "${BASH_SOURCE[0]}") && pwd)"
 
 switch_cluster() {
@@ -130,6 +132,9 @@ MANAGE_MINIKUBE="${MANAGE_MINIKUBE:-true}"
 # If true and client exe is kubectl, then two kind instances will be installed/uninstalled by these scripts
 MANAGE_KIND="${MANAGE_KIND:-false}"
 
+# Image of the kind cluster
+KIND_NODE_IMAGE="${KIND_NODE_IMAGE:-}"
+
 # Minikube options - these are ignored if MANAGE_MINIKUBE is false
 MINIKUBE_DRIVER="kvm2"
 MINIKUBE_CPU=""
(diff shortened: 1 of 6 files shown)


hack/istio/multicluster/env.sh
```diff
@@ -18,8 +18,6 @@ if [ "${HACK_ENV_DONE:-}" == "true" ]; then
   return 0
 fi
 
-set -u
```
jmazzitelli, line 21: i don't think we want to get rid of this. This is a good validation feature because it helps catch instances where the code is using a variable that was never defined (or it catches a typo in a variable name). This has helped me catch many bugs that I otherwise would have missed.

If you expect a variable may not be defined in a particular place in the code, then simply use the syntax `${VAR:-}`.

hack/setup-kind-in-ci.sh
```diff
@@ -373,6 +369,10 @@ setup_kind_multicluster() {
     mkdir -p "${certs_dir}"/keycloak
   fi
 
+  if [[ -n "${KIND_NODE_IMAGE}" ]]; then
```
jmazzitelli, line 372: not sure why the double-brackets are for... `-n` test need only single bracket
```suggestion
  if [ -n "${KIND_NODE_IMAGE}" ]; then
```

jmazzitelli (changes requested):
One change is probably trivial (the double-bracket thing), but I think we should keep `set -u` because without it, it too easily allows for bugs to creep into these scripts.

hack/istio/multicluster/env.sh
ferhoyos: Thanks @jmazzitelli, I did not know the syntax  `${VAR:-}` for unbound variables. I have restored the `set -u` flag again.

jmazzitelli:
We should not merge this until all CI tests are green (because this PR affects the tests themselves), and there is a failure here in the integration tests.

ferhoyos:
Agreed. The cypress test issue is happening in all PRs, so we need to fix and rebase before approval.

jmazzitelli:
The CI bug is fixed - rebase this PR and let's see if this all goes green

jmazzitelli (approved):
lgtm - all tests pass

PR #7943, opened 2024-11-19
PFT traffic graph animation

### LINKED ISSUE #7934: support trafficPoint animation in PFT graph impl
In Kiali v2 we defer to the built-in PFT animation, which uses animated dashed edges of various speeds. This approach is clean easy to understand, and seems less compute intensive than our prior approach. But it's not as rich, and multiple users have said they prefer our previous "particle" (aka "point") animation, used with cytoscape.

This issue is for offering the old animation as an option in V2 (possibly the default).  I feel like it is doable but I wasn't able to get things working in an initial attempt. I'm not very savvy with thins canvas/context stuff.  Here are some initial findings to possibly help a future try.

The prior animation really doesn't have too much to do with cytoscape itself.  The dots, diamonds, speeds, and density are all created and calculated in kiali code. The animation is rendered independently from the graph, basically rendered on top of the edges, on a separate canvas.  What made this doable was a fairly small cytoscape extension called [cytoscape-canvas](https://github.com/classcraft/cytoscape.js-canvas/blob/master/src/cytoscape-canvas.js), which provided a child-canvas, from which you can get a '2d' context, and render everything directly.

The kiali animation code is found in `TrafficRenderer` and `TrafficPointRenderer`.  In PFT we basically took the code that generates the speeds, and used it to set animation speeds on the PFT edges.  You can see this in `GraphPF`, look for the `Effect` activated by `showTrafficAnimation`.

The question is how to adapt this to PFT. To do so we probably need to again access a 2d canvas we can use to do our rendering. I'm not sure how to do it. If we could get that handle, I'm then not sure whether we would want to render it completely independently, like we do now, or per-edge, possibly adding logic to `StyleEdge`.

PFT has the idea of `Layers`. And you can add custom layers. In an experiment, I created a new ANIMATION_LAYER called "animation", that attempts to be a new layer to which we could write over the graph. In this draft PR you can see a custom layer created here:

https://github.com/kiali/kiali/pull/7933/files#diff-6015c1275b7b3433002b4bd677c35a321c9ff75fd9fa95566b9462551d5b4cfeR368

I can use this layer to successfully write over the graph, you can see in a test here that I'm able to cover the edge in another edge that applies an edge tag of 'foo'.

https://github.com/kiali/kiali/pull/7933/files#diff-084c0ab7a7e65af2ce6d6cddb9bd653c7a340917bfd630cf8833648364527636R169-R192

Also here you can see an attempt to just generate a green rectangle using a 2d context, but it doesn't seem to do anything. I'm stopping, as I've hit a dead end. I'm not sure we need a new Layer, maybe we can somehow just get PFT's top-level container, like cytoscape offers with `cy.container()`, and then just adapt `cytoscape-canvas` (it has open licensing).

ferhoyos:
### Describe the change

Bring back the particle animation to the PFT traffic graph.

### Steps to test the PR

Check the Traffic Animation option in the `Display` dropdown of the traffic graph and verify that traffic is represented by circles moving along the edges.

### Automation testing

N/A

### Issue reference

Closes #7934

### DIFF
--- a/frontend/src/pages/GraphPF/styles/styleEdge.tsx
+++ b/frontend/src/pages/GraphPF/styles/styleEdge.tsx
@@ -152,15 +152,28 @@ const StyleEdgeComponent: React.FC<StyleEdgeProps> = ({ element, ...rest }) => {
   const moveY = endPoint.y - startPoint.y;
 
   const moveAnimation = keyframes({
-    from: { transform: 'translateX(0)' },
-    to: { transform: `translateX(${moveX}px) translateY(${moveY}px)` }
+    from: { translate: '0' },
+    to: { translate: `${moveX}px ${moveY}px` }
   });
 
   const circleStyle = kialiStyle({
     fill: 'white',
     stroke: 'blue',
     animationName: moveAnimation,
-    animationDuration: '1s',
+    animationDuration: '2s',
+    animationFillMode: 'forwards',
+    animationTimingFunction: 'linear',
+    animationIterationCount: 'infinite'
+  });
+
+  const rectStyle = kialiStyle({
+    fill: 'white',
+    stroke: 'red',
+    transformBox: 'fill-box',
+    transformOrigin: 'center',
+    rotate: '45deg',
+    animationName: moveAnimation,
+    animationDuration: '2s',
     animationFillMode: 'forwards',
     animationTimingFunction: 'linear',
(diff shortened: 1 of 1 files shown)


jshaughn:
Closing this in favor of https://github.com/kiali/kiali/pull/7955, which applies the approach shown here into a full PR.

PR #7944, opened 2024-11-19
[1.89] Use system pool certs when talking to external OAuth server

nrfox:
### Describe the change

Adds the system pool certs to the http client that is used to do the oauth token exchange.

### Steps to test the PR

Regression test for multi-cluster Kiali with `openshift` auth strategy: https://github.com/openshift-service-mesh/sail-operator/blob/e4100417b8dbd5a57847769d97677fa7048d27b4/docs/ossm/kiali-multi-cluster/README.md


jmazzitelli (approved):
This is still in Draft mode. Since it is a backport to a PR I already approved, I'll approve this as well.

nrfox:
This actually doesn't need to be backported. Closing...

PR #7945, opened 2024-11-19
Use system pool certs when talking to external OAuth server

nrfox:
### Describe the change

Adds the system pool certs to the http client that is used to do the oauth token exchange.

### Steps to test the PR

Regression test for multi-cluster Kiali with `openshift` auth strategy: https://github.com/openshift-service-mesh/sail-operator/blob/e4100417b8dbd5a57847769d97677fa7048d27b4/docs/ossm/kiali-multi-cluster/README.md


jmazzitelli:
Is this not a thing for openid? (i.e. should we use the system certs for openid auth, too? This PR only tweeks the openshift auth).

nrfox:
> Is this not a thing for openid? (i.e. should we use the system certs for openid auth, too? This PR only tweeks the openshift auth).

It works a little different for openid. If you are using a custom CA we allow you to mount it at `/kiali-cabundle/openid-server-ca.crt` and Kiali will use that.
https://github.com/kiali/kiali/blob/e6e8e3f43b7abf9fcc61f46989964ca45448a13c/handlers/authentication/openid_auth_controller.go#L933-L946

If you don't mount a custom CA, then `certPool` is `nil` and the default behavior for the `http.Client` in golang when the `certPool` is `nil` is to read the system certs. So we're already implicitly doing this for openid.

For `openshift`, we want to combine the CA cert of the openshift cluster, which we know because it is mounted with the remote kiali secret, with the system certs. We may also want to allow users to specify a custom CA but I didn't add support for that with this change.

nrfox:
Rebasing with master to fix CI

jmazzitelli (approved):
lgtm

PR #7946, opened 2024-11-19
Use OCI image for keycloak

nrfox:
### Describe the change

Use the oci image of the keycloak charts. Necessary as of Nov 18th: https://blog.bitnami.com/2024/10/bitnami-helm-charts-moving-to-oci.html.

### Steps to test the PR

CI passes.


jshaughn (approved):
Thanks for fixing!

PR #7949, opened 2024-11-21
[bookinfo] Support usage of ingress gateway from different namespace as istio control plane in the traffic generator

mkralik3:
### Describe the change

Support usage of ingress gateway from different namespace as istio control plane in the traffic generator


jmazzitelli (approved):
lgtm

Issue #7954, opened 2024-11-22
envoy metrics tab missing when using user-monitoring prometheus in OpenShift


jmazzitelli:
On OpenShift, if you install Kiali, Istio, bookinfo normally, in the Kiali UI's Workload page for details-v1 workload, you'll see Envoy>Metrics tab with the envoy metrics.

If you then switch to using the user-montioring Prometheus that is built into OpenShift, the Envoy>Metrics tab disappears.

Here's how to see this (note - you need your OpenShift pull secret file on the filesystem so you can pass it to the crc hack script via the -p option).

1. Put your pull secret on the file system - I will assume it is in `$HOME/pull-secret.txt`
2. Install everything (CRC/OpenShift, Istio, Bookinfo demo, Kiali):
```bash
hack/crc-openshift.sh start --enable-cluster-monitoring true -p $HOME/pull-secret.txt
oc login -u kubeadmin -p kiali https://api.crc.testing:6443 --insecure-skip-tls-verify
podman login --tls-verify=false -u kubeadmin -p $(oc whoami -t) default-route-openshift-image-registry.apps-crc.testing
hack/istio/install-istio-via-istioctl.sh -cp openshift
hack/istio/install-bookinfo-demo.sh -c oc -tg
make build build-ui cluster-push operator-create kiali-create
```
3. After Kiali starts, log into the Kiali UI (username and password are both `kiali`), go to the "details-v1" workload, and look at the Envoy tab - you should see a Metrics sub-tab showing you envoy metrics.
```bash
xdg-open https://kiali-istio-system.apps-crc.testing/console/namespaces/bookinfo/workloads/details-v1?tab=envoy
```

![image](https://github.com/user-attachments/assets/cdbac3cf-8908-4ce5-94bc-3ca23453bc3f)

4. Now that you see it working, switch over to using the user-monitoring Prometheus that is built into OpenShift (NOTE! If you somehow get a "failed to verify certificate" error in the server logs, you need [this PR](https://github.com/kiali/kiali/pull/7956) and then pass in `-pi true` to this command in addition to the other options)
```bash
hack/use-openshift-prometheus.sh -n bookinfo
```

5. Wait for the Kiali pod to restart (follow its progress via `oc get pods -n istio-system -l app=kiali`). Once it restarts, then go back to the Envoy tab and see the Metrics sub-tab is now missing:

```bash
[...]

PR #7956, opened 2024-11-23
[hack] when setting up user-monitoring on openshift, be able to connect to Prom insecurely

jmazzitelli:
This was thought to be needed when working on https://github.com/kiali/kiali/issues/7954 - however, it doesn't seem to be required. I was getting an "failed to verify cert" error at one point, and this was needed. But I cannot repeat that failure. Leaving this PR open since it could be useful and may be needed under some situations.

### DIFF
--- a/hack/use-openshift-prometheus.sh
+++ b/hack/use-openshift-prometheus.sh
@@ -16,6 +16,7 @@ KIALI_CR_NAMESPACE=""
 MESH_LABEL="mymesh"
 NAMESPACES=""
 NETWORK_POLICIES="false"
+PROM_INSECURE="false"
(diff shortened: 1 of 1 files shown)


jmazzitelli:
This might not be needed. I see Kiali able to query prom on CRC without this. However, I do see that we set `external_services.prometheus.auth.ca_file` to `/kiali-cabundle/service-ca.crt` (I think the operator does that). It could be our script is overwriting that and setting it to empty string. If that is true, than I need to change this script so it leaves that ca_file setting alone.

hhovsepy (approved):
Verified, works fine on remote OCP cluster.

PR #7961, opened 2024-11-26
[graph] more tweaks to layout/reset

### LINKED ISSUE #7935: [graph] when zoomed-in "reset view" not resizing correctly
"Reset view" should perform a layout AND fit to the current screen. When zoomed-in The layout does not always fit inside the current screen.  Similar behavior can be seen in other situations, like graph hide.

For example:

[Screencast From 2024-11-18 21-34-15.webm](https://github.com/user-attachments/assets/e7ef22ac-6c89-48f2-9ce0-25a49c5705bd)

jshaughn:
Closes #7935 (again)

Increase the delay before performing a resize fit.  This seems to be enough in several tries to break it.

Also update react-resize-detector to the highest version prior to dropping support for react 17.

There are still times when PFT seems to just slightly lose its scale, I can't figure how to fix or avoid it.

To test you can just try various things, combining resize, changing layouts, zooming and panning, graph hide, side panel collapse/expand, etc.  Unless something horrible happens it's probably OK, even if it does sort of start laying out slightly too big.

### DIFF
--- a/frontend/src/pages/GraphPF/GraphPF.tsx
+++ b/frontend/src/pages/GraphPF/GraphPF.tsx
@@ -326,7 +326,7 @@ const TopologyContent: React.FC<{
       if (layoutInProgress === LayoutType.Resize) {
         setTimeout(() => {
           controller.getGraph().fit(FIT_PADDING);
-        }, 250);
+        }, 500);
       } else {
         controller.getGraph().fit(FIT_PADDING);
       }
--- a/frontend/src/pages/GraphPF/GraphPF.tsx
+++ b/frontend/src/pages/GraphPF/GraphPF.tsx
@@ -322,6 +322,8 @@ const TopologyContent: React.FC<{
     }
 
     if (layoutInProgress !== LayoutType.LayoutNoFit) {
+      controller.getGraph().fit(FIT_PADDING);
+
       // On a resize, delay fit to ensure that the canvas size updates before the fit
       if (layoutInProgress === LayoutType.Resize) {
         setTimeout(() => {
--- a/frontend/src/pages/GraphPF/GraphPF.tsx
+++ b/frontend/src/pages/GraphPF/GraphPF.tsx
@@ -82,7 +82,7 @@ const DEFAULT_NODE_SIZE = 40;
 const ZOOM_IN = 4 / 3;
 const ZOOM_OUT = 3 / 4;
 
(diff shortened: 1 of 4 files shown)


jshaughn:
> we can fit the graph twice...

Great idea.  Applied and pushed!

frontend/src/pages/GraphPF/GraphPF.tsx
```diff
@@ -322,11 +322,13 @@ const TopologyContent: React.FC<{
     }
 
     if (layoutInProgress !== LayoutType.LayoutNoFit) {
+      controller.getGraph().fit(FIT_PADDING);
```
ferhoyos, line 325: This needs some refactoring. If the layout type is not Resize, you will fit the layout twice without any delay.

My suggestion (simply remove the else statement in the condition):

```
if (layoutInProgress !== LayoutType.LayoutNoFit) {
      controller.getGraph().fit(FIT_PADDING);

      // On a resize, delay fit to ensure that the canvas size updates before the fit
      if (layoutInProgress === LayoutType.Resize) {
        setTimeout(() => {
          controller.getGraph().fit(FIT_PADDING);
        }, 500);
      }
    }
```

jshaughn: @ferhoyos OK, I cleaned it up.  I made one additional small tweak to the padding size, but I don't see anything else to do at this time.

frontend/src/pages/GraphPF/GraphPF.tsx
```diff
@@ -886,6 +887,10 @@ const TopologyContent: React.FC<{
                   zoomOutCallback: () => {
                     controller && controller.getGraph().scaleBy(ZOOM_OUT);
                   },
+                  // currently unused
+                  fitToScreenCallback: () => {
+                    controller.getGraph().fit(80);
```
ferhoyos, line 892: Why do you hardcode 80 here? It should be the FIT_PADDING value right?

jshaughn: good grief, I need a holiday, it seems.  That code isn't executed, but yes.  I'll push the change.

jshaughn: done.

ferhoyos (approved):
LGTM

PR #7964, opened 2024-11-26
Support bookinfo ingress gw on OCP in install-testing-demos script

mkralik3:
- Support bookinfo ingress gw on OCP in install-testing-demos script (moved gateway_yaml before if-else and used it for openshift case too )
- Add route for bookinfo ingress gateway in bookinfo ns
- Use that route in traffic generator when exists (1st, it tries `istio-ingressgateway` route (in ingressgateway/controlplane ns), 2nd it tries bookinfo-gateway-istio (in bookinfo app ns), 3rd it uses a direct route to productpage

### DIFF
--- a/hack/istio/install-bookinfo-demo.sh
+++ b/hack/istio/install-bookinfo-demo.sh
@@ -366,7 +366,8 @@ sleep 4
 # Expose the OpenShift routes
 if [ "${IS_OPENSHIFT}" == "true" ]; then
   $CLIENT_EXE expose svc/productpage -n ${NAMESPACE}
-  $CLIENT_EXE expose svc/istio-ingressgateway --port http2 -n ${ISTIO_NAMESPACE}
+  $CLIENT_EXE expose svc/istio-ingressgateway --port http2 -n ${INGRESS_NAMESPACE} --name=istio-ingressgateway
+  $CLIENT_EXE expose svc/bookinfo-gateway-istio --port=http -n ${NAMESPACE} --name=bookinfo-gateway-istio
 fi
 
 echo "Bookinfo Demo should be installed and starting up - here are the pods and services"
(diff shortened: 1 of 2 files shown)


hack/istio/install-bookinfo-demo.sh
```diff
@@ -399,12 +400,21 @@ fi
...
   if [ "${IS_OPENSHIFT}" == "true" ]; then
-    echo "Determining the route to send traffic to"
+    echo "Determining the route to send traffic to, 1st, trying istio-ingressgateway route in ${INGRESS_NAMESPACE} namespace"
+    # first, try istio-ingressgateway in istio-system
     INGRESS_ROUTE=$(${CLIENT_EXE} get route istio-ingressgateway -o jsonpath='{.spec.host}{"\n"}' -n ${INGRESS_NAMESPACE})
-    while [ -z "${INGRESS_ROUTE}" ]; do
+    if [ -z "${INGRESS_ROUTE}" ]; then
       sleep 1
-      INGRESS_ROUTE=$(${CLIENT_EXE} get route productpage -o jsonpath='{.spec.host}{"\n"}' -n ${NAMESPACE})
-    done
```
nrfox, line 407: I'm guessing this was a while loop before to give the ingress route time to be populated? We may need to check more than once before giving up and going onto bookinfo-gateway-istio. Maybe wait for the ingress status to have a `host` for some reasonable timeout?
```
kubectl wait '--for=jsonpath={.status.ingress[].host}' --timeout=10s route istio-ingressgateway -n ${INGRESS_NAMESPACE}
```

mkralik3: ah, yea. 
However, to support JSONPath condition without value, some newer kubectl/oc is needed, I am not sure which version is used in our CI

nrfox: Ok maybe then just keep the while loop but exit after ~10 tries?

nrfox (approved):

hack/istio/install-bookinfo-demo.sh
mkralik3: It works on our CI with
```
oc version
Client Version: 4.16.6
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: 4.17.4
Kubernetes Version: v1.30.5
```
so I left it in that way.

PR #7981, opened 2024-12-09
[security] update path-to-regexp

jshaughn:
Resolves OSSM-8528 / CVE-2024-52798 

This will need to be backported to v1.73 for OSSM 2.6.

### DIFF
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -9821,12 +9821,7 @@ path-scurry@^2.0.0:
     lru-cache "^11.0.0"
     minipass "^7.1.2"
 
-path-to-regexp@0.1.10:
-  version "0.1.10"
-  resolved "https://registry.yarnpkg.com/path-to-regexp/-/path-to-regexp-0.1.10.tgz#67e9108c5c0551b9e5326064387de4763c4d5f8b"
-  integrity sha512-7lf7qcQidTku0Gu3YDPc8DJ1q7OOucfa/BSsIwjuh56VU7katFvuM8hULfkwB3Fns/rsVF7PwPKVw1sl5KQS9w==
-
-path-to-regexp@^1.7.0:
+path-to-regexp@0.1.10, path-to-regexp@^1.7.0, path-to-regexp@^1.9.0:
   version "1.9.0"
   resolved "https://registry.yarnpkg.com/path-to-regexp/-/path-to-regexp-1.9.0.tgz#5dc0753acbf8521ca2e0f137b4578b917b10cf24"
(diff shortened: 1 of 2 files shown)


frontend/yarn.lock
```diff
@@ -9821,17 +9816,10 @@ path-scurry@^2.0.0:
     lru-cache "^11.0.0"
     minipass "^7.1.2"
 
-path-to-regexp@0.1.10:
-  version "0.1.10"
-  resolved "https://registry.yarnpkg.com/path-to-regexp/-/path-to-regexp-0.1.10.tgz#67e9108c5c0551b9e5326064387de4763c4d5f8b"
-  integrity sha512-7lf7qcQidTku0Gu3YDPc8DJ1q7OOucfa/BSsIwjuh56VU7katFvuM8hULfkwB3Fns/rsVF7PwPKVw1sl5KQS9w==
-
-path-to-regexp@^1.7.0:
```
ferhoyos, line 9829: This CVE doesn't affect the version 1.x of `path-to-regexp` library. Since this entry has the caret ^, I would update it to the latest 1.x version just in case.

ferhoyos (changes requested):
This CVE only affects version 0.x of `path-to-regexp` dependency. Since it is a fixed version imported from `express` library, I would update `express` to the latest version instead of adding `path-to-regexp` entry in resolutions. In my opinion, the less libraries we have in the resolution section, the better.

frontend/yarn.lock
jshaughn: I saw that but then was a bit confused by all of the versions offered by this package: https://www.npmjs.com/package/path-to-regexp?activeTab=versions

Do you suggest `^1.9.0` ?

jshaughn: On the assumption that it is preferred, I updated to 1.9.0.

ferhoyos: Sorry @jshaughn, I think I have not been clear enough 🙂  

In my opinion, the **resolutions** section should only include those dependencies that really need a fixed version, which is not the case for `path-to-regexp`. Then, what to do to update the dependency?

- **For version 1.x of `path-to-regexp`**:  
  The dependency has a caret (`^`), so to update it to the latest version, just remove the entry in `yarn.lock`, execute `yarn update`, and let the yarn magic happen.

- **For version 0.x of `path-to-regexp`**:  
  The dependency does not have a caret, so the above process does not work here. If you look further, this dependency is imported by `express`, which updates `path-to-regexp` to the latest version `0.1.12` in its latest release — [comparison link](https://github.com/expressjs/express/compare/4.21.1...4.21.2).  
  In this case, what you need to do is update `express` to the latest version removing the entry in `yarn.lock` and executing `yarn update`.

ferhoyos (approved):
LGTM

PR #7982, opened 2024-12-09
[graph] fix traffic animation with very small error rates

jshaughn:
fix traffic animation issue where a very small error rate resulted in a uniform stream of error trafficpoints.

Related to #7934 

@josunect This should resolve the issue we saw with the very small error rate resulting in all error points.  Note that with the fix you may not see any error trafficPoints, the error rate of 0.3 is very small and may get rounded down to 0 in the new approach.


josunect (approved):
Looks good!

image

image

PR #7983, opened 2024-12-09
[security] update nanoid

jshaughn:
Resolves OSSM-8538 / CVE-2024-55565

This will need to be backported to v1.73 for OSSM 2.6.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -159,6 +159,7 @@
     "typescript": "^4.9.5"
   },
   "resolutions": {
+    "nanoid": "^3.3.8",
     "nth-check": "^2.0.1",
     "postcss": "^8.4.31",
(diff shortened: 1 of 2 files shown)


ferhoyos (approved):
The change works. However, I think it is better to keep the `resolutions` section as short as possible, including there only those dependencies that really need a fixed version (e.g., two different major versions that get into conflict), which is not the case for `nanoid`. The list would now have more than 25 entries if we added each dependency version to be fixed.

To update `nanoid` to the latest version without modifying the `resolutions` section:

Since `nanoid` has a caret (`^`), to update it to the latest version, just remove the entry in `yarn.lock`, execute `yarn update`, and let yarn's update the dependency to the latest version.

ferhoyos (approved):
LGTM

PR #7986, opened 2024-12-10
Fix install testing demos when GATEWAY_HOST was not set

mkralik3:
### Fix for issue
Bookinfo app doesn't work when it is installed by `install-testing-demos.sh` script.

### Describe the change

```
  elif [ -v GATEWAY_HOST ]; then
```
is always true, since `-v` check if a variable [was set, not if it is empty or not](https://www.gnu.org/savannah-checkouts/gnu/bash/manual/bash.html#Bash-Conditional-Expressions)
It is always true, since by default, GATEWAY_HOST is an empty string ( https://github.com/kiali/kiali/blob/master/hack/istio/install-testing-demos.sh#L27 )

So when I run `install-testing-demos.sh` without gateway host param, GATEWAY_HOST is an empty string and the script sets
```
    hosts:
    - ""
```
to the Gateway resource which is created in the temp file (e.g. `/tmp/tmp.8LV8qsrLq`R) which is passed and will be used in `./install-bookinfo-demo.sh` script.

However that Gateway is not valid according to istio validation webhook:
```
on is invalid: 2 errors occurred:
	* short names (non FQDN) are not allowed
	* empty domain name not allowed
```
so bookinfo app and traffic generator don't work.

### Fix 
There should be check `-z` instead, which checks whether GATEWAY_HOST is not an empty string (True if the length of string is zero.)
Check that gateway host is not empty:
```
elif ! [ -z "$GATEWAY_HOST" ]; then
```

cypress tests from upstream pipeline after this fix: https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/kiali/job/test-jobs/job/kiali-cypress-tests/3356/

### DIFF
--- a/hack/istio/install-testing-demos.sh
+++ b/hack/istio/install-testing-demos.sh
@@ -124,7 +124,7 @@ if [ "${DELETE_DEMOS}" != "true" ]; then
   gateway_yaml=""
   if [ "${USE_GATEWAY_API}" == "true" ]; then
     gateway_yaml="${ISTIO_DIR}/samples/bookinfo/gateway-api/bookinfo-gateway.yaml"
-  elif [ -v GATEWAY_HOST ]; then
+  elif ! [ -z "$GATEWAY_HOST" ]; then
     gateway_yaml=$(mktemp)
     cat << EOF > "${gateway_yaml}"
 apiVersion: networking.istio.io/v1
@@ -200,7 +200,7 @@ EOF
     "${SCRIPT_DIR}/install-sleep-demo.sh" -c kubectl -in ${ISTIO_NAMESPACE} -a ${ARCH} ${AMBIENT_ARGS_BOOKINFO}
   fi
 
-  if [ -v "${GATEWAY_HOST}" ]; then
+  if [[ -z "$GATEWAY_HOST" && "${USE_GATEWAY_API}" != "true" ]]; then
     # Assume that the '*' is used for hosts if the gateway host is not specified.
     # Some front-end tests have conflicts with the wildcard host in the bookinfo-gateway. Patch it with the host resolved for the traffic generator.
     ISTIO_INGRESS_HOST=$(${CLIENT_EXE} get cm -n bookinfo traffic-generator-config -o jsonpath='{.data.route}' | sed 's|.*//\([^\:]*\).*/.*|\1|')


jshaughn (commented):
There is a second use `-v` at line 203, does that also need to change?

mkralik3:
@jshaughn good catch, I didn't look elsewhere :D

mkralik3:
however, according to the comment there
```
  if [ -v "${GATEWAY_HOST}" ]; then
    # Assume that the '*' is used for hosts if the gateway host is not specified.
    # Some front-end tests have conflicts with the wildcard host in the bookinfo-gateway. Patch it with the host resolved for the traffic generator.
```
that part of code should be executing when GATEWAY_HOST is empty, not when GATEWAY_HOST is not empty as in the previous occurrence. (probably type from origin commit )

mkralik3:
Verified that 

- `./install-testing-demos.sh` (no gateway api or gateway host specified) with cypress test suite from upstream istio pipeline https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/kiali/job/test-jobs/job/kiali-cypress-tests/3360/ 

- `./install-testing-demos.sh -gw true` (gw api is selected) with integration test suite from downstream pipeline (since upstream istio pipeline uses the old way of installing via bookinfo script only so it is not affected by this PR) https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/kiali/job/test-jobs/job/kiali-integration-tests/3136/ (only expected failures since testing v2.1 instance with master branch, I saw these issues before this PR so not related to this) 

- `./install-testing-demos.sh -g 'istio-ingressgateway-istio-system.apps.user-rhos-d-3.servicemesh.rhqeaws.com'` (gateway host is specified), installed manually and run cypress tests on it https://master-jenkins-csb-servicemesh.apps.ocp-c1.prod.psi.redhat.com/job/kiali/job/test-jobs/job/kiali-cypress-tests/3362/ 

Manually verified that:

- when `-g` param is used (gateway is explicitly specified), a host in the bookinfo virtual service will not be updated at the end according to traffic generator
- when `-g` param is not used, a host in the bookinfo virtual service will be updated according to the same host as traffic generator uses
- when `gw` param is used, gateway api resources are used and no update in virtual service ( since gw api doesn't have that resource so it doesn't exist )

work correctly :+1:  Ready to merge

jshaughn (approved):
Good work, @mkralik3 !

Issue #7958, opened 2024-11-25
Time duration component in kiosk mode does not handle the URL correctly


ferhoyos:
### Describe the Bug

In Kiosk mode, duration and refresh settings are handled by a time modal since the toolbar dropdowns are not shown.

This time duration component does not handle the URL parameters correctly, failing to read/write the `duration`/`refresh`/`rangeDuration` parameters.

### Expected Behavior

The time duration component should:
1. Load the appropriate time settings from the URL parameters.
2. Update the URL parameters when the user changes the time settings values.

### Steps to Reproduce the Bug

1. Navigate to any workload detail page (e.g., productpage). Verify that the `duration` and `refresh` URL parameters are set (e.g., `/console/namespaces/bookinfo/workloads/productpage-v1?duration=60&refresh=60000`).

2. Enable kiosk mode by appending `&kiosk=true` to the URL and refresh the page. The time duration component should appear on the right side of the screen.

   ![Time Modal in Kiosk Mode](https://github.com/user-attachments/assets/ce0ac0a5-434f-43ba-a041-109fc0d8b429)

3. Modify the `refresh` parameter in the URL to, for example, `&refresh=30000`, and refresh the page. The time duration component should load the new value (30s), but it does not.

4. Click on the time duration component to open the time modal. Modify any values and click the **Confirm** button. The modifications do not update the corresponding parameters in the URL.

   ![Time Modal Confirmation](https://github.com/user-attachments/assets/94242f3e-7776-49ef-b004-781b9fdf722e)

### Environment

- **Kiali Version:** master  
- **Istio Version:** 1.24.0  
- **Kubernetes Implementation:** Kind  
- **Kubernetes Version:** 1.31.0

PR #7991, opened 2024-12-11
Handle URL in kiosk for time duration modal

josunect:
### Describe the change

Handle URL in kiosk for time duration modal

### Steps to test the PR

Go to: http://localhost:3000/console/namespaces/bookinfo/workloads/productpage-v1?duration=604800&refresh=30000&kiosk=true&tab=logs&rangeDuration=604800

Make sure that the selector is: 
image

Also in the modal: 
image

Update the modal data: 
image

Make sure it changes in the URL. 

Note: For Overview & Traffic, the duration parameter is used. For the other tabs, rangeDuration. Where rangeDuration is used, the "to" and "from" parameters can also be used:

http://localhost:3000/console/namespaces/bookinfo/workloads/productpage-v1?duration=604800&refresh=60000&kiosk=true&tab=traces&from=1733912384867&to=1733912700115

image

### Automation testing

I think this will be validated in OSSMC tests (Can be done in a follow up PR)

### Issue reference

Fixes https://github.com/kiali/kiali/issues/7958

### DIFF
--- a/frontend/src/components/Time/TimeDurationModal.tsx
+++ b/frontend/src/components/Time/TimeDurationModal.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { Button, Form, FormGroup, Modal, ModalVariant, TooltipPosition } from '@patternfly/react-core';
 import { UserSettingsActions } from '../../actions/UserSettingsActions';
-import { HistoryManager, URLParam } from '../../app/History';
+import { HistoryManager, location, URLParam } from '../../app/History';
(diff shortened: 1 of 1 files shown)


ferhoyos (approved):
The PR works fine, thanks Josune!

The only minor comment is that all URL variables are calculated on every render but are only used during the component's initialization (in the getInit methods and React.useEffect initialization). It might be more efficient to consolidate everything (state initialization and dispatch) into a single React.useEffect, but it may not be worth the effort—up to you.

ferhoyos (approved):
LGTM

Issue #7937, opened 2024-11-19
In ambient mode, graph not contain gateway traffic


erikwoo:
### What do you want to discuss?
ingressgateway contain prom metrics istio_request_totoal with reporter=source
When istio is ambient, kiali would query with filter reporter=~destination|waypoint. 

### What does it improve / What problem does it solve?
Should we remove the reporter promql filter so that kiali graph can still display gateway traffic?

jshaughn:
@erikwoo , the graph querying approach is fairly complex and any changes need to be heavily considered. To start, please just supply some replication steps, so we can confirm exactly what you are seeing. Thanks!

josunect:
@erikwoo I was trying to reproduce this in Ambient mesh and Kiali is showing the ingress traffic: 

![image](https://github.com/user-attachments/assets/87bb03de-51d6-483d-9729-fcf291d90cee)
![image](https://github.com/user-attachments/assets/e4ee79b1-ec4b-4100-a76a-7b386bdc5534)

![image](https://github.com/user-attachments/assets/2228e30a-4d8d-4af7-91cc-a5b14ba2041b)

Another example with Ambient and Gateway traffic: 
![image](https://github.com/user-attachments/assets/77426092-0e43-488e-9710-56e7f0006e4c)


It would be really helpful if you can send us some replication steps for this issue, thank you!

erikwoo:
@josunect thank you for your replying.
![image](https://github.com/user-attachments/assets/33ed2df8-8b03-4d3b-90b2-6ac6879dd836)

try filtering with only http traffic

josunect:
[...]

PR #7992, opened 2024-12-11
[ambient] handle gateway-to-ambient traffic in the graph

jshaughn:
HTTP traffic from a gateway to an ambient workload travels via HBONE to ztunnel and reports L7 telem only as source proxy. Our ambient handling depends on L7 telem from a waypoint, but there is no waypoint involved. Because there is no waypoint or Destination proxy involved, we were missing this source-reported telemetry. The only way to solve this seems to be adding more queries, specfically to pick up this traffic :-(.

Notes:
- namespace graphs should now be capturing the data as long as the namespace "IsAmbient".
- node graphs should now be capturing the traffic as long as the node's namespace "IsAmbient".
- responseTime should be available on these edges as long as the namespace "IsAmbient".
- aggregate node handling (aka requestClassification/operation nodes) is set as a TODO until it's clear someone needs the support.
- security appender did not add queries for two reasons:
  - it seems HBONE is not setting connection_security or principal attributes, so adding a query to use thattelem is not useful.
  - if the user has TCP traffic selected, we are already able to utilize the information provided by ztunnel.
- throughput appender did not add queries because the default is already using request (i.e. source) telem.

Fixes https://github.com/kiali/kiali/issues/7937

### DIFF
--- a/graph/api/api_test.go
+++ b/graph/api/api_test.go
@@ -3907,7 +3907,7 @@ func ambientWorkloads(t *testing.T) *business.Layer {
 			},
 		}}
 
-	ns := kubetest.FakeNamespace("bookinfo")
+	ns := kubetest.FakeNamespaceWithLabels("bookinfo", map[string]string{"istio.io/dataplane-mode": "ambient"})
 	k8s := kubetest.NewFakeK8sClient(k8spod1, k8spod2, k8spod3, k8spod4, ns)
 	conf := config.NewConfig()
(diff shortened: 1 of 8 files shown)


jshaughn:
With new queries to pick up gateway->ambient (hbone) traffic:

Namespace Graph (only one namespace selected):
image

Node Graph:
image

josunect (commented):
LG! 
image

josunect (approved):

Issue #7997, opened 2024-12-13
GRPC Jaeger client using old tag for istio multi cluster


josunect:
### Describe the bug

GRPC Jaeger client is using old tag for istio multi cluster, it never detects if the traces are using the tracing tag, which provides the error: 
![image](https://github.com/user-attachments/assets/a8897ed7-213e-4191-87c3-914d47db8159)

### What are the steps to reproduce this bug?

1. setup tracing with jaeger and user_grpc: true
2. Go to traces
3. The message will appear

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:** Master (2.2 snapshot)
- **Istio version:** 1.24
- **Kubernetes impl:** 
- **Kubernetes version:** 
- **Other notable environmental factors:**

PR #7998, opened 2024-12-13
Fix cluster tag issue, improve error messages

josunect:
### Describe the change

- Fix cluster tag issue for Jaeger
- improve error messages
- Improve cluster tag detention method: Instead of doing a query for tags, use a query without the tag and check for the cluster tags in the result. It could happen that the service which we are querying doesn't have traces for the local cluster. The service is chosen randomly, as Jaeger does not allow to do a query for all the services (https://github.com/jaegertracing/jaeger/issues/423)

### Steps to test the PR

It can be tested with: 

With the Kiali config (I think it is the default): 
```  tracing:
    enabled: true
    provider: "jaeger"
    internal_url: "http://localhost:16685/jaeger"
    use_grpc: true
```

### Automation testing
cypress tests updated. Multi cluster should have the tags configured: 
image

### Issue reference

Ref. #7997

### DIFF
--- a/frontend/cypress/integration/featureFiles/service_details_multicluster.feature
+++ b/frontend/cypress/integration/featureFiles/service_details_multicluster.feature
@@ -37,7 +37,7 @@ Feature: Kiali Service Details page for remote cluster
 
   Scenario: See graph traces for ratings service details
     And user sees trace information
-    And an info message "Loading traces for all clusters. Tracing is not configured to store traces per cluster." is displayed
+    And an info message "Loading traces for all clusters. Tracing is not configured to store traces per cluster." is not displayed
     When user selects a trace
     Then user sees trace details
 
@@ -55,6 +55,6 @@ Feature: Kiali Service Details page for remote cluster
     And user is at the details page for the "service" "bookinfo/ratings" located in the "east" cluster
     Then user does not see any inbound and outbound traffic information
 
-  Scenario: See no Inbound Metrics for a service, which is not present in the specific cluster. 
+  Scenario: See no Inbound Metrics for a service, which is not present in the specific cluster.
     And user is at the details page for the "service" "bookinfo/ratings" located in the "east" cluster
     Then user does not see "Inbound" metrics information for the "east" "ratings" "service"
--- a/frontend/cypress/integration/featureFiles/service_details_multicluster.feature
+++ b/frontend/cypress/integration/featureFiles/service_details_multicluster.feature
@@ -37,7 +37,6 @@ Feature: Kiali Service Details page for remote cluster
 
   Scenario: See graph traces for ratings service details
     And user sees trace information
-    And an info message "Loading traces for all clusters. Tracing is not configured to store traces per cluster." is not displayed
     When user selects a trace
     Then user sees trace details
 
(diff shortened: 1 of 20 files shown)


tracing/jaeger/http_client.go
```diff
@@ -55,21 +54,24 @@ func NewJaegerClient(client http.Client, baseURL *url.URL) (jaegerClient *Jaeger
...
 		ignoreCluster = true
 	} else {
-		ignoreCluster = false
+		if r != nil && len(r.Data) == 0 {
+			log.Debugf("[HTTP Jaeger] Error getting query for tracing. cluster tags will be disabled. No data returned for: %s", urlTraces.String())
+			ignoreCluster = true
+		} else {
+			if includeJaegerClusterTag(r.Data) {
+				ignoreCluster = false
+			}
```
nrfox, line 73: Just so I understand, we make an initial query for traces to see if they have the cluster tag present and if they don't then for the rest of the queries we assume there's no cluster tag? Since Istio ~1.21 the cluster tag is now always present by default unless explicitly removed. I don't think we need `ignoreCluster` at all anymore?

Otherwise deciding whether or not to include the cluster tag this way may introduce problems. What if there cluster tag is present but there's just no tracing data yet?

josunect: I think in that case the only risk would be if there are be no traces in Istio < 1.21 (That I don't think it is supported anymore), or if the cluster tag changes for some reason. In that case we could create a config option to have a workaround if this changes? Or being able to disable it if for some reason the tag is not present?

nrfox: For metrics we rely on all the istio standard labels to be present: https://istio.io/latest/docs/reference/config/metrics/#labels which includes cluster. If someone explicitly removes that then likely Kiali queries wouldn't work. Can we treat traces the same way? cluster is no different than the other default tags that we rely on like `app` or `service`. The only difference is that it was added later than the others.

josunect: So, then we would use the cluster tag always? We were using it just in case of multi cluster. But I guess that would help when there are traces from other clusters, but maybe Kiali is configured for single cluster.

nrfox: Right I think Kiali has always used cluster in metrics queries: https://github.com/kiali/kiali/blob/54f4f79e1769bb2b1c650a3aef228df3dc019b54/graph/telemetry/istio/istio.go#L138

josunect: @nrfox updated to use always the cluster tag. 
Also, I've modified the tests to validate there are not Traces where the workload/service/app from the specific cluster doesn't have traffic

frontend/cypress/integration/featureFiles/service_details_multicluster.feature
```diff
@@ -46,6 +45,10 @@ Feature: Kiali Service Details page for remote cluster
     When user selects a trace
     Then user sees span details
 
+  Scenario: Don't see tracing info after selecting a trace
+    And user is at the details page for the "service" "bookinfo/ratings" located in the "east" cluster
+    Then user see no traces
```
nrfox, line 50: The user sees no traces because there's no traffic for this service in the `east` cluster?

nrfox (approved):

frontend/cypress/integration/featureFiles/service_details_multicluster.feature
nrfox: Never mind I read your comment:
> Also, I've modified the tests to validate there are not Traces where the workload/service/app from the specific cluster doesn't have traffic

Issue #8000, opened 2024-12-16
KIA1104 should only shows when there is one route destination but has explicit weight assigned and less than 100 on tcp/tls route


zirain:
### Describe the bug

<img width="1360" alt="image" src="https://github.com/user-attachments/assets/668687ed-c037-43aa-92d0-f5d0b8c4d595" />

### Expected Behavior

should not report this warning

### What are the steps to reproduce this bug?

1. install istio and kiali
2. apply tcp-echo demo
3. apply `samples/tcp-echo/tcp-echo-all-v1.yaml`, then check 

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:** v2.2
- **Istio version:** master
- **Kubernetes impl:**  
- **Kubernetes version:** 
- **Other notable environmental factors:**

hhovsepy:
Thank you for logging and the fix @zirain

PR #8001, opened 2024-12-16
Added missing tests for KIA1104

hhovsepy:
### Describe the change

Missing not provided route Weight check for TCP and TLS Route types in VirtualServices while KIA1104 Validations. (HTTP Route was working fine).
This was causing odd 'Warning KIA1104 The weight is assumed to be 100 because there is only one route destination' then VS had empty Weight in TCP and TLS Routes.
Added the missing similar checks in TCP and TLS routes (aligned to HTTP route).

### Steps to test the PR

Create "samples/tcp-echo/tcp-echo-all-v1.yaml"
You will not see a Warning KIA1104 on that VS.
Screenshot From 2024-12-16 09-39-46

### Automation testing

Added unit tests.

### Issue reference
https://github.com/kiali/kiali/issues/8000


josunect (approved):
LGTM!

PR #8005, opened 2024-12-16
[dual stack] Support dual stack by being able to display multiple IPs for a service

### LINKED ISSUE #8004: [dual stack] support the fact that a service may have muliple ip addresses
Looking at the code, Kiali seems to assume a single service IP, whereas in Dual Stack it could have two IPs, one for each IpFamily.  I'm not sure this is a problem functionally, but may be a display issue when presenting information to the end user.

jshaughn:
Closes #8004

This parses out more IP information from the core_v1.Service, passes it back to the client, and now Service detail will display the multiple IP familes, as needed.

### DIFF
--- a/business/checkers/virtualservices/route_checker_test.go
+++ b/business/checkers/virtualservices/route_checker_test.go
@@ -42,6 +42,61 @@ func TestServiceMultipleChecks(t *testing.T) {
 	assert.NoError(validations.ConfirmIstioCheckMessage("virtualservices.route.singleweight", vals[0]))
 	assert.Equal(vals[0].Severity, models.WarningSeverity)
 	assert.Equal(vals[0].Path, "spec/http[0]/route[0]/weight")
+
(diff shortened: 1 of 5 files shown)


hhovsepy:
Tried to reproduce on dualstack Kind and remote OCP clusters, with Istio instaleld in dualstack mode https://istio.io/latest/docs/setup/additional-setup/dual-stack/
It still shows the IPv6 address only:
Screenshot From 2024-12-19 17-28-53
This is coming from kube cache loaded service:
Screenshot From 2024-12-19 15-02-36
However in Service in Kubernetes it has both ips:
Screenshot From 2024-12-19 15-03-01

hhovsepy:
For IPv4 and IPv6 only :
Screenshot From 2024-12-19 17-54-34
Screenshot From 2024-12-19 17-54-17

hhovsepy (approved):

PR #8008, opened 2024-12-17
[hack] KinD setup - get podman network info

jmazzitelli:


hhovsepy (approved):
Verified locally, tried with dualstack services,
Screenshot From 2024-12-20 09-36-02
 regression tested

PR #8018, opened 2024-12-23
Avoid duplicated tarballs

ferhoyos:
### Describe the change

Currently, `yarn.lock` contains some entries that download the same tarball file. To avoid these duplications, I have merged the entries to ensure consistency. These duplicated tarballs can cause some build systems, such as Konflux, to fail

### Steps to test the PR

Verify that CI test pass

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -163,9 +163,9 @@
     "postcss": "^8.4.31",
     "workbox-webpack-plugin": "^7.1.0"
   },
+  "packageManager": "yarn@1.22.22",
   "engines": {
-    "node": ">=20.0.0",
-    "yarn": ">=1.0.0"
+    "node": ">=20.0.0"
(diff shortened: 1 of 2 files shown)


jmazzitelli:
Isn't the yarn.lock file auto-generated? If so, will this change be reverted when yarn regenerates this file?

ferhoyos:
Well, that's partially correct. `yarn.lock` is auto-generated the first time. Then, any change in `package.json` updates the `yarn.lock` file to make it compatible with the `package.json` file.

Said that, direct changes to `yarn.lock` file are correct if they are compatible with the current `package.json` file.

In this case we are just merging `yarn.lock` entries with the same tarball, so the dependency graph used by `yarn` is the same.

jmazzitelli (approved):

PR #8020, opened 2024-12-24
Update service with no traces

josunect:
### Describe the change

The service ratings for the east cluster, which is not supposed to have traffic, has one initial tracing for some reason, causing a CI flake. 
Update the service for one that is not generating traces. 

Verify that the logic is correct: 

image

image

### Steps to test the PR

CI passes. Also, the tests can be run locally for: 


ferhoyos (approved):
LGTM

Issue #8025, opened 2024-12-27
Irregular metrics loading error 


dan-m8t:
### Describe the bug

I updated Kiali from 1.89 to 2.3.0 in my test environment. As some changes happened to namespace access I migrated to discoverySelectors as the documentation said. I use Istio 1.23.3 in ambient mode.

When I put some traffic onto my workloads I can sometimes see metrics and sometimes not, I am not sure what causes this behaviour.
It just says: 
```
 Error loading metrics: Cannot access namespace data: Namespace [my-namespace] in cluster [Kubernetes] is not accessible to Kiali
```
![grafik](https://github.com/user-attachments/assets/4d3ebd71-9504-409c-ba2f-05553ff07dd3)

When I click on "refresh" inside the graph view metrics are displayed again. I cannot see any log entries here, I enabled debug logging for kiali but it states no errors.

All I get is this, regardless of Kiali UI throws an error or not:

```
{"level":"debug","time":"2024-12-27T13:04:14Z","message":"Scraping istiod for debug info"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"[ValidationsReconciler] Started reconciling "}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"[ValidationsReconciler] Finished reconciling in 3ms"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"Reading session cookie: kiali-token-Kubernetes"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"Reading session cookie: kiali-token-Kubernetes"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"Reading session cookie: kiali-token-Kubernetes"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"Reading session cookie: kiali-token-Kubernetes"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"Reading session cookie: kiali-token-Kubernetes"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"Grafana is disabled in Kiali by configuration"}
{"level":"debug","time":"2024-12-27T13:04:19Z","message":"Tracing is disabled in Kiali by configuration"}
{"level":"debug","time":"2024-12-27T13:04:20Z","message":"Reading session cookie: kiali-token-Kubernetes"}
{"level":"debug","time":"2024-12-27T13:04:20Z","message":"Reading session cookie: kiali-token-Kubernetes"}
[...]

PR #8027, opened 2024-12-30
Add the namespace labels for FilterNamespacesWithDiscoverySelectors

josunect:
### Describe the change
This issue is not always reproducible, but i think it might happen when the namespace is not found in the cache by `GetClusterNamespace`. This method uses `FilterNamespacesWithDiscoverySelectors` but it doesn't always have all the information needed to verify if the namespace is accessible, because the namespaces labels are missing:

image

### Steps to test the PR

I tested with Ambient, but I think it also might happen in a non Ambient Mesh. 
Install istio, kiali and bookinfo, set the discovery selectors: 

```
deployment:
  cluster_wide_access: true
  discovery_selectors:
    default:
      - matchExpressions:
        - key: istio.io/dataplane-mode
          operator: Exists
      - matchExpressions:    
        - key: istio-injection
          operator: Exists  
```

Go to a workload, metrics and let the refresh button. It needs to wait until the cache is empty to be able to reproduce the error. 
When applied this PR, the error should not happen.
A way to "force" this error, can be commenting the code: 

https://github.com/kiali/kiali/blob/master/business/namespaces.go#L274

And see how it always returns namespace not accessible. 

### Automation testing

N/A

### Issue reference

Fixes https://github.com/kiali/kiali/issues/8025

### DIFF
--- a/business/namespaces.go
+++ b/business/namespaces.go
@@ -275,10 +275,6 @@ func (in *NamespaceService) GetClusterNamespace(ctx context.Context, namespace s
 		return &ns, nil
 	}
 
-	if !in.isAccessibleNamespace(models.Namespace{Name: namespace, Cluster: cluster}) {
-		return nil, &AccessibleNamespaceError{msg: "Namespace [" + namespace + "] in cluster [" + cluster + "] is not accessible to Kiali"}
-	}
-
 	var result models.Namespace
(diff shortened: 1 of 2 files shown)


jmazzitelli (dismissed):
LGTM

jmazzitelli (approved):

Issue #6455, opened 2023-08-07
[hack] Keycloak 22+ does not work with multi-cluster hack scripts


nrfox:
The 22+ release of keycloak does not work with the multi-cluster hack scripts setup. There's something wrong with the https setup and/or openid redirect from kiali.

```Error when fetching OpenID provider's metadata: mismatch between the configured issuer_uri (https://keycloak-192-168-50-48.nip.io/realms/kube) and the exposed Issuer URI in OpenId provider metadata (http://keycloak-192-168-50-48.nip.io/realms/kube)```

A temporary workaround is to use the 21+ release of keycloak.

jshaughn:
@nrfox Is this becoming more, or less, urgent?

nrfox:
This looks like it was fixed already by: https://github.com/kiali/kiali/pull/8030

PR #8030, opened 2025-01-02
Update keycloak image and container

### LINKED ISSUE #8029: [hack script] Use keycloak lastest image
Revert this change https://github.com/kiali/kiali/pull/8028 where the previous version of Keycloak helm chart is used to being able to correctly set the admin user and password.

Using version 24.3.1 in the `keycloak.sh` script. 

### What is the new behavior?
Remove the version. Before this, make sure the admin username and password are correctly set and Keycloak is not using the Bitnami defaults.

josunect:
### Describe the change

The admin user/pwd variables of the Keycloak container (21.1.2) doesn't look compatible with the ones updated in the charts in the latest version (24.3.2).
With the defaults, it looks like everything is working, but I just wondering if there is any issue to use that specific container image. 

### Steps to test the PR
`hack/run-integration-tests.sh --test-suite frontend-multi-primary --setup-only true`

### Automation testing
The changes are affecting frontend-multi-primary environment

### Issue reference

Fixes #8029

### DIFF
--- a/hack/keycloak.sh
+++ b/hack/keycloak.sh
@@ -111,7 +111,7 @@ EOF
   echo "Creating keycloak deployment"
   helm upgrade --install --wait --timeout 15m \
   --namespace keycloak \
-   keycloak oci://registry-1.docker.io/bitnamicharts/keycloak \
+   keycloak oci://registry-1.docker.io/bitnamicharts/keycloak --version 24.3.2 \
   --reuse-values --values - <<EOF
 auth:
   createAdminUser: true
--- a/hack/keycloak.sh
+++ b/hack/keycloak.sh
@@ -111,7 +111,7 @@ EOF
   echo "Creating keycloak deployment"
   helm upgrade --install --wait --timeout 15m \
   --namespace keycloak \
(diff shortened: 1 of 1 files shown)


nrfox:
This was pinned because of https://github.com/kiali/kiali/issues/6455. But if the tests are passing then this may not be an issue with the latest image. I'm unsure if we always want to use the latest image though. If there's any breaking change in the charts then it will break our CI unexpectedly. Perhaps we can keep the version pin to 24 but remove the image pin to 21 and have some kind of recurring task to update this dep?

josunect:
> This was pinned because of #6455. But if the tests are passing then this may not be an issue with the latest image. I'm unsure if we always want to use the latest image though. If there's any breaking change in the charts then it will break our CI unexpectedly. Perhaps we can keep the version pin to 24 but remove the image pin to 21 and have some kind of recurring task to update this dep?

Thanks for the reference, @nrfox !  I haven't see it. 
Yes, there is a risk on that. I think it sounds good, but I'm not sure how this recurrent task could be implemented.

nrfox (approved):

Issue #7932, opened 2024-11-18
Formally support previous versions of Istio in Kiali releases


nrfox:
### What do you want to improve?

When a new version of Kiali is released, only the latest version of Istio at the time of release is supported. Older versions of Istio, while they may work anyways, are not formally supported. This means when performing a revision based upgrade of Istio, Kiali may not operate correctly during the upgrade.

### What is the new behavior?

Kiali should be backward compatible 2 versions of Istio to match Istio's backward compatibility.

When users want to upgrade from one version of Istio to the next using a revision based strategy, they should first upgrade their Kiali instance to match the desired Istio version, then deploy their second Istio revision. Kiali should work with both the old revision and the new revision so that it can continue to be used both during the upgrade and after.

jshaughn:
This is going to be necessary for us to make the Mesh page revision-based update support usable.  Promoting to sprint backlog.

jshaughn:
Thinking on this, it seems the main change is to just update what we say about compatibility, and the associated compatibility chart, on kiali.io.  In general, Kiali has been backward compatible, but we haven't wanted to do additional testing to prove that things are actually working. So, in a addition to the documentation change we need to establish some sort of sanity checking, via test automation, that Kiali and the Kiali operator are working against the same versions listed as supported by upstream Istio. Typically latest, latest-1, latest-2.

Note that we can say something like, "with the noted exceptions", meaning, if we do encounter a difficult back-compat issue, I think we could document the problem rather than spending a large effort on the support.  This is mainly to formalize what we already believe to be true, but have been reluctant to say, that Kiali is backward compatible, and do some level of testing to back up that claim.

jshaughn:
@ScriptingShrimp , Do you have any ideas on how we could automate testing for older versions of Istio?  Our merge CI runs head against the latest Istio (I think), but before a Kiali release we'd basically need to also run against the latest-1 and probably latest-2.  @jmazzitelli , this would be true for molecule tests as well, I think.

jmazzitelli:
> this would be true for molecule tests as well, I think.

The molecule tests support testing with specific Kiali CR `spec.version` values, so it is possible to run the tests against older versions of Kiali Server. Just need to ensure the CI environment is installed with the Istio version we are expecting to test with.

ScriptingShrimp:
[...]

PR #8039, opened 2025-01-08
CI molecule test action now tests with multiple Istio versions

jmazzitelli:
Runs molecule tests on environments with different Istio versions installed (by default will test with the last 3 Istio minor releases).

part-of https://github.com/kiali/kiali/issues/7932


jmazzitelli:
Tested and it works - saw it loop around the last 3 minor Istio versions - still tests with helm and OLM, only now it does that 3 times (with the 3 Istio releases). Here's my fork's action showing it is green: https://github.com/jmazzitelli/kiali/actions/runs/12678134457

jshaughn (dismissed):
Kind of a weird default but should be OK.

jmazzitelli:
This runs nightly. It will run all molecule tests with the operator installed via helm and installed via OLM. It will test on three Istio versions (the latest, the one previous to that, and the one previous to that).. so it works like this:

```
FOR Istio versions "X", "X-1", "X-2" DO
  1. Create new cluster and install Istio
  2. Use Helm installer
    a.Run all molecule tests
  3. Use OLM installer
    a. Run all molecule tests
DONE
```

jmazzitelli:
I'm testing in my fork. The test failed after almost exactly 6 hours of run time. I'm not sure, but I think github canceled the test run because it took too long. So I'm guessing we can't run all molecule tests for 3 istio versions. We could either:

1. Only run a small subset of molecule tests or...
2. Only run 2 against istio versions or...
3. Have separate CI actions - one for the latest istio version, one for the previous version (v-1), and one for the previous version before that (v-2).

I think #3 is probably the best since it would still give us a full test run but yet cover all the istio versions we want. I'll play with that idea in my fork and see how it can be done.

jshaughn:
> I think 3 is probably the best since it would still give us a full test run but yet cover all the istio versions we want. I'll play with that idea in my fork and see how it can be done.

This may also be best for isolating a problem on one version, independent of the other runs.

jmazzitelli:
The current code in this PR now performs three nightly runs - one on the latest minor version of Istio (V), one on minor version V-1, and one on minor version V-2. You can manually run this and tell it which Istio version you want by giving it a minor version "offset" - if you say you want to run with version offset of "0", that means you want the latest minor version of Istio. If you specify an offset of "1", that means you want 1 minor version prior to the latest version. Offset of "2" means 2 minor versions prior to the latest. This way you don't have to know what the actual versions of Istio there are, and nightly builds will immediately start using any new releases (picking up the latest z-stream of the minor release to be installed) because each run will ask github what istio versions are available and will determine which one to use.

Can you run these parallel? These are triggered by cron schedules - right now they are run 2 hours apart (3 runs at 2am, 4am, 6am). We could make these, say, 1 minute apart and they would essentially be concurrent. I don't think it matters since these are run overnight so concurrent or not, who cares?

Do they run if there were no changes to the operator code that day? Yes. Even if the operator code doesn't change, these nightly tests still run - which is a good thing because a new Istio version might have been released earlier that day, so testing the current code on a newly released Istio would be something we'd want.

jshaughn (approved):
Good!

Issue #8034, opened 2025-01-06
Workload Traffic tab navigation leaves Overview tab confused


jshaughn:
1. Install Bookinfo
2. Navigate to `productpage-v1` workload
3. Click `Traffic` tab
4. In the `Actions` column  click `View metrics` for the `reviews-v2` row
4.1 This correctly shows the `reviews-v2` inbound metrics, and the breadcrumb is correct
5. Click the `Overview` tab

The breadcrumb still correctly shows `reviews-v2`, but the `Overview` page is incorrectly showing `productpage-v1`

PR #8040, opened 2025-01-08
Fix issue with workload detail navigation

jshaughn:
Fixes https://github.com/kiali/kiali/issues/8034

When navigating from within a workload detail to a different workload detail, ensure that the new workload is fetched. This is the main bug reported in this issue. But furthermore, protect against a similar situation for app detail and service detail.

### Steps to test the PR

See issue for repro steps.  I removed what I think is an unnecessary `await` in one place. Make sure you are ok with the async stuff.  I applied basically the same logic across the three scenarios.

### DIFF
--- a/frontend/src/pages/AppDetails/AppDetailsPage.tsx
+++ b/frontend/src/pages/AppDetails/AppDetailsPage.tsx
@@ -75,29 +75,30 @@ class AppDetails extends React.Component<AppDetailsProps, AppDetailsState> {
     const cluster = HistoryManager.getClusterName() || this.state.cluster;
     const currentTab = activeTab(tabName, defaultTab);
 
-    if (
+    const mustFetch =
+      cluster !== this.state.cluster ||
       this.props.appId.namespace !== prevProps.appId.namespace ||
       this.props.appId.app !== prevProps.appId.app ||
       this.props.lastRefreshAt !== prevProps.lastRefreshAt ||
-      currentTab !== this.state.currentTab ||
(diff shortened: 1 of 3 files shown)


jshaughn:
@ferhoyos I didn't realize this PR had gained a conflict.  I rebased and it should be resolved and ready for review.

ferhoyos:
that's fine @jshaughn. I didn’t have time to look at it today, I’ll review it tomorrow.

ferhoyos (approved):
The issue is fixed and the code looks good to me.

However, I find it a bit confusing that the `View metrics` link behaves differently depending on the context. In the workloads and services, it redirects to a different workload, whereas in applications, it opens a different tab within the same application. I believe this behavior is intentional, as inbound and outbound metrics can be filtered by application but not by workload. Still, it feels a bit odd that the same link behaves differently depending on whether it's for a workload or an application.

jshaughn:
> However, I find it a bit confusing that the View metrics link behaves differently depending on the context. In the workloads and services, it redirects to a different workload, whereas in applications, it opens a different tab within the same application.

I totally agree, @ferhoyos.  In fact, I was questioning whether my changes were correct because I became confused when I noticed the behavior, and there was no onscreen help to let me know what was going on.  Only after seeing the differences in the generated URLs did I understand what we were trying to do.

I'm going to merge this fix, but maybe I'll add some help in the future.  I'm not sure how often this feature is even used, so not sure we should put a lot of effort into it.

PR #8041, opened 2025-01-08
replace use of crypto/md5 with crypto/sha256

jshaughn:
replace use of crypto/md5 with crypto/sha256
- regenerate the expected graph test output files due to the new element ids produced by the new hashing

Closes https://issues.redhat.com/browse/OSSM-8373


jmazzitelli (approved):
- code changes look innocent enough.
- unit tests pass.
- I installed on minikube and everything seems to work

jshaughn:
@nrfox I'm going to merge this based on mazz's approval.  If you see anything then let me know, but it seems pretty straightforward.  The main diff is that the hash values are now longer strings.

PR #8042, opened 2025-01-09
Catch OpenShift login error and display an error

### LINKED ISSUE #7925: [Multicluster] When login to remote cluster fails, no visible error appears

Remote cluster login error handling for `openshift` strategy.

When login to a remote cluster fails, you are redirected back to the overview page but no error message is shown. The error message is contained in the URL but this is not visible to users.

You can test this by giving by setting the remote `OAuthClient` to a bad cluster name in the url.

### What is the new behavior?

An error message is shown when login to a remote cluster fails with the reason for failure and ideally a hint for how to fix it.

leandroberetta:
Fixes #7925.

### DIFF
--- a/frontend/src/routes/WildcardRoute.tsx
+++ b/frontend/src/routes/WildcardRoute.tsx
@@ -11,7 +11,7 @@ export const WildcardRoute = (): JSX.Element => {
   const cluster = searchParams.get('cluster');
 
   if (openshiftError) {
-    const clusterMessage = cluster ? `to cluster ${cluster}` : 'to cluster';
+    const clusterMessage = cluster ? `to cluster "${cluster}"` : 'to cluster';
     AlertUtils.addError(
       `Openshift authentication ${clusterMessage} failed`,
       undefined,
--- a/frontend/src/routes/WildcardRoute.tsx
+++ b/frontend/src/routes/WildcardRoute.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import { Navigate, useSearchParams } from 'react-router-dom-v5-compat';
+
+import * as AlertUtils from '../utils/AlertUtils';
+
+export const WildcardRoute = (): JSX.Element => {
+  const [searchParams] = useSearchParams();
+  const openshiftError = searchParams.get('openshift_error');
+
+  if (openshiftError) {
+    AlertUtils.addError(
+      'Openshift authentication to remote cluster failed',
+      undefined,
+      undefined,
+      undefined,
+      openshiftError
+    );
+  }
+
+  return <Navigate to="/overview" replace />;
+};
--- a/frontend/src/routes/WildcardRoute.tsx
+++ b/frontend/src/routes/WildcardRoute.tsx
@@ -3,13 +3,17 @@ import { Navigate, useSearchParams } from 'react-router-dom-v5-compat';
 
 import * as AlertUtils from '../utils/AlertUtils';
 
+const OVERVIEW_ROUTE = '/overview';
+
 export const WildcardRoute = (): JSX.Element => {
   const [searchParams] = useSearchParams();
   const openshiftError = searchParams.get('openshift_error');
+  const cluster = searchParams.get('cluster');
 
   if (openshiftError) {
+    const clusterMessage = cluster ? `to cluster ${cluster}` : 'to cluster';
     AlertUtils.addError(
-      'Openshift authentication to remote cluster failed',
+      `Openshift authentication ${clusterMessage} failed`,
       undefined,
       undefined,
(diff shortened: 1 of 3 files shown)


frontend/src/routes/WildcardRoute.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { Navigate, useSearchParams } from 'react-router-dom-v5-compat';
+
+import * as AlertUtils from '../utils/AlertUtils';
+
+export const WildcardRoute = (): JSX.Element => {
+  const [searchParams] = useSearchParams();
+  const openshiftError = searchParams.get('openshift_error');
```
nrfox, line 8: Could searching for the `openshift_error` query param to add this alert affect other parts of the UI that might be using the same query param like the login page? 
https://github.com/kiali/kiali/blob/dadd63d4bec2c7a306a58a299f7a61145bfcdb8d/frontend/src/pages/Login/LoginPage.tsx#L182-L184

frontend/src/routes/WildcardRoute.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+  if (openshiftError) {
+    AlertUtils.addError(
+      'Openshift authentication to remote cluster failed',
```
nrfox, line 12: I'm not sure I would call this a "remote" cluster although it is remote to Kiali itself. It would be nice if we could put the cluster name in the alert message.

frontend/src/routes/WildcardRoute.tsx
leandroberetta: I don't think it's causing troubles as I think the query parameter is used in two separate phases but, I think we can change the key name of the code to something like you're proposing to avoid any issues.

Also, I can add the cluter name for sure.

frontend/src/routes/WildcardRoute.tsx
leandroberetta: I guess is a remote cluster, right, or is there any other case?. I will add the cluster name.

nrfox: > I guess is a remote cluster, right, or is there any other case?. I will add the cluster name.

It's "remote" to where Kiali is installed but in terms of Istio deployments it could be another primary in a multi-primary deployment so I'm worried that calling it "remote" could confuse people.

frontend/src/routes/WildcardRoute.tsx
leandroberetta: I was reviewing the key name change, and I saw that the generation of that query param occurs only in:

https://github.com/leandroberetta/kiali/blob/master/handlers/authentication/openshift_auth_controller.go#L88

And, really, any of those errors can happen when:

1. You login to Kiali (with the Login page)
2. You login to a "remote to Kiali" cluster

And as I mentioned, they occur at different stages (I guess)

1. When you're login into Kiali (pretty normal use case), you potentially will get one of these errors and stay in that login page (error is shown in the form)
2. When you're already logged in into Kiali and want to login to an another cluster, you will get one of these errors in the overview page (default landing page), and this is when we are picking the query param and generating an alert.

Also I want to confirm than once the alert is generated, the query parameter is gone so refreshing the page won't  fire an alert again.

So, cluster name can be added but I don't think it's ok to change the name of the query parameter.

leandroberetta:
I've added the cluster name, which is the cluster name is in the code, and it seems to be gather from the url, for that reason displays "west2", because that is the typo that generates the error:

image

I don't believe yet there is a way to get the intended cluster name. At least is a hint, meaning, user can tell that is not a good clustername.

Anyway I will spend some more time trying to get the good one (west).

frontend/src/routes/WildcardRoute.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const OVERVIEW_ROUTE = '/overview';
+
+export const WildcardRoute = (): JSX.Element => {
+  const [searchParams] = useSearchParams();
+  const openshiftError = searchParams.get('openshift_error');
+  const cluster = searchParams.get('cluster');
+
+  if (openshiftError) {
+    const clusterMessage = cluster ? `to cluster ${cluster}` : 'to cluster';
```
nrfox, line 14: nit: Quotes may make the name more obvious
```suggestion
    const clusterMessage = cluster ? `to cluster "${cluster}`" : 'to cluster';
```

nrfox (dismissed):
Looks good! Just 1 nit but non-blocking.

We'll also want an e2e cypress test for this that has the `@openshift` tag but that can be a follow up change.

nrfox (approved):

PR #8043, opened 2025-01-09
Do not send a stack trace to the user, limit exposure to the kiali po…

jshaughn:
Do not send a stack trace to the user, limit exposure to the kiali pod logs.  The only occurrence of this was in the graph panic handling, from what I can see.

Closes https://issues.redhat.com/browse/OSSM-8374

After the Change:

Screenshot From 2025-01-09 11-32-45

Testing this is a bit hard, because it requires a graph error, which is an unexpected condition. I just added a forced Panic in the server's graph gen code to provide the screenshot above.  I think the change clear from the simple code change.


jmazzitelli (approved):
lgtm

PR #8045, opened 2025-01-09
run integration tests on different istio versions

jmazzitelli:
This runs all integration tests 2 times nightly - at 2am it runs with the minor version previous to the latest version; at 4am it runs with the minor version that is 2 previous versions back.

So if the latest version of Istio is 1.25.4 and the previous two minor versions are 1.24.8 and 1.23.0 then those last two are run at 2am and 4am respectively.

You can alternatively run this CI action manually - you can either give a specific Istio version to test with (e.g. "1.24.2") or you can specify an offset to the latest minor version. If you give an offset of "1" it will run with the Istio minor version previous to the current minor version. "2" will use the version before that one, and so on. You can opt to specify negative offsets if you find that more intuitive - it doesn't matter. "-1" offset is the same as "1" offset - its the first minor version prior to the latest. Same with "-2" and "2" - its the second minor version prior to the latest.

### DIFF
--- a/.github/workflows/test-istio-version.yml
+++ b/.github/workflows/test-istio-version.yml
@@ -3,24 +3,23 @@ name: Test Istio Version
 on:
   schedule:
   # These are in UTC time.
-  # If you change any of these, you must also change the switch statment in the determine-istio-version-to-use task.
+  # If you change any of these, you must also change the case statment in the determine-istio-version-to-use task.
   - cron: '0 2 * * *'
   - cron: '0 4 * * *'
-  - cron: '0 6 * * *'
   workflow_dispatch:
     inputs:
       istio_version:
-        description: "The version of Istio to test with. Format is either #.#.# or #.#-dev. If specified, this overrides istio_minor_version_offset."
+        description: "istio_version: The version of Istio to test with. Format is either #.#.# or #.#-dev. If specified, this overrides istio_minor_version_offset."
         required: false
         default: ""
         type: string
       istio_minor_version_offset:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-istio-version.yml
```diff
@@ -20,12 +32,36 @@ jobs:
...
+      id: determine-istio-version-to-use
+      run: |
+        if [ -z "${{ github.event.inputs.istio_version }}" ]; then
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" && -n "${{ github.event.inputs.istio_minor_version_offset }}" ]]; then
+            OFFSET="${{ github.event.inputs.istio_minor_version_offset }}"
+            # Convert to absolute value - we want a positive offset, though some people might find it more intuitive to say "-1" offset for the previous version
+            OFFSET=$(( OFFSET < 0 ? -OFFSET : OFFSET ))
+          else
+            case "${{ github.event.schedule }}" in
+              "0 2 * * *") OFFSET=0 ;;
```
jshaughn, line 48: not sure we need to run the latest istio version nightly because we already run it on every PR. Not a big deal, though.

jshaughn (commented):
LG, one comment

.github/workflows/test-istio-version.yml
jmazzitelli: I removed the third cron schedule - no longer testing the latest version nightly

jshaughn (approved):
LG!

Issue #8033, opened 2025-01-03
[ambient] Error reading logs when user doesn't have permissions for ztunnel


josunect:
### Describe the bug
When a user doesn't have permissions on the ztunnel workload and selects the ztunnel options, there is error as the user doesn't have permissions. 

![image](https://github.com/user-attachments/assets/eac1e8a9-1b92-4363-8cb5-b06416c7a59d)

### Expected Behavior
If the user doesn't have permissions to see the ztunnel logs, it should be no able to see the ztunnel option. 

### What are the steps to reproduce this bug?

1. Install a environment with Kiali and Ambient and openId and create a role with access to just bookinfo.
`hack/setup-kind-in-ci.sh -a openid -ab true`
`kubectl delete clusterRoleBinding kiali-group-viewer`
`kubectl apply -f https://github.com/josunect/yamls/blob/main/roleBookinfo.yaml`
`kubectl apply -f https://raw.githubusercontent.com/josunect/yamls/refs/heads/main/roleBindingBookinfoGroup.yaml`

3. Go to a workload details
4. Click in ztunnel logs

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:** master
- **Istio version:** 1.24.2
- **Kubernetes impl:** 
- **Kubernetes version:** 
- **Other notable environmental factors:**

nrfox:
How is access to a workload with sidecar logs gated? If you have access to the namespace then you can see the logs for that workload?

At the very least we need to do the same here for ztunnel. Check for the user's namespace access. Ensure user can't see logs for workloads in namespaces where they don't have read access.

josunect:
> How is access to a workload with sidecar logs gated? If you have access to the namespace then you can see the logs for that workload?

Yes, that is correct. But, the sidecar is the the same pod than the one the user is checking.
> 
> At the very least we need to do the same here for ztunnel. Check for the user's namespace access. Ensure user can't see logs for workloads in namespaces where they don't have read access.

Right, so, in that case, if the user doesn't have access to ztunnel, we shouldn't be showing the [ ] ztunnel logs check? I mean, hide that check if the user doesn've have permissions in advance.

jshaughn:
> Right, so, in that case, if the user doesn't have access to ztunnel, we shouldn't be showing the [ ] ztunnel logs check? 

[...]

PR #8046, opened 2025-01-10
Use SA to read ztunnel logs

josunect:
### Describe the change

Use SA to read ztunnel logs. Then, they are filtered by the workload name/namespace, so just the access logs will be shown

### Steps to test the PR

Install Istio Ambient.
Use kiali token strategy. 
Create roles for a test sa just for the bookinfo: 
```
kubectl apply -f https://raw.githubusercontent.com/josunect/yamls/refs/heads/main/roleBookinfo.yaml
kubectl apply -f https://raw.githubusercontent.com/josunect/yamls/refs/heads/main/roleBindingBookinfo.yaml
```

ztunnel logs should be visible when login with a token for the test SA: 
image

### Automation testing

### Issue reference

Fixes https://github.com/kiali/kiali/issues/8033

### DIFF
--- a/business/layer.go
+++ b/business/layer.go
@@ -123,7 +123,7 @@ func newLayer(
 	temporaryLayer.RegistryStatus = RegistryStatusService{kialiCache: cache}
 	temporaryLayer.TLS = TLSService{discovery: discovery, userClients: userClients, kialiCache: cache, businessLayer: temporaryLayer}
 	temporaryLayer.Svc = SvcService{config: *conf, kialiCache: cache, businessLayer: temporaryLayer, prom: prom, userClients: userClients}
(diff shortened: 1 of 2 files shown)


jshaughn (commented):
Sorry, as a user that constantly uses `anonymous` login, can you give me the additional steps necessary to generate the service account and token.  I tried to do it but the token I generated gave me `Login unsuccessful: not enough privileges to login`.

josunect:
I've created the sa and then the token with:

```
kubectl create sa test
kubectl create token test
```
Hopefully this works!

jshaughn (approved):
Looks good, I can see the `ztunnel` logs for `productpage`, but I have no access to the `istio-system` namespace on the namespace selector.

PR #8055, opened 2025-01-13
be able to configure /api endpoint to require user authentication

jmazzitelli:
When auth.strategy is not anonymous, users can still see the /api endpoint version and config information. This new config allows users to disable that unauthenticated access, requiring users to authenticate with the server in order to see the /api info.

```yaml
server:
  require_auth: true
```

operator PR: https://github.com/kiali/kiali-operator/pull/865

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -178,6 +178,7 @@ type Server struct {
 	Port                       int           `yaml:",omitempty"`
 	Profiler                   Profiler      `yaml:"profiler,omitempty"`
 	StaticContentRootDirectory string        `yaml:"static_content_root_directory,omitempty"`
+	RequireAuth                bool          `yaml:"require_auth,omitempty"` // when true, unauthenticated access to api/ endpoint is not allowed
 	WebFQDN                    string        `yaml:"web_fqdn,omitempty"`
 	WebPort                    string        `yaml:"web_port,omitempty"`
 	WebRoot                    string        `yaml:"web_root,omitempty"`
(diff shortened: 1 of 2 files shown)


jmazzitelli:
how to test (note: when I say, "Check Kiali URL" in the steps below, you likely will need to create a port-forward to the kiali server - you can use the hack script `hack/kiali-port-forward.sh` to do this):

1. You need the operator PR, too: https://github.com/kiali/kiali-operator/pull/865
2. Install a k8s cluster with Istio - I will assume you are on minikube - adjust the next steps as appropriate to your cluster
3. Build, push, deploy the operator and server: `make -e CLUSTER_TYPE=minikube build build-ui test cluster-push operator-create kiali-create`
4. Check Kiali URL http://localhost:20001/kiali/api - you should be able to see it because by default auth.strategy is anonymous
5. Edit the Kiali CR and set `spec.auth.strategy` to `token`, then wait for server pod to restart 
6. Check Kiali URL http://localhost:20001/kiali/api - you should be able to see it because by default server.require_auth is false, so even though auth.strategy is token you can still see this info as an unauthenticated user.
7. Edit the Kiali CR and set `spec.server.require_auth` to `true`, then wait for server pod to restart 
8. Check Kiali URL http://localhost:20001/kiali/api - you should not be able to see it; you should get an "unauthenticated" message because auth.strategy is not anonymous and you told the server to require authentication for this endpoint.
9. Create a token for the Kiali SA: `kubectl -n istio-system create token kiali-service-account` and copy the long token string that is output onto the clipboard (you will paste it into the login screen on the next step)
10. Go to the Kiali login page (http://localhost:20001/kiali) and see that it asks you for a token to login. Paste that long token string you got in the last step and see that you can successfully log in.
11. Check Kiali URL http://localhost:20001/kiali/api - you should be able to see it now that you are authenticated.

jshaughn (approved):
All good, checked /api and also ensured a typically secure endpoint remained secure.

Issue #8058, opened 2025-01-14
WorkloadGroup Validations.


hhovsepy:
WorkloadGroup needs to have validations such as:

- [x] ServiceAccount existence
- [ ] Labels duplicated in other WG

PR #8062, opened 2025-01-14
Added Validations for WorkloadGroups, fixed issue with references.

hhovsepy:
### Describe the change

Added 2 new Validation Warning messages for Workload Groups:
"Warning KIA1701 Service Account not found in this namespace".
"Warning KIA1702 More than one Workload Group with duplicate labels found in the same namespace" 
Kiali does not list ServiceAccounts directly from the system but is listing them from the Pods.
That is why the notification severity is Warning not a Error.

### Steps to test the PR

Create a WorkloadGroup with non existing Service Account, or with Service Account from other namespace, and with labels which exist in another WG in the same namespace.

```
kind: WorkloadGroup
metadata:
  name: ratings-vm
  namespace: bookinfo
spec:
  metadata:
    labels:
      app: ratings-vm
  template:
    serviceAccount: bookinfo-ratings
---
kind: WorkloadEntry
metadata:
  name: ratings-vm
spec:
  address: 2.2.2.2
  labels:
    class: vm
    app: ratings-vm
    version: v3
  serviceAccount: bookinfo-ratings
  network: vm-us-east
---
kind: WorkloadGroup
metadata:
  name: ratings-vm2
  namespace: bookinfo
spec:
  metadata:
    labels:
      app: ratings-vm
  template:
    serviceAccount: wrong
---
kind: WorkloadEntry
metadata:
  name: ratings-vm2
  namespace: bookinfo
spec:
  address: 2.2.2.2
  labels:
    class: vm2
    app: ratings-vm2
    version: v4
  serviceAccount: bookinfo-ratings
  network: vm-us-east
---
kind: WorkloadGroup
metadata:
  name: ratings-vm-no-entry
  namespace: bookinfo
spec:
  template:
    serviceAccount: bookinfo-ratings
```
Screenshot From 2025-01-20 12-38-35

### Automation testing

Added unit tests.

### Issue reference
https://github.com/kiali/kiali/issues/8058
Docs: https://github.com/kiali/kiali.io/pull/853

### DIFF
--- a/business/checkers/authorization/principals_checker_test.go
+++ b/business/checkers/authorization/principals_checker_test.go
@@ -56,10 +56,10 @@ func TestRegexPrincipalNotFound(t *testing.T) {
 	assert.NotEmpty(vals)
 	assert.Len(vals, 2)
 	assert.Equal(models.ErrorSeverity, vals[0].Severity)
-	assert.Error(validations.ConfirmIstioCheckMessage("authorizationpolicy.nodest.principalnotfound", vals[0]))
+	assert.NoError(validations.ConfirmIstioCheckMessage("authorizationpolicy.source.principalnotfound", vals[0]))
 	assert.Equal("spec/rules[0]/from[0]/source/principals[0]", vals[0].Path)
 	assert.Equal(models.ErrorSeverity, vals[1].Severity)
-	assert.Error(validations.ConfirmIstioCheckMessage("authorizationpolicy.nodest.principalnotfound", vals[1]))
+	assert.NoError(validations.ConfirmIstioCheckMessage("authorizationpolicy.source.principalnotfound", vals[1]))
 	assert.Equal("spec/rules[0]/from[0]/source/principals[1]", vals[1].Path)
 }
 
@@ -91,10 +91,10 @@ func TestNotPresentServiceAccount(t *testing.T) {
 	assert.NotEmpty(vals)
 	assert.Len(vals, 2)
 	assert.Equal(models.ErrorSeverity, vals[0].Severity)
-	assert.Error(validations.ConfirmIstioCheckMessage("authorizationpolicy.nodest.principalnotfound", vals[0]))
+	assert.NoError(validations.ConfirmIstioCheckMessage("authorizationpolicy.source.principalnotfound", vals[0]))
 	assert.Equal("spec/rules[0]/from[0]/source/principals[0]", vals[0].Path)
 	assert.Equal(models.ErrorSeverity, vals[1].Severity)
-	assert.Error(validations.ConfirmIstioCheckMessage("authorizationpolicy.nodest.principalnotfound", vals[1]))
+	assert.NoError(validations.ConfirmIstioCheckMessage("authorizationpolicy.source.principalnotfound", vals[1]))
 	assert.Equal("spec/rules[0]/from[0]/source/principals[1]", vals[1].Path)
 }
 
@@ -112,7 +112,7 @@ func TestRemoteClusterServiceAccount(t *testing.T) {
 	assert.NotEmpty(vals)
 	assert.Len(vals, 1)
 	assert.Equal(models.WarningSeverity, vals[0].Severity)
-	assert.Error(validations.ConfirmIstioCheckMessage("authorizationpolicy.nodest.principalremote", vals[0]))
+	assert.NoError(validations.ConfirmIstioCheckMessage("authorizationpolicy.source.principalremote", vals[0]))
 	assert.Equal("spec/rules[0]/from[0]/source/principals[0]", vals[0].Path)
 }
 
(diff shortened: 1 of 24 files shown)


business/references/workloadentry_references.go
```diff
@@ -34,11 +34,11 @@ func (n WorkloadEntryReferences) getConfigReferences(we *networking_v1.WorkloadE
 	}
 	weLabelSet := labels.Set(we.Spec.Labels)
 	for _, wg := range n.WorkloadGroups {
-		if wg.Spec.Template == nil || wg.Spec.Template.Labels == nil {
+		if wg.Spec.Metadata == nil || wg.Spec.Metadata.Labels == nil {
```
ferhoyos, line 37: According to https://istio.io/latest/docs/reference/config/networking/workload-entry/, labels of `WorkloadEntry` are set in the field `spec.Labels`, not in `spec.Metadata.Labels`

business/workloads.go
```diff
@@ -942,7 +942,7 @@ func (in *WorkloadService) fetchWorkloadsFromCluster(ctx context.Context, cluste
 		defer wg.Done()
 		var err error
 		if in.isWorkloadIncluded(kubernetes.WorkloadGroupType) {
-			wgroups, err = kubeCache.GetWorkloadGroups(namespace, labelSelector)
+			wgroups, err = kubeCache.GetWorkloadGroups(namespace, "")
```
ferhoyos, line 945: Why don't you need labelSelector here?

business/checkers/authorization/principals_checker_test.go
ferhoyos, line 1: Are these changes related to workload groups?

ferhoyos (changes requested):
Nice job @hhovsepy, a pair of validations well tested with multiple scenarios. I have been playing a bit with the WorkloadGroup definitions and the validations seem to work as expected :slightly_smiling_face:.

I think there is a small mismatch in the workload entry CR.

Not related to this PR, I have a pair of suggestions about Workload Group support in Kiali:
- It is a bit confusing that all Workload Groups status are `Not Ready` because they don't have pods. I assume that we are limited and we cannot know if they are ready, but I think it is better to show an `Unknown` status instead of `Not Ready`.

image

The same in the Workload group details page:
image


- Since only the overview tab of a Workload Group shows any information, maybe it would be better to hide the rest of the tabs and just show the overview one. Another option is to show a message in all the tabs that the feature is not available in workload groups (like we already do for logs):

image

business/workloads.go
hhovsepy: will add a comment in code,
so WorkloadGroups do not have labels, but labels are in Spec.Metadata, so filtering here will bring always empty result

business/checkers/authorization/principals_checker_test.go
hhovsepy: not directly, just there was a issue in code which started to appear after workload group debugging

business/references/workloadentry_references.go
hhovsepy: yes for WorkloadEntry it takes `weLabelSet := labels.Set(we.Spec.Labels)` , but work WorkloadGroup it takes `wg.Spec.Metadata.Labels`, unless I do not see what is the problem in code

ferhoyos: oh yes, you are right, I saw the workload entry reference above and I did not notice that you are looping through the workload groups.

ferhoyos (dismissed):
Thanks Hayk for the clarifications, now the PR is ready to be merged.

business/workloads.go
hhovsepy: you are right, added label back and included the search in kubeCache

hhovsepy:
@ferhoyos tabs will be hidden in App details page as well:
Screenshot From 2025-01-29 12-30-32

ferhoyos:
Thanks, Hayk, for the changes! Now it looks much better. 

The issue I see is that the application details page doesn’t explain why the information is limited (e.g., there is no workload group badge anywhere). Additionally, there is no application type like in the workloads, which makes the message somewhat confusing.

My two cents:
- Change the message to something like: "Kiali provides limited information about this application because it contains a non-Kubernetes workload."
- Add the WG or a similar badge to the workload to indicate which one is non-Kubernetes.

hhovsepy:
Thanks @ferhoyos, changed the message, but the 'W' badge is remained used since we don't display 'WG' or other workload types anywhere.
Screenshot From 2025-01-30 08-14-05

ferhoyos (approved):
I didn't mean to substitue the `Workload` icon by the `Workload group` one, just adding an icon to the right indicating that it is a workload group or another icon that indicates that it is an unsupported workload.

image

Anyway, reading the new message it is quite clear that the limited information provided is due to the workload type. Approving the PR.

Issue #8064, opened 2025-01-15
integration test flake - jq error - Cannot iterate over null


jmazzitelli:
Running integration tests on Istio 1.23, this failure occurs (see [CI action output here](https://github.com/kiali/kiali/actions/runs/12780433251/job/35626845775#step:8:815))
(though I don't think Istio version matters - I see this [in the 1.22 run also](https://github.com/kiali/kiali/actions/runs/12761060680/job/35567625600#step:8:759))
```
[INFO] Kiali is ready.
[INFO] Waiting for Kiali server to respond externally to health checks at http://172.18.255.72/kiali
[INFO] Kiali server is healthy
jq: error (at <stdin>:1): Cannot iterate over null (null)
Error: Process completed with exit code 5.
```

I can't figure out where this error happens, because we call so many different hack scripts and functions and the output doesn't make it easy to figure out. But `[INFO] Kiali is ready.` is coming from the end of an invocation of `hack/setup-kind-in-ci.sh`. The next line that it output (`[INFO] Waiting for Kiali server to respond externally to health checks...`) is coming from `hack/run-integration-tests.sh` function `ensureKialiServerReady()`. That function outputs `[INFO] Kiali server is healthy`.

So we need to find the place where `hack/setup-kind-in-ci.sh` is executed, followed by a call to `ensureKialiServerReady()` followed by a call to `jq` which is where the error is occurring. I'm not 100% sure, but I think we need to look [here](https://github.com/kiali/kiali/blob/v2.4.0/hack/run-integration-tests.sh#L405), where I think the bug is happening [here](https://github.com/kiali/kiali/blob/v2.4.0/hack/run-integration-tests.sh#L286) in `hack/run-integration-tests.sh` function `ensureMulticlusterApplicationsAreHealthy()` when setting `has_http_200` variable:
```
has_http_200=$(echo "$response" | jq '[.applications[] | select(.name=="reviews" and .cluster=="west" and .health.requests.inbound.http."200" > 0)] | length > 0')
```

jmazzitelli:
Yes, I think that is where the problem happens. A quick test shows that if the Kiali URL response doesn't have `applications[]`, then jq will produce the error - e.g. if I ask jq to look for "foo" (which is missing in the response), then the same error occurs:

```
$ has_http_200=$(echo "$response" | jq '[.foo[] | select(.name=="reviews" and .cluster=="west" and .health.requests.inbound.http."200" > 0)] | length > 0')
[...]

PR #8065, opened 2025-01-15
[hack] integration test should not fail on missing applications list

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8064


jshaughn (approved):
script magic

Issue #8075, opened 2025-01-23
Show Istio configurations for K8sGateways


josunect:
### What do you want to improve?
Show the K8s gateways in the istio config for the workloads. 

### What is the current behavior?
At the moment, the k8s gateways are not show. 
This is an example for waypoint proxies: 

![Image](https://github.com/user-attachments/assets/bd3c8cbe-5189-4d0d-a1e8-26fc3be6c4a4)

And the Istio configuration: 

![Image](https://github.com/user-attachments/assets/5f10d745-883f-428d-823f-4935787ae5e1)

### What is the new behavior?
Show the Istio configuration for the k8s gateways

PR #8077, opened 2025-01-23
Added support of listing K8sGateways for Workloads.

hhovsepy:
### Describe the change

In K8sGateway Workload details page, list K8s Gateway in Istio config:
Screenshot From 2025-01-23 15-25-24

### Steps to test the PR

Setup ambient environment.
Or in K8s Gateway setup environment create a K8sGateway which will cause Workload generated.
Go to 'waypoint' or any K8s Gateway Workload details.
You will see the Istio Config list not empty, but will show those objects which name matches the `gateway.networking.k8s.io/gateway-name` label of workload.

### Automation testing

### Issue reference
https://github.com/kiali/kiali/issues/8075

### DIFF
--- a/business/apps.go
+++ b/business/apps.go
@@ -162,8 +162,7 @@ func (in *AppService) GetClusterAppList(ctx context.Context, criteria AppCriteri
 		for _, wrk := range valueApp.Workloads {
 			joinMap(applabels, wrk.Labels)
 			if criteria.IncludeIstioResources {
-				wSelector := labels.Set(wrk.Labels).AsSelector().String()
-				wkdReferences = append(wkdReferences, FilterWorkloadReferences(wSelector, *istioConfigList, cluster)...)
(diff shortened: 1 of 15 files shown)


josunect (commented):
Nice improvement, thanks @hhovsepy ! 

I think it works as expected but I have a couple of questions:

- I can see the istio config from the workload view: 
image
but not from the service: 
image

For the contrary, in the productpage is the opposite: 
image

In the service: 
image

- The other question is, shouldn't the workload be added as a reference? Or this is just for other Istio configs? I think it would be useful to have it linked also here.
image

I think we should include waypoint tests in the cy, but can be done in a separate PR.

hhovsepy:
Thank you @josunect for the review, right there are still many inconsistency in configs and links, I will resolve them within this PR.

josunect (approved):
LG!

Issue #8081, opened 2025-01-24
Error: namespace not found in Multi Cluster


josunect:
### Describe the bug

In Kiali multi-cluster, when there is a namespace into one of the cluster that does not exist in the remote, and we select this namespace in the traffic graph, it creates a 500 error turning into a "NS not found". This call is done by the validations:

In this example, travels was created just in the remote. When it is selected in the ns selector from the traffic map, there is an error: 
![Image](https://github.com/user-attachments/assets/c0e1928f-3f3a-497c-bc7f-aa6947f05341)

`https://kiali-istio-system.apps.user-rhos-01-05.servicemesh.rhqeaws.com/api/istio/validations?namespaces=travel-agency,travel-control,travel-portal,sample,bookinfo,default,istio-cni,istio-system&clusterName=east`

![Image](https://github.com/user-attachments/assets/c89c5122-0b91-47cf-bd76-5e57844052a0)

### Expected Behavior
Validations should be done for the namespaces that exist on that cluster (Or no error should be returned)

### What are the steps to reproduce this bug?

1. Install istio and kiali in multi-primary (This is probable happening in primary remote)
2. Create a namepsace in the west cluster 
3. Go to the traffic graph 
4. Select all the namespaces

### Environment
*Learn about how to determine versions [here](https://kiali.io/docs/faq/general/#how-do-i-determine-what-version-i-am-running).*

- **Kiali version:**  2.1.1
- **Istio version:** OSSM 3 TP2
- **Kubernetes impl:** 
- **Kubernetes version:** 
- **Other notable environmental factors:**

nrfox:
Needs backport to 2.4

PR #8082, opened 2025-01-24
Added filter namespaces per cluster in Graph Validations

hhovsepy:
### Describe the change

In multicluster installation, in Graph pages when all namespaces are selected, Kiali UI sends `validations` request to backend per cluster with all namespaces sent as a param.
So those namespaces which exist in 'east' and are requested for 'west', cause 500 errors in response.
Fixed the frontend by adding filter namespaces per cluster in Graph Validations queries.

### Steps to test the PR

1.  Install istio and kiali in multi-primary (This is probable happening in primary remote)
2.  Create a namepsace in the west cluster
3.  Go to the traffic graph
4.  Select all the namespaces

### Automation testing

### Issue reference
https://github.com/kiali/kiali/issues/8081

### DIFF
--- a/frontend/src/utils/Common.ts
+++ b/frontend/src/utils/Common.ts
@@ -17,12 +17,21 @@ export const arrayEquals = <T>(a1: T[], a2: T[], comparator: (v1: T, v2: T) => b
 export const namespaceEquals = (ns1: Namespace[], ns2: Namespace[]): boolean =>
   arrayEquals(ns1, ns2, (n1, n2) => n1.name === n2.name);
 
-export const namespacesPerCluster = (
+/**
+ * Returns a list of namespace names that are both active and belong to the specified cluster.
+ * If the full list of namespaces is not provided, return the names of active namespaces directly.
+ *
+ * @param activeNss - List of currently active namespaces.
+ * @param allNss - Full list of namespaces (optional). This is needed to figure out namespaces per cluster.
+ * @param cluster - Target cluster to filter namespaces by.
+ */
+export const namespacesForCluster = (
   activeNss: Namespace[],
   allNss: Namespace[] | undefined,
   cluster: string
 ): string[] => {
   const activeNames = new Set(activeNss.map(ns => ns.name));
+  // If allNss is not provided, return the names of active namespaces directly
   if (!allNss) {
     return activeNss.map(ns => ns.name);
   }
--- a/frontend/src/utils/Common.ts
+++ b/frontend/src/utils/Common.ts
@@ -17,6 +17,22 @@ export const arrayEquals = <T>(a1: T[], a2: T[], comparator: (v1: T, v2: T) => b
 export const namespaceEquals = (ns1: Namespace[], ns2: Namespace[]): boolean =>
   arrayEquals(ns1, ns2, (n1, n2) => n1.name === n2.name);
 
+export const namespacesPerCluster = (
+  activeNss: Namespace[],
+  allNss: Namespace[] | undefined,
+  cluster: string
+): string[] => {
+  const activeNames = new Set(activeNss.map(ns => ns.name));
+  if (!allNss) {
+    return activeNss.map(ns => ns.name);
+  }
+
+  // filter allNss for namespaces that match the selected active names and the cluster
+  return removeDuplicatesArray(
+    allNss.filter(ns => activeNames.has(ns.name) && ns.cluster === cluster).map(ns => ns.name)
+  );
+};
+
 export function groupBy<T>(items: T[], key: keyof T): { [key: string]: T[] } {
   return items.reduce(
     (result, item) => ({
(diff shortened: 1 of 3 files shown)


handlers/namespaces.go
```diff
@@ -126,13 +126,11 @@ func ConfigValidationSummary(w http.ResponseWriter, r *http.Request) {
...
 	validationSummaries := []models.IstioValidationSummary{}
 	for _, ns := range nss {
+		// for multicluster validation summaries, all selected namespaces are sent in a request per all clusters, so ignore not found namespaces
 		istioConfigValidationResults, errValidations := business.Validations.GetValidationsForNamespace(r.Context(), cluster, ns)
-		if errValidations != nil {
-			log.Error(errValidations)
-			RespondWithError(w, http.StatusInternalServerError, errValidations.Error())
-			return
+		if errValidations == nil {
+			validationSummaries = append(validationSummaries, *istioConfigValidationResults.SummarizeValidation(ns, cluster))
```
nrfox, line 132: I don't think we should ignore _all_ errors here. Doesn't the problem stem from the frontend sending a bad request to the backend? The frontend is asking for validations for namespaces it shouldn't. Is it possible to fix that on the frontend? Namespaces are cluster specific and we should know which namespaces exist on which clusters on the frontend and only send those namespaces in with the query param.

hhovsepy: Then we should rework the frontend, the namespaces selection box, it is not filtering namespaces per cluster and showing all merged, and while selecting, it makes a list of namespaces strings.

hhovsepy: Which is a scope of another task https://github.com/kiali/kiali/issues/5932

nrfox: > In backend, anyway exiting the loop in case of error and returning that error was wrong.

Why was that wrong? If there's an issue getting validations shouldn't the backend return an error instead of silently fail? In this case, the frontend is giving the backend an invalid list of namespaces. Maybe we should treat authorization errors, where the user doesn't have access to a namespace that exists, as 404 so that we don't leak namespace names.

The backend can do an access check for the namespaces requested and filter out the ones that the user doesn't have access to. If the user provides namespaces that don't exist in the query param post filtering, the backend can either return a 400 or silently return results without those namespaces.

You're right, we should address the namespace selector separately in #5932 but I think even without the the frontend should have both 1. the list of selected namespaces 2. the list of namespaces per cluster. Before you send a request to the backend the frontend would need to filter out 2. with 1. instead of just using the list of selected namespaces and sending a request per cluster.

jshaughn: Looking at the requests in Josune's screenshot, you can see that the client is asking for all namespaces for each cluster. I agree with Nick, that's wrong and should cause an authorization error in the backend.  The client should have full AccessibleNamespace info for the user, and should be able to narrow the namespaces as needed. There are lots of places in the UI code, like in the graph appenders, that need to protect against this sort of back-end violation. Looking at this, I think it's a client-side fix.

hhovsepy: Right, thanks for the comments, have fixed the frontend to filter selected namespaces per cluster before sending requests.

frontend/src/utils/Common.ts
```diff
@@ -17,6 +17,22 @@ export const arrayEquals = <T>(a1: T[], a2: T[], comparator: (v1: T, v2: T) => b
 export const namespaceEquals = (ns1: Namespace[], ns2: Namespace[]): boolean =>
   arrayEquals(ns1, ns2, (n1, n2) => n1.name === n2.name);
 
+export const namespacesPerCluster = (
```
jshaughn, line 20:
```suggestion
// namespacesPerCluster takes an array of <namespaceNames> and an array of <namespaces> (with cluster info) 
// and returns the subset of namespaceNames that are found on the target cluster. If <namespaces> is
// undefined all <namespaceNames> are returned.
export const namespacesPerCluster = (
```

Minimally, I think this utility could use a comment, because it's not really clear what it's doing or what it expects.  I think also, `namespacesForCluster` would be better that `namespacesPerCluster`.

But maybe more than that, it may be over-complicated because I think you could use `store.getState().namespaces.namespacesPerCluster` to immediately get the namespaces for the target cluster, and then filter them by the activeNamespace names.

hhovsepy: Have renamed the method.
However the `store.getState().namespaces.namespacesPerCluster` is empty, as this variable is set outside of the graph. But let me check if I can fix that and reuse.

josunect (commented):
Validated in primary-remote: 
image

frontend/src/utils/Common.ts
jshaughn: hmm, I thought that was set at login.  Note that I think JSON.stringify of a map just shows `{}`, even if the map has data.  Anyway, if I'm wrong then what you have should be fine.

PR #8083, opened 2025-01-24
Add waypoints and gateways to mesh page

### LINKED ISSUE #8054: [Mesh Page] visualize waypoints
Part-of https://github.com/kiali/kiali/issues/7826

Istio ambient users need to have insight into their Waypoint deployments. In addition to workload detail, it could be good for users to be able to visualize this infrastructure in the Mesh Page.  Note that Waypoint proxies are actually Gateways, it may make sense to expand this to visualize all gateways in the Mesh page.

jshaughn:
Closes #8054

This PR adds gateway and waypoint support to the Mesh page. There is a new Display menu offering these two options, both off by default.

The additional nodes made the Mesh side panel look unorganized, so the PR also improves the look and organization of the Mesh side panel. It adds analogous support to Cluster and Namespace side panels, showing the infra nodes narrowed to that box.

The waypoint and gateway nodes have connecting edges to the control plane with matching revision. By default, neither are shown.

@ferhoyos , @nrfox , thanks for your review feedback. If you'd like to review the PR please add yourself as a reviewer.


josunect (commented):
Validated: 

image

A couple of things that I've noticed in the summary panel graph of the Mesh:
- The api version for the waypoints is unknown
- In this example, there are 2 waypoints with the default name (waypoint) in different namespaces, but is not possible to know where. Maybe it would be possible to add the information here?

Multi cluster: 
image

image

I think that, in multi cluster, it would be also nice to have the cluster to make a difference between the resources that are named the same: 

image

(But those are minor comments that can be added in iterations)

jshaughn:
> The api version for the waypoints is unknown

I'm removing `api version` for Waypoints.  We only have the waypoint workload at this point, not the actual Waypoint config.  To get the api version for the Waypoint definition you need to navigate to the workload detail, via the link, and then to the linked config.

frontend/src/pages/Mesh/target/TargetPanelCommon.tsx
```diff
@@ -184,6 +184,221 @@ export const renderNodeHeader = (
...
+});
+
+export const renderNodeLink = (meshData: MeshNodeData, style?: string): React.ReactNode | undefined => {
+  let displayName, key, link: string;
+  let pfBadge: PFBadgeType;
+
+  switch (meshData.infraType) {
+    case MeshInfraType.GATEWAY:
+      const name = `${meshData.infraName}-${meshData.infraData?.spec?.gatewayClassName}`;
+      link = `/namespaces/${encodeURIComponent(meshData.namespace)}/workloads/${encodeURIComponent(name)}`;
```
josunect, line 221: If the gatewayClassName is undefined, it could lead to an incorrect link. This could happen if you create a Gateway instead of a k8s Gateway:
image
Probably could be better removing the last part of the name when it is undefined?

jshaughn: good catch, I'll make the change.

jshaughn:
> In this example, there are 2 waypoints with the default name (waypoint) in different namespaces, but is not possible to know where [from the summary]. Maybe it would be possible to add the information here?

I played with this a little bit, but for now I'm going to leave it ambiguous. I want the summary to be relatively short and clean, it's more about the numbers and the user can use the topology, as well as find/hide to zoom in and get details on a particular infra object.  Also, it's not totally clear to me how often the infra items will have the same name. We can revisit later, or in response to a user RFE.

josunect:
With revisions: 
image

josunect (approved):
LGTM!

jmazzitelli (approved):
fine

PR #8088, opened 2025-01-28
Use sail in frontend tests

nrfox:
### Describe the change

Use sail in frontend test setup.

### Steps to test the PR

Automated tests pass.

### Automation testing

Included updates to automation.

### Issue reference

Part of #7830

### DIFF
--- a/hack/istio/install-testing-demos.sh
+++ b/hack/istio/install-testing-demos.sh
@@ -125,6 +125,9 @@ if [ "${DELETE_DEMOS}" != "true" ]; then
   if [ "${USE_GATEWAY_API}" == "true" ]; then
     gateway_yaml="${ISTIO_DIR}/samples/bookinfo/gateway-api/bookinfo-gateway.yaml"
   elif ! [ -z "$GATEWAY_HOST" ]; then
+    # TODO: Ideally this wouldn't go in istio-system. It'd either go in the same namespace as the app
+    # or in its own namespace but doing that would require rewriting some tests and this is faster...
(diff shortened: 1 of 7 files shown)


hack/istio/install-testing-demos.sh
```diff
@@ -125,6 +125,9 @@ if [ "${DELETE_DEMOS}" != "true" ]; then
   if [ "${USE_GATEWAY_API}" == "true" ]; then
     gateway_yaml="${ISTIO_DIR}/samples/bookinfo/gateway-api/bookinfo-gateway.yaml"
   elif ! [ -z "$GATEWAY_HOST" ]; then
+    # TODO: Ideally this wouldn't go in istio-system. It'd either go in the same namespace as the app
```
mkralik3, line 128: Would it be a lot of work to refactor cypress test suite? ( only namespace and gateway name will change so ideally only names should be changed. or is there anything else ? )

nrfox: We discussed this offline but some of the frontend tests do rely on things like a certain number of objects in a table for a given namespace, the actual `Gateway` and `VirtualService` resource being present, etc. so it's probably more work than doing a find/replace.

mkralik3: yes :+1:  , we can try to move istio-gateway to bookinfo namespace as a part of different namespace

ScriptingShrimp (approved):
LGTM! Shall we backport to 2.4?

nrfox:
> Shall we backport to 2.4?

Yes thanks for the reminder

PR #8114, opened 2025-02-03
[cypress] Update mesh cypress testing to be flexible

jshaughn:
Update mesh cypress testing to be flexible given the tracing test deployment, which may be jaeger or tempo, and may be deployed in the CP namespace, its own namespace, or externally.

Note that our cypress tests do require one of the two to be deployed, it will not pass in an env with no tracing configured.

This fixes https://issues.redhat.com/browse/OSSM-8763

Testing
Upstream CI testing should handle the jaeger deployment option.  This should also be tested in an environment using Tempo for tracing.

### DIFF
--- a/frontend/cypress/integration/common/mesh.ts
+++ b/frontend/cypress/integration/common/mesh.ts
@@ -45,7 +45,7 @@ When('user selects mesh node with label {string}', (label: string) => {
       assert.isTrue(controller.hasGraph());
 
       const { nodes } = elems(controller);
-      const node = nodes.find(n => n.getLabel() === label);
+      const node = nodes.find(n => n.getLabel().toLowerCase() === label.toLowerCase());
       assert.exists(node);
 
       const setSelectedIds = state.meshRefs.setSelectedIds as (values: string[]) => void;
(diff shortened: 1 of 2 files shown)


jshaughn:
This should be backported to Kiali v2.4.

frontend/cypress/integration/common/mesh.ts
```diff
@@ -53,6 +53,26 @@ When('user selects mesh node with label {string}', (label: string) => {
...
+    .should('have.length', 1)
+    .then($graph => {
+      const { state } = $graph[0];
+
+      const controller = state.meshRefs.getController() as Visualization;
+      assert.isTrue(controller.hasGraph());
+
+      const { nodes } = elems(controller);
+      const node = nodes.find(n => n.getLabel() === 'jaeger' || n.getLabel() === 'Tempo');
+      assert.exists(node);
```
ScriptingShrimp, line 69: `assertexpected undefined to exist` scenario fails here with OCP + sail setup. Will DM you details.

ScriptingShrimp (changes requested):

frontend/cypress/integration/common/mesh.ts
jshaughn: Pushed a fix [I think]

ScriptingShrimp (approved):
Passing on OCP in upstream. 
Also passing on OSSM3.0, we can cherrypick fix there.

LGTM!

Issue #8084, opened 2025-01-27
Unlike other resources, DeploymentConfig (and some others) is not cached causing confusing behavior to the user


jmazzitelli:
Unlike other resources, DeploymentConfig is not cached: https://github.com/kiali/kiali/blob/f4b8190355d3abe7f798698ff0eded71125eeec9/business/workloads.go#L874-L879.

We use the userclient to fetch those objects directly from the API every time we need them which means if the user had LIST for all namespaces but no GET permissions for DeploymentConfig in that namespace, the request would fail. This is unlike other resource access Kiali does and results in weird behavior that is confusing to the user.

PR: https://github.com/kiali/kiali/pull/8118

jshaughn:
I wonder if this is true for all of the workloads we exclude by default:
- "CronJob"
- "DeploymentConfig"
- "Job"
- "ReplicationController"

nrfox:
> I wonder if this is true for all of the workloads we exclude by default:
> 
>     * "CronJob"
> 
>     * "DeploymentConfig"
> 
>     * "Job"
> 
>     * "ReplicationController"

Probably... look for `userclient.Get<ResourceName>` in `workloads.go`. iirc there should be a comment on them like: ` // No cache for DeploymentConfigs`.

nrfox:
> What's different, we use the kiali perms for cached config?

iirc we'd need to introduce a new library and refactor how we handle the openshift resources and we just haven't done that yet.

> Anyway, any reason we don't want to cache these if they are configured to be included? Going one step further, given our caching, do we still need to exclude workload types? I know we did this originally for perf reasons.

At least `Job` is an ephemeral resource so you might want to have the ability to exclude those.

jmazzitelli:
Here's how to fix this (I tried it and it seems to work). The question is - is this what we really want? What is the reason why we do not cache these objects that then require us to obtain the objects using the user client?

```diff
diff --git a/business/workloads.go b/business/workloads.go
index 1f44e584b..a067f5221 100644
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -815,7 +815,8 @@ func (in *WorkloadService) fetchWorkloadsFromCluster(ctx context.Context, cluste
 		return nil, err
 	}
 
-	userClient, ok := in.userClients[cluster]
+	// we've already established the user has access to the namespace; use SA client to obtain namespace resource info
+	client, ok := in.kialiSAClients[cluster]
 	if !ok {
[...]

PR #8118, opened 2025-02-04
some resources are not in the cache - use SA client to get them

jmazzitelli:
fixes: https://github.com/kiali/kiali/issues/8084

Right now, the original code does this:

1. Check if user has access to namespace
2. Get resources from cache if we know the resources are cached (e.g. NOT DeploymentConfigs)
3. Get resources directly from k8s API using the USER client if we know the resources are NOT cached (e.g. DeploymentConfigs)

This PR changes 3 so the code now does this: "get resources directly from k8s API using the <ins>*SA client*</ins> if we know the resources are NOT cached"

The original code is wrong because all the other resources are obtained from the cache, and those resources in the cache were obtained via the SA client; but yet for these potentially excluded workload resources (such as DeploymentConfigs and CronJobs which are not cached), the code was using the USER client to obtain them from the k8s API. This produced confusing errors for the user and goes against the namespace access [policy](https://kiali.io/docs/configuration/rbac/#introduction) of Kiali.

### DIFF
--- a/business/workloads.go
+++ b/business/workloads.go
@@ -817,7 +817,8 @@ func (in *WorkloadService) fetchWorkloadsFromCluster(ctx context.Context, cluste
 		return nil, err
 	}
 
-	userClient, ok := in.userClients[cluster]
+	// we've already established the user has access to the namespace; use SA client to obtain namespace resource info
+	client, ok := in.kialiSAClients[cluster]
 	if !ok {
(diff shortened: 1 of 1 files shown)


jmazzitelli:
If you were to do all those test steps with the current release, at step 14 you will see this on the graph page (and a similar error on the overview page):
image

hhovsepy:
Checking on a OCP cluster meanwhile.

nrfox (approved):

Issue #8007, opened 2024-12-17
Kiali uses a lot of CPU (in validations)


rolandkool:
### Describe the bug

On a cluster with around 1600 namespaces, but a discovery selector that only selects around 380 of them and cluster wide view set to true, the CPU usage of the Kiali pod is permanently around 2500m. Although I have no reference to compare it with, this seems a bit excessive. This is even with any clients interacting with the UI.

![image](https://github.com/user-attachments/assets/7acaf779-3d2e-4291-8fa0-9586d63a27ab)

The graph shows a drop in usage. With cluster wide view disabled, CPU load is around 4000m.

### Expected Behavior

I'd like to understand if this a reasonable amount of CPU for our setup or if there is an issue with Kiali that causes it to use so much CPU.

### What are the steps to reproduce this bug?

Nothing in particular.
Cluster has around 1600 namespaces in total, but only 380 are selected with a discovery selector. Cluster wide view is enabled. Anonymous mode enabled.

### Environment

- **Kiali version:** 2.2.0
- **Istio version:** 1.23.2 asm 2
- **Kubernetes impl:** GKE
- **Kubernetes version:** 1.30
- **Other notable environmental factors:**

Issues/PRs for this Epic:
- [x] https://github.com/kiali/kiali/pull/8119
- [x] https://github.com/kiali/kiali/pull/8194
- [x] https://github.com/kiali/kiali/pull/8206
- [x] #8211 
- [x] #8210 
- [ ] [pre-compute validation keys for istio config objects](https://github.com/kiali/kiali/pull/8335)

rolandkool:
[profile.gz](https://github.com/user-attachments/files/18168277/profile.gz)

I've attached a 15 seconds CPU profile.

rolandkool:
Deployed another instance on a cluster with similar amount of total namespaces, but only 25 are included in the discovery selector. CPU usage is permanently around 600m.

jshaughn:
I think currently the Kiali Discovery Selector is gating the namespaces available to, or can be seen, by the user.  But I don't think it has the ability to reduce the number of k8s watches that are in place in cluster-wide, to manage caching. And so I think that is why you see more of a constant consumption based on the total namespaces.  

I think the workaround is to not use cluster-wide, despite the fact that it is often the more efficient option.  I think we will need to investigate whether we can apply Discovery Selectors to also minimize cache impact.

cc @jmazzitelli , @nrfox

rolandkool:
[...]

PR #8119, opened 2025-02-04
Make validations reconcile interval configurable

nrfox:
### Describe the change

Makes the reconcile interval (how often Kiali validates all the Istio resources) configurable.

### Steps to test the PR

Recommendations for how to test this PR. Reminder that each PR should also include a "Test" label.

### Automation testing

### Issue reference

Relates to #8007


hhovsepy:
This works fine locally, and right, at this step we should not disable the reconciliation at all as described in step1, until  the step2 with events is implemented.

config/config.go
```diff
@@ -748,6 +752,7 @@ func NewConfig() (c *Config) {
 				IstiodPollingIntervalSeconds:      20,
 				RootNamespace:                     "istio-system",
 				UrlServiceVersion:                 "",
+				ValidationReconcileInterval:       util.AsPtr(time.Minute),
```
hhovsepy, line 755: Wouldn't it better to accept the ValidationReconcileInterval in seconds, as in milliseconds it takes a big numbers and probably it will not be necessary configure that precise level. Like IstiodPollingIntervalSeconds we have in seconds.

hhovsepy: I see now, it accepts strings.

hhovsepy (approved):
Lgtm, tested locally, and the Warning message also works fine when process takes longer than the interval.

hhovsepy:
Run on OCP cluster via Jenkins pipeline, this test is failing after several attempts: `Config wizard.Create multiple K8s Gateways with colliding hostnames and port combinations and check for a reference. Then delete one of them and the reference should be gone.`
Kiali Istio Config wizard -- Create multiple K8s Gateways with colliding hostnames and port combinations and check for a reference  Then delete one of them and the reference should be gone (failed)

nrfox:
> Run on OCP cluster via Jenkins pipeline, this test is failing after several attempts: `Config wizard.Create multiple K8s Gateways with colliding hostnames and port combinations and check for a reference. Then delete one of them and the reference should be gone.`

This is because the default reconcile interval for validations is now `60s` whereas before it was `10s` so things like this will be more noticeable. For the CI pipeline in Github Actions I manually set the reconcile interval on the Kiali to something low to account for this in the tests: https://github.com/kiali/kiali/pull/8119/files#diff-ba865888ca30193e590825deb05adae9feb87778d5a3a72800ceb1e9d349804eR185

nrfox:
@hhovsepy in other words, however that Jenkins pipeline is deploying Kiali and running the tests against it, the Kiali CR will now need to include setting `validation_reconcile_interval` to a lower value than the default.

hhovsepy:
I think we need to modify that test to request the Istio config details page for "gatewayapi-1" object after deleting the conflict one, which will trigger force validations, then continue testing scenario.

jshaughn (approved):
Visually looks good.  If the reconcile takes longer than the interval it starts again immediately, right?  I guess it's ok because users can increase the interval as needed.  Is there an operator PR?

nrfox:
>  If the reconcile takes longer than the interval it starts again immediately, right?

Yes and if there's an error or conflict in the reconcile it will also start again but, unlike the periodic reconcile, if there's an error or conflict then there's an exponential backoff to the retries.

nrfox:
Merging since the test failure appears to be a known flake.

nrfox:
> Is there an operator PR?

No since the operator will pass through any config you set. We only need an operator PR if we want to update the default in the operator but there's already a default set in the server so IMO it's not necessary.

Will need a docs update though to document the config option and maybe an FAQ entry around improving the performance of Kiali.

jmazzitelli:
> We only need an operator PR if we want to update the default in the operator but there's already a default set in the server so IMO it's not necessary.

For almost every setting, we never rely on the server defaults. Yes, the server has defaults set, but in the past that was mostly just to support the unit tests. The operator defines the settings and the defaults for customers.

Unless there is a reason not to (other than, "config.go has a default set") then for consistency's sake we should define it in the operator and the CRD / example documentation.

Our checklist for adding a new server config option is here: https://github.com/kiali/kiali-operator/blob/master/DEVELOPING.adoc#are-you-altering-a-kiali-server-configuration-setting

jshaughn:
I think I'd prefer if we updated the operator for all but purposefully undocumented settings.  The default config map should include all public options.

nrfox:
> For almost every setting, we never rely on the server defaults.

Why is that? Otherwise we have to set the same default in 4 different places:

1. Operator
2. Helm charts
3. Server
4. Documentation

With there being some risk of skew between all of them if we update them one but not the others. There's also the risk that the unit tests are relying on default behavior defined in `config.go` but that is set to something else in the operator.

Having to update both 3. and 4. is unavoidable for now but at least keeping the source of truth in the server would prevent behavior differences between deployments.

I'm fine with updating the operator and helm... just asking why.

jmazzitelli:
I'll explain the thought process here.

> Why is that? Otherwise we have to set the same default in 4 different places:
> 
> 1. Operator

The[ defaults yaml](https://github.com/kiali/kiali-operator/blob/v2.5.0/roles/default/kiali-deploy/defaults/main.yml) is where you put it - if it is not needed anywhere else, that's the only operator place. So that isn't too much of a burden (though it is an extra burden to be sure).

> 2. Helm charts

In fact, you rarely have to do anything to the helm charts. Uness the new setting actually affects the deployment, service, or something of that nature, you do not touch anything in the helm charts (no helm chart PR needed). That's what [this comment](https://github.com/kiali/helm-charts/blob/v2.5.0/kiali-server/values.yaml#L11) means. But if the setting does change or configure something in the resources the helm charts lay down, then regardless of whether the server has it defined in config.go or not, and regardless of whether the operator has it defined or not, it would still have to go in the server helm chart anyway to get the server helm chart to deploy things correctly.

> 3. Server

The config.go defaults are really just set (at least this was the original intention) so that when you run unit tests or integration tests (where the server is not installed by the operator) things work and can pass. Now they kinda are a de facto source of truth (which they really shouldn't be because that causes problems for customer support and customer debugging, which I'll get to shortly). But they are really (or should be) considered just the "fallback" defaults if things are not set in the ConfigMap (which is where the operator puts everything).

> 4. Documentation

This has to be updated with the new setting/default value anyway, regardless of where the defaults are set (config.go or operator defaults yaml). These are in the operator repo (the CRD schema and example yamls) so you have to create a PR there anyway.

---

So really there is only that extra place (in the operator) where it might be said that it isn't needed. The others are always needed (when required - as in the server helm chart case, or when unit tests need them - as in the config.go case).

Now, the question is - why? Here's the thought process and why we've been doing this since very early on.

Num 1. If you do not set it in the operator defaults, that means the ConfigMap will not have it set. It will simply not exist in the ConfigMap. Now, take off your developer's hat, and put on your customer support engineer hat (or just put on your customer hat on). Something is broken in Kiali, and you do not know why. "_Hey, let me go check what my settings are - maybe something isn't right._" You look at the ConfigMap yaml - you see nothing about "reconcile interval" (or, pick your favorite setting). You have no idea (as a customer support rep, or a customer) what the default is - you can't expect them to go looking into the codebase to find it (they aren't going to know to look [here in config.go](https://github.com/kiali/kiali/blob/v2.5.0/config/config.go#L676)). That person might not even know that setting even exists! BUT! If that person sees "reconcile interval: 10s" , it might make a light bulb go off - "_huh.. I didn't know that setting existed.. I wonder what this does. That 10s seems fast, and reconciling validations is where I'm seeing a problem. Let me investigate this... maybe let me set this to 60s and see if it makes a difference_".

In addition, how much easier it is when we get a bug report we can just ask the customer, "_send me your ConfigMap yaml_" and you can immediately see all the values (even the default values) right there in one place. You don't have to remember a setting exists and you don't have to think to yourself, "_hmm.. that setting is missing so it should be taking the default value, but I can't remember what the default value is. Drat, now I have to go look in config.go and read the code to find out what the default is_".

In short, the point is - this isn't to make development easier, it is mainly to make customer support easier, and for making it easier for a customer to at least know what his full configuration is (and maybe even help them solve their problem on their own). "ConfigMap" should show all.

 Num 2. Here's an ancillary reason - Ansible easily breaks when trying to access variables that are not defined (Ansible will bomb with a failure and thus cause the Kiali CR reconciliation to repeat if a variable that is not defined is accessed). To set the value in the default yaml means you really never have to worry about adding `<setting> is defined` everywhere in the code. (Actually, you will find that in lots of places anyway, but that's only because the code is written to be paranoid; it checks these things even though that "is defined" should always evaluate to true.) But in the case where someone writes Ansible code in the operator and simply forgets that "is defined" check, making sure you define the default helps avoid unnecessary bugs cropping up in the code.

That's the long winded answer for why we do what we do here as it related to defaults. And, again, it really is only putting it in that one place where it technically isn't needed (the operator defaults.yaml). Those other 3 places, would need to be updated regardless. So.. 3 places, 4 places.. what's the difference? :)

ferhoyos:
> @hhovsepy in other words, however that Jenkins pipeline is deploying Kiali and running the tests against it, the Kiali CR will now need to include setting validation_reconcile_interval to a lower value than the default.

This is to be done to Jenkins pipeline too (both upstream and downstream). 

Hayk and me have implemented different fixes for that Cypress test failing in Openshift (https://github.com/kiali/kiali/pull/8130 and https://github.com/kiali/kiali/pull/8142), but I think the definitve change is to set the `validation_reconcile_interval` to a lower value in the Kiali CR in Jenkins pipeline. This setting also affects the CRD validation cypress tests, increasing the total test suite runtime by 20 minutes in Jenkins

